PR #1716, opened 2019-09-10
writes -P profile into a txt file in the pier

joemfb:
The `-P` argument makes urbit send itself `SIGPROF` every 10ms. Sample profile data is interpolated into a structure on the home road, which is summarized and printed when urbit exits gracefully. (Note that this profile will be useless without the CPU_DEBUG build option).

As of `v0.8.0`, the profiling became much more awkward. The multi-process shutdown doesn't include affordances for sending the profiling data over IPC, so it just prints to the terminal after urbit appears to have shutdown. This PR changes the profiling printout, unconditionally writing it to a file in the pier (at `.urb/put/profile/~DATE.txt`).

### DIFF
--- a/pkg/urbit/include/noun/trace.h
+++ b/pkg/urbit/include/noun/trace.h
@@ -112,7 +112,7 @@
     /* u3t_damp(): print and clear profile data.
     */
       void
-      u3t_damp(void);
+      u3t_damp(FILE* fil_u);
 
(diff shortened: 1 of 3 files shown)


jtobin:
@joemfb I've reworded the commit description and applied it as a124dfe61f01712ad608c8e7327965d94f073bfa (I find the component prefixes useful as a minor sanity check when cherry-picking, so like to have them there).

PR #1717, opened 2019-09-10
removes extra allocations from u3a_atom->gmp and reverse conversions

joemfb:
Our conversions to and from GMP structs and our indirect atoms (a heap-allocated array of `uint_32_t`) used `alloca()` and copied twice out of an apparent abundance of caution. This PR removes the unnecessary intermediate copy from both conversions.


jtobin:
I've reworded the commit description to add a `noun: ` prefix and have committed as d0c26edd24d8ece227991222d58121eb5daa05bd.

PR #1719, opened 2019-09-10
fixes U3_EVENT_TIME_DEBUG event-timing printfs

joemfb:
I introduced a scoping issue at some point when refactoring these printfs. (Maybe we should turn them on CI ... Not an isolated problem, we have way too many `#ifdef`'s.)

### DIFF
--- a/pkg/urbit/worker/main.c
+++ b/pkg/urbit/worker/main.c
@@ -506,15 +506,13 @@ _worker_work_live(c3_d evt_d, u3_noun job)
   u3A->now  = u3k(now);
 
 #ifdef U3_EVENT_TIME_DEBUG
-  {
-    struct timeval b4, f2, d0;
-    gettimeofday(&b4, 0);
+  struct timeval b4, f2, d0;
+  gettimeofday(&b4, 0);
 
-    if ( c3__belt != u3h(u3t(ovo)) ) {
-      c3_c* txt_c = u3r_string(u3h(u3t(ovo)));
+  if ( c3__belt != u3h(u3t(ovo)) ) {
+    c3_c* txt_c = u3r_string(u3h(u3t(ovo)));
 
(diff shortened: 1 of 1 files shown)


jtobin:
(Reworded description to add a `worker: ` prefix and applied as a2dcabd38e3614120c3cf4f9183f1cdbcf688cb1.)

joemfb:
If this convention is going to be enforced, please have an associated taxonomy of labels.

Also, I'd really rather we don't make a pattern of closing pull requests and merging changes out of band. It makes it harder to browse the repository history.

jtobin:
On Wed, Sep 11, 2019 at 10:01:37AM -0700, Joe Bryan wrote:
> If this convention is going to be enforced, please have an associated
> taxonomy of labels.

IMO these don't have to be super precise (as in, drawn from a specific
list).  I think that would make contributing unnecessarily onerous.  I'm
mostly concerned about a rough pointer to what is being touched in a
given commit.  Stuff like "vere", "arvo", "eyre", "chat", etc. works
fine -- a judgment call here should typically be sufficient.

The reason for the prefix is that 1) it provides a rough but useful
piece of information when browsing or cherry-picking individual commits,
and, similarly, 2) it makes for more useful and informative release
notes when assembled with 'git shortlog', as I've been doing as of late
(and hope to do more of).

> Also, I'd really rather we don't make a pattern of closing pull
> requests and merging changes out of band. It makes it harder to browse
> the repository history.

Fear not -- I don't usually do this, nor do I expect it will become a
habit.  In this case it was to reword your commits, but I'll ask you to
do that in the future.

philipcmonk:
Agree that creating such a taxonomy is probably not worth it, especially since it's likely to attract a lot of bikeshedding.

You can still convince Github the PR was merged if you force-push the rebased branch here to the same name.

PR #1722, opened 2019-09-11
u3: rewrites +muk jet to skip heap allocations where possible

joemfb:
In the spirit of #1717, I've been looking for other places where we allocate more frequently than is necessary. The jet for `+muk` (murmur32) is a good example.

Along the way, I realized some of the preconditions in the `+muk` gate appear to be spurious. At least, they should be: `+muk` as implemented can only hash up to 31 bytes, and we routinely need to hash more than that for `+mug`. This condition effectively creates a jet mismatch in `+mug`, which formally calls `+muk`. (The byte-length assertion feels overly restrictive, but it's not the same level of problem.

Rewriting the jet was trivial, so I went ahead and did so. Confirming that `+muk` is correct without these assertions is less so, so I've just added an explanatory comment. I'll open an issue to track this.

/cc @frodwith

### DIFF
--- a/pkg/urbit/jets/c/muk.c
+++ b/pkg/urbit/jets/c/muk.c
@@ -6,59 +6,77 @@
 
 /* functions
 */
-  u3_noun
-  u3qc_muk(u3_atom seed,
-           u3_atom len,
-           u3_atom key)
+u3_noun
+u3qc_muk(u3_atom seed,
+         u3_atom len,
+         u3_atom key)
+{
+  c3_w key_w = u3r_met(3, key);
+
+  //  XX revisit
+  //
+  //    the first two conditions are explicitly asserted in +muk;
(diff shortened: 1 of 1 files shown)


belisarius222:
Does the byte-length restriction mean that the interpreter would bail:fail or otherwise crash if we tried to use a large atom as a map key or compare two large atoms for equality?  If so, I'm a little surprised that this doesn't happen routinely.  If not, I'm not sure I understand the situation properly.

joemfb:
That's right.

belisarius222:
Ok, that makes sense.  I hadn't realized this code wasn't being run.

—
~rovnys-ricfer
https://urbit.org

joemfb:
It's used for `@p` scrambling in `+ob`, but I think that's it.

PR #1731, opened 2019-09-13
Lockfile

benjamin-tlon:
Implemented lockfile with correct handling of SIGINT (^C) and SIGTERM (sent by `kill`).

### DIFF
--- a/pkg/king/app/Main.hs
+++ b/pkg/king/app/Main.hs
@@ -87,7 +87,7 @@ import UrbitPrelude
 import Data.RAcquire
 
 import Arvo
-import Control.Exception hiding (evaluate, throwIO)
+import Control.Exception (AsyncException(UserInterrupt))
 import Data.Acquire
 import Data.Conduit
 import Data.Conduit.List hiding (replicate, take)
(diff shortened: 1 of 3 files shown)


benjamin-tlon:
How do you mean? @orphaned lockfiles?

Right now, we just error if the lockfile exists. The existing code kills the other executable first, and then starts. We can easily implement that behavior here, if you think that's worth doing.=

Automatically killing the existing process seems like questionable behavior to me, but it might be the right thing. What do you think @joemfb?

joemfb:
It makes sense when urbit is just a CLI program, but seems a little harder to justify if it's to be a long-running daemon. Enforcing a single daemon-process per pier is obviously important, but I don't (currently) have strong feelings about exactly how that's done.

eglaysher:
When I talk about orphaned lockfiles, I'm talking about when a hard shutdown happens on the computer running Urbit causing a reboot. My understanding is that good lockfile implementations write the pid of the current process to the lockfile, and when another process starts up and tries to acquire the lock, it reads the lock file to see if there's a pid with that process id first and if not, that's a sign that the lockfile is an orphan. (The current Urbit gets this right, though it handles this by also trying to SIGTERM the other process, which is arguably wrong.)

Or: the Acquire monad isn't safe to power cords.

PR #948, opened 2018-03-19
Charm

frodwith:
This marks the third attempt to slay the ford cache wiping errors.

PR #941, opened 2018-03-06
Runtime overflows

frodwith:
Resolves issue #936.

PR #1741, opened 2019-09-14
u3: unifies unifying-equality unification

joemfb:
u3 has three difference noun-equality functions, which overlap non-obviously in their behavior. This PR combines them all into one, and refactors that one to be (hopefully) clearer.

I won't attempt to exhaustively characterize the old behavior. Here is the state of things as of this PR:

- one noun-equality function, `u3r_sing()`
- never unify on the home road
- always unify on inner roads
- recurse up into senior roads to unify as necessary (unless GC is enabled)

With regards to that last item: if either noun pointer is on our road, we decrement that reference (potentially freeing a noun) and use the other (always trying to preserve the noun that's deeper in the heap). If neither noun pointer is on our road, we recurse up into senior roads, as we want to unify wherever it's possible. We can't free senior memory, as there may be uncounted junior references to it (this is how the nested road model intersects with reference counting), so we're forced to leak a reference when we unify in senior memory. Ignoring for a moment the affect on memory pressure, it's safe to leak references on inner roads, as the memory will be reclaimed when we return to the home road. But we don't recurse into senior memory when GC is enabled; we obviously can't leak if anyone's watching.

The old implementation tended to also (temporarily) leak references on the current road, and the different implementations did not all recurse into senior roads. This likely had a minor performance benefit, as decrementing reference counts is not free, but it obviously increased memory pressure. There's no way to reclaim this leaked memory without exiting from the road in question, so we can't speculatively leak and then cleanup when we run out of memory. As we frequently have issues with memory pressure, this doesn't seem to be the right tradeoff for now. We may eventually want this behavior in certain contexts, in the same way that we likely want a simple bump-allocator for bounded computational contexts.

I have not yet profiled this PR against the master branch in detail. Boot and basic operations appear roughly unchanged when timing them manually (for instance, booting with these changes appears to take 1-2s longer than without).

Memory pressure appears to be significantly improved during boot with normal build options (https://pastebin.com/hNwvy0fy), but only slightly improved with GC enabled (https://pastebin.com/h5kJKQjj). This suggests that unification in senior memory is quite important.

There are a handful of ways that our unifying equality could still be improved:

- we don't (directly) unify indirect atoms
- we don't unify on the home road
- we depend on (temporary) memory leaks for unification in senior memory

We currently only unify the head and tail pointers of cells. We could defer the unification of indirect atoms in a couple ways: by maintaining a hashtable from an atom to it's preferred replacement, or by adding a replacement pointer to the underlying struct (this could also be used for leak-free senior unification). It's not clear if the time/space tradeoffs of either option are worthwhile, more measurement and testing is needed.

It seems like it should be safe to unify in the home road in the worker process, as we have very few noun references and can track them. This is not the case in the daemon, where it would be much harder to be sure. And our sample profiling `-P` jumps between an inner road and the home road to interpolate samples. (It's not at all clear to me how it can be safe to allocate on the home road in a signal handler while a computation is pending in an inner road).

And we could stop depending on memory leaks for senior unification by keeping a to-be-freed collection of some sort on a road. (This might require allocating in senior memory, which again, is n't something I know how to do safely. But we are already doing it in a limited capacity).

This PR improves the current situation by cleaning up the implementation and limiting memory pressure, and the cost of some incremental compute performance. I'll try to better quantify that performance cost with some additional profiling.

/cc @frodwith


joemfb:
I should note that @frodwith replaced the original recursive unifying-equality with an explicit stack (and memoization for especially-large comparisons) in #920, and refined that implementation in #941 and #948.

One more needed: we likely need interrupt/signal exclusion in this unification implementation, as it mutates the road pointer.

As the CI, I'm investigating the PR build failure, which looks like a real SIGSEGV. The push build looks like a successful build that just doesn't terminate, which we've seen periodically. But I've restarted it a couple of times and it's hung every time. Not sure what to make of that ...

joemfb:
The CI problems here are real: the short answer is that we depend on "unify-and-leak" semantics to  accurately promote values from inner roads. #1753 resolves this, and will unblock these changes.

joemfb:
With #1753 merged, this PR is now ready for review.

pkg/urbit/noun/allocate.c
```diff
@@ -1492,6 +1492,103 @@ u3a_use(u3_noun som)
...
+  c3_t bsr_t = ( c3y == u3a_is_senior(rod_u, *b) );
+  c3_t nor_t = ( c3y == u3a_is_north(rod_u) );
+  c3_t own_t = ( rod_u == u3R );
+
+  //  both are on [rod_u]; gain a reference to whichever we keep
+  //
+  if ( !asr_t && !bsr_t ) {
+    //  keep [a]; it's deeper in the heap
+    //
+    //    (N && >) || (S && <)
```
belisarius222: This comment is more like what I was expecting the code to do.  Doesn't heap depth change direction based on whether the road is north or south?

pkg/urbit/noun/allocate.c
```diff
@@ -1492,6 +1492,103 @@ u3a_use(u3_noun som)
...
+void
+u3a_wed(u3_noun* a, u3_noun* b)
+{
+  if ( *a != *b ) {
+    u3_road* rod_u = u3R;
+
+    //  while not at home, attempt to unify
+    //
+    //    we try to unify on our road, and retry on senior roads
+    //    until we succeed or reach thee home road.
```
belisarius222: thee -> the

pkg/urbit/noun/allocate.c
joemfb: The condition on the next line is `( (*a > *b) == nor_t )`, ie `( ((*a > *b) && nor_t) || (!(*a > *b) && !nor_t) )`.

pkg/urbit/noun/allocate.c
joemfb: Reach for thine home road

pkg/urbit/noun/allocate.c
belisarius222: Oh, right.  I misread that.

pkg/urbit/noun/retrieve.c
```diff
@@ -301,115 +248,25 @@ _eq_pop(c3_ys mov, c3_ys off)
...
+_song_atom(u3_atom a, u3_noun b)
 {
   u3a_atom* a_u = u3a_to_ptr(a);
 
-  if ( !_(u3a_is_atom(b)) ||
-      _(u3a_is_cat(a)) ||
-      _(u3a_is_cat(b)) )
+  if ( (c3n == u3a_is_atom(b)) ||
+       (c3y == u3a_is_cat(a)) ||
+       (c3y == u3a_is_cat(b)) )
```
belisarius222: How could `b` be a cat if it's not an atom?  Is this for `u3_post`s?

belisarius222: or did I mix up dogs and cats ...

joemfb:
@belisarius222, I pushed a couple more comments and a little refactoring. PTAL.

belisarius222:
LGTM

PR #1746, opened 2019-09-17
Makes a progress bar on booting a new ship.

eglaysher:
Also prints the name of the ship you're trying to boot, along with
whether its a fake or not. Also fixes a regression where we were
trying to acquire a lockfile before the pier directory was created.


benjamin-tlon:
Merging this so that I can make changes on top of it.

PR #1747, opened 2019-09-17
Make king shutdown cleanly on a new line.

eglaysher:
This makes canceling the persistance thread not rethrow an exception
which kills the process, and makes the terminal driver write one
final newline before giving control back to the terminal so the
bash prompt writes to its own line.

### DIFF
--- a/pkg/king/lib/Vere/Pier.hs
+++ b/pkg/king/lib/Vere/Pier.hs
@@ -7,18 +7,18 @@ module Vere.Pier
 import UrbitPrelude
 
 import Arvo
-import Vere.Pier.Types
 import System.Random
+import Vere.Pier.Types
 
 import System.Directory   (createDirectoryIfMissing)
 import System.Posix.Files (ownerModes, setFileMode)
 import Vere.Ames          (ames)
 import Vere.Behn          (behn)
+import Vere.Clay          (clay)
 import Vere.Http.Client   (client)
 import Vere.Http.Server   (serv)
 import Vere.Log           (EventLog)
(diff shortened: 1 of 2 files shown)


eglaysher:
A sorta question: it still takes several seconds between `^D` and getting back to the bash prompt. All the `cancel` calls on `Async`s happen serially since `cancel = throwTo (asyncThreadId a) AsyncCancelled <* waitCatch a`.

One thing Chrome did was have a two-phase shutdown where every subsystem received a start shutting down signal, and then after everything had been sent that, only then did it start blocking on each service to shutdown. How hard would it be to make an RSAcquire, which took a `start`, a `signal`, and a `stop`? `signal` would perform a `throwTo` to each thread, and `stop` would perform the `waitCatch`, allowing for a more parallel shutdown.

benjamin-tlon:
I don't think `RSAcquire` is the right abstraction, but something like that would be possible. Keep a set of things that can be eagerly asked to shutdown, and do those all eagerly.

I suspect that the sequential nature of the shutdown sequence is not the problem here. More, it's just that something is taking too long to shut down. It's probably `Serf` that's the problem.

benjamin-tlon:
The right thing here is not to replace `cancel >> wait` with `cancel` (we want the persist thread to be stopped before we tear down the event log).

How about `cancel >> waitCatch`? That way we wont re-throw.

eglaysher:
As mentioned in person, cancel is `throw ... <* waitCatch`. Also, fixed merge conflict.

eglaysher:
(Ben approved this over the shoulder.)

PR #1748, opened 2019-09-18
Hacky External Terminals

benjamin-tlon:
This branch implements external terminals in a very hacky way:

- Record all terminal events since start of KH process.
- When external terminals connect, dump all recorded events to external terminal.

In effect, this allows multiple client connections to *the same* terminal.

This is not ideal, obviously. I think this is a good idea as a short-term hack because:

- I need external terminal support in order to make the king a daemon.
- I need the king to be a daemon in order to support the client flow.
- Implementing external terminal support is hard.
- This approach bites off half the work needed to properly implement external terminals.

### DIFF
--- a/pkg/king/app/CLI.hs
+++ b/pkg/king/app/CLI.hs
@@ -68,6 +68,7 @@ data Cmd
     = CmdNew New Opts
     | CmdRun Run Opts
     | CmdBug Bug
+    | CmdCon Word16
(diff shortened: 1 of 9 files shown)


benjamin-tlon:
@eglaysher said "[something something] LGTM"

ohAitch:
Hey it's almost like tmux ;)

PR #1751, opened 2019-09-19
chat: message processing time down to 60ms

loganallenc:
I've been doing performance optimization of the Chat app in my la-odyssey branch within my new services architecture. After doing some profiling, I realized message processing times were abysmally slow (300ms per message). 90% of that time was spent doing Ford mark conversion.

On the odyssey branch, I was able to 10x performance (down to 30ms) by removing mark conversion and making this simple modification to Eyre that removes the ability to send pokes across ships from the browser. We don't use that functionality anyway, so I think this is an acceptable change.

Today I backported the changes I made in Odyssey to Hall and the current chat app. Not a 10x performance improvement, but 5x is still enough to warrant including in an OTA. Would love to see this come out soon in an OTA!


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1356,7 +1356,7 @@
           ^-  move
           :^  duct  %pass  /channel/poke/[channel-id]/(scot %ud request-id.i.requests)
           =,  i.requests
-          [%g %deal `sock`[our ship] `internal-task:gall`[app %punk mark %json !>(json)]]
```
ixv: This change will break our other web apps that have not been similarly ported to this paradigm. What you should probably do instead is implement an alternative poke api in channel.js and make both options available.

loganallenc:
From re-doing tests, looks like the %cast mark conversion from JSON to JSON happening within %punk takes very little time. On odyssey, still at 30ms even with %punk.

philipcmonk:
Yeah, a conversion from a mark to itself should not even build the mark, so I wouldn't expect that to slow it down.

PR #1753, opened 2019-09-20
u3: hashtable improvements, road promotion refactor

joemfb:
This PR was motivated by problems encountered in #1741. In short: promoting values out of inner roads is complicated, and was made much more complicated by our use of hashtables. (This manifested in #1741 as an existential dependence on leaking references while unifying nouns.)

This PR phases transitions from inner to outer roads; first we promote them (and don't do anything else with them), then we do whatever we need to do with them. Specifically, there are no hashtable lookups or equality comparisons during the promotion phase (`u3a_take()` and friends). To do this, I added hashtable promotion (`u3h_take()` and parameterized `u3h_take_with()`) and merge (`u3h_uni()`). I also refactored the internals of the bytecode interpreter and jet dashboard program/registration/cache promotion, splitting into these phases ("take" and "reap").

Along the way, I performed some minor refactoring of the hashtable implementation. Once this PR is merged, I'll update #1741 to include these changes. We'll then be able to remove the `u3h_gut()`, which will be entirely equivalent to `u3h_get()`.

I'm deeply indebted to @frodwith for his help in designing and making these changes.

### DIFF
--- a/pkg/urbit/include/noun/allocate.h
+++ b/pkg/urbit/include/noun/allocate.h
@@ -90,6 +90,16 @@
         u3p(struct _u3a_fbox) nex_p;
       } u3a_fbox;
 
+    /* u3a_jets: jet dashboard
+    */
+      typedef struct _u3a_jets {
+        u3p(u3h_root) hot_p;                  //  hot state (home road only)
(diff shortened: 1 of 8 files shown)


joemfb:
Merging so I can unblock #1741

PR #1754, opened 2019-09-24
dill: vane verbosity levels

belisarius222:
Implements three error reporting levels for Dill's `%crud` task:
- `%hush`: no output
- `%soft`: one line of output, just containing the `@tas` reason for failure
- `%loud`: full output

There is a new `|knob {error-tag} {level}` generator to reset verbosity levels.  All error tags default to `%loud` except for Ames `%hole`, which defaults to `%soft` so that packet spam doesn't clutter up the console.


philipcmonk:
You should be able to trigger a crash in other vanes by either ^C-ing an event sent to them (eg a sync to clay) or introducing a `!!` in them.  If you want a nondeterministic error specifically, add a `(bex (bex 256))` and ^C it.

The other vanes handle their own cruds and forward them to dill as `%flog` events.  You may need to know specifically that it's an error if `%flog` is used for other stuff.  In that case, add another case, eg `%clog` to represent an error.

philipcmonk:
Note this should correctly handle errors indirected through behn, since those errors are sent to the vane, which then usually `%flog`s them to dill.

philipcmonk:
Looks like CI was had connectivity issues to nix or something.  Restarted the builds and they're looking good.

belisarius222:
Ok, I'll try to get a Behn event to crash.

Thanks for restarting the build.  I tried that a few times last night, to no avail.  If it's working now, it must have been a connectivity error or something like it.

belisarius222:
I wrote this to only limit Dill's %crud tasks, which are often wrapped in %flog tasks.  A %flog %text, for example, shouldn't be truncated.  Fortunately error reporting seems to consistently use %crud as it should (except for Gall, which just +slog's all its own errors).

philipcmonk:
Ah, I see how that works.  But how can you distinguish which vane sent you the error?  Doesn't the %flog unwrapping obscure that?

belisarius222:
I inspect the top wire of the duct.  This should work except in cases where the vanes %slip a %crud to Dill, which they often do.  It should be easy to change that to a %pass, though, if that doesn't break anything.

—
~rovnys-ricfer
https://urbit.org

belisarius222:
Behn->Ames %crud's come into Dill on Unix ducts, it turns out, because Ames %slip's them to Dill.

philipcmonk:
Doesn't the `%flog` unwrapping in rewrite the duct to `u.hey.hall` anyway?

belisarius222:
Well, to be more precise, when Behn sends a %wake with an error to Ames (because Ames or something downstream had crashed on a previous attempt to deliver the %wake), Ames %slip's a %crud to Dill.

belisarius222:
Ach, you're right.

Not sure how to handle this, other than to think looking into the ducts was a mistake.  Maybe each vane should just send its letter to Dill manually in the card?  It seems a little wrong, but it's not magic and won't fall prey to these kinds of idiosyncrasies.

philipcmonk:
Putting a term in the card seems fine.  It's also an easy way to get more precise verbosity if we want (eg give certain errors a tag not equal to their vane name).

belisarius222:
Alternatively, we could filter based on the error tag instead of the vane sender.  That would be %wake for Ames wakeup failures.  Not sure what tag comes out of Ames packet cruds; I could check in Vere to find out.

belisarius222:
Dill's %crud task already has a `@tas` tag in addition to the tang, which is what I was referring to as the error tag.

belisarius222:
I'll do it this way and set the Ames tags that come from Unix to quiet by default.

belisarius222:
Rewrote it this way, PTAL.

Also note that Vere still produces the `worker_send_replace 230 crud` messages no matter what Arvo does.

belisarius222:
Also note that I did not write a state adapter for Dill, assuming we'll breach this in pretty soon.

jtobin:
@belisarius222 Want to fixup/squash those intermediate commits as some rebase training?

(I can also do this for you -- I already did locally, like this:

```
$ git rebase -i 442fdc3fe
```



plus an additional fixup, resulting in the following commits:

```
f12a34cdf zuse, ames, dill: look at error tag, not duct
e89a64bfa hood, dill: add support for unix %cruds
24194d3e9 hood, gall: |knob {vane} {level} to adjust verbosity
f9c1491d6 dill: fix bugs with vane verbosity
2f576a410 dill: three log levels
```

if you don't update this branch today I'll just merge the one I've rebased in.)

belisarius222:
It should really be two commits: one for Dill and associated changes in Zuse and Gall, and another that adds the |knob generator to Hood.  I can do that today.

—
~rovnys-ricfer
https://urbit.org

belisarius222:
@jtobin I rewrote history.  PTAL

jtobin:
:+1: I added a 'hood' prefix to the second commit and will sort out the pill conflict.

PR #1757, opened 2019-09-24
Added tank and path json encoders, array to set, and string as @tas to zuse/hoon

loganallenc:
I found myself copying these functions over and over into new JSON libs inside of the odyssey branch, so I'd love to get these included as standard library additions instead of having to do that!

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -5493,6 +5493,16 @@
       |=  a/^time
       =-  (numb (div (mul - 1.000) ~s1))
       (add (div ~s1 2.000) (sub a ~1970.1.1))
+    ::                                                  ::  ++path:enjs:format
+    ++  path                                            ::  string from path
+      |=  a=^path
+      ^-  json
+      [%s (spat a)]
+    ::                                                  ::  ++tank:enjs:format
+    ++  tank                                            ::  tank as string arr
+      |=  a=^tank
+      ^-  json
+      [%a (turn (wash [0 80] a) tape)]
     --  ::enjs
   ::                                                    ::  ++dejs:format
   ++  dejs                                              ::  json reparser
@@ -5506,6 +5516,10 @@
       |=  jon/json  ^-  (list _(wit *json))
       ?>  ?=({$a *} jon)
       (turn p.jon wit)
+    ::                                                  ::  ++as:dejs:format
+    ++  as                                              ::  array as set
+      |*  a=fist
+      (cu ~(gas in *(set _$:a)) (ar a))
     ::                                                  ::  ++at:dejs:format
     ++  at                                              ::  array as tuple
       |*  wil/(pole fist)
@@ -5630,6 +5644,9 @@
       |*  {a/cord b/*}
       =>  .(+< [a b]=+<)
       [(rash a fel) b]
+    ::                                                  ::  ++pa:dejs:format
+    ++  pa                                              ::  string as path
+      (su ;~(pfix net (more net urs:ab)))
     ::                                                  ::  ++pe:dejs:format
     ++  pe                                              ::  prefix
       |*  {pre/* wit/fist}
@@ -5637,6 +5654,11 @@
     ::                                                  ::  ++sa:dejs:format
     ++  sa                                              ::  string as tape
       |=(jon/json ?>(?=({$s *} jon) (trip p.jon)))
+    ::                                                  ::  ++se:dejs:format
+    ++  se                                              ::  string as aura
+      |=  aur=@tas
+      |=  jon=json
+      ?>(?=([%s *] jon) (slav aur p.jon))
     ::                                                  ::  ++so:dejs:format
     ++  so                                              ::  string as cord
       |=(jon/json ?>(?=({$s *} jon) p.jon))
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:85de6b4640a1dddeda908e6d111e2f6139447d376db3bdb4573f5bc370ddce82
-size 15756068
+oid sha256:6205835e39ce2a3f995c042ac3b46a34e492c464db20b6e7604b08b828744215
+size 15831246
--- a/pkg/arvo/lib/hall-json.hoon
+++ b/pkg/arvo/lib/hall-json.hoon
@@ -62,13 +62,8 @@
     =-  (pairs (turn ~(tap by a) -))
     |*  {k/* v/*}  [(b k) (c v)]
   ::
-  ++  lank                                              :::  tank as string arr
-    |=  a/tank
-    ^-  json
-    a+(turn (wash [0 80] a) tape)
-  ::
   ++  dank                                              :::  tank
-    |=  a/tank
+    |=  a=^tank
     ?:  ?=($leaf -.a)  (frond %leaf (tape p.a))
     ::  TODO: No idea what a plum is, but it was recently added
     ::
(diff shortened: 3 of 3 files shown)


ixv:
A dejs conversion for paths would be nice too.

pkg/arvo/sys/zuse.hoon
```diff
@@ -5618,6 +5632,9 @@
     ::                                                  ::  ++sa:dejs:format
     ++  sa                                              ::  string as tape
       |=(jon/json ?>(?=({$s *} jon) (trip p.jon)))
+    ::                                                  ::  ++se:dejs:format
+    ++  se                                              ::  string as @tas
+      |=(jon=json ?>  ?=([%s *] jon) (scot %tas p.jon))
```
joemfb: This seems like a bit of a strange operation -- extracting a JSON string, then serializing it as `@tas`. Can you elaborate on your use-case(s)? It seems like it'd be more common to want to parse the extracted string with `+slav`.

pkg/arvo/sys/hoon.hoon
```diff
@@ -6244,6 +6244,7 @@
   ~/  %scow
   |=(mol/dime ~(rend co %$ mol))
 ++  slat  |=(mod/@tas |=(txt/@ta (slaw mod txt)))
+++  slan  |=(mod/@tas |=(txt/@ta (need (slaw mod txt))))
```
joemfb: Is this intended for use with `+se:dejs:format`?

loganallenc:
Added dejs conversion for paths @ixv.

pkg/arvo/sys/hoon.hoon
loganallenc: No, this is intended for use with this sort of formatter. I used this 3 times, so I figured put it in zuse.

https://github.com/urbit/urbit/blob/ff7b2ab7d4f8e4c8daf28b474482b6874301effd/pkg/arvo/lib/chat-json.hoon#L14

loganallenc: https://github.com/urbit/urbit/blob/ff7b2ab7d4f8e4c8daf28b474482b6874301effd/pkg/arvo/lib/invite-json.hoon#L9

pkg/arvo/sys/zuse.hoon
loganallenc: Use case is simple: I want to go from JSON to a dock, or to an app-name, etc.

https://github.com/urbit/urbit/blob/ff7b2ab7d4f8e4c8daf28b474482b6874301effd/pkg/arvo/lib/invite-json.hoon#L152

loganallenc: (note the name is different in that library)

joemfb: Do you have some kind of parsing or validation in mind? What do you want to happen if you're sent JSON that doesn't conform to the "symbol constraint" of `@tas` (ie, lower-case alphanumeric, leading alpha)?

pkg/arvo/sys/hoon.hoon
belisarius222: These functions don't look wrong, but +slan, +slav, and +slat all look like cruft to me.  +slaw is the only function that has any meat here; the others either curry it, wrap it in an assertion, or both.

I feel like we'd be better served by a combinator that wraps a function in a +need call.  Maybe +slaw should also be defined curried, since uncurrying is (at least a little) more convenient than currying.  That way if you're composing these functions into your own combinators, you can stay functional, but we won't need so many functions that are almost exactly the same.

belisarius222: Something like this (unless we "bake" it):
```
++  want
  |*  f=$-(* (unit))
  |*  x=*
  (need (f x))
```

pkg/arvo/sys/zuse.hoon
```diff
@@ -5474,6 +5475,16 @@
...
+    ::                                                  ::  ++path:enjs:format
+    ++  path                                            ::  string from path
+      |=  a=^path
+      ^-  json
+      [%s (spat a)]
+    ::                                                  ::  ++tank:enjs:format
+    ++  tank                                            ::  tank as string arr
+      |=  a=^tank
+      ^-  json
+      [%a (turn (wash [0 80] a) tape)]
```
belisarius222: Looks like +wash produces a $wall already.  We shouldn't have to coerce to $tape for each line in the $wall, which will have O(n) performance.

joemfb: It's calling `+tape:enjs:format`, which will render each line to a JSON string inside thiis array.

pkg/arvo/sys/hoon.hoon
joemfb: Or, you know, fix currying and function composition.

pkg/arvo/sys/zuse.hoon
```diff
@@ -5611,13 +5626,19 @@
...
+      (su ;~(pfix net (more net urs:ab)))
     ::                                                  ::  ++pe:dejs:format
     ++  pe                                              ::  prefix
       |*  {pre/* wit/fist}
       (cu |*(* [pre +<]) wit)
     ::                                                  ::  ++sa:dejs:format
     ++  sa                                              ::  string as tape
       |=(jon/json ?>(?=({$s *} jon) (trip p.jon)))
+    ::                                                  ::  ++se:dejs:format
+    ++  se                                              ::  string as @tas
```
belisarius222: It seems a little strange to me to hard-code the aura here, as opposed to letting it be passed in as a parameter.  Something like this (I think it could stay dry).
```
++  se
  |=  aura=@tas
  |=  jon=json
  ?>  ?=([%s *] jon)
  (scot aura p.jon)
```

pkg/arvo/sys/zuse.hoon
joemfb: @loganallenc, to be clear, this seems wrong regardless of your use-case. I don't see any reason why you would want to serialize inside your (re-)parser here. In the valid case, the serialization is an identity function, and in the invalid case, it's *still* an identity function (see `(scot %tas '1FOO')`). I could see asserting `(sane @tas)`, or sanitizing the input to make it a valid `@tas`.

joemfb: /cc @belisarius222

pkg/arvo/sys/zuse.hoon
loganallenc: Great idea. Thanks!

pkg/arvo/sys/zuse.hoon
loganallenc: Will switch to using `slav`.

pkg/arvo/sys/zuse.hoon
loganallenc: Any reason why you wouldn't just specify it as

`
++  se
  |=  [aura=@tas jon=json]
  ?>  ?=([%s *] jon)
  (slav aura p.jon)
`
?

loganallenc: Formatting was messed up, but the idea is why the second gate?

loganallenc: Nevermind, Joe explained its use in function combinators such as cu.

loganallenc: Completed this.

pkg/arvo/sys/hoon.hoon
joemfb: @loganallenc, is this use case still relevant given the now-general form of `+se:dejs:format`?

loganallenc: @joemfb no longer relevant. Have removed it from my most recent push.

Should be ready to accept or review again at this point, addressed all listed concerns.

@belisarius222

pkg/arvo/sys/zuse.hoon
belisarius222: oh ok

pkg/arvo/sys/hoon.hoon
belisarius222: > Or, you know, fix currying and function composition.

I suspect the trickiest part would be letting wet gates nest inside dry gates, or equivalent functionality.  Makes think fully-baked Hoon might be a good idea ...

loganallenc:
Modified commit message.

jtobin:
I've rebased against master and updated the pill and get a bunch of test failures as a result.  Which seems strange, as going from the diff alone these just appear to be additive changes.  Investigating..

loganallenc:
Thanks for testing this Jared. I'll look into this further and fix any problems being caused.

PR #1758, opened 2019-09-24
Removed Hall references from DNS apps

loganallenc:
Added TODOs for adding a systems notification service to replace its notify functions.

### DIFF
--- a/pkg/arvo/app/acme.hoon
+++ b/pkg/arvo/app/acme.hoon
@@ -1,4 +1,4 @@
-/-  asn1, hall
+/-  asn1
 /+  base64, der, primitive-rsa, *pkcs, *jose
 =,  eyre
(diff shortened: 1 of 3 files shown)


philipcmonk:
Maybe at least replace with `+slog` for now?

joemfb:
Is there some reason :hall *needs* to be removed before all it's use-cases are replaced?

joemfb:
Of course, that question is for both this and #1761.

jtobin:
I won't merge this.  There simply isn't enough context in the PR, the change as it stands by itself is dubious (it basically just breaks stuff), etc.

If you have another PR or PRs in mind that would depend on this, you should just PR the lot of that work together, or at least in larger and more meaningful chunks.

PR #1760, opened 2019-09-24
Remove Hall references from Hood.

loganallenc:


pkg/arvo/lib/hood/drum.hoon
```diff
@@ -393,7 +386,7 @@
 ++  se-dump                                           ::  print tanks
   |=  tac/(list tank)
   ^+  +>
-  ?.  se-ably  (se-hall tac)
+  ?.  se-ably  !!
```
philipcmonk: Replace this with `((slog tac) +>.$)` so it gets printed somewhere.

ixv:
I still think we don't want to move forward with this until talk has been ported over.

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -471,21 +464,13 @@
...
-  ::(se-emit 0 %poke /drum/hall [our.hid %hall] (said:hall our.hid %drum now.hid eny.hid tac))
-::
 ++  se-text                                           ::  return text
   |=  txt/tape
   ^+  +>
   ?.  ((sane %t) (crip txt))  :: XX upstream validation
     ~&  bad-text+<`*`txt>
     +>
-  ?.  se-ably  (se-hall [%leaf txt]~)
+  ?.  se-ably  !!
```
philipcmonk: Oh also here

belisarius222:
@loganallenc Looks like you're removing old Hall before adding new Hall.  I feel like it would be better git-wise to do the opposite, since there would hopefully be fewer intermediate states of the repo that don't work.  Not sure if this is actionable advice at present.

pkg/arvo/lib/hood/drum.hoon
loganallenc: Updated. Thanks Phil.

jtobin:
> Looks like you're removing old Hall before adding new Hall. I feel like it would be better git-wise to do the opposite, since there would hopefully be fewer intermediate states of the repo that don't work.

+1.  Propose one atomic change, or several more useful atomic changes.  I don't think this is useful as is, and won't merge it.

Additionally, you need to add more context to these PRs -- I shouldn't "just have to know" that you have other work planned, basically.  Describe the changes, reference related material, etc.  I would argue that you should be treating a PR to Urbit in much the same way as if you were making a PR to Nix, git, GHC, bitcoin, the Linux kernel, etc.  None of those projects would accept something like this, and we shouldn't either.

loganallenc:
I had previously removed Hall and Talk from the apps that are started up upon the ship starting. I have now updated this branch to remove Hall dependencies from Hood but Hall and Talk are still started up upon boot. This should be an acceptable change since Hood shouldn't depend on Hall anyway.

PR #1761, opened 2019-09-24
Remove Hall and Chat app, everything except Hall's sur file

loganallenc:


joemfb:
I'm opposed to this change, barring some very compelling explanation. `:talk` may be our most widely-used app. I understand a replacement is "eventually forthcoming", but I don't understand why we would need to remove it beforehand.

belisarius222:
I agree with @joemfb.  It seems to me the thing to do is add in the new set of apps, temporarily unused and using temporary names if they conflict with old names.  Then, once we're ready, do one atomic commit that switches from old to new, deleting old files, updating dispatchers, and renaming as necessary.

This seems like the kind of change that we should be able to perform without the repo ever going through a broken state.

philipcmonk:
If the goal is to individually review these changes, then another way to do this is to make these PRs into a staging branch, then PR the whole thing as one atomic PR.

loganallenc:
@joemfb Talk is not a priority in this upcoming breach, and can be added in as an OTA, since it is simply a CLI view on top of `chat-store`. @Fang- will be working on this shortly as soon as he’s back from vacation.

vvisigoth:
To be clear, whenever this goes in, it needs to have parity on the cli.
—
~poldec-tonteg
http://urbit.org

ohAitch:
Not a priority = usage of urbit('s most common app) after this breach is
projected to be low? :^)

loganallenc:
@belisarius222 Removing Chat functionality from master for the space of a single commit does not seem negligent to me. Chat-store is ready to be PRed and merged as soon as #1763 goes in. I do not think that we should have a policy of considering the removal of an app in an atomic manner from master equivalent to breaking master.

ohAitch:
"Removed in a breach, and will be missing until a future PR is composed and OTA'd" seems somewhat larger than "for the space a single commit". Are there downsides to @philipcmonk's "do so in a staging branch" proposal?

jpt4:
> Removing Chat functionality from master for the space of a single commit does not seem negligent to me.

No reason to introduce unnecessary chinks in the armor, and there seems a procedure in place to do achieve the same thing without even a temporary degradation.

Fang-:
Agreed with all of the above that this seems a bit premature. This should probably happen as part of the "re-implement hall as services" PR?

jtobin:
I won't merge this, for the reasons given above and similar to those I've given in #1760.

PR #1763, opened 2019-09-25
Added store, hook, and JSON marks for both Groups and Permissions

loganallenc:
Added working and tested stores, hooks, and JSON marks. These have TODOs in them to no-op in a few cases where they don't currently (they send an update even though they did nothing), but apart from that, these are done.

Also, once some version of #1757 goes in, the group-json and permission-json can be refactored to use the functions in standard lib instead of using their own implementations. In the meantime, putting this out for review.

### DIFF
--- a/pkg/arvo/app/group-hook.hoon
+++ b/pkg/arvo/app/group-hook.hoon
@@ -1,0 +1,259 @@
+::  group-hook: allow syncing group data from foreign paths to local paths
+::
+/-  *group-store, *group-hook
+|%
++$  move  [bone card]
+::
++$  card
+  $%  [%diff [%group-update group-update]]
+      [%quit ~]
+      [%poke wire dock [%group-action group-action]]
+      [%pull wire dock ~]
+      [%peer wire dock path]
+  ==
+::
++$  state
+  $%  [%0 state-zero]
+  ==
+::
++$  state-zero
+  $:  synced=(map path ship)
+      boned=(map wire (list bone))
+  ==
+::
+--
+::
+|_  [bol=bowl:gall state]
+::
+++  this  .
+::
+++  prep
+  |=  old=(unit state)
+  ^-  (quip move _this)
+  ?~  old
+    [~ this]
+  [~ this(+<+ u.old)]
+::
+++  poke-group-hook-action
+  |=  act=group-hook-action
+  ^-  (quip move _this)
+  ?-  -.act
+      %add
+    ?.  =(src.bol our.bol)
+      [~ this]
+    =/  group-path  [%group path.act]
+    =/  group-wire  [(scot %p ship.act) group-path]
+    ?:  (~(has by synced) path.act)
+      [~ this]
+    =.  synced  (~(put by synced) path.act ship.act)
+    :_  (track-bone group-wire)
+    ?:  =(ship.act our.bol)
+      [ost.bol %peer group-wire [ship.act %group-store] group-path]~
+    [ost.bol %peer group-wire [ship.act %group-hook] group-path]~
+  ::
+      %remove
+    =/  ship  (~(get by synced) path.act)
+    ?~  ship
+      [~ this]
+    ?:  &(=(u.ship our.bol) =(our.bol src.bol))
+      ::  delete one of our own paths
+      =/  group-wire  [(scot %p our.bol) %group path.act]
+      :_  this(synced (~(del by synced) path.act))
+      %+  weld
+        (pull-wire group-wire path.act)
+      ^-  (list move)
+      %+  turn  (prey:pubsub:userlib [%group path.act] bol)
+      |=  [=bone *]
+      ^-  move
+      [bone %quit ~]
+    ?:  |(=(u.ship src.bol) =(our.bol src.bol))
+      ::  delete a foreign ship's path
+      =/  group-wire  [(scot %p u.ship) %group path.act]
+      :_  this(synced (~(del by synced) path.act))
+      (pull-wire group-wire path.act)
+    :: don't allow
+    [~ this]
+  ::
+  ==
+::
+++  peer-group
+  |=  pax=path
+  ^-  (quip move _this)
+  ?~  pax  !!
+  ?.  (~(has by synced) pax)  !!
+  =/  grp=(unit group)  (group-scry pax)
+  ?~  grp  !!
+  :_  this
+  [ost.bol %diff [%group-update [%path u.grp pax]]]~
+::
+++  diff-group-update
+  |=  [wir=wire diff=group-update]
+  ^-  (quip move _this)
+  ?:  =(src.bol our.bol)
+    (handle-local diff)
+  (handle-foreign diff)
+::
+++  handle-local
+  |=  diff=group-update
+  ^-  (quip move _this)
+  ?-  -.diff
+      %keys    [~ this]
+      %path    [~ this]
+      %bundle  [~ this]
+      %add
+    :_  this
+    %+  turn  (prey:pubsub:userlib [%group pax.diff] bol)
+    |=  [=bone *]
+    ^-  move
+    [bone %diff [%group-update diff]]
+  ::
+      %remove
+    :_  this
+    %+  turn  (prey:pubsub:userlib [%group pax.diff] bol)
+    |=  [=bone *]
+    ^-  move
+    [bone %diff [%group-update diff]]
+  ::
+      %unbundle
+    :_  this(synced (~(del by synced) pax.diff))
+    %+  weld
+    ^-  (list move)
+    %+  turn  (prey:pubsub:userlib [%group pax.diff] bol)
+    |=  [=bone *]
+    ^-  move
+    [bone %diff [%group-update diff]]
+    ^-  (list move)
+    %+  turn  (prey:pubsub:userlib [%group pax.diff] bol)
+    |=  [=bone *]
+    ^-  move
+    [bone %quit ~]
+  ::
+  ==
+::
+++  handle-foreign
+  |=  diff=group-update
+  ^-  (quip move _this)
+  ?-  -.diff
+      %keys    [~ this]
+      %bundle  [~ this]
+  ::
+      %path
+    ?~  pax.diff
+      [~ this]
+    =/  ship  (~(get by synced) pax.diff)
+    ?~  ship
+      [~ this]
+    ?.  =(src.bol u.ship)
+      [~ this]
+    :_  this
+    :~  (group-poke pax.diff [%unbundle pax.diff])
+        (group-poke pax.diff [%bundle pax.diff])
+        (group-poke pax.diff [%add members.diff pax.diff])
+    ==
+  ::
+      %add
+    ?~  pax.diff
+      [~ this]
+    =/  ship  (~(get by synced) pax.diff)
+    ?~  ship
+      [~ this]
+    ?.  =(src.bol u.ship)
+      [~ this]
+    :_  this
+    :~  (group-poke pax.diff diff)
+    ==
+  ::
+      %remove
+    ?~  pax.diff
+      [~ this]
+    =/  ship  (~(get by synced) pax.diff)
+    ?~  ship
+      [~ this]
+    ?.  =(src.bol u.ship)
+      [~ this]
+    :_  this
+    :~  (group-poke pax.diff diff)
+    ==
+  ::
+      %unbundle
+    ?~  pax.diff
+      [~ this]
+    =/  ship  (~(get by synced) pax.diff)
+    ?~  ship
+      [~ this]
+    ?.  =(src.bol u.ship)
+      [~ this]
+    :_  this(synced (~(del by synced) pax.diff))
+    :~  (group-poke pax.diff diff)
+    ==
+  ::
+  ==
+::
+++  quit
+  |=  wir=wire
+  ^-  (quip move _this)
+  =/  wir  `(list @tas)`wir
+  =/  =ship  (slav %p &1:wir)
+  =.  wir  ?^  wir  t.wir  ~
+  =.  wir  ?^  wir  t.wir  ~
+  ?:  (~(has by synced) wir)
+    =/  group-path  [%group wir]
+    =/  group-wire  [(scot %p ship) group-path]
+    :_  (track-bone group-wire)
+    [ost.bol %peer group-wire [ship %group-hook] group-path]~
+  ::  no-op
+  [~ this]
+::
+++  reap
+  |=  [wir=wire saw=(unit tang)]
+  ^-  (quip move _this)
+  ?~  saw
+    [~ this]
+  =/  wir  `(list @tas)`wir
+  =/  =ship  (slav %p &1:wir)
+  =.  wir  ?^  wir  t.wir  ~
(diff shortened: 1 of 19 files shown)


pkg/arvo/app/group-hook.hoon
```diff
@@ -0,0 +1,270 @@
...
+  ^-  (quip move _this)
+  ?-  -.diff
+      %keys
+    [~ this]
+      %path
+    [~ this]
+      %bundle
+    [~ this]
+  ::
+      %add
```
philipcmonk: Personally I'd use mixed kingside/queenside here:

```
?-  -.diff
      %keys    [~ this]
      %path    [~ this]
      %bundle  [~ this]
      %add
    :_  this
    %+  turn  (prey:pubsub:userlib [%group pax.diff] bol)
    ...
==
```

pkg/arvo/app/group-hook.hoon
```diff
@@ -0,0 +1,270 @@
...
+  ?:  (~(has by synced) wir)
+    =/  group-path  [%group wir]
+    =/  group-wire  [(scot %p ship) group-path]
+    :_  (track-bone group-wire)
+    [ost.bol %peer group-wire [ship %group-hook] group-path]~
+  ::  no-op
+  [~ this]
+::
+++  reap
+  |=  [wir=wire saw=(unit tang)]
```
philipcmonk: If you don't expect this to fail often, might not be the worst idea to `+slog` out the error.

pkg/arvo/app/group-hook.hoon
```diff
@@ -0,0 +1,270 @@
+/-  *group-store, *group-hook
```
philipcmonk: A brief comment wold be useful.  In particular, it's not immediately obvious what "synced" means.  My first guess would have been that it's a local copy of a foreign group, but it looks like actually it means that a foreign group has been used as input for a local group.  Thus if `~bud` is "synced" to `~dev`'s `/hello` group, then `~bud`'s `/hello` group will be a superset of `~dev`'s.  Is that correct?

pkg/arvo/app/group-hook.hoon
```diff
@@ -0,0 +1,270 @@
...
+      ::  delete a foreign ship's path
+      =/  group-wire  [(scot %p u.ship) %group path.act]
+      :_  this(synced (~(del by synced) path.act))
+      (pull-wire group-wire path.act)
+    :: don't allow
+    [~ this]
+  ::
+  ==
+::
+++  peer-group
```
philipcmonk: Am I correct in understanding that this exists solely so that foreign group-hook apps can peer this because we don't allow foreign apps to peer group-store?

pkg/arvo/app/permission-store.hoon
```diff
@@ -0,0 +1,195 @@
...
+  :^  ~  ~  %noun
+  ?-(kind.u.pem %black !has, %white has)
+::
+++  poke-permission-action
+  |=  action=permission-action
+  ^-  (quip move _this)
+  ?.  =(src.bol our.bol)
+    [~ this]
+  ?-  -.action
+      %add
```
philipcmonk: Would use kingside

pkg/arvo/sur/permission-store.hoon
```diff
@@ -0,0 +1,25 @@
...
++$  permission-map  (map path permission)
+::
++$  permission-update
+  $%  [%create =path =permission]
+      [%delete =path]
+      [%add =path who=(set ship)]
+      [%remove =path who=(set ship)]
+  ==
+::
++$  permission-action
```
philipcmonk: A comment describing the difference between `%add`, `%remove`, `%allow`, and `%deny` could be useful.  Seems to be "`%add` (`%remove`) adds (removes) to the set of specially-handled ships; `%allow` (`%deny`) is same as `%add` (`%remove`) if `%white` else same as `%remove` (`%add`).

pkg/arvo/app/permission-group-hook.hoon
```diff
@@ -0,0 +1,179 @@
...
+  |=  [wir=wire diff=group-update]
+  ^-  (quip move _this)
+  ?-  -.diff
+      %keys
+    [~ this]
+      %bundle
+    [~ this]
+      %path
+    ::  set all permissions paths
+    =/  perms  (~(get by relation) pax.diff)
```
philipcmonk: The logic here could be simplified using the jug operations (since `relation` is actually a jug):

```
    =/  perms  (~(get ju relation) pax.diff)
    :_  this
    %+  turn  ~(tap in perms)
    |=  =path
    ^-  move
    (permission-poke path [%add path members.diff])
```

Same for the other cases.

pkg/arvo/app/permission-group-hook.hoon
```diff
@@ -0,0 +1,179 @@
...
+  =/  checked-paths
+    %+  skim  perm-list
+    |=  pax=path
+    ^-  ?
+    ?~  (permission-scry pax)
+      %.n
+    %.y
+  ?.  =((lent checked-paths) perm-length)
+    [~ this]
+  =/  perms  (~(get by relation) group)
```
philipcmonk: This is also a jug operation, not sure whether it's clearer:

```
  =/  perms  (~(get ju relation) group)
  ::  if relation does not exist, create it and subscribe.
  :-  ?^  perms
        ~
      =/  group-path  [%group group]
      [ost.bol %peer group-path [our.bol %group-store] group-path]~
  =.  perms  (~(uni in perms) permissions)
  this(relation (~(put by relation) group perms))
```

pkg/arvo/app/permission-group-hook.hoon
```diff
@@ -0,0 +1,179 @@
+/-  *group-store, *permission-store, *permission-group-hook
```
philipcmonk: A brief comment describing what exactly is an association, I think I had the direction backwards in my mind.  To be clear, this lets you say "any time I add a person to this group also add them to this set of permissions groups".  For example, you could have five chat rooms all whitelisted to one particular group, and when you add someone to that group they'll have permission to join all five rooms.  Is that correct?

pkg/arvo/app/permission-group-hook.hoon
```diff
@@ -0,0 +1,179 @@
...
+    |=  =path
+    ^-  move
+    (permission-poke path [%delete path])
+  ::
+  ==
+::
+++  quit
+  |=  wir=wire
+  ^-  (quip move _this)
+  ::  TODO: implement resubscribe handler
```
philipcmonk: With this kind of rare error that you're explicitly not handling yet, I recommend a one-line printf so if people report bad behavior you can guess why.

pkg/arvo/lib/group-json.hoon
```diff
@@ -0,0 +1,30 @@
...
+  ^-  ^json
+  %+  frond  %group-initial
+  %-  pairs
+  %+  turn  ~(tap by grp)
+  |=  [=path =group]
+  ^-  [@t ^json]
+  :-  (spat path)
+  (sa group ship:enjs:format)
+::
+++  sa                                                :::  set as array
```
philipcmonk: I assume `+sa`, `+as`, and `+pa` will be removed from here once that lands in zuse.

pkg/arvo/app/group-store.hoon
```diff
@@ -0,0 +1,169 @@
...
+      %bundle
+    (handle-bundle action)
+  ::
+      %unbundle
+    (handle-unbundle action)
+  ::
+  ==
+::
+++  handle-add
+  |=  act=group-action
```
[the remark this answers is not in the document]
loganallenc: If I did this, (send-diff path.act act) in the move emitter of handle-bundle, add, etc would have to reconstitute the original action noun. Possible, but not preferable.

pkg/arvo/app/group-hook.hoon
```diff
@@ -0,0 +1,270 @@
...
+++  group-poke
+  |=  [pax=path action=group-action]
+  ^-  move
+  [ost.bol %poke pax [our.bol %group-store] [%group-action action]]
+::
+++  group-scry
+  |=  pax=path
+  ^-  (unit group)
+  =.  pax  ;:  weld
+    `path`/=group-store/(scot %da now.bol)
```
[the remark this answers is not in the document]
loganallenc: It works for ~bus, yeah.

pkg/arvo/app/group-store.hoon
```diff
@@ -0,0 +1,169 @@
...
+      %remove
+    (handle-remove action)
+  ::
+      %bundle
+    (handle-bundle action)
+  ::
+      %unbundle
+    (handle-unbundle action)
+  ::
+  ==
```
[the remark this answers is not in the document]
loganallenc: I like this, thanks

pkg/arvo/app/group-store.hoon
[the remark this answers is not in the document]
ohAitch: The latter should just nest with the broader action type, no?

pkg/arvo/app/group-hook.hoon
```diff
@@ -0,0 +1,270 @@
...
+  ==
+::
++$  state-zero
+  $:  synced=(map path ship)
+      boned=(map wire (list bone))
+  ==
+::
++$  poke
+  $%  [%group-action group-action]
+      [%group-hook-action group-hook-action]
```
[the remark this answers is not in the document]
loganallenc: Also great suggestion! Will do this.

pkg/arvo/app/group-store.hoon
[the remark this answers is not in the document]
philipcmonk:
```
|=  act=[%add members=group pax=path]
```

would let you still refer to it as `act` but not require asserting that `-.act` is `%add`.

pkg/arvo/lib/group-json.hoon
loganallenc: Yup.

pkg/arvo/app/permission-group-hook.hoon
loganallenc: I need to refactor this function anyway. I have a strong aesthetic distaste for jugs, but I agree the logic of this function could be simplified.

pkg/arvo/app/permission-group-hook.hoon
loganallenc: Yes, that’s correct.

pkg/arvo/sur/permission-store.hoon
loganallenc: Yes, your understanding is right. Will add comments

pkg/arvo/app/permission-group-hook.hoon
philipcmonk: I don't feel strongly about jugs

pkg/arvo/app/group-hook.hoon
loganallenc: If ~bud adds ~dev/help to his synced groups, then ~dev/help will remove ~bud/help and write the contents of ~dev/help on top of the empty ~bud/help.

philipcmonk: Ah, so there's effectively one global namespace and you can choose to use your own version of a particular group path or someone else's.

pkg/arvo/app/group-hook.hoon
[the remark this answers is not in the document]
loganallenc: On second thought, just removed the %group-hook-action poke. I generally don't specify %poke-in since it's easy to tell from the arms.

pkg/arvo/app/group-hook.hoon
loganallenc: Yes.

pkg/arvo/app/permission-group-hook.hoon
loganallenc: This is no longer relevant since I'm correctly handling the ++reap case. It correctly no-ops.

pkg/arvo/app/group-store.hoon
```diff
@@ -0,0 +1,169 @@
...
+::
+++  handle-add
+  |=  act=group-action
+  ^-  (quip move _this)
+  ?>  ?=(%add -.act)
+  ?~  pax.act
+    [~ this]
+  :: TODO: use dif to calculate whether anything changed before sending diff
```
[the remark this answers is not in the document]
loganallenc: Doing this. Thanks.

pkg/arvo/app/group-hook.hoon
loganallenc: I expect it to fail whenever a person's permissions are revoked. Not sure if that counts as often?

pkg/arvo/app/permission-store.hoon
loganallenc: Done.

pkg/arvo/app/permission-group-hook.hoon
loganallenc: What is the difference here? Will just return empty set if perms does not exist at the path?

pkg/arvo/app/permission-group-hook.hoon
loganallenc: Finished the refactor. Feels nicer without need for this.

pkg/arvo/app/permission-group-hook.hoon
philipcmonk: Yeah, a `(jug foo bar)` is conceptually a sparse map of all possible `foo` to sets of `bar`s.  So it says that a `foo` not having an entry in the map is the same as having an entry of "empty set".  Often this is a more natural representation if you don't specifically want to handle the case of "have an entry in the map but it's an empty set".  So it tends to reduce branching, as here.

But it's basically an aesthetic decision; choose what seems clearer to you.

pkg/arvo/app/group-hook.hoon
philipcmonk: Seems like if you tried to subscribe to someone else's group and failed, you might want a message saying that it failed.  That way you might see something like "insufficient permissions" instead of it just looking like it hung.

pkg/arvo/app/permission-group-hook.hoon
loganallenc: I'll keep it as is. Thanks for explaining.

loganallenc:
@philipcmonk Yes, your understanding of the user stories is correct.

pkg/arvo/app/group-hook.hoon
Fang-: > A brief comment wold be useful. 

Agreed all of the files in this PR could use a short explanation of what they do at the top of the file.

pkg/arvo/app/group-hook.hoon
loganallenc: Added this. Thanks.

pkg/arvo/app/group-hook.hoon
loganallenc: Added comments to all of the actions in sur files and to the top of the app files.

loganallenc:
@philipcmonk @ixv Ping. All tests pass, all comments responded to.

loganallenc:
@jtobin Ready!

PR #1774, opened 2019-09-28
u3: refactors memoization patterns for safer reference-counting

joemfb:
Specifically, this change avoids uncounted references in the "decapitated" memoizing compiler jet stubs.

These issues have been latent in the code for a long time, but they were "activated" by the unifying-equality changes in #1741, and surfaced in #1757.

### DIFF
--- a/pkg/urbit/include/noun/zave.h
+++ b/pkg/urbit/include/noun/zave.h
@@ -12,22 +12,27 @@
   ***
   ***  Memo functions RETAIN keys and transfer values.
   **/
-    /* u3z_find*(): find in memo cache.
+    /* u3z_key*(): construct a memo cache-key.  Arguments retained.
     */
-      u3_weak u3z_find(c3_m, u3_noun);
-      u3_weak u3z_find_2(c3_m, u3_noun, u3_noun);
-      u3_weak u3z_find_3(c3_m, u3_noun, u3_noun, u3_noun);
-      u3_weak u3z_find_4(c3_m, u3_noun, u3_noun, u3_noun, u3_noun);
+      u3_noun u3z_key(c3_m, u3_noun);
+      u3_noun u3z_key_2(c3_m, u3_noun, u3_noun);
+      u3_noun u3z_key_3(c3_m, u3_noun, u3_noun, u3_noun);
+      u3_noun u3z_key_4(c3_m, u3_noun, u3_noun, u3_noun, u3_noun);
 
-    /* u3z_save*(): save in memo cache.
+    /* u3z_find*(): find in memo cache. Arguments retained
     */
-      u3_noun u3z_save(c3_m, u3_noun, u3_noun);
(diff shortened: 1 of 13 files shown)


loganallenc:
Thank you king!

belisarius222:
@joemfb I don't quite understand this.  I see that you refactored the memo-cache find operations to work the same way no matter what the key is, and you separated key generation into different routines depending on length of the key tuple.  So it looks like you moved the variation from the find() operation into the key generation.  I'm not sure how this relates to refcounting, though -- or maybe this was a cosmetic change, and the actual fix was more subtle.

Would you mind pointing me at a place where the old code lost a reference?  Looking through it, nothing jumped out at me, and the new code looks almost exactly the same to me.

joemfb:
The change relates to key "lifetimes". In the old pattern, the key was constructed in `u3z_find*`, gaining a reference to each noun argument. The memoization cache was then searched for the key, after which the key was lost (deallocating it, and thereby losing the reference to each noun argument). If there was no cache entry for the key, the raw nock was evaluated, and the product stored in the memoization cache via `u3z_save*` (wherein the key was reconstructed, and the same allocation/reference-counting/deallocation sequence occurred).

In the new pattern, the key is constructed and returned. If there's a cache hit, the reference to the key must be lost; if there's a miss, the same key is used to save the nock product (without de/re-allocation). This seems like a superior approach, as it doesn't involve redundant allocations or churn the reference counts unnecessarily. It's also a *safe* pattern, in contrast to the old way.

When a cell is deconstructed with `u3r_at()`, `u3r_mean()`, or just `u3h()`/`u3t()`, the resulting reference into the content of the cell is *not* counted (ie, `u3r_*` functions *retain* their arguments *and* their return). Quoting `u3.md`: "It is normal and healthy to have these uncounted C references, but they must be treated with care." In particular, the more you do with them *or* the containing noun from which you extracted them, the more care you must have. And it's hard to imagine doing more than passing the containing noun into an arbitrary evaluation, which is exactly what the old partial-memoization pattern did.

In other words, in the "decapitated jets" there are uncounted C references to each of the individual components of the cache keys. This was and is fine, right until we passed the core (`+ut`) from which they were extracted into the bytecode compiler/interpreter (`u3n_nock_on()`). Any arbitrary subtree of that core could be unified with a senior copy of that core (and hopefully will, if one exists). This was not a practical problem in the past, because we almost always *leaked the old reference* when we unified. But I changed that in #1741. So the fix (and safe pattern) is to explicitly take a reference to anything that we need to still exist after the nock evaluation.

belisarius222:
Thanks for the explanation.  I understand this a lot better now, and I approve of it.  The problem with doing unsafe things like that in Vere is that they don't compose well.  Once there are two or three unsafe things happening at once, the possibility for error goes up dramatically.  Removing unsafe patterns in favor of safer ones should help us get a handle on Vere.

—
~rovnys-ricfer
https://urbit.org

Issue #1577, opened 2019-03-07
Possible bug removing the head of a qeu


yosoyubik:
It seems that the logic for removing the head of a queue doesn't behave as it should:

```
> =a (~(gas to `(qeu @)`~) [4 3 2 1 ~])
> [`(qeu @)`a ~(top to a)]
[{1 2 3 4} [~ u=4]]
> =a ~(nap to a)
> [`(qeu @)`a ~(top to a)]
[{1 2 3} [~ u=3]]
> =a ~(nap to a)
> [`(qeu @)`a ~(top to a)]
[{1 3} [~ u=3]]
> =a ~(nap to a)
> [`(qeu @)`a ~(top to a)]
[{3} [~ u=3]]
> =a ~(nap to a)
> [`(qeu @)`a ~(top to a)]
[{} ~]
```
I thought this might have been an issue with the pretty-printer, but this might the balancing arm acting up.

If we have this tree
```
          2
        /   \
      1       3
```
The head is 3, so far so good.
```
> =a (~(gas to `(qeu @)`~) [3 2 1 ~])
> a
[n=2 l={1} r={3}]
> ~(top to a)
[~ u=3]
```
but if we remove the head:
```
> =a ~(nap to a)
> a
[n=1 l={} r={3}]
> ~(top to a)
[~ 3]
```
we get this (?):
```
          1
        /   \
      ~       3
```
This could be where the problem begins:

[bal(a ^+(a [p.b q.b r.a]))](https://github.com/urbit/arvo/blob/master/sys/hoon.hoon#L1871)

If instead of using the right branch of the tree ([p.b q.b **r.a**], replicating the head) 
we use the node ([p.b q.b **[n.a ~ ~]**)
```
> a
[n=2 l={1} r={3}]
> =b ~(get to ?~(a ~ l.a))
> b
[p=1 q={}]
> ~(bal to ^+(a [p.b q.b [?~(a ~ n.a) ~ ~]]))
[n=2 l=[n=1 l={} r={}] r=~]
> `(qeu @)`~(bal to ^+(a [p.b q.b [?~(a ~ n.a) ~ ~]]))
{1 2}
> ~(top to `(qeu @)`~(bal to ^+(a [p.b q.b [?~(a ~ n.a) ~ ~]])))
[~ u=2]
```
we get this tree:
```
          2
        /   \
      1      ~
```

I'm having some issues introducing this changes in hoon.hoon:
```
[%hoon-compile ~raplep-nammeb]
[...]

Issue #1274, opened 2019-05-10
cc-release: booting ship fails


jtobin:
I've tried to take cc-release for a whirl but am having trouble getting it moving.

Doing a simple `make install` appears to work fine on both Debian 10 and NixOS, but I observe roughly the same problem when trying to boot `~nidsut-tomdun`:

Debian:

```
jtobin@pollux:~/urbit$ urbit -w nidsut-tomdun -k $KEYFILE
~
urbit 0.8.0
boot: home is /home/jtobin/urbit/nidsut-tomdun
loom: mapped 2048MB
lite: arvo formula 3685298
lite: core b41b6d6
lite: final state b41b6d6
Assertion '0 != u3_Host.ops_u.url_c' failed in vere/daemon.c:555

bail: oops
bailing out
urbit: noun/manage.c:663: u3m_bail: Assertion `0' failed.
Aborted (core dumped)
```

NixOS:

```
[jtobin@betelgeuse:~/urbit]$ urbit -w nidsut-tomdun -k $KEYFILE
~
urbit 0.8.0
boot: home is /home/jtobin/urbit/nidsut-tomdun
boot: open file limit: Invalid argument
loom: mapped 2048MB
lite: arvo formula 3685298
lite: core b41b6d6
lite: final state b41b6d6
Assertion '0 != u3_Host.ops_u.url_c' failed in vere/daemon.c:555

bail: oops
bailing out
urbit: noun/manage.c:663: u3m_bail: Assertion `0' failed.
Aborted (core dumped)
```

In both cases I get the assertion failure in `vere/daemon.c:555`, but there is a slight difference in that the latter case gets me an `open file limit: invalid argument`.

EDIT:

Peeking at `vere/daemon.c` indicates that this has something to do with the pill URL; trying to bypass and supply a solid pill with `-B bin/solid.pill` gets me a little further, but eventually it aborts with a core dump:

```
jtobin@pollux:~/urbit$ urbit -w nidsut-tomdun -k $KEYFILE -B bin/solid.pill
~
urbit 0.8.0
boot: home is /home/jtobin/urbit/nidsut-tomdun
loom: mapped 2048MB
lite: arvo formula 3685298
lite: core b41b6d6
lite: final state b41b6d6
boot: loading pill bin/solid.pill
boot: no azimuth snapshot specified
boot: retrieving latest block
boot: retrieving ~nidsut-tomdun's public keys
boot: verifying keys
boot: retrieving galaxy table
boot: retrieving network domains
[...]

PR #1777, opened 2019-09-29
tests: unit tests for pure algorithms

yosoyubik:
This submits the work for the following bounty: https://grants.urbit.org/bounties/1322185260-unit-tests-for-pure-algorithms

TL;DR: It provides four test files `qeu.hoon`, `map.hoon`, `set.hoon` and `differ.hoon` with unit tests for all arms under `+to`, `+by`, `+in` and `+differ`.

@jtobin I hope I haven't messed up the commit history too much but following the "Contribution hygiene" guidelines, but I thought that it would make sense to see what files were added, following the definition of "atomic" commits.

There are some things with this PR (the fact that some tests fail) that I would like to explain here (there are comments in the testing files with longer explanations). 

The gist of it is that I believe that I found some issues with some of the arms. I made the mistake in the past of too enthusiastically thinking that I found a bug, when [it was just the wrong comment](https://github.com/urbit/urbit/issues/1577#issuecomment-483845590) so I'll try to be cautious here:

The weird behavior belongs to `+apt:to` arm. For example:

```
~(apt to (~(gas to *(qeu @)) ~[1 2 3]))
>  %.n
```
 Looking at the implementation of the +apt arm I can see that the heap property is regarding siblings, not parent-children. I proposed a new arm `+check` in the test suite to fix it. @joemfb I can see that you worked [on this before](https://github.com/urbit/arvo/pull/1180#issuecomment-497096181) so if you could confirm if there's something wrong with it I'll appreciate it.

This is more of a detour into CS theory, but the `+qeu` is defined as Treap in the docs. It's been a while since i looked at data structures, but I remembered a Treap been ordered like a BST. The implementation of the `+qeu` in hoon.hoon does not follow that constrain (only the heap order) so the order of the leaves follow a FIFO queue. So I'm not sure if I would call this a Treap? Also, there was some misunderstandings of my part in the past regarding how the `+qeu` was used by ames, and I know that @philipcmonk and @belisarius222 looked at it carefully so probably they can correct me again. My question would be, if ames perform a search in `+puq` traversing left or right based on a comparison [`?:  (gth num p.n.puq)`](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/ames.hoon#L428) wouldn't we miss some packets?

Another issue (very minor I believe since the arm is not used anywhere) comes from `+uno:by`. The general union (in the sense of modifying the value of the element of the intersection based on the product of a gate) doesn't work:

```
((~(uno by (my ~[[1 9] [7 3] [8 5]])) (my ~[[1 2] [7 2]])) |=([k=@ v=@ w=@] (add v w)))
> [n=[p=8 q=5] l=~ r=[n=[p=7 q=2] l={} r={[p=1 q=2]}]]
```
I proposed a new arm called `+general-union` in the test set that return the proper result. The same issue affects `+uni:by` and `+uni:in` but it only means that there is an extra branch that is never called (because `(mor a b)` is non-strict so cases where `=(a b)` are `%.y` are contained in the `%.y` branch from testing the product of `+mor`.

Lastly, I'd like to comment about aesthetics. A careful reviewer will notice that the four test files don't follow the same pattern structure. If one researches testing suites from other famous open source projects, they will find that the tests are explicit in the sense that the space they take in a file is proportional to the number of things tested (i.e. very long files). You can see this pattern in some of my files (`+map` for example). In order to reduce the repetition of the test data for each arm, I declared all the data in a core at the top of the file with short names (that could remind some people of out-of-date naming conventions). This might have the problem of hiding the information of what is actually being tested (e.g. what are the elements in this array?). If this is the agreed consensus among Hoon overlords, I'll be happy to refactor it.

Now, if we come to `+qeu` I wanted to make sure that my proposed arms were thoroughly tested, without repeating the declaration of the same test data everywhere, so I decided to go with the same structure as `+map` with the twist that, if you are checking if the product of a gate is `%.y` or `%.n` and your test fails, you won't know which one was the cause, because the output will be:
```
expected: %.y
actual: %.n
```
so what I did was to tag each entry in the test suite and use it in `expect-eq` to get something like this:
```
expected: [%tri %.y]
actual: [%tri %.n]
```
I'm wondering if adding a `term` to the expect arms would be a good improvement for the `+test` environment, when tracking which case is failing...

I can't believe that I wrote all this, and half of it was about aesthetics of code styling. Hopefully this won't hide the fact that I had a lot of fun working on this bounty :)

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:585c83b1389eb11336a53b6cb711d33d5d510b8c982efd3fe526cc09ecfbaa91
-size 15602738
+oid sha256:2f5fab8f444a35101e8667761f9ea87f9a9ae8e62bf9bf2275f568f49698a43f
+size 15815965
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1365,8 +1365,6 @@
         ?:  (gor n.b n.a)
           $(a [n.a $(a l.a, b [n.b l.b ~]) r.a], b r.b)
         $(a [n.a l.a $(a r.a, b [n.b ~ r.b])], b l.b)
-      ?:  =(n.a n.b)
-        [n.b $(b l.b, a l.a) $(b r.b, a r.a)]
       ?:  (gor n.a n.b)
         $(b [n.b $(b l.b, a [n.a l.a ~]) r.b], a r.a)
       $(b [n.b l.b $(b r.b, a [n.a ~ r.a])], a l.a)
@@ -1634,8 +1632,6 @@
         ?:  (gor p.n.b p.n.a)
           $(a [n.a $(a l.a, b [n.b l.b ~]) r.a], b r.b)
         $(a [n.a l.a $(a r.a, b [n.b ~ r.b])], b l.b)
-      ?:  =(p.n.a p.n.b)
-        [n.b $(b l.b, a l.a) $(b r.b, a r.a)]
       ?:  (gor p.n.a p.n.b)
         $(b [n.b $(b l.b, a [n.a l.a ~]) r.b], a r.a)
       $(b [n.b l.b $(b r.b, a [n.a ~ r.a])], a l.a)
@@ -1653,14 +1649,12 @@
         b
       ?:  (mor p.n.a p.n.b)
         ?:  =(p.n.b p.n.a)
-          [n.b $(a l.a, b l.b) $(a r.a, b r.b)]
+          :+  [p.n.a (meg p.n.a q.n.a q.n.b)]
+            $(b l.b, a l.a)
+          $(b r.b, a r.a)
         ?:  (gor p.n.b p.n.a)
           $(a [n.a $(a l.a, b [n.b l.b ~]) r.a], b r.b)
         $(a [n.a l.a $(a r.a, b [n.b ~ r.b])], b l.b)
-      ?:  =(p.n.a p.n.b)
-        :+  [p.n.a (meg p.n.a q.n.a q.n.b)]
-          $(b l.b, a l.a)
-        $(b r.b, a r.a)
       ?:  (gor p.n.a p.n.b)
         $(b [n.b $(b l.b, a [n.a l.a ~]) r.b], a r.a)
       $(b [n.b l.b $(b r.b, a [n.a ~ r.a])], a l.a)
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1750,12 +1750,11 @@
   =|  a/(tree)  ::  (qeu)
   |@
   ++  apt                                               ::  check correctness
-    ^-  ?
-    ?~  a  &
     |-  ^-  ?
-    ?~  l.a  &
-    ?~  r.a  &
-    &((mor n.l.a n.r.a) $(a l.a) $(a r.a))
+    ?~  a  &
+    ?&  ?~(l.a & ?&((mor n.a n.l.a) $(a l.a)))
+        ?~(r.a & ?&((mor n.a n.r.a) $(a r.a)))
+    ==
   ::
   ++  bal
     |-  ^+  a
--- a/pkg/arvo/tests/sys/hoon/differ.hoon
+++ b/pkg/arvo/tests/sys/hoon/differ.hoon
@@ -1,0 +1,386 @@
+::  Tests for +differ (a suite of Hunt-McIlroy diff and merge algorithms)
+::
+/+  *test
+::
+=,  differ
+::  Testing arms
+::
+|%
+::  ++berk:differ: invert diff patch
+::
+++  test-berk  ^-  tang
+  ::  An inverted diff between %a and %b can be checked
+  ::  by patching %b and obtaining %a
+  ::
+  ;:  weld
+    ::  (some) test examples adapted from:
+    ::  https://github.com/gioele/diff-lcs/blob/master/test/test_diff-lcs.rb
+    ::
+    =/  a  "abcehjlmnp"
+    =/  b  "bcdefjklmrst"
+    =/  diff-a-b  (lusk a b (loss a b))
+    =/  diff-b-a  (lusk b a (loss b a))
+    =/  patch-a  (lurk b (berk diff-a-b))
+    =/  patch-b  (lurk a (berk diff-b-a))
+    %+  expect-eq
+      !>  a^b
+      !>  patch-a^patch-b
+    ::
+    =/  a  "abcde"
+    =/  b  "ae"
+    =/  diff-a-b  (lusk a b (loss a b))
+    =/  diff-b-a  (lusk b a (loss b a))
+    =/  patch-a  (lurk b (berk diff-a-b))
+    =/  patch-b  (lurk a (berk diff-b-a))
+    %+  expect-eq
+      !>  a^b
+      !>  patch-a^patch-b
+    ::
+    =/  a  "ae"
+    =/  b  "abcde"
+    =/  diff-a-b  (lusk a b (loss a b))
+    =/  diff-b-a  (lusk b a (loss b a))
+    =/  patch-a  (lurk b (berk diff-a-b))
+    =/  patch-b  (lurk a (berk diff-b-a))
+    %+  expect-eq
+      !>  a^b
+      !>  patch-a^patch-b
+    ::
+    =/  a  "vxae"
+    =/  b  "wyabcde"
+    =/  diff-a-b  (lusk a b (loss a b))
+    =/  diff-b-a  (lusk b a (loss b a))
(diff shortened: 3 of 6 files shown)


jtobin:
Fantastic stuff!

belisarius222:
@yosoyubik I'm pretty sure you found an actual bug this time :)

Hoon abuses the term "treap".  A `+qeu` is indeed a FIFO queue in the shape of a binary tree using the double-mug `+mor` as vertical ordering and insertion order as horizontal ordering.

Ames abuses the notion even further by declaring `puq` as a `+qeu` but then using a custom ordering function, turning the data structure into an ordered map where the packet sequence number (really more like [message-num fragment-num]) is the key, providing horizontal ordering, and the vertical ordering remains `+mor`.  To answer your question about `.puq` traversal, this should indeed allow for deletion of all packets within a range of sequence numbers without breaking any invariants.

I've made this a bit more explicit in new Ames, with the `+ordered-map` data structure:
https://github.com/urbit/urbit/blob/alef-testnet/pkg/arvo/sys/vane/alef.hoon#L69

Adding a `term` to `+expect-eq` is not a bad idea.  Maybe we could even print it after the test name on failure, or even as an extension of the test path.  I like that better than adding some kind of statefulness into the `+expect` interface, like incrementing a counter for each call to `+expect-eq`.  If you haven't noticed, though, our testing API could definitely use some work.  If you have more thoughts about how you'd like to interact with the test runner, I'd be interested to hear them.

yosoyubik:
Thanks @jtobin :)

I spent some time scratching my head trying to figure out how this was exactly a treap ("did i miss any magical incantation on how to order the elements?"). I had some rambling comments on the `+test/qeu` about the treap ordering and such but removed them now that this is more clearly understood.

I'll keep you posted if I come up with more ideas about the test runner. (PS: new ames code+comments is a pleasure to read @belisarius222)

I'm running into some issues creating a pill (I have added two more commits here to fix #1778 and #1779). This is the output of `sh/update-solid-pill`:
<details>
  <summary>show output</summary>

```
++ nix-build nix/ops -A solid --no-out-link
these derivations will be built:
  /nix/store/397nzdjbp67smcd63ijbnnryzswlmvfm-fakezod.drv
  /nix/store/iy9p6467p520wvhzk5i9pfqj6ak3a652-solid.drv
building '/nix/store/397nzdjbp67smcd63ijbnnryzswlmvfm-fakezod.drv'...
+ '[' -z '' ']'
+ /nix/store/c66id8xskcgpl9z126ixp05lpk2dc7xp-urbit/bin/urbit -d -F zod -B /nix/store/z68p20m7p0wxg2rhbz24994qwnv9wp36-solid.pill /nix/store/z9kbb703inwyrlv8m1dpniwck2xkx7z3-fakezod
~
urbit 0.8.2
boot: home is /private/var/folders/5x/z4085zxn1hjghh56s7j9dsf40000gn/T/nix-build-fakezod.drv-0//nix/store/z9kbb703inwyrlv8m1dpniwck2xkx7z3-fakezod
boot: running as daemon
loom: mapped 2048MB
lite: arvo formula 53d04a6a
lite: core 65cd7496
lite: final state 65cd7496
boot: loading pill /nix/store/z68p20m7p0wxg2rhbz24994qwnv9wp36-solid.pill
loom: mapped 2048MB
boot: protected loom
live: logical boot
boot: installed 262 jets
work: play 1
boot: ship: ~zod (fake)
work: (1)| boot
work: (1)| pill: 254bddf3
+ check
++ herb /nix/store/z9kbb703inwyrlv8m1dpniwck2xkx7z3-fakezod -d 3
Traceback (most recent call last):
  File "/nix/store/hanyd8bkk1p8509vvd9r4n1y75jn4kyz-herb/bin/herb.py", line 413, in <module>
    with open(os.path.join(args.pier, ".http.ports")) as ports:
IOError: [Errno 2] No such file or directory: '/nix/store/z9kbb703inwyrlv8m1dpniwck2xkx7z3-fakezod/.http.ports'
+ '[' 3 -eq '' ']'
/nix/store/g9chp09nag8na3w4p4cv0sfkj9hsjm0p-builder.sh: line 13: [: : integer expression expected
+ echo 'Boot failure.'
Boot failure.
+ kill 79668
/nix/store/g9chp09nag8na3w4p4cv0sfkj9hsjm0p-builder.sh: line 22: kill: (79668) - No such process
+ true
+ exit 1
+ exitHandler
+ exitCode=1
+ set +e
+ '[' -n '' ']'
+ ((  1 != 0  ))
+ runHook failureHook
++ shopt -po nounset
+ local 'oldOpts=set +o nounset'
+ set -u
+ local hookName=failureHook
+ shift
+ local 'hooksSlice=failureHooks[@]'
+ local hook
+ for hook in "_callImplicitHook 0 $hookName" ${!hooksSlice+"${!hooksSlice}"}
+ _eval '_callImplicitHook 0 failureHook'
++ type -t '_callImplicitHook 0 failureHook'
+ '[' '' = function ']'
+ set +u
+ eval '_callImplicitHook 0 failureHook'
++ _callImplicitHook 0 failureHook
++ set -u
++ local def=0
++ local hookName=failureHook
++ case "$(type -t "$hookName")" in
+++ type -t failureHook
++ '[' -z '' ']'
++ return 0
+ set -u
+ eval 'set +o nounset'
++ set +o nounset
+ return 0
+ '[' -n '' ']'
+ exit 1
builder for '/nix/store/397nzdjbp67smcd63ijbnnryzswlmvfm-fakezod.drv' failed with exit code 1
cannot build derivation '/nix/store/iy9p6467p520wvhzk5i9pfqj6ak3a652-solid.drv': 1 dependencies couldn't be built
error: build of '/nix/store/iy9p6467p520wvhzk5i9pfqj6ak3a652-solid.drv' failed
+ pkg=
```
</details>

jtobin:
> I'm running into some issues creating a pill

I've seen this error before but can't remember exactly what it refers to.  Can you check that the pills in your `bin/` directory are actual pills, and not git LFS pointers?

yosoyubik:
They look like real files...

```
-rw-r--r--   1 15M  Oct  2 10:45 solid.pill
-rw-r--r--   1 12M  Oct  2 10:45 brass.pill
-rw-r--r--  1 7.3M  Oct  2 10:45 ivory.pill

```
but probably they are not?:
```
> git lfs ls-files 
bb0a7faca3 * bin/brass.pill
535dc0cc19 * bin/ivory.pill
585c83b138 * bin/solid.pill
```
This doesn't return anything `git lfs pointer --check --file bin/*.pill`

jtobin:
The similar error I encountered was in #1274, and indeed had to do with the pill really being a git LFS pointer.  But yeah, in your case they don't look to be LFS pointers.. 🤔 

It looks like I'm able to create a pill from your branch, in the meantime, so I can just PR an update over to you as soon as it finishes.

jtobin:
(Don't worry about the actual pill conflict here, I'll sort that out on merge.)

yosoyubik:
Very happy that this has been merged but.. since no reviewers have said anything about it, and I'm pretty sure that I missed something, I'm wondering wether we should wait to hear comments from others?

jtobin:
I've given it a look over myself -- the PR might be missing my formal green checkmark, but I definitely approve of the changes! 😄

There's quite a bit of code, so I primarily just skimmed the diff (giving more or less attention to certain areas).  It looks to be of very high quality.  In addition, the assorted comments that you, @belisarius222, and @joemfb provided bolstered my opinion that this is a strict improvement, so I was happy to merge it as-is.

To your point, though -- I'd sort of forgotten this work is part of a grant, so I suppose someone will want to see it reviewed with some consideration of the spec posted there.  I can give it a closer ad-hoc look for you just to verify that's all good?

yosoyubik:
thanks @jtobin! :) yes, my comment was in reference to the the bounty; one wants to follow all the formal requirements when dealing with sensitive work like this one ;)

>  I can give it a closer ad-hoc look for you just to verify that's all good?

I'm not in a hurry to get this approved, but if you have time that'd be awesome!

joemfb:
FWIW, we should definitely make sure to have PR approval before merging.

To be clear about the status of these combined changes:

- This PR contains a host of new test suites, a cosmetic refactoring of +uni:in and +uni:by, and corrections to +uno:by and +apt:to.
- #1799 contains a general, simple refactoring (`%=`) of +in/+by/+to, along with aforementioned +uni/+uno changes.
- @jtobin fixed the spurious `+qeu` clam test while merging

These combined changes should resolve #1778 and #1779.

To take full advantage of #1799, the `+in` and `+by` jets at least be refactored to follow the corrected `+uni` algorithm, and perhaps be further refactored to use `%=` (ie, `u3i_edit()`). But no changes to the jets are strictly necessary, as all current (enabled) `+in` and `+by` jets produce the same result as the relevant hoon/nock, even if the implementations happen to differ. Put another way, the only semantic changes in either PR were to `+uno:by` and `+apt:to`, neither of which is jetted.

Is all of this correct? And is there any other outstanding tasks related to this effort?

yosoyubik:
Thanks @jtobin for your review and comments!

@joemfb I believe your summary is correct and, to my knowledge, there are no more tasks related to this.

PR #1782, opened 2019-10-01
arvo: trim vane in response to memory pressure

joemfb:
Following on #1773, this PR fills out (some of) the `%trim` memory-pressure notification handlers.

- in arvo proper, we clear compiler caches at high priority
- in %eyre, we close all inactive channels
- in %ford, we clear half of the cache(s) at low priority, all at high
- in %gall, we clear compiler caches

Some potential improvements are noted in comments; for instance, we may want to cancel active %ford builds or close active %eyre channels at high priority.

%clay remains a good candidate for further improvements, as it could discard or tombstone foreign desks. But I think reclaiming from its blob store would require a GC-style pass over the entire state.

/cc @philipcmonk, @eglaysher

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1910,8 +1910,42 @@
     [~ http-server-gate]
   ::  %trim: in response to memory pressure
   ::
+  ::    Cancel all inactive channels
+  ::    XX cancel active too if =(0 trim-priority) ?
+  ::
   ?:  ?=(%trim -.task)
-    [~ http-server-gate]
+    =/  event-args  [[our eny duct now scry-gate] server-state.ax]
+    =*  by-channel  by-channel:(per-server-event event-args)
+    =*  channel-state  channel-state.server-state.ax
+    ::
+    =/  inactive=(list @t)
+      =/  full=(set @t)  ~(key by session.channel-state)
+      =/  live=(set @t)
+        (~(gas in *(set @t)) ~(val by duct-to-key.channel-state))
+      ~(tap in (~(dif in full) live))
+    ::
+    ?:  =(~ inactive)
+      [~ http-server-gate]
+    ::
+    =/  len=tape  (scow %ud (lent inactive))
+    ~>  %slog.[0 leaf+"eyre: trim: closing {len} inactive channels"]
+    ::
+    =|  moves=(list (list move))
+    |-  ^-  [(list move) _http-server-gate]
+    =*  channel-id  i.inactive
+    ?~  inactive
+      [(zing (flop moves)) http-server-gate]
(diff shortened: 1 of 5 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1910,8 +1910,29 @@
...
+      =/  live=(set @t)
+        (~(gas in *(set @t)) ~(val by duct-to-key.channel-state))
+      ~(tap in (~(dif in full) live))
+    ::
+    =|  moves=(list (list move))
+    |-  ^-  [(list move) _http-server-gate]
+    ?~  inactive
+      ~>  %slog.[0 leaf+"eyre: trim: closing inactive channels"]
+      [(zing (flop moves)) http-server-gate]
+    =^  mov  server-state.ax  (on-channel-timeout i.inactive)
```
eglaysher: If you're raw calling on-channel-timeout, I don't think this clears the behn timer which is going to call it in the future.

joemfb: ah, thanks.

philipcmonk:
Gall looks good.

For %clay, it's true that to reclaim anything from the blob store we'd have to do a GC pass.  I'd collect a list of "roots" (heads of desks we want to keep), then recurse through each commit's parents to get a list of all the commits we need.  Then go through each commit and get each blob it refers to, including any blobs those blobs refer to.  Then just keep that.

I'm not immediately sure how to decide whether we can forget about a foreign desk's data.  Possibly parts of the state machine assume we keep a cache.  Hard to say exactly what's safe to purge.

In practice, I don't think Clay gets too big (78MB on ~wicdev-wisryt which has been up this whole era, though I've only made a few changes locally), so probably not worth risking it right now.

pkg/arvo/sys/vane/eyre.hoon
joemfb: Done, PTAL

joemfb:
Thanks, @philipcmonk, that all makes sense. I'll hold off for now.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1910,8 +1910,42 @@
...
+      [(zing (flop moves)) http-server-gate]
+    ::  discard channel state, and cancel any active gall subscriptions
+    ::
+    =^  mov  server-state.ax  (on-channel-timeout:by-channel channel-id)
+    ::  cancel channel timer
+    ::
+    =/  channel  (~(got by session.channel-state) channel-id)
+    =?  mov  ?=([%& *] state.channel)
+      :_  mov
+      (cancel-timeout-move:by-channel channel-id p.state.channel)
```
eglaysher: Brute force, but good enough.

PR #1783, opened 2019-10-01
app: added chat store, hook, marks, and view.

loganallenc:
Hi, here's the big change! This is a complete and functioning chat application backend and frontend containing all desired features except: invites and a chat-cli.

Hall has been replaced by a new approach to userspace architecture outlined in this document: https://docs.google.com/document/d/1hS_UuResG1S4j49_H-aSshoTOROKBnGoJAaRgOipf54/edit?usp=sharing

Features included in this PR:
- Sending chat messages with message types (%text, %code, %url)
- Setting a chatroom to read and keeping track of unread message count.
- Fine-grained permissions for read-access and write-access to chatrooms.
- Joining foreign chatrooms from the chat UI.
- Leaving chatrooms.
- Paginating message queries so as to have fast load times.

Invites will come in a second PR from me once this is merged.
The new %chat-cli will come in a separate PR from Mark afterwards, and then we will be able to remove %hall and %talk entirely.

This replaces the old chat.hoon file, but not %hall or %talk. This PR does not affect %hall or %talk.


pkg/arvo/sur/chat-store.hoon
```diff
@@ -0,0 +1,47 @@
+|%
+::
+++  serial  @uvH
+::
++$  letter
+  $%  [%text text=cord]
+      [%url url=cord]
+      [%code expression=cord output=(list tank)]
```
joemfb: FWIW, I'm not sure the simplification of message types here is a win. The missing types (from hall) are invites, replies, `@` messages (a la IRC `/me`), long-form/attachments, and app-messages. Of those, only the last can be trivially reconstructed and need not be a separate type. Replies, `@` messages, and long-form all encode user intent, in a way that can be transparently reconstructed. And invites are yet-to-be-built ...

pkg/arvo/sur/chat-store.hoon
```diff
@@ -0,0 +1,47 @@
...
+::
++$  config
+  $:  owner=ship
+      length=@
+      read=@
+  ==
+::
++$  mailbox
+  $:  =config
+      envelopes=(list envelope)
```
joemfb: I'm a little surprised this is a list rather than a map. What's the purpose of the serial uid, and how does it relate to the position on an envelope in the mailbox list?

.gitignore
```diff
@@ -16,3 +16,4 @@ release/
 **/*.swp
 **/*.swo
 **/*-min.js
+**/package-lock.json
```
joemfb: Far be it from me to kick of a JS build systems discussion, but I think dependency-pinning is the right approach here, and shouldn't be abandoned.

pkg/arvo/sur/chat-store.hoon
loganallenc: Hi Joe. In my opinion, long-form attachments should be written into another service. Similarly, I think threaded data structures and replies should be written as other services. Keep in mind that this is a purposefully opinionated and simple chat data structure, and not the only one that could exist. I think a threaded data structure will be one of the next types of data stores to be written, in fact, but I do not think it will be made to cater to a chat service. Instead, I think it will be written for a microblogging-type UI, which imo threading is well-suited for.

pkg/arvo/sur/chat-store.hoon
loganallenc: The serial UID makes it much easier for us to build client features such as de-duplication of pending messages, and a list seems like a natural fit for this ordered data structure.

pkg/arvo/app/chat-store.hoon
```diff
@@ -0,0 +1,259 @@
...
+    |=  [=bone *]
+    [bone %diff %chat-update act]
+  ::
+    ^-  (list move)
+    ?.  |(=(%create -.act) =(%delete -.act))
+      ~
+    %+  turn  (prey:pubsub:userlib /keys bol)
+    |=  [=bone *]
+    [bone %diff %chat-update act]
+  ==
```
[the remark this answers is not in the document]
loganallenc: We do need to be unconditionally sending to these paths, as they are simply different means of receiving updates from different parts of the data. Thanks for the recommendation about zing! I like the style of this. Thanks!

.gitignore
loganallenc: Noted. If I get more pushback on this, will revert that part of the change.

jtobin: I agree that these should probably be kept in place.  We want as much determinism and certainty as we can get in the Landscape builds, and the lockfiles help with that.

pkg/arvo/app/chat-view.hoon
```diff
@@ -0,0 +1,328 @@
...
+  ::
+     [%'~chat' *]
+    :_  this
+    [ost.bol %http-response (html-response:app index)]~
+  ==
+::
+++  poke-json
+  |=  jon=json
+  ^-  (quip move _this)
+  ?.  =(src.bol our.bol)
```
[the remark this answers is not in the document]
Fang-: +1, both here and in stores. This is also important for chat-cli to achieve parity with talk when it comes to moon behavior.

pkg/arvo/sur/chat-store.hoon
Fang-: At the very least we should just get `@` messages back in. They're a pretty core and simple message type, and get you huge gains in expressiveness.

The reason this isn't moving towards your hellworld of "talk to some friends on Chat, others on Microblog" is that you can, theoretically trivially, write a hook to pull their data into a single store, right? Wouldn't that still be easier if there's more overlap between what the message types allow for?

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,372 @@
+/-  *permission-store, *chat-hook
```
Fang-: Same as in the other PR, these should all have brief comments at the top.

Fang-:
Also the marks should probably go in under `/mar/chat/` (and with the `chat-` prefix removes from their filenames. Pretty sure you can just move/rename those without anything breaking, because of how mark paths/names work.

ohAitch:
I would like to echo Jared's perception here: as someone not internal to Tlon, I find myself lacking in insight on the decision process going into this change, and its benefits. (The clearest feature described is the removal of talk's primary(terminal) UI, which was [vetoed](https://github.com/urbit/urbit/pull/1761#issuecomment-535255671) by @vvisigoth on the previous "delete chat" PR)

pkg/arvo/app/chat-store.hoon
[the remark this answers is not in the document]
Fang-: `/updates` seems to always get everything though, just like `/all`. Looks like the only difference is that `/all` sends initial state on-peer, while `/updates` doesn't?

The subscriber can just choose to ignore the `%chat-initial`, and "data, then updates to data" is (I believe) the expected behavior of gall peers, so `/updates` should probably just go?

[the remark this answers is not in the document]
loganallenc: Your understanding of them is correct, but I think it's better to have a ++peer that operates as /updates does. If we don't want the whole state to be sent, then it shouldn't be sent (shouldn't need to be ignored).

loganallenc:
@jtobin Hello Jared, I added information to my PR. I think the discussion of expectations relating to what information should or should not go into a PR is a topic is better taken up in emails on urbit-dev than on this PR. Thanks for noting your own expectations.

.gitignore
loganallenc: Thanks for noting this, will revert that part of the change.

pkg/arvo/sur/chat-store.hoon
loganallenc: @ messages are unnecessary. They could potentially be added later, but it is not a show-stopping priority for me. I have nothing against them in principle, apart from thinking they are cruft.

ohAitch:
Thank you, the amended description is a lot clearer about "hall has been replaced" being a multi-part project.

pkg/arvo/app/chat-store.hoon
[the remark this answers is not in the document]
loganallenc: Thanks, converted to zing.

pkg/arvo/app/chat-view.hoon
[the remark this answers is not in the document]
loganallenc: Changed, good point. I wasn't thinking of this.

pkg/arvo/app/chat-view.hoon
```diff
@@ -0,0 +1,328 @@
...
+  ::  then quit the subscription
+  :_  this
+  :~  [ost.bol %diff %json (inbox-to-json (truncate-inbox all-scry))]
+      [ost.bol %quit ~]
+  ==
+::
+++  peer-updates
+  |=  pax=path
+  ^-  (quip move _this)
+  ?.  =(src.bol our.bol)  !!
```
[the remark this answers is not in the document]
loganallenc: Thanks, changed.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+::
+++  reap
+  |=  [wir=wire saw=(unit tang)]
+  ^-  (quip move _this)
+  ?~  saw
+    [~ this]
+  =/  wir  `(list @tas)`wir
+  =/  =ship  (slav %p &1:wir)
+  =.  wir  ?^  wir  t.wir  ~
+  =.  wir  ?^  wir  t.wir  ~
```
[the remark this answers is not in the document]
loganallenc: Thanks, I didn't really understand =^ before this.

pkg/arvo/app/chat-view.hoon
```diff
@@ -0,0 +1,328 @@
...
+  [ost.bol %poke / [our.bol %permission-group-hook] pok]
+::
+++  envelope-scry
+  |=  pax=path
+  ^-  (list envelope)
+  =.  pax  ;:  weld
+    `path`/=chat-store/(scot %da now.bol)/envelopes
+    pax
+    `path`/noun
+  ==
```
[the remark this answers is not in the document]
loganallenc: Thanks!

pkg/arvo/app/chat-hook.hoon
loganallenc: Good point, thanks.

.gitignore
loganallenc: Done.

jtobin:
@loganallenc I still don't think the amended PR description answers the question of why this should replace the existing chat app.  Feel free to contest me if you think I'm off base, but in my estimation you haven't presented any case that this is superior to the status quo, and there are a number of reasons to suggest it may not be.

It would be helpful if you could describe exactly how this app compares with the existing chat app.  As mentioned in other comments, various features of the current app, including `@` messages, appear to be missing, so it's far from clear that this is a net improvement.  Re: the features you've listed there -- how do these compare to the features of the current chat application?  It's not obvious to me.

Related to all this are questions of reliability and such.  This PR changes 62 files, which is a big change indeed.  The code appears to be of high quality, but there don't appear to be any tests, and it's not clear to what degree this stuff has been put through the ringer.  The existing chat app is battle-tested, and its warts are well-known.  Why should this one be considered reliable enough to replace it?

I apologize if this all seems onerous or picky, but these are real concerns, and I'm wary of merging this PR without seeing them addressed.  What I want to avoid is swapping the enemy one knows for a (potentially gnarlier) enemy one doesn't.  If you can make a convincing case that these changes constitute a strict improvement, I'll be happy to merge them in, but the onus is on you to make that case.

(Tangentially, I think you're jumping the gun in claiming that :hall has been replaced.  You are *proposing* to replace :hall.  The only comment on the (draft, unratified in any sense) UP you linked is one from me, which, FWIW, suggested adding a more concrete demonstration that the proposed model is superior.)

Fang-:
> I think you're jumping the gun in claiming that :hall has been replaced. You are proposing to replace :hall

I was indeed under the impression that we want this to go in, for now, as a realistic-scale implementation of the linked, proposed userspace architecture. More of a tech demo than a direct replacement, though the intent is to _eventually_ replace Hall fully. At the very least, we want this to see some real usage before doing so.

jtobin:
> I was indeed under the impression that we want this to go in, for now, as a realistic-scale implementation of the linked, proposed userspace architecture. More of a tech demo than a direct replacement, though the intent is to eventually replace Hall fully. At the very least, we want this to see some real usage before doing so.

Right -- I don't mean to imply that an implementation of proposed architecture shouldn't go in, I just mean that until it *does* go in, and :hall goes out, then :hall has not actually been replaced.  IMO it would be fair to say "we're planning to deprecate :hall," but not "we've deprecated :hall." 

(I hesitated to add that comment because I didn't want to derail the present PR, but the point did seem to be worth elaborating on.)

vvisigoth:
@jtobin correct me if I'm wrong, but it seems that you are pointing out four issues blocking a merge.

1. Better explanation of the rationale for this PR
2. Parity re the pat message type
3. Tests
4. Staging issues

(N.B. in the below I use "`%hall`" to mean the back end message service, "`%talk`" to be the cli client for same and "`%chat`" to be a separate client for "`%hall`" supporting the web UI).

1) I will let @loganallenc amend the PR description, but there are plenty of known issues with currently extant `%hall`, many of which are exposed by trying to build other clients (besides `%talk`) of `%hall`. You'll remember that this was the rationale for even having the `%hall`/`%talk` split in the first place. This also puts us on a better footing to have groups that can be used across applications. The PR description can be amended to reflect this for posterity. 

2) I'm inclined to agree with @loganallenc that supporting what is essentially a display preference in the back end (as opposed to merely parsing it in the client) is almost a MECE violation. But then again, maybe someone can argue me out of this position, so my vote is to put it back in because you can't die on every hill.

3) I see how adding some `%ph` tests could be useful, but I don't recall that being communicated as a standard (more as a future possibility). If you want tests for this, please elucidate what you'd like, especially for userspace code.

4) You point out that many files are changed here, but almost all of the changes are additive or are changes to the `%chat` client, which is the web UI. These changes I consider to be in the purview of Interface and have been reviewed for parity with the existing UI. There is nothing in this (with the exception to a one-line change to `%drum`) that touches the existing`%hall`/`%talk` system.

4 I consider a non-issue and is basically a misunderstanding. 2, I would go either way on. 3 would be nice, but I'd kind of like in the context of the whole proposed system (including invites), and if you're going to delay merge for tests, please be very clear about what you'd like. 1 I agree that the rationale should be briefly summarized here.

jtobin:
I feel I've been very clear about my concerns here:

> You haven't presented any case that this is superior to the status quo, and there are a number of reasons to suggest it may not be.

> What I want to avoid is swapping the enemy one knows for a (potentially gnarlier) enemy one doesn't.

I've also made several suggestions as to how my concerns can be assuaged:

> It would be helpful if you could describe exactly how this app compares with the existing chat app.

> Re: the features you've listed there -- how do these compare to the features of the current chat application?

> Urbit already has a complete and functioning chat application -- what makes this one any different, or more desirable?

> The existing chat app is battle-tested, and its warts are well-known. Why should this one be considered reliable enough to replace it?

Or, equivalently:

> If you can make a convincing case that these changes constitute a strict improvement, I'll be happy to merge them in.

Basically: the thing that's being replaced should be compared with the thing that's replacing it.

(Presence of tests, merits of `@` as a feature, etc. -- none of those particular details are all that important *per se*, and any of them can certainly be overlooked.)

Fang-:
> supporting what is essentially a display preference in the back end (as opposed to merely parsing it in the client) is almost a MECE violation

> merely parsing it in the client

But then you have to get clients to agree on what text turns into what kind of message and then implement that parsing, instead of relying on data structures provided by the store they all use. This is my biggest complaint against any "just use a `@t` bro" efforts.

jtobin:
> There is nothing in this (with the exception to a one-line change to %drum) that touches the existing%hall/%talk system.

To be clear, I'm aware of this, and am referring to the replacement of `app/chat.hoon` with chat-{store, hook, etc.}.

joemfb:
A couple things:

Apps have generally not been seriously unit-tested, and it's quite awkward to do so. Aquarium seems much better suited to kernel-level testing (given that it operates/asserts at the event/effect level). Improvements in this area would be very worthwhile, but I don't think they can or should be demanded at this stage.

Lots of work is commissioned/scheduled/planned/assigned within Tlon, particularly when it comes to userspace. That work should be explained when it's published/PR'd, but it seems too much to ask that its very existence be justified at that stage. Process complaints/critiques/proposals of this nature should be raised internally.

PR #1785, opened 2019-10-03
app: added chat store, hook, marks, and view.

loganallenc:
On the system level, Arvo is architecturally separated into modules with clear, single responsibilities. On the userspace level, the same does not hold. This has lead to hard-to-use data stores (hall) and applications that try to take on too many responsibilities at once (also hall).

As userspace continues to grow, needs of applications will overlap. Permission management is an obvious example of a task most applications will want to perform (forum membership, location sharing). The development costs of hand-rolling this for every new application are high. The UX costs, where permission systems in each application are subtly different, are even higher.

By defining attributes of good "userspace infrastructure", and specifying development patterns for working with that infrastructure, we hope to encourage a more modular userspace with a more consistent user experience.

Hall is being replaced by a new approach to userspace architecture outlined in this document: https://docs.google.com/document/d/1hS_UuResG1S4j49_H-aSshoTOROKBnGoJAaRgOipf54/edit?usp=sharing

Features included in this PR:

- Sending chat messages with message types (%text, %code, %url, %me (old @ messages))
- Setting a chatroom to read and keeping track of unread message count.
- Fine-grained permissions for read-access and write-access to chatrooms.
- Joining foreign chatrooms from the chat UI.
- Leaving chatrooms.
- Paginating message queries so as to have fast load times.
- Invites will come in a second PR from me once this is merged.
- The new %chat-cli will come in a separate PR from Mark afterwards, and then we will be able to remove %hall and %talk entirely.

This replaces the old chat.hoon file, but not %hall or %talk. This PR does not affect %hall or %talk.

Checklist for this PR to be ready to merge:

- [x] Finish changing wire format to /mailbox/~ship/path.
- [x] Test again with cross-ship communication.
- [x] Receive another review from either @ixv, @joemfb, or @philipcmonk.


vvisigoth:
FYI @loganallenc and I discussed keeping support for the `@` feature as a requirement here, given that it's a crowd favorite.

pkg/arvo/app/chat-view.hoon
```diff
@@ -0,0 +1,327 @@
...
+        (group-poke [%unbundle group-write])
+        (chat-poke [%delete path.act])
+    ==
+  ::
+  ==
+::
+++  peer-initial
+  |=  pax=path
+  ^-  (quip move _this)
+  ?.  =(src.bol our.bol)  !!
```
loganallenc: ?>(team:title instead)

vvisigoth:
@loganallenc reminder that we discussed namespacing old chat.hoon so that we have an "escape hatch" when we roll this out.

loganallenc:
@joemfb @philipcmonk @ixv This is ready for review. I've tested it in a variety of cases, and it works in all the cases I can think of.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,366 @@
...
+      [~ this]
+    =.  synced  (~(put by synced) path.act ship.act)
+    :_  (track-bone chat-path)
+    [ost.bol %peer chat-path [ship.act %chat-hook] chat-path]~
+  ::
+      %remove
+    =/  ship  (~(get by synced) path.act)
+    ?~  ship
+      [~ this]
+    ?:  &(=(u.ship our.bol) (team:title our.bol src.bol))
```
ixv: Why use team:title for one of these cases and not the other?

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,366 @@
...
+    ==
+    (pull-wire [%mailbox path.act])
+  ::
+  ==
+::
+++  peer-mailbox
+  |=  pax=path
+  ^-  (quip move _this)
+  ?>  ?=([* ^] pax)
+  ?.  (~(has by synced) pax)  !!
```
ixv: !!

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,366 @@
...
+    ?~  shp  ~
+    ?.  =(src.bol u.shp)  ~
+    [(chat-poke diff)]~
+  ::
+  ==
+::
+++  quit
+  |=  wir=wire
+  ^-  (quip move _this)
+  ~&  chat-hook-quit+wir
```
ixv: Should remove ~&'s

pkg/arvo/app/chat-store.hoon
```diff
@@ -0,0 +1,238 @@
...
+  ?>  (team:title our.bol src.bol)
+  ::  we now proxy all events to this path
+  [~ this]
+::
+++  peer-mailbox
+  |=  pax=path
+  ^-  (quip move _this)
+  ?>  (team:title our.bol src.bol)
+  =/  box=(unit mailbox)  (~(get by inbox) pax)
+  ?~  box  !!
```
ixv: got:by does this

pkg/arvo/app/permission-group-hook.hoon
```diff
@@ -67,22 +65,19 @@
   =/  permissions
     %-  silt
     %+  turn  ~(tap in permission-paths)
-    |=  [=path =kind]
-    path
+    |=([=path =kind] path)
```
ixv: Rather than turn set into a list and then back into a set you can use run:in

pkg/arvo/app/chat-hook.hoon
loganallenc: Changed this to an assert. The comment %chat-hook-reap covers the desire for a comment appended to the stacktrace.

pkg/arvo/app/chat-store.hoon
loganallenc: Modified here and in other places. Thanks.

pkg/arvo/app/chat-hook.hoon
loganallenc: These are valuable input for the user, as these cases are rare and only really initiated by semantic actions on other ships (a chatroom being deleted, or a user being banned).

pkg/arvo/app/permission-group-hook.hoon
loganallenc: Thanks! Great idea. Will do.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,364 @@
...
+  |=  jon=json
+  ^-  (quip move _this)
+  (poke-chat-action (json-to-action jon))
+::
+++  poke-chat-action
+  |=  act=chat-action
+  ^-  (quip move _this)
+  ?>  ?=(%message -.act)
+  ::  local
+  :_  this
```
joemfb: It's strange to produce the core first when there are state updates below. I guess it works because `+this` is an arm and it's evaluation is deferred, but I don't think it's a good pattern -- it'd be better to put it after the state updates and move-construction at each "return" point. (I have no objections to the use of `:_` in such places)

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,364 @@
...
+    :_  (track-bone chat-path)
+    [ost.bol %peer chat-path [ship.act %chat-hook] chat-path]~
+  ::
+      %remove
+    =/  ship  (~(get by synced) path.act)
+    ?~  ship
+      [~ this]
+    ?:  &(=(u.ship our.bol) (team:title our.bol src.bol))
+      ::  delete one of our.bol own paths
+      :_
```
joemfb: There should be a blank `::` comment line after the comment, and then the `%_` should be at least indented. It's short enough that I'd be inclined to:

```
:_  %_  this
      synced  (~(del by synced) path.act)
      boned  (~(del by boned) [%mailbox path.act])
    ==
...
```

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,364 @@
...
+        synced  (~(del by synced) path.act)
+        boned  (~(del by boned) [%mailbox path.act])
+      ==
+      %-  zing
+      :~  (pull-wire [%mailbox path.act])
+          (delete-permission [%chat path.act])
+          ^-  (list move)
+          %+  turn  (prey:pubsub:userlib [%mailbox path.act] bol)
+          |=  [=bone *]
+          ^-  move
```
joemfb: You shouldn't need both this `^-  move` and the `^-  (list move)` -- either should be enough for both documentary purposes and the type system.

pkg/arvo/app/chat-hook.hoon
joemfb: NB, the comment comment applies broadly.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,364 @@
...
+    :_  this
+    [ost.bol %peer /permissions [our.bol %permission-store] /updates]~
+  [~ this(+<+ u.old)]
+::
+++  poke-noun
+  |=  a=*
+  ^-  (quip move _this)
+  ~&  synced
+  ~&  boned
+  [~ this]
```
joemfb: If this arm is just for debug printfs, it'd be better to require an explicit `%debug` or some-such payload.

Also, it's worth reiterating that `~&` is only for developer debug printfs. If the user needs to see something, that should be produced as a move (a `%flog` to %dill is simplest -- some notification channel remains desirable).

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,364 @@
...
+      %journal
+    :~  (permission-poke (sec-to-perm read-perm %black))
+        (permission-poke (sec-to-perm write-perm %white))
+    ==
+  ::
+      %mailbox
+    :~  (permission-poke (sec-to-perm read-perm %white))
+        (permission-poke (sec-to-perm write-perm %black))
+    ==
+  ::
```
joemfb: Spacing comments should only ever be infix in "sub-hoons".

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,364 @@
...
+  ^+  this
+  =/  bnd  (~(get by boned) wir)
+  ?^  bnd
+    this(boned (~(put by boned) wir (snoc u.bnd ost.bol)))
+  this(boned (~(put by boned) wir [ost.bol]~))
+::
+++  pull-wire
+  |=  pax=path
+  ^-  (list move)
+  ?>  ?=([* ^] pax)
```
joemfb: This assertion looks unnecessarily specific: `?=(^ pax)` should be enough as you only ever reference `t.pax`.

pkg/arvo/app/group-hook.hoon
```diff
@@ -184,35 +157,32 @@
...
-    =/  group-wire  [(scot %p ship) group-path]
-    :_  (track-bone group-wire)
-    [ost.bol %peer group-wire [ship %group-hook] group-path]~
-  ::  no-op
-  [~ this]
+  =^  =ship  wir
+    ?>  ?=([* ^] wir)
+    [(slav %p i.wir) t.t.wir]
+  ?.  (~(has by synced) wir)
+    ::  no-op
```
joemfb: this comment seems unnecessary, given how many no-ops there are.

pkg/arvo/lib/chat-json.hoon
```diff
@@ -0,0 +1,251 @@
...
+  =,  ^?  dejs-soft:format
+  =/  exp  ((ot expression+so ~) a)
+  %-  need
+  ?~  exp
+    [~ '' ~]
+  :+  ~  u.exp
+  =/  res  ((ot output+(ar dank) ~) a)
+  ?^  res
+    u.res
+  p:(mule |.([(sell (slap !>(..^zuse) (ream u.exp)))]~))  ::TODO  oldz
```
joemfb: This `+eval` currently allows scry, and should have a more useful subject -- see my comments in #1803

pkg/arvo/app/chat-hook.hoon
loganallenc: What state updates? The =. statements apply to arguments to the gate that are then passed in a move.

pkg/arvo/app/chat-hook.hoon
loganallenc: Thanks, this is better stylistically.

pkg/arvo/app/chat-hook.hoon
loganallenc: You're right, thanks.

pkg/arvo/app/group-hook.hoon
loganallenc: Good point, thanks.

pkg/arvo/app/chat-hook.hoon
loganallenc: Good point, thanks.

pkg/arvo/app/chat-hook.hoon
joemfb: Ah, I just saw the `=.` ... my bad.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,353 @@
...
+  =/  ship  (~(get by synced) path.act)
+  ?~  ship
+    ~
+  ?.  =(u.ship our.bol)
+    ~
+  ::  scry permissions to check if write is permitted
+  ?.  (permitted-scry [(scot %p src.bol) %chat (weld path.act /write)])
+    ~
+  =.  author.envelope.act  src.bol
+  =.  when.envelope.act  now.bol
```
joemfb: Stylistically, I'd recommend:

```
=:  author.envelope.act  src.bol
    when.envelope.act  now.bol
  ==
```

pkg/arvo/app/chat-hook.hoon
loganallenc: Good point. Thanks.

pkg/arvo/app/chat-hook.hoon
loganallenc: Thanks, made the change in all new files.

pkg/arvo/app/chat-hook.hoon
loganallenc: Thanks for noting. Removed the poke-noun arms. The other ~& references stayed for debugging purposes if users run into problems. I should look more into %flog.

pkg/arvo/lib/chat-json.hoon
loganallenc: This is a straight copy of %hall-json. I would certainly love it to be the best version of eval it could possibly be, but I don't understand how it works.

ohAitch: Expanding this out a bit, which lines are particulatly confusing?
```
1  ::  u.exp=@t
2  =<  p  ^-  (each tang tang)
3  %-  mule  |.
4  =-  [-]~  :: tank -> (list tank)
5  %-  sell
6  %+  slap  !>(..zuse)
7  (ream u.exp)
```

And yes, to reiterate my comments on the "mule suddenly has scry by default" PR: if this used to be secure, and the exact same code is no longer secure, I am annoyed.

pkg/arvo/app/chat-view.hoon
```diff
@@ -0,0 +1,328 @@
...
+    :_  this
+    :~  (chat-hook-poke [%remove path.act])
+        (group-poke [%unbundle group-read])
+        (group-poke [%unbundle group-write])
+        (chat-poke [%delete path.act])
+    ==
+  ::
+  ==
+::
+++  peer-initial
```
philipcmonk: What's here should work most of the time, but FWIW the interaction between `/initial` and `/updates` isn't quite perfect (at least as used in `subscription.js`).  If a message comes in after the subscription to `/initial` is processed but before the subscription to `/updates` is, then it'll be skipped altogether.

We've done "get limited backlog and future updates" poorly in the past.  The two guarantees you'd like to have are: (1) updates come on a single canonical path and (2) never miss an update after the first message in the backlog.  The old `/updates/<from-number>` approach violates (1), and this approach can occasionally violate (2).  I believe the platonic form is basically:

- subscribe to `/updates`.  This immediately gives you a message number `n`, representing the number of messages in the channel.  It also gives you any new updates as they come in.
- subscribe to `/initial/(scot %ud (sub n 100))/(scot %ud n)`, which responds immediately with messages between those two numbers.

This way, you're already listening to `/updates` by the time you're getting the backlog, so you'll never miss a message.  The disadvantage of this approach is that you have to do the subscriptions consecutively and potentially reorder messages if `/updates` gives you results before `/initial`, so the client code is a bit more complicated.

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -46,77 +41,57 @@ export class ChatScreen extends Component {
...
-        this.updateNumPeople();
         this.scrollToBottom();
         this.updateReadInterval = setInterval(
           this.updateReadNumber.bind(this),
           1000
         );
         this.updateReadNumber();
       });
-    } else if (!(state.station in props.configs)) {
+    } else if (Object.keys(props.inbox) > 0) {
```
jtobin: It looks like this will always be `false`.

pkg/interface/chat/src/js/api.js
```diff
@@ -28,131 +73,245 @@ class UrbitApi {
...
-      }]
+  chatHookAction(data, mark = "chat-hook-action") {
+    this.action("chat-hook", mark, data);
+  }
+
+  chatHookAddOwned(path, security) {
+    let data = {};
+    data['add-owned'] = {
+      path, security
+    };
```
jtobin: This is fine, but FWIW it might be simpler to just do:

```
let data = {
  add-owned: { path, security }
}
```

here and below.

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -170,104 +145,75 @@ export class ChatScreen extends Component {
...
-    let chatMessages = reversedMessages.map((msg, i) => {
-      // Render sigil if previous message is not by the same sender
-      let gamAut = ['gam', 'aut'];    
+    let pendingMessages = 
+	    props.pendingMessages.has(state.station) 
+      ? props.pendingMessages.get(state.station) : [];
+	
+    pendingMessages.map(function(value) {
+      return value.pending = true;
+    })
```
jtobin: Does this do what you expect?  It looks like you want to set the `pending` attribute, but a comparable example doesn't seem to work: 

```
> let foo = [{ p: false, q: true }, { p: true, q: false }]
> foo.map(function(value) { return value.p = true })
[ true, true ]
```

pkg/interface/chat/src/js/components/join.js
```diff
@@ -2,59 +2,37 @@ import React, { Component } from 'react';
...
 
   componentDidMount() {
     const { props } = this;
-    let station = props.match.params.ship + '/' + props.match.params.station;
+    let station = `${props.match.params.ship}/${props.match.params.station}`;
 
-    if (station in props.configs) {
-      props.history.push(`/~chat/${station}`);
+    if (station in props.inbox) {
+      props.history.push(`/~chat/room${station}`);
```
jtobin: Should this be `/~chat/room/${station}`?  Or alternatively should `station` be prefixed with a `/` (as it is below)?

pkg/interface/chat/src/js/components/lib/chat-input.js
```diff
@@ -329,4 +198,29 @@ export class ChatInput extends Component {
       </div>
     );
   }
+
+  render() {
+    const { props, state } = this;
+
+    let writePermission = props.permissions[`/chat${props.station}/write`];
```
jtobin: Just a minor note: it might be error-prone to include the `/` prefix in `props.station`.

pkg/interface/chat/src/js/components/lib/invite-element.js
```diff
@@ -0,0 +1,124 @@
...
+
+    let aud = [];
+    let isValid = true;
+    if (state.members.length > 2) {
+      aud = state.members.split(',')
+        .map((mem) => mem.trim())
+        .map(deSig)
+        .map((mem) => {
+          return `~${mem}`;
+        });
```
jtobin: I'm not sure if a JS runtime will optimise this away, but it may be more efficient to move from:

```
foo.map(f).map(g).map(h)
```

to

```
foo.map(x => h(g(f(x))))
```

Additionally, you don't need the braces and `return` in the last `.map`.  You can just do:

```
.map(mem => `~${mem}`)
```

pkg/interface/chat/src/js/components/lib/sidebar-item.js
```diff
@@ -25,14 +25,26 @@ export class SidebarItem extends Component {
...
+  getLetter(lett) {
+    if ('text' in lett) {
+      return lett.text; 
+    } else if ('url' in lett) {
+      return lett.url;
+    } else if ('code' in lett) {
+      return lett.code.expression;
+    } else {
+      return '';
+    }
```
jtobin: A sequence of ternary operators might be useful here:

```
  'text' in lett
? lett.text
: 'url' in lett
? lett.url
[..]
```

pkg/interface/chat/src/js/components/member.js
```diff
@@ -5,130 +5,117 @@ import urbitOb from 'urbit-ob';
...
+          </div>
+          <div className="w-50 fr pa2 pl3">
             <p className="label-regular gray mb3">
-              Everyone with permission to see this chat.
+              {modReadText}
             </p>
-            {listMembers}
+            { window.ship === deSig(props.match.params.ship) ?
+              (  <InviteElement
+                   path={`/chat${state.station}/read`}
```
jtobin: I've noticed that some paths are `/chat/room/station` while others are `/chat/station`.  This is probably intended, but it's worth verifying things behave as you expect.

pkg/interface/chat/src/js/components/new.js
```diff
@@ -51,91 +57,90 @@ export class NewScreen extends Component {
...
+    let aud = [];
+    let isValid = true;
+    if (state.invites.length > 2) {
+      aud = state.invites.split(',')
         .map((mem) => mem.trim())
-        .map(deSig);
+        .map(deSig)
+        .map((mem) => {
+          return `~${mem}`;
+        });
```
jtobin: Similar comment to my other `.map` reduce observation (ha ha, get it).

pkg/interface/chat/src/js/components/new.js
```diff
@@ -51,91 +57,90 @@ export class NewScreen extends Component {
...
-        this.setState({
-          inviteError: false,
-          idError: false,
-          success: true,
-          invites: ''
-        }, () => {
-          props.setSpinner(true);
-          props.api.chat(actions);
-        });
+    // TODO: don't do this, it's shitty
```
jtobin: Should it not be done, then? 😄

pkg/interface/chat/src/js/components/join.js
jtobin: (Alternatively, maybe `props.match.params.ship` contains a prefixing `/`, but this seems error-prone.)

pkg/interface/chat/src/js/components/member.js
jtobin: Yeah, after seeing more of this code I'm guessing it's intended.  But nonetheless, worth another look.

pkg/interface/chat/src/js/api.js
loganallenc: Maybe my syntax highlighter is weird, but I always thought that didn’t work.

pkg/interface/chat/src/js/components/lib/invite-element.js
loganallenc: Seems cleaner, will try it

pkg/interface/chat/src/js/components/lib/sidebar-item.js
loganallenc: I don’t really like the style of using nested ternaries in JS. I generally go for my JS to be “explicit, needlessly verbose, and dumb as bricks”

pkg/interface/chat/src/js/components/new.js
loganallenc: Lol will look into this

pkg/interface/chat/src/js/components/chat.js
jtobin: Are you sure?  It looks like `props.inbox` should always be an object. `Object.keys` returns an array, which you're comparing against `0` (hence false).  Should there be a `.length` here, or am I missing something?

```
> let foo = { one: 1, two : 2 }
> let bar = {}
> Object.keys(foo) > 0
false
> Object.keys(bar) > 0
false
```

pkg/interface/chat/src/js/components/chat.js
jtobin: My bad, this *mutates* `pendingMessages` and doesn't actually care about the returned value `[true, true]`.  Gotcha.

pkg/interface/chat/src/js/components/join.js
jtobin: Yeah fair enough.  This particular one may still be missing the prefix, though; note that `station` here is defined via:

```
    let station = `${props.match.params.ship}/${props.match.params.station}`;

    if (station in props.inbox) {
      props.history.push(`/~chat/room${station}`);
    }
```

but in `components/chat.js`, `station` has a manual `/` prefix:

```
station: `/${props.match.params.ship}/${props.match.params.station}`,
```

pkg/interface/chat/src/js/api.js
jtobin: You might be right in that you may need quotes around `add-owned`, but otherwise it's ok.

pkg/interface/chat/src/js/components/lib/sidebar-item.js
jtobin: That's fine too.  I like the ternaries because you get a single expression, rather than a bunch of statements, but I get the motivation behind the alternative.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,359 @@
...
+  ?-  -.diff
+      %keys    [~ this]
+      %config  [~ this]
+      %create  [~ this]
+      %read    [~ this]
+      %delete
+    ?.  (~(has by synced) path.diff)
+      [~ this]
+    :_  this(synced (~(del by synced) path.diff))
+    :-  (chat-poke diff)
```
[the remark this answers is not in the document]
loganallenc: Good point. I think it's unnecessary.

pkg/arvo/lib/chat-json.hoon
```diff
@@ -0,0 +1,251 @@
...
+::
+++  update-to-json
+  |=  upd=chat-update
+  =,  enjs:format
+  ^-  json
+  %+  frond  %chat-update
+  %-  pairs
+  :~
+    ::
+    ::  %message
```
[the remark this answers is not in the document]
loganallenc: removed all of them

pkg/interface/chat/src/js/components/chat.js
loganallenc: Ah, I should add a `.length`. Good catch!

pkg/interface/chat/src/js/components/join.js
loganallenc: Good catch. There was indeed an error here.

loganallenc: Yeah, this one in particular seems to be an error

pkg/interface/chat/src/js/api.js
loganallenc: aesthetic, closing

pkg/interface/chat/src/js/components/lib/invite-element.js
loganallenc: looks nasty but it works?

pkg/interface/chat/src/js/components/new.js
loganallenc: After looking at it, I think I should continue doing the shitty thing until we get designers to look at it and come up with a better UX around fine-grained permission models being generated.

For context, what it's doing is...

Parsing which security type the chatroom will be, then taking the ships in the invite box, inviting them and granting them permission to the channel.

So you can't create a channel that already has banned users. This seems fine to me and retains the semantic around how the New chatroom page has functioned, but it's not taking full advantage of the security features. Seems fine, if a bit ugly in the code.

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
loganallenc: removed and it all works properly. good catch. it was just no-oping.

pkg/arvo/app/chat-view.hoon
loganallenc: Seems like I'd much prefer to just not cut off the subscription to %chat-view, and continue ferrying updates along through a single %chat-view subscription, since %chat-view is already subscribed to the /updates path.

pkg/interface/chat/src/js/components/chat.js
loganallenc: Not only that, it needed to be changed to .length === 0 rather than > 0. Caught this while testing. Thanks!

pkg/arvo/app/chat-view.hoon
loganallenc: This approach was successful and cleaner. Thanks for the motivation to change it!

pkg/arvo/lib/chat-json.hoon
loganallenc: I don't know what mule, sell, or slap do (apart from a vague hand-wavey "compile"),  or what the first argument to slap is in line 6.

philipcmonk: Read bottom-to-top, this is:

- parse `u.exp` to its ast (`+hoon`)
- compile that ast against the type of `..zuse` (standard library) and then run it against that actual value as well, producing a vase (pair of type and value)
- pretty-print that vase to a tank
- add a null so it's a singleton list of tanks (tang)
- run all of the above in a virtual environment so that if it crashes deterministically (syntax error in +ream, compile error in +slap, runtime error in +slap, prettyprint error in +sell) we get `[%| stack-trace=tang]` and if it succeeds we get `[%& result=tang]`
- get the tail of that value (which is a tang in both cases)

This gets you a tang of the result of the code or a stack trace of its failure.

Not that it matters, it's mostly fine to cargo cult this.

joemfb: It is *mostly* fine, expect for my repeated requests that you (or @Fang-) replace it with the implementation in the comments of #1803. This is a) to ensure a consistent subject for evals, and b) to prevent scry.

As @ohAitch notes, the issue of scry in `+mule` has been somewhat haphazardly handled, and these APIs need to be refactored. I haven't done this yet, so we need to jump through some extra hoops.

loganallenc: @joemfb Would you be okay with this PR merging in the existing ++eval that Hall also uses, then in a subsequent PR merging in the new one?

joemfb: I guess, if there's some reason that the state-threading is just too onerous for now. At the very least, it should match the current hall-json:

https://github.com/urbit/urbit/blob/32052396d8a85ac7f2479e689dcce7ea2e266c9a/pkg/arvo/lib/hall-json.hoon#L575-L585

But the primary use-case of `+eval` in chat messages is sharing code snippets for support or pedagogy. It's not helpful in either situation for different clients to use different subjects.

Fang-: Ah, heads up, I just committed this to the chat-cli branch. 7f0142d

loganallenc: Now using the eval function in the hall-json file you mentioned @joemfb. Thanks for posting that Mark. In your branch, we can include the updated eval functions for both JS chat client and chat-cli.

pkg/arvo/app/chat-hook.hoon
joemfb: @loganallenc, the `+poke-noun` arms appear to still be present.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -0,0 +1,359 @@
...
+    %-  ~(gas by *(map [ship path] bone))
+    %+  turn  ~(tap by sup.bol)
+    |=([=bone anchor=[ship path]] [anchor bone])
+  %-  zing
+  %+  turn  ~(tap in who)
+  |=  check-ship=ship
+  ?:  (permitted-scry [(scot %p check-ship) pax])
+    ::  if ship is permitted to be subscribed, do nothing
+    ~
+  :: if ship is not permitted, quit their subscription
```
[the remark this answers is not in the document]
joemfb: This comment still needs a double space, and subsequent blank comment line.

FWIW, I don't think you need to comment both branches. The "if not permitted" explanation implies the "if permitted" explanation.

pkg/arvo/app/chat-hook.hoon
joemfb: The overly-specific pattern-match is still present.

pkg/arvo/app/chat-hook.hoon
loganallenc: Thanks for mentioning. The commit I just pushed removed them.

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
loganallenc: Thanks, changing.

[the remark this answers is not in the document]
loganallenc: It's been removed.

pkg/arvo/app/chat-hook.hoon
loganallenc: Removed cast in latest commit.

pkg/arvo/app/chat-hook.hoon
loganallenc: It's now been removed.

loganallenc:
@joemfb Will keep that in mind, thanks.

@jtobin This is ready for merge.

PR #1788, opened 2019-10-03
Remove Fuzz Testing Option

brendanhay:
Removes the `-f` (Fuzz testing) flag. My assumption is that this behavior is neither used nor implemented in a sufficiently useful manner, please correct me + add context if that is not the case.


brendanhay:
(There's also a call to seed the random number generator for fuzz testing under `main.c`)

PR #1793, opened 2019-10-04
Add Option to Disable Terminal/TTY Assumptions

brendanhay:
Currently `-d` (daemon mode) disables terminal/tty assumptions and calls `fork()` to background the daemon process. In the hosting environment the former is required/desirable as no terminal is attached/present for containerized ships - but the parent process is required to stay in the foreground, as PID 1 within the container.

This PR toggles the terminal guards under `term.c` using the new `-t` option, and keeps the existing `-d` option to enable the call to `fork()`, with the latter implying the former. That is:

*  `-d` calls `fork()` and disables terminal assumptions.
*  `-t` only disables terminal assumptions.

An additional guard on `-t` has been placed around the terminal column/row calculation in `term.c:u3_term_get_blew`, which is in response to a `SIGABRT` that occurs when no terminal is attached.

### DIFF
--- a/pkg/urbit/vere/term.c
+++ b/pkg/urbit/vere/term.c
@@ -99,7 +99,7 @@ u3_term_io_init()
 {
   u3_utty* uty_u = c3_calloc(sizeof(u3_utty));
 
-  if ( c3y == u3_Host.ops_u.dem ) {
+  if ( c3y == u3_Host.ops_u.tem ) {
     uty_u->fid_i = 1;
 
     uv_pipe_init(u3L, &(uty_u->pop_u), 0);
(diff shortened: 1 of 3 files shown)


pkg/urbit/vere/term.c
```diff
@@ -919,7 +919,7 @@ u3_term_get_blew(c3_l tid_l)
   c3_l           col_l, row_l;
 
   struct winsize siz_u;
-  if ( uty_u && (0 == ioctl(uty_u->fid_i, TIOCGWINSZ, &siz_u)) ) {
+  if ( c3y == u3_Host.ops_u.tem && uty_u && (0 == ioctl(uty_u->fid_i, TIOCGWINSZ, &siz_u)) ) {
```
joemfb: Please split this across multiple lines in the (inconsistently enforced) house style:

```
if ( (foo) &&
      (bar) &&
      (baz) )
{
  // stuff
}
```

Issue #1790, opened 2019-10-04
|unlink after crud kills the drum


Fang-:
Not sure what happened here, but seems worth documenting.

After committing a change to my app file, I get a `crud` as in #1481.

```
: /~zod/home/234/app/klat3/hoon
worker_send_replace 570758 crud
```

Knowing that app wasn't going to update again, I ran `|unlink %klat3` so I could start over with `%klat4`. But instead of cleanly unlinking, I got the following (comments mine, for conciseness):

```
[ ! /~zod/home/0/app/hood/hoon:<[118 3].[119 20]>
  ! /~zod/home/0/app/hood/hoon:<[119 3].[119 20]>
  ! /~zod/home/0/app/hood/hoon:<[119 9].[119 19]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[177 3].[180 20]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[177 16].[180 20]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[177 20].[177 27]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[476 3].[479 35]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[477 3].[479 35]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[477 7].[477 18]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[477 11].[477 18]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[359 3].[362 36]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[360 3].[362 36]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[361 3].[362 36]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[362 3].[362 36]>
  ! /~zod/home/0/lib/hood/drum/hoon:<[362 4].[362 36]>
  [%ap-lame %hood %diff-sole-effect-drum-phat]
]  ::  4 times
> |unlink %klat3
crud: %reap event failed
/~zod/home/0/app/hood/hoon:<[118 3].[119 20]>
/~zod/home/0/app/hood/hoon:<[119 3].[119 20]>
/~zod/home/0/app/hood/hoon:<[119 9].[119 19]>
/~zod/home/0/lib/hood/drum/hoon:<[184 3].[187 49]>
[%drum-unauthorized [%our ~zod] %src ~zod]
[...]

Issue #1481, opened 2017-12-18
Old version of application stays running after update


Fang-:
I (on ~palfun-foslup) and others (like ~tonlur-sarret) are running into an issue where we have received the latest version of talk, but don't seem to be running it. We can see changed/added application logic in the hoon file (in both base and home), but talk acts as if that weren't there.

We've tried editing the file manually, to make it re-sync. And it does, but still doesn't seem to affect the running code.

May or may not be related to urbit/arvo#501.

vvisigoth:
@jtobin @philipcmonk any insight into this?

philipcmonk:
I haven't seen it in a few months, but I wouldn't be surprised if it's still lurking around somewhere.  I don't know what could cause that subscription to fall off.

Fang-:
Perhaps useful, @matildepark had reported this to me previously, and I've now experienced it myself. This issue seems to occur alongside (or because of) a `crud`.

Consider this dojo output, while updating an app whose `+prep` contains a printf that always fires.

```
> |commit %home
>=
: /~zod/home/129/app/klat/hoon
worker_send_replace 219236 crud
> |commit %home
>=
: /~zod/home/130/app/klat/hoon
```

We'd expect to see the printf here after both edits, but we don't. It had been firing fine previously.  
Incidentally, the `crud` happened on/after the first `|commit` that resulted in successfully-compiling app code, after a long sequence of error-hunting commits.

philipcmonk:
Behn eats %drip errors, which is almost certainly what is happening there.
I've fixed this on the philip/mall-real branch, but that won't land in this
breach.

I'm out for the rest of this week, but it would be easy to add a handler
for %wake in behn that actually printed an error if it exists.

Fang-:
> Behn eats %drip errors, which is almost certainly what is happening there.

Can confirm:

```
> |commit %home
>=
: /~zod/home/61/app/stream/hoon
worker_send_replace 99410 crud
drip failed
nest-fail
/~zod/home/~2019.10.16..05.49.49..104e/sys/vane/gall:<[2.345 9].[2.345 51]>
/~zod/home/~2019.10.16..05.49.49..104e/sys/vane/gall:<[2.344 9].[2.345 51]>
/~zod/home/~2019.10.16..05.49.49..104e/sys/vane/gall:<[2.343 9].[2.345 51]>
/~zod/home/~2019.10.16..05.49.49..104e/sys/vane/gall:<[2.342 7].[2.371 27]>
[...]

PR #1795, opened 2019-10-04
behn: print error message if %drip event fails

joemfb:
Per @philipcmonk's suggestion in #1481, as discussed in #1790. This change *does not* handle the error of a failing %drip, as there's no way to route that error notification right now. But printing it will at least let us debug...


philipcmonk:
LGTM, don't have my github password handy so can't formally approve it.

Issue #1778, opened 2019-09-29
bug: issue in +apt:to checking correctness of a +qeu


yosoyubik:
Checking the correctness of a `+qeu` returns false in some cases:
```
~(apt to (~(gas to *(qeu @)) ~[1 2 3]))
>  %.n
> (~(gas to *(qeu @)) ~[1 2 3])
[n=2 l={3} r={1}]
> (mor 3 1)
%.n
> (mug (mug 1))
1.405.103.437
> (mug (mug 3))
1.576.941.407
```

I believe that the issue comes when the order of insertion places elements with lower double mug hash on the right of the tree (compared to the left). `+apt:to` compares with `+mor` between siblings, but ignores the parent `(mor n.l.a n.r.a)`. Here is how is done:

```
++  apt
  ^-  ?
  ?~  a  &
  |-  ^-  ?
  ?~  l.a  &
  ?~  r.a  &
  &((mor n.l.a n.r.a) $(a l.a) $(a r.a))
```
A longer explanation of can be found in this #1777 with a proposed arm that checks the priority order between parent-children. Here is the code:

```
++  check
      |=  a=(qeu)
      ^-  ?
      ?~  a  &
      |-  ^-  ?
      ?~  l.a  &
      ?~  r.a  &
      ?&  ?&((mor n.a n.l.a) $(a l.a))
          ?&((mor n.a n.r.a) $(a r.a))
      ==
```
If this is confirmed to be the case I can create a separate PR with this fix.

belisarius222:
That old code definitely looks wrong to me.  You are correct that vertical order only applies between parent node and children, not between siblings.

+qeu horizontal ordering is hysteretic.  This means sibling ordering is not a function solely of the present contents and is therefore unverifiable in +apt.  Vertical ordering can be checked, though, since that does not depend on history.

I think your code misses some cases, though.  If the left subtree is ~ but the right subtree is not, we still need to check `(mor n.a n.r.a)`.  Same is true with left and right switched.

Also, stylistically, you should know that ?& is n-ary, not binary, so you don't need to nest the runes.

Look at +apt:in.  Your code should probably look like a shorter version of that, since the vertical order checks are the same; you're just deleting the horizontal order checks.

Your code should replace the |to core's +apt arm instead of introducing a new function, since |by and |in also use +apt as their validity checks and this change would be fixing a bug, not changing functionality.

joemfb:
[...]

PR #1799, opened 2019-10-06
hoon: replaces manual +tree cons with explicit %= updates in +in/+by/+to

yosoyubik:
See [here](https://github.com/urbit/urbit/issues/1779#issuecomment-537616988) for context.

These arms have been tested with the test suite (waiting comments from reviewers; subject to be modified/expanded) provided in this PR #1777


pkg/arvo/sys/hoon.hoon
```diff
@@ -1286,10 +1286,10 @@
       ?.  (mor n.a n.b)
         $(a b, b a)
       ?:  =(n.b n.a)
-        [n.a $(a l.a, b l.b) $(a r.a, b r.b)]
+        a(l $(a l.a, b l.b), r $(a r.a, b r.b))
       ?:  (gor n.b n.a)
-        %-  uni(a $(a l.a, b [n.b l.b ~]))  $(b r.b)
-      %-  uni(a $(a r.a, b [n.b ~ r.b]))  $(b l.b)
+        %-  uni(a $(a l.a, b b(r ~)))  $(b r.b)
```
belisarius222: `$(a l.a, b b(r ~))` can be further reduced to `$(a l.a, r.b ~)`.  That would take us all the way to making the smallest possible incision into these nouns, so I think it would be better -- maybe a bit clearer too.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1349,15 +1349,13 @@
...
       ?:  (mor n.a n.b)
         ?:  =(n.b n.a)
-          [n.b $(a l.a, b l.b) $(a r.a, b r.b)]
+          b(l $(a l.a, b l.b), r $(a r.a, b r.b))
         ?:  (gor n.b n.a)
-          $(a [n.a $(a l.a, b [n.b l.b ~]) r.a], b r.b)
-        $(a [n.a l.a $(a r.a, b [n.b ~ r.b])], b l.b)
-      ?:  =(n.a n.b)
-        [n.b $(b l.b, a l.a) $(b r.b, a r.a)]
+          $(a a(l $(a l.a, b b(r ~))), b r.b)
```
belisarius222: These four changed lines can be further reduced to `$(l.a $(a l.a, r.b ~), b r.b)`, etc.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1515,15 +1513,15 @@
         ~
       ?:  (mor p.n.a p.n.b)
         ?:  =(p.n.b p.n.a)
-          [n.b $(a l.a, b l.b) $(a r.a, b r.b)]
+          b(l $(a l.a, b l.b), r $(a r.a, b r.b))
         ?:  (gor p.n.b p.n.a)
-          %-  uni(a $(a l.a, b [n.b l.b ~]))  $(b r.b)
-        %-  uni(a $(a r.a, b [n.b ~ r.b]))  $(b l.b)
+          %-  uni(a $(a l.a, b b(r ~)))  $(b r.b)
```
belisarius222: These can be reduced just like in `+int:in`.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1583,19 +1581,19 @@
...
     =+  f=$(a r.a, b -.e)
-    [-.f [n.a +.e +.f]]
+    [-.f a(l +.e, r +.f)]
   ::
   ++  run                                               ::  apply gate to values
     |*  b/gate
     |-
     ?~  a  a
-    [n=[p=p.n.a q=(b q.n.a)] l=$(a l.a) r=$(a r.a)]
+    a(n n.a(q (b q.n.a)), l $(a l.a), r $(a r.a))
```
belisarius222: `a(q.n (b q.n.a), l $(a l.a), r $(a r.a))`

This line replaces so much of the noun that it might actually be better as an autocons:
`[[p.n.a (b q.n.a)] $(a l.a) $(a r.a)]`

pkg/arvo/sys/hoon.hoon
```diff
@@ -1773,23 +1766,23 @@
...
+      l.a(r $(l.a r.l.a))
+    r.a(l $(r.a l.r.a))
   ::
-  ++  nap                                               ::  removes head
+  ++  nap                                               ::  removes root
     ?>  ?=(^ a)
     ?:  =(~ l.a)  r.a
     =+  b=get(a l.a)
-    bal(a ^+(a [p.b q.b r.a]))
+    bal(a ^+(a a(n p.b, l q.b)))
```
belisarius222: I wonder if you could shorten this:
`bal(n.a p.b, l.a q.b)`

pkg/arvo/sys/hoon.hoon
yosoyubik: Ah yes, some of those `b b(...` were bothering me so that's definitively better.  We can push it a bit further and reduce this `$(a a(l $(a l.a, r.b ~)), b r.b)` to this `$(l.a $(a l.a, r.b ~), b r.b)`

pkg/arvo/sys/hoon.hoon
ohAitch: In that case I'd suggest the autocons version with faces this is originally replacing

pkg/arvo/sys/hoon.hoon
yosoyubik: Yeah, this seems to work.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1583,19 +1581,19 @@
...
     ?~  a  a
-    [n=[p=p.n.a q=(b q.n.a)] l=$(a l.a) r=$(a r.a)]
+    a(n n.a(q (b q.n.a)), l $(a l.a), r $(a r.a))
   ::
   ++  rut                                               ::  apply gate to nodes
     |*  b/gate
     |-
     ?~  a  a
-    [n=[p=p.n.a q=(b p.n.a q.n.a)] l=$(a l.a) r=$(a r.a)]
+    a(n n.a(q (b p.n.a q.n.a)), l $(a l.a), r $(a r.a))
```
yosoyubik: This one looks similar to `+rut`, we are replacing too many things

pkg/arvo/sys/hoon.hoon
```diff
@@ -1641,25 +1637,23 @@
...
-        :+  [p.n.a (meg p.n.a q.n.a q.n.b)]
-          $(b l.b, a l.a)
-        $(b r.b, a r.a)
+          $(a a(l $(a l.a, b b(r ~))), b r.b)
+        $(a a(r $(a r.a, b b(l ~))), b l.b)
       ?:  (gor p.n.a p.n.b)
-        $(b [n.b $(b l.b, a [n.a l.a ~]) r.b], a r.a)
-      $(b [n.b l.b $(b r.b, a [n.a ~ r.a])], a l.a)
+        $(b b(l $(b l.b, a a(r ~))), a r.a)
+      $(b b(r $(b r.b, a a(l ~))), a l.a)
```
yosoyubik: I'm replacing all instances of formats like `b b(...`, in this case it gets reduced to this: `$(r.b $(b r.b, l.a ~), a l.a)`

ohAitch:
"To use new nock edit operation" is an interesting way to put it, since on
one level this is rewriting them to use the hoon %= operation which is
anything but new(iirc it's one of the core runes that has existed since
hoon was called watt), and on another level most of these arms are jetted.
Definitely a good stylistic change though! (Except in a couple places/could
be taken further in others, see comments by Ted)

yosoyubik:
thanks @ohAitch I'm expanding it further, following Ted's comments and cases like this: https://github.com/urbit/urbit/pull/1799#pullrequestreview-297852376

About the "new Nock edit operation" I claim ignorance and refer to the background info: https://github.com/urbit/urbit/issues/1779#issuecomment-537616988

pkg/arvo/sys/hoon.hoon
belisarius222: @ohAitch I think you're right, especially since this is wet.  The type system might actually need the faces.

ohAitch: "Need" is a social construct; but it's certainly the case that if you call the faceless `run` two times in a row, then the second one won't be able to pull `q.n.a`, because the first didn't give it a `q` or an `n` 😉

(Unless you cast `a` back to a map in the middle, which we certainly find ourselves needing to do on occasion, though I think usually for more subtle reasons than this one)

ohAitch: I suppose actually, I completely agree with framing things like iron core nesting as the type system "demanding" or "compelling" specific behavior; but in the particular case of putting faces on data structures in wet gates(or dry gates samples, or other casts), the type system is just doing exactly what you told it to.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1641,25 +1637,23 @@
         b
       ?:  (mor p.n.a p.n.b)
         ?:  =(p.n.b p.n.a)
-          [n.b $(a l.a, b l.b) $(a r.a, b r.b)]
+          :+  [p.n.a (meg p.n.a q.n.a q.n.b)]
```
belisarius222: I don't like that the `=(p.n.b p.n.a)` check is only run if `+mor` produces `%.y`.  Seems somewhat brittle w.r.t. the details of `+mor`.  If it started to behave more like `+lth` than `+lte`, could that change the behavior here?  If the keys are equal, I don't think we should care about where exactly `+mor` switches from `%.n` to `%.y`.

I think we should first check equality, and then only run `+mor` if the keys aren't equal.

pkg/arvo/sys/hoon.hoon
belisarius222: Sure, I mostly meant that in practice we're better off putting faces on the product so that passing the result into other wet code will be less likely to run into problems.  I think that's pretty similar to what you're saying.

pkg/arvo/sys/hoon.hoon
yosoyubik: That makes sense. Philip expressed similar concerns when we talked about this issue. I'll update the `+uni` arms as well

yosoyubik: I have just pushed this change. I believe I have addressed your other comments as well, but let me know if i missed something. Thanks!

pkg/arvo/sys/hoon.hoon
```diff
@@ -1583,13 +1581,13 @@
...
     =+  f=$(a r.a, b -.e)
-    [-.f [n.a +.e +.f]]
+    [-.f a(l +.e, r +.f)]
   ::
   ++  run                                               ::  apply gate to values
     |*  b/gate
     |-
     ?~  a  a
-    [n=[p=p.n.a q=(b q.n.a)] l=$(a l.a) r=$(a r.a)]
+    [[p.n.a (b q.n.a)] $(a l.a) $(a r.a)]
```
yosoyubik: Forgot to add back the faces here, updating it asap

jtobin:
@yosoyubik I see one test failure after resolving the pill conflict:

```
FAILED  /sys/hoon/molds/test-qeu
expected: ~
actual: [~ {97 98 100 99}]
```

Is this expected here?

yosoyubik:
I'm wondering if that test was made with the previous implementation of `+apt:to` in mind. The longer explanation is here: https://github.com/urbit/urbit/issues/1778

Tl;DR: The test is expecting that qeu to not be correct... when it is.

The test: `[98 [97 ~ ~] [100 ~ [99 ~ ~]]]` is a correct queue if we look at vertical ordering: `(mor 98 97)`, `(mor 98 100)` & `(mor 100 99)` all return `%.y` so vertical ordering is correct.

The previous implementation of `+apt:to` checked only horizontal ordering between siblings, in this case that would fail: `(mor 97 100)` returns `%.n`, but that is not how you check correctness of hoon treaps.

I'm probably wrong about half of this, but I think the solution would be to modify that test with a proper "incorrect" `+qeu`, for example: `((soft (qeu)) [97 [98 ~ ~] [100 ~ [99 ~ ~]]])`. Vertical ordering is not correct with any of the children.

ohAitch:
`(soft (tree))` probably, since `(qeu)` is `$|`d with `+-apt`

yosoyubik:
It might be better here to roll back `+apt:to` and restore its brokenness, and update the conflicting test, together with the others, as part of this PR: #1777, which has the correct implementation of `+apt:to` ?

jtobin:
@yosoyubik Works for me! 😄

yosoyubik:
Ok! I'll restore `+apt:to` then. Hopefully this is the last time I force-push, at some point this won't hold it anymore :D

joemfb:
It seems to me that it'd be best to merge tests before (or at the same time as) changes to the operations that are being tested. Just a procedural note -- I trust that these changes are passing the tests in question.

ohAitch:
Re: jets, do we _have_ "replace r of prq, making a copy first if you don't hold the only refcount and the replacement isn't `.=`" helper functions exposed by libnoun? I'm sure the bytecode interpreter does something along those lines internally given the strength of the push for nock 4.

joemfb:
@ohAitch, the function in question is `u3i_edit(noun, axis, replacement)`. It's quite straightforwardly the "nock opcode `%10`" function: replace `axis` of `noun` with `replacement`, mutating rather than constructing a copy of each cell leading to `axis` where the refcount is one.

yosoyubik:
Thanks @joemfb I can see how phrasing the change like this can be misleading. I'll reword the commit/PR. How about this: "autocons in treap arms replaced with %=".

> It seems to me that it'd be best to merge tests before (or at the same time as) changes to the operations that are being tested. Just a procedural note -- I trust that these changes are passing the tests in question.

I tested these changes against the test suite from #1777... and they were failing due to the `+apt:to` arm not checking properly for queues. Modifying it here produces a test failure (there is a test in `%/tests/sys/hoon/molds` that expects a proper queue to be formed incorrectly; see https://github.com/urbit/urbit/pull/1799#issuecomment-539185254) so I was thinking to fix that test in #1777 since it seems to be within the scope of the PR. 

I agree that this would probably be better if it's merged together with #1777

joemfb:
That failing test is my fault -- I added it along with the incorrect `+apt:to` implementation. It should be replaced at the same time as `+apt:to` (just a simple check that the `+qeu` mold calls `+apt:to` when clammed).

As for "autocons", I meant to mention it too, as it's also misleading or just wrong here. A hoon *cons* (ie, cell-construction) is properly `:-`, while `[a b c]` is irregular syntax for `:*(a b c)` -- n-ary cons. "Autocons" is our name for the fact that any pair of nock formulas is a valid nock formula, any pair of `+hoon` AST nodes is a valid AST node, &c. In other words, structures with "autocons" semantics support an implicit cons operation. If you look in `+open:ap`, the `+hoon` macro-expansion/desugaring in the compiler, you'll see that the "cons" runes (`:-`, `:*`, &c) simply expand to cells. (This is the primary use-case for `$^` molds -- `+nock`, `+hoon`, and other structures with autocons semantics are built from a `$%` within a `$^`.)

All of that to say that "autocons" doesn't quite describe the patterns being replaced here -- it's true in a technical sense with regards to the implementation within the compiler, but the code in question consists of explicit cons runes/irregular-forms. I'd say something like "replaces manual +tree cons with explicit %= updates in +in/+by/+to" (carefully worded to acknowledge your point that `+qeu` is not a treap!).

It does seem simplest to just combine this with #1777. The other option is target both PRs against an integration branch which can be PR'd against master when complete (that's our usual approach for larger changes that require intermediate review). I'll defer to @jtobin.

jtobin:
> It does seem simplest to just combine this with #1777. The other option is target both PRs against an integration branch which can be PR'd against master when complete (that's our usual approach for larger changes that require intermediate review). I'll defer to @jtobin.

Either of these are perfectly fine, to me.

There's also the octopus merge option.. 🤔

jtobin:
(I've merged both of these branches together, updated pills, tweaked the "bad test" as suggested, and tested & merged the result to master.)

PR #1807, opened 2019-10-09
Chat: minor fixes re feedback

Fang-:
As per @jtobin's feedback. See commit messages for changelog.


loganallenc:
This looks good to me

PR #1809, opened 2019-10-09
Claz: batch invites

Fang-:
For ship distribution purposes, we want to be able to send out large amounts of planet invites (using Delegated Sending). This PR implements logic in Claz for generating those transactions.

It takes a `.txt` file as input, with every line being `~ship,~ticket,0xaddress`. To generate these for some planets under some star, a generator is provided.  
`|claz-invites ~marzod 1 10 %/invites/txt` does this for ~marzod's first through tenth child.

That file can be taken out of the Urbit and used to, for example, print out invite wallets.

Those wallets won't be useful until you've sent the matching invite transactions.  
`:claz [%generate %/out/eth-txs %ropsten 0xsender %invites ~palfun-foslup %/invites/txt]` generates transactions for sending those using ~palfun-foslup as the invite-sender. (Make sure they have the correct/enough invites!)

Lastly, this sneaks in an addition to keygen.hoon that matches what we've done in the js version.

### DIFF
--- a/pkg/arvo/app/claz.hoon
+++ b/pkg/arvo/app/claz.hoon
@@ -42,6 +42,11 @@
       ::  %deed: deed ships based on json, assumes spawnable
       ::
       [%deed deeds-json=cord]
+      ::  %invites: sendPoint for every ship in ship,ticket,owner file
+      ::
+      ::    to generate such a file, try +claz|invites ~star 1 10
+      ::
+      [%invites as-who=ship file=path]
       ::  %lock-prep: prepare for lockup by transfering ships to the ceremony address
       ::
       [%lock-prep what=(list ship)]
@@ -77,6 +82,8 @@
       [%transfer-ship who=ship to=address]
       [%set-transfer-proxy who=ship proxy=address]
       [%adopt who=ship]
+    ::
+      [%send-point as=ship point=ship to=address]
   ==
 ::
 ::  monadic structures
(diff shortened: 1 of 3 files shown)


vvisigoth:
This looks cool! To be clear, how would I then submit those transactions?

Fang-:
You can use /app/send-txs for signing and sending. From its inline docs:

```
::  sign txs for gasses of 2 and 11 gwei; (~ for default gwei set)
::  store at path
::    :send-txs [%sign %/txs %/txs/eth-txs %/pk/txt ~[2 0]]
::
::  send all but first 50 txs from path
::    :send-txs [%send %/txs/txt 50]
```

vvisigoth:
Just need to make sure that the ship has the proper creds.
—
~poldec-tonteg
http://urbit.org

vvisigoth:
This LGTM, but waiting for @philipcmonk on code review, since I'm not familiar with these codez

pkg/arvo/app/claz.hoon
```diff
@@ -484,6 +493,8 @@
     %transfer-ship  (transfer-ship:dat +.call)
     %set-transfer-proxy  (set-transfer-proxy:dat +.call)
     %adopt  (adopt:dat +.call)
+  ::
```
philipcmonk: Why the extra `::`?  Is it because it's not an azimuth-native call?

Fang-: Yeah, because it goes to a contract different than the Ecliptic. Note that we separate it out [in the type definition](https://github.com/urbit/urbit/pull/1809/files#diff-bd1093e2e574caa62df0552afecc07fcR85-R86) as well.

Fang-:
> It'd be nice if there was a way to sign raw transactions that didn't require that. I wonder if there are off-the-shelf tools for this

Hmm, most search results that come up are "just do it using this code". It wouldn't be too hard to add to Bridge, I think. It already has all the login and (most of the) signing logic we want.  
@vvisigoth Is there room in Bridge's feature-space for this, or would you rather a separate tool?

vvisigoth:
I've long wanted the feature in Bridge to sign/send arbitrary tx. Added to
the backlog!

A
—
~poldec-tonteg
http://urbit.org

vvisigoth:
Is there any way with claz to make sure that these invites are not for already-claimed planets?

Fang-:
Not currently, but I should try patching that into the invites procedure, seems good to have.

Issue #1811, opened 2019-10-09
%invalid-outstanding-connection - %eyre heartbeat problem?


loganallenc:
@jtobin @eglaysher @pkova Not sure if this relates to the heartbeat PR, but I've never seen this state before and my console is spamming them repeatedly. It's a no-op, so it isn't breaking anything, but it's annoying and a sign that something is wrong.

jtobin:
Are you seeing this on mainnet?  #1687 hasn't actually made it into a release yet, so if you're seeing this there, it must be due to something else.

loganallenc:
Seeing it on master.

jtobin:
Should be fixed in #1814.

PR #1687, opened 2019-09-02
eyre: add channel \n heartbeat every 20 seconds

pkova:

PR #1814, opened 2019-10-10
eyre: fix missing cancel-heartbeat-move cases

### LINKED ISSUE #1812: Eyre's %born event fails, so can't connect via http
I killed my urbit then brought it up again, and the `%born` event is failing, so the http server doesn't get set up properly.  Trace: https://pastebin.com/YRpWs4h7

Because of this, I can't access landscape.

pkova:
Addresses #1811.

I can reproduce the issue by shutting down a pier before disconnecting from landscape and then restarting it. This results in `%invalid-outstanding-connection` every 20 seconds since the heartbeat move is never cancelled.

This is due to me being sloppy and only canceling heartbeat moves on `%disconnect`. Now we cancel heartbeats in `on-cancel-request` (fixing the above case, this is called on %born) and in `on-channel-timeout`. Let me know if there are other cases that I have missed.

This might fix #1812 as well, not sure of how to reproduce that one, though.

I'm still having trouble with nix (as mentioned [here](https://github.com/urbit/urbit/pull/1687#issuecomment-538802626))  so I can't run tests nor create pills locally. I suspect some tests will fail here, so I'll just force push fixes here if needed.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1148,9 +1148,22 @@
       ::
       ~&  [%canceling-cancel duct]
       ::
+      =/  maybe-session
+        (~(get by session.channel-state.state) u.maybe-channel-id)
+      ?~  maybe-session  [~ state]
+      ::
+      =/  heartbeat-cancel=(list move)
+        ?~  heartbeat.u.maybe-session  ~
+        :~  %^  cancel-heartbeat-move
+              u.maybe-channel-id
+            date.u.heartbeat.u.maybe-session
+          duct.u.heartbeat.u.maybe-session
+        ==
+      ::
       =/  expiration-time=@da  (add now channel-timeout)
       ::
-      :-  [(set-timeout-move u.maybe-channel-id expiration-time) moves]
+      :-  %+  weld  heartbeat-cancel
+        [(set-timeout-move u.maybe-channel-id expiration-time) moves]
       %_    state
           session.channel-state
         %+  ~(jab by session.channel-state.state)  u.maybe-channel-id
@@ -1158,7 +1171,7 @@
         ::  if we are canceling a known channel, it should have a listener
         ::
         ?>  ?=([%| *] state.channel)
-        channel(state [%& [expiration-time duct]])
+        channel(state [%& [expiration-time duct]], heartbeat ~)
       ::
           duct-to-key.channel-state
         (~(del by duct-to-key.channel-state.state) duct)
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1613,6 +1626,9 @@
       |=  channel-id=@t
       ^-  [(list move) server-state]
       ::
+      ?~  connection-state=(~(get by connections.state) duct)
```
pkova: This connection state checking shouldn't be needed with the new cancels, but this should fix the infinite behn loop state experienced by at least @loganallenc from my previous PR #1687.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1148,17 +1148,30 @@
       ::
       ~&  [%canceling-cancel duct]
       ::
+      =/  maybe-session
+        (~(get by session.channel-state.state) u.maybe-channel-id)
+      ?~  maybe-session  [~ state]
+      ::
+      =/  heartbeat-cancel=(list move)
+      ?~  heartbeat.u.maybe-session  ~
```
eglaysher: Indent this block under the `=/`

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1645,6 +1661,14 @@
               session.channel-state
             (~(del by session.channel-state.state) channel-id)
           ==
+      =/  heartbeat-cancel=(list move)
```
eglaysher: Same here.

loganallenc:
Thanks for jumping on this @pkova.

pkg/arvo/tests/sys/vane/eyre.hoon
```diff
@@ -1038,9 +1038,14 @@
       call-args=[duct=~[/http-get-open] ~ %cancel-request ~]
       ^=  expected-moves
         ^-  (list move:http-server-gate)
-        ::  closing the channel restarts the timeout timer
+        ::  closing the channel cancels the sse heartbeat
+        ::  (initialized in results5 above) and restarts the timeout timer
```
loganallenc: remove the '5' character please.

pkova:
Fixed the indentation @eglaysher. Also, it looks like I don't have sufficient github permissions on this repository to request reviewers on my PR:s. I'll try to mention the relevant people in the PR text from now on.

jtobin:
> I'll try to mention the relevant people in the PR text from now on.

I'll make a mental note to tag appropriate reviewers for PRs I spot
needing them in the future.

PR #1817, opened 2019-10-11
ames: skip timers if unix duct is unset

joemfb:
This fixes a bug whereby %behn timers were being set on an empty duct, and the subsequent %wake was being routed to unix as an effect. These are safe to drop, as any subsequent messages sent or packets received will reset the timer, which will pick up all "pending" packets. (There's also a fallback %ames timer in vere, but it's presence doesn't change this analysis.)


belisarius222:
LGTM

PR #1818, opened 2019-10-11
app: added permission hook and %chat-view join action

loganallenc:
This commit adds the capability to sync permissions from chat hosts to their foreign clients. Clients can now know in their UI whether the chat they're in is read-only or not, etc.

Have tested that this successfully quits subscriptions, works with adding and removing permissions repeatedly, etc.

As a part of this, I also cleaned up the API file in the chat JS and removed all unused commands, as well as added a new %join command to %chat-view that syncs down both the foreign chatroom and the foreign permissions.

Note: Updated %chat-cli's join command to work in the same way that the client JS does by sending a %join action to %chat-view.

Thanks!


pkg/arvo/app/permission-hook.hoon
```diff
@@ -0,0 +1,282 @@
...
+      (~(put in (~(got by access-control) access.act)) owned.act)
+    =.  access-control
+      (~(put by access-control) access.act access-paths)
+    =/  perm-path  [%permission owned.act]
+    :_  (track-bone perm-path)
+    [ost.bol %peer perm-path [our.bol %permission-store] perm-path]~
+  ::
+      %add-synced
+    ?>  (team:title our.bol src.bol)
+    =/  perm-path  [%permission path.act]
```
Fang-: I'd put this closer to the move construction down below.

pkg/arvo/app/permission-hook.hoon
```diff
@@ -0,0 +1,282 @@
...
+    ~
+  (quit-subscriptions kind pax who u.access-paths)
+::
+++  handle-foreign
+  |=  diff=permission-update
+  ^-  (quip move _this)
+  ?-  -.diff
+      %create  (change-foreign-permission path.diff diff)
+      %add     (change-foreign-permission path.diff diff)
+      %remove  (change-foreign-permission path.diff diff)
```
Fang-: Does this not work if you check on `?(%create %add %remove)` all at once? I think the axes of `path.` line up for those cases, so iirc should be fine.

pkg/arvo/sur/permission-hook.hoon
```diff
@@ -0,0 +1,15 @@
+|%
++$  permission-hook-action
+  $%  ::  %add-owned: make a permission set accessible to foreign ships
+      ::  who are allowed by that same permission set
```
Fang-: +1 for /sur comments. (:

pkg/arvo/app/permission-hook.hoon
```diff
@@ -0,0 +1,282 @@
...
+  |=  [=bone *]
+  [bone %diff %permission-update upd]
+::
+++  track-bone
+  |=  wir=wire
+  ^+  this
+  =/  bnd  (~(get by boned) wir)
+  ?^  bnd
+    this(boned (~(put by boned) wir (snoc u.bnd ost.bol)))
+  this(boned (~(put by boned) wir [ost.bol]~))
```
Fang-: From looking at `+pull-wire` below (the only other place `boned` gets read) it doesn't seem like order matters here. So you could replace the above with `(~(add ja boned) wir ost.bol)` and be done with it. (It appends to the head of the list, but again, don't think that matters.)

pkg/arvo/app/permission-hook.hoon
```diff
@@ -0,0 +1,282 @@
...
+::
+++  pull-wire
+  |=  pax=path
+  ^-  (list move)
+  ?>  ?=([* ^] pax)
+  =/  bnd  (~(get by boned) pax)
+  ?~  bnd  ~
```
Fang-: Similarly, you could consider using `(~(get ja boned) pax)` here. (Or rather, `%+  turn  (~(get ja boned) pax)` below.)

pkg/arvo/app/permission-hook.hoon
```diff
@@ -0,0 +1,282 @@
...
+    ?~  owner-access
+      [~ this]
+    ?:  &(=(ship.u.owner-access our.bol) (team:title our.bol src.bol))
+      ::  delete one of our.bol own paths
+      :_  %_  this
+            synced  (~(del by synced) path.act)
+            boned  (~(del by boned) [%permission path.act])
+          ::
+            access-control
+              (~(del by access-control) access-control.u.owner-access)
```
Fang-: The indentation here is off.

pkg/arvo/app/permission-hook.hoon
Fang-: (Why are we using a list over a set here anyway? What's the case where bones appear twice?)

pkg/arvo/app/chat-view.hoon
```diff
@@ -156,25 +156,40 @@
...
+            (group-poke [%add read.act group-read])
+            (group-poke [%add write.act group-write])
+            (chat-poke [%create our.bol path.act])
+            (chat-hook-poke [%add-owned pax security.act])
+        ==
+        (create-security [%chat pax] security.act)
+        :~  (permission-hook-poke [%add-owned group-read group-read])
+            (permission-hook-poke [%add-owned group-write group-read])
+        ==
+    ==
```
Fang-: Can't you make these a flat list in the code, instead of `zing`ing?

pkg/arvo/app/permission-hook.hoon
loganallenc: It doesn't nest, much to my dismay.

pkg/arvo/sur/permission-hook.hoon
loganallenc: ❤️

pkg/arvo/app/chat-view.hoon
loganallenc: No, because (create-security) returns a list.

pkg/arvo/app/permission-hook.hoon
loganallenc: What should it be?

pkg/arvo/app/permission-hook.hoon
loganallenc: Will do, thanks.

loganallenc: done.

pkg/arvo/app/permission-hook.hoon
loganallenc: @pilfer-pandex plz fix

pkg/arvo/app/permission-hook.hoon
Fang-: Reversed:

```
  access-control
(~(del by access-control) ...
```

Because when the hoon is under the "case", you're doing ...queenside? instead of kingside form. (Hope I got that right, but you get the idea.)

pkg/arvo/app/chat-view.hoon
Fang-: Oops, of course. If order around that doesn't matter too much you could do a `:*` with `create-security` at the end. If it does, then out of luck for this I guess.

pkg/arvo/app/permission-hook.hoon
loganallenc: Can you have multiple peers with the same wire on a single bone? I didn't know the answer to this question, so I made it a list instead of a set.

loganallenc: I avoid jars and jugs like the plague because they're not 'common' data structures.

pkg/arvo/app/permission-hook.hoon
loganallenc: Thanks! Fixed it and pushed.

pkg/arvo/app/chat-view.hoon
loganallenc: Order does matter, unfortunately.

pkg/arvo/app/permission-hook.hoon
Fang-: I believe bones are unique for all subscriptions on your app. @jtobin might be able to confirm?

pkg/arvo/sur/chat-view-imports.hoon
```diff
@@ -0,0 +1,4 @@
+/-  *permission-store, *permission-hook, *group-store, *permission-group-hook
+|%
+--
```
philipcmonk: Ha, never seen that pattern before.  Is this only used in one place?  If so, is it really a necessary bit of indirection?  Makes it a bit harder to see where stuff is coming from.  If you're worried about 80 cols, I think you can split `/-` across multiple lines.

pkg/arvo/sur/chat-view-imports.hoon
```diff
@@ -0,0 +1,4 @@
...
+
```
philipcmonk: Blank line

pkg/arvo/sur/chat-view-imports.hoon
ohAitch: Pfft, `|% --` definitely shouldn't be valid hoon - if just the `/-` won't work, maybe try `.`?

loganallenc: If you look at Ford's parser for /- runes, it does not support multiple lines. It also does not support multiple /- statements per file. In the interest of being below 80 characters per line, I'm going to keep this the same for now. Eventually, it would be nice for /- to support gaps with an end to the list as ==.

https://github.com/urbit/urbit/blob/2bb3c5a551d473fa94f555c1206f3fac88542ae9/pkg/arvo/sys/vane/ford.hoon#L893

pkg/arvo/sur/chat-view-imports.hoon
loganallenc: Thanks, getting rid of.

pkg/arvo/app/permission-hook.hoon
```diff
@@ -0,0 +1,282 @@
...
+  %-  zing
+  %+  turn  ~(tap in who)
+  |=  check-ship=ship
+  ^-  (list move)
+  %-  zing
+  %+  turn  ~(tap in access-paths)
+  |=  access-path=path
+  ^-  (list move)
+  =/  bne  (~(get by sup) [check-ship [%permission access-path]])
+  ?~(bne ~ [u.bne %quit ~]~)
```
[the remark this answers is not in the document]
loganallenc: Thanks, +murn is clearer.

pkg/arvo/sur/chat-view-imports.hoon
philipcmonk: I'm pretty sure `gaw` includes a newline.  This seemed to work in a generator:

```
/-  a,
    b
```

loganallenc: ah, great. trying that now.

loganallenc: that is VERY irregular syntax for hoon :(

loganallenc: okay, that does indeed work. removed the sur/chat-view-imports file and added the syntax you recommended.

loganallenc:
@jtobin @philipcmonk this should be ready to go, I think. Thanks for the reviews on the weekend.

PR #1820, opened 2019-10-14
build: urbit derivation debug metadata

brendanhay:
Rather than reconstructing the `urbitExe` path/argument string using a debug conditional in every ops expression, attaches the following information to the `urbit` derivation's `.meta` attribute:

* `urbit.meta.debug` - whether this derivation is a debug build.
* `urbit.meta.bin` - the binary path/name, such as `/nix/store/urbit/bin/urbit` or `/nix/store/urbit/bin/urbit-debug`.
* `urbit.meta.flags` - any flags this particular build should be invoked with, such as `-g`.
* `urbit.meta.exe` - the combination of `urbit.meta.bin` and `urbit.meta.flags`.

The `nix/pkgs/image` expression has also been moved into `nix/ops` to take advantage of these changes as it also makes use of this debug metadata, as well as an update to the `sh/image` build script.

### DIFF
--- a/nix/ops/brass/default.nix
+++ b/nix/ops/brass/default.nix
@@ -1,19 +1,11 @@
-{ pkgs, tlon, deps, pier, arvo, debug }:
-
-let
-
-  urbitExe = if debug
-             then "${tlon.urbit-debug}/bin/urbit-debug -g"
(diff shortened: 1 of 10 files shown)


jtobin:
👍 👍

PR #1825, opened 2019-10-15
chat: changed a =: to a =. to fix message storage bug

loganallenc:
Messages were not being stored properly in the chat-store. I did not realize that all subject modifications in =: were all in parallel. Changed to =. so that the subject modifications happen correctly (in sequence) to store messages.

### DIFF
--- a/pkg/arvo/app/chat-store.hoon
+++ b/pkg/arvo/app/chat-store.hoon
@@ -192,12 +192,11 @@
     =/  =hoon  (ream expression.letter)
     letter(output (eval bol hoon))
   =:  length.config.u.mailbox  +(length.config.u.mailbox)
-      number.envelope.act  length.config.u.mailbox
+      number.envelope.act  +(length.config.u.mailbox)
       envelopes.u.mailbox  (snoc envelopes.u.mailbox envelope.act)
-      inbox  (~(put by inbox) path.act u.mailbox)
   ==
-  :_  this(inbox inbox)
-  (send-diff path.act act)
+  :-  (send-diff path.act act)
+  this(inbox (~(put by inbox) path.act u.mailbox))
 ::
 ++  handle-read
   |=  act=chat-action
@@ -206,11 +205,9 @@
   =/  mailbox=(unit mailbox)  (~(get by inbox) path.act)
   ?~  mailbox
     [~ this]
-  =:  read.config.u.mailbox  length.config.u.mailbox
-      inbox  (~(put by inbox) path.act u.mailbox)
-  ==
-  :_  this(inbox inbox)
-  (send-diff path.act act)
+  =.  read.config.u.mailbox  length.config.u.mailbox
+  :-  (send-diff path.act act)
+  this(inbox (~(put by inbox) path.act u.mailbox))
 ::
 ++  update-subscribers
   |=  [pax=path act=chat-action]


loganallenc:
I checked all of the other stores, hooks, and views to see if they were making =: parallel modifications. None were, so this shouldn't crop up again.

joemfb:
@loganallenc, I should've made sure this was clear on review when I asked you to switch.

pkg/arvo/app/chat-store.hoon
```diff
@@ -194,8 +194,8 @@
   =:  length.config.u.mailbox  +(length.config.u.mailbox)
       number.envelope.act  length.config.u.mailbox
       envelopes.u.mailbox  (snoc envelopes.u.mailbox envelope.act)
-      inbox  (~(put by inbox) path.act u.mailbox)
   ==
+  =.  inbox  (~(put by inbox) path.act u.mailbox)
   :_  this(inbox inbox)
```
philipcmonk: Pretty sure the `(inbox inbox)` is redundant.

pkg/arvo/app/chat-store.hoon
```diff
@@ -206,9 +206,8 @@
   =/  mailbox=(unit mailbox)  (~(get by inbox) path.act)
   ?~  mailbox
     [~ this]
-  =:  read.config.u.mailbox  length.config.u.mailbox
-      inbox  (~(put by inbox) path.act u.mailbox)
-  ==
+  =.  read.config.u.mailbox  length.config.u.mailbox
+  =.  inbox  (~(put by inbox) path.act u.mailbox)
   :_  this(inbox inbox)
```
philipcmonk: Also here

pkg/arvo/app/chat-store.hoon
loganallenc: yeah, you're right. will make the change

pkg/arvo/app/chat-store.hoon
```diff
@@ -194,10 +194,9 @@
   =:  length.config.u.mailbox  +(length.config.u.mailbox)
       number.envelope.act  length.config.u.mailbox
       envelopes.u.mailbox  (snoc envelopes.u.mailbox envelope.act)
```
joemfb: To be clear, envelopes are zero-indexed, correct? (Ie, the number of the latest envelope in a mailbox is always one one less than the length of the mailbox.)

FWIW, I would be inclined to store the envelopes reversed, both for easy update (cells cons instead of `+snoc` append which traverses the whole list) and because you're more likely to read recent messages. I should've picked up on this in the original review. It's probably not worth changing OTA, but worth considering in the future.

loganallenc: No, it's 1-indexed. It turns out the change over to =: messed up quite a few things. I've now changed this to increment as well.

loganallenc: I don't know. Maybe. Seems reasonable.

joemfb: I'd argue that it should be 0-indexed (for consistency with our list gates if nothing else), but this is also not the time or place. Sorry for the insufficiency of my earlier review!

vvisigoth:
@jtobin can you make sure that this gets merged and OTA'd as soon as you're back up and at it?

PR #1844, opened 2019-10-17
updates vere memory management heuristics and debug info

joemfb:
This PR restores the old heuristic of purging vere caches every 1k events. We don't make sufficiently efficient use of free space to just depend on the memory pressure thresholds.

Additionally, this updates the `|mass` handler to also print the size of the free lists, and adds static memory measurement on startup, for cases where we can't evaluate `|mass` (currently run on every startup where < 1/2 of the loom is contiguously free).

### DIFF
--- a/pkg/urbit/worker/main.c
+++ b/pkg/urbit/worker/main.c
@@ -898,6 +898,21 @@ _worker_poke(void* vod_p, u3_noun mat)
   }
 }
 
+/* _worker_static_grab(): garbage collect, checking for profiling. RETAIN.
+*/
+static void
+_worker_static_grab(void)
+{
+  c3_assert( u3R == &(u3H->rod_u) );
+
+  fprintf(stderr, "work: measuring memory:\r\n");
+  u3a_print_memory(stderr, "total marked", u3m_mark(stderr));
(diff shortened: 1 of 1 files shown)


joemfb:
Thanks, @belisarius222.

I should mention that this is a continuation of the work from #1773. I shouldn't have removed the 1K cache threshold there, the only advantage is possible performance improvements but the possible disadvantages are substantial. I continue to want to improve our (or just establish any) cache management, but that needs to be done systematically, not incrementally. My apologies to all affected.

philipcmonk:
LGTM, I'll switch ~zod to this so it doesn't eat further into its contiguous free space.

philipcmonk:
This brought ~zod's memory usage down to 130MB, though it's got 970MB of free lists.  That should be stable.

joemfb:
I don't think we need to be worried about those free lists. But if we are, we can replay with these changes and they'll be smaller as a result.

joemfb:
@jtobin, we'll want to release this (and anything else intervening) as `v0.9.1`.

philipcmonk:
yes, and we should emphasize that people should upgrade their vere so they don't run out of memory.

PR #1845, opened 2019-10-18
Hook up command line flags

eglaysher:
This builds a HasPierConfig typeclass for reading, and threads it through most of Vere. Then I used this to implement the following command line flag options:

- `-O`: offline only; disables all networking

- `-L`: localhost only networking; only binds and accepts packets from localhost

- `-x`: I implemented the big todo in Main.hs so that we now run a booted ship unless we are passed the `-x` option, which exits immediately. I also implemented/tested that this works with normal running.

- `-N`: dry run mode now works where we disable ames messages and writing to the event log. I tested it by booting a comet, shutting the comet down, starting it up in -N, writing a command to the dojo, restarting the comet again in normal mode, and ensuring that the previous command wasn't in up arrow shell history.


eglaysher:
Are you sure? The urbit help file says that `-d` is daemon mode. It looks like '-D' claims to be "dry" in the code comments, but is documented in the help as "recompute from events," which is not what I'd expect dry-run to do.

I can hook up king's recompute state from all events to a flag if you want, but several of these things should probably be hooked up to `--long-style-flags` instead of single letter unix ones.

joemfb:
You're right, my bad -- `-D` is dry-run, and, as usual, the usage strings are actively misleading.

We should have a replay mode/flag (probably gated by a prompt/warning that it may take a *long* time), but combining that with dry-run (which can't update state) is only helpful for runtime development, so it probably doesn't need a flag.

eglaysher:
What this update does:

- Speeds up king's running of urbit-worker by more than an order of magnitude by always disabling hashboard like vere does.

- Implements `--full-log-replay` which deletes the current snapshot and replays the event log from the beginning.

- Moves from `-N` to the long option `--dry-run` as discussed in person.

philipcmonk:
Yay for long options!

eglaysher:
Ping @benjamin-tlon

benjamin-tlon:
LGTM

Issue #1540, opened 2019-08-09
Some crypto functions produce results in reverse byte order


Fang-:
Reference result for SHA hashes of the empty, zero-length string:

```
SHA-1:   0xda39a3ee5e6b4b0d3255bfef95601890afd80709
SHA-256: 0xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
```

The output our stdlib functions give:

```
> `@ux`(shan 0)  ::  SHA-1
0xda39.a3ee.5e6b.4b0d.3255.bfef.9560.1890.afd8.0709
> `@ux`(shax 0)  ::  SHA-256
0x55b8.5278.1b99.95a4.4c93.9b64.e441.ae27.24b9.6f99.c8f4.fb9a.141c.fc98.42c4.b0e3
```

`shan` is returning the correct result.  
`shax` is returning the correct result... in reverse byte order. To match reference results, the result has to be put through `(swp 3 res)`. (Or, to make sure cases with leading zero-bytes are handled correctly, `(lsh 3 (sub 32 (met 3 res)) (swp 3 dat))`, where `32` is the expected byte-length of the result.)

This inconsistency is present in many different crypto functions. I have verified this to be the case for the SHA-2 functions (`shax`, `shaz`, etc), HMAC functions (`hmc:scr:crypto`) PBKDF2-HMAC (`pbk:scr:crypto`), and of course all functions that directly return a result generated by these.

Of course, reverse byte order would be correct if we were expecting `@t` results, but we aren't. All the crypto functions take and produce `@`, so one would expect their results for given hex literal inputs to be equivalent to those defined in official test vectors.  
If that means developers (or a wrapper arm) need to do a `swp` on their `@t` input, so be it.

For the BIP32 work I'm doing currently, I'll depend on wrappers around the problematic functions to make sure we're always dealing with expected results.  
A longer term solution is to *actually* fix this in the offending functions themselves (and their jets), which also requires updating all code that depends on them. We could get away with throwing `swp`s all over the place as described above, but that might not be optimal.

PR #1853, opened 2019-10-18
Add unit tests for SHA encryption and related arms

ken-meier:
This tests each arm referenced in [https://urbit.org/docs/reference/library/3d/](url), based on official NIST test vectors wherever possible. A few notes:

Many failures at the moment because of the issue noted in #1540. shal and shay both take reversed byte order input and produce reversed byte order output (presumably due to cords being little endian). So, all sha256/512 arms fail against official test vectors. Tests with cord input produce expected output but in reverse byte order. Tests with hex input fail entirely.  Tests could be modified to accommodate this (reverse byte order of hex input and all expected output), but I've left them as written for now as I'm not sure how intentional this is, since shan(SHA1) works differently. shan takes little endian input and produces big endian output. This means it passes tests with cord input, but fails hex input test vectors. One way or the other I imagine we'd want some consistency here.

Randomness functions and sham have simple tests based on the documentation at the moment. If shal/shay change their arbitrary outputs will probably also change.


joemfb:
Thanks for these!

IMO, #1540 is not phrased correctly -- all these functions *should* take little-endian arguments and produce little-endian values. It's possible we can just remove `+shan` (it is deprecated, after all, and appears to be unused). If we need a SHA1 implementation for some integration, it could be reintroduced (with little-endian output) as a userspace library.

I'm curious if others feel otherwise.

Fang-:
Fair that the phrasing isn't as good as it can be. At the time, [big-endian wrapper functions](https://github.com/urbit/urbit/blob/c6b4317aa902a0327bca4e2bed132ed2813d6f9f/pkg/arvo/sys/hoon.hoon#L3585) (including an [alternative sha1 implementation](https://github.com/urbit/urbit/blob/c6b4317aa902a0327bca4e2bed132ed2813d6f9f/pkg/arvo/sys/hoon.hoon#L3606)) were added to the stdlib to account for leading zero bytes, and behave identical to the way all crypto specs I've worked with describe their intended behavior. (ie, passing in `0x1234` treats it as the value `0x1234` rather than `0x3412`). ([`+keccak`](https://github.com/urbit/urbit/blob/c6b4317aa902a0327bca4e2bed132ed2813d6f9f/pkg/arvo/sys/zuse.hoon#L3914-L3917) and [`+hmac`](https://github.com/urbit/urbit/blob/c6b4317aa902a0327bca4e2bed132ed2813d6f9f/pkg/arvo/sys/zuse.hoon#L4221-L4224) work primarily with big-endian as well.)

joemfb:
My criticism was also poorly phrased! And I had forgotten about your wrapper functions. (BTW, I think it'd be more accurate to say that you want `0x1234` to be treated as the byte-vector `[0x34 0x12 ~]`. Your example would be 4-bit, not 8-bit, and characterizing that as *the value* assumes an answer to the question under dispute ...)

@ken-meier, should update these tests to point to the big-endian SHA-* wrappers in the `|sha` core of `hoon.hoon`.

FWIW, I think that our defaults for atom-gates should be little-endian arguments and products, with big-endian wrappers where they're needed for interop. And we should remove or replace anything that's inconsistent in its endianness. But all of that's out of scope for this PR.

ken-meier:
Retargeted tests against the `sha` wrappers and moved into hashes.hoon as requested.

Added tests for sha1 with length as that's now a thing.
Tests with length/hex arguments now all pass.
Tests with cord inputs fail as the wrapper functions flip the already little-endian data. Might need to test if arguments are `@t` or not?
Derivative functions such as `shad` fail because there are no wrappers for them, though presumably they could be implemented fairly easily.

ohAitch:
We do have contiguous-string-split-across-line syntax! See `++po` https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/hoon.hoon#L4043-L4078

joemfb:
Woah, I had totally failed to internalize that!

In that case, I recommend something like:

```
> =/  a  'abc\
         /def'
  [(met 3 a) `@t`(swp 3 a)]
[6 'fedcba']
```

ken-meier:
Cleaned the tests up and added a few containers for the remaining functions modeled after the existing ones in hoon.hoon. Feel free to include or exclude those - mainly wanted to ensure that all the tests actually pass.

ken-meier:
Cut the changes to hoon.hoon and tests that relied on them, to be made into a separate PR.

jtobin:
@ken-meier Cheers, I've squashed and merged these commits (manually, via the command line), so will close this PR.  

You might want to have a quick look over our [git hygiene notes](https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md#hygiene) before making the next bunch of changes you have planned.

ken-meier:
Thanks, will do. Mind you, I imagine a lot of the hygiene requires having a
clue what you're doing with git, which I don't. :P  Gettin' there, though.

jtobin:
On Thu, Nov 07, 2019 at 05:46:08AM -0800, ken-meier wrote:
> Thanks, will do. Mind you, I imagine a lot of the hygiene requires
> having a clue what you're doing with git, which I don't. :P  Gettin'
> there, though.

No worries.  Happy to help out here as well.

Issue #1863, opened 2019-10-21
publish: resubscribe on subscription close


philipcmonk:
When a subscription closes with `%quit`, we should automatically resubscribe unless we expected it to close.  In #1859 the subscriptions were closed because there were more than 20 updates to send at once, but they didn't resubscribe, so from the publisher's perspective all the subscribers were gone and from the subscriber's perspective all the old posts were still there, but no new ones show up.

The usual principle is "resubscribe on %quit, cancel on failed %reap"

@ixv

Issue #1834, opened 2019-10-16
Publish %serve seems to have become no-op


Fang-:
I had originally imported my old notebook by (copying in the files, then) using the `%serve` action. The posts showed up fine, but I hit #1801.

In an attempt to solve that, I deleted the notebook, and followed a flow proposed by ~winter that would supposedly not hit #1801.

1. Create notebook using the web UI
2. Unserve the notebook
3. Copy in files, commit
4. Serve the notebook

However, it wasn't picking up on the copied-in posts. The notebook remained empty in the UI. Even retrying without steps 1-2 didn't work. It feels like `%serve`, for this notebook, has turned into a no-op.

Trying similar flows with notebooks with different names also fails. From looking at `%serve`, I'd _guess_ it's waiting on ford responses that will never come, but I don't actually know.

@ixv please advise.

ixv:
Just tried that exact procedure on a fakezod and it worked just fine. Would you be willing to send me the files that you used so I can try it with them?

Fang-:
Sent!

PR #1865, opened 2019-10-22
Publish: 2 seperate fixes

ixv:
Fix one is for https://github.com/urbit/urbit/issues/1834
There previously was no logic set the permissions correctly in the %serve and +import cases, resulting in people not being able to subscribe to blogs that have been imported in this manner.

Fix two is for https://github.com/urbit/urbit/issues/1863
The old gall 20 limit ~~bug~~ feature. We now automatically resubscribe when we receive a quit. And in the cases of the app intentionally sending a quit, we have converted that action to a poke, asking the other end nicely to unsubscribe from us.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -1244,6 +1244,8 @@
   ::
       %serve
     :: XX specialize this check for subfiles
+    ?.  =(our.bol src.bol)
+      [~ this]
     ?:  (~(has by pubs.sat) coll.act)
       [~ this]
     =/  files=(list path)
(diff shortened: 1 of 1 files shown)


philipcmonk:
For "asking them nicely to unsubscribe", this seems to only happen in peer handling, another option is to simply crash so that they get a negative `%reap`.  Then you can handle that `%reap` to give feedback that the subscription failed.  This is how nu chat does it: https://github.com/urbit/urbit/blob/3210acd4ded48306a219f60e584a92e8f98a971a/pkg/arvo/app/chat-hook.hoon#L136

and

https://github.com/urbit/urbit/blob/3210acd4ded48306a219f60e584a92e8f98a971a/pkg/arvo/app/chat-hook.hoon#L255-L267

ixv:
Ah thanks, I didn't realize that would work. Will do.

pkg/arvo/app/publish.hoon
```diff
@@ -1390,6 +1420,13 @@
   ::
   ==
 ::
+++  quit-collection
+  |=  wir=wire
+  ^-  (quip move _this)
+  =/  pax=path  (weld /collection wir)
+  :_  this
+  [ost.bol %peer pax [src.bol %publish] pax]~
```
Fang-: If your in-app convention is to always use a `wire` identical to the subscription `path`, you could get away with implementing a more generic `++quit` and not thinking about it in the future. But, right now the only subscriptions you do are `/collection/*` ones, so it doesn't matter much.

You could do `[%collection wir]` instead of `weld`ing. But that, too, doesn't matter very much.

ixv:
@Fang-: Yes, doing %unserve and then %serve again will work.

@jtobin: Done

Issue #1829, opened 2019-10-16
Console prompt doesn't use terminal's full width


Fang-:
Observe:

![cli](https://user-images.githubusercontent.com/3829764/66907855-518c5000-f00a-11e9-9859-3de5d4b3fb0a.gif)

Expected behavior is for this to continue adding input on the right-hand side until reaching the edge of the terminal window, and only then starting to scroll earlier input out of view.

Resizing the terminal while Urbit's running has given me similar issues in the past, but that isn't the case here: the process was started in a terminal of this same size.

Note that this happens for both dojo and chat-cli.

philipcmonk:
I saw this the other day as well.

joemfb:
Resizing the console also doesn't fix it. I've seen this before with ordering issues on boot, but I've never seen it not reset on resize. A quick check with |verb shows the %blew event being passed to %gall (presumably to drum in :hood, as it should be). Further debugging is needed.

joemfb:
The terminal measurement was incorrectly gated when `-t` (disable tty) was added. I missed it on review. Fix incoming.

jtobin:
Fixed by the linked PR.

PR #1867, opened 2019-10-22
vere: restore terminal dimensions measurement

joemfb:
The condition, added in #1793 (as part of `-t`, which disables features that require a tty), was precisely backwards. I missed it on review.


joemfb:
fixes #1829

Issue #1352, opened 2019-07-17
=type syntax shadows all the time


philipcmonk:
`=type` is super convenient, but about 20% of the time the shadowing is super frustrating.  This is obviously bad design.  The old solution to this was to give a unique canonical name for each type (eg `pax` for `path`), but that requires twice the memorization for no benefit.

Many languages keep types and variables in separate namespaces.  This lets you quit having to worry so much about names because there's a canonical answer to "what should a variable of type `t` be called?".  One common way to do this is say that all types start with a capital letter.  Then a variable of type `State` is canonically called `state`.  This is very reasonable, and solves the problem elegantly.

It's against the hoon aesthetic to use capital letters, so I propose suffixing `-` to the type name.  Thus, a variable of type `state` becomes `state-`.  We don't ever use suffixed `-`, so this is easily read as "the variable of type `state`.  I propose using this both at define time and use time.  This solves the problem of "I see variable `a-point` but can't find the definition because it's `=a=point`.

This doesn't syntactically collide with the old way, so we could add this without removing the prefix `=` and see which wins out in practice.

This turns:

```
++  new-event
  |=  [=ship =udiff:point]
  ^+  this-su
  =/  =point  (~(gut by pos.zim.pki) ship *point)
  =/  =diff   (udiff-to-diff:^point udiff point-)
  (public-keys:feel %diff ship diff)
```

into:

```
++  new-event
  |=  [ship- udiff:point-]
  ^+  this-su
  =/  point-  (~(gut by pos.zim.pki) ship- *point)
  =/  diff-   (udiff-to-diff:point udiff- point-)
  (public-keys:feel %diff ship- diff-)
```

I don't actually know how to implement the postfix parsing without running into memory issues.  Prefix is much easier, and would be more in line with the rest of our syntax.  I don't care, as long as we stop shadowing.

ohAitch:
This puts a finger pretty well on my unease with shadow syntax!

For parsing, we probably just want to make a @tas parser that accepts the
trailing - and filters it out in this case(/ rejects otherwise): surely we
want udiff-:point and not udiff:point-

ohAitch:
(To be specific: rejects it in arm names, accepts it in references, converts to foo-=foo in tile syntax)

My one concern would be two-level usage like `mold-`, but that seems easy enough to avoid via renaming / not using the abbreviated face form, if that pops up

baudtack:
[...]

PR #1868, opened 2019-10-22
gall: fix bug in +ap-peek

jtobin:
fc7901d2 refactored much of +ap-peek, but introduced a bug in the process.  The relevant diff from that commit is as follows:

```
         =/  slammed
-        =/  =path  [ren tyl]
-        =/  =vase  !>((slag p.u.cug path))
-        (ap-slam q.u.cug p.arm vase)
+        =/  index  p.u.maybe-arm
+        =/  term  q.u.maybe-arm
+        =/  =vase
+          =/  =path  [term tyl]
+          =/  raw  (slag index path)
+          !>  raw
+        (ap-slam term p.arm vase)
```
Note that [ren tyl] was replaced with [term tyl], where 'term' and 'ren' are not equal.  This commit merely rights that wrong.

(/cc @philipcmonk as this is probably still relevant on your static gall branch)

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:8409523d9ada2b962b1f9ac4c996ff957f27c273d373e6ca50197e6abb66bacd
-size 9063441
+oid sha256:78a42e19573a5726c1045efe31ecdb6821afe7d5cf2ba7b34a06a3f56980c9d2
+size 9071562
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1287,11 +1287,11 @@
       ::
       =/  slammed
         =/  index  p.u.maybe-arm
-        =/  term  q.u.maybe-arm
+        =/  name  q.u.maybe-arm
         =/  =vase
           =/  =path  [term tyl]
           !>  (slag index path)
-        (ap-slam term p.arm vase)
+        (ap-slam name p.arm vase)
       ::
       =^  possibly-vase  ap-core  slammed
       ?:  ?=(%.n -.possibly-vase)


loganallenc:
Hi Jared, I’m curious as to how the bug was being surfaced. Also, have you tested that the chat app still works? It utilizes +peek extensively.

jtobin:
From @joemfb in an email re: the DNS collector app:

> the read interface (scry/+peek) to :dns-collector is not working. It
> worked last era, so maybe some aspect of app-scry changed in the
> recent %gall refactor?

I took a peek (ha ha, get it) et voilà, there was indeed a shadowing
issue, so *presumably* this caused the regression in :dns-collector.

I hadn't checked the effect of this fix on anything downstream, but I
did boot up Landscape on a fakezod after seeing your comment and clicked
around / did some stuff in chat.  Nothing seemed awry.

It's difficult for me to tell what the effect of the shadowing bug is,
exactly.  If you take a look at the code, it seems like it may produce
some unwanted effect whenever the index is zero -- otherwise the
shadowed term will just be lopped off by `slag`.

joemfb:
Worth noting that this has also broken `:acme`.

belisarius222:
I think new Gall might overuse this kind of shadowing, because it makes this kind of bug a lot more likely.  I was actually worrying about a very similar shadowing in some other Gall arm last night; I think that one is ok, but it is a bit of a footgun.

I wonder if the criterion could be: if you're shadowing a variable with a variable of a different type, it's ok, but if they're the same type, you're opening yourself up to errors.  Of course, that has the problem that if you change the types later and they happen to match up, you could accidentally get back into the bad situation.  There might be a better way of thinking about this.

Maybe we should just try not to shadow variables in general.

joemfb:
I agree that this shadowing feature is being overused. It seems nice in small arms or gates, where the meaning of some value is unambiguously conveyed by the name of its type. I also think a/b/c or p/q/r are fine in similar situations. The shadowing style is worse than ultra-lapidary when it creates a false sense of clarity; and that's easily enough done when you have a couple `+term`'s, for instance.

ohAitch:
Perhaps it would be good to limit use to |= but not =/ then?

philipcmonk:
Well, the usual anti-shadowing measure is =foo=bar, which doesn't work in =/, where it would be most useful.

I think the rule violated here is that "term" is too general for this usage.  Some general principles:

- `=foo` is good if the scope is small, especially less than five lines.
- `=foo` is good if it's specific, like `=care`.  There's likely only one +care you're talking about.  If there's more than one, both should have a more specific name.
- `=foo` is bad if it's too general of a type.  `=term` is rarely a good idea unless you actually only care about it being a term (eg in a few stdlib operations).  `=path` is occasionally useful, but never if you actually mean something more specific like `=wire`.
- If you find yourself using `=general-type`, consider creating another type as a documentary aid.  If a noun, consider adding an aura.  For example: `=term` is bad, `app-name=term` is good, `++  app-name  @tasappname` followed by `=app-name` is best.  `=@app-name` is for those who like to live dangerously.

ohAitch:
Ah did that ever make it in? I thought foo-bar names popping up without
ever being textually declared as such had been deemed too hard to follow

joemfb:
Previously discussed in #1352

PR #1869, opened 2019-10-22
contributing: fix typos in git blame and git bisect [ci skip]

pkova:


jtobin:
I contest that these are typos! :smile:

I always refer to git subcommands in this fashion; this is actually the way they appear in the git man pages.  For example:

```
       Please see git-commit(1) for alternative ways to add content to a
       commit.
```

or

```
SEE ALSO
       git-status(1) git-rm(1) git-reset(1) git-mv(1) git-commit(1) git-
       update-index(1)
```

Arguably they should actually be suffixed with the man pages section, e.g. `git-blame(1)` and `git-bisect(1)`, in our contributing doc.  I think I avoided doing this so as not to seem overly pedantic, but I'd be keen on it.. 😝

pkova:
Well fuck you're right! Never mind.

PR #1877, opened 2019-10-23
Dr. Tobin is prescribing you more pills

jtobin:
This PR makes a number of changes around pill management.  See the individual commit descriptions for more details, but the TLDR is:

* has `/sh/release` spit out a brass pill artifact as `urbit-vx.y.z.pill`
* changes the pill format in the vere bootstrap URI from `urbit-$URBIT_VERSION.pill` to `urbit-v$URBIT_VERSION.pill`
* adds a Nix derivation for Arvo specialised to Ropsten (i.e. a testnet), as well as derivations for brass & ivory testnet pills
* has `sh/release` spit out testnet pills as well

This automates a lot of mundane tasks that are currently required when one sets up a testnet.

### DIFF
--- a/nix/pkgs/arvo-ropsten/builder.sh
+++ b/nix/pkgs/arvo-ropsten/builder.sh
@@ -1,0 +1,26 @@
+source $stdenv/setup
+
+cp -r $src tmp
+chmod -R u+w tmp
+
+ZUSE=tmp/sys/zuse.hoon
+AMES=tmp/sys/vane/ames.hoon
+ACME=tmp/app/acme.hoon
+
+# replace the mainnet azimuth contract with the ropsten contract
+sed --in-place \
+  's/\(\+\+  contracts  \)mainnet\-contracts/\1ropsten-contracts/' \
+  $ZUSE
+
+# increment the %ames protocol version
+sed -r --in-place \
+  's/^(=\+  protocol\-version=)([0-9])/echo "\1$(echo "(\2+1) % 8" | bc)"/e' \
+  $AMES
+
+# use the staging API in :acme
+sed --in-place \
+  's_https://acme-v02.api.letsencrypt.org/directory_https://acme-staging-v02.api.letsencrypt.org/directory_' \
+  $ACME
+
+cp -r tmp $out
+chmod -R u+w $out
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -298,7 +298,7 @@ _main_getopt(c3_i argc, c3_c** argv)
            && u3_Host.ops_u.url_c == 0
            && u3_Host.ops_u.git == c3n ) {
     u3_Host.ops_u.url_c =
-      "https://bootstrap.urbit.org/urbit-" URBIT_VERSION ".pill";
+      "https://bootstrap.urbit.org/urbit-v" URBIT_VERSION ".pill";
   }
   else if ( u3_Host.ops_u.nuu == c3y
            && u3_Host.ops_u.url_c == 0
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: build build-all install cross release test pills clean
+.PHONY: build build-all install cross release test pills ropsten-pills clean
 
 build:
 	nix-build -A urbit -A herb --no-out-link
@@ -22,6 +22,9 @@ pills:
 	sh/update-solid-pill
 	sh/update-brass-pill
 	sh/update-ivory-pill
+
+ropsten-pills:
+	sh/create-ropsten-pills
 
 interface:
 	sh/build-interface
--- a/nix/ops/default.nix
+++ b/nix/ops/default.nix
@@ -2,10 +2,11 @@
 
 let
 
-  tlon  = import ../pkgs { inherit pkgs; };
-  arvo  = tlon.arvo;
-  herb  = tlon.herb;
-  urbit = if debug then tlon.urbit-debug else tlon.urbit;
+  tlon         = import ../pkgs { inherit pkgs; };
+  arvo         = tlon.arvo;
+  arvo-ropsten = tlon.arvo-ropsten;
+  herb         = tlon.herb;
+  urbit        = if debug then tlon.urbit-debug else tlon.urbit;
 
   bootbrass = ../../bin/brass.pill;
   bootsolid = ../../bin/solid.pill;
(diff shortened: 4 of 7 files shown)


sh/release
```diff
@@ -14,10 +14,22 @@ fi
...
 for plat in linux64 darwin
 do
   sh/cross urbit "$plat"
 
   echo "packaging release/urbit-$plat-$ver.tgz"
   (cd cross/$plat; tar cz .) > release/urbit-$plat-$ver.tgz
 done
+
+# release pill
+cp bin/brass.pill release/urbit-v$ver.pill
```
joemfb: You may want `nix-build brass` here instead -- the brass pill in `bin/` is not updated on any kind of regular basis.

ohAitch:
How redundant are brass and solid pills? Should this only run on commits
touching sys / be cached so as to be built from the most recent commit
touching sys? (Are solid pills even still built in CI?)

joemfb:
Currently, no pills are built in CI. If this PR is merged as is, brass and ivory pills with testnet constants (notably, the azimuth contract on Ropsten) would be built in CI.

nix/pkgs/arvo-ropsten/builder.sh
```diff
@@ -0,0 +1,26 @@
...
+ACME=tmp/app/acme.hoon
+
+# replace the mainnet azimuth contract with the ropsten contract
+sed --in-place \
+  's/\(\+\+  contracts  \)mainnet\-contracts/\1ropsten-contracts/' \
+  $ZUSE
+
+# increment the %ames protocol version
+sed -r --in-place \
+  's/^(=\+  protocol\-version=)([0-9])/echo "\1$(echo "\2+1" | bc)"/e' \
```
joemfb: This would rotate the ames version on every commit, right? I'm not sure we want to do that; the relevant conditions for a testnet is that a) the version doesn't match the livenet and b) the version doesn't match other testnets. If we want to automatically provisions networks from a given commit, I think we'll want to handle network isolation at the infrastructure level.

joemfb:
Solid pills are faster to boot, so they're used more frequently in development. Brass is the correct bootstrap sequence.

jtobin:
> We run sh/release on ever push to urbit/urbit, or to any open PR there.

Ah.  I hadn't realised this was being run on every push, but you're right.  It's certainly not necessary to build these pills for every commit -- I'm only interested in doing so for tags.

Is anyone interested in building & preserving the release binaries for every push?

> You may want nix-build brass here instead -- the brass pill in bin/ is not updated on any kind of regular basis.

Similarly I'd expected this was only being run for tags, when "all pills are up to date" is a safer assumption.

nix/pkgs/arvo-ropsten/builder.sh
jtobin: Yeah the `arvo-ropsten` derivation is the same as `arvo`, except having a testnet configuration (with the protocol number incremented by one).

I'm interested in producing and maintaining a long-lived testnet, so "mainnet + 1" feels appropriate as a canonical protocol number for that.  For ephemeral/short-lived testnets we could simply then avoid both the mainnet and canonical testnet protocol numbers, managing that elsewhere as you recommend.

(This is a bit out-of-bounds -- I haven't made a proposal for a canonical testnet, so an ad-hoc UP might be useful after the fact.  Also, this will obviously need to be patched to avoid incrementing out of bounds, but it would do the trick for the time being.)

ohAitch:
+4 maybe? (Or well, xor 4)

+ 1 sounds like it just sets you up for remnants of the previous testnet
accidentally sending packets to an era.

jtobin:
> +4 maybe? (Or well, xor 4)

Yeah, I'd thought (something like) this as well but filed it under "implement later."  I suppose I may as well do it now, though.. 😄

nix/pkgs/arvo-ropsten/builder.sh
jtobin: (It may be better and more Nix-appropriate to apply a proper patch here as well, rather than hack at the source with sed.)

jtobin:
I've since removed the functionality I'd previously added to `sh/release`.  Now this just includes:

* the same original change to the vere bootstrap pill URI
* the same additional Nix derivations, except `arvo-ropsten` changes the %ames protocol number with appropriate wraparound
* an additional script, `sh/create-ropsten-pills`, plus a corresponding `ropsten-pills` target in the Makefile

This is the functionality I'm most interested in at the moment anyway -- I'll think about changing `sh/release` elsewhere.

pkg/urbit/daemon/main.c
```diff
@@ -298,7 +298,7 @@ _main_getopt(c3_i argc, c3_c** argv)
            && u3_Host.ops_u.url_c == 0
            && u3_Host.ops_u.git == c3n ) {
     u3_Host.ops_u.url_c =
-      "https://bootstrap.urbit.org/urbit-" URBIT_VERSION ".pill";
+      "https://bootstrap.urbit.org/urbit-v" URBIT_VERSION ".pill";
```
joemfb: This naming-scheme change remains planned?

jtobin:
On Thu, Oct 24, 2019 at 11:32:26AM -0700, Joe Bryan wrote:
> A question about pill renaming:
>
> > @@ -298,7 +298,7 @@ _main_getopt(c3_i argc, c3_c** argv)
>             && u3_Host.ops_u.url_c == 0
>             && u3_Host.ops_u.git == c3n ) {
>      u3_Host.ops_u.url_c =
> -      "https://bootstrap.urbit.org/urbit-" URBIT_VERSION ".pill";
> +      "https://bootstrap.urbit.org/urbit-v" URBIT_VERSION ".pill";
>
> This naming-scheme change remains planned?

That's right.  The actual git release tags are v0.9.0, v0.9.1, etc., so
having the 'v' prefix stripped from the bootstrap pills seems imperfect.
This aligns them with the way the release binary tarballs are named,
e.g. urbit-linux64-v0.9.1.tgz.

pkg/urbit/daemon/main.c
joemfb: Fine by me, we'll just need to take this into account if we push another hotfix this era.

joemfb: @jtobin, reminder that the pill needs to be renamed to account for this scheme.

jtobin:
On Wed, Oct 30, 2019 at 02:29:30PM -0700, Joe Bryan wrote:
> @jtobin, reminder that the pill needs to be renamed to account for
> this scheme.

Merci.  Haven't forgotten -- just been planning to wait until the v0.9.2
release.  I was initially going to push that as a hotfix, but I may just
release it on Monday along with the weekly Arvo update.

PR #1882, opened 2019-10-24
jael: provide edge-triggered breach notification

### LINKED ISSUE #1852: Ames doesn't recognize breach if it comes in a `%full` update
Jael updates can be `%full` or `%diff`, and Ames only registers a breach if it comes from a `%diff`.  One way we can get a full is if you run `:azimuth-tracker|listen ~ %ship ~zod` or similar.  We should track the most recent rift, and if we get a `%full` or `%diff` that takes us to a higher rift, we should consider that a breach, just like in clay and gall.  Arguably Jael should have a synthetic subscription path for just that purpose.

philipcmonk:
Until now, clients of Jael have had to store the first-seen rift if they
want to reliably detect breaches.  Otherwise, they would get a false
positive if they heard an old message about a breach (eg if you kick
azimuth-tracker).  Clay and Gall did this correctly, but Ames did not.

Jael already maintains this state, so I added a notification to the
existing subscription that happens whenever it notices a breach (a diff
or full where the new rift is greater than the old one).

Because this is an issue on the live network, I wrote state adapters
for Gall and Clay.  The Gall one just removes the rift from our state,
but the Clay one is much more involved because we have to upgrade
instances of the clad monad that are possibly in progress.
Specifically, since more input is possible than before, we must wrap any
in-progress instances of the monad in a function that handles the
potential new input from Jael.  This temporarily preservers a copy of
the old kernel, but only until the current commit/merge/update has
completed.

The real solution for Clay is to factor out those IO-heavy instances to
userspace tapp/async/imp/threads, and if an upgrade happens in the
middle, you should simply restart them.

Fixes #1852

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -299,8 +299,7 @@
 ::  Foreign desk data.
 ::
 ++  rung
-          $:  rit=rift                                  ::  rift of 1st contact
-              rus=(map desk rede)                       ::  neighbor desks
+          $:  rus=(map desk rede)                       ::  neighbor desks
           ==
 ::
 ::  Hash of a commit, for lookup in the object store (hut.ran)
@@ -533,12 +532,6 @@
       $:  @tas                                          ::  by any
           $>(%crud vane-task)                           ::  XX strange
   ==  ==                                                ::
---
-::
-::  Old state types for ++load
-::
-=>  |%
-+$  raft-1  raft
 --  =>
 ::  %utilities
 ::
@@ -2663,8 +2656,6 @@
       ::  save foreign +rede
       ::
       =/  run  (~(gut by hoy.ruf) her *rung)
-      =?  rit.run  =(0 rit.run)
-        (fall (rift-scry her) *rift)
       =/  rug  (~(put by rus.run) syd red)
       ruf(hoy (~(put by hoy.ruf) her run(rus rug)))
     ::  save domestic +room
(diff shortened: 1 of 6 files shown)


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4374,16 +4352,209 @@
...
+        ==
+      ::
+      ++  upgrade-clad-1
+        |*  a=mold
+        |=  e-form=(eval-form-1 a)
+        :-  effects.e-form
+        ^+  *form:(clad ,a)
+        |=  cin=clad-input
+        ?:  ?=([%j %public-keys %breach *] sign.cin)
+          [~ ~ %wait ~]
```
belisarius222: I expected a %fail here, not a %wait.  It looks like this will just ignore a breach notification.  Do we not need to kill transactions involving a foreign ship that breaches while the transaction is still ongoing?  Or is Clay's input from a foreign ship always atomic for a given transaction?

jtobin:
On Wed, Oct 23, 2019 at 09:37:53PM -0700, Philip Monk wrote:
> Because this is an issue on the live network, I wrote state adapters
> for Gall and Clay.

It's really great to see these.  I don't think I've seen any state
adapters written yet -- I never got around to writing that :hood adapter
in the last era, myself.

In the near future it seems like we may want to write vane adapters
using state_{K}, state_{K-1}, etc. for some Arvo kelvin K?  That can be
refined further if/when we later want to version the individual vanes
with kelvins.

pkg/arvo/sys/vane/clay.hoon
philipcmonk: There's no real way it should get a public-keys response, since unlike in tapps, all the moves are on a particular wire prefix and we only route responses on that wire.

The transactions do need to be killed, but that happens elsewhere, where we actually handle the breach notification.

philipcmonk:
> In the near future it seems like we may want to write vane adapters
> using state_{K}, state_{K-1}, etc. for some Arvo kelvin K?

That would be reasonable once we're in the habit of using Kelvin versions for this sort of thing.

Issue #1871, opened 2019-10-22
Publish data confusion


ngzax:
Recently I was editing documents and changing *only the body* of a single post and it changed the title of it and another post! 

I am not 100% sure, but I think it is related to using "Previous" and "Next" to navigate between the posts and then clicking "Edit" I think it is getting crossed up somehow. This is because if I always navigate back to the top this issue doesn't seem to happen.

mattlevan:
This happened to me as well!

PR #1886, opened 2019-10-25
publish: correctly set title and body when navigating with next/prev

ixv:
Addresses this issue: https://github.com/urbit/urbit/issues/1871


ixv:
Yes, it needs a refactor. Much of the complexity comes from needing to handle confusing edge cases around permissions, which is very dependent on the backend architecture, so probably the time to do this will be after the backend refactor.

PR #1887, opened 2019-10-25
eyre: fix wire=path terminology

ohAitch:


ixv:
Can you explain the difference here? I've heard people allude to the different meaning of these words but I've never actually heard it explained.

philipcmonk:
A wire is structurally a path, but it's more specific. It's exactly the
thing you add on a %pass move and the thing you receive back when you get a
response. Thus, a duct is a stack of wires.

jtobin:
> A wire is structurally a path, but it's more specific

This is IMO one of the most unfortunate qualities about present-era Hoon -- that there is no convenient mechanism for distinguishing fundamentally different things, like paths and wires, at the type level.  The result is that e.g. paths and wires are not *actually* different things, but that any difference ascribed to them is simply a matter of convention.

The following would be a cardinal sin in Haskell:

```
type Dollar  = Double
type Celsius = Double
type Hertz   = Double
```

It would be a cardinal sin because the type synonyms don't provide any information to the compiler, at all -- they just serve as mnemonics for the programmer.  The result is that you can pull nonsense like this:

```
calibrateMyThermostat :: Celsius -> Celsius
calibrateMyThermostat temperature = someAdjustment * temperature

calculateMyTaxes :: Dollar -> Dollar
calculateMyTaxes income = someRate * income

whatIsThisIDontEven :: Dollar -> Hertz
whatIsThisIDontEven = calculateMyTaxes . calibrateMyThermostat
```

What any self-respecting Haskeller would do is avail of so-called *newtypes* that allow one to distinguish otherwise-identical things at the type level:

```
newtype Dollar  = Dollar Double
newtype Celsius = Celsius Double
newtype Hertz   = Hertz Double
```

Then it's impossible to mix things up without being very deliberate about doing so.

We could do this in Hoon via tagged unions, of course, but I suspect the performance implications would be pretty nasty in certain areas, wires and paths included (if not, we should replace all synonyms with tagged unions immediately).  A newtype abstraction in Haskell has zero runtime cost, as all newtypes are erased during compilation -- Hoon could really use an equivalent.

(N.b. type synonyms in Haskell do have valid uses, of course.  One doesn't necessarily want to have to type out something like `StateT (Bar, [Baz]) (ReaderT Qux IO) a` all the time.)

ohAitch:
I believe you are describing the atom aura system, for your particular
example? But yeah "noun auras" have been on the back burner for a while now.

philipcmonk:
I do think we should use atom auras a lot more than we do (would be helped
by allowing hep in the name), and noun auras would also be good.

I think auras in general are a nice solution that usually prevents that
sort of error without getting in your way too much

Issue #1729, opened 2019-09-12
Sending malformed hoon to a ship via herb renders herb unusable


baudtack:
When you send a command to a running ship via herb

```
[ dkerschner@swampert:~/urbit ]% herb -d "(add 40 2)"                                                                                                                
WARNING <module> 411 - No pier or port specified, looking on port 12321
42

[ dkerschner@swampert:~/urbit ]%
```

This works. However if what is sent is malformed, such as

```
[ dkerschner@swampert:~/urbit ]% herb -d "(add 40 2)e"  
```

Your ship will produce 

```
syntax error
{1 11}
/~zod/home/0/app/dojo/hoon:<[993 11].[994 63]>
/~zod/home/0/app/dojo/hoon:<[990 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[989 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[988 13].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[988 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[987 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[975 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[974 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[973 9].[1.044 11]>
/~zod/home/0/app/dojo/hoon:<[973 5].[1.057 26]>
/~zod/home/0/app/dojo/hoon:<[972 5].[1.057 26]>
/~zod/home/0/app/dojo/hoon:<[1.146 17].[1.146 28]>
/~zod/home/0/app/dojo/hoon:<[1.146 9].[1.146 28]>
/~zod/home/0/app/dojo/hoon:<[1.145 9].[1.146 28]>
/~zod/home/0/app/dojo/hoon:<[1.164 44].[1.165 21]>
[%poke-lens source=[%dojo command='(add 40 2)e'] sink=[%stdout ~]]
/~zod/home/0/app/dojo/hoon:<[1.164 25].[1.165 21]>
```

The herb process then appears to hang and you will no longer be able to send herb request to that ship.

When you send a command to a running ship via herb

```
[ dkerschner@swampert:~/urbit ]% herb -d "(add 40 2)"                                                                                                                
WARNING <module> 411 - No pier or port specified, looking on port 12321
42

[ dkerschner@swampert:~/urbit ]%
```

[...]

PR #1888, opened 2019-10-26
lens: clear state in +poke-noun

ryjm:
if you use herb to send a dojo command that doesn't compile, the lens state doesn't clear and herb no longer works. this just adds a simple way to clear that state.

### DIFF
--- a/pkg/arvo/app/lens.hoon
+++ b/pkg/arvo/app/lens.hoon
@@ -234,6 +234,8 @@
   |=  a=*
   ^-  (quip move _this)
   ~&  poke+a
+  ?:  ?=  %cancel  a
+    [~ this(job.state ~)]
   [~ this]
 ::
 --


jfranklin9000:
I've tested this and it worked for me, but is the real solution is to not get in this state in the first place?

You'll want to preface your commit comment with `lens:` (or maybe `herb:` or `arvo:`?).

ohAitch:
How did this end up with a single global (unit command) instead of one per
request, anyway? Does that mean that if you run two blocking herb commands
simultaneously, one of them now fails?

joemfb:
This PR relates to #1729

@ohAitch, this lens app is a basic stub I wrote to keep `herb` working for CI after the eyre rewrite. It should absolutely support multiple, concurrent requests, have better (or any) error handling, &c. But it's (barely) sufficient for the minimal purpose it serves.

/cc @baudtack, who was considering improvements.

ryjm:
> This isn't the right approach -- `:lens` needs to handle invalid :dojo expressions by canceling the command and giving an HTTP 500 response on "negative acknowledgements" (ie, `+coup` where the `(unit tang)` is not null).

Seems pretty easy - I'll make an attempt sometime this week unless someone else gets to it first.

jtobin:
This has been sitting with no activity for a month, and it doesn't look like the approach here is the way to go, so I'll close it for the time being.  Feel free to patch and reopen.

PR #1889, opened 2019-10-27
chat-hook: use sender's timestamp

Fang-:
No longer overwrite messages' timestamp on-receive, instead keeping whatever
timestamp was set by the sender.

This behavior matches that of the late Hall.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -69,9 +69,7 @@
   ::  scry permissions to check if write is permitted
   ?.  (permitted-scry [(scot %p src.bol) %chat (weld path.act /write)])
     ~
-  =:  author.envelope.act  src.bol
-      when.envelope.act  now.bol
-  ==
+  =.  author.envelope.act  src.bol
   [ost.bol %poke / [our.bol %chat-store] [%chat-action act]]~
 ::
 ++  poke-chat-hook-action


loganallenc:
no! We should never trust sender’s timestamp.

loganallenc:
Please revert this promptly. There are maintainers of different files in this project for a reason.

Fang-:
I told you @jtobin. (;

> no! We should never trust sender’s timestamp.

@loganallenc I was under the impression Urbit was a high-trust network? Why would I not trust sender's timestamps?  
In practice these turn out to be more accurate representations of "time message was sent" than setting it locally on-receive, what with network slowness, timeouts, etc.

loganallenc:
You don’t trust the sender’s timestamp because they can say anything they can put any value into that field. A conversation can vary significantly in meaning if the order of the messages changes. In a “hosted chat” setting, you explicitly trust the host only.

Fang-:
And I'm saying the host should trust the senders. If someone's spoofing their own timestamps to try and mess up message history, kick 'em.

If the assumption here is that timestamps are used for message ordering on-screen, then maybe we need separate fields for time-sent and time-received. (But that's going to be rejected by all parties involved as unnecessary growth.)

I'm not willing to die for this PR, but just saying that my chat experience has been (lightly) negatively impacted by not having sent-at timestamps.

philipcmonk:
> I'm not willing to die for this PR, but just saying that my chat experience has been (lightly) negatively impacted by not having sent-at timestamps.

Can you describe this negative impact?

Fang-:
The network and key propagation being what they are, sometimes messages come in that were sent days in the past. Being able to distinguish those from messages of the current conversation keeps you from "talking to a ghost" in that sense, and is also a useful cue that "this ship had trouble but seems fixed".

(Again, pretty trivial stuff. This becomes more important, I guess, when people start doing Urbit over IPoAC and other high-latency networks.)

loganallenc:
@Fang- I don't think we need to retain time-sent and time-received. Individuals should not be able to spoof their timestamps. It's that simple. Remove the capability, and no one has to worry about it. Don't modify code to make it less secure.

loganallenc:
Much less modify files and merge them without the maintainer being able to review it when it's a non-critical fix. This sets a bad precedent which is, in my view, more important to address than the specific content of this PR.

philipcmonk:
The other reason to use the server timestamps is that any time you have multiple conceptions of time you have issues, especially if you try to compare them.  It's not uncommon for machines to be hours or days out of sync with each other.  It's fine to use their own conception of time for display, but not for anything formal, like ordering.

I think a two-time solution is reasonable (display the original timestamp in the order received by the server), but if it's a single time and it's ever used for comparing to other times, it should be the server time.  In hall was the client time used for ordering or was the ordering just based on when it was received?

jtobin:
On Mon, Oct 28, 2019 at 04:35:38PM -0700, Logan wrote:
> Much less modify files and merge them without the maintainer being
> able to review it when it's a non-critical fix. This sets a bad
> precedent which is, in my view, more important to address than the
> specific content of this PR.

My bad gents, I got this one wrong.  I'd interpreted this as a fix (as
it patched things to match the the behaviour of :hall) and only reviewed
it for correctness, not whether the feature was desirable or not (which
in general I would never do).

Since I thought it was more along the lines of a fix, I figured it would
be worth pulling into the weekly release -- I wouldn't have stepped in
if I'd correctly intuited it was the latter.

I'll be more careful in the future!

loganallenc:
Thanks @jtobin!

Fang-:
> In hall was the client time used for ordering or was the ordering just based on when it was received?

Ordering/sequence number was based on when it was received. Timestamps were generally for display only, though web client did use timestamps for ordering once upon a time.

_looks for a "comment & close" button_

jtobin:
Reverted in e77237d1a10c41d6686c52d8a248507cb4b856b7.

loganallenc:
Thanks again! Sorry for making a fuss.

PR #1892, opened 2019-10-30
app: Bitcoin Node JSON RPC API

yosoyubik:
This is still a work in progress, but I'd like to get some feedback to check if this is going in the right direction. 

You can see a small demo [here](https://youtu.be/k-TjZfMLqOc)

There are unit tests that check the proper request and response format but these have been tested manually, not based on the output from the RPC node, since it's not clear from the spec if, for example, some of the attributes returned by the API will be present. These are by no means comprehensive or even correct (that might even be a grant on its own...)

There is an attempt at an integration test in the app folder but that one only checks that the proper call to `%iris` is made. I would imagine that a proper integration test in this case (call `%iris` get the response back from the RPC node) would use the `+ph` test framework instead of the normal `+test`?

The authentication data for the RPC node is hardcoded in the +prep arm (this should be removed and included via integration with %sole) and the wallet path needs to be added as part of the state of the hook app (when there is only one wallet created there is no need to use it)

One last thing is the logic of the store app to stay in sync with the node. For example, it seems that a wallet with no name is created by default when the node is booted up. That will make the state already incompatible (if we create a wallet from the hook app, two will be present in the RPC node but only one in our "local" urbit store). You can see this in the demo video. Maybe a sync call needs to be made when the store app is booted?

Probably all of these has already been considered, solved and declared case closed. In any case I'd love to hear what the general direction is since I probably took too many liberties here.


loganallenc:
Wow! This is fantastic. Great job! Did a quick read-through, and the overall structure / patterns are great / exactly the direction we were going for.

I’d recommend the hook not have any “wallet” state and instead use scries to the btc-node-store to validate that there is a wallet in the store that a given action will be performed upon.

The hook is a good place to store authentication information for the BTC node, though.

bioy-casares:
It's aliiiive.

yosoyubik:
Thanks @loganallenc! :) I'll give some thought to using scries. I imagine there could be a default wallet that can be set up by the user, and the hook will always rely on the default one to make calls (it probably can be set up in a future sole-based terminal flow where we can also ask for the auth keys)

Btw, I forgot to add that I left a ton of comments on the `%/sur/` files from the RPC API spec, mostly so reviewers can compare with the actual description of the call, but of course, those should be removed/modified after we make sure it work. I don't guarantee that I copy-pasted the right ones though ;)

loganallenc:
> Maybe a sync call needs to be made when the store app is booted?

You likely wouldn't run this when the store app is booted, but rather when the hook app is booted.

pkg/arvo/lib/btc-node-json.hoon
```diff
@@ -0,0 +1,1080 @@
...
+      ::  the type of json (%s, %b...) only when unit is non empty so
+      ::  this is like a reverse +bond (applied to the non-null case)
+      ::  or a +bif that doesn't return a +unit
+      ::
+      ++  ferm  |*([a=(unit) t=term] ?~(a ~ t^u.a))
+      ::
+      ::  We want to make sure that numbers are @ud but JSON
+      ::  defines numbers as @ta, so we need to do a re-parse
+      ::
+      ++  feud  |*(a=(unit) ?~(a ~ n+(scot %ud u.a)))
```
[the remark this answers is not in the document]
yosoyubik: And that's why this needs proper tests... :) thanks for catching this one!

yosoyubik:
Thanks @philipcmonk! :)
 
> However, you could also not mock the node and just use the pass-through behavior of the default iris mock, so you could test against your running bitcoin node. This couldn't be run in CI because it doesn't have a separate bitcoin node.

This sounds more like what I was looking for. I wanted it mainly for development, so I didn't have to run the calls manually, but  I feel that these might be necessary since we are dealing with responses that we don't control (see bellow)

> How much testing is required depends on what sorts of issues you've run into. If you don't feel much would be gained by full integration tests compared to the unit tests, then the unit test are probably sufficient.

I ran into couple of issues comparing the expected response from the spec to the actual response. For example, regarding optional values. These are specified in the spec, but I discovered that [some were not marked as optional but were not returned](https://github.com/urbit/urbit/pull/1892/commits/19508dd23959687e5c8c932292485d7fad7578c5#diff-6b2ffcde7e5c82fe4c89ddd67e4bab3fR1071) in some cases.

Something I forgot to mention. I assumed that not all the responses are supposed to be stored (some don't return anything and others are just internal to the bitcoin node; e.g. `+encrypt-wallet`). In other cases it's not clear what the target of our action is. If we want to list labels, addresses... etc, the node has that information, should we then replicate that in our store?  Do we do a "are we sync" check and then subsequent calls target the store instead?

I need to do a bit of thinking around the "sync" feature of the hook app, mainly what it's exactly going to be synced, which means adding to my homework list exploring the existing integration with Ethereum to search for similarities :)

pkg/arvo/lib/btc-node-json.hoon
```diff
@@ -0,0 +1,1080 @@
...
+          ((slug |=(a=[@ @t] (cat 3 a))) (star hep) alp)
+      ::
+      ++  bip125
+        ::  This would have been more elegant:
+        ::  ['bip125-replaceable' (cu term so)]
+        ::  but a term is not of type ?(%yes %no %unknown)
+        ::  so a manual convertion seems the only(?) way.
+        ::
+        |=  t=@t
+        ^-  bip125-replaceable
```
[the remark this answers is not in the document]
yosoyubik: Right, keep forgetting that molds can be used as gates for validation. I knew there was a more elegant way to handle this :) thanks!

pkg/arvo/lib/btc-node-json.hoon
```diff
@@ -0,0 +1,1080 @@
...
+        ?:  =(t 'yes')
+          %yes
+        ?:  =(t 'no')
+          %no
+        ?:  =(t 'unknown')  !!
```
yosoyubik: This should be `?.` but will probably removed in favor of using the mold directly

philipcmonk:
> I wanted it mainly for development

I used pH extensively for exactly that purpose when I built the Ethereum integration.  I did later build a mock eth node, but only after getting it right on real geth and parity.  pH is a bit hairy (i've cleaned up the architecture on another unfinished branch, but it works basically the same) and the output is loud, but for your purpose you can probably get away with looking at the other tests in +manual-tests and modifying them to do something simple (as long as your output is "real" output and not `~&` -- we can't intercept those).

The default behavior is for all HTTP traffic to pass through to the real web, so you don't have to do anything there as long as your bitcoin node is running.  So, add a test to +manual-tests that's a variation on:

```
:+  %btc-test
  ~[~bud]
;<  ~  bind:m  (raw-ship ~bud ~)
;<  ~  bind:m  (just-events (dojo ~bud ":btc-node-hook [%list-wallets ~]"))
;<  ~  bind:m  (wait-for-dojo ~bud "wallets=<||>")
(pure:m ~)
```

And repeat those two two dojo lines as necessary to test stuff.  The commands to get it started are:

```
|start %ph
:ph|init
:aqua +solid
:ph|run %btc-test
```

If it doesn't finish (eg there's an error so it hangs), run `:ph|cancel`.

When you change files, they won't be loaded into the guest pill automatically.  Run `:aqua [%swap-files ~]` to change the files without having to re-run `+solid`.

Let me know if you run into trouble with this.

yosoyubik:
Thanks @philipcmonk I'll come back to this next week and let you know if I run in to any issues.

yosoyubik:
@philipcmonk I have addressed your suggestions (and in the meantime finished adding types for part 3 of the BTC Node grant; we are close to 50% completion! (: ) and started using ph.

When you say "real output" that's a sole-effect right? Would adding `~&` cause problems to parse output from dojo? (I would expect that they'd be ignored... mainly to avoid removing/adding those printfs)

philipcmonk:
> When you say "real output" that's a sole-effect right? Would adding ~& cause problems to parse output from dojo? (I would expect that they'd be ignored... mainly to avoid removing/adding those printfs)

Yes, sole-effect will give you real output.  If you're in a context where you can't easily do that, you can also send `%flog` moves to dill.  `~&` is a nock hint, so it's entirely outside the system -- it won't interfere with anything, but you also can't make pH look for it.

> Don't I need to add this as well to the manual-test to start the apps?

Yes, you'll need to start the apps and do any kind of setup you might need.

ohAitch:
feature request: ++mule variant that intercepts ~&s to a list it gives you
at the end/crash of the computation :P

(this would also be neat for web dojo work, though _that_ might want to be
more of a per-event stderr handler. I guess pH would also want to just
install one to the inner arvo?)

philipcmonk:
While we're at it let's make a +mule variant that intercepts nock 13 and turns it into ames packets, no *arbitrary effects*.  This is fun!

ohAitch:
sorry I'm not keeping up with the times, is nock 13 .^

philipcmonk:
Doesn't exist, I'm just saying that you can of course encode arbitrary effects with mule.  You don't actually need new operations, it's sufficient to use hints, as you suggest with ~&.

yosoyubik:
tl;dr: Is there a way to connect to %sole without modifying the prompt?

PS: I think I'll go ahead and send %flog moves, it might faster than finding out on my own :) 

I guess that when i get a %peer-sole, there is no way to not have a dedicated prompt for the app (I've tried not sending any moves, or setting a prompt with %vis=%.n without luck) If I do, either I need to do a `^X` from the ph test (which... I haven't found how to) or I won't be able to poke my app since dojo has cycled and anything I do from ph now will be "sent" to the BTC app (and I don't want to deal with sole-actions yet)

ohAitch:
Yes, I caught the sarcastic criticism on the capacity to use hint
interception for evil, not good; and was attempting to point out that when
.^ing files off another ship is for all intents and purposes sending ames
packets as a side effect. Non-arbitrary ones! But the proposed
hint-collector mechanism wouldn't be arbitrary either, and I do remember at
some point writing a chat demo that was two comets looping on each other's
clay

(As for 13, as we all know by all rights the scry operator should live in
slot 11, and nock 5 was the last true nock; but insofar as there's heresy
running around that considered it justifiable to bump from 11 to 12 in the
name of performance optimization, well that's halfway to 13 already)

Alternately, reducing generality/powerful from putting it in ++mock: how
would you feel about being able to set a ~& destination on a per-/event/
level? In eyre, taking an incoming http request whose session has requested
this, and bouncing it back out as a formal effect for "send me this again
but with all slog output bytes written immediately to this open
event-stream socket", should cover sole / general landscape tile
console-logs for debugging purposes. (Spill them back to the terminal if
the socket breaks). And I think pH could achieve something similar with the
setTimeout(0) causal isolation mechanism, though there the destination
would be a loopback buffer.

yosoyubik:
PPS: sending %flog works; yay! Now i just need to work with pH... it seems like none of my apps are starting. I see this:
```
"\1b[49m\1b[39m\1b[49m\1b[39m"                                                                   │[09:39:44] Starting 'urbit-co
"\1b[49m\1b[39m~bud:dojo> \1b[49m\1b[39m\1b[49m\1b[39m\1b[49m\1b[39m"                            │py'...
"\1b[49m\1b[39m~bud:dojo> \1b[49m\1b[39m\1b[49m\1b[39m|start %btc-node-store\1b[49m\1b[39m"      │[09:39:44] Finished 'urbit-co
"~bud: > |start %btc-node-store"                                                                 │py' after 61 ms
"\1b[49m\1b[39m~bud:dojo> \1b[49m\1b[39m\1b[49m\1b[39m|start %btc-node-store\1b[49m\1b[39m"      │[09:39:56] Starting 'urbit-co
"\1b[49m\1b[39m~bud:dojo> \1b[49m\1b[39m\1b[49m\1b[39m\1b[49m\1b[39m"                            │py'...
"~bud: >="                                                                                       │[09:39:56] Finished 'urbit-co
"\1b[49m\1b[39m~bud:dojo> \1b[49m\1b[39m\1b[49m\1b[39m\1b[49m\1b[39m"
```
but not `activated app home/btc-node-store`. Even if I explicitly wait for it:

```
    :+  %btc-test
      ~[~bud]
    ;<  ~  bind:m  (raw-ship ~bud ~)
    |=  pin=ph-input
    ?:  =(%init -.q.uf.pin)
      [& (dojo ~bud "|start %btc-node-store") %wait ~]
    ?:  (is-dojo-output ~bud who.pin uf.pin "activated app home/btc-node-store")
      [& ~ %done ~]
    [& ~ %wait ~]
```

philipcmonk:
Yeah, flog is probably best for this.  There isn't a way to send sole effects without taking a prompt.  You can send ctrl-x in ph with something like `(just-events (ctrl ~bud %x))`.

I think in some cases the "activated app" line is actually printed via ~& (or +slog which is the same thing).  We're not great with our ~& hygiene.

philipcmonk:
If you have a debugging query or something, then you could just send that and if it responds you'll know it started.

yosoyubik:
Thanks @philipcmonk You are right, that's what drum sends: `?.  se-ably  ((slog [%leaf txt]~) +>.$)`

I'll keep trying; never surrender! :)

yosoyubik:
I've pushed some changes addressing @philipcmonk's comments and scring a "default-wallet" from the store app as suggested by @loganallenc.

This also includes parts 3 and 1 of the BTC bounty.

The functionality as of now is the same as in the demo video. The only change is that after creating a new wallet you need to switch to it: e.g.`:btc-node-store|command [%switch-wallet 'local']`

This is necessary because when the node has more than one wallet, RPC calls related to wallets need to specify which one you referring too, and there seems to be an issue with the default wallet that is created initially (called just ' ', so an empty string) see [here](https://github.com/urbit/urbit/pull/1892/commits/39cffcb5bec7cada162fc968e50512e492da0365#diff-4a945d220cb4b524774bd1ae71341e21R169)

I've tested some of the calls using pH but I'm thinking that the wallet section is the most important one in the short term (probably next thing to create as seen [here](https://www.dropbox.com/s/bu1pp9378iv30eo/btc-wallet.png?dl=0) @bioy-casares?
so I'm planning to make sure all wallet calls work and integrate them properly in the store app.

I've used `@t` for all numeric/float values (fees, btc amounts...) but I've seen @pkova's works with the `@rd` parser so I'll be happy to update those once that work is completed.

yosoyubik:
All RPC calls for bounties 1, 3, 4 have been tested manually and work locally using the -regtest option. This is the [config file](https://github.com/yosoyubik/urbitcoin/blob/master/bitcoin.conf) I've used to test everything locally.

Or it can also be ran manually: 
```
bitcoind  -server=1 -rpcuser=urbitcoiner -rpcport=18443 -rpcpassword=urbitcoiner  -regtest  -rpcallowip=127.0.0.1 -daemon
```
Once both hook and store are started, credentials need to be manually set:
```
|start %btc-node-store
|start %btc-node-hook
> =credentials :*  %credentials
    'http://127.0.0.1:18443/'
    :~  ['Content-Type' 'application/json']
          ::  user/password: urbitcoiner/urbitcoiner
          ::
          ['Authorization' 'Basic dXJiaXRjb2luZXI6dXJiaXRjb2luZXI']
==  ==
> :btc-node-hook|command credentials
```
If you create more than one wallet, you can target any of the new ones with this command:
```
> :btc-node-store|command [%switch-wallet 'local']
```
There's currently a bug that I haven't been able to find where, if there are more than one wallet created, a "/wallet" endpoint needs to be appended to the URL (e.g. "127.0.0.1/wallet/local"), but that means that you can't target the "default" wallet (just ''). Doing a curl with that URL works though... ("127.0.0.1/wallet"). Any other wallet with a name works fine.

I've kept the store app minimal, storing just the names of the wallets plus some metadata. The rest of the calls to the RPC node are printed to the dojo. Possibly future grants, or consensual decisions on what belongs in the store, will address its current spartan state.

I started an unsuccessful attempt of using the new pH framework for testing all these calls, but ended up just doing it manually. My unsatisfactory work (which does not work) can be found [here](https://github.com/yosoyubik/urbitcoin/blob/master/src/ted/ph/ph-btc.hoon).

I haven't used the new `@rd` parser, so kept all float numbers as `@t`, in order to get this out of the way asap, and then later, when all the bounties are merged, update the remaining usages of float to their proper representation. I'll also be happy to be told to shut up about later work, and get it done now :)

yosoyubik:
I forgot to mention that `fim:ag` doesn't parse [`%bech32`](https://en.bitcoin.it/wiki/Bech32) addresses:

```
> `@uc`(rash '2N2jNiPqH9aVTMVnW45EyABURmHeTgYi72F' fim:ag)
0c2N2jNiPqH9aVTMVnW45EyABURmHeTgYi72F
> `@uc`(rash 'bcrt1qkuv4mcl3g40xsdcwjxsw2d7hm8yavgsrel56qq' fim:ag)
{1 14}
```
Unless there's another parser hidden somewhere, I've just tagged those address separately from `@uc` , not ideal,(`?(@uc [%bech32 @t])` but I thought it still could be of use while the parser gets updated to take care of those addresses. (`?(@uc [%bech32 @t])`

pkg/arvo/app/btc-node-store.hoon
```diff
@@ -0,0 +1,146 @@
...
+=<  |_  =bowl:gall
+    +*  this      .
+        btc-core  +>
+        bc        ~(. btc-core bowl)
+        def       ~(. (default-agent this %|) bowl)
+    ::
+    ++  on-init
+      ^-  (quip card _this)
+      :-  ~
+      %=    this
```
Fang-: Generally use `%_` when you can, it prevents you from accidentally changing the types of things.

pkg/arvo/app/btc-node-store.hoon
```diff
@@ -0,0 +1,146 @@
...
+      ::
+          %btc-node-store-command
+        (store-command !<(btc-node-store-command vase))
+      ==
+      ::
+      ++  store-action
+        |=  action=btc-node-store-action
+        ^-  (quip card _this)
+        =^  cards  state
+          ?+  -.action  ~|  [%unsupported-action -.action]  !!
```
Fang-:
```suggestion
          ?+  -.action  ~|([%unsupported-action -.action] !!)
```

Slightly more immediately obvious that it's just one thing (instead of something that flows over to the next line).

pkg/arvo/app/btc-node-store.hoon
```diff
@@ -0,0 +1,146 @@
...
+              %switch-wallet  (handle-switch:bc +.command)
+          ==
+        [cards this]
+      --
+    ++  on-watch  on-watch:def
+    ++  on-leave  on-leave:def
+    ::  +on-peek: read from app state
+    ::
+    ::    default-wallet  .^(@t %gx /=btc-node-store=/default-wallet/noun)
+    ::    n-wallets       .^(@ud %gx /=btc-node-store=/n-wallets/noun)
```
Fang-: Interface documentation, sweet! I'd recommend putting this at the top of the file (alongside more info on how to use this, optionally) for discoverability reasons, but we don't really have any standards around this yet.

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,354 @@
...
+    ++  on-save   !>(state)
+    ++  on-load
+      |=  old=vase
+      `this(state !<(state-zero old))
+    ::
+    ++  on-poke
+      |=  [=mark =vase]
+      ^-  (quip card _this)
+      |^
+      ?+    mark    (on-poke:def mark vase)
```
Fang-: imo it looks better to indent the "main" block of the `|^`, either one level, or putting is directly after the `|^` like you would with `:~`.

You might also consider moving the `+hook-action` and `+hook-command` into the `+bc` below, to keep the gall interface core here as simple as possible. Up to you though. Again, no well-established standards here yet.

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,354 @@
...
+  =;  dere
+    =+  res=((ar dere) a)
+    ?~  res  (need (dere a))
+    [%batch u.res]
+  |=  a=json
+  ^-  (unit response:rpc:jstd)
+  =/  res=(unit [@t json])
+    ::  TODO  breaks when no id present
+    ::
+    ((ot id+so result+some ~) a)
```
Fang-: Yeah, those are stupid cases to deal with. You'll want to parse for both individually, I guess? Can always just revert to doing `?>  ?=(%o a)` and then handling the `p.a` map directly.

(`a` is a terrible name and you know it. (; )

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,354 @@
...
+  |=  btc-resp=btc-node-hook-response
+  ^-  (quip card _state)
+  :_  state
+  ^-  (list card)
+  ;:  weld
+      ::  Print term for each succesful RPC request (used by :ph)
+      ::
+      ^-  (list card)
+      [%pass / %arvo %d %flog [%text (trip -.btc-resp)]]~
+      ?+    -.btc-resp  ::  ~|  [%unsupported-response -.btc-resp]  !!
```
Fang-: Just pull this big ol' `?+` out of the `weld` call. You could do `%+  weld  [%pass ...]~` I think, with the `?+` on the next line. Perhaps the `%flog`ing and actual response handling should be in separate arms in the first place.

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,354 @@
...
+  ^-  card
+  :*  %pass   pax
+      %agent  [our.bowl %btc-node-store]
+      %poke   [%btc-node-store-action !>(act)]
+  ==
+::
+++  default-wallet
+  =-  .^(@t %gx -)
+  %+  weld  /(scot %p our.bowl)/btc-node-store
+  /(scot %da now.bowl)/default-wallet/noun
```
Fang-: Fun fact, `.^` is n-ary, but boils down to two arguments. It can do list building for you, in that way. Sort of like a `:*`. ie:

```hoon
.^  @t
  %gx
  (scot %p our.bowl)
  %btc-node-store
  (scot %da now.bowl)
  /default-wallet/noun
==
```

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,354 @@
...
+++  n-wallets
+  =-  .^(@ud %gx -)
+  %+  weld  /(scot %p our.bowl)/btc-node-store
+  /(scot %da now.bowl)/n-wallets/noun
+::
+++  endpoint-url
+  |=  [act=btc-node-hook-action]
+  ?.  ?|  =(-.act %abandon-transaction)
+          =(-.act %add-multisig-address)
+          =(-.act %backup-wallet)
```
Fang-: You probably want `?=` on all of these checks, to more explicitly check for/specialize the type of the `act`.

pkg/arvo/lib/btc-node-json.hoon
```diff
@@ -0,0 +1,2115 @@
...
+    ::
+    ++  vin-parser
+      =-  (ar (ou -))
+      :~  =-  ['txid' (uf ~ (mu -))]
+          (cu to-hex so)
+          ['vout' (uf ~ (mu ni))]
+          =-  ['scriptSig' (uf ~ (mu (ot -)))]
+          :~  ['asm' so]
+              =-  ['hex' -]
+              (cu to-hex so)
```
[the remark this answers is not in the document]
yosoyubik: I think I was pushing for too much code symmetry :D

pkg/arvo/sur/btc-node-hook.hoon
```diff
@@ -0,0 +1,1566 @@
...
+++  btc-rpc
+  |%
+  +$  request
+    $%
+    ::  Blockchain
+    ::
+        ::  %getbestblockhash: Returns the hash of the
+        ::  best (tip) block in the longest blockchain.
+        ::
+        [%get-best-block-hash ~]
```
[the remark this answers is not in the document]
yosoyubik: d'oh! I used the RPC method instead of the actual hoon tag (this is from copy-pasting from the BTC docs...)

pkg/arvo/sur/btc-node-hook.hoon
```diff
@@ -0,0 +1,1566 @@
...
+        [%get-addresses-by-label label=@t]
+        ::  %get-address-info: Return information about the given bitcoin
+        ::  address.
+        ::
+        [%get-address-info =address]
+        ::  %get-balance: Returns the total available balance.
+        ::
+        $:  %get-balance
+            $?  ~
+                $:  dummy=(unit %'*')
```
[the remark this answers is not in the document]
yosoyubik: It... should not? tbh i don't remember the exact reasoning behind why the union. The main idea is avoiding something like `[%method ~ ~ ~ ~ ~ ~ ~ ~]` but a `unit` works fine, the only change is that then you need to `[%method [~ [p q ~ ...]])]`. 

But I think units is more clear, so I'll use that instead.

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: I'm going to remove the flogging since its purpose was to detect successful calls in pH, but I ended up not using it.

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: This is not my code you honor! :D
I just realized that I removed the comment that pointed to its [source](https://github.com/urbit/arvo/pull/973/commits/be296f6897ca6c96225d844912d7e92ea93ea75a#diff-32b4dc798ca3aa3aa5c9fbd963627d4eR1941) but it's still in my own repo:
https://github.com/yosoyubik/urbitcoin/blob/4a75a1b0736d7e02271fa7165365fcb7516b7adf/src/app/btc-node-hook.hoon#L287

yosoyubik: I think I'll end up using this:

```
  =/  res=[id=(unit @t) res=(unit json)]
    %.  jon
    =,  dejs:format
    =-  (ou -)
    :~  ['id' (uf ~ (mu so))]
        ['result' (un some)]
    ==
```
If id or result are not present you'll get units.

pkg/arvo/sur/btc-node-hook.hoon
```diff
@@ -0,0 +1,1566 @@
...
+      ::  FIXME: a list for ux and map for the cell
+      ::  produces a fish-loop
+      ::
+      $?  ::  (for verbose = false)
+          ::
+          @ux                      :: (list @ux)
+          ::  (for verbose = true)
+          ::
+          [id=@ux =mem-pool]       :: (map id=@ux mem-pool)
+      ==
```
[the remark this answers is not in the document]
yosoyubik: I've moved all the `%-  unit` `%-  list` to the lines bellow.
I think you are right, at least when I experimented using lists and maps, all my attempts at making it work were in vane. 

Let me see if I follow your suggestion. The response is either a json array `(list @ux)`, or a json object `(list [@ux *])`, so it seems to me that the type will need to be a union, unless that, in the case of an array, you always produce `*mem-pool` for the tail of the pair?

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: First suggestion, done.
Second suggestion... I might leave it as it, for the time being (I'd like to get this out of the way asap (: ) But I agree with you that keeping the gall core as spartan as possible is desirable.

pkg/arvo/sur/btc-node-hook.hoon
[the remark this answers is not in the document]
ohAitch: Are they complex types? imo this should be a `$@` not a `$?`, which has no semantic difference but hammers home "fork between atom and cell".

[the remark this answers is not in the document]
Fang-: > it seems to me that the type will need to be a union

Yes, but the type system is tripping up over that. So instead, I'm suggesting using `$%`, something like:

```hoon
+$  mem-pool-response
  $%  [%soft ids=(list @ux)]
      [%loud pool=(map @ux mem-pool)]
  ==
```

My additional suggestion was to provide a library function that `$-(mem-pool-response (list @ux))`, so that if you just want the `(list @ux)` from the response, you don't have to re-implement the destructuring every time. This may or may not make sense given the context, apply your best judgement.

ohAitch:
++each perhaps?

pkg/arvo/sur/btc-node-hook.hoon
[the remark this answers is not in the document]
yosoyubik: Ah right, my confusion came from me thinking that you were referring to the "current" one (list of either atom or cell). That doesn't seem to have issues with the type system?

I might end up using @ohAitch' (`$@`instead of `$?`), but leaving it as a list for both (I'll need to do a bit of rewriting to make it into a map). Thanks for the suggestions!

yosoyubik:
Thanks @Fang- for your feedback! I believe that I have addressed all your comments and pushed the updates accordingly; let me know if there's something missing.

This also includes a template for unit tests for the wallet section that is not meant to be considered a comprehensive collection (although the ones included work) but rather a placeholder for a maybe future set of proper tests.

I'll be interested to know if anybody manages to run a dummy node in regtest mode and make it work? It "should" be doable with the instructions I left in previous comments but the interface for the hook app is very minimal, so there's some manual configuration (switch wallets after creating a new wallet... etc) that's needed before it actually works.

Fang-:
Thank you! I'll try playing with this sometime next week.

yosoyubik:
Awesome! Btw @Fang- I've only included the stuff relevant for the BTC grant, but this branch is very outdated, so I don't know what's the best way to bring it up to date. You might be better off trying things out from my local branch: https://github.com/yosoyubik/urbitcoin

pkg/arvo/lib/btc-node-json.hoon
```diff
@@ -0,0 +1,3129 @@
...
+      ^-  cord
+      =/  a=tape  (scow %if ip)
+      =/  b=json  (numb:enjs:format port)
+      ?>  ?=([%n *] b)
+      ?~  a  ''
+      (cat 3 (cat 3 (crip t.a) ':') p.b)
+    ::
+    ++  to-wall
+      |=  =tape
+      ^-  wall
```
Fang-: Huh, I'd expect this to be in the stdlib. Might be worth adding (in a separate PR), next to `+to-wain`.

pkg/arvo/lib/btc-node-json.hoon
```diff
@@ -0,0 +1,3129 @@
...
+                ['hex' (un (cu to-hex so))]
+              ::
+                ['req-sigs' (uf ~ (mu ni))]
+              ::
+                ['type' (un so)]
+              ::
+                :-  'addresses'
+                =-  (uf ~ (mu -))
+                (ar (cu addr-type-validator so))
+        ==  ==
```
Fang-: For cases like this, it's fine to keep the simpler ones grouped together, like so:

```hoon
:~  ['asm' (un so)]
    ['hex' (un (cu to-hex so))]
    ['req-sigs' (uf ~ (mu ni))]
    ['type' (un so)]
  ::
    :-  'addresses'
    =-  (uf ~ (mu -))
    (ar (cu addr-type-validator so))
==
```

This way, you cleanly delineate the complex codeblocks, but don't stretch out the simple ones unnecessarily.

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+    --
+::
+|_  =bowl:gall
+::
+++  httr-to-rpc-response
+  |=  hit=httr:eyre
+  ^-  response:rpc:jstd
+  ~|  hit
+  :: ?.  ?=(%2 (div p.hit 100))
+  ::   fail+hit
```
Fang-: This should probably be `=(2 (div p.hit 100))`. Do we not want to enable this for some reason?

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+  ?:  ?=([^ * ~] res)
+    `[%result [u.id.res ?~(res.res ~ u.res.res)]]
+  ~|  jon
+  ::%-  need
+  :: ((ot id+so error+(ot code+no message+so ~) ~) jon)
+  `[%error (handle-error jon)]
+::
+++  handle-error
+  |=  =json
+  ?~  json  ['' '' '']
```
Fang-: Would love a typecast here. Maybe just include the `%error` tag, and cast to `$>(%error response:rpc:jstd)`.

(It's really more like `+parse-error`, isn't it? Kinda surprised we don't have a generic parser for rpc responses in the stdlib. Clostest thing is [this in /lib/ethio](https://github.com/urbit/urbit/blob/6ef08962efa5aef197fdf961261e918075491d13/pkg/arvo/lib/ethio.hoon#L107-L119).)

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+      :-  'error'
+      =-  (uf ['' ''] -)
+      =-  (cu |*(a=(unit) ?~(a ['' ''] u.a)) (mu (ou -)))
+      :~  ['code' (uf '' no)]
+          ['message' (uf '' so)]
+  ==  ==
+::
+++  http-response
+  |=  [=wire response=client-response:iris]
+  ?.  ?=(%finished -.response)
```
Fang-: Please add a `^-` for sanity guarantees. (This is technically already checked higher up in the stack, but casting inside every non-trivial `|=` just makes life easier.)

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+    ^-  (list card)
+    ~&  [%address-info +.btc-resp]
+    ~
+  ::
+      %get-balance
+    ^-  (list card)
+    :_  ~
+    :*  %pass  /  %arvo  %d  %flog
+        %text  "amount={(trip +.btc-resp)}"
+    ==
```
Fang-: What's the reason for the fancy printing on this? Everything else just does `~&`. Maybe `+slog` is easier.

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+    ^-  (list card)
+    ~&  [%transactions +.btc-resp]
+    ~
+  ::
+      :: %list-unspent
+      :: %list-wallet-dir
+  ::
+      %list-wallets
+    ^-  (list card)
+    [(btc-node-store-poke /list [%list-wallets ~])]~
```
Fang-: So, a node response is telling us to... print the wallets to the screen? Why?

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+    %btc-node-store
+    (scot %da now.bowl)
+    /n-wallets/noun
+  ==
+::
+++  endpoint-url
+  |=  [act=btc-node-hook-action]
+  ?.  ?|  ?=([%abandon-transaction *] act)
+          ?=([%abort-rescan *] act)
+          ?=([%add-multisig-address *] act)
```
Fang-: I'd simplify to `?=(%tag -.act)` for all of these.

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+          ?=([%wallet-lock *] act)
+          ?=([%wallet-passphrase *] act)
+          ?=([%wallet-passphrase-change *] act)
+          ?=([%wallet-process-psbt *] act)
+      ==
+    endpoint
+  ?:  ?=([?(%dump-wallet %import-wallet) filename=@t] act)
+    (crip "{(trip endpoint)}wallet/{(trip filename.act)}")
+  ::  FIXME: fails when default-wallet is '' and more than 1 wallet
+  ::  has been created
```
Fang-: Considering the default `default-wallet` is `''`, this seems pretty common. Probably want to fix this prior to merge?

I don't understand why we have the empty name as default, anyway. Is that mirrored on the node side? I'm not very surprised it's giving issues with `wallet/` as top-level endpoint vs `wallet/` as pointing to the wallet with the empty name. (If I'm interpreting this situation correctly.)

pkg/arvo/app/btc-node-hook.hoon
```diff
@@ -0,0 +1,373 @@
...
+  ::  after a wallet is created, in order to use it.
+  ::
+  ::  e.g.  > :btc-node-store|command [%switch-wallet 'local']
+  ::
+  ?:  (lte n-wallets 1)
+    endpoint
+  %-  crip  %+  weld
+    "{(trip endpoint)}wallet/"
+  ?:  =('' default-wallet)  ~
+  "{(trip default-wallet)}"
```
Fang-: Removing the `?:` here is fine, since the end result is the same. (And then you get to remove the weld too. You could `:((cury cat 3) endpoint 'wallet/' default-wallet)` if you don't want to go back and forth between cord and tape, I guess.)

pkg/arvo/sur/btc-node-hook.hoon
```diff
@@ -0,0 +1,2489 @@
...
+        [%remove-pruned-funds txid=@ux]
+        ::  %rescan-blockchain  Rescan the local blockchain for wallet related
+        ::  transactions.
+        ::
+        [%rescan-blockchain start-height=(unit @ud) stop-height=(unit @ud)]
+        ::  %send-many: Send multiple times.
+        ::  Amounts are double-precision floating point numbers.
+        ::
+        $:  %send-many
+            dummy=%''
```
Fang-: I think you want `%$`, which is equivalent, but more-correct notation for "empty constant".

pkg/arvo/sur/btc-node-hook.hoon
```diff
@@ -0,0 +1,2489 @@
...
+            $=  groups
+            %-  list
+            %-  list
+            %-  list
+            $:  address
+             ::
+                amount=@t
+              ::
+                label=(unit @t)
+        ==  =
```
Fang-: Should this be done within this PR?

pkg/arvo/sur/btc-node-store.hoon
```diff
@@ -0,0 +1,180 @@
...
+        ::
+        [%switch-wallet @t]
+      ::
+        ::  Loads an external wallet
+        ::
+        [%load-wallet @t]
+      ::
+        ::  TODO: do the actual syncing
+        ::
+        [%sync ~]
```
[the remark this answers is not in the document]
yosoyubik: I think that we should follow Bcoin's structure and segregate the node and wallet logic, so it might be better to postpone that so we have a clear idea of what goes in the wallet %hook/%store apps.

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: This is a leftover of the `+ph` testing. I'll fix it

pkg/arvo/app/btc-node-store.hoon
```diff
@@ -0,0 +1,149 @@
...
+    ++  on-leave  on-leave:def
+    ::  +on-peek: read from app state
+    ::
+    ++  on-peek
+      |=  =path
+      ^-  (unit (unit cage))
+      ?+  path  (on-peek:def path)
+          [%x %default-wallet ~]  ``noun+!>(default-wallet)
+          [%x %n-wallets ~]       ``noun+!>(~(wyt by wallets))
+      ==
```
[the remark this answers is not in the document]
yosoyubik: It was a way to control for when there are more than one wallets created, since then the endpoint format used to call the node needs to change. When we have only one wallet (on the BTC node) the calls that handle wallet logic have in this format: `http://127.0.0.1:8332/` but if there is more than one wallet we need to append `wallet`or it will fail: `http://127.0.0.1:8332/wallet/<THE_WALLET>`. I'm gonna check all of these again to make sure that this behavior holds.

pkg/arvo/sur/btc-node-hook.hoon
```diff
@@ -0,0 +1,2489 @@
...
+    ::      - %decode-psbt
+    ::
+    +$  sig-hash
+      $?  %'NONE'
+          %'SINGLE'
+          %'ALL|ANYONECANPAY'
+          %'NONE|ANYONECANPAY'
+          %'SINGLE|ANYONECANPAY'
+          %'ALL'
+      ==
```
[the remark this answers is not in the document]
yosoyubik: I think that the RPC API[1] requieres the type of the signature to be send exactly like that. We really don't need for the type to match it but it was just to avoid having a conversion functions

- [1] https://bitcoincore.org/en/doc/0.19.0/rpc/rawtransactions/signrawtransactionwithkey/

pkg/arvo/sur/btc-node-store.hoon
```diff
@@ -0,0 +1,180 @@
...
+    ==
+  ::
+  +$  update
+    $%  ::  Wallet name has changed
+        ::
+        [%wallet-name name=@t]
+        ::  Updates wallet attributes
+        ::  FIXME: all attrs might need to be units...
+        ::
+        [%wallet-attrs name=@t attr=wallet-attr]
```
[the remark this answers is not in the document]
yosoyubik: I would leave this out of this PR until we start working on the wallet

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: Re-adding the handle-error (called now `parse-error`) here instead so we get a better error message without http headers.

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: This should ask the full node to list the wallets that it contains. The problems comes when we can have inconsistencies with the wallet the %store app has. Now, I agree that this kinda doesn't make much sense and will need to be addressed when deciding what's really the purpose of the store, and with the sync behavior in place. Fo this first pass I just assumed that the control is made only from you urbit (and to show the connection with the %store app) so the consistency is maintained, but of course, this breaks without the syncing feature. I can also ignore what's in the %store and print the response from the full node.

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: The full node creates a wallet called  `''`, so when only that wallet exists we can just use the url without specifying the wallet name. If we create another wallet, then we need to append `/wallet/<wallet_name>`. If we want to address the `''` wallet, the curl example here works `('http://127.0.0.1:18443/wallet/'`) but I couldn't make this work when creating a `request:http` (I get a `404` with no other message). I had assumed that I was not creating the `request:http` the right way, or that there was another way to do it because the curl works:
```
curl --location --request POST 'http://127.0.0.1:18443/wallet/' \       
--header 'Content-Type: application/json' \
--header 'Authorization: Basic dXJiaXRjb2luZXI6dXJiaXRjb2luZXI=' \
--data-raw '{"id":"curltext","method":"listtransactions","params":[]}'
```
but the `request:http` does not:
```
[ method=%'POST'
  url='http://127.0.0.1:18443/wallet/'
  header-list=~[[key='Content-Type' value='application/json'] [key='Authorization' value='Basic dXJiaXRjb2luZXI6dXJiaXRjb2luZXI']]
  body=[~ [p=82 q='{"params":[],"id":"list-transactions","jsonrpc":"0.2","method":"listtransactions"}']]
]
```

pkg/arvo/sur/btc-node-hook.hoon
yosoyubik: whoops, I completely missed this comment and now I can't see where was pointing at. Was it about `@uc` not parsing segwit addresses? If so I would say "later..."

pkg/arvo/app/btc-node-hook.hoon
yosoyubik: Correction: This is not a problem when different wallets are created. This issue comes when multiple wallets are loaded at the same time. This is a bit more complicated then since wallets can be loaded/unloaded without the gall app knowing about it. If we just had that sync feature...

loganallenc:
What's the status on this?

yosoyubik:
@loganallenc This (and the next) comment sums up the (I believe) last remaining issue: https://github.com/urbit/urbit/pull/1892#discussion_r394333359

TL;DR: When >1 wallets are loaded on the full node, a `curl` to `http://127.0.0.1:18443/wallet/` address works 

<details><summary>expand</summary>

```
curl --location --request POST 'http://127.0.0.1:18443/wallet/' \ 
--header 'Content-Type: application/json' \
--header 'Authorization: Basic dXJiaXRjb2luZXI6dXJiaXRjb2luZXI=' \
--data-raw '{"id":"curltext","method":"listunspent","params":[]}'
```

<code>{"result": ...}</code>
</p>
</details>

but the same POST request from your urbit gives a 404:

<details><summary>expand</summary>

```
[ method=%'POST'
  url='http://127.0.0.1:18443/wallet/'
  header-list=~[[key='Content-Type' value='application/json'] [key='Authorization' value='Basic dXJiaXRjb2luZXI6dXJiaXRjb2luZXI']]
  body=[~ [p=72 q='{"params":[],"id":"list-unspent","jsonrpc":"0.2","method":"listunspent"}']]
]

> :btc-node-hook|action [%list-unspent ~]
>=
error in %btc-node-hook
%arvo-response
! /~zod/home/0/app/btc-node-hook/hoon:<[107 39].[107 51]>
! /~zod/home/0/app/btc-node-hook/hoon:<[107 37].[107 51]>
! /~zod/home/0/app/btc-node-hook/hoon:<[107 23].[107 52]>
! /~zod/home/0/app/btc-node-hook/hoon:<[107 17].[107 53]>
! /~zod/home/0/app/btc-node-hook/hoon:<[107 3].[129 21]>
! [p=404 q=~[[p='date' q='Thu, 02 Apr 2020 04:43:08 GMT'] [p='content-length' q='0'] [p='content-type' q='text/html; charset=ISO-8859-1']] r=~]
! /~zod/home/0/app/btc-node-hook/hoon:<[106 3].[129 21]>
! /~zod/home/0/app/btc-node-hook/hoon:<[105 3].[129 21]>
! /~zod/home/0/app/btc-node-hook/hoon:<[155 5].[158 23]>
! /~zod/home/0/app/btc-node-hook/hoon:<[154 3].[163 42]>
! /~zod/home/0/app/btc-node-hook/hoon:<[153 3].[163 42]>
! /~zod/home/0/app/btc-node-hook/hoon:<[151 3].[163 42]>
! /~zod/home/0/app/btc-node-hook/hoon:<[150 3].[163 42]>
! /~zod/home/0/app/btc-node-hook/hoon:<[94 31].[94 63]>
! /~zod/home/0/app/btc-node-hook/hoon:<[93 9].[95 11]>
! /~zod/home/0/app/btc-node-hook/hoon:<[92 7].[96 19]>
! /~zod/home/0/app/btc-node-hook/hoon:<[91 7].[96 19]>
! /~zod/home/0/app/btc-node-hook/hoon:<[90 7].[96 19]>
! /~zod/home/0/lib/verb/hoon:<[72 21].[72 48]>
! /~zod/home/0/lib/verb/hoon:<[72 3].[73 15]>
! /~zod/home/0/lib/verb/hoon:<[71 3].[73 15]>
! /~zod/home/0/lib/verb/hoon:<[70 3].[73 15]>
! /~zod/stage/~2020.3.24..22.34.00..f53d/sys/vane/gall:<[1.213 9].[1.213 47]> 
```
</details>

My guess was that there was something wrong with how I was making the POST request, or how iris handles that request, but I haven't figured out where the issue is exactly. I'm assuming that if the curl works, the same request should work also from within urbit, but I could be wrong.

Fang-:
@yosoyubik oh! I think I know what this is.

See first, [our use of `de-purl`](https://github.com/urbit/urbit/blob/master/pkg/urbit/vere/cttp.c#L550) in `cttp.c`.

Then, observe:

```hoon
> (de-purl:html 'http://127.0.0.1:18443/wallet/')
[~ [p=[p=%.n q=[~ 18.443] r=[%.n p=.127.0.0.1]] q=[p=~ q=<|wallet|>] r=~]]
::
> (de-purl:html 'http://127.0.0.1:18443/wallet')
[~ [p=[p=%.n q=[~ 18.443] r=[%.n p=.127.0.0.1]] q=[p=~ q=<|wallet|>] r=~]]
```

The trailing slash is considered insignificant, and discarded as such. I'd guess that if you ran the curl without the trailing slash, it wouldn't succeed.

This has... been the case since forever, actually. Of course it's erroneous behavior, since there is at least a semantic difference between `/a` and `/a/`, and very often a technical one as well.

If `cttp.c` doesn't throw any curveballs at me, I can probably get this done today. Will keep you posted! (As per https://github.com/urbit/urbit/pull/2589#issuecomment-602828046 you might need to compile a new binary yourself until we put out a new release, if you want to test it out.)

yosoyubik:
> I'd guess that if you ran the curl without the trailing slash, it wouldn't succeed.

That's exactly right, it actually doesn't throw any errors or anything.

This is great! Thanks @Fang- for looking into it, and very happy to see this one near completion... at least for now ;)

Fang-:
Try pulling in #2656, rebuilding the binary with the updated pill there, and let me know if this fixes your issue.

yosoyubik:
> let me know if this fixes your issue.

It does! I'm adding a couple more changes in the hook/store to load external wallets that have not been created locally (but exists in the remote full node).
 
Thanks a lot @Fang-!

yosoyubik:
I'm done testing the RPC API calls on mainnet (disclaimer: I've *just* tested that the RPC calls don't fail (i.e. correct input format) when sent to the full node), now up to date with the latest release (v0.19)

I haven't been able to get a response from some of these, several due to %dojo or %eyre crashing on a (I believe) very long output message (for: `%get-block-hash `%0`, `%get-block hash `%2`, `%get-block-template` I got things like `(recover: dig: over)` or `(allocate: reclaim: half of 37 entries)`). 

I've also tested BTC grant 2 (control, generating, mining, and  network) and found some errors in these two api calls: `%get-mining-info`, `%get-network-hash-ps` due to the `ne:dejs:format` parser not being able to handle an input with the following format :

```
> (ne:dejs:format [%n p='8.660650844542463e+19'])
{1 19}
syntax error
```

Once the parser is updated to handle that case the API call should work fine.

Fang-:
> I haven't been able to get a response from some of these due to %dojo (or %eyre (?)) crashing on a (I believe) very long output message (for: `%get-block-hash [~ %0]`, `%get-block hash [~ %2]`, `%get-block-template` I got things like `(recover: dig: over)` or `(allocate: reclaim: half of 37 entries)`). 

Wait, but `%get-block-hash` should only be giving you a block hash, right? What is it receiving/processing/printing that's so big it stack overflows?

yosoyubik:
Whoops, I meant `%get-block` (it has different levels of verbosity). 

https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblock/

Fang-:
> It returns a pretty long hex-encoded string

Well it certainly shouldn't be `dig: over`ing on parsing/handling that! Maybe it's looping somewhere? Curious to hear how far it gets/where it actually trips up.

yosoyubik:
Let me investigate.

(this is what's doing rn:
Screen Shot 2020-05-27 at 4 38 49 PM )

joemfb:
You can disable the stack overflow handler (with `#define NO_OVERFLOW` in `noun/manage.c`) and get core dumps instead of hoon stack traces. That can be particularly helpful if the hoon traces are so big that they won't render.

yosoyubik:
I've narrowed down the error to  the parser in`+de-json:html`:
```
=/  jon=json  (need (de-json:html q:(need r.hit)))
```
Any suggestions on what's the appropriate way to deal with this?

Fang-:
Of course it's json parsing. ):

Can you post an example value of `q:(need r.hit)` for which this happens? I want to try poking at this a bit.

yosoyubik:
It's a ~3MB hex encoded string... 

https://gist.github.com/yosoyubik/c52a88533b3b5877c0e3e07bda64acf6

I was hitting something similar while working on the Canvas app (sending ~1MB SVG element, so I had to implement a pseudo chunking and then reassemble it later when all the chunks had arrived.

Fang-:
Hmm. Json string parsing (and many other things) depend on `+star`, which depends on `+stir`. Neither its hoon code nor its jet are tail recursive, so I guess these can stack overflow on long inputs.

I think the only way around this is to rewrite the jet (and optionally the hoon, but doesn't really matter) to be tail recursive. Can someone confirm/deny? Seems like a non-trivial rewrite, but maybe I'm overlooking a simple solution.

ohAitch:
Can definitely confirm that on my machine,
- `(scan (reap 0x7.ff00 'a') (cold ~ (star next)))` works fine
- `(scan (reap 0x7.ff29 'a') (cold ~ (star next)))` spins for a bit but resolves
- `(scan (reap 0x7.ff2a 'a') (cold ~ (star next)))` stack overflows

The jet seems annoying to rewrite tail-recursive since it's a left-fold, honestly the actual architecture change before we're comfortable slinging around megabytes would be making parsers work on cords (and probably a reference-counted atom slice implementation so `++cut` doesn't always copy). There's no excuse for taking the 3MB hex string and allocating it as a 192-bit-per-character linked list only to drop it back into a `[%s @t]` again - maybe with some JSON escape sequence decoding, but in this case verbatim!

<sub>Okay I don't completely understand how the special cell allocator works, I'm _assuming_ tape layout is still a 6-word size+rc+mug+car+cdr+size situation.</sub>

joemfb:
@ohAitch, I mostly agree with all that. Cells are definitely still arranged as you describe, so even if this were tail recursive, it'd still use an absurd amount of memory while processing. It might still be worth working on the `+stir` jet, but I'd have to think about how to do that.

Ideally, both a tape and a cord could be represented as a byte array at runtime, but that requires indirection facilities in the memory model that we simply don't have. I'm not sure how much harder it would be than a refcounted atom slice, which would require similar facilities to be broadly useful (the latter would be more directly helpful for `+cue` though).

As much as I hate to say it, the simplest thing to do now to fix both the stack overflow and the memory usage might be to directly jet the json parser.

ohAitch:
I'd recommend first taking a shot at implementing `de-json` as cord-to-cord in hoon before an outright jet; possibly with something gross that detects long chunks in unescaped quotes, substitutes them out with placeholders during the "trip", and then after parsing with the general parser tree-walks them back into place.

vvisigoth:
@joemfb @philipcmonk 

I'd like to figure out a way to get this merged, either by hacking around this error, or scheduling the jet rewrite. Any suggestions on a way forward here?

joemfb:
Here's a tail-recursive `+stir`:

```
++  stir
  ~/  %stir
  |*  {rud/* raq/_=>(~ |*({a/* b/*} [a b])) fel/rule}
  ~/  %fun
  |=  tub/nail
  ^-  (like _rud)
  =+  ^=  [lef wag]
    =|  lef=(list [p=hair q=[~ u=_(need q:(fel tub))]])
    |-  ^+  [lef *(like _rud)]
    =+  vex=(fel tub)
    ?~  q.vex
      [lef wag=[p.vex [~ rud tub]]]
    $(lef [vex lef], tub q.u.q.vex)
  ::
  |-  ^-  (like _rud)
  ?~  lef
    wag
  =*  vex  i.lef
  ?>  ?=(^ q.wag)
  %=  $
    wag  [(last p.vex p.wag) [~ (raq p.u.q.vex p.u.q.wag) q.u.q.wag]]
    lef  t.lef
  ==
```

The translation to u3 in the jet should be mechanical (and can take advantage of call-site caching). I'd be happy to do the rest of the development and staging if this is a priority.

Feedback requested.

/cc @ohAitch @frodwith

vvisigoth:
@frodwith @ohAitch do you have any feedback here by any chance?

frodwith:
Seems like the shortest path to victory. I didn't look at the tail recursive +stir too hard, but a couple of general notes:
* If this is cleanly a left fold, it would be nice to write it in terms of +roll and make that tail recursive instead.
* Maybe we should (also?) jet json parsing per se, since we do an awful lot of it.

frodwith:
In case it's relevant, https://github.com/urbit/urbit/pull/3355

yosoyubik:
I finally had some time to test this. Good news is that the rewrite of the jets has fixed the issue with the json parsing, but now it crashes in the cord parser.

I've tried with both `(cu |=(a=@t (rash a hex)) so)` and `(su hex)`. This is for the ~3Mb hex-encoded string I mentioned before https://github.com/urbit/urbit/pull/1892#issuecomment-635312635

There is also another issue that comes from `~&`'ing the output of `:btc-node-hook|action [%get-block hash [~ %2]]` that causes a `allocate: reclaim: half of 20 entries`. The response here is an object with info about [the block and all the transactions ](https://bitcoincore.org/en/doc/0.19.0/rpc/blockchain/getblock/)in it. (I'm guessing that this is too big to just print it to dojo)

Any ideas?

PS: Next month this PR will turn one year old!

joemfb:
Is this `+to-hex` in `lib/btc-node-json/hoon`? Can you post something about the codepath taken by that 3MB hex string, or even just the stack-trace of the crash?

I'm not surprised that the printer falls over on values that large. You might be able to get away with formatting it as a tank yourself and printing with `~_` (bypassing the automatic, type-driven formatting of `~&`), but that whole hint-based ("slog") printing pathway is really just for debug output.

yosoyubik:
> Is this +to-hex in lib/btc-node-json/hoon?

Yes, that's right (Here is where it crashes: https://github.com/urbit/urbit/pull/1892/files#diff-8e33a9dd288d524a754bd838751a4a85R88).

I have a python server giving me the hex string as json and am using this snippet so I don't need to connect to the btc full-node every time.

```
/-  spider
/+  strandio
=,  strand=strand:spider
^-  thread:spider
|=  arg=vase
=/  m  (strand ,vase)
^-  form:m
;<  =json  bind:m  (fetch-json:strandio "http://127.0.0.1:8000/")
=/  data=@t
  %.  json
  =,  dejs:format
  %-  ot
  ['data' (su hex)]~  :: doing ['data' so] works fine
:: ~&  data
(pure:m !>(%done))
```
I don't see a stack trace if I just let it run (it shows `«http»` for a while and then `allocate: reclaim: half of 20 entries`, although not every time I've tried; other times it spins for way too long...). If I force the crash with ctrl+c I get the error in `(su hex)` which I guess will come from `(rash p.jon sab)`, as defined in`+su`.

ohAitch:
Sounds like the hex parser, also, is not doing so hot on the allocating
tons of memory front. Have the stir changes made it in yet, or will they
not help?

yosoyubik:
I initially tested this from the [PR branch](https://github.com/urbit/urbit/pull/3433) (`jb/stir-tail`) and then again from `release/next-vere`, so that one should (?) have all the changes to +stir, +roll and +reel.

PS: I'm making the binary with `make build && make install`; do i need to do something else to build it?

joemfb:
@yosoyubik, sorry for the delay.

It's worth noting that our parser combinators are not exactly fast in general, and some amount of overhead is just baked in due to the number of calls. They're also not really factored with performance in mind -- they work pretty well for parsing hoon particularly, and small inputs in general. But their performance degrades dramatically with the length of input, much more so than one would expect.

I've been researching these performance/memory-usage issues. There two pervasive problems that are implicated here: our atom-slice operation (`u3r_chop()`) is inefficient, and a few of the low-level atom jets perform unnecessary heap allocations. While I'm not sure if those are the only issues, they go a long way towards explaining what you've found here.

I'm working on both issues, but it may take me some time to have fixes ready (along with suitable test harnesses, which are really needed for this kind of low-level code). So, in the meantime, I've implemented direct hex en/decoding and jetted it, in #3576 and #3577. Please take look at those implementations and confirm that there behavior is suitable for your use-case here. I've taken care to preserve leading zeros where relevant (which I believe was your goal of the various left/right-shifts in your implementation here).

yosoyubik:
This is great, thanks @joemfb! I'll test it right asap and let you know.

yosoyubik:
The latest commits should resolve the main problems this PR has faced regarding the crashes coming out of the parsers. I've also added a change for the case where we are referencing the genesis block, since that entails the non-existence of a previous block.

I believe this turns this PR into a mergeable state.

Thanks to everyone involved!

yosoyubik:
Almost forgot:

Happy birthday Bitcoin Node JSON RPC API Pull Request!

joemfb:
Many happy returns!

yosoyubik:
Some instructions to run this, if someone is interested:

```
> |start %btc-node-hook-store
>=
> |start %btc-node-hook-hook
>=
> :btc-node-hook|command [%credentials 'http://BITCON_FULL_NODE_URL:8332/' 'RPC_API_USER' 'RPC_API_PASS']
>=
credentials updated...
> :btc-node-hook|action [%get-difficulty ~]
>=
[%get-difficulty n='19997335994446.11']
> :btc-node-hook|action [%get-mempool-info ~]
>=
[%get-mempool-info size=44.269 bytes=22.967.672 usage=66.402.688 max-mem-pool=100.000.000 mem-pool-min-fee='0.00006984' min-relay-tx-fee='0.00000001']
...
```

tacryt-socryp:
Already merged out of band in different form.

Issue #1707, opened 2019-09-06
Cannot re-update Arvo Network DNS Proxy if IP changes


matildepark:
I moved from AWS to GCP after a whole lot of fumbling about with EBS snapshots. 

I send a `:dns|request [my-new-ip]` and hit a crash:

```
~haddef-sigwen[dns]: request for DNS sent to ~zod
~haddef-sigwen[dns]: awaiting response from ~zod
tapp command failed in app/dns
  %crash
/~haddef-sigwen/home/0/app/dns/hoon:<[236 37].[236 39]>
%mismatch-binding-wat-do
/~haddef-sigwen/home/0/app/dns/hoon:<[236 7].[236 39]>
/~haddef-sigwen/home/0/app/dns/hoon:<[235 5].[249 19]>
/~haddef-sigwen/home/0/app/dns/hoon:<[233 5].[249 19]>
/~haddef-sigwen/home/0/app/dns/hoon:<[232 5].[249 19]>
/~haddef-sigwen/home/0/app/dns/hoon:<[230 3].[250 5]>
/~haddef-sigwen/home/0/app/dns/hoon:<[228 3].[250 5]>
/~haddef-sigwen/home/0/app/dns/hoon:<[226 3].[250 5]>
/~haddef-sigwen/home/0/app/dns/hoon:<[225 3].[250 5]>
/~haddef-sigwen/home/0/app/dns/hoon:<[224 3].[250 5]>
/~haddef-sigwen/home/0/lib/tapp/hoon:<[527 18].[527 63]>
/~haddef-sigwen/home/0/lib/tapp/hoon:<[523 9].[529 11]>
/~haddef-sigwen/home/0/lib/tapp/hoon:<[522 9].[529 11]>
/~haddef-sigwen/home/0/lib/tapp/hoon:<[521 9].[529 11]>
```

Advice? I suspected at first I needed to clear its state but it might be a different issue, now that I think about it.

vvisigoth:
cc @joemfb any insight into this?

joemfb:
I suspect this was a successful request after a failed request, and the initial failure wasn't handled correctly. A fix (for that state transition, at least) is in #1894

jtobin:
Should be resolved by the linked PR.

PR #1314, opened 2019-07-02
moves :dns scry for ames domains in :dns|auto generator

joemfb:
... working around the current inability to scry within a tapp (see `+mule`). This fixes a bug introduced by an oversight in urbit/arvo#1197

PR #1894, opened 2019-10-30
fixes :acme and :dns notifications

joemfb:
These apps used `:hall` as their notification channel, but `:hall` is deprecated and no longer running by default. This PR updates the apps (and lib/tapp components) to support and use `%flog` output cards for all notifications.

### DIFF
--- a/pkg/arvo/app/dns.hoon
+++ b/pkg/arvo/app/dns.hoon
@@ -277,7 +277,7 @@
     ::  XX details
     ~&  %dns-ip-request-failed
     %-  (slog u.error.sign)
-    (pure:m state)
+    (pure:m state(requested ~))
   ::  re-subscribe if (involuntarily) unsubscribed
   ::
       %quit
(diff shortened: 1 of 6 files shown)


pkg/arvo/app/dns.hoon
```diff
@@ -94,6 +94,16 @@
       loop(try +(try))
     ::
     ++  hall-app-message
+      |=  [app=term =cord =tang]
```
loganallenc: Why not just remove this or modify its name?

loganallenc:
Could you also remove the /-  hall imports in both acme and dns?

joemfb:
These changes are intended for OTA, so I've left all those types. In general, it can be tricky to remove entries from a `$%` without full state adapters. And adapting anything besides the persistent tapps is non-trivial, to say the least.

joemfb:
Updated this further to revert #1314 (no longer necessary) and fix #1707 (failed requests weren't cleared from the :dns state).

ohAitch:
Why were they going to hall in the first place, was the idea like to
centralize notifications cross-ship?

PR #1895, opened 2019-10-30
eth-watcher 2: azimuth-trackerer

Fang-:
As previously discussed with @philipcmonk, @vvisigoth.

Uses the logic existing in current azimuth-tracker to re-implement eth-watcher, which can now look at Ethereum nodes for _any_ events, as opposed to exclusively a subset of the Azimuth contract's events.

Azimuth-tracker itself gets transformed into a client of this new eth-watcher. It should have maintained its previous behavior, so no Jael changes are required.

For the "azimuth-tracker -> new azimuth-tracker" diff, see this PR's diff of its `/app` file.
For the "azimuth-tracker -> new eth-watcher" diff (or at least, the first version of it), see cda68b9.

I've given this some rudimentary local testing, it seems to keep up with the Ethereum mainnet just fine. But it seems wise to run this on a testnet for a little while.

Note that this does not have state adapters, and so is not fit for OTA.


jtobin:
> Note that this does not have state adapters, and so is not fit for OTA.

👌

pkg/arvo/lib/ethio.hoon
```diff
@@ -0,0 +1,105 @@
...
+  ::
+  ++  retry
+    |*  result=mold
+    |=  [crash-after=(unit @ud) computation=_*form:(async:stdio (unit result))]
+    =/  m  (async:stdio ,result)
+    =|  try=@ud
+    |^  |-  ^-  form:m
+        =*  loop  $
+        ?:  =(crash-after `try)
+          (async-fail:stdio %retry-too-many ~)
```
philipcmonk: I wonder whether this might be a bad idea.  We don't want to just give up, we actually want to keep trying forever, though probably after yielding to a new event.  I guess that's basically what your "set a timer regardless of what happens" thing does.

pkg/arvo/app/eth-watcher.hoon
```diff
@@ -1,571 +1,391 @@
...
-::
-++  eye
-  $:  config
-      latest-block=@ud
-      filter-id=@ud
-      poll-timer=(unit @da)
-      snapshot
-      sap=history
-  ==
+=>  |%
```
philipcmonk: Do you think these should go in lib/ethio?

Fang-: Probably, yeah! I moved them over. May or may not want to do `|_  =url` instead of taking it as a separate argument for each arm, but I think this is fine for now at least.

(There's also a secondary task here of "figure out what to do with eth stdlib". We have multiple places in zuse for Ethereum types and functions, and there's even some overlap... Worth thinking about how to do cleanup here, especially now that jael doesn't explicitly depend on Ethereum anymore.)

pkg/arvo/lib/ethio.hoon
Fang-: Yeah, I think this is "safe enough" for the reason you describe. Should keep an eye out for issues with this, but don't think it's blocking.

Fang-:
OTA could indeed theoretically be done, but I'll punt on that until we know more about the state & future of the current network.

Also this PR contains drum changes, and I have bad memories of updating _anything_ relating to hood OTA...

PR #1896, opened 2019-10-30
apps: remove %hall and %talk

loganallenc:
These applications are now deprecated, so if there are no objections I'd like to remove them. I'll leave the sur/hall.hoon file so that the %dns and %acme apps don't need the intensive state adapter updates that Joe mentioned in #1894.


Fang-:
_pours one out_
It was fun while it lasted. ):

galenwp:
We appreciate all your hard work, @Fang-

joemfb:
The presence of :hall's log/save/load files in this diff leads me to ask: does any of that functionality needs to be implemented in the new chat apps? In the same vein, do these apps support herb's import/export?

loganallenc:
The new %chat-store and %chat-hook should certainly have their own import/export written. Will this to my to-do.

matildepark:
Chat apps, like languages, evolve. This human, this good. When we have applications that aren’t chat apps on Urbit, then, only then, will we will have transcended ourselves.

Issue #1714, opened 2019-09-10
%~ doesn't always check type of sample


philipcmonk:
Consider:

```
=/  a
  |_  b=@
  ++  c  +(b)
  --
~(. a [3 4])
```

Expected behavior: nest-fail on the last line (we gave a cell as an argument where it expected an atom.

Actual behavior:  produces the core.  If you try to use the resulting core by pulling `c` from it, it nest-fails.

joemfb:
`%~` currently reduces to `%=` via `%*`. There's a comment saying it should be replaced with (a commented out) `%:`, which is currently broken. This area definitely needs some attention.

joemfb:
Potentially related, from #1899 

- `$:perk` fails (mull)
- `(perk ~)` succeeds
- `$:$:=>(perk .(+6 ~))` succeeds
- `$:(perk ~)` fails (nest)

And another example from @philipcmonk:

- `ut(+< 53)`
` ~(. ut 53)` and `.(+<.ut 53)` both succeed

PR #1899, opened 2019-10-31
dojo: Add tab completion

philipcmonk:
This is initial support for type-aware tab completion.  When you hit tab, it tries to complete the word you're in the middle of using a face or arm in the subject at that point in the code.  It also shows all possible matches and their associated types.  It's nearly instantaneous.  Notes:

- It advances to the longest common prefix, so if you hit tab on `ab` and the only possible results are `abcde` and `abcdz`, then it'll write `abcd` and print both out (with their types).

- If there are fewer than ten matches, it prints the type along with the face.  Printing types is too slow to use all the time, but with ten it's essentially instantaneous.

- The match closest in the subject to you (i.e. smallest axis number) is displayed lowest (closest to your focus).

Examples below, where `<TAB>` represents me hitting tab while my cursor is at that position (the line with the `<TAB>` is not preserved in the actual output).

```
~zod:dojo> eth<TAB>
-----
ethereum        #t/<11.qcl {<3.ltb 27.ipf 7.ecf 36.uek 92.bjk 247.ows 51.mvt 126.xjf 41.mac 1.ane $141> <21.yeb 27.ipf 7.ecf 36.uek 92.bjk 247.ows 51.mvt 126.xjf 41.mac 1.ane $141>}>
ethereum-types  #t/<3.ltb 27.ipf 7.ecf 36.uek 92.bjk 247.ows 51.mvt 126.xjf 41.mac 1.ane $141>
~zod:dojo> ethereum
~zod:dojo> |=  zong=@ud  z<TAB>
-----
zing  #t/<1.dqs {* <126.xjf 41.mac 1.ane $141>}>
zap   #t/<1.iot {tub/{p/{p/@ud q/@ud} q/""} <1.rff {daf/@t <247.ows 51.mvt 126.xjf 41.mac 1.ane $141>}>}>
zuse  #t/$309
zong  #t/@ud
~zod:dojo> |=  zong=@ud  zo<TAB>
-----
zong  #t/@ud
~zod:dojo> |=  zong=@ud  zong
~zod:dojo> <TAB>
hoon-version
trel
quip
pole
unit
qual
lone
... about 600 more lines ...
unity
html
zuse
eny
now
our
~zod:dojo>
```

Functionally, this is in a state where I'd be comfortable shipping it.  It doesn't interfere with anything if you don't press tab, and it's perfectly OTA-able.  I do think its output is a little verbose, but that can be tuned over time as people try it and determine what feels good in practice.

Additional notes:

- There are plenty of similar systems for other languages, but my most direct inspiration is Idris's editor tools.  This is implemented for the dojo, but I actually want it in my editor, which is why the meat is all defined in a library.  I've only tested on dojo one-liners, so I don't know the performance on large blocks of code.

- The default type printer isn't great for this use case.  In particular,
	- Cores should not print anything about their context
  - The `#t/` should go away
  - If it looks like a gate, we should print its return value
  - Maybe special handling for molds, but if the above is done, then for example `bone` is  `* -> @ud`.

- The worst part about our wing ordering is that it really screws up tab completion.  You want to do `point.owner-address` instead of `owner-address.point` because that lets you type `point.ow<TAB>`.  I weakly prefer reading it how we do it now, but it's really not great.  You could do an (dojo-specific?) alternate syntax of `point;owner-address`; this is a simple transformation.

- Regardless of the above, this should handle the case where we're in the middle of defining a wing; it doesn't right now.

- When a variable is shadowed, we show both of them.  We should probably show the shadowed one with a `^`.

- We probably shouldn't print out hundreds of results.  Maybe just the closest 50 with ellipses.

- This gets you any face in your subject, regardless of whether its type is reasonable.  We could limit that some by copying the `gol` logic in mint, so that if the pseudo-backward-inference engine happens to know what type it should be, you can filter the tab results according to if they nest in that type.  This would be "strongly type-aware".


matildepark:
Oh, this is pretty cool. I’d be interested in knowing how to get parity on this in the web dojo (how does this interact with sole?)

philipcmonk:
Three things will need to happen:

- Translate tab in the browser to the new `%tab` entry of `sole-action`.
- Give the output as a real effect instead of `~&`.
- Make sure I haven't abused the OT too much by including this out-of-band signal.  I *think* it'll work fine.

philipcmonk:
@yosoyubik made a nice video of how it looks in practice: https://asciinema.org/a/9A26r2pEnep943SWQTTJqrDW4

joemfb:
This is awesome! A couple notes:

- I think the list of suggestions should be a new `+sole-effect`. Drum could render it as it is now (using %dill effects), but smarter terminals or future web-dojo could overlay a pane of some kind. (That effect should support styles, too.)

- It seems like we could workaround the reversed wing and `=<` syntax by just searching differently if the cursor is on a `.` or `:`. 

- Refining the type printing and generally making this more type-aware might be a good opportunity to improve the xray library (although that might be a lot slower).

philipcmonk:
> I think the list of suggestions should be a new +sole-effect.

Agree, I think that'll be especially useful for webdojo.

> It seems like we could workaround the reversed wing and `=<` syntax by just searching differently if the cursor is on a `.` or `:`.

FWIW, `:` works "correctly", though of course it only makes sense if you've already typed the second part.  Maybe I can get used to habitually typing in reverse order and using the meta keys to jump around.

pkg/arvo/sur/sole.hoon
```diff
@@ -6,8 +6,9 @@
 ++  sole-action                                         ::  sole to app
   $%  ::  {$abo ~}                                      ::  reset interaction
       {$det sole-change}                                ::  command line edit
-      {$ret ~}                                         ::  submit and clear
-      {$clr ~}                                         ::  exit context
+      {$ret ~}                                          ::  submit and clear
+      {$clr ~}                                          ::  exit context
+      {$tab pos=@ud}                                    ::  tab complete
```
joemfb: Nit, but this mixes old and new mold syntax.

pkg/arvo/lib/auto.hoon
```diff
@@ -0,0 +1,238 @@
...
+  |=  [sut=type gen=hoon]
+  %+  bind  (find-type-mule sut gen)
+  |=  [id=term typ=type]
+  (search-prefix id (get-identifiers typ))
+::
+++  tab-list
+  |=  [sut=type code=cord]
+  %+  bind  (find-type sut (scan (replace-hax (trip code)) vest))
+  |=  [id=term typ=type]
+  (search-prefix id (get-identifiers typ))
```
joemfb: Could you add some comments detailing these different entrypoints and their basic strategies (ie, `#` vs `magic-spoon`).

ericfode:
For wings, could we let tab completion fill in the beginning of the line as well as the end?

pkg/arvo/sur/sole.hoon
philipcmonk: fixed

philipcmonk:
Made some improvements.  This addresses the server side of making this work with web dojo, so web dojo should be able to just send a sole-action of `[%tab pos=5]` to say "user asked for autocomplete at position 5".

Commit messages:

- dojo: add a better function printer

It's useful to know what a function takes and produces, so this changes the autocomplete type prettyprinter to emphasize those.  This also gives a nice syntax for molds.  Examples:

```
-----
add  {a/@ b/@} -> @
~zod:dojo> add
-----
term  * -> @tas
~zod:dojo> term
-----
sign-transaction  {tx/{nonce/@ud gas-price/@ud gas/@ud to/@ux value/@ud data/@ux chain-id/@ux} pk/@} -> @ux
~zod:dojo> sign-transaction🔑ethereum
-----
wind  {a/(* -> *) b/(* -> *)} -> * -> ?({$give p/*} {$pass p// q/*} {$slip p/*})
~zod:dojo> wind
```

- dojo, drum: give tab completion as true output

This stops slogging the tab completion and intead adds a +sole-effect
for tab completion output.  This is morally correct, and it lets dojo
clients show tab completions how they want.  For example, web dojo could
implement this as a drop-down box.

Another advantage is that this puts the rendering logic in drum, which
knows the width of the terminal.  Thus, we can make sure each match
takes no more than one line by truncating with ellipses.  If there's
only one match and it's already fully typed, then we display the whole
type.

- dojo, auto: move insert-magic logic to lib/auto

This changes the entry-points in lib/auto so that clients never have to
handle magic-spoon.  You can specify either a tape of code with a
position index or a preparsed hoon (presumably you ran +insert-magic
before parsing).

pkg/arvo/sur/sole.hoon
```diff
@@ -20,21 +21,22 @@
...
       {$det sole-change}                                ::  edit command
       {$err p/@ud}                                      ::  error point
       {$klr p/styx}                                     ::  styled text line
       {$mor p/(list sole-effect)}                       ::  multiple effects
-      {$nex ~}                                         ::  save clear command
+      {$nex ~}                                          ::  save clear command
       {$pro sole-prompt}                                ::  set prompt
       {$sag p/path q/*}                                 ::  save to jamfile
       {$sav p/path q/@}                                 ::  save to file
+      {$tab p/(list {=term =type})}                     ::  tab-complete list
```
joemfb: I'm afraid I must object. It's not appropriate to send types in effects -- they might be massive and unsuitable for serialization. And I think this effect should be suitable for general purpose completions (chat might autocomplete audiences, for instance). It seems like it should be a list of something like `[term cord]`, `[cord cord]`, or `[cord tank]`. The latter can still be rendered at the correct width by drum.

pkg/arvo/lib/auto.hoon
philipcmonk: Refactored and added some comments, lmk if it makes sense.

pkg/arvo/sur/sole.hoon
philipcmonk: Ah, yes.  I figured for serialization you could just have the mark convert it to a tank.  But the point of making it general-purpose for things other than types is definitely correct.  I'll change it to `[cord tank]`.

pkg/arvo/lib/easy-print.hoon
```diff
@@ -0,0 +1,481 @@
...
+++  dole
+  ^-  cape
+  =+  gil=*(set type)
+  =+  dex=[p=*(map type @) q=*(map @ wine)]
+  =<  [q.p q]
+  |-  ^-  {p/{p/(map type @) q/(map @ wine)} q/wine}
+  =-  [p.tez (doge q.p.tez q.tez)]
+  ^=  tez
+  ^-  {p/{p/(map type @) q/(map @ wine)} q/wine}
+  ?:  (~(meet ut sut) -:!>(*type))
```
joemfb: I just tried this branch out locally. Hitting tab on an empty buffer (or immediately before `:title`, for instance) results in a crash on this line.

philipcmonk: Whoops, that was working before.  I'm guessing this broke because now we render everything to tanks instead of only if fewer than 10; there's probably some type that breaks this.  Will investigate.

philipcmonk: Hmm, any idea why putting the meet call in +mule take it from "super fast" to "painfully slow"?

philipcmonk: FWIW, it's crashing for the same reason that `*pfix` crashes (don't know why that crashes).

joemfb: I would guess the +mule turns a cache hit into a miss.

pkg/arvo/lib/auto.hoon
```diff
@@ -0,0 +1,265 @@
...
+      %void      ~
+      [%atom *]  ~
+      [%cell *]
+    %+  weld
+      $(ty p.ty)
+    $(ty q.ty)
+  ::
+      [%core *]
+    %-  weld
+    :_  $(ty p.ty)
```
joemfb: Do you think you should limit the results by checking the variance here?

pkg/arvo/lib/easy-print.hoon
philipcmonk: It's actually quite strange to me that +pfix compiles at all, since it takes a noun and then passes that into +comp, which is expecting something that has a `q.vex`.  This seems like a bug in `+mint:et`, but I'm not sure I want to track that down right now.

I changed `+pfix` to have an appropriate sample type and the crash went away, so I'll check if anything else has the same problem.  If there's just a couple that I can fix, I'll do that and leave off the +mule.

pkg/arvo/lib/auto.hoon
philipcmonk: Good catch, I think you're right.

pkg/arvo/lib/easy-print.hoon
joemfb: +sfix is defined in the same way. This seems related to #1714

pkg/arvo/lib/auto.hoon
joemfb: Beautiful:

```
-----
de  {a/@ b/@} -> u(@)
en  {a/@ b/@} -> @
nu  <3?kyj 6.zxs 33.rft 27.ipf 7.ecf 36.uek 92.bjk 247.ows 51.mvt 126.xjf 41.mac 1.ane>
ex  <4?zto 6.zxs 33.rft 27.ipf 7.ecf 36.uek 92.bjk 247.ows 51.mvt 126.xjf 41.mac 1.ane>
dy  {a/@ b/@} -> @
as  <4?erk 6.zxs 33.rft 27.ipf 7.ecf 36.uek 92.bjk 247.ows 51.mvt 126.xjf 41.mac 1.ane>
~zod:dojo> :*acru:ames
```

pkg/arvo/lib/easy-print.hoon
philipcmonk: Github's UI is fritzing out on me, not sure if this message went through:

It's actually quite strange to me that +pfix compiles at all, since it takes a noun and then passes that into +comp, which is expecting something that has a q.vex. This seems like a bug in +mint:et, but I'm not sure I want to track that down right now.

I changed +pfix to have an appropriate sample type and the crash went away, so I'll check if anything else has the same problem. If there's just a couple that I can fix, I'll do that and leave off the +mule.

pkg/arvo/lib/auto.hoon
joemfb: `[tab pos=0]`

pkg/arvo/lib/easy-print.hoon
philipcmonk: I can fix +sfix, I haven't figured out +perk yet.  I checked all the initial letters and they work except `p` and `s`.

pkg/arvo/lib/auto.hoon
philipcmonk: now just give those samples meaningful names and we can get rid of the docs altogether!

philipcmonk: now just give those samples meaningful names and we can get rid of the docs altogether!

pkg/arvo/lib/easy-print.hoon
joemfb: The issue with +perk is strange: `$:perk` fails (mull), while `(perk ~)` succeeds. But `$:$:=>(perk .(+6 ~))` succeeds, while `$:(perk ~)` fails (nest).

philipcmonk: Going to keep looking at why `*perk` doesn't compile, but it occurs to me that maybe wet gates shouldn't show their output type, since it may change depending on the input.  In any case, wet gates should probably be written differently (maybe replace `->` with `~>`).

ohAitch:
For dry gates also, could we have a separator that isn't synonymous with
+5? ~> is also a rune but at least it's not common.

philipcmonk:
Sure, I'm open to anything that your eye scans quickly as a function.  `=>`, `~>`, and `-->` come to mind.  We could even do "⟶" to flush out the weak terminals.

pkg/arvo/lib/easy-print.hoon
philipcmonk: Eh... fixed!

https://github.com/urbit/urbit/blob/0014d1cf2b9a7b34a346801d91057fa50eccbab2/pkg/arvo/app/dojo.hoon#L1147-L1148

This feels like a compiler bug and I've spent a few hours trying to track it down without any success, so I'm going to declare it out of scope.

philipcmonk:
Okay, I've fixed various bugs and added a few more features.  I think I'm going to declare it good enough, feature-wise.  Lots more could and should be done, but I've got to get back to my day job.

Would love for someone to pick it up and add features, make it work for other dojo operations (like `+<TAB>` to list all your generators), and for chat-cli too.  Still want vim integration too.

joemfb:
@jtobin, this should be suitable for OTA. The conflict with :talk is noteworthy (deleted on master, along with :hall). We can OTA those deletions, but I think they'll have to be done manually on ~zod (`|rm %/app/talk/hoon` &c). Or we can not OTA them, in which case the change to :talk from this PR should be included in the update (even if it's not running, we don't want to ship changes that would break it if started).

PR #1901, opened 2019-10-31
Added invite app and invite functionality to %chat

loganallenc:
This additional functionality finally brings us up to parity with %hall. We now support sending invites to new chatrooms and joining by accepting the invite.


pkg/arvo/sur/invite-store.hoon
```diff
@@ -0,0 +1,34 @@
...
+::
++$  invite
+  $:  =path
+      =ship
+      recipient=ship
+      app=@tas
+      text=cord
+  ==
+::
++$  invitatory  (map serial invite)  ::  containing or conveying an invitation
```
Fang-: Why do we map these by serial? I'd expect to map them by ship. (ie, do uniqueness based on "ship and context" of invite, not "uid" of invite. If I invite you multiple times to `~paldev/secret`, then *maybe* you should get multiple notifications (on-receive), but there certainly shouldn't be duplicate invitations stored/managed in state.

pkg/arvo/app/invite-store.hoon
```diff
@@ -0,0 +1,186 @@
...
+  ?>  (team:title our.bol src.bol)
+  (poke-invite-action (json-to-action json))
+::
+++  poke-invite-action
+  |=  action=invite-action
+  ^-  (quip move _this)
+  ?>  (team:title our.bol src.bol)
+  ?-  -.action
+      %create
+    (handle-create action)
```
Fang-: Probably just one-line all of these, ie `%create  (handle-create action`.

pkg/arvo/app/invite-store.hoon
```diff
@@ -0,0 +1,186 @@
...
+  ==
+::
++$  state-zero
+  $:  =invites
+  ==
+::
++$  diff
+  $%  [%invite-initial invites]
+      [%invite-update invite-update]
+  ==
```
Fang-: Why isn't this definition in the `/sur` file in its entirety?

(I'd personally also just add `%initial` to `invite-update`, instead of having an entirely separate diff tag for it.)

pkg/arvo/app/invite-store.hoon
```diff
@@ -0,0 +1,186 @@
...
+::
+++  peek-x-invite
+  |=  pax=path
+  ^-  (unit (unit [%noun (unit invite)]))
+  ?~  pax
+    ~
+  =/  invitatory=(unit invitatory)  (~(get by invites) t.pax)
+  ?~  invitatory
+    ~
+  =/  uid=serial  (slav %uv i.pax)
```
Fang-: So a `/invite` scry path takes the form of `/invite/some-uid/invite-path`? Feels dirty to not have that go broad->specific. As in, following the data structures, `/invite/invite-path/some-uid`.

pkg/arvo/app/invite-store.hoon
```diff
@@ -0,0 +1,186 @@
...
+  ::
+      %accept
+    (handle-accept action)
+  ::
+      %decline
+    (handle-decline action)
+  ::
+  ==
+::
+++  handle-create
```
Fang-: Why not auto-create these the first time an `%invite` happens on some path? I'm not sure what use an explicit action & diff for "we have a new invites collection, but there's nothing in it" serves.

pkg/arvo/sur/invite-store.hoon
```diff
@@ -0,0 +1,34 @@
...
+::
++$  invite-action
+  $%  invite-base
+      [%accept =path uid=serial]
+  ==
+::
++$  invite-update
+  $%  invite-base
+      [%invitatory =invitatory]
+      [%accepted =path uid=serial =invite]
```
Fang-: Why do we re-send the invite here?

pkg/arvo/app/invite-hook.hoon
```diff
@@ -0,0 +1,61 @@
...
+++  invite-poke
+  |=  [pax=path action=invite-action]
+  ^-  move
+  [ost.bol %poke pax [our.bol %invite-store] [%invite-action action]]
+::
+++  invitatory-scry
+  |=  pax=path
+  ^-  (unit invitatory)
+  =.  pax
+    ;:(weld /=invite-store/(scot %da now.bol)/invitatory pax /noun)
```
Fang-: For wide form, the irregular `:(weld` is more common.

pkg/arvo/sur/invite-store.hoon
loganallenc: The `invites` path is not semantically the same type of path as the one in an individual invite.

For instance, the %chat app only wants to see invites sent to /chat, so it creates an `invitatory` at the /chat path. Then within that invitatory, invites are unique, so they each have a serial identifier.

pkg/arvo/app/invite-store.hoon
loganallenc: Good idea.

pkg/arvo/app/invite-store.hoon
loganallenc: It's specific to this app.

pkg/arvo/app/invite-store.hoon
loganallenc: A uid is fixed length and a path is variable length, so it makes more sense to prefix with uid than to suffix with it.

pkg/arvo/sur/invite-store.hoon
loganallenc: Because accepting an invite deletes the invite, and apps may want to trigger some effect based off of the %accepted event. (As the %chat-hook does)

pkg/arvo/sur/invite-store.hoon
loganallenc: I should just put comments in the sur file explaining this.

pkg/arvo/app/invite-store.hoon
loganallenc: We don't want to auto-create because we don't want to have to clean up random invite paths. We want individual `invitatories` to be managed by specific apps which display interfaces / have affordances for cleaning out invites.

pkg/arvo/app/invite-hook.hoon
loganallenc: I prefer `;:` because it's explicit

pkg/arvo/app/invite-store.hoon
loganallenc: It would be difficult to manage some disparate number of paths that could have invites being sent to them.

pkg/arvo/app/invite-store.hoon
Fang-: Sure, this app is the only one _sending_ the diff, but any subscriber can _receive_ the diff, in case it will still want the type.

If it's intended do only send to other invite-store instances, then: 1) that breaks store rules, and 2) it's not currently happening, so might as well take it out.

pkg/arvo/app/invite-store.hoon
Fang-: I agree that makes sense from an _implementation_ standpoint, but would say that it doesn't from an _interface_ standpoint. It breaks convention of pretty much all path usage, both pre- and post-urbit.

pkg/arvo/sur/invite-store.hoon
Fang-: Shouldn't it also be sent along in a `%decline` then?

pkg/arvo/app/invite-store.hoon
loganallenc: Idk man, none of the other stores have that convention. I prefer pokes and diffs to be defined in the specific file.

pkg/arvo/app/invite-store.hoon
loganallenc: Most urbit paths follow the prefix convention or do weird things where they flop the paths.

pkg/arvo/sur/invite-store.hoon
loganallenc: No, at least not until there's some use case. %declines don't cause an action.

pkg/arvo/sur/invite-store.hoon
Fang-: Okay, that makes sense. Still feel pretty icky about using serial identifiers though. Have you thought about doing `[=ship resource=path]` as identifiers? ie, for `/chat`, doing `[~tacryt-socryp /dm-with-tacryt]` for (the one, auto-deduplicated) invite relating to that specific chat.

pkg/arvo/app/invite-store.hoon
Fang-: > We want individual `invitatories` to be managed by specific apps

For `invitatory`, is the distinction between "does not exist" and "does not contain any invites" meaningful?

I'm arguing that it isn't. Readers of the invite-store _shouldn't_ care about that distinction, and since _every single invite-sending application_ will just be doing "create this" before doing "add things to it", the store should just generalize that into itself.

If you're saying the distinction _is_ meaningful, then yes, manual creation is justified. If so, let me know your case for this.

(Related, adding to `invitatory` path that doesn't exist currently just silently no-ops. Seems a bit dirty. Noteworthy that auto-creating as discussed above would get rid of the no-op.)

pkg/arvo/app/invite-store.hoon
Fang-: The path flopping is ancient and largely abandoned. And I only recall seeing it in comments, not in implementations. Reverse order is still better than arbitrary order though.

pkg/arvo/app/invite-store.hoon
loganallenc: The distinction is meaningful, because if you send to an `invitatory` that doesn't exist, you should not store the invite.

pkg/arvo/sur/invite-store.hoon
loganallenc: I like the serial identifiers, personally. They're unique, meaningful to the client, and duplicate invites are rare / a non-issue.

pkg/arvo/app/invite-store.hoon
loganallenc: It isn't arbitrary. It's specified.

pkg/arvo/sur/invite-store.hoon
Fang-: > meaningful to the client

Wait, maybe I misunderstand, but `+$  serial  @uvH` suggests something like `0v6.bmjjf.t2l6p.h5s63.oop3a.5lc54`. How would they be meaningful?

loganallenc: Meaningfully different from each other insofar as individual invites are differentiated from each other via the unique identifier, and they're already context-specific due to the `(map path invitatory` pattern.

pkg/arvo/sur/invite-store.hoon
```diff
@@ -0,0 +1,34 @@
+|%
+++  serial  @uvH
+::
++$  invite
+  $:  =path
+      =ship
```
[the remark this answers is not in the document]
loganallenc: Good recommendation. I made it `ship app path`.

loganallenc:
Can I get a review @philipcmonk or @jtobin? If possible, I'd love to get this into the OTA on Monday.

philipcmonk:
I'll review it today

pkg/arvo/app/invite-store.hoon
joemfb: @Fang-  is right, pokes and diffs are public interfaces.

pkg/arvo/app/invite-store.hoon
joemfb: Scry paths are interfaces, and should be designed as such.

pkg/arvo/app/invite-store.hoon
loganallenc: Okay. Will change.

pkg/arvo/app/invite-store.hoon
loganallenc: Okay. Will change.

loganallenc: Have changed this now.

pkg/arvo/app/invite-store.hoon
loganallenc: Have changed this now.

pkg/arvo/app/invite-hook.hoon
ohAitch: Could you elaborate on how irregular form is not explicit?

loganallenc: Irregular form is nice when you need to save line space or when the expression is nested in such a way that fewer runes adds to readability. In this case, one fewer characters won’t affect the line break, but does make it easier to search in the Docs / etc if someone wants to know what it does. Thus, regular form is slightly more legible to a newcomer.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -38,10 +40,11 @@
...
   ^-  (quip move _this)
-  ?~  old
-    :_  this
-    [ost.bol %peer /permissions [our.bol %permission-store] /updates]~
-  [~ this(+<+ u.old)]
+  :_  ?~(old this this(+<+ u.old))
+  :~  (invite-poke [%create /chat])
+      [ost.bol %peer /invites [our.bol %invite-store] /invitatory/chat]
+      [ost.bol %peer /permissions [our.bol %permission-store] /updates]
+  ==
```
joemfb: Automatically subscribing on `+prep` like this could result in many duplicate subscriptions. The duct behind the bone (`ost.bol`) could be an ames packet, a key stroke, a %clay sync, a %behn timer ....

pkg/arvo/app/invite-hook.hoon
joemfb: This may all be true, but as a guideline it's not really compatible with hoon. There's tons of irregular syntax, it's used everywhere, and it simply must be learned. There may be some corners of the irregular syntax that can safely be ignored, but `:(foo a bc)` absolutely idiomatic.

pkg/arvo/app/invite-store.hoon
joemfb: That's true, but tautological. The question is: should it be meaningful? The answer depends on a) what invites are (for), and b) how clients/consumers are expected to interact with the invitiverse.

pkg/arvo/sur/invite-store.hoon
joemfb: Like the indirection between invite and invitatory, whether this is righteous seems to depend on what invites are (for), and how clients should use these components. If you've already been invited to something, should it be a state transition if you're invited to it again?

pkg/arvo/app/chat-hook.hoon
loganallenc: This is surprising news to me. I will have to look into this, as it may be an issue.

pkg/arvo/sur/invite-store.hoon
loganallenc: Likely yes, because you may have left the chatroom (specific case) or deleted the data associated with the subscription and canceled the subscription the invite had given you access to previously. Thus you’re being re-invited.

ohAitch:
Do invites expire, then?

loganallenc:
@ohAitch They do not expire within a given time period, but when they are either accepted or declined, they are deleted.

pkg/arvo/app/chat-hook.hoon
philipcmonk: This is probably true.  The easiest way to verify it would be to start a fakezod and fake ~bud, change the app on ~bud and |merge that change to ~zod, then change the file again locally on ~zod.  I believe this will give you three separate subscriptions (one for the |start keystroke, one for the ames packet, and one for the |commit keystroke).  It is possible that they could end up on the same bone (maybe they're all really on the |start duct because that's what kicks off the ford live build?), so I'd verify before changing.

loganallenc:
Thanks everyone!

Responsibilities are as follows:
- invite-store: acts as an independent data store for invite information. We assume invites will be used in different contextual ways for different applications, but grant the inviter the ability to send the invitee metadata about what app on the inviter’s ship should be subscribed to upon acceptance of the invite, and what path. The invite-store talks only to the local ship. It receives pokes, and pushes out diffs. That is all.
- invite-hook: The invite hook acts as a gateway for sending and receiving invites from foreign ships. It doesn’t use permissions at all, and just allows any ship to send us an invite so long as we have a corresponding invitatory. If we did not have an invitatory for the invite, then we wouldn’t have an app that could interpret the invite, so we shouldn’t store it.
- chat-hook: the additions here just establish an invitatory for chat and subscribe to it. When we receive an event from the invite-store telling us that we have accepted a chat, we promptly join the chatroom specified in the invite.

ohAitch:
I would like to second the suggestion to crash instead of NOP if
invite-hook receives an invalid invite

loganallenc:
@ohAitch I don’t have an opinion on that one, so I’ll just go ahead and make the change.

jtobin:
I haven't gone ahead and merged this yet due to the comments re: looking into the issue @joemfb mentioned, but rest assured I can OTA it later in the week.

pkg/arvo/app/chat-hook.hoon
loganallenc: Okay, the newest commit fixes this issue for the chat-hook. Thanks! @jtobin This should be ready to be OTA'ed.

jtobin: @loganallenc :+1: I'm going to make a hotfix today that will contain just enough changes to hopefully fix the current network problems, but will follow it up with another update containing these changes later in the week.

joemfb:
@loganallenc, this has been OTA'd but was not automatically started. Next time, if you need it to be, you'll have to explicitly stage that (in some other app already known to be running).

PR #1902, opened 2019-10-31
Pool group hook

Fang-:
Depends/builds on #1895. Targets that branch since that hasn't been merged into master yet.

Implements `/app/pool-group-hook` which, if you were invited by someone, creates and maintains an `/invite-peers` group that contains all ships invited by the same person as you. In other words, your siblings in the invite tree.

(CI failure because muh pills I think. cc @jtobin)

### DIFF
--- a/pkg/arvo/lib/ethio.hoon
+++ b/pkg/arvo/lib/ethio.hoon
@@ -102,4 +102,17 @@
     =,  dejs-soft:format
     (ot id+so error+(ot code+no message+so ~) ~)
   --
+::  +read-contract: calls a read function on a contract, produces result hex
+::
+++  read-contract
+  |=  [url=@t proto-read-request:rpc:ethereum]
+  =/  m  (async:stdio ,@t)
+  ;<  =json  bind:m
+    %^  request-rpc  url  id
+    :+  %eth-call
+      ^-  call:rpc:ethereum
+      [~ to ~ ~ ~ `tape`(encode-call:rpc:ethereum function arguments)]
+    [%label %latest]
+  ?.  ?=(%s -.json)  (async-fail:stdio %request-rpc-fail >json< ~)
+  (pure:m p.json)
 --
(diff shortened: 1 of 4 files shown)


loganallenc:
This is cool! It definitely fits to the convention around what %hooks should be used for. I don't understand the control flow of apps written in the %tapp convention or what half of these runes mean, so I can't review for anything other than that. Looks good!

Fang-:
For reference: [tapp aka async monad docs](https://urbit.org/docs/tutorials/hoon/async-monad/)

jtobin:
> CI failure because muh pills I think.

Yup -- 400e386 updated %zuse but not the pill.  I added one for you.

This would actually be better to squash into the commit that changed %zuse itself.  I can do this for you, but it will involve rebasing your branch, so gimme the ok first.

(Also: the commit that changed %zuse should be prefixed with `zuse:`, because that's what it's touching.)

Fang-:
Go ahead and rebase + rename. (:

jtobin:
« C'est fait ».

pkg/arvo/lib/ethio.hoon
```diff
@@ -102,4 +102,17 @@
...
+  =/  m  (async:stdio ,@t)
+  ;<  =json  bind:m
+    %^  request-rpc  url  id
+    :+  %eth-call
+      ^-  call:rpc:ethereum
+      [~ to ~ ~ ~ `tape`(encode-call:rpc:ethereum function arguments)]
+    [%label %latest]
+  ?.  ?=(%s -.json)  (async-fail:stdio %request-rpc-fail >json< ~)
+  (pure:m p.json)
 --
```
philipcmonk: Does this work fine without a newline?

Fang-: Works on my machine. But I'll add a newline to this in #1895 just in case.

jtobin:
It looks like `eth-watcher-2` is ready to go into master, so I'll start the merge train here..

PR #1910, opened 2019-11-04
Language Server

philipcmonk:
This PR pairs with the initial commit of hoon-language-server [here](https://github.com/urbit/hoon-language-server) and urbit/hoon.vim#18.

This adds a basic language server app.  It communicates over HTTP with hoon-language-server, which communicates over stdio with any editor plugin that speaks the Language Server Protocol.  Eventually we should make this app speak JSON RPC directly, so that if any proxy is needed it can be a dumb proxy (eg TCP to HTTP2).  I've included configuration information in the hoon.vim readme for vim.

It has three basic features:

- Autocomplete (using the same engine as #1899)
- Syntax checking
- Rune snippets

All three are displayed in this demo:

[asciicast](https://asciinema.org/a/vn1IF29vjgj4tlmjS7vSWtwHW)

Lots more should be done in terms of polish and new features, but this is minimally viable.  I've disabled autocomplete and syntax checking for files larger than 1000 lines because it's slow, but if you don't care about rune snippets being available quickly, then you may as well enable it.  Since they don't currently depend on any dynamic data, arguably rune snippets should be a static portable file, like TextMate snippets.  We may add dynamic snippets later (eg if we parse how many arguments a function has).

I hacked together something for VS code which proves that it works, but I don't know that ecosystem well enough to make a real plugin for it.  I followed [this tutorial](https://code.visualstudio.com/api/language-extensions/language-server-extension-guide) to modify [this code](https://github.com/microsoft/vscode-extension-samples/blob/master/lsp-sample/client/src/extension.ts) to refer to hoon-languge-server instead of server.js, changed the transport to `TransportKind.stdio`, and I added `runtime: "/usr/local/bin/python3` to the `run` and `debug` server options.


galenwp:
This is so, so cool

belisarius222:
I don't think I'll have time for a full review of this before my imminent vacation, which technically starts today.  I have two comments so far:
1)  +runes in /rune-snippets/hoon will call +malt to reconstruct its constants every time, which is probably not *that* slow but really should be moved to run once at linking time.
2)  I can't wait to use this!

pkg/arvo/lib/language-server/rune-snippet.hoon
```diff
@@ -0,0 +1,541 @@
...
+  =,  enjs:format
+  %-  pairs
+  :~  'label'^(tape rune)
+      'insertTextFormat'^(numb 2)
+      'insertText'^(tape text)
+  ==
+::
+++  runes
+  %-  malt
+  :~  :-  "|$"
```
joemfb: I'll echo @belisarius222's comment, and further note that constant-folding is trivial with `^~`:

```
> !=  (malt [1 2] [3 4] ~)
[8 [9 5.596 0 511] 9 2 10 [6 [7 [0 3] 1 1 2] [7 [0 3] 1 3 4] 7 [0 3] 1 0] 0 2]
> !=  ^~((malt [1 2] [3 4] ~))
[1 [1 2] 0 [3 4] 0 0]
```

pkg/arvo/app/dojo.hoon
```diff
@@ -1055,6 +1057,11 @@
     ::  ~&  [%his-clock ler.cal]
     ::  ~&  [%our-clock ven.say]
     =^  dat  say  (~(transceive sole say) cal)
+    ::  speedrun mode
+    ::  =;  res
+    ::    ?.  ?=(%del -.dat)
+    ::      res
+    ::    (he-tab:res +(p.dat))
```
joemfb: Is this just auto-tab on input? Any reason to include it?

pkg/arvo/lib/trie.hoon
```diff
@@ -0,0 +1,29 @@
+|%
```
joemfb: This doesn't appear to be used. Is it relevant to this PR?

pkg/arvo/lib/language-server/rune-snippet.hoon
joemfb: Bump for constant folding.

pkg/arvo/app/dojo.hoon
philipcmonk: yeah.  it's exhilarating, but impractical; removed.

pkg/arvo/lib/trie.hoon
philipcmonk: There were two times that I wanted a trie, so I wrote one.  One use case was an attempt at partial caching of the parsing, but the gains were only about 20% so I nixed it.  A trie would be the right data structure for filtering the identifiers in your subject according to the prefix you've typed, but I haven't used it because so far a list is fast enough.

As it's unused, I've removed it.  If we need a trie in the future, we can pull it out of git history.

pkg/arvo/lib/language-server/rune-snippet.hoon
philipcmonk: Done

pkg/arvo/app/dojo.hoon
joemfb: Something to remember, maybe it should be a configurable mode. For fun, if nothing else.

pkg/arvo/lib/trie.hoon
joemfb: Makes sense. The entire autocomplete flow seems ripe for optimization and caching. But it's impressively fast already.

PR #1914, opened 2019-11-04
gall: fix virtualization of agent compilation

joemfb:
This fixes the %gall crashes plaguing the network (reported in #1911), but fixing the virtualization of agent compilation. Instead of un-virtualized crashes in %gall, agent compilation failure will now (again) be caught (and reported as negative acknowledgements, where appropriate).

/cc @philipcmonk

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1390,14 +1390,11 @@
       |=  =term
       ^-  [(each vase tang) _ap-core]
       ::
-      =/  compiled
+      =/  virtual
         =/  =type  p.running-state.current-agent
         =/  =hoon  [%limb term]
-        (~(mint wa cache.current-agent) type hoon)
-      ::
-      =/  virtual
-        =/  trap  |.(compiled)
-        (mule trap)
+        %-  mule
+        |.  (~(mint wa cache.current-agent) type hoon)
       ::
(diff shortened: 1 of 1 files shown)


jtobin:
This one is rough.  I haven't yet understood why the two differ -- is it due to some effectful use of the caching compiler?

joemfb:
It has nothing to do with `+mint:wa`, and everything to do with `|.`. Before this change, the compilation is performed outside the `|.`, which is simply wrapped around the result (and then virtualized in `+mule`. After this change, the call to the compiler is deferred inside the `|.`, which is then evaluated virtually inside the `+mule`. Compare:

The dereference is virtualized:

```
> =>  a=(add 2 2)  !=  |.(a)
[[1 0 3] 0 1]
```

The addition is virtualized:

```
> =>  ..add  !=  |.((add 2 2))
[[1 8 [9 36 0 3] 9 2 10 [6 [7 [0 3] 1 2] 7 [0 3] 1 2] 0 2] 0 1]
```

philipcmonk:
This is a great reason why it's nice to set off runes from other identifiers.  You can always factor out the arguments to a function, but you can't guarantee that for all runes.  A trivial example is `?:(=(0 1) !! b)`.  If you factor that as `=/(t !! ?:(a t b))` it'll crash.

Another way to refactor this would have been to just move the `|.` inside `compiled`.

Issue #1898, opened 2019-10-31
Ford error when attempting to change the names of types in %chat


tacryt-socryp:
@belisarius222 Could you take a look at this? Joe has no idea what is causing it.

I started by booting a fakezod from master. I then copied in the updated files from my [branch](https://github.com/urbit/urbit/tree/la-chat-type-renames). I'm attempting to test "OTA" conditions where I simply rename the types within the chat app with no changes to the content.

Ford fails on committing the files, and I'm quite confused by it.

<img width="612" alt="Screen Shot 2019-10-30 at 5 33 55 PM" src="https://user-images.githubusercontent.com/1377557/67909197-99979080-fb3b-11e9-9067-51cb415c1f90.png">

matildepark:
Also ran into this today working on dojo stuff. All I did was move a line with a sur import. It worked on creating a new zod, but all other commits had no effect on the first ship after that error.

belisarius222:
Looking

belisarius222:
Looks like a Ford bug.  This line failed:
https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/ford.hoon#L5559

That means the build to parse a scaffold from lib/chat-json wasn't in the state during a rebuild of the app, but it should have been.  Not sure why it wasn't.  Part of the stack trace is missing here because +reduce has a `!.` around it.

I'll make a PR to make the `!.` narrower so we can get better forensics on this.  In the meantime, don't delete the piers that had this error.  We'll want to load the new Ford code onto them to try to reproduce.

belisarius222:
Hmm, I'm not sure this is the best way to debug it.  @loganallenc can you send me your pier?  I want to mess with it.

loganallenc:
Thanks for the fast response. I’ll get you a pier tomorrow.

joemfb:
#1915 includes not a fix, but a workaround. I think a real fix will require propagating %drip errors back to %ford.

matildepark:
I think Fusion was meant to specifically address stuff like this (or at least so that we don't need to goad anymore)? @belisarius222

belisarius222:
Yeah, I'd be inclined to close this unless someone has experienced something like it after ford-fusion.

—
~rovnys-ricfer
https://urbit.org

PR #1915, opened 2019-11-05
gall: adds %goad to force agent rebuilds

joemfb:
This PR follows #1914 in our efforts to fix the livenet. Specifically, it's a way to work around broken live builds in %ford, as reported in #1898.

%ford builds can get stuck if %ford crashes when processing the response from %clay -- the response is intermediated through %behn via a %drip task, which provides isolation but not (yet) error reporting. When this happens, %ford thinks it's waiting for %clay, and %clay thinks it's already responded. Both are technically correct.

To work around this state for %gall agents, we add a `%goad` task to %gall, which can cause it to goad %ford into restarting the build. The task can apply to one specific agent, or all that are running. %gall will cancel the current live build (as it may be stuck) and start a new one. Alternately, we can set `force` to `%.y` if the build is in an invalid state and can't be cleaned up by %ford. (This has been observed in testing, and is included for near-term durability. %ford should either a) never get into such a state, or b) be able to get out of it.)

The interface is straightforward:

```
> |goad-gall
> |goad-gall, =force &
> |goad-gall, =agent %hood
> |goad-gall, =force &, =agent %hood
```

The recommendation is simply to run `|goad-gall`, and only force or target specific apps as necessary. The `force` option will result in logically duplicate %ford builds (although one maybe blocked, in which case it's mostly irrelevant).

If one were so unlucky as to find `:hood` in this state (build blocked, so updates will not be applied to it), this command can *not* get you unstuck -- the command would be added in an update to :hood, which is precisely not an option. For such unlucky ones, this PR also includes a `:goad` app.

To use it, simply `|start %goad`. The equivalent of `|goad-gall` is then `:goad 1`. The equivalent of `|goad-gall, =force &` is `:goad %force`. No agent-specific option is provided.

I've tested this PR on (a local-only copy of) my planet; after the borked OTA, `:chat-cli` and `:dojo` were updated, while `:hood` was not. It restored full functionality. These changes also need to be tested on a ship in the opposite scenario (for instance, `:hood` updated, `:dojo` not) to confirm it works in either scenario.

### DIFF
--- a/pkg/arvo/sys/vane/dill.hoon
+++ b/pkg/arvo/sys/vane/dill.hoon
@@ -8,10 +8,11 @@
 --                                                      ::
 =>  |%                                                  ::  console protocol
 ++  axle                                                ::
-  $:  $1                                                ::
+  $:  $2                                                ::
       hey/(unit duct)                                   ::  default duct
       dug/(map duct axon)                               ::  conversations
       lit/?                                             ::  boot in lite mode
+      dog/_|                                            ::  auto-goad
       $=  hef                                           ::  other weights
       $:  a/(unit mass)                                 ::
           b/(unit mass)                                 ::
@@ -46,7 +47,10 @@
           $>(%wegh task:able:ames)                      ::
       ==                                                ::
       $:  %b                                            ::
-          $>(%wegh task:able:behn)                      ::
+          $>  $?  %wait                                 ::
+                  %wegh                                 ::
+              ==                                        ::
+          task:able:behn                                ::
       ==                                                ::
       $:  %c                                            ::
           $>  $?  %merg                                 ::  merge desks
@@ -76,6 +80,7 @@
       $:  %g                                            ::
           $>  $?  %conf                                 ::
                   %deal                                 ::
+                  %goad                                 ::
                   %wegh                                 ::
               ==                                        ::
           task:able:gall                                ::
@@ -100,7 +105,10 @@
               gift:able:ames                            ::
       ==  ==                                            ::
       $:  %b                                            ::
-          $%  $>(%mass gift:able:behn)                  ::
+          $%  $>  $?  %mass                             ::
+                      %wake                             ::
+                  ==                                    ::
+              gift:able:behn                            ::
               $>(%writ gift:able:clay)                  ::  XX %slip
       ==  ==                                            ::
       $:  %c                                            ::
@@ -149,13 +157,46 @@
         ^-  {(list move) axle}
         [(flop moz) all(dug (~(put by dug.all) hen +<+))]
       ::
+      ++  auto                                          ::  stage automation
+        ^+  .
+        ?.  dog.all  .
+        =.  dog.all  |
+        (pass /auto/one [%b %wait +(now)])
+      ::
+      ++  auto-wake                                     ::  resume automation
+        |=  [=wire error=(unit tang)]
+        ?+  wire
+          ?~  error
+            ~|(behn-bad-wake+wire !!)
+          (crud %wake u.error)
+        ::
+            [%auto %one ~]
+          ?~  error
+            ~&  %behn-goad
+            (pass / [%g %goad force=| ~])
+          ::  %goad crashed, wait again, then force
+          ::
+          ~&  %behn-goad-retry
+          %.  [/auto/two [%b %wait +(now)]]
+          pass:(crud %goad u.error)
+        ::
+            [%auto %two ~]
+          ?~  error
+            ~&  %behn-goad-again
+            (pass / [%g %goad force=& ~])
+          ::  %goad crashed again, bail out
+          ::
+          ~&  %behn-goad-fail
+          (crud %goad u.error)
+        ==
+      ::
       ++  call                                          ::  receive input
         |=  kyz/task:able
         ^+  +>
         ?+    -.kyz  ~&  [%strange-kiss -.kyz]  +>
           $flow  +>
           $harm  +>
-          $hail  (send %hey ~)
+          $hail  auto:(send %hey ~)
           $belt  (send `dill-belt`p.kyz)
           $text  (from %out (tuba p.kyz))
           $crud  ::  (send `dill-belt`[%cru p.kyz q.kyz])
@@ -293,7 +334,7 @@
             $c  '6'
             $w  '7'
             ~  '9'
-          ==
+           ==
         --
       ::
       ++  heft
@@ -371,7 +412,7 @@
         (deal / [%pump ~])
       ::
       ++  take                                          ::  receive
-        |=  sih/sign
+        |=  {tea/wire sih/sign}
         ^+  +>
         ?-    sih
             {?($a $b $c $e $f $g $i $j) $mass *}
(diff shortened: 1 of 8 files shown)


joemfb:
If there's time, I'll try to automate the application if this fix as part of the update process.

jtobin:
Nice -- I'll try borking/unborking a fakezod with this.

philipcmonk:
Automation would be good for people like me who can't use the console (hood
updated correctly but dojo did not) but eg landscape works. I think hood
will upgrade again if it receives new files. Another option is to poke the
app via http or lens.

joemfb:
As @philipcmonk notes, my instructions won't work if your :dojo is broken. In that case:

```
herb ./your-ship -p hood -d '+hood/start %goad'
herb ./your-ship -p goad -d '1'
herb ./your-ship -p goad -d '%force'
```

joemfb:
@philipcmonk, I've added %goad automation in %dill. PTAL.

pkg/arvo/sys/vane/dill.hoon
```diff
@@ -149,14 +157,44 @@
...
       ++  call                                          ::  receive input
         |=  kyz/task:able
         ^+  +>
         ?+    -.kyz  ~&  [%strange-kiss -.kyz]  +>
           $flow  +>
           $harm  +>
-          $hail  (send %hey ~)
-          $belt  (send `dill-belt`p.kyz)
+          $hail  auto:(send %hey ~)
+          $belt  (send `dill-belt`p.kyz)
```
philipcmonk: Trailing whitespace

joemfb: fixed

joemfb:
Rebased onto master and updated pill.

joemfb:
I should note that the automation is tied to the `%hail` event -- after applying the changes, you need to shutdown and restart. (The alternative is to check the flag on every `+abet` or equivalent, which seems excessive ...)

pkg/arvo/sys/vane/dill.hoon
```diff
@@ -100,8 +105,11 @@
               gift:able:ames                            ::
       ==  ==                                            ::
       $:  %b                                            ::
-          $%  $>(%mass gift:able:behn)                  ::
-              $>(%writ gift:able:clay)                  ::  XX %slip
+          $%  $>  $?  %mass                             ::
+                      %wake                             ::
+                      %writ                             ::  XX %slip
+                  ==                                    ::
+              gift:able:behn                            ::
```
joemfb: This refactoring is wrong -- my bad.

pkg/arvo/sys/vane/dill.hoon
```diff
@@ -149,13 +157,43 @@
...
+            ~|(behn-bad-wake+wire !!)
+          (crud %wake u.error)
+        ::
+            [%auto %one ~]
+          ?~  error
+            ~&  %behn-goad
+            (pass / [%g %goad force=| ~])
+          ::  %goad crashed, wait again, then force
+          ::
+          (pass /auto/two [%b %wait +(now)])
```
philipcmonk: Would be nice to print u.error here.  It did get to behn-goad-again on my ship.

pkg/arvo/sys/vane/dill.hoon
joemfb: It still works fine in normal operation, but fails boot. Looks like `$>` needs to assert that the tag actually exists somehow.

pkg/arvo/sys/vane/dill.hoon
joemfb: That'll be a crash in ford due to invalid internal state (ie, `+got:by` on an anchor or sub-build). I figured it'd be confusing to print the crash since we're continuing, but it's a one-time thing so it's probably not a big deal. I'll add it.

pkg/arvo/sys/vane/dill.hoon
joemfb: I guess it does when given a single tag, but not when combined with `$?`. Not sure if this pattern is righteous ....

philipcmonk:
To add the files I used:

```
herb ./your-ship -p hood -d '+hood/mount %'
herb ./your-ship -p hood -d '+hood/commit %home'
```

I got a crash on in ford on line 1596, similar to before.  However, when I restarted my urbit, it printed `%behn-goad` and `%behn-goad-again`, then worked!  Now I can use my dojo again, and I have tab completion.

joemfb:
Awesome!

I'm fixing the aforementioned %writ type error on boot, and making sure we don't %goad and boot. Updates momentarily.

pkg/arvo/sys/vane/dill.hoon
philipcmonk: That makes sense.  Ideally you'd recognize that specific error and not print it, but it doesn't matter much.

joemfb:
Ok, I've made the requested fixes, confirmed that everything is correct on first boot, and that the OTA problems are (still) fixed on my ship when I sync in these changes. The dev cycle time for small changes to the kernel is brutal ...

PTAL once more.

ohAitch:
Sounds like what happens when you remove transactionality from things!

jtobin:
On Tue, Nov 05, 2019 at 03:05:24PM -0800, Joe Bryan wrote:
> The dev cycle time for small changes to the kernel is brutal ...

Is it ever.  I've become desensitized to it, but it's awful.

philipcmonk:
Anything that you can do in pH you should because the cycle is much smoother.  You can swap vanes in a pill without recompiling everything, and you can script the test such that it's one click to do everything, which makes everything less frustrating and faster.  pH tests look kind of ugly now, but they're much cleaner in the static gall branch.

This one would have been a little hard to do in pH though.  If we ever get around to hoisting an existing ship into aqua...

jtobin:
On Tue, Nov 05, 2019 at 11:35:03PM -0800, Philip Monk wrote:
> Anything that you can do in pH you should because the cycle is much
> smoother.  You can swap vanes in a pill without recompiling
> everything, and you can script the test such that it's one click to do
> everything, which makes everything less frustrating and faster.  pH
> tests look kind of ugly now, but they're much cleaner in the static
> gall branch.

+1.  I haven't been doing much sustained kernel hacking since the %gall
refactor (the bulk of which was done as early as April) but recall
:aqua/:ph being fantastic there.

I used it once since, when finishing up that refactor a couple of months
ago.  There was a slight transaction cost in that I had to re-learn how
everything worked, but it was well worth paying.

belisarius222:
`:goad 1` nest-fail's for me on latest master on a comet https://github.com/urbit/urbit/commit/ee0de1a65780aad4bfb8ec1a19e9727c9e01b7c9, and `|goad-gall` also nest-fails.  `:goad %force` succeeds, but I'm not sure it does anything.

Traceback, including +trouble output: https://pastebin.com/WG9c5LJi

The comet can talk to other ships, reload its kernel, and generally seems to be working ok.  Merging from ~marzod %kids worked when I used `=gem %that`.  Not sure what the problem is here.  Maybe it's a stale Zuse version, possibly due to the Ford reload bug: https://github.com/urbit/urbit/issues/2109

joemfb:
@belisarius222, it looks like it's `:goad %go` as of #2097.

belisarius222:
Ok, that ran successfully on my comet.

PR #1916, opened 2019-11-05
behn: print %drip errors in the correct order

joemfb:
As one does ...


joemfb:
Abstractly, I think putting the most specific trace last is more convenient. But we almost always do the opposite, so this one stood out.

PR #1917, opened 2019-11-05
vere: virtualize tank printing

joemfb:
I saw some crashes due to bad tanks while developing #1915. We probably should've been doing this all along ... I'll update with details if I can reproduce.


joemfb:
This is hanging in CI, and also appears to be much slower than before. Let's leave it open until I track down more details.

jtobin:
I'm going to dismiss the approval while this is sitting here (the 'approved' tag in the PR list keeps catching my eye).

joemfb:
Closing, this is not a priority. I've been unable to reproduce the crash in question.

PR #1920, opened 2019-11-05
Convert apps to static gall

ixv:
Making this draft in order to convert apps one by one and have them reviewed without making a PR for each one.


philipcmonk:
let's use commit messages of the form "mall: convert permission-store"

philipcmonk:
The conversion looks good

ixv:
> let's use commit messages of the form "mall: convert permission-store"

Sure thing.

Fang-:
Are these going to be _duplicated_, or _moved_? If the latter, we should do the removal of the old `/app` file in the same commit, so that git can identify file `mv`s correctly and show us the "actual" diffs.

philipcmonk:
They will be moved ultimately.  My intention has been to do the "mall -> gall" step in one go at the end, so I figured I'd delete everything then.  That's less important now that it compiles, though.

However, I don't believe git has a real idea of "move" other than "a file disappeared and another one showed up somewhere else that's similar", and it operates on snapshots rather than diffs.  Thus when they're deleted later, the diff against master will show them as moves.  To verify this I deleted app/dojo (converted a while ago) and you can see it registers here as a move: https://github.com/urbit/urbit/compare/philip/scratch?expand=1#diff-33e907d8fb6b8fde3bd76cae78305c65

However, I have no objection to deleting the app files in the same commit as adding the age file.

joemfb:
It's really hard (impossible?) to stage incrementally that way. But maybe we can fixup the history in a merge commit (or sequence of commits: move unchanged, then merge ported versions over thee top) when everything's ready to switch.

philipcmonk:
Some style improvements for stores: https://github.com/urbit/urbit/commit/dadd6474a90c63181fd82a86d07e6ab1e05d9395

Avoids defining `state` in two places and all shadowing.

ixv:
updated to new style

philipcmonk:
- [ ] acme
- [ ] aqua-ames
- [ ] aqua-behn
- [ ] aqua-dill
- [ ] aqua-eyre
- [ ] aqua
- [ ] azimuth-tracker
- [ ] chat-cli
- [ ] chat-hook
- [ ] chat-store
- [ ] chat-view
- [ ] claz
- [ ] clock
- [ ] cop
- [ ] dns-bind
- [ ] dns-collector
- [ ] dns
- [ ] dojo
- [ ] eth-manage
- [ ] eth-watcher
- [ ] example-tapp-fetch
- [ ] example-tapp-subscribe
- [ ] gaze
- [ ] group-hook
- [ ] group-store
- [ ] hood
- [ ] language-server
- [ ] launch
- [ ] lens
- [ ] modulo
- [ ] permission-group-hook
- [ ] permission-hook
- [ ] permission-store
- [ ] ph
- [ ] publish
- [ ] send-txs
- [ ] test
- [ ] time
- [ ] weather

philipcmonk:
@loganallenc @ixv Here's an example of an app converted without a separate core: https://github.com/urbit/urbit/blob/46d89dd/pkg/arvo/age/dns-collector.hoon.

Compare the original: https://github.com/urbit/urbit/blob/46d89dd/pkg/arvo/app/dns-collector.hoon

ohAitch:
This isn't making any attempt to preserve app subscription etc state right,
it's a full network breaching change?

philipcmonk:
That's correct. I could imagine pushing this ota, but it would be a lot of
work.

philipcmonk:
The arms have been renamed on philip/mall-real, so pull that in here.

philipcmonk:
I refactored the gift/sign/task types a little bit so they don't have extraneous options or data.  The main difference you'll see is that +on-agent now takes a +sign:agent:mall instead of a +gift:agent:mall.

philipcmonk:
I also added a lib/verb, so you can do:

```
/+  verb
%+  verb  &
<your-agent>
```

This will print everything that happens, and you can toggle it with `:app +verb`.  The `&` determines if it's on by default.

I also converted chat-cli.

ixv:
@loganallenc currently working on invite store, modulo and launch

loganallenc:
I can convert %chat-hook. Will start on that today or tomorrow.

philipcmonk:
For anything that is a server, I'm going to clean that up a little bit within the next day, so I can give you a clean example to work off (will be clock).

philipcmonk:
I've fixed the issue where scry crashes weren't virtualized.

philipcmonk:
FWIW, the merge conflicts with this and m/convert-apps are not very helpful because for a couple of files I get a "conflict" of the entire file.  This happened because m/convert-apps is based on earlier versions of ixv/convert-apps, so when changes were made and rebased into earlier commits, they formally conflict with other changes by @Fang-.

You can avoid this by following the rule that if anybody is working off your branch, you don't rebase.  This results in more merge commits, but they're meaningful merges.  They preserve the actual history of the files, which makes conflict resolution much easier.  A linear flow is not always the simplest, and `git log --graph --oneline` is very good for understanding what led to what.

Rebase is fine and good when nobody's building off it, of course.

ohAitch:
It's interesting to think what a clay-backed git would look like: obviously clay can only store concrete branch reflogs instead of mutable "history", so on top of that the dag of currently-canonical beaks probably gets stored in a file looking very similar to the `rebase -i` view. (Along with commit messages etc)
The obvious merge strategies are "walk the dag gitwise to find common ancestors ignoring clay" and "run a naïve clay merge ignoring all git metadata(except to update it at the end)", but I wonder if there are any hybrid strategies that would be able to avoid the conflict mess described.

PR #1924, opened 2019-11-06
build: use HEAD in sh/create-ropsten-pills

jtobin:
If we're not on a branch, e.g. if we've checked out an arbitrary revision, then 'git branch --show-current' sensibly won't give us any output, and thus the pills produced by sh/create-ropsten-pills will be misnamed.

Consuming HEAD with git-rev-parse has the same effect here, and should work in all cases of interest.

### DIFF
--- a/sh/create-ropsten-pills
+++ b/sh/create-ropsten-pills
@@ -2,7 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


jtobin:
(@Fang- I ping you because you're probably online, and can probably sanity check this.)

PR #1928, opened 2019-11-08
hoon: Extend gol check through cores

philipcmonk:
This extends `gol` "backward-inference" typechecking to thread through cores.  Recall that `gol` is used exclusively for receiving more specific error messages; these changes should have no effect on programs which already compile successfully.

Before, this would type-fail on the second `|%`.

```
!:
^+  ^?
    |%
    ++  foo  *@ud
    --
|%
++  foo
  ?:  =(1 1)
    2
  %foo
--
```

With these changes, it gives a mint-nice at `%foo`.  It will also give you explicit errors if you have the wrong number/names of arms, including which arms it expects.

This is becoming much more important with static gall, since it's the first time we've used core subtyping so extensively and in userspace.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -10013,13 +10013,13 @@
   ++  hemp
     ::  generate formula from foot
     ::
-    |=  [hud/poly gen/hoon]
+    |=  [hud/poly gol/type gen/hoon]
     ^-  nock
     ~+
     =+  %hemp-141
     ?-  hud
-      $dry  q:(mint %noun gen)
-      $wet  q:(mint(vet |) %noun gen)
+      $dry  q:(mint gol gen)
+      $wet  q:(mint(vet |) gol gen)
     ==
   ::
   ++  laze
@@ -10043,7 +10043,7 @@
       ^-  (unit noun)
       %+  bind  (~(get by tal) axe)
       |=  gen/hoon
-      %.  [hud gen]
+      %.  [hud %noun gen]
       hemp(sut (core sut [nym hud %gold] sut [[%lazy 1 ..^$] ~] dom))
     ::
     %-  ~(gas by *(map @ud hoon))
@@ -10167,8 +10167,11 @@
   ++  mine
     ::  mint all chapters and feet in a core
     ::
-    |=  [mel/vair nym/(unit term) hud/poly dom/(map term tome)]
+    |=  [gol/type mel/vair nym/(unit term) hud/poly dom/(map term tome)]
     ^-  (pair type nock)
+    |^
+    =/  log  (chapters-check (core-check gol))
+    =/  dog  (get-tomes log)
     =-  :_  [%1 dez]
         (core sut [nym hud mel] sut [[%full ~] dez] dom)
     ^=  dez
@@ -10178,10 +10181,12 @@
       ~
     =/  dov/?(~ ^)
       =/  dab/(map term hoon)  q.q.n.dom
+      =/  dag  (arms-check dab (get-arms dog p.n.dom))
       |-  ^-  ?(~ ^)
       ?:  ?=(~ dab)
         ~
-      =+  vad=(hemp hud q.n.dab)
+      =/  gog  (get-arm-type log dag p.n.dab)
+      =+  vad=(hemp hud gog q.n.dab)
       ?-    dab
         {* ~ ~}   vad
         {* ~ *}   [vad $(dab r.dab)]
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -10167,8 +10167,12 @@
   ++  mine
     ::  mint all chapters and feet in a core
     ::
-    |=  [mel/vair nym/(unit term) hud/poly dom/(map term tome)]
+    |=  [gol/type mel/vair nym/(unit term) hud/poly dom/(map term tome)]
     ^-  (pair type nock)
+    |^
+    =/  gol  (core-check gol)
+    =.  gol  (chapters-check gol)
```
joemfb: I don't like the shadowing here, seems likely to be confusing (in an already dense corner of the compiler). Maybe `gol/type` and `log/gol-type` or somesuch.

pkg/arvo/sys/hoon.hoon
```diff
@@ -10178,10 +10182,13 @@
       ~
     =/  dov/?(~ ^)
       =/  dab/(map term hoon)  q.q.n.dom
+      =/  dag  (get-arms dog p.n.dom)
+      =.  dag  (arms-check dab dag)
```
joemfb: Another option here (and above) would be to combine some of these calls:

```
=/  gol  
=/  log  (chapters-check (core-check gol))
...
=/  dag  (arms-check dab (get-arms dog p.n.dom))
```

I realize this nit-picky -- no strong feelings on my part ...

pkg/arvo/sys/hoon.hoon
```diff
@@ -10194,6 +10201,119 @@
...
+    ::
+    ::  all the possible types we could be expecting.
+    ::
+    +$  gol-type
+      $~  %noun
+      $@  %noun
+      $%  [%cell p=type q=type]
+          [%core p=type q=coil]
+          [%fork p=(set gol-type)]
+      ==
```
joemfb: Are we sure that there's never any need to declare molds in separate cores/chapters from their use-sites? I *think* the seminoun system made that pattern unnecessary, but I haven't verified that.

pilfer-pandex:
Joe, how does the seminoun system affect this? I've actually been long
somewhat confused about the way in which "type aliases" defined in a core
can "come into scope"; in more traditional systems this would require a
form of dependent typing. (Can I use any gate in a ^- to cast to its return
type?)

—
~pilfer-pandex

pkg/arvo/sys/hoon.hoon
philipcmonk: I don't understand, what about this affects that?

joemfb:
@pilfer-pandex, yes. `^-(a b)` is just `^+(*a b)`. `*a` is `^*(a)`, which, in the case of a gate amounts to pulling `$` without changing the sample:

```
> (ream '*a')
[%kttr p=[%like p=~[%a] q=~]]
> ~(open ap (ream '*a'))
[%ktsg p=[%tsld p=[%limb p=%$] q=[%wing p=~[%a]]]]
```

For example:

```
> ^-  trip  "foo"
"foo"
```

Seminouns come into play with `^~`, which propagates constants through compile-time nock evaluation:

```
> !=  (add 2 2)
[8 [9 36 0 1.023] 9 2 10 [6 [7 [0 3] 1 2] 7 [0 3] 1 2] 0 2]
> !=  ^~  (add 2 2)
[1 4]
```

pkg/arvo/sys/hoon.hoon
joemfb: You've defined `$gol-type` in the same core (and chapter) as the arms that use it. That used to be frowned upon. As far as I know it's not a problem anymore (but when it was, I didn't understand it well) ...

pilfer-pandex:
So if you don't use ^~, you don't use seminouns? The advanced types section
of the website docs seems to suggest otherwise, unless I'm missing
something?

—
~pilfer-pandex

joemfb:
There's a seminoun in the `$coil` of `%core` is used by `+musk` (the seminoun interpreter) for compile-time nock 9 evaluation. The only call-site is `+blow:ut`, ie, `^~`.

jtobin:
(Barring objection, I'll squash these commits together before merging.)

philipcmonk:
Yeah, that's fine. I made them on the mall branch other commits in between,
so I just cherry-picked them over.

pilfer-pandex:
Jared, I’d like to note for the record that I actually didn’t approve this PR.

pilfer-pandex:
I wanted to have a longer conversation about the balance of complexity and how it’s factored. I wasn’t expecting to reject, clearly, but I am somewhat frustrated that this was merged without my green light.

philipcmonk:
I did expect for this to wait on @pilfer-pandex, and I'm still happy to
have that conversation and make changes as necessary.

I think we've abused the reviewer list in the past to also mean "take a
look", just tagging as "cc" is probably better for. Maybe we should say
that if we tag multiple reviewer, we expect a review from all of them?
OTOH, sometimes we do want to wait until like four people have seen it, but
they're not necessarily in a position to approve it. So maybe something
like this policy:

- If you don't want a PR to block on someone, don't request a review from
them. You may "cc" tag them.
- If anyone has requested changes, then it shouldn't be merged.
- If it's waiting on a review, it shouldn't be merged.
- It must have at least one approval to be merged.
- If you are a reviewer and you don't want to approve for whatever reason,
but you also don't want to block, you can review as "comment".

Exceptions can of course be made when there's good reason.

jtobin:
On Sat, Nov 09, 2019 at 09:50:53AM -0800, pilfer-pandex wrote:
> Jared, I’d like to note for the record that I actually didn’t approve
> this PR.

Apologies, my bad.

Typically I'll merge things if there's a single credible-looking review
on them.  I do this because often several people are tagged as
reviewers, and one or more of them don't show up, so, historically, it
has simply been more expedient to just merge as soon as something looks
to be reasonably vetted.

However, this is the second time in recent memory that someone has felt
"skipped," so I don't think the "merge if there's a single review"
policy is going to be productive to keep going forward.

I think I should probably move to a "require an ACK from every tagged
reviewer" model immediately.  I don't think there's any particular rush
to get things merged in, and this will prevent this sort of thing from
happening in the future.

jtobin:
On Sat, Nov 09, 2019 at 11:40:16AM -0800, Philip Monk wrote:
> I think we've abused the reviewer list in the past to also mean "take
> a look", just tagging as "cc" is probably better for. Maybe we should
> say that if we tag multiple reviewer, we expect a review from all of
> them?

Yeah, precisely.  Going forward, I'll generally treat the formal set of
GitHub reviewers as the canonical list of people who should approve PRs,
and will hassle reviewers who are late to the party, or warn that I
might merge without their review if they don't show up soon.

I've looked into GitHub's support for a CODEOWNERS file in the past; it
might be time to put together one of those as well.

PR #1929, opened 2019-11-09
eyre: subscribe to apps for responses

philipcmonk:
This removes the %http-response special case from gall.  In its place,
we implement a subscription regime with the following steps:

- Agent sends %connect to Eyre
- Eyre pokes agent with %handle-http-response, including unique eyre-id
- Agent passes %start-watching to Eyre with eyre-id and unique app-id
- Eyre subscribes to agent on /http-response/app-id
- Agent produces a %http-response-header fact followed by 0 or more
  %http-response-data facts and possibly a %http-response-cancel fact
- Agent produces a %kick to close the subscription, which Eyre
  interprets as completion of the message.

This works when there is data.  There is currently a bug where if the
response has no data in total (as in the case of a naked 404), no
response will be sent.

This also includes lib/http-handler, which implements a convenient
interface for agents that want to respond immediately with all the data.
This lets them avoid carrying extra state to keep track of pending
requests.

This should really have access to your state and the ability to change
it.  Perhaps a more minimalist design would be better: just keep track
of the requests, then hand it off to +on-watch when eyre is ready to
receive responses.  It's not clear how to pass in the request data in
+on-watch.

I'm most interested in review of the Eyre flow instead of lib/http-handler,
but ideas for how to make that more convenient are welcome.


joemfb:
That was really fast! I've skimmed these changes, and will give them a closer look soon.

What jumps out at me is the order of operations: I think %eyre should subscribe to the agent first, then %poke %handle-http-request, a la dojo and sole. Those apps currently use the subscription bone as a session, which obviously won't work here. But subscribing first allows the agent to give the http-response as a synchronous response to the request (kicking the subscription immediately), and this is the most common case.

The immediate question that occurs to me: on path should %eyre subscribe? It should be unique to the request, of course: /http-request/{sequence-number} seems simplest. %eyre already has server/connection/request sequence numbers in the duct, but they're not very useful there. They could move into the event card, or eyre could just count itself -- globally, per agent, or per binding (in which case the binding would also have to be accounted for in the path). Since agents don't have access to bones, we need a new pattern to associate pokes with active subscriptions. %handle-http-request could include the path, which would unblock this use case. All pokes could gain a wire, which would be really disruptive. Or maybe something like (unit path) could go in the bowl to signify that an agent-action corresponds to some active subscription (i'm not sure about the implications wrt subscription cardinality).

It'd be really nice to a) support synchronous production of http responses and b) establish a general pattern for these kinds of subscription-related actions, usable by the console subsystem at least. But maybe I'm being greedy ... Thoughts?

philipcmonk:
In this case, you could technically do the subscription before the poke since eyre is the only consumer and can guarantee that it doesn't reuse ids.  Maybe it's worth taking advantage of this, but it's not true of the general pattern, like dojo needs to deal with.

In general, the only person that can assign a unique path to a subscriber is the publisher.  So, you need to ask for an id before you subscribe to it (putting the id in the path).  Then, you pass the id along with every related poke.  This also has the nice property that you can pass the id to other people if they want to participate, which you can't do with duct/bone/wire-based systems.

One shortcut would be @loganallenc's proposal for adding a cage to `%watch` tasks, as a means of giving information about the subscription you want set up.  This would allow synchronous responses to the request.

However, I think proper library functions can make the experience seem synchronous.  For example, lib/http-handler should probably not take a function, so instead of storing `(map id response)` it would store `(map id request)`, and then when it notices the subscription start, it calls `+on-poke` with the request, which can now respond immediately.  This is probably cleaner than how I have it now, because it would simulate the flow you describe, state access comes for free, and you don't have to awkwardly define a function outside of the agent proper.  You can use virtually the same function for the console; you might even be able to share code.

In general, I like implementing this sort of nicety inside a library function because it lets the fundamental model stay very clean (it's all built on vanilla pubsub+poke).

ohAitch:
*concern about removing bones intensifies*

Maybe we could have some sort of hybrid system, where bone is ?(~ @),
~ being the "multicast" bone that updates everyone on a path / DWIMs poke
responses by duct, and @ (implied nonzero) is a gall-generated subscriber
id? Maybe not even from the whole duct, and rather from an optional
discriminator field in / new flavor of %peer.

It just seems notable that 6 months ago this just worked, because it's the
precise thing ducts are for architecturally, and now "establish 1:1
connection to app" functionality has been removed from gall(and sounds like
eyre). Like, it is in fact more work in eyre to route things if you're not
allowed to overload wires to mean individual events, and should maybe be
pushed inside the cards somewhere, but don't make *userspace* have to
syn-synack-ack things!

(Etc insert long-standing musings about how maybe we should have a
[key=duct ret=duct] to distinguish "UI rectangle somewhere ultimately
responsible for this event" from "arbitrary stack frames with transient
information not unique to the former + destination", aka
"session"/"keystroke" - traditionally it's meant the former, this has some
problematic constraints that have caused some unilateral switchover the
latter, but we kinda need both)

ohAitch:
> In general, the only person that can assign a unique path to a subscriber
is the publisher.

This in particular doesn't sit well with me, it's like auto-incrementing
database fields vs UUIDs. (Relatedly, a subscriber can assign itself a
/web/(shaf %app eny) perfectly fine). That particular implementation might
be slow and/or cause less pretty debug output however, which is why my
"hang on, this used to work before you removed the thing that was in active
use for it" focus.

pkg/arvo/age/clock.hoon
```diff
@@ -0,0 +1,64 @@
...
+    =/  back-path  (flop site.request-line)
+    ?~  back-path
+      ''
+    i.back-path
+  ::
+  ?~  back-path
+    not-found:gen
+  ?:  =(name 'tile')
+    (js-response:gen tile-js)
+  not-found:gen
```
ohAitch:
```
  ?+  (flop site.request-line)  not-found:gen
    [%tile *]  (js-response:gen tile-js)
  ==
```

also, am I right to understand `server` is implicitly exporting a three-letter named _core_ with a bunch of helper functions? that was kind of hard to follow, I realize we're moving away from three-letter names for local variables but imo we shouldn't move towards them for other things

also also this seems like kind of a lot of logic for "on `GET /~clock/js/tile.js` serve `/app/clock/js/tile.js`"(or well also on `POST` or `HEAD` or `OPTIONS` :/ ), this seems useful config to keep bundled somewhere near the app but I'm not sure "custom handler" is the place. Unless this code is trying to gesture at like, \~prefetching\~ JSON into the initial JS fetch?

philipcmonk: First, compare what's on master now: https://github.com/urbit/urbit/blob/a480e3aafd483d21aa2d589c1363ad2a502f32ef/pkg/arvo/app/clock.hoon

Second, yeah, I'm not a big fan of using `*server` like that.  I feel like "app" and "gen" aren't too bad for application and generator, but only if you know where to look for them.  But I'm converting a bunch of stuff to static gall right now, so I've declared those sorts of changes to be out of scope for now.

Third, how about what I just pushed.  Except for `/+  http-handler`, `%-  http-handler` and `+on-poke`, the rest is just registering the tile with `%launch` (which could abstracted into another library if desired).

philipcmonk:
I just pushed the "store requests in the library, not responses" change.  Now just adding `/+  http-handler` and `%-  http-handler` lets you treat it exactly as though by the time you receive the poke the subscription has already been set up.  Thus, from the perspective of the app, "when you get poked with `[path inbound-request:eyre]`, give updates on that path at your leisure."

> It just seems notable that 6 months ago this just worked, because it's the
> precise thing ducts are for architecturally, and now "establish 1:1
> connection to app" functionality has been removed from gall(and sounds like
> eyre).

It didn't "just work", it caused constant headaches.  It requires every subscriber to keep around a bunch of bones just so it cancel that subscription later and avoid resubscribing to the same path on a different duct.

With this architecture, the subscriber doesn't have to that in almost all cases.  The only time they have to store an id is if (1) they need to associate a poke with a subscription, like in dojo and eyre, and (2) it needs to happen on some event other than a subscription response (where the id could be put in the wire).

The complexity and additional state of bones is definitely not worth it if all you gain is the ability to do what a simple library can do.

Plus this approach is more powerful if you need to share these sessions.  For example, in the analogous dojo case you usually want web dojo to have a separate session from the console.  But maybe sometimes you don't.  Here, you just need to share the session id.  In the bones case, you need to somehow register on the server that these two bones are linked and anything done by one should affect the other, which sounds suspiciously like a shared subsription path.

> don't make *userspace* have to syn-synack-ack things!

You say "userspace" as though only user code can run there.  This problem goes away with a simple library to abstract over the back-and-forth.

> This in particular doesn't sit well with me, it's like auto-incrementing
> database fields vs UUIDs. (Relatedly, a subscriber can assign itself a
> /web/(shaf %app eny) perfectly fine). That particular implementation might
> be slow and/or cause less pretty debug output however, which is why my
> "hang on, this used to work before you removed the thing that was in active
> use for it" focus.

I understand this concern, and you're correct that in many cases you can shortcut this flow by "guessing" a new path.  I'm intentionally trying to do it in the most general and obviously-correct way first.

pkg/arvo/age/clock.hoon
ohAitch: Okay yup, completely missed that this was a copy of an existing file ty

I am alas only able to compare to "back in my day", where requests for static assets weren't being misrouted to apps in the first place. (I see there's also login logic there, so not completely static, but it seems like a lot of work to go for a boolean auth/noauth flag)

ohAitch:
> > It just seems notable that 6 months ago this just worked, because it's the
> > precise thing ducts are for architecturally, and now "establish 1:1
> > connection to app" functionality has been removed from gall(and sounds like
> > eyre).
> 
> It didn't "just work", it caused constant headaches. It requires every subscriber to keep around a bunch of bones just so it cancel that subscription later and avoid resubscribing to the same path on a different duct.

I said it worked for 1:1 connections XD
(which is why the flow I'm proposing has bones available but not mandatory, keeping access to the current gall "automagically manage who my subscribers are" logic)

Though, could you point me to the code that was managing bones for cancelling? I'm almost sure that either the source bones represented individual subscription causes, in which case you do in fact want to separately notify them of cancellations / only unsubscribe when the last one cancels, or it should set them to `0` if the most specific thing you could say about the subscription cause is "somewhere downstream of this app getting started" (and no messages needed to be sent directly to the bone).

> 
> Plus this approach is more powerful if you need to share these sessions. For example, in the analogous dojo case you usually want web dojo to have a separate session from the console. But maybe sometimes you don't. Here, you just need to share the session id. In the bones case, you need to somehow register on the server that these two bones are linked and anything done by one should affect the other, which sounds suspiciously like a shared subsription path.

"maybe sometimes you don't" given dojo in particular has stateful vector clock logic, almost certainly you do, unless you want to buffer-for-replay every message since the start of the connection. Is there a specific motivating example here?

Like, if you legitimately want shared sessions I would in fact recommend the /generic-path/[guid] thing, what I want to argue against is the removal of sessions that are architecturally known not to be shared.

> > don't make _userspace_ have to syn-synack-ack things!
> 
> You say "userspace" as though only user code can run there. This problem goes away with a simple library to abstract over the back-and-forth.
> 
> I understand this concern, and you're correct that in many cases you can shortcut this flow by "guessing" a new path. I'm intentionally trying to do it in the most general and obviously-correct way first.

Put me down as skeptical for "a poke, an eyre-specific pass, and only then a peer" being the obviously correct way to start a subscription.

philipcmonk:
> Put me down as skeptical for "a poke, an eyre-specific pass, and only then a peer" being the obviously correct way to start a subscription.

I mean correctness not as in ideal but as in "doesn't go wrong", and I mean "obviously" in the sense that I don't have to make any subtle arguments.

> "maybe sometimes you don't" given dojo in particular has stateful vector clock logic, almost certainly you do, unless you want to buffer-for-replay every message since the start of the connection. Is there a specific motivating example here?

It's true that dojo as written depends on being 2-party, but I don't think that's a fundamental limitation.  For catching up, you just need one of the parties to give you their state.

However, this isn't the primary motivation.  The primary motivation is that it reduces the complexity of the model by reducing the number of primitives, and this in fact leads to a lesser cognitive load.  You don't have to keep track bones/wires/ids except in those cases where it's obvious why you need to (I need a handle for a request that I may stream answers to over time).

It increases the load on "userspace", but if that load is paid entirely in libraries, then that's ideal.  It's more flexible than building it into the system, it isn't required to understand them unless you're using that specific feature, and they're just "macros" on top of a small set of primitives (poke and subscribe).

pkg/arvo/age/clock.hoon
philipcmonk: There's a generator path in eyre that should probably be used for this when it's static.  The only part that "needs" an app is registering the generator (the `%bound` request) and registering the tile with launch.  And both of those are stateless so could be done by a more powerful version of generators (possibly imps will morph into this).

ohAitch:
> > Put me down as skeptical for "a poke, an eyre-specific pass, and only then a peer" being the obviously correct way to start a subscription.
> 
> I mean correctness not as in ideal but as in "doesn't go wrong", and I mean "obviously" in the sense that I don't have to make any subtle arguments.
> 

I don't think the argument for "in the preexisting system, this could all be done in one request" is subtle. If bones exist, having to write up redundant extra logic(even once in libraries and once again in eyre) and keep track of corresponding state, introduces complexity that can cause things to go wrong. This is a fairly clear sign that ducts/bones/"current continuation" are a useful concept that should continue to exist.

> It's true that dojo as written depends on being 2-party, but I don't think that's a fundamental limitation. For catching up, you just need one of the parties to give you their state.

it's a distributed system, there's no "just" giving someone your state. Short of explicitly proxying through hood, at which point you can be back to one bone; or in the case of tile copying over a variable, in which case it makes more sense for the copy to be by reference, and at that point you can borrow the ID you're using to talk to eyre as well - by duplicating the subscription data locally from one connection.

> 
> However, this isn't the primary motivation. The primary motivation is that it reduces the complexity of the model by reducing the number of primitives, and this in fact leads to a lesser cognitive load. You don't have to keep track bones/wires/ids except in those cases where it's obvious why you need to (I need a handle for a request that I may stream answers to over time).
> 
> It increases the load on "userspace", but if that load is paid entirely in libraries, then that's ideal. It's more flexible than building it into the system, it isn't required to understand them unless you're using that specific feature, and they're just "macros" on top of a small set of primitives (poke and subscribe).

I don't think I follow the argument for "the underlying system has capabilities [X, Y], let us only expose the api for Y, so that libraries can rebuild X in terms of Y by sending 3x as many requests. this reduces total cognitive load understanding the system"

like, I do agree many apps shouldn't touch bones! But that itself sounds like a load to be paid entirely in libraries, have things that want to be bone-agnostic use a "manage subscriptions" library that keeps track of all ducts' inbound and outbound subscriptions, instead of presumably at present that logic being in gall. Then you actually don't have to keep track of wires except when you need to.

philipcmonk:
> I don't think I follow the argument for "the underlying system has capabilities [X, Y], let us only expose the api for Y, so that libraries can rebuild X in terms of Y by sending 3x as many requests. this reduces total cognitive load understanding the system"

> like, I do agree many apps shouldn't touch bones! But that itself sounds like a load to be paid entirely in libraries, have things that want to be bone-agnostic use a "manage subscriptions" library that keeps track of all ducts' inbound and outbound subscriptions, instead of presumably at present that logic being in gall. Then you actually don't have to keep track of wires except when you need to.

The argument is that "simple system with libraries to add features" is better than "complex system with libraries to dumb it down".

For one, this allows faster experimentation to find what are the really useful features, since it doesn't require consensus that it's worth putting in gall.  For example, I added a [lib/verb](https://github.com/urbit/urbit/blob/philip/mall-real/pkg/arvo/lib/verb.hoon) that lets you run `:app +verb` to print out what's happening in that app just by adding `/+  verb  %+  verb  &` at the top.  That doesn't depend on the bone thing specifically, but it's an example of the general principle that a small and diamond-like foundation is valuable because it's easier to build on.

Of course, the principle can be taken too far: it could make the system itself useless, so the complexity is just shifted and not removed.  But the complexity of bones has not proven worth it, based on the fact that they're connected to a large number of bugs, especially ones which are not discovered until they're on the network.

> it's a distributed system, there's no "just" giving someone your state.

It's fine if you start getting updates at the same time (or have already started getting updates).  So, eg right now drum subscribes to dojo, and webdojo should subscribe to dojo as well.  If they're subscribed to the same path, then all you have to do to keep them both up to date is (1) give dojo's version of the state to anyone on initial subscription and (2) when any of the three parties makes a change, send it to all subscribers.  Unless I'm misunderstanding OT and it's actually limited to a vector of two people?

ohAitch:
The current OT implementation definitely is just two-party! It could
theoretically be extended to n-party, though "send it to all subscribers"
gets difficult if you and the original owner of the session you're
"borrowing" are each an independent browser tab; seems much simpler to have
dojo itself bridge two two-way OT connections into a a 3-way V.
Discriminating by some sort of "actual individual client connection" ID,
doesn't have to be a bone but it's not the same thing as a shared session
either - or rather, I dispute the claim that a session in the former sense
frequently translates cleanly to the latter, you might want it also but
it's a different mechanism from normal single-client subscriptions.

pkg/arvo/age/clock.hoon
loganallenc: This looks nice to me as a straightforward port of existing functionality.

joemfb:
I'll boil-down and restate my objection to `%start-watching`: in this architecture, the source of inbound http requests can *NOT* be abstracted over. Since the server agent needs to kick the client/proxy *thing* (agent/vane), that *thing* must be identified and addressable.

(In all fairness, the %gall/%eyre interface that this is revising was also special-cased to %eyre due to the presence of %http-response as a sibling of %diff (now %fact). So this doesn't introduce a new special-case so much as it just moves the special-case around. And this is probably better than the special-case that was there before.)

We've talked before about the nice properties of the current relation between %dill and userspace. If I had to articulate the key strength of that system, I'd claim it's the fact that it completely abstracts over the source of some input, and therefore the "route" on which to respond. Dojo doesn't care if its input comes from drum or lens, drum doesn't care if its input comes from %dill or some other ship's drum over the network. It seems obviously right to extend this property to the http-server architecture as well.

The necessary feature for this property is the ability to associate an input with a stream of output. This is currently provided by bones -- in @ohAitch's characterization, the ability to specify the "current continuation". My proposal is to allow a %poke to be explicitly, systematically associated with some subscription path. @philipcmonk's suggestion is that this association should be built ad-hoc where desired, while this PR uses a special %eyre task to avoid forcing http-server agents to construct this ad-hoc association.

The general pattern of changes in static-gall is to enable control and abstraction in userspace by making things more explicit and low-level (at least in some sense): input and state as vases, %pass/%give directly addressing vanes, &c. But the total removal of bones seems to be in tension with this trend: it's no longer possible for agents to address the "current continuation". This is a drastic change, as the "current continuation" (by way of %pass/%give on ducts) is the core mechanism of Arvo (and, as of %alef, will be the core mechanism of the network). In short, the presence of bones can be abstracted over by libraries, while their absence cannot.

I think the use-cases that I care about are covered by some notion of binding a %poke to a subscription. That is less powerful than bones; there is no notion of some specific instance of a subscription, and therefore no systematic guarantee of an exclusively two-party "session", but that property can be approached by using a sufficiently unique subscription path (which can be constructed by the "client" with entropy, or by the "server" with an explicit handshake). This approach overloads the notion and purpose of a subscription path, as it's no longer merely a semantic namespace within the agent. That makes it less than ideal, but it's the best I can come up with without bones.

In summary, I think binding an input to an output stream is a necessary general userspace capability. If we can't achieve this without bones, then I think we have to seriously discuss restoring them, perhaps optionally.

philipcmonk:
You can make this a general purpose interface, and that's actually the full form of this approach, so it's worth describing.  I've got this working for spider, so I'll describe it there.

The idea is that you want to run an imp with some arguments and get the result.  This way you can take any arbitrarily long IO-heavy computation and spin it out into an imp, only persisting the result when it returns to you.

Of course, when you spawn an imp, you want to get the result of that, and not anybody else's imp, so you need a unique subscription path.  You could do it two ways:

- ask for an id by subscribing to /next-id, which will return one id and then %kick the subscription
- subscribe to /imp-result/[id] and poke with [id imp-filename arguments]
- receive result (or failure) of the imp from the subscription

Or, to avoid the first round-trip, if you're diligent about cycling your entropy:

- subscribe to /imp-result/[eny] and poke with [eny imp-filename arguments]
- receive result (or failure) of the imp from the subscription

I implemented the first one, but due to its complexity I'm inclined to switch to the second both in the eyre and the imp case.  It's straightforward to translate this to the eyre case, and this doesn't depend in any way on the identity of the caller, so it should satisfy your generality condition.


> I think the use-cases that I care about are covered by some notion of binding a %poke to a subscription.

Let me clarify: I'm good with a %poke being associated with a subscription path, but not with an individual subscription.  My concern with bones as proposed by @ohAitch as I understand them is that they're a clear violation of pubsub -- specifically the principle that you publish results to a path and not an individual subscriber.

If gall is really going to provide a pubsub interface, then it needs to translate *everything* to that as a clean layer.  Letting some of what's underneath leak through makes it significantly more complex.

ohAitch:
"bones as proposed by ohAitch" is I suppose a way to put "ducts as used for
all event processing for all of arvo's existence, and bones as a 1:1 opaque
duct id as used by gall for most of that existence". Like, buggy code has
been written using bones, buggy code has not yet been written using the
novel gall subscription unification mechanism, this does not translate into
enough of a vote of confidence to outright remove the former feature
instead of gradually moving things over as appropriate.
(Maybe new-talk is in the new model? My impression of new talk so far is
"fails to print backlog", which I guess could be due to philosophical
objections to tracking individual subscribers)

Once it is proven that bones are useless, then it could be argued their
absence is "clean". But being forced to build a general purpose interface
on top of a constrained interface on top of a general purpose interface
that did the original thing, is not a reduction of complexity however. HTTP
microservice style "you have only %poke, and can provide exactly one
response within 30 seconds max" would be an even smaller layer after all :P

The /path/[eny] thing seems as aforementioned obnoxious to log but overall
good. Having the /path/[eny] outside the body of the poke seems like a nice
quality-of-life tweak(look we're doing REST~), in particular for writing
poke type marks. In the non-imp case(and possibly the imp case) you'd want
that to be the full path and not just an id, but that's another minor
"worst case discriminate on the body / mark" concern.

At any rate, it certainly has precedent in api connectors(mb just the
twitter one?), which have not survived but iirc for the most part worked
fine.

>  (unit path) could go in the bowl to signify that an agent-action
corresponds to some active subscription

was the original proposal, and in fact does not associate with an
individual subscription - except insofar as the http-handler library should
abort duplicate subscribers on an expected-unique path, or agent-actions
bearing a path that's not subscribed. And maybe it shouldn't even, previous
OT discussion notwithstanding a /read-only/ streaming copy of sole,  tee
 style, is perfectly possible in this model if you just ignore everything
referencing the prompt line.

philipcmonk:
Okay, I pushed a new version.  Unlike the last version, this one is:

- Stateless in both eyre and the app.
- Synchronous without requiring a library to delay the poke
- Source-agnostic -- anybody who subscribes to /http-response/id and sends a poke %handle-http-request with that id will get the same interface eyre does.  Could be interesting build a little proxy app that counts requests and logs them or something.

The flow looks like this:

- Eyre creates a unique `@ta`.  This could be anything, and you can include helpful diagnostic info if you want.  I'm using `(cat 3 'eyre--' (scot %uv (sham duct)))`.  This makes a clean connection between vane-space ducts and app-space ids.
- Eyre subscribes to /http-response/eyre-id and pokes with mark `%handle-http-request` and data `[eyre-id inbound-request]`.
- App accepts `/http-response/*` subscriptions silently
- App handles %handle-http-request request by giving results on `/http-response/[given-eyre-id]`.
- If the request is cancelled before the app has responded, Eyre closes the subscription with `%leave`.
- When app is done responding to request (which will usually be immediately but needn't be), it `%kick`s the subscription.
- Eyre passes the `%fact`s to vere as it gets them, and completes the connection when it receives the `%kick`.

In another branch I've got the same system working for the imp case I described earlier.

@eglaysher this is ready to review.  Questions:

- How can I test cancelled http requests?  I want to make sure the duct is what I expect.  Also, I'm assuming there's no information being passed in the inbound-request, is that right?  Else I would need to poke it in.
- Any idea why giving results with no total data hangs instead of producing the result?  The main change is that in the past eg not-found 404 was a single `%start` http-event and is now 3 http-events: one `%start` with empty data, one `%continue` with empty data and not complete, and one `%continue` with empty data and complete.  It seems to do the same thing if I remove the middle `%continue`.

joemfb:
It would be cool to build a system agent that translated this pattern of http-requests to %iris. We would then have total http client/server pub-sub abstraction in userspace, which would pay dividends for integration testing and proxying use-cases.

ohAitch:
Sounds great!

As a small nitpick, I'd go for (cat 3 'eyre_' (scot %uv (sham duct))) over
'eyre--': the '--' separator convention is useful for @tas clean contexts,
which encoded @uv decidedly is not.

(other ways to spell that would be (pack /eyre/(scot %uv (sham duct))) and
(pack (dray /tas/uv %eyre (sham duct))), but that's really a style thing.
There's also the possibility of outright ~(rent co %many $/tas/%eyre
$/uv/(sham) ~) for ._eyre_0va.sdfjkl__ which imo would be overkill)


+1 on joe's suggestion of not having in-bound redundant subscription state
management, though I'm not sure I follow

> This would still require an explicit %cancel if an agent needed to do so
after giving a partial response

In this case, aren't the bytes of the previous response chunks and for that
matter HTTP headers already sent? What are you canceling as distinct from
"successfully" closing the TCP connection, possibly with an in-band encoded
error?

joemfb:
@ohAitch, `--` as a convention to distinguish `foo-bar` / `foo-bar-baz` / `foo-bar--baz-bud`? Does anything parse this convention?

As for cancel, you're right that the headers and maybe some portion of the body could already have been sent. It just seems right to distinguish successfully-completed multipart response from unable-to-complete multipart response. I think the latter should be an early TCP socket close in HTTP 1/1, I'd have to check what it should be in HTTP 2.

ohAitch:
I dearly hope not, but it's popped up once or twice! (Honestly the
difference between a "convention" and "format" is probably whether or not
it has any semantic consequences, vs being enforced by and beneficial for
code review)

Insofar as the transport has an "abnormal termination" lever that does seem
useful to expose, legit. (Probably worded more strongly than just %cancel;
and nack'ing the initial poke by crashing should hopefully also work - with
eyre cleaning up the peer)

philipcmonk:
> The presence of a subscription and its explicit lifecycle makes the multipart response management in +http-event somewhat superfluous.

Yes, the structure is too general now.  However, apps don't see that structure, and it's what's sent to vere to and iris, so I don't think it's worth the trouble to switch it now with King Haskell ready to land soon, and given time constraints on mall work.

> I don't see any great advantage in maintaining structural continuity between userspace and arvo/unix APIs

Yeah, as implemented it no longer maintains structural continuity.  Ack failures are translated to `%cancel`s, `%kick` is success, and you can give an explicit `%cancel` fact if you want to cancel.

> Rather than an ad-hoc binding between the %poke payload and the subscription path, I'd like to see poke-on-a-path become a primitive, with (unit path) in the bowl. %gall would nack pokes on paths not corresponding to one-or-more already-bound subscriptions, and this whole pattern would become systematic.

I'm unsure about this, so I figure the best approach is to implement it ad-hoc a few times (here, imps, dojo).  That should help characterize the feature.

> `(cat 3 'eyre_' (scot %uv (sham duct)))`

I tend to agree, except I sort of feel like paths "should" only be whatever parses as a path, and `/eyre--0v1.23456` does while you have to do `/'eyre_0v1.23456'`.  I don't feel strongly, though.

ohAitch:
oh huh definitely would expect the latter to parse, maybe you do want
._eyre_thing__ then
or at least like ~.eyre_thing - if the rule is "valid coin", why is the --
version working again?

joemfb:
```
> ~.eyre_0v1.23456
~.eyre_0v1.23456
> (scan "~.eyre_0v1.23456" nuck:so)
[%$ p=[p=~.ta q=1.099.465.366.323.144.161.535.485.289.855.333]]
```

But the common entrypoints don't work:

```
> (slaw %ta ~.eyre_0v1.23456)
~
> (slay ~.eyre_0v1.23456k)
~
```

Because of this:

```
> (trip ~.eyre_0v1.23456)
"eyre_0v1.23456"
```

ohAitch:
You'd want `~.~.foo` to represent `'~.foo'` right

my question is given [path parsing](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L12059) b[o](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L12062)t[t](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L12087)o[m](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L12068)s [o](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L12081)u[t](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L6110) in [++sym](https://github.com/urbit/urbit/blob/a480e3a/pkg/arvo/sys/hoon.hoon#L5667) for a leading letter, which in turn is defined as `/[a-z][0-9a-z-]*/`, why does `/eyre--0v1.2345` work

ohAitch:
(and presumably "actually works" as in `=(/eyre-... /'eyre-...')`, don't have a repl in front of me alas but would appreciate further research here)

joemfb:
Ah, you're right about `~.~.XX`.

That path doesn't work, it fails to parse at the `.`:

```
> (ream '/http-response/eyre--0v1.23456')
{1 25}
syntax error
> (ream '/http-response/~.eyre--0v1.23456')
[%clsg p=~[[%sand p=%tas q=8.037.768.698.167.340.501.673.090.249.832] [%sand p=%ta q=18.445.967.935.322.515.397.216.109.716.101.564.542.590]]]
```

philipcmonk:
Ah, quite.  In that case, I'll forget about the literal parser and just use `eyre_0v1.23456`.

ohAitch:
There's a further complication here that ++stab is more liberal and _will_
accept /a.b I think

but if the idea is these should be usable as literals / not blow up when
you put the pretty-printed versions of themselves in the dojo, then my
updated recommendation to philip is either (scot %ta) or the full on
coin-tuple syntax
(unless the pretty-printer /'x_y' escapes them properly, in which case the
short one seems fine)

joemfb:
Mark it down as another not-quite-right:

```
> /http-response/'eyre_0v1.23456'
[%http-response 'eyre_0v1.23456' ~]
> `path`/http-response/'eyre_0v1.23456'
/http-response/eyre_0v1.23456
```

`(scot %ta ...)` is superior:

```
> `path`/http-response/(scot %ta 'eyre_0v1.23456')
/http-response/~.eyre_0v1.23456
```

ohAitch:
Huh on further reflection I do vaguely remember putting special handling
for this in the path printer, maybe it was insufficiently paranoid though.
At any rate ~. isn't  actually more characters than quotes…

philipcmonk:
```
> (scot %ta (cat 3 'eyre_' (scot %uv (sham *duct))))
~.~.eyre_0v4.7j7tl.qs19r.9fa4k.3tc3j.r825p
> `path`/(scot %ta (cat 3 'eyre_' (scot %uv (sham *duct))))
/~.eyre_0v4.7j7tl.qs19r.9fa4k.3tc3j.r825p
> `path`/~.eyre_0v4.7j7tl.qs19r.9fa4k.3tc3j.r825p
/~.eyre_0v4.7j7tl.qs19r.9fa4k.3tc3j.r825p
```

Good enough for me

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -912,14 +909,9 @@
...
-      ^-  note
-      :^  %m  %deal  [our our]  :-  app.action.u.connection
-      ::
-      ^-  task:agent:mall
-      :*  %poke
-          %handle-http-cancel
-          !>(inbound-request.u.connection)
+      :*  duct  %pass  /watch-response/[eyre-id]
+          %m  %deal  [our our]  app.action.u.connection
+          %leave  ~
```
eglaysher: This probably doesn't matter since I assume there's no real usages of `handle-http-cancel`, but the application does get the ability to handle the leave event, right?

philipcmonk: Yes, the application can handle the `%leave` event, though it can't reply to it (which is why I can get rid of the section with your comment about explicitly ignoring the ack).

PR #1926, opened 2019-11-08
Fix url explaining how to add swap space

thunderrabbit:
When running urbit on a machine without enough swap space, the message was

> see urbit.org/docs/getting-started/installing-urbit/#swapfor adding swap space

So I fixed the URL and added a space before "for"

PR #1930, opened 2019-11-09
vere: fix url explaining how to add swap space

thunderrabbit:
When running urbit on a machine without enough swap space, the message was

see urbit.org/docs/getting-started/installing-urbit/#swapfor adding swap space

So I fixed the URL and added a space before "for"

This PR supercedes #1926 and hopefully fits the conventions better.


loganallenc:
Nice quality of life fix!

PR #1933, opened 2019-11-11
clay: fix meet-conflict crash

philipcmonk:
When clamming was changed to crash instead of coerce, this idiom broke.  Fortunately, `_` works fine.

fixes urbit/docs#714


ohAitch:
wait _~ still does coerce? that's confusing

joemfb:
It's not so much that `$_` coerces, it just clams to a constant. So it won't crash regardless of the sample.

philipcmonk:
Ah yes, I forgot we added `+key:by`; will change.

ohAitch:
Right, all other molds no longer clam and now er gauge, while $_ still
clams. The consistent handling of this would be to have $_ coerce to a
constant, and then crash unless the input was that exact constant

(Unless ofc molds are after all a useful syntactic capability that we still
want in the system...)

joemfb:
@ohAitch, "gauge" terminology predates me. But there's an argument that `$_` and `$-` (and the not-yet-parsed core interfaces in `+spec`) should all behave as you say. It seems like the most inconsistent part of the new crash-first story for molds.

ohAitch:
sorry it's not like an Official Term, I just meant the new crash-first story

joemfb:
@philipcmonk, `+key:by` producing a set is obviously correct, but I find myself usually wanting a list, which leads me to `~(tap in ~(key by *(map)))` or `(turn ~(tap by *(map)) head)`, neither of which is particularly satisfying ...

joemfb:
A particular problem with `$_` now is the pattern of using it to change the bunt: `_|` or `_1`, for instance. The right way to write those now would be `$~(| ?)` and `$~(1 @ud)`, but I haven't done that, or seen it anywhere.

ohAitch:
I think you mean one of the other new $ runes and $~ is an iron gate? But
yeah _1 being equivalent to @ moldwise sounds right to me, I guess we don't
want "check equal to sample" and rather "check fits general type of
sample"; mostly it seems bad for _%1 (or in the motivating case _~) to
traditional-sense-mold when ,%1 (resp. ,~) crashes

joemfb:
No, I mean `$~(a b)`, which explicitly sets the bunt of `b` to `a`.

```
> *@ud
0
> (@ud 10)
10
> *$~(1 @ud)
1
> ($~(1 @ud) 10)
10
```

`$-(a b)` is `_|~(a *b)`.

```
> $-(@ @)
<1.ngv {* {our/@p now/@da eny/@uvJ} <21.bal 27.kva 7.ylp 36.uek 92.otg 247.tye 51.mvt 126.xjf 41.mac 1.ane $141>}>
> *$-(@ @)
<1|akq {@ @ @}>
> ($-(@ @) dec)
<1|akq {@ @ @}>
> (($-(@ @) dec) 0)
0
> (($-(@ @) dec) 10)
0

```

`,%1` used to be `_%1`, but now it's built-in:

```
> *%1
%1
> (,%1 1)
%1
> (,%1 0)
> *_%1
%1
> (_%1 1)
%1
> (_%1 0)
%1
```

Not quite optimal ...

ohAitch:
Ah yup you're right, am I thinking of $-(a b) or sth?

philipcmonk:
Yes, $- is equivalent to both $_  ^|  |=  and $_  |~ and was formerly $+

ohAitch:
So it used to be |+ $+ and is now |~ $- then?

philipcmonk:
Yeah, exactly.

PR #1939, opened 2019-11-12
arvo: add +trouble generator (for troubleshooting)

jfranklin9000:
I'll PR this now so others can help work on it. 

We'll want to output the comments about what to compare on ethercan.io and how to do it.

ht: pcmonk, ohAitch, Fang

### DIFF
--- a/pkg/arvo/gen/trouble.hoon
+++ b/pkg/arvo/gen/trouble.hoon
@@ -1,0 +1,37 @@
+:-  %say
+|=  [[now=time * bec=beak] ~ ~]
+=*  our      p.bec
+=/  sponsor  (sein:title our now our)
+:-  %noun
+=<
+:~
+  [%base-hash .^(@uv %cz (pathify ~.base ~))]
+  [%home-hash .^(@uv %cz (pathify ~.home ~))]
+  ::
+  (info %our our)
+  (info %sponsor sponsor)
+  (info %dopzod ~dopzod)
+  ::
+  ["Compare lifes and rifts to values here:"]
+  ["https://etherscan.io/address/azimuth.eth#readContract"]
+  ["  life - getKeyRevisionNumber"]
+  ["  rift - getContinuityNumber"]
+==
+|%
+++  pathify
+  |=  [a=@ta b=(unit ship)]
+  ^-  path
+  =/  o=@ta  (scot %p our)
+  =/  n=@ta  (scot %da now)
+  ?~  b  ~[o a n]
+    ~[o a n (scot %p u.b)]
+::
+++  info
+  |=  [=term =ship]
+  :*  term
+      ship=ship
+      point=(crip (slag 2 (scow %ui ship)))
+      life=.^(* %j (pathify ~.life `ship))
+      rift=.^(* %j (pathify ~.rift `ship))
+  ==
+--


philipcmonk:
Very nice!  Two nice-to-haves:

- Maybe check specifically for ~dopzod's current life as well.  This would identify the issue that's happening a lot right now that you boot with keys for your sponsor chain, but not for anyone else.

- Could be good to go ahead and print out the `@ui` of your ship and sponsor, to make it faster to check their life on etherscan.

jfranklin9000:
I used `+dedot` instead of `@ui` because i think `0i123` looks funny.  I couldn't get `+murn` to work in `+dedot` so I wrote my own.

Is it possible to get the etherscan stuff in the generator?  Could I scry azimuth-tracker?

What is the type of `/===`?  I want to write a helper function that returns these.

philipcmonk:
I believe the ability to run an http request in a generator was removed with the eyre rewrite a few months ago.  Unfortunately azimuth-tracker is part of what we're trying to check, and it would only have the info if they've already run `:azimuth-tracker|listen ~ %app %azimuth-tracker`.

`/===` is of type `path`.  Technically it's `[@ta @ta @ta ~]` because it's a literal, but it nests within `path`.

jfranklin9000:
Here's the final output:
```
> +trouble
[ [%base-hash 0v1d.p3qi7.rllqm.i56i4.9tuuq.s5nqv.dlkca.7ipak.71gil.5v6ov.12jv5]
  [%home-hash 0ve.9secj.n8js1.tlisi.a1h8a.s4j6u.m7io2.bvv78.mha5m.kt0m2.436h1]
  [ %our
    ship=~hacduc-tombyl-socnes-fonnyl--danryt-disdyt-tocfel-marzod
    point='232291692758278314518123759985998954752'
    life=1
    rift=1
  ]
  [%sponsor ship=~marzod point='256' life=3 rift=0]
  [%dopzod ship=~dopzod point='4608' life=1 rift=0]
  "Compare lifes and rifts to values here:"
  "https://etherscan.io/address/azimuth.eth#readContract"
  "  life - getKeyRevisionNumber"
  "  rift - getContinuityNumber"
  ~
]
```
I can squash the commits if needed.

pkg/arvo/gen/trouble.hoon
```diff
@@ -0,0 +1,46 @@
...
+  ==
+::
+++  dedot
+  |=  =ship
+  ^-  cord
+  =/  tap=tape  (trip (scot %ud ship))
+  =|  lis=tape
+  |-  ?~  tap  (crip (flop lis))
+  ?:  =(i.tap '.')  $(tap t.tap)
+  $(lis [i=i.tap t=lis], tap t.tap)
```
Fang-: You can always just:

```hoon
> ((d-co:co 1) ship)
"1245952"
```

joemfb: Or:

```
(slag 2 (scow %ui ship))
```

joemfb:
@jfranklin9000, this is not strictly necessary, but if you can have pretty-printed output if you produce `%tang` (ie, `(list tank)`) instead of `%noun`. You'd just cons `%leaf` onto the head of each tape, and then either reverse them in source or `+flop` them as you produce them.

ohAitch:
Seems like a good starting point!

The next step if you're up to it is to build this out into |trouble, which
prints the exact same things(probably with just ~&), an then has a hood
++poke-helm-trouble or w/e run all the |hi and so on effects mentioned in
the initial post. (This can probably also do the etherscan api request
discussed in his thread? Though having that particular one just print
directions seems fine honestly)

ohAitch:
Another quality-of-life thing here might be truncating each %cz to just the
first meaningful segment (0v1.23abc), kind of like git short hashes. I
think that would be (rsh 0 45 x)?

ohAitch:
with 256-225 = 31 bits left over, i.e. exactly the same chance of false positives as a mug collision :P
(as trivia, this is also why `@uvI` hashes start `0vx.` half the time and `0v1x.` the other half, 256 is almost 255 which would be exactly 51 base-32 digits)

jtobin:
> I'll have an updated version later today.

:+1: I'll hold off merging until then.

philipcmonk:
@jtobin I don't know when you're planning the next OTA, but do make sure this goes in, even if it's not in its final form.  Would be a big boon for support.

jfranklin9000:
I got rid of `+dedot` but I kept the `cord` output for point numbers.
I kept the full hashes because that's how they're reported in the OTA announcements.
I'll leave the `%tang` and `|trouble` stuff for later.
I squashed all the commits.
It's ready to go AFAIC.

PR #1941, opened 2019-11-12
chat-cli: send invites on-;invite

Fang-:
This augments permission management with invite sending, when setting "positive"
permissions. This matches talk's behavior.

Also implements `+full:tr`, which renders as `~ship/path`, even for local targets.

@loganallenc This does more or less what the chat-js does, please confirm this is all sane. We generate the UID from chat-in-question and target, to avoid duplicate invites piling up if they're re-sent but neither accepted nor deleted.

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -10,7 +10,7 @@
 ::    and trust it to take care of the rest.
 ::
 /-  *chat-store, *chat-view, *chat-hook,
-    *permission-store, *group-store,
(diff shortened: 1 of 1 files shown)


loganallenc:
Question: Are invites displayed anywhere from %chat-cli?

Fang-:
No not yet. I imagine you want to subscribe to the invites store and just print out whenever something new comes in.

Seems like less-trivial work (especially wrt accept semantics), so figured I'd just get this bare minimum in for now.

loganallenc:
Ok! Sounds good. In that case, looks good to me.

Issue #1875, opened 2019-10-22
chat: unread counter does not always appear


ixv:

loganallenc:
If you see this again, can you console.log(store.state.inbox); ?

jtobin:
This has been fixed by #1948.

PR #1948, opened 2019-11-13
chat: add history functionality. can optionally request backlog.

### LINKED ISSUE #1858: chat: feature request: load backlog when joining a chat
Please do this, its sucks to not have this.

loganallenc:
Upon chat creation, you can specify whether subscribers to the chat are allowed to ask for portions of the chat message backlog. If you specify that they cannot, then they cannot.
If you specify that they can, then they can optionally request a portion of the backlog starting from an index specified by the last path segment. For instance: /~zod/urbit-help/20 would fetch every message from index 20 forward.

I have added a %messages event to batch messages to prevent frequent pump-locks upon high message bandwidth, and am batching messages within backlog requests (500 per batch).

The joiner can request whether they want to receive history for a channel or not. This allows a user to quickly join a channel regardless of whether the channel has a long backlog or not.

Fixes #1858 and #1875.

PS: Joe, I've fixed your issue with dropped messages by implementing the start index in backlog request. I've tested this with repeated pump-locks, and no messages are lost.


pkg/arvo/app/chat-cli.hoon
```diff
@@ -363,13 +365,20 @@
         ;~  (glue ace)
           (tag %create)
           security
-          ;~(plug path (punt ;~(pfix ace glyph)))
+          path
+          (punt ;~(pfix ace glyph))
+          (punt ;~(pfix ace ;~(pose (cold %| bar) (cold %& pad))))
```
Fang-: Pretty sure this won't parse correctly, demanding at least two spaces after the `path` (you're punting, but the `glue`'s still there!). You want:

```hoon
          ;~  plug
            path
            (punt ;~(pfix ace glyph))
            (punt ;~(pfix ace ;~(pose (cold %| (jest 'n')) (cold %& (jest 'y')))))
          ==
```

Untested, but you get the idea. Note also the `y` and `n` instead of `& |` to avoid ambiguity with the glyph parsing. (Talk had a parser-primitive for that, should probably bring it back.)

pkg/arvo/app/chat-cli.hoon
```diff
@@ -363,13 +365,20 @@
...
         ;~((glue ace) (tag %delete) path)
         ;~((glue ace) (tag %invite) rw path ships)
         ;~((glue ace) (tag %banish) rw path ships)
       ::
-        ;~((glue ace) (tag %join) ;~(plug targ (punt ;~(pfix ace glyph))))
+        ;~  (glue ace)
+          (tag %join)
+          targ
+          (punt ;~(pfix ace glyph))
+          (punt ;~(pfix ace ;~(pose (cold %| bar) (cold %& pad))))
```
Fang-: Same as above.

I tried my hand at making an `(cmd-with-args %cmd ~[required] ~[optional])` parser building function but never could get it to compile. Should probably do another attempt, these are getting out of hand.

pkg/arvo/app/chat-cli.hoon
```diff
@@ -613,18 +622,22 @@
...
-      ::  write
-      ?-  security
-        ?(%channel %mailbox)  ~
-        ?(%village %journal)  [our-self ~ ~]
+          ::  write
+          ?-  security
+            ?(%channel %mailbox)  ~
+            ?(%village %journal)  [our-self ~ ~]
+          ==
+          ?~(allow-history %.y u.allow-history)
```
Fang-: `+fall` exists. But if we're gonna do defaulting to constants, probably do that in the parser instead of in the logic. Same for `;join`.

pkg/arvo/sur/rw-security.hoon
```diff
@@ -0,0 +1,8 @@
+|%
++$  rw-security
```
Fang-: Can we please not do unrelated renaming/restructuring in the same commits as feature changes?

pkg/interface/chat/src/js/components/lib/chat-input.js
```diff
@@ -25,20 +25,22 @@ export class ChatInput extends Component {
...
-          Date.now(),
-          {
-            text: `${Date.now()}`
-          }
-        );
-      }
-      setTimeout(closure, 1000);
-    };
-    setTimeout(closure, 2000);*/
+    //let closure = () => {
```
Fang-: lol @ these changes

pkg/arvo/sur/chat-store.hoon
```diff
@@ -30,16 +30,24 @@
...
   $%  [%create =ship =path]         ::  %create: create a mailbox at ~ship/path
       [%delete =path]               ::  %delete: delete a mailbox at path
       [%message =path =envelope]    ::  %message: append a message to mailbox
       [%read =path]                 ::  %read: set mailbox to read
   ==
 ::
++$  chat-action
+  $%  ::  %messages: append a list of messages to mailbox
+      ::
+      [%messages =path envelopes=(list envelope)]
```
Fang-: Shouldn't `%message` just be deprecated in favor of this?

Fang-: Also updating types that go over the air has historically given problems when ships on latest and latest-1 tried to communicate. Did you test this case on a fakenet, or are these changes going to be breached in?

pkg/arvo/app/chat-hook.hoon
```diff
@@ -33,42 +34,62 @@
...
-|_  [bol=bowl:gall state-one]
+|_  [bol=bowl:gall state-two]
 ::
 ++  this  .
 ::
 ++  prep
-  |=  old=(unit state-both)
-  ^-  (quip move _this)
+  |=  old=(unit versioned-state)
+  |^  ^-  (quip move _this)
```
Fang-: What's this about?

pkg/arvo/app/chat-hook.hoon
```diff
@@ -157,14 +181,61 @@
...
   ?~  box  !!
   :_  this
-  [ost.bol %diff %chat-update [%create (slav %p i.pax) pax]]~
+  :-  [ost.bol %diff %chat-update [%create (slav %p i.pax) pax]]
+  ?:
+    ?&
+      ?=(^ backlog-start)
+      (~(got by allow-history) pax)
+    ==
+    (paginate-messages pax u.box u.backlog-start)
```
Fang-: Why do we paginate instead of just sending whatever's requested? I don't think we fall over on big messages anymore.

pkg/arvo/sur/rw-security.hoon
loganallenc: I think this is a small, positive change...

pkg/arvo/sur/chat-store.hoon
loganallenc: I expect this to work just fine once all ships have updated given that my state import functions have been tested on ships with older state versions. As for this network message issue, seems like it'll only be a problem for ships on latest vs latest-1, which shouldn't last very long?

loganallenc: As for %message, I think it's semantically useful to have a "singular version". Idk, arguable.

pkg/arvo/app/chat-hook.hoon
loganallenc: It scopes the migrate state function within this core to show that it's only used here.

pkg/arvo/app/chat-hook.hoon
loganallenc: Idk... Big messages scare me.

pkg/arvo/app/chat-cli.hoon
loganallenc: I know very little about this type of parsing, so I was cargo-culting. Thanks for your suggestions.

pkg/arvo/sur/rw-security.hoon
Fang-: And I agree! Just saying it's bad practice to include it in a commit that does something completely unrelated. The change isn't at all detectable from the commit message, and it pollutes the diff of the change the commit is claiming to make, which is messier for review.

Not a big deal in this particular case, but gotta stay vigilant. (:

pkg/arvo/app/chat-hook.hoon
Fang-: Oh, durr, I read this as a `^|` somehow. Nevermind!

loganallenc:
I don't care about supporting the "negative index" case yet. I would prefer to keep the interface simple with a "request everything", "request to be updated with everything since I was last in the room", or "request only new messages, no old messages". This is a simpler api for the end user than providing them a combinatorial explosion of options about how much backlog to request.

pkg/arvo/app/chat-cli.hoon
ohAitch: I think you want ++fuss from hoon.hoon

pkg/arvo/sur/chat-store.hoon
Fang-: Yeah I don't doubt that app state will be fine.

> it'll only be a problem for ships on latest vs latest-1, which shouldn't last very long?

You think too highly of the current network. (^:

iirc the consequences were pretty harsh last time we messed this up. Big stack traces, unusable chats, etc. Worth testing and maybe even holding off until breach for.

---

> I think it's semantically useful to have a "singular version". Idk, arguable.

If it's semantics, then that difference should be explicit (instead of `%singular` and `%plural`). Implementation-wise, I've found it nice to stick to singular when you can, but just go all-in on plural when you can't. Having both is just interface bloat.

joemfb: This kind of change is supposed to work OTA -- if the message types become incompatible, the subscription is cleared (or the sender backs off, as appropriate). In practice, that process is incomplete and not particularly tuned, so it can be quite noisy.

A simple solution is to just use a different mark for backlog subscriptions. The receiver won't make the subscription if they don't have the mark, and the new type won't mess with current subscriptions. Then you can combine the types in the next breach.

pkg/arvo/sur/rw-security.hoon
joemfb: Separate commits endorsed.

pkg/arvo/app/chat-hook.hoon
joemfb: Pagination is generally righteous IMO, at least in cases where a client can practically make use of the individual pages. But there also aren't any problems with big messages that I'm aware of.

pkg/arvo/sur/chat-store.hoon
loganallenc: This simple solution feels very messy to me... I'd much prefer noise in the dojo than making another mark.

joemfb: Tfw a developer is unwilling to make sacrifices for the experience of their users...

pkg/arvo/sur/rw-security.hoon
loganallenc: I have now split this up into different commits. Laborious, but worth it!

joemfb: Positively virtuous!

pkg/arvo/app/chat-cli.hoon
loganallenc: cool, will use +fall

loganallenc: done

pkg/arvo/app/chat-cli.hoon
loganallenc: done, thanks!

Fang-: This uses `+fuss` now, but the code below doesn't, and both still want too many spaces.

loganallenc: I've tested out the ;create parser, and it only wants a single space in between each option. Currently doing ;join

Fang-: Oh huh, sorry, I think gh was showing me outdated code. (Or I was interpreting it as the old code.) This does indeed seem good. Carry on! (:

pkg/arvo/app/chat-cli.hoon
loganallenc: this is done

loganallenc:
Can I get another pass on this? @joemfb @philipcmonk @Fang-

pkg/interface/chat/src/js/components/lib/chat-input.js
Fang-: Probably just remove these.

pkg/arvo/sur/chat-store.hoon
Fang-: Seconding @joemfb's sentiment here. Whatever's messy about it we can clean up on next breach (that is, right now in a separate, breach-flagged PR). We're busy enough with support as-is, don't need more loud errors scaring people.

Fang-:
Will do another code pass tomorrow!

pkg/arvo/sur/chat-store.hoon
loganallenc: If you can't add a type to a tagged union OTA, then our support for OTAs doesn't exist imo.

vvisigoth: Call it what you will. I've seen the fall out of not doing this and it sucks and will mean emails that someone else has to respond to. Please do this and let's wrap this up.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -110,20 +131,23 @@
...
     (create-permission [%chat path.act] security.act)
   ::
       %add-synced
     ?>  (team:title our.bol src.bol)
-    =/  chat-path  [%mailbox (scot %p ship.act) path.act]
+    =/  chat-path=path  [%mailbox (scot %p ship.act) path.act]
     ?:  (~(has by synced) [(scot %p ship.act) path.act])
       [~ this]
     =.  synced  (~(put by synced) [(scot %p ship.act) path.act] ship.act)
+    =/  history=path  ?:(ask-history.act /0 /~)
```
Fang-: Wait, do we not want the empty path `/` here for the second case? Pretty sure it's more common to do `/some/resource` and then parse the absence of a trailing `/index`, instead of doing `/some/resource/nothing` explicitly. Certainly have never seen suffix `/~` before.

pkg/arvo/app/chat-hook.hoon
Fang-: _Maybe_ this is unresolved since pagination here is a data-sending mechanism rather than part of the data-retrieving query. Since big messages aren't a problem, maybe just pull this logic out and send backlog all at once, to reduce complexity. But I don't feel super strongly.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -263,7 +341,7 @@
     ?.  =(u.shp src.bol)
       [~ this]
     :_  this(synced (~(del by synced) path.diff))
-    :-  (chat-poke diff)
+    :-  (chat-poke [%delete path.diff])
```
Fang-: Isn't this equivalent to the previous code (passing in `diff` as-is)?

Same for the `%message` case below. What am I missing?

loganallenc: Type nesting changes for `chat-base`, `chat-update`, and `chat-action`.

pkg/arvo/app/chat-cli.hoon
```diff
@@ -39,12 +39,14 @@
...
-      [%create chat-security path (unit glyph)]     ::  create chat
+      ::
+      ::  create chat
+      [%create rw-security path (unit glyph) (unit ?)]
       [%delete path]                                ::  delete chat
       [%invite ?(%r %w %rw) path (set ship)]        ::  allow
       [%banish ?(%r %w %rw) path (set ship)]        ::  disallow
     ::
-      [%join target (unit glyph)]                   ::  join target
+      [%join target (unit glyph) (unit ?)]          ::  join target
```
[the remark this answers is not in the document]
loganallenc: In that case, I will keep it unit.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -33,42 +34,62 @@
...
+        synced.sta  synced
+        allow-history.sta  (create-allow-history synced)
+        invite-created  %.y
+    ==
+    this(+<+ sta)
+  ::
+  ++  create-allow-history
+    |=  synced=(map path ship)
+    ^-  (map path ?)
+    (~(run by synced) |=(shp=ship %.n))
```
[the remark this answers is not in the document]
loganallenc: thanks, will do.

pkg/interface/chat/src/js/components/lib/chat-input.js
loganallenc: done

pkg/arvo/app/chat-store.hoon
```diff
@@ -198,6 +199,35 @@
...
+  =*  envelope       i.envelopes.act
+  =*  letter         letter.envelope
+  =?  letter  &(?=(%code -.letter) ?=(~ output.letter))
+    =/  =hoon  (ream expression.letter)
+    letter(output (eval bol hoon))
+  =:  length.config.u.mailbox  +(length.config.u.mailbox)
+      number.envelope  +(length.config.u.mailbox)
+      envelopes.u.mailbox  (snoc envelopes.u.mailbox envelope)
+  ==
+  =.  inbox  (~(put by inbox) path.act u.mailbox)
```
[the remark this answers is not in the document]
philipcmonk: Isn't the difference that this doesn't send the diffs individually?  You could factor out the duplicated code, though.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -157,14 +181,57 @@
...
+  ^-  (list move)
+  =/  moves=(list move)  ~
+  =/  end  (lent envelopes.mailbox)
+  ?:  |((gte start end) =(end 0))
+    moves
+  =.  envelopes.mailbox  (slag start `(list envelope)`envelopes.mailbox)
+  |-  ^-  (list move)
+  ?~  envelopes.mailbox
+    moves
+  ?:  (lte end 5.000)
```
philipcmonk: 5000 is a lot, have you tested that many to make sure it doesn't bail:meme?

pkg/arvo/app/chat-hook.hoon
```diff
@@ -157,14 +181,57 @@
 ++  peer-mailbox
   |=  pax=path
   ^-  (quip move _this)
+  ?>  ?=(^ pax)
+  =/  last  (dec (lent pax))
+  =/  backlog-start=(unit @ud)
+    %+  rush
+      (snag last `(list @ta)`pax)
+    dem:ag
+  =>  .(pax `path`(oust [last last] `(list @ta)`pax))
```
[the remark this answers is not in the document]
loganallenc: yeah, TMI. I'll change to last 1.

pkg/arvo/app/chat-store.hoon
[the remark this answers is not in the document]
loganallenc: I've now factored out duplicated code.

pkg/arvo/app/chat-hook.hoon
loganallenc: It's fine either way, but I actually like the /~ convention.

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
loganallenc: done

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
loganallenc: done

pkg/arvo/app/chat-hook.hoon
loganallenc: Works just fine, just tested it.

pkg/arvo/sur/chat-store.hoon
loganallenc: Working on this next.

loganallenc: I've tested it with the new "chat-two" marks, and it all works great.

loganallenc:
Should be completely finished @Fang-.

pkg/arvo/lib/chat-json.hoon
```diff
@@ -141,6 +134,23 @@
...
+::
+++  update-to-json
+  |=  upd=chat-update
+  =,  enjs:format
+  ^-  json
+  %+  frond  %chat-update
+  %-  pairs
+  :~
+    ?:  =(%message -.upd)
+      ?>  ?=(%message -.upd)
```
Fang-: Maybe I missed this before, but I'm curious: isn't just `?:  ?=(%message -.upd)` practically equivalent here? I don't recognize any TMI-sensitive situations in the code below.

loganallenc: Yes, it is equivalent. I'll make a note to go modify all of the JSON marks with that expression rather than the ?>  expression. (it's used across a variety of marks that i've written)

loganallenc:
@jtobin This is ready to merge. Thanks!

PR #1956, opened 2019-11-15
vere: consistent use of my-planet and my-comet in Simple Usage

jfranklin9000:


loganallenc:
Reasonable. Good cleanup!

PR #1957, opened 2019-11-15
zuse, build: remove default testnet config

jtobin:
Returns the target %zuse contract configuration to mainnet, and also tweaks the 'arvo-ropsten' build to use %alef instead of %ames.

@belisarius222 I have some existing scripting for handling testnets such that one doesn't need to hard-code any testnet config in git.  It's desirable to patch things here so I can simply reuse that stuff.

### DIFF
--- a/nix/ops/default.nix
+++ b/nix/ops/default.nix
@@ -67,14 +67,6 @@ rec {
     pier = zod;
   };
 
-<<<<<<< HEAD
-  image = import ./image {
-    inherit pkgs urbit;
-    pill = bootsolid;
-  };
-
-||||||| merged common ancestors
-=======
   ivory-ropsten = import ./ivory {
     inherit pkgs herb urbit;
     arvo = arvo-ropsten;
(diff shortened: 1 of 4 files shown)


nix/ops/default.nix
```diff
@@ -67,14 +67,6 @@ rec {
     pier = zod;
   };
 
-<<<<<<< HEAD
```
belisarius222: Oh whoops, good catch.  Sorry about that

jtobin:
On Fri, Nov 15, 2019 at 11:23:29AM -0800, Ted Blackman wrote:
> LGTM.  Using a regex to change the ames protocol version seems a
> little risky, but we were apparently already doing it that way.

Agreed, but it works for the time being.  Perhaps I could create a
proper patchfile and apply that (some bells and whistles would then at
least fire if the patch didn't apply), but it seems I'll still have to
tweak the thing dynamically based on the protocol version present in the
source.  Likely a bit safer, though.

PR #1966, opened 2019-11-17
hoon: refactor royl float parsers to separate arms

pkova:
Immediately useful for implemeting json `@rd` parsing, which is basically
`++royl-rd` minus pfix sig. The increased separation also allows for running
stuff like `(rash '3.22e-47' royl-rn:so)` from the dojo.

Not sure who should look at this, ping @jtobin.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -6124,47 +6124,54 @@
     ==
   ++  royl
     ~+
-    =+  ^=  moo
-      |=  a/tape
+    ;~  pose
+      (stag %rh royl-rh)
+      (stag %rq royl-rq)
+      (stag %rd royl-rd)
+      (stag %rs royl-rs)
+    ==
+  ::
+  ++  royl-rh  (cook rylh ;~(pfix ;~(plug sig sig) (cook royl-cell royl-rn)))
+  ++  royl-rq  (cook rylq ;~(pfix ;~(plug sig sig sig) (cook royl-cell royl-rn)))
+  ++  royl-rd  (cook ryld ;~(pfix sig (cook royl-cell royl-rn)))
+  ++  royl-rs  (cook ryls (cook royl-cell royl-rn))
+  ::
+  ++  royl-rn
+    =/  moo
+      |=  a=tape
       :-  (lent a)
       (scan a (bass 10 (plus sid:ab)))
-    =+  ^=  voy
-      %+  cook  royl-cell
-      ;~  pose
-        ;~  plug
-          (easy %d)
-          ;~  pose  (cold | hep)  (easy &)  ==
-          ;~  plug  dim:ag
-            ;~  pose
-              ;~(pfix dot (cook moo (plus (shim '0' '9'))))
-              (easy [0 0])
-            ==
-            ;~  pose
-              ;~  pfix
-                (just 'e')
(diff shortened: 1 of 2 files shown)


jtobin:
*cracks knuckles*

Is the original caching behaviour still preserved here?  It looks like `~+` previously caused `moo` and `voy` to be retained locally in the subject, and presumably that was useful for backtracking, but now it doesn't appear to be the case.

(I'd better not get too ahead of myself here and tag @philipcmonk or @Fang- for a proper review instead.. :smile:)

pkg/arvo/sys/hoon.hoon
```diff
@@ -6124,47 +6124,52 @@
...
-              ;~  pfix
-                (just 'e')
-                ;~(plug ;~(pose (cold | hep) (easy &)) dim:ag)
-              ==
-              (easy [& 0])
-            ==
+    ;~  pose
+      ;~  plug
+        (easy %d)
+        ;~  pose  (cold | hep)  (easy &)  ==
```
Fang-:
```suggestion
        ;~(pose (cold | hep) (easy &))
```

pkg/arvo/sys/hoon.hoon
```diff
@@ -6124,47 +6124,52 @@
...
+              ;~(plug ;~(pose (cold | hep) (easy &)) dim:ag)
+            ==
+            (easy [& 0])
           ==
         ==
-        ;~  plug
-          (easy %n)
-          (cold ~ (jest 'nan'))
+      ==
+      ;~  plug
```
Fang-: You'll want to put in some `::` to more clearly separate these parts of the parser (here and further down), since they're all big and multi-line. ie:

```hoon
      ==
    ::
      ;~  plug
```

pkg/arvo/sys/hoon.hoon
```diff
@@ -6124,47 +6124,52 @@
...
+      (stag %rs royl-rs)
+    ==
+  ::
+  ++  royl-rh  (cook rylh ;~(pfix ;~(plug sig sig) (cook royl-cell royl-rn)))
+  ++  royl-rq  (cook rylq ;~(pfix ;~(plug sig sig sig) (cook royl-cell royl-rn)))
+  ++  royl-rd  (cook ryld ;~(pfix sig (cook royl-cell royl-rn)))
+  ++  royl-rs  (cook ryls (cook royl-cell royl-rn))
+  ::
+  ++  royl-rn
     =+  ^=  moo
```
Fang-:
```suggestion
    =/  moo
```

pkg/arvo/sys/hoon.hoon
```diff
@@ -6124,47 +6124,52 @@
...
       |=  a/tape
```
Fang-:
```suggestion
      |=  a=tape
```

pkova:
Fixed the styling stuff!

joemfb:
The `~+`  memoization doesn't preserve the incremental computations, but the product of pulling the arm, key'd by the subject and formula. In this case, it's caching the construction of the parser itself, not the parsing. This refactoring should be strictly equivalent from a cache standpoint.

jtobin:
Whoops -- I forgot to update your remote before merging, so I got the old commit.  I'll revert and apply the updated one.

PR #1969, opened 2019-11-19
Modernise syntax throughout Arvo type definitions

jtobin:
Minor housekeeping on arvo.hoon and related.

Changes:

* The u3 spec was out of date, describing a since-removed +keep arm as a member of the Arvo structural core.  Additionally, the hardcoded formula axes were out-of-sync with the current ones.

* Fixes the deprecated comment style that existed in Arvo's header, rewriting all comments verbatim in the modern "breathing" style.

* Modernises syntax in Arvo's type definitions.  That is:
  * replaces old syntax (kel, ker, fas) with modern (sel, ser, tis)
  * replaces all `++` arms with `+$` or `+*` as appropriate
  * changes instances of `$foo` to `%foo`

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -1,85 +1,83 @@
-::::::  ::::::::::::::::::::::::::::::::::::::::::::::::::::::
-::::::  ::::::    Postface                              ::::::
-::::::  ::::::::::::::::::::::::::::::::::::::::::::::::::::::
+::  postface
+::
 ~>  %slog.[0 leaf+"%arvo-assembly"]
 =-  ~>  %slog.[0 leaf+"%arvo-assembled"]
     -
-=<  ::
-    ::  Arvo formal interface
-    ::
-    ::    this lifecycle wrapper makes the arvo door (multi-armed core)
-    ::    look like a gate (function or single-armed core), to fit
-    ::    urbit's formal lifecycle function. a practical interpreter
-    ::    can ignore it.
-    ::
-    |=  [now=@da ovo=*]
-    ^-  *
-    ~>  %slog.[0 leaf+"arvo-event"]
-    .(+> +:(poke now ovo))
-::::::  ::::::::::::::::::::::::::::::::::::::::::::::::::::::
-::::::  ::::::    volume 3, Arvo models and skeleton    ::::::
-::::::  ::::::::::::::::::::::::::::::::::::::::::::::::::::::
+=<
+::  arvo formal interface
+::
+::    this lifecycle wrapper makes the arvo door look like a gate to fit
+::    urbit's formal lifecycle function. a practical interpreter
+::    can ignore it.
+::
+|=  [now=@da ovo=*]
+^-  *
+~>  %slog.[0 leaf+"arvo-event"]
+.(+> +:(poke now ovo))
+::
 =>
+::  volume 3, arvo models and skeleton
+::
 |%
-++  arch  {fil/(unit @uvI) dir/(map @ta $~)}            ::  fundamental node
-++  arvo  (wind {p/term q/mill} mill)                   ::  arvo card
-++  beam  {{p/ship q/desk r/case} s/path}               ::  global name
-++  beak  {p/ship q/desk r/case}                        ::  garnish with beak
-++  bone  @ud                                           ::  opaque duct
-++  case                                                ::  version
-          $%  {$da p/@da}                               ::  date
-              {$tas p/@tas}                             ::  label
-              {$ud p/@ud}                               ::  sequence
-          ==                                            ::
-++  desk  @tas                                          ::  ship desk case spur
-++  dock  (pair @p term)                                ::  message target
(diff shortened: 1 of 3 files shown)


pkg/arvo/sys/arvo.hoon
```diff
@@ -1,85 +1,177 @@
...
-    ::    urbit's formal lifecycle function. a practical interpreter
-    ::    can ignore it.
-    ::
-    |=  [now=@da ovo=*]
-    ^-  *
-    ~>  %slog.[0 leaf+"arvo-event"]
-    .(+> +:(poke now ovo))
-::::::  ::::::::::::::::::::::::::::::::::::::::::::::::::::::
-::::::  ::::::    volume 3, Arvo models and skeleton    ::::::
-::::::  ::::::::::::::::::::::::::::::::::::::::::::::::::::::
```
philipcmonk: Isn't the "volume" reference part of a larger organization scheme for hoon/zuse/arvo?

pkg/arvo/sys/arvo.hoon
```diff
@@ -1,85 +1,177 @@
...
-++  wind                                                ::  new kernel action
-          |*  {a/mold b/mold}                           ::  forward+reverse
-          $%  {$pass p/path q/a}                        ::  advance
-              {$slip p/a}                               ::  lateral
-              {$give p/b}                               ::  retreat
-          ==                                            ::
-++  wire  path                                          ::  event pretext
+::  +vile: reflexive constants
+::
++$  vile
```
philipcmonk: As long as there's not tons of items, I find this style excessively verbose compared to:

```
::  +vile: reflexive constants
::
::    typ: -:!>(*type)
::    duc: -:!>(*duct)
::    pah: -:!>(*path)
::    mev: -:!>([%meta *vase])
::
$:  typ=type
    duc=type
    pah=type
    mev=type
==
```

pkg/arvo/sys/arvo.hoon
```diff
@@ -1,85 +1,177 @@
...
-          (pair cord (each noun (list mash)))           ::
-++  mill  (each vase milt)                              ::  vase+metavase
-++  milt  {p/* q/*}                                     ::  metavase
-++  monk  (each ship {p/@tas q/@ta})                    ::  general identity
-++  muse  {p/@tas q/duct r/arvo s/@ud}                  ::  sourced move
-++  move  {p/duct q/arvo}                               ::  arvo move
-++  ovum  {p/wire q/curd}                               ::  typeless ovum
-++  pane  (list {p/@tas q/vase})                        ::  kernel modules
-++  pone  (list {p/@tas q/vise})                        ::  kernel modules old
+::  +arch: fundamental node
```
philipcmonk: It seems to me that the long-form comment style is helpful because it allows longer descriptions, not for its own sake.  The previous version of this is significantly more readable to me (comment-wise; the syntax modernizations are righteous).

jtobin: I do think all of these things deserve longer-form commentary.  E.g. `+arch` -- "fundamental node" is uninformative, to me, whereas some longer exposition on it being a blob ref / directory listing would be more appropriate.  This struck me as moving partway towards that goal.

However, maybe it'd be better to add more useful comments first, *then* modernise their style..

pkg/arvo/sys/arvo.hoon
jtobin: If it is, I'm afraid this is the only surviving work. 😄

loganallenc:
Praise be! Next step: @pilfer-pandex updating the language to remove the old syntax such that it isn’t valid Hoon.

pkg/arvo/sys/arvo.hoon
jtobin: Having ruminated on this a little more: I do think the breathing comment style is an improvement, even absent any additional meat in the actual comments.  But it's certainly a controversial change, so I'll revert it for the time being.

pkg/arvo/sys/arvo.hoon
jtobin: Resolving as I've reverted the breathing comments for now.

pkg/arvo/sys/arvo.hoon
jtobin: I've retained the "volume 3" notice for now (though as I mention, other volumes seem to be MIA, or are implicit).

pkg/arvo/sys/arvo.hoon
```diff
@@ -1,85 +1,83 @@
...
+::
+::    this lifecycle wrapper makes the arvo door look like a gate to fit
+::    urbit's formal lifecycle function. a practical interpreter
+::    can ignore it.
+::
+|=  [now=@da ovo=*]
+^-  *
+~>  %slog.[0 leaf+"arvo-event"]
+.(+> +:(poke now ovo))
+::
```
joemfb: The former indentation of this block is idiomatic, and should be restored.

jtobin:
I'm going to close this one -- I may recycle the `u3` spec edits, but I think the other stuff is both too minor & contentious to bother with here.

Issue #1904, opened 2019-11-01
eyre: login without redirect, "not redirecting properly"


Fang-:
You won't naturally end up in this situation, but if you manually navigate to `/~/login` (as opposed to, say , `/~/login?redirect=/`), and then log in, it'll give you the following:

<img width="878" alt="image" src="https://user-images.githubusercontent.com/3829764/68029984-b6a59e00-fcb8-11e9-86fa-21d0fb1f4eb3.png">

This should either default to redirecting to `/`, or simply close the page on-login if no redirect is specified.

loganallenc:
:( agreed on defaulting to redirect to /

philipcmonk:
If you go to `/` then put in the wrong code, then put in the right one, it seems to redirect to a blank page.

matildepark:
Both of these issues are fixed! See #3000.

PR #1970, opened 2019-11-19
Link: social bookmarking

Fang-:
Draft because remote comms are untested, and `link-server-hook` + web extension are still pending. What's in the PR right now should be ready for review regardless.

I haven't been able to test foreign-ship interactions with this yet because I ran into [an ames crash](https://github.com/urbit/urbit/blob/1c398594c7f6bde87b51ecc0e399409364779e25/pkg/arvo/sys/vane/ames.hoon#L805). This also happens for regular `|hi`s so @philipcmonk I guess this means I'm waiting on alef integration?


joemfb:
@Fang-, I suspect that the %gall refactor is implicated in this issue. This is the same crash as #1831? I've been meaning to look into it, but haven't yet.

Fang-:
@joemfb Yes, seems to be the same crash.

pkg/arvo/app/link-store.hoon
```diff
@@ -0,0 +1,172 @@
...
+::    /local-pages/[some-group]    all pages we saved by recency
+::    /submissions/[some-group]    all submissions by recency
+::
+/+  *link, default-agent, verb
+::
+|%
++$  state-0
+  $:  %0
+      by-group=(map path links)
+      by-site=(map site (list [path submission]))
```
loganallenc: looks good. these two types of indices / keys to query on make sense

pkg/arvo/gen/group-store/remove.hoon
```diff
@@ -0,0 +1,10 @@
+::  group-store|remove: remove members from a group
+::
+/-  *group-store
+:-  %say
+|=  $:  [now=@da eny=@uvJ =beak]
+        [[=path members=(list ship) ~] ~]
+    ==
+:-  %group-action
```
loganallenc: great additions! love these generators. Could you put them under the folder `groups` instead of `group-store`? It would give an interface akin to the publish interface you put forward. Thanks!

pkg/arvo/app/link-store.hoon
```diff
@@ -0,0 +1,172 @@
...
+::
+::  scry and subscription paths:
+::
+::    /local-pages/[some-group]    all pages we saved by recency
+::    /submissions/[some-group]    all submissions by recency
+::
+/+  *link, default-agent, verb
+::
+|%
++$  state-0
```
loganallenc: why name it 0 instead of zero? curious, since it seems uncommon for us to include numbers in variable names

pkg/arvo/app/link-store.hoon
```diff
@@ -0,0 +1,172 @@
...
+::
+=|  state-0
+=*  state  -
+::
+%+  verb  &
+^-  agent:gall
+=<
+  |_  =bowl:gall
+  +*  this  .
+      do    ~(. +> bowl)
```
loganallenc: do? Perhaps this could be better named.

pkg/arvo/app/link-store.hoon
```diff
@@ -0,0 +1,172 @@
...
+    ^-  (quip card _this)
+    [~ this(state !<(state-0 old))]
+  ::
+  ++  on-poke
+    |=  [=mark =vase]
+    ^-  (quip card _this)
+    ?>  (team:title [our src]:bowl)  ::TODO  /lib/store
+    =^  cards  state
+      ?+  mark  (on-poke:def mark vase)
+        %json         (do-action:do (action:de-json !<(json vase)))
```
loganallenc: beautiful

pkg/arvo/app/link-store.hoon
Fang-: imo using words in place of digits for versioning is terrible. `state-0` is in line with its tag, `%0`, and also happens to keep width-consistency with `-1` through `-9`.

pkg/arvo/gen/group-store/remove.hoon
Fang-: I would if I could, but `%say` generators send their output as a poke to the app based on the path they're in. I can't put it into `groups` because there is no `/app/groups` to catch the resulting poke.

pkg/arvo/app/link-store.hoon
Fang-: Note that querying by site isn't _actually_ supported yet in the scry/peer paths, but trivial to put in once we do need it. Leaving it out for now as nothing uses it.

Fang-:
[Here's a demo video](https://pal.dev/share/tlon/link-demo.mp4) that shows:
- being prompted to configure the web extension
- being prompted to login (and subsequent redirect fail as per #1904, really want this page to just self-close)
- saving a page (to `/private`) using the web extension
- fetching a json containing pages we saved to `/private`
- scrying for all submissions recorded under `/private`

That `/private` path is just a temporary solution while the web extension is still super dumb. It can of course be any arbitrary path.

cc @vvisigoth @urcades

pkg/arvo/app/link-store.hoon
Fang-: As in the verb. My general structuring of new-gall apps is that the agent core is just handling gall interface tasks, and then the core below it actually _does_ all the application logic. Hence `do`. I think it reads quite nicely in most cases, `get-local-pages:do` etc. Happy to listen to suggestions for better names, if you have them.

pkg/arvo/app/link-store.hoon
philipcmonk: Personally I generally use words for numbers when there's only a particular set of them (where it's really an enum) and numerals where they're actually numbers (like here where they can go on forever).

pkg/arvo/app/link-store.hoon
philipcmonk: Another option is liberal use of |^.  For example, +do-action, +add-page, and +hear-submission could be in a |^ of on-poke.  The other two are used in both +on-peek and +on-watch, though.  You have to weigh the advantages of tighter scoping and easier state management against the reduced flexibility if you later want to call the function from another arm.

pkg/arvo/app/link-proxy-hook.hoon
```diff
@@ -0,0 +1,231 @@
...
+::    whatever's returned by the scry at that path, but perhaps that should
+::    become part of the stores standard anyway.
+::
+/-  *link, group-store
+/+  default-agent, verb
+|%
++$  state-0
+  $:  %0
+      ::TODO  we use this to detect "first sub started" and "last sub left",
+      ::      but can't we use [wex sup]:bowl for that?
```
philipcmonk: Spider depends on depends on `wex` and `sup` being correct, so those should work fine.

Fang-:
Note for @jtobin: some work still pending, please hold.

jtobin:
On Wed, Nov 20, 2019 at 01:55:08AM -0800, Fang wrote:
> Note for @jtobin: some work still pending, please hold.

This is a PR to philip/mall-real, so you're better off warning
@philipcmonk. ;-)

pkg/arvo/app/link-listen-hook.hoon
```diff
@@ -62,11 +62,20 @@
...
   ::
+  ++  on-arvo
+    |=  [=wire =sign-arvo]
+    ^-  (quip card _this)
+    ?.  ?=([%g %done *] sign-arvo)
+      (on-arvo:def wire sign-arvo)
+    ?~  error.sign-arvo  [~ this]
+    =/  =tank  leaf+"{(trip dap.bowl)}'s message went wrong!"
+    %-  (slog tank tang.u.error.sign-arvo)
+    [~ this]
```
Fang-: @philipcmonk I was kind of surprised to need to implement this (to catch `[%g %done ...]`), would've expected network messages to apps ack via `+on-agent` instead...

philipcmonk:
Uh, yeah, that looks weird. How are you getting a %done, especially from
%gall? Any network acks should come through on-agent.

Fang-:
idk man, I'm asking you. (^: I don't think I'm doing anything special here.

For what it's worth, the `verb` output for this was:

```hoon
%link-listen-hook: on-arvo on wire /out/~wanzod/link-proxy-hook/links/~wanzod/some/path, [%g %done]
```

where `/links/~wanzod/some/path` is the wire used internally by link-listen-hook.

pkg/arvo/gen/group-store/remove.hoon
joemfb: `:foo|bar` is short for `:foo +foo/bar`, so you could, but it'd be awkward.

philipcmonk:
Yeah, somehow an ack is being triggered that's supposed to go to ames (if the ack is supposed to happen at all) but is being routed to an app.  Probably a bad interaction between alef's new acks and me getting rid of bones.  I'll find and fix that.

philipcmonk:
What's an easy way to reproduce this?

Fang-:
> What's an easy way to reproduce this?

Hmm, so I tried getting you simple reproduction steps, but... I can't actually reproduce this anymore? I can't trigger an `+on-arvo` call in the flow that was causing it previously.

Perhaps I had something stale in my local environment? Would that even make sense here? I'll remove the `+on-arvo` implementation for now, and keep an eye for this problem going forward.

pkg/arvo/app/link-server-hook.hoon
```diff
@@ -0,0 +1,230 @@
...
+++  handle-http-request
+  |=  [eyre-id=@ta =inbound-request:eyre]
+  ^-  (list card)
+  ::NOTE  we don't use +require-authorization because it's too restrictive
+  ::      on the flow we want here.
+  ::
+  ?.  ?&  authenticated.inbound-request
+          =(src.bowl our.bowl)
+      ==
+    ::TODO  `*octs -> ~ everywhere once no-data bug is fixed
```
Fang-: Note to self: this is now fixed on latest `philip/mall-real`.

Fang-:
Just a heads up that I'm not sure we want this to go OTA yet. Not that it'd harm anyone, but it's not Ready yet either.

jtobin:
> Just a heads up that I'm not sure we want this to go OTA yet. Not that it'd harm anyone, but it's not Ready yet either.

Yeah, my bad.  I  should have dismissed the old review (left over from/approved for `philip/mall-real`) on this one.

PR #1973, opened 2019-11-20
soto: mounts at ~dojo

### LINKED ISSUE #1972: URL of web dojo is /~soto even though the tile is called Dojo
I recommend we call the web dojo url path /~dojo instead of /~soto for consistency. Thanks!

matildepark:
Closes #1972. 

Since Soto is the name of the relay, its location in Landscape should be
the application it relays to: Dojo. This commit amends the mount paths
and source paths accordingly.


jtobin:
@loganallenc ping

PR #1975, opened 2019-11-20
Convert more apps to static gall

Fang-:
This is based on and targets #1920.

Conversion of claz is still pending, but can go in as its own PR just fine.


ixv:
I just pushed a change to ixv/convert-apps that adds an id to sole-action, since soto needs to send them over the wire.

Fang-:
Gonna go ahead and mark this ready for review. Still need to do claz (and, uh, gaze I guess) but it's not strictly necessary for review. Can get its own PR if needed.

ixv:
The sole-action thing effects chat-cli, no?

pkg/arvo/app/pool-group-hook.hoon
```diff
@@ -0,0 +1,199 @@
...
+::
+++  leave-spider
+  |=  [=path]
+  ^-  card
+  [%pass [%running path] %agent [our.bowl %spider] %leave ~]
+::
+++  new-tid
+  |=  eny=@uv
+  ^-  @t
+  :((cury cat 3) dap.bowl '--' (scot %uv eny))
```
philipcmonk: I find it convenient for the `tid` to be a something that gets printed as a literal that can be typed in, for example for use with `+spider/tree` and `:spider|stop`.  If you `(scot %ta)` this then it'll work.  Also I've been using `_` instead of `--`, but that's less important.

pkg/arvo/app/pool-group-hook.hoon
```diff
@@ -0,0 +1,199 @@
...
+  ::
+  ++  on-poke   on-poke:def
+  ++  on-peek   on-peek:def
+  ++  on-watch  on-watch:def
+  ++  on-leave  on-leave:def
+  ++  on-fail   on-fail:def
+  --
+::
+|_  =bowl:gall
+++  five   5
```
ixv: debug code?

pkg/arvo/app/pool-group-hook.hoon
Fang-: Ah, that's all fair. Fix incoming!

pkg/arvo/app/pool-group-hook.hoon
Fang-: Oops yes.

pkg/arvo/app/pool-group-hook.hoon
```diff
@@ -141,7 +140,7 @@
 ++  new-tid
   |=  eny=@uv
   ^-  @t
-  :((cury cat 3) dap.bowl '--' (scot %uv eny))
```
philipcmonk: It needs to be `(scot %ta :((cury cat 3) dap.bowl '_' (scot %uv eny)))`.  The `0v1.` part is what doesn't parse.

Fang-: Huh, weird oops here. Thanks for catching. (:

philipcmonk:
I mucked around with rebase and cleaned up the history to "good enough".  Merged in 2e8b46836

PR #1978, opened 2019-11-21
zuse: add +name:title for (moon) identity

Fang-:
Just a little something to make "working with moons" easier.


jtobin:
You'll need to update the solid pill, my good man (and squash it into e1be585).

Issue #1664, opened 2019-08-25
Landscape builds should be improved


jtobin:
Presently `sh/build-interface` is underpowered; it assumes gulp and npm installations and also doesn't handle any installation of required npm packages on its own.  `make interface` should, in one command, install everything needed to compile Landscape's CSS and JS files -- and then compile them.

(Like our other builds, this is probably best handled via Nix.)

jtobin:
Re: conversation in #1918, the compiled JS/CSS files should not be renamed from the `-min.js` convention (and the associated apps should be patched to use these `-min.js` files).

matildepark:
We moved to Webpack and only require a single build command now, but we still don't do what OP suggests:

> in one command, install everything needed to compile Landscape's CSS and JS files -- and then compile them.

matildepark:
@jtobin How are you feeling about this? The build process has changed significantly, we don't ship compiled JS in git anymore, development workflows use a dev-server and deployment uses a CDN. Webpack is installed with the `npm install`, so the build command only assumes npm.

jtobin:
Definitely safe to retire this one. :+1:

PR #1980, opened 2019-11-22
soto: ignores backspace at position 0

matildepark:
@ixv caught this. The keypress handler would check if backspace was past position 0
before deleting one character at that position. If you were at position
0, then it would drop down to the 'else' and send the key value as a
string, throwing off the state between front and back end. This commit
fixes the issue, ignoring Backspace at position 0.


ixv:
LGTM

joemfb:
@matildepark, @jtobin, I just got `:soto` in the recent OTA, and this bug still present. It looks like this commit went into the update, so I'm not sure what went wrong ...

matildepark:
@joemfb Since there were multiple commits with built JS (but not each other's changes), presumably #1973 took precedence and the two commits weren't rebuilt for the OTA. Depends what @jtobin's build process is like these days (I recall he built the JS himself before deploying? It was a while ago, and we've changed a lot since then).

joemfb:
That makes sense. I also appear to be missing the tab completion.

matildepark:
That wasn't patched in, but it could point to me being wrong about the build thing (and the keycodes operating differently wherever you're on?). What's the browser and OS? I'll try and replicate this.

joemfb:
Chrome 77 on MacOS 10.14.

matildepark:
I can replicate … I’ll look into what’s going on and patch shortly. The rebuilt JS will help too — running from master only has the backspace commit’s built JS, not both.

— 
~haddef-sigwen
http://urbit.org

jtobin:
> Since there were multiple commits with built JS (but not each other's changes), presumably #1973 took precedence and the two commits together in a built JS format weren't pushed for the OTA. Depends what @jtobin's build process is like these days (I recall he built the JS himself before deploying? It was a while ago, and we've changed a lot since then).

Indeed, I recompile the JS before every release, so this *should* contain all the committed :soto changes.  I do remember running `make interface` a few times, but I had to tweak the release candidate a couple of times to drop some commits, so it's possible the updated compiled JS fell out.

(The release process is still an overly manual affair -- I am steadily automating things as I get the opportunity, but much remains to be automated.)

jtobin:
If it *is* missing I can hotfix it today.

jtobin:
If it helps diagnose the issue one way or the other, this is appearing in the compiled JS:

```
{if("Backspace"===e.key&&this.props.cursor>0)return qs.doEdit({del:this.props.cursor-1}),qs.setState({cursor:this.props.cursor-1}
```

matildepark:
@jtobin That code fragment wasn't what was patched here in #1980 — this PR adds "Backspace" as an ignored key so that if cursor is *at* 0, it gets caught and ignored instead of sent as a string.

I think this patch wasn't incorporated in the OTA, just #1973's changes. #1988 rebuilds everything and fixes the issue @joemfb brings up here.

jtobin:
@matildepark Aha.  I've figured this out. 

I just checked, and `make interface` on `v0.9` doesn't produce any different compiled JS for me.  So I did (attempt to) update it accordingly.  But, per #1664, `make interface` is really only half a build; it assumes that one has already `npm install`'d a bunch of stuff before running it.  I hadn't done that since `:soto` landed, so the relevant JS wasn't being compiled at all, and no errors were thrown.

I'll patch this up and hotfix things lickety-split.

jtobin:
(`:soto` also hadn't been added to `sh/build-interface`.  That'd also do it. 😄)

PR #1982, opened 2019-11-22
zuse: implement json number to @rd parsing

pkova:
As prepared by #1966.

Note that this is the first thing in the codebase to parse floats using something else than `++royl`, and is therefore the only place where the parser is not cached with `~+`. That may or may not be a problem, someone smarter than me could weigh in on this!

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:bccc640d030d2ec49b24a5f3413b305cb4823f1b5bf86f5d0dbfe7f6dc38ac2a
-size 8951763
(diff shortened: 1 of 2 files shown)


yosoyubik:
This is great! 

Wished I had this before; I'll have to replace all my `no:dejs:format` to `ne:dejs:format`  :D

Have you defined all fees, amounts... etc with `@rd` for the BTC grant?

pkova:
Tragically JSON allows for leading zeroes in exponents and therefore reusing the `@rd` parser does not work. Closing this, will need a different approach.

@yosoyubik I needed this for `difficulty` and `networkhashps` in `getmininginfo`. We should talk about the btc stuff at some point, I'll email you later.

PR #1984, opened 2019-11-22
u3: adds jam variants and jammed state management

joemfb:
This PR adds a new implementation of `+jam`, refactors the old implementation, adds operations to safely jam/cue the entire persistent state to/from a file, and automatically does in response to memory pressure. The primary motivation is to be more resilient to memory pressure, by deduplicating the entire state and recovering space for inner-road computations from our free lists (of which the current allocator makes particularly poor use.) These changes are factored to be broadly useful for future endeavors, such as jammed snapshots/state-exports, more efficient jam/cue, and cleaner patterns of recursive noun traversal.

The traversal from @frodwith's jam is factored out into `u3a_walk_fore()`, as we know have three separate uses for it. The current jam is preserved as `u3s_jam_fib()`, and still used in the jet and for IPC. All `+jam` operations now live in `serial.c`, as does `+cue`. The new operations are `u3s_jam_met()`, to calculate jam-bitwidth and back-references, and `u3s_jam_buf()`, to serialize into a buffer without incremental (or re-) allocation. `u3s_jam_file()` combines the preceding with `mmap()` and `msync()` calls to write the buffer into a file.

`u3a_walk_fore()` is named to anticipate an unimplemented `u3a_walk_hind()` or somesuch, which would implement that traversal pattern used in `u3a_take()`, `u3r_mug()`, and elsewhere. The key difference is that cell processing therein is primarily performed *after* sub-trees, whereas `+jam` requires only a reader pass and can cells *before* subtrees. There's also `u3a_walk_fore_unsafe()`, which elides road stack-overflow checks for performance, and is used by `u3s_jam_fib()`, preserving the current behavior. It's also *safely* used by `u3s_jam_buf()`, as the relevant checks were already performed in the preceding call to `u3s_jam_met()`.

I've added `u3m_rock_load()`/`u3m_rock_stay()`/`u3m_rock_drop()` to support saving and restoring the state; they respectively import, export, and delete a "rock" (jammed state export) from `.urb/roc/$EVT.jam`. `u3m_wipe()` throws away the contents of the loom and then reinitializes it from scratch. Since `+cue` performs many ephemeral allocations (and, as previously mentioned, we make poor use of our free-lists), `u3m_rock_load()` cues the rock in an inner road.

`u3e_hold()`/`u3e_drop()`/`u3e_fall()` respectively backup the memory-image checkpoint, discard the backup, and restore the backup (in case something went wrong in the intervening period). `u3e_wipe()`, currently unused, is an alternative approach: it just wipes the memory-image files in place.

To support automatically "packing" the state in response to memory-pressure, I've refactored event success/failure handlers in the worker. We pack on `bail:meme`, or after any event that triggers a `%trim` memory-pressure notification. There are lots of ways this could be improved: we could pack *after* the `%trim` event is processed; we could pack and then retry events that `bail:meme`; we could trim and then pack after `bail:meme`; &c. The efficiency of this automatic packing could certainly be improved as well. We write the "rock" into a file, then read the file, then delete file; we could just skip the file. But, for forward-looking reasons, I wanted the paths in and out of the file to exist and work. This is a very heavyweight operation, and it should be pretty rare.

There's a bunch of follow-on work related to these changes that should be done. Off the top of my head:

- use `u3a_walk_fore()` where suitable in other noun traversals
- write `u3a_walk_hind()`, use in `u3a_take()`, `u3r_mug()`, &c
- diagnose and fix our poor usage of free lists
- profile `u3s_jam_fib()` against the combination of `u3s_jam_met()` and `u3s_jam_buf()`, choose one or develop heuristics to switch between them
- write a bit-addressed cue without ephemeral allocation
- update jam/cue to use uint64_t for bit-indexing (requires a new u3r_chop or equivalent), as we can currently only jam/cue ~536MB serialized
- add an effect to trigger memory packing; allow the daemon to ask the worker to pack
- pack and retry after bail:meme
- make pack more efficient (at least by skipping the file)
- truncate the event log in coordination with a "rock"

I forced this to run on (a local-only copy of) ~master-morzod:

- before: https://pastebin.com/wgp8K59J
- after: https://pastebin.com/8kmtG1Je

Some noteworthy changes:

- overall memory usage from 262MB to 175MB
- free-lists from 179MB to 33MB
- the `|mass` payload itself ("total userspace") from 218MB to 150MB
- %clay from 113MB to 85MB
- %ford from 25MB to 16MB
- %gall from 31MB to 16MB
- %jael from 170MB to 119MB

I have not yet run this on a ship with much more significant memory usage, but I intend to do so. I don't have an estimate for how much stack/heap the first jam traversal consumes, but it's certainly possible for a loom to be so full that we can't perform this operation.

Finally, these changes are compatible with the live network, and should go out as a hotfix release.

/cc @frodwith, @benjamin-tlon


philipcmonk:
Nice work!  Is there any easy way to force this to run?

> %jael from 170MB to 119MB

Actually that's all the vanes.  %jael goes from 723KB to 616KB

joemfb:
Ah, good catch, thanks.

There's no way to trigger it right now, but it'd be pretty easy to add `|pack`. Happy to do so if there's interest.

philipcmonk:
As mentioned out-of-band, booting a fake ~bud, running ~pack, and restarting it makes the daemon think it's ~zod and also forces `u3_Host.ops_u.net` to `c3n` even if you don't include `-L`.

joemfb:
@philipcmonk, I've fixed this identity issue on `|pack`.

The pier identity was being retrieved from both the database "header" and the serf. The serf used to need to have identity and a fake-bit, but that was made obsolete by the "header" we setup in LMDB. I was indecisive about removing that data from the serf and the IPC protocol and left it in at the time. I've since become convinced that it's a layer violation, but was waiting to replace it until there was time to reform that protocol more generally. This bug is a motivating event, as I don't want to continue propagating that layer violation (into the jammed "rock", for instance).

I've pushed two commits, the first ignores/defaults the "single-home" data sent in the IPC protocol, but leaves the structures intact. The second removes the "single-home" data from the protocol itself. This protocol is an internal convention, with no other (public users), so I think it's ok to just change it incrementally. If others disagree, I can revert to the defaulted info.

/cc @eglaysher

joemfb:
I had thought the identity data was simply being ignored in `pier.c`, and, when I tested these changes on a copy of ~master-morzod, I failed to notice this:

```
$ ../urbit/pkg/urbit/build/urbit -L messsing-with-morzod/
~
urbit 0.9.2
...
work: play 359812
ames: czar: ~zod on 31337 (localhost only)
```

philipcmonk:
Does this fix the issue where it assumes -L? Maybe that was happening
because the fake bit defaulted to true?

joemfb:
Yeah, that's what it was:

https://github.com/urbit/urbit/blob/master/pkg/urbit/vere/pier.c#L1349-L1353

philipcmonk:
The test mentioned before (boot fakebud, reboot) still resets our identity to ~zod.

I've merged this with `rc` on the `philip/merg` branch.  The fix shouldn't affect that, but consider working off that (and whatever you do, don't just rebase the fix here).

philipcmonk:
I've merged this into `rc`, so it'll be marked as merged when `rc` lands on master.

PR #1991, opened 2019-11-26
zuse: implement json number to @rd parsing

pkova:
Like #1982 but better.

This parser is not cached with `~+`, if that matters let me know.

@jtobin

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -5453,6 +5453,39 @@
     |=  cis/coin
     ?.  ?=({$$ case} cis)  ~
     `(unit beam)`[~ [who dex `case`p.cis] (flop t.t.t.pax)]
+  ::
+  ++  json-rn                                           ::  json to rn parser
+    %+  knee  *rn  |.
+    ;~  plug
+      (easy %d)
+      ;~(pose (cold | hep) (easy &))
+      ;~  plug  dim:ag
+        ;~  pose
+          ;~  pfix  dot
+            %+  sear
+              |=  a=tape
+              =/  b  (rust a dum:ag)
+              ?~  b  ~
+              (some [(lent a) u.b])
+            (plus (shim '0' '9'))
+          ==
+          (easy [0 0])
+        ==
+        ;~  pose
+          ;~  pfix
+            (just 'e')
+            ;~  plug
+              ;~(pose (cold | hep) (easy &))
+              ;~  pose
+                ;~(pfix (plus (just '0')) dim:ag)
+                dim:ag
+              ==
+            ==
+          ==
+          (easy [& 0])
+        ==
+      ==
+    ==
   ::                                                    ::  ++enjs:format
   ++  enjs  ^?                                          ::  json encoders
     |%
@@ -5571,9 +5604,9 @@
     ::                                                  ::  ++ne:dejs:format
     ++  ne                                              ::  number as real
       |=  jon/json
-      ^-  (unit @rd)
-      ::  please implement me, it's not that hard!
-      !!
+      ^-  @rd
+      ?>  ?=([%n *] jon)
+      (rash p.jon (cook ryld (cook royl-cell:^so json-rn)))
     ::                                                  ::  ++ni:dejs:format
     ++  ni                                              ::  number as integer
       |=  jon/json
@@ -5784,7 +5817,8 @@
     ++  ne                                              ::  number as real
       |=  jon/json
       ^-  (unit @rd)
-      !!
+      ?.  ?=([%n *] jon)  ~
+      (rush p.jon (cook ryld (cook royl-cell:^so json-rn)))
     ::
     ++  ni                                              ::  number as integer
       |=  jon/json
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d0d7e3807c3159b1483220caf14fc2b96af65c05f2f761c4351c8614dc7db3cf
-size 9634290
+oid sha256:17e7708b65df64f3d8a69f360b5eccdb96dec87dd1e1c60658be834b40f066ce
+size 9627978


pkg/arvo/sys/zuse.hoon
```diff
@@ -5453,6 +5453,35 @@
     |=  cis/coin
     ?.  ?=({$$ case} cis)  ~
     `(unit beam)`[~ [who dex `case`p.cis] (flop t.t.t.pax)]
+  ::
+  ++  json-rn                                           ::  json to rn parser
```
pkova: It's kinda silly to put this parser under `++format`, if someone figures out a better place to put it let me know. Doesn't feel great to duplicate it under ´ne:dejs-soft:format` and `ne:dejs:format` either.

pkg/arvo/sys/zuse.hoon
```diff
@@ -5453,6 +5453,35 @@
...
+    ;~  plug
+      (easy %d)
+      ;~(pose (cold | hep) (easy &))
+      ;~  plug  dim:ag
+        ;~  pose
+          ;~  pfix  dot
+            %+  cook
+            |=(a=tape :-((lent a) (scan a dum:ag)))
+            (plus (shim '0' '9'))
```
joemfb: A parser shouldn't crash, so this should be `+sear` and `+rust` instead of `+cook` and `+scan`. And `%+` should use backstep indentation:

```
%+  sear
  |=(a=tape [(lent a) (rust a dum:ag)])
(plus (shim '0' '9'))
```

pkg/arvo/sys/zuse.hoon
```diff
@@ -5453,6 +5453,35 @@
     |=  cis/coin
     ?.  ?=({$$ case} cis)  ~
     `(unit beam)`[~ [who dex `case`p.cis] (flop t.t.t.pax)]
+  ::
+  ++  json-rn                                           ::  json to rn parser
+    ;~  plug
```
joemfb: We haven't done this consistently, but it's nice to explicitly cast a parser so it's obvious what it produces. A parser is a wet gate from `+nail` to `+edge`, and `+like` is a mold-builder for a specific type of `+edge`. If I'm reading this correctly, it should be:

```
^-  $-(nail (like dn))
```

pkg/arvo/sys/zuse.hoon
```diff
@@ -5570,10 +5599,11 @@
       ?~(jon ~ (some (wit jon)))
     ::                                                  ::  ++ne:dejs:format
     ++  ne                                              ::  number as real
+      =,  ^so
```
joemfb: This looks to be unused.

pkg/arvo/sys/zuse.hoon
```diff
@@ -5782,9 +5812,11 @@
       ?~(jon (some ~) (bind (wit jon) some))
     ::
     ++  ne                                              ::  number as real
+      =,  ^so
```
joemfb: As does this.

pkg/arvo/sys/zuse.hoon
pkova: I'm getting %redo-match with your suggestion. Fixed the backstep indentation and dumb `:-` usage though.

My `scan` was inspired by `++royl-rn` in hoon.hoon
```
=/  moo
  |=  a=tape
  :-  (lent a)
  (scan a (bass 10 (plus sid:ab)))
```

pkg/arvo/sys/zuse.hoon
pkova: Fixed!

pkg/arvo/sys/zuse.hoon
pkova: It's used below, but only in one place so I removed this import.

pkg/arvo/sys/zuse.hoon
ohAitch: Doesn't `$-` break parser composition? If you can still plug the resulting parser that's great, otherwise iirc `++knee` is the conventional way to add this sort of annotation

pkg/arvo/sys/zuse.hoon
joemfb: Ah, *that* `+so`, I was picturing `+so:dejs:format` or somesuch... Regardless, I do think the full path is better in this case.

pkg/arvo/sys/zuse.hoon
joemfb: @ohAitch, you're right. I've used this convention in some of my own parsers, but only at the top level, where they were never being composed.

@pkova, sorry about this -- you can just delete the line, or replace it with

```
%+  knee  *rn  |.
```

pkg/arvo/sys/zuse.hoon
joemfb: Ah, that's 'cause my suggestion was dumb ... ;)

It should be this:

```
%+  sear
  |=  a=tape
  =/  b  (rust a dum:ag)
  ?~  b  ~
  (some [(lent a) u.b])
(plus (shim '0' '9'))
```

The pattern in `+royl-rn` is also wrong. Parsers shouldn't crashes, but `+scan` does.

pkova: Done!

pkg/arvo/sys/zuse.hoon
pkova: And done!

pkg/arvo/sys/zuse.hoon
ohAitch: I can see the argument, but note that `dum:ag` is basically `(bass 10 (plus (shim '0' '9')))`? So there aren't any actual inputs it crash on.

Ofc this redundancy kinda suggests a parser of
```hoon
=/  bass10  =|(a=@ |=(b=(list @) ?~(b a $(a (add (mul 10 a) i.b), b t.b))))
%+  cook
  |=(a=(list @u) [(lent a) (bass10 a)])
(plus din:ab)
```
for both. (This probably wasn't originally done bc the `bass` code exists as a combinator but not as a plain function. Except `++can` for bex'en I guess)

Alternately you could
```hoon
=/  prelen
  |=  [a=tape b=tape]
  ?: =(a b) 0
  ?~  a  ~|(%b-must-suffix-a !!)
  +($(a t.a))
=/  with-length
  |*  a=rule
  |=  b=nail
  =/  c  (a b)
  ?~  q.c  c
  c(p.q :_(p.q.c (prelen q.b q.q.c)))
::
(with-length dum:ag)
```
but that seems like rather overkill for two uses :)

joemfb: @ohAitch, there are various use-cases in userspace: my DER parser specifically, but really anything that parses to +octs or equivalent. So this would be worth doing at some point.

ohAitch: No complaint from me, it's definitely the version that best isolates parser implementation from declarative grammar specification~

PR #1993, opened 2019-11-27
Chat functionality: integration tests complete

loganallenc:
Chat applications have been converted to static gall style and have been integration tested.

Notably, chat-hook now has a %backlog path for subscribing to messages from the backlog and a %mailbox path for subscribing to ongoing updates.

I may have uncovered a bug with removing wires after negative %watch-acks. Everything else works great and has been integration tested with the chat front-end.

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -80,11 +80,14 @@
   ::
   ++  on-init
     ^-  (quip card _this)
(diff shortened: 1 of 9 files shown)


philipcmonk:
Oh, and the %watch-ack bug was real, and after I fixed it that uncovered a separate bug in alef's handling of negative acknowledgments, which has also been fixed.  When you get a chance, try that test again.

PR #1996, opened 2019-11-27
Release candidate

philipcmonk:
Fuller release notes will come later, but here's a release candidate for the next breach.  This primarily includes the twin efforts of "static gall" and "new ames".  Some highlights include:

- Ames extends duct mechanics across network, so you can "pass" and "give" over the network.
- Ames has framework for communicating backpressure.  Currently, the rule is to cancel subscriptions if "not responding; still trying" and at least 5 messages in queue, but we may tune this as necessary.
- Ames supports arbitrary-length negative acks (possible cause of OWC).
- Ames protocol simplified
- A new rune, `!<`, has been introduced.  It takes a mold and a vase, dynamically checks the type in the vase matches the mold, and produces the value typed with that type.  In other words, ``!<(some-type !>(`some-type`value))`` is just `value`.
- Gall apps made "static", which means they are just an instance of a particular type, namely `agent:gall` in sys/zuse.hoon.
- Because they're a static type, all gall apps have exactly 10 arms to handle the different sorts of input they might receive.  This is much more explicit.
- Due to being a static type, it's much easier to create library functions that augment apps.  For example, lib/verb is a function from `agent` to `agent` that wraps the agent with printfs describing every piece of input it receives.  It can be turned on or off with `:app +verb`.
- Bones have been removed from apps.
- Gall is about 1000 lines shorter.
- The concept of "threads" is introduced, replacing tapps/async.  These are temporary processes that operate within what Haskell would call the IO monad with a process tree.
- pH has been rewritten from an app to a library used by threads.  Thus, each pH test is now an individual thread.

There is some preliminary documentation on agents, threads, and ph here: 
[agent.txt](https://github.com/urbit/urbit/files/3896064/agent.txt)
[thread.txt](https://github.com/urbit/urbit/files/3896065/thread.txt)
[ph-static.txt](https://github.com/urbit/urbit/files/3896063/ph-static.txt)

Feel free to review a part at a time.  "New Ames" changes primarily ames.hoon, with additional changes in clay, jael, and gall to support the new interface.  Static gall has five main components: gall.hoon (+ap core rewritten), app/*.hoon (apps converted to the new style), app/spider.hoon (thread manager, replacing tapps/async), ted/*.hoon (threads, mostly converted from tapps), and ted/ph/*.hoon (new-style ph tests, which are just threads instead of a hand-rolled monad).

The only known issues are:
- ~~Comets aren't working correctly.  @belisarius222 is working on this and expects to solve it soon.~~
- ~~@Fang- has reported that eth-watcher isn't working correctly as of today, but this hasn't been nailed down.  It could be due to us upgrading our Ethereum nodes to be ready for Istanbul.~~ Determined to be a timing issue; it used to be safe to reduce the time between jael updates.
- ~~Backpressure was working, but a recent change broke it.  I believe it needs to add the trigger for checking for %dead when enqueueing a new message.~~ I believe this works correctly.
- ~~If you leave a chat room for a while and come back, it doesn't give you all the backlog because it only resubscribes to the "new posts" path.  It should retrieve all backlog since its last known post.  @loganallenc is working on this.~~


pkg/arvo/lib/hood/drum.hoon
```diff
@@ -80,24 +79,21 @@
...
         %eth-watcher
         %azimuth-tracker
+        %ping
     ==
+  ?:  lit
+    ~
   :~  %lens
+      %clock
       %dojo
       %modulo
```
ixv: %modulo app has been removed because its functionality was very minor and made sense to just fold it into %launch.

philipcmonk: thanks, removed

pkg/arvo/sys/arvo.hoon
```diff
@@ -693,26 +694,49 @@
               =^  rey  +>+  (^load +<)
               [rey +>.$]
     ::
-    ++  peek  |=(* (^peek ;;([@da path] +<)))      ::  46
+    ++  peek  |=  *                                     ::  46
+              =/  rob  (^peek ;;([@da path] +<))
+              ?~  rob  ~
+              ?~  u.rob  ~
+              [~ u.u.rob]
```
joemfb: I don't think this is right for the top-level `+peek` arm. At least, it's not obvious to me that the runtime *shouldn't* be able to distinguish between blocking and failing reads. But I see that it's consistent with what was here before. Filing away for later ...

pkg/arvo/sys/hoon.hoon
```diff
@@ -10619,6 +10637,10 @@
         {$zpmc *}
       =+  vos=$(gol %noun, gen q.gen)       ::  XX validate!
       [(nice (cell (play p.gen) p.vos)) (cell (play(sut dox) p.gen) q.vos)]
+    ::
+        {$zpgl *}
+      ::  XX is this right?
```
joemfb: It looks right

pkg/arvo/sys/hoon.hoon
```diff
@@ -6782,6 +6782,7 @@
   ::                                            ::::::  special
     {$zpcm p/hoon q/hoon}                               ::  !,
     {$zpbn p/hoon}                                      ::  !>
+    {$zpgl p/spec q/hoon}                               ::  !<
```
joemfb: This naming is inconsistent with the rest of hoon.hoon wrt pronounceables (ban/led vs gal/gar)

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -807,6 +808,7 @@
   ::  gate that produces the +per-server-event core from event information
   ::
   |=  [[our=@p eny=@ =duct now=@da scry=sley] state=server-state]
+  =/  eyre-id  (scot %ta (cat 3 'eyre_' (scot %uv (sham duct))))
```
joemfb: Noted elsewhere, but worth noting again: %eyre doesn't not currently ensure duct consistency for the lifetime of a channel. There's a good chance this will at least create duplicate subscriptions. If used as an id for an app with aa dojo-like interface, duct mismatches may also cause crashes.

pkg/urbit/worker/main.c
```diff
@@ -422,6 +439,7 @@ _worker_lame(c3_d evt_d, u3_noun now, u3_noun ovo, u3_noun why, u3_noun tan)
     u3_noun lef = u3nc(c3__leaf, u3kb_weld(u3i_tape("bail: "),
                                            u3qc_rip(3, why)));
     u3_noun nat = u3kb_weld(u3k(tan), u3nc(lef, u3_nul));
+    _worker_send_tang(1, u3k(nat));
```
joemfb: This and the above changes will result in un-hushable error output on every event replacement.

pkg/urbit/Makefile
```diff
@@ -45,6 +46,23 @@ mrproper: clean
 
 ################################################################################
 
+include/ca-bundle.h:
+	@echo XXD -i $(SSL_CERT_FILE)
+	@cat $(SSL_CERT_FILE) > include/ca-bundle.crt
+	@xxd -i include/ca-bundle.crt > include/ca-bundle.h
+	@rm include/ca-bundle.crt
```
joemfb: It looks like this was inadvertently added. This is the old approach, now handled through nix dependencies and the `./configure` script.

pkg/urbit/Makefile
```diff
@@ -45,6 +46,23 @@ mrproper: clean
...
+
+include/ivory.h:
+	@echo XXD -i $(IVORY)
+	@cat $(IVORY) > u3_Ivory.pill
+	@xxd -i u3_Ivory.pill > include/ivory.h
+	@rm u3_Ivory.pill
```
joemfb: Same as `include/ca-bundle.h:` above

nix/pkgs/arvo-ropsten/builder.sh
```diff
@@ -4,18 +4,18 @@ cp -r $src tmp
 chmod -R u+w tmp
 
 ZUSE=tmp/sys/zuse.hoon
-AMES=tmp/sys/vane/ames.hoon
+ALEF=tmp/sys/vane/alef.hoon
```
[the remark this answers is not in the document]
jtobin: Fixed in 25fbb5c.

pkg/arvo/lib/hood/helm.hoon
```diff
@@ -34,36 +34,23 @@
...
-          {$flog wire flog:dill}                        ::
-          [%knob wire @tas ?(%hush %soft %loud)]        ::
-          {$nuke wire ship}                             ::
-          [%serve wire binding:eyre generator:eyre]     ::
-          {$poke wire dock pear}                        ::
-          {$rest wire @da}                              ::
-          {$wait wire @da}                              ::
-          {$rekey wire life ring}                       ::
-          {$moon wire ship udiff:point:able:jael}       ::
-      ==                                                ::
```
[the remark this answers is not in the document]
philipcmonk: I do think it would be useful to have a pattern for doing this.  It shouldn't be too hard.

pkg/arvo/sys/arvo.hoon
philipcmonk: Yeah, I agree with that.  Worth revisiting when we start using peek more from the runtime.

pkg/arvo/sys/hoon.hoon
philipcmonk: I actually don't think it is quite right for the same reason that `!>` is currently mishandled in +mull.  Both of those should give a mull-bonk error if the actual type is not exactly the formal type.  We can't generate the correct `!<` nest check or `!>` +burp type polymorphically, since it depends on the actual type.

pkg/arvo/sys/vane/eyre.hoon
philipcmonk: This is only for handle-http-request (ie lib/server-style) pokes, so apps don't have access to the duct/bone anyway.  Subscriptions to apps use a separate interface that's handled entirely by eyre and is indistinguishable to the app from a normal subscription.  I assume eyre ensures duct consistency for these.

For example, soto serves only static html/js/css/png via this mechanism.  To maintain a session with dojo, it generates a random id and passes that along with the subscription and pokes:

https://github.com/urbit/urbit/blob/e3005eaffa31f269467a8da31c31f972c5367f1f/pkg/interface/soto/src/index.js#L9
https://github.com/urbit/urbit/blob/e3005eaffa31f269467a8da31c31f972c5367f1f/pkg/interface/soto/src/js/subscription.js#L15
https://github.com/urbit/urbit/blob/e3005eaffa31f269467a8da31c31f972c5367f1f/pkg/interface/soto/src/js/subscription.js#L27
https://github.com/urbit/urbit/blob/e3005eaffa31f269467a8da31c31f972c5367f1f/pkg/interface/soto/src/js/api.js#L33-L35

@matildepark fyi

pkg/urbit/worker/main.c
philipcmonk: These can just be removed, right @belisarius222 ?

pkg/arvo/sys/hoon.hoon
philipcmonk: you're right, I'll change to zpld.

pkg/arvo/sys/hoon.hoon
joemfb: So, a check that sut and dox mutually nest for both `!;` and `!<`? In a sense, these runes implicitly and silently dry out the core, whereas such a check would force the caller to do so.

belisarius222:
Yes, that's just a git wart

—
~rovnys-ricfer
https://urbit.org

pkg/arvo/sys/hoon.hoon
jfranklin9000: we want gal/gar not led/ban

joemfb: Yes, but we require consistency. The AST nodes should be migrated all at once; it would be confusing for them to simultaneously use two incompatible schemes

pkg/arvo/sys/arvo.hoon
```diff
@@ -693,26 +694,49 @@
...
+                    %0  p.res
+                    %1  $(res [%2 leaf+"blocked" >((list path) p.res)< ~])
+                    %2
+                  ~_  leaf+"deterministic error"
+                  |-
+                  ?~  p.res
+                    !!
+                  ~_  i.p.res
+                  $(p.res t.p.res)
+                ==
```
joemfb: @philipcmonk, on second thought, aren't these changes now superfluous? %gall virtualizes app invocations and sets the scry handler. This approach works to a) pin `.^` for the duration of an event and b) enables `.^` in vanes. The former is desirable, but only a partial solution to a pinned scry, while the latter is more debatable. It seems like this should be reverted.

philipcmonk: You're correct, I'll remove it.

pkg/urbit/include/noun/vortex.h
```diff
@@ -13,8 +13,8 @@
         u3_noun wen;                      //  current time, as text
         u3_noun sev_l;                    //  instance number
         u3_noun sen;                      //  instance string
-        u3_noun our;                      //  identity
-        u3_noun fak;                      //  c3y is fake
+        u3_noun our;                      //  identity     XX Remove
+        u3_noun fak;                      //  c3y is fake  XX Remove
```
joemfb: Since this is going into a breach, these members can now be removed from this struct. (Doing so in a hotfix would break the memory image.)

pkg/arvo/sys/zuse.hoon
```diff
@@ -1,3 +1,4 @@
+!:
```
joemfb: This should be removed, IMO

philipcmonk:
Both done

PR #2001, opened 2019-11-28
eth: two enhancements

Fang-:
Put these in while working on porting userspace Ethereum tools to static gall. Figured it might be nice to get these into the release candidate if possible.

- Add support for batch requests to `/lib/ethio`. If any request in the batch fails, the entire batch is considered a lost cause.
- Add support for custom refresh rates to `/app/eth-watcher`. Useful for low-priority watchdogs, or cases where you need a long timeout time to deal with big requests/responses.

### DIFF
--- a/pkg/arvo/app/azimuth-tracker.hoon
+++ b/pkg/arvo/app/azimuth-tracker.hoon
@@ -88,7 +88,7 @@
   ^-  card:agent:gall
   =/  args=vase  !>
     :*  %watch  /[dap]
-        url.state  launch:contracts:azimuth
+        ~m5  url.state  launch:contracts:azimuth
         ~[azimuth:contracts:azimuth]
         (topics whos.state)
     ==
--- a/pkg/arvo/app/eth-watcher.hoon
+++ b/pkg/arvo/app/eth-watcher.hoon
@@ -6,10 +6,6 @@
 =,  able:jael
 ::
 =>  |%
-    ++  refresh-rate  ~m5  ::  ~m5
-    --
-::
-=>  |%
     +$  card  card:agent:gall
     +$  app-state
       $:  %0
@@ -34,14 +30,14 @@
 ::
 =>  |%
     ++  wait
-      |=  now=@da
+      |=  [=path now=@da time=@dr]
       ^-  card
-      [%pass /timer %arvo %b %wait (add now refresh-rate)]
+      [%pass [%timer path] %arvo %b %wait (add now time)]
     ::
     ++  wait-shortcut
-      |=  now=@da
+      |=  [=path now=@da]
       ^-  card
-      [%pass /shortcut %arvo %b %wait now]
+      [%pass [%timer path] %arvo %b %wait now]
     ::
     ++  poke-spider
       |=  [=path our=@p =cage]
@@ -70,8 +66,7 @@
 ::
 ++  on-init
   ^-  (quip card _this)
-  ::  start update timer loop
-  [[(wait now.bowl) ~] this]
+  [~ this]
 ::
 ++  on-save   !>(state)
 ++  on-load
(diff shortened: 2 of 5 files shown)


philipcmonk:
I pushed this to the testnet OTA and everything seems fine.

pkg/arvo/app/azimuth-tracker.hoon
```diff
@@ -88,7 +88,7 @@
   ^-  card:agent:gall
   =/  args=vase  !>
     :*  %watch  /[dap]
-        url.state  launch:contracts:azimuth
+        ~m5  url.state  launch:contracts:azimuth
```
philipcmonk: This line was actually wrong, the refresh-rate goes after the url.

philipcmonk:
Some things I discovered while trying to run it through the ph tests:

- ph tests didn't support the batch requests
- azimuth-tracker was sending the wrong poke type
- timers didn't restart after the first time they fire
- when we detected a hung thread, we killed it but didn't start a new one

None of these are architectural issues, but two of them would have broken key propagation.  I think the lesson to learn here is that all changes to eth-watcher must be run through pH.  We should use pH tests for more of the system, but the one area where we actually have thorough pH tests is networking and breach-related tests, so let's never break those.

(the basic formula for running the pH tests is `|start %aqua` `:aqua +solid` `-ph-all`)

philipcmonk:
Oh, another lesson is that the `=;` construction to set a timer regardless of errors and the "loop without moving on to the next timer" construction were rather subtle and I should have commented them.

ohAitch:
How hard would it be for "must be run" to turn into "CI"? :3

philipcmonk:
The main thing is just that even the reduced list that's in `-ph-all` takes like 40 minutes to run.  Maybe worth it, though.  Of course, it could be parallelized, and parts of it could be cached better.

Fang-:
> * timers didn't restart after the first time they fire
> * when we detected a hung thread, we killed it but didn't start a new one

Oops, yes, I discovered this also. Your fix for the `%wake` case is more or less identical to what I implemented before noticing you had already pushed a fix of your own to `rc`. Sorry for the extra work here, and thank you for patching that up! (:

> another lesson is that the =; construction to set a timer regardless of errors and the "loop without moving on to the next timer" construction were rather subtle

I think the `=;` was clear enough, but I messed up making the transition from "one timer for everything" to "one timer per watchdog".

ohAitch:
Hm, 40 minutes seems tbh fine for an OTA; how would people feel about a
"release" branch that deploys are cut from and tracks master, but PRs to it
must pass more hardcore tests?
https://docs.travis-ci.com/user/conditional-builds-stages-jobs/#conditional-builds

Or you can rely on discipline ofc ;)

ohAitch:
Though as you say, nix caching e.g. "pier built from
lib/ph/runner-commit.txt" + "primed with a boot from the last commit that
touched hoon.hoon" + "further primed with a boot from the last commit that
touched arvo" might do wonders here. How close is the new compile cache to
making the second of two +solid's(in separate events) be <30 sec? (Might
also want to strap on a parse-cache, at least in ph land and specifically
for hoon.hoon "ride-parsed")

What all else is time-consuming, do you have a like -DHGETTO profile of the
various tests you could paste? <s>Does ames take forever to sync the fs to
children on network tests? Mining comets?</s>

jtobin:
> 40 minutes seems tbh fine for an OTA; how would people feel about PRs to  master  from  rc  having to pass more hardcore tests?

I've gotten a start on something like this.  Not in CI,  exactly, but more hardcore automated testing that should run before tagging a release.  I typically cherry-pick commits over to a release branch (`v0.8`, `v0.9`, etc.) locally and then tag a release, so what I want is a script that will wrap the tagging conditional on passing the meatier suite.

At present I put OTA-targeted releases through a basic ringer that tests the update applies cleanly, but I hope to add more to this, per @philipcmonk's commentary.

PR #2005, opened 2019-12-01
app: add types for btc-rpc control, generating, mining and network

pkova:
For part 2 (https://grants.urbit.org/bounties/1577544668) of the btc rpc grant.

@yosoyubik did a lot of heavy lifting on the plumbing for this, so I'm only including my `lib` and `sur` stuff for now.

These should eventually be merged on @yosoyubik's work (#1892). This also depends on my `@rd` parsing on #1991.

I'm curious as to what people think should go to the `btc-node-store` from these sections. Not actually sure we care about any of these calls for now!

The calls have been manually tested against two local regtest nodes (`getblocktemplate` requires two networked nodes). I'll write similar unit tests to Jose's work next.

### DIFF
--- a/pkg/arvo/lib/btc-node-json.hoon
+++ b/pkg/arvo/lib/btc-node-json.hoon
@@ -1,66 +1,623 @@
-/-  *btc-node-store
+/-  *btc-node-hook
 |%
 ++  rpc
-  =,  ^rpc
-  |%
-  ++  request-to-rpc
-    =,  enjs:format
-    |=  req=request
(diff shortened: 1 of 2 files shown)


pkova:
Made a PR to Jose's branch at yosoyubik/urbit#2 which contains my stuff here but reworked to match with his conventions.

yosoyubik:
thanks @pkova! I hope to get it merged with my stuff tomorrow

pkova:
My stuff is in #1892, closing this.

PR #2018, opened 2019-12-04
soto: tab + android bug fixes

### LINKED ISSUE #2012: soto: tab complete moves cursor strangely
(am on `rc`, but I also saw this on livenet)

If you type `eth` then <TAB>, it correctly completes up to `ethereum` and shows the two options, but it puts the cursor back at the beginning of the line, and you can't move it forward again by clicking, though arrow keys work.

### LINKED ISSUE #2015: soto: "Unidentified" on mobile
Typing "2" in soto on mobile (Android/Chrome) rewrites to "U2nidentified".

matildepark:
- Catches "Unidentified" keys as seen on Android, which should rectify Dojo behaviour on mobile Android browsers. Closes #2015.
- Adds a click handler to set the cursor by clicking.
- Adds a quick catch in sole.js's transpose for insertion events at or above the current cursor. Closes #2012.

I asked @ohAitch about other sole inserts, but it's ~rare outside Clay and the CLI~ (it's actually a false positive in Clay, sorry); he's right that it would probably produce strange effects if a sole application was inserting and deleting behind the cursor. I can't think of a situation that would necessitate that thus far, and it makes tab correction work much smoother, so I think it's acceptable as a tradeoff.


pkg/interface/soto/src/js/components/lib/sole.js
```diff
@@ -57,9 +57,11 @@ export class Share {
           switch (Object.keys(dex)[0]) {
             case 'del': if (at < dex.del) { dex.del++; } break;
             case 'ins': if ((at < dex.ins.at) ||
-                              ((at === dex.ins.at) && !(cha <= dex.ins.cha))) {
+                           ((at === dex.ins.at) && !(cha <= dex.ins.cha))) {
                 dex.ins.at++;
-              } break;
+              } else if (at >= dex.ins.at) dex.ins.at = at;  //NOTE possibly unpredictable behaviour
```
loganallenc: please put this on a newline and within curly braces.

matildepark: Done.

PR #2020, opened 2019-12-04
Ames sponsor scry

belisarius222:
Don't crash on a %public-keys response from Jael that includes a sponsor change but no keys.  Instead, ignore that update.  When we need the sponsor later, scry the up-to-date sponsor from Jael.


belisarius222:
Closing, git snafu.

PR #2021, opened 2019-12-04
ames: scry for sponsor and don't crash on jael response

belisarius222:
Don't crash on a %public-keys response from Jael that includes a sponsor change but no keys. Instead, ignore that update. When we need the sponsor later, scry the up-to-date sponsor from Jael.


philipcmonk:
NB: this has been deployed to ~zod

jtobin:
On Wed, Dec 04, 2019 at 03:54:01PM -0800, Philip Monk wrote:
> NB: this has been deployed to ~zod

Merci.  I'll reconcile the git / ~zod states today.

PR #2022, opened 2019-12-05
eth-watcher: be quieter about parse errors

philipcmonk:
This is a fix for those transient loud errors where the node is giving `~` in the response to the block number request.  This parses without crashing, retrying if the parse fails.

This also silences spider's crash printing.  That's the responsibility of whoever started the thread -- maybe you expected it to crash.  This is why that stack trace was printing twice.

This is fine to OTA

### DIFF
--- a/pkg/arvo/lib/ethio.hoon
+++ b/pkg/arvo/lib/ethio.hoon
@@ -126,28 +126,30 @@
   =/  m  (strand:strandio ,block)
   ^-  form:m
   |^
+  %+  (retry:strandio ,block)  `10
+  =/  m  (strand:strandio ,(unit block))
+  ^-  form:m
   ;<  =json  bind:m
     %+  request-rpc  url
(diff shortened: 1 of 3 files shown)


pkg/arvo/lib/ethio.hoon
```diff
@@ -126,28 +126,30 @@
...
+  (pure:m (parse-block json))
   ::
   ++  parse-block
     |=  =json
-    ^-  block
-    =<  [[&1 &2] |2]
-    ^-  [@ @ @]
+    ^-  (unit block)
+    =<  ?~(. ~ `[[&1 &2] |2]:u)
+    ^-  (unit [@ @ @])
```
Fang-: Probably better to just slap some faces on here? The tree addressing here is nothing complex, but it's still tree addressing. (I'd also use `=;` or `=-` here, but no strong feelings.)

PR #2030, opened 2019-12-05
build: support for releasing cross-compiled debug binaries

brendanhay:
`sh/cross` can now be invoked as follows:

```
sh/cross urbit linux64
sh/cross urbit darwin

sh/cross urbit-debug linux64
sh/cross urbit-debug darwin
```

Note: this seems orthogonal to #2029, as it to builds a _separate_ `urbit-debug` binary.

### DIFF
--- a/nix/release.nix
+++ b/nix/release.nix
@@ -19,9 +19,13 @@ let
   ge-additions = env:
     import ./pkgs/ge-additions/cross.nix env;
 
-  urbit = env:
-    import ./pkgs/urbit/release.nix env
-      { ent = ent env; ge-additions = ge-additions env; debug = false; name = "urbit"; };
+  urbit = { env, debug }:
+    import ./pkgs/urbit/release.nix env {
+      inherit debug;
+      name         = if debug then "urbit-debug" else "urbit";
+      ent          = ent env;
+      ge-additions = ge-additions env;
+    };
 
   builds-for-platform = plat:
     plat.deps // {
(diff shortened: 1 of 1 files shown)


jtobin:
It is morally better to have a specific urbit-debug binary, and IIRC
there used to be one of these.  I gather that the rationale for #2029 is
that in practice we pretty much always want debug symbols at present.

Given we do always want them, perhaps best is to include both
derivations, but to temporarily include debug symbols in the basic urbit
derivation anyway (w/a comment acknowledging this explicitly).

philipcmonk:
Note that urbit-debug is more than just debug symbols, so having both is still useful.

joemfb:
It's specifically worth noting that `urbit-debug` sets the `MEMORY_DEBUG` build option, which changes the structures in the memory image. The `urbit-debug` can only be used with a pre-existing pier if the log is replayed. (We could work around this with the `|pack` implementation ...)

As it stands, `urbit-debug` is the PR-build CI configuration, and not directly useful for much else.

joemfb:
@brendanhay, I should've asked: what's the use-case for this? It might be that it's fine as-is, or you might really need a somewhat-different set of build options.

brendanhay:
I deploy it a sort of smoke test to the hosting platform, admittedly only
using fake ships. Your point about MEMORY_DEBUG means in the longer term
this would cease to be as useful and having a binary named urbit-debug
might possibly even be misleading - I’d rather come up with either a
singular urbit-debug binary that wouldn’t produce incompatible piers or
better nomenclature for naming the binaries produced by various debug
options - maybe urbit-debug-ci etc.

PR #2034, opened 2019-12-05
ames: don't crash on missing queued larval event

philipcmonk:
@jtobin, this is what I pushed to ~zod today.  The diff below is somewhat misleading, the actual changes are what was in philip/p2p, but apparently this commit (cherry-picked from bee0b580) was already on ~zod.  Feel free to close this after the next OTA that contains these changes or do whatever you want with the history.

Incidentally, the git action would have been clearer if these changes hadn't been rebased on top of philip/p2p since it would have been a simple history-preserving merge, no cherry-picks.  I guess that's the cost of a revisionist linear history.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -894,9 +894,15 @@
         [~ larval-gate]
       ::  if crashed, print, dequeue, and set next drainage timer
       ::
-      ::    TODO: cleanup duplicate iteration logic
-      ::
       ?^  error.sign
+        ::  .queued-events should never be ~ here, but if it is, don't crash
+        ::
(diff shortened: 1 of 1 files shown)


jtobin:
This was a very unusual commit -- I think it snuck onto ~zod while @belisarius222 was fixing the breach issues, but it got committed to another branch in git after I had already created a new pill and continued booting the rest of the network.  I cherry-picked it to master yesterday to include it in the hotfix OTA that's sitting (so it's also on `v0.10`) but didn't merge the whole branch it was on, because that branch had a couple of WIP commits in it that I'd previously cleaned up.

So yeah.  Very weird case attributable to post-breach frenzy, should basically never happen.

PR #2035, opened 2019-12-05
drum: print tangs in order

philipcmonk:
Tangs were being printed in reverse order because they weren't updated after I removed the double-reversing action of hood to bring it in conformance with accepted standards.

OTA-able

### DIFF
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -438,7 +438,7 @@
   |=  tac/(list tank)
   ^+  +>
   =/  wol/wall
-    (zing (turn (flop tac) |=(a/tank (~(win re a) [0 edg]))))
+    (zing (turn tac |=(a/tank (~(win re a) [0 edg]))))
   |-  ^+  +>.^$
   ?~  wol  +>.^$
   ?.  ((sane %t) (crip i.wol))  :: XX upstream validation


joemfb:
@philipcmonk, this broke `+cat` and `+tree`, which both produce tangs in the reverse order they expect them to be printed. I thought that was expected (and what this PR was restoring); now I'm now confused about what it means to print tangs "in order".

philipcmonk:
Tangs in order are the same as moves in order: the head is printed before
the tail. Maybe this didn't change when I fixed hood's move order, but it
seems to me we should be consistent.

If [1 2 3] prints 3 2 1 I think that's wrong. There is always a mismatch
between wanting to add tanks/moves by consing on the front and wanting to
print/execute by popping off the front, so a flop will always be necessary
somewhere. The difference is when you don't cons one at a time but rather
make a whole list. That case is much easier to think about in what I call
"natural order".

ohAitch:
That makes sense if the flop of stack traces has been delegated to ++mock,
but is a fundamental API change that warrants review of every reference to
++tang, not a quickfix.

joemfb:
I agree with @ohAitch, and should've checked when I reviewed. This breaks a bunch of things that produce tangs, documentation for %say generators, &c.

philipcmonk:
This is news to me, probably because eg slog prints stuff in order. I
thought this was restoring previous behavior, but I didn't check anything
else that uses it. If we revert this, I can fix the specific thing I was
doing by putting a flop in strand-fail.

Worth making sure at some point that we're consistent with our use of tang.

jtobin:
On Sat, Dec 07, 2019 at 02:38:53PM -0800, Joe Bryan wrote:
> @philipcmonk, this broke `+cat` and `+tree`, which both produce tangs
> in the reverse order they expect them to be printed.

On Sat, Dec 07, 2019 at 03:52:40PM -0800, Joe Bryan wrote:
> This breaks a bunch of things that produce tangs, documentation for
> %say generators, &c.

FWIW: however this gets resolved, this would be an ideal time to add a
couple of basic tests for these things to prevent similar regressions in
the future.

Issue #357, opened 2015-07-08
++  mink jet mismatch: %1 block autocons


ohAitch:
++mink, as written, supports having two blocking requests in a cell, e.g.

```
> (mink 0^[[11 1 20] [11 1 21]] ,~)
::  ought to be
[%1 ~[20 21]]
```

This is not reflected in u3's idea of nock execution semantics(actual result `[%1 ~[20]]`), which causes causing e.g. a `(turn paths file)` to induce O(`(lent paths)`) network round-trips in ford instead of all the blocks being computed in one pass.

ohAitch:
@philipcmonk which, apparently, it doesn't at the C level P:

ohAitch:
Should this just be patched in `++mock` with a comment documenting the desired future behavior?

jtobin:
@joemfb Was this one resolved by #2285 afterwards?

joemfb:
There's a fix on a branch (`na-release/next-vere`), but it'll require a breach to release it.

joemfb:
Fixed in #2285, finally released.

PR #2046, opened 2019-12-06
single-block +mink

frodwith:
Our virtual nock interpreter +mink defines the semantics of the virtual nock operator 12 (the .^ rune in hoon, often called "scry" or "wish") as well as the semantics of exception handling. You give +mink a subject, a formula, and a gate for doing namespace lookups. Instead of returning a product, +mink returns one of:
  [%0 product]
  [%1 (list blocked-paths)]
  [%2 stack-trace]

+mink is supposed never to crash while running the nock formula, but can crash while executing a scry if the scry gate itself crashes.

Why a list of blocked paths? Some nock operators compute one thing from one subject. Some compute two (autocons, equality, nock(2)). +mink takes that opportunity (or should, the current implementation is subtly wrong for 5) to compute both arguments (as long as they do not crash) and, if they both block, concatenate the blocks together. This has the effect that sometimes you would be able to report more of the things that would block in a program - if i had a cons of two scrys, i could report that both of them block, giving the caller of +mink the opportunity to fetch more things before calling me again. So far, so good.

The problem appears with the +mink jet. Naturally we jet +mink by running the normal nock interpreter. There are two ways to support "block welding" in a nock interpreter. One is as +mink does, by having the return type include a case for blocks, and carefully welding blocks - that is by "directly" jetting mink. Now you have two nock interpreters (although nock can simply call mink with a crashing scry gate and unpack the result, much like +mack in hoon.hoon).

Another approach (the one taken by all current implementations) is to use an interpreter that relies more on an external runtime - specifically, stack trace handling is external (not part of the return type) and blocked scrys become exceptions. the +mink jet then becomes the exception handler for these block exceptions. This approach restricts us to a single block, however, since throwing the exception generally means we are done computing and won't try (for example) the other side of an autocons to find more blocks. If we want to correctly jet mink (without what amounts to a jet mismatch), we also need to trap for the block exceptions any time we need to compute two things from the subject - for example, we need to wrap a try around each of the head and tail sides of an autocons and weld them together if they are blocks. In vere this would mean descending into two inner roads every time we cons a cell. In other environments the cost is less extreme, but nonzero. Optimizations are possible (for instance not trapping if the subformulas don't contain 12s), but at complexity cost.

The current situation is a jet mismatch, and therefore unacceptable. Either +mink should be changed to report only a single block, enabling the exception handling regime described above without jet mismatches, or the underlying interpreter (at least the one run during +mink) should be structured more like mink in order to properly weld and propagate block lists without unacceptable performance overhead. The jet calling convention would also likely need to change to accommodate the interpreter structure. I feel it would be less disruptive simply to limit +tone (the +mink result type) to a single block in its block case, even with the downside of being theoretically able to report less blocks to the mink caller in some cases. We are practically limited to a single block at present anyway.

This pull requests implements that strategy. In addition, the current implementation of +mook and +mush use the clam form of calling a mold as a gate, expecting the mold to bunt if the sample does not conform to the mold. Alas, that behavior has changed; these functions are updated accordingly to accomodate the change in mold semantics. They will no longer spuriously crash when not given carefully typed input.

I have also taken the liberty of improving some of the names. Since a primary purpose of +mink is to be read, it should be readable.

A word to the wise for those following discussion on the urbit-dev mailing list: this PR does not address stack trace nondeterminism.


jtobin:
Note: the commit history is wonky because @frodwith branched from a release tag.  I'm gonna help clean it up.

joemfb:
This will close #357..

jtobin:
Superceded by #2285.

PR #2057, opened 2019-12-08
ames: honor -p for non-galaxies

joemfb:
This was never properly wired up in cc-release. Also adds port printfs for all ships.

### DIFF
--- a/pkg/urbit/vere/ames.c
+++ b/pkg/urbit/vere/ames.c
@@ -417,13 +417,17 @@ _ames_io_start(u3_pier* pir_u)
     c3_c* imp_c = u3r_string(imp);
     c3_y  num_y = (c3_y)pir_u->who_d[0];
 
+    if ( 0 != por_s ) {
+      u3l_log("ames: czar: -p %d ignored\n", por_s);
+    }
+
     por_s = _ames_czar_port(num_y);
 
     if ( c3y == u3_Host.ops_u.net ) {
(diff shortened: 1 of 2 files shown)


ohAitch:
Yeah like, unless you know what you're doing you don't want -p anyway, for
galaxies this should print a loud warning about "the port you _probably_
want is nnnnn" but I think still work?

joemfb:
I can believe it would be slightly more convenient to have that flexibility galaxies. Not sure how much it matters, given the external configuration overhead. @brendanhay, what say you?

noops888:
I would like this option on planets, stars, or even moons so that urbit is not using a random port every time it's run - this makes firewall configuration a lot easier.

brendanhay:
Ideally the `-p` option would be honored for any ship class - but I think loud warnings are in order when used with a Galaxy to ensure the operator does in fact, know what they are doing.

For hosting, having the ability to assign ports using `-p` to any ship is mildly desirable as it removes the need to unobsfucate/fynd the `@p` within internal configuration (say, containers using hardcoded `-p 13337`) - leaving the responsibility of ensuring that `@p + 13337` is exposed externally left up to external configuration (load balancers, firewalls, etc.) or the hosting operator.

joemfb:
The aye's have it. @belisarius222, PTAL.

(Closes #2062)

jtobin:
I'll merge this as I'm judging by @belisarius222's original review here, plus the comment in the linked issue, that he'll find the solution here agreeable.

PR #2061, opened 2019-12-09
vere: print arg to c3_malloc() et. al. on failure

belisarius222:
This should help debug the mysterious "memory lost" printfs we've been seeing on our infrastructure nodes.

### DIFF
--- a/pkg/urbit/include/c/defs.h
+++ b/pkg/urbit/include/c/defs.h
@@ -86,21 +86,27 @@
     /* Asserting allocators.
     */
 #     define c3_free(s) free(s)
-#     define c3_malloc(s) ({          \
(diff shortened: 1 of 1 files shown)


belisarius222:
I've tried `%u`, `%z`, `%zu`, `%lu`, and `%i` paired with `c3_w`, and none compile.  I think it's because the inputs to this macro vary between `unsigned int`, `unsigned long`, and I think `unsigned long long`.  I don't think using `c3_d` will break anything, so unless there's something I'm missing here, I think we should just leave this as is.

joemfb:
Those allocation functions should only take up to `SIZE_MAX`, if we're exceeding that, that could be our problem right there. So t'll be useful for debugging as is.

PR #2066, opened 2019-12-09
Jet `repn`

benjamin-tlon:


joemfb:
@benjamin-tlon, I think this should be relatively straightforward with two passes and `u3r_chop()`.

benjamin-tlon:
@joemfb Would that be fast? I don't understand `chop`.

I'm planning to just do it in the same way as `ripn`, by manually constructing an atom.

joemfb:
It's just arbitrary bloq-addressed XOR from an atom into a word-array. I'm sure it could be faster, but we use it for a ton of things and it seems fast enough.

Here's a (completely untested) guess at what I think this would be:

```
c3_w  len_w = u3qb_lent(x);
c3_w  bit_w = bits * len_w;
c3_w* wor_w = u3a_slaq(0, bit_w);
c3_w    i_w = 0;

while ( u3_nul != x ) {
  u3r_chop(0, 0, bits, i * bits, wor_w, u3h(x));
  x = u3t(x);
  i_w++;
}

return u3a_malt(wor_w);
```

benjamin-tlon:
That seems pretty reasonable, actually @joemfb.

benjamin-tlon:
@joemfb: This is done, though. It should be a little bit faster.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+  ((w) & ((1 << (n)) - 1))
+
+/*
+  Divide, rounding up.
+*/
+#define DIVCEIL(x,y) \
+  (x==0) ? 0 :       \
+  1 + ((x - 1) / y);
+
+u3_noun u3qc_repn(u3_atom bits, u3_noun blox) {
```
joemfb: House style:

```
(qualifier) type
fun(args)
{
...
}
```

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+    if ( !_(u3a_is_cat(bits) || bits==0 || bits>31) ) {
+        return u3m_bail(c3__fail);
+    }
+
+    //
+    //  Calculate input and output size.
+    //
+    c3_w num_blox = u3qb_lent(blox);
+    c3_w bit_widt = num_blox * bits;
+    c3_w wor_widt = DIVCEIL(bit_widt, 32);
```
joemfb: I won't hold you to CVC, but I must insist on the "hungarian" suffixes for all non `u3_noun`/`u3_atom` variables.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+    //
+    //  Bits are pushed into the `acc` register and flushed to the buffer
+    //  once full.
+    //
+    //  acc  register
+    //  use  number of register bits filled (used)
+    //  cur  next buffer word to flush into.
+    //
+
+    c3_w acc=0, use=0, *cur=buf;
```
joemfb: Seems like this would be cleaner in its own block

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+
+#   define FLUSH() *cur++=acc; acc=use=0
+#   define SLICE(sz,off,val) TAKEBITS(sz, val) << off
+
+    for (c3_w i=0; i<num_blox; i++) {
+        u3_noun blok_n = u3h(blox);
+        blox = u3t(blox);
+
+        if (!_(u3a_is_cat(blok_n)))
+            return u3m_bail(c3__fail);
```
joemfb: I must request braces, and -- my own personal preference -- `if ( c3n == u3a_is_cat(blok_n) )`. The "loobean" conversions drive me nuts.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+
+        c3_w blok = blok_n;
+
+        for (c3_w rem_in_blok=bits; rem_in_blok;) {
```
joemfb: It seems like this really wants to be `while( rem_in_block )`.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+            c3_w rem_in_acc = 32 - use;
+            if (rem_in_blok == rem_in_acc) {            //  EQ
+                acc |= SLICE(rem_in_blok, use, blok);
+                FLUSH();
+                rem_in_blok = 0;
+            } else if (rem_in_blok < rem_in_acc) {      //  LT
```
joemfb: brace/else on separate lines.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+                FLUSH();
+            }
+        }
+    }
+
+    //
+    //  If the last word isn't fully used, it will still need to be
+    //  flushed.
+    //
+    if (use) FLUSH();
```
joemfb: I'd be more comfortable with braces here too, given that macro expansion.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+
+    return u3a_malt(buf);
+}
+
+u3_noun u3wc_repn(u3_noun cor) {
+  u3_noun bits, blox=0;
```
joemfb: no reason to initialize either of these.

pkg/urbit/jets/c/repn.c
```diff
@@ -0,0 +1,112 @@
...
+  if ( (c3n == u3r_mean(cor, u3x_sam_2, &bits, u3x_sam_3, &blox, 0)) ||
+       (c3n == u3ud(bits)) )
+  {
+    return u3m_bail(c3__exit);
+  }
+
+  return u3qc_repn(bits, blox);
+}
+
+u3_noun u3kc_repn(u3_atom bits, u3_atom blox) {
```
joemfb: Function signature style here and above.

joemfb:
A note for the future, *not* work to be done now:

At some point, we need to do a pass over the jets and start returning `u3_none` if we can't handle the parameters, instead of `bail:fail`. The jet dashboard then falls back to the nock.

Similarly, implementing this with `u3r_chop` would handle more inputs. So, for example, this jet would ideally have its current fast path, a more general slower path, and then maybe even fall back to nock for some inputs.

/cc @frodwith

pkg/urbit/jets/c/repn.c
benjamin-tlon: Thanks. Looks like this one was actually a bug.

pkg/urbit/jets/c/repn.c
benjamin-tlon: The for loop is cleaner. Variable doesn't escape scope. This is a pretty standard use of `for`.

benjamin-tlon:
@joemfb, could you resolve any remaining nits and merge this?

joemfb:
Sure, I can make a pass over it this evening.

benjamin-tlon:
Thanks a lot! @joemfb

PR #2068, opened 2019-12-10
gall: remove message type from wire

philipcmonk:
%leave over the network didn't work because we included the message type
in the wire from gall, so the duct for the initial %watch and the %leave
were different.  We need to know the message type so we can route the
acknowledgment as %poke-ack, %watch-ack, or no-op.

This moves this piece of information to a piece of state, where we queue
up the message types per [duct wire].  Ames guarantees that
acknowledgments will come in order.

This also includes an easy state adapter.  The more interesting part of
the upgrade is that we likely have outstanding subscriptions with the
old wire format.  The disadvantage of storing information in wires is
that it can't be upgraded in +load.  So, here we listen for updates on
the old wire format, and when we get them we kill the old subscription,
so that it will be recreated with the new wire format.

As an aside, this is a good example of what we mean when we say
subscriptions may be killed at any time, so apps must handle this case.

Finally, this fixes the "attributing" ship to ~zod for agent requests.
This information was ignored for agent requests, but including it causes
spurious duct mismatches.

This is OTA-able.  cc @Fang-, who reported the bug.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -54,7 +54,7 @@
 ++  state
   $:  :: state version
       ::
-      %1
+      %2
       :: agents by ship
       ::
       =agents
@@ -75,6 +75,9 @@
   $:  ::  system duct
       ::
       system-duct=duct
+      ::  outstanding request queue
+      ::
+      outstanding=(map [wire duct] (qeu remote-request))
       ::  foreign contacts
       ::
       contacts=(set ship)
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -596,16 +601,27 @@
...
+          =/  stand
+            %+  ~(gut by outstanding.agents.state)  [sys+wire hen]
+            *(qeu remote-request)
+          ~|  [sys+wire=wire hen=hen stand=stand outs=outstanding.agents.state]
+          =^  rr  stand  ~(get to stand)
+          [rr (~(put by outstanding.agents.state) [wire hen] stand)]
+        ::  non-null case of wire is old, remove on next breach after
+        ::  2019/12
+        ::
+        [(remote-request i.t.t.t.wire) outstanding.agents.state]
```
belisarius222: I think this should have a ;; on the call to +remote-request

philipcmonk: Done

PR #2069, opened 2019-12-10
vere: adds tests for newt ipc framing, fixes infinite loop

joemfb:
This PR builds on #2042 (and really should've been included there). It includes further refactoring of `newt.c` to enable testing of the serialization and parsing routines, said tests, and a fix for an infinite loop in the parser.

### DIFF
--- a/pkg/urbit/vere/newt.c
+++ b/pkg/urbit/vere/newt.c
@@ -31,8 +31,6 @@
 #include "all.h"
 #include "vere/vere.h"
 
-#undef NEWT_VERBOSE
-
 /* _newt_gain_meat(): add a block to an existing message
 */
 static void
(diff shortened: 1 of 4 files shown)


pkg/urbit/vere/newt.c
```diff
@@ -219,6 +218,9 @@ _newt_consume(u3_moat* mot_u)
       if ( 8ULL <= mot_u->len_d ) {
         _newt_gain_mess(mot_u);
       }
+      else {
+        break;
+      }
```
joemfb: This fixes the bug introduced in #2042. If we have "stray bytes", but not enough to extract a message length, we must break out of the loop.

pkg/urbit/vere/newt.c
```diff
@@ -256,6 +258,7 @@ _newt_read_cb(uv_stream_t*    str_u,
   u3_moat* mot_u = (void *)str_u;
 
   if ( 0 > len_i ) {
+    c3_free(buf_u->base);
```
joemfb: This is really just a question of code hygiene, as any errors here are fatal.

PR #2071, opened 2019-12-10
ames: recover from mismatched message nums

philipcmonk:
We've seen issues where the message-num of the head of live.state is
less than current.state.  When this happens, we continually try to
resend message n-1, but we throw away any acknowledgment for n-1 because
current.state is already n.  This halts progress on that flow.

We don't know what causes us to get in this bad state, so this adds an
assert to the packet pump that we're in a good state, run every time
the packet pump is run.  When this crashes, we can turn on |ames-verb
and hopefully identify the cause.

This also adds logic to +on-wake in the packet pump to not try to resend
any messages that have already been acknowledged.  This is just to
rescue ships that currently have these stuck flows.

(Incidentally, I'd love to have a rr-style debugger for stuff like this.
Just run a command that says "replay my event log watching for this
specific condition and then stop and let me poke around".)

This is OTA-able.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -768,7 +768,7 @@
   $%  [%hear =message-num =fragment-num]
       [%done =message-num lag=@dr]
       [%halt ~]
-      [%wake ~]
+      [%wake current=message-num]
   ==
 ::  $packet-pump-gift: effect from |packet-pump
 ::
@@ -2222,6 +2222,7 @@
     =~  (dispatch-task task)
         feed-packets
         (run-packet-pump %halt ~)
+        assert
         [(flop gifts) state]
     ==
   ::  +dispatch-task: perform task-specific processing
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2391,6 +2392,16 @@
     =.  message-pump  (give i.packet-pump-gifts)
     ::
     $(packet-pump-gifts t.packet-pump-gifts)
+  ::  +assert: sanity checks to isolate error cases
```
belisarius222: Actually, I wonder if we'd be better off just printing here to make sure ships don't get stuck from crashing on this -- especially since you added the recovery logic.

philipcmonk:
But if we print, we probably won't get enough information. This should only
get that one flow stuck, and that's been the case on several ships already
with no adverse affects.

I tested this on ~wicdev, and it correctly fixed the issue that already
existed.

PR #2072, opened 2019-12-10
Check Arvo compatibility on boot

jtobin:
Adds a basic check for Arvo compatibility in _worker_work_boot & u3_worker_boot, via a _worker_check_zuse helper.

We hardcode the version of Arvo/%zuse that this version of Vere was developed against in _CVX_ZUSE and do a simple comparison to +zuse (via +wish) to check if we know how to run the supplied Arvo core.  At present +zuse tracks the kernel API, so any change to it represents a backwards-incompatible change from Vere's perspective.

The present check simply verifies that the versions are compatible and refuses to run the core otherwise.

Of note: I've used u3m_reclaim() to clear the wish cache before retrieving +zuse on restart -- otherwise we can wind up with a stale version of %zuse floating around.  This feels like overkill; @joemfb advised that there was likely a better way to handle this via u3m_soft(), but I wasn't able to figure out a satisfying way to do that.  Any pointers?

### DIFF
--- a/pkg/urbit/worker/main.c
+++ b/pkg/urbit/worker/main.c
@@ -24,6 +24,8 @@
 #include "all.h"
 #include <vere/vere.h>
 
+#define ZUSE 309
+
     typedef struct _u3_worker {
       c3_w    len_w;                        //  boot sequence length
       u3_noun roe;                          //  lifecycle formulas
@@ -120,6 +122,41 @@ void _worker_space(FILE* fil_u,  c3_w n)
 {
   for (; n > 0; n--)
     (fprintf(fil_u," "));
+}
+
+/* _worker_read_zuse(): read +zuse via +wish.
+*/
+static u3_noun
+_worker_read_zuse(u3_noun non)
+{
+  return u3v_wish("zuse");
+}
+
+/* _worker_check_zuse(): check arvo compatibility.
+*/
+void
+_worker_check_zuse(void)
+{
+  u3_noun pro = u3m_soft(0, _worker_read_zuse, c3y);
+
+  if ( u3_blip != u3h(pro) ) {
+    fprintf(stderr, "fatal error: +zuse not present\n");
+    exit(1);
+  }
+  else {
+    u3_noun ver = u3k(u3t(pro));
+
+    if ( ver != ZUSE ) {
+      fprintf(stderr,
+              "work: zuse incompatible (have %u, need %u)\r\n",
+              ver,
+              ZUSE
+             );
+      exit(1);
+    }
+  }
+
+  u3z(pro);
 }
 
 /* _worker_print_memory(): print memory amount.
@@ -735,6 +772,8 @@ _worker_work_boot(c3_d evt_d, u3_noun job)
     u3V.mug_l  = u3r_mug(u3A->roc);
     u3A->ent_d = u3V.dun_d;
 
+    _worker_check_zuse();
+
     u3l_log("work: (%" PRIu64 ")| core: %x\r\n", evt_d, u3V.mug_l);
   }
   else {
@@ -972,6 +1011,8 @@ u3_worker_boot(void)
   if ( 0 != u3V.dun_d ) {
     u3V.mug_l = u3r_mug(u3A->roc);
     nex_d    += u3V.dun_d;
+
+    _worker_check_zuse();
   }
   else {
     u3V.mug_l = 0;


pkg/urbit/worker/main.c
```diff
@@ -24,6 +24,8 @@
 #include "all.h"
 #include <vere/vere.h>
 
+#define _CVX_ZUSE 309
```
joemfb: This particular naming convention is local to `vortex.c` -- caps 'cause it's a constant, _cv because it's static/local to `vortex.c`. Feel free to name this however you see fit!

pkg/urbit/worker/main.c
```diff
@@ -735,6 +753,15 @@ _worker_work_boot(c3_d evt_d, u3_noun job)
     u3V.mug_l  = u3r_mug(u3A->roc);
     u3A->ent_d = u3V.dun_d;
 
+    {
+      u3_noun ver = u3v_wish("zuse");
+      u3m_p("zuse", ver);
```
joemfb: I don't think we need to print these. But if you want to, they should go back over the pipe with `u3l_log()`

pkg/urbit/worker/main.c
```diff
@@ -122,6 +124,22 @@ void _worker_space(FILE* fil_u,  c3_w n)
     (fprintf(fil_u," "));
 }
 
+/* _worker_check_zuse(): check arvo compatibility.
+*/
+void
+_worker_check_zuse(u3_noun ver)
+{
+  if ( ver > _CVX_ZUSE ) {
+    u3l_log("work: zuse incompatible (need vere < v" URBIT_VERSION ")\r\n");
```
joemfb: I think we *should* print the version numbers here. Have/need or something like that.

joemfb: I should add, these printfs should use `stderr` over `u3l_log()`; the latter goes over the pipe, and we'll exit immediately, so they might not get there.

jtobin:
> It's a little awkward; something like this:
> 
> [..]
>
> `u3_noun pro = u3m_soft(0, _worker_read_zuse, c3y);`

Ah, the use of the dummy argument here helps.  I was trying to figure out how to use u3m_soft() to call u3v_wish, which expects a C string as its argument rather than a noun (contra u3m_soft).

jtobin:
I may have to ignore the check on the %vega event for now as getting a new binary with some kind of decent Arvo compatibility check onto the network needs to be done pronto.

What's best is perhaps to merge this as-is (since what has been added is a strict improvement, and at least the existing additions have been approved in principle), tag a Vere release candidate, and deploy it to Tlon's ships so that we're genuinely running something that's performing a reasonable compatibility check.  The %vega piece can be added in short order to a non-candidate release, and deployed in turn.

jtobin:
Or rather -- what I can do is refrain from merging this to master now, but cherry-pick the relevant existing changes over to the release branch to construct the candidate.  That should be a little more palatable.

belisarius222:
So ... does this prevent OTA updates that decrement a Kelvin version?  That seems pretty restrictive, since not all kernel updates affect Vere.  I worry that this will cause us to either a) require everybody to reinstall their binaries when we OTA, or b) dodge decrementing Kelvins properly, to avoid that.

What was the rationale for not just requiring a Kelvin at least as modern as the one Vere is expecting?  Maybe I've got that backward somehow, but I don't get why we should require the versions to match exactly.

—
~rovnys-ricfer
https://urbit.org

joemfb:
It just means either a) we should not make kelvin-changing upgrades to zuse/arvo/hoon unless/until we are also changing the arvo/vere interface (broadly considered), or b) if we need to OTA such changes, we'll need to intercept reads to %zuse in Arvo's `+wish` arm. As for supporting a range of kelvins in vere, we should totally do that. But we just don't right now. A good opportunity will be the next time we change the arvo/vere interface.

jtobin:
@joemfb I just realised that the check here actually fails when booting from a brass pill:

```
fatal error: +zuse not present
pier: work error: end of file
pier: work exit: status 1, signal 0
```

Presumably the +wish is occurring before %zuse has actually been installed?

jtobin:
(This wasn't an issue when deploying to Tlon's ships, since they'd already been booted.)

joemfb:
That makes sense, it's only with a solid pills that %zuse is present at the end of the lifecycle sequence.

I intend to change this in the near future, and make the lifecycle sequence always take only one event. Another nice approach would be to make arvo's metamorphosis out of the larval stage observable with an effect. And, finally, `+peek` should be exposed over the IPC barrier. Once all these things are done, checking the appropriate conditions can trivially move to the daemon.

For now, since the %zuse installation is not directly observable, we'll have to infer it indirectly to support both pills. In a solid boot, `u3V.len_w` will be `1`, with brass, it'll be `5`. That lifecycle sequence is immediately followed by identity and entropy events, which is followed by %zuse (at this point, the larval stage is discarded). So the right place to check during boot is on the effects of event `u3V.len_w + 3`, or on event `u3V.len_w + 4`.

ohAitch:
I would like to second the expectation that having the check be an =
instead of a < will cause underuse of kelvin decrementing; especially in
combination with the full continuity commitment. In theory, zuse and hoon
are supposed to be always backward-compatible, with any colder kelvin
supporting all functionality exposed by its predecessors - "we do not break
userspace", in linux parlance. The difference between theory and practice
of course being…

(this was btw also broadly my objection to calling the confusingly-named
"thread" system %gall, without afaik providing any direct adapter that
would work on urbit/examples etc)

belisarius222:
I'm having trouble following your line of reasoning, @ohAitch.  In theory, we're not supposed to break userspace, ok.  And I think your ellipses imply that in practice, we might still break userspace.  But how does that relate to this check in Vere?

—
~rovnys-ricfer
https://urbit.org

ohAitch:
The purpose of the check in vere doing exact pinning is because zuse will
break compatibility. If zuse were following proper kernel kelvin
append-only semantics, and keeping all old interfaces around, ++zuse ≤
MAX_ZUSE would be sufficient to check that the old interfaces are present.
The cost would be maintaining adapters - the transition stage code between
multi homing and single homing was not exactly "clean"

(I'm also not super following why we're calling the arvo.hoon version
number "zuse", given zuse is the set of standard library arms exposed to
hoon programs - this may be part of my pushback here. If we were pinning
specifically arvo.hoon, it would be feasible to add arms to zuse and
decrement the zuse version number such that userspace could signal
requiring these arms, without also requiring everyone to recompile vere)

joemfb:
@ohAitch, I generally agree. Vere only cares about a) nock, b) Arvo's structural interface (+peek, +poke, &c), and c) the arvo-level events and effects (or maybe the necessary subset thereof).  %zuse and its version includes much more than that. It would be better to check against more granular versions, and factoring the kernel into more suitable levels remains a goal.

However, we would still run into similar situations when it comes to determining validity. Any straightforward versioning scheme for events and effects will depend on hoon at least, but hoon could change dramatically without invalidating those structures. In such a case, you'd want to pretend that you're using the warmer/higher version, when in fact you're using the colder/lower version. Which is exactly what I've proposed we do with this scheme.

ohAitch:
I mean I think in this thread I've seen proposals to outright intercept u3v
calls to ++zuse

which seems like a rather hacky way around having a separate ++arvo,
defined initially as "the last ++zuse bump relevant to vere"

jtobin:
On Sun, Dec 15, 2019 at 11:19:37PM -0800, Anton Dyudin wrote:
> I would like to second the expectation that having the check be an =
> instead of a < will cause underuse of kelvin decrementing; especially
> in combination with the full continuity commitment.

Of note, I think you're absolutely correct here re: underuse.  As I see
it, we've picked up the most conservative possible mechanical
implementation of versioning -- with kelvin decrements occurring on
backwards-incompatible kernel interface changes -- just to get the ball
(re)rolling.

This is acceptable to me for the time being, but IMO it should not be a
permanent (or even prolonged) state of affairs.  I expect the mechanical
implementation to become more refined early in the new year.

jtobin:
I'll close this one for the time being.  I plan to get back to this general line of work in a week or two, but I suspect I'll want to look into a more involved compatibility check (and possible upgrade scheme).  I've since reverted these changes on `v0.10`.

Issue #1945, opened 2019-11-13
memory lost error in newt.c


belisarius222:
```
; ~zod not responding still trying
; ~zod is ok
[ ! /~rovnys-ricfer/home/0/app/weather/hoon:<[106 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[108 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[110 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[111 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[114 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[115 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[117 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[118 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[122 3].[130 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[122 17].[125 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[122 40].[125 5]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[123 5].[123 47]>
  ! /~rovnys-ricfer/home/0/app/weather/hoon:<[123 15].[123 47]>
  [%ap-lame %weather %http-response]
]
; ~zod not responding still trying
; ~zod is ok
Assertion '!"memory lost"' failed in vere/newt.c:120

bail: oops
bailing out
Assertion failed: 0 (noun/manage.c: u3m_bail: 667)
pier: work error: stream closed                   \«http»
pier: work exit: status 0, signal 6
```

On restart, I get a ton of %canceling-nonexistant-channel prints and then a %behn-goad.  It's now been spinning on <<behn>> for a few minutes.

belisarius222:
Huh, looks like that's actually a c3_malloc() call in newt.c that failed:
https://github.com/urbit/urbit/blob/master/pkg/urbit/vere/newt.c#L120

belisarius222:
It got past the <<behn>> spinning and errored out in weather.hoon again in the same way; now it's doing %behn-goad-again and printed a bunch of stuff including a %reset-restart, then dumped me back into the dojo, and the web UI is back up.

[...]

PR #2076, opened 2019-12-10
vere: plug a memory leak in IPC message parsing

joemfb:
This builds further on #2069 et all. As far as I can tell, this leak has always been present on `cc-release` and related branches. Pretty obvious in retrospect ...

### DIFF
--- a/pkg/urbit/vere/newt.c
+++ b/pkg/urbit/vere/newt.c
@@ -181,6 +181,8 @@ _newt_poke_mess(u3_moat* mot_u)
     u3_noun mat = u3i_bytes((c3_w)len_d, buf_y);
     mot_u->pok_f(mot_u->vod_p, mat);
   }
+
+  c3_free(buf_y);
 }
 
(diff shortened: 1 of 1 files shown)


joemfb:
This should *actually* fix #1945.

vvisigoth:
👑
—
~poldec-tonteg
http://urbit.org

philipcmonk:
So you think those piers have been running out of system memory?

joemfb:
We were leaking the buffer for IPC message we received, and we push a lot of data through that pipe. But I can't guarantee that's the only problem... I'm looking for a leak checker we can run in CI.

ohAitch:
I hear it ran fine under valgrind once upon a time

joemfb:
@ohAitch, I was able to get it to run (for the daemon and test suite) by moving the loombase on linux. It doesn't like tracing both processes, so I'm still trying to get it to run on the worker process.

PR #2078, opened 2019-12-11
ames: don't forward up the sponsorship chain

philipcmonk:
This is *actually* why the galaxies are under so much load.  They're in
a forwarding loop with their stars, and this breaks the loop.


vvisigoth:
👑
—
~poldec-tonteg
http://urbit.org

PR #2081, opened 2019-12-11
Ames trace by ship

belisarius222:
Adds |ames-sift generator that passes a %sift $task to ames with a list of ships to track.  If null, Ames prints for all ships (still subject to the submodule filtering of |ames-verb); otherwise, Ames prints just for those ships selected in the arguments to the generator.

Includes a state adapter, which I've run locally without issue.  Should be OTA'able.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1771,7 +1771,8 @@
   ++  send-blob
     |=  [for=? =ship =blob]
     ::
-    %-  (trace rot.veb ship |.("send-blob: to {<ship>}"))
+    =/  final-ship  ship
+    %-  (trace rot.veb final-ship |.("send-blob: to {<ship>}"))
     |-
     |^  ^+  event-core
         ::
@@ -1792,10 +1793,10 @@
           (try-next-sponsor sponsor.peer-state)
         ::
         ?~  route=route.peer-state
-          %-  (trace rot.veb ship |.("no route to:  {<ship>}"))
+          %-  (trace rot.veb final-ship |.("no route to:  {<ship>}"))
           (try-next-sponsor sponsor.peer-state)
         ::
-        %-  (trace rot.veb ship |.("trying route: {<ship>}"))
+        %-  (trace rot.veb final-ship |.("trying route: {<ship>}"))
         =.  event-core
           (emit unix-duct.ames-state %give %send lane.u.route blob)
         ::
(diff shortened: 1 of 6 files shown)


philipcmonk:
Will review this soon, but in the meantime, can you get this to pass CI and clean up the commit messages?  (At least appropriate prefixes)

belisarius222:
Sure

belisarius222:
Cleaned up the commits.  Not sure what's up with CI, since this is working just fine on my machine.  Making a pill to see if that helps.

belisarius222:
@philipcmonk CI passed and commits are cleaned up; please review at your leisure

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1733,10 +1792,10 @@
           (try-next-sponsor sponsor.peer-state)
         ::
         ?~  route=route.peer-state
-          %-  (trace rot.veb |.("no route to:  {<ship>}"))
+          %-  (trace rot.veb ship |.("no route to:  {<ship>}"))
```
philipcmonk: I think this should be gated on the original `ship`.  If I want to trace the routes tried to send to `~wicdev-wisryt`, I don't want that not printing that it tried `~marzod` just because I didn't list it.  You'll need to store the original ship at the top of +send-blob for this purpose.

Same below

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1032,26 +1069,37 @@
...
-          [%2 ^ames-state]
+          [%2 ames-state-2]
+          [%3 ^ames-state]
       ==
   ^+  ames-gate
   ?-    -.old-state
-      %2
+      %3
     ames-gate(ames-state +.old-state)
+  ::
```
philipcmonk: This looks trivial enough to be fine, but I generally upgrade one at a time with a sequence of `=?`s, else it can't be tested without trying upgrades from each of the old versions -- a quadratic problem.  Example:

https://github.com/urbit/urbit/blob/0c95c2d4dd3a7d1caccccd1412d0b2fdec104d49/pkg/arvo/sys/vane/gall.hoon#L1554-L1566

belisarius222:
@philipcmonk PTAL

jtobin:
One other commit-structure note: avoid using standalone commits for pill changes.  I previously advocated this, but have since changed my tune.  I now think they're in the same category as WIP commits, and should just be squashed into the commit that contains the corresponding source change.

(No need to fix this one -- I'll squash it for you.)

PR #2089, opened 2019-12-12
Fix typo in +put:in.

eglaysher:
This fixes +put:in so that it works without the correct jet.
There's a mismatch where the hoon code is wrong and the jet is
correct, so that when we try to run this on alternate interpreters
which may not have the +in jets, things won't work.


joemfb:
@eglaysher, my apologies for missing this in the original review! I spent last night pouring over `+put:by`, so I'm confident this fix is correct ...

yosoyubik:
Whoops, apologies for this :( Thanks for the fix and for looking into it!

jtobin:
@eglaysher I've patched up the commit here by adding a `hoon: ` prefix and also by adding an updated solid pill.

PR #2090, opened 2019-12-13
vere: disables accelerated http-server graceful shutdown

### LINKED ISSUE #2032: seg faults when using https
I've gotten multiple segmentation faults, on two different planets on two different servers, that seem to only happen when I'm using landscape over https

address 0x159 out of loom!
loom: [0x36000000 : 0xb6000000)
Assertion '0' failed in noun/events.c:129

bail: oops
bailing out
Assertion failed: 0 (noun/manage.c: u3m_bail: 667)
                                                  Segmentation fault

joemfb:
This fixes #2032. Rather, it fixes our only consistent reproduction thereof. I'll keep looking for there underlying cause, but a) the http-server i/o driver needs to be rewritten, and b) that's a pretty low-priority task, given the approach of king haskell.

### DIFF
--- a/pkg/urbit/vere/http.c
+++ b/pkg/urbit/vere/http.c
@@ -437,11 +437,14 @@ _http_hgen_send(u3_hgen* gen_u)
   else {
     h2o_send(rec_u, vec_u, len_w, H2O_SEND_STATE_FINAL);
 
-    u3_h2o_serv* h2o_u = req_u->hon_u->htp_u->h2o_u;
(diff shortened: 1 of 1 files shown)


jtobin:
On Thu, Dec 12, 2019 at 05:30:49PM -0800, Joe Bryan wrote:
> /cc @jtobin, this should go into the next scheduled hotfix.

ack'd

joemfb:
Closing, as #2101 has the correct fix.

PR #2094, opened 2019-12-13
Eth tooling: eth-sender

Fang-:
Formerly send-txs. Can still sign eth-txs at multiple gas prices, fan transactions to multiple nodes, wait for confirmation between transaction batches (now of user-specified size).

The previous nonce reading implementation was broken beyond belief and has been taken out. ~Can be reimplemented once RLP decoding is in the stdlib. I have a half-written version of that, will try to bring that up to spec and chuck it in there soon.~
I replaced it with a proper decode-the-transaction approach.

The version of `/ted/eth/send-txs` that just did `!!` for failure cases instead of properly `strand-fail`ing was a bit more straightforward, but its traces were huge... ):


pkg/arvo/app/eth-sender.hoon
```diff
@@ -0,0 +1,297 @@
+::  eth-sender: utility for signing & sending eth-txs
+::
+::    usage:
```
jtobin: Usage documentation.  Damn fine.  *Damn* fine. 👏

PR #2095, opened 2019-12-14
goad: recompile apps after changes to /sys

philipcmonk:
OTAs commonly end up in an inconsistent state if apps depend on changes
to /sys.  For example, the %sift changes break on OTA because %spider
needs to be reloaded so that it's aware of the new thread type.  This
adds a %goad app, which reloads all apps after every change to /sys.

Getting this to start OTA is nontrivial, but this pattern should work
for apps in the future.  The changes to clock shouldn't generally be
necessary; they are only necessary here because we can't rely on hood to
start goad, since hood fails to compile if it's run before zuse is
reloaded.  Once goad is active, this will cease to be a problem.

### DIFF
--- a/pkg/arvo/app/clock.hoon
+++ b/pkg/arvo/app/clock.hoon
@@ -22,8 +22,20 @@
   :~  [%pass / %arvo %e %connect [~ /'~clock'] %clock]
       [%pass /clock %agent [our.bowl %launch] %poke launcha]
   ==
-++  on-save   on-save:def
-++  on-load   on-load:def
+::  bootstrapping to get %goad started OTA
+::
+++  on-save   !>(%1)
+++  on-load
+  |=  old-state=vase
+  =/  old  !<(?(~ %1) old-state)
+  =^  cards  this
+    ?.  ?=(~ old)
+      `this
+    :_  this  :_  ~
+    [%pass /behn %arvo %b %wait +(now.bowl)]
+  ::
+  [cards this]
+::
 ++  on-poke
   |=  [=mark =vase]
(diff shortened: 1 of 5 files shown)


pkg/arvo/lib/hood/drum.hoon
```diff
@@ -212,6 +213,11 @@
     %drum-set-boot-apps  =;(f (f !<(_+<.f vase)) poke-set-boot-apps)
   ==
 ::
+++  on-load
+  |=  %1
+  =<  se-abet  =<  se-view
+  (se-born %home %goad)
+::
```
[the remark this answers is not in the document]
philipcmonk: Then we don't know if `lit` was set, so in development ships we'll start a bunch of apps we were trying to avoid.  We maybe should keep that in the state.  I also feel like you should be able to "suspend" an app, and it shouldn't get automatically restarted, though we don't support that yet.

I'm going to leave this for now so we can get the OTA out today, but next time we should consider a more general solution.

pkg/arvo/app/clock.hoon
```diff
@@ -63,6 +75,13 @@
 ++  on-arvo
   |=  [=wire =sign-arvo]
   ^-  (quip card:agent:gall _this)
+  ?:  ?=(%wake +<.sign-arvo)
+    :_  this  :_  ~
+    ?^  error.sign-arvo
+      :_  this  :_  ~
```
joemfb: It looks like this duplicates the state and null-terminator in the product.

philipcmonk: indeed, mis-paste

PR #2101, opened 2019-12-15
vere: fix use-after-free in http responses

joemfb:
As discovered with clang's ASan. This supersedes #2090.

The thing I had missed, causing #2032, is that `h2o_send()` with `H2O_SEND_STATE_FINAL` *synchronously* finalizes/disposes the request and all associated resources.

There's a second use-after-free fixed in this PR. The response generator is disposed before the request, but I trying to inspect it on request-dispose in order to cancel the request in %eyre. That is now handled through the response generator as it should be.


belisarius222:
LGTM

PR #2107, opened 2019-12-16
King Haskell

benjamin-tlon:


benjamin-tlon:
There is a problem with replays, possibly introduced by Phill's PR (which I reviewed). Ran into this when helping Isaac with his PR.

This needs to be fixed pre-merge.

PR #2113, opened 2019-12-18
Moons, routing, and key propagation

philipcmonk:
A few changes related to making moons work, reducing load on ethereum nodes, and improving jael's ship-to-ship capabilities.  Particularly:

- Adds "eager" option to eth-watcher.  If true, behavior is as now.  If false, we *only* progress by "zooming".  This means we'll generally be about 130 blocks behind, because zooming can't detect reorgs.  I believe this delay (about 30 minutes) is acceptable.  Currently, it's set so that galaxies are "eager" and everything else is not.  Longer term, we should use ship-to-ship Jael so that only info about your sponsor has the 30 minute delay because the rest comes from your galaxy (via your star), which is eager.
- Fixes some accidentally dropped cards in eth-watcher
- Increase ping frequency from once per ~m5 to ~s30.  Many sites recommending setting your UDP NAT pinhole timeout to between 30 and 60 seconds, so this ensures a constant connection under most NAT conditions.  This was required for moons to work reliably when I was running a planet and moon behind a NAT.  Notably, ping time on old ames was ~s32.
- Routing between moons and planets doesn't work the way we were trying to do it.  In reviewing old Ames to determine how it did this, it turns out that old Ames only ever routed via galaxies or direct -- never through stars or planets.  This looks like a bug, since it creates a list of intermediate sponsors only to apparently incidentally throw it away, but if you "fix" the bug you get the forwarding loops we had earlier this era.  So, this reintroduces that behavior, since it's "safe", though inefficient.  This should be revisited as an optimization problem once a proper test suite exists.
- Because of these changes, moons now seem to work fine.
- Includes a partial attempt to make ship-to-ship Jael work correctly.  It seems to work as well as it did last era, but it still doesn't handle the case where your sponsor (or other source of Jael data) breaches.


pkg/arvo/sys/zuse.hoon
```diff
@@ -7626,7 +7630,7 @@
       ::  #  constants
       ::
       ::  contract addresses
-      ++  contracts  mainnet-contracts
+      ++  contracts  ropsten-contracts
```
belisarius222: gotta switch this back to mainnet

pkg/arvo/gen/azimuth-sources.hoon
```diff
@@ -0,0 +1,4 @@
+:-  %say
+|=  [[now=@da *] *]
+:-  %noun
+.^(state-eth-node:jael j//=sources/(scot %da now))
```
Fang-: wh- why is this valid path syntax?

And why not just the more common `%j /=sources` (or even `/j/=sources`) here?

pkg/arvo/sys/vane/jael.hoon
```diff
@@ -308,11 +308,15 @@
...
       ::
-      ::  get everything from azimuth-tracker because jael subscriptions
-      ::  seem to be flaky for now
-      ::
       =.  +>.$  (poke-watch hen %azimuth-tracker nod.own.pki)
       =.  +>.$
+        ::  get everything from azimuth-tracker because jael subscriptions
+        ::  seem to be flaky for now
+        ::
+        ?:  &
```
Fang-: I feel like such things should be marked with an `XX` also, for that "temporary and should be fixed" feel.

pkg/arvo/sys/vane/jael.hoon
```diff
@@ -307,8 +307,20 @@
       ::
```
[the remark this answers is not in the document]
Fang-: This comment of mine was outdated on-post, lol.

pkg/arvo/sys/zuse.hoon
jtobin: On this point: I think we should avoid leaving testnet config lying around in the history at all.  I don't think there's ever a need to do this -- for testnet pills there's `make ropsten-pills`, and for convenience during development one can produce a testnet-configured Arvo via:

```
$ nix-build nix/pkgs -A arvo-ropsten --no-out-link
```

and that can be copied into a running testnet ship, etc.

jtobin:
I've rebased this branch locally and removed the testnet config changes from the history, so I'll merge that one instead.

pkg/arvo/gen/azimuth-sources.hoon
philipcmonk: You can't do `/j/=sources` because the `=`'s don't get counted right (ie it becomes %home instead of the ship name).

This syntax was considered best practice once upon a time because the vane is "a part of the path" in the sense that it's required to make it a referentially transparent name.  I still prefer that, but I guess that's changed.

philipcmonk:
I don't understand what happened in this rebase, but it appears to have caused a significant divergence between master and what's on mainnet.  At least this commit has been lost, not sure if there are others: https://github.com/urbit/urbit/commit/dd198d03dce22513cd511981e768f129361924e0

This is why os1 ships behind a NAT work very inconsistently.

vvisigoth:
whoa, good catch.
—
~poldec-tonteg
http://urbit.org

matildepark:
nb. OS1-rc is >500 commits behind master, and it, I think, uses the Contacts FE branch as its base, which started work in the latter half of Q4 2019. I'm unsure about its history of rebasing master back into it.

philipcmonk:
Yeah, I checked this is an issue on master as well.

PR #2114, opened 2019-12-18
hoon: support =/ (and friends) without face

Fang-:
This has been getting on my nerves for the longest time, and today I finally snapped. Turns out it really isn't that hard!

This updates the `+wise` parser, used for `=;` and `=^` (and everything that compiles down to them, which is just `=/` I think) to accept molds (`+wyde`) without a specified face.

This allows for constructions such as:

```hoon
=/  [x=@ y=@]
  [5 6]
-  ::  [x=5 y=6]
```

Note that if you want to use "`some-type`" this way, you need to prefix with the `,` to avoid having it parsed as a face instead.

```hoon
=/  ,octs
  [2 'hi']
-  ::  [p=2 q='hi']
```

I've gone ahead and replaced all instances of `=+  ^-` with `=/`, and similarly cleaned up my own todo's for doing the same with `=-  ...  ^-` and `=;`.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:802628572edd2f8305dbbb723215fec89f46a9a63f40f4a0421cfb38f0285b0b
-size 9646301
+oid sha256:7f7f34e9ca7fe2b45c96e34e1bbef09801cb85c19c6a5114b57b7560019bf788
+size 9646645
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7f7f34e9ca7fe2b45c96e34e1bbef09801cb85c19c6a5114b57b7560019bf788
-size 9646645
+oid sha256:23ec7d497dd53f180b8753bd196f92695c7d60dad52f2b57061fc8642cdae4e2
+size 9645510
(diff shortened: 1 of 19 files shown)


ohAitch:
fwiw =/ ,foo seems like a code smell, in that particular case =+ ^- seems
more explicit about "we're doing a weird thing that is going to introduce
nonlexical lexical bindings"

love the [x=@ y=@] common case tho

belisarius222:
I think I like this. I do find myself wanting something like it pretty frequently.

=/ ,foo doesn't seem any less explicit than =+ ^- foo. They're both unambiguous.

—
~rovnys-ricfer
https://urbit.org

ohAitch:
I think I mean something like "uses =+", which with this change gets
relegated _exclusively_ to low-level "we're adding limbs to your subject
and not telling you which ones"

As opposed to being an understated character insertion off the very common
"=/ foo"

anyway definitely seems correct to support it for consistency, just
something to frown on in code review :P

philipcmonk:
Big fan of this, and thanks so much for migrating existing usage.

jtobin:
ping @pilfer-pandex

philipcmonk:
Note that while this is theoretically perfectly OTAable, it should be thoroughly tested before OTA.  For example, if the reset doesn't happen, then all code that uses it will be broken.

philipcmonk:
This is *not* OTA-able as written, since we can't compile the new hoon.hoon with the old compiler.

Fang-:
It _would_ be OTA-able if we exclude the syntactic changes to hoon.hoon, right? Seems like relatively straightforward staging, which could even just be at the commit instead of file level (ie 21ac0e5 first, then ea7c1db should apply fine).

PR #2116, opened 2019-12-18
chat: validate chat names

### LINKED ISSUE #2092: Cannot create Channel
Hey guys, 

I just tried to create a channel and it seems like Landscape has issues with it:
Screenshot 2019-12-13 at 10 25 29

matildepark:
Add a regex check for non-lowercase, non-slash, non-hyphen characters,
or numerals, and a quick check for starts-with-slashes, double-slashes,
to a boolean before creating a chat at that path.

Closes #2092.


ohAitch:
might want a 0-9 in there for alphanumeric

also, was there an issue with --

matildepark:
Ah, you're right! I keep forgetting you can totally use numbers.

The only fail case I've truly seen is when you start with a slash — it breaks chat.js outright even though the path gets created.

jtobin:
I'll shortcut @loganallenc here as this one looks clear enough.

Issue #2109, opened 2019-12-17
Ford cache doesn't invalidate correctly


philipcmonk:
See #2099.  Until this is fixed OTAs can't safely depend on changes to zuse.

One repro is to try to boot from https://bootstrap.urbit.org/urbit-v0.10.1-old.pill on livenet.  If you boot from it with a comet, run `|merge %base ~marzod %kids, =gem %that`.  If `-eth-watcher` crashes on line 12, that means everything works.  If it crashes in lib/ethio, that means it's using an old version of zuse.  You may have to run `|start %goad` and/or `:goad %go` to make sure it tried to recompile all the apps.

PR #2117, opened 2019-12-18
ford: clear build results on +load

belisarius222:
Fixes https://github.com/urbit/urbit/issues/2109 by clearing the cache whenever Ford reloads.  Note that this also assumes Ford will be reloaded every time Zuse reloads.  That's an assumption we might want to relax eventually, but for now this, in combination with the recent :goad call from :dill, should fix OTAs.

I've confirmed that this reloads Ford properly on a fakezod.  Testing the bug repro on a comet locally at the moment.

### DIFF
--- a/bin/brass.pill
+++ b/bin/brass.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:18d492d912068e7fefef48006105d39c1c8f56aa756b7aeae48387c2254c1b91
-size 7153239
+oid sha256:407a763f44eb91db0dd4a1ec2dbd12ed4332b48decefd3999c4313844daa2c0b
+size 7226043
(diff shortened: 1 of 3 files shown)


belisarius222:
I haven't been able to trigger the reproduction exactly, but I did boot a comet off that URL, load in new Ford, and |reset a couple times, in addition to reloading Zuse and Ford a bunch of times on fakezods.

It all seems to be working swimmingly, although to be on the safe side I think we might also want to add a :goad whenever Ford reloads, not just when Zuse reloads.

jtobin:
> It all seems to be working swimmingly, although to be on the safe side I think we might also want to add a :goad whenever Ford reloads, not just when Zuse reloads.

@belisarius222 I couldn't tell if you wanted to add the `:goad` call before this one gets merged.  It seems like a strict improvement as-is, in the meantime, so I'm inclined to pull the change in anyway.

belisarius222:
I think this should be merged as-is

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
Goad listens to /sys, so it'll fire whenever ford (or indeed any vane)
changes. This is more than necessary, but safe.

PR #2118, opened 2019-12-18
build: allow baking multiple pills into docker images

brendanhay:
By baking (potentially) multiple pills into an image, we can provide mainnet vs ropsten images. It is still up to the operator to pass along the pill path(s) to the entrypoint.

For example, using Docker:

```
docker run --tty urbit -B /share/brass.pill -J /share/ivory.pill ...
```

The main `image` attribute still uses only the solid pill and `image-ropsten` has been provided with brass and ivory pills.

Additionally, herb has been added to the image for convenience.

### DIFF
--- a/nix/ops/default.nix
+++ b/nix/ops/default.nix
@@ -74,8 +74,13 @@ rec {
   };
 
   image = import ./image {
-    inherit pkgs urbit;
-    pill = bootsolid;
+    inherit pkgs herb urbit solid;
+  };
+
+  image-ropsten = import ./image {
+    inherit pkgs herb urbit;
+    brass = brass-ropsten;
+    ivory = ivory-ropsten;
   };
 
 }
--- a/nix/ops/image/default.nix
+++ b/nix/ops/image/default.nix
@@ -1,6 +1,17 @@
-{ pkgs, urbit, pill }:
+{ pkgs
+, herb
+, urbit
+, solid ? null
+, brass ? null
+, ivory ? null
+}:
 
-pkgs.dockerTools.buildImage {
+let
+  link = pill: path:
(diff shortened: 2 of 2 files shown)


ohAitch:
Oof, seems like a lot of extra megabytes for one hex constant. And I
wouldn't trust gzip to pick it up either given ++jam's bitwise nature…

Hm actually, given the dag backreference mechanism, would there be any way
to cons all six of these into an everything.pill, and pull them out by axis
at boot time? (alternately are the pills eclipsed by the vere binary size
st this is not a real concern?)

joemfb:
We should really just inject the contract addresses in a boot event.

brendanhay:
In case it wasn't clear the two image attributes have the following pills baked into their respective docker images:

* `image` has _only_ the solid pill under `/share/solid.pill`.
* `image-ropsten` has _both_ the brass and ivory pills under `/share/brass.pill` and `share/ivory.pill`, respectively.

`image` is (potentially) distributed via `https://hub.docker.com` and `image-ropsten` is used by hosting code to deploy testnets. In the latter's case, the image is actually only downloaded once per phyiscal cluster node and booted multiple times.

In terms of resulting image sizes, there's far larger gains to be had by building against `musl-linux64` or using the cross-compilation expressions. Having differing pills due to contract addresses is annoying but not impeding anything currently. If the addresses were configured via something that didn't require a build - life would be better(er).

ohAitch:
Re: musl-linux, aren't the dynamically linked images going to have smaller
image _diff_ sizes? Like the test nodes will already have all the docker
layers from the previous testnet, right?

But yeah it sounds like small-o of other things either way.

Issue #1903, opened 2019-10-31
clay: %u product-type is unusable


joemfb:
It produces `~` if a case exists, and blocks if you ask for a non-existent case, while accepting but ignoring the path:

```
> .^(* %cu /==/(scot %da now))
0
> .^(* %cu /==/(scot %da now)/foo)
0
> .^(* %cu /==/(scot %da now)/foo/bar)
0
> .^(* %cu /==/(scot %da now)/foo/bar/baz)
0
```

The product type is not useful, as it's not a constant:

```
.^(~ %cu /==/(scot %da now))
%nets-failed
%nets-need
$~
%nets-have
@n
```

It seems like this would be much more useful if it checked the path.

philipcmonk:
Formally, the idea is that every path exists (indeed, it exists as both a "file" and a "directory") so long as the case exists.  Then, you can use .^(arch %cy %/foo) to get an +arch which will tell you if the path has any contents and/or children.

I could see an argument of using %u for checking whether a path has content, but I don't know the natural answer to what if it's a "directory" but not a "file".

joemfb:
Ah, I didn't realize this. The `%y` care already covers this use-case. `%u` doesn't seem particularly useful, as the only information it can convey is whether it blocked or not (and, at least in an app, you won't be able to check!). In any case, this is not an issue.

joemfb:
Revising that, the product of `%u` remains an issue.

joemfb:
#1534 is somewhat relevant.

philipcmonk:
Ah, yeah, it shouldn't block when the case doesn't exist.  I sort of feel like the point of `%u` is to check whether the scry will block, but of course we can't produce that answer in a referentially-transparent way.

Maybe it would be better to support `.^(? %u %/test-case)` to check whether `test-case` exists right now (and crash if the formal case is not equal to now).

ohAitch:
What about `?`, which is `|` if the desk doesn't exist and the case is in the past(or otherwise prior to `/==0`), `&` if it already does/did, and blocks until the referenced time otherwise? I can see the argument for "reject non-time case", while "label or number always returns `&` resolving the build whenever relevant" is coherent it may not be exactly useful.

If it's not doing `?=(^ file:.^(arch cy+...))` it should probably crash on non-`/` spur rather than ignoring it though.

philipcmonk:
fixed in #2119

Issue #1546, opened 2019-08-09
!> should produce a +hypo


belisarius222:
Right now the `!>` rune produces a vase of its argument, losing all type information about the tail.

I think it would be a little better for, say, `!>(*case:clay)` to produce a `(hypo case:clay)` instead of ``[-:!>(*case:clay) `*`*case:clay]``. Whatever the original value's type is, the type of the new pair will always nest in `+vase`, and we could avoid [this kind of repetitive code in Clay](https://github.com/urbit/arvo/blob/master/sys/vane/clay.hoon#L2533-#L2534).

It might even make vases easier to print nicely.

philipcmonk:
FWIW, you can get this effect with `!;(*type value)`

PR #2119, opened 2019-12-18
clay: implement %u care

Fang-:
Previously, it would always produce ~, regardless of the path asked
about.

Now, it produces a loobean, based on whether or not a file exists at the
specified path.

This largely mimics the surrounding code. It takes care to specify the actual output type (with a fake `%loob` mark) in the `^-` of `+read-u`.
Specifying just `cage` there allows us to do a simple `!>` on the actual result. However, with the `hypo` construction, this doesn't nest. I'm not sure why, so we just do as the surrounding code does: get the type using the type spear, and put in the result noun as-is. This can probably be simplified, happy to take suggestions.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:d2c08caf6f69f9fa349003da923d0b6507a8b6df763a0ee491f195a937630843
-size 9620244
+oid sha256:4185a43f4acbcc6f70fb9ac7c68b4b04a08545ff9a20f5c61d1542feb48fdb5a
+size 14078527
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -3890,12 +3890,18 @@
     ::  at any of its children.
     ::
     ++  read-u
-      |=  {yon/aeon pax/path}
-      ^-  (unit (unit (each {$null (hypo ~)} lobe)))
-      =+  tak=(~(get by hit.dom) yon)
-      ?~  tak
+      |=  [yon=aeon pax=path]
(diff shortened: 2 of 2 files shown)


belisarius222:
`!>` produces just a plain vase, not a hypo.  This means the tail of the result is just %noun, not whatever type it was originally.  I have a github issue complaining about this: https://github.com/urbit/urbit/issues/1546

belisarius222:
Is it ok that we produce a fake mark here?  Also, I think it might be better to call the mark %flag instead of %loob.

Fang-:
Note that other existing `+read-` arms already produce fake marks. `%file-list` in `+read-t`, `%dome` in `+read-v`.

We could take the approach that `+read-p` takes and just use a `%noun` mark?

belisarius222:
Well, we use those other scrys that produce fake marks, and they seem to work fine, so this should be fine too.

Fang-:
Alright. I'll rename to `%flag` as suggested (as consistent with `@f`) and rebase onto master.

> The original conceit here was that every file/directory exists as long as that commit exists.

Really makes me think.

joemfb:
relevant discussion in #1903

Fang-:
Thanks for the reference @joemfb. It seems the implementation here isn't too far off from the last two suggestions there. `%u` as implemented here is, indeed, pretty much `?=(^ fil:.^(arch %cy path))`. Perhaps that means we don't need it at all, but having a shorthand scry for it in `%u` feels kinda nice to me.

philipcmonk:
If %u is "always true or crash", then it's basically useless.  I think this is a more useful definition of it.

joemfb:
I agree that we `%u` should be changed or removed. I don't have strong feelings about this, and @philipcmonk is better suited to judge the API. I just wanted to share that context.

PR #2135, opened 2019-12-23
webchat: render "me" messages in italics

Fang-:
Tiny fix. No idea how/why this got lost.

image


matildepark:
> +70,520

`gulp bundle-prod`, ~palfun!

Fang-:
If we can ship a 50MB `.mp4`, we can ship a couple lines of unminified JS.

matildepark:
I think the regression happened during the chat-store Interface rewrite. I mean, it was a rewrite from scratch though. Seems expected to see a regression in the little nooks and crannies. (I think @me messages still present as no message in the sidebar.)

Issue #1710, opened 2019-09-09
Arvo could use a proper build


jtobin:
I recently closed #1658, which sought to ignore the compiled CSS and JS files from Landscape that are currently tracked in git -- this issue is a continuation of that.

At present, as tracked in git, Arvo has no build.  Except in practice it sort of does, in that one has to build the aforementioned Landscape files externally to Arvo and dump them into the Arvo tree.  Similarly, we keep a number of "heavier" pills checked into git LFS that, for lack of a better term, go above and beyond what is technically required of them (see this [discussion on slim pills, etc.](https://groups.google.com/a/urbit.org/forum/#!topic/dev/Cz222LxMYOE) on urbit-dev).

Keeping more of these build artifacts in revision control than is absolutely necessary is a dubious practice.  They are non-diffable, so they cause frequent merge conflicts; ensuring their state is synced with source on any given commit is clumsy, manual, and error-prone; they are difficult to audit, and thus they pose security vulnerabilities.  And so on.

There's probably no way around tracking a so-called slim pill in the repository itself, but these additional build artifacts should not be tracked, and we should have a one-step build that produces them.

jtobin:
(Worth linking #1664.)

jtobin:
I think this one can be closed now that Landscape's minified JS is not included in Arvo's source.

PR #2138, opened 2019-12-24
interface: use production build of react

matildepark:
We were shipping development builds of React in production. Now we're not. Affects Chat, Publish, Soto. (Should be pulled into Contacts and Links branches.)

- Also adds parity for rollup's 'named exports' so that indigo-react can be used seamlessly in future builds of all applications without running into build errors.

Apparently the performance increase can be 3-5x, but I don't really know how to measure that with much accuracy. From just using the "Performance" tab in Chrome, basic actions in a 6s window went from 981ms of scripting to 103ms. We'll see how it actually feels to use.


matildepark:
I agree, although ... I'm unsure how to perform a partial rollup query and then throw the remainder of the work to another task. What do you suggest? My initial thought is to restructure the gulp pipe work into a common function that each variant calls...

jtobin:
> I'm unsure how to perform a partial rollup query and then throw the remainder of the work to another task.

I went over this tonight and cleaned things up a bit using ⚡️ *functional programming* :zap:.  WIll put finishing touches on it tomorrow and push it out.

jtobin:
> WIll put finishing touches on it tomorrow and push it out.

Please hold, with a bit more work I can solve this together with the long-standing #1664 and #1710..

litpub:
There's an issue related to this @jtobin @matildepark which is that `sucrase` has a production config that is not enabled. See: https://github.com/alangpierce/sucrase/pull/270

On HEAD of os1-rc this results in:

1. ~12kb delta on the chat app
2. Not including the personal filepaths of developers in the production build (if you go look at the production builds you'll see personal filepaths baked in - https://raw.githubusercontent.com/urbit/urbit/b6120ac9fc640c43e9dae8a9ec95fc923cb4f692/pkg/arvo/app/chat/js/index.js ctrl+f for `matilde` here). Basically, the `createElement` in React is designed to accept a `__source` prop for local debugging.

This can be solved by doing 

``` 
gulp.task('jsx-transform', function(cb) {
  return gulp.src('src/**/*.js')
    .pipe(sucrase({
      transforms: ['jsx'],
      production: true
    }))
    .pipe(gulp.dest('dist'));
});
```

But it should be sensitive to only the production build, though. I suspect the use of `         rollupReplace({'process.env.NODE_ENV': JSON.stringify('production')}),` in this PR could be pulled out and made more generally available for reference by other tasks so prod builds always use NODE_ENV = production across any task?

matildepark:
@litpub I believe @jtobin was about to wrap a functional-oriented rewrite of the gulp scripts in the near-term that may address this ...

jtobin:
On Tue, Feb 18, 2020 at 08:14:05PM -0800, matildepark wrote:
> @litpub I believe @jtobin was about to wrap a functional-oriented
> rewrite of the gulp scripts in the near-term that may address this ...

Yeah, this (+ some associated stuff) has been mostly done and sitting
for awhile now.  Will try to wrap it up this week.

loganallenc:
This is a small and pleasant quality of life improvement that decreases the size of our minified JS files. I think this should be merged today, as it has no dependencies on any other work.

matildepark:
Since @jtobin is out and refactoring his Landscape build process I approve of this merged for the time being. It's definitely a performance improvement.

ixv:
Seems fine, conflicts need to be fixed though.

matildepark:
Are one of you able to resolve them, test the build and merge it if all’s well? @loganallenc @ixv They seem straightforward, just a bit occupied atm

—
~haddef-sigwen
https://urbit.org

loganallenc:
fixed

loganallenc:
@ixv ready

PR #2144, opened 2019-12-27
chat: add dark mode

matildepark:
Adds custom classes and declarations for dark mode color scheme.
Uses [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query to check preference.

Note that prefers-color-scheme is set to 'dark' when:
- The operating system (mobile or desktop) is in 'dark mode' or
- Your web browser is set to 'dark mode'

Screen Shot 2019-12-27 at 12 10 30 PM
Screen Shot 2019-12-27 at 12 12 35 PM


pkg/interface/chat/src/css/custom.css
```diff
@@ -141,4 +141,46 @@ h2 {
   .h-100-minus-48-xl {
     height: calc(100% - 48px);
   }
+}
+
+/* dark */
+
```
urcades: Another thing to consider for a dark theme are images set at a lower opacity, perhaps around 60%.

This is another eyesight saver — nothing punches a hole in a dark theme like a bright ass image coming into view.

urcades: Everything else seems solid tho, excited to see this live.

jtobin:
ping @loganallenc

jtobin:
Gonna merge this one as I reckon there's not much controversy around it.

PR #2150, opened 2020-01-04
hoon: implement signed int parsing

pkova:
Needed this for advent of code day 5, seems useful enough.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -5605,6 +5605,15 @@
 ++  but  (cook |=(a/@ (sub a '0')) (shim '0' '1'))      ::  binary digit
 ++  cit  (cook |=(a/@ (sub a '0')) (shim '0' '7'))      ::  octal digit
 ++  dem  (bass 10 (most gon dit))                       ::  decimal to atom
+++  dsm                                                 ::  decimal to signed int
+  %+  cook
+    |=  [s=? n=@]
+    ^-  @s
+    ?:  s
+      (mul n 2)
+    (dec (mul n 2))
+  ::
+  ;~(plug ;~(pose (cold | hep) (easy &)) dem)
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -5605,6 +5605,15 @@
 ++  but  (cook |=(a/@ (sub a '0')) (shim '0' '1'))      ::  binary digit
 ++  cit  (cook |=(a/@ (sub a '0')) (shim '0' '7'))      ::  octal digit
 ++  dem  (bass 10 (most gon dit))                       ::  decimal to atom
+++  dsm                                                 ::  decimal to signed int
+  %+  cook
+    |=  [s=? n=@]
+    ^-  @s
+    ?:  s
+      (mul n 2)
+    (dec (mul n 2))
```
Fang-: You can replace this with just `new:si`, which is the existing stdlib function for "unsigned to signed", and matches the function signature used here.

pkg/arvo/sys/hoon.hoon
```diff
@@ -5605,6 +5605,15 @@
 ++  but  (cook |=(a/@ (sub a '0')) (shim '0' '1'))      ::  binary digit
 ++  cit  (cook |=(a/@ (sub a '0')) (shim '0' '7'))      ::  octal digit
 ++  dem  (bass 10 (most gon dit))                       ::  decimal to atom
+++  dsm                                                 ::  decimal to signed int
```
Fang-: Might be nice if we can maintain the `cvc` style naming here. Maybe `sem`?

ohAitch:
Yeah this definitely goes into a library not hoon.hoon: 4i is for lexing
/the hoon grammar/ specifically

pkova:
Thanks guys, I'll be back when I figure out where to put this.

`@s` parsing seems to be done in `tash:so`.

PR #2163, opened 2020-01-10
king: add partial-replay and --dry-from

philipcmonk:
partial-replay allows you to replay up to event N and make a checkpoint
there.  --dry-from replays to event N and then starts in dry-run mode
from that point, so you can attach a console.

The intended debug flow is to use partial-replay to make a checkpoint
near when a bug is triggered, then use event-browser and --dry-from to
drill down and find exactly what triggered the bug.

Unfortunately, if the bug was triggered by crash handling, you do not
have the original event, only the injected replacement event.  This is
still useful, though.

Please particularly review for style, I am unfamiliar with Haskell idioms.


benjamin-tlon:
Oh! Didn't see this.

Email me next time?

benjamin-tlon:
I'll update this branch tomorrow.

pkg/hs/king/lib/Ur/King/Main.hs
```diff
@@ -134,12 +135,12 @@ toSerfFlags CLI.Opts{..} = catMaybes m
...
 toPierConfig pierPath CLI.Opts{..} = PierConfig
     { _pcPierPath = pierPath
-    , _pcDryRun   = oDryRun
+    , _pcDryRun   = (oDryRun || isJust oDryFrom)
     }
 
 toNetworkConfig :: CLI.Opts -> NetworkConfig
 toNetworkConfig CLI.Opts{..} = NetworkConfig
-    { ncNetworking = if oDryRun then NetworkNone
+    { ncNetworking = if (oDryRun || isJust oDryFrom) then NetworkNone
```
benjamin-tlon: All of this logic is pretty messy. There should really be a proper transformation pass from the CLI options into a data structure representing all of config options.

Not your fault, though.

pkg/hs/king/lib/Ur/King/Main.hs
```diff
@@ -185,8 +189,8 @@ runOrExitImmediately getPier oExit =
 tryPlayShip :: ( HasLogFunc e, HasNetworkConfig e, HasPierConfig e
                , HasConfigDir e
                )
-            => Bool -> Bool -> Serf.Flags -> RIO e ()
-tryPlayShip exitImmediately fullReplay flags = do
+            => Bool -> Bool -> Maybe Word64 -> Serf.Flags -> RIO e ()
+tryPlayShip exitImmediately fullReplay playFrom flags = do
```
benjamin-tlon: This is bad code. If you look at the type of this function, you have no idea what the arguments are for. Generally, the solution to this is to create a record or to use newtypes to make the types more descriptive.

Again, not your fault. This was already bad, just making style notes since you asked for them.

pkg/hs/king/lib/Ur/Vere/Pier.hs
```diff
@@ -130,18 +130,36 @@ booted pill lite flags ship boot = do
...
+resumed event flags = do
     top    <- view pierPathL
+    tap    <- case event of
+                Nothing -> pure top
+                Just ev -> rio $ do
+                  snap <- getSnapshot top ev
+                  case snap of
+                    Nothing -> pure top
+                    Just sn -> pure sn
+    rio $ logTrace $ displayShow tap
```
benjamin-tlon: Minor style nits: generally wouldn't make a big deal out of this, but you asked for style feedback.

The `snap` variable is useless. It doesn't need a name since it's clear what comes out of `getSnapshot`. I would write this with `\case`:

```

    tap    <- case event of
                Nothing -> pure top
                Just ev -> rio $ do
                  getSnapshot top ev >>= \case
                    Nothing -> pure top
                    Just sn -> pure sn
```

The `pure top` bit is repeated. This can be made cleaner with `MaybeT`:

```
import Control.Monad.Trans.Maybe
...
    tap    <- fmap (fromMaybe top) $ rio $ runMaybeT $ do
                  ev <- MaybeT (pure event)
                  MaybeT (getSnapshot top ev)
```

Your code is fine, but it's faster for me to read after these changes:

- Ok, I'm doing something that might fail and I will fall back to `top` in any edge case.
- Ok, first I'll fail if `event` is `Nothing`.
- Ok, Then I'm getting the snapshot, which might also fail.

pkg/hs/king/lib/Ur/Vere/Pier.hs
```diff
@@ -130,18 +130,36 @@ booted pill lite flags ship boot = do
...
+getSnapshot :: String -> Word64 -> RIO e (Maybe String)
+getSnapshot top last = do
+  createDirectoryIfMissing True $ top <> "/.partial-replay/"
+  snapshots <- listDirectory $ top <> "/.partial-replay/"
+  let nums = Ur.Prelude.mapMaybe readMay snapshots :: [Int]
+  let snaps = sort $ Ur.Prelude.filter (\n -> n <= (fromIntegral last)) nums
+  case (Data.List.uncons $ reverse snaps) of
+    Nothing      -> pure Nothing
+    Just (sn, _) -> pure $ Just $ top <> "/.partial-replay/" <> (show sn)
+
```
benjamin-tlon: I would write it like this:

```
getSnapshot :: forall e. FilePath -> Word64 -> RIO e (Maybe FilePath)
getSnapshot top last = do
    lastSnapshot <- lastMay <$> listReplays
    pure (replayToPath <$> lastSnapshot)
  where
    replayDir        = top </> ".partial-replay"
    replayToPath eId = replayDir </> show eId

    listReplays ∷ RIO e [Word64]
    listReplays = do
        createDirectoryIfMissing True replayDir
        snapshotNums <- mapMaybe readMay <$> listDirectory replayDir
        pure $ sort (filter (<= fromIntegral last) snapshotNums)
```

pkg/hs/king/lib/Ur/Vere/Serf.hs
```diff
@@ -437,12 +441,21 @@ replayJobs serf lastEv = go Nothing
...
+    let numEvs = case last of
+                   Nothing -> ssNextEv ss
+                   Just la -> la - (ssNextEv ss) + 1
+
+    lastEv <- case last of
+                Nothing -> Log.lastEv log
+                Just la -> pure la
+
     runConduit $  Log.streamEvents log (ssNextEv ss)
+               .| CC.take (fromIntegral numEvs)
```
benjamin-tlon: Nice

pkg/hs/king/lib/Ur/Vere/Serf.hs
```diff
@@ -437,12 +441,21 @@ replayJobs serf lastEv = go Nothing
...
     ss <- handshake serf (Log.identity log)
 
-    lastEv <- Log.lastEv log
+    let numEvs = case last of
+                   Nothing -> ssNextEv ss
+                   Just la -> la - (ssNextEv ss) + 1
+
+    lastEv <- case last of
+                Nothing -> Log.lastEv log
+                Just la -> pure la
```
benjamin-tlon:
```
lastEv <- last & maybe (Log.lastEv log) pure
```

pkg/hs/king/lib/Ur/Vere/Pier.hs
benjamin-tlon: I don't quite understand what this does by looking at the code. Could this be named `lastPartialReplaySnapshot`?

pkg/hs/king/lib/Ur/King/Main.hs
```diff
@@ -286,6 +290,42 @@ collectAllFx top = do
...
+      snap <- Pier.getSnapshot top last
+      case snap of
+        Nothing -> pure ()
+        Just sn -> do
+          let start = top <> "/.partial-replay/"
+          liftIO $ system $ "cp -r \"" <> sn <> "\" \"" <> tmpDir <> "\""
+          pure ()
+
+    tmpDir :: FilePath
+    tmpDir = top <> "/.partial-replay/" <> show last
```
benjamin-tlon: top </> ".partial-replay" </> show last

pkg/hs/king/lib/Ur/King/Main.hs
```diff
@@ -286,6 +290,42 @@ collectAllFx top = do
...
+        logTrace "Done replaying events!"
+  where
+    fetchSnapshot :: RIO e ()
+    fetchSnapshot = do
+      snap <- Pier.getSnapshot top last
+      case snap of
+        Nothing -> pure ()
+        Just sn -> do
+          let start = top <> "/.partial-replay/"
+          liftIO $ system $ "cp -r \"" <> sn <> "\" \"" <> tmpDir <> "\""
```
benjamin-tlon: This is gross and can be broken (file paths maybe contain `"`). However, there doesn't seem to be anything off-shelf for copying directories.

benjamin-tlon: `storeSnapshot`?

benjamin-tlon: Should probably make sure `.partial-replay` directory exists first.

pkg/hs/king/lib/Ur/King/CLI.hs
```diff
@@ -204,11 +209,10 @@ new = do
 
 opts :: Parser Opts
 opts = do
-    oAmesPort  <- option auto $ metavar "PORT"
+    oAmesPort  <- optional $ option auto $ metavar "PORT"
                              <> short 'p'
                              <> long "ames"
                              <> help "Ames port number"
-                             <> value Nothing
                              <> hidden
```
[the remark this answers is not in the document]
philipcmonk: It didn't work before, IIRC it couldn't actually parse what it needed to.

pkg/hs/king/lib/Ur/Vere/Pier.hs
philipcmonk: Let me make sure I understand the types:

- `fmap` is in `RAcquire`, so everything after the first `$` needs to be an `RAcquire e (Maybe a)`.
- `rio` takes an `RIO a` and converts to `RAcquire e a`, so we now need an `RIO (Maybe a)`.
- `runMaybeT RIO` takes a `MaybeT RIO a` and converts to `RIO (Maybe a)`, which brings us to the `do` block, which uses to the `MaybeT RIO a` monad.
- We get the event number by creating a `MaybeT RIO a` with `MaybeT` applied to a `RIO (Maybe a)`, so the `pure` is in the `RIO` monad.  This step is necessary since `event` is a `Maybe a`, so this is will short-circuit to `top` here if `event` is `Nothing`.
- Then we get the snapshot by doing the same thing, except instead of `pure` we use `getSnapshot` (because there's actually IO).

pkg/hs/king/lib/Ur/King/Main.hs
philipcmonk: `Pier.getSnapshot` does create `.partial-replay` it it's not there, though it's not obvious that it would do that.

philipcmonk:
Okay, I made those changes, PTAL @benjamin-tlon

pkg/hs/king/lib/Ur/Vere/Pier.hs
benjamin-tlon: Yep! That's all correct.

This is concept heavy code, so I might not write this if I was trying to make my code accessible to people less familiar with these idioms. Also, if the logic itself is complicated, sometimes I'll still prefer the explicit-nested-pattern-matching styles.

But, once you're familiar, you mostly don't have to think through this. I'll do basically exactly the process that you wrote down if I made a mistake writing something like this, but when I read it, I already know it type checks so I just filter out most of the conversions in my mind, so I basically just see `fromMaybe top ... rio ... maybeT ... ev <- event ...getSnapshot top ev`.

pkg/hs/king/lib/Ur/King/Main.hs
benjamin-tlon: Right. It seems wrong to rely on that.

Also, `Pier.getSnapshot` doesn't actually need to do that. It can just handle the case where the directory doesn't exist and return `Nothing`.

benjamin-tlon:
LGTM :+1:

benjamin-tlon:
Merge? @philipcmonk

philipcmonk:
Fine by me. I figured you'll do it since it's sort of your branch?

PR #2166, opened 2020-01-12
ames: actually maintain direct routes

belisarius222:
There was a typo in the routing logic that was comparing equality against a value where it should have been doing a pattern match.  The value compared against contained the literal `*` gate, which would never match `route.peer-state`, so this condition was always true, meaning the fix that had added this  extra condition (https://github.com/urbit/urbit/commit/5406f0609273fa4db1bd84bf600a401d1c114ae2#) did not actually change the behavior from what it been previously.

If only this had caused a type error ...


belisarius222:
@philipcmonk PTAL, I'm pretty sure I got it right this time

PR #2169, opened 2020-01-14
u3: clear cached mugs on cell edit

joemfb:
This PR fixes a bug wherein cached mugs were wrongfully retained on editing cells, producing a jet mismatch in `+mug`. @eglaysher found the mismatched values while tracking down a potentially related bug in jaque.

The extent of the impact of fixing this bug is not yet clear; many things depend on mugs, and it's possible that something was depending on a (deterministically) wrong mug. More testing will be required before these changes can be released.

/cc @frodwith


frodwith:
lol whoops!

joemfb:
This PR was reverted on master, but it should not have been. It can't be released as a hotfix (without the ability to force event-log truncation), so it'll have to go into the next breach.

PR #2177, opened 2020-01-16
nix: enable ASan during tests

joemfb:
This PR enables ASan unconditionally in the relevant nix derivations, so as to run it in CI. The right approach is to parameterize the build and test derivations so that the appropriate flags and environment variables can be threaded through. (I only know how to do that in a dumb, explicit way, so I've held off. Maybe `pkg.overrideAttrs` could help?)

This currently includes merges of various open PRs on which it depends, I'll rebase those out once they're merged.

### DIFF
--- a/nix/ops/fakeship/builder.sh
+++ b/nix/ops/fakeship/builder.sh
@@ -2,11 +2,19 @@ source $stdenv/setup
 
 set -ex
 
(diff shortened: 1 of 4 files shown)


tacryt-socryp:
Is this PR still relevant? @joemfb

PR #2179, opened 2020-01-21
/lib/dbug agent wrapper

Fang-:
This implements `/lib/dbug`, an agent wrapper in the style of `/lib/verb`, that gives you some generic ways of looking at application state. More concretely, you can look at the entire bowl, or see if there are any subscriptions matching your specified conditions.

Example:

```hoon
> :chat-store +dbug %incoming ship+our
[ i=[duct=[i=/g/use/chat-view/~zod/out/~zod/chat-store/updates t=~[/d //term/1]] ship=~zod path=/updates]
  t=[i=[duct=[i=/g/use/chat-cli/~zod/out/~zod/chat-store/chat-store t=~[/d //term/1]] ship=~zod path=/updates] t=~]
]
```

The lists don't always print very prettily (ie above, they explicitly include list faces), but trying to cast those always kills the pretty-printer in a weird way. In any case, this should suffice for now.

I went ahead and included this in a couple applications, also adding `/lib/verb` where it wasn't in use yet.

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -12,7 +12,7 @@
 /-  *chat-store, *chat-view, *chat-hook,
     *permission-store, *group-store, *invite-store,
     sole-sur=sole
-/+  sole-lib=sole, chat-eval, default-agent, verb,
+/+  sole-lib=sole, chat-eval, default-agent, verb, dbug,
     auto=language-server-complete
 ::
 |%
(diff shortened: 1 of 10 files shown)


philipcmonk:
Nice.  Another nice one, not always applicable but often, would be "print application state", which you can do generically because you have it as a vase (use +sell).

Fang-:
Good call @philipcmonk, I added that in!

PR #2188, opened 2020-01-24
removing black border

jyng:
Removing `bb` from images

@urcades @matildepark

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/message.js
+++ b/pkg/interface/chat/src/js/components/lib/message.js
@@ -71,7 +71,7 @@ export class Message extends Component {
           ></img>
         );
         return (
-          <a className="f7 lh-copy v-top bb word-break-all"
+          <a className="f7 lh-copy v-top word-break-all"
             href={letter.url}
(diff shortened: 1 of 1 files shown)


matildepark:
This removes the border from all links, always.

matildepark:
You want to refactor it so that it applies `bb` only if ytMatch and imgMatch both come in false

matildepark:
You should also test and ensure it actually works that way in your dev environment

pkg/interface/chat/src/js/components/lib/message.js
```diff
@@ -71,7 +71,7 @@ export class Message extends Component {
           ></img>
         );
         return (
```
jyng: @matildepark my understanding of the code is that it, if imgMatch, it returns the the classes in className. It seems to render that correctly?

image

matildepark:
Shit, you're right, sorry — was a little quick on the draw from my phone.

jyng:
@jtobin updated commit message to be more descriptive.

Issue #2180, opened 2020-01-22
Ford crashes during new publish state transition


ixv:
When I try to `|commit %home` the files in the branch `ixv/publish-remix` into a ship based on master, which has already existing publish notebooks something crashes. These crashes do not happen if the ship does not have a notebook.

In the change, publish does the following things in the on-load arm:
- kills old ford builds
- kicks old subscribers
- leaves old subscriptions
- moves old publish files
- creates groups for notebooks
- creates invite store entry
- subscribes to permission-store, and invite-store
- reads from clay
- sends invites to old subscribers

It seems the problematic part of this is the killing of old ford builds.

Something about these crashes is non-deterministic, because over the course of many reproductions it crashes differently.

The first crash is this: https://pastebin.com/FSBVs6fJ
The second crash is this, and this one happens over and over again: https://pastebin.com/xuggUpBN
The third crash is this: https://pastebin.com/KgD0RKmF

The third one is interesting because it appears to be depending on an old version of the publish sur file.

philipcmonk:
Publish makes heavy use of Ford's ability to include dependencies.  @belisarius, want to take a look at this?

belisarius222:
Yeah, I'm taking a look at this tonight.

Fang-:
@ixv We got an (nondeterministic) compilation success upon committing the new publish files, which proceeded to run the `+on-load`. This resulted in the following output:

```hoon
error in %publish
%load-failed
! /ct/~zod//~2000.1.1/web/publish
/ct/~zod//~2000.1.1/web/publish
error in %publish
%arvo-response
! /~zod/home/0/lib/default-agent/hoon:<[63 3].[63 5]>
! "unexpected system response %f to %publish on wire /post/urbit-only-exclusive-tweets/hilarious-guy-computing-from-boston-freaks-out-when-he-sees-urbit-explicit-language"
! /~zod/home/0/lib/default-agent/hoon:<[62 3].[63 5]>
! /~zod/home/0/app/publish/hoon:<[306 7].[306 28]>
! /~zod/home/0/app/publish/hoon:<[305 5].[339 7]>
! /~zod/home/0/app/publish/hoon:<[304 5].[339 7]>
! /~zod/home/~2020.1.14..16.26.07..02c8/sys/vane/gall:<[1.207 9].[1.207 47]>
```

(With the `error in %publish` and everything below it repeating a bunch of times.)

[...]

PR #2192, opened 2020-01-25
ford: dequeue orphans

belisarius222:
@ixv recently uncovered a bug (https://github.com/urbit/urbit/issues/2180) in Ford that caused certain rebuilds to crash.  @Fang- and I believe this change should fix the bug, and we have confirmed that the reproduction that used to fail about two thirds of the time now has not failed at all in the ten or so times we've run it since then.  @Fang- is still running more tests to confirm the fix with more certainty.

It turned out the cause was that (depending on the rebuild order, which is unspecified and should not need to be specified), Ford could enqueue a provisional sub-build to be run but then, later in the same +gather call, discover that the sub-build was in fact an orphan and delete it from `builds.state` accordingly.  Then when Ford tried to run the sub-build, it would have already been deleted from the state, so Ford would crash when trying to process its result in `+reduce`.

The fix was to make sure that when we discover a provisional sub-build is orphaned, dequeue it from `candidate-builds` and `next-builds` to make sure we don't try to run it.  I'm about 95% sure this fix completely solves the bug.

I worried a bit about whether asynchronicity could break this check somehow -- if we run +gather across multiple Ford activations because something blocked, could we still end up running an orphaned build?  But I think the answer is no, because whenever Ford determines a build is an orphan, it'll make sure we don't try to run it in that event.  If a future event were to enqueue the build to be run, it could only be because some other build had established this build as a sub-build, in which case it would have to add it to the state again before enqueueing it.  If this analysis holds, as I believe it does, we're in the clear.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:e6232e5bcb64be057ccd9b4c23bbd636814d50778e4f3a01fbf2ee0042fe3d44
-size 9635072
+oid sha256:8ca3d44142cebd027a8c7d7326342f5ca2fca921034293ac1fd05646ea00cc30
+size 9636208
--- a/pkg/arvo/sys/vane/ford.hoon
+++ b/pkg/arvo/sys/vane/ford.hoon
@@ -5817,6 +5817,11 @@
       ?:  verified.build-relation
         ~
       `sub
+    ::  dequeue orphans in case we were about to run them
+    ::
+    =/  orphan-set        (~(gas in *(set ^build)) orphans)
+    =.  next-builds       (~(dif in next-builds) orphan-set)
+    =.  candidate-builds  (~(dif in candidate-builds) orphan-set)
     ::  remove links to orphans in :build's +build-status
     ::
     =^  build-status  builds.state


belisarius222:
Note that this only directly addresses the first and second failure cases in @ixv's bug report.  This seems to have also fixed the third failure case, since we're not seeing it anymore, but @Fang- and I are less confident about that.  We haven't been able to reproduce the third failure case with this patched Ford, though.

ixv:
I just tried this out, and got another error. This one appears to be deterministic, doing the same thing over and over gives the same set of stack traces. https://pastebin.com/Fn7cizXy

Not sure if this is a new error, or another manifestation of the same one. Should I make a new issue for this?

belisarius222:
@ixv I think that new crash is a different issue, but I don't fully understand it yet.  Ford is crashing on a Clay subscription update because it's no longer storing the duct it used to start the Clay subscription.  I think this is from a race condition.

My guess is that the :publish app had canceled the live build associated with this duct, causing Ford to delete its state related to the duct and emit a move to Clay to cancel the subscription, but that Clay, earlier in the same Arvo event, had already emitted subscription update moves to Ford before this happened.  So now Ford gets a subscription update on some unknown duct and crashes.

I think we should add some printfs to confirm this.  If this is in fact the cause of the crash, we should be able to fix it by just having Ford ignore spurious Clay updates instead of crashing on them.

belisarius222:
@eglaysher I think we need to rerun builds that are only provisionally in the tree of the root build associated with the newly updated Clay subscription.  Until we finish rerunning the whole tree, we don't know which sub-builds will turn out to be orphaned and which ones were actually needed.

This is because our rebuilds traverse up the tree from the leaves to the root, in hopes of minimizing the number of rebuilds that need to be run.  I suspect rebuilding downward from the root could yield a simpler system overall, but it would be a somewhat major change.

PR #2193, opened 2020-01-25
Ford +got-build helper

belisarius222:
Replaced manual calls to `(~(got by builds.state) build)` with a new `+got-build` helper function that prints a helpful error message on failure.

Depends on https://github.com/urbit/urbit/pull/2192


eglaysher:
(Also, sorry, I was confused by the inclusion of the other patch in this patch.)

PR #2198, opened 2020-01-28
Restore toplevel LICENSE.txt file.

eglaysher:
Urbit is MIT licensed, but the LICENSE.txt file appears to have
disappeared from the toplevel directory in c86879e833.


jtobin:
:+1: Good catch.

PR #2199, opened 2020-01-28
Ford: ignore spurious Clay responses

belisarius222:
Due to asynchronicity, Ford can receive responses from Clay to requests that it has already attempted to cancel.  This PR removes some overzealous assertions that this wouldn't happen, which should fix @ixv's crash from https://github.com/urbit/urbit/pull/2192#issuecomment-578887032.

This is ready for review, but please don't merge this just yet.  I need to cherry-pick this off master so it doesn't depend on the orphan dequeueing.


belisarius222:
Closing in favor of https://github.com/urbit/urbit/pull/2200

PR #2204, opened 2020-01-29
Autocomplete II: Electric Boogaloo

liam-fitzgerald:
Brings back rune and name autocomplete. 

Changes hover/complete to look for symbols in the current ford build, so imports and the like are accounted for. This is disabled on files in app/* as their payload is typically opaque.


philipcmonk:
Ideally you want the subject at the exact cursor position.  For example, before `=/  abc  57` hovering and naming of `abc` should give nothing, and after it (while it's still in the subject) it should be find it.

There's two parts of that, one is keeping track of how the subject changes in hoon.  Autocomplete in the dojo does this, it's not clear to me right now if you're using that mechanism.  The second part is making sure we start with the right subject.  This is `-:!>(..zuse)` plus anything included by the ford prelude.  This is true for *all* ford builds (app/lib/sur/gen/ted/mar).

One way to get the ford prelude would be to parse the file ford-wise, then replace the `hoon` at the end with `.` (which is `[%cnts [& 1] ~]`, or possibly more legibly `!,(*hoon .)`).  When you run that build, you should get exactly the prelude, then you just need to trace the subject through the hoon.

liam-fitzgerald:
> Ideally you want the subject at the exact cursor position. For example, before `=/ abc 57` hovering and naming of `abc` should give nothing, and after it (while it's still in the subject) it should be find it.
> 
> There's two parts of that, one is keeping track of how the subject changes in hoon. Autocomplete in the dojo does this, it's not clear to me right now if you're using that mechanism. The second part is making sure we start with the right subject. This is `-:!>(..zuse)` plus anything included by the ford prelude. This is true for _all_ ford builds (app/lib/sur/gen/ted/mar).
> 
> One way to get the ford prelude would be to parse the file ford-wise, then replace the `hoon` at the end with `.` (which is `[%cnts [& 1] ~]`, or possibly more legibly `!,(*hoon .)`). When you run that build, you should get exactly the prelude, then you just need to trace the subject through the hoon.

Can you elaborate on tracing the subject through the hoon?

ohAitch:
Might be interesting to see what comes out if you stick a `~?(trace-type [p.gen sut] ...)` around
https://github.com/urbit/urbit/blob/4915ceb96bdd56f5b1f25504d1c68f8d41161540/pkg/arvo/sys/hoon.hoon#L10944

(and the corresponding `trace-type` flag somewhere relevant, and set it in the language-server with `(%*(nest ut sut sut, trace-type &) gen)`)

philipcmonk:
If you look in `+he-tab` in `app/dojo.hoon`, you'll see that it does three things:

- insert a "magic marker" into the text with `+insert-magic:language-server-complete`
- get the autocomplete candidates with `+tab-list-hoon:language-server-complete`, which calls `+find-type` to find the type of the subject at the magic marker (assuming the whole thing has subject `sut` to start with).  Then it gets the identifiers available from the subject there, and looks for ones that match the prefix.
- get the autoadvance longest-prefix with `+advance-hoon:language-server-complete`, which calls `+find-type` and finds the longest match among the same set of identifiers as in `+tab-list-hoon` (there's some duplication of work here).

So `+find-type` is the part that traces through a `hoon` and gives you the subject at a given point.  The other part you need is to get the original subject right, which is what I mentioned where the ideal solution is to parse the file with `language-server-parser` and replace the `hoon` at the en (ie the body of the program) with `[%cnts [& 1] ~]` and send that to ford, so it'll produce exactly the subject that it would have started with.

Does that make sense?

PR #2195, opened 2020-01-25
Fix +to-wain:format null byte handling

pkova:
Fixes #1755

PR #2207, opened 2020-01-29
publish: fix new-comment name parsing for @da-case

### LINKED ISSUE #2085: Publish titles starting with numbers still give trouble
There was a similar issue before. Hitting something similar now.

Creating a post with a title "123" works fine, but then trying to post a comment to it result in a "syntax error {1 1}". Prefixing the title with an alphabetical character prevents this from occurring.

pkova:
Fixes #2085.

The root of the problem is that the frontend uses `stringToSymbol` (interface/publish/src/js/lib/util.js) to attempt converting any old post titles to `@tas` post id:s, inserting hyphens and such. If the title starts with a number, it just defaults to generating the current `@da`.

Why this fix works:

```
((su:dejs:format sym) [%s '~2020.1.29..16.45.41'])
{1 1}
syntax error
`@tas`(so:dejs:format [%s '~2020.1.29..16.45.41'])
%~2020.1.29..16.45.41
```

Everything works now, but given what [I've recently learned](https://github.com/urbit/urbit/pull/2195#issuecomment-578901192) I'm left wondering whether that's an actual `@tas`, though. 

```
`@tas`~2020.1.29..16.45.41
```

relinks the dojo. The dojo parser accepts typing the symbol out, though.

### DIFF
--- a/pkg/arvo/mar/publish/action.hoon
+++ b/pkg/arvo/mar/publish/action.hoon
@@ -69,7 +69,7 @@
       %-  ot:dejs
       :~  who+(su:dejs fed:ag)
           coll+so:dejs
-          name+(su:dejs sym)
+          name+so:dejs
           content+so:dejs
       ==
     ::


ixv:
This looks good, thanks.

joemfb:
Some context:

`(su sym)` fails because `'~2020.1.29..16.45.41'` is not a valid symbol (`@tas`), while `+so` succeeds because it is a string.

These are the same thing:

```
> `@tas`'~2020.1.29..16.45.41'
%~2020.1.29..16.45.41
> ^-(@tas ^-(@ '~2020.1.29..16.45.41'))
%~2020.1.29..16.45.41
```

Any atom can be cast to any aura if it's first cast to `@`, and the irregular cast syntax automatically does so (arguably a misfeature!).

This unlinks dojo because it crashes the pretty-printer, attempting to print the date as if it were LSB utf-8 (note that it's casting the `@da` itself, not the `@t`-serialized `@da`):

```
> `@tas`~2020.1.29..16.45.41
!!
```

Note the validation results, and compare the bytes:

```

> ((sane %t) '~2020.1.29..16.45.41')
%.y
> ((sane %ta) '~2020.1.29..16.45.41')
%.y
> ((sane %tas) '~2020.1.29..16.45.41')
%.n
> 
> ((sane %t) ~2020.1.29..16.45.41)
%.n
>
> `@ux`~2020.1.29..16.45.41
0x8000.000d.2ccf.c435.0000.0000.0000.0000
> `@ux`'~2020.1.29..16.45.41'
0x3134.2e35.342e.3631.2e2e.3932.2e31.2e30.3230.327e
```

pkg/arvo/mar/publish/action.hoon
```diff
@@ -69,7 +69,7 @@
       %-  ot:dejs
       :~  who+(su:dejs fed:ag)
           coll+so:dejs
-          name+(su:dejs sym)
```
Fang-: If I understand correctly, this is `post` in the `action:publish` type. Not sure why we use `name` here.

Regardless, publish [uses that as a `@tas`](https://github.com/urbit/urbit/blob/master/pkg/arvo/app/publish.hoon#L1139), so accepting technically-not-valid-term strings into it might be ill-advised. (See also @joemfb's comment.)

That said, @ixv has already given his blessing, and publish will relaunch with all new code in the near future anyway, so this is probably fine.

pkova: Note that posts are already saved to clay with `@da`-style id:s when their titles start with numbers, which is why in the mark for new-post 9 lines above `name` is parsed with `so`. 

I do agree that it's incorrect, but invalid `@tas` post id:s are already accepted regardless of this PR being merged or not. You just can't comment on them.

Of course none of this matters very much with what I hear of the publish rewrite, I'm fine with you guys closing this or merging this as you wish!

PR #2208, opened 2020-01-29
OS1: adds publish new post page

ixv:


pkg/interface/publish/src/js/components/root.js
```diff
@@ -78,20 +78,39 @@ export class Root extends Component {
...
-            sidebarShown={true}
-            notebooks={state.notebooks}>
-              <Notebook
-              notebooks={state.notebooks}
-              view={view}
-              ship={ship}
-              book={notebook}/>
-            </Skeleton>
-          )
+          if (view === "new") {
```
matildepark: Ooh, I like this a lot — but I don't think there's a mount for this param on the back-end, so while navigating to the "new" view will work fine, going directly to it won't work. Should be added I suppose.

matildepark: Just tested and it does work, so nvm! Cool cool cool. Let me look at testing...

matildepark:
:D 

Jan-29-2020 15-07-54

It needs to have [mobile-specific nav](https://github.com/urbit/urbit/blob/master/pkg/interface/chat/src/js/components/chat.js#L227) but I can do that in the polish round.

PR #2209, opened 2020-01-29
link: listening & comments

Fang-:
Broadly, the commits herein address two issues.

- Comments weren't being synced by link-listen-hook at all.  
2e8fb88 makes that so, but depends on 69c86d7, which is an ugly hack that would not have been needed had the link-store API rewrite from the `m/uplink-broad` branch been in use. That work requires a partial rewrite of the frontend logic, however, so that will have to wait for a bit longer. The hack is "good enough for now" and I fully intend to see its removal through.

- Link-listen-hook would not retry if rejected due to out-of-date group definition.  
8723dc6 implements retry logic with exponential backoff, so that we attempt to recover from rejected subscriptions.

Commits not directly addressed above are largely administrative in nature. This includes a couple cherry-picked commits from master relating to `/lib/dbug`, which proved useful for debugging here.

(CI is failing because the branch this is based on fails CI. There are no errors relating to this PR in the logs.)

### DIFF
--- a/pkg/arvo/app/link-listen-hook.hoon
+++ b/pkg/arvo/app/link-listen-hook.hoon
@@ -19,6 +19,21 @@
       ::      makes adding any state in the future easier.
   ==
 ::
++$  target
+  $:  what=%local-pages  ::TODO  %annotations
+      who=ship
+      where=path
+  ==
+++  wire-to-target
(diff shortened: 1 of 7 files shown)


vvisigoth:
@Fang- merge away!

PR #2213, opened 2020-01-30
chat-js: added contacts metadata display to chat

loganallenc:
If there is a group associated with a particular chat, show member custom sigil colors and nicknames from contacts data.


vvisigoth:
@loganallenc let's merge this.

PR #2217, opened 2020-01-31
Make release tarballs extract to a release directory

### LINKED ISSUE #1400: Urbit distribution tarball poorly packaged
The urbit distribution tarball  https://bootstrap.urbit.org/urbit-linux64-v0.8.0.tgz
has all of the files in the root of the tarball instead of inside a directory. This is a sin.

jtobin:
Fixes #1400.

Release tarballs previously didn't include a release directory, so the files would be unceremoniously dumped into the user's current directory when extracted.  No longer!

Before:

```
$ tar xzf urbit-linux64-v0.10.3.tgz
$ ls
urbit    urbit-linux64-v0.10.3.tgz    urbit-terminfo    urbit-worker
```

After:

```
$ tar xzf urbit-linux64-v0.10.3.tgz
$ ls
urbit-linux64-v0.10.3    urbit-linux64-v0.10.3.tgz
$ ls urbit-linux64-v0.10.3
urbit    urbit-terminfo    urbit-worker
```

### DIFF
--- a/sh/release
+++ b/sh/release
@@ -12,12 +12,22 @@ else
   ver="$(git rev-parse HEAD)"
 fi
 
+traced () {
+  echo '$' "$@" >&2; "$@"
+}
+
 mkdir -p release
 
 for plat in linux64 darwin
 do
   sh/cross urbit "$plat"
 
+  tmp=$(mktemp -d)
+  mkdir -p $tmp/urbit-$plat-$ver
+  traced cp -r cross/$plat/* $tmp/urbit-$plat-$ver
+
   echo "packaging release/urbit-$plat-$ver.tgz"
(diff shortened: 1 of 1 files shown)


jtobin:
There's a hiccup here when I test it with the actual `urbit-terminfo` directory; I get a 'too many levels of symbolic links' error from `cp`, which is a first.  Will resolve..

jtobin:
I can't reproduce the `Too many levels of symbolic links` error outside of my own machine, and importantly it doesn't appear to occur in CI.  So I'm not terribly concerned about it.

jtobin:
Pushing this through; I've confirmed that everything works by running the packaged release binaries on mainnet.

PR #2222, opened 2020-02-02
zuse: add crypto test vectors

lukechampine:
Hi! I've been assigned the [crypto tests bounty](https://grants.urbit.org/bounties/1097913361-unit-tests-for-zuse-cryptographic-functions), so this is my first step towards completing that.

Right now I just have tests for AES ECB mode. I wasn't sure if I should open separate PRs for each algorithm, or if I should submit everything together in one big PR. I also don't know how many tests cases are appropriate for each function; for AES ECB, the FIPS-197 specification only includes a handful of cases, but I could easily generate more from e.g. the Go stdlib. ([wycheproof](https://github.com/google/wycheproof) seems like a good source too, but they don't have vectors for every algorithm.)

Anyway, since I'm fresh out of Hoon 101, I figured it would be best to submit some code as early as possible so that I can get feedback on style/testing conventions going forward. Let me know how I can improve! :)

Checklist of arms to test:

- [x] `aes`
    - [x] ECB (`ecba`, `ecbb`, `ecbc`)
    - [x] CBC (`cbca`, `cbcb`, `cbcc`)
    - [x] CTR (`ctra`, `ctrb`, `ctrc`)
    - [x] SIV (`siva`, `sivb`, `sivc`)
    - [x] CMAC (`maca`, `macb`, `macc`)
- [x] `ed` (ed25519)
    - [x] `puck`
    - [x] `sign`
    - [x] `veri`
- [x] `scr` (scrypt)
    - (vectors added, but currently infeasible to run)
- [x] `hmac`
    - [x] `hmac-sha1`, `hmac-sha256`, `hmac-sha512`
- [x] `secp256k1`
    - (tested indirectly via `bip32`)
- [x] argon2 (already done)
- [x] blake2b (already done)
- [x] keccak (already done)
- [x] pbkdf (already done)
- [x] ripemd (already done)


pkg/arvo/tests/sys/zuse/crypto/aes.hoon
```diff
@@ -0,0 +1,109 @@
...
+  =,  i.ves
+  ;:  weld
+    ::  encrypt
+    %+  expect-eq
+        !>  out
+        !>  `@ux`(~(en ecba key) in)
+    ::  decrypt
+    %+  expect-eq
+        !>  in
+        !>  `@ux`(~(de ecba key) out)
```
lukechampine: This feels like it could be made much DRYer -- the `do-test-vectors-ecbX` functions are identical except for whether they use `ecba`, `ecbb`, or `ecbc`. Is there a way to accept the algorithm as an argument?

Fang-: Good question! Perhaps this?

```hoon
+$  vector  [key=@ux in=@ux out=@ux]
::
++  run-vectors-on
  |*  [ecbx=_ecba vectors=(list vector)]
  ^-  tang
  ::  rest of the logic. maybe use +turn in place of manual looping?
  ::  example encryption call:
  (~(en ecbx key) in)
::
++  test-vectors-ecba
  %+  run-vectors-on  ecba
  :~  some
      vectors
  ==
```

This depends on the fact that all the `+ecbx:aes:crypto` doors have the same shape/arms (as `+ecba`), but I think that's a sane & safe assumption.

pkg/arvo/tests/sys/zuse/crypto/aes.hoon
```diff
@@ -0,0 +1,109 @@
...
+  :~
+    :+  0x0
+      0x0
+    0xdc95.c078.a240.8989.ad48.a214.9284.2087
+  ::
+    :+  0x0
+      0x1
+    0x530f.8afb.c745.36b9.a963.b4f1.c4cb.738b
+  ::
+    :+  0x1.0203.0405.0607.0809.0a0b.0c0d.0e0f.1011.1213.1415.1617.1819.1a1b.1c1d.1e1f
```
Fang-: Try to avoid going wider than 80 characters. You can multi-line `@ux` literals like so:

```hoon
:+   0x1.0203.0405.0607.0809.0a0b.0c0d.0e0f.
    1011.1213.1415.1617.1819.1a1b.1c1d.1e1f
  etc
etc
```

pkg/arvo/tests/sys/zuse/crypto/aes.hoon
lukechampine: ah, that's clever! Tested locally and it seems to work

lukechampine:
>Separate commits in the same PR seem like good style.

SGTM -- wouldn't want such a nice PR number to go to waste. I can add a checklist to the PR body so that we can track progress (assuming @jtobin agrees).

pkg/arvo/tests/sys/zuse/crypto/aes.hoon
lukechampine: Done -- will standardize on 64 hex digits per line

pkg/arvo/tests/sys/zuse/crypto/aes.hoon
lukechampine: Can you elaborate on how to use `turn` here? I can sort of picture how it would work (map the encrypt and decrypt checks over each vector), but I can't get it to work. Here's what I have:
```hoon
++  do-test-vectors
  |*  [ecbx=_ecba ves=(list vector)]
  ^-  tang
  =<  (weld (turn ves enc) (turn ves dec))
  |%
  ++  enc
    |=  v=vector
    %+  expect-eq
      !>  out.v
      !>  `@ux`(~(en ecbx key.v) in.v)
  ++  dec
    |=  v=vector
    %+  expect-eq
      !>  in.v
      !>  `@ux`(~(de ecbx key.v) out.v)
  --
```

Fang-: That seems pretty close. The only thing you're really missing is a `+zing` to turn the resulting `(list tang)` into a `tang`. (`tang` is a `(list tank)`, and `+zing` turns a list of lists into a single list.)

With `|^` as shorthand for `=<  ...  |%`, and using `+category` from `/lib/test` for fancier failure printing, I'd do something like this:

```hoon
++  do-test-vectors
  |*  [ecbx=_ecba ves=(list vector)]
  ^-  tang
  |^  %+  weld
        %+  category  "encrypting"
        (zing (turn ves enc))
      %+  category  "decrypting"
      (zing (turn ves dec)))
  ::
  ++  enc
    |=  vector
    %+  expect-eq
      !>  out
      !>  `@ux`(~(en ecbx key) in)
  ::
  ++  dec
    |=  vector
    %+  expect-eq
      !>  in
      !>  `@ux`(~(de ecbx key) out)
  --
```

(If that nest-fails around the `zing`s, it may need a cast to the expected type.)

jtobin:
On Sun, Feb 02, 2020 at 11:47:35AM -0800, Fang wrote:
> Separate commits in the same PR seem like good style. Maybe prefix
> with `test:` instead of `zuse:`. (@jtobin do you agree?)

Yep, I would avoid prefixing with 'zuse:' unless one is actually
touching %zuse proper.

pkg/arvo/tests/sys/zuse/crypto/aes.hoon
```diff
@@ -0,0 +1,413 @@
...
+    :+
+      0x8563.2d07.c6e8.f37f.950a.cd32.0a2e.cc93
+      14
+      0x40c0.2b96.90c4.dc04.daef.7f6a.fe5c
+    ::
+    :^    0x7f7e.7d7c.7b7a.7978.7776.7574.7372.7170.
+            4041.4243.4445.4647.4849.4a4b.4c4d.4e4f
+        :~  0x11.2233.4455.6677.8899.aabb.ccdd.eeff.
+            dead.dada.dead.dada.ffee.ddcc.bbaa.9988.
+            7766.5544.3322.1100
```
lukechampine: this test case is currently failing, because this AD has leading zeros. What's the easiest way to prepend leading zeros to a byte string? I noticed that some functions take the length as a separate argument (for example, manually hashing this AD with "maca `40" produces the correct output), but there's no way to supply the length of the AD in the `siva` API.

Also, if I happen to discover any test inputs that cause crashes, should I treat those as security vulnerabilities and report them through some secure channel, or is it fine to post them here?

joemfb: You could skip it, or factor it into a call to the SIV internals where you can pass the length.

In general, you never need to handle leading zeros in urbit, except when integrating with 3rd-party systems. The siv{a|b|c} gates are currently for urbit cryptography, and don't have a 3rd-party integration use-case.

lukechampine:
I've added a checklist, but I'm not sure exactly which arms need to be tested. Some are clearly part of the "public interface" (like `maca:aes:crypto`), while others seem to be merely internal helpers (like `doub:aes:crypto`). Can someone fill in the sections marked with a `?` with the arms I should test (or at least, test first)?

pkg/arvo/sys/zuse.hoon
```diff
@@ -3357,7 +3357,7 @@
         ~/  %en
         |=  txt/@
         ^-  (trel @uxH @ud @ux)
-        =+  [k1=(rsh 5 3 key) k2=(end 5 3 key)]
+        =+  [k1=(rsh 5 6 key) k2=(end 5 6 key)]
```
lukechampine: I found a bug!

Do I need to compile a pill for this? Not sure how that works

pkg/arvo/sys/zuse.hoon
```diff
@@ -3357,7 +3357,7 @@
         ~/  %en
         |=  txt/@
         ^-  (trel @uxH @ud @ux)
-        =+  [k1=(rsh 5 3 key) k2=(end 5 3 key)]
+        =+  [k1=(rsh 6 3 key) k2=(end 6 3 key)]
```
lukechampine: I found a bug! `sivb` was previously unusable; it would split the key in the wrong place and then crash because the key was the wrong size.

Do I need to compile a pill for this? Not sure how that works

joemfb: Good catch! `+crub` uses `+siva` and `+sivc`, but nothing uses `+sivb`. I'd guess it's always been broken.

You'll need to create and commit a new solid pill at `bin/solid.pill`. You can use `sh/update-solid-pill`.

lukechampine: Cool, seems like it worked!

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
```diff
@@ -0,0 +1,156 @@
...
+    |=  vector
+    %+  expect-eq
+      !>  sig
+      !>  `@ux`(sign msg seed)
+  ::
+  ++  check-veri
+    |=  vector
+    %+  expect-eq
+      !>  %.y
+      !>  (veri sig msg pk)
```
lukechampine: Is there a shorthand for `expect-eq %.y`? If not, might be convenient to have a bare `expect` arm.

joemfb: Definitely! Feel free to add it to `lib/test`.

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
```diff
@@ -0,0 +1,156 @@
...
+  ::
+  ++  vectors
+    (turn vectors-raw swp-vec)
+  ++  swp-vec
+    |=  vector
+    :^  (swp 3 seed)
+        (swp 3 pk)
+        (swp 3 msg)
+        (swp 3 sig)
```
lukechampine: this feels ugly; how can I clean this up? I want to map `(cury swp 3)` over each element of each vector.

btw, it took me a while to realize that my tests were failing because I wasn't correcting for endianness...I guess it's more convenient to operate on curves this way, but it definitely tripped me up. *Especially* because `shan` and `shaz` have different endiannness...es. I thought I was going crazy when the `shan` output was clearly correct and the `shaz` output was not. 😅

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
```diff
@@ -0,0 +1,156 @@
...
+       41e9.89f7.914f.90cc.2c7f.ff35.7876.e506.
+       b50d.334b.a77c.225b.c307.ba53.7152.f3f1.
+       610e.4eaf.e595.f6d9.d90d.11fa.a933.a15e.
+       f136.9546.868a.7f3a.45a9.6768.d40f.d9d0.
+       3412.c091.c631.5cf4.fde7.cb68.6069.3738.
+       0db2.eaaa.707b.4c41.85c3.2edd.cdd3.0670.
+       5e4d.c1ff.c872.eeee.475a.64df.ac86.aba4.
+       1c06.1898.3f87.41c5.ef68.d3a1.01e8.a3b8.
+       cac6.0c90.5c15.fc91.0840.b94c.00a0.b9d0
+      0xaab.4c90.0501.b3e2.4d7c.df46.6332.6a3a.
```
lukechampine: style question: these columns are unintentionally aligned, because the bitstring starts with three leading zeros. Should I indent differently in that event?

Fang-: You generally use line separators to help distinguish between longer blocks in cases where it's potentially confusing. Seems appropriate here.

```hoon
:^    short
    also-short
  a-little-bit-
  longer
::
  weird-
indentation
```

lukechampine:
So, I added test vectors for scrypt, and they appear to be correct for small parameters, but anything larger takes far too long. The Go `scrypt` package can check all of the vectors in under 10 seconds, but the hoon `scrypt` takes many minutes to check even one vector.

Is this because the `scrypt` arms aren't jetted? And what should I do with the test file I wrote?

pkg/arvo/lib/bip32.hoon
```diff
@@ -212,5 +212,5 @@
 ::
 ++  hash160
   |=  d=@
-  (ripemd-160:ripemd:crypto 256 (sha-256:sha d))
+  (ripemd-160:ripemd:crypto 32 (sha-256:sha d))
```
lukechampine: found another bug!

I don't need to build a new pill for this, because it's not in `sys`, right?

jtobin: Yup, that's right.

lukechampine:
Ok, I've rebased onto master and built a new pill. So far, I've added test vectors for AES, HMAC, Ed25519, scrypt, and BIP32, and I've found and fixed 2 bugs. What's next? Are there additional arms I should test?

joemfb:
@lukechampine, I still need to make time for a full review, but your checklist looks reasonable to me.

As for scrypt, there are jets in the source tree and built, but they're not included in the jet dashboard. A little history sleuthing shows them being added in b812d16b0aff755289ff7399e7ecbf94fc4eb8cd, and included all the way up to 9c22f7e08b33290b6cfb3cd7bc5880bd780a75c5. I don't know if they were intentionally removed due to some issue, or simply fell by the wayside during a major refactor. It'd be great to have them re-enabled and tested, but that could be done separate from this PR and bounty.

joemfb:
It looks like the CI failure here is just an issue with the build cache.

/cc @benjamin-tlon, @jtobin

pkg/arvo/tests/sys/zuse/crypto/scrypt.hoon
```diff
@@ -0,0 +1,70 @@
...
+    |=  vector
+    %+  expect-eq
+      !>  (swp 3 out)
+      !>  `@ux`(hsh pw salt n r p 64)
+  ::
+  ++  vectors
+    ::  TODO: until scrypt has been jetted, we can only test the
+    ::  first vector; the others do not finish in a reasonable
+    ::  amount of time.
+    %+  scag  1  ^-  (list vector)
```
lukechampine: should delete this when the `scrypt` jet is re-enabled

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
joemfb: @lukechampine, it is kind of ugly, but there's no general, typed way to map over arbitrary-width tuples (You could do it over raw nouns, although i'd advice against it.)

Little-endian is the right default for hoon atom operations, including crypto IMO, but the inconsistencies you mentioned are really bad. It'd be nice to get them cleaned up.

pkg/arvo/tests/lib/bip32.hoon
```diff
@@ -0,0 +1,179 @@
...
+      ==
+  ::
+  ++  check-id
+    |=  vector
+    %+  expect-eq
+      !>  id
+      !>  ^-  @ux
+         =,  (from-seed mk)
+         =,  (derive-path dp)
+         identity
```
joemfb: It's actually pretty cool that this works, but, IMO, it's an abuse of the `=,` namespace-ing system. It'd be better to be explicit about the stateful threading, something like:

```
=/  bip  (from-seed mk)
=.  bip  (derive-path.bip dp)
identity.bip
```

Or inline:

```
identity:(derive-path:(from-seed mk) dp)
```

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
```diff
@@ -0,0 +1,156 @@
...
+      !>  sig
+      !>  `@ux`(sign msg seed)
+  ::
+  ++  check-veri
+    |=  vector
+    %-  expect
+      !>  (veri sig msg pk)
+  ::
+  ++  vectors
+    (turn vectors-raw swp-vec)
```
joemfb: Not strictly necessary, but a neat trick: you can turn this into a constant via compile-time evaluation with `^~`:

Compare:

```
> =<  -  |%  ++  foo  (turn bar |=(@ +(+<)))  ++  bar  (gulf 0 4)  --
[[8 [9 3.075.070 0 1.023] 9 2 10 [6 [7 [0 3] 1 0] 7 [0 3] 1 4] 0 2] 8 [9 96.126 0 1.023] 9 2 10 [6 [7 [0 3] 9 4 0 1] 7 [0 3] 8 [1 0] [1 4 0 6] 0 1] 0 2]
```

with

```
> =<  -  |%  ++  foo  ^~  (turn bar |=(@ +(+<)))  ++  bar  ^~  (gulf 0 4)  --
[[1 0 1 2 3 4 0] 1 1 2 3 4 5 0] 
```

pkg/arvo/tests/lib/bip32.hoon
lukechampine: even better, I think, would be to factor out the derivation from the subtests. But I wasn't sure how to accomplish that.

pkg/arvo/tests/lib/bip32.hoon
```diff
@@ -0,0 +1,179 @@
...
+    |=  vector
+    %+  expect-eq
+      !>  cc
+      !>  ^-  @ux
+         =,  (from-seed mk)
+         =,  (derive-path dp)
+         chain-code
+  ::
+  ++  vectors
+    :~
```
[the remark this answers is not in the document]
lukechampine: Should I do that for all of the tests I added?

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
lukechampine: neat! I think it's appropriate here.

lukechampine:
Comments addressed. Too bad the pill keeps getting outdated. :/ I'll wait until final approval before updating it again.

jtobin:
On Fri, Feb 28, 2020 at 08:53:05AM -0800, Luke Champine wrote:
> Comments addressed. Too bad the pill keeps getting outdated. :/ I'll
> wait until final approval before updating it again.

Don't worry about racing the pill.  That's a battle you can't win.  I
think Aesop might have written something about this..

(But seriously, fear not, I'll handle the inevitable pill conflict on merge.)

pkg/arvo/tests/sys/zuse/crypto/ed25519.hoon
ohAitch: Personally I'd just collapse to one line and call it a day, it's not _that_ much code, but:

ye of little faith, check out ++scanf and ++raid in zuse. If what you have is a fixed-length tuple of the shape `|-($@(@ [@ $]))`, then a wet gate something like `|*(a=|-($@(@ [@ $])) ^+(a ?@(a (swp 3 a) [(swp 3 -.a) (..$ +.a)])))` will monomorphize itself repeatedly for the different tuple lengths, producing a tuple of the same width at each step. (in this case same exact type even, hence `^+ a`.) 

To reiterate, for just four elements the explicit version is much more readable

```
[(swp 3 seed) (swp 3 pk) (swp 3 msg) (swp 3 seed)]
```

or if you must

```
=/  f  |=(a=@ (swp 3 a))
[(f seed) (f pk) (f msg) (f seed)]
```

joemfb: Very nice! Looks like `(..$ +.a)` is what makes this go, forcing the re-monomorphization.

It requires a little casting to make it work (and a `$;` for the manual mold construction):

```
> =swp-tup |*(a=$;(|-($@(@ [@ $]))) ^+(a ?@(a (swp 3 a) [(swp 3 -.a) (..$ +.a)])))
> (swp-tup `@`%foo)
6.713.199
> (swp-tup `@`%foo `@`%bar)
[6.713.199 6.447.474]
> (swp-tup `@`%foo `@`%bar `@`%baz)
[6.713.199 6.447.474 6.447.482]
```

ohAitch: `(swp 3 %foo)` does not have the value `%foo` and so cannot be cast back to its constant type, indeed :P

ohAitch: I'd recommend e.g. `@ta` to verify the output clearer; and prefix `? ` since the type being right is the important part

joemfb:
```
> =swp-tup |*(a=$;(|-($@(@ [@ $]))) ^+(a ?@(a (swp 3 a) [(swp 3 -.a) (..$ +.a)])))
> ? (swp-tup ~.foobarbaz)
  @ta
~.zabraboof
> ? (swp-tup ~.foo ~.bar ~.baz)
  {@ta @ta @ta}
[~.oof ~.rab ~.zab]
```

PR #2219, opened 2020-01-31
Publish: subscribe screen implementation

vvisigoth:
Keeping it dumb.

Hasn't been tested E2E, but all actions are submitted correctly. If there are errors, it'll be on the back end.

Screen Shot 2020-01-31 at 3 25 27 PM

PR #2223, opened 2020-02-03
basic join screen

vvisigoth:
LIke https://github.com/urbit/urbit/pull/2219 but without the cruft.

### DIFF
--- a/pkg/interface/publish/src/js/components/lib/join.js
+++ b/pkg/interface/publish/src/js/components/lib/join.js
@@ -1,13 +1,125 @@
 import React, { Component } from 'react'
+import classnames from 'classnames';
+import { Route, Link } from 'react-router-dom';
+import urbitOb from 'urbit-ob';
 
 //TODO textarea + join button to make an api call
 export class JoinScreen extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      book: '/',
+      error: false
+    };
+
+    this.bookChange = this.bookChange.bind(this);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+      const { props, state } = this;
+      if (state.book in props.notebooks) {
+        props.history.push('/~publish');
+      }
+   }
+
+  onClickJoin() {
+    const { props, state } = this;
+
+    let text = state.book;
+    // an error condition to prevent double joins?
+    if (text in props.notebooks ||
+        text.length === 0) {
+      this.setState({
+        error: true,
+      });
+      return;
+    }
+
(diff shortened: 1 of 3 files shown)


pkg/interface/publish/src/js/components/lib/join.js
```diff
@@ -1,13 +1,125 @@
...
+      book: '/',
+      error: false
+    };
+
+    this.bookChange = this.bookChange.bind(this);
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+      const { props, state } = this;
+      if (state.book in props.notebooks) {
```
matildepark: props.notebooks looks like this:

```
{
  "notebooks": {
    "~zod": {
      "test": {
        "title": "test",
        "about": "test",
        "num-unread": 0,
        "num-notes": 1,
        "date-created": 1580765730419
      }
    }
  }
}
```

Because `state.book` is checking entire notebook titles against the top level properties of the object, not the notebooks nested in those properties, it's checking against notebook owners.

So, basically what this line does, given that notebooks example above, is that if I type `~zod` it boots me back to the home screen.

pkg/interface/publish/src/js/components/lib/join.js
```diff
@@ -1,13 +1,125 @@
...
+        props.history.push('/~publish');
+      }
+   }
+
+  onClickJoin() {
+    const { props, state } = this;
+
+    let text = state.book;
+    // an error condition to prevent double joins?
+    if (text in props.notebooks ||
```
matildepark: Likewise, this error check will be set to `true` if I type `~zod`, or any other notebook owner I'm subscribed to.

vvisigoth:
Whoops, good catch.
—
~poldec-tonteg
http://urbit.org

pkg/interface/publish/src/js/components/lib/join.js
```diff
@@ -1,13 +1,134 @@
...
+          <p className="f9 gray2 mb4">Notebook names use lowercase, hyphens, and slashes.</p>
+          <textarea
+            ref={ e => { this.textarea = e; } }
+            className="f7 mono ba b--gray3 b--gray2-d bg-gray0-d white-d pa3 mb2 db"
+            placeholder="~zod/dream-journal"
+            spellCheck="false"
+            rows={1}
+            onKeyPress={e => {
+              if (e.key === "Enter") {
+                this.onClickJoin();
```
matildepark: You'll want `e.preventDefault()` to stop it from entering a new line in the textarea here.

PR #2227, opened 2020-02-04
publish group paths

ixv:
Changed the publish api so that group paths are passed through from the frontend rather than generated on the backend (except in cases where they must be generated on the backend).

also included a minor bugfix in this PR.


vvisigoth:
This looks good. Question, though, under what circumstances does the path need to be mad on the backend?

ixv:
When you import a notebook with clay and the paths are nulled, or you have an old version of the publish-info mark. It defaults to /~/publish/<ship>/<notebook>/write or /read

vvisigoth:
ah got it.

vvisigoth:
Go ahead and merge.

PR #2230, opened 2020-02-05
Validate group names

vvisigoth:
Check if group names conform to @tas


joemfb:
This doesn't appear to capture the "leading-alpha" constraint of `@tas`. Or ensure that the whole string is match. This should match all valid `@tas` strings:  `/^[a-z][a-z0-9-]*$/`

matildepark:
Oh! God, that's a good catch — thanks. I forgot about the leading alpha. We should port that to Chat's validation check too, @vvisigoth.

vvisigoth:
Oh nice. I'll update this.
—
~poldec-tonteg
http://urbit.org

vvisigoth:
Looks like I was mistaken and this needs to just be a `@ta`

joemfb:
Even easier: `/^[a-z0-9~_.-]*$/`

vvisigoth:
boom
—
~poldec-tonteg
http://urbit.org

vvisigoth:
@matildepark plz look

matildepark:
It’s fine offhand but I think it should be everywhere a group can potentially be created. Chat, Publish.

joemfb:
While I'm nosing around, the `new Regexp()` is superfluous; it's for runtime compilation, but this expression is static. The literal works just fine: `/foo/.test('foo')`.

loganallenc:
@vvisigoth please reflect these changes onto publish and chat, otherwise looks good. thanks!

vvisigoth:
OK I updated chat to use this but the new publish stuff isn't in yet, so I can't update.

PR #2233, opened 2020-02-05
link: subscriptions for the frontend

Fang-:
In which we leave web 1.0 behind.

There's some known issues/things I'm still working on, but figured I should open a PR to give @matildepark a chance to start reviewing this. I'm fine with this being merged as-is, too, it seems pretty functional.

- [ ] loading directly into a submission doesn't work
- [ ] comment counts don't always update
- [ ] unused index logic floating around (wrt comments)
- [ ] should probably move towards a more consistent/persistent route path scheme. "nth link on mth page" works, but can change out from under you.


matildepark:
> should probably move towards a more consistent/persistent route path scheme. "nth link on mth page" works, but can change out from under you.

Yes, it absolutely should. That routing path sucks, but I took it up because there wasn't really a way to just grab "arbitrary link [in base64] within this path" with the API. Ideally we just change the route to reflect that, if that's now available...

pkg/interface/link/src/js/api.js
```diff
@@ -91,135 +91,84 @@ class UrbitApi {
...
-        "link-update": {
-          commentPage: {
-            path: path,
-            linkPage: page,
-            index: index,
-            comPageNo: commentPage,
-            data: responseData.page
+  getCommentsPage(path, url, page) {
+    //TODO factor out
+    // encode the url into @ta-safe format, using logic from +wood
```
matildepark: I know it's here as a TODO, but yeah this would be good as a utility function in `/lib/util.js`

Fang-:
> will pull this and give some basic tests after standup.

Feel free to merge if it looks good after your tests!

pkg/arvo/app/link-view.hoon
```diff
@@ -0,0 +1,303 @@
...
+::
+%+  verb  |
+^-  agent:gall
+=<
+  |_  =bowl:gall
+  +*  this  .
+      do    ~(. +> bowl)
+      def   ~(. (default-agent this %|) bowl)
+  ::
+  ++  on-init
```
matildepark: This doesn't [launch the tile](https://github.com/urbit/urbit/blob/os1-rc/pkg/arvo/app/link-server-hook.hoon#L98).

pkg/interface/link/src/js/components/link.js
```diff
@@ -18,33 +18,35 @@ export class LinkDetail extends Component {
...
-      if (!comments) {
-        api.getComments(props.path, props.data.url, props.page, props.link);
+    // if we have no preloaded data, and we aren't expecting it, get it
+    if (this.props.page != 0 && (!this.props.data || !this.props.data.url)) {
+      api.getPage(this.props.path, this.props.page);
+    } else {
+
+      // if we have preloaded our data,
+      // but no comments, grab the comments
+      if (!!this.props.comments) {
```
matildepark: Going directly to a submission page fails because of this: it can't get length of undefined on api.getCommentsPage, so it's passing an undefined URL — so, we want to wait for the URL. Might just need the same check at L48 here.

matildepark: Yeah, this is one of those "on a plane" clunky things I think — Links in general needs a cleanup and refactor it didn't get after "minimum viable working thing". :/ We'll see what we can do.

vvisigoth:
morge

matildepark:
I'm gonna build it first so we can run it in os1-rc.

PR #2240, opened 2020-02-06
link fe: various fixes and improvements

Fang-:
Mainly:

- Include submission URL (base64-encoded) in the route path, so that pageloads on a specific submission can actually work. (Note that we don't take out the existing `:index` part of the path, it's still useful for the regular flow.)
- Keeps displayed comment counts up-to-date.
- Fix issues around the combination of pagination and receiving updates.
- Misc refactoring/cleanup.

Please see commit messages for more granular and detailed changelog.


Fang-:
(I, uh, did `gulp bundle-prod` because I remember being admonished for not doing so prior to PRing. From the looks of it I guess we don't actually want to do that for non-release branches? (something something binary files discussion, you know the drill))

matildepark:
Cursed name.

> (I, uh, did gulp bundle-prod because I remember being admonished for not doing so sometime previously. From the looks of it I guess we don't actually want to do that for non-release branches? (something something binary files discussion, you know the drill))

This is sort of an unspoken thing at the moment — on the one hand, I think it makes it harder to review code if the line diffs for the compiled, but unminified JS are also replicated; on the other, if you compile and minify it, it produces merge conflicts if several PRs are on the same application (as with this and #2238), and of course it produces these massive diff numbers if the previous binary was unminified.

matildepark:
Hmm, so this is odd — `contact-update`s still come in, but they're not being handled, put into state. If you refresh, they do appear in the `contact-initial` of course. But I remember them updating automatically, and I don't know when this regression happened.

matildepark:
Honestly, I don't know — there's not even a contact-update reducer here, so am I remembering incorrectly? Did contact updates change recently? 

I think what needs doing is just to add [this reducer](https://github.com/urbit/urbit/blob/os1-rc/pkg/interface/contacts/src/js/reducers/contact-update.js) to the stack of reducers. Would you mind adding that here?

pkg/interface/link/src/js/components/root.js
```diff
@@ -94,7 +99,7 @@ export class Root extends Component {
               )
             }}
           />
-          <Route exact path="/~link/(popout)?/:ship/:channel/:page/:index/(comments)?/:commentpage?"
+          <Route exact path="/~link/(popout)?/:ship/:channel/:page/:index/:encodedUrl/(comments)?/:commentpage?"
```
matildepark: So, this looks like this:

`~link/~/default/0/0/aHR0cDovLzRjaGFuLm9yZw==`

The `/0/0` there presents the same issue as the previous route — that is, that the permalink is a bit ephemeral. Maybe just strip the `:index` here.

pkg/interface/link/src/js/components/lib/link-item.js
```diff
@@ -68,7 +70,7 @@ export class LinkItem extends Component {
             : "~" + props.ship}</span>
           <span className="f9 inter gray2 pr3 v-mid">{this.state.timeSinceLinkPost}</span>
           <Link to=
-          {"/~link" + props.popout + "/" + props.channel + "/" + props.page + "/" + props.index}
+          {"/~link" + props.popout + "/" + props.channel + "/" + props.page + "/" + props.index + "/" + encodedUrl}
```
matildepark: You'd remove it from here too.

pkg/interface/link/src/js/components/links-list.js
```diff
@@ -67,12 +74,12 @@ export class Links extends Component {
 
       return (
         <LinkItem
-        key={timestamp}
+        key={time}
         title={title}
         page={props.page}
         index={link}
```
matildepark: And here.

pkg/interface/link/src/js/components/link.js
```diff
@@ -198,7 +197,7 @@ export class LinkDetail extends Component {
             commentPage={props.commentPage}
             members={props.members}
             popout={props.popout}
-            url={props.data.url}
+            url={props.url}
             linkPage={props.page}
             linkIndex={props.link}
```
matildepark: And here (and why do these have different names? I wish I got to refactor this.)

pkg/interface/link/src/js/components/root.js
Fang-: The permalink doesn't care about the `:index` here. But it's still used for the "click nth link from the list" flow, because I didn't completely overhaul the state structure. (iow, the "access by url" logic doesn't really know where to put the submission in the current state (because pages), so it just doesn't put it anywhere outside itself. (And state modification is considered a no-no outside of reducers anyway?))

matildepark: Going to `~link/~/default/0/2/aHR0cDovLzRjaGFuLm9yZw==` crashes, unable to find the `time` of `undefined`, so it seems to still pull its information from the `nth-link from list` flow, not the `:encodedUrll`?

Fang-: Cannot reproduce, works fine for me. Are you running an older version of `/app/link-view` or `/lib/link` maybe?

no-repro

matildepark: Nope — I mean to change the second number in the URL and go directly to it. If that index isn't on the page (say, `19` in your screencap's case) it'll crash.

matildepark: In this case imagine someone loads their browser history and clicks a page that supposedly corresponds to the permalink in question (after all, that's why we even have the encodedUrl param, isn't it?), it'll load some random link that corresponds to that index on that page.

vvisigoth:
Yes please imagine this.
—
~poldec-tonteg
http://urbit.org

pkg/interface/link/src/js/components/root.js
Fang-: Oh I see, can repro now. I'll work to fix this!

(But also, some quick playing around shows it _does_ work if I navigate to a url with any arbitrary _in-range_ index number. And since the bounds there only ever _increase_, your case _shouldn't_ happen in practice?)

matildepark:
I see now you mentioned the index thing from the start, ha — I'm still waking up clearly — but it seems like we need to redo that entire flow if we're going to actually solve this route problem since it always pulls from index; seemingly disregarding the permanent, encoded URL param.

pkg/interface/link/src/js/components/root.js
matildepark: We should be defensive and never allow a crash to happen if we can help it, right?

Fang-: As said,

> I'll work to fix this!

Just wanted to clarify the limits of the issue & my implementation.

> seems like we need to redo that entire flow if we're going to actually solve this route problem since it always pulls from index; seemingly disregarding the permanent, encoded URL param

We don't, because as described above, it can work, _does_ work, and only needs a little more cleaning up elsewhere I'm pretty sure. Will report back after working on this.

Fang-:
@matildepark should be all good now.

no-more-repro

matildepark:
Oh wait, contact-update reducer tho?

Fang-:
Oops, lost sight of that one. On it!

Fang-:
Huh, looks like everyone literally just copies each other's files in. Seems... sub-optimal.

Also sub-optimal is my fix for popout links being broken. They work now, but we really shouldn't be manually rebuilding those all over the place. Tried my hand at a refactor for that, but seems a bit too hairy to do quickly.

Merge commit contains a rebuild. (Did just a dev build now, but that puts me at at the opposite side of the diff size? /shrug)

matildepark:
Maybe a <PopoutLink> component that, if we’re not in a popout, takes the current location, splices “popout/“ into the string after the indexOf “~link/“, and makes a <Link> component to that?

Fang-:
I'll give that a shot. In the mean time, please merge if this is approved now. (:

matildepark:
It was fun being Mr. Manager.

PR #2245, opened 2020-02-07
link: hotfix - add "prop.groupPath" to link-submit.js

matildepark:
#2244 introduced a bug where you couldn't submit links because it used the old prop name in a child component. Quick fix to return it to functionality.


matildepark:
It's not been a good day for matilde ...

PR #2247, opened 2020-02-07
%spot type annotation for hoon

liam-fitzgerald:
Adds a `[%spot *]` type to the `note` type annotation definition.
These are added when the %dbug hoon is encountered. Done to enable
jump to definition in the language-server.
Adds support for the `%spot` type annotation to lib/xray

Vere-side, this is broken because the jet has not been updated. Guidance on this would be very much appreciated.


philipcmonk:
This makes sense to me.  It'll certainly be very nice to have "jump to definition".  Is the `%dbug` hoon the one that happens with `!:`?

Which jet(s) specifically don't respect this?

@joemfb you know these type annotations better than I do, does this seem right to you?

Also @pilfer-pandex, this has no semantic effect (this just adds another kind of "type hint"), but you should look at this.

liam-fitzgerald:
> This makes sense to me. It'll certainly be very nice to have "jump to definition". Is the `%dbug` hoon the one that happens with `!:`?

Yes
> Which jet(s) specifically don't respect this?

The mint:ut jet, I've disabled it to generate the pills, which is why CI is passing.

pilfer-pandex:
@philipcmonk, it is !:, although this is not immediately clear from reading the parser, sadly.

The code itself seems fine on first pass. I have some questions though:
- Can anyone explain why %hint on $type has two types in it?
- Can you say a bit more about how these hints are consumed such that this information is surfaced to the language server?
- I don't know anything about xray; why are there changes there?

@liam-fitzgerald, something is very off if you need to remove the jet in order to make this change work. Since about 6 months ago, the compiler jets only have caching logic in them: after checking a cache, they simply forward back to the hoon code. So if you're concerned about logic mismatch, don't be. I'm actually shocked it compiles at all without the caching in the jet. Can you say more about what the conflict with the jet is?

liam-fitzgerald:
> * I don't know anything about xray; why are there changes there?
xray is used to collect information about a type including rounding up the type hints. It's used for pretty-printing in lib/pprint.hoon. It also is supposed to collect inline documentation for a type, but as those structures are unused that functionality does not work. 

> * Can you say a bit more about how these hints are consumed such that this information is surfaced to the language server?
Language server can get the type of the symbol under the cursor. The plan is to xray the type, which should return the spot hint for that type, thus revealing the file/position of the type definition. This should also dovetail nicely if/when inline documentation arrives as xray should surface that as well.



> @liam-fitzgerald, something is very off if you need to remove the jet in order to make this change work. Since about 6 months ago, the compiler jets only have caching logic in them: after checking a cache, they simply forward back to the hoon code. So if you're concerned about logic mismatch, don't be. I'm actually shocked it compiles at all without the caching in the jet. Can you say more about what the conflict with the jet is?
Currently trying to work this out. There were some pill shenanigans that were preventing a build, and so I wondered if it was actually a jet mismatch and removed the hint but that may be unrelated to the real issue which was probably me trying to use a solid pill for hoon dev. Will investigate further

liam-fitzgerald:
> @liam-fitzgerald, something is very off if you need to remove the jet in order to make this change work. Since about 6 months ago, the compiler jets only have caching logic in them: after checking a cache, they simply forward back to the hoon code. So if you're concerned about logic mismatch, don't be. I'm actually shocked it compiles at all without the caching in the jet. Can you say more about what the conflict with the jet is?

Update: it was a pill issue, have just pushed with the jet enabled again

joemfb:
@pilfer-pandex, the `$note` includes the type of the subject (`p.p.sut`), and the type of the subsequent expression (`q.sut`). (The annotation is at `q.p.sut`.)

joemfb:
@liam-fitzgerald, do you have a version of the language server somewhere that consumes these changes? Would it be fine for the %spot hint inclusion to be gated by a compile-time flag?

liam-fitzgerald:
>@liam-fitzgerald, do you have a version of the language server somewhere that consumes these changes? Would it be fine for the %spot hint inclusion to be gated by a compile-time flag?

Yes and yes. However, the issue is then surfacing that gate in ford, through which the langserver does all its builds. 

The notion of a debug build is fairly generalisable and shouldn't necessarily be restricted to hoon builds. As such, I think we should add a new %trace silk that takes a silk and enables debug artifacts (spot hints, doc comments) for that silk. 

cc: @belisarius222

liam-fitzgerald:
Should be ready for merge now. The associated changes to ford will come in a followup PR.

joemfb:
I untagged @philipcmonk, as he's unavailable for the time being. @jtobin, this should be all set.

jtobin:
@liam-fitzgerald I'll get around to this lickety-split.

PR #2251, opened 2020-02-07
link: minimum viable un/seen status

Fang-:
Colors the relative time on items in the link list green if we haven't interacted with them yet. Allows you to clear this on everything at once with a simple button.

unread

I want to do more work on this to get unread counts in the sidebar (and hide the "mark all as seen" button if there's nothing new), but want to get this merged in so that @vvisigoth has a stable backend state type to test against.


Fang-:
> Seems straightfoward, I would just take out the bundled js.

I, uh, ok. I'll just stop touching compiled js from now on unless specifically instructed.

Thanks, merging!

PR #2258, opened 2020-02-10
chat-hook: evaluate result-less %code messages

### LINKED ISSUE #2257: Dojo code entered in landscape chat does not show result in landscape
I have seen multiple times when someone enters code into their landscape chat properly preceeded by # and they see the output but no one else sees the output.

We just saw an instance of this in /hoontards but I have seen it else where. The command that we used most recently was

#(trim 1 "abcde")

Fang-:
Prior to sending them out to the messaging target.

Fixes #2257.


loganallenc:
LGTM

PR #2259, opened 2020-02-10
link: unread counters in sidebar

Fang-:
Adds little unread counters to the link counts in the sidebar. (Disappear if there's nothing new.)

image


matildepark:
The mocks for Publish's unread counter should be followed here (`2 unread` instead of `(2)`), but I suppose I shouldn't necessarily hold this back because of that.

Fang-:
image

PR #2261, opened 2020-02-10
publish: "new post" button controlled in state

matildepark:
Previously, if you created a new notebook, you won't see "new post" until you refresh; it's done once on render.

This moves the check for write permissions to its own function, and calls that function on component mount and update; it also short circuits the check if the notebook belongs to your ship. 

This is a tradeoff; I can't think of a good edge case where you own a notebook you can't write to at the moment (*especially* outside a pre-existing group context, which is the only time when a write group will be created and thus not appear in time for the update check — and the only time in which the check for the write group prop fails anyway is *immediately after* creating a notebook without a pre-existing group).

### DIFF
--- a/pkg/interface/publish/src/js/components/lib/notebook.js
+++ b/pkg/interface/publish/src/js/components/lib/notebook.js
@@ -11,6 +11,9 @@ import { Settings } from './settings';
 export class Notebook extends Component {
   constructor(props){
     super(props);
+    this.state = {
+      canWrite: false
+    };
 
     this.onScroll = this.onScroll.bind(this);
     this.unsubscribe = this.unsubscribe.bind(this);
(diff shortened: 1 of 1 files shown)


pkg/interface/publish/src/js/components/lib/notebook.js
```diff
@@ -41,11 +44,31 @@ export class Notebook extends Component {
...
 
   componentDidMount() {
+    this.canWriteCheck();
     if (this.props.notebooks[this.props.ship][this.props.book].notes) {
       this.onScroll();
     }
   }
 
+  componentDidUpdate(prevProps) {
+    if (prevProps !== this.props) {
```
ixv: Isn't this condition a problem for deeply nested objects?

matildepark: We're not checking the whole object; just checking if any of its first levels get initialised.

matildepark:
To the best of my understanding:

It renders the component once and keeps it around, unless you give each instance of the component a key in `root.js`; some things reactively update when a prop updates. If I `.map` a prop or a variable assigned to a prop, the prop updating will reactively change it in the view. For some reason, this was a one-off render (I think it didn't even update if you clicked away and came back?); this PR's changes aren't even really reactive, so much as it's catching the cases when the underlying data is initialised and making it check the status. 

When we create a write group, the component doesn't seem to care or notice and thus update the render ... probably because no variable we use is directly assigned to it? I think? It's possible we could make this more directly reactive than this if we are clever with assignment, but I think @loganallenc would have a better idea than me.

pkg/interface/publish/src/js/components/lib/notebook.js
loganallenc: You asked the right question, and you had the right response! :) I love this

loganallenc:
Let me know the conditions under which you want it to update and I can make a follow-up commit on this branch.

loganallenc:
I have time today, so feel free to message me if it's easier to convey.

matildepark:
@loganallenc We want the view to update if `this.props.groups` changes; upon change, it should check if the write group exists. If we're in the write group, display the button.

The main replication test I can think of is just creating a notebook without a pre-existing group; it will have to create a write group, transition to the notebook.js view, and notebook.js will have to see that the group has been added, and look for our ship in it.

matildepark:
I ... can no longer reproduce "new post" not appearing until refresh on os1-rc. Closing.

PR #2269, opened 2020-02-11
eyre: add new os1 login screen

matildepark:
Kudos to @ixv for the help.

Screen Shot 2020-02-10 at 9 01 11 PM

Screen Shot 2020-02-10 at 9 01 19 PM

cc @urcades

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -501,7 +501,7 @@
                 ; Bridge ↗
               ==
               ;a
-                =href  "https://urbit.live"
+                =href  "https://urbit.org/using/install/#id"
                 =style  "margin-left: 8px; color: #2aa779;"
                 =target  "_blank"
                 ; Purchase
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -350,104 +350,165 @@
   ;html
     ;head
       ;meta(charset "utf-8");
-      ;title:"Sign in"
+      ;meta(name "viewport", content "width=device-width, initial-scale=1, shrink-to-fit=no");
+      ;title:"OS1"
       ;style:'''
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -350,104 +350,165 @@
...
+            ;p.pr2.pr0-ns.pb2
+              ;a(href "https://bridge.urbit.org", target "_blank")
+                ;span.dn.dib-ns.pr1:"Open"
+                ; Bridge ↗
+              ==
+              ;a
+                =href  "https://urbit.live"
+                =style  "margin-left: 8px; color: #2aa779;"
+                =target  "_blank"
+                ; Purchase
```
loganallenc: This should not be here. Please remove this.

matildepark: Tagging @urcades

urcades: Oh yes — this was meant to point to a page in our docs about purchasing planets, rather than a direct link to some market that we don't own.

I think instead, we can link to this page:

https://urbit.org/using/install/

and anchor link to the "Getting an Urbit ID" subheader.

matildepark: Mk, on it. I recall some oral discussion about "leads you to Urbit Live or whatever" but nothing specific ...

urcades: Yeah I actually think this would be a valid test, if we did A/B/whatever testing — obviously optimizing for the option that gets more people into Urbit is the preferred route, but I have no idea what that might be atm.

My hunch is that this is a decent route to provide for teaching more people about what obtaining an Urbit ID is, but idk — would love to get more data that supports a more firm position.

urcades: cc @vvisigoth : Figure you might be interested in this train of thought ^

PR #2271, opened 2020-02-11
patp now uses ~, and inter on nickname

jyng:
Screen Shot 2020-02-10 at 7 38 51 PM
Screen Shot 2020-02-10 at 7 38 25 PM

Would be curious if there's a better way to write this logic @loganallenc @matildepark


pkg/interface/chat/src/js/components/lib/message.js
```diff
@@ -167,10 +167,10 @@ export class Message extends Component {
       let color = "#000000";
       if (contact) {
         name = (contact.nickname.length > 0)
-          ? contact.nickname : props.msg.author;
+          ? contact.nickname : "~" + props.msg.author;
         color = `#${uxToHex(contact.color)}`;
-      } 
-
+        console.log(name);
```
matildepark: This can go.

jyng: lmao my bad

pkg/interface/chat/src/js/components/lib/message.js
```diff
@@ -167,10 +167,10 @@ export class Message extends Component {
       let color = "#000000";
       if (contact) {
         name = (contact.nickname.length > 0)
```
matildepark: Cleaner? Idk, you can do \`~${props.msg.author}\` if you use backtick quotes like that, I guess.

matildepark:
But honestly this looks concise enough; I wouldn’t necessarily worry about making it more and more clean. Inline is best unless it detracts from legibility or has excessive logic, in which case you’d start refactoring it outside the flow or in its component (if it’s a component, anyway). This is fine inline and the ternaries are basic.

pkg/interface/chat/src/js/components/lib/message.js
jyng: Hmm, could wee a variable that prepends the sig? This would make it easier to QA.

matildepark: (Idk why my phone made it do this a line too high.) Honestly this is fine as is imo but I’ll wait and see what the other reviewers think too.

jyng:
Ok will send this PR to the boss person

loganallenc:
Jimmy, please run gulp bundle-prod to minify these files, and rebase to remove the "large unminified versions". Otherwise looks good.

pkg/interface/chat/src/js/components/lib/message.js
loganallenc: remove this console.log

pkg/interface/chat/src/js/components/lib/message.js
loganallenc: I prefer Matilde's version of this as well

pkg/interface/chat/src/js/components/lib/message.js
matildepark: I think he removed it already in 8f580aa3bb4e275e433bfe2dcd0c17d8e5f95faa

jyng:
Do I have to rebase to remove? You just want to me to minify the files in `dist/` right?

loganallenc:
Merged and done. Thanks!

PR #2272, opened 2020-02-11
link fe: various fixes causing improvements

Fang-:
Please see commit messages for full changelog. They should be reviewable individually.

Primarily fixes:
- No unread counts on non-private channel
- "NaN links" after first link submission
- Permanent "0 contributors"

### DIFF
--- a/pkg/interface/link/src/js/lib/util.js
+++ b/pkg/interface/link/src/js/lib/util.js
@@ -14,6 +14,17 @@ export function uuid() {
   return str.slice(0,-1);
 }
 
+export function getContactDetails(contact) {
+  contact = contact || {
+    'nickname': '',
+    'avatar': 'TODO',
+    'color': '0x0'
+  };
+  const nickname = contact.nickname || '';
+  const color = uxToHex(contact.color || '0x0');
+  return {nickname, color};
+}
+
 // encodes string into base64url,
 // by encoding into base64 and replacing non-url-safe characters.
 //
@@ -117,6 +128,8 @@ export function deSig(ship) {
   return ship.replace('~', '');
 }
 
+//TODO look at uxToHex wonky functionality
+//TODO what does "wonky functionality" refer to?
 export function uxToHex(ux) {
   let value = ux.substr(2).replace('.', '').padStart(6, '0');
   return value;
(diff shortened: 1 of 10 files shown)


pkg/interface/link/src/js/lib/util.js
```diff
@@ -117,6 +128,8 @@ export function deSig(ship) {
   return ship.replace('~', '');
 }
 
+//TODO look at uxToHex wonky functionality
+//TODO what does "wonky functionality" refer to?
```
matildepark: @loganallenc has more here — essentially I *think* this utility had a weird thing going on where it wasn't returning the hex value set, but usually a digit off? I think it might have had something to do with how its returned value operated with the view? I think Contacts uses this utility, but it might have a fix; I know he removed the `#` from the Sigil Color field entirely at least partly because of this.

matildepark:
Regarding e801316:

> We were passing in a contacts object, which may not always contain all
group members.

I don't know if this is actually true. The correct choice is passing the metadata-less group store information, yes; but if you're in a (non-sig) group, you're in contacts — just with no metadata set. AFAIK. Another @loganallenc question. 

I believe the member counts being incorrect was just previous iterations of Contacts not including the owner in the group; but in that case, they wouldn't be in the group set, either. It wasn't specifically omitted from the contacts object. It's moot because it's still an improvement but I'm unsure about this specific rationale cited here.

matildepark:
Granted, it seemed non-fixed when you went into this. I don't know. (Doing testing now...)

Fang-:
> I don't know if this is actually true. The correct choice is passing the metadata-less group store information, yes; but if you're in a (non-sig) group, you're in contacts — just with no metadata set. 

I observed it to be true. It seems to me then that if no metadata is set for a ship, then no object is returned for it. Contact groups that I added ships to that weren't live at the time weren't included in the intial-contacts, and neither was I (the user ship) until I set metadata for myself.

matildepark:
Ah, I've disproved myself — possibly functionality changed or I was wrong this whole time, but there it is — contacts not added to the object but in the group.

pkg/interface/link/src/js/lib/util.js
loganallenc: Yes, please update to use my modified uxToHex util in the contacts JS

loganallenc:
https://github.com/urbit/urbit/pull/2272#issuecomment-584756786

Contacts do not show up in the contacts object unless the person has already shared their metadata. They will show up in the group regardless.

PR #2275, opened 2020-02-11
chat-cli: no longer allow %mailbox and %journal chats

Fang-:
Also adds in some casts to ensure sane action data. The chat-view-action
changed in this commit should've been nest-failing.

@loganallenc this (still?) doesn't automagically do the host prefixing on these paths. Should I be doing that during construction of the `%create` action here?

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -37,6 +37,9 @@
 +$  glyph  char
 ++  glyphs  "!@#$%^&()-=_+[]\{};'\\:\",.<>?"
 ::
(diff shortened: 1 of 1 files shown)


Fang-:
@loganallenc I guess that's a no to my question then?

Fang-:
Also @loganallenc `;invite`/permission management still works on `r | w | rw` basis, touching the permission store at `.../read` and `.../write` paths. I'm guessing that needs to be changed too? Do I just pull out the r/w distinction and touch at unmodified paths?

loganallenc:
Touch at unmodified paths, yes.

PR #2285, opened 2020-02-13
single-block +mink 2

jtobin:
(This PR replaces #2046.  @frodwith's text below is pasted verbatim from there.)

Our virtual nock interpreter +mink defines the semantics of the virtual nock operator 12 (the .^ rune in hoon, often called "scry" or "wish") as well as the semantics of exception handling. You give +mink a subject, a formula, and a gate for doing namespace lookups. Instead of returning a product, +mink returns one of:
[%0 product]
[%1 (list blocked-paths)]
[%2 stack-trace]

+mink is supposed never to crash while running the nock formula, but can crash while executing a scry if the scry gate itself crashes.

Why a list of blocked paths? Some nock operators compute one thing from one subject. Some compute two (autocons, equality, nock(2)). +mink takes that opportunity (or should, the current implementation is subtly wrong for 5) to compute both arguments (as long as they do not crash) and, if they both block, concatenate the blocks together. This has the effect that sometimes you would be able to report more of the things that would block in a program - if i had a cons of two scrys, i could report that both of them block, giving the caller of +mink the opportunity to fetch more things before calling me again. So far, so good.

The problem appears with the +mink jet. Naturally we jet +mink by running the normal nock interpreter. There are two ways to support "block welding" in a nock interpreter. One is as +mink does, by having the return type include a case for blocks, and carefully welding blocks - that is by "directly" jetting mink. Now you have two nock interpreters (although nock can simply call mink with a crashing scry gate and unpack the result, much like +mack in hoon.hoon).

Another approach (the one taken by all current implementations) is to use an interpreter that relies more on an external runtime - specifically, stack trace handling is external (not part of the return type) and blocked scrys become exceptions. the +mink jet then becomes the exception handler for these block exceptions. This approach restricts us to a single block, however, since throwing the exception generally means we are done computing and won't try (for example) the other side of an autocons to find more blocks. If we want to correctly jet mink (without what amounts to a jet mismatch), we also need to trap for the block exceptions any time we need to compute two things from the subject - for example, we need to wrap a try around each of the head and tail sides of an autocons and weld them together if they are blocks. In vere this would mean descending into two inner roads every time we cons a cell. In other environments the cost is less extreme, but nonzero. Optimizations are possible (for instance not trapping if the subformulas don't contain 12s), but at complexity cost.

The current situation is a jet mismatch, and therefore unacceptable. Either +mink should be changed to report only a single block, enabling the exception handling regime described above without jet mismatches, or the underlying interpreter (at least the one run during +mink) should be structured more like mink in order to properly weld and propagate block lists without unacceptable performance overhead. The jet calling convention would also likely need to change to accommodate the interpreter structure. I feel it would be less disruptive simply to limit +tone (the +mink result type) to a single block in its block case, even with the downside of being theoretically able to report less blocks to the mink caller in some cases. We are practically limited to a single block at present anyway.

This pull requests implements that strategy. In addition, the current implementation of +mook and +mush use the clam form of calling a mold as a gate, expecting the mold to bunt if the sample does not conform to the mold. Alas, that behavior has changed; these functions are updated accordingly to accomodate the change in mold semantics. They will no longer spuriously crash when not given carefully typed input.

I have also taken the liberty of improving some of the names. Since a primary purpose of +mink is to be read, it should be readable.

A word to the wise for those following discussion on the urbit-dev mailing list: this PR does not address stack trace nondeterminism.


jtobin:
(Per @joemfb's comment on the superceded PR, this should resolve #357.)

frodwith:
Re @joemfb:
* Dunno what's up with the failing ford test yet.
* the old implementation of +mush should be removed before merge
* This is a reversion to the intended behavior as originally written. Arguably there could be a default tank (which is what would have happened before) instead of skipping weird crash items. The behavior was changed unintentionally when the behavior of molds was changed. I think the burden of proof lies with the idea that it should crash. Leaving that aside, the whole point of calling mock is to trap for exceptions and get a stack trace if it crashed. Please make the case that assembling the stack trace should be able to crash, that seems crazy to me.
* In addition, the current "pass through" of hints is subtly wrong and should be changed to use a constant [1 clue] and [1 product] to avoid spuriously product-hinting tones. This is the fabled "extra commit" that I'm supposed to add to this PR, although it doesn't really block a merge if I don't get around to it soon. Just making note of it here.

jtobin:
Re: the previously-failing test, @frodwith linked it to a gate called exclusively by `%ford` that was was still expecting the old tone format.  Should be fixed in d6b28db.

pkg/arvo/sys/hoon.hoon
```diff
@@ -6317,119 +6317,191 @@
...
+        =/  clue  $(formula clue.formula)
+        ?.  ?=(%0 -.clue)  clue
+        =;  next
+          ?.  ?=(%0 -.next)  next
+          :-  %0
+          .*  subject
+          [11 [tag.formula 1 product.clue] 1 product.next]
+        %=  $
+          formula  next.formula
+          trace    ?.  ?=  ?(%hunk %hand %lose %mean %spot)
```
joemfb: This is a good spot for `=?`.

pkg/arvo/sys/hoon.hoon
```diff
@@ -6317,119 +6317,191 @@
...
+        ?.  ?=(%0 -.path)  path
+        =/  result  (scry product.ref product.path)
+        ?~  result
+          [%1 product.path]
+        ?~  u.result
+          [%2 [%hunk (mush product.path)] trace]
+        [%0 u.u.result]
+      ==
+  ::
+  ++  mush
```
joemfb: This implementation doesn't appear to match `_n_mush()` in nock.c, which would be a jet mismatch in `+mink`. A straightforward port looks something like:

```
|=  path=*
^-  tank
:+  %rose
  [mid="/" hed="/" tal=""]
|-  ^-  (list tank)
?@  path  ~
:-  [%leaf ?^(-.path "" (trip -.path))]
$(path +.path)
```

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1358,7 +1358,7 @@
       =/  res  (mock [run %9 2 %0 1] (sloy ski))
       ?-  -.res
         %0  [%& !<(step:agent [-:!>(*step:agent) p.res])]
-        %1  [%| (turn p.res |=(a=* (smyt (path a))))]
+        %1  [%.n (smyt (path p.res)) ~]
```
joemfb: Should be `%|` for consistency.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1369,7 +1369,7 @@
       =/  res  (mock [run %9 2 %0 1] (sloy ski))
       ?-  -.res
         %0  [%& !<((unit (unit cage)) [-:!>(*(unit (unit cage))) p.res])]
-        %1  [%| (turn p.res |=(a=* (smyt (path a))))]
+        %1  [%.n (smyt (path p.res)) ~]
```
joemfb: Should also be `%|`

pkg/arvo/sys/hoon.hoon
```diff
@@ -6440,43 +6512,45 @@
...
+             leaf+(scow %p (mug dat))
+      $mean  ?@  dat  [leaf+(rip 3 dat) rep]
+             =/  mac  (mack dat -.dat)
+             ?~  mac  [leaf+"####" rep]
+             =/  sof  ((soft tank) u.mac)
+             ?~  sof  rep
+             [u.sof rep]
+      $spot  =/  sof=(unit spot)  ((soft spot) dat)
+             ?~  sof  rep
+             :_  rep
```
joemfb: I dislike have so many virtualization contexts in an inner loop, and re-entrant calls into them (`+soft` eventually calls `+mook`). Catching crashes in the stack trace production is very different from catching crashes in arbitrary code -- it's basically for shaking out bugs in the compiler or runtime. Stack trace production can already be pretty slow, and this is likely to make it slower. At the same time, it's more elegant -- and probably more correct -- to not crash here.

Regardless of the virtualization, it'd be nice to refactor the non-standard recursion here. This would easily translate to `+turn` or `+murn`, and the result should be simpler.

ohAitch:
reason number n+1 why removing mold-as-in-soft capabilities from the type
system was a bad idea…

joemfb:
This PR is now fully up-to-date. I resolved my own requested changes in #2900, and performed final cleanup and a bugfix here. These charges now target the "next breach" version of vere, and include:

- a rewrite of +mink, with "single-block" scry semantics, hint pass-thru, and generally improved clarity and style
- new "unitary" mule-style static virtualization in `+mole` (with use of the same in `+soft`)
- normalized stack-trace rendering, without potentially-crashing clams or reentrancy
- proper compilation of nock 6, allowing for invalid formulas in unevaluated branches
- modern style in all the virtualization gates

PR #2286, opened 2020-02-13
OS1: added members and settings page to publish

ixv:


pkg/interface/publish/src/js/components/lib/new-post.js
```diff
@@ -98,7 +100,7 @@ export class NewPost extends Component {
             popout={props.popout}
           />
           <button
-            className="v-mid w-100 mw7 tl pl4 h1"
+            className={"v-mid w-100 mw7 tl h1"+submitButtonPadding}
```
matildepark: You’ll need a space in one of these strings or it will concatenate the strings directly and make a monstrous monoclass

ixv: There is a space in submitButtonPadding though

matildepark: Oh, sorry you’re right. I saw it wrong x_x

pkg/arvo/app/publish.hoon
```diff
@@ -703,49 +706,29 @@
 ++  handle-permission-update
   |=  upd=permission-update
   ^-  (quip card _state)
-  ?+  -.upd
+  ?.  ?=(?(%remove %add) -.upd)
```
loganallenc: Don't send invites upon being removed from a blacklist

pkg/interface/publish/src/js/components/lib/subscriber-item.js
```diff
@@ -1,12 +1,149 @@
...
+
+    let width = 0;
+    let options = [];
+    if (this.props.section === 'participants') {
+      if (this.props.readPath === this.props.writePath) {
+        width = 258;
+        let url = `/~contacts${this.props.writePath}`;
+        options = [
+          <a key={0} className="tl pointer" href={url}>
+            Manage this group in the contacts view
```
loganallenc: Flagging this as something where the API logic and view logic should be broken apart. Pass in function calls to APIs as props, and options as props, rather than hardcoding

pkg/interface/publish/src/js/components/lib/settings.js
```diff
@@ -1,41 +1,39 @@
...
+    this.props.history.push('/~publish');
+  }
+
+  render() {
+    if (this.props.host.slice(1) === window.ship) {
+      return (
+        <div className="flex-column">
+          <p className="f9 mt3 lh-copy db">Delete Notebook</p>
+          <p className="f9 gray2 db mb4">
+            Permenantly delete this notebook. (All current members will no longer see this notebook)
```
matildepark: Permanently*

ixv:
Unless I'm misunderstanding, Logans point was not that the view dropdown should be split out from subscriber item into its own component, but that the API calls in subscriber-item should be defined in subscribers, and passed in to subscriber-item.

matildepark:
Oh. Personally I think both. We're going to use that dropdown element in other places.

pkg/interface/publish/src/js/components/lib/new-post.js
ixv: I just realized this change only looks good when the screen width is small enough, otherwise it looks bad. Going to revert this, not sure what the right solution is, but it shouldn't have been in this PR in the first place.

pkg/arvo/app/publish.hoon
loganallenc: good

pkg/interface/publish/src/js/components/lib/dropdown.js
```diff
@@ -0,0 +1,73 @@
...
+
+    return (
+      <div className="options relative dib"
+           ref={(el) => {this.optsButton = el}}>
+        <button className="pr3 mb1 pointer br2 pa2 pr4"
+                style={{backgroundColor: optionsColor}}
+                onClick={this.toggleDropdown}>
+          {this.props.buttonText}
+        </button>
+        <div className="absolute flex flex-column pa4 ba b--gray4 br2 z-1 bg-white"
```
loganallenc: position absolute wtf @matildepark

matildepark: The dropdown list is supposed to be directly below the dropdown prompt without taking up space on the layer below — absolute drops it immediately below outside the box model. This seems like the way you'd do that at first glance here. Can you explain the wtf

matildepark: I think the `flex flex-column` classes would just be overwritten by the inline style here btw @ixv

PR #2287, opened 2020-02-13
hoon: clarify comments for +rep:in and +rep:by

Fang-:
This has bitten me often enough that it's actually _really_ annoying now. I'm pretty sure this comment is at least _equally_ correct, and hopefully _more_.

"Replace" suggests this function either produces an updated set/map when done,
like `+snap`, or changes all values in-place, like `+turn`. In truth, it's more
similar to `+roll`, which does reduction/accumulation.

("Reduce" specifically was chosen because it maintains the mnemonic relation to
the arm name.)

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1337,7 +1337,7 @@
       a(r c)
     c(l a(r l.c))
   ::
-  ++  rep                                               ::  replace by product
+  ++  rep                                               ::  reduce to product
     |*  b/_=>(~ |=({* *} +<+))
     |-
     ?~  a  +<+.b
@@ -1592,7 +1592,7 @@
       a(r d)
     d(l a(r l.d))
   ::
-  ++  rep                                               ::  replace by product
+  ++  rep                                               ::  reduce to product
     |*  b/_=>(~ |=({* *} +<+))
     |-
     ?~  a  +<+.b


pilfer-pandex:
These functions take the initial accumulator from the initial sample of the passed-in gate? That's... :/

Fang-:
That's actually fine most of the time. But a real pain when it isn't... `$_` can help here, but has also caused trouble in the past. (With `+roll` in old `/app/hall`, to be specific. But we were `$_`ing a core/context, so...)

pilfer-pandex:
It's fine when the starting value you want is the same as the bunt value of
the type. Otherwise, you're in for some trouble. The worst is when you want
something different from the bunt but don't realize or remember. Then your
code will silently go wrong.

—
~pilfer-pandex

joemfb:
@Fang-, that problem was from before a complete rewrite of the mold system. I'd be very surprised if it's still an issue.

belisarius222:
@pilfer-pandex All of hoon's fold-like functions use this idiom.  I agree with your analysis that this can be a footgun, but I don't think I've ever run into that particular problem.

The problem I usually run into with Hoon's folding functions is that the accumulator is often a stateful core that's a mutant copy of part of the subject, but the original subject is still in lexical scope.  This makes it easy to forget to call the mutant core's arms instead of the stale original subject's arms.

Issue #2206, opened 2020-01-29
Haven't been able to join a notebook for about two months


somlus-savlev:
Hello, somlus-savlev here. 

Somewhere along the line (about 2 months ago, probably 2-3 weeks after the last network breach), the notebook ~minder-folden/the-new-era stopped updating altogether for me, so I unsubscribed and then could never resubscribe.  I get quite a long error seen here:

https://pastebin.com/XLWsWQiV

when I type in the usual command in dojo:

:publish|subscribe ~minder-folden %the-new-era

Unsubscribing with

:publish|unsubscribe ~minder-folden %the-new-era

and then resubscribing similarly does not work.  Any ideas on how to get reconnected or what is going on here? I see ~minder-folden just fine and communicate regularly with him in his /party chat.

philipcmonk:
Can you run `+timers` and give the output?  @belisarius this may be a stuck flow.

sdubyam:
You bet, here you go:

https://pastebin.com/BjYbMSUN

risruc-habteb:
I'm seeing the same issue with the same notebook.
I noticed through chat that I haven't gotten updates in about 1 month, so decided to try and unsub/resub.

Here is the dojo output upon unsubcribing and attempting to resubscribe.
https://hastebin.com/rehijasifa.pl

custom-jonathan:
If there's anything you need `~minder-folden` to run, let me know and I'll try it out.

belisarius222:
Hm, not sure what would cause the 'ack for nonexistent subscription' crash.  Maybe an Ames error-handling gone wrong, or even a Gall error-handling gone wrong?

That does seem like a lot of timers, on a lot of different Ames flows.  @philipcmonk do you have an idea of what that indicates?

risruc-habteb:
fwiw, here is my +timers output.
https://hastebin.com/pucicokizu.coffeescript

pepijndevos:
I'm also unable to join the Hoon school notebook:
```
:publish &publish-action [%subscribe ~rapfyr-diglyt %hoon-school-assignments]
out={[p=[wire=/collection/hoon-school-assignments ship=~rapfyr-diglyt term=%publish] q=[acked=%.n path=/collection/hoon-school-assignments]]}
subscribe wire not unique
%publish
/collection/hoon-school-assignments
[p=~rapfyr-diglyt q=%publish]
error in %publish
%watch-not-unique
! subscribe wire not unique
! %publish
[...]

Issue #2176, opened 2020-01-15
Unable to ;join specific channel, but other co-located channels work


paulie-g:
Posting as per ~master-morzod's request. I am unable to ;join ~somlus-savlev/hoontards, although |hi'ing him works and I can join, read from and write to other channels on ~somlus-savlev. The channel itself works fine from a different urbit comet instance at the same time.

See session output from the cli: https://pastebin.com/raw/xT2NfdGW

paulie-g:
See #2140 - looks like the same, or at least similar, issue.

Output of +trouble from the concerned planet ship:
```
> +trouble
[ [%base-hash 0v19.t3qr1.o5q90.j9160.7778e.qubos.egc2s.a4enc.6d0cs.nle6c.gu1mu]
  [%home-hash 0v19.t3qr1.o5q90.j9160.7778e.qubos.egc2s.a4enc.6d0cs.nle6c.gu1mu]
  [%our ship=~haclex-moppyx point='3260582959' life=[~ 1] rift=[~ 1]]
  [%sponsor ship=~nibfeb point='35887' life=[~ 3] rift=[~ 2]]
  [%dopzod ship=~dopzod point='4608' life=[~ 1] rift=[~ 0]]
  "Compare lifes and rifts to values here:"
  "https://etherscan.io/address/azimuth.eth#readContract"
  "  life - getKeyRevisionNumber"
  "  rift - getContinuityNumber"
  ~
```

joemfb:
@Fang-, is there a way to clear out the duplicate subscription? Is this what `:chat-cli %connect` is supposed to do?

Fang-:
`:chat-cli %connect` is completely unrelated. You might want to try `:chat-hook &chat-hook-action [%remove /hoontards]`? Or maybe with `[%remove /~somlus-savlev/hoontards]` instead. I think that should tell the chat-hook to stop being subscribed to that, which should clear the subscription, after which you can try a `;join` again.

cc @loganallenc for sanity check.

philipcmonk:
There's two subscriptions: one to get the backlog, then one for new messages.  This is crashing because the backlog subscription is opened twice.  Possibly `%remove` only cancels the "new messages" subscription?

If so, @loganallenc it should just cancel both -- it doesn't hurt to cancel a subscription that doesn't exist.

loganallenc:
Interesting. I'll look into this and fix it.

paulie-g:
[...]

Issue #2153, opened 2020-01-06
Empty Recent Tab


custom-jonathan:
https://youtu.be/8AbRM1swpOY

Fang-:
Is this maybe a case of #2079?

custom-jonathan:
I just unsubscribed from all notebooks, clicked the home icon to get back to my dashboard, clicked back to publish and the recent page is still blank (not showing any of my posts, either).

Resubscribing shows the notes in recent now, but my own posts aren't present.

custom-jonathan:
Posted a new note and it's showing in recent, but none of my preexisting notes are showing. #weird

ixv:
closing as pre-publish-rewrite issue

Issue #2079, opened 2019-12-11
publish: unsubscribe removes all posts from "Recent"


philipcmonk:
@ixv 

As reported on urbit-help and verified on my comet, if you subscribe to a few notebooks, then unsubscribe with `:publish &publish-action [%unsubscribe ~ship %notebook]`, it clears the "Recent" page.  Refreshing doesn't bring it back, and there are no errors in the js console.

Unsubscribing from all notebooks and resubscribing seems to work to recover.

ixv:
closing as pre-publish-rewrite issue

PR #2288, opened 2020-02-14
gall: subscription acknowledgment fix

philipcmonk:
Should fix many cases of "received 2nd watch-ack" errors.

outstanding.agents.state is a queue of what sort of message we sent to a
foreign app.  We use it so that when the acknowledgment comes back we
know whether to treat it as a watch-ack, poke-ack, or neither.  We used
to put this info in the wire, but this gave us a different ames flow,
which meant %leave and %watch didn't get associated (causing #2079).

The error was that when when retrieving the item from the queue, we put
the new 1-item-shorter queue back in outstanding.agents.state at a
different wire than it came from, so the queues never actually got
shorter, and acknowledgments of the wrong sort were commonly produced.
This caused problems mainly in situations where we poke and peer on the
same wire, and possibly when a subscription was cancelled.

Possibly related to #2206 and #2176.  I would expect this bug to cause
those issues, but I haven't verified the converse.  Also possibly
related to #2153 and #2079.

First commit is the fix, second makes it OTA-able.


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -610,12 +610,16 @@
...
+          =/  full-wire  sys+wire
           =/  stand
-            %+  ~(gut by outstanding.agents.state)  [sys+wire hen]
-            *(qeu remote-request)
-          ~|  [sys+wire=wire hen=hen stand=stand outs=outstanding.agents.state]
+            %+  ~(gut by outstanding.agents.state)  [full-wire hen]
+            ::  default is do nothing; should only hit if cleared queue
+            ::  in +load 3-to-4
+            ::
+            (~(put to *(qeu remote-request)) %leave)
```
belisarius222: Why do we insert a %leave into the queue of outstanding requests if it was empty?  I get that we might have cleared the queue during migration, but I don't see how this would help.  I feel like I'm misunderstanding something here.

We could be in the middle of some other request that isn't a %leave, right?  ... in which case when we hear the ack on whatever that request was, we'll misinterpret it as an ack on this %leave that we never actually sent.

philipcmonk: Yeah, `%leave` isn't correct, but it's impossible to know for certain the right answer.  This feels conservative because it'll just give no acknowledgment at all.

It occurs to me this could screw up `%ping` because it's expecting a poke-ack, and if it doesn't get it, then it might hang.  Don't really have a good answer.  I could just fix it to `%poke`, which also feels not great, but it's less likely to cause a real problem I guess.

belisarius222:
I feel like +load needs to handle this. It could make a separate piece of state that holds the old outstanding requests so they can be dealt with correctly later, and then it would clear the outstanding request state.

Or maybe all the apps we have that could have been susceptible to this bug should clear their request state and reinitialize somehow.

I think we don't have to solve this in the general case, since there are only a few apps that could have been affected.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
We don't know how many of the items in the queue are already acked, so it's impossible to know what the right answer is.

The set of apps affected by this is any that have pending (ie unacknowledged) subscription or poke requests.  Most of the time that'll only be messages to another ship which is offline.  Maybe the best answer is to give both acknowledgments?  That way at least the right thing will happen, even if the wrong thing does too.

philipcmonk:
Okay, I made it give both acks if we don't know which it should be.  I tried this on a few ships and didn't run into a problem.

PR #2291, opened 2020-02-14
metadata-store: associate groups with resources within apps, and allow various types of queries

loganallenc:
Added first pass of metadata-store, a data store for application metadata and mappings 
between groups and application data (chatrooms, publish notebooks, etc).

I will add documentation to this before merging it, but I wanted to put out this version for review. The hook will be started next and will be in a separate PR, then integration work will be merged in as its own PR.

### DIFF
--- a/pkg/arvo/app/metadata-store.hoon
+++ b/pkg/arvo/app/metadata-store.hoon
@@ -1,23 +1,10 @@
 ::  metadata-store: data store for application metadata and mappings 
 ::  between groups and application data (chatrooms, publish notebooks, etc)
 ::
+/-  *metadata-store
 /+  default-agent
 |%
 +$  card  card:agent:gall
-+$  group-path  path
-+$  app-name  @tas
-+$  app-path  path
-+$  metadata
-  $:  title=@t
-      description=@t
-      color=@ux
-  ==
-::
-+$  metadata-action
-  $%  [%add =group-path =app-name =app-path =metadata]
-      [%remove =group-path =app-name =app-path]
-  ==
-+$  metadata-update  metadata-action
 ::
 +$  versioned-state
   $%  state-zero
@@ -37,10 +24,10 @@
 ^-  agent:gall
 =<
   |_  =bowl:gall
-  +*  this        .
+  +*  this           .
       metadata-core  +>
-      mc          ~(. metadata-core bowl)
-      def         ~(. (default-agent this %|) bowl)
+      mc             ~(. metadata-core bowl)
+      def            ~(. (default-agent this %|) bowl)
   ::
   ++  on-init            on-init:def
   ++  on-save   !>(state)
@@ -65,7 +52,8 @@
     |^
     =/  cards=(list card)
       ?+  path  (on-watch:def path)
-          [%all ~]  (give %metadata-initial !>(associated))
+          [%all ~]      (give %metadata-initial !>(associated))
+          [%updates ~]  ~
       ==
     [cards this]
     ::
--- a/pkg/arvo/app/metadata-store.hoon
+++ b/pkg/arvo/app/metadata-store.hoon
@@ -1,0 +1,164 @@
+::  metadata-store: data store for application metadata and mappings 
+::  between groups and application data (chatrooms, publish notebooks, etc)
+::
+/+  default-agent
+|%
++$  card  card:agent:gall
++$  group-path  path
++$  app-name  @tas
++$  app-path  path
++$  metadata
+  $:  title=@t
+      description=@t
+      color=@ux
+  ==
+::
++$  metadata-action
+  $%  [%add =group-path =app-name =app-path =metadata]
+      [%remove =group-path =app-name =app-path]
+  ==
++$  metadata-update  metadata-action
+::
++$  versioned-state
+  $%  state-zero
+  ==
+::
++$  state-zero
+  $:  %0
+      associated=(map [group-path app-name app-path] metadata)
+      group-indices=(jug group-path [app-name app-path])
+      app-name-indices=(jug app-name [app-path group-path])
+      app-path-indices=(map [app-name app-path] group-path)
+  ==
+--
+::
+=|  state-zero
+=*  state  -
+^-  agent:gall
+=<
+  |_  =bowl:gall
+  +*  this        .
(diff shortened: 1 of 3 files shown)


pkg/arvo/sur/metadata-store.hoon
```diff
@@ -0,0 +1,23 @@
+|%
++$  group-path  path
++$  app-name  @tas
++$  app-path  path
```
Fang-: These feel like they might want an encapsulating `+$  resource  [=app-name =app-path]`.

pkg/arvo/app/metadata-store.hoon
```diff
@@ -0,0 +1,172 @@
...
++$  versioned-state
+  $%  state-zero
+  ==
+::
++$  state-zero
+  $:  %0
+      =associated
+      group-indices=(jug group-path [app-name app-path])
+      app-name-indices=(jug app-name [group-path app-path])
+      app-path-indices=(map [app-name app-path] group-path)
```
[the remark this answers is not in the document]
loganallenc: Are you saying that app-path-indices should be a jug? Until we have user stories around multiple groups associated with a singular app-path, I beg to differ.

pkg/arvo/sur/metadata-store.hoon
loganallenc: I like that.

pkg/arvo/sur/metadata-store.hoon
```diff
@@ -0,0 +1,23 @@
...
++$  associated  (map [group-path app-name app-path] metadata)
```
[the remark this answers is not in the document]
loganallenc: It's good.

pkg/arvo/app/metadata-store.hoon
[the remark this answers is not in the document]
Fang-: That's what I'm saying.

The `associations` data structure has keys of `[group-path app-name app-path]`. This means that you can issue `%create` actions to store, say, both `[/a %x /y]` and `[/b %x /y]`. Since these both point to the resource `[%x /y]`, the latter action will _overwrite_ the former one in `app-path-indices`, even though they're both getting stored in `associations`. Your indices are no longer accurate!

I understand we have no use cases for "multiple groups per resource" right now, but if we don't let you get into that situation in the first place I see no harm in having the store just support it already, so we don't have to go back and change this in the future.

If we insist on not supporting it on the metadata-store level, then we should add a check for `%create` actions that prevents such "duplicates". (The questions for apps that _rely_ on this store for state storage then becomes, what if it doesn't want to store my state? But if they shouldn't be getting into that position in the first place, maybe that's fine...)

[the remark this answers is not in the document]
loganallenc: Okay, changed it to a jug.

pkg/arvo/app/metadata-store.hoon
```diff
@@ -73,15 +89,14 @@
...
     ^-  (unit (unit cage))
     ?+  path  (on-peek:def path)
-        [%x %all ~]               ``noun+!>(associated)
-        [%x %group-indices ~]     ``noun+!>(group-indices)
-        [%x %app-name-indices ~]  ``noun+!>(app-name-indices)
-        [%x %app-path-indices ~]  ``noun+!>(app-path-indices)
+        [%y %associations ~]      ``noun+!>(associations)
+        [%y %group-indices ~]     ``noun+!>(group-indices)
+        [%y %app-indices ~]       ``noun+!>(app-indices)
+        [%y %resource-indices ~]  ``noun+!>(resource-indices)
```
Fang-: I, uh, don't think this is how you use `%y` cares? They're for getting a "directory listing" of the given path. In context of apps, I think this means that if you support a `/something/[your-resource]` scry, then doing a `%y` scry on `/something` will give you all values of `your-resource` for which there's something to be scried out using the more specific path.

ie you want to give _keys_, not _values_. In that regard, certainly the `/associations` one here should be a `%x`, and probably the other ones as well. If you want to actually implement a `%y`, you'd do `~(key by associations)` etc.

pkg/arvo/app/metadata-store.hoon
```diff
@@ -73,15 +89,14 @@
...
         [%x %metadata @ @ @ ~]
       =/  =group-path  (stab i.t.t.path)
```
Fang-: Noticing this now (late, sry), but just `stab` here is kind of surprising. It works on `(stab '/some/path')`, but that cord isn't actually path-legal, since it contains slashes. (Imagine your frontend sending a path-as-a-string that's `'/outer/some/path/deeper'`. Good luck trying to figure out if the middle thing is supposed to be `/some/path` or `/some` or whatever.)

In my proposal in the email thread, I included `(scot %t path)` for encoding, and `(slav %t part.of.path)` for decoding. Those are crucial, because they turn your `@t` `'/some/path'` into `@ta` `~.~~~2f.some~2f.path`, which _is_ path-legal. You'll run into issues with this if you don't do that safety-conversion.

pkg/arvo/sur/metadata-store.hoon
```diff
@@ -1,23 +1,26 @@
...
++$  app-path      path
++$  resource      [=app-name =app-path]
++$  associations  (map [group-path resource] metadata)
 ::
 +$  metadata
   $:  title=@t
       description=@t
       color=@ux
+      date-created=@da
+      creator=@p
```
Fang-: Kinda hesitant to just casually throw these in, but can't make a solid case against them...

loganallenc: Came from a discussion with Gavin. Easy to revise these over time, but we wanted to add a few obvious ones.

pkg/arvo/app/metadata-store.hoon
loganallenc: Done, thank you.

pkg/arvo/app/metadata-store.hoon
loganallenc: Done, thank you.

ixv: If it's going to return full keys rather than path segments, maybe it should be %t rather than %y.

PR #2293, opened 2020-02-14
king: Use newer static-haskell-nix script for KH build.

benjamin-tlon:
Running this build now, not yet sure if works.

First build on a new machine takes hours.


benjamin-tlon:
Make sure to run `cachix use static-haskell-nix` before trying to build this.

Issue #2024, opened 2019-12-05
Chat doesn't scroll in firefox


philipcmonk:
Don't see an actual issue for this, and it comes up often.  This has been broken for a while now, is there any fix on the horizon?

matildepark:
It comes from nested flexbox. See [here](https://stackoverflow.com/questions/28636832/firefox-overflow-y-not-working-with-nested-flexbox) and [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1043520). 

I can't find the ticket, but I recall @loganallenc saying that it's essentially a browser bug on Flexbox spec, not our bug — I mean, Safari and Chrome both handle it.

(FWIW, when reimplementing chat I tried the hack provided in those queries but to no avail. We would basically have to rethink the layout entirely, I think, avoiding nesting flexboxes.)

philipcmonk:
So our answer is... don't support Firefox?  If it's been a bug for five years, Firefox isn't going to fix it any time soon, which makes it our bug.

loganallenc:
Yes. We do not support Firefox.

philipcmonk:
There's a few ways to get a scrollbar, the most straightforward of which is to spit this div into two nested ones, where the outer one has `overflow-y-scroll` and the inner one has `flex flex-column-reverse`.

https://github.com/urbit/urbit/blob/73cffdb28ff3baf2f11c376ae867725dee3ca5ba/pkg/interface/chat/src/js/components/chat.js#L255-L258

What I don't know is how to get it to open at the bottom instead of the top and auto-advance down when new messages come in.

litpub:
I've been exploring another option, which is to throw out flex-column-reverse, and use the existing `scrollToBottom` when new messages come in.

This shows it working:

![hoax](http://g.recordit.co/HtU0Dy8C5J.gif)

But it's a bit of a hoax in that I'm just calling `scrollToBottom` in all `componentDidUpdate` calls. I'm trying to get it to be smarter, and only do so on new message receipt, but it appears that `prevProps.envelopes.slice(0).length` and  `props.envelopes.slice(0).length` are always the same value

litpub:
https://github.com/litpub/urbit/commit/14f9d44fff4e71fa0bdc48ac4980df0b5c793f2f

litpub:
This is merged into os1-rc now

matildepark:
[...]

PR #2295, opened 2020-02-15
Firefox scroll support for chat

litpub:
Attempts a fix for https://github.com/urbit/urbit/issues/2024 to support scrolling in chat for Firefox.

Currently, the flex-box layout uses `column-reverse` to implement the expected behaviors of a chat interface. The problem is that Firefox has a 5-year-old bug relating to the use of scroll support and `column-reverse`.

One approach, this one, is to just use `column` instead, and to deal with having to force scroll events on new messages.

This PR:

- Works by just using column
- un-reverses messages, now that `column-reverse` is gone
- Adds a named parameter to `scrollToBottom` so that you can control animation
- Calls `scrollToBottom` in `componentDidUpdate` so that we always scroll to the bottom on new messages
- Reorders the scroll element so that it appears below the message fragments in the DOM, as a consequence of dropping `column-reverse`
- Tidies up `render` a bit, extracts building the messages array into `messages`
- *Does not* implement the scroll lock behavior of chrome/safari on Firefox - I thought the existing function was too long, and should be refactored, and I wanted to get feedback first

Notable changes and tradeoffs:

1. New chats will render the first message at the top of screen, like so
Screenshot 2020-02-14 13 09 38

2. Others that I might not know about?

If this is a bad approach, lmk, I'm not fragile - I just want to fix the damn thing so I can use FF.


litpub:
Also should I include the built pkg/arvo/app/chat/js/index.js or not? Seems kind of dangerous in a PR.

loganallenc:
thanks! first off, run gulp bundle-prod and do a commit --amend and force push so that you're only committing the minified version of index.js

litpub:
@loganallenc done

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -168,13 +175,10 @@ export class ChatScreen extends Component {
    }
  }
 
- render() {
+ messages() {
```
loganallenc: Naming. messagesWithPending() would be better

matildepark:
> New chats will render the first message at the top of screen, like so

Have you tried `align-items: flex-end` (should just be `items-end` in indigo-static classes)? Unsure if that would introduce issues with a full chatlog but might work.

litpub:
Thanks for round of feedback all - will push some changes shortly.

benjamin-tlon:
Hell yes.

loganallenc:
@litpub any word? no rush, just don't want to drop this.

litpub:
@loganallenc revisiting now

litpub:
@loganallenc force pushed - also the CSS file is diffing now, but I assume that's because a PR was merged without gulp build production somewhere else.

PR #2296, opened 2020-02-15
OS1: Update build interface script

ixv:


Fang-:
(Though the commit msg here doesn't match the required style, should probably be prefixed with "interface:".)

PR #2301, opened 2020-02-18
chat-cli: respect path structure during ;create

Fang-:
Chat paths need to be manually created with the /~ship/path structure now.

(Something something types.)

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -762,7 +762,7 @@
       :-  %chat-view-action
       !>  ^-  chat-view-action
       :*  %create
-          path
+          [(scot %p our-self) path]
           security
           ::  ensure we can read from/write to our own chats
           ::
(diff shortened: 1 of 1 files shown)


loganallenc:
This doesn’t seem to support /~/ paths.

Fang-:
@loganallenc right. Villages correspond directly to a group (that `%chat-view` creates for us), so that can have a normal path, but channels don't correspond to a group, so need to signal they're not to be used by group-based logic (with the `/~` prefix). Right? idk why `%chat-view` doesn't make that decision for us... The security model is right there in the `%create` action.

loganallenc:
The decision for whether or not a group is a managed group is not strictly tied to the %security. There are three cases.

/~zod/managed-group-case - security is village
/~/~zod/unmanaged-group-case - security is village
/~/~zod/unmanaged-group-case - security is channel

Fang-:
What's the use case for unmanaged villages?

Fang-:
Update: talked with @loganallenc out of band. The use case is ad-hoc messaging groups, of course. Chat-cli should indeed expose these _in addition to_ the more formal messaging groups. I'll have to think for a little bit about how to best expose this.

loganallenc:
Closing this, as it's superceded by your other PR...

PR #2304, opened 2020-02-18
chat: immediately send read event for own messages

### LINKED ISSUE #2182: web chat unread on my own messages
Whenever I post a chat message using web chat, it marks that channel as unread.  This might only happen if I move quickly to another channel after I post.

pkova:
Fixes #2182.

The gist of this PR is to send read receipts immediately after our own messages. We also ignore unread messages until this read receipt request completes to remove annoying blinking in the ui.

### DIFF
--- a/pkg/interface/chat/src/js/api.js
+++ b/pkg/interface/chat/src/js/api.js
@@ -77,6 +77,16 @@ class UrbitApi {
     });
   }
 
+  addReadReceiptPending(station) {
+    const m = store.state.myReadReceiptsPending.set(station, true);
+    store.setState({myReadReceiptsPending: m})
+  }
+
+  removeReadReceiptPending(station) {
+    const m = store.state.myReadReceiptsPending.set(station, false);
+    store.setState({myReadReceiptsPending: m})
+  }
+
   groupsAction(data) {
     this.action("group-store", "group-action", data);
   }
@@ -98,7 +108,7 @@ class UrbitApi {
   }
 
   chatAction(data) {
-    this.action("chat-store", "json", data);
+    return this.action("chat-store", "json", data);
   }
 
   chatMessage(path, author, when, letter) {
@@ -121,7 +131,7 @@ class UrbitApi {
   }
 
   chatRead(path, read) {
-    this.chatAction({ read: { path } });
+    return this.chatAction({ read: { path } });
   }
 
   chatViewAction(data) {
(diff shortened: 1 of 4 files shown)


pkg/interface/chat/src/js/store.js
```diff
@@ -18,6 +18,7 @@ class Store {
       spinner: false,
       sidebarShown: true,
       pendingMessages: new Map([]),
+      myReadReceiptsPending: new Map([]),
```
pkova: This is truly terrible naming I think, but couldn't come up with anything better! This is a map of station -> boolean, indicating whether we are waiting on a response to our own read receipt request.

pkg/interface/chat/src/js/api.js
```diff
@@ -98,7 +108,7 @@ class UrbitApi {
   }
 
   chatAction(data) {
-    this.action("chat-store", "json", data);
+    return this.action("chat-store", "json", data);
```
pkova: I need to access the promise to know when the read receipt request completes, so I added these return statements to `api.js`. Maybe everything in this file should be returning it?

Fang-: Agree that seems like good practice for `api.js` files.

pkg/interface/chat/src/js/api.js
```diff
@@ -77,6 +77,16 @@ class UrbitApi {
...
   }
 
+  addReadReceiptPending(station) {
+    const m = store.state.myReadReceiptsPending.set(station, true);
+    store.setState({myReadReceiptsPending: m})
+  }
+
+  removeReadReceiptPending(station) {
+    const m = store.state.myReadReceiptsPending.set(station, false);
+    store.setState({myReadReceiptsPending: m})
```
matildepark: Both of these probably shouldn't directly manipulate the state from api.js — I would make a fake `chat-update` object and throw it to `store.handleEvent` like the [sidebarToggle does](https://github.com/urbit/urbit/blob/9751bb9daadac160300ffa313be511803b348e79/pkg/interface/chat/src/js/api.js#L185). We want state manipulation to happen in one place (the reducers), basically.

(I know I wrote [the other function that has this behaviour](https://github.com/urbit/urbit/blob/9751bb9daadac160300ffa313be511803b348e79/pkg/interface/chat/src/js/api.js#L68), but I know better now. Probably, both should be changed...)

pkg/interface/chat/src/js/store.js
matildepark: I've been sitting here for a few minutes now thinking about it. `pendingReadReceipts`?

loganallenc:
This should be done in the store, not in the UI. It may already be being done. The read logic needs a once-over because it seems to have some inconsistencies.

pkova:
I'll be back.

PR #2306, opened 2020-02-18
Hoon: add repn function and jet

ixv:
This code has been cherry-picked off the os1-rc branch, because I ran into some problems when testing the OTA flow which requires us to stage things more carefully.

The problems I had were, first the %contact-view app failed to compile trying to find +repn. Then it noticed the changes to the kernel, and did a reset, but it appears ford was still hanging on to an old version of it, because after the reset it still failed to find repn.

To fix this I had to call:
```
|wipe-ford 100
|keep-ford 1 1
:goad %go
```

and then make and commit a change to %contact-view to trigger a recompile.

I was talking about this with @philipcmonk and he mentioned we may also need to make a change to ford to wipe its cache whenever it changes, to make the os1 OTA completely smooth, iirc.


joemfb:
Previously in #2066 and #2254

ixv:
I think the problem with waiting is that theres no guarantee that that successive updates pushed over the air will be applied one at a time. Having a longer time gap between when this goes out and when OS1 goes out makes it much more likely to be a smooth upgrade for the most of the network, though there is still no guarantee.

philipcmonk:
While the code change is small, this is an update with a high potential for issues, so it would be good to push it out soon so that any issues can be resolved before os1 is released.  It's a good test case for the issue @belisarius222 is working on where ford's cache doesn't get properly cleared (since apps/libs fail to rebuild here because the old kernel doesn't have +repn), which IMO is something that absolutely has to be fixed before we can push out os1.  @belisarius222, do you think this PR should wait until the ford cache issue is fixed?

While it's not ideal to have the commits duplicated, I think it's a bad idea to release stuff that's not on master, since then you have to keep track of what's released but still has to be merged.  If it was just a day or two it might be worth it, but os1 is a large release, so best not to confuse it.  If os1-rc was my private branch I would probably rebase it on master after these commits go in, but since people are collaborating on it, that's also a bad idea. The duplicate commits won't cause a real problem.

belisarius222:
This is a good question.  If the :goad commands fix this reliably, maybe we should add those commands to the OTA to get it out the door relatively safely?  I'm picturing some code in +prep in :hood or dojo that would run those commands after the OTA completes, maybe deferring the emission of the moves to a later event to prevent race conditions with Ford.

I'm not sure this is easy to do, but if it is, it might be worth it.  Open to other approaches here.

I'm still trying to figure out how Ford can avoid this problem in a principled way.  I suspect the issue here is that we don't know the order in which Clay subscribers will be notified of a file change.  This means Ford could receive file change notices before :hood hears the kernel change and causes the kernel to reload.

My best idea so far is to have it respond to changes to files in its %reef build (hoon, arvo, or zuse) by going into a mode where it enqueues build and rebuild requests.  It will stay in this mode until its +load is called, at which point it will exit that mode, clear all build state, and start chewing through the queued build requests, one per event.

Maybe this isn't quite right, though, since there's all this :goad and whatnot once the kernel does reload, which should cause Ford to cancel all of its live builds.  So maybe when the %reef changes, we just enter turtle mode and refuse to do anything until +load is called.  This is getting pretty close to just wiping the Ford state completely on reload... 

I don't think we can easily have Ford cancel its open requests when the %reef changes, since that could trigger apps to send rebuild requests, which would put us into an infinite loop of Ford rejecting build requests and apps making them again.  Then again, maybe apps should really be giving up if their build request dies a few times in a row.

Still thinking this through.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
How bad would it be to rip out the `pit` short-circuiting?  Then it would get the correct zuse from the start, right?

belisarius222:
Removing pit short-circuit would increase boot time, but shouldn't be a problem elsewhere. I'm not sure I fully understand how the short-circuit relates to this problem, though. Maybe that would in fact be enough ...

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
I don't know why the caches aren't getting properly cleared, but when it tries to rebuild apps before ford, it gets the new app files, it just doesn't get the new zuse, since `=(now date.build)`, so it uses `pit` (which hasn't been updated because ford hasn't been reloaded).  It then seems to install this into its cache, which breaks it ever after (until the next change to zuse).

Increasing boot time isn't great, but it's not that bad if it'll get in this under control.  We can optimize boot time later.

belisarius222:
Ah, I see, that conditional can get the wrong answer during update. Ripping out the pit pass-through should fix this, then. I agree, we can improve performance after getting correctness here.

—
~rovnys-ricfer
https://urbit.org

jtobin:
@ixv and @philipcmonk, re: your comments:
 
> I think the problem with waiting is that theres no guarantee that that successive updates pushed over the air will be applied one at a time. Having a longer time gap between when this goes out and when OS1 goes out makes it much more likely to be a smooth upgrade for the most of the network, though there is still no guarantee.

> While the code change is small, this is an update with a high potential for issues, so it would be good to push it out soon so that any issues can be resolved before os1 is released. 
> 
> [..]
> 
> While it's not ideal to have the commits duplicated, I think it's a bad idea to release stuff that's not on master, since then you have to keep track of what's released but still has to be merged.

I explained myself somewhat poorly.  When I said "these don't technically have to be committed to master first in order to release them first," I meant "first" in the sense of "appearing deeper in the history."  We should avoid releasing things that haven't yet been merged to master, without a doubt.

Similarly, I agree it would be nice to get these changes released sooner than later -- they probably should have been originally PR'd to master (at which point we could have released them) and then master merged into os1-rc, but I don't mean to opine on what could have been done.

In this case, I agree we should release these changes sooner than later.  But -- we can still do that when the os1-rc branch is merged to master.  My preferred ordering just would be: merge os1-rc to master, replay these changes on `v0.10` and deploy them OTA / release a new Vere version, then release other OS1 changes as required, whenever.

On that note: is there any reason not to just PR os1-rc to master now instead of cherry picking these commits over?  We don't have to wait for OS1 to be 100% ready before merging its development branch to master, because obviously it's getting merged in eventually, so why wait.

loganallenc:
What's the status on this?

pkg/arvo/lib/base64.hoon
```diff
@@ -59,7 +59,7 @@
     'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/'
   ::
   |^  |=  bs=octs  ^-  cord
-      =/  [padding=@ blocks=(list word24)]
+      =+  ^-  [padding=@ blocks=(list word24)]
```
joemfb: This looks like an accidental revert. It can be fixed without pushing new pills.

ixv: Huh, I didn't realize that works.

joemfb: See #2114

jtobin:
Closed in favour of the commits in #2365.

PR #2310, opened 2020-02-19
chat, publish: invite-search surfaces valid ship name as result

matildepark:
If we don't have any ship peers in the search results, and the search term is a valid ship name, show it as a result.

Screen Shot 2020-02-18 at 11 29 23 PM


loganallenc:
What happens when we do have the ship as a ship peer? Do we show their contact?

matildepark:
> What happens when we do have the ship as a ship peer? Do we show their contact?

They get returned in the `shipResults` array as usual. `invite-search.js` at present does not surface Contacts metadata in the results, just group paths and the names of ships that we're in groups with.

The mocks surface that in a sidenote, "aka [name1], [name2]", and I can do that as well. It seems out of scope of this PR, but I can add it to the backlog for polish.

PR #2311, opened 2020-02-19
Adds autoresize functionality to chat input

litpub:
I got tired of the tiny chat window, so I fixed it.

This makes the <textarea> auto resize. It doesn't add any dependencies, so it doesn't have _all_ of the affordances you might expect, but it works well enough for 18 lines.

Notably:
1. Debounces the resize function by 16ms so that it only happens once per frame if we're drawing at 60hz. Otherwise, the input is laggy.
2. It increases in size, but does not _decrease_ until it resets when you send a message. Having auto-decrease will probably require adding a dependency that solves this more comprehensively. On balance, I think this works well enough.
3. Restricts the max height so that it doesn't take over the whole window, just gives you a little more breathing room if you've got a lot to say.

Demo of 1 and 2:

ezgif-7-e9b789db6d3e


loganallenc:
Nice addition, well-executed.

PR #2312, opened 2020-02-19
chat: Supports translating /n into new <p> tags

litpub:
A third webchat ergonomics thing: I was annoyed that entering line breaks with shift+return in the input did not produce corresponding line breaks in the render, even though in the chat-cli, I see them.

This PR supports line breaks by splitting the text from the letter on `\n`, and producing separate `<p>` elements.

Screenshots below show this behavior on input/output in a before/after sequence.

Screen Shot 2020-02-18 at 10 38 48 PM
Screen Shot 2020-02-18 at 10 38 51 PM
Screen Shot 2020-02-18 at 10 39 30 PM
Screen Shot 2020-02-18 at 10 39 32 PM


loganallenc:
Are you getting warnings in the console about the keys not being unique for the p tags? I haven’t tested it yet, but I’d be curious whether any warnings showed up, as that sort of thing is common. I’ll test tomorrow.

litpub:
@loganallenc no because each envelope->letter is its own component, and each letter only has one `text|code|url` so the keys are always unique in their leaf of the render tree.

litpub:
Incidentally, there is a warning about a `<div>` tag inside of a `<p>` on this page already, but it's on HEAD of os1-rc, not specific to this PR

PR #2313, opened 2020-02-19
WIP: Uncursed King

benjamin-tlon:
Don't use terminfo.


benjamin-tlon:
This seems to work. Review and further testing would be helpful.

PR #2317, opened 2020-02-20
publish: add spinner, transition behaviours

matildepark:
Polish batch:

- new-post.js had `mw7` and the other views use `mw6`, so the views weren't consistent in use. Amended to all use `mw6`.
- Sidebar sorter uses the custom dropdown component. Dropdown component gets an `align` prop to flush itself left or right against its parent.
- Sidebar was too long to see all notebooks if you have a lot of notebooks. This is calc'd against the UI height now.
- header-bar moved back into skeleton a la Chat. Spinner added as an animation on the Home button.
Feb-20-2020 00-17-24
- header-bar sigil is now lined up properly with the ship name, and blends into its background regardless of theme.
- New posts, new notebooks, new comments all have proper button disabling transitions and spinner functionality.
- Adding some fixes for `-m` and `-l` viewports for sidebar and note.js navigation back to notebooks — the latter is using a `w-90-l w-90-m` shim for now...

### DIFF
--- a/pkg/arvo/app/publish/index.hoon
+++ b/pkg/arvo/app/publish/index.hoon
@@ -16,7 +16,6 @@
   ==
 ::
   ;body
-    ;div#header.w-100;
     ;div#root.w-100.h-100;
     ;script@"/~publish/index.js";
   ==
--- a/pkg/interface/publish/src/css/custom.css
+++ b/pkg/interface/publish/src/css/custom.css
@@ -200,7 +200,7 @@ a {
   transform: rotate(180deg);
   position: absolute;
   right: 4px;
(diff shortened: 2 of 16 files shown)


Fang-:
This might be the cutest spinner we've ever had.

ixv:
Checked out branch, built solid pill and ran gulp to make sure everything was up to date. Got this:
image

matildepark:
Oh, I think I know the issue — I think I forgot to add a single file to the commit. -_-; Standby!

—
~haddef-sigwen
https://urbit.org

matildepark:
Ok, try now @ixv

PR #2319, opened 2020-02-20
chat: blend uncustomised sigils

matildepark:
If the sigil has not been given a colour by its owner, blend it into the background for both themes.

(I also caught a case where we don't sig planet names if a member is in a chat but not a contact, so I added a sig in that case.)

Screen Shot 2020-02-20 at 1 35 51 AM


matildepark:
I tagged @urcades for design review because this is unspec'd, it just feels intuitive to me.

matildepark:
Yes, precisely, it emphasises that this is system information, not a personal choice — so if you use `#000000` it will show, well, black, as your choice. Everywhere I can thus far, I use the blending on system panels that use the sigil.

PR #2324, opened 2020-02-20
OS1 OTA: chat groups & permissions

Fang-:
As usual, see commit messages for deets. Primarily, this PR does two things:

- Make chat-hook depend on metadata-store for permission checks (specifically, to find the group(s) associated with the chat it wants to check permissions on).
- Add on-upgrade logic to chat-hook that will migrate "legacy" `/chat/~host/name` group (& permission) paths to the new `/~/~host/name` format. (We use the `/~` prefix here, signifying an "unmanaged" group, because these groups were created hidden from the user, and thus should remain as such.)

I've done some light testing on this, and nothing obvious breaks. That said, I want to put some more stress on this and see if it still holds up. Hence the draft status. Still, a preliminary review seems good to have here, given the nature of all this.

Thanks @loganallenc for advising on this. Please take a look!

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -598,6 +598,17 @@
 ++  chats-of-group
   |=  =group-path
   ^-  (list path)
+  ::  if metadata-store isn't running yet, we're still in the upgrade ota phase.
+  ::  we can't get chats from the metadata-store, but can make assumptions
+  ::  about group path shape, and the chat that would match it.
+  ::TODO  remove me at some point.
+  ::
+  ?.  .^(? %gu (scot %p our.bol) %metadata-store (scot %da now.bol) ~)
+    ?:  ?=([%'~' @ ^] group-path)
+      ~&  [%assuming-ported-legacy-group group-path]
+      [t.group-path]~
+    ~&  [%weird-group group-path]
+    ~
   %+  murn
     ^-  (list resource)
     =-  ~(tap in (~(gut by -) group-path ~))
@@ -616,6 +627,17 @@
 ++  groups-of-chat
   |=  chat=path
   ^-  (list group-path)
+  ::  if metadata-store isn't running yet, we're still in the upgrade ota phase.
+  ::  we can't get groups from the metadata-store, but can make assumptions
+  ::  about chat path shape, and the chat that would match it.
+  ::TODO  remove me at some point.
+  ::
+  ?.  .^(? %gu (scot %p our.bol) %metadata-store (scot %da now.bol) ~)
+    ?:  ?=([@ ^] chat)
+      ~&  [%assuming-ported-legacy-chat chat]
+      [%'~' chat]~
+    ~&  [%weird-chat chat]
+    ~
   =-  ~(tap in (~(gut by -) [%chat chat] ~))
   .^  (jug resource group-path)
     %gy
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -2,7 +2,8 @@
 ::  mirror chat data from foreign to local based on read permissions
 ::  allow sending chat messages to foreign paths based on write perms
 ::
-/-  *permission-store, *chat-hook, *invite-store
+/-  *permission-store, *chat-hook, *invite-store,
+    *metadata-store, *permission-group-hook, *group-store
 /+  *chat-json, *chat-eval, default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
@@ -142,8 +143,8 @@
     ~
   ?.  =(u.ship our.bol)
     ~
-  ::  scry permissions to check if write is permitted
-  ?.  (permitted-scry [(scot %p src.bol) path.act])
+  ::  check if write is permitted
+  ?.  (is-permitted src.bol path.act)
     ~
   =:  author.envelope.act  src.bol
       when.envelope.act  now.bol
@@ -206,8 +207,8 @@
   ^-  (list card)
   ?>  ?=(^ pax)
   ?>  (~(has by synced) pax)
-  ::  scry permissions to check if read is permitted
-  ?>  (permitted-scry [(scot %p src.bol) pax])
+  ::  check if read is permitted
+  ?>  (is-permitted src.bol pax)
   =/  box  (chat-scry pax)
   ?~  box  !!
   [%give %fact ~ %chat-update !>([%create pax])]~
@@ -224,8 +225,8 @@
   =/  pas  `path`(oust [last 1] `(list @ta)`pax)
   ?>  ?=([* ^] pas)
   ?>  (~(has by synced) pas)
-  ::  scry permissions to check if read is permitted
-  ?>  (permitted-scry [(scot %p src.bol) pas])
+  ::  check if read is permitted
+  ?>  (is-permitted src.bol pas)
   %-  zing
   :~  [%give %fact ~ %chat-update !>([%create pas])]~
       ?.  ?&(?=(^ backlog-start) (~(has by allow-history) pas))  ~
@@ -286,20 +287,25 @@
   ++  handle-permissions
     |=  [kind=?(%add %remove) pax=path who=(set ship)]
     ^-  (list card)
-    ?>  ?=([* *] pax)
-    =/  owner  (~(get by synced) pax)
+    %-  zing
+    ^-  (list (list card))
+    %+  turn
+      (chats-of-group pax)
+    |=  chat=path
+    ^-  (list card)
+    =/  owner  (~(get by synced) chat)
     ?~  owner  ~
     ?.  =(u.owner our.bol)  ~
     %-  zing
     %+  turn  ~(tap in who)
     |=  =ship
-    ?:  (permitted-scry [(scot %p ship) pax])
+    ?:  (is-permitted ship chat)
       ?:  ?|(=(kind %remove) =(ship our.bol))  ~
       ::  if ship has just been added to the permitted group,
       ::  send them an invite
-      ~[(send-invite pax ship)]
+      ~[(send-invite chat ship)]
     ::  if ship is not permitted, kick their subscription
-    [%give %kick [%mailbox pax]~ `ship]~
+    [%give %kick [%mailbox chat]~ `ship]~
   ::
   ++  send-invite
     |=  [=path =ship]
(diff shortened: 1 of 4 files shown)


Fang-:
Worth noting: the upgrade logic here depends on the metadata-store being started (it gets scried), which doesn't seem guaranteed if the drum changes happen in the same commit. Maybe this means we need a multi-phase OTA, maybe there's some way to force this to work the way we want it.

loganallenc:
first pass skim looks great, will do an in-depth review

joemfb:
@Fang-, you can scry to check if an app is started: `.^(? gu+/=dojo=)`. IDK if there's a practical fallback here if it hasn't been, you might have to set a timer and try again later.

ixv:
Note that os1-after is a temporary testing branch. You can PR into it if your just looking for a way to test this, but if this is actually ready it should go into os1-rc.

Fang-:
@joemfb that's very good to know, thanks! I think we can do a dumb fallback in that case, that just makes the same assumptions the upgrade logic in general does.

@ixv tfw the main development branch doesn't exactly match the eventual live environment.  
I went and moved this over to os1-rc, in any case.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -51,8 +54,137 @@
...
+      %~  tap  in
+      %^  scry:cc  (set path)
+        %chat-store
+      /keys
+    |^  |=  chat=path
+        ^-  (list card)
+        =/  host=ship  (slav %p (snag 0 chat))
+        =/  newp=permission  (unify-permissions chat)
+        =/  old-group=path  [%chat chat]
+        =/  new-group=path  [%'~' chat]
```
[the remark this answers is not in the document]
loganallenc: examples: /~ship/urbit-help-0afsda or /~ship/0afsda

pkg/arvo/app/chat-hook.hoon
```diff
@@ -464,19 +601,88 @@
...
+    %gy
+    (scot %p our.bol)
+    %metadata-store
+    (scot %da now.bol)
+    /resource-indices
+  ==
 ::
-++  permitted-scry
-  |=  pax=path
+::NOTE  this assumes permission paths match group paths
```
loganallenc: this is a large reason for why I'm thinking that these two stores should be combined in our next refactor

pkg/arvo/app/chat-hook.hoon
```diff
@@ -464,19 +601,88 @@
...
+++  is-permitted
+  |=  [who=ship chat=path]
   ^-  ?
-  .^(? %gx ;:(weld /=permission-store/(scot %da now.bol)/permitted pax /noun))
+  %+  lien  (groups-of-chat chat)
+  |=  =group-path
+  %^  scry  ?
+    %permission-store
```
loganallenc: single line this %^

pkg/arvo/app/chat-hook.hoon
```diff
@@ -51,8 +54,137 @@
...
+    ::
+    ++  delete-group
+      |=  [host=ship group=path]
+      ^-  card
+      ::  if we host the group, delete it directly
+      ::
+      ?:  =(our.bol host)
+        %^  make-poke  %group-store
+          %group-action
+        !>  ^-  group-action
```
loganallenc: unnecessary cast

pkg/arvo/app/chat-hook.hoon
```diff
@@ -51,8 +54,137 @@
...
+        [%unbundle group]
+      ::  else, just delete the sync in the hook
+      ::
+      %^  make-poke  %permission-hook
+        %permission-hook-action
+      !>  ^-  permission-hook-action
```
loganallenc: unnecessary cast

pkg/arvo/app/chat-hook.hoon
```diff
@@ -51,8 +54,137 @@
...
+      |=  chat=path
+      ^-  permission
+      =/  read=(unit permission)   (get-permission chat %read)
+      =/  write=(unit permission)  (get-permission chat %write)
+      ?.  &(?=(^ read) ?=(^ write))
+        ~&  [%missing-permission chat read=?=(~ read) write=?=(~ write)]
+        [%white [(slav %p (snag 0 chat)) ~ ~]]
+      ::  village: exclusive to writers
+      ::
+      ?:  &(?=(%white kind.u.read) ?=(%white kind.u.write))
```
loganallenc: 97 through 109 would be much clearer if it was a ?+ branching off of [kind.u.read kind.u.write]

pkg/arvo/app/chat-hook.hoon
```diff
@@ -51,8 +54,137 @@
...
+            %permission-hook-action
+          !>  ^-  permission-hook-action
+          [%add-owned group group]
+      ==
+    ::
+    ++  record-group
+      |=  [group=path chat=path]
+      ^-  card
+      =/  =metadata
+        ~|  [%weird-chat-path chat]
```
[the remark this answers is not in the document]
Fang-: We can theoretically crash on the `snag` and `slav` calls below. If that happens, I want to see what the unexpected path we're dealing with is.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -51,8 +54,137 @@
...
+        %*  .  *metadata
+          title         (snag 1 chat)
+          date-created  now.bol
+          creator       (slav %p (snag 0 chat))
+        ==
+      %^  make-poke  %metadata-store
+        %metadata-action
+      !>  ^-  metadata-action
```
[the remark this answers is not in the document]
Fang-: I had previously been bitten by `!>`ing without actually checking the type of the noun I was putting in, leading to crashes on the receiver side. Casting the contents of a `!>`, so you get to use the type system for sanity-checks. If the action type ever changes, we'll get a nest-fail instead of this silently breaking.

Fang-:
> but the logic looks pristine.

Hah, you're in for a ride friend. See also email comms. Chat subscriptions stay in place, which I'm pretty sure is safe, but groups/permission ones obviously don't, and I'm not sure how to treat those. Should we rely on invites to reestablish those?

loganallenc:
We should either go all or nothing. We have two options:
- Neither chat subscriptions nor group/permission subscriptions stay in place
- Both of them stay in place

I'd opt for the former, in which we send a chat invite to reestablish these subscriptions and do the simpler approach of just killing subscriptions to start with.

Fang-:
Then you have to answer the question: what happens if the host has OTA'd already, but the client hasn't yet, and the client accepts an invite?

The invite structure contains the exact path to subscribe to, but we only send one invite, with just the chat path (I think?), so how can it know to subscribe to the new group/permissions from that?

loganallenc:
I'd suggest that we send some text within the invite that denotes to the UI that it should not be displayed unless it's on the newest version, and do an OTA in the near-term that includes an update to the UI to not display those invites.

Fang-:
I'm hesitant to get into situations where people who don't get the OTA* would be left completely in the dark, so don't want hosts to kill incoming chat subscriptions on-upgrade.  
(There's apparently still a good handful of people on the network who aren't getting OTAs properly, so I'm not sure if we should _hard-depend_ on pre-OS1 updates to hold back "prematurely received" invites. If current webchat already renders invite texts, then we can just include a user-readable thing.)

Can you elaborate on what would be wrong with the middle option? (ie maintain chat subscriptions, re-establish group/permission subscriptions using the invite mechanism.) I've thought this over a bit and it seems like we can just use the invite mechanism for groups/perms, without any need to touch chat subscriptions at all.  
If it's a case of "we already have the chat so the invite does nothing", then that seems like behavior we should just fix in OS1.

(Also "groups/perms" above will have to start including metadata subscriptions, right?)

pkg/arvo/app/chat-hook.hoon
Fang-: idk about you, but I know which of these I'd rather read. I'm not convinced any of these fall into "too tall" territory.

```hoon
  |=  =group-path
  (scry ? %permission-store [%permitted (scot %p who) group-path])
 
::  vs
 
  |=  =group-path
  %^  scry  ?
    %permission-store
  [%permitted (scot %p who) group-path]
```

pkg/arvo/app/chat-hook.hoon
Fang-: Good call, this is much better. (Kinda sad the type system isn't smart enough to let you write this as a `?-`, but w/e.)

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
Fang-: Using this patterns comes with the assumption that nobody will ever directly touch resource paths ever again. I get that in the OS1 frontend that's basically always the case, but what about command line use cases?

What's the rationale for doing this again? Preventing creation of resources at the same path? Seems like something the view layer should take care of, maybe? We probably shouldn't be renaming resources more than strictly necessary, in any case, but we can apply this going forward.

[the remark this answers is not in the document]
loganallenc: As per our discussion, let's not apply this change yet, and wait until we are doing refactors in the coming months to make these types of changes.

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
loganallenc: :( I dislike it stylistically but understand your rationale and think it's reasonable. We need a shorthand for this

pkg/arvo/app/chat-hook.hoon
loganallenc: I'd rather read the above version. Doesn't matter that much, keep it.

pkg/arvo/app/chat-hook.hoon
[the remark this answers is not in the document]
loganallenc: As per our discussion, let's wait to make decisions around UIDs. I think it's likely we'll move to them, but we don't need to do it now.

Fang-:
@loganallenc Added invite sending logic as discussed.

Worth noting that this doesn't send invites for public channels. We could look at existing incoming subscriptions and parse those out, if we really wanted to?

Other than that, good to merge if latest additions look good to you.

vvisigoth:
wait wat

I thought this was going to stay parallel for a while?
—
~poldec-tonteg
http://urbit.org

loganallenc:
I'm going to work on top of this to do %metadata-hook integration and some updates / tests of %chat-view. Mark's handed it off at this point and is free to work on %links.

vvisigoth:
understood.
—
~poldec-tonteg
http://urbit.org

PR #2325, opened 2020-02-20
chat-store: immediate handle-read on our %message

### LINKED ISSUE #2182: web chat unread on my own messages

pkova:
Fixes #2182.

Same functionality as #2304, but as Logan pointed out this is better done in the store instead of the frontend.


loganallenc:
Looks great

PR #2326, opened 2020-02-20
metadata-hook: a hook for sending metadata over the network

loganallenc:
Added metadata-hook. This hook networks the metadata-store by making all resources related to a particular group-path accessible.

One can subscribe to the metadata-hook upon a group-path and receive updates any time a resource changes associated with that group-path.

I'll add better documentation before merging this, but am mostly interested in a quick review / sanity check of the functionality / logic itself.

I'll be moving into testing and integration with `chat` with my next set of PRs.

### DIFF
--- a/pkg/arvo/app/metadata-hook.hoon
+++ b/pkg/arvo/app/metadata-hook.hoon
@@ -4,7 +4,7 @@
 /+  default-agent
 |%
 +$  card  card:agent:gall
-++  versioned-state
++$  versioned-state
   $%  state-zero
   ==
 ::
@@ -107,8 +107,27 @@
 ++  watch-group
   |=  =path
   ^-  (list card)
+  |^
   ?>  (~(has by synced) path)
-  ~
+  ?>  (is-permitted [(scot %p src.bowl) path])
+  %+  turn  ~(tap by (metadata-scry path))
+  |=  [[=group-path =resource] =metadata]
+  ^-  card
+  [%give %fact ~ %metadata-update !>([%add group-path resource metadata])]
+  ::
+  ++  is-permitted
+    |=  pax=^path
+    ^-  ?
+    =.  pax
+      ;:(weld /=permission-store/(scot %da now.bowl)/permitted pax /noun)
+    .^(? %gx pax)
+  ::
+  ++  metadata-scry
+    |=  pax=^path
+    ^-  associations
+    =.  pax  ;:(weld /=metadata-store/(scot %da now.bowl)/group pax /noun)
+    (need .^((unit associations) %gx pax))
+  --
 ::
 ++  fact-metadata-update
   |=  [wir=wire fact=metadata-update]
@@ -160,12 +179,22 @@
 ++  kick
   |=  wir=wire
   ^-  (quip card _state)
-  [~ state]
+  :_  state
+  ?+  wir  !!
+      [%updates ~]
+    [%pass /updates %agent [our.bowl %metadata-store] %watch /updates]~
+  ::
+      [%group @ *]
+    ?.  (~(has by synced) t.wir)  ~
+    =/  =ship  (~(got by synced) t.wir)
+    ?:  =(ship our.bowl)
+      [%pass wir %agent [our.bowl %metadata-store] %watch wir]~
+    [%pass wir %agent [ship %metadata-hook] %watch wir]~
+  ==
 ::
 ++  watch-ack
   |=  [wir=wire saw=(unit tang)]
   ^-  (quip card _state)
-  ?~  saw  [~ state]
   ?>  ?=(^ wir)
-  [~ state(synced (~(del by synced) t.wir))]
+  [~ ?~(saw state state(synced (~(del by synced) t.wir)))]
 --
--- a/pkg/arvo/app/metadata-hook.hoon
+++ b/pkg/arvo/app/metadata-hook.hoon
@@ -1,5 +1,8 @@
 ::  metadata-hook: allow syncing foreign metadata
 ::
+::  watch paths:
+::  /group/%group-path                      all updates related to this group
+::
 /-  *metadata-store, *metadata-hook
 /+  default-agent
 |%
@@ -86,12 +89,11 @@
       %remove
     =/  ship  (~(get by synced) path.act)
     ?~  ship  [~ state]
-    ?:  &(?!(=(u.ship our.bowl)) ?!((team:title our.bowl src.bowl)))
+    ?:  &(!=(u.ship src.bowl) ?!((team:title our.bowl src.bowl)))
       [~ state]
     :_  state(synced (~(del by synced) path.act))
     %-  zing
     :~  (unsubscribe [%group path.act] u.ship)
-        ?.  &(=(u.ship our.bowl) (team:title our.bowl src.bowl))  ~
         [%give %kick ~[[%group path.act]] ~]~
     ==
   ==
@@ -108,18 +110,22 @@
   |=  =path
   ^-  (list card)
   |^
-  ?>  (~(has by synced) path)
-  ?>  (is-permitted [(scot %p src.bowl) path])
+  ?>  =(our.bowl (~(got by synced) path))
+  ?>  (is-permitted src.bowl path)
   %+  turn  ~(tap by (metadata-scry path))
   |=  [[=group-path =resource] =metadata]
   ^-  card
   [%give %fact ~ %metadata-update !>([%add group-path resource metadata])]
   ::
   ++  is-permitted
-    |=  pax=^path
+    |=  [=ship pax=^path]
     ^-  ?
     =.  pax
-      ;:(weld /=permission-store/(scot %da now.bowl)/permitted pax /noun)
+      ;:  weld
+          /=permission-store/(scot %da now.bowl)/permitted
+          [(scot %p ship) pax]
+          /noun
+      ==
     .^(? %gx pax)
   ::
   ++  metadata-scry
--- a/pkg/arvo/app/metadata-hook.hoon
+++ b/pkg/arvo/app/metadata-hook.hoon
@@ -1,0 +1,89 @@
+::  metadata-hook: allow syncing foreign metadata
+::
+/-  *metadata-store
+/+  default-agent
+|%
++$  card  card:agent:gall
+::
+++  versioned-state
+  $%  state-zero
+  ==
+::
++$  state-zero
+  $:  %0
+      synced=(map [group-path resource] ship)
+  ==
+--
+=|  state-zero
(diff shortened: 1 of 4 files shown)


pkg/arvo/app/metadata-store.hoon
```diff
@@ -171,6 +176,14 @@
   :-  [group-path [app-name app-path]]
   (~(got by associations) [group-path [app-name app-path]])
 ::
+++  metadata-for-group
+  |=  =group-path
+  %-  ~(gas by *(map [^group-path resource] metadata))
+  %+  turn  ~(tap in (~(got by group-indices) group-path))
+  |=  =resource
+  :-  [group-path resource]
```
Fang-: Why do we include the `group-path` in the result here? In every place you get this result, you already know the `group-path`, and it's constant, so including it in the map key seems unnecessary.

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+    ?>  (team:title our.bowl src.bowl)
+    =/  =path  [%group path.act]
+    ?:  (~(has by synced) path.act)  [~ state]
+    :_  state(synced (~(put by synced) path.act ship.act))
+    [%pass path %agent [ship.act %metadata-hook] %watch path]~
+  ::
+      %remove
+    =/  ship  (~(get by synced) path.act)
+    ?~  ship  [~ state]
+    ?:  &(?!(=(u.ship our.bowl)) ?!((team:title our.bowl src.bowl)))
```
Fang-: Shouldn't this first one use `src.bowl` instead of `our.bowl`?

The other actions assert permission conditions, but this one just no-ops. Probably `?>` here as well.

(Also you can just `!=(u.ship src.bowl)` etc, `!` is irregular form for `?!` negation.)

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+++  poke-hook-action
+  |=  act=metadata-hook-action
+  ^-  (quip card _state)
+  |^
+  ?-  -.act
+      %add-owned
+    ?>  (team:title our.bowl src.bowl)
+    :-  ~
+    ?:  (~(has by synced) path.act)  state
+    state(synced (~(put by synced) path.act our.bowl))
```
Fang-: It's a little bit confusing that `%add-owned` puts items in to `synced` just like `%add-synced` does... but I think that at least follows the pattern used by other existing hooks?

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+      %remove
+    =/  ship  (~(get by synced) path.act)
+    ?~  ship  [~ state]
+    ?:  &(?!(=(u.ship our.bowl)) ?!((team:title our.bowl src.bowl)))
+      [~ state]
+    :_  state(synced (~(del by synced) path.act))
+    %-  zing
+    :~  (unsubscribe [%group path.act] u.ship)
+        ?.  &(=(u.ship our.bowl) (team:title our.bowl src.bowl))  ~
+        [%give %kick ~[[%group path.act]] ~]~
```
Fang-: I don't understand why we do the `team` check here, or any of these, really. If we remove it, shouldn't we always kill all subscriptions to it?

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+      [%pass path %agent [our.bowl %metadata-store] %leave ~]~
+    [%pass path %agent [ship %metadata-hook] %leave ~]~
+  --
+::
+++  watch-group
+  |=  =path
+  ^-  (list card)
+  |^
+  ?>  (~(has by synced) path)
+  ?>  (is-permitted [(scot %p src.bowl) path])
```
Fang-: Won't you agree it makes more sense for `+is-permitted` to have a `|=  [=ship =group-path]` signature? Expecting the _caller_ to do scry path serialization doesn't seem very sane to me.

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+  ::
+  ++  give
+    |=  [=path upd=metadata-update]
+    ^-  (list card)
+    [%give %fact ~[[%group path]] %metadata-update !>(upd)]~
+  ::
+  ++  poke
+    |=  act=metadata-action
+    ^-  (list card)
+    [%pass / %agent [our.bowl %metadata-store] %poke %metadata-action !>(act)]~
```
Fang-: `+forward` and `+store` might be more appropriate here. These aren't generic "build a card" arms.

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+    ?:  =(ship our.bowl)
+      [%pass wir %agent [our.bowl %metadata-store] %watch wir]~
+    [%pass wir %agent [ship %metadata-hook] %watch wir]~
+  ==
+::
+++  watch-ack
+  |=  [wir=wire saw=(unit tang)]
+  ^-  (quip card _state)
+  ?>  ?=(^ wir)
+  [~ ?~(saw state state(synced (~(del by synced) t.wir)))]
```
Fang-: Don't we want to do the same `%kick`ing of incoming subscriptions we do in the `%remove` action case?

pkg/arvo/app/metadata-hook.hoon
loganallenc: I really like this pattern actually, as it keeps track of our "hosted resources" in the same data structure and format as our "synced resources". Ted recommended it when I was first experimenting with hooks, and it really stuck.

pkg/arvo/app/metadata-hook.hoon
loganallenc: Yes, you're right. I'll make these changes.

pkg/arvo/app/metadata-hook.hoon
loganallenc: No. We only need to kill subscriptions if we are the host of the resource.

pkg/arvo/app/metadata-hook.hoon
loganallenc: Sure, that's fine.

pkg/arvo/app/metadata-hook.hoon
loganallenc: I prefer give and poke, as they're literal. Given that they're already scoped to this particular arm, I see no issue with the shorthand names.

pkg/arvo/app/metadata-hook.hoon
loganallenc: No, because we aren't going to receive a "not-permitted" case if we are the host of the resource, and we would only kick subscriptions if we were the host of the resource.

pkg/arvo/app/metadata-store.hoon
loganallenc: We're returning an "associations" type that's specifically scoped to the group.

pkg/arvo/app/metadata-hook.hoon
joemfb: pass/give would be parallel, as would poke/fact, or update/action.

pkg/arvo/app/metadata-store.hoon
Fang-: Ohh, I see. You probably want `~(gas by *assocations)` then, for clarity.

pkg/arvo/app/metadata-hook.hoon
Fang-: The `+watch-group` seems to allow incoming subscriptions on arbitrary resources though, as long as they exist in `synced`. Perhaps we want a `?>  =(our.bowl (~(got by synced) path))` there?

pkg/arvo/app/metadata-hook.hoon
Fang-: Okay, I think I get that.

It looks like this `+watch-ack` only works for the group subscriptions, though, because of the assertion on the wire. We use `/updates` as our wire for watching the metadata-store, but use this for handling watch-acks always, so I'm a bit confused on how this doesn't insta-crash.

pkg/arvo/app/metadata-hook.hoon
```diff
@@ -0,0 +1,200 @@
...
+  ?+  wir  !!
+      [%updates ~]
+    [%pass /updates %agent [our.bowl %metadata-store] %watch /updates]~
+  ::
+      [%group @ *]
+    ?.  (~(has by synced) t.wir)  ~
+    =/  =ship  (~(got by synced) t.wir)
+    ?:  =(ship our.bowl)
+      [%pass wir %agent [our.bowl %metadata-store] %watch wir]~
+    [%pass wir %agent [ship %metadata-hook] %watch wir]~
```
Fang-: Can't you use `[ship %metadata-hook]` in both these cases? Doesn't look like there's a need for the check here.

pkg/arvo/app/metadata-hook.hoon
loganallenc: The team check can be removed, but I'll add the assert.

pkg/arvo/app/metadata-hook.hoon
loganallenc: The wir would be a cell? Even in the /updates case? [%updates ~]

pkg/arvo/app/metadata-hook.hoon
loganallenc: No, because if we're subscribing to a path on our ship, then we want to subscribe to our store.

Fang-: Durr, I hadn't seen the `-hook` `-store` difference. Disregard me!

pkg/arvo/app/metadata-hook.hoon
Fang-: Man, I shouldn't be doing review comments early in the morning.

But then, `(~(del by synced) ~)` doesn't make a lot of sense. We should definitely check for `?=([%group ^] wir)` instead, and only do the deletion logic in that case, so it's clear that that subscription flow is what this logic belongs to.

pkg/arvo/app/metadata-store.hoon
loganallenc: That doesn't work, so I'm going to keep it as is.

loganallenc: Figured it out!

loganallenc:
@Fang- This should be ready to go. Addressed all your concerns, added comments, etc.

PR #2327, opened 2020-02-20
contacts: add spinner, invite-search, and blend unknown sigils

matildepark:
- Adds current spinner and correct behaviour. Removes one redirect-on-null case that prevented the "creating a new group" flow from completing.

- Blends sigils for group members who are not yet contacts. Blends the root identity sigil, too (and removes the root identity's nickname showing as the root identity's name, too — making it feel more like system information, like the ship name in top right.) @urcades 

image

- Adds invite-search.js to Contacts for creating new groups and adding new contacts to a group.

During the work, I discovered that the component was never actually performing correct behaviour — the peer list would only have `~zod` in it when I knew other members existed in my groups. This is due to spread syntax flattening the group sets into strings, instead of adding all values in a set to the peers set. This is now fixed.

- Adds a `groupResults` boolean to the invite-search component to exclude groups from results.

- Patched chat and publish's invite-search component accordingly for the new prop and the bug fix.


matildepark:
Seems like CI failed on a 502? 
image

PR #2332, opened 2020-02-21
Experiement: Format with brittany (and stylish-haskell)?

benjamin-tlon:


benjamin-tlon:
I tried out `brittany` on some of our Haskell modules. `brittany` is basically `gofmt` for Haskell.

While, I prefer my own indentation in many cases, the output here looks quite reasonable. Why do other people think?

@jtobin 
@eglaysher 
@joemfb 
@brendanhay 
@pilfer-pandex

brendanhay:
In the hosting code I run `stylish-haskell` and that's it. None of the "premier" Haskell formatters, such as [ormolu](https://www.tweag.io/posts/2019-10-11-ormolu-first-release.html) or `brittany` produce particularly pleasant code to read, IMO, and `brittany` particularly seems to have a bunch of edge cases that end up either idiosyncratic or unformatted. I'd personally rather a "when in rome" approach to formatting. 🤷‍♂

eglaysher:
The output in king haskell seems sane, minus the handling of the non-idiomatic if/else cases in the terminal code which are a straight translation from C.

+1 to having _some_ code formatter since the gains from having `gofmt` or [`clang-fmt`](https://chromium.googlesource.com/chromium/src/+/master/docs/clang_format.md) are huge. Several areas of the chromium code base have `PRESUBMIT.py` checks which enforce that the code was formatted.

joemfb:
I prefer 2 spaces to 4, so this is an improvement on that front. (And ASCII's probably better too.)

I'm not able to evaluate the degree to which style before or after these changes is idiomatic. And I don't have any opinions about haskell formatting. But this seems to have lost some amount of intentionality in line grouping, and to attack the right margin more aggressively.

jtobin:
There are some things I like about this formatting (e.g. no unicode usage), but others seem strange (e.g. one-space indentation in areas -- I've never seen this convention anywhere).  Outside of the no-unicode restriction I don't think it really improves materially over the existing style.

No strong opinions here.  I'm also a fan of When-In-Rome, so if I need to adapt to `brittany` usage here that's fine with me.  

FWIW, I don't use anything for my stuff, `stylish-haskell` included, since none of them format Haskell code correctly (i.e., in my own preferred way). 😜

benjamin-tlon:
> In the hosting code I run `stylish-haskell` and that's it. None of the "premier" Haskell formatters, such as [ormolu](https://www.tweag.io/posts/2019-10-11-ormolu-first-release.html) or `brittany` produce particularly pleasant code to read, IMO, and `brittany` particularly seems to have a bunch of edge cases that end up either idiosyncratic or unformatted. I'd personally rather a "when in rome" approach to formatting.

I kinda agree with the "when in Rome" approach in general, but I've mostly failed to get people to adopt a consistent style in the King Haskell codebase. Using an autoformatter would resolve this issue quickly and cleanly.

I'm guessing you've come to this opinion about `brittany` after some real experience trying it? Since this PR, I've been using it to format sections of code, just to save time. So far, I don't always like how it chooses to do things, but there has yet to be any cases where the output is bad that I can't live with it. In contrast, I think `hindent` is awful.

benjamin-tlon:
> There are some things I like about this formatting (e.g. no unicode usage), but others seem strange (e.g. one-space indentation in areas -- I've never seen this convention anywhere). Outside of the no-unicode restriction I don't think it really improves materially over the existing style.
> 
> No strong opinions here. I'm also a fan of When-In-Rome, so if I need to adapt to `brittany` usage here that's fine with me.
> 
> FWIW, I don't use anything for my stuff, `stylish-haskell` included, since none of them format Haskell code correctly (i.e., in my own preferred way).

It's pretty normal to treat `where` as a half-indent. So, with two-space indentation, you get one-space indented where. This is one reason why I've always preferred 4-space indents.

benjamin-tlon:
So @jtobin @brendanhay:

Instead of thinking of this proposal as "Let's commit to using this on all of our Haskell code", think of this as: "If I use this to resolve formatting discrepancies in King Haskell, am I going to regret it? Do you have any experience trying to use these tools?"

brendanhay:
I'm fairly sure `hindent` and others aside from `brittany` have had development cease in favour of `ormolu`. I recommend reading the Tweag related blog posts on it and comparing the output to `brittany`. On a personal level I loathe some of the design decisions, but it is definitely more consistent than `brittany` - which I've run on large code bases and tended to produce some fairly irregular formatting in certain cases.

benjamin-tlon:
I'll check out `ormolu` @brendanhay. Haven't seen it before.

benjamin-tlon:
Tried out `ormolu` and output seems really awful.

benjamin-tlon:
We can revisit this conversation later, for now I'm going to soft-commit to `brittany`. Not automatically used in CI, but recommended and used to set precedent for new code in `urbit-king`.

PR #2333, opened 2020-02-21
OS1: fix publish ford OTA issues

ixv:
This PR fixes the issue described here: https://github.com/urbit/urbit/issues/2321

Of crucial importance: This needs to go on the network at the same time as the rest of the OS1 stuff. If this goes on the network as a separate step, it will not work.

cc-ing @jtobin to make sure you see this. With the recent talk about how parts of feature branches can go in at different times, I want to make sure this one does not go in at a different time.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -196,7 +196,7 @@
           =/  book=notebook-info
             [title.old '' =(%open comments.old) / /]
           =+  ^-  [grp-car=(list card) write-pax=path read-pax=path]
-            (make-groups book-name group-pax ~ %.n %.n)
+            (make-groups:main book-name group-pax ~ %.n %.n)
           =.  writers.book      write-pax
           =.  subscribers.book  read-pax
           =/  inv-car  (send-invites book-name (~(get ju old-subs) book-name))
--- a/pkg/arvo/sys/vane/ford.hoon
+++ b/pkg/arvo/sys/vane/ford.hoon
@@ -173,15 +173,11 @@
   ==  ==
 --
 |%
-::  +axle: overall ford state
+::  +axle: overall ford state, tagged by version
 ::
 +=  axle
-  $:  ::  date: date at which ford's state was updated to this data structure
-      ::
-      date=%~2018.12.13
-      ::  state: all persistent state
-      ::
-      state=ford-state
+  $%  [%~2018.12.13 state=ford-state]
+      [%~2020.2.21 state=ford-state]
   ==
 ::  +ford-state: all state that ford maintains
 ::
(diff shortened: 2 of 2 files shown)


pkg/arvo/app/publish.hoon
```diff
@@ -196,7 +196,7 @@
           =/  book=notebook-info
             [title.old '' =(%open comments.old) / /]
           =+  ^-  [grp-car=(list card) write-pax=path read-pax=path]
-            (make-groups book-name group-pax ~ %.n %.n)
```
belisarius222: What's this change about?

ixv: Another publish OTA issue that I encountered before the ford issue. I can make it its own PR if you prefer.

belisarius222: I'm ok with it being in this PR, but I don't know how to review it.  Was it just a namespace type where +make-groups really needs to be +make-groups:main, or was it something else?

ixv: Yes thats exactly it. Without `:main` the call to make-groups had a bunted bowl, which caused problems when it tried to use the current date in a particular way.

belisarius222: Ok

ixv:
Actually the more I think about this, the less I understand how it works. How is it okay to flush ford state like this?

belisarius222:
It works because of the specifics of this OTA.  :publish is the only Ford client other than Gall itself that's currently maintaining live Ford builds, which are the only part of the Ford state that strictly need to be maintained across an update.  Since :goad gets Gall to create new live builds on agents after the vanes reload, Gall will reinitialize correctly after the Ford state is wiped, and we should be  in the clear.

belisarius222:
(and because new :publish does not make live builds, so only Gall needs to do them after /sys reload)

joemfb:
We'll want to subsequently OTA an update to ford that restores state preservation on +load, right?

belisarius222:
Ford's semantics are almost entirely stateless.  It should always be safe to throw away caches, but Ford needs to know which ducts have requested which builds.  A build can be either %once or %live.

If a build is %once, Ford only maintains state for it if it blocks on a scry, so that once the scry completes, Ford can run the rest of the build and send the result back on the stored duct.  We can safely assume no %once builds are blocked when we start the OTA, since right now we don't have any builds whose scrys could block.

If a build is %live, Ford maintains state for it until the requester cancels it by sending a %kill task, or until Ford decides to send an %incomplete response.  The proper way to do this OTA would be to have :publish send a bunch of %kill's to Ford to clear its live builds.  The problem with this is that Ford has a correctness issue where sending a %kill is likely to corrupt Ford's state, which endangers the OTA.  So this hack achieves an equivalent result by completely jettisoning the old Ford state and relying on :goad to get Gall to recreate the new live builds on agents.

It's definitely a hack, but it's worth keeping in mind that because Ford's semantically meaningful state is so minimal, dropping Ford state and starting fresh is not such a crazy thing to do if we get into a bind.

belisarius222:
@joemfb I think that will depend on how quickly we can fix the Ford bug that causes us to mistrust sending %kill's.  Once we can lean on Ford to maintain its state correctly, we should definitely reinstate the real +load.  Until then, though, I think this is likely to be safer.

Hopefully we won't need to do this again, but if we need to push out an emergency OTA to fix bugs with this one, I think we should clear the Ford state again.

philipcmonk:
@joemfb is correct that this should be a one-time thing, not on every update.  In future, there could be other live builds, or there could be a pending build, and we shouldn't throw that away unless we consciously decide to do that (like now).  We can do it again if we need to clear it again.

The way to do this normally is to increment the state version number without changing the rest of the state type, then if the old state has the old version number, don't migrate it to the new state.

belisarius222:
Yeah, that sounds good.

ixv:
Oh, I didn't see Phil's comment before making that change. Should I do that instead?

belisarius222:
Hm, I guess that's more principled, yeah.  Probably a good idea.

—
~rovnys-ricfer
https://urbit.org

PR #2337, opened 2020-02-21
link-fe: truncate titles properly

matildepark:
Link items in the detail view and in the list view now properly truncate link titles to the maximum width of the page container.

The issue has to do with the implicit `min-width: auto` of flex items that allow them to overflow their parents. Since we have nested flexboxes, if one overflows *its* parent, then all its children will obey the overflow in turn. In this case, I had to attribute a `min-width` to both the link item to restrict it to the page container, and then add another `min-width` to the skeleton, which was overflowing its parent in turn.

image

### DIFF
--- a/pkg/interface/link/src/js/components/lib/link-detail-preview.js
+++ b/pkg/interface/link/src/js/components/lib/link-detail-preview.js
@@ -99,12 +99,12 @@ export class LinkPreview extends Component {
           className={"w-100 tc " + (ytMatch ? "embed-container" : "")}>
           {embed}
         </div>
-        <div className="flex flex-column ml2 pt6">
+        <div className="flex flex-column ml2 pt6 flex-auto">
           <a href={props.url} className="w-100 flex" target="_blank">
-            <p className="f8 truncate" style={{overflow: "visible"}}>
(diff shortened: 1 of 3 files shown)


pkg/interface/link/src/js/components/lib/link-detail-preview.js
```diff
@@ -99,12 +99,12 @@ export class LinkPreview extends Component {
           className={"w-100 tc " + (ytMatch ? "embed-container" : "")}>
           {embed}
         </div>
-        <div className="flex flex-column ml2 pt6">
+        <div className="flex flex-column ml2 pt6 flex-auto">
           <a href={props.url} className="w-100 flex" target="_blank">
-            <p className="f8 truncate" style={{overflow: "visible"}}>
+            <p className="f8 truncate flex-auto">
```
Fang-: The `overflow: "visible"` was there so that text didn't get cut off on the y-axis, but I guess the flex also accounts for that?

matildepark: It doesn't look cut off in the screenshot. I am pretty sure it does do the same thing, but feel free to check.

PR #2341, opened 2020-02-22
chat, contacts, link-fe, publish: black borders on focused input

matildepark:
Part of the Indigo spec. Text boxes, when focused, get a black border.

I threw it into Links' multi-line input too, but that one's more blurry as to whether it matches the specification (cc @urcades):
Feb-21-2020 20-46-01


urcades:
This is totally good I think.

PR #2344, opened 2020-02-24
Bump LMDB mapsize on non-macOS, use generic builder for urbit

jtobin:
(Update):

Per discussion in the comments, it seems fine to just `ifdef` the log limit size appropriately, so I've done that and removed the large-log-limit derivation.  The use of the generic builder for the urbit derivation is retained.

---

From #2211:

> Adds a derivation that uses a patched pkg/urbit/vere/lmdb.c that employs a sixty gigabyte log limit, instead of the default forty. This is useful for ships e.g. ~zod that receive a lot of traffic on mainnet.
> 
> Also removes the use of builder.sh in the urbit derivation, replacing it with the generic stdenv builder available to mkDerivation. installPhase captures all the important stuff that builder.sh previously did, and use of the generic builder allows stages of the build to be more easily overridden.

I wound up reverting this because it proved to mess up builds on Linux.

The problem turned out to be in the `./configure` called by the implicit `configurePhase` when switching to the generic stdenv builder.  That uses `sh` by default, IIRC, so `configure` needs to be called with `bash` instead.


benjamin-tlon:
I don't feel like I have enough context on this.

This is creating another executable that supports larger log files? We're using a patch file to have a variant which uses this larger log file?

What are the implications of this? Can you run this executable on an existing log file? Can you run the old variant on a large log file? If you can't do some of these things, what happens when you do?

How does this interact with King Haskell? Should this include a change to KH as well?

If this is a problem worth solving now, should we instead just dig in and figure out how to grow the log as needed?

benjamin-tlon:
How does this affect the release builds? They have a separate derivation IIRC.

jtobin:
> This is creating another executable that supports larger log files? We're using a patch file to have a variant which uses this larger log file?

Yes and yes.  The binary is only intended for use on ~zod on mainnet.

The motivation here is that ~zod now regularly exceeds the 40G limit in every era.  It's presently running a Vere with a 60G limit as-is -- it's just that this binary was assembled in hacky firefighting fashion (i.e. by tweaking lmdb.c and building a new binary on-the-box).  This has had to be done now several times, IIRC, so if we need special-case binary with a higher log limit going forward, it's desirable to be able to produce it mechanically.

(Come to think of it: I don't know why the 40G default limit was chosen instead of 60G, and if we can't just bump the default limit in Vere to 60G or even higher.  @joemfb?)

> What are the implications of this? Can you run this executable on an existing log file? Can you run the old variant on a large log file? If you can't do some of these things, what happens when you do?

Yeah, you can run this on a log that was originally produced by a binary with the 40G limit.  I can't comment on what happens if you try to run a >=40G log with a <40G binary, but I presume it's something along the lines of what @ixv reported previously (via email):

> Got the message: "lmdb: failed to commit event 163813629: MDB_MAP_FULL: Environment mapsize limit reached"
> 
> When I restarted it, it played through a bunch of events then printed the same message, along with: "Failed to persist event. Exiting to prevent corruption"

Continuing:

> How does this interact with King Haskell? Should this include a change to KH as well?

Potentially, but in the sense of this change, only if/when we're going to run KH on ~zod.  If there's no reason we can't just bump the default limit to 60G or higher, we should probably just do this across the board.

> If this is a problem worth solving now, should we instead just dig in and figure out how to grow the log as needed?

Possibly, but it's probably not too pressing.  At present it's for a one-off case; I don't think any ships other than ~zod tend to bump up against the 40G limit yet.

jtobin:
> How does this affect the release builds? They have a separate derivation IIRC.

They're unaffected -- this just provides a new `urbit-large-log` derivation that can be constructed on-demand via e.g. `nix-build -A urbit-large-log`.

joemfb:
IIRC, the only issues with setting a log size > 40GB have been on MacOs. Maybe we could just #ifdef the larger limit directly in the source?

/cc @eglaysher

eglaysher:
I don't see any issue with just ifdefing a larger size on OSes other than Mac.

jtobin:
Yeah that works for me.  I'll tweak this to preserve the use of the generic builder (instead of builder.sh) as that at least makes it really easy to override the derivation in the future if we need to.

joemfb:
Note that this CI failure is just timeout(s) due to #2322, and can be ignored.

PR #2345, opened 2020-02-24
Copy changes to publish and contacts

jyng:
Screen Shot 2020-02-24 at 11 01 36 AM
Screen Shot 2020-02-24 at 11 01 10 AM


pkg/interface/contacts/src/js/components/lib/contact-card.js
```diff
@@ -489,35 +489,36 @@ export class ContactCard extends Component {
     //TODO "Share card" if it's /me -> sends to /~/default of recipient
     return (
       <div className="w-100 h-100 overflow-hidden">
-        <div className="w-100 bg-white bb b--gray4">
+        <div className="flex justify-between w-100 bg-white bb b--gray4 ">
```
matildepark: ? We're centring these in the middle now?

ixv:
Please don't commit 170k lines of js blob.

pkg/interface/contacts/src/js/components/lib/contact-card.js
jyng: No, this is so that the button groups are flush left or right

jyng: It makes more sense to have the remove button separated from the primary actions.

urcades: Yeah I pointed this out in a github issue or asana task last week — it was way, way too easy to delete remove oneself/another person from a group

pkg/interface/contacts/src/js/components/lib/contact-card.js
```diff
@@ -499,38 +499,36 @@ export class ContactCard extends Component {
...
+          <div className="flex">
+            <button
+              onClick={() => {
+                if (props.share) {
+                  this.shareWithGroup();
+                } else {
+                  this.editToggle();
+                }
+              }}
+              className={`ml3 mt2 mb2 f9 pa1 ba br2 pointer b--black ` + ourOpt}>
```
matildepark: Why did you put `ba` back into these? I don't think this is OS1 in mock.

jyng: These were buttons the last time I synced with the os1-rc branch. I can remove these.

matildepark: You might want to do it like

```
git checkout os1-rc
git pull
git checkout jy/copy
git merge os1-rc
```

I don't want to have regressions in UI from this, basically...

matildepark: Though it may just be this one thing that is sticking around.

urcades: I personally adore the look of ba + br styling in Urbit, but afaik we're keeping them unstyled per @matildepark 's obervation

matildepark: Per OS1 mocks:

image

If you want to change that, I think we should probably be consistent with it, but I think this is ultimately design purview, @urcades.

urcades: No, let's not change these things, just keep them unstyled text for the time being.

matildepark:
This is approved and can be merged pending that `ba` removal from the buttons, basically.

jyng:
> This is approved and can be merged pending that `ba` removal from the buttons, basically.

Done.

PR #2349, opened 2020-02-25
Arvo and Landscape build overhauls

### LINKED ISSUE #1664: Landscape builds should be improved

### LINKED ISSUE #1710: Arvo could use a proper build

jtobin:
This is a fairly large overhaul of the build for Arvo, primarily by way of adding a derivation for Landscape.  At the same time, it makes a number of changes to the Landscape builds themselves.  It also subsumes #2138.

The background is explained in [this urbit-dev post](https://groups.google.com/a/urbit.org/d/msg/dev/bJJNVTjaLMM/dOKVL9jTCQAJ).  The TLDR from there is:

> * I've created a Nix derivation for Landscape.  'nix-build -A landscape'
>   will now create a build consisting of the compiled build artifacts
>   (i.e. JS, CSS) used in Arvo.
>
>   I've updated the various Landscape subproject gulpfiles (as well as
>   the individual package metadata) accordingly.  They no longer contain
>   tasks that work with files in the pkg/arvo directory; instead, each
>   'src' or 'dest' is always a file in the subproject itself, with the
>   subproject's 'build' directory being used for intermediate
>   reads/writes and the 'dist' directory being used for final build
>   artifacts.
>
>* I've made the Landscape derivation an input to the Arvo derivation.
>  This means that when Nix builds Arvo, it first builds Landscape, and
>   then uses the result of the Landscape builds to populate the CSS/JS in
>   app/chat, app/weather, app/soto, and so on.
>
>   I've removed these build artifacts from the Arvo subtree in git.  The
>   result is no more index.{css, js} in pkg/arvo/app/chat/{css, js}, so
>   there's no more need for people to run 'gulp bundle-prod' before
>   making a PR, and thus no more spurious merge conflicts there.
> 
>   It's worth noting that pill builds are unaffected, since the pill
>   derivations use Arvo as an input, and Arvo uses Landscape as an input.
>   This was previously an issue that prevented #1658 from being merged.
> 
> There are some other smaller changes (e.g. gulpfile syntax modernisation
> and pruning, renaming of tasks to use this_case instead of this-case),
> but they are minor.

Check the rest of that post and the commit log here for additional details.

The best way to review this, from a Landscape developer's perspective, is probably just to give it a whirl.  I encourage you to check out the branch and do the following to get a feel for the changes:

* type a plain-old `gulp` in a Landscape app directory to have it watch for file changes (or alternatively, manually use `gulp bundle_dev` or `gulp bundle_prod` after changing files -- note the underscores)
* modify some JS, CSS, etc. files as you normally would
* when you're ready to test your changes on a running development ship, use `sh/load-arvo $PIER`.  Note that `sh/load-arvo` will need `herb` installed; run `make install` to install it, if you haven't yet done so.

The only caveat that I can think of is that you'll need a sufficiently modern `gulp-cli` installation to support the updated gulp syntax.  I recommend managing your node installation with [nvm](https://github.com/nvm-sh/nvm), installing node `v12.16.1`, and then doing `npm install --global gulp-cli`.

This completely resolves #1710, and partially resolves #1664.  Longer term we'll want to pull node dependency management into Nix itself.


matildepark:
I'm guessing we will just have to merge `master` back into `os1-rc` immediately after this gets merged. I'm grateful to see this PR, solves a big problem, will test.

nix/pkgs/arvo-ropsten/builder.sh
```diff
@@ -24,3 +24,62 @@ sed --in-place \
...
+rcp $landscape/soto/index.css $out/app/soto/css/index.css
+rcp $landscape/soto/index.js $out/app/soto/js/index.js
+rcp $landscape/soto/tile.js $out/app/soto/js/tile.js
+
+# weather
+
+mkdir -p $out/app/weather/js
+
+rcp $landscape/weather/tile.js $out/app/weather/js/tile.js
+
```
matildepark: We also have `link` now in os1-rc, so when we do pull this in, we will have to adjust this file for that.

nix/pkgs/landscape/builder.sh
```diff
@@ -0,0 +1,54 @@
...
+rcp $src/soto/dist/tile.js $out/soto/tile.js
+rcp $src/soto/dist/index.css $out/soto/index.css
+
+# weather
+
+mkdir -p $out/weather
+
+rcp $src/weather/dist/tile.js $out/weather/tile.js
+
+chmod -R u+w $out
```
matildepark: Rather, all of these builders. Haha.

matildepark:
First goings at it:

`sh/build-landscape`, `make`, `nix/store/wgrwgrgw/bin/urbit -F zod -B "bin/solid.pill" -A "pkg/arvo"` boots into a blank landscape, so that clearly isn't how you do this anymore.

OK. Maybe I need to `sh/load-arvo zod`?

```
sh/load-arvo zod
+ test 1 -eq 1
+ ship=zod
+ test -d zod/.urb
+ test -f zod/.vere.lock
+ herb zod -p hood -d '+hood/autoload |'
sh/load-arvo: line 13: herb: command not found
```

Hmm.

nix/pkgs/arvo-ropsten/builder.sh
ixv: contacts as well

jtobin:
> sh/load-arvo: line 13: herb: command not found

Ah, you'll need to have `herb` installed.  `make install` should cover it.

> nix/store/wgrwgrgw/bin/urbit -F zod -B "bin/solid.pill" -A "pkg/arvo"

pkg/arvo is now Arvo *source*, not a built (i.e., with Landscape included) Arvo.  That now lives somewhere in the Nix store -- as it always has, but now more meaningfully.

The impact is that you should specify the location in the Nix store for use with `-A`, instead of pkg/arvo. Try this:

```
$ ARVO=`nix-build -A arvo --no-out-link`
$ urbit -F zod -B bin/solid.pill -A $ARVO
```

nix/pkgs/arvo-ropsten/builder.sh
jtobin: :+1: :+1: I can make any required adjustments on os1-rc when we pull these changes into it.

loganallenc:
Glad this is happening. Likely won't have time to test and review for the next few days. I trust y'alls judgment

joemfb:
It should be possible to turn this into a nix derivation and not require tools/dependencies to be installed into the $PATH, right?

jtobin:
> It should be possible to turn this into a nix derivation and not require tools/dependencies to be installed into the $PATH, right?

There are two things here that are happening externally to Nix:

* gulp-cli, npm, various node dependencies, etc.  This absolutely should all be pulled into Nix, but it's a fair bit more work, and, having never managed node stuff with Nix, I'm not sure how to best handle it yet.  I've reserved it for the next stage of this work (i.e. the remainder of #1664).  This does do a lot of the groundwork to make that easier.

* herb is just required for loading things into a running ship via `sh/load-arvo`.  This isn't actually a part of any build -- it's just a convenient development tool.

matildepark:
FTR I will get to this but it may have to wait until we're in proper QA and can afford the integration work having our build process completely overhauled would entail.

ixv:
Alright, finally got around to checking this out. A couple points.

The arvo nix store thing is kind of annoying. Putting it in the $ARVO variable is nicer than referencing the nix store itself, though its still annoying. Doing that in my .bashrc file would smooth this over, but I'm not really sure how to do that since it has to be run in the urbit directory. 

Even if I smooth it over that way, its not friendly for newcomers because knowing to do that becomes this piece of implicit tribal knowledge. I'm imagining the pain of trying to get a less technical person set up for development. I guess this is the kind of thing that dirrenv and lorri are supposed to help with, but we can't expect everyone to have a set up like that.

./sh/load-arvo worked the first time, but it seems to change the permissions of my ship's mounted desk. The second time I run it, it made me go through this:
image
and giving it y rather than n just gives me permission denied.

That's as far as I got in testing this.

jtobin:
On Fri, Feb 28, 2020 at 01:28:51PM -0800, ixv wrote:
> The arvo nix store thing is kind of annoying. Putting it in the $ARVO
> variable is nicer than referencing the nix store itself, though its
> still annoying.

I don't think it's so bad.  The environment variable is not necessary;
in practice I suppose we'd just do one-liners.  You can compare the
syntax directly:

```
urbit -F zod -B bin/solid.pill -A pkg/arvo
urbit -F zod -B bin/solid.pill -A `nix-build -A arvo --no-out-link`
```

so, 25 extra characters, and not too much more obscure.  You'd probably
want to type it in once and then CTRL+R to get it from the shell
history.

> Even if I smooth it over that way, its not friendly for newcomers
> because knowing to do that becomes this piece of implicit tribal
> knowledge. I'm imagining the pain of trying to get a less technical
> person set up for development.

I agree that requiring that kind of setup wouldn't be acceptable, yeah.
But again, I think the one-liner is fine.

> ./sh/load-arvo worked the first time, but it seems to change the
> permissions of my ship's mounted desk. The second time I run it, it
> made me go through this:
> image
> and giving it y rather than n just gives me permission denied.

That's no good.  I'll see if I can fix it..

jtobin:
Just some comments that came to my mind after looking at the comparison I made in my reply to @ixv:

```
urbit -F zod -B bin/solid.pill -A pkg/arvo
urbit -F zod -B bin/solid.pill -A `nix-build -A arvo --no-out-link`
```

Using `-A pkg/arvo` as we do at present sort of pretends that we don't have an Arvo build, in either a de-facto or technical sense.  On the de-facto side: we manually compile a bunch of JavaScript and CSS in pkg/interface, place it into pkg/arvo, and call the result 'Arvo'.  This is a build, even if we don't call it one.

On the pedantic side: Arvo literally already has a Nix build, which is why both of:

```
urbit -F zod -B bin/solid.pill -A pkg/arvo
urbit -F zod -B bin/solid.pill -A `nix-build -A arvo --no-out-link`
```

will do the exact same thing on master right now.  The changes in this PR 1) stop us from pretending that we *don't* have a build, and 2) reconcile the de-facto Arvo build with the actual one.

(This technique is already useful for getting a testnet config on master.  There is never a need to alter pkg/arvo in order to tweak testnet settings, for example -- one can always boot a ship in a testnet configuration using the `arvo-ropsten` derivation, via:

```
urbit -F zod -B bin/solid.pill -A `nix-build -A arvo-ropsten --no-out-link`
```

I've mentioned this somewhere before, but I don't think it's well-disseminated.)

jtobin:
> ./sh/load-arvo worked the first time, but it seems to change the permissions of my ship's mounted desk. The second time I run it, it made me go through this: and giving it y rather than n just gives me permission denied.

Made a tweak to `sh/load-arvo` in 0429af3 that should take care of this (not wiping `home` on disk, swapping `rsync` for `cp`).  Nice catch.

ixv:
Last fix works for me.

I'm ambivalent about the ergonomics of this. If we're using this oneliner which needs to be run in the urbit directory, we might as well make some nicer scripts for booting. That doesn't need to block this PR though.

I'm personally fine with this going in, but I'd like to see another person to review this before it does.

jtobin:
> If we're using this oneliner which needs to be run in the urbit directory, we might as well make some nicer scripts for booting.

Good point.  We could add a simple `boot-dev` script that just takes a ship (and maybe optionally a git revision) as its argument and just calls:

```
urbit -F $SHIP -B bin/solid.pill -A `nix-build -A arvo --no-out-link`
```

then booting a development ship becomes e.g. `sh/boot-dev zod`, and loading changes becomes `sh/load-arvo zod`?

loganallenc:
@jtobin Can you rebase this pull request on master and separate it into one that simply moves all the files from interface to landscape, another that makes the new nix build scripts, and another that makes the modifications to gulp? This has gotten very far behind master at this point, and we just now have time to think about reviewing this and what these changes entail.

loganallenc:
Closing this for now.

PR #2350, opened 2020-02-25
"extra commit"

frodwith:
The long-awaited extra commit I owe. Also removes +mush.


joemfb:
Links #2285 for posterity.

joemfb:
BTW, the CI failure is spurious. We just need to make the cache push optional for PRs from forked repos.

PR #2353, opened 2020-02-25
publish: add 'edit note' functionality

matildepark:
- Add a wildcard to the endpoint mount for notes so we can handle params
- Removes an errant console.log
- Create an "Edit Post" component with a similar presentation as New Post
- Routes it into Note view and conditionally checks if the user is in the write group before presenting the 'edit' link

There were some inline styles that duplicated indigo-static classes in new-post, I believe, so I swapped in those classes too.

## Additional notes
Editing a note and then viewing the note returns the note but with a new line appended to the start and end. I can't see why this would originate in front-end: the state and value changes don't have this new line. 

I can only think this happens as part of the processing of the note after the API submission. Do you know why this does this?

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -1225,7 +1225,7 @@
       (~(get by subs) who.act book.act)
     ?~  book
       ~|("nonexistent notebook: {<book.act>}" !!)
-    =/  not=(unit note)  (~(get by notes.u.book) note.act) 
+    =/  not=(unit note)  (~(get by notes.u.book) note.act)
     ?~  not
       ~|("nonexistent note: {<note.act>}" !!)
     =?  tile-num  !read.u.not
@@ -1613,7 +1613,7 @@
     (manx-response:gen (index u.book-json))
   ::
   ::  single note, with initial 50 comments, wrapped in html
-      [[~ [%'~publish' %note @ @ @ ~]] ~]
+      [[~ [%'~publish' %note @ @ @ *]] ~]
     =/  host=(unit @p)  (slaw %p i.t.t.site.url)
     ?~  host
       not-found:gen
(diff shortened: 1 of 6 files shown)


matildepark:
Do you know what's up with the new line thing? The issue is that they accumulate, haha. Edited notes just get more and more new lines after being submitted...

ixv:
Oh yes, we add a new line when submitting a new post or editing a post.

https://github.com/urbit/urbit/blob/os1-rc/pkg/arvo/app/publish.hoon#L1032
https://github.com/urbit/urbit/blob/os1-rc/pkg/arvo/app/publish.hoon#L1106

This is an artifact of when we used udon, which would crash if it didn't have a trailing newline. It can be removed now that we're not doing udon anymore.

ixv:
Didn't test the case where the original poster is not the person editing. I see now that we don't prevent that, maybe we should?

matildepark:
Why? It makes sense to me that giving someone write access means giving someone write access. If you want to edit a note someone else started, you’d expect this would be what write access would do.

ixv:
Okay, I don't feel strongly about it, just checking that it's intentional behavior.

matildepark:
@urcades ? @vvisigoth ? Any objections? The obvious one is that your edit of the post won’t be live updated if it changes from under you. We can try and catch it and merge it into our prompt maybe? Dunno though.

matildepark:
@ixv I still get `%unterminated-line` when I remove those lines. Are you sure this is no longer udon?

ixv:
Ah, you're right, it's a more complicated than just removing those lines. I have a fix which involves reworking the way we do frontmatter parsing, which will go in with my publish OTA stuff.

matildepark:
OK. Then the edit change is done and this can just be merged.

PR #2357, opened 2020-02-26
links, publish: blend member sigils

matildepark:
- If a user is a member, but not a contact (no metadata shared), their sigil is unstyled and blended as in Chat and Contacts. This is now incorporated into Publish and Links.
- Links had one too many `flex-auto`s and thus while long titles were truncated properly, short titles pushed their hostname too far to the edge. This is amended and it all seems to work properly now for both cases.


matildepark:
If they’re in our groups but not in our contacts, they’re a member! Member and contact are the two words I’ve seen so far but it’s very ad hoc.

PR #2358, opened 2020-02-26
Visual QA: Publish comments-component, vertical padding fix

urcades:
Previously our comments component was spaced `mt9` from the post body, and was situated too close to the bottom of the viewport and the comments themselves, when populated.

This value has been edited to `mv9`, to give it some breathing room.

### DIFF
--- a/pkg/interface/publish/src/js/components/lib/comments.js
+++ b/pkg/interface/publish/src/js/components/lib/comments.js
@@ -58,7 +58,7 @@ export class Comments extends Component {
 
     return (
       <div>
-        <div className="mt8">
+        <div className="mv8">
           <div>
             <textarea style={{resize:'vertical'}}
               ref={(el) => {this.textArea = el}}


urcades:
Simple PR — was considering bundling more "Visual QA" changes into it, but unsure if we have a particular style/approach to how we map PRs to individual Asana tasks.

urcades:
cc @matildepark re: the above — should I try to avoid making PRs with such tiny change footprints (a single react styling class)?

matildepark:
@urcades If you were to ask @Fang- he'd say the smaller and more atomic the PR, the better. I think that's generally sound, just use judgment as to moderating the bandwidth you expect of others.

vvisigoth:
Is OK to have tiny PR. It's not a competition ;p
—
~poldec-tonteg
http://urbit.org

urcades:
This is ready to merge pending stamp from @matildepark or @jyng

matildepark:
I honestly feel like this could be closed since its changes are in #2364 and we're reviewing that as part of this.

urcades:
@matildepark good point — I'll do just that.

PR #2360, opened 2020-02-26
invite search: fuzzy search ships by nicknames, show ship nicknames

matildepark:
Submitting this PR for feedback and then I'll port the changes to all apps, so don't merge immediately.

This is an implementation that creates a map of all our contacts and every nickname we have for them; on search, we filter the array of nicknames and if we have any partial match, throw their ship into the search results.

Screen Shot 2020-02-26 at 5 31 30 PM

Separately, in the render side we append all nicknames we know of. Both nicknames and long ship names get truncated.

Screen Shot 2020-02-26 at 5 31 50 PM

In this implementation the performance will worsen the more contacts we have. At that point, we will probably want to debounce the search input.

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/invite-search.js
+++ b/pkg/interface/chat/src/js/components/lib/invite-search.js
@@ -93,7 +93,6 @@ export class InviteSearch extends Component {
           return e.toLowerCase().includes(searchTerm);
         });
         if (match.length > 0) {
-          console.log("found")
           if (!(contact in shipMatches)) {
             shipMatches.push(contact);
           }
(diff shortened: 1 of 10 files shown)


pkg/interface/chat/src/js/components/lib/invite-search.js
```diff
@@ -33,18 +34,32 @@ export class InviteSearch extends Component {
...
       }
+      if (this.props.contacts[group]) {
+        let groupEntries = this.props.groups[group].values();
+        for (let member of groupEntries) {
+          if (this.props.contacts[group][member]) {
+            if (contacts.has(member)) {
+              contacts.get(member).push(this.props.contacts[group][member].nickname);
+            }
+            else {
+              contacts.set(member, [this.props.contacts[group][member].nickname]);
```
loganallenc: Looks great, the performance characteristics of this are fine. People aren't going to have 100s or 1000s of contacts for a while.

PR #2361, opened 2020-02-27
OS1: publish ota-ability fixes

ixv:
This PR fixes various issues encountered in the process of upgrading publish from the old version.

In order:
- fixes a problem where it was making a %sing clay request rather than a %next, which required a second change to the file state before it detected the initial change
- set the tile num for invites and unread notes properly upon upgrading
- No longer leave notebooks on upgrade, since the other side will %kick you anyway and send invites for you to resubscribe. Gettting the update first and leaving too soon would mean you don't get the invite.
- Ignore subscription updates on the old /collection/* subscription path, just prevents uneccesary stack traces in some cases, functionally its the same.
- Stop reading in the date-created field at file change time, rather than just reading it from the file's frontmatter, which caused post dates to be reset to the time of upgrade.

In the process of doing an OTA there is still a crash in the case that ship has a notebook already, but it is a spurious one, caused by the fact that app changes are processed before vane changes when you commit. When ford is loaded, and goad recompiles all the apps, the crash does not happen again.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -105,7 +105,7 @@
     ::      send invites to all previously subscribed ships
     ::
     |^
-    =/  rav  [%sing %t [%da now.bol] /app/publish/notebooks]
+    =/  rav  [%next %t [%da now.bol] /app/publish/notebooks]
     =/  tile-json
       (frond:enjs:format %notifications (numb:enjs:format 0))
     =/  init-cards=(list card)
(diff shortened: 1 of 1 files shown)


loganallenc:
What happens if the other side continues sending you updates because they haven't received their OTA yet?

ixv:
We just no-op. The old publish subscribed on the /collection/* wire, now we subscribe on /subscribe/*. See: 8616bf0

PR #2366, opened 2020-02-27
arvo: refactors internals, adds error-handling

joemfb:
This PR refactors arvo's internal engines, and adds new mechanisms by which vanes can propagate error notifications.

The primary changes are:

- `|wink` refactored into `|me` and `|va`
- `|is` refactored into `|le`
- top-level structures are modernized, grouped, and documented
- new-style error notifications are added, in addition to the current style

The new engines -- namespaced in `|part` -- are inspired to varying degrees by the unreleased, incomplete `neo/arvo` (deleted in 7d4b35c86bdb54db5b13ed49e3027ac7eedb8e6f). These changes were made to support the error-notification implementation, and generally improve the quality of arvo's internals. No changes have been made to arvo's external interface or persistent structures, so this extensive refactoring is still suitable for OTA release without staging or adaptation.

The new error notifications involve changes to arvo's internal loop, and the interface between arvo and its vanes:

- `$goof` is added, a new error-notification structure (similar to `$ares` from the deep past)
- `%hurl` is added, a new kernel action to propagate a `$goof` along with a `%pass` or `%give`
- `(unit goof)` is added to the sample of vanes' `+call` and `+take` arms

Inside the vanes, new error notifications are "downcast" to the old style (ie, `%crud` everywhere but %ames, where they become `%crud` or `%hole`). The only change in the error-handling behavior of the vanes is in %behn's `%drip` handling. Errors therein are now propagated to the intended recipients (arriving in their `+take` arm, where they'll be merely printed).

Finally, the worker process is updated to send new error-notification events, including both the `$goof` (ie, bail mote and stack trace) and the original event.

This PR represents an incomplete but viable snapshot of the error-handling work. Additional changes are needed in the vanes (to handle more error notifications, or more fully) and the runtime, specifically the IPC protocol and I/O drivers (to precisely handle errors in error notifications). Additional improvements to arvo also follow from this work, most notably around upgrade. All such changes will be more disruptive than these, and harder to handle without strict versioning coordination between arvo and the runtime. These changes are a foundation upon which incrementally better error-handling can be built for the live network, while larger efforts continue in the background.

Calling the error-notification structure `$goof` is somewhat ... well, goofy. Some other candidates include fail, ruin, crud, flaw, lack, and miss. Feedback is requested.

The changes to arvo and the vanes must be released together, but no intermediate staging is needed, and the runtime changes need not be correlated. Since almost every commit in this PR merits a new pill, but none individually require one, I've departed from the recommended approach and saved the pill update for the end.


pkg/arvo/sys/arvo.hoon
```diff
@@ -21,65 +21,156 @@
...
+      ::    NB:  task: a forward action
+      ::         sign: a sourced $gift
+      ::
+      [note gift]
+  $%  ::  %hurl: action failed
+      ::  %pass: advance
+      ::  %slip: lateral
+      ::  %give: retreat
+      ::
+      [%hurl =goof wite=$;($>(?(%pass %give) $))]
```
ixv: wite here refers to the %pass or %give that failed?

joemfb: Right, the idea is to send it again, along with the error. See the `%drip` handling in %behn for an example.

I'm not sure what that mold has to be as complicated is, but obvious simplifications don't work (`wite=$>(?(%pass %give) $)` or `wite=$>(?(%pass %give) (wite note gift))`). Something's probably a little off in the `$spec` macro expansion, and the new explicit recursion specs aren't wired up in the parser.

pkg/arvo/sys/arvo.hoon
```diff
@@ -21,65 +21,156 @@
...
+      [%tas p=@tas]
+      [%ud p=@ud]
+  ==
++$  cage  (cask vase)
+++  cask  |$  [a]  (pair mark a)
++$  desk  @tas
++$  dock  (pair @p term)
++$  mark  @tas
++$  ship  @p
+++  sink  (trel bone ship path)
```
belisarius222: sink should be +$ too

pkg/arvo/sys/arvo.hoon
```diff
@@ -21,65 +21,156 @@
...
+::
++|  %global
+::
+::  $arch: fundamental node
+::  $beak: global context
+::  $beam: global name
+::  $bone: opaque duct handle
+::  $case: global version
+::  $cage: global metadata
+::  +cask: global data builder
```
belisarius222: I think 'global' is misleading for these, and 'metadata' too.
I propose:
```
$case: revision reference
$cage: data with mark and type
$cask: marked data builder
```

pkg/arvo/sys/arvo.hoon
```diff
@@ -21,65 +21,156 @@
...
++$  curd  (cask)
++$  duct  (list wire)
+++  hobo
+  |$  [a]
+  $?  $%  [%soft p=*]
+      ==
+      a
+  ==
++$  goof  [mote=term =tang]
++$  monk  (each ship (pair @tas @ta))
```
belisarius222: Maybe it's out of scope, but while we're messing with Arvo, it would be nice to remove this (incomplete) definition of a person's name from the kernel.  I don't know how hard it would be to switch scrys to use `(unit (set ship))` instead of `(unit (set monk))`, but it might be pretty quick.

pkg/arvo/sys/arvo.hoon
```diff
@@ -21,65 +21,156 @@
...
++|  %implementation
+::
+::  +mash: producing mass
+::  $mass: memory usage
+::  $pane: kernel modules
+::  $pone: kernel modules old
+::  $vane: kernel module
+::  $vile: reflexive constants
+::
+++  mash  |=(* (mass +<))
```
belisarius222: mash seems unused

joemfb: It's used for the recursion in `$mass`, but it seems unnecessary. As does the explicit bunt. Probably artifacts of an old mold refactoring. I'll clean this up.

pkg/arvo/sys/arvo.hoon
joemfb: It's definitely bad, but I'd prefer to handle the whole scry-sample at once. FWIW, the neo scry-sample is just `[(unit (set ship)) case=term =beam]`.

pkg/arvo/sys/arvo.hoon
joemfb: I think the "global" in `$case` goes well with `$beak` and `$beam`, which seem correct. "Revision" feels more specific than "version", and that could be misleading.

But "global metadata" is definitely a stretch though, as is "global" as regards marks. How about even more normcore:

```
::  $cage: marked vase
::  +cask: marked data builder
```

pkg/arvo/sys/arvo.hoon
```diff
@@ -156,6 +247,553 @@
...
+    ++  refine-move
+      |=  vax=vase
+      ^-  (pair move worm)
+      ~>  %mean.'bad-move'
+      =^  hip  sac  (~(nell wa sac) p.vax)
+      ?.  hip
+        ~>(%mean.'not a cell' !!)
+      =/  duc
+        ~>  %mean.'bad-duct'
+        ::  XX used to be a nest-check and clam, seemed excessive
```
belisarius222: clamming this duct is fine; don't need the comment

pkg/arvo/sys/arvo.hoon
```diff
@@ -21,65 +21,154 @@
...
+      ::
+      [note gift]
+  $%  ::  %hurl: action failed
+      ::  %pass: advance
+      ::  %slip: lateral
+      ::  %give: retreat
+      ::
+      [%hurl =goof wite=$;($>(?(%pass %give) $))]
+      [%pass =wire =note]
+      [%slip =note]
```
belisarius222: I see you decided to leave %slip in there.  Do we still use it anywhere?  Also, what happens if a %slip fails, since it can't get wrapped in a %hurl?

pkg/arvo/sys/arvo.hoon
```diff
@@ -156,6 +245,549 @@
...
+      ?>  hip
+      =^  typ  sac  (~(slot wa sac) 2 tiv)
+      =.  sac  (~(neat wa sac) typ.vil [%& typ])
+      ::
+      ::  support for meta-meta-cards has been removed
+      ::
+      ?>  ?=(meta q.tiv)
+      [[%| q.tiv] sac]
+      ::
+      :: =/  mut
```
belisarius222: I don't like this commented-out code.  We can always dig this out of git if we need to include it.

pkg/arvo/sys/arvo.hoon
```diff
@@ -156,6 +245,549 @@
...
+            q.bed
+            `coin`[%$ r.bed]
+            (flop s.bed)
+        ==
+      ::
+      ::  |spin:plow:va: move statefully
+      ::
+      ++  spin
+        |=  [hen=duct eny=@uvJ dud=(unit goof)]
+        :: =/  duc  !>(hen)
```
belisarius222: I think this comment should either be removed or have context added.  I'm pretty sure I know what it means (that [duc.vil hen] is similar to !>(hen)), but I don't find it clarifying.

pkg/arvo/sys/arvo.hoon
```diff
@@ -156,6 +245,549 @@
...
+          %pass
+        =*  wire  wire.ball.move
+        =*  duct  duct.move
+        =*  vane  vane.note.ball.move
+        =*  task  task.note.ball.move
+        ::
+        ~?  &(!lac !=(%$ vane.gem))
+          :-  (runt [(dec depth.gem) '|'] "")
+          :^  %pass  [vane.gem vane]
+            ?:  ?=(?(%deal %deal-gall) +>-.task)
```
belisarius222: Why is %deal-gall special-cased here?  If it's because the move contains a vase, we might want to check for Ford builds here too.

pkg/arvo/sys/arvo.hoon
joemfb: We `%slip %init` to all the vanes during the "legacy boot" event. And depend on the resulting duct(s), IIRC. We're not gonna recover from a failure in that event. Sane vane bootstrap is definitely on my list, but it could get messy (and might not be backwards compatible), so I'm saving it for later.

pkg/arvo/sys/arvo.hoon
joemfb: I minor hedge on removing meta-meta cards; I wanted to show what it would be post-refactoring. Of course, meta-meta card reduction has been broken since a335974008d795ee5b9cb941b7f2404fe1069aeb, so it's not very risky to remove it. I'll delete the comment.

pkg/arvo/sys/arvo.hoon
joemfb: Hedging against the `$vile` cached reflexives. I don't think they're actually necessary, but I need to work through any implications wrt type-of-type migrations first. I'll remove the comment (and replace `=*  err  !>(dud)` with a cons of a type from `$vile`).

pkg/arvo/sys/arvo.hoon
joemfb: This is just ported straight from `+jack:is`, preserving the current behavior of `|verb`, which  has been documented, and is expected by some (small number of) people. I think `|verb` could be significantly improved, and also think we should probably have short and long printing routines in %zuse for all vanes' tasks/gifts.

joemfb:
Requested changes are made, conflicts resolved, pills updated. PTAL.

Some final things to note:

There are three ways that these changes will affect users and developers:

- Error messages in stack traces will be somewhat different

I don't think this is a big deal, developers can add ~| to arvo if necessary.

- The vane-specialized and jet-hinted activation wrappers have been removed

Likewise, vanes can be jet-hinted if more details are needed in profile output.

- Error reporting and control via `|knob` may be different

Most notably, `%hole` cards will never sent from unix with an updated binary. The typical path for an unprocessable %ames packet will be `%hear` -> `%crud` (internally converted to %hole) -> %warn -- assuming crashes in the first two events. %warn prints a single line in the kernel, and can't be turned on or off. The full traces will still be captured in the log, but those are obviously not easily available.

`|knob` and error-printing in general will need refinement in this overall error-handling effort. I think the state of affairs as of these changes is fine, but it may be a little disrupting to users.

Finally, I'm leaning towards replacing `$goof` with `$flaw`. The specific use-case for now will be bail-mote and stack-trace, but the general spirit of the structure is an error described by a "symbolic category" or "very short description" and a full, printable description. I expect over time that we'll use traces less, and explicitly-constructed error categories and descriptions more. Feedback appreciated.

belisarius222:
Hm, so does this mean users will need to run |knob %hear %hush to shush Ames packet dropping?  People do run that command a lot, so it's important for them to at least be able to run something to turn that stuff off.  I wouldn't be opposed to adding a shim to |knob to enable people to keep running |knob %hole %hush for this purpose, if we need to.

I prefer %goof to %flaw, since I think of "flaw" as an inherent problem, whereas a "goof" could be a one-time failure.  I'm not terribly opposed to replacing [mote=term =tang] with some kind of $% at some point, but I don't think it's worth it for now.

It might not ever be worth it; it's quite ergonomic to have the same error type everywhere.  Compare, for example, Rust code that has to transform error types in a bunch of places.  Hoon doesn't have typeclass-like automatic coercion either, so I suspect we're better off sticking with $tang — unless we start doing a bunch of programmatic dispatch based on the kind of error, but I don't think we have much of that right now.

—
~rovnys-ricfer
https://urbit.org

joemfb:
No, they wouldn't `|knob %hear`, there will be no full report for them to hush. The current replacement flow is %hear -> %hole -> %crud -> warn, and it's the third event that is being controlled by `|knob` in this scenario. The flow in this PR is %hear -> %crud -> %warn, with %warn being a temporary stand-in for a %bail IPC message that punts the problem to the I/O driver.

As for goof/flaw, I'm not suggesting that the structure change, or that the error types be reified in a $%. I'm just pointing out the broad usability of that structure, and how I expect it's use to change over time. Goof seems over-specialized to a description of a transient mistake, like a crash.

joemfb:
Post-merge, this is hitting the gateway timeout on boot, not in the tests ...

joemfb:
I wonder if it's possible to keep the pit short-circuit in %ford, but only for the initial builds during boot.

belisarius222:
I'm not sure how to do that in a way that doesn't make later bugs very likely.  For example, if we hard-coded Ford to check whether the Clay revision was less than 2, or whatever the criterion would be there to check if we're "booting" -- then I wouldn't be surprised if there were issues with pills or some other way that these things could get out of sync.  Maybe that's an unfounded worry, but I'm not sure.

Maybe we could have Jael send Ford an %init, and Ford would copy its .pit into some %reef caches at whatever Clay revision we're at at that time?  We'd have to be sure Ford was compiled using the same Clay revision as the one Clay was at, I think, in order to prevent getting the wrong answer.

ohAitch:
Baking all boot compile caches clay blob store hashing etc into the pill
seems desirable

Like, in an ideal world there's no reason for booting a fresh |moon to take
multiple seconds, at least on localhost? All the information it needs
should exist on the parent, and most of that is just "here's the nock and
++type everything compiles to at the current version"; there's correctness
concerns wrt hoon self-compilation actually converging on the actual
version of the interpreter you end up on, but in a multihome hosting
environment they should be possible to short-circuit after the first time.
including for booting fresh dev ships, or non-final-release ci

</vaporware>

joemfb:
@ohAitch, I strongly agree, but I'm not entirely sure how to do some of it. And getting close will probably require significant changes to the bootstrap procedures. Already merited, of course.

joemfb:
@belisarius222, `%init` wouldn't work right now, as it happens in the "legacy boot" event, before the initial userspace commit. I don't know %ford's internals too well, maybe there's an easy way to check for the first %reef build. Hardcoding cases 1/2 might be fine. But the right way to do the pit short-circuit is for arvo to persist the source it's running, so %ford can scry and compare. I aim to add that minimal arvo filesystem soon(tm).

belisarius222:
Yeah, comparing source (or receiving source with pit) is the right way to solve this.  Should we just do that?  Working on that is a better use of time than this Ford hack, although it's not good to let this PR languish either.

If not, I could add a hack to Ford to track whether we've built %reef on this desk before, and if not, use the .pit.  I think it whould be relatively straightforward.

—
~rovnys-ricfer
https://urbit.org

joemfb:
@belisarius222, I don't want to churn interfaces for short-term workarounds. And I don't want to add state adapters here, or disrupt the outer layers of arvo.

I've added back the pit-shortcircuit for %home and %base (which does get built first, you were right). This makes boot fast again. The only risk is making a bad pill (kernel source mismatch), which is already a risk. It should just be avoided. This seems fine for now, IMO, unless you'd prefer to prime the cache to the same effect. I just don't want to be blocked on this problem anymore.

belisarius222:
Ok, no this looks fine.  Glad it worked.

joemfb:
@jtobin, thanks for checking. Was this a new fake-ship from the arvo tagged release? 

I've pulled the %ford changes into #2384, and will reintegrate here once that's done. Pushing these changes OTA should work (and is why this milestone has been PR'd), but it still needs to be tested. As for `[%poke %bad-wire ...]` certain implicit constraints have been made explicit, and it looks like we're violating them in some cases. IMO, the best thing would be to track those cases down and either fix or special-case (ie, adapt) them.

joemfb:
@jtobin, the `[%poke %bad-wire ...]` error was caused by the `%vega` reset notification using an invalid wire. I've fixed this, and added code to dynamically fix-up the `%vega` that will come from the old kernel on `|reset`. The `find` errors on `|reload` are most likely due to the kernel `|reset` failing -- `|reload` doesn't recompile the kernel, so it was attempting to load new vanes into an old kernel.

That being said, I'm not quite ready for this to be merged. I'd like to redo the master merge once  #2384 is in, and clean things up generally. And I want to do a review myself.

Also worth noting, this does not need to be in or block the os1 release. Due to the issues we've been having with OTA's, it might be best to hold it until after os1. But we can discuss that separately.

joemfb:
Ok, this is cleaned up and ready on my end. @belisarius222, please take a final look.

I've tested committing these changes to a fake-zod booted from the `v0.10` branch. When #2247 is included, the OTA fails in %clay and %gall (`%load-lost`), due to type-of-type changes (The old type nests in the new, but the need to mutually nest for vanes that store vases.) When #2247 is excluded the upgrade applies cleanly.

@jtobin, I haven't tried merging this and then cherry-picking the merge commit; I'm not sure if that's sufficient to exclude the changes in question.

jtobin:
On Wed, Mar 04, 2020 at 11:11:48AM -0800, Joe Bryan wrote:
> @jtobin, I haven't tried merging this and then cherry-picking the
> merge commit; I'm not sure if that's sufficient to exclude the changes
> in question.

No sweat, I'll check it out before merging.

joemfb:
I did one last re-merge (there was a legitimate, minor conflict in %ford as of os1-rc).

jtobin:
The cherry-picked merge doesn't include the %spot hint changes, but it *does* include the %ford changes that can't yet go out OTA (per #2333 -- presumably this popped up in the conflict you encountered).  

I think it's ok to merge this, as I don't plan to release any more non-surgical-hotfix updates prior to the OS1 release.  This can probably just go out with OS1 proper.

jtobin:
Alternatively, I can probably resolve the %ford conflict that occurs in the cherry-pick to take just the change relevant to this PR, and see if I can get it into the last release prior to OS1.

I'll give it a quick test just to check -- if anything looks dodgy I'll just hold off.

jtobin:
> I'll give it a quick test just to check -- if anything looks dodgy I'll just hold off.

This appears fine.  Will merge and push a new Arvo release candidate.

PR #2367, opened 2020-02-27
contacts: add group details page, metadata-store reducer

matildepark:
To re-target `master` once #2365 is merged. Until then, this is just for review!

image

- Quick display fix on the 'delete' button which was too close to the edge.
- Quick typo fix in metadata-store that led to `/all` subscriptions failing.
- Wrote a metadata reducer for the initial load and for `add` events; there are likely delete and change events but I cannot produce one!

The metadata data structure I wrote for the front-end is a map, keyed by group path, whose value is an object with keys of the channel name and a value of the full object. Like so:

image

This makes it simple to display and programmatically link to.

## Additional notes

We pass all these "Open" links to the `/join` path of the respective application, because each application has a handler that redirects to a channel if we're already subscribed, or initiates a subscription to the channel in question if not, all seamlessly. (Links doesn't have this because it doesn't have a join flow yet, **but it will**...)

The metadata store will thus (theoretically) show you every channel the group has, even ones you're not in yet, and you can join them seamlessly. Pretty cool!

@loganallenc warned me that if we haven't started a chat, the subscription will fail. I cannot replicate this: I get an empty `associations` object. Maybe he ran into the typo bug too. It all runs smoothly.

The reason that the titles in the screenshot are the paths of the channels is that we fall back to the app-path name if we don't have a title for the channel.

### DIFF
--- a/pkg/interface/contacts/src/js/components/lib/group-detail.js
+++ b/pkg/interface/contacts/src/js/components/lib/group-detail.js
@@ -1,0 +1,57 @@
+import React, { Component } from 'react';
+import { uxToHex } from '/lib/util.js';
+
+export class GroupDetail extends Component {
+  render() {
+    const { props } = this;
+
+    let responsiveClass =
+      props.activeDrawer === "details" ? "db" : "dn db-ns";
+
+    let groupPath = props.path || "";
+
+    let groupChannels = props.channels.get(groupPath) || {};
+
+
+    let channelList = <div/>
+
+    channelList = Object.keys(groupChannels).map((channel) => {
+      let channelObj = groupChannels[channel];
+      if (!channelObj) {
+        return false;
+      }
+      let title = channelObj.metadata.title || channelObj["app-path"] || "";
+      let color = uxToHex(channelObj.metadata.color) || "000000";
+      let app = channelObj["app-name"] || "Unknown";
+      let channelPath = channelObj["app-path"];
+      let link = `/~${app}/join${channelPath}`
+      app = app.charAt(0).toUpperCase() + app.slice(1)
+      return(
+        <li
+        key={channel}
+        className="f9 list flex pv2 w-100">
+        <div className="dib"
+          style={{backgroundColor: `#${color}`, height: 32, width: 32}}
+        ></div>
+        <div className="flex flex-column flex-auto">
+          <p className="f9 inter ml2 w-100">{title}</p>
+          <p className="f9 inter ml2 w-100"
+          style={{marginTop: "0.35rem"}}>
+            <span className="f9 di mr2 inter">{app}</span>
+            <a className="f9 di green2" href={link}>Open</a>
+          </p>
+          </div>
+        </li>
+      )
+    })
+    return (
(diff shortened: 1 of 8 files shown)


pkg/interface/contacts/src/js/components/lib/group-detail.js
```diff
@@ -0,0 +1,64 @@
...
+
+
+    let channelList = <div/>
+
+    channelList = Object.keys(groupChannels).map((channel) => {
+      let channelObj = groupChannels[channel];
+      if (!channelObj) {
+        return false;
+      }
+      let title = channelObj.metadata.title || channelObj["app-path"] || "";
```
loganallenc: Perfect

pkg/interface/contacts/src/js/subscription.js
```diff
@@ -26,6 +26,10 @@ export class Subscription {
       this.handleEvent.bind(this),
       this.handleError.bind(this),
       this.handleQuitAndResubscribe.bind(this));
+    api.bind('/all', 'PUT', api.authTokens.ship, 'metadata-store',
+    this.handleEvent.bind(this),
```
loganallenc: Indent these

pkg/interface/contacts/src/js/store.js
```diff
@@ -11,6 +12,7 @@ class Store {
     this.state = {
       contacts: {},
       groups: {},
+      channels: new Map,
```
loganallenc: Is there a reason not to just call this associations?

matildepark: At first I had it as `metadata`, but realised that didn't explain itself to a dev trying to read this code later and went with "channels" because that's what this is ultimately for. I can see why the preferred tact could be just to keep the same terminology between front-end and back-end though, I can amend.

matildepark:
I'll address style and polish the case where there are no channels so new groups can be introduced better.

matildepark:
Caught the style and also added an empty explainer:

image

vvisigoth:
Navigating to "Share Contact" on contact view crashes app

Contacts

with error

```
Uncaught TypeError: Cannot read property 'pathname' of undefined
    at Zh.render (index.js:1)
    at Ov (index.js:1)
    at Lv (index.js:1)
    at Gv (index.js:1)
    at HTMLUnknownElement.e (index.js:1)
    at Object.y (index.js:1)
    at N (index.js:1)
    at hg (index.js:1)
    at Ky (index.js:1)
    at Yy (index.js:1)
index.js:1 The above error occurred in the <Zh> component:
    in Zh (at root.js:174)
    in div (at skeleton.js:26)
    in div (at skeleton.js:16)
    in div (at skeleton.js:14)
    in Rh (at root.js:165)
    in t (at root.js:151)
    in div (at root.js:41)
    in t (created by t)
    in t (at root.js:40)
    in L (at src/index.js:15)
```

matildepark:
Lol I know why, ok one sec.

loganallenc:
@matildepark can I go ahead and merge this?

vvisigoth:
everything looks nice!

PR #2369, opened 2020-02-28
Unmanaged Chats Work Properly and Support Metadata Syncing

loganallenc:
This should be ready to go. It's been tested heavily.

Joining unmanaged chats works, receiving invites to unmanaged chats works, title and descriptions work, metadata syncing works, etc


pkg/interface/chat/src/js/components/new.js
```diff
@@ -220,19 +228,29 @@ export class NewScreen extends Component {
...
           <textarea
             className={idClasses}
-            placeholder="secret-chat"
+            placeholder="The coolest chat"
             rows={1}
             style={{
               resize: "none"
             }}
-            onChange={this.idChange}
+            onChange={this.titleChange}
```
matildepark: Oh, I found one. This should be descriptionChange, ha.

pkg/interface/chat/src/js/components/new.js
```diff
@@ -38,9 +41,18 @@ export class NewScreen extends Component {
     }
   }
 
-  idChange(event) {
+  titleChange(event) {
+    let asciiSafe = event.target.value
+      .replace(/[^a-z0-9~_.-]/g, "-");
```
matildepark: This replaces capitals with `-`

loganallenc: fine by me for a first pass. if it's easy to make a better one, i'd love to see it. i'm not the best at regex

pkg/interface/chat/src/js/components/new.js
loganallenc: good catch, thanks!

pkg/interface/chat/src/js/components/new.js
matildepark: you could just stick a `toLowerCase()` on it before the regex then so we preserve the letters in the path name. This is the path name, right? Not the title itself?

loganallenc: Yes, path name must be ascii safe. That’s a good idea

PR #2372, opened 2020-02-28
ford: remove ship and desk from %reef +rain path

belisarius222:
This should allow %reef builds on %base and %home desks to share cache.

### DIFF
--- a/pkg/arvo/sys/vane/ford.hoon
+++ b/pkg/arvo/sys/vane/ford.hoon
@@ -4658,29 +4658,22 @@
       ::
       ?.  ?=([~ %success %scry *] zuse-scry-result)
         (wrap-error zuse-scry-result)
-      ::  omit case from path to prevent cache misses
       ::
-      =/  hoon-path=path
-        /(scot %p ship.disc)/(scot %tas desk.disc)/hoon/hoon/sys
       =/  hoon-hoon=(each hoon tang)
         %-  mule  |.
-        (rain hoon-path ;;(@t q.q.cage.u.hoon-scry-result))
+        (rain /hoon/hoon/sys ;;(@t q.q.cage.u.hoon-scry-result))
       ?:  ?=(%| -.hoon-hoon)
         (return-error leaf+"ford: %reef failed to compile hoon" p.hoon-hoon)
       ::
-      =/  arvo-path=path
-        /(scot %p ship.disc)/(scot %tas desk.disc)/hoon/arvo/sys
       =/  arvo-hoon=(each hoon tang)
         %-  mule  |.
-        (rain arvo-path ;;(@t q.q.cage.u.arvo-scry-result))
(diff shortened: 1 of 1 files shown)


ohAitch:
arguably the desk should be a runtime "biblical" constant; removing it
entirely will cause much confusion looking through hoon or zuse stack
traces when your source is a different version, though to an extent this
already happened with case zeroing

perhaps it's time for clay to grow a `@uv/absolute/file/path/hoon`
interface, keyed off the hash of the file? (keyed off hash of the desk
would be neater in some sense but much less compatible with this sort of
caching; honestly the path could be entirely presentational and ignored,
like blog post or amazon product "readable" url slugs)

belisarius222:
I doubt we'll lose important debugging information from this change.

@liam-fitzgerald's work to put file locations in types suggests we at least want some kind of programmatically readable path in there, so that an editor can jump-to-definition.  I suspect as desks get more self-contained, being reminded which desk something happened in will also become less useful.  Arguably, for kernel code the pair of spur (or maybe even just a $term like %zuse, %arvo, or %hoon) and Kelvin version should be an unambiguous file reference.

joemfb:
@belisarius222, I don't see how this would help with slow boot due to #2322, as I don't think we build anything but %home during boot. It might help with `|mount`'ing desks other than %home.

I think we want the paths in the stack traces. As long as it's possible to run different versions of a piece of code, it's a help to know where the source came from. Even in the kernel, as |autoload or the equivalent could be set on different desks.

We could have remove the path entirely from %spot, and have a separate hint that sets the path for subsequent %spots. Or maybe a biblical would work, I'm not entirely sure.

Adding %spot to type is mostly orthogonal -- it's behind a flag, and probably shouldn't be on in general. And those types won't be available to traces.

belisarius222:
Right, it wouldn't help with boot.  It would help with later %reef builds across desks, though, which currently can't share any caching at all because the desk information is in there.

In over two years of working on Urbit, I'm pretty sure that not once have I made use of the ship or desk in a stack trace for debugging purposes.  Have you?  Maybe @ixv has on live ships.  Also keep in mind this is just for %reef builds.  Builds of other files still have the full beam (with zeroed-out case) as their +rain paths.

Knowing that an error happened on a particular line in hoon vs. zuse is certainly useful, but knowing it happened on this ship (which it always did) or in some particular desk ... doesn't seem anywhere near as useful as being able to share kernel build results across desks.

This "biblical" idea is possibly compelling, although I'm not sure anyone knows how to implement that, so I wouldn't count on it coming into existence anytime soon.

Being able to load kernel source and pit from Arvo itself should fix this issue, so I could just drop it in favor of that.

—
~rovnys-ricfer
https://urbit.org

loganallenc:
Status on this?

belisarius222:
dead in the water

PR #2373, opened 2020-02-28
soto: swap out tile images

urcades:
The previous Tile.png for the Dojo module (Soto) was a stylized rendering of "barcen" and could be seen as confusing. Fixing up the image to appear more *actually* glyph-like.


urcades:
Just looking for a stamp of approval from @matildepark or @vvisigoth , whoever gets to it first!

PR #2375, opened 2020-02-29
hoon: +side and +sill vase splitters

belisarius222:
`+side` is the inverse of `+slop`, and `+sill` splits a vase of a list-shaped noun into a list of vases, which I suspect Arvo might want to use for decoding moves and Clay might want to use for destructuring Ford build results.

Both arms preserve type information and specialize for the shape that was discovered.  If the shape doesn't match, they crash.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -12083,6 +12083,21 @@
   ^-  vase
   [[%cell p.hed p.tal] [q.hed q.tal]]
 ::
+++  side                                                ::  split vase to cell
+  |=  vax=vase
+  ^-  [hed=vase tal=vase]
+  ?@  q.vax  !!
+  =.  p.vax  (~(fuse ut p.vax) [%cell %noun %noun])
+  [(slot 2 vax) (slot 3 vax)]
+::
+++  sill                                                ::  split vase to list
+  |=  vax=vase
+  ^-  (list vase)
+  ?~  q.vax
+    ~
+  =/  [hed=vase tal=vase]  (side vax)
+  [hed $(vax tal)]
+::
 ++  slot                                                ::  got axis in vase
   |=  {axe/@ vax/vase}  ^-  vase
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -12083,6 +12083,21 @@
   ^-  vase
   [[%cell p.hed p.tal] [q.hed q.tal]]
 ::
+++  side                                                ::  split vase to cell
+  |=  vax=vase
+  ^-  [hed=vase tal=vase]
+  ?@  q.vax  !!
```
joemfb: I think this is better as an assertion: `?>  ?=(^ q.vax)` or `?<  ?=(@ q.vax)`.

pkg/arvo/sys/hoon.hoon
```diff
@@ -12083,6 +12083,21 @@
...
+  =.  p.vax  (~(fuse ut p.vax) [%cell %noun %noun])
+  [(slot 2 vax) (slot 3 vax)]
+::
+++  sill                                                ::  split vase to list
+  |=  vax=vase
+  ^-  (list vase)
+  ?~  q.vax
```
joemfb: This is really just `?:  =(0 q.vax)`. I wonder if we should be checking that the type is `~` as well:

```
?>  (~(nest ut [%atom %n `0]) & p.vax)
```

But I'm not sure how much it matters...

loganallenc:
Status?

belisarius222:
Deprioritized.  Could close if this is cluttering up the /pulls page.

—
~rovnys-ricfer
https://urbit.org

tacryt-socryp:
Any thoughts on bringing this home or closing it?

tacryt-socryp:
Ping.

belisarius222:
I think I should remove +sill and then get this in.

tacryt-socryp:
Ping.

belisarius222:
Non-urgent.  Closing.

PR #2377, opened 2020-02-29
group-hook: sync metadata for synced groups

Fang-:
In the "new" model, metadata is tightly linked with group usage. But the
metadata for any given group wasn't being synced to its members.

This adds logic to the group-hook that, for local groups, exposes metadata when
the group gets exposed, and for foreign groups, syncs in metadata when the group
itself gets synced in.

For local groups, we also set up permissions, to ensure metadata is available. @loganallenc there's no need to sync in permissions for remote groups, right?

### DIFF
--- a/pkg/arvo/app/group-hook.hoon
+++ b/pkg/arvo/app/group-hook.hoon
@@ -1,6 +1,6 @@
 ::  group-hook: allow syncing group data from foreign paths to local paths
 ::
-/-  *group-store, *group-hook
+/-  *group-store, *group-hook, *metadata-hook, *permission-group-hook
 /+  default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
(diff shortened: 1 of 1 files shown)


loganallenc:
I’ve wanted to (and am planning for tomorrow) try to achieve similar functionality by writing an addition to the contact-view that syncs down metadata for managed groups upon the %join action. What are the tradeoffs between these two approaches?

loganallenc:
For now, I’d like to avoid the permission-group-hook coupling, at the least. That seems premature... The group-hook / metadata-hook coupling also seems unnecessary? I’d prefer to be able to interact with these components independently, but maybe have library functions that send moves to both of them for common application lifecycles.

Fang-:
Closing as per out-of-band conversation with @loganallenc. tldr: a contact-view approach is more consistent/sane on the short-term. Going forward, some of these things are going to need to be refactored/unified. Details pending.

PR #2370, opened 2020-02-28
os1: dark mode pass

matildepark:
Screen Shot 2020-02-28 at 1 25 10 AM

- Incorporates dark theme into Contacts and Publish. 
- Also adds some changes to invite search component for blending across modes and filtering out empty nickname strings.
- Incorporates invite search into the 'members' component of Chat too, invites seem to work properly so all good.

PR #2378, opened 2020-03-01
link: metadata integration

Fang-:
Chunky commits this time, because of big, intertwining changes. In brief:
- make link hooks use metadata-store for resource discovery & syncing
- add aggregate action to link-view for creating new collection, with maybe a new (unmanaged) group + invites
- update frontend to use metadata for resource discovery and detail rendering, invites

6c19b05 is more or less the same thing as #2368, except for groups instead of apps.

Depends on whatever @loganallenc's implementation of #2377 ends up being. Doesn't sync links properly without that.

Putting this out here for preliminary review. Remaining work:
- [ ] Support creating real groups from the new collection screen. Checkbox currently looks weird and does nothing, "unmanaged" only.
- [ ] Member management screen for collections?
- [ ] Minor UI fixes
- [ ] Support `/~link/join/resource` route as linked to be contacts frontend
- [ ] Currently always syncs in all of a group's collections. Maybe we want explicit opt-in, even if you're already in the group?

### DIFF
--- a/pkg/arvo/app/link-listen-hook.hoon
+++ b/pkg/arvo/app/link-listen-hook.hoon
@@ -1,21 +1,33 @@
 ::  link-listen-hook: get your friends' bookmarks
 ::
-::    on-init, subscribes to all groups on this ship. for every ship in a group,
-::    we subscribe to their link's local-pages and annotations
-::    at the group path (through link-proxy-hook),
-::    and forwards all entries into our link as submissions and comments.
+::    subscribes to all %link resources in the metadata-store.
+::    for all all ships in groups associated with those resources, we subscribe
+::    to their link's local-pages and annotations at the resource path (through
+::    link-proxy-hook), and forward all entries into our link-store as
+::    submissions and comments.
 ::
-::    if a subscription to a group member fails, we assume it's because their
-::    group definition hasn't been updated to include us yet.
+::    if a subscription to a target fails, we assume it's because their
+::    metadata+groups definition hasn't been updated to include us yet.
 ::    we retry with exponential backoff, maxing out at one hour timeouts.
+::    to expede this process, we prod other potential listeners when we add
+::    them to our metadata+groups definition.
 ::
-/-  *link, group-store
-/+  default-agent, verb, dbug
+/-  *metadata-store, *link, group-store
+/+  metadata, default-agent, verb, dbug
 ::
 |%
 +$  state-0
   $:  %0
       retry-timers=(map target @dr)
+      ::  reasoning: the resources we're subscribed to,
+      ::             and the groups that cause that.
+      ::
+      ::    we don't strictly need to track this in state, but doing so heavily
+      ::    simplifies logic and reduces the amount of big scries we do.
+      ::    this also gives us the option to check & restore subscriptions,
+      ::    should we ever need that.
+      ::
+      reasoning=(jug [ship app-path] group-path)
   ==
 ::
 +$  what-target  ?(%local-pages %annotations)
@@ -52,7 +64,7 @@
   ++  on-init
     ^-  (quip card _this)
     :_  this
-    [watch-groups:do]~
+    ~[watch-metadata:do watch-groups:do]
   ::
   ++  on-save  !>(state)
   ++  on-load
@@ -63,26 +75,27 @@
   ++  on-agent
     |=  [=wire =sign:agent:gall]
     ^-  (quip card _this)
-    ?:  ?=([%groups ~] wire)
-      =^  cards  state
+    =^  cards  state
+      ?+  wire  ~|([dap.bowl %weird-agent-wire wire] !!)
+          [%metadata ~]
+        (take-metadata-sign:do sign)
+      ::
+          [%groups ~]
         (take-groups-sign:do sign)
-      [cards this]
-    ?:  ?=([%links ?(%local-pages %annotations) @ ^] wire)
-      =^  cards  state
+      ::
+          [%links ?(%local-pages %annotations) @ ^]
         (take-link-sign:do (wire-to-target t.wire) sign)
-      [cards this]
-    ?:  ?=([%forward ^] wire)
-      =^  cards  state
+      ::
+          [%forward ^]
         (take-forward-sign:do t.wire sign)
-      [cards this]
-    ?:  ?=([%prod *] wire)
-      ~|  [%weird-sign -.sign]
-      ?>  ?=(%poke-ack -.sign)
-      ?~  p.sign  [~ this]
-      %-  (slog [leaf+"failed to prod" u.p.sign])
-      [~ this]
-    ~|  [dap.bowl %weird-wire wire]
-    !!
+      ::
+          [%prod *]
+        ?>  ?=(%poke-ack -.sign)
+        ?~  p.sign  [~ state]
+        %-  (slog leaf+"prod failed" u.p.sign)
+        [~ state]
+      ==
+    [cards this]
   ::
   ++  on-poke
     |=  [=mark =vase]
@@ -122,8 +135,65 @@
 ::
 ::
 |_  =bowl:gall
++*  md  ~(. metadata bowl)
 ::
-::  groups subscription
+::  metadata subscription
+::
+++  watch-metadata
+  ^-  card
+  [%pass /metadata %agent [our.bowl %metadata-store] %watch /app-name/link]
+::
+++  take-metadata-sign
+  |=  =sign:agent:gall
+  ^-  (quip card _state)
+  ?-  -.sign
+    %poke-ack     ~|([dap.bowl %unexpected-poke-ack /metadata] !!)
+    %kick         [[watch-metadata]~ state]
+  ::
+      %watch-ack
+    ?~  p.sign  [~ state]
+    =/  =tank
+      :-  %leaf
+      "{(trip dap.bowl)} failed subscribe to metadata store. very wrong!"
+    %-  (slog tank u.p.sign)
+    [~ state]
+  ::
+      %fact
+    =*  mark  p.cage.sign
+    =*  vase  q.cage.sign
+    ?.  ?=(%metadata-update mark)
+      ~|  [dap.bowl %unexpected-mark mark]
+      !!
+    %-  handle-metadata-update
+    !<(metadata-update vase)
+  ==
+::
+++  handle-metadata-update
+  |=  upd=metadata-update
+  ^-  (quip card _state)
+  ?+  -.upd  [~ state]
+      %associations
+    =/  socs=(list [=group-path resource])
+      ~(tap in ~(key by associations.upd))
+    =|  cards=(list card)
+    |-  ::TODO  try for +roll maybe?
+    ?~  socs  [cards state]
+    =^  caz  state
+      =,  i.socs
+      ?.  =(%link app-name)  [~ state]
+      (listen-to-group app-path group-path)
+    $(socs t.socs, cards (weld cards caz))
+  ::
+      %add
+    ?>  =(%link app-name.resource.upd)
+    (listen-to-group app-path.resource.upd group-path.upd)
+  ::
+      %remove
+    ?>  =(%link app-name.resource.upd)
+    (leave-from-group app-path.resource.upd group-path.upd)
+  ==
+::
+::  groups subscriptions
 ::
 ++  watch-groups
   ^-  card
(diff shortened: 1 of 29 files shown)


pkg/interface/link/src/js/components/lib/invite-search.js
```diff
@@ -0,0 +1,369 @@
...
+          <span className="db gray2">Participants</span>
+          {groups} {ships}
+        </div>
+      );
+    }
+
+    return (
+      <div className="relative">
+        <img
+          src="/~chat/img/search.png"
```
matildepark: You're going to want to copy in a search icon into the `pkg/arvo/link/img` folder, no cross-pollination of assets...

Also maybe copy a new version of this file in, some patches to it got put into os1-rc with #2370 I think.

pkg/interface/link/src/js/components/new.js
```diff
@@ -0,0 +1,252 @@
...
+      //TODO account for state.createGroup
+      let submit = api.createCollection(
+        appPath,
+        state.title,
+        state.description,
+        target
+      );
+      submit.then(() => {
+        api.setSpinner(false);
+        console.log('TODO nav', `/~link/list/0${appPath}`);
```
matildepark: `props.history.push` is what you want. passing `{...props}` to new will provide that method.

pkg/interface/link/src/js/components/root.js
```diff
@@ -93,44 +122,47 @@ export class Root extends Component {
...
+                  resource={resource}
                   popout={popout}
                   sidebarShown={state.sidebarShown}
                   />
                 </Skeleton>
               )
             }}
           />
-          <Route exact path="/~link/(popout)?/:ship/:channel/:page/:index/:encodedUrl/(comments)?/:commentpage?"
+          <Route exact path="/~link/(popout)?/item/:page/:index/:commentpage/:encodedUrl/:resource([^#]+)"
```
matildepark: Er, I'm actually now confused as to what resource correlates to; if it's a collection path that the item belongs to, this is definitely out of order.

pkg/arvo/app/link-listen-hook.hoon
```diff
@@ -1,21 +1,33 @@
 ::  link-listen-hook: get your friends' bookmarks
 ::
-::    on-init, subscribes to all groups on this ship. for every ship in a group,
-::    we subscribe to their link's local-pages and annotations
-::    at the group path (through link-proxy-hook),
-::    and forwards all entries into our link as submissions and comments.
+::    subscribes to all %link resources in the metadata-store.
+::    for all all ships in groups associated with those resources, we subscribe
```
loganallenc: all all

pkg/interface/link/src/js/components/lib/channels-item.js
```diff
@@ -18,7 +18,7 @@ export class ChannelsItem extends Component {
       : null;
 
     return (
-      <Link to={"/~link" + props.link}>
+      <Link to={"/~link/list/0" + props.link}>
```
loganallenc: Weird path format?

pkg/interface/link/src/js/components/new.js
```diff
@@ -0,0 +1,252 @@
...
+
+    let idClasses =
+      "f7 ba b--gray3 b--gray2-d bg-gray0-d white-d pa3 db w-100 " +
+      "focus-b--black focus-b--white-d ";
+
+    let idErrElem = (<span />);
+    if (state.idError) {
+      idErrElem = (
+        <span className="f9 inter red2 db pt2">
+          Chat must have a valid name.
```
loganallenc: Chat

pkg/interface/link/src/js/components/new.js
```diff
@@ -0,0 +1,252 @@
...
+      submit.then(() => {
+        api.setSpinner(false);
+        console.log('TODO nav', `/~link/list/0${appPath}`);
+      })
+    });
+  }
+
+  render() {
+    const { props, state } = this;
+    let inviteSwitchClasses = (state.security === "village")
```
loganallenc: Does links have blacklist security support?

pkg/interface/link/src/js/components/new.js
```diff
@@ -0,0 +1,252 @@
...
+      if (station in props.resources) {
+        console.log('TODO nav', '/~chat/room' + station);
+      }
+    }
+  }
+
+  titleChange(event) {
+    let asciiSafe = event.target.value.toLowerCase()
+      .replace(/[^a-z0-9-]/g, "-");
+    this.setState({
```
loganallenc: Random number appended. Love it!

pkg/interface/link/src/js/components/lib/invite-search.js
Fang-: Ah, yeah. I copied this over from chat, should've checked for such things more thoroughly.

pkg/interface/link/src/js/components/new.js
Fang-: Link fe doesn't use `history` though? Not sure how to pull that in properly, but I can try mirroring what chat does, assuming that doesn't mean reworking routing entirely.

pkg/interface/link/src/js/components/root.js
Fang-: Yes, it's a collection path. But it's a collection _path_. You can't put that in the middle of another path and parse it out sanely. So if you ever have multi-part paths there (ie `/webdev/web1`, you kind of need it to be at the end.

cc @loganallenc (re "weird path format")

pkg/interface/link/src/js/components/new.js
Fang-: As per discussion on asana, whitelist only. (So yes, this should be cleaned up. The whole new.js should, in any case.)

pkg/interface/link/src/js/components/new.js
matildepark: From Link-fe:
image

It comes with `react-router`. :)

pkg/interface/link/src/js/components/root.js
loganallenc: Seems strange to me too btw. I don’t know the data structures of links well enough to understand what each item corresponds to, but this is... certainly unconventional.

matildepark: > But it's a collection path. You can't put that in the middle of another path and parse it out sanely.

This is why Chat, Notebooks, etc don't say `:path` params but `:ship/:name` and why I am starting to think we should disallow slashes in names — I think the most sane, consistent pattern we can do across all apps is dictate static identifiers that follow `host/group/channel` nesting. I'm a bit unsettled that we are deliberating this now.

vvisigoth: I'm all for reducing the degrees of freedom wherever possible here; especially where there is precedent with chat and notebooks. FWIW the multi-channel per group functionality will cover what (ostensibly) the use case of multi-segment path collection IDs.

Let's make sure we don't leave standup today without putting this to bed.

Fang-: As per standup: it isn't that important but I'll go do the consistent thing and make frontend-side assumptions about what resource paths look like.

ohAitch:
Arvo solves nested paths in some places by @ta coin encoding (iirc / to _
and _ to ~- and ~ to ~~), but this might be obnoxious on the frontend

vvisigoth:
@Fang- is this ready for review?

vvisigoth:
@Fang- LMK when this is ready for review?

Fang-:
@vvisigoth bonking my head against the last couple issues (settings screen flow), should be all ready by the time you wake tomorrow.

vvisigoth:
Sounds good! Thanks.
—
~poldec-tonteg
http://urbit.org

Fang-:
Alright, ready for re-review. I think I hit all of the flows, nothing about this is horribly broken afaict.

image
idk what's up with the toggle button on the "new" page, I'm using the same code that chat does, and they show up fine there.

matildepark:
Custom css ~~~

pkg/interface/link/src/js/components/settings.js
```diff
@@ -0,0 +1,301 @@
...
+        </div>
+      );
+    }
+
+    return (
+      <div className="h-100 w-100 overflow-x-hidden flex flex-column white-d">
+        <div
+          className="w-100 dn-m dn-l dn-xl inter pt4 pb6 pl3 f8"
+          style={{ height: "1rem" }}>
+          <Link to="/~chat/">{"⟵ All Chats"}</Link>
```
matildepark: Link back to ~link, All Collections.

pkg/interface/link/src/js/components/settings.js
```diff
@@ -0,0 +1,301 @@
...
+          <SidebarSwitcher
+            sidebarShown={this.props.sidebarShown}
+            popout={this.props.popout}
+          />
+          <Link to={makeRoutePath(props.resourcePath, props.popout)}
+          className="pt2">
+            <h2
+              className="mono dib f9 fw4 v-top"
+              style={{ width: "max-content" }}>
+              {props.resource.title}
```
matildepark: One comment for all four views: using `mono` on user-dictated name.

image

So, is it cool and good to put the user's dictated name in the header? I don't know, maybe? But we don't do that in Chat, we put the endpoint in the header. Either: 
1. this has to change to Inter and we add user dictated names to the header in Chat or
2. This displays the endpoint name to match Chat

@urcades makes the call.

Fang-: ftr, displaying the endpoint name will give you `/it-s-a-collection-4897` or similar.

pkg/interface/link/src/js/components/lib/member-element.js
```diff
@@ -0,0 +1,52 @@
...
+    api.groups.remove(props.groupPath, [`~${props.ship}`]);
+  }
+
+  render() {
+    const { props } = this;
+
+    let actionElem;
+    if (props.ship === props.owner) {
+      actionElem = (
+        <p className="w-20 dib list-ship black white-d f8 c-default">
```
matildepark: I think this is missing the `list-ship` custom class from Chat's `custom.css`

matildepark: That is, adding those declarations in Link's custom.css, ha.

pkg/interface/link/src/js/components/new.js
```diff
@@ -0,0 +1,239 @@
...
+      })
+    });
+  }
+
+  render() {
+    const { props, state } = this;
+
+    let createGroupClasses = state.createGroup
+      ? "relative checked bg-green2 br3 h1 toggle v-mid z-0"
+      : "relative bg-gray4 bg-gray1-d br3 h1 toggle v-mid z-0";
```
matildepark: I think we want `checked` and `toggle` classes from Chat's `custom.css` added to Link's `custom.css` too.

pkg/interface/link/src/js/components/settings.js
```diff
@@ -0,0 +1,301 @@
...
+    const isManaged = ('/~/' !== props.groupPath.slice(0,3));
+
+    let deleteButtonClasses = (props.amOwner) ? 'b--red2 red2 pointer bg-gray0-d' : 'b--grey3 grey3 bg-gray0-d c-default';
+    let leaveButtonClasses = (!props.amOwner) ? "pointer" : "c-default";
+
+    let deleteClasses = 'dib f9 black gray4-d bg-gray0-d ba pa2 b--black b--gray0-d pointer';
+    let deleteText = 'Remove this collection from your collection list.';
+    let deleteAction = 'Remove';
+    if (props.amOwner && isManaged) {
+      deleteText = 'Delete this collection from the group. This deletes it for everyone!';
```
matildepark: Probably just use the phrasing from Chat's settings. Exclamation marks! are not ideal.

Same for below copy too...

pkg/interface/link/src/js/components/settings.js
urcades: Option 1 is preferred.

Very aware we want to retain the context for where a chat "lives" with respect to its host ship, so we'll need to think about how to obviously surface this info at some point, whether the endpoint is displayed onHover or below the name in a newly structured header or otherwise.

Filing this under something I need to think about for OS1+ updates.

pkg/interface/link/src/js/components/root.js
```diff
@@ -26,62 +33,167 @@ export class Root extends Component {
...
+      state.invites['/link'] : {};
+
     return (
-      <BrowserRouter>
+      <BrowserRouter><Switch>
         <Route exact path="/~link"
           render={ (props) => {
             return (
               <Skeleton
-                active="channels"
```
matildepark: This `active` switch stuff was for mobile. I think `rightPanelHide` might make it all redundant, probably, but there should be a catch in `skeleton.js` that isn't there now? Now the root page on mobile looks like this:

image

pkg/interface/link/src/js/components/lib/channel-sidebar.js
```diff
@@ -81,15 +61,15 @@ export class ChannelsSidebar extends Component {
...
-          <h2 className={"f8 f9-m f9-l f9-xl " +
-           "pt1 pt4-m pt4-l pt4-xl " +
-           "pr4 pb3 pb3-m pb3-l pb3-xl " +
-           "pl3 pl4-m pl4-l pl4-xl " +
-           "black-s gray2 white-d c-default " +
-           "bb b--gray4 b--gray2-d mb2 mb0-m mb0-l mb0-xl"}>
-             Your Collections
-             </h2>
-          {privateChannel}
+          <div className="w-100 bg-transparent pa4 bb b--gray4 b--gray1-d"
```
matildepark: add `b--gray1-d` to custom.css in the dark section.

pkg/interface/link/src/js/components/lib/invite-search.js
```diff
@@ -0,0 +1,369 @@
...
+            left: 12
+          }}
+        />
+        <textarea
+          ref={e => {
+            this.textarea = e;
+          }}
+          className={
+            "f7 ba b--gray3 b--gray2-d bg-gray0-d white-d pa3 w-100" +
+            " db focus-b--black focus-b--white-d"
```
matildepark: Add the `focus-b--white-d` stuff to custom.css in the dark section.

pkg/interface/link/src/js/components/settings.js
matildepark: Then I'm adding option 1 to internal kanban and this is resolved — just remove `mono` @Fang-

matildepark:
Otherwise, functionally seems good! Works well across it and Contacts/Groups, everything is in place. Good stuff.

matildepark:
Oh right you'll also want to target master now.

matildepark:
Oh another thing for @urcades. So when you use a `pa4` which should make the top nav look nice and even ... it's unaligned with the rest of the nav:

image

In Chat what I did was this enforced `padding-bottom` on the nav that ... makes the nav buttons no longer vertically centered.

image

Fang-:
Alright, I think I got everything, including the top nav alignment, and a couple small things not caught by review.

PR #2379, opened 2020-03-01
chat + contacts: integrate managed groups with metadata

loganallenc:
- Can create managed groups using contacts application or chat application
- Unmanaged group flow in chat is untouched
- Must join group through contacts before being able to join via chat (there is a need for a follow-up commit to hide invites for resources associated with groups that we have not joined yet)
- Chat removal / group removal seems to work but still throws the %ack-for-nonexistent channel prints? Not sure where that's coming from.

Still need to test:
- Complex cases where you join, are kicked, and rejoin groups and their associated resources repeatedly

Update to the PR:
- Permission-hook now automatically makes permissions for a managed group subscribable. This finishes the use case requirements needed for links and chat metadata to sync properly.


matildepark:
This seems premature but looks fine — still needs the new.js added with the title and desc submission. Most of the contacts JS changes are just rewriting the reducer around objects and not a map but I'm unsure why it was necessary.

loganallenc:
I was experiencing a crash in Contacts JS upon creating a group (because groups now automatically have a metadata entry for themselves added), and was unable to solve it (either because I don't understand `Map` well enough or for some other reason) without rewriting that reducer logic.

pkg/interface/contacts/src/js/components/lib/group-detail.js
```diff
@@ -10,36 +10,35 @@ export class GroupDetail extends Component {
...
-
-    let channelList = <div/>
-
-    channelList = Object.keys(groupChannels).map((channel) => {
-      let channelObj = groupChannels[channel];
-      if (!channelObj) {
-        return false;
+    channelList = Object.keys(channelsForGroup).map((key) => {
+      let channel = channelsForGroup[key];
+      console.log(key, channel);
```
matildepark: errant console.log

loganallenc: removed thanks

pkg/arvo/app/contact-view.hoon
```diff
@@ -126,18 +133,24 @@
...
+    %+  weld
     :~  (group-poke [%bundle path.act])
         (contact-poke [%create path.act])
         (contact-hook-poke [%add-owned path.act])
         (group-hook-poke [%add our.bol path.act])
         (group-poke [%add (~(put in ships.act) our.bol) path.act])
+        (perm-group-hook-poke [%associate path.act [[path.act %white] ~ ~]])
+        (permission-hook-poke [%add-owned path.act path.act])
     ==
+    (create-metadata path.act title.act description.act)
```
Fang-: I don't think you need the `weld` here (and for `%delete` below). You can just `:*` (in place of `:~`) the current list, and add `(create-metadata ...)` as the last item. It's like `:-`ing all these in front of the list that `+create-metadata` produces.

pkg/arvo/app/contact-view.hoon
```diff
@@ -218,6 +231,51 @@
...
+  ^-  (list card)
+  =/  =metadata
+    %*  .  *metadata
+        title         title
+        description   description
+        date-created  now.bol
+        creator       our.bol
+    ==
+  :~  (metadata-poke [%add path [%contacts path] metadata])
+      (metadata-hook-poke [%add-owned path])
```
Fang-: Note: it seems that with metadata (and most other stores) we have consistent rules about when and how to expose its data to others. Might be worth thinking about baking that into the hook directly.

pkg/arvo/app/contact-hook.hoon
```diff
@@ -242,7 +247,10 @@
         %delete
       =.  synced  (~(del by synced) path.fact)
       :_  state
-      [(group-poke [%unbundle path.fact])]~
+      :~  (group-poke [%unbundle path.fact])
+          (metadata-hook-poke [%remove path.fact]) 
+          (metadata-poke [%remove path.fact [%contacts path.fact]])
+      ==
```
Fang-: Can you explain why, for (local) creation, metadata management lives in the view, but for deletion, it happens in the hook?

pkg/interface/contacts/src/js/api.js
```diff
@@ -69,7 +69,14 @@ class UrbitApi {
   }
 
   contactCreate(path, ships = []) {
-    return this.contactViewAction({ create: { path, ships }});
+    return this.contactViewAction({
+      create: {
+        path,
+        ships,
+        title: '',
+        description: ''
```
Fang-: Do we not want to let the user input these?

loganallenc: We do, but I didn’t want to do that work in this PR

pkg/arvo/app/contact-hook.hoon
loganallenc: Within contacts, we’ve followed a pattern in which:
- %contact-view sets everything up and has a delete action that cleans everything up
- alternatively, if you delete the group in group-store that contacts depends on, %contact-hook will clean up all of the associated data

pkg/arvo/app/contact-view.hoon
loganallenc: This is a good idea. Agreed that there’s probably something here.

pkg/arvo/app/contact-view.hoon
loganallenc: I prefer the weld.

pkg/arvo/app/chat-view.hoon
```diff
@@ -247,18 +250,40 @@
...
+      %+  turn  ~(tap in u.group)
+      |=  =ship
+      ?:  =(ship our.bol)  ~
+      [(send-invite app-path ship)]~
     ::  do not create a managed group if this is a sig path or a blacklist
     ::
-    ?:  =(security %channel)
-      ~[(group-poke [%bundle path])]
+    ?:  =(sec %channel)
+      :~  (group-poke [%bundle path])
```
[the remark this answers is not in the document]
loganallenc: I suppose we could pass a boolean instead of the /~, but I don’t think that makes it much better

pkg/arvo/app/chat-view.hoon
```diff
@@ -247,18 +250,40 @@
...
+    |=  [=path app-path=path sec=rw-security ships=(set ship) title=@t desc=@t]
     ^-  (list card)
-    ?^  (group-scry path)  ~
+    =/  group  (group-scry path)
+    ?^  group
+      %-  zing
+      %+  turn  ~(tap in u.group)
+      |=  =ship
+      ?:  =(ship our.bol)  ~
+      [(send-invite app-path ship)]~
```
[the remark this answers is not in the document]
loganallenc: I should add an assert that says if group-path is anything other than ~, ships must be empty. Neither the front-end or the backend supports that case, so we should make it explicit.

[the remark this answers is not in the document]
Fang-: Should probably just assert that on the type level then, by doing `$%([%group =path] [%ships ships=(set ship)])` or something.

pkg/arvo/app/chat-view.hoon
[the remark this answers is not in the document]
Fang-: Oh, right, you still want the unmanaged whitelist case.

pkg/arvo/app/contact-view.hoon
Fang-: Then can you please indent their first argument properly? (:

pkg/arvo/app/chat-view.hoon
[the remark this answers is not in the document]
loganallenc: Good idea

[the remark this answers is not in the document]
loganallenc: Ended up asserting that either group-path = app-path OR members (ships) is empty, so won't be including this in the type for now.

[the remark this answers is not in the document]
loganallenc: I think that's more explicit about what the actual expectations are.

pkg/arvo/app/contact-view.hoon
loganallenc: ok

loganallenc:
Should be good to go.

PR #2384, opened 2020-03-02
ford: restores pit-shortcircuit, but only during boot

joemfb:
This PR is a rethink of #2322 -- slightly less principled, but much more performant. It restores %ford's "pit" short-circuit for kernel builds, but only during the boot process. Bugs caused by faulty "pit" short-circuit during normal operation will still be fixed as they were in #2322, but boot and `+test` are both much faster. These changes are necessary to restore well-functioning CI, and keep our kernel cycle-times somewhat reasonable.

These changes were made on #2366, but should go in separately (and maybe first). I'll update that PR once this is complete and merged.

### DIFF
--- a/pkg/arvo/sys/vane/ford.hoon
+++ b/pkg/arvo/sys/vane/ford.hoon
@@ -4658,6 +4658,23 @@
       ::
       ?.  ?=([~ %success %scry *] zuse-scry-result)
         (wrap-error zuse-scry-result)
+      ::
+      ::  short-circuit to .pit during boot
+      ::
+      ::    This avoids needing to recompile the kernel if we're asked
+      ::    to build %hoon one the home desk, at revision 1 or 2.
+      ::
+      ?:  ?&  =(our ship.disc)
+              ?=(?(%base %home) desk.disc)
+          ::
+              =/  =beam
+                [[ship.disc desk.disc [%da date.build]] /hoon/hoon/sys]
+              =/  cass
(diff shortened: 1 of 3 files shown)


pkg/arvo/sys/vane/ford.hoon
```diff
@@ -4658,6 +4658,23 @@
       ::
       ?.  ?=([~ %success %scry *] zuse-scry-result)
         (wrap-error zuse-scry-result)
+      ::
+      ::  short-circuit to .pit during boot
+      ::
+      ::    This avoids needing to recompile the kernel if we're asked
+      ::    to build %hoon one the home desk, at revision 1 or 2.
```
eglaysher: How does this deal with zuse changes where zuse.hoon differs in the arvo snapshot from the pill?

joemfb: You mean when `zuse.hoon` in the initial filesystem (`%into` event) in the pill is different from the `zuse.hoon` that was used to build %zuse in arvo (`%veer` event)? The answer to that is "not well". For now, I think  "don't make bad pills". Longer: it shouldn't be possible to get into this state, which is why arvo should keep a little mini filesystem in sync with the kernel. That's the approach from `neo/arvo`, and I've ported it and forward on a branch: https://github.com/urbit/urbit/blob/5d9c11e71243ebb9271b4ae9da4119a44b0762a9/pkg/arvo/gen/deep.hoon. I intend to implement that in arvo in my next round of changes.

PR #2385, opened 2020-03-02
Add command-line flags to configure HTTP and HTTPS ports.

benjamin-tlon:


pkg/hs/urbit-king/lib/Urbit/Vere/Http/Server.hs
```diff
@@ -396,41 +416,90 @@ app env sId liv plan which req respond =
...
+        throwIO (CantOpenPort p)
+      Right ps -> do
+        logTrace (display ("Opened port " <> tshow (fst ps)))
+        pure ps
+
+  bindTo = if soLocalhost then "127.0.0.1" else "0.0.0.0"
+
+  getBindAddr :: W.Port -> IO SockAddr
+  getBindAddr por =
+    Net.getAddrInfo Nothing (Just bindTo) (Just (show por)) >>= \case
```
benjamin-tlon: Previous behavior was a bug. This actually *always* ignored the suggested ports and used whatever free ports were available.

benjamin-tlon:
@joemfb @philipcmonk @ixv @eglaysher 

This might be a good time to review HTTP port behavior in general.

The behavior for (insecure HTTP) after this change is:

- If user specified a port, use that.

  - If we can't open it crash.

- Otherwise try a small list of preferable ports.

  - If we are a fake ship, those ports are [80, 8080, 8081, 8082, 8083, 8084, 8085]

  - If we are a real ship, those ports are [8080, 8081, 8082, 8083, 8084, 8085]

  - If we can't open any of those ports, just ask the OS for any free port.

benjamin-tlon:
> Other than frowning at more lens usage, this LGTM.
> 
> > If we are a fake ship, those ports are [80, 8080, 8081, 8082, 8083, 8084, 8085]
> 
> Did you mean this the other way?

Yeah, I meant the other way.

benjamin-tlon:
@jtobin did

PR #2387, opened 2020-03-03
metadata-hook: support group members other than host creating/editing shared resources

loganallenc:
Pretty small change, but nice quality of life improvement.

NOTE: this is good to keep in mind for links and publish, because the preference is now to send all metadata actions through %metadata-hook so that we have automatic support for foreign chatrooms / publish blogs / links collections being shown in the group's available channels. (not just the host's channels)

### DIFF
--- a/pkg/arvo/mar/metadata/action.hoon
+++ b/pkg/arvo/mar/metadata/action.hoon
@@ -1,0 +1,7 @@
+/-  *metadata-store
+|_  act=metadata-action
+++  grab
+  |%
+  ++  noun  metadata-action
+  --
+--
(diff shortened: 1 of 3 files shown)


loganallenc:
Note: I have one change incoming that makes it so that "leaving a chat that you are not the host of" doesn't affect the metadata of that chat.

pkg/arvo/mar/metadata/action.hoon
```diff
@@ -0,0 +1,7 @@
+/-  *metadata-store
+|_  act=metadata-action
+++  grab
```
Fang-: Maybe also write the trivial `+noun:grow`, for completeness' sake.

loganallenc: Matilde’s PR includes a grow arm to noun and json already. I’ll merge this when I get back from workout at noon

loganallenc: Alternatively, feel free to merge sooner. The conflict in that file is trivial

PR #2388, opened 2020-03-03
chat: settings — edit channel metadata, shortcode copying

matildepark:
- Adds sig-prepended, unmanaged chats to the regex that linkifies channels when mentioned in chat
- Adds a "copy shortcode" section to settings with the channel name for easy copying
- Dark mode linkified chats had a black border underneath them instead of white. No longer.
- Took a detour in ~hell~ lib/metadata-json.hoon and wrote JSON to metadata action parsers for add and remove actions, because I can do that now
- Constructed the API calls in the front-end for revising metadata
- Added 'update-metadata' responses to the metadata reducer (it just overwrites the entry in the map)
- Added edit fields for title, description, and color; the color field is regex validated for 6 hexadecimal characters; all fields pre-fill with current metadata and are disabled and faded if you aren't the owner of the chat; would appreciate a further test for people who can easily do multi-ship testing, but it shouldn't be errant

### DIFF
--- a/pkg/interface/chat/src/js/api.js
+++ b/pkg/interface/chat/src/js/api.js
@@ -183,7 +183,7 @@ class UrbitApi {
 
   metadataAction(data) {
     console.log(data)
-    return this.action("metadata-store", "metadata-action", data);
+    return this.action("metadata-hook", "metadata-action", data);
   }
 
   metadataAdd(appPath, groupPath, title, description, dateCreated, color) {
@@ -196,11 +196,11 @@ class UrbitApi {
           "app-name": "chat"
         },
         metadata: {
-          title: title,
-          description: description,
-          color: color,
+          title,
+          description,
+          color,
           'date-created': dateCreated,
-          creator: creator
+          creator
         }
       }
     })
--- a/pkg/interface/chat/src/js/api.js
+++ b/pkg/interface/chat/src/js/api.js
@@ -182,7 +182,6 @@ class UrbitApi {
   }
 
   metadataAction(data) {
-    console.log(data)
     return this.action("metadata-hook", "metadata-action", data);
   }
 
--- a/pkg/interface/chat/src/js/api.js
+++ b/pkg/interface/chat/src/js/api.js
@@ -181,6 +181,28 @@ class UrbitApi {
     });
   }
 
+  metadataAction(data) {
+    return this.action("metadata-store", "metadata-action", data);
+  }
+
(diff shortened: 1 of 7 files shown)


matildepark:
Probably "copied" shouldn't be exclamation mark'd. @urcades

urcades:
Yeah, good call

pkg/interface/chat/src/js/api.js
```diff
@@ -181,6 +181,31 @@ class UrbitApi {
     });
   }
 
+  metadataAction(data) {
+    console.log(data)
+    return this.action("metadata-store", "metadata-action", data);
```
loganallenc: Send this to metadata-hook. (My most recent PR notes that this is now the expected behavior)

pkg/interface/chat/src/js/api.js
```diff
@@ -181,6 +181,31 @@ class UrbitApi {
...
+      add: {
+        "group-path": groupPath,
+        resource: {
+          "app-path": appPath,
+          "app-name": "chat"
+        },
+        metadata: {
+          title: title,
+          description: description,
+          color: color,
```
loganallenc: Did you know that you can just write ‘color,’ instead of ‘color: color,’?

matildepark: Wtf? I mean I guess I get it because passing a single argument to the object wouldn’t have any other key but it assumes the argument is the key name?

matildepark:
This stuff doesn’t have spinner behaviour. I meant to do that before PR. Let me do that before merge with these changes.

pkg/interface/chat/src/js/api.js
loganallenc: yeah, it makes the variable name the key

matildepark:
Clear

pkg/interface/chat/src/js/api.js
```diff
@@ -181,6 +181,31 @@ class UrbitApi {
     });
   }
 
+  metadataAction(data) {
+    console.log(data)
```
loganallenc: console log :(

matildepark: NO

loganallenc:
lmao i’m merging this

matildepark:
Note for onlookers that the metadata actions are going to be broken in this branch until #2387 gets merged in the morning since the metadata-hook changes to take these API actions are in that PR.

PR #2392, opened 2020-03-04
eyre: remove padding from 'access' input

matildepark:
More tiny polish things as I remember them ... in this case, the access input's subheader was spaced odd to me. @urcades confirmed it should go, so here it is.

I feel we should ship a pill with our eyre changes, but haven't done that before. So, what do.


loganallenc:
@matildepark for future reference, run:
`
sh/update-ivory-pill
sh/update-brass-pill
sh/update-solid-pill
`

PR #2397, opened 2020-03-04
contacts: prune unmanaged "share" flow, rename to "groups"

matildepark:
- `~contacts` -> `~groups`
- Renamed the title in the tile and on the HTML page accordingly.
- Removed the dummy `Share` button in the root identity view
- Removed `/~/default` from the group sidebar
- Rename "Root Identity" to "Your Identity"


matildepark:
Included the built file by accident, standby ...

matildepark:
OK, pulling that compiled file is a gigantic pain in the ass because of the simultaneous renaming. I wonder if I should just `gulp bundle-prod` it to bring the diff down?

loganallenc:
I'd recommend not moving the files from app/contacts to app/groups. The files are still being served to the browser by contact-view.

loganallenc:
Everything else about this change looks good.

matildepark:
You mean in pkg/arvo? And still mounting at `~groups` in Landscape?

loganallenc:
Correct.

pkg/interface/groups/src/js/components/lib/add-contact.js
```diff
@@ -63,7 +63,7 @@ export class AddScreen extends Component {
     return (
       <div className="h-100 w-100 flex flex-column overflow-y-scroll">
         <div className="w-100 dn-m dn-l dn-xl inter pt1 pb6 pl3 pt3 f8">
-          <Link to={"/~contacts" + props.path}>{"⟵ All Contacts"}</Link>
+          <Link to={"/~groups" + props.path}>{"⟵ All Contacts"}</Link>
```
loganallenc: All Groups :)

matildepark: All Groups is the first sidebar. This mobile nav takes you back to the second one, of the contacts in the group.

loganallenc: kk

PR #2399, opened 2020-03-04
publish: integration with metadata

loganallenc:
Tested:
- Publish notebooks are added and deleted properly within managed groups that existed prior to notebook creation OR that were created through publish
- Unmanaged notebooks are created and invites are sent out properly
- Join for managed notebooks works
- Join for unmanaged notebooks works

Notes:
- Haven't looked into OTA compatibility yet but with this implementation it might just work out of the box. Need to test, but this PR should be good to go as is.


ixv:
When you make a notebook from a pre-existing group it sends invites and adds it to everyones metadata hook. But if you navigate to the notebook from contacts, it automatically subscribes, without removing the invite. If you then accept the invite you get crash from trying to subscribe twice on the same wire.

Joining unmanaged notebooks seems to work fine for me.

Agree, OTA might just work, but definitely test.

In another PR you made some changes to the make-groups arm. Some of the call sites pass empty strings to it when they don't need to.
In this case: https://github.com/urbit/urbit/commit/bb0f56f17aa292b63da14b1b515c103463852248#diff-847f2df2c61ecac74f5d59a96be8b890L199 It has the title.
In this case: https://github.com/urbit/urbit/commit/bb0f56f17aa292b63da14b1b515c103463852248#diff-847f2df2c61ecac74f5d59a96be8b890L566 it has both title and description.

Not sure if you were planning on leaving those ~&s in, but if so one of them should be a ~|

loganallenc:
This is ready to go.

PR #2400, opened 2020-03-05
invite search: proper padding, blending of group name

matildepark:
Was starting work on incorporating the group names into the search and started with this — the blending of results was padded inconsistently and all black on dark mode. This should be solid now.

Outside this PR, I wanted to inquire about two issues with incorporating group names and slate it for later work:
1. We don't get group metadata from our subscriptions and we would have to add another subscription and reducer set to get them
2. Reducers are done a bit inconsistently at the moment anyway (Chat + Notebooks + hopefully the others using a Map, Contacts using an object), and throwing associations to the invite search component would basically expect a Map. Hopefully since Contacts' invite search usage disables group searching it wouldn't even come up (we could just throw a `new Map` in as a prop), but it's something that may come up once other apps have their own implementation of the metadata store.

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/invite-search.js
+++ b/pkg/interface/chat/src/js/components/lib/invite-search.js
@@ -205,7 +205,7 @@ export class InviteSearch extends Component {
     ) {
       let groupHeader =
         state.searchResults.groups.length > 0 ? (
-          <p className="f9 gray2 ph3">Groups</p>
+          <p className="f9 gray2 ph3 pb2">Groups</p>
         ) : (
           ""
         );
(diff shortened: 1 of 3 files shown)


loganallenc:
Question: for group names, must you start by writing the “~zod” prefix or can you start by typing in the “metadata title” of the group or the second segment of the path “chat-room”? Preferably, I’d like to see it work with all three.

matildepark:
It uses the `includes` method to filter of an array of the group names (includes on each string, not on the array itself), so it's a partial match anywhere. It's pretty fuzzy and lenient.

matildepark:
Though right now there's no "metadata title" at all because I'm trying to figure out how we would solve that without getting a second subscription to get those names and integrate it all into a reducer just for the search component...

matildepark:
Honestly would prefer the back-end provided `group-name` alongside `app-path`, `app-name` and `group-path` over a second subscription and reducer for the whole contacts metadata object.

loganallenc:
Just add the other subscription. It’s all good

matildepark:
nehhhhhhhhhh

loganallenc:
Extra subscriptions are cheap, performance-wise. It’ll be a while before they become a bottleneck in any way, particularly since our APIs already paginate.

loganallenc:
If you look at how Channel.js works, a single Channel has multiple subscriptions. It isn’t like you’re opening multiple sockets. It’s all shared on a single connection.

matildepark:
Just thinking about how to make that not suck for the store construction. The associations map is based on our app-path from the chat subscription. Do we just make another associations map in the store? That feels bad.

matildepark:
Can I get a stamp on this to have this in for tomorrow's work?

PR #2405, opened 2020-03-06
OS1: metadata restructuring across applications, group metadata editing

matildepark:
- Rewriting the metadata data structure to be nested objects in Chat, etc. @loganallenc mentioned that a map wasn't *wrong*, but specialised for iteration order, and I inferred that in these cases you should just fall back to a more general object. 

The reason why it's nested now is because we have two subscriptions to metadata-store: one for the app we're using, and one for the metadata of the group itself. This is in preparation for some changes to invite-search to surface the group's *name* instead of its endpoint; and in general we want the data structure to be *consistent* if we're going to be building multiple, reusable components that consume the same data.

- Chat headers now use the chat's title, not the endpoint. Some polish here to clip off the starting `/` if the title is blank and it falls back to naming it the endpoint in the sidebar and header.
- Some buttons in Groups didn't transition to dark mode correctly; they're also `f9` now to match the other apps
- Add a nav back to the group details view from the contact card view in Groups
- In general Groups got some padding / spacing polish across its sidebars
- Rewrote the metadata structure in Links to be the same nested object pattern; while its idiosyncrasies led to some cleaner assignments, they didn't line up with the back-end in a way that was replicable across all applications and it also wasn't built for having multiple metadata subscriptions.
- Some tweaks to the Links channel headers with `lh-solid`
- Added group titles and descriptions to the group detail view in Groups; also added a group settings page for the group owner to edit the metadata of the group itself. Colour is exempt right now as we don't surface group colour anywhere — but earlier mocks had them for groups, so we could probably work it in as we go.

image

- Publish adds a subscription to the metadata-store for group metadata — it won't get another one for Notebook metadata as it's the source of truth for that, but we want the group metadata on hand for the upcoming component change.


pkg/interface/chat/src/js/reducers/metadata-update.js
```diff
@@ -14,10 +14,14 @@ export class MetadataReducer {
   associations(json, state) {
     let data = _.get(json, 'associations', false);
     if (data) {
-      let metadata = new Map;
+      let metadata = state.associations;
       Object.keys(data).map((channel) => {
         let channelObj = data[channel];
-        metadata.set(channelObj["app-path"], channelObj);
+        let app = data[channel]["app-name"];
+        if (!metadata[app]) {
```
loganallenc: if(!(app in metadata)) is much safer

pkg/interface/chat/src/js/components/sidebar.js
```diff
@@ -39,10 +39,10 @@ export class Sidebar extends Component {
         let author = !!msg ? msg.author : '';
         let when = !!msg ? msg.when : 0;
 
-        let title = box;
-        if ((props.associations.has(box)) && (props.associations.get(box).metadata)) {
-        title = (props.associations.get(box).metadata.title)
-          ? props.associations.get(box).metadata.title : box;
+        let title = box.substr(1);
+        if ((props.associations.chat[box]) && (props.associations.chat[box].metadata)) {
```
loganallenc: use (box in props.associations[chat]) for safety

pkg/interface/chat/src/js/components/settings.js
```diff
@@ -234,6 +234,13 @@ export class SettingsScreen extends Component {
         text = "Leaving...";
       }
 
+      let title = props.station.substr(1);
+
+      if ((props.association) && (props.association.metadata)) {
```
loganallenc: “metadata“ in props.associations for safety

pkg/interface/chat/src/js/components/root.js
```diff
@@ -210,6 +215,8 @@ export class Root extends Component {
               };
               let popout = props.match.url.includes("/popout/");
 
+              let association = (associations.chat[station])
```
loganallenc: station in props.associations[chat] for safety

pkg/interface/chat/src/js/components/settings.js
matildepark: Wait, I don't follow why this one isn't safe. The && won't hit the second check unless the first one is initialised at all, and this is just a property check, so it won't hit `can't find metadata of undefined`

matildepark: Basically the pattern I use does assume it's being passed an object — and if it isn't, it'll probably crash. And I can see why the `in` operator catches that. But I can't think of a situation where it wouldn't be either uninitialised or an object...hm. I will do it anyway but want to figure out the rationale for safety here

loganallenc: At this point it’s just an pre-rational embedded heuristic I use as a means of overly defensive programming within a hostile JS environment, I’ll try to find resources relating to it.

loganallenc: Alright, I’ve come to the conclusion that I just have an irrational fear of object accessors, I can’t find anything.

matildepark: I will do it anyway

Fang-: Potentially of interest: [`?.`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

(not suggesting using this right now, of course. just sharing)

matildepark: I’m aware of option chaining but I’m unsure if we use Babel and iirc native support is patchy.

loganallenc: ?. is a lovely convention in swift, hopefully it'll become widely used in JS.

matildepark:
Standby @ixv, I'm gonna do the safety stuff and triple wrap this...

matildepark:
Ready to go!

matildepark:
... oh wait standby again I found a bug

matildepark:
Done, good to go again ...

PR #2410, opened 2020-03-06
launch: subscribe to invite-view, show group invite count in tile

matildepark:
- Open a subscription to invite-view in launch
- Append the initial object to the data we pass to child tiles
- Add the invite count to the groups tile

## Additional notes

This doesn't account for invite-updates — we would need to write some proper reducers for launch. Do we want to do that now? Later?

### DIFF
--- a/pkg/interface/groups/src/js/components/new.js
+++ b/pkg/interface/groups/src/js/components/new.js
@@ -119,7 +119,7 @@ export class NewScreen extends Component {
               "focus-b--black focus-b--white-d"
             }
             rows={1}
-            placeholder="Two trumpeteers and a microphone"
+            placeholder="Two trumpeters and a microphone"
             style={{
               resize: "none",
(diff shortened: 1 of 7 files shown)


loganallenc:
Adding a reducer step in the store would probably be a good thing.

matildepark:
Alright, I'll take a sec and write a brief reducer for the invite stuff — the other ones dynamically update by directly overwriting themselves I think...

loganallenc:
Exactly, yeah... That should stay for now, but eventually we'll probably want to rethink it

matildepark:
Reduces invites correctly. Interesting that since if we create a group when making a channel, it all comes in as group invites, which makes sense really. Chat has zero data coming in for invites, even if it's an unmanaged invite?...

matildepark:
ping @loganallenc

PR #2411, opened 2020-03-06
`urbit-atom` 32 bit and GHCJS support.

benjamin-tlon:


benjamin-tlon:
This seems right, but still doesn't work correctly on GHCJS. Need to bring back the tests.

benjamin-tlon:
Maybe GHCJS `Natural`'s are broken? https://github.com/ghcjs/ghcjs/commit/c3922ede190a7dda6c0b673a3b7441cf41afcc9c

benjamin-tlon:
Alpine Linux build works just fine. Basically the same `--host` option, same source URL, also using `musl`. What's the difference?

https://pkgs.alpinelinux.org/package/edge/testing/armv7/libsigsegv

benjamin-tlon:
Happy to take review on this, but mostly just want to get it merged.

benjamin-tlon:
Who can merge this? @eglaysher @jtobin @ixv

eglaysher:
(I can't, otherwise I would.)

PR #2415, opened 2020-03-07
invite search: surface group metadata in results

matildepark:
- If we have group metadata, organise our initialised index of groups as an array of `["endpoint", "title"]`
- When searching, filter against both indexes of the array with `||` so partial matches against either return the group as a result
- Show the title if we have it, otherwise show the endpoint
- If group results are disabled then group metadata can be exempted as a prop without a resulting crash. 

Screen Shot 2020-03-06 at 11 01 39 PM

If I can flex for a second: I wrote this entire thing in one go, cold, and it just worked. I felt really powerful.

## Additional notes

I don't know if calling `.includes()` on a string that isn't allocated in an array causes a crash (see L101 of invite-search.js in these commits) — but thus far we don't pass groups without the associated group metadata so ... 

I think maybe if we skip the dedupe we do by making an array from the set, and just passed the endpoint twice in those cases, it might fix that edge case. What do you think?

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/invite-search.js
+++ b/pkg/interface/chat/src/js/components/lib/invite-search.js
@@ -31,7 +31,20 @@ export class InviteSearch extends Component {
 
   peerUpdate() {
     let groups = Array.from(Object.keys(this.props.groups));
-    groups = groups.filter(e => !e.startsWith("/~/"));
+    groups = groups.filter(e => !e.startsWith("/~/"))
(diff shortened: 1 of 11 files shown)


loganallenc:
Thought: instead of “selected entity”...
“selected groups or ships”

matildepark:
ping @loganallenc

urcades:
I 100% realize this was already approved, just +1ing the changes

PR #2416, opened 2020-03-07
ota logic updates for chat and contacts

loganallenc:
I can write more about this later, but please review these fixes. They make otas work for chat and contacts. Need to fix an issue created while I was rebasing this onto master, but otherwise is ready for review.


loganallenc:
Looks like I really botched the merge yesterday... Closing this and will re-open with a more sanely merged version.

Issue #2413, opened 2020-03-07
+apt:by should prevent duplicate keys


belisarius222:
This `;;` should fail, since the map has duplicate keys, and while I checked that `+apt:by` is indeed getting run, it doesn't prevent this invalid map from being instantiated.  This should be a straightforward check to add, I think.
```
> ;;((map @ @) [[1 1] [[1 2] ~ ~] ~])
{[p=1 q=2] [p=1 q=1]}
> ~(key by ;;((map @ @) [[1 1] [[1 2] ~ ~] ~]))
{1}
> ~(val by ;;((map @ @) [[1 1] [[1 2] ~ ~] ~]))
~[1 2]
```

yosoyubik:
I was thinking to keep an array while traversing the tree in `+apt`, add keys, sort it and check for duplicates but what do you think about something like this straight out of your example?

```
      ++  apt                                               ::  check correctness
        =|  [l=(unit) r=(unit)]
        ?.  =(~(wyt in key) (lent val))  |
        |-  ^-  ?
        ?~  a   &
        ?&  ?~(l & (gor p.n.a u.l))
            ?~(r & (gor u.r p.n.a))
            ?~(l.a & ?&((mor p.n.a p.n.l.a) $(a l.a, l `p.n.a)))
            ?~(r.a & ?&((mor p.n.a p.n.r.a) $(a r.a, r `p.n.a)))
        ==
```

PR #2417, opened 2020-03-07
hoon: +apt:by check for duplicate keys

yosoyubik:
Fixes https://github.com/urbit/urbit/issues/2413

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1493,10 +1493,12 @@
     =|  {l/(unit) r/(unit)}
     |-  ^-  ?
     ?~  a   &
-    ?&  ?~(l & (gor p.n.a u.l))
-        ?~(r & (gor u.r p.n.a))
-        ?~(l.a & ?&((mor p.n.a p.n.l.a) $(a l.a, l `p.n.a)))
-        ?~(r.a & ?&((mor p.n.a p.n.r.a) $(a r.a, r `p.n.a)))
+    ?&  ?~(l & &((gor p.n.a u.l) !=(p.n.a u.l)))
+        ?~(r & &((gor u.r p.n.a) !=(u.r p.n.a)))
+        ?~  l.a   &
+        &((mor p.n.a p.n.l.a) !=(p.n.a p.n.l.a) $(a l.a, l `p.n.a))
+        ?~  r.a   &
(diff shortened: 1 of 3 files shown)


yosoyubik:
Thanks @belisarius222!

belisarius222:
Ah, right, as long as everything is ordered, checking that each pair of neighbors is unequal should be enough to ensure global key uniqueness.  Yeah, ok, that's worth doing.

—
~rovnys-ricfer
https://urbit.org

yosoyubik:
Thanks @joemfb! Definitively more elegant than my solution :)  I think I've addressed your solution correctly, but let me know if something is amiss. I have spaced the code so it fits within the 80-column margin but I can rearrange it differently.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1493,10 +1493,12 @@
...
-    ?&  ?~(l & (gor p.n.a u.l))
-        ?~(r & (gor u.r p.n.a))
-        ?~(l.a & ?&((mor p.n.a p.n.l.a) $(a l.a, l `p.n.a)))
-        ?~(r.a & ?&((mor p.n.a p.n.r.a) $(a r.a, r `p.n.a)))
+    ?&  ?~(l & &((gor p.n.a u.l) !=(p.n.a u.l)))
+        ?~(r & &((gor u.r p.n.a) !=(u.r p.n.a)))
+        ?~  l.a   &
+        &(&((mor p.n.a p.n.l.a) !=(p.n.a p.n.l.a)) $(a l.a, l `p.n.a))
+        ?~  r.a   &
+        &(&((mor p.n.a p.n.r.a) !=(p.n.a p.n.r.a)) $(a r.a, r `p.n.a))
```
joemfb: `&(...)` groups an arbitrary number of expressions, so there's no need to nest those; I'd just  `&((mor ...) !=(...) $(...))`. Sorry for misleading with my earlier comment!

The overall `?&` is pretty substantial now -- I'd be tempted to try different formatting and balance it more (for instance, it might be improved with all the `?~` in tall form). But that's pretty nitpicky, this formatting is also fine as is.

yosoyubik: Sounds good! thanks @joemfb I've tried with all `?~` in tall form but couldn't come up with something that didn't look that much better compared to the previous one. I'll go ahead with your first suggestion.

loganallenc:
What's the status on this PR? @joemfb @yosoyubik

PR #2419, opened 2020-03-08
chat: upgraded chat-hook to move old chats to /~/chat-path

loganallenc:
chat: ota attempt

chat: ota triggers chat-store to tell chat-hook to send cards to update chat-store's state

contact-view: commented out avatars and base64

chat: cleaned up commits


pkg/arvo/app/chat-hook.hoon
```diff
@@ -13,8 +13,14 @@
...
 ::
-+$  state-1  [%1 state-base]
++$  state-1  [%1 new-state]
 +$  state-0  [%0 state-base]
++$  new-state
+  $:  synced=(map path ship)
+      invite-created=_|
+      allow-history=(map path ?)
+      loaded-cards=(list card)
+  ==
```
Fang-: Questionable naming here. Using the `state-base` in this type definition can help make clear what's actually changing. (This would also help simplify the new-state construction below.)

```hoon
+$  new-state
  $:  loaded-cards=(list card)
      state-base      
  ==
```

ixv:
Looks like this will handle the case that chat-hook loads before chat-store, but not the reverse. If chat-store loads first it will send the poke, which will hit chat-hooks default-agent poke handler.

Pinging @philipcmonk because I want to know if the app load order on OTA is actually nondeterministic, since part of this PR is about handling said nondeterminism if it exists.

loganallenc:
It's easy to make it work in the opposite order, but I am interested to know if we expect determinism from this.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -637,12 +654,7 @@
   ::  about group path shape, and the chat that would match it.
   ::TODO  remove me at some point.
   ::
-  ?.  .^(? %gu (scot %p our.bol) %metadata-store (scot %da now.bol) ~)
-    ?:  ?=([%'~' @ ^] group-path)
-      ~&  [%assuming-ported-legacy-group group-path]
-      [t.group-path]~
```
Fang-: Is this not still sane to assume? (With change to just `group-path`, of course.) Same below.

pkg/arvo/app/chat-store.hoon
```diff
@@ -35,8 +34,12 @@
...
   ++  on-save   !>(state)
   ++  on-load
-    |=  old=vase
-    `this(state !<(state-zero old))
+    |=  old-vase=vase
+    =/  old  !<(versioned-state old-vase)
+    ?:  ?=(%1 -.old)
+      [~ this(state old)]
+    :_  this(state [%1 inbox.old])
+    [%pass /lo-chst %agent [our.bowl %chat-hook] %poke %noun !>(%store-load)]~
```
Fang-: `/lo-chst`  👀

ixv:
The hook sends data to the store, if the store isn't updated when that happens it fails, so Logan made the store send a trigger to the hook to tell it when its updated. But the problem is now it fails in the opposite case, since if the store updates before the hook, the hook won't process the trigger.

Fang-:
> The hook sends data to the store, if the store isn't updated when that happens it fails

How would it fail? Am I overlooking some API change we made to it?

ixv:
The %create action changed, iirc.

loganallenc:
Isaac is right, yes. The create action changed, which was causing OTAs to fail.

pkg/arvo/app/contact-view.hoon
```diff
@@ -180,21 +180,21 @@
...
-    ~&  pav+pav
-    ~&  name+name
-    =/  contact  (contact-scry `path`(weld pav [name]~))
-    ?~  contact  not-found:gen
-    ?~  avatar.u.contact  not-found:gen
-    =*  avatar  u.avatar.u.contact
-    =/  decoded  (de:base64 q.octs.avatar)
-    ?~  decoded  not-found:gen
-    [[200 ['content-type' content-type.avatar]~] `u.decoded]
+::      [%'~groups' %avatar @ *]
```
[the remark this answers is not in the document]
loganallenc: I'd much rather comment it out.

pkg/arvo/app/chat-hook.hoon
loganallenc: We don't want to do anything different in the "~" group case than the non-sig group case.

pkg/arvo/app/chat-hook.hoon
loganallenc: u right

pkg/arvo/sur/chat-hook.hoon
```diff
@@ -4,7 +4,7 @@
   $%  ::  %add-owned: make a chatroom accessible to foreign ships
       ::  specified by the rw-security model
       ::
-      [%add-owned =path security=rw-security allow-history=?]
+      [%add-owned =path allow-history=?]
```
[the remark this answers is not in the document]
loganallenc: Done

pkg/arvo/app/chat-hook.hoon
loganallenc: Done

loganallenc:
@ixv in the new commit, it will work in either order. thanks!

ixv:
So now if chat-hook updates first, it sends moves that crash, and then gets a trigger to resend them when it will work. If chat-store upgrades first, it receives the moves twice.

I'm suspicious of this, and I don't really have the bandwidth to test it right now. I'll merge if @Fang- gives the approval.

loganallenc:
I'm like 95% sure that these operations are idempotent.

loganallenc:
I'll do another test-pass to verify before this is merged.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -79,7 +77,7 @@
           ^-  (list (list card))
           %+  turn  ~(tap in keys)  generate-cards
         ==
-    [~ this(state upgraded-state)]
+    [loaded-cards.upgraded-state this(state upgraded-state)]
```
Fang-: If the hook upgrades first, the store moves in this are going to crash, right?

I think the thing to do here is just track whether one condition (of "store upgraded" and "hook upgraded") has already been met in a flag. Then, if we're trying to set the flag to true, but it already is, kick off the upgrade logic.

loganallenc:
If chat-store upgrades first, it actually only sends the moves once, because the trigger won't affect chat-hook in that case.

loganallenc:
So in "hook first" upgrade condition, it sends moves that all crash, then is triggered when chat store upgrades to send moves that work properly.
In "store first" upgrade condition, the moves are sent by the hook but the hook never receives a trigger.

PR #2423, opened 2020-03-09
hoon: fix ket not working past tistar

### LINKED ISSUE #1951: Cannot ^ket past =* tistar alias face
Either I'm seriously misunderstanding how this is supposed to work or something is seriously amiss here. Behold:

```hoon
> =+  a=1
  =+  b=2
  =*  a  b
  [a ^a ^^a]
[2 2 2]
```

It just... eats the `^`? I would've expected this resolve as `[2 1 find-error]`.

pkova:
Fixes #1951.

The definition of tistar is `{$tstr p/(pair term (unit spec)) q/hoon r/hoon}`. In the case of code like `=+  a=1  =+  b=2  =*  a  b  ^a` this becomes 
```
[%tstr p=[p=%a q=~] q=[%wing p=~[%b]] r=[%wing p=~[[%.n p=1 q=[~ %a]]]]]
```
In this code path `u.u.tyr` is the the wing at face q above, but of course the number of kets (p=1) gets dropped.

If the number of kets is 0, we do whatever we did before. Otherwise we ignore whatever is in the alias map and resolve the original term with one ket removed. We also modify the subject by removing the alias so that we're not infinite looping.

On master:
```
> =+  a=1  =+  b=2  =*  a  b  ^a
2
> =+  a=1  =+  b=2  =*  a  b  ^^^^^^^^^^a
2
```

With this change:
```
> =+  a=1  =+  b=2  =*  a  b  ^a
1
> =+  a=1  =+  b=2  =*  a  b  ^^^^^^^^^^a
-find.^^^^^^^^^^a
-find.^^^^^^^^^a
```

Now this is deep compiler stuff so it's very possible my fix is complete nonsense! Not sure who should take a look at this, help me @jtobin or @Fang-  or someone who knows.


pkova:
Thanks Joe, you're a boss. Fixed.

pilfer-pandex:
This will take some time to load back into my mind. On cursory inspection, this code seems fine, but given how intricate name resolution is, I won't feel comfortable that I've done my due dilligence on this until I've gone through and traced the types and the logic here. I don't expect any problems given that joe's stamped it, but still.

Also, sorry joe, this is not the hardest part of the compiler. That's the oc core, used in mint:et.

pkg/arvo/sys/hoon.hoon
```diff
@@ -9865,6 +9865,8 @@
                   next
                 ?~  u.tyr
                   $(sut q.sut, lon [~ lon], p.heg +(p.heg))
+                ?:  !=(0 p.heg)
+                  $(sut q.sut, lon [~ lon], p.heg (dec p.heg))
```
pilfer-pandex: You know, on reading the fond code for myself, I arrived at exactly the same question you had @joemfb: what are we modelling with the nullability of the hoon in an alias (u.tyr here)? The increment to the ^ count is extremely strange. Suppose you have a subject [◊/1 a/2 a/3]. If ◊ is a tune with a mapped to ~, then the wing a resolves to 3. These ~ aliases then seem to be a way to introduce nonlocal skipping behavior in name resolution — why do we want this functionality? In any event, it does seem to be the most consistent to place the ?: after this ?~  u.tyr, as OP has, which answers my remaining question about this code.

pkg/arvo/sys/hoon.hoon
```diff
@@ -9865,6 +9865,8 @@
                   next
```
pilfer-pandex: @joemfb What are "bridges" (q.zot, processed in next)? Are they emitted by anything? In general, what part of the parser produces tunes?

pkg/arvo/sys/hoon.hoon
joemfb: My earlier guess that this nullability/skip was an old artifact is wrong, it's specifically called out in https://urbit.org/docs/reference/hoon-expressions/advanced/. It's still strange: as far as I can tell, nothing uses this. And the code as written appears to include a double-skip, bypassing the `%face` (replacing `sut` with `q.sut`) *and* incrementing the skip count (`p.heg`).

pkg/arvo/sys/hoon.hoon
joemfb: Bridges are "imports", resulting from `=,`. The `%tune` AST node is produced in the macro-expansion of `%tstr`, and a tune is constructed (with `+busk`) when inferring the type of a `%tscm`.

pkova:
Did some spelunking in the git history and came across the suspicious commit 1fba410c4b5a6931d0c09eb5be0db8f9b02fd661. I like Curtis' original code better and it still works so I took the liberty of pushing it here. It seems like this feature was originally disabled due to some jet shenanigans.

pkova:
As for the mystery of the unit case hoon, git history reveals the following:

[Previously](https://github.com/urbit/urbit/blob/1832c14cca22bc3c7e2782a504c7b270ec95f395/arvo/hoon.hoon#L340)  `$face` was defined as `{$face p/$@(term tool) q/span}`. [`++tool` ](https://github.com/urbit/urbit/blob/1832c14cca22bc3c7e2782a504c7b270ec95f395/arvo/hoon.hoon#L348-L351) contained a `(map term (unit port))` instead of `(map term (unit hoon))` as it is today. [`++port`](https://github.com/urbit/urbit/blob/1832c14cca22bc3c7e2782a504c7b270ec95f395/arvo/hoon.hoon#L134) is some kind of search path which is clearly a unit value. The commit af35cff1d6711d52f9d4334b12b26c04416416b1 changes the alias system to use hoons (twigs) instead of ports, but continues using units, probably by oversight.

joemfb:
The unit(s) look the same to me before and after those changes. As I mentioned to @pilfer-pandex, the unit/skip feature is specifically called out in https://urbit.org/docs/reference/hoon-expressions/advanced/. So I think it's intentional, just not currently used by anything. (And I'm still not sure I understand what would uses it ...)

jtobin:
(FYI, I'll take care of the pill conflict here & merge after #2522 goes in.)

PR #2427, opened 2020-03-09
chat: only send invites for unmanaged chats

loganallenc:


vvisigoth:
king
—
~poldec-tonteg
http://urbit.org

PR #2431, opened 2020-03-10
publish: naive visual fix for publish body scrollbar-overlap

urcades:
As mentioned in the commit title, please keep in mind this is a "very naive" fix to the mentioned issue of "scrollbar overlap" over the publish body content.

This definitely needs more work in order to make the interface properly responsive and legible on larger/smaller screens, but it serves as a proof of itent for laptop-sized screens.

Will discuss more in PR comments.

### DIFF
--- a/pkg/interface/publish/src/js/components/lib/comment-item.js
+++ b/pkg/interface/publish/src/js/components/lib/comment-item.js
@@ -53,7 +53,7 @@ export class CommentItem extends Component {
 
     return (
       <div>
-        <div className="flex mv3 bg-white bg-gray0-d">
+        <div className="flex mv3 bg-white bg-gray0-d w-50 w-100-m center">
         <Sigil
           ship={commentData.author}
           size={24}
@@ -66,7 +66,7 @@ export class CommentItem extends Component {
           </div>
           <div className="f9 gray3 pt1">{date}</div>
         </div>
-        <div className="f8 lh-solid mb8 mb2">
+        <div className="f8 lh-solid mb8 mb2 w-50 w-100-m w-100-s center">
           {content}
         </div>
       </div>
--- a/pkg/interface/publish/src/js/components/lib/note.js
+++ b/pkg/interface/publish/src/js/components/lib/note.js
@@ -165,7 +165,7 @@ export class Note extends Component {
               popout={props.popout}
               sidebarShown={props.sidebarShown}
             />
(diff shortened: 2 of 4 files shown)


urcades:
OK, so just to get this immediately out there: If you review the code, you'll see I liberally applied the classes of `w-50 w-100-m w-100-s center` to basically all affected components I wanted to generally render in the center of the publish interface, at 50% the width of the container.

This 'fix' allows one to use the entire width of the publish interface as a scrollable element, rather than constrain scrollability to the actual body content itself. This UX improvement doubles as a visual improvement, as having a scrollbar butted up right next to the content of a post is very unsightly.

However, this fix obviously presents readability issues on viewports smaller and larger than a laptop screen, but I pushed these changes into this PR so people could get a sense for how this fix should more or less pan out, compared to the present implementation (which is also documented in this [Asana task](https://app.asana.com/0/1159034235108489/1165080933121593)).

Looking for some input on how to make these interfaces more responsive and less likely to collide with other components.

matildepark:
My reflexive guess is that this should really just be a matter of moving `overflow-y-scroll` into the parent (and `w-100 mw6` or etc. scales responsively a bit more nicely than the step down fixed width catches?) but I will look at this in a bit!

pkg/interface/publish/src/js/components/lib/comment-item.js
```diff
@@ -66,7 +66,7 @@ export class CommentItem extends Component {
           </div>
           <div className="f9 gray3 pt1">{date}</div>
         </div>
-        <div className="f8 lh-solid mb8 mb2">
+        <div className="f8 lh-solid mb8 mb2 w-50 w-100-m w-100-s center">
```
matildepark: For all of these: `w-100-s` isn't a class we have, you'd want `w-100 w-50-l w-50-xl` or something.

pkg/interface/publish/src/js/components/lib/note.js
```diff
@@ -165,7 +165,7 @@ export class Note extends Component {
               popout={props.popout}
               sidebarShown={props.sidebarShown}
             />
-            <Link className="f9 w-100 w-90-m w-90-l mw6 tl" to={baseUrl}>
+            <Link className="f9 w-50 w-100-m w-100-s center tl" to={baseUrl}>
```
matildepark: Careful — this regresses an issue the previous weird class stuff fixed on `-m` viewports:

image

PR #2434, opened 2020-03-10
chat-cli: support group-based chats

Fang-:
From 22cc6ae:

Re-enables chat creation, touches up invite logic, and makes everything work with the new "un/managed" attribute of chats.

Changes the +target type, so requires state transition. We use that opportunity to clean up our messages mirror (memory reclamation).

"Unmanaged" chats should work the same as they did before.
Group-based chats are secondary citizens (seriously, just use the group web-ui), but supported by prepending "group " to whatever target you want to use. ie:
```
;join group ~marzod/secret-club  ::  join a group-based chat
;group ~marzod/secret-club       ::  target a group-based chat
```

The latter case should be rarely needed, as glyphs remember this attribute of their bound target.
 
Creating a group alongside a chat is supported through:
`;create village-with-group /cool-kids`
 
You can then invite to that group (and by extension the associated chat) by doing:
`;invite group /cool-kids ~rinsed-walrus`

Submitting for review, but there's still some small amount of pending work:
- [x] Test OTA state transition of this some more
- [x] Subscribe to & render invites as they come in

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -803,14 +803,9 @@
         =;  perm=(unit permission)
           ?~(perm ~ `?=(%white kind.u.perm))
         ::TODO  +permission-of-target?
-        .^  (unit permission)
-            %gx
-            (scot %p our-self)
-            %permission-store
-            (scot %da now.bowl)
-            %permission
-            (snoc path %noun)
-        ==
+        %^  scry-for  (unit permission)
+          %permission-store
+        [%permission path]
       ?~  whitelist
         ~&  [%weird-no-permission path]
         ~
@@ -1000,12 +995,9 @@
       :_  ~
       %-  print-more:sh-out
       =/  all
-        ::TODO  refactor
-        ::TODO  remote scries fail... but moon support?
-        .^  (set path)
-            %gx
-            /(scot %p our-self)/chat-store/(scot %da now.bowl)/keys/noun
-        ==
+        %^  scry-for  (set path)
+          %chat-store
+        /keys
       %+  turn  ~(tap in all)
       %+  cork  path-to-target
       |=  target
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/chat-cli.hoon
```diff
@@ -17,8 +17,15 @@
...
-  $:  grams=(list mail)                             ::  all messages
+::
++$  versioned-state
+  $%  state-1
+      state-0
+  ==
+::
++$  state-1
+  $:  %1
+      grams=(list mail)                             ::  all messages
```
loganallenc: Why are we storing grams when we could scry them on-demand from chat-store? Seems unnecessary.

Fang-: idk how you would solve for message selection/expansion in that case though. This `(list mail)` contains messages from all chats. They get rendered in the same stream, and numbered sequentially.

`;;;` gets you the third-most-recent message, as plucked from this. I don't think you can make such queries against chat-store. Maybe you could "get all messages from everything, merge-sorted-by-timestamp", but performance aside I doubt the ordering that gives you is guaranteed to be identical to the `%fact`/rendering order.

loganallenc: Alternatively you could make the `chat-cli` "get message interface" take both a channel and a message number, and then you wouldn't have any real problem with either rendering order, which would be guaranteed, or performance, which would be a single scry, and quite reasonable.

Fang-: Giving separate numeric indicators for messages from different channels seems potentially confusing when we're rendering them in the same stream. I don't disagree that the current implementation is a bit of a hack, but consider fixing that out of scope for now.

loganallenc: Okay. I feel that this should be prioritized after the release. @vvisigoth Storing *double* chat's state seems entirely unnecessary and like it will impede our ability to properly export/import chat data down the line.

Fang-:
Should be good to go in pending approval on this one new commit.

PR #2437, opened 2020-03-11
chat: splice, URL test and send mixed content messages

matildepark:
Cuts up a message by gaps in the message, removes the URLs and reattaches the message content with the gaps intact; sends the URLs as separate messages after it.

image

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/chat-input.js
+++ b/pkg/interface/chat/src/js/components/lib/chat-input.js
@@ -133,12 +133,11 @@ export class ChatInput extends Component {
 
   isUrl(string) {
     try {
-      const urlObject = new URL(string);
-      //NOTE we check for a host to ensure a url is actually being posted
-      //     to combat false positives for things like "marzod: ur cool".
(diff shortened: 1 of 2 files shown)


pkg/interface/chat/src/js/components/lib/chat-input.js
```diff
@@ -151,14 +150,35 @@ export class ChatInput extends Component {
...
     );
+
+    for (let each of URLs) {
+      props.api.chat.message(
+        props.station,
+        `~${window.ship}`,
+        Date.now(),
+        each
+      );
+    }
```
Fang-: I thought the intended behavior was, for a message like "hey i like https://urbit.org it's really cool", to send as three messages in the following order:

- `%text` hey i like
- `%url` https://urbit.org
- `%text` it's really cool

Looks like the current implementation always sends with `%url`s after smooshing the rest of the content into a single `%text`, regardless of where they appear in the text. In the above example, that gets you "hey i like it's really cool". Seems not good?

matildepark:
Can adjust, just gotta figure out how to do the timestamp++ thing.

PR #2442, opened 2020-03-11
chat-hook: ota fix and improvement on backlog fetching

loganallenc:
Have tested this with ~bus and with ~zod. The issue was that my previous testing was on ~zod, where the bunted our.bol value of 0 was actually working with no issue. For the recreate-chat pokes to be going through, I needed to be calling the arms within the cc core.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -109,10 +109,10 @@
       =/  old-mailbox=mailbox
         (need (scry:cc (unit mailbox) %chat-store [%mailbox chat]))
       =*  enves  envelopes.old-mailbox
-      :~  (chat-poke [%delete chat])
-          (chat-poke [%create new-chat])
-          (chat-poke [%messages new-chat enves])
-          (chat-poke [%read new-chat])
+      :~  (chat-poke:cc [%delete chat])
+          (chat-poke:cc [%create new-chat])
+          (chat-poke:cc [%messages new-chat enves])
+          (chat-poke:cc [%read new-chat])
           %^  make-poke  %chat-hook  %chat-hook-action
           !>  ^-  chat-hook-action
           ?:  =(our.bol host)  [%add-owned new-chat %.y]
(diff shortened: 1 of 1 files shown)


ixv:
I'm blocked on testing this on the live net by this: https://github.com/urbit/bridge/issues/410

loganallenc:
Can I get a merge?

loganallenc:
The next bit of work that I've mentioned to you should be reviewed and go in separately.

ixv:
Ok

pkg/arvo/app/chat-hook.hoon
```diff
@@ -366,8 +366,11 @@
       =/  chat-path  [%mailbox path.act]
       :_  state
       [%pass chat-path %agent [ship.act %chat-hook] %watch chat-path]~
-    ::  TODO: only ask for backlog from previous point
-    =/  chat-history  [%backlog (weld path.act /0)]
+    =/  mailbox=(unit mailbox)  (chat-scry path.act)
+    =/  chat-history=path
+      :-  %backlog
+      %+  weld  path.act
+      ?~(mailbox /0 /(scot %ud (lent envelopes.u.mailbox)))
```
Fang-: I think you explicitly want the sequence number registered with the last-heard message, instead of the length? If I hadn't previously gotten history, or never got full history, or am not entirely in sync history-wise for any other reason, then `lent` here will differ from the one on the host side.

loganallenc: lent would be strictly less than or equal to the number of messages in the channel, so I’m not concerned. Worst case is a small number of duplicated messages, which is much better than previous behavior.

Fang-: > lent would be strictly less than or equal to the number of messages in the channel

I've now observed this not holding true for channels which get deleted, then recreated under the same name/path. It's making me miss out on messages whenever a reconnect happens. Using latest-heard message number here is more correct. (Doesn't fix the load-backlog-on-join case though.)

PR #2447, opened 2020-03-11
chat-hook: added ability to subscribe to chat-hook state from the front-end

loganallenc:
Context: When someone kicks you from a chat, you lose permission to post to that chat and to receive updates, but you still have the chat history. However the experience as a user is quite unpleasant, because there has up to now been no indication in the front-end that you had lost the permission to post to the chat. Now, we show a message explaining that you have been disconnected from the host for some reason, and show a button giving you the chance to try to resubscribe. I think this will smooth over the Chat OTA quite a bit, but it really needs a design pass.

The API works great, tested locally and remotely with adding members to a group, deleting them to kick their subscriptions, adding them back, then getting them to try resubscribing using the button, etc.


ixv:
Can you post pictures for the designers?

urcades:
Posting three thoughts, all similar, with small distinctions:

opt01
opt02
opt03

The primary idea here is to inject a - very visually distinct - component into the chat stream that indicates the current state of one's connection with the host.

The first image depicts a bit of a longer explainer, and a rightward CTA to attempt a reconnect.
The second image just left-aligns the entire thing, and drops some explainer in lieu of a reconnect button.
The third option basically automatically counts down and attempts an automatic reconnect, possibly every _n_ seconds.

Rather than sourcing this component into the stream as a 'box'-looking alert, we could also do this via plain ol' text, and possibly color it red or orange – something more visible than the yellow i'm using here.

A last detail that could be used in isolation, or as another indicator, is the message input text itself, which signals that one has been disconnected from the chat.

urcades:
cc @galenwp + @g-a-v-i-n  + @jyng , in case y'all have opinions on the three ideas. We currently don't really have a component that handles a case like this, would be good fodder for Indigo 1.1.1 component dev.

loganallenc:
I love option 2!

loganallenc:
https://drive.google.com/file/d/1Y4BzpP8gIC1lzvOTrxmGx-9MRhzrZVP0/view?usp=sharing

Here's a video. Implemented the design, hope it looks nice.

loganallenc:
Current design works properly. Looks like you have to refresh for the reducer to parse the `%messages` event properly in the UI, but that seems fine. It would be hard to implement a change to the reducer that didn't break backlog fetching.

loganallenc:
@ixv I no longer send invites upon OTA. Can't recreate the backlog duplication issue you were seeing in normal conditions. See video for an example... Will test in OTA conditions and see if it duplicates backlog in a way that's different than I'd expect otherwise?

urcades:
> https://drive.google.com/file/d/1Y4BzpP8gIC1lzvOTrxmGx-9MRhzrZVP0/view?usp=sharing
> 
> Here's a video. Implemented the design, hope it looks nice.

Beautiful — seeing this makes me realize there are some unrelated-to-issue states we should account for at some point in the future, for a more informative chat experience, but this will serve us well for a while.

loganallenc:
Paired with Isaac. We got to the bottom of an issue where chat-hook wasn't resubscribing to chat-store when its subscription with chat-store was kicked.

loganallenc:
Angsting out here but finally got a test that works completely.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -109,7 +103,8 @@
       =/  old-mailbox=mailbox
         (need (scry:cc (unit mailbox) %chat-store [%mailbox chat]))
       =*  enves  envelopes.old-mailbox
-      :~  (chat-poke:cc [%delete chat])
+      :~  (chat-poke:cc [%delete new-chat])
```
Fang-: Why are we concerned about these things already existing? Does anything create with `/~` prefix rn? (Commit says "fix ota chat backlog duplication" but I don't understand how we'd end up with duplication in the first place.)

If they exist, they'll pop up in the `turn  ~(tap in keys)` routine, meaning we'll create `/~/~` chats too. _If_ we want to deal with them, we should probably filter them out there and send deletes only?

loganallenc: Depending on the order that the OTA occurs in and the order that the hook and store come online, the events may be sent twice, so we needed to make sure they were idempotent. They’ve been tested and we’re sure they’re idempotent.

PR #2454, opened 2020-03-13
chat: split and send mixed messages sequentially

matildepark:
Hot and cold in that order. This amends the behaviour of messages
with text and URLs to submit them in order of found URL,
instead of batching the URLs at the end.

Same message, before and after:

Screen Shot 2020-03-12 at 9 44 41 PM

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/chat-input.js
+++ b/pkg/interface/chat/src/js/components/lib/chat-input.js
@@ -150,33 +150,43 @@ export class ChatInput extends Component {
       return;
     }
 
-    let URLs = [];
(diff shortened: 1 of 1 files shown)


matildepark:
These PRs have a tradeoff anything in vaguely URL format register as a positive. Why? Because the URI schema part of the regex is optional. But since we have people dumping error messages, hashes, etc that are being split up and made into weird links, I wonder if we should just make a URI schema mandatory to register as a positive (HTTP, HTTPS, DAT etc etc etc etc). It's a one character change. What do you think @Fang- ?

PR #2455, opened 2020-03-13
chat, link: preview for channel colour picker

matildepark:
Uses the same logic as Groups' sigil colour picker for submission, too.

image


matildepark:
I agree. AA asked for it to be the same behaviour as groups so let him decide.

vvisigoth:
To be clear, I just wanted the color preview behavior to be the same.
Submission logic can be the same as context (form).
—
~poldec-tonteg
http://urbit.org

matildepark:
Amended.

PR #2480, opened 2020-03-13
chat: groupify

Fang-:
Allows users to upgrade invite-only chats into full-fledged groups (with an associated chat). Relies on largely on existing chat-view logic to move things around (delete the old chat, create the new one, copy messages & metadata over).

Frontend copy might need another pass, but shouldn't be awful as-is.

### DIFF
--- a/pkg/interface/chat/src/js/components/settings.js
+++ b/pkg/interface/chat/src/js/components/settings.js
@@ -76,7 +76,22 @@ export class SettingsScreen extends Component {
     props.api.setSpinner(true);
 
     this.setState({
-      isLoading: true
+      isLoading: true,
+      loadingText: (deSig(props.match.params.ship) === window.ship)
+        ? 'Deleting...'
+        : 'Leaving...'
+    });
+  }
+
+  groupifyChat() {
+    const { props, state } = this;
+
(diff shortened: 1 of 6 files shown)


pkg/interface/chat/src/js/components/settings.js
```diff
@@ -106,6 +121,33 @@ export class SettingsScreen extends Component {
...
+      props.permission.kind === 'white';
+
+    if (!ownedUnmanagedVillage) {
+      return null;
+    } else {
+      return (
+        <div>
+          <div className={"w-100 fl mt3 "}>
+            <p className="f8 mt3 lh-copy db">Convert Chat</p>
+            <p className="f9 gray2 db mb4">Convert this chat into a group + associated chat.</p>
```
matildepark: I wouldn't use `+` in UI copy but God help me if I block this PR over this. We can come back and copy pass in the next few weeks.

Fang-: I can literally just use web editor to do this in one second. You could've used the gh suggestion button and I could've single-click applied it. This stuff is trivial lol, fixing rn.

Issue #2465, opened 2020-03-13
ship has not received groups invite


belisarius222:
~wicdev-wisryt invited ~rovnys-ricfer to a group, but ~rovnys hasn't received the invite after maybe fifteen minutes.  The ships can |hi each other just fine, and ~rovnys's connectivity to other ships also seems good.

liam-fitzgerald:
can also replicate with ~wicdev-wisryt inviting ~racfyr-simfur-midryl-maptud--tabfep-sidlen-sabwyx-marzod
|hi also works fine and connectivity is fine

philipcmonk:
Everyone says they're not getting the invites.

<img width="908" alt="Screen Shot 2020-03-12 at 10 59 22 PM" src="https://user-images.githubusercontent.com/5553176/76594030-2a508680-64b5-11ea-81f2-1846db7e5490.png">

reitraced:
i also had this issue with getting invites on my comet from a friend whos planet survived the ota

loganallenc:
Welp! Will look at this in the morning.

reitraced:
ok a note: after breaching my planet i made a group and invites work fine for it. idk if its because i breached or what but i just find it weird

PR #2484, opened 2020-03-13
contact-hook: re-establish connection to group-store if subscription …

loganallenc:
…failed during ota

This fixes the issue where members added to a group would occasionally not be sent an invite.


philipcmonk:
Have we tested this on a ship that loaded them correctly originally?

ixv:
Yes, both one that loaded contact-hook before group-store and one that loaded group-store before contact-hook.

philipcmonk:
CC #2465

PR #2493, opened 2020-03-14
chat: use more precise regex for splitting messages

matildepark:
See some chatter in #2454.

Current URL regex is getting too many false positives, cutting up debug conversations. Replacing with one that only matches positive if there's a schema in front:

- www.google.com doesn't match. img.jpg doesn't match. a base hash won't match.
- https://google.com will match.
- dat://wogjerpgorejgiuohg3oigh3g03ihtq3.onion will match.

cc @ixv

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/chat-input.js
+++ b/pkg/interface/chat/src/js/components/lib/chat-input.js
@@ -134,8 +134,7 @@ export class ChatInput extends Component {
   isUrl(string) {
     try {
       let websiteTest = new RegExp(''
-        + /[(http(s)?):\/\/(www\.)?a-zA-Z0-9@:%._\+~#=]{2,256}/.source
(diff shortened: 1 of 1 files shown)


matildepark:
It looks for HTTP or HTTPS and I want to support DAT and IPFS.

matildepark:
Also the whole thing is a capture group for true/falsey, then within it

- In first capture group it matches any number of alphanumeric characters
followed by ://
- The rest are any number of ASCII safe characters with a bunch of
backslashes to escape them

It's really a lot dumber than the other one other than that it enforces the
:// prefix and assumes as long as we have word://ascii.safe/characters it's
a URI.

I don't know why you're asking me to explain my own changes though.

Fang-:
> I don't know why you're asking me to explain my own changes though.

Can't review if I don't understand! This seems sane then, merge at your leisure.

matildepark:
I can't, it's to master.

PR #2498, opened 2020-03-14
Lsp parser

liam-fitzgerald:
First pass at a recoverable parser for the language server. Would appreciate a review for architecture, as I feel I'm not using hoon's polymorphism to the fullest.

cc: @philipcmonk


zalberico:
I think a good LSP will be useful for devex, but I think this is unlikely to get attention or merge anytime soon.

PR #2517, opened 2020-03-15
arvo: add web app manifest to launch app

fakezod:
This commit adds a web app manifest to the launch app so that Landscape can be added to the smartphone home screen and be used standalone (i.e. seperate from the browser) …

### DIFF
--- a/pkg/arvo/app/launch/index.hoon
+++ b/pkg/arvo/app/launch/index.hoon
@@ -3,11 +3,13 @@
   ;head
     ;title: Home
     ;meta(charset "utf-8");
-    ;meta
-      =name     "viewport"
-      =content  "width=device-width, initial-scale=1, shrink-to-fit=no";
-      ;link(rel "stylesheet", href "/~launch/css/index.css");
-      ;link(rel "icon", type "image/png", href "/~launch/img/Favicon.png");
+    ;meta(name "viewport", content "width=device-width, initial-scale=1, shrink-to-fit=no");
+    ;meta(name "apple-mobile-web-app-capable", content "yes");
+    ;meta(name "apple-mobile-web-app-status-bar-style", content "default");
+    ;meta(name "apple-touch-fullscreen", content "yes");
+    ;link(rel "stylesheet", href "/~launch/css/index.css");
+    ;link(rel "icon", type "image/png", href "/~launch/img/Favicon.png");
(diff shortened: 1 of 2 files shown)


matildepark:
Have you tested? I don't think we route the mounting of arbitrary `json` files for the ~launch endpoint in launch.hoon (this is why I inlined it in Chat)

fakezod:
After very lots of compile errors, I fear it’s not possible to inline the `data:application/manifest+json,{…}` string in Sail/Udon?

Some “quick fix” would be converting `launch.hoon` to `launch.html`, but the wildcard loading of tile scripts, see below, is quite convenient.

```
|=  scripts=marl
;html
  […]
    ;script@"/~/channel/channel.js";
    ;script@"/~modulo/session.js";
    ;*  scripts
    ;script@"/~launch/js/index.js";
  […]
```

So maybe the best way to solve this would be to, as you wrote above, route the mounting of arbitrary `.json` files for the `/~launch` endpoint in `launch.hoon`, what do you think?

matildepark:
Hi, yeah probably. I've honestly never done that — I've only really mounted png stuff. Let's ask @Fang-.

Fang-:
> So maybe the best way to solve this would be to, as you wrote above, route the mounting of arbitrary .json files for the /~launch endpoint in launch.hoon, what do you think?

That seems sanest to me, yeah. You'd be adding a case to [this `?+`](https://github.com/urbit/urbit/blob/1fe7ef8617f456022ee3738b68ea6f770e6e0abd/pkg/arvo/app/launch.hoon#L90), which already handles serving `index.js` and such, to serve `manifest.json` also. To follow the existing pattern, you import its contents at the top of the `launch.hoon` as is already done [for `index.js`](https://github.com/urbit/urbit/blob/1fe7ef8617f456022ee3738b68ea6f770e6e0abd/pkg/arvo/app/launch.hoon#L7-L13).

That should get you on your way! lmk if you run into trouble.

matildepark:
@fakezod Any luck? If not, we should close this in the next few days.

fakezod:
@matildepark Unfortunately no luck, yet. Wanted to give this another try after finishing Hoon School … I guess it would be incredibly simple if I would know some Hoon.

loganallenc:
Not to discourage, because I'd love for this work to go in, but I'm going to close this for now.

PR #2518, opened 2020-03-15
chat-hook: clear all subscriptions on watch-nack

Fang-:
We would remove the entry from state, but not clean up its subscriptions.

Found this trying to join a chat I was in that got deleted, then re-created. It yelled `subscribe wire not unique` at me for `/mailbox/~/~host/chat`.

@loganallenc Perhaps this change should also be made for the `%store` case? But that doesn't do any subscription cleaning rn so I'm less sure. Please confirm sanity on this in general.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -609,10 +609,7 @@
   ::
       [%backlog @ @ @ *]
     =/  pax  `path`(oust [(dec (lent t.wir)) 1] `(list @ta)`t.wir)
-    =.  synced  (~(del by synced) pax)
-    :_  state
-    :-  [%give %fact [/synced]~ %chat-hook-update !>([%initial synced])]
-    %.  ~
+    %.  (poke-chat-hook-action %remove pax)
     %-  slog
     :*  leaf+"chat-hook failed subscribe on {(spud pax)}"
         leaf+"stack trace:"
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -605,7 +605,7 @@
   ?~  saw  [~ state]
   ?+  wir  [~ state]
       [%store @ *]
-    [~ state(synced (~(del by synced) t.wir))]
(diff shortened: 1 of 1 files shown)


loganallenc:
It should be made for both cases, yes. This looks good. Weird to me that the subscription wasn’t already being removed by the watch-nack itself??

loganallenc:
@philipcmonk should the subscription be removed upon watch-nack?
@Fang- The subscription should be removed upon backlog nack as well

philipcmonk:
watch-nack should remove the subscription, yes, I'd be surprised if it doesn't.  @Fang-, do you mean you're removing other subscriptions as well as the one that was negatively acknowledged?

Fang-:
@loganallenc The watch-nack removes (or implies removal of) the subscription that's getting nacked only, but as far as `%remove` functionality is concerned, we always want the pair of `/mailbox` and `/backlog` subscriptions to be removed. (So @philipcmonk: that's exactly right.)

https://github.com/urbit/urbit/blob/1dd2b77222cf7447f94020b97d02a182129c1326/pkg/arvo/app/chat-hook.hoon#L361-L362

We're not strictly guaranteed to have both of those open, but we want to close whichever ones we have, and `%leave` on non-existing wire is safe.

Also realizing now that the `/0` there doesn't match the new "only get the backlog we missed" logic. @loganallenc I'm having trouble reasoning about backlog subscriptions, am I correct in guessing that just doing `?~(mailbox /0 /(scot %ud (lent envelopes.u.mailbox)))` there also is correct, or does that miss some cases? `/backlog` subscriptions get kicked after initial response right? In that case we shouldn't ever _need_ to `%leave` it outside of the watch-nack case, which implies we didn't receive backlog... I think.

jtobin:
(Haven't merged yet as there still look to be a few questions in @Fang-'s last comment.)

Fang-:
Those questions are not strictly related to this PR, as they already existed, and fall outside of the scope of "handle watch-nack properly".

That most recent commit certainly needs some touching up though. Not sure how that happened. Will force-push an updates commit message, then merge.

PR #2519, opened 2020-03-15
groups: enable sigil foreground color customization

pkova:
Bridge had spoiled me on this.

Also fixes a bug where if you type an uppercase letter A-F to the sigil color box you enter an annoying infinite request loop until you refresh.

### DIFF
--- a/pkg/arvo/app/contact-store.hoon
+++ b/pkg/arvo/app/contact-store.hoon
@@ -5,10 +5,15 @@
 +$  card  card:agent:gall
 +$  versioned-state
   $%  state-zero
+      state-one
   ==
 ::
 +$  state-zero
   $:  %0
+      rolodex=rolodex-old
+  ==
++$  state-one
+  $:  %1
       =rolodex
   ==
 +$  diff
@@ -16,7 +21,7 @@
   ==
 --
 ::
-=|  state-zero
+=|  state-one
 =*  state  -
 %-  agent:dbug
 ^-  agent:gall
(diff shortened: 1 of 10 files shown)


matildepark:
Oh my goodness.

matildepark:
We’ll have to adjust every app to make use of this too but just wow.

galenwp:
I feel like the right way to do this is to provide a palette, otherwise these are going to often look quite bad.

loganallenc:
I feel like we should offer maximum customizability to our users, and also provide the ability for a user to turn off display of colors entirely.

matildepark:
I agree that we want maximum customizability. I see no reason to kneecap this out of user distrust ... I believe people can arrive at beauty in ways I couldn’t fathom by leashing them. 

Turning off colours is interesting. I can imagine how to implement but we’d need a settings store, no?

reitraced:
oh wow i could finally do that black on pink ive wanted to do since os1 came out

eglaysher:
I will also speak in favor of this feature, though not being on interface, I have no idea about the implementation.

loganallenc:
So far the code looks quite reasonable and good. I'll do a full review once we finish the OS1 release, as we're trying to pause new features until we're well-pleased that the current functionality / user stories are smooth and work properly.

loganallenc:
Hi, we're going to include automatic logic for making the sigil white when on a dark-colored background and black when on a light-colored background. Closing this, as it complicates the contacts UI.

PR #2522, opened 2020-03-16
git surgery & doc updates re: new branching/release procedure

jtobin:
I recently discussed moving to a new branch/release process with some folks at Tlon.  The relevant excerpt from the internal email thread is as follows:

```
All -- I spent some time musing over the branching/release process
earlier and came up with the following scheme, which I think is superior
to what we're doing now, or what we were doing before.

It's basically a mix of the old master/hotfix/next model and the current
release tag-based model.  In particular, we maintain both 'master' and
'next' branches, and use ample release tags on master (i.e. as is
currently done on the release branch) to do away with the old hotfix
branch.

In short:

* maintain 'master' and 'next' branches.  contributors should usually
  begin work from master, unless there's a good reason not to, and
  should usually base their PRs on master.

* 'master' consists of known-good, blessed, ready-to-go work.  releases
  are made by tagging commits on master.  *at all times*, the diff
  between the latest release and master should constitute a valid OTA
  update.

* 'next' consists of work for the next era; i.e., that which cannot be
  deployed OTA.  breaches follow merges of 'next' to 'master'.
```

The only changes currently on `master` that prevent us from simply starting to do this are those reverted in this PR.  As per the [standard recommendation](https://github.com/git/git/blob/master/Documentation/howto/revert-a-faulty-merge.txt), I'll revert that revert on the new `next` branch.


jtobin:
I'm going to go ahead and merge this as it has already been informally ratified and will simplify near-term release work, but I just wanted to make a PR for added visibility (& to offload a test run to Travis).  Future releases can just be made by tagging commits on master.

/cc @Fang- @ixv @joemfb @philipcmonk

philipcmonk:
LGTM

joemfb:
Thanks, @jtobin.

PR #2524, opened 2020-03-16
chat-cli: absolute parity wrt "shortcodes"

### LINKED ISSUE #2503: Cannot type ~/ to ;join or ;leave chat in cli
**Describe the bug**
The format for public chat rooms changed and now include `~/` in front of `~ship-name/chat-topic`. In cli, typing the format `~ship-name/chat-topic` is allowed and typing the format `~/~ship-name/chat-topic` is not allowed.

**To Reproduce**
1. Open Terminal
2. Start urbit 
3. Type `ctrl` + `x` to go to `chat-cli` if not there already
4. Type `;join ~/~dopzod/urbit-help`
5. Terminal beeps after `~` and unable to type `/`

**Expected behavior**
Ablility to type `~/~ship-name/chat-topic` in `chat-cli` in order to `;join` or `;leave` a public chat.

**Screenshots**
n/a

 - OS: MacOS 10.15.3
 - Browser: chrome
 - Base hash: 0v8.rpvar.1t9h7.4es5r.irauk.niaqo.7do90.bpc1r.nihtd.ufdgt.8ehjl

**Smartphone (please complete the following information):**
n/a

Fang-:
Downgrades unmanaged chat paths to their uglified versions, as used by
web chat. Removes "group-based" indicators and makes them implicit in
shorter paths instead. Fixes #2503.

Includes a just-in-case addition of fixing configured `width` if it's not sane. Observed it being 0 on development ships a small amount of times. Couldn't reproduce, but want to add this as a precaution: should patch up anyone left in insane state.

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -185,7 +185,9 @@
   ::
   ^-  state-1
   ?-  -.u.old
-    %1  u.old(bound (~(gas by *(map target glyph)) ~(tap by bound.u.old)))
+      %1
+    =?  width.u.old  =(0 width.u.old)  80
+    u.old(bound (~(gas by *(map target glyph)) ~(tap by bound.u.old)))
   ::
       ?(~ ^)
     :-  %1
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -634,12 +634,19 @@
     ++  tag   |*(a=@tas (cold a (jest a)))  ::TODO  into stdlib
     ++  ship  ;~(pfix sig fed:ag)
     ++  path  ;~(pfix net ;~(plug urs:ab (easy ~)))  ::NOTE  short only, tmp
+    ::  +mang: un/managed indicator prefix
+    ::
+    ++  mang
+      ;~  pose
+        (cold %| (jest '~/'))
+        (cold %& (easy ~))
+      ==
     ::  +tarl: local target, as /path
     ::
     ++  tarl  (stag our-self path)
-    ::  +tarx: local target, maybe group
+    ::  +tarx: local target, maybe managed
     ::
-    ++  tarx  ;~(plug (fuss 'group ' '') path)
+    ++  tarx  ;~(plug mang path)
     ::  +tarp: sponsor target, as ^/path
     ::
     ++  tarp
@@ -650,7 +657,8 @@
     ++  targ
       ;~  pose
         ;~  plug
-          (fuss 'group ' '')
+          mang
+        ::
           ;~  pose
             tarl
             tarp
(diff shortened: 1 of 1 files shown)


matildepark:
I wouldn't feel comfortable reviewing this.

joemfb:
These new managed/unmanaged chat-identifier paths seem to be unwieldy, and a bad user-experience. Is it strictly necessary that the identifier itself always include this content? Wouldn't it be possible to have a fully-qualified chat identifier just be `~ship/name`, and look up whether it's associated with a group or not?

Fang-:
@joemfb Apparently this change is desired for consistency with web ui. Strong agree though, see also my comment here: https://github.com/urbit/docs/issues/820#issuecomment-599240722

vvisigoth:
Yes, we should change this, as this was a convention established before the
introduction of the metadata store, but let's support the existing UX while
we figure it out.

@Mark <mark@tlon.io> if you want to look into changing this convention in
order for this to go faster, please do. Otherwise it's going in the backlog.
—
~poldec-tonteg
http://urbit.org

belisarius222:
I could see this being resolved in a couple different directions, but one thing that seems strange to me is the idea of a chat that *is not* part of a group.  What does that even mean?  It suggests an incompletely defined ontology.  Are groups the foundation of the ontology, or does the foundation consist of individual "channels", such as notebooks, chats, and link pages, and groups are more like metadata associated with them?

The clearer our ontology is, the clearer we'll be able to make the UI.  I suspect what we want is for the group to be the base.  If you want a DM or other informally created chat channel, the system could create and automatically name a group for you, and then if you started that channel, you could rename it later if you want to reify it more.

I think this would have benefits for things like making UI affordances for switching between links, chat, and blog.  If you start DMing someone, the system could automatically add links to an associated links page for that group, for example, so later you could quickly find some link one of you shared.  I think this sort of thing makes more sense when all social activity is tied to a first-class group.

—
~rovnys-ricfer
https://urbit.org

matildepark:
Ted — see the docs conversation for the current thoughts around groups and chat. An unmanaged chat *is* a group, as in, it sets itself up in the group store, but it’s hidden from the Groups UI as a proper group — it’s an app-specific, ad hoc set of peers.

—
~haddef-sigwen
https://urbit.org

Fang-:
> if you want to look into changing this convention in order for this to go faster, please do

@vvisigoth I mean, I want to, but given how relatively deeply this path structure thing is integrated into logic across our current userspace stack I'm not sure it makes sense to do this as its own change, rather than as part of the broader refactoring of that stack. I'll definitely keep an eye out for easy-outs here, though, as I document the current state & think about the refactoring.

matildepark:
It’s only integrated in Chat’s web UI — Notebooks doesn’t use this and Links doesn’t have a join flow where you use shortcodes. One way or the other would’ve been established over the next two weeks for all apps.

—
~haddef-sigwen
https://urbit.org

loganallenc:
@vvisigoth These two path formats are not artifacts of previous decisions, but still serve purposes within the backend for differentiating between whether to request to sync down permissions and groups of the same name or not when joining a chat, for example.

PR #2529, opened 2020-03-17
links: add soundcloud embeds

matildepark:
- Gets the embed HTML from the Soundcloud [oembed API](https://developers.soundcloud.com/docs/oembed). 
- Sets the HTML in our state once we get it since we're waiting on a promise; passes it into the DOM as HTML directly as a fallback from the local `embed` variable (theoretically unsafe if we assume Soundcloud's oembed API gets attacked?).

In general I think what we really want is to stop manually regexing for these embeds and to get the Links back-end to get the metadata for us — so we can support anything and everything that returns an embed as part of OpenGraph.

For now this is ok. I think @vvisigoth just wants it in.

Screen Shot 2020-03-16 at 9 23 02 PM

### DIFF
--- a/pkg/interface/link/src/js/components/lib/link-detail-preview.js
+++ b/pkg/interface/link/src/js/components/lib/link-detail-preview.js
@@ -7,7 +7,8 @@ export class LinkPreview extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      timeSinceLinkPost: this.getTimeSinceLinkPost()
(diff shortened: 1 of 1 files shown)


matildepark:
True. Thinking about how to write that without those also creating multiple requests. Hmm

matildepark:
From testing this only fetches once, works in inter-link navigation and works on direct link access, so I think we're good?

PR #2534, opened 2020-03-17
ames: print num-fragments in packet verbage

philipcmonk:
Gives you a poor man's progress bar.  For example, to determine how much
of an OTA you've downloaded from your sponsor, run:

|ames-sift (sein:title our now our)
|ames-verb %rcv

and then to turn it off:

|ames-verb


vvisigoth:
wow
—
~poldec-tonteg
http://urbit.org

PR #2535, opened 2020-03-17
Aarch64 Release Build (~botter-nidnul)

benjamin-tlon:


ohAitch:
> The job exceeded the maximum log length, and has been terminated.

is the only issue with this CI? is the build not supposed to take half an hour?

joemfb:
Full cross-compilation takes way longer than CI limits -- we depend heavily on the build cache. #3374 is almost done, and will bring many improvements.

joemfb:
Note also that @botter-nidnul currently maintains Aarch64 binaries here: https://botter-nidnul.github.io/.

botter-nidnul:
There were other, broader issues with upstreaming my AArch64 patches. The big one was getting the new nixcrpkgs to work with OSX, but that [should be pretty easy to fix now](https://github.com/urbit/urbit/pull/2537#issuecomment-660536851) if someone wants to tackle it.

A new problem has emerged with king-haskell: it uses hsc2hs, and [hsc2hs doesn't like cross-compilation](https://github.com/haskell/hsc2hs#id1). I was only able to get a static urbit-king binary by [modifying haskell.nix](https://github.com/botter-nidnul/haskell.nix/tree/aarch64-musl) and building natively on AArch64. So I think CI for AArch64 king-haskell would require a separate AArch64 build server.

ohAitch:
Hm, can the hsc2hs bits be quarantined to a module that is built manually and then cached, or is the relevant ffi surface area constantly in flux? Heck, build it under qemu, "creating, compiling, and running a C program" doesn't sound lethally expensive if all it's doing is generating a sane amount of Haskell source.

joemfb:
#3374 dispenses with nixcrpkgs, and builds MacOS binaries on a MacOS instance. We should revisit this once it's in.

@ohAitch, your comments seem reasonable to me. @pilfer-pandex, @eglaysher?

eglaysher:
I'm not sure what you're asking, Joe, but I'd love to have an aarch64 build covered on the bots, and I assume we could just get a hosted raspberry pi to act as a build server, like we have a mac mini in the cloud to be a nix darwin build server.

joemfb:
To the degree that it's practical, it'd be good for FFI-stuff in king-haskell to be amenable to cross-compilation. So I'm just forwarding on the report that it's not.

brendanhay:
The use of `hsc2hs` arises in what appears to be a copy-paste of the [lmdb](https://hackage.haskell.org/package/lmdb) package from hackage with modifications to the build options and `lmdb` C library verbatim in `cbits`. I assume all of this was in aid of static/cross compilation originally.

1. The `lmdb` package has only 2 total uses of `#{let}` and `#{def}` which are incompatible with `hsc2hs --cross-compilation` per the [README](https://github.com/haskell/hsc2hs#cross-compilation). Lowest effort might be trying out the `--cross-compilation` flag.
2. Write a `lmdb` wrapper that is more amenable to cross compilation.
3. Add an `aarch64` build agent to go along with the `linux64` and `darwin64` agents. I actually think this is a good mid term target as if we have `aarch64` build derivations in tree - we officially support it - so it should be part of CI. 

My preference would be to throw $$ at the problem and obtain a dedicated machine for any given architecture. The high-cardinality of a mixture of static vs shared builds, cross-compiation targets, and so on quickly becomes a mess. The derivations for simply building on different architectures are complex enough without then having to differentiate build vs host platforms, along with the various whack `buildInputs` vs `nativeBuildInputs` and similar problems, IMO.

tacryt-socryp:
Any updates? @brendanhay @joemfb

Issue #2067, opened 2019-12-10
intermittent memory mapping errors on AArch64


botter-nidnul:
I used nix to build urbit for my Chromebook (inside Crostini running Debian 9) and virtual private server (Debian 10) which are both Arm64 (uname -m outputs "aarch64"). Everything else seems fine, but on both systems there are occasional memory related errors when booting piers which are resolved by simply booting the pier again. 

There are two types of errors I've seen, the first more common than the second.

1st:
```
~
urbit 0.10.1
boot: home is /home/urbit/botter-nidnul
loom: mapped 2048MB
urbit: malloc.c:2401: sysmalloc: Assertion `(old_top == initial_top (av) && old_size == 0) || ((unsigned long) (old_size) >= MINSIZE && prev_inuse (old_top) && ((unsigned long) old_end & (pagesize - 1)) == 0)' failed.
Aborted
```
2nd:
```
~
urbit 0.10.1
boot: home is /home/urbit/botter-nidnul
loom: mapped 2048MB
lite: arvo formula 7c50e147
lite: core 6c9faaf4
lite: final state 6c9faaf4
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/255.606.784
Assertion '!"loom: corrupt"' failed in noun/allocate.c:123

bail: oops
bailing out
realloc(): invalid pointer
pier: work error: end of file
pier: work exit: status 0, signal 6
```

botter-nidnul:
Just got a 3rd type:

```
~
urbit 0.10.1
boot: home is /home/urbit/botter-nidnul
loom: mapped 2048MB
lite: arvo formula 7c50e147
lite: core 6c9faaf4
lite: final state 6c9faaf4
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/256.671.744
boot: installed 262 jets
address 0x600000006 out of loom!
loom: [0x36000000 : 0xb6000000)
Assertion '0' failed in noun/events.c:129

bail: oops
bailing out
pier: work error: end of file
pier: work exit: status 0, signal 11
```

botter-nidnul:
I've been playing around with my static AArch64 binaries (dozens of boots on Android, fewer on my Raspberry Pi and ThunderX VPS) and this just doesn't appear to happen anymore. I suspect it's because the dependencies for building these static binaries are different (musl) and more recent (gcc) and whatever bug was causing this platform specific issue has been patched.

martijnbolt:
[...]

PR #2537, opened 2020-03-17
[WIP] Update nixpkgs and nixcrpkgs.

benjamin-tlon:
I previously just copied the file from `nixcrpkgs` to avoid increasing the size of our repo, but in practice this is no good. There is no sane way to pull in upstream changes with that approach.

So, instead, this pulls in the latest `nixcrpkgs` using `git subtree`, reapplies our local changes, and gets things working again with the updated dependencies.

```
$ git rm -r nix/nixcrpkgs
COMMIT
$ git subtree add --prefix nix/nixcrpkgs https://github.com/pololu/nixcrpkgs.git master
# That happened at '0ccf0f7b716fcc5d750ae59017addeca8ba0d3b7'
...
```


brendanhay:
Having run into the same issue I think it'd just be better to fork `nixcrpkgs` and then work on getting modularisation changes pushed back upstream from our fork. Is the subtree really necessary?

benjamin-tlon:
We have a fork for nixpkgs.

I don't really care one way or the other @ subtree.

I didn't even consider getting changes pushed upstream, but that would definitely help with maintenance if possible.

benjamin-tlon:
This is not related to CI being broken AFAIK.

benjamin-tlon:
This is not an easy change, and it wont be done for some time unless someone is down to throw a bunch of days/weeks at it.

There are a few possible solutions:

1. Throw a bunch of time at getting the new nixcrpkgs working on OSX. This probably means manually creating the OSX SDK tarball (as per nixcrpkgs instructions) and finding some way to make it available to our build infrastructure.

  - Once that's done. Continue to get everything working again with the new nixcrpkgs.
  - Finally, sort out the right way to include nixcrpkgs in our repo and untangle the current mess.
  
2. Stop cross-compiling OSX builds. This will mean writing an entirely different build system that statically links all of our dependencies, but doesn't statically link against OSX system dependencies.

3. Stop releasing statically linked binaries at all, and require all users to use Nix. This would make everything much easier and faster, but would require all users to have nix. Probably not viable for the time being.

botter-nidnul:
> 3. Stop releasing statically linked binaries at all, and require all users to use Nix. This would would remove a huge amount of complexity and make everything much faster. However, all users would need to have nix installed in order to get Urbit, and would kill our ARM support (I think). This is probably not viable for the time being, but should be discussed.

This wouldn't be very disruptive to Urbit on AArch64. Nix fully supports 64-bit ARM, and the install process is exactly the same as x86 architectures. Nix even works [on Android](https://github.com/t184256/nix-on-droid) thanks to PRoot. The only thing this would kill would be [my ugly non-PRoot termux hack](https://github.com/botter-nidnul/urbit/tree/termux-v0.10.7) which is going to be [killed by Google's security paranoia soon anyway](https://github.com/termux/termux-packages/wiki/Termux-and-Android-10). But I think the recent linux binary performance improvements make the Android non-PRoot version superfluous, so it's no real loss.

I do have two caveats: 

First: there is [a bug I've seen with the dynamic builds for AArch64](https://github.com/urbit/urbit/issues/2067#issuecomment-596423781) but I think [merging this pull request](https://github.com/urbit/urbit/pull/2991) would fix it.

Second: king-haskell is going to be a pain to get statically building for AArch64. I'm not sure it can be done with static-haskell-nix without more modifications than it would be worth, it might be better to switch to [haskell.nix](https://input-output-hk.github.io/haskell.nix/motivation/). I'll continue to play around with it, and see what I'm able to get working.

Just wanted to let you know Urbit's status with regard to Nix on ARM, I'm not taking a stance on stopping release of statically linked binaries.

brendanhay:
> Second: king-haskell is going to be a pain to get statically building for AArch64. I'm not sure it can be done with static-haskell-nix without more modifications than it would be worth, it might be better to switch to haskell.nix. I'll continue to play around with it, and see what I'm able to get working.

FWIW we're actually using `haskell.nix` elsewhere, internally. It's even building `king-haskell` static binaries for use inside containers. Having only used it for AMD64 targets - can I take that statement as "It'd be easier for me to maintain ARM support if you used `haskell.nix`"?

botter-nidnul:
>"It'd be easier for me to maintain ARM support if you used haskell.nix"?

Yes, I think so. Static ARM binaries should just be another type of cross compilation environment that can be inherited, with something like `pkgsNative.lib.systems.examples.aarch64-multiplatform-musl`. 

(The big advantage of haskell.nix over static-haskell-nix is that it will work with versions of ghc 8.8.1 and above, when AArch64 became a fully supported platform and some long standing platform specific bugs with ghc on ARM disappeared. Also, it will be easily to compile for AArch64 from x86_64, since static-haskell-nix doesn't play well with cross compilation)

botter-nidnul:
> There are a few possible solutions:
> 
> 1. Throw a bunch of time at getting the new nixcrpkgs working on OSX. This probably means manually creating the OSX SDK tarball (as per nixcrpkgs instructions) and finding some way to make it available to our build infrastructure. There are probably copyright implications.

@benjamin-tlon I was playing around on github today, and I noticed that [nixcrpkgs seems to have been updated](https://github.com/pololu/nixcrpkgs/commit/f6981c6d79ec3879926d37ca352f4b7c3372e9bc) with exactly this in mind. Should be possible to pull the OSX SDK tarballs in from something like https://github.com/phracker/MacOSX-SDKs now?

I'd volunteer to try to get this working with Urbit, but I have no Apple hardware to test it on because I'm a second generation anti-Apple bigot and I don't want to end the family tradition. (Though I am tempted to try building Urbit to run on a jailbroken iphone - I think that should be possible with an updated nixcrpkgs and an updated OSX SDK)

matildepark:
For the record @botter-nidnul, I’m not sure @benjamin-tlon is likely to respond as he is no longer with us. It would be good to have someone own this PR, but it’s a big set of changes ...

tacryt-socryp:
Is this PR still valid? @joemfb @brendanhay

joemfb:
We're no longer using `nixcrpkgs`, so this PR is obsolete.

Issue #2542, opened 2020-03-18
Large text in creating Publishing post breaks out of container … (Firefox)


fakezod:
**Describe the bug**
Writing a larger than input container height text in `Publishing` -> `New Post` as well as pasting such a text will result in the input container breaking out of the surrounding frame. As far as I’m concerned, this bug does not happen in Chromium.

Sidenote: How about some placeholder in the body text input field?

**Screenshot**
<img width="967" alt="Screenshot 2020-03-18 at 13 35 18" src="https://user-images.githubusercontent.com/62216503/76961290-681e3680-691d-11ea-8118-672704972b6f.png">

**Screencast**
![publishing-in-firefox](https://user-images.githubusercontent.com/62216503/76961159-25f4f500-691d-11ea-8392-6c08325c2ec8.gif)

**To Reproduce**
1. Go to `Publishing`
2. Click on `New Post`
3. Write lots of text …

**Expected behavior**
The input field should not break out of the surrounding frame/border.

**Desktop:**
 - macOS `10.14.6`
 - Firefox `v74`
 - base hash `0vf.djnhb.aguv9.3u4hg.54bj5.4e6mi.abspq.99s36.p5fro.qmv7l.65gau`

PR #2544, opened 2020-03-18
publish: stop post input area from overflowing parent container in firefox

ixv:
Fixes https://github.com/urbit/urbit/issues/2542

For some reason calc behaves differently on firefox, so I stopped using it. Tested on Firefox, Chrome, Safari and Brave.

### DIFF
--- a/pkg/interface/publish/src/css/custom.css
+++ b/pkg/interface/publish/src/css/custom.css
@@ -93,7 +93,14 @@ a {
 
 .NewPost {
   width: 100%;
-  height: calc(100% - 115px);
+  height: calc(100vh - 174px);
+  display: flex;
+  padding-top: 8px;
+}
+
+.EditPost {
+  width: 100%;
+  height: calc(100vh - 120px);
   display: flex;
   padding-top: 8px;
 }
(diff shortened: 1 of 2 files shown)


matildepark:
Was it always doing this? (Probably we need to pass `flex-auto` to children of `flex` parents to prevent overflow of parent containers.)

image

pkg/interface/publish/src/css/custom.css
```diff
@@ -93,9 +93,20 @@ a {
 
 .NewPost {
   width: 100%;
-  height: calc(100% - 115px);
   display: flex;
   padding-top: 8px;
+  top: 115px;
+  bottom: 0px;
+  position: absolute;
```
matildepark: Actually, it's probably because of this, lol

matildepark:
Doing this with absolute positioning I think also ends up regressing this responsive check: 
image

ixv:
Good catch, fixed. Turns out we don't need to do position absolute after all. While firefox doesn't like calc with mixed % and px, its fine with mixed vh and px, I guess.

PR #2546, opened 2020-03-18
Chat: move chat to existing group

Fang-:
Expands on the "groupify" functionality added in #2480 by allowing you to merge invite-only chats into an existing group.

UI/copy here might need some light touching up. Currently just display the group search, lets you continue without selecting a group (old behavior, creates a new group based on the chat), or shows you the "inclusive" toggle if you do select a target group (new behavior, merges chat into existing group).

convert

### DIFF
--- a/pkg/interface/chat/src/js/components/settings.js
+++ b/pkg/interface/chat/src/js/components/settings.js
@@ -181,7 +181,6 @@ export class SettingsScreen extends Component {
     const { props, state } = this;
 
     const chatOwner = (deSig(props.match.params.ship) === window.ship);
-    console.log(chatOwner, props.match.params.ship, window.ship);
 
     const ownedUnmanagedVillage =
       chatOwner &&
@@ -192,9 +191,8 @@ export class SettingsScreen extends Component {
       return null;
     } else {
       let inclusiveToggle = <div/>
-      console.log('tg', state.targetGroup);
       if (state.targetGroup) {
-        //TODO toggle component
+        //TODO toggle component into /lib
         let inclusiveClasses = state.inclusive
           ? "relative checked bg-green2 br3 h1 toggle v-mid z-0"
           : "relative bg-gray4 bg-gray1-d br3 h1 toggle v-mid z-0";
@@ -207,7 +205,7 @@ export class SettingsScreen extends Component {
               onChange={this.changeInclusive}
             />
             <span className="dib f9 white-d inter ml3">
-              Inclusive
+              Add all members to group
             </span>
             <p className="f9 gray2 pt1" style={{ paddingLeft: 40 }}>
               Add chat members to the group if they aren't in it yet
--- a/pkg/interface/chat/src/js/components/settings.js
+++ b/pkg/interface/chat/src/js/components/settings.js
@@ -5,6 +5,7 @@ import { Route, Link } from "react-router-dom";
 
 
 import { ChatTabBar } from '/components/lib/chat-tabbar';
+import { InviteSearch } from '/components/lib/invite-search';
 import SidebarSwitcher from './lib/icons/icon-sidebar-switch';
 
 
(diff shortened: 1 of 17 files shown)


matildepark:
If you're going to rewrite invite search, you should update all the copies of it across all apps.

pkg/interface/chat/src/js/components/settings.js
```diff
@@ -171,15 +191,56 @@ export class SettingsScreen extends Component {
...
+        inclusiveToggle = (
+          <div className="mt4">
+            <input
+              type="checkbox"
+              style={{ WebkitAppearance: "none", width: 28 }}
+              className={inclusiveClasses}
+              onChange={this.changeInclusive}
+            />
+            <span className="dib f9 white-d inter ml3">
+              Inclusive
```
matildepark: `Add all members to group` and keep the body text and this is fine
```suggestion
              Add all members to group
```

pkg/interface/chat/src/js/components/settings.js
```diff
@@ -171,15 +191,56 @@ export class SettingsScreen extends Component {
     if (!ownedUnmanagedVillage) {
       return null;
     } else {
+      let inclusiveToggle = <div/>
+      console.log('tg', state.targetGroup);
+      if (state.targetGroup) {
+        //TODO toggle component
```
matildepark: seems like it's here

ixv:
I tried the case where members of the chat are not in the group, and the inclusive toggle is on. The ship gets invited to the group successfully, but the old chat persists with the "your ship has been disconnected" notice. Joining the new chat manually works fine, but then trying to leave the old chat manually gives me this: 
image

Fang-:
> the old chat persists with the "your ship has been disconnected" notice

I don't think we can (or should) push deletion in the current model. Maybe a `%moved` diff for cases like these is appropriate, but seems way out of scope.

Should we maybe send invites for the chat, in addition to invites for the group?

pkg/interface/chat/src/js/components/settings.js
Fang-: I meant like a `component/lib/toggle.js`. Will clarify.

ixv:
Yeah, I don't have a huge problem with this. If you fix the issue with leaving those old chats, I'm fine with this going in.

Fang-:
> If you're going to rewrite invite search, you should update all the copies of it across all apps.

Did we decide not to have a `/pkg/interface/lib` at some point, or have we just not gotten around to it yet?

@ixv this _seems_ to be a chat/metadata bug, where chat deletion pushes deletion of everything (metadata, group) but the chat itself. Those things are relied upon during local deletion, and so it crashes. I'll investigate a bit further and try to find an appropriate fix here...

matildepark:
I think the goal is to use an npm package for that and it’s what’s in progress among the rest of interface right now. Also I’m unsure we can really reach out of the directory of the npm package like that (since we have package.json in all the apps individually). Maybe if we refactor it all into one big one? And be better with gulp? But don’t want to step on @jtobin’s jtoes...

Fang-:
Correction: we never have the group, but we do have metadata, with a pointer to the group. I think this means the solution is to just, if the metadata isn't there, forego deleting the (non-existent) associations etc.

@ixv Please look this over carefully. Maybe the Real Solution is fixing some other problem, but I don't think this change does any harm, and at least supports deletion for the current situation.

ixv:
Looks good, and I tested it and it works fine.

jtobin:
> I think the goal is to use an npm package for that and it’s what’s in progress among the rest of interface right now. Also I’m unsure we can really reach out of the directory of the npm package like that (since we have package.json in all the apps individually). Maybe if we refactor it all into one big one? And be better with gulp? But don’t want to step on @jtobin’s jtoes...

I think it does make sense to refactor them into one big 'urbit-landscape' package or whatever.  I might be able to just go ahead and do this since I need to update #2349 for the post-OS1 stuff anyway.

PR #2548, opened 2020-03-18
groups: delete group members without shared metadata; delete groups

### LINKED ISSUE #2483: Delete Groups functionality
You also can't currently remove members from a group until they've shared their contact details. These are frontend issues, I think.

matildepark:
Closes #2483.

What it says on the tin ... the behaviour when you create a group in another app and then delete it in groups is neat — it seems to just seamlessly keep working, but retains its non-sig-prepended path.

### DIFF
--- a/pkg/interface/groups/src/js/api.js
+++ b/pkg/interface/groups/src/js/api.js
@@ -85,6 +85,12 @@ class UrbitApi {
     });
   }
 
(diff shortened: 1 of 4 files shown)


loganallenc:
Looks good.

galenwp:
Thanks!

PR #2562, opened 2020-03-19
interface: add eslint to interface and lint `launch` and `chat`

### LINKED ISSUE #2555: Adding .eslint or .prettier to `pkg/interface`
Hey,

Wondering if there's interest in adding an `.eslint` and/or `.prettier` file to `pkg/interface`.
The code style seems to vary quite a lot and there are some line breaks here and there that could be removed.
More consistentcy overall would be a good thing I think, can look into adding it if there's interest!

nijynot:
Adds `eslint`, `eslint-plugin-react` and `babel-eslint` to the root of `pkg/interface`.
You can find the eslint rules in `pkg/interface/.eslintrc.json`.

As of now, only `interface/launch` has been linted.
If this looks good, I can continue on the other urbit-apps modules.

Resolves #2555.

*Update:*
The .eslintrc.json files is partially inspired by this file https://github.com/handshake-org/hsd/blob/master/.eslintrc.json and airbnb's react eslint settings.

Only linted the `chat` and the `launch` apps. Other apps are untouched for now to prevent an even bigger PR. Have tested building the PR and commiting it to a fake planet, and does seem to work correctly.

No logic should have been touched, only code style should have been changed. All linting has been done manually by using Atom and some plugins to display visual cues when a lint error occurs.

To run linting by a command line, do this
```
$ cd pkg/interface
$ npm install
$ npm run lint
```
, also added linting tips to the CONTRIBUTING.md file in the `interface` folder.

### DIFF
--- a/pkg/interface/.eslintrc.json
+++ b/pkg/interface/.eslintrc.json
@@ -1,0 +1,147 @@
+{
+  "env": {
+    "browser": true,
+    "es6": true,
+    "node": true
+  },
+  "extends": [
+    "eslint:recommended",
+    "plugin:react/recommended"
+  ],
+  "settings": {
+    "react": {
+      "version": "^16.5.2"
+    }
+  },
+  "parser": "babel-eslint",
+  "parserOptions": {
+    "ecmaVersion": 10,
+    "requireConfigFile": false,
+    "sourceType": "module"
+  },
+  "root": true,
+  "rules": {
+    "array-bracket-spacing": ["error", "never"],
+    "arrow-parens": [
+      "error",
+      "as-needed",
+      {
+        "requireForBlockBody": true
+      }
+    ],
+    "arrow-spacing": "error",
+    "block-spacing": ["error", "always"],
+    "brace-style": ["error", "1tbs"],
+    "camelcase": [
+      "error",
+      {
+        "properties": "never"
+      }
+    ],
+    "comma-dangle": ["error", "never"],
+    "eol-last": ["error", "always"],
+    "func-name-matching": "error",
+    "indent": [
+      "off",
+      2,
+      {
+        "ArrayExpression": "off",
+        "SwitchCase": 1,
+        "CallExpression": {
+          "arguments": "off"
+        },
+        "FunctionDeclaration": {
+          "parameters": "off"
+        },
+        "FunctionExpression": {
+          "parameters": "off"
+        },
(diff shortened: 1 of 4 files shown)


matildepark:
This looks fine to me but I have never used a repository-wide linter.

nijynot:
Rebased and ready for review.

nijynot:
The longer this is open, the more rebasing will need to be done. 😬
Is there something blocking this PR or is there something else on your mind?

If it's too risky of a PR as there's a lot of lines changed and could have unintended side-effects, I could close this PR. For example, Bitcoin Core has a rule where no style change PRs are ever approved because it has potential unintended side effects, so I do understand if this should be closed.

An alternative that I know could be done is to *only* merge the .eslintrc files and the package.json files `pkg/interface`. This keeps the unintended side-effects to a minimum and style changes can be done with time, instead of a single big PR. Let me know what you think.

matildepark:
It's just that we're in the midst of a lot of work getting the OS1 release out (as opposed to our released work in alpha) and rule of thumb is to not rebase in the middle of a release sprint. I would say ... let's slate this for two weeks from now, *then* rebase and we'll evaluate?

nijynot:
@matildepark Thanks for the update, yeah that sounds good!
Will wait for a better timing when it fits you.

matildepark:
Hi! Picking this up.

So instead of playing rebase tennis, I think you should do this:

> An alternative that to this PR would be to only merge the .eslintrc files and the package.json files pkg/interface (eslint and all of its dependencies). This keeps the unintended side-effects to a minimum and style changes can be done with time instead of a single big PR. Let me know what you think.

I can't guarantee that no activity will happen on the JS files this week; there's still some cleanup work, so, putting the linter in place and letting it be organically integrated (and/or deliberately accelerated on our part) seems ideal.

nijynot:
> putting the linter in place and letting it be organically integrated (and/or deliberately accelerated on our part) seems ideal.

Agreed, have updated the PR to only include the lint rules, updated docs, and the required node_modules in `pkg/interface`.

matildepark:
Oddly in VSCode it complains about not having the eslint-plugin-react module but if I run eslint in pkg/interface from Terminal it’s fine. Probably just need it to not use the global one … or maybe the other way around.

— 
~haddef-sigwen
https://urbit.org

PR #2564, opened 2020-03-19
links: adjust sigil size to 38

matildepark:
In links, the `pt1` used to separate the `LinkItem` matter — title and author/date/comments — stretches the sigil, being sized at 36, to 36x38. It results in this:

image

I decided to just size it up to 38px instead of futzing with the line-heights of the matter.

image


urcades:
stamp

PR #2566, opened 2020-03-19
chat fe: disable advanced groupify

Fang-:
Introduced in #2546, the new functionality seems able to induce weird
behavior causing messages to be processed twice. Disabling this
functionality on the frontend until that has been fixed.

image


pkg/interface/chat/src/js/components/settings.js
```diff
@@ -222,6 +222,7 @@ export class SettingsScreen extends Component {
               Convert this chat into a group with associated chat, or select a
               group to add this chat to.
```
ixv:
```suggestion
              Convert this chat into a group with associated chat.
```

should change the copy here

PR #2571, opened 2020-03-20
@p autocomplete for chat

liam-fitzgerald:
Autocompletes the @ps of moons and above when tab is hit. If no unambiguous completion is found, matches are shown to the user, which they can then tab through and hit enter to confirm.
Fetches @ps from recent messages to the channel and prioritises more recent message senders.

Demo:
Screen Shot 2020-03-20 at 3 22 09 pm
Screen Shot 2020-03-20 at 3 21 56 pm
Screen Shot 2020-03-20 at 3 20 56 pm
Screen Shot 2020-03-20 at 3 20 32 pm


liam-fitzgerald:
cc: @Fang-

matildepark:
Empty chatrooms get you `undefined` hehe

image

Lemme spin up a few more to play with

matildepark:
- OK, even when I have a member in my chatroom the autocomplete never functions, it just replaces what I was typing with `undefined`, no prompt above.
- I was considering the behaviour of this over breakfast; at first I thought, "perhaps this should function a bit more like Discord, so no tab, it just opens the prompt on `~`," but I realised we were heading toward greater parity between chat-cli and chat-fe's keybinds, so matching tab was superior.
- However I do think we probably want to integrate nicknames into here a la invite search. Opinions? @urcades

matildepark:
Oh, because it goes off of recent messages, not the entire group. OK. Once I typed a message I showed up I think it should probably have both, but prioritise the former?

Hitting `tab` without any `~` gives an error — we probably want to only proc `tab` if we're trying to autocomplete.

This is dope in any case.

urcades:
Ah, this is beautiful — I was hoping we'd get around to this in a formally-integrated manner at some point, but the sooner we get this in the better. To @matildepark 's point, we definitely should be affording for the ability to search by both @p and self-designated name.

I have some further out gestures for an interface of this type in the following mocks:

Sig-Completion

From the looks of it, the implementation above captures 80% of these mocks, but I'd like to keep in mind we should think about more complex information eventually being sourced here, such as:

1. @p
2. Self-designated name
3. Programmatic status (online or offline)
4. AIM-style self-written status
5. etc.

This is sick!

urcades:
Also, I think your hunches towards the autocomplete/keyboard interactivity is correct, @matildepark – this should more or less match what's found in other applications with robust user-matching/filtering: tab-completion, up/down arrow navigation, the works.

liam-fitzgerald:
@urcades what should happen if the background color matches the users display color?

liam-fitzgerald:
Displays & matches on nicknames. Nickname search is slightly fuzzier than patp search (see below). up/down and tab/shift+tab can navigate.
Screen Shot 2020-03-21 at 2 08 47 pm


Screen Shot 2020-03-21 at 2 09 06 pm

urcades:
@liam-fitzgerald apologies for not getting back to you in time, but:

1. Regarding user's color choice: There are likely a few ways to accomplish this, but my gut check had been to "darken" or "desaturate" each color by a set amount (this is best expressed by the HSL color model), or to add a pure black drop shadow with zero spread to each name, and to make the opacity of the text itself %80 such that even lighter colors would appear a readable 'darker' color.

A visual examples of the intended effect:

Screen Shot 2020-03-21 at 5 50 48 PM

Each of the colored names themselves are set to 80% opacity here, with a 100% opacity black 'shadowing' name located directly 'under' them.

If this effect ends up being too hard/annoying to reasonably build, we should set all names to plain ol' black.

urcades:
The implementation in the above comments are amazing though — the pop-overs look gorgeous.

matildepark:
Is this ready for another review @liam-fitzgerald ?

liam-fitzgerald:
Playing with the user supplied color contrast. Tried a bunch of different techniques, ended up on this one because imo it works nicely across dark/light modes and is essentially CSS only. 

It's a black drop shadow with transparency like @urcades suggested but applies a contrast and saturate filter, to help w/ dark mode legibility. 

Ready for new review! @matildepark 
btw, there's an issue with the `.hover-bg-gray1-d` class in that it sets the `color` and not the `background-color`. Is the unminified css kept somewhere?

Screen Shot 2020-03-24 at 10 45 15 am

Screen Shot 2020-03-24 at 10 45 26 am

matildepark:
Oof, you’re right. It’s here:

https://github.com/urbit/urbit/blob/master/pkg/interface/chat/src/css/custom.css#L279

Let me pull this PR up but feel free to patch this too …

— 
~haddef-sigwen
https://urbit.org

PR #2572, opened 2020-03-20
various: correctly validate @p

### LINKED ISSUE #2556: invite search: non-hyphenated ships register as valid
Screen Shot 2020-03-19 at 11 07 03 AM

Screen Shot 2020-03-19 at 11 07 06 AM

liam-fitzgerald:
Checks that the @p is correctly formatted with hep. 
Fixes #2556


matildepark:
image

Except not so legit in practice. Sorry. -_-;

liam-fitzgerald:
> image
> 
> Except not so legit in practice. Sorry. -_-;

Which app is this? I'm unable to repro this

jtobin:
Hmm.  I think it'd be better to fix this properly in urbit-ob (and resolve the long-standing urbit/urbit-ob#17).

liam-fitzgerald:
will do!

PR #2580, opened 2020-03-20
publish: add pending state for comments

matildepark:
- Much like the chat variant, shows your last pending comment in faded opacity below; and still allows you to type another one while waiting for it. Posting that second comment, however, will still be disabled while you're waiting.

Performance question: To get it to be a seamless update (not popping in and out), I handle it in `componentDidUpdate`, but I can't get it to only do this when the latest comment has changed; if I add another `[0]` to that `Object.keys()`, it never processes — even though the strings ostensibly are different between the previous latest comment and our current; if I just compare `Object.keys()`, it always processes, on every keystroke, because it's comparing arrays, and of course the arrays are different assignments.

Likewise, handling pending comments in our Comments component has the side effect of losing the currently pending comment when we navigate away (though ostensibly it will post and be shown within a few seconds anyway). I assumed from a UX standpoint that people, unlike chat, won't be juggling tons of pending comments — otherwise this has to be rewritten into the reducer as Chat does. I tried the simpler approach first.

### DIFF
--- a/pkg/interface/publish/src/js/components/lib/comment-item.js
+++ b/pkg/interface/publish/src/js/components/lib/comment-item.js
@@ -29,6 +29,7 @@ export class CommentItem extends Component {
     });
   }
   render() {
+    let pending = !!this.props.pending ? "o-60" : "";
     let commentData = this.props.comment[Object.keys(this.props.comment)[0]];
     let content = commentData.content.split("\n").map((line, i)=> {
       return (
@@ -55,7 +56,7 @@ export class CommentItem extends Component {
     }
 
     return (
-      <div>
+      <div className={pending}>
         <div className="flex mv3 bg-white bg-gray0-d">
         <Sigil
           ship={commentData.author}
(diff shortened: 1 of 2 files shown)


ixv:
I get this error when I try to make a new note or when I navigate to an old note: 
image

matildepark:
Eesh, will fix.

— 
~haddef-sigwen
https://urbit.org

matildepark:
Should be okay now …

— 
~haddef-sigwen
https://urbit.org

ixv:
Works okay, the only nitpick I have is that the pending comment date says "50 years ago" 
image

matildepark:
God, I am just off today. Let me see.

— 
~haddef-sigwen
https://urbit.org

matildepark:
Seemed like since I divided by 1000 (to get seconds, not miliseconds) it might’ve already compensated somewhere; so it was  displaying “the time in seconds, divided by 1000, from now”? It says “just now” now.

— 
~haddef-sigwen
https://urbit.org

PR #2582, opened 2020-03-20
chat-view: limit initial scrollback to 25 msgs

Fang-:
Provided "load backlog on-scroll" actually works, you don't really need
all that much on initial load.

### DIFF
--- a/pkg/arvo/app/chat-view.hoon
+++ b/pkg/arvo/app/chat-view.hoon
@@ -109,6 +109,8 @@
       [[%give %fact ~ %json !>(*json)]~ this]
     (on-watch:def path)
     ::
+    ++  message-limit  25
+    ::
     ++  truncated-inbox-scry
       ^-  inbox
       =/  =inbox  .^(inbox %gx /=chat-store/(scot %da now.bol)/all/noun)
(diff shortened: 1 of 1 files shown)


loganallenc:
Loading backlog on-scroll doesn’t currently work, so this shouldn’t be merged yet.

Fang-:
This came up during the last two standups. It currently takes me ~2.5 minutes to load the chat page. Pretty sure we want to take this trade-off for now. @vvisigoth can you confirm?

loganallenc:
Have you confirmed a measurable perf improvement?

Fang-:
Would have applied this to ~palfun, but it has clay mount/commit issues, so can't confirm the change there.

Tried this out on a fakezod (which generally have a faster baseline than old, live ships). Impact seems negligible if you only have a chat or two, but once you start adding many, the difference becomes more significant. Measured (with very crude tools) 4 seconds -> ~1 second load times, for 15 chats with >100 messages in backlog.

Fang-:
Resolved sync issues on ~palfun and tried this out, brought it down from ~2.5 minutes to just over a minute.

loganallenc:
Okay good. this should be good to go.

galenwp:
The only problem with this is that when the scrollback breaks you're left with v little context.

Fang-:
Fix for that scrollback bug Real Soon Now™! See #2606.

Issue #2586, opened 2020-03-21
Clogged Cue


custom-jonathan:
**Describe the bug**
Can't delete notebook posts, OTA not working, and other problems I think are stemming from a clay issue.

**To Reproduce**
I wasn't getting an OTA, and notebooks weren't accepting new posts, so I ran `|cancel` which immediately triggered `<<behn>>` as outlined in #2574. Once I got it working the limbo posts came through, so I didn't want a bunch of garbage posts hanging out in the notebook. So I connected via FTP, and then tried deleting via the FTP fileviewer, and then went to the client and tried `|commit %home` but the posts would still be there. I then tried to navigate directly via dojo. `=dir %/app/publish/notebooks/the-library/` and then try `|rm %/post-name/udon` and it would say it's removed. But then I'd run `+ls %` and the post would still be there. I navigated up to home with `=dir` and then came back down to the notebook and the post was still there.  Then I thought about trying to unmount `%home` and then remount it. So I ran `|unmount %home` and the FTP viewer showed an empty "minder-folden" folder. Then I went back to dojo and ran `|mount %home` and got this error: 

```
> |mount %home                                                                  |
  /~minder-folden/home/~2020.3.16..15.26.38..05ed/sys/vane/ford                 |
  <[4.842 18].[4.842 40]>                                                       |
nest-fail                                                                       |
ford: %slit failed:                                                             |
- want                                                                          |
{ { now/@da                                                                     |
    eny/@uvJ                                                                    |
    bec/{p/@p q/@tas r/?({$da p/@da} {$tas p/@tas} {$ud p/@ud})}                |
  }                                                                             |
  {pax// pot/?($~ {v/@tas $~})}                                                 |
  $~                                                                            |
}                                                                               |
- have                                                                          |
{ { now/@da                                                                     |
    eny/@uvJ                                                                    |
    bec/{p/@p q/@tas r/?({$da p/@da} {$tas p/@tas} {$ud p/@ud})}                |
  }                                                                             |
  {$home $~}                                                                    |
  $~                                                                            |
}                                                                               |
ford: %call failed type calculation                                             |
~minder-folden:dojo>    
```

So then I wanted to see if maybe something was stuck, so I ran `|cancel` and I get this error at the top:

```
[%clay-idle cue-length=35]
> |cancel
>=                                                                                                    crud: %wake event failed
bail: exit
take: failed
  /~minder-folden/home/~2020.3.16..15.26.38..05ed/sys/vane/clay
[...]

PR #2595, opened 2020-03-23
clay: add option to skip queue

philipcmonk:
Makes it so that |cancel %force skips the next thing in the queue if
you're not in the middle of something.  If you are in the middle of
something, it skips the thing you're in the middle of (just like naked
|cancel).

This should resolve issues where |cancel doesn't drain the queue.


vvisigoth:
nice!
—
~poldec-tonteg
http://urbit.org

philipcmonk:
Confirmed this worked for ~minder in #2586

PR #2601, opened 2020-03-24
publish: delete notes

matildepark:
- Adds the ability to delete notes for note authors, within the note, beside the 'edit' button.
- Adds a slight tweak to line up mobile navigation between the notebook view and the all notebooks view — it was affected by inner component padding, so it's pulled out of that padding.

(Also rearranged `componentWillMount` in note.js since it was in a weird position in the code.)


vvisigoth:
wow
—
~poldec-tonteg
http://urbit.org

PR #2605, opened 2020-03-25
server: add basic cache-control for faster load speeds

loganallenc:
We weren't doing basic server-side caching of our assets, which was significantly slowing down our pageload speed. Added a simple policy where assets we frequently change have a 1 day timeout and assets that don't frequently change have a 1 week timeout.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

### DIFF
--- a/pkg/arvo/lib/server.hoon
+++ b/pkg/arvo/lib/server.hoon
@@ -57,15 +57,18 @@
 ++  gen
   |%
   ::
+  ++  max-1-da  ['cache-control' 'max-age=86400']
+  ++  max-1-wk  ['cache-control' 'max-age=604800']
+  ::
   ++  html-response
     |=  =octs
     ^-  simple-payload:http
-    [[200 ['content-type' 'text/html']~] `octs]
+    [[200 [['content-type' 'text/html'] max-1-wk ~]] `octs]
   ::
   ++  js-response
     |=  =octs
     ^-  simple-payload:http
-    [[200 ['content-type' 'text/javascript']~] `octs]
+    [[200 [['content-type' 'text/javascript'] max-1-da ~]] `octs]
   ::
   ++  json-response
     |=  =octs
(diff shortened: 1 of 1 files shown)


loganallenc:
Have tested it, and it does in fact improve perf by a few seconds for initial pageload times on mobile internet.

ohAitch:
Stupid question, what happens when you want to change them?

Do we already implement If-Modified-Since?

loganallenc:
We don’t implement if-modified-since. I’m interested in putting significantly longer max-age times and making all built files other than index.html named purely by hash. Another way to ensure we’re fetching files only if it has changed.

ohAitch:
Seems legit, and I think already to an extent something browsers do to an
extent; as long as this change doesn't land before then!

(I suppose if it does I'm not exactly developing on urbit but, this seems
likely to break any workflows that involve |autocommit'ing in pngs or w/e)

loganallenc:
For the record we hardly ever change the PNGs or fonts, hence my lack of concern for putting a long expiration date on them.

belisarius222:
Isn't there some caching trick where you can add a query string that's the hash of the file contents, so you can still have real names for the files but also get content-addressing to make caching easier?

—
~rovnys-ricfer
https://urbit.org

ohAitch:
"Cache busting" is the general search term you want here

belisarius222:
Ah yes, that's the term.  Good ol' referential transparency...

—
~rovnys-ricfer
https://urbit.org

ixv:
How does this behave with publish's state injection stuff?

loganallenc:
I'm not doing any caching on manx-responses and json-responses, which are what publish uses.

matildepark:
Heads up for all other Landscape developers that you have to force refresh after every commit now, lol

loganallenc:
You can disable caching by clicking inspect element > networking > disable caching

ohAitch:
I'm definitely more concerned about "HTML OTA updates take a week to propagate" than local development, yeah

matildepark:
Can’t think of a single time we changed the HTML files let alone OTA...?

philipcmonk:
As feedback, I just had to force refresh to get the new sidebar groups changes.

philipcmonk:
Oh, that's the 1 day timeout.  That's probably fine I guess?  As long as we don't ship backend changes that break the previous backend.

PR #2606, opened 2020-03-25
Scrollback fixed and tested in Brave, Safari, and Chrome

loganallenc:
Initial mailbox length changed to 50 because if there are no scrollbars, it's impossible to "scroll back" to request more, and on a macbook pro on regular screen zoom, 25 short messages don't trigger a scrollbar.

Scrollback works and has been tested.


loganallenc:
I suggest reviewing commit-by-commit, otherwise my indentation fixes make it hard to figure out what changed.

vvisigoth:
can we OTA this EOD @ixv ?
—
~poldec-tonteg
http://urbit.org

ixv:
yes

pkg/arvo/app/chat-view.hoon
```diff
@@ -109,7 +109,7 @@
       [[%give %fact ~ %json !>(*json)]~ this]
     (on-watch:def path)
     ::
-    ++  message-limit  25
+    ++  message-limit  50
```
Fang-: Is this because, on taller screens, this wouldn't provide enough messages to enable the scrollbar, and that somehow breaking the logic below? Seems like maybe we should be fixing that case instead of just throwing more messages at the problem, since this may not be enough for people with small font sizes, or vertical monitor setups or w/e.

loganallenc: Not just on taller screens, on a regular macbook pro, if the messages are all sent by a single author and are all only one line, it doesn't provide enough messages. This is a reasonable fix for now, as it's an edge case.

Fang-: > This is a reasonable fix for now, as it's an edge case.

So we're negatively impacting performance across the board, to salvage the edge case. I see.

(I mean it's not gonna be as bad as the original `100` message-limit, but still...)

matildepark:
Seems like we could auto-fire the “get backlog" event if the screen is so tall we can’t scroll? 

It’s true that we don’t know the source of the performance impact rn but if it’s the ship, and each additional chat we’re subscribed to scales performance based on the size of the requested initial page (50 vs 25), then this has a potential performance impact around (number of chats you are subscribed to * 2)?

(Idk, I know *about* Big O but not enough to phrase it like that.)

— 
~haddef-sigwen
https://urbit.org

loganallenc:
Okay expect blazing fast speeds now

loganallenc:
Good suggestion @matildepark

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -14,288 +14,299 @@ import { deSig } from '/lib/util';
...
+ 
+    if (
+      prevProps.match.params.station !== props.match.params.station ||
+      prevProps.match.params.ship !== props.match.params.ship
+    ) {
+      this.hasAskedForMessages = false;
+      
+      if (props.envelopes.length < 100) {
+        this.askForMessages();
+      }
```
Fang-: idk if I'm reading this correctly, please confirm: for the "newly created chat case" is this just gonna be trying to load in scrollback on every update until we hit 100 messages?

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -84,26 +89,32 @@ export class ChatScreen extends Component {
   askForMessages() {
     const { props, state } = this;
 
+    if (props.envelopes.length === 0) {
+      setTimeout(() => {
+        this.askForMessages();
+      }, 500);
```
ixv: please explain

loganallenc: In the case that you have refreshed the browser or directed the browser directly to a specific chatroom, you haven’t set up your subscriptions / received envelopes yet. Thus, props.envelopes.length === 0. Thus, to successfully askForMessages, we set a timeout to call askForMessages again once we’ve loaded the envelopes properly. The “chatroom has no messages ever and we call this method twice a second” edge case seems like a reasonable tradeoff.

pkg/interface/chat/src/js/components/chat.js
loganallenc: No, if you notice, this case is only being called upon switching chatrooms. So any time you switch chatrooms and the chatroom you’ve switched to has fewer than 100 messages, it will askForMessages()

loganallenc: No, if you notice, this case is only being called upon switching chatrooms. So any time you switch chatrooms and the chatroom you’ve switched to has fewer than 100 messages, it will askForMessages()

PR #2609, opened 2020-03-26
group-hook: fix resubscribe bug

### LINKED ISSUE #2578: one `planet` see different `links`/content than other `planet` within same `group`
**Describe the bug**
I've joined a `group` with my 2 `planets` and realized that the `first planet` see more `links` in a `links collection` than the `second planet`.

**Expected behavior**
Every planet in the `group` should see the same content (e.g. `links` in a `links collection`).

 - MacOS 10.15.4
 - Chrome
 - 0v11.n1t2j.o6f9m.jnh36.ldprs.ksorj.7gpj4.bo1qk.h5a2q.4il4o.nv70v (first planet)
 - 0v11.n1t2j.o6f9m.jnh36.ldprs.ksorj.7gpj4.bo1qk.h5a2q.4il4o.nv70v (second planet)

**Additional context**
The `second planet` was invited to the `group` a few days after the `first planet`.

Fang-:
This fixes a bug where group-hook wouldn't resubscribe correctly upon getting kicked. Now uses the correct group path for that resubscribe, instead of the whole wire.

Also adds on-load logic that checks to see if we have a subscription in place for every `synced` entry, and starts a new subscription if we don't.

Couldn't help myself and touched up the kick and watch-nack logic a bit to be more clear and correct. `=*` for pointing to specific parts of the wire that we care about, and the watch-nack case now makes sure it's actually relevant to the current configuration before we delete the entry. (The `=/  =ship` was sitting unused, pretty sure this is what it was intended for.)

I would test this on ~palfun but it's still spinning behn from today's OTA. Happy to wait that out and give it a shot before we merge, but logic seems fine from testing on a fake ship.

Fixes #2578's root cause.

### DIFF
--- a/pkg/arvo/app/contact-hook.hoon
+++ b/pkg/arvo/app/contact-hook.hoon
@@ -273,13 +273,38 @@
         %contacts
       =/  owner  (~(got by synced) path.fact)
       ?>  =(owner src.bol)
-      %+  weld
-      :~  (contact-poke [%delete path.fact])
-          (contact-poke [%create path.fact])
+      =/  have-contacts=(unit contacts)
+        (contacts-scry path.fact)
+      ?~  have-contacts
+        ::  if we don't have any contacts yet,
+        ::  create the entry, and %add every contact
+        ::
+        :-  (contact-poke [%create path.fact])
+        %+  turn  ~(tap by contacts.fact)
+        |=  [=ship =contact]
+        (contact-poke [%add path.fact ship contact])
+      ::  if we already have some, decide between %add, %remove and recreate
+      ::  on a per-contact basis
+      ::
+      %-  zing
+      %+  turn
+        %~  tap  in
+        %-  ~(uni in ~(key by contacts.fact))
+        ~(key by u.have-contacts)
+      |=  =ship
+      ^-  (list card)
+      =/  have=(unit contact)  (~(get by u.have-contacts) ship)
+      =/  want=(unit contact)  (~(get by contacts.fact) ship)
+      ?~  have
+        [(contact-poke %add path.fact ship (need want))]~
+      ?~  want
+        [(contact-poke %remove path.fact ship)]~
+      ?:  =(u.want u.have)  ~
+      ::TODO  probably want an %all edit-field that resolves to more granular
+      ::      updates within the contact-store?
+      :~  (contact-poke %remove path.fact ship)
+          (contact-poke %add path.fact ship u.want)
       ==
-      %+  turn  ~(tap by contacts.fact)
-      |=  [=ship =contact]
-      (contact-poke [%add path.fact ship contact])
     ::
         %add
       =/  owner  (~(got by synced) path.fact)
--- a/pkg/arvo/app/group-hook.hoon
+++ b/pkg/arvo/app/group-hook.hoon
@@ -74,20 +74,26 @@
       ?~  p.sign
         [~ this]
       %-  (slog u.p.sign)
-      ?>  ?=([@ @ *] wire)
-      =/  =ship   (slav %p i.wire)
-      =.  synced.state  (~(del by synced.state) t.t.wire)
+      ?>  ?=([@ %group ^] wire)
+      =/  =ship  (slav %p i.wire)
+      =*  group  t.t.wire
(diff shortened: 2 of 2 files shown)


Fang-:
~palfun just finally got out of OTA <<behn>> hell. I'll give this a test run there, just to confirm, then merge if successful.

Fang-:
This doesn't seem to create cards where it's needed. Debugging!

Fang-:
Wasn't using the `synced.old`, even though `synced` as pulled from the app core is just gonna be `~` in `+on-load`.

I'm observing something weird in ~palfun now though. Groups did get updated to contain newly-added members, and subscriptions are once again in place, but I'm missing some of them from the contacts UI. They still exist in group-store, but are missing from contact-store...

Hadn't looked at contacts UI post-ota until after I tried out this PR, so might not be related. Seems unlikely to be related, but I'll investigate further. Thoughts, @loganallenc?

loganallenc:
Weird. Contacts UI merges the contacts and groups sets to show its listings. Contacts are not created in the store until their profile has been shared.

Fang-:
Well, I'm missing some groups (& profiles) that were present previously. Like, they seem to have just disappeared from contact-store... The groups that are still there (in contact-store) contain their old data it seems, but the others are 100% gone.

loganallenc:
Sounds like they got unbundled somehow

Fang-:
Okay, this seems definitely related, because now the only group I have left in contacts is the one for which it didn't need to re-open a subscription for.

Opening a subscription to an existing group shouldn't get us an `%unbundle` right? Even if it did, that should affect group-store too, not just contact-store.

Fang-:
I think [this](https://github.com/urbit/urbit/blob/0d3e409f7c14d0e66e02aa6bfe447373823242c6/pkg/arvo/app/group-hook.hoon#L172) is the culprit.

@loganallenc why are we issuing `%unbundle` and `%bundle` here? Don't we just want to calculate the diff between `%path`'s "this is current group" and the group we have locally? This wouldn't be too bad if `%unbundle` was _just_ a group-store state change, but it notifies other apps (ie contacts), and they straight-up just dump group-related state when that happens. A `%bundle` is then no longer enough to get them back into action, because they probably aren't looking for updates to the deleted thing anymore.

loganallenc:
That unbundle is absolutely necessary for these subscriptions to work properly. If we delete a group, we should delete the contacts, and vice versa.

Fang-:
But a `%path` group-update doesn't signify deletion, just an authoritative initial response of "this is the current group definition"?

Sent you a meeting invite, call might be easier to discuss this.

loganallenc:
Ah, I see. That definitely looks like a mistake.

Fang-:
Phew! I have a fix for this. Want to give it the proper real-world test. Gonna try restoring ~palfun's contacts to its original glory, kill some subs to get back to a close-to-original state, then apply and see what happens.

(Wish I could mark this as a draft-status PR now, that merge button is way too green. For ops lurkers, just to be clear: DO NOT MERGE, this is actually kinda harmful to push out as-is.)

Fang-:
@loganallenc just noticed this while trying to debug ~palfun back into a sane state: maybe a similar mistake was made [in contact-hook](https://github.com/urbit/urbit/blob/0d3e409f7c14d0e66e02aa6bfe447373823242c6/pkg/arvo/app/contact-hook.hoon#L273)? Though that's not gonna impact much beyond contacts itself.

loganallenc:
We need to get rid of that one at the same time. Thanks for finding these. I put them in so we’d never end up in a state where we had more contacts / group members than we were supposed to from previous changes made while we were disconnected but didn’t think about the networking logic I had put in that would delete reactively.

Fang-:
With 4551c74, the group-hook changes proved themselves to be safe on live ~palfun.

Note that in 058dcdf we still `%remove`-then-`%add` individual contacts. I didn't want to do the broader-scale thing here, but we should probably add an `%all` edit or something similar at some point.

ixv:
I see this has an approval, but I also see a whole lot a comments after the approval. Can this go in? Please advise.

loganallenc:
Yes, please merge.

PR #2616, opened 2020-03-28
chat-js: improve subscription speed by doing two rounds of subscriptions and no longer subscribing to groups

loganallenc:
Groups and permissions have a 1:1 mapping so there's no need to even subscribe to groups.

Perf improvement of about 1-2 seconds on initial pageload!


matildepark:
T-two seconds is a lot!

pkg/interface/chat/src/js/components/lib/chat-input.js
```diff
@@ -505,29 +476,4 @@ export class ChatInput extends Component {
...
-        if (writePermission.who.has(window.ship)) {
-          return this.writeAccessRender();
-        } else {
-          return this.readOnlyRender();
-        }
-      }
-    } else {
-      return this.writeAccessRender();
-    }
-  }
```
matildepark: I guess we don’t have read only rooms anymore? Thinking. Thinking. Wait, what if you ban someone? What do they see?

pkg/interface/chat/src/js/components/lib/invite-search.js
```diff
@@ -49,15 +49,17 @@ export class InviteSearch extends Component {
...
+
+    Object.keys(this.props.permissions).map(group => {
+      if (this.props.permissions[group].who.size > 0) {
+        let groupEntries = this.props.permissions[group].who.values();
         for (let member of groupEntries) {
           peerSet.add(member);
         }
       }
+
       if (this.props.contacts[group]) {
```
matildepark: Wary of letting this component become idiosyncratic. Any way to pass the permissions as a “group” prop?

loganallenc: Can do that. Will update the PR tomorrow to just pass the set in as “group”

pkg/interface/chat/src/js/components/lib/chat-input.js
loganallenc: We currently show a “you are disconnected from this chatroom” box in that case

matildepark: Ok then whatev.

matildepark:
I did the thing where I approved too early. lol. You know — awaiting the one change above, this can merge ... will poke when we're ready ...

matildepark:
LGTM, @ixv ready to go

loganallenc:
ready to go

PR #2617, opened 2020-03-28
apps: added a duplicate channel.js for caching, updated script tags to point to it

loganallenc:
Transition to serving Channel.js from an app rather than from Eyre directly. Given that we've already merged cache-control into index.html, we can't remove the channel-js from Eyre yet, so this commit serves a duplicate channel.js from an app with max-1-day caching in the meantime.

I think this is the right decision, as caching this file removes 2s from time of pageload to receiving data.

In a week or two, we could remove channel-js from Eyre for good.

What say you?

### DIFF
--- a/pkg/arvo/app/chat/index.html
+++ b/pkg/arvo/app/chat/index.html
@@ -23,7 +23,7 @@
   </head>
   <body>
     <div id="root"/>
-    <script src="/~/channel/channel.js"></script>
+    <script src="/~channel/channel.js"></script>
     <script src="/~modulo/session.js"></script>
     <script src="/~chat/js/index.js"></script>
   </body>
--- a/pkg/arvo/app/contacts/index.html
+++ b/pkg/arvo/app/contacts/index.html
@@ -10,7 +10,7 @@
   </head>
   <body>
     <div id="root" />
-    <script src="/~/channel/channel.js"></script>
+    <script src="/~channel/channel.js"></script>
     <script src="/~modulo/session.js"></script>
     <script src="/~groups/js/index.js"></script>
   </body>
--- a/pkg/arvo/app/launch.hoon
+++ b/pkg/arvo/app/launch.hoon
@@ -11,6 +11,13 @@
   /|  /js/
       /~  ~
   ==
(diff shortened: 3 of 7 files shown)


matildepark:
Two seconds is a lot.

philipcmonk:
I'm a little confused.  Are you moving it to an app so that it can be cached?  Why is app stuff cacheable but not eyre stuff?

loganallenc:
Yes, I’m moving it to an app so it can be cached. Eyre’s return static-file-on-duct doesn’t seem to take in HTTP headers as an argument, as far as I can tell. I also think it’s better to move the channel-js outside of Eyre anyhow.

ohAitch:
How does eyre take 2 seconds to serve 200 lines of javascript?

loganallenc:
You can test it yourself if you'd like. I'm in a cabin with satellite internet and am accessing my server through GCloud.

matildepark:
Interesting new variant on the seven proxies meme

—
~haddef-sigwen
https://urbit.org

philipcmonk:
Definitely a fan of doing performance work while on a satellite connection.

@eglaysher How hard would it be to allow caching for channel.js in eyre?

It feels a little strange to attach this to app/launch, but I have no strong objection to it.  If it's hard to change this is Eyre, I think this is fine.

eglaysher:
I don't think it would be that hard to add headers to eyre's static serving.

loganallenc:
I find it strange that this JS client is included in Eyre at all, as it is just one of many means by which one could communicate to Eyre’s API. Imo it should be moved to userspace regardless.

loganallenc:
Agreed that it likely shouldn't live in Launch longterm, but it seems good enough for now TM. If no objections from Elliot, can I get a merge today or tomorrow? We're seeking to do a feature freeze on Tuesday or Wednesday before all the demo stuff Friday.

philipcmonk:
Good point.  We do want to it to be an API usable in different environments and languages, so sometime we should clean it up for that purpose.

loganallenc:
@ixv can I get a merge?

PR #2618, opened 2020-03-28
chat: reversed order that messages are stored in for faster performance

loganallenc:
This PR reverses the order that messages are stored in within %chat-store. This provides a major improvement to the time complexity of joining chatrooms and should mildly reduce the amount of time spent to send messages as well.

This has been tested to make sure there are no regressions to scrollback performance and that the OTA of this works properly on existing ships.

Note: I removed the complicated manual backlog pagination going on in %chat-hook because ames should already be paginating this noun adequately, and if it's not, then that's an ames bug.


pkg/arvo/app/chat-view.hoon
```diff
@@ -122,10 +122,9 @@
     ++  truncate-envelopes
       |=  envelopes=(list envelope)
       ^-  (list envelope)
-      =/  length  (lent envelopes)
-      ?:  (lth length message-limit)
+      ?:  (lth (lent envelopes) message-limit)
         envelopes
-      (slag (sub length message-limit) envelopes)
+      (scag message-limit envelopes)
```
belisarius222: You shouldn't need the length check here.  +scag will no-op if the list length is less than the supplied index.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -402,41 +402,15 @@
   ?>  (~(has by synced) pas)
   ::  check if read is permitted
   ?>  (is-permitted src.bol pas)
+  =/  =mailbox  (need (chat-scry pas))
```
belisarius222: This is a nit, but since you don't use `.mailbox`, I'd just do this:
```
=/  envs  envelopes:(need (chat-scry pas))
```

pkg/arvo/app/chat-hook.hoon
```diff
@@ -402,41 +402,15 @@
...
+  =*  envs  envelopes.mailbox
   %-  zing
   :~  [%give %fact ~ %chat-update !>([%create pas])]~
       ?.  ?&(?=(^ backlog-start) (~(has by allow-history) pas))  ~
-      (paginate-messages pas (need (chat-scry pas)) u.backlog-start)
+      [%give %fact ~ %chat-update !>([%messages pas 0 (lent envs) envs])]~
```
belisarius222: It seems unnecessary to pass both `(lent envs)` and `envs`, and it costs a traversal to calculate the length.  I think we'd be better off not passing the length around.  I might be wrong, depending on how this fits into the system at large, and it's also not really relevant to the changes in this PR.

loganallenc: We use the same %messages data structure when passing on-scroll events, which do use both the start and end index.

loganallenc: We could add an action / update independently for %backlog, but that seems like a premature change. I'd like to see how the network likes this change first.

pkg/arvo/app/chat-view.hoon
loganallenc: Cute, will make da change

pkg/arvo/app/chat-hook.hoon
loganallenc: I like this! Thanks

joemfb:
The reversal is righteous, IMO.

As for the pagination, one reason to do so in the app and not (just) in %ames is that sending smaller messages lets your peer start reacting to those messages faster. IDK if that's why it was done this way here, but it's worth considering regardless.

pkg/arvo/app/chat-store.hoon
```diff
@@ -182,6 +188,7 @@
   :-  (send-diff path.act act)
   state(inbox (~(del by inbox) path.act))
 ::
+
```
Fang-:
```suggestion
```

pkg/arvo/app/chat-store.hoon
```diff
@@ -240,12 +242,12 @@
     letter(output (eval bol hoon))
   letter
 ::
-++  append-envelope
+++  prepend-envelope
```
Fang-: idk, it's still (semantically) appending to the message log, even if that (technically) means prepending to the list.

pkg/arvo/app/chat-store.hoon
```diff
@@ -206,15 +213,10 @@
   ?~  envelopes.act
     :_  state(inbox (~(put by inbox) path.act u.mailbox))
     %+  send-diff  path.act
-    :*  %messages
-        path.act
-        (sub length.config.u.mailbox (lent evaluated-envelopes))
-        length.config.u.mailbox
-        evaluated-envelopes
-    ==
+    [%messages path.act 0 (lent evaluated-envelopes) evaluated-envelopes]
```
Fang-: I think we still want to pass the old-style indexes? ie the ones that match message numbers?
Your range here is always going to be fixed to start at `0`, which doesn't really tell you anything about the messages.

Arguably we don't want to send them at all, because they're already in the message's numbers. I guess this is somewhat similar to the argument @belisarius222 makes in chat-hook.

(We don't have any ask-for-messages-by-index watch/scry paths, do we? If we did, we'd want those to use message numbers, not list indexes.)

pkg/arvo/app/chat-store.hoon
loganallenc: That's very confusing and I want to avoid weird naming like that.

loganallenc:
As discussed in OS1 meeting yesterday, we're moving to only sending the last 10,000 messages of the backlog for now. Will include that in this PR.

ixv:
Can you test that on urbit-help before you pick a number? I think urbit-help is something like 40,000 messages, so I don't know if quartering the load time will be enough.

loganallenc:
Sure.

ixv:
Can you remove the compiled js please?

loganallenc:
Still needs some more testing, but yes, when I'm done I'll remove it.

loganallenc:
@ixv Please review. This has been tested on ~tacryt-socryp extensively now. I ran it as it would run in an OTA, then went through all the JS on every chatroom to make sure scrollback worked in all conditions.

pkg/arvo/app/chat-store.hoon
Fang-: This still stands I think. With these changes, it looks like `+peek-x-envelopes` expects you to know about the reverse ordering in storage, which 1) leaks an implementation detail and 2) isn't practically usable as implemented right now, because to get at _specific_ messages, you need to know about the total that exists, and adjust the indexes you're asking about accordingly. (Though maybe the negative number addressing is still usable?)

If there ever was a time to document these interfaces, even just one-liners in app file headers, this is probably it.

loganallenc: It works. If we want to do refactors for developer ergonomics, let's do them separately.

ixv:
I booted 2 ships from v1.0.12, made a chat and sent a bunch of messages in it. Then I copied in these changes and a bunch of scrollback wen't missing.

loganallenc:
Gotcha, so in your test you hadn't copied in the new JS, and then you tried it with the new JS and it worked as expected?

ixv:
Thats right.

loganallenc:
Great! morge plz

PR #2619, opened 2020-03-28
/who JSON endpoint

liam-fitzgerald:
The /who endpoint responds with the name of the ship in JSON. Necessary to validate the success of DNS binding in Bridge to enable one-click login.

- I added the %serve card to `launch`, unsure if this is the best place for it or not.

### DIFF
--- a/pkg/arvo/app/launch.hoon
+++ b/pkg/arvo/app/launch.hoon
@@ -57,10 +57,14 @@
 |_  bol=bowl:gall
 +*  this  .
     def   ~(. (default-agent this %|) bol)
+    launch-who  [%pass /who %arvo %e %serve [~ /who] q.byk.bol /gen/who/hoon ~]
 ++  on-init
   ^-  (quip card _this)
   :_  this(state *[%2 state-two])
-  [%pass / %arvo %e %connect [~ /] %launch]~
+  :~
+    launch-who
+    [%pass / %arvo %e %connect [~ /] %launch]
+  ==
 ::
 ++  on-save  !>(state)
 ::
@@ -68,6 +72,7 @@
   |=  old=vase
   ^-  (quip card _this)
   =/  old-state  !<(versioned-state old)
+  =|  cards=(list card)
   |-
   ?-    -.old-state
       %0
@@ -80,9 +85,9 @@
           (~(del by path-to-tile.old-state) /primary)
           first-time.old-state
       ==
-    $(old-state [%2 new-state])
+    $(old-state [%2 new-state], cards [launch-who cards])
   ::
-      %2  [~ this(state old-state)]
+      %2  [(flop cards) this(state old-state)]
   ==
 ::
 ++  on-poke
--- a/pkg/arvo/app/launch.hoon
+++ b/pkg/arvo/app/launch.hoon
@@ -61,9 +61,8 @@
 ++  on-init
   ^-  (quip card _this)
   :_  this(state *[%2 state-two])
-  :~
-    launch-who
-    [%pass / %arvo %e %connect [~ /] %launch]
+  :~  launch-who
+      [%pass / %arvo %e %connect [~ /] %launch]
   ==
 ::
 ++  on-save  !>(state)
@@ -76,14 +75,15 @@
   |-
   ?-    -.old-state
       %0
-    $(old-state [%1 tiles.old-state data.old-state path-to-tile.old-state %.n])
+    $(old-state [%1 tiles data path-to-tile %.n]:old-state)
   ::
       %1
     =/  new-state=state-two
-      :*  (~(del in tiles.old-state) [%contact-view /primary])
-          (~(del by data.old-state) %contact-view)
-          (~(del by path-to-tile.old-state) /primary)
-          first-time.old-state
+      =>  [old-state .]
+      :*  (~(del in tiles) [%contact-view /primary])
+          (~(del by data) %contact-view)
+          (~(del by path-to-tile) /primary)
+          first-time
       ==
     $(old-state [%2 new-state], cards [launch-who cards])
   ::
--- a/pkg/arvo/app/launch.hoon
+++ b/pkg/arvo/app/launch.hoon
@@ -48,6 +48,9 @@
   ==
 ::
 +$  card  card:agent:gall
+++  launch-who
+  |=  =desk
+  [%pass /who %arvo %e %serve [~ /who] desk /gen/who/hoon ~]
 --
 ::
 =|  [%2 state-two]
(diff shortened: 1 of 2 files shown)


liam-fitzgerald:
cc: @Fang-

pkg/arvo/gen/who.hoon
```diff
@@ -0,0 +1,28 @@
...
+::
+|=  [[now=@da eny=@ bek=beak] ~ ~]
+|=  [authorized=? =request:http]
+^-  simple-payload:http
+=/  url=(unit pork:eyre)
+  (rush url.request apat:de-purl:html)
+::
+::  url doesn't match expected binding from :launch
+::
+?.  ?=(^ url)
```
Fang-: It's customary to just use `?^` or `?~` for these kinds of typechecks. Same for the check on `t.q.u.url` below.

```suggestion
?~  url
```

pkg/arvo/gen/who.hoon
```diff
@@ -0,0 +1,28 @@
...
+::
+?.  ?=(~ t.q.u.url)
+  [[%404 ~] ~]
+=/  response=json
+  (frond:enjs:format %who (ship:enjs:format p.bek))
+:-
+  :-  %200
+  :~
+    ['content-type' 'application/json']
+    ['access-control-allow-origin' 'https://bridge.urbit.org']
```
Fang-: Here and in launch's `+on-init`:

```suggestion
  :~  ['content-type' 'application/json']
      ['access-control-allow-origin' 'https://bridge.urbit.org']
```

Opening this to just Bridge is fine for the time being. I need to write that CORS handling proposal...

pkg/arvo/app/launch.hoon
```diff
@@ -40,36 +41,55 @@
...
     %=  this
-      state  [%1 tiles.old-state data.old-state path-to-tile.old-state %.n]
+      state  [%2 tiles.old-state data.old-state path-to-tile.old-state %.n %.y]
     ==
   ::
-      %1  this(state old-state)
+      %1
+    :-  ~[launch-who]
+    %=  this
+      state  [%2 tiles.old-state data.old-state path-to-tile.old-state first-time.old-state %.y]
```
Fang-: Since the only things used here are constants and things from `old-state`, you could

```suggestion
      state  [%2 tiles data path-to-tile first-time %.y]:old-state
```

Also applies to the `%0` case I guess.

pkg/arvo/app/launch.hoon
```diff
@@ -40,36 +41,55 @@
...
 ::
-=|  state-one
+=|  state-two
 =*  state  -
 %-  agent:dbug
 ^-  agent:gall
 |_  bol=bowl:gall
 +*  this  .
     def   ~(. (default-agent this %|) bol)
+    launch-who  [%pass /who %arvo %e %serve [~ /who] q.byk.bol /gen/who/hoon ~]
```
Fang-: Feels a bit weird to have this as an alias. I know the extra arm would make the agent core fail to nest, and launch doesn't use an extra core to work around that yet... Probably fine?

pkg/arvo/gen/who.hoon
```diff
@@ -0,0 +1,28 @@
...
+?.  ?=(^ url)
+  ~|  [%invalid-url url.request]  !!
+?.  ?=([%who *] q.u.url)
+  ~|  [%unknown-url url.request]  !!
+::
+::
+?.  ?=(~ t.q.u.url)
+  [[%404 ~] ~]
+=/  response=json
+  (frond:enjs:format %who (ship:enjs:format p.bek))
```
Fang-: The minimalist in me wants to just send `(scow %p p.bek)` over the wire as-is instead of a neat json bundle, but I guess this is slightly more palatable to clients?

pkg/arvo/app/launch.hoon
```diff
@@ -40,36 +41,55 @@
       path-to-tile=(map path @tas)
       first-time=?
   ==
++$  state-two
+  $:  %2
+      tiles=(set tile:launch)
+      data=tile-data:launch
+      path-to-tile=(map path @tas)
+      first-time=?
+      setup-who=_%.y
```
Fang-: Doesn't look like we need to keep this flag around? We don't use it anywhere, and idk what we'd use it for.

pkg/arvo/app/launch.hoon
liam-fitzgerald: could put in the type declaration core?

pkg/arvo/gen/who.hoon
liam-fitzgerald: Sending just the string makes any change to the payload a breaking change, so I'd rather not. But up to you

pkg/arvo/app/launch.hoon
Fang-: That seems slightly better, yeah.

Fang-: That seems slightly better, yeah.

Fang-:
Except for, uh, non-trivial conflicts with launch.hoon?

loganallenc:
I’d recommend rebasing on master.

pkg/arvo/app/launch.hoon
```diff
@@ -60,29 +63,33 @@
...
-          (~(del by path-to-tile.old-state) /primary)
-          first-time.old-state
+      =>  [old-state .]
+      :*  (~(del in tiles) [%contact-view /primary])
+          (~(del by data) %contact-view)
+          (~(del by path-to-tile) /primary)
+          first-time
       ==
-    $(old-state [%2 new-state])
+    $(old-state [%2 new-state], cards [(launch-who q.byk.bol) cards])
```
Fang-, line 90: launch version `%2` is already on the livenet I believe (see: [version](https://github.com/urbit/urbit/blob/93c881cec99d79b73c0f6b3ff43dd9766423604d/pkg/arvo/app/launch.hoon) from the release tagged commit), so this isn't gonna get run for people who are up to date already. You'll probably want to add a version `%3`.

pkg/arvo/app/launch.hoon
```diff
@@ -60,29 +63,33 @@
...
-    $(old-state [%1 tiles.old-state data.old-state path-to-tile.old-state %.n])
+    $(old-state [%1 tiles data path-to-tile %.n]:old-state)
   ::
       %1
     =/  new-state=state-two
-      :*  (~(del in tiles.old-state) [%contact-view /primary])
-          (~(del by data.old-state) %contact-view)
-          (~(del by path-to-tile.old-state) /primary)
-          first-time.old-state
+      =>  [old-state .]
```
Fang-, line 84: This is what `=,` does, effectively.

PR #2633, opened 2020-03-31
clay: don't underflow when patch produces empty file

### LINKED ISSUE #2602: Clay: issue with committing in modified JS files
Repro steps:

1. Boot up ~zod or ~bus.
2. Mount
3. Turn on |autocommit %home
4. Copy in changed chat JS files via gulp.

Screen Shot 2020-03-24 at 6 53 31 PM
Screen Shot 2020-03-24 at 6 53 55 PM
Screen Shot 2020-03-24 at 6 54 43 PM

philipcmonk:
Fixes #2602

I haven't been able to reproduce this bug, but I can see how it could
happen in theory.  This should resolve the issue.


loganallenc:
Thanks so much :)

PR #2636, opened 2020-04-01
dojo: add autocomplete for various dojo syntaxes

liam-fitzgerald:
Adds autocomplete for
- :app poke syntax
- :app|gen poke generator syntax
- +gen naked generator syntax
- =var variable syntax

Mines generators for their `+help` strings

Demo:
Screen Shot 2020-04-01 at 11 26 37 am
Screen Shot 2020-04-01 at 11 26 02 am
Screen Shot 2020-04-01 at 11 25 40 am
Screen Shot 2020-04-01 at 11 25 25 am

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -1226,7 +1226,7 @@
       ^-  (unit [term tank])
       ?.  =(variable (end 3 (met 3 variable) name))
         ~
-      `[name (xsell q.cage)]
+      `[name (sell q.cage)]
     ::
     ++  complete-gen-poke-to-app
       |=  [app=term gen=term]
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -1115,7 +1115,7 @@
     =/  fore-pos-diff  (sub fore-pos pos)
     =+  vex=((full parse-command-line:he-parser) [1 1] txt)
     ?.  ?=([* ~ [* @ %ex *] *] vex)
-      res
+      (he-tab-not-hoon pos :(weld buf (tufa buf.say) "\0a"))
     =/  typ  p:(slop q:he-hoon-head !>(..dawn))
     =/  tl  (tab-list-hoon:auto typ p.q.q.p.u.q.vex)
     =/  advance  (advance-hoon:auto typ p.q.q.p.u.q.vex)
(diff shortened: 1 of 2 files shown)


belisarius222:
It might not be too hard to add |generator syntax for :hood generators while you're at it.

liam-fitzgerald:
> It might not be too hard to add |generator syntax for :hood generators while you're at it.

Done. Now I now longer have to remember the difference between |reset and |reboot!
Screen Shot 2020-04-01 at 12 30 38 pm

pkg/arvo/app/dojo.hoon
```diff
@@ -1168,6 +1168,140 @@
...
+    ::
+    ++  complete-variable
+      |=  variable=term
+      %+  complete  variable
+      %+  murn  ~(tap by var)
+      |=  [name=term =cage]
+      ^-  (unit [term tank])
+      ?.  =(variable (end 3 (met 3 variable) name))
+        ~
+      `[name (xsell q.cage)]
```
philipcmonk, line 1229: Any particular reason you chose `xsell` over `sell`?

pkg/arvo/app/dojo.hoon
```diff
@@ -1168,6 +1168,140 @@
...
+      =/  pfix=path
+        /(scot %p our.hid)/[q.byk.hid]/(scot %da now.hid)/gen/[app]
+      ::
+      %^  tab-generators:auto  pfix  `app
+      %+  murn
+        ~(tap by dir:.^(arch %cy pfix))
+      |=  [=term ~]
+      ?.  =(gen (end 3 (met 3 gen) term))
+        ~
+      ?~  =<(fil .^(arch %cy (weld pfix ~[term %hoon])))
```
philipcmonk, line 1248: Why the inconsistency between this and the `dir` access above (where you used infix `:`)?

liam-fitzgerald: The parser didn't like `:.^(` inside tall form, but was fine inside wide form.

pkg/arvo/app/dojo.hoon
liam-fitzgerald: None in particular, I just skimmed the dojo source until I found what I thought was the prettyprinting arm

pkg/arvo/app/dojo.hoon
philipcmonk: Huh, okay.  It would probably work if you wrap the right side with `[]`, but that's not really prettier, so this is fine.

pkg/arvo/app/dojo.hoon
philipcmonk: Ah, `xsell` is what you get from prefixing `????` or `?????`.  It's probably fine, but it hasn't been used as much as our main prettyprinter `sell`, so it could crash or print stuff differently.

galenwp:
This looks very cool to me

PR #2637, opened 2020-04-01
os1: add dark mode style for app name in headers

matildepark:
Style tweaking in #2615 lacked `white-d` for text in the header, so text in the header shows as black in dark mode. This PR remedies that.


loganallenc:
oh good

PR #2640, opened 2020-04-01
chat: DMs repositioned below grouped chats

matildepark:
I'm in so many direct messages that my grouped chats are buried, regardless of filtering down to seeing just them. Realising the error of my ways, this PR repositions the DMs *below* our (potentially filtered down) set of grouped chats.

cc @urcades

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/group-item.js
+++ b/pkg/interface/chat/src/js/components/lib/group-item.js
@@ -17,7 +17,30 @@ export class GroupItem extends Component {
     let first = (props.index === 0) ? "pt1" : "pt4"
 
 
-    let channelItems = channels.map((each, i) => {
+    let channelItems = channels.sort((a, b) => {
+      if (props.index === "/~/") {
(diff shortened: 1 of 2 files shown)


urcades:
Spiritual Stamp!

matildepark:
Standby, I'd like to have DMs sort by latest and all others by alphabetical, so I'm gonna add that here...

Fang-:
Should other apps put unmanaged things last, too? (Not as much pressure there, since they likely don't have as many unmanaged resources, but consistency.)

matildepark:
I honestly don’t see it as a big deal; Chat is a special case in my head because of how it treats managed and unmanaged resources. The other modules seem more lassez-faire (and considering it tends to be personal content that goes unmanaged in those modules, might do better above shared content). @urcades ?

matildepark:
I just think of these modules are different spatial areas; to a user they don’t think of the underlying content as the same primitive as a chat, just metadata...idk

PR #2644, opened 2020-04-01
groups: show tile icon, not full color; alphabetically sort group channels

matildepark:
Upon @galenwp request. Border color is the channel color.

Pasted Graphic


loganallenc:
What's the fallback if we don't have the image?

matildepark:
How would we programmatically see if we have it? There's no fallback because it splices it directly into the `img src`.

loganallenc:
ah, just maintain an explicit object mapping name to image, and when we don't know anything about the app, should be fine to show a blank square with the colored border (better than a "missing image" placeholder)

matildepark:
The `app` variable defaults to Unknown, thus Unknown.png — maybe we just add an "Unknown" image?

matildepark:
See that.

galenwp:
I love this

matildepark:
This shouldn’t be merged yet. I forgot dark mode. Give me an hour, I’ll eat dinner...

matildepark:
OK, fixed. Added a tinted background too to reflect the colour choice — the border color by itself looked weird with the default black on dark mode.

galenwp:
Is this going in today?

ixv:
Yes

galenwp:
Whew

matildepark:
Ideally … the images are slightly canted left, @urcades is uploading new ones.

--
~haddef-sigwen
https://urbit.org

PR #2645, opened 2020-04-01
clock: ensure launch is listening correctly

### LINKED ISSUE #2585: Clock infrequently receives latlong
**Describe the bug**
Haven't seen this filed yet, but it's been on my mind. The latlong is seemingly infrequently passed to the clock; if I submit my latlong to Weather, the data passed to Clock remains `null` in React Dev Tools, not the string corresponding to the latlong.

However, if you boot a fresh dev ship and set your weather, you'll find it works properly.

**To Reproduce**
1. If your ship is affected, set your latlong in weather.
2. Wait for the clock to change.

Fang-:
In the wild, ships that were live pre-OS1 still had launch subscriptions
open to the clock on the `/tile` path, instead of the currently-used
`/clocktile` path. Additionally, launch state for the clock tile seemed
incomplete.

Here, we simply re-`%add` the clock to launch, during `+on-load`.

Note that launch [currently](https://github.com/urbit/urbit/blob/069ec72f8ad4c834ea5a04aa2749f3d787b87632/pkg/arvo/app/launch.hoon#L110) does not clean up old subscriptions on
path change. For the pre-OS1 case, the old path is no longer in use,
rendering the subscription harmless. For cases where the correct
subscription was already in place, it'll print a %watch-wire-not-unique,
but doesn't do any harm besides that.

I thought I'd just go and fix that launch logic real quick, but it's a bit more involved. To look up path from tile name, you kind of want [this](https://github.com/urbit/urbit/blob/069ec72f8ad4c834ea5a04aa2749f3d787b87632/pkg/arvo/app/launch.hoon#L39) to be a map instead of a set. Could write state transition logic, or maybe just do a search through the set instead, but don't want to let myself go on this. Happy to make some change there if requested, though.

Tested as working for the problem case on ~palfun-foslup. Tested as working for the "already works" case on a fakeship.

Closes #2585.

### DIFF
--- a/pkg/arvo/app/clock.hoon
+++ b/pkg/arvo/app/clock.hoon
@@ -1,4 +1,4 @@
-/+  *server, default-agent, verb
+/+  *server, default-agent, verb, dbug
 /=  tile-js
   /^  octs
   /;  as-octs:mimes:html
@@ -17,6 +17,7 @@
 +$  state-zero  [%0 data=json]
 --
 %+  verb  |
+%-  agent:dbug
 =|  state-zero
 =*  state  -
 ^-  agent:gall
--- a/pkg/arvo/app/clock.hoon
+++ b/pkg/arvo/app/clock.hoon
@@ -35,13 +35,21 @@
   ==
 ::  bootstrapping to get %goad started OTA
 ::
-++  on-save   !>(%2)
+++  on-save   !>(%3)
 ++  on-load
   |=  old-state=vase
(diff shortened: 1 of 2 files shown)


matildepark:
Closes #2585 ?

Fang-:
Oops, yes, pretty sure it does. Editted.

Fang-:
I thought @ixv made these but apparently @loganallenc has more ownership. Review approved, I guess. (;

PR #2648, opened 2020-04-01
os1: reimplement spinner and behaviours as part of app layouts

matildepark:
Screen Shot 2020-03-31 at 7 32 40 PM

- Reimplements the spinner outside of the header bar as a contextual part of the layout. All components that make API calls now reflect their status between front-end and back-end with an `awaiting` property in their state. This is passed to the spinner component to track when it should appear, alongside custom classes and text.
- Fixes issue where Publish settings would flash blank between edits and blank the other inputs as they were being edited — as props change, we atomically update our local state (reflected in the inputs) instead of all at once.
- Adds pending comments to Links to match Publish.
- Migrates the "mark all as seen" button in Links to the settings page.
- Fixes an issue where changing a description for a channel in links would never actually work because its code was pre-metadata store refactor.
- Fixes an issue where the back arrows in web Dojo were purple when visited.
- In Groups, sigil color is now set `onBlur`, not automatically, to match all other color fields.


matildepark:
The failed PR check seems unrelated?

galenwp:
@loganallenc can we get a review on this and get it out?

matildepark:
I’m resolving and testing one particular conflict to ensure it’s resolved.


~haddef-sigwen
https://urbit.org

PR #2650, opened 2020-04-01
Uruk mandelbrot

eglaysher:
This is a port of the hoon mandelbrot generator to moon and to uruk. I've been running this locally with `moon exec --new example/mandelbrot.moon`, but as you said, this is slow.

This adds a few jets and hooks up every jet needed by the generator into the global table.


benjamin-tlon:
Nice! Looks great.

PR #2659, opened 2020-04-02
group filter: properly pass /contacts invites and enumerate them

matildepark:
- In some apps, invites were still being filtered to app specific invites. This passes the full object and routes them to sidebar and header-bar appropriately
- Header-bar was checking for `contacts` not `/contacts`, thus it never found anything.
- If it happened to even miraculously get a proper contacts invites object, it wouldn't work because `.length` on an object returns `undefined`. What we want is `Object.keys(invites).length` to get an array we can measure. This fixes that in all applications.


matildepark:
let me resolve these conflicts now...

PR #2665, opened 2020-04-03
launch: handle change in url parsing

ixv:
Fixes the issue with the homepage breaking. I think this was cause by this PR: https://github.com/urbit/urbit/pull/2656

### DIFF
--- a/pkg/arvo/app/launch.hoon
+++ b/pkg/arvo/app/launch.hoon
@@ -151,7 +151,7 @@
     ?+  site.request-line
       not-found:gen
     ::
-        ~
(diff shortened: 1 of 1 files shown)


ixv:
Yes, I'm holding off announcing this because I want to make sure nothing else went wrong first.

Fang-:
Oops! Sorry about this. I tested some app frontends with that change, but actually missed the homepage.

Wondering if this "completely empty path" case should be getting caught and just produce the empty list again. Seems a bit weird, semantically, for it to be `['' ~]`.

PR #2667, opened 2020-04-03
chat-js: transition to new chat upon chatSynced changing in join or new chat screen

loganallenc:
Small change, improves perceived performance of join and new chat screen moving to chatroom as soon as chatSynced (%chat-hook property) changes.


ixv:
please dont commit built js

loganallenc:
When did that convention begin? Can remove it.

loganallenc:
Removed the built js

PR #2686, opened 2020-04-07
gall: added hack to reduce processing time for chat messages by 30%

loganallenc:
After a discussion with Ted, I decided to try to remove Ford entirely from the hot-path of chat message sending / receiving to improve performance. Received about a 50% speed improvement for receiving a chat poke, and about a 15% improvement to receiving a chat subscription update from my tests.

Here for review!

Note: This is a hack meant to improve performance until the point at which Ford pinto is ready.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -773,6 +773,60 @@
   ++  mo-apply
     |=  [agent=term =routes =deal]
     ^+  mo-core
+    ::  TODO: Remove this horrific hack when ford pinto comes!
+    =>  |%
+        +$  serial  @uvH
+        ::
+        +$  letter
+          $%  [%text text=cord]
+              [%url url=cord]
+              [%code expression=cord output=(list tank)]
+              [%me narrative=cord]
+          ==
+        ::
+        +$  envelope
+          $:  uid=serial
+              number=@
+              author=ship
+              when=time
+              =letter
+          ==
+        ::
+        +$  config
+          $:  length=@
+              read=@
+          ==
+        ::
+        +$  mailbox
+          $:  =config
+              envelopes=(list envelope)
+          ==
+        ::
+        +$  inbox  (map path mailbox)
+        ::
+        +$  chat-configs  (map path config)
+        ::
+        +$  chat-base
+          $%  [%create =path]
+              [%delete =path]
+              [%message =path =envelope]
+              [%read =path]
+          ==
+        ::
+        +$  chat-action
+          $%  ::  %messages: append a list of messages to mailbox
+              ::
+              [%messages =path envelopes=(list envelope)]
+              chat-base
+          ==
+        ::
+        +$  chat-update
+          $%  [%keys keys=(set path)]
+              [%config =path =config]
+              [%messages =path start=@ud end=@ud envelopes=(list envelope)]
+              chat-base
+          ==
+        --
     ::
     =/  =path
       =/  ship  (scot %p attributing.routes)
@@ -783,9 +837,19 @@
       [p q]:beak
     ::
     ?:  ?=(%raw-poke -.deal)
-      =/  =schematic:ford  [%vale ship-desk +.deal]
-      =/  =note-arvo  [%f %build live=%.n schematic]
-      (mo-pass path note-arvo)
+      ::  TODO: Remove this horrific hack when ford pinto comes!
+      ?+  mark.deal
+        =/  =schematic:ford  [%vale ship-desk +.deal]
+        =/  =note-arvo  [%f %build live=%.n schematic]
+        (mo-pass path note-arvo)
+      ::
+          %chat-action
+        =/  =cage  [%chat-action !>(;;(chat-action noun.deal))]
+        =/  new-deal=^deal  [%poke cage]
+        =/  app  (ap-abed:ap agent routes)
+        =.  app  (ap-apply:app new-deal)
+        ap-abet:app
+      ==
     ::
     ?:  ?=(%poke-as -.deal)
       =/  =schematic:ford  [%cast ship-desk mark.deal [%$ cage.deal]]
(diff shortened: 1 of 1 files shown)


loganallenc:
The structures in question reside in sur/chat-store.hoon, and looking at the history of the file, it hasn’t changed in 2 months. I’d be perfectly satisfied not to change or update for a long while, personally.

matildepark:
Might be worth flagging some comments in there to keep static until Pinto.

Fang-:
> I’d be perfectly satisfied not to change or update for a long while

We have general app api work coming up. New (or even just restructured) actions/updates aren't entirely out of the question.

Regardless we should at the very least test it, and maybe wrap in a `mule` & handle more gracefully as @philipcmonk recommends.

> Received about a 50% speed improvement for receiving a chat poke, and about a 15% improvement to receiving a chat subscription update from my tests.

What numbers do these percentages derive from? Is this a notable user-side improvement? This hack is sufficiently gnarly that I'd be hesitant to go ahead with it if the numbers aren't very big and/or can be gained elsewhere.

loganallenc:
I’ll email you the flame graph. And yes, notable user-side improvement.

ohAitch:
> looking at the history of the file, it hasn’t changed in 2 months. I’d be
perfectly satisfied not to change or update for a long while, personally.

To me this reads like "full-network continuity breach hereby planned in
approximately 2 months, maybe more like 8." Presumably a rolling one to not
break certain uh commitments, still effectively full-network if the vast
majority of people are, well, using chat.

loganallenc:
@ohAitch There’s no need to breach. Update process for chat types as follows:
1. Send out Ford Pinto update and remove this hack from Gall. No changes to chat types.
2. Wait a month for the update to propagate to the vast majority of network users.
3. Change desired chat types.

ohAitch:
Hopefully chat-action already contains a version number(or could orthogonally have a version number $% unioned in), but yeah fair

loganallenc:
@philipcmonk Would you be okay with us taking the approach of renaming the chat marks if we want to update them and me leaving a comment in sur/chat-store to not update the marks without renaming them?

If so, then I've tested this PR adequately. It's running on the live network and doing great.

If the mule thing is a blocker, I'm not sure how to do that but I can figure it out.

philipcmonk:
Committing to change the mark name if the format changes would help, yes.  I think if you put the `;;`s inside a +mule, then I'd be comfortable pushing this out.  I'm tempted to say we could cut the changes in `+mo-handle-ames-response` since they provide a 15% improvement and only on the client side, but I'll leave that up to you.

`+mule` is easy to use.  If you have a trap that produces an `A`, then it produces an `(each A tang)` where deterministic errors (like `!!` or `;;` failure, but not ^C or bail:meme) are caught and turn into the `[%| tang]` case.  Try these at the dojo:

```
;;(^ 5)
(mule |.(;;(^ 5)))
!:  (mule |.(;;(^ 5)))
(mule |.(;;(@ 5)))
```

The basic idea is that we really don't want gall to crash, especially not repeatedly.  Ie if you kill gall with a ^C that'll generally just kill the ames event and the other side will retry.  But if it's crashing deterministically, this can problematic, as in the case we ran into in July (?).

ohAitch:
Doesn't ++soft still exist, and do basically the above?

philipcmonk:
Yes and yes.  Feel free to use that.

loganallenc:
The latest version has been tested and works great.

It removes the %chat-update version of the hack since the 15% improvement wasn't substantial, and runs the %chat-action version of the hack virtualized using `+soft`.

I tried to upload new pills as well but my bandwidth tonight doesn't seem to be able to handle it.

loganallenc:
@philipcmonk Please take a look, and if all good, would love for this to go out in the OTA tomorrow evening.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -783,9 +837,24 @@
...
-      (mo-pass path note-arvo)
+      ::  TODO: Remove this horrific hack when ford pinto comes!
+      ?+  mark.deal
+        =/  =schematic:ford  [%vale ship-desk +.deal]
+        =/  =note-arvo  [%f %build live=%.n schematic]
+        (mo-pass path note-arvo)
+      ::
+          %chat-action
+        =/  chat-act=(unit chat-action)
+          ((soft chat-action) ;;(chat-action noun.deal))
```
philipcmonk, line 848:
```suggestion
          ((soft chat-action) noun.deal)
```

The call to `+soft` replaces the `;;`.  As written this would still crash if the noun didn't match the type.  Feel free to try examples in the dojo.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -845,18 +914,17 @@
...
     ^+  mo-core
-    ::
     ?-    -.ames-response
         ::  %d: diff; ask ford to validate .noun as .mark
         ::
         %d
       =/  =wire  /sys/rep
       ::  agents load their code from the %home desk, including marks
       ::
+      ::
```
philipcmonk, line 924:
```suggestion
```

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -783,9 +837,24 @@
...
+      ?+  mark.deal
+        =/  =schematic:ford  [%vale ship-desk +.deal]
+        =/  =note-arvo  [%f %build live=%.n schematic]
+        (mo-pass path note-arvo)
+      ::
+          %chat-action
+        =/  chat-act=(unit chat-action)
+          ((soft chat-action) ;;(chat-action noun.deal))
+        ?~  chat-act
+          ~&  %gall-raw-poke-chat-action-failed
```
philipcmonk, line 850: I would add `agent` and `attributing.routes` to this printf so we know which ship is sending to which app.

loganallenc:
@philipcmonk Made the requested changes! Thanks.

loganallenc:
@ixv This is good to go, please make pills and merge.

PR #2687, opened 2020-04-07
chat-js: resubscribe on error, not just on quit

loganallenc:


loganallenc:
@ixv can I get a merge?

PR #2693, opened 2020-04-07
various: alphabetise associations accesses the object

### LINKED ISSUE #2672: chat, publish, links, groups: sidebar not alphabetically sorting group titles
Possible bug in `alphabetiseAssociations` utility not accessing the title from the metadata object correctly and falling back to the path, because the sidebar is sorted alphabetically by `@p`. Alternatively an issue in how we handle the alphabetised associations object after the fact.

matildepark:
Previously we were checking for a title for the association by accessing
a non-existent property of the key we were using to iterate through
the object. What we want to do is access the iterated object to find
that title, and so this commit does that.

Because of this issue, group titles weren't being used to sort groups in all sidebars except Groups (which has a bespoke sort) — it was silently defaulting to the paths, sorting by `@p`.

Fixes #2672.

### DIFF
--- a/pkg/interface/chat/src/js/lib/util.js
+++ b/pkg/interface/chat/src/js/lib/util.js
@@ -138,14 +138,14 @@ export function alphabetiseAssociations(associations) {
   Object.keys(associations).sort((a, b) => {
     let aName = a.substr(1);
     let bName = b.substr(1);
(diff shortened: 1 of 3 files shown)


matildepark:
ping @loganallenc

PR #2694, opened 2020-04-08
invite-search: allow keyboard nav + selection

liam-fitzgerald:
Allows the user to navigate potential invitees with the arrow keys and
with shift+tab and tab. Enter adds the currently selected invitee. 

Looks like unminified tile js was committed for publish, which would explain the massive diff there. Should I not be minifying? 

cc: @matildepark


matildepark:
Apparently my email didn't send. Basically, yeah, remove the compiled JS and rebase.

PR #2695, opened 2020-04-08
publish: sanitise markdown snippets in preview

liam-fitzgerald:
Runs the snippet through ReactMarkdown instead of rendering it as text.
We restrict the allowed nodes in the render to pure text, so the snippet is not
overly visually heavy. At the moment, any links will be rendered as pure text instead of excluding them. Demo:
Screen Shot 2020-04-08 at 6 08 36 pm
Screen Shot 2020-04-08 at 6 08 48 pm

cc: @matildepark


matildepark:
Same as the last, can you rebase without the compiled JS?

Issue #2190, opened 2020-01-24
king: ames doesn't work after computer sleeps


philipcmonk:
@benjamin-tlon 

(From a binary a couple weeks ago, disregard if fixed). When my Mac falls asleep for long enough (overnight), we seem to stop sending packets until I restart the pier.  The timers look fine, so it says "not responding" for everyone I was trying to talk to (eg my sponsor tree).

PR #2698, opened 2020-04-08
Make king handle socket exceptions in ames

eglaysher:
This is an attempt at fixing #2190 . Since the socket can change between threads, stuff it in a TVar.

I'm not 100% sure that I've thought of everything here, so I'm asking for your advice. Do you think this approach works? Is there a better way to set up the types?


eglaysher:
(Fixed linked bug.)

PR #2700, opened 2020-04-08
invite-search: handle suggestions for short searches

### LINKED ISSUE #2635: In Links, invite search for groups only works if you prefix with the ship name, if the group name is 2 characters or less
**Describe the bug**
In Links, invite search for groups only works if you prefix with the ship name

**To Reproduce**
1. Make a group.
2. Go to new page in link collection, and try to make a link collection referencing it.

**Expected behavior**
I expected to type in the group name and see it in the list.

liam-fitzgerald:
Triggers invitee suggestions after one character of input. To
compensate for the potential explosion of suggestions, we
tighten our group searching criteria and truncate the
ship suggestions on short searches. Also addresses a bug where names of
groups were not being downcased before search.

Fixes #2635

cc: @matildepark

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/invite-search.js
+++ b/pkg/interface/chat/src/js/components/lib/invite-search.js
@@ -127,8 +127,8 @@ export class InviteSearch extends Component {
           isValid = false;
         }
 
-        if (shipMatches.length === 0 && isValid) {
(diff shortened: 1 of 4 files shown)


matildepark:
I just noticed this is still open. @liam-fitzgerald would you mind integrating a #2719 fix into this?

PR #2701, opened 2020-04-09
ford: print source line on syntax error

lukechampine:
Thanks to @belisarius222 for walking me through this. :)

Previous behavior:
```
> +hello 'world'
ford: %hood: syntax error at [12 13] in /~zod/home/0/gen/hello/hoon
ford: %core on /~zod/home/0/gen/hello/hoon failed:
~zod:dojo>
```

New behavior:
```
> +hello 'world'
ford: %hood: syntax error at [12 13] in /~zod/home/0/gen/hello/hoon
(crip (weld \ "hello, " (trip txt)))
------------^
ford: %core on /~zod/home/0/gen/hello/hoon failed:
~zod:dojo> 
```

The last line still reads "backwards" to me; I think it would make more sense at the top. But it wasn't obvious to me how to achieve that, since that line is produced in a separate build step.

### DIFF
--- a/pkg/arvo/sys/vane/ford.hoon
+++ b/pkg/arvo/sys/vane/ford.hoon
@@ -3165,10 +3165,12 @@
       ?~  q.parsed
         =/  =path  (rail-to-path source-rail)
         %-  return-error
-        :-  :-  %leaf
-            %+  weld  "ford: %hood: syntax error at "
-            "[{<p.p.parsed>} {<q.p.parsed>}] in {<path>}"
-        ~
+        =/  lyn  p.p.parsed
+        =/  col  q.p.parsed
+        :~  leaf+(runt [(dec col) '-'] "^")
+            leaf+(trip (snag (dec lyn) (to-wain:format q.q.as-cage)))
+            leaf+"ford: %hood: syntax error at [{<lyn>} {<col>}] in {<path>}"
+        ==
       ::
       (return-result %success %hood p.u.q.parsed)
     ::


philipcmonk:
Love this

lukechampine:
It would be great to have something like this for stack traces too, like `nest-fail`, but I'm not sure how that would work, since the elements of a stack trace can each span multiple lines. I suppose that even just printing the source of the innermost element would go a long way in aiding debugging, though. Something like:
```
> +hello 'world'
/~zod/home/~2020.4.9..20.26.27..0e29/sys/vane/ford:<[4.822 18].[4.822 59]>
/~zod/home/0/gen/hello/hoon::[9 1].[12 35]>
/~zod/home/0/gen/hello/hoon::[10 1].[12 35]>
/~zod/home/0/gen/hello/hoon::[11 1].[12 35]>
/~zod/home/0/gen/hello/hoon::[12 1].[12 35]>
/~zod/home/0/gen/hello/hoon::[12 7].[12 34]>
(weld 'hello, ' (trip txt)))
nest-fail
ford: %slim failed: 
ford: %ride failed to compute type:
ford: %plan failed: 
ford: %core on /~zod/home/0/gen/hello/hoon failed:
```

belisarius222:
With stack traces, it's not always obvious whether the part that needs to be fixed is at the beginning or end of its range of lines and columns.  So the naive thing to do would be to print the whole range of lines and columns in the stack trace.

That's probably too long, though, since sometimes it's hundreds of lines.  We could truncate anything more than, say, five lines at the top and five lines at the bottom, or just print the first and last lines as bookends.  We could also just try one of those, and see how it feels.

—
~rovnys-ricfer
https://urbit.org

ohAitch:
Separately stack traces should not be hundreds of lines; I remember at one
point I got code in that truncated any concentric "telescoping" section to
only the first outermost and last most specific entry, but that had its own
problems.

lukechampine:
>We could truncate anything more than, say, five lines at the top and five lines at the bottom, or just print the first and last lines as bookends.

I would imagine that in the vast majority of cases, this would provide sufficient context to identify the problem. If the innermost stack element spans many lines, it's probably just a long `:~` (or similar), so the omitted lines wouldn't add much information, since it's the list as a whole that's causing the error, not a specific element.

philipcmonk:
For runtime errors, the whole stack is quite useful.  Often times the error is in some generic little function and I really want to know where that function was called.

For compile-time type errors, the main time the whole stack is useful is when compiling wet gates, since I believe you can tell whether it's the definition or the call-site that's the problem.

I'm quite wary of removing stack trace lines in general.  @ohAitch's proposal of removing telescoping lines is probably good though, especially if replaced by `<X similar lines>`.

ohAitch:
Well, the previous implementation is available at
https://github.com/urbit/urbit/blob/c78ed1b/sys/hoon.hoon#L5298-L5317 for
the curious

PR #2703, opened 2020-04-09
chat-fe: copy patp to clipboard on author name click

liam-fitzgerald:
Copies patp to clipboard when user clicks on the nickname/patp of the author of a message.

cc: @matildepark

### DIFF
--- a/pkg/interface/chat/src/js/components/lib/message.js
+++ b/pkg/interface/chat/src/js/components/lib/message.js
@@ -2,7 +2,7 @@ import React, { Component } from 'react';
 import { Sigil } from '/components/lib/icons/sigil';
 import classnames from 'classnames';
 import { Route, Link } from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


pkg/interface/chat/src/js/components/lib/message.js
```diff
@@ -181,7 +181,15 @@ export class Message extends Component {
             <div className="hide-child" style={paddingTop}>
               <p className="v-mid f9 gray2 dib mr3 c-default">
                 <span
-                  className={contact.nickname ? null : "mono"}
+                  className={"pointer " + (contact.nickname ? null : "mono")}
+                  ref="author"
+                  onClick={() => {
+                    writeText(props.msg.author);
+                    this.refs.author.innerText = "Copied";
```
matildepark, line 188: This means that if the name is `mono`, the "Copied" will also be `mono`. We want the `Copied` to always be Inter, I think.

PR #2707, opened 2020-04-09
publish: eagerly store your own changes

ixv:
This PR changes publish so that when we make changes, whether its adding, editing, or deleting a post or comment, it will immediately store that upon receiving the poke, even if we are not the host of the notebook, and immediately send that update to the frontend.

If we aren't the host, and our poke to the host crashes, we handle that by either deleting the post/comment in the case of an add, or restoring the old one, in the case of an edit or delete. We also then send the opposite update to the frontend, eg: a delete if it was a failed add, etc.

The frontend reducers for adding notes and comments have been updated to be able to deduplicate updates, since in the normal case the frontend will now receive 2 updates, one pre- and one post-confirmation.

Additionally, this PR contains some cleanup to make it easier to write. One was the merging of books and subs, into a single field called books. The other was to eliminate the OS1 +onload stuff as its basically dead code now, I expect I'll get some pushback on that though.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -119,17 +119,52 @@
   ++  on-load
     |=  old=vase
     ^-  (quip card _this)
-    =/  old-state=versioned-state  !<(versioned-state old)
+    =/  old-state=(each versioned-state tang)
+      (mule |.(!<(versioned-state old)))
     =|  cards=(list card)
     |^
-    ?-  -.old-state
+    ?:  ?=(%| -.old-state)
+      =/  zero  !<(state-zero old)
+      =/  rav  [%next %t [%da now.bol] /app/publish/notebooks]
+      =/  tile-json
+        (frond:enjs:format %notifications (numb:enjs:format 0))
+      =/  init-cards=(list card)
+        :~  [%pass /read/paths %arvo %c %warp our.bol q.byk.bol `rav]
+            :*  %pass  /permissions  %agent  [our.bol %permission-store]  %watch
+                /updates
+            ==
+            (invite-poke:main [%create /publish])
+            :*  %pass  /invites  %agent  [our.bol %invite-store]  %watch
+                /invitatory/publish
+            ==
+            [%give %fact [/publishtile]~ %json !>(tile-json)]
+        ==
+      =+  ^-  [kick-cards=(list card) old-subs=(jug @tas @p)]  kick-subs
+      =/  inv-scry-pax
+        /(scot %p our.bol)/invite-store/(scot %da now.bol)/invitatory/publish/noun
+      =/  inv=(unit invitatory)  .^((unit invitatory) %gx inv-scry-pax)
+      =|  new-state=state-two
+      =?  tile-num.new-state  ?=(^ inv)
+        ~(wyt by u.inv)
+      %=  $
+          old-state  [%& %2 new-state]
+      ::
+          cards
+        ;:  weld
+          (kill-builds pubs.zero)
+          kick-cards
+          init-cards
+          (move-files old-subs)
+        ==
+      ==
+    ?-  -.p.old-state
         %1
       %=  $
-          -.old-state  %2
+          -.p.old-state  %2
       ::
           cards
         %-  zing
-        %+  turn  ~(tap by books.old-state)
+        %+  turn  ~(tap by books.p.old-state)
         |=  [name=@tas book=notebook-2]
         ^-  (list card)
         =/  group-host=(unit @p)
@@ -146,25 +181,25 @@
     ::
         %2
       %=  $
-          old-state
+          p.old-state
         =/  new-books=(map [@p @tas] notebook)
           %-  %~  uni  by
-            %-  ~(run by subs.old-state)
+            %-  ~(run by subs.p.old-state)
             |=  old-notebook=notebook-2
             ^-  notebook-3
             (convert-notebook-2-3 old-notebook)
           ^-  (map [@p @tas] notebook)
-          %-  ~(rep by books.old-state)
+          %-  ~(rep by books.p.old-state)
           |=  [[key=@tas val=notebook-2] out=(map [@p @tas] notebook)]
           ^-  (map [@p @tas] notebook)
           %+  ~(put by out)
             [our.bol key]
           (convert-notebook-2-3 val)
-        [%3 our-paths.old-state new-books tile-num.old-state [~ ~]]
+        [%3 our-paths.p.old-state new-books tile-num.p.old-state [~ ~]]
       ==
     ::
         %3
-      [cards this(state old-state)]
+      [cards this(state p.old-state)]
     ==
     ::
     ++  convert-comment-2-3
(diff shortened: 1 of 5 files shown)


pkg/arvo/app/publish.hoon
```diff
@@ -127,149 +141,62 @@
...
-        %noun
-      ?:  =(%print-state q.vas)
-        ~&  state
-        [~ this]
-      ?:  =(%print-bowl q.vas)
-        ~&  bol
-        [~ this]
-      [~ this]
+    ?+  mar
+      (on-poke:def mar vas)
```
loganallenc, line 199: Just a minor style thing, but any reason not to put this in-line? Seems pretty standard

pkg/arvo/app/publish.hoon
```diff
@@ -1005,14 +951,161 @@
...
+    =/  host=@p  (slav %p i.t.t.wir)
+    =/  book-name  i.t.t.t.wir
+    =/  note-name  i.t.t.t.t.wir
+    =/  book  (~(get by books) [host book-name])
+    ?~  book
+      [~ state]
+    =/  note  (~(get by notes.u.book) note-name)
+    ?~  note
+      [~ state]
+    =.  notes.limbo   (~(put by notes.limbo) [host book-name note-name] u.note)
```
loganallenc, line 971: How is the limbo state rendered? Is it cleared regularly? Is it retried?

pkg/interface/publish/src/js/reducers/primary.js
```diff
@@ -60,26 +60,29 @@ export class PrimaryReducer {
     let book   = Object.keys(json[host])[0];
     let noteId = json[host][book]["note-id"];
     if (state.notebooks[host] && state.notebooks[host][book]) {
-      if (state.notebooks[host][book]["notes-by-date"]) {
-        state.notebooks[host][book]["notes-by-date"].unshift(noteId);
-      } else {
-        state.notebooks[host][book]["notes-by-date"] = [noteId];
-      }
-
       if (state.notebooks[host][book].notes) {
```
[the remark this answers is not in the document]
ixv: Agree that this is an issue throughout the reducer, and all the publish reducers in general, but its a seperate piece of work from this PR.

pkg/arvo/app/publish.hoon
ixv: It's not rendered anywhere. In the case of event failures, edits and deletes remove the thing stored in limbo and put it in our regular state. Failed adds remove the thing from our regular state and put it in limbo. Currently there's no way non-involved way of getting it back out of limbo, but I figured it would be better to at least store it and give savvy users that option than just swallowing it.

pkg/arvo/app/publish.hoon
ixv: I have no opinion

pkg/interface/publish/src/js/reducers/primary.js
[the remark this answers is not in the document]
loganallenc: yup true

pkg/arvo/app/publish.hoon
loganallenc: Hmm. Okay. Could you add a TODO relating to cleaning them up somehow or presenting them in some "weird limbo pending" interface later?

pkg/arvo/app/publish.hoon
```diff
@@ -245,7 +238,12 @@
     |=  [wir=wire sin=sign:agent:gall]
     ^-  (quip card _this)
     ?-    -.sin
-        %poke-ack   (on-agent:def wir sin)
+        %poke-ack
+      ?~  p.sin
+        (on-agent:def wir sin)
```
Fang-, line 243: While iirc the default behavior for no-error `%poke-ack`s is "do nothing", you explicitly want "do nothing" here, not whatever the default behavior is.

```suggestion
      ?~  p.sin  [~ this]
```

pkg/arvo/app/publish.hoon
Fang-: > [Limbo state is] not rendered anywhere.

Huh. Wasn't that the whole point of this? If you're _just_ storing it locally, and not rendering it, that doesn't do anything to improve the UX.

pkg/arvo/app/publish.hoon
```diff
@@ -953,6 +951,151 @@
...
+  |=  wir=wire
+  ^-  (quip card _state)
+  ?+  wir
+    [~ state]
+  ::  new note failed, stash it in limbo
+  ::
+      [%forward %new-note @ @ @ ~]
+    =/  host=@p  (slav %p i.t.t.wir)
+    =/  book-name  i.t.t.t.wir
+    =/  note-name  i.t.t.t.t.wir
```
Fang-, line 964: You generally want `=*` for referring to specific path elements like this.

pkg/arvo/app/publish.hoon
```diff
@@ -962,41 +976,98 @@
     =/  pax=path  /app/publish/notebooks/[book.act]/publish-info
     :_  state
     [(write-file pax %publish-info !>(new-book)) cards]
+  ::  %new-note
   ::
       %new-note
```
Fang-, line 1124: Having a bit of trouble parsing b990140. `%new-note` and `%new-comment` flows don't seem to be touching `limbo` at all. Is this correct? If yes, then why were these sections changed? It's not immediately apparent to me.

(It's a bit late here, forgive me if this should be obvious. But also a couple more comments here wouldn't be amiss. These sections feel pretty long.)

pkg/arvo/app/publish.hoon
ixv: @loganallenc 
Maybe I should just add a rudimentary poke %noun and scry interface for managing this. Probably better than leaving it.

@Fang- 
We use the normal non-limbo structure for pending posts/comments, in order to reuse the read endpoint code, rather than have to rewrite all of that. Limbo is for storing data that allows us to recover from failures that happen in a pending state.

pkg/arvo/app/publish.hoon
```diff
@@ -51,14 +51,25 @@
 ::
 +$  state-two
   $:  our-paths=(list path)
-      books=(map @tas notebook)
-      subs=(map [@p @tas] notebook)
+      books=(map @tas notebook-2)
+      subs=(map [@p @tas] notebook-2)
```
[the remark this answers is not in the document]
ixv: I think it's not in scope for this PR, given that publish has always been this way, and this is the first time I'm hearing this concern.

pkg/arvo/app/publish.hoon
Fang-: Hmm, interesting. My naively thought up implementation here was to store pending posts/comments separately, as you are doing, but then include them in the already-existing scry/subscription endpoint results by simply welding them on top of the "confirmed" lists.

pkg/arvo/app/publish.hoon
```diff
@@ -51,14 +51,25 @@
...
   ==
 ::
++$  state-three
+  $:  our-paths=(list path)
+      books=(map [@p @tas] notebook)
+      tile-num=@ud
+      $=  limbo
+      $:  notes=(map [@p @tas @tas] note)
+          comments=(map [@p @tas @tas @da] comment)
+      ==
```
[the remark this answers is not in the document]
ixv: No strong reason, just wanted to perturb things as minimally as possible.

pkg/arvo/app/publish.hoon
ixv: That would be a more invasive change, requiring a bunch of library functions and reducer logic to change.

pkg/arvo/app/publish.hoon
ixv: Does my answer to your question above explain this sufficiently? I can go into more detail. I will add some comments too.

Fang-: Yeah I think that explains it. Comments would still be good. Thanks! (:

ixv:
Just pushed some commits to do the following
- restore the OS0 -> OS1 on-load logic
- added the ability to scry limbo state and flush it
- fixed issue with edits where if you make multiple edits in a row, you can get updated from the remote host with old edits
- some style changes and comments

@Fang- I didn't change to =* for path segments because I use =/ for that everywhere else in the file.

loganallenc:
I can't load the publish new page even after rebuilding from latest JS. I get an error.

ixv:
Yeah that's the mousetrap issue.

pkg/arvo/app/publish.hoon
```diff
@@ -1376,7 +1381,10 @@
     =/  pax=path  /app/publish/notebooks/[book.act]/publish-info
     :_  state
     [(write-file pax %publish-info !>(new-info)) cards]
-  ::  %edit-note
+  ::  %edit-note:
+  ::    If poke is from us, eagerly store new note in books, and place the old
+  ::    note in limbo. If poke is to us, save file, otherwise forward the poke.
+  ::    If forwarded poke fails, old note is restored from limbo.
```
Fang-, line 1430: This helps clarify what this is used for. Would've also liked to see comments about the logic flow break up the section of code that follows, but this gives good context at least.

Limbo-based revert behavior for these edits is very straightforward to understand. But then you look at the create cases:

> Failed adds remove the thing from our regular state and put it in limbo. Currently there's no way non-involved way of getting it back out of limbo, but I figured it would be better to at least store it and give savvy users that option than just swallowing it.

If that's the angle you want to take (prevent loss of changes/data), shouldn't edits, on-restore, be _swapped_ with the old version from limbo, instead of _replaced_?

Arguably, at that point, you get very close to wanting full-on revision control, which is clearly out of scope here...

ixv:
Exposing the limbo state is a design question I feel. We've discussed in the past having drafts, and something like that feels exactly like where these limbo posts actually belong. Thats a larger change and way out of scope for this though.

PR #2714, opened 2020-04-10
chat: profile overlay and DMs

liam-fitzgerald:
Adds a profile overlay to display information about the user on hover. Hides and shows profile overlay on sigil click. Shows the overlay for 2 seconds after click. If the user moves their mouse into the overlay then it will remain open as long as the mouse is inside the overlay. Adds DMs, which are just unmanaged chats whose last path component starts with dm-- followed by the patp of the other user. 

cc: @urcades @matildepark 
Demo: 
Screen Shot 2020-04-10 at 6 06 26 pm
Screen Shot 2020-04-10 at 7 44 33 pm


urcades:
@liam-fitzgerald this is looking gorgeous — a wonderful suggestion made by @matildepark was to provide an additional affordance for editing one's own identity using this interface:

sigil-hover

We could simply co-opt the button style that uses "Send message" but have it be Black/gray.

urcades:
For the DM-ing interface something I'm considering, but have yet to fully resolve, is that for wherever we end up placing the "DM" button (at the top row of Chat options, vs. within the chat list itself), rather than kicking users into a form-like "send message" screen, we instead lighten up the interaction and build a popover field that immediately focuses a user's keyboard into a ship search interface:

os2-dm

(Illustration is simply showing how this field would appear with a DM button located in the chat list — it would obviously always "drop" downward if located in the top menu ... this overall placement and its UX is up for consideration per design review with the team)

cc @matildepark

urcades:
A huge disclaimer for the above image — please do not focus on / ask about the surrounding changes in chat interface, including new buttons/menus/whatever — everything is still in flight and in the middle of changing.

matildepark:
Probably we put a + beside the group titles to make new chats for those groups specifically; and New Chat either makes a Group DM or a Group proper depending on one single toggle (with a chat path that programmatically matches the DM spec here if it’s with one other person, so the Send Message button always redirects to it from the New Screen).

urcades:
Another note after playing with this is that I think these would actually be best displayed as an onClick-triggered event — onHover feels slightly too finnicky, and I could imagine it being slightly annoying when hovering over the interface, possibly triggering many of these to appear.

liam-fitzgerald:
Definitely agree re: summoning logic. However, at the moment it dismisses a couple of seconds after click, unless you hover over it. I'm unsure if this is the best dismissal logic. What's your take?

urcades:
Hm, given the density of information on the card, that's altogether not terrible, but I think a "best" approach, in terms of limiting side-affect interactions that aren't prompted by user input, would be to render 1) clicking outside of the popover and 2) typing `Esc` as the only means one could use to dismiss this overlay. 

If the above happens to be difficult for any reason, extending the onHover display to maybe 5 seconds would be perfectly fine.

liam-fitzgerald:
Sounds good. With the DM popover, there's no implied group context, so where should the names + colors come from?

urcades:
A black background and white sigil feels appropriate here —
monochromatic is good.

—
~fabled-faster
https://urbit.org/

liam-fitzgerald:
Screen Shot 2020-04-11 at 8 06 05 pm

liam-fitzgerald:
I've made the container have `position: relative` (in order to show overlays) which appears to break scrollback. The most recent commit is a fix for this.
 cc: @loganallenc you're the scrollback guru  does this look okay?

loganallenc:
This PR now depends on (and has conflicts with) the reverse chat message PR due to the scrollback change. Hopefully the reverse message PR can go in Monday and we can just rebase.

matildepark:
The “new DM” vs “new Chat” distinction needs to be considered a bit. I do think it’s important to segregate the DM creation flow so that they have the same underlying path, so we can programmatically link to them anywhere.

Maybe we just, in new chat, if it’s a new invite only unmanaged chat with one other person, enforce the programmatic path underneath.

pkg/interface/chat/src/js/components/lib/ship-search.js
```diff
@@ -0,0 +1,339 @@
...
+  }
+
+  render() {
+    const { state, props } = this;
+
+    return (
+      <div
+        ref={this.popoverRef}
+        style={{ top: "150%", left: "-80px" }}
+        className="b--gray2 b--solid ba absolute bg-white bg-gray0-d shadow-5"
```
matildepark, line 284: Idk, do we use drop shadows? @urcades It seems like removing the shadow and letting the border speak would be enough.

image

pkg/interface/chat/src/js/components/lib/profile-overlay.js
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    this.props.onDismiss();
+  }
+
+  render() {
+    const { contact, ship, color, topSpace, bottomSpace } = this.props;
+
+    let top, bottom;
+    debugger;
```
matildepark, line 40: 🔫 delete this

pkg/interface/chat/src/js/components/lib/profile-overlay.js
```diff
@@ -0,0 +1,94 @@
...
+              to={`/~chat/new/dm/~${ship}`}
+              className="b--green0 b--green2-d b--solid bw1 green2 mt3 tc pa2 pointer db"
+            >
+              Send Message
+            </Link>
+          )}
+          {isOwn && (
+            <a
+              href="/~groups/me"
+              className="b--gray3 b--solid bw1 black white-d mt3 tc pa2 pointer db"
```
matildepark, line 85: Try `className="b--black ba black white-d mt3 tc pa2 pointer db"`

pkg/interface/chat/src/js/components/lib/profile-overlay.js
```diff
@@ -0,0 +1,94 @@
...
+            <Link
+              to={`/~chat/new/dm/~${ship}`}
+              className="b--green0 b--green2-d b--solid bw1 green2 mt3 tc pa2 pointer db"
+            >
+              Send Message
+            </Link>
+          )}
+          {isOwn && (
+            <a
+              href="/~groups/me"
```
matildepark, line 84: We don't want to go to our root identity. We want to go to our card in the group in Groups. 

We can see that path from `props.association["group-path"]` in chat.js, for example. If the associated group is `~zod/mygroup` then it's just `~groups/view/~zod/mygroup/zod` (if you haven't shared your card with the group yet, does `view` still render? I think it would.)

If it starts with `/~/` then yeah, go to `/me`.

loganallenc:
Will review tomorrow or Sunday.

pkg/interface/chat/src/js/components/lib/ship-search.js
```diff
@@ -0,0 +1,339 @@
...
+          suggestEmpty
+        />
+      </div>
+    );
+  }
+}
+
+// import { useEffect } from 'react';
+
+// // via https:// usehooks.com/useOnClickOutside/
```
loganallenc, line 319: We will *never* incorporate hooks into any of our apps, so please remove this commented out code, as it's not a viable route that we would implement.

pkg/interface/chat/src/js/components/new.js
```diff
@@ -124,6 +124,10 @@ export class NewScreen extends Component {
       }
     });
 
+    if(state.ships.length === 1 && state.security === 'village' && !state.createGroup) {
+      props.history.push(`/~chat/new/dm/${aud[0]}`);
+    }
+
     if (!isValid) {
```
loganallenc, line 131: While we're making changes in new.js, let's remove the "unmanaged invite-only chat" case.

liam-fitzgerald: Can you elaborate?

matildepark: Essentially we're simplifying over time: 
- Unmanaged chats will be synonymous with direct messages, which are invite-only and created through the DM flow
- All other chats people are in will be within a group

**Right now** since we don't have open groups you can just paste a shortcode into, people use "lobby chats" (public unmanaged chats). By removing the invite-only unmanaged chat flow from new.js, we start to simplify these distinctions until groups are updated and we can remove unmanaged chats from the new.js screen entirely.

matildepark: It would mean, in this case, removing the invite only toggle and assuming it's always a public chat when we make a new unmanaged chat from the "New Chat" screen (managed chats stay invite only of course).

liam-fitzgerald:
reviews addressed, merged master and fixed and tested on FF.

pkg/interface/chat/src/js/components/lib/profile-overlay.js
```diff
@@ -0,0 +1,97 @@
...
+        </div>
+        <div className="pv3 pl3 pr2">
+          {contact && contact.nickname && (
+            <div className="b white-d">{contact.nickname}</div>
+          )}
+          <div className="mono gray2">{cite(`~${ship}`)}</div>
+          {!isOwn && (
+            <Link
+              to={`/~chat/new/dm/~${ship}`}
+              className="b--green0 b--green2-d b--solid bw1 green2 mt3 tc pa2 pointer db"
```
matildepark, line 80: `bw1` -> `ba`

matildepark:
Weird behaviour when close to the bottom.

image

pkg/interface/chat/src/js/components/lib/profile-overlay.js
```diff
@@ -0,0 +1,97 @@
...
+            >
+              Send Message
+            </Link>
+          )}
+          {isOwn && (
+            <a
+              href={identityHref}
+              className="b--black ba black white-d mt3 tc pa2 pointer db"
```
matildepark, line 88: Needs `b--white-d`

pkg/interface/chat/src/js/components/lib/ship-search.js
urcades: Yeah, let's null this drop shadow for now — there's no spec for their use in the latest Indigo works, and I think if we assemble our interfaces with clarity in mind in a general sense, we might not need them for a while.

liam-fitzgerald:
> Weird behaviour when close to the bottom.
> 
> image

what browser is this on?

matildepark:
Can’t remember if it was Firefox or Brave.

— 
~haddef-sigwen
https://urbit.org

liam-fitzgerald:
behaviour close to the bottom should be fixed and all review comments addressed

PR #2717, opened 2020-04-10
chat: fix onscroll for safari

matildepark:
Chrome's `navigator.userAgent` includes both Safari and Chrome in it, but Safari's doesn't. So we change `&&` to `||`.

(Also adds a space between a multi-line class declaration so we can get overflow-x-scroll working again for the chat tab bar in specific cases.)


matildepark:
Oh, I see I was seeing the nesting in the wrong way, ha. OK, closing.

PR #2718, opened 2020-04-10
chat: add firefox-specific chat window

matildepark:
It turns out that Firefox can retain the same `flex-column-reverse` implementation as long as it's wrapped within a non-flex `<div>` with `height: 100%;`. 

In implementing the window that way, it regressed behaviour in all other browsers such that the window no longer automatically scrolled to the bottom on sending a message.

So, I put a shim where, for Firefox, it's nested within that `<div>` with those conditions; for everything else, it isn't. Gross, but this is such a longstanding issue.

I also set the check for "did we get the backlog" to be a tad more lenient — it seems to be waiting for a difference of 200 messages? Fresher chatrooms wouldn't have 200 messages, but would have more than 20. I also make it just push us to the bottom on that initial backlog load.

Since Chat is a major use case, please do test. It seems to work the same way as our pre-existing browsers since the implementation is largely unchanged; but I'm still worried about edge cases. Also tagging @Fang- as our Firefox advocate.


matildepark:
I'm getting some inconsistent "scrolling to bottom upon sending a message" in Firefox for chatrooms with multiple people and inconsistent asking for messages if someone else owns the room. That latter one I don't get — the room I own auto-loads the backlog properly.

loganallenc:
This will have conflicts with my “reverse chat message order PR”. Nothing major, but it touches some of the same lines. Will review anyway, but hopefully we can rebase it on top once my PR goes in.

Fang-:
This doesn't autoscroll to bottom at all for me. Haven't been able to make it do that under any conditions. Firefox 75.0b11.

(Also doesn't seem to load scrollback beyond the initial handful of messages, but can repro that in the pre-PR UI also, so probably not related. (This is weird, because it mostly works on my ship on mainnet.))

> r*base it on top 

Please always just merge the target branch into your own, to preserve "make compatible with conflicting changes" changes as part of history explicitly.

matildepark:
Closing in favour of future work in #2733.

PR #2720, opened 2020-04-10
channel.js: added way to handle event source errors more smoothly, and implemented in chat

loganallenc:
Figured out that there was an unhandled error condition in Channel.js. Handled it. Tested an implementation within chat where chat can reset itself and re-establish subscriptions in no-connectivity scenarios. It works.

### DIFF
--- a/pkg/arvo/app/launch/js/channel.js
+++ b/pkg/arvo/app/launch/js/channel.js
@@ -1,5 +1,17 @@
 class Channel {
   constructor() {
+    this.init();
+    this.deleteOnUnload();
+
+    //  a way to handle channel errors
+    //
+    //
+    this.onChannelError = (err) => {
+      console.error('event source error: ', err);
+    };
+  }
+
+  init() {
     //  unique identifier: current time and random number
     //
     this.uid =
@@ -40,8 +52,10 @@ class Channel {
     //    disconnect function may be called exactly once.
     //
     this.outstandingSubscriptions = new Map();
+  }
 
-    this.deleteOnUnload();
+  setOnChannelError(onError = (err) => {}) {
+    this.onChannelError = onError;
   }
 
   deleteOnUnload() {
@@ -196,8 +210,9 @@ class Channel {
     }
 
     this.eventSource.onerror = e => {
-      console.error("eventSource error:", e);
       this.delete();
+      this.init();
+      this.onChannelError(e);
     }
   }
 
--- a/pkg/interface/chat/src/js/store.js
+++ b/pkg/interface/chat/src/js/store.js
@@ -9,7 +9,20 @@ import { LocalReducer } from '/reducers/local.js';
 
 class Store {
   constructor() {
-    this.state = {
+    this.state = this.initialState();
+
+    this.initialReducer = new InitialReducer();
+    this.permissionUpdateReducer = new PermissionUpdateReducer();
+    this.contactUpdateReducer = new ContactUpdateReducer();
+    this.chatUpdateReducer = new ChatUpdateReducer();
+    this.inviteUpdateReducer = new InviteUpdateReducer();
+    this.metadataReducer = new MetadataReducer();
+    this.localReducer = new LocalReducer();
+    this.setState = () => {};
+  }
+
+  initialState() {
+    return {
       inbox: {},
       chatSynced: {},
       contacts: {},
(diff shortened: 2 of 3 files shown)


pkg/interface/chat/src/js/store.js
```diff
@@ -9,7 +9,20 @@ import { LocalReducer } from '/reducers/local.js';
...
+    this.contactUpdateReducer = new ContactUpdateReducer();
+    this.chatUpdateReducer = new ChatUpdateReducer();
+    this.inviteUpdateReducer = new InviteUpdateReducer();
+    this.metadataReducer = new MetadataReducer();
+    this.localReducer = new LocalReducer();
+    this.setState = () => {};
+  }
+
+  initialState() {
+    return {
```
matildepark, line 25: Any reason for programming it to ignore future “initial” objects instead of reducing them to diffs? Or even clobbering the props of the same type? We might drop updates this way, no?

loganallenc: I'm not sure what you mean. `Subscription` sends `Store` a clear event that resets `Store` entirely to its initial state. Then `Subscription` re-establishes all subscriptions. It's the equivalent of a page reload.

matildepark: Oh! Ok. I thought we were just dropping future initials based on the glance here. Ok cool

loganallenc:
This is good to go. A nice quality-of-life fix for the future is to make failed actions re-send automatically...

pkg/arvo/app/launch/js/channel.js
```diff
@@ -196,8 +210,9 @@ class Channel {
     }
 
     this.eventSource.onerror = e => {
-      console.error("eventSource error:", e);
       this.delete();
+      this.init();
+      this.onChannelError(e);
```
Fang-, line 215: Uh, I thought the channel design was made so that you can reestablish a connection with the _same_ channel after error, which avoids opening new subscriptions on the backend, and thus doesn't get you superfluous initial results. Doing a full delete and init here seems to run counter to that.

(And even if this is sane, you might want to call `onChannelError` before doing all of that, for the expected order of operations.)

I'd expect the best behavior here to be something along the lines of checking `eventSource.readyState`, and reestablishing the connection (maybe through `connectIfDisconnected`) if that's `EventSource.CLOSED`. You probably don't want to throw away client-side state. This could lose you poke n/ack handling, forces you to reestablish subscriptions, etc.

I'm not yet as familiar with this as I'd like to be, so I'm only 80% sure. @eglaysher care to take a look to confirm/deny my suspicions?

loganallenc: No, if the channel failed, I think you want to treat it just like you would a page refresh. You need the "superfluous" initial results to receive any information you lost while the eventSource was down. You want to run `onChannelError` after a new channel has been initialized so that you can be sure that you have a new Channel that you can work with to re-establish subscriptions, since this one will have been cleaned up already.

Fang-: Makes sense re call order. Maybe `onChannelReconnect` or similar would be more clear then.

> You need the "superfluous" initial results to receive any information you lost while the eventSource was down.

I was under the impression channel.js actively acks events as they come in, enabling Eyre to hold on to events we haven't received, to then send those on reconnect.

https://github.com/urbit/urbit/blob/cc16e905afda468afe0cebd36e1a713fb9f6fd74/pkg/arvo/sys/vane/eyre.hoon#L233-L241

loganallenc: Nope. If you look at the TODO in +channel-js in Eyre, it says: "Must send acks to the server"

loganallenc: This issue is separate from the current PR. The current PR is ready to be merged.

Fang-: > Nope. If you look at the TODO in +channel-js in Eyre, it says: "Must send acks to the server"

Oh, you're right. Agreed it's not _strictly_ part of "reconnect on errors", but not/having that todo in heavily influences implementation and behavior here.

Seems to me like we should make an action item for implementing that todo, and then (assuming this gets merged) revisit the work here to account for that actually working as intended. I'll make sure this gets noted somewhere.

loganallenc: Thanks, I completely agree.

ixv:
Does this need to be implemented for the other apps?

loganallenc:
It doesn't change the behavior for the other apps at all.

Issue #2671, opened 2020-04-03
Cannot open new notebook 


josh612:
**Describe the bug**

When attempting to open a new notebook, no response after clicking 'create notebook'

**To Reproduce**
1. Go to 'Publishing'
2. Click on 'new notebook'
3. Enter a name and description (optional; bug is the same without entering description)
4. Click on 'create notebook'

**Expected behavior**
A new notebook should be opened

**Screenshots**
If applicable, add screenshots to help explain your problem. If possible, please also screenshot your browser's dev console. Here are [Chrome's docs](https://developers.google.com/web/tools/chrome-devtools/open) for using this feature.

**Desktop (please complete the following information):**
 - OS: [e.g. MacOS 10.15.3]
 - Base hash of your urbit ship. Run ` .^(@uv %cz /=base=)` in Dojo to see this.

**Smartphone (please complete the following information):**
 - Device: [desktop]
 - OS: [ubuntu 19.10]
 - Browser [Bug reproduced in firefox and chromium]
 - Base hash 0vr.o87oq.g9lsl.l4ced.nau1s.uhn6h.e6la8.bbd8e.koaa3.u4tro.7ca6n but also seen in earlier hashes

**Additional context**
I was able to open a new notebook more than a week ago.  There doesn't appear to be any information in the javascript console when it is opened on the notebook page.  When asking about it in the Discord chat I was encouraged by a Tlon employee to fill out a github issue so here we are, in the thick of it together.

ixv:
Very strange. Can you run `:publish +dbug [%state '~(key by books)']` in the dojo and let me know if you see the notebook you tried to create in there?

josh612:
![Screenshot from 2020-04-04 10-52-41](https://user-images.githubusercontent.com/1130568/78417255-f0f9d980-7662-11ea-9de0-c797f39a08a9.png)

The notebook I tried to create is not there (although a couple of older notebooks I was able to create earlier appear to have showed up).

ixv:
Hm okay, your base hash is out of date by the way, but it shouldn't matter since publish hasn't really updated since then.

Can you run the following command in dojo?
`:publish &publish-action [%new-book %foo 'foo' '' %.y /~/foo ~ %.n %.n]`
This should make a notebook, so we can tell if this is a frontend or backend issue.

josh612:
[...]

PR #2724, opened 2020-04-13
publish: install mousetrap before we use it....

ixv:
This PR https://github.com/urbit/urbit/pull/2694 introduced mousetrap to publish without adding it to the package.json, resulting in this issue: https://github.com/urbit/urbit/issues/2671#issuecomment-612756244


matildepark:
A ha. I should’ve caught this in review. We’re going to need this in Links and Groups too, then.

ixv:
No, they have it already.

PR #2728, opened 2020-04-13
link: improve page load performance

Fang-:
By doing more or less the same thing that chat does, where it doesn't load in big backlog on page-load, but instead only does so when you actually go to view the resource.

Also does some smaller refactoring things. Cleans up the pagination logic in link-view a little bit, and refactors "message display" as used for the link fe landing page and "loading" messages into its own component.

### DIFF
--- a/pkg/arvo/app/link-listen-hook.hoon
+++ b/pkg/arvo/app/link-listen-hook.hoon
@@ -288,12 +288,30 @@
 ++  handle-metadata-update
   |=  upd=metadata-update
   ^-  (quip card _state)
-  ?.  ?=(%remove -.upd)  [~ state]
-  ?>  =(%link app-name.resource.upd)
-  =?  listening
-      ?=(~ (groups-from-resource:md %link app-path.resource.upd))
-    (~(del in listening) app-path.resource.upd)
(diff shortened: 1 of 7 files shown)


vvisigoth:
king
—
~poldec-tonteg
http://urbit.org

Fang-:
(Numbers observed here differ per the amount of collections you have, but my test case with eight collections went from ~10 seconds to <1 second in `+give-initial-submissions` for the intitial-pageload subscription result.)

matildepark:
This looks good to me but haven't tested yet.

ixv:
When I create a collection and invite invite someone, and accept that invite on the other ship, the collection doesn't then show up in that ships sidebar.

matildepark:
I can reproduce what @ixv is reporting — it's added to `associations`, but not `listening`.

Fang-:
Hmm, get the feeling this might be an unintended side-effect from #2708. We should still auto-opt-in to unmanaged groups I guess? Will take a deeper look and push a fix.

Fang-:
Kinda outside of PR scope, but went and pushed a fix here. Please take a look.

matildepark:
ping @loganallenc

PR #2733, opened 2020-04-14
chat: add firefox-specific chat window 

liam-fitzgerald:
supersedes: #2718 

Firefox treats the reflow in the scroll container weirdly so for the
scrollback case, we save the position to restore later, when the backlog
messages come in. For the 'locked at bottom' case we add a new state
variable and rescroll based on that.

cc: @matildepark @Fang-


Fang-:
It doesn't scroll quite to the bottom for me, which then causes it to not "stick" to the bottom on subsequent messages.

chat-scroll

(The flickering at the start is an unrelated issue which I'll file for now.)

Also the way chats start at the top is kind of jarring in that, after initial pageload, on taller screens, you _always_ have too few messages to fill the entire screen. Everything you're already seeing/reading then gets pushed to the bottom when scrollback loads in. Not a PR killer, but if it's possible to avoid, maybe we should do it?

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -78,12 +83,22 @@ export class ChatScreen extends Component {
...
         this.setState({ scrollLocked: false }, () => {
           this.scrollToBottom();
         })
       }
+    } else if (
+      navigator.userAgent.includes("Firefox") &&
+        props.length !== prevProps.length &&
+        state.scrollBottom
+    ) {
+      this.scrollToBottom();
```
matildepark, line 101: Ha, if this is here perhaps remove what I did on L91:L94 here — Firefox specific scrolling to bottom should be definitely cordoned off in a user agent flag so we can organise and remove it when we can properly simplify this.

matildepark:
> It doesn't scroll quite to the bottom for me, which then causes it to not "stick" to the bottom on subsequent messages.

So here's my question: is this because of pending messages having a different height? Wouldn't it fire twice: once when we send our message, then again when we receive it? Or are we doing a `delete` of the pending message that skews our scrollHeight?

matildepark:
I almost wonder if the Firefox scroll to bottom issue here isn't just a weird side effect of #2737, which should be solved first.

liam-fitzgerald:
This should work correctly, more testing welcome. It is a little bit janky, because it has to manually reset the scroll position every time the scroll container might have changed height. It also starts chats at the top, when they don't fill the top


Screen Shot 2020-04-17 at 3 20 15 pm

matildepark:
I'm gonna look at this today.

Issue #2731, opened 2020-04-14
Publish update comment stacktrace


liam-fitzgerald:
**Describe the bug**
Publish errors with `take:failed` when you try to edit a comment. If you try to send the edit poke over channel.js, vere crashes and dumps core.



**To Reproduce**
1. Create a new publish notebook
2. Create a new note 
3. Create a comment
4. Attempt to update comment through dojo e.g.
```
:publish &publish-action [~zod %notebook %note '~2020.4.14..03.42.16..f575.afaf.8594.2917' 'Baz']
```
**Expected behavior**
Should update correctly

**Stack Trace**
```
crud: %belt event failed
bail: exit
wire=/commit/home/ergo
  bar-stack
~[
  ~[
    / g
      use
      publish
      ~zod
      write
      ~zod
      home
      ~2020.4.14..03.59.37..0646.ae3a.3a8e.7147
      app
      publish
      notebooks
      test
      foo
      ~2020.4.14..03.42.16..f575.afaf.8594.2917
      publish-comment
    /d
    //term/1
  ]
  ~[
    /c/commit/home/ergo
    / g
      use
      publish
      ~zod
      write
      ~zod
      home
      ~2020.4.14..03.59.37..0646.ae3a.3a8e.7147
      app
      publish
      notebooks
      test
      foo
      ~2020.4.14..03.42.16..f575.afaf.8594.2917
      publish-comment
    /d
    //term/1
]                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                
  ~[
    / g
      use
      publish
      ~zod
      write
      ~zod
      home
      ~2020.4.14..03.59.37..0646.ae3a.3a8e.7147
      app
      publish
      notebooks
      test
      foo
      ~2020.4.14..03.42.16..f575.afaf.8594.2917
      publish-comment
    /d
    //term/1
  ]
  ~[
    /c/commit/home/checkout
    / g
      use
      publish
      ~zod
      write
      ~zod
      home
      ~2020.4.14..03.59.37..0646.ae3a.3a8e.7147
      app
      publish
      notebooks
      test
      foo
      ~2020.4.14..03.42.16..f575.afaf.8594.2917
      publish-comment
    /d
    //term/1
  ]
  ~[
    / g
      use
      publish
      ~zod
      write
      ~zod
      home
      ~2020.4.14..03.59.37..0646.ae3a.3a8e.7147
      app
      publish
      notebooks
      test
      foo
[...]

PR #2736, opened 2020-04-14
publish: fix bug in comment mark

ixv:
A fix for https://github.com/urbit/urbit/issues/2731

cc @liam-fitzgerald I would set you as the reviewer but thats not an option for some reason. Let me know if this works.


liam-fitzgerald:
Makes the draft PR work, so LGTM.

PR #2739, opened 2020-04-14
[WIP] Uruk Inlining

benjamin-tlon:
Cleaning this up before experimenting with inlining.


pkg/hs/urbit-uruk/example/mandelbrot.moon
```diff
@@ -661,8 +668,5 @@
   %+  turn  (mandelbrot w h)
   build-ppm-line
 
-:: Doing a 20x20 render takes 4m40s.
+:: This takes around 1m30s.
```
loganallenc, line 671: wild

PR #2740, opened 2020-04-15
chat: leave chats associated with group on group leave

liam-fitzgerald:
When a group is left, the permissions no longer exist and the mailbox
subscription is kicked. Before attempting to resubscribe, we simply
check if we still have permissions and if not, leave the chat.

cc: @matildepark

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -571,9 +571,6 @@
       [%mailbox @ *]
     ~&  mailbox-kick+wir
     ?.  (~(has by synced) t.wir)  [~ state]
-    ?.  (is-permitted our.bol t.wir)
-      :_  state
-      ~[(chat-view-poke %delete t.wir)]
     ~&  %chat-hook-resubscribe
     =/  =ship  (~(got by synced) t.wir)
     =/  mailbox=(unit mailbox)  (chat-scry t.wir)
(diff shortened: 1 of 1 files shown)


loganallenc:
@Fang- @urcades We could instead put these chats into “archive” mode. I’d be happy to work on such a feature, if we had a story around it, but I think it’s generally good to think through how users interact with “old” data.

matildepark:
I definitely email replied to this. Ah well:

> It’s a bit premature; some iteration of a fully considered contacts flow (including leaving associated channels when leaving a group and having leaving a group work) should be in place first. I wouldn’t hold this back because of that, though the archive work in a few weeks could iterate on this behaviour.

urcades:
I mentioned this in our standup, but the general note I'd like to affix to this issue is that this realm of functionality (handling how one interacts with a group) should be associated with the larger story of "How someone in Urbit navigates the internal and external structures of a Group" — this is less of an individual "feature" we should implement, and more of a networked series of affordances we embed into groups, as a larger project.

Matilde mentioned that this could be seen as a subset of a "fully considered contacts flow", and I'd more or less underline this statement, with the additional detail that the "Group" construct/abstraction as a whole is undergoing evolution — nothing that will throw a wrench into any of our implementation efforts I believe, but it's good to keep in mind the larger structure we're working with.

liam-fitzgerald:
bump @loganallenc

loganallenc:
This should go into the backlog subscription instead of the `chat-hook-remove`, not the mailbox subscription. Otherwise, looks fine.

jtobin:
@liam-fitzgerald note that a conflict has popped up on this one.

PR #2746, opened 2020-04-15
chat-hook: on %remove, try our best to pull subs

Fang-:
Previously, we were removing the relevant entry from the `synced` map
before calling `+pull-wire`, which requires an entry to still be there.
This lead to subscriptions not actually being pulled, commonly leading
to "subscribe wire not unique" errors on re-join.

In addition to fixing that, `%remove` actions now try to pull the
subscription regardless of whether they have an entry in the `synced`
map or not. `%leave` is always safe, and we might want to clean up
subscriptions that shouldn't be there anymore in the first place.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -358,8 +358,6 @@
       %+  weld  path.act
       ?~(mailbox /0 /(scot %ud (lent envelopes.u.mailbox)))
     :_  state
-    ::TODO  we shouldn't include the trailing /0 or similar in the wire,
-    ::      we never use it anywhere, and it complicates other logic. see #2746
     :~  [%pass chat-history %agent [ship.act %chat-hook] %watch chat-history]
         [%give %fact [/synced]~ %chat-hook-update !>([%initial synced])]
     ==
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -358,6 +358,8 @@
       %+  weld  path.act
       ?~(mailbox /0 /(scot %ud (lent envelopes.u.mailbox)))
     :_  state
+    ::TODO  we shouldn't include the trailing /0 or similar in the wire,
+    ::      we never use it anywhere, and it complicates other logic. see #2746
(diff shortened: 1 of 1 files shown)


loganallenc:
This is great. Possible to go even further by iterating through the subscription list. Keep in mind unsubscribing from [%backlog (weld path /0)] doesn’t necessarily remove the backlog subscription, as it could have been made on [%backlog ... /55] or some other number.

Fang-:
Isn't the real fix there to just not include the `/55` or w/e in the wire?

(That won't help with existing cases, so maybe we want both that _and_ "look at open subscriptions"?)

Fang-:
On second thought, that seems like slightly more invasive change: you'd want to change this logic, but keeping that backwards compatible is slightly non-trivial.

https://github.com/urbit/urbit/blob/65cad93b1b00215bd1e18a9b63666e9046beaf7a/pkg/arvo/app/chat-hook.hoon#L577-L578

loganallenc:
If you can come up with a lovely solution using a watch and then poke or some weird pattern like that to replace putting the number in the wire, feel free. It'd be more correct.

Fang-:
Oh, actually, we do use it. Implicitly, a couple lines down:

https://github.com/urbit/urbit/blob/65cad93b1b00215bd1e18a9b63666e9046beaf7a/pkg/arvo/app/chat-hook.hoon#L584-L586

(Maybe it's more appropriate to scry into chat-store for the `(lent envelopes)` again? But that's not quite correct either, shouldn't change anyway, and I'm much more comfortable not touching this/doing weird OTA logic here.)

This code is... not the clearest, imo, so I pushed a small patch that makes this use better naming and more appropriate runes here (and in the very similar `watch-ack` case below).

@loganallenc please review.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -358,22 +358,30 @@
       %+  weld  path.act
       ?~(mailbox /0 /(scot %ud (lent envelopes.u.mailbox)))
     :_  state
+    ::TODO  we shouldn't include the trailing /0 or similar in the wire,
+    ::      we never use it anywhere, and it complicates other logic. see #2746
```
loganallenc, line 362: This comment is untrue, as we do in fact use it.

PR #2750, opened 2020-04-15
dbug: avoid printing nothing at all

Fang-:
Previously, if filtering subscriptions turned up no results, nothing was
printed. With this change, we explicitly print "no matching subscriptions"
instead.

Wondering if we should do something to make the result stand out more, too. Since it gets printed _above_ the command in dojo scrollback, people seem to sometimes just miss it.


Fang-:
This also adds leading `> ` to every line, which I'm not a big fan of (wrt copy-pasting outputs), but this is certainly better than what we had.

PR #2754, opened 2020-04-15
link: leave collection on leaving associated group

liam-fitzgerald:
Automatically leaves a collection when the user is removed from a group.

cc: @matildepark

### DIFF
--- a/pkg/arvo/app/link-listen-hook.hoon
+++ b/pkg/arvo/app/link-listen-hook.hoon
@@ -344,9 +344,11 @@
   =*  loop-whos  $
   ?~  whos  loop-socs(socs t.socs)
   =^  caz  state
-    ?:  ?=(%remove -.upd)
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/link-listen-hook.hoon
```diff
@@ -399,8 +399,11 @@
 ++  leave-from-peer
   |=  [=app-path =group-path who=ship]
   ^-  (quip card _state)
+  ::  we left/got kicked from the group
+  ::  so leave collection
```
Fang-, line 403: Because this is logic specific to group changes, and not part of "ignoring a peer" per se, I think this is better at home in `+handle-group-update`'s handling of the `%remove` case? Right now the `+listen-to-*` and `+leave-from-*` arms deal exclusively with lower-level subscription management, all the "when to auto-join/leave" semantics happen in the update/action arms.

PR #2756, opened 2020-04-16
chat-js: input and render markdown in chat messages

liam-fitzgerald:
Renders markdown in text messages but not url or code messages. Swaps out the textarea for a CodeMirror in markdown mode. Resolves an issue related to the input shrinking with the content.

'me' and code messages disable the markdown highlighting. I had a crack at writing a codemirror mode for hoon, but decided that should probably come in a later PR. (haunted by dreams of jupyter-like functionality in publish).

I'm unsure whether or not we should allow the full range of markdown elements, what are your thoughts? It currently supports everything except links and images

cc: @matildepark  @urcades 

demo: 
Screen Shot 2020-04-16 at 8 28 54 pm
Screen Shot 2020-04-16 at 8 28 31 pm


loganallenc:
@Fang-  “me” messages should be removed and replaced with Markdown italics. They complicate the ontology of messaging unnecessarily.

urcades:
This is deeply cool to see — amazing job getting this working. 

I believe the best method we could adopt here for styling would be to limit our input to markdown that doesn't affect the line height of the text — limiting the styling capability here to strong/bold, italics, code fencing, strikethrough (if we had it), links (if it was possible) etc. would likely be the best way to keep the rendering of chat's messages sane and legible.

I actually think rendering quotes (and tall code blocks) would be incredibly cool as well, but I'd want to be sure that their representation as taller blocks of messaging felt right amongst other text content, and maintained a consistent-looking line height.

matildepark:
I can say right now that we’ll probably remove headers — we use them for Hoon code snippets. 

As to message types ... yes, this is a product discussion. They’re an old concept but we have to decide what we retain or not if we use styled content.

urcades:
> I can say right now that we’ll probably remove headers — we use them for Hoon code snippets. As to message types ... yes, this is a product discussion. They’re an old concept but we have to decide what we retain or not if we use styled content.
> […](#)
> On Apr 16, 2020, at 11:33 AM, Logan ***@***.***> wrote: ﻿ @Fang- “me” messages should be removed and replaced with Markdown italics. They complicate the ontology of messaging unnecessarily. — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

Yeah, all headers are absolutely out.

TBH, we should dismantle our current mechanism for hoon execution in chat and replace it with \```markdown code enclosure```\ that can be executed after it's been posted to chat. In the cli, we could perhaps adopt some 'triggering' glyph similar to our current usage of `;`, which would run code on a specified line.

Fang-:
Generally agree that we should only support the minimum viable markdown subset for expressive chat styling. Most messaging apps that support styling in the first place do _italics_, **bold**, and sometimes ~~strikethrough~~, not much else. 

> “me” messages should be removed and replaced with Markdown italics

> They’re an old concept but we have to decide what we retain or not if we use styled content.

Uh. I don't see how they relate to styled content. We happen to render them as italics messages in web UI, but that's largely arbitrary. Arguably they should be rendered more significantly different from regular messages. For example:

image

They're fundamentally a different thing than a regular chat message. Saying an emotive action is just a chat message seems like overaggressive flattening of our data structures around this.

> They complicate the ontology of messaging unnecessarily.

As per the above, I'd argue they aren't unnecessary. It's also not obvious to me how they _complicate_ the messaging ontology. Handling these isn't costly at all.

If you turn distinct message types into flat markdown, you remove the possibility for client agency over rendering decisions, or at least make that much more complicated. I see no reason to turn "chat as a protocol" into "chat as a pipe for sending markdown".

> TBH, we should dismantle our current mechanism for hoon execution in chat and replace it with \```markdown code enclosure```

I was actually going to suggest "remove `%url` and `%code` messages and just replace them with markdown" as a joke. I think this isn't good at all, for the reasons I mention above, and running posted code on demand instead of exclusively on the author's ship is a non-starter.

matildepark:
Personally my grievance is that @ is completely nonsensical as a /me syntax; /me is as old as IRC, people expect it, and we replace it with a nonintuitive, and thus forgotten thing. 

I’d retain “me” messages; style them specifically, not running them through the markdown processor (or maybe we just prefix them with specific markdown content and feed them into the same processor to be consistent); but use /me. 

> “But we don’t use /commands” 

Neither does hangouts or iMessage and yet they both have this.

Basically we want everything to be fed into the same processor for display. Maybe it means pre-processing received message types into markdown syntax in the client and throwing them into the display.

Fang-:
> Personally my grievance is that @ is completely nonsensical as a /me syntax; /me is as old as IRC

This is fair. We may or may not want to retain @ (alongside /me) for people who have already learned.

> Basically we want everything to be fed into the same processor for display.

What does this mean? Isn't this just [message.js](https://github.com/urbit/urbit/blob/master/pkg/interface/chat/src/js/components/lib/message.js)? Feeding _everything_ through the markdown parser as a last rendering step seems like it'd restrict your freedom as far as "how to render this" goes.

matildepark:
If you type /me ...

Strip /me from the message. 
Send it as a “me” type.

When you get it...
Add italics (*) to the content.
Render that content in the markdown viewer.

If you paste a link...

Send it separately as a URL message. 

When you get it...
If it’s not media, wrap it in markdown inline link syntax. 
Render that in the viewer.

If it is media,
If image, use markdown image inline.
If audio or video, inline link markdown syntax the URL, and attach our media viewer underneath for rendering an optional embed.

Keep the message itself client agnostic, is my point.

urcades:
> If you turn distinct message types into flat markdown, you remove the possibility for client agency over rendering decisions, or at least make that much more complicated. I see no reason to turn "chat as a protocol" into "chat as a pipe for sending markdown".

I'm actually in total agreement w/ this stance — I don't think of chats, and future conversation mechanisms, as being markdown pipes, more or less what I'm advocating for are better (possibly more standardized?) affordances for sending typed messages, rather than the `@` and `#` syntax that (afaik) is pretty idiosyncratic.

> running posted code locally & on demand instead of exclusively on the author's ship is a non-starter.

Regarding this, is this a nonstarter for entirely security reasons? I understand this concern but I'm curious to what degree we could mitigate security concerns ... imagining an environment (admittedly, a very naive environment) where one could demonstrate hoon in an interactive manner seems pretty cool ... but I understand entirely if it's too broad an attack surface.

Fang-:
> is this a nonstarter for entirely security reasons?

Basically yes.

We sandbox code execution of course, even currently for locally submitted code, so you can't accidentally scry your private key into a chat message or anything. But what if there's a vulnerability there?

Besides, sandboxing doesn't stop anyone from sending code snippets that take hours to execute, or cause trouble in some other way. As a user, I shouldn't be expected to know what is/isn't safe to run. So just make the author run their own code, and display the results provided by them. Never trust remote code.

liam-fitzgerald:
I dislike hiding message types behind weird syntax gates, imo it's super unintuitive and it's hard to use on mobile. We should just use a button on the right side of the input box to select a message type, like the below designs. Maybe use /me because it's a time-honored irc staple, but #/@ strike me as unnecessary
Screen Shot 2020-04-17 at 11 13 54 am

matildepark:
To clarify, the change this would mean is

@ -> /me
# -> “code” button

I mean, I dig that. The “image” button would incorporate the S3 work @loganallenc is doing and essentially return with a URL that it sends.

—
~haddef-sigwen
https://urbit.org

belisarius222:
I feel like an option here that we should keep in mind is keeping chat messages strictly plain text, and then having conventions for how clients render them, possibly including /me and inline markdown syntax.

—
~rovnys-ricfer
https://urbit.org

matildepark:
I said this.

> Keep the message itself client agnostic, is my point.

I don’t think anyone disagrees.

—
~haddef-sigwen
https://urbit.org

urcades:
Yeah, I'm pro-agnosticism with respect to the actual messages sent — I
would expect any "markdown" formatting applied via the OS1 client would
-not- be visible in a CLI interface, for example.

—
~fabled-faster
https://urbit.org/

liam-fitzgerald:
block level markdown elements are now disabled, and toggle for code editor has been implemented. 

Screen Shot 2020-04-17 at 9 30 00 pm
Screen Shot 2020-04-17 at 9 29 50 pm
Screen Shot 2020-04-17 at 9 29 44 pm
Screen Shot 2020-04-17 at 9 29 23 pm
Screen Shot 2020-04-17 at 9 29 14 pm
Screen Shot 2020-04-17 at 9 29 10 pm

pkg/interface/chat/src/js/components/lib/chat-input.js
```diff
@@ -275,14 +277,6 @@ export class ChatInput extends Component {
 
   messageChange(editor, data, value) {
 
-    const mode = editor.getOption('mode');
-    const specialFormat = value.startsWith('#') || value.startsWith('@')
```
Fang-: Can't we support both the button and the `#` prefix? I get that we're not going hard on keyboard shortcuts yet, but right now I can send some code in between talking about the code without taking my hands off the keyboard. We probably shouldn't degrade that into "mouse only".

liam-fitzgerald: Absolutely, will put it back in. I'd like to visit keyboard shortcuts more broadly in the near future though.

matildepark:
Imagine # toggled the code button if it’s the start of the input. We have mousetrap; the two “this is code” methods should interrelate.

liam-fitzgerald:
Merged master && retested with firefox, everything looks good.

matildepark:
@loganallenc iirc Gavin’s even written an indigo react component for markdown editors.

PR #2757, opened 2020-04-16
test: +expect-yes, +expect-crash

belisarius222:
Adds two test helpers:
- `+expect-yes` produces a non-null `$tang` if a boolean expression produces `%.n`
- `+expect-crash` produces a non-null `$tang` if the given trap doesn't crash

### DIFF
--- a/pkg/arvo/lib/test.hoon
+++ b/pkg/arvo/lib/test.hoon
@@ -1,6 +1,20 @@
 ::  testing utilities meant to be directly used from files in %/tests
 ::
 |%
+++  expect-yes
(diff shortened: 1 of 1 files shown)


lukechampine:
What's the difference between `expect-yes` and `expect`?

belisarius222:
Oh, nothing much... I didn't realize we had more than just `+expect-eq`.  My `+expect-crash` is also basically the same as the existing `+expect-fail`.  I think we'll want to add failure message arguments to these at some point, but I'll just close this for now.

PR #2758, opened 2020-04-16
chat-cli: account for reverse message ordering

Fang-:
#2618 changed the way chat-store orders messages. We need to account for that
when processing/printing. This is the minimum viable fix.

@loganallenc I'm 100% sure the inbox _scry_ has the reverse ordering, but the `%messages` update one I'm only 80% sure of. /sur file comments for ordering on these would be a nice aid.

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -235,7 +235,7 @@
   =*  path  p.n.inbox
   =*  mailbox  q.n.inbox
   =/  =target  (path-to-target path)
-  =^  cards-n  state  (read-envelopes target envelopes.mailbox)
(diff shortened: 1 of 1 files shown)


loganallenc:
%messages is reverse-ordered. It’s just a scry.

Fang-:
> It’s just a scry.

I'm... not sure I understand? It's somewhat dynamically constructed. See also:

https://github.com/urbit/urbit/blob/18878727b0f75c1c0f80cf518fa593b3ada788f5/pkg/arvo/app/chat-store.hoon#L221-L222

In any case, please review. (:

loganallenc:
Looks fine. When will you look into removing messages from state?

Fang-:
Does this appease you? cb01ff2

Spent 30 minutes working on rendering logic and have already had enough of it. Will return to this... Soon™.

PR #2762, opened 2020-04-17
various: set sigil foreground based on contrast

matildepark:
Sets sigil foregrounds to black when the [contrast ratio](https://contrast-ratio.com/) is 1.68 or lower. This essentially means it only flips to black when the background colour is *very* close to white:

Screen Shot 2020-04-17 at 7 13 53 PM
Screen Shot 2020-04-17 at 7 14 02 PM

But it lets you set the coveted 'inverted sigil', too:

Screen Shot 2020-04-17 at 7 31 19 PM


ohAitch:
hex color length 8 = RGBA?

matildepark:
You mean in the way it’s sliced? It’s because we pass # prefixed hex colours to the sigil component.

matildepark:
@liam-fitzgerald Solved this by restricting the sigil colour input to 6 valid, hexadecimal characters (0-9, a-f, case insensitive). Also linted contact-card.js. Please stamp

PR #2765, opened 2020-04-18
chat-fe: add backlog loading indicator

liam-fitzgerald:
Adds a backlog loading indicator that displays when we have not yet
received the backlog from the host.

Demo:
Screen Shot 2020-04-18 at 5 17 40 pm 

Screen Shot 2020-04-18 at 5 18 02 pm

### DIFF
--- a/pkg/interface/chat/src/js/components/chat.js
+++ b/pkg/interface/chat/src/js/components/chat.js
@@ -7,6 +7,7 @@ import { Route, Link } from "react-router-dom";
 import { store } from "/store";
 
 import { ResubscribeElement } from '/components/lib/resubscribe-element';
+import { BacklogElement } from '/components/lib/backlog-element';
 import { Message } from '/components/lib/message';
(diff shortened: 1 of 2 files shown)


liam-fitzgerald:
cc: @urcades

matildepark:
You can probably do this without creating a separate component — you can do it simply with the Spinner component as used elsewhere.

—
~haddef-sigwen
https://urbit.org

loganallenc:
I kind of like it being separated that way. Chat.js is big enough as is...

matildepark:
No, I mean, Spinner.js lets you pass classes and a “when to show this” condition such that the component seems equivalent. See the uses of it in Groups, etc.

—
~haddef-sigwen
https://urbit.org

jtobin:
@liam-fitzgerald a wild conflict has appeared

liam-fitzgerald:
r*based

PR #2771, opened 2020-04-20
meta: add note re: announcing default pill

### LINKED ISSUE #2735: boot pill hashes aren't listed anywhere
It would help with troubleshooting, and would reassure a lot of anxious new users, if the urbit-os version and hash of the boot pill was posted somewhere whenever it was updated. Maybe an announcement in urbit-dev when this was done?

jtobin:
Resolves #2735.

/cc @philipcmonk @ixv @joemfb @Fang- 

(A note: in general I don't think that we should change the pill corresponding to a Vere release after making it available, but this does happen from time to time.  In such cases we should just post the appropriate update to urbit-dev.)

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -119,6 +119,9 @@ this:
 ```
 urbit-vx.y.z
 
+Note that this Vere release will by default boot fresh ships using an Urbit OS
+va.b.c pill.
+
 Release binaries:
 
 (linux64)
@@ -138,9 +141,11 @@ Contributions:
 
 The same schpeel re: release candidates applies here.
 
-Do not include implicit Urbit OS changes in Vere releases.  This used to be
-done, historically, but shouldn't be any longer.  If there are Urbit OS and
-Vere changes to be released, make two releases.
+Note that the release notes indicate which version of Urbit OS the Vere release
+will use by default when booting fresh ships.  Do not include implicit Urbit OS
+changes in Vere releases; this used to be done, historically, but shouldn't be
+any longer.  If there are Urbit OS and Vere changes to be released, make two
+separate releases.
 
 ### Deploy the update
 


philipcmonk:
Agree that the vere release notes are probably a good place for it.  Also agree that we should probably avoid updating the pill too much, but I do think there are legitimate reasons to upgrade the pill independent of vere versions.  When we do this, maybe an ANNOUNCE email is appropriate, and maybe a pill tag?  What we really should do is make it so the content in the pill matches one of the OTAs.  This is simply a matter of making sure the pills and OTAs have the same policy regarding including or excluding README and a couple other similar files.

We've a couple instances recently where we needed to release a new pill.  I recall three:

- OS1.  When we released OS1, there was no corresponding Vere release (the pill for 0.10.4 was the "appetizer" OTA, which was still on OS0).  We wanted a new pill because we didn't want people to boot into OS0 and then wait to be upgraded to OS1 until they got the update from their sponsor.  This is a bad user experience, but also we didn't know if there would be issues from an upgrade that large, and we didn't want to have everyone showing up and immediately running into upgrade troubles.
- Post-OS1 fixer-upper.  Something was wrong with the OS1 pill that was pushed.  I forget exactly what, but it made Urbit effectively unusable for new users, so we had to push a new pill.  This was my bad, I should have tested the new pill on the live network instead of just a fakezod.
- Bad OTA.  One of the OTAs commonly dropped people into the "404" issue and the next one didn't always get them out of it.  We pushed a new pill to stop the tide of people running into that.

Fang-:
> we should probably avoid updating the pill too much

Can you elaborate on why this is? You describe the OS0->OS1 case, where the difference between "pill" and "latest" would've been huge, but I don't see why that (desire to give people latest right away) wouldn't apply to smaller changes and bugfixes too.

jtobin:
> Can you elaborate on why this is?

Yeah, now that I think on it: I've forgotten the *precise* reason we avoid doing this.

IIRC, the problem with updating the pill was that a ship would boot with pill version X, but then still receive an OTA for any available version < X from its sponsor; this would be problematic for all the reasons that OTAs are currently problematic.  But the exact mechanics here escape me.

jtobin:
Also agreed with @philipcmonk that at present there are all sorts of good reasons to update the bootstrap pill from time to time, as listed in his comment.

joemfb:
Our state transitions only run forward. If OTA's were atomic, this would just mean that "reverts" would fail. As it stands, if your sponsor is behind the pill you used to boot, all bets are off.

PR #2776, opened 2020-04-21
contacts: added "url" avatar support and finished "octet" avatar fetching support

loganallenc:
@urcades As soon as you're ready, let's collab on a front-end for this! Shouldn't be too hard.


pkg/arvo/sur/contact-store.hoon
```diff
@@ -1,10 +1,26 @@
 /-  *identity
 |%
-+$  rolodex  (map path contacts)
++$  rolodex-0  (map path contacts-0)
```
Fang-, line 3: Currently, we never expose older-version data structures in `/sur` files, instead always containing them in the `/app`, sometimes even in _just_ the `+on-load`. Is there a reason these need to be public, or can we hide them away?

pkg/arvo/lib/contact-json.hoon
```diff
@@ -15,46 +15,62 @@
...
+    :-  %s
+    %-  crip
+    %-  zing
+    :~  "/~groups/img"
+        (trip (spat path))
+        "/"
+        (trip (scot %p ship))
+        "."
+        (trip file-extension.u.avat)
+    ==
```
Fang-, line 71: Can you explain how the `%octt` case of `+avatar` is intended to be used? I'm confused.

pkg/arvo/sur/contact-store.hoon
loganallenc: Good point.

pkg/arvo/lib/contact-json.hoon
loganallenc: Yes. In the first version of this that we release, I’d like to only support %url on the front-end due to space-constraints in vere. However, ideally, you’d encode an image as base64, upload it to your urbit as a poke, it would be decoded from base64 in the %json mark, then stored in contact-store. When converting back to %json, you pass a relative URL through which the binary file can be downloaded by the client. I’ve tested this whole flow and it works, just want to hide it away for a little while until the space-constraints are less salient.

Fang-: I see. Thanks!

Might want to consider just leaving out this case entirely from implementation, for the time being. "Half-in and intentionally not in use" isn't the best state to put things in imo.

Also, when this does go in, we may want to write these to clay instead of storing in state. Both to provide savvy users more control, and to play nicely with standardized filesystem serving that Eyre's slated to get.

(We'll probably want to impose a limit on the size of avatars we accept regardless of memory pressure?)

loganallenc: Disagree, as I view Clay’s future as being a filesystem exclusively for source files.

loganallenc: I also think it’s a good experimental capability for us to have, and disabling from the front-end is just fine.

loganallenc:
@Fang- @matildepark pls review, would love to get this merged

matildepark:
Little for me to review, you just delete some old boilerplate in FE.

urcades:
Yo @loganallenc !

Regarding the way we could adapt the Drag n' Drop pattern described in the chat S3 upload flow, an incredibly dumb way to implement this feature without modifying the existing look and feel of the page would be to do something like this:

Avatar Upload

Top row describes clicking the "upload image" button, Bottom row describes a drag and drop action into the profile area.

The thing is, is that I kind of hate the way the current profile is structured, and given how we'd like to integrate some "settings"-like functions into a combined Profile and ID section of OS 2, I've been experimenting with the following:

Screen Shot 2020-04-22 at 12 14 55 PM

Screen Shot 2020-04-22 at 12 14 36 PM

I paste these here not to detract from the actual thing we're working on, but to give a sense of how I can see the interaction evolving — basically having avatar editing capability be more obviously co-located with the representation of the avatar itself, instead of a janky form that clutters up the rest of the profile information.

If it's not too much additional implementation work, I think it'd be worth exploring a similar interaction for OS 1.1/2, where we embed this capability into the avatar-representation itself.

loganallenc:
These are beautiful! In the near-term, I envision the first release that we do of avatar support only including "URL" avatars, not the full drag and drop experience. I'm curious as to specific guidance on how to do that interaction of setting a URL, rendering a preview, and allowing unsetting a URL in the existing contact card.

pkg/arvo/app/contact-view.hoon
```diff
@@ -181,21 +181,16 @@
...
-::    =/  decoded  (de:base64 q.octs.avatar)
-::    ?~  decoded  not-found:gen
-::    [[200 ['content-type' content-type.avatar]~] `u.decoded]
+      [%'~groups' %avatar @ *]
+    =/  =path  (flop t.t.site.url)
+    ?~  path  not-found:gen
+    =/  contact  (contact-scry `^path`(snoc (flop t.path) name))
+    ?~  contact  not-found:gen
+    ?~  avatar.u.contact  not-found:gen
+    ?.  ?=(%octt -.u.avatar.u.contact)  not-found:gen
```
Fang-, line 190: Wondering if we could serve a redirect or something in case it's a `%url` avatar?

loganallenc: Hmm... Good idea!

Fang-:
Neat!

loganallenc:
@matildepark Ready for morge

urcades:
@loganallenc : 

Each row represents a slightly different image-saving model: 

Top row is a "save on field blur"-style experience, bottom row requires explicit button press to save the avatar state. The two of these require a slight modification of the sigil/avatar location, left-aligned to best make room for the fields, and to possibly slowly nudge the interface to an OS 2-like state.

Avatar Upload II

loganallenc:
Your wish is my command. On it!

matildepark:
Unff, they look so good.

L — remember that this will require appropriate integration in every app FE too; they all currently hardcode sigils, though I imagine the code to swap them out is pretty easy. Just saying, if you were trying to get that out ASAP, you can still just leave the FE to me.

— 
~haddef-sigwen
https://urbit.org

loganallenc:
Appreciate it Matilde! I’m planning on doing the contact card form tomorrow, but if you wouldn’t mind doing the Sigil / avatar integration afterwards that would be lovely.

PR #2777, opened 2020-04-21
publish: groupify notebooks

ixv:
This PR adds the ability to turn a notebook into a group, or attach it to an existing group.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -1702,7 +1702,87 @@
     :~  [%give %fact [/primary]~ %publish-primary-delta !>(act)]
         [%give %fact [/publishtile]~ %json !>(jon)]
     ==
+  ::  %groupify
+  ::
+      %groupify
+    ?.  (team:title our.bol src.bol)
+      ~|("action not permitted" !!)
+    =/  book  (~(get by books) our.bol book.act)
+    ?~  book
+      ~|("nonexistent notebook: {<book.act>}" !!)
+    ::
+    =/  old-write      writers.u.book
+    =/  old-read  subscribers.u.book
+    ?>  ?=([%'~' ^] old-write)
+    =/  destroy-old-groups=(list card)
+      :~  (group-poke [%unbundle old-write])
+          (group-poke [%unbundle old-read])
+          (group-hook-poke [%remove old-write])
+          (group-hook-poke [%remove old-read])
+          (perm-group-hook-poke [%dissociate old-write (sy old-write ~)])
+          (perm-group-hook-poke [%dissociate old-read (sy old-read ~)])
+          (perm-hook-poke [%remove old-write])
+          (perm-hook-poke [%remove old-read])
+      ==
+    ::
+    ?~  target.act
+      :: create new group from subscribers
+      =.  writers.u.book      (slag 1 writers.u.book)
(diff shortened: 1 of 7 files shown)


pkg/interface/publish/src/js/components/lib/settings.js
```diff
@@ -84,6 +89,112 @@ export class Settings extends Component {
...
+      disabled: true,
+      type: 'Converting'
+    }, (() => {
+      window.api.action("publish", "publish-action", {
+        groupify: {
+          book: props.book,
+          target: state.targetGroup,
+          inclusive: state.inclusive,
+        }
+      }).then(() => this.setState({awaiting: false}));
```
matildepark, line 118: We don't use `awaiting` in state for this view, we use `disabled`. (These are used to disable other fields until the poke returns and to in FE mark when to show the `Spinner` component).

pkg/interface/publish/src/js/components/lib/settings.js
```diff
@@ -84,6 +89,112 @@ export class Settings extends Component {
...
+
+  changeInclusive(event) {
+    this.setState({ inclusive: !!event.target.checked });
+  }
+
+  groupifyNotebook() {
+    const { props, state } = this;
+
+    this.setState({
+      isLoading: true,
```
matildepark, line 108: We don't use `isLoading` here either, this can be removed.

pkg/arvo/app/publish.hoon
```diff
@@ -1702,8 +1702,88 @@
...
+          [%give %fact [/primary]~ %publish-primary-delta !>(del)]
+          (group-poke [%add dif-peeps u.target.act])
+      ==
+    :: kick subscribers who are not already in group
+    =/  del=notebook-delta  [%edit-book our.bol book.act u.book]
+    :_  state(books (~(put by books) [our.bol book.act] u.book))
+    %+  weld  destroy-old-groups
+    ^-  (list card)
+    :+  [%give %fact [/notebook/[book.act]]~ %publish-notebook-delta !>(del)]
+      [%give %fact [/primary]~ %publish-primary-delta !>(del)]
```
Fang-, line 1770: It looks like these lines are equivalent between the "add new subscribers to group" and "kick subscribers not in group" cases. If so, you could just move the `?:` into the tail of the `:+` here and cut down a bunch of LoC.

Fang-:
(Also, don't be afraid to let your code breathe a bit. Nothing wrong with double or multi-line `=/`s, and the "flag-style" comments have started growing on me.)

jfranklin9000:
Just curious, what's a 'flag-style" comment?  Can you post a small example?

Fang-:
@jfranklin9000 Just giving the comments an extra line of `::` underneath them. This helps space out the code a bit more, making for a more obvious "section break"/descriptor.

```hoon
=+  some=code
::  next section does this
::
(do this)
```

ixv:
Addressed both of your reviews and fixed an issue where we weren't sending metadata updates.

pkg/arvo/app/publish.hoon
```diff
@@ -1741,33 +1740,53 @@
               title.u.book
               description.u.book
           ==
+          %-  metadata-poke
```
Fang-, line 1743: You probably want backstepped `::` around your long-form `:~` entries at this point. Same applies below.

PR #2779, opened 2020-04-21
gall: clean up API, remove cache

belisarius222:
Gall was full of vacuous worms.  I also cleaned up its $task interface and removed the |wash-gall generator, which was only useful as a placebo.


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1575,52 +1549,18 @@
...
-    ::
-    =>  (mo-boot:initialised q.dock q.task)
-    mo-abet
-  ::
+      %conf  mo-abet:(mo-boot:mo-core dap.task our %home)
       %deal
-    =/  =sock  p.task
-    =/  =term  q.task
-    =/  =deal  r.task
+    =/  [=sock =term =deal]  [p q r]:task
```
lukechampine, line 1556: this is neat -- Hoon-style "destructuring!"

belisarius222: Yeah, you can even do more complex tree shapes, as long as they're finite:
```hoon
=/  [[=ship =desk =case:clay] =spur]  *beam
`path`[(scot %p ship) desk (scot case) (spud (flop spur))]
```

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1575,52 +1549,18 @@
...
-    =/  =dock  p.task
-    =/  =ship  p.dock
-    ?.  =(our ship)
-      ~&  [%gall-not-ours ship]
-      [~ gall-payload]
-    ::
-    =>  (mo-boot:initialised q.dock q.task)
-    mo-abet
-  ::
+      %conf  mo-abet:(mo-boot:mo-core dap.task our %home)
```
joemfb, line 1554: This removes the ability to run apps anywhere other than `%home` -- that seems pretty drastic.

belisarius222:
We don't have to deploy this, or even merge it. I mostly wanted to separate it out as a unit of work in case we did want to do deploy it. Since OTAs are risky these days, probably better not to do it. I could close this.

A couple weeks ago, Philip and I decided to remove the multi-desk app building at least temporarily to make it easier to make OTAs atomic -- atomicity across desks is nontrivial.

We'll definitely want the ability to run user code on multiple desks in the future, but we don't use that feature now, so we thought we'd get to working atomicity, then figure out how to run agents on multiple desks.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
I don't think this can be OTA'd without adapting the running agents, since they're covariant in `+note-arvo`.  This is straightforward (throw away `%wash` and reduce `%conf`), but it also won't be necessary in Ford Fusion, since we'll be rebuilding all apps when gall updates.

The rest of it looks fine.  The question of how to run apps on other desks is not resolved.  For example, if the definition of `+sign-arvo` or `+note-arvo` changes in one desk, how can the same gall handle both agents?  Further research is needed, so in the meantime, I think it's best to not even attempt to do it.

belisarius222:
I think I'll just close this, and it can just go in as part of the Ford Fusion work.

PR #2784, opened 2020-04-23
channel-js: fixed long-standing issue with subscription errors not being propagated

loganallenc:


matildepark:
Is there a pre-existing issue here (filed or not)? What is this doing and fixing?

loganallenc:
There's not a pre-existing issue filed, but the issue has existed since the original eyre-light upgrade. I think I'm the only person who noticed the behavior, but the gist is that subscription errors don't actually end up calling the "err" function of the particular outstandingSubscription because the error is received in the `connectIfDisconnected` method as a poke error. I fixed this with the above code and now it works properly every time. You can easily recreate the behavior by breaking the chat-json library or another JSON conversion library and then noting that an exception is thrown in the JS console but the "err" function that you set in subscription.js is not called.

Fang-:
> the error is received in the `connectIfDisconnected` method as a poke error

I don't understand what you mean by this. Can you elaborate?

loganallenc:
obj.response === 'poke' evaluates to true despite the error actually corresponding to a subscription.

Fang-:
I see. The only place Eyre sends `'poke'` results is [when forwarding a `%poke-ack`](https://github.com/urbit/urbit/blob/a4fb1a12dea4fd6e78cbcbb827398276b52f8edb/pkg/arvo/sys/vane/eyre.hoon#L1601). idk how the behavior your describe can happen, but seems like an Eyre bug, not a channel.js one, right?

loganallenc:
Probably an `%eyre` bug, yeah. This fixes the behavior in channel.js now, though. Perhaps the bug can be logged for `%eyre` and this can be reverted afterwards, but this is merge-ready and fixes the problem.

Fang-:
Actually, cannot reproduce.

image

Putting a `!!` in one of the json conversion arms just gives me the above which correctly comes in as a 'subscribe' response. Am I repro-ing wrong?

(Should also note that current channel.js behavior instantly retries on getting that error, leading to a _very_ fast-paced stream of subscription attempts being sent to the urbit. It should... back off, maybe?)

loganallenc:
plz morge. the backoff should occur in subscription.js

loganallenc:
Ping.

Fang-:
Still haven't been able to reproduce this, and still think the fix isn't The Right One, but don't let me stop you I guess...

loganallenc:
Thanks.

PR #2787, opened 2020-04-24
groups-js: added ability to set avatar images

loganallenc:
This could be a little prettier, but it's entirely functional and matches the style / uses the same elements as the rest of the contact card.


matildepark:
You didn't say, so I finished picking up from here, I'll slate it on another branch.

PR #2788, opened 2020-04-24
os1: set and show avatars in all apps

matildepark:
Supersedes #2787.

Adds the display for set avatars in all apps on top of the FE setting of avatars. Tested in unmanaged and managed channels, each with set or unset avatars as applicable, for all apps, and it looks fine.

Lints the files we touch, so sorry for the diff.

### DIFF
--- a/pkg/interface/groups/src/js/components/lib/contact-card.js
+++ b/pkg/interface/groups/src/js/components/lib/contact-card.js
@@ -116,7 +116,7 @@ export class ContactCard extends Component {
         if (
           (state.avatarToSet === '') ||
           (
-            !!props.contact.avatar &&
+            Boolean(props.contact.avatar) &&
             'url' in props.contact.avatar &&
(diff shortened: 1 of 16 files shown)


pkg/interface/groups/src/js/components/lib/contact-card.js
```diff
@@ -200,14 +235,6 @@ export class ContactCard extends Component {
...
         break;
       }
-      case 'removeAvatar': {
-        this.setState({ awaiting: true, type: 'Removing from group' }, (() => {
-            api.contactEdit(props.path, ship, { avatar: null }).then(() => {
-              this.setState({ awaiting: false });
-            });
-        }));
-        break;
-      }
```
matildepark, line 210: Note that we had two cases for the removeAvatar case introduced in 15404e0989683151ffe0fe8daf6323e9059863e6, heh. We just remove the stubbed one I wrote originally here.

PR #2553, opened 2020-03-19
gall: remove pending messages from breached ship

philipcmonk:
See #2485 for context.

When a ship breaches, we remove all messages that have yet to be
delivered to an app (eg if it's not yet started).  We also add
|gall-sear to do this manually, but this shouldn't be needed in normal
operation.

Finally, to unblock ~zod and ~bus on mainnet, we sear one particular
ship automatically on loading hood.  It cannot be done manually because
no userpace changes can be made until it's unblocked.

Tested on a localhost copy of ~bus.

PR #2800, opened 2020-04-24
Minimal S3 support

loganallenc:
s3-store supports:
- being configured to have all of the credentials needed to access S3 buckets
- being configured to set "currently active bucket"
- local subscriptions
- Integration with the Groups application
- Integration with the Chat application

Nice to have eventually:

a storage tile
a mini-app interface for configuring your S3 based on pre-set templates (AWS settings, DigitalOcean settings, GCloud, etc)


pkg/arvo/mar/s3/action.hoon
```diff
@@ -0,0 +1,8 @@
+/+  *s3-json
+|_  act=action
+++  grab
+  |%
+  ++  noun  action
+  ++  json  |=(jon=^json (json-to-action jon))
```
Fang-, line 6:
```suggestion
  ++  json  json-to-action
```

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -223,7 +224,7 @@
   ==
 ::
 ++  on-load
-  |=  ver=?(%1 %2)
+  |=  ver=?(%1 %2 %3)
```
Fang-, line 227: This `+on-load` is... quite something. At the very least you'll need to update `/app/hood` to actually call this.

Should probably, for `%3`, only start the s3-store? Though I guess we can't guarantee that the others didn't get skipped over, because of how OTA fast-forward right now...

philipcmonk: The usual pattern is a series of

```
=?  state  =(ver %1)
  ...(ver %2)
=?  state  =(ver %2)
  ...
```

People do in practice skip large numbers of commits.  On mainnet right now the only update that has to be staged is the compiler upgrade in january.  You can take the original pill, jump to that commit, then jump to the present, and it'll just work.

loganallenc: Can someone verify that I updated hood properly? I think I did and it works well, but it's always nice to have another pair of eyes.

philipcmonk: I don't think so.  I think you need to add another case to +on-load in app/hood or else this won't get called.  There's no magic that calls +on-load in drum right now, we just manually call it as necessary.  Also the overall state in app/hood is what needs its version number bumped.  Basically, the flow is:

- Bump app/hood's whole state's version number in +on-save and +hood-old
- In app/hood's +on-load, use this version number change as a trigger to do anything you need to do.  In this case, that's call +on-load in lib/hood/drum
- In lib/hood/drum, start the apps

loganallenc: Thanks. I've modified it. I tested it by booting a fakezod on master, then copying in the changes and saw `%hood` successfully reload with the `%s3-store` started.

pkg/arvo/mar/s3/action.hoon
loganallenc: Done! Love it, will do the same for other apps in the near-term.

loganallenc:
Ping @ixv @Fang-

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -250,7 +252,22 @@
     =<  (se-born %home %link-store)
     =<  (se-born %home %link-proxy-hook)
     =<  (se-born %home %link-listen-hook)
-    (se-born %home %link-view)
+    =<  (se-born %home %link-view)
+    (se-born %home %s3-store)
+  ::
+      %3
+    =<  se-abet  =<  se-view
+    =<  (se-emit %pass /kiln %arvo %g %sear ~wisrut-nocsub)
```
Fang-, line 260: This... clears pending queues for... some planet under ~litzod? I see this was done for previous version transitions, but I have no idea why. Was it just a weird hack that got cargo-culted forward?

philipcmonk: See #2553.  Was used to resolve a particular issue on a couple galaxies, but can be removed now.

matildepark:
I intend to test today btw, so hold a sec.

loganallenc:
@matildepark Can I get a morge?

matildepark:
I have an s3 bucket lol, I can DevOps with the best of em, unless you mean Urbit-side.

loganallenc:
Okay. You have to set a ton of parameters to get it. The bucket has to be public-read, allow CORS from * origins, and allow * headers. You have to on the Urbit side set the endpoint, the bucket, and two different API keys.

Fang-:
Might be nice to get those setup instructions commented into the s3-store for the record, and then docs/config UI at some later point in time.

pkg/interface/chat/src/js/components/lib/chat-input.js
```diff
@@ -345,16 +360,23 @@ export class ChatInput extends Component {
             onChange={(e, d, v) => this.messageChange(e, d, v)}
           />
         </div>
+        <div className="ml2"
```
matildepark, line 363: These images are too big! And too close. Compare with mock with this:

image

image

- The two images should be 16x16, not 24x24. 
- The codeEval image needs to be padded.
- The image here doesn't have the multi-shade bg, it probably wasn't included in the export. Also ensure the export is 4x.

You'll also want to ensure the two buttons are consistent; probably they should both be white:

image

matildepark:
Also important to note:

- This implementation only works with [endpoints that use signature v2](https://docs.aws.amazon.com/general/latest/gr/s3.html), not v4, when using AWS S3.
- For some reason clicking my uploaded image from the chat prompts my browser to download it. But I view it in Chat fine when it's directly embedded. Weird.

pkg/interface/chat/src/js/components/lib/chat-input.js
loganallenc: @urcades Can I get an image asset with squared off borders?

urcades: Hopefully this can be accounted for in this issue/feature as a small patch to another feature, but the latest iteration of these that Gav refined actually retain a small radii on 'em:

icn-img

Additionally, our current icon for "code" is incorrect in that it states "centis" is a rune, when it most definitely is not. If we can replace this with a more standardized `>` symbol, that'd be wonderful:

icn-code

If this additional replacement is out of scope for this issue, I'm happy to add this in and PR myself!

urcades: So actually, this might be best expressed that rather than replace the "send image" button with a squared off version, we should replace the actually-incorrect "send code" button to normalize it with this work currently being done.

Fang-: > our current icon for "code" is incorrect in that it states "centis" is a rune, when it most definitely is not

It most definitely is! You might be remembering my feedback about a previous iteration (maybe of the dojo tile?) that used `=%`, which is indeed not a rune.

(sad to see `%=` go...)

urcades:
Oops, my bad — reversing my distinction of that event, thanks for
correcting me! Regardless, for the reason that runes are a little hard to
see when rendered as 16x16 icons, we came to the conclusion in a review
long ago that maintaining a `>` for "running" code in an interface like
this might be our best bet.

—
~fabled-faster
https://urbit.org/

pkg/interface/chat/src/js/components/lib/chat-input.js
loganallenc: responded to design comments

loganallenc:
Screen Shot 2020-05-01 at 7 07 49 PM

loganallenc:
@matildepark nothing is blocking a merge now

matildepark:
@loganallenc Would you mind adding brief documentation on use for posterity?

loganallenc:
I’d like to work with the docs team on that as opposed to including it in the code if you don’t mind.

matildepark:
Alright, suit yourself

Issue #2248, opened 2020-02-07
add a startup tip saying to press ctrl-x to switch apps


reitraced:
i know this idea might just put the urbit discord out of business but i think it would be beneficial to add like a startup message on the initial boot that reminds you how to switch apps
something like:
`tip: use ctrl-x to switch between chat-cli and dojo`
idk
is it even possible?

galenwp:
I like this idea. There also used to be a +help generator that would print
some helpful text. Could be worth reviving that if it has fallen into
disrepair.

belisarius222:
Oh, we could even include an arm called "help" in Zuse, or injected into the dojo subject by dojo itself, that would resolve to some help text.  That way someone could type "help" into dojo and actually get something useful.  This should be pretty easy to add.

—
~rovnys-ricfer
https://urbit.org

jtobin:
`+help` still exists, and is somewhat useful, but it could definitely use some minor TLC.

reitraced:
revisiting this, still very much want it
being a guide in the discord has made me appreciate this idea more
like 9/10 lost souls who stumble into #ship-starting-support saying their ships hanged on boot are actually just in chat-cli and dont know how to switch to dojo

botter-nidnul:
Even just fixing the docs at [https://urbit.org/using/install/](https://urbit.org/using/install/) would help some - it tells people that `When your ship is finished booting, you will see the ~sampel-palnet:dojo> prompt.` Does this ever happen currently, or is it 100% chat-cli when boot finishes?

botter-nidnul:
https://github.com/urbit/urbit.org/pull/379

reitraced:
ayyy nice on that pr!
lets see what comes of this

Fang-:
> Does this ever happen currently, or is it 100% chat-cli when boot finishes?

It, uh, depends on your `@p`, iirc? It's a drum boot order thing, which goes through a set containing "`%app` on `~ship`" descriptions. Set ordering depends on the contents, and so this is gonna vary between ships.

loganallenc:
Ping

Issue #2681, opened 2020-04-06
Chat-cli subscribe error on boot


philipcmonk:
Getting error `subscribe failed from %chat-cli on wire /invites` when booting for the first time.  Looks like invite-store no longer accepts this path but chat-cli still asks for it on boot.

Base hash: 4a6ue (mainnet pill) and eh7oh (solid pill for urbit-os-v1.0.11)

matildepark:
@Fang- @loganallenc I feel like this got fixed somewhere but can't find the PR if it did.

Fang-:
Uh, chat-cli subscribes on a sane path:

https://github.com/urbit/urbit/blob/39c707e596d2be0fc921d34ef765fb54b480b74b/pkg/arvo/app/chat-cli.hoon#L248-L250

But I'm guessing because of boot order shenanigans, the "invitatory" with that name hasn't been created when this happens, and invite-store rejects the subscription because of that. Seems like the same class of underlying issue as https://github.com/urbit/urbit/issues/2248#issuecomment-601639995 has.

PR #2805, opened 2020-04-24
drum: sane boot order

liam-fitzgerald:
Fixes issues related to drum boot ordering
- Changes drum's boot order to be deterministic regardless of the ship
you're on.
- Sorts cli connections so that the user always boots into %dojo
regardless of their @p

Fixes: #2248 #2681

### DIFF
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -356,7 +356,7 @@
   ++  priorities
     ^-  (list (set @))
     :~
-      (sy ~[%dojo])  ::  ensure dojo connects first
+      :: set up stores with priority: depended on, but never depending
       %-  sy
       :~  %permission-store
           %chat-store
@@ -366,7 +366,8 @@
           %invite-store
           %metadata-store
       ==
-      (sy ~[%chat-hook])  :: ensure chat-cli can sub to invites
+      :: ensure chat-cli can sub to invites
+      (sy ~[%chat-hook])
     ==
   ++  sort-by-priorities
     =/  priorities  priorities
@@ -389,7 +390,10 @@
   %+  roll
     %+  sort
       ~(tap in eel)
-    |=([[@ a=term] [@ b=term]] (aor a b))
+    |=  [[@ a=term] [@ b=term]]
+    ?:  =(a %dojo)  %.n
+    ?:  =(b %dojo)  %.y
+    (aor a b)
   =<  .(con +>)
   |:  $:{gil/gill:gall con/_.}  ^+  con
   =.  +>.$  con
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -332,26 +332,31 @@
 ++  se-adit                                           ::  update servers
   ^+  this
   |^
-  %+  reel
+  =/  servers=(list well:gall)
     (sort ~(tap in ray) sort-by-priorities)
-  |=  [=well:gall that=_this]
-  ^+  that
-  =/  =wire  [%drum p.well q.well ~]
+  |-
+  ?~  servers
+    this
+  =/  wel=well:gall
+    i.servers
+  =/  =wire  [%drum p.wel q.wel ~]
   =/  hig=(unit (unit server))
-    (~(get by fur) q.well)
-  ?:  &(?=(^ hig) |(?=(~ u.hig) =(p.well syd.u.u.hig)))
-    that
+    (~(get by fur) q.wel)
+  ?:  &(?=(^ hig) |(?=(~ u.hig) =(p.wel syd.u.u.hig)))
+    this
   =.  fur
-    (~(put by fur) q.well ~)
-  =.  that
-    (se-text "activated app {(trip p.well)}/{(trip q.well)}")
-  %-  se-emit
-  [%pass wire %arvo %g %conf [our.hid q.well] our.hid p.well]
+    (~(put by fur) q.wel ~)
+  =.  this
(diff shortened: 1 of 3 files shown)


pkg/arvo/lib/hood/drum.hoon
```diff
@@ -386,7 +386,10 @@
 ::
 ++  se-adze                                           ::  update connections
   ^+  .
-  %+  roll  ~(tap in eel)
+  %+  roll
+    %+  sort
+      ~(tap in eel)
+    |=([[@ a=term] [@ b=term]] (aor a b))
```
Fang-, line 392: > always boots into `%dojo`

until we add `%zork` or other later-in-alphabet app to the default connections list. Might be worth making `%dojo` explicitly last here, so we don't run into this again.

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -329,23 +330,66 @@
...
-  =.  +>.$  con
-  =+  hig=(~(get by fur) q.wel)
-  ?:  &(?=(^ hig) |(?=(~ u.hig) =(p.wel syd.u.u.hig)))  +>.$
-  =.  +>.$  (se-text "activated app {(trip p.wel)}/{(trip q.wel)}")
-  %-  se-emit(fur (~(put by fur) q.wel ~))
+  ^+  this
+  |^
+  =/  servers=(list well:gall)
+    (sort ~(tap in ray) sort-by-priorities)
+  |-
```
Fang-, line 337: I appreciate the restyling here, the code style around these parts is... quite something[.](http://www.employmentking.co.uk/wp-content/uploads/2011/11/disgust.jpg)

Was gonna ask about the manual looping vs `+roll`, but I remember struggling with getting that to behave nicely myself. To confirm, does just `|=  [wel=well:gall _this]` not cut it for the `+roll` gate?

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -329,23 +330,66 @@
...
+    %-  se-emit
+    [%pass wire %arvo %g %conf [our.hid q.wel] our.hid p.wel]
+  $(servers t.servers)
+  ::
+  ++  priorities
+    ^-  (list (set @))
+    :~
+      (sy ~[%dojo])  ::  ensure dojo connects first
+      %-  sy
+      :~  %permission-store
```
Fang-, line 361: Maybe:

```suggestion
      ::  ensure dojo connects first
      ::
      (sy ~[%dojo])
    ::
      ::  set up stores with priority: depended on, but never depending
      ::
      %-  sy
      :~  %permission-store
```

We also sort for dojo below though, I think this is just boot order rather than connect order, which for dojo doesn't really matter?

pkg/arvo/lib/hood/drum.hoon
liam-fitzgerald: yeah `roll` was being funny

liam-fitzgerald:
while I fix these up, do you think it's worth going thru and deleting all of the %u scries ?

Fang-:
Hmm, good one. If they're not too tangled up with other logic, might be nice to clean up the OTA-related existence checks.

liam-fitzgerald:
Perhaps there's a problem between keyboard and chair, but I can't see the comment you left @philipcmonk

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -329,23 +330,65 @@
...
+  =/  hig=(unit (unit server))
+    (~(get by fur) q.well)
+  ?:  &(?=(^ hig) |(?=(~ u.hig) =(p.well syd.u.u.hig)))
+    that
+  =.  fur
+    (~(put by fur) q.well ~)
+  =.  that
+    (se-text "activated app {(trip p.well)}/{(trip q.well)}")
+  %-  se-emit
+  [%pass wire %arvo %g %conf [our.hid q.well] our.hid p.well]
```
philipcmonk, line 349: Hmm, somehow this comment got eaten.

This doesn't preserve the changes in the accumulator because we never read from `that`.  In particular, `se-text` and `se-emit` add moves to our state, but they get ignored the next iteration of `reel`.  You can resolve this by adding `=.  this  that` after the `|=`, or by making sure all functions are called like `se-text:that`, or by unrolling to a manual `|-`.

This bug will present as "works to start one app but not multiple apps", which is the kind of bug that may not be found until it's on livenet and is often hard to pinpoint.  I really hate bugs where I *know* I did the right thing, just the moves/state changes got silently dropped somewhere else in the code.  Personally, I shy away from ever using `this` as the accumulator in `reel`/`roll` for exactly this reason.

liam-fitzgerald: So your preference would be for recursion here?

philipcmonk: Yeah, I would just do a `|-` here.

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -386,8 +391,8 @@
     %+  sort
       ~(tap in eel)
     |=  [[@ a=term] [@ b=term]]
-    ?:  =(a %dojo)  %.y
-    ?:  =(b %dojo)  %.n
+    ?:  =(a %dojo)  %.n
+    ?:  =(b %dojo)  %.y
```
philipcmonk, line 395: Was this backward?

liam-fitzgerald: indeed

liam-fitzgerald:
Pls merge

PR #2814, opened 2020-04-28
chat: improve namespacing of chat-* types

liam-fitzgerald:
Namespacing proposal:
- Never prefix a type with the name of the app. e.g. always `action` never `chat-hook-action`
- All files in /lib and /sur should have ^? on the main core. ^?  makes the payload opaque, which means you can't pull in imports indirectly. (e.g. you can't import lib/chat-hook to get at `action:sur:chat-hook-lib`)
- Each app should have precisely one /sur and one /lib file
- App-associated files in /sur should be imported as
```
/-  hook=chat-hook :: inside the context of chat
/-  chat-hook  :: outside the context of chat
```
- Associated files in /sur should be imported as
```
/+  hook-lib=chat-hook
/+  chat-hook-lib=chat-hook
```
- JSON conversion functions should be under +dejs or +enjs inside the associated /lib file. 
- Imports with * should be avoided as much as is reasonably possible, especially in /app files.

Implements the above proposal for the chat-* apps.

If everyone's happy with this, I'd like to extends this style to the rest of the codebase


matildepark:
(Informal GitHub etiquette: only tag reviewers if they're required; cc optional reviews. This isn't codified yet iirc, but it helps clarify when merging.)

pkg/arvo/mar/chat/view-action.hoon
```diff
@@ -1,11 +1,11 @@
...
+/+  *chat-view
+|_  act=action
 ++  grab
   |%
-  ++  noun  chat-view-action
+  ++  noun  action
   ++  json
     |=  jon=^json
-    (json-to-view-action jon)
+    (action:dejs jon)
```
loganallenc, line 9: You can just put `++  json  action:dejs`

pkg/arvo/mar/chat/hook-action.hoon
```diff
@@ -1,11 +1,12 @@
...
+/+  *chat-hook
+|_  act=action
 ++  grab
   |%
-  ++  noun  chat-hook-action
+  ++  noun  action
   ++  json
     |=  jon=^json
-    (json-to-hook-action jon)
+    (action:dejs jon)
```
loganallenc, line 9: can replace with `++  json  action:dejs`

pkg/arvo/mar/chat/action.hoon
```diff
@@ -1,11 +1,12 @@
...
-|_  act=chat-action
+|_  act=action
 ++  grab
   |%
-  ++  noun  chat-action
+  ++  noun  action
   ++  json
     |=  jon=^json
-    (json-to-action jon)
+    (action:dejs jon)
```
loganallenc, line 10: Same

pkg/arvo/app/chat-hook.hoon
```diff
@@ -3,8 +3,11 @@
 ::  allow sending chat messages to foreign paths based on write perms
 ::
 /-  *permission-store, *chat-hook, *invite-store, *metadata-store,
-    *permission-hook, *group-store, *permission-group-hook  ::TMP  for upgrade
-/+  *chat-json, *chat-eval, default-agent, verb, dbug
+    *permission-hook, *group-store, *permission-group-hook,  ::TMP  for upgrade
+    hook=chat-hook,
```
Fang-, line 7: This is importing `/sur/chat-hook` twice now.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -3,8 +3,11 @@
...
+    store=chat-store,
+    view=chat-view
+/+  *chat-eval, default-agent, verb, dbug, store-lib=chat-store
```
Fang-, line 10: Wondering if we should just go and move the things in chat-eval into the store lib...

pkg/arvo/app/chat-store.hoon
```diff
@@ -1,6 +1,7 @@
 :: chat-store: data store that holds linear sequences of chat messages
 ::
-/+  *chat-json, *chat-eval, default-agent, verb, dbug
+/-  store=chat-store
```
Fang-, line 3: We discussed importing "native" types (and maybe libs) with `*`, and thought we agreed on that? Having `:store` suffixes, inside of the store, feels a little bit silly to me. (Same goes for the other apps.)

pkg/arvo/lib/chat-store.hoon
```diff
@@ -0,0 +1,210 @@
...
+  =,  enjs:format
+  |%
+  ::
+  ++  letter
+      |=  =^letter
+      ^-  json
+      =,  enjs:format
+      ?-  -.letter
+          %text
+      (frond %text s+text.letter)
```
Fang-, line 14:
```suggestion
  ++  letter
    |=  =^letter
    ^-  json
    =,  enjs:format
    ?-  -.letter
        %text
      (frond %text s+text.letter)
```

Looks like this file in its entirety of `+enjs` is using 4 instead of 2 spaces for its indentation. (ie the `?:`s below)

pkg/arvo/lib/chat-view.hoon
```diff
@@ -0,0 +1,51 @@
...
+  |%
+  ++  action
+    |=  jon=json
+    ^-  ^action
+    =,  dejs:format
+    =<  (parse-json jon)
+    |%
+    ++  parse-json
+        %-  of
+        :~  [%create create]
```
Fang-, line 14:
```suggestion
    ++  parse-json
      %-  of
      :~  [%create create]
```

etc.

pkg/arvo/app/chat-store.hoon
liam-fitzgerald: I've been flipping back and forth on this.  I like having `:store` suffixes because 
1. using the autoname syntax doesn't shadow the type.
2. It preserves a semblance of referential transparency inside the chat context. It's not uncommon to have to make changes to both the -store and -hook and I think preserving the namebindings is useful to prevent unnecessary context switches
3. Types aren't really that hard up for space in most common syntactical situations, so identifier length is not really an issue here.

liam-fitzgerald:
> One thing I neglected to mention in conversation, @liam-fitzgerald, is that maybe lib and sur files for the stores themselves shouldn't have the `-store` suffix. Idea here being that the things in them are data structures and associated helpers for the _entire suite_ of apps, not just the store.
> So you'd get, for example, `envelope:chat` and `action:chat` in `/sur/chat.hoon`.
> 
> Wondering what you think of this. It does rub up against "1 file per app" a little bit, in that the relation is less literal.

I kinda dislike this, because it implies (at least to me) that `action:chat` and `action:chat-hook` are somehow dependent, or that `action:chat-hook` is a subtype of `action:chat`. This is true for the chat apps, but is not necessarily the case for all apps. Also wondering how this works in relation to an incomplete 'suite' of apps. e.g. `permission-group-hook`.

ohAitch:
Personally I would push for the ^? being part of the _ford /- * syntax_, so you can still structure things as multiple cores but have to be explicit about things. It's not being able to get at things at all that's an issue, it's them being spilled into a uniform namespace: if you see action:chat-sur:chat-lib due to both being qualified imports, you can follow that something funky is going on.

pkg/arvo/app/chat-store.hoon
ohAitch: "Each file can get one sur `*` or one lib `*`, after that things get too confusing" seems like a reasonable compromise. Otoh `-lib` suffixes appearing anywhere are a sign that something has gone obviously wrong, and imo a fairly strong argument for a "stacking" model of "a library will frequently subsume the structure by the same name, pull in only one of them" - I don't believe e.g. `inbox-to-configs:store` would be overly unclear compared to `inbox-to-configs:store-lib`.

pkg/arvo/mar/chat/hook-update.hoon
```diff
@@ -1,13 +1,14 @@
...
 ++  grow
   |%
-  ++  json  (hook-update-to-json upd)
+  ++  json  (update:enjs upd)
   --
 ::
 ++  grab
   |%
-  ++  noun  chat-hook-update
+  ++  noun  update
```
ohAitch, line 11: Something I don't love here is these are in fact being sent over the wire as `%chat-hook-update`, which might warrant some verbosity in at least the marks and !<s for consistency. (Alternately the mark system should perhaps have some manner of namespacing… even like, `%chat-hook--update` would make intermittent usage of the bare name less awkward)

pkg/arvo/app/chat-store.hoon
Fang-: Having libs and matching surs under the same namespace sounds pretty nice, yeah. Lacking support for simultaneous `/-  name=thing` and `/+  name=thing`, doesn't seem too unreasonable to just expose the types with the lib, under the assumption you're gonna need them anyway.

pkg/arvo/app/chat-hook.hoon
liam-fitzgerald: done!

pkg/arvo/lib/chat-view.hoon
```diff
@@ -0,0 +1,50 @@
+/-  *chat-view, *rw-security
+|%
```
Fang-, line 2: Hmm, but with the `^?` removed, this would still expose the `*rw-security`. Wondering if there's some middle ground here where we can _selectively_ expose only the types (core) "native" to the lib...

liam-fitzgerald: could we just cons the sur import and the library core together?

Fang-: Oh yeah, that seems to work.

You're gonna have to `=, chat-view` instead of importing as `*chat-view` though, which is bad-style use of `=,`. I guess the ends justify the means here.

pkg/arvo/lib/chat-store.hoon
```diff
@@ -0,0 +1,235 @@
...
+  =,  enjs:format
+  |%
+  ::
+  ++  letter
+    |=  =^letter
+    ^-  json
+    =,  enjs:format
+    ?-  -.letter
+        %text
+    (frond %text s+text.letter)
```
Fang-, line 16:
```suggestion
        %text
      (frond %text s+text.letter)
```

The backsteps on these are still a bit too aggressive.

loganallenc:
Liam the indenting you “fixed” is wrong...

liam-fitzgerald:
Pls approve @loganallenc @Fang-

loganallenc:
Morge

matildepark:
You have a merge conflict here @liam-fitzgerald

pkg/arvo/lib/chat-store.hoon
```diff
@@ -0,0 +1,235 @@
...
+=,  sur
+|%
+++  enjs
+  =,  enjs:format
+  |%
+  ::
+  ++  letter
+    |=  =^letter
+    ^-  json
+    =,  enjs:format
```
loganallenc, line 13: unnecessary

pkg/arvo/lib/chat-store.hoon
```diff
@@ -0,0 +1,235 @@
...
+        [%number (numb number.envelope)]
+        [%author (ship author.envelope)]
+        [%when (time when.envelope)]
+        [%letter (letter letter.envelope)]
+    ==
+  ::
+  ++  config
+    |=  =^config
+    ^-  json
+    =,  enjs:format
```
loganallenc, line 47: unnecessary

pkg/arvo/lib/chat-store.hoon
liam-fitzgerald: What do you mean?

liam-fitzgerald: oh right the tiscom

matildepark:
This has a merge conflict again.

When should this even be merged? And what target? It seems like these can go out whenever; they're more for developer benefit than user fixing.

Fang-:
> When should this even be merged? And what target?

afaic this can be merged when conflict is resolved, onto a `release/next-userspace` or `release/next-ota`.

matildepark:
ping @liam-fitzgerald Mind fixing the conflicts and targeting `release/next-userspace` for this and the other namespacing PRs?

liam-fitzgerald:
Don’t merge yet I have to port the virtualisation changes

matildepark:
RIP sorry

You can make another PR for `release/next-userspace`?

PR #2815, opened 2020-04-28
chat-js: only show scrollbars if overflow

liam-fitzgerald:
Uses overflow: auto on the code attachments so that scrollbars only
appear if the content overflows the container. Adds borders on sent Hoon
code to improve visual separation. 
Demo:
Screen Shot 2020-04-28 at 12 53 44 pm

Screen Shot 2020-04-28 at 12 56 12 pm

cc: @urcades

### DIFF
--- a/pkg/interface/chat/src/css/custom.css
+++ b/pkg/interface/chat/src/css/custom.css
@@ -71,7 +71,7 @@ h2 {
 }
 
 .clamp-attachment {
-  overflow: scroll;
+  overflow: auto;
   max-height: 10em;
   max-width: 100%;
 }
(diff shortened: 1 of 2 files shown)


urcades:
@liam-fitzgerald this is nice! Much more readable than the current output we render.

One question: What determines the height of the containers that are now scrollable?

matildepark:
@urcades looks like this:

https://github.com/urbit/urbit/pull/2815/files#diff-3168c42d5901c606b146add7e4181ca2R75

loganallenc:
@matildepark plz morge

PR #2819, opened 2020-04-28
[WIP]: Regicide

benjamin-tlon:
- [x] Don't build the Daemon from `serf` build. (Can be re-added later as separate package)
- [ ] Remove daemon code from serf source code, and remove it's dependencies.
- [ ] Use King Haskell for tests/CI.
- [ ] Use King Haskell executable in release bundles.
- [ ] Write a wrapper shim that emulates old command line arguments for now (until docs can be updated).


joemfb:
Unsurprisingly, I object. If we're gonna do this, it should be `pkg/u3` with noun and the jets, and then separate `u3-serf` and `u3-king` packages with appropriate dependences. But I've got a branch in progress with tons of changes, and that kind of movement will need be either afteer or on top of those...

benjamin-tlon:
Most important part of this PR is to get the tests and releases working with KH, can figure out how to untangle the `vere` and `serf` packages later.

joemfb:
Obsolete/incomplete, closing.

PR #2823, opened 2020-04-29
chat-js: fix race condition in new-dm flow

### LINKED ISSUE #2792: Made a new chat DM, sent a message while it still said “chat is being restored”
And for some reason it then navigated out of the chatroom? And the chatroom then didn’t load? Weird.

liam-fitzgerald:
Creating a new DM might redirect the user to the chat before it is
actually created. Fixed by waiting for confirmation before redirecting.

Fixes #2792


liam-fitzgerald:
Ping @loganallenc

Issue #2817, opened 2020-04-28
Chat module's "~"-mention list is rendered incorrectly


urcades:
I believe this is a regression, but from which point/version, I'm unsure.

See the image below:

<img width="593" alt="Screen Shot 2020-04-28 at 11 33 05 AM" src="https://user-images.githubusercontent.com/1195363/80506876-292fc780-8944-11ea-9567-3378a1cde65d.png">

The autocomplete interface is experiencing three distinct bugs that may be related:
1. User-set "avatar" sigil bounding boxes are getting stretched horizontally
2. In the case a user has set their own name, their actual patp is being pushed way, way too far to the right
3. These items are being set to 14px, when they'd look a bit more proportional set to our 12px font size (f9, I believe)

A visual recall of how these user lists are generally intended to be rendered:

<img width="220" alt="Screen Shot 2020-04-28 at 11 39 28 AM" src="https://user-images.githubusercontent.com/1195363/80507585-fb974e00-8944-11ea-824d-f96a584655b7.png">

matildepark:
Seems like we should assign usernames a max-width and `truncate` them (and add `flex-shrink-0` to the `@p` just to be safe, heh). But I won't micromanage the task at hand.

Issue #2804, opened 2020-04-24
Coinciding date and unread header reflows poorly on mobile


liam-fitzgerald:
See title

![image](https://user-images.githubusercontent.com/46801558/80252973-bb716a80-86bb-11ea-8328-56abe084e408.png)

liam-fitzgerald:
Cc: @urcades what are your thoughts?

urcades:
hi @liam-fitzgerald  — this is a great catch!

Two thoughts here: 

1. If we can eliminate the left "green line" and simply align the "New messages below" text with the leftmost edge of the content (the leftmost edge of the avatar) and adjust the line + "Today" label accordingly, that would be great:

<img width="352" alt="Screen Shot 2020-04-26 at 7 25 27 PM" src="https://user-images.githubusercontent.com/1195363/80322610-6127f500-87f4-11ea-8e55-8d7334953d11.png">

2. If the above proves to be too annoying to to get looking good, we can change the styling such that the "day" marker and "new message" marker are stacked: 

<img width="341" alt="Screen Shot 2020-04-26 at 7 26 37 PM" src="https://user-images.githubusercontent.com/1195363/80322605-5bcaaa80-87f4-11ea-8e90-c14064680874.png">

Issue #2797, opened 2020-04-24
Chat message unread markers appears too often / disappears too quickly


Fang-:
**Describe the bug**
When navigating to a channel with unread messages, the green line that marks the beginning of unread messages disappears very quickly after navigating to the channel. It's immediately respecting the "last read msg nr" sent by the backend.

**Expected behavior**
The marker to hold its place until I navigate away from the channel again. Also, to not briefly reappear for every new message that comes in.

This is on latest, `0vj.57gf5.1qk6u.r75oq.dlvm4.mkj21.ufv0v.7cor6.pu1b9.cmib5.ontk0`.

matildepark:
From @urcades:

> "New Messages" visual indicator requires some adjustments:
> 1. The indicator is a bit over-enthusiastic and pops in even when I'm on the page – it should only really show up after I've been perceived as "being away" from the screen/chat for a while (this is vague and needs clarification, I need to study the actual behavior when the chats are busy, it's too quiet/slow on the weekends)
> 2. I still start at the bottom of the scroll, presumably where the last* new message is located — ideally, we'd place a user alongside the first* new message to have been posted, and clear the notifier when a user either dismisses the message manually, or scrolls to the bottom of the scroll, and has effectively "read" the messages in sequential order

matildepark:
I'll also add that the "disappear when scrolling through the midpoint" makes chats with < 20 messages or chats that have just gotten a message not dismiss the notifier automatically.

Issue #2782, opened 2020-04-23
Chat doesn't scroll to bottom on new messages


eglaysher:
**Describe the bug**
Under conditions I don't understand, chat will stop scrolling to the bottom of the screen. If you're at the end of the newest messages, and a new message arrives, you won't scroll to the bottom.

**To Reproduce**
1. Go to a chat channel
2. ????
3. Receive new messages
4. See that the scrollbar goes up but that you don't scroll to the bottom.

**Expected behavior**

**Screenshots**
If applicable, add screenshots to help explain your problem. If possible, please also screenshot your browser's dev console. Here are [Chrome's docs](https://developers.google.com/web/tools/chrome-devtools/open) for using this feature.

**Desktop (please complete the following information):**
 - OS: Mac 10.14.5, but I can repro on Windows, too.
 - Browser: Chrome Version 80.0.3987.163
 - 0v1e.va6cc.cks0i.8dhed.sre6p.pi67c.4b9he.d9c5s.ii3v2.c2v76.i2ffu

**Additional context**
This isn't a new problem and I've run into this for a few months, but assumed that you knew about it but then I mentioned it to ~palfun and he said it was unknown.

Fang-:
Also experiencing this now on Firefox, post-OTA. Probably unrelated cause though? Seems like unread markers, or typing very long chat messages, can get me unstuck from the bottom without actually scrolling myself, causing the frame to not scroll down on subsequent messages.

PR #2825, opened 2020-04-29
chat-js: fix scrolling and unread behaviour

liam-fitzgerald:
- Fixes autoscroll to the bottom on new messages
- Changes behaviour of unread markers: The chat loads in centered at the unread marker, and sets the chat as read when the user scrolls to the bottom. This causes the 'x new messages since y' overlay to disappear, however the green line break does not disappear until the user navigates away and back to the chat. Additionally, if all the messages in a chat have been read and a new message comes in whilst the chat is open, the new message is automatically marked as read
- Hides line on left side of 'new messages below' marker on small screens
- Passes prop.size into the flexBasis property of the sigil, so that it flexes correctly

cc: @urcades 

fixes: #2778
fixes: #2782
fixes: #2797
fixes: #2804
fixes: #2817


matildepark:
Does this address complaints 2 and 3 of #2817?

matildepark:
Also how does this address #2778? It's pretty widespread on livenet for repros ofc. Does this fix the "pin chat to bottom" behaviour? Just confused as to what's going on for our scroll lock behaviour here.

loganallenc:
What happens when the unread marker is more than 220 messages old? (In other words, further back than a single fetch of backlog messages would give us.

pkg/interface/chat/src/js/components/chat.js
```diff
@@ -89,21 +96,21 @@ export class ChatScreen extends Component {
...
        props.envelopes.length !== prevProps.envelopes.length ||
        getNumPending(props) !== this.lastNumPending ||
        state.numPages !== prevState.numPages)
     ) {
-      if(state.scrollBottom) {
-        setTimeout(() => {
-          this.scrollToBottom();
-        })
-      } else {
+      this.scrollToBottom();
```
liam-fitzgerald, line 112: here we try to scrollToBottom upon loading any new messages. Safari not scrolling to bottom on load is a result of a race condition where we try to scroll to the bottom before any messages load in. 
@matildepark

liam-fitzgerald:
> What happens when the unread marker is more than 220 messages old? (In other words, further back than a single fetch of backlog messages would give us.

Good catch. On one hand I'd like to fetch all unread messages, but I wonder if this is feasible. What's your take?

liam-fitzgerald:
> Does this address complaints 2 and 3 of #2817?

yes

urcades:
@matildepark + @liam-fitzgerald 

I sent this to Liam this morning, but figured it'd be well to just post it so everyone is on the same page with my thinking:

Screen Shot 2020-04-29 at 11 36 27 AM

The tweet I reference in particular is [here](https://twitter.com/rsnous/status/1251810672544845826?s=20), and it rings deeply true to me — in many platforms, twitter being the most egregious one, scrolling is often a "real labor" labor performed with no respect being given to it by the platform as such. 

Ideally, in -all- cases, no matter what, I'd heavily prefer we do whatever we can to respect our user's scrolling labor and never take it away from them — the only times we should automatically progress the scroll is when they are already situated near the bottom of the chat, with a small threshold of what constitutes "near the bottom".

The example I provided Liam is that perhaps the height of a single line of text, or an avatar, or something in that realm (the max of this threshold perhaps being half the height of the viewport) is enough to keep a users "pinned to the bottom". We should assume that any distance further than this constitutes a deliberate act to stay situated in the scroll wherever they are, and respect that positioning.

The only other case in which we'd auto-scroll a user is if they input text into the messaging-send-field, and send it, or post an image, or code, or whatever else we allow a user to send in the future — posting to the mainline view of a chat should assume that one is responding to the latest messages, which are always located at the bottom of a scroll.

Let me know if these thoughts seem sane/amenable to how Chat is currently built — I have no idea how to address the case Logan mentioned, regarding fetching "further than the backlog history". My naive answer to this would involve loading in whatever history is necessary to situate a user where they last left off, but I'm unsure if that's possible or ergonomic or involves longer load/wait times.

loganallenc:
> > What happens when the unread marker is more than 220 messages old? (In other words, further back than a single fetch of backlog messages would give us.
> 
> 
> 
> Good catch. On one hand I'd like to fetch all unread messages, but I wonder if this is feasible. What's your take?

I doubt it would be slow to do so. Go for it!

loganallenc:
I’m not a fan of doing this in the askForMessages method. I’d rather a method that wraps that method or calls that method.

loganallenc:
King, looks good. I’ll test it tomorrow.

liam-fitzgerald:
Ping @loganallenc

PR #2826, opened 2020-04-29
king: add --ames-latency option

liam-fitzgerald:
Adds --ames-latency option, which specifies a time in milliseconds that
the king should wait before sending an ames event to the serf. 

Motivation:
In practice, one often wants to mimic unreliable network conditions 
between fake ships on their own machine. This allows developers to 
do that. 

(Please be hard on style, I'm a haskell neophyte)

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
@@ -31,6 +31,7 @@ data Opts = Opts
     , oHttpPort     :: Maybe Word16
     , oHttpsPort    :: Maybe Word16
     , oLoopbackPort :: Maybe Word16
+    , oAmesLatency  :: Maybe Int
     }
   deriving (Show)
 
@@ -220,6 +221,13 @@ opts = do
       <> long "ames"
       <> help "Ames port"
       <> hidden
+
+    oAmesLatency <-
+      optional
+      $  option (map (*1000) auto)
+      $  metavar "DELAY"
+      <> long "ames-latency"
+      <> help "Milliseconds before sending ames event"
 
     oHttpPort <-
       optional
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
@@ -174,6 +174,10 @@ ames inst who isFake enqueueEv stderr =
                 bindSock socketVar
               Right (bs, addr) -> do
                 logTrace $ displayShow ("(ames) Received packet from ", addr)
+                mLatency <- view (networkConfigL . ncAmesLatency)
+                case mLatency of
+                  Nothing -> pure ()
+                  Just latency -> threadDelay $ latency
                 case addr of
                     SockAddrInet p a -> atomically (enqueueEv $ hearEv p a bs)
                     _                -> pure ()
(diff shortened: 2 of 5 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
```diff
@@ -174,6 +174,10 @@ ames inst who isFake enqueueEv stderr =
                 bindSock socketVar
               Right (bs, addr) -> do
                 logTrace $ displayShow ("(ames) Received packet from ", addr)
+                mLatency <- view (networkConfigL . ncAmesLatency)
+                case mLatency of
+                  Nothing -> pure ()
+                  Just latency -> threadDelay $ latency
```
eglaysher, line 180: So, because we're reading a from a socket, this threadDelay will also block further reading of packets from the socket, not just enqueuing them for ames. As long as this is what you intend and this is a debugging option that'll never get turned on "for real" in the wild, it's probably ok, but I want to check that you mean this.

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -221,6 +222,13 @@ opts = do
       <> help "Ames port"
       <> hidden
 
+    oAmesLatency <-
+      optional
+      $  option (map (*1000) auto)
+      $  metavar "DELAY"
+      <> long "ames-latency"
+      <> help "Milliseconds before sending ames event"
+
```
eglaysher, line 231: You probably want `<> hidden` here so that this option doesn't show up on the brief line.

benjamin-tlon:
A King Haskell PR! Exciting!

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
benjamin-tlon: I wouldn't do the `(* 1000)` here. Command-Line arguments data structure should closely map to actual user input. Do the conversion later, unless it's too annoying. Right now the code isn't setup for this (there is no function to convert from CLI parameters to runtime options. This already hurts in a few places now, so it's ripe for doing).

This is probably fine, just commenting here b/c there's a better way to do this.

pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
benjamin-tlon: I agree with @eglaysher here: I would do something like:

Replace this:

`atomically (enqueueEv $ hearEv p a bs)`

With this:

```
async $ do
    threadDelay latency
    atomically (enqueueEv $ hearEv p a bs)
```

On the other hand, this is a kinda dirty hack to test latency, so it probably doesn't matter that much either way.

benjamin-tlon: This isn't really important, but a minor style comment since you asked for it:

Instead of this:

```
mLatency <- view (networkConfigL . ncAmesLatency)
case mLatency of
  Nothing -> pure ()
  Just latency -> threadDelay $ latency
```

I would write this:

```
view (networkConfigL . ncAmesLatency) >>= \case
  Nothing      -> pure ()
  Just latency -> threadDelay latency
```

Or even

```
view (networkConfigL . ncAmesLatency) >>= maybe (pure ()) threadDelay
```

benjamin-tlon:
A comment on this PR in general:

It is possible to do this kinda thing at the OS-level. Maybe it would be better to do it there instead of adding more code for adding latency to Ames specifically?

liam-fitzgerald:
> A comment on this PR in general:
> 
> It is possible to do this kinda thing at the OS-level. Maybe it would be better to do it there instead of adding more code for adding latency to Ames specifically?

This is probably the better idea. Closing.

PR #2840, opened 2020-05-03
gall: smaller %watch-not-unique print

Fang-:
Instead of printing all outgoing subscriptions for the app, only print the
subscription whose wire we're trying to re-use.

In the case of wire re-use, I usually don't care about all those other subscriptions...

(Given this is a kernel PR, and our current OTA strategy around those, this might need to be retargeted at something other than master...)


Fang-:
> let's not merge until we have a place for /sys things which are not urgent.

Alright, merging into `release/next-sys` now.

PR #2841, opened 2020-05-04
groups: fix self destruct on kick

### LINKED ISSUE #2816: Urbit group deleted spontaneously
Hello I created ship "hiddev-dannut" to host the new Hooniverse group.  A member filsur-sanmut could not see it, even though at some point he had to have had access because he was listed as a colored sigil with access.  I clicked on the sigil and saw the expanded sigil page with his full ship name and the "Remove from group" link in the upper right. upon clicking that link, the entire group disappeared.  It was confirmed to have disappeared from my other ship which was a member.

I recreated the Hooniverse group on that ship and invited a couple of former members. Those members immediately had access to all chat and group links (without re-creating them) and could post and be seen in those chat channels. 

So in summary, a group got deleted from removing a member, and then once the group was re-created the chats and links were all there automatically.  Only the members were gone. I am hoping that re-inviting the members does not lead to more of this disappearing group error.

Here is the error that filsur-sanmut got:

https://pastebin.com/nJ1b10Nr

In case that is helpful.

liam-fitzgerald:
Fixes a bug where kicking a member with contact info set would delete the group. Also fixes a bug where a group member leaving would not be propagated to the rest of the group

fixes #2816

### DIFF
--- a/pkg/arvo/app/group-hook.hoon
+++ b/pkg/arvo/app/group-hook.hoon
@@ -215,7 +215,9 @@
     =/  ship  (~(get by synced.state) pax.diff)
     ?~  ship  [~ state]
     ?.  =(src.bol u.ship)  [~ state]
-    ?.  (~(has in members.diff) our.bol)  [~ state]
+    ?.  (~(has in members.diff) our.bol)
+      :_  state
+      [(group-poke pax.diff diff)]~
     =/  changes  (poke-group-hook-action [%remove pax.diff])
     :_  +.changes
     %+  welp  -.changes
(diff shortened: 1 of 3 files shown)


matildepark:
woah dope

baudtack:
Please merge this as soon as it's tested to resolve the problem. It needs deployed as soon as possible as hoon school just imploded because of this.

Fang-:
Is there anything at all we can do to OTA-fix the situation for people who hit this, or is the group completely gone? Could we look at permissions, or contacts? cc @loganallenc @liam-fitzgerald

matildepark:
We’ll be pushing something tonight, @baudtack. 

@Fang- do you want to jump into the ship and +dbug? Was it dopzod?

loganallenc:
This should be merged immediately.

Fang-:
> @Fang- do you want to jump into the ship and +dbug? Was it dopzod?

@baudtack's case lives on his own ship. I don't think I have direct access to anything where this occurred. @loganallenc and probably @liam-fitzgerald are more familiar with the logic here anyway, and might be able to y/n answer the "is anything even possible" question.

Issue #2843, opened 2020-05-04
Add SVG support to Chat.


sigilante:
Is it anything more than adding svg|SVG here?

https://github.com/urbit/urbit/blob/03d80b8c6301d2bbbe1a89563af5908856faf443/pkg/interface/chat/src/js/components/lib/message.js#L82

matildepark:
I think that would be it, yeah. `<img>`ing svg files works fine iirc.

matildepark:
Even that regex could remove half those terms by making it case insensitive, haha. I think that's just adding `i` after the backslash at the end.

matildepark:
Merged in #2846.

PR #2846, opened 2020-05-04
Add SVG support to img tags in Chat.

sigilante:
#2843 

I'll let you wizards add the case-insensitivity i at the end, JS isn't my scene.


sigilante:
While we're on the subject, jpg? should be fine in the middle but that starts to get weird.  Wish you could do a Unix-style scan of the file header without having to access the whole thing.

Fang-:
> While we're on the subject, jpg? should be fine in the middle

I'd say you could just `/(jpg|png|etc)($|\?.*)/` to match either end, of `?` and whatever more might follow. Maybe that's a bit too permissive?

PR #2849, opened 2020-05-05
contacts-js: fix setting of avatar prop

liam-fitzgerald:
Passes an empty avatar prop through correctly.


loganallenc:
What is your intention here, like what are you trying to fix? I’m not sure this will work the way you expect.

liam-fitzgerald:
If an avatar is not set, then the avatar is sent as `{ url: null }` instead of just `null`. This causes a parsing error when JSON conversion is attempted

loganallenc:
Gotcha, okay!

ixv:
@liam-fitzgerald merge conflict

matildepark:
@liam-fitzgerald Target feat/spa?

loganallenc:
Ping, please clean up these pull requests

liam-fitzgerald:
bug has already been fixed

PR #2850, opened 2020-05-05
publish: allow crash case to no-op instead

ixv:
When we remove metadata for a notebook, if the metadata does not exists, we previously crashed. Now we just no-op. This isn't directly related to any issues seen on the network, but prevented certain recovery tactics I tried to use when fixing ~bitpyx.

In general the publish %warp loop is fragile to these crashes, if you receive an update from clay and crash before you can set the next read, publish will not recover from that. This should be fixed more rigorously at some point. I'm unsure if the answer is to remove all such possible crashes, or to stop using clay entirely, which is why I have only fixed this one case for now.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -1857,8 +1857,9 @@
   ::
       %remove
     =/  app-path  [(scot %p author.del) /[book.del]]
-    =/  group-path  (group-from-book app-path)
-    [(metadata-poke [%remove group-path [%publish app-path]])]~
+    =/  group-path=(unit path)  (group-from-book app-path)
+    ?~  group-path  ~
+    [(metadata-poke [%remove u.group-path [%publish app-path]])]~
   ==
   ::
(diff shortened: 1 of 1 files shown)


ixv:
Thanks, handled that crash case and the other one in this arm as well. Also added a debug poke to manually reset the warp loop as a temporary support measure for now.

loganallenc:
Question: (not trying to hold this up) why not also run the %warp reset on-load?

ixv:
This is an issue that may afflict people multiple times, so it's neccesary to put it in the on-poke. But it does not afflict everyone on the network, so putting it in the on-load additionally is unnecessary.

Putting it on the on-load would solve the problem automatically for some small amount of people who happen to have run into the problem soon enough before an app update that they need not bother to do the poke, but doing it on every app update may get a bit slow for everyone else, the more notebook data they have, since it causes publish to reprocess everything they have.

So it's a minor benefit, and causes a minor annoyance, meh.

PR #2852, opened 2020-05-05
ames: Deduplicate long messages

philipcmonk:
This is essentially what was pushed to ~marzod and ~marnus over the weekend.

This reduces the memory usage of Ames in some situations by delaying splitting messages into fragments until the last moment.  It also checks whether new messages longer than 1KB exist in queues to other ships or flows.  This deduplicates large messages, for example OTA updates that may need to be sent to many ships.


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3377,6 +3415,17 @@
   |=  [=symmetric-key plaintext=shut-packet]
   ^-  @
   ::
+  =.  meat.plaintext
```
belisarius222, line 3418: could be a =?

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2222,7 +2255,8 @@
       =/  =message-pump-state
         (~(gut by snd.peer-state) bone *message-pump-state)
       ::
-      =/  message-pump    (make-message-pump message-pump-state channel)
+      =/  message-pump
```
belisarius222, line 2258: It doesn't look like we need to thread peers.ames-state into this core anymore.

philipcmonk: GOOD point

philipcmonk:
@belisarius222 Can I get a re-review?

philipcmonk:
Whoops, accidentally merged it anyway.  Uh, let me know if something's wrong

PR #2853, opened 2020-05-05
+scot and +slaw jets

eglaysher:
This patch builds partial jets for common cases in +slaw and +scot, as well as revamping the +po jets.


eglaysher:
In the git history, there's a jet for `+ob`, but I gave up trying to get all the refcounts correct and just used a hook instead. The `+fein` and `+fynd` gates are lifted to the toplevel core in `%tri` because there appears to be a bug in the hook system where it can't deal with gates which aren't directly in the core being jetted; `fein:ob` doesn't work.

eglaysher:
I believe I've fixed most of what's immediately actionable: I made u3a_string() and changed how they were freed, I fixed the one instance of fail that I found, and I moved most failures to u3_none.

The printing code in hoon is tape first, and I went to copy that structure.

pkg/arvo/sys/hoon.hoon
```diff
@@ -2133,7 +2133,13 @@
...
   ::    3g: molds and mold builders                     ::
   ::                                                    ::
-~%  %tri  +  ~
+~%  %tri  +
+  ==
+    %year  year
+    %yore  yore
+    %fein  fein
+    %fynd  fynd
+  ==
```
joemfb, line 2142: I don't think it's right to rearrange these cores for incrementally easier hooking. If you just declare `%ob` here, the hint `+ob` and declares hooks on it for `%fein` and `%fynd`, you can hook them in the jets in two steps: hook the core for `+ob`, then hook the product of that for `+fein` or `+fynd`.

pkg/urbit/jets/a/mod.c
```diff
@@ -10,15 +10,10 @@
...
-    if ( b == 3 && a == 2684227708 ) {
-      printf("dword at 0x27ff84ff8 is %" PRIu64 "\r\n", *(c3_d *)0x27ff84ff8);
-      *(c3_d *)0x27ff84ff8 = 25;
-      printf("see, we modified it\r\n");
-    }
-#endif
     if ( 0 == b ) {
       return u3m_bail(c3__exit);
+    } else if ( _(u3a_is_cat(a)) && _(u3a_is_cat(b)) ) {
+      return a % b;
```
joemfb, line 16: Nice catch! I bet there's quite a bit of this sort of low-hanging fruit lying around. I'd like to really audit these jets at some point ...

pkg/urbit/jets/e/scow.c
```diff
@@ -0,0 +1,385 @@
...
+    return u3nq('~', a, b, u3nc(c, 0));
+  }
+
+  u3_atom dyy = u3qc_met(4, u3k(sxz));
+  if (!_(u3a_is_cat(dyy))) {
+    return u3_none;
+  }
+
+  u3_noun list = 0;
+  for (u3_atom imp = 0; imp != dyy; ++imp) {
```
joemfb, line 206: I'm concerned about the use of `u3_atom` here. It doesn't look like `imp` is used for anything but separator counting, so it can be of any type. I see now that you `return u3_none` early on indirect atom, so this is basically just a style concern, but using `u3_atom` instead of `c3_w` implies that the value might be an indirect atom.

pkg/urbit/jets/e/scow.c
```diff
@@ -0,0 +1,385 @@
...
+  do {
+    if (between == 3) {
+      list = u3nc('.', list);
+      between = 0;
+    }
+
+    list = u3nc(u3ka_add(u3qa_mod(ud, 10), '0'), list);
+    between++;
+    ud = u3ka_div(ud, 10);
+  } while (c3n == u3r_sing(ud, 0));
```
joemfb, line 251: You can count on any atoms 31-bits or smaller being indirect, and just compare to 0 directly.

pkg/urbit/jets/e/scow.c
```diff
@@ -0,0 +1,385 @@
...
+
+u3_noun
+u3we_scow(u3_noun cor)
+{
+  u3_noun mod;
+  u3_noun atom;
+
+  if (c3n == u3r_mean(cor, u3x_sam_2, &mod,
+                      u3x_sam_3, &atom, 0) ||
+      !_(u3a_is_cat(mod))) {
```
joemfb, line 314: This direct atom check is what I was referring to previously; it doesn't seem like a failure condition to me. IMO, this code should `return u3_none` if the aura is indirect.

pkg/urbit/jets/e/slaw.c
```diff
@@ -0,0 +1,508 @@
...
+  TRY_GET_SYLLABLE(o);
+  ENSURE_NOT_END();
+  TRY_GET_SYLLABLE(p);
+
+  if (*cur != 0) {
+    // We've parsed all of a comet shape, and there's still more in the
+    // string. Error.
+    u3a_free(c);
+    return u3_none;
+  }
```
joemfb, line 262: I missed this yield case earlier, this looks right and should like us parse arbitrary-length `@p`. My bad!

pkg/urbit/jets/e/slaw.c
```diff
@@ -0,0 +1,508 @@
...
+
+u3_noun
+u3we_slaw(u3_noun cor)
+{
+  u3_noun mod;
+  u3_noun txt;
+
+  if (c3n == u3r_mean(cor, u3x_sam_2, &mod,
+                      u3x_sam_3, &txt, 0) ||
+      !_(u3a_is_cat(mod))) {
```
joemfb, line 486: Same aura-directness issue as above.

pkg/urbit/jets/e/scow.c
joemfb: Also, failing the sample check here is a deterministic error.

pkg/urbit/jets/e/scow.c
eglaysher: Changed to c3_w.

pkg/urbit/jets/e/scow.c
eglaysher: Changed here and in the other %u printing functions.

pkg/urbit/jets/e/slaw.c
eglaysher: I actually changed this in response to your previous request. I've changed the comment from "Error" to "Bail to the interpreter" instead.

pkg/urbit/jets/e/scow.c
eglaysher: Given that I've misunderstood what you've requested a few times now, is this what you want?

eglaysher:
I think I've fixed most of what you've wanted; please check the aura thing. PTAL.

pkg/urbit/jets/e/slaw.c
```diff
@@ -0,0 +1,514 @@
...
+  u3_noun txt;
+
+  if (c3n == u3r_mean(cor, u3x_sam_2, &mod,
+                      u3x_sam_3, &txt, 0)) {
+    return u3m_bail(c3__exit);
+  }
+
+  if (!_(u3a_is_cat(mod))) {
+    return u3_none;
+  }
```
joemfb, line 494: Redundant given the default case below.

pkg/urbit/jets/e/scow.c
```diff
@@ -0,0 +1,390 @@
...
+
+  do {
+    if (between == 3) {
+      list = u3nc('.', list);
+      between = 0;
+    }
+
+    list = u3nc(u3ka_add(u3qa_mod(ud, 10), '0'), list);
+    between++;
+    ud = u3ka_div(ud, 10);
```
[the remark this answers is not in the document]
eglaysher: I don't understand this part. We're progressively dividing an input number repeatedly. There is no guarantee that the output of div is a direct atom. While it will eventually become a direct atom (and we can do a != 0 check for that), the intermediaries need to be consumed. 

This was both why I added a u3k(ud) at the top, so that that extra refcount could then be consumed here.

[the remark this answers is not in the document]
joemfb: You're right, I was scanning too fast. Would you leave a comment to that effect on the `u3k()`?

pkg/urbit/jets/e/scow.c
```diff
@@ -0,0 +1,390 @@
...
+  return u3nc('~', list);
+}
+
+static
+u3_noun
+_print_ud(u3_atom ud)
+{
+  // number of characters printed "between" periods.
+  int between = 0;
+  u3_atom list = 0;
```
joemfb, line 239: `list` is a cons list, so it should be `u3_noun`

pkg/urbit/jets/e/slaw.c
```diff
@@ -0,0 +1,514 @@
...
+  while (1) {
+    // Parse 4 hex digits
+    int one, two, three, four;
+    PARSE_HEX_DIGIT(one);
+    PARSE_HEX_DIGIT(two);
+    PARSE_HEX_DIGIT(three);
+    PARSE_HEX_DIGIT(four);
+
+    int current = (one << 12) + (two << 8) + (three << 4) + four;
+    list = u3nc(current, list);
```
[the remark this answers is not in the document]
eglaysher: This is a good catch.

eglaysher:
OK, I've gone through everything, but I have one question regarding the comments I didn't address: I'm unsure about whether I should really be removing the u3k() in ud, uv, uw, etc.

ohAitch:
The last change request I see is "comment purpose", rather than "remove".
(If you wanted to actually get rid of it, I suppose you could directly use
gmp_printf on indirect atoms, which may churn through less space - u3ka_div
may already be optimized for unique references however. For *uw* at least
I'm pretty sure O(1) auxiliary space is straightforward.)

pkg/urbit/jets/e/slaw.c
```diff
@@ -0,0 +1,514 @@
...
+    }
+
+    if (cur[0] > '9' || cur[0] < '0') {
+      u3a_free(c);
+      u3z(total);
+      return u3_none;
+    }
+
+    total = u3ka_mul(total, 10);
+    total = u3ka_add(total, cur[0] - '0');
```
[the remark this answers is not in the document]
joemfb: This declaration is still wrong

pkg/urbit/jets/tree.c
```diff
@@ -883,6 +888,10 @@ static u3j_core _141_qua_d[] =
 
   { "mink", 7, _141_qua_mink_a, 0, _141_qua_mink_ha },
   { "mule", 7, _141_qua_mule_a, 0, _141_qua_mule_ha },
+
+  { "scot", 7, _141_qua_scot_a, 0, 0 },
+  { "scow", 7, _141_qua_scow_a, 0, 0 },
+  { "slaw", 7, _141_qua_slaw_a, 0, 0 },
```
joemfb, line 894: In keeping with the other new-jet PRs, please stub out the arm hash declarations in this file.

PR #2854, opened 2020-05-05
claz: correct lockup registration arg order

Fang-:
We were confusing "rate" and "rateUnit". These are "stars per tick" and
"duration of a tick" respectively, not the other way around.

The naming on those, in the contracts, is... not optimal.

(Looking at the logic for `rateUnit` calculation for `+register-conditonal`, maybe that isn't quite right either? Doubt we'll ever use it, but...)

### DIFF
--- a/pkg/arvo/app/claz.hoon
+++ b/pkg/arvo/app/claz.hoon
@@ -417,14 +417,17 @@
   :*  to
       (mul windup-years yer:yo)
       stars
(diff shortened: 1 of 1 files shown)


matildepark:
```
%start-test-cores
~
+ herb ./ship -p test -d ':-  %cores  /'
gall: not running %test yet, got %poke
```

Looks like CI is failing on #2858, fixed by #2859.

Issue #2858, opened 2020-05-05
|start doesn't start the app


matildepark:
The only thing that has touched hood has been #2805, so it probably comes from that.

<img width="330" alt="image" src="https://user-images.githubusercontent.com/20846414/81125132-56ace000-8f05-11ea-8a3b-f4cc4fee62a7.png">

PR #2859, opened 2020-05-05
drum: fix broken app start

liam-fitzgerald:
If we find an agent that has already been launched then we continue
iterating through the list of apps instead of stopping

Fixes: #2858


belisarius222:
Do we know when this broke?

—
~rovnys-ricfer
https://urbit.org

ixv:
A couple days ago, it hasn't made it onto the network.

liam-fitzgerald:
#2805 @belisarius222

matildepark:
I only merged it last night.

belisarius222:
Ok, thanks

—
~rovnys-ricfer
https://urbit.org

PR #2866, opened 2020-05-06
docs: include line to save urbitrc

tylershuster:
While following the interface documentation, I followed it literally and failed due to the step "Save this file as urbitrc" missing. This changes that. Ignore this if you don't want a PR for it. Just thought it should be mentioned.


matildepark:
The current version of interface's contribution guidelines [does say this](https://github.com/urbit/urbit/blob/master/pkg/interface/CONTRIBUTING.md) now. Sorry for leaving you hanging.

PR #2867, opened 2020-05-06
Per-ship Drum subscriptions

lukechampine:
Huge thanks to @philipcmonk for explaining in detail how to accomplish this.

Previously, `|link` couldn't be used across ships, because Drum always subscribes on `/sole/drum`, regardless of what ship it's talking to. This PR changes Drum to subscribe on `/sole/drum_SHIPNAME`, so that updates can be routed to the appropriate ship.

This means that you can run `|link ~zod %dojo`, hit Ctrl-X, and be dropped into zod's Dojo prompt! This could be used for e.g. administrating one ship from another; to allow a developer to debug your ship in-situ, etc.

Of course, for security reasons, apps generally reject pokes from other ships. So I also added access controls to Dojo: in addition to checking the peer against `team:title`, it will look for the ship in an access control list. So it's basically `ssh` for Urbit, leveraging the built-in identity system instead of needing to copy public keys around. :)

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -1386,10 +1386,6 @@
       ~&  acl
       `state
     ::
-        %json
-      ~&  jon=!<(json vase)
-      `state
-    ::
         %wipe
       ~&  %dojo-wipe
       =.  hoc
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -1372,16 +1372,16 @@
       =+  !<([=id =command:lens] vase)
       he-abet:(~(he-lens he hid id ~ (~(got by hoc) id)) command)
     ::
-        %grant-access
+        %allow-remote-login
       =/  who  !<(@p vase)
       `state(acl (~(put in acl) who))
     ::
-        %revoke-access
+        %revoke-remote-login
       =/  who  !<(@p vase)
       :_  state(acl (~(del in acl) who))
       [%give %kick ~ `who]~
     ::
-        %list-access
+        %list-remote-logins
       ~&  acl
       `state
     ::
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -1378,9 +1378,8 @@
     ::
         %revoke-access
       =/  who  !<(@p vase)
-      =/  =path  /sole/(cat 3 'drum_' (scot %p who))
       :_  state(acl (~(del in acl) who))
-      [%give %kick [path]~ `who]~
+      [%give %kick ~ `who]~
     ::
         %list-access
       ~&  acl
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -12,9 +12,10 @@
 =>  |%                                                  ::  external structures
     ++  id  @tasession                                  ::  session id
     ++  house                                           ::  all state
-      $:  $5
+      $:  $6
           egg/@u                                        ::  command count
           hoc/(map id session)                          ::  conversations
+          acl/(set ship)                                ::  remote access whitelist
       ==                                                ::
     ++  session                                         ::  per conversation
       $:  say/sole-share                                ::  command-line state
@@ -1349,9 +1350,12 @@
   !>(state)
 ::
 ++  on-load
-  |=  =old-state=vase
-  =/  old-state  !<(house old-state-vase)
-  `..on-init(state old-state)
+  |=  old=vase
+  ?:  ?=(%6 +<.old)
+    `..on-init(state !<(house old))
+  =/  old-5  !<([%5 egg=@u hoc=(map id session)] old)
+  =/  =house  [%6 egg.old-5 hoc.old-5 *(set ship)]
+  `..on-init(state house)
 ::
 ++  on-poke
   |=  [=mark =vase]
@@ -1359,6 +1363,7 @@
   =^  moves  state
     ^-  (quip card:agent:gall house)
     ?+  mark  ~|([%dojo-poke-bad-mark mark] !!)
+    ::
         %sole-action
       =/  act  !<(sole-action vase)
       he-abet:(~(he-type he hid id.act ~ (~(got by hoc) id.act)) act)
@@ -1366,6 +1371,20 @@
         %lens-command
       =+  !<([=id =command:lens] vase)
       he-abet:(~(he-lens he hid id ~ (~(got by hoc) id)) command)
+    ::
+        %grant-access
+      =/  who  !<(@p vase)
+      `state(acl (~(put in acl) who))
+    ::
+        %revoke-access
+      =/  who  !<(@p vase)
+      =/  =path  /sole/(cat 3 'drum_' (scot %p who))
+      :_  state(acl (~(del in acl) who))
+      [%give %kick [path]~ `who]~
+    ::
+        %list-access
+      ~&  acl
+      `state
     ::
         %json
       ~&  jon=!<(json vase)
@@ -1390,8 +1409,9 @@
 ++  on-watch
   |=  =path
   ^-  (quip card:agent:gall _..on-init)
-  ~?  !=(our.hid src.hid)  [%dojo-peer-stranger src.hid]
-  ?>  (team:title our.hid src.hid)
+  ?>  ?|  (team:title our.hid src.hid)
+          (~(has in acl) src.hid)
+      ==
   ?>  ?=([%sole @ ~] path)
   =/  id  i.t.path
   =?  hoc  (~(has by hoc) id)
--- a/pkg/arvo/gen/dojo/grant.hoon
+++ b/pkg/arvo/gen/dojo/grant.hoon
@@ -1,0 +1,9 @@
+::  grant: allow a ship to link to dojo
+::
+/?    310
+::
+:-  %say
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        [[=ship ~] ~]
+    ==
+[%grant-access ship]
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -633,8 +633,9 @@
 ::
 ++  se-peer                                           ::  send a peer
   |=  gyl/gill:gall
+  =/  =path  /sole/(cat 3 'drum_' (scot %p our.hid))
   %-  se-emit(fug (~(put by fug) gyl ~))
-  [%pass (en-gill gyl) %agent gyl %watch /sole/drum]
+  [%pass (en-gill gyl) %agent gyl %watch path]
 ::
 ++  se-pull                                           ::  cancel subscription
   |=  gyl/gill:gall
@@ -662,6 +663,8 @@
     |=  act/sole-action
     ^+  +>
     (ta-poke %sole-action !>(act))
+  ::
+  ++  ta-id  (cat 3 'drum_' (scot %p our.hid))        ::  per-ship duct id
   ::
   ++  ta-aro                                          ::  hear arrow
     |=  key/?($d $l $r $u)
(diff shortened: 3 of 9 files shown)


pkg/arvo/app/dojo.hoon
```diff
@@ -1349,16 +1350,33 @@
...
 ++  on-load
-  |=  =old-state=vase
-  =/  old-state  !<(house old-state-vase)
-  `..on-init(state old-state)
+  |=  old=vase
+  ?:  ?=(%6 +<.old)
+    `..on-init(state !<(house old))
+  =/  old-5  !<([%5 egg=@u hoc=(map id session)] old)
+  =/  =house  [%6 egg.old-5 hoc.old-5 *(set ship)]
+  `..on-init(state house)
```
lukechampine, line 1358: I hacked together this pretty sad looking state adapter -- feedback much appreciated

pkg/arvo/app/dojo.hoon
```diff
@@ -1349,16 +1350,33 @@
...
     ^-  (quip card:agent:gall house)
     ?+  mark  ~|([%dojo-poke-bad-mark mark] !!)
+    ::
+        %noun
+      ?+    q.vase  ~|([%dojo-poke-bad-noun vase] !!)
+          [%grant-access @p]
+        `state(acl (~(put in acl) +.q.vase))
+      ::
+          [%revoke-access @p]
+        `state(acl (~(del in acl) +.q.vase))
```
lukechampine, line 1373: I believe this also needs to kick any active subscriptions; currently, once you `|link`, you remain linked even after your access is revoked. Is that as simple as returning a `%kick` card here?

pkg/arvo/app/dojo.hoon
```diff
@@ -1390,8 +1408,10 @@
 ++  on-watch
   |=  =path
   ^-  (quip card:agent:gall _..on-init)
-  ~?  !=(our.hid src.hid)  [%dojo-peer-stranger src.hid]
-  ?>  (team:title our.hid src.hid)
+  ?>  ?|
+    (team:title our.hid src.hid)
+    (~(has in acl) src.hid)
+    ==
```
lukechampine, line 1414: wasn't sure how to format this. Should probably restore the `%dojo-peer-stranger` printf too.

pkg/arvo/app/dojo.hoon
philipcmonk: This is basically correct, certainly fine for a single upgrade.  More complex ones are usually structured like this:

https://github.com/urbit/urbit/blob/69b6495d3d4c8cd3d81c5842a2cc90bf8059d63c/pkg/arvo/sys/vane/gall.hoon#L1665-L1683

pkg/arvo/app/dojo.hoon
philipcmonk: Yes, this is what the `(unit ship)` in `%kick` is used for.  You don't have to specify the path.

pkg/arvo/app/dojo.hoon
philipcmonk: This is canonical

```
?>  ?|  (team:title our.hid src.hid)
        (~(has in acl) src.hid)
    ==
```

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -663,6 +664,8 @@
     ^+  +>
     (ta-poke %sole-action !>(act))
   ::
+  ++  ta-duct  (cat 3 'drum_' (scot %p our.hid))      ::  per-ship duct
```
philipcmonk, line 667: This isn't technically a "duct", though it's similar in spirit.  I would call it `+ta-tag`, `+ta-id`, `+ta-pipe`, or something like that.

ohAitch: style suggestion `(pack /drum/(scot %p our.hid))` should do the same thing
(or `(pack (dray /tas/p %drum our.hid))` but that's honestly overkill)

ohAitch: …who appear to no longer be around.
[These guys](https://github.com/urbit/arvo/blob/master@%7B2019-1-1%7D/sys/vane/eyre.hoon#L210-L218), though I could swear were in zuse at some point. Consider this suggestion penalized by the requirement to reintroduce library code

pkg/arvo/app/dojo.hoon
joemfb: This is great! It'd be nice to add generators (and probably marks) for these administrative pokes, then people could `:dojo|grant ~zod` or `:dojo|revoke ~zod`. You just put the generators in `gen/dojo/` to enable match that syntax.

pkg/arvo/app/dojo.hoon
lukechampine: Curious: How is it that Dojo didn't already have a state adapter, given that its state was on version 5? Do "builtin" apps not need full 0->1, 1->2, 2->3... state adapters because they're updated OTA?

pkg/arvo/lib/hood/drum.hoon
lukechampine: gotcha, I called it a duct because of how `sole-action` is defined:
```
++  sole-action                                         ::  sole to app
  $:  id=@ta                                            ::  duct id
    $=  dat
    ...
```
but I see how `id` would be a better name

pkg/arvo/app/dojo.hoon
lukechampine: Done. `%kick` still doesn't seem to be working, though; `~zod` remains linked to `~nec` even after being kicked. Not sure what's going on.

pkg/arvo/app/dojo.hoon
```diff
@@ -1349,23 +1350,40 @@
...
+      =/  who  !<(@p vase)
+      :_  state(acl (~(del in acl) who))
+      [%give %kick ~ `who]~
+    ::
+        %list-access
+      ~&  acl
+      `state
     ::
         %json
       ~&  jon=!<(json vase)
```
lukechampine, line 1389: btw, should this `%json` mark be removed? I don't see what it would be useful for.

philipcmonk: I suspect this was added for Soto but ultimately %sole-action was used directly.  @matildepark Do you know if that's true?

matildepark: We didn't touch dojo.hoon for Soto, as I remember it.

matildepark: The blames on that line go back a-ways for lens, talk, dojo conversions for mall.

pkg/arvo/app/dojo.hoon
philipcmonk: Hmm, it looks like it actually only kicks the subscriptions you specify, and the ship is just a filter on top of that.  I guess just specify the subscription path.  I think at some point we should change gall to have the semantics I thought it had.

https://github.com/urbit/urbit/blob/69b6495d3d4c8cd3d81c5842a2cc90bf8059d63c/pkg/arvo/sys/vane/gall.hoon#L1108-L1118

pkg/arvo/app/dojo.hoon
philipcmonk: It was upgraded in earlier continuity eras, this is the first state adapter since this era started in December.  I guess we deleted the adapters but didn't reset the version number to 0.

pkg/arvo/app/dojo.hoon
philipcmonk: Ah, lens makes sense.

Yes, pretty sure this can just be removed.

pkg/arvo/app/dojo.hoon
lukechampine: Great, it's working now -- zod gets unlinked as soon as nec runs `:dojo|revoke ~zod`.

Unfortunately this means that we now have `(cat 3 'drum_' (scot %p ship))` floating around in *four* separate places, two of those in Gall apps. I'd really like to consolidate those if possible.

ohAitch: uhh, I'd just kick all paths for that ship - as stands I think the acl also lets them connect _web_ dojo which then doesn't get kicked
(or any other hypothetical future client)

pkg/arvo/gen/dojo/grant.hoon
```diff
@@ -0,0 +1,9 @@
+::  grant: allow a ship to link to dojo
+::
+/?    310
+::
+:-  %say
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        [[=ship ~] ~]
+    ==
+[%grant-access ship]
```
lukechampine, line 9: do we want to print anything here? Maybe even prompting the user to confirm? I'm slightly concerned about unsuspecting users getting tricking into running `:dojo|grant ~scammer`

ohAitch: maybe just more explicit `:dojo|allow-remote-login`

ohAitch: or `:dojo|trust`

ohAitch: (for bonus points, allow e.g. `:dojo|trust 0c1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa` for setting up piñata planets, which you redeem as `:~sipnym-sipnym/dojo|redeem-trust our .^(@ cx+%/private/key)` from a comet, sending a message signed with the BTC address key that the sending comet be allowed to log in)

pkg/arvo/app/dojo.hoon
philipcmonk: This is true.  This is really a bug in gall.  There's code to clear all subscriptions from a ship if list of paths is null, but it's inside that `+turn`, which means it doesn't ever get run.  We should handle the `?~  target-path` case in `+ap-ducts-from-paths` before the `+zing` instead of in `+ap-ducts-from-path` (note pluralization).  In fact, `+ap-ducts-from-path` shouldn't take a `(unit path)` at all, but a `path`.

@lukechampine I can get to this in the next few days, or you're welcome to do it.  Just test it by taking away the subscription path in your `%kick`.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1110,31 +1110,32 @@
...
-        ?:  =(target-path `path)
-          `duct
-        ~
-      ?~  target-path
-        ?:  =(target-ship `ship)
+        ?:  =(target-path path)
           `duct
         ~
-      ?:  &(=(target-path `path) =(target-ship `ship))
+      ?:  &(=(target-path path) =(target-ship `ship))
```
lukechampine, line 1138: I think these conditionals could be merged now; the desired logic is: "if target-path is path AND (target-ship is ship OR target-ship is ~)". Is there a unit logic helper that can express this concisely?

pkg/arvo/app/dojo.hoon
lukechampine: I think I got it! Implementation can probably be cleaned up a bit, but it appears to work.

pkg/arvo/sys/vane/gall.hoon
philipcmonk: Not a commonly used one, and this is complex enough that I'd optimize for readability.  I think what you have here is fine, the other option that comes to mind is a tall-form `?&` with wide-form `?|`.  That could be more readable.

lukechampine: ok, let's go with that. I was thinking something like:
```
      ?:  &(=(target-path path) =((fall target-ship ship) ship))
        `duct
      ~
```
but it's probably best to be explicit about checking for `~`.

philipcmonk: Oh, I guess that would work.  For me, the explicit check is clearer, but either is fine.

pkg/arvo/gen/dojo/grant.hoon
philipcmonk: I like the `:dojo|trust` or `:dojo|allow-remote-login`

lukechampine: Let's do `allow-remote-login`. This isn't an operation that should be done frequently or lightly, so a long, scary name is preferable to a short one with multiple interpretations.

PR #2868, opened 2020-05-06
Add ContentType to uploaded s3 images.

### LINKED ISSUE #2865: s3: set Content-Type for uploaded media
Clicking on an uploaded image forces it to download, not view in the browser. [This appears to be linked to how we set the Content-Type on upload](https://stackoverflow.com/questions/14150854/aws-s3-display-file-inline-instead-of-force-download).

tylershuster:
Fixes #2865 

When uploading files to S3, the `ContentType` is not set as per the docs [here](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#upload-property).

This results in the S3 object not having the proper header set (instead it is application/octet-stream) so the images when clicked are downloaded instead of viewed natively.

Fortunately, the buffer object passed to the upload function contains the mime type as one of its properties.

Therefore, extracting this and passing it to S3 sets the ContentType properly and fixes the issue.

### DIFF
--- a/pkg/interface/chat/src/js/lib/s3.js
+++ b/pkg/interface/chat/src/js/lib/s3.js
@@ -33,7 +33,8 @@ export default class S3Client {
       Bucket: bucket,
       Key:  filename,
       Body: buffer,
-      ACL: 'public-read'
+      ACL: 'public-read',
+      ContentType: buffer.type
     };
     return new Promise((resolve, reject) => {
       if (!this.s3) {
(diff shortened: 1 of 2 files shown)


matildepark:
Would you mind rebasing your changes onto `feat/spa`? The applicable files would be in `pkg/interface/src/apps` there, with the same architecture inside.

tylershuster:
I hope you don't mean `git rebase` because I am not good with computer, but I can make a commit with the same changes

matildepark:
`git rebase` is a swear word around these parts. If you were to change the branch you start off working from and then make these changes to that branch, it would basically have to be a fresh PR, I think.

PR #2871, opened 2020-05-07
Branching scheme

philipcmonk:
We shouldn't release all changes to the network at the same pace, at least for now.  This is a branching scheme which gives developers more control over when their changes will be released.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -88,7 +88,7 @@ the code may lack state adapters, or it may not properly handle outstanding
 subscriptions.  It could also be code which is planned to be released only upon
 a breach (network-wide or rolling).
 
-In this case, the code maybe PR'd to a `na-release/` branch.  All rules are the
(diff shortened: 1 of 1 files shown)


MAINTAINERS.md
```diff
@@ -1,5 +1,113 @@
...
+another.
+
+### Non-OTAable release branches
+
+In some cases, work is completed which cannot be OTA'd as written.  For example,
+the code may lack state adapters, or it may not properly handle outstanding
+subscriptions.  It could also be code which is planned to be released only upon
+a breach (network-wide or rolling).
+
+In this case, the code maybe PR'd to a `na-release/` branch.  All rules are the
```
lukechampine, line 91:
```suggestion
In this case, the code may be PR'd to a `na-release/` branch.  All rules are the
```

philipcmonk: Thanks

PR #2875, opened 2020-05-08
Initial debug dashboard

Fang-:
Initial feature-set:

- Apps: list all apps, look at state and subscriptions if they use `/lib/dbug`, subscribe to (subset of) events if they use `/lib/verb`.
- Spider: list the thread tree, kill individual threads.
- Ames: view nearly all state for all peers and their flows.
- Behn: view timers.
- Eyre: view bindings.
- Can configure a passcode, using which unauthenticated strangers can view your ship's debug page also.

I tried merging in @philipcmonk's Pottery, but that depends on yet-unreleased clay changes, it seems?

Super quick demo, including passcode-based auth:

debug-dash

It isn't the prettiest, but it works!

Note that this touches zuse and various vanes in order to make the debug info we need scryable. As such, this might have to wait for livenet release until ford fusion.

Happy to take feature requests on this over time.


galenwp:
king

jtobin:
Confirmed to be an absolute gawd.

urcades:
perfect

vvisigoth:
great work @Fang- !!

philipcmonk:
Will look at this line-by-line in a bit, but CI says ames doesn't compile

philipcmonk:
I think the CI failures are just not having a pill

philipcmonk:
How do you run this?  I made a pill, ran `gulp watch` in `pkg/interface/dbug`, and it added a couple js files and a css file, but going to `/~debug` still gives a 404

Fang-:
Dumb Q, but you didn't mention explicitly: did you `|start %dbug`? If yes, I'll try and walk through the clean-slate flow tomorrow (or monday?) and see what's up here.

philipcmonk:
Yes, I did that, and I can load the JS and JSON files.  I think it's just missing an index.html or something

philipcmonk:
Now it loads to a blank page titled "Links", with console error "cannot use import statement outside a module" pointing to dbug/src/index.js

Fang-:
Oh right I never fixed the `<title>` on this thing. I'll take a more in-depth look soon!

philipcmonk:
Oh, gulp watch is giving an error

```
[17:25:41] Finished 'tile-js-bundle-dev' after 2.37 s
'/reducers/initial' is imported by dist/js/store.js, but could not be resolved – treating it as an external dependency
'/components/lib/message-screen' is imported by dist/js/components/root.js, but could not be resolved – treating it as an external dependency
{ Error: Could not resolve '../components/subscriptions' from dist/js/views/apps.js
    at error (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:10149:30)
    at ModuleLoader.handleMissingImports (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:17218:17)
    at ModuleLoader.<anonymous> (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:17269:26)
    at Generator.next (<anonymous>)
    at fulfilled (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:40:28)
  code: 'UNRESOLVED_IMPORT',
  message:
   'Could not resolve \'../components/subscriptions\' from dist/js/views/apps.js',
  name: 'Error',
  stack:
   'Error: Could not resolve \'../components/subscriptions\' from dist/js/views/apps.js\n    at error (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:10149:30)\n    at ModuleLoader.handleMissingImports (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:17218:17)\n    at ModuleLoader.<anonymous> (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:17269:26)\n    at Generator.next (<anonymous>)\n    at fulfilled (/Users/philip/urbit/pkg/interface/dbug/node_modules/rollup/dist/rollup.js:40:28)',
  __safety: undefined,
  _stack: undefined,
  plugin: 'gulp-better-rollup',
  showProperties: true,
  showStack: false }
```

philipcmonk:
Not urgent for me, feel free to get to it later

loganallenc:
Note: JS should be merged into the SPA as opposed to in this piecewise fashion.

Great work!

Fang-:
@philipcmonk I don't know how it successfully compiled for me when it was straight-up missing a component. lmk if this works for you.

> Note: JS should be merged into the SPA as opposed to in this piecewise fashion.

@loganallenc I'd be down for making use of the more generic components, code-wise, but I don't think merging it into the single-page _experience_ is a good idea. This does some funky things due to requirements around serving to non-authed visitors (like leaving out the header bar), and probably doesn't want to be a first-class OS1.1 citizen.

philipcmonk:
> lmk if this works for you

Getting the same error and `find pkg/interface/dbug | grep subscriptions` still gives nothing

philipcmonk:
This works for me now.  I loaded it on a -L copy of ~wicdev-wisryt, and I've been having a blast looking at what ~wicdev has been doing all this time.  Some notes, feel free to triage according to the time you wish to spend on this:

- Loading the state of some apps, like chat-store, takes forever (not sure if it finishes)
- Killing threads doesn't work.  Tested by running `-time ~s10` and clicking "kill".  Got console error `api.killThread is not a function at spider.killThread`
- Feature request: lots of useful ames state, but when you have your filters just right if you want to refresh you have to re-enter all your filters.  You could either put the filters into state (app/localstorage/url) or put a specific refresh button to be able to re-request whatever info you're looking at without clearing the filters.
- Feature request: for sending bones, filter on "active" (unsent messages or fragments)
- When generating a passcode, print (even if just ~&) a message describing what information is available to whoever you're giving it to: subscriptions, chat room names (past and present), neighbor names, etc, as well as how to revoke it.  Over time we can narrow this info but for now we can at least make them aware of it.
- Are bones sorted by... hex string or something?  They seem to go "0, 256, 1, 257, 4, 260, ..."
- In ames, if you have the nth ship expanded and then you change the filter, it shows the nth ship as "Loading...".  It should just close it if it's a different ship.
- Would be good to decode lanes: try `((soft ,[@tas @if @ud]) (cue lane))`, but we don't guarantee the lane format -- it's an opaque blob used by the runtime.  Galaxy lanes won't fit this type, for example.
- In `rcv`, nacks should be sorted by number.

- I didn't think about this until now, but I actually think it would be good to transpose bones and snd/rcv.  There's more structure between them than we're letting on.  @belisarius222 should check me, but basically, bones which are 0 mod 4 are "forward" bones (which have ducts), 1 mod 4 are "backward" bones.  2 and 3 mod 4 are naxplanation channels for each of those.  Forward and backward bones that are "next to each other" are not related, but the `snd` and `rcv` data for each bone is related and should be displayed next to each other (of course, you don't always have both for each bone).  Possibly naxplanation bones should be in the same box.

Loosely, `snd` on a bone is about the messages sent on it and their acknowledgments received; `rcv` is about the messages received on a bone and their acknowledgments sent.  Some bones you only send messages -- eg if you're just sending pokes and getting acknowledgments, there will be no `rcv` entry.  Some bones you only receive messages -- eg if you're only receiving pokes.  Some bones you will do both -- eg subscriptions (watch/leave upstream and fact/kick downstream).  Of course, you may also poke and subscribe on the same bone (in fact, if they're on different bones, they're not guaranteed to arrive in order).

custom-jonathan:
Butter my biscuit and call me impressed.

belisarius222:
Collating naxplanation bones with their associated normal bones sounds useful. Each flow gets four bone numbers, of which it uses three: one for forward (outgoing / call) messages, one for backward (incoming / take) messages, and a naxplanation bone.

I think just listing all bones in numeric order would colocate related information pretty well. It would be nice to add the duct as a header for a flow, if it has one. Some flows were started by the peer, in which case we don't have an associated duct (they send %plea's, we respond with %boon's). If we sent the first message, though (we send %plea's, they send %boon's), the duct on which we sent that is probably the most natural identifier for that flow. The ossuary for that peer stores the mapping from bone to duct, so it should be pretty straightforward to display the duct.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
The confusing part is that the forward and backward bones at `n` and `n+1` are not related (where `n = 0 mod 4`), but the naxplanation bone at `n+2` is related to `n`.  There should be no `n+3` bone since backward bones don't have nacks.

Is that correct?  If so, then I think you ideally want to display `n`'s `snd` and `rcv` entries next to `n+2`'s `snd` and `rcv` entries.  You can display `n+1`'s `snd` and `rcv` entries somewhere else (needn't even be nearby; you could do all forward bones with their naxplanations followed by all backward bones).

belisarius222:
Hm, I think that's almost entirely right.  On flows where we can send %plea's, we can also receive %boon's and receive naxplanations.  On flows where we can receive %plea's, we can send %boon's and send naxplanations.  So yeah, I think displaying the snd bones separately from the rcv bones makes sense, and then showing each naxplanation bone next to its primary bone could be nice.

—
~rovnys-ricfer
https://urbit.org

belisarius222:
Sorry, I meant separating all forward bones from backward bones makes sense.  Both kinds can have naxplanations; forward bones can receive naxplanations, and backward bones can send them.

—
~rovnys-ricfer
https://urbit.org

Fang-:
Got everything except the in/active filtering. Will try to get to that tomorrow. In the mean time, please verify this forward/backward flow separation is what you guys meant.

image

philipcmonk:
This is great, seems to work well.  The "query state" textbox is great.  `-hi ~bud` is great because it lets you test ames refresh, behn timers, and spider kill.

I believe the layout of forward/backward bones is correct.  It looks like you read the duct for bone `n+1` from `n`, but backward bones just don't have ducts (or rather, the duct is stored on the other ship).

Fang-:
> It looks like you read the duct for bone n+1 from n, but backward bones just don't have ducts (or rather, the duct is stored on the other ship).

Ah, I misunderstood talk about these bones above, then. Fixed!

Also allow you to pass "active" into the filter now, and get only active flows, as requested. Made an adjustment to the list search so that "active hood" searches for both "active" and "hood" in the item key, instead of literally "active hood". Should be more useful this way.

A bit unsure around behavior wrt unsent messages. Here's a screenshot after trying to `|hi ~dopzod` twice. The "current" and "next" display as expected, but "total unsent" and the message list don't seem to quite line up. Should I look for an error on my end?

image

Fang-:
Gonna try and throw some more Eyre endpoints in, please hold on merge. (But of course do just merge if I don't get back before sys OTA wants to go out.)

Fang-:
Adds support for inspecting Eyre's open channels, connections and sessions. Could prove useful.

image

(No, those are not the actual session keys/cookies. Yes, it looks like Eyre doesn't clean up expired cookies.)

Gonna wait with pill updating until #2926 gets merged, for which I'll need to resolve a merge conflict around behn's scry interface anyway.

Fang-:
I, uh, pushed a merge commit, but it got lost during gh downtime I think... 58e85fa

loganallenc:
JSON encoding / decoding should be put into a lib with the enjs and dejs arms following the new styleguide.

Fang-:
*curses at pill conflict*

> JSON encoding / decoding should be put into a lib with the enjs and dejs arms

Been thinking about this, and have come to the conclusion that "it's weird".

For the "userspace infrastructure" case, this usually makes sense. You're putting conversions for the data in a lib, so that it's more easily used by clients (as in, "client" agents).

Here, some data comes from vanes, some from apps, the generated json doesn't always include all the data, and cares about structure only for the sake of display in the debug dashboard. This makes the conversion here a bad fit for use outside the app.
(Even putting it into `/lib/dbug` instead of, say, `/lib/ames` etc. would be too much exposure imo.)

So, idk, a lib doesn't seem strictly better than putting them in the app itself, in this case. If it ever turns out we want json conversion for behn timers in other places, we can always write a better version and factor it out.

Issue #2878, opened 2020-05-08
Landscape Chat breaks for all viewers when rending large number as @t using #


rabsef-bicrym:
**Describe the bug**
~rabsef-bomb 2.0 causes all users in room to have rendering issues on front end of chat if they are/were in the room @ time of execution of bomb (see below for execution steps)

**To Reproduce**
1. Open a chat room
2. Attempt to cast a large number as a @t using the code entry method - I suggest `#`@t`~rabsef-bicrym`
3. Reload your browser
4. All chats will disappear from the left (you can still access them in cli)

**Expected behavior**
Error prone code being passed to the frontend of chat should be excluded from execution and rejected

**Screenshots**
[Screenshot of console](https://imgur.com/qz7PYFr)

**Desktop (please complete the following information):**
Any browser, any OS (attempted from phone and computer on multiple devices)

**Additional context**
sorry.

ohAitch:
Ooh have we tried putting right-to-left overrides in chat

somlus-savlev:
hiddev-dannut has the OTA but still has the same issue. Hard refresh with Ctrl-F5 does not correct the issue.

PR #2879, opened 2020-05-08
chat: virtualize message-to-json conversion

Fang-:
Bad `@t` text might crash and fully halt the json conversion, preventing any
data from being delivered to the frontend.

Straight-up virtualizing the entire thing feels a bit aggressive, but seems like the safest bet here. Maybe the issue is just in the `%code` `%output` rendering, which calls `+wash`, but this way we don't need to know anything about how _any_ of the `enjs` functions convert to strings.

Targeting master directly, as a proposal to push this out with some priority. The estimated amount of people affected by #2878 (which this fixes) is 20-50, but that number could grow at any time, and the issue's pretty big-impact.

### DIFF
--- a/pkg/arvo/lib/chat-json.hoon
+++ b/pkg/arvo/lib/chat-json.hoon
@@ -41,9 +41,16 @@
   (fall ((ot output+(ar dank) ~) a) ~)
 ::
 ++  lett
+  =,  enjs:format
   |=  =letter
   ^-  json
-  =,  enjs:format
+  =;  result=(each json tang)
+    ?-  -.result
+      %&  p.result
+      %|  (frond %text s+'[[json rendering error]]')
+    ==
+  %-  mule
+  |.
   ?-  -.letter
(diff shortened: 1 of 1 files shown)


loganallenc:
Uh... what is a “bad @t”

ohAitch:
`@t`7 would do it I suspect

Fang-:
Yeah, try running `` `@t`our`` or such a thing in dojo. It tries to print as a `@t`, but since some of the numbers in there aren't valid/printable characters, it crashes.

Fang-:
Got @loganallenc's approval out of band. Gonna get this out OTA now!

somlus-savlev:
FYI ~hiddev-dannut has the OTA but still shows the same behavior with no chats appearing.  Ctrl-F5 does not fix the behavior.

ohAitch:
Putting that in particular in the mule to localize the error seems desirable

PR #2882, opened 2020-05-08
links: namespace types

liam-fitzgerald:
Namespaces the links types according to the proposal in #2814 . I've made one minor change though, which is that the /sur/link-store file only contains types specific to the store. e.g. `action`, `update` and `initial`. All the other types should go in sur/link which should be imported with a tar. I've made this change because of the sheer amount of types that links uses. If this is ok, I'll revise the chat PR to match this. 

cc: @loganallenc @ixv


pkg/arvo/lib/link-store.hoon
```diff
@@ -1,7 +1,10 @@
 ::  link: social bookmarking
```
Fang-, line 1: I don't think I understand why this entire file was renamed from `/lib/link` to `/lib/link-store`. Makes sense for the json conversions, but the other functions here mostly operate on suite-wide data, rate?

liam-fitzgerald: none of these are used outside the link store though. imo the idea of 'suite-wide' data is kinda vague because the store itself should always store the 'suite-wide' data

pkg/arvo/lib/link-store.hoon
```diff
@@ -92,7 +95,7 @@
     [a b]
   --
 ::
-++  en-json
+++  enjs
   =,  enjs:format
   |%
```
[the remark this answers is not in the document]
liam-fitzgerald: probably a good idea

liam-fitzgerald:
should this be retargeted?

pkg/arvo/lib/link-store.hoon
Fang-: You're right that it is a bit vague. My thinking there is that, even if you're interacting only with one of the hooks, you're still pulling in the core data types and working with those. Pulling in the store lib for that (which could very well contain helpers for interacting with the store) makes is slightly less obvious what you're actually touching. But it's also not the biggest deal.

Fang-:
next-userspace seems right for this, yeah. This seems good to go afaic. The "suite-wide" thing can always be revisited if it turns out it's a useful distinction later.

PR #2887, opened 2020-05-11
chat: virtualize message-to-json only where needed

Fang-:
This is a follow-up to #2879, which was written a bit hastily. 

Previously we were virtualizing all message rendering operations, but it's only
+tank:enjs:format that we're worried about here.

Doing this more narrowly also allows us to specify a slightly more specific
error message.


Fang-:
Resolved merge conflict. @loganallenc please stamp.

Fang-:
CI failure seems spurious again? Probably fine to merge.

PR #2890, opened 2020-05-11
Add ContentType to uploaded s3 images (SPA)

### LINKED ISSUE #2865: s3: set Content-Type for uploaded media

tylershuster:
Dupe of #2868 for SPA version

FIxes #2865


matildepark:
Oh *yeah* we only do that in one place on spa. God we're good.

PR #2892, opened 2020-05-12
cttp: no '&' before first url query var

ryan-lingle:
this is more of an issue (and a seemingly failed attempt at patching it)

cttp seems to be incorrectly adding an ampersand to the first url query variable

ex:
```
http://127.0.0.1:23000/rates?currencyPairs=BTC_USD&storeID=examplestore
```

turns into:
```
http://127.0.0.1:23000/rates?&currencyPairs=BTC_USD&storeID=examplestore
```


loganallenc:
@joemfb @eglaysher please take a look

ryan-lingle:
thanks @joemfb. merged!

PR #2893, opened 2020-05-12
gall: send breach notifications to agents

ixv:
This PR adds an arm to the definition of agent, +on-rift, which gets called when an apps subscribers or subscribees breaches.

cc @philipcmonk I haven't tested this yet, this is just a draft PR. I just want to get some feedback because I didn't entirely understand the email you sent me about how to do the upgrade logic.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1067,26 +1067,36 @@
       |=  =ship
       ^+  ap-core
       =/  in=(list [=duct =^ship =path])
-        ~(tap by incoming.subscribers.current-agent)
-      |-  ^+  ap-core
-      ?^  in
-        =?  ap-core  =(ship ship.i.in)
-          =/  core  ap-load-delete(agent-duct duct.i.in)
-          core(agent-duct agent-duct)
-        $(in t.in)
-      ::
+        %+  skim
+          ~(tap by incoming.subscribers.current-agent)
+        |=([duct who=^ship path] =(who ship))
       =/  out=(list [[=wire =^ship =term] ? =path])
-        ~(tap by outgoing.subscribers.current-agent)
-      |-  ^+  ap-core
-      ?~  out
-        ap-core
-      =?  ap-core  =(ship ship.i.out)
-        =/  core
-          =.  agent-duct  system-duct.agents.state
-          =/  way  [%out (scot %p ship) term.i.out wire.i.out]
-          (ap-specific-take way %kick ~)
+        %+  skim
+          ~(tap by outgoing.subscribers.current-agent)
+        |=([[wire who=^ship term] ? path] =(who ship))
+      ::
+      =.  ap-core
+        %+  roll  in
+        |=  [[=duct *] _ap-core]
+        =/  core  ap-load-delete(agent-duct duct)
         core(agent-duct agent-duct)
-      $(out t.out)
+      =.  ap-core
+        %+  roll  out
+        |=  [[[=wire =^ship =term] ? path] _ap-core]
+          =/  core
+            =.  agent-duct  system-duct.agents.state
+            =/  way  [%out (scot %p ship) term wire]
+            (ap-specific-take way %kick ~)
+          core(agent-duct agent-duct)
+      ::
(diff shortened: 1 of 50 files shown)


ixv:
Another thing I'm unsure about, aside from the upgrade logic, is what to do with the stack trace if on-rift crashes. You can see I just stick it in a variable `maybe-tang` and do nothing with it.

joemfb:
Is it necessary that this be unconditional? It seems like %gall has enough information to handle most cases itself, and apps could subscribe for notifications if they have additional breach-handling requirements.

I believe `+on-fail` should be called with the trace in case of error. Which raises another question: if we're going to change the app-interface, should we also upgrade `+on-fail to take a `$goof` (and maybe the original input too)?

loganallenc:
Might be better as an addition to `+on-arvo` than its own arm.

ixv:
@joemfb I hadn't considered that, that makes sense to me, though I'm not sure what you mean by %gall being able to handle most cases itself. Are you proposing something beyond what it currently does, of just kicking and unsubbing from breach ships?

belisarius222:
I think @philipcmonk had some thoughts about this too

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1067,26 +1067,36 @@
...
+            =/  way  [%out (scot %p ship) term wire]
+            (ap-specific-take way %kick ~)
           core(agent-duct agent-duct)
-        $(in t.in)
       ::
-      =/  out=(list [[=wire =^ship =term] ? =path])
-        ~(tap by outgoing.subscribers.current-agent)
-      |-  ^+  ap-core
-      ?~  out
+      ?:  &(?=(~ in) ?=(~ out))
```
Fang-, line 1092: Shouldn't `+on-rift` be called regardless of the amount of incoming and outgoing subscriptions?

Fang-:
An imagined use case for this was "send re-invite to group, to members who breached". The existence/use of export-import logic (on the breacher's side) would be a superior solution to most such use cases, in that they could re-establish a connection with the group themselves.

Accounting for that, I'm having some trouble imagining very many use cases for breach notification handling in applications. An explicit subscription for breach notifications might be better then. (And any boilerplate arising from that could likely be handled in a library.)

pkg/arvo/sys/vane/gall.hoon
philipcmonk: Which ships should you receive notifications about?  Gall hears about breaches for any ships it's talked to, but I don't think an app should receive notifications about a ship it hasn't talked to.

philipcmonk:
> Another thing I'm unsure about, aside from the upgrade logic, is what to do with the stack trace if on-rift crashes. You can see I just stick it in a variable maybe-tang and do nothing with it.

@joemfb is correct this should go into +on-fail.  Look at the call sites of +ap-error to see other places this is done.

> Might be better as an addition to +on-arvo than its own arm.

I suggested doing this as a separate arm because I assumed this would be unconditional, and if you don't send any %arvo moves, then you don't get anything in +on-arvo.  You could unconditionally send stuff into +on-arvo anyway.  The line between "new arm" and "case of existing arm" is fuzzy, though.

> An explicit subscription for breach notifications might be better then.

I don't think a subscription to Jael quite works because you really want to know if you're getting notified before or after Gall, and probably what subscriptions Gall culled.  So Gall should be in the loop somewhere. You could either subscribe directly to Gall or have Gall intercept the subscription request to Jael, which isn't a great precedent to set (it breaks the rule that if you send the same thing from different apps, the same thing happens), though it was quite common in dynamic gall.

Subscribing directly to gall makes sense, but we don't really have a place for such a move, unless you just do `[%rift app=term]` task in gall, making no attempt to only allow that app to call it.  Then +ap-breach just gives the result (including the culled subscriptions) instead of invoking the core directly.  I think either this or the current +on-rift are good options.

> Accounting for that, I'm having some trouble imagining very many use cases for breach notification handling in applications.

I'm actually not familiar with the use cases either.

ixv:
Closing this for now, maybe we'll revisit it after Liam's import/export stuff lands, if we find out theres more we need to do.

PR #2894, opened 2020-05-12
zuse: allow yel in chrd:de-xml:html

pkova:
Allow the literal " (double-quote) character to appear in character data sections of xml, for example `<a>"hello"</a>`. Following the spec at https://www.w3.org/TR/2006/REC-xml11-20060816/#dt-chardata.

The xml parsing adventure continues.

Ping @Fang-.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:66c71e256923835bf9aee19f6bf7ea301f187230fe9cb4d7c695a98cb612019b
-size 12884154
+oid sha256:749dbf182ef35fd4324538966b0afeb89a22d0477e39e626df440fc49fd7fdda
+size 12884083
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -6493,7 +6493,7 @@
     ::                                                  ::  ++chrd:de-xml:html
     ++  chrd                                            ::  character data
       %+  cook  |=(a/tape ^-(mars ;/(a)))
-      (plus ;~(less yel ;~(pose (just `@`10) escp)))
+      (plus ;~(pose (just `@`10) escp))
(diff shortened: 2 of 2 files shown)


joemfb:
I'm a little hesitant about this -- obviously quotes are valid in XML, but they were explicitly excluded here. I don't know the `$manx` (et al) system well enough to have any idea if this change might lead to problems elsewhere. But it's worth noting that the `&quot;` entity is specifically supported, bi-directionally. I suspect these arms would be better-named `+en-xhtml` and `+de-xhtml`:

```
> (de-xml:html '<foo>&quot;bar&quot;</foo>')
[~ [g=[n=%foo a=~] c=~[[g=[n=%$ a=~[[n=%$ v="\"bar\""]]] c=~]]]]
> (en-xml:html (need (de-xml:html '<foo>&quot;bar&quot;</foo>')))
"<foo>&quot;bar&quot;</foo>"
```

@ohAitch, do you have insights into this area?

pkova:
Some git spelunking below.

Origin of the line: https://github.com/urbit/urbit/commit/5987ff2f8d86a2c0f1c52cfcbe8e10d66b94c6d5#diff-d6e03a3cd4d7330235230da89c6d4999R911, throwing away both " and '. This makes sense to me, since the parser clearly assumes that the quotes only appear in their escaped form (`&quot;` and `&apos;`). This was changed to the current line in https://github.com/urbit/urbit/commit/3796d56de7#diff-d6e03a3cd4d7330235230da89c6d4999R709, where for some reason the single-quote is now allowed. In any case either quote does not need to be escaped inside chardata according to the XML spec.

No idea about XHTML or whether this breaks something elsewhere!

ohAitch:
My thoughts mostly mirror "this parser is aimed at html more than xml, but
also hasn't been used in anger".

pkova:
Closing this so i get my master branch back lol!

tacryt-socryp:
@joemfb Any chance you can merge this?

PR #2895, opened 2020-05-12
Jet +by and +in gates

eglaysher:
This jets a bunch of gates inside +by and +in. (With a guest appearance for +find, which was also unjetted.)

This speeds up boot, but only by 5 seconds. The speedup is real, but I'm actually a bit disappointed with the gains compared to the effort.


ohAitch:
At some point we'll have data structure jets and not completely thrash the
cache…

pkg/urbit/jets/d/by_apt.c
```diff
@@ -0,0 +1,88 @@
...
+    return c3y;
+  }
+
+  u3_noun n_a, l_a, r_a;
+  u3_noun p_n_a, q_n_a;
+  u3x_trel(a, &n_a, &l_a, &r_a);
+  u3x_cell(n_a, &p_n_a, &q_n_a);
+
+  u3_noun u_l;
+  if (c3y == u3r_cell(l, 0, &u_l)) {
```
joemfb, line 19: I think these `u3r_cell()` calls (and subsequent `u3r_trell()` calls) are a subtle mismatch with the hoon. `?~` is not `?@`, but really just `?:(=(0 ...) ...)`.

It seems that each of these conditionals should be shaped like so:

```
if ( u3_nul != x ) {
  u3x_cell();
  ...
} 
```

pkg/urbit/jets/d/by_apt.c
```diff
@@ -0,0 +1,88 @@
+/* j/4/by_apt.c
+**
+*/
+#include "all.h"
+
+static
+c3_y _by_apt(u3_noun a, u3_noun l, u3_noun r)
```
joemfb, line 7: This should be `c3_o` for loobean, not `c3_y` for arbitrary byte.

pkg/urbit/jets/d/in_apt.c
```diff
@@ -0,0 +1,66 @@
...
+{
+  if (a == 0) {
+    return c3y;
+  }
+
+  u3_noun n_a, l_a, r_a;
+  u3x_trel(a, &n_a, &l_a, &r_a);
+
+  u3_noun u_l;
+  if (c3y == u3r_cell(l, 0, &u_l)) {
```
joemfb, line 17: Same comment re: conditionals as for `+apt:by` above.

pkg/urbit/jets/tree.c
```diff
@@ -1635,28 +1649,46 @@ static c3_c* _141_two__in_ha[] = {
...
-  // static c3_c* _141_two__by_uni_ha[] = {0};
+  static u3j_harm _141_two__by_uni_a[] = {{".2", u3wdb_uni, c3y}, {}};
+  static c3_c* _141_two__by_uni_ha[] = {0};
 
 static u3j_core _141_two__by_d[] =
-  { { "bif", 7, _141_two__by_bif_a, 0, _141_two__by_bif_ha },
+  { { "all", 7, _141_two__by_all_a },
+    { "any", 7, _141_two__by_any_a },
+    { "apt", 7, _141_two__by_apt_a },
+    { "bif", 7, _141_two__by_bif_a, 0, _141_two__by_bif_ha },
```
joemfb, line 1676: I kinda hate to ask, but it would be better to maintain consistency in these declarations and stub out the hash vectors for all the new arms here. I'll probably go through and fill updated hashes again at some point. Until we've replaced hashboard with the new jetpack, we should fully maintain this file as is.

joemfb:
Looks like this needs to be merged-with/rebased-onto master for the `|start` fix. And of course there are pill conflicts all over, and there will be more as these PRs get merged ...

PR #2899, opened 2020-05-13
jets: hook up preexisting +repn jet.

eglaysher:
The C code for +repn was written, but never hooked up.

### DIFF
--- a/pkg/urbit/include/jets/w.h
+++ b/pkg/urbit/include/jets/w.h
@@ -66,6 +66,7 @@
     u3_noun u3wc_rep(u3_noun);
     u3_noun u3wc_rev(u3_noun);
     u3_noun u3wc_rip(u3_noun);
+    u3_noun u3wc_repn(u3_noun);
     u3_noun u3wc_ripn(u3_noun);
     u3_noun u3wc_rsh(u3_noun);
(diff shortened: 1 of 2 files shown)


philipcmonk:
@benjamin-tlon You wrote this, right?  Was there a reason it wasn't turned on?

joemfb:
It was added (and turned on) in #2066, as part of the OS1 work. There were issues with the management of the relevant branches -- that and OTA questions lead to #2254 and #2306. The tree.c parts of #2066 are notably missing from from the final OS1 PR: #2365.

joemfb:
The CI failure here was strange and presumably spurious; the test app just didn't start for some reason. Re-running the build to check.

joemfb:
Welp, the same CI failure occurred. Hard for me to imagine how improved base64 performance could impact start of the `:test` app. @philipcmonk, any ideas about what might've go wrong in app initialization?

matildepark:
Does this branch have 54969afdd25d0843f16797b0302e1f0179122daf in it? If it precedes that commit then apps won't `|start`.

eglaysher:
Rebased this on top of current master, which has the `|start` fix, and made stub hash declarations.

PR #1020, opened 2018-08-10
crypto (blake, ripe) jetted

BernardoDeLaPlaz:

PR #2902, opened 2020-05-15
Jet AES-SIV mode

eglaysher:
@belisarius222 mentioned that he saw that `+{en,de}:sivc:aes:crypto` were taking a millisecond each and appeared to be unjetted. This vendors libaes_siv and jets the `+{en,de}:siv?:aes:crypto` gates against it.

### DIFF
--- a/pkg/urbit/jets/e/aes_siv.c
+++ b/pkg/urbit/jets/e/aes_siv.c
@@ -28,7 +28,7 @@ static void u3r_bytes_reverse(c3_w    a_w,
 
 static u3_noun _siv_en(c3_y* key_y,
                        c3_w keysize,
-                       u3_atom ads,
+                       u3_noun ads,
                        u3_atom txt)
 {
   AES_SIV_CTX* ctx = AES_SIV_CTX_new();
@@ -61,12 +61,10 @@ static u3_noun _siv_en(c3_y* key_y,
   c3_y* txt_y = u3a_malloc(txt_w);
   u3r_bytes_reverse(0, txt_w, txt_y, txt);
 
-  c3_w iv_w = 16;
-  c3_y* iv_y = u3a_malloc(iv_w);
-  c3_w out_w = txt_w;
-  c3_y* out_y = u3a_malloc(out_w);
+  const c3_w iv_w = 16;
+  c3_y iv_y[iv_w];
+  c3_y* out_y = u3a_malloc(txt_w);
   if ( 0 == AES_SIV_EncryptFinal(ctx, iv_y, out_y, txt_y, txt_w) ) {
-    u3a_free(iv_y);
     u3a_free(out_y);
     u3a_free(txt_y);
     AES_SIV_CTX_free(ctx);
@@ -77,24 +75,21 @@ static u3_noun _siv_en(c3_y* key_y,
   AES_SIV_CTX_free(ctx);
 
   // Read the first 16 bytes as the "iv"
-  u3_noun iv_n = u3i_bytes(16, iv_y);
-  u3_noun msg_n = u3i_bytes(txt_w, out_y);
+  u3_noun iv = u3i_bytes(16, iv_y);
+  u3_noun msg = u3i_bytes(txt_w, out_y);
 
   // Reverse byte order for output
-  u3_noun rev_iv_n = u3qc_rev(3, iv_w, iv_n);
-  u3_noun rev_msg_n = u3qc_rev(3, txt_w, msg_n);
+  u3_noun rev_iv = u3kc_rev(3, iv_w, iv);
+  u3_noun rev_msg = u3kc_rev(3, txt_w, msg);
 
-  u3a_free(iv_y);
   u3a_free(out_y);
-  u3z(iv_n);
-  u3z(msg_n);
 
-  return u3nt(rev_iv_n, txt_w, rev_msg_n);
+  return u3nt(rev_iv, u3i_words(1, &txt_w), rev_msg);
 }
 
 static u3_noun _siv_de(c3_y* key_y,
                        c3_w keysize,
-                       u3_atom ads,
+                       u3_noun ads,
                        u3_atom iv,
                        u3_atom len,
                        u3_atom txt)
@@ -125,17 +120,16 @@ static u3_noun _siv_de(c3_y* key_y,
     ads = u3t(ads);
   }
 
-  c3_w txt_w = u3r_word(0, len);  // ?
+  c3_w txt_w = u3r_word(0, len);
   c3_y* txt_y = u3a_malloc(txt_w);
   u3r_bytes_reverse(0, txt_w, txt_y, txt);
 
-  c3_w iv_w = 16;
-  c3_y* iv_y = u3a_malloc(iv_w);
+  const c3_w iv_w = 16;
+  c3_y iv_y[iv_w];
   u3r_bytes_reverse(0, 16, iv_y, iv);
 
   c3_y* out_y = u3a_malloc(txt_w);
   if ( 0 == AES_SIV_DecryptFinal(ctx, out_y, iv_y, txt_y, txt_w) ) {
-    u3a_free(iv_y);
     u3a_free(out_y);
     u3a_free(txt_y);
     AES_SIV_CTX_free(ctx);
@@ -150,27 +144,26 @@ static u3_noun _siv_de(c3_y* key_y,
   AES_SIV_CTX_free(ctx);
 
   // Read the first 16 bytes as the "iv"
-  u3_noun msg_n = u3i_bytes(txt_w, out_y);
+  u3_noun msg = u3i_bytes(txt_w, out_y);
 
   // Reverse byte order for output
-  u3_noun rev_msg_n = u3qc_rev(3, txt_w, msg_n);
+  u3_noun rev_msg = u3kc_rev(3, txt_w, msg);
 
-  u3a_free(iv_y);
   u3a_free(out_y);
-  u3z(msg_n);
 
-  return u3nc(0, rev_msg_n);
+  return u3nc(0, rev_msg);
 }
 
 
 u3_noun
 u3qea_siva_en(u3_atom key,
-              u3_atom ads,
+              u3_noun ads,
               u3_atom txt)
 {
   c3_y key_y[32];
-  AES_KEY key_u;
-  c3_assert(u3r_met(3, key) <= 32);
+  if (u3r_met(3, key) > 32) {
+    return u3_none;
+  }
 
   u3r_bytes_reverse(0, 32, key_y, key);
   return _siv_en(key_y, 32, ads, txt);
@@ -194,14 +187,15 @@ u3wea_siva_en(u3_noun cor)
 
 u3_noun
 u3qea_siva_de(u3_atom key,
-              u3_atom ads,
+              u3_noun ads,
               u3_atom iv,
               u3_atom len,
               u3_atom txt)
 {
   c3_y key_y[32];
-  AES_KEY key_u;
-  c3_assert(u3r_met(3, key) <= 32);
+  if (u3r_met(3, key) > 32) {
+    return u3_none;
+  }
 
   u3r_bytes_reverse(0, 32, key_y, key);
   return _siv_de(key_y, 32, ads, iv, len, txt);
@@ -229,12 +223,13 @@ u3wea_siva_de(u3_noun cor)
 
 u3_noun
 u3qea_sivb_en(u3_atom key,
-              u3_atom ads,
+              u3_noun ads,
               u3_atom txt)
 {
   c3_y key_y[48];
-  AES_KEY key_u;
-  c3_assert(u3r_met(3, key) <= 48);
+  if (u3r_met(3, key) > 48) {
+    return u3_none;
+  }
 
   u3r_bytes_reverse(0, 48, key_y, key);
   return _siv_en(key_y, 48, ads, txt);
@@ -258,14 +253,15 @@ u3wea_sivb_en(u3_noun cor)
 
 u3_noun
 u3qea_sivb_de(u3_atom key,
-              u3_atom ads,
+              u3_noun ads,
               u3_atom iv,
               u3_atom len,
               u3_atom txt)
 {
   c3_y key_y[48];
-  AES_KEY key_u;
-  c3_assert(u3r_met(3, key) <= 48);
+  if (u3r_met(3, key) > 48) {
+    return u3_none;
+  }
 
   u3r_bytes_reverse(0, 48, key_y, key);
   return _siv_de(key_y, 48, ads, iv, len, txt);
@@ -294,12 +290,13 @@ u3wea_sivb_de(u3_noun cor)
 
 u3_noun
 u3qea_sivc_en(u3_atom key,
-              u3_atom ads,
+              u3_noun ads,
               u3_atom txt)
 {
   c3_y key_y[64];
-  AES_KEY key_u;
-  c3_assert(u3r_met(3, key) <= 64);
+  if (u3r_met(3, key) > 64) {
+    return u3_none;
+  }
 
   u3r_bytes_reverse(0, 64, key_y, key);
   return _siv_en(key_y, 64, ads, txt);
@@ -324,14 +321,15 @@ u3wea_sivc_en(u3_noun cor)
 
 u3_noun
 u3qea_sivc_de(u3_atom key,
-              u3_atom ads,
+              u3_noun ads,
               u3_atom iv,
               u3_atom len,
               u3_atom txt)
 {
   c3_y key_y[64];
-  AES_KEY key_u;
-  c3_assert(u3r_met(3, key) <= 64);
+  if (u3r_met(3, key) > 64) {
+    return u3_none;
+  }
 
   u3r_bytes_reverse(0, 64, key_y, key);
   return _siv_de(key_y, 64, ads, iv, len, txt);
--- a/pkg/urbit/jets/e/aes_siv.c
+++ b/pkg/urbit/jets/e/aes_siv.c
@@ -89,7 +89,69 @@ static u3_noun _siv_en(c3_y* key_y,
   return u3nt(rev_iv_n, txt_w, rev_msg_n);
 }
 
+static u3_noun _siv_de(c3_y* key_y,
+                       c3_w keysize,
+                       u3_atom ads,
+                       u3_atom iv,
+                       u3_atom len,
+                       u3_atom txt)
+{
+  AES_SIV_CTX* ctx = AES_SIV_CTX_new();
+  if ( 0 == AES_SIV_Init(ctx, key_y, keysize) ) {
+    AES_SIV_CTX_free(ctx);
+    return u3m_bail(c3__exit);
+  }
 
+  while (u3_nul != ads) {
+    c3_w ad_w = u3r_met(3, u3h(ads));
+    c3_y* ad_y = u3a_malloc(ad_w);
+    u3r_bytes_reverse(0, ad_w, ad_y, u3h(ads));
+
+    c3_w ret = AES_SIV_AssociateData(ctx, ad_y, ad_w);
+    u3a_free(ad_y);
+
+    if ( 0 == ret ) {
+      AES_SIV_CTX_free(ctx);
+      return u3m_bail(c3__exit);
+    }
+
+    ads = u3t(ads);
+  }
+
+  c3_w txt_w = u3r_word(0, len);  // ?
+  c3_y* txt_y = u3a_malloc(txt_w);
+  u3r_bytes_reverse(0, txt_w, txt_y, txt);
+
+  c3_w iv_w = 16;
+  c3_y* iv_y = u3a_malloc(iv_w);
+  u3r_bytes_reverse(0, 16, iv_y, iv);
+
+  c3_y* out_y = u3a_malloc(txt_w);
+  if ( 0 == AES_SIV_DecryptFinal(ctx, out_y, iv_y, txt_y, txt_w) ) {
+    u3a_free(iv_y);
+    u3a_free(out_y);
+    u3a_free(txt_y);
+    AES_SIV_CTX_free(ctx);
+
+    // Dcryption failed or signature bad.
+    return 0;
+  }
+
+  u3a_free(txt_y);
+  AES_SIV_CTX_free(ctx);
+
+  // Read the first 16 bytes as the "iv"
+  u3_noun msg_n = u3i_bytes(txt_w, out_y);
+
+  // Reverse byte order for output
+  u3_noun rev_msg_n = u3qc_rev(3, txt_w, msg_n);
+
+  u3a_free(iv_y);
+  u3a_free(out_y);
+  u3z(msg_n);
+
+  return u3nc(0, rev_msg_n);
+}
 
 
 u3_noun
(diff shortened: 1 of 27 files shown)


eglaysher:
This builds, passes tests, and I've tested that one fake ship can `|hi` another fake ship.

eglaysher:
Also, unlike the other jetting patches, I think this one can actually go out this era? The `~/` hints are already in `hoon.hoon`. You should be able to just run a new interpreter.

eglaysher:
Changed the libaes-siv library so it's using the openssl allocator and then set the openssl allocator to use the loom allocators in the worker startup path.

Let's talk speedup. On doing the initial sync from ~zod to ~marzod (stopwatch from seeing `beginning sync from %kids on ~zod to %base` to `sync succeeded from %kids on ~zod to %base`), it takes 27s without this patch and takes 23s with it.

eglaysher:
The build failure looks spurious. How do I retry the build? I went on to Travis and there doesn't appear to be a retry button?

joemfb:
You have to login to travis-ci, then there are "retry build" and "retry job" buttons. I restarted the failing job (sub-build).

pkg/urbit/noun/manage.c
```diff
@@ -1633,6 +1634,11 @@ u3m_boot(c3_c* dir_c)
   */
   u3m_init();
 
+  /* In the worker, set the openssl memory allocation functions to always
+  ** work on the loom.
+  */
+  CRYPTO_set_mem_functions(u3a_malloc, u3a_realloc, u3a_free);
```
joemfb, line 1640: I'm really surprised this worked. From the OpenSSL docs:

```
 int CRYPTO_set_mem_functions(
         void *(*m)(size_t, const char *, int),
         void *(*r)(void *, size_t, const char *, int),
         void (*f)(void *, const char *, int))
```


And from the build output:

```
noun/manage.c: In function 'u3m_boot':
noun/manage.c:1640:28: warning: passing argument 1 of 'CRYPTO_set_mem_functions' from incompatible pointer type [-Wincompatible-pointer-types]
   CRYPTO_set_mem_functions(u3a_malloc, u3a_realloc, u3a_free);
                            ^~~~~~~~~~
In file included from noun/manage.c:10:0:
/nix/store/nx7qi0av2h4bfw76fakm6b8a73g27vzx-openssl-1.1.1-x86_64-linux-musleabi/include/openssl/crypto.h:261:5: note: expected 'void * (*)(size_t,  const char *, int) {aka void * (*)(long unsigned int,  const char *, int)}' but argument is of type 'void * (*)(size_t) {aka void * (*)(long unsigned int)}'
 int CRYPTO_set_mem_functions(
     ^~~~~~~~~~~~~~~~~~~~~~~~
noun/manage.c:1640:40: warning: passing argument 2 of 'CRYPTO_set_mem_functions' from incompatible pointer type [-Wincompatible-pointer-types]
   CRYPTO_set_mem_functions(u3a_malloc, u3a_realloc, u3a_free);
                                        ^~~~~~~~~~~
In file included from noun/manage.c:10:0:
/nix/store/nx7qi0av2h4bfw76fakm6b8a73g27vzx-openssl-1.1.1-x86_64-linux-musleabi/include/openssl/crypto.h:261:5: note: expected 'void * (*)(void *, size_t,  const char *, int) {aka void * (*)(void *, long unsigned int,  const char *, int)}' but argument is of type 'void * (*)(void *, size_t) {aka void * (*)(void *, long unsigned int)}'
 int CRYPTO_set_mem_functions(
     ^~~~~~~~~~~~~~~~~~~~~~~~
noun/manage.c:1640:53: warning: passing argument 3 of 'CRYPTO_set_mem_functions' from incompatible pointer type [-Wincompatible-pointer-types]
   CRYPTO_set_mem_functions(u3a_malloc, u3a_realloc, u3a_free);
                                                     ^~~~~~~~
In file included from noun/manage.c:10:0:
/nix/store/nx7qi0av2h4bfw76fakm6b8a73g27vzx-openssl-1.1.1-x86_64-linux-musleabi/include/openssl/crypto.h:261:5: note: expected 'void (*)(void *, const char *, int)' but argument is of type 'void (*)(void *)'
 int CRYPTO_set_mem_functions(
     ^~~~~~~~~~~~~~~~~~~~~~~~
```

The default functions are not OPENSSL_malloc/realloc/free, but CRYPTO_malloc/realloc_free.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,356 @@
...
+                       u3_atom txt)
+{
+  AES_SIV_CTX* ctx = AES_SIV_CTX_new();
+  if ( 0 == ctx ) {
+    return u3m_bail(c3__fail);
+  }
+
+  if ( 0 == AES_SIV_Init(ctx, key_y, keysize) ) {
+    AES_SIV_CTX_free(ctx);
+    return u3m_bail(c3__exit);
```
joemfb, line 42: It's not at all clear to me that `AES_SIV_Init()` failing can only correspond to deterministic failures represented in the hoon/nock. If that's clear to you, can you add a comment explaining the failure conditions? And if it's not, please switch this to `c3__fail`.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,356 @@
...
+    c3_w ad_w = u3r_met(3, u3h(ads));
+    c3_y* ad_y = u3a_malloc(ad_w);
+    u3r_bytes_reverse(0, ad_w, ad_y, u3h(ads));
+
+    c3_w ret = AES_SIV_AssociateData(ctx, ad_y, ad_w);
+    u3a_free(ad_y);
+
+    if ( 0 == ret ) {
+      AES_SIV_CTX_free(ctx);
+      return u3m_bail(c3__exit);
```
joemfb, line 55: Same failure-determinism concern.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,356 @@
...
+  c3_w iv_w = 16;
+  c3_y* iv_y = u3a_malloc(iv_w);
+  c3_w out_w = txt_w;
+  c3_y* out_y = u3a_malloc(out_w);
+  if ( 0 == AES_SIV_EncryptFinal(ctx, iv_y, out_y, txt_y, txt_w) ) {
+    u3a_free(iv_y);
+    u3a_free(out_y);
+    u3a_free(txt_y);
+    AES_SIV_CTX_free(ctx);
+    return u3m_bail(c3__exit);
```
joemfb, line 74: Same failure-determinism concern.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,356 @@
...
+                       u3_atom txt)
+{
+  AES_SIV_CTX* ctx = AES_SIV_CTX_new();
+  if ( 0 == ctx ) {
+    return u3m_bail(c3__fail);
+  }
+
+  if ( 0 == AES_SIV_Init(ctx, key_y, keysize) ) {
+    AES_SIV_CTX_free(ctx);
+    return u3m_bail(c3__exit);
```
joemfb, line 110: Same failure-determinism concern.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,356 @@
...
+    c3_w ad_w = u3r_met(3, u3h(ads));
+    c3_y* ad_y = u3a_malloc(ad_w);
+    u3r_bytes_reverse(0, ad_w, ad_y, u3h(ads));
+
+    c3_w ret = AES_SIV_AssociateData(ctx, ad_y, ad_w);
+    u3a_free(ad_y);
+
+    if ( 0 == ret ) {
+      AES_SIV_CTX_free(ctx);
+      return u3m_bail(c3__exit);
```
joemfb, line 123: Same failure-determinism concern.

eglaysher:
(For those curious about all the tryjobs here, there seems to be a weird issue where we're running out of loom, even though I've reverted things back to the state where it previously passed, and I'm trying to figure out what's going on.)

eglaysher:
OK. So this passes entirely when I disable attaching the openssl allocator to the loom, even though I previously had that working previously. I have no idea what's going on.

eglaysher:
I've minimized the previous changes of builds down to whether the `CRYPTO_set_mem_functions` line is activated or not.

- When I have allocator replaced, I get a "address 0 out of loom!", which I
assume is just choking on a null pointer somewhere.

  - Example: https://travis-ci.org/github/urbit/urbit/jobs/691470871

- When I disable setting the allocator, things work:

  - Example: https://travis-ci.org/github/urbit/urbit/jobs/691481597

My question is how do I go about debugging why this crashes on the CI? Is there a way to get a stack trace out of Travis on this (I assume) null pointer dereference?

joemfb:
Core dumps used to be automatic in the old CI setup. The gdb script is still around (`sh/print-core-backtrace`), but it doesn't appear to be wired up to anything in nix.

You can run these nix builds locally; is this crash not reproducible on MacOS? If so, it's probably simplest to just run them on an ubuntu vm, so as to have a similar environment to CI. I can't remember if the core dumps are still created when the tests within nix. If not, that might be an easy fix, but you can also just manually run the tests via herb, or even directly on a ship.

eglaysher:
It appears that OpenSSL calls whatever malloc function passed in to allocate several internal structures on first use and assign the returned memory to static variables. After adding debug logging to show the balance of mallocs vs frees, there's always 2 more mallocs than frees, and the first two mallocs are of sizes that are never allocated again / aren't the sizes of the structures used in libaes_siv.

We have something called `u3a_malloc`, but it isn't a conformant malloc because the memory it allocates can be freed without calling `u3a_free`--doing so is the whole point of the function.

Can you just run the initialization code and then switch the allocator? Nope. After the first call to `CRYPTO_malloc()`, [you can no longer switch the allocator](https://docs.huihoo.com/doxygen/openssl/1.0.1c/mem_8c_source.html).

Do we have to heap allocate the structures used in siv mode? Can't we just modify libaes_siv? We could, but that doesn't help because the openssl structures that libaes_siv allocates are sealed C structs.

We effectively can't change the allocator for OpenSSL, at least to anything loom like.

ohAitch:
Surely we could hand it a pointer to a shim, which just calls a vere-side
function pointer we *can* change dynamically?

joemfb:
I think the specific problem here is that we're not explicitly initialization OpenSSL, so these initial allocations aren't happening on the home road. If they were, their locations and lifetimes would be stable (barring something like `|pack` which already has to deal with these sort of complexities).

But even if we do that, we'll run into other problems. OpenSSL is notoriously difficult to fully uninitialize: https://wiki.openssl.org/index.php/Library_Initialization. Per their docs, it's just not possible to account for every single allocation.

Along the lines that @ohAitch has suggested, it seems that we could further special-case OpenSSL's use of the loom. On the home road, we would need to track additional roots to be able to clean up on exit or resume (similar to `u3j_ream()`). On inner roads we could just pass-thru to u3a_malloc(). This is obviously a lot of complexity.

I owe you an apology, @eglaysher, for suggesting this would be straightforward. I had run into these OpenSSL-cleanup issues in the past, attempting to make vere play nice with valgrind, but I forgot all about it in this context.

eglaysher:
OK. The secret to why this is still failing after trying to warm up the aes_siv stuff is that openssl is allocating memory in caches in the preexisting openssl code: it allocates today and not on the loom.

eglaysher:
OK, assuming this squash/cleanup patch goes green, this is now ready for review. The problem was not openssl code, but was usage of non-loom malloc in preexisting, non-aes encryption code.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+    c_y[i_w] = hi;
+    c_y[b_w - i_w - 1] = lo;
+  }
+
+  return;
+}
+
+static u3_noun _siv_en(c3_y* key_y,
+                       c3_w keysize,
+                       u3_atom ads,
```
joemfb, line 31: S/b `u3_noun`

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+
+    ads = u3t(ads);
+  }
+
+  c3_w txt_w = u3r_met(3, txt);
+  c3_y* txt_y = u3a_malloc(txt_w);
+  u3r_bytes_reverse(0, txt_w, txt_y, txt);
+
+  c3_w iv_w = 16;
+  c3_y* iv_y = u3a_malloc(iv_w);
```
joemfb, line 65: This could just be on the stack, since it's a fixed size.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+  u3a_free(txt_y);
+  AES_SIV_CTX_free(ctx);
+
+  // Read the first 16 bytes as the "iv"
+  u3_noun iv_n = u3i_bytes(16, iv_y);
+  u3_noun msg_n = u3i_bytes(txt_w, out_y);
+
+  // Reverse byte order for output
+  u3_noun rev_iv_n = u3qc_rev(3, iv_w, iv_n);
+  u3_noun rev_msg_n = u3qc_rev(3, txt_w, msg_n);
```
joemfb, line 85: It's worth noting that the `+rev` jet leaves much to be desired. It might be better to have a `u3i_bytes_reverse()` here.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+    AES_SIV_CTX_free(ctx);
+    return u3_none;
+  }
+
+  u3a_free(txt_y);
+  AES_SIV_CTX_free(ctx);
+
+  // Read the first 16 bytes as the "iv"
+  u3_noun iv_n = u3i_bytes(16, iv_y);
+  u3_noun msg_n = u3i_bytes(txt_w, out_y);
```
joemfb, line 81: Noun variable names shouldn't have a type suffix.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+  // Reverse byte order for output
+  u3_noun rev_iv_n = u3qc_rev(3, iv_w, iv_n);
+  u3_noun rev_msg_n = u3qc_rev(3, txt_w, msg_n);
+
+  u3a_free(iv_y);
+  u3a_free(out_y);
+  u3z(iv_n);
+  u3z(msg_n);
+
+  return u3nt(rev_iv_n, txt_w, rev_msg_n);
```
joemfb, line 92: This should be `u3i_words(1, &txt_w)`

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+  }
+
+  c3_w txt_w = u3r_met(3, txt);
+  c3_y* txt_y = u3a_malloc(txt_w);
+  u3r_bytes_reverse(0, txt_w, txt_y, txt);
+
+  c3_w iv_w = 16;
+  c3_y* iv_y = u3a_malloc(iv_w);
+  c3_w out_w = txt_w;
+  c3_y* out_y = u3a_malloc(out_w);
```
joemfb, line 67: `out_w` is only ever used here, `txt_w` is used subsequently. The alias doesn't seem to accomplish anything.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+  u3a_free(out_y);
+  u3z(iv_n);
+  u3z(msg_n);
+
+  return u3nt(rev_iv_n, txt_w, rev_msg_n);
+}
+
+static u3_noun _siv_de(c3_y* key_y,
+                       c3_w keysize,
+                       u3_atom ads,
```
joemfb, line 97: S/b `u3_noun`

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+
+    ads = u3t(ads);
+  }
+
+  c3_w txt_w = u3r_word(0, len);  // ?
+  c3_y* txt_y = u3a_malloc(txt_w);
+  u3r_bytes_reverse(0, txt_w, txt_y, txt);
+
+  c3_w iv_w = 16;
+  c3_y* iv_y = u3a_malloc(iv_w);
```
joemfb, line 133: Small and fixed size, could be on the stack.

pkg/urbit/jets/e/aes_siv.c
joemfb: Also, these could use the `u3k` convention to simply refcounts.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+  u3_noun msg_n = u3i_bytes(txt_w, out_y);
+
+  // Reverse byte order for output
+  u3_noun rev_msg_n = u3qc_rev(3, txt_w, msg_n);
+
+  u3a_free(iv_y);
+  u3a_free(out_y);
+  u3z(msg_n);
+
+  return u3nc(0, rev_msg_n);
```
joemfb, line 162: Same comments as above wrt the `_n` suffix, `+rev` jet, refcount convention, &c

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+
+u3_noun
+u3qea_siva_en(u3_atom key,
+              u3_atom ads,
```
joemfb, line 168: s/b `u3_noun`

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+       c3n == u3ud(txt) ) {
+    return u3m_bail(c3__exit);
+  } else {
+    return u3qea_siva_en(key, ads, txt);
+  }
+}
+
+u3_noun
+u3qea_siva_de(u3_atom key,
+              u3_atom ads,
```
joemfb, line 197: s/b `u3_noun`

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+
+
+u3_noun
+u3qea_siva_en(u3_atom key,
+              u3_atom ads,
+              u3_atom txt)
+{
+  c3_y key_y[32];
+  AES_KEY key_u;
+  c3_assert(u3r_met(3, key) <= 32);
```
joemfb, line 173: This is a `bail:oops`, which could theoretically be handled like any `bail:fail`, but currently aborts the process. What is the hoon/nock behavior in this case?

pkg/urbit/jets/e/aes_siv.c
joemfb: Same for all subsequent `_siv*_en` functions.

pkg/urbit/jets/e/aes_siv.c
joemfb: Same for all subsequent `c3_assert()` calls.

pkg/urbit/jets/e/aes_siv.c
joemfb: Same for all subsequent `_siv*_de` functions.

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+
+u3_noun
+u3qea_siva_en(u3_atom key,
+              u3_atom ads,
+              u3_atom txt)
+{
+  c3_y key_y[32];
+  AES_KEY key_u;
```
joemfb, line 172: This is unused. (Same for all other `AES_KEY` declarations.)

pkg/urbit/jets/e/ripe.c
```diff
@@ -17,15 +17,7 @@
...
-    static EVP_MD_CTX* con_u = NULL;       // context
-
-    /* build library context object
-       we do this once (and only once)
-    */
-
-    if (NULL == con_u) {
-      con_u = EVP_MD_CTX_create();
-    }
+    EVP_MD_CTX* con_u = EVP_MD_CTX_create();
```
joemfb, line 20: This is a great catch. It was added in #1020, and I'm embarrassed to say that I approved it ...

With this resolved, have you confirmed that there are no longer any "persistent" OpenSSL allocations that would require special de-initialization on exit? (And therefore, special accommodations in crash recovery?)

pkg/urbit/jets/e/ripe.c
```diff
@@ -37,23 +29,28 @@
     ret_w = EVP_DigestInit_ex(con_u, rip_u, NULL);
     if ( 1 != ret_w ) {
       u3a_free(dat_y);
+      EVP_MD_CTX_destroy(con_u);
       u3l_log("\rripe jet: crypto library fail 1\n");
       return u3m_bail(c3__exit);
```
joemfb, line 34: While you're in here, this should be `bail:fail`

pkg/urbit/jets/e/ripe.c
```diff
@@ -37,23 +29,28 @@
...
     }
 
     ret_w = EVP_DigestUpdate(con_u, (void*)dat_y, wid);
     u3a_free(dat_y);
     if (1 != ret_w) {
+      EVP_MD_CTX_destroy(con_u);
       u3l_log("\rripe jet: crypto library fail 2\n");
       return u3m_bail(c3__exit);
```
joemfb, line 42: `bail:fail`

pkg/urbit/jets/e/ripe.c
```diff
@@ -37,23 +29,28 @@
...
     }
 
     ret_w = EVP_DigestFinal_ex(con_u, sib_y, &sil_w);
     if ( 1 != ret_w ) {
+      EVP_MD_CTX_destroy(con_u);
       u3l_log("\rripe jet: crypto library fail 3\n");
       return u3m_bail(c3__exit);
```
joemfb, line 49: `bail:fail`

pkg/urbit/jets/e/aes_siv.c
```diff
@@ -0,0 +1,357 @@
...
+  c3_w iv_w = 16;
+  c3_y* iv_y = u3a_malloc(iv_w);
+  c3_w out_w = txt_w;
+  c3_y* out_y = u3a_malloc(out_w);
+  if ( 0 == AES_SIV_EncryptFinal(ctx, iv_y, out_y, txt_y, txt_w) ) {
+    u3a_free(iv_y);
+    u3a_free(out_y);
+    u3a_free(txt_y);
+    AES_SIV_CTX_free(ctx);
+    return u3_none;
```
joemfb, line 73: Commenting for posterity ...

As we've discussed previously/elsewhere, the normal way to handle these failures in a jet is to simply `bail:fail` (since not all failure cases are deterministic). One advantage of bailing is that no local cleanup is necessary.

This pattern of partial-jetting for error-handling purposes is not wrong, but requires to doing more work on failure cases. In other jets, that more work could be so disproportionate as to be intractable. But this could has been unjetted all along, so we know that's not case. This pattern seems a little too cute, but not enough to be a problem.

pkg/urbit/jets/e/aes_siv.c
eglaysher: I'm going to ignore this given the preexisting usage of rev in a lot of our other jet code.

eglaysher: (The u3kc comment otoh, I acted on.)

pkg/urbit/jets/e/aes_siv.c
eglaysher: This pattern was taken from the other AES jets. There actually are no sizing checks anywhere in the hoon code. I've changed these to also be partial failures and to fallback to the nock.

pkg/urbit/jets/e/ripe.c
eglaysher: Yes. I hooked the allocator and ensured that the number of outstanding ssl allocations returned to zero on a whole +test.

(Bisecting with that code with +test was how I zeroed in on this being the problem which was causing the crashing.)

eglaysher:
OK. PTAL.

PR #2909, opened 2020-05-19
http-server-store: dynamic static webpage serving

ixv:
Adds static site functionality to landscape.hoon

Two landscape actions are added, %serve-dir and %unserve-dir, they let you set and unset a url base path associated with a clay base path. All files under that clay base path will be served at the corresponding url.

A new authorization function is added that just does a simple passthrough of the response, rather than calling a handler on the request. We use this at a finer level of granularity, requiring authorization for OS1 stuff, but not for statically served stuff.


pkg/arvo/app/landscape.hoon
```diff
@@ -41,68 +46,155 @@
...
+    (handle-http-request req)
   ==
   ::
+  ++  landscape-action
+    |=  act=action
+    ^-  (quip card _this)
+    ?-  -.act
+        %serve-dir
+      :-  ~
+      this(serving (~(put by serving) url-base.act clay-base.act))
```
Fang-, line 65: Shouldn't this check for conflicts, and nack/crash if the binding is already in use?

pkg/arvo/app/landscape.hoon
```diff
@@ -41,68 +46,155 @@
...
-          [%'~dojo' *]      (get-file-at /app/landscape [t.site u.ext]:req-line)
-          [%'~groups' *]    (get-file-at /app/landscape [t.site u.ext]:req-line)
-          [%'~landscape' %js %session ~]
-        (some (as-octt:mimes:html "window.ship = '{+:(scow %p our.bowl)}';"))
-      ::
-          [%'~landscape' *]
-        (get-file-at /app/landscape [t.site u.ext]:req-line)
+    ?+  site.req-line  (get-custom req-line)
+    ::
+        [%'' %'index' ~]
```
Fang-, line 91: You probably want to `$?([%'' %index ~] [%'~link' *] etc)` all of these cases, so you don't duplicate their logic.

Curious about what the reasoning is for hardcoding these for the "core" apps, instead of having them set this up (like the custom case) during `+on-init`.

pkg/arvo/app/landscape.hoon
ixv: Ah good point.

pkg/arvo/app/landscape.hoon
ixv: Fair point about eliminating duplicate logic.

wrt to hardcoding, maybe @loganallenc can speak to that, that was a part I left unchanged from the old version. Was there any reason for hardcoding other than the fact that we just didn't have the ability to dynamically set these yet?

loganallenc: Nope, no reason for hardcoding other than lack of dynamicity. I was gonna tear that out and just have on-init set up all the bindings

loganallenc:
Looking great! One comment: there should be an action handler for setting the home-url-prefix in the configuration, as well as a subscription endpoint such that Landscape can subscribe to receive that home-url-prefix as a JSON string.

loganallenc:
Going to start working on this branch. Thanks @ixv!

pkg/arvo/app/http-server-store.hoon
```diff
@@ -0,0 +1,220 @@
...
+        |=(pax=path [pax [/app/landscape %.n]])
+      ==
+  :~  (connect /)
+      (connect /'~landscape')
+      (connect /'~chat')
+      (connect /'~link')
+      (connect /'~dojo')
+      (connect /'~groups')
+      (connect /'~publish')
+  ==
```
ixv, line 43: If you're serving the root path, why do you also need to serve these other paths. Also doesn't this contradict the earlier conversation we had about not hard-coding these?

loganallenc: They aren't hard-coded. They're dynamically registered in the state just like any other path would be and can be unbound at any time. Imagine a case in which you wanted to unbind the / path from the %http-server-store to use in some other context. If you don't already have the specific app paths bound, then suddenly everything breaks. If you do have the specific app paths bound, then everything still works as normal. Also, the `get-suffix` arm you had didn't work to serve the SPA until I made this change, so it isn't just theoretical...

pkg/arvo/app/http-server-store.hoon
```diff
@@ -0,0 +1,220 @@
...
+    ?.  ?|  ?&(?=(^ url-prefix) =((need url-prefix) i.site.req-line))
+            =(url-prefix ~)
+        ==
+      not-found:gen
+    ::
+    ?:  ?=([%'~landscape' %js %session ~] site.req-line)
+      %+  require-authorization-simple:app
+        inbound-request
+      %-  js-response:gen
+      (as-octt:mimes:html "window.ship = '{+:(scow %p our.bowl)}';")
```
ixv, line 129: This is my fault, but I don't think we actually need this if we have the +who generator.

pkg/arvo/app/http-server-store.hoon
ohAitch: Surely "not hard-coded" would involve, even in the initialization logic, pulling from the filesystem or sth rather than having an explicit list in the source code twice?

ixv: Okay, I understood something else by "dynamicity", I thought you wanted to have each app poke this one. This seems fine.

pkg/arvo/app/http-server-store.hoon
```diff
@@ -0,0 +1,220 @@
...
+      ==
+    ::
+    ++  get-clay-path
+      |=  pax=path
+      ^-  (unit [path ?])
+      =/  first-try  (match-clay-path pax (~(del by serving) /))
+      ?^  first-try  first-try
+      =/  root  (~(get by serving) /)
+      ?~  root  ~
+      (match-clay-path pax (~(gas by *^serving) [[/ u.root] ~]))
```
Fang-, line 164: This is... special-casing the `/` path, because every possible path is a suffix of it? You could probably rewrite `+match-clay-path` to find the _most specific_ match instead, so you don't have to do two passes here, and can also override with more specific content in other places, where necessary.

Eyre behaves this way too, wrt bindings.

pkg/arvo/app/http-server-store.hoon
loganallenc: Mark is right, yeah. Will move them to the individual apps.

pkg/arvo/app/http-server-store.hoon
```diff
@@ -0,0 +1,220 @@
+/-  *http-server-store
```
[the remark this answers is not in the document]
loganallenc: It is spiritually a store. I like `file-server-store` though.

pkg/arvo/app/http-server-store.hoon
```diff
@@ -0,0 +1,220 @@
...
+        %unserve-dir
+      :-  [%pass url-base.act %arvo %e %disconnect [~ url-base.act]]~
+      this(serving (~(del by serving) url-base.act))
+    ::
+        %toggle-permission
+      ?.  (~(has by serving) url-base.act)
+        ~|("url is not bound" !!)
+      =/  [clay-base=path public=?]  (~(got by serving) url-base.act)
+      :-  ~
+      this(serving (~(put by serving) url-base.act [clay-base !public]))
```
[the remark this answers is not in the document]
loganallenc: I’d prefer this not use that feature of Clay, as I prefer this approach. 🤷🏻‍♂️

pkg/arvo/app/http-server-store.hoon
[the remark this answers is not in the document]
Fang-: > It is spiritually a store.

What about the "serving data to the outside world" part? The only data it stores is configuration around what clay files to serve at what endpoints. If anything, it's a hook, but really it's just an application... "App" still exists as an architectural category, we shouldn't force ourselves to not use it.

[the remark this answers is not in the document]
loganallenc: Fair. `file-server` it is.

pkg/arvo/app/http-server-store.hoon
```diff
@@ -0,0 +1,220 @@
...
+    ?.  =(method.req %'GET')  not-found:gen
+    =?  req-line  ?=(~ ext.req-line)
+      [[[~ %html] ~['index']] args.req-line]
+    ?>  ?=(^ ext.req-line)
+    ?~  site.req-line
+      not-found:gen
+    =*  url-prefix  landscape-homepage-prefix.configuration
+    ?.  ?|  ?&(?=(^ url-prefix) =((need url-prefix) i.site.req-line))
+            =(url-prefix ~)
+        ==
```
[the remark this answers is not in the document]
loganallenc: The specificity is good imo. Normally I wouldn't like it, but I think it's positive here.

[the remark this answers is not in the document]
ohAitch: I'd break it up as
```
?~  landscape-homepage-prefix.configuration
  not-found:gen
?.  =(u.landscape-homepage-prefix.configuration i.site.req-line)
  not-found:gen
```
since the actual content of the if is pretty trivial. (And ban six-syllable 25-character field names so things don't get this awkward.)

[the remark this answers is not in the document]
ohAitch: …hm okay that doesn't work because `~` is the found case, `?. ?~(thing & =(u.thing req-line))`? In which case the fall makes sense ignore me ^-^

pkg/arvo/app/http-server-store.hoon
loganallenc: This is good as is. It's not doing any extra work with this approach.

ohAitch, line 164: > `*^serving`

Oh hey it's our friend "don't dump `=autonames` of unqualified types into the subject themselves unqualified, shadowing just makes everything confusing".

Also, would it perchance work to just

```suggestion
      (match-clay-path pax (my [/ u.root] ~))
```

pkg/arvo/app/http-server-store.hoon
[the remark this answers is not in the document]
Fang-: The middle ground I originally crossed over, and that @ohAitch almost arrived at, is just to do the `~` case check first, so you can do the value comparison more easily:

```hoon
?|  ?=(~ url-prefix)
    =(u.url-prefix i.site.req-line)
==
```

pkg/arvo/app/http-server-store.hoon
loganallenc: I like autonames.

pkg/arvo/app/http-server-store.hoon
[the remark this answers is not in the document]
loganallenc: Thanks, good suggestion.

loganallenc:
Should be good to go.

pkg/arvo/app/contact-view.hoon
```diff
@@ -28,20 +34,27 @@
...
+        ==
     ==
   ::
-  ++  on-save   on-save:def
+  ++  on-save   !>(state)
   ++  on-load
     |=  old=*
     ^-  (quip card _this)
-    :_  this
+    :_  this(dir-served %.y)
```
Fang-, line 51: Not sure why we need to store such a flag in state. It'd be no problem to just check for the pre-`%0` version (vase contains only `~`), and only run this if we're upgrading from that.

As it stands, the upgrade logic will get run on every `+on-load`, which may be undesirable, especially since it also `%disconnect`s some eyre endpoint that might have gotten reused elsewhere.

pkg/arvo/app/file-server.hoon
```diff
@@ -75,8 +70,12 @@
...
-      ?:  (~(has by serving) url-base)
-        ~|("url already bound to {<(~(got by serving) url-base.act)>}" !!)
+      =/  served  (~(get by serving) url-base)
+      ?^  served
+        ?.  =(clay-base.act clay-base.u.served)
+          ~|("url already bound to {<(~(got by serving) url-base.act)>}" !!)
+        :-  [%pass url-base %arvo %e %connect [~ url-base] %file-server]~
+        this(serving (~(put by serving) url-base [clay-base.act public.act]))
       :-  [%pass url-base %arvo %e %connect [~ url-base] %file-server]~
       this(serving (~(put by serving) url-base [clay-base.act public.act]))
```
Fang-: Being able to write this as a single `?:` aside, this now lets you double-connect on the same `url-base` as long as you're serving different files? But forbids you from serving the same files in multiple locations? The `%connect` here will fail if it's already bound by something...

None of that seems very sane to me. Is this intentional? Can you explain?

(Just realized you can actually scry into eyre to find out if there's a binding in place, but it's not the best interface. I'll try and put something nicer there soon, would be nice to be able to do checks on that.)

Fang-: (idk why it says "outdated" on this, I can still see the quoted code in the diff...)

loganallenc:
Thanks for being thorough. You're right, I had forgotten to add entries to some of the `+on-init`s. Fixing now.

loganallenc:
All issues resolved. Thanks!

pkg/arvo/app/file-server.hoon
loganallenc: Fixed the issue this solved via an alternative means (making +on-load arms stateful everywhere), reverted back to prior implementation.

pkg/arvo/app/chat-view.hoon
```diff
@@ -111,11 +128,13 @@
...
     (on-arvo:def wire sign-arvo)
   ::
-  ++  on-save  on-save:def
+  ++  on-save  !>(state)
   ++  on-load  
-    |=  old=*
+    |=  old-vase=vase
     ^-  (quip card _this)
-    :_  this
+    =/  old  ((soft state-0) q.old-vase)
```
Fang-, line 135: Cool, this actually works. (:

You could've also unpacked the vase into the `$@(~ state-0)` type instead of softing, but it does the trick.

pkg/arvo/app/contact-view.hoon
Fang-: I think you're fine to remove the flag from state, with your new changes, the logic only gets run once. (The `+on-init` path doesn't even set it, anyway.)

loganallenc:
Oh I had already made that change but hadn’t pushed. Will push and merge. Thanks!

PR #2914, opened 2020-05-20
os1: migrate app data to new group

ixv:
Adds a thread for doing all the pokes necessary to migrate chats/link collections/notebooks to a new group.

### DIFF
--- a/pkg/arvo/ted/migrate-group.hoon
+++ b/pkg/arvo/ted/migrate-group.hoon
@@ -1,0 +1,79 @@
+/-  spider, *metadata-store, *publish
+/+  strandio
+=,  strand=strand:spider
+^-  thread:spider
+|=  arg=vase
+=/  m  (strand ,vase)
+^-  form:m
+=/  [og-path=path ng-path=path ~]  !<([path path ~] arg)
+;<  bol=bowl:spider  bind:m  get-bowl:strandio
+::
+=/  gs-path  /(scot %p our.bol)/group-store/(scot %da now.bol)
+=/  ms-path  /(scot %p our.bol)/metadata-store/(scot %da now.bol)
+=/  p-path   /(scot %p our.bol)/publish/(scot %da now.bol)
+::
+=/  og=(unit (set ship))
+  .^((unit (set ship)) %gx :(weld gs-path og-path /noun))
+?~  og
+  (pure:m !>("no such group: {<og-path>}"))
+=/  ng=(unit (set ship))
+  .^((unit (set ship)) %gx :(weld gs-path ng-path /noun))
+?~  ng
+  (pure:m !>("no such group: {<ng-path>}"))
+::
+=/  ass=associations  .^(associations %gx :(weld ms-path /group og-path /noun))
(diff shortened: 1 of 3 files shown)


pkg/arvo/ted/migrate-group.hoon
```diff
@@ -0,0 +1,79 @@
...
+::
+|-
+=*  loop  $
+?~  ass-list
+  (pure:m !>("done"))
+=/  [[g-path=group-path res=resource] meta=metadata]  i.ass-list
+?.  =(our.bol creator.meta)
+  loop(ass-list t.ass-list)
+?.  =(g-path og-path)
+  loop(ass-list t.ass-list)
```
Fang-, line 35: Could this be an assertion instead, since we ask for metadata specifically about the `og-path` group?

pkg/arvo/ted/migrate-group.hoon
```diff
@@ -0,0 +1,79 @@
...
+  .^((unit (set ship)) %gx :(weld gs-path og-path /noun))
+?~  og
+  (pure:m !>("no such group: {<og-path>}"))
+=/  ng=(unit (set ship))
+  .^((unit (set ship)) %gx :(weld gs-path ng-path /noun))
+?~  ng
+  (pure:m !>("no such group: {<ng-path>}"))
+::
+=/  ass=associations  .^(associations %gx :(weld ms-path /group og-path /noun))
+=/  ass-list=(list [[group-path resource] metadata])  ~(tap by ass)
```
Fang-, line 25:
```suggestion
=/  assoc-list=(list [[g-path=group-path res=resource] meta=metadata])
  %~  tap  by
  .^(associations %gx :(weld ms-path /group og-path /noun))
```

If you add faces to the list type here, you can just `=,` them in below, instead of `=/`ing and writing out the type twice. Feels better imo, but also nbd.

pkg/arvo/ted/migrate-group.hoon
```diff
@@ -0,0 +1,79 @@
...
+    :~  :*  %pass  /poke  %agent
+            [our.bol %publish]
+            %poke  %publish-action
+            !>
+            :*  %edit-book
+                i.t.app-path.res
+                title.book
+                description.book
+                comments.book
+                `[ng-path ~ %.y %.n]
```
Fang-, line 65: I guess this `%edit-book` takes care of adding the association to `ng-path`, then? A bit awkward to need to provide the other details to do only that, may or may not be worth its own action.

Looking at publish code, it also sets up a new permission-group-hook mirroring. Doesn't that mean we need to remove the old one?

ixv: No, because permission-group-hook deletes its state when group-store receives an %unbundle.

ixv:
Changes as requested @Fang- please re-review.

Please also note that this is probably going to be very short lived code, since it will probably be made irrelevant when the new groups stuff lands. It's just meant to address some problems we're currently having on the network. So I care less about having the style reviewed, than about having the functionality reviewed.

PR #2915, opened 2020-05-20
behn: switch to ordered-map

philipcmonk:
We were using an ordered list for timers, which becomes quite slow when you have a lot of timers and are mostly adding timers to the end of the list.  This switches it to a priority queue, lifted straight from ames.  Note the length of the blue "set-timer" trace here.  Horizontal scales are not the same; the whole event went from 8ms to 2.5ms and the call to +set-timer went from 5.6ms to 0.06 ms, a 99% reduction.  This ship has approximately 1000 timers and log_2(1000) is 10, so we'd expect a speedup of about 100x, which is what we see.

Screen Shot 2020-05-19 at 9 32 39 PM

Screen Shot 2020-05-19 at 9 32 32 PM


loganallenc:
Could you please move this to zuse instead?

belisarius222:
We will eventually move it to Zuse, but it'll be less likely to cause problems when deployed to the network if we copy it into Behn for now.  There's already a TODO in Ames to move it into Zuse, but we should minimize changes to Zuse until Ford Fusion lands.

—
~rovnys-ricfer
https://urbit.org

loganallenc:
Okay. Given the TODO, sounds fine for now given the constraint around upgrading zuse.

philipcmonk:
Since our next /sys update will touch zuse, I've moved +ordered-map into zuse.

PR #2916, opened 2020-05-20
jets: fix broken jets so ~marzod can boot.

eglaysher:
This fixes an off by one error in ship name printing, and disables
the +apt:{by,in} jets. I've tried debugging the +apt jets and am
making no progress other than finding a bug in the recursion; this
isn't enough to get ~marzod to boot, though.


joemfb:
For posterity, this fixes a bug from #2853, and works-around another from #2895.

PR #2923, opened 2020-05-21
hoon: make +slab respect variance (according to passed in mode) and return false

### LINKED ISSUE #2861: +slab is broken
`+slab`, the standard library arm that purports to produce a flag indicating whether a limb exists in a type, is broken in two ways.  First, it fails to honor core variance, yielding false positives.  Second, in any case where it should produce `%.n`, it crashes instead.

Consider a lead core with one arm, `+foo`.  Trying to access anything other than 'foo' from this core should fail, and it does:
```
> =<  zing  ^?  |%  ++  foo  0  --
-find.zing
```
However, `+slab`, which purports to produce a flag indicating whether a limb exists in a type, cheerfully reports that `zing` can be called, since it would be accessible if the core variance rules were ignored (`+zing` is in the standard library, which became this core's context due to the `|%`):
```
> %+  slab  'zing'  =<  p  !>  ^?  |%  ++  foo  0  --
%.y
```
If you call `+slab` with anything that isn't in the type, it just crashes, instead of producing `%.n`.  Consider a core with the same battery, but without the standard library:
```
> %+  slab  'zing'  =<  p  !>  ^?  =>  ~  |%  ++  foo  0  --
-find.zing
```

pilfer-pandex:
Fixes #2861.


pkg/arvo/sys/hoon.hoon
```diff
@@ -12014,8 +12014,9 @@
   [p.gun (slum q.gat q.sam)]
 ::
 ++  slab                                                ::  test if contains
-  |=  {cog/@tas typ/type}
-  =(& -:(~(find ut typ) %free [cog ~]))
+  |=  {cog/@tas way/?(%read %rite %both) typ/type}
```
pilfer-pandex, line 12017: I forbid %free because it violates the principle of information hiding.

belisarius222: Ok.  A comment explaining that would be good, then.  While you're at it, I think you might as well modernize the syntax here.

joemfb:
It looks like `ford-fusion` doesn't currently have any changes to `hoon.hoon`. If we want to preserve the option to push it OTA, we probably  want this PR to target another branch.

philipcmonk:
I think it would be good to avoid changing hoon.hoon during the OTA of ford-fusion if we can manage it.  This can be targeted against a new `release/next-hoon`.

belisarius222:
Oh, I did make a minor change to hoon.hoon to clean up +ride printfs.  I'll revert that so we can OTA this more easily.

—
~rovnys-ricfer
https://urbit.org

joemfb:
@philipcmonk, the original changes here were released with ford-fusion, so this can be closed?

joemfb:
Closing, as this branch was released with ford fusion (merged in 3daecd8c1).

PR #2924, opened 2020-05-21
publish: fix publish-info mark quote escaping

ixv:
If you made a notebook with a `'` in the title or description, the mark wouldn't parse it properly, when converting from `%mime` to `%publish-info`. This fixes that.

### DIFF
--- a/pkg/arvo/mar/publish/info.hoon
+++ b/pkg/arvo/mar/publish/info.hoon
@@ -31,7 +31,7 @@
     ++  old-parser
       ;~  plug
         (key-val (jest 'owner: ~') fed:ag)
-        (key-val (jest 'title: ') (cook crip (star qit)))
+        (key-val (jest 'title: ') (cook crip (star prn)))
         (key-val (jest 'filename: ') sym)
         %+  key-val  (jest 'comments: ')
           ;~(pose (jest %open) (jest %closed) (jest %none))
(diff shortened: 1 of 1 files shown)


pkg/arvo/mar/publish/info.hoon
```diff
@@ -45,8 +45,8 @@
       ==
     ++  new-parser
       ;~  plug
```
ohAitch, line 47: `;~ (glue gaq)` instead of having `sfix ... gaq` in `key-val`? Would let the "optional trailing newline" `pose`s be cleaned up to just `sfix ... (punt gaq)`
Alternately, you could define `++  eol  ;~(pose gaq (full (easy ~)))` for line-or-EOF termination

pkg/arvo/mar/publish/info.hoon
```diff
@@ -45,8 +45,8 @@
...
-        (key-val (jest 'title: ') (cook crip (star qit)))
-        (key-val (jest 'description: ') (cook crip (star qit)))
+        (key-val (jest 'title: ') (cook crip (star prn)))
+        (key-val (jest 'description: ') (cook crip (star prn)))
         %+  key-val  (jest 'comments: ')
           (cook |=(a=@ =(%on a)) ;~(pose (jest %on) (jest %off)))
```
ohAitch, line 51: `(fuss %on %off)`? Also all the `jest`s might want to just be inside `key-val`

Fang-:
These are good recs, @ohAitch. Upon closer inspection the parsers here could do with a light rewrite. Happy to review/accept a PR with these changes!

PR #2926, opened 2020-05-22
drum: relink to chat-cli

philipcmonk:
This is mostly a merge of release/link-dojo and release/behn-fast.  The last commit is the only unreviewed code here, it's worth looking at.  Commit message:

The subscription changes in drum broke existing subscriptions.  This
worked alright (though loud) for dojo, but it left chat-cli "frozen"
unless you manually unlinked/relinked.  This does that automatically.

It also includes a refactoring of +on-load in drum, to avoid vain
repetition.


pkg/arvo/sys/vane/behn.hoon
```diff
@@ -335,7 +335,7 @@
     ~
   ?.  ?=(%timers syd)
     [~ ~]
-  [~ ~ %noun !>((turn (tap:timer-map timers) head))]
+  [~ ~ %noun !>(>(turn (tap:timer-map timers) head)<)]
```
joemfb, line 338: Worth noting that the debug dashboard (#2875) makes the opposite change here.

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -225,51 +225,27 @@
...
+      =<  (se-born %home %contact-view)
+      =<  (se-born %home %link-store)
+      =<  (se-born %home %link-proxy-hook)
+      =<  (se-born %home %link-listen-hook)
+      =<  (se-born %home %link-view)
+      (se-born %home %s3-store)
+    .
+  ?>  ?=(%4 ver)
+  =>  (se-drop:(se-pull our.hid %dojo) | our.hid %dojo)
+  (se-drop:(se-pull our.hid %chat-cli) | our.hid %chat-cli)
```
Fang-, line 248: It's not clear to me how this _reconnects_, after pulling the existing subscription. Dojo automatically re-links in `+se-drop`, but don't you want a manual `se-link` here for chat-cli?

pkg/arvo/sys/vane/behn.hoon
Fang-: Considering a couple things there might need to be updated to match, I think I should just resolve this as a merge conflict on #2875's side.

(Why we ever did, or would want to, produce a tank as a scry result in the first place, idk...)

pkg/arvo/lib/hood/drum.hoon
philipcmonk: `eel` is the set of apps we want to be connected to, and `fug` is the set of apps we believe we are connected to.  `+se-link` just puts it in `eel`, and later it gets linked in `+se-adze` (called in `+se-abet`).  `+se-drop` removes it from `fug`, and if the `pej` ("permanent") argument is yes, then it'll remove it from `eel` as well.  Since it's no, we don't, and it gets restarted.

`+se-pull` unsubscribes (which you don't need to do if you're unlinking because of a negative watch-ack, for example).  `+se-nuke` is almost exactly what I have except `pej` is yes.  `+se-drop` special cases dojo so that even if you run `|unlink %dojo`, you still resubscribee.

Fang-: Thanks for the explanation, this seems to check out!

ohAitch: Should the 10 se-born apps be collected into a `:~`? For that matter, is there an explicit "apps that should be running" data structure this could be drawing from?

PR #2932, opened 2020-05-24
shoe: console application library

Fang-:
In which we invent footwear that's actually comfortable to walk in.

Sole expects applications to do non-trivial logic, but applications often just want to implement some text commands. `/lib/shoe` can take care of that lower-level work.

Wrap as `((shoe command-type) your-agent)`, where `command-type` is the type of CLI commands for your app, and `your-agent` is a standard gall agent, with five extra arms: `+command-parser`, `+tab-list`, `+on-command`, `+on-connect`, and `+on-disconnect`. The extra arms will be called by `/lib/shoe` whenever it needs to deal with sole events.

Further documentation is in the lib itself, and the example `/app/shoe`.

This PR includes a commit for making chat-cli use sole. Modifies existing version upgrade logic instead of creating a new version, since `%2` hasn't gone out onto the network yet. Would be nice to get this in alongside that if possible, we need the re-link here too.

Going forward, we should try adding `shoe-effect`s along the lines of "draw table row with this data" or similar, allowing parts of output rendering to be taken up by `/lib/shoe`.

(Took the liberty of merging master into `release/next-sys` prior to these commits, to avoid a merge conflict in chat-cli.)

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -11,12 +11,11 @@
 ::
 /-  view=chat-view, hook=chat-hook,
     *permission-store, *group-store, *invite-store,
-    *rw-security, sole-sur=sole
-/+  sole-lib=sole, default-agent, verb, dbug, store=chat-store,
-    auto=language-server-complete
+    *rw-security, sole
+/+  shoe, default-agent, verb, dbug, store=chat-store
 ::
 |%
-+$  card  card:agent:gall
++$  card  card:shoe
 ::
 +$  versioned-state
   $%  state-2
@@ -35,7 +34,6 @@
       settings=(set term)                           ::  frontend flags
       width=@ud                                     ::  display width
       timez=(pair ? @ud)                            ::  timezone adjustment
-      soles=(map @ta sole-share:sole-sur)           ::  console state
   ==
 ::
 +$  state-1
@@ -49,7 +47,7 @@
       settings=(set term)                           ::  frontend flags
       width=@ud                                     ::  display width
       timez=(pair ? @ud)                            ::  timezone adjustment
-      cli=state=sole-share:sole-sur                 ::  console state
+      cli=state=sole-share:sole                     ::  console state
       eny=@uvJ                                      ::  entropy
   ==
 ::
@@ -63,7 +61,7 @@
       settings=(set term)                           ::  frontend flags
       width=@ud                                     ::  display width
       timez=(pair ? @ud)                            ::  timezone adjustment
-      cli=state=sole-share:sole-sur                 ::  console state
+      cli=state=sole-share:sole                     ::  console state
       eny=@uvJ                                      ::  entropy
   ==
 ::
@@ -111,13 +109,15 @@
 ::
 %-  agent:dbug
 %+  verb  |
-^-  agent:gall
+%-  (agent:shoe command)
+^-  (shoe:shoe command)
 =<
   |_  =bowl:gall
   +*  this       .
       talk-core  +>
       tc         ~(. talk-core bowl)
       def        ~(. (default-agent this %|) bowl)
+      des        ~(. (default:shoe this command) bowl)
   ::
   ++  on-init
     ^-  (quip card _this)
@@ -139,25 +139,9 @@
     =^  cards  state
       ?+  mark        (on-poke:def mark vase)
         %noun         (poke-noun:tc !<(* vase))
-        %sole-action  (sh-in:tc !<(sole-action:sole-sur vase))
       ==
     [cards this]
   ::
-  ++  on-watch
-    |=  =path
-    ^-  (quip card _this)
-    =^  cards  state  (peer:tc path)
-    [cards this]
-  ::
-  ++  on-leave
-    |=  =path
-    ^-  (quip card _this)
-    :-  ~
-    ?.  ?=([%sole @ ~] path)
-      this
-    this(soles (~(del by soles) i.t.path))
-  ::
-  ++  on-peek   on-peek:def
   ++  on-agent
     |=  [=wire =sign:agent:gall]
     ^-  (quip card _this)
@@ -181,8 +165,33 @@
       ==
     [cards this]
   ::
+  ++  on-watch  on-watch:def
+  ++  on-leave  on-leave:def
+  ++  on-peek   on-peek:def
   ++  on-arvo   on-arvo:def
   ++  on-fail   on-fail:def
+  ::
+  ++  command-parser
+    |=  sole-id=@ta
+    parser:sh:tc
+  ::
+  ++  tab-list
+    |=  sole-id=@ta
+    tab-list:sh:tc
+  ::
+  ++  on-command
+    |=  [sole-id=@ta =command]
+    =^  cards  state
+      (work:sh:tc command)
+    [cards this]
+  ::
+  ++  on-connect
+    |=  sole-id=@ta
+    ^-  (quip card _this)
+    [[prompt:sh-out:tc ~] this]
+  ::
+  ++  can-connect     can-connect:des
+  ++  on-disconnect   on-disconnect:des
   --
 ::
 |_  =bowl:gall
@@ -247,10 +256,6 @@
       grams  known  count
       bound  binds  audience
       settings  width  timez
-    ::
-      %+  ~(put by *(map @t sole-share:sole-sur))
-        (cat 3 'drum_' (scot %p our.bowl))
-      state.cli.u.old
     ==
   ::
   ?>  ?=(%2 -.u.old)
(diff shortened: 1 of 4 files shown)


pkg/arvo/lib/shoe.hoon
```diff
@@ -0,0 +1,341 @@
...
+  ::
+  ++  on-save   !>([shoe-inner=on-save:og shoe-self=state])
+  ::
+  ++  on-load
+    |=  old-state=vase
+    ^-  (quip card:agent:gall agent:gall)
+    ::  we could be upgrading from a shoe-less app, in which case the vase
+    ::  contains inner application state instead of our +on-save.
+    ::  to distinguish between the two, we check for the presence of our own
+    ::  +on-save faces in the vase.
```
Fang-, line 176: RFC, wondering if there's a better way to do this. Having this rely on the faces in `+on-save` feels a bit weird.

urcades:
Yes!

pkg/arvo/lib/shoe.hoon
ohAitch: Agree, case types should generally be *preserved*, not drive logic - dynamically dispatching calls to arms is sometimes the correct move, but even there much of the point of static gall was to do so less. (The double vase also feels weird, though I suppose short of `slop` it'd be hard to get away from.)

How would you feel about an explicit `!>([%shoe-app inner=on-save:og shoe-state=state])` with corresponding 
`?.  ?=([%shoe-app ^])`? Seems as unlikely to collide, and is in a real sense an outer layer state version.

pkg/arvo/app/chat-cli.hoon
```diff
@@ -779,37 +641,29 @@
...
-      ^-  (list card)
-      ::  echo commands into scrollback
-      ?.  =(`0 (find ";" buf))  ~
-      [(note:sh-out (tufa `(list @)`buf)) ~]
-    :_  cards
-    %+  effect-to:sh-out  sole-id
-    :-  %mor
-    :~  [%nex ~]
-        [%det cal]
+  ++  tab-list
```
liam-fitzgerald, line 646: I wonder if we could annotate the parser with tab completion instructions and then get nicer tab completion, e.g. completing `;leave ~dopzod/urbi`

pkg/arvo/lib/shoe.hoon
Fang-: > with corresponding `?. ?=([%shoe-app ^])`

How do you unpack (`!<`) to get to be able to do the `?=` though? Or did you mean to do that check as a vase operation also?

ohAitch: Sorry yep that's not valid hoon, `?=([%shoe-app ^] q.old-state)`. Having checked the value, we'll know the type lines up(I don't thing we're at risk of getting a fork type), so you don't even have to `sped` just `!<` directly like with the current `worn`.

ohAitch: And then like, `=^  old-inner  state  +:!<([%shoe-app vase state-0] old-state)`

pkg/arvo/app/chat-cli.hoon
```diff
@@ -111,13 +109,15 @@
...
-^-  agent:gall
+%-  (agent:shoe command)
+^-  (shoe:shoe command)
 =<
   |_  =bowl:gall
   +*  this       .
       talk-core  +>
       tc         ~(. talk-core bowl)
       def        ~(. (default-agent this %|) bowl)
+      des        ~(. (default:shoe this command) bowl)
```
philipcmonk, line 120: Might be reasonable to use `default:shoe` as a total replacement for `default-agent` so you don't have to think about whether an arm comes from `+agent` or `+shoe`.

pkg/arvo/app/chat-cli.hoon
Fang-: > annotate the parser

Wonder if you have a concrete idea of what this'd look like.

Applications could also dynamically generate the `+tab-list`, but that could result in very large suggestion sets. Maybe this should be a nested list instead?

(I don't care too deeply about improving this within this PR, but happy to facilitate future experimentation/development. (: )

pkg/arvo/lib/shoe.hoon
```diff
@@ -0,0 +1,341 @@
...
+  ++  tab-list
+    ~
+  ::
+  ++  on-command
+    |=  [sole-id=@ta command=command-type]
+    [~ shoe]
+  ::
+  ++  can-connect
+    |=  sole-id=@ta
+    ?|  =(~zod src.bowl)
```
philipcmonk, line 120: Always Trust Content from ~zod

I assume this was for testing?  Should be removed.

pkg/arvo/lib/shoe.hoon
```diff
@@ -0,0 +1,341 @@
...
+::
++$  card
+  $%  card:agent:gall
+      [%shoe sole-ids=(list @ta) effect=shoe-effect]  ::  ~ sends to all soles
+  ==
+::  $shoe-effect: easier sole-effects
+::
++$  shoe-effect
+  $%  [%sole effect=sole-effect]
+      ::TODO  complex screen-draw effects
```
philipcmonk, line 30: I like having this layer to put more complex screen-drawing before deciding to support it through the whole stack.

pkg/arvo/lib/shoe.hoon
philipcmonk: I think @ohAitch's solution is pretty good.  Although it's cheapest if you know the whole type ahead of time, it's fine to look into a vase's value or even `!<` to more and more specific types, especially when the performance doesn't much matter, as in `+on-load`.  Technically the former is a special case of the latter: `q.vase` is just `!<(* vase)`.

Nested vases are fine if the semantics are actually "this is dynamic data which contains data that is also dynamic even in the context where the original data is static", as is the case here.

pkg/arvo/lib/shoe.hoon
```diff
@@ -0,0 +1,341 @@
...
+  ++  tab-list
+    |~  sole-id=@ta
+    ::  (list [@t tank])
+    *(list (option:auto tank))
+  ::
+  ++  on-command
+    |~  [sole-id=@ta command=command-type]
+    *(quip card _^|(..on-init))
+  ::
+  ++  can-connect
```
philipcmonk, line 51: I like requiring them to implement `+can-connect` like this.  Makes you have to *specifically* say "I'm open to communication from other ships".  I don't like that our other arms fail open, maybe we should do something similar for those eventually.

pkg/arvo/app/shoe.hoon
```diff
@@ -0,0 +1,64 @@
+::  shoe: example usage of /lib/shoe
```
philipcmonk, line 1: Example apps are Good

pkg/arvo/app/chat-cli.hoon
ohAitch: I'd expose `on-tab` as a `$~([sole-id=@ta buf=(list @c) pos=@u] to-send=tape)` hook more generically, and have tab-list based handling be a library arm (or even a chat-specific arm, since dojo probably _would_ want to dynamically scry on `:app|` or w/e)

~~also `ost=bone` should be in the bowl instead of having to thread ids everywhere~~ I realize this ship has sailed

ohAitch: As for the mechanics of completion-oriented parsers - I suspect at some point we'll want a set of extended parser combinators, which deal in `(pair (unit thing-you-wanted) (list highlightable-token))` as opposed to just `(unit thing-you-wanted)`(or rather, as a further refinement/variant on `edge`). For LSP if anything, esp if we can run a pure-function highlighter without urbit's persistence latency constraints 😉.

Or even some intermediate structures, "well the cursor is on the second of three arguments for a generator that wants a `@da` there"; this may be just a dream ofc. It would be nice to have tab close all open parens whatever shape they may be, at least.

pkg/arvo/app/chat-cli.hoon
Fang-: Hmm, I've been going back and forth on whether I like that or not. Leaning towards not liking it.

Having to think about the distinction between the `+agent` and `+shoe` arms doesn't seem like the biggest deal, and I kind of like the emphasis on the "standard" vs "non-standard" separation.

(And yes, technically all of these arms come from `+shoe`, because they have the augmented `card` type... but the `+agent` ones still nest fine.)

Ultimately I don't think this library has any business saying anything about regular gall agent arms (outside of the cases it was designed to handle), even if that's just saying "look at these widely used defaults". The type system (wrt cores) already forces this into knowing more about `+agent` than it wants to, let's not add insult to injury.

This has been entirely too many words, because I don't feel very strongly about this, and it doesn't sound like you do either. Happy to discuss further, or let it be.

philipcmonk: I don't feel strongly either, let's leave it.

ohAitch:
I do have slight concerns over us growing three of these libraries and any
app can use only one, due to how the arms are arranged

But this bridge can certainly be crossed at a later date.

Fang-:
> three of these libraries

You're thinking of `verb` and `dbug`? Those take only standard agent cores, so should layer nicely _on top of_ anything that produces a standard agent core.

That's exactly the problem (and justified concern) with `shoe` though. Just now noticed that by necessity `dbug` wraps around the `shoe` instead of the inner app, making it much harder to get at application state. This could be remedied by having shoe (or apps themselves) implement a `/dbug/state` scry endpoint, for `dbug` to try and use before falling back to `+on-save`... but doesn't seem very elegant.

If it was _only_ the extra arms you might get away with taking the `+agent` core and the `+shoe` core as separate arguments, but they all also produce a superset of the `+card:agent`...

PR #2936, opened 2020-05-25
zuse: improve json to rn parsing

### LINKED ISSUE #2935: Float parser for JSON (+ne:dejs:format) fails when the input includes a +
**Describe the bug**
`+ne:dejs:format` doesn't parse floating values with a `'+'` 
(e.g.: `[%n p='8.660650844542463e+19']`).

**To Reproduce**
```
> (ne:dejs:format [%n p='8.660650844542463e+19'])
{1 19}
syntax error
```

pkova:
Allow the exponent token to be 'e' or 'E', allow the explicit sign of '+'.
Following the spec at https://json.org.

Fixes the bug Jose discovered in https://github.com/urbit/urbit/pull/1892#issuecomment-633428939.

Closes #2935.

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -5538,9 +5538,9 @@
         ==
         ;~  pose
           ;~  pfix
-            (just 'e')
+            (mask "eE")
             ;~  plug
-              ;~(pose (cold | hep) (easy &))
+              ;~(pose (cold | hep) (cold & lus) (easy &))
               ;~  pose
                 ;~(pfix (plus (just '0')) dim:ag)
                 dim:ag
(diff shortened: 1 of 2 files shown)


yosoyubik:
Thanks @pkova! :)

pkg/arvo/sys/zuse.hoon
```diff
@@ -5538,9 +5538,9 @@
         ==
         ;~  pose
           ;~  pfix
-            (just 'e')
+            ;~(pose (just 'e') (just 'E'))
```
Fang-, line 5541: `+just` is deprecated in favor of `+jest` I believe. But also, consider using `+mask`.

```suggestion
            (mask "eE")
```

pkova:
Done.

matildepark:
This is gonna be weird but: consider putting "Fixes #2935" in the commit itself for this / future patches; now that we have the rolling branch model, it leaves the issue in question open when this is merged, because it's not being merged to master. Even if its commits get merged into master later.

**But**, if you put "Fixes #2935" in the commit, it will auto-close the issue when the commit itself is in master, no matter when it hits master.

pkova:
And done.

ohAitch:
I don't think just is /deprecated/ per se, it's a hot path that parses a
single character without having to go into list prefix logic.

Issue #2688, opened 2020-04-07
Issue bunting types made with bucgar


liam-fitzgerald:
## Description
Bunting a type made with bucgar ($>) crashes if the bunt of the original type `q` doesn't nest underneath the new type.

## Reproduction
```
~zod:dojo> *$>(%zpzp hoon)
[%zpzp ~]

~zod:dojo> *$>(%cnhp hoon)
ford: %ride failed to execute:
```
Note that both types work fine for casts in the dojo. 
```
~zod:dojo> `$>(%zpzp hoon)`[%zpzp ~]
[%zpzp ~]

~zod:dojo>`$>(%cnhp hoon)`[%cnhp *hoon *hoon]
[%cnhp p=[%zpzp ~] q=[%zpzp ~]]
```
And also as samples and products to a gate iff it doesn't set off the above issue

```
~zod:dojo> =f |=  a=$>(%zpzp hoon)  -.a
~zod:dojo> (f [%zpzp ~])
%zpzp
~zod:dojo> (f [%cnhp *hoon *hoon])
nest-fail
~zod:dojo> =h |=  a=$>(%cnhp hoon)  -.a
ford: %ride failed to execute
~zod:dojo>=g |=  =hoon
                         ^-  (unit $>(%zpzp ^hoon))
                         ?:  ?=(%zpzp -.hoon)  
                            `hoon
                         ~
~zod:dojo>(g [%zpzp ~])
[~ [%zpzp ~]]
~zod:dojo>(g [%cnhp *hoon *hoon])
nest-fail
~zod:dojo> =x |=  =hoon
                         ^-  (unit $>(%cnhp ^hoon))
                         ?:  ?=(%cnhp -.hoon)
                           `hoon
                         ~
~zod:dojo> (x [%cnhp *hoon *hoon])
[~ [%cnhp p=[%zpzp ~] q=[%zpzp ~]]]

```

joemfb:
If you use `$>` or `$<`, you should just wrap it in a `$~` and set the bunt explicitly. There's no straightforward way to propagate `$>` constraints backward into the bunting logic since specs are macro-expanded before types are inferred.

ohAitch:
Huh I thought ++gain / ++lose were core type system primitives? Does this
just not apply to bunts?

joemfb:
[...]

PR #2937, opened 2020-05-26
groups refactor: one store to rule them all

liam-fitzgerald:
New and Improved Groups!
Headline features:
- Open (blacklist) and invite (whitelist) access control
- Admins and moderators
- Permissions now rolled into group-store
Blockers:
- Publish has not been updated to handle the new groups
- Frontend 
- State adapters
- groupify

Open questions:
- Groupify should now just be adding a contacts object to the pre-existing group?
- Should writers to unmanaged publish notebooks have a separate tag, or piggyback one of the privileged roles?
- How 2 state migrate? Some groups will have a group and permission to migrate from, others may just have a permission

cc: @Fang- @loganallenc can you check changes to links + chat respectively
@ixv


pkg/arvo/lib/group-store.hoon
```diff
@@ -0,0 +1,297 @@
...
+        ban-ships+(as ship)
+        replace+policy
+    ==
+  ++  group-id
+    %-  ot
+    :~  ship+ship
+        name+so
+    ==
+  ::
+  ++  remove-group
```
liam-fitzgerald, line 254: is there a nicer way to do this? `%remove-group` needs the `~` at the end to make sure the axes match so we can always access group-id when given a `action:group-store`

philipcmonk: One option is to transpose the `group-id` and the tag.  Thus, it's `$:  =group-id  $%  [%remove-group ~] 
 ...  ==`.  You still want the `~`, but some prefer this style since it's more explicit that every case needs a `group-id`.  This may violate local conventions, though, which would take precedence.

pkg/arvo/sur/group-store.hoon
```diff
@@ -1,19 +1,37 @@
...
-      [%bundle pax=path]                           :: create group at path
-      [%unbundle pax=path]                         :: delete group at path
-  ==
+::    %add-group: add a group
+::    %add-members: add members to a group
+::    %remove-members: remove members from a group
+::    %add-tag: add a tag to a set of ships, creating the tag if it doesn't exist
+::    %remove-tag:
+::      remove a tag from a set of ships. If the set is empty remove the tag
+::      from the group.
```
Fang-, line 11: What does it mean to "create" a tag? We don't track them explicitly as their own set of possibilities (only implicitly as the set of tags that are actively in use), so feels weird to talk about them that way.

pkg/arvo/lib/group.hoon
```diff
@@ -0,0 +1,98 @@
...
+  |=  [=path =ship]
+  %+  scry-for
+    ?
+  %+  welp
+    [%groups path]
+  /join/[(scot %p ship)]
+::
+++  can-join
+  |=  [=group-id =ship]
+  =-  ~&  -  -
```
Fang-, line 76:
```suggestion
```

pkg/arvo/lib/group.hoon
```diff
@@ -0,0 +1,98 @@
+/-  *group, *metadata-store
```
Fang-, line 1: +1 for most of the functions here, these are good and useful! Looks like they all pull data from the store though, which makes me wonder why they're not in `/lib/group-store`. You might still leave the call-with-data equivalent in here, if doing those operations on group nouns themselves turns out to be common.

Also could consider not exposing the `-path` functions at all, if you want the dev/user to work with `group-id`s as much as possible.

matildepark:
Might want to rebase this off of feat/spa so we can get rid of the unrelated commits here, heh.

pkg/arvo/lib/group.hoon
```diff
@@ -0,0 +1,98 @@
...
+      %gx
+      (scot %p our.bowl)
+      %contact-store
+      (scot %da now.bowl)
+      (snoc `^path`[%contacts path] %noun)
+    ==
+  ?~  contact
+    %.n
+  %.y
+++  is-managed
```
Fang-, line 93: These are kinda weird. Scry into contact-store? Check for existence?

I mean, I get what it's doing, but a comment would be very useful here. And we probably don't want to depend on contact-store for this, anyway? Did we not consider including a `hidden=?` flag in groups?

pkg/arvo/app/group-store.hoon
```diff
@@ -33,50 +69,59 @@
...
   ::
   ++  on-peek
     |=  =path
     ^-  (unit (unit cage))
     ?+  path  (on-peek:def path)
-      [%x *]  ``noun+!>((~(get by groups) t.path))
+        [%x %groups @ @ ~]
+      =/  group-id
+        (group-id:de-path:store t.t.path)
+      ?~  group-id  ~
```
Fang-, line 108: (I always trip myself up on this but) I think `~` is for "no answer", whereas `[~ ~]` is for "invalid request"/"doesn't exist", which I'm pretty sure what you intend to say here.

pkg/arvo/app/group-store.hoon
```diff
@@ -1,26 +1,62 @@
...
+::
+::
+::    ## Scry paths
+::
+::    /y/groups:
+::      A listing of the current groups
+::    /y/groups/[group-id]/tag-queries:
+::      A listing of the tag queries for a group
+::    /x/groups/[group-id]:
+::      The group itself
```
Fang-, line 16: No subsequent `%x` for "ships with tag"? (:

(The "tag query" phrasing feels pretty ugly to me (here and everywhere), too, so I'd probably do something like this:)

```hoon
/y/groups/[group-id]/tags        ::  get tags used in group
/x/groups/[group-id]/tags/[tag]  ::  get ships by tag
```

pkg/arvo/app/group-store.hoon
```diff
@@ -85,85 +130,322 @@
...
+    =.  members.group
+      (~(put in members.group) our.bol)
+    =.  policy.group   policy
+    =.  groups
+      (~(put by groups) group-id group)
+    :_  state
+    (send-diff %add-group group-id policy)
+  ::  +add-members: add members to group
+  ::
+  ::    no-op if group does not exist
```
Fang-, line 216: Shouldn't it be pretty loud about this though? Silent failures aren't very nice...

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,98 @@
...
+|%
+::  $group-id: unique identifier for a group
+::
++$  group-id  [=ship =term]
+::  $groups: a mapping from group-ids to groups
+::
++$  groups  (map group-id group)
+::  $tag: an identifier used to identify a subset of members
+::
+::   Tags may be used and recognised differently across apps.
```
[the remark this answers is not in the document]
matildepark: I think that was @loganallenc? Unless @liam-fitzgerald also wrote a proposal gist.

pkg/arvo/sur/group-hook.hoon
```diff
@@ -1,12 +1,26 @@
...
-                                     :: available for other ships to sync
-                                     :: if ship is foreign, delete any local
-                                     :: group at that path and mirror the
-                                     :: foreign group at our local path
-                                     ::
-      [%remove =path]                :: remove the path.
+::  $action: request to change group-hook state
+::
+::    %add:
+::      if ship is ours make group available to sync, else sync foreign group
```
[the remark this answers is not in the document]
loganallenc: It’s important to keep this pattern around such that work on federation, CRDTs, etc goes smoother when/if we implement it for this store. See Harrison’s chat work for an example of us moving towards those affordances.

pkg/arvo/sur/group.hoon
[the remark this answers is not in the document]
loganallenc: I’m fine with moving to `?(tag [app=term tag=term])`.

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,98 @@
...
+::
+::    These roles are
+::    %admin: Administrator, can do everything except delete the group
+::    %moderator: Moderator, can add/remove/ban users
+::    %janitor: Has no special meaning inside group-store,
+::    but may be given additional privileges in other apps.
+::    %member: Ordinary member, this tag is implied if the user is not in any
+::    of the other roles
++$  role-tag
+  ?(%admin %moderator %janitor %member)
```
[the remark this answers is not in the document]
loganallenc: Agreed

pkg/arvo/app/group-store.hoon
```diff
@@ -85,85 +130,322 @@
...
+  ^-  rank
+  =/  size
+    (met 3 ship)
+  ?:  (lte size 2)
+    %0
+  ?:  (lte size 4)
+    %1
+  ?:  (lte size 8)
+    %2
+  %3
```
ixv, line 144: You can use the standard lib function clan:title instead of this.

pkg/arvo/app/group-store.hoon
ixv: I agree with this, I think the only time you want to be silent about failures is if being loud gives away information you don't want to give away. But since only our ship can send this poke, we should go loud.

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,98 @@
...
+      =tag-queries
+      =policy
+  ==
+::  $rank: ship class by length
+::
+::    0: galaxy or star -- 2  bytes
+::    1: planet         -- 4  bytes
+::    2: moon           -- 8  bytes
+::    3: comet          -- 16 bytes
++$  rank  ?(%0 %1 %2 %3)
```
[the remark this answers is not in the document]
liam-fitzgerald: oh I nicked this from ames i wasn't aware of the existence of `rank:title`

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,98 @@
...
+::  $policy: access control for a group
+::
+++  policy
```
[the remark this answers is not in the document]
liam-fitzgerald: This was more so i could bunt the policies individually, but i guess it's not that hard to just write `[%open ~ ~]`

pkg/arvo/lib/group.hoon
liam-fitzgerald: We'd be duplicating this information across stores then, which is part of what caused the whole permissions/groups mess in the first place. I'm not 100% opposed to it, just think we should minimise duplication across stores where possible.

pkg/arvo/app/group-store.hoon
liam-fitzgerald: Considering nixing the `tag-query` scry tbh because we can just implement it as a scry helper in `lib/group`. Are subsequent scries in the same invocation cached? Maybe could help performance.

pkg/arvo/sur/group.hoon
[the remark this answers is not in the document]
Fang-: Hmm, I was gonna suggest you might be able to `*$>(%open policy)`, but that doesn't work. Not sure why, I'm guessing my intuition about `$>` doesn't quite line up with reality.

pkg/arvo/lib/group.hoon
Fang-: I think the mess there was that it was implicit in paths/identifiers, which was hard and unnecessary to track.

The root of that information would be what the _group_ was to be used for. You only ever care about this on the group level. A chat is just a chat, but it may be associated with a visible or hidden group. As such, you just have apps refer to the data here, and don't end up with duplication.

Does that sound about right?

pkg/arvo/sur/group-hook.hoon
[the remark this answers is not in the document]
Fang-: > It’s important to keep this pattern around

You mean the "explicit add for local" pattern, right?

We're doing a rewrite for correctness and reduction in complexity. If we want to account for, idk federated groups or something, we should probably do so explicitly rather than leave old and otherwise undesirable patterns standing.

pkg/arvo/sur/group.hoon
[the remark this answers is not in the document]
liam-fitzgerald: #2688 might be of interest

pkg/arvo/sur/group-hook.hoon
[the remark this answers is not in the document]
liam-fitzgerald: I'd be concerned about state migration if we split in group-hook in two. Additionally as groups doesn't have a `-view` agent, there are some cross-cutting concerns that complicate a bifurcation (invites etc.).

pkg/arvo/app/publish.hoon
```diff
@@ -955,14 +988,16 @@
...
-  .^(? %gx full-pax)
+  =/  =group-id
+    (need (group-id:de-path:group-store writers.book))
+  =/  role=(unit role-tag)
+    (role-for-ship:grup group-id who)
+  ?~  role
+    %.n
+  ?.  ?=(%write mod)
+    %.y
+  !=(%member u.role)
```
ixv, line 1000: A downside of doing this with role tags is that you can't have different notebooks in a group with different posting permissions.

pkg/arvo/sur/group-hook.hoon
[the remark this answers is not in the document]
loganallenc: This isn't an undesirable pattern. I will say that I think groups should be split into a `push-hook` and a `pull-hook` now akin to Harrison's work as opposed to waiting to do so until later.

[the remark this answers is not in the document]
loganallenc: Happy to discuss a specific strategy.

pkg/arvo/lib/group.hoon
loganallenc: There should be no need for `is-managed` now that we have tagged-queries for groups.

pkg/arvo/mar/group/initial.hoon
```diff
@@ -1,14 +1,13 @@
-/+  *group-json
```
loganallenc, line 1: This file is no longer necessary. Please incorporate `%initial` into the %group-update.

loganallenc, line 1: This file is no longer necessary. Please incorporate `%initial` into the %group-update.

loganallenc: Sorry for duplicating the comment! Weird internet

pkg/arvo/app/publish.hoon
liam-fitzgerald: revamped with per-notebook role tags

pkg/arvo/app/link-proxy-hook.hoon
```diff
@@ -209,25 +204,25 @@
...
   ^-  (quip card _state)
   :_  state
-  ?.  ?=(%remove -.upd)  ~
+  ?.  ?=(%remove-members -.upd)  ~
   ::  if someone was removed from a group, find all link resources associated
   ::  with that group, then kick their subscriptions if they're no longer
   ::
   %-  zing
-  %+  turn  (app-paths-from-group:md %link pax.upd)
+  %+  turn  (app-paths-from-group:md %link (group-id:en-path:group-store group-id.upd))
```
Fang-, line 220: Maybe we want to update `+app-paths-from-group:md` to take a `group-id` as argument instead?

pkg/arvo/app/link-view.hoon
```diff
@@ -205,19 +205,37 @@
...
+  =/  =group-id
+    (need (group-id:de-path:group-store path.invite.upd))
+  :~
+      ::  add self
+      :*  %pass
+          [%group-add path.invite.upd]
+          %agent  [ship.group-id %group-hook]
+          %poke  %group-action
+          !>  ^-  action:group-store
+          [%add-members group-id (sy our.bowl ~) ~]
```
Fang-, line 217: The `%add-members` here feels a little bit weird. Have you considered having a special action for "I am joining this group"?

pkg/arvo/app/link-view.hoon
```diff
@@ -278,50 +294,40 @@
     ==
   ?:  ?=(%group -.members)  ~
   ::  if the group is "real", make contact-view do the heavy lifting
-  ::
+  ::  TODO: verify
```
Fang-, line 297: Verified y/n?

pkg/arvo/app/link-view.hoon
```diff
@@ -278,50 +294,40 @@
...
+      (sham group-path eny.bowl)
+    :*  our.bowl
+        %group-hook
+        group-path
+        ship
+        title
+    ==
   ::  for "unmanaged" groups, do it ourselves
   ::
+
```
Fang-, line 323:
```suggestion
```

pkg/arvo/app/link-view.hoon
```diff
@@ -388,8 +396,8 @@
   ^-  (list card)
   :-  %^  do-poke  %group-store
         %group-action
-      !>  ^-  group-action:group-store
-      [%add ships group-path]
+      !>  ~ :: ^-  action:group-store
+      ::  [%add ships group-path]
```
Fang-, line 400: What's up with this?

pkg/arvo/sur/group-hook.hoon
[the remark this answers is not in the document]
Fang-: Is this happening y/n?

pkg/arvo/app/group-store.hoon
Fang-: This still remains, I think.

pkg/arvo/sur/group-store.hoon
```diff
@@ -1,20 +1,39 @@
...
-+$  group-update
-  $%  [%initial =groups]
-      [%keys keys=(set path)]                      :: keys have changed
-      [%path members=group pax=path]
-      group-action
++$  action
+  $%  [%add-group =group-id =policy hidden=?]
+      [%add-members =group-id ships=(set ship) tags=(set tag)]
+      [%remove-members =group-id ships=(set ship)]
+      [%add-tag =group-id =tag ships=(set ship)]
```
Fang-, line 20: Not sure how to feel about tags being included in the `%add-members` action. Especially with the "members can add themselves sometimes" from a previous comment, this seems to add a lot of complexity. Think it should strictly be in `%add-tag` regardless.

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,94 @@
...
++$  group-id  [=ship =term]
+::  $groups: a mapping from group-ids to groups
+::
++$  groups  (map group-id group)
+::  $group-tag: an identifier used by groups
+::
+::    These tags should have precise semantics, as they are shared across all
+::    apps.
+::
++$  group-tag  ?(role-tag)
```
Fang-, line 14: Looks like `group-tag` can just go for now, it doesn't really add anything on top of `role-tag`. Naming of the latter is more precise, too.

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,94 @@
...
+  |%
+  ::
+  +$  policy
+    $%  invite
+        open
+    ==
+  ::  $diff: change group policy
+  +$  diff
+    $%  diff:invite
+        diff:open
```
Fang-, line 59: Maybe you want these to be tagged with the policy tag also, to avoid diff tag conflicts. (Not the case _right now_ but, you know, structure sanity.)

```suggestion
    $%  [%invite diff:invite]
        [%open diff:open]
```

pkg/arvo/sur/group-hook.hoon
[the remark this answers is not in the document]
liam-fitzgerald: yeah sounds good

pkg/arvo/app/link-view.hoon
liam-fitzgerald: What would that give us over and above what we already get from `%add-members`?

pkg/arvo/sur/group.hoon
liam-fitzgerald: what if we want to add more tags later?

pkg/arvo/app/link-proxy-hook.hoon
liam-fitzgerald: should probably be done when links is changed to use identifiers tbh

pkg/arvo/app/link-view.hoon
Fang-: Cleaner separation of the flows/use cases for the action. Right now there's a bunch of implicit logic that comes with "add members", where you have to go "admins can do this, but also randos can do this, but only if they add themselves, and only if the policy allows it".

pkg/arvo/sur/group.hoon
Fang-: We move to `+$  group-tag  ?(role-tag new-tag)`, and the existing uses of `role-tag`s will still nest fine. The current `group-tag` is exactly identical to `role-tag` and so superfluous, is what I'm saying.

matildepark:
If this is still for after Fusion, ensure you target `release/after-fusion` and base it on the Fusion branch.

liam-fitzgerald:
should be ready for a final review. Note that this doesn't include the FE changes, which are in a seperate branch for git-related reasons.

liam-fitzgerald:
Can I get another pass @Fang- @ixv @loganallenc

pkg/arvo/lib/group-store.hoon
```diff
@@ -0,0 +1,455 @@
+/-  *group, sur=group-store
+/+  resource
+^?
+=<  [. sur]
+=,  sur
+|%
+::
+++  dekebab
```
Fang-, line 8: >different names between hoon and json

I get it, kebab case is a pain to work with in js, but...  
At least this isn't doing them all by hand. (^:

pkg/arvo/app/group-store.hoon
```diff
@@ -30,50 +67,201 @@
...
+        =/  group=^path
+          (en-path:resource rid)
+        [(spat group) ~]
+      ``noun+!>(arch)
+    ::
+        [%x %groups %ship @ @ ~]
+      =/  rid=(unit resource)
+        (de-path-soft:resource t.t.path)
+      ?~  rid   ~
+      ``noun+!>((peek-group u.rid))
```
Fang-, line 256: Maybe with the addition of #3001 it might be worth writing marks for these, instead of using `%noun` everywhere?

pkg/arvo/app/chat-store.hoon
```diff
@@ -9,18 +9,20 @@
...
 +$  state-two     [%2 =inbox:store]
 +$  state-three   [%3 =inbox:store]
++$  state-four    [%4 =inbox:store]
 +$  admin-action
   $%  [%trim ~]
   ==
 --
 ::
-=|  state-three
+=|  state-four
```
loganallenc, line 25: This needs a rebase.

pkg/arvo/app/contact-view.hoon
```diff
@@ -1,16 +1,21 @@
...
+    *permission-group-hook,
+    *permission-hook,
+    pull-hook,
+    push-hook
+/+  *server, *contact-json, default-agent, dbug, verb,
+    grpl=group, mdl=metadata, resource,
+    group-store
 ::
-/+  *server, *contact-json, default-agent, dbug
+
```
loganallenc, line 18: blank line

pkg/arvo/app/contact-view.hoon
```diff
@@ -40,9 +46,7 @@
     :_  this
     :~  [%pass /updates %agent [our.bowl %contact-store] %watch /updates]
         (contact-poke:cc [%create /~/default])
-        (group-poke:cc [%bundle /~/default])
         (contact-poke:cc [%add /~/default our.bowl *contact])
-        (group-poke:cc [%add [our.bowl ~ ~] /~/default])
```
loganallenc, line 45: Why are you removing the default group?

pkg/arvo/sur/group.hoon
```diff
@@ -0,0 +1,93 @@
...
+::   Tags may be used and recognised differently across apps.
+::   for example, you could use tags like `%author`, `%bot`, `%flagged`...
+::
++$  tag  $@(group-tag [app=term tag=term])
+::  $role-tag: a kind of $group-tag that identifies a privileged user
+::
+::    These roles are
+::    %admin: Administrator, can do everything except delete the group
+::    %moderator: Moderator, can add/remove/ban users
+::    %janitor: Has no special meaning inside group-store,
```
loganallenc, line 25: I think that roles should be %group-store specific, and tags should be used within apps for everything else. I think it's best to remove the %janitor option.

pkg/arvo/app/contact-view.hoon
liam-fitzgerald: What purpose does it serve now?

matildepark:
At present? If you set your root identity, it prefills your contact card to share.

However, I think @loganallenc wanted to leave the door open for the ad hoc sharing of contact information as well.

—
~haddef-sigwen
https://urbit.org

pkg/arvo/app/contact-view.hoon
loganallenc: It’s a placeholder for ad-hoc contact sharing. Perhaps better served by a TODO

pkg/arvo/sur/group.hoon
liam-fitzgerald: Roles won't really be group-store specific though. For example, you can imagine chat would want to restrict pinning a message to certain roles. Having a new tag for each permission on each app would lead to a combinatorial explosion of tags, most of which would probably be duplicating %group-store's roles anyway.

loganallenc:
Note: in the last commit you introduce a non-atomic operation with option chaining to add yourself to the group. We should avoid those whenever possible

matildepark:
- Merge conflict
- This can be amended to the `~ship/chat-name` format, though honestly I would probably just remove the 'join chat' screen UI entirely in favour of people joining public groups (@urcades?):

image

We would still need to keep its logic for the auto-join functionality but essentially the only unmanaged chats left should be DMs.

- Not a good group name placeholder:

image

The options button in Group Settings has no dark css:

image

(Is this using the 'dropdown' component? It should — tho you may need to move `dropdown.js` out of Publish and into `src/components`)

I'm gonna try the OTA case now ...

matildepark:
The OTA seemingly keeps hanging and crashing on hood. Anyone know what's going on?

```
gall: reloading %hood
crud: %into event failed
bail: exit
[%poke p=%into]
%leap
[%feck p=%veer]
%leap
[%feck p=%veer]
%leap
[%feck p=%veer]
%leap
wire=/sys/lyv
bar-stack=~[~[/d //term/1] ~[//sync/0v5.vijv5] ~[//sync/0v5.vijv5]]
take: failed
[%gall-take-failed /sys/lyv]
%load-failed
nest-fail
/app/hood/hoon:<[47 7].[47 39]>
/app/hood/hoon:<[47 3].[59 23]>
/app/hood/hoon:<[46 3].[59 23]>
```

urcades:
@matildepark I'm 100% for removing that "join" flow from interface existence

matildepark:
You can essentially restore shortcodes if all shortcodes refer to groups. 

However I can more easily imagine flows for unmanaged notebooks and whatever. Perhaps they’re just private, if unmanaged, and if you invite someone in, that sneaks an invite in for them, but there’s no join flow for that directly.

—
~haddef-sigwen
https://urbit.org

urcades:
Ah, I'm so very pro- shortcodes coming back into play — while they lasted, they really gave the system this peculiar sense of self-reference that I felt was powerful

loganallenc:
I have no preference other than that we reduce the number of flows and try to make everything internally consistent without making open groups unwieldy to join (not saying anything proposed necessarily would, just that it’s something to keep in mind).

matildepark:
I’m in total agreement. We need to pare down flows significantly, even if it means we eliminate ad hoc stuff. I think we just need to re-evaluate the Publish product and everything else would be okay?

—
~haddef-sigwen
https://urbit.org

urcades:
These both seem like fine assessments to me!

—
~fabled-faster
https://urbit.org/

custom-jonathan:
Yep, I've booted ~minder with ~roswet. I've left the group with ~minder and resent invitation from ~roswet. Whole cocktail of attempts.

And if the new groups clears it all up, I'm all good waiting it out. Everything else seems to be running smoothly, so I'm not in a bad spot. Good to know there's a plan!

matildepark:
DMs seem to be broken?

matildepark:
profile-overlay.js and others have lots of these conditional checks based on `/~/`:

```javascript
    const identityHref = group['group-path'].startsWith('/~/')
      ? '/~groups/me'
      : `/~groups/view${group['group-path']}/${window.ship}`;
```

In this case it links you to Groups where you can edit your group identity for an otherwise hidden group. It would be good to sweep for these checks.

matildepark:
Launch will also have to be amended — its welcome.js links you to 'join our lobby' at `~/~dopzod/urbit-help` when what we want is to link you to join the Urbit Community group directly.

timlucmiptev:
Was playing around a bit with this, and it seems in `push-hook.hoon` that `poke-update` is passing the card to a hard-coded `%group-store` rather than using `store-name.config`:
https://github.com/urbit/urbit/blob/64065d720c49c9065ee3deccf9f5a131c3ef6b80/pkg/arvo/lib/push-hook.hoon#L209

Disregard if that's intended behavior.

liam-fitzgerald:
> Was playing around a bit with this, and it seems in `push-hook.hoon` that `poke-update` is passing the card to a hard-coded `%group-store` rather than using `store-name.config`:
> 
> https://github.com/urbit/urbit/blob/64065d720c49c9065ee3deccf9f5a131c3ef6b80/pkg/arvo/lib/push-hook.hoon#L209
> 
> Disregard if that's intended behavior.

good catch!

matildepark:
`associations.contacts` retains old path:

image

matildepark:
:( More OTA problems. Tried on a `-L` version of haddef.

1. Direct Messages retain their sig-prepended paths in our `inbox`, but `associations.chat` migrates the path to the new path structure. All DMs don't load when you try to go to them.

Screen Shot 2020-07-21 at 2 15 13 PM
Screen Shot 2020-07-21 at 2 14 16 PM

2. All notebooks don't have their `subscribers-group-path` migrated to the new path structure. The Publish sidebar organises notebooks by `subscribers-group-path`, so all notebooks are considered 'unmanaged'.

image

matildepark:
What do old groups default to — private or public? Is there an easy way to change a group's privacy setting? (Asking bc we'll want UC to be public.)

matildepark:
Honestly, my bad for not reading — it looks like `%change-policy` in `sur/group-store` is what we want here.

PR #2938, opened 2020-05-26
vere: handle failed cue of private key file

lukechampine:
This is my first time touching vere -- hopefully I got it *mostly* right 😬

### DIFF
--- a/pkg/urbit/vere/daemon.c
+++ b/pkg/urbit/vere/daemon.c
@@ -612,7 +612,14 @@ _boothack_key(u3_noun kef)
 
     //  +seed:able:jael: private key file
     //
-    seed = u3ke_cue(u3k(u3t(des)));
+    u3_noun pro = u3m_soft(0, u3ke_cue, u3k(u3t(des)));
+    if ( u3_blip != u3h(pro) ) {
+      u3l_log("dawn: unable to cue private key\r\n");
+      exit(1);
+    }
+    seed = u3k(u3t(pro));
+    u3z(pro);
+
     //  local reference, not counted
     //
     ship = u3h(seed);


pkg/urbit/vere/daemon.c
```diff
@@ -612,7 +612,13 @@ _boothack_key(u3_noun kef)
 
     //  +seed:able:jael: private key file
     //
-    seed = u3ke_cue(u3k(u3t(des)));
+    seed = u3m_soft(0, u3ke_cue, u3k(u3t(des)));
+    if ( u3_blip != u3h(seed) ) {
+      u3l_log("dawn: unable to cue private key\r\n");
+      exit(1);
+    }
+    seed = u3t(seed);
```
joemfb, line 620: This logic looks right, but it's leaking the reference to the product of `u3m_soft()`. You need to gain a reference to the tail of the product of `u3m_soft()`, and lose a reference to the product itself. This is tedious, and tends to require lots of local variables (which has its own implications for code style ....).

Something like this:

```
{
  u3_noun pro = u3m_soft(...);

  if ( u3_blip != u3h(pro) ) {
    // fatal error, refcounts ignored
  }
  else {
    foo = u3k(u3t(pro));
    u3z(pro);
  }
}
```

lukechampine: thanks -- reassigning to `seed` was probably bad style anyway. Think I got it now.

philipcmonk:
I actually don't think `cue` crashes on invalid input (perhaps it should?), it just return `0` usually (always?).

philipcmonk:
Oh nvm I saw your comment that those jets supposedly do crash in those instances.

philipcmonk:
Actually, under what circumstances does `u3m_soft` produce an error when just running the code produces `0`?

joemfb:
`0` should generally never be an error return on a function over nouns (units maybe excepted, depending on what is meant by "error"). Some partial functions return `u3_weak`, with `u3_none` ("not-a-noun") as a sentinel value. 

`+cue` is partial due to back-references (see `(need (~(get by m) q.d))`), and again due to `+rub`, which is partial over atom lengths larger than the available input (see `?<  (gth c m)`). Those deterministic crashes are `%exit` bails in the runtime.

A call to `u3m_bail()` is equivalent to throwing an exception -- `return u3m_bail(c3__exit);` does not in fact return to the caller, but to the enclosing `u3m_soft()` or equivalent, via `_longjmp()`. The product of `u3m_soft()` is roughly `$%([%0 p=*] goof)`, where `$goof` is `[@term tang]`. If the function in a `u3m_soft()` call returns, the product will be `[%0 return-value]`; if the function bails, the product will be `[bail-mote stack-trace]`.

philipcmonk:
Okay, so what's confusing me is that `cue` considers a lot of things to be valid that I wouldn't.  Is there any way to say "if there's extra data, crash"?

```
> (cue eny)
7
> (cue eny)
/app/dojo/hoon:<[847 7].[852 22]>
/app/dojo/hoon:<[848 7].[852 22]>
/app/dojo/hoon:<[850 7].[852 22]>
/app/dojo/hoon:<[851 7].[852 22]>
/app/dojo/hoon:<[852 7].[852 22]>
> (cue eny)
0
> (cue eny)
0
> (cue eny)
0
> (cue eny)
[0 1 0]
> (cue eny)
0
> (cue eny)
3
> (cue eny)
/app/dojo/hoon:<[847 7].[852 22]>
/app/dojo/hoon:<[848 7].[852 22]>
/app/dojo/hoon:<[850 7].[852 22]>
/app/dojo/hoon:<[851 7].[852 22]>
/app/dojo/hoon:<[852 7].[852 22]>
```

ohAitch:
I think that would make the recursive "spec" definition somewhat more
complex, though may still be worthwhile

joemfb:
There's currently no constraint in `+cue` itself that it consume all the input. When cue-ing random data, you have a 50% chance of that data being read as an atom (ie, if the low bit is 0), and 25% chances for a cell (next bit 0) or backreference (next bit 1), respectively. A top-level backreference will always be invalid, while a cell will need to be followed by two bitwise-contiguous valid entries. Atoms are length-prefixed, so you'll only ever read up to the length specified, which can certainly be less then the size of the input. And given the nature of "atomic encoding", you may also be "reading" zeros off the end of an atom.

(Tangentially: @benjamin-tlon pointed out that if the length-prefix encoding were reversed (N 1s to encode the bit-width of the atom, instead of N 0s) and backreferences removed, cue would be total over all atoms.)

I could imagine a top-level assertion in `+cue` requiring that we've read at least the bit-width of the input out of the input. But I don't know that it's very important -- we should be clamming the output of +cue anywhere we require a particular structure, and any type-safety concerns should be enforced by the clam.

ohAitch:
You could also replace back references with "forward references" that say
the following noun should be considered spliced in n bits in the future, if
you want to keep the dag encoding space asymptotics - would be a mess to
encode tho, as compression schemes go.

Or just remove the "back reference must be to an aligned noun" constraint,
and consider the bitstream to also have inf /leading/ zeroes. Really tho
"cue never crashes" seems like overkill, at most we want "most cues are
distinct nouns" - a slight tweak we could do to the current paradigms is
consider the (nonzero parts of the) bitstream to be an improper list of cue
nouns, which would make appending to the event log slightly prettier.

ohAitch:
On the elegance side - if you swap the cell encoding to 1 1 (("here follows
one thing, and after it one other thing")) and back-references to 1 0, you
could *count* the first zero of either the 1 0 or simple atom 0 as part of
the length-length encoding. This would make encoding atom 0 cost one bit
more 0 1 -> 0 1 0, but encoding any other atom cost one bit less:
- 01 1 vs 0  01 1 for 1
- 001 0 01 vs 0 001 0 01 for 2
- 001 0 11 vs 0 001 0 11 for 3
- 001 1 001 vs 0 001 1 001 for 4
etc

In an optimization sense, I suppose this boils down to "are most atoms
jammed 0"; or approximately, "what percentage of data sent over the wire is
maps/sets"

PR #2939, opened 2020-05-26
SPA Cleanup: publish HTTP registration, store/subscription/reducer deduplication, mark cleanup

loganallenc:
- Cleaned up issue where Groups was being navigated to via an `a` link instead of a `Link`, initiating a new pageload.
- Cleaned up issue where the `publish` app was still in charge of sending `html-response` for the `/~publish` endpoint instead of the `landscape` app.
- Moved stores, reducers, subscriptions, and api objects into the top-level directory and modified them to extend `Base` objects that supply common methods and functionality. Moved the current global stores used for the header bar into this structure as well.
- Cleaned up the mark pattern where `initial` marks are different than `update` marks and added `%initial` updates to the tagged unions where appropriate.
- Namespaced `publish` JSON updates as `publish-update` to match how all the other apps' data formats and reducers work. For `publish` fetched JSON responses, namespaced behind `publish-response` tag.
- Remove notifications from `launch` for now since the new `launch` structure doesn't involve subscriptions to apps for data and the current unread numbers need to be reworked into a `notification-store` anyway.

Known issues:
- This was present in the previous version as well, but browsers only support 4 simultaneous event sources. If you navigate back and forth 4 times, Channel requests stop loading. Need to discuss how to fix this. Otherwise, works great!
- Links API file doesn't incorporate enough data-hiding.
- Dojo API / subscription have not been updated to follow the new pattern because they're... strange.


loganallenc:
Happy to. I'll resolve merge conflicts from this branch with feat/spa, merge, then rebase your branch and finish it. Thanks!

matildepark:
> Cleaned up issue where Groups was being navigated to via an a link instead of a Link, initiating a new pageload.

Oh, the conflict might be because I did some of that in b0a9ed041a2e0fec7df07354655e846edbcf1436.

PR #2940, opened 2020-05-26
publish: send less data on kick/resubscribe

ixv:
Now when you get kicked, check your local data to find the date of the last update you received and send that with the wire. The other ship, when receiving a watch, if it has a date, will only send back data from after that date.

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -362,7 +362,7 @@
     ?+    pax  (on-watch:def pax)
         [%http-response *]  [~ this]
     ::
-        [%notebook @ ~]
+        [%notebook @ *]
       =^  cards  state
         (watch-notebook:main pax)
       [cards this]
@@ -458,8 +458,13 @@
           [%subscribe @ @ ~]
         =/  who=@p     (slav %p i.t.wir)
         =/  book=@tas  i.t.t.wir
+        =/  wen=(unit @da)  (get-last-update:main who book)
+        =/  pax=path
+          ?~  wen
+            /notebook/[book]
+          /notebook/[book]/(scot %da u.wen)
         :_  this
-        [%pass wir %agent [who %publish] %watch /notebook/[book]]~
+        [%pass wir %agent [who %publish] %watch pax]~
       ::
           [%permissions ~]
         :_  this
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/publish.hoon
```diff
@@ -542,6 +547,63 @@
...
+  ^-  @da
+  =?  out  (gth last-edit.note wen)
+    last-edit.note
+  =.  out
+    %-  ~(rep by comments.note)
+    |=  [[@da =comment] out=_out]
+    ?:  (gth date-created.comment out)
+      date-created.comment
+    out
+  out
```
Fang-, line 568: You could just use `+max` instead of hand-rolling comparisons.

```suggestion
  %+  max  out
  %+  max  last-edit.note
  %-  ~(rep by comments.note)
  |=  [[@da =comment] out=_out]
  (max out date-created.comment)
```

pkg/arvo/app/publish.hoon
```diff
@@ -542,6 +547,63 @@
...
+    ?~  comments.not
+      out
+    (~(put by out) nom not)
+  ==
+::
+++  merge-notebooks
+  |=  [base=notebook diff=notebook]
+  ^-  notebook
+  %=  base
+      notes
```
Fang-, line 596: Don't we want to update things like title, description etc also? Those might have changed in our absence too.

So, maybe this should be `%=(diff ...)` instead.

pkg/arvo/app/publish.hoon
```diff
@@ -1963,7 +2042,11 @@
               date-created.data.del
           ==
       ==
+    =/  book  (~(get by books) host.del book.del)
     =^  cards  state
+      ?~  book
+        (emit-updates-and-state host.del book.del data.del del sty)
+      =.  data.del  (merge-notebooks u.book data.del)
       (emit-updates-and-state host.del book.del data.del del sty)
```
Fang-, line 2050: You might want `=?  data.del  (~(has by books) host.del book.del)` here instead. (Also, I generally prefer putting the `=/` at a more local "scope", closer to actual use.)

pkg/arvo/app/publish.hoon
ixv: Good catch. I also just noticed this won't handle the case of deleted notes.

pkg/arvo/app/publish.hoon
```diff
@@ -542,6 +547,59 @@
...
+    ?~  comments.not
+      out
+    (~(put by out) nom not)
+  ==
+::
+++  merge-notebooks
+  |=  [base=notebook diff=notebook]
+  ^-  notebook
+  %=  diff
+      notes
```
Fang-, line 592: ftr we discussed out of band that we're fine (for now) with not propagating deletion in these cases. Deletion in general will be picked up whenever the "archival" work happens.

Might still be nice to have a `::NOTE` here explaining that this is deliberate.

PR #2941, opened 2020-05-27
behn: fix bug with timers at same date

belisarius222:
Changes Behn's `.timers` data structure to maintain a queue for each date at which we have timers, preserving insertion order.

### DIFF
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -20,19 +20,14 @@
       ==
     ::
     +$  behn-state
-      $:  %1
-          timers=(tree [timer ~])
+      $:  %2
+          timers=(tree [key=@da val=(qeu duct)])
           unix-duct=duct
           next-wake=(unit @da)
           drips=drip-manager
       ==
     ::
-    ::  use lth instead of lte so that if same date, goes after
-    ::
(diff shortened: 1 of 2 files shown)


belisarius222:
`-time ~s2`, `+timers`, and `-ph-child-sync` work fine.  I'm not sure if any of our current pH tests exercise code that would send multiple timers at the same date, though.

pkg/arvo/sys/vane/behn.hoon
```diff
@@ -202,33 +196,70 @@
...
+    (emit-doze `date.u.first)
+  ::  +pop-timer: dequeue and produce earliest timer
+  ::
+  ++  pop-timer
+    ^+  [*timer timers.state]
+    =^  [date=@da dux=(qeu ^duct)]  timers.state  (pop:timer-map timers.state)
+    =^  dut  dux  ~(get to dux)
+    :-  [date dut]
+    ?:  =(~ dux)
+      +:(del:timer-map timers.state date)
```
philipcmonk, line 221: This is unnecessary and could be just `timers.state` because we updated it with `+pop`, right?

belisarius222: Oh, you're right.

PR #2948, opened 2020-05-28
hoon: update on-save

philipcmonk:
Whoops, all updates are relinking dojo/chat-cli.  Not an urgent problem.


philipcmonk:
You're correct, I'll r*base

PR #2950, opened 2020-05-28
ames: add |ames-wake

philipcmonk:
Somehow we ended up with flows which expected to awaken but did not wake
up.  This was likely caused by the error in r920j OTA, urbit-os-v1.0.18,
insufficiently mitigated in somam, urbit-os-v1.0.19.

This adds a command which ensures that every flow has an active timer.
I expect this to be needed only once, but it's a pretty general tool, so
it's worth keeping.

I've included an unused @t parameter to more easily add simple debug
commands to ames without having to add a new task

Includes #2948 because we don't want to keep reloading hood.  Officially this is considered a "backport" cherry-pick since that has already been merged into a release branch.

### DIFF
--- a/pkg/arvo/app/hood.hoon
+++ b/pkg/arvo/app/hood.hoon
@@ -140,7 +140,7 @@
   `..on-init
 ::
 ++  on-save
-  !>([%4 lac])
+  !>([%5 lac])
 ::
 ++  on-load
   |=  =old-state=vase
--- a/pkg/arvo/gen/hood/ames-wake.hoon
+++ b/pkg/arvo/gen/hood/ames-wake.hoon
@@ -1,0 +1,5 @@
+::  Set timers for any ames flows that lack them
+::
+:-  %say
+|=  [^ ~ ~]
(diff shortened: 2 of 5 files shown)


Fang-:
Do we not want to run this on-ames-upgrade to auto-fix for existing cases?

philipcmonk:
You can't send moves on vane upgrade, but I suppose we could do it in helm.

philipcmonk:
It's actually really hard to do safely in hood, this has been the cause of several of our OTA fiascos.  Using a new task in the same event it's added fails nondeterministically, depending on the order of app updates vs kernel updates.  If hood is updated before zuse/ames, then it crashes painfully.  I *think* we goad enough right now that it would work, but I don't think it's worth the risk.

Best to just say for the next week, "if you're worried people suddenly stopped receiving some of your messages around the release of `somam`, run `|ames-wake`".  This command is fast, idempotent, and it tells you if it did anything (`stirring` will be non-null).

Fang-:
Certainly not worth the risk then. Thanks for clarifying!

Issue #2952, opened 2020-05-28
Breach recovery issues


philipcmonk:
TL;DR if a new ship joined urbit-help (or any channel) before it finished with azimuth-tracker's initial update (about 10 minutes after boot), then it may have (50% chance) ended up in a broken state where it doesn't receive messages from urbit-help.  Workaround is to ;leave ~/~dopzod/urbit-help from chat-cli and then try to join again (possibly twice).

I'm writing this up because I want to make sure everyone understands how breaches are handled and their caveats.  Not that ignorance caused this issue — the two issues that came up are effectively bugs in my code.  But hopefully this will help people identify the sorts of problems that can arise during breach handling.

First, when a breach happens, Jael notifies its subscribers by turning over ~(tap in yen), which means they get notified in mug order (effectively random).  Consider what happens if Gall is notified before Ames.  Then an app hears a %kick, so it decides to resubscribe, so it sends a %watch via Ames, which enqueues that message and perhaps gives an effect to unix.  Then, before the event finishes, Ames hears about the breach, so it throws away all its state related to that ship — including the brand-new %watch.

Now, Gall thinks it's sent a %watch, so it has an unacknowledged outgoing subscription — but Ames knows nothing about it, and it'll never reach ~dopzod.  This is bad.  If Ames hears about the breach before Gall, then it works fine though.

You can resolve this by unsubscribing and resubscribing manually.  To solve it generally, I think the best solution is to send the notification first to Ames, then to Clay, then to Gall.  This way, gall apps can depend on the entire kernel knowing about the breach already.  Ames and Clay must be aware that later parts of the system don't know about the breach yet.

This is what happened with ~moltes-bitheb.  This is why it has many "breach peer" notifications early in its scrollback.  This suggests the second issue, which is that we shouldn't interpret the initial download of keys as evidence of a breach — we're just getting their keys for the first time, no need to throw anything away.  The code for deciding whether to notify a breach is small; you're welcome to identify the bug as an exercise:

```
      =/  maybe-point  (~(get by pos.zim) who)
      =/  =point  (fall maybe-point *point)
      ::  if changing rift upward, then signal a breach
      ::
      =?    ..feel
          ?&  ?=(%rift -.a-diff)
              (gth to.a-diff rift.point)
          ==
        %+  public-keys-give
          (subscribers-on-ship who)
        [%breach who]
      ...
```

[...]

PR #2953, opened 2020-05-28
jael: Breach-related fixes

philipcmonk:
Still testing, but seems correct so far.  See #2952 for details.  Also copying the description of the first commit:

We inspect the wire of our subscriber to see if we need to produce the
result as a %public-keys or a %boon.  This is bad -- we should proxy the
subscription to avoid this need, but this doesn't make that change yet.

%pubs is an old name that doesn't exist anymore (last existed around
September 2019).  The new version is /public-keys, but it's worked so
far because /public-keys has only one item in the path, so it missed the
conditional.  This commit makes the intent more clear.

The [%a @ @ *] could be just [%a @ *], but I leave it to reduce the
chance of breaking stuff.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1432,6 +1432,9 @@
     ::
     ?:  already-pending
       event-core
+    ::  NB: we specifically look for this wire in +public-keys-give in
+    ::  Jael.  if you change it here, you must change it there.
+    ::
     (emit duct %pass /public-keys %j %public-keys [n=ship ~ ~])
   ::  +send-blob: fire packet at .ship and maybe sponsors
   ::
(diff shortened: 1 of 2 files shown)


philipcmonk:
These passed -ph-breach-sync, -ph-breach-multiple, and -ph-breach-sudden, so I'm fairly confident there's no regression.  This does not actually fix the problem of ignoring historical breaches, since the initial update comes as a long series of diffs, applied one at a time, so we do in fact have an entry in `pos.zim` by the time we hear about the breach.  However, I started several moons with a pill from this and joined urbit-help before hearing its keys, and they all got the backlog correctly, so the bug we were originally trying to fix is in fact fixed.  I tried this on master and verified it was broken.

This is an improvement, so I think it's worth deploying as-is.  Ignoring historical breaches will have to wait for another day.

PR #2954, opened 2020-05-29
behn: fix +unset-timer

philipcmonk:
It was throwing away timers sometimes if you cancelled a timer at the
same date as other timers.


philipcmonk:
I pushed this to ~dopzod a while ago and it seems to be working.  I intend to deploy it soon.

Issue #2677, opened 2020-04-04
compacting loom on boot -> bail meme repeatedly


matildepark:
What good timing!

```
matilde@matilde:~$ ./urbit haddef-sigwen/
~
urbit 0.10.4
boot: home is /home/matilde/haddef-sigwen
loom: mapped 2048MB
lite: arvo formula 56587b84
lite: core 715fdde3
lite: final state 715fdde3
loom: mapped 2048MB
boot: protected loom
live: loaded: GB/1.602.945.024
boot: installed 262 jets

work: contiguous free space: MB/544.569.464
work: measuring memory:
work: play 5352376
---------------- playback starting----------------
pier: replaying events 5352376 through 5352389
  kernel: GB/1.224.866.996
total arvo stuff: GB/1.224.866.996
  warm jet state: KB/96.900
  cold jet state: MB/149.833.456
  hank cache: B/372
  battery hash cache: B/288
  call site cache: B/100
  hot jet state: KB/127.744
total jet stuff: MB/150.058.860
  bytecode programs: MB/1.727.192
  bytecode cache: KB/212.432
total nock stuff: MB/1.939.624
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
total marked: GB/1.376.865.804
free lists: MB/226.046.960
sweep: GB/1.376.865.804

.............work: measuring memory:
.
---------------- playback complete----------------
ames: live on 0
  kernel: GB/1.231.381.740
total arvo stuff: GB/1.231.381.740
  warm jet state: KB/96.900
  cold jet state: MB/149.833.456
  hank cache: B/288
  battery hash cache: B/288
  hot jet state: KB/127.744
total jet stuff: MB/150.058.676
  bytecode cache: B/288
total nock stuff: B/288
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
total marked: GB/1.381.441.028
free lists: MB/228.918.312
sweep: GB/1.381.441.028

work: compacting loom
bail: meme
bailing out
pier: work error: end of file
pier: work exit: status 0, signal 6
```

joemfb:
I wish I knew what was using so much memory in your kernel instance. Unfortunately, we don't yet have tools to get past this state. ("compacting loom" is `|pack`, which is the best we have. We need a version that's read-only wrt the current loom.)

I'm afraid a personal breach is your only option at this time.

matildepark:
[...]

PR #2957, opened 2020-05-29
serf: disable automatic |pack

joemfb:
`|pack` (added in #1984) only works up to some threshold of memory usage (as currently implemented), so it's unsafe to run automatically. Restoring this kind of automatic memory compaction remains a goal, but it'll require more significant development efforts.

Examples of this problem abound -- #2677 is representative.


ohAitch:
Generational garbage collection is coming in fits and starts

PR #2958, opened 2020-05-29
spa: fix bugs in chat and groups, fix metadata reducer

loganallenc:
All actions in the following have been tested:
- Chat has been tested to work
- Groups have been tested to work

Current State:
- Haven't tested Links
- Publish doesn't work
- Launch works but the weather / clock tile APIs are not hooked up properly

Note: This concludes my JS work on the SPA for a bit. Moving my attention towards fixing the HTTP redirect bug in Eyre and then `linear-store` and `graph-store` afterwards.

### DIFF
--- a/pkg/interface/src/api/chat.js
+++ b/pkg/interface/src/api/chat.js
@@ -1,4 +1,5 @@
 import BaseApi from './base';
+import { uuid } from '../lib/util';
 
 export default class ChatApi {
   constructor(ship, channel, store) {
--- a/pkg/interface/src/api/groups.js
+++ b/pkg/interface/src/api/groups.js
@@ -21,7 +21,7 @@ export default class GroupsApi {
 
     this.group = {
       add: helper.groupAdd.bind(helper),
-      delete: helper.groupRemove.bind(helper)
+      remove: helper.groupRemove.bind(helper)
     };
 
     this.invite = {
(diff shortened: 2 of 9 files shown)


matildepark:
Some weird stuff going on with this PR Logan ... it seems to have a diff commit history than `feat/spa` now that you force pushed `feat/spa`. Many conflicts also

loganallenc:
Will fix, won’t be too hard.

matildepark:
ping @loganallenc

matildepark:
This is blocking two different lines of work, @loganallenc — can you please rebase this PR?

loganallenc:
on it

loganallenc:
It's good to go now. Thanks.

matildepark:
It turns out that the changes in fec5646229d371e11ec648f22d29182bfd7d6836 are why everything is broken. 

The metadata-reducer in every application keys by `app-path` so we can get the metadata for each individual item, and that commit reduces it down to `group-path` and then overwrites the path with the latest `app-path` item's metadata. Thus Chat does not work either. I would revert this PR but it would retain the bugs in question here.

@loganallenc I have to ask you to amend the changes you made to the metadata-reducer to rekey it to `app-path`s; it's broken the entire suite of applications and should not have been pushed.

PR #2959, opened 2020-05-30
eyre: augmented cookie handling

### LINKED ISSUE #1617: Landscape silently dies when auth expires
As per #1612 cookies expire fast/often. If this happens while you still have a Landscape window open, it will stop functioning, without telling you. Functionality can only be restored by refreshing and logging in.

That is, Chat won't load new messages or submit whatever you're sending. In Publish, posting notes or comments will just silently hang forever.

(This has overlap with that Asana task about Landscape not giving you error messages whenever anything fails.)

Fang-:
Primarily:

- make sure we actually clear expired cookies from state, instead of letting them accrue indefinitely
- extend cookie/session duration on-use in regular HTTP requests; fixes #1617 by virtue of channel acks being PUT requests

See commit messages for more commentary.

Note that this doesn't _explicitly_ fix the symptom mentioned in #1617 where channels stay open even after a session expires. But considering the "extend on ack" nature of the implementation here though, I don't think there's any need to track a `(map session-id channel-id)` for explicit cleanup or anything. RFC though, maybe there's security implications there we want to take care of.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -84,7 +84,7 @@
 ++  axle
   $:  ::  date: date at which http-server's state was updated to this data structure
       ::
-      date=%~2019.10.6
+      date=%~2020.5.29
       ::  server-state: state of inbound requests
       ::
       =server-state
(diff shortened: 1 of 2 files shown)


Fang-:
Just realized the CI failures for this are, in fact, not spurious. Will update the tests to account for the setting of expiry timers and aggressive including of fresh cookies.

PR #2960, opened 2020-05-30
publish: update groupify UI

ejdiv:
this UI confused me because it always says "create group" even when you try to "add to group". this PR is for making the string update to "add to group" when a group is selected.

This also fixes the header UI to match others on this screen.

if the different header sizes were intentional, let me know and i can remove that change!


matildepark:
Dope!

Some comments tho:
1. Can you target `feat/spa`? You'll just make the changes [here](https://github.com/urbit/urbit/blob/feat/spa/pkg/interface/src/apps/publish/components/lib/settings.js) instead.
2. Please don't commit compiled JS.

ejdiv:
thanks for the review, @matildepark 

i apparently totally forgot how to use git and screwed up this PR beyond repair - #2968 is the new PR with just the JS changes in the new file location.

btw, when you say "don't commit compiled JS", do you mean not to run `gulp bundle-prod` as [mentioned here?](https://github.com/urbit/urbit/blob/master/pkg/interface/CONTRIBUTING.md)

PR #2966, opened 2020-06-02
Groupify tweak 2

ejdiv:


matildepark:
You probably need to refresh your copy of feat/spa, it was force pushed today. Try checking it out and `git reset --hard origin/feat/spa` and THEN originate your branch from it

ejdiv:
i see, thanks! i actually made a new PR at #2968  that seems to work. sorry for the spam

PR #2969, opened 2020-06-02
eyre: 303 redirect on successful login

liam-fitzgerald:
Changes the HTTP status code of the redirect that occurs upon a
successful login from 307 to 303. 307 preserves the method of the
original request, so the redirected request is a POST. With the new SPA,
this causes a 404 as app/file-server validates the method of the
request, something that did not happen in earlier versions of landscape.
303 instead changes the method to always produce a GET request.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:5a22f2797c2cbbff171cb0a4c14afaca4353f41f49f1fcdc5f5dbf3e8056a7e3
-size 13813374
+oid sha256:50821a0866e2ed901404a5c7bba9c4e2acb14a9658e32fd4719a642df77ff5e7
+size 13813430
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -979,7 +979,7 @@
       =/  actual-redirect  ?:(=(u.redirect '') '/' u.redirect)
       %-  handle-response
       :*  %start
-          :-  status-code=307
+          :-  status-code=303
           ^=  headers
             :~  ['location' actual-redirect]
                 ['set-cookie' cookie-line]
--- a/pkg/arvo/tests/sys/vane/eyre.hoon
+++ b/pkg/arvo/tests/sys/vane/eyre.hoon
@@ -613,12 +613,12 @@
             ^-  (hypo sign:http-server-gate)  :-  *type
             :*  %g  %unto  %fact
                 %http-response-header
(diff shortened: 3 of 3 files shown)


matildepark:
We're going to need this in `feat/spa` which might make that branch a `next-sys` release itself ... @loganallenc

liam-fitzgerald:
We can just comment out the method validation in app/file-server for now and then remove that when this OTAs?

loganallenc:
I’m okay with making `feat/spa` dependent on `next-sys` if y’all are... otherwise that sounds fine

philipcmonk:
If you can avoid changing sys, then the OTA is more likely to go smoothly and faster, since they'll only have to compile the apps that changed, and only once.  A change like this is pretty simple, so I wouldn't expect it to fail, it'll just take longer to process.  However, there are other things in release/next-sys that are more risky, and you don't want to unnecessarily make feat/spa depend on that.

It's reasonable to put in a shim so you don't have to depend on changes to sys, but if you feel it's better to make the change, then what I would do is PR the commit to *both* feat/spa and release/next-sys.  This way whichever one gets released first will make the change, and the other one won't have to.

You don't want to pull in unnecessary changes, so the cleanest thing to do is r*base this commit on top of `git merge-base origin/feat/spa origin/release/next-sys`, which is the most recent ancestor of both branches (v1.0.19 right now).  That way it's the exact same commit going into both branches.

Does that make sense?  This kind of situation frequently causes git-related headaches if not properly thought through, so I'm happy to elaborate on any of it.

matildepark:
@liam-fitzgerald I agree with @philipcmonk's course of action here. This has a conflict with the pill; if you can rebase, create the pill and then PR this into both `next/sys` and `feat/spa` branches we can remedy this one error.

liam-fitzgerald:
i've rebased on top of `git merge-base origin/feat/spa origin/release/next-sys` but i've still got a pill conflict?

philipcmonk:
The rebase looks correct, thanks.

There will be a merge conflict in the pill since you can't have a pill that will satisfy both sides.  The correct way to resolve this is by performing the merge at the command line, with something like these commands:

```
git checkout release/next-sys
git merge origin/lf/get-eyre-redirect
git checkout --ours :/bin/solid.pill
(make a new pill and copy it into bin/solid.pill)
git add :/bin/solid.pill
git commit
git push
```

You can use `sh/merge-with-custom-msg origin/lf/get-eyre-redirect 2969` in place of the `git merge` for a nicer commit message.  And then do the same thing for feat/spa.

ohAitch:
> `(make a new pill and copy it into bin/solid.pill)`

don't suppose we have a script for this?

joemfb:
Yes, `sh/update-solid-pill`.

Issue #2649, opened 2020-04-01
Feature Request: logout all sessions


custom-jonathan:
The ability to close all other landscape sessions connected to my ship.

matildepark:
As of #2970 this is just a front end task to surface the functionality. Adding to our slate ...

PR #2970, opened 2020-06-02
eyre: provide logout endpoint

Fang-:
Set up, by default, on /~/logout.

Sending a POST request to this expires the current session and redirects to the login page. If the "all" key is set in the request body, expires all open sessions.

This provides the backend for #2649. @urcades might want to think about integrating these actions into the frontend. For now, I just added a pair of buttons to the debug dashboard.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1012,7 +1012,7 @@
       ::
       =/  response=$>(%start http-event:http)
         :*  %start
-            response-header=[307 ['location' '/~/login']~]
+            response-header=[303 ['location' '/~/login']~]
             data=~
             complete=%.y
         ==
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -2445,7 +2445,7 @@
       %~2019.10.6
     =^  success  bindings.server-state.old
       %+  insert-binding
-        [[~ /~/logout] ~ [%logout ~]]
+        [[~ /~/logout] [/e/load/logout]~ [%logout ~]]
       bindings.server-state.old
     ~?  !success  [%e %failed-to-setup-logout-endpoint]
     %_  $
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -835,6 +835,9 @@
         %authentication
       (handle-request:authentication secure address request)
     ::
+        %logout
+      (handle-logout:authentication authenticated request)
+    ::
         %channel
       (handle-request:by-channel secure authenticated address request)
     ::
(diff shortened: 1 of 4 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2404,12 +2443,16 @@
     %~2020.5.29  ..^$(ax old)
   ::
       %~2019.10.6
+    =^  success  bindings.server-state.old
+      %+  insert-binding
+        [[~ /~/logout] ~ [%logout ~]]
```
Fang-, line 2448: @eglaysher not entirely sure how bad it is to use the `~` duct here, but didn't want to just make one up either. In practice only Eyre itself can undo these bindings anyway, so this shouldn't be the end of the world, right?

eglaysher: That's probably correct, but having null ducts around seems like the sort of thing which eventually causes chaos...

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1000,6 +1003,38 @@
...
+    ::  +handle-logout: handles an http request for logging out
+    ::
+    ++  handle-logout
+      |=  [authenticated=? =request:http]
+      ^-  [(list move) server-state]
+      ::  whatever we end up doing, we always redirect to the login page
+      ::
+      =/  response=$>(%start http-event:http)
+        :*  %start
+            response-header=[307 ['location' '/~/login']~]
```
eglaysher, line 1015: There was a patch this morning about 303 vs 307. (I don't actually know which one is better.)

pkg/arvo/sys/vane/eyre.hoon
philipcmonk: A null duct is definitely risky and weird.  If you ever pass on that duct and then someone gives a response, that'll go straight to unix instead of eyre, giving strange errors.  If you give on that duct... undefined behavior?  Worst of all, if someone stumbles on an empty duct in the wild, you can't exactly grep for `~`.

I'd hand-create a greppable duct for this purpose.  Something like ~[/logout] should be fine.  Still don't give on that duct or it'll go to unix, but at least if you pass on it it'll come back correctly, and it's greppable.

Fang-: The only place the binding ducts are used is in verifying binding "ownership", so that only the original creator of it can delete it. This seems.. potentially fragile, but that's a discussion for another time. I'll put a fake duct there for now, thanks all!

pkg/arvo/sys/vane/eyre.hoon
Fang-: Oops, good catch. 307 preserves the request type, whereas 303 becomes a GET request to the new location. Here, too, that's more appropriate.

PR #2971, opened 2020-06-04
eyre: 303 redirect on successful login

liam-fitzgerald:
Changes the HTTP status code of the redirect that occurs upon a
successful login from 307 to 303. 307 preserves the method of the
original request, so the redirected request is a POST. With the new SPA,
this causes a 404 as app/file-server validates the method of the
request, something that did not happen in earlier versions of landscape.
303 instead changes the method to always produce a GET request.

(pill is conficting halp @philipcmonk )


liam-fitzgerald:
self morging because #2969 was approved

PR #2972, opened 2020-06-04
zuse: add subset arm to ordered-map

tacryt-socryp:
Added an exclusive range subset arm to ordered-map that runs in ~ O(2*log n ) time. It passes all unit tests. Also added a get arm that I took from behn.


pkg/arvo/sys/zuse.hoon
```diff
@@ -7950,6 +7950,60 @@
...
+        ?:  =(key.n.a u.c)
+          (nip a(l ~))
+        ::  traverse to find key
+        ?:  (compare key.n.a u.c)
+          ::  found key to the left of start
+          (del-span (nip a(l ~)) b c)
+        ::  found key to the right of start
+        =/  lef  $(a l.a)
+        =.  l.a  lef
+        a
```
philipcmonk, line 8000: Feels a little strange to have two ways of recursing, would probably be cleaner to just choose one.  My preference would be the `$` recursion.

Also I feel these few lines would be cleaner as `a(l $(a l.a))` or at least `=.  l.a  $(a l.a)  a`

pkg/arvo/sys/zuse.hoon
```diff
@@ -7950,6 +7950,60 @@
     ?:  (compare key.n.a key.n.b)
       $(l.b $(b l.b, r.a ~), a r.a)
     $(r.b $(b r.b, l.a ~), a l.a)
+  ::
+  ::  +subset: take a range excluding start and/or end and all elements
+  ::  outside the range
```
philipcmonk, line 7964: Are we sure exclusive is most useful here?  Are there cases where we'd want inclusive?  Would be pretty easy to do I think, just don't `nip` in the `=` case.

pkg/arvo/sys/zuse.hoon
```diff
@@ -7950,6 +7950,60 @@
...
+    ++  del-span
+      |=  [a=(tree item) b=?(%start %end) c=(unit key)]
+      ^-  (tree item)
+      ?~  a  a
+      ?~  c  a
+      ?-  b
+          %start
+        ::  found key
+        ?:  =(key.n.a u.c)
+          (nip a(l ~))
```
philipcmonk, line 7992: This is fragile if we allow different keys to compare the same way.  You would have to recurse as well: `$(a (nip a(l ~))`

loganallenc:
@philipcmonk I've gotten this as far as I can without pairing with someone. All tests pass when ordered-map is moved to a library, but it mysteriously fails to compile %ames and %behn when the subset arm is added to ordered-map in zuse.

pkg/arvo/sys/zuse.hoon
loganallenc: I'll clean up the recursion and look at your other comments.

pkg/arvo/sys/zuse.hoon
loganallenc: I'm not sure I understand why we would allow different keys to compare the same way. That seems... bad?

pkg/arvo/sys/zuse.hoon
loganallenc: Done.

pkg/arvo/sys/zuse.hoon
loganallenc: Exclusive is most useful in the case I'm wanting to use it in, in which I keep track of the last event I've received, and ask for all events after that one. I'm not sure if it's always more useful.

loganallenc:
Ready for merge whenever!

loganallenc:
@belisarius222 Can I get a review?

tacryt-socryp:
Thanks @belisarius222! Made the changes you requested. Since you're on vacation, I'll dismiss your review. @philipcmonk This is ready for merge.

PR #2976, opened 2020-06-04
eyre: expire channels with their sessions

Fang-:
Associates channels with the authentication sessions that opened them, and deletes the channel when its associated session expires.

This should fix the weird behavior (and potential security issue) described in https://github.com/urbit/urbit/issues/1617#issuecomment-612839162, and allow the frontend to more immediately report on loss of connectivity. (Of course, #2959 should make that much less likely in the first place.)


Fang-:
Last remaining CI failure seems unrelated. (It ran for 50 minutes, wow.) Merging.

PR #2978, opened 2020-06-05
gall: properly handle empty outstanding ack queue

### LINKED ISSUE #2977: gall: Outstanding ack queue empty
A ship was regularly getting `arvo: replacement event (for %hear) failed` every 30 seconds, and it turned out to be an unusual crash in gall on line 629: https://pastebin.com/Mc3mJnMi

https://github.com/urbit/urbit/blob/79bbaec5e8b1d745136dd21ab9db77188eecec4a/pkg/arvo/sys/vane/gall.hoon#L622-L630

`outstanding` is a queue of the type of acknowledgment we expect to receive based on what type of message we sent.  We add the queue when we send a message and dequeue when we receive the ack.  We received an ack, but there was nothing in the queue.

When adding this queue it was initially empty, so we have a fallback `%missing` case, which doesn't crash and just sends both types of acknowledgment.  In the most recent OTA, we had to clear `outstanding` (see #2872 and #2873), which means we expected to hit this case again.

However, we only add `%missing` if the queue for this wire/duct doesn't exist, not if the queue is empty.  I'm guessing that there were two outstanding messages at the moment of the upgrade (it probably happened to have an outstanding ping at the time of the OTA, and when the OTA was applied it reloaded app/ping, which sends a second one).

You can see from that code above that the first ack would correctly trigger the `%missing` case, but it would add an empty queue at the wire/duct, so the second ack would crash.  The solution is to also use the `%missing` case if the queue exists but is empty.

Impact is probably relatively minimal other than occasional crashes, but it wouldn't surprise me if this caused a few flows to get stuck.

philipcmonk:
Fixes #2977, see there for rationale

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:58deed8e9b8cd2c84d092cb8f638b9881cb0d12b97f7c719339e3604c9e9d1d2
-size 13826033
+oid sha256:503a9a3cfbbe70866b89eee752253627f6799d12cb872f0ca3c3b8fc3b024837
+size 13825839
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -620,12 +620,14 @@
         ?~  t.t.t.wire
           =/  full-wire  sys+wire
           =/  stand
-            %+  ~(gut by outstanding.agents.state)  [full-wire hen]
-            ::  default is do nothing; should only hit if cleared queue
-            ::  in +load 3-to-4
-            ::
+            (~(gut by outstanding.agents.state) [full-wire hen] ~)
+          ::
+          ::  default is to send both ack types; should only hit if
+          ::  cleared queue in +load 3-to-4 or +load-4-to-5
+          ::
+          =?  stand  ?=(~ stand)
             (~(put to *(qeu remote-request)) %missing)
-          ~|  [full-wire=full-wire hen=hen stand=stand outs=outstanding.agents.state]
+          ~|  [full-wire=full-wire hen=hen stand=stand]
           =^  rr  stand  ~(get to stand)
           [rr (~(put by outstanding.agents.state) [full-wire hen] stand)]
         ::  non-null case of wire is old, remove on next breach after


philipcmonk:
Based on more stuff I'm seeing on the network, I actually think this would be really good to get out.  This causes crashes which are handled very slowly (a couple seconds, because the trace is huge, even though it doesn't get printed), and I wouldn't be surprised if it causes some ames flakiness.

Not sure if it makes sense to push this out as a hotfix.

belisarius222:
If we've only seen this on one ship, I'd say we should wait until the next OTA to push this out, but if we've seen it hit many ships, it could be worth a hotfix.

philipcmonk:
I found it on ~siddef today by accident while investigating another bug and
that's the only reason I noticed it on ~hiddev-dannut. I expect many cases
of "arvo replacement event" are this. Obviously not all, but we need to be
in the habit of occasionally checking the stack traces for those with |verb.

sipwex-barfus:
Hey guys it looks like I might be seeing this myself however I can't be 100% sure. I was having Hoon School notebook troubles about a month ago and turned my urbit off for that time as a result due to getting busy with some other things. Upon booting a few days ago I realized I was on a much older hash 

```
  [%base-hash 0vr.o87oq.g9lsl.l4ced.nau1s.uhn6h.e6la8.bbd8e.koaa3.u4tro.7ca6n]
  [%home-hash 0vr.o87oq.g9lsl.l4ced.nau1s.uhn6h.e6la8.bbd8e.koaa3.u4tro.7ca6n]
  [%sponsor ship=~litpub point='1446' life=[~ 3] rift=[~ 0]]
  [%dopzod ship=~dopzod point='4608' life=[~ 2] rift=[~ 1]]
```

I then tried the following since even after leaving it on for quite a while I wasn't getting any updates:
`|unsync %base (sein:title our now our) %kids` and `|merge %base ~marzod %kids` to force the OTA update which I heard was out along with updating my binary from 0.10.3 to 0.10.5 

```[ [%base-hash 0vr.v6l8h.ff979.hpr0m.uea2r.02fkf.26us8.or5j9.rgr57.6bgbi.9rbpn]
  [%home-hash 0vr.v6l8h.ff979.hpr0m.uea2r.02fkf.26us8.or5j9.rgr57.6bgbi.9rbpn]
  [%our ship=~sipwex-barfus point='2865038758' life=[~ 3] rift=[~ 2]]
  [%sponsor ship=~litpub point='1446' life=[~ 3] rift=[~ 0]]
  [%dopzod ship=~dopzod point='4608' life=[~ 2] rift=[~ 1]]
  "Compare lifes and rifts to values here:"
  "https://etherscan.io/address/azimuth.eth#readContract"
  "  life - getKeyRevisionNumber"
  "  rift - getContinuityNumber"
  ~
]
```

Before these upgrades(both OTA and binary) I was able to load up landscape and even message on a few chats however the vast majority of the chats had 0 messages which I found strange as they were previously very active before I shut off my urbit. Even after having landscape up for a period of time no chat would load up any other message. Additionally the biggest issue now is that the CLI incessantly displays

```arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
; ~todset-partug is ok
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
arvo: replacement event (for %hear) failed
; ~laptun-formul not responding still trying
arvo: replacement event (for %hear) failed````

and it takes minutes to load up landscape or to open any page on landscape. The CLI is essentially frozen and it is more or less impossible to even type a command in it unless I do it right on bootup. Somewhat at a loss but if this hotfix fixes this that would be amazing.

philipcmonk:
Hmm, you could load the changes directly into your ship.  I would do basically this:

- Start your ship with the `-L` flag, which will start it with Ames networking turned off, so you'll stop getting those crashes (but won't be able to talk to other ships)
- `|mount %` and clone this repo
- `git checkout urbit-os-v1.0.21`
- `cp urbit/pkg/arvo/* pier/home/`
- `|commit %home`

Then it'll apply the latest OTA and those errors should mostly go away.  At that point the trick will be making sure later OTAs don't fail.  I'd do this:

- `|cancel` until `cue-length=0`
- `|merge %base our %home, =gem %that`
- `|merge %base (sein:title our now our) %kids, =gem %this`

This should get you ready to receive future OTAs, so just `|unsync` wherever you're synced for them, and then `|sync` again.

Don't forget to run start your ship without `-L` so you can talk to ships again.

sipwex-barfus:
Appreciate it @philipcmonk ! 

Just ran and finally got arvo assembled etc. I jumped the gun a bit and did `|cancel until cue-length=0` and `|merge %base our %home, =gem %that` before the OTA was over, absent-mindedly copy and pasting. I hope that isn't an issue? I have most of the logs from when I ran `|commit %home` and came across quite a few errors, unsure if they are of critical importance or not. https://pastebin.com/edYvyjgq

I no longer see the `arvo: replacement event (for %hear) failed` error but I just have behn now spinning with CPU usage at 99% on the cli. Is this normal/to be expected or did something go wrong along the way?

> "so just |unsync wherever you're synced for them, and then |sync again."
Lastly at what point/how should I do this?  (Sorry not familiar with sync'ing and unsyncing process)

matildepark:
@philipcmonk Was this already deployed?

philipcmonk:
Yes.  If someone's still having issues related to this, please make an issue.

PR #2979, opened 2020-06-05
publish: fix store reducers

liam-fitzgerald:
fixes store reducers for SPA

### DIFF
--- a/pkg/arvo/app/publish.hoon
+++ b/pkg/arvo/app/publish.hoon
@@ -2124,7 +2124,6 @@
       [[[~ %json] [%'publish-view' %notebooks ~]] ~]
     %-  json-response:gen
     %-  json-to-octs
-    %+  frond:enjs:format  %publish-response
     (notebooks-map:enjs our.bol books)
   ::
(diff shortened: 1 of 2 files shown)


loganallenc:
Please append publish-response back to the Hoon files and instead update the JS reducers to grab the object from within publish-response such that we have namespaced updates like all the other apps and reducers.

liam-fitzgerald:
We still have namespaced updates though, it's just that the namespace is coming from the frontend instead of the backend. Because the publish reducers switch on json.type instead of `Object.keys(json)[0]` and adds the variables from the request into the store event, the pattern we use elsewhere feels awkward imo. Besides HTTP requests are implicitly namespaced because the store update is tightly coupled to the request itself. For example, this would be the reducer if the namespace comes from the backend.
```javascript
//  fetchNotesPage
.then((json) => {
  let res = json['publish-response'];
  res.type = 'notebook';
  res.host = host;
  res.notebook = book;
  res.startIndex = start;
  res.length = length;
  this.handleEvent({
    data: json
  });
});
  ```

PR #2980, opened 2020-06-05
interface: improve webpack dev config

liam-fitzgerald:
Fixes an issue where only the first ship would receive the updated files
from webpack. Enables watch mode. If herb is set to true in the urbitrc,
then webpack will commit the desk for each change on every ship.


matildepark:
yay

liam-fitzgerald:
(you'll need to have `herb` in your `$PATH` for the integration to work)

PR #2991, opened 2020-06-08
build: update nixpkgs and support gcc 8

brendanhay:
The currently pinned nixpkgs commit in `nix/nixpkgs.nix` is from December 18, 2018. Updating this to a recent nixpkgs commit means GCC is updated from version `7.3.0` to `8.3.0`. 
The new GCC coupled with `-Wall` raises [format truncation](https://gcc.gnu.org/onlinedocs/gcc-8.4.0/gcc/Warning-Options.html#index-Wformat-truncation) warnings from `snprintf` usage that are then treated as errors via `-Werror`.

I've manually bumped the buffer sizes according to the warning's recommendations in serf and u3 related code. The most commonly recommended alternative seems to be checking the `snprintf` return code for `< 0` (allocation failure) or `> 0` (number of truncated bytes).

Additionally, linking fails with a missing `pthread` dependency so I've been added this to `configure` - please verify.


brendanhay:
This won't be able to be merged until we can get CI to proceed. The updated version of nixpkgs requires a newer version of `nix >= 2.2` than travis provides.

joemfb:
@brendanhay, CI is fixed on master as of #3005 -- you can merge or rebase this.

abathur:
I'm not sure if it helps with the situation here, but just wanted to let you know that Nix support is working on Travis again, and the default version is updated to 2.3.6, as of about an hour ago.

botter-nidnul:
This seems to fix https://github.com/urbit/urbit/issues/2067

Issue #2883, opened 2020-05-09
chat: image load borks scroll position


liam-fitzgerald:
**Describe the bug**
In chat, images loading in on slow connections bork the scroll position

**To Reproduce**
1. Open an image heavy chat on a mobile device
2. As the images load in, your scroll position will be disrupted

**Expected behavior**
Scroll position should be stable as images load in.

matildepark:
I guess what we want is to wait until everything is 100% loaded before processing scroll behaviour?

matildepark:
Turns out this is specifically a Safari issue because [scroll anchoring](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-anchor/Guide_to_scroll_anchoring) generally is on by default and prevents content reflow outside the viewport. [Status is unclear](https://bugs.webkit.org/show_bug.cgi?id=171099).

PR #2997, opened 2020-06-10
various-js: improve scrolling experience on iOS

liam-fitzgerald:
Removes use of vh units, as they don't play well with safari
(https://bugs.webkit.org/show_bug.cgi?id=141832). Adjusts layouts
accordingly. Fixes the bug causing scrolling jank in iOS safari


galenwp:
I'll be very happy if mobile scrolling is in fact fixed!

matildepark:
Different bug, I think, @galenwp. This is SPA specific and would’ve been cumulative. The mobile bug you refer to is #2883.

PR #3001, opened 2020-06-11
eyre: support scrying into userspace over http

Fang-:
Minimum viable scries-over-http inside. We support scrying for any mark (filetype), but do need that mark to be able to convert to `%mime`. If the scry result is of a different mark than was asked for, a conversion for those will need to exist as well.

I suspect the dependence on "real" marks will be a little bit annoying, considering the variety of types that scry results may give. These should have enough overlap with types that are already in use though, so writing a `+grow` to json or w/e should generally still be trivial.

Note that this only allows scrying into gall. Eyre doesn't currently expose data from anywhere else, instead focusing entirely on userspace (channels/apps, generators). Exposing vane scry endpoints seems way out of line with that, so I'm opting out of it.  
"But clay scries!" Considering the frequent need for caching headers, unauthenticated requests, fallback files and other such things, fileserving is probably better implemented as a dedicated `action:eyre`.

Additionally, this only allows scrying with the `%x` care. `%y` might be nice to support, but we don't enforce a consistent result type from it. We could treat it the same way we treat `%x`, but that feels a _little_ weird. This seems to bleed into general "enforce care usage style in userspace" work.

RFC. Do these limitations seem sane for the time being?

(Note that if `release/next-sys` goes out without `ford-fusion` merged in, the `+load` logic here will need to be updated.)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -856,11 +857,71 @@
...
+    ::  attempt to find conversion gates to the requested mark, then to mime
+    ::
+    =/  tub=(unit tube:clay)
+      ?~  ext.req
+        (find-tube mark %mime)
+      ?~  to-mim=(find-tube u.ext.req %mime)  ~
+      ?:  =(mark u.ext.req)  to-mim
+      ?~  to-ext=(find-tube mark u.ext.req)   ~
+      `|=(=^vase (u.to-mim (u.to-ext vase)))
+    ?~  tub  (error-response 500 "no tube from {(trip mark)} to mime")
```
philipcmonk, line 895: Perhaps should say "via {<ext.req>}" or something.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -856,11 +857,71 @@
...
       (handle-request:by-channel secure authenticated address request)
+    ::
+        %scry
+      (handle-scry authenticated address request(url suburl))
     ::
         %four-oh-four
       %^  return-static-data-on-duct  404  'text/html'
       (error-page 404 authenticated url.request ~)
     ==
+  ::  +handle-scry: respond with scry result, 404 or 500
```
loganallenc, line 868: Can you describe some use-cases this would serve? So far I think I can imagine it being used to serve a `graph` as an RSS feed.

Fang-:
Some more changes to this incoming: want to move the userspace scry mark conversion into gall where it belongs, and only respond to GET requests on the scry endpoint.

Also just noticed this permits unauthenticated requests, which it certainly shouldn't right now. Always intended on just `403`ing those for now, but it slipped through.

re @loganallenc:

> Can you describe some use-cases this would serve?

RSS feeds sound cute! But note that this currently (lacking any real namespace-wide permissioning solution) will only respond to authenticated requests.

The shorter-term intended use-case is for easier data fetching on the frontend. ie sending scries on pageload, instead of opening subscriptions. I imagine this might also make it easier to get at specific bits of data from the frontend: implementing a scry endpoint vs implementing a subscription path.

Transitioning to doing that now will allow easier performance gains in the future, when such requests could be handled statelessly by the runtime (no disk write overhead), and eventually even in parallel with your ship's normal process.

> What do you see the next steps being?

On the frontend side, in the short term (after ford-fusion goes live), would be good to see these get used for static data requests as described.

I just got a-ok'd to work on runtime-level support for scries, which hooks into and/or overlaps with this, making the first steps towards potential performance gains described above.

Fang-:
> I’d love to see permissions thought through more

Hah. If it's any consolation, "figure out unauthenticated request handling" has been on my Eyre to-do list for a while now. The current state of things there has been a real pain-point in my personal projects. It's just such a big can of worms that it hasn't seemed wise to open yet.

Whenever scries get reified (Soon™?), that'll give us more surface area/contexts for thinking about unauthenticated requests, and more ways to ensure whatever we settle on is the Correct approach. I already have a hunch, but all in due time...

joemfb:
Merge master back into this branch to fix CI (see #3005)

PR #3004, opened 2020-06-11
group-hook: add /synced scry

liam-fitzgerald:
Adds a simple /synced scry to ease the migration of splitting group-hook
into group-push-hook and group-pull-hook.

Needs to be OTA'd before new groups is rolled out. Ping @ixv

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -286,7 +286,13 @@
     ==
   ::
   ++  on-leave  on-leave:def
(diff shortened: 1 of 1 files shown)


loganallenc:
Maybe my github is acting out but it’s showing that this change was performed to chat-hook? I’ll go use a different client and verify

matildepark:
It is indeed performing the change to chat-hook. @liam-fitzgerald ?

matildepark:
Liam confirmed fine to close out of band.

PR #2946, opened 2020-05-28
[WIP] King Haskell New IPC

benjamin-tlon:
Stubbed out:

- [x] Handle replacement events (stubbed out now b/c interface can't
  handle unparsed nouns)
- [x] Handle IPC errors by killing serf process.
- [x] Handle `peek` and `pack` in `swimming` flow.
- [x] Documentation for `Urbit.Vere.Serf.IPC`.
- [x] Unstub slog/stder/dead callbacks on serf config.
- [x] Remove GoodParse hack in newRunCompute.
- [x] Bring back tank printing.
- [x] Handle serf stderr message correctly.
- [x] Bring back `logEvent`.
- [x] Snapshots should block until that event is commited to disk.

King-Haskell specific features:

- [x] Re-implement `collectFX` flow.
- [ ] Test new `collectFX` flow

Performance:

- [x] Batching during replay.
- [x] Batching during normal operation.

Polish:

- [x] Cleanup batching flow.
- [x] Think through how to shutdown the serf on exception.
- [x] Logging for new IPC flow.
- [x] Logging for boot sequence.
- [x] Bring back progress bars.
- [x] Hook up error callbacks to IO Drivers.

Unrelated bugs:

- [x] Handle ^C in connected terminals.
- [x] terminal driver seems to have a race condition when spinner changed
      too quickly.
- [x] King should shutdown promptly on ^C. Always takes 2s in practice.

PR #3006, opened 2020-06-11
u3/serf: ipc-redux: refactoring and bugfixes 

joemfb:
I did this work recently on top of #2946 (which was probably a bad idea). PR'ing it now for posterity and legibility -- otherwise it'll be buried in merge commits.

### DIFF
--- a/pkg/urbit/include/noun/aliases.h
+++ b/pkg/urbit/include/noun/aliases.h
@@ -84,6 +84,11 @@
 #     define u3nt(a, b, c)     u3i_trel(a, b, c)
 #     define u3nq(a, b, c, d)  u3i_qual(a, b, c, d)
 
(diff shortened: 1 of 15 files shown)


joemfb:
change summary:

refactors the newt IPC implementation:

- print errors to stderr (necessary for the serf)
- allocate less on write
- have defined sync/async modes for multiple messages in one read

refactors `u3i_*`

- general source cleanup
- rewrites `u3i_chubs()`., removing inappropriate `c3_malloc()` call
- adds sampling-profiler allocator labels to all the u3i_* atom allocator functions
- implements `u3i_list()` constructor

- fixes a (latent) bug in `u3r_mug_words()`

- fixes a memory leak in the rewritten serf

PR #3010, opened 2020-06-12
Bart Info app

ronreg-ribdev:
This is a Landscape app that implements the BART scheduler search functionality as per the Urbit Grants bounty https://grants.urbit.org/bounties/1026701722-bart-train-tracker-landscape-tile

I don't consider this code to be quite complete and am waiting for feedback from Tlon on some details of the frontend, so this is a draft PR.

@urcades


matildepark:
Very late response but out of band we decided it was acceptable to accept a grant from a GitHub repository as the distribution model.

PR #2900, opened 2020-05-15
hoon: further virtualization improvements

joemfb:
This PR includes my review feedback for #2285, modernizes the style of the remaining virtualization gates, and adds new unitary (trace-less) virtualization gates (`+mole` and friends).

/cc @frodwith

PR #3011, opened 2020-06-13
arvo: compile hoon/arvo in separate roads

philipcmonk:
Adds +mure to run a trap in a separate road.  This should eventually be
just a hint.

Vega was running inside a mule, but since +load was called within vega,
the new kernel was all run within the same mule, so it didn't actually
get to reclaim the space after hoon compiled.

We verified this with printfs in u3m_fall.  On the test ship (from
mainnet) which had 800MB used, vega was taking interior free space from
950MB to 450 over the course of compiling hoon, then each vane would go
from about 450 to 350 and then back to 450 once it finished (which
proves they were correctly isolated).  With this change, after hoon
compiles the free space goes back up to 950MB.  This gives us a lot more
space to compile OTAs.

We had to slightly refactor the logic for doubly-recompiling hoon, since
+mure as written produces a `?(!! _trap)`, and you can't find faces in the
result of the trap.  We could bake mure, but that's rather awkward.  I
wonder if there's a way to fix this as a wet gate.

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -1213,7 +1213,7 @@
   ::
   =/  raw
     ~&  [%hoon-compile `@p`(mug hun)]
-    (mure |.((ride %noun hun)))
+    (road |.((ride %noun hun)))
   ::  activate the new compiler gate, producing +ride
   ::
   =/  cop  .*(0 +.raw)
@@ -1230,7 +1230,7 @@
     ?:  =(nex hoon-version)
       [raw cop]
     ~&  [%hoon-compile-upgrade nex]
-    %-  mure  |.
+    %-  road  |.
     =/  hot  (slum cop [%noun hun])
     [hot .*(0 +.hot)]
   ::  extract the hoon core from the outer gate (+ride)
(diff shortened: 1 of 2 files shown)


belisarius222:
We should add a comment to +mure, but the code LGTM

philipcmonk:
Where is mure used in that?  `grep -r mure pkg/{urbit,arvo}` finds nothing.

belisarius222:
A road hint would be better, but we don't have one.  Since this is so similar to +mule, +mack, +mute, etc., I think it belongs in hoon.hoon, since it'll make it easier to reorganize all these arms later.

Maybe we could rename it +murk?

—
~rovnys-ricfer
https://urbit.org

joemfb:
My bad! I really thought I had typed #2900 ....

joemfb:
@belisarius222, it doesn't need to be anywhere in particular, it's just a helper around `+mule`. And it doesn't require any new jet tricks. I don't think this should need to exist in the stdlb, so I think it'd be better to not put it there. Do you have plans to use it elsewhere?

philipcmonk:
I'll defer to that one!  That's more similar to mule/mute anyway.  I feel like this isn't really in that category because semantically it's not virtualizing anything -- semantically it's just a hint, more akin to +slog than anything.  Perhaps `+road` or `+rome`?

belisarius222:
It's used in both Arvo and Clay, and I suspect it's likely to be used in other places that could produce a lot of garbage.  It's true that it doesn't require jet tricks, although its sole purpose is language-level memory management, which makes me feel like it belongs in hoon.hoon at least spiritually.

Since it's in Arvo, it needs to be in either hoon.hoon or arvo.hoon.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
I changed the name to `+road`.  As @belisarius222 mentioned, this is used in fusion's Clay, so it needs to be somewhere accessible or duplicated, and I think it's a generic enough utility to be worth having in the standard library.

Issue #2908, opened 2020-05-19
webchat freezes/crashes browser when loading huge scrollback


Fang-:
**Describe the bug**
Trying to display a channel with 20k unread messages makes the browser slow to a crawl and never get out.

**To Reproduce**
1. Join a popular channel
2. Don't view it for a while
3. Try to view it

**Expected behavior**
It to load only the handful of messages around the last-read marker, and load more recent ones on scroll-down.

Alternatively, to not auto-upscroll to the last-read marker if there are >1k (or something) unread messages.

joemfb:
Another option would be to paginate both forward and backward above some threshold.

matildepark:
> Alternatively, to not auto-upscroll to the last-read marker if there are >1k (or something) unread messages.

FWIW this is what other chat clients do. I can look into this ...

ohAitch:
"throw away scroll position unpredictably" probably not the play here, you want a prominent "scroll to bottom button" for the option but mostly like need to never mount 50k DOM nodes

loganallenc:
> Another option would be to paginate both forward and backward above some threshold.

I believe we may do this already unless the implementation changed.

belisarius222:
~rovnys-ricfer has run out of loom, partially due to Eyre's channels using over 140MB, which doubled to 280MB after a process restart.  @joemfb's ship just ran out of memory too, with 800MB in Eyre.  I wonder if big messages from Urbit to the web client on reload could cause unfinished HTTP responses to build up in Eyre?

For the past week or so, trying to load urbit-help would try to load tens of thousands of messages most times I clicked on it, which would often kill the browser tab.  If all those messages were enqueued from Gall to Eyre at once and they built up per channel, maybe that would explain these bail:meme's.

joemfb:
I waited out the `~h12` channel timeout, restarted my ship, and %eyre is now using 24 bytes for channels. It's nice that that kind of recovery is possible, but this space leak in %eyre is still very concerning -- a little more memory consumption and my ship wouldn't have been able to boot up and process timers ...

It does look like chat message loading is paginated, so I assume the problem is that the frontend makes lots of paginated requests to load the whole backlog.

joemfb:
And of course, now I have > 800MB in free lists, which is its own kind of reward ...

loganallenc:
I believe the problem is that in many cases messages are replicated and sent to all of the open channels. Could we do de-dupe nouns like we do in ames?

joemfb:
[...]

Issue #2897, opened 2020-05-13
Unread count doesn't reflect group filter


galenwp:
**Describe the bug**
If I'm filtered on a specific group, in Chat, my unread count (in the `<title>`) increments when messages arrive in other channels.

**To Reproduce**
1. Go to Chat, filter on Group A
2. Receive some messages in Group B
3. You don't see those messages, nothing appears unread, but your unread count is going up

**Expected behavior**
I think the expected behavior is that the unread count in the title is just what you see on the page. I could also see it being something like [filtered unread]/[total unread] or something.

matildepark:
Patched in https://github.com/urbit/urbit/commit/5757fd8f5d28db111ee3cfa614c5f96e5b322c8b, coming in with the SPA.

Issue #2889, opened 2020-05-11
Publish: 'Manage this group in the contacts view' is a broken link


galenwp:
**Describe the bug**
On the subscribers page we provide a dropdown with a link, but the link is broken.

**To Reproduce**
1. Go to a publish notebook
2. Click on 'Subscribers'
3. Click the 'Options' dropdown, then the link
4. See a 404

**Expected behavior**
This should go to `/~groups/~dopzod/tlon` but instead it goes to `/~contacts/~dopzod/tlon`

**Screenshots**
<img width="549" alt="Screen Shot 2020-05-11 at 8 08 28 AM" src="https://user-images.githubusercontent.com/1467211/81577832-d5b77380-935e-11ea-841b-fb4dd1033716.png">

matildepark:
As referenced above, I patched this on feat/spa and this issue will close when feat/spa merges.

PR #3025, opened 2020-06-17
Landscape: Single-page application rearchitecture

matildepark:
- The build process for each Landscape app has been consolidated into a single build process for all Landscape apps, using Webpack instead of Gulp. It can be run by using `npm run build:dev` inside `pkg/interface`.
- All apps have been migrated into a single application, deduplicating all common UI components across the interface.
- Front-end API, subscription and reducers have been rearchitected into a central location, deduplicated, and permissioned per application (such that within Chat, the API is no longer accessible via `window.api` and able to send authenticated actions to any app as it likes).
- All custom CSS in each application was deduplicated and centralised in preparation for `indigo-react` integration.
- `ContentType` has been added to S3 uploads such that they do not prompt a download window (#2865, #2890) 
- Links unread counts no longer overlap (#2833).
- Publish URL to Groups view has been fixed (#2889).
- Third party applications can now add their applications to the home screen by just passing their name, icon URL, and a URL to refer, to `%launch`; they can also easily mount their endpoints using `%file-server`, without having to write `on-watch` boilerplate for handling HTTP requests. More details to come for developers.
- Landscape can optionally be hosted at `~landscape` and allow you to slot a static page at the root directory instead.
- Unread counts in title bar now filter out the currently filtered channels, when viewing only a specific group's channels (#2897).
- Chat no longer scrolls to top of scrollback if the backlog exceeds 1,000 new messages (#2908).
- In chats, images are sized at a maximum of 250px to prevent scroll jumping on Safari (#2883).
- Code blocks now overflow properly and scroll within their container when too wide (#2829).
- The code and s3 buttons in Chat are now graphically aligned.

cc @liam-fitzgerald


ryjm:
will existing landscape apps still sorta work? and will there be any guidance on how to migrate them? i know, i know, the perils of being on the bleeding edge, but i'd really like to spend more time in hoon and much less in react :grimacing:

matildepark:
> will existing landscape apps still sorta work?

Should be fine.

> and will there be any guidance on how to migrate them?

I'm working on it! Just cleaning up create-landscape-app today and tomorrow, and when we OTA (quite soon) I'll write some advice on migration for existing third-party apps.

Essentially instead of your Hoon boilerplate having HTTP request handling etc etc. you just let `%file-server` handle it, so a lot of boilerplate is not necessary. 

You can also go without back-end code entirely if you like; you can just poke `%file-server` to mount an endpoint to the respective folder in /app and it will mount all the HTML, JS, PNG files in the folder. You also can just poke `%launch` to add a basic tile with only an image and a destination URL.

The tradeoff is that widget-style tile apps are basically no longer possible without writing it directly into `pkg/interface/src/apps/launch`.

ryjm:
> > will existing landscape apps still sorta work?
> 
> Should be fine.
> 
> > and will there be any guidance on how to migrate them?
> 
> I'm working on it! Just cleaning up create-landscape-app today and tomorrow, and when we OTA (quite soon) I'll write some advice on migration for existing third-party apps.

beautiful. i think i forked from a pretty old version of create-landscape-app, but i'm optimistic. the main thing i'm worried about is the boilerplate for the OS1 style interface - for srrs, i ended up porting a lot of the frontend code for publish and tweaking the css, and i'd love to get rid of all the hacky stuff i did there... but i'm hoping that the app will still work while i do that.

philipcmonk:
Have we tried this on `-L` copies of real ships?  I've found that to be a good way of shaking out OTA bugs.

I copied this into a `-L` copy of ~wicdev and now my homepage is blank (even after force-refresh).  Here's the trace: https://pastebin.com/nrd20pib.  I just ran `|commit %home` and then at the end there you can see me trying a manual `:goad %force`, but that had no effect except to print an error again.  I think file-server is failing to start because something's already bound: `"url already bound to [clay-base=/app/landscape public=%.n]"`.

matildepark:
Hey! That error (I think) shouldn’t mean anything — I want to ask how you merged it in. If you literally took this branch, you’ll need to compile the JS (`npm run build:dev`). I’ve tried the OTA case on ~sitden and @loganallenc has tested a few OTA cases today with the last patches he wrote.

philipcmonk:
You're correct, it works when I do that.  Thanks!

The fact that +on-load keeps crashing every time on you :goad %force is bad, because it means that app can't reload.  On fusion, this will mean no OTAs will succeed until the on-load succeeds.  Not a showstopper for this, but it will need to not crash before fusion goes out.

matildepark:
It would seem imperative to tackle now, then. @loganallenc @liam-fitzgerald would you mind investigating? It seems like `%soto` just on-loads every single load.

liam-fitzgerald:
On it!

tylershuster:
Is the ~landscape url documentation anywhere? I was just dreaming of this feature yesterday

matildepark:
I'll try to look into it today and write something up.

loganallenc:
I don’t think this part has been fully integrated on the front-end yet to function properly. Needs some additional love.

tylershuster:
If you can point me in the right direction I might try giving it some

matildepark:
@tylershuster As you can see [here](https://github.com/urbit/urbit/blob/master/pkg/arvo/app/file-server.hoon#L91-L94) you can poke file-server to set the Landscape homepage prefix as a [unit of terms](https://github.com/urbit/urbit/blob/master/pkg/arvo/sur/file-server.hoon#L6) (you know, `%name` etc). When we [subscribe](https://github.com/urbit/urbit/blob/master/pkg/interface/src/subscription/launch.js) to Launch, it [should provide that prefix](https://github.com/urbit/urbit/blob/master/pkg/arvo/app/file-server.hoon#L195). We don't really specifically [reduce it](https://github.com/urbit/urbit/blob/master/pkg/interface/src/reducers/launch-update.js), but we do take it in the initial object wholesale, so we should have access to it inside state.

We would then need to use that [and make our router relative to the prefix](https://github.com/urbit/urbit/blob/master/pkg/interface/src/App.js#L82) for Launch. We'd need the [back nav to update too](https://github.com/urbit/urbit/blob/master/pkg/interface/src/components/StatusBar.js#L55-L66).

On top of all this, in our current state this isn't possible because the store will wipe as we navigate and we'll lose track of the prefix. But after #3038 this should be fine?

ryjm:
> The tradeoff is that widget-style tile apps are basically no longer possible without writing it directly into pkg/interface/src/apps/launch.

is this only when you're using file server, or in general? how would you go about porting a custom tile? here's what i had to do get srrs working with a basic tile if it's useful for anyone else: https://github.com/ryjm/srrs/commit/d5ff0ec079dee56683a33c2e979fda176d8482ec

PR #3026, opened 2020-06-18
gall: import/export

liam-fitzgerald:
Adds import/export functionality to gall.

#### tl;dr

with this PR, you can backup your ship, breach it and restore its state seamlessly

#### Description

Adds %a care to gall scry. This returns a `+lore` representing the current state of gall at the time.

Adds the `%book` task to gall.  %book restores gall to a previously
exported state.

%book should be given to gall by unix on boot, to set the state of the
gall apps before they are allowed to start. To facilitate this, we add
+solid-backup and +brass-backup generators. This builds their respective
pills, but appends a %book task to the userspace ova. Thus we always
package the backup with a copy of the desk at the time. This is to
ensure that the type of the saved state matches the type of the state
from ++on-save. We need these types to match because we stitch the type
from ++on-save to the raw data from the backup, as we cannot serialise
+type.

%book restores the state and incoming and outgoing subscribers. All of
the outgoing subscriptions receive a %kick. For the incoming
subscriptions we call +on-leave.

An %book that is received once apps have booted is ignored.

Changes the type of $agent:gall to include a breached flag in +on-load.
Updates agents to conform to this type.

If the breached flag is %.y, then all vane state has been wiped and must
be recreated. Most agents can simply ignore this flag. See
link-listen-hook for an example of recreating the vane state.

Includes a ph integration test for validating that this works without having to backup and then breach.

#### Usage:

in the dojo:
```
.backup/pill +brass-backup
```
Will create a new backup and save it at `{pier}/.urb/put/backup.pill`. (`+solid-backup` will also work, but takes far longer)

to restore, upon booting:
```
./urbit -w hastuc-dibtux -k hastuc-dibtux-10.key -B {path to backup}
```

cc: @Fang- @loganallenc @ixv have I missed any vane state that I need to recreate?

also: can we merge master into release/next-sys? I don't see any reason master should be ahead of release/next-sys


vvisigoth:
yes
—
~poldec-tonteg
http://urbit.org

Fang-:
You added a `bin/brass-a.pill`. Is this intentional?

With the current gall logic, what happens if the `+on-load` fails? ie the app forgot about the type, or the type was too new somehow (could happen if you boot with older pill). Does it just crash on the import, and proceed past it? Lacking other import-from-lore affordances, maybe you want to fail boot in that case?


As part of my "runtime scries" work, it's already possible to scry the `lore` out and jam it to disk. I'm trying to extend support for this specific case, by allowing you to point to that jam-file, and append its `lore` to the boot sequence.

I've started out by trying to manually construct the `%book` ovum and appending it to whatever pill was used. Now I'm wondering though, how sane "construct ovum in vere" actually is. Maybe we want to be able to scry the full ovum out of gall instead? Having a vane know how to build this seems less strange than having the runtime know, but it's still not ideal...
RFC @joemfb or others.

liam-fitzgerald:
> You added a `bin/brass-a.pill`. Is this intentional?

no, whoops.

> With the current gall logic, what happens if the `+on-load` fails? ie the app forgot about the type, or the type was too new somehow (could happen if you boot with older pill). Does it just crash on the import, and proceed past it? Lacking other import-from-lore affordances, maybe you want to fail boot in that case?

This is the point of packaging the ovum inside the pill. In theory, the on-load should never fail because it's already been run against the current state. This is unavoidable due to our inability to serialise +type, necessitating the stitching of the type of +on-save onto the raw noun.

> As part of my "runtime scries" work, it's already possible to scry the `lore` out and jam it to disk. I'm trying to extend support for this specific case, by allowing you to point to that jam-file, and append its `lore` to the boot sequence.

We need to package the pill with the ovum as mentioned earlier, so in general we can't do this kind of unconstrained import

> I've started out by trying to manually construct the `%book` ovum and appending it to whatever pill was used. Now I'm wondering though, how sane "construct ovum in vere" actually is. Maybe we want to be able to scry the full ovum out of gall instead? Having a vane know how to build an ovum seems less strange than having the runtime know, but it's still not ideal...

My vote is to put the pill generation logic in ++zuse and then the runtime can generate the backup by way of the ivory pill. I have no solid opinion on whether we should scry the lore or the ovum out of gall, so will defer to others on this.

Fang-:
I'm not able to use this successfully, by the way. Booting off the backup pill ends up in lots of `subscribe wire not unique`, which ultimately prevents me from subscribing to my own dojo.

Building an entirely new pill from the user's system still seems not ideal to me. To copy some discussion from emails we had about this previously:

> > Wouldn't [building pill from the user's ship] break recovery in the "I messed around with /sys too hard and broke core functionality" case?
> > ...Though maybe, at that point, you should've been smart enough to make backups ahead of time anyway?
>
> you could in theory splice the %rmem ovum onto a known valid brass pill that was built off a desk where the types of the gall agent states are the same as the broken desk. This would address the /sys problem although I don’t really think this is a use case worth caring too hard about.

I'm not fully convinced we should discard the "broken sys" case. The problem here, with not being able to guarantee that the app can handle the old state data, seems like it might be resolved by 1) using an up-to-date pill and 2) having apps keep old state adapters around.
(1) we should probably be doing anyway(?), and (2) we could shrink in scope if we say this might not support importing from a ten-year-old pier.

The email thread didn't go anywhere past this either, so RFC I guess.

liam-fitzgerald:
> Building an entirely new pill from the user's system still seems not ideal to me. To copy some discussion from emails we had about this previously:
> 
> > > Wouldn't [building pill from the user's ship] break recovery in the "I messed around with /sys too hard and broke core functionality" case?
> > > ...Though maybe, at that point, you should've been smart enough to make backups ahead of time anyway?
> > 
> > 
> > you could in theory splice the %rmem ovum onto a known valid brass pill that was built off a desk where the types of the gall agent states are the same as the broken desk. This would address the /sys problem although I don’t really think this is a use case worth caring too hard about.
> 
> I'm not fully convinced we should discard the "broken sys" case. The problem here, with not being able to guarantee that the app can handle the old state data, seems like it might be resolved by 1) using an up-to-date pill and 2) having apps keep old state adapters around.
> (1) we should probably be doing anyway(?), and (2) we could shrink in scope if we say this might not support importing from a ten-year-old pier.

The problem is not entirely 'can the app handle old state data', it's largely a product of our inability to serialize $type in any meaningful way. This means we can't write the vase to disk, and thus have to find a $type to graft onto our raw noun. I'd be happy to see a solution for this that doesn't involve a full pill build, but it seemed to be the only reasonable way to do this.

ohAitch:
I mean, the pill _is_ writing the vase to disk right?

matildepark:
Ongoing inquiry: does this maintain subscriptions? It seems like "yes and no"?

joemfb:
I've been meaning to write some feedback for this, sorry for the delay.

This is pretty neat work, and functionality that we desperately need. But I have some concerns with the approach taken here, and the constraints of deploying this practically.

The bundling of the update in a new pill (created on the old ship) creates a trusted bootstrap sequence for the "evil vase" that %gall will construct for each app. But that binding effectively means that this tool can only be used to "re-bootstrap" a ship into a state as close as possible to it's previous state, but also new. This is a substantial constraint, and means that this tool cannot be used for any kind of breaching update (rolling or otherwise), or to span any changes to gall agent implementations. Since this design leaves no room for adapters, or upgrades of any kind, any such changes would have to be subsequently delivered to the new ship over-the-air.

Other designs are definitely possible. For one thing, we can absolutely serialize a vase -- the existence of `|pack` (and `+solid` pills) is a proof-by-construction. For another, we can inject events into specific positions in the bootstrap sequence without including them in the pill itself. And it seems at least conceivable that ford could produce the input source (with full dependencies) for each app, if it turned out that we needed to reconstruct the type of the state and avoid serializing a full vase.

This work should also be contrasted to @eglaysher's earlier work adding import/export to `:lens` and `herb`. The biggest difference between these approaches is that his work requires an application to opt-in with ad-hoc functionality, while this design does not. But the strength of the ad-hoc approach is that it leaves room for adapters and upgrades (and re-uses the mark system, so no special type provenance/reconstruction is necessary).

While automatic is generally preferable to manual, I'm not sure it is in this case -- "restoring" the state of an application post-breach feels like a necessarily ad-hoc task. Somethings should cross the divide, others should not, and application-specific logic will be necessary to get that right.

Finally, there's the question of deployment. Using scry to extract the data seems right, and is (at least somewhat) more likely to be continue to work in suboptimal scenarios (which are likely if a user is individually breaching). Still, it'd be best to have the data available before failure occurs, or it may not be possible to extract it for migration. So I think we'll want to consider some (opt-in?) automation around such functionality, and should consider building it into the runtime and pier itself.

In summary, I think we'll need to establish tighter requirements for this feature. I'd especially like to hear feedback from @philipcmonk and @belisarius222.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1395,9 +1404,87 @@
...
+    ::
+    ++  ap-new-state
+      ^?  |.
+      =^  init-cards  agent.current-agent
+        on-init:ap-agent-core
+      =/  old-state=(unit [state=* wex=boat sup=bitt])
+        (~(get by lore.state) agent-name)
+      ?~  old-state
+        [init-cards agent.current-agent]
+      :: dill's dependency on hood means we skip it
```
joemfb, line 1422: This limitation should be explicitly noted as well.

matildepark:
Superseded by #3604?

liam-fitzgerald:
> Superseded by #3604?

yep!

PR #3027, opened 2020-06-18
soto: run +on-load migration once

liam-fitzgerald:
runs state adapter precisely once.


matildepark:
Thanks for getting to this so promptly Liam!

PR #3029, opened 2020-06-19
clay: make reachable-takos linear instead of exponential

philipcmonk:
When merging, +reachable-takos is called roughly once per merge commit
in the ancestry of the new commit.  +reachable-takos was exponential in
the number of merge commits in the ancestry of the commit it's looking
at, due to mishandling of the accumulator.  This makes it linear.

Of course, linear x linear is still quadratic, which is not great.  I
doubt +reachable-takos can be made asymptotically better, but
+reduce-merge-points/+find-merge-points probably can.  50 merge commits
already gives about 14.000 iterations through the loop in
+reachable-takos.  Another option is to try to memoize this somehow, but
a simple ~+ is insufficient since `s` is usually different.

In local tests on macOS with a -L copy of ~wicdev-wisryt, this speeds up
OTAs significantly.  The majority of time was spent on this.  This is especially
true for updates which didn't involve changes to /sys, since those involve a
lot of compiling.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -2515,13 +2515,17 @@
     ++  reachable-takos                                 ::  reachable
       |=  p/tako
       ^-  (set tako)
+      =|  s=(set tako)
+      |-  ^-  (set tako)
+      =.  s  (~(put in s) p)
       =+  y=(tako-to-yaki p)
-      %+  roll  p.y
-      =<  .(s (~(put in *(set tako)) p))
-      |=  {q/tako s/(set tako)}
-      ?:  (~(has in s) q)                               ::  already done
-        s                                               ::  hence skip
-      (~(uni in s) ^$(p q))                             ::  otherwise traverse
+      |-  ^-  (set tako)
+      ?~  p.y
+        s
+      ?:  (~(has in s) i.p.y)
+        $(p.y t.p.y)
+      =.  s  ^$(p i.p.y)
+      $(p.y t.p.y)
     ::
     ::  Get the data at a node.
     ::


belisarius222:
LGTM.

I wonder if we could eventually memoize the reachable takos by storing them in the Clay state as a cache that could be cleared and recomputed if needed.  The results shouldn't change over time, so I'd imagine we could cache the results; even if that doesn't improve the asymptotics, just amortizing the cost over multiple events by eagerly computing +reachable-takos immediately on every commit might be a better user experience.

philipcmonk:
Potentially true.  Of course, then you're trading quadratic time for quadratic space, which may or may not be a good tradeoff.  Might be a sweet spot in the middle where you cache every 50th merge commit or something.  I feel like "find a merge base" must be a well-studied problem.  I wonder how git does it.

belisarius222:
Do we need to find every single reachable tako for a commit?  I feel like another overall approach would be to search backward through time lazily until we find one merge base, then maybe do a full traversal between that time and the present to make sure there aren't closer merge bases.

—
~rovnys-ricfer
https://urbit.org

ohAitch:
My mind wanders to a combination of a topologically sorted vector of all
commits we've ever seen, skip list per independent chain of history, and
maps to jump between them by commit id; I think this could get you
quasilinear time and space, but it would involve a lot of duct tape.

philipcmonk:
I expect something like that is possible.  You want to be careful because you need to find all possible mergebases (which could be more than one if it's a criss-cross merge).  Right now we find all reachable takos from one parent (linear) and call it `r`, and then we trace down the ancestry of the other parent, storing in `s` any takos which are in `r` (and interrupting the recursion).  This is also linear, so we're at linear + linear, which is fine.  Note however that in the common case where you're getting repeated merges from `%base` to `%home`, the size of `s` will be linear in the number of commits to `%base` (and may in fact be equal to the set of commits in `%base`).

Now, we take `s` and get the reachable takos for each of them, and check if any of them are in each other's ancestry, keeping them if so.  Since `s` is linear and reachable-takos is linear, this is quadratic.  This is in `+reduce-merge-points`.

It seems like a better algorithm would be to lazily compute reachable-takos, and delete from our worklist as soon as we find out that it's reachable.  Best case, you find the most recent common ancestor immediately and delete everything else from your worklist (linear).  Worst case, you go in exactly backward order, saving no time (quadratic).  But it's mug ordered, so you're likely to get average case, which is that you delete half of your worklist every time, which I think is `n*log(n)`.

You could probably do better if you don't randomize the order since you probably have a guess about which is more recent, but you don't have any guarantees.  Sort by date would probably work excellent in practice, but it's not a guarantee.

ohAitch:
Does sorting by /depth/ have any more solid guarantees than date?

philipcmonk:
Depth from head?  Not really, since your history can be a long sequence of merges from unrelated branches, or they could be related but you did something weird like merge the commits in reverse order.

ohAitch:
No, from tails, summing on merges(optionally deduplicating past
merge-base). Just memoize the count for every commit.

Any merge-base will necessarily have the same number by virtue of being the
same commit; in the common case with fast-forward, or almost fast-forward,
you don't/barely touch the branch that hasn't/really updated before hitting
the merge base; when you have a criss-cross, it should still traverse each
side only once.

"Date in logical time", if you will.

philipcmonk:
I think you're right, that would be pretty fast.

philipcmonk:
This went in the Ford Fusion OTA

PR #3034, opened 2020-06-19
chat-store: added %trim command as a hack that runs once +on-load and again every 5.000 messages per mailbox

loganallenc:
Would love for this to go out in an OTA prior to Fusion so we could trim some of the duplicate message state. Thanks for making the initial version @Fang-!

This appears to have some side effects relating to message sequence numbers and unread counts, so I'm looking into that now.

Update:
After a %trim the message numbers are now guaranteed to be both contiguous and start from 0.

### DIFF
--- a/pkg/arvo/app/chat-store.hoon
+++ b/pkg/arvo/app/chat-store.hoon
@@ -8,14 +8,19 @@
   $%  state-zero
       state-one
       state-two
+      state-three
   ==
 ::
-+$  state-zero  [%0 =inbox:store]
-+$  state-one   [%1 =inbox:store]
-+$  state-two   [%2 =inbox:store]
++$  state-zero    [%0 =inbox:store]
++$  state-one     [%1 =inbox:store]
++$  state-two     [%2 =inbox:store]
++$  state-three   [%3 =inbox:store]
++$  admin-action
+  $%  [%trim ~]
+  ==
 --
 ::
-=|  state-two
+=|  state-three
 =*  state  -
 ::
 %-  agent:dbug
@@ -34,14 +39,18 @@
   ++  on-load
     |=  old-vase=vase
     =/  old  !<(versioned-state old-vase)
-    ?:  ?=(%2 -.old)
+    ?:  ?=(%3 -.old)
       [~ this(state old)]
+    ?:  ?=(%2 -.old)
+      :_  this(state [%3 inbox.old])
+      [%pass /trim %agent [our.bowl %chat-store] %poke %noun !>([%trim ~])]~
     =/  reversed-inbox=inbox:store
       %-  ~(run by inbox.old)
       |=  =mailbox:store
       ^-  mailbox:store
       [config.mailbox (flop envelopes.mailbox)]
-    [~ this(state [%2 reversed-inbox])]
+    :_  this(state [%3 reversed-inbox])
+    [%pass /trim %agent [our.bowl %chat-store] %poke %noun !>([%trim ~])]~
   ::
   ++  on-poke
     ~/  %chat-store-poke
@@ -52,6 +61,7 @@
       ?+  mark  (on-poke:def mark vase)
         %json         (poke-json:cc !<(json vase))
         %chat-action  (poke-chat-action:cc !<(action:store vase))
+        %noun         (poke-noun:cc !<(admin-action vase))
       ==
     [cards this]
   ::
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/chat-store.hoon
```diff
@@ -197,8 +232,11 @@
     [~ state]
   =.  letter.envelope.action  (evaluate-letter [author letter]:envelope.action)
   =^  envelope  u.mailbox  (prepend-envelope u.mailbox envelope.action)
-  :-  (send-diff path.action action(envelope envelope))
-  state(inbox (~(put by inbox) path.action u.mailbox))
+  :_  state(inbox (~(put by inbox) path.action u.mailbox))
+  ?:  =((mod number.envelope 5.000) 0)
+    :-  [%pass /trim %agent [our.bol %chat-store] %poke %noun !>([%trim ~])]
+    (send-diff path.action action(envelope envelope))
+  (send-diff path.action action(envelope envelope))
```
Fang-, line 239: idk why we always defer this to new events instead of just running them in-place. What's the reasoning there?

Also every 5.000 messages seems very infrequent, especially per channel. I'd put this at more like 500 or something. But also, considering `+handle-messages` exists, you could skip some of these checks, especially in cases where you're, you know, resubscribing.

pkg/arvo/app/chat-store.hoon
```diff
@@ -34,14 +39,18 @@
   ++  on-load
     |=  old-vase=vase
     =/  old  !<(versioned-state old-vase)
-    ?:  ?=(%2 -.old)
+    ?:  ?=(%3 -.old)
```
philipcmonk, line 42: `+on-load`s get really unwieldy when written with `?:` since they grow quadratically with the number of revisions.  A better pattern is successive `=?`s, like here:

https://github.com/urbit/urbit/blob/4fded00005770a84a53ff77a81ba71353f84b4bd/pkg/arvo/sys/vane/gall.hoon#L1674-L1695

This way you upgrade from previous versions one at a time and only need to think about how to advance from the previous revision to the new one.

philipcmonk:
Is this ready/going to be ready for the Fusion OTA on Monday?

loganallenc:
Yes, I’ll respond to comments early today and target it to the ford-fusion release branch.

pkg/arvo/app/chat-store.hoon
```diff
@@ -8,14 +8,19 @@
...
+      state-three
   ==
 ::
-+$  state-zero  [%0 =inbox:store]
-+$  state-one   [%1 =inbox:store]
-+$  state-two   [%2 =inbox:store]
++$  state-zero    [%0 =inbox:store]
++$  state-one     [%1 =inbox:store]
++$  state-two     [%2 =inbox:store]
++$  state-three   [%3 =inbox:store]
```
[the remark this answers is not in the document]
loganallenc: Yes.

pkg/arvo/app/chat-store.hoon
```diff
@@ -148,6 +158,27 @@
     [~ ~ %noun !>((swag [start (sub end start)] envelopes))]
   ==
 ::
+++  poke-noun
+  |=  nou=admin-action
+  ^-  (quip card _state)
+  ~&  %trimming-chat-store
+  :-  ~
+  :-  %3
```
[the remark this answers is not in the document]
loganallenc: Good point.

pkg/arvo/app/chat-store.hoon
loganallenc: 5.000 seems like a good frequency. 5.000 messages really don't take up that much space.

pkg/arvo/app/chat-store.hoon
loganallenc: Good point.

loganallenc:
Should be good to go. Responded to all the style comments.

pkg/arvo/app/chat-store.hoon
loganallenc: As for deferring to new events, I think it's a cleaner pattern for this type of operation.

loganallenc:
Ping

matildepark:
I see https://github.com/urbit/urbit/pull/3060/commits/ac864bd42a0db364a7221c8abdfdd496c9e0326a in #3060.

loganallenc:
Chatted with Ted about this last night. Closing this PR, as its contents have been merged, tested, and approved within the ford-fusion PR.

PR #3037, opened 2020-06-20
interface: correct order of arguments in resubscribe logic

liam-fitzgerald:
should fix event source reconnection issues


matildepark:
It’s kind of hilarious that this is why

PR #3038, opened 2020-06-22
interface: global (typed) store

liam-fitzgerald:
Unifies persistent state across the different apps into one global store.
Dynamically subscribes based on position in router.
Types a significant majority of the store/api/subscriptions so it should now be easier to gradually introduce TS in to components.


galenwp:
This will keep state around while navigating, right? That's pretty great.

matildepark:
@galenwp This subscribes to everything on launch and passes state to each app instead, yes. Good work Liam, a big change. Will test today.

matildepark:
Weather seems stuck in error state — doesn't have its props.

image

matildepark:
Sidebar switcher state persists as you navigate — so you can't actually use Chat if, for example, it's closed when you navigate to it.

matildepark:
@g-a-v-i-n

liam-fitzgerald:
fyi: `npm run tsc` will run the typescript checker, without actually transpiling the code (we still use webpack + babel for this). `npm run tsc:watch` does the same in watch mode.

 @g-a-v-i-n lmk if you have any issues with the tsconfig.json as it stands. I'd like to leave `noImplicitAny` as false for now, and then once we have most of the codebase converted, we can turn it on and fix anything that's still missing. My reasoning is that the noImplicitAny errors require you to systematically work from the leaves of the dependency tree and convert things methodically, rather than the ad-hoc approach that'd probably be best here.

matildepark:
Getting this in Dojo:

```
[%missing-subscription-in-unsubscribe /channel/subscription/1592856998710-55258f/47]
[%missing-subscription-in-unsubscribe /channel/subscription/1592856998710-55258f/48]
```

matildepark:
Make a group, set some metadata in your contact card, press 'share'. It works immediately but the spinner leaves us spinning on "sharing with group" ... but this appears to also be in the SPA, so maybe out of scope here.

liam-fitzgerald:
can't repro `%missing-subscription-in-unsubscribe` but i suspect it's a race condition because we don't wait for a positive ack on the subscription. Fixing it requires changing channel.js to notify the caller when a subscription is acked. I'll do this in a seperate PR. Otherwise, those errors can be treated as spurious for now.

matildepark:
Just the conflict left then. Merge release/next-userspace back in?

PR #3041, opened 2020-06-23
jets: remove peek, play, and fond jets and cap memo cache

philipcmonk:
Three commits.  The middle one is the real one, the other two are useful debugging tools, disabled.

---

noun: add functions to count size of noun

Adds a few functions to count the size of nouns in the current road.
Since this marks the nouns (high bit of refcount), you need to
"discount" them immediately after to unmark them.  Parallel functions
exist for the counting the size of a hashtable.

It would nice to hook this up to a hint, but these are useful to have
available to run in the debugger or by inserting callsites as necessary.
It's also possible to hook them up to the +jam jet gated on a special
value.

---

jets: cap memo cache and remove peek, play, and fond jets

With these changes, about 90% less memory and 15% less time is needed to
compile hoon.hoon.  The produced noun is within 3% of the same size,
which suggests this results in little if any duplication.

These are three of the four most commonly hit +ut jets.  The other is
+nest, which cannot be un-memoized without taking much longer to compile
(it didn't finish in my test).  These four jets combined for 2.3 million
out of the 2.4 million cache entries, the other +ut jets combine for
less than 100k, and literal ~+ accounted for about 50k entries.

This also caps the memo cache at 50k entries.  Even with these jets not
memoized, the memo cache grows to 357k entries and 122 MB.  Capping at
50k entries has no effect on time and reduces memory usage of the hash
table to about 25MB.  Entries are reclaimed with the clock algorithm,
which seems to be sufficient for this use.

---

jam: add commented-out functionality to count size of atom

This is a convenient way to count memory usage of noun by simplying
running `(jam 1.337 noun-1 noun-2 ... ~)`.  This should
be a hint, but for debugging this is sufficient.


philipcmonk:
`release/next-vere` has the change that makes the lmdb mapsize 1TB -- have we determined that's safe to release?

pkg/urbit/jets/e/jam.c
```diff
@@ -6,6 +6,40 @@
...
+    u3a_print_memory(stderr, "memoization cache", mem_w);
+    u3h_root* har_u = u3to(u3h_root, u3R->cax.har_p);
+    u3l_log("memoization entries: %d\r\n", har_u->use_w);
+    c3_w diff = (u3R->hat_p < u3R->cap_p) ?
+                  (u3R->cap_p - u3R->hat_p) :
+                  (u3R->hat_p - u3R->cap_p);
+		u3a_print_memory(stderr, "unused free", diff);
+    return tot_w;
+  }
+#endif
```
joemfb, line 41: For this kind of thing, we should really have a de-optimized hint opcode in the interpreter for easy extensibility. And we need to track bytecode versions so we can add new opcodes online.

pkg/urbit/noun/allocate.c
```diff
@@ -1721,6 +1721,156 @@ u3a_mark_noun(u3_noun som)
...
+        siz_w += new_w;
+        if ( _(u3du(som)) ) {
+          siz_w += u3a_discount_noun(u3h(som));
+          som = u3t(som);
+        }
+        else return siz_w;
+      }
+    }
+  }
+}
```
joemfb, line 1871: These are great! Should be really handy going forward.

pkg/urbit/jets/e/jam.c
```diff
@@ -6,6 +6,40 @@
...
+    u3h_discount(u3R->cax.har_p);
+
+    u3a_print_memory(stderr, "total", tot_w);
+    u3a_print_memory(stderr, "memoization cache", mem_w);
+    u3h_root* har_u = u3to(u3h_root, u3R->cax.har_p);
+    u3l_log("memoization entries: %d\r\n", har_u->use_w);
+    c3_w diff = (u3R->hat_p < u3R->cap_p) ?
+                  (u3R->cap_p - u3R->hat_p) :
+                  (u3R->hat_p - u3R->cap_p);
+		u3a_print_memory(stderr, "unused free", diff);
```
joemfb, line 38: I forgot about this in #3012, but there are macros or functions for each of the loom regions/categories:

- `u3a_heap()` measures the heap
- `u3a_temp()` measures the stack
- `u3a_open()` measures contiguous free space
- `u3a_idle()` measures the free lists

joemfb:
The 1TB LMDB max-mmap is safe, as in it works on all platforms. it's unsafe, as in your core dumps will reflect the full percentage of used memory (on linux, anyway -- on macos, i think they'll be the full mapped address space).

Unmapping the database on SIGABRT is an unsafe hack, as it requires either ignoring or canceling the commit thread (if active), which is undefined in a signal handler. On ipc-redux, i've disabled core dumps in the vere king and removed the SIGABRT handler. On macos, there are console.app crash reports that include backtraces where possible; on linux, we can run under gdb (ignoring SIGSEGV, which is impossible on macos).

We should probably forward-port those changes if the 1TB limit will be released before ipc-redux.

pkg/urbit/jets/tree.c
```diff
@@ -575,7 +557,6 @@ static u3j_hood _141_pen__ut_ho[] = {
   { "fine",   49086 },
   { "fire",       4 },
   { "fish",    6006 },
```
eglaysher, line 559: Since jet core identity does an equality check on the hooks, is this OTAable? My model is that if you won't recognize the compiler jets anymore...

pkg/urbit/noun/manage.c
```diff
@@ -496,7 +496,7 @@ _pave_south(c3_w* mem_w, c3_w siz_w, c3_w len_w)
 static void
 _pave_parts(void)
 {
-  u3R->cax.har_p = u3h_new();
+  u3R->cax.har_p = u3h_new_cache(50000);
```
eglaysher, line 499: When I was exploring how the cache worked, I had to set this to 800,000 for an urbit to be able to boot. However, I also didn't remove `+play`, etc, and if those are the majority of the cache entries, maybe this is OK.

philipcmonk: I was able to boot a solid pill with 50k without issue, but I'll check a brass pill.  Uncapped, removing these caches reduced the number of cache entries for compiling hoon.hoon by about 85%.

Another thing I'd like to do is just make this configurable with a CLI flag, so if something goes wrong we have a quick backup.  I know how to add a CLI flag to the daemon, is there an example I can look at of one that gets passed to the worker?

pkg/urbit/jets/tree.c
philipcmonk: Didn't realize that.  I've tested disabling the jets on a -L copy of ~wicdev, but not deleting the jets.  I'll do that, and if it doesn't work I'll re-add the jets as pass-throughs, or maybe there's a way to have the hook but no jet implemented.

eglaysher: I believe you can delete the jet registrations without deleting the hook registrations.

philipcmonk:
Re: asymptotically absurd w/o fond, I wouldn't rule out it *currently* being asymptotically absurd.  The fact that +play is called more than a million times is hard to believe.  Nock is dangerously becoming fast enough that you don't notice asymptotic absurdity for a while.

philipcmonk:
For example, IIUC +fond's recursion doesn't hit the cache, so a subject of `[1 ..zuse]` won't benefit from previously checking in `..zuse`.  If true, it wouldn't shock me if we got massive gains just be moving the cache check inside the recursion.

pkg/urbit/jets/tree.c
joemfb: You can definitely have the hook without the jet.

But i also think there's no problem in removing the hook (I think there would be if we were running under hashboard, but we're not). I think the only thing the `u3j_hood*` declarations are used for with fast-hint registration is lazy name->axis lookup when you actually try to resolve a hook. (And I think this because the hook declarations in `tree.c` are pretty stale ...)

@frodwith, can you confirm?

pkg/urbit/noun/manage.c
joemfb: All the current arguments are just flags -- they get packed into a bitmask and sent to the worker. But there is one argument with a value on `ipc-redux`, and I just hacked it in, see: dfa818a069dd6ba201b4cfbe2e171275e9f044af.

pkg/urbit/jets/tree.c
joemfb: Turns out I was wrong and @eglaysher is right -- I obviously misunderstood this part of the system. Please restore these hooks.

philipcmonk: Hmm, strangely I can run this on ~wicdev and build a pill with no issues.  Regardless, better to leave the hooks in.

pkg/urbit/noun/manage.c
philipcmonk: I was able to boot with a brass pill

pkg/urbit/jets/tree.c
joemfb: When you try that, do the existing `+ut` jets still fire?

joemfb: I just talked to @frodwith some more about this, and I now recant my previous recantation. The combination of hashboard and fast-hints is fairly complicated and difficult to reason about it. And our implementation of fast-hints was glued onto hashboard after the fact; this area definitely needs some work in light of @frodwith's more up-to-date designs.

My earlier claim, that the presence/absence of the hooks is irrelevant (except lazily, in cases where they are actually called), is also not quite right. It looks like it's the hooks in your parent core that are used to match a battery to an already registered driver. And registrations are updated by fast hints. So @eglaysher and I were each about half right.

philipcmonk: So best to just leave in the hooks?

joemfb: Yes. But it doesn't really matter either way -- getting things working in hashboard again would be a significant chunk of work regardless.

pkg/urbit/noun/manage.c
philipcmonk: As long as it's not optional, I guess I can just add it as `argv[4]`, and merging that with ipc-redux will just be incrementing a few indices.

philipcmonk:
Okay, let's not release the mapsize changes until they're safe.

I restored the hooks and added a -C argument to control the memo cache.  If this looks good, we can release it ASAP.

pkg/urbit/jets/e/jam.c
philipcmonk: Cool, I switched it

philipcmonk:
This has the same CI error as several other branches, something to do with permissions and cachix.  Do we have a solution for that?  Does it interfere with uploading the static binaries?

joemfb:
It's fixed on master, you'll just need to merge back or rebase.

joemfb:
No CI deployments take place if the build fails.

pkg/urbit/noun/manage.c
```diff
@@ -496,7 +497,7 @@ _pave_south(c3_w* mem_w, c3_w siz_w, c3_w len_w)
 static void
 _pave_parts(void)
 {
-  u3R->cax.har_p = u3h_new();
+  u3R->cax.har_p = u3h_new_cache(u3_Host.ops_u.hap_w);
```
joemfb, line 500: u3_Host is a `vere` global, it really shouldn't be used in u3. But this isn't a new problem (the json tracing follows the same pattern), so it doesn't need to be fixed here.

philipcmonk:
This is already up-to-date with master

pkg/urbit/worker/main.c
```diff
@@ -1035,6 +1036,7 @@ main(c3_i argc, c3_c* argv[])
   */
   {
     sscanf(wag_c, "%" SCNu32, &u3C.wag_w);
+    sscanf(hap_c, "%" SCNu32, &u3_Host.ops_u.hap_w);
```
joemfb, line 1039: It'd be nice to have a warning here if the value is below some threshold (but i'm not sure what that threshold should be).

joemfb:
Oh, sorry. CI has broken a couple of times recently, for different reasons. The last one has to do with updates to the official nix environment in travis-ci, and was fixed in #3046 (see https://github.com/urbit/urbit/pull/3046/files#diff-354f30a63fb0907d4ad57269548329e3R15). Probably easiest to just duplicate the change here.

In `before_install`:

```
- echo "trusted-users = root travis" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon
```

philipcmonk:
Cool, I'm going to release this now.

PR #3046, opened 2020-06-23
king: Build release tarballs with both king and vere.

eglaysher:
This adds a new build stage called combine which takes the results
of the previous compile builds and packages them up into one
release tarball per platform.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,15 @@
+stages:
+  - compile
+  - combine
 jobs:
   include:
-  - os: linux
+  - stage: compile
+    os: linux
     language: nix
     nix: 2.3.6
     before_install:
       - git lfs pull
+      - echo "trusted-users = root travis" | sudo tee -a /etc/nix/nix.conf && sudo pkill nix-daemon
     install:
       - nix-env -iA cachix -f https://cachix.org/api/v1/install
     script:
@@ -14,7 +19,8 @@ jobs:
       - make release
       - sh/ci-tests
 
-  - os: linux
+  - stage: compile
+    os: linux
     language: generic
     env: STACK_YAML=pkg/hs/stack.yaml
     cache:
@@ -31,7 +37,8 @@ jobs:
       - stack test
       - sh/release-king-linux64-dynamic
 
-  - os: osx
+  - stage: compile
+    os: osx
     language: generic
     sudo: required
     env: STACK_YAML=pkg/hs/stack.yaml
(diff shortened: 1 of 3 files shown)


eglaysher:
Hi, this is an attempt to move forward on releasing king-haskell by placing the urbit-king binaries in the release tarball. This is done by building a new stage, which aggregates the build artifacts from the previous stages.

.travis.yml
```diff
@@ -1,10 +1,18 @@
+stages:
+  - compile
+  # Don't run the combine stage in pull requests, because deploy is disabled there.
+  - name: combine
+    if: env(TRAVIS_PULL_REQUEST) = false
```
joemfb, line 5: This doesn't appear to be working. This conditional DSL appears to only have limited env access (per https://docs.travis-ci.com/user/conditions-v1). But, per that same link, I think `type != pull_request` should work.

eglaysher: I changed it, but where did you see this not working? I looked at the build in this PR and it looks like it was excluded properly.

joemfb: It didn't run here: https://travis-ci.org/github/urbit/urbit/builds/701432802. Looks good in the latest: https://travis-ci.org/github/urbit/urbit/builds/701439095.

joemfb:
For posterity:

As of these changes, two tarballs are built on every commit:

```
$ tar -tvf vere-596f089e2014db44ed6eeb52edefcb64af86dda7-darwin.tgz 
drwxrwxr-x  0 travis travis      0 Jun 23 14:59 vere-596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/
-r-xr-xr-x  0 travis travis 6829392 Jun 23 14:59 vere-596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/urbit
-r-xr-xr-x  0 travis travis 6628896 Jun 23 14:59 vere-596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/urbit-worker
$ tar xzvf 596f089e2014db44ed6eeb52edefcb64af86dda7-darwin.tgz 
x 596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/
x 596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/urbit
x 596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/urbit-king
x 596f089e2014db44ed6eeb52edefcb64af86dda7-darwin/urbit-worker
```

The naming conventions per commit aren't great, but that's a pre-existing issue.

On *every* tag (NB: this also applies to the OS1/interface suite of tags), the same two tarballs will be built, and named with the tag (`vere-urbit-v.10.6-darwin.tgz` and `urbit-v.10.6-darwin.tgz`).

.travis.yml
eglaysher: I see. I didn't understand that there's a difference between the tryjob for the commit itself and the tryjob for the merge commit associated with this PR.

Issue #3035, opened 2020-06-19
Debug dashboard partially broken


Fang-:
- "Interactive" actions, like subscribing to an app's `verb`, or poking spider, fail because `window.urb is undefined`.
- Tried loading ames details for ~dopzod, but the page crashed with `JSON fetch on /~debug/ames/peer/dopzod.json failed: TypeError: t.key.includes is not a function`

matildepark:
First is fixed by #3048. Honestly unsure about the latter one so far, though — it does seem to work on ~haddef when I go directly to it.

Fang-:
I think I might already have fixed that in #2951, actually? But I don't see that change on master, could it have gotten lost somewhere along the merge/release process?

matildepark:
It wasn’t lost, release/next-userspace just hasn’t been merged into master and released since May 2.

matildepark:
With both PRs now in master, closing this.

PR #3048, opened 2020-06-24
dbug: init new channel on start

matildepark:
Landscape rearchitecture removes `window.urb` from userspace. This initiates it as a fresh channel at dbug start.

Partially addresses #3035.


matildepark:
Don't know why CI is failing — all tests passed when you look at the log.

PR #3049, opened 2020-06-24
vere: add -X flag for running a scry

Fang-:
Tries to get a scry result from a /vanecare/desk/path formatted path,
and jams the result to disk (.urb/put/) if it succeeds.
Optionally use -Y to specify a name for the resulting file.

Wrote this as an exercise, but seems potentially useful for getting data out of piers without booting them.

### DIFF
--- a/pkg/urbit/vere/pier.c
+++ b/pkg/urbit/vere/pier.c
@@ -949,25 +949,26 @@ _pier_on_scry_done(void* vod_p, u3_noun nun)
   else {
     u3l_log("pier: scry succeeded\n");
 
-    u3_atom jam = u3qe_jam(res);
-    c3_w siz_w = u3r_met(3, jam);
-    c3_y* dat_y = c3_malloc(siz_w);
-    u3r_bytes(0, siz_w, dat_y, jam);
-
-    c3_c* nam_c = u3_Host.ops_u.puk_c;
-    if (!nam_c) {
-      nam_c = "scry";
+    c3_c* pac_c = u3_Host.ops_u.puk_c;
+    if (!pac_c) {
+      pac_c = u3_Host.ops_u.pek_c;
     }
+
+    u3_noun pad;
+    {
+      u3_noun pax = u3do("stab", u3i_string(pac_c));
+      c3_w len_w = u3kb_lent(u3k(pax));
+      pad = u3nt(c3_s4('.','u','r','b'),
+                 c3_s3('p','u','t'),
+                 u3qb_scag(len_w - 1, pax));
+      u3z(pax);
+    }
+
     c3_c fil_c[2048];
-    snprintf(fil_c, 2048, "%s/.urb/put/%s.jam", pir_u->pax_c, nam_c);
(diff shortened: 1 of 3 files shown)


belisarius222:
I think this will be very useful.  I'd say the default filename in .urb/put/ should be the full path of the scry request, not just "scry".

I can also see this eventually becoming able to specify a different case in the request, which would mostly be useful for reading old Clay files.

pkg/urbit/vere/pier.c
```diff
@@ -965,7 +1006,26 @@ _pier_on_lord_live(void* vod_p)
...
+    if (u3_Host.ops_u.pek_c) {
+      u3_noun pex = u3do("stab", u3i_string(u3_Host.ops_u.pek_c));
+      u3_noun car;
+      u3_noun dek;
+      u3_noun pax;
+      if ( c3n == u3r_trel(pex, &car, &dek, &pax)
+        || c3n == u3a_is_cat(car) )
+      {
+        u3m_p("pier: invalid scry", pex);
+        _pier_on_scry_done(pir_u, c3__null);
```
joemfb, line 1018: This should be `u3_nul` (ie, `~` -- `c3__null` is a symbol(ie, `%null`).

Fang-:
Good idea re filepath, @belisarius222. It's in 2976b53. Almost considered writing my own recursive `mkdir`, but found that `u3_walk_save` already does this for you. Wrestled with its interface for a bit, but pretty sure I got it right.

@joemfb non-trivial change, please take another look. (:

pkg/urbit/vere/pier.c
```diff
@@ -935,6 +935,48 @@ _pier_on_lord_bail(void* vod_p)
...
+
+    c3_c* pac_c = u3_Host.ops_u.puk_c;
+    if (!pac_c) {
+      pac_c = u3_Host.ops_u.pek_c;
+    }
+
+    u3_noun pad;
+    {
+      u3_noun pax = u3do("stab", u3i_string(pac_c));
+      c3_w len_w = u3kb_lent(u3k(pax));
```
joemfb, line 960: This could just be `u3qb_lent(pax)`, which is equivalent but doesn't thrash the count.

Fang-: Huh, I could swear I ran into that not existing, but it's there. Maybe I got confused. Will fix!

Fang-:
idk, the loc didn't change much, and the complexity here is largely due to `u3_walk_save`'s interface. As you suggested in chat, I'll probably take a stab at writing a wrapper for that, to resolve the complexity creep here.

The command itself indeed only runs a single scry, but you could imagine using this for writing scripts that do `y` scries and pluck out specific sets of data or whatever.  
Maybe that's assuming too much? Once we got complexity down it's probably not too big a deal to just leave it in, and see if this goes unused/unappreciated or not. Can always dumb it down again later, there's no dependencies here.

Fang-:
Merging. As discussed out of band, will do the refactoring work in its own PR.

PR #3050, opened 2020-06-24
behn, vere: some improvements

Fang-:
Primarily:

- Rewrites behn.hoon's scry interface to use cares, not misuse the desk, provide more endpoints, and expose ducts only on an explicitly debug-oriented endpoint.
- Rewrites behn.c's "backstop" logic to scry into behn for the next timer instead of setting one for ten minutes into the future. (But still falls back to doing so if the scry doesn't succeed.)

Note that f714d90 touches arvo to make `%$` be considered a valid value for the desk. `+slaw` doesn't parse the empty string as a `%tas`. Arguably that's a bug in `+slaw`, but it's parsing behavior for `%tas` is a bit funky already and I'd rather not make the broader-scope change right now.

@joemfb I made some notes during our call yesterday about some oddities in behn.hoon's implementation, but chose not to do the fixes for those right now. They seem better off cushioned in other work. Happy to still do the safe ones right now if you think it's worth it.

(Is the convention for vere commits just the `vere:` prefix always? `behn:` would get confusing wrt arvo vs runtime changes... Maybe `behn.c:`?)

### DIFF
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -430,11 +430,14 @@
       [~ ~]
     :^  ~  ~  %noun
     !>  ^-  (list @da)
-    %-  zing
-    %+  turn  (tap:timer-map timers)
-    |=  [date=@da q=(qeu duct)]
-    ?:  (gth date u.til)  ~
-    (reap ~(wyt in q) date)
+    =/  tiz=(list [date=@da q=(qeu duct)])
+      (tap:timer-map timers)
+    |-  ^-  (list @da)
+    ?~  tiz  ~
+    ?:  (gth date.i.tiz u.til)  ~
+    %+  weld
+      (reap ~(wyt in q.i.tiz) date.i.tiz)
+    $(tiz t.tiz)
   ==
 ::
 ++  stay  state
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -389,18 +389,53 @@
 ++  scry
   |=  [fur=(unit (set monk)) ren=@tas why=shop syd=desk lot=coin tyl=path]
   ^-  (unit (unit cage))
+  ::  only respond for the local identity, %$ desk, current timestamp
   ::
-  ?.  ?=(%& -.why)
+  ?.  ?&  =(&+our why)
+          =([%$ %da now] lot)
+          =(%$ syd)
+      ==
     ~
-  ?.  ?=(%timers syd)
-    [~ ~]
-  =/  tiz=(list [@da duct])
+  ::  /bx/debug/timers  (list [@da duct])  all timers and their ducts
+  ::  /bx/timers        (list @da)         all timer timestamps
+  ::  /bx/timers/next   (unit @da)         the very next timer to fire
+  ::  /bx/timers/[da]   (list @da)         all timers up to and including da
+  ::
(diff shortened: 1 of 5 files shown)


philipcmonk:
Is there any way that `_behn_scry_cb` might not be called?  For example, are we sure we get a response if a nondeterministic error happens during the scry?  The most important thing with timers is to make progress.

I suppose the idea is for this to eventually replace `%doze` altogether`?

belisarius222:
@joemfb and I discussed Behn some today.  We've been thinking it would be nice if none of the I/O drivers required runtime scrying to function properly, using scrying purely for optimization purposes.

pkg/arvo/sys/vane/behn.hoon
```diff
@@ -389,18 +389,53 @@
...
+      [%timers @ ~]
+    ?~  til=(slaw %da i.t.tyl)
+      [~ ~]
+    :^  ~  ~  %noun
+    !>  ^-  (list @da)
+    %-  zing
+    %+  turn  (tap:timer-map timers)
+    |=  [date=@da q=(qeu duct)]
+    ?:  (gth date u.til)  ~
+    (reap ~(wyt in q) date)
```
joemfb, line 437: Since this interface is for reading batches, the full traversal seems like a bit too much. Those timers get really big!

Once this is merged with @loganallenc's ordered-map range-queries, that'd probably be the best way to implement this. In the meantime, it'd be nice to have (four-letter) `+take-while` (and `+drop-while`, while we're at it). Barring that, I think i'd just manually traverse the list up to `til`.

pkg/arvo/sys/vane/behn.hoon
```diff
@@ -389,18 +389,53 @@
...
     ~
-  ?.  ?=(%timers syd)
-    [~ ~]
-  =/  tiz=(list [@da duct])
+  ::  /bx/debug/timers  (list [@da duct])  all timers and their ducts
+  ::  /bx/timers        (list @da)         all timer timestamps
+  ::  /bx/timers/next   (unit @da)         the very next timer to fire
+  ::  /bx/timers/[da]   (list @da)         all timers up to and including da
+  ::
+  ?.  ?=(%x ren)  ~
```
joemfb, line 404: This seems right and proper to me. (I'm tempted to ask you to implement `%by` as well, but that's probably beest saved for a more systematic scry-cleanup effort.)

joemfb:
@philipcmonk, any crash in an external scry is currently (and has always been, as far as I can tell) an assertion failure in the runtime (specifically, in the serf). So the callback won't be called, but neither will anything else ...

This is not any kind of serious solution -- deterministic crashes in a scry handler are arguably a bug, but non-deterministic crashes are always possible (at least SIGINT and OOM). I don't think the caller/callback should get the trace though; I'm inclined to say that crashes should just be converted to `~` (ie, no response, already encompassing invalid paths and internal blocking). Some could certainly be retried, but I hesitate to proliferate that logic in every call-site.

joemfb:
@Fang-, I'm reverting the changes to the i/o driver on IPC redux. There are subtle interactions between the ordering of the scry requests, events, and effects, which manifests as a) too many scry requests in normal operation, and b) a scry loop on first boot.

We can and will sort this out and revive these changes, but it requires more attention than I have to give at the moment, as I'm prepping that branch for release ASAP.

Fang-:
Fair game! Let's talk about this next week.

PR #3054, opened 2020-06-26
libsigsegv: disable stack vma check (or: Make Linux Fast)

philipcmonk:
This is a draft PR pending testing stack overflow handling and nix hygiene.  Can someone who knows nix take a look at this?  @brendanhay @eglaysher 

This patches libsigsegv to not check the stack vma on Linux, since that
involves reading procfs, and we make very heavy use of sigsegv.  This
eliminates most of urbit's performance discrepancy between Linux and
MacOS.  These are the benchmarks used; note this is a local MBP vs a
cloud Linux server, and the MBP is almost certainly faster hardware.

We take two benchmarks, one of which decrements 10 million times and the
other simply allocates 125MB of memory.  These are the results:

cpu-heavy ==  =/ n 10.000.000 |-(?~(n n $(n (dec n))))
mem-heavy == =a (bex 1.000.000.008)

macos, cpu-heavy: 6 seconds
macos, mem-heavy: 1 second

linux-before, cpu-heavy: 30 seconds
linux-before, mem-heavy: 160 seconds

linux-after, cpu-heavy 9 seconds
linux-after, mem-heavy 1.3 seconds

This represents a 3x speedup for the cpu-heavy operation and a 120x
speedup for the memory-heavy operation.

This check was used to try to distinguish stack overflow from other
forms of segmentation fault.  In the comments in src/handler-unix.c, it
describes three heuristics it uses, depending on what's available from
the OS.  In the linux-i386 case, all three are availble, so we simply
disable the slow one.  Further testing is required to determine if it
continues to acceptably distinguish stack overflows.

Not sure I'm doing the nix stuff correctly, but this works for `make`.
Need to test release and nix-shell.

### DIFF
--- a/nix/pkgs/urbit/default.nix
+++ b/nix/pkgs/urbit/default.nix
@@ -16,9 +16,14 @@ let
     exe   = ''${meta.bin} ${pkgs.lib.strings.concatStringsSep " " meta.flags}'';
   };
 
+  sigseg =
+    pkgs.libsigsegv.overrideAttrs (oldAttrs: rec {
+      patches = [ ./libsigsegv_fix.patch ];
+    });
+
   deps =
     with pkgs;
-    [ curl gmp libsigsegv ncurses openssl zlib lmdb ];
+    [ curl gmp sigseg ncurses openssl zlib lmdb ];
 
   vendor =
     [ argon2 softfloat3 ed25519 ent ge-additions libaes_siv h2o scrypt uv murmur3 secp256k1 sni ivory-header ca-header ];
--- a/nix/pkgs/urbit/libsigsegv_fix.patch
+++ b/nix/pkgs/urbit/libsigsegv_fix.patch
@@ -1,9 +1,3 @@
-source $setup
-
-tar -xf $src
-
-cd libsigsegv-$version
-patch -p1 << 'HEREDOC'
 --- a/src/fault-linux-i386.h    2020-06-25 23:46:02.099235491 +0000
 +++ b/src/fault-linux-i386.h    2020-06-25 23:45:48.679156892 +0000
 @@ -18,6 +18,7 @@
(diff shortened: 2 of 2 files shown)


philipcmonk:
Adding `alloca(10000000000)` to the jam jet (gated on a special value) correctly gave `dig: over` and didn't crash the process.  This means at least some stack overflows are correctly handled.

philipcmonk:
Compiling large things seems less affected, presumably since once it has marked a block dirty, it doesn't hit SIGSEGV until it's checkpointed.  Specifically, for `.urbit/pill +solid` on Linux I get 6:57 on v0.10.5, 5:15 on v0.10.6, and 5:02 on this branch.  This is compared to 2:50 on my Mac.  This is pretty well in line with the 6 seconds/9 seconds ratio of Mac to Linux for the cpu-heavy benchmark after these changes, which suggests this particular task was never slowed down that much by SIGSEGV, and the apparent speed difference between the OSes is likely just better hardware.

joemfb:
For the release builds, I think it'd be simplest to just edit the derivation (`nix/nixcrpkgs/pkgs/libsigsegv/default.nix `) and have it always apply the same patch. The whole package is new in our fork/subtree of nixcrpkgs, so it's probably not worth the trouble to overlay as if the package had come from upstream.

philipcmonk:
I pushed a change that makes this work for release binaries, so functionally nix is fine now.

I'll note this change only applies to Linux on i386 and x86_64.  In particular it's not enabled on arm since I don't have a device to test it on.  If someone wants to test it there, it should be just a matter of adding `#define HAVE_STACKVMA 0` to `fault-linux-arm.h` (or even `fault-linux.h`), compiling that, and testing `=a (bex 1.000.000.000)` before and after to see if you get a dramatic speedup.

philipcmonk:
This is ready for review now

brendanhay:
:shipit:

PR #3058, opened 2020-06-26
metadata-store: add cleanup utility

ixv:
Adds a utility to metadata-store to remove by group-path

Theres a problem on the mainnet where if you migrate a group to a new ship, you can get lots of old useless metadata lying around, which then can cause the frontend to misrepresent what channels are present in a group. 

This issue will have to be revisited at some point to fix the root problem. Probably better to wait for the new groups stuff to land first though. For now, this PR just gives people the ability to manually erase metadata by group-path, which can be used to solve this problem.

### DIFF
--- a/pkg/arvo/app/metadata-store.hoon
+++ b/pkg/arvo/app/metadata-store.hoon
@@ -61,9 +61,27 @@
     ^-  (quip card _this)
     ?>  (team:title our.bowl src.bowl)
     =^  cards  state
-      ?:  ?=(%metadata-action mark)
+      ?+  mark  [~ state]
+          %metadata-action
         (poke-metadata-action:mc !<(metadata-action vase))
-      (on-poke:def mark vase)
+          %noun
+        =/  val=(each [%cleanup path] tang)
+          (mule |.(!<([%cleanup path] vase)))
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/metadata-store.hoon
```diff
@@ -61,9 +61,27 @@
     ^-  (quip card _this)
     ?>  (team:title our.bowl src.bowl)
     =^  cards  state
-      ?:  ?=(%metadata-action mark)
+      ?+  mark  [~ state]
```
belisarius222, line 64: Shouldn't this fall back to `+on-poke:def` instead of a direct no-op?

ixv: Fixed, thanks,

belisarius222:
merged into ford-fusion https://github.com/urbit/urbit/pull/3060

ixv:
Thanks Ted.

Fang-:
> `%remove-group-data` might be clearer than `%cleanup`, but looks fine otherwise.

Alright, people are running this command when they shouldn't. Let's _please_ be good about naming these kinds of commands going forward. Dangerous tools should have dangerous names.

PR #3059, opened 2020-06-26
arvo: behn scry interface updates

joemfb:
This PR is just a cherry-pick of the arvo changes from #3050, retargeted at the appropriate branch (for separate OTA).


joemfb:
I'll revert the original commits on `ipc-redux` so that the vere changes can be released on a separate schedule.

philipcmonk:
Hm, looks like this didn't get merged and now has conflicts.  @Fang- or @joemfb, can you determine what should happen here?

Fang-:
As discussed with @joemfb, pulled in the ames.hoon changes from #3087 also. Should be good to go, but if either of you could take another look and stamp it that'd be nice.

PR #3060, opened 2020-06-27
Ford Fusion

philipcmonk:
This is a significant refactoring and rearchitecting of our build system and over-the-air update system.  Further documentation is necessary, but highlights include:

- The Ford vane, which was responsible for the build system, has been removed.
- A new build system has been developed in Clay, in the +ford arm.  Hereafter "Ford" will be taken to mean the portion of Clay which deals with building files and marks.
- Updates are fully atomic.  For example, if an update to hoon.hoon causes an app to fail to compile, then the commit/merge will not apply.
- Since updates are atomic, the Clay write queue has been removed.
- The new build system is built on three types of requests: build a file, build a mark object, and build a mark conversion gate.  These may be requested via individual request, subscription request, or scry from Clay just like the contents of a file, using the `%a`, `%b`, and `%c` marks respectively.
- Clay `%warp` responses are now synchronous for local desks and asynchronous for foreign desks.
- The `%base` desk is being retired.  See below for the new pattern.
- Ford's cache is simplified.  Now its only cache is that all files at the current revision for each of our desks is cached once built, until one of its dependencies changes.
- Reduces the total lines of code in the vanes by 1/3, from 21k to 14k
- Various non-Ford changes, including:
  - Landscape stores all app data in a single global store on the front end
  - Chat deduplicates its data, reducing memory usage
  - Eyre clears expired sessions/cookies/channels more cleanly
  - Eyre adds a logout endpoint, exposed in the debug dashboard
  - Eyre respects "forwarded" header from localhost for proxying
  - Publish sends less data on kick/resubscribe
  - Debug dashboard fixed in various cases
  - Shoe allows you to auto-run commands as you enter them
  - `+slab` respects core variance
- Various fixes and improvements

Re: new desk pattern, updates used to follow this pattern:

```
    ~zod/base ---------> ~zod/home
        |
        v
    ~zod/kids
        |
        v
   ~marzod/base -------> ~marzod/home 
        |
        v
   ~marzod/kids
        |
        v
~wicdev-wisryt/base
        |
        v
~wicdev-wisryt/base ---> ~wicdev-wisryt/home
```

Now, they follow this pattern:

```
    ~zod/home
        |
        v
    ~zod/kids ---------> ~marzod/home
        |
        v
   ~marzod/kids -------> ~wicdev-wisryt/home
        |
        v
~wicdev-wisryt/kids
```

+trouble still reports a `%base-hash`, which is the hash of the most recent commit which we have downloaded/cached of our parent's `%kids` desk.  This performs roughly the same function as the old base hash.


pkg/arvo/sys/arvo.hoon
```diff
@@ -968,289 +1017,297 @@
...
+            (poke:soul now ovo)
+  ::
+  ++  wish  |=(* (^wish ;;(@ta +<)))             ::  22
+  --
+::  $pram: upgradeable state
+::  +dram: upgrade $pram, compatible
+::
++$  pram
+  $~  [%345 nub=*debt:le:part]
+  $%  [%345 nub=debt:le:part]
```
joemfb, line 1054: Kelvin versions are great. I'm a uuuge fan of kelvin versions. But this is not a valid kelvin version, as it violates the stack invariants (see `%zuse`). I humbly request that this be replaced with a date constant, until such time as we can normalize the versioning across the whole kernel.

ohAitch:
Is it possible to |mount e.g. ~marzod/kids? Presumably %base-hash is just
"whatever |syncs says /=home= is pulling from"?

ohAitch:
Have we generally established the undesirability of e.g. stars running apps
they're not redistributing to all child planets? Obviously they don't at
present, but in general present use of the desk system is fairly minimal
and approximately isomorphic to "two layer filesystem, 'github' and 'local
changes'"

philipcmonk:
+trouble might be hard coded to sponsor right now, you're right it should
vary with who we're actually synced from.

The new pattern still has stars' %home desks out of the distribution chain.
~zod's home also doesn't need to be in the chain, and in general you may
want to commit to those two desks separately. Unless you're making
sufficiently breaking hoon.hoon changes, you can just reverse the direction
of that sync, ie kids -> home.

I haven't tried mounting a foreign desk, if it doesn't work we should make
it work. Of course a workaround is merging it to a new local desk.

belisarius222:
@pilfer-pandex I redid the fix-slab work, which had two bugs: if the name was not found in the type, `+slab` crashed instead of producing false, which I fixed by replacing `+find` with `+fond`; and its result was negated because the `?=` was looking for a `%|` instead of a `%&`.

@joemfb The Clay test was failing because of the `+slab` problems, so now it passes again.  I also switched Arvo to use a date for its version instead of the bogus Kelvin.

sigilante:
@belisarius222 @philipcmonk What does this do to Ford runes?  Can you tell me where to look for the source of those and I'll document?

belisarius222:
@sigilante https://github.com/urbit/docs/issues/905#issuecomment-653233698

Here's a decent entry point into the source for how the Ford runes get parsed and run:
https://github.com/urbit/urbit/blob/1adca22298a93f0a4b2f1a35701f445c4fc271ca/pkg/arvo/sys/vane/clay.hoon#L754

PR #3064, opened 2020-06-29
ipc redux

joemfb:
This PR replaces the urbit runtime IPC protocol, and significantly refactors (approaching "rewrites") the implementations on both ends.

The new IPC protocol:

- avoids head-of-line blocking
- includes full error-handling for all events
- supports arbitrary reads from the urbit namespace.

The vere "king" (i/o & event persistence, the `urbit` binary) and "serf" (compute & state persistence, the `urbit-worker` binary) have been rewritten to support and take advantage of the new protocol. New features include:

- improved event scheduling, with per-source (i/o driver) prioritization
- a capped queue (currently 1K) for ames packets
- the ncurses/terminfo dependency has been removed (run static `urbit` binaries out of your `$PATH`!)
- periodic ames crash reports (no more unconditional `arvo: replacement event (for %hear) failed`)
- arbitrary scry with `-X` and `-Y`
- status info with SIGINFO (where supported) or SIGUSR1
- partial-replay and save a "rock" ("portable snapshot", from `|pack`) with `-n`
- `load state from a "rock" with -r`

The clean-room rewrite of the "king" process in haskell ("king-haskell") has been updated to support the new IPC protocol, and is being moved to a pre-release status -- it's now included in every release bundle as the `urbit-king` binary. There are additional improvements here as well (most notably, some initial support for running multiple piers within one king process), but I'll leave the exegesis to others, as I was involved only the protocol upgrade. (NB, the PR includes the changes from the `khrc` branch.)

Additional notes:

The low-level IPC atom-framing implementation (`vere/newt.c`) has been rewritten. The result is simpler and more efficient with regards to allocation. A bug wherein completed messages were not immediately delivered has been fixed -- this is a likely candidate for (at least some of) the periodic reports of process hangs or non-responsiveness.

The `u3i_*` functions (`noun/imprison.c`, the higher-level interface to the allocator) have been refactored for clarity, with `u3i_chubs()` specifically rewritten for correctness (can't use malloc!) and efficiency.

A latent-but-substantial bug in `u3r_mug_words()` has been fixed. That function produced incorrect results for any input of more than one word, but was fortunately never being called with such an input.

Core dumps have been disabled from the vere king, since it unconditionally maps 1 TB of memory for the LMDB event log. MacOS users/developers can resort to the Console.app crash reports, but linux denizens simply must use gdb. (To smoothly run urbit in gdb, ignore SIGSEGV, and disable "tty assumptions" with the `-t` argument.)

Status:

Such a large set of changes is difficult to review, and risky to release. These changes run happily under ASan and Valgrind (with a smaller LMDB mapping). The only known memory leaks are one-time allocations, either ignored (varies CLI arguments) or disposal-deferred (the HTTP servers) on shutdown. Additionally, these changes (all but the last few commits) have been running on several high-traffic infrastructure nodes (Tlon-operated galaxies/stars) for a couple of weeks without (serious) incident.


joemfb:
@philipcmonk, bump for review.

PR #3066, opened 2020-06-29
chat fe: don't embed webms as images

Fang-:
Fixes this:

image


matildepark:
Target `release/next-userspace`, I just merged master back into it.

Fang-:
Oops, good catch.

PR #3067, opened 2020-06-30
spa: restore previous line-height, padding

matildepark:
Setting a line-height across the board of 1.5 was patched during SPA development with a few padding tweaks here and there; but in general we decided to just restore the inconsistent line-height we had before and set a consistent line-height during the Indigo refactor work in progress in FE.


matildepark:
I wanna give this an in-depth look one last time before merging ...

PR #3070, opened 2020-06-30
bitcoin: connects wallet client to local bcoin node

yosoyubik:
@matildepark This is still broken; I have just updated this with the branch where I had added support to connect to a local bcoin node.


matildepark:
OK, I'll also merge master into the bitcoin branch while I'm at it.

matildepark:
I think I need to undo the work I put on this and start over on it — keeping its state local, not globalised, while still matching the SPA architecture (api / store / subscription folders).

PR #3071, opened 2020-06-30
nix: handle vere version in 'urbit' derivation

jtobin:
Previously, one needed to edit the URBIT_VERSION variable in pkg/urbit/configure before building in order to specify a new version of Vere.  From this change forward, one should instead edit the "version" attribute in nix/pkgs/urbit/default.nix.  The URBIT_VERSION variable is then simply assigned the value of that attribute when building.

The motivation here is to reflect the Vere version in its derivation as well.  By specifying 'pname' and 'version' attributes, the stdenv builder will automatically name the derivation 'pname-version'.  This leads to derivations named e.g. 'urbit-0.10.7' and 'urbit-debug-0.10.7', rather than the plain 'urbit' and 'urbit-debug'.

cc @philipcmonk

### DIFF
--- a/nix/pkgs/urbit/default.nix
+++ b/nix/pkgs/urbit/default.nix
@@ -29,8 +29,11 @@ let
     [ argon2 softfloat3 ed25519 ent ge-additions libaes_siv h2o scrypt uv murmur3 secp256k1 sni ivory-header ca-header ];
 
   urbit = pkgs.stdenv.mkDerivation {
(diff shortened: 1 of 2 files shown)


jtobin:
(CI appears to be failing due to an environment config issue. )

jtobin:
I'm going to close this one -- the end result is desirable, but I don't think what I've done here is quite the right approach anyway.  Will revisit.

PR #3072, opened 2020-06-30
bitcoin: SPA refactoring

yosoyubik:
@matildepark I managed to adapt the wallet to use the same patterns as the others Landscape apps, so I don't think you'll have to undo your previous work. There are couple of UI issues (loading the tile, custom CSS styles...), but I can sync the SPV node on the browser, derive the xpubkey and store it on the ship, so that should be good to get you going for now.


pkg/arvo/app/wallet.hoon
```diff
@@ -383,34 +359,34 @@
...
-      [%'~bitcoin' %img @t *]
-    =/  name=@t  i.t.t.site.url
-    =/  img  (~(get by bitcoin-png) name)
-    ?~  img
-      not-found:gen
-    (png-response:gen (as-octs:mimes:html u.img))
-  ::
-      [%'~bitcoin' *]  (html-response:gen index)
-  ==
+:: ++  poke-handle-http-request
```
loganallenc, line 362: Note: This entire arm should be removed and replaced with %file-server.

pkg/arvo/app/wallet.hoon
```diff
@@ -11,107 +11,107 @@
...
+:: /=  style
+::   /^  octs
+::   /;  as-octs:mimes:html
+::   /:  /===/app/bitcoin/css/index
+::   /|  /css/
+::       /~  ~
+::   ==
+:: /=  bitcoin-png
+::   /^  (map knot @)
+::   /:  /===/app/bitcoin/img  /_  /png/
```
loganallenc, line 114: Note: these can all be removed, not just commented out.

pkg/arvo/app/landscape/index.html
```diff
@@ -21,6 +21,16 @@
   </head>
   <body>
     <div id="root"/>
+    <script src="/~wallet/js/bcoin.js"></script>
```
loganallenc, line 24: Note: these scripts should just be included in the main App bundle, not loaded separately

yosoyubik: I'll look into it but this has been lingering since the old build flow with gulp. I couldn't get that to work so adding them separately was my only successful attempt, but maybe it'll be a bit easier with webpack. fyi, these files were created with a [bundler](https://github.com/chjj/bpkg) used by bcoin so there might be some issues integrating it with webpack.

pkg/arvo/app/wallet.hoon
yosoyubik: whoops, I had collapsed all these comments and forgot to delete them. thanks!

matildepark:
Not sure how we got into such a conflict tussle here ...

matildepark:
Well, while we're waiting on a merge conflict (I think the issue is that you completely rebased and force push'd instead of just adding a 'addressing review comments' commit), you can just delete tile.js, and add a poke to launch on-init. It [looks like this](https://github.com/urbit/create-landscape-app/blob/master/full/urbit/app/smol.hoon#L15) but the second, `''` is the location of the image file. See launch.hoon for more examples.

PR #3076, opened 2020-06-30
kiln: Small fixes

philipcmonk:
Produces conflict list on mate-conflict, and also makes +trouble look at where we're actually getting OTAs from instead of guessing that it's our parent.  This still crashes if you haven't received any clay updates from them.

### DIFF
--- a/pkg/arvo/app/hood.hoon
+++ b/pkg/arvo/app/hood.hoon
@@ -39,7 +39,14 @@
   [d this]
 ::
 ++  on-leave  on-leave:def
-++  on-peek   on-peek:def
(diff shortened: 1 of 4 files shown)


belisarius222:
Why does this crash?  I'd expect it to just produce a base-hash of 0 if we haven't had an OTA yet.

philipcmonk:
It crashes because we crash on any request to a ship we don't have any clay data for.  It might be a pain to refactor clay's scry handling, or maybe there's an easy way

belisarius222:
Ok.  We can address that issue separately, then.

PR #3078, opened 2020-07-01
bitcoin: fix bug when deriving a new @uc for remote ships

yosoyubik:
@matildepark I tested a full flow here (sync node from seed, request address, derivation, payment, balance update) so it should have most of the necessary stuff for the initial UI work.

### DIFF
--- a/pkg/arvo/app/wallet.hoon
+++ b/pkg/arvo/app/wallet.hoon
@@ -71,13 +71,6 @@
           (handle-json:bc !<(json vase))
         [cards this]
       ::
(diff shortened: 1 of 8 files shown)


matildepark:
My webpack chokes on the bcoin js libs now ... I wonder if there's npm packages for these.

yosoyubik:
Yeah, I'd really like to get that fixed. bcoin is recommended to be included using git+https from the master branch, but I never got that to work with gulp's build flow. I just went with what it was faster to have it up and running. I hope that webpack is easier to integrate with this.

PR #3087, opened 2020-07-02
ames.c: stateless forwarding

Fang-:
Opening as draft because there's non-trivial remaining work (see below), but seems good to start getting some eyeballs on this.

This implements "stateless forwarding" in the ames io driver. If we receive a sane-looking packet, and we are not the intended receiver, we attempt to forward it without processing it as an event.

We accomplish this by scrying into ames for a `/peers/~receiver/forward-lane`. If the scry succeeds, and the resulting `(unit lane)` holds a value, we forward the packet onto that lane.

Considering most of the work that galaxies do is forward packets, this should help boost their performance and reduce the growth of their event legs.

Remaining work, as listed among the TODOs:
- [x] rewrite the scry endpoint in ames,
  - RFC. The current logic feels a bit nasty, and doesn't obviously match `+send-blob`. Perhaps the right thing to do here is to rewrite `+send-blob` to be more explicit about the kind of routing it's _actually_ doing, and split out the route discovery into a function that can be called by the scry logic too.
- [x] implement a cap on the amount of pending to-forward packets we hold,
- [x] properly clean up pending packets on-exit,
- [x] handle version mismatches in the sponsee case,
  - RFC. On livenet we just drop those, and I'm a bit hesitant to blow up scope of this PR by starting to scry for sponsorship chains etc. Would we be ok with punting on this?
- [x] track counters around filtering & forwarding packets,
- [ ] test for memory leaks more aggressively,
  - Address sanitizer no longer complains, but I want more certainty. I haven't been able to get Valgrind to play nice, I may want someone to pair with me on this.
- [x] figure out why recent ipc-redux changes made scry results come back as `[~ %noun ~ res]` instead of `[~ %noun res]`,
- [x] remove some items from the `u3_ames` struct, now that it stores a pointer to the `u3_pier`.

And ultimately:

- [ ] Run on a livenet ship or two for a while and see what happens.


matildepark:
huge

Fang-:
Ping @philipcmonk @belisarius222 for the RFCs here.

pkg/arvo/sys/arvo.hoon
```diff
@@ -191,7 +191,7 @@
   ?.  ?=({@ @ @ @ *} u.pux)  ~
   =+  :*  hyr=(slaw %tas i.u.pux)
           fal=(slaw %p i.t.u.pux)
-          dyc=(slaw %tas i.t.t.u.pux)
+          dyc=?~(i.t.t.u.pux (some %$) (slaw %tas i.t.t.u.pux))
           ved=(slay i.t.t.t.u.pux)
```
belisarius222, line 195: Won't this line crash if `i.t.t.u.pux` is `~`?

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -691,35 +691,65 @@
...
+      ::TODO  refactor wrt +send-blob to share routing logic
+      ::  find lane for u.who, or their galaxy
+      ::
+      :^  ~  ~  %noun
+      !>  ^-  (unit lane)
+      =/  ship-state  (~(get by peers.ames-state) u.who)
+      ?.  ?=([~ %known *] ship-state)
+        ~
+      =/  peer-state  +.u.ship-state
+      ?.  =(~ route.peer-state)  ::NOTE  avoid tmi
```
belisarius222, line 740: Why do we need to avoid TMI here?  I don't see `route.peer-state` used anywhere else.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -691,35 +691,65 @@
...
+    =/  who  (slaw %p i.t.tyl)
     ?~  who  [~ ~]
     ?~  peer=(~(get by peers.ames-state) u.who)
       [~ ~]
-    ``noun+!>(u.peer)
+    ?+  t.t.tyl  [~ ~]
+      ~  ``noun+!>(u.peer)
+    ::
+        [%forward-lane ~]
+      ::TODO  (list lane) instead?
```
belisarius222, line 730: I'm concerned that a list of lanes could be used for amplification attacks.  Relay attacks are bad enough, but at least R0 isn't above 1.

pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +550,112 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
 
+/* _ames_put_packet(): add packet to queue, drop old packets on pressure
+*/
+static void
+_ames_put_packet(u3_ames* sam_u,
+                 u3_noun  msg,
+                 u3_noun  lan)
+{
+  u3_noun wir = u3nc(c3__ames, u3_nul);
+  u3_noun cad = u3nt(c3__hear, u3nc(c3n, lan), msg);  // trnasfer, unchanged
```
belisarius222, line 561: typo: trnasfer -> transfer

pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +550,112 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
+      u3k(lon); u3k(bod);
+      u3z(old);
+    }
+
+    //  only replace the lane if it was ~
+    //
+    if (u3_nul == lon) {
+      u3z(lon);
+      lon = u3nt(u3_nul, c3n, u3k(pac_u->ore));
+      //TODO  ore is always opaque lane. what about the sender=galaxy case?
```
belisarius222, line 596: ames.hoon ignores galaxy ip/port lanes that vere sends it, so I'm not sure it matters

pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +550,112 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
+      nal = c3y;
+    }
+
+    pac_u->bod_u.con = u3ke_jam(u3nc(lon, bod));
+  }
+
+  //  if we updated the origin lane, serialize the body to update the mug hash
+  //
+  //TODO  refactor into serialize_packet?
```
belisarius222, line 605: It does seem suboptimal to serialize the packet twice.  If we don't change the lane, we shouldn't have to serialize at all, right?  Just re-send the packet we got ... or do we not store the original bytes we receive?  It's probably not really a performance problem, but the code duplication is suspect.

pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +550,112 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
+
+    pac_u->hed_u.mug_l = u3r_mug(bod) & ((1 << 20) - 1);
+
+    c3_free(bod_y);
+    u3z(bod);
+  }
+
+  pac_u->sam_u->fow_d++;
+  if ( 0 == (pac_u->sam_u->fow_d % 1000) ) {
+    u3l_log("ames: forwarded %" PRIu64 " total\n", pac_u->sam_u->fow_d);
```
belisarius222, line 624: This seems like it would print a lot on a live infrastructure node.  Maybe decrease the frequency?

pkg/arvo/sys/arvo.hoon
Fang-: I'm not sure I understand. The `?~` is handling the `~` case explicitly, how would it crash?

pkg/arvo/sys/vane/ames.hoon
Fang-: It's used implicitly by the `$(peer-state +.u.next)` a couple lines down.

pkg/arvo/sys/vane/ames.hoon
philipcmonk: IMO a list of lanes is definitely going to be desirable at some point, mainly so you can say "i'm reachable at these different locations", where those locations might be different ip/port (in case of load balancer) or protocol than the outgoing.  R0 is still <= 1 as long as those destinations aren't themselves forwarding

pkg/urbit/vere/io/ames.c
```diff
@@ -446,38 +667,146 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+      c3_w hed_w = (byt_y[0] <<  0)
+                 | (byt_y[1] <<  8)
+                 | (byt_y[2] << 16)
+                 | (byt_y[3] << 24);
+
+      hed_u.ver_y = hed_w & 0x7;
+      hed_u.mug_l = (hed_w >> 3) & ((1 << 20) - 1);
+      hed_u.sac_y = (hed_w >> 23) & 0x3;
+      hed_u.rac_y = (hed_w >> 25) & 0x3;
+      hed_u.enc_o = (hed_w >> 26) & 0x1;
```
philipcmonk, line 711: This I think should also be `>> 27`

pkg/urbit/vere/io/ames.c
```diff
@@ -446,38 +667,146 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+    else if (c3y == sam_u->see_o) {
+      c3_y sen_y = 2 << hed_u.sac_y;
+      c3_y rec_y = 2 << hed_u.rac_y;
 
-    u3_auto_peer(
-      u3_auto_plan(&sam_u->car_u,
-                   u3_ovum_init(0, c3__a, wir, cad)),
-      0, 0, _ames_hear_bail);
+      u3_noun sen = u3i_bytes(sen_y, bod_y);
+      u3_noun rec = u3i_bytes(rec_y, bod_y + sen_y);
```
philipcmonk, line 730: I think these get leaked if we're not forwarding.  This won't affect stars, galaxies, or the half of planets which are direct atoms, but it will moons and comets.

pkg/arvo/sys/arvo.hoon
belisarius222: Oh, this is really working around a weirdness in path parsing, I guess, where if there's an empty path element, `+slaw` won't parse it to a `@tas`, even though `%$` is a valid `@tas` whose numeric value is zero.  So this is fine, just a bit counterintuitive.

pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +550,112 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
+    //  only replace the lane if it was ~
+    //
+    if (u3_nul == lon) {
+      u3z(lon);
+      lon = u3nt(u3_nul, c3n, u3k(pac_u->ore));
+      //TODO  ore is always opaque lane. what about the sender=galaxy case?
+      nal = c3y;
+    }
+
+    pac_u->bod_u.con = u3ke_jam(u3nc(lon, bod));
```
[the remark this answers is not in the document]
Fang-: Hah, nice catch. I actually shot myself in the foot with this during refactoring. Doing `u3qe_cue` and explicit `u3z` ended up being the right call in the end.

Fang-:
> Not sure why we wouldn't drop version mismatches in the sponsee case. Seems fine.

@philipcmonk Doesn't that affect the "OTA that updates ames protocol" case? Though, there's other complications there too, right now we only scry the version out of ames once...

Fang-:
Addressed all review feedback. It's not yet clear to me protocol version mismatches are always safe/sane to drop.

@philipcmonk @belisarius222 please take another look when you get the chance. Thanks!

pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +598,76 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
+  _ames_panc_free(pac_u);
+}
+
+/*  _ames_lane_scry_cb(): learn lane to forward packet on
+*/
+static void
+_ames_lane_scry_cb(void* vod_p, u3_noun nun)
+{
+  u3_panc* pac_u = vod_p;
+  u3_weak  las = u3r_at(15, nun);  //TODO  why [~ %noun ~ lane]
```
joemfb, line 648: @Fang-, this was just a refactoring bug: introduced in a53f688d01262403a5c61bb9a68ca243cf080411 and fixed in 406c86568900b7be0c2bea0239fe8806a7a0f010. Please merge `ipc-redux` back into this branch and correct the axis.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -738,42 +738,72 @@
...
+  ::
+  ?.  ?&  =(&+our why)
+          =([%$ %da now] lot)
+          =(%$ syd)
+      ==
+    ~
+  ::  /ax/protocol/version           @
+  ::  /ax/peers                      (map ship ?(%alien %known))
+  ::  /ax/peers/[ship]               ship-state
+  ::  /ax/peers/[ship]/forward-lane  (unit lane)
```
Fang-, line 763:
```suggestion
  ::  /ax/peers/[ship]/forward-lane  (list lane)
```

joemfb:
Oof, sorry @Fang-!

PR #3091, opened 2020-07-03
chat-store: don't trim on %messages, either

Fang-:
#3083 removed auto-trimming when receiving individual messages, but didn't touch the `%messages` case.

It's... not nice that there's duplicate logic for `%message` and `%messages`, but that's probably out of scope here.

(Targetting master, maybe we want to push this out as a hotfix.)


Fang-:
Going into next-userspace instead, considering it's Friday, a holiday, and I'm gonna be afk over the weekend.

PR #3099, opened 2020-07-07
Memory compaction

philipcmonk:
This compacts memory on the home road, allowing you to reclaim contiguous free space.

Read [this](https://github.com/urbit/urbit/blob/69f8602e55c9467f6a8b714af5cd9ecd7224eb04/doc/spec/u3.md) if you're interested in the basics of how our memory allocator works.  Importantly, while you can free memory on the home road (permament memory, vs inner roads which exist and are freed during an event), it's unlikely to be at the edge of the heap, so you end up with framgented memory.  Processing events happen in inner roads, which only exist in the contigous free space between the home road's stack and heap.

Some ships end up with badly fragmented home road heaps, such that most of their loom is unusable.  |pack solves this by serializing, deduplicating, and deserializing, but it's quite slow, doesn't work for too much data (which is when you want it!), and it allocates on the heap as it's working, which fails if you have too little contiguous free space -- the very problem it's trying to solve.

This algorithm does *not* deduplicate since that's a relatively complex operation to do efficiently, and you need heap space somewhere (hopefully not on the loom).  Instead, it simply compacts the memory.  The basic algorithm (u3a_compact in pkg/urbit/noun/allocate.c) is:

- Sweep through the heap with a pointer starting at the beginning of the heap.  Write the current value of that pointer to the trailing size in the box and increment the pointer by the size of the box.  This is the intended new address of that box.  Note that everything in the heap is a "box" which starts and ends with its size and also includes a reference count (if zero it's free memory).
- Trace through the heap starting from the same roots used for |mass and |pack, rewriting all pointers by looking at the destination of the pointer for what we just put in the trailing size word.
- Sweep through the heap again, actually moving each box to the new address, and writing the correct size to the trailing size word.  Since we only ever move left (to lower addresses), this is safe to do in place.

This is fast, not much slower than `|mass`, and it uses no heap space, only the C stack.

This is a draft PR since it's not hooked up to a convenient interface.  To run this for now, download the static binaries here:

Linux: https://bootstrap.urbit.org/c184752f8dd9ed0b0c8175570f9c9209f78b1fcc-linux64.tgz
MacOS: https://bootstrap.urbit.org/c184752f8dd9ed0b0c8175570f9c9209f78b1fcc-darwin.tgz

Then run with the `-C 1337` flag.  This will compact and then immediately exit.  It will end with `pier: serf error: end of file` or similar, which is normal.  Run with a standard binary after that and run |mass -- you should see that "free lists" are very small (a few MB at most, all created since starting the ship again).

### DIFF
--- a/nix/nixcrpkgs/pkgs.nix
+++ b/nix/nixcrpkgs/pkgs.nix
@@ -24,10 +24,6 @@ rec {
     inherit crossenv;
   };
 
-  ncurses = import ./pkgs/ncurses {
-    inherit crossenv;
-  };
-
   pdcurses = import ./pkgs/pdcurses {
     inherit crossenv;
   };
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Serf/IPC.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Serf/IPC.hs
@@ -19,9 +19,10 @@
   +$  gang  (unit (set ship))
   +$  writ
     $%  $:  %live
-            $%  [%exit cod=@]
+            $%  [%cram eve=@]
+                [%exit cod=@]
                 [%save eve=@]
-                [%pack eve=@]
+                [%pack ~]
         ==  ==
         [%peek mil=@ now=@da lyc=gang pat=path]
         [%play eve=@ lit=(list ?((pair @da ovum) *))]
@@ -107,7 +108,8 @@ data Serf = Serf
 data Live
   = LExit Atom -- exit status code
   | LSave EventId
-  | LPack EventId
+  | LCram EventId
+  | LPack ()
  deriving (Show)
 
 data Play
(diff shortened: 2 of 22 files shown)


joemfb:
This implementation now replaces `|pack`. Additionally, it can be run directly via the urbit-worker binary:

```
./build/urbit-worker pack path/to/pier
```

The old `|pack` implementation is no longer available from within dojo. It can be reconstructed (over multiple process lifetimes) with the `-n` and -r` arguments (replay up to N and save a jammed snapshot "rock" and load from the same, respectively). The same operations are available directly in the urbit-worker as "cram" and "queu". Run `./build/urbit-worker` without arguments for more usage.

---

I need to make a further review/refactoring pass over the `u3a_compact()` implementation itself, then this PR can come out of draft status.

ohAitch:
Uh, does this mean the existence of urbit-worker is user-facing now? Is
there any good way to thread things back through to be runnable from the
dojo again, even if there has to be a "to continue packing, urbit must
exit. Restart it by running `urbit {piername-which-we-know}`" step? (Though
presumably the whole point of a worker is you can start/stop it
automatically, at least in a one dispatcher many hosted urbits scenario)

joemfb:
@ohAitch, all these capabilities are available to the "king" process, and could be straightforwardly made available to `:dojo` et al (a la `|pack`, `|mass`, &c). It's perfectly reasonable to stop and restart worker processes. But it's a nontrivial amount of work to do this well in our current C async state machines. It's much simpler to deploy this kind of cleanup code in a synchronous environment.

I'd also be happy to ship an `urbit-utils` or somesuch thing. Or just add recovery subcommands to the main binary itself. This approach is just short term expedience. I will be doing much more work in this area, and would welcome your feedback and recommendations. But I want to ship what we have, in a minimally viable and accessible way, ASAP.

Issue #2623, opened 2020-03-30
Sigils in profile overlay should clickthrough to profile in groups


custom-jonathan:
I'd love to have a member's sigil be clickable to their group profile details from within a group chat.

SetKam:
Agree on this one

matildepark:
Addressed partially in #2714 though it's possible we could have clicking on the overlay's sigil links to the profile within Groups. @liam-fitzgerald

matildepark:
(@liam-fitzgerald Feature, not a bug!)

PR #3104, opened 2020-07-08
chat: profile overlay links to group profile

matildepark:
See the request in #2623.

Also discovered and fixed a bug where we were passing `props.association` as `props.group` to the Message component, which passes that all the way down through `OverlaySigil` and `ProfileOverlay`, so the conditional check for hidden group was just silently producing `undefined` while there was no association prop to check, causing crashes on `ProfileOverlay` render.


matildepark:
pong @liam-fitzgerald

Issue #3102, opened 2020-07-07
|reset is broken on vanilla master


tacryt-socryp:
Type `|reset` into a freshly booted ~zod or ~bus. Get a %net-need doesn't match a %net-have.

philipcmonk:
Here's a trace of it failing with the laconic bit unset: https://pastebin.com/U5GLAULe

The issue is that gall's molting is triggered by a sign from gall/dojo instead of like usual from clay.  So gall tries to `%huck` the sign through behn so it can molt and then process the sign.  However, the sign includes the source vane which makes it trivially type-fail, but more generally `%huck` requires to add the sign type to behn's gifts.  Maybe we could make behn's gifts include sign-arvo or something, but that's a bit strange.

In general |reset isn't needed on master, but this is a real bug in gall that will likely bite us sooner or later.

cc @belisarius222

belisarius222:
This is my fault.  I'll fix it.

I do think hucking is the right thing to do here.  If we need to add $sign-arvo to behn's gifts, then I think that's just the price we pay for using it as a generic relay.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
One possibility is to have an explicit Behn sign `[%heck sign-arvo]` which Gall explicitly calls its own +take with.

belisarius222:
That's a cool idea.

—
~rovnys-ricfer
https://urbit.org

PR #3107, opened 2020-07-08
zuse,behn,gall: fix %huck and |reset

belisarius222:
This fixes https://github.com/urbit/urbit/issues/3102 by modifying Behn's `%huck` facility to accept and produce a `$sign-arvo` directly instead of using Arvo's meta-vase reduction.

This has the potential downside that Behn's `$task` now depends on `$sign-arvo`.  I'm not aware of any immediate problems from doing that, but maybe it could make future Zuse changes more difficult somehow, so I'm flagging it here.

I have tested `|reset` but not the other use of `%huck` in Gall.  I also confirmed that after applying this fix, changing Zuse still causes a successful `+molt`.

### DIFF
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -75,9 +75,9 @@
   ::    Useful if you want to continue working after other moves finish.
   ::
   ++  huck
-    |=  mov=vase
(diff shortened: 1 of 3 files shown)


philipcmonk:
I ran this on ~bus and it worked, so LGTM!

PR #3108, opened 2020-07-08
-build-{cast,file,mark}: accept full beams; load from any desk

belisarius222:
Take a full beam in `-build-file`, `-build-cast`, and `-build-mark` threads so that they can run Ford builds from any desk.

They're now callable from dojo like this:
```
=f -build-file %/lib/sole/hoon
=c -build-cast /=work=/mime/txt
=m -build-mark %/mime
```

### DIFF
--- a/pkg/arvo/ted/build-cast.hoon
+++ b/pkg/arvo/ted/build-cast.hoon
@@ -5,8 +5,9 @@
 |=  arg=vase
 =/  m  (strand ,vase)
 ^-  form:m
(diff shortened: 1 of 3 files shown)


belisarius222:
I thought about that a little, but I'm not sure there's a convenient way to assemble a beak in the dojo. Some other program could use the +build-file et. al. arms from /lib/strandio directly if it's assembled its own beak.

—
~rovnys-ricfer
https://urbit.org

PR #3115, opened 2020-07-09
tab completion in chat-cli completes without prefix char

ryjm:
Describe the bug
In chat-cli, seem like the command tab completion triggers without the necessary prefix char (; in chat-cli) but not with it.

To Reproduce
Hit tab at a clear prompt in chat-cli, see completions for ; commands. Enter ; and hit tab, no more completions.

Expected behaviour
Should see completions after ;

Notify maintainers
@Fang-

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -177,9 +177,7 @@
   ::
   ++  tab-list
     |=  sole-id=@ta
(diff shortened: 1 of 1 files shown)


Fang-:
Oh huh, good catch!

I think the better fix might be to add the prefix `;` straight into the commands as listed in `+tab-list:sh`, and no longer write them as `%constants`, but as the `'cords'` they really are. The way they're currently written was a mistake.

PR #3123, opened 2020-07-14
Initial TTracker commit

nicholasforbes:
This is the initial commit for the MBTA T tracker. https://grants.urbit.org/bounties/2039941382-boston-t-train-tracker-landscape-tile

A few notes. 

1. One of the requirements for this bounty is a trip planner. The MBTA API does not support this functionality. I spent a few days trying to think about how to use the API to accomplish this but eventually emailed the MBTA dev team and they replied with 

"Hi, we currently do not provide our own API for this. Transit trip planning can be a very complex task, and we'd recommend using off-the-shelf solutions for it. The simplest option would likely be Google's Directions API, though this comes at a cost per request. OpenTripPlanner is a good free and open-source option (it powers the trip planner on mbta.com), but requires some amount of setup on your own infrastructure. Hopefully one of these makes sense for your use case."

I was thinking maybe I could scrape the https://www.mbta.com/trip-planner page, I have attempted this but ran into a few issues. I am going to post a longer question about this to the urbit-dev channel. I have built out the UI for this feature and will know where to go after I send the email.

2. I am going to make a few cosmetic changes to this repo over the next few days and I also want to add a place for in the UI for api key management. 

3. This is my first ever open source submission (Urbit is quite inspiring!). I'm not entirely sure if I put all of the required files in the correct place so please let me know if you would like me to change anything or if I screwed anything up in the submission process.

Thanks,
Nick


matildepark:
I know it does tell you to do this but this really shouldn't be a PR. I personally think a GitHub repo submission is fine. I suggested that for #3010. @rmariani

rmariani:
Yeah a personal repo would be great, @nicholasforbes. Someone will need to review just the same there, though.

PR #3125, opened 2020-07-14
nix: fix boot-ship.nix

ryjm:


matildepark:
Honestly unsure who to tag in but @jtobin knows Nix stuff, I think ...

ryjm:
i think boot-ship.nix was just never updated to reflect the new nix organization - had this in my git stash for a while.

PR #3126, opened 2020-07-14
interface: add hot reload, dev server

matildepark:
Currently our Landscape build process relies on autocommitting with herb, or committing to your ship manually each time you make a change:

- You run `npm run build:dev`
- It compiles and copies the JS binary into your ship's pier
- You commit the new JS blob
- You shift-refresh the page so it isn't cached.

This PR amends this flow; you now run `npm run start` to start a webpack-dev-server. 

When you make changes to your source code, the index.js binary has hot reload enabled, and thus, refreshes the page for you. No further work necessary.

When you're actually shipping code, `npm run build:prod` continues to output the binary in `pkg/arvo`.

### DIFF
--- a/pkg/interface/config/webpack.dev.js
+++ b/pkg/interface/config/webpack.dev.js
@@ -22,26 +22,6 @@ class UrbitShipPlugin {
       'UrbitShipPlugin',
       async (compilation) => {
         const src = path.resolve(compiler.options.output.path, 'index.js');
-
-        return Promise.all(this.piers.map((pier) => {
-          const landscapeHtml = path.resolve(pier, 'app/landscape/index.html');
-          fs.readFile(landscapeHtml, 'utf8', (err, data) => {
-            if (err) {
-              return console.log(err);
-            }
-            const result = data.replace('/~landscape/js/index.js', 'http://localhost:9000/index.js');
-
-            if (data === result) {
-              return;
-            }
-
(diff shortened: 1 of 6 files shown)


pkg/interface/config/webpack.dev.js
```diff
@@ -77,12 +99,12 @@ module.exports = {
...
   devtool: 'inline-source-map',
-  // devServer: {
-  //   contentBase: path.join(__dirname, './'),
-  //   hot: true,
-  //   port: 9000,
-  //   historyApiFallback: true
-  // },
+  devServer: {
+    contentBase: path.join(__dirname, '../dist'),
+    hot: true,
```
liam-fitzgerald, line 104: we can probably get real hot reload by proxying 404s to localhost:80 or wherever. This would kinda fall down on a multi ship dev env though.

matildepark: ? maybe we can include that but with a comment to disable in multi-ship setups? I don't use multi-ship and it seems rare for FE development.

matildepark: I mean it's used for like, cross-ship chats and whatnot but usually the second ship doesn't receive the FE changes, it's just to model incoming stuff.

matildepark:
I'm going to amend this to not overwrite Landscape's html and use the dev server proxy method (and document this), standby...

PR #3129, opened 2020-07-15
interface: add HMR to hot reloading config

liam-fitzgerald:
Adds HMR to the hot reloading config. JS changes no longer require a refresh and appear instantly after the compilation has finished.

Bonus round: you can point `URL` at your own planet to test JS changes there.

### DIFF
--- a/pkg/interface/src/App.js
+++ b/pkg/interface/src/App.js
@@ -1,3 +1,5 @@
+import { hot } from 'react-hot-loader/root';
+import 'react-hot-loader';
 import * as React from 'react';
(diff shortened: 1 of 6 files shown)


pkg/interface/src/App.js
```diff
@@ -152,3 +154,5 @@ export default class App extends React.Component {
   }
 }
 
+export default hot(App);
```
matildepark, line 157: Does this affect production?

g-a-v-i-n: Also this is very very cool.

PR #3130, opened 2020-07-15
interface: show base hash

liam-fitzgerald:
Makes the base desk hash accessible through scry and then shows it in the status bar. Unfortunately, we can't scry this in the login page because we don't yet have a session to do scries with. 

Preview (see dhk61 in top right):
Screen Shot 2020-07-15 at 2 47 44 pm

Screen Shot 2020-07-15 at 2 48 03 pm

### DIFF
--- a/pkg/arvo/app/file-server.hoon
+++ b/pkg/arvo/app/file-server.hoon
@@ -223,7 +223,28 @@
   ==
 ::
 ++  on-leave  on-leave:def
-++  on-peek   on-peek:def
+++  on-peek   
+  |=  =path
+  ^-  (unit (unit cage))
+  |^
+  ?+  path  (on-peek:def path)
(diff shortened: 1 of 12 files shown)


matildepark:
I would move it to the bottom left of the Launch screen.

matildepark:
Oh also place the "fixes" stuff in the commit so it only closes the issue once it hits master, of course.

liam-fitzgerald:
Screen Shot 2020-07-16 at 10 17 40 am
Screen Shot 2020-07-16 at 10 17 21 am

updated preview cc @urcades

urcades:
yo @liam-fitzgerald ! this is beautiful/very ideal — it looks perfect in that corner.

I think there are some more vague-than-not ideas on how to embed this into the friendly command line drop down, but i can see this working for a very long time

galenwp:
Can we make it monospace?

Issue #3132, opened 2020-07-15
Paste broken in web dojo


vvisigoth:
**Describe the bug**
Pasting something into the web dojo breaks input.

**To Reproduce**
1. Go to ~dojo
2. try to paste in some text
3. try to edit pasted text
4. nothing works

**Expected behavior**
Pasted text doesn't behave differently than normal text

**Additional context**
An error is shown when you paste
```
Uncaught ReferenceError: regeneratorRuntime is not defined
```

matildepark:
Browser?

matildepark:
Oh it's looking like [Babel did this](https://github.com/babel/babel/issues/5085), ok let me see how to amend the build.

PR #3134, opened 2020-07-15
interface: add babel-transform-runtime

matildepark:
See #3132. Prior to this, we hit errors due to Babel not being able to transform async functions for browsers.


matildepark:
ping @g-a-v-i-n

PR #3135, opened 2020-07-16
landscape clock: render sun+moon after night

finled-bansel:
The clock renders the sun+moon before rendering the part of the image for the night section.

This ends up hiding the moon behind the night part of the image.

Layer the sun+moon after the night section to make the moon visible during night time.


matildepark:
ping @g-a-v-i-n

PR #3136, opened 2020-07-16
Code block rendering

tylershuster:
This PR adds a new dependency (bleh) on `react-syntax-highlighter` and uses the `ReactMarkdown` component's `renderer` prop to add a renderer to code blocks (not inline code). It uses a listener on `window.matchMedia` to ensure that dark mode is maintained reactively.

[Code block rendering](https://marpem-files.sfo2.digitaloceanspaces.com/code-test.png)

Note that the second message is not highlighted because it is not fenced, and that inline code blocks are not highlighted, and language declarations are respected.

### DIFF
--- a/pkg/interface/src/apps/chat/components/lib/message.js
+++ b/pkg/interface/src/apps/chat/components/lib/message.js
@@ -4,6 +4,43 @@ import { uxToHex, cite, writeText } from '../../../../lib/util';
 import moment from 'moment';
 import ReactMarkdown from 'react-markdown';
 import RemarkDisableTokenizers from 'remark-disable-tokenizers';
+import SyntaxHighlighter from 'react-syntax-highlighter';
+import { darcula, atomOneLight } from 'react-syntax-highlighter/dist/esm/styles/hljs';
+
+export class CodeBlock extends Component {
+  static defaultProps = {
+    language: 'markdown',
+  }
+
+  constructor() {
+    super();
+    const scheme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+    this.state = {
+      scheme
+    };
+  }
+
+  componentDidMount() {
+    const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
+    darkModeMediaQuery.addListener((event) => {
+      const scheme = event.matches ? 'dark' : 'light';
+      if (scheme !== this.state.scheme) {
+        this.setState({ scheme });
+      }
+    });
+  }
+
+  render() {
+    const { language, value } = this.props;
+    const style = this.state.scheme === 'dark' ? darcula : atomOneLight;
+    return (
+      <SyntaxHighlighter language={language} style={style}>
+        {value}
+      </SyntaxHighlighter>
+    );
+  }
+}
+
 
 const DISABLED_BLOCK_TOKENS = [
   'indentedCode',
@@ -28,6 +65,7 @@ const DISABLED_INLINE_TOKENS = [
 const MessageMarkdown = React.memo(
   props => (<ReactMarkdown
               {...props}
+              renderers={{code: CodeBlock}}
               plugins={[[RemarkDisableTokenizers, { block: DISABLED_BLOCK_TOKENS, inline: DISABLED_INLINE_TOKENS }]]}
             />));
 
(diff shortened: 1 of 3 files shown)


matildepark:
I feel like there's a better solution here between `indigo-react` and the groups refactor at `lf/groups-refactor` (I'm trying to get it in `next-userspace` faster ...). Hoping @liam-fitzgerald and @g-a-v-i-n have an idea about the best way to contribute this.

pkg/interface/src/apps/chat/components/lib/message.js
```diff
@@ -4,6 +4,43 @@ import { uxToHex, cite, writeText } from '../../../../lib/util';
 import moment from 'moment';
 import ReactMarkdown from 'react-markdown';
 import RemarkDisableTokenizers from 'remark-disable-tokenizers';
+import SyntaxHighlighter from 'react-syntax-highlighter';
+import { darcula, atomOneLight } from 'react-syntax-highlighter/dist/esm/styles/hljs';
+
```
liam-fitzgerald, line 9: This ought to be in a seperate file.

pkg/interface/src/apps/chat/components/lib/message.js
```diff
@@ -4,6 +4,43 @@ import { uxToHex, cite, writeText } from '../../../../lib/util';
...
+export class CodeBlock extends Component {
+  static defaultProps = {
+    language: 'markdown',
+  }
+
+  constructor() {
+    super();
```
liam-fitzgerald, line 16: Should be `super(props)`

pkg/interface/src/apps/chat/components/lib/message.js
```diff
@@ -4,6 +4,43 @@ import { uxToHex, cite, writeText } from '../../../../lib/util';
...
+    const scheme = window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light';
+    this.state = {
+      scheme
+    };
+  }
+
+  componentDidMount() {
+    const darkModeMediaQuery = window.matchMedia('(prefers-color-scheme: dark)');
```
liam-fitzgerald, line 24: A media query watcher will be set up for each message being rendered. It'd be ideal to only do this once. When the groups refactor is merged, this probably should be a part of the state and handed down to the component in a prop.

pkg/interface/src/apps/chat/components/lib/message.js
liam-fitzgerald: and if you're feeling generous, typescript would be nice but not a blocker for this to be merged.

tylershuster: Got it. Move `CodeBlock` into a separate file?

liam-fitzgerald: Yep!

tylershuster:
I tried to work with CodeMirror because that would reduce dependencies but I ran into two problems: One, getting CodeMirror's height to adapt to the contents (probably wouldn't be too hard) and two, mixing markdown and code modes is kind of tricky. If it renders markdown, it doesn't parse it; it just highlights it and leaves the markup. And it does highlighting for the code fencing, but it leaves the ticks and language definition.

pkg/interface/src/apps/chat/components/lib/message.js
tylershuster: Ah, that is pretty expensive. Maybe for now, put in `ChatScreen` and pass it down. Or, just set the mode once. But then it doesn't change on MQ change, which can look weird but only for a little bit.

liam-fitzgerald: yeah for now put it in ChatScreen

matildepark: Why not rebase on top of `lf/groups-refactor`, add the media plumbing as a prop and target `release/next-userspace` here? It seems odd to have a change immediately obsolesced within days ...

tylershuster: Want me to implement the MQ in groups-refactor and add this component in there? Code highlighting can wait a couple days

g-a-v-i-n:
Not sure about this one because I'd prefer to handle this through indigo react.

tylershuster:
> Not sure about this one because I'd prefer to handle this through indigo react.

Can you point me in the right direction? I'm happy to PR there

tylershuster:
I'm going to close this and delete the branch so I can work on other things. Seems like this work needs to be done elsewhere. If @g-a-v-i-n or @matildepark wants to point me in the direction of where to implement it in indigo-react I'd be happy to do the work there.

PR #3138, opened 2020-07-17
interface: show connection status indicator

liam-fitzgerald:
- Shows a connection status indicator in the status bar
- If the reconnection times out and enters the 'Disconnected' state, then the indicator can be clicked to manually reconnect
- Fixes a bug in the reconnect logic where app-specific subscriptions were not re-established.

cc: @urcades 
[demo](https://ewr1.vultrobjects.com/urbit/disconnect.mov)

### DIFF
--- a/pkg/arvo/app/landscape/js/channel.js
+++ b/pkg/arvo/app/landscape/js/channel.js
@@ -9,6 +9,9 @@ class Channel {
     this.onChannelError = (err) => {
       console.error('event source error: ', err);
     };
+    this.onChannelOpen = (e) => {
+      console.log('open', e);
+    };
   }
 
   init() {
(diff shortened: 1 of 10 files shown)


matildepark:
Shh don't tell Galen

liam-fitzgerald:
I've tried this on livenet minus the channel.js change, which is not super critical to the reconnection functionality itself.

urcades:
yeah, this is amazing

galenwp:
Definitely happy to see this. 

Looking at your demo: can we make it obvious that clicking 'Disconnected' will reconnect you? either by changing the copy ('Disconnected - click to try to reconnect' is my off-the-cuff attempt) or at least by making it cursor: pointer on hover?

Thanks!

ohAitch:
Throw in a ⟳ perhaps

liam-fitzgerald:
Screen Shot 2020-07-21 at 10 58 05 am
new disconnect message. w/ `cursor: pointer`

matildepark:
Yeah, let's be more sparse. What you want is a simple call to action: `Reconnect  ⟳` in red.

PR #3139, opened 2020-07-18
glob: Distribute landscape's index.js outside clay

philipcmonk:
As seen on tv: https://www.youtube.com/watch?v=kY3lmbqcXZc

This implements the %glob app, which includes a hash of a directory and downloads the contents from bootstrap.urbit.org.  This is specifically so that landscape can cheaply distribute updates to its javascript, which is 10MB minified.  There's little reason to keep old versions of this around, so storing it in Clay is unnecessary, and takes a lot of memory.

This is a minimum viable implementation that could/should be generalized in many ways, especially:

- Allow convenient registering of additional blobs to download
- Add option to choose where to download blobs from
- Add option to download from an app on another ship instead of over http
- Add option to refer to blob by name instead of hash so you don't have to update the source to get an update.  Possibly this should only be used when downloading from another ship; otherwise you have to worry about separately signing the update.

To update the JS for release, you should copy it into `app/landscape/js/index.js` of a running fakezod and run `:glob|make`.  This will put a `glob-0v*.glob` file in .urb/put, which should be uploaded to bootstrap.urbit.org (without changing the name, which is a hash of its contents).  Then, you must update +hash at the top of `app/glob.hoon` with the hash in the filename.

For development, #3126 means you don't touch this at all.  Simply `cd pkg/interface; npm install; npm run start` and navigate to localhost:9000.


philipcmonk:
Right now I've got it set up to automatically use this for index.js only, but it would be pretty straightforward to use it for everything in app/landscape.  Do you think that's better?  About 98% of the space is in the js file.

Yes, `:glob|make` takes what's in your app/landscape/js/index.js and serializes it into a glob in .urb/put.  Possibly it would be useful to let that take a path argument to make globs out of arbitrary directories.

Couple more things:

- `:glob %kick` restarts the download, which is useful in testing but shouldn't be needed.  If the download fails or gets stuck, glob will retry after 30 minutes.

- For reference, on my machine when you kick off the download it takes a few seconds to download in the background and then about 1 second of <<http>> processing.

ohAitch:
"The urbit frontend binary is so big it doesn't fit in the urbit
filesystem" certainly is a place to have ended up in.

What size are the actual byte diffs between the blobs? Could a
https://bup.github.io / rsync style rolling-hash splitter be used to
recover some sanity, assuming the vast majority of the 10Mb are
dependencies that the vast majority of the time are unchanged? (And if we
can get the diff/patch logic to pointer-unify matching atom chunks in the
resulting ropelike structure)

matildepark:
> "The urbit frontend binary is so big it doesn't fit in the urbit filesystem" certainly is a place to have ended up in.

It was actually a lot bigger before the SPA branch — it's one 10mb blob instead of 4+ of them. But we're at the point now where it isn't Ford + OTAs bailing ships, it's the rapidly-updating, versioned-with-back-end, immediately-disposable blob pinning everyone's ship to kelvin 200 before it bursts the loom.

Clay isn't suitable for large object storage at the moment. Thus the s3-store. Maybe in the future it will be — at the moment what comes first is user experience in the short to mid-term: performance, stability. And we keep the door open for those iterations as we go ...

> What size are the actual byte diffs between the blobs? 

I believe @philipcmonk had a rationale against diffing JS but can't recall.

ohAitch:
One could also go for "implement enough of a wasm interpreter in hoon to ship duktape and run all the js compiler/linker tooling from source locally", but that seems less practical along a number of fronts. Conversely this PR is imminently \~practical~, at the cost of introducing a post-boot dependency on "bootstrap".urbit.org and generally running from the sign that something is fundamentally architecturally broken when your general-purpose file system can't store large files - specific files you actually want to use even!

ohAitch:
Hm
The term "large object store" generally does not refer to objects that probably fit in your laptop processor's L3 cache. Like, offloading a 5GB video to S3 because that legitimately occupies a percent of your local disk - sure, reasonable! But the sizes urbit is struggling with here are incredibly tiny; yes, even if you publish thirty revisions of the thing.

I understand the like, 31-bit address space constraints the current implementation is working under(after all with this PR the "disposable" blobs still end up permanently in your event log, that just doesn't cause anything the slightest problem). But to me keeping the door open would look like some workaround keeping distribution and storage fundamentally within the system, not achieving stability by undermining core value propositions like "independent network".

philipcmonk:
I haven't looked closely at what changes in the result of compiling/minifying a bunch of js -- it wouldn't surprise me if small changes in the source cause large changes in the minified js by eg shifting all the gensym'd names.  I do suspect that compression algorithms could be devised to reduce the size significantly further -- even gzip would probably help a lot.

> One could also go for "implement enough of a wasm interpreter in hoon to ship duktape and run all the js compiler/linker tooling from source locally", but that seems less practical along a number of fronts. 

That is the correct answer, and that's not practical for now.

> Conversely this PR is imminently ~practical~, at the cost of introducing a post-boot dependency on "bootstrap".urbit.org and generally running from the sign that something is fundamentally architecturally broken when your general-purpose file system can't store large files - specific files you actually want to use even!

> The term "large object store" generally does not refer to objects that probably fit in your laptop processor's L3 cache.

It literally does, though.  We pulled the pills out of git back when they were only 1MB.  We used a hand-rolled file store, but nowadays we use the Git Large File Store.  I believe it was a mistake to call Clay a filesystem since that implies it's the place to put all your data.  "The purpose of a system is what it does" and all that.  And of course, you don't actually want to store every version of your compiled JS client, just like you don't actually want to download every pill that has existed.

Other than trying to heuristically compress compiled data, the only in-Clay solution I can imagine is tombstoning commits or files, which we'll want at some point, but if you're *not* keeping the history or doing semantic merges, why should it be in Clay?  I don't think we should focus on adding features to Clay to support keeping compiled binaries in it, since that's so far from its normal use, and we shouldn't be doing that anyway.

> But to me keeping the door open would look like some workaround keeping distribution and storage fundamentally within the system, not achieving stability by undermining core value propositions like "independent network".

I don't see it as undermining independence.  The hash comes from Clay from Ames (hence signed by your sponsor), so it doesn't much matter where the bits come from.  It'll be easy to send this app-to-app over Ames -- would that satisfy your "independent network" criteria?

ohAitch:
> I haven't looked closely at what changes in the result of compiling/minifying a bunch of js -- it wouldn't surprise me if small changes in the source cause large changes in the minified js by eg shifting all the gensym'd names. I do suspect that compression algorithms could be devised to reduce the size significantly further -- even gzip would probably help a lot.

I expect some fairly straightforward build step determinism tweaks could help on that front. For one thing, [conventional wisdom](https://medium.com/hackernoon/the-100-correct-way-to-split-your-chunks-with-webpack-f8a9df5b7758) is to split a bundle.vendor.js and bundle.main.js - though aimed at reuse of browser caches than lightening the load of updating your webserver, exactly.

(Similarly it would certainly be possible to throw some padding into pills so that when jam'd they come out as mostly plaintext and don't shift around by unaligned numbers of bits all the time, but that ship has perhaps sailed. And those do have a fundamentally binary nock portion.)

> > The term "large object store" generally does not refer to objects that probably fit in your laptop processor's L3 cache.
> 
> It literally does, though. We pulled the pills out of git back when they were only 1MB. We used a hand-rolled file store, but nowadays we use the Git Large File Store.

You may recall I objected strenuously to that move also :)
Servers owned by Microsoft are worse than cashix or the bootstrap S3 bucket are worse than something actually distributed(nb. we do not have a compelling example of something actually distributed.) But also ultimately the bootstrapping process has to involve a measure of existing infrastructure, it's much less a concern than everything after the first `ovo` hits the deck.

> I believe it was a mistake to call Clay a filesystem since that implies it's the place to put all your data. "The purpose of a system is what it does" and all that. And of course, you don't actually want to store every version of your compiled JS client, just like you don't actually want to download every pill that has existed.

Truly it was a mistake because it produces the impression that urbit is an OS with a filesystem, if we're really talking user experiences.
> 
> Other than trying to heuristically compress compiled data, the only in-Clay solution I can imagine is tombstoning commits or files, which we'll want at some point, but if you're _not_ keeping the history or doing semantic merges, why should it be in Clay? I don't think we should focus on adding features to Clay to support keeping compiled binaries in it, since that's so far from its normal use, and we shouldn't be doing that anyway.

Fundamentally this raises the question "where then do you keep your files". The PR presents one answer: ad-hoc as variables, single level store means you won't lose them, and indexing/navigability who needs that?
I can see a future where there's a [fossil(4)](https://9fans.github.io/plan9port/man/man4/fossil.html) to clay's [venti(7)](https://9fans.github.io/plan9port/man/man7/venti.html), perhaps entire desks only accessible by the current time. Doesn't have quite the ring of "typed revision-controlled globally referentially-transparent filesystem" though.
>
> I don't see it as undermining independence. The hash comes from Clay from Ames (hence signed by your sponsor), so it doesn't much matter where the bits come from. It'll be easy to send this app-to-app over Ames -- would that satisfy your "independent network" criteria?

It would, yes. Integrity is far from than nothing, but availability is also kind of important - I believe before this change you can run urbit with TCP fire-walled off after boot, possibly with a local ETH proxy? Without reliance on centralized Tlon servers to receive software updates, at any rate.

Of course trying to do this robustly runs the risk of reimplementing 5-10% of (the clay part of) clay in userspace, but at worst this is an opportunity to reflect on the inelegance of your solutions, with the bits as you say staying available within the system.

philipcmonk:
Alright, I tested this on a localhost copy of ~wicdev and it worked fine.  I'm comfortable deploying this after review.

> I can see a future where there's a fossil(4) to clay's venti(7)

Thanks for the reference, I hadn't seen venti before, and I haven't looked closely enough at fossil.

matildepark:
Let's do this!

PR #3143, opened 2020-07-21
interface: two design tweaks

matildepark:
If I can squeeze these in here I'd like to — 

- base hash uses `mono`
- group-detail has a correction for the line-height of its items


matildepark:
As for the deletion of those imports — they're not used, it's just lint / cleanup.

PR #3146, opened 2020-07-22
chat-js: save unsent message for each station

lukechampine:
Previous behavior:

1. Type message in chat without submitting it
2. Navigate to other chat
3. Navigate back
4. Message has been cleared

New behavior: message is not cleared. This matches the behavior of Discord, Slack, etc.


galenwp:
hero

PR #3152, opened 2020-07-23
group-store: remove scries from OTA logic

liam-fitzgerald:
Removes scries such that the OTA will succeed with a molting gall.

cc: @tacryt-socryp  can you try this on your ship as well?


matildepark:
Testing on a local before I stamp, standby

PR #3165, opened 2020-07-23
file-server: base hash is same as +trouble

tacryt-socryp:
Tested on a fake ship. +trouble and base-hash in landscape are the same upon boot.


philipcmonk:
If we want it to be just the end, it should just be `(end 0 25 ...)` instead of `(end 3 3)` (ie last 25 bits instead of 24), right?

matildepark:
Yes, ha. I‘ll try that in a separate PR...

matildepark:
Superseded by #3166.

Issue #3159, opened 2020-07-23
Base hash displayed in Landscape doesn't match +trouble


sivner-figbus:
**Describe the bug**
As of today's OTA (it's 20200723 1400 UTC) the basehash displayed in dojo in response to "+trouble" does not match that displayed on the front page of Landscape.

**To Reproduce**
1. Go to 'dojo'
2. Type '+trouble'
3. Observe '0vk.c90mr.psja5.f7gfu.e1fm1.dgt61.9iv7q.6ripd.dklhc.vr8lo.jp5fi'
4. Go to Landscape
5. Observe bottom left corner of screen says '3p5fi'

**Expected behavior**
basehash reported in both places should match

**Screenshots**

![image](https://user-images.githubusercontent.com/67953485/88294597-5cd87e80-ccf4-11ea-971c-71b9f4cb1d3a.png)
![image](https://user-images.githubusercontent.com/67953485/88294621-6661e680-ccf4-11ea-9354-41f9c27fdc9d.png)

matildepark:
~sorpex presents a hint:

![image](https://user-images.githubusercontent.com/20846414/88306266-8683ac80-ccd8-11ea-8c4e-bf0f19d3147b.png)

PR #3166, opened 2020-07-23
file-server: fix base-hash trimming

matildepark:
See #3159. Maintains parity with +trouble, tested by booting a fake ~zod and a fake ~marzod.

### DIFF
--- a/pkg/arvo/app/file-server.hoon
+++ b/pkg/arvo/app/file-server.hoon
@@ -269,7 +269,7 @@
   ==
 ::
 ++  on-leave  on-leave:def
-++  on-peek   
+++  on-peek
   |=  =path
   ^-  (unit (unit cage))
   |^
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/file-server.hoon
```diff
@@ -287,10 +287,10 @@
       *@uv
     =/  parent  (scot %p ship.u.ota)
     =+  .^(=cass:clay %cs /[parent]/[desk.u.ota]/1/late/foo)
-    %^  end  3  3
+    %^  end  0  25
     .^(@uv %cz /[parent]/[desk.u.ota]/(scot %ud ud.cass))
   --
-  
+
```
tacryt-socryp, line 293: Hi, if you get rid of this newline, it looks good to go to me.

matildepark: That line was there before...? But alright...

PR #3169, opened 2020-07-24
metadata-store,hook: ignore bad paths in on-load

liam-fitzgerald:
should resolve the failure to process the most recent OTA on ~bitbet-bolbel. The path is malformed on ~bitbet because of what i assume to be migration jank. However, ships other than ~bitbet have reported the OTA failing and so I'm drafting this PR until I can confirm that we can safely discard that metadata.

### DIFF
--- a/pkg/arvo/app/metadata-store.hoon
+++ b/pkg/arvo/app/metadata-store.hoon
@@ -85,15 +85,16 @@
         old  [%2 +.old]
       ::
           cards
-        %+  turn
+        %+  murn
           ~(tap in ~(key by group-indices.old))
         |=  =group-path
-        ^-  card
-        =/  rid=resource
-          (de-path:resource group-path)
-        ?:  =(our.bowl entity.rid)
-          (poke-md-hook %add-owned group-path)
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
Was this a parsing issue of the path into resource id? It looks as though you're discarding the path upon failure. This seems dangerous if it's a local resource, but if it's foreign, we could easily ask them to delete and download state anew if it broke... I wish we had archiving support already for archiving local resources with broken metadata. LGTM though, I doubt we'll get a better solution until then.

liam-fitzgerald:
> Was this a parsing issue of the path into resource id? It looks as though you're discarding the path upon failure. This seems dangerous if it's a local resource, but if it's foreign, we could easily ask them to delete and download state anew if it broke... I wish we had archiving support already for archiving local resources with broken metadata. LGTM though, I doubt we'll get a better solution until then.

Yeah this is why I want to get an n > 1 sample of this error before this goes in.

liam-fitzgerald:
fine to merge: if a member processed the ota before the host of the group did, then the member may have duplicate metadata.

matildepark:
I trust u

PR #3170, opened 2020-07-24
chat-cli: revive chat creation and add group association

liam-fitzgerald:
Fixes chat creation through chat-cli, and allows users to associate new chats with groups

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -72,7 +72,7 @@
 +$  glyph  char
 ++  glyphs  "!@#$%^&()-=_+[]\{}'\\:\",.<>?"
 ::
-+$  nu-security  ?(%channel %village %village-with-group)
++$  nu-security  ?(%channel %village)
 ::
 +$  command
   $%  [%target (set target)]                        ::  set messaging target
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/chat-cli.hoon
```diff
@@ -741,15 +741,22 @@
...
+              gyf=(unit char)
+              allow-history=(unit ?)
+          ==
       ^-  (quip card _state)
-      =/  with-group=?     ?=(%village-with-group security)
+      =/  with-group=?     ?=(^ ugroup)
       =/  =target          [with-group our-self path]
       =/  real-path=^path  (target-to-path target)
+      =/  group-path=^path  ?~(ugroup ship+real-path (en-path:resource u.ugroup))
+      ?<  &(?=(%channel security) with-group)
```
Fang-, line 755: The parser still allows this, right? You might want to put a `+sear` or whatever around it to ensure people cannot trigger this assertion. That's not gonna make for the prettiest parser though...

matildepark:
@liam-fitzgerald @Fang- I assume we're all addressed here? Can we merge this?

liam-fitzgerald:
should be good

PR #3173, opened 2020-07-24
vere: cleans up handling of fatal db errors

joemfb:
The error printfs in the database weren't all properly formatted, and top-level initialization failure was not being handled (resulting in a segmentation fault, as opposed to an error message and `exit(1);`).

/cc @botter-nidnul


joemfb:
An example of the prior, incorrect behavior: https://gist.github.com/botter-nidnul/241b1d654d3e4f953d6714698f175c47

botter-nidnul:
Much cleaner error
https://gist.github.com/botter-nidnul/541ece18b5f82c9a5213ceee1073cdce

joemfb:
@philipcmonk, we need those because we put the terminal in raw mode, so newlines don't reset the cursor position.

PR #3174, opened 2020-07-24
ames.c: stateless forwarding

Fang-:
For prior discussion, see #3087.

Reopening here because Github doesn't let you reopen PRs that target a deleted branch, and doesn't let you change the merge target for closed PRs. Wonderful!

Remaining work:
- [x] Resolve merge conflicts.
- [x] Integrate @joemfb's feedback.
- [x] Test on ~nus.
- [x] Test on ~nus again post-fixes.


pkg/urbit/vere/io/ames.c
```diff
@@ -435,6 +598,76 @@ _ames_hear_bail(u3_ovum* egg_u, u3_noun lud)
...
+{
+  pac_u->sam_u->fow_d++;
+  if ( 0 == (pac_u->sam_u->fow_d % 10000) ) {
+    u3l_log("ames: forwarded %" PRIu64 " total\n", pac_u->sam_u->fow_d);
+  }
+
+  {
+    u3_noun pac = _ames_serialize_packet(pac_u, c3y);
+    while (u3_nul != las) {
+      _ames_ef_send(pac_u->sam_u, u3h(las), u3k(pac));
```
joemfb, line 632: This pattern leaks the outer cells in the cons-list `las`, but not (the head and tail of) the individual list items. You'll want to keep an extra pointer to `las`, `u3k()` both noun arguments to `_ames_ef_send()`, and then `u3z()` your saved pointer to the original `las`.

pkg/urbit/vere/io/ames.c
```diff
@@ -30,21 +33,57 @@
...
     c3_o          imp_o[256];           //  imperial print status
+    c3_o          see_o;                //  can scry
+    c3_o          fit_o;                //  filtering active
+    c3_y          ver_y;                //  protocol version
+    c3_d          vet_d;                //  version mismatches filtered
+    c3_d          mut_d;                //  invalid mugs filtered
+    u3_panc*      pac_u;                //  packets pending forwards
+    c3_d          foq_d;                //  forward queue size
+    c3_d          fow_d;                //  forwarded count
+    c3_d          fod_d;                //  forwards dropped count
```
joemfb, line 55: These new counters should all be printed in `_ames_io_info()` (which is for SIGINFO/SIGUSR1).

pkg/urbit/vere/io/ames.c
```diff
@@ -232,6 +302,99 @@ u3_ames_encode_lane(u3_lane lan) {
...
+  c3_o nal_o = c3n;
+
+  //  update the body's lane, if desired
+  //
+  if (c3y == dop_o) {
+    //  unpack (jam [(unit lane) body])
+    //
+    u3_noun lon, bod;
+    {
+      u3_noun old = u3qe_cue(pac_u->bod_u.con);
```
joemfb, line 333: This should be virtualized (with `u3m_soft()`), or invalid input will bring down the process. Inability to cue signifies bad input, so this should happen at a point where failure can guide control flow, perhaps inside the conditionals of `_ames_recv_cb()`.

pkg/urbit/vere/io/ames.c
```diff
@@ -446,38 +679,149 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+    pas_o = c3n;
+  }
+
+  //  ensure the protocol version matches ours
   //
-  if (  (0 < nrd_i)
-     && (0 == (0x7 & *((c3_w*)buf_u->base))) )
+  if ( c3y == pas_o
+    && (c3y == sam_u->fit_o)
+    && (sam_u->ver_y != (0x7 & *((c3_w*)buf_u->base))) )
```
joemfb, line 694: The packet header should be parsed prior to this point, and this check should be performed against the version in the resulting `u3_head` struct.

pkg/urbit/vere/io/ames.c
```diff
@@ -30,21 +33,57 @@
...
+    c3_y rac_y;                         //  receiver class
+    c3_o enc_o;                         //  encrypted?
+  } u3_head;
+
+/* u3_body: ames packet body
+*/
+  typedef struct _u3_body {
+    u3_noun sen;                        //  sender
+    u3_noun rec;                        //  receiver
+    u3_noun con;                        //  (jam [origin content])
```
joemfb, line 73: To the degree that it's practical, it'd be better to save c structures over nouns here. The sender and receiver could easily be `c3_d[2]`.

As structured, the content could be a pointer to a byte buffer, at the cost of some minor memory management responsibilities. But if the content is to be cue'd earlier in the process, it might not be practical to save byte buffer(s) (barring more flexible de/serialization routines than the current ones).

Fang-:
@joemfb thanks for your comments! I think I addressed all of them. Haven't done a third pass for refcounts yet, so I won't be surprised if I missed something, but ptal.

Fang-:
Just ran a little test of this on ~nus. @philipcmonk's suspicions were right, the prioritizing of scries over events, and the sheer amount of forwards we want to be doing, means we don't actually get to process regular events at all.

The forwarding load is _substantial_, I should note. By the time it had processed 20k forwards successfully, it had dropped 44k forwards on the floor due to forward-queue limits. We're nowhere close to keeping up.

Judging visually, the rate of dropping forwarding packets wasn't very different from the rate we're current getting on the all-packet queue. Might even have been dropping a bit more/faster. But perhaps that's to be expected?

We can probably get better by caching lanes in the king, even if just for a minute, but I imagine we'll want to do something about scries clogging the IPC pipe regardless.

pkg/urbit/vere/io/ames.c
```diff
@@ -705,14 +1079,44 @@ _ames_io_exit(u3_auto* car_u)
...
+  else if ( (c3n == u3a_is_cat(ver))
+         || (7 < ver) ) {
+    u3m_p("ames: strange protocol", nun);
+    sam_u->ver_y = 0;
+  }
+  else {
+    sam_u->ver_y = ver;
+  }
+
+  sam_u->fit_o = c3y;
```
joemfb, line 1104: This function leaks `nun`.

pkg/urbit/vere/io/ames.c
```diff
@@ -232,6 +301,107 @@ u3_ames_encode_lane(u3_lane lan) {
...
+      u3z(sen); u3z(rec);
+    }
+    memcpy(pac_y + 4 + sen_y + rec_y, bod_u->con_y, bod_u->con_w);
+
+    //  if we updated the origin lane, we need to update the mug too
+    //
+    if (c3y == nal_o) {
+      u3_noun bod = u3i_bytes(sen_y + rec_y + bod_u->con_w, pac_y + 4);
+      pac_u->hed_u.mug_l = u3r_mug(bod) & ((1 << 20) - 1);
+      u3z(bod);
```
joemfb, line 384: This should just be `_ca_mug_body(sen_y + rec_y + bod_u->con_w, pac_y + 4)` -- the atom allocation isn't necessary. (Also, you may need to reorder that addition, or cast the `c3_y` arguments to `c3_w`).

pkg/urbit/vere/io/ames.c
```diff
@@ -445,39 +686,164 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+  //  unpack the packet header
+  //
+  else {
+    c3_w hed_w = (byt_y[0] <<  0)
+               | (byt_y[1] <<  8)
+               | (byt_y[2] << 16)
+               | (byt_y[3] << 24);
+
+    hed_u.ver_y = hed_w & 0x7;
+    hed_u.mug_l = (hed_w >> 3) & ((1 << 20) - 1);
```
joemfb, line 708: The mask should just be `0xfffff`, maybe with `((1 << 20) - 1)` in a comment if you'd like.

pkg/urbit/vere/io/ames.c
```diff
@@ -445,39 +686,164 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+    u3r_chubs(0, 2, sen_d, sen);
+    u3z(sen); u3z(rec);
+
+    con_y = c3_malloc(con_w);
+    memcpy(con_y, bod_y + sen_y + rec_y, con_w);
+
+    //  ensure the content is cue-able
+    //
+    u3_noun con = u3i_bytes(con_w, con_y);
+    u3_noun pro = u3m_soft(0, u3qe_cue, con);
```
joemfb, line 762: I don't think it actually matters (due to the road transition), but this should be `u3ke_cue()`, to re-emphasize that `con` is being transferred. (It might also be better to just pass the `u3i_bytes()` return to further emphasize that, and avoid the intermediate variable.)

pkg/urbit/vere/io/ames.c
```diff
@@ -445,39 +686,164 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+    if (u3_blip != u3h(pro)) {
+      pas_o = c3n;
+    }
```
joemfb, line 765: This could just be `pas_o = __(u3_blip != u3h(pro));`, or `(u3_blip != u3h(pro)) ? c3n : c3y;`, or somesuch.

pkg/urbit/vere/io/ames.c
```diff
@@ -445,39 +686,164 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+        pac_u->nex_u = sam_u->pac_u;
+        sam_u->pac_u->pre_u = pac_u;
       }
+      sam_u->pac_u = pac_u;
 
-      cad = u3nt(c3__hear, u3nc(c3n, lan), msg);
+      //  if the recipient is a galaxy, their lane is always &+~gax
+      //
+      if ( (rec_d[1] == 0) && (256 > rec_d[0]) ) {
+        _ames_forward(pac_u, u3nc(u3nc(c3y, rec_d[0]), u3_nul));
```
joemfb, line 825: This should have a cast (`(c3_y)rec_d[0]`) for clarity.

pkg/urbit/vere/io/ames.c
```diff
@@ -721,21 +1125,26 @@ u3_auto*
...
-  if ( c3y == sam_u->fak_o ) {
+  if ( c3y == sam_u->pir_u->fak_o ) {
     u3_Host.ops_u.net = c3n;
   }
 
+  //  scry the protocol version out of arvo
+  //
+  u3_lord_peek_last(pir_u->god_u, u3_nul, c3_s2('a', 'x'), u3_nul,
+                    u3nt(u3i_string("protocol"), u3i_string("version"), u3_nul),
+                    sam_u, _ames_prot_scry_cb);
```
joemfb, line 1147: This should be moved to `_ames_io_talk()`.

pkg/urbit/vere/io/ames.c
```diff
@@ -30,21 +33,58 @@
...
+    c3_y* con_y;                        //  (jam [origin content])
+  } u3_body;
+
+/* u3_panc: deconstructed incoming packet
+*/
+  struct _u3_panc {
+    u3_ames* sam_u;                     //  ames backpointer
+    u3_panc* pre_u;                     //  previous packet
+    u3_panc* nex_u;                     //  next packet
+    u3_noun  ore;                       //  origin lane
```
joemfb, line 83: This could be `u3_lane`, which would move all nouns after the `ames.c` state.

pkg/urbit/vere/io/ames.c
```diff
@@ -30,21 +33,58 @@
...
     c3_d          fal_d;                //  crash count
     c3_w          imp_w[256];           //  imperial IPs
     time_t        imp_t[256];           //  imperial IP timestamps
     c3_o          imp_o[256];           //  imperial print status
+    c3_o          see_o;                //  can scry
+    c3_o          fit_o;                //  filtering active
+    c3_y          ver_y;                //  protocol version
+    c3_d          vet_d;                //  version mismatches filtered
+    c3_d          mut_d;                //  invalid mugs filtered
+    u3_panc*      pac_u;                //  packets pending forwards
```
joemfb, line 52: There's no need to repeat the typedef above, `struct _u3_panc*` should work here.

joemfb:
@Fang, my apologies for the slow review!

pkg/urbit/vere/pier.c
```diff
@@ -60,13 +98,32 @@ _pier_work_send(u3_work* wok_u)
   {
     u3_ovum* egg_u;
     u3_noun    ovo;
+    u3_pico* pic_u;
 
-    while (  len_w-- && car_u && (egg_u = u3_auto_next(car_u, &ovo)) ) {
+    while ( len_w && car_u && (egg_u = u3_auto_next(car_u, &ovo)) ) {
+      len_w--;
```
Fang-, line 104: Initial test on ~nus resulted in the same old unresponsive ship, seemingly exclusively busy with packet forwarding. After this specific change, the ship is responsive again.

The commit message doesn't mention this, but I now think this was just integer underflowing in the `len_w == 0` case, leading to enqueuing a ton of scries further down in this function. (Of course, this also fixes handling of the "no next ovum" case, but that was less terribly wrong.)

vvisigoth:
nice
—
~poldec-tonteg
http://urbit.org

pkg/urbit/vere/pier.c
joemfb: Good catch! I got a little too cute with the one-liner ...

pkg/urbit/vere/pier.c
```diff
@@ -60,13 +98,32 @@ _pier_work_send(u3_work* wok_u)
...
       car_u = egg_u->car_u;
+
+      //  interleave scry requests
+      //
+      if ( len_w
+        && (pic_u = _pier_peek_next(pir_u)) )
+      {
+        len_w--;
+        u3_lord_peek_pico(god_u, pic_u);
+        c3_free(pic_u);
```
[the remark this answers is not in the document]
joemfb: It is a bit odd. The pattern I've followed for objects with a lifetime that spans multiple modules is to put constructors/destructors (init/free functions) in `ward.c`. A `u3_pico` constructor has a variable number of arguments, so you'd probably just want to `c3_calloc()` it in the constructor. The destructor would also `u3z` all noun references in the struct, so you'd have to `u3k` what you use in `lord.c`. And `_pier_peek_new()` would become `_pier_peek_plan()` and only add the struct to the linked list.

Fang-:
> This PR has been open a long time, and I've asked you for a lot of things that are borderline out-of-scope... I have one more

Unbelievable, this guy... (;

Let's talk about this after the music club, if you have time!

pkg/urbit/vere/lord.c
```diff
@@ -349,6 +349,8 @@ _lord_plea_peek_bail(u3_lord* god_u, u3_peek* pek_u, u3_noun dud)
 {
   u3_pier_punt_goof("peek", dud);
 
+  pek_u->fun_f(pek_u->ptr_v, u3_nul);
+
```
joemfb, line 353: You'll also need to remove the `_lord_bail(god_u);` line at the end of this function.

pkg/urbit/vere/io/ames.c
```diff
@@ -232,6 +297,106 @@ u3_ames_encode_lane(u3_lane lan) {
...
+    //  now we can serialize the head
+    //
+    u3_head* hed_u = &pac_u->hed_u;
+    c3_w     hed_w = hed_u->ver_y
+                   | (hed_u->mug_l << 3)
+                   | (hed_u->sac_y << 23)
+                   | (hed_u->rac_y << 25)
+                   | (hed_u->enc_o << 27);
+
+    memcpy(pac_y, &hed_w, 4);
```
joemfb, line 391: This assumes little-endian, and should be commented as such. (Alternately, the 4 bytes could be individually shifted/masked off of `hed_w` and assigned to `pac_y`.)

pkg/urbit/vere/io/ames.c
```diff
@@ -445,39 +681,161 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
+        _ames_forward(pac_u, u3nc(u3nc(c3y, (c3_y)rec_d[0]), u3_nul));
+      }
+      //  otherwise, if there's space in the queue, scry the lane out of ames
+      //
+      else {
+        u3_noun pax = u3nq(u3i_string("peers"),
+                           u3dc("scot", 'p', u3i_chubs(2, rec_d)),
+                           u3i_string("forward-lane"),
+                           u3_nul);
+        u3_pier_peek_last(sam_u->pir_u, u3_nul, c3_s2('a', 'x'),
```
joemfb, line 826: You should add `c3__ax` to `motes.h`.

pkg/urbit/vere/io/ames.c
```diff
@@ -601,6 +985,12 @@ _ames_io_talk(u3_auto* car_u)
 
     u3_auto_plan(car_u, u3_ovum_init(0, c3__a, wir, cad));
   }
+
+  //  scry the protocol version out of arvo
+  //
+  u3_pier_peek_last(car_u->pir_u, u3_nul, c3_s2('a', 'x'), u3_nul,
```
joemfb, line 991: `c3__ax` here too.

PR #3178, opened 2020-07-24
chat-hook: recreate all foreign dms as local dms

### LINKED ISSUE #3164: chat: migrated direct messages only function for owner
- If you owned the DM, the other user(s) appear in the members list and appear normally. You can send messages. Because they post to your ship, it appears functional.
- Other users in the DM cannot post any messages. They don't receive any, either.

This is a difficult problem to even isolate because if you delete the DM as the owner, the chat persists in their sidebar, and reinviting them will just forward to the pre-existing inbox with the same issues.

You will have to mutually delete your DMs and re-establish them to get them working again.

tacryt-socryp:
About to test this, but putting out for review early. Should fix #3164.

Update: Tested, and confirmed that it is not destructive to working DMs.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -399,10 +399,16 @@
     ~
   =/  =ship  (slav %p i.t.path)
   ?:  =(ship our.bol)
+    ::  local dm, no need to do cleanup
+    ~
+  ?:  ?=(^ (groups-of-chat path))
+    ::  correctly initialized, no need to do cleanup
+    ::
     ~
   :-  =-  [%pass /fixdm %agent [our.bol %chat-view] %poke %chat-view-action -]
       !>  ^-  action:store
       [%delete path]
+  =/  new-dm  /(scot %p our.bol)/(crip (weld "dm--" (trip (scot %p ship))))
   =/  mailbox=(unit mailbox:store)  (chat-scry path)
   ?~  mailbox
     ~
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/chat-hook.hoon
```diff
@@ -383,6 +385,52 @@
...
+  ?:  =(ship our.bol)
+    ::  local dm, no need to do cleanup
+    ~
+  ?:  ?=(^ (groups-of-chat path))
+    ::  correctly initialized, no need to do cleanup
+    ::
+    ~
+  :-  =-  [%pass /fixdm %agent [our.bol %chat-view] %poke %chat-view-action -]
+      !>  ^-  action:store
+      [%delete path]
```
Fang-, line 410: I'm likely missing some context here, so maybe these are dumb questions, but:

Am I reading this correctly in that it will fully delete history for affected DMs?

https://github.com/urbit/urbit/blob/c96705e755f424e90887c656c452ada28797806b/pkg/arvo/app/chat-view.hoon#L235-L240

What's stopping us from keeping them around (in the store, at least), and/or rejoining the foreign DMs?

tacryt-socryp: Note that we scry out the messages and move them to the newly created local DM.

Fang-: Oh oops, I missed that. What about reconnecting to the foreign DMs though? Is that off the table because of permission migration issues or similar?

PR #1661, opened 2019-08-25
zuse: make of-wain:format use linear space (or "Make commits fast")

philipcmonk:
You know how writing large files to Clay seems to take forever?  It's not your imagination, hoon.hoon takes 19 seconds on my machine. This PR reduces that by 93% to 1.3 seconds.

Turns out `+of-wain:format` used quadratic space.  It converts a list of lines into one big atom.  Its algorithm was more or less:

- set result = 0
- if no more lines, return result
- else, concatenate result and the next line, then recurse

Of course, each time it hits step 3 (once per line), it allocates a new big atom.  For hoon.hoon, that's 17000 allocations and writes, averaging 250KB each, for a total allocation/write of 3.4 GB.

Memory-wise, the solution is to do a single allocation and paste in all the lines at once.  Hoon doesn't let us access memory directly, but vere has a jet for `+can` which does this.  This is a good example of a virtuous jet:  while it doesn't reduce the time complexity, it does reduce the space complexity (write complexity?) from quadratic to linear.

`+of-wain:format` was used in several other places, including pretty much any mark that gets written as plain-text to unix.  These should speed up, but it'll probably be less noticeable because the files are smaller.

PR #466, opened 2015-10-28
fix ++lore and ++role

chc4:
Fixes both the ++lore jet and both arms in hoon.hoon. I'd like someone to verify that I didn't do something terribly wrong with the jet, since it's my first one commited.

(Does this pull request have a pill attached? I solided one and added it to the commit, but it's not showing the github UI...)

Fixes #241

PR #3180, opened 2020-07-24
vere,arvo: accept embedded nulls in to-wain:format

frodwith:
Prior to this commit, there was a jet mismatch in to-wain (formerly
called lore, and still jetted under that name). 0 bytes in the middle of
a cord caused the jet to crash, whereas the hoon simply treated them as
the end of cord and truncated the output. The history of this behavior
is fraught with controversy. This commit rectifies the current mess with
the following rationale: Null bytes are valid ASCII/UTF-8, and \n\n in
the input will cause null list items in the output, so nulls are (for
the purposes of to-wain) allowed in cords. Trailing nulls cannot be
represented because of the nature of atoms, but that is outside the
scope of to-wain's concern. Therefore to-wain should simply measure the
cord and split on newlines, and do nothing fancy at all with nulls.

In addition, the hoon for to-wain was written in an inefficient style
that produced a lot of intermediate garbage atoms via rsh and cat. This
commit's implementation measures once and cuts once, so to speak, and so
avoids the intermediate garbage.  Quick benchmarks suggest it is about
20x faster than the old hoon, but still orders of magnitude slower than
the jetted code. to-wain is the workhorse for the txt mark, so we should
still prefer to have a jet.

The old jet is left wired up under %lore, and should be removed when
support for the old, unupgraded zuse is no longer necessary. A new jet
with matching null handling has been wired up under the name %leer.


pkg/urbit/jets/e/lore.c
```diff
@@ -3,6 +3,30 @@
...
+  _lore_cut(c3_w pos_w, c3_w len_w, u3_noun src)
+  {
+    if ( 0 == len_w ) {
+      return 0;
+    }
+    else {
+      // as an optimization, we alloc small lines on the stack
+      // (only small to prevent stack overflow)
+      c3_t  big_t = len_w > 255;
+      c3_y* buf_y = big_t ? u3a_malloc(len_w) : alloca(len_w);
```
philipcmonk, line 16: this is the perfect chance to introduce a "technical" reason for the 80 column rule!

ohAitch:
NUL is not valid @t, and whatever cast a bytestring containing null to @t is making a type error. Like, `((sane %t) 'text\00more-text')` is false, this is not ambiguous. 

NUL is of course valid ascii/utf-8, but @t is not "ascii/utf-8", it's "utf-8, [minus control characters <32, with the exception of \n](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/hoon.hoon#L4289)". \n\n produces empty strings, which are not the same thing as strings containing (nonexclusively) null bytes.

The hoon implementation should imo be changed to match the jet: it's being passed invalid input, and as formally crashing is allowed it's preferable to sinking in the "to/from wain round trip except if you had a null followed by a newline in which case oops" special case you mentioned. What was the motivating case for the new behavior?

ohAitch:
Note that I do support `\xx` string syntax staying in, even though it's primarily useful for constructing invalid @t, because hoon as a language plays fast and loose with auras sometimes, and that's okay. But a function rejecting input outside its domain seems legitimate and where-feasible desirable behavior - it's the jet mismatch that's the bug, because jets should not mismatch even on incorrectly typed data, including outright using raw nock to call `to-wain` on a *cell*. (Which I think this has previously come up with, like, not checking a `bloq` was a direct atom)

frodwith:
@joemfb tells me that this "no embedded nuls" standard is inconsistently applied in other places, and after discussion we agreed that this was least likely to do something the user wasn't expecting.

I noticed the jet mismatch while working on jets for cl-urbit. The only motivation for the change in behavior is rectification of the mismatch. I don't have a strong opinion on whether it should crash, but the jet and the hoon should definitely match :)

To take @ohAitch's side for a moment, the sample is @ and the return type is (list @t). That suggests that while the sample may contain embedded nuls, the elements of the returned list should be validated for (sane %t)-ness. Can someone make a ruling on this?

ohAitch:
Hm, that seems to be even *more* hardcoding in "nulls immediately before newlines are special and silently stripped".

What are some other places that, like, are making decisions based on the nulls? My bias is definitely "when the jet and the hoon substantially different, fix the hoon, since the jet is what the rest of the system has been written and tested against; separately, if you hit a bug in jetted code fix both"

ohAitch:
Also, if you change the sample from @ to @t, what type-errors does the build report? Because @t is defintiely, like, the prima facie sample I would expect.

joemfb:
@ohAitch, I'm sympathetic to this argument (and believe I made it myself the last time this issue came up in #2195). But I'm no longer convinced.

I think considering null bytes exclusively is wrong here. As you point out, all sorts of atoms are not valid `@t` per `(sane %t)`? Should `of-wain` validate utf-8?

I believe the reason nulls were considered at all in this code is simply that they were used to check for the end of the input. There's no reason to use that pattern with an atom, and once you are traversing based on a measurement, there's no reason to specifically consider nulls at all.

I think it's important that we develop a consistent story around string validation (`@t`, `@c`, `tape`, &c), which we currently do not have at all. I have a branch somewhere with an improved implementation of `+sane`, and I'd like to integrate (and jet!) that, and start aggressively calling it. Then we could work
on cleanup up various operations on `@t` and `@c`, some of which choke on invalid input, while others do not. But barring such a comprehensive effort, I see no reason to perform a very limited validation step in this particular function.

joemfb:
While we're at it, what are `+lune`/`+nule` supposed to accomplish that `+to-wain`/`+of-wain`? Just the enforcement of a trailing newline? FWIW, they're both currently unused.

joemfb:
#1661 is related (operationally, as concerns efficient atom traversal). #466 is relevant history.

joemfb:
Also, per `+ruth`, we should really be calling `+sane` in all the atom clam gates. I'd be interested to try that, once `+sane` is cleaned up and jetted.

ohAitch:
No, I do not think of-wain has a *duty* to validate its input; I think it has a choice to either treat it as valid input or report an error, as convenient. And that the convenient thing to do here is to report embedded nulls as an error, because otherwise (of-wain (to-wain)) would not round trip, and contra "my side" to ignore all other validation concerns, because they are not directly related to its task of "produce a list of lines, that comprise all input bytes sans 0x10 bytes which are interpreted as delimiters".

(list @t) cannot encode lines with trailing null bytes, therefore it should not purport to have converted a @t with line-trailing null bytes correctly; I suppose we could try to catch specifically \~line-trailing\~ null bytes, but that seems kind of silly, since they're not supposed to be there in the first place.

ohAitch:
ruth, extension of `sane` to non-text types, etc are commendable; but I do not see why they should gate correct functionality of this gate, which ideally one would be able to quickcheck as `(of-wain (to-wain a))` for all nouns `a` producing either `a` or an error.

joemfb:
That's a pretty compelling argument -- nock-level bijection in these kinds of conversion gates seems worth pursuing. Alternatively, we could better document the limitations of these gates by adjusting the types from `$-(@ (list @t))` to `$-(@t (list @t))` and the respective inversions.

@philipcmonk, what do you say?

ohAitch:
Changing the type seems pretty obvious until we either remember or hit the reason it wasn't already @t, whatever the null behavior

philipcmonk:
I would like `{to,of}-wain` to invert each other.  We've gone back and forth on that, particularly as it relates to trailing newlines, but I believe that has been resolved.  `+lune`/`+nule` were another way of handling the trailing newline.

The argument for specially rejecting nulls (or nulls-at-end-of-line) is compelling -- they're the only thing that breaks the bijection.  I'm fine with either of those last two solutions (rejecting null or taking `@t` input) or both.

frodwith:
I'm taking the position that a `@t` sample is sufficient. One of the advantages of having types is not compiling checks, because it is faster not to check.

It might be desirable to explicitly cast the product of ++of-wain to @t to emphasize the inversion, but I haven't included that.

joemfb (changes requested):
I support these changes in their current state. For release staging purposes, I have to ask that this PR be split into two, separately targeting arvo and vere. These changes will go into separately tagged releases (released from separate branches). And we'll want to release the new jet ASAP, so that users will be more likely to have it when the hoon changes are OTA'd.

This PR could stay open for one half. And it's probably easiest to make a second branch from this one and then interactive-rebase/edit the commits on each side. I'd be happy to help with staging.

joemfb:
Closing in favor of #3523 and #3525.

PR #3181, opened 2020-07-24
Unified SidebarInvite component

tylershuster:
This PR deduplicates the sidebar invite component and converts it to TypeScript.

Instead of explicitly declaring behaviors for `onAccept` and `onDecline`, the new component takes `Function` props for them. It tries to show the `text` property of the `invite: Invite` prop and if not available shows the `path` property instead.

In addition, it `position: sticky`s the invite to the top of the sidebar, which prevents invites from being ignored if the user has scrolled too far down the sidebar, which is what I started this PR to do in the first place.

I think this adequately accounts for the current uses of sidebar-invite, though there is certainly more work that could be done.


matildepark:
Oh hey dope. You may want to get into the practice of not PRing from your fork's `master` branch, that prohibits you from doing other stuff.

tylershuster:
Oh hey yeah you definitely caught me there.

matildepark:
Cc @g-a-v-i-n re: indigo-react

I had an immediate hesitant response because we're about to refactor the codebase in indigo-react components but this aspect of componentisation I think might be actually fine. It's higher order than the work we'll be doing?

pkg/interface/src/components/SidebarInvite.tsx
```diff
@@ -1,33 +1,26 @@
...
-    this.props.api.invite.decline('/chat', this.props.uid);
-  }
-
+export class SidebarInvite extends Component<{invite: Invite, onAccept: Function, onDecline: Function}, {}> {
   render() {
     const { props } = this;
 
     return (
-      <div className='w-100 bg-transparent pa4 bb b--gray4 b--gray1-d'>
+      <div className='w-100 bg-white pa4 bb b--gray4 b--gray1-d z-5' style={{position: 'sticky', top: 0}}>
```
liam-fitzgerald, line 9: probably needs `bg-gray0-d` as well. `position: sticky` is a nice touch

tylershuster:
@liam-fitzgerald Good tip. I added the dark mode background.

PR #3182, opened 2020-07-24
serf: adds thaw: a recovery command to jam/cue cold jet state

joemfb:
This PR adds `thaw` - an experimental memory recovery command - to the urbit-worker binary.

The "cold" state in the jet dashboard abstractly represents all registered jets. In an implementation that supports `%fast` hints (dynamic jet registration), the cold state is derived from the execution history, and effectively permanent. It is, therefore, a kind of intentional space leak. `u3` includes *hashboard* -- a dashboard implementation with pre-registered (out-of-band) jets, matched and dispatched by cryptographic battery hash. That general approach (modulo some details, and the choice of value or cryptographic hash) is obviously correct, and the only long-term solution. But we're currently just using fast hints, for a variety of short-term reasons.

Given sufficient care, it seems possible to treat jet registrations as weak references to the registered cores, and prune the cold state as batteries "expire". But this is tricky to implement in an environment without actual weak references (or guaranteed unification/interning), and there are some open conceptual questions as well. In the meantime, `thaw` should be able to help.

It works by simply jamming and cue'ing the cold state, which is currently not at all unified. In other words, it's like the old `|pack` (now `cram` and `queu`), but only for the cold state. By working with less state, it should be usable in more scenarios then the old `|pack`, but it retains the fundamental weaknesses of that implementation.

The end-result of `thaw` is a completely unified cold state, completely dis-unified from any and all actively used cores. Counterintuitively, this means it may increase memory usage on ships that did not have many old registrations. The registrations will be incrementally re-unified with active cores as those cores are registered, or can be re-unified in one shot with `cram` and `queu`.

In other words, this is an experimental tool that may relieve memory pressure on ships under a particular kind of duress, but is otherwise probably not generally useful.

/cc @frodwith

### DIFF
--- a/pkg/urbit/worker/main.c
+++ b/pkg/urbit/worker/main.c
@@ -288,6 +288,143 @@ _cw_queu(c3_i argc, c3_c* argv[])
   }
 }
 
+/* _cw_thaw(); jam/cue cold jet dashboard
+*/
+static void
+_cw_thaw(c3_i argc, c3_c* argv[])
+{
+  c3_assert( 3 <= argc );
+
+  c3_c* dir_c = argv[2];
+  c3_w  pre_w;
+
+  u3m_boot(dir_c);
+
+  pre_w = u3a_open(u3R) + u3a_idle(u3R);
+
+  // u3_serf_grab();
+  // u3a_print_memory(stderr, "urbit-worker: thaw: open", u3a_open(u3R));
+  // u3a_print_memory(stderr, "urbit-worker: thaw: idle", u3a_idle(u3R));
+
+  u3m_reclaim();
+
+  fprintf(stderr, "urbit-worker: thaw: reclaim\r\n");
+
+  // u3a_print_memory(stderr, "urbit-worker: thaw: open", u3a_open(u3R));
+  // u3a_print_memory(stderr, "urbit-worker: thaw: idle", u3a_idle(u3R));
+
+  {
+    u3p(u3h_root) bak_p;
+    u3_noun cod = u3j_stay();
+    c3_d  len_d = u3s_jam_met(cod, &bak_p);
+    c3_w  byt_w, wor_w, len_w;
+
+    fprintf(stderr, "urbit-worker: thaw: measured cold state %" PRIu64 "\r\n", len_d);
+
+    if ( len_d > 0xffffffffULL ) {
+      fprintf(stderr, "urbit-worker: thaw: overflow c3_w: %" PRIu64 "\r\n", len_d);
+      exit(1);
+    }
+
+    //  length in bytes a la u3i_bytes
+    //
+    byt_w = (c3_w)(len_d >> 3ULL);
+    if ( len_d > (c3_d)(byt_w << 3) ) {
+      byt_w++;
+    }
+
+    //  length in words
+    //
+    wor_w = (c3_w)(len_d >> 5ULL);
+    if ( len_d > (c3_d)(wor_w << 5) ) {
+      wor_w++;
+    }
+
+    //  byte-length of word-length
+    //
+    len_w = 4 * wor_w;
+
+
+    {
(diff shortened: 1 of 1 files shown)


philipcmonk:
In principle this makes sense.  I tried running it on a -L copy of ~wicdev which had 274MB in its cold jet state, and it says it lost 22MB, then when I start it up again, it goes really really slow (like 50x slower, but it does make progress), with sampled backtraces in _song_x_cape in u3r_sing in _ch_node_git in _cj_sink.  IOW, it's looking up jet registrations and having to trace out those comparisons.  I don't think it'll unify them either because it's on the home road.

joemfb:
What's the cold state after (per `|mass` or `urbit-worker grab $pier`) after it ran?

Those printfs are confusing; "lost 22MB" means that the overall memory usage *grew* -- as opposed to "reclaim X". Those numbers are calculated based on free space (open + idle), but I just realized I could give more precise measurements with your count/discount functions ...

I saw something similar on a local test ship: the cold state went from 6MB to 26MB. But it promptly reunified them and went right back to 6MB (it's possible that such unification is only happening due to re-registration though). I expect that kind of behavior on a new ship where the registrations are basically all active, but would've expected much better numbers on a ship with stale registrations.

I can make copies of some struggling infrastructure nodes and experiment further. The tradeoffs may not be worth it here -- I just whipped this PR together since it was fairly obvious how to do so, and seemed like it could help ...

philipcmonk:
Before:

```
total arvo stuff: MB/719.850.036
  warm jet state: KB/99.044
  cold jet state: MB/274.566.124
  hank cache: B/288
  battery hash cache: B/288
  hot jet state: KB/126.364
total jet stuff: MB/274.792.108
  bytecode programs: KB/456.188
  bytecode cache: KB/51.832
total nock stuff: KB/508.020
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
total marked: MB/995.150.488
free lists: MB/953.569.996
sweep: MB/995.150.488
```

After (and after letting it run for like 10 minutes):
```
...
total userspace: MB/702.517.904
  kernel: KB/211.272
total arvo stuff: KB/211.272
  warm jet state: KB/97.152
  cold jet state: MB/253.445.492
  hank cache: B/372
  battery hash cache: B/288
  call site cache: B/100
  hot jet state: KB/126.372
total jet stuff: MB/253.669.776
  bytecode programs: MB/1.195.432
  bytecode cache: KB/108.884
total nock stuff: MB/1.304.316
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
space profile: KB/9.084
total marked: MB/957.712.676
free lists: MB/979.986.788
sweep: MB/957.712.676
```

Output of run:

loom: mapped 2048MB
boot: protected loom
live: loaded: GB/1.937.719.296
boot: installed 268 jets
urbit-worker: thaw: reclaim
urbit-worker: thaw: measured cold state 273365722
urbit-worker: thaw: jammed cold state
urbit-worker: thaw: free'd jet state
urbit-worker: thaw: allocated atom
urbit-worker: cue'd cold state
urbit-worker: booted jet dashboard
urbit-worker: loaded cold state
urbit-worker: reamed jet dashboard
urbit-worker: reamed bytecode state
urbit-worker: thaw: lost: MB/22.019.860
```

joemfb:
That's significantly less unification than I expected. Outside of variation in`%spot` hints, I don't have any explanation for why that would be the case.

Thanks for checking, this might be a dud.

philipcmonk:
I wonder if the variation in the spot hints is actually important.  If there's anything deep in hoon.hoon with a !: or something, then the whole kernel is probably not actually important.

joemfb:
There's 5, and `|ar:ut` is fairly deep ...

joemfb:
Closing in favor of #3235

PR #3183, opened 2020-07-24
interface: add theme placeholders

matildepark:
Adding placeholder themes for current Indigo usage in Landscape so that our refactor work will match. The main difference is the predominance of `#333` as a background colour in dark mode.

(Once we finish refactoring, we can unveil `@tlon/indigo-light` and the like and update our UI...)

The light theme is fairly untouched because its additions should fit in a bit better, but we can tweak.

cc @g-a-v-i-n


matildepark:
ping @tacryt-socryp

matildepark:
It just eliminates a bug where indigo-react versioning breaks builds while letting us start refactoring.

PR #3184, opened 2020-07-24
king: use c vere's heuristics for finding the serf binary

### LINKED ISSUE #3172: kh: Running urbit-king didn't use the urbit-worker that I downloaded
I had an `urbit-worker` in `~/.nix-profile/bin`, which is in my path, so when I ran `./urbit-king run wicdev` it tried to use that one instead of the downloaded one.  It worked when I ran it as `PATH=.:$PATH ./urbit-king run wicdev`.  Maybe we should check the current directory even if it's not in the `$PATH`?

cc @pilfer-pandex

pilfer-pandex:
Fixes #3172

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
@@ -25,6 +25,8 @@ import Urbit.King.App
 import Urbit.Vere.Pier.Types
 
 import Control.Monad.STM      (retry)
+import System.Environment     (getExecutablePath)
+import System.FilePath        (splitFileName, (</>))
 import System.Posix.Files     (ownerModes, setFileMode)
 import Urbit.EventLog.LMDB    (EventLog)
 import Urbit.King.API         (TermConn)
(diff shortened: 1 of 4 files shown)


joemfb:
@pilfer-pandex, I took the liberty of retargeting this against (the closest thing to) the appropriate release branch.

pilfer-pandex:
So it'll go out before ipc-redux and get merged into same?

—
~pilfer-pandex

pilfer-pandex:
WRONG. Do not merge.

pilfer-pandex:
Apparently, haskell provides a function for what we want! http://hackage.haskell.org/package/base-4.14.0.0/docs/System-Environment.html, implementation https://gitlab.haskell.org/ghc/ghc/-/blob/master/libraries/base/System/Environment/ExecutablePath.hsc.

The previous implementation I gave here was very wrong because getArgs in Haskell always returns the argument list *excluding* the program, in an ostensible attempt to be platform-independent.

Anyway, there are still some problems with properly threading the arguments into the serf, so I'm putting this into draft so it doesn't accidentally get merged.

pilfer-pandex:
I retracted this as draft last week because I ran into issues running the command

```
./urbit-king -F zod
```

wherein urbit-worker complains that it has the wrong number of arguments.

Silly me, this is actually not an issue (except for the confusing error message), because the correct command to run was in fact

```
./urbit-king zod -F zod
```

and the "issue" I describe also happens on master., lol.

Accordingly, I reintroduce this PR without change.

pilfer-pandex:
Update: the actual problem was that the serf had its arg layout changed, but I was using an old serf binary. The substance of the previous comment remains: this is good to go.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -122,17 +124,25 @@ runSerf
...
     , scStdr = \txt -> slog (txt <> "\r\n")
     , scDead = pure () -- TODO: What can be done?
     }
+  getSerfProg :: IO FilePath
+  getSerfProg = do
+    (path, filename) <- splitFileName <$> getExecutablePath
+    pure $ case filename of
+      "urbit"      -> path </> "urbit-worker"
+      "urbit-king" -> path </> "urbit-worker"
+      _            -> "urbit-worker"
```
joemfb, line 145: The current vere logic is `concat(argv[0], "-worker")`. If we end up running in king-haskell in a similar manner to the current CI debug mode (wherein the binaries are prefixed with `debug-` or something), we'll need to revisit this logic.

I'm not sure we ever want to fallback to looking up `urbit-worker` in `$PATH`. But that code is not going to run in any current deployment scenario, so we can work out that behavior later.

ohAitch:
How obnoxious would it be to try to duct tape them both into the same
binary, and run one from a `--worker` argument mode?

joemfb:
That'd be pretty trivial for the *vere* king/serf (they actually used to be the same binary). I'm sure it's possible to do that with the haskell king and vere serf, through some kind of FFI, but it'd probably be a major pain. libu3 is not exactly friendly to embedding ...

PR #3185, opened 2020-07-24
kiln: get germ for actual desk

philipcmonk:
This causes new ships to get a `%merge-no-merge-base` error after downloading the initial ota while merging into their own kids desk.  Not that big a deal since it merges correctly into their home desk.  For anyone who runs into this, you can workaround by manually running `|merge %kids (sein:title our now our) %kids, =gem %that`.


philipcmonk:
I fixed this in `philip/clay` (new protocol) the other day, maybe that was it.

PR #3188, opened 2020-07-25
clay: Refactor over-the-wire protocol to reduce network and memory usage

philipcmonk:
In Clay on the network now, any time we sync from a new desk, we have to download the entire history of that desk, even if know most of it because we've synced from a related desk.  Nearly all desks share some history, and most share almost all history.  Subsequent fetches from that desk do make sure to only send data that it hasn't already sent you, but the list of blobs which have been sent is not shared between desks or ships.

This change makes it so that when we fetch a desk (technically: make a %many request, which is caused by a `%cv` request to a foreign ship for a time/version after the time when we last got a full update from that desk), we first fetch the commits without the blobs, then we scan through the referenced blob hashes and separately request each of the blobs which we don't already have in our object store.  If any of those blobs turn out to be deltas on other blobs, we recursively download the parent if we haven't got it in our object store.  Once all the referenced blobs are in the object store, we apply the commits and notify subscribers.

The state machine is very precise, but I don't think it's fragile.  I'll do more testing, particularly around delta blobs, but I've tested this on fake ships as well as on livenet moons, starting from a master pill and after receiving the OTA from the planet.  This tested the upgrade process, which worked fine.

Regarding the upgrade process, we add a version number to `riff` (over-the-wire request).  If the version number is omitted, we assume the subscriber doesn't speak the new protocol, so we send all blobs eagerly, exactly as on the network now.  If the version number is `%1`, then we know we can send just the commits and wait for them to ask for the blobs.

Note this upgrade process does not work in reverse.  If a ship which speaks the new protocol needs to sync from a ship which does *not* yet speak the new protocol, the request will fail since it isn't expecting a version number.  I consider this to be acceptable since this is an uncommon case which will be resolved by the upstream ship updating, but it's worth noting.

This change has the following resource usage benefits:
- Reduced network traffic, since already-known blobs won't have to be sent again
- Reduced memory usage on the sender's side, since they won't have to queue up a giant update (130MB at time of writing) to every new subscriber
- If the subscriber does not have any of the data, the sender will still have to send 130MB, but it still doesn't have to queue it up at once.  Since the subscriber asks for blobs individually, the peak ames memory usage for each subscriber will be the size of the largest file that must be sent.
- If the subscriber requests a giant update and then goes offline before finishing the download, we are no longer stuck with a 130MB blob of data in our ames queue that we'll never be able to get rid of.

I believe this will be particularly impactful for 3rd party apps.  If you install a 3rd party app by merging from their desk, you no longer have to download 130MB plus the data in the app; rather, you only download the new and changed files.

Further work:
- We could add a snapshot of ~zod's `%kids` desk (or just the blobs in it) next to the pill so the initial OTA is faster
- Requesting a blob by hash could be optimized by having intermediate ships recognize they don't need to forward the request; rather, they can answer it themselves.  This is related to the idea of "named data networking".  Relatedly, this would be a reasonable place for a distributed hash table.


ohAitch:
Re: having blob data cached in the pill, weren't we already doing this with
the textual file contents? In which case the blob hashes should line up,
and not need to be re-requested. But possibly I'm misremembering something.

Re: recursive fetching, how does git handle this? I'm pretty sure it's a
request->metadata->response_metadata->blobs flow with no further back and
forth. Perhaps when the server gets a request for a blob it has stored as a
delta, it should just materialize it to a direct blob, unless its parent is
also being requested; with the delta-parent sent along as metadata so the
compression can be reconstructed if it is available, and maybe some
common-case optimization around "client sends along the most recent
revision of the desk it has, which it expects deltas to be against"

philipcmonk:
> Re: having blob data cached in the pill, weren't we already doing this with
> the textual file contents? In which case the blob hashes should line up,
> and not need to be re-requested. But possibly I'm misremembering something.

Yes, but the initial filesystem is about 4MB now and pretty stable, while the whole history is 130MB and linearly growing (even if slower now that %glob is deployed).

> Re: recursive fetching, how does git handle this? I'm pretty sure it's a
> request->metadata->response_metadata->blobs flow with no further back and
> forth. Perhaps when the server gets a request for a blob it has stored as a
> delta, it should just materialize it to a direct blob, unless its parent is
> also being requested; with the delta-parent sent along as metadata so the
> compression can be reconstructed if it is available, and maybe some
> common-case optimization around "client sends along the most recent
> revision of the desk it has, which it expects deltas to be against"

It would be interesting to look how git does this.  I suspect materializing (when can't prove they have the parent) is a good choice if you have a routine on the other side for "try to compact my object store by judiciously dematerializing", which is on my list as something I want to build soon.

ohAitch:
Hah, unpopular opinion: do initial merges from /(sein)kids= actually need
all the history? Esp if you're unrolling deltas to present snapshots.

"git clone --depth=1", if you will. Especially if it saves you a factor of
30 space unrelated to states your ship has actually observed!

philipcmonk:
Well, you want the history so that if you merge another branch you can diff against a real merge base.  But of course the mergebase is in the other branch's history, so you could get that commit (and only that commit) at the time of merge, as long as you have commit ancestry.  Does `git clone --depth=1` give you all the commits just not all the blobs?  Because I suspect that's what we'd want.

This falls under the general category of "tombstoned" clay commits, though perhaps that suggests too much permanency -- no reason you couldn't revive them by getting blobs back.  I hope we support this one day, it's not too hard in theory.

philipcmonk:
One thing I'll note is that sending blobs one at a time sacrifices any structural sharing between objects.  I don't believe there's any structural sharing to speak of between objects right now, but it is a limitation.

philipcmonk:
Here's the tests I've run, which I think is sufficient, except of course final tests at release time when merged with whatever else is going into the release.

- On fake ~zod and ~marzod, merge back and forth with `|ames-verb %rcv` on, making sure it's sending the appropriate blobs
- On fake ~zod and ~marzod booted from a master pill, commit changes to ~zod and syncing to ~marzod, to check the upgrade works correctly across the network.
- On real moons booted from a master pill and up to date from ~wicdev-wisryt (ie with all 130MB of real data), commit changes to one moon and on the other moon switch `|ota` to that one.  The merge back and forth between a few moons to verify correct behavior.  This tests correct behavior with the real on-the-network clay data.
- On real moons where the recipient was booted with a *new* pill.  This tests correct behavior of downloading the entire initial OTA on the new protocol, which is especially important because don't generate near so many `%delta` changes as before.
- I successfully ran `-ph-child-sync`, `-ph-child-update`, `-ph-change-sys`, and `-ph-breach-sync`.  Had to make a few unrelated changes to get upt to date with ford fusion.  Notably these work now without taking forever/increasing ames MTU, since it doesn't need to download any blobs except the ones that changed.  These are good because they're standard and easy to run, but they're not exhaustive and do not test applying these changes to a ship on a previous pill.  I actually think you could build a pH test for "boot with this pill, but then commit the new files" for a pretty generic OTA test.

This is a fairly thorough battery of tests, but for anything touching clay's protocol I wouldn't want to do any less.  Ford Fusion helps make sure stuff compiles, but it won't catch protocol issues.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -2765,12 +2883,19 @@
         $(aeon +(aeon))
       ::  produce updated state
       ::
+      =/  =rave  rave:(~(got by bom.u.ref) inx)
+      ?>  ?=(%many -.rave)
```
joemfb, line 2887: Small point of style, but I think the diff would be clearer if these lines were added in the `lim` update expression below, since that's the only place they're used.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3245,16 +3361,21 @@
     ::  ones we found before `a`.  Then convert the takos to yakis and also get
     ::  all the data in all the yakis.
     ::
+    ::  What happens if you run an %init merge on a desk that already
+    ::  had a commit?
```
joemfb, line 3365: The theoretical answer would seem to be "exactly the same thing as if you ran a `%that` merge, except the commits in your previous revision(s) have been orphaned. How does that fail in practice?

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3939,7 +4060,14 @@
     =^  for  req
       ?:  ?=(%warp -.req)
         [~ req]
-      :-  ?:(=(our who.req) ~ `who.req)
+      ::  ?:  =(our who.req)
+      ::    [~ [%warp wer.req rif.req]]
```
joemfb, line 4064: These lines can be deleted.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3957,8 +4085,14 @@
...
     ::
-    ?>  ?=({%question *} pax)
-    =+  ryf=;;(riff res)
+    ?:  ?=([%backfill *] pax)
+      =+  ;;(=fill res)
+      =^  mos  ruf
+        =/  den  ((de our now ski hen ruf) our desk.fill)
+        abet:(give-backfill:den +.fill)
+      [[[hen %give %done ~] mos] ..^$]
+    ?>  ?=([%question *] pax)
```
joemfb, line 4094: This could use a comment for vertical separation.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4303,6 +4521,35 @@
...
+      ::
+      =/  her=ship   (slav %p i.t.tea)
+      =/  =desk      (slav %tas i.t.t.tea)
+      =/  index=@ud  (slav %ud i.t.t.t.tea)
+      ::
+      =^  mos  ruf
+        =/  den  ((de our now ski hen ruf) her desk)
+        abet:abet:(take-backfill:(foreign-update:den index) blob)
+      [mos ..^$]
+    ==
```
joemfb, line 4551: This PR is not the time or the place... But I'd like to work through these error semantics with you sometime soon.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -2003,6 +2018,7 @@
       ::  bob's.
       ::
       ?:  ?=(%init germ)
+        ?>  ?=(~ bob-yaki)
```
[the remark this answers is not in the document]
philipcmonk: I, uh, forgot to address these review comments before releasing.  Oh well!

Yes it's a new invariant that `%init` can only be used for the first commit, but I think it was already load-bearing.  I don't think +data-twixt-takos works correctly with sufficient disconnectedness; I would expect downloading such desks to omit some relevant commits and blobs.  I have very occasionally (like 5 times in the last year?) heard of someone fetching a desk and somehow not having all the relevant blobs, and I wonder if this was the cause.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -2689,36 +2705,138 @@
...
+      ::  same desk).
+      ::
+      ?:  ?|  (~(has by lat.ran) i.need.sat)
+              (~(has by have.sat) i.need.sat)
+          ==
+        $(need.sat t.need.sat)
+      ::  Otherwise, fetch the next blob
+      ::
+      =/  =fill  [syd i.need.sat]
+      =/  =wire  /back-index/(scot %p her)/[syd]/(scot %ud inx)
```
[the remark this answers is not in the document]
philipcmonk: Yes each request is a new flow.  This is not new behavior.  The argument for putting them on separate flows is that some requests may be quite long while others are quick.  Consider if you request a large desk and, while it's downloading, you decide to request its `%cz` hash.  There's no reason those should interfere with each other.

pkg/arvo/sys/vane/clay.hoon
philipcmonk: The set of yakis produced is `+reachable-takos` (set of reachable ancestors via parentage -- not looking at aeons) of `b` (often the head of the desk) subtract `+reachable-takos` of `a` (usually the last commit we know for sure the foreign ship has).  Suppose there's an `%init` merge `c` between `a` and `b`, and `x` is the aeon before `c`.  Then, the descendants of `a` whose aeon is lte `x` will *not* be produced and sent over the wire, so the foreign ship will not get all the commits it needs.

pkg/arvo/sys/vane/clay.hoon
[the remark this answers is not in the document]
joemfb: All else being equal, sure. But flows are permanent, and this seems like unbounded proliferation. Since it's not strictly necessary, I'd be inclined to limit them to `/our/her/desk`. It's not new behavior, but it's worth nailing down these norms at some point.

pkg/arvo/sys/vane/clay.hoon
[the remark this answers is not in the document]
joemfb: No worries, none were intended to block release. It seems fine to limit `%init` to true initialization; `%that` covers the relevant "force merge" case for pre-existing desks.

PR #3192, opened 2020-07-27
Added error boundary

tylershuster:
[React Documentation](https://reactjs.org/docs/error-boundaries.html)

This is pretty much the most basic implementation of an error boundary, inspired by the Notebook Subscribers page bug.

I put an error boundary in the skeleton of every app (with a skeleton), and the boundary gets access to the router, so if a single route errors out you can still navigate away.

I think there's more work to be done here but this will keep the whole thing from crashing if there's an error inside one of the components.

For instance, this could go inside each tab of the notebook page, around the whole `<Content>` component, etc. it just seems that with things in the flux that they are this was the right scope to implement this.


matildepark:
I wonder if there's a cool way to integrate this with the 404 component, making it into a generic 'error' component that takes an error code, text and log and catches errors.

pkg/interface/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,46 @@
...
+class ErrorBoundary extends Component<
+  RouteComponentProps,
+  { error?: Error }
+> {
+  constructor(props) {
+    super(props);
+    this.state = { error: undefined };
+    const { history } = this.props;
+    history.listen((location, action) => {
+      if (this.state.error) {
```
liam-fitzgerald, line 14: How does this handle intra-app navigations? i.e. if the app crashes, I might click back which might keep me in the app that crashed?

pkg/interface/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,46 @@
...
+    if (this.state.error) {
+      return (
+        <div className="pa4">
+          <h1 className="mb4">Oh no</h1>
+          <p className="mb4">{this.state.error.message}</p>
+          <p>
+            <details>
+              <summary>Stack trace</summary>
+              <pre>{this.state.error.stack}</pre>
+            </details>
```
liam-fitzgerald, line 37: You probably want to make it such that you can navigate out of the app from inside the error trace.

pkg/interface/src/components/ErrorBoundary.tsx
tylershuster: If you 1) visit a working page, 2) visit a broken page and see the error and 3) click on the sidebar to a new page then 4) click the back button, you'll see the same error, yes. But it will have been recomputed because you navigated to a new page, the router calculated where it was supposed to be, and ran into an error *again*. 
The error boundary exists to *keep* the app from crashing — just to catch faulty components. If something outside the boundary crashes then tough luck — WSOD 4 U.

tylershuster:
@matildepark That seems possible. Like `Error<{code: number, error?: Error}>`?

tylershuster:
@matildepark @liam-fitzgerald Alright, good suggestions. I made the ErrorBoundary a 500 error code. It's maybe not technically correct but I don't think HTTP status codes were really made for this architecture. Let me know what you think.

matildepark:
Don’t give Urbit errors HTTP codes. The 404 issue is because it’s a valid 404 — file-server serves Landscape for all first level routes so it passes the catch to front-end. But Urbit JS crashes aren’t HTTP.

— 
~haddef-sigwen
https://urbit.org

tylershuster:
True.

matildepark:
I want to hold this and test this sometime tomorrow for UX stuff so this won't go out tonight, just btw.

tylershuster:
Good grief I'll stop now. How to merge commits

matildepark:
1. Go to localhost:9000/404
2. Click 'go back'
3. Go to groups and click on a group in sidebar
4. ErrorBoundary crashes React, ironically

image

tylershuster:
This was dumb. I used some boilerplate that included `React.Children.only` which crashes if there's more than one child. I think that's the last of the boilerplate, anyhow.

matildepark:
OK, I'm gonna try and instigate a crash. Might just edit some code to undo the subscribers change I guess, lol

pkg/interface/src/components/Error.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+class ErrorComponent extends Component<ErrorProps> {
+  render () {
+    const { code, error, history, description } = this.props;
+    return (
+        <div className="pa4 inter tc flex flex-column items-center justify-center w-100 h-100">
+          <h1 className="mb4 fw2 f2" style={{
+            fontFeatureSettings: '\'zero\' 1',
+          }}>
+            {code ? code : 'Oh No'}
```
matildepark, line 18:
```suggestion
            {code ? code : 'Error'}
```

matildepark:
Big step up! This is great. 

Screenshot 2020-07-28 at 3 15 27 PM

matildepark:
OK, one last thing. Can you edit your commits to [match our hygiene guidelines](https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md#hygiene)?

just `git merge-base master error-boundary`, get that hash and

`git rebase -i <hash>`

Merge commits that correspond to one thing and reword the others, once you've done that `git push -f`.

PR #3201, opened 2020-07-28
vere: change lmdb mapsize for aarch64

botter-nidnul:
Adds aarch64 lines to `pkg/urbit/configure` and defines a new macro, `U3_CPU_aarch64`.

That macro is used in `pkg/urbit/vere/disk.c` to change the lmdb mapsize from 1TB to 60GB for aarch64 on linux, which was the previous (working) behavior #3014 

This fixes a problem with the mapsize on aarch64, tangentially discussed in #3173 

Uses #if defined instead of #ifdef in anticipation of macos on aarch64 being a thing before the end of the year.

@joemfb


botter-nidnul:
well, that would explain why I needed to add aarch64 lines to pkg/urbit/configure to get `make release` to work but not to get `make build` to work...🤦

pkg/urbit/configure had a broken cpu architecture detection script, and if $HOST wasn't set it marked the CPU type as "unknown". fixed now.

PR #3204, opened 2020-07-28
chat-hook: identify DMs correctly

liam-fitzgerald:
The previous fix assumed that unmanaged chats are prepending with sig, which is no longer the case. Unfortunately, this means we can no longer reliably disambiguate unmanaged chats from managed ones. Instead, we check for `dm--` prefix. This has the potential risk of returning false negatives.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -20,8 +20,10 @@
       state-3
       state-4
       state-5
+      state-6
   ==
 ::
++$  state-6  [%6 state-base]
 +$  state-5  [%5 state-base]
 +$  state-4  [%4 state-base]
 +$  state-3  [%3 state-base]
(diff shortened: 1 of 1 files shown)


matildepark:
Future DMs have no means, but before groups refactor you literally couldn’t make unmanaged villages with multiple people. I think it therefore captures all remaining cases it targets.

—
~haddef-sigwen
https://urbit.org

Fang-:
Isn't the identification for unmanaged chats (which includes/is DMs) now technically "chats whose group is [hidden](https://github.com/urbit/urbit/blob/4d288f2e4e38ed0a483345d77dee34ebde8a1364/pkg/arvo/sur/group.hoon#L43)"?

liam-fitzgerald:
> Isn't the identification for unmanaged chats (which includes/is DMs) now technically "chats whose group is [hidden](https://github.com/urbit/urbit/blob/4d288f2e4e38ed0a483345d77dee34ebde8a1364/pkg/arvo/sur/group.hoon#L43)"?

Yes, but in this particular case, we've lost the group/metadata for the migrated DM, and so we can't use this.

PR #3205, opened 2020-07-28
gen: add validate-ship generator

liam-fitzgerald:
+validate-ship scries for the groups the ship belongs to and then
validates that the associated syncs are setup in their hooks.

In future, this should be expanded to validate all of our foreign-key type relations, their associated syncs and path structure where relevant.

### DIFF
--- a/pkg/arvo/gen/validate-ship.hoon
+++ b/pkg/arvo/gen/validate-ship.hoon
@@ -1,0 +1,108 @@
+::  Validate userspace state
+::
+::  /hoon/validate-ship/gen
+::
+/+  resource
+:-  %say
+|=  $:  [now=@da eny=@uvJ =beak]
+        [~ ~]
+    ==
+:-  %noun
+=|  issues=(list tape)
+|^
+=.  issues
+  check-group-syncs
+issues
+::
+++  scry-synced
+  |=  app=term
+  ^-  (set path)
+  %~  key  by
+  .^  (map path ship)
+    %gx
+    (scot %p p.beak)
+    app
+    (scot %da now)
+    /synced/noun
+  ==
+::
+++  scry-grp-push-syncs
+  ^-  (set resource)
+  .^  (set resource) 
+    %gx
+    (scot %p p.beak)
+    %group-push-hook
+    (scot %da now)
+    /synced/noun
+  ==
+::
+++  scry-grp-pull-syncs
+  ^-  (set resource)
+  %~  key  by
+  .^  (map resource ship)
+    %gx
+    (scot %p p.beak)
+    %group-pull-hook
+    (scot %da now)
+    /synced/noun
+  ==
+::
+++  report-missing-syncs
+  |*  [app=term syncs=(set)]
+  ^+  issues
+  %+  weld
+    issues
+  %+  turn
+    ~(tap in syncs)
+  |=(thing=* "{<app>}-hook missing: {<thing>}")
+::  +check-group-syncs: validate sync for group
+::
+::    ensure that syncs are correctly set up for each
+::    hook that has group associated resources
+++  check-group-syncs
+  ^+  issues
+  =/  groups=(set resource)
+    %-  sy
+    %+  turn
+      %~  tap  in
+      %~  key  by
+      =<  dir
+      .^  arch
+        %gy
+        (scot %p p.beak)
+        %group-store
+        (scot %da now)
+        /groups
+      ==
+    (corl de-path:resource stab)
+  =/  [ours=(list resource) theirs=(list resource)]
+     (skid ~(tap by groups) |=(rid=resource =(p.beak entity.rid)))
+  =/  pull-syncs=(set resource)
+    scry-grp-pull-syncs
+  =/  push-syncs=(set resource)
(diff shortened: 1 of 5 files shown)


pkg/arvo/gen/validate-ship.hoon
```diff
@@ -0,0 +1,108 @@
...
+  ==
+::
+++  report-missing-syncs
+  |*  [app=term syncs=(set)]
+  ^+  issues
+  %+  weld
+    issues
+  %+  turn
+    ~(tap in syncs)
+  |=(thing=* "{<app>}-hook missing: {<thing>}")
```
Fang-, line 57: Isn't this gonna print `thing` as a plain noun? You probably want this to be a wet gate too.

```suggestion
  |*(thing=* "{<app>}-hook missing: {<thing>}")
```

pkg/arvo/gen/validate-ship.hoon
```diff
@@ -0,0 +1,108 @@
...
+::  +check-group-syncs: validate sync for group
+::
+::    ensure that syncs are correctly set up for each
+::    hook that has group associated resources
+++  check-group-syncs
+  ^+  issues
+  =/  groups=(set resource)
```
tacryt-socryp, line 64: I suppose this isn't the place to discuss this, but I'm very surprised group-store is still a `(map path group` instead of `(map resource group)`. I can't believe I didn't catch it in review... This sort of magic shouldn't be necessary to get a `(set resource)`.

pkg/arvo/gen/validate-ship.hoon
```diff
@@ -0,0 +1,108 @@
...
+        [~ ~]
+    ==
+:-  %noun
+=|  issues=(list tape)
+|^
+=.  issues
+  check-group-syncs
+issues
+::
+++  scry-synced
```
tacryt-socryp, line 17: Metadata and contacts should at some point be moved onto `push-hook` and `pull-hook` (and away from `path` and onto `resource` identifiers) as well... But yeah, not related to this PR

pkg/arvo/gen/validate-ship.hoon
tacryt-socryp: Ideally, `report-missing-syncs` wouldn't be a wet gate and would just take either a `(set path)` or a `(set resource)`, and would normalize all input to one or the other, then we wouldn't need `thing=*` at all.

philipcmonk:
If these are generic sanity tests (ie tests which depend on the dynamic data of your ship, vs unit tests which use mock data), do you think it would be useful to run this on every OTA, and reject the OTA if it fails?  This would be pretty easy to do.  Could put it in an existing or a new `%sane` app and subscribe to clay's `%/app` (or specify a list of relevant apps).  This subscription would be fired on every update in the same event, so if you crash on failure it'll reject the commit.  Maybe the crash needs to be nondeterministic, but that's easy to arrange.

pkg/arvo/gen/validate-ship.hoon
liam-fitzgerald: It's because a %gy scry must produces an `arch` i.e. `[dir=(map @ta ~) fil=(unit @uvI)]` so we encode the resource into a path and then encode it into a cord. Internally, group-store still uses a `(map resource group)`

liam-fitzgerald:
> If these are generic sanity tests (ie tests which depend on the dynamic data of your ship, vs unit tests which use mock data), do you think it would be useful to run this on every OTA, and reject the OTA if it fails? This would be pretty easy to do. Could put it in an existing or a new `%sane` app and subscribe to clay's `%/app` (or specify a list of relevant apps). This subscription would be fired on every update in the same event, so if you crash on failure it'll reject the commit. Maybe the crash needs to be nondeterministic, but that's easy to arrange.

This is a really good idea, and would definitely help both debugging strange state migration, but also making `-L` tests more reliable. Thus addressing part of my groups OTA postmortem. Question: could we add a scry interface to gall to peek into the active subscription state? Thus we could ensure that all the subscriptions are setup as we expect them to be.

philipcmonk:
> Question: could we add a scry interface to gall to peek into the active subscription state? Thus we could ensure that all the subscriptions are setup as we expect them to be.

We could, though if your apps use /lib/dbug, i think you can just scry that app for `/dbug/subscriptions/noun`

pkg/arvo/gen/validate-ship.hoon
Fang-: > a %gy scry must produces an arch

Is this actually enforced? I don't think it is. Not convinced `arch` exactly the right structure for gall `%y` scries, or at the very least it exposes more weirdness around using paths in paths. "encode into path into cord" (and then back again) gives me bad vibes.

liam-fitzgerald: Apparently not, I didn't know that. Will change to a `(set resource)`

matildepark:
(Last known status: Waiting on an eventual rewrite to migrate into an app.)

matildepark:
Superseded by #3416, right @liam-fitzgerald? We can close if so.

liam-fitzgerald:
yep!

PR #3209, opened 2020-07-28
Hid vertical overflow on group join box

tylershuster:
The vertical overflow on this box has always bugged me. I assume you didn't make it an `<input />` because you wanted the scrollability. If 'nowrap' is set, 'overflow-y-hidden' follows naturally...right?


matildepark:
I believe Publish’s join component worked properly here but the others didn’t. I was very selectively precise, haha. But since users only see Groups input, this is solid.

PR #3212, opened 2020-07-29
interface: refactored chat message component into smaller pieces

tacryt-socryp:
Starting a process of tackling hairy parts of our interface code to bring them under control :)


matildepark:
I forgot message.js was so huge.

tylershuster:
Beat me to the punch

PR #3216, opened 2020-07-29
Interface: Fixes #3195 URL message parsing

### LINKED ISSUE #3195: Chat: Linkification of urls adds \n's, ignores backticks and breaks multiline messages
**Describe the bug**
If you write the message `test http://example.com test` in chat it'll convert it to:

>test
>[http://example.com](http://example.com)
>test

and it turns a multiline message like:

```
test http://example.com
test
```
into:

>test
>[http://example.com test](http://example.comtest)

that is, it appends the first word of the second line to the end of the url

additionally, it linkifies urls regardless of whether they're in backticks and so breaks formatting of code blocks

**To Reproduce**
1. Try the above examples in chat

**Expected behavior**
It'd be preferable if:
a) urls were linkified inline rather than adding \n's
b) multiline messages should not produce broken urls or have their formatting altered
c) urls in codeblocks should not be linkified

 - OS: arch linux 64-bit
 - Browser: qutebrowser
 - Base hash: jp5fi

tylershuster:
This works, but someone might want to check my math, so to speak.
There are two problems being fixed here:
One, the code split on spaces, which caused problems if a URL was followed by a newline
Two, the parser didn't check if we were in a codeblock. This is sort of a lame procedural method but it gets the job done

### DIFF
--- a/pkg/interface/src/apps/chat/components/lib/chat-input.js
+++ b/pkg/interface/src/apps/chat/components/lib/chat-input.js
@@ -209,29 +209,60 @@ export class ChatInput extends Component {
       return;
     }
     let message = [];
-    editorMessage.split(' ').map((each) => {
-      if (this.isUrl(each)) {
-        if (message.length > 0) {
-          message = message.join(' ');
-          message = this.getLetterType(message);
-          props.api.chat.message(
-            props.station,
-            `~${window.ship}`,
-            Date.now(),
-            message
-          );
-          message = [];
-        }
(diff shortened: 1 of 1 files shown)


matildepark:
This *mostly* works but it doesn't work for the \` sample-code-text https://google.com \` case, which produces

image

tylershuster:
This works now

matildepark:
Apologies for the wait, will test now ...

tylershuster:
LMK if it works and I'll squash

PR #3217, opened 2020-07-29
jael, eyre: Add a way to reset the web login code

cmarcelo:
Jael now stores a `step` that is combined with the original salt to
produce a new code.  A `%step` card is used to increment that value,
and effectively resetting the keys.  Because the first `step` is zero,
the first code is the same as before.

Eyre was changed to be notified with `%code-changed` so it can forget
old cookies and their associated sessions.

A new generator was added |code, that does both querying and
resetting the code

    |code             :: shows current code, step and help
    |code %reset      :: changes the code

The old +code generator still works correctly.


galenwp:
cool

pkg/arvo/sys/vane/jael.hoon
```diff
@@ -35,10 +35,18 @@
...
++$  state-0
+  $:  %0
+      pki=state-pki                                     ::
+      etn=state-eth-node                                ::  eth connection state
+  ==                                                    ::
++$  state-1
+  $:  %1
       pki=state-pki                                     ::
       etn=state-eth-node                                ::  eth connection state
+      step=@ud
```
philipcmonk, line 49: Organizationally, I would put this next to `lyf` in `own.pki`.

cmarcelo: Done. That had some ripple effect on the changes to state versioning (now `state-pki` is versioned), so might be worth to take another look.

cmarcelo:
The build error is a bit odd, I wonder if zuse is being correctly rebuilt.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1960,6 +1960,11 @@
     ::
       closed-connections
     ==
+  ::
+  ?:  ?=(%code-changed -.task)
+    ~>  %slog.[0 leaf+"eyre: code changed, throwing away old cookies"]
+    =.  authentication-state.server-state.ax  *authentication-state
+    [~ http-server-gate]
```
Fang-, line 1967: To make sure that the relevant `channel-state`, timers, gall subscriptions get cleaned up, you'll want to call `+discard-channel:by-channel` for every channel in every session. (This is the part of logout logic I was thinking about when we DMed earlier.)

cmarcelo: Done inline in the handler for `%code-changed`. Could you take another look?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1960,6 +1960,31 @@
...
+    ~>  %slog.[0 leaf+"eyre: code-changed: throwing away old cookies"]
+    =.  authentication-state.server-state.ax  *authentication-state
+    ::
+    =/  event-args  [[our eny duct now scry-gate] server-state.ax]
+    =*  by-channel  by-channel:(per-server-event event-args)
+    =*  channel-state  channel-state.server-state.ax
+    ::
+    =/  channel-ids=(list @t)  ~(tap in ~(key by session.channel-state))
+    ?:  =(~ channel-ids)
+      [~ http-server-gate]
```
Fang-, line 1974: I don't think it's very important to catch the "do nothing" case early, the overhead from the couple lines below is extremely minimal. Might be cleaner to remove this, so you only have a single exit point in this block of code.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1960,6 +1960,31 @@
...
+    ::
+    =/  len=tape  (scow %ud (lent channel-ids))
+    ~>  %slog.[0 leaf+"eyre: code-changed: closing {len} channels"]
```
Fang-, line 1977: I'd put either this, or the "throwing away old cookies" one above. Both seems like a bit much. You might consider rephrasing as "old cookies and existing sessions" though.

cmarcelo:
@philipcmonk @Fang- 

The CI failure shows an error about what seems to be the lack of a `%step`, but that's defined on zuse. It seems that particular phase of the CI is not picking it up...

For the reference `/lib/hood/helm/hoon::[87 3].[87 34]>` points to:

```
  (emit %pass / %arvo %j %step ~)
```

Here is the partial log of the error, I can't see zuse being re-compiled (but perhaps was supposed to in an earlier stage of CI).

```
pier: replaying events 1-7
arvo: assembly
arvo: assembled
arvo: formal event
arvo: formal event
arvo: metamorphosis
gall: direct morphogenesis
arvo: metamorphosed
arvo: formal event
arvo: formal event
/lib/hood/helm/hoon::[21 1].[258 3]>
/lib/hood/helm/hoon::[22 1].[258 3]>
/lib/hood/helm/hoon::[23 1].[258 3]>
/lib/hood/helm/hoon::[83 3].[87 34]>
/lib/hood/helm/hoon::[84 3].[87 34]>
/lib/hood/helm/hoon::[85 3].[87 34]>
/lib/hood/helm/hoon::[87 3].[87 34]>
nest-fail
road: new
[%error-building /lib/hood/helm/hoon]
[%error-building /app/hood/hoon]
/lib/hood/helm/hoon::[21 1].[258 3]>
/lib/hood/helm/hoon::[22 1].[258 3]>
/lib/hood/helm/hoon::[23 1].[258 3]>
/lib/hood/helm/hoon::[83 3].[87 34]>
/lib/hood/helm/hoon::[84 3].[87 34]>
/lib/hood/helm/hoon::[85 3].[87 34]>
/lib/hood/helm/hoon::[87 3].[87 34]>
nest-fail
road: new
[%error-building /lib/hood/helm/hoon]
[%error-building /app/hood/hoon]
gall: not running %hood yet, got %poke
gall: not running %hood yet, got %watch
pier: (7): play: done
---------------- playback complete ----------------
ames: live on 31519 (localhost only)
gall: not running %hood yet, got %poke
gall: not running %hood yet, got %poke
http: web interface live on http://localhost:8080
pier (14): live
+ check
++ herb /nix/store/gga43fh7lwizj408yl66v1nwvcwfb1z5-fakebus -d 3
ames: larva: drain
ames: metamorphosis
gall: not running %lens yet, got %watch
gall: not running %lens yet, got %poke
```

Fang-:
You probably want to build and commit a `bin/solid.pill` that contains the updated zuse.

`.pill +solid` from dojo, puts the result in `your-pier/.urb/put.pill`. Alternatively, just `./sh/update-solid-pill`.

cmarcelo:
> You probably want to build and commit a `bin/solid.pill` that contains the updated zuse.
> 
> `.pill +solid` from dojo, puts the result in `your-pier/.urb/put.pill`. Alternatively, just `./sh/update-solid-pill`.

That did the trick. Thanks!

philipcmonk:
Sorry for the delay, but I just tested it and it works great.  Thanks!

vvisigoth:
Would it be some sort of security problem to return the _new_ +code when you run |reset? I did this in web term and it locked me. Since I'm hosted, I'm basically hosed unless I run some kubectl magic.

vvisigoth:
cc @Fang- @philipcmonk

Fang-:
Doesn't hosting have a magic "give me the `+code`" button?

Regardless, I'm not a security expert, but considering we might be clearing all sessions because of a compromised `+code`, wouldn't sending the new code out to connected sessions potentially leak it to the attacker?

PR #3223, opened 2020-07-29
interface: refactored chat input

tacryt-socryp:
Have tested the following:
- Avatar support
- /me, code, url, text messages
- S3 Upload
- switching code mode
- Saving written text input when switching chatrooms

Note: It would be very difficult to do this refactor with the PatPAutocomplete in place, so I removed it. Lmk if that's unacceptable. That's the only regression.


matildepark:
@urcades thoughts on regressing this?

@liam-fitzgerald You wrote the ship search — if you see an easy way to factor it in here, let us know ...

Will review, test in a bit!

urcades:
@matildepark + @tacryt-socryp : It's totally fine, not life-or death if it's removed for a short/mid-while.

Given how unergonomic typing out patPs tends to be, I think an interface like this is quite important to retain in the longer term though.

liam-fitzgerald:
Seeing as we only use the autocomplete in the ChatInput now, (it used to be also in the DM menu), we could probably get away with using CodeMirror's builtin [autocomplete](https://codemirror.net/demo/complete.html). It'd be fun to expand this to include groups as well if you tried to autocomplete on `~hastuc-dibtux/` or similar.

urcades:
Yeah, this reminds me that input completion of some sort will need to make
it into the Friendly Command Line at some point, with respect to
autocompleting commands, group/people shortcodes, and the like specified by
Gav and Jimmy. This is crucial, because an omnibox without basic
autocompletion isn't the effortless/powerful/obvious interface we desire it
to be.

—
~fabled-faster
https://urbit.org/

matildepark:
>  not life-or death if it's removed for a short/mid-while

Generally speaking we end up finding any temporary change lasts longer than you'd think ...

> we could probably get away with using CodeMirror's builtin autocomplete

This sounds kinda cool. @tacryt-socryp if you're into this it could be as simple as `Array.from`ing some patps from the channel into an autocomplete custom dictionary prop.

tacryt-socryp:
I feel as though autocomplete should be punted until we move over to the new indigo editor...

urcades:
> I feel as though autocomplete should be punted until we move over to the new indigo editor...

Yeah, I'm very OK with us holding off until indigo is live

pkg/interface/src/apps/chat/components/lib/chat-editor.js
```diff
@@ -0,0 +1,134 @@
...
+      theme: 'tlon' + codeTheme,
+      lineNumbers: false,
+      lineWrapping: true,
+      scrollbarStyle: 'native',
+      cursorHeight: 0.85,
+      placeholder: props.inCodeMode ? 'Code...' : props.placeholder,
+      extraKeys: {
+        'Enter': () => {
+          this.submit();
+        }
```
matildepark, line 111: We have more keys defined here — `Shift-3` to toggle code mode for example (and only if the message was empty). 

Though honestly, I'm unsure if anyone used this shortcut — it was meant to call back to how code used to be prefixed with `#`, and it's not documented for users.

@urcades

pkg/interface/src/apps/chat/components/lib/chat-input.js
```diff
@@ -339,96 +179,54 @@ export class ChatInput extends Component {
...
+          onUnmount={props.onUnmount}
+          message={props.message}
+          placeholder='Message...' />
+        <div className="ml3 mr3"
+             style={{
+               height: '16px',
+               width: '16px',
+               flexBasis: 16,
+               marginTop: 10
+             }}>
```
matildepark, line 208: This is a little off.

Before (livenet right now):

Screen Shot 2020-08-03 at 7 37 58 PM

After (localhost:9000):

Screen Shot 2020-08-03 at 7 37 47 PM

matildepark:
Also might be worth putting a comment at the top of ship-search.js since it will be entirely unused, if we're going to restore it later. If we're not, it can just be deleted.

pkg/interface/src/apps/chat/components/lib/chat-editor.js
urcades: Are you thinking we remove this? It's a strange input expectation, there's some crossed wires here re: how we treat markdown in general in this input, the code entry mechanism, etc.

matildepark: I think it would be fine to just remove it because it's indeed a strange expectation and in general I'd like to make UX less, well, unintuitive. But I was in any case just flagging that it was removed so we could assent to that

urcades: Yeah — I'm in favor of removing

tacryt-socryp: Ok gonna leave this as is

pkg/interface/src/apps/chat/components/lib/chat-input.js
tacryt-socryp: Reverting to ml2 and mr2

tacryt-socryp:
This is good to go as well (includes the chat window changes)!

pkg/interface/src/apps/chat/components/lib/content/text.js
```diff
@@ -3,7 +3,6 @@ import { Link } from 'react-router-dom';
 import ReactMarkdown from 'react-markdown';
 import RemarkDisableTokenizers from 'remark-disable-tokenizers';
 import urbitOb from 'urbit-ob';
-import { Link } from 'react-router-dom';
```
matildepark, line 5: This was preventing a compile?

tacryt-socryp: I fixed the missing Link import in my branch as well, so the two imports of the same thing made it impossible to compile.

matildepark:
Screenshot 2020-08-06 at 6 16 23 PM

😕

tacryt-socryp:
@urcades The 'CodeEval' asset and the ImageUpload asset are not the same size. It appears that the ImageUpload asset contains the whole button and border in the asset, whereas the CodeEval asset is having its border drawn manually with CSS. This makes it difficult to make these two assets the same size. Could I get some support on this, perhaps a new CodeEval asset with a border around it?

matildepark:
@tacryt-socryp That was actually done on purpose because of issues with light and dark mode (plus the code eval toggle lights up, and needs to light up in both modes, not just invert).

Chat-input on master has these level, what threw them off? 

image

tacryt-socryp:
In safari, current master's icons do not look the same size to me:
Screen Shot 2020-08-06 at 3 32 35 PM

tacryt-socryp:
@matildepark does this look the same? I think it has the same styles as master now

PR #3228, opened 2020-07-30
chat-hook: fix synced map getting out of sync with outgoing subscriptions

tacryt-socryp:
Somehow in the big groups OTA, it became possible for the outgoing subscription map to get out of sync with the "synced" map. In normal operation, this is not allowed. This occurred via unsafe modification of the synced map. This is the root cause of the DM issue. I have made a fix that makes the outgoing subscriptions properly reflect the synced map.

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -21,8 +21,10 @@
       state-4
       state-5
       state-6
+      state-7
   ==
 ::
++$  state-7  [%7 state-base]
 +$  state-6  [%6 state-base]
 +$  state-5  [%5 state-base]
 +$  state-4  [%4 state-base]
@@ -52,7 +54,7 @@
   $%  [%chat-update update:store]
   ==
 --
-=|  state-6
+=|  state-7
 =*  state  -
 ::
 %-  agent:dbug
@@ -81,8 +83,14 @@
     =/  old  !<(versioned-state old-vase)
     =|  cards=(list card)
     |-
-    ?:  ?=(%6 -.old)
+    ?:  ?=(%7 -.old)
       [cards this(state old)]
+    ?:  ?=(%6 -.old)
+      =.  cards
+        %+  weld  cards
+        ^-  (list card)
+        [%pass /s %agent [our.bol %chat-hook] %poke %noun !>(%fix-out-of-sync)]~
+      $(-.old %7)
     ?:  ?=(?(%3 %4 %5) -.old)
       =.  cards
         %+  weld  cards
@@ -327,7 +335,7 @@
       ?+  mark  (on-poke:def mark vase)
           %json         (poke-json:cc !<(json vase))
           %chat-action  (poke-chat-action:cc !<(action:store vase))
-          %noun         (poke-fix-dms:cc %fix-dms)
+          %noun         (poke-noun:cc !<(?(%fix-dm %fix-out-of-sync) vase))
       ::
           %chat-hook-action
         (poke-chat-hook-action:cc !<(action:hook vase))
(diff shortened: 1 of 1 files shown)


ohAitch:
~single sources of truth~

matildepark:
@tacryt-socryp This is marked as draft, what's left here?

If @Fang- says the phrase "Seems sane", do you think we could bundle this today @philipcmonk? I know you're doing in-depth tests on a `/sys` OTA. This is one of those fixes that is better sooner than later.

philipcmonk:
The OTA is compiling on ~zod right now, so better to do it as a separate release.

matildepark:
Noted!

—
~haddef-sigwen
https://urbit.org

tacryt-socryp:
Mark tested it and it worked

pkg/arvo/app/chat-hook.hoon
```diff
@@ -327,7 +335,7 @@
       ?+  mark  (on-poke:def mark vase)
           %json         (poke-json:cc !<(json vase))
           %chat-action  (poke-chat-action:cc !<(action:store vase))
-          %noun         (poke-fix-dms:cc %fix-dms)
+          %noun         (poke-noun:cc !<(?(%fix-dm %fix-out-of-sync) vase))
```
Fang-, line 338: > `%fix-dms` -> `%fix-dm`

So, the previous fix didn't run in the first place? (^:

Kinda glad tbh, I think this new one fixes most cases and is less invasive.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -389,51 +397,80 @@
...
-          %.y
-          %.n
-      ==
-    ::
-      =-  [%pass /fixdm %agent [our.bol %chat-store] %poke %chat-action -]
-      !>  ^-  action:store
-      [%messages new-dm envelopes.u.mailbox]
+  ?-  a
+      %fix-dm    (fix-dm %fix-dm)
+      %fix-out-of-sync  (fix-out-of-sync %fix-out-of-sync)
```
Fang-, line 407: You could remove the `|=` from the top of these arms and just write `fix-dm` etc here, without calling them with constants.

pkg/arvo/app/chat-hook.hoon
```diff
@@ -81,8 +83,14 @@
...
     =|  cards=(list card)
     |-
-    ?:  ?=(%6 -.old)
+    ?:  ?=(%7 -.old)
       [cards this(state old)]
+    ?:  ?=(%6 -.old)
+      =.  cards
+        %+  weld  cards
+        ^-  (list card)
+        [%pass /s %agent [our.bol %chat-hook] %poke %noun !>(%fix-out-of-sync)]~
```
Fang-, line 92: This continues to weird me out. Decoupling into a separate event like this means that the OTA succeeds even if this horribly crashes or w/e.

pkg/arvo/app/chat-hook.hoon
tacryt-socryp: The old one did in fact run, as the arm expected %fix-dms.

matildepark:
Cool. I will probably ship this tn then.

pkg/arvo/app/chat-hook.hoon
matildepark: Is this a "don't crash if we're OTAing during a next-sys OTA" thing?

Fang-: It would only crash if you scried, right? This doesn't scry.

joemfb: A poke to self happens in the same event. If this set a timer, it would affect OTAs as you describe.

FWIW, I'm also not a fan of the poke-to-self pattern in general)

pkg/arvo/app/chat-hook.hoon
Fang-: Yeah, but we were sending a `%fix-dm` instead:

https://github.com/urbit/urbit/blob/5e65abcb7afc3e6c60e67b1f3ea817f606e92a5e/pkg/arvo/app/chat-hook.hoon#L90

tacryt-socryp: Yes, but we ignored the contents of the noun.

Issue #2983, opened 2020-06-05
We should probably do some kind of smart scrolling


galenwp:
Try this:

- Open a chat
- Scroll through the backlog until you've got a few hundred messages
- Inspect an element

It looks like we're actually rendering every message that just got put on the page. If a new message comes in, or we trigger a re-render somehow, we're going to re-render everything — which is slow.

I've noticed that if I leave urbit-help (or any other really active chat) open for too long (and it fills with new messages) my CPU will struggle.

I think the right thing to do is to switch to a react scrolling technique where we only render what's 'on screen'. Something like this: https://bvaughn.github.io/react-virtualized/#/components/List perhaps?

matildepark:
I really want to focus on this this week because Urbit Community is c h o n k y

tylershuster:
I think this is the right approach: https://github.com/bvaughn/react-virtualized/blob/master/docs/creatingAnInfiniteLoadingList.md

It's going to require a bit of surgery though. Just bookmarking this til later

matildepark:
Hey so @tacryt-socryp does the new chat implementation dynamically load and unload pages? If so this issue is probably met.

tacryt-socryp:
It does not, no.

matildepark:
Oh, I see. If you scroll to the bottom it seemingly now only shows the latest page; otherwise, if you scroll up, it just keeps loading elements. So not smart scrolling, but some sort of unload if you're just keeping up with conversation.

tylershuster:
https://github.com/petyosi/react-virtuoso

This might be better than react-window. Explicitly says it doesn't need to know the size ahead of time

matildepark:
Chat implementation drafted in #3357.

matildepark:
Fixed in #3357.

PR #3229, opened 2020-07-30
Interface: Add smart scrolling to group member list

tylershuster:
This uses `react-window`, which is a minimal rewrite of `react-virtualized` mentioned by @galenwp  in #2983.
Basically it combines the "members" and "contacts" into an array, and given an index, either returns members first or contacts if it's shown all the members.

It's not pretty but...it works?

### DIFF
--- a/pkg/interface/src/apps/groups/components/lib/contact-sidebar.tsx
+++ b/pkg/interface/src/apps/groups/components/lib/contact-sidebar.tsx
@@ -1,5 +1,7 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
+import { FixedSizeList as List } from 'react-window';
+
 import { ContactItem } from './contact-item';
 import { ShareSheet } from './share-sheet';
 import { Sigil } from '../../../../lib/sigil';
@@ -23,6 +25,7 @@ interface ContactSidebarProps {
 }
 interface ContactSidebarState {
 	awaiting: boolean;
+  memberboxHeight: number;
 }
 
 
@@ -31,9 +34,20 @@ export class ContactSidebar extends Component<ContactSidebarProps, ContactSideba
   constructor(props) {
     super(props);
     this.state = {
-      awaiting: false
+      awaiting: false,
+      memberboxHeight: 0
     };
+    this.memberbox = this.memberbox.bind(this);
   }
+
+  memberbox(element) {
+    if (element) {
+      this.setState({
+        memberboxHeight: element.getBoundingClientRect().height
+      })
+    }
+  }
+
   render() {
     const { props } = this;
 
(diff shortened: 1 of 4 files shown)


pkg/interface/src/apps/groups/components/lib/contact-sidebar.tsx
```diff
@@ -166,8 +169,20 @@ export class ContactSidebar extends Component<ContactSidebarProps, ContactSideba
           >Channels</Link>
           {shareSheet}
           <h2 className="f9 pt4 pr4 pb2 pl4 gray2 c-default">Members</h2>
-          {contactItems}
-          {groupItems}
+          {this.memberbox.current ? <List
+            height={this.memberbox.current.getBoundingClientRect().height}
+            className="flex-auto"
+            itemCount={contactItems.length + groupItems.length}
+            itemSize={48}
```
tylershuster, line 176: `itemSize={48}` is based on a padding of 6 around a sigil of 32 which is not 48 damn it

pkg/interface/src/apps/groups/components/lib/contact-sidebar.tsx
```diff
@@ -166,8 +169,20 @@ export class ContactSidebar extends Component<ContactSidebarProps, ContactSideba
           >Channels</Link>
           {shareSheet}
           <h2 className="f9 pt4 pr4 pb2 pl4 gray2 c-default">Members</h2>
-          {contactItems}
-          {groupItems}
+          {this.memberbox.current ? <List
```
tylershuster, line 172: We see if the parent has been mounted yet first so we can set the height to that

matildepark:
On ~bitbet-bolbel I get ... nothing?

image

matildepark:
Also, man, this should probably also be in `GroupView` because the Group Settings stalls for about ten seconds on UC.

tylershuster:
> On ~bitbet-bolbel I get ... nothing?

*On* ~bitbet-bolbel? Like...as? This is UC, right? What steps did you take to get this failure?

matildepark:
> _On_ ~bitbet-bolbel? Like...as?

Yes.

> What steps did you take to get this failure?

1. Sign in.
2. Go to Groups.
3. Click Urbit Community. Blank.
4. Click "Add to Group" or refresh the page — members show up.

tylershuster:
Weird, but confirmed.

tylershuster:
Ok check it. For the `GroupView`, this may be something to run past @urcades because I don't know that the current approach (setting it to 500px):

pic

is exactly A  E  S  T  H  E  T  I  C  .

Basically it's a "window" into a list. It's not exactly...if something comes into screen, load it. Which I don't know is possible. LMK if you know of a library, but this seems to be how they're all made.

matildepark:
Wanna squash these commits?

PR #3231, opened 2020-07-30
interface: add new omnibox navigation and header bar

matildepark:
Screenshot 2020-07-30 at 7 23 03 PM

- Since we reference "is this a Tlon app?" in multiple places, "default apps" is refactored into a constant in a library.
- Removes group filter component from entire interface, store, api, etc. If you had a selectedGroup in `localStorage`, it doesn't matter — we also remove all the group filtering from the sidebars.
- Adds an omnibox of all apps (from launch's "basic tile" data); subscribed channels (from metadata-store) and some basic commands (hard-included as per the suite) and populate a group index separately from other subscriptions
- Omnibox is responsive for mobile users; Omnibox also works in dark mode.
- Omnibox responds to keyboard navigation; catches are also added for CodeMirror in Chat so you can still open it from Chat.
- New header bar includes a simplified "Groups" button — it gets a nice badge when you have a group invite.


matildepark:
If you aren't on alpha hosting, this is totally usable with your live ship btw, using `npm run start`. Please do flag issues with your actual data and we'll get it sorted.

liam-fitzgerald:
very excited to have this in. Testing now.

pkg/interface/src/App.js
```diff
@@ -70,6 +73,10 @@ class App extends React.Component {
     this.api.local.setDark(this.themeWatcher.matches);
     this.themeWatcher.addListener(this.updateTheme);
     this.api.local.getBaseHash();
+    Mousetrap.bind(['command+l', 'ctrl+l'], (e) => {
+      e.preventDefault();
+      this.api.local.setOmnibox();
```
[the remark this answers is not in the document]
liam-fitzgerald: ctrl-L works on safari, but cmd-L does not. I think safari may intercept the cmd-L before we see it, so this might have to be changed
cc: @jyng @g-a-v-i-n

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,241 @@
...
+
+  componentDidUpdate(prevProps) {
+    if (prevProps !== this.props) {
+      this.setState({ index: index(this.props.associations, this.props.apps.tiles) });
+    }
+
+    if (prevProps && this.props.show && prevProps.show !== this.props.show) {
+      Mousetrap.bind('escape', () => this.props.api.local.setOmnibox());
+      document.addEventListener('mousedown', this.handleClickOutside);
+      this.input.focus();
```
liam-fitzgerald, line 35: focus on show doesn't work on safari

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,241 @@
...
+        this.props.api.local.setOmnibox();
+      }
+    };
+
+    if (evt.key === 'ArrowDown') {
+      this.setNextSelected();
+    }
+
+    if (evt.key === 'ArrowUp') {
+      this.setPreviousSelected();
```
liam-fitzgerald, line 60: you probably want `evt.preventDefault()` on the arrow keys to prevent it interfering with the text input.

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,241 @@
...
+        })
+      );
+    });
+
+    this.setState({ results: results });
+  }
+
+    setPreviousSelected() {
+    const current = this.state.selected;
+    const flattenedResults = Array.from(this.state.results.values()).flat();
```
liam-fitzgerald, line 131: wrapping doesn't work

pkg/interface/src/components/OmniboxResult.js
```diff
@@ -0,0 +1,80 @@
...
+  render() {
+    const { icon, text, subtext, link, navigate, selected, dark } = this.props;
+
+    let invertGraphic = {};
+
+    if (icon.toLowerCase() !== 'dojo') {
+      invertGraphic = (!dark && this.state.hovered) ||
+      selected === link ||
+      (dark && !(this.state.hovered || selected === link))
+        ? { filter: "invert(1)" }
```
liam-fitzgerald, line 27: we do this a lot, @g-a-v-i-n can we put this in indigo-react sometime.

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,241 @@
...
+  search(event) {
+    const { state } = this;
+    const query = event.target.value;
+    const results = this.initialResults();
+
+    this.setState({ query: query });
+
+    // wipe results if backspacing
+    if (query.length === 0) {
+      this.setState({ results: results });
```
liam-fitzgerald, line 104: this should also reset `state.selected`, otherwise you have jank where it selects the most recent selection, after you've cleared it.

pkg/interface/src/components/Omnibox.js
matildepark: I mean does arrow up do anything on text input?

pkg/interface/src/components/OmniboxResult.js
matildepark: This stuff is just because Dojo breaks the pattern. What we want is to put the app icons into `<Icon>` and then dispose of all the image stuff and invert patterns, because it handles the fill for us.

pkg/interface/src/components/Omnibox.js
liam-fitzgerald: yeah it moves the cursor to the start for arrow up and to the end for arrow down.

pkg/interface/src/components/OmniboxResult.js
liam-fitzgerald: oh sweet.

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,241 @@
...
+        this.navigate(this.state.selected);
+      } else {
+        this.navigate(Array.from(this.state.results.values()).flat()[0].link);
+      }
+    }
+  }
+
+  handleClickOutside(evt) {
+    if (this.props.show && !this.omniBox.contains(evt.target)) {
+      this.props.api.local.setOmnibox();
```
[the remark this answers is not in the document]
matildepark: Can I disagree? 
Jul-30-2020 20-52-57

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,241 @@
...
+          position='absolute'
+          top='0'
+          right='0'
+          zIndex='9'
+          display={props.show ? 'block' : 'none'}
+        >
+          <Row justifyContent='center'>
+            <Box
+              mt='20vh'
+              width='50vw'
```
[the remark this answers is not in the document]
matildepark: Good call!

pkg/interface/src/App.js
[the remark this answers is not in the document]
matildepark: Blech. https://stackoverflow.com/questions/37371508/disabling-keyboard-shortcuts-using-javascript-not-working-in-safari

pkg/interface/src/components/Omnibox.js
matildepark: Fun fact: Safari needs a `touchstart` event first. Well. OK. Apple just wants me to add another line here.

pkg/interface/src/components/Omnibox.js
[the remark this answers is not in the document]
liam-fitzgerald: I'm thinking about it from the mobile perspective, where you can't easily clear an input and you don't have access to the escape shortcut.

tacryt-socryp:
Will review and test tomorrow morning, excited to check this out.

vvisigoth:
hell yeah
—
~poldec-tonteg
http://urbit.org

matildepark:
pong @tacryt-socryp 

I might also do a tweak to change the group asset, it's abnormally small in the list.

pkg/interface/src/App.js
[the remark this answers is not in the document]
matildepark: Ping @g-a-v-i-n and @jyng once more — Cmd+L is intercepted on Safari but Ctrl+L isn't. Do you have additional suggestions for what we can use or are you content with falling back to Ctrl?

tacryt-socryp:
Also: Why does the Leap button have a keyboard shortcut in the button? Is the keyboard shortcut going to be hardcoded? I have Cmd+L bound to open up Launchpad on MacOS, but I still have to see this non-working keyboard shortcut? Do we have to make this the keyboard shortcut? Also, what will non-mac users do? Surely they won't be able to use the command key. @urcades

tacryt-socryp:
Note the white line as bottom border on the blank results page on Safari

Screen Shot 2020-08-05 at 11 18 47 AM

tacryt-socryp:
Note the horizontal scrollbar on Safari... not good:
Screen Shot 2020-08-05 at 11 19 54 AM

tacryt-socryp:
Empty icon for links that is misaligned with the header
Screen Shot 2020-08-05 at 11 22 03 AM

tacryt-socryp:
no icon for groups
Screen Shot 2020-08-05 at 11 22 37 AM

tacryt-socryp:
navigated to anthony's publish digest and got this error
Screen Shot 2020-08-05 at 11 23 23 AM

pkg/interface/src/App.js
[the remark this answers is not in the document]
tacryt-socryp: I think Ctrl+L is better than Cmd+L, but I also don't think we should show the shortcut on the button. Except maybe in an onboarding flow??

pkg/interface/src/App.js
```diff
@@ -4,6 +4,8 @@ import * as React from 'react';
 import { BrowserRouter as Router, Route, withRouter, Switch } from 'react-router-dom';
 import styled, { ThemeProvider, createGlobalStyle } from 'styled-components';
```
tacryt-socryp, line 6: I think we should break this file into an outer wrapper file that handles all the logic, and then have the Router with routes in a separate file (with the router being a child component of the wrapper file). Starting to get big enough that that makes sense.

pkg/interface/src/components/Omnibox.js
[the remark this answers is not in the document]
tacryt-socryp: I prefer clearing it as well.

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,254 @@
...
+            (result.host !== null ? result.host.includes(query) : false));
+        })
+      );
+    });
+
+    this.setState({ results: results });
+  }
+
+    setPreviousSelected() {
+    const current = this.state.selected;
```
tacryt-socryp, line 136: indent

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,254 @@
...
+              mt='20vh'
+              width='max(50vw, 300px)'
+              maxWidth='600px'
+              borderRadius='2'
+              backgroundColor='white'
+              ref={(el) => {
+                this.omniBox = el;
+              }}
+            >
+              <input
```
tacryt-socryp, line 234: I think this input should be made into its own component to better separate concerns of this file.

pkg/interface/src/components/StatusBar.js
```diff
@@ -1,82 +1,117 @@
...
+            ↩
+          </Text>
+          <Text ml={2} color='black'>
+            Leap
+          </Text>
+          <Text display={mobile ? 'none' : 'inline-block'} ml={4} color='gray'>
+            {metaKey}L
+          </Text>
+        </Box>
+        {connection === 'disconnected' && (
```
tacryt-socryp, line 81: Connection logic should be its own component

urcades:
> Also: Why does the Leap button have a keyboard shortcut in the button? Is the keyboard shortcut going to be hardcoded? I have Cmd+L bound to open up Launchpad on MacOS, but I still have to see this non-working keyboard shortcut? Do we have to make this the keyboard shortcut? Also, what will non-mac users do? Surely they won't be able to use the command key. @urcades

I can answer these individually, will cc @g-a-v-i-n + @jyng to provide further answers/context if they see fit, given they shipped the current working version of this interface:

> Why does the Leap button have a keyboard shortcut in the button?

To provide a very, very, very obvious touchpoint for how to use a crucial, possibly core navigational construct in the system. My hope is that we could offer up a systems setting to remove shortcut helpers across the interfaces where we provide them, but for the time being it seems OK to provide these helpers for this interface's early life.

> Is the keyboard shortcut going to be hardcoded?

My hope is that all keyboard commands/shortcuts will be end-user configurable. So no, it shouldn't be.

> I have Cmd+L bound to open up Launchpad on MacOS, but I still have to see this non-working keyboard shortcut?

I empathize, and actually Cmd/Ctrl+L are both used in most modern browsers to access their search/omnibox fields, and this overlap is something we're very aware of — I'm unsure if Gav + Jimmy thought it was a point that needed reinvestigating. Related to the above point, this shortcut should be user-configurable though, and the interface should reflect the configuration a user makes to this interface.

> Do we have to make this the keyboard shortcut?

See above!

> Also, what will non-mac users do? Surely they won't be able to use the command key. 

I believe it was always in the spec to have this shortcut indicator/label reflect CMD if on macOS, and Ctrl if on linux/windows. If this is not the case, it's being implemented incorrectly, or was specified without this crucial detail.

matildepark:
Ohh! I think I know what you're seeing with the icons being poof; we add some assets to pkg/arvo/landscape/img and you're running on your livenet ship, so no asset (I think for Groups and Home).

matildepark:
> navigated to anthony's publish digest and got this error
> Screen Shot 2020-08-05 at 11 23 23 AM

I can't join Arroyo Digest; so I think it was pre his latest breach and thus not migrated to get tags. Also outside scope of PR, would be separate ticket.

matildepark:
> I believe it was always in the spec to have this shortcut indicator/label reflect CMD if on macOS, and Ctrl if on linux/windows. If this is not the case, it's being implemented incorrectly, or was specified without this crucial detail.

It does indeed say "Ctrl+L" if you're not running Mac. iOS and the like don't show any shortcut indicator (no keyboard + iPad flags itself as a Mac).

pkg/interface/src/components/Omnibox.js
matildepark: Are we able to retain the ref if it's a separate component?

pkg/interface/src/App.js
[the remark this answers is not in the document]
g-a-v-i-n: i would consider using a lib to smooth this over https://github.com/jaywcjlove/hotkeys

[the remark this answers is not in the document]
matildepark: Wouldn't Mousetrap do this just as well? https://medium.com/@subwaymatch/comparing-keyboard-shortcut-libraries-in-react-mousetrap-vs-hotkeys-634877b4af9e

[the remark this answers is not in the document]
matildepark: I can see if stoppropagation would help this as well ...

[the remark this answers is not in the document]
g-a-v-i-n: mousetrap may not deal with edgecases a purpose-built library might.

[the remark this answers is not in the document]
matildepark: Tested, hotkeys has the same issue.

matildepark:
Alright, addressed everything but the Cmd+L issue in Safari (other libraries don't fix it); refactoring Router out of App.js (outside PR scope and my first go at that introduced some weird issues).

Either we tell Safari users to Ctrl+L (I can just add a flag for that) or we change it to like Ctrl+K or something (like Discord, Slack, bla bla bla). I like the idea of `Ctrl+/` because it matches urbit.org.

pkg/interface/src/components/Omnibox.js
```diff
@@ -0,0 +1,251 @@
...
+         categoryResult.push(<Box key={i} width='max(50vw, 300px)' maxWidth='600px'>
+        <Rule borderTopWidth="0.5px" color="washedGray" />
+        <Text gray ml={2}>{category.charAt(0).toUpperCase() + category.slice(1)}</Text>
+          {each}
+        </Box>);
+      }
+    });
+
+    return (
+      <>
```
tacryt-socryp, line 216: What is the advantage of wrapping the Box in this?

matildepark: I think that's from when I was working with Gavin; when you do that it's short hand for `<Fragment>` so it doesn't wrap it in a div or the like. But because we have the Box anyway, it's moot, let me double check and push

urcades:
> Alright, addressed everything but the Cmd+L issue in Safari (other libraries don't fix it); refactoring Router out of App.js (outside PR scope and my first go at that introduced some weird issues).
> 
> Either we tell Safari users to Ctrl+L (I can just add a flag for that) or we change it to like Ctrl+K or something (like Discord, Slack, bla bla bla). I like the idea of `Ctrl+/` because it matches urbit.org.

`Ctrl+/` is fine, I think. It was the only other option we explored in our drifts on this function.

matildepark:
OK! Moved to Cmd+L, works in Safari fine. Also discovered a bug where dismissing the Omnibox prevented it from reappearing until you clicked in again. So I globally bound it with a plugin for Mousetrap that lets you open the omnibox in all locations without having to declare it in all textboxes.

matildepark:
Anything left @tacryt-socryp?

Issue #3190, opened 2020-07-26
Crash on startup due to memory issues, not fixed with |pack


deploradude:
**Describe the bug**
Receive this on startup, indicating full use of memory.  Only difference in recent days is importing ~400 entries of legacy internet 'blog.  Loaded once and used |pack as recommended, but then failed again.  Now bails every time.  Cannot run |mass as cannot access dojo.

urbit 0.10.8
boot: home is /home/span/urbit/docteg-mothep
loom: mapped 2048MB
lite: arvo formula 50147a8a
lite: core 590c9d56
lite: final state 590c9d56
loom: mapped 2048MB
boot: protected loom
live: loaded: GB/2.133.983.232
boot: installed 268 jets
---------------- playback starting ----------------
pier: replaying events 8752750-8754405
eyre: no channel to cancel ~[//http-server/0vr.gigd1/6/2]
pier: (8753250): play: done
pier: (8753751): play: done
|«play»
bail: meme
bailing out
pier: serf error: end of file

**To Reproduce**
Steps to reproduce the behaviour:
1. boot urbit ship after doing whatever changed the state.

**Expected behaviour**
Expect ship to boot without incident

**Screenshots**
n/a; see above for output

**System (please supply the following information, if relevant):**
 - OS: [e.g. macOS, linux64, FreeBSD]
 - 0.10.8
 - base hash: jp5fi

**Additional context**
Mirrored legacy internet 'blog to Publish.  Only recent difference.

**Notify maintainers**
n/a

deploradude:
Ubuntu 20

philipcmonk:
It's very strange that it looks like it's running out of memory while replaying events.  If they didn't run out of memory the first time, why would they the second time?  Maybe it's because we send several events at once, and it doesn't have room to store them?  cc @joemfb 

Can you run `./urbit-worker grab PIER`?  That runs |mass without having to actually start up, so hopefully that works.  Also try running `./urbit-worker pack PIER` and send the output of both of those.

deploradude:
grab output:


`loom: mapped 2048MB
boot: protected loom
live: logical boot
boot: installed 268 jets
serf: measuring memory:
  warm jet state: B/288
  cold jet state: B/288
  hank cache: B/288
  battery hash cache: B/288
  hot jet state: KB/124.408
total jet stuff: KB/125.560
  bytecode cache: B/288
total nock stuff: B/288
  memoization cache: B/288
total road stuff: B/288
total marked: KB/126.136
[...]

Issue #3179, opened 2020-07-24
urbit crashes with core file, then unable to restart


mdecerbo:
**Describe the bug**
A long-running urbit 0.10.7 session crashed with a core dump and issued `bail:meme` and `pier: work error: end of file` when I try to restart. I downloaded urbit 0.10.8 and got another core dump with different error messages. 

**To Reproduce**
1. Notice that a core dump appeared at 0153 CET and a long-running urbit process terminated.
2. Attempt to restart with `urbit $SHIPNAME`
3. Get a failure and another core dump.
4. Download urbit 0.10.8 and repeat attempt to restart.
5. Encounter another failure and core dump.

**Expected behaviour**
Expected to successfully enter a new urbit session.

**Screenshots**
Here's the relevant console output (ship name elided for privacy):
```
~
urbit 0.10.7
boot: home is /path/to/$SHIP_NAME
loom: mapped 2048MB
lite: arvo formula 297b51b3
lite: core 12b850ec
lite: final state 12b850ec
loom: mapped 2048MB
boot: protected loom
live: loaded: GB/2.138.275.840
boot: installed 268 jets

work: contiguous free space: MB/9.238.652
work: measuring memory:
work: play 1800718
---------------- playback starting----------------
pier: replaying events 1800718 through 1801816
  kernel: GB/1.213.377.168
total arvo stuff: GB/1.213.377.168
  warm jet state: KB/107.060
  cold jet state: MB/96.425.320
  hank cache: B/288
  battery hash cache: B/288
  hot jet state: KB/125.376
total jet stuff: MB/96.658.332
  bytecode programs: KB/766.224
  bytecode cache: KB/83.296
total nock stuff: KB/849.520
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
total marked: GB/1.310.885.344
free lists: MB/827.358.232
sweep: GB/1.310.885.344

...[snip]...
---------------- playback complete----------------
ames: live on 0
eyre: no channel to cancel ~[//http-server/0v11.ul8b3/1.758/2]
eyre: no channel to cancel ~[//http-server/0v11.ul8b3/18/1]
eyre: canceling ~[//http-server/0v11.ul8b3/1.757/5]
eyre: no channel to cancel ~[//http-server/0v11.ul8b3/157/4]
eyre: no channel to cancel ~[//http-server/0v11.ul8b3/36/1]
eyre: canceling ~[//http-server/0v11.ul8b3/4.296/5]
[...]

Issue #3117, opened 2020-07-10
Pack doesn't work in some cases where I assume it should


forfel-norfel:
I've been running the `ipc-redux-replay-recovery` binary from https://github.com/urbit/urbit/issues/2903#issuecomment-629553431 on macOS to recover, running it in batches of 1m events, and it's worked well so far, until getting to 12720438.

Apparently I care more about fighting bugs than actually getting my planet running, since I opted not to breach back in May, or at any point since then.

I started experimenting with replaying different batches of events. I replayed to 12.7m instead of 13m, then 12.71m. I save the chk folder at each successful manual checkpoint and restore the version that corresponds to the -r flag for each run. I found that certain combinations of checkpoints will work where similar ones will not.

##### `./urbit -r 12710000 -n 12719998 forfel-norfel` succeeds:
```
~
urbit 0.10.4
boot: home is /Volumes/Urbit/urbit-exp/forfel-norfel
loom: mapped 2048MB
lite: arvo formula 28f92f5d
lite: core 682b3a42
lite: final state 682b3a42
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/651.427.840
boot: installed 262 jets
serf: rock: forfel-norfel/.urb/roc/12710000.jam found
boot: protected loom
live: loaded: MB/651.427.840
rock: load: MB/209.506.472
serf (12710000): compacted loom
pier: replay till 12719998
---------------- playback starting ----------------
pier: replaying events 12710001-12719998
...
ames: ~dev: send ack-2 [1 fragment-num=34.827 num-fragments=79.489]
pier: (12719998): play: done
---------------- playback complete ----------------
pier: replay barrier reached, packing
serf (12719998): saving rock
serf: measuring memory:
  kernel: MB/364.252.164
total arvo stuff: MB/364.252.164
  warm jet state: KB/65.204
  cold jet state: MB/97.220.300
  hank cache: B/288
  battery hash cache: B/288
  hot jet state: KB/125.868
total jet stuff: MB/97.411.948
  bytecode programs: KB/564.560
  bytecode cache: KB/59.796
total nock stuff: KB/624.356
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
total marked: MB/462.288.792
free lists: MB/198.375.608
sweep: MB/462.288.792

pier: pack complete, shutting down
```

##### `./urbit -r 12710000 -n 12719999 forfel-norfel` succeeds:
```
[...]

PR #3235, opened 2020-07-31
u3: off-loom global deduplication

joemfb:
The PR adds off-loom global state unification and memory compaction. It works via a hash-consing allocator, backed by malloc. We reallocate and deduplicate the entire persistent state, wipe the loom, and then re-reallocate on the loom.

The new allocator is in a new `ur` module, and has no dependencies on `u3`. It should be an entirely separate library, but is included in the `u3` build for shot-term expedience. As an experiment -- since it's conceptually separate from u3 -- this module is written in a conventional style. (It's an experiment, because it remains to be seen if the cost of context-switching overrides the conventional clarity.) This is effectively a first, rough draft, and all APIs and implementation details are subject to change.

This PR is a basic, minimum viable integration of this new tooling, availably exclusively as an emergency recovery tool in the `urbit-worker` binary:

```
./urbit-worker uniq path/to/pier
```

Reconciling this tooling with the existence of the loom defragmenter (ie, the new `|pack` added in #) and the old `|pack` (ie, current `-n` and `-r` partial replay arguments) remains a top priority. The exact shape of that integration is yet to be determined; the next steps will be to implement jam and cue (universal noun de/serialization) on top of this noun representation.

But even in this minimal state, this tool allows us fully utilize the loom up to its 2GB limit, by maximizing structural sharing and limiting fragmentation. Since it runs off-loom, it can handle arbitrary amounts of data -- up to the limits of the underlying machine.

Relevant history: #1984, #2957, #3099

Obsoletes: #3182

Ameliorates #3190, #3179, #3117, #3116, #3051, #2903, #2896, #2677, #2456, #2283, #2108, #1767, #1629, and likely others. (Full fixes will require further integration, and, in some cases, improvements to related functionality.)

Many thanks to @frodwith, who designed the hash-consing allocator, and guided me in its implementation.


galenwp:
king

joemfb:
To anyone looking to run this before it's released, the latest static binaries from this branch are here:

- [linux](https://bootstrap.urbit.org/993f4b83f4df7a65f76ce03df93871fbf59f659c-linux64.tgz)
- [macos](https://bootstrap.urbit.org/993f4b83f4df7a65f76ce03df93871fbf59f659c-darwin.tgz)

Just download one of those, unpack the tarball (`tar -xzvf 993f4b83f4df7a65f76ce03df93871fbf59f659c-linux64.tgz`), and run `./993f4b83f4df7a65f76ce03df93871fbf59f659c-linux64/urbit-worker uniq path/to/your/pier`.

philipcmonk:
Read this in some detail, and it looks great!  Agree with @Fang- that a couple comments on the data structures could be useful; for me it all clicked into place when I understood the relation between `_cu_loom` and `_ur_root_t`.

joemfb:
My apologies for the delay -- I'm afraid I let the scope of this effort expand significantly. This PR now includes:

- an off-loom, hash-consing noun representation and allocator
- routines to (destructively) move permanent state off the loom and back
- off-loom jam and cue implementations
- backed by off-loom bitstream read/write modules
- and several hashtables
- use of the above for taking/loading portable snapshots
- additional implementations of jam/cue, with various permutations of loom usage
- use of the same on both sides of the ipc pipe, and in the (c) king's persistence layer
- a new +cue jet
- tests and benchmarks for (portions of) the above

As a result of these features, we can fully utilize the loom up to its 2GB limit (by completely deduplicating the persistent state),
and take/load portable snapshots of a size up to (and, in fact, well beyond) the limitations of the loom.

Additionally, the new jam and cue implementations generally exceed the performance of the old by a factor of 2-10x, depending on a variety of details. In the most extreme cases (a full, highly deduplicated, loom), portable snapshot performance exceeds the old implementation by many hundreds of times (due to guaranteed, preemptive unification, as opposed to exponential re-comparison of duplicate, equal subtrees in the worst case).

Incredibly unscientific results follow:

The (incredibly basic) jam/cue microbenchmarks (`make bench`) simply jam or cue a representative %ames effect (in an ipc message wrapper) with different implementations, 20K times in a loop, timing with `gettimeofday()`. Representative output from a run on my machine (2017 macbook pro):

```
jam microbenchmark:
  jam og: 48 ms
  jam xeno: 27 ms
  jam cons: 24 ms
  jam cons unsafe: 11 ms

cue microbenchmark:
  cue og: 270 ms
  cue atom: 63 ms
  cue xeno: 32 ms
  cue xeno unsafe: 26 ms
  cue test: 21 ms
  cue test unsafe: 15 ms
  cue cons: 59 ms
  cue re-cons: 90 ms

cue virtual microbenchmark:
  cue virtual og: 280 ms
  cue virtual atom: 38 ms
```

Here's the result of some simple trace profiling of the changes to the ipc layer. These numbers are taken from local usage of a fake ship,
wherein I mashed on the keyboard in dojo, sent myself a few small messages in webchat dm, and refreshed the page a few times. The jam/cue timing was logged in the `chrome://tracing` format, and averaged in the ui. Note, again, that personal results may vary -- past performance is no indicator of future success:

- serf jam (effects and logging output) improved 15-22x
- serf cue (input events) improved ~3.5x
- king jam (input events) improved ~1.5x
- king jam (input events, persistence) improved ~1.5x
- king cue (input events, replay) improved ~1.8x
- king cue (effects and logging output) improved ~12x
- king process startup (ivory pill cue) improved ~4x.

----------------------------

The new portable snapshots have been wired into the -n/-r replay arguments (and `urbit-worker` commands: `cram` and `queu`). Global deduplication with jamming a portable snapshot is now available via `urbit-worker meld` (since `uniq` is not a verb), and will be available via `|meld` -- the arvo-side implementation will be in a separate PR. (The relevant changes -- fairly small -- will also need to be ported to king-haskell.)

There are two more pieces of integration that should be done.

The new %ames stateless-forwarding implementation in vere should make special use of the new jam/cue implementations (most notably the non-allocating `ur_cue_test()` for packet validation) -- that's simple, and can be PR'd after these changes are merged.

The global deduplicator (meld) should be automatically deployed via memory pressure heuristics. At the same time, various currently-fatal error conditions (`bail:meme`) should become non-fatal. This will require memory pressure notifications to flow back from the serf to the king, and the i/o driver event error-handlers to be integrated with overall memory-pressure tracking and mediation.

As this is a large PR, I've requested multiple reviewers, with the following breakdown in mind:

- @Fang-: changes to the c king (predominantly in `vere/`)
- @philipcmonk: changes to the u3 library (`noun/`) and serf implementation (`worker/`)
- @brendanhay: new off-loom modules (`include/ur`, `ur/`, `tests/ur_tests.c`)

@brendanhay, note that the new modules in the `ur_` namespace are effectively a separate library, with no dependencies except murmur3. I bundled them in `pkg/urbit` for development convenience, but I'd be happy to split them into a separate library now if you have any strong opinions about how that should be built. I expect them to eventually become a separate library in any case.

Fang- (commented):
> I let the scope of this effort expand significantly.

Nobody is surprised. (; Amazing work, regardless!

> routines to (destructively) move permanent state off the loom and back

To be clear, "destructively" here just means temporarily wiping the loom, right? It's not like you're losing any data.

I gave the king-related code a look. The changes mostly make sense to me. The only issue I found is in a printf. I could complain about alignment inconsistencies in variable declaration, but it's really not that bad.

There's a couple new `XX`s that you added. I'm assuming we're fine punting on all of those. You could consider writing some of them up as contributor-friendly issues, maybe someone will pick them up. (Thinking specifically of the "check if dictionary can be shrunk" ones, which seem fairly self-contained. Perhaps there's others.)

galenwp:
Exciting

brendanhay:
I'd leave `ur` as-is for now ( - will `urcrypt` subsume a direct `murmur3` dependency at all? Either way we can look at how to build/split `ur` proper separately from this PR.

frodwith:
I wasn't planning on including murmur in urcrypt, as it's not a
cryptographic hash, but I could be convinced otherwise.

Fang-:
For the record, just rescued ~palfun-foslup from `bail: meme` using this! Good stuff.

 `urbit-worked meld` wasn't blazing-fast by any means, still took 10-15 minutes, but it got the job done. If those are anomalous numbers (for a standard-ish DO droplet), let me know. I have a backup of the pre-meld `chk` files in case they're interesting.

brendanhay (approved):
I've looked through the off-loom modules - other than the shock of being confronted with 'normal' C code in the urbit repository, it looks good to me.

joemfb:
For anyone needing to run this, the latest static binaries from this branch are now:

- [linux](https://bootstrap.urbit.org/a6a7b557a95cd246430ce68809d8188d40ae28f4-linux64.tgz)
- [macos](https://bootstrap.urbit.org/a6a7b557a95cd246430ce68809d8188d40ae28f4-darwin.tgz)

Just download one of those, unpack the tarball (`tar -xzvf a6a7b557a95cd246430ce68809d8188d40ae28f4-linux64.tgz`), and run `./a6a7b557a95cd246430ce68809d8188d40ae28f4-linux64/urbit-worker meld path/to/your/pier`.

philipcmonk (approved):
There's a lot here, but I think it all makes sense.  With occasional `|meld` (which will be automatic in trim eventually), this essentially gets us to the point where all memory leaks are "true" leaks in the kernel/app code (or a few hashtables), which is a great milestone.  I remember when we finally got to zero C memory leaks on the loom (around when we wrote `|mass`), and that made a huge difference.  This is the next level!

I reviewed noun/ and worker/ and they look good to me.

PR #3236, opened 2020-07-31
interface: add settings screen

liam-fitzgerald:
Adds a settings screen which allows for customisable backgrounds, the hiding of avatars and nicknames and the order of tiles on the homescreen. Includes affordances for configuring S3, and logging out. Persists state to localStorage. 

[demo](http://ewr1.vultrobjects.com/urbit/Screen%20Recording%202020-07-31%20at%205.03.35%20pm.mov)

Screen Shot 2020-07-31 at 5 10 03 pm

Screen Shot 2020-07-31 at 5 08 27 pm


pkg/interface/src/reducers/local.ts
```diff
@@ -1,18 +1,36 @@
...
-import { LocalUpdate } from '../types/local-update';
+import { LocalUpdate, BackgroundConfig } from '../types/local-update';
 
-type LocalState = Pick<StoreState, 'sidebarShown' | 'selectedGroups' | 'dark' | 'baseHash'>;
+type LocalState = Pick<StoreState, 'sidebarShown' | 'selectedGroups' | 'baseHash' | 'hideAvatars' | 'hideNicknames' | 'background' | 'dark'>;
 
 export default class LocalReducer<S extends LocalState> {
+    rehydrate(state: S) {
+      try {
+        const json = JSON.parse(localStorage.getItem('localReducer') || '');
```
tacryt-socryp, line 11: Clever and good approach to storing visual settings.

pkg/interface/src/themes/light.ts
```diff
@@ -162,7 +162,7 @@ const theme = {
   ],
   // z-index
   zIndices: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
-  breakpoints: ["768px", "1024px", "1440px", "2200px"],
+  breakpoints: ["34.375em", "46.875em", "60em"]
```
tylershuster, line 165: yes

matildepark:
Dojo needs its background made transparent.

image

matildepark:
Groups, on the other hand, has too much transparency.

image

matildepark:
It would be dope to let the user set "light / dark / auto" theme settings (I find the dark mode looks dope with the background, for eg.) but I think we need to wrap refactoring into indigo-react first because it's not simple to override `prefers-color-scheme`...

matildepark:
image

This is a weird way to do this ... @urcades do we have a "drag and drop" design anywhere?

matildepark:
image
This is a good place for letting the user upload a background via s3 — see the avatar upload in Groups for an example ...

matildepark:
Finally — the clock uses some `transparent` colours that look bizarre with backgrounds.

image

Now that we `matchMedia` you'll want to pass the dark boolean to the clock to set `white` vs. the `gray0` backgrounds, instead of just falling back to transparents...

tylershuster:
Re @matildepark comments on transparency perhaps checking for `@supports(background-filter: blur())` and then making a frosted pane effect and falling back to a semi-transparent one

liam-fitzgerald:
cc: @urcades thoughts?
Screen Shot 2020-08-04 at 12 18 43 pm

urcades:
> image
> 
> This is a weird way to do this ... @urcades do we have a "drag and drop" design anywhere?

We do not, but I'm happy to make one using existing parts

urcades:
> cc: @urcades thoughts?
> Screen Shot 2020-08-04 at 12 18 43 pm

I can see this looking pleasant with a refreshed sidebar look and feel, but we'd need to tweak the translucency/blurring precisely to get visibility/accessibility/readability correct. Even with this example, the green text buttons can get a little washed out and hard to read – I'd only imagine this gets accentuated further with more vibrant background images.

We can revisit this as we assemble a new sidebar structure — transparency should be something we allow users to set parametrically at some point, possibly — Might be good to @g-a-v-i-n here, as I'd imagine we'd want to think of a wholesale way to begin introducing transparency/translucency into indigo, so we aren't doing this piecemeal.

urcades:
So tl;dr, let's keep that sidebar background boring and white100 for the time being — we'll think about how to go about this in a systematic manner.

tylershuster:
I apologize in advance if this comes off as a bit rambling, but allowing the user to customize settings is like feeding the proverbial mouse.

The cookie is allowing the user to change their background. Obviously, and intentionally, this changes the whole feel of the interface. There's no guaranteeing one look going with that background. So keeping everything white is the safe way to go.

Of course, they didn't choose the background because they want it to be hidden, so you give them some translucency settings. But again as Ed points out, the same setting won't work for everyone.

It seems to me that inherent to the capacity to change the interface is a set of *semantically variable* classes, as opposed to the current *utility* classes. E.g. `.bg-translucent` where the particular properties of that class can be set via interface. This too is a cookie because how much control do you give? Do you go full MySpace or do you just give them a slider to control `n` properties of `x` classes?

tl;dr I agree with Ed and thank you for coming to my ted talk

pkg/interface/src/themes/light.ts
g-a-v-i-n: hang on, why did we go with ems here?

g-a-v-i-n:
willing to think about transparent backgrounds a bit more but agree with ed.

tacryt-socryp:
No unnecessary bells and whistles should be a policy from our team as a whole.

pkg/interface/src/themes/light.ts
liam-fitzgerald: to match our current breakpoints in indigo-static

matildepark: Actually @liam-fitzgerald brought up out of band that there is no way to do anything responsively with a minimum of 768 also, this seems like a valid concern.

liam-fitzgerald:
Screen Shot 2020-08-05 at 1 46 43 pm

[drag and drop demo](http://ewr1.vultrobjects.com/urbit/dnd-demo.mov)

pkg/interface/src/themes/light.ts
g-a-v-i-n: Ok, I need to check on this because it seems counter to the [styled-system documentation](https://styled-system.com/responsive-styles), unless I'm misunderstanding these docs. I'd also prefer we use px for breakpoints. You can just multiply ems by 16 to get their px values.

tylershuster: Not exactly. Really it should be rems but at this point there won't be any divergence. I always use rems because that way the interface scales in proportion to whatever the user sets their font size at. But I don't know what kind of guidelines this project follows. I've actually transitioned to setting font-size in millimeters at root and everything else to rem which bases it off that. That grounds it in physical reality.

pkg/interface/src/apps/profile/components/lib/DisplayForm.tsx
```diff
@@ -0,0 +1,230 @@
+import React, { useCallback } from "react";
+
```
tacryt-socryp, line 2: Should be broken into individual files.

pkg/interface/src/apps/profile/components/lib/S3Form.tsx
```diff
@@ -0,0 +1,191 @@
...
+  Text,
+  Menu,
+  MenuButton,
+  MenuList,
+  MenuItem,
+} from "@tlon/indigo-react";
+
+import { Formik, Form } from "formik";
+import * as Yup from "yup";
+import GlobalApi from "../../../../api/global";
```
tacryt-socryp, line 19: Should be broken into individual components

matildepark:
Time to resolve some conflicts! @liam-fitzgerald 

@urcades Do we have a gear icon or something for a settings button? Post-omnibox there's no way to access settings.

urcades:
One exists for export here:
https://www.figma.com/file/H1RAHV4KscSTnvrIiL0z8C/Indigo?node-id=312%3A147

This has come across in a meeting with Liam, but I'm unsure if it was
socialized more broadly: The mechanism of "System Settings" access should
effectively be via the user's profile — the idea is to conjoin the notion
of one's root identity, and one's ship settings into one location. "You"
are effectively "the ship", and so it seems to make sense to have settings
reside as a section within one's "profile".

Additionally, settings should totally be searchable in the omnibox.

—
~fabled-faster
https://urbit.org/

matildepark:
Yes it should! Just remember to stick that in the index in lib/omnibox.js @liam-fitzgerald.

Yes, his initial go at this had the settings linked by clicking your sigil, which is no longer in the header bar anywhere...

—
~haddef-sigwen
https://urbit.org

matildepark:
Like, it could be *exclusively* accessed by leaping to "Profile", but seems unfriendly?

urcades:
Yeah, I totally see this not being legible at first bat. A part of me
thinks that editing the terminology from "Profile" to "Ship" would fix
this, but we'd be reifying one of our weird neologisms into
top-of-interface status.

—
~fabled-faster
https://urbit.org/

liam-fitzgerald:
@urcades that link is dead, do you have another?

tacryt-socryp:
I like the idea of keeping the word Profile but having a settings cog button to the right of it that's always present. Seems fine.

matildepark:
An icon in the text? I vote for sigil icon, “Profile and Settings” copy.

tacryt-socryp:
I meant a separate cog button that directs to settings. "Profile and settings" seems too heavy to me...

matildepark:
Oh you’re talking about the header. I mean as an omnibox entry. The profile and settings are one thing is the issue.

tacryt-socryp:
Is this ready?

matildepark:
Awaiting a UX change and then I’ll do a basic test. I asked @urcades to dump a definitive mock here so we can wrap it.

urcades:
@matildepark  + @liam-fitzgerald Apologies for the wait, needed to cross reference w/ gav + jimmy:

Screen Shot 2020-08-12 at 9 32 39 PM

The above are the two ways we can imagine "profile + settings" being embedded in a single "ship" button. We've gone over this quite a bit but we generally share the belief that most people will expect some configuration interface to exist where they see their ship name.

In addition, it'd be good to surface this same interface here, in the Leap menu:

Screen Shot 2020-08-12 at 11 02 49 PM

I realize the default Leap menu doesn't exist just yet as we shipped MVP, but we should consider the above menubar listing/capability to be surfaced in the default menu as well.

liam-fitzgerald:
Screen Shot 2020-08-17 at 5 42 13 pm
 

ready to merge!

Screen Shot 2020-08-17 at 5 42 25 pm

matildepark:
Can you match [this](https://github.com/urbit/urbit/pull/3334#issuecomment-675075797)? Also conflict. Oof.

matildepark:
I believe post-Chat refactor this was dropped as a prop. 
image

matildepark:
hiding avatars and nicknames also appears to not be included in Links, Publish, full stop. All apps should include the boolean!

matildepark:
This also needs to pull root identity out of Groups and put its 'contact card' editing into settings.

urcades:
Yes — I'm not sure if it has immediate bearings in the way we're
structuring settings, but we should really be conceptually extracting the
idea of the "root profile" completely away from groups — it's no longer a
structure of groups, it is a standalone representation of the ship a user
is operating, it's external cloaking/presentation, etc.

To the various points raised about profile development and externalization
Galen has been bringing up recently, we should really consider the notion
of "one's profile" to be a very core structure that is externalized in
other applications, is the "prototype" object of representation so to speak.


—
~fabled-faster
https://urbit.org/

matildepark:
As of #3337 merge, this is now unblocked.

liam-fitzgerald:
ready to go! @matildepark

matildepark:
You resolved the conflict, but did you address these:

> This is amazing. Except one thing blocking the merge ...
> 
> Screen Shot 2020-08-18 at 1 18 33 PM
> 
> Screen Shot 2020-08-18 at 1 18 41 PM

> I believe post-Chat refactor this was dropped as a prop.
> image

> hiding avatars and nicknames also appears to not be included in Links, Publish, full stop. All apps should include the boolean!

> This also needs to pull root identity out of Groups and put its 'contact card' editing into settings.
> 
> So remove:
> 
> image
> 
> And put it in this sidebar as "Your Identity" above "ship settings":
> 
> image
> 
> Same right-hand panel.

liam-fitzgerald:
whoops, on it!

matildepark:
Missing from profile-overlay

image

matildepark:
People with nicknames set still have Inter applied even if nicknames are disabled.

image

matildepark:
And Your Identity should come first.

image

matildepark:
That seems to be it. Everything else looks slick and these are tiny changes.

liam-fitzgerald:
done!

pkg/arvo/app/landscape/index.html
```diff
@@ -24,6 +24,5 @@
     <script src="/~landscape/js/channel.js"></script>
     <script src="/~landscape/js/session.js"></script>
     <script src="/~landscape/js/bundle/index.f58fbbc4b037bb976a2a.js"></script>
-    <script src="https://sdk.amazonaws.com/js/aws-sdk-2.1.12.min.js"></script>
```
matildepark, line 27: I have no idea how or why this got dropped, but I restored it in next-userspace.

PR #3241, opened 2020-07-31
Remove lodash dependency

tylershuster:
Tell me if this isn't welcome, but

I think it's twenty twenty and transpiled (Java/Type)Script can do a lot more than when lodash was invented. It's not like landscape leans heavily on it. There are a *few* places where the vanilla replacement is more awkward but I think in general native JavaScript is more expressive. It's a little bit lighter.

### DIFF
--- a/pkg/interface/src/apps/chat/components/lib/chat-input.js
+++ b/pkg/interface/src/apps/chat/components/lib/chat-input.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import _ from 'lodash';
 import moment from 'moment';
 import { UnControlled as CodeEditor } from 'react-codemirror2';
 import CodeMirror from 'codemirror';
@@ -339,12 +338,10 @@ export class ChatInput extends Component {
         classes={sigilClass}
         />;
 
-    const candidates = _.chain(this.props.envelopes)
-      .defaultTo([])
-      .map('author')
-      .uniq()
-      .reverse()
-      .value();
+    const candidates = this.props.envelopes
+      .map(envelope => envelope.author)
+      .filter((value, index, self) => self.indexOf(value) === index)
+      .reverse();
 
     const codeTheme = state.code ? ' code' : '';
 
--- a/pkg/interface/src/apps/chat/components/lib/ship-search.js
+++ b/pkg/interface/src/apps/chat/components/lib/ship-search.js
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import _ from 'lodash';
 import urbitOb from 'urbit-ob';
 import Mousetrap from 'mousetrap';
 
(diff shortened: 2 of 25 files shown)


matildepark:
@tacryt-socryp @liam-fitzgerald How do you feel about removing the lodash dependency? I believe you both made use of it; I'm pretty neutral on this.

liam-fitzgerald:
:( 

in defense of lodash:

> I think it's twenty twenty and transpiled (Java/Type)Script can do a lot more than when lodash was invented.

This is unequivocally true. However, there are still huge holes in the JS stdlib. Map/reduce/filter over objects is still awkward, requiring `Object.keys` or `Object.entries`. Objects as a generic key value datastructure are pretty fundamental and I'd hate to lose these tools. `lodash/fp` is also neat for point free style. Furthermore, lodash's typings are *really* good and prevent you from having to reason about TS generics and avoid casts. e.g. lodash's `_.pick`, when correctly typed looks roughly like this. 
```typescript
type Pick<O, K extends keyof O> = { [X in keyof O]: X extends K ? O[X] : never };

function pick<O, K extends keyof O>(obj: O, keys: O[]): Pick<O,K>
```
These type signatures are awkward to read and write and should be avoided if we can.

Lodash is not particularly heavy, and so removing it feels kind of unnecessary. 

RFC though, interested to hear others opinions.

liam-fitzgerald:
Screen Shot 2020-08-04 at 11 36 18 am

A visualisation of our deps. FWIW, they all gzip to 337KB so i don't think we have a particularly onerous dependency issue right now.

pkg/interface/src/apps/chat/components/lib/chat-input.js
```diff
@@ -339,12 +338,10 @@ export class ChatInput extends Component {
...
 
-    const candidates = _.chain(this.props.envelopes)
-      .defaultTo([])
-      .map('author')
-      .uniq()
-      .reverse()
-      .value();
+    const candidates = this.props.envelopes
+      .map(envelope => envelope.author)
+      .filter((value, index, self) => self.indexOf(value) === index)
```
liam-fitzgerald, line 343: this is more verbose && probably less efficient than _.uniq

liam-fitzgerald: also this doesn't handle the envelopes is undefined case

pkg/interface/src/apps/chat/components/lib/ship-search.js
```diff
@@ -119,35 +118,30 @@ export class ShipSearch extends Component {
 
       return (
         hay.startsWith(needle) ||
-        _.some(_.words(hay), s => s.startsWith(needle))
+        hay.split(/\s/).some(s => s.startsWith(needle))
```
liam-fitzgerald, line 121: this is not technically equivalent, but it's close enough. Leaves in punctuation and numbers.

pkg/interface/src/apps/publish/app.js
```diff
@@ -46,19 +45,17 @@ export default class PublishApp extends React.Component {
 
     const notebooks = props.notebooks ? props.notebooks : {};
 
-    const unreadTotal = _.chain(notebooks)
-      .values()
-      .map(_.values)
-      .flatten() // flatten into array of notebooks
+    const unreadTotal = Object.values(notebooks)
+      .map((owner) => Object.values(owner))
```
liam-fitzgerald, line 49: Another bug bear: point free style is pretty difficult with JS stdlib.

pkg/interface/src/components/InviteSearch.tsx
```diff
@@ -177,7 +178,7 @@ export class InviteSearch extends Component<
       const shipIdx = shipMatches.findIndex((ship) => ship === selected);
       const staleSelection = groupIdx < 0 && shipIdx < 0;
       if (!selected || staleSelection) {
-        const newSelection = _.get(groupMatches, '[0][0]') || shipMatches[0];
+        const newSelection = (groupMatches[0] && groupMatches[0].length) ? groupMatches[0][0] : shipMatches[0];
```
liam-fitzgerald, line 181: should be able to use the elvis operator here instead

tylershuster:
Seems like a cowboy hunting dependencies might set his sights on codemirror instead 🤔

T

pkg/interface/src/lib/util.js
```diff
@@ -88,7 +86,7 @@ function hexToDec(hex) {
 }
 
 export function hexToRgba(hex, a) {
-  const [r,g,b] = _.chunk(hex, 2).map(hexToDec);
+  const [r,g,b] = /^([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex).map(hexToDec);
```
liam-fitzgerald, line 89: probably want to include uppercase here.

ohAitch:
But syntax highlighting! …I wonder how big a hacked-down read-only build of
codemirror would be

tylershuster:
FWIW, I don’t particularly have a strong opinion about removing lodash. Just going for low hanging fruit. But if it’s not ripe I won’t pull

T

tylershuster:
@liam-fitzgerald your comments are all well-taken. I think some of my edits are fine, but not really enough for a PR. I would love to see Landscape done purely functionally but I don't think it's the time for that even if it is the Right Thing to Do. Unless @tacryt-socryp has some really crushing arguments in favor of removing lodash, I'm going to close this and say thank you for your consideration.

PR #2874, opened 2020-05-08
chat-js: fix scroll slam

liam-fitzgerald:
Locks the scroll upon scrolling half a screenful or 200px from the
bottom, whichever comes first. Correspondingly unlocks the scroll upon
scrolling inside this distance. 

fixes #2864 

cc: @urcades

Issue #2864, opened 2020-05-06
Chat slams to bottom on getting a new message


matildepark:
We only want Chat to slam to bottom when *we* send a message, not when we get one, as per @urcades.

loganallenc:
what? why? @urcades that’s distinctly different than previous expectations about this

matildepark:
@loganallenc [Here](https://github.com/urbit/urbit/pull/2825#issuecomment-621296359)'s the comment in question.

loganallenc:
From my reading, his comment coheres with my expectations around the scroll behavior, which is that if it’s close to the bottom of the screen, and a new message appears, it should scroll to the bottom. Otherwise, scroll position is unchanged.

matildepark:
![image](https://user-images.githubusercontent.com/20846414/81198572-a2a46700-8f8f-11ea-9a37-306cb3c8bf62.png)

matildepark:
This is beside the point of this bug, which is to say, high-traffic rooms are more than a little unfriendly to scrollback reading at the moment. I'm trying to think about how urgent this is to fix and get out.

urcades:
> From my reading, his comment coheres with my expectations around the scroll behavior, which is that if it’s **close to the bottom of the screen**, and a new message appears, it should scroll to the bottom. Otherwise, scroll position is unchanged.

Emphasizing the bit here I think is under contention via this issue, which is what we define as being "close to the bottom of the chat" — my thought is that our definition of "close" should be relatively small, as to Matilde's point, you don't want to be trying to backtrace a conversation that's only ½ or 1 viewport away from the bottom of the scroll, only to have it snap back into position when someone adds as new message.

I suggested in the original comment that we limit the "snap to bottom" affordance to be the height of one comment, which still seems mostly correct — it's hard to know for sure without experimenting with how this affordance feels in a busy channel.

loganallenc:
My suggestion to Liam was a heuristic of 100px - 200px from bottom.

matildepark:
#3242 should fix this.

galenwp:
As was just being discussed: this does seem to be continuing to occur. Did #3242 go out?

matildepark:
Yes, it went out last week. The current behaviour is that the scroll is locked unless you're on the newest page (20 messages). I am assuming you're experiencing it no matter how far back you are?

urcades:
This conversation got me thinking, rather than a page, what if the
threshold of stickiness was literally a single message/post? Like if you
[...]

PR #3242, opened 2020-07-31
interface: refactored chat window into header, window, and scroll container

tacryt-socryp:
Maintained all functionality, just split it into different files to organize things.

As far as I can tell from testing, everything works in an equivalent manner. Big change, so please help out with testing!


pkg/interface/src/apps/chat/components/lib/chat-messages.tsx
```diff
@@ -0,0 +1,96 @@
+import React, { Component, Fragment } from "react";
+import moment from "moment";
+
+import { Message } from "./message";
+
+type IMessage = Envelope & { pending?: boolean };
+
+
+export class ChatMessages extends Component {
```
liam-fitzgerald, line 9: typing props would be nice. e.g. `extends Component<ChatMessagesProps>`

pkg/interface/src/apps/chat/components/lib/chat-header.js
```diff
@@ -0,0 +1,68 @@
...
+              {title}
+            </h2>
+          </Link>
+          <ChatTabBar
+            {...props}
+            station={props.station}
+            numPeers={group.length}
+            isOwner={deSig(props.match.params.ship) === window.ship}
+            popout={props.popout}
+            api={props.api}
```
liam-fitzgerald, line 62: some of these props are made redundant by `{..props}`

pkg/interface/src/apps/chat/components/lib/chat-messages.tsx
```diff
@@ -0,0 +1,96 @@
...
+      if (unread > 0 && i === unread - 1) {
+        return (
+          <Fragment key={msg.uid}>
+            {messageElem}
+            <div
+              ref={this.setUnreadMarker}
+              className="mv2 green2 flex items-center f9"
+            >
+              <hr className="dn-s ma0 w2 b--green2 bt-0" />
+              <p className="mh4">New messages below</p>
```
liam-fitzgerald, line 65: might be nice to refactor into `renderUnreadDivider()` and `renderDayBreakDivider()` or the like.

pkg/interface/src/apps/chat/components/lib/chat-header.js
```diff
@@ -0,0 +1,68 @@
+import React, { Component, Fragment } from "react";
+import { Link } from "react-router-dom";
+
+import { ChatTabBar } from "./chat-tabbar";
+import { SidebarSwitcher } from "../../../../components/SidebarSwitch";
+import { deSig } from "../../../../lib/util";
+
+
+export class ChatHeader extends Component {
```
liam-fitzgerald, line 9: TS would be nice, but not a dealbreaker

pkg/interface/src/apps/chat/components/lib/chat-header.js
```diff
@@ -0,0 +1,68 @@
...
+    const { props } = this;
+    const isinPopout = props.popout ? "popout/" : "";
+    const group = Array.from(props.group.members);
+    let title = props.station.substr(1);
+    if (props.association && "metadata" in props.association) {
+      title =
+        props.association.metadata.title !== ""
+          ? props.association.metadata.title
+          : props.station.substr(1);
+    }
```
liam-fitzgerald, line 21:
```suggestion
    if (props.association && 
      "metadata" in props.association && 
      props.association.metadata.tile !== "") {
      title = props.association.metadata.title
    }
```

liam-fitzgerald:
> * need to set scroll lock properly when inactive (do we actually need this feature? seems unnecessarily bespoke 


cc: @urcades 
FWIW, I really like the current scrolling behaviour and I'd be hesistant to change it, given the number of iterations we did on this.

pkg/interface/src/apps/chat/components/lib/chat-header.js
tacryt-socryp: I don't want to pass in unnecessary props.

urcades:
In order to make a semblance of a call here, I need to know what's precisely meant by:

> need to set scroll lock properly when inactive (do we actually need this feature? seems unnecessarily bespoke

Specifically, what a "scroll lock" is — and what we meant when we say "current scrolling behaviour".

Is this discussion with respect to the **literal mechanic** how we 'freeze' people's chat where they last left off, and keep them in-place for when they revisit a chat? 
Is this discussion with respect to the "_n_ new messages since last visit" bar that populates a chat when this mechanism is triggered? Are we discussing the whole thing here?

If we're discussing the mechanic as a broad structure: I'd argue for its inclusion in this refactor seeing as "respecting the scroll-journey of users" is a frequently abused anti-pattern various feed-based interfaces leverage, generally for the purpose of showing ads and algorithmic reshuffled feeds and the like. Seeing as we're not about that, I feel rather strongly this is a mechanic we should try to keep intact — not only in chat, but as a pattern replicated in a variety of scrolling contexts we might end up building in modules.

If this isn't what we're discussing here, please let me know!

urcades:
Deeply into this PR regardless, these separations seem incredibly good/sane for the purposes of further development — we should replicate this in all other modules we develop.

matildepark:
> need to set scroll lock properly when inactive (do we actually need this feature? seems unnecessarily bespoke

I also don't follow this. Currently our scroll doesn't lock properly (see #2874) and slams to the bottom regardless of where you are. 

We basically want the scroll to float without being pinned to bottom unless we're at the bottom or if we send a message. If there's logic in there that seems very bespoke, it also likely is currently non-functional, I think.

pkg/interface/src/apps/chat/components/lib/chat-header.js
liam-fitzgerald: you're passing them in anyway with `{...props}` on line 57.

tacryt-socryp: Good point, got rid of them!

pkg/interface/src/apps/chat/components/lib/chat-messages.tsx
tacryt-socryp: Killed this whole file.

pkg/interface/src/apps/chat/components/lib/chat-header.js
tacryt-socryp: TS seems excessive for a component with no logic.

matildepark:
Some very weird scroll behaviour. Scroll up and it immediately pulls you back down a page at a time.
Aug-06-2020 16-02-41

Brave.

Also the "mark as read" indicator doesn't dismiss once you hit bottom; only once you click it manually.

tacryt-socryp:
I've only tested in Safari. Will look at Brave behavior ☹️

matildepark:
Well it would probably affect all Blink browsers. I bet Firefox works great.

tacryt-socryp:
@matildepark fixed!

matildepark:
Say I want to see what happened yesterday; I scroll a step into the backlog and look at the mid-point of it. I get forced out and bumped back down into the latest page. If I scroll *really* far this doesn't happen but otherwise this is pretty consistent.

Aug-06-2020 17-26-11

tacryt-socryp:
@matildepark Fixed it, I had introduced a bug with resetting pageNum to 1 in my last commit.

matildepark:
Works on Brave. Scroll load in Firefox works, with the caveat that it jumps *straight* to the top of the loaded backlog page. Don't know if scroll slamming is averted in Firefox. But mostly addresses #2864, maybe put "Fixes ####" in one of these commits so the issue auto-closes once it hits master.

tacryt-socryp:
I'll just manually close it myself..

PR #3243, opened 2020-08-01
lib-hooks: document with comments

liam-fitzgerald:
documents both the helper libraries for building hooks.

### DIFF
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -1,3 +1,24 @@
+::  lib/pull-hook/hoon: helper for creating a push hook
+::  
+::   lib/pull-hook is a helper for automatically pulling data from a
+::   corresponding push-hook to a store. 
+::   
+::   ## Interfacing notes:
+::
+::   lib/push-hook provides the inner door with its own separate state.
+::   The inner door may interact with the library by producing cards or
+::   scrying. Do not pass any cards on a wire beginning with /helper as
+::   these wires are reserved by this library. Any watches/pokes/peeks
+::   not listed below will be routed to the inner door.
+::
+::   ##  Subscription paths
+::
+::   /tracking: The set of resources we are pulling
+::
+::   ##  Pokes
+::
+::   %pull-hook-action: Add/remove a resource from pulling.
+::
 /-  *pull-hook
 /+  default-agent, resource
 ::
@@ -5,13 +26,24 @@
 |%
 +$  card   card:agent:gall
 ::
+::  $config: configuration for the pull hook
+::
+::    .store-name: name of the store to proxy subscriptions to.
+::    .update-mark: mark that updates will be tagged with
+::    .push-hook-name: name of the corresponding push-hook
+
 +$  config
   $:  store-name=term
       update=mold
       update-mark=term
       push-hook-name=term
   ==
+::  
+::  $state-0: state for the pull hook
 ::
+::    .tracking: a map of resources we are pulling, and the ships that
+::    we are pulling them from.
+::    .inner-state: state given to internal door
 +$  state-0
   $:  %0
     tracking=(map resource ship)
@@ -37,7 +69,7 @@
   |*  config
   $_  ^|
   |_  bowl:gall
-  ::
+  ::  from agent:gall
   ++  on-init
     *[(list card) _^|(..on-init)]
   ::
@@ -77,7 +109,9 @@
     *[(list card) _^|(..on-init)]
   ::  +on-pull-nack: handle failed pull subscription
   ::
-  ::    This arm is called when a pull subscription fails.
+  ::    This arm is called when a pull subscription fails. lib/pull-hook
+  ::    will automatically delete the resource from .tracking by the
+  ::    time this arm is called.
   ::
   ++  on-pull-nack
     |~  [resource tang]
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -5,10 +5,10 @@
 ::   
 ::   ## Interfacing notes:
 ::
-::   The inner door may interact with the library by producing cards or
-::   scrying. Do not pass any cards on a wire beginning with /helper as
-::   these wires are reserved by this library. Any watches/pokes/peeks
-::   not listed below will be routed to the inner door.
+::   The inner door may interact with the library by producing cards.
+::   Do not pass any cards on a wire beginning with /helper as these
+::   wires are reserved by this library. Any watches/pokes/peeks not
+::   listed below will be routed to the inner door.
 ::
 ::   ##  Subscription paths
 ::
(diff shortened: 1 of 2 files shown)


pkg/arvo/lib/pull-hook.hoon
```diff
@@ -1,17 +1,49 @@
+::  lib/pull-hook/hoon: helper for creating a push hook
```
Fang-, line 1:
```suggestion
::  lib/pull-hook: helper for creating a pull-hook
```

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -1,17 +1,49 @@
...
+::  
+::   lib/pull-hook is a helper for automatically pulling data from a
+::   corresponding push-hook to a store. 
+::   
+::   ## Interfacing notes:
+::
+::   lib/push-hook provides the inner door with its own separate state.
```
Fang-, line 8: The phrasing here feels a bit confusing, but all it's really saying is that the inner agent's `+on-save` result gets threaded through to its `+on-load` transparently, right? Might not even be worth mentioning, if it just preserves expected behavior. But ymmv.

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -1,17 +1,49 @@
...
+::   The inner door may interact with the library by producing cards or
+::   scrying. Do not pass any cards on a wire beginning with /helper as
```
Fang-, line 10: >or scrying

I don't see any special scry handling in this. If there was, you'd want to mention the prefix path used for these.

But also, this seems potentially problematic. At the very least you want to pass it onto the underlying agent, right?

https://github.com/urbit/urbit/blob/d91de5f2769060e81bd46c696f373064a9bf9033/pkg/arvo/lib/pull-hook.hoon#L246

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -1,17 +1,49 @@
...
 ::
 |%
 +$  card   card:agent:gall
 ::
+::  $config: configuration for the pull hook
+::
+::    .store-name: name of the store to proxy subscriptions to.
+::    .update-mark: mark that updates will be tagged with
+::    .push-hook-name: name of the corresponding push-hook
+
```
Fang-, line 34:
```suggestion
::
```

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -1,17 +1,49 @@
...
 /+  default-agent, resource
 ::
 ::
 |%
 +$  card   card:agent:gall
 ::
+::  $config: configuration for the pull hook
+::
+::    .store-name: name of the store to proxy subscriptions to.
+::    .update-mark: mark that updates will be tagged with
```
Fang-, line 32: Suggestion:

```suggestion
::    .store-name: name of the store to send subscription updates to
::    .update-mark: mark that updates will be tagged with
```

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -77,7 +109,9 @@
     *[(list card) _^|(..on-init)]
   ::  +on-pull-nack: handle failed pull subscription
```
Fang-, line 110: I'd recommend moving these specialized arms to be _above_ the standard gall ones, so you can find them more quickly when reading top-to-bottom.

pkg/arvo/lib/push-hook.hoon
```diff
@@ -1,15 +1,55 @@
+::  lib/push-hook/hoon: helper for creating a push hook
```
Fang-, line 1:
```suggestion
::  lib/push-hook: helper for creating a push-hook
```

pkg/arvo/lib/push-hook.hoon
```diff
@@ -1,15 +1,55 @@
...
+::  
+::   lib/push-hook is a helper for automatically pushing data from a
+::   local store to the corresponding pull-hook on remote ships. It also
+::   proxies remote pokes to the store.
```
Fang-, line 5: > automatically pushing data from a local store to ... remote ships

This is why I don't like the "pull-push" phrasing. You're not so much _pushing_ as you are answering requests. Am I alone in thinking this is weird?

pkg/arvo/lib/push-hook.hoon
```diff
@@ -1,15 +1,55 @@
...
+::
+::   ## Interfacing notes:
+::
+::   lib/push-hook provides the inner door with its own separate state.
+::   The inner door may interact with the library by producing cards or
+::   scrying. Do not pass any cards on a wire beginning with /helper as
```
Fang-, line 11: There are no special scries here, either. But at least it passes `+on-peek` calls through to `og` properly!

pkg/arvo/lib/pull-hook.hoon
Fang-: (Also `/helper` is a bit unhelpful as far as descriptive prefixes go. It's the generic name, but you might want to have multiple helper libraries that do different things, etc. etc.)

pkg/arvo/lib/push-hook.hoon
```diff
@@ -1,15 +1,55 @@
...
+::   these wires are reserved by this library. Any watches/pokes/peeks
+::   not listed below will be routed to the inner door.
+::
+::   ##  Subscription paths
+::
+::   /resource/[resource]: Receive initial state and updates to
+::   .resource. .resource should be encoded with de-path:resource from
```
Fang-, line 18:
```suggestion
::   .resource. .resource should be encoded with en-path:resource from
```

pkg/arvo/lib/push-hook.hoon
liam-fitzgerald: I do think the listen/proxy dichotomy is clearer, but pull/push is clearer.

pkg/arvo/lib/pull-hook.hoon
liam-fitzgerald: Oh the plan was to have `/helper` be generally a reserved prefix for helper libraries. i.e. pull-hook uses /helper/pull-hook and push-hook uses /helper/push-hook as their wire prefixes.

pkg/arvo/lib/push-hook.hoon
tacryt-socryp: I like having symmetrical, short names. Lmk if you have recommendations

Fang-: Symmetrical in what sense? Character length, or semantics?

I think that's less important than being clear about what these things do. Best alternative I got is hear/show.

matildepark:
@Fang- anything left here?

Fang-:
The comments still say the wrapper agents can be interacted with by scrying, even though they don't do any scry handling themselves. That should be the last thing left though!

matildepark:
@liam-fitzgerald It's been a month for a comment change, let's get this sweeped up

Issue #3239, opened 2020-07-31
Landscape Javascript Cache Busting


tylershuster:
**Summary**: Append base hash to `<script>` tags to bust Javascript caches

**Details**:
Currently JavaScript is loaded like this:

https://github.com/urbit/urbit/blob/baaea9a2a447564fa185af36f537c71180663f35/pkg/arvo/app/landscape/index.html#L24-L26

As a result of the file having the same URL, OTA Javascript changes often take up to a day to resolve if not cleared (and caching can be unreasonably sticky).

Adding the base hash as a query string, e.g. `<script src="/~landscape/js/channel.js?bmhb4"></script>` would change the URL of the script every OTA and therefore a) bust caches while b) retaining the proper caching found here: https://github.com/urbit/urbit/blob/baaea9a2a447564fa185af36f537c71180663f35/pkg/arvo/app/file-server.hoon#L197

However, this requires a few things which I'm not knowledgeable enough to do and am therefore tagging @Fang- and @tacryt-socryp.

Primarily it requires `index.html` to become `index.hoon`
- This requires a change in how `%file-server` serves landscape, with the mechanics of which I am unfamiliar.
- This also requires a pretty funky change to the manifest to render it in udon, which might be better done by dynamic loading. I'm unsure how that affects its functionality. I'm happy to explore these options.

The other, bad option, is to load minimal Javascript first to scry the ship, get the hash, and dynamically load the full JavaScript afterward.

liam-fitzgerald:
Tbh, i'm in favour of moving caching client-side. Network level caching behaviour can vary pretty widely between browsers. (We can retain server-side caching if we want, we'll just have to remove it on index.html)

I don't think it's necessary that index.html become index.hoon because there's already a manual intervention to update the glob hash, we should just update index.html in that glob commit.

I'm not 100% on this so RFC i guess

ohAitch:
As long as we're distributing them by glob anyway, might as well
/~landscape/js/channel.bmhb4.js rather than /~landscape/js/channel.js?bmhb4

tacryt-socryp:
I would like to have 30 day caching for JS + CSS files and no caching for index.html. No need for an index.hoon. Agree with Liam.

liam-fitzgerald:
closed by #3249

PR #3249, opened 2020-08-03
interface: do not cache index.html and cache bust JS

liam-fitzgerald:
- Strips index.html of all caching headers. 
- Webpack now outputs a hashed filename, .e.g. `index.11130a034c1b98f912ee.js`
- Updated glob to handle new bundle name.
- Updates MAINTAINERS.md to instruct the deployer to update the index.html to pull in the latest JS. (PR for janeway otw)
- Remove index.js being treated as both the entry point and a chunk, which was causing the bundle to be twice as large as it should be.


matildepark:
Always flexin...

tylershuster:
Fixes #3239

matildepark:
Doesn't this break the dev server?

liam-fitzgerald:
> Doesn't this break the dev server?

not anymore

liam-fitzgerald:
fwiw this also opens the door to bundling our dependencies separately, which could be a nice perf improvement.

matildepark:
pong @g-a-v-i-n also this has a conflict

tacryt-socryp:
Have you tested the OTA case for this? I worry that it would break the webpage by unbinding 
/~landscape/index.js

matildepark:
Aye, might be worth just doing a once-over on an -L ship; my guess is that it will certainly blank at first, because of a residually cached index.html, but work afterward (and for all subsequent updates).

—
~haddef-sigwen
https://urbit.org

liam-fitzgerald:
OTA case tested. works fine

matildepark:
And if we merge this we won't be able to run the dev server against livenet right? The proxy won't block the original landscape.

liam-fitzgerald:
it does not break the dev server for livenet or for new ships.

matildepark:
I meant “until we OTA this” heh

liam-fitzgerald:
> I meant “until we OTA this” heh

No it still works.

tacryt-socryp:
Good to go.

matildepark:
> > I meant “until we OTA this” heh
> 
> No it still works.

Just a heads up that this is straight up not true; doing `npm run start` now against a livenet ship and it's serving old landscape instead of the dev server.

matildepark:
Worked around this by changing L55 in webpack.dev.js back to this for the time being so we proxy against the actual location of the old index.js:

```
      '/~landscape/js/index.js': {
```

PR #3255, opened 2020-08-06
interface: lint adds max-lines, max-lines per function

matildepark:
- Linter now treats files above 200 lines as having an error.
- Linter now treats functions above 50 lines as having an error.

Blanks and comments don't count.

You will notice (or think you are very very smart) that the linter doesn't run on TypeScript files at the moment. This is because it needs, basically, its own parser. I investigated this; and moved to a TS parser, but it began to call TypeScript errors on JS files (like dictating return types). I tried a few overrides, which got complicated fast. If any JS programmers in the audience know about ESLinting mixed codebases, by all means, have at it.


tylershuster:
I get the below when I clone this and run `npm run lint` in `pkg/interface`. Am I doing something wrong?

```
ESLint couldn't find the config "eslint/recommended" to extend from. Please check that the name of the config is correct.

The config "eslint/recommended" was referenced from the config file in "/Users/tyler/Code/sandbox/ts-parser/pkg/interface/.eslintrc.json".
```

matildepark:
Oh, I probably made a typo when I was undoing ts stuff...good catch.

Fang-:
I guess the reasoning here is "big files are always complex"? I'm not sure how well that holds. For example, JSX blocks can get pretty tall, but are usually just dumb "pls display this" lists.

Have you considered starting with higher limits and then gradually lowering over time? There's an opposite-end-of-the-spectrum problem here where your logic is too spread out to track, you don't want to push yourself into that either.

matildepark:
I'd be fine with a moderate middle-ground (350?). This flags most of our codebase as is; and the code smell gets stronger on a rather quadratic curve. I'll let the other reviewers weigh in before merging.

tacryt-socryp:
350 is too large. 250 is acceptable, 300 is a file that's already getting scary.

matildepark:
So we want 300 to flag as an error then. Yeah?

tacryt-socryp:
> I guess the reasoning here is "big files are always complex"? I'm not sure how well that holds. For example, JSX blocks can get pretty tall, but are usually just dumb "pls display this" lists.
> 
> 
> 
> Have you considered starting with higher limits and then gradually lowering over time? There's an opposite-end-of-the-spectrum problem here where your logic is too spread out to track, you don't want to push yourself into that either.

Worth noting that so long as you aren't passing data back up from children to parents often, it's very difficult to get this problem in React due to the unidirectional data flow of our Redux-like store pattern.

g-a-v-i-n:
Like this idea, even if its not going to be a hard stop necessarily. Agree with tacryt re: size. If logic is too spread out to track then the logic is probably poorly compartmentalized. Frontend components should really address appearance only, with as little business logic as possible. Props should be the API which controls the appearance. Following this model is part of why the React team introduced hooks and is less enthusiastic about class components.

matildepark:
Alrighty, so we have a gentler hard cap, TS linting (thanks @tylershuster) and some buy-in. @liam-fitzgerald didn't object in our original discussion, so I think this is fine to go ahead and merge.

PR #3256, opened 2020-08-06
Interface: Added typescript linting

tylershuster:
This adds typescript linting. This is an initial shot, but it doesn't crash when you run `npm run lint` so

@matildepark


matildepark:
You wanna target my other PR so we can do this as an all-in-one?

matildepark:
Of course then you’ll have to include its changes (this PR undoes them)

PR #3259, opened 2020-08-06
publish: send reads on nav, distinguish unreads

matildepark:
A bandaid for the @galenwp in the audience.

During the SPA refactor we did a rather large renovation of the Publish back-end / front-end interaction pattern. It worked via API calls on navigation instead of the PHP-style state injection. I wrote the logic to pre-empt sending read actions before we knew what we were looking at, by ensuring we had state instantiated first. Here I did so clumsily, so that reads would never send. My bad.

I also went ahead and decided to follow Links in how it marks "unread content": unread posts now have their dates written in green. No "mark all as read" button yet (we'll have to do [something like this](https://github.com/urbit/urbit/blob/master/pkg/interface/src/apps/dojo/components/input.js#L90-L99) where we reduce an array of notes with Promises, waiting for each subsequent Promise to resolve before proceeding with the next, which would be very slow and not ideal in Publish's case, potentially).

### DIFF
--- a/pkg/interface/src/apps/publish/components/lib/note.js
+++ b/pkg/interface/src/apps/publish/components/lib/note.js
@@ -12,7 +12,8 @@ export class Note extends Component {
   constructor(props) {
     super(props);
     this.state = {
-      deleting: false
+      deleting: false,
+      sentRead: false
     };
     moment.updateLocale('en', {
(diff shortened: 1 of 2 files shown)


galenwp:
Thank you!

PR #3265, opened 2020-08-08
chat: Fixed over-large images

tylershuster:
This fixes an issue where very small images would become scaled up, and instead just limits images to a max size.

I also linted the file while I was in there.


matildepark:
I can't review this because your release/next-userspace is significantly behind ours — I think this file isn't even there anymore. Can you pull from upstream and rebase this PR?

pkg/interface/src/apps/chat/components/lib/content/url.js
```diff
@@ -1,8 +1,6 @@
 import React, { Component } from 'react';
 
-
-const IMAGE_REGEX = 
-  /(jpg|img|png|gif|tiff|jpeg|JPG|IMG|PNG|TIFF|GIF|webp|WEBP|webm|WEBM|svg|SVG)$/;
+const IMAGE_REGEX = new RegExp(/(jpg|img|png|gif|tiff|jpeg|webp|webm|svg)$/i);
```
tylershuster, line 3: Same formats, just case-insensitive.TiFf

matildepark:
For future reference this should be `chat:`

PR #3266, opened 2020-08-09
hoon: add +lead and +late

Fang-:
For tacking some head or tail onto some noun, respectively.

I find myself doing things in the vein of `(turn my-list |=(=item [%constant item]))` often enough that I wanted a little function to just be there for me. Appropriately filed under the "functional hacks" section.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -262,6 +262,9 @@
 ++  tail  |*(^ ,:+<+)                                   ::  get tail
 ++  test  |=(^ =(+<- +<+))                              ::  equality
 ::
+++  lead  |*(h=* |*(* [h +<]))                          ::  put head
+++  late  |*(t=* |*(* [+< t]))                          ::  put tail
+::
 ::  #  %containers
 ::
 ::    the most basic of data types


belisarius222:
I've also found myself wanting these frequently.

pkg/arvo/sys/hoon.hoon
```diff
@@ -262,6 +262,9 @@
 ++  tail  |*(^ ,:+<+)                                   ::  get tail
 ++  test  |=(^ =(+<- +<+))                              ::  equality
 ::
+++  lead  |*(h=* |*(* [h +<]))                          ::  put head
```
pilfer-pandex, line 265: Why |*?

pilfer-pandex:
Cons and snoc would be the traditional names for these operations, but they're already taken. I also kind of like these; nice wordsmithing.

pkg/arvo/sys/hoon.hoon
Fang-: You'll want the types of whatever nouns you pass in to be preserved, so that the eventual result can nest correctly, right?

pilfer-pandex: GrrrrRRRr. Fine.

Issue #1332, opened 2019-07-12
Mention that Landscape is running in Urbit boot messages


msutherl:
Urbit currently doesn't give any indication that Landscape/Modulo exists. Wondering if it would be reasonable to include a message at the end of the boot sequence that says something like:

`Modulo running on https://localhost:[port]`

I suppose the most elegant way to do this would be from the Modulo app itself.

ixv:
Sure thats definitely possible. It would be somewhat redundant with the `http: live (insecure, public) on 8080` print though.

joemfb:
> activated app home/modulo

*technically speaking* ...

vvisigoth:
AKSHUALLY

msutherl:
@ixv maybe you're right. For consistency – or perhaps grotesque inconsistency – could be something like:

```
http: live (insecure, public) on 8080
modulo: live on localhost:8080
```

But there's something gross about that.

@joemfb perhaps the **right way** to do this would be to let each app publish its own "I'm activated" message.

```
activated app home/modulo: live on localhost:8080 🍅
```

joemfb:
yikes

msutherl:
@joemfb wait, you're right – that would be awful. The more I think about this the more I think apps shouldn't do this, even though it causes a temporary annoying UX gap.

joemfb:
It's just that the thing that's started on 8080 or whatever is a general-purpose HTTP server, those "activated" messages tell you what applications are automatically started for you, etc. It doesn't feel right to specialize those basic system notifications for onboarding.

At the same time. I do think that urbit needs some kind of first-boot wizard/tour/etc. The UX is not friendly, and there should be some way to direct people to the happy path.

msutherl:
@joemfb I had immediately dismissed the idea as unpalatable, but with your blessing, a dedicated "welcome" message that can easily be disabled by the purist would be great. Where would one ideally insert such a thing?

joemfb:
The fully general answer to this is actually really hard. I talked to @galenwp about this awhile ago, but haven't had a chance to circle back. There isn't any central orchestration of all the layers of the system, so it's kind of hard to even answer when "boot" is complete (for some expansive definition of "boot"). But this doesn't to be fully general.

I'd say that if the message is short it can be completely in the console. If it's long, I'd make it available from the login page (maybe with associated server-side state to track if it's been dismissed, etc.), and just have a short message in the console directing the user to the longer one.

[...]

PR #3267, opened 2020-08-10
king: eyre: add missing printouts with ports

### LINKED ISSUE #3175: urbit-king doesn't print the http port in the dojo on boot
**Describe the bug**
urbit-king doesn't print its http ports in the dojo when it boots, unlike the c king which says:
http: web interface live on http://localhost:8081
(or whatever)

**To Reproduce**
1. spin up pier with urbit-king
2. look at what it spits out in the terminal

**Expected behaviour**
It should behave like the c king

**Screenshots**
```
> urbit-king run -p 42067 piers/lagtuc-socleg-tinnus-napbus
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/469.991.424
boot: installed 268 jets
~lagtuc-socleg-tinnus-napbus:dojo> 
```

 - OS: linux64
 - 0.10.8 (urbit-king)
 - jp5fi

pilfer-pandex:
fixes #3175

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Eyre.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Eyre.hs
@@ -176,8 +176,9 @@ startServ
   -> Bool
   -> HttpServerConf
   -> (EvErr -> STM ())
+  -> (Text -> RIO e ())
   -> RIO e Serv
-startServ multi who isFake conf plan = do
+startServ multi who isFake conf plan stderr = do
   logDebug (displayShow ("EYRE", "startServ"))
 
   let vLive = meaLive multi
@@ -270,6 +271,10 @@ startServ multi who isFake conf plan = do
       fil = pierPath <> "/.http.ports"
 
   logDebug $ displayShow ("EYRE", "All Servers Started.", srvId, por, fil)
+  for secPor $ \p ->
+    stderr ("http: secure web interface live on https://localhost:" <> tshow p)
+  stderr ("http: web interface live on http://localhost:" <> tshow insPor)
+  stderr ("http: loopback live on http://localhost:" <> tshow lopPor)
 
   pure (Serv srvId conf lop ins mSec por fil vLive)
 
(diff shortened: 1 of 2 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Eyre.hs
```diff
@@ -270,6 +271,10 @@ startServ multi who isFake conf plan = do
       fil = pierPath <> "/.http.ports"
 
   logDebug $ displayShow ("EYRE", "All Servers Started.", srvId, por, fil)
+  for secPor $ \p ->
+    stderr ("http: secure web interface live on https://localhost:" <> tshow p)
```
eglaysher, line 275: You probably don't want to mention "localhost" in these. Vere doesn't and most of the time, a user won't be accessing it via 127.0.0.1. In the case of the secure web interface, it's wrong since the SSL certificate is extremely unlikely to match "localhost".

joemfb: I don't love these messages, but it looks to me like they match #3092

eglaysher: Ah, it looks like I was looking at an out-dated master locally. Up to you ~pilfer if you want to make this change.

pilfer-pandex: Joe's right that I just copied verbatim the behavior of C here (except for adding the word "secure" in the https case), including the fact that the text "localhost" is hardcoded. I agree that it's a bit weird.

I'm happy to do whatever you guys want here; just let me know what it is.

Fang-: If it is, somehow, easy to know what IP/domain we're available on, do put that in there.

The reason I put "localhost" in #3092 is because I just wanted to fix the complaints from https://github.com/urbit/urbit/issues/1332#issuecomment-513029707, and figuring those things out was much more involved than I cared for. If you're not running locally, you're likely to understand the implications of "localhost" here.

pilfer-pandex:
How would the king even know that it's at ship.urbit.org? For this, it'd
have to know 1) what ip address it is at and 2) what domain name(s) point
to that ip. And what if there's some complex http proxying scheme going on,
like there usually is? "What domain I represent," IIRC, is generally
implemented as a *piece of configuration* that is input to the program by
the complex proxying environment that runs it, if the program needs to
know, right? Also, if billy is running his ship on our hosting system, does
he even have access to the cli and these printouts?

Fang-:
It was more of a rhetorical suggestion I guess, for all the reasons you described. I think Eyre keeps around some config state wrt this, but in practice "It's Hard" and we can't "Really Know" unless we're explicitly told.

Of course at that point, and indeed in the hosting environment too, what message gets printed here doesn't really matter.

(Didn't mean to hold this up, it's fine to go in afaic!)

joemfb:
Note that your urbit already needs to be told this in order to request a TLS certificate from LetsEncrypt. And king-haskell needs to be told this when running multiple ships in a single process with a shared http server (in order to route requests appropriately). It's not at all out of the question for this information to be available, but the requirements would need to be nailed down and the interfaces/data-flow would need to be designed.

pilfer-pandex:
Seems like we have consensus that this is fine for now, so I'll merge. Thanks guys.

Issue #3280, opened 2020-08-11
Leap interface background dark mode: Make dark


urcades:
Hello — I noticed in dark mode landscape, when invoking leap, I receive the following:

<img width="1502" alt="Screen Shot 2020-08-11 at 2 46 04 AM" src="https://user-images.githubusercontent.com/1195363/89866120-fde08980-db7c-11ea-90b7-96a8082bfa0a.png">

If we change this to `background-color: rgba(0, 0, 0, 0.3);`, we get the following:

<img width="1502" alt="Screen Shot 2020-08-11 at 2 48 08 AM" src="https://user-images.githubusercontent.com/1195363/89866238-33857280-db7d-11ea-93f5-8ab3e8c36f19.png">

It feels more visually correct for a dark mode!

Curious on the takes of dark mode aficionados.

tacryt-socryp:
I want the dark background to always be that dark tbh

matildepark:
Have to look into how we do this. Because we use `gray` as the background-color, which is based on opacities, it of course does the opposite in each direction.

Issue #3279, opened 2020-08-11
leap: apps are below subscriptions in search results


mike-cr:
Search results in leap seem to be sorted commands \> subscriptions \> apps. If you search "chat" it returns all chat subscriptions, so if you're trying to go to the chat app you have to scroll past a large list to get to it. I think apps should be at the top or at least above subscriptions.

Issue #3277, opened 2020-08-11
search: newline inserted in chat input when navigating to channel with keyboard


mike-cr:
**Describe the bug**
When using the ctrl+/ search (omnibox?), if you select the desired chat using the keyboard (\<tab\>, \<enter\>), the \<enter\> is inserted in the chat input so your cursor starts on the second line.

**To Reproduce**
1. ctrl+/
2. type "general"
3. \<tab\> to select the general chat, \<enter\> to open it
4. Observe your cursor is on the second line of the chat input

**Expected behavior**
the \<enter\> should not be inserted in the chat input

**Desktop (please complete the following information):**
 - OS: arch linux 64-bit
 - Browser: qutebrowser (QTWebEngine aka the chromium engine) and firefox
 - Base hash: dg0qj

Issue #3274, opened 2020-08-11
omnibox: shows all channels regardless of if you're actually subscribed


matildepark:
- If you try and click on a channel you're aware of but not *in* then nothing happens.

Either we want to filter this down (how?) or catch it as a join request in all our apps if you go directly to a channel you aren't subscribed to (probably could do this).

PR #3286, opened 2020-08-11
leap: fixes #3278

### LINKED ISSUE #3278: search: input must be lowercase
**Describe the bug**
If you type "General" rather than "general" the search box it doesn't find the channel. It seems input must be entirely lowercase for it to produce results

**To Reproduce**
1. ctrl+/
2. type "General"
3. find nothing
4. type "general"
5. find the channel

**Expected behavior**
search input should be case insensitive

 - OS: arch linux 64-bit
 - Browser: qutebrowser
 - Base hash: dg0qj

tylershuster:
Fixing a few bugs that shipped with leap. On the list:

- [x] #3277
- [x] #3278
- [x] #3279
- [ ] #3280
- [ ] bonus: #3274 

I'll watch those issues to see if anything develops

### DIFF
--- a/pkg/interface/src/components/Omnibox.js
+++ b/pkg/interface/src/components/Omnibox.js
@@ -70,6 +70,7 @@ export class Omnibox extends Component {
     }
 
     if (evt.key === 'Enter') {
+      evt.preventDefault();
       if (this.state.selected !== '') {
         this.navigate(this.state.selected);
       } else {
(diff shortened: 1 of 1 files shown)


pkg/interface/src/components/Omnibox.js
```diff
@@ -9,6 +9,8 @@ import OmniboxResult from './OmniboxResult';
 import { cite } from '../lib/util';
 
 export class Omnibox extends Component {
+  searchCategories = ['commands', 'apps', 'groups', 'subscriptions'];
```
tylershuster, line 12: DRY so order can be changed or appended

matildepark:
lol I'm doing these very same fixes Tyler.

matildepark:
I actually have all but #3274 written and committed, just writing proper autojoin logic to catch errors.

tylershuster:
I didn't think you'd even be up.

PR #3287, opened 2020-08-11
leap: small fixes

matildepark:
(Everyone calls it "leap" and "leap" is four characters, I guess I'll start calling it "leap" after this.)

- Enter no longer propagates to inputs you navigate to.
- Search queries are now case insensitive.
- Results are sorted 'apps', 'commands', 'groups', 'subs'
- Background darkens in both light and dark mode (using `scales.black30` instead of `lightGray`, which swaps between white and black in each mode with some opacity)

Still pining after #3274, but it needs some investigation; I'm getting infinite loops trying to track down how to execute, and these are self contained.

@tylershuster wrote a duplicate PR of these as I was writing them. Sorry Tyler.

### DIFF
--- a/pkg/arvo/app/landscape/index.html
+++ b/pkg/arvo/app/landscape/index.html
@@ -4,7 +4,7 @@
     <title>OS1</title>
     <meta charset="utf-8" />
     <meta name="viewport"
-      content="width=device-width, initial-scale=1, shrink-to-fit=no"/>
+      content="width=device-width, initial-scale=1, shrink-to-fit=no,user-scalable=0"/>
       <meta name="apple-mobile-web-app-capable" content="yes" />
       <meta name="apple-touch-fullscreen" content="yes" />
       <meta name="apple-mobile-web-app-status-bar-style" content="default" />
--- a/pkg/arvo/app/landscape/index.html
+++ b/pkg/arvo/app/landscape/index.html
@@ -4,7 +4,7 @@
     <title>OS1</title>
     <meta charset="utf-8" />
     <meta name="viewport"
-      content="width=device-width, initial-scale=1, shrink-to-fit=no,user-scalable=0"/>
+      content="width=device-width, initial-scale=1, shrink-to-fit=no,maximum-scale=1"/>
       <meta name="apple-mobile-web-app-capable" content="yes" />
       <meta name="apple-touch-fullscreen" content="yes" />
       <meta name="apple-mobile-web-app-status-bar-style" content="default" />
--- a/pkg/interface/src/components/Omnibox.js
+++ b/pkg/interface/src/components/Omnibox.js
@@ -217,7 +217,7 @@ export class Omnibox extends Component {
 
     return (
         <Box
-          backgroundColor='lightGray'
+          backgroundColor='scales.black30'
           width='100vw'
           height='100vh'
           position='absolute'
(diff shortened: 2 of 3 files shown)


pkg/interface/src/components/Omnibox.js
```diff
@@ -119,6 +120,8 @@ export class Omnibox extends Component {
       return;
     }
 
+    query = query.toLowerCase();
+
     ['commands', 'subscriptions', 'groups', 'apps'].map((category) => {
```
tylershuster, line 125: I do think DRYing these out into a class property would reduce future headaches

matildepark: Can you explain?

tylershuster: See https://github.com/urbit/urbit/pull/3286/files#r468706373

matildepark: I mean we aren't really repeating ourselves, though; this same `.map` style is in the `search()` behaviour, but it doesn't matter what order we search for results in. So the order of displaying results is in one place, in the `render()`. (It should probably be outside the `render()`, but that's besides the point.

We're only mentioning the order of results once, so directly above instead of in the environment seems preferable.

tylershuster: That follows. The other issue that it fixes is if future things get added into the search, but I suppose that would imply a larger rewrite

tacryt-socryp:
These changes look sane and good. Any fix for the viewport zooming on iOS?

matildepark:
Oh, and the keypress too, right. I will look into it and add to this.

matildepark:
There you go @tacryt-socryp

matildepark:
oh right shoot one more tiny thing, i'll set spell check to false

pkg/arvo/app/landscape/index.html
```diff
@@ -4,7 +4,7 @@
     <title>OS1</title>
     <meta charset="utf-8" />
     <meta name="viewport"
-      content="width=device-width, initial-scale=1, shrink-to-fit=no"/>
+      content="width=device-width, initial-scale=1, shrink-to-fit=no,user-scalable=0"/>
```
tylershuster, line 7: I think people will have a problem with this. See https://stackoverflow.com/a/46254706

matildepark: Who pinch zooms in Landscape atm? I don't think you'll be pinch zooming anywhere.

tylershuster: I have a client named Chas who would disagree. He experiences pretty much every accessibility problem known to man, and I just don't think disabling pinch-to-zoom is a good approach. I mean, you tha boss, but Chas has helped me to realize that these things *matter* to people. I don't know that it's going to be a problem tomorrow or even this year, but this is a global change.

matildepark: Fair enough, I'll test, I'll test...

tacryt-socryp: Personally I think we should disable pinch to zoom.

matildepark: Idk. I get the idea that [aged users](https://accessiblewebsiteservices.com/accessible-web-development-how-to-allow-pinching-to-zoom-text/), for example, may need to look at text 2x. I'm happy to resolve this direct issue (omnibox autofocusing) while leaving the door open here.

PR #3290, opened 2020-08-11
chat: fixes #3288

### LINKED ISSUE #3288: chat: all messages start with newlines
When sending a message from latest (post-leap) webchat, all messages get prefixed with the newline character.

This doesn't show up in web ui, where the browser just eats leading whitespace, but chat-cli diligently renders this:

image

Arguably leading and trailing newlines shouldn't be rendered, but they certainly shouldn't be sent either.

(Additional, likely unnecessary detail: you _will_ notice this in webchat when sending URLs: it'll split your message into a blank line, and the URL.)

tylershuster:
Only preface the message with a newline if we're on subsequent lines


matildepark:
@Fang- — I'm on hosting so I can't actually test this against chat-cli. Would you mind?

PR #3292, opened 2020-08-11
chat: don't show green line when unread is on page 1

tacryt-socryp:


tacryt-socryp:
Worth noting: we now only show the green unread marker when there is more than one page of unread messages.

matildepark:
Yes, I noted this. It's all about tradeoffs, isn't it...

PR #3294, opened 2020-08-11
leap: set default select behavior

tylershuster:
This adds one new behavior: the first result is selected, unless a result that *still* matches the current input has previously been selected

It also fixes a minor bug concerning the order of results and tab-to-select behavior

Furthermore, it follows the new linting convention by breaking the rendering of results into a separate method

### DIFF
--- a/pkg/interface/src/components/Omnibox.js
+++ b/pkg/interface/src/components/Omnibox.js
@@ -23,6 +23,7 @@ export class Omnibox extends Component {
     this.control - this.control.bind(this);
     this.setPreviousSelected = this.setPreviousSelected.bind(this);
     this.setNextSelected = this.setNextSelected.bind(this);
+    this.renderResults = this.renderResults.bind(this);
   }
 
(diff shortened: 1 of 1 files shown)


matildepark:
(Let's just start going with `leap:` and keep it all consistent.)

matildepark:
> This adds one new behavior: the first result is selected, unless a result that still matches the current input has previously been selected

Had difficulty parsing this. I believe it's like, "if you type in a query, every time it will highlight option 1; however, if you select option number 4, then go back and search the same query, option 4 will be highlighted. If you type in a query, select option 4, and then type in a different query, option 1 will be highlighted."

tylershuster:
The description is hard. If you type "mem" and you have "memes," "memory," and "memetics" available "memes" will be highlighted. Then let's say you tab to "memetics." If you then hit "e," "memetics" will stay highlighted. If you then type "s," "memes" will be highlighted because it's the only result that matches. 

Backspacing still clears.

PR #3295, opened 2020-08-11
interface: autojoining known channels

matildepark:
Hi all — this is a RFC, since this PR produces some weird behaviour and quirks.

In #3274, I noted that since we iterate through metadata-store in the Leap prompt, we see channels our group is aware of but we aren't necessarily subscribed to. Right now, if you go to any of those options, nothing happens — you get a default template page in all apps, without any data. 

The ideal case is that if you go to these channels, our apps just automatically grab them for you and subscribe. 

In this PR, I attempt to check if we have an open subscription to the app in question; if we do, and we aren't subscribed to the channel, go join it via the `/join` route.

This *mostly* works, but ...

## Known issues
- If any channel is *invalid* — most commonly this is because the host breached, or the channel was deleted, but the metadata was preserved — basically every app we have can't handle it. Chat and Links go into infinite loops between join and the channel page; Publish idles on the join screen.

We used to say "check your console for a stack trace" (itself *not* a solution to just writing error case handling for joining channels) but we no longer have stack traces in the console. We don't produce any notion of an error in FE or console at all, in fact. 

Somewhere in the APIs being rewritten twice during the SPA work and the global store rewrite, we no longer report `console.error()`, so writing a `catch()` for the rejected promise and throwing an error in FE is not possible. @liam-fitzgerald @tacryt-socryp Any ideas here?

- Links, no matter what, loops at least a few times. We want to wait for the resource to be thrown into `listening` before proceeding to the page, but instead we make the API call and proceed upon it succeeding, which leads to a delay of a few seconds, and thus on first pageload the links are doubled or tripled. This resolves on a refresh.

- But that said, going directly to a Links page now redirects you to the join prompt as well, even if already listening for it, suggesting the `subscribed` constant I'm defining here isn't even tripping correctly. Confusing.

@Fang- suggestions on catching this correctly? We'll be migrating this soon, I know, but you wrote a lot of this FE too.

### DIFF
--- a/pkg/interface/src/apps/chat/components/lib/chat-window.tsx
+++ b/pkg/interface/src/apps/chat/components/lib/chat-window.tsx
@@ -20,7 +20,7 @@ export class ChatWindow extends Component {
     };
 
     this.hasAskedForMessages = false;
-    
+
     this.dismissUnread = this.dismissUnread.bind(this);
     this.scrollIsAtBottom = this.scrollIsAtBottom.bind(this);
     this.scrollIsAtTop = this.scrollIsAtTop.bind(this);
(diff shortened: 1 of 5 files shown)


matildepark:
Took a shower. I'm overthinking this. We have a wonderful device for autojoining any channel, regardless of subscription, and it's what we use in Groups — the `/join` route. Will just change everything in omnibox to it and test.

The console.error is still outstanding, but perhaps out of scope?

PR #3302, opened 2020-08-12
interface: broke out routes from main App.js

tacryt-socryp:
This will make it easier for us to move all of the apps in TwoPaneApp into a single layout component. No functionality changed.


matildepark:
Intriguing.

PR #3308, opened 2020-08-13
interface: restructured outer directories into views and logic

tacryt-socryp:
The outer src directory was getting extraordinarily messy with the combination of the SPA transition, the global store, and the move over to TypeScript. Reorganized into the following outer structure:
- views
  - components
  - apps
  - css
  - themes
  - app.js
- logic
  - api
  - lib
  - reducers
  - store
  - subscription
- types
- index.js

No functionality has been changed. It builds and renders properly. Internally within these folders I think we could still use some more restructuring, but I think this is a good step forward.


pkg/interface/src/views/apps/chat/components/lib/chat-header.js
```diff
@@ -3,7 +3,7 @@ import { Link } from "react-router-dom";
 
 import { ChatTabBar } from "./chat-tabbar";
 import { SidebarSwitcher } from "../../../../components/SidebarSwitch";
-import { deSig } from "../../../../lib/util";
+import { deSig } from "../../../../../logic/lib/util";
```
tylershuster, line 6: this is a bit deep in'it? maybe we should do something like this https://github.com/entwicklerstube/babel-plugin-root-import

tacryt-socryp: That would be a great quality of life change

tylershuster: I'll PR after this gets merged

matildepark:
I've honestly never heard of a 'logic' folder before, but I haven't exactly looked at a lot of codebases.

PR #2928, opened 2020-05-22
&publish-info parser cleanup

ohAitch:
On review, getting rid of all the `last-modified: ~` is not necessarily a pure improvement; going the other way, here's a sketch of a less idiomatic approach:

```
=,  userlib
=+  [t=(cook crip (star prn)) p=;~(pfix sig fed:ag) da=;~(pfix sig (cook year when:so))]
%-  parsf:scanf
[;"""
  owner: {p}
  title: {t}
  filename: {sym}
  comments: {(perk %open %closed %none ~)}
  allow-edit: {(perk %post %comment %all %none ~)}
  date-created: {da}
  """]
```

**Note:** submitting as a draft because this has not been tested including for syntax errors, the device I was running urbit on got water-damaged QQ 
(Also I don't know where to get copies of the files this is supposed to be parsing, at any rate)

PR #3311, opened 2020-08-14
dojo: restore `??` and `???` syntax

ohAitch:
Unsure what happened here, we appears to have ended up at a `?` and a `?????` and nothing in between, documented as "val/type/hoon/xray".

Grabbed original implementations out of git history, there was an ominous warning about "filtering" but I don't have the context for what it means - these operations cannot have side effects, so worst case you can just not run them. (Possibly after rebooting a hung pier?)

Contains some questionable type hackery around the 79aa1cb redefinition of `$?` to disallow multiple "complex" types, if anyone can come up with a real solution here I'll be overjoyed if you push commits to that effect.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -589,9 +589,7 @@
                   ?-  p.p.mad
                     %0  ~
                     %1  [[%rose [~ "  " ~] (skol p.q.cay) ~] maar]
-                    ::  XX  actually print something meaningful here
-                    ::
-                    %2  [[%rose [~ "  " ~] *tank ~] maar]
+                    %2  [[%rose [~ "  " ~] (dy-show-type-noun p.q.cay) ~] maar]
                     ::%3  handled above 
                     %4  ~
                     %5  [[%rose [~ "  " ~] (xskol p.q.cay) ~] maar]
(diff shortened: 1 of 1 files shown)


ohAitch:
Should I also be reformatting this to have commit message bodies cf #2928? This is mostly a glorified "uncomment some code" PR.

Fang-:
> Should I also be reformatting this to have commit message bodies

If you want to give more details, explanation or references (ie historical commits etc), that's always nice, but I wouldn't say they're mandatory.

ohAitch:
Makes sense, I think for historical record I'll just leave a note \~here on this PR~: most of the code reintroduced was either [commented out by](https://github.com/urbit/urbit/commit/79aa1cb#diff-95e3157683c97ac8eaa6db465e5c352fR1157-R1189) or otherwise [present as of](https://github.com/urbit/urbit/blob/79aa1cb/app/dojo.hoon#L1133-L1153) aforementioned `$?` commit 79aa1cb. (And got cleaned out in a separate later one.) There's also a small [patch](https://github.com/urbit/urbit/pull/3311/commits/dda3864) to account for thread syntax.

The part that could use eyes is a three-line [creative shim](https://github.com/urbit/urbit/pull/3311/files#diff-efaa874d5295aa406381fc8493f63577R663-R665) of bunting old $?, as (to simplify) `?(hoon @ta)` is no longer a valid mold.

philipcmonk:
I don't know why that fishes, but I confirm your code works and that `$?` fails as you say.  Good enough for this PR.

PR #3312, opened 2020-08-14
interface: made head reactive

tylershuster:
Made head reactive. Also took out the tracking of unseenTotal on each app because it didn't seem to do anything?

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -7700,6 +7700,24 @@
       "resolved": "https://registry.npmjs.org/react-fast-compare/-/react-fast-compare-2.0.4.tgz",
       "integrity": "sha512-suNP+J1VU1MWFKcyt7RtjiSWUjvidmQSlqu+eHslq+342xCbGTYmC0mEhPCOHxlW0CywylOC1u2DFAT+bv4dBw=="
     },
+    "react-helmet": {
+      "version": "6.1.0",
+      "resolved": "https://registry.npmjs.org/react-helmet/-/react-helmet-6.1.0.tgz",
+      "integrity": "sha512-4uMzEY9nlDlgxr61NL3XbKRy1hEkXmKNXhjbAIOVw5vcFrsdYbH2FEwcNyWvWinl103nXgzYNlns9ca+8kFiWw==",
+      "requires": {
+        "object-assign": "^4.1.1",
+        "prop-types": "^15.7.2",
+        "react-fast-compare": "^3.1.1",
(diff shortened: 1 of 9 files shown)


matildepark:
I still don't understand the benefit here. In general your PR should be clear about the issue, this particular solution, etc.

In general Landscape team has a stance of "if we don't need the dependency, don't add the dependency." So I'm curious as to what we get here going forward.

tylershuster:
The issue is that the head content is not a function of the ship state. It ships with index.html and doesn't update. 
The current solution is to update bits of the head piecemeal which is not how react works. It sucks that react doesn't ship with it but here we are.

The favicon PR made a little bit of the head reactive, but the whole thing should be, because it's a part of the interface. Here's what we get now without using `document.` writes:
- unread counts
- favicon

 Here's what we could make reactive to state/preferences in the future:
- app manifest including colors and ship name
- app name
- user zoom settings
- dynamic scripts/styles

matildepark:
(Big Conflict again, but I'd like general buy-in for this sort of change, regardless.)

ohAitch:
- Looking at the implementation, updating bits of the head piecemeal is invariably what `<Helmet/>` also ends up [doing](https://github.com/nfl/react-helmet/blob/master/src/HelmetUtils.js#L380-L423)
- Perhaps a less controversial change would be to pull out specifically the existing title-setting (and favicon-setting?) logic into one in-repo helper component, and separate the question of whether to replace it with a more featureful external one into a separate PR
- Downvote on `<React.Fragment></React.Fragment>` everywhere, just use `<></>`
- Not relevant to existing discussion, but if anyone is just tuning in the [whitespace-insensitive view](https://github.com/urbit/urbit/pull/3312/files?w=1) of the PR is much easier to follow

ohAitch:
The main benefit I can personally see to a declarative `<Title/>` component, besides consistency, is on unmount setting the title *back* to whatever it was before if it hasn't been rewritten to a third thing in the interim.

matildepark:
Resolve conflict? I'll review soon. I'm coming around to this, it looks cleaner.

tylershuster:
Will do, and will ditch the Fragments

liam-fitzgerald (approved):
having thought about it, I'm a fan of this change. The usecases for react-helmet seem justified and solve problems, especially around routing. Would appreciate further buy-in though.

matildepark:
Sorry for the new conflict...

matildepark:
Can you clean up the commits a bit? Will merge after.

tylershuster:
ahh fuck

PR #3313, opened 2020-08-14
interface: div is not a self-closing tag

tylershuster:
a bit pedantic but


ohAitch:
Really taking the X out of XHTML

matildepark:
I don't know what this means but I can't win here

tylershuster:
inde.xhtml

ohAitch:
Aight to be less cryptic: A+ correct change to achieve nonbroken behavior in the broken web ecosystem ^-^

PR #3314, opened 2020-08-14
interface: makes the embed button a button

tylershuster:
how to use indigo-react

### DIFF
--- a/pkg/interface/src/apps/chat/components/lib/content/url.js
+++ b/pkg/interface/src/apps/chat/components/lib/content/url.js
@@ -84,6 +84,7 @@ export default class UrlContent extends Component {
             border={1}
             borderRadius={2}
             display='inline-block'
+            fontSize={0}
             style={{ cursor: 'pointer' }}
             opacity={.8}
             px={2}
--- a/pkg/interface/src/apps/chat/components/lib/content/url.js
+++ b/pkg/interface/src/apps/chat/components/lib/content/url.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import { Box } from '@tlon/indigo-react';
 
 const IMAGE_REGEX = new RegExp(/(jpg|img|png|gif|tiff|jpeg|webp|webm|svg)$/i);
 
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
Try using Button instead of Box, despite the difference in visual style.

tylershuster:
I tried. I tried again this morning. The problem is that the button is display: flex and because of how it's written here:
https://github.com/urbit/indigo-react/blob/master/src/buttons/Button.tsx#L187
I can't seem to override that with class `dib`, or `display="inline-flex"` or anything. 

The leap button uses `<Box>` which is why I used it here.

matildepark:
Indigo-react doesn’t use atomic css, it used styled system props, so usually it would just be display=“inline-flex” but I believe the Button component doesn’t take display. I’m unsure if that’s changing next ver. Might be.
—
~haddef-sigwen
https://urbit.org

tylershuster:
Yeah, it doesn't take display. I'm not very good at reading styled-system yet but it *looks* like @g-a-v-i-n has already implemented it here: 

https://github.com/urbit/indigo-react/blob/ga/1.2/src/Button.tsx

matildepark:
The PR itself's visual change, for those of you at home: 
image

pkg/interface/src/apps/chat/components/lib/content/url.js
```diff
@@ -79,11 +80,19 @@ export default class UrlContent extends Component {
           >
             {content.url}
           </a>
-          <a className="bs ml2 f7 pointer lh-copy v-top"
-             onClick={e => this.unfoldEmbed()}
+          <Box
+            border={1}
+            borderRadius={2}
+            display='inline-block'
```
matildepark, line 86:
```suggestion
            display='inline-block'
            fontSize={0}
```

image

tylershuster: much better

matildepark:
Also want to note that you can pass `display: 'inline-block'` or the like as a `style` param like cursor; but it has automatic padding. It can break up the line-height rhythm of chat; it is seemingly meant as pure UI, not as part of copy.

image

tylershuster:
ok, the inline thing was kind of a red herring, because I tried the wrong prop. But the padding does foul up the rhythm of chat and the current implementation looks the best for now

matildepark:
Conflict now that the new organisation is in the branch, sorry. -_-;

matildepark:
Might as well take the opportunity to reword the commit above to fit the guidelines. (And this only affects chat, not all of interface).

tylershuster:
great! I'm going to close this and make a new PR when ready

PR #3318, opened 2020-08-14
interface: makes embed button a button

tylershuster:
**Problem**: a) The current "[embed]" next to Youtube URLs is not clearly a button and b) does not clearly describe its state

**Solution**: Make it a button using `indigo-react`'s styled component that reacts to state

**Caveat**: Due to active development on `indigo-react`, there is not a method that is both semantic and stylistically appropriate. This opts for semantics and documents where it is hacked

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/content/url.js
+++ b/pkg/interface/src/views/apps/chat/components/lib/content/url.js
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import { Button } from '@tlon/indigo-react';
 
 const IMAGE_REGEX = new RegExp(/(jpg|img|png|gif|tiff|jpeg|webp|webm|svg)$/i);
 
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
Instead of "stringly typed" values, please use a boolean. Otherwise, LGTM!

tylershuster:
I'd be happy to but I don't know what you're referring to

matildepark:
The `state.unfold` is a boolean.

tacryt-socryp:
Oh, I misinterpreted what was happening from the small diff Github was showing me. All good!

tylershuster:
Did I do something incorrectly?

matildepark:
I mean I'd like to see it since you changed the implementation, but I'll have to stash what I'm doing first.

tylershuster:
sec I'll send a screenshot

tylershuster:
preview

matildepark:
OK i'll merge after CI

PR #3319, opened 2020-08-14
interface: adds absolute path resolution to imports

tylershuster:
**Problem**: Deeply nested import structure often leads to lines like `../../../../../../views/components/component`

**Solution**: Use the `babel-plugin-root-import` plugin to resolve `~/` to the `src` directory.

### DIFF
--- a/pkg/interface/.babelrc
+++ b/pkg/interface/.babelrc
@@ -1,0 +1,14 @@
+{
+  "plugins": [
+    [
+      "babel-plugin-root-import",
+      {
+        "paths": [
+          {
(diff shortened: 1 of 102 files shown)


liam-fitzgerald:
you'll want to add [this](https://github.com/entwicklerstube/babel-plugin-root-import/issues/114) to tsconfig.json

tylershuster:
Just so I'm clear, what problem are you encountering? It seems to build fine for me.

tylershuster:
I added the lines but it doesn't change how it works for me

Issue #3177, opened 2020-07-24
king haskell's log file is spammy and will fill up the disk.


eglaysher:
I've been running king haskell for about 13 megabytes in ~20 minutes. Looking at the file, it looks like it's printing all the JSON being sent over the wire.

PR #3320, opened 2020-08-15
king: configure logging via cli flags

pilfer-pandex:
It took a surprising amount of iteration to get this right.

Addresses #3177.


pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -44,6 +46,19 @@ data Opts = Opts
...
   deriving (Show)
 
+-- | Options for the logging subsystem.
+data Log = Log
+    { lTarget  :: Maybe (LogTarget FilePath)
+    , lLevel   :: LogLevel
+    }
+  deriving (Show)
+
+data LogTarget a
```
eglaysher, line 56: Why is this `a` instead of just putting `FilePath` in `LogFile`? Are you planning on reusing this somewhere else?

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -494,15 +494,15 @@ router slog waitFx Drivers {..} = do
 
 logEvent :: HasLogFunc e => Ev -> RIO e ()
 logEvent ev = do
-  logTrace $ "<- " <> display (summarizeEvent ev)
```
eglaysher, line 497: I thought `logTrace` was supposed to be spammy than `logInfo`. But apparently not. It looks since like logTrace is a LevelOther, it's more important than errors. So yeah.

pkg/hs/urbit-king/lib/Urbit/King/App.hs
```diff
@@ -73,36 +76,50 @@ instance HasKingId KingEnv where
...
+
+withLogFileHandle :: FilePath -> (Handle -> IO a) -> IO a
+withLogFileHandle f act =
+  withFile f AppendMode $ \handle -> do
     hSetBuffering handle LineBuffering
     act handle
 
+defaultLogFile :: IO FilePath
+defaultLogFile = do
+  logDir <- getAppUserDataDirectory "urbit"
```
eglaysher, line 117: Consider `getXdgDirectorySource XdgCache "urbit"`. I was really surprised when it dumped log files in a new `~/.urbit/` originally.

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
pilfer-pandex: Yeah, actually I do, but it's kind of gross. in logTarget in Main.hs, I take a Maybe (LogTarget FilePath) to a LogTarget (Maybe FilePath). In the argument, Nothing encodes "log to the default location"; in the result, LogFile Nothing encodes "log to the default file (which you have to run an IO action to find)"

pkg/hs/urbit-king/lib/Urbit/King/App.hs
pilfer-pandex: I saw this Xdg business in the docs, but I didn't really understand it, so I passed over it. I'm happy to follow your suggestion though.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
pilfer-pandex: With this at LogOther "trace", there's actually no way to turn it off. Note that logInfo is almost unused in the king, and it sits between debug and warn. Since the full event printing (next line) is at debug, this is a good level for this printout.

pkg/hs/urbit-king/lib/Urbit/King/App.hs
pilfer-pandex: Okay I understand it now.

PR #3321, opened 2020-08-15
chat: account for digits in group name links

tylershuster:
**Problem**: When a group name has digits in it, it does not automatically link when sent on its own line in chat

**Solution**: Make the regex capture digits too


tylershuster:
Also multiple dashes (emojis, possibly)

PR #3322, opened 2020-08-15
leap: better handling of existing focus

### LINKED ISSUE #3297: Leap Search Interface is overridden by Dojo prompt when summoned in Dojo Module
**Describe the bug**
As mentioned in the issue title, I used the Leap interface to access Dojo, and when invoking it to go elsewhere in Landscape, my cursor was placed into the Dojo's input, even as the Search input was visible.

**To Reproduce**
1. Go to Dojo via any means (I used Leap)
2. Immediately attempt to use the Leap incantation
3. See error: Text insertion point should be in Dojo's prompt, but Leap should be summoned regardless. Attempting to type will result in text being input into the Dojo prompt.

**Expected behavior**
I expect the text insertion point/focus to be placed into the leap interface, rather than the Dojo's

Video:

https://www.dropbox.com/s/9f3f2rf1r6pouh2/dojo.mov?dl=0

tylershuster:
See #3297 and #3323. Provides expected behavior on leap invocation and cancellation, especially within the dojo. Returns the user to previous input when cancelled.

### DIFF
--- a/pkg/interface/src/logic/api/local.ts
+++ b/pkg/interface/src/logic/api/local.ts
@@ -28,7 +28,7 @@ export default class LocalApi extends BaseApi<StoreState> {
     });
   }
 
-    setOmnibox() {
+  setOmnibox() {
     this.store.handleEvent({
       data: {
         local: {
--- a/pkg/interface/src/logic/reducers/local.ts
+++ b/pkg/interface/src/logic/reducers/local.ts
@@ -3,7 +3,7 @@ import { StoreState } from '../../store/type';
 import { Cage } from '../../types/cage';
 import { LocalUpdate } from '../../types/local-update';
 
-type LocalState = Pick<StoreState, 'sidebarShown' | 'omniboxShown' | 'dark' | 'baseHash'>;
+type LocalState = Pick<StoreState, 'sidebarShown' | 'omniboxShown' | 'dark' | 'baseHash' | 'suspendedFocus'>;
 
 export default class LocalReducer<S extends LocalState> {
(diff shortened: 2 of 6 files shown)


matildepark:
Shouldn’t all text do this? Or is it that the initial propagation lets Dojo override?

tylershuster:
Oh, I actually forgot about that because I couldn't reproduce it, But I do think that's the case. @urcades what browser are you using?

tylershuster:
As it turns out, this seems to be a Safari-specific problem. I was getting a problem where the `/` of `cmd+/` got entered in dojo, and this does fix that. But Safari doesn't like giving focus if another element is focused.

urcades:
Ooo, this is wonderful! Thank you for addressing this general issue. Lmk if I can provide any other tests.

tylershuster:
Fixes #3323

tylershuster:
As it happens, CodeMirror components handle leap behavior properly. I don't know exactly why. Unless anyone objects, I'm going to reconfigure dojo/input.js to use the same component as chat-editor.js

matildepark:
I assume this PR is waiting on the Dojo change?

tylershuster:
Turns out  that was a red herring. The actual issue was inconsistent handling of "setSelectionRange" which was causing the dojo to always take focus. I think the whole experience is much more fluid and consistent now

matildepark:
Dojo's a scary place.

tylershuster:
Yeah, honestly I noped out of replacing the input with CodeMirror once I started to see what was happening. Turned out for the best.

PR #3324, opened 2020-08-15
chat: made channel-item link

tylershuster:
**Problem this solves**: Currently the `ChannelItem` in `chat` uses `div`s and `p`s to list the channels. This causes some problem for keyboard users and is not semantic.

**Expected behavior** (all the other apps function this way):


matildepark:
We use `<Link>` in the other sidebars. And quite shortly we’ll be making it a single component anyhow.

tylershuster:
Oh, true.

PR #3327, opened 2020-08-17
groups: join fixes

liam-fitzgerald:
- contact-hook: no-op on accepted invite

The contact-hook was attempting to join a group upon it's invite being
accepted. However, the join poke is also sent from the frontend, causing
a potential race condition. Changes contact-hook to no-op on an
%invite-update.

- group-store: disallow joining a non-existent group
- pull-hook: no-op on pulling a resource we already have

Should address issues with joining groups

### DIFF
--- a/pkg/arvo/app/contact-hook.hoon
+++ b/pkg/arvo/app/contact-hook.hoon
@@ -164,10 +164,7 @@
           (fact-group-update:cc wire !<(update:group-store q.cage.sign))
         [cards this]
       ::
-          %invite-update
-        =^  cards  state
-          (fact-invite-update:cc wire !<(invite-update q.cage.sign))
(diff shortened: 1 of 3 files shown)


liam-fitzgerald:
> These all seem like fine changes.
> 
> > The contact-hook was attempting to join a group upon it's invite being
> > accepted. However, the join poke is also sent from the frontend, causing
> > a potential race condition. Changes contact-hook to no-op on an
> > %invite-update.
> 
> I guess the reasoning for fixing this on the backend rather than frontend is that you want it to be robust against poorly implemented clients? Isn't the "join when we accept the invite" behavior still desirable?

I wanted to it this way, so that if for whatever reason the join errors out, the user still has an invite and can try again.

PR #3334, opened 2020-08-17
statusBar: rewrite in indigo, line-height tweak

matildepark:
Migrated the rest of the status bar layout to flexbox, indigo-react components. Set line height to zero and attempted to center all elements properly.

Screenshot 2020-08-17 at 12 26 36 PM
Screenshot 2020-08-17 at 12 26 47 PM

(Also patches rounded corners into Leap's results, too.)

### DIFF
--- a/pkg/interface/src/views/components/Omnibox.js
+++ b/pkg/interface/src/views/components/Omnibox.js
@@ -192,7 +192,13 @@ export class Omnibox extends Component {
 
   renderResults() {
     const { props, state } = this;
-    return <Box maxHeight="400px" overflowY="scroll" overflowX="hidden">
+    return <Box
+            maxHeight="400px"
+            overflowY="scroll"
+            overflowX="hidden"
+            borderBottomLeftRadius='2'
+            borderBottomRightRadius='2'
+           >
       {this.getSearchedCategories()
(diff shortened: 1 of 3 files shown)


urcades:
This is lovely! Vertical spacing is vastly improved.

A last thing we should consider here is either vertically centering the "Reconnecting" label, or aligning it with one of the corners.

urcades:
I'm checking up on this latter observation w/ design team, here's some options I'll be suggesting:

Screen Shot 2020-08-17 at 3 39 20 PM

matildepark:
So in this case we push Groups to left hand and put Settings on right hand + ship name?

—
~haddef-sigwen
https://urbit.org

urcades:
@matildepark I'm actually discussing that right now as well, I was going to write to ignore my actual placement here, and focus on the reconnecting alert positioning, but it might be good to address the general ordering of these areas themselves!

urcades:
I believe we have chosen this configuration:

Screen Shot 2020-08-17 at 3 44 41 PM

matildepark:
Mk, we'll have to incorporate this into the Settings PR too, then. I'll push Reconnecting to the left for now.

PR #3335, opened 2020-08-17
file-server: fix #3047, incorrect path matching

### LINKED ISSUE #3047: file-server: Landscape is mounted at all first-level directories regardless of other endpoints
In https://github.com/urbit/create-landscape-app/issues/42 a user reported that going straight to `/~myapp` gave a 404, but `/~myapp/index.html` worked, with malfunctioning behaviour. The latter was due to the react-router not anticipating the `index.html` of the URL. The former is importantly not an Eyre 404, but Landscape's:

image

Thus when you go direct to `/~myapp` it is serving Landscape's JS bundle — even if `/~myapp` was added to `%file-server`.

tacryt-socryp:


tacryt-socryp:
@Fang- Check out lines 156-160. Don’t you think Eyre should strip blank list items ‘’ from the paths it hands us?

tacryt-socryp:
@matildepark any discussion with mark about eyre semantics should not block the contents of this PR. Should be good to go!

matildepark:
I just like to wait for CI these days...

Fang-:
> @Fang- Check out lines 156-160. Don’t you think Eyre should strip blank list items ‘’ from the paths it hands us?

URL parsing was updated to have the current behavior precisely to be able to _retain_ blank list items. It can be important to tell the difference between `/some/page` and `/some/page/` (note the trailing `/`, resulting in the last list item being "blank"), the two are not equivalent.

matildepark:
@tacryt-socryp Might be an introduced bug from this PR: going to paths deeper than one-level (in this case, `~chat/room/~zod/dm—zod`) produce a Eyre-level 404.

PR #3339, opened 2020-08-18
urbit-king: update urbit-hob to v0.3.3

jtobin:
v0.3.2 included an important bug fix that should be pulled in here.


eglaysher:
But you also may want to target release/next-vere instead of master.

Issue #3338, opened 2020-08-18
Chat: email-style ">" blockquote markdown should be rendered as quotes


mike-cr:
Everyone seems to already use "\>" to indicate a quotation in chat, it'd be nice if this were rendered

> like this

and multi-line quotes 

\> like
\>
\> this

were rendered
> like
>
> this

Issue #2828, opened 2020-04-30
Chat markdown PR broke emoji rendering 😰


tacryt-socryp:

matildepark:
Can't reproduce

![image](https://user-images.githubusercontent.com/20846414/80673484-6055c400-8a7d-11ea-8746-86106a06442b.png)

loganallenc:
![image](https://user-images.githubusercontent.com/1377557/80676457-c7c34200-8a84-11ea-959f-4e27108a5f76.jpeg)

loganallenc:
On mobile

liam-fitzgerald:
codemirror doesn't seem to play nice with iOS. Hopefully when indigo-react components land, we can swap it out for the MarkdownEditor.

matildepark:
That’s ... also codemirror though

matildepark:
From [CodeMirror release notes](https://codemirror.net/doc/releases.html), as of 5.55.0 / June 21:

- The editor no longer overrides the rendering of zero-width joiners (allowing combined emoji to be shown).

Next release of Landscape should patch this since we `npm install` on release.

matildepark:
Unfortunately still an issue, so I was wrong about that potential remedy.

matildepark:
Fixed in #3774.

Issue #3325, opened 2020-08-16
chat: codeblocks should have a different background colour


mike-cr:
it'd be cool if backtick'd code had a different background colour `like this` to make it distinguishable

PR #3343, opened 2020-08-18
chat: several design tweaks

matildepark:
- Not the sidebar PR yet, sorry @galenwp.
- Adds blockquotes (#3338). @urcades?

Screen Shot 2020-08-18 at 5 20 37 PM

- Code mode swaps placeholder and text with Source Code Pro.
- Add code messages now use Source Code Pro instead of being reset to Consolas.
- Updated codemirror to 5.55.0 to test my hypothesis from #2828.

### DIFF
--- a/pkg/interface/package.json
+++ b/pkg/interface/package.json
@@ -11,7 +11,7 @@
     "@tlon/indigo-light": "^1.0.3",
     "@tlon/indigo-react": "^1.1.15",
     "classnames": "^2.2.6",
-    "codemirror": "^5.51.0",
+    "codemirror": "^5.55.0",
     "css-loader": "^3.5.3",
     "formik": "^2.1.4",
     "lodash": "^4.17.15",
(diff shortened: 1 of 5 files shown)


urcades:
Ah, proper quoting is a nice addition!

matildepark:
Threw #3325 into here. 

image

liam-fitzgerald (approved):
LGTM!

tylershuster:
This is going to make Urbit Dialogues look like this:

quotes

urcades:
whoa ... We'll need another demarcation, huh.

A thought here, Inter natively allows for glyph-formation in a really cool
way: we could transition Dialogue turn-passing with `->`, which turns into
a beautiful arrow.

—
~fabled-faster
https://urbit.org/

matildepark:
Do Urbit Dialogues do `>>` or `> > >`? Why does the markdown renderer process `>>>` as three blockquotes? (Was @Fang- right all along?!)

urcades:
I believe Dialogues uses ">>>", no spaces.

I'll suggest to them we use → , since this is processed by inter in a pleasing manner.

—
~fabled-faster
https://urbit.org/

Issue #3282, opened 2020-08-11
Long group names render behind the "copy" prompt in the "share" box in "group settings


sivner-figbus:
![image](https://user-images.githubusercontent.com/67953485/89905884-93623480-dbe2-11ea-9b6a-9235b873c165.png)

**To Reproduce**
1. Create a group with a long name 
2. Click on 'groups'
3. Click on the name of the group in the list at the left of the screen.
4. Click on "group settings"

**Expected behavior**
The text/prompt in the "share a shortcode to join this group" box should not be rendered over the top of text.

**Desktop (please complete the following information):**
 - OS: Mac OS X 10.15.5
 - Browser Version 1.11.104 Chromium: 84.0.4147.105 (Official Build) (64-bit)
 - Same in Safari Version 13.1.1 (15609.2.9.1.2)
 - Base hash of your urbit ship. 0v1i.s1lmp.1uiuf.e8mtd.mu0lr.m54qh.b9gm3.qa7im.pemgd.3uh25.dg0qj

PR #3346, opened 2020-08-18
groups: pad shortcode prompt to avoid graphic bug

matildepark:
See #3282. Also refactors shortcode render into a function.

Screen Shot 2020-08-18 at 5 59 28 PM


liam-fitzgerald (approved):
LGTM!

Issue #3326, opened 2020-08-16
Chat: Clicking on a participant's name no longer gives feedback that it's been copied to clipboard


mike-cr:
**Describe the bug**
In chat, if you click on `~tinnus-napbus` it copies `tinnus-napbus` to the clipboard, but there's absolutely no feedback that it's done so. I believe it used to say "copied" or something but it doesn't any more.

**To Reproduce**
1. Go to chat
2. Click on someone's name
3. Note nothing seems to happen
4. Paste and see it did in fact copy

**Expected behavior**
It should give you some kind of feedback that the ship name has been copied to clipboard

**Desktop (please complete the following information):**
 - OS: archlinux 64bit
 - Browser qutebrowser & firefox
 - Base hash: ah5b8

PR #3347, opened 2020-08-18
chat: restore 'copied' notice on copying names

matildepark:
See #3326. Now in a nice, stateless flavour.


matildepark:
ping @tacryt-socryp

tacryt-socryp:
Must check if ref is null before changing innerText both in the initial closure and in the setTimeout closure.

matildepark:
> Also must set unique key on message in this div so that we don't have the potential for extraneous Copied messages.

Wait, where am I setting `key={props.key}`? The message component itself has a key passed to it from chat-message.

matildepark:
ping @tacryt-socryp

tacryt-socryp:
More explicitly: check if nameSpan is null in the closure and this will be good to go

matildepark:
Aight I think I did. It functions normally.

matildepark:
can I get a stamp here? @tacryt-socryp @liam-fitzgerald

liam-fitzgerald (approved):
LGTM!

Issue #3211, opened 2020-07-28
soto: renders offscreen on iOS safari


matildepark:
Hard to replicate unless you test on an iPhone — in local environments, regardless of browser, it appears to calculate its height correctly; on an iPhone it clearly oversizes the viewport such that you can never see the input. I had a viable solution by giving mobile `calc(100% - 110px)` on the main container instead of the current `calc(100% - 45px)` but it of course renders poorly in local environments, and likely is just about wrangling flexbox to get the correct height of the viewport.

matildepark:
@tylershuster Do you know offhand about this one? I think it could possibly be because Safari autofocuses and it changes the height, pushing it offscreen? But it's only testable on the device; everything works fine in the web environment.

matildepark:
Fixed in #3348.

PR #3348, opened 2020-08-19
dojo: fixes whitescreen on mobile

tylershuster:
DWISOTT. 

Does this by affecting the whole app, though. IMO not a bad solution as it prevents the need for `calc(100vh - $headerheight)` hacks in the future.

Also hides the ship name on mobile to save space.

Also learned how to serve on my local network for phone testing so that's cool.


tylershuster:
#3211

PR #3350, opened 2020-08-19
natpmp: change nat-pmp startup behaviour.

eglaysher:
- Only tries to perform port forwarding if we detect we're on a private network, with explicit always try
  and never try command line flags.

- Fixes problems with the library we use where it could wait up to 4 minutes for a timeout. 

- Fixes a problem in the binding layer where if we failed to get a public address, we'd block Ames
  from ever successfully sending a packet.


eglaysher:
I've tested this on my own AWS instance running `~littel-ponnys`. Without an explicit --no-port-forwarding, this does add a 2 second startup delay before it times out on trying to find a router, but I believe the way to handle this is for our infrastructure scripts to pass that argument to king, since the gains for home users are so big when you have direct routes.

eglaysher:
(Q: Now that this works on cloud instances, should I add 172.16.x.x/12 to include all private networks?)

pilfer-pandex:
And sorry about the mix-up last night.

PR #3353, opened 2020-08-20
king: upgrade king haskell to ghc 8.8.3

brendanhay:
Given the GHC and stack resolver in use were coming up on 2~ years old, along with a plethora of stale dependencies - this PR upgrades `pkg/hs/stack.yaml` to the latest stackage LTS which is based on GHC `8.8.3`. (Ideally we'd move to `8.10.2` directly - but that proved a bridge too far, for now.)

There are two changes + possibly still open questions:

* The use of `Control.Monad.Fail.fail` within the `Web3` monad - for which there is no `MonadFail` instance. This primarily rears its head in `Urbit.Vere.Dawn` which previously used `fail` to signal errors. Since this simply specialised to `IO` / `userError` - I've replaced it as such.
* The `network` package long ago deprecated `Network.Socket.inet_addr` - it was bemusing to see the "workaround" for this in `Urbit.Vere.Eyre.Serve` was to set `-Wno-deprecations`? I've replaced it with `getNameInfo` - but this will probably need the mosting eyes/testing of all the changes.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
@@ -16,8 +16,6 @@
   TODO How to detect socket closed during server run?
 -}
 
-{-# OPTIONS_GHC -Wno-deprecations #-}
-
 module Urbit.Vere.Eyre.Serv
   ( ServApi(..)
   , TlsConfig(..)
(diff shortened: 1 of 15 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
```diff
@@ -140,8 +138,12 @@ openFreePort hos = do
       Right ps -> pure (Right ps)
  where
   doBind sok = do
-    adr <- Net.inet_addr hos
-    Net.bind sok (Net.SockAddrInet Net.defaultPort adr)
+    adr <-
+      Net.getAddrInfo (Just Net.defaultHints) (Just hos) Nothing >>= \case
+        [] -> error ("unable to determine numeric hostname from " ++ hos)
+        x:_ -> pure (Net.addrAddress x)
```
eglaysher, line 144: This will probably work, but did you manually test that it does?

brendanhay: In the process of going through various tests on Linux, hold off from any kind of merge until verified.

botter-nidnul:
Excellent news for king haskell on aarch64. 🙏

eglaysher:
Also, the travis jobs are timing out. I suspect you'll need to do something regarding having cachable prebuilts of the stack dependencies, but I don't understand how that works.

brendanhay:
I'll take care of it.

pilfer-pandex:
Why are we giving up on this??

brendanhay:
Work continues in #3374.

pilfer-pandex:
Yeah, I saw that *after* I commented. Sorry for frivolously pinging. :(

PR #3354, opened 2020-08-20
natpmp: when reopening laptop, reestablish a connection to the router.

eglaysher:
If the user closes their laptop lid, they may lose their connection to the router. Handle the case where that connection has gone away.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Ports.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Ports.hs
@@ -69,6 +69,14 @@ portLeaseLifetime = 15 * 60
 portRenewalTime :: Word32
 portRenewalTime = portLeaseLifetime - 60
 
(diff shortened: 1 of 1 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Ports.hs
```diff
@@ -69,6 +69,14 @@ portLeaseLifetime = 15 * 60
 portRenewalTime :: Word32
 portRenewalTime = portLeaseLifetime - 60
 
+-- Number of retries before we give up on performing nat operations.
+maxRetries :: Int
+maxRetries = 3
+
+-- How long to wait between retries.
+networkRetryDelay :: Int
+networkRetryDelay = 5 * 1000000
```
pilfer-pandex, line 78: if you like you can put _s in the number to make its size more clear? Looks like it's 10e6?

PR #3355, opened 2020-08-20
+roll and +reel stack space

frodwith:
The jet for +roll was tail recursive, but +reel for sure consumed stack space to iterate backwards over the input list. Both jets have been made into while loops, and +reel uses a vector to remember the list instead of the call stack.

### DIFF
--- a/pkg/urbit/jets/b/reel.c
+++ b/pkg/urbit/jets/b/reel.c
@@ -10,14 +10,11 @@
             u3_noun b)
   {
     u3_noun pro = u3k(u3x_at(u3x_sam_3, b));
-    if ( u3_nul == a ) {
-      return pro;
-    }
-    else {
+    if ( u3_nul != a ) {
       u3j_site sit_u;
       u3_noun  i, t = a;
       c3_w     j_w, len_w = 0, all_w = 89, pre_w = 55;
-      u3_noun* vec = u3a_malloc(all_w);
+      u3_noun* vec = u3a_malloc(all_w * sizeof(u3_noun));
 
       // stuff list into an array
       do {
@@ -30,7 +27,7 @@
             // grow vec fib-wise
             all_w += pre_w;
             pre_w = len_w;
-            vec = u3a_realloc(vec, all_w);
+            vec = u3a_realloc(vec, all_w * sizeof(u3_noun));
           }
           vec[len_w++] = i;
         }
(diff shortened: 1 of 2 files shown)


ohAitch:
I'm not exactly fluent in C / perf, but would it make any sense to do a first pass over the linked list to just measure its length, vs the fibonacci in-place growth pattern?

frodwith:
You'd have to traverse it twice, then.

ohAitch:
Right, how bad is it to chase a pointer you just did(which is probably still in cache? unless it's big?) vs reallocating more space a bunch of times(which is probably in the same place? unless it's big?)
I suppose your meaning is "bad" :P

frodwith:
The only real way to find out is to benchmark, I suppose, but traversing a list is O(n) pointer chases, and the reallocing happens a lot less than O(n) (some coefficient involving the golden mean, i suppose), so I think the fib-growth wins for large inputs.

pkg/urbit/jets/b/reel.c
```diff
@@ -3,34 +3,44 @@
...
+      u3_noun  i, t = a;
+      c3_w     j_w, len_w = 0, all_w = 89, pre_w = 55;
+      u3_noun* vec = u3a_malloc(all_w * sizeof(u3_noun));
+
+      // stuff list into an array
+      do {
+        if ( c3n == u3r_cell(t, &i, &t) ) {
+          u3a_free(vec);
+          return u3m_bail(c3__exit);
+        }
```
joemfb, line 24: I think this can just be `u3x_cell()` -- we don't need to clean up our allocations if we bail. And this code already depends on that being the case, consider the refcount of `pro`.

pkg/urbit/jets/b/reel.c
```diff
@@ -3,34 +3,44 @@
...
+          }
+          vec[len_w++] = i;
+        }
+      } while ( u3_nul != t );
+
+      // now we can iterate backwards
+      u3j_gate_prep(&sit_u, u3k(b));
+      for ( j_w = len_w; j_w > 0; ) {
+        pro = u3j_gate_slam(&sit_u, u3nc(u3k(vec[--j_w]), pro));
+      }
```
joemfb, line 40: Nitpicky, but I think this'd be clearer as `while ( len_w > 0 )` and `vec[--len_w]` ... The author decides these things, of course.

joemfb:
Regardless of the approach, avoiding intermediate cell allocation in `+reel` seems like a win. However it's done there should (eventually) be replicated in the jets for `+turn`, `+murn`, the relevant treap operations, &c.

Issue #3332, opened 2020-08-17
Spacing between last message and chat input box is too tight


galenwp:
**Describe the bug**
The spacing between the last chat message and the chat input box is too small

**To Reproduce**
1. Go to Chat, select a channel
2. Notice the spacing between the last message and the input box

**Expected behavior**
More spacing

**Screenshots**
The spacing when entering a channel is like this:
<img width="199" alt="Screen Shot 2020-08-17 at 8 57 03 AM" src="https://user-images.githubusercontent.com/1467211/90416845-d8b1b680-e067-11ea-8363-ab1e6f1f8311.png">

But if you scroll down a bit it's like this:
<img width="227" alt="Screen Shot 2020-08-17 at 8 57 14 AM" src="https://user-images.githubusercontent.com/1467211/90416865-e0715b00-e067-11ea-9848-5d49d74c5fbe.png">

The second one is correct. I find myself scrolling down intentionally often to make it look right.

tylershuster:
`pb2` is the problem here:
https://github.com/urbit/urbit/blob/b0d252fa7633ba030c555f690594d85c0d59b36f/pkg/interface/src/apps/chat/components/lib/chat-scroll-container.js#L104

matildepark:
Well no padding would still have the zero spacing as displayed. I think we want, what, mt2 on the chat input?

—
~haddef-sigwen
https://urbit.org

tylershuster:
I was just pointing out where the gap comes from. mt2 works. I can PR unless you're already on it

tylershuster:
actually mt2 leaves a gap on top of the chatbox so there's a white bar between it and the text as you scroll. Except you can actually fix it in FireFox because it has the wrapper. Lol. Normally I would do like `.chat:last-of-type { padding-bottom: .25rem }` but that's not really possible.

matildepark:
Ironically I believe here it’s the first child?

— 
~haddef-sigwen
https://urbit.org

matildepark:
But yeah you don’t want to remove the padding though. It’s that the autoscroll now doesn’t pin the container, but focuses the chat message element?

— 
~haddef-sigwen
https://urbit.org

tylershuster:
Right, fist-child but in any case. The autoscroll basically does "scrollintoview" which works because the entire message is now visible, but there's padding on the bottom of the wrapper element. It's not trying to scroll the *wrapper element* into view, so it ignores the `pb2` at its bottom.

[...]

PR #3357, opened 2020-08-20
interface: changed virtual scroll implementation and added to chat

tylershuster:
This PR changes the virtualization component to react-virtuoso, which has the advantage of calculating layout instead of needing the height to be set explicitly.

This makes it possible to virtualize the chat scroll.

The features I'm aware of are at least at parity in this PR, if not better. It no longer uses flow-reverse-column and mostly does away with the idea of pages (see caveats below). I deleted some concepts and renamed others because *I think* they were artifacts of the history of this component.

There is a known issue where (on my ship) messages were out of order with their `number` (as in, I could dbug chat-store and the messages would not have the same number as if I loaded the room on a comet. urbit dialogues specifically) and for whatever reason would load properly in the old implementation but not in the new. I would appreciate some insight there.

I would appreciate a review from @tacryt-socryp, @g-a-v-i-n, @liam-fitzgerald, and/or @matildepark when you get some time. Any comments would be appreciated. I'm putting this in draft because I know it needs some more work to really be polished but I'd like some feedback before moving forward.


matildepark:
Wow, big changes here.

>  It no longer uses flow-reverse-column and mostly does away with the idea of pages (see caveats below).

Cheers, will peruse.

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -1,25 +1,64 @@
...
+import { Envelope, IMessage } from "~/types/chat-update";
+import { RouteComponentProps } from "react-router-dom";
+import { Patp, Path } from "~/types/noun";
+import { Contacts } from "~/types/contact-update";
+import { Association } from "~/types/metadata-update";
+import { Group } from "~/types/group-update";
+import GlobalApi from "~/logic/api/global";
+
+// This component gets passed an envelopes prop with 20 envelopes
+// Why? No one knows.
```
liam-fitzgerald, line 17: this is because the initial chat-view subscription returns 20 envelopes for all chats. More are requested when the chat is selected

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -1,25 +1,64 @@
...
+  station: any;
+  api: GlobalApi;
+}
+
+interface ChatWindowState {
+  numPages: number;
+}
+
+export class ChatWindow extends Component<ChatWindowProps, ChatWindowState> {
+  private fetchPending: boolean;
```
liam-fitzgerald, line 51: if you're declaring these members, you can also initialise them here and save you from doing it in the constructor. (It will also infer the type)

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -126,68 +142,104 @@ export class ChatWindow extends Component {
...
+    // can't reach the bottom, so just don't set the initial
+    // index if fewer than 20 messages exist in the chat
+    const initialIndex =
+      mailboxSize - 1 < INITIAL_LOAD
+      ? undefined
+      : unreadCount // otherwise if there are unread messages
+        ? mailboxSize - unreadCount - 1 // put the one right before at the top
+        : mailboxSize - 1; // Just load it at the end
+    
+    const overscan = 100;
```
liam-fitzgerald, line 202: should probably be declared outside of the function scope as an all caps constant.

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -126,68 +142,104 @@ export class ChatWindow extends Component {
...
-                previousMsg={messages[i - 1]}
-                nextMsg={messages[i + 1]}
-                association={props.association}
-                group={props.group}
-                contacts={props.contacts} />
-            ))
-          }
-        </ChatScrollContainer>
+        <BacklogElement isChatLoading={isChatLoading} />
+        <ResubscribeElement {...{ api, host: ship, station, isChatUnsynced}} />
```
liam-fitzgerald, line 210: woah I didn't realise you could do this

liam-fitzgerald (approved):
Thanks for this! Looks good to me, but a review from @tacryt-socryp would be nice as he just refactored some of this.

matildepark:
Does this load the entire room? UC General was blank in testing for quite a while ... before it suddenly jumped 1,300 messages up. The entire scrollback does seem to be there...

matildepark:
It's pretty nice. The cutoff for unloading seems to be like 75% of the actual chat window height, weirdly. And yes, we need @tacryt-socryp to scrutinise this...

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
tylershuster: Yeah, it's the same feature as 
const { props } = this

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
tylershuster: Oh ok, so this is an arvo feature not a landscape one

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
tylershuster: Sorry, which are you referring to? Point me in the direction of a component with similar syntax if you don't mind.

tylershuster:
@matildepark It doesn't load the entire room. Basically it gets given the initial 20, fetches the `DEFAULT_BACKLOG_SIZE`, and then when you scroll up there's a few things that happen. One, if you're within `DEFAULT_BACKLOG_SIZE` of the top OR by some twist of fate you've actually *reached* the top, it loads `DEFEAULT_BACKLOG_SIZE` messages.

Again, there is some weirdness with message ordering but it seems to be a glitch in my ship. If anyone knows, I've asked tacryt about using the `number` property of the message once `graph-store` launches, because that will make this bulletproof.

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
liam-fitzgerald: `private fetchPending = false;`

matildepark:
pong @tacryt-socryp 

(Also conflict @tylershuster)

tacryt-socryp:
Strangely enough, in Safari, when you scroll up using the mouse, the entire page doesn't render at all.
Screen Shot 2020-09-02 at 1 31 45 PM

tacryt-socryp:
Another strange artifact where the page partially renders:
Screen Shot 2020-09-02 at 1 32 59 PM

tacryt-socryp:
A bit strange. I'm unsure of whether this is polished enough to go out, though it seems due to the library's behavior, not the implementation. This seems like a product decision to me, not a technical one. The code style is good, but if it does go in, please remove the comments.

tacryt-socryp:
@urcades @matildepark Up to you.

tylershuster:
I agree it's not ready to go out. I will check it mor thoroughly in other browsers. If you think the code style is acceptable I will probably just make small tweaks to get it to work.

matildepark:
Ping when it's more polished and we'll take another look.

urcades:
I'll also provide feedback when item is updated.

tylershuster:
Ok, turns out I was fetching messages incorrectly. This seems to work very nicely. Check out the [video here](https://marpem-files.sfo2.digitaloceanspaces.com/Scrolling.mov).

@urcades you can see some of the behavior I mentioned here. The library allows for scroll placeholders, and I took a rough stab at them here, drawing sigils every five messages and pseudo-randomly setting the length of the message elements. It's a new design pattern so I would appreciate your input.

I would also appreciate some insight on how you would prefer scroll slamming and idleness to works. It can now load and render at an arbitrary point in the list (I tested it with 300+ unread messages and it always loaded at the correct point), but I don't know expected/desired behavior around what to do when a new message is added.

There's a little to clean up and some conflicts to resolve, but I'm much happier with the way this works.

matildepark:
@tylershuster @urcades is out for a week (but I've seen him post around anyway) so I'm happy to handle in his absence.

> I would also appreciate some insight on how you would prefer scroll slamming and idleness to works. It can now load and render at an arbitrary point in the list (I tested it with 300+ unread messages and it always loaded at the correct point), but I don't know expected/desired behavior around what to do when a new message is added.

See [previous](https://github.com/urbit/urbit/pull/2825#issuecomment-621296359) [conversations](https://github.com/urbit/urbit/issues/2864#issuecomment-624736105) on this matter (since May?...). Basically we only want to pin us to the bottom if we're already at / near the bottom (1-2 messages, 10% of the viewport, whatever), or if we send a message.

Will test...

matildepark:
> @matildepark It doesn't load the entire room. Basically it gets given the initial 20, fetches the `DEFAULT_BACKLOG_SIZE`, and then when you scroll up there's a few things that happen. One, if you're within `DEFAULT_BACKLOG_SIZE` of the top OR by some twist of fate you've actually _reached_ the top, it loads `DEFEAULT_BACKLOG_SIZE` messages.
> 
> Again, there is some weirdness with message ordering but it seems to be a glitch in my ship. If anyone knows, I've asked tacryt about using the `number` property of the message once `graph-store` launches, because that will make this bulletproof.

Are you *sure*? In Safari you're sitting on a blank screen for about ~30 seconds for General on UC, the scrollbar is *long* and you can scroll all the way to April 7 without any wait.

matildepark:
This screen blanking seems a lil aggressive also, it makes me lose track of where I was when I scrolled up. Seemingly fires at the smallest scroll up. 
Sep-06-2020 19-01-50

tylershuster:
@matildepark regarding the initial load: Loading a room I already have been in before loads instantly. The scrollbar is long as a *feature* — it knows the `mailboxSize` so it guesses how long based on the elements it has. It updates when it finds them. I just subscribed to a new room and it took forever to join, but the `fetchMessages` method didn't even fire. Then I loaded it in a new browser (checked Safari too) and stuff was fast. Can you give me some more details? If you use React DevTools you can inspect what a given component has. Scrolling really is wicked fast with this implementation.

As for the scrolling, I agree and that should be easy to change.

Thanks for the feedback on slamming. I think that's pretty easy to implement

matildepark:
Idk, it's simple as "I open Safari to localhost:9000 and click General". [Vid](https://drive.google.com/file/d/1c28_rlAemratIwk2DtycMf3tGdBMv4Fq/view?usp=sharing). We just want it to show the latest 20 instead of making us wait for the whole thing. Idk what else it's doing; rooms with less history take less time.

matildepark:
Also want to note that we seemingly never fire the 'read' action when we hit bottom. Even clicking 'mark as read' does nothing?

tylershuster:
Oh f that is debugging code.

tylershuster:
I'll have to investigate the loading behavior.

tylershuster:
You were right — partially because of some bad math on my part, and partially because the reducer didn't dedupe messages after loading, so you could end up with as many copies of the same message as times you had loaded the channel in a session.

Working on slamming behavior now.

tylershuster:
Alright, racking up the conflicts but the implementation is much smoother now thanks to a) fixing the way that chat-update reduces messages and b) debouncing fetches so it doesn't try and grab new messages until you stop scrolling

tylershuster:
Also added a new feature where you can click the unread notice to scroll to that position

pkg/interface/src/logic/reducers/chat-update.ts
```diff
@@ -49,8 +50,11 @@ export default class ChatReducer<S extends ChatState> {
   messages(json: ChatUpdate, state: S) {
     const data = _.get(json, 'messages', false);
     if (data) {
-      state.inbox[data.path].envelopes =
-        state.inbox[data.path].envelopes.concat(data.envelopes);
+      state.inbox[data.path].envelopes = _.unionBy(
```
tylershuster, line 53: This prevents dupes

pkg/interface/src/logic/api/chat.ts
```diff
@@ -12,7 +12,7 @@ export default class ChatApi extends BaseApi<StoreState> {
    * Fetch backlog
    */
   fetchMessages(start: number, end: number, path: Path) {
-    fetch(`/chat-view/paginate/${start}/${end}${path}`)
+    return fetch(`/chat-view/paginate/${start}/${end}${path}`)
```
tylershuster, line 15: Allows for post-fetch behavior

pkg/interface/src/views/apps/chat/components/lib/chat-message.tsx
```diff
@@ -2,16 +2,15 @@ import React, { PureComponent, Fragment } from "react";
...
+import { Envelope } from "~/types/chat-update";
+import _ from "lodash";
 
 export const ChatMessage = (props) => {
   const {
     msg,
     previousMsg,
     nextMsg,
-    isLastUnread,
+    isFirstUnread,
```
tylershuster, line 13: This is because we render in the correct order now

pkg/interface/src/views/components/RemoteContent.tsx
```diff
@@ -116,7 +116,10 @@ export default class RemoteContent extends Component<RemoteContentProps, RemoteC
         </>
       );
     } else if (isOembed && remoteContentPolicy.oembedShown) {
-      this.loadOembed();
+      if (!this.state.embed) {
```
tylershuster, line 119: Fixes an uncaught bug where it would just fetch and fetch and fetch

tylershuster:
Cleaned up; behaves very nicely now

matildepark:
Good stuff, will test.

matildepark:
- The placeholder needs a gentler gray, basically, it needs to get out of the way, it flashes quick and it's tolerable if it's barely noticeable
- Chat doesn't seem to keep us pinned at bottom if a new message comes in and we're at the very bottom? It shows the "1 new message" but tbh I wasn't in focus, I have to send the incoming message in another window, so possibly it's working as intended if it's based on focus

Otherwise, yeah a total improvement imo

tylershuster:
Hm, good points. I'll clean up the placeholder and account for dark mode after the sun sets. The pinning worked last time I checked; maybe I changed something.

I don't think I changed the padding, but maybe reversing the scroll direction did something.

matildepark:
It also does a clean flash and load of messages on navigating to Settings or per room (as it doesnt in livenet). I wonder how to prevent a re-render if we don't need to.

liam-fitzgerald:
Just tried this! This is mad nice. I have no qualms with it, other than the flash of message that occurs when i nav to a channel.

tylershuster:
Alright, I fixed the padding thing (also fixes #3332). The flash is unavoidable, I think. The one downside of not using the flex-reverse implementation. Though I did reduce its severity.

matildepark (approved):
I'm content. If anything is left here we can sort out in the next few days.

PR #3359, opened 2020-08-21
chat: updated Leave Group copy

tylershuster:
The current "Leave Group" copy says "You will need to request access again" which after the groups refactor is not accurate. Changing that is the primary purpose of this PR.

I'm also wondering about two things: one of which I changed, the other I didn't.

1. `React.PureComponent`: I don't have a lot of experience with these but it seems like it might be appropriate here. If here, than quite a few other places. I've been in the habit of using functional components but *they say* these have performance benefits.

2. The other is maybe an @urcades question: Why do we gray out the unusable buttons instead of just removing them? Seems like in especially this case you wouldn't want to have both "Leave Group" and "Delete Group" on the same screen, ever. I have used the settings page to reference a chat description before but it's hard to read anyway. Maybe we should just add the description and get rid of the text box. Off-topic but RFC.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/delete-button.js
+++ b/pkg/interface/src/views/apps/chat/components/lib/delete-button.js
@@ -1,51 +1,54 @@
-import React, { Component } from 'react';
+import React, { memo } from 'react';
 
+export const DeleteButton = memo(({ isOwner, station, changeLoading, association, contacts, api }) => {
+  const leaveButtonClasses = (!isOwner) ? 'pointer' : 'c-default';
+  const deleteButtonClasses = (isOwner) ? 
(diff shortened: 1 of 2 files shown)


matildepark:
pong @tacryt-socryp

matildepark:
> Why do we gray out the unusable buttons instead of just removing them?

IIRC the stance at the time was that it informally shows that we're a member, not an admin; and what the owner has them set to? I don't know. I have seen that pattern before, can't remember where.

> I have used the settings page to reference a chat description before but it's hard to read anyway. Maybe we should just add the description and get rid of the text box.

"Surface chat description" below the title was always on the table, just hasn't been done. I should add that to the list ...

Issue #3363, opened 2020-08-24
links should support posting via S3


liam-fitzgerald:
Links should support posting via S3, e.g. I upload something and it's stored in my S3 bucket and link is posted to the collection.

tylershuster:
Do you think this is like the whole collection becomes a dropzone where you can drag a file in?

liam-fitzgerald:
Just a button a la current chat implementation would be fine for now

PR #3367, opened 2020-08-24
links: added s3 integration

tylershuster:
Click the button; upload the files.

Also deduped the `<S3Upload />` component.

#3363


tylershuster:
Oops, this should probably use a different icon...

tylershuster:
Progress bar would be nice too

tylershuster:
@urcades do you have a generic "file upload" icon?

matildepark:
@tylershuster Try picking one from [here](https://github.com/urbit/indigo-react/blob/master/src/Icon/iconIndex.tsx) and using the `<Icon>` indigo-react component (though note that not all of them are in 1.1, which we have in use here. We'll move to 1.2 after some of Liam's PRs are in...)

urcades:
@tylershuster yes, this is an Indigo thing — if you note a lack of icons available, feel free to make a request/suggestion and they can be worked on.

tylershuster:
@urcades As far as I can tell, the existing Image icon isn't in here — and a generic "File" icon would be groovy as well. I can swing it with "ArrowNorth" but expanding the S3 connector beyond images means that it should have some other file types.

Part of the work I'm doing here involves a) deduplicating the S3Upload component and b) adding some new parameters to it to account for context, such as `size` and `accept` and forthcoming `icon`

matildepark:
Interesting. We don't want to remove the image prohibition from Chat yet because we need media previews for what we let you upload. So basically, if a URL is a vid then a nice lil vid preview, etc.

matildepark:
That said I think ArrowNorth is fine for now? But we use that "image" icon for all other instances, even in settings screen PR I think.

tylershuster:
Yeah, I added an `icon` prop but the current problem is that the image icon is an image

tylershuster:
And an `accept` prop

urcades:
cc @g-a-v-i-n : any thoughts on the suggested icon additions mentioned above? I think we had a catch-all file icon here:

Screen Shot 2020-08-25 at 9 24 26 AM

Unsure if you witnessed this in the repo @tylershuster — as a placeholder, the above might work fine (it'd actually be great to get a screenshot of the context you're working in), but I see what you mean re: cooking up some specialized icons for cases like this.

tylershuster:
I didn't see those in the repo. Check this out:
[Video](https://marpem-files.sfo2.digitaloceanspaces.com/Screen%20Recording%202020-08-25%20at%206.31.13%20AM.mov)

urcades:
@tylershuster this reminded me that we had some nascent indigo components that afford for a form/input of this type, at least partially:

Screen Shot 2020-08-25 at 10 44 48 AM

You can imagine how we might conjoin fields to build something akin to what you are updating/improving here. It looks like Gav has specified that we use a text-button in these fields, which seems good. This update still prompts us to have icons for this sort of field/action though.

g-a-v-i-n:
I would punt on the icon and just use <Action /> with descriptive text like 'Upload File'.

tylershuster:
Ok check this out. `S3Upload` now takes `children`. Otherwise defaults to an up arrow.

tylershuster:
For those of you at home:

[video](https://marpem-files.sfo2.digitaloceanspaces.com/Screen%20Recording%202020-08-25%20at%202.10.51%20PM.mov)

matildepark:
... oh it was targeting `settings-screen` for some reason.

tylershuster:
Yeah. I asked @liam-fitzgerald and he agreed because of the s3 changes he had made

liam-fitzgerald:
settings screen is now in release/next-userspace, so it should be able to go in there.

matildepark:
I can't seem to reopen it or change the base branch.

Issue #3371, opened 2020-08-25
Joining chats is broken on `release/next-userspace`


liam-fitzgerald:
Joining chats is broken on `release/next-userspace`. Joining via url does nothing, and clicking join chat shows a stack trace with `props not defined`

cc: @tacryt-socryp

matildepark:
Just gotta pass the prop to the component, yeah? Easy fix.

—
~haddef-sigwen
https://urbit.org

matildepark:
Closing because the purview of this PR was release/next-userspace, not master.

PR #3372, opened 2020-08-25
chat: restore join functionality

matildepark:
See #3371. 

New join behaviour in release/next-userspace had a few issues (no `station` prop passed; using `props` as if `const { props } = this` was there when it wasn't; and a double `/` in the route it redirected you to). All resolved. You can join chats again.


liam-fitzgerald (approved):
LGTM!

Issue #1923, opened 2019-11-06
Nix derivations are somewhat unsound


jtobin:
Many of our Nix derivations depend on the existence of cached pills in bin/.  Except: the files that are present in bin/ are typically only pills if we assume the presence of 1) a git repository, that 2) is non-bare, with 3) a git-lfs configuration installed, and where 4) git's smudge filter has been activated on the LFS pointer files.

This can invalidate our derivations.  In particular: if one uses Nix's 'fetchgit' or 'fetchWithGithub' to grab our source from elsewhere, and then imports e.g. the `urbit` derivation from it, then the derivation will refer to LFS pointers, and Nix will fail to build it.

giacomocariello:
Bump.

joemfb:
@brendanhay, this is fixed by `fetch-github-lfs.nix`, right?

brendanhay:
Yeah, as fixed as we'll probably get with LFS. 😏

ohAitch:
%lull ascendant!

PR #3374, opened 2020-08-25
build: ghc 8.8.4, haskell.nix, nix lfs pills, and ci changes

brendanhay:
This PR makes a number of pervasive nix and related changes in order to provide a better CI caching experience and facilitate upgrading various Haskell related dependencies, the latter requiring the former due to Travis CI timeouts (> 50 minutes) every time we upgrade the stackage resolver and effectively wipe the Travis cache.

## CI Changes

Firstly, by building all of the Haskell project dependencies within nix via [haskell.nix](https://input-output-hk.github.io/haskell.nix/) those dependencies are then also push/pull-able to our nix `cachix` cache. Secondly, since we're using nix there is little point to continually spinning up a new container, restoring an on disk cache from S3, running what might be a tiny change, and then saving said cache back to remote storage - all of which I assume is what Travis does.

This PR changes the CI setup to a [Hercules CI](https://hercules-ci.com/about/architecture/) GitHub integration which notifies our own NixOS server to build all the top-level attributes listed in `ci.nix`. This means we only rebuild what has actually changed, incrementally, replacing the need to save/restore any on container/disk state with `/nix/store`. Successful builds of any expression are then pushed to `cachix`. The server(s) could in theory be set as substituters for developer's local machines, too.

The intended specifications for running the NixOS build agents are AMD Ryzen 5/7 3600/3700 6/8 core, 64GB RAM, 2TB NVMe  bare metal 🤘🏽 machine(s). For the purposes of this PR the build agent is running on a `n1-standard-4` machine with 4 cores, 16GB RAM, 375GB SSD, in `us-central1a` - which comes to 5x the price of the bare metal option.

No work has yet been done to port the `.travis.yaml` OSX cross compilation or deployment related steps - both the Travis and Hercules integrations can be simultaneously active which will allow us to compare the CI options.

## Additional Changes

* All remote sources such as our own forks of `argon2`, `uv`, `h2o`, and so on are declared in `nix/sources.json`. This manifest is (optionally) maintained via the [niv](https://github.com/nmattia/niv) program and the remote sources are accessible via the `pkgs.sources.<...>` attribute. All related expressions have been updated to utilise this. 
* All top-level (`nixpkgs.nix`, `nixcrpkgs.nix`, etc.) package specifications are now parameterised with `pkgs` and `nixpkgs.nix` itself is parameterised on `sources`, allowing downstream dependents to override sources as they desire. Probably more work needs to be done here as I (re)attempt it from another project.
* Pills are now resolvable via nix expressions. This allows us to reference pills in nix expressions for remote CI builds, among other use cases. It specifically avoids the previous case where you need a `.git` repository and had to assume someone had run `git lfs {install,pull}` prior to running any `nix build` that was dependent on pills. It also allows a downstream dependent of the nix expressions to reference pills. We'll see how robust the actual implementation of the impure fetcher is ...
* `herb`'s build expression had to be located with the other `nix/...` build expressions due to a bizarre CI file path bug.
* `static-stack2nix` has been replaced with [haskell.nix](https://input-output-hk.github.io/haskell.nix/). You can observe the mechanisms for a static `musl64` `urbit-king` build in `ci.nix` and `pkg/hs/default.nix`.

This work supersedes #3353 since it contains the stackage resolver and `haskell.nix` updates in addition to the changes which allow the build to complete without the 50 minute Travis timeout observed in that PR.


jtobin:
Looks like this should resolve #1923.

eglaysher:
First immediate feedback: make release isn't working on at least osx. The immediate first problem is a mismatch between 'pkgs' and 'nixpkgs' inside of importing nixcrpkgs in release.nix, but it looks like there are additional problems like sources not getting threaded through things.

brendanhay:
Thanks, still working through various issues - I'll unmark it as draft when I'm slightly more convinced things aren't broken.

eglaysher:
Hmm. Isn't building by default on my linux/nixos machine, since it looks like `nix.useSandbox` is on by default, and the workaround for `--option sandbox relaxed` is being rejected by nix-build. Probably will have to change configuration.nix for that.

brendanhay:
What's the rejection error?

I'm still trying to figure out if the fetcher can avoid using `__useChroot = false`.

eglaysher:
It was complaining that I couldn't pass that option since I wasn't a trusted user. (I did a nixos-rebuild switch to remove the useSandbox = "relaxed"; option to try to get you an exact quote, but since the derivation is built now, it's succeeding.)

brendanhay:
@eglaysher @jtobin The `fetchlfs` fetcher is now a fixed-output derivation that calculates the `sha256` by parsing the LFS pointer for the `oid`, since silly me - they're one and the same, which is rather convenient. I probably should of read the LFS specification more thoroughly before attempting this ..

I think this ended up being a pretty reasonable solution. We don't have to duplicate the LFS binary blob in the nix store (more than once) if you've run `git lfs pull` out of band - they'll just transparently be used as is - otherwise if they're still pointer files per the original "are the first 7 characers == `version`" heuristic, then we proceed with downloading the remote metadata to obtain the actual storage url to retrieve the binary data from.

See: [fetchlfs.nix](https://github.com/urbit/urbit/blob/7ecf1b2ea0bceab94722ca8924200b8c76bf5900/nix/fetchlfs/default.nix)

brendanhay:
(Note: no `sandbox = "relaxed"` shenanigans are required anymore.)

PR #3376, opened 2020-08-25
clay: if commit would be a no-op, don't commit it

### LINKED ISSUE #3317: |autocommit makes thousands of blank revisions
**Describe the bug**

Turning on `|autocommit` steadily increments the clay revision, regardless of any change occurring.

**To Reproduce**

```hoon
> |autocommit
# echo >>zod/home/dojo.hoon
: /~zod/home/4/dojo/hoon
<<some time passes>>
# echo >>zod/home/dojo.hoon
: /~zod/home/1.049/dojo/hoon
```

**Expected behaviour**
The next revision after revision 4 being revision 5, and there being no interim NOP revisions.

**Screenshots**
See above, ish.

 - OS: linux64 [busybox](https://github.com/asssaf/urbit-fakezod-docker)
 - Urbit v0.10.7 OS v1.0.25 😛
  (don't think the IPC rewrite actually touched this much?)
 - `%base` hash 0v1m.o19aa...cms4k, similarly dating back to the existence of /=base=

**Additional context**
N/A

**Notify maintainers**
@philipcmonk @joemfb

philipcmonk:
fixes #3317

Tested with both `|commit` and `|autocommit`

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -1365,7 +1365,6 @@
     |^
     |=  [updated=? =yoki =rang]
     ^+  ..park
-    =/  abort  ..park
     =:  hut.ran  (~(uni by hut.rang) hut.ran)
         lat.ran  (~(uni by lat.rang) lat.ran)
       ==
@@ -1374,12 +1373,12 @@
         %&  q.p.yoki
         %|  (~(run by q.p.yoki) |=(=lobe |+lobe))
       ==
-    =/  old-lobes=(map path lobe)
+    =/  old-yaki
       ?:  =(0 let.dom)
-        ~
-      q:(aeon-to-yaki:ze let.dom)
+        *yaki
+      (aeon-to-yaki:ze let.dom)
     =/  [deletes=(set path) changes=(map path (each page lobe))]
-      (get-changes old-lobes new-data)
+      (get-changes q.old-yaki new-data)
     ~|  [from=let.dom deletes=deletes changes=~(key by changes)]
     ::
     ::  promote ford cache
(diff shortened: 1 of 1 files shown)


ohAitch:
And they say hoon doesn't have `try{}except` mechanisms ;)

One thing that may be good here is extracting out the bit in the middle into an arm whose name clarifies what state is and isn't being changed, so the metaphorical `TRANSACT … ABORT` aren't 70 lines apart. Right now(looking at them on github on mobile, not strictly representative) they *almost* fit on the same page
3EE91E4F-62ED-48B9-9353-50C51D11E514

philipcmonk:
Could be done, but the changes are pretty various: object store, commit store, move list, reef, and ford caches at least.  Although the print logic should probably go at the end anyway, I'll do that now.

ohAitch:
Do we perhaps want the ford caches, if the change is an edit of the raw mime bytes that doesn't affect the parsed representation, and will keep coming in in future attempted-commits?

philipcmonk:
Probably yes, and I suppose the changes to the object store to this point are either nil or they actually made a change, so we could just leave those.  Maybe we don't need to throw anything away.

Also I think I should make sure the commit's parent is the previous (aeon-wise) commit, because they could add several commits and the last one is not a merge commit, and it doesn't happen to change anything compared to the previous commit.  We don't have an interface for doing that, but if someone did do it, it shouldn't be a no-op

ohAitch:
As long as the changes to the object store are actually zero and not just logically zero, i.e. get `sing`ed properly by the comparison.

philipcmonk:
The worry with the ford cache would be throwing away too much.  Looking at the ford cache promotion logic, I don't think we will, at least in the common case.

> As long as the changes to the object store are actually zero and not just logically zero, i.e. get singed properly by the comparison.

We always prefer the existing copy of items in the stores.  Semantically, this is so we don't overwrite a direct with a delta (potentially creating a loop), but it also means we keep the older object, which is nicer on pretty much any garbage regime.

philipcmonk:
Made those changes, @belisarius222 another small review please

Issue #3358, opened 2020-08-20
interface: chat overflows y-axis when unbreakable text is entered


tylershuster:
Example: 
![Thanks ~ravmel](https://marpem-files.sfo2.digitaloceanspaces.com/overflow.png)

I think the solution would be to `white-space: pre-wrap`

matildepark:
Are the crypto kids dumping their keys?

tylershuster:
Galen pasted a base64 image actually

matildepark:
Should only close when it hits master. This time it’s ok.

tylershuster:
except when it's not

tylershuster:
Also happens in inline code blocks

PR #3377, opened 2020-08-26
chat: break overflowing one-word messages

matildepark:
See #3358. Sometimes you dump a base64 string and it makes the entire chat scroll horizontally. It happens.

Adds `overflow-wrap: break-word` to text messages. We also use indigo-react instead of adding another atomic CSS declaration; I also discovered the only thing using the old class declaration was a code overflow-wrap fix that ... isn't currently in use anyway, so it's unnecessarily in the CSS. Deleted.


liam-fitzgerald (approved):
LGTM!

PR #3378, opened 2020-08-26
clay: expose mergebase as base-hash

philipcmonk:
Change the definition of base-hash to be the mergebase of %home with the
OTA source.  This means it's the most recent successfully-applied
update, which is usually the most important information.

Add sour-hash, which is the hash of the most recently *downloaded*
update, regardless of whether it applied successfuly (ie the old
base-hash).

Add a summary of the various hashes at the top of gen/trouble.

I'll note this takes a few seconds now, since +find-merge-points isn't all that optimized (see #3029)

### DIFF
--- a/pkg/arvo/app/file-server.hoon
+++ b/pkg/arvo/app/file-server.hoon
@@ -1,5 +1,5 @@
 /-  srv=file-server, glob
-/+  *server, default-agent, verb, dbug
+/+  *server, default-agent, verb, dbug, version
 |%
 +$  card  card:agent:gall
 +$  serving  (map url-base=path [=content public=?])
@@ -291,24 +291,11 @@
 ++  on-peek
   |=  =path
   ^-  (unit (unit cage))
-  |^
   ?+  path  (on-peek:def path)
-    [%x %clay %base %hash ~]  ``hash+!>(base-hash)
+      [%x %clay %base %hash ~]
+    =/  versions  (base-hash:version [our now]:bowl)
+    ``hash+!>(?~(versions 0v0 (end 0 25 i.versions)))
   ==
-  ::  stolen from +trouble
-  ::  TODO: move to a lib?
-  ++  base-hash
-    ^-  @uv
-    =+  .^  ota=(unit [=ship =desk =aeon:clay])
-            %gx  /(scot %p our.bowl)/hood/(scot %da now.bowl)/kiln/ota/noun
-        ==
-    ?~  ota
-      *@uv
-    =/  parent  (scot %p ship.u.ota)
-    =+  .^(=cass:clay %cs /[parent]/[desk.u.ota]/1/late/foo)
-    %^  end  0  25
-    .^(@uv %cz /[parent]/[desk.u.ota]/(scot %ud ud.cass))
-  --
 ++  on-agent  on-agent:def
 ++  on-fail   on-fail:def
 --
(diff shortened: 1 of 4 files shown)


matildepark:
Could you bring [file-server](https://github.com/urbit/urbit/blob/master/pkg/arvo/app/file-server.hoon#L300-L310) to parity with the new heuristic?

philipcmonk:
Done.  However, this takes about 3 seconds on my moon, which isn't great since it's run whenever you go home.  I'll think a bit about how to make it faster in common cases.

philipcmonk:
Alternately, if the home page didn't block on it, that wouldn't be too bad.

matildepark:
It should just be a promise, but presumably Arvo could be blocking serving the diff in the initial subscription to Launch because it's processing the peek

ohAitch:
How does "about 3 seconds" change as revisions pile up? Presumably it's not
3 seconds on a brand new desk with 1 version?

philipcmonk:
@ohAitch Correct, it's instantaneous on a new ship.  The moon i've been testing on is fairly new, but of course it has all the OTAs since December.  I could try it on my planet, but 3 seconds is already slow enough that I'm going to review #3029 and see if I can make it faster.

philipcmonk:
Okay, I rewrote the +find-merge-points algorithm, and it seems to be much faster (and probably still correct!).  On that moon, instead of 3 seconds it's essentially instantaneous.  I didn't try the previous version on my planet, but this version is very fast on it, certainly less than a second.

I don't know how to concisely explain the difference in the algorithms, but the main thing is that we now aggressively check whether we're in ancestry of another mergebase candidate.  This allows us to short-circuit instead of creating redundant candidates (ie candidates who are in each others' ancestry) which need to be cleaned up in +reduce-merge-points (which was quite expensive).

I still need to test the algorithm under some additional merge conditions (criss-cross in particular).

ohAitch:
Urbit port of https://github.com/DRMacIver/minithesis when ;)

philipcmonk:
Added some comments, and tested a criss-cross merge with this:

```
|merge %cross our %home
*%/test/hoon 'hi'
*%%%/cross=/toast/hoon 'hey'
|merge %cross our %home
|merge %home our %cross, =cas ud+(dec -:.^([@ud *] %cw /=cross=))
.^((list @uv) %cs %/base/~zod/cross)

~[
  0v15.q2ftg.eu4l7.elm2f.kuuvv.hm0ee.too42.qc0tk.6kif9.r6rkt.rjegn
  0v16.ck026.70acj.pipl5.fou6k.5d8e3.qqt5g.49ubu.elm28.ce5q4.45405
]
```

This correctly identified two mergebases.  I think this is ready to go.

philipcmonk:
@belisarius222 could i get a review when you get a chance?

belisarius222:
Yeah, sorry, this slipped through the cracks

belisarius222 (approved):
LGTM.  It took me a while to think through the various cases in `+find-merge-points`, but it looks correct to me.

Fang-:
Looks like the merge of this (f9665c0) failed CI. Some `nest-fail` in `/lib/hood/kiln`. Not immediately clear to me why the PR CI on the commits here wouldn't catch that...

joemfb:
This PR doesn't included an updated solid pill, could be related. As for why it failed after the merge but not earlier, CI on a PR is not triggered when the target branch is updated, but only on source branch changes.

philipcmonk:
I think that issue came from #3404, where I changed +germ without pushing a new pill.  I just pushed a new pill, which resolved this.

PR #3379, opened 2020-08-26
file-server: fix %file-server by adding SPA flag and threading is-file through request

tacryt-socryp:
…h is-file through request

### DIFF
--- a/pkg/arvo/app/file-server.hoon
+++ b/pkg/arvo/app/file-server.hoon
@@ -2,25 +2,23 @@
 /+  *server, default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
-+$  serving  (map url-base=path [=content public=?])
++$  serving    (map url-base=path [=content public=? single-page=?])
 +$  content
   $%  [%clay =path]
       [%glob =glob:glob]
   ==
-+$  state-base
-  $:  =configuration:srv
+::
++$  state-3
+  $:  %3
+      =configuration:srv
       =serving
-  ==
-+$  state-2
-  $:  %2
-      state-base
   ==
 --
 ::
 %+  verb  |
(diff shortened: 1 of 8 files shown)


matildepark:
Tested with create-landscape-app, patched for the new boolean. I'm getting 404s on the image files, but they're definitely in app/testing/img/.

image

pkg/arvo/app/file-server.hoon
```diff
@@ -113,14 +122,17 @@
...
+        (~(put by serving) url-base clay+clay-base.act public.act spa.act)
+      ==
     ::
         %serve-glob
       =*  url-base  url-base.act
       ?:  (~(has by serving) url-base)
         ~|("url already bound to {<(~(got by serving) url-base.act)>}" !!)
       :-  [%pass url-base %arvo %e %connect [~ url-base] %file-server]~
-      this(serving (~(put by serving) url-base glob+glob.act public.act))
+      this(serving (~(put by serving) url-base glob+glob.act public.act %.y))
```
matildepark, line 135: So we pass `%.y` as the spa flag on `:glob`, which won't have the file — it mounts its state as `bundle.[hash].js`. Given the `?&` for spa and is-file flags, will this play nice?

tacryt-socryp: It works fine so long as the file name is lowercase. This is intended behavior given constraints around clay paths.

matildepark: I just mean because glob doesn't put it in clay, it pokes file-server to mount directly from state AFAIK.

tacryt-socryp: Not a concern given that is-file overrides that flag.

PR #3380, opened 2020-08-26
[WIP]: Build all dependencies to osx urbit statically.

eglaysher:
This is more asking if the approach I'm taking is viable.

For the urbit binary, I've gone and changed things so that it takes all of its dependencies explicitly, instead of just fishing them out of pkgs. Then I've made a sysdeps.nix, which takes pkgs and makes a set of the urbit binary's dependencies from nixpkgs with overrides to make them compile statically. The result of this is a binary which has no dependencies on things in /nix/store/...except for libSystem.dylib. Next up is trying to replace the nixified osx libc with the /usr/lib/ version.

(Note: something happened to the libaes_siv build on OSX between testing this locally and then rebasing onto tot of bh/hercules-ci, where here's no files in the libaes_siv output. This happens on the current bh/hercules-ci branch without my patch. Not sure what's up there.)


brendanhay:
Overriding the packages as you've done seems like the right idea, although possibly `sysdeps.nix` would be a better fit as an overlay? We could pass the `nix/overlays/static.nix` as an argument to `import ./nix/nixpkgs.nix { overlays = ... }`, for example. This would allow us to continue to referring to `pkgs`, without introducing another `sysdeps` binding/import that you need explicitly pass to every sub-expression and remember to refer to. The necessary packages in the `pkgs` set would be transparently overridden and the appropriate flags added/enabled/disabled.

`nix/overlays/statix.nix`:

```nix
final: prev: {
  curl = prev.curl.override { ... };
  zlib = prev.zlib.override { ... };
  ...
}
```

`nix/nixpkgs-static.nix`:

```nix
let
  overlays = map import [
    ./overlays/static.nix
  ];
in import ./nixpkgs.nix { inherit overlays; }
```

`nix/release.nix`:

```nix
{ pkgs ? import ./nixpkgs.nix { } }:

# Usual release imports here which take `pkgs` and call pkgs.curl, etc.
let
  urbit = import ./pkgs/urbit { inherit pkgs; };
in { ... }
```

And to build dynamic vs static:

```
nix build -f nix/release.nix 
nix build -f nix/release.nix --arg pkgs 'import ./nix/nixpkgs-static.nix { }'
# ^ Or conditionally do this per platform in release.nix.
```

See [overlays](https://nixos.wiki/wiki/Overlays) for more information, or look under `nix/overlays` in `ylem` for more examples.

eglaysher:
The immediate problem with doing this as an overlay is that the modified zlib gets seen by everything because basically everything relies on zlib. So far it's rebuilt binutils, bash, cmake and python...and since it's reubilding python, that implies that it will rebuild ghc since python is a nativeBuildInputs to that derivation.

I agree that this way is a bit uglier, but it has the advantage that these rebuilt versions only affect the Urbit binary.

eglaysher:
OK, so just raw injecting a call to `install_name_tool` fixed up the final dependency on a `/nix/...` path, but this should be triggered a different way instead of always doing this during the install phase, regardless of the build type. How would you go about adding optional code to `installPhase`?

The other difference (which I don't think matters), is that the resulting binaries depend on being dynamically linked to CoreFoundation because of curl...but I don't understand why. First, it actually shouldn't matter; it's not a nix path and it looks like it is specified with `@rpath@`, so it should work anywhere. But it is a difference that I'm unable to account for and every attempt at trying to get curl to not link in CoreFoundation has been unfruitful.

brendanhay:
@eglaysher @joemfb Sorry I somehow missed this until Joe pointed me at it - you could do something like:

```
installPhase = ''
  <always do this stuff ...>
'' + lib.optionalString bool ''
  <conditionally do this>
'';

Or you could do it inline using `${ ... }`.

brendanhay:
I see you've already done that! 😅

PR #3381, opened 2020-08-26
arvo/README: update maintainers

matildepark:
RFC as to current Arvo maintainers. Correct me where I'm wrong? @philipcmonk @joemfb


philipcmonk:
These changes look correct to me

PR #3382, opened 2020-08-27
Links s3

tylershuster:
dupe #3367


pkg/interface/src/views/apps/links/app.js
```diff
@@ -152,6 +152,22 @@ export class LinksApp extends Component {
                   listening={listening}
                   api={api}
                 >
+<<<<<<< HEAD
```
liam-fitzgerald, line 155: ????

tylershuster: f

tylershuster:
Ok, this should be good. I'm making some changes to spinner because it needs them and suits my purpose.

matildepark:
Still needs parity ... 

before:

image

currently:

image

tylershuster:
Ok, it's just a few pixels off now, but it uses classes instead of explicit pixels.

urcades:
I have a very small design tweak/proposal here, let me know if it's possible/makes sense in the context of this PR:

Screen Shot 2020-09-02 at 4 55 18 PM

As a reference point, the activity being described here can be observed in are.na's block input:

arena

tylershuster:
I like this, but I wonder about it being in this commit. It's already changing a bit of how the S3 component works. I'm happy to turn these designs into reality but I'm getting a little nervous about the current scope.

matildepark:
It would be good to contain this and do the design tweak / additional functions separately yes.

—
~haddef-sigwen
https://urbit.org

urcades:
That's fine! We can ignore this comment and I can situate it in an incremental update. Will check with @matildepark on where best to locate this feedback.

tylershuster:
Would probably pair well with https://github.com/urbit/landscape/issues/74

urcades:
Ah, perfect call. Thanks, will locate this feedback there.

matildepark (approved):
Clean commits and I'll merge

tylershuster:
These are some messy commits

Issue #3196, opened 2020-07-27
Eyre errors, %eyre-no-channel


galenwp:
**Describe the bug**
I see regular stream of 

```
crud: %eyre-no-channel event failed
id='1595502659818-a9a43e'
```

in my scrollback

**To Reproduce**
Steps to reproduce the behaviour:
Not sure exactly, but I imagine it's due to something Landscape-related

**Expected behaviour**
Not seeing these errors

**System (please supply the following information, if relevant):**
 - OS: Ubuntu
 - Vere and Urbit OS versions: 10.7 / `jp5fi`

matildepark:
Still getting this, Galen? I think this was a side effect of an infinite loop we stumbled into propagating group updates on loop. It seems to have quieted on my ship.

— 
~haddef-sigwen
https://urbit.org

matildepark:
Ah, I still see some. Sorry — hosted ships have a slight delay it seems.

matildepark:
I think you're right that this happens during use of Landscape. In particular this happens when I navigate to another chat — we send a request to mark the chat read every time we click on a chat and scroll down or up IIRC, and that seems to instigate this for me.

tacryt-socryp:
This occurs when a channel is sent a request from the JS but that channel does not exist or has already been cleaned up. The question here is: why is a nonexistent channel being sent requests? It doesn't cause any problems for this to occur, but we should start automatically establishing a new channel occasionally if we receive enough errors. Hmm... Will look into this behavior. Can't recreate on my ship.

tacryt-socryp:
Marking as not an issue, as it’s expected (and fine) behavior.

galenwp:
It's not really 'fine', since my console is unusable. It seems like this should either be able to be silenced, or shouldn't happen in the first place. Happy to show someone my pier, it happens constantly.

joemfb:
@tacryt-socryp, from reading the %eyre source, I think is the opposite of the situation you describe. The `%eyre-no-channel` message is produced in `+emit-event`, which is only ever called when processing a subscription update from `%gall`. In other words, this happens when %gall thinks %eyre is subscribed to it, but %eyre doesn't (no longer?) know(s) about that subscription.

#3248 is at least partly related.

Fang-:
> but %eyre doesn't (no longer?) know(s) about that subscription.

Yeah. The issue here seems to be that eyre forgot to clean up a gall subscription when it closed a channel.

[...]

PR #3387, opened 2020-08-27
eyre: prevent %eyre-no-channel errors

Fang-:
This is a band-aid fix for #3196. The root cause for that still eludes me. All of the involved code "looks good to me".

Still, the symptoms of #3196 are annoying, so we address those directly. We detect the case that would lead to the error and, instead of trying to talk to the now-gone channel, we `%leave` the subscription that gave us the update. Since this requires new details from the wire, we cannot do this for existing cases, and instead rely on gall to forget all of eyre's subscriptions, "just this once".

Tested on a copy of @galenwp's ship, who was suffering from this issue.

idk what's up with the hercules failures, but it doesn't look like they're relevant here?

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -11,9 +11,9 @@
 ::  $move: Arvo-level move
 ::
 +$  move  [=duct move=(wind note-arvo gift-arvo)]
-::  $state-6: overall gall state, versioned
+::  $state-7: overall gall state, versioned
 ::
-+$  state-6  [%6 state]
++$  state-7  [%7 state]
 ::  $state: overall gall state
 ::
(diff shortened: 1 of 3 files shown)


eglaysher:
Don't worry about the hercules failures, you can ignore those for now until the changes to our nix derivations are merged.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1732,7 +1759,7 @@
 ::
 ++  stay
   ^-  spore
-  =;  eggs=(map term egg)  state(yokes eggs)
+  =;  eggs=(map term egg)  [- | +]:state(yokes eggs)
```
philipcmonk, line 1762: > `[- | +]`

cute

PR #3388, opened 2020-08-27
hoon: remove ^% ketcen

Fang-:
This came up iirc during lunch call some time ago. "Test mode" is old, seemingly unimplemented, and entirely unused. So we should just remove it.


philipcmonk:
Does this upgrade cleanly?  I suspect it wouldn't, since it makes +type smaller.

pilfer-pandex:
@philipcmonk, this patch doesn't touch $type in any way. ^% is a directive to +ut: recur processing the subhoon with the `fab` flag set to true. In the compiler commentary, I say about `fab`:

> Comments refer to this flag as demarcating a “fabrication mode” or “test mode,” which can be turned off (or on?) by ^%. Although it is duly passed into ap and ax, the value is never read in live code, and has no effect on the operation of the compiler.

I think you're thinking of %hint in $type, which is unrelated.

Given the code I looked at, I don't see how this can cause upgrade problems. Lmk if I'm wrong, Philip.

philipcmonk:
The %hold type includes a +hoon, so +type is covariant in +hoon.  This is one of the reasons I don't love %hold, but it isn't a spurious dependency -- if you're migrating a vase to the new kernel and it has a %hold type with %ktcn, we do need to explicitly strip that.

Could be a fun exercise to do this upgrade, though.

```
> |commit %home
>=
vega: compiling hoon
ride: parsing
ride: compiling
ride: compiled
ride: parsing
ride: compiling
ride: compiled
vega: compiling kernel ~mocbex-fogseg
ride: parsing
ride: compiling
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
fund: in part, parent 33b1b99f not found at 7
ride: compiled
vega: extracting core
arvo: assembly
arvo: assembled
vega: +load
ride: parsing
ride: compiling
ride: compiled
ride: parsing
ride: compiling
ride: compiled
vega: evolving
zuse: ~dotdex-saldyl
vane: %a ~maglet-masfet
vane: %b ~nocsed-radtul
crud: %into event failed
[%poke p=%into]
%leap
[%feck p=%lyra]
vega: ruin
[%failed-vane-activation %b]
%load-lost
nest-fail
veer: %b
[%feck p=%veer]
%leap
[%feck p=%veer]
%leap
[%feck p=%veer]
%leap
%load
```

Ames succeeded because it holds no vases and i think we trust the we trust the vane vases to be the same?  Behn has vases in its state though and so that fails.

pilfer-pandex:
!@#$!#$%^@$%^@#$ holds grrr

I totally forgot about that. Well pointed out. :(

>

ohAitch:
"hooks to iterate through every single vase in the system" is definitely desirable longterm: why else have `++typo` and friends.

Fang-:
I think the answer is yes, but to confirm: would retargeting this change onto `release/next-breach` make this PR legit without further work?

joemfb:
`na-release/next-vere`, but yes. That branch is a bit behind, I'll be updating it with several merges soon. In the meantime, you can base changes off of `jb/na-release-merge`.

joemfb (changes requested):
`na-release/next-vere` is now (mostly) up-to-date, but `na-release/runes` is the better target, as it also has the corrected hoon AST labels (per the standard pronunciations).

These changes LGTM, but I'd like to request another: `fab` should be removed from the `|ut` context as well -- it's only purpose was to be set by `%ktcn`. This change would eventually need to be reflected in the memoizing compiler jets (see #3974), but it shouldn't break anything for purposes of this PR.

Fang-:
This should be it, assuming the jets don't need any more work.

joemfb (approved):
LGTM

PR #3390, opened 2020-08-27
interface: added graph reducer, store state, and subscription

tacryt-socryp:
Works fine and has no functionality changes. Will append and test scrying functionality to this PR, then we can get it merged with no functionality changes. Will make it easier to review the Links migration as its own thing.

### DIFF
--- a/pkg/interface/src/logic/reducers/graph-update.js
+++ b/pkg/interface/src/logic/reducers/graph-update.js
@@ -1,20 +1,14 @@
 import _ from 'lodash';
 
 
-const OrderedMap = (arr = []) => {
-  let map = new Map(arr);
-//  map[Symbol.iterator] = function* () {
-//      yield* [...this.entries()].sort((a, b) => a[1] - b[1]);
-//  };
-  return map;
-};
-
 export const GraphReducer = (json, state) => {
   const data = _.get(json, 'graph-update', false);
   if (data) {
     keys(data, state);
     addGraph(data, state);
+    removeGraph(data, state);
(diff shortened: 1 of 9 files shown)


matildepark:
Looks nice and clean so far, but I'll wait for the remainder too.

tacryt-socryp:
@belisarius222 is there an efficient way to convert between a mop and a map?

belisarius222:
No.  It might be good to add conversion routines from mop to map and back, although arguably converting to list as an intermediate is ok.

tacryt-socryp:
@matildepark All scries have been tested to work. The reducer isn't finished, but far enough that I can move to Links interface integration.

pkg/interface/src/logic/reducers/graph-update.js
```diff
@@ -0,0 +1,145 @@
+import _ from 'lodash';
+
+
+const OrderedMap = (arr = []) => {
+  let map = new Map(arr);
+//  map[Symbol.iterator] = function* () {
```
belisarius222, line 6: Why is this commented out?

matildepark:
> Will append and test scrying functionality to this PR, then we can get it merged with no functionality changes.

Wasn't the scope of this PR intentionally limited?

tacryt-socryp:
Yep! Good point. Removed that commit. This is good to go. I'd like to merge it into release/next-userspace now.

matildepark:
Merge #3352 while you're at it. I'm gonna review the s3 one now ...

PR #3391, opened 2020-08-27
profile: log out of all sessions

matildepark:
This was hidden in urbit-os-v1.0.44 until we could ensure it actually
logged you out of all sessions. It requires a hidden input to function,
so it's added here.

See #2649 for initial feature request. Sorry for the wait, @custom-jonathan!


custom-jonathan:
Awesome! Better now than never. <3

liam-fitzgerald (approved):
LGTM!

Issue #3283, opened 2020-08-11
Unread Notebook post count misaligned if title occupies >1 line


sivner-figbus:
**Describe the bug**
In Landscape, I have a misaligned unread message count for a Notebook which has a title that overflows onto 2 lines. It is easier to show than to describe:

![image](https://user-images.githubusercontent.com/67953485/89901468-05377f80-dbdd-11ea-9853-93db28b9d60f.png)

**To Reproduce**
Create a notebook in "publish" with a long name.
Create at least one post within the notebook.
View the notebook from another ship, in order that an unread posts count is displayed

![image](https://user-images.githubusercontent.com/67953485/89907216-48492100-dbe4-11ea-9df8-2ccd47fd1f3d.png)

**Expected behavior**
IDK but this doesn't look right, and gets messy if another group name is displayed directly underneath

 - OS: Mac OS X 10.15.5
 - Browser Version 1.11.104 Chromium: 84.0.4147.105 (Official Build) (64-bit)
 - Same in Safari Version 13.1.1 (15609.2.9.1.2)
 - Base hash of your urbit ship. 0v1i.s1lmp.1uiuf.e8mtd.mu0lr.m54qh.b9gm3.qa7im.pemgd.3uh25.dg0qj

matildepark:
Blocked on #3337 merge; may resolve, will test after.

PR #3392, opened 2020-08-27
publish: fix unread counts for long titles

matildepark:
Re: #3283

The Publish refactor partially addressed the issue in question, but produced some flexbox squishing of the unread count instead. We give the title some padding on the right and disable flex-shrink to compensate.

Screenshot 2020-08-27 at 3 34 06 PM


liam-fitzgerald (approved):
LGTM!

PR #3393, opened 2020-08-27
chat: break up inline code

matildepark:
Importantly, doesn't affect the white-space of %code messages.

See #3358.


liam-fitzgerald (approved):
LGTM!

PR #3397, opened 2020-08-27
chat: only show other participant in dms

tylershuster:
Pretty simple implementation: just tests to see if the title of the chat is going to be the ship name surrounded on either side with ` <-> `, and if so, replaces that same regex with nothing. Works for all my chats:

secrets


matildepark:
Most of the weird complexity of https://github.com/urbit/urbit/commit/3290db961506da9cb2687a377823d6f2d4d662d6 was because I wanted to cover the case where DMs shows "~ravmel-ropdyl, ~fallyn-balfus" for eg. 

But this sort of dumb approach, punting group DMs to use their own name, seems good.

PR #3399, opened 2020-08-28
chat: added oembed parser

tylershuster:
This PR adds oembed parsing to chat. OEmbedding is an open protocol to specify embedding content in pages from a URL. Examples include YouTube and Twitter.  Here's a video of it working:

[Video](https://marpem-files.sfo2.digitaloceanspaces.com/Screen%20Recording%202020-08-27%20at%203.08.04%20PM.mov)

Sick, right?

## Problems This Fixes
- There's currently an exception for youtube to iframe its content, which is Weird given urbit's position toward MEGACORP. This implements an open protocol that youtube *also* implements. Incidentally their oembed response is just the iframe that chat already uses.
- Obviously, it vastly expands the possibilities for oembeds
- It replaces the links of oembeddable content with their title, if available. I had someone ask me about doing this via the api and ew.

Ok here's some

## Caveats
- It uses https://noembed.com/ to bypass the CORS restrictions that would happen with a vanilla implementation
  - Counter-caveat a: This could be avoided by proxing requests through arvo.
  - Counter-caveat b: A *lot* of people use this service. Assuming the above solution is eventually implemented, this wouldn't result in a significantly diminished experience if noembed were to go down.
- It makes a request to a third party without explicit permission
  - Counter-caveat: It already does this; see: images.

## Future Work
I've always imagined this feature being paired with a new section in settings called "Remote Content."

```
[ ] Load images
[ ] Load videos
[ ] Load audio
[ ] Load oembeds
```

And we default to images being on, I guess, because that's what people are used to. The idea though is that your browser shouldn't load any domains except the one you're on.

Anyway I see these two features as going hand-in-hand.

RFC


tacryt-socryp:
I prefer this approach to remote content to the current one. Looks excellent.

tylershuster:
Reviewing, I think we probably don't want to render any remote content (even under `dn`) without clicking embed.

matildepark:
Might be worth componentising this somehow. Links uses similar stuff.

matildepark:
Are you intending to revise the PR or submit subsequent changes separately?

tylershuster:
Revising today or tomorrow. I’ve been outside. 

T

matildepark:
That’s what I like to hear!

—
~haddef-sigwen
https://urbit.org

tylershuster:
This is my first pass at a new local setting. Seems to work, unimplemented as of yet.

Next step is to incorporate these new settings into a) `views/apps/chat/compontents/lib/content/url.js` and b) `views/apps/links/components/lib/link-detail-preview.js`

Comments appreciated

tylershuster:
Alright, there's actually something to look at here. Some styling needs to be included, but I think the basic logic is solid. Defaults to loading everything except oembeds.

matildepark:
Will review in the morning.

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+              <Checkbox
+                label="Load Audio"
+                id="audioShown"
+              />
+              <Checkbox
+                label="Load Videos"
+                id="videoShown"
+              />
+              <Checkbox
+                label="Load Oembeds"
```
matildepark, line 81:
```suggestion
                label="Load embedded content"
```

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+              <Checkbox
+                label="Load Images"
+                id="imageShown"
+              />
+              <Checkbox
+                label="Load Audio"
+                id="audioShown"
+              />
+              <Checkbox
+                label="Load Videos"
```
matildepark, line 77:
```suggestion
                label="Load video files"
```

(Do we even have this?)

matildepark: So we do.

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+            <Box color="black" fontSize={1} mb={3} fontWeight={900}>
+              Remote Content Policy
+            </Box>
+            <Box>
+              <Checkbox
+                label="Load Images"
+                id="imageShown"
+              />
+              <Checkbox
+                label="Load Audio"
```
matildepark, line 73:
```suggestion
                label="Load audio files"
```

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+            gridTemplateColumns="1fr"
+            gridTemplateRows="audio"
+            gridRowGap={3}
+          >
+            <Box color="black" fontSize={1} mb={3} fontWeight={900}>
+              Remote Content Policy
+            </Box>
+            <Box>
+              <Checkbox
+                label="Load Images"
```
matildepark, line 69: Load images

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+    >
+      {(props) => (
+        <Form>
+          <Box
+            display="grid"
+            gridTemplateColumns="1fr"
+            gridTemplateRows="audio"
+            gridRowGap={3}
+          >
+            <Box color="black" fontSize={1} mb={3} fontWeight={900}>
```
matildepark, line 64: You should do font stuff in a separate component, `<Text>`.

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+              Remote Content Policy
+            </Box>
+            <Box>
```
matildepark, line 67: You'll want a `<Text gray>` that says "Embed remote content within this Landscape session."

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,93 @@
...
+      {(props) => (
+        <Form>
+          <Box
+            display="grid"
+            gridTemplateColumns="1fr"
+            gridTemplateRows="audio"
+            gridRowGap={3}
+          >
+            <Box color="black" fontSize={1} mb={3} fontWeight={900}>
+              Remote Content Policy
```
matildepark, line 65: "Remote Content" is fine.

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
tylershuster: Tell that to whomever wrote `DisplayForm.tsx` ah ah ah

tylershuster: Really though, what should I use for `Text` settings? All the other sections on this page use `Box`

tylershuster:
Ok, this is good feedback. I added a few options to the `<RemoteContent>` component that increase its contextability (contextualizationability?). Then I applied those in chat and in links, to more or less maintain the existing behavior with new behavior added on top of it.

matildepark:
Screenshot 2020-09-02 at 5 18 40 PM

YouTube embeds now cause horizontal overflow on mobile.

Screenshot 2020-09-02 at 5 19 10 PM

Links should not have the 'collapse' button available at all, basically.

matildepark:
How did this conflict?

tylershuster:
Some jerk pushed a change to links earlier today. something to do with s3 i don't know

matildepark:
ok will test, just clean some of the commits here and we'll get going probably

matildepark (approved):
Feels very nice. Some slight padding issues with the settings console here, but we're probably just doing a holistic pass later anyway.

tylershuster:
Can you give this a second look? I've had some merging issues this evening.

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
matildepark, line 81: We lost these changes.

matildepark:
That seems to be the only regression, settings copy.

tylershuster:
Ok, that was kind of weird, but as long as it works now.

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -0,0 +1,94 @@
...
+              <Checkbox
+                label="Load images"
+                id="imageShown"
+              />
+              <Checkbox
+                label="Load audio files"
+                id="audioShown"
+              />
+              <Checkbox
+                label="Load field files"
```
matildepark, line 77: What are field files ...

tylershuster: ihaventhadmycoffee.jpg

PR #3404, opened 2020-08-28
clay: add safe merge strategies

philipcmonk:
See gen/hood/merge/help.txt for an overview of all merge strategies.

- `|merge` with no arguments now prints out usage and a description of merge strategies
- Added `%take-this` and `%take-that` strategies, which are the same as the old `%this` and `%that` except they preserve files which only exist on one side or the other.  The primary use-case for `%take-that` is "forcing" updates without obliterating locally-added files, such as notebooks or 3rd party apps.
- Added `%meet-this` and `%meet-that` strategies, which are the same as `%meet`, except if both sides touched the same file (since their most recent common ancestor), prefer this or that desk.  Note this operates on the file level -- even if the changes inside the file don't conflict, we take the whole file from one side or the other.  It would be nice to add `%mate-this` and `%mate-that` strategies which operate within the file.
- Renamed `%this` and `%that` to `%only-this` and `%only-that` to emphasize that they remove anything not in the other desk.
- Note `%take-this` and `%take-that` are 2-way merges (like `%only-this` and `%only-that`).  This means they only look at the content in the commits, not their ancestry.  This means they are suitable for cases where you don't have a unique mergebase (`%merge-no-merge-base` and `%criss-cross`) and generally makes them blunt weapons.  By contrast, `%meet-this` and `%meet-that` *do* require a mergebase and are a sort of 3-way merge.
- Switch the default merge strategy for desks on their first revision from `%that` to `%take-that`.  This should alleviate some issues where people add files (eg make a notebook) before the first OTA arrives.

### DIFF
--- a/pkg/arvo/gen/hood/merge/help.txt
+++ b/pkg/arvo/gen/hood/merge/help.txt
@@ -69,7 +69,7 @@ you're guaranteed to have exactly the files in the source commit plus
 any files you separately added.
                                                                       
 We speak of merging into a destination *desk* from a source *commit*
-becaus while you can only merge on top of a desk, you can merge from
+because while you can only merge on top of a desk, you can merge from
 historical commits.  For example,
                                                                       
     |merge %old our %home, =cas ud+5, =gem %init
--- a/pkg/arvo/gen/hood/merge/help.txt
+++ b/pkg/arvo/gen/hood/merge/help.txt
@@ -1,0 +1,110 @@
+Usage:
+
+    |merge %destination-desk ~source-ship %source-desk
+    |merge %destination-desk ~source-ship %source-desk, =gem %strategy
+    |merge %destination-desk ~source-ship %source-desk, =cas ud+5
+
+We support various merge strategies.  A "commit" is a snapshot of
+the files with a list of parents plus a date.   Most commits have
+one parent; a "merge" commit is a commit with two parents.  The
+%home desk starts with an initial commit with no parents; commits
+with several parents ("octopus merges") are possible but we don't
+generate them right now.
+                                                                      
+Unless otherwise specified, all of the following create a new commit
+with the source and destination commits as parents.
+                                                                      
+Several strategies need a "merge-base".  They find it by identifying
+the most recent common ancestor of the two desks.  If none, fail
+with %merge-no-merge-base; if there are two or more, pick one.
+                                                                      
+%init: the only way to create a desk.  Not a true merge, since it
+simply assigns the source commit to the destination.
+                                                                      
+%fine: if source or destination are in the ancestry of each other,
+use the newer one; else abort.  If the destination is ahead of the
+source, succeed but do nothing.  If the source is ahead of the
+destination, assign the next revision number to the source commit.
+Some call this "fast-forward".
+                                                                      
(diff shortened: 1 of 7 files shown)


pkg/arvo/gen/hood/merge/help.txt
```diff
@@ -0,0 +1,110 @@
...
+%meet: combine changes, failing if both sides changed the same file.
+Specifically, take diff(merge-base,source) and
+diff(merge-base,destination) and combine them as long as those diffs
+touch different files.
+                                                                      
+%mate: combine changes, failing if both sides changed the same part
+of a file.  Identical to %meet, except that some marks, like %hoon,
+allow intelligent merge of changes to different parts of a file.
+                                                                      
+%meld: combine changes; if both sides changed the same part of a
```
belisarius222, line 39: Does this mean %meld doesn't apply either change, and just goes with the old version?

pkg/arvo/gen/hood/merge/help.txt
```diff
@@ -0,0 +1,110 @@
...
+                                                                      
+The most common merge strategy is %mate, which is a normal 3-way
+merge which aborts on conflict.
+                                                                      
+%take-that is useful to "force" an OTA.  After running %take-that,
+you're guaranteed to have exactly the files in the source commit plus
+any files you separately added.
+                                                                      
+We speak of merging into a destination *desk* from a source *commit*
+becaus while you can only merge on top of a desk, you can merge from
```
belisarius222, line 72: typo: becaus

pkg/arvo/gen/hood/merge/help.txt
philipcmonk: That's correct.  This makes it convenient to construct a 3-way diff later on, but it's somewhat non-intuitive

belisarius222: Ah, I see, this is used for annotating conflicts on top of it.  Makes sense.

pkg/arvo/gen/hood/merge/help.txt
```diff
@@ -0,0 +1,110 @@
...
+Note this is a normal %only-that merge, which means you're creating a
+*new* commit with the old *contents*.
+                                                                      
+%meld is rarely used on its own, however if you specify %auto or
+omit the merge strategy, %kiln will run a %meld merge into a scratch
+desk and then annotate the conflicts there.
+                                                                      
+If you resolve merge conflicts manually, for example by mounting the
+desks, copying the files in unix and then running |commit, you
+should usually run an %only-this merge.  This will not change the
```
belisarius222, line 92: As a note for the future, I think we should eventually automate this workflow for development.  It could be related to the `-work` thread.

philipcmonk: Good idea.  Would love it to to just pop up your `$EDITOR`, but that's probably a bridge too far...

belisarius222 (approved):
LGTM other than that one typo

philipcmonk:
Fixed, and also fixed a small bug in `%mate` that I noticed while trying to update a 3rd party app: https://github.com/urbit/urbit/pull/3404/commits/c7480c9b0700274afdb92a009fc01f53ba08e9fb

PR #3405, opened 2020-08-29
ames.c: better stateless forwarding

### LINKED ISSUE #3345: ames.c stateless forwarding issues
- [ ] If the scry fails once (with a `u3_none` result), it stops trying stateless forwarding forever. We don't expect the scry to fail, there's a bug there that needs fixing, but we should also just be robust against this, probably.

- [ ] It doesn't cache lanes, so it's only about as fast as stateful forwarding.

- [ ] Just observed ~nus getting into a state where it wasn't processing new forwards anymore, and exclusively  dropping them. (As evidenced by lack of "forwarded x total" prints, while the "dropped x forwards total" prints continued.) Printing driver info wasn't very interesting, the "forwards pending" queue was filled, as it always is... Are we maybe not cleaning entries in that up properly, accumulating "dead" entries in the queue until we no longer have room for "live" ones? Wouldn't be too surprised if it turns out I messed something up about the double-linked list manipulations...

Fang-:
Fully fixes #3345:

- No longer fail to do bookkeeping on the scry queue in the "scry says we have no lane" case. This would lead to the queue "filling up" with non-existent entries, eventually resulting in all forward requests to be dropped on the floor.

- No longer give up on scrying at the first sight of failure. Instead, only do so if multiple lane scries fail in succession.

- Cache lane scry results for better performance. 

I had the pre-caching version running on ~nus for a while. On average, for every forward request it completed, it would drop three.  
Now, with caching, it seems to keep up with network traffic just fine. It's running at 100% cpu usage continuously to do so, but it's no longer dropping large amounts of packets (only dropping any briefly after booting, due to the empty cache).

### DIFF
--- a/pkg/urbit/vere/io/ames.c
+++ b/pkg/urbit/vere/io/ames.c
@@ -113,10 +113,10 @@ _ames_pact_free(u3_pact* pac_u)
 */
 static void
 _ames_panc_free(u3_panc* pac_u) {
-  if (0 != pac_u->nex_u) {
+  if ( 0 != pac_u->nex_u ) {
     pac_u->nex_u->pre_u = pac_u->pre_u;
   }
-  if (0 != pac_u->pre_u) {
+  if ( 0 != pac_u->pre_u ) {
     pac_u->pre_u->nex_u = pac_u->nex_u;
   } else {
     c3_assert(pac_u == pac_u->sam_u->pac_u);
@@ -312,7 +312,7 @@ _ames_lane_from_sockaddr(struct sockaddr_in* add_u)
 
 /* _ames_lane_into_cache(): put las for who into cache, including timestamp
 */
-void
+static void
 _ames_lane_into_cache(u3p(u3h_root) lax_p, u3_noun who, u3_noun las) {
   struct timeval tim_tv;
   gettimeofday(&tim_tv, 0);
@@ -324,11 +324,11 @@ _ames_lane_into_cache(u3p(u3h_root) lax_p, u3_noun who, u3_noun las) {
 
 /* _ames_lane_from_cache(): retrieve lane for who from cache, if any & fresh
 */
-u3_weak
+static u3_weak
 _ames_lane_from_cache(u3p(u3h_root) lax_p, u3_noun who) {
   u3_weak lac = u3h_git(lax_p, who);
 
-  if (u3_none != lac) {
+  if ( u3_none != lac ) {
     struct timeval tim_tv;
     gettimeofday(&tim_tv, 0);
     u3_noun now = u3_time_in_tv(&tim_tv);
@@ -336,7 +336,7 @@ _ames_lane_from_cache(u3p(u3h_root) lax_p, u3_noun who) {
 
     //  consider entries older than 2 minutes stale, ignore them
     //
-    if (120000 > u3_time_gap_ms(u3k(den), now)) {
+    if ( 120000 > u3_time_gap_ms(u3k(den), now) ) {
       lac = u3k(u3h(lac));
     } else {
       lac = u3_none;
@@ -359,7 +359,7 @@ _ames_serialize_packet(u3_panc* pac_u, c3_o dop_o)
 
   //  update the body's lane, if desired
   //
-  if (c3y == dop_o) {
+  if ( c3y == dop_o ) {
     //  unpack (jam [(unit lane) body])
     //
     u3_noun lon, bod;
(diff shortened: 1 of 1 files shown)


belisarius222:
I'm curious what's using so much CPU.  Do you know if it's the worker or the daemon?

Fang-:
htop lists 100% for the daemon, with the worker varying, sometimes sitting at %0, sometimes spiking up, but usually hovering in the 0-30% range.

belisarius222:
Ok, so the scrying is not the limiting factor.  I guess it's probably the noun manipulation (jam, cue, etc.), but it would be interesting to profile it and check.  Do you have any approximate numbers on throughput, as measured in forwards/second?

Fang-:
> Do you have any approximate numbers on throughput, as measured in forwards/second?

It takes about 110 seconds for ~nus to forward a million packets. In other words, ~9.1k forwards per second. It may or may not be able to go faster, it hasn't been pressured into dropping any packets at least...

belisarius222:
Ok, so roughly 100 microseconds per packet.  Since that does include at least one jam and cue, I wonder how much @joemfb's upcoming work on those will improve performance here.

I'll also note that the packet processing is an embarrassingly parallel operation, so it shouldn't be too hard to load-balance if we wanted more throughput.

philipcmonk:
100 us is better than a 10x improvement, nice job.  And good point about how this can be parallelized, though that's probably easier in king haskell.

ohAitch:
Huh, how is a galaxy ending up responsible for 9k RPS anyway, don't stars do the routing?

pkg/urbit/vere/io/ames.c
```diff
@@ -308,6 +310,43 @@ _ames_lane_from_sockaddr(struct sockaddr_in* add_u)
...
+    if (120000 > u3_time_gap_ms(u3k(den), now)) {
+      lac = u3k(u3h(lac));
+    } else {
+      lac = u3_none;
+    }
+  }
+
+  u3z(who);
+  return lac;
+}
```
joemfb, line 348: I'd make these functions `static`.

pkg/urbit/vere/io/ames.c
```diff
@@ -652,23 +690,42 @@ _ames_lane_scry_cb(void* vod_p, u3_noun nun)
...
     _ames_panc_free(pac_u);
   }
-  //  if there is a lane, forward the packet on it
-  //
-  else if (u3_nul != las) {
-    _ames_forward(pac_u, u3k(las));
+  else {
+    if (0 < pac_u->sam_u->saw_d) {
+      pac_u->sam_u->saw_d--;
+    }
```
joemfb, line 710: Do we want any successful scry to just reset this counter to 0? I had thought this feature was mostly to detect ames namespace mismatches, are you picturing it more as general purpose backoff?

pkg/urbit/vere/io/ames.c
```diff
@@ -771,29 +828,44 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
     //
     //TODO  drop oldest item in forward queue in favor of this one.
     //      ames.c doesn't/shouldn't know about the shape of scry events,
     //      so can't pluck these out of the event queue like it does in
     //      _ames_cap_queue. as such, blocked on u3_lord_peek_cancel or w/e.
     //
-    if ( (1000 < sam_u->foq_d)
-      && !(rec_d[1] == 0 && (256 > rec_d[0])) )
+    if ( (u3_none == lac) && (1000 < sam_u->foq_d) )
     {
```
joemfb, line 852: This opening braces should be on the preceding line now.

pkg/urbit/vere/io/ames.c
```diff
@@ -771,29 +828,44 @@ _ames_recv_cb(uv_udp_t*        wax_u,
...
 
       sam_u->fod_d++;
-      if ( 0 == (sam_u->fod_d % 100000) ) {
+      if ( 0 == (sam_u->fod_d % 10000) ) {
         u3l_log("ames: dropped %" PRIu64 " forwards total\n", sam_u->fod_d);
       }
     }
+    //  if we know there's no lane, drop the packet
+    //
+    else if (u3_nul == lac) {
```
joemfb, line 862: There's some inconsistent conditional styling in this file now. House style is to have whitespace around the conditional expression (ie, `( u3_nul == lac )`).

pkg/urbit/vere/io/ames.c
```diff
@@ -1132,6 +1208,23 @@ u3_ames_io_init(u3_pier* pir_u)
   sam_u->fit_o    = c3n;
   sam_u->foq_d    = 0;
 
+  //NOTE  some numbers on memory usage for the lane cache
+  //
+  //    assuming we store:
+  //    a (list lane) with 1 item, 1+8 + 1 + (6*2) = 22 words
+  //    and a @da as timestamp,                       5 words
```
joemfb, line 1215: I think `@da` should also be 8 words, as it's always 128-bits.

joemfb:
@belisarius222, I think we can boost this quite a bit with jam/cue improvements.

@ohAitch, that remains the goal, but it's not the current behavior.

philipcmonk:
@ohAitch as far as I can tell stars have actually never been a useful part of the routing hierarchy.  This surprised us when we ported the routing logic to new Ames last year.

PR #3406, opened 2020-08-29
urbit/.github: adding new design issue template

urcades:
Creating a baseline structure for design issue-reporting in Landscape.

Primary differences from the OS 1 issue report are as follows:

1. Explicit direction to include images/screenshots in the report
2. Explicit direction to reference the original design if one is the point designer on a given design project
3. Explicit direction to reference an Indigo component's implementation instruction if the observed issue relates to Indigo component mis-implementation

This seems to capture a large chunk of how we could/should be structuring design QA reporting.

### DIFF
--- a/.github/ISSUE_TEMPLATE/design-issue-report.md
+++ b/.github/ISSUE_TEMPLATE/design-issue-report.md
@@ -24,7 +24,9 @@ Example:
 **Expected design behavior**
 A clear and concise description of how you expected the design to behave, look, feel, or be implemented.
 
-If you are a designer involved directly as the point-person of the observed design issue, please include screenshot(s) of the intended design in your description of expected behavior. Link to the associated design if it exists.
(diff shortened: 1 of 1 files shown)


matildepark:
I'd stick this in os1-team (becoming Landscape soon), since it's a product-oriented / feature board, and refer people from the template selection screen to it. I personally treat urbit/urbit as bug-centric, but I'd like to also see it be Arvo specific, we're starting to become a gigantic presence on it.

matildepark:
Poke — can be transferred to urbit/landscape. Probably call it "Report a design issue", I would prefer urbit/urbit to be functionality issues...we can link to it from here.

urcades:
Cool, I can close this PR and migrate the template over.

urcades:
PR has been re-stated in urbit/landscape: https://github.com/urbit/landscape/pull/78

PR #3408, opened 2020-08-31
zuse: allow xml declaration in +de-xml:html

liam-fitzgerald:
Allows for parsing of an xml declaration. Throws it away, as the
datastructure doesn't have anywhere to put the result.

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -6311,9 +6311,10 @@
     |_  ent/_`(map term @t)`[[%apos '\''] ~ ~]
     ::                                                  ::  ++apex:de-xml:html
     ++  apex                                            ::  top level
-      =+  spa=;~(pose comt whit decl)
+      =+  spa=;~(pose comt whit)
       %+  knee  *manx  |.  ~+
-      %+  ifix  [(star spa) (star spa)]
+      %+  ifix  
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/zuse.hoon
```diff
@@ -6311,7 +6311,7 @@
     |_  ent/_`(map term @t)`[[%apos '\''] ~ ~]
     ::                                                  ::  ++apex:de-xml:html
     ++  apex                                            ::  top level
-      =+  spa=;~(pose comt whit)
+      =+  spa=;~(pose comt whit decl)
       %+  knee  *manx  |.  ~+
       %+  ifix  [(star spa) (star spa)]
```
Fang-, line 6316: The XML spec requires declarations appear only as the very first element in a document. Adding it to the "spacing" parser, allowing it to occur multiple times, and even at the end of the document, doesn't seem entirely correct to me.

Considering doing `;~(plug (punt decl) (star spa))` as the prefix parser you give to `ifix`.

Fang- (commented):
Discarding it feels a little bit questionable, since it doesn't let you get the original contents from the resulting noun. Character encoding details could be in there... but are those ever relevant in practice?

We might be fine just ignoring that for now, and addressing it if it ever becomes an issue. I'd make this conform a bit more tightly to the spec though, see below.

Fang- (approved):
This seems fine for now!

PR #3409, opened 2020-08-31
interface: update title eagerly

liam-fitzgerald:
By the decree of @galenwp 

Updates title whilst tab is in the background

cc: @tylershuster this look ok to you?


tylershuster:
Mos def

PR #3414, opened 2020-08-31
chat: cite patps in DM title

liam-fitzgerald:
Shortens comets and moons in the DM title


matildepark:
lol asked and answered

PR #3415, opened 2020-09-01
meta: revise developer onboarding, demarcate landscape apps

matildepark:
- Adds external links to refer Landscape feature requests to urbit/landscape; developer questions to urbit-dev.
- Revises the README at root to explain very succinctly the components that comprise Urbit, which components are in this repo, which are elsewhere.
- Adds a link to Arvo's maintainers in the CONTRIBUTING document. Revises contributing pathways slightly, since we do not at this moment provide maintainers' emails. Maybe we should, but we don't right now.
- Adds a README to pkg/interface with basic information.
- Adds `[landscape]` to all the applications that make up Landscape in `pkg/arvo/app`.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,8 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Landscape feature request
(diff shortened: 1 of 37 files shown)


liam-fitzgerald (approved):
Looks good to me! I wonder whether its worth allocating maintainers for the Landscape apps, such that new contributors know who to ask for different parts of the system?

PR #3418, opened 2020-09-01
landscape: updated subscriptions to scry also

baudtack:
I've made these changes and tested them out and didn't find anything broken. Some of these, s3 for example, have no real subscription that makes sense other than /all. That should probably be changed.

### DIFF
--- a/pkg/interface/src/logic/subscription/global.ts
+++ b/pkg/interface/src/logic/subscription/global.ts
@@ -43,7 +43,8 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
     groups: []
   };
   start() {
-    this.subscribe('/all', 'invite-store');
+    this.scry('/all', 'invite-store');
+    this.subscribe('/updates', 'invite-store');
     this.subscribe('/groups', 'group-store');
     this.subscribe('/primary', 'contact-view');
     this.subscribe('/all', 'metadata-store');
--- a/pkg/interface/src/logic/subscription/global.ts
+++ b/pkg/interface/src/logic/subscription/global.ts
@@ -45,11 +45,23 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
   start() {
     this.scry('/all', 'invite-store');
     this.subscribe('/updates', 'invite-store');
+
+    this.scry('/groups', 'group-store');
     this.subscribe('/groups', 'group-store');
+
     this.subscribe('/primary', 'contact-view');
-    this.subscribe('/all', 'metadata-store');
+
+    this.scry('/associations', 'metadata-store');
+    this.subscribe('/updates', 'metadata-store');
+
+    this.scry('/credentials', 's3-store');
+    this.scry('/configuration', 's3-store');
     this.subscribe('/all', 's3-store');
+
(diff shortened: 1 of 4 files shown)


liam-fitzgerald (changes requested):
The return value of the scry never gets passed to the store? See `getBaseHash()` for how this works

matildepark:
> The subscribe and initial data requests should be synchronous, thus should occur simultaneously. This is not the best place for a performance optimization with scries, unless we do a scheme similar to what graph-store will do:
> 
> Subscribe to /updates, which sends a %keys update upon initial subscribe to "prime" the data store.

Can you clarify this?

pkg/interface/src/logic/subscription/global.ts
```diff
@@ -43,12 +43,25 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
     groups: []
   };
   start() {
-    this.subscribe('/all', 'invite-store');
+    this.scry('/all', 'invite-store');
+    this.subscribe('/updates', 'invite-store');
+
+    this.scry('/groups', 'group-store');
     this.subscribe('/groups', 'group-store');
```
matildepark, line 50: Well, you'll want /updates not /groups here. You're getting /groups from the scry, why make two requests... that is, if group-store has an updates subscription.

pkg/interface/src/logic/subscription/global.ts
```diff
@@ -43,12 +43,25 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
...
+
     this.subscribe('/primary', 'contact-view');
-    this.subscribe('/all', 'metadata-store');
+
+    this.scry('/associations', 'metadata-store');
+    this.subscribe('/updates', 'metadata-store');
+
+    this.scry('/credentials', 's3-store');
+    this.scry('/configuration', 's3-store');
     this.subscribe('/all', 's3-store');
```
matildepark, line 59: We should probably just write an /updates subscription ... or maybe just tell it to scry from s3-store after setting it in `profile`, because it's unlikely to change very often.

pkg/interface/src/logic/subscription/global.ts
```diff
@@ -43,12 +43,25 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
...
+
+    this.scry('/tiles', 'launch');
+    this.scry('/first-time' 'launch');
     this.subscribe('/all', 'launch');
```
matildepark, line 63: Again, `/all` subscriptions undermine the purpose of scrying for this. They return the same data.

baudtack: The only other subscription available in launch is currently /keys which does not appear to include what you actually want when updates happen. But I don't understand any of these apps so maybe it does.

pkg/interface/src/logic/subscription/global.ts
baudtack: It does not currently.

tacryt-socryp:
Let's discuss this in today's midweek check-in.

matildepark:
We ended up deciding against this for the moment but I think we're going to come back to it.

PR #3419, opened 2020-09-01
shoe: tables

Fang-:
I needed this for a thing I'm working on. Figured tables would be a good first "complex screen-draw effect" for /lib/shoe.

Table data is a list of header names, a list of width restrictions, and a list of rows, defined as lists of `dime`s.  
Row contents are rendered as per the aura in the `dime`. This also determines their alignment (left/right), and how they break to fit smaller-width columns.

Note that the comment describes `%table` as "sortable, filterable data". This obviously isn't gonna fly in our current shoe/sole client, but it hints at a not-far-off future where we expose `shoe-effects` directly, letting capable clients render them in more dynamic, context-appropriate ways.

`%row` was added because `%table` necessitated implementing it. Whether it's a good fit for the shoe "protocol" remains to be seen.

The bulk of this is rendering code. A bit nasty, but I hope it's not _too_ bad.

image

I briefly looked into updating chat-cli to use `%row` for its message rendering. That would allow us to remove its line-breaking logic and... not much else, actually. Chat-cli is kind of particular about details here, so refactoring it to use `%row` seems like more trouble than it's worth.

@syzygyzer I feel pretty ambivalent about including these new commands in the example app (and continuing to do so with new additions in the future). On the one hand, it's real code to look at. On the other hand, it adds complexity to the "dumb example" codebase. What do you think?  
Regardless, these commands should be described in the reference documentation, so consider yourself prompted for an update. (;

### DIFF
--- a/pkg/arvo/app/shoe.hoon
+++ b/pkg/arvo/app/shoe.hoon
@@ -6,7 +6,11 @@
 /+  shoe, verb, dbug, default-agent
 |%
 +$  state-0  [%0 ~]
-+$  command  ~
++$  command
+  $?  %demo
+      %row
+      %table
+  ==
 ::
 +$  card  card:shoe
 --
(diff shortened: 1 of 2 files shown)


philipcmonk (approved):
This is great!

PR #3423, opened 2020-09-02
hotfix: prevent joining a host that we already host

liam-fitzgerald:
If a chat is leapt to before joined chats are loaded, then the frontend will attempt to join a chat that we already host.

Fixed in chat-view, lib/pull-hook and the frontend

### DIFF
--- a/pkg/interface/src/views/apps/chat/app.tsx
+++ b/pkg/interface/src/views/apps/chat/app.tsx
@@ -194,6 +194,11 @@ export default class ChatApp extends React.Component<ChatAppProps, {}> {
             render={(props) => {
               let station = `/${props.match.params.ship}/${props.match.params.station}`;
 
+              // ensure we know joined chats
+              if(!chatInitialized) {
+                return null;
+              }
+
             return (
               <Skeleton
                 associations={associations}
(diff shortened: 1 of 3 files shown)


pkg/interface/src/views/apps/chat/app.tsx
```diff
@@ -194,6 +194,11 @@ export default class ChatApp extends React.Component<ChatAppProps, {}> {
             render={(props) => {
               let station = `/${props.match.params.ship}/${props.match.params.station}`;
 
+              // ensure we know joined chats
+              if(!chatInitialized) {
+                return null;
+              }
```
matildepark, line 200: Wouldn't this just blank screen people leaping to an unmanaged chat during first session? If we just no-op the join on back-end do we need this?

liam-fitzgerald: It'll blank screen them until it loads, ensuring the leap still works

matildepark: is chatInitialised tied to a reactive prop?

liam-fitzgerald: yeah

PR #3424, opened 2020-09-02
graph-view: draft of thread architecture

liam-fitzgerald:
This PR adds two threads for graph-store, create & join. Moving forward, I think we should use threads for semantic actions that operate on several hooks and stores. This has several benefits:
- We can annotate the precise failure cases of the threads. This would address urbit/landscape#70 . You can potentially even imagine trying to `|hi` a ship before you join its group/resource and failing if we cannot `|hi` the ship inside some timeout. 
- Easier juggling of card ordering. For example,  to join an unmanaged chat you must first join its group. In app/chat-view this means we have to encode our current join request onto a wire, send the join card on that wire, and then pick up the rest of the chat joining process in +on-agent after it's finished.
- Rollback logic can be encapsulated inside the threads, allowing these transactions to complete or fail atomically. 
- It makes the stores more accessible from the command line, and better defines the API for the applications.

Downsides:
- We still have to write a `app/graph-view` in order to juggle JSON and vase conversions, but the logic should remain in the threads.
- Threads can get interrupted, but most of the threads should be short-lived enough that this isn't a problem in practice. 

Drafting to get feedback before I proceed further into this architecture:
RFC: @Fang- @belisarius222 @philipcmonk @tacryt-socryp


tacryt-socryp:
I think this architecture is excellent. I think with JSON conversion / marks for these actions, there will be no need for a `%graph-view` app.

joemfb:
> Easier juggling of card ordering. For example, to join an unmanaged chat you must first join its group. In app/chat-view this means we have to encode our current join request onto a wire, send the join card on that wire, and then pick up the rest of the chat joining process in +on-agent after it's finished.

I just want to note that, in gall agents, these kind of multi-part transactions are best managed be representing them explicitly in the agent state. The wire then becomes a symbolic reference to that state, and the whole operation becomes much more observable, manageable, interruptible, &c. Of course, the complexity scales geometrically with the number of individual operations involved in the multi-part transaction (and the overall number of multi-part transactions the agent implements), so threads become desirable as the numbers go up.

pkg/arvo/ted/graph/create.hoon
```diff
@@ -0,0 +1,57 @@
...
+::
+=/  =metadata
+  %*  .  *metadata
+    title  title.action
+    description  description.action
+    date-created  now.bowl
+    creator  our.bowl
+  ==
+=/  act=metadata-action
+  [%add group-path [%graph (en-path:resource rid.action)] metadata]
```
tacryt-socryp, line 52: app-name should be passed in as a part of the poke, and will be exactly the same as the existing ones:

`%chat`, `%publish`, `%link`

pkg/arvo/lib/graph-view.hoon
```diff
@@ -0,0 +1,62 @@
...
+      ==
+    ::
+    ++  groupify  !!
+    ++  invite    !!
+    ++  associated
+      %-  of
+      :~  group+dejs:resource
+          policy+policy:dejs:group-store
+      ==
+
```
tacryt-socryp, line 59: blank line

pkg/arvo/lib/graph-view.hoon
```diff
@@ -0,0 +1,62 @@
...
+    ==
+    ::
+    ++  create
+      %-  ot
+      :~  resource+dejs:resource
+          title+so
+          description+so
+          app+app
+          associated+associated
+      ==
```
tacryt-socryp, line 33: `::` in between

pkg/arvo/ted/graph/create.hoon
```diff
@@ -0,0 +1,57 @@
...
+  =/  m  (strand ,resource)
+  ?:  ?=(%& -.associated)
+    (pure:m p.associated)
+  =/  =action:group-store
+    [%add-group rid p.associated %&]
+  ;<  ~  bind:m  (poke-our %group-store %group-action !>(action))
+  ;<  ~  bind:m
+    (poke-our %group-push-hook %push-hook-action !>([%add rid]))
+  (pure:m rid)
+
```
tacryt-socryp, line 19: blank line

pkg/arvo/ted/graph/create.hoon
```diff
@@ -0,0 +1,57 @@
...
+    date-created  now.bowl
+    creator  our.bowl
+  ==
+=/  act=metadata-action
+  [%add group-path [%graph (en-path:resource rid.action)] metadata]
+;<  ~  bind:m  (poke-our %metadata-hook %metadata-action !>(act))
+;<  ~  bind:m
+  (poke-our %metadata-hook %metadata-hook-action !>([%add-owned group-path]))
+(pure:m !>(~))
+
```
tacryt-socryp, line 57: blank line

pkg/arvo/ted/graph/delete.hoon
```diff
@@ -0,0 +1,61 @@
...
+  |=  rid=resource
+  =/  m  (strand ,(unit resource))
+  ;<  pax=(unit path)  bind:m
+    %+  scry:strandio   ,(unit path)
+    ;:  weld
+      /gx/metadata-store/resource/graph
+      (en-path:resource rid)
+      /noun
+    ==
+  (pure:m (bind pax de-path:resource))
```
tacryt-socryp, line 18: `::` in between arms

pkg/arvo/ted/graph/delete.hoon
```diff
@@ -0,0 +1,61 @@
...
+  |=  rid=resource
+  =/  m  (strand ,(unit resource))
+  ;<  ugroup=(unit group)  bind:m
+    %+  scry:strandio   ,(unit group)
+    ;:  weld
+      /gx/group-store/resource/graph
+      (en-path:resource rid)
+      /noun
+    ==
+  (pure:m (need ugroup))
```
tacryt-socryp, line 29: `::` in between arms

pkg/arvo/ted/graph/delete.hoon
```diff
@@ -0,0 +1,61 @@
...
+++  delete-graph
+  |=  rid=resource
+  =/  m  (strand ,~)
+  ;<  ~  bind:m
+    (poke-our %graph-push-hook %push-hook-action [%remove rid.action])
+  ;<  ~  bind:m
+    (poke-our %graph-store %graph-update [%archive-graph rid.action])
+  (pure:m ~)
+--
+^-  thread:spider
```
tacryt-socryp, line 39: `::` in between cast and core

pkg/arvo/ted/graph/delete.hoon
```diff
@@ -0,0 +1,61 @@
...
+?~  ugroup  (fail:m %nonexistent)
+?.  hidden.u.ugroup
+  (delete-graph rid.action)
+;<  ~  bind:m
+  (poke-our %group-push-hook %pull-hook-action [%remove rid.action])
+;<  ~  bind:m
+  (poke-our %group-store %group-action !>([%remove-group rid.action])
+;<  ~  bind:m  (delete-graph rid.action)
+(pure:m !>(~))
+
```
tacryt-socryp, line 61: blank line

pkg/arvo/ted/graph/join.hoon
```diff
@@ -0,0 +1,52 @@
...
+  =/  m  (strand ,(unit resource))
+  ;<  pax=(unit path)  bind:m
+    %+  scry:strandio   ,(unit path)
+    ;:  weld
+      /gx/metadata-store/resource/graph
+      (en-path:resource rid)
+      /noun
+    ==
+  (pure:m (bind pax de-path:resource))
+--
```
tacryt-socryp, line 19: `::` in between

pkg/arvo/ted/graph/join.hoon
```diff
@@ -0,0 +1,52 @@
...
+  %+  poke-our  %group-pull-hook
+  pull-hook-action+!>([%add ship.action rid.action])
+::
+;<  ~  bind:m
+  %+  poke-our  %metadata-hook
+  metadata-hook-action+!>([%add-synced ship.action rid.action])
+;<  ~  bind:m  
+  (poke-our %graph-pull-hook %pull-hook-action !>([%add ship.action rid.action]))
+(pure:m !>(~))
+
```
tacryt-socryp, line 52: blank line

pkg/arvo/ted/graph/leave.hoon
```diff
@@ -0,0 +1,61 @@
...
+  |=  rid=resource
+  =/  m  (strand ,(unit resource))
+  ;<  pax=(unit path)  bind:m
+    %+  scry:strandio   ,(unit path)
+    ;:  weld
+      /gx/metadata-store/resource/graph
+      (en-path:resource rid)
+      /noun
+    ==
+  (pure:m (bind pax de-path:resource))
```
tacryt-socryp, line 18: `::` in between arm declarations

pkg/arvo/ted/graph/leave.hoon
```diff
@@ -0,0 +1,61 @@
...
+  |=  rid=resource
+  =/  m  (strand ,(unit resource))
+  ;<  ugroup=(unit group)  bind:m
+    %+  scry:strandio   ,(unit group)
+    ;:  weld
+      /gx/group-store/resource/graph
+      (en-path:resource rid)
+      /noun
+    ==
+  (pure:m (need ugroup))
```
tacryt-socryp, line 29: `::` in between arm declarations

pkg/arvo/ted/graph/leave.hoon
```diff
@@ -0,0 +1,61 @@
...
+++  delete-graph
+  |=  rid=resource
+  =/  m  (strand ,~)
+  ;<  ~  bind:m
+    (poke-our %graph-pull-hook %pull-hook-action [%remove rid.action])
+  ;<  ~  bind:m
+    (poke-our %graph-store %graph-update [%archive-graph rid.action])
+  (pure:m ~)
+--
+^-  thread:spider
```
tacryt-socryp, line 39: `::` in between core and cast

pkg/arvo/ted/graph/leave.hoon
```diff
@@ -0,0 +1,61 @@
...
+?~  ugroup  (fail:m %nonexistent)
+?.  hidden.u.ugroup
+  (delete-graph rid.action)
+;<  ~  bind:m
+  (poke-our %group-push-hook %pull-hook-action [%remove rid.action])
+;<  ~  bind:m
+  (poke-our %group-store %group-action !>([%remove-group rid.action])
+;<  ~  bind:m  (delete-graph rid.action)
+(pure:m !>(~))
+
```
tacryt-socryp, line 61: blank line

tacryt-socryp (commented):
Looks good. No need to include the hooks.

liam-fitzgerald:
Should be good to go!

philipcmonk:
This all makes sense to me.  One thing to emphasize about threads is that they can hang -- if you really need them to terminate, you need to set a timer.  Honestly, %spider should let you set an optional timeout.  That would simplify %eth-watcher and %glob -- and those are a good template for the feature that should be added to %spider.

Additionally, threads can fail at any time -- for example, if an OTA arrives in the middle that causes %spider to reload, all outstanding threads will be killed.  You absolutely need to make sure that if it fails in the middle it doesn't leave the system in an inconsistent state.  The easiest way to do this is to accumulate all the changes and apply them once at the end, but that's not always possible.

liam-fitzgerald:
superseded by #3479

PR #3432, opened 2020-09-04
links, chat: add drag and drop

tylershuster:
This PR addresses https://github.com/urbit/landscape/issues/74

Current implementation works only for links, looking for feedback before continuing to chat.

[Here's a video](https://marpem-files.sfo2.digitaloceanspaces.com/links-pr.mov) of it working for 

a) autogenerated oembed titles
b) clicking upload to push to S3
c) dragging and dropping to push to S3

@urcades, design thoughts?

@tacryt-socryp When uploading that file (the video), I got the following response:

```
ETagMissing: No access to ETag property on response. Check CORS configuration to expose ETag header.
    uploadPart managed_upload.js:562
    send request.js:369
    callListeners sequential_executor.js:106
    emit sequential_executor.js:78
    emit request.js:688
    transition request.js:22
    runTo state_machine.js:14
    runTo state_machine.js:26
    transition request.js:38
    emit request.js:690
```

Do you know why this happened?


urcades:
This is looking excellent, here's a small bit of feedback for making the drag and drop state of the field more explicit:

Screen Shot 2020-09-04 at 6 42 38 PM

This should be relatively clear, but if you have any questions about the above, let me know — basically we should lighten up the form to our lightest grey background and add a single label making the interactive state hyper-clear.

tylershuster:
@urcades Thanks, that looks great. I made the changes. Also made the field pasteable with the caveats we're discussing in https://github.com/urbit/landscape/issues/75

matildepark:
This works for me except pasting image data straight up, which gives the ETag header error. [This comment](https://github.com/minio/minio/issues/7289#issuecomment-511223212) implies there's another config thing to set? Idk

pkg/interface/src/views/apps/links/components/lib/link-submit.tsx
```diff
@@ -0,0 +1,256 @@
...
+      linkValid = URLparser.test(`http://${linkValue}`);
+      if (linkValid) {
+        linkValue = `http://${linkValue}`;
+      }
+    }
+
+    this.setState({ linkValid, linkValue });
+
+    if (linkValid) {
+      console.log('splarb is valid', linkValue)
```
matildepark, line 82: 'splarb'?

tylershuster: It's a nonsense word I've taken to using that allows you to do leakproof filtering.

tylershuster:
Interesting — how big was the image you tried to paste?

matildepark:
1.5mb iirc

tylershuster:
Hm. I'll look into it. But if it was generally a pleasant experience I'll move onto integrating it into chat

tylershuster:
I added the implementation to chat. You drag into the whole chat window because a) I thought targeting the chat input was a bit of a small target and b) it acted funny anyway.

Basically there's two strange behaviors: 
- The upload icon can be for any files now, so that might need to change
- The aforementioned ETag issue, which I'll investigate now

But it's ready for review otherwise

tylershuster:
Ok, now it accepts image stream pastes into chat, but to prevent someone accidentally pasting something they didn't want and it turning into a kerfuffle, it a) automatically uploads and sends dragged images and b) automatically uploads pasted images and sets the value of the editor to the URL

tylershuster:
Well, I can confirm that the ETag issue exists currently so it's not a regression but if we're going to turn this into a drag-and-drop-into-S3 solution it definitely needs to not fail for large files

tylershuster:
https://github.com/aws/aws-sdk-js/issues/2083

pkg/interface/src/views/apps/links/components/lib/link-submit.tsx
```diff
@@ -0,0 +1,253 @@
...
+                e.preventDefault();
+                this.onClickPost();
+              }
+            }}
+            value={this.state.linkValue}
+          />
+        </div>
+          <input
+            type="text"
+            className="pl2 bg-transparent w-100 f8 pv2 white-d"
```
matildepark, line 214: I don't think `pv2` does anything here

pkg/interface/src/views/apps/links/components/lib/link-submit.tsx
```diff
@@ -0,0 +1,253 @@
...
+               this.s3Uploader.current.inputRef.current.click();
+            }}>upload</span> a file, or paste a link here
+          </span>}
+          {!this.state.disabled ? <S3Upload
+            ref={this.s3Uploader}
+            configuration={this.props.s3.configuration}
+            credentials={this.props.s3.credentials}
+            uploadSuccess={this.uploadSuccess.bind(this)}
+            uploadError={this.uploadError.bind(this)}
+            className="dn absolute pt4 pb2 pl2 w-100"
```
matildepark, line 193:
```suggestion
            className="dn absolute pt3 pb2 pl2 w-100"
```

pkg/interface/src/views/apps/links/components/lib/link-submit.tsx
```diff
@@ -0,0 +1,253 @@
...
+              if (e.key === 'Enter') {
+                e.preventDefault();
+                this.onClickPost();
+              }
+            }}
+            value={this.state.linkTitle}
+          />
+          {!this.state.disabled ? <button
+            className={
+              'bg-transparent f8 flex-shrink-0 pr2 pl2 pt2 pb4 ' + activeClasses
```
matildepark, line 234:
```suggestion
              'bg-transparent f8 flex-shrink-0 pr2 pl2 pt2 pb3 ' + activeClasses
```

matildepark (commented):
Link input is kinda weirdly padded (I think the suggestions here fix that) and otherwise it's good to go

pkg/interface/src/views/apps/links/components/lib/link-submit.tsx
```diff
@@ -0,0 +1,253 @@
...
+      <div
+        className={`relative ba br1 w-100 mb6 ${focus}`}
+        onDragEnter={this.onDragEnter.bind(this)}
+        onDragOver={e => {e.preventDefault();this.setState({ dragover: true})}}
+        onDragLeave={() => this.setState({ dragover: false })}
+        onDrop={this.onDrop}
+      >
+        {this.state.dragover ? <SubmitDragger /> : null}
+        <div className="relative">
+          {(this.state.linkValue || this.state.urlFocus || this.state.disabled) ? null : <span className="gray2 absolute pl2 pt4 pb2 f8" style={{pointerEvents: 'none'}}>
```
matildepark, line 179:
```suggestion
          {(this.state.linkValue || this.state.urlFocus || this.state.disabled) ? null : <span className="gray2 absolute pl2 pt3 pb2 f8" style={{pointerEvents: 'none'}}>
```

Dropped this one sorry

pkg/interface/src/views/apps/links/components/lib/link-submit.js
```diff
@@ -1,153 +0,0 @@
...
-      /((?:([\w\d\.-]+)\:\/\/?){1}(?:(www)\.?){0,1}(((?:[\w\d-]+\.)*)([\w\d-]+\.[\w\d]+))){1}(?:\:(\d+)){0,1}((\/(?:(?:[^\/\s\?]+\/)*))(?:([^\?\/\s#]+?(?:.[^\?\s]+){0,1}){0,1}(?:\?([^\s#]+)){0,1})){0,1}(?:#([^#\s]+)){0,1}/
-    );
-
-    const validURL = URLparser.exec(link);
-
-    if (!validURL) {
-      const checkProtocol = URLparser.exec('http://' + link);
-      if (checkProtocol) {
-        this.setState({ linkValid: true });
-        this.setState({ linkValue: 'http://' + link });
```
matildepark, line 48: You dropped this logic, now if you type `4chan.org` it doesn't prepend `http://` to it. We want it agnostic to URI schema (gopher:// etc should work); but as it is in the PR right now 4chan has no URI schema at all and the link opens in the same tab as `http://localhost:9000/~link/4chan.org`.

tylershuster: Hm good point. I don't know how it got dropped.

tylershuster: Ok, I remember why I did this. I changed the regex because a URL like https://www.youtube.com/watch?v=ADiDaOVTixs&feature=youtu.be doesn't get registered and that's what I was testing. I just changed it back to what you had the regex as before and we can deal with the problem separately

matildepark (approved):

pkg/interface/src/views/apps/links/components/lib/link-submit.js
matildepark: O_o weird

matildepark:
I'll wait for CI here and then merge, brb

PR #3433, opened 2020-09-04
u3: rewrites +stir jet, eliminating recursion and optimizing allocations

joemfb:
Stack overflow in `+stir` (when parsing large inputs with `+star`) has been blocking #1892. For clarity, I'll PR a similar rewrite of the hoon separately.

The first commit implements the traversal I posted here: https://github.com/urbit/urbit/pull/1892#issuecomment-673680540. In the second, I realized that we could save all but the last of the accumulator cell allocations in the left-fold. And the optimizations in the last commit are inspired by #3355.

/cc @frodwith


joemfb:
I've further updated this PR with a new implementation of the road-stack API, factored to avoid north/south branches and (possibly) redundant overflow checks.

joemfb:
@belisarius222, bump for review.

belisarius222 (approved):
LGTM

PR #3435, opened 2020-09-05
links, publish: display rich text comments

tylershuster:
Fixes https://github.com/urbit/landscape/issues/77

Deduplicates some of the message parsing logic.

Does not fix some of the fragmentation/non-indigoness between apps, but should be just as easy to fix later


matildepark:
This is simpler than I thought it would be

matildepark (approved):
Looks quite nice on livenet, helps componentise quite a bit ... soon we should componentise comments across Links and Publish too but you get the idea.

tylershuster:
Yes, I was thinking once the store is isomorphic we can standardize a component in indigo

PR #3436, opened 2020-09-05
term: move escape sequence generation out of dill

Fang-:
This is a first step towards command line ~~r~~evolution.

Presently, when rendering styled text, dill inserts ansi escape sequences before passing it on to term.c. There's no reason for dill to know about such earthly customs though, so here we move the "styled text noun" -> "string with escape sequences" conversion into term.c.

Drum made some assumptions around escape codes to specify "true" cursor position, adjusting for escape characters that might have gotten inserted. That, too, has now been moved into the runtime, letting drum only concern itself with the semantic/visible cursor position.

Note that you can use the new term.c just fine with old dill/drum, but the reverse won't work. Concretely, styled text (this includes, always, the prompt) will render as blank lines instead. We'll likely want to push out the vere changes early, then wait a good amount of time before releasing the OTA that starts making use of them. (Alternatively, version negotiation?)

I should probably look into making a similar change in king haskell...


pkg/urbit/vere/io/term.c
```diff
@@ -1036,6 +1042,169 @@ u3_term_ef_ctlc(void)
   _term_it_refresh_line(uty_u);
 }
 
+/* _term_it_put_color(): put ansi color id on lin_w. RETAINS col.
+*/
+static void
+_term_it_put_color(c3_w* lin_w,
```
joemfb, line 1048: I think this should use the hoon type name (`_term_it_put_tint()`).

pkg/urbit/vere/io/term.c
```diff
@@ -1036,6 +1042,169 @@ u3_term_ef_ctlc(void)
...
+                   u3_noun col)
+{
+  switch ( col ) {
+    default:
```
joemfb, line 1052: I think you can just remove the default case here.

pkg/urbit/vere/io/term.c
```diff
@@ -1036,6 +1042,169 @@ u3_term_ef_ctlc(void)
...
+                        u3_noun dec)
+{
+  switch ( dec ) {
+    default:
+    case u3_nul: *lin_w = '0'; break;
+    case c3__br: *lin_w = '1'; break;
+    case c3__un: *lin_w = '4'; break;
+    case c3__bl: *lin_w = '5'; break;
+  }
+}
```
joemfb, line 1078: `_term_it_put_deco()`, default case.

pkg/urbit/vere/io/term.c
```diff
@@ -1036,6 +1042,169 @@ u3_term_ef_ctlc(void)
...
+    }
+  }
+
+  //  allocate enough memory for every display character, plus styles
+  //
+  //NOTE  we use max 20 characters per styl for escape codes:
+  //      3 for opening, 4 for decorations, 4 for colors, 4 for closing,
+  //      and 5 as separators between decorations and colors.
+  //
+  c3_w* lin_w = c3_malloc( 4 * (lec_w + (20 * tuc_w)) );
```
joemfb, line 1106: For clarity, I think this should be `sizeof(c3_w) * ...`, here and in the `c3__lin` case of `_term_ef_blit()`.

pkg/urbit/vere/io/term.c
```diff
@@ -1036,6 +1042,169 @@ u3_term_ef_ctlc(void)
...
+  c3_w sap_w = 0;
+  {
+    u3_noun nub = tub;
+    while ( u3_nul != nub ) {
+      u3_noun nib   = u3t(u3h(nub));
+      u3_noun tyl   = u3h(u3h(nub));
+      u3_noun dec   = u3h(tyl);
+      u3_noun bag   = u3h(u3t(tyl));
+      u3_noun fog   = u3t(u3t(tyl));
+      c3_o    tyl_o = c3n;
```
joemfb, line 1121: I think this would be clearer as

```
u3_noun tyl, nib, dec, bag, fog;
u3x_cell(u3h(nub), &tyl, &nib);
u3x_trel(tyl, &dec, &bag, &fog);
```

And the horizontal spacing can be reduced.

pkg/urbit/vere/io/term.c
```diff
@@ -1036,6 +1042,169 @@ u3_term_ef_ctlc(void)
...
+      if ( (u3_nul != dec) || (u3_nul != bag) || (u3_nul != fog) ) {
+        tyl_o = c3y;
+      }
+
+      //  write style escape sequences
+      //
+      if ( c3y == tyl_o )
+      {
```
joemfb, line 1129: brace on same line as conditional.

joemfb (changes requested):
This looks great! It's all really nice and clear: much simpler than I expected it could be.

Since you've structured this with backwards compatibility, I think the changes to dill should go into a separate PR. The vere changes can go out ASAP (as can future changes to king-haskell).

Ideally we could notify users to upgrade the runtime via version negotiation, but it's probably not the worse thing to just a) wait, and then b) tell any stragglers to upgrade when they ask why their console isn't working. (Note that it won't crash or anything, it'll just not display the prompt properly.)

Some style nits below:

pkg/urbit/vere/io/term.c
Fang-: From a hoon types perspective, sure. But C can't enforce those, and I thought it was good practice to always have a default in `switch`es. It's more robust against future `+tint` changes, too; implementation assumes this function's gonna put a character there, leaving that byte of memory un-set seems not good.

joemfb: You're right, fall-thru is good behavior here. Not sure what I was thinking ...

PR #3439, opened 2020-09-06
king: change logging from tracing to info in Wai.hs

eglaysher:
I noticed that the king's text log file kept filling up, and it's
mostly 'sending chunk "\n"' from eyre. This changes the log level
of every partial send over an active eyre channel to info, instead
of always.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
@@ -179,7 +179,7 @@ streamBlocks env init getAct = send init >> loop
 
   send "" = pure ()
   send c  = do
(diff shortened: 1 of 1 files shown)


brendanhay:
When you say it will 'always' log referring to `logTrace` and replace it with `logInfo` - doesn't that seem a little inverted? Wouldn't tracing information be ignored by running with a less verbose logging level? For example, if I was to run this how would I turn logging of partial chunks on - or off?

It just seems a little strange to log partial chunks at [LogInfo](https://hackage.haskell.org/package/rio-0.1.18.0/docs/RIO.html#t:LogLevel) - our logging seems a little wonky in general.

brendanhay (approved):

eglaysher:
logTrace is actually higher than logError because of how the sum type is ordered, because logTrace is an alias around custom types. I agree the details of our logging system are wonky.

PR #3441, opened 2020-09-07
dbug: attempt to scry for state

Fang-:
If the underlying app implements a `/x/dbug/state` endpoint in `+on-peek`,
dbug will now use that for `%state` evaluation. Falls back to the vase
provided by `+on-save` if the peek fails.

This allows apps and (perhaps more usefully) wrapper agents to provide
customized vases to `/lib/dbug`.

Updates `/lib/shoe` to make use of this, properly prepending the wrapped
app's vase to shoe's own, instead of including it as-is.


philipcmonk (approved):
This is a good feature.

PR #3442, opened 2020-09-07
hoon: avoid potential shadow in +late

Fang-:
Here's a fun one. Follow-up to #3266.

Previously:

```hoon
> ((late 42) *(list))
mull-grow
-find.t
find-fork
```

By using tree instead of face addressing, we avoid stepping into faces from the
argument noun.

Updates `+lead` for visual symmetry, and the much rarer `-find.h` case.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -262,8 +262,8 @@
 ++  tail  |*(^ ,:+<+)                                   ::  get tail
 ++  test  |=(^ =(+<- +<+))                              ::  equality
 ::
-++  lead  |*(h=* |*(* [+>+< +<]))                       ::  put head
-++  late  |*(t=* |*(* [+< +>+<]))                       ::  put tail
+++  lead  |*(* |*(* [+<.^$ +<]))                          ::  put head
+++  late  |*(* |*(* [+< +<.^$]))                          ::  put tail
 ::
 ::  #  %containers
 ::
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -262,8 +262,8 @@
 ++  tail  |*(^ ,:+<+)                                   ::  get tail
 ++  test  |=(^ =(+<- +<+))                              ::  equality
 ::
-++  lead  |*(* |*(* [+<.^$ +<]))                          ::  put head
-++  late  |*(* |*(* [+< +<.^$]))                          ::  put tail
+++  lead  |*(* |*(* [+>+< +<]))                          ::  put head
+++  late  |*(* |*(* [+< +>+<]))                          ::  put tail
 ::
 ::  #  %containers
 ::
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -262,8 +262,8 @@
 ++  tail  |*(^ ,:+<+)                                   ::  get tail
 ++  test  |=(^ =(+<- +<+))                              ::  equality
 ::
-++  lead  |*(h=* |*(* [h +<]))                          ::  put head
-++  late  |*(t=* |*(* [+< t]))                          ::  put tail
+++  lead  |*(h=* |*(* [+>+< +<]))                       ::  put head
+++  late  |*(t=* |*(* [+< +>+<]))                       ::  put tail
 ::
 ::  #  %containers
 ::


pkg/arvo/sys/hoon.hoon
```diff
@@ -262,8 +262,8 @@
 ++  tail  |*(^ ,:+<+)                                   ::  get tail
 ++  test  |=(^ =(+<- +<+))                              ::  equality
 ::
-++  lead  |*(h=* |*(* [h +<]))                          ::  put head
-++  late  |*(t=* |*(* [+< t]))                          ::  put tail
+++  lead  |*(h=* |*(* [+>+< +<]))                       ::  put head
+++  late  |*(t=* |*(* [+< +>+<]))                       ::  put tail
```
joemfb, line 266: IMO, you should also remove the `h`/`t` faces from the sample of the outer gate. And it might be clearer to use `+<.^$` over `+>+<`. (Not sure about that last part ...)

Fang-: Good call on removing the faces. To my mild surprise, your stylistic suggestion isn't purely stylistic though:

```hoon
> ? ((lead-lark 2) 3)
  {@ud @ud}
[2 3]
> ? ((lead-ktbc 2) 3)
  {* @ud}
[2 3]
```

joemfb: Wow, that's pretty bad.

```
> `@`(|*(* +<) 1)
1
> `@`(|*(* +<.$) 1)
nest-fail
```

joemfb: Even worse:

```
> `@`(|*(@ +<.$) 1)
1
> `@`(|*(@ +<.$) [1 1])
<<unlinks-dojo>>
```

joemfb:
@Fang-, does this PR need to be reverted to its earlier state (ie, 8531c83005e109532897c8f2be24ac92f68408b7)?

Fang-:
8531c83 wasn't quite right either. Should be good now though!

pilfer-pandex (approved):
Do you have a model for why putting faces on things breaks wet gates?

Fang-:
> Do you have a model for why putting faces on things breaks wet gates?

I'm not gonna have the best answer for you, so ask around, but I intuit this as being "type information falls into the wet gate and overrides/shadows locally defined values".

pilfer-pandex:
> so ask around

No one that I've asked so far at the company has a good model of how wetness actually works, which is concerning. Perhaps, @ohAitch: do you know?

belisarius222:
I don't think "putting faces on things breaks wet gates" is exactly right.  Plenty of wet gates have faces on their samples, and the issue with using faces in `+lead` is not a bug, but rather expected, if maybe nonintuitive, behavior.  There was also a bug discovered with using `+>.^$` nest-failed -- that's a [separate issue](https://github.com/urbit/urbit/issues/3600), which @pkova seems to have just fixed.

The issue with using the 'h' face in the sample in +lead is that wet gates use the type of the call-site sample to resolve its faces.  So if +lead uses 'h' as the face of the sample in its outer gate like this,
```
++  lead  |*(h=* |*(* [h +<]))
```
then if the inner gate's sample has an `h` face in it, that inner sample's face will get resolved there.  This is a footgun, but not a bug.  +mull checks that every face that the wet gate accesses resolves to the same axis for the call-site sample type as for the default sample type.

That check is needed so that wet gates can be jetted (either that, or you'd basically need to pass in an interface core instead of raw data).  It's sort of like row polymorphism on the sample type, in that you can call a wet gate with a sample containing any set of unused faces at any axes, as long as the faces that the wet gate accesses resolve to the same axes as in the default sample.

ohAitch:
Oh dear. Secrets held between czyar' and zod.

This case is fairly straightforward, you're trying to compile `|=((list) [+< t]))` and that's a find-fork.

More generally you either want faces on both the `|*`s or neither of them: otherwise you get `|*(t=* |*(* [+< t]))` compiled as [[1 [[0 0b110] [0 0b1.1110]]] [1 0] [0 1]] but then it gets specialized(I'm futzing with the evaluation order some) as `|=(t=@ |=((list) [+< t]))` i.e. `|=(t=@ |=((list) [+< t.+<]))` compiled as [[1 [[0 0b110] [0 0b1**011**]]] [1 0] [0 1]], leading to a wet shape mismatch. Alternately you could go for `|*(t=* |*(* [+< t.+>]))`

…except actually `((|*(t=* |*(* [+< t])) 2) t=1)` produces `[t=1 2]` just fine, so no I *don't* know what's going on, and someone should perhaps read through the relevant compiler passes enough to get an actual gears model.

ohAitch:
Ted coming in with the far clearer explanation! Do *you* know why `((lead 1) h=2)` `((lead 1) [h=2 3])` etc do compile?

belisarius222:
No, I'm also surprised by your `[t=1 2]` result.  I would have expected `[t=1 1]`.  I ran it for myself to be sure, though, and got the same thing.  Maybe it's something similar to #3600  ?

ohAitch:
I mean you *definitely* shouldn't get `[t=1 1]`, there's no code to actually `[+< +<]`(in the physical core vs wet-checking compile of the gate). But one would hope the specialized one is in fact checked and emits an error.

belisarius222:
Oh yeah, the +mull check should fail, since the `t` is in a different place — unless the `t` somehow gets resolved when the outer gate compiles, before +mull checks the inner gate's call site compatibility?

—
~rovnys-ricfer
https://urbit.org

ohAitch:
But then it wouldn't complain about the list either!

PR #3443, opened 2020-09-07
groups: add group deletion confirmation

tylershuster:
Addresses https://github.com/urbit/landscape/issues/68

Doesn't use indigo yet. the `destructive` prop doesn't quite work as intended and figured it would be better to get a fix that worked fine out rather than have people accidentally delete groups in the meantime


matildepark:
It probably works in 1.2 but we don't use 1.2 yet.

tylershuster:
Should I close or wait?

matildepark:
We can merge as is, just gotta test.

matildepark (approved):
Does what it says

PR #3444, opened 2020-09-08
publish: don't timeout, show groupify

liam-fitzgerald:
- Does not timeout publish join flow
- Exposes groupify in settings UI


matildepark:
Delete notebook should go at bottom of the list.

Groupify doesn't let you choose the group it's added to? (see Chat, Links for ex.)

matildepark (approved):

PR #3445, opened 2020-09-08
chat: added ability to blur chat editor

tylershuster:
Addresses https://github.com/urbit/landscape/issues/82


matildepark (approved):

matildepark:
btw you can do "Fixes urbit/landscape#82" in the commit and it will work fine, close the issue when it hits master, etc.

PR #3446, opened 2020-09-08
publish: first draft of graph-store migration

liam-fitzgerald:
State adapter to migrate publish to graph-store.

Just want to check that I understand the graph-store API correctly.

Open questions:
- How should titles be set on notes? We can't attach a metadata entry to a particular graph node. Tempted to just parse the metadata out of the text contents (this is what publish already does). Alternatively, we could put a metadata entry into contents and use that? If we opt for parsing the metadata, should we check the validity inside the publish mark?

- I'm leaving the state untouched, so that a) we don't have any irretrievable data loss and b) we can port read indicators forward when notification-store is shipped.

Pending:
- Graph hook implementation

### DIFF
--- a/pkg/arvo/mar/graph/validator/publish.hoon
+++ b/pkg/arvo/mar/graph/validator/publish.hoon
@@ -9,8 +9,15 @@
   ++  noun
     |=  p=*
     =/  ip  ;;(indexed-post p)
-    ?>  (gte 2 (lent index.p.ip))
-    ip
+    |^
+    ?+    index.p.ip  !!
+      [@ @ ~]  ip
+    ::  top level post must have title as first content
+        [@ ~]
+      ?>  ?=(^ contents.p.ip)
+      ?>  ?=(%text -.i.contents.p.ip)
+      ip
+    ==
   --
 ::
 ++  grad  %noun
--- a/pkg/arvo/mar/graph/validator/publish.hoon
+++ b/pkg/arvo/mar/graph/validator/publish.hoon
@@ -1,0 +1,17 @@
+/-  *post
+|_  i=indexed-post
+++  grow
+  |%
+  ++  noun  i
+  --
+++  grab
+  |%
+  ++  noun
+    |=  p=*
+    =/  ip  ;;(indexed-post p)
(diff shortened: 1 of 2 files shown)


tacryt-socryp:
Keep in mind that contents within a post is a list, not a singular entry. I think the best thing to do is make contents have two %text entries: the first is the title, the second is the body.

liam-fitzgerald:
> Keep in mind that contents within a post is a list, not a singular entry. I think the best thing to do is make contents of a "publish note" (toplevel node in the graph) have two %text entries: the first is the title, the second is the body.

What about other metadata, like whether comments are enabled?

tacryt-socryp:
I'm okay with this being configured per notebook. perhaps we add an optional key-value map of data of types %boolean, %integer, %date... to metadata-store.

pkg/arvo/mar/graph/validator/publish.hoon
```diff
@@ -9,8 +9,15 @@
...
     =/  ip  ;;(indexed-post p)
-    ?>  (gte 2 (lent index.p.ip))
-    ip
+    |^
+    ?+    index.p.ip  !!
+      [@ @ ~]  ip
+    ::  top level post must have title as first content
+        [@ ~]
+      ?>  ?=(^ contents.p.ip)
+      ?>  ?=(%text -.i.contents.p.ip)
```
tacryt-socryp, line 18: I'm curious: do you think that contents should be more heavily constrained, specifically as [%text %text]~? It is currently not especially constrained: the list can be of any length, and can have any type in the subsequent list items.

liam-fitzgerald: We have to be able to render the other content types, so I don't think it really makes much of a difference from a development perspective, and it seems unnecessary to limit publish like this. I'd love to see publish as a jupyter style notebook for hoon and I'm sure there will be other usecases that I haven't thought of

liam-fitzgerald:
ready to go into release/graph-publish

liam-fitzgerald:
cc: @tacryt-socryp

tacryt-socryp:
This is obsoleted by #3473.

Issue #3383, opened 2020-08-27
settings: don’t work on mobile


matildepark:
The sidebar appears, but the items do nothing. If this is written as an onClick then it may not process touch events?

PR #3448, opened 2020-09-08
profile: fix iOS safari behaviour, add react-helmet

matildepark:
- The box for the nav was buried under some flexboxes. Adds z-index to fix behaviour on iOS Safari, profile orientation (it worked in landscape?). (#3383)
- Adds react-helmet to the routes since it appeared just before merge.


liam-fitzgerald (approved):
LGTM!

PR #3449, opened 2020-09-08
chat: check for cited participants in sidebar

### LINKED ISSUE #3438: chat: cite ships in sidebar
https://github.com/urbit/urbit/commit/11b913b7468ea4fedd3f2c08939cae0d31bdb311

tylershuster:
Fixes #3438

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/group-item.js
+++ b/pkg/interface/src/views/apps/chat/components/lib/group-item.js
@@ -1,13 +1,13 @@
 import React, { Component } from 'react';
 import { Link } from 'react-router-dom';
 import { ChannelItem } from './channel-item';
-import { deSig } from "~/logic/lib/util";
+import { deSig, cite } from "~/logic/lib/util";
 
 export class GroupItem extends Component {
   render() {
     const { props } = this;
     const association = props.association ? props.association : {};
-    const DEFAULT_TITLE_REGEX = new RegExp(`(( <-> )?~${deSig(window.ship)}( <-> )?)`);
+    const DEFAULT_TITLE_REGEX = new RegExp(`(( <-> )?~(?:${window.ship}|${deSig(cite(window.ship))})( <-> )?)`);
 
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/apps/chat/components/lib/group-item.js
```diff
@@ -1,13 +1,13 @@
...
 import { ChannelItem } from './channel-item';
-import { deSig } from "~/logic/lib/util";
+import { deSig, cite } from "~/logic/lib/util";
 
 export class GroupItem extends Component {
   render() {
     const { props } = this;
     const association = props.association ? props.association : {};
-    const DEFAULT_TITLE_REGEX = new RegExp(`(( <-> )?~${deSig(window.ship)}( <-> )?)`);
+    const DEFAULT_TITLE_REGEX = new RegExp(`(( <-> )?~(?:${deSig(window.ship)}|${deSig(cite(window.ship))})( <-> )?)`);
```
matildepark, line 10: `cite` deSigs already iirc

matildepark: it returns a sig'd ship no matter what

matildepark: ... well now I'm confused

tylershuster: I don't know but I tested it on my planet and on a comet that had previously not exhibited correct behavior and now they both do

matildepark: so: 
- `window.ship` doesn't have a sig so `deSig` isnt really necessary
- `cite` will take a sig'd or desig'd ship and return a sig'd ship
- the DM format *does* use sigs on both ships

so maybe just remove `deSig` from window.ship and it will make more sense? since we want to force the sig in both contexts and we have to desig to have the start of the regex apply to the query regardless

matildepark (approved):

PR #3451, opened 2020-09-08
leap: use a instead of Link, if not default app

matildepark:
Logging out currently doesn't work via Leap because we only use
react-router. This adds selection and navigation logic that passes
both app name and link href and compares app name against the default
app lib. This makes leap extensible for third party apps, too.


liam-fitzgerald (approved):
LGTM!

PR #3452, opened 2020-09-08
error: add background style

matildepark:
If you've set an image background, it can make it hard to read 404s or
error messages. This adds the default bg color.


liam-fitzgerald (approved):
LGTM!

PR #3453, opened 2020-09-08
interface: embeds on by default

matildepark:
By @vvisigoth request.

cc @tylershuster: this isn't set anywhere else, is it?


tylershuster:
No...but...uh...this is almost-arbitrary script injection

WARNED - LORAX

vvisigoth:
TBC this isn't a commentary on the righteousness of opt-out embeds, it's
just going to be interpreted as a regression if something that previously
worked doesn't anymore.
—
~poldec-tonteg
http://urbit.org

tylershuster:
Ok, fair enough. What could be worse than allowing Google in, which is already the case?

liam-fitzgerald (approved):
LGTM!

PR #3454, opened 2020-09-08
chat: prevent message rerender

tylershuster:
This makes `ChatMessage`  a `PureComponent` so that it won't rerender if a shallow comparison of its props determines that they are equal. Previously a rerender of `ChatWindow` (if new messages came in) would force a rerender of all its children.


matildepark (approved):
Fixes the issue of embeds collapsing on mobile (and apparently on some desktops? idk)

PR #3455, opened 2020-09-08
interface: always show home button

matildepark:
This was to go out with the last feature OTA, but regressed in #3236, so it's reinstated here.


liam-fitzgerald (approved):
LGTM!

PR #3456, opened 2020-09-08
all: added fullscreen button

tylershuster:
RFC. I always have urbit open and on a mac it's nice to have something take up the full 'space' or whatever they're called. Feels nice.

### DIFF
--- a/pkg/interface/src/views/components/StatusBar.js
+++ b/pkg/interface/src/views/components/StatusBar.js
@@ -74,6 +74,16 @@ const StatusBar = (props) => {
           <Sigil ship={props.ship} size={24} color={"#000000"} classes="dib mix-blend-diff" />
           <Text ml={2} display={["none", "inline"]} fontFamily="mono">{props.ship}</Text>
         </StatusBarItem>
+        {'requestFullscreen' in document.documentElement ? <StatusBarItem ml={2} onClick={() => {
+          if (document.fullscreenElement) {
+            document.exitFullscreen();
+            document.fullscreenElement = null;
+          } else {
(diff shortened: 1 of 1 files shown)


matildepark:
I don't think it's really necessary to have in the status bar, honestly?

tylershuster:
Think a hotkey would be better?

galenwp:
Isn't this normally done by making the browser fullscreen?

tylershuster:
You still get browser chrome, at least in Firefox

matildepark:
That's all this does — it's a button in the header bar to tell the browser to full screen.

tylershuster:
Not quite...make your browser full screen versus pasting this in:
`document.addEventListener('click', () => document.documentElement.requestFullscreen())`
and then clicking on the page.

tylershuster:
(I am not attached to this, but I do think it's neat)

galenwp:
In chrome there's no chrome. I guess I'd have to see it. I'm always trying to remove as many buttons as possible. @urcades can also weigh in.

matildepark:
pinging @urcades who just got back

urcades:
I did some digging and it turns out this is a specifically annoying firefox trait while full-screening the browser. I still need to test Edge browser, but currently i'm witnessing Safari and Chrome both initialize chrome-less fullscreen as desired. 

I'd be cautious to introduce another non-visual (keyboard shortcut triggered)  feature specifically tailored to one browser, but I'd prefer we resist implementing new buttons/interface for a feature that should really be handled by the browser/OS.

tylershuster:
Alright, Firefox maximalists strike out again

PR #3457, opened 2020-09-09
chat: show pending messages again

tylershuster:
A side effect of #3357 was that pending messages didn't show up properly because, having no `number` attribute, the smarter system didn't know where to render them. This PR sorts the pending messages by timestamp and appends them to the list of messages to render

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
+++ b/pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
@@ -232,10 +232,13 @@ export class ChatWindow extends Component<ChatWindowProps, ChatWindowState> {
     const debouncedFetch = _.debounce(this.fetchMessages, 500).bind(this);
 
     envelopes
-      .concat(stationPendingMessages)
       .forEach((message) => {
         messages[message.number] = message;
       });
+    
+    stationPendingMessages.sort((a, b) => a.when - b.when).forEach((message, index) => {
+      messages[mailboxSize + index + 1] = message;
+    });
     
     return (
       <Fragment>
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (approved):
LGTM!

matildepark:
Sep-08-2020 20-09-37

tylershuster:
Hm. I had that issue but then I thought I fixed it

tylershuster:
Hm...can't replicate.

matildepark:
I can, in any chat ... using Brave ofc. Seems like the numbering eats the latest message's number until it comes in and then it sorts out. Are you sending to yourself? It would be too fast to notice

Sep-08-2020 20-16-33

tylershuster:
I'm using my comet (and then I turned it off so it *really* can't resolve). Looks like it might just be an off-by-one problem

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -232,11 +232,14 @@ export class ChatWindow extends Component<ChatWindowProps, ChatWindowState> {
     const debouncedFetch = _.debounce(this.fetchMessages, 500).bind(this);
 
     envelopes
-      .concat(stationPendingMessages)
       .forEach((message) => {
         messages[message.number] = message;
       });
     
+    stationPendingMessages.sort((a, b) => a.when - b.when).forEach((message, index) => {
+      messages[mailboxSize + index] = message;
```
matildepark, line 240: I mean, indexes start at 0. So it'll start at the latest message.

tylershuster: Damn, I'm sure that's it.

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -250,7 +253,7 @@ export class ChatWindow extends Component<ChatWindowProps, ChatWindowState> {
         {messages.length ? <VirtualList
           ref={this.virtualList}
           style={{height: '100%', width: '100%', visibility: this.state.initialized ? 'initial' : 'hidden'}}
-          totalCount={mailboxSize}
+          totalCount={mailboxSize + stationPendingMessages.length}
```
matildepark, line 256: Is this being passed to the messages inside? The last non-pending message has `pb3` when the pending should have `pb3`.

tylershuster: good point

matildepark (approved):

PR #3458, opened 2020-09-09
graph-store: fix signatures & marks

liam-fitzgerald:
Fix signing nodes with graph-store


liam-fitzgerald:
ping: @tacryt-socryp

tacryt-socryp:
What was the issue before?

liam-fitzgerald:
a) `/=app=` syntax doesn't work on ford fusion
b) the mold for the foreign deed scry is incorrect it's a `[life pass (unit @ux)]` not a unit of the same
c) said deed scry crashes if we don't have the users key, necessitating scrying for their lyfe (unitized life) to check whether we have their key

tacryt-socryp:
Sounds good. I had only tested it for local deeds.

tacryt-socryp (approved):

PR #3460, opened 2020-09-09
leap: safety check state.selected

matildepark:
- Selected checks were crashing leap if no results were available. This just checks more safely.


liam-fitzgerald (approved):
LGTM!

PR #3461, opened 2020-09-09
chat: hotfix new message indicator

tylershuster:
Doesn't show the "new message" indicators if the chat isn't idle. The sidebar still shows as unread briefly, but I don't know that this is a problem with the scroll implementation...


tylershuster:
@liam-fitzgerald @matildepark

matildepark:
Will test in a bit

matildepark (approved):
I am satisfied

PR #3467, opened 2020-09-09
vere: support rendering +stub as ansi escape codes

Fang-:
Adds support to `term.c` for a `%klr` blit, containing a +stub describing
styled text.

Dill will start making use of this in a separate PR (#3468), for release
cutting reasons.

Supersedes #3436. See that for prior discussion. @joemfb forgive me for squashing review corrections into this one commit. All your comments should have been addressed, in any case.


joemfb (approved):
LGTM

PR #3468, opened 2020-09-09
dill: send styled text to runtime as-is

Fang-:
Depending on the additions to term.c made in #3467 allows dill to
forget about ansi escape codes, and pass styled text nouns straight on
to vere.

Also removes a bit of logic from drum, which assumed things about the
rendering of escape codes to adjust cursor positioning. Now it simply
states the semantic cursor position, letting the runtime deal with the
potential influence of styling.

Supersedes #3436, containing just the dill changes from there, unmodified.

Targeting `release/next-dill`, which shouldn't be pushed out OTA until #3467 (and the matching king haskell change) has been out in the wild for a while. Deploying this to older clients will make their command line prompts and some output lines fail to render.

### DIFF
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -551,7 +551,6 @@
 ++  se-show                                           ::  show buffer, raw
   |=  lin/(pair @ud stub)
   ^+  +>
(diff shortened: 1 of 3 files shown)


joemfb (approved):
LGTM

PR #3470, opened 2020-09-10
Use logInfo in place of logDebug in most situations

pilfer-pandex:
The previously mostly unused "info" level sits between debug and warn. I've moved most things, especially startup trace messages, from debug to info. The idea is if you run with --stderr --log-info, you should get as chatty a ship as possible while still having it be usable. This means in particular that you don't get printouts on every event or http exchange. This would have made my debugging session with @ixv yesterday way better.

One problem remains, which is that on startup you get a bunch of zig-zag as multiple different threads log at the same time. This is a deep issue with RIO logging, which simply fails to guard the egress of logs in a manner appropriate for a concurrent environment.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
@@ -194,9 +194,9 @@ tryBootFromPill oExit pill lite ship boot = do
  where
   bootedPier vSlog = do
     view pierPathL >>= lockFile
-    rio $ logDebug "Starting boot"
+    rio $ logInfo "Starting boot"
     sls <- Pier.booted vSlog pill lite ship boot
-    rio $ logDebug "Completed boot"
+    rio $ logInfo "Completed boot"
     pure sls
 
 runOrExitImmediately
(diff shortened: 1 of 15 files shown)


pilfer-pandex:
@brendanhay are these ci/hercules things yours? What do I need to do to fix?

matildepark:
### ＵＳＥ

eglaysher (approved):
1) Edit the title for future searchability

2) This LG for now. If we're focusing on usability on stderr, in the long run we probably want to do a full pass on the messages themselves to ensure they all have a consistent format, instead of some having UPPER CASE NAMES, some having no subsystem names, and some displaying as a renders of tuples of strings.

3) Addressing the logging issue might be hard. RIO locks down their `LogOptions` structure so you cannot build your own `logSend`, which would let us redirect the logging to a function that locks or does STM or something.

pilfer-pandex:
Thanks.

1. Use.
2. Agree. I had thoughts of a system that would "push" context names into the LogEnv as you enter modules, but that might be too much fun.
3. I know :/ Spent several hours reading that code.

PR #3471, opened 2020-09-10
vere: fix crash, cleanup error messages

joemfb:
- v0.10.8 introduced a bug, wherein the vere king would crash if the serf crashed. This was just a dumb mistake in the pier shutdown state machine (advancing the state before dispatching off of it).
- the error message printed on unexpected serf shutdown (manifesting as EOF on the ipc pipe) used to use the default libuv error strings, but this was unhelpful and confusing, so it's been replaced with more specific messages on each side of the pipe.


brendanhay (approved):
LGTM :shipit:

PR #3472, opened 2020-09-10
arvo: adds new maintenance commands: |trim and |meld

joemfb:
`%meld` is the name of the new global deduplicator, added in #3235). Support for the effect is also added in that PR, but there's no need to block this addition, or otherwise coordinate the release of these changes -- if the effect is unsupported, the user will merely see a `kick: lost` printf.

The `%trim` effect (triggering memory reclamation via a subsequent `%trim` event) has been long supported, but there was no way to manually trigger it. I've taken the liberty of adding that as well. The triggering task to %dill is named `%crop`, as `%trim` is already reserved for the global memory pressure event.


philipcmonk (approved):
Looks good

PR #3473, opened 2020-09-10
publish: graph-store rewrite

liam-fitzgerald:
Rewrites publish entirely in graph-store
- Metadata store is now the canonical source of a notebook's title etc.
- Publish remains, but should be tombstoned after this migration successfully OTAs

#### Regressions
These mostly don't fit into the conceptual model of how graph-store works, or are planned for future slates of work
- Unread counters (will be revived with notificiation-store)
- Updating comments/posts (no support from graph-store)
- Disabling of comments is gone (no place to hold such app-specific metadata)
- Lazy loading of notebooks, it currently loads all relevant notes/comments upon entering a notebook. This will need to be resolved before merge.

#### Untested
These are both blocked on a functioning set of hooks for graph-store
- Migrations have not been tested
- Join/leave functionality has not been tested


tacryt-socryp:
I would love to discuss this in depth today if you have time, M.

matildepark:
We probably all should. I will book it for after midweek.

tacryt-socryp:
This needs to merge in release/next-userspace.

pkg/arvo/mar/graph/validator/publish.hoon
```diff
@@ -0,0 +1,43 @@
...
+      ip
+    ::  specific revision
+    ::  first content is the title
+    ::  revisions are numbered by the revision count
+    ::  starting at one
+        [@ %1 @ ~]
+      ?>  ?=(^ contents.p.ip)
+      ?>  ?=(%text -.i.contents.p.ip)
+      ip
+
```
tacryt-socryp, line 28: blank line

pkg/arvo/mar/graph/validator/publish.hoon
```diff
@@ -0,0 +1,43 @@
...
+    ::  container for comments
+        [@ %2 ~]
+      ?>  ?=(~ contents.p.ip)
+      ip
+    ::  comment
+      [@ %2 @ ~]  ip
```
tacryt-socryp, line 34: Do we want to limit the kinds of content that can go into comments?

pkg/interface/src/logic/lib/publish.ts
```diff
@@ -0,0 +1,117 @@
+import { Post, GraphNode, TextContent, Graph, NodeMap } from "~/types";
+
+const buntPost = (): Post => ({
+  author: '',
+  contents: [],
+  hash: null,
+  index: '',
+  signatures: [],
+  'time-sent': 0
+});
```
tacryt-socryp, line 10: Perhaps these "generic post functions" could be moved into a generic `lib/post.ts` library.

pkg/arvo/app/publish.hoon
```diff
@@ -242,8 +244,100 @@
...
+    ++  note-to-node
+      |=  =note
+      ^-  node:graph-store
+      =/  =index:graph-store
+        ~[date-created.note]
+      =/  contents=(list content:graph-store)
+        :~  [%text title.note]
+            [%text file.note]
+        ==
+      :_  (comments-to-children index note)
```
tacryt-socryp, line 317: We need the `%1` child for revisions and the `%2` pattern for comments in the publish mark to be reflected here. It is not currently. The top-level node should have null contents and the `%1` and `%2` child that is in the mark.

pkg/arvo/app/graph-store.hoon
```diff
@@ -513,7 +513,7 @@
     =/  =ship  (slav %p i.t.t.path)
     =/  =term  i.t.t.t.path
     =/  =index:store
-      (turn t.t.t.t.path |=(=cord (slav %ud cord)))
+      (turn t.t.t.t.path |=(=cord (rash cord dem:ag)))
```
tacryt-socryp, line 516: Why?

tacryt-socryp (changes requested):
The Publish migration and Publish graph validation mark do not match.

pkg/arvo/app/graph-store.hoon
liam-fitzgerald: To avoid the dots in the `@ud` syntax i.e. `1.000.000`

Fang-: Wait, but this is in the scry path? I don't think we should start introducing non-native syntax/encodings into those. If the concern here is generating scry URLs on the client side, implementing `scot` in "airlock" libraries is perfectly reasonable.

liam-fitzgerald:
can I get another pass @tacryt-socryp ? I've brought it inline with the new groups layout and fixed your previous comments.

tacryt-socryp:
I'll give a very thorough pass tomorrow during the day. Thanks for updating this!

pkg/interface/src/logic/lib/publish.ts
```diff
@@ -0,0 +1,101 @@
+import { Post, GraphNode, TextContent, Graph, NodeMap } from "~/types";
+import { buntPost } from '~/logic/lib/post';
+
+export function newPost(
+  title: string,
```
tacryt-socryp, line 5: Very good. I like this whole file.

pkg/interface/src/views/apps/publish/components/NoteNavigation.tsx
```diff
@@ -30,45 +31,60 @@ function NavigationItem(props: {
   );
 }
 
+function getAdjacentId(graph: Graph, child: number, backwards = false): number | null {
```
tacryt-socryp, line 34: I wonder if we should make a scry to receive a list of all the IDs in the graph. Would make this easier to paginate.

pkg/interface/src/views/landscape/components/GroupsPane.tsx
```diff
@@ -134,9 +134,9 @@ export function GroupsPane(props: GroupsPaneProps) {
         render={(routeProps) => {
           const { app, host, name } = routeProps.match.params;
           const appName = app as AppName;
-          const isShip = app === "link";
-          const appPath = `${isShip ? '/ship/' : '/'}${host}/${name}`;
-          const association = isShip ? associations.graph[appPath] : associations[appName][appPath];
+          const isGraph = app === "link" || app === 'publish';
```
tacryt-socryp, line 137: This logic should be a function in lib/util.

tacryt-socryp (commented):
First pass.

pkg/arvo/mar/graph/validator/publish.hoon
```diff
@@ -0,0 +1,46 @@
...
+      ?>  ?=(~ contents.p.ip)
+      ip
+    ::  specific revision
+    ::  first content is the title
+    ::  revisions are numbered by the revision count
+    ::  starting at one
+        [@ %1 @ ~]
+      ?>  ?=([* * ~] contents.p.ip)
+      ?>  ?=(%text -.i.contents.p.ip)
+      ?>  ?=(%text -.t.i.contents.p.ip)
```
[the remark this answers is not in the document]
liam-fitzgerald: Well we don't offer any affordances for anything other than posting and displaying text.

pkg/interface/src/views/apps/publish/components/NoteNavigation.tsx
liam-fitzgerald: Good idea

liam-fitzgerald:
cc: @ixv @Fang- would appreciate a sanity check on the backoff logic in the latest commit

pkg/arvo/app/publish.hoon
```diff
@@ -484,8 +506,41 @@
...
+      ::  we are safe to add it to the pull-hook
+      :_  this(migrate (~(del by migrate) rid))
+      ~[(poke-graph-pull:main %add entity.rid rid)]
+    ::  if nacked, then set a exponential backoff and retry
+    =/  nack-count=@ud
+      +((~(gut by migrate) rid 0))
+    :_  this(migrate (~(put by migrate) rid nack-count))
+    ::  (bex 19) is roughly 6 days
+    =/  wakeup=@da
+      (add now.bol (bex (min 19 nack-count)))
```
Fang-, line 529: Friendly reminder:

```hoon
> `@ux`~s1
0x1.0000.0000.0000.0000
```

```suggestion
    ::  (bex 19) is roughly 6 days
    =/  wakeup=@da
      (add now.bol (mul ~s1 (bex (min 19 nack-count))))
```

You might want to consider lower-bounding nack-count as well, retrying in 1, 4, 8.. etc seconds is unlikely to give different results.

Fang-: Also do we not want to stop trying after, say, 30 days or something?

joemfb: Here's the exponential backoff function in `:acme`:

https://github.com/urbit/urbit/blob/b866ff0e3d0f66ea753553f98530bc801090cd20/pkg/arvo/app/acme.hoon#L423-L431

I think we should put this in zuse somewhere.

tacryt-socryp:
notes:
- Notebook creation works
- New post works
- Edit post route doesn't work
- Clicking an already existing notebook doesn't work:
Screen Shot 2020-10-20 at 2 23 51 PM
- Notebook leave and deletion now works

matildepark:
Seems like either `contacts` or `node` isn't being passed to the `Notebook` route in question?

tacryt-socryp:
We need edit functionality to be finished and the migration to be thoroughly tested, but otherwise this all looks great.

matildepark:
Many such conflicts @liam-fitzgerald

pkg/arvo/app/publish.hoon
```diff
@@ -373,842 +502,57 @@
...
-    ::
-        [%view-bind ~]
-      [~ this]
-    ==
+    ?.  ?=([%graph-migrate *] wire)
+      (on-arvo:def wire sign-arvo)
+    =/  rid=resource
+      (de-path:resource t.wire)
+    ?>  ?=([%b %wake *] sign-arvo)
+    ~?  ?=(^ error.sign-arvo)
```
tacryt-socryp, line 543: never seen this rune before???

matildepark: The elusive conditional sigpam

image

ohAitch: `=?` and `~?` come up surprisingly often

tacryt-socryp:
Can confirm no crashes in migration. Unsure if logic around networked publish stuff works yet. Will test that next.

tacryt-socryp (approved):
Tested extensively with Liam and Matilde. LGTM.

PR #3475, opened 2020-09-10
jets: remove ut_fish, ut_fuse, ut_crop

jtobin:
This jet was previously decapitated in 61b957cc5b1, meaning that its formal logic was stubbed out and that it simply passes through a partially-memoizing Vere-level cache during evaluation.

My understanding is that only +ut arms involving so-called worklist algorithms -- usually indicated by the presence of 'gil', but here 'vot' -- need to be cached in this way.  The results of these algorithms are cached with keys that omit any reference to the worklist state itself, typically instead being based on the Hoon version, arm name, and sample value (hence *partial* memoization).

In the case of +fish, this partial memoization doesn't appear to accomplish much; tracking the keys generated for memoization of +fish during the lite boot sequence via a solid pill, for example, yields 5499 entries, 5498 of which are unique.  In other words, almost everything is a cache miss.

I've measured the performance of (lite) boot and +solid on a fakezod with and without the ut_fish jet, and there doesn't seem to be any appreciable difference either in time expenditure or memory usage.  A non-fish-jetted fakezod boots from either a solid or brass pill without any issue.  I've also been running ~nidsut-tomdun on mainnet using a non-fish-jetted binary for several days with normal use without having noticed anything out of the ordinary.

**EDIT**:

I've since removed ut_fuse and ut_crop as well, for which the story is similar.


jtobin:
(Just as an update, I traced +fish cache key creation earlier during a combined lite boot and +solid sequence; the result was 29665 unique entries out of a total of 29814.)

tacryt-socryp:
Is anything blocking this from being merged? @joemfb

jtobin:
With the benefit of having done further work in this area: this could be merged, or not (though I think the tree hashes would need to be updated now).  It doesn't really matter much either way; the remaining compiler jets *cannot* be removed until the Hoon-level cache is finished, so removing some benign compiler jets now and some non-benign compiler jets later really has the same effect as just removing them all later.

I'd probably just wait until they can all be removed and close this PR.  I'll do that, though feel free to reopen if for some reason someone would like to prune these jets now.

PR #3476, opened 2020-09-10
chat: Scroll behavior

tylershuster:
Some issues have arisen from the new chat implementation.

This PR mostly accounts for rendering weirdness. Some significant changes:
- The unread marker is now absolutely positioned so that its removal does not affect the scroll height and cause jank
- the envelope type is now correctly typed
- chat messages have been highly purified for fewer rerenders
- better idle handling on the chat window
- chat-window uses flex-column-reverse again, but doesn't need two versions for browsers. the same feature handles marking of unread messages more sane
- oembed fetches have been memoized for fewer network requests
- chat window now idles on window blur, meaning you get unread notifications when tab is in background

There is still some significant jank as you scroll up. Basically, I can blame all our problems on chat-store not being validated. Let's say there's a room with 100 messages.

A call to `/chat-view/paginate/80/90` *should* return messages with numbers 10-20, but there's no guarantee that it does. On the host it *probably* will, but we've seen problems everywhere from messages that should be `123456789` being numbered `678912345` to `987654321` That makes seeking accurately impossible, so you will see some strange behavior the higher you go in the scrollback. This will not be a problem with `graph-store`, or so I'm told, but I assume chat will be the last to migrate. Food for thought.


tylershuster:
Oh, I meant to make this a draft. Might work better as is, but should be tested. I'm just going to run this on my machine until I'm satisfied

galenwp:
Do we think this also handles the case where we start our scroll somewhere in the backlog when loading a channel?

tylershuster:
Yeah, it's supposed to. But again, having inconsistent return values from chat-view/paginate really throws a wrench in anything relating to the backlog. It might be worth having a superhooner take a look at what it would take to make that more consistent.

matildepark:
Chat migration is basically two weeks off from now.

tylershuster:
That's not that far...I think you should try this — I think it's better than the current livenet and minor changes after graph-store will make it perfect.

matildepark:
The scrolling up behaviour now seems to be "it's a lil slow and choppy" but it doesn't push us around. So yeah, an improvement. I had a case where the unread messages banner immediately dismissed, but then in other rooms worked normally.

matildepark:
Choose your character 
Sep-10-2020 17-36-00

matildepark:
Still has some inconsistent "page loads mid-backlog" cases, but they're not like, concretely reproducible for me.

I want to keep this open until closer to next deployment for continued tweaking and @liam-fitzgerald and/or @vvisigoth can run this PR and add cases.

liam-fitzgerald:
[this looks odd](http://ewr1.vultrobjects.com/urbit/rendering-issue.mov)

tylershuster:
Ok, this is a big change. I bit the bullet and used the `react-virtualized` library — it needs some cleanup to get rid of the old implementation but I think it's good enough for testing. This library rerenders *much* less aggressively than the previous, which accounted for the behavior both of you have documented here. It does this by `position: absolute`ing everything which means that it has to keep track of cell heights, and those caches need to be cleared when message content changes for whatever reason. But it's looking good. At least enough that I would like to get any feedback from the dedicated squad of alpha testers (har har).

matildepark:
I will try this! But late tonight Eastern.

matildepark:
You can't refresh or load directly on a chatroom page. 
image

tylershuster:
I DMed you about this. This is weird, and looks like it broke a limit that the app was already pushing. Looking into how to reduce nested updates.

matildepark:
If you have many unreads then all the pages load in overlapping like that 2018 screenshot

image

tylershuster:
Ok, the overlapping one is fixed — height cache is cleared for messages that are fetched.

matildepark:
The (livenet-style) scroll anchoring jank returns in rooms with many unreads (this time ~90+) for some reason. It also doesn't jump up to the unread position, you have to click it.

matildepark:
I'm so confused as to why it's so inconsistent with behaviour though, let me keep trying it tonight.

tylershuster:
Latest commit refactors into a totally new implementation. It still has some work to do, but I think the overall experience is leagues ahead of where it's been. Tested in Firefox, Chrome, and Safari, and they all work.
The secret is `container { transform: scaleY(-1); } child { transform: scaleY(-1); }` — this is the Correct way to tell a browser to calculate its scroll distance from the bottom. It also intercepts the `wheel` event to compensate for the inversion.

Current issues mostly regard a non-optimized fetching of messages, resulting in a blinking loading bar.

As mentioned to @matildepark, the sidebar DMs seem to be rearranging when loaded. But unless I'm mistaken I haven't changed the sidebar.

Remaining work to do includes:
- Aforementioned backlog fetching performance
- reimplementing sigil-overlay scroll behavior
- implementing the other two virtualized scroll windows using this method and removing the old external libraries

pkg/interface/src/views/apps/chat/components/lib/chat-window.tsx
```diff
@@ -267,7 +213,7 @@ export class ChatWindow extends Component<ChatWindowProps, ChatWindowState> {
     const messages: Map<number, Envelope | IMessage> = new Map();
 
     envelopes
-      .sort((a, b) => b.when - a.when)
+      .sort((a, b) => a.when - b.when)
```
matildepark, line 216: This is the reason for the DMs being in a weird order. We saw this last time someone tried a flexbox implementation that didn't use column-reverse I think? Or something. We fixed it using the tweak in #2751. Just do that in the root component.

matildepark:
The chat window shrinks as you go up? 
Sep-14-2020 19-26-02

matildepark:
Backlog load loop.

Sep-14-2020 19-30-11

matildepark:
I caused a conflict by accident here...

In any case the scrolling itself feels natural; haven't captured tests for mobile or for avoiding slamming cases yet; just has the quirks aforementioned.

tylershuster:
Ok, these two issues are due to some sloppy math on my part regarding buffering and overscanning, so I'm patching those now

tylershuster:
Ok, there is one caveat here relating to the misnumbering issue, and that's when you scroll to the top in a room where the index is off, you might get a gap at the top and it will keep trying to load the missing messages.

Other than that the issues should be fixed.

matildepark:
Testing as of 12df5a2 — no scroll slamming when a new message comes in. Scrolling down doesn't even reveal it — I have to scroll up, *then* scroll down. Maybe it thinks we're still pinned to bottom.

tylershuster:
Yup, needed to recompute. Should work now

matildepark:
Oh ho ... this may be the one @tylershuster. It does everything and feels so smooth.

@liam-fitzgerald @vvisigoth Care to test drive?

@tylershuster Resolve conflict and clean up commits? Let's get this ready ...

tylershuster:
Ok, I think it's kosher

liam-fitzgerald:
- embeds autoclose when you scroll
- managed to reproduce a weird scroll 'vibration' when you scroll past an animated gif and it's loaded, but not visible [demo](http://ewr1.vultrobjects.com/urbit/scroll-jank.mov)

looks really nice otherwise!

matildepark:
> embeds autoclose when you scroll

This one seems resolved now.

There's a mobile case for backlog loading I linked out of band. I'm gonna try the animated GIF case.

DMs moving out of order when they are loaded is either still there or returned, I don't recall noticing it last time.

tylershuster:
Some explanation: this implementation (and all virtualizers) have a buffer zone on either side of the window. When you saw it, Liam, the buffer zones were independent areas, meaning that when an object scrolled into them, they rerendered, hence the autoclosing. I believe this was also the case for the gif issue. What is with you people and short viewports (I kid).

This issue should be fixed by reconciling both the calcuated buffers and the visible items into a single area, where the embeds won't autoclose until they have reached at least a screen's height outside of the current window. I also fixed a typo which was resulting in the gap being calculated based on the average height, producing jitters.

However this does bring up an interesting point — Sometimes I will start playing media and continue listening as I scroll down the page. The very concept of virtualization is incompatible with this user story — cutting renders means cutting renders.

matildepark:
> However this does bring up an interesting point — Sometimes I will start playing media and continue listening as I scroll down the page. The very concept of virtualization is incompatible with this user story — cutting renders means cutting renders.

I can accept that. I can't think of any other messaging product that let you just scroll indefinitely while playing an embed.

tylershuster:
I forgot to mention my possible solution was to use portals to create an offscreen region where open embeds go.

matildepark:
No need to get fancy just yet — we can do that in another release. It would be cool to plan for tonight.

tylershuster:
Oh, this should fix #3482

matildepark:
I will note that while the merge conflict is resolved the changes in #3347 are not reflected here.

tylershuster:
Wtf I put them in and tested last night

matildepark:
> Wtf I put them in and tested last night

I think that those changes and the DM changes were perhaps dropped during rebase.

matildepark:
I do have a tiny tweak slightly out of scope it would be good to add here — reactmarkdown in content/text.js can take `parserOptions` to pass [here](https://github.com/remarkjs/remark/tree/main/packages/remark-parse) if you're feeling up to stopping `>greentext without a gap` from processing as a blockquote, etc.

tylershuster:
@urcades @matildepark WDYT about this?

[video](https://marpem-files.sfo2.digitaloceanspaces.com/Screen%20Recording%202020-09-15%20at%202.04.10%20PM.mov)

The purpose is to prevent jumping around when you first load a channel, while it's figuring out which messages should be where. It waits until it's all loaded and calculated, or 3 seconds, whichever comes first.

matildepark:
Replace the white with the placeholder that’s in current chat?

—
~haddef-sigwen
https://urbit.org

tylershuster:
Ok my dudes

tylershuster:
Oh, and be sure to check the group pages that were previously virtualized

matildepark:
reword commits! will test!

matildepark (approved):

PR #3478, opened 2020-09-10
links: graph-store rewrite

tacryt-socryp:
Completed:
- Link submission
- Link commenting
- Rendering data coming from graph-store
- Sidebar rendering (required create thread)
- Settings metadata editing
- Finished remove / delete support
- Hooks
- Joining (requires join thread to be done)

Edit: Completely done


liam-fitzgerald (changes requested):
couple of nits, looks good otherwise

pkg/interface/src/views/apps/links/components/new.tsx
```diff
@@ -0,0 +1,107 @@
...
+          name,
+          description,
+          group
+        );
+      } else {
+        await props.api.graph.createUnmanagedGraph(
+          resourceId,
+          name,
+          description,
+          { open: {
```
liam-fitzgerald, line 46: This differs from the current behaviour of unmanaged links, which are always invite-only

liam-fitzgerald (changes requested):
Fetch-on-every render still an issue from the old review and one small nit, but otherwise looks good!

pkg/interface/src/views/apps/links/components/new.tsx
tacryt-socryp: Behavior is now the same, thanks!

tacryt-socryp:
@liam-fitzgerald I've fixed the fetch every render issue with `useEffect` as you suggested. Thanks!

belisarius222 (approved):
Hoon LGTM.  I don't really have enough context on the JS to review thoroughly.

tacryt-socryp:
Currently, joining from the Groups page doesn't work, as the metadata tag "graph" doesn't correspond to the use case... Going to work on adding the proper tag to graphs in the `%create` thread, and making the redirect route in `/~graph` check for that tag before redirecting to the proper app.

matildepark:
Aren't all the routes changing in the other PR anyway? I don't know what the new join routes are in that one.

tacryt-socryp:
This PR is good to go as is. I need to do some `%metadata-store` work to make the “Channels” page work normally with graph-store resources, and that does block release of the new Links version, but is not important within context of this PR.

PR #3479, opened 2020-09-11
graph-view backend

liam-fitzgerald:
- Adds a HTTP interface to spider accessible by POSTing JSON to `/spider/[input-mark]/[thread-name]/[output-mark].json`
- Adds create,join,leave,delete,groupify threads to act as a view for graph-store
- Expands the JS graph API to accomodate these threads
- Now with bonus error-handling, returning a meaningful error code on bad request/offline ship/bad permissions


pkg/arvo/lib/graph-view.hoon
```diff
@@ -0,0 +1,80 @@
+/-  sur=graph-view
+/+  resource, group-store
+^?
+=<  [sur .]
+=,  sur
+|%
+++  app-to-mark
```
tacryt-socryp, line 7: We shouldn't need this. The mark should just be specified directly as a %tas.

pkg/arvo/lib/graph-view.hoon
```diff
@@ -0,0 +1,80 @@
...
+    ++  create
+      %-  ot
+      :~  resource+dejs:resource
+          title+so
+          description+so
+          app+app
+          associated+associated
+      ==
+    ::
+    ++  app
```
tacryt-socryp, line 37: There shouldn't be an `app-name` type. We should just return a `tas`

pkg/arvo/lib/graph-view.hoon
```diff
@@ -0,0 +1,80 @@
...
+      %-  ot
+      :~  resource+dejs:resource
+          app+app
+      ==
+    ::
+    ++  join
+      %-  ot
+      :~  resource+dejs:resource
+          app+app
+          ship+(su ;~(pfix sig fed:ag))
```
tacryt-socryp, line 62: Why do we need ship? Could we not use the resource's ship?

pkg/arvo/sur/graph-view.hoon
```diff
@@ -0,0 +1,47 @@
...
+|%
+::  $associated: A group to associate, or a policy if it is unmanaged
+::
++$  associated
+  $%  [%group rid=resource]
+      [%policy =policy]
+  ==
+::  $app: An app that is associated to a graph-store mark
+::
++$  app-name  ?(%chat %publish %links)
```
tacryt-socryp, line 13: Should be removed for more genericity.

pkg/arvo/ted/graph/create.hoon
```diff
@@ -0,0 +1,60 @@
...
+^-  form:m
+=+  !<(=action:graph-view arg)
+?>  ?=(%create -.action)
+;<  =bowl:spider  bind:m  get-bowl:strandio
+::  Add graph to graph-store
+::
+?.  =(our.bowl entity.rid.action)
+  (strand-fail:strandio %bad-request ~)
+=/  =update:graph
+  [%0 now.bowl %add-graph rid.action *graph:graph `(app-to-mark:graph-view app.action)]
```
tacryt-socryp, line 35: Should just be a term.

pkg/arvo/ted/graph/create.hoon
```diff
@@ -0,0 +1,60 @@
...
+::
+=/  =metadata
+  %*  .  *metadata
+    title  title.action
+    description  description.action
+    date-created  now.bowl
+    creator  our.bowl
+  ==
+=/  act=metadata-action
+  [%add group-path [app.action (en-path:resource rid.action)] metadata]
```
tacryt-socryp, line 56: `app.action` should stick around just to be used for this field in metadata-store, and we shoud add `mark.action` for the other use case.

tacryt-socryp (approved):
Small change is necessary, but other than that this looks great.

pkg/arvo/lib/graph-view.hoon
liam-fitzgerald: Was more in the interest of maintaining a bijection between the JSON and the actual interface of the app

pkg/arvo/app/spider.hoon
```diff
@@ -394,6 +463,25 @@
...
+  =-  (fall - `state)
+  %+  bind  
+    (~(get by serving.state) tid)
+  |=  [eyre-id=@ta output=mark]
+  :_  state(serving (~(del by serving.state) tid))
+  %+  give-simple-payload:app:server  eyre-id
+  ^-  simple-payload:http
+  :_  ~  :_  ~
+  ?.  ?=(http-error:spider term)
+    ((slog tang) 500)
```
belisarius222, line 478: It would be nice to add a prefix tank to this saying it was an HTTP thread failure.  Not blocking.

pkg/arvo/app/spider.hoon
```diff
@@ -402,7 +490,24 @@
...
   =/  fail-cards  (thread-say-fail tid term tang)
   =^  cards  state  (thread-clean yarn)
-  [(weld fail-cards cards) state]
+  =^  http-cards  state  (thread-http-fail tid term tang)
+  [:(weld fail-cards cards http-cards) state]
+::
+++  thread-http-response
+  |=  [=tid =vase]
+  ^-  (quip card ^state)
+  =-  (fall - `state)
```
belisarius222, line 499: How would we get an HTTP response from a nonexistent thread?  This seems like it should be an assertion, not a silent no-op.

pkg/arvo/ted/graph/join.hoon
```diff
@@ -0,0 +1,74 @@
+/-  spider, graph-view, graph=graph-store, *metadata-store, *group
+/+  strandio, resource
+=>
+|% 
+++  strand  strand:spider
+++  fail    strand-fail:strand
+++  poke  poke:strandio
+++  poke-our   poke-our:strandio
+::
+++  check-live
```
belisarius222, line 10: Why does this have a timeout loop?  I think this should be handled by Ames, not here.

belisarius222 (changes requested):

belisarius222 (approved):

PR #3480, opened 2020-09-11
vere: refactors terminal output implementation for efficiency

joemfb:
And, hopefully, for clarity...

This code could still use lots of improvement; in particular, the output state machine seems unnecessarily complex. But I've tried to remove layers of unnecessary indirection, reallocation, measurements of statically-sized outputs, &c. I think the biggest win is the direct utf-32 to utf-8 conversation, and the most questionable change is attempting synchronous writes.

These changes are not particularly high-priority -- I'm mostly just scratching at a longstanding itch. I'd be happy to make any changes you'd like.

### DIFF
--- a/pkg/urbit/vere/io/term.c
+++ b/pkg/urbit/vere/io/term.c
@@ -485,11 +485,39 @@ _term_it_set_line(u3_utty* uty_u,
                   c3_w     sap_w)
 {
   u3_utat* tat_u = &uty_u->tat_u;
-  u3_noun    txt = u3do("tuft", u3i_words(wor_w, lin_w));
-  c3_w     byt_w = u3r_met(3, txt);
   c3_y*    hun_y = (c3_y*)lin_w;
+  c3_w     byt_w = 0;
 
-  u3r_bytes(0, byt_w, hun_y, txt);
+  //  convert lin_w in-place from utf-32 to utf-8
+  //
+  //    (this is just a hand-translation of +tuft)
+  //
+  {
+    c3_w car_w, i_w;
+
+    for ( i_w = 0; i_w < wor_w; i_w++ ) {
+      car_w = lin_w[i_w];
+
+      if ( 0x7f >= car_w ) {
+        hun_y[byt_w++] = car_w;
+      }
+      else if ( 0x7ff >= car_w ) {
+        hun_y[byt_w++] = 0xc0 ^ ((car_w >>  6) & 0x1f);
+        hun_y[byt_w++] = 0x80 ^ (car_w & 0x3f);
+      }
+      else if ( 0xffff >= car_w ) {
+        hun_y[byt_w++] = 0xe0 ^ ((car_w >> 12) &  0xf);
+        hun_y[byt_w++] = 0x80 ^ ((car_w >>  6) & 0x3f);
+        hun_y[byt_w++] = 0x80 ^ (car_w & 0x3f);
+      }
+      else {
+        hun_y[byt_w++] = 0xf0 ^ ((car_w >> 18) &  0x7);
+        hun_y[byt_w++] = 0x80 ^ ((car_w >> 12) & 0x3f);
+        hun_y[byt_w++] = 0x80 ^ ((car_w >>  6) & 0x3f);
+        hun_y[byt_w++] = 0x80 ^ (car_w & 0x3f);
+      }
+    }
+  }
 
   c3_free(tat_u->mir.lin_y);
   tat_u->mir.lin_y = hun_y;
@@ -498,8 +526,6 @@ _term_it_set_line(u3_utty* uty_u,
   tat_u->mir.sap_w = sap_w;
 
   _term_it_show_line(uty_u, wor_w, sap_w);
-
-  u3z(txt);
 }
 
 /* _term_it_show_more(): new current line.
--- a/pkg/urbit/vere/io/term.c
+++ b/pkg/urbit/vere/io/term.c
@@ -169,13 +169,13 @@ u3_term_log_init(void)
     //  configure input escape sequences
     //
     //    NB: terminfo reports the wrong sequence for arrow keys on xterms.
-    //
-    {
-      uty_u->ufo_u.inn.kcuu1_u = TERM_LIT_BUF("\033[A");  //  terminfo reports "\033OA"
-      uty_u->ufo_u.inn.kcud1_u = TERM_LIT_BUF("\033[B");  //  terminfo reports "\033OB"
-      uty_u->ufo_u.inn.kcuf1_u = TERM_LIT_BUF("\033[C");  //  terminfo reports "\033OC"
-      uty_u->ufo_u.inn.kcub1_u = TERM_LIT_BUF("\033[D");  //  terminfo reports "\033OD"
-    }
+    //    disabled, currently unused
+    // {
+    //   uty_u->ufo_u.inn.kcuu1_u = TERM_LIT_BUF("\033[A");  //  terminfo reports "\033OA"
+    //   uty_u->ufo_u.inn.kcud1_u = TERM_LIT_BUF("\033[B");  //  terminfo reports "\033OB"
+    //   uty_u->ufo_u.inn.kcuf1_u = TERM_LIT_BUF("\033[C");  //  terminfo reports "\033OC"
+    //   uty_u->ufo_u.inn.kcub1_u = TERM_LIT_BUF("\033[D");  //  terminfo reports "\033OD"
+    // }
 
     //  Load old terminal state to restore.
     //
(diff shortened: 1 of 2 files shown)


pkg/urbit/vere/io/term.c
```diff
@@ -143,60 +148,33 @@ u3_term_log_init(void)
       uv_pipe_open(&(uty_u->pop_u), uty_u->fid_i);
     }
 
-    //  Load terminfo strings.
+    //  configure output escape sequences
+    //
+    //    (as reported by the terminfo database we bundled)
```
Fang-, line 153: I don't understand this comment. The sequences below are literals in our code, not loaded in from terminfo, right? Where does the terminfo database file come into play?

pkg/urbit/vere/io/term.c
```diff
@@ -13,6 +13,11 @@
 #include "all.h"
 #include "vere/vere.h"
 
+//  macros for string literal args/buffers
+//
+#define TERM_LIT(s)      sizeof(s) - 1, (const c3_y*)(s)
+#define TERM_LIT_BUF(s)  uv_buf_init(s, sizeof(s) - 1)
```
Fang-, line 19: The `- 1` here are subtracting the trailing null byte, I'm guessing?

pkg/urbit/vere/io/term.c
```diff
@@ -358,87 +322,91 @@ _term_write_cb(uv_write_t* wri_u, c3_i sas_i)
...
-    c3_y* buf_y = c3_malloc(old_u->len_w);
-
-    memcpy(buf_y, old_u->hun_y, old_u->len_w);
-    buf_u = uv_buf_init((c3_c*)buf_y, old_u->len_w);
-
-    c3_free(old_u);
-  }
-  _term_it_write_buf(uty_u, buf_u);
+  uv_buf_t buf_u = uv_buf_init((c3_c*)hun_y, len_w);
+  _term_it_write(uty_u, &buf_u, 0);
```
Fang-, line 391: You could consider using `_term_it_dump_buf` here for consistency. You use it (instead of manually passing a `0` pointer) everywhere else.

pkg/urbit/vere/io/term.c
```diff
@@ -526,10 +494,12 @@ static void
 _term_it_show_more(u3_utty* uty_u)
 {
   if ( c3y == u3_Host.ops_u.tem ) {
-    _term_it_write_str(uty_u, "\n");
-  } else {
-    _term_it_write_str(uty_u, "\r\n");
+    _term_it_dump(uty_u, TERM_LIT("\n"));
+  }
+  else {
+    _term_it_dump(uty_u, TERM_LIT("\r\n"));
```
Fang-, line 573: Might be a good time to ask: what's the deal with the "terminal assumptions" flag, and why does it influence how we print newlines?

pkg/urbit/vere/io/term.c
```diff
@@ -464,37 +445,63 @@ _term_it_show_cursor(u3_utty* uty_u, c3_w cur_w)
...
-  }
-  uty_u->tat_u.mir.len_w = len_w;
-  uty_u->tat_u.mir.sap_w = sap_w;
+  _term_it_dump(uty_u, tat_u->mir.byt_w, tat_u->mir.lin_y);
+
+  //  XX refactor to avoid updating state
+  //
+  tat_u->mir.cus_w += wor_w;
+  tat_u->mir.wor_w = wor_w;
+  tat_u->mir.sap_w = sap_w;
```
Fang-, line 494: Not sure what's going on here. Don't you only need to update these alongside `lin_y` (as done in `_term_it_set_line`)? If `lin_y` doesn't change, why would these? Guessing there's some relation to the bytes<->words measurements I'm not fully grokking.

I guess we still update `cus_w` regardless because rendering puts our cursor at the end of whatever we rendered...

pkg/urbit/vere/io/term.c
```diff
@@ -485,11 +485,39 @@ _term_it_set_line(u3_utty* uty_u,
...
   u3_utat* tat_u = &uty_u->tat_u;
-  u3_noun    txt = u3do("tuft", u3i_words(wor_w, lin_w));
-  c3_w     byt_w = u3r_met(3, txt);
   c3_y*    hun_y = (c3_y*)lin_w;
+  c3_w     byt_w = 0;
 
-  u3r_bytes(0, byt_w, hun_y, txt);
+  //  convert lin_w in-place from utf-32 to utf-8
+  //
+  //    (this is just a hand-translation of +tuft)
```
Fang-, line 526: Nice! Does this feel faster?

Any reason not to throw this into a jet?

pkg/urbit/vere/io/term.c
```diff
@@ -333,17 +335,48 @@ _term_it_write(u3_utty*  uty_u,
...
+  //
+  uv_buf_t     fub_u = { .base = buf_u->base, .len = buf_u->len };
+  uv_stream_t* han_u = (uv_stream_t*)&(uty_u->pop_u);
+  c3_i         ret_i;
 
-  wri_u->data = ptr_v;
+  //  try to write synchronously
+  //
+  while ( 1 ) {
+    ret_i = uv_try_write(han_u, &fub_u, 1);
```
Fang-, line 348: Under what conditions would this fail?

Fang- (approved):
Thanks for small commits, this was nice to review.

No real objections here, these all seems like good QoL changes. I do have a bunch of questions for you, for both existing and new code, to make sure I understand these things properly.

> I'd be happy to make any changes you'd like.

This seems fine to go in as-is, but if you're still feeling the itch, please do go wild. (;  
Even if that's just aggressively writing `XX refactor this by doing x, y, z` all over the place, to make sure this file continues to be pulled in a single/coherent direction as I do more work on it in the future.

pkg/urbit/vere/io/term.c
joemfb: There are a variety of error conditions for a bad call (bad parameters, invalid handle state, &c), there are failures in the underlying `write(2)` call that can bubble up, there are partial writes, and then there is `UV_EAGAIN`, which means that the stream is still connecting, or has asynchronous writes already queued. If we always use `uv_try_write()` first, and the stream is always already "connected" (by virtue of being a `uv_pipe_t`), I'm not sure that we'll ever end up with asynchronous writes.

Arguably, the terminal should just be completely synchronous, but that's a significantly larger change.

pkg/urbit/vere/io/term.c
joemfb: It's hard to tell from just mashing the keyboard locally. I need to run this on a slower cloud VM, I expect it'll reduce or remove the prompt flickering that used to occur.

I lot of stuff in Arvo's text output codepath could stand to be jetted. I'd rather not call a jet here even if I had one, since I want to mutate the buffer in place rather than reallocate. But this could be factored into a function used here and in a jet. This conversion could also be performed earlier in the process, while building up the initial char buffer and interpolating escape codes, but I'm not sure that would make much of a difference.

Note also that I didn't rewrite the inverse translation in `+taft` -- that's conditionally called if the (utf-8) input requires it, so it seems like less of a priority. But it wouldn't be too hard.

pkg/urbit/vere/io/term.c
joemfb: We currently have to (re-)set these values, as some sequences of output operations change them (ie, `_show_clear()` followed by `_show_line()`).

pkg/urbit/vere/io/term.c
joemfb: `u3_Host.ops_u.tem` (the `-t` flag) disables "terminal assumptions": specifically, the assumption that we're connected to an interactive tty. Setting that option means that we just append output, and avoid refreshing the prompt, ringing the bell, &c. We may want to disable style-rendering entirely as well.

pkg/urbit/vere/io/term.c
joemfb: Originally, the urbit binary was built on/far specific environments and we used the local terminfo database. When we switched to static binaries, there were cross-compilation and compatibility issues, so we bundled a terminfo database. But our escape code usage is very minimal and broadly compatible, so I removed the terminfo integration entirely in v0.10.8. I'll update the comments.

joemfb:
Updated, PTAL.

pkg/urbit/vere/io/term.c
Fang-: Would I be correct in interpreting that as "logging mode" in a sense? Disabling style seems the right thing there, yes.

Fang- (approved):
Thanks for the answers! Corrections and comments look good. Merge at your leisure. (:

pkg/urbit/vere/io/term.c
joemfb: Yeah, that's exactly right. It's "non-interactive foreground mode", while "daemon mode" is "non-interactive background mode".

Issue #3483, opened 2020-09-14
webdojo: channel / subscription initialises inconsistently


matildepark:
It usually takes a few navigations back and forth to get the prompt to show on Dojo, indicating our subscription has started. There's just some sequencing to sort out ...

matildepark:
I no longer encounter this at all, so closing as of #3484.

PR #3484, opened 2020-09-14
soto: call resetControllers on mount

matildepark:
It might address #3483 but too early to tell. Try a reproduction on livenet and in this branch by navigating quickly from apps to Dojo. On livenet, you can navigate between apps fine as fast as you want; but Dojo doesn't like it if you go to it quickly in sequence after another app and doesn't initiate.

I sequence the `resetControllers()` in the onMount and explicitly state what its `this` scope is to prevent any potential issue where `resetControllers()` was called too quickly or just not processing at all (since `this` might not be bound to the component?). It seems to be more reliable this way.

### DIFF
--- a/pkg/interface/src/views/apps/dojo/app.js
+++ b/pkg/interface/src/views/apps/dojo/app.js
@@ -20,7 +20,7 @@ export default class DojoApp extends Component {
     this.store.setStateHandler(this.setState.bind(this));
 
     this.state = this.store.state;
-    this.resetControllers();
+    this.resetControllers = this.resetControllers.bind(this);
   }
 
   resetControllers() {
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/apps/dojo/app.js
```diff
@@ -20,7 +20,7 @@ export default class DojoApp extends Component {
     this.store.setStateHandler(this.setState.bind(this));
 
     this.state = this.store.state;
-    this.resetControllers();
+    this.resetControllers = this.resetControllers.bind(this);
```
tacryt-socryp, line 23: This bind is unnecessary (a no-op, given that the function is not used in any closures that would otherwise lose the `this` context)

tacryt-socryp (approved):

pkg/interface/src/views/apps/dojo/app.js
matildepark: so it is ...

PR #3486, opened 2020-09-15
vere: fixes -P sample profiling

joemfb:
... which I had broke somewhere in the lead up to `v0.10.8`, but adding "allocator" annotations to the `u3i_*()` atom allocation functions, but not accounting for their early returns.

This PR also unconditionally disables sample profiling in the king -- there's currently no way to capture the output, and it'll be easy to turn back on if we ever decide we want to and add such a mechanism.


belisarius222 (approved):
LGTM

PR #3494, opened 2020-09-16
webchat: fix slow scrolling

### LINKED ISSUE #3492: webchat: scrolling is "slow"
After latest update, scrolling through chat messages is slow. The same amount of scrollwheel movement now results in much slower scroll speed.

This affects the messages part of the window, but the sidebar with channels seems unaffected.

Watch me try to scroll as fast I can:
scroll-ree

This is on Firefox 80.0b5.

pkova:
Fixes #3492

Most browser give deltaY as pixels on the wheel-event, Firefox gives it in lines (deltamode is 1, not 0: https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent/deltaMode). There are also major inconsistencies with what resolution various hardware and browser combinations fire the event, for example this bug only occurs for me on Firefox AND when using a mouse instead of a touchpad. The picture below logs a single mousescroll with that setup (with Chrome the values are the same).

Screenshot 2020-09-16 at 19 24 24

Now I hate adding dependencies as much as the next guy, but this library is just https://gist.github.com/akella/11574989a9f3cc9e0ad47e401c12ccaf, and solves the problem quite nicely.

### DIFF
--- a/pkg/interface/package.json
+++ b/pkg/interface/package.json
@@ -40,7 +40,8 @@
     "suncalc": "^1.8.0",
     "urbit-ob": "^5.0.0",
     "urbit-sigil-js": "^1.3.2",
-    "yup": "^0.29.3"
+    "yup": "^0.29.3",
+    "normalize-wheel": "1.0.1"
   },
   "devDependencies": {
     "@babel/core": "^7.9.0",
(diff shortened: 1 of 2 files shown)


pkova:
Ping @Fang- @tylershuster

matildepark:
Wow. Beast. @Fang- care to test?

tylershuster:
I haven't experienced the issue, but this implementation looks solid

Fang-:
Can confirm this fixes #3492 for me! Thanks @pkova. (:

matildepark (approved):

PR #3499, opened 2020-09-16
chat: fixes dancing timestamps

### LINKED ISSUE #3493: webchat: fade-in on timestamps causes elements to be displaced slightly
Observed after on/after latest update. Watch closely:

jiggle-ree

The movement is like half a pixel. You can't make it out from the gif, but things get blurry while they're displaced. Notice also the artifact on the sigil near the bottom. In the second half of the gif, it looks like this only applies to items below whatever is hovered on, but this is not the case. Items further up on the page are still affected.

This is on Firefox 80.0b5.

tylershuster:
fixes #3493

@Fang- can you confirm? I only saw it in Safari but this fixes it there

### DIFF
--- a/pkg/interface/src/views/components/VirtualScroller.tsx
+++ b/pkg/interface/src/views/components/VirtualScroller.tsx
@@ -283,7 +283,7 @@ export default class VirtualScroller extends PureComponent<VirtualScrollerProps,
     
     const indexesToRender = Array.from(visibleItems.keys());
 
-    const transform = origin === 'top' ? 'scaleY(1)' : 'scaleY(-1)';
+    const transform = origin === 'top' ? 'scale3d(1, 1, 1)' : 'scale3d(1, -1, 1)';
 
     const render = (index) => {
       const measure = (element) => {


matildepark:
Indeed does address it in Safari.

Fang-:
It's a bit more subtle now, but still occurs for me on Firefox 80.0b. 

Noticed that the sigil artifact is visible when zoomed out to 90% (first gif) which I normally browse with, but not when at 100% zoom (ie, no zoom, second gif).

more-jiggle-ree
-
bigger-more-jiggle-ree

tylershuster:
I can replicate in Firefox Developer Edition, even with this fix, and now I see the pixel artifact. I want to chalk this up to a rendering engine update that hasn't shipped to stablefox yet

Fang-:
That's fair game, we can see if this makes its way into stable and solve as needed.

matildepark:
> I can replicate in Firefox Developer Edition, even with this fix, and now I see the pixel artifact.

Wait so does that mean you develop on Firefox bleeding edge?

tylershuster:
I develop on Firefox stable, which is why I hadn't seen this before. I hadn't paid attention to the "b" in fang's browser version but when I did I downloaded developer edition and saw it there

matildepark:
So why would we wait? If beta breaks this?

tylershuster:
I think we should deploy this fix merely because it fixes Safari.

My intuition tells me that Firefox beta breaks this because they are working on rendering engine changes that are not ready to deploy — as evidenced by this issue.

I assume that they will account for this behavior by the time it ships to stable.

Fang-:
Just upgraded to 81.0b3 and can no longer reproduce... even on livenet!

tylershuster:
I can definitely see it on 81.0b9

tylershuster:
FWIW:
FF: No -> No
FF Dev: Sometimes -> Sometimes
Chrome: No -> No
Chrome Canary: No -> No
Safari: Yes -> No
Safari Tech Preview: No -> No

matildepark (approved):

PR #3507, opened 2020-09-17
all: Fixes joining navigation jank

### LINKED ISSUE #3503: all: Join URLs can't be paged back through
When using `/join` URLs to join a channel, you can't use the browser's back button to return to the previous page, because it tries to autojoin again.

I'm working on fixing this, I just wanted something to reference.

### LINKED ISSUE #3504: chat: immediately loads blank screen instead of requested channel
I know why this is happening, just need something to reference. Basically the `/join` screen immediately tries to load instead of waiting for the `api.chat.join` promise to fulfill

tylershuster:
Uses the `replaceState` feature of the history API to change-in-place the page
location instead of pushing a new entry, which prevents backward navigation.
Also irons out some inconsistent behavior on joining.

Fixes #3504, fixes #3503, fixes some unfiled bugs

### DIFF
--- a/pkg/interface/src/logic/api/chat.ts
+++ b/pkg/interface/src/logic/api/chat.ts
@@ -81,7 +81,7 @@ export default class ChatApi extends BaseApi<StoreState> {
    * If we don't host the chat, then it just leaves
    */
   delete(path: Path) {
-    this.viewAction({ delete: { 'app-path': path } });
+    return this.viewAction({ delete: { 'app-path': path } });
   }
 
   /**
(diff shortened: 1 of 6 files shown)


pkg/interface/src/logic/api/chat.ts
```diff
@@ -81,7 +81,7 @@ export default class ChatApi extends BaseApi<StoreState> {
    * If we don't host the chat, then it just leaves
    */
   delete(path: Path) {
-    this.viewAction({ delete: { 'app-path': path } });
+    return this.viewAction({ delete: { 'app-path': path } });
```
matildepark, line 84: I was wondering why leaves never processed...

matildepark (approved):

PR #3510, opened 2020-09-18
all: noop on unmounted fetch embed error

### LINKED ISSUE #3500: chat: embeds inconsistently appear as either embeds or links
Hard to reproduce — it's inconsistent. Is there something about our rendering logic that sidesteps the embed parser?

tylershuster:
fixes #3500


matildepark:
Took a few tries, but finally got it consistent? Reproduction case for me is just, on livenet, let an embedded link scroll far enough that it's unloaded and scroll it back down, you'll see it lost its embed. In this PR's case, the embed never breaks.

matildepark (approved):

PR #3511, opened 2020-09-18
chat: fixes scroll keys

tylershuster:
fixes #3498

PR #3512, opened 2020-09-18
chat: make backlog height only reflect loaded messages

### LINKED ISSUE #3496: chat: backlog scroll bar shouldn't reflect the entire history
Not a bug so much as preference:

While the scrollbar being as long as the chat history is theoretically neat;
the user stories for wanting to access that part of the past are not well defined;
Users find the behaviour of using it to be 'lying' or 'broken';

Therefore,
It would be better if it just reflected the page / loaded messages we have, instead.

tylershuster:
fixes #3496

Commented out to avoid merge collisions when #3511 merges.

Also only drafting because this should probably also show some loading indicator when loading more backlog

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
@@ -20,6 +20,7 @@ import { BacklogElement } from "./backlog-element";
 const INITIAL_LOAD = 20;
 const DEFAULT_BACKLOG_SIZE = 100;
 const IDLE_THRESHOLD = 64;
+const MAX_BACKLOG_SIZE = 1000;
 
 type ChatWindowProps = RouteComponentProps<{
   ship: Patp;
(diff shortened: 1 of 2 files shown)


tylershuster:
Alright, now it tells you when a fetch is pending. Review to see if it's annoying, appropriate, etc

matildepark:
It shouldn't really overlap like this — and it occurs more frequently than I was expecting.

image

What I was witnessing when I made the suggestion was that I could hit the top of the scrollbar and have no indication it was going further in the backlog — but in this implementation, it's not waiting for a pagination cue to load backlog, it's dynamically loading in backlog messages? Which is actually superior, and wouldn't really require any notice, so perhaps we can remove the notice...

tylershuster:
VirtualScroller has an `OVERSCAN_SIZE` property (100), which is the number of rows back in the log it should look. If I'm hearing you correctly I'll just remove the `|| this.state.fetchPending` from the `BacklogElement`

matildepark:
I'm just saying that if we're dynamically pulling in messages then it's ok to remove the 'past messages' banner rather than be told about stuff that should feel sort of invisible or seamless.

tylershuster:
Ok, that is what we're doing. The banner will only appear on the first load, as is the current case. When you hit the top of the backlog if you scroll fast enough you may see a brief "stick" as it fetches the messages in the background but I don't think anyone's going to lose sleep over it.

matildepark:
We're sending `NaN`s?

image

tylershuster:
well shit

tylershuster:
Helps if you give `MAX_BACKLOG_SIZE` a value

matildepark (approved):

PR #3513, opened 2020-09-18
chat: fixes unread marker behavior

### LINKED ISSUE #3495: chat: "thin green line" causes horizontal overflow
Specifically a problem on mobile.

tylershuster:
fixes #3495, ~introduces new "collapse-on-read" behavior to prevent jumping~

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/lib/ChatMessage.tsx
@@ -1,6 +1,7 @@
 import React, { Component, PureComponent } from "react";
 import moment from "moment";
 import _ from "lodash";
+import { Box } from "@tlon/indigo-react";
 
 import { OverlaySigil } from './overlay-sigil';
 import { uxToHex, cite, writeText } from '~/logic/lib/util';
@@ -12,14 +13,10 @@ import RemoteContent from '~/views/components/RemoteContent';
 
 export const DATESTAMP_FORMAT = '[~]YYYY.M.D';
 
-export const UnreadMarker = React.forwardRef(({ dayBreak, when, style }, ref) => (
-  <div ref={element => {
-    setTimeout(() => {
-      element.style.opacity = '1';
-    }, 250);
-  }} className="green2 flex items-center f9 absolute w-100" style={{...style, opacity: '0'}}>
+export const UnreadMarker = React.forwardRef(({ dayBreak, when }, ref) => (
+  <div ref={ref} className="green2 flex items-center f9 absolute w-100 left-0">
     <hr className="dn-s ma0 w2 b--green2 bt-0" />
-    <p className="mh4">New messages below</p>
+    <p className="mh4" style={{ whiteSpace: 'normal' }}>New messages below</p>
     <hr className="ma0 flex-grow-1 b--green2 bt-0" />
     {dayBreak
       ? <p className="gray2 mh4">{moment(when).calendar()}</p>
@@ -39,18 +36,19 @@ interface ChatMessageProps {
   msg: Envelope | IMessage;
   previousMsg: Envelope | IMessage | undefined;
   nextMsg: Envelope | IMessage | undefined;
-  isFirstUnread: boolean;
+  isLastRead: boolean;
   group: Group;
   association: Association;
   contacts: Contacts;
-  unreadRef: React.RefObject<HTMLDivElement>;
   hideAvatars: boolean;
   hideNicknames: boolean;
   remoteContentPolicy: LocalUpdateRemoteContentPolicy;
-  className: string;
+  className?: string;
   isPending: boolean;
   style?: any;
   scrollWindow: HTMLDivElement;
+  isLastMessage?: boolean;
+  unreadMarkerRef: React.RefObject<HTMLDivElement>;
 }
 
 export default class ChatMessage extends Component<ChatMessageProps> {
@@ -72,11 +70,10 @@ export default class ChatMessage extends Component<ChatMessageProps> {
       msg,
       previousMsg,
       nextMsg,
-      isFirstUnread,
+      isLastRead,
       group,
       association,
       contacts,
-      unreadRef,
       hideAvatars,
       hideNicknames,
       remoteContentPolicy,
@@ -84,7 +81,9 @@ export default class ChatMessage extends Component<ChatMessageProps> {
       isPending,
       style,
       measure,
-      scrollWindow
+      scrollWindow,
+      isLastMessage,
+      unreadMarkerRef
     } = this.props;
   
     const renderSigil = Boolean((nextMsg && msg.author !== nextMsg.author) || !nextMsg || msg.number === 1);
@@ -92,7 +91,7 @@ export default class ChatMessage extends Component<ChatMessageProps> {
 
     const containerClass = `${renderSigil
       ? `w-100 flex flex-wrap cf pr3 f7 pt4 pl3 lh-copy`
-      : `w-100 flex flex-wrap cf pr3 hide-child`} ${isPending ? ' o-40' : ''} ${className}`
+      : `w-100 flex flex-wrap cf pr3 hide-child`} ${isPending ? 'o-40' : ''} ${isLastMessage ? 'pb3' : ''} ${className}`
 
     const timestamp = moment.unix(msg.when / 1000).format(renderSigil ? 'hh:mm a' : 'hh:mm');
 
@@ -116,15 +115,19 @@ export default class ChatMessage extends Component<ChatMessageProps> {
       scrollWindow
     };
 
+    const unreadContainerStyle = {
+      height: isLastRead ? '1.66em' : '0',
+    };
+
     return (
       <div ref={this.divRef} className={containerClass} style={style} data-number={msg.number}>
-        {dayBreak && !isFirstUnread ? <DayBreak when={msg.when} /> : null}
+        {dayBreak && !isLastRead ? <DayBreak when={msg.when} /> : null}
         {renderSigil
           ? <MessageWithSigil {...messageProps} />
           : <MessageWithoutSigil {...messageProps} />}
-        {isFirstUnread
-          ? <UnreadMarker ref={unreadRef} dayBreak={dayBreak} when={msg.when} style={{ marginTop: (renderSigil ? "-17px" : "-6px") }} />
-          : null}
+        <Box fontSize='0' position='relative' width='100%' overflow='hidden' style={unreadContainerStyle}>{isLastRead
+          ? <UnreadMarker dayBreak={dayBreak} when={msg.when} ref={unreadMarkerRef} />
+          : null}</Box>
       </div>
     );
   }
--- a/pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
@@ -48,10 +48,12 @@ interface ChatWindowState {
   fetchPending: boolean;
   idle: boolean;
   initialized: boolean;
+  lastRead: number;
 }
 
 export default class ChatWindow extends Component<ChatWindowProps, ChatWindowState> {
   private virtualList: VirtualScroller | null;
+  private unreadMarkerRef: React.RefObject<HTMLDivElement>;
   
   INITIALIZATION_MAX_TIME = 1500;
 
@@ -61,18 +63,20 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
     this.state = {
       fetchPending: false,
       idle: true,
-      initialized: false
+      initialized: false,
+      lastRead: props.unreadCount ? props.mailboxSize - props.unreadCount : Infinity,
     };
     
     this.dismissUnread = this.dismissUnread.bind(this);
-    this.initialIndex = this.initialIndex.bind(this);
     this.scrollToUnread = this.scrollToUnread.bind(this);
     this.handleWindowBlur = this.handleWindowBlur.bind(this);
     this.handleWindowFocus = this.handleWindowFocus.bind(this);
     this.stayLockedIfActive = this.stayLockedIfActive.bind(this);
-    this.firstUnread = this.firstUnread.bind(this);
+    this.dismissIfLineVisible = this.dismissIfLineVisible.bind(this);
+    this.lastRead = this.lastRead.bind(this);
     
     this.virtualList = null;
+    this.unreadMarkerRef = React.createRef();
   }
 
   componentDidMount() {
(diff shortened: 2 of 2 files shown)


matildepark:
nit (small viewport) 
image

matildepark:
I like the lil animation but I suppose we should get design input here. @urcades ?

pkg/interface/src/views/apps/chat/components/lib/ChatMessage.tsx
```diff
@@ -116,15 +112,20 @@ export default class ChatMessage extends Component<ChatMessageProps> {
...
       <div ref={this.divRef} className={containerClass} style={style} data-number={msg.number}>
-        {dayBreak && !isFirstUnread ? <DayBreak when={msg.when} /> : null}
+        {dayBreak && !isLastRead ? <DayBreak when={msg.when} /> : null}
         {renderSigil
           ? <MessageWithSigil {...messageProps} />
           : <MessageWithoutSigil {...messageProps} />}
-        {isFirstUnread
-          ? <UnreadMarker ref={unreadRef} dayBreak={dayBreak} when={msg.when} style={{ marginTop: (renderSigil ? "-17px" : "-6px") }} />
-          : null}
+        <div className="f9 relative w-100 overflow-hidden" style={unreadContainerStyle}>{isLastRead
```
matildepark, line 126: you can get similar results with `<Box fontSize='0' position='relative' width='100%' overflow='hidden' style={unreadContainerStyle}>` from indigo-react. Why? Well we just want to phase out the atomic CSS and it makes less for us to rewrite later. Just stuff to consider when writing these divs.

urcades:
> I like the lil animation but I suppose we should get design input here. @urcades ?
> 
> Sep-17-2020 23-51-17

This is actually quite pleasant/satisfying to see, but I'm of the opinion we need to be hyper-standardized and generally principled with how we begin to introduce things like animation, sound, movement, etc. into Landscape — for ex., if we were to introduce this as a standard into chat, I'm curious how we could begin to build it as a replicable animation-expectation that one might see in other places, like comments in publish and links, in yet-to-be-built interfaces like notifications and group activity feeds, etc.

This is a note we should probably take down as a point of standardization for Indigo — cc'ing @g-a-v-i-n for this purpose. 

Do we imagine standardized animation components making their way into Indigo? What might a Landscape-brand animation curve look like? These are some questions I'd be interested in answering at some point before introducing animation into Landscape.

g-a-v-i-n:
Please don't add animation.

tylershuster:
I get that it's a bit of a pandora's box, but here's the problem I'm trying to solve: The unread marker *should* have padding around it. However, the unread marker *should* also disappear. Just removing it causes jumping. In this case, animation isn't a flourish, but a functional transition. What do you suggest to solve the problem?

g-a-v-i-n:
Got it. Here's my proposal for what to do.

- The marker should not appear if the user already has the channel open.
- The market should only disappear when the user has navigated away from the channel.

tylershuster:
Great. And what if a user sends a message to a channel with previous unreads? Seems a bit odd not to mark that channel as read. I have a bit of obsession with "Inbox Zero" so to speak, so having unread markers in the tab title nags at me, and I imagine I'm not alone.

tylershuster:
One thought I've had is to move the "New messages below" to the *right* side to avoid conflicts with text, and then it wouldn't need padding.

matildepark:
Seems ok if it just stays around the “last time you navigated here this was the unreads” place. Let’s try it and test it out.

— 
~haddef-sigwen
https://urbit.org

Fang-:
(Animation looks and feels bad on my e-ink device!)

tylershuster:
I will remove the animation.

Currently the green line is a function of the room's unread count. Seems to me that it should work this way, which requires some new modeling, and is possibly outside of the scope of this PR:

1. Message received in background
2. I navigate to page
3. Message is marked as read
  - "new message" banner goes away
  - Unread count in title goes away
  - line persists, perhaps with different copy
4. I navigate to different channel
5. I come back. There is no line

matildepark:
> One thought I've had is to move the "New messages below" to the _right_ side to avoid conflicts with text, and then it wouldn't need padding.

Design tends veer away from right-aligned text, I'm afraid.

> Currently the green line is a function of the room's unread count.

Ah, I wasn't aware of this. Honestly, I would be fine with the slight jump at this stage, but your proposal is the endgoal, I think; perhaps some sort of "if input is focused, incoming messages are auto-read" but that may produce the jump as it is.

tylershuster:
Ok, if @g-a-v-i-n and @urcades are fine with my proposal it won't take long to implement this behavior. What's one more state variable on `ChatWindow`?

urcades:
I highly recommend checking out the way slack functions, as I believe it performs exactly to the description Gav outlined above — "clearing" a "read state" is a function of one's visitation to a channel. Please check this example sourced from slack:

Kapture 2020-09-18 at 15 49 10

The above .gif occurs as I've landed into a channel, with new messages prior to my arrival. The line sticks there no matter what new messages are sent, by myself, or the other person speaking in there.

Kapture 2020-09-18 at 15 51 54

In this .gif, I navigate away after about ~20 seconds — nevermind my poking around, please be patient with the .gif

Navigating away from the channel and clicking back immediately clears the 'new' state. 

This, as a model, seems nice and simple, and seems to prevent our interface jittering about without introducing animation of any sort.

tylershuster:
This fits into the outline I described. However in addition to the line, Urbit has the "X New Messages since When" banner at the top, and the title bar number

tylershuster:
Ok, it now follows the model I outlined above. The banner and the unread marker in the title disappear on read as before (reaching the bottom of a chat), but the line stays until the channel is reloaded

jfranklin9000:
I haven't been using Landscape much lately but I will opine anyway.
My preference is to make the user manually mark what has been read
and not try to guess.  I would like per-message granularity, with islands
of read messages in a sea of unread messages.
But I'll only ask for this: Can I get one marker that only I can set/clear
(and a key binding or button to find it)?

matildepark:
> I haven't been using Landscape much lately but I will opine anyway.
> My preference is to make the user manually mark what has been read
> and not try to guess. I would like per-message granularity, with islands
> of read messages in a sea of unread messages.
> But I'll only ask for this: Can I get one marker that only I can set/clear
> (and a key binding or button to find it)?

This sort of thing will become far more feasible once notifications land in the short-term.

(I'm actually also realising we had a regression where it will jump up 1000+ messages of backlog when we want it to just not jump us up if it's that big, as the lag is quite long..)

tylershuster:
Not quite sure what you mean by this regression

matildepark:
See a2970ce92947d75466579864b436f81356e17f3c for previous backlog fetch behaviour.

tylershuster:
Roger, will fix

tylershuster:
Ok, but let's do that over here: https://github.com/urbit/urbit/pull/3512

tylershuster:
I consider this pr complete within the bounds of its scope. What @jfranklin9000 suggests would be nice and I agree but requires further engineering

matildepark:
Yes, will test anon. I think @urcades wanted to as well

matildepark:
The branch as it is right now appears to be missing this part:

> In this .gif, I navigate away after about ~20 seconds — nevermind my poking around, please be patient with the .gif
> 
> Navigating away from the channel and clicking back immediately clears the 'new' state.
> 
> This, as a model, seems nice and simple, and seems to prevent our interface jittering about without introducing animation of any sort.

I believe we still only send reads once we hit the bottom — in this case we would want to send a read action once the 'new messages below' banner is in view, since it's not tied to the read state anymore, and persists until the next time we navigate to the chat.

tylershuster:
This introduces some problems, in my mind. First, if the read state is cleared as soon as the marker becomes visible, the banner becomes just a blip. Second, it means that if I accidentally visit a room or just visit it to reference the backlog, I tell the channel that I am caught up with reading it when I am not.

matildepark:
> First, if the read state is cleared as soon as the marker becomes visible, the banner becomes just a blip.

But the banner should really only ever be visible to the user if the backlog is longer than what we're looking at. You don't want the banner to show if the "new messages below" line is visible — they're redundant.

This may mean we just don't want the backlog to jump up (? @urcades ?) so that people click the banner or scroll up to catch up, instead.

> Second, it means that if I accidentally visit a room or just visit it to reference the backlog, I tell the channel that I am caught up with reading it when I am not.

For the case where there are, like, three new messages, yes absolutely you want the unread state to gray out when you click the room. Making people force the scroll to the bottom seems odd in that case. This means we just want it to be "read" once the user sees the line, full stop. The line itself persists until we navigate away.

jfranklin9000:
> Second, it means that if I accidentally visit a room or just visit it to reference the backlog, I tell the channel that I am caught up with reading it when I am not.

Indeed, this ("I've read this") must be a manual UX action.  We're conflating "it showed up on my monitor" with "I've read it."

urcades:
There are a lot of vectors being discussed here, some which really won't find any meaningful purchase until we've developed a proper notifications-store and applied it to chat. 

Just going to address some key points of agreement so we can get this nice and cinched up as a fix:

>But the banner should really only ever be visible to the user if the backlog is longer than what we're looking at. You don't want the banner to show if the "new messages below" line is visible — they're redundant.
>
> This may mean we just don't want the backlog to jump up (? @urcades ?) so that people click the banner or scroll up to catch up, instead.

Yes to all of this. Another way of stating this mechanic is that the `30 new messages since 13:45 ... Mark as Read` banner should really only appear if the number of messages that have been posted since a user's last visit exceeds a significant/meaningful amount. I'm not really sure what this amount should be, but I'd personally expect this banner to show up when the number of posted messages since my last visit exceeded the viewport's height — which is obviously incredibly variable. We should attempt to approximate this heuristic in the most engineering-efficient manner possible.

As I was reading through this, a thought came to mind — if we refashion the `30 new messages since 13:45 ... Mark as Read` banner to effectively act as a "Click me to scroll back in the chat log to the unread line" button, does this mean that we always situate a user at the bottom of a scroll when they nav to a channel? I think this is possibly a good idea, but I'd hate to have the scroll reset to the bottom every time I visit a channel. While we might not necessarily want to attempt situating users where they "last left off" anymore, we still need a small amount of scroll memory as a user navigates around chats or other channels. I'd like for someone to explicitly address this concern when they get the time. I have a vague sense this might be best explained/discussed over a call, as it's a subtle mechanic. If any of you have access to slack channels, I highly recommend clicking around to see how they handle channel navigation/unread states.

> For the case where there are, like, three new messages, yes absolutely you want the unread state to gray out when you click the room. Making people force the scroll to the bottom seems odd in that case. This means we just want it to be "read" once the user sees the line, full stop. The line itself persists until we navigate away.

I agree completely with this particular take.

urcades:
I wanted to quickly provide an interesting reference regarding scroll memory. Please watch:

Kapture 2020-09-20 at 01 28 12

In this slack's `general` channel, I scrolled up a bit until a hit a picture of a bike.

I navigated to another channel, and scrolled around a bit, and navigated back to the original channel.

The scroll's state was sustained to a very high degree of accuracy. Even when I nudged the scroll of the motorcycle to a sliver of it's position in the channel, and navigated + scrolled another channel for quite a while, its scroll state was respected.

When I refreshed the page, the scrolls were all re-situated at the bottom of a scroll.

My hunch is that if slack's unread markers were displaying, this scroll behavior would be retained exactly. No matter what read/unread state you first enter a channel during a session, it will always display the most recent messages/bottom of the scroll initially. A user has to intentionally choose to click a button to descend into the chat's history, but while this session is still active, their scroll state will be respected either way. If the session ends (due to refresh, or some expiration mechanic), the scroll memory is forgotten and a user must view the channel from the bottom/most recent messages again.

This seems ... more or less a sane approach.

tylershuster:
I like all this, but it relies on some significant engineering, and I need to work on airlock this week. I think I'm going to back out all non-#3495-related changes to this PR and maybe we can discuss changes to behavior at a later point. I'm very interested, but this has clearly turned into something bigger than I was expecting.

matildepark:
You can triage out remembering scroll position … one thing at a time.
—
~haddef-sigwen
https://urbit.org

tylershuster:
Maybe it would help if I wrote a detailed description of what the system actually does now. As far as I know I didn't change anything in my implementation but I know it's been in flux.

tylershuster:
Ok @urcades it is easy enough to determine a user's screen height. So if someone loads a channel with 100 unread messages, do we place them at the top or the bottom of them?

urcades:
Given what I had described in my last message, in this new scheme I'd
expect a user to be situated at the bottom of the scroll, every time, while
viewing a new stack of unread messages.

tylershuster:
Making sure I understand this correctly:

a) If a user enters a room with (a lot of) unread messages, the window is positioned at the bottom of them, and a banner at the top says "A lot of unread messages since a certain time" — they can click the banner to scroll to the top of the new messages, where the green line will be waiting for them. The banner disappears, but the green line remains until next time they enter the channel.

b) if a user enters a channel with a few unread messages, the banner never appears, but the green line shows.

The one thing I haven't been able to intuit from this is *When does the `read` message get sent to the back-end?* Options are 
1) Upon entering the room
2) Upon seeing the green line
3) Only upon manual interaction

matildepark:
Try 2 in this PR, deploy it with 2, we can adjust if it's weird. But I am p sure it will be fine.

urcades:
> Try 2 in this PR, deploy it with 2, we can adjust if it's weird. But I am p sure it will be fine.

What Matilde said! Let's try this out at first. More nuanced directions (mark on user interaction) can be taken when notifications are implemented, but for the time being we should stick with something that's more or less automatic and hopefully simple.

tylershuster:
Ok, this feels pretty natural. This should be good to go.

urcades:
So excited – thanks for the work in this realm, it's huge.

—
~fabled-faster
https://urbit.org/

matildepark (approved):
The behaviour does feel nice; it's a little weird when there's a ton of backlog because it's not clear how to dismiss the unreads. I was looking at how Discord does this and basically will dismiss if you see the line *or* if you scroll and hit bottom (it pushes you up like a single page worth of unreads in this case). But I still think this is an improvement and we can keep honing away at the ideal behaviour here.

tylershuster:
The "Mark as read" button should still work, which is a nice last resort if you really don't want to deal with it

matildepark:
I actually do notice now that if you navigate to a room with unreads, see the green line, navigate away, and unreads come in, so you go back to the room, the green line never comes back?

tylershuster:
Hm...I can't replicate this given your description. Can you provide a video?

matildepark:
Can no longer replicate. I had loaded the entire backlog, then left, and came back. Could be the main thing to peek at.

matildepark:
... I meant to merge the other one. But this was an edge case and we have a few days before OTA, so no harm really.

PR #3514, opened 2020-09-18
chat: fixes drag-and-drop bugs

### LINKED ISSUE #3482: chat: drag and drop uploading visually 'flashes' as I move my cursor in the hover-area
**Describe the bug**
As I drag and drop an image into a chat channel (to upload it into the chat), I witness a visual "flashing" of the dragging/mouseDown state:

drag-upload

I suspect this occurs as one drags their mouse and the attached file over text in the chat, but I'm not 100% sure.

**To Reproduce**
1. Go to any chat-channel
2. Drag an image/file from your desktop into the chat...
3. ... as you drag the file, experiment with where dragging triggers a visual "flash" as seen in my image.

**Expected behavior**
As I drag a file from my host OS/computer into Landscape's present interface, I'd expect no visual "bugging" or "flashing" to occur at all. Just a static state-change as I perform the action.

A more nuanced design stance: We are building a machine/computer that should exude "stability" or "solidity", like a mechanical object with exceedingly tight affordances/articulation. State-changes in the interface which jitter or hop or tremble while a user interacts with them resemble a latch which doesn't close in a satisfying manner, or a button with too much wiggle before it actually activates whatever function it's linked to. We should try to reduce any visual bugs like this in any state-change interfaces we build in landscape, especially ones that take over most of the screen.

 - OS: Chrome 85.0.4183.83 (Official Build) (64-bit)
- e135b

tylershuster:
Fixes #3482 and an unfiled bug where files would post twice on safari

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/chat.tsx
+++ b/pkg/interface/src/views/apps/chat/components/chat.tsx
@@ -79,7 +79,7 @@ export class ChatScreen extends Component<ChatScreenProps, ChatScreenState> {
   }
 
   onDragEnter(event) {
-    if (!this.readyToUpload() || !event.dataTransfer.files.length) {
+    if (!this.readyToUpload() || (!event.dataTransfer.files.length && !event.dataTransfer.types.includes('Files'))) {
       return;
     }
     this.setState({ dragover: true });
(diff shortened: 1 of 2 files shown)


matildepark:
Woah the drag and drop menu even appears in Safari now.

tylershuster:
There's a menu?

matildepark (approved):

matildepark:
Er. Like. The lil dragger appearance.

tylershuster:
Yeah I guess Safari actually has a better implementation of the standard, which is probably what discord uses

PR #3517, opened 2020-09-18
all: exclude ships from invite search when shipResults is false

### LINKED ISSUE #3508: invite search: disabling shipResults still returns ships
When we pass `shipResults={false}` it removes 'ships' from the prompt and should remove all ship results.

However, as seen on the 'new group chat' screen ...

image

tylershuster:
fixes #3508

### DIFF
--- a/pkg/interface/src/views/components/InviteSearch.tsx
+++ b/pkg/interface/src/views/components/InviteSearch.tsx
@@ -151,26 +151,26 @@ export class InviteSearch extends Component<
           );
         });
 
-        for (const contact of state.contacts.keys()) {
-          const thisContact = state.contacts.get(contact) || [];
-          const match = thisContact.filter((e) => {
-            return e.toLowerCase().includes(searchTerm);
-          });
-          if (match.length > 0) {
-            if (!(contact in shipMatches)) {
-              shipMatches.push(contact);
-            }
+      for (const contact of state.contacts.keys()) {
+        const thisContact = state.contacts.get(contact) || [];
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/components/InviteSearch.tsx
```diff
@@ -151,26 +151,26 @@ export class InviteSearch extends Component<
...
-            if (!(contact in shipMatches)) {
-              shipMatches.push(contact);
-            }
+      for (const contact of state.contacts.keys()) {
+        const thisContact = state.contacts.get(contact) || [];
+        const match = thisContact.filter((e) => {
+          return e.toLowerCase().includes(searchTerm);
+        });
+        if (match.length > 0) {
+          if (!(contact in shipMatches) && props.shipResults) {
```
matildepark, line 160: This is a one-liner change, and it fixes the issue. 

But I honestly don't see *how*; this line is meant to, if we're typing a ship we don't know of yet, but it's a valid patp, show it anyway — 

However, on livenet we see that all results just show normally?

image

matildepark (approved):

pkg/interface/src/views/components/InviteSearch.tsx
tylershuster: As I understand it, this code has two steps: 
1. Search peers
2. Search contacts
It was the second that didn't have the `props.shipResults` check

matildepark: OK, I misunderstood where we were in the checks.

PR #3520, opened 2020-09-18
ames: print failed scrys

belisarius222:
This should help us debug failing scry requests related to stateless forwarding.  `|ames-verb %for` turns on printing in the `~` case.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -756,6 +756,8 @@
           =([%$ %da now] lot)
           =(%$ syd)
       ==
+    ?.  for.veb.bug.ames-state  ~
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Looks like a printf to me!

Someone remind me why we have both `a+b` and `a/b` for `[%a b]` shorthand.

belisarius222:
Because we never got around to removing the older `a+b` syntax.

—
~rovnys-ricfer
https://urbit.org

joemfb:
Merging, as I want to make sure this makes the next release.

ohAitch:
Directly related to us having both `a/b` and `a=b` syntax: not all of the former of each has been cleaned out of the codebase.

ohAitch:
(actual removal will ofc want a `!?` and `/?` dependency bump, but we should at least be able to turn it _off_)

PR #3521, opened 2020-09-18
graph import/export generators

belisarius222:
To export a graph:
```
+graph-store/export-graph our %graph-name
```
The output can be routed to a jammed Unix noun by using :dojo's `.file/ext` syntax.  This file can then be copied into a Unix mount of an Urbit desk, where it should be given the `.graph-update` file extension so that the Clay file has the right mark.

To import a graph from a Clay file:
```
:graph-store|import-graph %new-graph-name %/full/path/to/graph-update
```

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -474,6 +474,20 @@
       now.bowl
     [%add-graph [ship term] `graph:store`p.u.result q.u.result]
   ::
+      ::  note: near-duplicate of /x/graph
+      ::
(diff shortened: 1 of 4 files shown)


tacryt-socryp:
Merge base should be release/link

belisarius222:
Thanks, I thought I'd set that, but apparently not.  Fixed.

tacryt-socryp (approved):
Please remove the printf on line 452 of graph-store. Otherwise, looks great!

belisarius222:
Fixed, thanks

PR #3522, opened 2020-09-18
vere: fix typo in secp reco jet

frodwith:
ecdsa-raw-recover had an erroneous jet. This commit fixes the critical
error so that the jet returns the right result according to the secp
secp256k1 spec, although it still does not match the hoon. The hoon is
subtly wrong, and will be addressed in a future commit.

### DIFF
--- a/pkg/urbit/jets/e/secp.c
+++ b/pkg/urbit/jets/e/secp.c
@@ -207,7 +207,7 @@ u3qe_reco(u3_atom has,
   ret = secp256k1_ecdsa_recover(ctx_u,                      /* IN: context */
                                 & puk_u,                     /* OUT: pub key */
                                 & sig_u,                        /* IN: signature */
-                                (const c3_y *) & has);        /* IN: message has */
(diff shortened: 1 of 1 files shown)


pkg/urbit/jets/e/secp.c
```diff
@@ -207,7 +207,7 @@ u3qe_reco(u3_atom has,
   ret = secp256k1_ecdsa_recover(ctx_u,                      /* IN: context */
                                 & puk_u,                     /* OUT: pub key */
                                 & sig_u,                        /* IN: signature */
-                                (const c3_y *) & has);        /* IN: message has */
+                                has_y);                       /* IN: message has */
```
joemfb, line 210: That's brutal. Looks like it would work for direct atoms, which is probably how it got past smoke-testing.

joemfb (approved):
LGTM. (And sure for the PR target churn, I had to update `release/next-vere` and temporarily messed it up.)

PR #3524, opened 2020-09-18
term: don't print styles in non-interactive mode

Fang-:
As per https://github.com/urbit/urbit/pull/3480#discussion_r487219868.


joemfb (approved):
LGTM

Issue #3489, opened 2020-09-15
Date for prev / next publish posts i wrong


galenwp:
**Describe the bug**
The previous and next links for publish posts have incorrect dates on them. 

**Screenshots**
<img width="555" alt="Screen Shot 2020-09-15 at 8 49 02 AM" src="https://user-images.githubusercontent.com/1467211/93234425-1ec66c80-f731-11ea-9e97-cccc3f6a74f3.png">

PR #3526, opened 2020-09-19
publish: correct timestamps on note navigation

matildepark:
Addresses #3489.

Because we didn't access `date` through the function's `props` it was `undefined`; moment, too shy to say anything, just decided to use the timestamp for *right this second* instead.


liam-fitzgerald (approved):
LGTM!

PR #3529, opened 2020-09-19
various: s3 dragger uses indigo-react

matildepark:
The dragger overlay doesn't have dark mode styles; this rewrites the component in indigo-react so that its colour scheme changes to match.

cc @urcades — anything we need to distinguish this further?

Screenshot 2020-09-18 at 10 08 00 PM

### DIFF
--- a/pkg/interface/src/views/components/s3-upload.tsx
+++ b/pkg/interface/src/views/components/s3-upload.tsx
@@ -1,5 +1,5 @@
 import React, { Component } from 'react'
-import { Icon } from "@tlon/indigo-react";
+import { Box, Text, Icon } from "@tlon/indigo-react";
 
 import S3Client from '~/logic/lib/s3';
 import { Spinner } from './Spinner';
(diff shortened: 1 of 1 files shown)


urcades:
This looks just as expected: spiritual stamp

pkg/interface/src/views/components/s3-upload.tsx
```diff
@@ -1,16 +1,30 @@
...
+    top='0'
+    bottom='0'
+    left='0'
+    right='0'
+    position='absolute'
+    backgroundColor='white'
+    height='100%'
+    width='100%'
+    display='flex'
+    alignItems='center'
```
liam-fitzgerald, line 20: It's probably cleaner to write this a `<Center>` because it automatically sets the height, flex, and alignment props to the one you want.

liam-fitzgerald (changes requested):
Tiny nit, looks good otherwise.

pkg/interface/src/views/components/s3-upload.tsx
matildepark: It apparently does not support `backgroundColor` in 1.1.5 so we don't get the colour scheme stuff here.

liam-fitzgerald: Ah don't worry about it then!

liam-fitzgerald (approved):
LGTM!

PR #3533, opened 2020-09-19
chat: fixes misses pending messages

### LINKED ISSUE #3497: chat: pending messages sometimes not appearing in chat
Not *consistently* an issue, it appears to happen in rooms with big backlogs and thus higher probability of messages being out of order.

Suspected to be a chat-store dependent issue, or else our virtual scroll implementation should enforce front-end side numbering? (If it does, does that just produce weirder issues?)

tylershuster:
fixes #3497


matildepark:
... oh ok cool

matildepark (approved):
Limited cases I can test with, but it does what it should. Why's `mailboxSize` a different size than `envelopes.length`

tylershuster:
`mailboxSize` is every message. `envelopes.length` is the number of currently browserside messages. A dumb mistake on my part.

PR #3535, opened 2020-09-19
chat: sort message by number instead of timestamp

### LINKED ISSUE #3534: Chat: Message attributed to incorrect author
**Describe the bug**
I replied to a post in a chat, with a URL, followed by some words.

In the console, both URL and words are correctly attributed to me.

As viewed in Landscape, in the chat, the words are attributed to the previous writer.

**To Reproduce**
I cannot reproduce this on demand.

In the screenshots below, the console screenshot correctly shows the order of events:

~botter-nidnul writes
"the rumor with[...]"
I reply with the URL, and the text "is good - on[...]"

image

image

(images have been edited to block some text, I believe the bug remains clear)

Issue remains after a shift+reload.
Issue remains if I use an incongito mode window. 
Issue remains if I use Chrome instead.
Issue remains if I use Safari on my iphone instead.

However the chat messages display correctly in the "Ursus chat" IOS Urbit client app.

 - OS: [e.g. MacOS 10.15.5]
 - Browser [Brave 1.13.86]
 - Base hash of your urbit ship. 0v1c.o253s.lpvsm.t4fr4.en39p.mq1rj.qjnrk.tjnrl.c5t1i.o68th.naoqh

tylershuster:
fixes #3534

This is correct behavior.
May cause issues with some people who have messages stored out of order on their computer.
Leaving and rejoining the channel should fix it for them.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
@@ -228,7 +228,7 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
     let lastMessage = 0;
     
     [...envelopes]
-      .sort((a, b) => a.when - b.when)
+      .sort((a, b) => a.number - b.number)
       .forEach(message => {
         messages.set(message.number, message);
         lastMessage = message.number;


Fang-:
Kinda tempted to revive discussion from #1889 here. If sorting by sequence number is the correct behavior (and it probably is) then timestamps are purely cosmetic, and so we should be fine with senders setting them themselves.

Might be moot wrt the transition to graph-store though, depending on how that presently handles these things.

matildepark:
This is the scary PR! Curious if we can get @tacryt-socryp take — but probably, since graph-store will necessitate this(?), it's a bandaid we should take off.

matildepark:
Testing and comparing backlog on General and Watercooler between haddef and test ship, they're identical. So I think the original bug was sort of rare as it was?

matildepark:
~minweb-datsyt is the only person I know who had the backwards chat problem, so I asked them to check out this PR and try it out — 

> ok so, chats that were backwards are still backwards, and leaving/rejoining does fix them

Which I think is acceptable enough given most people did not get this issue and this is easily resolvable.

PR #3538, opened 2020-09-21
all: sets scrollbar colors

tylershuster:
fixes urbit/landscape#87

Also fixes reversed scroll dragging in Chrome...?

### DIFF
--- a/pkg/interface/src/views/App.js
+++ b/pkg/interface/src/views/App.js
@@ -39,6 +39,24 @@ const Root = styled.div`
   }
   display: flex;
   flex-flow: column nowrap;
+
+  * {
+    scrollbar-width: thin;
+    scrollbar-color: ${ p => p.theme.colors.gray } ${ p => p.theme.colors.white };
+  }
+  
+  /* Works on Chrome/Edge/Safari */
+  *::-webkit-scrollbar {
+    width: 12px;
+  }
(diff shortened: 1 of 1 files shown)


matildepark:
> Also fixes reversed scroll dragging in Chrome...?

*How*?

tylershuster:
The line that specifies the scrollbar width seems to invoke some deep magic at the core of Chromium's implementation that makes it think a bit harder about what it's doing. Remove that line and you don't get any of the fixes.

tylershuster:
FF Light:
image
FF Dark:
image
Chrome Light:
image
Chrome Dark:
image

tacryt-socryp:
Safari?

tylershuster:
Safari is Webkit, so it looks the same as Chrome

tylershuster:
(These are webkit prefixes, which Chrome respects)

matildepark:
Safari works fine with "always show scrollbar" on, as said — 

image

And it indeed does fix the backwards scroll stuff. I do not even begin to get it.

matildepark (approved):

tylershuster:
Maybe it takes control over an implementation that's normally controlled by the OS

matildepark:
Dark mode much cleaner (livenet on right). 
image

PR #3539, opened 2020-09-21
update/rewrite secp256k1 hoon code

frodwith:
The previous version of this code had incorrect jet matching semantics (constants in a sample that affect the result but are not checked by the jet) and didn't match the jet w.r.t. the "recovery id" (v) part of the signature. The code has been restructured and tests added.


joemfb:
This has been merged out-of-band, into `na-release/candidate`.

Issue #3268, opened 2020-08-10
OS 1 Visual Bug: Publish list item line-height


urcades:
Noticed a strange discrepancy in the third level of publish lists:

<img width="514" alt="Screen Shot 2020-08-10 at 1 13 08 PM" src="https://user-images.githubusercontent.com/1195363/89810659-67c04b00-db0b-11ea-835c-fa88e894f793.png">

This spacing is a lot more than I'd expect

<img width="566" alt="Screen Shot 2020-08-10 at 1 13 05 PM" src="https://user-images.githubusercontent.com/1195363/89810664-6a22a500-db0b-11ea-9650-7ed169a57351.png">

Upon inspection, it also looks like this element is turned into a nested `<ul>` for some reason

matildepark:
Blocked on #3337 merge.

PR #3540, opened 2020-09-22
interface: bump indigo to 1.2.6

liam-fitzgerald:
Bumps indigo to a hypothetical v1.2.5. This version should be the result of merging https://github.com/urbit/indigo-react/pull/63 into master. 

- Introduces ProseMirror for editing of publish notes

Should wait for the indigo changes to be merged, then I'll change the package.json ver to 1.2.5


pkg/interface/src/views/apps/profile/components/lib/BackgroundPicker.tsx
```diff
@@ -1,9 +1,14 @@
...
+  ManagedRadioButtonField as Radio,
+  ManagedTextInputField as Input,
+} from "@tlon/indigo-react";
 
-import GlobalApi from '~/logic/api/global';
-import { S3State } from '~/types';
-import { ImageInput } from '~/views/components/ImageInput';
+import GlobalApi from "~/logic/api/global";
+import { S3State } from "~/types";
+import { ImageInput } from "~/views/components/ImageInput";
```
matildepark, line 11: I believe our linter for pkg/interface uses single quotes; we should avoid committing this stuff back and forth in future.

pkg/interface/src/views/components/Tab.tsx
```diff
@@ -0,0 +1,31 @@
...
+);
+
+export const Tabs = ({ children, ...rest }: Parameters<typeof Row>[0]) => (
+  <Row
+    bg="white"
+    mb={2}
+    justifyContent="stretch"
+    alignItems="flex-end"
+    {...rest}
+  >
```
matildepark, line 27: Needs `cursor: pointer`.

matildepark: Also the notebook list items aren't respecting the margin bottom here. Is it an inline item?
Screen Shot 2020-09-22 at 1 33 30 PM

matildepark:
More margin tweaks to make (livenet, then dev environment):

Screen Shot 2020-09-22 at 1 36 41 PM

Screen Shot 2020-09-22 at 1 36 47 PM

matildepark:
@g-a-v-i-n @urcades 

image

g-a-v-i-n:
From small to big:

- The button should just say 'Publish'
- Can you give me more info on why the button is offset vertically (or vic versa)? It might be Indigo related
- Does the markdown rendering in the editor match the rendering on the post?
- This is not using indigo markdown style. They only exist in Figma
- Very cool to see the 1.2 bump, nice work Liam
- Why is the 1.2 bump in the same PR as ProseMirror?
- None of the controls on prosemirror have been designed, because this is out of cycle and was never designed

Summary:

Don't like the idea of adding prosemirror without design work, plus seems like 2 separate PRs to me
Happy to see indigo 1.2 in the wild

matildepark:
If I had to guess (since Liam's sleeping) I think it was introduced because of some ambient design bugs with Publish layout (#3268 etc) due to us using CodeMirror; and indigo deprecating its editor introduced a decision to make as to what bugs to stomp or whatnot. But yeah it introduces a lot to consider.

matildepark:
The input and the button seem to be in independent divs? The input has 8px margin-top and the button has no margin, but its parent div does, 4px. So they're separated by 4px.

g-a-v-i-n:
Got it, that's def an indigo issue, will fix for 1.2.5 release today.

I'm happy to do some work on text editing (after notifications) because I think it's a good move in the right direction.

liam-fitzgerald:
So should I update this PR to use CodeMirror?

matildepark:
Yeah just maintain parity with livenet for now.

liam-fitzgerald:
Should be good now
cc: @g-a-v-i-n there's some Radio/Checkbox breakage, but it appears to be resolved by adding `display: flex` to the Indicator component. 
Screen Shot 2020-09-23 at 1 15 10 pm

matildepark:
still blocked on
- the breakage you mentioned haha

> The button should just say 'Publish'

(Not "Publish to [xxxaax]")

- Buttons and input are now topsy turvy in the opposite direction

image

We lost a border here — on purpose?

image

g-a-v-i-n:
It would be really helpful to have a line permalink for some of these screenshots. I can't repro this checkbox/radio button issue on the Indigo test page.

matildepark:
Sure thing. Here's the locations in the latest commit's tree:

- [Breakage in settings](https://github.com/urbit/urbit/blob/1f95c40a24093807116bf4fa2156ba61125b1a2f/pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx#L61-L83).
- [Publish](https://github.com/urbit/urbit/blob/1f95c40a24093807116bf4fa2156ba61125b1a2f/pkg/interface/src/views/apps/publish/components/lib/NoteForm.tsx#L32-L59).
- [Uploading an image form](https://github.com/urbit/urbit/blob/1f95c40a24093807116bf4fa2156ba61125b1a2f/pkg/interface/src/views/components/ImageInput.tsx#L61-L88).

matildepark:
Unsure but it's possible you would use 1.2.6 and the `TwoUp` component for the elements in `NoteForm`?

liam-fitzgerald:
bump @matildepark be nice to have this before the weekend

matildepark:
I agree. I am going to test it anon!

—
~haddef-sigwen
https://urbit.org

matildepark:
Weird stretch here — 
image

If you're sleeping I'll try to patch this myself after the next few meetings, I can't seem to rig it the way I want atm

g-a-v-i-n:
I would not use TwoUp here, not what its for.

matildepark:
We can try a regular Row and see how it goes.

—
~haddef-sigwen
https://urbit.org

g-a-v-i-n:
ngl... I'm optimistic about Row

matildepark (approved):

PR #3542, opened 2020-09-22
permission: fully deprecate and remove unnecessary logic

tacryt-socryp:
This has been unused for a few months now. Might as well get rid of unnecessary logic.


belisarius222 (approved):
LGTM

PR #3543, opened 2020-09-22
link: joining graph resources from the groups screen works properly

tacryt-socryp:


matildepark (approved):
lgtm, reading it over — the `export default` is curious since that usually produces a breaking change in the build process between, say, `import { LinksApp } from LinksApp` or `import LinksApp from LinksApp`; but the latter is already in use, so I guess it just works anyhow. Happy to see the catch for rendering app icons (even if it's a temporary shim)

belisarius222 (approved):
LGTM

PR #3544, opened 2020-09-22
links: full migration to graph-store

tacryt-socryp:
This PR includes:
- migration of Links data to Graph store
- archival of Links data within Graph store
- an HTTP interface added to spider
- threads that act as the functionality necessary for a Graph view
- validator mark for Link data
- JS interface for Links moved to Graph store


matildepark:
I emailed you a list of regressions. Did you address those? For most you can just pull from release/next-userspace the exact logic. You can't upload files to s3 or drag and drop into this, last I tested.

tacryt-socryp:
Note: I have tested OTAing this to ~tacryt-socryp and it works well.

matildepark:
Do we have confirmation that the following case works in full?

1. Livenet ship has link collections
2. Links into graph store migration occurs during the OTA 
3. The ship is able to export and unarchive the collection in a group they run.

tacryt-socryp:
I have now fixed all the UI regressions other than a single one with the ordering of the OrderedMap js type, which I'm not sure how to fix. I imagine Liam will have thoughts when he comes online.

I have tested the OTA and it worked on ~tacryt. I have not tested export or unarchive. Have pinged Ted and we're discussing / testing that part soon.

pkg/interface/src/views/apps/links/components/lib/link-preview.js
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+import { cite } from '~/logic/lib/util';
+import RemoteContent from "~/views/components/RemoteContent";
+
+import moment from 'moment';
+
+
+const URLparser = new RegExp(
+    /((?:([\w\d\.-]+)\:\/\/?){1}(?:(www)\.?){0,1}(((?:[\w\d-]+\.)*)([\w\d-]+\.[\w\d]+))){1}(?:\:(\d+)){0,1}((\/(?:(?:[^\/\s\?]+\/)*))(?:([^\?\/\s#]+?(?:.[^\?\s]+){0,1}){0,1}(?:\?([^\s#]+)){0,1})){0,1}(?:#([^#\s]+)){0,1}/
```
liam-fitzgerald, line 9: is there any reason for this (giant, unmaintable) regex instead of say 
``` typescript
let hostname = '';
try {
  const parsed = new URL(url);
  hostname = parsed.hostname;
} catch(e) {}

File: pkg/arvo/sur/metadata-store.hoon [hunk 2]
```diff
@@ -1,16 +1,18 @@
...
-      description=@t
-      color=@ux
-      date-created=@da
-      creator=@p
+  $:  title=cord
+      description=cord
+      =color
+      date-created=time
+      creator=ship
+      module=term
```
liam-fitzgerald, line 15: This is missing a corresponding change to `api/metadata.ts` to include the `module` field, will cause all metadata updates from the client to fail.

liam-fitzgerald (changes requested):
Read through, seems good, most of this has already been reviewed. Just the metadata api stuff blocking this, and a q about URL parsing.

pkg/interface/src/views/apps/links/components/lib/link-preview.js
matildepark: [This isn't an introduced change](https://github.com/urbit/urbit/blob/master/pkg/interface/src/views/apps/links/components/lib/link-detail-preview.js#L61-L63), it's on livenet.

If it's the one I originally wrote — I believe we want to parse the hostname for, eg.`gopher://` or `ipfs://` or whatever schema and this doesn't? Or there was some other weird catch? Ancient history really. But out of scope here, better suited for later refactor work

liam-fitzgerald: Ok sweet, no worries then, I wasn't aware it was already there.

pkg/arvo/sur/metadata-store.hoon
tacryt-socryp: Fixed this.

matildepark (approved):

philipcmonk:
I'm not sure if this is exactly where it happened, but the link-store upgrade logic is a real pain to work with right now because its state now contains cards and vases, which are difficult to work with.  In particular, including cards in your state means any change to the card type means you need a state adapter.  Usually you can work around this by reading in as `*` instead of `card` and `;;`ing to the new card type, but in this case those cards themselves include vases (for poking), which means you can't `;;`.

I'm going to see what I can do about this in this case, but the rule to keep in mind is that keeping either cards or vases in your state is risky, and mixing them is even worse.

As a side note, this was caught specifically by testing on a -L copy of ~wicdev -- it worked fine on fresh fake ships booted from a master pill, since they had no link data to speak of.  This is why using real data is important!

philipcmonk:
I suppose the answer for this case is that we can say the cards were "already used" and are just hanging around in our state for no reason (ie they could be discarded in +on-arvo).  So, in `+on-load` we will expect `[%1 cards=*]` and just throw away anything we get.

I argue this is fine because:

- We don't expect "long range" upgrades past `%1` because this upgrade is link-store's final goodbye
- The cards should already have been sent, and if they weren't we don't have any way to rescue anyway

PR #3545, opened 2020-09-22
chat: fixes message jump

tylershuster:
@matildepark @vvisigoth 

This should fix the issue you noticed where posting an image would cause the chat to jump. It was caused by an old method that did more harm than good in the new implementation.


matildepark (approved):
Works in the case mentioned

matildepark:
@tylershuster This has reintroduced scroll jumping when one is scrolling down.

tylershuster:
Can you describe the behavior more?

matildepark:
I’m on Safari, mind — 

Go to a chat with a backlog. Click the banner to jump up to the line and scroll down. It will push you up as you scroll. Most egregious if there’s lots of images. Can record probably...

—
~haddef-sigwen
https://urbit.org

tylershuster:
No, I can imagine the behavior you're describing. It doesn't have anything to do with this PR though. I'll file an issue with some explanation why it's happening so I can get to it

PR #3546, opened 2020-09-22
chat: fixes unread marker on blur

tylershuster:
Just some light refinements on how the unread markers behave. Now when you refocus the window, it will mark messages as read if you're at the bottom, thus clearing the banner and the tab icon, but the line will remain.


tylershuster:
This is not finished, oops

tylershuster:
Now, if you leave the window, and messages come in, the marker will appear (presumably you're not looking at it so the jump doesn't matter), and when you revisit the window, the marker will stay there but the messages will be marked as read (when you're at the bottom).

matildepark (approved):
Feels nice.

PR #3548, opened 2020-09-23
leap: destructure selection logic variables

matildepark:
A very tiny PR while I'm debugging Leap in Safari 14 — just amending some of these repeated assignments into a single destructure makes the selection logic way easier to follow? — places emphasis on *where* in the array we assign in each case instead of just seeming like code.


liam-fitzgerald (approved):
You can literally never have enough destructuring. LGTM!

Issue #3437, opened 2020-09-05
king-haskell: very high cpu usage after awhile


eglaysher:
This morning I found the king haskell process taking up 50% to 200% CPU on a multicore machine, with a rough average of 150%. It had been up for about 20 hours. This doesn't appear to be causing any keyboard latency in dojo, and appears to have no side effects.

Restarted the process and waited 10 minutes, didn't reoccur so I don't think it was a network packet storm or something.

eglaysher:
(This happened again after being up for a few days.)

eglaysher:
This is still happening, and I've rebuilt my urbit-king binary with --profiling, and am going to run KH with `+RTS -s -hc -p` for a couple of days to see if I can track this down better.

eglaysher:
Very basic profiling for 10 minutes. Will also let it run for a day.

```
>>=.\                                 Urbit.Noun.Cue                   lib/Urbit/Noun/Cue.hs:(134,39)-(136,32)                   8.2    3.0
primitive                             Control.Monad.Primitive          Control/Monad/Primitive.hs:195:3-16                       5.5    8.0
fmap.\                                Urbit.Noun.Cue                   lib/Urbit/Noun/Cue.hs:(107,36)-(109,35)                   4.9    8.8
<*>.\                                 Urbit.Noun.Cue                   lib/Urbit/Noun/Cue.hs:(116,44)-(119,37)                   4.8    9.2
throwSocketErrorIfMinus1RetryMayBlock Network.Socket.Internal          Network/Socket/Internal.hsc:(162,1)-(163,53)              4.4    0.3
putStr                                ClassyPrelude                    src/ClassyPrelude.hs:587:1-31                             4.1    4.7
>>=.\                                 Urbit.Noun.Jam                   lib/Urbit/Noun/Jam.hs:(233,35)-(235,28)                   3.7    1.1
findSafeSlots.go                      Data.HashTable.ST.Basic          src/Data/HashTable/ST/Basic.hs:(570,5)-(629,48)           2.3    0.6
*>.\                                  Urbit.Noun.Cue                   lib/Urbit/Noun/Cue.hs:(122,43)-(124,20)                   1.9    1.3
writeArray                            Data.HashTable.Internal.Array    src/Data/HashTable/Internal/Array.hs:43:1-25              1.8    1.8
<*>.\                                 Urbit.Noun.Jam                   lib/Urbit/Noun/Jam.hs:(215,40)-(218,37)                   1.8    2.8
[...]

PR #3549, opened 2020-09-23
Fix a bunch of strictness issues in King Haskell

eglaysher:
These are a grab bag of space leak issues that I noticed when trying to fix the increasing heap size after I started doing heap profiling. It should be easier to review this PR commit by commit.

I left my Urbit on with this patch for about 20 hours, and the heap usage as recorded with `+RTS -N -s -hc -p -L500` is now flatline instead of up and to the right.

The patch which gives 90% of the results is the `fromNoun`/`NFData` patch, which made noun conversion thunks disappear from the heap profile as a hot spot, and also appears to have cut the amount of PINNED memory significantly.

Partial fix for #3437.


pkg/hs/urbit-king/lib/Urbit/Vere/Term/Logic.hs
```diff
@@ -86,7 +86,7 @@ step st@St{..} = \case
     word = fromIntegral
 
     record :: Text -> St -> St
-    record t st@St{..} = st { sHistory = trim (sHistory |> t) }
+    record t st@St{..} = st { sHistory = force $ trim (sHistory |> t) }
```
eglaysher, line 89: `record !t ...`

pilfer-pandex, line 89: You shouldn't need `force` here, because seqs are strict in their length.

pkg/hs/urbit-noun-core/lib/Urbit/Noun/Core.hs
```diff
@@ -37,6 +37,12 @@ data Noun
     = NCell Int Word Noun Noun
     | NAtom Int Atom
```
eglaysher, line 38: Remove NFData instances and make `!Atom` and `!Noun !Noun`.

eglaysher (commented):
TODO list from in call with pilfer

pkg/hs/urbit-noun-core/lib/Urbit/Noun/Convert.hs
```diff
@@ -88,13 +88,13 @@ fromNoun :: FromNoun a => Noun -> Maybe a
 fromNoun n = runParser (parseNoun n) [] onFail onSuccess
   where
     onFail p m  = Nothing
-    onSuccess x = Just x
+    onSuccess x = Just $ force x
```
pilfer-pandex, line 91: can't hurt: do `onSuccess !x` and on the other two.

we also need a convention that everything we FromNoun have strict fields.

pilfer-pandex (commented):
Review mostly by phone. Outcome: Run an experiment where we don't have nfdata, but do have ! on the nouns and cells in NCell, NAtom, ! on all the fields of data structures which are FromNoun. If this works, do that; otherwise, do this nfdata thing even though we don't fully understand why it works.

eglaysher:
The experiment was a success. The combination of StrictData on all data structure files and liberally sprinkling `!` on  the noun and parsing stuff got the same result as making everything NFData.

pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE StrictData #-}
```
pilfer-pandex, line 1: I have mixed feelings about this. Basically, it makes readability worse (you have to remember that the meaning of a data definition in these files is different), but writability better (basically no chance of accidentally introducing laziness when you add a new ev/ef). It's not clear to me what the right call is. For what it's worth, I've only ever seen this used as a project-wide choice. But I'll sign off on it for now.

pkg/hs/urbit-king/lib/Urbit/Vere/Serf/IPC.hs
```diff
@@ -1,3 +1,5 @@
+{-# LANGUAGE StrictData #-}
```
pilfer-pandex, line 1: Since IPC defines not only the ipc types (analogous to ev/ef types) but also complex dataflows, I feel less confident about this than I do about ev/ef. What if billy, in the course of implementing some dataflows, introduces an intermediate datatype? One way to discharge this concern would be to move the writ,plea,etc definitions into a separate module IPC.Types.

pilfer-pandex (approved):
Fabulous. I had a feeling that this would work, and it's great when your models are confirmed. Thanks so much for running the experiment, Elliot.

I have some concerns about the use of LANGUAGE StrictData, documented above and below. Before shipping, please think about them and do what seems appropriate.

PR #3550, opened 2020-09-23
meta: add link to file design issues

matildepark:
While Landscape source code is bundled with the Arvo repository, not all Landscape issues belong in the urbit repo. That is, urbit/urbit should concern itself with functionality and livenet issues, and QA or design-specific reports should be handled in its own forum. 

Commits can still close cross-repo issues using the repo format, eg. urbit/landscape#25.

cc @urcades


matildepark:
ping @jtobin @philipcmonk

philipcmonk (approved):
Sorry missed this

PR #3553, opened 2020-09-24
eyre: remove superfluous connection-state check

pkova:
This was originally introduced by me in #1814 to address #1811. Eyre was not canceling heartbeat timers on all relevant events making it easy to end up with an infinite behn loop. This check allowed ships that entered an infinite loop to recover, as per my comment at https://github.com/urbit/urbit/pull/1814#discussion_r333477482. Otherwise it's not necessary.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:84d815edefd8cf64598cce4b06bd1b3038a265222ada9d45c193a817ce5580bf
-size 6297329
+oid sha256:9b572ea1bc3bafc9ac7ff1a0fcecca16320bc2bbe6d6964cc76a2bc36f201975
+size 6299579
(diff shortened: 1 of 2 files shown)


pkova:
Ping @eglaysher

eglaysher (approved):
LGTM. In the future, you should probably use the reviewer request system on the right hand side of the PR because otherwise I don't get a label in my inbox.

matildepark:
I don't think contributors [can](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/requesting-a-pull-request-review).

> Note: Pull request authors can't request reviews unless they are either a repository owner or collaborator with write access to the repository.

But marking the PR with a cc should be suitable enough?

eglaysher:
(Is there any other action that I need to take here? ie, can @pkova merge the approved patch, or is that something else external people can't do?)

matildepark:
Only 'core' members of our org can. Normally maintainer of the section merges I think.

PR #3554, opened 2020-09-24
zuse: make octs value @ instead of @t

Fang-:
Because these things aren't always printable/valid `@t`.

There's maybe some argument to be made here around `@t` indicating a specific byte order, but `octs` gets used for both cords and arbitrary bytestreams, so...

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -79,7 +79,7 @@
 ::
 ::    TODO: Rename to +mime once the current +mime and +mite are gone. The
 ::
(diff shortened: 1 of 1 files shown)


ohAitch:
Hah, did we run into a use case where they actually weren't?

joemfb:
@ohAitch, I've run into this a few times serializing arbitrary data (in the LetsEncrypt integration, for example) and was complaining about it to @Fang-.

There's also the `$byts` type in hoon.hoon, used in the crypto implementations and documented to be MSB-first. We should similarly establish the byte-order of `$octs` and reconcile those types.

joemfb (approved):
LGTM

PR #3557, opened 2020-09-24
vere: webslogs

Fang-:
This adds an http endpoint (`/~/slog`) to the existing http server in vere, on top of the bindings that eyre makes. It does SSE, sending pre-processed lines of output to all open connections whenever the runtime receives a `%slog` from the worker.

This can be useful for gathering debugging output from ships in hosting contexts, and can bring web dojo closer to the real terminal experience.

The matching web dojo PR is following Soon™, but this should be ready for initial review on its own.

~We... may or may not want to hold off on merging this until we put authentication on it. We don't have an http auth mechanism in the runtime yet, but there's probably something we can do along the lines of scrying into eyre with "is this a valid session cookie"?~  
Authentication is in as of d0d9437.

### DIFF
--- a/pkg/urbit/vere/io/http.c
+++ b/pkg/urbit/vere/io/http.c
@@ -91,6 +91,7 @@ typedef struct _u3_h2o_serv {
   typedef struct _u3_hfig {
     u3_form*         for_u;             //  config from %eyre
     struct _u3_hreq* seq_u;             //  open slog requests
+    uv_timer_t*      sit_u;             //  slog stream heartbeat
   } u3_hfig;
 
 /* u3_httd: general http device
@@ -107,6 +108,7 @@ static void _http_serv_start_all(u3_httd* htd_u);
 static void _http_form_free(u3_httd* htd_u);
 
 static const c3_i TCP_BACKLOG = 16;
+static const c3_w HEARTBEAT_TIMEOUT = 20 * 1000;
 
 /* _http_close_cb(): uv_close_cb that just free's handle
 */
@@ -1911,6 +1913,27 @@ _http_stream_slog(void* vop_p, c3_w pri_w, u3_noun tan)
   u3z(tan);
 }
 
+/* _http_seq_heartbeat_cb(): send heartbeat to slog streams and restart timer
+*/
+static void
+_http_seq_heartbeat_cb(uv_timer_t* tim_u)
+{
+  u3_httd* htd_u = tim_u->data;
+  u3_hreq* seq_u = htd_u->fig_u.seq_u;
+
+  if ( 0 != seq_u ) {
+    u3_noun dat = u3nt(u3_nul, 1, c3_s1('\n'));
+    while ( 0 != seq_u ) {
+      _http_continue_respond(seq_u, u3k(dat), c3y);
+      seq_u = seq_u->nex_u;
+    }
+    u3z(dat);
+  }
+
+  uv_timer_start(htd_u->fig_u.sit_u, _http_seq_heartbeat_cb,
+                 HEARTBEAT_TIMEOUT, 0);
+}
+
 /* _reck_mole(): parse simple atomic mole.
 */
 static u3_noun
@@ -2081,6 +2104,12 @@ u3_http_io_init(u3_pier* pir_u)
 
   pir_u->sop_p = htd_u;
   pir_u->sog_f = _http_stream_slog;
+
+  uv_timer_t* sit_u = c3_malloc(sizeof(*sit_u));
+  sit_u->data = htd_u;
+  uv_timer_init(u3L, sit_u);
+  uv_timer_start(sit_u, _http_seq_heartbeat_cb, HEARTBEAT_TIMEOUT, 0);
+  htd_u->fig_u.sit_u = sit_u;
 
   //  XX retry up to N?
   //
--- a/pkg/urbit/vere/io/http.c
+++ b/pkg/urbit/vere/io/http.c
@@ -824,6 +824,8 @@ _http_req_prepare(h2o_req_t* rec_u,
   return seq_u;
 }
 
+/* _http_seq_continue(): respond to slogstream request based on auth scry result
+*/
 static void
 _http_seq_continue(void* vod_p, u3_noun nun)
 {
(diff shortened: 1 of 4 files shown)


pkg/urbit/vere/io/http.c
```diff
@@ -1732,6 +1915,88 @@ _http_ef_http_server(u3_httd* htd_u,
...
+        wol = u3dc("wash", u3nc(0, col_l), u3k(tan));
+        u3z(blu);
+      }
+
+      if ( u3_none != wol ) {
+        u3_noun low = wol;
+        u3_atom txt = u3_nul;
+        while ( u3_nul != low ) {
+          u3_atom lin = u3qc_cat(3, u3qc_rap(3, u3k(u3h(low))), c3_s2('\n', '\n'));
+          txt = u3qc_cat(3, txt, u3qc_cat(3, u3i_string("data:"), lin));
```
joemfb, line 1957: These refcounts don't look quite right -- callers to `q` jet functions retain their references. I also suspect that this loop could be refactored to simply cons/weld a list of atoms and only concatenate (rap) at the end. I'd have to work through the details though.

Another possibility would be send each line separately via `_http_continue_respond()`.

pkg/urbit/vere/io/http.c
```diff
@@ -1857,11 +2122,34 @@ _http_io_exit(u3_auto* car_u)
   //   _http_serv_close(htp_u);
   // }
 
-  // XX close u3_Host.fig_u.cli_u and con_u
+  {
+    u3_noun dat = u3nt(u3_nul, 25, u3i_string("data:urbit shutting down\n\n"));
```
joemfb, line 2126: I think that's a little long to hardcode the count. I think it'd be better to measure explicitly here.

pkg/urbit/vere/io/http.c
```diff
@@ -1857,11 +2122,34 @@ _http_io_exit(u3_auto* car_u)
...
+{
+  u3_httd* htd_u = (u3_httd*)car_u;
+  c3_y sec_y = 0;
+  u3_hreq* seq_u = htd_u->fig_u.seq_u;
+  while ( 0 != seq_u ) {
+    sec_y++;
+    seq_u = seq_u->nex_u;
+  }
+  u3l_log("      open slogstreams: %d\n", sec_y);
+}
```
joemfb, line 2151: This is great, we should add more metrics at some point.

pkg/urbit/vere/pier.c
```diff
@@ -971,10 +971,22 @@ _pier_on_lord_slog(void* ptr_v, c3_w pri_w, u3_noun tan)
...
   }
   else {
-    u3_pier_tank(0, pri_w, tan);
+    u3_pier_tank(0, pri_w, u3k(tan));
+    if ( 0 != pir_u->sog_f ) {
+      pir_u->sog_f(pir_u->sop_p, pri_w, tan);
+    }
+    else {
+      u3z(tan);
+    }
```
joemfb, line 989: I'd refactor this slightly so that `u3z(tan)` is only called once, at the end of the function.

joemfb (changes requested):
This looks good, I've picked a few nits below.

One higher-level note: I think we need to explicitly decide to reserve a part of the url namespace for the runtime (`/~/` is fine, if maybe prime real-estate). We just want to make sure not to step on our own toes going forward.

pkg/urbit/vere/io/http.c
```diff
@@ -728,6 +801,123 @@ typedef struct _h2o_uv_sock {         //  see private st_h2o_uv_socket_t
...
+{
+  //  try to find a cookie header
+  //
+  u3_weak coo = u3_none;
+  {
+    u3_noun hed = _http_heds_to_noun(rec_u->headers.entries,
+                                     rec_u->headers.size);
+    u3_noun deh = hed;
+    while ( u3_nul != deh ) {
+      if ( c3y == u3r_sing_c("cookie", u3h(u3h(deh))) ) {
```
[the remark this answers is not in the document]
Fang-: Yes! Was surprised it wasn't something we had to do manually, but this gives `cookie` even when the browser sends `Cookie`.

Fang-:
> I think we need to explicitly decide to reserve a part of the url namespace for the runtime (`/~/` is fine, if maybe prime real-estate)

Eyre already uses `/~` for some of its own reserved urls:

https://github.com/urbit/urbit/blob/e262403436cbae5ce37749416ab9d8c0c05270d9/pkg/arvo/sys/vane/eyre.hoon#L1959-L1962

Maybe we want to go with `/~~` prefix for "runtime can use these for whatever" urls.

joemfb:
How about `/~_~/`?

joemfb:
Whatever the prefix is, I think we'll want to modify %eyre to prevent binding it.

Fang-:
@joemfb I think I addressed most of your comments. #3647 will prevent conflicting bindings by userspace. Please take another look.

pkg/urbit/vere/io/http.c
```diff
@@ -728,6 +801,126 @@ typedef struct _h2o_uv_sock {         //  see private st_h2o_uv_socket_t
...
+        coo = u3k(u3t(u3h(deh)));
+        //TODO  http2 allows the client to put multiple 'cookie' headers
+        break;
+      }
+      deh = u3t(deh);
+    }
+    u3z(hed);
+  }
```
joemfb, line 897: Should've mentioned this earlier -- this search could use `h2o_find_header()` (and maybe `h2o_lookup_token()`) instead of allocating all the headers first. That could be a nice performance improvement, but this is also unlikely to be any kind of hot-loop. Maybe leave a comment?

joemfb (changes requested):
A couple comment requests, otherwise LGTM.

pkg/urbit/vere/io/http.c
Fang-: Good catch! I'm not very familiar with the various offerings of h2o, but it should've occurred to me to just look for these. It's not a hot path, but still.

Should be all good now!

joemfb (approved):
LGTM

PR #3558, opened 2020-09-24
link-store: fix issue with bad paths

tacryt-socryp:


matildepark:
Is this in release/link? Does it need to still be open?

PR #3560, opened 2020-09-24
webdojo: slogs

Fang-:
Connects to the SSE stream on `/~/slog` to receive printf output from the
runtime, and pretends those are incoming `%txt` effects.

Depends on #3557, which introduces the endpoint used here. Probably shouldn't be merged until that has gone out in a vere release? But this fails gracefully, so maybe we can pass on juggling pending PRs.

### DIFF
--- a/pkg/interface/src/views/apps/dojo/subscription.js
+++ b/pkg/interface/src/views/apps/dojo/subscription.js
@@ -19,9 +19,11 @@ export default class Subscription {
 
   setupSlog() {
     const slog = new EventSource('/~/slog', { withCredentials: true });
+    let available = false;
 
     slog.onopen = e => {
       console.log('slog: opened stream');
+      available = true;
     }
 
     slog.onmessage = e => {
@@ -30,11 +32,13 @@ export default class Subscription {
 
     slog.onerror = e => {
       console.error('slog: eventsource error:', e);
-      window.setTimeout(() => {
-        if (slog.readyState !== EventSource.CLOSED) return;
-        console.log('slog: reconnecting...');
-        this.setupSlog();
-      }, 10000);
+      if (available) {
+        window.setTimeout(() => {
+          if (slog.readyState !== EventSource.CLOSED) return;
+          console.log('slog: reconnecting...');
+          this.setupSlog();
+        }, 10000);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/apps/dojo/subscription.js
```diff
@@ -14,6 +14,36 @@ export default class Subscription {
...
+
+    slog.onmessage = e => {
+      this.handleEvent({ txt: e.data });
+    }
+
+    slog.onerror = e => {
+      console.error('slog: eventsource error:', e);
+      if (!available) {
+        this.handleEvent({ txt:
+          'landscape: no printf stream. bad connection or old binary.'
```
matildepark, line 37: Makes webdojo itself messier than necessary honestly. This should be a `console.log`.

matildepark:
I feel weird approving stuff that is awaiting next-sys changes. The "PR is blocked on a next-sys release" case isn't well covered IMO.

Fang-:
> But this fails gracefully, so maybe we can pass on juggling pending PRs.

With the requested change, this even fails _silently_. Unless the user is looking at browser console.

In this particular case I don't think it's a big deal to merge this ahead of the runtime-side dependency. But it's your call!

matildepark:
slog reconnects every few seconds ... 

image

Fang-:
Oops, looks like I threw a baby out with the bathwater in b12f892. Will repair!

philipcmonk:
> The "PR is blocked on a next-sys release" case isn't well covered IMO.

Without following the full context, generally that just means it should be PR'd against next-sys.  Of course it'll look a little different once landscape and arvo are fully decoupled.

matildepark:
I generally agree. Once the two are decoupled, though, we are essentially pinned to an Arvo release (maybe a Git submodule? Idk) so we’d have to think about changes with dependencies. Perhaps just strongly versioning.

philipcmonk:
Yeah I think you'll have a branch that depends on future arvo changes which is separate from the one which doesn't.  That branch can even have changes which are *required* to make landscape work with the new version -- we can still unidirectionally couple by having anyone distributing landscape (eg ~zod) only send the new arvo changes with the corresponding landscape changes.  Other software distributors should do the same thing.

Fang-:
@matildepark this fails more gracefully now. Please take another look. Afaic this can just slide into the next-userspace queue, but your call on when/where to merge this.

matildepark:
Whoops. Will look tn.

matildepark (approved):
Fails gracefully.

Incidentally this makes Dojo overflow scroll horizontally, which I guess was on purpose.

Before / after:

image


image

PR #3561, opened 2020-09-24
kh: abolish unicode syntax in favor of ascii

Fang-:
I haven't done much (king) haskell work, but I did spot this crime against developers. Everyone I've spoken to about it was in favor of returning to ascii syntax, so I did a bunch of find-replace, and this is the result.


tylershuster:
https://github.com/tonsky/FiraCode people

pilfer-pandex (approved):
Thank you.

PR #3563, opened 2020-09-25
Restore thread and Indigo import fix

liam-fitzgerald:
Adds a thread to restore an archived graph to full functioning, and fixes imports that broke in the indigo 1.2 upgrade


tacryt-socryp (approved):
LGTM

PR #3564, opened 2020-09-25
eyre: send 'no content' headers with no content

tylershuster:
Ok, full disclosure I didn't test this. But I know so little about this system I figured I should just run it all by you and tell you why.

I'm tired of seeing this in the console and it suggests an error:
```
XML Parsing Error: no root element found
Location: http://localhost:1701/~/channel/1601004556-791863
Line Number 1, Column 1:
```

So, in `eyre.hoon` wherever I saw `data=~` and `status-code=200` I changed it to 204, as per [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)

?


matildepark:
(I feel like this should be `release/next-sys`.)

tylershuster:
Yeah true maybe. Got to clean up some conflicts thayn.

tylershuster:
FFS will redo

PR #3565, opened 2020-09-25
eyre: send 'no content' headers with no content

tylershuster:
Take 2!

Ok, full disclosure I didn't test this. But I know so little about this system I figured I should just run it all by you and tell you why.

I'm tired of seeing this in the console and it suggests an error:
```
XML Parsing Error: no root element found
Location: http://localhost:1701/~/channel/1601004556-791863
Line Number 1, Column 1:
```

So, in `eyre.hoon` wherever I saw `data=~` and `status-code=200` I changed it to 204, as per [the spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204)

?

### DIFF
--- a/pkg/arvo/tests/sys/vane/eyre.hoon
+++ b/pkg/arvo/tests/sys/vane/eyre.hoon
@@ -996,7 +996,7 @@
                 %give
                 %response
                 %start
-                :-  204
+                :-  200
                 :~  ['content-type' 'text/event-stream']
                     ['cache-control' 'no-cache']
                     ['connection' 'keep-alive']
--- a/pkg/arvo/tests/sys/vane/eyre.hoon
+++ b/pkg/arvo/tests/sys/vane/eyre.hoon
@@ -996,7 +996,7 @@
                 %give
                 %response
                 %start
-                :-  200
+                :-  204
                 :~  ['content-type' 'text/event-stream']
                     ['cache-control' 'no-cache']
                     ['connection' 'keep-alive']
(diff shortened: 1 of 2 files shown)


Fang-:
You'll want to update the eyre tests to reflect the new expected results (ie just change `200` to `204` in a couple places). This seems fine to me otherwise. Do give this a test run and see if it _actually_ resolves the issue though. (;

tylershuster:
Well, I think I updated the test code but "To run all tests in /tests, run +test in dojo. +test /some/path would only run all tests in /tests/some/path." does not work. And there's no `test.hoon` in gen so...is there a new process for testing?

Fang-:
I think it's `-test %/tests/your/test` now.

tylershuster:
Thank you sir. Somehow borked this commit but the test does succeed.

tylershuster:
Any chance of this going out in next-sys?

pkg/arvo/tests/sys/vane/eyre.hoon
```diff
@@ -996,7 +996,7 @@
                 %give
                 %response
                 %start
-                :-  200
+                :-  204
```
Fang-, line 999: This one shouldn't expect a `204` response, because there is data in there! Eyre is failing to meet this expectation, so I guess that's good.

Fang- (changes requested):
There's an incorrect expectation in the tests, but otherwise looks good.

pkg/arvo/tests/sys/vane/eyre.hoon
tylershuster: Weird. That shouldn't have slipped in there.

Fang- (approved):
Thank you! (:

PR #3566, opened 2020-09-25
metadata-store: rebuild indices

liam-fitzgerald:
The OTA deployment of urbit-os-v1.0.50 unintentionally wiped the indices
during the state migration. This caused widespread permissions issues,
as agents were unable to tell which group is associated with what channel.
Fortunately, the indices are a simple transform of the associations noun,
which was kept intact. This commit introduces a state migration to rebuild
the indices, in order to restore connectivity on the network.


Fang- (approved):
Looks good!

PR #3569, opened 2020-09-25
dojo: parse %as in ++parse-value instead of ++parse-build

### LINKED ISSUE #1476: "+foo &bar" in dojo syntax-errors after &
Should, uh, not do that, running generators on mark-conversion results is totally valid. Minor parser bug.

I think the motivating example was `|init-auth-google &json .^(json %cx %/keys/json)`?

pkova:
Allows things like `+hello &helm-hi 'hi'`. Fixes #1476.


pkova:
Ping @Fang-

Fang- (approved):
Mark-converted data is indeed a value. Good fix!

PR #3577, opened 2020-09-25
vere: adds jets for en/decoding base16

joemfb:
Jets for the new implementations in #3576.

### DIFF
--- a/pkg/urbit/jets/e/base.c
+++ b/pkg/urbit/jets/e/base.c
@@ -9,21 +9,18 @@ const c3_y hex_y[16] = { '0', '1', '2', '3', '4', '5', '6', '7',
 u3_noun
 u3qe_en_base16(u3_atom len, u3_atom dat)
 {
-  c3_w len_w, byt_w;
-
   if ( c3n == u3a_is_cat(len) ) {
     return u3m_bail(c3__fail);
   }
   else {
-    len_w = (c3_w)len;
-    byt_w = len_w * 2;
+    c3_w     len_w = (c3_w)len;
+    u3i_slab sab_u;
 
-    if ( (byt_w / 2) != len_w ) {
-      return u3m_bail(c3__fail);
-    }
-    else {
-      c3_y* buf_y = u3a_malloc(byt_w);
-      c3_y* dat_y = buf_y;
+    u3i_slab_bare(&sab_u, 4, len_w);
+    sab_u.buf_w[sab_u.len_w - 1] = 0;
+
+    {
+      c3_y* buf_y = sab_u.buf_y;
       c3_y  inp_y;
 
       while ( len_w-- ) {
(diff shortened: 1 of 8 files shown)


joemfb:
Once #3612 is merged, I'll update this PR to use the slab abi and only allocate once per call.

pkg/urbit/noun/retrieve.c
```diff
@@ -1107,6 +1107,39 @@ u3r_mp(mpz_t   a_mp,
...
+u3r_short(c3_w  a_w,
+          u3_atom b)
+{
+  c3_assert( u3_none != b );
+  c3_assert( c3y == u3a_is_atom(b) );
+
+  if ( c3y == u3a_is_cat(b) ) {
+    switch ( a_w ) {
+      case 0:  return b & 0xffff;
+      case 1:  return b >> 16;
```
Fang-, line 1124: I'd expect `0` to be no data at all, and `1` to be two bytes (which is what a short is, right?)... so I'm probably misunderstanding what this is supposed to be doing.

Fang- (commented):
Looks fine, but there's parts I don't fully get. Probably missing something trivial/unexpected.

pkg/urbit/noun/retrieve.c
joemfb: `a_w` is a zero-based index into the atom, considered as an array of shorts (ie, uint16). In the case of a direct atom, `u3r_short(0, a)` is the 16 least-significant bits, and `u3r_short(1, a)` is the 16 most significant bits.

This is the same interface as `u3r_bytes()` and `u3r_words()`, modulo the difference in size.

pkg/urbit/jets/e/base.c
```diff
@@ -0,0 +1,160 @@
...
+
+  {
+    c3_w byt_w = u3r_met(3, inp);
+
+    if ( byt_w & 1 ) {
+      ret_o = _of_hex_odd(inp, len_w, byt_w, buf_y);
+    }
+    else {
+      ret_o = _of_hex_even(inp, len_w, buf_y);
+    }
```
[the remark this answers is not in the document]
joemfb: The hex-decoded length in bytes is equal to the hex-encoded length in bytes, divided by two, rounding up. Ie, `u3r_met(3, output) == u3r_met(4, input)` -- two bytes of hex-encoded input correspond 1 byte of output. If there are an even number of bytes, we can simply process them two at a time (reading them as shorts), but if there are odd-number of bytes of input, we must specially handle the most significant, and our pairs of input bytes span input shorts.

I'll add a comment.

pkg/urbit/noun/retrieve.c
Fang-: > his is the same interface as `u3r_bytes()` and `u3r_words()`

`_byte` and `_word`, right?

This makes sense now, in any case. Thanks!

joemfb: Right, that's what I meant.

joemfb: I thought about adding `u3r_shorts()` as well, but I don't have an immediate use-case, and it raises (new) endian-ness questions.

Fang- (approved):
Elaboration and changes look good! (:

PR #3579, opened 2020-09-25
vere: adds/uses apis for repeated deserialization

joemfb:
This PR adds handle-based apis for eliding allocations in contexts where we'll be repeating deserializing inputs. The previous approach was an unsafe API that let the caller supply the context (but required that the caller maintain invariants that could not be enforced). The new approach simply maintains the invariants. The handles are now used on both ends of the ipc pipe, in the persistent layer, and in ames' stateless-forwarding.


Fang- (approved):
Thanks for the nice commit separation as always. As far as I can tell, this all seems good!

PR #3588, opened 2020-09-25
landscape: channel.js plumbing acks through for subscriptions

baudtack:


pkg/arvo/app/landscape/js/channel.js
```diff
@@ -210,6 +212,9 @@ class Channel {
         if (obj.hasOwnProperty("err")) {
```
Fang-, line 212: There's a comment just above here (that gh won't let me select) that says "only notify on err". So that can be removed.

pkg/arvo/app/landscape/js/channel.js
```diff
@@ -210,6 +212,9 @@ class Channel {
...
           funcs["err"](obj.err);
           this.outstandingSubscriptions.delete(obj.id);
+        } else if (obj.hasOwnProperty("ok")) {
+          funcs["subAck"](obj);
+          this.outstandingSubscriptions.delete(obj.id);
```
Fang-, line 217: I don't think we want to delete the subscription when we get an ack for it, right? We still want to have its `event` callback around, and maybe it can error out later.

Fang- (changes requested):

Fang-:
(Also, you're free to just append commits to your PR instead of opening a new one. If you must insist on squashing history, there's always `git commit --amend` and `git push --force-with-lease`.)

pkg/arvo/app/landscape/js/channel.js
baudtack: fixed.

Fang- (approved):
Changes look good. Thanks!

PR #3590, opened 2020-09-25
interface: open dev server to network

tylershuster:
This opens up the dev server to network connections, for instance on a mobile device on the same wifi network.


matildepark:
`meta` is for the repo / admin; `interface` is for "landscape-wide" changes

matildepark (approved):
Preserves the localhost:9000 case and makes mobile testing simpler.

PR #3591, opened 2020-09-26
shoe: don't auto-run discontinuous inputs

Fang-:
Previously, up-arrowing into (or otherwise retrieving) a command from history
that automatically runs on-input would directly run the command again,
preventing the user from up-arrowing past the auto-run command into further
history.

With this change, shoe detects discontinuous inputs (sole's `%set` edit), and
refuses to auto-run the parsed command in those cases.

Wrote an alternative fix here: 4267761, but ultimately decided against leaving auto-run commands out of history entirely. I imagine the behavior here fits user expectations better. Maybe this needs to be configurable, but I'm be hesitant to make that change without clear demand for it.

### DIFF
--- a/pkg/arvo/lib/shoe.hoon
+++ b/pkg/arvo/lib/shoe.hoon
@@ -243,7 +243,10 @@
         %+  rose  (tufa buf.cli-state)
         (command-parser:og sole-id)
       ?:  ?=(%& -.res)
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
Looks good.  I do think it's better to still add to history and just not auto-run.

PR #3595, opened 2020-09-26
Implement stateless forwarding in the king

pilfer-pandex:
Our version of #3174 and #3405.

When the king receives a packet, he parses it and determines if it is for a ship other than himself. If so, he forwards it without involving the serf, except that he from time to time scries the serf for information on the routes ("lanes"/"AmesDests") stored for ships he wishes to forward to. Additionally, he drops packets that are malformed, have the wrong version number, or which are bound for ships that we don't have a route to.

Test plan.

Change the logDebug to a logInfo on L258 and add the following code at L259 of Ames.hs:
```
if encode pkt == b
  then logInfo $ displayShow ("ames: packet match")
  else logInfo $ displayShow ("ames: packet mismatch", encode pkt, b)
```
Throughout, verify that we have only "packet match" messages.

1. Run your own ship with this version of king haskell. Make sure everything looks good and we have no packet parsing errors.

2. Run the following transcript
```
$ ./urbit-king new -F zod
$ ./urbit-king new -F marzod
$ ./urbit-king new -F panzod
zod> |hi ~marzod
marzod> |hi ~zod
panzod> |hi ~zod
panzod> |hi ~marzod
```
Look to see that all the printouts look good. In particular, for the hi from panzod to marzod, we should see a packet arrive at zod with sndr panzod, rcvr marzod, and origin Nothing, and we should subsequently see a packet arrive at marzod with sndr panzod, rcvr marzod, and origin Just <stuff>.

3. [testing the cache system]. Modify LaneCache.hs to have an expiry of only 3 seconds. Add putStrLns to each of the three possible outcomes of `byCache`. To the start function in Ames.hs add code which spawns a thread that, every .5s scries for the route to marzod (Ship 256). Start the zod from part 2 of this test plan and watch the printouts. You should see first that there is a cache miss, then that there are a bunch of cache hits, then a cache expiry.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
@@ -135,7 +135,7 @@ ames'
   :: HasPierEnv e
   => Ship
   -> Bool
-  -> (Time.Wen -> Gang -> Path -> (Maybe (Term, Noun) -> IO ()) -> STM ())
+  -> (Time.Wen -> Gang -> Path -> IO (Maybe (Term, Noun)))
   -> (Text -> RIO e ())
   -> RIO e ([Ev], RAcquire e (DriverApi NewtEf))
 ames' who isFake scry stderr = do
@@ -189,7 +189,7 @@ ames
   => e
   -> Ship
   -> Bool
-  -> (Time.Wen -> Gang -> Path -> (Maybe (Term, Noun) -> IO ()) -> STM ())
+  -> (Time.Wen -> Gang -> Path -> IO (Maybe (Term, Noun)))
   -> (EvErr -> STM PacketOutcome)
   -> (Text -> RIO e ())
   -> ([Ev], RAcquire e (NewtEf -> IO ()))
@@ -209,7 +209,7 @@ ames env who isFake scry enqueueEv stderr = (initialEvents, runAmes)
   start :: RIO e AmesDrv
   start = do
     mode <- rio (netMode isFake)
-    cache <- laneCache scryLane
+    cachedScryLane <- cache scryLane
 
     aTurfs   <- newTVarIO Nothing
     aDropped <- newTVarIO 0
@@ -220,7 +220,7 @@ ames env who isFake scry enqueueEv stderr = (initialEvents, runAmes)
     aRecvTid <- queuePacketsThread
       aDropped
       aVersion
-      (byCache cache)
+      cachedScryLane
       (send aUdpServ aResolvr mode)
       aUdpServ
 
@@ -230,21 +230,22 @@ ames env who isFake scry enqueueEv stderr = (initialEvents, runAmes)
 
   trackVersionThread :: HasLogFunc e => TVar (Maybe Version) -> RIO e (Async ())
   trackVersionThread versSlot = async $ forever do
-    env <- ask
+    scryVersion >>= \case
+      Just v -> do
+        v0 <- readTVarIO versSlot
+        atomically $ writeTVar versSlot (Just v)
+        if (v0 == Just v)
+          then logInfo $ displayShow ("ames: proto version unchanged at", v)
+          else stderr ("ames: protocol version now " <> tshow v)
 
-    scryVersion \v -> do
-      v0 <- readTVarIO versSlot
-      atomically $ writeTVar versSlot (Just v)
-      if (v0 == Just v)
-        then logInfo $ displayShow ("ames: proto version unchanged at", v)
-        else stderr ("ames: protocol version now " <> tshow v)
+      Nothing -> logError "ames: could not scry for version"
 
     threadDelay (10 * 60 * 1_000_000)  -- 10m
 
   queuePacketsThread :: HasLogFunc e
                      => TVar Word
                      -> TVar (Maybe Version)
-                     -> (Ship -> (Maybe [AmesDest] -> RIO e ()) -> RIO e ())
+                     -> (Ship -> RIO e (Maybe [AmesDest]))
                      -> (AmesDest -> ByteString -> RIO e ())
                      -> UdpServ
                      -> RIO e (Async ())
@@ -259,7 +260,7 @@ ames env who isFake scry enqueueEv stderr = (initialEvents, runAmes)
 
         if pktRcvr == who
           then serfsUp p a b
-          else lan pktRcvr $ \case
+          else lan pktRcvr >>= \case
             Just (dest:_) -> forward dest $ encode pkt
               { pktOrigin = pktOrigin <|> Just (ipDest p a) }
             _ -> logInfo $ displayShow ("ames: dropping unroutable", pkt)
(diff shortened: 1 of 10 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Ames/Packet.hs
```diff
@@ -0,0 +1,102 @@
...
+    let sndrRank     = shiftR  head 23 .&. 0b11
+    let rcvrRank     = shiftR  head 25 .&. 0b11
+    let pktEncrypted = testBit head 27  &  not  -- loobean
+    -- verify checksum
+    lookAhead $ do
+      len  <- remaining
+      body <- getBytes len
+      -- XX mug (marked "TODO") is implemented as "slowMug" in U.N.Tree. Ominous
+      -- Also, toNoun will copy the bytes into an atom. We probably want a mugBS
+      let chk = fromIntegral (mug $ toNoun $ MkBytes body) .&. (2 ^ 20 - 1)
```
Fang-, line 57: So, is this super slow right now, or what? Once you're reasonably confident in this as a whole, you should try putting it on ~nus and see if it can keep up.

pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
```diff
@@ -194,36 +206,85 @@ ames env who isFake enqueueEv stderr = (initialEvents, runAmes)
...
+            Just (dest:_) -> forward dest $ encode pkt
+              { pktOrigin = pktOrigin <|> Just (ipDest p a) }
+            _ -> logInfo $ displayShow ("ames: dropping unroutable", pkt)
+
+      Right pkt -> logInfo $ displayShow ("ames: dropping ill-versed", pkt, ver)
+
+      Left e -> logInfo $ displayShow ("ames: dropping malformed", e)
+
+    where
+      serf'sUp p a b =
```
Fang-, line 272: \>[serf's up](https://i.imgur.com/1rkt3Qo.png)

I know haskell syntax allows this, but is it considered fair game, style-wise, to put `'` in there? Wouldn't this be more legible as `enqueuePacket`, matching the `enqueueEv` it calls out to?

Fang- (commented):
This feels pretty small compared to the C code. Nice!

I don't see anything surprising or weird here. I'd recommend taking this for a quick spin on ~nus (instance `bigger-nus`, mind) to see how it performs under real load.

pkg/hs/urbit-king/lib/Urbit/Vere/Ames/Packet.hs
pilfer-pandex: On reflection, I think that mug is only appreciably slow for nonatomic nouns, but I could be wrong about this. I think that you propose a good test. How do we track whether it is keeping up?

pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
pilfer-pandex: It's not done commonly, but it's also not considered unstylish. Like if I saw a name with "dont" in it, I'd probably tell you to insert the apostrophe.

I get what you're saying with enqueue, but this name makes immediately clear where the thing is going, not merely that it is going into some queue. I also now reserve packet for `Packet`s and this thing takes a ByteString (you'll notice I also renamed `sendPacket` to `send` for this reason).

pkg/hs/urbit-king/lib/Urbit/Vere/Ames/Packet.hs
Fang-: ~nus was dropping packets without stateless forwarding. If it's not getting pressured into dropping packets when running stateless forwarding (I see kh has printfs for this too), then it's keeping up. (:

joemfb (changes requested):
I've gone over this closely several times now, the implementation looks correct, and is nice and clean.

I have three concerns with this functionality, none of them directly the responsibility of this new code:

- the scry handler throws on exception on a `%bail` response (ie, an error in the serf while evaluating the request)

`recvPeek` in `Serf/IPC.hs` should be updated to produce `Nothing` in the `%bail` case -- this matches the current handling in vere

- the `slowMug` implementation

The mug in this case should be able to take the bytestring we already have. But the allocation is not the only thing that's suspect -- the inner implementation details of slowMug also are a little strange (the resize operation in `muk`?). (And, frankly, this also applies to a variety of other basic atom operations in `Noun/Tree.hs`: most notably `met` being a right-shift loop instead of a GMP builtin.)

It might not need to block this PR, but optimizing mug and cleaning up the basic atom operations should be flagged as a priority.

- the protocol version handling

This is a complicated topic, and the details are not fully baked. I've laid out the concerns in #3764 (specifically, in the comments in 67f06ce5e5dbd7912ba8a6b9055c2dc28614a411). I don't expect this to be solved now or here, but it would be good to include some comments (in `queuePacketsThread`) noting that there are situations where we'll likely need to enqueue packets we don't know how to parse.

pilfer-pandex:
Thanks so much for so thoroughly lending this code your careful eye. I'll call your bullet points 1-3. In response:

1. Ah thanks for reminding me of that. Will do.
2. I think we should accept this as is for now and prioritize fixing mug (and also my fully understanding mug) in a new PR. Quite likely, we'll want the mug to be cached in noun representation, rather than the other hash that's there right now. I'll want to talk with ben to understand more about why that hash is there anyway. Changes to the noun representation are a big deal and don't belong here.
3. My understanding of your comment in `67f06ce` is as follows: When we have packets from old versions, we should still accept and even kingside-forward them (with the last perhaps holding only for the most recent previous version). When we have packets from future versions, we must enqueue them. In short, if a packet doesn't parse, it should be enqueued rather than dropped. Additionally (and this does not apply in the context of today's pr anyway), when the version is incremented, we should keep around the code that parses the previous version and still support forwarding for that. So the action item in this pr is to change the [code that drops invalid packets](https://github.com/urbit/urbit/blob/359d6c881b4ff0fbb5604be20c30e2cae2871789/pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs#L270) to instead serf them up.

@joemfb, does the foregoing resolution plan sound good to you?

joemfb:
@pilfer-pandex, yes on 1 and 2.

re: 3, I'm not asking you to change this code. Just to add a comment in/by `queuePacketsThread` that this system is supposed to smoothly handle protocol upgrades, but does not currently, and will need to be revisited. I put down my thoughts in more detail in #3764, but that's just like my opinion, man. The details still need to be worked through, it's just important to flag protcol-version handling as Not Done.

pkg/hs/urbit-king/lib/Urbit/Vere/Ames.hs
```diff
@@ -194,36 +206,86 @@ ames env who isFake enqueueEv stderr = (initialEvents, runAmes)
...
+    ver <- readTVarIO vers
+
+    case decode b of
+      Right (pkt@Packet {..}) | ver == Nothing || ver == Just pktVersion -> do
+        logDebug $ displayShow ("ames: bon packet", pkt, showUD $ bytesAtom b)
+
+        if pktRcvr == who
+          then serfsUp p a b
+          else lan pktRcvr >>= \case
+            Just (dest:_) -> forward dest $ encode pkt
```
joemfb, line 264: To prevent forwarding loops (#3778), the forwarding logic needs to check if the destination lane is a galaxy (and if we are a galaxy) and exclude our selves. See e59863b in #3764 for an example.

joemfb (changes requested):
One more requirements issue came up today:

pilfer-pandex:
Addressed joe's review comments and the request to skip lanes that would forward to a galaxy that is ourself. Reran my test plan described above.

joemfb (approved):
I'm sorry for the delay. While there's more work to be done in this area (tracking stats, performance improvements to `+mug`, &c), this PR is complete as is, and all feedback has been addressed. (It'd be good to get another test on a live galaxy.)

LGTM

PR #3599, opened 2020-09-28
landscape: add woff2 support, serve Inter & Source Code Pro

tylershuster:
Fixes https://github.com/urbit/landscape/issues/92 except it doesn't. What am I missing?

(does not reference inter.woff2 in CSS yet, just trying to get it to serve woff2 files)


pkova:
My guess would be `pkg/interface/src/views/css/fonts.css` and `pkg/interface/dbug/src/css/fonts.css`.

matildepark:
Well the first thing I’d do is go directly to ~landscape/fonts/inter.woff2 and see if it’s there.

tylershuster:
Right yeah I went to the URL and got a 404. Haven't added it to css yet

pkova:
This is broken because the url extension parsing in zuse is broken. I made a PR #3603.

tylershuster:
Cheers that puts me back on track

matildepark:
You may want to target next-sys since this depends on the PR in there. (And maybe serve Source Code Pro while you're at it.)

tylershuster:
Good point on the target. I've based off it. I'm adding SCP but I've run into some issues I'm working through with fang

tylershuster:
@Fang- Here's what I'm working with. Hopefully you can reproduce my problem (For spectators, the problem is that trailing zeroes get trimmed off the end when the woff2 is served)

tylershuster:
Should note for posterity that we have an issue where uppercase files don't get served

Fang-:
> the problem is that trailing zeroes get trimmed off the end when the woff2 is served

This probably happens at the point of storage in clay, because we define its canonical type to be an atom. If we care about leading zeroes (trailing zeroes become leading because muh byte ordering) then the type should be octs!

Defining the mark as such, and updating file-server to scry for `octs` instead of `@`, seems to work:

```hoon
|_  dat=octs
++  grow 
  |% 
  ++  mime  [/font/woff2 dat] 
  -- 
++  grab 
  |% 
  ++  mime  |=([=mite =octs] octs) 
  ++  noun  octs
  -- 
++  grad  %mime 
--
```

We should probably consider making file-server filetype agnostic, and just always scrying for `(snoc path.content.u.content %mime)`, instead of manually reconstructing the octs and content-type... but that's outside of scope here.

tylershuster:
So what needs to be changed besides this mark?

Fang-:
> updating file-server to scry for `octs` instead of `@`

So, something slightly tidier than:

```hoon
        ::  etc
        ?.  .^(? %cu scry-path)  [not-found:gen %.n]
        ?:  ?=([~ %woff2] ext.req-line)
          :_  public.u.content
          [[200 [['content-type' '/font/woff2'] ~]] `.^(octs %cx scry-path)]
        =/  file  (as-octs:mimes:html .^(@ %cx scry-path))
        ::  etc
```

tylershuster:
You'll make a hooner of me yet

tylershuster:
If I update `file-server` to the following:
```
        ?.  .^(? %cu scry-path)  [not-found:gen %.n]
        =/  file  (as-octs:mimes:html .^(@ %cx scry-path))
        :_  public.u.content
        ?+  ext.req-line  not-found:gen
            [~ %js]     (js-response:gen file)
            [~ %css]    (css-response:gen file)
            [~ %png]    (png-response:gen file)
            [~ %woff2]  [[200 [['content-type' 'font/woff2'] ~]] `.^(octs %cx scry-path)]
```

I get

```
%nets-failed
%nets-need
[p=@ud q=@]
%nets-have
@
arvo: scry-lost
```

matildepark:
Since the Zuse change is in master this can target userspace again

PR #3602, opened 2020-09-28
readme: update arvo test command

Fang-:
This wasn't updated to match "recent" changes to how testing works.


jtobin (commented):
Trivial, but may I recommend the slight tweak:

```
To run all tests in `/tests`, run `-test %/tests` in dojo. To run only the tests in `/tests/some/path`, use `-test %/tests/some/path`.
```

I just find that reads a little more naturally.

jtobin (approved):

PR #3603, opened 2020-09-28
zuse: improve url extension parsing

pkova:
Previously extensions ending with numbers such as `.woff2` were not recognized.

Observe:
```
(rash '/~landscape/fonts/inter.woff2' apat:de-purl:html)
[p=~ q=<|~landscape fonts inter.woff2|>]
```

Now: 
```
(rash '/~landscape/fonts/inter.woff2' apat:de-purl:html)
[p=[~ ~.woff2] q=<|~landscape fonts inter|>]
```

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -6563,8 +6563,13 @@
       ?^  t.rax
         [p.pok [ire q.pok]]:[pok=$(rax t.rax) ire=i.rax]
       =/  raf/(like term)
-          =>  |=(a/@ ((sand %tas) (crip (flop (trip a)))))
-          (;~(sfix (sear . sym) dot) [1^1 (flop (trip i.rax))])
+        %-  ;~  sfix
+              %+  sear
+                |=(a/@ ((sand %ta) (crip (flop (trip a)))))
+              (cook |=(a/tape (rap 3 ^-((list @) a))) (star aln))
+              dot
+            ==
+        [1^1 (flop (trip i.rax))]
       ?~  q.raf
         [~ [i.rax ~]]
       =+  `{ext/term {@ @} fyl/tape}`u.q.raf
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7a5ffa86349b0f418d7d5a5f69a773d29b1d48d80d657058ff5abb0572f5187f
-size 6409970
+oid sha256:b4e0acf6459d5d17f2dff4de71e9ddd0732148f00bd13d0c092fdfa14774a015
+size 6332407


pkg/arvo/sys/zuse.hoon
```diff
@@ -6512,8 +6512,13 @@
       ?^  t.rax
         [p.pok [ire q.pok]]:[pok=$(rax t.rax) ire=i.rax]
       =/  raf/(like term)
-          =>  |=(a/@ ((sand %tas) (crip (flop (trip a)))))
-          (;~(sfix (sear . sym) dot) [1^1 (flop (trip i.rax))])
+        %-  ;~  sfix
+              %+  sear
+                |=(a/@ ((sand %tas) (crip (flop (trip a)))))
```
pkova, line 6517: Does anyone know why the extension has to be a valid @tas here? This will still break on extensions like `.7z`.

tylershuster: docs says "a term or @tas, an atomic ASCII string which obeys symbol rules: lowercase and digit only, infix hyphen, first character must be a lowercase letter." — not that the docs are up to date, but maybe this is a clue

pkova:
Don't know who to summon so I'll summon @Fang-.

pkg/arvo/sys/zuse.hoon
matildepark: I don't think something like a term would change all that frequently really.

pkova: The extension in `++pork` is defined as `(unit @ta)`, though, so I'm wondering why there's a coercion here to a more restrictive type.

joemfb: In general, the intention is to map file extensions to marks inside urbit. This generally makes us compatible with only a subset of the outside world.

A similar example is the use of atoms to encode arbitrary binary data. If we require precise compatibility with outside systems is required, as in the case of hash functions and other cryptographic use-cases, the possibility of trailing null bytes must be handled with a structure such as `$octs`. Similarly we have have base16 encoding a la `@ux`, and base16 encoding a la #3576.

pkg/arvo/sys/zuse.hoon
```diff
@@ -6512,8 +6512,13 @@
...
+              (cook |=(a/tape (rap 3 ^-((list @) a))) (star aln))
```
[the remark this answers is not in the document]
Fang-: Also, to make sure I'm clear on this: `sym` -> `(star aln)` is your only real change, right? And that affects the `.woff2` case because it's parsed as `2ffow.`, correct? And then `.7z` example from above still won't pass, because it's sanity-checked as `7z`, which isn't a valid `@tas`.

pkg/arvo/sys/zuse.hoon
Fang-: Hmm, not sure what to make of this. "File extensions as marks" is nice, but would prevent us from, well, implementing a world-compatible http client. Perhaps we should only start caring about this wherever url -> clay path conversions or similar things happen?

pkg/arvo/sys/zuse.hoon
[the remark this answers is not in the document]
pkova: It's parsed as `2ffow` without the `dot`, but yeah you get the gist.

I was trying to find the correct parser here, `urs:ab` and similar don't work in this context since they include `dot` and we're inside `;~  sfix  ...  dot` here. Not sure if `aln` is the correct one tho.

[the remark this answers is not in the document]
Fang-: Ohh, right. That's a bit awkward, yeah. I'm struggling to find an RFC that says anything about file extensions. Seems like they might just be treated as part of the path, following the same parsing rules? Though, I don't think I've ever seen an extension fancier than `aln` in the wild...

tylershuster:
What's going on with this?

Fang-:
I got prodded about this last week and then didn't actually follow up. Sorry!

@pkova go ahead and make this parse for `@ta`, as defined in the `+pork` type, replacing the `%tas` with a `%ta` check instead. Maybe this means you can get rid of the `flop`ping here, too.

"Marks are file extensions" is cute, but clay paths are `(list @ta)`, which means they're already incompatible with `mark` defined as `@tas`. I imagine we want marks to be `@ta`, but that's obviously outside of scope here.

pkova:
Fixed.

Didn't spend too much time investigating the uses of `flop` here, but just naively removing them does not work.

Fang- (approved):
Looks good! Thanks for this @pkova. (:

PR #3416, opened 2020-09-01
Applied schizoanalysis: Sanity checking for landscape

liam-fitzgerald:
%sane: sanity checker for the landscape suite of applications

Userspace currently uses certain identifiers as foreign keys, and
expects those foreign keys to exist in a number of locations. 
These foreign key relationships are prone to breaking during OTAs 
and there are enough of them that they rarely get tested for 
manually. %sane is a gall app that will check the validity of
these relationships.

Pokes:
  %off - turn off checking on a reload
  %lax - check on reload, printing errors
  %strict - check on reload, aborting commit if invalid
  %check-print  - run a check now, printing errors
  %check-crash - run a check now, crashing nondeterministically if invalid 
    
Currently validates:
  - Entries in {contact,metadata,group} stores are in sync with 
     their hooks
  - Each group has its associated metadata and contacts
 
 Future validation:
- Subscriptions are correctly setup for hooks
- Graph store integration

PR #3604, opened 2020-09-28
data migration

ixv:
Here is a draft PR for data migration. The goal of this work is to be able to recover data from your ship when you breach and easily import it into your fresh ship.

Unfortunately I didn't realize graph-store was already finished and had replaced links. So this PR needs graph-store migrations before it can go in. It will need some minor changes to lens/herb once I finish that, so I'll commit those changes on top.

If you're interested in testing this, theres a handful of threads for setting up several ships, filling them with data, exporting, breaching, importing, and testing the subscription flows between the landscape apps post-import. Be forewarned, these threads are finicky as hell, since I'm somewhat abusing the aqua/ph system by keeping fleets and azimuth state around after threads "finish", in order to avoid the cost of doing full end-to-end tests for incremental changes.

The basic flow of the tests is
- `-ph-migrate-init` to boot ~zod, ~bus and ~web
- `-ph-migrate-start` to start the apps on the ships, because doing it in the init thread doesn't work for some reason
- `-ph-migrate-make-data` to fill the ships with landscape data
- `+aqua-export ~zod` (or bus or web), to export data from all the landscape apps
- `-ph-migrate-breach ~zod` (or bus or web) breaches this ship. this is the most finicky step, you have to run `:azimuth-tracker|kick` on the non-breached ships to make sure they get the update, and often `-time ~s1` to get timers unstuck.
- `-ph-migrate-import ~zod exported-data` imports the data that you exported previously
- `-ph-migrate-test-import` sends some additional data through the channels set up in the `make-data` step, to ensure everything has resumed properly

If you do trouble yourself to actually try these tests, let me know if you run into any problems, I'll be happy to help you through it.


matildepark:
> Unfortunately I didn't realize graph-store was already finished and had replaced links. So this PR needs graph-store migrations before it can go in. It will need some minor changes to lens/herb once I finish that, so I'll commit those changes on top.

Note that Publish is on its way out shortly too (see #3473) and Chat in the next few weeks. Metadata-store's data structure changed a bit, dunno if that affects this.

belisarius222:
FYI there is now some prior art on graph migration, designed mostly for archived link collections: https://github.com/urbit/urbit/pull/3521

Fang- (commented):
I looked this over and don't have strong feelings about anything here. Some of the hook import stuff looks like it'd only really work if the store import happens too, but I guess that's what the `%import-all` is for.

I'm assuming the only approved use case of this is "I want/need to start with a fresh pier"? What's the thinking around supporting imports from older versions of the app (in case an OTA happens while my I'm sunk)?

Also, how does this deal with importing data from ~A into ~B? That is, the "I want to use a new identity" case?

ixv:
Thats right, this is for the use case of breaching a ship while maintaining the same landscape state. You wouldn't use it for individual apps.

If apps upgrade, we can add multiple cases to the +poke-import arms to handle the different versions. I think we can do that in the future, I don't think we need to do it to handle cases from prior to the release of this utility.

It isn't going to work for importing into a different ship, to do that we would need to let all the other ships with which you have subscriptions know about the upgrade. This could be something to look into, but not in scope for this work.

pkg/arvo/gen/aqua-export.hoon
```diff
@@ -0,0 +1,51 @@
...
+|^
+:-  %noun
+(export-all who)
+::
+++  export-app
+  |=  [who=ship agent=term]
+  ^-  @
+  %-  need
+  ;;  (unit @)
+  .^  *
```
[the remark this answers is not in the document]
ixv: I'm not sure why, honestly, but when I do it that way I get:
image

ixv:
Hm, after thinking about your point about app upgrades some more, I suppose theres a potential case where app upgrades end up requiring different app data to be imported. Maybe the thing to do is to export all data from all apps, and on import, just no-op for most apps.

pkg/arvo/gen/aqua-export.hoon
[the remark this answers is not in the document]
joemfb: You're peeking into aqua, so you're calling the `+peek` arm of arvo, which means your output will be untyped.

pkg/arvo/app/metadata-store.hoon
```diff
@@ -321,6 +329,12 @@
       %remove  (handle-remove group-path.act resource.act)
   ==
 ::
+++  poke-import
+  |=  jammed=@
+  ^-  (quip card _state)
+  =/  sty=state-4  ;;(state-4 (cue jammed))
```
liam-fitzgerald, line 335: Needs to be upgraded for current state number

pkg/arvo/app/chat-hook.hoon
```diff
@@ -644,6 +654,20 @@
     ==
   ==
 ::
+++  poke-import
+  |=  jammed=@
+  ^-  (quip card _state)
+  =/  sty=state-8  ;;(state-8 (cue jammed))
```
liam-fitzgerald, line 660: Also needs to be upgraded for current state number

liam-fitzgerald (changes requested):
Some type errors have crept in bc of the how long this has been up for. Otherwise, I have no issues with the code or the architecture. You probably need support for (group|graph)-(push|pull)-hooks, but otherwise looks good to me! (I'm happy to help with the lib/hooks, if you think they ought to be supported there)

pkg/arvo/app/chat-hook.hoon
```diff
@@ -469,7 +473,13 @@
     ==
   ::
   ++  on-leave  on-leave:def
-  ++  on-peek   on-peek:def
+  ++  on-peek
+    |=  =path
+    ^-  (unit (unit cage))
+    ?+  path  (on-peek:def path)
+        [%x %export ~]
+      ``noun+!>((jam state))
```
Fang-, line 481: Friendly reminder that we have a `%jam` mark. Probably nice to use that here.

pkg/arvo/app/chat-hook.hoon
Fang-: Why isn't this just `versioned-state` or w/e?

Fang- (commented):
Not sure why you pinged me for review, considering nothing changed since I left my previous one. (;

> I suppose theres a potential case where app upgrades end up requiring different app data to be imported. Maybe the thing to do is to export all data from all apps, and on import, just no-op for most apps.

But earlier you said:

> You wouldn't use it for individual apps.

So I'm a little bit confused now.

> If apps upgrade, we can add multiple cases to the +poke-import arms to handle the different versions.

The suggestion to just import the old version into state, and use on-load logic to upgrade, immediately comes to mind. If the on-load works, it works, right? But maybe there's concerns here about opening, then closing & re-opening old-style subscriptions or something?

> The alternative, I think, would be solidifying our userspace architecture patterns to the point where its certain you would only need to import to the store, or whatever.

This is, of course, the ideal. Once an "archived" flag or w/e makes it into stores, hooks can probably use that to decide what needs to be subscribed to, and what can remain disconnected.

matildepark:
<details>
<summary>Notes from import/export meeting, w/ @liam-fitzgerald @tacryt-socryp @ixv </summary>

- LA: Contact-view is being moved to a collection of threads instead of a single app and is currently slated to be done before the breach, so it’s relevant for Isaac to know that so that once we finish that work, he can keep track of it and spin up a thread to do that initialisation instead of what he’s currently doing, which is currently correct, which is poking the contact-view.
- LA: IV, similarly, have you seen the Publish migration code with the backoff? (No.) LF can fill you in, but the pattern you’re using to import remote metadata hook subscriptions is fragile and it would benefit from using the same pattern that Liam used from having periodic watches and backoffs.
- LF: You keep rewatching until you get a positive watch-ack with exponential backoff; once you get it, you then add it to the hook.
- IV: Doesn’t it eventually go through, regardless?
- LF: In the global breach case, sure, it’ll keep going until everything comes back online; in the individual breach case, it may take them some time before they actually restore all their stuff and before they restore their stuff, that’ll get nacked and it’ll fuck up your day.
- IV: Problem here, because in watch-ack, if they haven’t restored yet, we’ll get a nack right? (Yes.) And it’ll just be deleted from our synced. (Yes.) So it sounds like we need to add some extra state for things we’re resuming, I guess. (Yeah.)
- LA: IV, I understand what you’re doing so far and I think that it makes sense and if we do it right it’s not even fragile. My only comment so far is — have you considered actually having those import ack pokes go into a specific importer app that then has the corresponding logic instead of putting it into these apps?
- IV: In those cases the app needs to subscribe to something else so the apps do need to have some of their logic.
- LA: Some of the apps that have to subscribe to things have poke APIs that tell them to subscribe to things. All the hooks, you can poke it and it will watch what you’re telling it to watch, which is standardised in the push/pull hook.
- IV: The other thing is that it needs to inject all this state.
- LA: But what if we separated that process of sending pokes to various places, and watches with backoffs, into its own thing, and had each of the apps you’re putting state into just put state into them?
- IV: What’s the main benefit of that?
- LA: The logic would be in one place; particularly backoff logic could be reused. The way we did it in Publish, we didn’t even have the app itself send the watch until we had made the Publish app successfully watch. So once we got a watch-ack on Publish, then we sent the poke to graph-pull-hook and said “you should go watch [that resource] because we know for a fact they’ve finished their import”. If we want a similar pattern here, we’d have to do it in a library and implement it everywhere or … yeah. Just a suggestion. I don’t know how hard it’d be to do.
- IV: Also unsure. Will think some more, but the hard deadline of this gives me some pause.

**Contact-store**

- IV: Loads the state.

**Contact-hook**

- IV: Looks like it does the same thing as metadata-store, except if it’s ours we send the data to the store.
- IV: This might be an error in metadata-hook now that I think about it.
- LA: Might be a circular watch because you’d watch yourself on metadata-hook.
- IV: No, if its ours we do nothing, that’s the problem.
- LF: When we modernise metadata-store, do we want to move to the pattern where one path, you get everything on it, like group-store does? So we can avoid this? (Sure?)

**Invite-store**

- IV: Just loads everything because it doesn’t need to set up subscriptions to anything. I think there were changes to this as well?
- LA: It does the same stuff, just restructured. It just doesn’t use paths now.

**Chat-store**

- IV: Don’t know if it’s relevant, if it’s migrating … (Worth looking at.)
- IV: Loads everything, does nothing else.

**Chat-hook**

- IV: Same pattern as the other hooks; either watches our store or the foreign hook, and it watches on the mailbox path, which I think is right.
- LA: What if, before you initialise group-store, you initialise contact-view with an atom that just had the groups you wanted to send out the poke-creates for?
- IV: I’m sure that could work.
- LA: Then it would be like, the actual app itself would send the pokes and then … yeah.
- IV: As long as there’s no incongruities between the state of contacts and groups — there shouldn’t be, right? (No. Though LA can’t guarantee, sometimes they’ve gotten out of sync.)

Skipping Publish.

**Graph-store**

- IV: I’m most worried about this because I don’t know it as well as the other apps.
- IV: It jams the state for export. For poke-import, iterates through all the graphs and either pokes the push-hook if it’s ours with an add, or pokes the pull-hook with an add if it’s not ours.
- LA: You’re missing one thing — we need to set up clay subscriptions for any marks we’re using — see how it’s a marked graph?
- IV: Is there any logic for that already in this app? `(/sing`)
- LA: It’s in the poke pull create, there’s another `%sing` (add-graph). 
- IV: For each graph, we check the mark and sing that. (So we have it in the hot cache. Not necessary, just for performance.)

**Questions?**

- LA: I’m concerned about incongruities between invite-store, metadata-store, and contact-store.
- IV: It works right now, but those incongruities will just be preserved.
- LA: I don’t think they’ll cause more issues with your importing/exporting than they already do; but we have a good opportunity to have a clean slate that doesn’t have our old — currently we’re in a situation where we have made all these various decisions and done so many iterations, and at times messy changes and — 
- IV: The question I have is, what’s the rule then? We could do what you said with contact-store and use that as the basis for everything, for group-store and metadata-store. That could probably work. It’s just a matter of which apps’ data gets the axe.
- LA: I think we should axe — I think we should take in the group-store state, contact-store state and metadata-store state and anything that doesn’t have all three should be removed.
- IV: How exactly do unmanaged groups work these days?
- LA: It’s just a hidden flag, a boolean, in group-store, otherwise they’re normal.
- IV: You still have contact stuff and metadata stuff? (Metadata, not contacts.)
- MP: Is there an edge case where, if nobody set a contact there wouldn’t be an entry in contact-store? (It should have a container anyway.)
- IV: That sounds fine.
- LA: It would just be nice.
- LA: The clay mark thing is important; the reorganisation into its own app isn’t important but would be neat, and the validation of data thing is important, and if you don’t have time for it, you should tell us so we could put time to it ourselves.
- IV: Something I thought about and was never really able to figure out completely was doing this in a library somehow. Like in one of the hook libraries, but I don’t know if that would work.
- LA: I think the only special cased one so far is group-store, currently.
- IV: Why?
- LA: The stores just take in the state and do nothing … except group-store. The hooks just take in the state and send the appropriate subscription request. Those could actually just be libraries and the special case could be in contact-view or something.
- IV: I don’t know if that’s better than your other proposals, though.
- LF: I do think it should be an agent, but other than that it looks pretty good. I’m wondering whether there’s some kind of — if this problem and the whole “data validation app, %sane” thing is the same. If we have some kind of …
- IV: Does this run periodically or do you poke it to run? (You poke it.)
- LA: I feel as though this could literally just use the formula we’ve discussed in #3416, I feel like it would, if we checked to make sure we had group-store, metadata and contact store in the managed case and just metadata-store and group-store in unmanaged case, we could just run that and that would be it.
- IV: It depends how much confidence you have in those apps to not get into this state. (LA feel like they don’t happen now, it’s standardised and in few places.)
- IV: Liam, what’s the PR for the Publish stuff? (It’s done, it’s in release.) 
- LA: Cmd+F `bex`.

</details>

joemfb (changes requested):
I'd like to request (emphasis on *request*) a couple changes to this approach:

The individual app exports, and the combined export in `:lens`, should be tagged structures in a union (of one). It could be a constant `@da`, `%0`, or whatever, but they should be discriminable (the app states should probably just use their state version number).

Jamming/cueing each individual export is unnecessary, and will prevent some amount of structural sharing. The individual `/gx/.../export` handlers can simply produce a noun, the `+poke-import` arms can take a noun.

Neither of these changes would affect #3942

philipcmonk:
How do I run the ph tests here?  A few files have changed, and `-ph-migrate-groups` has a syntax error, is it not necessary?

ixv:
It is necessary, checking it in was a mistake. Also, @liam-fitzgerald pointed out some issues with the way I was handling exponential backoff in graph-store, and I have a fix coming shortly, but if you run the tests right now everything will resume connectivity properly except for graph-store.

The way to run these tests is in the order:
- init
- start
- make-groups
- make-chats
- make-graphs
- breach (for each ship, or all at once depending on args)
- start
- post-import-groups
- post-import-chats
- post-import-metadata-contacts
- post-import-graphs

The basic idea of it is the same as explained in the OP, the only difference is the make-data and test-import threads were broken up to deal with a smaller set of apps.

Also you might find useful the helper threads `-ph-migrate-send-his` to send a |hi from each ship to every other ship, and  `-ph-start-drivers` which just starts the vane threads and loops infinitely.

ixv:
I should note, these aren't exactly "tests" in the sense of telling you directly if something has gone wrong. I've found it more expedient to just check the debug state of apps manually after doing various operations, since reading in state in this environment can be tricky.

pkg/arvo/app/graph-store.hoon
```diff
@@ -792,9 +816,37 @@
...
   ==
 ::
-++  on-agent  on-agent:def
+++  on-agent
+  |=  [=wire =sign:agent:gall]
+  ^-  (quip card _this)
+  ?.  ?=([%try-rejoin @ *] wire)
+    (on-agent:def wire sign)
+  ?.  ?=(%watch-ack -.sign)
+    [~ this]
```
liam-fitzgerald, line 837: this should crash, so that the subscription automatically dies when it receives it's first `%fact`

ixv: I just realized I meant to do a %leave when it gets the watch-ack, that should be sufficient right?

philipcmonk:
This is going out tomorrow unless someone objects, @liam-fitzgerald @joemfb

joemfb:
I'm happy as long as the exports are version-tagged, and that appears to be the case for all of them. I haven't reviewed any of the (re)subscription logic or aqua changes.

liam-fitzgerald:
I'm happy with this as it stands

PR #3606, opened 2020-09-28
landscape: increase z-index of dropdown search

### LINKED ISSUE #3583: links: group dropdown appears below button
image

tylershuster:
fixes #3583

I figure this is safe because in general you want dropdowns to be above other things. LMK if there's a better pattern to use here.


matildepark (approved):
lgtm

PR #3607, opened 2020-09-28
landscape: delay uploading briefly to account for browser inconsistencies

### LINKED ISSUE #3589: chat: paste image stream no longer works in Firefox
Pasting an image (not a URL, a stream) from the clipboard in macOS into chat from Firefox no longer works. It works on other browsers and it works in the Links app.

tylershuster:
fixes #3589

Due to inconsistencies in how browsers handle "change" events, it is necessary to make a brief delay on S3 uploader to prevent double-uploading


matildepark (commented):
Safari -> Dragging and dropping dumps the image straight into Chat. Pasting the image stream returns a URL in the input.

Firefox (79, then updated to 81) -> Dragging and dropping returns a URL in the input. Pasting the image stream does nothing.

tylershuster:
My bad. I forgot this whole thing came about because Firefox doesn't like change handlers being fired asynchronously.. This should work.

matildepark (approved):
Behavioural parity established between Webkit and Gecko

PR #3609, opened 2020-09-28
landscape: self-host s3 sdk

tylershuster:
fixes https://github.com/urbit/landscape/issues/91

Does what it says. Get out of my house amazon I invite you you don't invite me


tylershuster:
Wait...this isn't even needed. It still works if I remove the S3 package. Must be included with the node package. If someone can confirm, I'll just PR a deletion instead.

matildepark:
Lol yeah we appear to remove it in b3139f0e52a2393ebf4e7a70cf9b9b50ef413e0b but then I restored it in e76e919755b2477d824acaa65641a001192d8128 like a dummy.

matildepark:
Just PR a deletion and we'll test quick but I think it was a mistake on my part.

PR #3611, opened 2020-09-28
landscape: remove externally loaded aws

tylershuster:


matildepark (approved):
Works. Didn't even need it here.

PR #3612, opened 2020-09-28
u3: replaces atom slab-allocation api for efficiency

joemfb:
Typical usage patterns of the old atom "slab" api lead to unconditional, unnecessary heap allocations. This PR replaces that API: a `u3i_slab` struct is used to track a slab allocation, and it includes one word of internal space, letting us allocations.

Additionally, this PR replaces some unnecessary use of `u3r_chop()` (XOR of arbitrary bit-slices of atoms) with straight copies. And `u3s_jam_fib()` has been refactored to use the slab api, enabling us to avoid re-allocation of the output atom in the `+jam` jet.

(Note to self: this new API needs to be documented in u3.md)

/cc @frodwith


Fang- (approved):
Most of my comments got resolved as the commits went on. New api seems a bit more legible, too. A couple comments/suggestions, but no show-stoppers here.

pkg/urbit/noun/imprison.c
```diff
@@ -3,228 +3,461 @@
...
+    return (c3_w)u3m_bail(c3__fail);
+  }
+  else {
+    *out_d = byt_d;
+    return wor_w;
+  }
+}
+
+/* _ci_slab_init(): initialize slab with heap allocation.
+**              NB: [len_w] must be >0
```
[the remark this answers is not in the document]
joemfb: The call-sites are factored to already check, and this is basically necessary due to the design of this api. But I've added a debug assertion that will be checked in CI.

pkg/urbit/noun/imprison.c
```diff
@@ -3,228 +3,461 @@
...
+      _ci_slab_init(sab_u, wor_w);
     }
-    else if ( a_w == 3 ) {
-      return (b_y[0] | (b_y[1] << 8) | (b_y[2] << 16));
+
+    sab_u->byt_d = byt_d;
+
+    //  XX refactor
+    //
+    memset(sab_u->buf_y, 0, (size_t)wor_w * 4);
```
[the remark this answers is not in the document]
joemfb: Good call.

pkg/urbit/noun/imprison.c
```diff
@@ -3,228 +3,453 @@
...
 */
-u3_noun
+u3_atom
 u3i_mp(mpz_t a_mp)
 {
-  c3_w  pyg_w = mpz_size(a_mp) * ((sizeof(mp_limb_t)) / 4);
-  c3_w* buz_w = u3a_slab(4 * pyg_w);
+  size_t   siz_i = mpz_sizeinbase(a_mp, 2);
+  u3i_slab sab_u;
+  u3i_slab_init(&sab_u, 0, siz_i);
```
joemfb, line 431: NB: the old implementation not only allocated unconditionally, but it measured the gmp integer with very low resolution, and often over-allocated. On 64-bit systems (ie, the only systems we currently support), `mp_limb_t` is 64 bits, so measurements were calculated in 128-bit increments (`(number-of-limbs * 2) * 4)`).

ohAitch: > On 64-bit systems (ie, the only systems we currently support)

I could swear urbit ran on 32bit, ARM, Raspberry Pi. Isn't the 2GB loom limit driven by pointers needing to fit in 32 bit registers?

joemfb: The ARM/Raspi stuff is all 64-bit. As far as I know, urbit hasn't worked on 32-bit architectures as long as I've been around. There's no reason it can't, the loom limitations are orthogonal. It just hasn't been a priority.

PR #3613, opened 2020-09-29
leap: use indigo-react icons

matildepark:
- Removes all the image logic gunk and just uses indigo-react icons for Leap results. (One step closer to getting PNGs out of Landscape.)
- Trims up OmniboxResult of some unnecessary duplicate logic for essentially a single color ternary.

cc @urcades @g-a-v-i-n 

Screen Shot 2020-09-28 at 9 13 01 PM
Screen Shot 2020-09-28 at 9 12 48 PM

### DIFF
--- a/pkg/interface/src/logic/lib/omnibox.js
+++ b/pkg/interface/src/logic/lib/omnibox.js
@@ -68,7 +68,7 @@ const appIndex = function (apps) {
     });
   // add groups separately
   applications.push(
-    result('Groups', '/~groups', 'groups', null)
+    result('Groups', '/~groups', 'Groups', null)
   );
   return applications;
 };
--- a/pkg/interface/src/views/components/leap/OmniboxResult.js
+++ b/pkg/interface/src/views/components/leap/OmniboxResult.js
@@ -26,43 +26,20 @@ export class OmniboxResult extends Component {
   }
 
   getIcon(icon, dark, selected, link) {
-    // graphicStyle is only necessary for pngs
-    //
-    //TODO can be removed after indigo-react 1.2
-    //which includes icons for apps
-    let graphicStyle = {};
-
-    if (icon.toLowerCase() !== 'dojo') {
-      graphicStyle = (!dark && this.state.hovered) ||
-        selected === link ||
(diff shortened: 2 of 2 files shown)


liam-fitzgerald:
consider setting `color="transparent"` and `stroke="black"` otherwise these icons look kinda wack

matildepark:
Ooo. OK one sec, will replace images. I knew they looked wack.

liam-fitzgerald:
LGTM!

urcades:
The Dojo carat/arrow is looking a little thick (its line weight is not equivalent to the other icons), but otherwise these are looking rather good. We should check if that's actually what's in Indigo, or if the icon is being transformed for some reason.

For reference:

Screen Shot 2020-09-28 at 9 52 49 PM

Have we shipped the icon-ified versions of sigils yet? We'd likely be using that in place of the fully-rendered sigil, so the end result isn't so busy.

Other than these two details, this is looking quite nice – really excited to start getting vector assets into Landscape.

urcades:
Actually, I take the last image example back, I believe we'd likely be using this icon as the symbol for "Dojo":

Screen Shot 2020-09-28 at 9 54 40 PM

If for whatever reason, this isn't in Dojo (I'm unaware of how versioning works, or how it maps to Figma Indigo), we can opt to use the simple rightward carat:

Screen Shot 2020-09-28 at 9 54 51 PM

matildepark:
[I don't think it is at present.](https://github.com/urbit/indigo-react/blob/master/src/Icon/iconIndex.tsx#L382) Sorry for being trigger happy on the merge here.

matildepark:
> The Dojo carat/arrow is looking a little thick (its line weight is not equivalent to the other icons), but otherwise these are looking rather good. We should check if that's actually what's in Indigo, or if the icon is being transformed for some reason.

It's because we're using a stroke on that one when the colour would do more precisely (as we do with the arrow for Log Out). I guess I should add yet another conditional for that case, probably.

g-a-v-i-n:
Something is wrong with the icons if they require you to use stroke. I'll fix this in next release, and add the new icons.

PR #3614, opened 2020-09-29
interface: indigo-react statusbar icons

matildepark:
We still had some hacky solutions for the statusBar icons, so we swap in the indigo-react versions; `Icon` no longer allows you to use `position` so I fixed badges while I was at it.

Screen Shot 2020-09-28 at 9 33 03 PM

cc @g-a-v-i-n


liam-fitzgerald (approved):
LGTM!

PR #3615, opened 2020-09-29
meta: simplify contributing guides

matildepark:
- The interface dev environment is very simple to set up. This PR attempts to better exemplify that.
- Because there have been a few situations in which users get tripped up on fakezods using boot pills, I add instructions to use the solid pill and Arvo directory from the repository.

### DIFF
--- a/pkg/interface/CONTRIBUTING.md
+++ b/pkg/interface/CONTRIBUTING.md
@@ -10,11 +10,12 @@ applications. Landscape applications will usually make good use of Gall, but
 it's not strictly required if a Landscape application is not interacting with
 ships directly.
 
-## Contributing to Landscape applications
+## Starting the dev environment
 
-To begin developing on Landscape, find the `urbitrc-sample` file found
-at `urbit/pkg/interface/config/urbitrc-sample`. Copy it as `urbitrc`.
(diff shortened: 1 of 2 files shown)


philipcmonk (approved):
both changes look good, glad CONTRIBUTING isn't so intimidating

jtobin (approved):
:+1: :+1:

pkg/interface/CONTRIBUTING.md
```diff
@@ -26,130 +27,17 @@ module.exports = {
...
-
-```javascript
-module.exports = {
-  URL: 'http://localhost:80'
-}
-```
-
-Then get everything installed:
+Change the URL to your livenet ship (if making front-end changes) or keep it the
+same (if [developing on a local development ship][local]). Then, from
```
Fang-, line 31: My local ships have never bound to `80` because I have something else running locally. Might want a sidenote here about using whatever port the ship prints on boot.

matildepark: *1 minute ago* god damn it mark

matildepark: I'm willing to accept the tradeoff that anyone with something else running locally knows that ports exist

Issue #3617, opened 2020-09-29
Can't change name of existing chat


timlucmiptev:
**Describe the bug**
When I try to change the name of an existing chat, it fails with the following error in devtools:
```
Uncaught (in promise) gall: poke-as cast fail :metadata-hook [a=%json b=%metadata-action]
syntax error
{1 1}
/lib/metadata-json/hoon:<[52 5].[52 21]>
/lib/metadata-json/hoon:<[51 5].[52 21]>
[%key 'color']
[%key 'metadata']
[%key 'add']
/lib/metadata-json/hoon:<[29 7].[29 23]>
/lib/metadata-json/hoon:<[29 3].[68 5]>
/lib/metadata-json/hoon:<[28 3].[68 5]>
/lib/metadata-json/hoon:<[27 3].[68 5]>
/mar/metadata/action/hoon:<[14 5].[14 25]>
[%grab %json %metadata-action]
```

**To Reproduce**
1. Have a chat in a group
2. Open devtools to observe console
3. Change its name

**Desktop (please complete the following information):**
 - OS: MacOS 10.15.6
 - Browser: Brave
 - Base hash: `oeaff`

Issue #3605, opened 2020-09-28
chat: group links with periods in them don't create links


tylershuster:
for example: ~forhet-losreb/comp.lang

How can we make the criteria for autolinking the same as the name verification check?

tylershuster:
Ok, this is actually different. You can create a group with a period in it, but if you try to directly visit the *same URL* that you get when you create the group, you get a white screen, presumably because it picks up the extension and tries to parse it as a file. Maybe we need to prevent the creation of groups with a period in them, but Arvo doesn't prevent this, so maybe it's a problem in how the router parses URLs. RFC.

matildepark:
We should just replace the dots with hyphens on FE, probably.

—
~haddef-sigwen
https://urbit.org

Issue #3592, opened 2020-09-26
Can't convert chat to group chat


timlucmiptev:
**Describe the bug**
When I try to convert a non-group chat and add it to a group of mine, I get the error:
```
“inclusive is not defined”
Stack trace
ReferenceError: inclusive is not defined
    at r.value (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:460:244518)
    at r.value (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:460:245505)
    at Io (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:69875)
    at Yo (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:69670)
    at gs (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:105469)
    at cl (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:96716)
    at sl (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:96641)
    at Qs (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:93671)
    at https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:319:45313
    at t.unstable_runWithPriority (https://timlucmiptev.space/~landscape/js/bundle/index.9f00eb9b1c58d2b1bd3c.js:327:3844)
If this is unexpected, email support@tlon.io or submit an issue.
```

**To Reproduce**
1. Create a new chat (not in a group)
2. Go to that chat's settings and add it to an existing group you own
3. See error

**Desktop (please complete the following information):**
 - OS: MacOS 10.15.6
 - Browser: Brave
 - Base hash: `oeaff`

PR #3618, opened 2020-09-29
interface: metadata UI fixes

matildepark:
- "Module" and "colour" were in the wrong order for the poke that sets the name of a resource, causing a poke fail. (#3617)
- Replaces `~` and `.` in new group / resource paths with `-` to avoid weird behaviours. (#3605)
- We didn't pass `station` nor `changeLoading` to the groupify button causing groupifying chats to fail. (#3592)


liam-fitzgerald (approved):
LGTM!

PR #3620, opened 2020-09-29
publish: disable cursor position reset on rerender

matildepark:
Addresses #3593.

Due to the refactor in #3540 we now use `useCallback`, and this prompts a rerender every single time we make a keystroke, pushing the cursor to the end. `autoCursor={false}` tells CodeMirror to not reset the cursor on a value change. We are going to want to prevent the rerender entirely, but I am thus far unsure how to finagle hooks — initial tips point to adding a [useRef](https://stackoverflow.com/questions/55045566/react-hooks-usecallback-causes-child-to-re-render).

(h/t to @tylershuster for pointing out the obvious fix)


liam-fitzgerald (approved):
LGTM!

Issue #3608, opened 2020-09-28
Malformed chat-hook state is preventing OTA


Fang-:
Copying over from #3581 for better visibility:

Trying to pull down the latest OTA fails for me with the following error. Apparently ~palfun hosts chat(s?) with no group associated with it.

https://github.com/urbit/urbit/blob/e17e508379fac7b87a331f15d47bdd72aeec6dbf/pkg/arvo/app/chat-hook.hoon#L912-L918

```
%load-failed
/app/chat-hook/hoon:<[917 3].[918 11]>
/app/chat-hook/hoon:<[915 3].[918 11]>
/app/chat-hook/hoon:<[914 3].[918 11]>
/app/chat-hook/hoon:<[107 25].[107 50]>
/app/chat-hook/hoon:<[107 9].[113 79]>
/app/chat-hook/hoon:<[106 9].[113 79]>
/app/chat-hook/hoon:<[105 9].[113 79]>
/app/chat-hook/hoon:<[103 9].[113 79]>
/app/chat-hook/hoon:<[102 9].[113 79]>
/app/chat-hook/hoon:<[101 9].[113 79]>
/app/chat-hook/hoon:<[100 7].[114 18]>
/app/chat-hook/hoon:<[99 7].[114 18]>
/app/chat-hook/hoon:<[91 7].[114 18]>
/app/chat-hook/hoon:<[90 5].[189 7]>
/app/chat-hook/hoon:<[88 5].[189 7]>
/app/chat-hook/hoon:<[87 5].[189 7]>
/app/chat-hook/hoon:<[86 5].[189 7]>
/app/chat-hook/hoon:<[85 5].[189 7]>
/app/chat-hook/hoon:<[84 5].[355 7]>
/app/chat-hook/hoon:<[83 5].[355 7]>
```

So I poked around the state of these things a bunch. This was kinda painful, chat-hook has no scry endpoints.

In any case, interesting findings. Every locally hosted chat in chat-store has a matching group. For chat-hook however, these two do not have matching groups:

```hoon
{ /~palfun-foslup/dm--sipnes-hanhes
  /~palfun-foslup/dm--naltyc-wornes
}
```

But then, we look at what chats are in chat store, and what groups are in groups-store. We see these entries:

```hoon
{ /~palfun-foslup/dm--~sipnes-hanhes
  /~palfun-foslup/dm--~naltyc-wornes
}
```

Note the `~` in front of the names.

[...]

PR #3625, opened 2020-09-30
eyre: minimal CORS support

Fang-:
Implement minimal CORS handling into Eyre, [as proposed on urbit-dev](https://groups.google.com/a/urbit.org/g/dev/c/bb82dwEJGzM/m/q2JjNSx5BwAJ).

Includes a `+cors-registry` generator for easily viewing the current CORS state, and `|cors-approve` and `|cors-reject` for allowing and denying origins access, respectively.

idk what's up with hood all of a sudden, but I needed b277b82 to get it to play nice and compile regardless of my other changes here.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:9b572ea1bc3bafc9ac7ff1a0fcecca16320bc2bbe6d6964cc76a2bc36f201975
-size 6299579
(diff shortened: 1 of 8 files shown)


vvisigoth:
Any idea when we can get this on-network?

A
—
~poldec-tonteg
http://urbit.org

Fang-:
Whenever we release `release/next-sys`. I think we said on Monday that we were due for that anyway. But we're hard-blocked by #3608.

matildepark:
That issue will almost certainly be cleared by Monday. I want a userspace OTA tomorrow that will have that in there.

PR #3626, opened 2020-09-30
interface: new groups layout + routing + skeleton

liam-fitzgerald:
This PR is a complete overhaul of the way that apps are laid out and accessed in landscape. App channels[1] are now contained within their associated groups. Channels that are unmanaged now live in the 'Home' space. 
Other changes: 
- Allows for custom sorting of the channels
- Brings chat messages down to 12px fontSize i.e. what we use everywhere else to ensure consistency
- Adds a popover engine using react portals, accessible using the `Dropdown` component. 

NB: needs cleanup to remove routes and views that are now unused. 
[1]: I use channel to mean 'resource of an app' e.g. a chat or notebook or collection.


vvisigoth:
pumped
—
~poldec-tonteg
http://urbit.org

pkg/interface/package.json
```diff
@@ -9,7 +9,7 @@
     "@reach/menu-button": "^0.10.5",
     "@reach/tabs": "^0.10.5",
     "@tlon/indigo-light": "^1.0.3",
-    "@tlon/indigo-react": "1.2.6",
+    "@tlon/indigo-react": "urbit/indigo-react#lf/1.2.8",
```
matildepark, line 12: The secret Liam changes ...

matildepark:
We're going to have to change the omnibox index for all these new routes I think

pkg/interface/src/views/components/ShipSearch.tsx
```diff
@@ -0,0 +1,159 @@
...
+            px={2}
+            py={1}
+            border={1}
+            borderColor="washedGrey"
+            color="black"
+            fontSize={0}
+            mt={2}
+            mr={2}
+          >
+            <Text fontFamily="mono">{cite(s)}</Text>
```
matildepark, line 150: Btw for future reference you can just do `<Text mono>` iirc

pkg/interface/package.json
liam-fitzgerald: PR'd https://github.com/urbit/indigo-react/pull/71

matildepark:
Might be worth merging next userspace back in to, eg. clean up Leap etc

tacryt-socryp (approved):

PR #3628, opened 2020-10-01
eyre: add authentication checker scry endpoint

Fang-:
Lets you check whether a specific `Cookie` header value string constitutes an authenticated request.

eg: `/ex/=//=/authenticated/cookie/(scot %t 'cookie-string')`

Intended for use in the runtime, for example with #3557.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -2422,6 +2422,14 @@
         %approved  ``noun+!>((~(has in approved.cors-registry) u.origin))
         %rejected  ``noun+!>((~(has in rejected.cors-registry) u.origin))
       ==
+    ::
+        [%authenticated %cookie @ ~]
+      ?~  cookies=(slaw %t i.t.t.tyl)  [~ ~]
+      :^  ~  ~  %noun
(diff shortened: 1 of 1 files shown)


ohAitch:
This still won't let you open a webslog in the *middle* of an event right? There's no "we interrupt this broadcast to compute a high-priority scry" worker IPC type?

ohAitch:
I guess it wouldn't hurt to like, cache these for 10 minutes - you probably want a proactive "a session logged out so recheck all currently active connections" event anyway, even outside of webslog.

joemfb:
@ohAitch, that's right, we'd have to buffer slog output to support such a use-case. I believe king-haskell already does so, but I'm not sure it's important that we do so in vere.

eglaysher (approved):

Fang-:
> you probably want a proactive "a session logged out so recheck all currently active connections" event anyway, even outside of webslog.

This is a good call. Expanding the vane-io interface for such things feels a little bit icky to me, but I'm not sure why. I think there's unanswered questions around the scope/style of those interfaces in general...

The simpler fix might be to just run that check on a five-minute timer. You're usually not gonna have very many of these open.

PR #3630, opened 2020-10-01
chat-hook: remove app-path from hook state when it has no correspondi…

tacryt-socryp:
…ng group, and do not crash in prior upgrade if chat has no group

### DIFF
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -23,8 +23,10 @@
       state-6
       state-7
       state-8
+      state-9
   ==
 ::
++$  state-9  [%9 state-base]
 +$  state-8  [%8 state-base]
 +$  state-7  [%7 state-base]
 +$  state-6  [%6 state-base]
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/chat-hook.hoon
```diff
@@ -85,8 +87,20 @@
     =/  old  !<(versioned-state old-vase)
     =|  cards=(list card)
     |-
-    ?:  ?=(%8 -.old)
+    ?:  ?=(%9 -.old)
       [cards this(state old)]
+    ?:  ?=(%8 -.old)
```
matildepark, line 92: The people encountering this issue haven't successfully processed the last state migration, so wouldn't their state still be at `%7`?

tacryt-socryp: Yes, hence the change to the logic in the `%7` upgrade case so as not to crash when encountering a chat without a corresponding group.

Fang-: You could probably get away with including this in the upgrade logic for `%7` then, considering that upgrade hasn't happened yet for everyone who has syncs that would be deleted by this.

Also you're deleting from `synced` but not `allow-history`? I get the latter is only for things we're hosting, but the former contains those too, right?

PR #3633, opened 2020-10-02
landscape: better embed checking

tylershuster:
@matildepark Would you try to break this? I can't, but I don't see why this change would cover every use case.


tylershuster:
This does fix an error I'm seeing on livenet

matildepark:
Merge next-userspace back into this so we can test it with the new layout? Might be able to sneak this in.

tylershuster:
Ok, it'll be an hour or so

tylershuster:
Wait maybe that was it

matildepark (approved):
Prevents false positive embeds, but we still have the issue of embeds unloading when out of view

PR #3634, opened 2020-10-02
leap: error trapping, fixes icon bug

tylershuster:
This fixes an unfiled bug where any leap results including a command in them would crash the whole app because the lowercase app name would fail to be recognized as an icon, and leap was not contained in an error boundary.

For future error trapping it also wraps the main components of the app in error boundaries.


tylershuster:
Now filed as #3635

matildepark (approved):

PR #3639, opened 2020-10-02
interface: indigo-reacting components, removing popout logic

matildepark:
- Migrate most of the components in the global components folder to indigo-react
- Because this made the old chat / links menus look weird, migrated their settings pages to indigo-react too
- Removing popout from everywhere in the interface, every conditional piece of display logic etc


liam-fitzgerald (approved):
LGTM!

PR #3644, opened 2020-10-05
arvo: restore :time app

joemfb:
This PR restores the `:time` app, which is a simple, low-resolution tool for timing a computation in the dojo (estimated via the elapsed time between the beginning of the computation, and the arrival of a subsequent timer event schedule for `+(now)`).

It's usefulness had been greatly reduced by the integration of `cc-release` (in v0.8.0), due to issues with event-timestamp granularity. It was removed with the integration of static-gall and spider, replaced by the `-time` thread. But the latter is less useful, as it doesn't support arbitrary expressions. And the event-timestamp issues have been resolved in `v0.10.8`.

### DIFF
--- a/pkg/arvo/app/time.hoon
+++ b/pkg/arvo/app/time.hoon
@@ -1,0 +1,37 @@
+::
+/+  default-agent, verb
+::
+|%
+::
++$  card  card:agent:gall
+--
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
Nice

PR #3646, opened 2020-10-05
interface: removed unused Chat components

tacryt-socryp:
Still have two TODOs to figure out DM creation and missing chat redirect.

Edit: Solved the DM creation and redirect to DM flow.

*This is good to go.*


pkg/interface/src/views/apps/chat/components/lib/profile-overlay.js
```diff
@@ -85,7 +87,7 @@ export class ProfileOverlay extends PureComponent {
           <div className="mono gray2">{cite(`~${ship}`)}</div>
           {!isOwn && (
             <Link
-              to={`/~chat/new/dm/~${ship}`}
+              to={`/~todo/~${ship}`}
```
tacryt-socryp, line 90: TODO

pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
```diff
@@ -127,7 +127,8 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
     const { isChatMissing, history, envelopes, mailboxSize, stationPendingMessages, unreadCount } = this.props;
 
     if (isChatMissing) {
-      history.push("/~chat");
+      //  TODO: fix this and push a different route
+      history.push("/~404");
```
tacryt-socryp, line 131: TODO

tacryt-socryp: Actually, I think this is the correct behavior here, to show a 404.

matildepark:
Probably, because UnjoinedResource handles all other join cases I think. Actually does Leap?

—
~haddef-sigwen
https://urbit.org

pkg/interface/src/views/apps/chat/components/lib/profile-overlay.js
liam-fitzgerald: I'd suggest `/~groups/home/new/dm/<patp>`

pkg/interface/src/views/apps/chat/components/lib/ChatWindow.tsx
liam-fitzgerald: Agreed

liam-fitzgerald (approved):
LGTM!

matildepark:
Sorry, probably shouldn't be touching chat while you're doing this. (Conflict, pull in the target)

pkg/interface/src/views/apps/chat/components/lib/profile-overlay.js
tacryt-socryp: Performed the logic inline, seems neater, since this is the only place we can create a new DM.

PR #3647, opened 2020-10-06
eyre: prevent binding in reserved namespaces

Fang-:
Disallows registering bindings (through `%connect` and `%serve`) that would capture
traffic on paths starting with `/~` (Eyre's) or `/~_~` (runtime's, as of cc389c5).

Note that we don't touch `+insert-binding`, which is used by Eyre internally to
set up bindings in its own namespace.


eglaysher (approved):
`~_~`

PR #3648, opened 2020-10-06
interface: removed unused links files

tacryt-socryp:


pkg/interface/src/views/apps/links/components/link-item.js
```diff
@@ -33,7 +33,7 @@ export const LinkItem = (props) => {
     ? <img src={props.avatar} height={36} width={36} className="dib" />
     : <Sigil ship={`~${author}`} size={36} color={'#' + props.color} />;
 
-  const baseUrl = props.baseUrl || `/~link/${resource}`;
+  const baseUrl = props.baseUrl || `/~404/${resource}`;
```
matildepark, line 36: Why not just `props.baseUrl`

matildepark (approved):

matildepark:
Hang on for Publish until I PR some fixes for a few crashes shortly

pkg/interface/src/views/apps/links/components/link-item.js
tacryt-socryp: This is in the case that props.baseUrl is null or undefined. Better to explicitly redirect to /~404 imo

matildepark:
Ok ping when you're ready to merge this

tacryt-socryp:
It's ready to go.

PR #3649, opened 2020-10-06
publish: crashes and regressions

matildepark:
- Fix unmanaged notebooks (or any notebook in a group with no contacts) crashing
- Fix a crash in the groupify menu for notebooks
- Prevent duplicate results from showing when searching an exact patp
- Send 'read' actions when loading a note


vvisigoth:
yes
—
~poldec-tonteg
http://urbit.org

tacryt-socryp (approved):
lgtm

PR #3650, opened 2020-10-06
publish: prevId -> nextId

ryjm:
`nextUrl` updated to the id of the next note rather than the previous note.

@matildepark @tylershuster


matildepark (approved):
Good catch

PR #3651, opened 2020-10-06
launch: fixing layout

tylershuster:
Big changes to how tiles are rendered. Still needs more work but this is better than it was.

~Also incidentally starts to fix some of the weird text-sizing in chat, awaiting further indigoification~

### DIFF
--- a/pkg/interface/src/views/apps/launch/components/tiles/basic.js
+++ b/pkg/interface/src/views/apps/launch/components/tiles/basic.js
@@ -1,7 +1,6 @@
 import React  from 'react';
 import classnames from 'classnames';
-import { Link } from 'react-router-dom';
-import defaultApps from '~/logic/lib/default-apps';
+import { Text, Icon } from '@tlon/indigo-react';
 
 import Tile from './tile';
 
@@ -9,47 +8,22 @@ export default class BasicTile extends React.PureComponent {
   render() {
     const { props } = this;
 
-    const children = (
-      <span>
-        <p className={
-           classnames('absolute f9',
-           { 'black white-d': props.title !== 'Dojo',
-             'white': props.title === 'Dojo' })}
-           style={{ left: 8, top: 8 }}
-        >
-          {props.title}
-        </p>
-         <img
-           className={classnames('absolute',
-           { 'invert-d': props.title !== 'Dojo' })}
-           style={{ left: 38, top: 38 }}
-           src={props.iconUrl}
-           width={48}
-           height={48}
-         />
-      </span>
-    );
-
-    const routeList = defaultApps.map((e) => {
-      return `/~${e}`;
-    });
-
-    const tile =  ( routeList.indexOf(props.linkedUrl) !== -1 ) ? (
-      <Link className="w-100 h-100 db pa2 no-underline" to={props.linkedUrl}>
-       {children}
-      </Link>
-    ) : (
-      <a className="w-100 h-100 db pa2 no-underline" href={props.linkedUrl}>
-       {children}
-      </a>
-    );
-
     return (
-      <Tile>
-        <div className={classnames('w-100 h-100 relative ba b--gray3 b--gray2-d bg-gray0-d br2',
-        { 'bg-white': props.title !== 'Dojo',
-          'bg-black': props.title === 'Dojo' })}
-        >{tile}</div>
+      <Tile
+        bg={props.title === 'Dojo' ? '#000000' : 'white'}
+        to={props.linkedUrl}
+      >
+        <Text color={props.title === 'Dojo' ? '#ffffff' : 'black'}>
+          {props.title === 'Dojo'
+            ? <Icon
+              icon='ChevronEast'
+              color='transparent'
+              color='#fff'
+              style={{ position: 'relative', top: '.3em'}}
+            />
+            : null
+          }{props.title}
+        </Text>
       </Tile>
     );
   }
--- a/pkg/interface/src/views/apps/launch/components/tiles/clock.js
+++ b/pkg/interface/src/views/apps/launch/components/tiles/clock.js
@@ -63,13 +63,21 @@ const circle = (ctx, x, y, r, from, to, fill) => {
   ctx.fill();
 };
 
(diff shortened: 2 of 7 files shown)


matildepark:
> Also incidentally starts to fix some of the weird text-sizing in chat

Hey man! You said this was going to be atomic. What did you change chat to look like, we may want to preserve those classes, which were added like yesterday

tylershuster:
Ok fine I fatfingered the `git add` command. I don't think I changed it at all. Did you check out the`f8`/`f9` removals? I was trying to fix an issue where URLs rendered weirdly small. Was that recently added? I suppose I can just back out

matildepark:
URLs do render at 14px for whatever reason on next-userspace yeah

matildepark:
Should be 16px which should be ... f8?

tylershuster:
I recant

pkg/interface/src/views/apps/launch/components/tiles/basic.js
```diff
@@ -1,55 +1,24 @@
...
-        { 'bg-white': props.title !== 'Dojo',
-          'bg-black': props.title === 'Dojo' })}
-        >{tile}</div>
+      <Tile
+        bg={props.title === 'Dojo' ? 'black' : 'white'}
+        to={props.linkedUrl}
+      >
+        <Text color={props.title === 'Dojo' ? 'white' : 'black'}>
+          {props.title === 'Dojo'
+            ? <Icon icon='ChevronEast' color='transparent' stroke='white' display='inline' style={{ position: 'relative', top: '.3em'}} />
```
matildepark, line 18: For future ref you should try to cut lines at 80 columns (though I myself am guilty of this)

matildepark, line 18: Oh also this one doesn't need `stroke`, you can get away with just `color='white'` here to make it thinner

pkg/interface/src/views/apps/launch/components/tiles/basic.js
```diff
@@ -1,55 +1,24 @@
...
-    );
-
     return (
-      <Tile>
-        <div className={classnames('w-100 h-100 relative ba b--gray3 b--gray2-d bg-gray0-d br2',
-        { 'bg-white': props.title !== 'Dojo',
-          'bg-black': props.title === 'Dojo' })}
-        >{tile}</div>
+      <Tile
+        bg={props.title === 'Dojo' ? 'black' : 'white'}
```
matildepark, line 13: This reverses in dark mode when you really just want it to always be black regardless I think. I think you can just pass a hex color instead of using the theme's keywords in that case

image

pkg/interface/src/views/apps/launch/components/tiles/clock.js
```diff
@@ -63,13 +63,21 @@ const circle = (ctx, x, y, r, from, to, fill) => {
   ctx.fill();
 };
 
+const circleClip = (ctx, x, y, r, from, to, fill) => {
+  ctx.globalCompositeOperation = 'xor';
+  circle(ctx, x, y, r, from, to, fill);
+  ctx.globalCompositeOperation = 'source-over';
+};
+
```
matildepark, line 71: Produces this effect in light mode with a dark background

image

tylershuster: That doesn't look good

pkg/interface/src/views/apps/launch/components/tiles/weather.js
```diff
@@ -207,13 +173,9 @@ export default class WeatherTile extends React.Component {
...
       >
-          <p className="f9 absolute"
-            style={{ left: 8, top: 8 }}
-          >
-            Weather
-          </p>
-        <p className="absolute w-100 flex-col f9"
-        style={{ bottom: 8, left: 8, cursor: 'pointer' }}
+          <p className="f9">Weather</p>
+        <p className="w-100 flex-col f9"
```
matildepark, line 177: Now a little too high 
Screenshot 2020-10-06 at 3 54 25 PM

pkg/interface/src/views/apps/launch/components/tiles/weather.js
```diff
@@ -136,67 +103,66 @@ export default class WeatherTile extends React.Component {
...
+          className="f9 black white-d pointer"
           onClick={() =>
             this.setState({ manualEntry: !this.state.manualEntry })
           }
         >
           &lt;&#45;
         </a>
         {secureCheck}
-        <p className="f9 pt5">
+        <Text mt='auto' pb={1}>
```
matildepark, line 132: Seems a little too low now? Should be at the top of the tile but below the arrow I think

Screenshot 2020-10-06 at 3 54 28 PM

pkg/interface/src/views/apps/launch/components/tiles/clock.js
tylershuster: This is due to making the hole transparent. It just responds to light/dark mode, nothing else. Should I just fill in the hole? I don't see a good way to detect background color, especially with images.

matildepark: This is what I usually use `mix-blend-mode` for but it needs a background so you'd need to like *go up the DOM* and inherit background colors or set it according to a prospective contrast ratio? Idk man elaborate shit

matildepark: If you want to just call it for now you can plug the hole, it's not ideal but better than the messy case here

tylershuster: Ok, I think I will. In the future I'm going to rewrite this in SVG so we can use mix-blend-mode (and to draw the moon phase), but I'll just plug the hole. Don't tell Ed, he'll be heartbroken

tylershuster:
Do you get notifications on every force-push? Sorry if you do

matildepark:
OK pull in next-userspace and move the `stroke`s to `color`, sorry, ha

matildepark (approved):

PR #3652, opened 2020-10-06
La/cleanup publish

tacryt-socryp:
- fixes route redirects
- removes unused files


pkg/interface/src/views/apps/publish/components/CommentInput.tsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 import * as Yup from "yup";
 import { Formik, FormikHelpers, Form, useFormikContext } from "formik";
-import { AsyncButton } from "../../../../components/AsyncButton";
+import { AsyncButton } from "../../../components/AsyncButton";
```
matildepark, line 4: we can just use `~/` etc and save the trouble of thinking about this

tacryt-socryp: True, but at three ..s I think it's okay to stay as is for now...

tylershuster: The rubric I try to follow here is "is this coming from core, or does it pertain to the local scope?"

matildepark (approved):

PR #3653, opened 2020-10-06
La/cleanup groups

tacryt-socryp:
- Deletes old groups files
- Moves new Groups files out of ~/views/apps/groups into ~/views/landscape


pkg/interface/src/views/apps/chat/components/profile-overlay.js
```diff
@@ -91,7 +91,7 @@ export class ProfileOverlay extends PureComponent {
 
     const identityHref = group.hidden
       ? '/~profile/identity'
-      : `/~groups/view${association['group-path']}/${window.ship}`;
+      : '/~404';
```
matildepark, line 94: Which means if we click on someone's sigil in the overlay we now get a 404? We should just not link to anything in that case.

pkg/interface/src/views/landscape/components/Content.js
```diff
@@ -42,7 +42,7 @@ export const Content = (props) => {
         <Route
           path='/~groups'
```
matildepark, line 43: Dropped a route?

pkg/interface/src/views/landscape/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+    this.props.subscription.startApp('graph');
+    
+  }
+
+  componentWillUnmount() {
+    this.props.subscription.stopApp('groups')
+    this.props.subscription.stopApp('chat')
+    this.props.subscription.stopApp('publish');
+    this.props.subscription.stopApp('graph');
+  }
```
matildepark, line 42: If you're in Landscape most of the time does it make sense to really open and close all of these just navigating to the home screen? Seems like we're finally at "subscribe to everything and never leave" life?

pkg/interface/src/views/landscape/index.tsx
```diff
@@ -0,0 +1,91 @@
...
+        props.invites['/contacts'] : {};
+    const s3 = props.s3 ? props.s3 : {};
+    const groups = props.groups || {};
+    const associations = props.associations || {};
+    const { api } = props;
+
+
+    return (
+      <Switch>
+        <Route path="/~groups/ship/:host/:name"
```
matildepark, line 64: Oh, you just haven't rewritten these yet anywhere, ok.

pkg/interface/src/views/landscape/index.tsx
tacryt-socryp: u rite m8

tacryt-socryp:
Should be ready to go. ;)

pkg/interface/src/views/landscape/index.tsx
matildepark: I can see a bunch of removals but don't we want to initiate these subscriptions in a central place? It looks like we'd just start already started subscriptions a lot.

tacryt-socryp: Pretty sure startApp is idempotent

matildepark:
You still have to change the index logic here, etc:

https://github.com/urbit/urbit/blob/3d1ca6fb424b3a8b903992b33dec8495fd0dc50f/pkg/interface/src/logic/lib/omnibox.js#L112-L118

pkg/interface/src/views/landscape/index.tsx
matildepark: It just printfs iirc

matildepark: I mean console.logs, not like, in Dojo

tacryt-socryp:
Done! I think this should work for omnibox.

pkg/interface/src/logic/lib/omnibox.js
```diff
@@ -4,7 +4,7 @@ import { cite } from '~/logic/lib/util';
   const indexes = new Map([
     ['commands', []],
     ['subscriptions', []],
-    ['groups', []],
+    ['landscape', []],
```
matildepark, line 7: this has ramifications for the Omnibox component, since we return this object to the state there. It will also rename the header to "Landscape" in the results, so do we really need to rename this?

tacryt-socryp: Uh no we don't need to rename it. Idk how omnibox works...

matildepark: tl;dr we basically make a list here corresponding to headers in the results, as well as to categories in the returned map

https://github.com/urbit/urbit/blob/cbe58f05199154bfc34caf8f9789485d9669f728/pkg/interface/src/views/components/leap/Omnibox.js#L54-L56

and then iterate through it

https://github.com/urbit/urbit/blob/cbe58f05199154bfc34caf8f9789485d9669f728/pkg/interface/src/views/components/leap/Omnibox.js#L149-L161

and make header'd results with it

https://github.com/urbit/urbit/blob/cbe58f05199154bfc34caf8f9789485d9669f728/pkg/interface/src/views/components/leap/Omnibox.js#L225-L251

matildepark: Arguably you could replace `getSearchedCategories` with the keys of `state.index` and it would be fine tho — but it would still write "landscape" as a header heh

tacryt-socryp: Uh... can you do this omnibox part? Seems delicate

matildepark: incoming

matildepark (approved):

PR #3655, opened 2020-10-06
launch: display group invites on home screen

matildepark:
The new layout has no way to do this, so we iterate them as tiles with 'accept' and 'reject' iconography.

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -78,7 +78,7 @@ export default class LaunchApp extends React.Component {
               weather={props.weather}
             />
           </Box>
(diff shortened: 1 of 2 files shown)


tacryt-socryp:
Looks lovely. Perhaps we can ensure they are sorted at the beginning?

matildepark:
They're in a separate map before the groups iteration, so they always come first.

tacryt-socryp (approved):

vvisigoth:
can a brotha get a screenshot
—
~poldec-tonteg
http://urbit.org

matildepark:
This with some additional 'accept' and 'reject' text at the bottom

Messages Image(3749392606)

PR #3658, opened 2020-10-06
u3: improve efficiency of list jets by cons'ing in-order

joemfb:
This PR adds `u3i_defcons()`, a deferred cell constructor, and uses it in the list jets. This replaces a pattern of constructing lists in reverse order, and flopping them when complete, and therefore avoids an unnecessary round cell allocations in each. This optimization is often called [tail-recursion modulo cons](https://en.wikipedia.org/wiki/Tail_call#Tail_recursion_modulo_cons). We may want to consider adding it to the bytecode interpeter at some point in the future.

### DIFF
--- a/pkg/urbit/jets/b/flop.c
+++ b/pkg/urbit/jets/b/flop.c
@@ -3,46 +3,29 @@
 */
 #include "all.h"
 
+u3_noun
+u3qb_flop(u3_noun a)
+{
+  u3_noun i, t = a, b = u3_nul;
 
-/* functions
-*/
-  u3_noun
-  u3qb_flop(u3_noun a)
-  {
-    u3_noun b = 0;
-
-    while ( 1 ) {
-      if ( u3_nul == a ) {
-        return b;
-      }
-      else if ( c3n == u3du(a) ) {
-        u3z(b);
-
-        return u3m_bail(c3__exit);
-      }
-      else {
-        b = u3nc(u3k(u3h(a)), b);
-        a = u3t(a);
-      }
-    }
-  }
-  u3_noun
-  u3wb_flop(u3_noun cor)
-  {
-    u3_noun a;
-
-    if ( u3_none == (a = u3r_at(u3x_sam, cor)) ) {
-      return u3m_bail(c3__exit);
-    } else {
-      return u3qb_flop(a);
-    }
-  }
-  u3_noun
-  u3kb_flop(u3_noun a)
-  {
-    u3_noun b = u3qb_flop(a);
-
-    u3z(a);
-    return b;
+  while ( u3_nul != t ) {
+    u3x_cell(t, &i, &t);
+    b = u3nc(u3k(i), b);
   }
 
+  return b;
+}
+
+u3_noun
+u3wb_flop(u3_noun cor)
+{
+  return u3qb_flop(u3x_at(u3x_sam, cor));
+}
+
+u3_noun
+u3kb_flop(u3_noun a)
+{
+  u3_noun b = u3qb_flop(a);
+  u3z(a);
+  return b;
+}
--- a/pkg/urbit/jets/b/murn.c
+++ b/pkg/urbit/jets/b/murn.c
@@ -29,8 +29,9 @@ u3qb_murn(u3_noun a, u3_noun b)
 
         if ( u3_nul != res ) {
           *lit = u3i_defcons(&hed, &tel);
-          *hed = u3t(res);
+          *hed = u3k(u3t(res));
           lit  = tel;
+          u3z(res);
         }
       }
       while ( u3_nul != t );
--- a/pkg/urbit/jets/b/murn.c
+++ b/pkg/urbit/jets/b/murn.c
@@ -3,54 +3,50 @@
 */
 #include "all.h"
 
-  u3_noun
-  _murn_in(u3j_site* sit_u, u3_noun a)
-  {
-    if ( 0 == a ) {
-      return a;
-    }
-    else if ( c3n == u3du(a) ) {
-      return u3m_bail(c3__exit);
-    }
-    else {
-      u3_noun one = u3j_gate_slam(sit_u, u3k(u3h(a)));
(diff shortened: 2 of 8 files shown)


pkg/urbit/jets/b/flop.c
```diff
@@ -3,46 +3,29 @@
...
-
-        return u3m_bail(c3__exit);
-      }
-      else {
-        b = u3nc(u3k(u3h(a)), b);
-        a = u3t(a);
-      }
-    }
+  while ( u3_nul != t ) {
+    u3x_cell(t, &i, &t);
```
ohAitch, line 12: Calling u3x_cell with an out pointer of its argument is well-defined right? Bc it's a pointer passed by-value

pkg/urbit/jets/b/murn.c
```diff
@@ -3,54 +3,50 @@
...
+      u3_noun  res, i, t = a;
+
+      do {
+        u3x_cell(t, &i, &t);
+
+        res = u3j_gate_slam(&sit_u, u3k(i));
+
+        if ( u3_nul != res ) {
+          *lit = u3i_defcons(&hed, &tel);
+          *hed = u3t(res);
```
ohAitch, line 32: Per the original `_murn_in`, you want `u3k(hed)` and `u3z(res)`, I don't think `u3t` does anything to reference counts.

pkg/urbit/noun/imprison.c
```diff
@@ -465,6 +465,37 @@ u3i_vint(u3_noun a)
...
+*/
+u3_cell
+u3i_defcons(u3_noun** hed, u3_noun** tel)
+{
+  u3_noun pro;
+  u3t_on(mal_o);
+
+#ifdef U3_CPU_DEBUG
+  u3R->pro.cel_d++;
+#endif
```
ohAitch, line 479: Should this just move to `u3a_celloc` instead of duplicating it?

pkg/urbit/noun/imprison.c
```diff
@@ -465,6 +465,37 @@ u3i_vint(u3_noun a)
...
+
+  {
+    c3_w*     nov_w = u3a_celloc();
+    u3a_cell* nov_u = (void *)nov_w;
+
+    nov_u->mug_w = 0;
+
+    //  XX zero-initialize head and tail?
```
ohAitch, line 487: Can't hurt, we're about to write to it again relatively soon right? `u3_none` may be more appropriate, not sure what the pitfalls of storing one in memory are.

ohAitch (commented):
One suspected memory leak, rest looks generally sane.

pkg/urbit/jets/b/flop.c
joemfb: Right, the noun reference (first argument) is passed by value.

pkg/urbit/jets/b/murn.c
joemfb: Good call, this leaks the outer cell.

pkg/urbit/noun/imprison.c
joemfb: It certainly can't hurt. But can it help? This API is somewhat unsafe compared to `u3i_cell()` (as the function comment says, you *must* stored something in both head and tail pointers). We spend a massive percentage of runtime allocating cells, so any memory stores that we can elide in that codepath seem worth pursuing.

ohAitch: I claim the strong form of "can't hurt", that it will not measurably affect performance.

All current callers immediately write to `hed`, and my expectation is two consecutive stores to the same location in L1 are much less expensive than the generic memory store, as are ones to two consecutive words. (The shadowed one is hopefully outright optimized out when inlining, I know gdb has sth like https://godbolt.org/ which could be used to check).

Probably the "correct" course here is to profile how much time the PR saves overall when replaying a representative ship(or w/e fleet sandbox test code), and rerun that both with and without the returning uninitialized memory speedup. Otoh the change in general is obviously beneficial for both speed and readability, so that may be not otherwise be warranted. @philipcmonk thoughts?

ohAitch:
+1 on trying to get this into the bytecode, esp if the `#` operator implementation doesn't already do sth very similar for the non-unique pointer case.

\<hobbyhorse>also just cons in general, since the operator in question is a travesty and should be removed from the spec and left as an interpreter implementation suggestion, oh how to treat `[[0 2] ...]` &c\</hobbyhorse>

pkg/urbit/noun/imprison.c
eglaysher: If you prefill these, it might be nice to fill them with dead beef instead of 0 so improper usage crashes instead of being an implicit null terminated list.

eglaysher (approved):

pkg/urbit/noun/imprison.c
ohAitch: Is `u3__none` a reasonable deadbeef, or do we have some other canonical non-`Option`-oriented sentinel?

joemfb: `u3_none` (not-a-noun) is the right sentinel value. I've updated this to initialize to it under U3_MEMORY_DEBUG. It's a bug and misuse of the api to ever miss a store through one of these pointers, and the only reason to initialize them is to help catch such a bug.

joemfb:
Here's same basic measurements of the cell allocation savings in this PR:

To take this, I built with U3_CPU_DEBUG to count cell allocations (and added some code to the runtime to print the number of cell allocations on serf exit and shutdown the ship as soon as replay completes, both of which should probably have cli arguments). I prepared a couple different event logs (brass boot and initialization, solid boot, playing around in the dojo, dm'ing myself through web-chat), and then replayed the logs on both `release/next-vere` and this branch.

brass boot (28 events):

- before `1.446.477.804`
- after `1.392.536.184`

savings of `53.941.620`, or roughly 3%.


solid boot, plus dojo and webchat (886 events):

- before: `2.791.775.948`
- after: `2.744.122.652`

savings of `47.653.296`, or roughly 1.7%.

PR #3659, opened 2020-10-07
interface: more design nits

matildepark:
Just pushing this because I can see you're working on the same tasks. -_-


liam-fitzgerald (approved):
LGTM!

Issue #3657, opened 2020-10-06
u3: |fl jets have memory leaks, double-count GMP integers


joemfb:
As discussed in #3656, `u3i_mp()` *transfers* its return, but the `|fl` jets incorrectly double-count these references in several places.

ohAitch:
Rust rewrite of all the jets when ;)
We already have https://github.com/rsaarelm/urbit-u3-rs!

joemfb:
Fix released (memory leaks, not rust rewrite)

PR #3660, opened 2020-10-07
u3: fixes incorrect double ref-counting in |ff jets

joemfb:
This PR includes fixes for the obvious, visually-identified double-counted references mentioned in #3657. It's a draft, as this code has no tests, I'm not very familiar with it. I'll revisit once I've sufficiently tested manually. (Unit tests for our floating point implementation would be very nice to have ...)


joemfb:
I haven't had a chance to write tests for this code. But this just fixes obviously broken refcounting protocols. I don't think it's worth holding this fix on a higher bar of correctness is established.

Fang- (approved):
Can confirm that did indeed look very broken. Much better now!

PR #3661, opened 2020-10-07
interface: various fixes for new groups layout 

liam-fitzgerald:
A host of changes to address design criticism and QA discoveries. Highlights:
- Virtualised, searchable contact list
- Return of new/join groups screen (now with 200% more indigo)
- Made invitePopover match figma design, including an email field for inviting with urbit ID
- iOS scrolling is now working
- No longer weird empty space at bottom of screen on iOS safari


pkg/interface/src/views/apps/launch/components/tiles/tile.js
```diff
@@ -1,30 +1,25 @@
...
     if (to) {
-      if (routeList.indexOf(to) === -1) {
-        childElement= (<Link to={to}>{childElement}</Link>);
-      } else {
-        childElement= (<a href={to}>{childElement}</a>);
-      }
-      
+      childElement= (<Link to={to}>{childElement}</Link>);
+    } else if (href) {
+      childElement= (<a href={href}>{childElement}</a>);
```
matildepark: Should preserve the old logic here. It’s for third party apps which need to be a href’d.

matildepark: Or are you just passing href for those separately

pkg/interface/src/views/apps/publish/components/NotebookRoutes.tsx
```diff
@@ -70,6 +71,9 @@ export function NotebookRoutes(
           const { noteId } = routeProps.match.params;
           const note = notebook?.notes[noteId];
           const noteUrl = relativePath(`/note/${noteId}`);
+          if(!note) {
+            return <Center height="100%"><LoadingSpinner /></Center>;
+          }
```
matildepark, line 76: The only thing in this commit is spinner behaviour for no note case? But the note never loads in, last I looked

pkg/interface/package.json
```diff
@@ -35,6 +35,7 @@
     "react-oembed-container": "^1.0.0",
     "react-router-dom": "^5.0.0",
     "react-virtuoso": "^0.20.0",
+    "react-visibility-sensor": "^5.1.1",
```
tylershuster, line 38: Is this needed?

matildepark (approved):
Tweaked a few regressions. In we go!

PR #3663, opened 2020-10-07
chat: fix and standardize font size

tylershuster:
This PR standardizes chat font-size at Indigo 1, resulting in the same size as livenet (16px / `f9`).

To do this, it rewrites many chat components in Indigo, but there are still a few utility classes left.

It also catches an uncaught `next-userspace` bug where `ChatWindow` would retain its current scroll position and unread status when switching between stations, presumably due to some upstream memoization.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -139,6 +139,7 @@ export default class ChatMessage extends Component<ChatMessageProps> {
         className={containerClass}
         style={style}
         data-number={msg.number}
+        mb={1}
       >
         {dayBreak && !isLastRead ? <DayBreak when={msg.when} /> : null}
         {renderSigil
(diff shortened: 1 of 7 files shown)


matildepark:
We have some sort of half-dipped sigils when they don't have a colour set (i.e. when we're using `mix-blend-mode`)

image

matildepark:
This also uses `black` when the spec is to use `gray` as before (see [tlon.io](https://tlon.io/)).

tylershuster:
You mean like the text seen here?

image
 
This looks...odd...

image

Also should note that the tlon.io homepage has spacing between messages of the same user, so ... is that canon now?

matildepark:
I don't know, bc 16px is also not in spec. @urcades @g-a-v-i-n can we make a call?

tylershuster:
Better but a bit squinty

image

g-a-v-i-n:
lets go with 12 for now.

tylershuster:
So...gray, 12, with or without gaps between messages by the same user?

matildepark:
With gaps as per the mock, presumably!

matildepark:
(Which means the chat input will have to also be 12)

matildepark:
conflict

tylershuster:
Alright, hope everybody's prescription glasses are up to date

matildepark:
clean up commits? you have one that just says 'merged'

matildepark:
Chonky button 
image

tylershuster:
oh lawd he expandin

matildepark (approved):
Thanks!

PR #1089, opened 2018-12-06
fix another type variance bug

joshuareagan:
This PR corresponds to: https://github.com/urbit/arvo/pull/914

It was a bit of an adventure figuring out how to get the jet working right, so please double-check the reference counting.  (I did check for leaks, and I believe there are none...)

If you have questions about why I chose this kind of fix I'm happy to answer them.

PR #3664, opened 2020-10-07
hoon: preserve callsite type on wet core in ++felt

### LINKED ISSUE #3600: Wet gate type definition used instead of callsite type, when addressed through $
This is a clumsy title. Feel free to reword if you can do better.

From https://github.com/urbit/urbit/pull/3442#discussion_r495496085:

```
> ? (|*(* +<) 1)
  @ud
1
> ? (|*(* +<.$) 1)
  *
1

> `@`(|*(* +<) 1)
1
> `@`(|*(* +<.$) 1)
nest-fail

> `@`(|*(@ +<.$) 1)
1
> `@`(|*(@ +<.$) [1 1])
<<unlinks-dojo>>
```

pkova:
Fixes #3600.

cc @pilfer-pandex @joemfb

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -9740,18 +9740,7 @@
     ^-  (pair type type)
     (~(mull et hyp rig) gol dox)
   ::
-  ++  felt
-    ~/  %felt
-    |=  lap/opal
-    ^-  type
-    ?-  -.lap
-      %&  p.lap
-      %|  %-  fork
-          %+  turn  ~(tap in q.lap)
-          |=  [a=type *]
-          ?>  ?=([%core *] a)
-          [%core q.q.a q.a]
-    ==
+  ++  felt  !!
   ::                                                    ::
   ++  feel                                              ::  detect existence
     |=  rot/(list wing)
@@ -9800,7 +9789,12 @@
       ==
     ::
         %&
-      =.  sut  (felt q.p.mor)
+      =.  sut
+        =*  lap  q.p.mor
+        ?-  -.lap
+          %&  p.lap
+          %|  (fork (turn ~(tap in q.lap) head))
+        ==
       =>  :_  +
           :*  axe=`axis`1
               lon=p.p.mor
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:4268cfff119e27f4d755eaf3a6571c1a6a712b29dc9d4a65930cd5dcf6468492
-size 6376654
+oid sha256:42a1cfce064fc42f930bf34fcb53478a186ff880e9c95ca9042a4ee27a8aa25d
+size 6286780


pkg/arvo/sys/hoon.hoon
```diff
@@ -9735,7 +9735,11 @@
           %+  turn  ~(tap in q.lap)
           |=  [a=type *]
           ?>  ?=([%core *] a)
-          [%core q.q.a q.a]
+          :+  %core
+            ?:  =(q.p.q.a %wet)
```
pilfer-pandex, line 9739: It's more idiomatic to express this as a ?- on q.p.q.a.

pilfer-pandex (commented):
Just as a quick sanity check, can you talk through what `feel` does, and why this change has the desired effect?

joemfb:
I'll need to think about this more before I'm prepared to comment. For anyone else who's doing the same, prior changes to this code are at urbit/arvo#914 and #1089, and the related discussion is at urbit/arvo#891.

pkova:
The way I debugged this was the following, the code examples are all from before my patch.

First, let's start with the examples from the issue:
```
=first (~(mint ut %noun) %noun (ream '(|*(* +<.$) 1)'))
=second (~(mint ut %noun) %noun (ream '(|*(* +<) 1)'))
```
The nock is the same obviously, but the types are different:
```
-:first
#t/*
-:second
#t/@ud
```
The types expanded are the following: 
<pre>
first:
[ %hold
  [ %core
    [%cell [%atom %ud 0] %noun]
    [0 %wet %gold]
    [%cell %noun %noun]
    [[%full 0] 0 6]
    [0 0 [0 %cnts <b>[[%.y 6] %$ 0]</b> 0] 0 0]
    0
    0
  ]
  %cnts
  <b>[[%.y 6] %$ 0]</b>
  0
]
second:
[ %hold
  [ %core
    [%cell [%atom %ud 0] %noun]
    [0 %wet %gold]
    [%cell %noun %noun]
    [[%full 0] 0 6]
    [0 0 [0 %cnts <b>[[%.y 6] 0]</b> 0] 0 0]
    0
    0
  ]
  %cnts
  <b>[[%.y 6] 0]</b>
  0
]
</pre>
That's some ghetto manual noun pretty printing but you should get the gist, I bolded the differences. Why are these types different?
```
=a `type`-:first
=b `type`-:second

?:  ?=(%hold -.a)  (~(mint ut p.a) %noun q.a)  a
[#t/* q=[%0 p=6]]
?:  ?=(%hold -.b)  (~(mint ut p.b) %noun q.b)  b
[#t/@ud q=[%0 p=6]]
```
Tracking this down led me to `++fond` where the subject `sut` eventually ends up as the type. The subject changes from `#t/<1.tyd [@ud *]>` to `#t/<1.tyd [* *]>` on this line: https://github.com/urbit/urbit/blob/30be2ce6af2bcc28b96ae0fcd01de63a6476adbd/pkg/arvo/sys/hoon.hoon#L9792
which is the only place `hoon.hoon` calls `++felt`.

In the buggy case the argument to `++felt` is 
```
[%.n p=2 q={[#t/<1.tyd [@ud *]> q=[%wet p=[%cnts p=~[[%.y p=6] %$] q=~]]]}]
```
and `a` in the `turn` becomes 
```
[ %core
  #t/[@ud *]
    q
  [ p=[p=~ q=%wet r=%gold]
    #t/[* *]
    r=[p=[mask=[%full blocks={}] data=[0 6]] q={[p=%$ q=[p=~ q={[p=%$ q=[%cnts p=~[[%.y p=6] %$] q=~]]}]]}]
  ]
]
```
Where the original code loses the `@ud` type information.

Now obviously this kind of reasoning does not allow me to be super confident in the correctness of my patch, which I probably should've been clear about from the start! But it does work and doesn't seem to break any stuff that I've tried.

joemfb (changes requested):
I can explain the context and history here.

`+felt` is just a helper for general wing resolution in `+fond`. It's used to apply the subject from the previous name-resolution step (ie, of the subsequent limb in the wing). `+fond` has an overly-complicated product type (`$pony`), but the relevant portion is a `$palo` or `(pair vein opal)`, where `$vein` is a subject-axis trace from resolving a wing and `$opal` is a (wrapper around) the type of the product of that wing. `+felt`'s only purpose is to convert the `$opal` to a `$type` for subsequent limb resolution.

`+felt` was rewritten in urbit/arvo#914 to fix a variance-enforcement bug. The old version called `+fire` (which, unsurprisingly, computes the product of firing an arm of a core), but this was not correct for name-resultion: `+fire` unconditionally sets the core variance to `%gold`, because an arm can read anything in the core's context. urbit/arvo#914 fixed that bug, but introduced a new one by substituting the actual and formal subjects. The discussion of that PR notes that this was to work around a compilation issue in %clay, but it also introduced the bug this PR is aiming to solve.

I spent time on this today, and have found and fixed the (very-longstanding) offending expression in %clay (which never should've compiled): #3689.

Noting that this won't work until that PR is in (you could base your changes on top), I believe the body of `+felt` should simply be:

```
|=  lap=opal
^-  type
?-  -.lap
  %&  p.lap
  %|  (fork (turn ~(tap in q.lap) head))
==
```

That's simple enough that it doesn't seem to merit a separate gate, so I'd be inclined to inline that expression in `+fond`:

```
=.  sut
  =*  lap  q.p.mor
  ?-  -.lap
    %&  p.lap
    %|  (fork (turn ~(tap in q.lap) head))
  ==
```

I'd stub out `+felt` (ie, `++  felt  !!`), but keep the arm in `|ut` to avoid messing with the hooks.

joemfb:
BTW @pkova, nice job tracking this down! I discovered #3600, but was honestly dreading the effort to figure out where the wrong core subject was being used.

joemfb:
The miscompilation documented in #3600 is fairly straightforward: the relationship between the formal payload of a wet core and the actual payload is never verified through a nest check. Instead, soundness of wet cores is enforced only through call-site type inference (ie, `+mull:ut`) -- it's never acceptable to access the subject of a wet core using the type of it's formal payload, as that type is never directly checked.

(I continue to suspect that this is too flexible, and that it should be possible to require that actual payloads of wet cores nest in their formal payloads. But this would at least require a change to the nest semantics for higher-ordered wet cores, since a wet core nests in nothing but itself. And I think @ohAitch may disagree with me ...)

pkova:
At the risk of stating something obvious (noticed #3691), the way I debugged this was just copy-pasting duplicates of all relevant arms in `hoon.hoon` and stuffing them full of print statements. Otherwise the printing breaks everything on the `hoon.hoon` recompile. So I didn't actually run `(~(mint ut %noun) %noun (ream '(|*(* +<.$) 1)'))`, I ran `(~(my-mint ut %noun) %noun (ream '(|*(* +<.$) 1)'))`, and was working with `++my-fond`, `++my-felt` and so on. Gotta love pure functions.

pkova:
Fixed and #3689 is in.

pilfer-pandex:
I've got to reload all the content around this into my head before I can properly sign off on it. Most likely will happen next week. @joemfb, can I assume by your request and the foregoing conversation that you're good with this if I am?

joemfb (approved):
Yeah, these changes LGTM.

joemfb:
@pilfer-pandex, bump for review.

philipcmonk:
Is this ready for merge?  @pilfer-pandex I think you were going to review?

PR #3668, opened 2020-10-07
launch: visual edits to tile borders and clock colors

urcades:
I made a few tiny edits to the colors used on the clock sun's outer border (to change it from blue to a low-opacity black), and lightened up the tile borders.


urcades:
Actually, I was observing the latest changes made to next-userspace and prefer some options made there, will base another similar PR off that one.

PR #3670, opened 2020-10-08
eyre: lower +channel-timeout

liam-fitzgerald:
Eyre does not reap active channels with an associated eventstream.
+channel-timeout is only used when we lose the eventstream, to allow for
a client to re-establish a subscription without having to resend updates
that had already been received. However, this feature is unused in landscape
and the length of the timeout has been causing bail:memes as of late. As such,
this commit changes +channel-timeout to reap more often to avoid potential
memory pressure issues

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:15a8b98d7eb45b168edfc1fe346bc79781a2e81a8d0b9fa7db87ef527b84c66b
-size 6319407
+oid sha256:24459ded4f143fae9e942d5b4096ed05ff65db7fffc1f97bb69c6f1a3601a28b
+size 6328344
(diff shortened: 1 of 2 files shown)


tylershuster:
This must be why I have a gig of channels

matildepark:
> If it doesn't, does that mean that, whenever I drive through a tunnel, my urbit re-initiates all subscriptions etc once it regains a connection?

IIRC, Yes — we clear and re-reduce the initial update.

eglaysher (changes requested):
I am against this for the reasons palfun said above.

A better workaround, assuming that this can't be fixed in landscape, is to add to eyre the same subscription backoff logic that ames uses, so that a subscription that has a large queue of unacked messages breaks the individual subscription, which should be a detectable state in the channel, so that the channel queues don't become so large in the first place.

tacryt-socryp:
I think this can be closed.

PR #3671, opened 2020-10-08
channel.js: eagerly ack subscription updates

liam-fitzgerald:
Changes channel.js to eagerly ack subscription updates. Also fires the delete action in beforeunload instead of unload, to buy us some more time.


Fang-:
(Not a showstopper but) wondering how many additional http requests this sends and what that does to perceived performance and event log growth. (I am once again asking: log truncation when? (; )

tacryt-socryp (approved):

PR #3674, opened 2020-10-08
hark: JSON encoders, marks, bugfixes

liam-fitzgerald:
- Adds JSON encoders/decodeers
- Fixes a bug in lib/metadata where paths would not be encoded properly
- Fixes a bug where the metadata helper door was not given the correct sample
- Adds action, update marks
- Adds a scry path to retrieve the `unread-mop` given an app-resource and group-resource
 
(Apologies for merging in release/next-userspace, was going to do UI stuff but ended up doing this to get it working)


matildepark:
> Fixes a bug in lib/metadata where paths would not be encoded properly
> Fixes a bug where the metadata helper door was not given the correct sample

👀

belisarius222 (approved):
LGTM.  Thank you for the various bugfixes.

PR #3677, opened 2020-10-08
groups: pass hideAvatars / nicknames to participants list

matildepark:
- Re-establishes avatars in the participants list if they are set
- Passes the hide props for both nicknames and avatars to respect user settings and avoid a regression


urcades (approved):

urcades:
Stamp!

PR #3678, opened 2020-10-08
launch: edited colors+values for clock, weather, and tile borders

urcades:
Made color value edits to a few launch elements, that's all.


matildepark (approved):
So you did

PR #3679, opened 2020-10-08
groups: remove remnant groups from recent

matildepark:
- We get recent groups from localStorage, but this produces edge cases where we leave a group, but retain it in our recentGroups array
- We don't filter this array for what groups we're currently in before rendering it, so "empty" recents are rendered
- This just filters that array with a safety check'd call inside the `associations.contacts` object.


urcades (approved):
Stamp

Issue #3645, opened 2020-10-05
bail meme: out of loom


Zaphod101010:
**Describe the bug**
Ship crashes with error 
```~hodler-pinryx:dojo> \«ames»
bail: meme
bailing out
pier: serf error: end of file
~hodler-pinryx:dojo>
address 0x458 out of loom!
loom: [0x200000000 : 0x280000000)
Assertion '0' failed in noun/events.c:129

bail: oops
bailing out
Assertion failed: (0), function u3m_bail, file noun/manage.c, line 677.
[1]    99384 abort      ./urbit hodler-pinryx
```

I used the Mac binary to run meld from: https://github.com/urbit/urbit/pull/3235#issuecomment-693215064

Output of meld:
```
a6a7b557a95cd246430ce68809d8188d40ae28f4-darwin/urbit-worker meld ./hodler-pinryx
loom: mapped 2048MB
boot: protected loom
live: loaded: GB/1.515.700.224
boot: installed 268 jets
serf: measuring memory:
  kernel: GB/1.309.939.032
total arvo stuff: GB/1.309.939.032
  warm jet state: KB/116.388
  cold jet state: MB/145.005.616
  hank cache: B/372
  battery hash cache: B/288
  call site cache: B/100
  hot jet state: KB/126.496
total jet stuff: MB/145.249.260
  bytecode programs: MB/1.078.368
  bytecode cache: KB/107.576
total nock stuff: MB/1.185.944
  trace buffer: B/36
  memoization cache: B/288
total road stuff: B/324
total marked: GB/1.456.374.560
free lists: MB/59.300.944
sweep: GB/1.456.374.560

hash-cons arena:
  root: B/144
  atoms (445039):
    refs: MB/10.284.580
    data: MB/164.433.182
    dict: MB/27.967.368
  total: MB/202.685.130
  cells (35463450):
    refs: MB/781.763.380
    dict: GB/3.439.758.872
  total: GB/4.221.522.252
total: GB/4.424.207.526

serf: measuring memory:
  kernel: MB/968.540.908
total arvo stuff: MB/968.540.908
  warm jet state: KB/84.400
  cold jet state: MB/56.481.148
  hank cache: B/288
  battery hash cache: B/288
  hot jet state: KB/124.408
total jet stuff: MB/56.690.532
  bytecode cache: B/288
total nock stuff: B/288
  memoization cache: B/288
total road stuff: B/288
total marked: GB/1.025.232.016
free lists: KB/150.220
[...]

PR #3680, opened 2020-10-08
eyre: implement a max-unacked-events counter and discard channels that hit it

tacryt-socryp:


matildepark:
(Does this mean that if we initialise enough subscriptions at the same time on FE, just the initial updates going unacked will auto-reap the channel, causing a loop when we reconnect?)

tacryt-socryp:
Not if we ack eagerly. Initial subscriptions don't actually happen "at the same time" from Arvo's perspective.

eglaysher:
This is closer than the other patch, but still isn't what I think should happen.

There's a thing in the gall/ames interface which closes a gall subscription if there are too many unacked packets, I believe its called `%clog`. What I requested on the other patch was to extend this from ames to eyre.

If a single subscription on a channel fills up with say, 50 messages, the desired behaviour is to notify the gall app that the subscription is dead, and then add a %quit message into the channel for this subscription. Now, if the client reappears, the client gets a normal %quit message on the subscription and can resubscribe, the channels themselves don't become huge, and there's less work being done on possibly gone channels in gall.

tacryt-socryp:
`+discard-channel` already cleans up all related `%gall` subscriptions. Why would we need anything more than this?

eglaysher:
I'm saying there's a difference between canceling a subscription and canceling a channel. I'm saying that you probably want to cancel individual subscriptions when they get overflowed instead of discarding the entire channel, and that you want to do this because a reconnection of a channel should see the quit and restablish the subscription where it left off.

ohAitch:
To put it in other words, if there's 9 subscriptions that are updating once every 5 minutes and 1 subscription with all your chat messages, going through a tunnel should keep the 9 alive even if the chat queue overflows. Instead of closing and reopening them repeatedly as you go through subsequent tunnels, most times not even delivering a message between subscription and unsubscription.

belisarius222:
I'd like to note for completeness that Ames %clog works on a per-ship basis. If any publications to a peer have five or more outgoing messages and the peer hasn't communicated with us at all for 30 seconds, then Ames sends Gall a %clog for that peer, and Gall kills all the subscriptions.

https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/gall.hoon#L1153

Maybe that should be done on a per-channel basis (per-flow in Ames) instead, so that Gall could kill an individual subscription. The current Ames %clog system is very much a minimum viable flow control system.

—
~rovnys-ricfer
https://urbit.org

tacryt-socryp:
I understand the argument that one would want to do this on a subscription basis rather than a channel basis, but I disagree. Acks are done on a channel basis, not a subscription basis. It seems that this implementation has parity with existing ames' per-ship clog tracking, which to me seems good enough to alleviate the problem we have been seeing.

belisarius222:
I just talked to Logan about this, and I think his approach makes sense.  By far the most common cause of the client failing to ack a subscription update is a break in connectivity, which will affect all subscriptions in the channel.  Maybe we'd be leaving a little bit of performance on the table if the client fails to ack for other reasons, such as slowness or a bug within some particular channel (like what @ohAitch described), but that doesn't seem like an important case to optimize for.

Ames's per-ship %clog system, while simple, does seem to work pretty well in practice as far as I can tell, and similar reasoning applies there.  So I think this is fine.

vvisigoth:
A big question is how coupled we want eyre to be to this implementation of landscape. I think we need to keep things as general as possible to make sure that others can build clients for urbit without reverse engineering Landscape. 

@philipcmonk can you sound off/make a decision here?

Fang-:
Since I was tagged for review on this (and only just noticed), I'll leave my two cents here real quick.

I side with @eglaysher, some reasons already discussed in #3670.  

Changing Eyre to be more lenient towards (channel.js's) bad client behavior, instead of fixing that behavior, seems to be working from the wrong direction. In a multi-client world, which we need to start acting like we already live in, a client that makes your ship bail-meme is a client you'd immediately switch out for a better one.

Doesn't #3671 fix the root cause here? @eglaysher's proposal here seems righteous, we'll still want to do it some day. But it might also not be _needed_ if we just ship the client-side fix. (In addition, perhaps, to my proposal of storing channel-ids in browser state and closing previously-used ones on-pageload, to work around on-unload logic being unreliable.)

joemfb:
I think it's very important that %eyre be robust against a miswritten client. Obviously, landscape should be as well-behaved as it can be. But if we're going to promote "airlock"-style architectures, these will problems will continually reoccur.

Having some kind of back-pressure handling in %eyre seems like the top priority, with how frequently we run into issues with channels filling up. I agree with @eglaysher that it would be best for that handling to be as precise as possible (per subscription, rather than per channel). But that's significantly more work than this PR. And, as @belisarius222 points out, %ames currently handles back-pressure with very low-granularity.

So I think we should move forward with this change, and schedule work in %ames/%eyre/%gall to replace the back-pressure implementations with a much more granular version.

belisarius222:
I don't understand.  What about this is landscape-specific?  My understanding is that each channel is a representation of a connection between client and server, with multiple subscriptions inside it.  Landscape does this, but so would any other client, right?  Unless a client is creating a new connection for each subscription, which is possible but not the way the API is intended to be used -- and in that case, limiting per-channel and limiting per-subscription have indistinguishable behavior.

The problem being addressed by limiting the number of unacked outgoing messages, whether per channel or per subscription, is that if a client fails to ack a message, that takes up space on the server until Eyre somehow knows that the data is no longer relevant, at which point it can delete the data.  Limiting the number of outgoing messages effectively puts an upper bound on the size of that space leak in Eyre.

There are two options for this limit: either per-channel, or per-subscription.  How do these compare?

Since each channel is a connection, a per-channel limit places an upper bound on Eyre's space usage for each connection.  When a connection dies, only a certain amount of data will be added across all the subscriptions within that connection before the limit is hit.

A per-subscription limit bounds the space usage for each subscription.  This means upper bound of the space leak for each connection is N*X, where X is the limit per subscription, and N is the number of subscriptions.

Which of these is more useful?  I think it depends on what you expect to cause unacked messages.  I think we can divide this into two categories: connection failures, and bugs.  Either the client can't send an ack because the connection is dead, or it doesn't send an ack because its logic is wrong, like the current situation.  Eyre needs to protect itself from client bugs (a broken client should have some difficulty causing a bail:meme), but it should optimize for the case of a working client.

A connection failure prevents acks on any subscription within the dead channel, so an upper bound on the whole channel is the clearest and most direct way to handle this failure.

As the case of bugs, some bugs might cause only certain subscriptions to fail, but others, like the current bad behavior where the client defers all its acks, affect all subscriptions within the channel.  If a bug causes just one subscription in the channel to stop acking, then the requirement for Eyre is just that it shouldn't bail:meme because of this.  A per-channel limit will fill up if one subscription is dead, preventing a bail:meme, so that should satisfy the requirement.

Is there any advantage from granular per-subscription limits?  I don't really see any.  I do see some advantages to per-channel limits: it's a simpler state with less supporting code, and it's a direct handling of the most common failure case -- a per-subscription limit would only indirectly handle a connection failure, by repeatedly limiting many individual subscriptions.

joemfb:
Simply put, re-establishing a subscription involves work; if some that work is unnecessary it would be better to not do that part.

Judging by my console output, %ames back-pressure notifications happen all the time. These involve a bunch of re-subscription events, which re-send a bunch of data I already have, making both peers do a bunch of work they shouldn't have to do. (Note also that mishandled initial responses from re-subscriptions seem to be the cause of the message duplication bugs in web-chat and the chat-store. So not only are both peers doing unnecessary work, but clients are storing extra copies of data, and, in the case of web-chat, displaying it.)

belisarius222:
> re-subscription events, which re-send a bunch of data I already have
This is interesting, and it suggests to me that the way we're thinking about resubscription is suboptimal.  If I disconnect for a while and lose a Gall subscription, when I resubscribe I should ideally only receive updates that happened after the peer killed my subscription -- data should not have to be re-sent, generally, other than the resubscribe messages themselves, which should be small.

That being said, Ames's %clog does seem like it might be a bit trigger-happy.  Notably, Ames should in theory actually be able to (roughly, and up to duplication) calculate the memory usage of its outbound messages, since they're jammed, and use that for the upper bound rather than number of messages.  The current limit is five messages, which for chat is usually less than five kilobytes -- not really worth killing subscriptions over.

I guess there is an advantage to per-subscription limits, which is clearer in Ames's case: ideally subscriptions don't interfere with one another, so if one subscription gets blocked (which can really only happen because of some kind of error on the subscriber, I think), that shouldn't prevent other subscriptions from making progress.

Since Urbit is single-threaded, I'm not sure how often this happens in practice.  As long as the subscriber sends at least one ack packet to the publisher every thirty seconds, the %clog won't get triggered, so even if one subscription is stuck, as long as any one subscription is making progress, no subscriptions will be killed.  I think almost every %clog is caused by a ship being unresponsive across all Ames flows, not from an individual Ames flow getting stuck.

On the other hand, if Ames killed per subscription rather than per ship, then you're definitely right that only that one subscription would need to be restarted, which would reduce the number of roundtrips.  And if resynchronization sends down repeat data, then that's even worse, since that has to be done on every subscription every time one of them dies.

Let's say we implemented per-subscription limits in Eyre now.  Since the client creates a new channel every time it reconnects, I think per-subscription limits would cause the same number of new subscriptions as per-channel, since the client would restart all its subscriptions anyway.  I don't think that should be the case forever, but I think it validates your suggestion do Eyre limits the easy way for now (per-channel).

tacryt-socryp:
It's worth noting that graph-store's `update-log` model opens up the possibility of sending very little duplicate data upon resubscribe from clients, by storing previous state (and the date of the last update) in local storage.

belisarius222:
Some longer-term thoughts about subscriptions and flow control:

I wonder if we're not making a sharp enough distinction between temporary connection loss and permanent data loss.  For example, in Ames, why do we need to send network messages to kill and restart Gall subscriptions when a connection flickers out and back on?  For any app to work properly, the subscriber needs to know how to resubscribe starting at an arbitrary point in the stream of updates from the publisher.  However, since there was no breach, both sides know which updates have been pushed down.  There's something very strange about requiring apps to manually resubscribe in this case, or even requiring network traffic about flow control in the first place.

For web clients, we also don't make this distinction.  I think most reconnects happen in browser tabs that still remember everything the server had told them in the previous connection, so reinitializing from scratch seems like unnecessary work.  I'm not really sure why a client tries to create a channel with a new id on reconnect.  If it remembers the previous channel id, it could use that, and then the server could reactivate the channel (if it had even deactivated it in the first place) and just push down the unacked messages without all these resubscribes.

Of course, web clients can come and go -- if a web client forgets its state, that's analogous to a breach: full initialization is required, and any total ordering guarantees are at least partially broken.  As long as all persistent state has the server as its source of truth, as it should, then this should reduce to a deoptimization, not a user-perceived data loss, with the exception of optimistic UI for latency compensation.  Basically, you might have to re-send any messages you've sent that haven't been confirmed by the server, which the UI generally handles properly now by keeping the message grayed out until the server has persisted it.

I think the alternative to requiring manual resubscription makes a subscription more like an event log, or maybe snapshot and event log.  When a ship first subscribes, the publisher would send it the current snapshot of the subscription state, or maybe a stale snapshot and all the updates since then.  Then the publisher streams new updates to the subscriber as they come in.  If outgoing updates start to back up on the server, 

This also suggests we might want two kinds of subscriptions: hysteretic and non-hysteretic -- path-dependent or not.  For a hysteretic subscription, a subscriber needs to operate on every update in order and can't skip any.  For a non-hysteretic subscription, all the client needs to know is the current state.  A contrived example of a hysteretic subscription client subscribes to a changing bit and counts the number of times it changes.  A non-hysteretic subscription client subscribes to a changing bit and just stores its latest value.

I think this distinction is related to the differences between Operational Transform and CRDTs, and we could similarly embed either system into the other.  Path-dependence could be built into a non-path-dependent subscription model by including every update into the current state (this is analogous to a CRDT); non-path-dependent subscriptions could be embedded into a path-dependent subscription model trivially, I think.  Each "diff" could either represent a change from the previous state, or it could be a full snapshot of the current state.

There is an asymmetry between these two approaches.  Publishers of hysteretic subscriptions can't do flow control using anything like the %clog system -- the publisher isn't allowed to forget any of the updates; publishers of non-hysteretic subscriptions only push down updates as an optimization; they could just push down a new snapshot every time.  So flow control is easier in this respect if the system assumes, as it does now, that subscriptions are non-hysteretic.

If the system assumed subscriptions were hysteretic, then I think it wouldn't need the concept of resubscription at all -- each update in a subscription could have a referentially transparent identifier (e.g. a permanent name for the subscription, then a sequence number), and a client would know it was up to date when it had retrieved (possibly through scry) every item in the subscription.

One thing we could do is support both kinds of subscriptions at the system level -- I think we'd want the client's subscribe request to include a hysteresis bit that would be required to match the bit the publisher would be required to provide when exposing the publication endpoint.  Not sure we'd gain anything from allowing the client and server to have different ideas about whether a publication is hysteretic.

Or maybe we continue to support only formally non-hysteretic subscriptions -- I'm not sure only supporting hysteretic subscriptions is a valid option, since there are plenty of cases where the publisher should not have to hold onto the full history forever.  I will note, though, that treating some non-hysteretic subscriptions as formally hysteretic could be more convenient -- chat comes to mind; a chat host shouldn't delete old chat messages, and graph-store already gives each message a referentially transparent identifier to make resynchronization easier and to better support scrying.

tacryt-socryp:
If we wish to remove the channel model within Eyre and logically have a single channel per subscription, then I think that reaping per subscription makes sense. Within Eyre's current model of having multiple subscriptions per channel, reaping per channel is the only solution that makes sense. This PR should go in as is, and discussion of rearchitecting Eyre should happen elsewhere.

philipcmonk:
> (Does this mean that if we initialise enough subscriptions at the same time on FE, just the initial updates going unacked will auto-reap the channel, causing a loop when we reconnect?)

This is something I'd be concerned about.  There's a very good reason Ames has the rule "5 messages AND 30 seconds" -- if you're sending a lot of events, but they're getting acked quickly, then you're fine to keep sending them, and in particular it's common during initial subscription to send many messages at once.  I don't know if landscape would run into this right now, but this happened with the old gall 20-message backpressure constantly, requiring everyone to write separate batch backlog messages.

I don't care much either way about cancelling the specific subscription vs the whole channel.  I don't have time to read the backlog on this tonight, I'll get back to it if you guys don't make a decision.

joemfb:
If %eyre's back-pressure trigger were similarly "100 events unacknowledged for ~m1" (or whatever), we could also reduce the frequency of eager client acks. The logic would be something like: send an ack if we've received messages that we haven't yet acked, with that function called via `setInterval()` every 10|20|30 seconds.

philipcmonk:
This seems like a useful discussion, but also this addresses a significant issue on the network that needs to be fixed quickly.  I think we should release this since its implementation is simpler than the alternative, and then revisit this later.  @eglaysher @Fang- Is that fine?  Can you review the implementation?

>  guess there is an advantage to per-subscription limits, which is clearer in Ames's case: ideally subscriptions don't interfere with one another, so if one subscription gets blocked (which can really only happen because of some kind of error on the subscriber, I think), that shouldn't prevent other subscriptions from making progress.

FWIW, I saw this happen months ago.  One flow gets stuck, so the other flows continually %clog.  I forget if we mitigated that somehow (eg only %clog on new messages and not on retries?) or just fixed the relevant subscriptions.

> If the system assumed subscriptions were hysteretic, then I think it wouldn't need the concept of resubscription at all -- each update in a subscription could have a referentially transparent identifier (e.g. a permanent name for the subscription, then a sequence number), and a client would know it was up to date when it had retrieved (possibly through scry) every item in the subscription.

The biggest change this implies is that you no longer "send a bunch of updates".  Instead, the subscriber requests updates one at a time, and sometimes it takes longer to fetch because the update doesn't exist yet.

Fang-:
> this addresses a significant issue on the network that needs to be fixed quickly. I think we should release this since its implementation is simpler than the alternative, and then revisit this later. @eglaysher @Fang- Is that fine?

I don't think this is a significant issue anymore, as of #3671. That client-side fix doesn't make Eyre robust against "bad" clients, but at least takes the pressure off, giving us time to do the right thing here.

...Which I still believe is the per-subscription variation on this. The primary point of Eyre's channel design is that it allows reconnecting gracefully after connectivity loss. If you nuke channels when they aren't getting acked for a minute or so (100 messages in a couple minutes seems pretty plausible), you're forced to re-establish all your subscriptions/data from scratch again. The proposed alternative just gives you a `%quit`, like you would get in a gall context, letting you recover much more cleanly/easily.

tacryt-socryp:
#3671 is still a significant issue for anyone who refreshes frequently, as those channels may not be properly cleaned up by the unload event to delete the channel. Those channels will still be a significant space leak.

Fang-:
@tacryt-socryp Quoting from above, and as brought up in an interface meeting earlier, there's a client-side fix for this that should be implemented regardless of what happens in this PR.

> (In addition, perhaps, to my proposal of storing channel-ids in browser state and closing previously-used ones on-pageload, to work around on-unload logic being unreliable.)

tacryt-socryp:
The server-side fix makes the client-side fix unnecessary, as we will be robust against poorly behaving clients.

philipcmonk:
> I don't think this is a significant issue anymore, as of #3671.

Hmm, I feel like I've still been seeing people complain about large eyre memory usage.  @joemfb You've been handling those cases I think, can you confirm either way?

joemfb:
I don't have hard numbers, but we get a lot of `bail:meme` reports, and I think the root cause is almost always this -- #3740 is a recent example, https://github.com/urbit/urbit/issues/3645#issuecomment-709631355 has many more. (The only other active cause I'm aware of is `|autocommit` problems, as in #3317.)

I don't think it's a good idea to rely on a well-behaved client. I'm especially wary of more sophisticated browser implementations, like "beforeunload()" hooks or recovering channels to close from local storage. I imagine extensions or local configuration could interfere with either or both.

It seems like it would be best to add some kind of timer to the change in this PR. If that's too complicated to release quickly, I'd be in favor of releasing this as is. Or maybe with a slightly higher limit -- do we know a) how many initial updates we tend to send per subscription , and b) how many subscriptions we tend to open per channel?.

tacryt-socryp:
We tend to send one initial response per subscription, and to open 5-10 per channel immediately upon channel instantiation. (It's now ~10, with graph-store, it will be reduced to ~5)

Fang-:
Superseded by #3749.

PR #3681, opened 2020-10-08
leap: add 'home' to leap index, default results

matildepark:
- Adds 'home' (our DM / personal space) to the leap index + default results when nothing is searched

cc @urcades @g-a-v-i-n — 

Design-wise I use "Circle" for the icon, not "Home", because we use "Circle" to denote the home space in the tile; and we use the "Home" icon to ...  bring us to the launch screen. I think this is gonna need smoothing at some point, but I just wanted to be consistent.

image

### DIFF
--- a/pkg/interface/src/logic/lib/omnibox.js
+++ b/pkg/interface/src/logic/lib/omnibox.js
@@ -30,6 +30,7 @@ const commandIndex = function (currentGroup) {
 
 const otherIndex = function() {
   const other = [];
+  other.push(result('Home', '/~landscape/home', 'home', null));
   other.push(result('Profile and Settings', '/~profile/identity', 'profile', null));
   other.push(result('Log Out', '/~/logout', 'logout', null));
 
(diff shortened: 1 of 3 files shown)


urcades:
Yeah, the icon is actually a really good question, kind of up in the air for me — I felt that the circle was a fine temp/placeholder representation because circles have always embodied the little urbit diagram, but Home could probably be a little more legible/distinguished than a circle. Not sure atm though.

matildepark:
We just don't want the 'home' icon to go to two different places, is the thing

urcades:
Basically, I'd stamp this for the pure inclusion, but if @g-a-v-i-n has a good idea/an existing thought for how to represent the "home/unmanaged group", we should just use w/e he has in mind.

urcades:
> We just don't want the 'home' icon to go to two different places, is the thing

Yeah, totally understand

urcades:
Stamp

urcades (approved):

PR #3684, opened 2020-10-08
-graph-leave: fix action type

belisarius222:


matildepark (approved):
Thank you both

philipcmonk:
Without having more context, I will note that threads commonly expect a trailing `~` since this allows you to run them from the dojo directly

tacryt-socryp:
A quirk of spider's HTTP interface, one that likely should be fixed, is that it hands a singular vase to the thread it invokes, not a list of vases.

PR #3687, opened 2020-10-08
leap: restore app index, remove from default

matildepark:
Fixes a temporary regression where apps were no longer indexed in Leap; I believe it also fixes an unfiled bug where third party apps crash Landscape due to, what I can only assume, is new behaviour in indigo-react when no icon string is specified (it used to have a default?).

Doesn't list the apps in the default results list (we'll probably want "Recent Groups" here but requires some additional logic); also uses the new Dojo icon.


liam-fitzgerald (approved):
LGTM!

PR #3688, opened 2020-10-09
u3: run gc after each unit test suite

joemfb:
This PR enables garbage-collection in all of our runtime unit suites (except for `noun_tests`, which would need a ton of cleanup). It also removes redundant tests, uses a more consistent testing protocol (continue testing after failure, return results), and generally cleans things up.

### DIFF
--- a/pkg/urbit/include/jets/k.h
+++ b/pkg/urbit/include/jets/k.h
@@ -5,6 +5,7 @@
   /** Tier 1.
   **/
     u3_noun u3ka_add(u3_noun a, u3_noun b);
+    u3_noun u3ka_dec(u3_atom a);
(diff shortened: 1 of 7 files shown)


Fang- (approved):
This seems fine. The indirection/mismatch between urbit-style (0 is failure) and unix-style (0 is success) status codes is a little bit confusing at first glance, but probably consistent with the entire rest of the codebase. Is there a particular reason to not just use `c3_o` though, besides an abstract desire to support more than two status codes or w/e?

joemfb:
My original reason for using `int` in the tests was `&=`, but we honestly overuse loobeans, probably due to the cognitive cots of switching back and forth. At some point, I'll probably bite the `stdbool.h` bullet.

PR #3689, opened 2020-10-09
clay: refactors +cancel-request, consistently accounting for TMI

joemfb:
This PR fixes a long-standing miscompilation in %clay, introduced in 8f66b6ff5f2c864f3477e4b029bc7d0dc472a0a7, found by experimenting with #3664, first (obliquely) reported in https://github.com/urbit/arvo/pull/914#discussion_r239579011.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -2654,25 +2654,28 @@
   ::  we remove it from `ref` and tell the foreign ship to cancel as well.
   ::
   ++  cancel-request                                    ::  release request
-    ^+  .
+    ^+  ..cancel-request
     =^  wos/(list wove)  qyx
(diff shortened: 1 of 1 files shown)


pilfer-pandex:
Can you help me understand what the miscompilation was, why it was a miscompilation, and how the fix works?

As you know, I have a longstanding interest in actually understanding wetness.

joemfb:
"Miscompilation" is probably hyperbole, and the title of this PR was wrong. By the time I opened it, I had convinced myself that the subject axes were wrong in the `?=(~ ref)` branch, but that's actually no the case.

#3664 fixes a mis-compilation bug in wet-gate subject access, but there are no wet cores involved in this code. The problem was simply that the specialized subject type was not being consistently accounted for in each branch (ie, "too-much-information"), but the compiler was not enforcing this due to subject access occurring through a core (and using the formal payload rather than the actual payload).

philipcmonk (approved):
LGTM

PR #3690, opened 2020-10-09
eyre: explicitly permit proposed request in cors

### LINKED ISSUE #3676: cors-registry: send allowed methods and headers in preflight
When authenticated, PUTting to a channel from an allowed origin results in this response:

```
Access to XMLHttpRequest at 'http://localhost:8080/~/channel/1602164113-8cb5f1' from origin 'http://localhost:1701' has been blocked by CORS policy: Method PUT is not allowed by Access-Control-Allow-Methods in preflight response. 
```

Subscribing to that channel yields this one:

```
Access to fetch at 'http://localhost:8080/~/channel/1602164113-8cb5f1' from origin 'http://localhost:1701' has been blocked by CORS policy: Request header field cache-control is not allowed by Access-Control-Allow-Headers in preflight response. 
```

These are both set here:

https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/eyre.hoon#L597-L598

[According to MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Methods), `*` is not allowed as a response for authenticated responses.

Fang-:
Instead of always providing a wildcard for the allowed methods and headers, now echoes back the method and headers that the client said it's going to use, if any.

@tylershuster can you give this a try and confirm it fixes #3676?

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -558,12 +558,13 @@
   ++  request
     |=  [secure=? =address =request:http]
     ^-  [(list move) server-state]
+    =*  headers  header-list.request
     ::  for requests from localhost, respect the "forwarded" header
     ::
     =?  address  =([%ipv4 .127.0.0.1] address)
-      (fall (forwarded-for header-list.request) address)
+      (fall (forwarded-for headers) address)
     ::
-    =/  host  (get-header:http 'host' header-list.request)
+    =/  host  (get-header:http 'host' headers)
     =/  [=action suburl=@t]
       (get-action-for-binding host url.request)
     ::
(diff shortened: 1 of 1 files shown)


eglaysher (approved):

tylershuster:
Yes uh

How do you test from next-sys? I've only ever accidentally done it. I'm familiar with the parts but I seen unable to put them together this morn

Fang-:
@tylershuster checkout this branch, then `urbit -F dev -A pkg/arvo -B bin/solid.pill`. Probably run `|reset` after booting because the pill wasn't updated with this change. And then just, send some requests around.

tylershuster:
Works like a dream!

PR #3691, opened 2020-10-09
hoon: enables need/have type printfs on nest-fail

joemfb:
I think their time has come. While investigating #3664 and developing #3689, I found that I had to enable these to make progress. Even though I was printing absolutely massive types, performance was not an issue. There's still much to be done to improve our type printing, but I think it's worth enabling these regardless.

This PR also removes the "road:new" stack-trace message on crashes in `+road` -- I don't think that conveys anything useful to users or developers, as `+road` is a general-purpose virtualization tool with no specific semantics. (Furthermore, its entire purpose is to behave as if the computation were not virtualized, modulo the allocator properties of the runtime. So it really shouldn't be observable, and should probably just be a hint in the future.)

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -12060,7 +12060,7 @@
   =/  res  (mule trap)
   ?-  -.res
     %&  p.res
-    %|  (mean leaf+"road: new" p.res)
+    %|  (mean p.res)
   ==
 ::
 ++  slew                                                ::  get axis in vase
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
LGTM.  I agree that `+road` should eventually just be a hint, so printing there is a mistake.

ohAitch:
Oh hey, I remember this being one of the main things in my local dev patchset! If perf is good and we don't like swallow any of these in expected crashes, this is absolutely correct.

PR #3692, opened 2020-10-09
interface: show unmanaged invites in home workspace

matildepark:
There's no way to actually accept a DM at the moment. This PR iterates through non-group invites in the home workspace (since it's the only case that still uses the invites-only system).

Design review since I rewrote the component in indigo-react @urcades @g-a-v-i-n — 

Screen Shot 2020-10-09 at 2 34 33 PM


tacryt-socryp (approved):

matildepark:
Merging early as per @vvisigoth

PR #3693, opened 2020-10-09
vere: updates ames to only print network send failures once

joemfb:
This resolves an irritating issue wherein the ames i/o driver spams error messages if the network goes down, the machine goes to sleep, &c.


Fang- (approved):
Looks good!

PR #3696, opened 2020-10-13
interface: various bundle size optimisations

liam-fitzgerald:
This PR makes some changes to landscape's build configuration in order to decrease bundle size. Total reduction is from 10MB to 2MB. Have no hard numbers, but this should improve landscape's responsiveness across the board. 

##### disable source maps on production builds

Source maps have never worked correctly on production builds, and the
way that they are bundled into the JS itself negatively impacts our time
to first paint. As such, we disable them for a significant bundle size
improvement

##### do not bundle extra locales

Landscape is not localised in any fashion and we never change the
default locale. As such, the locales that moment.js includes in the
bundle are redundant and can be removed

#####  rewrite lodash imports with babel
Due to lodash's packaging mechanism, it is unable to treeshake unless
you always import from it in a certain fashion. This commit adds a babel
plugin to rewrite these imports for us.

### DIFF
--- a/pkg/interface/src/views/apps/chat/ChatResource.tsx
+++ b/pkg/interface/src/views/apps/chat/ChatResource.tsx
@@ -1,6 +1,7 @@
 import React, { useRef, useCallback } from "react";
 import { RouteComponentProps } from "react-router-dom";
 import { Col } from "@tlon/indigo-react";
+import _ from 'lodash';
 
 import { Association } from "~/types/metadata-update";
 import { StoreState } from "~/logic/store/type";
--- a/pkg/arvo/app/glob.hoon
+++ b/pkg/arvo/app/glob.hoon
@@ -85,20 +85,33 @@
       %glob-make
     :_  this
     =/  home=path  /(scot %p our.bowl)/home/(scot %da now.bowl)
-    =+  .^(=tube:clay %cc (weld home /js/mime))
+    =+  .^(=js=tube:clay %cc (weld home /js/mime))
+    =+  .^(=map=tube:clay %cc (weld home /map/mime))
     =+  .^(arch %cy (weld home /app/landscape/js/bundle))
-    =/  bundle=path
-      %-  need
-      ^-  (unit path)
+    =/  bundle-hash=@t
+      %-  need 
+      ^-  (unit @t)
       %-  ~(rep by dir)
-      |=  [[file=@t ~] out=(unit path)]
+      |=  [[file=@t ~] out=(unit @t)]
       ?^  out  out
-      ?.  =((end 3 5 file) 'index')
-        ~
-      `/[file]/js
-    =+  .^(js=@t %cx :(weld home /app/landscape/js/bundle bundle))
-    =+  !<(=mime (tube !>(js)))
-    =/  =glob:glob  (~(put by *glob:glob) bundle mime)
+      ?.  ?&  =((end 3 6 file) 'index.')
+              !=('sj.' (end 3 3 (swp 3 file)))
+          ==
+        out
+      ``@t`(rsh 3 6 file)
+    =/  js-name
+      (cat 3 'index.' bundle-hash)
+    =/  map-name
+      (cat 3 js-name '.js')
+    =+  .^(js=@t %cx :(weld home /app/landscape/js/bundle /[js-name]/js))
+    =+  .^(map=@t %cx :(weld home /app/landscape/js/bundle /[map-name]/map))
+    =+  !<(=js=mime (js-tube !>(js)))
+    =+  !<(=map=mime (map-tube !>(map)))
+    =/  =glob:glob
+      %-  ~(gas by *glob:glob)
+      :~  /[js-name]/js^js-mime
+          /[map-name]/map^map-mime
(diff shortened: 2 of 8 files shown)


matildepark:
We should really be in a production mindset where end-users shouldn't expect to hit bugs at all; and all new and recent features are compiled as 'dev' environment to opted-in users. (And I guess as it stands, the pre-existing bug reports don't even make use of the source maps?) Having a fast, 2mb blob for the average person seems forward-thinking to me.

ohAitch:
Source maps are useful, source maps _in the same file_ are maybe a bit much. Perhaps host them on S3 or sth since they're not part of the cryptographic chain of custody in the same way code you actually run is? With a flag on the server for whether to even include them in the sail.

tylershuster:
It seems to me that Urbit is going to need localization eventually but as long as it's English only, removing other languages seems sane until we can take a principled approach.

I agree with removing source maps, with some caveats. It would be nice if there was a flag to "use source maps" so we could get post-mortems on crashes as Ted mentions. The system being in its infancy as it is, we more or less expect crashes to occur. However, having the dev channel for those intrepid enough to test for bugs should buffer that. I wonder if we shouldn't hold off on removing source maps until the dev channel pattern manifests. I would like to see sane stack traces in the Error Boundary.

liam-fitzgerald:
I agree about the desirability of source maps in production, however they should be external and hosted on some S3 bucket. This realistically should happen in CI, so I've punted doing so until we have that. 

Considering the dev channel concerns raised here it might be worth discussing ways we can get better stack traces and richer debug info from users. Perhaps we could put a github icon in the status bar (dev channel only ofc) and allow users to click that to log an issue. We could even serialise relevant state and take a screenshot. This would come in a .zip to be attached to the issue which would allow users to post an issue without being hassled by devs for more info.

matildepark:
> Considering the dev channel concerns raised here it might be worth discussing ways we can get better stack traces and richer debug info from users. Perhaps we could put a github icon in the status bar (dev channel only ofc) and allow users to click that to log an issue. We could even serialise relevant state and take a screenshot. This would come in a .zip to be attached to the issue which would allow users to post an issue without being hassled by devs for more info.

😍 This is actually a great idea because we can get whether this is dev or prod easily. (Though we also hook into that for HMR build?)

galenwp:
I would *love* something like an auto-issue link to exist in Landscape. 

But also, issues require github accounts. For some people, this might not be something they have. Could be interesting to use a notebook / chat channel.

matildepark:
In the case where people are bleeding edge testing something that might itself be broken is that wise? We could just pipe a specific QA email address to an issue for the "opt-in to our bleeding edge stuff but doesn't have Github" case.

matildepark:
This is in the weeds as it is, though. Going to get some quick check-ins on our approach for this in the upcoming meeting but I have a decent idea where we're headed, will update after.

liam-fitzgerald:
Have rigged the source maps to be served externally, such that the browser will fetch them in order to display an error or somesuch. As a bonus, this fixes sourcemaps in production. 

cc: @philipcmonk made some changes to glob, do they look ok?

pkg/interface/src/views/apps/chat/ChatResource.tsx
```diff
@@ -1,6 +1,7 @@
 import React, { useRef, useCallback } from "react";
 import { RouteComponentProps } from "react-router-dom";
 import { Col } from "@tlon/indigo-react";
+import _ from 'lodash';
```
matildepark, line 4: ???

liam-fitzgerald: Oh this import was missing

matildepark:
(I forgot to update on this.)

After the Landscape meeting today: Basically we believed that even if source maps weren't working in prod, we haven't gotten to a point where we outgrew them specifically because we don't have the dev release cycle yet; so we wanted to just leave this open until it was necessary.

But: as @liam-fitzgerald has just pushed a fix for production here that includes the source maps out of the bundle, we can merge this in now. Pending @philipcmonk on glob I think.

pkg/interface/src/views/apps/chat/ChatResource.tsx
matildepark: So `clearUnsent` is going to fail on livenet atm?

liam-fitzgerald: It doesn't, for reasons I do not understand

matildepark: lodash is so smart

tylershuster: Well, `_` is available globally for whatever reason

liam-fitzgerald: oh lmao this is probably the lodash import rewriter breaking things, but we really shouldn't be using UMD globals ever

philipcmonk (approved):
These glob changes look fine

matildepark (approved):
Approved since we now retain source maps for production errors

PR #3698, opened 2020-10-13
Fix zigzag in --stderr logging

pilfer-pandex:
Add the following to `pier` in Vere/Pier.hs:

```
void $ acquireWorker "crazy log tester" $ forever $ do
    logError "wow amazing wow"
    threadDelay 200_000
```

and run the king with --stderr. Before this change, the output will look something like

```
~zod:chat-cli: wow amazing wow
                              wow amazing wow
                                             wow amazing wow
                                                            wow amazing wow
```

and after it looks like

```
~zod:chat-cli: wow amazing wow
wow amazing wow
wow amazing wow
wow amazing wow
```

This should make the more verbose logging modes substantially more usable and less scary.

Note: It's still an open question in my mind whether this is the thing we should be doing, or if we should instead be sending "stderr" logs to the terminal multiplexing system. Even if the latter, this is such a win in the short term that I think it should be accepted.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/King/App.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/App.hs
@@ -29,6 +29,7 @@ where
 import Urbit.King.Config
 import Urbit.Prelude
 
+import RIO                    (logGeneric)
 import System.Directory       ( createDirectoryIfMissing
                               , getXdgDirectory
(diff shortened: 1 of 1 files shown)


eglaysher (approved):
I suspect that we should be sending stderrs to the terminal muxing system, but this is straightforward better in the short term.

PR #3699, opened 2020-10-13
groups: properly direct mobile navigation in home

matildepark:
- Mobile navigation out of a DM presents a blank sidebar at the moment
- We check for the home URL (probably a better workspace utility for this?) and pass workspace instead of selected group for constructing our back links to remedy this issue


liam-fitzgerald (approved):
LGTM!

PR #3701, opened 2020-10-14
u3: re-re-enable +repn jet

joemfb:
This just backports #2899 from `na-release/next-vere`.


eglaysher (approved):
I approve my own code!

PR #3703, opened 2020-10-14
u3: add more jets for list and treap operations

joemfb:
This PR is straight backport of #2895 from `na-release/next-vere` (the hoon changes are also present in #3702, targeting `release/next-sys` for easy OTA release).

This PR is currently a draft, due to the issues fixed in #2916 (which cannot be simply backported). I intend to PR any necessary changes into this branch (hopefully fixing the underlying bug, but re-disabling the treap-verification jets if that is not practical), at which point this can come out of draft status.

PR #3712, opened 2020-10-14
u3: cleanup new treap jets, rewrite and enable old

### LINKED ISSUE #328: {int,uni}:{in,by} should be jetted
There are some jets for these that "worked" at some point, but they have a couple of issues.  Firstly, they may have various memory leaks, which is obviously unacceptable.  Secondly, I'm not sure if the correct hashes are used in the right places -- specifically hor vs gor.

joemfb:
This PR augments #3703 with bug fixes, efficiency improvements, and style consistency. It also fixes #328, with rewrites of the `+int:in` and `+int:by` jets.

efficiency:

- `+zing` now cons's lists in-order with `u3i_defcons()` (as in #3658)
- `+find` no longer uses unnecessary arbitrary-precision increment (as in #3666)
- a handful of `|by` map jets are now tail-recursive in their handling of the right subtree

bugs fixes:

- `+find` was wrong (consider `(find "bcd" "abc")`)
- `+apt:in` and `+apt:by` did not properly preserve unchanged left/right parents when recursing.
- `+int:in` had incorrect refcount handling
- `+int:by` was just wrong, and also had incorrect refcount handling

While rewriting `+int:in`/`+int:by`, I adhered closely to the hoon implementation. This was somewhat frustrating, as they mutually differ unnecessarily, and each has their own redundancies: `+int:in` has  redundant conditional branches when it recurs with swapped arguments, while `+int:by` repeats equality checks and produces identical results in both `+mor`-guarded branches. It would be nice to clean those up, but, for now, I think it'll be easier to maintain these jets if they map directly to the hoon.


eglaysher (approved):
Nice. Sorry about the jet I wrote which was wrong.

joemfb:
I've confirmed that I can boot fake stars under fake galaxies (and with jet equivalence testing enabled), so the issues of #2916 are resolved. I've added a regression to test for `+find`, but have not been able to come up with appropriate input for `+apt:in` or `+apt:by`.

PR #3714, opened 2020-10-14
interface: design feedback tweaks

matildepark:
- Fixes the chat input top border being cut off on dark mode Safari; rewrites the containers in indigo-react.
- Properly pins the unread notice; truncates the text for small browsers (urbit/landscape#106).
Screen Shot 2020-10-14 at 6 05 54 PM

- Because we don't have a new DM route to handle the logic, just removes 'send message' from Participants view for now. (We need a utility to redirect and/or create a DM.)
- Stubs out the email invite prompt as per @vvisigoth.
- Fixes the styling of the channel dropdown in dark mode (urbit/landscape#109).
- Removes the secondary, 'recent groups' sort from the launch screen (urbit/landscape#111).
- Fixes an unfiled bug where the box-shadow was set to `rgba(0,0,0,0.1)` producing a weird, dark border in dark mode. We just pass the theme colour and inherit it, as we do elsewhere.

cc @urcades

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
@@ -7,6 +7,7 @@ import tokenizeMessage, { isUrl } from '~/logic/lib/tokenizeMessage';
 import GlobalApi from '~/logic/api/global';
 import { Envelope } from '~/types/chat-update';
 import { Contacts, S3Configuration } from '~/types';
+import { Row } from '@tlon/indigo-react';
 
 interface ChatInputProps {
   api: GlobalApi;
@@ -187,12 +188,16 @@ export default class ChatInput extends Component<ChatInputProps, ChatInputState>
         />;
 
     return (
-      <div className={
-             "cf items-center flex black white-d bt b--gray4 b--gray1-d bg-white"  +
-             "bg-gray0-d relative"
-           }
-           style={{ flexGrow: 1 }}
-           >
+      <Row
+        alignItems='center'
+        position='relative'
+        flexGrow='1'
+        flexShrink='0'
+        borderTop='1'
+        borderTopColor='washedGray'
+        backgroundColor='white'
+        className='cf'
+      >
         <div className="pa2 flex items-center">
           {avatar}
         </div>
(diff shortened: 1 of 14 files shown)


matildepark:
lol noticing some weirdness even in that screenshot with the name + timestamp overflow, might append another design tweak commit later

urcades:
Question — which branch is the currently-public OTA being sourced from? It's still master, yes?

matildepark:
Yes.

pkg/interface/src/views/apps/chat/components/ChatInput.tsx
```diff
@@ -187,12 +188,16 @@ export default class ChatInput extends Component<ChatInputProps, ChatInputState>
...
     return (
-      <div className={
-             "cf items-center flex black white-d bt b--gray4 b--gray1-d bg-white"  +
-             "bg-gray0-d relative"
-           }
-           style={{ flexGrow: 1 }}
-           >
+      <Row
+        alignItems='center'
+        position='relative'
```
liam-fitzgerald, line 193: `position='relative'` is the default now on block level indigo components I believe

liam-fitzgerald (approved):
LGTM!

urcades:
Quick question: Are you working with text-fixes in this PR? I just checked it out and it looks like we're using 14px and black text ... which is wonderful lol

urcades:
I'm just making double sure that this is in fact getting OTA'd soon, a little tired of the complaints, just need to get the fixes in, etc.

matildepark:
> Are you working with text-fixes in this PR? I just checked it out and it looks like we're using 14px and black text ... which is wonderful lol

I don't believe I changed the chat message colour, so weird. I can tho if you want.

This is probably going out tomorrow-ish.

urcades:
Yes! That's be lovely — i have no idea why my checked out branch is showing this edit, my vscode has been a bit wonky these last few days.

urcades:
Another thing I'm just noticing now: 

Screen Shot 2020-10-14 at 6 58 46 PM

Is this a simple z-index fix we can ship in this PR as well? If it's more complicated, I can roll it up into another issue, no sweat.

matildepark:
It's not complicated, I just forgot lol I kept trying to replicate

PR #3719, opened 2020-10-15
link-store: throw away state if it's stale cards

philipcmonk:
It's burdensome to upgrade cards and vases, and in practice they will be stale anyway, so we throw them away.

Per https://github.com/urbit/urbit/pull/3544#issuecomment-708880202

### DIFF
--- a/pkg/arvo/app/link-store.hoon
+++ b/pkg/arvo/app/link-store.hoon
@@ -4,7 +4,9 @@
 /+  store=link-store, graph-store, default-agent, verb, dbug
 ::
 |%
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
LGTM.

This situation is regrettable.  If we weren't just turning this app into a potato, I think we might want to add a pupal phase to hold the cards as an extra piece of state in a pupal wrapper core, until the agent can molt back into its adult form.  We'd want to use a narrow type for the cards, with no vases or reference to `$card:agent:gall`.

PR #3722, opened 2020-10-15
landscape: check workspace in channel menu

matildepark:
Fixes an unfiled bug via @galenwp where accessing channel settings (or deleting the channel) in the home workspace redirected back to the hidden group path instead of the home workspace, wiping the sidebar content.


liam-fitzgerald (approved):
LGTM!

PR #3728, opened 2020-10-15
Revert "king: fix zig-zag in stderr logging"

eglaysher:
This reverts commit e9ed97f8a82fb43992f8d1a16503e4565ea1b040.

This breaks slogs; we are constantly overwriting the last line in our boot output. I suspect that you really will have to send the --stderr messages to the terminal system.


pilfer-pandex (approved):
Sadface. Thanks for catching, E.

PR #3734, opened 2020-10-16
interface: use icon sigils for <16px renders

matildepark:
As per mockups. Tagging designers since the changes are 'passing a prop', but when and where we should use the icons is the question here.

Screenshot 2020-10-16 at 11 51 36 AM

### DIFF
--- a/pkg/interface/src/logic/lib/sigil.js
+++ b/pkg/interface/src/logic/lib/sigil.js
@@ -21,12 +21,14 @@ export const Sigil = memo(({ classes = '', color, foreground = '', ship, size, s
   return ship.length > 14
     ? (<Box
         backgroundColor='black'
+        borderRadius={icon ? '1' : '0'}
         display='inline-block'
         height={size}
         width={size}
        />) : (
        <Box
         display='inline-block'
+        borderRadius={icon ? '1' : '0'}
         flexBasis={size}
         backgroundColor={color}
         padding={padding}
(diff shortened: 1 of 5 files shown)


g-a-v-i-n:
Nice. There should be 2px padding around the glyph, with a 16px width and height. 

Screen Shot 2020-10-19 at 10 33 14 AM

matildepark:
Padded!

image

matildepark:
Should this button have a smaller sigil anyway? 
image

urcades:
the simplified sigils look beautiful in a chat where they are widely customized, like in urbit-community. I think per the indigo spec, we should give them a 1px radius in the chat view.

In the designs created thus far, the top bar sigil is fully 16px square, with no background color — the sigil itself is edge-to-edge 16px, but still iconified. Example/comparison here:

Screen Shot 2020-10-19 at 3 21 07 PM

Screen Shot 2020-10-19 at 3 21 21 PM

urcades:
Another comparison here:

Screen Shot 2020-10-19 at 3 22 49 PM

matildepark:
OK, can change some implementation stuff ... seems that the top bar sigil colour should be a big faded too.

urcades:
it should be using `gray` I believe

matildepark:
OK, addressed the padding and unpadded the statusbar sigil, set it to 16px. The partially transparent colour is a bit beyond me for the moment — unsure how to ensure the sigil render inherits our indigo-react theme instead of, well, css variable `gray`. I think this is ok to merge for a pass.

urcades (approved):

PR #3736, opened 2020-10-16
interface: enforce border radius

matildepark:
image

image

Addresses urbit/landscape#114.

cc @urcades


liam-fitzgerald (approved):
LGTM!

PR #3737, opened 2020-10-16
publish: prevent post input overflowing layout on safari

matildepark:
See urbit/landscape#117.

image

Would appreciate a once-over on Firefox and the like, but seems to look fine on all sizes on Webkit.


matildepark:
ping @liam-fitzgerald

liam-fitzgerald (approved):
LGTM!

PR #3738, opened 2020-10-16
threads: make spider HTTP threads compatible with dojo threads, and fix the %graph-leave thread

tacryt-socryp:


liam-fitzgerald (approved):
LGTM! 
I do question the behaviour of having threads take a list of their arguments. Taking a tuple seems more inline with the prototypical hoon behaviour of lists only when you need to recurse over them. It creates a strange impedance mismatch between marks and the vases that they correspond to. I do think that rectifying the inability to call HTTP threads from dojo is more important than this qualm, so approved!

ohAitch:
I mean, could dojo threads receive a list of cages, instead of a vase of a
list? (Or generator-style fixed-length "list", as far as the type is
concerned)

philipcmonk:
> I mean, could dojo threads receive a list of cages, instead of a vase of a
> list?  (Or generator-style fixed-length "list", as far as the type is
> concerned)

Who's to say `-thread %a %b` is `[!>(%a) !>(%b) ~]` instead of `!>((list @ta))`?  As far as I can tell, we mirror exactly what generators get, except as a vase.

While this PR should go in for now, I've come to agree threads and generators should probably not take a "list" of arguments.  I think that was incorrectly imported from earth, and so far hardly any generators take variable length arguments.  Unlike in bash, it's quite convenient to create structured data, so we tend to use that more, and even anything that would have taken variable-length arguments changes just slightly, to require a trailing `~`.  Better to think of generators/threads as functions of one argument, like everything else.

PR #3739, opened 2020-10-16
clay: speed up +find-merge-points

### LINKED ISSUE #3735: Landscape doesn't load any content
**Describe the bug**
After the latest OTA `ht5ec` loading Landscape only shows home and profile tiles. After that, from dojo, I can see %eyre engaging («http») but pretty much blocks everything, so one can only ctrl+Z and boot again.

**Screenshots**

Screen Shot 2020-10-16 at 5 38 38 PM

Screen Shot 2020-10-16 at 5 57 47 PM

**Desktop**
 - OS: MacOS 10.14.6
 - Browser: Brave

**Additional context**

I ran into the `end of file` error but used `|pack` and `|mass` as explained here https://github.com/urbit/urbit/issues/3645#issuecomment-709631355 and after that got the issues with Landscape. I've also noticed that running `+trouble` sometimes takes a very long time (tried starting with and without `-L`)

Without launching Landscape, I start receiving messages and everything seems to work (except when using `+trouble`)

philipcmonk:
In certain cases +find-merge-points was very slow.  Specifically, the
`done` set was meant to avoid checking the same commit repeatedly, but
it didn't catch the case where a commit was added to the worklist that
was already in that worklist.

Secondly, the worklist was stored as a list but used as a queue, which
resulted in a lot of unnecessary welding.  We change it to a qeu.

Fixes #3735

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -2460,7 +2460,7 @@
     =/  ali-takos  (reachable-takos:ze r.ali-yaki)
     ::  Tako worklist
     ::
-    =/  takos=(list tako)  ~[r.bob-yaki]
+    =/  takos=(qeu tako)  [r.bob-yaki ~ ~]
     ::  Mergebase candidates.  Have proven they're common ancestors, but
     ::  not that they're a most recent
     ::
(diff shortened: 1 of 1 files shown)


matildepark:
Were you thinking of deploying this as a hotfix? Seems pressing.

philipcmonk:
Yes I will deploy it soon

vvisigoth:
King
—
~poldec-tonteg
http://urbit.org

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -2488,20 +2489,24 @@
...
+        outer-loop
       bases-loop(base-list t.base-list)
     ::  Append parents to list and recurse
     ::
-    =/  bob-yaki  (~(got by hut.ran) i.takos)
-    outer-loop(takos (weld t.takos (skip p.bob-yaki ~(has in done))))
+    =/  bob-yaki  (~(got by hut.ran) tako)
+    =/  new-candidates  (skip p.bob-yaki ~(has in done))
+    %_  outer-loop
+      done   (~(gas in done) new-candidates)
```
belisarius222, line 2507: If I'm reading this right, `done` is somewhat misleadingly named, and it's really more like the set of commits we've marked as being relevant.  The old code didn't add to `done` here, but if my understanding is correct, then this should be a performance improvement and not break functionality.

belisarius222 (commented):
LGTM as long as I properly understood the purpose of `done`

pkg/arvo/sys/vane/clay.hoon
philipcmonk: You're right, `done` should be named `seen` or `enqueued` or something.  As the comment specifies, it's the ones we've already checked OR are already in our list to check.  If we don't care about the order, we could make `takos` (our worklist) a set and check against that directly, but I think we're best off processing the worklist in breadth-first order, which is what the queue does.

PR #3743, opened 2020-10-16
profile: add html color picker

tylershuster:
Fixes https://github.com/urbit/landscape/issues/118


matildepark:
There's that @tylershuster magic

matildepark (changes requested):
This breaks the previous style — 

image

I suggest perhaps `position: absolute`-ing it inside the input as `position: relative` and then applying the same border and border-radius to complete the input, as before.

tylershuster:
Hm, this must be not Firefox. I'll do some browser testing

matildepark:
That's Brave Dev on macOS

tylershuster:
Ok, slightly different approach but it looks good (better) and works the same now

tylershuster:
FYI Brave's color picker is superior

matildepark (approved):
lgtm!

Issue #3704, opened 2020-10-14
groups: any user can add channels


matildepark:
Until we have a concrete group role story we should really just hide the ‘new channel’ button for any user who isn’t an admin.

tacryt-socryp:
Related: group admins should have the ability to remove channels from their group surfaced more readily.

tacryt-socryp:
We do currently hide the new channel button for non-admins. Do we have a concrete group role story, or is that being fleshed out by design for Q1?

urcades:
@tacryt-socryp designs are complete for this! They have yet to receive a round of critique, but structurally, I think the approach is rather solid.

See the following prototype:

https://www.figma.com/proto/ovD1mlsYDa0agyYTdvCmGr/Landscape?node-id=3405%3A0&viewport=398%2C13%2C0.22694168984889984&scaling=min-zoom

urcades:
In the "Admins" version of the group, you should see in the "Channels" section the ability to set channel creation limitations.

urcades:
To be a bit more clear about the permissions 'story' @tacryt-socryp :

Currently the permissioning being pushed is a very simple model by which we (still) have "Admins" and "Members" only.

Channel creation is determined against these two roles, where the toggle is binary:

1. Only admins can create channels
2. All members can create channels

The included detail here is that at any time, the original creator of a channel can delete/archive/remove their channel, but Admins can delete/archive/remove anyone's channel, for moderation purposes.

Channel creators can set write permissions (I am currently ignoring read permissions for the time being, for simplicity purposes) for their channels which are separate from the role distinction made between Admins and normal members. Write permissions can be edited at any time in a channel's settings.

urcades:
Let me know if you have any questions here — the linked interface reflects the above description, but feel free to leave questions/comments directly in the figma prototype.

tacryt-socryp:
This sort of "permissioning toggle" seems like an appropriate item to place in `%metadata-store`. Would you like to chat sometime about moving this forward? I've let it lie fallow a bit too long.

urcades:
@tacryt-socryp  Yeah, we should definitely schedule some time to chat, or even chat during midweek. I'm crystalizing a model around this, interfaces included, and it's worth giving it a look-over.

matildepark:
Progress is currently being tracked in https://github.com/urbit/landscape/issues/197.

PR #3745, opened 2020-10-17
landscape: only show 'new channel' if admin

matildepark:
- Checks if we're in the home workspace or if we're admin of the group when displaying the 'new channel' button.

This is, of course, a lie if we don't also prevent this on back-end, but it's a stopper for now. Part 1 of resolving #3704.


liam-fitzgerald (approved):
LGTM!

PR #3746, opened 2020-10-18
eyre: reduce memory usage of unacked channel events

Fang-:
...by over 90%, in the observed case.

Eyre keeps channel events in state until they get acked by a connected client, so that the client can catch up and resume operating after unexpected disconnects. Currently, Eyre stores the fully serialized versions of those events, in the format in which they're sent over the wire.

This PR makes Eyre lazy about serializing events coming from gall agents, delaying both the noun -> json, and json -> event-stream conversions until they're actually needed. Instead of storing serialized event-stream data in state, we simply store the `sign`s. This generally takes up less memory, and potentially makes use of noun deduplication to reduce the footprint even further.

The cost we pay is needing to reserialize in the "reconnected to channel with unacked events" case, potentially making that slower. That seems like a fair trade-off to make, especially considering the addition of a "clog" strategy to prevent build-up, like #3680, or the more sophisticated variant discussed therein (which I intend to write next).

Some numbers, from adding a thousand messages to chat-store and observing them as individual events in the front-end. (To observe more clearly, I flipped the order of vanes in `|mass` logic so that gall/chat-store would get counted first. That might not have mattered much, as far as illustrating effectiveness goes!)

Baseline:
```
%chat-store: KB/318.200
%channels: KB/165.160
```

After 1k messages, on master:  
```
%chat-store: KB/359.468
%channels: MB/17.636.968
```

After 1k messages, with 8def1db (json nouns in event queue):  
```
%chat-store: KB/359.376
%channels: MB/1.405.944
```

After 1k messages, with 7e5f29c ("native" marks/data in event queue):  
```
%chat-store: KB/359.376
%channels: MB/1.396.504
```

"Compacting" the unacked events from 17.6 to ~1.4 MB seems pretty worthwhile. 7e5f29c seems to have much less relative impact, so we could debate whether its bit of added complexity is worth it, but keep in mind that the numbers might skew differently for apps with less straightforward `$-(noun json)` in their marks.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1214,10 +1214,9 @@
           events
         =^  head  queue  ~(get to queue)
         =,  p.head
-        =/  sign=(unit sign:agent:gall)
-          (channel-event-to-sign channel-event)
-        ?~  sign  $
-        $(events [(event-json-to-wall id (sign-to-json request-id u.sign)) events])
+        ?~  sign=(channel-event-to-sign channel-event)  $
+        ?~  json=(sign-to-json request-id u.sign)       $
+        $(events [(event-json-to-wall id u.json) events])
       ::  send the start event to the client
       ::
       =^  http-moves  state
@@ -1363,7 +1362,7 @@
           :^  duct  %pass
             (subscription-wire channel-id request-id ship app)
           :*  %g  %deal  [our ship]  app
-              `task:agent:gall`[%watch-as %json path]
+              `task:agent:gall`[%watch path]
           ==
         ::
         =.  session.channel-state.state
@@ -1469,6 +1468,8 @@
       ::
       =?  moves  ?=([%| *] state.u.channel)
         ^-  (list move)
+        ?~  json=(sign-to-json request-id sign)
+          moves
         :_  moves
         :+  p.state.u.channel  %give
         ^-  gift:able
@@ -1476,7 +1477,7 @@
         ::
             ^=  data
             %-  wall-to-octs
-            (event-json-to-wall event-id (sign-to-json request-id sign))
+            (event-json-to-wall event-id u.json)
         ::
             complete=%.n
         ==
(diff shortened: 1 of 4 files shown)


tylershuster:
Get in mah belleh

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1511,30 +1464,27 @@
...
-        :-  (weld "id: " (format-ud-as-integer event-id))
-        %+  turn  json-text
-        |=  =tape
-        (weld "data: " tape)
       ::  if a client is connected, send this event to them.
       ::
       =?  moves  ?=([%| *] state.u.channel)
         ^-  (list move)
+        ?~  json=(sign-to-json request-id sign)
+          moves
```
Fang-, line 1472: Hmm. @eglaysher this means we might "skip" event IDs, as far as the client can tell. Considering our acking assumes but doesn't depend on contiguous numbers, and the SSE spec doesn't require IDs to be of a specific shape/order, I think this should be fine. But please do give me a second opinion!

eglaysher: That probably isn't an issue.

Fang-: Ah, here's the missing piece of the puzzle: when you `%watch-as`, gall kicks the subscription if a conversion isn't available/functioning. So I think the answer here is "check if `sign-to-json` works, if it doesn't, `%kick` the subscription".

eglaysher (approved):
Minus not understanding the clay scrying details, I think this should work. Thank you.

pkg/arvo/sys/vane/eyre.hoon
Fang-: Alright, fix or this in 5bebff3, which is in the other PR because it's at the intersection of these two changesets, and I have a conflict-averse personality. Apologies for tangling these things up so much. @eglaysher please take a look.

eglaysher: (Not sure what you're asking for, but the changes in 5bebff3 look reasonable enough as error handling.)

Fang-:
@philipcmonk if you could take look at the use of clay scries here just in case, much appreciated. Feel like we might want stdlib functions for these things if their use becomes more common, but we can see about that.

philipcmonk (approved):
The scries look correct, and storing without vases should make this easy to upgrade.  I agree that we shouldn't depend on data -> noun conversions preserving structural sharing if we can avoid it.

PR #3754, opened 2020-10-20
hoon: add +succ gate

liam-fitzgerald:
Adds an +inc (increment) gate. Whilst .+ exists, it is impossible to use
in a point free style and doesn't compose functionally. 

Should this be in functional hacks instead of unsigned int math?


ohAitch:
the `~/` is a bit on the nose

pilfer-pandex (approved):

joemfb (changes requested):
As discussed out of band, I'm somewhat opposed to adding this. I think we should have a consistent story about when we're going to additionally expose runes as gates, and not just add them haphazardly. (If we wanted them all, we could codegen them in the compiler.)

People should use `.+` directly wherever they can -- if the gate is present, I expect it'll be used more than necessary. Putting it in "functional hacks" (maybe as `+succ`) might better convey that.

tacryt-socryp:
@liam-fitzgerald please apply Joe's suggestion so we can get this merged

tacryt-socryp (approved):
@joemfb only waiting for a stamp from you now

tacryt-socryp:
@philipcmonk all review suggestions were applied in this PR, should be good to go

PR #3755, opened 2020-10-20
groups: add 'kick from' action

matildepark:
- Adds 'kick' action for admins, because sometimes you need to just give a ship a kick (if breached, etc.)


liam-fitzgerald (approved):
LGTM!

(how nice is `StatelessAsyncAction` & friends)

PR #3756, opened 2020-10-20
groups: new channel only shows ship search at home

matildepark:
- The 'ship search' prompt in the new channel menu shows in all workspaces; but it's only really necessary in the home workspace.
- In this PR we just hide it away when we're inside a group workspace, so inviting participants only happens from one place (the invite button in the sidebar)


liam-fitzgerald (commented):
LGTM!

liam-fitzgerald (approved):

PR #3757, opened 2020-10-20
groups: allow search to have sigs

matildepark:
If you type `~haddef-sigwen` into the participant search, nothing shows up. If you type `haddef-sigwen`, I do indeed show up. 

In this PR we just trim off the query's sig if it starts with it.

(Also fixes an unfiled bug where the search was case sensitive by accident, because we were comparing the `nickname.toLowerCase()` against the raw, `search` query, not `s` which is `search` transformed to `lowercase` itself.)


liam-fitzgerald (approved):
LGTM!

Issue #3748, opened 2020-10-19
Joining channel in a group navigation error ("Cannot read property 'envelopes' of undefined")


Stillstellung:
**Description**
Navigating away from a channel mid join attempt to do something else in Urbit Landscape, then navigating back to it, results in an error message.

**Reproduction Steps**
1) try to join channel in group
2) channel join taking long time
3) navigate to another element of group
4) click channel originally trying to join

**Expected behavior**
Un-joined channel is in a state where user can attempt to join again

**Screenshots**
If applicable, add screenshots to help explain your problem. If possible, please also screenshot your browser's dev console. Here are [Chrome's docs](https://developers.google.com/web/tools/chrome-devtools/open) for using this feature.

**Stack Trace**
```Error "Cannot read property 'envelopes' of undefined"
TypeError: Cannot read property 'envelopes' of undefined
    at ad (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:461:83906)
    at Zo (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:57929)
    at gs (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:104168)
    at cl (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:96716)
    at sl (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:96641)
    at Qs (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:93671)
    at https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45313
    at t.unstable_runWithPriority (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:322:3844)
    at Hi (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45022)
    at Wi (https://sigdur-timlus.tlon.network/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45258)
```

PR #3758, opened 2020-10-20
chat: prevent crashes if join is interrupted

matildepark:
Addresses #3748.

We destructure envelopes and config from the station inside our inbox unsafely; and if you navigate away mid-join, you can sidestep the join flow and see the view before the props are instantiated — causing a consistent crash when you navigate to the room, until the full backlog hits our ship and the messages load in on a subsequent refresh.

By safely instantiating (more elvis operators!) we can avoid the crash, which nicely enough shows the "Past messages are being restored" notice while we wait, which I haven't yet seen in nu-groups, so it seems to be functioning normally.


liam-fitzgerald (approved):
LGTM!

Issue #3673, opened 2020-10-08
Float jets: detect oversized arguments


ohAitch:
**Describe the bug**
When passing an e.g. 32-bit float core an atom >32 bits wide, it blindly operates on the bottom word, without first checking the input size is sane.

**To Reproduce**
```
> (add:rs .10 `@`.~10.4)
.-1.0737417e8
```

**Expected behaviour**
```
> (add:rs .10 `@`.~10.4)
bad-float
dojo: hoon expression failed
```

**Additional context**
Obviously casting a double to an f32 is a convoluted example, but if lagoon turns into a full-blown `@rss` tensor type (guaranteed to be multiple words long for nonscalars), this would be a useful sanity check.

**Notify maintainers**
@joemfb and @philipcmonk seemed supportive of my previous float ticket #3624

joemfb:
This doesn't appear to be (specifically) a jet issue, the hoon appears to just `+cut` out the low bits of the input (in `+sea:ff`). I'm not sure this design is a problem, the operations are just defined on a larger domain than necessary. That could be a nice property, allowing the jet to avoid a branch. But the sanity check could also be nice. I'm not sure which is better.

/cc @frodwith

ohAitch:
Fair, I had not been meaning to imply a _mismatch_ here, just semantics change-request.

PR #3759, opened 2020-10-20
[hoon @r] add .0j1 complex float syntax

ohAitch:
Stores real then complex part in the obvious LSB manner: `=(.123j0 .123)`, `=(.3j-4 (cat 5 .3 .-4))`

Parsing feature-flagged as no support has been implemented in the math
cores, which will just ignore/truncate off the complex part.

In dojo,

    =enable-complex-float  &

to try out the syntax. (Or just go nuts with,
```
> ? (div:rs .15 .8)
  @rs
.1.875
> ? .~1
  @rd
.~1
> `@ux`(div:rs .15 .8)
0x3ff0.0000
> `@ux`.~1
0x3ff0.0000.0000.0000
> `@rs`.~1
.0j1.875
```
etc)

Steps to leave draft status:
- [ ] Figure out if we want a `@rsr` `@rdr` etc odor for "real, as in actually real", and implement that in the parser / otherwise resolve the TODO
- [ ] Update all the `ff` library functions and `++r[sdhq]` jet-variants to correctly calculate complex results, or to reject complex inputs per #3673 leaving correct implementations for future work; or at least figure out a more principled way to do the "opt into syntax that breaks all your math" flag.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -114,9 +114,10 @@
 ::  |parser-at: parsers for dojo expressions using :dir as working directory
 ::
 ++  parser-at
-  |=  [our=ship dir=beam]
+  |=  [our=ship dir=beam no-complex=?]
   |%
   ++  default-app         %hood
+  ++  vang                %*(. ^vang no-complex no-complex)
   ++  hoon-parser         (vang | (en-beam:format dir))
   ++  our                 p.dir
   ::
@@ -324,7 +325,9 @@
     dir(r [%da now.hid])
   ::
   ++  he-beak    `beak`[p q r]:he-beam
-  ++  he-parser  (parser-at our.hid he-beam)
+  ++  vang  %*(. ^vang no-complex !enable-complex-float)
+  ++  he-parser  (parser-at our.hid he-beam !enable-complex-float)
+  ++  enable-complex-float  (~(has by var) %enable-complex-float)
   ::
   ++  dy                                                ::  project work
     |_  dojo-project                                    ::
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -2265,12 +2265,12 @@
   ::
 ::
 ++  fl                                                  ::  arb. precision fp
-  =/  {{p/@u v/@s w/@u} r/$?($n $u $d $z $a) d/$?($d $f $i)}
+  =/  {{p/@u v/@s w/@u} r/$?($u $d $z $a $n) d/$?($d $f $i)}
     [[113 -16.494 32.765] %n %d]
   ::  p=precision:     number of bits in arithmetic form; must be at least 2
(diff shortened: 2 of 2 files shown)


sigilante:
(This question presses me because it impinges on how to set up vector types in the matrix library #3536 .)

There are several possible configurations for a complex number type:

1. Complex values have a separate new aura:  `@qd`
2. Complex values are nested inside of floating-point auras:  `@rdc`, etc.
3. Complex values live in the matrix library as a new aura:  `@lcd`
4. Complex values live in the matrix library only as vector/matrix elements:  `@lvcd`

##  Separate New Aura

Complex values are stored as real and imaginary parts inside of a single atom as upper and lower bytes.

An unoccupied aura like `@q` would be used, since the logical choices `@c` and `@i` are already occupied with incompatible types.

Pros:

- Values with the aura has separate handling logic.

Cons:

- Values with the aura don't sit next to similarly-used values such as `@r`.


##  `@r` Nested Aura

Complex values are stored as an extension of the floating-point auras.

Pros:

- Complex-valued logic can live beside the real-valued logic.

Cons:

- Introduces aura complexity relatively close to the language primitives.


##  Matrix Library Aura

In addition to the matrix library types (currently `@lv` for vectors and `@lm` for matrices), we can introduce `@lc` as a container aura for `@lcs`, etc.

Pros:

- Complex values are handled by optimized routines in the matrix library.

Cons:

- Still requires case 4.

##  Matrix Library Value

In this case, complex values are implemented only within the matrix library as entries in a complex-valued vector or matrix.  That is, the primitive complex value is the special case of a 1×1 matrix or a 1× vector, such as `@lvcd` (for vector-complex-double).

Pros:

- Implements complex numbers in the matrix library where they are most likely to be useful.

Cons:

- Requires matrix library to become part of Hoon or Zuse if complex types are going to become Hoon primitives.  (Not necessarily a con, but a design decision.)

---

I personally favor case 3, which is a superset of case 4, but I welcome cases for other configurations.

sigilante:
I should also mention the `@rt` structure that @ohAitch mentioned here:  https://github.com/urbit/urbit/pull/3536#discussion_r505775785

ohAitch:
`@q` is currently occupied by an "unscrambled `@p`" format, though I don't think literally anything ended up using it.

sigilante:
Doh!  Yes, of course.

philipcmonk:
(didn't mean to close this, apparently github deletes PR branches even if other PRs are targeted to them and doesn't retarget the open PRs?)

ohAitch:
I think there's a checkbox for it or sth?

philipcmonk:
normally when you have PR A->B and B->master and then you merge B into master, then the first becomes A->master automatically.  maybe that only applies to the master branch, or maybe it doesn't work when you delete the branch as you're merging

joemfb:
I think you can reopen it if you restore the branch, then retarget elsewhere.

PR #3761, opened 2020-10-20
hoon, sur/sole, dill, drum: Add 24-bit true color

crides:
This commit adds 24bit true color capabilities to `sole-effect` for
those terminal supporting it (which most modern terminal does). As Urbit
does not do `termcap` detection, this also does not attempt that. But on
terminals that doesn't support true color (e.g. linux console), the
color would be truncated to the nearest achievable approximate.

The changes roughly follows what is done for normal 8-bit colors (`styx`
and related). The definition lives in `hoon`, the structural changes in
`sur/sole`, (in my case) passing the color structure through `drum` as
is, and coverting to escape sequences in `dill`.

Type/Arm names should be revised to fit Urbit/Hoon style.

cc related component maintainers: @pilfer-pandex @joemfb @philipcmonk

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1972,7 +1972,9 @@
 +$  tile  ::  XX: ?@(knot (pair styl knot))
           ::
           cord
-+$  tint  ?($r $g $b $c $m $y $k $w $~)                 ::  text color
++$  tint
+  $@  ?($r $g $b $c $m $y $k $w $~)                     ::  text color
+  [r=@uxD g=@uxD b=@uxD]                                ::  24bit true color
 ::
 ::  A `plum` is the intermediate representation for the pretty-printer. It
 ::  encodes hoon-shaped data with the least amount of structured needed
--- a/pkg/arvo/sys/vane/dill.hoon
+++ b/pkg/arvo/sys/vane/dill.hoon
@@ -211,7 +211,6 @@
               (bg ~)
               (fg ~)
           ==
-        ::
         ++  ef  |=(a/^deco (scap (deco a)))             ::  ANSI effect
         ::
         ++  fg  |=(a/^tint (scap (tint a)))             ::  ANSI foreground
@@ -219,13 +218,13 @@
         ++  bg                                          ::  ANSI background
           |=  a/^tint
           %-  scap
-          =>((tint a) [+(p) q])                         ::  (add 10 fg)
+          =>((tint a) [+(-) +])                         ::  (add 10 fg)
         ::
         ++  scap                                        ::  ANSI escape seq
-          |=  a/$@(@ (pair @ @))
+          |=  a/$@(@ (list @))
           %-  (list @c)
           :+  27  '['                                   ::  "\033[{a}m"
-          ?@(a :~(a 'm') :~(p.a q.a 'm'))
+          ?@(a :~(a 'm') (snoc a 'm'))
         ::
         ++  deco                                        ::  ANSI effects
           |=  a/^deco  ^-  @
@@ -238,19 +237,25 @@
         ::
         ++  tint                                        ::  ANSI colors (fg)
           |=  a/^tint
-          ^-  (pair @ @)
+          ^-  (list @)
           :-  '3'
-          ?-  a
-            $k  '0'
-            $r  '1'
-            $g  '2'
-            $y  '3'
-            $b  '4'
-            $m  '5'
-            $c  '6'
-            $w  '7'
-            ~  '9'
-           ==
+          ?@  a
+            :_  ~
+            ?-  a
+              $k  '0'
+              $r  '1'
+              $g  '2'
+              $y  '3'
+              $b  '4'
+              $m  '5'
+              $c  '6'
+              $w  '7'
+              ~  '9'
+             ==
+          =/  r  (~(sit fe 3) r.a)
+          =/  g  (~(sit fe 3) g.a)
+          =/  b  (~(sit fe 3) b.a)
+          "8;2;{<r>};{<g>};{<b>}"
         --
       ::  XX move
       ::


Fang-:
You didn't tag me, but I've been doing (and continue to do) work in this area, so I'll comment anyway. (;

First of all, this is a very exciting change, thank you!

For ease of upgrade, have you considered re-using the existing types, and fitting this color information into `%klr`, instead of defining a new sole & dill effect? I think you might be able to fit this information into `tint` without breaking existing usage:

```hoon
+$  tint
  $@  [r=@ud g=@ud b=@ud]
  ?($r $g $b $c $m $y $k $w $~) 
```

Second, I must apologize for not pointing this out before, but please take a look at #3467 and #3468 (and I guess #3487). tldr: the rendering of ANSI escape sequences got factored out into the runtime. Dill now just sends the `%klr` as-is.

This isn't necessarily important for this PR, we could probably push it out regardless of the state of the above and have it immediately available. But the `%color` rendering will have to be implemented in both runtimes by the time #3468 gets pushed to the network. Happy to help you with doing this, or do the work in your stead if you're not comfortable with C and/or Haskell.

crides:
Thanks for looking at the pr!

I have actually looked into using existing types, and I haven't actually thought of combining `tint` and `true-color` into a single type. In this case a new structure is indeed not necessary, but I've been confused by the way `styx` is set up from the start. putting an extra `styx` at the end is just kind of weird [1], and I don't want people to be confused like me when they use this interface. That said, merging the two interface is the more important thing.

Regarding #3467 and #3468, I have heard about that from you earlier, but I completely forgot about it. I don't think anyone is (or that I'm just not aware) doing color related work, so true color is probably not in a hurry. In terms of runtime implementations, I don't have a problem with writing C, and I've also learned Haskell several times (although I've never written anything in it), but I definitely need time before I can get used to the urbit way of writing C or Haskell.

[1]: as if multiple effects need to be applied? But then that should only be for `deco`, as fg and bg can be specified at the same time. So why isn't `deco` in a `list`?

joemfb:
@crides, this is a good step. I agree that we should reconcile these types with the existing design, and that the runtime is the right place for rendering. Regardless, I'll try to leave a code review here when I can.

The initial discussion surrounding these types is at urbit/arvo#212, and the initial implementation at urbit/arvo#216. The linked "fora" discusssion is archived at https://web.archive.org/web/20171214195005/https://urbit.org/fora/posts/~2016.8.2..21.19.29..2ab8~/, and also available in a `:publish` notebook in the Urbit Community group, but it's kind of hard to find old stuff with so many posts.

@Fang-, your `$@` is backwards, but I otherwise agree. I don't know if hex or decimal is more desirable for these colors, but, either way, the aura should also include the appropriate bitwidth suffix:

```
+$  tint
  $@  ?($r $g $b $c $m $y $k $w $~)
  [r=@udD g=@udD b=@udD]
```

It shouldn't be too hard to stage such a change for OTA update -- `$styx` is just not used too many places. It would probably be better good to do that before #3468 is released.

crides:
Since I should be basing my efforts on #3467 and #3468, and that I'm currently branched off from `master`, where should I be rebasing on top of?

Fang-:
> I've been confused by the way `styx` is set up from the start

It's definitely a bit awkward to use. I've been lightly thinking about ways to make it better, but haven't settled on anything serious yet... Changes to it seem out of scope here, in any case.

> I might try doing that myself over the next weeks, at least for the C version.

Great! For C, this might be as simple as updating `_term_it_put_tint` and its callsites to account for these fancier colors.

https://github.com/urbit/urbit/blob/c6de028fc26e01a6f81690a9f61d5919369a12ed/pkg/urbit/vere/io/term.c#L1125

Careful though, you might want to make sure to allocate a bit more memory:

https://github.com/urbit/urbit/blob/c6de028fc26e01a6f81690a9f61d5919369a12ed/pkg/urbit/vere/io/term.c#L1180-L1184

> where should I be rebasing on top of?

For the short-term arvo-side changes, base off of `release/next-sys`.  
For the runtime changes, please work off `release/next-vere` (and test that by running `release/next-dill` as your arvo version, with your `%klr` changes merged in).


> your `$@` is backwards

\**shakes fist at insufficiently clever compiler*\*

Re hex vs decimal, `@uxD` is probably the right call, yeah.

crides:
Per @Fang- 's suggestion, this is a `dill`-only implementation of the change, rebased on top of `release/next-sys`. The 24-bit color is squashed into `tint`.

crides:
Sorry I haven't done any serious work on any large git repo. I feel like I screwed up a little bit. I guess that was because I'm rebasing to an older branch. I think I drop the 20-something commits between mine and `release/next-sys`. Is that correct?

joemfb:
Yeah, interactive rebase to drop the commits would work.

Since it's just one commit, it's probably a little easier to just cherry-pick your commit onto a temporary branch. Here's how I would do that (assuming your fork is `origin`, and our repo is `upstream`):

```
git checkout -b tmp upstream/release/next-sys
git cherry-pick 8456e8a
git branch -f 24bit-colors tmp
git checkout 24bit-colors
git branch -d tmp
git push --force-with-lease origin 24bit-colors
```

Fang- (approved):
Nice! I have no objection to putting this in as-is. You might want to amend the commit message though, to reflect the "just hoon/dill" approach that you switched over to.

crides:
Yeah sure, I should revise the commit message

joemfb:
Do care about fallback from 24-bit to basic text color? In other words, style/colors will be in the console protocol, and can be rendered or not based on the capabilities of the runtime/connected-terminal/&c. Do we want to have ternary configurability there (full style, minimal style, no style)? I believe this question also came up in the linked discussions, but I don't think it was answered.

If we do, extending `$tint` is not right, we'd probably want to extend both `$styl` and `$stye`. But if we don't, this design is more elegant.

/cc @eglaysher

eglaysher:
We probably want fallback, but I believe the fallback would be handled by the terminal driver. (If the user was required to submit multiple configurations, the most likely outcome is that they'll submit no configurations.)

crides:
I don't really have much experience in terms of the standard way to do this, but from my own experience, because Urbit is solely an interactive program, there's little chance the input/output will be piped from/to other programs. So I don't see the need in having the option for that reason.

In terms of terminal capabilities, I think most terminals today support at least 256 colors anyway, and the behavior that I observed is that those that can't support higher color resolutions will just truncate the colors to the nearest representable color.

But that's only from my limited observation anyway.

I've also amended the commit messages, so that it reflects that this is only in `dill`.

joemfb:
That's good to hear re: color truncation -- fallback may already be automatic in the scenarios we care about. (And it shouldn't be too hard to implement if we find that we need it somewhere.)

"Urbit is solely an interactive program" is wrong -- vere has the `-t` and `-d` flags (disable interactive tty assumptions, and that + fork into the background, respectively), and king-haskell has even more substantial "daemon mode" affordances, remote terminal connections, &c. This is a large part of the motivation to move the style rendering into the runtime via the console protocol, so that runtime/execution/rendering environments have the context and control to behave appropriately.

Fang-:
I've included this in #4055, which adds support for this to the runtimes.
Closing in favor of that. Thank you for kicking this off, @crides. (:

crides:
Sorry I did I have enough time to work on the runtime stuff, but thanks for adding this! We'll have a more colorful Urbit in the future!

PR #3762, opened 2020-10-20
chat: timestamps are lined at top of messages

matildepark:
Before:

image

After:

image

cc @tylershuster is there some reason we were using `items-center` on all these besides this timestamp?


tylershuster:
Not as far as I know, no. I don't think I touched that line. Looks good to me.

matildepark:
Oh, git blame was just being very thorough since you touched basically everything in that file.

urcades (approved):
Stamp!

PR #3764, opened 2020-10-20
vere: refactors ames i/o driver

joemfb:
With a focus on packet-parsing, stateless forwarding, and galaxy address resolution.

The packet parsing (original in `_ames_recv_cb()`) did not sufficiently protect against strange lengths, and could underflow the buffer-length (which was signed). The parsing logic was also mixed together with the forwarding logic. These things have been factored out from each other.

The packet parsing and serialization involved a bunch of unnecessary allocations and representation changes, these have been mostly replaced with direct implementations (but there's still a bunch of unnecessary allocations in the higher-level packet flows).

The protocol-version checking/enforcement is incomplete in various ways, this PR includes comments documenting various scenarios we need to support to varying degrees.

I have not tested these changes on the livenet -- I'll do so after source level issues are resolved.

### DIFF
--- a/pkg/urbit/vere/io/ames.c
+++ b/pkg/urbit/vere/io/ames.c
@@ -238,6 +238,45 @@ _ames_ship_of_chubs(c3_d sip_d[2], c3_y len_y, c3_y* buf_y)
   memcpy(buf_y, sip_y, c3_min(16, len_y));
 }
 
+/* _ames_etch_pack(): serialize packet header and body.
+*/
+static c3_w
+_ames_etch_pack(u3_head* hed_u,
+                u3_body* bod_u,
+                c3_o     mug_o,
+                c3_y**   out_y)
+{
+  //  start with the body
+  //
+  c3_y  sen_y = 2 << hed_u->sac_y;             //  sender len
+  c3_y  rec_y = 2 << hed_u->rac_y;             //  receiver len
+  c3_w  bod_w = sen_y + rec_y + bod_u->con_w;  //  body len
+  c3_w  len_w = 4 + bod_w;                     //  packet len
+  c3_y* pac_y = c3_malloc(len_w);
+  c3_y* bod_y = pac_y + 4;
+
+  _ames_ship_of_chubs(bod_u->sen_d, sen_y, bod_y);
+  _ames_ship_of_chubs(bod_u->rec_d, rec_y, bod_y + sen_y);
+
+  {
+    c3_y* con_y = bod_y + sen_y + rec_y;
+    memcpy(con_y, bod_u->con_y, bod_u->con_w);
+  }
+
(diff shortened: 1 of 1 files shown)


pkg/urbit/vere/io/ames.c
```diff
@@ -116,31 +120,195 @@ _ames_pact_free(u3_pact* pac_u)
...
+  c3_assert( 0 == hed_u->ver_y );
+
+  buf_y[0] = hed_w & 0xff;
+  buf_y[1] = (hed_w >>  8) & 0xff;
+  buf_y[2] = (hed_w >> 16) & 0xff;
+  buf_y[3] = (hed_w >> 24) & 0xff;
+}
+
+/* _ames_chub_bytes(): c3_y[8] to c3_d
+** XX move
```
Fang-, line 191: To where? Might be in scope for this.

pkg/urbit/vere/io/ames.c
```diff
@@ -748,188 +909,227 @@ _ames_lane_scry_cb(void* vod_p, u3_noun nun)
...
-    u3z(sen); u3z(rec);
-
-    con_y = c3_malloc(con_w);
-    memcpy(con_y, bod_y + sen_y + rec_y, con_w);
-
-    //  ensure the content is cue-able
+  if (  (4 > len_w)
+     || (c3n == _ames_sift_head(&hed_u, hun_y)) )
+  {
+    //  XX track stats
```
Fang-, line 1028: Probably trivial to add, and useful statistics to have going forward.

Fang- (approved):
This all seems fair, thanks for cleaning this up! Some suggestions on tiny `XX`s you might still want to pick up, but good either way.

pkg/urbit/vere/io/ames.c
joemfb: Added.

pkg/urbit/vere/io/ames.c
joemfb: There's a bunch of placed were we do this kind byte/integer un/packing -- it'd be nice to factor it out where possible. I've updated the comments.

joemfb:
This PR now includes a partial fix for #3778, by preventing galaxies from forwarding to themselves. (I say partial, because I think the %ames scry handler should be updated to always exclude itself from forwarding lanes.)

joemfb:
FWIW, I've tested this PR pretty extensively with fake ships (including running it under ASAN) but have still not tested it on the mainnet.

joemfb:
I've run this (merged with release/next-vere) on `~zod` for awhile now, monitoring closely. ~zod's forwarded > 2 million packets in ~m50, with single-digit-percentage CPU usage throughout most of that time.

PR #3767, opened 2020-10-21
clock: render in SVG, show moon phase

tylershuster:
image

### DIFF
--- a/pkg/interface/src/views/apps/launch/components/tiles/clock.js
+++ b/pkg/interface/src/views/apps/launch/components/tiles/clock.js
@@ -1,38 +1,28 @@
 import React from 'react';
 import moment from 'moment';
 import SunCalc from 'suncalc';
+import styled from 'styled-components';
 
 import Tile from './tile';
 
-const innerSize = 124; // clock size
+const VIEWBOX_SIZE = 100;
+const CX = VIEWBOX_SIZE / 2;
+const CY = VIEWBOX_SIZE / 2;
+const RADIUS = VIEWBOX_SIZE / 2;
+const CELESTIAL_BODY_SIZE = 16;
 
-// polar to cartesian
-// var ptc = function(r, theta) {
-//   return {
-//     x: r * Math.cos(theta),
-//     y: r * Math.sin(theta)
-//   }
-// }
-
-let timeTextColor = '#000000', dateTextColor = '#333333', background = '#ffffff';
-
-const dark = window.matchMedia('(prefers-color-scheme: dark)');
-
-if (dark.matches) {
-  timeTextColor = '#ffffff';
-  dateTextColor = '#7f7f7f';
-  background = '#333';
-}
-
-function darkColors(dark) {
-  if (dark.matches) {
-    timeTextColor = '#ffffff';
-    dateTextColor = '#7f7f7f';
-    background = '#ffffff';
+const ApplyClockBg = styled.div`
+  .background {
+    fill: ${p => p.theme.colors.white};
   }
- }
-
-dark.addListener(darkColors);
+  .time {
+    fill: ${p => p.theme.colors.black};
+    color: ${p => p.theme.colors.black};
+  }
+  .date {
+    fill: ${p => p.theme.colors.gray};
+  }
+`;
 
 const toRelativeTime = (date, referenceTime, unit) => moment(date)
   .diff(referenceTime, unit);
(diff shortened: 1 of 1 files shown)


urcades:
I see the utility for dimming the donut hole for readability reasons, but it detracts from the geometric starkness of the interface that I personally find very pleasing..

I think if we really wanted a utility-driven clock, we'd simply stick a numerals-only clock in the menubar area and be done with it — I think the homescreen-based clock is somewhat of a more artistic object that doesn't exactly need to be 'read' more than it needs to be 'perceived'.

urcades:
I'd be okay with retaining the "dimmed and more legible" donut hole if others felt strongly about it. 

I wouldn't die on a hill to make the clock **less** legible, heh. Just expressing the 'keep it an art object' take.

matildepark:
> I see the utility for dimming the donut hole for readability reasons, but it detracts from the geometric starkness of the interface that I personally find very pleasing..

Why not make it always black bg

urcades:
> > I see the utility for dimming the donut hole for readability reasons, but it detracts from the geometric starkness of the interface that I personally find very pleasing..
> 
> Why not make it always black bg

Oh, well to be clear, I'm pro no background at all — literally just a punched-out void that would allow you to see the background, so the "ring" of the clock is the shape that is emphasized. This obviously gets into low-legibility territory depending on the color/image of a user's background.

Basically, if others agree we like the clock for its artistic qualities, I'd say we should vouch for no background. 

If we believe the clock should meet a bar of accessibility/legibility, we should probably leave the background as-is.

matildepark:
> Oh, well to be clear, I'm pro no background at all — literally just a punched-out void that would allow you to see the background, so the "ring" of the clock is the shape that is emphasized. This obviously gets into low-legibility territory depending on the color/image of a user's background.
> Basically, if others agree we like the clock for its artistic qualities, I'd say we should vouch for no background.

Why not remove the time and date if the actual timetelling isn't important

urcades:
While typing the above, I realized if we had a more developed approach to on-hover interactions, the clock could be a pure visual ring, and could display the time on-hover, using a background matched to the light/dark theme.

urcades:
> Why not remove the time and date if the actual timetelling isn't important

Well, that's the primary object of debate here imo, I'm more or less trying to attain a temp check on whether we think this interface is more art/fun or intended to be a utility-based thing, that's all.

g-a-v-i-n:
The svg clock is very cool. Canvas was probably overkill. Could these changes be made?
- make the donut hole white. 
- make the time black. 
- make the date gray60. 

Thanks!

urcades:
> make the donut hole white

and w/e background color we use for dark mode, I presume?

g-a-v-i-n:
yes, but that should happen automatically via the theme

tylershuster:
You would think so. But because this is SVG, indigo colors aren't available automatically. Do you have an example in the `indigo-react` repo that you can point to? I need to set them manually and add a listener for dark mode.

matildepark:
> You would think so. But because this is SVG, indigo colors aren't available automatically. Do you have an example in the indigo-react repo that you can point to? I need to set them manually and add a listener for dark mode.

Just check the two values for `white` in the themes directory and use the listener logic as in the root application. Actually I think the clock has a listener already.

g-a-v-i-n:
You can give an svg element an id or className, and then target that in a styled component wrapper. Up to you.

```jsx

const ApplyDonutBg = styled.div`
   #donut {
       fill: ${p => p.theme.colors.white};
}
`
```
and
```jsx
<ApplyDonutBg>
<svg
        style={{ height: '100%', width: '100%'}}
        viewBox={`0 0 ${VIEWBOX_SIZE} ${VIEWBOX_SIZE}`}
        xmlns="http://www.w3.org/2000/svg"
        xmlnsXlink="http://www.w3.org/1999/xlink"
      >
...
 <circle cx={CY} cy={CY} r={RADIUS} fill={background} id={#donut} />
...
```

g-a-v-i-n (approved):

tylershuster:
fixing colon and rebasing

matildepark:
Loadin' up Brisbane, Australia time zone and ...

image

matildepark:
Reverting for now ...

tylershuster:
Well that's because they're all upside down over there

matildepark:
No I mean if you set your coordinates to Australia and also your date/time to Australia you'll see this, we just need to enforce a mask on that donut

PR #3769, opened 2020-10-21
king: implement -I and --inject-event-list

eglaysher:
This makes it so during `run`, you can inject a series of Events
into your Urbit. The `-I` injection was implemented in Vere, and
has been ported to KH, with the extended `--inject-event-list`
flag which handles a jammed list of events.

Also included is a sample generator which builds a list of events.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
@@ -228,15 +228,17 @@ pierPath :: Parser FilePath
 pierPath = strArgument (metavar "PIER" <> help "Path to pier")
 
 injectEvents :: Parser [Injection]
-injectEvents = many (InjectOneEvent <$> strOption ( short 'I'
-                          <> long "inject-event"
-                          <> metavar "JAM"
-                          <> help "Path to a jammed event")  <|>
-                     InjectManyEvents <$> strOption (
-                        long "inject-event-list"
-                     <> metavar "JAM LIST"
-                     <> help "Path to a jammed list of events"
-                    ))
+injectEvents = many $ InjectOneEvent <$> strOption
+                        ( short 'I'
+                       <> long "inject-event"
+                       <> metavar "JAM"
+                       <> help "Path to a jammed event"
+                       <> hidden)
+                  <|> InjectManyEvents <$> strOption
+                        ( long "inject-event-list"
+                       <> metavar "JAM_LIST"
+                       <> help "Path to a jammed list of events"
+                       <> hidden)
 
 serfExe :: Parser (Maybe Text)
 serfExe =  optional
--- a/pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
@@ -44,6 +44,7 @@ data Opts = Opts
     , oHttpPort     :: Maybe Word16
     , oHttpsPort    :: Maybe Word16
     , oLoopbackPort :: Maybe Word16
+    , oInjectEvents :: [Injection]
     }
   deriving (Show)
 
@@ -75,6 +76,11 @@ data Nat
   = NatAlways
   | NatWhenPrivateNetwork
   | NatNever
+  deriving (Show)
+
+data Injection
+  = InjectOneEvent   FilePath
+  | InjectManyEvents FilePath
   deriving (Show)
 
 data New = New
@@ -221,6 +227,17 @@ pillFromURL = PillSourceURL <$> strOption
 pierPath :: Parser FilePath
 pierPath = strArgument (metavar "PIER" <> help "Path to pier")
 
+injectEvents :: Parser [Injection]
+injectEvents = many (InjectOneEvent <$> strOption ( short 'I'
+                          <> long "inject-event"
+                          <> metavar "JAM"
+                          <> help "Path to a jammed event")  <|>
+                     InjectManyEvents <$> strOption (
+                        long "inject-event-list"
+                     <> metavar "JAM LIST"
+                     <> help "Path to a jammed list of events"
+                    ))
+
 serfExe :: Parser (Maybe Text)
 serfExe =  optional
     $  strOption
@@ -305,6 +322,8 @@ opts = do
       <> long "loopback-port"
       <> help "Localhost-only HTTP port"
       <> hidden
+
+    oInjectEvents <- injectEvents
 
     oHashless  <- switch $ short 'S'
                         <> long "hashless"
--- a/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
@@ -598,7 +598,7 @@ runShip (CLI.Run pierPath) opts daemon = do
 
     loadInjections :: [CLI.Injection] -> RIO PierEnv [Ev]
     loadInjections injections = do
-      perInjection :: [[Ev]] <- forM injections $ \case
+      perInjection :: [[Ev]] <- for injections $ \case
           CLI.InjectOneEvent filePath -> do
             logInfo $ display $ "boot: reading injected event from " ++
               (pack filePath :: Text)
--- a/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
@@ -190,7 +190,7 @@ tryBootFromPill
 tryBootFromPill oExit pill lite ship boot = do
   mStart <- newEmptyMVar
   vSlog  <- logSlogs
-  runOrExitImmediately vSlog (bootedPier vSlog) oExit mStart
+  runOrExitImmediately vSlog (bootedPier vSlog) oExit mStart []
  where
   bootedPier vSlog = do
     view pierPathL >>= lockFile
@@ -204,8 +204,9 @@ runOrExitImmediately
   -> RAcquire PierEnv (Serf, Log.EventLog)
   -> Bool
   -> MVar ()
+  -> [Ev]
   -> RIO PierEnv ()
-runOrExitImmediately vSlog getPier oExit mStart = do
+runOrExitImmediately vSlog getPier oExit mStart injected = do
   rwith getPier (if oExit then shutdownImmediately else runPier)
  where
   shutdownImmediately :: (Serf, Log.EventLog) -> RIO PierEnv ()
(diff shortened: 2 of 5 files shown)


pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -221,6 +227,17 @@ pillFromURL = PillSourceURL <$> strOption
 pierPath :: Parser FilePath
 pierPath = strArgument (metavar "PIER" <> help "Path to pier")
 
+injectEvents :: Parser [Injection]
+injectEvents = many (InjectOneEvent <$> strOption ( short 'I'
```
pilfer-pandex, line 231: Why not `$`?

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -221,6 +227,17 @@ pillFromURL = PillSourceURL <$> strOption
...
+                          <> long "inject-event"
+                          <> metavar "JAM"
+                          <> help "Path to a jammed event")  <|>
```
pilfer-pandex, line 234: `<|>` should be at start of next line

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -221,6 +227,17 @@ pillFromURL = PillSourceURL <$> strOption
...
+                     InjectManyEvents <$> strOption (
```
pilfer-pandex, line 235: `(` should be on next line, lined up with `<>`, by analogy with
```
[ blah
, blah
]

File: pkg/hs/urbit-king/lib/Urbit/King/Main.hs [hunk 4]
```diff
@@ -586,11 +588,28 @@ runShip (CLI.Run pierPath) opts daemon = do
...
       tryPlayShip
         (CLI.oExit opts)
         (CLI.oFullReplay opts)
         (CLI.oDryFrom opts)
         mStart
+        injections
+
+    loadInjections :: [CLI.Injection] -> RIO PierEnv [Ev]
+    loadInjections injections = do
+      perInjection :: [[Ev]] <- forM injections $ \case
```
pilfer-pandex, line 601: just use `for`; `forM` dates to before Applicative was a superclass of Monad.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -183,7 +183,7 @@ bootNewShip pill lite ship bootEv = do
   let logPath = (pierPath </> ".urb/log")
 
   rwith (Log.new logPath ident) $ \log -> do
-    logInfo "Event log onitialized."
+    logInfo "Event log initialized."
```
pilfer-pandex, line 186: Are you sure? "Onitialized" here means that's it's been turned on and initialized; you lose a shade of meaning.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -360,6 +361,29 @@ pier (serf, log) vSlog startedSig = do
   tExec <- acquireWorker "Effects" (router slog (readTQueue executeQ) drivz)
   tDisk <- acquireWorkerBound "Persist" (runPersist log persistQ execute)
 
+  -- Now that the Serf is configurated, the IO drivers are hooked up, their
```
pilfer-pandex, line 364: "configurated"

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -360,6 +361,29 @@ pier (serf, log) vSlog startedSig = do
...
+  -- starting events have been dispatched, and the terminal is live, we can now
+  -- handle injecting events requested from the command line.
+  unless (null injected) $ rio $ do
+    for_ (zip [1..] injected) $ \(num, ev) -> do
```
pilfer-pandex, line 368: I think the unless here is unnecessary. If the list is empty, `for` won't do anything.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -360,6 +361,29 @@ pier (serf, log) vSlog startedSig = do
...
+      logTrace $ display @Text ("Injecting event " ++ (tshow num) ++ " of " ++
```
pilfer-pandex, line 369: you're aware that logTrace is actually the HIGHEST priority, right?

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
```diff
@@ -586,11 +588,28 @@ runShip (CLI.Run pierPath) opts daemon = do
...
+          CLI.InjectOneEvent filePath -> do
+            logInfo $ display $ "boot: reading injected event from " ++
+              (pack filePath :: Text)
```
pilfer-pandex, line 604: isn't `pack` always `Text`?

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -221,6 +227,17 @@ pillFromURL = PillSourceURL <$> strOption
...
+                        long "inject-event-list"
+                     <> metavar "JAM LIST"
```
pilfer-pandex, line 237: maybe put an _ or  - in the metavar?

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -221,6 +227,17 @@ pillFromURL = PillSourceURL <$> strOption
...
+                     <> help "Path to a jammed list of events"
```
pilfer-pandex, line 238: I think you want these options to be `hidden` so they're not listed in the main usage string, only in the extended list of options.

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
```diff
@@ -77,6 +78,11 @@ data Nat
   | NatNever
   deriving (Show)
 
+data Injection
+  = InjectOneEvent   FilePath
+  | InjectManyEvents FilePath
+  deriving (Show)
```
pilfer-pandex, line 84: why is it necessary to support both of these?

pilfer-pandex (changes requested):
Mostly style, but you do need to make the new cli options hidden.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
eglaysher: Reverted.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
eglaysher: Yes. This is to match the printing of major events like "Resume Ship" above. Do you disagree that this warrants this amount of loudness?

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
eglaysher: We need the cast here because filePath is a FilePath, like all the times in this file where we have this exact same cast on `pack`.

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
eglaysher: Good catch. I keep forgetting about this since that's a weird default behaviour.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
pilfer-pandex: wat. I was not being serious

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
pilfer-pandex: For each event? What if there are a lot?

pkg/hs/urbit-king/lib/Urbit/King/CLI.hs
eglaysher: `-I` is a backwards compatibility concern with vere and there was a TODO to implement it, but it's also not exactly what you want for the use case we're using in production, where we want to give a sequence of events in a row for configuration reasons.

If we were starting from scratch, I wouldn't put `-I` in, but we aren't so I am.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
eglaysher: Please do not do that in a code review.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
eglaysher: Then it will give you sufficient debugging information about which event in the sequence errored.

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
pilfer-pandex: Elliot, pack is a function String -> Text, and FilePath = String. There are no casts in Haskell in the sense that there are in C++.

eglaysher: Feel free to try it locally. It does not compile, nor do any of the other instances where this occurs.

pilfer-pandex (approved):

eglaysher:
@joemfb , do you want to review this? I know you had opinions about how I was previously doing the event injection by putting in the boot sequence in `new`. This way doesn't do that and instead allows injection during each `run`.

pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs
```diff
@@ -360,6 +361,28 @@ pier (serf, log) vSlog startedSig = do
...
+        cb :: WorkError -> IO ()
+        cb = \case
+          RunOkay _         -> putMVar okaySig (Right ())
+          RunSwap _ _ _ _ _ -> putMVar okaySig (Right ())
+          RunBail goofs     -> putMVar okaySig (Left goofs)
+
+    io inject
+
+    takeMVar okaySig >>= \case
+      Left goof -> error ("Goof in injected event: " <> show goof)
```
joemfb, line 382: What is the effect of `error` in this case? Does it kill the process?

eglaysher: Yeah, it should unwind everything and kill the process. You can see another `error` earlier in this same function.

joemfb: This is a discrepancy with vere's behavior, which prints a stack trace but otherwise ignores errors here. I don't have strong feelings, and think we should let the hosting requirements drive this behavior, which presumably means killing the process. I just want to flag the discrepancy and confirm the requirements.

Otherwise, I have no issues with this approach (that aren't already captured in the TODO document re: parsing/validation).

eglaysher: @brendanhay Do you have an opinion on the above, whether we should kill a king process when we fail an injected event?

brendanhay: My preference would be an option/knob to choose the behaviour - otherwise I'd stick with vere's current behaviour.

For hosting the only current use-case is injecting non-critical events.

eglaysher: Given that the current vere behaviour is to not die, I went and changed it so that it just logs the error instead of killing the process.

joemfb (approved):
LGTM

PR #3770, opened 2020-10-21
chat: hoon eval whitespace, unread marker padding

matildepark:
1. Multi line hoon evaluations now have `white-space: pre` to preserve line breaks, etc.
2. Unread marker has a little bit more padding to even out its location in the log (was slightly too close to preceding messages). Still some weird cases when between authors but we may just have to equalise last message / next author padding amounts. (Also is rewritten in indigo-react.)

image


liam-fitzgerald (approved):
LGTM!

PR #3772, opened 2020-10-21
chat: performance

tylershuster:
The code is still a bit messy but the ideas in this are:

- [x] Rewrite sigils to use `Blob` URLs, one per sigil to allow caching and reduce render size
- [x] Indigoify along the way
- [x] Before/after profiling to make sure this use useful

### DIFF
--- a/pkg/interface/src/logic/lib/sigil.js
+++ b/pkg/interface/src/logic/lib/sigil.js
@@ -15,40 +15,76 @@ export const foregroundFromBackground = (background) => {
   return ((whiteBrightness - brightness) < 50) ? 'black' : 'white';
 };
 
-export const SigilURL = memoize(({ patp, size, icon, color}) => {
+const sigilBlobUrls = new Map();
+
+export const sigilBlobUrl = (vals) => {
+  const { patp, size, icon, color } = vals;
+  const key = `${patp}-${size}-${color}${icon ? `-icon` : ''}`;
+  if (sigilBlobUrls.has(key)) {
+    return sigilBlobUrls.get(key);
+  } else {
+    const url = SigilURL(vals);
+    sigilBlobUrls.set(key, url);
+    return url;
+  }
+};
+
+export const SigilURL = memoize(({ patp, size, icon, colors }) => {
   const sigil = sigiljs({
     patp,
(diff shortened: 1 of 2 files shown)


tylershuster:
There's now a global `Map` that keeps track of every unique sigil and the blob URL for it. It doesn't do garbage collection, but I mean it can only be 4 billion entries, right?

tylershuster:
Note that search is a *little* funky because when you search for something, the browser hits the unrendered version of it, then jumps to it, forcing a rerender and the node no longer exists. I'm hesitant to make a `PureComponent` version of the rendered text that can be shuffled in and out of containers, but I think it's the only way to get really nice search. Would appreciate impressions.

tylershuster:
I'm ready for other people's perception of this, whether whole or in piecemeal.

I can't generate objective data on performance (but I'm open to suggestions) but it *feels* faster, and it allows searching the entire loaded log of text, even if it is a little "bouncy" — i.e. you might have to do the search a few times as the component renders in. I think it's a good start at least.

pkg/interface/src/logic/lib/sigil.js
```diff
@@ -33,18 +63,14 @@ export const Sigil = memo(({ classes = '', color, foreground = '', ship, size, s
...
-        renderer: reactRenderer,
-        size: innerSize,
-        icon,
-        colors: [
-          color,
-          foregroundColor
-        ],
-        class: svgClass
-      })}
+      <img
```
matildepark, line 68: `BaseImage` in indigo-react? :p

matildepark (changes requested):
Sigils aren't looking quite right here. 

image

There's a strange behaviour where chat scrolling is slower the more messages there are; small DMs are zippy but UC rooms are pretty sluggish to scroll through. Unsure how to address that though.

tylershuster:
Ok, let's just go with this. If it works it works. Much fewer nodes with the SVGs as images.

tylershuster:
@matildepark

matildepark:
Will test.

matildepark:
Sigils still quite wonky, need padding and the surrounding border often pokes out.

Compare livenet vs. dev:

image


image

image

PR #3773, opened 2020-10-21
chat: create dm route, restore participants option

matildepark:
- Places DM creation / redirection logic at `/~landscape/dm/:ship`.
- Pulls 'allStations' prop plumbing out of all the chat components, remove DM creation logic from profile overlay, just redirects to the DM route.
- Restores "Send Message" option in Participants list.

**Still to be done**:

- A general 'DM creation' prompt — part of me wondered if it would be easier to just weave in "Direct Message" instead of "Chat" when in NewChannel.tsx in the home workspace, but I don't know how to make Formik less aggressive about required values (much as it is with Publish).

That is, we want it to only prompt us to name the chat if there's multiple members in the direct message; if there's only one participant, we want to just ship it over to the "new / redirect DM" route.

I also thought we could write a Popover for inputting a ship? Unsure atm how we could most succinctly write a modal to take a single ship and redirect us.

### DIFF
--- a/pkg/interface/src/views/apps/chat/ChatResource.tsx
+++ b/pkg/interface/src/views/apps/chat/ChatResource.tsx
@@ -114,7 +114,6 @@ export function ChatResource(props: ChatResourceProps) {
         group={group}
         ship={owner}
         station={station}
-        allStations={Object.keys(props.inbox)}
         api={props.api}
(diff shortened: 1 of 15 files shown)


liam-fitzgerald (approved):
LGTM!

Issue #3637, opened 2020-10-02
chat: funky editing behavior


tylershuster:
Start by watching this video:

https://marpem-files.sfo2.digitaloceanspaces.com/radbur-sivmus/Screen%20Recording%202020-10-02%20at%201.32.58%20PM.mov

I was copying text from one chat to another. I had formatted it previously, and I wanted to keep that formatting. I pasted it and began to add the markdown. But after I added the first backtick, the whole thing went haywire as you can see.

matildepark:
Can't replicate.

PR #3774, opened 2020-10-22
chat: fall back to native input on mobile

tylershuster:
In order to get "normal" (expected) text input behavior on mobile, typing needs to occur in a proper input element. 

https://github.com/codemirror/CodeMirror/blob/2571b4ad08e007c246d73bdd973892bf5dd7b02e/src/edit/options.js#L78

As you can see above, codemirror (despite exposing the option) does not allow changing the input type to a textarea on mobile. Therefore, expected behavior is impossible.

This PR detects mobile browsers and falls back to regular HTML input — no code highlighting, no image pasting. Fortunately, those aren't really things we need on mobile.

It uses a JavaScript proxy object so that existing CodeMirror behavior doesn't have to be altered.

The experience is miles ahead. I have tested it on iOS and Android (simulated via Android Studio)

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/chat-editor.js
+++ b/pkg/interface/src/views/apps/chat/components/chat-editor.js
@@ -3,7 +3,7 @@ import { UnControlled as CodeEditor } from 'react-codemirror2';
 import { MOBILE_BROWSER_REGEX } from "~/logic/lib/util";
 import CodeMirror from 'codemirror';
 
-import { Row } from '@tlon/indigo-react';
+import { Row, BaseInput } from '@tlon/indigo-react';
 
 import 'codemirror/mode/markdown/markdown';
 import 'codemirror/addon/display/placeholder';
@@ -35,6 +35,25 @@ const MARKDOWN_CONFIG = {
     linkHref: 'presentation'
   }
 };
+
+// Until CodeMirror supports options.inputStyle = 'textarea' on mobile,
+// we need to hack this into a regular input that has some funny behaviors
+const inputProxy = (input) => new Proxy(input, {
+  get(target, property) {
+    if (property in target) {
+      return target[property];
+    }
+    if (property === 'setOption') {
+      return () => {};
+    }
+    if (property === 'getValue') {
+      return () => target.value;
+    }
+    if (property === 'setValue') {
+      return (val) => target.value = val;
+    }
+  }
+});
 
 export default class ChatEditor extends Component {
   constructor(props) {
@@ -129,7 +148,11 @@ export default class ChatEditor extends Component {
         'Esc': () => {
           this.editor?.getInputField().blur();
         }
-      }
+      },
+      // The below will ony work once codemirror's bug is fixed
+      spellcheck: !!MOBILE_BROWSER_REGEX.test(navigator.userAgent),
+      autocorrect: !!MOBILE_BROWSER_REGEX.test(navigator.userAgent),
+      autocapitalize: !!MOBILE_BROWSER_REGEX.test(navigator.userAgent)
     };
 
     return (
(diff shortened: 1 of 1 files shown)


ohAitch:
There's supposed to be code highlighting in chat??
I will note I did try to paste an image on iOS yesterday, however it didn't work anyway, so taking it out doesn't seem too bad; maybe long-term a textarea onPaste handler would be in order.

tylershuster:
Uh...ok markdown highlighting.

ohAitch:
Ah yup indeed. Can confirm, markdown highlighting works fine on iOS chrome as does otherwise "normal" input behavior, though on safari I have had the selection glitches that presumably motivated this PR. (possibly something about being in desktop mode? It's an iPad)

tylershuster:
Selection glitches about (#3637) on CodeMirror. There's some on their repo too. Basically it's because of *how* controlled CodeMirror is.

This PR is motivated mostly by autocorrect and autocapitalize, honestly. (I should check to make sure this respects system settings) On my iPhone SE, chat is pretty much unusable if I want to keep up my appearance as someone who uses correct punctuation and spelling. There's some other weird behavior too, like holding down backspace, getting stuck in shift (or not locking in caps, I can't remember), and others that all result from not being a true input.

ohAitch:
Yep I definitely hit the holding down backspace thing! Just not on Chrome, for some reason, despite it presumably also having to use the same engine.

matildepark:
> (possibly something about being in desktop mode? It's an iPad)

iPad uses the Mac OS X user agent so you'll probably find the input always autofocusing as you navigate.

pkg/interface/src/views/apps/chat/components/chat-editor.js
```diff
@@ -142,19 +165,38 @@ export default class ChatEditor extends Component {
         paddingTop='8px'
         width='calc(100% - 88px)'
         className={inCodeMode ? 'chat code' : 'chat'}
+        color="black"
       >
-        <CodeEditor
+        {MOBILE_BROWSER_REGEX.test(navigator.userAgent)
```
matildepark, line 170: That said this will basically mean that iPad doesn't have the native HTML input lol. Thanks Apple for making iOS Safari vs. Desktop Safari illegible

ohAitch:
You can always set explicitly mobile mode, at least

tylershuster:
I've confirmed with the author of CodeMirror that this feature will not be coming back.

pkg/interface/src/views/apps/chat/components/chat-editor.js
```diff
@@ -142,19 +165,38 @@ export default class ChatEditor extends Component {
...
+          ? <input
+            type="text"
+            style={{ width: '100%', marginTop: '-8px', background: 'transparent', color: 'currentColor' }}
+            onKeyUp={event => {
```
matildepark, line 174:
```suggestion
           placeholder={inCodeMode ? "Code..." : "Message..."}
            onKeyUp={event => {
```

pkg/interface/src/views/apps/chat/components/chat-editor.js
```diff
@@ -142,19 +165,38 @@ export default class ChatEditor extends Component {
         paddingTop='8px'
         width='calc(100% - 88px)'
         className={inCodeMode ? 'chat code' : 'chat'}
+        color="black"
       >
-        <CodeEditor
+        {MOBILE_BROWSER_REGEX.test(navigator.userAgent)
+          ? <input
```
matildepark, line 171: you can import `BaseInput` from indigo-react for this.

custom.css had logic for whether `code` descended elements would pass Source Code Pro; now Code or Message doesn't change font face on this one, because that declaration is too specific. the BaseInput component could have `mono={inCodeMode}` to compensate. It also should have `fontSize='14px'` to match the chat input, in which case you'll want to ensure it's lined up properly with the sigil, etc.

tylershuster: `mono` doesn't seem to work on `BaseInput`...

matildepark: ohh right I guess that's specific to `Text`. You could do similar with `fontFamily` probably.

matildepark (approved):

ohAitch:
Okay so with this change on iOS chrome, not only is markdown highlighting gone as expected, but also it pops up an autofill bar directly over the input such that it's impossible to see what you're typing QQ
Would you like to fill a password

matildepark:
You can swipe up with your finger to see what you’re typing but I also wonder why safari isn’t doing that when it does that in other inputs.

ohAitch:
I can swipe up with my finger to see what I'm typing, but then release my finger, and not see what I'm typing again

tylershuster:
Looks like we just need to turn autocomplete off. That should be sane. I don't know why it didn't come up in testing.

matildepark:
> I can swipe up with my finger to see what I'm typing, but then release my finger, and not see what I'm typing again

This is definitely not acceptable, but my question is — why doesn't iPhone safari do this? It works perfectly fine there. I will try and reproduce at some point today.

> Looks like we just need to turn autocomplete off.

People do use that, though.

tylershuster:
> I think the fix here is to rise above CodeMirror

matildepark:
This issue seems to be with the native input?

tylershuster:
@ohAitch What device is that?

ohAitch:
iPad Pro 2

The safari presumably doesn't do that because it doesn't unconditionally pop up an autofill bar - which behavior I don't love from chrome but haven't figured out any way to turn off.

tylershuster:
Apparently Chrome handles `autocomplete` attributes differently.

@matildepark I don't think people *do* use autocomplete here — I don't think this is predictive typing, this is "look an input that you've previously entered text in; how about this text? (can also be credit cards)"

I'll try and work on an autocomplete-disabled version

matildepark:
Oh, I see.

PR #3775, opened 2020-10-22
clay: add %ce scry for static dais

philipcmonk:
If you know the type of a mark statically, then you can skip all the
vases and get a +says ("static dais").  This does not work if you don't
know the type at compile time since we don't polymorphize that way, so
in that case continue to use a +dais.

With this preamble:

```
=orig ~['line1' 'line2' 'line3']
=diff [%& 1] [%| ~['line2'] ~['newline2']] [%& 1] ~
```

With a dais, applying a diff looks like:

```
=/  dais  .^(dais:clay %cb %/txt)
!<(wain (~(pact dais !>(orig)) !>(diff)))

<|line1 newline2 line3|> 
```

And with a says:

```
=/  says  .^((says:clay wain (urge:clay cord)) %ce %/txt)
(~(pact says orig) diff)

<|line1 newline2 line3|> 
```

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:808cc6a28842ab8cde540de14cf0663ef3c6e7cff7c4c5bf5bbb0bb43ec1894d
-size 6329173
+oid sha256:3e4c7cbdf0d3115c3e507a63a6c0a1a7c275c02eed71adb8395e3df9d82487aa
+size 19047616
--- a/pkg/arvo/mar/txt.hoon
+++ b/pkg/arvo/mar/txt.hoon
@@ -23,6 +23,7 @@
 ++  grad
   |%
   ++  form  %txt-diff
+  ++  dorm  (urge cord)
   ++  diff
     |=  tyt/wain
     ^-  (urge cord)
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -611,6 +611,45 @@
         ^+  sam
         [p:bunt noun]
       --
+    ::  +get-says: get staticized mark definition
+    ::
+    ::  TODO: cache?
+    ::
+    ++  get-says
+      |=  mak=mark
+      ^-  [vase state]
+      =^  cor=vase  nub  (build-fit %mar mak)
+      =^  =dais  nub  (get-mark mak)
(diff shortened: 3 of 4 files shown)


ohAitch:
And for our next trick, static mark conversion gate scrys :)

Looks like the implementation still has a bunch of !< internally though,
any chance of eliding those and compiling the says from the mark cores
directly?

philipcmonk:
> And for our next trick, static mark conversion gate scrys :)

Yes, that would be good.

> Looks like the implementation still has a bunch of !< internally though,
> any chance of eliding those and compiling the says from the mark cores
> directly?

Yes, but then we should refactor +dais in terms of +says if possible. That's probably worth doing, to avoid a few nest checks.  @belisarius222 might see a cleaner way to do it, though.

belisarius222:
Hmm, yeah the performance will likely be better if we always start by building a `$says` (maybe I suggest renaming it, maybe to `$nave`?), and then transform that into a `$dais` if needed.  I think we want to cache both.  We should do the same for `$tube`, at least eventually.

If we add a cache line for this, though, I doubt the performance will be that much worse, so it might be worth putting in this way at first.

Huh, it turns out `+volt:dais` is never called.  It's unsafe anyway, so let's remove it.

belisarius222:
`$dais` is a door for mostly historical reasons; it would be better if it was a stateless `|%` core.  Only `+pact` takes a marked datum as an unambiguous initial argument, so I don't think it's worth turning the whole interface core into a closure around a particular value.  Here's a stab at a `|%` version of `$nave` for comparison.

```hoon
  +$  nave
    |$  [typ dif]
    $_
    |%
    ++  bunt  *typ
    ++  diff  |~(old=val new=val *dif)
    ++  form  *mark
    ++  join  |~([a=dif b=dif] *(unit (unit dif)))
    ++  mash
      |~  [a=[ship desk dif] b=[ship desk dif]]
      *(unit dif)
    ++  pact  |~([typ dif] *typ)
    ++  vale  typ
    --
```

ohAitch:
Surprised volt is never used, is it being done manually now that ford has
been merged to clay?

belisarius222:
@ohAitch it was only ever used in Ford and Clay ... if I remember right, it was used for constructing a `$page` (`[mark *]`) by applying deltas, after which we would assume the value came out as a valid value of that mark.  Now instead of assuming, we validate that when we do it.  See `+lobe-to-page` in Clay.

I doubt it's worth adding `+volt` just to optimize that operation, especially since it's hard to be completely sure an old delta will always be valid -- it's not a bad sanity check.

philipcmonk:
superseded by #4491

PR #3776, opened 2020-10-22
landscape: add 'admin only' check to groupify form

matildepark:
Part of #3704 work — just only lists groups we admin in the groupify list if we pass `adminOnly` as a prop.

Side note: Currently only Publish has a groupify form at all, so might want to take a sec and add it to the other unmanaged settings pages again...


liam-fitzgerald (approved):
LGTM!

PR #3779, opened 2020-10-22
Home tiles

tylershuster:
Depends on https://github.com/urbit/urbit/pull/3767, which is included here for an accurate rendering of the clock

@urcades this was pretty easy to implement. Would appreciate any feedback.

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -37,8 +37,7 @@ export default class LaunchApp extends React.Component {
           <Box
             ml='2'
             display='grid'
-            gridAutoRows='124px'
-            gridTemplateColumns='repeat(auto-fit, 124px)'
+            gridTemplateColumns='repeat(auto-fill, minmax(128px, 1fr))'
             gridGap={3}
             p={2}
           >
@@ -75,12 +74,9 @@ export default class LaunchApp extends React.Component {
               location={props.userLocation}
               weather={props.weather}
             />
+            <Box display={["none", "block"]} width="100%" gridColumn="1 / -1"></Box>
+            <Groups groups={props.groups} associations={props.associations} invites={props.invites} api={props.api}/>
           </Box>
-          <Groups
-            associations={props.associations}
-            groups={props.groups}
-            invites={props.invites}
-            api={props.api} />
         </Box>
         <Box
           position="absolute"
--- a/pkg/interface/src/views/apps/launch/components/Groups.tsx
+++ b/pkg/interface/src/views/apps/launch/components/Groups.tsx
@@ -35,17 +35,7 @@ export default function Groups(props: GroupsProps & Parameters<typeof Box>[0]) {
   };
 
   return (
-    <Box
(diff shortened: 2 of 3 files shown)


matildepark:
Conflict on launch/app.js ?

tylershuster:
resolving

tylershuster:
I should comment that this turns all tiles into a single grid — i.e. no separation between home/settings/dojo/clock/weather and groups.

matildepark:
Conflict again but is probably my fault this time.

> I should comment that this turns all tiles into a single grid — i.e. no separation between home/settings/dojo/clock/weather and groups.

Idk, we want this on mobile but not on desktop. In indigo components we usually do this eg. `display={["none", "block"]}` for breakpoint declarations, where 'none' is below 550px and 'block' is above. But in this case you'd essentially want the wrapper to not wrap on mobile. Maybe making it `inline` on mobile breakpoint?

urcades:
+1ing the sentiment that evening out the display of tiles was always
intended to be something we enforced on mobile-sized viewports, less so in
desktop-sized views.

Also, the separation of utility (single player) tiles from groups was
always intentional — we should try to retain this separation.

—
~fabled-faster
https://urbit.org/

tylershuster:
Alright, how about this? Just adds a blank box that spans all columns when it's visible.

matildepark:
@urcades @g-a-v-i-n 

mobile

Screen Shot 2020-10-27 at 1 35 22 PM

desktop

Screen Shot 2020-10-27 at 1 35 32 PM

I'm pretty sure what we want is for all layouts but mobile to look the same as livenet, with the mobile layout just slightly bigger tiles so they're level. My second question is "are the mobile tiles too big here"...

urcades:
I finally managed to pull this and check it out and I think the sizing of the tiles on mobile is fine. It's a perfect solution to ensuring the fit + finish of the mobile interface is at least  minimally up to par with the way the homescreen renders on larger devices.

And I want to emphasize, this was always the motivating factor to getting the grid 'balanced' – was for mobile to look not-lopsided. That's it!

With that statement in mind, when I checked out the homescreen on my laptop-sized device, I think it was quite clear the tiling scheme being used makes the home tiles quite large:

Screen Shot 2020-10-28 at 4 37 03 PM

Given that the impetus for grid-correcting was really to aid our mobile interface, I'd be fine leaving our current gridding scheme intact with 128px-sized 'large viewport' tiles, and a responsive two-column grid being enacted on mobile-sized devices.

The only exception here is for very, very small viewports, such as the iPhone 5S:

Screen Shot 2020-10-28 at 4 38 13 PM

This should still be two columns!

tylershuster:
Ok, I made a mistake here and left the tiles at `10rem` squares, which is definitely too big. I changed it back to `128px` and cleaned up the commits. I think you'll find that this addresses your concerns.

urcades:
Yeah! This is totally it.

—
~fabled-faster
https://urbit.org/

urcades (approved):
Big ol stamp

PR #3783, opened 2020-10-23
clock: svg with mask

tylershuster:
mask that donut


matildepark:
The changes here look smaller than #3767 somehow.

tylershuster:
oh shoots

tylershuster:
well damn it I'm having quite the time

PR #3784, opened 2020-10-23
graph-js: normalize keys in reducer

liam-fitzgerald:
Pre-graph-store links that were brought over use @da as their index,
however new links use a unix timestamp as their index. Pending a proper
fix, we instead just normalize all the indexes to be unix timestamps
inside the reducer.

### DIFF
--- a/pkg/interface/src/logic/reducers/graph-update.js
+++ b/pkg/interface/src/logic/reducers/graph-update.js
@@ -1,10 +1,20 @@
 import _ from 'lodash';
 import { OrderedMap } from "~/logic/lib/OrderedMap";
 
+const DA_UNIX_EPOCH = 170141184475152167957503069145530368000;
+const normalizeKey = (key) => {
+  if(key > DA_UNIX_EPOCH) {
+    // new links uses milliseconds since unix epoch
+    // old (pre-graph-store) use @da
+    return (1000 * (9223372036854775 + (key - DA_UNIX_EPOCH))) / 18446744073709551616;
+  }
+  return key;
+}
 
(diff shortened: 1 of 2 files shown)


joemfb:
Why aren't we using `@da` anymore? Seems fairly hostile to the rest of the system.

matildepark:
/me shrugs — outside the scope of this PR, which is a FE shim fix to fix Links before making a call on a back-end fix to standardise these times.

matildepark (approved):

liam-fitzgerald:
@joemfb We can't reliably do any manipulation of @da on the frontend, because even discarding the 64bits of subsecond timestamp, we don't have enough room in the JS number type (double, so 53 bits) to accurately manipulate and display timestamps. BigInt support just dropped in iOS 14, so we might be able to use `@da` in the future, but I don't think iOS 14 has sufficient traction yet to warrant a change. We usually do `@da` to timestamp conversion during the conversion to JSON, but we can't do this with graph-store, because it's not guaranteed that indexes are necessarily dates. (.e.g. publish uses the revision number as the index)

ohAitch:
Are there BigInt polyfills? Would it make any sense to pass da around as
{seconds,subseconds} objects?

matildepark:
image

Maybe [this Babel plugin](https://www.npmjs.com/package/@babel/plugin-syntax-bigint)?

liam-fitzgerald:
Well, looks like I'm gonna have to fix this in order to ship notifications. Will pull in a polyfill for bigint and add some utility functions for formatting and parsing @da

matildepark:
:/ OK ganbatte

Issue #3778, opened 2020-10-22
ames: fake ships are subject to forwarding loops


joemfb:
I booted a fake zod/binzod/master-morzod (off of `release/next-vere`). After a couple minutes, my fake ~zod had forwarded over 400K packets.

belisarius222:
As @joemfb and I discussed a little while ago, we should be able to fix this by adding a filter to Vere that rejects packets sent by our own process.  I think we could do this by filtering incoming packets from loopback on the port that we bound.

philipcmonk:
I mean, who's to say they were sent by our own process.  The thing to do is to find out what exactly those packets are and trace who sent/forwarded them and why.  For example, are stars and planets trying to forward any packets?  If so, something earlier was broken.

belisarius222:
Hmm ... yeah I think for some reason I was under the misapprehension that UDP loopback would just broadcast, which would inevitably cause a forwarding loop.

I feel like these loops usually happen when one ship assumes a different ship's former IP and port.  Maybe when we hear a new lane for a ship, we should make sure to dissociate that lane from any other ships?

joemfb:
I added debug printfs to #3764, and the cause was fairly obvious: galaxies can end up forwarding to themselves when they're forwarding on behalf of a ship they sponsor and don't have a route to that ship. I updated #3764 with to filter this case out of that runtime, but I believe the scry handler should be updated as well. I'll open a separate PR.

philipcmonk:
Sounds like this could be happening on mainnet, is that right?  Would it have happened pre-stateless forwarding?

joemfb:
This could definitely be happening on mainnet, but only with stateless forwarding I think:

https://github.com/urbit/urbit/blob/e3dcfc637291d922376eb84df87e9bfa6f9e8e7d/pkg/arvo/sys/vane/ames.hoon#L1564-L1569

ohAitch:
There's generally a sense of "closer", where you go from a ship to its
parent star to its parent galaxy to the destination galaxy to the
destination star to the destination ship, right? Where you can skip across
hops if you have a lane(esp if any two in that list are the same
intermediary), but should never go backwards or stay in place.

ohAitch:
Destination < prefix of that destination < prefix followed by unrelated
bytes < prefix followed by more unrelated bytes < shorter prefix down to
zero (< same shorter prefix followed by unrelated bytes, up to a moon of a
planet of a star of a completely different galaxy)

[...]

PR #3786, opened 2020-10-23
ames: updates forward-lane scry-handling to prevent routing loops

joemfb:
To help avoid routing loops (see #3778), this PR updates the %ames' `/peer/<ship>/forward-lane` scry handler to exclude ourselves from possible routing targets. To precisely match the forwarding semantics in `+send-blob:event-core`, it also produces both the peer's and peer's sponsoring galaxy's routes, when the route to the peer is indirect.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -785,23 +785,35 @@
       ~  ``noun+!>(u.peer)
     ::
         [%forward-lane ~]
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
LGTM

Fang- (approved):
Concur'd. Wondering if this and `+send-blob` should just share the logic (though some utility function), but not a pressing concern.

PR #2949, opened 2020-05-28
hoon: add doq, kel, ker, pal, pam, par, pat, sel, ser, soq, tic

jfranklin9000:
update "XX deprecated" comments to match:
      https://urbit.org/docs/tutorials/hoon/hoon-syntax/#reading-hoon-aloud
      restore mnemonics

Replaces https://github.com/urbit/urbit/pull/2930.

I will add `bus` to `buc`, etc commits as I finish them, but you can pull at any time.

PR #3787, opened 2020-10-23
hoon: update glyph names

philipcmonk:
Resolves conflicts in #2949, thanks to @jfranklin9000 for all the actual work.  This has languished since May since it was difficult to OTA correctly.  Since then, Ford Fusion made this sort of update easy and reliable, so I tested the OTA now and it works fine.

For the merge I did the dumbest possible thing, which leaves two notable areas using the old parser names: lib/language-server/parser and clay, since at that time the ford parsers were in ford.hoon.  The old parser names still exist, so maybe this is fine, else we could update those.


joemfb (approved):
LGTM

PR #3789, opened 2020-10-23
publish: don't insert newline

### LINKED ISSUE #3596: publish: editor inserts newline at beginning of document
It seems to be only upon editing, and the newline does not appear upon merely viewing, but to reproduce:

1) Create a document in publish.
2) Save
3) Edit document
4) Note newline at beginning of editor.

tylershuster:
Fixes #3596

Not sure if this needs to be done in `<Note />` as well, just assumed that it does because it draws from the same thing.

### DIFF
--- a/pkg/interface/src/views/apps/publish/components/EditPost.tsx
+++ b/pkg/interface/src/views/apps/publish/components/EditPost.tsx
@@ -14,7 +14,7 @@ interface EditPostProps {
 
 export function EditPost(props: EditPostProps & RouteComponentProps) {
   const { note, book, noteId, api, ship, history } = props;
-  const body = note.file.slice(note.file.indexOf(";>") + 2);
+  const body = note.file.slice(note.file.indexOf(";>") + 3);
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/apps/publish/components/Note.tsx
```diff
@@ -45,7 +45,7 @@ export function Note(props: NoteProps & RouteComponentProps) {
 
   const comments = note?.comments || [];
   const file = note?.file;
-  const newfile = file ? file.slice(file.indexOf(";>") + 2) : "";
+  const newfile = file ? file.slice(file.indexOf(";>") + 3) : "";
```
matildepark, line 48: Imagine a world where the fix in EditPost is applied and people start using it in their posts, no longer having the newline at the start. Won't this then cut off the first character, if we do this in both places?

tylershuster: Good point

matildepark (approved):

PR #3790, opened 2020-10-23
weather: fix api call

ryjm:
Keeping my one line fix streak alive.


matildepark (approved):
Oh. Well, good catch. (Apparently automatic detection was quietly broken.)

PR #3793, opened 2020-10-23
ames: validate cyphertext length

belisarius222:
When we receive an Ames packet, we've been malloc()ing a buffer of whatever size that packet requests, even before trying to decrypt it.  Our AES jets should be more careful, but we can also fix this particular problem now by validating the relevant sizes in Hoon.

This should eventually be moved to the `|acru` cryptosuite core, but I think we're better off doing that in a later refactor.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -3061,14 +3061,20 @@
   (en:crub:crypto symmetric-key (jam plaintext))
 ::  +decrypt: decrypt packet content to a $shut-packet or die
 ::
+::    TODO: alter |acru interface
+::
 ++  decrypt
   |=  [=symmetric-key ciphertext=@]
   ^-  shut-packet
+  =+  ;;([iv=@ len=@ cyf=@] (cue ciphertext))
+  ::  assert cyphertext length <= claimed length <= max bytes
   ::
-  ;;  shut-packet
-  %-  cue
-  %-  need
-  (de:crub:crypto symmetric-key ciphertext)
+  ?.  (lte (met 3 cyf) len)
+    ~|(ames-len-small+len^(met 3 cyf) !!)
+  ?.  (lte len 1.024)
+    ~|(ames-len-large+len !!)
+  ;;  shut-packet  %-  cue  %-  need
+  (~(de sivc:aes:crypto (shaz symmetric-key) ~) iv len cyf)
 ::  +encode-packet: serialize a packet into a bytestream
 ::
 ++  encode-packet


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3061,14 +3061,20 @@
...
-  ::
-  ;;  shut-packet
-  %-  cue
-  %-  need
-  (de:crub:crypto symmetric-key ciphertext)
+  =+  ;;([iv=@ len=@ cyf=@] (cue ciphertext))
+  ::  assert cyphertext length <= claimed length <= max bytes
+  ::
+  ?.  (lte (met 3 cyf) len)
+    ~|(ames-len-small+len^(met 3 cyf) !!)
```
joemfb, line 3073: This first assertion is actually redundant, the AES-CTR implementations already include it:

https://github.com/urbit/urbit/blob/ca3872fba46660e3638b377cdf4bd0f64b535a7f/pkg/arvo/sys/zuse.hoon#L3021

The absence of the same in the AES-SIV jet is a mismatch (as is the length truncation to `c3_w`).

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3061,14 +3061,20 @@
...
+  ?.  (lte len 1.024)
+    ~|(ames-len-large+len !!)
```
joemfb, line 3075: I think this would be clearer in `bloq=13`, to match the other packet-length operations in %ames.

joemfb (changes requested):

pkg/arvo/sys/vane/ames.hoon
belisarius222: I could do
```
?.  (lte len ^~((bex (sub 13 3))))
```
so that someone could change the packet length by string-replacing `13` with some other number.  Let me know if you know a better way to express this.

pkg/arvo/sys/vane/ames.hoon
belisarius222: Ah, ok so it sounds like this first assertion should be removed from Ames, relying instead on Zuse's crypto code to enforce this.  That sounds ok.  I'll remove this, then.

pkg/arvo/sys/vane/ames.hoon
joemfb: I don't think you need to check `len` at all here, I'd just `?>  (lte (met 13 cyf) 1)` or even `?>  ?=(?(%0 %1) (met 13 cyf))`.

joemfb: The decryption call validates the relationship between `len` and `cyf`.

pkg/arvo/sys/vane/ames.hoon
joemfb: To be clear, the length assertion is not reflected in the AES-SIV jet, which is an active mismatch.

belisarius222:
Superseded by https://github.com/urbit/urbit/pull/3894 and https://github.com/urbit/urbit/pull/3903.

PR #3794, opened 2020-10-23
u3: remove timestamp printing in u3v_time()

joemfb:
This PR is a substantial performance improvement for the vere king.

On every turn of the libuv event loop, we update a global `@da` timestamp, so that all i/o drivers can have a consistent view of time. The function that we use to do so (`u3v_time()`) was also formatting that same timestamp, through a nock call to hoon's atom (`$coin`) printer. That's an expensive call, with no relevant high level jets (at the moment, see #2853). But, even worse, that formatted timestamp was never read or used anywhere.

I discovered this issue while debugging the network forwarding loop reported in #3778. So I used the loop as a simple performance test. As the following flame graph makes clear, samples were dominated by `u3v_time()`.

before

Those samples were taken roughly every millisecond, for around 90 seconds. In that time, my fake ~zod forwarded over 400K packets.

Compare with timestamp printing removed:

after

In those roughly 30 seconds, my fake ~zod forwarded over 2 million packets.

### DIFF
--- a/pkg/urbit/include/noun/vortex.h
+++ b/pkg/urbit/include/noun/vortex.h
@@ -10,7 +10,7 @@
         c3_d    ent_d;                    //  event number
         u3_noun yot;                      //  cached gates
         u3_noun now;                      //  current time, as noun
-        u3_noun wen;                      //  current time, as text
+        u3_noun wen;                      //  current time, as text, XX remove
         u3_noun sev_l;                    //  instance number
         u3_noun sen;                      //  instance string
         u3_noun roc;                      //  kernel core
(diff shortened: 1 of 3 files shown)


belisarius222:
lol

philipcmonk (approved):
Very nice find!  This combined with #3786 will make a huge difference for infrastructure ships.  I'm optimistic that between the two we'll see ~zod's cpu usage below 100% for the first time in months.

This is definitely worth testing on ~zod and/or ~nus, and maybe worth pushing a release before too long.

matildepark:
"We stopped printing a timestamp we don't use and got a 1,400% performance improvement for forwarding packets"

jtobin:
hot like fire

Issue #3705, opened 2020-10-14
Leap results should prioritize exact matches


Fang-:
I'm in a couple groups with "Tlon" in the name. One of them is called _just_ "Tlon". When I type "Tlon" into leap, I'd expect the exact match to come up first, however:

<img width="176" alt="image" src="https://user-images.githubusercontent.com/3829764/95966643-ac1ddf00-0e0b-11eb-8565-029290b4bd46.png">

It lists the longer names first. This is incorrect from both a "best match" and "alphabetical order" perspective.

Additionally, when I'm already viewing the "Tlon" group, it shows the "create channel" action as well. And as a first option, even though it's a relatively rare action compared to just navigating around.

PR #3795, opened 2020-10-23
leap: return 'other' first, lowercase groups

matildepark:
- Return 'other' results first because "Home" was getting buried by channels when those types of results should be returned first (though exact results should also always be weighted, see #3705)
- Searching for group names was accidentally case sensitive; now it's not.


liam-fitzgerald (approved):
LGTM!

PR #3796, opened 2020-10-23
chat: disable autocomplete on mobile

tylershuster:
@ohAitch I can't test this without an iPad — simulator won't let me download an app. Can you try this?

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/chat-editor.js
+++ b/pkg/interface/src/views/apps/chat/components/chat-editor.js
@@ -172,6 +172,7 @@ export default class ChatEditor extends Component {
             fontSize="14px"
             style={{ width: '100%', background: 'transparent', color: 'currentColor' }}
             placeholder={inCodeMode ? "Code..." : "Message..."}
+            autoComplete={new RegExp(/Chrome/).test(navigator.userAgent) ? 'disabled' : 'off'}
             onKeyUp={event => {
               if (event.key === 'Enter') {
                 this.submit();
--- a/pkg/interface/src/views/apps/chat/components/chat-editor.js
+++ b/pkg/interface/src/views/apps/chat/components/chat-editor.js
@@ -172,7 +172,7 @@ export default class ChatEditor extends Component {
             fontSize="14px"
             style={{ width: '100%', background: 'transparent', color: 'currentColor' }}
             placeholder={inCodeMode ? "Code..." : "Message..."}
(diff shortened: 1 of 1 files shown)


matildepark:
Idk how he would be able to actually. But I can give it a shot on my own iPad. Let me, uh, install iPad chrome I guess.

tylershuster:
way to take one for the team

ohAitch:
(Sorry, not sure how, my urbits come in "hosted" and "docker, cli, haven't figured out how to get at the http ports")

matildepark:
0A7B3E91-5825-42E5-B6ED-F1C117519B06_1_101_o

No dice.

pkg/interface/src/views/apps/chat/components/chat-editor.js
```diff
@@ -172,6 +172,7 @@ export default class ChatEditor extends Component {
             fontSize="14px"
             style={{ width: '100%', background: 'transparent', color: 'currentColor' }}
             placeholder={inCodeMode ? "Code..." : "Message..."}
+            autoComplete={new RegExp(/Chrome/).test(navigator.userAgent) ? 'disabled' : 'off'}
```
matildepark, line 175: I think this doesn't work because Chrome [uses nearly the same user agent string as Safari](https://chromium.googlesource.com/chromium/src/+/master/docs/ios/user_agent.md). Try catching `CriOS` instead.

tylershuster: CriOS, desTROyer of wORldS

matildepark:
OK well, still nothing — 

85D29E66-54A7-44F7-8A26-1D518819CF3C_1_105_c

Since this is Chrome on iOS, which is just Safari on iOS — does it [even support this attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete)? It seems like it might not.

ohAitch:
Any chance of just detecting CriOS and reverting to codemirror? ^-^

(I noticed I can also no longer Shift-Enter on what is, after all, now a
single-line <input> element)

tylershuster:
Try this out. I changed it to a textarea which seems to solve our problems but can't verify on an iPad.

matildepark:
Afraid not.

IMG_0010

matildepark:
Still no solution here.

tylershuster:
You know I know it. I've been able to reproduce behavior stemming from the same underlying issue on my phone, so that puts us one step closer, but I haven't figured out the magic yet. Obviously it's possible with native elements, but I don't know what yet.

PR #3798, opened 2020-10-24
eyre: produce moves in correct order

Fang-:
I double-confused myself. The comment here made me think I was fine just storing the moves in reverse order, but of course I have to _produce_ them in the correct order. Also Arvo uses sane move order, so the test was wrong to begin with.

https://github.com/urbit/urbit/blob/235bfbbd2faf3184481558bad09108a8f69e91c3/pkg/arvo/sys/vane/eyre.hoon#L1050-L1052


eglaysher (approved):
Makes sense. I've done the move order wrong thing before too.

It would be great if we could solve this entire category of bugs with types, but oh well.

PR #3799, opened 2020-10-24
eyre: ignore facts directly after clog

Fang-:
When an application would send multiple facts during a single event, it was possible for one fact to trigger a clog, removing the subscription and sending a quit, but then subsequent facts still getting sent out at normal.  
Getting facts after receiving a quit seems slightly weird, especially considering clients sending out resubscribes based on the point at which they lost the subscription.

Here, we drop any facts for subscriptions we don't have registered in state, which should only happen in the described case.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1507,7 +1507,6 @@
       ?:  ?&  ?=(%fact -.sign)
               !(~(has by subscriptions.u.channel) request-id)
           ==
-        ~&  [%e %fact-without-subscription channel-id request-id]
         [~ state]
       ::  attempt to convert the sign to json.
       ::  if conversion succeeds, we *can* send it. if the client is actually
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1500,6 +1500,15 @@
       ?~  channel
         :_  state  :_  ~
         [duct %pass /flog %d %flog %crud %eyre-no-channel >id=channel-id< ~]
+      ::  it's possible that this is a fact emitted directly alongside a fact
+      ::  that triggered a clog & closed the subscription. in that case, just
+      ::  drop the fact.
+      ::
+      ?:  ?&  ?=(%fact -.sign)
+              !(~(has by subscriptions.u.channel) request-id)
+          ==
+        ~&  [%e %fact-without-subscription channel-id request-id]
+        [~ state]
       ::  attempt to convert the sign to json.
       ::  if conversion succeeds, we *can* send it. if the client is actually
       ::  connected, we *will* send it immediately.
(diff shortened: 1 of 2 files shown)


belisarius222:
Ames gives Gall a `%clog` sign, which Gall then uses to kill the subscription.  I think Eyre should do that too, since then it shouldn't receive later `%fact`s on the subscription and have to figure out what to do with them.

philipcmonk:
Pretty sure Ames has the same problem, there's no way to stop stuff already
in your move queue. Would be easy to test by pushing out 20 updates in one
invocation to an offline ship.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1500,6 +1500,15 @@
...
         :_  state  :_  ~
         [duct %pass /flog %d %flog %crud %eyre-no-channel >id=channel-id< ~]
+      ::  it's possible that this is a fact emitted directly alongside a fact
+      ::  that triggered a clog & closed the subscription. in that case, just
+      ::  drop the fact.
+      ::
+      ?:  ?&  ?=(%fact -.sign)
+              !(~(has by subscriptions.u.channel) request-id)
+          ==
+        ~&  [%e %fact-without-subscription channel-id request-id]
```
eglaysher, line 1510: I suspect this might be too noisy in practice. Also, what could the user actually do in response to this printf?

eglaysher (approved):
Whether you should send a clog event to gall or not, this part is also probably correct.

Fang-:
> Ames gives Gall a %clog sign, which Gall then uses to kill the subscription.

@belisarius222 looks like gall, on receiving a clog, closes all subscriptions for the specified ship though? So as-is, `%clog` doesn't make sense for this, considering we only care about a single subscription, and the ship is always `our`. Maybe if/when `%clog` becomes more granular, this might make sense?

pkg/arvo/sys/vane/eyre.hoon
Fang-: My thinking was, this should only happen directly after an `%e %clog` printf, and stand out as suspicious when they show up on their own. But you're right it's more worrying than informative, and could become too loud.

belisarius222:
Oh yeah, we modified Ames to check for clog on each %boon if the peer is offline, so that will trigger a %clog after the first few %boon's, eliciting a subscription kill move from Gall. And then because of Arvo's depth-first move ordering, Ames will receive the kill before the rest of the %boon's.

This is less of an issue for Ames because it doesn't try to reclaim flows, but if we ever do try to get it to reclaim flows, we should keep this in mind. It's sort of like a race condition.

So I guess Eyre does need to handle this case too, for the same reason.  It should probably let Gall know the subscription is dead, though, so that Gall doesn't continuously stream %fact's to nowhere afterward.  Since %clog is per-ship for the moment, Eyre should send something else.  Maybe just a %leave?

—
~rovnys-ricfer
https://urbit.org

Fang-:
> It should probably let Gall know the subscription is dead, though, so that Gall doesn't continuously stream %fact's to nowhere afterward.  Since %clog is per-ship for the moment, Eyre should send something else.  Maybe just a %leave?

@belisarius222 this was already implemented as part of #3749. (:

belisarius222:
@Fang- Ok, good

Issue #3800, opened 2020-10-25
Cannot delete user-created groups


vcs12322:
**Describe the bug**
If I create a group in the new landscape, then when I try to delete the group, it isn't working, and the tile remains and if you try to delete it again, the settings popup doesn't work.

**To Reproduce**
1. Click the Home tile and click the dropdown to create a Group
2. Go to that group's settings and click the red "Delete this group"
3. The group tile remains and when you try to go back to the group settings, there's an error saying:
“e.group is undefined”
Stacktrace:
jv@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:461:251976
Zo@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:57929
gs@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:104168
cl@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:96718
sl@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:96641
Qs@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:93671
Wi/<@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45313
t.unstable_runWithPriority@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:322:3844
Hi@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45022
Wi@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45260
Ui@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:45193
A@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:114342
Kt@http://localhost:8080/~landscape/js/bundle/index.24abd83f9c469aff2e02.js:314:22724

**Expected behavior**
I expected the group to be deleted.

**Desktop (please complete the following information):**
Pop OS 20.04
Firefox
~[0vj.qold4.9lvpo.19vs5.ighng.hgov7.1bbq9.87ks3.eeo4m.28moq.akkc6]

matildepark:
[...]

Issue #2992, opened 2020-06-08
metadata-store: retains group metadata after deletion


galenwp:
**Describe the bug**
When I go to filter on the 'Tlon' group, I see two results. Since we migrated the group at one point, I think this is the old one. But it no longer exists in my 'all groups' view.

**To Reproduce**
This seems specific to the Tlon group, and not to any deleted group

1. Click on the 'all groups' dropdown
2. Filter on 'Tlon'
3. See 'Tlon' twice
4. Click on the 'wrong' one and have the filter just make all groups disappear

**Expected behavior**
There should only be one Tlon group, since I'm only in one.

**Screenshots**
<img width="209" alt="Screen Shot 2020-06-08 at 6 04 54 AM" src="https://user-images.githubusercontent.com/1467211/84033966-81e47e00-a94e-11ea-9765-ce6c70562a6c.png">
<img width="323" alt="Screen Shot 2020-06-08 at 6 01 21 AM" src="https://user-images.githubusercontent.com/1467211/84033970-827d1480-a94e-11ea-8b49-cdf6374ee4c1.png">

While we're at it, it'd be nice if we didn't spellcheck the group name in the filter box.

matildepark:
We're clearly retaining the `metadata-store` entry for the group (which is what we search through in that component). @loganallenc comments on appropriate behaviour?

matildepark:
Amending title because this is a metadata-store issue and the group filter is no longer in Landscape.

matildepark:
Bumping this a bit because it's more pressing in our current groups layout, where links to the group persist but lead nowhere. @tacryt-socryp let's chat about how some of the metadata work you're thinking about might incorporate this?

matildepark:
Implicit part of the modernisation work in urbit/landscape#72.

tacryt-socryp:
Fixed in #4044.

ycl924:
I have metadata remaining from groups I left, and :metadata-store [%cleanup /~ship/groupname] isn't helping (it still shows up on the debug page). I think that's what preventing me from rejoining the groups as stated in the other issue. What can I do?

matildepark:
Given that all paths you're providing are seemingly doing *nothing* it's possible you're passing the wrong paths. That said I'm unsure a cleanup is necessary here. Metadata won't prohibit a group join.

tacryt-socryp:
@ycl924 you could try [%cleanup /ship/~insert-ship/insert-group]

[...]

PR #3810, opened 2020-10-27
launch: filter out metadata for deleted groups

matildepark:
A front-end fix for #3800 until urbit/landscape#72 lands.

The launch screen iterates through `associations.contacts` to list group tiles; by using our metadata-store instead of the group-store, it inherits the "group metadata isn't deleted" bug in #2992. To the user, this is confusing — groups they're kicked out of and groups they delete alike will still persist on the launch screen (but groups they manually leave will correctly tell metadata-store to delete the data?), so it appears as though the group itself cannot be deleted.

This PR passes our groups into that logic and filters the metadata to verify we only list groups we currently actually have.

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -76,7 +76,11 @@ export default class LaunchApp extends React.Component {
               weather={props.weather}
             />
           </Box>
-          <Groups associations={props.associations} invites={props.invites} api={props.api} />
+          <Groups
+            associations={props.associations}
+            groups={props.groups}
+            invites={props.invites}
+            api={props.api} />
         </Box>
         <Box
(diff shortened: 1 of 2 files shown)


liam-fitzgerald (approved):

tacryt-socryp:
We should be able to remove these entries using the same logic but in a hoon generator.

matildepark:
> We should be able to remove these entries using the same logic but in a hoon generator.

1. This is such a widespread issue that it should be done network-wide.
2. I consider it secondary work alongside the modernisation work for metadata-store.
3. In the meantime we need something out sooner than later to fix the UX here.

tacryt-socryp:
Yes, I think it should be done in conjunction with this JS work, not instead of it.

PR #3812, opened 2020-10-27
no more impostor comets

belisarius222:
Only treat an address as a comet if it's bigger than 64 bits.  Otherwise, a comet could impersonate a different kind of ship, which posed security problems.  This PR changes comet generation in Zuse, and comet validation in Ames.  I'm not aware of other places where this validation would be necessary, but it's possible others exist.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1046,6 +1046,9 @@
   ++  on-hear-open
     |=  [=lane =packet ok=?]
     ^+  event-core
+    ::  assert the comet can't pretend to be a moon or other address
+    ::
+    ?>  (lth 64 (met 0 sndr.packet))
     ::  if we already know .sndr, ignore duplicate attestation
     ::
     =/  ship-state  (~(get by peers.ames-state) sndr.packet)
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1048,7 +1048,7 @@
     ^+  event-core
     ::  assert the comet can't pretend to be a moon or other address
     ::
-    ?>  (lth 64 (met 0 sndr.packet))
+    ?>  ?=(%pawn (clan:title sndr.packet))
     ::  if we already know .sndr, ignore duplicate attestation
     ::
     =/  ship-state  (~(get by peers.ames-state) sndr.packet)
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -9176,7 +9176,7 @@
     =/  who=ship  `@`fig:ex:cub
     ::  disallow 64-bit or smaller addresses
     ::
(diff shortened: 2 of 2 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1056,6 +1056,9 @@
     ?>  ?=(@ content.packet)
     =+  ;;  [signature=@ signed=@]  (cue content.packet)
     =+  ;;  =open-packet            (cue signed)
+    ::  assert the comet can't pretend to be a moon or other address
+    ::
+    ?>  (gth 64 (met 0 sndr.packet))
```
ohAitch, line 1061: Should this also just use `clan:title`? We probably don't want 129-bit "comets" either…

pkg/arvo/sys/zuse.hoon
```diff
@@ -2173,6 +2173,74 @@
     ++  oath  @                                         ::  signature
     --  ::  pki
   --  ::  jael
+::                                                      ::::
+::::                      ++lenz                          ::  (1l) scry network
+  ::                                                    ::::
```
ohAitch, line 2178: This core at minimum seems unrelated to the commit message

belisarius222: oops, cross-talk from another line of work -- will remove, apologies

pkg/arvo/sys/vane/ames.hoon
belisarius222: unnecessary -- `sndr.packet` is read from a 128-bit slot in the packet, not from a jammed noun

belisarius222:
ok, cleaned up the git snafu

joemfb:
FWIW, i'd be really tempted to just write these as `?>  ?=(%pawn (clan:title some-ship))`.

ohAitch:
Have the "is a bitwidth check" in the comments and "is a ship class check" in the code instead of the other way around :P

belisarius222:
Ah, yeah I could add this check to `+clan:title` if it isn't implicitly there.  That would be cleaner.

belisarius222:
redid it using `+clan:title`

joemfb:
@pilfer-pandex, the same sanity check likely needs to be added to comet "mining" in king-haskell.

ohAitch:
I mean, you're not going to mine 64 trailing zero bits on /accident/, this
seems mostly important to validate on the receiving end.

(Really if mining specific addresses were remotely feasible comets wouldn't
make sense as a concept, the worst you could hope for is hitting a random
moon which may or may not be spawned - but with a mature network this could
become, at all a viable grieving route via birthday attacks)

belisarius222:
@ohAitch Mining a moon-length comet is extremely unlikely, but still invalid, so it's worth checking when trying to create a comet address -- otherwise it's wrong, and this is Kelvin-versioned kernel code that defines the spec.

@philipcmonk I've addressed review comments. PTAL

philipcmonk (approved):
LGTM

PR #3814, opened 2020-10-27
weather: change API

### LINKED ISSUE #2634: weather: dark sky API shuts down end-2021
Our weather app uses the Dark Sky API. Now that Apple has purchased it, they're [retiring the API](https://blog.darksky.net/dark-sky-has-a-new-home/). We will want to move off the API sometime in the next 18 months.

tylershuster:
Fixes #2634

### DIFF
--- a/pkg/interface/src/views/apps/launch/components/tiles/weather.js
+++ b/pkg/interface/src/views/apps/launch/components/tiles/weather.js
@@ -1,14 +1,43 @@
 import React from 'react';
 import moment from 'moment';
 import { Box, Icon, Text, BaseAnchor, BaseInput } from '@tlon/indigo-react';
+import ErrorBoundary from '~/views/components/ErrorBoundary';
 
 import Tile from './tile';
+
+export const weatherStyleMap = {
+  Sunny: 'rgba(67, 169, 255, 0.4)',
+  PartlyCloudy: 'rgba(178, 211, 255, 0.33)',
+  Cloudy: 'rgba(136, 153, 176, 0.43)',
+  VeryCloudy: 'rgba(78, 90, 106, 0.43)',
+  Fog: 'rgba(100, 119, 128, 0.12)',
+  LightShowers: 'rgba(121, 148, 185, 0.33)',
+  LightSleetShowers: 'rgba(114, 130, 153, 0.33)',
+  LightSleet: 'rgba(155, 164, 177, 0.33)',
+  ThunderyShowers: 'rgba(53, 77, 103, 0.33)',
+  LightSnow: 'rgba(179, 182, 200, 0.33)',
+  HeavySnow: 'rgba(179, 182, 200, 0.33)',
+  LightRain: 'rgba(58, 79, 107, 0.33)',
+  HeavyShowers: 'rgba(36, 54, 77, 0.33)',
+  HeavyRain: 'rgba(5, 9, 13, 0.39)',
+  LightSnowShowers: 'rgba(174, 184, 198, 0.33)',
+  HeavySnowShowers: 'rgba(55, 74, 107, 0.33)',
+  ThunderyHeavyRain: 'rgba(45, 56, 66, 0.61)',
+  ThunderySnowShowers: 'rgba(40, 54, 79, 0.46)',
+  default: 'transparent'
+};
+
+const imperialCountries = [
+  'United States of America',
+  'Myanmar',
+  'Liberia',
+];
 
 export default class WeatherTile extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      latlong: '',
+      location: '',
       manualEntry: false,
       error: false
     };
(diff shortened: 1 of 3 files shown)


tylershuster:
@urcades we now have the following codes for weather:

https://github.com/chubin/wttr.in/blob/master/lib/constants.py#L4-L51

Compare with:

https://github.com/urbit/urbit/blob/master/pkg/interface/src/views/apps/launch/components/tiles/weather.js#L56-L87

Would appreciate insight on what do do here — note most significantly that there aren't variants for night. We could decide on some semi-transparent colors and change the background based on light/dark mode.

pkg/interface/src/views/apps/launch/app.js
```diff
@@ -1,100 +0,0 @@
-import React from 'react';
```
matildepark, line 1: Why are we deleting this?

tylershuster: what

tylershuster: I don't know where that came rom

tylershuster:
To be more clear — I think we might try an approach where we just have a map like:
```
{
  Sunny: 'rgba(12,34,56,.6)'
  Cloudy: 'rgba(22,22,22, .6)'
  ...rest
}
```
And then have the background be `white` and the text be `black` and let the indigo theme switcher take care of it.

For reference, the unique weather conditions available via the new API are:

 - [ ] Sunny
 - [ ] PartlyCloudy
 - [ ] Cloudy
 - [ ] VeryCloudy
 - [ ] Fog
 - [ ] LightShowers
 - [ ] LightSleetShowers
 - [ ] LightSleet
 - [ ] ThunderyShowers
 - [ ] LightSnow
 - [ ] HeavySnow
 - [ ] LightRain
 - [ ] HeavyShowers
 - [ ] HeavyRain
 - [ ] LightSnowShowers
 - [ ] HeavySnowShowers
 - [ ] ThunderyHeavyRain
 - [ ] ThunderySnowShowers

urcades (approved):

matildepark:
@urcades are you able to provide a palette for the list provided?

urcades:
Ha, when I had last updated this tile, the only place I recorded the values
was in the tile itself, located here:

https://github.com/urbit/urbit/blob/b7e41522d4a50461d99da4e01bec6f6779307a66/pkg/interface/src/views/apps/launch/components/tiles/weather.js#L55-L88

I still believe these to be fine colors, but I can provide a plainer list
if one is needed.

—
~fabled-faster
https://urbit.org/

tylershuster:
I wil adapt and extrapolate those to fit the new weather descriptions and generate a grid of squares in both light and dark variants for your perusal.

urcades:
Yeah, the notion of light/dark variants is very interesting — I'd be very
happy to see how you interpret their differentiation!

I admittedly came up with these only in the context of light mode, and I'd
be happy to give reason for why they are the way they are, but frankly I
was just drawing from how I felt at the time, what the weather segments
reminded me of, etc. Feel free to be similarly subjective in the approach
you take.

—
~fabled-faster
https://urbit.org/

tylershuster:
This isn't particularly inventive, but it gives you something to go on. 

Screen Shot 2020-10-28 at 4 01 37 PM
Screen Shot 2020-10-28 at 4 01 46 PM

```
export const weatherStyleMap = {
  Sunny: 'rgba(67, 169, 255, 0.4)',
  PartlyCloudy: 'rgba(178, 211, 255, 0.33)',
  Cloudy: 'rgba(136, 153, 176, 0.43)',
  VeryCloudy: 'rgba(78, 90, 106, 0.43)',
  Fog: 'rgba(100, 119, 128, 0.12)',
  LightShowers: 'rgba(121, 148, 185, 0.33)',
  LightSleetShowers: 'rgba(114, 130, 153, 0.33)',
  LightSleet: 'rgba(155, 164, 177, 0.33)',
  ThunderyShowers: 'rgba(53, 77, 103, 0.33)',
  LightSnow: 'rgba(179, 182, 200, 0.33)',
  HeavySnow: 'rgba(179, 182, 200, 0.33)',
  LightRain: 'rgba(58, 79, 107, 0.33)',
  HeavyShowers: 'rgba(36, 54, 77, 0.33)',
  HeavyRain: 'rgba(5, 9, 13, 0.39)',
  LightSnowShowers: 'rgba(174, 184, 198, 0.33)',
  HeavySnowShowers: 'rgba(55, 74, 107, 0.33)',
  ThunderyHeavyRain: 'rgba(45, 56, 66, 0.61)',
  ThunderySnowShowers: 'rgba(40, 54, 79, 0.46)',
  default: 'transparent'
};
```

urcades:
This is rather pleasant, I like the shift across cool/warm flavors of gray
. . Consider this a stamp!

—
~fabled-faster
https://urbit.org/

pkg/interface/src/views/apps/launch/components/tiles/weather.js
```diff
@@ -248,12 +230,11 @@ export default class WeatherTile extends React.Component {
           display="flex"
           flexDirection="column"
         >
-          <Text color={weatherStyle.text}>{c.summary}</Text>
-          <Text color={weatherStyle.text}>{Math.round(c.temperature)}°</Text>
-          <Text color={weatherStyle.text}>{nextSolarEvent}</Text>
+          <Text>{c.weatherDesc[0].value.replace(/([a-z])([A-Z])/g, '$1 $2')}</Text>
+          <Text>{Math.round(c.temp_F)}℉</Text>
```
matildepark, line 234: We were doing `?=auto` to get Celsius in ~most of the world~ countries that use it. Any way to maintain that here?

tylershuster: Well both units are in there. We could just check if the user ~~is free~~ location is in the United States and switch on that

matildepark: There's a few others but seems like an ok heuristic for now.

tylershuster:
Ok, I think this does it. I didn't wait 3 hours for it to pull again

tylershuster:
Ok, when it does finally get around to pulling a new payload, it renders correctly. Until then it uses centigrade. I wonder...can we force it to pull a new one when the OTA comes?

tacryt-socryp:
As a part of this... is there any way to use zip code instead of lat/long?

tylershuster:
@tacryt-socryp According to `curl wttr.in/:help`:

```
    /paris                  # city name
    /~Eiffel+tower          # any location (+ for spaces)
    /Москва                 # Unicode name of any location in any language
    /muc                    # airport code (3 letters)
    /@stackoverflow.com     # domain name
    /94107                  # area codes
    /-78.46,106.79          # GPS coordinates
```

my zip code gets put in Russia, but my city name works.

As long as the back-end just stores a string, we could just allow any inputs with some suggestions like the above

matildepark:
I believe only FE validates the latlong.

matildepark:
Yeah, it looks like if you poke Weather with JSON it'll just store whatever string it's given.

tylershuster:
Do we want to do a sanity check to ensure that the string you've entered matches where you think it is? We could do that on FE with `fetch`

matildepark:
There's definitely some sort of sanity issue here because of how Weather currently works; if we have a location string stored but no weather data we say "Try again later" with no way to set the location. The easiest way I got? Return the weather location from wttr in the tile and offer the arrow to change the location if we're in that error state.

@urcades

matildepark:
I also have a question — if we're a user who does this migration, will it *immediately* get the new weather data or will we, for 3 hours, serve the old data to the FE? Because that could brick FE potentially.

tylershuster:
The second, hence my question about forcing an update. See also "`import ErrorBoundary from '~/views/components/ErrorBoundary';`"

tylershuster:
Can I have this one too

tylershuster:
Would this be a good one to target for the breach?

tylershuster:
And I forgot what we decided in the meeting. Are we just allowing any string and then if it doesn't show what you expect, you enter a new string?

tylershuster:
It now works as described. In the case that the user has entered a location, it now shows on the "set location" screen as a sanity check:

image

matildepark:
I'll try to break this sometime today

matildepark (commented):
Booted fakezod; set weather on next-userspace, applied these changes and restarted dev environment. Doesn't crash interface, just has an error. Presumably it will do this for 3 hours.

image

If I wipe location from react state I can set it successfully.

What I suggest:

- Fix the abnormally high amount of padding on this error tile state
- Allow users to set weather data from the error tile so they can just force a refresh

matildepark:
Though honestly you could just look for `weather.currently` or whatever matches a part of the old data structure, and if it matches, send another locationSubmit automatically; that would improve UX a lot.

tylershuster:
This is a good idea. FE versioned state wen. 

I'll implement this soon, maybe tonight.

tylershuster:
Ok, concerns addressed.

matildepark (approved):
Just works.

PR #3815, opened 2020-10-28
clock: add background to moon

tylershuster:
@urcades

### DIFF
--- a/pkg/interface/src/views/apps/launch/components/tiles/clock.js
+++ b/pkg/interface/src/views/apps/launch/components/tiles/clock.js
@@ -72,10 +72,19 @@ const Moon = ({ angle, ...props }) => {
         <rect x="-50" y="-50" height="200" width="200" fill="black" />
         <path d={dFromPhase(phase)} fill="white"/>
       </mask>
+      <circle
+        id="moonbg"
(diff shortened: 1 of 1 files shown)


tylershuster:
Can we get this out before the next OTA?

matildepark:
If @urcades @g-a-v-i-n are around ...

urcades:
Checking this out now

urcades:
Including the image of the interface for gav's sake, or anyone who doesn't want to pull the branch:

Screen Shot 2020-11-05 at 7 02 45 PM

Comparison .gif:
Kapture 2020-11-05 at 19 06 07

also the current phase of the moon, for anyone interested:

Screen Shot 2020-11-05 at 7 03 38 PM

I think this treatment is a fine update to the current state. It certainly keeps the shape intact, compared to the last version.

For future reference, for nearly entirely visual PRs like this one, it would speed up the review process massively if images like the ones above were provided, that way an explicit checkout process is necessitated. If there's worry that there may be edge cases, I think it's fine to ask that we check out a branch, but otherwise visuals are always welcome in a PR like this!

urcades (approved):

PR #3816, opened 2020-10-28
chat: fixes relative datestamps

tylershuster:
fixes https://github.com/urbit/landscape/issues/125


matildepark (approved):
Confirmed operational

PR #3823, opened 2020-10-28
king: cancel http servers immediately on shutdown.

### LINKED ISSUE #3822: king-haskell: shutting down doesn't free http port
If you shut down king haskell normally, and then start it up immediately after, the http port changes because the old port wasn't freed up correctly. It looks like we aren't properly shutting down the http server.

My current hypothesis about what's going on: In Eyre.hs, we do call a kill STM action which is supposed to signal to kill the server...but we don't actually wait for the result of that shutdown! If you look at the implementation of that STM action, it notifies a separate async which is just sitting around to cancel another thread instead.

eglaysher:
This should fix the problem where shutting down urbit-king, and then immediately starting it up again changes the http port. I hit this all the time and it is super annoying.

Fixes #3822


pilfer-pandex (approved):
awesome

PR #3828, opened 2020-10-29
groups-js: add ourselves to admin in group-update reducer

liam-fitzgerald:
Brings behaviour in line with group-store agent.

### DIFF
--- a/pkg/interface/src/logic/reducers/group-update.ts
+++ b/pkg/interface/src/logic/reducers/group-update.ts
@@ -103,7 +103,7 @@ export default class GroupReducer<S extends GroupState> {
       const resourcePath = resourceAsPath(resource);
       state.groups[resourcePath] = {
         members: new Set(),
-        tags: { role: {} },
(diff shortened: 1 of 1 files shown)


pkg/interface/src/logic/reducers/group-update.ts
```diff
@@ -103,7 +103,7 @@ export default class GroupReducer<S extends GroupState> {
       const resourcePath = resourceAsPath(resource);
       state.groups[resourcePath] = {
         members: new Set(),
-        tags: { role: {} },
+        tags: { role: { admin: new Set([window.ship]) } },
```
matildepark, line 106: My question is "does `addGroup` only apply to a group we create? What if we accept an invite to a group?"

liam-fitzgerald: That comes in as an initialGroup update

matildepark (approved):

PR #3829, opened 2020-10-29
Profile: use ColorInput for background

liam-fitzgerald:
Uses ColorInput instead of a regular text Input to bring background color selection inline with the rest of landscape


matildepark (approved):
We get a colour picker for free!

PR #3832, opened 2020-10-29
observe-hook: spin up a thread upon receiving a %fact from a subscription

tacryt-socryp:
This is a (properly compiling)  sketch of a hook that allows one to create side effects upon subscription updates from any `%gall` application within Urbit simply by setting up an observer that runs some `thread` any time it receives a `%fact`. This is necessary for us to make invite acceptance work alongside the `push/pull-hook`s, but I think it's generally useful. Posting for commentary.

Threads can already watch subscriptions themselves, but I think this utility makes the management of those threads far easier.


matildepark:
What's left on this @tacryt-socryp?

tacryt-socryp:
Just got this working less than 5 minutes ago actually. We can now successfully accept invites to unmanaged publish and links graphs. Now I just need to finish the edge cases (error handling).

pkg/arvo/app/thread-watch-hook.hoon
```diff
@@ -0,0 +1,199 @@
...
+    ?~  p.sign  [~ this]
+    ?+    wire  (on-agent:def wire sign)
+        [%thread-result @ ~]  [~ this]
+        [%observer @ ~]
+      =/  =time  (slav %da i.t.wire)
+      [~ this(observers (~(del by observers) time))]
+    ==
+  ::
+      %fact
+    ?+    wire  (on-agent:def wire sign)
```
liam-fitzgerald, line 156: I think this would be cleaner to route on wire before sign

pkg/arvo/app/thread-watch-hook.hoon
```diff
@@ -0,0 +1,199 @@
...
+      =/  =observer:sur  (~(got by observers) time)
+      =/  tid  (scot %uv (sham eny.bowl))
+      :_  this
+      :~  :*  %pass
+              [%thread-start i.t.wire tid ~]
+              %agent
+              [our.bowl %spider]
+              %poke
+              %spider-start
+              !>([~ `tid thread.observer (slop q.cage.sign !>(~))])
```
liam-fitzgerald, line 206: Maybe it's the right time to change this behaviour

tacryt-socryp:
Can confirm that I've tested positive and negative poke-acks, positive and negative watch-acks, %thread-fail, %thread-done, and all of the poke actions and they all both work properly and clean up state appropriately. This is ready to go in.

liam-fitzgerald (commented):
Some nitpicks, overall structure and layout looks good

philipcmonk (commented):
Looks pretty good, but I don't understand what `time` is used for.  Is it just a UUID?  If so, `eny` would be a better choice to give sub-event precision.

tacryt-socryp:
@philipcmonk @liam-fitzgerald Addressed review comments and tested it again. This is ready to go.

pkg/arvo/app/thread-watch-hook.hoon
tacryt-socryp: Done.

pkg/arvo/app/thread-watch-hook.hoon
tacryt-socryp: Not yet :)

tacryt-socryp:
@philipcmonk I moved away from time identifiers to `serial` identifiers for sub-event precision.

liam-fitzgerald (approved):

PR #3833, opened 2020-10-29
Revert "eyre: ignore facts directly after clog"

Fang-:
Reverts urbit/urbit#3799. It was accidentally merged into master, but should've targetted next-sys instead. Reverting to regain parity between master and livenet.


matildepark:
@Fang- Wait, does this mean @philipcmonk always runs master and caught this early? Or are you pushing this out?

Fang-:
This means there was code on master that had not been deployed. @philipcmonk was running what had been deployed (just like the rest of us), which did not include this, but will probably fix the issue he reported once it goes out.

PR #3834, opened 2020-10-29
nix: clarify lfs error message

ohAitch:


joemfb:
As of #3806, the relevant nix derivations no longer assume that they're running on top of a git working with lfs properly configured -- instead, they explicitly resolve the lfs pointer as needed.

ohAitch:
Nice

PR #3843, opened 2020-10-30
Antiautodefenestration

tylershuster:
don't do it @galenwp 

(two commits preventing accidental deletion of data — feel free to suggest more)

### DIFF
--- a/pkg/interface/src/logic/lib/util.js
+++ b/pkg/interface/src/logic/lib/util.js
@@ -327,11 +327,18 @@ export function numToUd(num) {
   )(num.toString())
 }
 
-export function usePreventWindowUnload(shouldPreventDefault) {
+export function usePreventWindowUnload(shouldPreventDefault, message = "You have unsaved changes. Are you sure you want to exit?") {
   React.useEffect(() => {
     if (!shouldPreventDefault) return;
-    const handleBeforeUnload = event => event.preventDefault();
+    const handleBeforeUnload = event => {
+      event.preventDefault();
+      return message;
+    }
     window.addEventListener("beforeunload", handleBeforeUnload);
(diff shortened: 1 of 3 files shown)


galenwp:
life saving

pkg/interface/src/views/apps/publish/components/MarkdownEditor.tsx
```diff
@@ -22,6 +24,17 @@ interface MarkdownEditorProps {
   onBlur?: (e: any) => void;
 }
 
+const PromptIfDirty = () => {
+  const formik = useFormikContext();
+  usePreventWindowUnload(formik.dirty);
+  return (
+    <Prompt
+      when={formik.dirty}
+      message="Are you sure you want to leave? You have with unsaved changes."
```
lycurgus, line 33: Tiny nit: "have unsaved" ?

pkg/interface/src/logic/lib/util.js
```diff
@@ -325,3 +326,12 @@ export function numToUd(num) {
     f.join('.')
   )(num.toString())
 }
+
+export function usePreventWindowUnload(shouldPreventDefault) {
+  React.useEffect(() => {
+    if (!shouldPreventDefault) return;
+    const handleBeforeUnload = event => event.preventDefault();
```
matildepark, line 333: This doesn't work on Chrome for whatever reason, you can still close the tab.

tylershuster: Looks like Chrome uses a slightly different approach. Give this a shot.

matildepark (approved):

matildepark:
lol @tylershuster this is a little too aggressive — fires a prompt if you submit a post normally

tylershuster:
Oh ha. hold plz

PR #3848, opened 2020-10-30
landscape: 10/30 QA fixes from design feedback

matildepark:
- Removes big sigil tile from Launch screen. (RIP)
- Chat messages are lined up with chat input.
- Nicknames use font-weight 500.
- Message containers use `pt='3'` not `pt2` to level the padding out between messages (and hopefully also level the unread marker).
- Code blocks are pleasantly rounded and have even padding.
- Tile borders use `scales.black20` (cc @g-a-v-i-n is this intentional for dark mode's being almost invisible then?)
- Remove reverse alphabetical sidebar sort.
- Default sorting is last updated.
- Invite icon uses 'Users' not 'createGroup'
- Mobile devices couldn't navigate back from 'new channel,' so this adds a back button.
- Only admins can now see the invite icon (#3803)


liam-fitzgerald (approved):
LGTM!

PR #3850, opened 2020-10-30
king: increase lmdb max mmap size for parity with vere

pilfer-pandex:
Previously, the mapsize was 10 GiB. This change increases it to 1 TiB, which should be the same that the c king uses.

When I tried to switch ~def over to kh for testing, I found that the ship would promptly crash because of this discrepancy.

### DIFF
--- a/pkg/hs/urbit-eventlog-lmdb/lib/Urbit/EventLog/LMDB.hs
+++ b/pkg/hs/urbit-eventlog-lmdb/lib/Urbit/EventLog/LMDB.hs
@@ -100,7 +100,7 @@ rawOpen :: MonadIO m => FilePath -> m Env
 rawOpen dir = io $ do
     env <- mdb_env_create
     mdb_env_set_maxdbs env 3
-    mdb_env_set_mapsize env (100 * 1024 * 1024 * 1024)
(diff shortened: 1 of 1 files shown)


joemfb (approved):
LGTM

botter-nidnul:
Ah, this explains why urbit-king suddenly stopped working on aarch64.

I can change it to 60 GiB, but I'm afraid I don't know enough about Haskell to do the equivalent of 
```
#if defined(U3_CPU_aarch64) && defined(U3_OS_linux)
        const size_t siz_i = 64424509440;
```

which I did in #3201

brendanhay:
We can do the same #if/def in the Haskell code either directly where `mdb_env_set_mapsize` is called or add a flag to the build/generated cabal macros, I'll take a look at it tomorrow.

botter-nidnul:
Did some experimenting to find the limit for the platform, since 60 GiB was chosen as the aarch64 mapsize just because it was the default for all linux platforms prior to the 1 TiB change.

I can't get `urbit` to run with the mapsize set any higher than 500 GiB (536870912000 bytes) and `urbit-king` maxes out at 253 GiB.

I don't really have any problem with keeping it at 60 GiB, since hopefully no one is running a galaxy on a Pi. But there's something interesting going on with the difference between the two.

PR #3854, opened 2020-11-02
launch: fixes tile padding

tylershuster:
fixes https://github.com/urbit/landscape/issues/129

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import Helmet from 'react-helmet';
+import styled from 'styled-components';
 
 import { Box, Row, Icon, Text, Center } from '@tlon/indigo-react';
 import { uxToHex, adjustHex } from '~/logic/lib/util';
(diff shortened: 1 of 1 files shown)


matildepark (changes requested):
image

Padding isn't helping here I think.

image

tylershuster:
feckin scrollbar

tylershuster:
Ok, this is a bit bold but I think it achieves what we want without sacrificing usability. The scrollbar will no longer be visible on the home screen but it will still be scrollable. This fixes the padding issues

pkg/interface/src/views/apps/launch/app.js
```diff
@@ -32,10 +41,10 @@ export default class LaunchApp extends React.Component {
         <Helmet>
           <title>OS1 - Home</title>
         </Helmet>
-        <Box height='100%' overflowY='scroll'>
+        <ScrollbarLessBox height='100%' overflowY='scroll'>
```
matildepark, line 44: TBH we're going to want this same thing on the sidebar probably. Separate PR.

matildepark (approved):

PR #3860, opened 2020-11-02
publish: fixes #3859, fixes #3844

### LINKED ISSUE #3844: publish: Editing a post causes endless spinner, but edit shows on refresh
Via @vvisigoth.

### LINKED ISSUE #3859: publish: endless spinner on submitting post update
**Describe the bug**
Updating an existing post doesn't terminate. But edit is propagated.

**To Reproduce**
1. Make a post
2. Click on "Update"
3. Update the post
4. Click save

**Expected behavior**
Edit is saved and returned to post screen

tylershuster:
This is the same fix as https://github.com/urbit/urbit/pull/3856

@liam-fitzgerald does this have something to do with the work you mentioned on bigints earlier? `===` checks are failing because of a JS variable type and I haven't looked deeply into it but my intuition tells me a) that's the issue and b) you might know something about a holistic pass to remedy the issue rather than adding `valueOf()` piecemeal

### DIFF
--- a/pkg/interface/src/logic/lib/publish.ts
+++ b/pkg/interface/src/logic/lib/publish.ts
@@ -87,7 +87,7 @@ export function getLatestRevision(node: GraphNode): [number, string, string, Pos
     return empty
   }
   const [title, body] = rev.post.contents as TextContent[];
-  return [revNum, title.text, body.text, rev.post];
+  return [revNum.toJSNumber(), title.text, body.text, rev.post];
 }
 
 export function getComments(node: GraphNode): GraphNode {


matildepark:
Tagging @liam-fitzgerald for possible holistic solution; if this thread is left pending around the next OTA I will merge this as is

pkg/interface/src/views/apps/publish/components/EditPost.tsx
```diff
@@ -37,7 +37,7 @@ export function EditPost(props: EditPostProps & RouteComponentProps) {
       await api.graph.addNodes(ship, book, nodes);
       await waiter(p => {
         const [rev] = getLatestRevision(p.note);
-        return rev === newRev;
+        return rev.valueOf() === newRev;
```
liam-fitzgerald, line 40: `getLatestRevision` in `lib/publish.ts` should be changed such that `return [revNum.toJSNumber(), title.text, body.text, rev.post]`; is the return value.

liam-fitzgerald (changes requested):
should change this so that the revision number is a JS number.

tylershuster:
Cheers @liam-fitzgerald — now we don't need https://github.com/urbit/urbit/pull/3856 but it won't hurt

liam-fitzgerald (approved):
LGTM!

PR #3861, opened 2020-11-02
Use new-style gate argument syntax in /mar

ronreg-ribdev:
Switch a bunch of .hoon files in /mar from using
the old `{}` syntax for specifying argument types, to the
newer `[argname=argtype]` syntax.

### DIFF
--- a/pkg/arvo/mar/css.hoon
+++ b/pkg/arvo/mar/css.hoon
@@ -14,7 +14,7 @@
   --
 ++  grab
   |%                                                    ::  convert from
-  ++  mime  |=({p/mite q/octs} (@t q.q))
+  ++  mime  |=([p=mite q=octs] (@t q.q))
   ++  noun  @t                                         ::  clam from %noun
   --
 ++  grad  %mime
--- a/pkg/arvo/mar/hoon.hoon
+++ b/pkg/arvo/mar/hoon.hoon
@@ -41,7 +41,7 @@
   --
 ++  grab
   |%                                            ::  convert from
-  ++  mime  |=({p/mite q/octs} q.q)
+  ++  mime  |=([p=mite q=octs] q.q)
   ++  noun  @t                                  ::  clam from %noun
   ++  txt   of-wain:format
   --
--- a/pkg/arvo/mar/html.hoon
+++ b/pkg/arvo/mar/html.hoon
@@ -17,7 +17,7 @@
 ++  grab  ^?
           |%                                            ::  convert from
           ++  noun  @t                                  ::  clam from %noun
-          ++  mime  |=({p/mite q/octs} q.q)             ::  retrieve form $mime
+          ++  mime  |=([p=mite q=octs] q.q)             ::  retrieve form $mime
           --
 ++  grad  %mime
 --
--- a/pkg/arvo/mar/js.hoon
+++ b/pkg/arvo/mar/js.hoon
@@ -15,7 +15,7 @@
   --
 ++  grab
   |%                                                    ::  convert from
-  ++  mime  |=({p/mite q/octs} (@t q.q))
+  ++  mime  |=([p=mite q=octs] (@t q.q))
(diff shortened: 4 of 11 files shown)


Fang- (approved):
Unless there is some arcane reason why `a/_""` should remain as `a=_""` instead of `a=tape`, this looks good to me. Thank you!

ronreg-ribdev:
@Fang- looks like CI failed? I can't tell from the travis error message whether this PR legitimately broke something or if it's a CI bug

joemfb:
It's a legit boot failure (which unfortunately does not currently include stack-trace). There might be a syntax error in one of these changes, or a problem with the pill that's checked.

To reproduce, try booting a fake-zod from this branch with `-A`. To track down the problem, boot a fake-zod without `-A` and use the `:test` app to build all the marks (`|start %test` and then `:test %marks`).

ronreg-ribdev:
> It's a legit boot failure (which unfortunately does not currently include stack-trace). There might be a syntax error in one of these changes, or a problem with the pill that's checked.
> 
> To reproduce, try booting a fake-zod from this branch with `-A`. To track down the problem, boot a fake-zod without `-A` and use the `:test` app to build all the marks (`|start %test` and then `:test %marks`).

I can run `./urbit-v0.10.8-linux64/urbit -A <my-local-urbit-checkout> -F nec` and reproduce a failure locally:
```play: bail: %fail
serf: fail: end of file
fish: “./urbit-v0.10.8-linux64/urbit -…” terminated by signal SIGABRT (Abort)
```

I'm not sure how to use the `:test` app in a fakezod to build the marks in my local checkout from source. Do I need to use the urbit binary built by `make build`? (and where in the local filesystem does that binary get put anyway?)

joemfb:
@ronreg-ribdev, you can use nix to install a locally-built urbit, run it directly out of the nix-store, or use `nix-shell` in `pkg/urbit` (and  `./configure`, `make -j 8`, run out of `./build`).

But you shouldn't need a new binary, nothing here should be affected one way or the other.

This will establish a baseline, ignoring changes in this PR (useful to rule out pre-existing failures):

```
path/to/urbit -F ~zod -B bin/solid.pill -c my-fake-zod
...boots...
> |start %test
> :test %marks
```

Then, you can `|mount %`, copy in your changed marks, `|commit %home`, and run `:test %marks` again.

ronreg-ribdev:
> @ronreg-ribdev, you can use nix to install a locally-built urbit, run it directly out of the nix-store, or use `nix-shell` in `pkg/urbit` (and `./configure`, `make -j 8`, run out of `./build`).
> 
> But you shouldn't need a new binary, nothing here should be affected one way or the other.
> 
> This will establish a baseline, ignoring changes in this PR (useful to rule out pre-existing failures):
> 
> ```
> path/to/urbit -F ~zod -B bin/solid.pill -c my-fake-zod
> ...boots...
> > |start %test
> > :test %marks
> ```
> 
> Then, you can `|mount %`, copy in your changed marks, `|commit %home`, and run `:test %marks` again.

When I try this on a fresh fakezod, I get the following failures:
```[%no-cast-from %publish-info %mime]   
[%error-building-cast %publish-info %mime]  
[%error-building-mark %publish-info]        
[%mark-failed i=%~.mar t=/publish-info]      ```

with a `%some-marks-failed` error message at the end of it.

Running the test with my changes results in a bunch more failures. This seems to be fixed by restoring the line in `mime.hoon` to `|=(a=_"" [/application/x-urb-unknown (as-octt:mimes:html a)])`. So I guess that *is* different from `a=tape` in some meaningful way, although I am not sure how. I just re-pushed with that change and hopefully it will build successfully now

Fang-:
Oh, uh, yeah I guess that'd fail. This is one of those "painfully obvious in hindsight" cases:

Using `tape` there refers to the name of the arm you're in, not the `tape` you expect from the standard library. Using `^tape` instead should also do the trick.

PR #3862, opened 2020-11-02
landscape: hide new channel container, daybreak in unread marker

matildepark:
- The new channel container had some padding leftover when its child was hidden, preventing users from clicking bottom channels; we hide the container itself now
- I was going to add `flex-shrink` to the daystamp in the unread marker, but it always provided useless data for its intended use case (eg. staggering the daybreak + unread marker instead of combining them, see below) so I just remove it.

Screen Shot 2020-11-02 at 3 43 01 PM


liam-fitzgerald (approved):
LGTM!

PR #3863, opened 2020-11-02
group switcher: full row links to group; group title is flex truncated

matildepark:
- Instead of clicking text, you can click the full row to go to the group in question in recent groups
- Group title was truncated in the dropdown with a max width of `131px`. By passing `min-width: 0` all the way up we can use flexbox to set the width and truncate the title depending on viewport. Previously, titles were cut off halfway through the viewport on mobile (bigger sidebar).

cc @liam-fitzgerald Anything I need to remember with the changes to dropdown here?


liam-fitzgerald (approved):
LGTM!

PR #3864, opened 2020-11-02
clay: replace ream with nouns and !, zapcom

Fang-:
Motivation for the change is performance improvements on the un-`^~`d uses of
`+ream`. Parsing turns out to be slow, making `+ream` slow in turn. So we construct
the hoon ast manually instead.

`!,` is arguably better style than ream, since it doesn't necessitate a `^~` for static
input, and lets syntax highlighting function properly.

For the investigated case, in `+get-cast`'s +grow flow, improves performance by
over 80%.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -561,7 +561,7 @@
         ++  vale
           |=  =noun
           ^+  sam
-          (slam (slap cor ^~((ream 'noun:grab'))) !>(noun))
+          (slam (slap cor !,(*hoon noun:grab)) !>(noun))
         ++  volt
           |=  =noun
           ^+  sam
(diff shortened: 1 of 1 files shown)


ohAitch:
It's certainly clearer.

If !, is showing up in vanes we should probably add a biblical for ++hoon,
so you can just say !,(tw(i g)) - probably the OTA-able version of this is
to just modify the parser to stag it onto a one-child syntax(and an
optional wide two-child form to cover the one existing use in also clay),
though that's certainly less clean than the way !> / !; work

ohAitch:
Oh actually the ! namespace is much less crowded than I remembered, given
?!(@) ?!(^) etc are pretty nonsensical - let me amend that to propose !~
for a one-arity rune that expands to biblical !,(*hoon a), by analogy to |~
/ $- being gates which are kind of like unevaluated code ;)

Idk enough bible for a four-letter name invoking the concept of risk of
overabstraction by operating on language, something tower of babel themed
perhaps? <s> @philipcmonk help </s>

belisarius222 (approved):
LGTM

joemfb (commented):
It's a small thing, but `[%limb a]` is probably preferable to [%wing ~[a]]` throughout.

Fang-:
Good catch @joemfb, I was matching the `!,`-generated AST a bit too closely.

philipcmonk (approved):

PR #3865, opened 2020-11-03
graph-push-hook: allow for node removal by author

liam-fitzgerald:


matildepark:
This should probably include FE changes in the PR?

liam-fitzgerald:
ruh roh

matildepark:
Is that prohibitive? We just want to avoid a situation where back-end and front-end changes are deviated and it's unclear which support which.

liam-fitzgerald:
no, I just forgot accidentally

pkg/arvo/app/graph-push-hook.hoon
```diff
@@ -32,6 +32,22 @@
...
       (is-admin:grp src.bowl i.group-paths)
   ==
+::
+++  is-allowed-remove
+  |=  [=resource:res indices=(set index:store) =bowl:gall]
+  ^-  ?
+  =/  gra   ~(. graph bowl)
+  ?.  (is-allowed resource bowl %.n)
+    %.n
+  %+  levy
```
tacryt-socryp, line 42: What does this do? The method description in the standard library is... not helpful.

tacryt-socryp: Oh, okay. It just returns yes if all of the result tests are yes... essentially an ?& across a list.

pkg/arvo/lib/graph.hoon
```diff
@@ -17,6 +17,18 @@
   %+  scry-for  update:store
   /graph/(scot %p entity.res)/[name.res]
 ::
+++  get-node
```
tacryt-socryp, line 20: Technically this is "got-node", as "get-node" would return a unit.

tacryt-socryp (changes requested):
One small change (get to got), otherwise looks good.

tacryt-socryp (approved):

PR #3867, opened 2020-11-03
gall: fix crashed %fact handling not removing subscription from yoke

liam-fitzgerald:
If the handling of a %fact crashes in the gall agent, we kill the
subscription with +ap-kill-down. However, +ap-kill-down did not remove
the subscription from the yoke. This causes spurious %watch-not-unique
errors, as the bowl may contain subscriptions that have been killed by
crashed %fact handling.

Postscript: This is almost certainly the cause of publish's mysterious 'kicked for no reason' behaviour and the subsequent `%watch-not-unique` on resubscribe.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7a5ffa86349b0f418d7d5a5f69a773d29b1d48d80d657058ff5abb0572f5187f
-size 6409970
+oid sha256:9ead5c73510b0e764d0acdfd8c9dd40c599a065eaf43e8b313399ab3ba5147fe
+size 6338075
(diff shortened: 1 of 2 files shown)


belisarius222 (approved):
LGTM

Good catch.  I believe this was my mistake, so mea culpa.

liam-fitzgerald:
ping @philipcmonk I'd like to get this fixed&deployed asap, it makes fixing broken subscriptions in OTAs somewhat unreliable

philipcmonk:
Sorry I missed this. Doesn't the kick remove from outbound.watches in
ap-specific-take? Is it trying to remove the wrong wire or dock or
something?

philipcmonk:
@liam-fitzgerald I don't understand why the `%kick` doesn't do this.

liam-fitzgerald:
on a second look, i was reading this code incorrectly

PR #3869, opened 2020-11-03
hark, graph-store: mentions support.

liam-fitzgerald:
- Adds a %mention type to +content:post
- hark-graph-hook scans for %mentions of our own ship instead of doing wacky text processing
- Links, publish now tokenise mentions before sending to backend
- Mentions in comments are highlighted and replaced with the nickname, if any.
Visual demo: 
Capture d’écran 2020-11-03 à 4 19 17 PM

### DIFF
--- a/pkg/arvo/app/hark-graph-hook.hoon
+++ b/pkg/arvo/app/hark-graph-hook.hoon
@@ -211,11 +211,9 @@
       ^-  ?
       ?.  mentions  %.n
       ?~  contents  %.n
-      ?.  ?=(%text -.i.contents)
+      ?.  ?=(%mention -.i.contents)
         $(contents t.contents)
-      =/  res
-         (find (scow %p our.bowl) (trip text.i.contents))       
-      ?^  res
+      ?:  =(our.bowl ship.i.contents)
         %.y
       $(contents t.contents)
     ::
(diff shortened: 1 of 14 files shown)


urcades:
Oh yeah

tacryt-socryp (approved):
Hoon looks great.

galenwp:
exciting

liam-fitzgerald:
Note for those following at home: This will be unable to be prereleased for beta testing on livenet, due to the change in the +content:post type and the subsequent mark validation change that will occur. As a result, I've PR'd this change against the original branch, with the intention of keeping the `lf/hark-redux` branch testable on livenet so that we can cleanup any issues there. 

cc: @matildepark

matildepark:
> As a result, I've PR'd this change against the original branch, with the intention of keeping the lf/hark-redux branch testable on livenet so that we can cleanup any issues there.

Except it has a different mentions implementation, and we'll leave a great deal of functionality untested if we do that. We can have a subnet of moons with these changes instead of operating against livenet content; so I'd like to merge this in anyway. @tacryt-socryp?

tacryt-socryp:
I'd like it to be merged and tested as a single PR as well.

PR #3870, opened 2020-11-03
lib-push-hook: fix behaviour of %remove

### LINKED ISSUE #3868: Irrelevant code or bug?
The =/  paths isn't used in the +remove gate here. I'm still wrapping my head around the logic so can't say whether this =/ should be removed or if there is something wrong here. Thought I'd mention it.

https://github.com/urbit/urbit/blob/e279b080495f3af2020aab8ea8e84a36d326a9ea/pkg/arvo/lib/push-hook.hoon#L285

liam-fitzgerald:
%remove did not properly iterate over the paths and as such did not
return a complete list of paths to kick.

Fixes #3868

### DIFF
--- a/pkg/arvo/lib/push-hook.hoon
+++ b/pkg/arvo/lib/push-hook.hoon
@@ -282,14 +282,15 @@
       |=  rid=resource
       =/  pax=path
         [%resource (en-path:resource rid)]
-      =/  paths=(list path)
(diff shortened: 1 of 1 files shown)


matildepark:
ping @tacryt-socryp

tacryt-socryp (approved):

Issue #3857, opened 2020-11-02
landscape: leaving a group dumps all channels, even if unjoined, into home workspace


devinmancuso:
**Describe the bug**
Click on the _Settings_ and _Refresh_ icons produce errors and incorrect function calls when accessing channels I have never joined from groups I am no longer a member of.

**To Reproduce**
1. Join a group by clicking on a link in a chat
2. Click on each channel in the group but do **not** join any of the channels in that group
3. Go to landscape home
4. Click on the newly created group tile 
5. Leave the group via button
6. Go to landscape home
4. Click on your Home tile
5. You should see the channels from that group greyed out in your channel list
6. Click on one of these channels on the LHS menu

**Expected behavior**
These channels should not appear in the list under 'home' as I never joined them and I left the group.

**Actual behavior**
With a channel selected after steps above
1. Clicking on the 'cog settings icon' will result in an error screen "“Cannot read property 'metadata' of undefined”"
2. Clicking on the 'refresh icon' will result the in 'new DM' modal appearing

**Screenshots**
<img width="745" alt="Screen Shot 2020-11-02 at 12 28 29 AM" src="https://user-images.githubusercontent.com/1478273/97846133-6d5ca580-1ca2-11eb-93cc-c58078b32f90.png">
(clicking on the refresh icon)

**Desktop (please complete the following information):**
 - OS: MacOS v10.15.7
 - Browser Chrome

matildepark:
tagging @tacryt-socryp for potential FE work to hide metadata for channels we never had (or any association with %sane work in #3416)

tacryt-socryp:
Going to hit this now by implementing a thread in `observe-hook`

tacryt-socryp:
Fixed in #3986.

PR #3871, opened 2020-11-03
graph-pull-hook: only archive if we have graph

liam-fitzgerald:
Scries to see if we already have the graph, and only then do we archive it. 
Unrelated, but how the hell does lib/graph work here if we aren't passing it the bowl?


matildepark:
Part of a fix for #3857.

matildepark:
ping @tacryt-socryp

tacryt-socryp (approved):

PR #3872, opened 2020-11-03
test: cleans up output, adds |trim and |meld

joemfb:
with intervening |mass measurements (the before/after `~&` printfs ensure that `stderr` output is fully captured).


brendanhay (approved):
Clearly the real point of this PR was typographic subversion.

PR #3873, opened 2020-11-03
verb: render printouts lazily

Fang-:
Instead of passing in a tape to print in case we need in, pass in a trap that
can be called to generate the tape on-demand.

Eagerly rendering printouts was costing us a lot of time when we didn't actually
need them, because the pretty-printer is slow.

### DIFF
--- a/pkg/arvo/lib/verb.hoon
+++ b/pkg/arvo/lib/verb.hoon
@@ -12,26 +12,26 @@
 ::
 ++  on-init
   ^-  (quip card:agent:gall agent:gall)
-  %-  (print bowl "{<dap.bowl>}: on-init")
+  %-  (print bowl |.("{<dap.bowl>}: on-init"))
   =^  cards  agent  on-init:ag
   [[(emit-event %on-init ~) cards] this]
 ::
 ++  on-save
   ^-  vase
-  %-  (print bowl "{<dap.bowl>}: on-save")
+  %-  (print bowl |.("{<dap.bowl>}: on-save"))
   on-save:ag
 ::
 ++  on-load
   |=  old-state=vase
   ^-  (quip card:agent:gall agent:gall)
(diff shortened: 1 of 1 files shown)


pkg/arvo/lib/verb.hoon
```diff
@@ -62,39 +62,40 @@
...
   ^-  (quip card:agent:gall agent:gall)
-  %-  (print bowl "{<dap.bowl>}: on-fail with term {<term>}")
+  %-  (print bowl |.("{<dap.bowl>}: on-fail with term {<term>}"))
   =^  cards  agent  (on-fail:ag term tang)
   [[(emit-event %on-fail term) cards] this]
 --
 ::
 ++  print
-  |=  [=bowl:gall =tape]
+  |=  [=bowl:gall tape=_^?(|.(*tape))]
```
joemfb, line 92: This can just be `tape=(trap tape)`.

pkg/arvo/lib/verb.hoon
```diff
@@ -62,39 +62,40 @@
...
   ^+  same
   =?  .  bowl-print
     %-  (slog >bowl< ~)
     .
   ?.  loud  same
-  %-  (slog leaf+tape ~)
+  %-  (slog leaf+(tape) ~)
```
joemfb, line 98: While I'm nitpicking, I think `$:tape` would be idiomatic here. (Also, the face `tape` has become questionable.)

belisarius222:
Yeah, note that Ames does something very similar, taking in a trap: https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/ames.hoon#L109

philipcmonk (approved):

pkg/arvo/lib/verb.hoon
ohAitch: `=mold` delenda est? 🤔

PR #3876, opened 2020-11-03
vere: set maximum http request to 512 MiB

joemfb:
The default was 1 GiB, which seems high.

(This is not a pressing issue, but I had made a note to change it, and I want to test hercules-ci with a PR from a fork.)


joemfb:
That's what I was afraid of ...

@brendanhay, it looks like only pushes to branches of `urbit/urbit` are picked up by CI. Do you know of anything we can do about this?

Fang- (approved):
Nice when h2o does the work for you. Looks like a sane config change to me! Nobody should be uploading data this big into their urbits right now anyway.

brendanhay:
@joemfb https://github.com/hercules-ci/support/issues/10

PR #3877, opened 2020-11-03
AsyncButton: protect against double submit

liam-fitzgerald:
Disables `AsyncButton` if the form is submitting, to prevent accidental double submits.

Design: it makes the new channel screen look this (note the different color of the button):
Capture d’écran 2020-11-04 à 5 19 35 AM

cc: @urcades

### DIFF
--- a/pkg/interface/src/views/components/AsyncButton.tsx
+++ b/pkg/interface/src/views/components/AsyncButton.tsx
@@ -29,7 +29,7 @@ export function AsyncButton({
   }, [status]);
 
   return (
-    <Button disabled={!isValid} type="submit" {...rest}>
+    <Button disabled={!isValid || isSubmitting} type="submit" {...rest}>
       {isSubmitting ? (
         <LoadingSpinner
           foreground={rest.primary ? "white" : 'black'}


matildepark (approved):
Seems fine — design feedback is sort of built-in for indigo components and disabled when submitting is expected

urcades:
Yeah, note the following Indigo structure for button styling:

[image: Screen Shot 2020-11-03 at 2.41.44 PM.png]

No dimming is implied here, but I think disabling the cursor via the css
property 'not-allowed'
<https://www.w3schools.com/cssref/playit.asp?filename=playcss_cursor&preval=not-allowed>
is enough to signal what's needed to be signaled here.

—
~fabled-faster
https://urbit.org/

tylershuster:
Image missing @urcades

urcades:
Ah, I forget the image discrepancy between responding via email and responding in github:

Screen Shot 2020-11-03 at 2 41 44 PM

matildepark (changes requested):
Holding off on merge until it resembles design — 
image

PR #3878, opened 2020-11-03
king: fix warnings; establish a stricter and more consistent warning policy

pilfer-pandex:
Following a complaint from @brendanhay, I've turned on -Wall in the package.yamls (except proto, and terminal-progress-bar, which doesn't even use stack) and fixed all the resulting failures and/or excluded certain warnings I found unproductive. Tested by rebuilding after `stack clean`.

Happy to receive feedback on what set of warnings is appropriate.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
           name: mars
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
-      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.1.2
+      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.2.0
         with:
           version: '290.0.1'
           service_account_key: ${{ secrets.GCS_SERVICE_ACCOUNT_KEY }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,7 +100,7 @@ jobs:
           name: mars
           authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
 
-      - uses: GoogleCloudPlatform/github-actions/setup-gcloud@0.2.0
+      - uses: google-github-actions/setup-gcloud@v0.2.0
         with:
           version: '290.0.1'
           service_account_key: ${{ secrets.GCS_SERVICE_ACCOUNT_KEY }}
--- a/pkg/hs/proto/lib/SimpleNoun.hs
+++ b/pkg/hs/proto/lib/SimpleNoun.hs
@@ -45,7 +45,8 @@ loob = \case
 
 textToAtom :: Text -> Atom
 textToAtom t = case N.textToUtf8Atom t of
-  N.A a -> a
+  N.A a   -> a
+  N.C _ _ -> error "textToAtom: nani!?"
 
 showA :: Atom -> String
 showA a = show (N.A a)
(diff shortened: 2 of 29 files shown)


jtobin:
+1 for "warnery"

brendanhay (approved):
LGTM :shipit:

brendanhay:
There's a few other issues like needing `-fno-warn-orphans` in test modules. I'm also unconvinced about disabling name shadowing and do bind warnings pervasively - is disabling those warnings on a per module basis an option?

pilfer-pandex:
Thanks for pointing out those issues with the tests. For some reason I did not catch by running `stack test`.

As for shadowing and do bind, I find these less problematic, but I'd love to hear why I'm wrong.
- With name shadowing, sometimes this is a thing you actively want to do, e.g. to signal you don't plan on using a variable anymore. (Arguably also, banning shadowing breaks a locality property and alpha equivalence, but whatever.)
- Is the fact that you've discarded a bind result that big a deal? If it's a mistake, won't you quickly find out because you'll find that you need to supply a value that you don't have?

brendanhay:
* Regarding unused do-binds: https://tech.freckle.com/2020/09/23/void-is-a-smell/
* Regarding shadowing: Yes, I agree - sometimes this is a thing you actively want to do, or the vagaries of do-notation leave you with various distasteful alternatives. That said, allowing shadowing by default hides _unintentional_ overloading - I typically leave the warning enabled by default and disable the warnings per module when it's actually necessary. YMMV. :shrug:

pilfer-pandex:
@brendanhay, I'm rather confused by the test failures here. I don't see how changing the warning settings in urbit-eventlog-lmdb could have changed how opening a db works. Additionally, these failures (except AppendHugeEvents, which also fails on release/next-vere on my machine) don't happen locally on my machine. Could it be a ci problem?

brendanhay:
@pilfer-pandex It looks like it's a `nix.useSandbox` issue on OSX specifically. By default your local `nix.conf` will have `use-sandbox = false` (you can check via `nix show-config`), but on CI we enabled sandboxing by default.

It's getting late here so I'll investigate and hopefully fix it tomorrow, CEST.

brendanhay:
(Also - we only enabled the sandboxing overnight, hence why you see the error now. Sorry!)

pilfer-pandex:
@brendanhay, this issue is still happening in this branch, but not release/next-vere. Do you have any insight into why? The latter branch is also running with sandboxing now, no?

I'm really baffled here.

brendanhay:
> Do you have any insight into why? 

Setting `nix.useSandbox = true` is not a build input to the tests derivation and doesn't cause a change to the output hash - meaning the build that succeeded on `release/next-vere` was with `nix.useSandbox = false` - since no inputs to the tests (ie. the king binary) have changed on that branch the tests haven't needed to be run. If I changed and GHC/Cabal options, for example, we'd have the same errors.

pilfer-pandex:
My understanding of your comment is that if I were to merge this, it would result in test failures because the superficial change in the lmdb package would cause the derivation to be run where previously it hadn't. Since the non-parametric sandbox setting has changed, this run will fail. So there will be test failures in release/next-vere if I merge, but there's nothing I can do within the scope of this patch to fix them. I'd like to get this in, but I don't want to "cause" failures in the main branch. I'm a bit unsure about what the right policy decision is here.

- should we turn sandboxing off?
- does some fix to the contents of the sandbox need to be made, so lmdb can run inside?
- does the merge of this patch actually have to wait for either of the above to go through?

I hope I'm not belaboring a silly point here; I just want to do the "right" thing. Should I merge?

brendanhay:
The right thing to do for all parties is to disable the tests for CI, get it merged, work on a longer term fix, and ensure the tests are run locally in the interim.

If you're fine with me pushing directly to this branch I'll push a change here to disable the tests only on CI, tomorrow.

pilfer-pandex:
Got it. Feel free to do whatever you want with this branch, and thanks so much.

PR #3879, opened 2020-11-03
vere: save snapshot upon completed replay

joemfb:
The previous behavior was to simply wait for the next snapshot timer, but it's better to save eagerly.


belisarius222:
I often Ctrl-D fakezods right after starting them because this feature was missing.

Fang- (approved):
Good change!

PR #3881, opened 2020-11-03
landscape: migrate global components to indigo-react

matildepark:
See urbit/landscape#50. Migrates the last pieces of our global components to indigo-react. In the process I found that a few of the remaining components weren't even used anywhere, so I took the opportunity to clear them out.

cc @tylershuster — VirtualScroller seems to work fine with `<Box>`, anything to be careful with?


matildepark:
Bigger diff than I expected, really.

pkg/interface/src/views/components/VirtualScroller.tsx
```diff
@@ -332,13 +333,13 @@ export default class VirtualScroller extends PureComponent<VirtualScrollerProps,
...
+      <Box overflowY='scroll' ref={this.setWindow.bind(this)} onScroll={this.onScroll.bind(this)} style={{ ...style, ...{ transform } }}>
+        <Box ref={this.scrollContainer} style={{ transform }}>
+          <Box style={{ height: `${origin === 'top' ? startgap : endgap}px` }}></Box>
           {indexesToRender.map(render)}
-          <div style={{ height: `${origin === 'top' ? endgap : startgap}px` }}></div>
-        </div>
-      </div>
+          <Box style={{ height: `${origin === 'top' ? endgap : startgap}px` }}></Box>
+        </Box>
+      </Box>
```
tylershuster, line 342: Code looks fine to me

tacryt-socryp (approved):

PR #3882, opened 2020-11-03
ren: remove the last renderer files from /ren

Fang-:
No longer in use.

Not sure who to tag on this, but considering half of these are publish-related, @ixv?


belisarius222 (approved):
Renderers are no longer supported by Ford.  I should have deleted these files during the Ford Fusion work.  LGTM

PR #3883, opened 2020-11-04
eyre: forego <atom> in favor of (scow %aura atom)

Fang-:
Pretty-printing is expensive, yet we do it whenever we construct the cookie string, at least once (but usually twice) per authenticated request.

Here we call out to the specific to-tape functions we need, instead of relying on the pretty-printer for converting... tapes to tapes, among other things.  
The primary gains come from the cookie-related instances, we update the others mostly for good style.

For the "receive request and immediately send response" case, that is processed synchronously within eyre and doesn't touch anything else (ie, client sends channel ack), speeds thing up by roughly 55%.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -462,7 +462,7 @@
     ==
     ;body
       ;h1:"Internal Server Error"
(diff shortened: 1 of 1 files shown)


eglaysher (approved):
Neat.

PR #3884, opened 2020-11-04
Ga/qa fixes

g-a-v-i-n:
- Resolves inconsistent spacing between status bar and home / content
- Leap positioning improved on mobile
- Reconnecting button refactored to use <Button />. Spinner added.
- Home is now "DMs + Drafts". I need extra review on that one because I also altered metadata around the codebase that seemed to point to this.

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -1693,9 +1693,9 @@
       "integrity": "sha512-3OPSdf9cejP/TSzWXuBaYbzLtAfBzQnc75SlPLkoPfwpxnv1Bvy9hiWngLY0WnKRR6lMOldnkYQCCuNWeDibYQ=="
     },
     "@tlon/indigo-react": {
-      "version": "1.2.12",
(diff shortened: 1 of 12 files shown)


matildepark (changes requested):
We somehow lost the status bar sigil by using Button instead of Row? Next userspace still has it showing correctly.

image

The Home changes seem to be in the right places, though wherever you see `icon="Circle"` we should probably change it to `icon="Boot"` to match.

g-a-v-i-n:
Who wrote this sigil component? I have no idea what this thing is doing.

matildepark:
It dates back to Modulo Chat, it's just been hacked at a lot since then.

g-a-v-i-n:
Why is mix-blend-mode used here?

matildepark:
Please see discussion in #2319.

g-a-v-i-n:
Ok, thanks. Think I found the issue.

matildepark (approved):

PR #3887, opened 2020-11-04
vere: remove dependency on nix-store path for static darwin binary

brendanhay:
Poached from @eglaysher's work in #3380.


joemfb (approved):
LGTM

PR #3891, opened 2020-11-05
landscape: restore groupify functionality

liam-fitzgerald:
Restores groupify functionality, accessible in the channel settings. Also comprehensively refactors our dropdown components and their internal logic.
Capture d’écran 2020-11-05 à 12 06 29 PM


matildepark (changes requested):
If you specify no group, it will redirect to `/~landscapenull/resource/` etc etc. and 404.

matildepark (approved):

PR #3894, opened 2020-11-05
ames: check cyphertext length

belisarius222:
Assert that the length of the cyphertext in an Ames packet must be between 1 and 1.024 bytes, inclusive.  An upcoming Vere PR will fix the jet mismatch in the AES-SIV crypto jet to prevent big mallocs.


joemfb (approved):
LGTM

joemfb:
@belisarius222, this is obsolete as of #4051, right?

belisarius222:
Yes.

PR #3895, opened 2020-11-05
lib-hooks: fix %kick bug

liam-fitzgerald:
- Fixes a bug where paths that had been kicked and rewatched would not receive new updates after receiving the backlog from the initial rewatch.
- Adds a ph test to validate that this rewatch-after-kick logic works correctly. 
- On migrate, kicks all rewatched paths so they can get the new updates


liam-fitzgerald:
cc: @ixv

tacryt-socryp (approved):

PR #3898, opened 2020-11-06
pull-hook: rewatch missing subscriptions

liam-fitzgerald:
The crashing of the %fact handling in the #3895 fix caused the subscriptions to be removed. Adds a migration case to readd any missing subscriptions


tacryt-socryp (approved):
Why were they missing in the first place?

liam-fitzgerald:
crashed fact handling because we weren't passing the bowl properly to the scry helper

PR #3899, opened 2020-11-06
 u3: refactors fatal exception handling in u3m_bail()

joemfb:
This PR refactors the nock runtime exception generator, `u3m_bail()`. A subset of bail motes (`%foul`, `%oops`, `%meme`) are always treated as fatal errors, but this was inconsistently handled: `abort()` in one case, `assert(0)` in another. (@brendanhay, I think this may have been the cause of the hanging-on-error behavior you observed, although I haven't been able to reproduce it.)

Additionally there was an assertion in the presence of a "namespace block" bail (`%need`), but that is obsolete (the interpreter bails with `[%1 (list path)]` in the case of a blocked namespace read).

### DIFF
--- a/pkg/urbit/noun/manage.c
+++ b/pkg/urbit/noun/manage.c
@@ -612,7 +612,6 @@ c3_w Exit;
 **    %intr               ::  interrupt
 **    %fail               ::  computability failure
 **    %over               ::  stack overflow (a kind of %fail)
-**    %need               ::  namespace block
 **    %meme               ::  out of memory
 **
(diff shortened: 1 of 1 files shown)


brendanhay (approved):
LGTM :shipit: 

Previously `assert(0)` should've called `abort()`, right? Or is that not a certainty?

joemfb:
It should've been, unless we were somehow building with assertions off. I don't yet have a better lead on those process hangs.

In any case, this was unnecessarily complex.

PR #3904, opened 2020-11-07
Webterm

Fang-:
[Demo video](https://pal.dev/share/webterm-demo.mov). Turns out I was running with the wrong binary, so there's no webslogs or styling in the "real" terminal. But it's 3am and you probably get the idea, so I'm not doing another take.

- Augments dill to allow subscriptions to arbitrary sessions' output.
- Redirects sessionless dill input to the default session.
- Adds `/app/herm`, a userspace proxy for dill's default session.
- Implements webterm, the landscape equivalent of `term.c`, in place of current webdojo. Supports styled text, opening URLs, a beep sound (@urcades -approved variant incoming), downloading files, and everything else you would expect.

Tagging @joemfb for arvo-side review. Note that the additions to the dill interface feel a bit half-baked. Next on my list is "make sessions more real", which should alleviate most concerns here.

Tagging @matildepark for landscape-side. Though as you might have noticed from the demo, frontend still needs some small tweaks. And I gotta replace "dojo" with "term" everywhere...


matildepark:
Very cool

urcades:
Yes

pkg/interface/src/views/apps/dojo/components/line.js
```diff
@@ -0,0 +1,66 @@
...
+          case 'y': prop.backgroundColor = 'yellow';  break;
+          case 'k': prop.backgroundColor = 'black';   break;
+          case 'w': prop.backgroundColor = 'white';   break;
+          default: console.log('weird back', part.stye.back);
+        }
+        if (Object.keys(prop).length === 0)
+        {
+          return part.text;
+        } else {
+          return (<Text mono fontSize='inherit' key={index+'-'+i} {...prop}>{part.text}</Text>);
```
liam-fitzgerald, line 49: drive-by: keys only need to be unique within their list, so you can drop `index` entirely as a prop here. Also, `React.memo` is a cleaner way to write this kind of component.

pkg/arvo/sys/zuse.hoon
```diff
@@ -1134,6 +1135,7 @@
           {$talk p/tank}                                ::
           {$text p/tape}                                ::
           {$veer p/@ta q/path r/@t}                     ::  install vane
+          [%view =duct]                                 ::  watch session blits
```
joemfb, line 1138: I'm not a fan of the ducts in these cards. It's not a pattern we've used elsewhere, it (at least arguable) conflicts with the purpose of ducts. It's a kind of fake, forward-looking modularity.

The design of %dill's namespace (`/dx/sessions//line`) suggests the intention of atomic session ids. This too is a kind of fake, forward-looking modularity, conflicting with the preceding instance, but arguably more defensible (what else would you do)?

I think it probably makes sense to reconcile these things now, and replace these ducts with atomic identifiers (`@tas` or `@ta`). We could use `%$` for the default session, as the new namespace does, and crash on any other identifiers until sessions are made real. Or, we could defer the session mechanism entirely, embracing The One Terminal and just send `[%view ~]` and `[%flee ~]` (ironically enough, the same noun).

pkg/interface/src/views/apps/term/components/input.js
```diff
@@ -0,0 +1,112 @@
...
+          id="term"
+          cursor={this.props.cursor}
+          onKeyDown={this.keyPress}
+          onClick={this.click}
+          onPaste={this.paste}
+          ref={this.inputRef}
+          defaultValue="connecting..."
+          value={prompt}
+        />
+      </div>
```
matildepark, line 107: You can do this like
```
<Row flexGrow='1' position='relative'>
<Box flexShrink='0'>
<Text mono id="prompt">
<BaseInput [etc etc]/>
```

pkg/interface/src/views/apps/term/store.js
```diff
@@ -0,0 +1,65 @@
...
...EEECUBYln03TTTdNBDZopopYvrTTdNa325mImNg3TTPV9q3pmY0xoO6bv3r00y+IDGid/9aaaZTGMuj9mpu9Mpio1dXrr5HERTZSmqU36A3CumzN/9Robv/Xx4v9ijkSRSNLQhAWumap82WRSBUqXStV/YcS+XVLnSS+WLDroqArFkMEsAS+eWmrUzrO0oEmE40RlMZ5+ODIkAyKAGUwZ3mVKmcamcJnMW26MRPgUw6j+LkhyHGVGYjSUUKNpuJUQoOIAyDvEyG8S5yfK6dhZc0Tx1KI/gviKL6qvvFs1+bWtaz58uUNnryq6kt5RzOCkPWlVqVX2a/EEBUdU1KrXLf40GoiiFXK///qpoiDXrOgqDR38JB0bw7SoL+ZB9o1RCkQjQ2CBYZKd/+VJxZRRZlqSkKiws0WFxUyCwsKiMy7hUVFhIaCrNQsKkTIsLivwKKigsj8XYlwt/WKi2N4d//uQRCSAAjURNIHpMZBGYiaQPSYyAAABLAAAAAAAACWAAAAApUF/Mg+0aohSIRobBAsMlO//Kk4soosy1JSFRYWaLC4qZBYWFRGZdwqKiwkNBVmoWFSJkWFxX4FFRQWR+LsS4W/rFRb/////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////VEFHAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAU291bmRib3kuZGUAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMjAwNGh0dHA6Ly93d3cuc291bmRib3kuZGUAAAAAAAAAACU=");
```
matildepark, line 23: man i sorta wanna stick this in a lib or something, huge base64 string in store.js lol

matildepark:
You should tombstone soto.hoon btw.

pkg/arvo/mar/belt.hoon
```diff
@@ -0,0 +1,29 @@
...
+    ^-  $-(^json belt:dill)
+    =,  dejs:format
+    %-  of
+    :~  aro+(su (perk %d %l %r %u ~))
+        bac+ul
+        ctl+`$-(json @c)`so
+        del+ul
+        met+`$-(json @c)`so
+        ret+ul
+        txt+`$-(json (list @c))`sa
```
joemfb, line 20: It's kind of wild that these casts work. But the code is wrong, `so` should be `(cu taft so)`. As for `%txt`, you may want to send an array of strings instead of a single string -- this would allow more flexibility in calculating the cursor position in the presence of combining codepoints. In which case, `sa` would be `(ar (cu taft so))`.

joemfb (changes requested):
This is great work -- I'm really excited to have parity in the browser console!

This PR obsoletes `:soto`, correct? Do we have a default/no-op agent with which we can replace it?

matildepark:
This certainly seems anomalous (prompt is invisible)

image

matildepark (changes requested):
Also you dropped an import!

```bash
ERROR in ./src/views/landscape/components/Content.js
Module not found: Error: Can't resolve '../../apps/dojo/app' in '/Users/matilde/git/urbit/pkg/interface/src/views/landscape/components'
 @ ./src/views/landscape/components/Content.js 28:0-42 47:46-53
 @ ./src/views/App.js
 @ ./src/index.js
```

Fang-:
> prompt is invisible

Ah. This is based off next-dill when it didn't have #3885 merged in yet. Having that, or just using next-vere, should fix this. I can merge that back into this branch for convenience.

pkg/arvo/mar/belt.hoon
Fang-: > you may want to send an array of strings instead of a single string -- this would allow more flexibility in calculating the cursor position in the presence of combining codepoints

Can you clarify a bit? Am I correct in reading "array of strings" as "array of strings that are, semantically, a single character"? I'm not sure what that would gain us over `(cu tuba sa)` here. Were you maybe thinking of doing the equivalent for _blits_ instead?

(I guess doing arrays for `%lin` blits (`(list @c)`) means we'd have to do them here, too, for consistency.)

pkg/arvo/sys/zuse.hoon
Fang-: Yeah, my plan here was for this to use ducts only as a means to get webterm working, and then to transition to using more formal session ids immediately afterward. This is indeed what the `/dx/sessions//line` hints at. iow the fakeness and forward-looking-ness are intentional.

I was gonna say, this is more a matter of whether or not to block the PR on this, but it seems that as long as I don't touch the runtime-facing interface, we can become 80% real with 20% of the work. I'll try and push on this a little harder. (:

joemfb: Real sessions will be awesome, but I don't mean to push you to do that now.

In case it wasn't clear, the suggestion of `[%view ~]` and `[%flee ~]` is simply for %dill to automatically use the duct of the single, axiomatically available terminal (a la `[%flog ...]`).

pkg/arvo/mar/belt.hoon
joemfb: `@c` is equivalent to `@t`, but the %dill protocol uses `(list @c)` (ie, `$tour`). The JSON equivalent of that is most likely an array of (usually individual) character strings. The reason to do that (and, I think, the reason for the factoring of the %dill protocol) is that combining codepoints equivalent to a logically-singular "character" can be controlled by the client. Note that cursor positions are calculated based on list items ...

If all that seems like too much to tackle now, you'll at least want to do something about this current `%txt` parser, which will only work for ASCII.

pkg/arvo/sys/zuse.hoon
Fang-: It's a good suggestion! Implemented as of the push just now.

pkg/arvo/mar/belt.hoon
Fang-: You are, once again, entirely right. Turns out cursor positioning was broken anyway. Fixed!

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -237,10 +237,11 @@
     =>  (se-born | %home %group-push-hook)
     (se-born | %home %group-pull-hook)
   =?  ..on-load  (lte hood-version %9)
-   (se-born | %home %graph-store)
+    (se-born | %home %graph-store)
   =?  ..on-load  (lte hood-version %10)
     =>  (se-born | %home %graph-push-hook)
-    (se-born | %home %graph-pull-hook)
+    =>  (se-born | %home %graph-pull-hook)
+    (se-born | %home %herm)
```
Fang-, line 244: Not sure what's up with the proliferation of versions here. The latest of these gets run on every `+on-load` call to hood, and remain silent even if the app has already been started.

pkg/arvo/mar/belt.hoon
ohAitch: `@c` is extremely not equivalent to `@t` what. `so:jo` produces `@t` not `@c`
`@t` is utf-8 encoded, `@c` is raw UTF-32 codepoints
```
> `@ux`~-~1f914.~21.~21.
0x21.0000.0021.0001.f914
> ? ~-1f914.~21.~21.
  @c
~-1f914.~21.~21.
> `@ux`(tuft ~-~1f914.~21.~21.)
0x2121.94a4.9ff0
> ? (tuft ~-~1f914.~21.~21.)
  @t
'🤔!!'
```

ohAitch: _In the case of ascii_, you can just `;;` between `(list @cF)` and `(list @tD)`(`tape`) and everything will appear to work.

joemfb: @ohAitch, I worded that very poorly. What I meant is that a) `@c` is of arbitrary length, like `@t`, and b) the `(list @c)` conveys cursor positions/glyphs (ie, it's not like a tape).

I endorse the changes in eb40e2feed0bbd1a357398672d78d766777d308f.

joemfb (approved):
The hoon side LGTM.

matildepark:
[Herm crashes on an attempted boot](https://hatebin.com/qaepmjjtfk). See extremely long stack trace. Also the prompt is still invisible.

joemfb:
Presumably these changes require a new pill.

Fang-:
Indeed it does. Apologies @matildepark, try again with this.

matildepark (commented):
- You'll want to add an icon catch in OmniboxResult to supply the Dojo icon. (you can see a bunch of examples for pointers there) 
- Leap navigates to ~term using `window.location` and not react-router, which implies it's not in `defaultApps` — either add a manual catch in Leap's `navigate()` or ensure defaultApps is using the same word for both the app and the title of the app (eg. "~terminal"), this sort of mismatch makes a lot of stuff in Landscape annoying (looking at you "Link / Links")
- Not everything is reflected, sadface.

Screenshot 2020-11-11 at 1 06 48 PM
Screenshot 2020-11-11 at 1 06 55 PM

urcades:
Hello all — I'm finalizing a larger map of the current state of Landscape, and I was able to quickly provide an indigo-spec reskin of the existing terminal interface:

Dojo

I only have two questions: 

1. Is this the correct place to bring up these design updates alongside shipping this PR? Or should I slot this into another landscape-related issue? cc @matildepark 

2. @Fang- was the audio file for terminal bell successfully situated within this PR? Let me know if there's remaining work I need to do to get that working.

Fang-:
> Not everything is reflected, sadface.

Webslogs will fix this. Running with a binary from `next-vere` should have them operate & show up fine. (Though I just realized I haven't tested them since making changes to blit format, so I'll give them another look-over.)

> reskin of the existing terminal interface

"blue for user input" is out of scope here, and something we'd want to implement on the dojo/per-app level anyway. (Web)term cannot distinguish one kind of display content from another, outside of the styling specified for that content by whoever produced it.

> was the audio file for terminal bell successfully situated within this PR?

Yes. Thanks! (:

matildepark (approved):

Fang-:
@urcades thinking about the design you presented a bit more, it's not entirely in line with what web term actually is. The prompt doesn't real, it's just a line in the scrollback.

It happens to behave the way it does because of details in drum (the thing that implements the prompt & application switching), but that could be subject to change as we continue work on the command line situation.

So, separating out the prompt visually is semantically correct right now, but that might change in the near-ish future, and it's not technically correct anyway.

PR #3905, opened 2020-11-08
vere: print error and exit if stdin is not a tty

joemfb:
We should tell people about `-t` instead of calling `c3_assert(0)`.


brendanhay (approved):
Nice 👌🏽

ohAitch:
Stupid question, what is the difference between `-t` and `-d`?

joemfb:
`-d` implies `-t`, and also forks into the background, exiting the foreground process once boot/restart is complete. This is convenient for scripting, and used in CI.

PR #3909, opened 2020-11-10
Links sigil size

tylershuster:
See https://github.com/urbit/landscape/issues/96#issuecomment-723200836 for an example of where this is heading


pkg/interface/src/views/apps/links/components/comment-item.tsx
```diff
@@ -0,0 +1,84 @@
+import React, { Component } from 'react';
```
matildepark, line 1: any reason you're de-componentising this

tylershuster: ...an accident? I don't think this is used anywhere, looks like a merge conflict

matildepark: oki

tylershuster:
image
image
image
image
image
image

tylershuster:
I realize the border isn't gray but...I can't get it to change colors. LMK why if you know why.

Also @urcades I thought these were the icon names that made sense (Copy, Menu), but they don't look quite like the mockup

urcades:
No worries — @g-a-v-i-n  and I synced on how we'd be handling Notebook/Collections "action" menus last week, and came upon a solution that - should - exist as a part of indigo now (menu styling) and doesn't involve icons:

Screen Shot 2020-11-16 at 1 18 19 AM

Also, we'll be using the "..." button rather than the "down chevron" to indicate these extra actions.

tylershuster:
Great; I'll worry about removing the icons and changing the menu icon. @g-a-v-i-n should I submit a separate PR for the indigo dropdown? Seems like this is a fairly consistent pattern across the interface.

tylershuster:
image
image

matildepark:
I'd just include the dropdown change here so we can holistically merge this.

tylershuster:
Sorry, to be more clear:

It seems like "Dropdown" is an Indigo pattern now and should be implemented in that repo, then instantiated here

matildepark:
> It seems like "Dropdown" is an Indigo pattern now and should be implemented in that repo, then instantiated here

@g-a-v-i-n thoughts on this?

tylershuster:
Ok, so as per the meeting today we're agreeing to go ahead with PRing the dropdown changes separately. Is this good to go as is?

matildepark:
I'll look through. You'll want to fix these commits in the meantime.

pkg/interface/src/views/apps/links/components/LinkItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+        lineHeight="tall"
+        display='flex'
+        flexDirection='column'
+        style={{ textDecoration: 'none' }}
+        href={contents[1].url}
+        width="100%"
+        target="_blank"
+        rel="noopener noreferrer"
+        p={2}
+        borderColor='washedGray'
```
matildepark, line 82:
```suggestion
        color='washedGray'
```

pkg/interface/src/views/apps/links/components/LinkItem.tsx
```diff
@@ -0,0 +1,136 @@
...
+        border={1}
+        borderRadius={2}
+      >
+        <Text overflow='hidden' style={{ textOverflow: 'ellipsis', whiteSpace: 'pre' }} mb={2}>{contents[0].text}</Text>
```
matildepark, line 86:
```suggestion
        <Text color='black' overflow='hidden' style={{ textOverflow: 'ellipsis', whiteSpace: 'pre' }} mb={2}>{contents[0].text}</Text>
```

matildepark (commented):
You can get around the borderColor shenanigans by using `color` and being specific with child text nodes. Otherwise this looks really nice, thank you.

matildepark:
Can you rebase the commits so I can merge?

tylershuster:
werkin on it

tylershuster:
Do another check before merging it, there's a few changes

tylershuster:
Agh, it doesn't work. sec. This has to do with unmanaged comments.

tylershuster:
Ok, now give it a shot

matildepark:
- Ah wait, there is one regression, we don't show previews for remote content links anymore. Can we restore that?
- Link pages are also a different width than the links lists, which is a little jarring. Can those be the same width?
- Not quite matching the mockup — we want Back to be bold with some margin to it a la:

image

tylershuster:
Ok, you'll want to check this one all around for `RemoteContent`. I think it looks really nice, but it does affect some global components

tylershuster:
image
image

urcades:
Beautiful

—
~fabled-faster
https://urbit.org/

tylershuster:
Without loading automatically turned on:

image

urcades:
Wow, I was actually just lamenting on how, while I loved the immediate visual unfolding, I felt weird about eviscerating the more hacker-news-feeling vibe to this interface, but you already considered this, it seems.

I'm in full support of this getting pushed, with the above in mind — I was worried this encroached on possible Posts functionality, but this still feels nicely unique, and evolves the module/channel type into a direction we've been prodding collections for a while.

matildepark:
Oof, a lot of moving parts. Ok, human selenium will do its best. I didn't know `ProfileOverlay` and the profile popover was here too! Have to try all this with diff content / avatars / etc

tylershuster:
I'm a little on the fence about the automatic unfolding in list view, because its appearance is tied to a global "automatically load images" toggle, which is nice in chat but can lose some of the "hacker news" vibe elsewhere, as you noted. One option you might consider while reviewing, @matildepark, is to only automatically expand when you view the item on its own.

As for the other parts, yes — all instances of someone else's sigils *should* now come complete with a popup card. This PR has certainly exceeded its scope, but I think it's all pretty necessary

urcades:
> This PR has certainly exceeded its scope, but I think it's all pretty necessary

Yeah, I feel like the observation here is that these 'simple' changes ended up being highly networked with a lot of other presentational aspects of  the interface ... good for future reference.

> I'm a little on the fence about the automatic unfolding in list view, because its appearance is tied to a global "automatically load images" toggle, which is nice in chat but can lose some of the "hacker news" vibe elsewhere, as you noted. 

This is exactly the core/source of my wariness here, glad to hear you feel it as well. For what it's worth, the end-goal for Collections was to have a mechanism by which the main view could be configured into a plain list, a more visual set of tiles/blocks, and even weirder/more free form structures. 

It's good to see this exist in a nascent form, but it clearly calls to mind we need an explicit toggle here for viewing the links, not rely on someone turning off "remote image fetching", which is an unrelated concern to how Collections present themselves.

We don't need to continue to bloat this issue with more adjacent development, but situating a "view preference" interface somewhere in here, inching towards Jimmy's total spec, is probably a good thing to situate next in Collections development.

matildepark:
Links is by far the least used module and automatic unfolding goes a very long way for its usability, so I don't think it's worth worrying about. Test upcoming.

vvisigoth:
thank you I want to love links

pkg/interface/src/views/apps/links/components/LinkItem.tsx
```diff
@@ -0,0 +1,154 @@
...
+        border={1}
+        borderRadius={2}
+        alignItems="flex-start"
+        overflow="hidden"
+      >
+        <RemoteContent
+          url={contents[1].url}
+          text={contents[0].text}
+          remoteContentPolicy={remoteContentPolicy}
+          style={{ marginTop: '-1px' }}
```
matildepark, line 88:
```suggestion
          style={{ marginTop: '-1px', alignSelf: 'center' }}
```

We want images to be centred in the item ... this suggestion however will also centre the title text as well. Any idea here?

pkg/interface/src/views/apps/links/components/LinkItem.tsx
```diff
@@ -0,0 +1,154 @@
...
+        color='washedGray'
+        border={1}
+        borderRadius={2}
+        alignItems="flex-start"
+        overflow="hidden"
+      >
+        <RemoteContent
+          url={contents[1].url}
+          text={contents[0].text}
+          remoteContentPolicy={remoteContentPolicy}
```
matildepark, line 87:
```suggestion
          remoteContentPolicy={remoteContentPolicy}
          unfold={true}
```

Removes "unfold" buttons so we see vids:

image

matildepark: Also these embeds should be 100% width if we can ... I can't seem to prompt it to do so, but Soundcloud does actually go full width.

pkg/interface/src/views/apps/links/components/LinkItem.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <Author
+        showImage
+        contacts={contacts}
+        ship={author}
+        date={node.post['time-sent']}
+        hideAvatars={hideAvatars}
+        hideNicknames={hideNicknames}
+        remoteContentPolicy={remoteContentPolicy}
+        group={group}
+        api={api}
```
matildepark, line 119: You're going to have to weave `history` in here because ProfileOverlay expects `history` for redirecting to the DM.

So you'll have to pass `history` from LinkResource to LinkItem to Author to OverlaySigil to ProfileOverlay, and so on anywhere you're using Author.

tylershuster: _case for state hooks intensifies_

tylershuster:
Ok, these have all been addressed

matildepark:
Testing, round two or three or whatever

matildepark:
image

Still lopsided.

tylershuster:
Ohhh, I see what happened there

pkg/interface/src/views/components/Author.tsx
```diff
@@ -1,28 +1,36 @@
...
 
-export function Author(props: AuthorProps) {
-  const { contacts, ship = '', date, showImage } = props;
-  let contact = null;
+export default function Author(props: AuthorProps) {
+  const { contacts, ship = '', date, showImage, hideAvatars, hideNicknames, remoteContentPolicy, group, api } = props;
+  const history = useHistory();
+  let contact;
   if (contacts) {
     contact = ship in contacts ? contacts[ship] : null;
```
matildepark, line 31: `contacts` here is assuming we have JUST the one group's contacts; but it's being passed the *entire* contacts prop. In LinkItem you will want to process the prop accordingly before passing it down. Note how this should say "galen" but doesn't — 

Screen Shot 2020-11-24 at 8 16 55 PM

matildepark:
I keep encountering an issue where I:

- Have a YouTube video as a link
- go to its comments, and the video embed is gone
- I go back to the links list and all video embeds are gone

Nov-24-2020 20-35-45


This doesn't happen in our music collection, which has an unbelievable like ~50 or so YouTube embeds that slow the browser to a crawl (we'll need pagination / lazy-loading badly but out of scope here?)

tylershuster:
Ugh, this looks like a memoization thing. But needs to be solved anyway.

I'm getting more and more skeptical of automatic unfolding....

tylershuster:
I think the basic problem is that `RemoteContent` isn't a `PureComponent` — it isn't purely a function of its props and state. It relies on a remote fetch. At least the fetch can be memoized (although, that's not really Correct)

matildepark:
Problem persists. `state.embed` becomes `undefined` as we rapidly `unmount` and `mount` it seems?

matildepark:
I think it's just because it happens in such quick succession.

tylershuster:
It seems that the memoized fetch was the culprit.

matildepark (approved):

PR #3910, opened 2020-11-10
graph-store: run-updates no longer pokes itself

liam-fitzgerald:
Runs all updates in one event, so as to not involve IO when it's unnecessary. 

Sidenote: The `update(resource.q resource)` pattern is suspicious to me, shouldn't we really be checking equality and crashing otherwise? Maybe useful for copying a graph I guess, but idk.

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -505,26 +505,23 @@
       ^-  (quip card _state)
       ?<  (~(has by archive) resource)
       ?>  (~(has by graphs) resource)
-      :_  state
-      %+  turn  (tap:orm-log update-log)
-      |=  [=time update=logged-update:store]
-      ^-  card
-      ?>  ?=(%0 -.update)
-      :*  %pass
-          /run-updates/(scot %da time)
-          %agent
-          [our.bowl %graph-store]
-          %poke
-          :-  %graph-update
-          !>
-          ^-  update:store
-          ?-  -.q.update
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/graph-store.hoon
```diff
@@ -505,26 +505,23 @@
...
-              %add-nodes          update(resource.q resource)
-              %remove-nodes       update(resource.q resource)
-              %add-signatures     update(resource.uid.q resource)
-              %remove-signatures  update(resource.uid.q resource)
-          ==
-      ==
+      =/  updates=(list [=time upd=logged-update:store])
+        (tap:orm-log update-log)
+      =|  cards=(list card)
+      |-  ^-  (quip card _state)
```
tacryt-socryp, line 511: Can you not do this with a `roll` on updates? I think it would be cleaner, since the type of the output result never changes (just a continual modification of the pair of `[cards state]`).

tacryt-socryp (changes requested):

pkg/arvo/app/graph-store.hoon
liam-fitzgerald: I dislike doing this for the following reason https://github.com/urbit/urbit/pull/2805#discussion_r416955058

tacryt-socryp: Makes sense.

tacryt-socryp (approved):

PR #3913, opened 2020-11-11
hark: first QA pass

matildepark:
Handful of FE tweaks for obvious QA fixes. Still need to amend some behaviours; see if we can show invites in the notifications menu (as per @vvisigoth), etc.


liam-fitzgerald (approved):
LGTM!

PR #3914, opened 2020-11-11
channel.js: Enqueue subscriptions to avoid hitting connection limit

liam-fitzgerald:
One of our chief performance issues is the way we make subscriptions, using a new PUT request for each watch request. This causes a waterfall-type effect (see image below) where we hit the max number of connections to our ship (usually 6 on desktop, 4 on mobile). Therefore, we now enqueue the bare minimum to load the homescreen (metadata, invites, weather, etc.) in a single batch. We then load in the rest of the data later, in another batch. 

This improved our time to paint a full launch from 9s to 3s over a localhost connection. Mobile clients are likely to benefit the most from this, given their generally lower connection limit.

Capture d’écran 2020-11-11 à 3 49 43 PM


pkg/arvo/app/landscape/js/channel.js
```diff
@@ -159,21 +187,29 @@ class Channel {
...
+        {action: "ack", "event-id": parseInt(this.lastEventId)}
+      ];
       if(j) {
         payload.push(j)
       }
       let x = JSON.stringify(payload);
       req.send(x);
 
+      this.outstandingJSON = [];
       this.lastEventId = this.lastAcknowledgedEventId;
```
Fang-, line 213: Might be better to factor this to come after/outside of the if-else here, since it's now (even more obviously) shared logic between the two branches.

pkg/interface/src/logic/subscription/global.ts
```diff
@@ -43,18 +43,27 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
...
     this.subscribe('/all', 'metadata-store');
-    this.subscribe('/all', 's3-store');
-    this.subscribe('/all', 'launch');
-    this.subscribe('/all', 'weather');
-    this.subscribe('/keys', 'graph-store');
-    this.subscribe('/updates', 'hark-store');
-    this.subscribe('/updates', 'hark-graph-hook');
-    this.subscribe('/updates', 'hark-group-hook');
-    this.subscribe('/updates', 'hark-chat-hook');
+    this.subscribe('/all', 'invite-store', true);
```
Fang-, line 47: If you want this to go in the same batch, shouldn't the `metadata-store` subscription have the `queue` flag set also?

pkg/interface/src/logic/subscription/global.ts
```diff
@@ -61,7 +61,7 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
       this.subscribe('/updates', 'hark-chat-hook', true);
       this.dequeue();
 
-    }, 1500)
```
Fang-: I'm not sure I understand why there's a delay here in the first place. Don't you always want the context's connection limit to be the case, rather than waiting on this timer in the "high connection limit" and "fast response" cases?

Fang- (commented):
>sneaking in the ack timer change

Sneaky! But good change regardless. Have you tested it behaves fine in the "I joined a new chat and am getting pushed 100+ messages" case?

Upon reflection, I'm not 100% confident the clog logic will even _let_ you get those. Right now, the logic triggers when "subscription open >30s and >50 events unacked", but in the above case maybe you want something more like ">50 events unacked for >30s". Let me know how this works out in practice!

pkg/arvo/app/landscape/js/channel.js
```diff
@@ -71,9 +75,16 @@ class Channel {
...
+    if(this.lastAcknowledgedEventId === this.lastEventId) {
+      return;
+    }
+    this.sendJSONToChannel();
+  }
+
   //  sends a poke to an app on an urbit ship
   //
-  poke(ship, app, mark, json, successFunc, failureFunc) {
+  poke(ship, app, mark, json, successFunc, failureFunc, queue = false) {
```
tacryt-socryp, line 87: I don't think this 'queue' variable should be a part of Channel.js's API. I think we should either handle the enqueue and dequeue logic *entirely* in Channel.js (and opaquely), or entirely in Landscape. I lean towards Channel.js handling the enqueue and dequeue logic completely. Mixing the responsibilities of Landscape and Channel should be avoided.

pkg/interface/src/logic/subscription/global.ts
tacryt-socryp: Why do we even need to batch the requests if they're in a queue?

tacryt-socryp (changes requested):

matildepark:
bump

pkg/arvo/app/landscape/js/channel.js
```diff
@@ -217,8 +257,13 @@ class Channel {
           funcs["subAck"](obj);
         }
       } else if (obj.response == "diff") {
-        // ack subscription
-        this.sendJSONToChannel();
+        // ensure we ack before channel clogs
+        if((this.lastEventId - this.lastAcknowledgedEventId) > 30) {
+          clearTimeout(this.debounceTimer);
```
tacryt-socryp, line 262: Identical code to this.clearQueue(), so replace it with the method call.

tacryt-socryp (approved):
LGTM, one nit that should be addressed before merge. Otherwise, bring on the perf gains!

vvisigoth:
plz
—
~poldec-tonteg
http://urbit.org

matildepark:
Merge conflict

matildepark:
Standby, will resolve

PR #3916, opened 2020-11-12
king: Rationalize implementation of mug

pilfer-pandex:
- implement a mugBS that doesn't copy
  - use it in ames packet processing
- switch the hash stored structurally in nouns to be the mug
  - actually compute it lazily (use of StrictData in this file means the hash was previously computed at noun construction time, despite appearances)


joemfb (approved):
LGTM

pilfer-pandex:
@brendanhay Here too.

PR #3922, opened 2020-11-12
hark: pass remotecontentpolicy to chat notifications

matildepark:
Fixes an interface crash when receiving rich media in notifications.

image

### DIFF
--- a/pkg/interface/src/views/apps/notifications/chat.tsx
+++ b/pkg/interface/src/views/apps/notifications/chat.tsx
@@ -34,8 +34,9 @@ export function ChatNotification(props: {
   contacts: Rolodex;
   groups: Groups;
   api: GlobalApi;
+  remoteContentPolicy: any;
 }) {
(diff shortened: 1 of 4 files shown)


matildepark:
Actually would this need to be passed to Archive too or archived rich media notifications would crash?

tylershuster:
@liam-fitzgerald we really have to make the useRemoteContentPolicy hooks; this prop stuff is insane

matildepark:
> Actually would this need to be passed to Archive too or archived rich media notifications would crash?

Just checked, archive works fine. OK this is good to go. Needs a stamp from anybody. I asked @philipcmonk a few hours ago about a potential next-sys deploy; if there isn't one this can go out basically asap but otherwise it's better to just ship with that.

Fang- (approved):
Not the one you asked, but you seem antsy to get this out. It looks Fine!

PR #3927, opened 2020-11-12
graph-store: amend %publish and %link validators to allow editing comments

### LINKED ISSUE #3846: publish, links: user can't edit or delete their own comments
A user can't edit or delete their own comments on notebook posts.

Expected Behavior:
A user should be able to delete or edit their own comments on notebook posts.

Repro:
1. Go to any existing notebook.
2. Go to a post in that notebook.
3. Post a comment.
4. Go back to your comment.  You''ll see a Delete option that can't be clicked, and there is no Edit/Update option.

 - OS: Windows 10
 - Browser: Brave, Chrome, Firefox
 - Base hash of your urbit ship: 4v63o

1e19282c6452ee8b77f8714c3b21f782

tacryt-socryp:
We now allow editing of comments in the `%graph-validator-link` and `%graph-validator-publish` marks, and have support for comment revisions. As a part of this, we throw away our existing update-logs, and change the behavior of `%graph-push-hook` to check for an empty update log. When we have an empty update log, we hand the requester of an initial-watch the whole graph with an `overwrite` field set to `%.y` so as to ensure they now have our latest data in our latest format. We wipe their update-log in this overwrite case, as it adds a great deal of complexity to attempt to update the update-logs in place to match new mark validators.

I have tested that the migration code in the `+on-load` does pass each of the corresponding `validator` marks on livenet. I need to update the frontend to work with the new publish and graph data structures, but the backend of this is complete.

Fixes #3846.

### DIFF
--- a/pkg/interface/src/views/components/Comments.tsx
+++ b/pkg/interface/src/views/components/Comments.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import bigInt from 'big-integer';
 import { Col } from '@tlon/indigo-react';
 import { CommentItem } from './CommentItem';
 import CommentInput from './CommentInput';
@@ -7,6 +8,7 @@ import GlobalApi from '~/logic/api/global';
 import { FormikHelpers } from 'formik';
 import { GraphNode } from '~/types/graph-update';
 import { createPost, createBlankNodeWithChildPost } from '~/logic/api/graph';
+import { getLatestCommentRevision } from '~/logic/lib/publish';
 import { LocalUpdateRemoteContentPolicy } from '~/types';
 import { scanForMentions } from '~/logic/lib/graph';
 
@@ -14,6 +16,8 @@ interface CommentsProps {
   comments: GraphNode;
   name: string;
   ship: string;
+  editCommentId: string;
+  baseUrl: string;
   contacts: Contacts;
   api: GlobalApi;
   hideAvatars: boolean;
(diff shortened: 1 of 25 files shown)


liam-fitzgerald (approved):
LGTM! lmk when the accompanying FE changes drop. I have no issue with dropping the update logs now, but I'd like to make clear my conceptual opposition to it in future, especially given chat on graph-store and any potential federation work.

tacryt-socryp:
I have opposition to it as well. Further, I oppose mark validator upgrades.

tacryt-socryp:
This is ready for frontend review @matildepark. Backend already reviewed. I've tested that both Links and Publish comment revisions work.

pkg/interface/src/views/components/Comments.tsx
```diff
@@ -40,22 +48,75 @@ export function Comments(props: CommentsProps) {
...
+        <CommentInput
+          onSubmit={onEdit}
+          label='Edit Comment'
+          loadingText='Editing...'
+          initial={commentContent}
         />
-      ))}
+      ) : null )}
+      {Array.from(comments.children).reverse()
+        .filter(([idx, _]) => idx.toString() !== props.editCommentId)
```
matildepark, line 103: Why do we pop it out when we're editing? It's sorta confusing in practice — could we just fade it as if it's "pending" while we're editing it?

tacryt-socryp: Sure, sounds fine.

PR #3930, opened 2020-11-13
launch: click to copy base hash

tylershuster:
Fixes https://github.com/urbit/landscape/issues/162


vvisigoth:
ThankS!

A
—
~poldec-tonteg
http://urbit.org

matildepark (approved):
The LoC here are misleading, we should just align the linter between your editor and eslint's config. Fine otherwise

PR #3933, opened 2020-11-15
hoon: update glyph names (continued)

jfranklin9000:
See: https://github.com/urbit/urbit/pull/3787

I tested by deleting the deprecated parsers in `hoon.hoon`, although that's not part of the PR.

I made a fakezod with this PR and the deletions.
```
urbit -A pkg/arvo -B bin/solid.pill -F zod
```

Then I made a new pill and made a fakezod from that too.
```
.fixup/pill +solid
^d

urbit -B zod/.urb/put/fixup.pill -F bus
```

This all worked.

I will start the conversion of `%bscl` to `%bccl`, etc.
I didn't commit a pill.
Is this the correct branch for this PR?

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -534,7 +534,7 @@
     ::     ;~(pfix ace ;~(plug i.opt $(opt t.opt)))
     ::   --
     ::
-    ++  group  ;~((glue net) ship sym)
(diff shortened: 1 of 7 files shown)


philipcmonk (approved):
Yes, this is the correct branch.  Looks good!

The conversion of the stems of +hoon is harder to deploy over the air, but we're planning a network breach in a couple weeks, so if you can get it done soon it would be trivial to include.

PR #3934, opened 2020-11-15
build: github actions workflow configuration

brendanhay:
Due to Hercules CI's non-support for automatically building external PRs and security concerns around leaking secrets when running artefact upload and other effects only using nix sandboxing, I spent a week or so migrating the Hercules migration to GitHub Actions.

In comparison to Hercules:

* External PRs will automatically be built.
* There is no Hercules evaluation step, immediately saving around 2-3~ minutes.
* All prior built artefacts are pulled from Cachix over HTTP. Previously they were just present on disk in the `/nix/store`. This seems to add around 1-5~ minutes of "cache setup" to each separate job.
* The Azure`DSv2` instances are only 2 core with a lower clock speed - in particular the Darwin builder seems to suffer.
* There is no need for sandboxing on the Darwin builder - this should alleviate the LMDB `mmap()` test issues observed in #3878 and #3019.

Overall the change is pretty simple - all of the work was done previously in preparing the derivations for Hercules. There's a bit of YAML nonsense, but due to GitHub actions support for Cachix it's a little simpler than the corresponding Travis' configuration, IMO.

i plan to add the bare-metal Linux and Darwin machines used as Hercules agents as additional build runners for these GitHub Actions in a separate PR. To avoid the security implications of running untrusted code they will be limited only to members of the GitHub organisation. The AArch64 support will then either use a bare-metal runner subject to the same caveat or I'll setup an Azure/AWS ARM instance to run containerised builds, which would support building external PRs.


joemfb (approved):
LGTM.

With these change, CI again has access to git commit metadata, tags, &c, so we could using those things to name release artifacts (or aliases to them), and change behavior between push/pr/tag builds. This is not urgent (we can do the next release(s) manually), but could make things smoother.

@philipcmonk, PTAL.

philipcmonk (approved):
Very nice!

brendanhay:
I'll start work on a separate PR for the analogous `deploy` step.

Issue #3929, opened 2020-11-13
Notifications show new notifications dot despite all being marked read


mike-cr:
**Describe the bug**
There's a blue dot next to the leap button despite all notifications being marked read. I think it might have got messed up by the since-fixed crash-on-remote-content bug

**To Reproduce**
1. Not sure...

**Expected behavior**
There should not be a blue dot if there are no unread notifications

**Desktop (please complete the following information):**
 - OS: Arch linux 64-bit
 - Browser qutebrowser (qtwebengine)
 - Base hash 0v14.pk93p.3v3pd.gomnc.ut49c.31ca7.aqqvh.sg72n.o24vc.qa0bk.tdo1o

Fang-:
I got this too, even before seeing that remote content crash, so I don't think it's related.

vvisigoth:
I've seen this but now can't repro @liam-fitzgerald what should we do if we see this in the wild to at least isolate if it's on front end or back end.

matildepark:
I can reproduce this often; I fix it by archiving all the notifications I can see, then refresh and come back and there's my new notification. I think #3935 fixes this though.

mike-cr:
I tried archiving everything and now it crashes:

```
Error
“Cannot read property 'group-path' of undefined”
Stack trace
TypeError: Cannot read property 'group-path' of undefined
    at by (https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:184:263175)
    at Ko (https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:37:57929)
    at gs (https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:37:104168)
    at cl (https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:37:96716)
    at sl (https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:37:96641)
    at Js (https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:37:93671)
    at https://landscape.tinnus-napbus.xyz/~landscape/js/bundle/index.b23d5c0044c48b1d2bab.js:37:45313
[...]

Issue #3924, opened 2020-11-12
hark: accepting gropu invite from inbox doesn't redirect to group or give confirmation of success


vvisigoth:
**Describe the bug**
If you accept a group invite in inbox, it just sits in the inbox. The tile does, however, appear on homescreen.

**To Reproduce**
1. Receive a group invite
2. Click accept
3. It disappears

**Expected behavior**
It should redirect to group (I think)

**Base**

`tdo1o`

tylershuster:
I can take this one

vvisigoth:
thanks @tylershuster

matildepark:
Fixed in #3935.

Issue #3920, opened 2020-11-12
[notifications] opening a notification marked read toggles its read state


mike-cr:
**Describe the bug**
If you have a notification in yoiur inbox that's marked as read, opening it will toggle it back to unread

**To Reproduce**
1. Go to notifications inbox
2. Click on a message you're already marked as read
3. observe it's now unread

**Expected behavior**
opening a read notification should not mark it unread

**Desktop (please complete the following information):**
 - OS: arch linux 64-bit
 - Browser qutebrowser (qtwebengine)
 - Base hash of your urbit ship. Run `+trouble` in Dojo to see this.

**Smartphone (please complete the following information):**
 - Base hash 0v1r.kh31a.3lh29.dkdc0.m61ud.o9v6j.dogmk.i83bn.e063g.k6vjo.lg6c2

**Additional context**
As an aside - is "archive" meant for read messages or for things I specifically want to save?

matildepark:
Fixed in #3935.

PR #3935, opened 2020-11-16
hark: rework caching to allow for better unread counts, read all by index

liam-fitzgerald:
Reworks the caching so that we keep a `(jug index @da)` as a cache of what indices have unread notifications in what timeboxes. Uses this cache to unify the unread handling and add a `%read-index` poke, which allows us to automatically read notifications when we mark a chat as read. Additionally, adds unread counts to group tiles as seen below.

Also includes assorted bugfixes:
- Cleans up UX of accepting a group invite from the invite (#3924 )
- Does not mark notifications as unread when their content is navigated to (#3920 )
- Fixes ordering and loading of notifications (#3929 )
- Fixes an issue where the color of the transparency on publish note notifications would not adjust with theme
- Enables scrollback on archived notifications

Capture d’écran 2020-11-16 à 12 25 51 PM

### DIFF
--- a/pkg/interface/src/logic/reducers/hark-update.ts
+++ b/pkg/interface/src/logic/reducers/hark-update.ts
@@ -6,19 +6,21 @@ import {
 } from "~/types";
 import { makePatDa } from "~/logic/lib/util";
 import _ from "lodash";
+import {StoreState} from "../store/type";
 
-type HarkState = {
-  notifications: Notifications;
-  archivedNotifications: Notifications;
-  notificationsCount: number;
-  notificationsGraphConfig: NotificationGraphConfig;
-  notificationsGroupConfig: GroupNotificationsConfig;
-  notificationsChatConfig: string[];
-};
+type HarkState = Pick<StoreState,
+  "notificationsChatConfig"
+  | "notificationsGroupConfig" 
+  | "notificationsGraphConfig" 
+  | "notifications"
+  | "notificationsCount"
+  | "archivedNotifications"
+  | "unreads">;
 
 export const HarkReducer = (json: any, state: HarkState) => {
   const data = _.get(json, "harkUpdate", false);
   if (data) {
+    console.log(data);
     reduce(data, state);
   }
   const graphHookData = _.get(json, "hark-graph-hook-update", false);
@@ -139,16 +141,31 @@ function reduce(data: any, state: HarkState) {
   timebox(data, state);
   more(data, state);
   dnd(data, state);
-  count(data, state);
   added(data, state);
-  graphUnreads(data, state);
+  unreads(data, state);
 }
 
-function graphUnreads(json: any, state: HarkState) {
-  const data = _.get(json, 'graph-unreads');
+function unreads(json: any, state: HarkState) {
+  const data = _.get(json, 'unreads');
   if(data) {
-    state.graphUnreads =  data;
+    data.forEach(({ index, unread }) => {
+      updateUnreads(state, index, x => x + unread);
+    });
   }
+}
+
+function updateUnreads(state: HarkState, index: NotifIndex, f: (u: number) => number) {
+      state.notificationsCount = f(state.notificationsCount);
+      if('graph' in index) {
+        const curr = state.unreads.graph[index.graph.graph] || 0;
+        state.unreads.graph[index.graph.graph] = f(curr);
+      } else if('group' in index) {
+        const curr = state.unreads.group[index.group.group] || 0;
+        state.unreads.group[index.group.group] = f(curr);
+      } else if('chat' in index) {
+        const curr = state.unreads.chat[index.chat.chat] || 0
+        state.unreads.chat[index.chat.chat] = f(curr);
+      } 
 }
 
 function added(json: any, state: HarkState) {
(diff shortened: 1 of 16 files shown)


liam-fitzgerald:
cc: @urcades lmk if design has any opinions on presenting unread counts in group tiles

urcades:
@liam-fitzgerald Yo!

So, our latest prototype depicts the home/launch screen as so:

Screen Shot 2020-11-15 at 11 43 55 PM

So, simple change: Let's make these blue. 

What needs to be determined now is how exactly this is set/toggled/configured! As I see it there are essentially three options:

1. "Display total update count on group tile" toggle located in the existing notification preferences:
Screen Shot 2020-11-16 at 12 31 37 AM

2. Display the same toggle in one's "Ship Settings":
Screen Shot 2020-11-16 at 12 50 27 AM

3. Push this update **without** the ability to toggle it: This essentially opts people into receiving visual tile notifications as long as they've subscribed to updates — we can think of this approach as analogous to the blue dot, which shows the moment you've subscribed to updates and receive an update. We could consider hiding the tile notifications as an extension to the "do not disturb" option, but this seems somewhat obscure.

I think of these three options, locating the setting in the notifications interface (option 1) is currently the most **obvious** approach given the settings that already exist in the inbox. It should probably be toggled "on" by default.

cc'ing @g-a-v-i-n here because he's led the work on notifications and likely has an opinion here!

urcades:
Something about approach #3 is also appealing to me, only for the reason that it's one less toggle/settings item for someone to worry about. If you've subscribed to notifications, you receive a tile count.

I can see people who are more calm-inclined to bemoan this, but anyone opting into notifications in the first place probably just wants to set and forget, not fiddle with toggles.

liam-fitzgerald:
worth noting that hark only tracks unread counts for graphs that have been subscribed to, so we'd have 3 effectively by default.

urcades:
That's really good to know, 3) seems that much more appealing now — I'd wait to hear gav on on any configuration thoughts he might have, still!

galenwp:
My take:

I'd like to see unread counts on tiles for groups where I'm not subscribed to notifications, I just think they should be grey instead of blue. There are groups I'm intentionally still in but don't follow closely. It could be nice to let these counts accumulate as a way of passively following groups without having them end up in your inbox.

vvisigoth:
FTR I think that liam is referring to _channels_ that have been subscribed
to. Is that correct @liam-fitzgerald?
—
~poldec-tonteg
http://urbit.org

pkg/interface/src/logic/reducers/hark-update.ts
```diff
@@ -6,19 +6,21 @@ import {
...
+  | "notificationsGraphConfig" 
+  | "notifications"
+  | "notificationsCount"
+  | "archivedNotifications"
+  | "unreads">;
 
 export const HarkReducer = (json: any, state: HarkState) => {
   const data = _.get(json, "harkUpdate", false);
   if (data) {
+    console.log(data);
```
matildepark, line 23: ( ͡° ͜ʖ ͡°)

pkg/interface/src/logic/reducers/hark-update.ts
```diff
@@ -6,19 +6,21 @@ import {
 } from "~/types";
 import { makePatDa } from "~/logic/lib/util";
 import _ from "lodash";
+import {StoreState} from "../store/type";
```
matildepark, line 9: linter asks that we put spaces between the brackets here — maybe start running eslint before committing?

matildepark (commented):
JS side reads fine

vvisigoth:
Any ETA on getting this out?

matildepark:
I’m asking L for a review today and testing tomorrow for tomorrow night deploy. I would like a few other fixes in as well for Hark and our graph apps.
—
~haddef-sigwen
https://urbit.org

vvisigoth:
""dope""
—
~poldec-tonteg
http://urbit.org

urcades:
Very Dope.

Quick note: 

We had a design review yesterday which touched on the above thoughts I posted for tile notifications.

Some edits to be performed immediately:

Screen Shot 2020-11-17 at 1 22 51 PM

We can opt to revert the color of these tile notifications to gray60/gray, for one.

And we can state plainly that for ease of development, these tile notifications are supplied as described in approach 3) stated above, and repeated + modified for clarity here:

User flow: 

1. I join a group, and its tile appears in my home screen. As long as I am unsubscribed to any channels within the group, I see no dot notification, and I see no tile notifications:

Screen Shot 2020-11-17 at 1 30 30 PM

2. The moment I subscribe to a channel within the group, I am asked if I'd like to subscribe to updates from the channel:

Screen Shot 2020-11-17 at 1 29 24 PM

3. If I keep the box unchecked, my state will be the same as Step 1 above. If I check the box, any new posts to the channel will result in items a) populating my updates inbox, b) triggering the appearance of the blue dot, and c) incrementing the tile notification by the amount of updates received

From here on out, the "Home Screen" can be used to quickly ascertain activity occurring in all groups you have joined, whose channels you have subscribed to.

urcades:
A near-term consideration here links to the statement @galenwp made above, which is to possibly introduce introduce the capability to display the TOTAL NUMBER of ALL channel updates within a group on a tile, while still only subscribing to a subset of this TOTAL NUMBER, this subset of which is sent to the Inbox.

matildepark:
@liam-fitzgerald This PR will also want to include the restoration of "sorting by last updated" behaviour in the sidebars for graph resources, if it doesn't already...

tacryt-socryp (approved):
LGTM, discussed with Liam

matildepark:
Conflict on this btw.

matildepark:
Good to merge?

liam-fitzgerald:
yep!

PR #3936, opened 2020-11-16
launch: puts base hash under tiles

tylershuster:
Fixes https://github.com/urbit/landscape/issues/166

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -64,31 +64,31 @@ export default function LaunchApp(props) {
           />
           <Box display={["none", "block"]} width="100%" gridColumn="1 / -1"></Box>
           <Groups groups={props.groups} associations={props.associations} />
+          <Box width="100%" gridColumn="1 / -1"></Box>
+          <Box 
+            justifySelf="start"
+            fontFamily="mono"
+            left="0"
+            bottom="0"
+            color="gray"
+            bg="white"
+            borderRadius={2}
+            fontSize={0}
+            p={2}
+            cursor="pointer"
+            onClick={() => {
+              writeText(props.baseHash);
+              setHashText('copied');
+              setTimeout(() => {
+                setHashText(props.baseHash);
+              }, 2000);
+            }}
+          >
+            {hashText || props.baseHash}
+          </Box>
         </Box>
       </ScrollbarLessBox>
-      <Box
-        position="absolute"
-        fontFamily="mono"
-        left="0"
-        bottom="0"
-        color="gray"
-        bg="white"
-        ml={3}
-        mb={3}
-        borderRadius={2}
-        fontSize={0}
-        p={2}
-        cursor="pointer"
-        onClick={() => {
-          writeText(props.baseHash);
-          setHashText('copied');
-          setTimeout(() => {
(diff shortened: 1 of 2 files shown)


Fang-:
Heads up that rendering it behind the tiles would prevent me from viewing it on mobile. (Maybe I could see through the gaps in the tiles while scrolling, but, yeah.)

signal-2020-11-16-194945

matildepark:
Maybe we should just stick it in the layout below all the tiles.

tylershuster:
I just put some padding at the bottom of the box so you can always scroll up to see it.

matildepark:
Sneaky little base hash! 
image

Honestly though, it's sounding like a better solution is just putting it in the layout. Though on desktop we don't want it to be halfway down the screen. Perhaps we do the `position: absolute` stuff only on larger viewports.

tylershuster:
image

matildepark:
> Though on desktop we don't want it to be halfway down the screen. Perhaps we do the position: absolute stuff only on larger viewports.

I still stand by this. It's a UI element that should be separated from the content layout; it can append to the bottom on mobile but in general we want to preserve how it acts now on desktop viewports. @urcades

tylershuster:
On larger viewports the  hash can still be within the realm of the group tiles. It certainly can on mine. Fact is, there's no concrete metric as to when to position it absolutely because even on an HD screen there could be so many groups that we start to see the overlapping behavior.

The solution would be to be something like

```
container { display: grid; min-height: 100% }
groups { stick-to: top }
hash { stick-to: bottom }
```

But there's not really a way to do this and keep the behavior we want with the tiles spanning the full width. You *should* be able to set `align-self: flex-end` on the hash and get it to go to the bottom of the container but alas, setting `align-contents: flex-start` on the container precludes any possibility of that. Frankly I don't see a problem with the screenshot in "sneaky little base hash" as long as scrolling all the way to the bottom lets you see it in full.

urcades:
Such a fiddly interface, huh — didn't realize this one would be giving us so much trouble..

I just spoke to Matilde about this and feel similarly to her that retaining the current behavior is fine — keeping the desktop-viewport-sized HUD quality of the base hash is fine.

On mobile however, let's skip juggling around z-indices and worrying about overlap at all — let's just stack the base hash as we do the tiles, and have it essentially render as a footer element in smaller viewports.

To be honest, while I admit I'm the one who flagged this as a usability concern (and it still is), I can imagine it being a moot issue when we divert group invites into the notifications inbox. Situating group invites on the home screen was always a temporary crutch/fix, and we really shouldn't sweat over the base hash positioning all that much.

urcades:
Clearer illustration of what I mean by "Keeping the existing large viewport behavior intact":

Kapture 2020-11-18 at 01 05 36

Note here, essentially all tiles, if overflowing the viewport, scroll as one unit — the header and footer are actually static elements, but the padding of the latter end of the tile-scroll is 16px, which gives it an even-looking appearance when the bottom of the tile scroll is reached.

urcades:
In contrast, the mobile view simply stacks the element:

Kapture 2020-11-18 at 01 21 31

Excuse the general roughness of these prototypes, they are simply serving to illustrate the positioning of the hash display with respect to the scroll, nothing more.

PR #3937, opened 2020-11-16
interface: bump indigo to 1.2.14, adjust flexes

liam-fitzgerald:
Bumps indigo to 1.2.14 and adjusts flexes to accomodate the `flex-shrink: 0` change. I realise this change might be slightly contentious, so I'd like to justify it in full.

### The problem with flex-shrink

With larger, layout-type components (think inputs in a form), you never want flex-shrink because you want it to scroll. Because of the unpredictable interaction of `min-{height,width}` and flex-shrink (behaviour differs very heavily on Safari, Chrome and FF), flex-shrink means that instead of having to test at a handful of screen width based on your media queries, the screen height is an important part of the equation, and layouts can break on only some combinations of screen dimensions. 

Furthermore, `flex-shrink: 1` (the user agent default as per the CSS spec), makes setting a percentage height/width ambiguous. In this PR there are some changes to heights to make containers size correctly. Setting `height="100%"` on a container that will *never* be the full height of its parent is ambiguous at best and downright confusing at worst. Instead, consider `flex-grow: 1` or a calculated height.

The number of deletions in this PR is also a decent argument for why flex-shrink should be off by default. 

cc: @tacryt-socryp @tylershuster

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -1693,9 +1693,9 @@
       "integrity": "sha512-3OPSdf9cejP/TSzWXuBaYbzLtAfBzQnc75SlPLkoPfwpxnv1Bvy9hiWngLY0WnKRR6lMOldnkYQCCuNWeDibYQ=="
     },
     "@tlon/indigo-react": {
-      "version": "1.2.13",
-      "resolved": "https://registry.npmjs.org/@tlon/indigo-react/-/indigo-react-1.2.13.tgz",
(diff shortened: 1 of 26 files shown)


matildepark (changes requested):
Clicking into "Message..." on mobile pushes the header up.

image

ColorInput now causes a horizontal overflow.

image

matildepark:
bump

liam-fitzgerald:
closing in favour of #4045

Issue #3925, opened 2020-11-12
hark: DM invites don't appear in inbox, still appear in DMs+Drafts


vvisigoth:
**Describe the bug**
If I am invited to a new DM, I don't receive a notification.

**To Reproduce**
1. Start a new DM
2. Invite shows up in DMs+Drafts
3. No invite in Inbox

**Expected behavior**
DM invites should appear in inbox

**Base**
`tdo1o`

matildepark:
I actually initially wrote it so that *all* invites of all types would go to notifications and then I decided to only do groups. Why? Uh, I don't know. So I can just restore that.

PR #3940, opened 2020-11-16
landscape: show all invites in inbox

matildepark:
- Moves `SidebarInvite` to components and calls it `InviteItem`
- Removes invite logic from the sidebar, moves all invites for all apps to the inbox instead
- Changes the check for 'a dot' accordingly

See #3925.

Screenshot 2020-11-16 at 4 16 57 PM

It would be cool if we could display if it was a chat or the like — but invite-store provides us with `graph-push-hook`, not any metadata, ahead of time.


vvisigoth:
yuge
—
~poldec-tonteg
http://urbit.org

liam-fitzgerald (approved):

PR #3942, opened 2020-11-17
vere: add flags for import and export

ixv:
Part of the [data migration](https://github.com/urbit/urbit/pull/3604) work. This PR adds -i and -o flags for import and export.

I moved the commit for the herb/lens changes from the linked PR to this one, because its a dependency for this work, and it looks like this work will land first. Even so, just using this naively in its current state will not work because the lens changes depend on changes to the apps in that PR. To test this, just change the `+on-peek` arm in lens to return a constant for `%export-all`.

The import flow, while gross, I believe does not break anything. Making an http request with no place to put the response appears to just cause the http server to reset.

The way to invoke this is 
to export: `urbit -o <pier>` (or to specify the output path: `urbit -o -Y ./path/to/archive.jam <pier>`)
to import: `urbit -i /path/to/archive.jam <pier>`

### DIFF
--- a/pkg/urbit/vere/io/fore.c
+++ b/pkg/urbit/vere/io/fore.c
@@ -69,6 +69,32 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
   u3z(ovo);
 }
 
+/* _fore_import(): form an ovum from jammed archive at [pax_c] and inject it.
+*/
+static void
+_fore_import(u3_auto* car_u, c3_c* pax_c)
+{
+  // With apologies
+  u3_noun arc = u3ke_cue(u3m_file(pax_c));
+  c3_c * b64_c = u3r_string(u3do("crip", u3do("en-base64:mimes:html", arc)));
+  c3_w siz = strlen(b64_c) + 120;
+
+  c3_c bod_c[siz];
+  snprintf(bod_c, siz,
+    "{\"source\": {\"import-all\": {\"base64-jam\": \"%s\"}}, \
+    \"sink\": {\"stdout\": null}}", b64_c);
+  u3_noun dat = u3nt(u3_nul, u3i_word(strlen(bod_c)), u3i_string(bod_c));
+
+  u3_noun req = u3nt(c3n,
+    u3nc(u3i_string("ipv4"), u3i_word(0x7f000001)),
+    u3nq(u3i_string("POST"), u3i_string("/"), u3_nul, dat));
+  u3_noun wir = u3nc(u3i_string("http-server"), u3_nul);
+  u3_noun cad = u3nc(u3i_string("request-local"), req);
+  u3_auto_peer(
+    u3_auto_plan(car_u, u3_ovum_init(0, c3__e, wir, cad)),
+    0, 0, _fore_inject_bail);
+}
+
 /* _fore_io_talk():
 */
 static void
@@ -105,6 +131,10 @@ _fore_io_talk(u3_auto* car_u)
   //
   if ( u3_Host.ops_u.jin_c ) {
     _fore_inject(car_u, u3_Host.ops_u.jin_c);
+  }
+
+  if ( u3_Host.ops_u.imp_c ) {
+    _fore_import(car_u, u3_Host.ops_u.imp_c);
   }
 }
 
--- a/pkg/urbit/vere/io/fore.c
+++ b/pkg/urbit/vere/io/fore.c
@@ -22,6 +22,17 @@ _fore_inject_bail(u3_ovum* egg_u, u3_noun lud)
   u3_ovum_free(egg_u);
 }
 
+/* _fore_import_bail(): handle failure on arbitrary injection.
+*/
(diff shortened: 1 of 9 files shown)


pkg/urbit/vere/io/fore.c
```diff
@@ -69,6 +69,32 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
   u3z(ovo);
 }
 
+/* _fore_import(): form an ovum from jammed archive at [pax_c] and inject it.
+*/
+static void
+_fore_import(u3_auto* car_u, c3_c* pax_c)
+{
+  // With apologies
+  u3_noun arc = u3ke_cue(u3m_file(pax_c));
```
joemfb, line 78: Both `u3m_file()` and `u3ke_cue()` can bail. It'd be nice to catch those and print a clean error message, but it's probably not urgent.

pkg/urbit/vere/io/fore.c
```diff
@@ -69,6 +69,32 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
...
+  c3_c * b64_c = u3r_string(u3do("crip", u3do("en-base64:mimes:html", arc)));
```
joemfb, line 79: This transfers `arc` and the product of `+en-base64`, but leaks the product of `+crip` (`u3r_string()` has retain semantics). You'll need to assign that to a variable and `u3z()` it once you're done.

pkg/urbit/vere/io/fore.c
```diff
@@ -69,6 +69,32 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
...
+  c3_w siz = strlen(b64_c) + 120;
+
+  c3_c bod_c[siz];
```
joemfb, line 82: This can overflow the stack on large inputs, it'd be better to `c3_malloc()`.

pkg/urbit/vere/io/fore.c
joemfb: You'll also need to call `c3_free()` on `b64_c`

pkg/urbit/vere/io/fore.c
```diff
@@ -69,6 +69,32 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
...
+  u3_noun dat = u3nt(u3_nul, u3i_word(strlen(bod_c)), u3i_string(bod_c));
+
+  u3_noun req = u3nt(c3n,
+    u3nc(u3i_string("ipv4"), u3i_word(0x7f000001)),
+    u3nq(u3i_string("POST"), u3i_string("/"), u3_nul, dat));
+  u3_noun wir = u3nc(u3i_string("http-server"), u3_nul);
+  u3_noun cad = u3nc(u3i_string("request-local"), req);
+  u3_auto_peer(
+    u3_auto_plan(car_u, u3_ovum_init(0, c3__e, wir, cad)),
+    0, 0, _fore_inject_bail);
```
joemfb, line 95: `_fore_inject_bail()` is the error callback here -- you'll probably want to define your own so that you can at least print a clear message if the import fails. If that should be a fatal error, you can call `u3_pier_bail(car_u->pir_u);`.

pkg/urbit/vere/io/fore.c
```diff
@@ -69,6 +69,32 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
...
 
+/* _fore_import(): form an ovum from jammed archive at [pax_c] and inject it.
+*/
+static void
+_fore_import(u3_auto* car_u, c3_c* pax_c)
+{
+  // With apologies
+  u3_noun arc = u3ke_cue(u3m_file(pax_c));
+  c3_c * b64_c = u3r_string(u3do("crip", u3do("en-base64:mimes:html", arc)));
+  c3_w siz = strlen(b64_c) + 120;
```
joemfb, line 80: This should be named `siz_w`.

joemfb (changes requested):
A few things with `_fore_import()`:

joemfb:
When it gets a response effect without a corresponding requests, http.c should just print `http: strange response`. Can you confirm?

ixv:
I get 
image

joemfb:
Ah, the wire parsing is failing, so `_http_io_kick()` returns `c3n` and we never get to `_http_ef_http_server()`. Also, I think that's just the server starting for the first time, not a reset. Had the ports already printed once?

In any case, this behavior is probably fine for now.

ixv:
Ah you're right, its starting for the first time, not sure why I thought otherwise.

joemfb (approved):
The memory-management changes LGTM.

Worth noting: I'm not sure how suitable `+en-base64:mimes:html` is for arbitrary binary data. I had some problems with it when I was writing `:acme`, which lead me to write `lib/base64`. I've been meaning to replace the one in %zuse with the lib. In any case, this isn't an urgent consideration, as it affects only the import flow.

PR #3947, opened 2020-11-17
Revert "chat: fixed embedded scroll item disability"

matildepark:
This reverts commit 3b5d3c6ff6a288cbf0d659cdb84183baa3bd8d6e.

@tylershuster This doesn't seem to actually fix the issue with reversed scrolling in embeds and introduces additional bugs for users when scrolling down, any objections here?

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/content/text.js
+++ b/pkg/interface/src/views/apps/chat/components/content/text.js
@@ -26,7 +26,7 @@ const DISABLED_INLINE_TOKENS = [
 
 const renderers = {
   inlineCode: ({language, value}) => {
-    return <Text mono backgroundColor='washedGray' style={{ whiteSpace: 'preWrap'}}>{value}</Text>
+    return <Text mono p='1' backgroundColor='washedGray' style={{ whiteSpace: 'preWrap'}}>{value}</Text>
   },
   code: ({language, value}) => {
     return <Text
--- a/pkg/interface/src/views/apps/chat/components/content/text.js
+++ b/pkg/interface/src/views/apps/chat/components/content/text.js
@@ -36,7 +36,7 @@ const renderers = {
               borderRadius='1'
               mono
               backgroundColor='washedGray'
-              overflowX='scroll'
+              overflowX='auto'
               style={{ whiteSpace: 'pre'}}>
               {value}
             </Text>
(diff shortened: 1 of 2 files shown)


tylershuster:
Well, the description said "embeds" because that was the original issue, but this also fixes the problem for "embedded" hoon execution content. To see what I mean, run some code that outputs multiline results, and then try scrolling across it

tylershuster:
And if you don't see what I mean, then I don't know what I'm talking about and you can delete this

liam-fitzgerald (approved):

matildepark:
Code blocks no longer have vertical overflow scrolling, it seems. Idk. @urcades @g-a-v-i-n is there a preference here

image

urcades:
> Code blocks no longer have vertical overflow scrolling, it seems.

This is wonderful imo, let vertically-large code blocks be free!

With respect to the actual block examples you posted above, I think following the current indigo spec is a good place to start:

Screen Shot 2020-11-18 at 11 24 27 AM

Note the equally-spaced margins around the text. It appears as though our current block code elements have uneven margins.

To note, Figma does grant a specific code-oriented view into how these elements are composed:

Screen Shot 2020-11-18 at 11 28 01 AM

Here, we see the text is 12pt, while the line height is 16pt — i'm unsure how this will change now that we're using 14pt text in chat (and in comments), but as our spec changes, this is a place to watch.

tylershuster:
Given the above, I'm happy to revert this commit

matildepark:
Note that the padding is indeed even:
Screen Shot 2020-11-18 at 3 30 38 PM

If you make code blocks `display: inline` then it seems to fix it but ... of course, that's not ideal.
image

matildepark:
Oh, I figured it out. Invisible scrollbar making the block big.

urcades:
Sorry, it wasn't apparent in the images I posted as they were multi-line, but there's the overall equality of padding/margins we'd like in larger blocks, but we also want to maintain this for single lines as well — this seems less of a literal padding fix, as you illustrated, and more of an x-height fix that needs to be considered:

Screen Shot 2020-11-18 at 3 48 58 PM

Re-using the screenshot you took, it's not the padding that's the issue here, it's that the x-height of the letters themselves are visually off-center.

An article that speaks to this:

https://stackoverflow.com/questions/25836191/vertical-alignment-based-on-x-height

matildepark:
It wasn't the Xheight, it was a forced scrollbar ("scroll" instead of "auto").

image

All good now.

urcades:
Oh yeah, this looks perf

Issue #3875, opened 2020-11-03
groups: no UX or indication that "sharing your contact" must be done first


matildepark:
- Join a group.
- Try to edit your information all at once.

In the case of sigil color, it'll autofill the field with a JS error. In the case of other fields, it'll seem to work fine until you Save, and it will spin forever.

You need to hit save with *no info typed in* first, then you can edit the avatar info.

But we should either do that invisibly for the user, or (perhaps preferably) surface to the user to opt-in to set a group profile before allowing content to be entered (you can imagine a button to "Set Profile" sending the first 'share' action and enabling the fields).

(Aside, we want the fields to be prefilled with our root identity info.)

PR #3951, opened 2020-11-18
landscape: fix broken sharing of contact information, prefill with root identity

matildepark:
I was wrong about the issue in #3875 — the issue was the rewrite of the `hexToUx` utility into TypeScript, which instead of returning a string dumped a JS blob into the field. Why it worked properly when the contact information existed, I don't know. So now you can set all your information before sharing your contact information, without having to share the empty contact first, as expected.

Also fixes an unfiled issue where the color input processing assumed all information began with `0x`, as it does in group colour metadata but not in contact card defaults (which just used `0`), so it would show a different colour than what you submitted because it would chop off the first two characters.

Instead of falling back to an empty contact card, it falls back to our root identity and says "Share Contact". The root identity page now explains what it does.

Screenshot 2020-11-17 at 11 28 03 PM
Screenshot 2020-11-17 at 11 36 39 PM

(Screenshots from two different dev environments, but you get the idea.)

### DIFF
--- a/pkg/interface/src/logic/lib/util.ts
+++ b/pkg/interface/src/logic/lib/util.ts
@@ -155,18 +155,14 @@ export function uxToHex(ux: string) {
   return value;
 }
 
-export const hexToUx: (hex: string) => string = f.flow(
-  f.split(""),
-  f.chunk(4),
-  f.map(
-    f.flow(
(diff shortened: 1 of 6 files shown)


liam-fitzgerald (approved):
LGTM!

urcades:
Stamp!

PR #3955, opened 2020-11-18
Custom group policy

xiphiness:
This PR adds a new group policy [%custom agent=term join-path=path]

There are some things that still need working out, namely around perhaps having two different marks and $%'s for actions and updates, or simply having two cases for %change-policy. Reason being, is that updates get turned into pokes by the pull-hook, and I have the title and description of the policy scry'd for in the action case, but are included in the update in the update case. I haven't made this uniform yet so it's not quite functional at the moment. I would like to get input on this.

This is related to antechamber, a group policy and gall agent that only allows galaxies and ships with a reciprocal alt-id claim from the ethereum claims contract with a galaxy, made for creating a group exclusive to senates and senate delegates.

Also, UI support is yet to be done.

I'm also thinking maybe other, standardized paths to scry for admin, moderator permissions, but maybe that's out of scope for now.


matildepark:
I was going to say "doesn't banRanks let you do this" but

>  ships with a reciprocal alt-id claim with a galaxy

Tagging @liam-fitzgerald since he would catch appropriate alternatives, reviews relating to group-store...

pkg/arvo/sur/group.hoon
```diff
@@ -51,13 +51,25 @@
...
+        [%open diff:open]
+        [%replace =create]
     ==
   ::  $diff: change group policy
   +$  diff
     $%  [%invite diff:invite]
         [%open diff:open]
         [%replace =policy]
     ==
+
```
liam-fitzgerald, line 72: empty newline

pkg/arvo/sur/group.hoon
```diff
@@ -89,5 +101,11 @@
           [%allow-ships ships=(set ship)]
       ==
     --
+  ++  custom
+    =<  custom-policy
+    |%
+      +$  custom-policy  [%custom agent=term join-path=path name=@t description=@t]
```
liam-fitzgerald, line 107: indentation on arms

pkg/arvo/sur/group-store.hoon
```diff
@@ -32,13 +32,18 @@
...
 ::    %expose: unset .hidden flag
 ::
-+$  action
-  $%  [%add-group =resource =policy hidden=?]
++$  action  (build-group-diff create:policy action:policy)
+
++$  diff  (build-group-diff policy diff:policy)
+
+++  build-group-diff
+  |$  [create diff]
```
liam-fitzgerald, line 40: Is there any reason you can't add a `metadata-path` to the policy, and then scry/watch for that to populate info? I don't particularly see anything wrong with even just leaving the title and description in, so we don't have this weird action/diff distinction.

pkg/arvo/mar/group/update.hoon
```diff
@@ -8,9 +8,4 @@
   %+  frond:enjs:format  'groupUpdate'
   (update:enjs upd)
   --
-++  grab
-  |%
-  ++  noun  update
```
liam-fitzgerald, line 13: ?

liam-fitzgerald (commented):
I find the action/update distinction kinda weird, but other than that it looks good.

xiphiness:
> I find the action/update distinction kinda weird, but other than that it looks good.

I'm thinking instead, to have a single $% that has a %change-policy and %policy-changed. the former will never be emitted as a diff, but be transformed into the latter, and the latter will never be an action

pkg/arvo/sur/group-store.hoon
xiphiness: see below comment

pkg/arvo/mar/group/update.hoon
xiphiness: artifact of splitting up update and action - with new idea, see below, I'd reinstate this

liam-fitzgerald:
Yeah that sounds ideal!

zalberico:
Closing as stale, can reevaluate if necessary.

PR #3956, opened 2020-11-18
eth abi parsing, sur, lib, and mar codegen

xiphiness:
This PR adds an eth ABI parser, and code generation of static types, marks, and encoding and decoding libraries for ethereum function calls and writes, event subscription and event updates.

lib/eth-abi does the parsing, lib/eth-abi-gen does the code generation, lib/eth-abi-magic has some extraneous methods and a modified version of abi:ethereum to do various noun manipulation and decoding that was not possible with existing zuse, which ends up being used by the generated libraries. The abi:ethereum code should probably be moved into zuse, but not the noun manipulation.

I have not exhaustively tested every ABI type yet, though have tested most. One thing in particular is I would like to get another pass on a decode-array-n that doesn't lose type information. I have one that works, but don't particularly like it. A similar thing ought to be done with the tuple case(which isn't present at all in zuse, but may be able to be reduced to an array-n case).

This probably deserves thorough unit tests in its own right, as we have for the zuse en/decoders.

### DIFF
--- a/pkg/arvo/lib/eth-abi-gen.hoon
+++ b/pkg/arvo/lib/eth-abi-gen.hoon
@@ -1,0 +1,635 @@
+/+  *eth-abi-parse
+|%
+::  TODO: for some reason u.name.petyp is null sometimes
+::  see petyp-to-bape hack
+  ++  damn-skim
+    |=  [es=(list event-input) b=?]
+    ^-  (list event-input)
+    %+  skim  es
+    |=  =event-input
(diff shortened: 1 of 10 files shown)


Fang-:
This is neat! All your comments sound right. I'll wait with more thorough review until you're done with the improvements here. For the time being: arms in cores don't get indented, only their contents. (;

zalberico:
Closing this one out as stale, we can reopen if work starts up on it.

PR #3962, opened 2020-11-19
gall: Allow deleting or suspending an app

philipcmonk:
NB: this has breaking changes without state adapters!  It also changes Gall's wire format with no adapters.  Hopefully this can go into the breach, else I can write those later.

Uninstalling or suspending apps is useful when trying out unstable builds or 3rd party apps.  There's a few ways we can implement this:

- `%gone`:
  - unsubscribe from all incoming and outgoing subscriptions
  - delete state
  - risks
    - outstanding pokes don't get acknowledged
    - outstanding kernel requests don't get canceled

- `%doze`:
  - don't unsubscribe
  - resolve to state (i.e. call +on-save)
  - queue deals and agent signs to be run once agent resumes
  - makes app no longer block updates
  - allows resuming later
  - risks
    - app becomes nonresponsive without notifying subscription clients
    - might miss responses from kernel

- %sleep
  - unsubscribe from all incoming and outgoing subscriptions
  - resolve to state (call +on-save)
  - queue deals and agent signs (or drop/nack?)
  - can wake up later
  - risks
    - weird non-functional state
    - if don't queue kernel signs, might miss important responses

In this PR I implement the first two, `%gone` and `%doze`, which I think covers most use cases.

This feature should be considered experimental -- I've tested it reasonably well, but deleting an app breaks an invariant that we've always had, so there could be far-away effects.  But we need to get this right before 3rd party software distribution lands, so we should start experimenting.

Interface is:

```
|fade %app         ::  delete app altogether
|doze %app         ::  suspend app
|doze %app %wake   ::  resume app
```

Also includes `|pass`, which takes an arbitrary `note-arvo` and sends it.  This makes the implementation of simple generators like `doze` easy since they don't require hood changes.

For a more expansive description of the tradeoffs involved in uninstalling gall apps, check out the first 20 minute of my recent stream: https://www.youtube.com/watch?v=PJ7DqlzRrqo

@belisarius222 please review both the design and code, @joemfb you can just review the design if you want.

### DIFF
--- a/pkg/arvo/gen/hood/fade.hoon
+++ b/pkg/arvo/gen/hood/fade.hoon
@@ -1,6 +1,6 @@
 ::  Drum: destroy app
 ::
-::::  /hoon/start/hood/gen
+::::  /hoon/fade/hood/gen
   ::
 /?    310
 ::
--- a/pkg/arvo/gen/hood/fade.hoon
+++ b/pkg/arvo/gen/hood/fade.hoon
@@ -1,0 +1,17 @@
+::  Drum: destroy app
+::
+::::  /hoon/start/hood/gen
+  ::
+/?    310
+::
+::::
+  ::
+:-  %say
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        [arg=[@ $@(~ [@ ~])] ~]
+    ==
+:-  %drum-fade
+?>  ((sane %tas) -.arg)
+?@  +.arg  [q.bec -.arg]
+?>  ((sane %tas) +<.arg)
+[-.arg +<.arg]
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -56,13 +56,13 @@
       live=?
       =stats
       =watches
-      =agent
+      agent=(each agent vase)
       =beak
       marks=(map duct mark)
   ==
 ::  $blocked-move: enqueued move to an agent
 ::
-+$  blocked-move  [=duct =routes =deal]
++$  blocked-move  [=duct =routes move=(each deal sign:agent)]
 ::  $stats: statistics
 ::
 ::    change: how many moves this agent has processed
@@ -126,7 +126,7 @@
       live=?
       =stats
       =watches
-      old-state=vase
+      old-state=(each vase vase)
       =beak
       marks=(map duct mark)
   ==
@@ -166,9 +166,11 @@
         ?~  apps  mo-core
         ~>  %slog.[0 leaf+"gall: upgrading {<dap.i.apps>}"]
         =/  ap-core  (ap-abut:ap:mo-core i.apps)
-        =^  tan  ap-core  (ap-install:ap-core `old-state.egg.i.apps)
-        ?^  tan
-          (mean u.tan)
+        =?  ap-core  ?=(%& -.old-state.egg.i.apps)
+          =^  tan  ap-core  (ap-install:ap-core `p.old-state.egg.i.apps)
+          ?^  tan
+            (mean u.tan)
+          ap-core
         =?  ap-core  wipe-eyre-subs.spore
           =/  ducts=(list ^duct)
             %+  skim  ~(tap in ~(key by inbound.watches.egg.i.apps))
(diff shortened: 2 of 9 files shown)


Fang-:
> Also includes `|pass`, which takes an arbitrary `note-arvo` and sends it. This makes the implementation of simple generators like `doze` easy since they don't require hood changes.

Strong contender for QoL Change Of The Year.

belisarius222:
> queue deals and agent signs to be run once agent resumes
I'll note that this is a potential space leak.  Not a showstopper, though, since this functionality is definitely useful.

pkg/arvo/gen/hood/fade.hoon
```diff
@@ -0,0 +1,17 @@
...
+/?    310
+::
+::::
+  ::
+:-  %say
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        [arg=[@ $@(~ [@ ~])] ~]
+    ==
+:-  %drum-fade
+?>  ((sane %tas) -.arg)
```
belisarius222, line 14: Why isn't `arg` just a `@tas` to begin with?

belisarius222: each piece of `arg`, rather

pkg/arvo/sys/zuse.hoon
```diff
@@ -1723,6 +1723,7 @@
           [$deal p=sock q=term r=deal]                  ::  full transmission
           [%goad force=? agent=(unit dude)]             ::  rebuild agent(s)
           [%sear =ship]                                 ::  clear pending queues
+          [%fade dap=term style=?(%gone %doze %wake)]   ::  put app to sleep
```
belisarius222, line 1726: I'd use something other than %wake here, since Behn already uses that.  %jolt, maybe?

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -53,16 +53,17 @@
 ::
 +$  yoke
   $:  control-duct=duct
+      nonce=@t
```
belisarius222, line 56: Are we committing to using nonces for app scrys now?  I wasn't sure we'd resolved that discussion.  I'm slightly concerned that running a hash function on every app invocation will be slow, but maybe that's not worth worrying about.

I'd be inclined to push this in one of two directions.  Either:
+ defer all nonce and scry work until later, or
+ go all the way and redo `$bowl` now to include the nonce, so that apps don't need to do an upgrade later.

belisarius222 (changes requested):
I think we shouldn't reuse %wake and %doze, since those show up for Behn in so many places.  I have some other questions and concerns, but probably nothing else pressing enough for me to want to hold up this work.

pkg/arvo/gen/hood/fade.hoon
philipcmonk: This is lifted exactly from `|start`.  Maybe they both should be updated to modern style.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -53,16 +53,17 @@
...
       live=?
       =stats
       =watches
-      =agent
+      agent=(each agent vase)
```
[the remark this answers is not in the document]
philipcmonk: Not sure what you mean.  The `each` is between an active app and one that's been suspended to its state for an indefinite period of time.  When gall goes into its pupal phase, this becomes and `(each vase vase)` to preserve the distinction so it doesn't try to rebuild a suspended app.

pkg/arvo/sys/vane/gall.hoon
philipcmonk: This is a different nonce.  The hash is run only once when starting an app and put in the duct used for all `%pass` moves from that app.  An example of the bug we're avoiding is:

- Start app `%foo`, which starts a timer for ~d1
- Delete app `%foo`
- Less than ~d1 later, start another copy of `%foo`
- `%foo` gets a timer response that it didn't expect because the timer was started by an earlier iteration of the app.

Several other kernel interfaces could cause this, like Clay subscriptions.  Until we make those legible to Gall, we can't cancel them automatically.

pkg/arvo/sys/zuse.hoon
philipcmonk: You're right these are too close to the other meanings of `%doze` and `%wait`.  I've changed them as you suggest, except I've left the user-visible names, since I think for that the original names make more sense.

joemfb (approved):
LGTM

PR #3963, opened 2020-11-19
hotfix: OTA bugs

liam-fitzgerald:
- Scroll on load
- Hide empty timeboxes
- Loosen property access in various places to fix crashes


matildepark (approved):
looks like option chaining to me

PR #3964, opened 2020-11-19
gen: add +sponsor generator

ronreg-ribdev:
Add a new generator `+sponsor` that encapsulates the logic of invoking `(sein:title our now our)` to get the sponsor of the current ship.

### DIFF
--- a/pkg/arvo/gen/sponsor.hoon
+++ b/pkg/arvo/gen/sponsor.hoon
@@ -1,0 +1,5 @@
+::  Print the sponsor of this ship
+:-  %say
+|=  [[now=time @ our=ship ^] * ~]
+:-  %ship
+(sein:title our now our)


pkg/arvo/gen/sponsor.hoon
```diff
@@ -0,0 +1,6 @@
+::  Print the sponsor of this ship
+:-  %say
+
```
tacryt-socryp, line 3: Blank lines are against the style guide. Otherwise, looks great!

tacryt-socryp (approved):

pkg/arvo/gen/sponsor.hoon
ronreg-ribdev: Fixed this!

philipcmonk:
One more thing: commit messages should have the form `component: message`.  There's no fixed taxonomy of component names, it's usually an app or vane name, but in this case I'd use `gen:`

PR #3972, opened 2020-11-19
webterm: migrate to indigo-react

matildepark:
I thought this was already done in the Term work, but it wasn't, so I just do it here. Tested on desktop and on iOS, works normally. Matches code styling in chat.

cc @urcades

### DIFF
--- a/pkg/interface/src/views/apps/term/app.js
+++ b/pkg/interface/src/views/apps/term/app.js
@@ -5,6 +5,8 @@ import Helmet from 'react-helmet';
 import { History } from './components/history';
 import { Input } from './components/input';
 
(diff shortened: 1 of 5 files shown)


urcades:
Stamp from me

Fang- (changes requested):
Neat, thanks for cleaning up my mess. That screenshot doesn't look quite right though, the output from `azimuth` should look like so:

image

Guessing something got lost in the classname->prop translation, but it's not immediately obvious to me what.

matildepark:
Cracked the dev environment open again and looks fine?

EBB399BB-D980-48E8-AB82-F57B5486A64B

Fang- (approved):
Ok, looks all good then! Thank you.

PR #3973, opened 2020-11-19
u3: adds ghetto +mook replacement

joemfb:
`+mook` processes stack frames into printable trace (`$tang`), but is currently unavailable to use during boot, which makes bootstrap debugging very painful. This adds very basic printing of stack frame values for `%spot` and  %mean (atomic only).


philipcmonk (approved):
Very useful, even for just mean and spot

PR #3974, opened 2020-11-19
u3: adds |ut battery to jet cache keys

joemfb:
This resolves a basic safety issue with the partial memoization in the decapitated jets.


joemfb:
This PR is against `release/next-vere`, since it's a compatible change that could go out at any time. But it's not urgent, and would be fine to go out with `na-release/next-vere`.

philipcmonk (approved):
Good check

PR #3978, opened 2020-11-20
hark: remove 'archived' view, use 'dismiss'

matildepark:
- Removes 'Archive' view from the inbox.
- Removed 'archivedNotifications' from all store, reducer logic
- Says 'Dismiss' instead of 'Archive'

image

cc @urcades 

Is there anything we used `archivedNotifications` for? Is this a thorough and concise surgery?

### DIFF
--- a/pkg/interface/src/logic/reducers/hark-update.ts
+++ b/pkg/interface/src/logic/reducers/hark-update.ts
@@ -10,11 +10,10 @@ import { StoreState } from "../store/type";
 
 type HarkState = Pick<StoreState,
   "notificationsChatConfig"
-  | "notificationsGroupConfig" 
-  | "notificationsGraphConfig" 
+  | "notificationsGroupConfig"
+  | "notificationsGraphConfig"
   | "notifications"
   | "notificationsCount"
-  | "archivedNotifications"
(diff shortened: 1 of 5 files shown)


liam-fitzgerald (approved):
Looks good, although if we are excising this from the frontend, I see no reason we should keep storing the archived notifications.

matildepark:
> Looks good, although if we are excising this from the frontend, I see no reason we should keep storing the archived notifications.

Yeah, just consider this pending @urcades stamp because this is just me following intuition and some feedback from people. Instigating a dialogue~

tacryt-socryp:
Personally I think it best not to store dismissed notifications, but that's just my 0.02

vvisigoth:
IMO storing archived notifications only makes sense if you can search them
(from my perspective using gmail inbox).
—
~poldec-tonteg
http://urbit.org

matildepark:
@urcades informed me that this is a topic of disagreement for design at the moment so we'll have to wait and see.

urcades:
Update: It looks like design (Gav and I, as Jimmy is Out) is aligned on this, and agrees this simplification is desirable for a variety of reasons.

Issue #979, opened 2018-05-14
First _pave_north() call writes outside loom bounds


eglaysher:
I tried compiling Urbit on my new POWER9 machine, running Linux.

The first obvious bug was a crash after mprotect() because of unaligned pages. However, after changing 
`u3a_page` to 16 for the 64kb pages on POWER9 chips, I'm still getting a crash on startup:

```
#0  __memset_power8 () at ../sysdeps/powerpc/powerpc64/power8/memset.S:254
#1  0x00000001000541e8 in _pave_north (mem_w=0x36000004, siz_w=356,
    len_w=536870911) at ../noun/manage.c:421
#2  0x0000000100054630 in u3m_pave (nuu_o=0 '\000', bug_o=1 '\001')
    at ../noun/manage.c:488
#3  0x00000001000580a8 in u3m_boot (nuu_o=0 '\000', bug_o=1 '\001',
    dir_c=0x100143d80 "fakezod", pil_c=0x0,
    url_c=0x1000c64b0 "https://bootstrap.urbit.org/latest.pill",
    arv_c=0x100143e40 "arvo/") at ../noun/manage.c:1665
#4  0x0000000100025348 in main (argc=7, argv=0x7ffffffff348)
    at ../vere/main.c:600
```

This crash in `_pave_north()` appears to be happening because the memset is scribbling outside the loom bounds. This appears to be happening on x86 too, but just happens to work. If you were to add the following checks (on either x86 or ppc64le):

```
/* after the mmap() in um3_init() */
fprintf(stderr, "map_v: %p - %p\n", map_v, map_v + len_w);
```

```
/* inside of _pave_north() before the memset */
fprintf(stderr, "pave-north: %p - %p \n", rod_u, rod_u + (4 * siz_w));
```

You'd get the following output:

```
map_v: 0x36000000 - 0xb6000000
loom: mapped 2048MB
pave-north: 0xb5fffa70 - 0xb61d8770 
```

Note how the range that is going to get memset in `_pave_north()` is partially outside the `map_v` range. I'm not quite sure what the right fix for this is though. The code seems fairly fiddly.

joemfb:
This has not been investigated. The issue might be obsolete, or it might only affect architectures on which we don't currently run (ie, POWER9). It's arguably low priority, but it's certainly not completed.

zalberico:
Meant to close as unplanned/won't fix (not completed) - it looked like an issue with POWER9, an architecture we don't support.

[...]

PR #3981, opened 2020-11-20
vere: adds versions to all pier artifacts

joemfb:
... including the checkpoint, checkpoint patch files, the event log, and jammed snapshots ("rock").

Additionally, this PR updates the format of the event log, both metadata and events. The metadata had an unnecessary layer of jamming, while the events had an unnecessary layer of cell allocation and a misalignment of events and the mug of the resulting kernel (both artifacts of the old IPC message format).

In cleaning up the persistent `u3_arvo` struct, I had to decide what to do with the i/o drivers' use of the "instance number" therein, as some use the instance number to create unique wires per process. While I suspect that this is entirely unnecessary, I'm not quite sure, and it's an easy thing to change in the future. I've implemented equivalent behavior with per-driver values rather than per-process.

This PR is a draft because the checkpoint version number is not yet useful -- it's the last member of the `u3_home` struct, which makes it quite useless for format discrimination. There's pointer math in the loom initialization, and I'm too tired to track down where it's gone wrong (but I suspect #979 maybe related); I'll revisit tomorrow.

@eglaysher, the event log format changes will need to be implemented in king-haskell as well.


eglaysher:
A question about the changes to lmdb here: is this just adding a version key/value pair to the meta table? I want to make sure I'm not missing something.

joemfb:
No, the key change is to the structure of the persisted events themselves.

Before, log entries were jammed cells, where the head was the mug of the kernel from the "previous" event, and the tail was the "current event".

Now, log entries are byte buffers, where the first 4 bytes are the mug of the kernel from the "current" event, and the remaining bytes are the jammed current event.

Fang-:
Changes so far look fine. I'm not very familiar with this part of vere, but nothing jumps out as weird, and the cleanup is good. Withholding approval until no longer WIP!

eglaysher:
I have a draft of the new format in KH (https://github.com/urbit/urbit/tree/eg/hs-vere-versions), but it looks like I'm having issues with the urbit-worker being unstable. Is that what you were talking about with regards to loom math?

joemfb:
No, this should be working in its current state -- I didn't push the failing changes. Do you have any more details about what you're seeing?

joemfb:
I have no idea why github is showing so many commits; here's the real diff: https://github.com/urbit/urbit/compare/b6edb8a4c...2edfc51?expand=1

matildepark:
Wouldn't merging this duplicate those commits? Wondering if a force push changed the tree somewhere in there.

joemfb:
I think it's just that the Ui state is broken: 
https://github.com/urbit/urbit/compare/na-release/candidate...jb/vere-versions?expand=1

joemfb:
"have you tried turning it off an on again?"

pkg/urbit/include/noun/vortex.h
```diff
@@ -5,22 +5,22 @@
   /**  Data structures.
   **/
     /* u3v_arvo: modern arvo structure.
+    **       NB: packed to perserve word alignment given [eve_d]
     */
-      typedef struct _u3v_arvo {
-        c3_d    ent_d;                    //  event number
+      typedef struct __attribute__((__packed__)) _u3v_arvo {
+        c3_d  eve_d;                      //  event number
```
philipcmonk, line 11: Press F for "bumping ent_d"

pkg/urbit/vere/io/ames.c
```diff
@@ -1318,10 +1319,14 @@ _ames_io_talk(u3_auto* car_u)
   u3_ames* sam_u = (u3_ames*)car_u;
   _ames_io_start(sam_u);
 
-  // send born event
+  //  send born event
   //
   {
-    u3_noun wir = u3nt(c3__newt, u3k(u3A->sen), u3_nul);
+    //  XX remove [sev_l]
```
philipcmonk, line 1325: Is this XX for now or later?

pkg/urbit/vere/io/http.c
```diff
@@ -1738,7 +1737,9 @@ _http_serv_start_all(u3_httd* htd_u)
 
     //  XX remove [sen]
     //
-    u3_noun wir = u3nt(u3i_string("http-server"), u3k(u3A->sen), u3_nul);
+    u3_noun wir = u3nt(u3i_string("http-server"),
+                       u3dc("scot", c3__uv, htd_u->sev_l),
```
philipcmonk, line 1741: Aren't these various `scot` calls rather expensive?

philipcmonk (approved):
Couple questions; no objections

pkg/urbit/vere/io/ames.c
joemfb: It's for later; the question is just: is it desirable for these wires/ducts to change with each process lifetime? I tend to think not, but I'm not sure. (While I'm thinking of it, I should really get rid of `%newt` ...)

pkg/urbit/vere/io/http.c
joemfb: Yeah, they are. Here and above, we should really call them once and store them in the driver/server/connection/request structs as strings. I'll note that for later.

PR #3985, opened 2020-11-20
sane: restructured to emit cards (for archiving and removing)

tacryt-socryp:
Not done yet, but this compiles, so it's enough for you to see where it's going. Working on automating the cleanup of hanging chads (metadata without a group, graphs without a group or metadata, group without metadata, etc).

Once done, this will fix the broken resources left over from #3857.

### DIFF
--- a/pkg/arvo/app/sane.hoon
+++ b/pkg/arvo/app/sane.hoon
@@ -28,246 +28,215 @@
 ::    
 ::
 /-  *metadata-store, contacts=contact-store, *group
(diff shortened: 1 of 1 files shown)


matildepark:
@liam-fitzgerald @tacryt-socryp If we're deploying this before breach this needs to be taken home for EOD tomorrow.

tacryt-socryp:
I still prefer the app restructuring here, but it's superceded by Liam's work last week.

PR #3989, opened 2020-11-20
publish: styles preview

tylershuster:
Fixes https://github.com/urbit/landscape/issues/182

I realize this changes some design (@jyng @urcades) but I think it's closer to the spec than before, and it basically just unifies it to use the same styling as on the inside, plus bolding the title...which I think is in spec? I mean it's all changing soon but this fixes it in the interim.

Before:
image

After:
image

### DIFF
--- a/pkg/interface/src/views/apps/publish/components/NotePreview.tsx
+++ b/pkg/interface/src/views/apps/publish/components/NotePreview.tsx
@@ -82,7 +82,7 @@ export function NotePreview(props: NotePreviewProps) {
           <Box color={isRead ? "gray" : "green"} mr={3}>
             {date}
           </Box>
-          <Box mr={3}><Text>{commentDesc}</Text></Box>
+          <Box mr={3}>{commentDesc}</Box>
           <Box>{rev.valueOf() === 1 ? `1 Revision` : `${rev} Revisions`}</Box>
         </Box>
       </Col>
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/apps/publish/components/NotePreview.tsx
```diff
@@ -62,8 +62,8 @@ export function NotePreview(props: NotePreviewProps) {
   return (
     <Link to={url}>
       <Col mb={4}>
-        <WrappedBox mb={1}>{title}</WrappedBox>
-        <WrappedBox mb={1}>
+        <WrappedBox mb={1}><Text bold>{title}</Text></WrappedBox>
+        <WrappedBox mb={1} className="md">
```
matildepark, line 66: I feel like we should defer to indigo-react's line-heights here instead of PRing more custom css classes?

tylershuster: I agree with you, but `md` is the class applied to the contents of the note once you open it.

matildepark:
boop @urcades

tylershuster:
While we're at it, should this be the same width as links?

urcades (approved):
@tylershuster ideally yes, the same width would be preferable

The updates here are looking good, I like

tylershuster:
image

tylershuster:
Compare to 

this

urcades:
Yeah, this is gorg, double stamp

PR #3991, opened 2020-11-21
arvo: tank reform

joemfb:
This PR:

- revises `$tank`, removing $plum from it and adding $cord.
- updates the `$tank` renderer (`|re`) to accommodate these changes
- additionally removes the `\/` windowing and utf32 conversions required by it (to avoid unsafely truncating a utf8 string)
- moves `$plum`, its renderer (`|plume`), and `$xray` back into userspace, as they're incomplete implementations, unused in the kernel
- moves the implicated styled text and output types back to %zuse and modernizing their syntax
- removes a handful of surrounding, unused structures (`$ares`, `$coop`, and `$disc`)

The `%leaf` branch of `$tank` is preserved as a convenience for string interpolation, which only works with tapes. The other tapes (in the `%rose` and `%plum` branches) should arguably be cords, but that's deferred (at least here) as a more disruptive change.


philipcmonk (approved):
Very nice!

PR #3995, opened 2020-11-21
landscape: fix link route breaking popover

ryjm:
When you're on a link page, hitting the settings popover will trigger an error since the LinkResource matches the path. This makes it so that it only matches numerical indexes.


matildepark:
King of the one-liners

matildepark (approved):
[TIL](https://stackoverflow.com/questions/37269736/react-router-how-to-constrain-params-in-route-matching). Works. Addresses #4001, too.

PR #3998, opened 2020-11-22
groups: better ship search validation

### LINKED ISSUE #3824: Groups: Invite modal input + submit bug
I'm relaying a bug noted by @belisarius222: 

In the "Invite to Group" popover interface, the following bug was noted:

- If you paste in a `@p` into the field, and press "Send", the modal disappears as though the invite is sent, but it does not send

It seems to be that one must actually select a filtered `@p` from the dropdown in order for a 'send' to actually register.

There are two UX issues here:

1. Valid `@p`s should be able to be sent regardless of whether they are input or selected from a list
2. An invalid `@p` should obviously error out the form (right now, anything can be sent) — we need to make it a cardinal rule that success states and fail states don't resolve in exactly the same way

tylershuster:
Fixes #3824

...or nearly does. There's a problem where when you've entered a ship, then you go to type in another one, it clears the current value (preventing you from selecting multiple). @liam-fitzgerald you know react hooks & formik better than I. Can you point out what I'm doing wrong in `ShipSearch.tsx`?

A hint: in `onChange`, `value` is set to a blank array when it fires for the first time after selecting a ship, although the component later on knows that there is something in `value`.

### DIFF
--- a/pkg/interface/src/views/components/ShipSearch.tsx
+++ b/pkg/interface/src/views/components/ShipSearch.tsx
@@ -1,5 +1,5 @@
-import React, { useMemo, useCallback } from "react";
-import { Box, Label, Icon, Text, Row, Col } from "@tlon/indigo-react";
+import React, { useMemo, useCallback, ChangeEvent, useState, SyntheticEvent, useEffect } from "react";
+import { Box, Label, Icon, Text, Row, Col, ErrorLabel } from "@tlon/indigo-react";
 import _ from "lodash";
 import ob from "urbit-ob";
 import { useField } from "formik";
@@ -10,6 +10,8 @@ import { Associations, Association } from "~/types/metadata-update";
 import { cite, deSig } from "~/logic/lib/util";
 import { Rolodex, Groups } from "~/types";
 import { HoverBox } from "./HoverBox";
+
+const INVALID_SHIP_ERR = "Invalid ship";
 
 interface InviteSearchProps {
   autoFocus?: boolean;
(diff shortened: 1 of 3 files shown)


liam-fitzgerald:
hope you don't mind me pushing over it. tl;dr is that the lack of separation between the textfield state and the selected state made things a little awkward. Refactored to store them separately and `useEffect` to set the actual field value any time either one changes.

liam-fitzgerald:
cc: @tylershuster can you review my changes?

tylershuster:
Sure...Where are they?

liam-fitzgerald:
oh lol i can't push to your repo
https://github.com/urbit/urbit/tree/invite-modal-submit

matildepark:
I also don’t see the pushed changes here. They exist in limbo …

— 
~haddef-sigwen
https://urbit.org

tylershuster:
Reference:

https://github.com/urbit/urbit/commit/3a3a1ac3d504a8c03ce5e609539492532965c312

liam-fitzgerald:
@tylershuster could you cherrypick the changes onto your branch?

tylershuster:
Yeah, this looks good. Like I said, I'm not too great with hooks yet so I'll study this as part of my education, but it does the job proper

matildepark:
This disabled state needs a tweak — 

image

I think if you pass `backgroundColor='white'` to AsyncButton -> Button you can force disabled state to always blend in here?

tylershuster:
How do we indicate it's disabled? If I pass bg="white" then it doesn't change when it becomes valid

matildepark:
All we care about is preventing the click. This popover doesn't use buttons like other UI prompts; it's really a button pretending to be an Action in this case. @urcades

urcades:
For reference, the latest "New DM" modal appears as such:

Screen Shot 2020-11-25 at 5 22 37 PM

@g-a-v-i-n should probably confirm here if Affirmative Buttons are universally aligned to the right side of the modal, or only right-aligned when another option is presented, ie [cancel] [send]

matildepark:
Every time lol

urcades:
Also: It's not necessary to pay mind to this entire design, but I've pasted this here because it is the latest art and depicts the state of the inactive indigo button.

tylershuster:
Ok, given the larger changes that need to be applied for the above design to work, how would we feel about shipping this for now?

image
image

urcades:
Yeah, this button state is good, stamp

matildepark:
Lightning round! How do we fix the error validation being rude when we're trying to choose an option? 

Nov-25-2020 17-53-26


Also can we make the text use `cursor: pointer` for the candidates too

tylershuster:
@liam-fitzgerald this is related to validateOnBlur...any ideas?

matildepark:
If the button disables anyway, why does it need the onBlur validation as well?

liam-fitzgerald:
just drop `setTouched(true)` in the `onBlur` handler, it'll behaviour fine because touched is set to true in the `onSelect` handler anyway.

tylershuster:
Ready for review

matildepark:
It's on the list for today ...

matildepark (approved):

matildepark:
Can you revise the commits?

tylershuster:
yes

pkg/interface/src/views/components/ShipSearch.tsx
```diff
@@ -45,25 +47,67 @@ const Candidate = ({ title, detail, selected, onClick }) => (
...
+        checkInput(true, ship);
+      } else {
+        checkInput(ship.length !== 1, undefined) 
+      }
+    },
+    [checkInput]
   );
 
+  const onBlur = useCallback(() => {
+    setInputTouched(true);
```
matildepark, line 86: is this still correct? i thought `setTouched` needs to be here?

matildepark: Oh it needed to be *dropped*, okay.

PR #4000, opened 2020-11-23
chat-cli: mvp graph-store migration

Fang-:
Makes chat-cli listen to graph-store rather than chat-store.

Along the way, cuts some features. No longer supports channel creation,
deletion, permission management or inviting. No longer lets you send
messages to multiple targets at once.

No longer displays messages from all joined chats, and no longer lets
you join or leave chats. Instead, chat-cli solely concerns itself with
viewing chat content you're already subscribed to (through use of
Landscape or other rich clients). `;view ~host/chat` to start printing
messages for that chat when they come in, `;flee` to toggle off. `;view` to
see what chats are enabled, `;chats` to see all available chats.

The removal of features acknowledges chat-cli's lack of real support for
the modern, group-centric state of userspace.
The intent of `;view` ux is to prepare chat-cli for a multi-session
terminal world.

~~Opening this PR in case it needs to go in asap, but will try and find some time in the next 24 hours to add actual multi-session support to this.~~ Multi-session support is in.

The approach here is one of "minimum viable migration". The code style is (or rather, remains) a bit rough around the edges, but fixing that is not in scope here.

Tested this a bit by playing around with graph generators, think it's probably good, but want to test again once @liam-fitzgerald's frontend changes are in just to make sure this plays nice in "wilder" contexts. Message rendering should work fine, but might look a tad janky.
(I assume an updated `/mar/graph-validator/chat.hoon`, with `notification-kind:grow` arm, will find its way into `next-userspace`.)

cc @syzygyzer for updating the messaging docs, though maybe we just want to remove chat-cli from the docs altogether for now.

### DIFF
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -369,6 +369,8 @@
       %remove-graph  (~(notice-remove se i.sez) +.q.upd)
     ::
         %add-nodes
+      ?.  (~(has in viewing.session.i.sez) resource.q.upd)
+        [~ session.i.sez]
       %+  ~(read-posts se i.sez)
         resource.q.upd
       (sort ~(tap by nodes.q.upd) ior)
--- a/pkg/arvo/app/chat-cli.hoon
+++ b/pkg/arvo/app/chat-cli.hoon
@@ -9,29 +9,48 @@
 ::    we concat the ship onto the head of the path,
 ::    and trust it to take care of the rest.
 ::
-/-  view=chat-view, hook=chat-hook,  *group,
-    *permission-store, *group-store, inv=invite-store,
-    sole
-/+  shoe, default-agent, verb, dbug, store=chat-store,
-    group-store, grpl=group, resource
+/-  *resource, post,
+    store=chat-store
+/+  shoe, default-agent, verb, dbug,
+    graph=graph-store, libgraph=graph
 ::
 |%
 +$  card  card:shoe
 ::
 +$  versioned-state
-  $%  state-2
+  $%  state-3
+      state-2
       state-1
       state-0
   ==
 ::
++$  state-3
+  $:  %3
+      ::TODO  support multiple sessions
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/chat-cli.hoon
```diff
@@ -710,177 +674,68 @@
...
     ::
     ::    this double-virtualizes and clams to disable .^ for security reasons
     ::
     ++  eval
       |=  [txt=cord exe=hoon]
-      (say %code txt (eval:store bowl exe))
+      ~&  %eval-tmp-disabled
+      [~ state]
+      ::TODO  why -find.eval??
+      :: (say %code txt (eval:store bowl exe))
```
Fang-, line 777: idk what's up with this, by the way. Nothing's shadowing `store`, nothing about this even changed, but it just won't resolve to the right thing. What dumb mistake could I be overlooking?

matildepark:
Thank you for your work @Fang-

matildepark:
Pointed at the current Chat -> graph-store migration location. @liam-fitzgerald can you review this?

liam-fitzgerald (approved):
This looks good! Pushed a couple of changes for things I found while testing, other than that LGTM!

ohAitch:
Is there at least still support for generators to manage channel state?
"Urbit is unusable without a web browser" feels deeply weird.

matildepark:
You haven't been able to have complete parity with Landscape's web front-end for quite a while now — since around OS1, I think. 

`:chat-cli` is from a point in time where we wrote IRC and left it at that; Landscape is a suite of Gall agents centred around the group pattern with an accompanying front-end, whose scope we've contained to the one web client as a team. Anything else is essentially a bonus until Landscape gets assigned a CLI maintainer. (Not to mention CLI parity is blocked on quite a few things to actually provide that experience. That said, @Fang- spends quite a bit of time these days working on the CLI experience, so I expect more improvements to come here.)

PR #3994, opened 2020-11-21
arvo: unflop the spur in scry

Fang-:
As requested, no longer use reversed paths in sley-style peek handling.

I think this is the last of the "reverse path" shenanigans.

PR #4010, opened 2020-11-24
arvo: version negotiation, reform

joemfb:
This PR adds kelvin-version negotiation between arvo and vere. Additionally, it reimplements arvo upgrade and refines arvo's interfaces, for future compatibility.

- version negotiation

On boot and every subsequent, vere injects a `%wyrd` event to arvo, declaring its own version, and the arvo kelvins down to which it supports. Arvo checks these versions, optionally responding with a `%wend` effect to signal downgrade (specified, unimplemented). Additionally, arvo stores the runtime-supported versions, and checks for compatibility on every upgrade, failing the upgrade when versions are unsupported.

Finally, arvo intercepts error notification events, setting a lag-bit in it's state if the error signified a failed upgrade. This bit can be read from arvo's namespace.

- event/effect routing reform

Recapitulating the %pass/%give pattern between the vanes, the runtime now explicitly routes events to their intended target. Concretely, this means that wire on external events begins with a vane letter, in all case except for events on arvo proper, and effect wires no longer begin with an empty segment.

- arvo-proper event reform

%pass to arvo is implemented, %give to the runtime -- but really to arvo -- is deprecated, and arvo events are separated into external ($wasp) and internal ($waif) structures (for instance, only vere can inject entropy via %wack). Also, %verb is no longer (only) a toggle, and %crud wraps an $ovum instead of a $card..

- filesystem

Arvo now maintains a trivial filesystem, tracking the source of anything in the kernel, and calculating deltas to drive upgrades.

- upgrade reform

Arvo upgrades via a kelvin-tagged structure ($heir), containing persistent state, in-progress state (moves, accumulated output), and an upgrade recipe. This allows complete upgrades to complete in one move. Since the structure is version-tagged, all upgrades go through the +load arm. Additionally, arvo-only kernel upgrades are statically-typed, and a bug in full-reset upgrades involving hoon kelvin changes has been fixed.

- scry reform

Arvo stores the timestamp of the last successful event in its persistent state, and supports external `+peek` at the current date through a deconstructed data structure. Arvo implements its own trivial portion of the namespace (`care=%$`). And the scry sample of the vanes has been cleaned up. (This depends on #3994, and currently includes a partial cherry-pick, which will be reverted).

This PR is complete except for two things:

- final kelvin assignment, deferred as it involves coordinated changes to all layers
- vane structures in %lull, deferred as it is basically part of the above

Note that both `:aqua` and king-haskell will need to be updated -- %wyrd version negotiation is mandatory, as are the new event/effect wire structures.


galenwp:
"king"

pkg/urbit/vere/lord.c
```diff
@@ -382,13 +378,25 @@ _lord_plea_peek_bail(u3_lord* god_u, u3_peek* pek_u, u3_noun dud)
...
 _lord_plea_peek_done(u3_lord* god_u, u3_peek* pek_u, u3_noun rep)
 {
+  //  XX review
+  //
+  if (  (u3_pico_once == pek_u->typ_e)
+     && (u3_nul != rep) )
+  {
+    u3_noun dat;
+
+    if ( c3y == u3r_pq(u3t(rep), c3__omen, 0, &dat) ) {
```
belisarius222, line 388: Can you explain what this is doing?

pkg/arvo/sys/arvo.hoon
```diff
@@ -913,464 +1409,284 @@
...
+                   van
+                 %+  roll  van.job
+                 |=  [[(cask hoof:part)] =_van]
+                 ^+  van
+                 ?>  ?=(^ bod)
+                 =/  way  (wilt p)
+                 (~(put by van) way (mint $:u.bod way /sys/vane/[p]/hoon q))
+               ==
+    ::
+        %whom  ..poke(who `p.gub)
```
belisarius222, line 1670: should assert lack of prior identity

Fang-:
I took the liberty of fixing `+sloy` to support the empty vane/care that we now use for scries into arvo. Figured it wasn't worth the PR overhead.

I'd leave style comments, but it's not that important. Just want to say: you really like your kingside huh. (;

PR #4016, opened 2020-11-25
hark-store: fix caching of merged notifications

### LINKED ISSUE #3977: hark: dot persists if notification is auto-dismissed by being in the resource
Replication — 

1. See notification for a DM or something. It's the only notification.
2. Click the notification. Dot disappears.
3. The person sends another DM while you're in the chat itself. Dot appears.
4. Go to notifications — no actual unread notification is apparent.

liam-fitzgerald:
If a notification came in that merged with a notification that had
already been read, then it would not be added to the cache properly and
therefore would not correctly send %read updates when %read-index was
called.

cc: @vvisigoth 

Fixes #3977

### DIFF
--- a/pkg/arvo/app/hark-store.hoon
+++ b/pkg/arvo/app/hark-store.hoon
@@ -166,11 +166,16 @@
         ?~  existing-notif
           notification
         (merge-notification:ha u.existing-notif notification)
+      =/  new-read=?
+        ?~  existing-notif
+          %.y
+        read.u.existing-notif
+      =.  read.new  %.n
       =/  new-timebox=timebox:store
         (~(put by timebox) index new)
       :-  (give:ha [/updates]~ %added last-seen index new)
       %_  state
-        +  ?~(existing-notif (upd-unreads:ha index last-seen %.n) +.state)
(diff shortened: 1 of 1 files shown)


tacryt-socryp (approved):
The merging semantic is quite complex... Wondering about whether we should have a meeting once I’m back online to thoroughly discuss desired semantics and means of achieving those semantics. We wouldn’t necessarily change anything, I’m just curious what the state space of possible design decisions for this are. Nonetheless, LGTM.

pkg/arvo/app/hark-store.hoon
```diff
@@ -166,11 +166,16 @@
...
+        read.u.existing-notif
+      =.  read.new  %.n
       =/  new-timebox=timebox:store
         (~(put by timebox) index new)
       :-  (give:ha [/updates]~ %added last-seen index new)
       %_  state
-        +  ?~(existing-notif (upd-unreads:ha index last-seen %.n) +.state)
+        +  ?.(new-read +.state (upd-unreads:ha index last-seen %.n))
         notifications  (put:orm notifications last-seen new-timebox)
       ==
```
Fang-, line 180: In awe of this `%_`.

PR #4017, opened 2020-11-25
hark: update mark conversion for editable comments

liam-fitzgerald:
Updated mark conversion so that edits to comments are notified
separately to new comments.

UX note (cc: @urcades ):
This will notify users if they're subscribed to the parent node i.e. if they've unmuted the book or link that the comment was left on. Should we expose following a comment for updates instead? e.g. click 'notify me on update' -> get notified when comment changes.

### DIFF
--- a/pkg/arvo/mar/graph/validator/link.hoon
+++ b/pkg/arvo/mar/graph/validator/link.hoon
@@ -5,8 +5,9 @@
   ++  noun  i
   ++  notification-kind
     ?+  index.p.i  ~
-      [@ ~]    `[%link 0]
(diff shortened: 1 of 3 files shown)


tacryt-socryp (approved):
Really straightforward. LGTM

urcades:
> This will notify users if they're subscribed to the parent node i.e. if they've unmuted the book or link that the comment was left on.

This seems correct — I think this is the ideal stance to take here, as per-comment/post notify toggling sort of exacerbates the issue we're finding now, with channel unmuting feeling tedious.

PR #4018, opened 2020-11-25
publish: surface writers in new channel creation

matildepark:
- If we have no writers set for the notebook in our group, everyone can write.
- Allow users to set writers on channel creation if the new channel screen is in a managed context. All added writers are verified they're ingroup✔️ on FE to prevent back-end crashes (`addTag` requires the added ships are already members).
- Surfaces current writers on the channel settings page, decorated in mono and cited.

Screen Shot 2020-11-24 at 11 11 15 PM


liam-fitzgerald (approved):
LGTM!

PR #4022, opened 2020-11-25
pull-hook: unconditionally restart missing subscriptions

liam-fitzgerald:
Restarts subscriptions missing from the bowl every on load


liam-fitzgerald:
superseded in favour of #3416

PR #4023, opened 2020-11-25
graph-store: add basic sanity checker

liam-fitzgerald:
A small subset of %sane's functionality to go out in the next OTA. When paired with #4022 , should fix all broken subscriptions for graph-store. I'm aware this a slight layering violation, but priority should lie with the stability of the livenet. We can transplant this logic out into `%sane` later.


liam-fitzgerald:
superseded in favour of #3977 being finished and ready

PR #4024, opened 2020-11-25
build: disable ci release tarball upload on pull request

brendanhay:
* Splits the tarball upload into a separate workflow that is only triggered when any kind of draft/create/update of a release occurs. Currently it just continues to upload an artefact named according to the pattern: `gs://bootstrap.urbit.org/ci/urbit-v${version}-${system}-${GITHUB_SHA:0:9}.tgz` - I assume you'll want this to match the release name?
* The (vere) tests potentially rerun for every push and when any pull request is initiated. I've disabled this behaviour so it'll only occur when a pull request is created/pushed to.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,11 +30,9 @@
 #
 #   https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
 
-name: Build, test, and upload urbit release tarball
+name: Build and test urbit binaries
 
-on:
-  push: null
-  pull_request: null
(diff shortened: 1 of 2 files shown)


brendanhay:
Looks like the tests were skipped when I created this PR - which is at odds with how I "imagined' things to work according to the documentation. I'll dig around and see if I missed something obvious.

joemfb (approved):
I still don't know exactly how I want CI to behave, wrt to both tests (gc?, ASan?) and release artifact triggers. But we can sort it out later. These changes fix the PR-from-fork and static-binary issues.

LGTM

PR #4027, opened 2020-11-25
king: port the new lmdb event log changes.

eglaysher:
This ports the new event log format changes to king, including the event browser.


eglaysher:
I've tested this by booting a few comets, deleting their event logs, and then making sure C vere can replay their logs. Can you think of other tests I should try?

joemfb (approved):
LGTM, thanks!

As long as we can boot with vere, replay with kh, and vice-versa, I'm happy.

PR #4030, opened 2020-11-26
lib/hooks: virtualise kick handler to ensure subscription integrity

liam-fitzgerald:
- Virtualises the kick handler for the inner door in lib/pull-hook, so
that crashes in the handler do not cause a dangling resource with no
subscription. Additionally, failed kicks now cause the sync to be
dropped into a failed-kicks map in the state, where we attempt to
recover from the failure in the on-load. Unfortunately, we can't catch
blocking scries, so we sanity check the path (if it's a gall scry, ensure
that the ship and case are [our now]:bowl).
- If the %kick handler of the inner-core crashes, then we never get to
resubscribe, thus leaving a dangling entry in the state with no
corresponding subscription. Updates the pull-hook-action pokes to not
crash when given a dangling entry.

### DIFF
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -158,6 +158,7 @@
         og   ~(. pull-hook bowl)
         hc   ~(. +> bowl)
         def  ~(. (default-agent this %|) bowl)
+    ::
     ++  on-init
       ^-  [(list card:agent:gall) agent:gall]
       =^  cards  pull-hook
@@ -311,6 +312,7 @@
     --
   |_  =bowl:gall
   +*  og   ~(. pull-hook bowl)
+  ::
   ++  mule-scry
     |=  [ref=* raw=*]
     =/  pax=(unit path)
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -49,13 +49,21 @@
       inner-state=vase
   ==
 ::
++$  base-state-1
+  $:  base-state-0
+      failed-kicks=(map resource ship)
+  ==
+::
 +$  state-0  [%0 base-state-0]
 ::
 +$  state-1  [%1 base-state-0]
 ::
++$  state-2  [%2 base-state-1]
+::
 +$  versioned-state 
   $%  state-0
       state-1
+      state-2
   ==
 ::
 ++  default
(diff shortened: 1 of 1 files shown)


pkg/arvo/lib/pull-hook.hoon
```diff
@@ -286,6 +311,58 @@
...
+        p.u.cas
+      now.bowl
+    ::  catch bad gall scries early
+    ?:  ?&  =((end 3 1 i.u.pax) %g)
+            ?|  !=(`our.bowl ship)
+                !=(dat now.bowl)
+            ==
+        ==
+      ~
+    ``.^(* u.pax)
```
tacryt-socryp, line 339: Can't this scry still crash, despite your safety checks? Is it run virtualized as well?

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -286,6 +311,58 @@
...
+  ++  handle-kick
+    |=  [rid=resource =ship]
+    ^-  (quip card _state)
+    =/  res=toon
+      (mock [|.((on-pull-kick:og rid)) %9 2 %0 1] mule-scry)
+    =/  pax=(unit path)
+      !<  (unit path) 
+      :-  -:!>(*(unit path)) 
+      ?:(?=(%0 -.res) p.res ~)
```
philipcmonk, line 348: That's some spicy hoon.  Seems legit, though, as in Gall.  It's the only way to intercept a scry and get a typed result afaik.

pkg/arvo/lib/pull-hook.hoon
philipcmonk: Yes, but hopefully it's less likely.

philipcmonk (approved):
This seems correct, and if it's working when you test the failing case then it should be fine.

pkg/arvo/lib/pull-hook.hoon
joemfb: Shouldn't that be `[~ p.res]`, assuming the product is a path?

joemfb: Also, if you're only every scrying for a path, it should be fine to skip the evil-vase construction and just clam.

philipcmonk: It produces a unit path, he has no need to distinguish between the app producing null or crashing.

joemfb: Ok. I'd still encourage clamming.

matildepark:
Ironically this PR is blocked on the one `::` comment @tacryt-socryp mentioned earlier — but also, does this affect import/export ... ? Versioned state changes I see.

PR #1123, opened 2019-01-07
switch u3r_mug from FNV to Murmur3

joemfb:
dual of urbit/arvo#965

PR #4031, opened 2020-11-26
mug: again

joemfb:
This PR replaces the `+mug` (31-bit non-zero noun hash) algorithm for cells. The previous cell `+mug` commutatively combined the head and tail via XOR, which was suboptimal... (see #1123 and urbit/arvo#965, *mea maxima culpa*)..

Additionally, this PR removes a spurious assertion in +muk, and replaces the seed-length assertion with 32-bit truncation.


philipcmonk:
How has this been tested?  In particular, have we done some tests comparing the jet with the hoon?

joemfb:
The smoke-tests that are checked in have been run independently against the hoon and C (by virtue of the staging order). I'll try to run with the jet-dashboard testing flags on the mug jet.

Issue #4007, opened 2020-11-24
File upload doesn't allow audio


galenwp:
**Describe the bug**
Uploading a `.wav` fails on a link

**To Reproduce**
1. Connect S3
2. Try uploading a .wav file to a link collection
3. Nothing happens

**Expected behavior**
The .wav being uploaded and linked to.

vvisigoth:
What is console error?
—
~poldec-tonteg
http://urbit.org

matildepark:
Can't replicate — 
<img width="959" alt="image" src="https://user-images.githubusercontent.com/20846414/100129212-4469a380-2e4f-11eb-8ab6-a6109d2bcfdc.png">

Possibly what you're referring to is that the .wav is *too big*, not that it's a .wav. @tylershuster Did we ever figure that one out?

tylershuster:
I think it's this https://github.com/aws/aws-sdk-js/issues/2083 but it's not solved

matildepark:
So we have to chunk upload?

tylershuster:
Yeah, I'm pretty sure that's the case.

galenwp:
Yeah it was like 200MB

tylershuster:
One thing getting in the way of me fixing this is that the S3 provider needs to expose the ETag header to process multipart files. [DigitalOcean does not have a GUI to do this](https://www.digitalocean.com/community/questions/spaces-unable-to-do-multipart-upload-in-js), which will get in the way of a lot of people's configuration.

matildepark:
Could we check for it and fall back? We could also refer people to these instructions.

Issue #3992, opened 2020-11-21
"Drop a file to upload" strange behaviour


mike-cr:
**Describe the bug**
If you highlight text in a chat channel then drag & drop it, the "drop a file to upload" overlay sometimes gets stuck and it's difficult to get rid of. It happens if you drop it somewhere it thinks is invalid where the cursor says no. It seems the invalid locations are related to the content beneath the overlay.

**To Reproduce**
1. Go to chat
2. Highlight some text
3. Drag and drop it somewhere it doesn't like
4. Observe you're now stuck with the "drop a file to upload" overlay

**Expected behavior**
Overlay should disappear after drop

**Screenshots**
![Peek 2020-11-21 19-44](https://user-images.githubusercontent.com/32793870/99869848-1576e600-2c33-11eb-895f-755bec8f018f.gif)

**Desktop (please complete the following information):**
 - OS: Arch Linux
 - Browser qutebrowser (qtwebengine aka chromium engine)
 - Base hash of your urbit ship: 0kb4n

**Additional context**
I tested this in firefox and it doesn't seem to have the problem as there are no invalid drop locations. Unsure if this is just a qutebrowser problem or chromium problem

mike-cr:
this problem crops up semi-regularly cos I tend to mindlessly highlight text as I'm reading a page (idk why) and I'll unintentionally drag & drop it

mike-cr:
ok I installed chromium and it doesn't have the problem either, so this is just qutebrowser being weird I guess. It actually seems like qutebrowser changes the cursor to invalid when it's moving rather than being related to the chat content -- it seems not to be able to drop while in motion. Dunno if this is something you guys can fix

tylershuster:
Yeah, this is possible to fix — probably don't need to worry about handling drag-and-drops when the user drags in text.

Fang-:
It's possible for me to reproduce in Firefox by pressing escape to cancel the drag. The cursor lets go of whatever I was dragging, but the overlay remains. I'm forced to navigate to the homescreen to dismiss it.

PR #4032, opened 2020-11-26
s3: upload behavior

tylershuster:
Tackling a few S3 issues:

- [x] #3992
- [x] https://github.com/urbit/landscape/issues/183
- [x] https://github.com/urbit/landscape/issues/191
- [x] #4039
- [ ] ~https://github.com/urbit/urbit/issues/4007~

Feel free to add others


tylershuster:
Might not be able to do #4007 due to larger concerns (see issue).

tylershuster:
Punting 4007 until later, as it's not strictly a code change, but perhaps more of a messaging one.

Would appreciate a review from @liam-fitzgerald and of course @matildepark 

This unifies all S3-related functions into two files:
1. `useS3`, the heavy lifter and only interface with the bucket. Eventually it will have the `S3` parameter removed to use global state hooking.
2. `withS3`, a light HOC around `useS3` that allows it to be used in class components, provided they have the `s3` prop (for now)

`SubmitDragger` and `useDrag` have been lightly touched to fit with this model.

`LinkSubmit` is now a functional component.

matildepark:
It's easy to imagine dragging and dropping more than just images into Publish, but it presents the upload as `filename.type` regardless of filetype. If it's not an image it shouldn't have that `!` so it presents as a link.

matildepark:
If it doesn't increase the scope too much ... I wonder if we could just embed uploaded multimedia anyway?

image

tylershuster:
Easy enough for the first. We already have the Mime. The second one is more of a remotecontentpolicy thing. I'd punt that to a more wholistic pass at the editor.

matildepark:
They call me the PROPWEAVER

pkg/interface/src/logic/lib/useDrag.ts
```diff
@@ -53,6 +80,18 @@ export function useFileDrag(dragged: (f: FileList) => void) {
...
+    const mouseleave = (e) => {
+      if (!e.relatedTarget && !e.toElement) {
+        setDragging(false);
+      }
+    };
+    document.body.addEventListener('mouseout', mouseleave);
+    return () => {
+      document.body.removeEventListener('mouseout', mouseleave);
+    }
+  }, [setDragging]);
```
liam-fitzgerald, line 93: `setDragging` is unnecessary

pkg/interface/src/logic/lib/useS3.ts
```diff
@@ -1,51 +1,96 @@
...
   );
 
-  return { canUpload, upload, uploadDefault };
-}
+  const uploadDefault = useCallback(async (file: File) => {
+    if (s3.configuration.currentBucket === "") {
+      throw new Error("current bucket not set");
+    }
+    return upload(file, s3.configuration.currentBucket);
+  }, []);
```
liam-fitzgerald, line 70: missing `s3` dependency

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
```diff
@@ -0,0 +1,230 @@
...
+import { hasProvider } from "oembed-parser";
+
+interface LinkSubmitProps {
+  api: GlobalApi;
+  s3: S3State;
+  name: string;
+  ship: string;
+};
+
+const LinkSubmit = (props: LinkSubmitProps) => {
```
liam-fitzgerald, line 18: seems a little long, wonder if it's maybe worth turning into a Formik form?

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
```diff
@@ -0,0 +1,230 @@
...
+    }
+    return link;
+  };
+
+  const onFileDrag = useCallback(
+    async (files: FileList | File[], e: DragEvent) => {
+      if (!canUpload) {
+        return;
+      }
+      uploadDefault(files[0]).then(setLinkValue);
```
liam-fitzgerald, line 94: not returning or awating a promise?

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+  const onPaste = useCallback(
+    (event: ClipboardEvent) => {
+      if (!event.clipboardData || !event.clipboardData.files.length) {
+        return;
+      }
+      event.preventDefault();
+      event.stopPropagation();
+      uploadDefault(event.clipboardData.files[0]).then(setLinkValue);
+    }, [setLinkValue]
```
liam-fitzgerald, line 120: uploadDefault missing from dependencies

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
```diff
@@ -0,0 +1,230 @@
...
+  );
+
+  const onKeyPress = useCallback(
+    (e) => {
+      if (e.key === 'Enter') {
+        e.preventDefault();
+        doPost();
+      }
+    }, []
```
liam-fitzgerald, line 129: doPost missing from dependencies, but because it's not memoised you can probably drop the useCallback here entirely

liam-fitzgerald (approved):
just some hook dependency issues, you're getting quite good with hooks though!

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
tylershuster: Not sure what you mean. This function doesn't need to return anything:
image

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
tylershuster: I just figured out hooks, but I haven't quite grokked formik. If I try more than one new tech per PR my head begins to spin a little.

tylershuster:
I'm blocked on developing this due to the iconindex and channelqueue things. Though, I think we should just push this set and I'll work on the embeds separately.

matildepark:
You can navigate the channelqueue thing by just developing off a fakezod like the old days.

matildepark:
You can also just comment out line 13 of base.ts in src/logic/subscription to prevent that channelqueue crash.

pkg/interface/src/views/apps/links/components/LinkSubmit.tsx
liam-fitzgerald: this callback is marked `async` so surely you'd use `await` instead of chaining promises? Either that or drop the `async`

tylershuster: Good point. Must have been an oversight.

tylershuster:
I will address Liam's one comment tonight, then this should be done￼

tylershuster:
what

matildepark:
Ha that was an accident. I restored the branch.

matildepark:
The dragger in Publish takes up the full screen — changing `position="static"` in to `position="relative"` in MarkdownField.tsx will fix this, but unsure if it has any ramifications (can't see any when I responsively play with it...)

matildepark (approved):
Fine otherwise — if you can address that one thing I'll merge. Remember to get commits into hygiene too.

PR #4036, opened 2020-11-28
various: start threads with unit of tuple args

Fang-:
As suggested in https://github.com/urbit/urbit/pull/3738#issuecomment-713031839.

Instead of a list of arguments with trailing `~`, dojo now passes threads and generators a a tuple of the arguments. If no arguments are given, still passes `~`. We update the generaters/threads themselves, and all other callsites, to match.

Notable changes include:
- In the `path` argument case, we can no longer distinguish between no arguments and `/` (without adding additional clarification, like `unit` or explicit trailing `~`).  This currently affects the pill generators and `|init-oauth2`, but the `/` case wasn't reasonable for those anyway.
- Similarly, the atom argument case is affected, making it harder to distinguish between no arguments and `0`. For `:dns|request` and `|moon` this doesn't matter. `|trim` can no longer send a `%trim` with priority 0, but arguably only the system should be able to send that anyway?
- Optional arguments in some cases lead to type errors. In `|mount`'s case, `?(path [path @tas])` is a `fish-loop`. We're forced to move the optional mount point name into the named arguments section.
- Threads and generators that take lists of arguments, like `+cat` and `-test`, now require  an explicit trailing `~` when called.

I didn't test _all_ of these, but I made sure to hit the non-trivial cases at least.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -728,6 +728,16 @@
     ::
     ++  dy-cage       |=(num=@ud (~(got by rez) num))   ::  known cage
     ++  dy-vase       |=(num=@ud q:(dy-cage num))       ::  known vase
+    ::
+    ++  dy-some
+      |=  src=(list dojo-source)
+      ^-  vase
+      ?~  src   !>(~)
+      %+  slop  !>(~)
+      |-
+      ?~  t.src  (dy-vase p.i.src)
+      (slop (dy-vase p.i.src) $(src t.src))
+    ::
     ++  dy-sore
       |=  src=(list dojo-source)
       ^-  vase
@@ -850,7 +860,7 @@
         [%pass /wool %agent [our.hid %spider] %watch /thread-result/[tid]]
       %-  he-card
       =/  =cage  ::  also sub
-        [%spider-start !>([~ `tid fil (dy-sore src)])]
+        [%spider-start !>([~ `tid fil (dy-some src)])]
       [%pass /wool %agent [our.hid %spider] %poke cage]
     ::
     ++  dy-make                                         ::  build step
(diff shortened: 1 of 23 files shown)


belisarius222:
> harder to distinguish between no arguments and 0

I don't think structural ambiguity is inevitable here.  Is there a good reason not to make the positional-arg sample a `(unit tuple)`, instead of a `$@(~ tuple)`?  As far as I can tell, the only downside would be that a generator would need an initial `~` in its sample type if it required positional arguments, which isn't onerous.

philipcmonk:
If we do that it should be unit of a vase instead of a vase of a unit.
Ergonomically it would be annoying to have to ?> ^ every time, but it is a
more precise way of saying it.

ohAitch:
Any reason not to do the vase of a unit?

philipcmonk:
It's "wrong" in the sense that the unit part isn't dynamically typed. It's
just that maybe you got an argument.

However, this pattern is probably worth preserving:

=+  !<  [~ =argument] vase

I guess technically if it's a unit of a vase you could still do:

=+  !<  =argument  +.vase

Fang-:
As per out of band discussion, the generators side of this is a more complicated question, but we can certainly do the proposed unit-ization of arguments for threads right now.

So, just pushed a commit that does that, reducing the scope of the PR accordingly.

philipcmonk (approved):
LGTM.  One of CI failed with HTTP 500 errors installing nix stuff, so I kicked them off again, but I'm sure they're fine

matildepark:
@Fang- You may want to review that all the threads reflect these changes now that release/next-userspace is merged into the candidate.

joemfb:
We can catch any stragglers in a follow up PR.

PR #4038, opened 2020-11-30
chat-fe: autodismiss mentions whilst in channel

liam-fitzgerald:
Automatically dismiss relevant mentions when a chat is marked as read

Fixes: #3977

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
@@ -182,6 +182,7 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
     if (this.props.unreadCount === 0) return;
     this.props.api.chat.read(this.props.station);
     this.props.api.hark.readIndex({ chat: { chat: this.props.station, mention: false }});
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
```diff
@@ -182,6 +182,7 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
     if (this.props.unreadCount === 0) return;
     this.props.api.chat.read(this.props.station);
     this.props.api.hark.readIndex({ chat: { chat: this.props.station, mention: false }});
+    this.props.api.hark.readIndex({ chat: { chat: this.props.station, mention: true }});
```
matildepark, line 185: Why do we dismiss mention notifications in a different way than regular ones?

liam-fitzgerald: Just a peculiarity of the API I guess. They're not merged with regular notifications, so we need to address them differently.

matildepark (approved):
Can reproduce on livenet, can confirm this fixes it in this branch.

PR #4042, opened 2020-11-30
aqua: rework how it uses azimuth

ixv:
In order to make it much easier to work with fleets of ships, aqua now stores azimuth state internally rather than relying on a thread.

### DIFF
--- a/pkg/arvo/app/aqua.hoon
+++ b/pkg/arvo/app/aqua.hoon
@@ -21,11 +21,15 @@
 ::  We get ++unix-event and ++pill from /-aquarium
 ::
 /-  aquarium
-/+  pill, default-agent
+/+  pill, default-agent, aqua-azimuth, dbug, verb
 =,  pill-lib=pill
 =,  aquarium
 =>  $~  |%
-    +$  state
+    +$  versioned-state
+      $%  state-0
+          state-1
+      ==
+    +$  state-0
       $:  %0
           pil=pill
           assembled=*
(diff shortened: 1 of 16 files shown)


pkg/arvo/app/aqua.hoon
```diff
@@ -428,17 +480,47 @@
...
+      %init-azimuth
+    =.  azi.piers  *az-state
+    start-azimuth-timer
+  ::
+      %spawn
+    =.  state  (spawn who.act)
+    this
+  ::
+      %breach
+    ::  should we remove the pier from state here?
```
philipcmonk, line 501: No, because some of our tests are specifically to check that once a breach notification comes the defunct ship can't be reached.

pkg/arvo/app/aqua.hoon
```diff
@@ -428,17 +480,47 @@
...
   %+  turn-events  events
   |=  [ae=aqua-event thus=_this]
   =.  this  thus
   ?-  -.ae
+  ::
       %init-ship
+    ::  XX  Note that the keys that get passed in are unused. The keys field
+    ::      should be deleted now that aqua is capable of managing azimuth state
+    ::      internally. Its been left this way for now until all the ph tests
+    ::      can be rewritten
```
philipcmonk, line 522: To make a fake ship you just pass in keys of `~` still, right?

philipcmonk (approved):
Looks good!  Should be much easier to do this kind of testing

pkg/arvo/app/aqua.hoon
ixv: Yeah, it should probably be changed to take in `fake=?` instead

PR #4043, opened 2020-11-30
vere: add -Z for choosing scry export format

Fang-:
For use with other flagday logic. Writes the current keyfile to `.urb/put/archive/keyfile.key`.

In the future we may want broader control over the "scry read" output format, instead of always writing a jamfile.


ohAitch:
Comets will live?

philipcmonk:
I think so, we will certainly test

Fang-:
As per out of band discussion, replaced this with the more generic thing. Use `-Z` to specify an aura to render the scry result as. Equivalent of -O is now  `urbit -X /j/vile -Y /keyfile -Z uw your-pier`. This gets you a `.txt` file instead of `.key` but that shouldn't matter.

@ixv is `-o` just `-X` with hardcoded paths, or am I missing something?

pkg/urbit/vere/pier.c
```diff
@@ -491,6 +491,23 @@ _pier_on_scry_done(void* ptr_v, u3_noun nun)
...
+    //
+    u3_atom out;
+    c3_c*   ext_c;
+    {
+      u3l_log("xxx\n");
+      u3_atom puf = u3i_string(u3_Host.ops_u.puf_c);
+      if ( c3y == u3r_sing(c3__jam, puf) ) {
+        out   = u3qe_jam(res);
+        ext_c = "jam";
+      }
```
philipcmonk, line 504: I think technically we leak `puf` in this branch, and maybe `res` overall.  Doesn't really matter for what we're doing, but if we reuse the code later.  At least a comment mentioning it would be good.

pkg/urbit/vere/pier.c
```diff
@@ -491,6 +491,23 @@ _pier_on_scry_done(void* ptr_v, u3_noun nun)
   else {
     u3l_log("pier: scry succeeded\n");
 
+    //  serialize as desired
+    //
+    u3_atom out;
+    c3_c*   ext_c;
+    {
+      u3l_log("xxx\n");
```
philipcmonk, line 499: printf

philipcmonk (changes requested):

pkg/urbit/vere/pier.c
Fang-: We're not leaking `res`, because it comes from `nun`, and we `u3z` that at the end of this function. Good catch on `puf` though, I'll just fix that!

Fang-:
Actually having some trouble with scries not succeeding in local testing now, but that remains even when reverting to pre-PR state, so I'm probably just doing something dumb? @ixv regardless give this a shot and lmk if you run into trouble.

pkg/urbit/vere/pier.c
joemfb: You can use `u3r_sing_c()` and avoid allocating `puf`.

joemfb: Nevermind, I guess you still need it for `+scot`. You'll need to `u3k` it there.

joemfb:
@Fang-, scry was broken. fixed in #3942. This should include those changes, and I think, target `release/next-vere`.

ixv:
Works for me, @joemfb, the change you're referring to seems to be in the base branch. @Fang- maybe you built the wrong branch?

ixv (approved):

joemfb:
Ok, good. Regardless, this needs to target a pre-breach branch and be available on the export side, right?

philipcmonk:
That's correct.  Works for me too, btw

Fang-:
Removed `-o` as per out-of-band discussion with @ixv, since you can get equivalent behavior by using `-X` and `-Y`.

pkg/urbit/vere/pier.c
Fang-: Weird, this wasn't `foul`ing on me without the `u3k`... But I believe you're right.

philipcmonk (approved):
Working great for me

PR #4047, opened 2020-12-01
clay, gall, jael: cleanup protocol versions

philipcmonk:
+riff-any is all clay requests except "backfill" requests.  Change to
`$%` from `$^`, which was used to distinguish originally non-versioned
requests.

+fill is backfill requests and had no version number, so we add one.

We do not have version numbers on responses since those are implied by
the request.  If someone requests at version `n` and you're at `n+1`,
you must respond in the format of `n`.

If someone requests at version `n+1` and you're at `n`, you crash;
though possibly you should be able to respond with message "I only know
up to `n`", in which case they may be able to re-request at `n`.  In
either case, the version of the response is dictated by the request.


philipcmonk:
Went ahead and did gall and jael too.  All tested except Jael, which you can't test on a fake network.  Will try moons in aqua, and I've marked down to test moons on the testnet, which will exercise this.

joemfb (approved):
LGTM

I'll merge shortly, out-of-band.

PR #4048, opened 2020-12-01
Ames flat packet format

belisarius222:
Revises the Ames packet format to be easier to parse and ideally save a few bits here and there.  Relays no longer need to understand `+jam` and `+cue`, raising the ceiling on relay speed.  Also, to save bandwidth, instead of including `[sndr rcvr sndr-life rcvr-life]` in the ciphertext for authentication purposes, that's now authenticated explicitly as an [AES-256 SIV](https://tools.ietf.org/html/rfc5297) "associated data vector" but not actually sent over the wire.

This packet format is called "flat" because it is less recursive than its predecessor.

```
32  bits:  header
48  bits:  (optional) 48-bit origin
4   bits:  sender life (mod 16)
4   bits:  receiver life (mod 16)
<variable> sender address
<variable> receiver address
128 bits:  SIV
16  bits:  ciphertext size
<variable> ciphertext
```

The header is:
```
3  bits: reserved
1  bit:  is-this-ames? (i.e. not using a different protocol, such as the planned remote scry protocol)
3  bits: ames protocol version (currently 0)
2  bits: sender address size
2  bits: receiver address size
20 bits: checksum (truncated `+mug`)
1  bit:  is-this-relayed? (if set, `origin` will be present in the body)
```

The `life-tick`s are not required for security, and are only present for quicker filtering of honest packets sent to or from a stale life.  Honesty about life information is ensured not by this field, but by the AD vector, which includes the full life data, not truncated.


philipcmonk:
Hate to be that guy, but... "WIP" can't be in commit messages, at least not at the beginning.  Don't need to squash or anything, but they should be reworded.  If you explicitly want to call out they aren't expected to compile, you can put it after the `ames:`

belisarius222:
Sure, I'll fix up the commit messages tomorrow.

—
~rovnys-ricfer
https://urbit.org

belisarius222:
Closed in favor of https://github.com/urbit/urbit/pull/4051

PR #4049, opened 2020-12-01
hoon: +rep/+rip reform, atom gate additions

joemfb:
This PR combines `+rep`/`+repn` and `+rip`/`+ripn` in the obvious way, giving each the same *bloq*/stride API as `+end`/`+lsh`/`+rsh`.

It also adds:

- `+rut`: turn over an atom, in a *bloq*/stride, into a list
- `+run`: turn over an atom into an atom, each with the same *bloq*/stride
- `+sew`: inverse of `+cut`


joemfb:
The additions are eminently jet-able (many of their allocations can be entirely avoided), but not yet jetted. The new `+rip`/`+rep` jets simply preserve the existing jet coverage, which is fairly minimal. These limitations will be rectified in the future.

pkg/arvo/sys/hoon.hoon
```diff
@@ -910,31 +900,42 @@
...
 ::
+++  run                                                 ::  +turn into atom
+  ~/  %run
+  |=  [a=bloq b=@u c=@ d=$-(@ @)]
+  ^-  @
+  (rep a b (rut a b c d))
+::
+++  rut                                                 ::  +turn into list
+  ~/  %rut
+  |=  [a=bloq b=@u c=@ d=$-(@ @)]
```
Fang-, line 923: Any particular reason for this not taking `$-(@ *)` instead? (I don't feel strongly either way, this is a pretty thin wrapper.)

pkg/urbit/jets/c/rep.c
```diff
@@ -3,92 +3,199 @@
...
+         u3_atom b,
+         u3_noun c)
+{
+  if ( 1 == b ) {
+    return _block_rep(a, c);
+  }
+
+  if ( 0 == a ) {
+    return _bit_rep(b, c);
+  }
```
Fang-, line 170: Don't you want the order of these checks the other way around? Or does it not matter much?

pkg/urbit/jets/tree.c
```diff
@@ -1866,10 +1855,10 @@ static u3j_core _141_two_d[] =
   { "muk", 59, _141_two_muk_a, 0, _141_two_muk_ha },
   { "rap", 7, _141_two_rap_a, 0, _141_two_rap_ha },
   { "rep", 7, _141_two_rep_a, 0, _141_two_rep_ha },
+  { "new-rep", 7, _141_two_rep_a, 0, _141_two_rep_ha },
   { "rev", 7, _141_two_rev_a, 0, _141_two_rev_ha },
   { "rip", 7, _141_two_rip_a, 0, _141_two_rip_ha },
-  { "repn", 7, _141_two_repn_a, 0, _141_two_repn_ha },
-  { "ripn", 7, _141_two_ripn_a, 0, _141_two_ripn_ha },
+  { "new-rip", 7, _141_two_rip_a, 0, _141_two_rip_ha },
```
Fang-, line 1861: I think you can remove these now.

Fang- (commented):
Looks good. Just a small cleanup thing, and a couple of questions.

pkg/arvo/sys/hoon.hoon
```diff
@@ -910,31 +900,42 @@
...
+  ^-  (list @)
+  (turn (rip a b c) d)
+::
+++  sew                                                 ::  stitch into
+  ~/  %sew
+  |=  [a=bloq [b=@u c=@u d=@] e=@]
+  ^-  @
+  %^  cat  a
+    (can a b^e c^d ~)
+  (rsh a (add b c) e)
```
ohAitch, line 933: I think the 0 semantics you want here are
```suggestion
  %^  mix  e
    (can a b^e c^d ~)
 (end a (add b c) e)
```
or sth; in particular, `(sew 3 [2 1 %a] '\00\00\00\00b')` should definitely be `'\00\00a\00b'` not `'\00\00ab'`.

(Arguably `c` should also be limited so the operation cannot _increase_ the `met a` from e to the result, but it definitely should never decrease it)

pkg/arvo/sys/hoon.hoon
ohAitch: now we just need one that round-trips back to a (possibly different-`bloq`) output atom buffer ;)

ohAitch:
Don't love the unification tbh; it'd be different if there was a single `(repn (mul (bex a) b) ...)` in the codebase this was replacing, but there isn't, so this just adds a lot of `rep 0 x` / `rep y 1` noise to all call sites for in practice differently used operations.

pkg/arvo/sys/hoon.hoon
joemfb: That's a good point, but a bad example:

```
> `@t`(sew 3 [2 1 %a] '\00\00\00\00b')
'\00\00a\00b'
> `@ux`'\00\00\00\00b'
0x62.0000.0000
> `@ux`(sew 3 [2 1 %a] '\00\00\00\00b')
0x62.0061.0000
```

A better example would be:

```
> `@ux`(sew 3 [2 1 %$] '\00\00\00\00b')
0x6200
> `@t`(sew 3 [2 1 %$] '\00\00\00\00b')
'\00b'
```

which should roundtrip. I had switched to `+cat` to avoid measuring `e` or combining right and left shifts, but that's just wrong here.

I think this gate is strictly more useful if it doesn't limit `c` to avoid growing `e` -- `+cut` (and many other operations) treat atoms as if they were of infinite length (note that `+can` already truncates `d` to `c` bloqs). The runtime can mutate the storage behind `e` if the usual conditions apply (on our road, refcount 1, &c) and the result would fit in `e`. Or it can realloc in place, or allocate and copy.

pkg/urbit/jets/tree.c
joemfb: good point, thanks!

pkg/urbit/jets/c/rep.c
joemfb: We want to operate bloq-aligned where possible. Both jets (and the underlying `u3r_chop()` function) need to be rewritten, this just preserves existing behavior.

pkg/arvo/sys/hoon.hoon
joemfb: I was focused on `+run`, but this is a good point; there's no reason to restrict `+rut`. It should also be wet, then.

pkg/arvo/sys/hoon.hoon
ohAitch: You're right, `e` has conceptually infinite trailing zeroes(as a practical matter, it might be a 32-bit word which is getting `sew 0`n/`sew 3`d) and there's no reason to prevent writes to those.

ohAitch:
…similarly are there _literally any_ calls to `run` or `rut` with a non-1 count, or the equivalent of `rip a not1` calls/patterns they would replace
Like I would tentatively support expanding the `repn`/`ripn` calling convention to be `repn 0 n a`/`ripn 0 n a` instead of the prod `repn n a`/`ripn n a`, but for the common case this is just silly. I'm reminded of when `~(tap by x)` took a list argument to tap into, because you could in theory ever need it, and it practice it was called as `(~(tap by x))` without exception.

ohAitch:
Like I do concur the `repn` naming is not ideal and `runn` would be worse.

Perhaps you could stash the more general jetted ones as
```
++  rip
  =;  n  |=([a=bloq b=(list @)] (n:rip a 1 b))
  ~/  %rip-n
  |=  [a=bloq b=@u c=(list @)]
  ...etc
```
and so on so the namespace is unpolluted while the normal "split/join/map bloqs" operation is convenient; not sure that precise form is worth readability cost, but anything to get away from `(rip 3 1 a)` for "split bytes".

joemfb:
@ohAitch, I'm sympathetic to this argument, and have gone back and forth in considering the changes here. I'm curious, do you feel the same way about `+end`/`+lsh`/`+rsh`/&c? I don't recall hearing complaints about their API.

To muddy the water further while we're on the topic of proliferating atom api variants ... an awful of these call sites involve flops. I suspect that some are unnecessary endianness twiddling, but others are presumably legitimate. From a performance standpoint, it's somewhat tempting to have inverse versions of these (`+per`, `+pir`?) in order to avoid allocating the intermediate list.

ohAitch:
I don't feel the same way about `end`/`lsh`/`rsh`, because there a count of non-1 (on nonzero block sizes) is a common operation. (There's 3 `lsh 3 1` out of a total 26 `lsh 3` for example.) `end x 1` I could maybe see an alias for?

Re: flops, what would be the performance of changing a flop before/after to a `swp` after/before? In particular for `run` you already want a fresh rc-1 buffer…
(From an api standpoint, if they are added I would also recommend "namespacing" as `rev:rip` or `swapped:rip` or w/e)

pkg/arvo/lib/base64.hoon
```diff
@@ -128,6 +128,6 @@
     =/  len  (sub (mul 3 (div (add lat dif) 4)) dif)
     :+  ~  len
     %+  swp  3
-    (repn 6 (flop (weld dat (reap dif 0))))
+    (rep 0 6 (flop (weld dat (reap dif 0))))
```
ohAitch, line 131: This one seems particularly egregious wrt "just do all these operations in atom-space",
`(repn 0 6 (flop (weld dat (reap dif 0)))` is I think the same as `(lsh 0 dif (repn 0 6 (flop dat)))`.

Which granted seems to still has a 6-bit swap followed by an 8-bit swap so I'm not sure how much can be done…

ohAitch: (Honestly we should be doing our _parsing_ in atom-and-offset space, wherein the string slice could be simd converted to bytes-of-u6 and then squeezed down from "bloq" 6 bits to bloq bytes in a single jetted operation, but that's a whole other can of worms.)

joemfb: `(lsh 0 (mul 6 dif) ...)`, I think. I completely agree wrt parsing.

joemfb:
Existing code is written to existing APIs -- there are no `(rep !0 !1 ...)` calls because no such API existed. `+repn` and `+ripn` were added to optimize the base64 conversions, which should just be directly jetted, imo.

But they were addressing a legitimate gap in this API, and not just for bitwise calls. `(end 3 2 ...)` and `(rsh 3 2 ...)` are not unheard of; it can be clearer to express some operation as some number of bytes or words or whatever.

`+swp` and `+rev` have pretty basic jets, those should definitely be more efficient. `+rev:rep`/`+rev:rip` is a good idea.

pkg/arvo/lib/base64.hoon
ohAitch: `(lsh  .2.585 dif ...)`, quite

ohAitch:
I mean, I agree it can be clearer for some operations, I just disagree that `rep`/`rip` are those operations. And again, I'm not opposed to speculatively having a `repn` that takes both with or without namespacing, just to this being the "default" usage without a single `(rep !0 !1 ...)` call - is there an actual `lsh`/`mix` loop hiding somewhere for concatenating e.g. three-byte chunks? Is there code you wish to write yourself, and what does it do?

ohAitch:
Like, an argument can be made for not having `bloq` at all and phrasing all atom manipulations using a width in bits; but in practice, `32` is tremendously a more common bitwidth than `48` or particularly `47`, and warrants semantic compression on that basis. Having to specify `8` as `3 1` blows up that compression.

joemfb:
Maybe all these bloq/stride gates should actually be `$@(bloq [bloq stride])`?

joemfb:
I chatted with @frodwith about this, and he has a competing proposal: move bloq and stride out of the sample and into the context, call with `%*`, and (maybe later) add syntax to make it nicer to call. For instance:

```
+$  bloq  @
+$  step  _1
++  end
  =|  [=bloq =step]
  |=  a=@
  (mod a (bex (mul (bex bloq) step)))
```

- `(end %foo)` gets a bit
- `%*($ end bloq 3, +< %foo)` gets a byte
- `%*($ end bloq 3, step 2, +< %foo)` gets two bytes

An initial proposal for irregular syntax:

- `(end{bloq 3} %foo)`
- `(end{bloq 3, step 2} %foo)`

---

Whereas, my more incremental proposal would be something like so:

```
+$  bloq  @
+$  step  _1
+$  bite  $@(bloq [=bloq =step])
++  end
  |=  [a=bite b=@]
  =/  [=bloq =step]  ?^(a a [a *step])
  (mod b (bex (mul (bex bloq) step)))
```

- `(end 3 %foo)` gets a byte
- `(end 3 2 %foo)` gets two bytes

Plz comment.

/cc @belisarius222, @Fang-, @philipcmonk

Fang-:
> `%*` for that

Why not just `(end(bloq 3) %foo)`?

joemfb:
`%=` makes changes first, then evaluates with wing (pull the arm in this case). For this pattern, you need to first pull the arm, then make the changes.

We already have an example of this, these all work:

- `(%*(. slog pri 1) leaf/"foo" ~)`
- `%*($ slog pri 1, +< [leaf/"foo" ~])`
- `(=+(slog -(pri 1)) leaf/"foo" ~)`

While this does not:

- ` (slog(pri 1) leaf/"foo" ~)`

joemfb:
I should further note that `%=` needs to work the way that it does for it to be useful in loops and state machines. Consider `$(i +(i)` ...

ohAitch:
Oh no as soon as we get rid of braces people try to bring them back in! I'd be pretty happy if braces stayed reserved for interpolation functionality, including possibly future path interpolation and more general macro syntax support. 

I think I'd prefer `(end [3 2] a)`/`(end [3 2] a)` to `(end 3 a)`/`(end 3 2 a)`, it seems semantically cleaner; some kind of "keyword argument" style may be appropriate, but only for values present in <90% of calls, so `(%*(. end n 2) 3 %foo)`. "Make it bad to call now and maybe if you must nicer later" rather misses that the core of my complaint is it's nice to call now and the pr makes it somewhat less nice.

ohAitch:
I could *maybe* be convinced that `cat` in particular should have some manner of default bloq of _3_, given how often that comes up, but definitely not 0. And that seems much more shaky as a default than "count of 1, as it's been for as long as the language has existed".

joemfb:
I've restaged these changes with `$bite` arguments. The keyword args approach is intriguing, but feels like too much to bite off right now.

I intend to update `+end`, `+lsh`, and `+rsh` to use `$bite` as well, but I may do that later in a separate PR.

PTAL.

pkg/arvo/sys/hoon.hoon
```diff
@@ -821,14 +832,14 @@
   ::
 ++  bex                                                 ::  binary exponent
   ~/  %bex
-  |=  a=@
+  |=  a=bloq
```
ohAitch, line 835: Disagree, if I e.g. `(bex 10)` I'm probably not using it as a stride size, and if I `(dec (bex (bex 5)))` the two bexes definitely have different types - hence unopinionated `@`.

pkg/arvo/sys/hoon.hoon
```diff
@@ -821,14 +832,14 @@
...
   ^-  @
   ?:  =(0 a)  1
   (mul 2 $(a (dec a)))
 ::
 ++  can                                                 ::  assemble
   ~/  %can
-  |=  [a=bloq b=(list [p=@u q=@])]
+  |=  [a=bloq b=(list [p=step q=@])]
```
ohAitch, line 842: I'd err on the side of keeping these `@u`, it's a count; but not a strong opinion.

pkg/arvo/sys/hoon.hoon
```diff
@@ -910,27 +915,38 @@
...
-  |=  [bits=@ud x=@]
-  ^-  (list @)
-  ?:  =(0 x)  ~
-  [(end 0 bits x) $(x (rsh 0 bits x))]
-::
 ++  rip                                                 ::  disassemble
   ~/  %rip
-  |=  [=bloq x=@]
+  |=  [a=bite b=@]
+  =/  [=bloq =step]  ?^(a a [a *step])
```
ohAitch, line 921: Though this in particular seems *much* less clear than `[a 1]`

pkg/urbit/jets/c/rip.c
```diff
@@ -1,6 +1,90 @@
...
+*/
+#define DIVCEIL(x,y) \
+  (x==0) ? 0 :       \
+  1 + ((x - 1) / y);
+
+/*
+  `ripn` breaks `atom` into a list of blocks, of bit-width `bits`. The
+  resulting list will be least-significant block first.
+
+  XX TODO This only handles cases where the bit-width is <= 32.
```
ohAitch, line 22: Is this `bloq <= 32` or `bloq <= 5`?

ohAitch (commented):
Overall reasonable, ran into one kind of concerning XX - I'm assuming it came from ripn but it's still in the hot path now.

pkg/arvo/sys/hoon.hoon
joemfb: `bloq` is just an alias for `@`, communicating that the atom is used as a power of two -- that example is the same as the `(met 0 (met 0 atom))` in `+mat`.

pkg/arvo/sys/hoon.hoon
joemfb: I'm aiming for `$bite` to be an idiom in these jets -- you pass the `bloq` and `step` or the `step` is defaulted for you (where that always means one). But I'm not opposed to writing that as `[a 1]` ...

pkg/urbit/jets/c/rip.c
joemfb: This is `bloq <= 5`. What I've done with `+repn`/`+rep` and `+ripn`/`+rip` is simply to preserve the existing capabilities, while updating the API (in what I thought would be a difficult-to-stage manner, but that's not really the case now). I intend to rewrite these jets with full argument coverage ASAP.

joemfb:
@ohaitch, in response to your earlier comment: `+run` is turn over an atom, into an atom of the same bloq/step. It would certainly be useful to have a version that produces a possibly-different bloq/step, but it seems worth favoring the uniform case (ie, array/vector update). I guess `+run` could take a `(pair bite (unit bite)` and handle both, but I'm curious what you think.

Fang- (approved):
Changes seem good! I don't feel strongly about the `step`/`bloq` vs `@` changes.

pkg/arvo/sys/hoon.hoon
ohAitch: `met` doesn't produce `bloq`! And indeed, the outer one is a bloq but the inner one isn't, `a blocksize is the power of 2 size of an atom. ie, 3 is a byte as 2^3 is 8 bits.` - bloq implies 2^2^n not just any n used as a power of 2.

pkg/urbit/jets/c/rip.c
ohAitch: ohh, this is the now-inaccurately named `bits` <= 32, not the actual `bloq` parameter, `rip 9`/(now `run 9` not yet independently jetted) will still work. That seems fine if not ideal yup.

ohAitch:
To clarify my comment, I had completely missed the existence of `run`, despite it being in a sense the motivation of this PR. Privileged uniform case seems completely fine, we can do whatever keyword argument things if we get around to those.

belisarius222 (approved):
Ames changes all LGTM.  Didn't review the rest.

PR #4051, opened 2020-12-01
Ames flat packet format

belisarius222:
(Supersedes https://github.com/urbit/urbit/pull/4048.  Commits have been cleaned up and applied to the latest na-release/candidate.)

Revises the Ames packet format to be easier to parse and ideally save a few bits here and there.  Relays no longer need to understand `+jam` and `+cue`, raising the ceiling on relay speed.  Also, to save bandwidth, instead of including `[sndr rcvr sndr-life rcvr-life]` in the ciphertext for authentication purposes, that's now authenticated explicitly as an [AES-256 SIV](https://tools.ietf.org/html/rfc5297) "associated data vector" but not actually sent over the wire.

This packet format is called "flat" because it is less recursive than its predecessor.

```
32  bits:  header
48  bits:  (optional) 48-bit origin
4   bits:  sender life (mod 16)
4   bits:  receiver life (mod 16)
<variable> sender address
<variable> receiver address
128 bits:  SIV
16  bits:  ciphertext size
<variable> ciphertext
```

The header is:
```
3  bits: reserved
1  bit:  is-this-ames? (i.e. not using a different protocol, such as the planned remote scry protocol)
3  bits: ames protocol version (currently 0)
2  bits: sender address size
2  bits: receiver address size
20 bits: checksum (truncated `+mug`)
1  bit:  is-this-relayed? (if set, `origin` will be present in the body)
```

The `life-tick`s are not required for security, and are only present for quicker filtering of honest packets sent to or from a stale life.  Honesty about life information is ensured not by this field, but by the AD vector, which includes the full life data, not truncated.

### DIFF
--- a/bin/ivory.pill
+++ b/bin/ivory.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:85a4e1625d528b5fdc88faeff4fd288a23d6fbf1c11a846fc8f8d5b3cd38370f
-size 2118873
+oid sha256:5af78cbe22c8311af47b070be953ca0ec50f383bc20140f38e3364aa56a4e2a2
+size 2314689
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:b3e9d1637efea5862a114ce320f6e14afeb1f9f418428e0490d9c079a9becc56
-size 6712875
+oid sha256:8e5d6fb3716e9e4dd3267a08aa70e8f782e722930eab3045a90c48b3e51a1668
+size 6853071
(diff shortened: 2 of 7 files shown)


belisarius222:
This crashes each time it tries to forward a packet:
```
fail in _cj_hank_fill (c3y == u3h(sat))
bail: fail
```

I didn't see this crash until I applied my changes to the latest na-release/candidate.  Not sure what's going on, but that print comes from this line: https://github.com/urbit/urbit/blob/na-release/candidate/pkg/urbit/noun/jets.c#L1210

belisarius222:
The crash seems to have been caused by a stale ivory pill, which failed during a call to the new `+scot` jet.  Updating the ivory pill stopped the crashing, and I also fixed some ames tests that had been broken by the recent changes to vane scry handlers.

Back to a no-known-bugs state.

philipcmonk (approved):
Looks good to me.  For testing we'll want this on the testnet as soon as possible.  One of the Github actions (macos-latest) failed and I couldn't find a log, so I'm re-running them to try and get logs.

philipcmonk:
CI succeeded this time.  That's the great thing about deterministic builds -- if it doesn't work the first time you can safely try again!

PR #4052, opened 2020-12-01
dojo: correct mark conversion scry path

Fang-:
Fixes an instance that hadn't accounted for #3994 yet. This broke `&json &blit bel+~` and other conversion cases.

Went over the rest of dojo again, but I think this is the only one I missed. All other calls take arguments from contexts that should already have unflopped spurs.


belisarius222 (approved):
LGTM

PR #4055, opened 2020-12-01
term: 24-bit %klr support

Fang-:
Includes and supersedes #3761. Based off #4049 because this makes one (1) `+rep` call.

Does what it says on the can. Extends `+tint` to allow for 3-byte color specification, and updates webterm and the runtimes to render those correctly.

image

Webterm's line spacing is a little bit sad, but I couldn't easily find the thing to tweak to make them fit _just right_. It's like half a pixel, too...

Tagging @joemfb for hoon and C changes, @pilfer-pandex for haskell changes, and @matildepark for js changes + optional help with that line spacing thing.


matildepark (approved):
LGTM — you can find our `lineHeight` prop values here: https://github.com/urbit/indigo-react/blob/master/src/themes/light.ts#L138-L143

Try passing a few and see what aligns it; alternatively you could pass it an integer like `1` which might fix too.

ohAitch:
(I note for completeness that the one rep call it makes is, as expected, a `rep x 1`)

Fang-:
@matildepark `1` isn't small enough to make it align. That is, I can only make the spacing bigger. Looking at browser inspector, it's not clear where the spacing is coming from. There's no margin/padding, either... The line has some space on top, but not enough to cover for the entire thing...

spacing-weird

matildepark:
I can crack it open, how do I reproduce this ~ rainbow

Fang-:
Use [this](https://gist.github.com/Fang-/53c7c49dcd4577ff759e8dd1a8faf056) as your `/app/shoe.hoon`, then type `demo` into it.

matildepark:
I can't build vere on this? Weird. End snippet of running `make`:

```
checking for sed... /nix/store/a0zcpf8i0cifrhc4hnp39lzjx03mq3h2-gnused-4.8/bin/sed
checking for grep... /nix/store/j2j2zqinkwb3mgvgap06krb6dv47v7q3-gnugrep-3.5/bin/grep
checking for egrep... /nix/store/j2j2zqinkwb3mgvgap06krb6dv47v7q3-gnugrep-3.5/bin/grep -E
checking for a BSD-compatible install... /nix/store/g77kammvqqn63v01ppig43m2yn2s1lqd-coreutils-8.31/bin/install -c
checking for gcc... clang
checking whether the C compiler works... no
configure: error: in `/private/var/folders/t8/l_hrkql95rs4hyqvbg21r4300000gp/T/nix-build-curl-7.72.0.drv-0/curl-7.72.0':
configure: error: C compiler cannot create executables
See `config.log' for more details
builder for '/nix/store/vfbnampdla31qlll8z9w7478z2q6fxfn-curl-7.72.0.drv' failed with exit code 77
building '/nix/store/p4jjiwkzd373m8g8h1da2d7x15c4x9ah-herb.drv'...
cannot build derivation '/nix/store/0jg4nkii2psnh1wljd9s9myqv87wmx1c-urbit-0.10.9-rc1.drv': 1 dependencies couldn't be built
error: build of '/nix/store/0jg4nkii2psnh1wljd9s9myqv87wmx1c-urbit-0.10.9-rc1.drv', '/nix/store/p4jjiwkzd373m8g8h1da2d7x15c4x9ah-herb.drv' failed
```

Fang-:
Strange, not immediately sure what's up. Just in case: did you run this from inside `nix-shell`, and do `./configure` first?

matildepark:
No, I just ran `make` in my repo directory.

joemfb:
Running `make` from the top level should work. Not sure what the issue is, you may need to update nix.

matildepark:
Upgraded from 2.3.6 to 2.3.9, no dice. Weird. Will look around.

matildepark:
Appears to be [trouble in paradise](https://github.com/NixOS/nixpkgs/issues/104580) if you're using Nix on Big Sur. I'll try to play with these changes on Debian for now.

matildepark:
OK so after discovering that this Pinebook Pro, while slow as hell, can compile and run vere, boot a ship and get the interface dev environment going ... and thus I can do my job on it, I was able to get this running and poke around across my computers.

If you thought "there is no possible way this is Firefox specific!" — guess what...

Screen Shot 2020-12-01 at 11 49 29 PM

No matter though — it seems that using `display="flex"` instead of `display="block` in line.js fixes this with no real change to any other browser.

Screen Shot 2020-12-01 at 11 51 41 PM

pkg/hs/urbit-king/lib/Urbit/Arvo/Effect.hs
```diff
@@ -144,6 +144,7 @@ data Tint
     | TintK
     | TintW
     | TintNull
+    | TintTrue Atom Atom Atom
```
joemfb, line 147: I suspect the haskell thing to do is to make these a `Word8` ...

joemfb (changes requested):
@Fang-, sorry for messing up your PR with a force-push. #4049 is now backwards compatible wrt `+rep`, so you can just rebase onto (and target) `na-release/candidate`. 

Otherwise, this all LGTM

pkg/hs/urbit-king/lib/Urbit/Arvo/Effect.hs
pilfer-pandex: That's right. Please do so.

pkg/hs/urbit-king/lib/Urbit/Arvo/Effect.hs
```diff
@@ -174,19 +175,22 @@ instance FromNoun Deco where
...
+    TintR          -> toNoun $ Cord "r"
+    TintG          -> toNoun $ Cord "g"
+    TintB          -> toNoun $ Cord "b"
+    TintC          -> toNoun $ Cord "c"
+    TintM          -> toNoun $ Cord "m"
+    TintY          -> toNoun $ Cord "y"
+    TintK          -> toNoun $ Cord "k"
+    TintW          -> toNoun $ Cord "w"
+    TintNull       -> Atom 0
+    TintTrue r g b -> Cell (Atom r) $ Cell (Atom g) (Atom b)
```
pilfer-pandex, line 187: If you like, you can use C and A instead of Cell and Atom here.

pkg/hs/urbit-king/lib/Urbit/Vere/Term.hs
```diff
@@ -417,10 +420,10 @@ localClient doneSignal = fst <$> mkRAcquire start stop
           [ intersperse ';' $ fmap termRenderDeco $ toList decoset
           , case back of
               TintNull -> []
-              tint     -> ['4', termRenderTint tint]
+              tint     -> cons '4' $ termRenderTint tint
```
pilfer-pandex, line 423:
```
'4' : termRenderTint tint

File: pkg/hs/urbit-king/lib/Urbit/Vere/Term.hs [hunk 4]
```diff
@@ -391,17 +391,20 @@ localClient doneSignal = fst <$> mkRAcquire start stop
...
+      TintG    -> ['2']
+      TintY    -> ['3']
+      TintB    -> ['4']
+      TintM    -> ['5']
+      TintC    -> ['6']
+      TintW    -> ['7']
+      TintNull -> ['9']
+      TintTrue r g b ->
+        mconcat ["8;2;", bite r, ";", bite g, ";", bite b]
+        where bite = show . flip mod 256
```
pilfer-pandex, line 407:
```
show . (`mod` 256)

pilfer-pandex (changes requested):
Word8 plus minor style suggestions

pkg/hs/urbit-king/lib/Urbit/Arvo/Effect.hs
ohAitch: pls tell me there's some way in haskell to say "TintR means `r`, TintG means `g`…" that doesn't use a line per character

ohAitch: I'm _assuming_ you can't just do a hoon/ ts-ish `Tint ("r" | "g" | "b" | "c" | "m" | "y" | "k" | "w")` type that would render the point moot?

pilfer-pandex: Unfortunately, no to both of these, honorable Anton.

pilfer-pandex (approved):

PR #4057, opened 2020-12-02
login: translates to indigo

tylershuster:
Fixes https://github.com/urbit/landscape/issues/154

There's one line commented out for pattern matching, because it doesn't get escaped properly. Otherwise, take a look!

@urcades @matildepark @Fang- 

Screen Shot 2020-12-01 at 8 24 11 PM
Screen Shot 2020-12-01 at 8 24 04 PM
Screen Shot 2020-12-01 at 8 23 53 PM
Screen Shot 2020-12-01 at 8 23 43 PM
Screen Shot 2020-12-01 at 8 22 14 PM
Screen Shot 2020-12-01 at 8 22 11 PM
Screen Shot 2020-12-01 at 8 22 05 PM


tylershuster:
Updated the disabled input to:

Screen Shot 2020-12-01 at 8 35 12 PM

matildepark:
I guess this should target ... `na-release/candidate`? It's sys files but you want this in the breach, right? @philipcmonk ?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -254,167 +254,152 @@
...
       ==
     ==
+    ;script:'''
+            var failSpan = document.querySelector('.failed');
+            if (failSpan) {
+              document.querySelector("input[type=password]").addEventListener('keyup', function (event) {
+                failSpan.style.display = 'none';
+              });
+            }
+            '''
```
matildepark, line 402: Looks like it still needs the "Access Bridge" etc. links at bottom?

tylershuster: https://github.com/urbit/landscape/issues/154#issuecomment-733989711

> Note that in the latest rev, the bottom links have been removed

matildepark: Welp you got me

urcades (approved):
Beautiful

tylershuster:
Will rebase when pcmonk chimes in

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -254,167 +254,152 @@
...
+      ;form(action "/~/login", method "post", enctype "application/x-www-form-urlencoded")
+        ;input
+          =type  "password"
+          =name  "password"
+          =placeholder  "sampel-ticlyt-migfun-falmel"
+          =class  "mono"
+          =required  "true"
+          =minlength  "27"
+          =maxlength  "27"
+          :: =pattern  "((\w{6})-?){4}"
```
Fang-, line 379: This can probably go?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -254,167 +254,152 @@
...
+          =autofocus  "true";
+        ;input(type "hidden", name "redirect", value redirect-str);
+        ;+  ?:  failed
+          ;span.failed
```
Fang-, line 383: I'd use a `?.` to put the shorter, `;span;` case first. I think you can just put `~` for that, too, instead of an empty span.

Fang- (approved):
Neat, CSS vars. This can be a little bit cleaner, but looks fine otherwise.

pkg/arvo/sys/vane/eyre.hoon
tylershuster: Well, I was hoping to do pattern matching on the code, but this fails parsing because of the () and {}

Fang-: You should be able to just `\{`?

pkg/arvo/sys/vane/eyre.hoon
tylershuster: One would hope you could sig it out but alas

pkg/arvo/sys/vane/eyre.hoon
tylershuster: Alas, alack, I report a nack

philipcmonk:
Yeah, na-release/candidate is good for the breach.  As always, merge instead of rebase if it's also going on another branch or you'll create duplicates.  If its ancestors shouldn't go on na-release/candidate now, then you can rebase onto master, then merge from that to whatever you want since everything is up-to-date with master.

pkg/arvo/sys/vane/eyre.hoon
Fang-: Have you tried: `"((\\w\{6})-?)\{4}";`?

```hoon
> ;input
  =pattern  "((\\w\{6})-?)\{4}";
[[%input [%pattern "((\\w\{6})-?)\{4}"] ~] ~]
```

pkg/arvo/sys/vane/eyre.hoon
Fang-: Huh. What kind of error do you get for that? It works from dojo, but maybe that's naive/missing some relevant context.

```hoon
> ;+  ?.  |  ~
      ;span.failed  "lmao";
  ==
[~ ~]
```

pkg/arvo/sys/vane/eyre.hoon
lycurgus: Can I humbly suggest `((?:[a-z]{6}-){3}(?:[a-z]{6}))` as the regex? It's a bit clunkier, but the existing one will match e.g. `sampelsampelsampelsampel` because the `-` is always optional. Gratuitous non-capture groups added for good measure (so we get a whole capture or nothing at all - not sure if that makes a difference for HTML validation anyway).

Also switched out `\w` for `[a-z]` because `\w` includes 0-9 which I imagine we don't want to consider valid in a code?

PR #4058, opened 2020-12-02
chat: graph-store

liam-fitzgerald:
Graphication of chat-store. Included (for no extra charge):
- VirtualScroller now runs on a `BigIntOrderedMap`.
- Native mentions instead of scanning messages individually
- Makes graph-store JSON conversions roughly invertible i.e. `(dejs(enjs(x)) ~= x` (still need to strip leading sigs from @ps)

- Uses this invertibility to standardise pending messages across all apps
- Adds a variety of scries to `graph-store` to accomodate chat access patterns
- Reworks hark to count unreads!!
- Unread counts are kept for comments, links, notes and messages.
- Unreads are tracked as a `(set index)` for links and notes (i.e. you can read a note without reading the others)
- Unreads are tracked as an `@ud` for comments and messages (i.e. reading will read all)
- Last updated is kept correctly for graph resources
#### Group tiles now show unread counts in addition to the number of notifications
Capture d’écran 2020-12-02 à 3 42 32 PM
unread comments on a note
Capture d’écran 2020-12-02 à 3 35 25 PM
unread link
Capture d’écran 2020-12-02 à 3 35 00 PM

Still todo: 
- [x] tombstone chat-store & friends
- [x] Finalize migration logic

cc: @tylershuster please look at new virtualScroller
@urcades


matildepark:
Holy diff. Very big

pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
```diff
@@ -256,27 +269,12 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
...
-      .sort((a, b) => a.when - b.when)
-      .forEach((message, index) => {
-        const idx = bigInt(index + 1); // To 1-index it
-        messages.set(bigInt(mailboxSize).add(idx), message);
-        lastMessage = mailboxSize + index;
-      });
-
     const messageProps = { association, group, contacts, hideAvatars, hideNicknames, remoteContentPolicy, unreadMarkerRef, history, api };
 
+    const keys = graph.keys().reverse();
```
tacryt-socryp, line 268: Why do you need to do this?

pkg/interface/src/views/apps/chat/ChatResource.tsx
```diff
@@ -66,7 +66,7 @@ export function ChatResource(props: ChatResourceProps) {
   const chatInput = useRef<ChatInput>();
 
   useEffect(() => {
-    props.api.graph.getGraph(owner,name);
+    props.api.graph.getNewest(owner, name, 10);
```
tacryt-socryp: Why only the newest 10? Surely we could handle a hundred or so?

matildepark: Looks like it got increased [here](https://github.com/urbit/urbit/pull/4058/files#diff-7c7cc8a40d849e76929b674f5a1020d03a7d7742caa3eaa53b2df1564b67c799R44-R49)?

pkg/interface/src/logic/subscription/global.ts
```diff
@@ -54,7 +54,7 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
     this.subscribe('/updates', 'hark-store');
     this.subscribe('/updates', 'hark-graph-hook');
     this.subscribe('/updates', 'hark-group-hook');
-    this.subscribe('/updates', 'hark-chat-hook');
+    this.startApp('chat');
```
tacryt-socryp, line 57: How did this sneak back in?

pkg/arvo/app/hark-graph-hook.hoon
```diff
@@ -59,9 +60,21 @@
...
   ^-  (quip card _this)
   :_  this(state !<(state-0 old))
-  ?:  (~(has by wex.bowl) [/graph our.bowl %graph-store])
-    ~
-  ~[watch-graph:ha]
+  %+  welp
+    ?:  (~(has by wex.bowl) [/graph our.bowl %graph-store])
+      ~
+    ~[watch-graph:ha]
+  %+  turn
```
tacryt-socryp, line 67: Perhaps graph-store should keep track of a list of validator marks currently in use, and provide subscription updates when that list changes? Would remove the need to hard code this sort of list. (Your use case of knowing when the marks change is righteous, though, certainly, and this is not a blocker, just a suggestion for the future)

pkg/arvo/lib/graph-store.hoon
```diff
@@ -198,14 +198,17 @@
...
       |=  [a=atom n=^node]
-      ^-  json
-      :-  %a
-      :~  (index [a]~)
-          (node n)
-      ==
+      ^-  [@t json]
+      :_  (node n)
+      =/  idx
+        (numb a)
```
tacryt-socryp, line 208: Such a short statement, surely it can go on the prior line...

pkg/arvo/lib/graph-store.hoon
```diff
@@ -222,14 +225,15 @@
...
       |=  [n=^index o=^node]
-      ^-  json
-      :-  %a
-      :~  (index n)
-          (node o)
-      ==
+      ^-  [@t json]
+      :_  (node o)
+      =/  idx
+        (index n)
```
tacryt-socryp, line 234: Same.

pkg/arvo/lib/graph-store.hoon
```diff
@@ -310,12 +314,12 @@
...
-      :~  [%empty ul]
-          [%graph graph]
-      ==
-    ::
+      |=  jon=json
+      ^-  ^internal-graph
+      ?~  jon
+        [%empty ~]
+      [%graph (graph jon)]
+          ::
```
tacryt-socryp, line 322: Why the spacing?

pkg/interface/src/views/apps/chat/ChatResource.tsx
tacryt-socryp: Gotcha. I went through all the commits individually to watch it evolve.

tacryt-socryp:
Going to test this now on a fakezod and fakebus.

matildepark:
Can this be tested yet? The migration isn't in this PR as of yet.

liam-fitzgerald:
btw @tacryt-socryp your IconIndex issue should be fixed with `rm -rf node_modules && npm i`

liam-fitzgerald:
(i've pushed a working glob if you can't get it to work for whatever reason)

pkg/arvo/app/chat-store.hoon
```diff
@@ -72,6 +81,74 @@
     ::
     ==
     ::
+    ++  nobody
+      ^-  @p
+      (bex 128)
```
tacryt-socryp: What made you decide on making a "null ship" vs discarding the data? Just curious, not strongly opinionated on the matter.

tacryt-socryp (approved):
LGTM, pending testing. The migration is very straightforward.

tacryt-socryp:
My problem has not been resolved by clearing Safari's cache, removing node_modules, reinstalling them with npm, and restarting the dev server. It is also present on localhost:80 as well as localhost:9000. Not sure why you aren't seeing it in your browser.
Screen Shot 2020-12-03 at 12 05 53 PM

matildepark:
I’m gonna look into this and hot fix next userspace if necessary, will merge it in

tacryt-socryp:
Worth noting: I'm on Safari.

matildepark:
The indigo-bump from 1.2.14 -> 1.2.15 had a typo that bumped it to 1.2.5, which had way fewer icons. Yes, it was really that dumb. This branch should work now.

pkg/arvo/app/chat-store.hoon
liam-fitzgerald: I'm just incredibly hesitant to discard data, given the amount of data that's generally in chats

matildepark: I think this hesitation is generally correct — discarding user data is essentially forbidden as per previous discussions, so we should really be eager to instead dump malformed data and provide the means by which it can be amended, re-imported or just otherwise archived.

matildepark:
Some conflicts here now that na-release/candidate is merged into release/next-userspace. Might want to merge userspace back in here to reflect that

tacryt-socryp:
I resolved the merge conflicts, but this can't be merged until a fix has been made to the virtualized kick handler. It now crashes when compiling that code, and I'm unsure of why. Investigating, but I don't know how much progress I'll make.

matildepark:
If this affects the candidate, can you PR any fix to the candidate in turn? We can just merge it back into userspace.

tacryt-socryp:
No, it comes from release/next-userspace but only appears now that the candidate has merged in.

matildepark:
But next userspace was merged into the candidate last night.

tacryt-socryp:
Fixed it!

tacryt-socryp:
Note: Making a new chat DM on this branch is broken.

joemfb:
Blocking semantics in `+mock` changed in these breach, see https://github.com/urbit/urbit/pull/4080#discussion_r536378863.

matildepark:
> Note: Making a new chat DM on this branch is broken.

Duplicate logic that hasn't been updated — 

https://github.com/urbit/urbit/blob/777b780d0567a6c656da3e5c057bf35c2916b823/pkg/interface/src/views/landscape/index.tsx#L33-L66

If no one gets to this I'll try to; I'm gonna test regular channel creation a bit with a few ships first ...

matildepark:
Hark may not be updated for this graph yet? Or something? Issue one is that the content is blank, issue two is that the chat is muted and shouldn't be delivering notifications

image

pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
liam-fitzgerald: so we can get the next and previous messages

matildepark:
Set up a DM and the hamburger menu on both ships says "Unmute this channel," so it must be muted — but all messages are sent to the notifications inbox.

matildepark:
These are being rendered sorta weird in an aesthetic way. Single string becomes one character per line

image

tacryt-socryp:
I have no clue how to do this `%chat-hook` merge. How are we going to handle `+poke-import` (flagday imports from the last era) when we are also migrating out of this whole app to `%graph-store`? Wasn't there some talk of "flag day must only move them to the first pill of the era" or something, such that we could remove the import code from the chat apps?

matildepark:
Wouldn't it migrate versioned state that would then be prompted to do the graph-store migration afterward?

tacryt-socryp:
It migrates to state-10, whereas the graph-store move occurs at state-11, but I don't think the poke-import would even work if the source code was already at state-11, given type checking.

philipcmonk:
I mean, it's all just nouns, right?  Lens will turn it into the state that chat-store has now, we just need to pass that to whoever knows how to install it into graph-store.  Remember we can change lens to recognize it's chat-store and do whatever it wants.  We all have `./tmp-flagday/archive.jam` files with "test" data which you can `-i` into a fake ship or `-L` copy.

tacryt-socryp:
Well, we're discussing the `+poke-import` of `%chat-hook` itself, which spits out a "state-10" and trying to figure out whether we can expect that they'll be upgrading into a state that starts at the beginning of the era, or whether we need to support upgrading the state both from `+poke-import` and from `+on-load`, which seems annoying.

philipcmonk:
I would really like to avoid saying that importing only works with a particular pill.  You're very familiar with the problem of people booting with an old pill -- you'll have a bunch of people boot into a broken ship that can't even get to urbit-help until they get an upgrade from their sponsor which probably hasn't migrated or was never online in the first place.

I really think we need to support either case.  Is it much harder than having +poke-import end by calling +on-load?  You might be able to get rid of a few watches or other unnecessary moves, but also maybe you don't need to.  Sure, you have to test both paths, but we're going to have both paths in real life anyway and they will both need to be tested.

liam-fitzgerald:
cc: @tacryt-socryp can I get a pass on the changes to the migration logic?

matildepark:
Crash on accepting an invite — 

image

ps:
@ps chiming in since I was tagged, don't have the bandwidth to go through this, looks like @matildepark has it covered

tacryt-socryp:
On the moon testnet, when I click on a chat to join it within your group, I cannot, as it shows a loading indicator instead of showing the "Join blah blah blah" dialogue.

urcades:
The issue Logan is noting is the one I continue to have, and had originally
reported.

—
~fabled-faster
https://urbit.org/

urcades:
Something I hadn't considered trying: Making my own group!

Please see if you can join my group: ~litwyn-binpes-tonlur-hidlyn/lunarians

—
~fabled-faster
https://urbit.org/

pkg/arvo/app/hark-store.hoon
```diff
@@ -84,9 +84,64 @@
...
+    |=  [=index:state-zero:store =notification:state-zero:store]
+    ^-  (unit [index:store notification:store])
+    =/  new-index=(unit index:store)
+      (convert-index-1 index)
+    =/  new-notification=(unit notification:store)
+      (convert-notification-1 notification)
+    ?~  new-index  ~
+    ?~  new-notification  ~
+    `[u.new-index u.new-notification]
+
```
tacryt-socryp, line 125: use a linter, blank lines are against the style guide

tacryt-socryp (approved):
Migration of `%hark-chat-hook` looks good.

pkg/arvo/app/graph-store.hoon
```diff
@@ -257,11 +257,17 @@
               ?&  !(~(has by archive) resource)
                   !(~(has by graphs) resource)
           ==  ==
+      =/  old-graph=(unit marked-graph:store)
```
tacryt-socryp, line 260: This is the wrong behavior. We cannot both have an archive of an old graph of a name AND have an active graph of the same name. Overwrite does what it says on the box. It overwrites the data.

tacryt-socryp (changes requested):

matildepark:
@urcades @tacryt-socryp's join error seems to be that I never ran `|public %home` on my moon, which I Forgot. User error in this case.

matildepark (approved):
LGTM after many edge case attempts on testnet (flagday straight into chat migration; DM deletion and recreation; massive backlog virtualisation).

PR #4060, opened 2020-12-02
hoon: updates +end, +lsh, and +rsh to use $bite

joemfb:
This PR extends that pattern from #4049 to the aforementioned jets -- atom gets that take both a `$bloq` and a `$step` now take a `$bite`, supporting an implicit default `$step` of one.


Fang- (approved):
Righteous change!

pkg/arvo/app/chat-cli.hoon
```diff
@@ -767,7 +767,7 @@
       :-  %chat-view-action
       !>  ^-  action:view
       :*  %create
-          (rsh 3 1 (spat path))
+          (rsh 3 (spat path))
```
ohAitch, line 770: Hmm, this has the unfortunate appearance of "rightshift by 3", possibly we want this just for `end`? Idk if you're sure of it put it in, I just notice that almost half the calls _do_ use both arguments, and most that don't are made less clear by removing `1`. (Granted a lot of them are right next to a matching `end`, making this case particularly bad.)

joemfb: I believe you answered this objection already in https://github.com/urbit/urbit/pull/4049#issuecomment-736851311:

> an argument can be made for not having bloq at all and phrasing all atom manipulations using a width in bits; but in practice, 32 is tremendously a more common bitwidth than 48 or particularly 47, and warrants semantic compression on that basis. Having to specify 8 as 3 1 blows up that compression.

ohAitch:
Ooh nice I've hit the fil thing before

pkg/arvo/app/chat-cli.hoon
ohAitch: If we keep shifting by 2 or 3 or 5 blocks, it no longer qualifies for "tremendously more common" :P
But yeah it's like fine.

ohAitch: (Esp if we keep shifting by like, `a`. My objection to `3 1` for rip/run/can is if it's 3 it's literally always 1, my softer objection for `end` is `end 3 1` was more common than all other `end` put together)

PR #4065, opened 2020-12-02
arvo: remove unused app files, libraries, and imports

tacryt-socryp:
We no longer use the following apps, so this PR removes them for the upcoming breach:
- group-hook
- link-listen-hook
- link-proxy-hook
- link-store
- link-view
- permission-group-hook
- permission-hook
- permission-store
- publish
- soto

And a whole host of associated mar files, lib files, gen files, and sur files.


Fang- (approved):
rip all of this code.

The removals here seem fine to me.

philipcmonk (approved):
Nice!

PR #4067, opened 2020-12-03
MAINTAINERS.md: add instruction for latest vere

matildepark:
By adding this instruction to our release flow, we can easily link to the latest Vere programmatically.

(I have already done this with v.0.10.8.)


philipcmonk (approved):
Yeah, I think this is good.

PR #4070, opened 2020-12-03
AsyncButton, StatelessAsyncs: protect against double submit

liam-fitzgerald:
Disables AsyncButton if the form is submitting, to prevent accidental double submits.

Supersedes #3877


matildepark:
I'm starting to feel averse to merging PRs that make us dependent on an ephemeral branch of indigo-react — would rather hold these until Indigo bumps

liam-fitzgerald:
no I don't want to merge these until the indigo changes are merged + published. It's just easier to do the PRs in lockstep.

Issue #4059, opened 2020-12-02
Copying @p in chat drops the “~”


timlucmiptev:
When I click a username in chat to copy the @p, the leading sig isn’t included. This makes it awkward to use the new notification feature to mention people.

matildepark:
I think what we really want is the return of the tab suggestions ...

PR #4074, opened 2020-12-04
release/next-userspace -> na-release/candidate

matildepark:
Merging next-userspace into the post-breach release. Please don't merge pre-breach OTA work into next-userspace.

Applicable changes for @Fang- release notes:

- Fixes an issue where crashes in the pull-hook could create a situation where the subscription is abandoned while the resource persists, never updating or resubscribing.
- Fixes an issue in Landscape where being mentioned in a chat while viewing that chat would not dismiss the notification, persisting until next refresh.
- Fixes an issue where metadata would not be deleted upon leaving a group, dumping all channels into "DMs + Drafts".
- Fixes an issue where users would stay subscribed to a graph resource after leaving a group.
- Aqua stores Azimuth state internally rather than relying on a thread.
- Channel.js now batches subscription requests, leading to a major performance improvement, especially on mobile devices, for initial load time.
- Kicking a user from a group now properly kicks them from the group and all graph resources, instead of silently leaving a ghost version of the group on their ship.

And all fixes as per my QA cleanup PR:

> * Adds 'cancel' button when editing a note. If the field is dirty, MarkdownField still prevents navigation.
> 
> * Adds a leading sig when copying patps from Chat (#4059)
> * Mobile input for Chat overflows vertically by using textarea instead of input.
> * Chat's Daybreak marker is padded evenly with other chat messages, removing `pt='3'` since `pt='3'` is already inherited by the renderSigil component.
> * Admins can no longer ban themselves.
> * Admins can no longer demote themselves if they are the only admin in the group.
> * Invite popover now has some margin on mobile devices.
> * Amends the 'Back' link size on mobile views of our root identity card to match current Indigo.
> * Profile now has icons, and its sidebar matches the styling of other sidebars (gray background, not blue).
> * Fixes an issue where if you set an avatar while sharing your contact for the first time — or if your root identity had an avatar set — you couldn't ever share your contact card. This is because contact-json wants `{ url: avatar }` but we were passing the URL as a straight string; whereas we caught that if we were _editing_ the card. Additionally our field should've passed blank strings to the backend as `null` for avatars, but it wasn't, so if you wiped the input field to try and fix it, it would still fail. (#4064)
> * If you set an avatar or nickname in a group or on your root identity, the contact card field now reflects that choice (if your ship settings allow it).


philipcmonk:
Do you want next-userspace to go out with the pre-breach OTA as well?

matildepark:
I don't. It can all just go out post-breach.

philipcmonk (approved):
LGTM

Issue #4015, opened 2020-11-25
Assertion 'c3y == u3a_north_is_normal(u3R, nov)' failed in noun/allocate.c:1161


lrettig:
I just encountered an error very similar to one that @joemfb described in #3928 as concerning:

```
[%mailbox-kick i=%~.mailbox t=[i=~.~dopzod t=/urbit-help]]
%chat-hook-resubscribe
; ~locref-somnyd not responding still trying
Assertion 'c3y == u3a_north_is_normal(u3R, nov)' failed in noun/allocate.c:1161

bail: oops
bailing out
pier: serf unexpectedly shut down
```

`bail:meme` can be a transient error, if computing some event pushes you to the memory threshold, but doesn't affect persistent state. `bail:meme` is always treated as a fatal error for a bunch of contingent implementation reasons that need to be fixed, to cover this scenario.

If `bail:meme` happens repeatedly, or on restart, the only solution is `|pack` or `|meld` (either in dojo, or directly through the `urbit-worker`).

The other error you posted is much more concerning:

```
Assertion 'c3y == (( c3y == nor_o) ? u3a_north_is_normal(u3R, nov) : u3a_south_is_normal(u3R, nov))' failed noun/allocate.c:1190
```

Outside of a couple recent reports, I've never seen this happen. That's a sanity check when promoting values from inner roads -- it failing either represents significant programmer error (somehow latent until now) or memory corruption.

`|pack` may help with some kinds of memory corruption, but it mostly just moves existing things around. `|meld`, if it works, is much more likely to help, as it reallocates everything. The best option we have, again, if the corruption is not so significant as to prevent it from running, is taking and loading a jammed snapshot (`cram` and `queu` in `urbit-worker` v0.10.9, run without arguments to see usage details).

If the corruption prevents these recovery tools from running, the only option is to replay the log (perhaps first restoring from an earlier, known-good snapshot).

_Originally posted by @joemfb in https://github.com/urbit/urbit/issues/3928#issuecomment-731375238_

vvisigoth:
huh. @philipcmonk @joemfb any insights here?

lrettig:
Forgot this one:

```
> +trouble
[ [%base-hash ~[0v7.t16i9.avhp4.pflq0.1c79d.rplo0.ho0i2.9ovj1.kddvj.e1snn.0kb4n]]
[...]

Issue #1224, opened 2019-03-13
stack smashing on Ctrl-C


belisarius222:
My planet has been almost entirely unresponsive since I booted it yesterday, so I finally Ctrl-C'd it to see what would happen. This is the result:

```
~rovnys-ricfer:talk[] |
recover: dig: intr
address 0xe265567c out of loom!
*** stack smashing detected ***: <unknown> terminated
```

Stack trace from pulling up the core dump:
```
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51      ../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7fb0f3c90800 (LWP 12557))]
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007fb0f24aa801 in __GI_abort () at abort.c:79
#2  0x00007fb0f24f3897 in __libc_message (action=action@entry=do_abort,
    fmt=fmt@entry=0x7fb0f2620988 "*** %s ***: %s terminated\n")
    at ../sysdeps/posix/libc_fatal.c:181
#3  0x00007fb0f259ecd1 in __GI___fortify_fail_abort (
    need_backtrace=need_backtrace@entry=false,
    msg=msg@entry=0x7fb0f2620966 "stack smashing detected")
    at fortify_fail.c:33
#4  0x00007fb0f259ec92 in __stack_chk_fail () at stack_chk_fail.c:29
#5  0x1c433e792893c900 in ?? ()
#6  0x0000000000000007 in ?? ()
#7  0x000000003a15e6b0 in ?? ()
#8  0x0000000000e5c6c2 in ?? ()
#9  0x0000000001e73763 in ?? ()
#10 0x0000000039971b08 in ?? ()
#11 0x000000003d9cdd8c in ?? ()
#12 0x0000000000040010 in ?? ()
#13 <signal handler called>
#14 0x000000003657f570 in ?? ()
#15 0x0000000100000006 in ?? ()
#16 0xc259baa200000000 in ?? ()
#17 0x0000000600afa82c in ?? ()
#18 0x0000000000000018 in ?? ()
#19 0x01798b63020e8b63 in ?? ()
#20 0x01fd91b9025352f6 in ?? ()
#21 0x0000000e01fa3ebc in ?? ()
[...]

PR #4076, opened 2020-12-04
u3: disable +scot and +scow jets

joemfb:
Disabling these jets (from #2853) resolved memory corruption issues (in the king process) on the testnet. More debugging will be needed to find the root cause (test coverage remains wanting, yet highly desirable). Disabling for now ...

### DIFF
--- a/pkg/urbit/jets/tree.c
+++ b/pkg/urbit/jets/tree.c
@@ -944,8 +944,11 @@ static u3j_core _141_qua_d[] =
   { "mole", 7, _141_qua_mole_a, 0, _141_qua_mole_ha },
   { "mule", 7, _141_qua_mule_a, 0, _141_qua_mule_ha },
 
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):

ohAitch:
\<small>
&nbsp; the only concrete parsing jet we should have is `cue`, everything else should be fixed through a better memory model / datastructures
\</small>

joemfb:
leans into the mike ... *No*

joemfb:
I should note that the corruption symptoms here matched both #1224 and #4015.

ohAitch:
Oh wait these are both a *printer* not a parser, then I'm just confused yeah. I'm assuming running these under asan was not illuminating?

joemfb:
Surprisingly, no. But some of the crashes were showing up in the road/bail setjmp/longjmp, and idk how well ASan can track that sort of thing.

PR #4078, opened 2020-12-04
vere: only compare to clay when committing files

### LINKED ISSUE #3240: clay: Compilation triggered by |commit misses some new files if previous |commit resulted in build failure
**Describe the bug**

Compilation triggered by commit seems to miss files that were changed in a previous commit that had a compilation fail.  See steps below:

**To Reproduce**

1. Add an extra note `[%another-case ~]` to eyre notes in zuse.hoon
2. `|commit`, this will result in a `mint-lost` failure, as expected.
3. Handle the new note in eyre.hoon (just do whatever %vega does or !!).
4. `|commit`, this will fail with `mint-vain` as if the new case in zuse did not exist

**Expected behaviour**

In step 4, I've expected it to compile correctly.

A workaround is to change BOTH files, that will force both to be recompiled, resulting in correct compilation.

**Additional context**

Making a new change to zuse, and doing another |commit results in the first `mint-lost` error, which suggests the old eyre source is being used.

It seems Clay is taking only the latest committed changes plus the latest correctly built state.

Fang-:
Instead of first comparing to a previously committed version of the
file, which may or may not have been fully accepted by clay.

Fixes #3240, though some of the discussion therein remains relevant.

### DIFF
--- a/pkg/urbit/vere/io/unix.c
+++ b/pkg/urbit/vere/io/unix.c
@@ -35,7 +35,6 @@ struct _u3_ufil;
     c3_c*             pax_c;            //  absolute path
     struct _u3_udir*  par_u;            //  parent
     struct _u3_unod*  nex_u;            //  internal list
-    c3_w              mug_w;            //  mug of last %into
     c3_w              gum_w;            //  mug of last %ergo
   } u3_ufil;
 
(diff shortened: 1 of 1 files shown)


joemfb:
This change itself is not exceedingly urgent, it can be easily made in a post-breach runtime release. So I'm probably going to hold off on loading the problem into my head for now ...  @philipcmonk is probably better situated for a quick review.

philipcmonk:
@Fang- please re-open again release/next-vere, looks like this was autoclosed by accident and I can't seem to rebase it

joemfb:
My bad, I forgot there were PRs still targeting this branch.

Fang-:
Gentle ping, either @philipcmonk or @joemfb for review.

philipcmonk (approved):
Looks correct

joemfb (approved):
LGTM

Sorry for the delay! I meant to read this file and be sure I understood it better in general. But this change is nicely isolated, only implicates a couple places, and seems obviously better.

PR #4080, opened 2020-12-04
pull-hook: fix handle-kick printouts

tacryt-socryp:
For some reason, on na-release/candidate, `lib/pull-hook` fails to compile. This addresses that issue.

### DIFF
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -332,7 +332,7 @@
         p.u.cas
       now.bowl
     ::  catch bad gall scries early
-    ?:  ?&  =((end 3 1 i.u.pax) %g)
+    ?:  ?&  =((end 3 i.u.pax) %g)
             ?|  !=(`our.bowl ship)
                 !=(dat now.bowl)
             ==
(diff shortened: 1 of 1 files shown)


pkg/arvo/lib/pull-hook.hoon
```diff
@@ -349,13 +350,16 @@
...
         :+  leaf+"failed kick handler, please report" 
           leaf+"{<rid>} in {(trip dap.bowl)}"
         ?:  ?=(%2 -.res)
           p.res
         ?>  ?=(%1 -.res)
-        (turn `(list *)`p.res (cork path smyt))
+        =/  paths=(unit (list path))  ((soft (list path)) p.res)
+        ?~  paths  ~
+        %+  turn  u.paths
+        (cork path smyt)
```
joemfb, line 362: This is no longer an untyped list of paths, it's a single untyped path (see `$toon`).

joemfb (changes requested):

pkg/arvo/lib/pull-hook.hoon
tacryt-socryp: Latest commit addresses that. LMK what you think.

joemfb (approved):
LGTM

philipcmonk (approved):

PR #4082, opened 2020-12-04
leap: position notification dot relatively

matildepark:
After indigo-react 1.2.15, Box no longer has `position:relative` by default.

While we caught this elsewhere, we dropped this in Leap results, aligning the
notification dot to the top left of the browser window instead.

Targeting na-release/candidate as it'll inevitably get merged downstream into next-userspace and this is otherwise a regression.


matildepark:
ping @tacryt-socryp @liam-fitzgerald

liam-fitzgerald (approved):

PR #4083, opened 2020-12-05
vere: report, not crash, on unreasonable -Y and -Z

Fang-:
Previously, we would crash on attempting to `+scot` cells, or parse
invalid paths. Now we catch those cases, and print legible error
messages instead.


joemfb (approved):
LGTM

PR #3636, opened 2020-10-02
hoon: comments, refactors $coin rendering, minor efficiency improvements

joemfb:
Spurred by a suggestion of @Fang-'s in #3576, I needed to reacquaint myself with the various helper cores in `|co`, the `$coin` renderer. I've found it hard to remember how those helpers work, so I went through them and added detailed comments.

Having read them closely, I noticed various inefficiencies (separate calls to `+div` and `+mod` instead of `+dvr`, a handrolled `+pow` implementation, &c). Those matter, as we run this code quite frequently. Once I made those changes I decided to modernize the surrounding syntax as well. Once I did that, I couldn't prevent myself from furthering refactoring parts of the implementation. Most of the changes are just mechanical reorganization, using new runes (like `=/` and `=?`).

Please review for style and comment clarity.

PR #4084, opened 2020-12-05
hoon: fixes logic bug in date printer (truncation)

joemfb:
Fixes a terrible bug i introduced into `@da` coin printing in #3636, wherein timestamps with *either* zero seconds/minutes/hours were printed as if all of their seconds/minutes/hours were zero (and truncated). Flagged by @philipcmonk.

Also fixes some calls to `+end` and +rsh` (api changed in #4060) that weren't updated when merged (in #4074).


philipcmonk (approved):
Tested and works for me

PR #4085, opened 2020-12-05
jael: don't signal breach if first hearing of ship

philipcmonk:
When you first boot, if you try talk to someone before your azimuth is
up-to-date (for example by import), then if they've ever breached
(twice) then you'll get a breach notification, cancelling your message.

This changes is it so that if we haven't heard anything about this ship,
we don't signal a breach.

The implementation complexity is primarily because we need
eth-watcher/azimuth-tracker to produce an update of a list instead of a
list of updates.  This way, Jael can keep a "state as of the beginning
of this move" variable to check when deciding whether to signal a
breach.

Ignore the changes to lib/test-fake-ship.nix and lib/aqua-azimuth, they're actually from another commit that already went through review.


pkg/arvo/sys/vane/jael.hoon
```diff
@@ -97,7 +97,7 @@
   $%  $>(%init vane-task)                               ::  report install
   ==  ==  ==                                            ::
 ::                                                      ::
-+$  peer-sign  [=ship =udiff:point]                     ::
++$  peer-sign  (list [=ship =udiff:point])              ::
```
Fang-, line 100:
```suggestion
+$  peer-sign  udiffs:point                             ::
```

pkg/arvo/sys/vane/jael.hoon
```diff
@@ -716,13 +716,19 @@
...
+    =/  original-pos  pos.zim.pki
+    |-  ^+  this-su
+    ?~  udiffs
       this-su
-    (public-keys:feel %diff a-ship u.a-diff)
+    =/  a-point=point  (~(gut by pos.zim.pki) ship.i.udiffs *point)
+    =/  a-diff=(unit diff:point)  (udiff-to-diff:point udiff.i.udiffs a-point)
+    =.  this-su
+      ?~  a-diff
+        this-su
```
Fang-, line 729:
```suggestion
    =?  this-su  ?=(^ a-diff)
```

pkg/arvo/sys/vane/jael.hoon
```diff
@@ -882,7 +889,8 @@
       ::  if changing rift upward, then signal a breach
       ::
       =?    ..feel
-          ?&  ?=(^ maybe-point)
+          ?&  (~(has by original) who)
+              ?=(^ maybe-point)
```
Fang-, line 893: May or may not want to update the comment (here and on 854) to point out the "and we have older keys for them" clause.

Fang- (commented):
A question and some nits, but nothing too serious.

pkg/arvo/app/azimuth-tracker.hoon
```diff
@@ -75,13 +75,15 @@
...
+  :-  [%give %fact ~[/] %azimuth-udiffs !>(udiffs)]
+  |-  ^-  (list card:agent:gall)
   ?~  udiffs
     ~
   =/  =path  /(scot %p ship.i.udiffs)
-  :*  [%give %fact ~[/] %azimuth-udiff !>(i.udiffs)]
-      [%give %fact ~[path] %azimuth-udiff !>(i.udiffs)]
-      $(udiffs t.udiffs)
-  ==
+  ::  Should really give all diffs involving each ship at the same time
```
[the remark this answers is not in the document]
philipcmonk: Yes, but also this branch is only used for moons right now, since it only applies if you're subscribed to a *particular* ship.

philipcmonk:
Good points, fixed the other stuff you pointed out

Fang- (approved):
Changes look good!

philipcmonk:
merged out of band (sorta... it's fine)

PR #4088, opened 2020-12-06
ames: don't crash on forward-lane scry

philipcmonk:
Ames crashed on this scry if it hadn't heard of the ship, so if you had five of them in a row it would give up on stateless forwarding.  This is possibly the reason for the occasional time we've seen forwarding revert on mainnet, but it's especially important in the new breach when the first ships will mostly be talking to ships that haven't migrated yet.


joemfb (approved):
Great catch! LGTM

Fang- (approved):
Concurred.

PR #4090, opened 2020-12-06
arvo: final cleanup, kelvins

joemfb:
This PR includes the following changes:

- updates pill structures to be discriminable
- generates ivory pills from userspace libraries
- persists vanes by full name
- addresses vanes by full name or abbreviation
- supports abbreviated or expanded scry perspectives (ex, %cx or ~.._clay_x__)
- replaces obsolete scry handlers and conversion
- restructures or removes legacy vane tasks (%init, %crud)
- removes the obsolete +come arm from the arvo interface
- moves vane strutures out of %zuse into %lull

It is complete, except for final kernel kelvin assignment.


pkg/arvo/sys/zuse.hoon
```diff
@@ -196,7 +196,7 @@
       [%crud p=@tas q=(list tank)]
       ::  boot completed (XX legacy)
       ::
-      [%init p=ship]
+      [%init ~]
```
Fang-, line 199: Does this mean the `XX legacy` qualifier can go?

pkg/arvo/sys/arvo.hoon
```diff
@@ -784,26 +869,21 @@
...
   ++  va
     =>  |%
-        ::  XX move next to +slap
-        ::
-        ++  swat
-          |=  [vax=vase gen=hoon]
-          ^-  (trap vase)
-          =/  gun  (~(mint ut p.vax) %noun gen)
-          |.([p.gun .*(q.vax q.gun)])
+        +$  vane-sample  [now=@da eny=@uvJ rof=rook]
```
Fang-, line 872: Curious what the reason is for passing `our` in through a top-level gate, as opposed to not having a top-level gate and just keeping it in this sample.

pkg/arvo/sys/arvo.hoon
```diff
@@ -1114,13 +1194,13 @@
...
         (peek:pith lyc %$ bem)
       ::
-      ::  XX vane and care are concatenated
+      ::  vane and care are concatenated
       ::
-      =/  lal  (end 3 cyr)
-      =/  ren  ;;(@t (rsh 3 cyr))
-      ?.  (~(has by van.mod) lal)
+      =/  way=term  (grow (end 3 cyr))
+      =/  car=term  (rsh 3 cyr)
```
Fang-, line 1200: Don't you want this to support `%clayx` style vane+care?

Fang- (commented):
All good changes so far! Some comments.

pkg/arvo/sys/zuse.hoon
joemfb: `XX somewhat legacy` ...

pkg/arvo/sys/arvo.hoon
joemfb: The main motivation is to only pass values that change on each invocation.

pkg/arvo/sys/arvo.hoon
joemfb: I'm not quite sure how to generically support all of `%$`, `%j`, `%cx`, %clayx`.

joemfb: Also, `%clayx` is *really* ugly.

joemfb: But, this current API doesn't support non-abbreviated names, which is a problem. Arguably, `(room)` should take `[gang route care beam]`, and this combined vane/care parsing should go into `+look`. I'll think about it.

pkg/arvo/gen/ivory.hoon
```diff
@@ -1,84 +1,33 @@
-::  Produce an ivory pill
-::
```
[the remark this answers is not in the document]
joemfb: I just wrote it elsewhere and then moved it -- I'll restore the comments.

pkg/arvo/sys/arvo.hoon
joemfb: Note that `|mass` is currently broken on this branch due to exactly this problem.

joemfb:
```
?=([~ %many [[%$ %tas @] [%$ %tas @] ~]] (rush ~.._ames_x__ ;~(pfix dot perd:so)))
```

Fang-: To be clear, I'm not wed to the `%clayx` format in particular, my concern is indeed about the non-abbreviated names that are now supported throughout the rest of arvo.

I think you _can_ discriminate between the term variants you gave, but you'd be making up janky-feeling restrictions like "only care-less scries using vane abbreviations". Using `%many` for this could be fun!

pkg/urbit/vere/pier.c
```diff
@@ -1584,64 +1584,88 @@ static u3_boot
...
+      exit(1);
+    }
 
-  if ( 0 != u3h(pro) ) {
-    fprintf(stderr, "boot: failed: unable to parse pill\r\n");
-    exit(1);
-  }
+    if ( c3y == u3r_sing_c("ivory", tag) ) {
+      fprintf(stderr, "boot: failed: unable to boot from ivory pill\r\n");
+      exit(1);
```
Fang-, line 1622: We now `u3_king_bail` prior to the other `exit` calls. Why do we not do that here?

pkg/arvo/sys/hoon.hoon
```diff
@@ -1905,17 +1905,17 @@
 ++  map
   |$  [key value]                                       ::  table
   $|  (tree (pair key value))
-  |=(a=(tree (pair)) ~(apt by a))
+  |=(a=(tree (pair)) ?:(=(~ a) & ~(apt by a)))
```
Fang-, line 1908: idk how this fixes the jet printfs, but seems a little bit wrong to fix those on the hoon side...

Fang-:
Scry reform reform looks good! Question about `pier.c` still stands, but it's pretty minor. Otherwise good to go in afaic, assuming @philipcmonk agrees.

pkg/arvo/sys/hoon.hoon
joemfb: This is a work around for the lack of hint pass-thru in the seminoun interpreter `|musk`. Calls to %fast-hinted functions from inside molds trigger the "parent not registered jet hint". And `|musk` would probably need to be changed significantly in order to pass %fast hints through.

pkg/urbit/vere/pier.c
joemfb: Good catch!

tacryt-socryp:
Your motion branch merged in a replacement of `%b` with `%behns` in chat-hook.hoon.

joemfb:
thanks, @tacryt-socryp, I'll push a fix.

joemfb:
@tacryt-socryp, chat-hook still fails to build on this branch (with `-find.action`).

philipcmonk:
also an 80 column violation on the imports...

philipcmonk:
For some reason `view` is coming from arvo instead of chat-view.  Presumably because one of the later imports doesn't `^?` its context

liam-fitzgerald:
dropping the `^?` in `sur/chat-view` will fix, but has the behavior changed?

philipcmonk:
Don't think that'll fix it, I think this branch must add +view to arvo

philipcmonk:
It's lib/migrate.  If you add a `^?` at the top it works.  Anything that's `*`-imported needs a `^?` or else it puts the kernel ahead in the subject

joemfb:
Yeah, that's my bad -- I thought I had grep'd for it. `$view` is the new concatenated-or-separated scry perspective: `$@(term [term term])`. I'll fix up the imports.

joemfb:
I hadn't refreshed the page, that's even better. I've confirmed fixing lib/migrate fixes the builds. Should I push the change here?

philipcmonk:
Sure

joemfb:
pulls in #1888

PR #4091, opened 2020-12-06
various: ethereum, azimuth, json-rpc out of zuse

Fang-:
Moves them into files in `/lib`, and updates dependents accordingly.

Other than renaming the existing `/lib/azimuth` to `azimuthio`, nothing too surprising.  
Arguably types should be plucked out of the ethereum & azimuth libs and moved into sur files, and maybe the ethereum lib should be split up into separate files for each of the main arms, but I'm fine with punting on all of that for now.

Based off & targets #4090, because that moved `+dawn` out of zuse into userspace, letting us pull the rest of these things out as well.


pkg/arvo/lib/vere.hoon
```diff
@@ -1,6 +1,7 @@
 ::  runtime support code
 ::
-=>  ..zuse  =>
+/+  ethereum, azimuth
+=>  [ethereum=ethereum azimuth=azimuth ..zuse]  =>
```
joemfb, line 4: This won't work, because of the way we build `lib/vere` in `+ivory`. I should really add a comment to the top noting that %ford runes aren't currently supported in `lib/vere`.

In the meantime, each of the relevant `|dawn` arms could just be modified to accept a contract address.

joemfb: It looks like it's just be `+czar:give`, `+point:give`, and `+turf:give`, I think it would be fine to just hardcode the contract address in the binary and pass it to those gates.

joemfb: Alternately, and more easily, you could keep just the addresses in %zuse.

Fang-: Right now zuse knows barely anything about azimuth anymore. Seems good to keep that as minimal as possible.

Since we need more than just the contract addresses, I ended up copying the relevant logic over from the libraries, directly into `/lib/vere`. Not the most elegant, but I'm assuming this is just a temporary solution.

joemfb: Oh, of course. This is probably fine for now (it's easy to update later). Or we could just hardcode the dependencies in `gen/ivory`.

pkg/arvo/gen/ivory.hoon
```diff
@@ -38,7 +40,35 @@
...
 ++  build-lib
-  |=  [sub=(trap vase) nam=term]  ^-  (trap vase)
+  |=  [sub=(trap vase) imp=? nam=term]  ^-  (trap vase)
   ~>  %slog.[0 leaf+"ivory: building /lib/{(trip nam)}"]
-  (swat sub (rain /lib/[nam]/hoon .^(@t cx+(welp lib /[nam]/hoon))))
+  =/  hun=hoon
+    %+  mist  /lib/[nam]/hoon
+    .^(@t cx+(welp lib /[nam]/hoon))
+  ?.  imp  (swat sub hun)
+  (swel sub [%ktts nam hun])
```
Fang-, line 49: @joemfb idk if this is strictly necessary or not, but using `+swel` for `/lib/vere` gave me a pill that caused the binary to crash.

Issue #4089, opened 2020-12-06
`make install` error


philipcmonk:
`make` works, but `make install` gives this error on Ubuntu, NixOS, and mac.

```
installing
post-installation fixup
patching script interpreter paths in /nix/store/mqiyh67czmisc856ib0j758pv582rz0w-urbit-debug-0.10.9-main-post
building '/nix/store/px7y2mc979bwpkbybh6j9fyi9icsdd9r-user-environment.drv'...
error: packages '/nix/store/mqiyh67czmisc856ib0j758pv582rz0w-urbit-debug-0.10.9-main-post/bin/urbit-worker' and '/nix/store/izjykmbqpw9xixcz14i6rwfmrnlrfz4c-urbit-0.10.9-main-post/bin/urbit-worker' have the same priority 5; use 'nix-env --set-flag priority NUMBER INSTALLED_PKGNAME' to change the priority of one of the conflicting packages (0 being the highest priority)
builder for '/nix/store/px7y2mc979bwpkbybh6j9fyi9icsdd9r-user-environment.drv' failed with exit code 1
error: build of '/nix/store/px7y2mc979bwpkbybh6j9fyi9icsdd9r-user-environment.drv' failed
make: *** [install] Error 100
```

joemfb:
I think this is caused by the presence of both `urbit` and `urbit-debug` derivations, which both produce a binary named `urbit-worker` (and also one named `urbit`?). I believe I heard that it can be worked by assigning priorities to the derivations.

/cc @eglaysher

PR #4092, opened 2020-12-07
nix: don't install urbit-debug by default.

eglaysher:
The urbit and urbit-debug derivations both produce an urbit-worker
binary, and they conflict.

I've been running this locally for a while, if there's no better fix, this could go in as a workaround.

References #4089

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ build:
 	nix-build -A urbit -A herb --no-out-link
 
 install:
-	nix-env -f . -iA urbit -iA urbit-debug -iA herb
+	nix-env -f . -iA urbit -iA herb
 
 release:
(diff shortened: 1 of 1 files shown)


joemfb:
I don't have a problem with this. @brendanhay?

brendanhay (approved):

ohAitch:
Isn't it supposed to make an `urbit-debug-worker` and look for it by excutable name?

joemfb:
@ohAitch, it used to work that way, but I'm not sure it's worth preserving. `urbit-debug` is really just for CI, and the whole design could use a refresh (checkpoints aren't compatible across urbit/urbit-debug, for instance).

PR #4093, opened 2020-12-07
clock: ignore cached date for sunrise

tylershuster:
Fixes https://github.com/urbit/landscape/issues/190


matildepark:
So we were always comparing the time from our last request (every ~3 hours)?

tylershuster:
Right.

tylershuster:
This way won't be accurate to the minute but it doesn't need to be

matildepark:
I can't trigger a reproduction in next userspace to test this, heh.

matildepark:
Ah ok it's when we pass the midnight threshold with a 3 hour cache. Really weird.

matildepark (approved):
lgtm

PR #4094, opened 2020-12-07
pull-hook: make wires more stable identifiers based on resource

tacryt-socryp:
Prevent proliferation of wires by removing path-specific details from wires when `%watch`ing from pull-hook.


liam-fitzgerald (approved):

liam-fitzgerald:
righteous

PR #4095, opened 2020-12-07
Add provenance to gall tasks & scrys - confine vanes from gall apps.

xiphiness:
I built this on na-release/candidate so that's the target branch I have, for now, not expecting to get this into the breach though.

I have some other PR's I need to attend to before I continue with this, but a brief outline:

In order for gall apps to appropriately be confined so that they may not arbitrarily read from clay, scry the keys of the jael, or send whatever tasks they choose to gall apps and vanes, I have systematically recorded provenance information into scrys and tasks from gall agents. Right now, provenance is a path, but will only ever be /arvo, /g/[agent], or /[vane] in it's current shape based on how gall.hoon records it. Vanes can arbitrarily choose their provenance - this seems fine, as vanes shouldn't ever really be third party and shouldn't ever act maliciously, and also one can imagine with a more confined userspace client eyre can safely forward provenance information from client actions and so it is good for it to describe its own provenance.

With this provenance information, gall can manage an ACL for vanes, and gall agents may choose whatever permissions system they want, be they acl's or capabilities, enforced by the provenance of tasks and scrys.

This also works for remote actions, and ought to be revisited in light of remote scry when that work matures. Of course, another ship could always fake the provenance of an action, but if one trusts a users *morals*, it is secure up to the person being naive enough to grant an arbitrary third party application to invoke some other application which invokes yours remotely.

Provenance is in the bowl of every gall agent, even for gifts (this is possible for gifts because the provenance is associated with the wire), and is also present in $bitt for all incoming subscriptions(outgoing can simply be constructed with /g/[ship.src.bol], for now), except in +on-peek where it is in the sample of the gate. (need to look into what bowls are when +on-peek is called... it shouldn't have entropy, but may it have other things?...)

I create a new gate in gall.hoon called +en-snare which accepts a path for provenance and returns a $slyt which may be passed into mock in +ap-mule and +ap-mule-peek. Right now, it's very simple, and does the same logic +scry does in gall.hoon, but rather for scrys targeting apps it uses the internal +mo-peek instead of routing it through the roof. However, it should check to make sure the scry'ing application may scry a vane when it's not going to a gall agent.

The main task which remains is confining the vanes, which is now possible with above affordances. It seems like the simplest way here is best, as one can always build more sophisticated permissions system on top of gall agents which have a very simple ACL permission to the vane. If a permission system proves itself to particularly strong in practice, it could be moved up the stack later. It seems like there should be a bespoke agent which has special privileges to update gall's acl on the vanes, and maybe even do some forwarding of tasks and scrys. The permissions may include updating the permissions themselves. The permissions should probably come out of a sensible analysis of all vane tasks and scrys - factoring in read vs write in certain cases, and having attention to detail to the tasks involved in certain vanes. A simple role-based solution here would do the trick, I think, associating roles with provenance paths.

tagging @philipcmonk since this came out of extensive discussion with him

n.b. don't expect this to work by merging into home - I tested this by commenting out the commits to home with herb in solid.sh. I haven't yet built a +load case for gall to migrate from state version %7

### DIFF
--- a/pkg/arvo/app/acme.hoon
+++ b/pkg/arvo/app/acme.hoon
@@ -357,7 +357,7 @@
   ++  on-watch  on-watch:def
   ++  on-leave  on-leave:def
   ++  on-peek
-    |=  =path
+    |=  [prov=path =path]
     ^-  (unit (unit [%noun vase]))
     ?+    path  ~
         [%x %domain-validation @t ~]
--- a/pkg/arvo/app/chat-hook.hoon
+++ b/pkg/arvo/app/chat-hook.hoon
@@ -189,7 +189,7 @@
       ?>  ?=(^ old-path)
       ?.  =('~' i.old-path)
         ~
-      [%give %kick ~[mailbox+old-path] ~]~
+      [%give %kick ~[mailbox+old-path] ~ ~]~
     ::
     ++  add-members-group
       |=  [=path ships=(set ship)]
@@ -531,7 +531,7 @@
   ?>  ?=(%message -.act)
   ::  local
   :_  state
-  ?:  (team:title our.bol src.bol)
+  ?:  (team:title our.bol ship.src.bol)
     ?.  (~(has by synced) path.act)
       ~
     =*  letter  letter.envelope.act
@@ -546,8 +546,8 @@
   ?~  ship  ~
   ?.  =(u.ship our.bol)  ~
   ::  check if write is permitted
-  ?.  (is-member:grp src.bol (group-from-chat path.act))  ~
-  =:  author.envelope.act  src.bol
+  ?.  (is-member:grp ship.src.bol (group-from-chat path.act))  ~
+  =:  author.envelope.act  ship.src.bol
       when.envelope.act  now.bol
   ==
   [%pass / %agent [our.bol %chat-store] %poke %chat-action !>(act)]~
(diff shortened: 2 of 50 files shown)


matildepark:
You'll have to remake this PR, @xiphiness, sorry about that — you can't change the target branch if the original branch is deleted I guess?

xiphiness:
c'est la vie - I rebased and refactored anyway, now the only breaking change is /lib/dbug.hoon, $bowl:spider and convert-bowl in app/spider.hoon

joemfb:
I'm sorry, I forgot that branch was still targeted. I've restored the reference. You'll most likely want to rebase off of master anyway, and then target `release/next-sys`. But you can do all that in this PR.

zalberico:
@belisarius222 @philipcmonk I'm not sure how to evaluate this one and it's a few years old, is this something we want to clean up and merge or should I close it? I'm moving it to the review column.

xiphiness:
never finished this, got distracted by other things. permission work is largely doing what this was intended to do.

xiphiness:
new desk based perms is probably a better model, though I still feel quite awkward about remote subs being able to be done by any remote app (unless this was changed)

always thought "peer is trustful but maybe dumb but not dumb enough to grant random app arbitrary perms" was a security model worth designing around, but some have disagreed with me on that

zalberico:
> never finished this, got distracted by other things. permission work is largely doing what this was intended to do.

Thanks for the super fast response - I didn't want to close this out without reaching out first since it looked like you put a lot of thought into it.

PR #4099, opened 2020-12-08
jael: deconstruct wires accurately

philipcmonk:
there's no correct way to deconstruct a duct, but we can at least do it accurately


joemfb (approved):
LGTM

PR #4100, opened 2020-12-08
vere: webslog atom tanks, and webflogs

Fang-:
As discussed out of band: fixes the webslog callback to handle cords correctly, fixes the flog callback, and sends flogs to the webslog callback.

### DIFF
--- a/pkg/urbit/vere/lord.c
+++ b/pkg/urbit/vere/lord.c
@@ -366,6 +366,8 @@ _lord_plea_slog(u3_lord* god_u, u3_noun dat)
 static void
 _lord_plea_flog(u3_lord* god_u, u3_noun dat)
 {
+  u3_pier* pir_u = god_u->cb_u.ptr_v;
+
   if ( c3n == u3a_is_atom(dat) ) {
     return _lord_plea_foul(god_u, c3__flog, dat);
   }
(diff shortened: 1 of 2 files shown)


pkg/urbit/vere/lord.c
```diff
@@ -366,17 +366,15 @@ _lord_plea_slog(u3_lord* god_u, u3_noun dat)
...
+    return _lord_plea_foul(god_u, c3__flog, dat);
   }
 
-  c3_c* tan_c = u3r_string(tan);
+  c3_c* tan_c = u3r_string(dat);
   u3C.stderr_log_f(tan_c);
   c3_free(tan_c);
-  u3z(dat);
+
+  god_u->cb_u.slog_f(god_u->cb_u.ptr_v, 0, dat);
```
joemfb, line 377: I don't think we want to print the `%flog` messages twice, isn't that what happens here?

joemfb (changes requested):
The slogstream fix looks good. And thanks for catching and fixing the uninitialized variable use in %flog handling.

But I'm don't think the logic change in %flog is quite right:

pkg/urbit/vere/lord.c
Fang-: Oh, you're right, I got my callbacks confused. Looks like this indeed double-prints in the real terminal. Do we care about explicitly doing `stderr_log_f` here, or are we fine with saying "let the slog callback handle this"?

Fang-: It adds an extra `\n`, which isn't very nice... Probably best to just call the webslog callback directly here.

joemfb: This looks good, we can tweak it later.

joemfb (approved):
LGTM

joemfb:
I had pushed a fix for the uninitialized variable to the target branch too. Should be simple to deconflict -- feel free to merge.

Fang-:
Alternatively we can add a `_pier_on_lord_flog` so that `lord.c` doesn't have to know about the webslog callback, but maybe that's too much indirection for that small benefit.

Issue #4102, opened 2020-12-08
Landscape Crashes on Missing Default Background Color


Sneagan:
**Describe the bug**
Landscape crashes when the background is set to an image (I think. See Additional context section).

**To Reproduce**
1. Load Landscape with a background image set.
2. Wait a second after render during which time the interface is intractable.
3. Experience crash

**Expected behavior**
Rendered interface should continue working.

**Desktop (please complete the following information):**
 - OS: Manjaro 20.2
 - Browser: Firefox, Brave
 - Base hash of your urbit ship: [%base-hash ~[0v14.25lr5.4ps7o.2e8qc.j8uo3.7lc4t.a4u4k.k39qk.4e7ff.mlrjc.3eo6o]

**Additional context**
Used `flagday` on a star and 2 ships. The only one experiencing this issue is the ship with an image background and the error appears to be happening [here](https://github.com/urbit/urbit/blob/master/pkg/interface/src/views/App.js#L111). The error is below:

```
Uncaught TypeError: Cannot read property 'color' of undefined
    at r.value (App.js:111)
    at r.value (App.js:141)
    at Pa (react-dom.production.min.js:182)
    at Na (react-dom.production.min.js:181)
    at gs (react-dom.production.min.js:263)
    at cl (react-dom.production.min.js:246)
    at sl (react-dom.production.min.js:246)
    at Js (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
```

matildepark:
If it's happening there then it means we don't have `/~/default` in contacts.

matildepark:
We could at least catch this with option chaining and an elvis operator to fall back but I'd be curious why your migration is missing a contact entry here. Possibly the ship saved no metadata for itself, so the migration script imported ... nothing and didn't instantiate that entry or something.

Sneagan:
I can see invites to things in notifications before the UI crashes. Not sure if I won't just hit another but immediately, but I made #4103 in case.

matildepark:
You can always run the dev environment against your ship fairly easily, see the CONTRIBUTING doc in pkg/interface.

Sneagan:
Ah great idea. Thanks.

Sneagan:
I used #4103 to skip this error and discovered what is probably the root issue. I am able to operate my ship normally, but my profile page does not render at all. [See here](https://imgz.org/i6eYBcKC/)

[...]

PR #4103, opened 2020-12-08
Add deeper property existence check to landscape background value setting

Sneagan:
Component: Landscape

Check more specifically for the existence of a `color` attribute before attempting to override the default of `#ffffff`.

### DIFF
--- a/pkg/interface/src/views/App.js
+++ b/pkg/interface/src/views/App.js
@@ -106,10 +106,7 @@ class App extends React.Component {
   }
 
   faviconString() {
-    let background = '#ffffff';
-    if (this.state.contacts.hasOwnProperty('/~/default')) {
-      background = `#${uxToHex(this.state.contacts['/~/default'][window.ship].color)}`;
-    }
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/App.js
```diff
@@ -108,7 +108,11 @@ class App extends React.Component {
   faviconString() {
     let background = '#ffffff';
     if (this.state.contacts.hasOwnProperty('/~/default')) {
-      background = `#${uxToHex(this.state.contacts['/~/default'][window.ship].color)}`;
+        if (this.state.contacts['/~/default'].hasOwnProperty(window.ship)) {
```
matildepark, line 111: Wouldn't the following be cleaner?

```
background = this.state.contacts?.['/~/default']?.[window.ship]?.color ? `#${uxToHex(this.state.contacts['/~/default'][window.ship].color)}` : '#ffffff';`
```

Also format the commit `interface: commit name`.

Sneagan: It would, I was just matching the existing check. Will update.

matildepark:
Closing this because it isn't the correct fix for this. We're interested in figuring out why this one specific case is producing a crash for you — continue in #4102

PR #4105, opened 2020-12-09
chat-hook: send right amount of backlog

ixv:


philipcmonk (approved):
Works on ~darrux-landes

PR #4106, opened 2020-12-09
proxy-hook: first draft

liam-fitzgerald:
The poke proxy hook forwards pokes onto the owner of the resource, and persists the pokes until they are acked.
- pending messages will persist across refreshes.
- more detailed liminal states can be shown in the UI, like "here's a list of groups you're currently trying to join, but can't because they're offline"

This a very rough first draft, as there are a bunch of open questions about this architecture:
- Should the proxy-hook handle foreign pokes as well?
- How do we format the unacked pokes so they're accessible in watches and peeks? Maybe it should be the  responsibility of the inner core to expose a reasonable (i.e. mark conversion friendly) API? I've hacked up some custom JSON for this, but I suspect it will prove insufficient for real use cases. 
- Is this even worth being a wrapper library, or should it be a standalone agent? Could probably use mark conversions for `resource-for-update`

### DIFF
--- a/pkg/arvo/lib/poke-proxy-hook.hoon
+++ b/pkg/arvo/lib/poke-proxy-hook.hoon
@@ -1,0 +1,219 @@
+/+  resource, default-agent, dbug
+|%
++$  card  card:agent:gall
+::
++$  state-0
+  $:  %0
+      waiting=(jar resource vase)
+      inner-state=vase
+  ==
+::
++$  config
+  [foreign=term =mark]
+::
+::  TODO: move to +zuse
+++  append-ju
+  |*  [a=(jar) b=* c=*]
+  =/  d=(list _c)
+    (~(get ja a) b)
+  (~(put by a) b (snoc d c))
+::
+++  tail-ju
+  |*  [a=(jar) b=*] 
+  =/  c=(list _?>(?=(^ a) ?>(?=(^ q.n.a) i.q.n.a)))
+    (~(get ja a) b)
+  %+  ~(put by a)  b 
+  ?~(c c t.c)
+::
+++  poke-proxy-hook
+  $_  ^|
+  |_  bowl:gall
+  ::  +on-proxied-poke: process a proxied poke
+  ++  on-proxied-poke
+    |~  vase
+    *[(list card) _^|(..on-init)]
+  ::
+  ++  resource-for-update
+    |~  vase
+    *(unit resource)
+  ::  from agent:gall
+  ::
+  ++  on-init
+    *[(list card) _^|(..on-init)]
+  ::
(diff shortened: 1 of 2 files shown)


matildepark:
ping @tacryt-socryp

pkg/arvo/lib/poke-proxy-hook.hoon
```diff
@@ -0,0 +1,219 @@
...
+      :_  this
+      =/  =rave:clay  [%next %b [%da now.bowl] /[mark.config]/json]
+      [%pass wire %arvo %c %warp our.bowl [%home `rave]]~
+    ==
+
+  ++  on-fail   on-fail:def
+  ++  on-peek   on-peek:def
+  --
+  ::
+  |_  =bowl:gall
```
tacryt-socryp, line 193: Shorter core should come before the agent core in this case, no?

pkg/arvo/lib/poke-proxy-hook.hoon
```diff
@@ -0,0 +1,219 @@
+/+  resource, default-agent, dbug
```
tacryt-socryp, line 1: I agree with you that it seems the best approach here is to make this its own app rather than a library. Seems unnecessary for us to make individual versions of this app for each use case, given how little is different between groups and graphs so far. Making this an individual app would also improve client-side ergonomics, as we can keep track of pending pokes via a singular subscription.

Something I'm curious about is the idea of chaining various pokes and then "inverting them" if one fails. We handle this via threads currently, but I see potential in this architecture to do something similar. For instance, sending a group-store join that "fails" and removes the group if any of the subscriptions to contacts, metadata, or group is nacked. That is a "pending complex transaction" which is more complicated than a chat message, but would be extraordinarily useful for communicating internal state to users.

tacryt-socryp (changes requested):
The idea has legs and should move forward, particularly as an app.

pkg/arvo/lib/poke-proxy-hook.hoon
liam-fitzgerald: Sadly, writing this as an app is thwarted by the cage of a cage conversion problem. I guess we could just emit the fact as json and do the conversion in the agent with some json conversions.

matildepark:
Is this superseded?

liam-fitzgerald:
Not superseded per se. Personally this should be revived when infra solves scrying inside a mark, as that would enable the ideal architecture for this app. We've solved the most pressing problems that this was designed to fix, but I definitely think this is worth doing, the UX benefit is too good to pass up.

zalberico:
Closing this for now, if/when it's the focus we can reopen it.

PR #4115, opened 2020-12-09
vere: commit the specified mountpoint only

Fang-:
Instead of committing all mountpoints regardless of what was specified when running `|commit`.

This came up in discussion on #3240. Couldn't find an issue for it.


joemfb (approved):
I support this.

philipcmonk (approved):
This will be nice

PR #4118, opened 2020-12-10
clock: add background to moon

tylershuster:
Fixes https://github.com/urbit/landscape/issues/186


pkg/interface/src/views/apps/launch/components/tiles/clock.js
```diff
@@ -202,9 +202,10 @@ class Clock extends React.PureComponent {
   }
 
   initGeolocation() {
-    if (typeof this.props.data === 'object') {
-
-      const { latitude: lat, longitude: lon } = this.props.data;
+    if (typeof this.props.data === 'string') {
+      const latlon = this.props.data.split(',');
+      const lat = latlon[0];
+      const lon = latlon[1];
```
matildepark, line 208: unrelated change?

tylershuster: looks like it don't it

matildepark (approved):
lgtm

PR #4119, opened 2020-12-10
observe-hook: spider wasn't always issuing thread-results, so we got %watch-not-unique. we now make unique wires, but spider's issue remains

tacryt-socryp:
Fixes #4114. We got a repro on Isaac's ship and resolved it by making this change. This issue may be representative of a deeper issue in `%spider` where spider may not be issuing `%thread-results` deterministically every time a thread is completed. Unsure. Regardless, no functionality was broken, and everything continues to work in this PR.

cc @philipcmonk

PR #4120, opened 2020-12-10
group-on-leave: soften de-path to fix thread and add thread back to observe-hook state

### LINKED ISSUE #2992: metadata-store: retains group metadata after deletion

tacryt-socryp:
Fixes #2992 for good. Depends on #4119.

### DIFF
--- a/pkg/arvo/app/observe-hook.hoon
+++ b/pkg/arvo/app/observe-hook.hoon
@@ -64,6 +64,9 @@
   =|  cards=(list card)
   |-
   ?:  ?=(%2 -.old-state)
+    =.  cards
+      :_  cards
+      (act [%watch %group-store /groups %group-on-leave])
     [cards this(state old-state)]
   ?:  ?=(%1 -.old-state)
     =.  cards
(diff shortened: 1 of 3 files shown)


matildepark:
It looks like it supersedes #4119.

liam-fitzgerald (approved):

PR #4121, opened 2020-12-10
landscape: add woff2 support, serve Inter & Source Code Pro

tylershuster:
Fixes urbit/landscape#92


pkg/arvo/app/file-server.hoon
```diff
@@ -216,9 +216,10 @@
         =/  file  (as-octs:mimes:html .^(@ %cx scry-path))
```
joemfb, line 216: @tylershuster, you're scrying for all the files as `@`, and then later scrying again for `%woff2` as `$octs`. In the case of a `%woff2` mark, the first one will fail and you'll never get to the second one.

tylershuster: Thank you, that's very clear.

tylershuster:
Ok, the syntax should be "slightly tidier" but I am not good with computer. Nonetheless, it works!

pkg/arvo/app/file-server.hoon
```diff
@@ -213,12 +213,15 @@
...
+          [[200 [['content-type' '/font/woff2'] ~]] `.^(octs %cx scry-path)]
         =/  file  (as-octs:mimes:html .^(@ %cx scry-path))
         :_  public.u.content
         ?+  ext.req-line  not-found:gen
-            [~ %js]    (js-response:gen file)
-            [~ %css]   (css-response:gen file)
-            [~ %png]   (png-response:gen file)
+            [~ %js]     (js-response:gen file)
+            [~ %css]    (css-response:gen file)
+            [~ %png]    (png-response:gen file)
```
joemfb, line 224: This whitespace change is now unnecessary.

pkg/arvo/lib/pill.hoon
```diff
@@ -77,7 +77,7 @@
   =?  hav  ?=(^ fil.lon)
       ::  XX  this whitelist needs to be reviewed
       ::
-      ?.  ?=  ?(%css %hoon %html %js %json %md %png %txt %udon %umd)
+      ?.  ?=  ?(%css %hoon %html %js %json %md %png %txt %udon %umd %woff2)
```
joemfb, line 80: Adding these files to the pill is fine, but the same scry-format and serialization need to be applied below. And this will be a new case; we have `@t` and `$wain`,  but no $`octs`.

pkg/arvo/mar/woff2.hoon
```diff
@@ -0,0 +1,12 @@
...
+  |% 
+  ++  mime  [/font/woff2 dat] 
+  -- 
+++  grab 
+  |% 
+  ++  mime  |=([=mite =octs] octs) 
+  ++  noun  octs
+  -- 
+++  grad  %mime 
+--
```
joemfb, line 12: You should have a trailing newline.

joemfb (changes requested):
I'm not too familiar with `:dbug` or `:file-server`, but I'm fine with this assuming it works.

The addition to the pill will require further changes, and I've noted a couple style nits:

pkg/arvo/app/dbug.hoon
```diff
@@ -400,7 +401,7 @@
 ++  get-file-at
   |=  [base=path file=path ext=@ta]
   ^-  (unit octs)
-  ?.  ?=(?(%html %css %js %png) ext)
+  ?.  ?=(?(%html %css %js %png %woff2) ext)
```
Fang-, line 404: The fileserving logic in `/app/dbug` is a holdover from when we didn't have `/app/file-server` yet, and should be ripped out. Your change here would run into the same issue 
@joemfb described in his comments (scrying for woff2 as `@`). You're fine to just revert your changes here, the debug dashboard doesn't serve/need fonts anyway.

pkg/arvo/lib/pill.hoon
tylershuster: TBH I don't even know what I'm doing in this file I just heard there were pngs

joemfb (approved):
The hoon changes LGTM.

Note that without the fonts in the pill, they will not be available in the UI until a ship gets an OTA. I guess whether that's ok is a requirements question for interface.

tylershuster:
bump

Fang-:
> Note that without the fonts in the pill, they will not be available in the UI until a ship gets an OTA. I guess whether that's ok is a requirements question for interface.

@vvisigoth thoughts? I'm inclined to just merge.

tacryt-socryp:
This LGTM as well.

tacryt-socryp (approved):

matildepark:
> Note that without the fonts in the pill, they will not be available in the UI until a ship gets an OTA.

*shrug* We'll keep a CDN fallback I suppose. @tylershuster 

In we go!

matildepark:
Oh wait, you make the JS changes here too. Well, we have a sec, if we could optimise with CDN fallback that would be ideal. First OTAs can be a half hour where new users would have Times New Roman Landscape™

tylershuster:
I should target `release/next-userspace` with that then?

matildepark:
Yes!

Issue #4130, opened 2020-12-11
landscape: mentions don't respect nickname setting


Fang-:
I have both nicknames and avatars disabled in my landscape UI. Yet mentions render as the nickname for the corresponding ship.

<img width="234" alt="image" src="https://user-images.githubusercontent.com/3829764/101926551-9b070980-3bd3-11eb-91d7-fa12c44793ee.png">

Interestingly, the nickname is included in the profile overview here (which is what I want, as per urbit/landscape#65), but isn't when clicking on the sigil next to their messages.

matildepark:
Oooh ok easy catch. 

— 
~haddef-sigwen
https://urbit.org

Issue #4127, opened 2020-12-11
landscape: chat day/section separator timestamps always now


Fang-:
This screenshot was taken at 11:17am. Coming back to this chat a couple minutes later, all the separator timestamps had updated to match the current time again.

<img width="242" alt="image" src="https://user-images.githubusercontent.com/3829764/101891589-8b230180-3ba2-11eb-9856-7df99925abbb.png">

(They're also not perfectly aligned with each other, but that's secondary.)

Vaguely remember something like this happening before, so maybe a regression?

matildepark:
Almost certainly just passing an undefined datestamp to moment.

PR #4132, opened 2020-12-11
chat: fixes two regressions

matildepark:
- Passes the correct timestamp to Daybreak (#4127)
- Passes CalmEngine props to Mention components (#4130)


tylershuster:
Hurry up with the state hooks already @tylershuster this is ridiculous

tacryt-socryp (approved):

Issue #4012, opened 2020-11-24
hark: adding nodes produces multiple notifications or unreads


vvisigoth:
**Describe the bug**
Instead of receiving one notification per comment, I get several

**To Reproduce**
1. Post a link in a collection
2. Someone makes a comment
3. Get ~3x notifications per comment

**Expected behavior**
One comment = one notification

![Screen Shot 2020-11-24 at 2 32 00 PM](https://user-images.githubusercontent.com/1609955/100148819-e34bcb00-2e62-11eb-8d14-1c1a0942be1c.png)

matildepark:
Somewhat separate but somewhat related to #3971.

liam-fitzgerald:
largely sure that this a result of stale versions of landscape. just forced the ota on ~bitpyx-dildus, any other hosts known to repro this issue?

vvisigoth:
@liam-fitzgerald  this is still an issue on ~bitpyx which is up to date

matildepark:
As per investigations with @liam-fitzgerald overnight as we watched the deploy go out, this appears to be an issue regarding the behaviour between hark-graph-hook and the graph-store itself. 

Multiple addNode events may occur in graph-store, which no-ops; hark-graph-hook listens to graph-store for updates. The duplicate updates are not in the log, but they are in hark-graph-hook.

This has been a latent issue with graph node notifications which has, as of #4058, become a higher priority, since it leads to unreads and notifications coming in high numbers. 

@liam-fitzgerald @tacryt-socryp I would like you to consider this priority one and if we can get a band-aid fix out to prevent surfacing these duplicates soon, we should.

PR #4137, opened 2020-12-11
graph-store: no-op if we already have *any* of the indexes in an %add-nodes event

tacryt-socryp:
For some reason, we receive 8-10 facts for every individual update that is performed to graph-store. I believe the root cause stems from the `lib/push-hook` or `lib/pull-hook` libraries, given the behavior, though it could also stem from the frontend if we are repeatedly sending events to the host. My guess is that the issue is in `%pull-hook` somehow. However, we now no-op aggressively within `%graph-store`, which is righteous, and we no longer spam notifications on livenet, a very noticeable symptom.

Unfortunately, we now query graph-store twice as much in order to determine if we already have a node. I don't expect noticeable slow-down, but it will definitely affect the time complexity of the `%add-nodes` operation.

The root cause should be investigated as soon as possible next week.

Fixes the symptom of #4012 but not the root cause.

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -297,10 +297,8 @@
       |^
       =/  [=graph:store mark=(unit mark:store)]
         (~(got by graphs) resource)
-      ::  no-op if we already have any of the nodes
-      ?:  (check-for-duplicates resource ~(key by nodes))
-        ::~&  nooped-due-to-duplicate-nodes+[resource nodes]
-        [~ state]
+      ~|  "cannot add duplicate nodes to {<resource>}"
+      ?<  (check-for-duplicates graph ~(key by nodes))
       =/  =update-log:store  (~(got by update-logs) resource)
       =.  update-log
         (put:orm-log update-log time [%0 time [%add-nodes resource nodes]])
@@ -316,24 +314,21 @@
       ==
       ::
       ++  check-for-duplicates
-        |=  [=resource:store nodes=(set index:store)]
+        |=  [=graph:store nodes=(set index:store)]
         ^-  ?
-        =|  has-duplicates=_|
+        =/  has-duplicates  %.n
         =/  node-list  ~(tap in nodes)
         |-
         ?~  node-list
           has-duplicates
-        =?  has-duplicates  (has-node resource i.node-list)  %.y
+        =.  has-duplicates
+          |(has-duplicates (has-node graph i.node-list))
         $(node-list t.node-list)
       ::
       ++  has-node
-        |=  [=resource:store =index:store]
+        |=  [=graph:store =index:store]
         ^-  ?
-        =/  parent-graph=(unit marked-graph:store)
-          (~(get by graphs) resource)
-        ?~  parent-graph  %.n
         =/  node=(unit node:store)  ~
-        =/  =graph:store  p.u.parent-graph
         |-
         ?~  index
           ?=(^ node)
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -297,6 +297,10 @@
       |^
       =/  [=graph:store mark=(unit mark:store)]
         (~(got by graphs) resource)
+      ::  no-op if we already have any of the nodes
+      ?:  (check-for-duplicates resource ~(key by nodes))
+        ::~&  nooped-due-to-duplicate-nodes+[resource nodes]
+        [~ state]
       =/  =update-log:store  (~(got by update-logs) resource)
       =.  update-log
         (put:orm-log update-log time [%0 time [%add-nodes resource nodes]])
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/graph-store.hoon
```diff
@@ -311,6 +315,37 @@
...
       ::
+      ++  check-for-duplicates
+        |=  [=resource:store nodes=(set index:store)]
+        ^-  ?
+        =|  has-duplicates=_|
+        =/  node-list  ~(tap in nodes)
+        |-
+        ?~  node-list
+          has-duplicates
+        =?  has-duplicates  (has-node resource i.node-list)  %.y
```
Fang-, line 326: Isn't this equivalent to:

```suggestion
        =.  has-duplicates  |(has-duplicates (has-node resource i.node-list))
```

If yes, why do we try the entire node-list, instead of returning `%.y` as soon as we find a duplicate?

pkg/arvo/app/graph-store.hoon
```diff
@@ -311,6 +315,37 @@
...
+        $(node-list t.node-list)
+      ::
+      ++  has-node
+        |=  [=resource:store =index:store]
+        ^-  ?
+        =/  parent-graph=(unit marked-graph:store)
+          (~(get by graphs) resource)
+        ?~  parent-graph  %.n
```
Fang-, line 334: Why don't we move this up into the start of `+check-for-duplicates`, and pass it in as an argument here? No need to `(~(get by graphs) resource)` on every node check for the same graph, right?

pkg/arvo/app/graph-store.hoon
```diff
@@ -297,6 +297,10 @@
       |^
       =/  [=graph:store mark=(unit mark:store)]
         (~(got by graphs) resource)
+      ::  no-op if we already have any of the nodes
```
Fang-, line 300: _Any_? Do we not still want to take action, if some of these are new to us?

belisarius222:
I think this should crash, not no-op.  The transaction failed if there are any non-duplicate nodes.

Fang- (changes requested):
Wondering if there's some graph walking we can do to take advantage of it being an _ordered_ graph, but not familiar with that, and seems like unnecessary optimization for the time being.

As stands though, I think you can get very cheap gains here. Some suggestions for easy performance gains below.

Also a question about when exactly we no-op. I'm not super familiar with graph-store and it's semantics, but no-oping when you get data you already know certainly seems justified.

tacryt-socryp:
Making appropriate changes. We will now crash instead of no-opping, as the transaction will have failed, and I am incorporating Mark's performance feedback. I don't think we can get a lot of benefit from the fact that it's an ordered-map, and think it's good to punt that to later (if at all).

Fang-:
> crash

If we're doing that on receiving a `%fact`, doesn't that kill our subscription? I could be mistaken, or just have it the wrong way around, but I remember getting bitten by gall behavior like that at some point.

tacryt-socryp:
@Fang- The crash will occur on processing a poke, not a fact, so luckily we can crash however we wish.

pkg/arvo/app/graph-store.hoon
tacryt-socryp: Correct, the transaction as a whole will have failed in that case.

pkg/arvo/app/graph-store.hoon
tacryt-socryp: Good point.

pkg/arvo/app/graph-store.hoon
tacryt-socryp: Also a good point. Addressed this also.

tacryt-socryp:
@Fang- Addressed all of review. Thanks!

pkg/arvo/app/graph-store.hoon
```diff
@@ -311,6 +313,31 @@
...
       ==
       ::
+      ++  check-for-duplicates
+        |=  [=graph:store nodes=(set index:store)]
+        ^-  ?
+        =/  node-list  ~(tap in nodes)
+        |-
+        ?~  node-list  %.n
+        ?:  (has-node graph i.node-list)  %.y
+        $(node-list t.node-list)
```
Fang-, line 323: imo it's nicer to just do `?|` or w/e instead of explicitly returning booleans, when doing boolean logic.

pkg/arvo/app/graph-store.hoon
Fang-: I think I'm missing context, because I don't understand why you don't still want to process the unknown/new content, but I'll believe it.

Fang- (approved):
Oh oops, I guess that posted already somehow.

This looks fine to me. A style comment, but good to go either way, afaict.

joemfb:
When we crash on this poke, we'll generate a negative acknowledgement. Who's poking us, and how will they handle that?

tacryt-socryp:
Either graph-pull-hook or graph-push-hook is poking us, and either way they'll ignore the nack. This is good as is.

PR #4138, opened 2020-12-11
debug: render queued ames acks correctly

Fang-:
Wrap the items in a `SearchableList` instead of throwing the objects into the DOM directly.

(Notable, debug dashboard still uses `gulp` instead of whatever the fancy modern landscape thing is. Having debug not rely on file-server or glob or the likes is nice, but it doesn't let you do the fancy dev setup, and the changes will accrue in clay history. I'll make an issue for this.)


ixv (approved):
We fixed our ames issue, which subsequently made this issue go away, so I can't really test this. The logic makes sense though, and if you tested it with the json dump I gave you, then I'm happy to approve this.

Fang-:
Good news! And I did test with the dump, thanks again for sharing that.

PR #4142, opened 2020-12-12
graph-pull-hook: silence issues temporarily

tacryt-socryp:
Silence issues with nacked `%poke-ack`s occurring in `%graph-pull-hook` due to the mysterious duplicate message bug.

If anyone can stamp this, that would be nice.


philipcmonk (approved):
Tempted to add a "if before tuesday" condition so it automatically turns back on.... but this is good

PR #4143, opened 2020-12-12
ames: Fix stuck flows caused by %strange-current

### LINKED ISSUE #2505: arvo: replacement event (for %hear) failed
**Describe the bug**
Repeated error message (perhaps) associated with segmentation fault (not captured but it was position out of loom):
`arvo: replacement event (for %hear) failed`

**To Reproduce**
Complex series of events - perhaps joining a chat w/o permissions started this - noticed it after OS1 upgrade but not immediately thereafter

**Expected behavior**
N/A - Would like to stop them from occurring to see if that prevents (potentially - not running long enough) nightly seg faults

**Screenshots**
image

 - OS: Ubuntu 18.4
 - Browser N/A
 - Base hash of your urbit ship. [%base-hash 0v8.rpvar.1t9h7.4es5r.irauk.niaqo.7do90.bpc1r.nihtd.ufdgt.8ehjl]

**Additional context**
*sad face*

philipcmonk:
@belisarius222 and I identified an issue about a year ago that caused flows to occasionally stick, so we added the `%strange-current` assertion to try to catch the issue earlier.  This has showed up from time to time in the last year, and @ixv and I worked out a fix.

An example of the problem is when we're sending messages 288 and 289 and we've received a message-level ack for 289 but not 288, but there are also fragments of 289 which have not been acknowledged, so they remain in our packet pump.  The fragments of 289 *should* have been removed from packet pump when the whole message was acknowledged, but for some reason they weren't.

The first commit in this PR simply runs `on-done` in the packet pump (this is what clears straggler fragments from the packet pump) as we process messages, so that if for some reason there are stragglers, they're removed.  As noted in the comment, this is slightly inefficient but is not a problem I believe.  I still wish I understand why they're not removed earlier.

The second commit adds tags to the bare numbers in ames-verb printfs, because it takes way longer to tell what's happening without them.

The third commit threads error traces a little further and prints them if we crash while handling a message-level ack.  Consistently crashing while handling a message-level ack is very bad since all we can do is retry -- if we skip giving that to a vane, then we violate the invariant that vanes will receive all acks in order.

fixes #2505

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2381,6 +2381,19 @@
     ::
     =.  queued-message-acks.state
       (~(del by queued-message-acks.state) current.state)
+    ::  clear all packets from this message from the packet pump
+    ::
+    ::    Note we did this when the original packet came in, a few lines
+    ::    above.  It's not clear why, but it doesn't always clear the
+    ::    packets when it's not the current message.  As a workaround,
+    ::    we clear the packets again when we catch up to this packet.
+    ::
+    ::    This is slightly inefficient because we run this twice for
+    ::    each packet and it may emit a few unnecessary packets, but
+    ::    but it's not incorrect.  pump-metrics are updated only once,
+    ::    at the time when we actually delete the packet.
+    ::
+    =.  message-pump  (run-packet-pump %done current.state lag=*@dr)
     ::  give %done to vane if we're ready
     ::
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
I'll approve this, because this bug is heinous and any fix that brings us to correctness should be celebrated.  Far be it from me to stand athwart history.

I'll also look into what goes wrong when +on-done is called on a non-current message.  That must be made to work.

PR #4145, opened 2020-12-12
dbug: make subscription items header match the table

jfranklin9000:
I changed the header to match the table instead of the other way around  due to my preference.
Untested.  Did not update `pkg/arvo/app/debug/js/index.js` or pill.  Not sure if this is the correct branch to PR against.
If the change is correct can someone take this and do a proper PR?


Fang-:
Change looks good, nice catch. This is indeed the sane order to stick with.

I'd target `release/next-userspace`. I think you can `gulp bundle-prod` to build the release `index.js`. You may or may not need to make sure your `urbitrc` points to an actual pier. You don't need to update the pill for this.

jfranklin9000:
Moved to https://github.com/urbit/urbit/pull/4163.

PR #4155, opened 2020-12-14
notifications: fix invite squashing

### LINKED ISSUE #4111: invites stick to top of notifications, but overlap
So their content visually overlaps as you scroll down. Wouldn't be as bad if they had backgrounds, but we probably don't want to pin them to top if there's this many, anyway.

ttttt

liam-fitzgerald:
- Set scrolling box as `relative`
- invite container is sticky, not the invites themselves
- fix spacing on border between invites

Fixes #4111


tacryt-socryp (approved):
LGTM

Issue #4004, opened 2020-11-24
eyre: cannot access Last-Event-ID parameter in eventsource


liam-fitzgerald:
We can't send custom headers on EventSource requests in the browser, so we can't use the channel resume functionality. We could ideally send this as a query parameter or something like that, so we can actually resume these.

Fang-:
Relevant reading is https://github.com/whatwg/html/issues/2177#issuecomment-332071504 (and the rest of that thread). tldr "needs implementer interest" but browser vendors think they have better things to do. Recommended solution that doesn't modify the server-side is to use `fetch()`-based event streaming.

Worth looking into, I'd avoid doing non-standard things on the Eyre side if we can help it...

tylershuster:
In my TS lib I've accessed `event.lastEventId` without problem. Is this something else?

ohAitch:
I think this is access *for writing* when resuming connections. Which in theory the browser is supposed to do for you?

liam-fitzgerald:
Yeah it's for resuming interrupted eventsources, instead of having to restart all subscriptions

PR #4157, opened 2020-12-14
eyre: allow last-event-id as query parameter

liam-fitzgerald:
Browsers do not support setting custom headers on EventSources, so we
allow the last-event-id to be sent as a query parameter. See below link
for reference.
https://github.com/whatwg/html/issues/2177#issuecomment-332071504

Fixes: #4004


Fang-:
Neat.

As per my comment in https://github.com/urbit/urbit/issues/4004#issuecomment-732502980, did you look into making `channel.js` use `fetch` instead? Is that unergonomic, unreliable, or in some other way unreasonable?

tylershuster:
Do you mean for EventSources? `fetch` isn't a solution for processing data from SSEs — a more modern way would probably be to use websockets, because SSE support is a bit lacking as you can see in the WhatWG discussion.

ohAitch:
Is it not? `fetch` exposes a [stream](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream), which can polyfill a less restricted EventSource API: https://github.com/Azure/fetch-event-source

Though I haven't evaluated the quality of that particular implementation in detail, so it might not literally be as easy as `npm install @microsoft/fetch-event-source`

tylershuster:
Interesting, this didn't exist when I was doing my research. Seems promising, and a clever use of the exposed stream. Would definitely have made my TS lib easier.

liam-fitzgerald:
hmm, closing in favour of a fetch based stream

PR #4158, opened 2020-12-14
chat: lazier chat loading

liam-fitzgerald:
Load smaller amounts of backlog when displaying a chat


vvisigoth:
Does this also apply on initial load?
—
~poldec-tonteg
http://urbit.org

tacryt-socryp:
@vvisigoth Yes, it would fetch 50 messages on initial load rather than 150.

tacryt-socryp (approved):

PR #4161, opened 2020-12-14
default.nix: fix typo in examples section

botter-nidnul:
should be `enableStatic` not `enableSatic`

Took me a couple of minutes to figure out why `nix-build -A urbit --argstr crossSystem x86_64-unknown-linux-musl --arg enableSatic true` was trying to build different things than `nix-build -A urbit --arg enableStatic true` even though they were supposed to be exactly equivalent commands.


joemfb (approved):
LGTM

PR #4163, opened 2020-12-15
dbug: make eyre Channels header match the table

jfranklin9000:
Supersedes https://github.com/urbit/urbit/pull/4145.

This PR updates `pkg/arvo/app/debug/js/index.js`.  It also updates ` pkg/arvo/app/debug/js/tile.js` because it also changed after the `gulp bundle-prod` command.

I did `npm install` and then `npm update` which caused an enormous diff to ` pkg/interface/dbug/package-lock.json` (also included in this PR).  The diff is so large I'm not sure if I've done something wrong.

I did confirm the changed header with a fakezod.

One thing I noticed in the JS console, _even before my changes_, is this warning:
```
Warning: Each child in a list should have a unique "key" prop.

Check the render method of `cf`. See https://fb.me/react-warning-keys for more information.
    in sf (at skeleton.js:37)
    in cf (at root.js:40)
    in t (at root.js:37)
    in t (at root.js:36)
    in t (created by t)
    in t (at root.js:36)
    in Unknown (at src/index.js:16) index.js:1:5731
```


Fang- (approved):
> I did npm install and then npm update which caused an enormous diff to pkg/interface/dbug/package-lock.json

Yeah, I guess `npm update` would do that. As long as nothing broke with that update, it's probably fine. This would also explain while the tile changes, even though there's barely any code in there, and it's not actually used.

> Warning: Each child in a list should have a unique "key" prop.

Guess we're not being entirely clean about setting element keys. This isn't a big deal, I don't think.

This should be fine. Thank you!

Issue #4133, opened 2020-12-11
hark: resources don't consistently update 'last' timestamp


matildepark:
This one confused me because it works properly within a single Landscape session — if you refresh and `console.log()` hark updates as they come in, a lot of resources' `last` are `946684800000`, that is, 01/01/2000 at 00:00 UTC, even if we've gotten messages from them post-update?

This doesn't appear to be enforced as a default on front-end, so I can only assume we aren't updating `last` on back-end (or are otherwise overwriting the `last` in state with the oldest unread we have for them).

It affects graphs created after #4058 and imported resources alike.

PR #4166, opened 2020-12-15
graph-store: run update log in correct order

liam-fitzgerald:
tapping an ordered map returns a list with the smallest keys first.
However, graph-store's update-log is defined backwards, such that the
list will have the most recent updates in the front of the list. We
therefore flop the update log before iterating over it, in order to
process the updates in the order they occurred. 

A partial mitigation for #4133 and probably a lower risk deployment that #4156


tacryt-socryp:
What? We should just change ordered-map to use the comparator in the tap function. This is an ordered-map level change.

tacryt-socryp:
If you add a TODO to switch out tap:ordered-map for a generic implementation, we can push this out as is.

liam-fitzgerald:
It already does use the comparator which is the issue. I don't think it's valid to use any other comparator for ordering the list, maybe we should just add bap:orm ie backwards tap

tacryt-socryp:
Oh, I understand now. Yes, I like the idea of making a backward order tap. It could be far more efficient than flopping. Could you just add a TODO comment here to revisit that, and we can merge as is?

tacryt-socryp (approved):

matildepark:
Is this superseded by the #4156 merge?

tacryt-socryp:
Not superceded, but the implementation of this (using +tap, then +flop) as opposed to using +bap should be updated. We should use +bap now.

matildepark:
Which would mean this would target next-sys too.

tacryt-socryp:
Nope, this can target master. The addition of +bap is already on livenet, it just isn't being used yet.

PR #4167, opened 2020-12-15
drum: update app index early when unlinking

Fang-:
This fixes an issue where ctrl-d-ing to unlink the current app would crash.

The added logic also happens in `+se-drop`, which will get called by
`+se-abet` if we did unlink an application. But `+se-agon` depends on the
index being sane, and may be called between `+se-klin` and `+se-abet`.


joemfb (approved):
This part of drum is frustratingly fiddly ...

LGTM

PR #4170, opened 2020-12-15
u3: revive hashboard

joemfb:
The PR brings hashboard up-to-date with hoon %140, adding and updating hooks and battery hashes. 

There's currently no command-line option to enable hashboard, the simplest way to run it is to just ensure that `u3_Host.ops_u.has` is set to `c3n` in `daemon/main.c`. I'll think about a better way to expose that control. I have also not tested switching to hashboard on a pier booted with only %fast-hint registration. Both are out of scope for this PR.

### DIFF
--- a/pkg/urbit/noun/jets.c
+++ b/pkg/urbit/noun/jets.c
@@ -1796,13 +1796,13 @@ _cj_minx(u3_noun cey, u3_noun cor)
 }
 
 static void
-_cj_print_tas(FILE* fh, u3_noun tas)
+_cj_print_tas(u3_noun tas)
 {
   c3_w  met_w = u3r_met(3, tas);
   c3_c* str_c = alloca(met_w + 1);
   u3r_bytes(0, met_w, (c3_y*)str_c, tas);
   str_c[met_w] = 0;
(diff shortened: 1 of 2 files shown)


pkg/urbit/noun/jets.c
```diff
@@ -1803,6 +1803,7 @@ _cj_print_tas(FILE* fh, u3_noun tas)
   u3r_bytes(0, met_w, (c3_y*)str_c, tas);
   str_c[met_w] = 0;
   fprintf(fh, "/%s", str_c);
+  u3l_log("/%s", str_c);
```
Fang-, line 1806: Looking at where this gets called, this is gonna go to _both_ strerr, and wherever 
`u3l_log` outputs to (also stderr?). Are you intentionally printing this twice?

joemfb: Good catch! The write to stderr is clobbered by the `u3l_log()` calls, but I didn't mean to leave it.

Fang- (approved):
Nothing else about this PR looks fishy, though note I didn't do the numerological deep-dive into these hashes...

PR #4171, opened 2020-12-15
interface: convert local state to hooks

tylershuster:
This is part one of a much more ambitious project, and basically serves to introduce the interface to the idea of using hooks for state.

This PR only converts the local state to hooks (who put baseHash in local state?) because performing a similar operation on "remote" (arvo) state will take a bit more refactoring, and it would be nice to have some groundwork upon which to build.

This converts the following properties to the new model, and removes all prop references to them: 
- Remote Content Policy
- Avatars Shown
- Nicknames shown
- Dark mode
- Background
- Omnibox Status

It exposes a new HOC: `withLocalState` for passing the appropriate properties to class-based components.

This would be a good place to discuss general patterns before I start on the larger, more comprehensive project.

There is currently a bug where profile overlays when clicked, position themselves relatively. I don't know why yet.

cc @liam-fitzgerald @tacryt-socryp


pkg/interface/src/views/components/Author.tsx
```diff
@@ -8,29 +8,29 @@ import OverlaySigil from "./OverlaySigil";
...
+  const { contacts, ship = '', date, showImage, group, api } = props;
+  const hideNicknames = useLocalState(state => state.hideNicknames);
   const history = useHistory();
   let contact;
   if (contacts) {
     contact = ship in contacts ? contacts[ship] : null;
   }
   const color = contact?.color ? `#${uxToHex(contact?.color)}` : "#000000";
-  const showNickname = !props.hideNicknames && contact?.nickname;
+  const showNickname = !hideNicknames && contact?.nickname;
```
liam-fitzgerald, line 33: This pattern should be refactored out into its own hook

pkg/interface/src/views/components/OverlaySigil.tsx
```diff
@@ -139,3 +133,5 @@ export default class OverlaySigil extends PureComponent<OverlaySigilProps, Overl
     );
   }
 }
+
+export default withLocalState(OverlaySigil);
```
liam-fitzgerald, line 137: It'd be nice if this was a seperate component that only passed down the props it needed

pkg/interface/src/logic/state/local.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { BackgroundConfig, RemoteContentPolicy } from "~/types/local-update";
+
+export type LocalState = {
+  hideAvatars: boolean;
+  hideNicknames: boolean;
+  remoteContentPolicy: RemoteContentPolicy;
+  dark: boolean;
+  background: BackgroundConfig;
+  omniboxShown: boolean;
+  suspendedFocus?: HTMLElement;
```
liam-fitzgerald, line 14: this should be an interface not a type

pkg/interface/src/views/apps/profile/components/lib/RemoteContent.tsx
```diff
@@ -47,13 +48,12 @@ export default function RemoteContentForm(props: RemoteContentFormProps) {
         } as FormSchema
       }
       onSubmit={(values, actions) => {
-        api.local.setRemoteContentPolicy({
-          imageShown: values.imageShown,
-          audioShown: values.audioShown,
-          videoShown: values.videoShown,
-          oembedShown: values.oembedShown,
+        setRemoteContentPolicy(state => {
```
liam-fitzgerald, line 51: `Object.assign(state, values)` maybe?

liam-fitzgerald (approved):
This looks really good. Can you talk about what needs to occur for remote state?

tylershuster:
Take a look at my post in `~wolref-podlex/foundation` "Discussions" "Modularization" for some high-level overview of the project.

The current problem is that Landscape's state is fundamentally tied to its API: https://github.com/urbit/urbit/blob/6499eb5fe0bd81c91f12d6f9ebcc6843b2ca7ac7/pkg/interface/src/logic/api/global.ts#L26-L30

Here's my current thoughts on how the specifics will be handled.

***

## useApi

There will still be a global API, but it will basically just be an instance of `window.channel`:

```js
import landscapeReducer from '~/logic/reducer';

const useApi = memoize(() => {
  const channel = new window.channel();
  channel.onMessage = landscapeReducer;
  return channel;
});
// You get the idea
```

## landscapeReducer

This handles moving all the data that gets received from the channel into the various stores. The idea behind this is that exactly how a client manages state is up to them, but the messages received from the back-end come in a predictable fashion, hence the external dependency on `@urbit/api`.

```js
const landscapeReducer = (data) => {
  if ('group-update' in data) {
    const groupState = useGroup.getState();
    useGroup.setState(groupState.set(data)); // Purported method, untested
  }
  ...
}
```

## useStore

Finally we reach the level of the stores, which you've already seen. At first I thought that the idea was that all the `*Action` methods on the `*Api` classes will turn into zustand state members. But with the mark conversion abstracted into another package, that's actually unnecessary. For example:

**Old:**
```js
api.hark.setMentions(mentions);
```

**New:**
```js
import { setMentions } from '@urbit/api/hark';
const api = useApi();
api.sendMessage(setMentions(mentions));
// Then the api gets the response back, and reduces the correct landscape state out of it
```

This does present an interesting case where there will be a period in which the API has sent information and has not received the new state. I'm open to patterns which allow us to temporarily mutate the local state until it can be confirmed against the new remote state, but I haven't thought of any yet.

pkg/interface/src/views/components/OverlaySigil.tsx
tylershuster: Do you have an example of this pattern?

pkg/interface/src/views/components/Author.tsx
tylershuster: Like...`useNickname(contact)`?

liam-fitzgerald: Yep

tylershuster:
To answer your questions about getmore and other methods that rely on state, I think the appropriate way to handle this is to make the methods that get put into the external library map strictly to Urbit's internals, and we can add utility methods that implement hooked state to provide extra utility like this￼

pkg/interface/src/views/components/OverlaySigil.tsx
liam-fitzgerald: It's not strictly necessary for this, given the infrequency with which the localState tends to change, but if we're serious about hookifying the whole state, then we could end up with a boatload of spurious rerenders. (Bridge currently has this problem in spades). Here's a nicer version of `withLocalState` that would cause less rerenders. Dunno if these types work, but you could definitely strongly type such a function. 
```typescript
function withLocalState<P, S extends keyof LocalState>(pickState: S[], Component: (compProps: P) => ReactNode) {
  return React.forwardRef((props: Omit<P,S>, ref) => {
    const localState = _.pick(useLocalState(), pickState);
    return <Component ref={ref} {...localState} {...props} />
  })
}
function withLocalState<P>(component: Component, relea

tylershuster: Ok, nice. I see what you mean. Zustand actually has a nice way of selectively using state members. I'll get this written up tonight.

pkg/interface/src/logic/state/local.tsx
tylershuster: If I do that I get this:
Screen Shot 2020-12-17 at 9 50 45 AM

How do I fix this?

tylershuster: https://github.com/pmndrs/zustand/issues/213

tylershuster:
Ok, feedback addressed:

- added `useShowNickname(contact: Contact): boolean` to utils
- implemented above hook
- allowed selective state picking in `withLocalState` HOC (would like some help abstracting it eventually into `withState(Component, State, members)`
- style nits

matildepark:
A change this sweeping and positive should go in sooner than later. I'd like @liam-fitzgerald to take a confirmation review here, firstly — 

Secondly, I'm still catching up on progress but FE work like this is a prime candidate for targetting release/next-js so it gets auto-deployed.  Standby on retargetting or merging while I catch up on matters and ping an update here...

liam-fitzgerald:
@tylershuster please don't force push after a review, makes github forget what i've reviewed.

tylershuster:
Roger that. I only changed the items mentioned above, if it helps at this point

matildepark:
next-js is current with master. I think this can be the first thing to go in there, but let #4198 get merged first. Seems like a string append won't take us long.

tylershuster:
...?

matildepark:
You know? Fair.

PR #4173, opened 2020-12-15
vere: set arvo verbosity based on presence of -v

### LINKED ISSUE #4107: v1.0-rc1 binary: Verbose Mode Stays On
**Description**
The `-v` flag puts the ship permanently into verbose mode, even when exiting the binary and restarting without `-v`.

**To Reproduce**
1. Get the v1.0-rc1 binary
2. Run a fakeship with `-v`: `./urbit -F zod -v`
3. Exit the ship, and run without `-v`

**Expected behaviour**
The 2nd time running, verbosity should be off. Instead, it continues.

 - OS: macOS
 - v1.0-rc1, UrbitOS v2.2
 - `a1ssv`

joemfb:
... instead of simply toggling whatever state was previous set. Fixes #4107


Fang- (approved):
Good fix!

PR #4176, opened 2020-12-16
hoon: crash with valid tank on %eror

### LINKED ISSUE #4169: Trying to print a bad tank crashes the urbit
Trying to evaluate the following expression in dojo (or including it in an app, lib, or elsewhere) crashes the urbit process:

```hoon
|%  ++  x  1  ++  x  2  --
```

```
fish: 'pkg/urbit/build/urbit nec' terminated by signal SIGABRT (Abort)
newt: write failed broken pipe
serf: pier error: broken pipe
```

I vaguely remember this resulting in a legible complaint before. Certainly haven't seen it crash like this before, but maybe I just never hit it?

joemfb:
Fixes #4169 (or, at least, that specific instance of it). We may still want to virtualize `$tank` printing in the king.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -8363,7 +8363,7 @@
         [%bust *]  ~(example ax %base p.gen)
         [%ktcl *]  ~(factory ax p.gen)
         [%dbug *]   q.gen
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Nice. Works on my machine!

PR #4178, opened 2020-12-16
zuse: add backward tap (+bap) to +ordered-map

tacryt-socryp:
This can be used to make #4166 run much more efficiently.


philipcmonk (approved):
TIL "bap" is a scrabble word.  Implementation looks fine, and seems like a good function to have (maybe for queues as well?).

PR #4181, opened 2020-12-17
azimuth: add local contracts

liam-fitzgerald:
Adds local contracts to azimuth.hoon to support local testnets. These addresses are static iff you use the same seed and deploy script from bridge.

### DIFF
--- a/pkg/arvo/lib/azimuth.hoon
+++ b/pkg/arvo/lib/azimuth.hoon
@@ -79,15 +79,7 @@
     ::
     ::  contract addresses
     ++  contracts  mainnet-contracts
-    ++  local-contracts
-      |%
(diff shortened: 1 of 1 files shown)


Fang- (commented):
Fine change! But I'd put these below the ropsten arm in this core, and respect the styling/ordering used for what's already there. (Also you might want to add `+public` for completeness.)

pkg/arvo/lib/azimuth.hoon
```diff
@@ -79,6 +79,7 @@
     ::
     ::  contract addresses
     ++  contracts  mainnet-contracts
+
```
Fang-, line 82: Hate to be that guy asking for a second round of style changes, but...

Fang- (approved):
Actually, I should just expedite this.

PR #4192, opened 2020-12-18
interface: fix missing archiveBox by removing Archive view

### LINKED ISSUE #4187: Unable to archive notifications
**Describe the bug**
Archiving a notification fails on a js error. The urbit-side update seems to happen, but something breaks in the ui.

**To Reproduce**
1. Go to Notifications
2. Click on Archive
3. See error

**Expected behavior**
Clicking archive and it, uh, archiving

**Screenshots**
Screen Shot 2020-12-18 at 9 10 03 AM

tacryt-socryp:
This fixes #4187 by removing the "archive" view in notifcations and the references to it in the reducers. The archive functionality had been partially removed from the frontend, but not all the way, which was causing JS errors when you tried to click the archive button on a notification. This resolves that issue and has been tested to work on `~tacryt-socryp`.

@philipcmonk I'd like to cut a release to hotfix this issue upon having this PR approved. Would you have time to chat? I'm new to doing these deploys and would like to get it right.

### DIFF
--- a/pkg/interface/src/views/apps/notifications/notification.tsx
+++ b/pkg/interface/src/views/apps/notifications/notification.tsx
@@ -100,7 +100,7 @@ function NotificationWrapper(props: {
         </StatelessAsyncAction>
         {!props.archived && (
           <StatelessAsyncAction name={time.toString()} onClick={onArchive} backgroundColor="transparent">
-            Read
+            Dismiss
           </StatelessAsyncAction>
         )}
       </Row>
--- a/pkg/interface/src/views/apps/notifications/notification.tsx
+++ b/pkg/interface/src/views/apps/notifications/notification.tsx
@@ -100,7 +100,7 @@ function NotificationWrapper(props: {
         </StatelessAsyncAction>
         {!props.archived && (
           <StatelessAsyncAction name={time.toString()} onClick={onArchive} backgroundColor="transparent">
-            Archive
+            Read
           </StatelessAsyncAction>
         )}
       </Row>
--- a/pkg/interface/src/logic/api/hark.ts
+++ b/pkg/interface/src/logic/api/hark.ts
@@ -196,12 +196,10 @@ export class HarkApi extends BaseApi<StoreState> {
     });
   }
 
-  getMore(archive = false) {
-    const offset = this.store.state[
-      archive ? 'archivedNotifications' : 'notifications'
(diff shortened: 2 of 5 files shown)


urcades (approved):
Big ol' stamp!

philipcmonk:
I'm going to put out a next-sys release shortly, I can include this if you want.  I'm happy to show you how to release stuff if you prefer, though.

tacryt-socryp:
If you can include this in the next-sys release, that would be perfect. Happy to jump on a call with you while you're going through the release process and learn some stuff though.

pkg/interface/src/views/apps/notifications/notification.tsx
```diff
@@ -100,7 +100,7 @@ function NotificationWrapper(props: {
         </StatelessAsyncAction>
         {!props.archived && (
           <StatelessAsyncAction name={time.toString()} onClick={onArchive} backgroundColor="transparent">
-            Archive
+            Read
```
Fang-, line 103: This might be a bit confusing, since the blue dot also indicates "read" state, but is separate from this. "Dismiss" might be more appropriate?

Fang- (approved):
This looks like it tears out various things related to archived notifications, yes.

I still see a `archivedNotifications` in the store, some `isArchive` logic that will presumably always be false, and a `partition`ing of notifications into archived and unarchived. Unclear what of that is truly unnecessary now, but if any of those jump out as clean-up-able, probably just get rid of them.

Also a small comment about the new button phrasing, but otherwise this looks good.

pkg/interface/src/views/apps/notifications/notification.tsx
urcades: Oh yes, I should have commented here — let's definitely go with dismiss.

I mentioned this in the comparison screens I had sent Liam, which I can repost here:

Before:
2020 12 17 00 53 00-Screen Shot 2020-12-16 at 7,52,37 PM

After:

2020 12 17 00 53 05-Screen Shot 2020-12-16 at 7,52,52 PM

tacryt-socryp: I can change the name to Dismiss. We should clean up the attributes in the store / etc when the archival functionality is removed from `%hark-store` itself, which is slated to occur early in Q1.

PR #4195, opened 2020-12-18
gall: don't molt on vega

philipcmonk:
We can't molt until clay has gotten its pork or else we'll build the old
app against the new kernel.  This ignores vegas, since we should get a
notification from clay on /sys/lyv.


belisarius222 (approved):
LGTM, hope it works

PR #4197, opened 2020-12-19
interface: document title

tylershuster:
This changes the name of the app across the board from "OS1" to "Landscape."

It also changes the document title as you navigate around the app, including surfacing notification counts.

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -1,5 +1,4 @@
 import React, { useState } from 'react';
-import Helmet from 'react-helmet';
 import styled from 'styled-components';
 
 import { Box, Row, Icon, Text } from '@tlon/indigo-react';
@@ -14,6 +13,7 @@ import ModalButton from './components/ModalButton';
 import { writeText } from '~/logic/lib/util';
 import { NewGroup } from "~/views/landscape/components/NewGroup";
 import { JoinGroup } from "~/views/landscape/components/JoinGroup";
+import { Helmet } from 'react-helmet';
 
 const ScrollbarLessBox = styled(Box)`
   scrollbar-width: none !important;
@@ -28,8 +28,8 @@ export default function LaunchApp(props) {
 
   return (
     <>
-      <Helmet>
-        <title>OS1 - Home</title>
+      <Helmet defer={false}>
+        <title>{ props.notificationsCount ? `(${String(props.notificationsCount) }) `: '' }Landscape</title>
       </Helmet>
       <ScrollbarLessBox height='100%' overflowY='scroll'>
         <Welcome firstTime={props.launch.firstTime} api={props.api} />
--- a/pkg/interface/src/views/apps/term/app.js
+++ b/pkg/interface/src/views/apps/term/app.js
@@ -46,8 +46,8 @@ export default class TermApp extends Component {
   render() {
     return (
       <>
-        <Helmet>
-          <title>OS1 - Terminal</title>
+        <Helmet defer={false}>
+          <title>{ props.notificationsCount ? `(${String(this.props.notificationsCount) }) `: '' }Landscape</title>
         </Helmet>
         <Box
           height='100%'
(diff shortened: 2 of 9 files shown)


tylershuster:
Fixes https://github.com/urbit/landscape/issues/203

tylershuster:
addresses https://github.com/urbit/landscape/issues/195

tylershuster:
addresses https://github.com/urbit/landscape/issues/201

matildepark:
If you pull out the eyre change as a separate PR we can deploy this to next-js.

tylershuster:
Great, should I squash now?

matildepark:
Yeah.

pkg/interface/src/views/apps/launch/app.js
```diff
@@ -29,7 +29,7 @@ export default function LaunchApp(props) {
   return (
     <>
       <Helmet>
```
matildepark, line 31: Needs `defer={false}`

pkg/interface/src/views/apps/term/app.js
```diff
@@ -47,7 +47,7 @@ export default class TermApp extends Component {
     return (
       <>
         <Helmet>
```
matildepark, line 49: Needs `defer={false}`

pkg/interface/src/views/landscape/components/GroupsPane.tsx
```diff
@@ -133,28 +134,36 @@ export function GroupsPane(props: GroupsPaneProps) {
...
-              selected={appPath}
-              {...props}
-              baseUrl={baseUrl}
-            >
-              <UnjoinedResource
-                graphKeys={props.graphKeys}
-                notebooks={props.notebooks}
-                inbox={props.inbox}
+            <>
+              <Helmet>
```
matildepark, line 146: Needs `defer={false}`

pkg/interface/src/views/landscape/components/GroupsPane.tsx
```diff
@@ -186,20 +195,26 @@ export function GroupsPane(props: GroupsPaneProps) {
...
-                p='4'
-              >
-                <Box p="4"><Text fontSize="0" color='gray'>
-                  {description}
-                </Text></Box>
-              </Col>
-              {popovers(routeProps, baseUrl)}
-            </Skeleton>
+            <>
+              <Helmet>
```
matildepark, line 201: Needs `defer={false}`

pkg/interface/src/views/landscape/components/Resource.tsx
```diff
@@ -34,47 +34,58 @@ export function Resource(props: ResourceProps) {
...
+    }
+  }
   return (
-    <Switch>
-      <Route
-        path={relativePath("/settings")}
-        render={(routeProps) => {
-          return (
+    <>
+      <Helmet>
```
matildepark, line 45: Needs `defer={false}`

pkg/interface/src/views/landscape/index.tsx
```diff
@@ -78,71 +77,76 @@ export default class Landscape extends Component<LandscapeProps, {}> {
...
-            const baseUrl = `/~landscape${groupPath}`;
-            const ws: Workspace = { type: 'group', group: groupPath };
-
-            return (
-              <GroupsPane workspace={ws} baseUrl={baseUrl} {...props} />
-            )
-          }}/>
-        <Route path="/~landscape/home"
+      <>
+        <Helmet>
```
matildepark, line 81: Needs `defer={false}`

matildepark (commented):
Careful — we want the number to update when we're out of the tab. Also seems like a lot of different places when the title doesn't change anywhere — why `landscape/index.tsx` *and* its child components?

tylershuster:
As for why both the parent and children:

On the home page we want it to say `(1) Landscape`
But on child pages we want `(1) Urbit Community`

and so forth

tylershuster:
Ok, deferments added and tested

pkg/interface/src/views/apps/launch/app.js
matildepark: What about this one?

tylershuster: You're too quick

PR #4220, opened 2021-01-05
notifications: clear state before processing initial update

ixv:
If a subscription restarts, hark will send an %unread update, which functions as an initial-state update, but the reducer just treats it as a normal update, adding unread counts rather than setting them. This PR fixes that by clearing the notifications state before processing the update.


liam-fitzgerald (commented):
LGTM!

liam-fitzgerald (approved):

PR #4198, opened 2020-12-19
Add GitHub Action & Workflow for glob + OTA

jtobin:
Adds an action that globs and OTAs relevant changes to a target ship, plus a workflow that invokes the action (deploying to `~lomlyx-lopsem-nidsut-tomdun`) whenever `release/next-js` is pushed or PR'd to.

Once this is merged to master, it can be activated on `release/next-js` by simply merging master into that.

cc @liam-fitzgerald @tacryt-socryp @vvisigoth @galenwp

PR #4222, opened 2021-01-05
meta: add landscape development stream information

matildepark:
As of #4198, we auto-deploy release/next-js to opted-in ships on network (I'm getting final word on the specific %desk before announcing this). 

Thus, merging into release/next-js is part of deploying to livenet (and really, merge conflicts should be pretty infrequent, since the `master <-> release/next-js -> release/next-userspace` flow is more or less unidirectional for cross-branch changes).

This includes an advisory to include a merge to next-js as part of deploy. We could sidestep this if GitHub Actions let us auto-merge master into next-js, but unsure about that yet.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -180,9 +180,9 @@ new fakezod with `urbit -F zod -B bin/solid.pill -A pkg/arvo`).  Run
 `:glob|make`, and this will output a file in `fakezod/.urb/put/glob-0vXXX.glob`.
 
 Upload this file to bootstrap.urbit.org, and modify `+hash` at the top of
-`pkg/arvo/app/glob.hoon` to match the hash in the filename of the `.glob` file. 
+`pkg/arvo/app/glob.hoon` to match the hash in the filename of the `.glob` file.
 Amend `pkg/arvo/app/landscape/index.html` to import the hashed JS bundle, instead
-of the unversioned index.js. Do not commit the produced `index.js` and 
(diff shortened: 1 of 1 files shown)


philipcmonk:
This applies to anything that has js changes, and it only copies the js changes to that ship, right?

matildepark:
As far as I'm aware, it globs and deploys like any other — the dev ship becomes your sponsor.

philipcmonk:
> Thus far this only includes `release/next-js`, which can only include livenet-compatible JavaScript changes.

Is this "if there are any js changes, they must be livenet-compatible" or "only js changes are allowed and they must be livenet-compatible"

matildepark:
Ah, okay:

- `release/next-js` is the branch to target when you are making front-end specific changes that do not necessitate an accompanying back-end change
- Full-stack features + changes live on `release/next-userspace` and follow the 6 week sprint pattern for release
- JS hotfixes may still go straight into master, but can target next-js depending on severity, and JS-specific features target next-js

next-js may be released more often, but that's still a matter of discussion for us. I will try to keep merging next-js into next-userspace.

philipcmonk (approved):
Cool.  Then, as you wrote, the main thing to do is push master changes to release/next-js soon after pushing to master.

matildepark:
@jtobin thinks an auto-merge is an easy action to incorporate so I would stand by before merging ... we can simplify the process here.

jtobin (approved):
LGTM, and note that #4232 should handle the desired auto-merge.

PR #4226, opened 2021-01-06
interface: fixes bug where profile overlay would crash due to hook use

tylershuster:
my bad.

this passes a second parameter to `useShowNickname` for manual override of `hideNicknames` state variable. lmk if there's a better way to do this.


matildepark:
Mind force-pushing this branch to force a re-test?

matildepark:
You want commit hygiene here anyway.

matildepark:
Excellent, thank you.

matildepark (approved):
Works as specified.

PR #4229, opened 2021-01-06
interface: better checking if channel should be marked unread

tylershuster:
Basically, check if the line is visible after the initialization period has passed, after which point we should know if the blue line is visible. Worked in testing for me.

Also some style nits and removed `console.log`s

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
@@ -92,7 +92,6 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
       if(this.props.scrollTo) {
         this.scrollToUnread();
       }
-
       this.setState({ initialized: true });
     }, this.INITIALIZATION_MAX_TIME);
   }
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
```diff
@@ -136,6 +135,10 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
       this.setState({ fetchPending: false });
     }
 
+    if (this.state.initialized !== prevState.initialized) {
+      this.dismissIfLineVisible();
+    }
```
tylershuster, line 140: This is the relevant bit

matildepark: Won't this fire if state.initialized is `false` and prevState's is `true`?

tylershuster: I don't think that would ever happen, but if there's some weirdness with switching chat windows, I changed it to make sure

matildepark:
Haven't stamped this as of yet because I can't quite replicate the problem. On my ship I see the new messages banner until I scroll up while the line is on screen, then it vanishes. Is this case best tested with "the chat has less than 20 messages" case?

tylershuster:
Does that behavior still occur with this patch applied?

matildepark:
Actually, yeah, it never gets marked as read and there's no line visible.

tylershuster:
Needs more systematic attention, and I deleted this fork.

PR #4232, opened 2021-01-06
Add a GitHub workflow that auto-merges master to release/next-js on push

jtobin:
Adds a workflow that, on any push to master, merges master to release/next-js.

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -2,7 +2,7 @@ name: merge
 on:
   push:
     branches:
-      - 'jt/merge-workflow'
+      - 'master'
 jobs:
   merge-to-next-js:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


matildepark:
So what happens if we merge next-js into master?

jtobin:
grey goo? :-)

Not really.  I haven't tested that behaviour, but it will just try to merge master back into next-js as per usual, creating a new merge commit on release/next-js if there's a material one to make.  And if not, it won't do anything ('already up to date').

matildepark (approved):

matildepark:
@jtobin The automated merge commit doesn't seem to initiate a glob action? Any ideas?

jtobin:
Hmm, apparently a workflow that's run using `GITHUB_TOKEN` [can't trigger other workflows](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token).

The solution seems to be to use a personal access token instead.  So, I'll probably create a custom bot account for this type of thing, assign it the required write permissions, and generate / use a token for that.

PR #4233, opened 2021-01-06
tally: add generator for measuring group activity

Fang-:
Scrapes groups and graphs for member counts and recent activity respectively,
and uses that to print out a "tally" of activity for all local groups and their
channels. Since this output is intended to be shared with other parties,
anonymizes resource identifiers by default.

Example output, from a ship hosting a couple low-activity groups.

```
> +tally

tallied your activity score! find the results below.
to show non-anonymized resource identifiers, +tally |
counted from groups and channels that you are hosting.
groups are listed with their member count.
channels are listed with activity from the past week:
  - amount of top-level content
  - amount of unique authors

the date is ~2021.1.6..17.41.04..0daa
you are in 3 group(s).
you are hosting 3 group(s):
group, 2627f969, 103
 chan, 574bcec1, chat, 0, 0
 chan, 6016c0a5, chat, 7, 1
 chan, 2a5141ef, chat, 11, 4
 chan, 3d4fe01a, chat, 0, 0
 chan, 63261571, chat, 0, 0
 chan, 6275e690, chat, 25, 5
 chan, 39b03c56, chat, 31, 6
 chan, 794f43a6, chat, 0, 0
group, 5db104dd, 10
 chan, 69e410fc, chat, 2, 2
group, 3f5d063f, 2
```

This isn't the prettiest thing I ever wrote, but it gets the job done.

As requested by @vvisigoth. Rationale is that we want some way for group hosts to opt in to sharing activity metrics, but don't want to auto-run anything on people's ships, and don't want to leak any details outside of that activity. Having copy-paste-able output like this will make that easier, and will also let hosts peek at a rudimentary activity overview without ever sharing anything.

### DIFF
--- a/pkg/arvo/gen/tally.hoon
+++ b/pkg/arvo/gen/tally.hoon
@@ -1,0 +1,122 @@
+/-  gr=group, md=metadata-store, ga=graph-store
+/+  re=resource
+!:
+:-  %say
+|=  $:  [now=@da eny=@uvJ =beak]
+        args=?(~ [shy=? ~])
+        ~
+    ==
+::
+=/  shy=?   ?~(args & shy.args)
+=*  our=@p  p.beak
+::
+|^
+=;  out=(list @t)
+  :-  %tang
+  %-  flop  ::NOTE  tang is bottom-up
+  :*  ''
+      'tallied your activity score! find the results below.'
+    ::
+      ?:  shy
+        'to show non-anonymized resource identifiers, +tally |'
+      'showing plain resource identifiers, share with care.'
+    ::
+      'counted from groups and channels that you are hosting.'
+      'groups are listed with their member count.'
+      'channels are listed with activity from the past week:'
+      '  - amount of top-level content'
+      '  - amount of unique authors'
+      ''
+      (snoc out '')
+  ==
+::  gather local non-dm groups, sorted by size
+::
+=/  groups=(list [local=? resource:re members=@ud])
+  %+  murn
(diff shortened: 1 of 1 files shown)


matildepark:
Seems to me like we want this out sooner than `release/next-userspace`? It is a slower moving release branch, currently timed for early Feb.

Fang-:
Sure, this can target another branch that goes out earlier.

vvisigoth:
Just want to confirm: are those group IDs hashes or similar? So that we can deduplicate across different ship reports?

Fang-:
Yes, they're just the mug hash of the group/channel ids, so they'll be unique-per-group as you want. ...But we only list locally hosted things (as per your spec, iirc), so you'll only see them show up again for new reports from the same ship.

vvisigoth:
Ah, can we put the mug of the groups that I'm in as well?

A

—
~poldec-tonteg
http://urbit.org

Fang-:
@vvisigoth done! Tiny example case:

```
the date is ~2021.1.8..21.51.37..d60d
you are in 3 group(s):
64095f07, 2ce8c9b5, 3ac1e2f0
you are hosting 3 group(s):
group, 3ac1e2f0, 1
group, 2ce8c9b5, 1
group, 64095f07, 1
 chan, 7ed5caef, chat, 3, 1
```

vvisigoth (approved):
j'approve

matildepark:
Base this off master, please. @Fang-

PR #4234, opened 2021-01-06
graph-delete-thread: remove entry from metadata-store, then remove entry from metadata-hook

### LINKED ISSUE #4160: Channel deletion does not get pushed to members
Created a channel in a group by mistake. Deleted it, and it disappeared from the sidebar on the host of the group & channel. But for members of the group, the channel remains in the sidebar.

tacryt-socryp:
Fixes #4160.

### DIFF
--- a/pkg/arvo/ted/graph/delete.hoon
+++ b/pkg/arvo/ted/graph/delete.hoon
@@ -33,7 +33,7 @@
   (pure:m (need ugroup))
 ::
 ++  delete-graph
-  |=  rid=resource
+  |=  [group-rid=resource rid=resource]
   =/  m  (strand ,~)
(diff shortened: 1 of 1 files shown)


matildepark:
I feel like a hotfix like this should target master and go out once it's stamped — granted we have 100% certainty of the fix.

tacryt-socryp:
I'm still testing, but I'm happy to rebase it to master once I'm done.

tacryt-socryp:
@matildepark Tested and reproduced the error, then applied this commit and tested and confirmed that the error is resolved. Am confident that this works and should be applied as a hotfix to master.

liam-fitzgerald (approved):

Issue #4211, opened 2021-01-02
graph-join thread retries too hard


dchem:
**Describe the bug**
`urbit-worker` is taking 99-100% of the CPU. This behavior started on its own three days ago.

As a consequence, landscape and group talks are either painfully slow or times out.

Running urbit with `urbit -v <shipname>` shows endless stream of:

```
["" %unix %wake /b/behn ~2021.1.2..20.29.55..e860]                                                                                                                                                                                                                                        
["|" %give %behn %wake i=/gall/use/spider/0wOO~2Q/~<redacted>/thread/<redacted>/wait/~2021.1.2..20.29.54..271a t=~[/dill //term/1]]                                                                                                                             
["||" %pass [%gall %b] [%wait /use/spider/0wOO~2Q/~<redacted>/thread/<redacted>/wait/~2021.1.2..20.29.56..6860] ~[/dill //term/1]] 
```

**To Reproduce**
Not sure I am able to reproduce it.

**Expected behaviour**
`urbit-worker` should not be consuming all available CPU at all times.

**System (please supply the following information, if relevant):**

OS:
```
~$ uname -a
Linux <redacted> 5.4.0-48-generic #52-Ubuntu SMP Thu Sep 10 10:58:49 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

Urbit:
```
~$ urbit -R
urbit 1.0
gmp: 6.2.0
sigsegv: 2.12
openssl: OpenSSL 1.1.1g  21 Apr 2020
libuv: 1.40.0
libh2o: 0.13.4
lmdb: 0.9.25
curl: 7.72.0
argon2: 0x13
```

Base Hash:
```
> +trouble
[ [%base-hash ~[0ve.e60ss.an2vu.33q7c.86vqp.tg29i.atsqk.871pj.ihukp.1cqif.74m9v]]
  [%sour-hash 0ve.e60ss.an2vu.33q7c.86vqp.tg29i.atsqk.871pj.ihukp.1cqif.74m9v]
  [%home-hash <redacted>]
  [%kids-hash 0ve.e60ss.an2vu.33q7c.86vqp.tg29i.atsqk.871pj.ihukp.1cqif.74m9v]
  [%glob-hash 0v5.hvt1e.ie7it.b7i7l.1r7jj.dn9ib %done]
  [%our ship=~<redacted> point='<redacted>' life=[~ 2] rift=[~ 1]]
[...]

PR #4237, opened 2021-01-06
spider: prefix thread ids with thread name

Fang-:
For thread ids generated by spider itself, includes the name of the thread,
for convenient identification.

Prompted by debug struggles in #4211.


philipcmonk (approved):
This is good.  Tempting to include whether it was started by http directly.

PR #4238, opened 2021-01-06
term: use 'this' in notificationscount

matildepark:
cc @tylershuster 

#4197 checks for props.notificationsCount in the interface, but Term doesn't pull props out of `this`.

Fixes urbit/landscape#221


tylershuster:
d'oh

tacryt-socryp (approved):

PR #4239, opened 2021-01-06
graph: make the join thread back off

Fang-:
Pushes the pathological case back a bit, but doesn't make it stop trying.

This compiles, but I haven't been able to actually run the specific code I changed. @liam-fitzgerald, @tacryt-socryp, please advise.

This is a _partial_ fix for #4211. The full fix is making this also stop trying at some point, but I'm not sure what the right behavior for that case would be.

### DIFF
--- a/pkg/arvo/ted/graph/join.hoon
+++ b/pkg/arvo/ted/graph/join.hoon
@@ -29,13 +29,15 @@
   ^-  form:m
   =/  pax
     (en-path:resource rid)
+  =/  hold=@dr  ~s0..8000
   |-  ^-  form:m
   =*  loop  $
   ;<  u-group=(unit group)  bind:m  
     (scry:strandio ,(unit group) (weld /gx/group-store/groups (snoc pax %noun)))
   ?^  u-group
     (pure:m ~)
-  ;<  ~  bind:m  (sleep:strandio `@dr`(div ~s1 2))
+  ;<  ~  bind:m  (sleep:strandio hold)
+  =.  hold  (min (mul hold 2) ~m5)
   loop
 ::
 ++  wait-for-md
@@ -44,13 +46,15 @@
   ^-  form:m
   =/  pax
     (en-path:resource rid)
+  =/  hold=@dr  ~s0..8000
   |-  ^-  form:m
   =*  loop  $
   ;<  groups=(set path)  bind:m  
     (scry:strandio ,(set path) /gy/metadata-store/group-indices) 
   ?:  (~(has in groups) pax)
     (pure:m ~)
-  ;<  ~  bind:m  (sleep:strandio `@dr`(div ~s1 2))
+  ;<  ~  bind:m  (sleep:strandio hold)
+  =.  hold  (min (mul hold 2) ~m5)
   loop
 --
 ::
--- a/pkg/arvo/ted/graph/join.hoon
+++ b/pkg/arvo/ted/graph/join.hoon
@@ -31,13 +31,14 @@
     (en-path:resource rid)
   =/  hold=@dr  ~s0..8000
   |-  ^-  form:m
+  ?>  (lte hold ~m5)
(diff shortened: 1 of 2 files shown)


matildepark:
Hotfix, should target master (until I come up with a "short-term fixes" release pattern). This is pretty important to fix anyway

liam-fitzgerald (approved):

Fang-:
Guess I should've marked this as draft huh.

> This compiles, but I haven't been able to actually run the specific code I changed. @liam-fitzgerald, @tacryt-socryp, please advise.

matildepark:
We just have very supportive reviewers!

joemfb:
I'm a fan of backoff ... but wouldn't it be better to make "attempting to join" an explicit state, so that subsequent join requests no-op? This state could (later) be surfaced in the UI as well.

Fang-:
That's definitely the better fix, if landscape team wants to pick that up. (They should!) I'm just here for the quick fix for an unacceptable situation.

matildepark:
> but wouldn't it be better to make "attempting to join" an explicit state, so that subsequent join requests no-op?

Isn't this what #4106 is for

tacryt-socryp:
I think the best approach for now is to include a one-time "kill all threads in spider" in +on-load, and ensure that we *do* stop retrying after an appropriate period of time (one week?)

belisarius222:
I agree with Joe.  I think this problem stems ultimately from a lack of idempotence -- clicking "join" should cause a piece of state to be updated, once, that we are now attempting to join.  Ames already does backoff; that should not be recapitulated at a higher layer.

It might be desirable to also set a timer after which the user will be notified that their join attempt has not yet succeeded.  I suspect we actually don't want this, and instead want a page where we can view pending operations, or a grayed out title in the sidebar, which brings the user to a "trying to reach ~so-and-so since yesterday" status page.  This would properly reflect the underlying system state without introducing unnecessary and fraught retry states in userspace.

If we do want a user notification timer, though, there should be at most one timer per join request, so it should be part of a separate thread that could be killed and restarted independently of the join request, which is persistent and not really cancelable -- the best that Ames could do there is to enqueue a leave request to be sent just after the join request, and I think that would require careful synchronization with userspace state.

This timer thread would need to be kicked off by an app that remembers whether it's trying to join or not, so that if someone kills all threads, the app can spawn a new thread to restart the timer.  

Another way of thinking about this is to consider the duration of a transaction.  It's probably good practice for any thread spawned by a landscape web request to complete within a single Arvo event -- otherwise, idempotence is more likely to be broken, since a subsequent request could spawn a second copy of the same thread.

A web-spawned thread that could past multiple events should at least be written with an informal proof of idempotence.  Or maybe :spider could be modified to spawn web threads in a way that enforces these invariants?  Not sure, but could be worth investigating.  Spider could store the set of ourstanding web requests and ignore duplicate requests, or something like that.

matildepark:
Since it seems like this is potentially prohibiting users from using Landscape and otherwise tanking ships with enough threads running, can we agree on a solution for the short-term?

liam-fitzgerald (approved):

matildepark:
With #4303 filed I'd like to target this for a deploy start of week. @tacryt-socryp @liam-fitzgerald If we can get just more eyes confirming resolution of issue that would be great.

tacryt-socryp:
Again, this works, but "pushing back the pathological case" doesn't seem to be much help, when we can clearly see that it would be best for this to stop retrying after a period of time.

tacryt-socryp:
This looks good to me as a stop-gap. Thanks for updating it! I built the file and looked through its "termination conditions". Seems reasonable and like a small/safe change.

tacryt-socryp (approved):
LGTM

matildepark:
@liam-fitzgerald Probably don't *directly* supersede alternative solutions next time but I will merge this!

Issue #4202, opened 2020-12-23
Chat: newlines are being turned into spaces again


mike-cr:
**Describe the bug**
if you compose a message like 
```
1
2
3
```
it becomes `1 2 3` when you send

**To Reproduce**
1. Go to chat
2. compose multi-line message
3. Send
4. Observe it's now one line

**Expected behavior**
newlines should be preserved

**Desktop (please complete the following information):**
 - OS: arch linux
 - Browser qutebrowser
 - Base hash 74m9v

matildepark:
For anyone watching, note that the new lines are indeed preserved — they're just being flexboxed into one line on render. Annoying.

matildepark:
Post #4252 just wanted to write the findings here.

So, our linear graphs have multi-content graph nodes. And this basically comes down to whether newlines should be within one `text` message or whether we should separate newline'd content with sibling `text` messages.

Right now on livenet we cut up messages per newline in subsequent `text` messages; but if you, say, mention someone and keep typing, that'll be one text message, one mention message, and another text message, intended as a single string.

My hypothesis is that we should stop splitting content this way and just insert newlines in the text message so we assume each message is one long string. We would have to ensure it renders the newlines properly on CLI.

tacryt-socryp:
This seems sane. I don't think there's a need for multiple contiguous %text elements. Best to use it for different categories of content than for formatting.

tylershuster:
I agree that each message should retain in a continuous entity when it is passed to the back end. However, I have opined elsewhere that graph store makes it sensible to add the different content types as sub nodes of the initial node, which would allow us to retain the special formatting￼. Does this seem sane for the chat application?￼

tylershuster:
```
{
  text: '~zod look at http://google.com',
  children: [
    { mention: 'zod' },
    { url: 'http://google.com' }
  ]
}
```

tylershuster:
This might even allow for a "gallery" style view of a chat when you could see all the sub nodes with "attachment" type

tacryt-socryp:
I'd prefer not to do nesting that way, and make the following a [%mention, %text, %url].

matildepark:
I’m hesitant to change up the graph validators if we can help it. So far this change it just front end.

tylershuster:
I'll take a stab at the flex implementation

matildepark:
This is actually a tokenize message change I think. 

— 
~haddef-sigwen
https://urbit.org

Fang-:
[...]

PR #4248, opened 2021-01-08
publish: renders line breaks

### LINKED ISSUE #4245: publish: line breaks ignored in comments
When you write a comment with line breaks, they are ignored when viewing that comment. They still exist when you go to edit the comment.

tylershuster:
fixes #4245

This is a bit of a novel approach — it reduces the content in `<MentionText>` down to a single child `<RichText>` whose markdown-formatted content contains specially-formatted (but hopefully logical) links, which are then parsed by `<RichText>` into actual `<Mention>` objects.

**Before:**
image

**After:**
image

### DIFF
--- a/pkg/interface/src/views/components/MentionText.tsx
+++ b/pkg/interface/src/views/components/MentionText.tsx
@@ -22,25 +22,16 @@ export function MentionText(props: MentionTextProps) {
   const { content, contacts, contact, group } = props;
 
   return (
-    <>
-      {_.map(content, (c, idx) => {
+    <RichText contacts={contacts} contact={contact} group={group}>
+      {content.reduce((accum, c) => {
         if ("text" in c) {
-          return (
-            <RichText
-              inline
-              key={idx}
-            >
-              {c.text}
-            </RichText>
-          );
+          return accum + c.text;
         } else if ("mention" in c) {
-          return (
-            <Mention key={idx} contacts={contacts || {}} contact={contact || {}} group={group} ship={c.mention} />
-          );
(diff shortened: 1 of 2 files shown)


matildepark:
Oh great! Fortuitous timing, I was going to look into #4202 and this is pseudo-related I guess (vis-a-vis proper rendering of these multi-child graph nodes with line breaks and mentions)

tylershuster:
Interesting, I don't believe that chat uses the same renderer but this may get us closer to The One True Renderer

matildepark:
It doesn't (though ChatMessage punts some rendering to the Mention component). I just mean Code Inspo. Reviewing in a sec

tylershuster:
Another possible relation: https://github.com/urbit/landscape/issues/200

matildepark (approved):
I thought this introduced a regression where remotePolicy isn't being passed to the comments (YouTube always renders as links), but that might be a regression due to the state hooks.

tylershuster:
Sec, There's one fiddly thing I want to try

matildepark:
I'll wait!

tylershuster:
Ok, this is just a slightly different approach. Before I was using `[~sampel-palnet](~sampel-palnet)` as a placeholder to be parsed out, now it's just `[~sampel-palnet]`. This is what remark calls a `linkReference` (https://github.com/remarkjs/remark/issues/258#issuecomment-305873317) and I think because we don't really support this functionality anywhere, the bonus we get in readability if something goes wrong, like the below image, is a more graceful fallback. (note this problem is unrelated and occurred before this PR)

image

PR #4252, opened 2021-01-08
chat: separate text nodes with line breaks

matildepark:
Getting creative with flexbox to break up text nodes as line-broken content while still flexing mentions onto the same line. See #4202.

Screen Shot 2021-01-08 at 3 17 34 PM

cc @tylershuster — mind looking through chat logs and mentions inside comments in other apps to see if this introduces any quirks?


tylershuster:
Looking now

tylershuster:
Er, this could be better with stuff like this:

**Entered**
```
Hello ~marpem
It is I, ~radbur-sivmus, your old
nemesis
```

**next-js**
image

**This commit**
image

But maybe that's a job for the fix for  https://github.com/urbit/landscape/issues/200



There is a more serious error here, that is best experienced by going to my comment in the "Grants, Bounties, Projects" chat on UC at 10:12 AM PST today, and seeing both the above problem breaking on mentions, as well as text overflowing horizontally.

matildepark:
Too smart by half.

matildepark:
Ick. I think to handle this we may want to generalise this content render with the work in #4248 and severely simplify Chat's components...

PR #4254, opened 2021-01-08
add basic browser notifications

tylershuster:
Fixes https://github.com/urbit/landscape/issues/195

This is sort of a draft to think about how this could work. Not that it doesn't actually work on its own, but a) I think that it probably shouldn't be implemented until part two of the state hooks projects gets implemented and b) it exposes some interesting considerations that haven't been undertaken yet.

This works by adding a button to the notification "preferences" tab that prompts you to allow notifications. After that, whenever a new notification comes in (not on the initial fetch), it pings you letting you know there is a new notification. As you can see [here](https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification#parameters), there are quite a few things we can do at this point.

Some of those are only made available via the use of service workers. This is a way we could alert people when they get notifications even when landscape isn't running, but the creation of a service worker would be a significant project on its own.

RFC @urcades @matildepark @liam-fitzgerald and of course anyone else

### DIFF
--- a/pkg/interface/src/logic/reducers/hark-update.ts
+++ b/pkg/interface/src/logic/reducers/hark-update.ts
@@ -193,7 +193,7 @@ function unreads(json: any, state: HarkState) {
     clearState(state);
     data.forEach(({ index, stats }) => {
       const { unreads, notifications, last } = stats;
-      updateNotificationStats(state, index, 'notifications', x => x + notifications);
+      updateNotificationStats(state, index, 'notifications', x => x + notifications, false);
       updateNotificationStats(state, index, 'last', () => last);
       if('count' in unreads) {
         updateUnreadCount(state, index, (u = 0) => u + unreads.count);
@@ -235,7 +235,7 @@ function updateUnreadCount(state: HarkState, index: NotifIndex, count: (c: numbe
   }
   const property = [index.graph.graph, index.graph.index, 'unreads'];
   const curr = _.get(state.unreads.graph, property, 0);
-  const newCount = count(curr)
+  const newCount = count(curr);
   _.set(state.unreads.graph, property, newCount);
 }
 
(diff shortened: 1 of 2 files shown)


matildepark:
Such a small diff for such a big world

galenwp:
oh hell yeah

matildepark:
I think the best way to start to unblock this is to talk about what we want to enable and how we want to present them. @urcades @g-a-v-i-n do we have mocks for local notification toasts?

urcades:
@matildepark Actual browser notifications have yet to be designed or structured, but I imagine we'd want them to remain as close to the specification/pattern we've established for our Updates inbox:

Screen Shot 2021-01-20 at 6 28 56 PM

I was doing a cursory overview of some notification formats and it appears as though they are pretty simple:

Screen Shot 2021-01-20 at 7 25 24 PM

Generally it looks like at max, we can surface the following fields:

```
title
source/from
body text
```

Since the 'body' of our notifications are quite variable, involving rollups and such, it might be good to simplify web notifications to something like this:

```
// title: title of triggering group/channel context
// body: Generic category of notification type, ie: "New Messages, New Group Invite, New DM, Mention, etc.
// 
// examples

Urbit Community | Community Formation
New Messages

// or

Tlon | Tlon General
~ravmel-ropdyl mentioned you
```

These are just a rough approximation — as the designer of the bulk of notifications, Gav may have a better idea on how to adapt our current message structure to simpler web notifications.

tylershuster:
This seems sane to start. Another question that needs answering regards *which* notifications come through in this manner. Because these are significantly less calm than a blue dot, does it stand to reason that only a subset of notifications come through in the browser? One option is the grid that you see around like this example from the USPS:
image

matildepark:
I would keep it the same as the blue dot. You should be opting into notifications if you want it to be noisy — and the blue dot is itself pretty noisy for my liking. I only use notifications for DMs and select notebooks.

urcades:
+1 matilde's thought here — i've added a small addition to our system settings prototype that acknowledges that a browser notifications hook exists, and I would essentially assume that it'd be a simple mirror of one's existing in-landscape notification settings, rather than a new subset:

Screen Shot 2021-01-24 at 9 13 33 PM

A note here is that the table that indicates one's landscape-wide channel notifications settings will likely require its own dedicated table view — I could see this getting pretty large.

tacryt-socryp:
What's blocking this?

urcades:
The people are crying out for notifications

—
~fabled-faster
https://urbit.org/

matildepark:
Uhh, I believe us doing it. We were unblocking this with the #4810 work but we would need fetch-based event stream actually in use and in the service worker. Then we implement a basic way of doing it

tylershuster:
Come help me put up this fence

tylershuster:
Soup or seeded by https://github.com/urbit/urbit/pull/4873

PR #4260, opened 2021-01-10
Reference urbit/urbit repository when submitting issues

jtobin:
At present issues opened by the 'Submit an issue' button are opened in the urbit/landscape repo, but since the relevant commit doesn't exist in that repository, GitHub fails to auto-link it.  This prepends `urbit/urbit@` to the rendered Landscape shorthash, which I *think* should cause GitHub to auto-link the commit appropriately.

### DIFF
--- a/pkg/interface/src/views/components/StatusBar.js
+++ b/pkg/interface/src/views/components/StatusBar.js
@@ -54,7 +54,7 @@ const StatusBar = (props) => {
           onClick={() => window.open(
             'https://github.com/urbit/landscape/issues/new' +
             '?assignees=&labels=development-stream&title=&' +
-            `body=commit:%20${process.env.LANDSCAPE_SHORTHASH}`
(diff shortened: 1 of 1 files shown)


matildepark (approved):
Seems so in my testing.

matildepark:
Can you use an `@` in a URL?

jtobin:
Works in simple test examples.  Digging deeper, [this answer](https://webmasters.stackexchange.com/a/508) will tell you accurately that `@` is technically reserved, and then says it should be escaped, but, per the linked [RFC 2396](https://www.ietf.org/rfc/rfc2396.txt), the text reads:

> These characters are called "reserved", since their usage within the URI component is limited to their reserved purpose.  If the data for a URI component would conflict with the reserved purpose, then the conflicting data must be escaped before forming the URI.

It seems to me that the data doesn't conflict with the reserved purpose, so escaping isn't necessary.

PR #4261, opened 2021-01-11
hark-store: fix dangling unreads

liam-fitzgerald:
Correctly remove unreads when nodes or graphs are removed. Additionally fix all current cases of hark-graph mismatch. 

#### graph-store: emit affected descended nodes in %remove-nodes

When %remove-nodes is sent as an update, it may refer to a parent node,
in which case the deletion of all the parent's children is implied. If
we hold onto references to nodes outside of graph-store, we are unable
to tell which of our references were affected, other than iterating over
every reference to check if one of the deleted indices is a prefix of
the reference. This is obviously undesirable, so we emit all of the
nodes we deleted, including the nodes deleted implicitly.

#### hark-store: cleanup dangling unreads

Creates a noun poke %fix-dangling, which iterates through the unreads
and dismiss those which refer to nodes which no longer exist.
Additionally, adds a migration which runs this poke.

#### graph-store: add node existence query

To facilitate next commit

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -869,6 +869,15 @@
     ^-  [index:store node:store]
     [~[atom] node]
   ::
+      [%x %node-exists @ @ @ *]
+    =/  =ship  (slav %p i.t.t.path)
+    =/  =term  i.t.t.t.path
+    =/  =index:store
+      (turn t.t.t.t.path (cury slav %ud))
+    =/  node=(unit node:store)
+      (get-node ship term index)
+    ``noun+!>(?=(^ node))
+  ::
       [%x %node @ @ @ *]
     =/  =ship  (slav %p i.t.t.path)
     =/  =term  i.t.t.t.path
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -418,43 +418,83 @@
       =/  =update-log:store  (~(got by update-logs) resource)
       =.  update-log
         (put:orm-log update-log time [%0 time [%remove-nodes resource indices]])
+      =/  [affected-indices=(set index:store) new-graph=graph:store]
+        (remove-indices resource graph (sort ~(tap in indices) by-lent))
       ::
-      :-  (give [/updates]~ [%remove-nodes resource indices])
+      :-  (give [/updates]~ [%remove-nodes resource (~(uni in indices) affected-indices)])
       %_  state
           update-logs  (~(put by update-logs) resource update-log)
           graphs
         %+  ~(put by graphs)
           resource
-        [(remove-indices resource graph ~(tap in indices)) mark]
+        [new-graph mark]
       ==
       ::
+      :: we always want to remove the deepest node first,
+      :: so we don't remove parents before children
+      ++  by-lent
+        |*  [a=(list) b=(list)]
+        ^-  ?
+        (gth (lent a) (lent b))
+      ::
       ++  remove-indices
+        =|  affected=(set index:store)
         |=  [=resource:store =graph:store indices=(list index:store)]
-        ^-  graph:store
-        ?~  indices  graph
+        ^-  [(set index:store) graph:store]
+        ?~  indices  [affected graph]
+        =^  new-affected  graph
+          (remove-index graph i.indices)
         %_  $
             indices  t.indices
-            graph    (remove-index graph i.indices)
+            affected  (~(uni in affected) new-affected)
+        ==
+      ::
+      ++  get-descendants
+        |=  =graph:store
+        =|  indices=(list index:store)
+        =/  nodes=(list [atom node:store])
+          (tap:orm:store graph)
+        %-  ~(gas in *(set index:store))
+        |-  =*  tap-nodes  $
+        ^+  indices
+        %-  zing
+        %+  turn
+          nodes
+        |=  [atom =node:store]
+        ^-  (list index:store)
+        %+  welp
+          (limo index.post.node ~)
+        ?.  ?=(%graph -.children.node)
+          ~
+        %_  tap-nodes
+          nodes  (tap:orm p.children.node)
         ==
       ::
       ++  remove-index
+        =|  indices=(set index:store)
         |=  [=graph:store =index:store]
-        ^-  graph:store
-        ?~  index  graph
+        ^-  [_indices graph:store]
+        ?~  index  [indices graph]
         =*  atom   i.index
         ::  last index in list
         ::
         ?~  t.index
-          +:`[* graph:store]`(del:orm graph atom)
+          =^  u-val  graph  (del:orm graph atom)
+          ?~  u-val  `graph
+          ?.  ?=(%graph -.children.u.u-val)
+            `graph
+          =/  new-indices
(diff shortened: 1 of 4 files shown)


matildepark:
Target master, this is going out once we hit 100% confidence

matildepark:
This cleared Jobs + Gigs' dangling unread on a -L'd haddef.

pkg/arvo/app/graph-store.hoon
```diff
@@ -418,43 +418,83 @@
...
+        %-  ~(gas in *(set index:store))
+        |-  =*  tap-nodes  $
+        ^+  indices
+        %-  zing
+        %+  turn
+          nodes
+        |=  [atom =node:store]
+        ^-  (list index:store)
+        %+  welp
+          (limo index.post.node ~)
```
tacryt-socryp, line 466: Best to use literal syntax rather than `+limo`.

pkg/arvo/app/graph-store.hoon
```diff
@@ -418,43 +418,83 @@
...
         %_  $
             indices  t.indices
-            graph    (remove-index graph i.indices)
+            affected  (~(uni in affected) new-affected)
+        ==
+      ::
+      ++  get-descendants
+        |=  =graph:store
+        =|  indices=(list index:store)
+        =/  nodes=(list [atom node:store])
```
tacryt-socryp, line 455: Best to single line this expression, and no need for the cast to `(list [atom node:store])`

pkg/arvo/app/graph-store.hoon
```diff
@@ -418,43 +418,83 @@
...
+          (tap:orm:store graph)
+        %-  ~(gas in *(set index:store))
+        |-  =*  tap-nodes  $
+        ^+  indices
+        %-  zing
+        %+  turn
+          nodes
```
tacryt-socryp, line 462: Best to collapse this line into the prior line

pkg/arvo/app/graph-store.hoon
```diff
@@ -418,43 +418,83 @@
...
+        %_  tap-nodes
+          nodes  (tap:orm p.children.node)
         ==
       ::
       ++  remove-index
+        =|  indices=(set index:store)
         |=  [=graph:store =index:store]
-        ^-  graph:store
-        ?~  index  graph
+        ^-  [_indices graph:store]
```
tacryt-socryp, line 476: Would prefer to have the type of this expression be `(set index:store)` rather than `_indices`. Just for clarity!

pkg/arvo/app/graph-store.hoon
```diff
@@ -418,43 +418,83 @@
...
+        ?~  index  [indices graph]
         =*  atom   i.index
         ::  last index in list
         ::
         ?~  t.index
-          +:`[* graph:store]`(del:orm graph atom)
+          =^  u-val  graph  (del:orm graph atom)
```
tacryt-socryp, line 482: Keeping with consistent naming, `new-graph` is better than `u-val`

pkg/arvo/app/hark-graph-hook.hoon
```diff
@@ -167,11 +167,42 @@
...
+      (turn ~(tap in indices) (lead rid))
+    :_  state(watching (~(dif in watching) to-remove))
+    =/  =tube:clay
+      (get-conversion:ha rid)
+    %+  roll
+      ~(tap in indices)
+    |=  [=index:graph-store out=(list card)]
+    =|  =indexed-post:graph-store
+    =.  index.p.indexed-post  index
+    =+  !<(u-notif-kind=(unit notif-kind) (tube !>(indexed-post)))
```
tacryt-socryp, line 194: Would this not be neater as

`=/  u-notif-kind=(unit notif-kind)  !<((tube !>(indexed-post)))`
?

pkg/arvo/app/hark-graph-hook.hoon
```diff
@@ -167,11 +167,42 @@
...
+    ?~  u-notif-kind  out
+    =*  notif-kind  u.u-notif-kind
+    =/  =stats-index:store
+      [%graph rid (scag parent-lent.notif-kind index)]
+    ?.  ?=(%each mode.notif-kind)  out
+    :_(out (poke-hark %read-each stats-index index))
```
tacryt-socryp, line 200: Neater to put this in tall form, I think.

pkg/arvo/app/hark-store.hoon
```diff
@@ -78,9 +84,16 @@
...
   ?-  -.old
-      %2  
+      %3  
     :-  cards
     this(-.state old, +.state (inflate-cache:ha old))
+    ::
+      %2
+    %_  $
+      -.old  %3
+      cards  :_(cards [%pass / %agent [our dap]:bowl %poke noun+!>(%fix-dangling)])
```
tacryt-socryp, line 94: Long line, neater in tall form.

pkg/arvo/app/hark-store.hoon
```diff
@@ -264,10 +277,43 @@
...
+  ++  poke-noun
+    |=  val=*
+    ?+  val  ~|(%bad-noun-poke !!)
+      %fix-dangling  fix-dangling
+      %print  ~&(+.state [~ state])
+    ==
+  ::
+  ++  fix-dangling
+    =/  graphs=(set resource)
+      get-keys:gra
```
tacryt-socryp, line 293: Should be single-lined, and the cast is unnecessary.

pkg/arvo/app/hark-store.hoon
```diff
@@ -264,10 +277,43 @@
...
+    :_  state
+    %+  roll
+      ~(tap by unreads-each)
+    |=  $:  [=stats-index:store indices=(set index:graph-store)]
+            out=(list card)
+        ==
+    ?.  ?=(%graph -.stats-index)  out
+    ?.  (~(has in graphs) graph.stats-index)
+      :_(out (poke-us %remove-graph graph.stats-index))
```
tacryt-socryp, line 302: Any reason you're avoiding writing this as
`[(poke-us %remove-graph graph.stats-index) out]`
?

pkg/arvo/app/hark-store.hoon
```diff
@@ -264,10 +277,43 @@
...
+    %+  weld  out
+    ^-  (list card)
+    %+  turn
+      ^-  (list index:graph-store)
+      %+  skip
+        `(list index:graph-store)`~(tap in indices)
```
tacryt-socryp, line 308: Do you need the cast on this line? I wouldn't think you do.

tacryt-socryp (changes requested):
Style changes, nothing major. Looks great. Will promptly change to approved once they're addressed.

pkg/arvo/app/hark-graph-hook.hoon
liam-fitzgerald: `!<` is binary no?

pkg/arvo/app/hark-store.hoon
liam-fitzgerald: i think it's cleaner to make the cons to the head of the list more obvious when read left to right, happy to change though

tacryt-socryp: All good

tacryt-socryp (approved):

PR #4263, opened 2021-01-11
hark-graph-hook: fix muting on %remove-graph

liam-fitzgerald:
*_watching was bunting to watching (i.e. a no-op), causing watching to
be wiped everytimne a graph was removed


tacryt-socryp (approved):
LGTM

PR #4265, opened 2021-01-11
chat: fix sizing on mobile

tylershuster:
fixes https://github.com/urbit/landscape/issues/230

uses code shamelessly swiped from https://codepen.io/shshaw/pen/c72208b279d750f1ab46b6a45bbd04d7 via https://css-tricks.com/auto-growing-inputs-textareas/


matildepark:
Looks good, but the font size (and I think line height) doesn't match the chat above. Also the clickable area to start typing is about half the field horizontally.

tylershuster:
Oh, duh. try this.

matildepark:
Clickable area still too small. If you use, say, your right thumb, you won't know why it's unresponsive.

tylershuster:
Ok, that's a grid layout weirdness. I added a listener to watch the parent element for focus

matildepark (approved):

PR #4266, opened 2021-01-11
channel.js: ack every 30 events *even if this.outstandingJSON is empty*

### LINKED ISSUE #4110: landscape: repeated clog on group-store subscription
I'm getting `%e %clogged` printfs relatively frequently. Decided to pop open the debug dashboard and see what specific subscription this is. Took a little bit before I caught it in the act, but:

image

(Note the right-hand "unacked events" column.)

Hitting refresh repeatedly sometimes shows the amount of unacked events for the group-store subscription climb rapidly, sometimes paired with `<<ames>>` flickering on the ship's cli. My best guess it's getting contact data updates or something for people who are (re)joining the groups I'm in, but why those events specifically aren't getting acked, I have no clue.

tacryt-socryp:
Hotfix to channel.js that *should* (currently being tested) vastly reduce `%eyre` clogs. Fixes #4110.

### DIFF
--- a/pkg/arvo/app/landscape/js/channel.js
+++ b/pkg/arvo/app/landscape/js/channel.js
@@ -63,7 +63,7 @@ class Channel {
   }
 
   resetDebounceTimer() {
-    if(this.debounceTimer) {
+    if (this.debounceTimer) {
       clearTimeout(this.debounceTimer);
       this.debounceTimer = null;
     }
@@ -203,7 +203,7 @@ class Channel {
       //
       let payload = [
         ...this.outstandingJSON, 
-        {action: "ack", "event-id": parseInt(this.lastEventId)}
+        {action: "ack", "event-id": this.lastEventId}
       ];
       if (j) {
         payload.push(j)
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (approved):
Looks correct, have you tested it?

tacryt-socryp:
Am currently testing it on `~tacryt` to see if it reduces clogs in the wild.

tacryt-socryp:
Oof, this issue was far worse than expected. We were incorrectly setting the `lastAcknowledgedEventId` and also trying to compare the values of a string and an integer with the `<` operand to determine whether to send the ack. In other words, we never acked under any conditions.

matildepark:
I'm content pushing this tomorrow but with the caveat that we cache bust channel.js to ensure people catch the fix here.

matildepark:
ping @tacryt-socryp ? This seems sooner the better if tacryt is clogged less.

tacryt-socryp:
This definitely made ~tacryt clog less

matildepark:
So can we cache bust it? Looks like one line in file server.

Fang- (approved):
Pretty wild. I think the two branches of the "have acknowledged last event" check could be merged, the branching moved to the construction of the payload specifically. Might make the flow here a bit more clear. But the logic itself is fine as-is.

tacryt-socryp:
Brief offline chat with @matildepark and we're moving ahead with this as a hotfix as is. not sure about deploy schedule

Issue #4219, opened 2021-01-05
chat: "new messages" banner uses oldest loaded timestamp, not last read


matildepark:
Will see "4 messages since ~2020.11.5" when the newest messages are from today — seems to change per room, so I can only imagine it's pulling the top from the envelopes we have instead of the actual newest unread message.

(Did Hark change its unread order? Regardless, might not be too hard to fix...)

PR #4267, opened 2021-01-11
chat: unreverse keys when getting unread msg

matildepark:
See #4219 for filed issue.

The unread message banner was getting its sourced unread message from the index of the last unread message; but this was itself called from a reversed `graph.keys()`, so if I have 9 unreads, it seems like it would get the 9th oldest loaded in, not the 9th newest?

If we call `graph.keys()` for this message without the reverse, the banner's timestamp functions as it should.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
@@ -258,7 +258,7 @@ export default class ChatWindow extends Component<ChatWindowProps, ChatWindowSta
     const messageProps = { association, group, contacts, unreadMarkerRef, history, api };
 
     const keys = graph.keys().reverse();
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (changes requested):
should be `this.props.unreadCount - 1` as the index, otherwise good

matildepark:
I disagree, that produces the latest unread's date, but we're saying "Messages since [the last message you read]". So if you last read at 5:22pm and the last unread is 5:30pm, it'll say "Messages since 5:30pm" when we want the conversation since that last message I read.

liam-fitzgerald:
hmm I always interpreted it as "messages since [earliest message you haven't read]"

liam-fitzgerald (approved):

matildepark:
Got an informal @urcades stamp on this distinction. In we go.

PR #4268, opened 2021-01-11
group-view (nee graph-view)

liam-fitzgerald:
Changes group joining to be handled by this agent. Joining a hidden group now automatically joins the channels inside that group. Replaces the old contact-view %join poke and also the current -graph-join thread for the unmanaged case. I'm of the opinion that with contact-view gone, we should look to the group-view agent and associated threads to handle more of the lifecycle of unmanaged graphs.


tacryt-socryp:
I'm still of the view that we can resolve this issue by simply using a thread that only makes a single request (no timers, no back-offs), and the poke-proxy-hook for keeping track of pending requests.

matildepark:
<details>
<summary>~2020.1.11 discussion on doing join requests</summary>

- LA: Why are threads bad for joining resources?
- LF: Join thread is bad because group-store works by waiting for the remote ack on the add-members poke before you can continue to join the group. Because of that, the thread can wait an indefinitely long amount of time, during which it might be reloaded; the thread might die; and because we launch the threads directly from the HTTP interface, we have no control. If the join thread just dies, we have no way to notify the user the join request failed because [x, y, z]. It’s because it’s a thread that we can’t; usually it’s guaranteed it happens in the event, but in this case, since we have to wait for the ack, it’s less viable.
- LA: I agree the current approach isn’t viable, starting from client and waiting for group ack and determine if it’s on a pending state or not. I don’t think we need as heavy an approach as having an app for handling these cases. 
    - What do you think of the following: the poke-proxy-hook emits a `%fact` to anyone that is subscribed to it when it receives any positive acknowledgment of its pending pokes, so when you send a poke, it goes into the pending pokes and the poke-proxy-hook and when the poke is acked or nacked, a diff is sent out, a fact is sent out, saying whether there was an ack or a nack on that particular poke, and given that system, we could more or less make the front-end send a — 
    - LF: request for all outstanding pokes?
    - LA: Yeah, we could do that for pending state. But we could also use it to determine whether the group-pull-hook request succeeded or failed, and then we could hook it up to the observe-hook so that the observe-hook would actually, upon receiving that positive acknowledgement saying you joined successfully, the observe-hook would join the metadata resource, since the metadata resource is the only one we would be joining upon group acceptance, since we would be joining the contacts. It would still be done in threads, but it would be upon the poke-proxy-hook completing its thing, and there would be no waiting in the threads.
- LF: One of the issues I have with that is that it still doesn’t address that, in theory, threads can be killed at any time, if spider is reloaded or anything. And it also, I think, makes the join flow, which is far and away the most complex state transition we do in Landscape, split up into three or four components (observe-hook, thread for metadata-hook after joining the group successfully, etc.). I’m wary of overusing the observe-hook with spooky action at a distance issues because of how observe-hook did something.
- LA: So, more or less, you’re saying you would prefer for us to keep all of the logic for joining in a single place: rather than splitting up across the poke-proxy-hook, observe-hook, and some particular thread — putting it in a single app.
- LF: The graph-join thread is 94 lines, and the graph-view thread, with more features, is 226 lines in tall code. The minor readability of a thread is not preferable to the downsides in a heavily networked scenario.
- LA: That makes sense. I was talking about group joining, and now we’re talking about graph joining, which currently has the issue. Group joining is about to get simpler, now that there’s only one chain of events, if that makes sense — you join the group, and if you join it successfully, you join the metadata. As opposed to also joining contacts. I think, though, that the one thing you said surrounding “what about some long-running thread that gets killed”, any thread started, where all it did was send a poke to metadata-store, would conclude in a single event, so there would be no waiting with it. I guess what I’m thinking right now is, if we have specific apps for group and graph joining, it seems like they will both be replicating some of the logic that could be in poke-proxy-hook, some of what this poke-proxy-hook is supposed to do would be made redundant by all these apps. I like the idea of a generic pending request app.
- LF: The other issue is, to counter the group/graph join distinction, you can get a lot more information out of doing it this way — in theory, under the thread model, if you wanted to indicate to the user that the poke failed because of a ban, how would we handle that?
- LA: A moderator tried to add someone as a member who the host had banned?
- LF: If they try to join the group. In the join thread, we just return a 401. How are you relaying that back to the user?
- LA: How do you rely back that some part of a transaction failed, joining the group failed? If poke-proxy-hook told us, “joining this group failed” … I could see us needing a graph-view and a group-view. I don’t know that it’s actually a view. I feel like we need some other name for this.
- LF: You could imagine it as a ‘join manager’ or whatever. If you read the hoon, it’s fairly isolated. We could get a join group, join graph, and then one of them could call the other or something like that.
- LA: It’s worth noting that what you’re describing is not a view, it’s not really concatenating actions in a straightforward way for making the client work better, it has a different purpose of acting as a state machine for pending actions? I think that’s cool, it seems like a type of hook maybe, but I feel reasonably convinced we don’t need threads for joining at this point. Does this obsolete poke-proxy-hook?
- LF: poke-proxy-hook works better in the case where generally all you’re doing is sending one action to a remote host; it’s relatively simple. What we can do is give them a list of unacked pokes and they can work it out from there. With the group-proxy-hook, you would see a bunch of add-members and other stuff waiting elsewhere, and you wouldn’t know which of these are ‘add-members’ — you wouldn’t look at poke-proxy-hook for the state of a long-running transaction. There’s things you want to do once, and more complicated actions you want to do that doesn’t fit well with poke-proxy-hook and one of the pokes will end up in one proxy-hook and another in another proxy-hook, because we can’t do one generic poke-proxy-hook due to restrictions on JSON conversions in the nested JSON conversion problem I was discussing. We can’t do a unified app poke-proxy-hook for that reason. We have to write a library wrapper to do most of it and some query patterns.
- LA: This seems like a Landscape join hook.

</details>

tacryt-socryp (approved):
LGTM

tacryt-socryp:
NOTE: Before merging, we need to have this added to `%hood`

liam-fitzgerald:
Sorry, should have been clearer. This still needs FE work to surface the group joining states. Not ready for merge yet.

liam-fitzgerald:
ready to go!

PR #4269, opened 2021-01-12
hark: address phantom dot on archive

liam-fitzgerald:
We simply did not update the notification count when a notification was archived. Fixed

Fixes https://github.com/urbit/landscape/issues/233


galenwp:
thanks!

matildepark (approved):

PR #4275, opened 2021-01-12
landscape: use indigo-react themes

matildepark:
Uses `@tlon/indigo-light` and `@tlon/indigo-dark` instead of our interim themes.

Screen Shot 2021-01-12 at 2 32 06 PM


matildepark:
Note that this doesn't include the upcoming typographic update in indigo-react, which isn't released yet. Once it is I can update this PR again.

g-a-v-i-n (approved):
very nice

matildepark:
Updated to 1.2.16. Notably, `Text` defaults to 14pt now, so if we didn't dictate, it got stepped up. Everything looks Big. @urcades asked me to just let it hit the devnet so we can clear it up in practice together.

Screen Shot 2021-01-12 at 2 55 22 PM

tacryt-socryp:
Incredible

tacryt-socryp (approved):

PR #4276, opened 2021-01-12
settings store

ixv:
Draft PR for the settings store. Still need to do frontend stuff.

Can someone tell me if I have the pattern right for scrying? I wasn't sure if I should return `~` or `[~ ~]` when scrying for a thing that doesn't exist, but might exist in the future.

### DIFF
--- a/pkg/arvo/app/settings-store.hoon
+++ b/pkg/arvo/app/settings-store.hoon
@@ -76,13 +76,13 @@
     ^-  (unit (unit cage))
     ?+  pax  (on-peek:def pax)
         [%x %all ~]
-      ``noun+!>(settings)
+      ``settings-data+!>(all+settings)
     ::
         [%x %bucket @ ~]
       =*  buc  i.t.t.pax
       =/  bucket=(unit bucket)  (~(get by settings) buc)
       ?~  bucket  [~ ~]
-      ``noun+!>(u.bucket)
+      ``settings-data+!>(bucket+u.bucket)
     ::
         [%x %entry @ @ ~]
       =*  buc  i.t.t.pax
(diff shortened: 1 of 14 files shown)


liam-fitzgerald:
Yeah in userspace you always want to use `[~ ~]`. The way I remember it is that ~ is "I don't know anything about this path" and and [~ ~] is I know that I don't have it. ~ is really only useful for scrying foreign ships in kernelspace, as I understand it.

pkg/arvo/app/settings.hoon
```diff
@@ -0,0 +1,167 @@
...
+  ++  on-watch
+    |=  pax=path
+    ^-  (quip card _this)
+    ?>  (team:title our.bol src.bol)
+    ?+  pax  (on-watch:def pax)
+        [%all ~]
+      [~ this]
+    ::
+        [%bucket @ ~]
+      =/  bucket-key  i.t.pax
```
liam-fitzgerald, line 62: Should use tistar instead

liam-fitzgerald (approved):
Minor quibble, looks good otherwise

matildepark:
ping @tacryt-socryp this could also use a review but later in the week is fine

pkg/arvo/app/settings-store.hoon
```diff
@@ -0,0 +1,167 @@
...
+      ``settings-data+!>(entry+u.entry)
+    ==
+  ::
+  ++  on-agent  on-agent:def
+  ++  on-leave  on-leave:def
+  ++  on-arvo   on-arvo:def
+  ++  on-fail   on-fail:def
+  --
+::
+|_  bol=bowl:gall
```
tacryt-socryp, line 102: I generally prefer shadowing `=bowl:gall` so as not to be able to accidentally reference the bunted bowl.

tacryt-socryp (approved):
LGTM, only one note on bowl shadowing vs `bol`.

PR #4278, opened 2021-01-13
hark-store: rework caching to correctly calculate unread totals

### LINKED ISSUE #236: :begin does not accept just a country code
Greetings, I just went through :begin and was asked a few questions for the %lord form, including the country/location question.

Here is the transcript:

        Where are you?  This is totally optional, but we'd like to
        know your vague general location.  You can enter nothing at
        all, just a country code, or country and postal code.
        Location at
        invalid input
        Location at/1000

So I would conclude that :begin does not currently accept just a country code.

If you think this issue is irrelevant at this stage, feel free to close it ;-) Just posting it because on the path toward a version "1.0", also usability improvements count for user acceptance.

liam-fitzgerald:
Hark, when calculating unread totals was ignoring notifications that
were under the same stats-index, but a different index. Reworks the
cache to calculate unread totals correctly.

Fixes urbit/landscape#236

### DIFF
--- a/pkg/arvo/app/hark-store.hoon
+++ b/pkg/arvo/app/hark-store.hoon
@@ -49,7 +49,7 @@
 ::  $cache: useful to have precalculated, but can be derived from state
 ::  albeit expensively
 +$  cache
-  $:  by-index=(jug stats-index:store @da)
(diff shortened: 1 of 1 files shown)


tacryt-socryp (approved):
LGTM

PR #4279, opened 2021-01-13
notifications: fix layout

tylershuster:
fixes https://github.com/urbit/landscape/issues/218
fixes https://github.com/urbit/landscape/issues/237

before:
image

after:
image

also fixes an unfiled bug where leap crashes if state.suspendedfocus is invalid

I don't understand a lot of how the inbox is set up, but I think this is sane, so ping @liam-fitzgerald

### DIFF
--- a/pkg/interface/src/views/apps/notifications/notification.tsx
+++ b/pkg/interface/src/views/apps/notifications/notification.tsx
@@ -18,6 +18,7 @@ import { GroupNotification } from "./group";
 import { GraphNotification } from "./graph";
 import { ChatNotification } from "./chat";
 import { BigInteger } from "big-integer";
+import { useHovering } from "~/logic/lib/util";
 
 interface NotificationProps {
   notification: IndexedNotification;
@@ -88,7 +89,7 @@ function NotificationWrapper(props: {
     return api.hark[func](notif);
   }, [notif, api, isMuted]);
 
-  const [visible, setVisible] = useState(false);
+  const { hovering, bind } = useHovering();
 
   const changeMuteDesc = isMuted ? "Unmute" : "Mute";
   return (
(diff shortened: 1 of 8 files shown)


matildepark:
@vvisigoth did report that bug but no one could repro. He reported it by emailing me.

matildepark:
Squeezing this in would be cool https://github.com/urbit/landscape/issues/148 (and see the others with that label)

tylershuster:
- [ ] https://github.com/urbit/landscape/issues/171
- [ ] https://github.com/urbit/landscape/issues/170
- [ ] https://github.com/urbit/landscape/issues/148
- [ ] https://github.com/urbit/landscape/issues/146

matildepark:
cc @urcades

tylershuster:
(from urbit/landscape#171)
image

pkg/interface/src/views/apps/notifications/notification.tsx
```diff
@@ -89,11 +88,22 @@ function NotificationWrapper(props: {
     return api.hark[func](notif);
   }, [notif, api, isMuted]);
 
+  const [visible, setVisible] = useState(false);
```
liam-fitzgerald, line 91: pendanticism, but this is a good candidate for a custom hook. 
.e.g.
```typescript
const { hovering, bind } = useHovering();
// in component, bind is simply { onMouseEnter, onMouseLeave }
<Box {...bind}>

liam-fitzgerald (approved):
LGTM! minor nitpick about refactoring, else very good!

pkg/interface/src/views/apps/notifications/notification.tsx
tylershuster: Great idea; I wondered how I should do this. I've added another commit which adds and implements this.

urcades:
these fixes look sane, whew — stamp!

Issue #3329, opened 2020-08-17
Publish: Pressing "enter/return" after a written list item should create a new list item


urcades:
This is funny to write out, so here are some clear visual examples.

This is what publish does now while writing lists:

![auto-list](https://user-images.githubusercontent.com/1195363/90367027-28778a00-e036-11ea-8a60-e51aa08b3b4f.gif)

This is the ideal behavior, I believe:

![good](https://user-images.githubusercontent.com/1195363/90367057-37f6d300-e036-11ea-8f51-79a5c3ba7899.gif)

This is such a basic text-editing affordance that in my mind, it's basically a bug that it's not implemented in our current text editor.

tylershuster:
a) new-post and edit-post should be deduped in indigo
b) Might we want to consider using ProseMirror? I know most of us are nerds but it seems like prosemirror offers a better editing experience for most people

urcades:
I'm basically pro-anything that gets us to a place where editing text is as comfortable as using iA writer, which is the experiential baseline imo. Surrounds, auto-lists, proper tabbing behavior, etc. would all be really nice to have.

tylershuster:
Then I'll double down on my recommendation of ProseMirror (specifically https://github.com/outline/rich-markdown-editor)— CodeMirror doesn't really have a concept of nestedness like we think about it in prose...it just pushes and pops states. The repo I linked is used in/created by getoutline.com which I've never used but my ten second trial confirmed that it does know how to do lists like you demoed. Give it a shot?

Without any judgement, I think the current use of codemirror is probably due to the current userbase's technical inclination and a better setup might be codemirror for code (dojo, whatever code-editing app we end up with) and this component or similar for everything else.

tylershuster:
https://open.nytimes.com/building-a-text-editor-for-a-digital-first-newsroom-f1cb8367fc21

urcades:
cc @g-a-v-i-n : Would be curious to hear your thoughts on the above ^

@tacryt-socryp : skimming this, it seems Really Nice. I recall seeing that NYT writeup on their text editor, and it seems at a glance to be incredibly good. The capability to annotate + build custom headers would be kind of amazing for the next iteration(s) of Publish, or whatever comes after/is derived from it.

urcades:
Sorry, @tylershuster meant to @ you in that last message, not ~tacryt

tylershuster:
Honored

g-a-v-i-n:
[...]

PR #4282, opened 2021-01-13
publish: automatically continue lists

matildepark:
See #3329 for filed issue.

Imports [continueList](https://codemirror.net/doc/manual.html#addons) CodeMirror addon and hooks into 'Enter' to prefix the next line with another list item if we're in a list.


urcades:
This is beautiful , yes, we are inching to perfection

tacryt-socryp (approved):
whoa

PR #4283, opened 2021-01-14
push, pull hooks: prevent forwarded pokes attack

liam-fitzgerald:


tacryt-socryp (approved):
Really great work catching this before it occurred in the wild. LGTM

PR #4284, opened 2021-01-14
hark-graph-hook: rewatch all DMs on upgrade

liam-fitzgerald:
Does what it says on the tin, to correct for an earlier bug that may have wiped the `watching` state


tacryt-socryp (approved):
LGTM, ran this on my moon: ~moldeb-datlux-tacryt-socryp.

PR #4288, opened 2021-01-14
chat: allow rendering newlines

matildepark:
See #4202. We make use of `display: block` and `inline-flex` to handle "mid-sentence mention" and "multiple paragraph" cases, retaining the line breaks entered but mostly rendering inline.

cc @tylershuster can you break it?


tylershuster:
As entered:
```
Hello ~marpem it is
I, ~radbur-sivmus 
your old nemesis
```
As rendered:
image

matildepark:
Defeated again.

PR #4289, opened 2021-01-14
chat: allow line breaks in markdown

matildepark:
See #4202 for filed issue.

Uses RemarkBreaks plugin to render line breaks as `<br>` instead of rendering as child text nodes.

OK, one last time @tylershuster...

image


tylershuster:
image

matildepark:
Can I get a stamp @liam-fitzgerald

liam-fitzgerald (approved):

PR #4291, opened 2021-01-15
landscape: QA from 1.5, round one

matildepark:
image

- Removes the empty circles from Hark; it's a dot or it's blank now.
- You can only mark notifications as read; no toggle.
- Removes some Chat cruft from the Notifications code.
- Hark dates use `@da`.
- Squares out the icons in the status bar.
- Sidebar uses the latest mockup.


urcades:
Howdy — posting up some of the more subtle/precision margin choices we've made for the next iteration of the sidebar:

margins

urcades:
Note: Until the group avatar is actually a part of a group's data we can surface, we can omit the space it'd take up, indicated in the image above

matildepark:
What’s the header font weight and size here? Looks bolder than 500 / 600

liam-fitzgerald (approved):
LGTM!

PR #4292, opened 2021-01-15
build: tag docker image as "latest"

eamsden:
When pushing the docker image, update the "latest" tag as well as
pushing a tag for the current Urbit version. This way users pulling
tloncorp/urbit:latest will always get an image with the latest point
release of Urbit.

@jtobin this is a convenience fix for the docker image release cc @jalehman

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -40,4 +40,8 @@ jobs:
 
       - run: |
           version="$(cat ./pkg/urbit/version)"
-          $(nix-build -A skopeo)/bin/skopeo --insecure-policy copy tarball:$(nix-build -A docker-image) docker://${{ secrets.DOCKERHUB_USERNAME }}/urbit:v$version
+          skopeo="$(nix-build -A skopeo)/bin/skopeo"
+          image="$(nix-build -A docker-image)"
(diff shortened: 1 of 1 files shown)


jtobin (approved):

.github/workflows/release-docker.yml
```diff
@@ -40,4 +40,8 @@ jobs:
 
       - run: |
           version="$(cat ./pkg/urbit/version)"
-          $(nix-build -A skopeo)/bin/skopeo --insecure-policy copy tarball:$(nix-build -A docker-image) docker://${{ secrets.DOCKERHUB_USERNAME }}/urbit:v$version
+          skopeo="$(nix-build -A skopeo)/bin/skopeo"
+          image="$(nix-build -A docker-image)"
+          $skopeo --insecure-policy copy tarball:$docker docker://${{ secrets.DOCKERHUB_USERNAME }}/urbit:v$version
```
eamsden, line 45: Checked it this morning and realized there is a typo: should be `tarball:$image` instead of `tarball:$docker`

I wish there was a better way to test GitHub CI actions.

eamsden: I will make one more PR to fix this @jtobin

jtobin: Ah, my bad for not catching it.  I went over the changes carefully, but not quite carefully enough. 😄

PR #4293, opened 2021-01-15
graph-store: log %add-graph

### LINKED ISSUE #4290: hark: inactive channels occasionally produce false unreads
Four independent confirmations of the issue in UC General.

> is anyone else on the bleedingedge branch getting a thing where just,,, random channels that havent been messaged in for weeks get an unread for some reason

> on uc the last thing on news from tlon becomes unread every so often, but the issue doesnt affect other channels

> News From Tlon in UC, some channels in litmus' art group that havent been used in a month or two, and the #go channel in freenode have all been victims of ghost unreads for me

Seemingly only happened today, so likely side effect from #4278?

liam-fitzgerald:
When a graph is joined, the subscriber receives an %add-graph update
that corresponds to the current state of the graph. This was not logged,
so if the graph did not recieve any updates before the subscription was
kicked, the whole graph would be have to be sent again. This would be
largely asymptomatic, however hark-store assumes that %add-graph will
only be emitted when a new graph is joined or created. As such, we log
this initial update, so that the push-hook can correctly ascertain the
backlog it needs to send and so %add-graph is only emitted once per
graph. This also opens the door for time-varying scries into
graph-store.

Fixes #4290

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -230,7 +230,7 @@
     ?>  ?=(%0 -.update)
     =?  p.update  =(p.update *time)  now.bowl
     ?-  -.q.update
-        %add-graph          (add-graph +.q.update)
+        %add-graph          (add-graph p.update +.q.update)
         %remove-graph       (remove-graph +.q.update)
         %add-nodes          (add-nodes p.update +.q.update)
         %remove-nodes       (remove-nodes p.update +.q.update)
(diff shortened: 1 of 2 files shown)


tacryt-socryp (approved):
Very good, I've been meaning to make this change, but you beat me to it!

matildepark:
https://github.com/urbit/urbit/commit/69a5675a86130ee85ec42c02d75bbde6688be4cb is in master, closing. (Use sh/merge-with-custom-msg next time to auto-close these, @liam-fitzgerald.)

PR #4299, opened 2021-01-17
docker-image: Pass `-t` flag to vere if not in an interactive container

eamsden:
Vere will not run a ship if STDIN is closed, unless passed the `-t`
flag. Hosted docker containers will not be interactive, but we would
like the docker image to be usable locally and interactively as well.

Thus, we check for STDIN being closed in the shell prior to launching
Vere, and if it is closed, pass the `-t` flag.

### DIFF
--- a/nix/pkgs/docker-image/default.nix
+++ b/nix/pkgs/docker-image/default.nix
@@ -5,6 +5,15 @@ let
 
     set -eu
 
+    # If the container is not started with the `-i` flag
+    # then STDIN will be closed and we need to start
+    # Urbit/vere with the `-t` flag.
+    ttyflag=""
+    if [ ! -t 0 ]; then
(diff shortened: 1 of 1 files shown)


ohAitch:
This really seems like it should live in vere itself, is there a reason we'd want "exit with an error message with what flag you should have passed" behavior? (I think the error message currently does not have that information, but presumably could if `isatty` was called on stdin prior using it)

eamsden:
@ohAitch I agree, but without it the Docker image can't actually be deployed to e.g. ECS. I'll open an issue against Vere but this should go in for now.

eamsden:
@ohAitch I opened a bug against Vere: https://github.com/urbit/urbit/issues/4302

jtobin (approved):

PR #4310, opened 2021-01-19
landscape: typography refresh sprint QA

matildepark:
cc @urcades 

- Prevent text clipping in group titles.
- Widen popover and profile menu size from 150px to 250px.
- Level the font size to 14px nearly everywhere.
- Chat editor padding is increased on mobile devices.
- Remove patp from status bar.
- Move 'new post' button to the header of resources, removes the metadata display in Publish resources.

Screenshot 2021-01-19 at 4 26 03 PM


liam-fitzgerald (approved):
LGTM!

PR #4314, opened 2021-01-20
ames: cache message +jam in +on-memo

belisarius222:
Tested with @tacryt-socryp.  Reduces number of calls to `+jam` when publishing chat messages down to one total, as opposed to one per subscriber.  Since the `~+` cache does not persist across Arvo events, this doesn't catch all cases of message duplication, but it catches the common use case of Gall `%fact`s being delivered to many subscribers in a single Arvo event.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -180,6 +180,9 @@
     $(index +(index), sorted [(~(got by fragments) index) sorted])
   ::
   (cue (rep 13 (flop sorted)))
+::  +jim: caching +jam
+::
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
Looks safe!  Do we know when exactly the memo cache gets cleared?  I thought it was on every road pop, but that might be wrong.  But if you tested that this calls jam less often (eg with a printf in jim), that's good enough for me.

tacryt-socryp:
We have tested that this does call jam less often (by using a printf in jim).

matildepark:
@philipcmonk I'm planning on wrangling an update together early tomorrow, so I'll target a hotfix branch and get this out then.

PR #4321, opened 2021-01-21
fixes urbit/landscape#252

tylershuster:
For lo


matildepark:
We can't do flex'd stickies? What about just `flexShrink={0}`?

tylershuster:
Alright that works too

matildepark (approved):

PR #4322, opened 2021-01-21
interface: react to deleted comments

### LINKED ISSUE #4272: deleting a comment on a link or note gives no feedback
**Describe the bug**
If you hit delete on your comment, nothing seems to happen. It does in fact delete it, but you have to navigate away and back again to see that it's gone

**To Reproduce**
1. Comment on link or note in collection or notebook
2. click delete
3. observe nothing seems to have happened
4. reopen comments and see it has in fact been deleted

**Expected behavior**
Comment should immediately disappear and also probably flash "deleted" or whatever

 - OS: arch linux
 - Browser qutebrowser
 - Base hash eljue

tylershuster:
fixes #4272

@liam-fitzgerald this works, is it sane?

### DIFF
--- a/pkg/interface/src/views/components/CommentItem.tsx
+++ b/pkg/interface/src/views/components/CommentItem.tsx
@@ -1,15 +1,13 @@
-import React, { useState } from 'react';
+import React, { useCallback } from 'react';
 import { Link } from "react-router-dom";
 import { Contacts } from '~/types/contact-update';
 import GlobalApi from '~/logic/api/global';
 import { Box, Row, Text } from '@tlon/indigo-react';
 import styled from 'styled-components';
 import Author from '~/views/components/Author';
-import { GraphNode, TextContent } from '~/types/graph-update';
-import tokenizeMessage from '~/logic/lib/tokenizeMessage';
+import { Post } from '~/types/graph-update';
 import { Group } from '~/types';
 import { MentionText } from '~/views/components/MentionText';
-import { getLatestCommentRevision } from '~/logic/lib/publish';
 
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/components/CommentItem.tsx
```diff
@@ -18,26 +16,26 @@ const ClickBox = styled(Box)`
...
-  const [revNum, post] = getLatestCommentRevision(comment);
+  const { ship, contacts, name, api, post, index, group } = props;
   const disabled = props.pending || window.ship !== post?.author;
 
-  const onDelete = async () => {
-    await api.graph.removeNodes(ship, name, [comment.post?.index]);
-  };
+  const onDelete = useCallback(async () => {
+    await api.graph.removeNodes(ship, name, [index]);
+  }, [post])
```
liam-fitzgerald, line 36: dependencies are wrong

liam-fitzgerald (changes requested):
minor dependency nitpick, else LGTM!

pkg/interface/src/views/components/CommentItem.tsx
tylershuster: D'oh

tylershuster:
Alright, I cleaned that up and the commit history

matildepark:
Honestly in my testing this seems just as spotty as the original behaviour — sometimes a delete just stays there until you navigate and come back; most of the time it disappears fairly quickly. On devnet comments do in fact disappear on a delete but not consistently, sometimes they hang around.

PR #4327, opened 2021-01-22
landscape: two hotfixes

matildepark:
- Formik was upgraded to 2.2.6, which broke form behaviours across the interface; we downgrade to the dictated peer version of 2.1.5 to keep parity with indigo-react. This is the critical fix, because it prevents DM and channel creation full stop.
- We ended up leaking additional Markdown tokens as of the line breaks fix deployed last week on devnet. Using RemarkBreaks and RemarkDisableTokenizers together, for whatever reason, breaks the latter, so you have to choose one or the other plugin. I fall back to disallowing the tokens for now, which strips their nodes (eg. `# this is header` will show as `this is header`), but it's better than rendering headers.

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -5,11 +5,11 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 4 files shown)


matildepark:
That diff is weird. Will never understand package-lock changes.

tacryt-socryp (approved):
Lord protect us from this hell

matildepark:
Forgot one more to stop `[]()` format.

PR #4328, opened 2021-01-22
meta: remove landscape template

matildepark:
In the interest of segregating our issues a bit more, I'll be transferring pre-existing Landscape issues to urbit/landscape. This amends our issue templates to follow suit.


philipcmonk (approved):
LGTM.  You could include something like the old "Chat, Publish, Links, Groups, Weather or Clock" list to help people who might not know what's a part of landscape

matildepark:
I was averse to that because Landscape isn't 'apps' to pilots, it's a few Gall agent with some interface 'templates', but I'll amend

PR #4331, opened 2021-01-25
notifications: lazy load correctly

liam-fitzgerald:
- Lazy loads notifications correctly
- Acknowledges the end of notifications (i.e. when they've all been loaded)

Fixes urbit/landscape#266


matildepark (approved):
This is more involved than I thought for a fix — how were we trying to load these before? Does this then help with https://github.com/urbit/landscape/issues/286? Still, quite an improvement, happy to stamp.

PR #4334, opened 2021-01-25
interface: add debug shim to store

liam-fitzgerald:
adds window.debugStore() to show past actions based on the tag and the
current state of some slice of the store. Logs the past 15 actions for
each tag.

cc: @tylershuster how can we get something similar for the zustand based store?

### DIFF
--- a/pkg/interface/src/logic/store/store.ts
+++ b/pkg/interface/src/logic/store/store.ts
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+
 import BaseStore from './base';
 import InviteReducer from '../reducers/invite-update';
 import MetadataReducer from '../reducers/metadata-update';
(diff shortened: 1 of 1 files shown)


tylershuster:
Yeah, I've been thinking I need to make an abstract class to build the state stores on top of anyway. If we intercept the `setState` function and add a `debugStore` function to that abstract class that all the state stores extend, we should be able to get a list of recent transformations per store

matildepark (changes requested):
This doesn't actually seem to function in practice. No matter where I navigate, what messages I send, or what notifications I dismiss, I get

image

liam-fitzgerald:
You have to pass it the mark of the update eg "harkUpdate"

matildepark (approved):
My bad.

PR #4335, opened 2021-01-26
peek date

gilletteonmeheleparim:
I needed to get current date in urbit format for Airlock metadata-store because it expects it only in this format `~2021.1.26..15.19.30..aa0f`

so I thought maybe i could scry current date from urbit but didn't find any way to do it so made this code change. I thought maybe hood was the right place but idk?

### DIFF
--- a/pkg/arvo/app/hood.hoon
+++ b/pkg/arvo/app/hood.hoon
@@ -193,8 +193,12 @@
   `..on-init
 ::
 ++  on-peek
-  |=  path
-  *(unit (unit cage))
+  |=  =path
+  ^-  (unit (unit cage))
+  ?+  path  (on-peek:def path)
+    [%x %date ~]
+    ``date+!>(now:bowl)
+  ==
 ::
 ++  on-agent
   |=  [=wire =sign:agent:gall]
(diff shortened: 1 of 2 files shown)


Fang-:
> I needed to get current date in urbit format for Airlock metadata-store because it expects it only in this format `~2021.1.26..15.19.30..aa0f`

You're not wrong in saying that an urbit can provide you with a string in this format, but I don't think it's the right approach here. You may want to render arbitrary dates, and you also just don't want to round-trip every time you need the current timestamp.

So, I'd strongly recommend writing an implementation of `@da` rendering in whatever context you're currently working in. This is the not-very-sexy but still important lower-level part of doing Airlock things: getting Earth to speak Martian, even if it's just a little bit.

The only thing to keep in mind is that Urbit deals with UTC times only, so make sure not to do any time localization during rendering.

(Yes, there should be an "Urbit library" for every language/environment that includes these kinds of conversions. Nobody has gotten around to it yet, but it'd be a fun little project.)

tylershuster:
This should be something that goes in `@urbit/api`...right?

gilletteonmeheleparim:
it's in https://github.com/urbit/urbit/blob/master/pkg/interface/dbug/src/js/lib/util.js#L14 currently
I wrote the same for my Kotlin Airlock now

PR #4338, opened 2021-01-27
added NPM packages

tylershuster:
This PR adds a few packages into the main urbit repo, as discussed offband. It doesn't modify landscape code, so let me know if another branch is more appropriate.

This adds the `pkg/npm` directory, and three packages to be published on npm:
- `http-api`: the library for connecting with eyre, taking parameters constructed in
- `api`: types and utilities for working with urbit
- `eslint-config`: the urbit eslint-config

Please take a look @liam-fitzgerald @matildepark @tacryt-socryp and let me know if other bits need to be in. There are a few `TODO`s, so I'm marking this as draft but in general it is ready for review.


pkg/npm/api/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "@urbit/api",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "types": "index.d.ts",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "Tyler Brown Cifu Shuster",
```
tylershuster, line 10: What should go here?

pkg/npm/api/tsconfig.json
```diff
@@ -0,0 +1,27 @@
...
+    "noUnusedLocals": false,
+    "noImplicitAny": false,
+    "noEmit": true,
+    "target": "es2015",
+    "module": "es2015",
+    "strict": true,
+    "jsx": "react",
+    "baseUrl": ".",
+    "paths": {
+      "~/*": ["src/*"]
```
tylershuster, line 18: Needs to use the babel-root-plugin

pkg/npm/http-api/.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
+name: CI
```
tylershuster, line 1: This file needs reworking @matildepark

pkg/npm/http-api/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2020 Tyler Brown Cifu Shuster
```
tylershuster, line 3: Probably not

pkg/npm/http-api/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@urbit/http-api",
+  "version": "1.1.0",
+  "license": "MIT",
+  "description": "Library to interact with an Urbit ship over HTTP",
+  "repository": "github:tylershuster/urbit",
```
tylershuster, line 6: Need to change this line

matildepark:
So RFC @liam-fitzgerald @tacryt-socryp @ixv — why isn’t this its own repo on the Urbit org? We have precedent with urbit/indigo-react; it discourages us from rapidly changing the API which we *shouldn’t* if other clients depend on the package. It means we have to update the API package stepwise with the livenet which is good. And it lets us get this complexity out of this repo where it seems to be pushing it.

liam-fitzgerald:
Yes, the http package should probably be moved out of repo. I'm wary of pulling the agent interfaces out into a seperate repo though, just because PRs in lockstep are a bad idea, as seen by some of the simultaneous indigo-landscape changes. We seem to be pretty all-in on the monorepo. Also worth noting that the monorepo doesn't prevent strict versioning of the npm packages.

matildepark:
Out of band conversation:

We seem to be concluded that this is a "flagship Airlock" and other Airlock libraries are descendent projects outside this repo.

Lingering TODOs need @tacryt-socryp @liam-fitzgerald commentary for graph methods.

tylershuster:
I think before I mark this as ready I want to deduplicate the type code so landscape relies on the newly created packages.

matildepark:
For the record [this seems like a fine package for automating publication](https://github.com/marketplace/actions/npm-publish). It will only publish if the version number is incremented in the appropriate file.

tylershuster:
@liam-fitzgerald was looking at Lerna; any thoughts?

matildepark:
It looks pretty heavy-duty; I might meet with him to ask about that.

liam-fitzgerald:
I suspect the complexity of lerna isn't worth it right now. But in the future where we package + version packages on a per-agent basis (i.e. `@urbit/graph @urbit/contact @urbit/groups` etc.) it will be. Let's just see how a lightweight system works, and we can switch to lerna if it's necessary

tylershuster:
Alright, anything you want to do? Once `release/next-userspace` lands I'll make the final changes and mark this

tylershuster:
@jtobin Are you the person I should talk to about implementing this: https://github.com/marketplace/actions/npm-publish#github-action

tylershuster:
Short of the deployment script mentioned above, which still needs implementation, this is ready to go. As part of splitting out the eslint config, I've started (not quite completed) to work down the linting errors in the codebase. Likely I will make another pass at this with the next PR.

In general though, this completely separates out the type system into its own folder. Right now you need to use `npm link` in order for it to all connect. I've put some of the utility poke functions into the `npm/api` folder but haven't implemented them because it needs the new HTTP API architecture in place to utilize. For this reason I would be fine holding off on publishing the future `@urbit/api` package until the state hook refactor which has been blocked on this PR.

liam-fitzgerald (changes requested):
Looks good! Get rid of the lockfiles, and this can go in.

jtobin:
@tylershuster for some insane reason that I don't care to delve into now, git is hanging when trying to fetch from your (and only your) fork.  I was gonna PR a workflow addition, but on second thought, it'd be better to add the workflow in another PR from the urbit/urbit repo itself anyway.  Secrets (like `NPM_TOKEN`) in GitHub Actions aren't accessible in PR's made from forks, so we wouldn't be able to test it here.

So: when this is merged I can put together the workflow script in a separate PR, if you'd like.  Is what's desired literally "publish these three packages to NPM on every push to master, so long as a version number has been incremented?"

tylershuster:
@jtobin yes that's the gist of it. I don't think we need to do anything more complicated than that for now. What you say sounds good.

tylershuster:
A solution that allows for other packages to be added later would not go amiss, either

matildepark:
Tagging @liam-fitzgerald for final stamp before merge here. @tylershuster:

> In general though, this completely separates out the type system into its own folder. Right now you need to use `npm link` in order for it to all connect. I've put some of the utility poke functions into the `npm/api` folder but haven't implemented them because it needs the new HTTP API architecture in place to utilize. For this reason I would be fine holding off on publishing the future `@urbit/api` package until the state hook refactor which has been blocked on this PR.

If I read that right and this changes the build process and adds a step (ie. `npm link`), you may want to add that to pkg/interface/CONTRIBUTING.

tylershuster:
No, you're skipping the "right now" — because `@urbit/api` doesn't exist on NPM you need to use `npm link` which allows you to use a local package as if it were remote. 

However, you're right that we may need to change the `CONTRIBUTING` doc, probably after the automation is actually set up. That's because if you want to develop on the API, you need to `npm link` it, or else it will get its code from the frozen (NPM) version instead of the malleable (local) version.

liam-fitzgerald (approved):

matildepark (changes requested):
In conversation with @liam-fitzgerald this is _not_ ready for merge because it's pending API changes made in next-userspace; @liam-fitzgerald will push those changes to the branch or PR against the fork's branch with those changes sometime today

PR #4340, opened 2021-01-27
interface: consistent hover events

jamesacklin:
Changes the mouse event handler so the notification dismissal control persists between actions, regardless of browser. 

Fixes urbit/landscape#265


liam-fitzgerald (approved):
LGTM! Confirmed to work!

PR #4341, opened 2021-01-27
contacts: update profile overlay

matildepark:
- Use mono for ship names in Leap
- Weaves the now-flattened contacts prop through Chat
- Adds the new look for the mini profile box with an editable status message and dropdown menu, just has a stub for the API action.

Questions for other JS wizards (@tylershuster @jamesacklin @liam-fitzgerald)

- I'm trying to be super clever with `contentEditable` so that if it's *our card* you can just click and change the status in the message, setting it on blur, but React seems unhappy about having a controlled component `contentEditable`.
- `history.push()` doesn't work in the dropdown inside the profile overlay. I do not know why. Other dropdowns don't have to do anything to the context.

image


tylershuster:
contenteditable is notorious for inconsistent behavior. Maybe you could try https://styled-components.com/docs/api#as-polymorphic-prop ?

tacryt-socryp (approved):

PR #4343, opened 2021-01-28
interface: adds serviceworker

tylershuster:
This adds a serviceworker with the boilerplate from https://developers.google.com/web/tools/workbox/guides/get-started and starts to cache images. Seems...faster?

Preliminary work to out-of-tab browser notifications and other wondrous splendiferousnesses

@liam-fitzgerald RFC

### DIFF
--- a/pkg/arvo/app/landscape/index.html
+++ b/pkg/arvo/app/landscape/index.html
@@ -24,6 +24,13 @@
     <div id="portal-root"></div>
     <script src="/~landscape/js/channel.js"></script>
     <script src="/~landscape/js/session.js"></script>
+    <script>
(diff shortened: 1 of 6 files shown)


jamesacklin (approved):
I'm not super-familiar with workbox, but this behaves as expected. It appears to properly cache images and display them when the browser is offline.

liam-fitzgerald (changes requested):
Looks ok, will have to add this to the glob so that it'll work in production. Will PR such a change myself.

tylershuster:
Great, I'll clean up the boilerplate

tylershuster:
Obviously this is just a start to get the fiddly bits out of the way to do the real work

PR #4344, opened 2021-01-29
interface: sigil color for comets

jamesacklin:
Comets need sigils with background colors too. Passes the color in to the Box used by comets (shipname > 14 chars).

Also a formatting change, but I noted the operative line with a comment.

Fixes urbit/landscape#316

### DIFF
--- a/pkg/interface/src/logic/lib/sigil.js
+++ b/pkg/interface/src/logic/lib/sigil.js
@@ -8,46 +8,58 @@ export const foregroundFromBackground = (background) => {
     g: parseInt(background.slice(3, 5), 16),
     b: parseInt(background.slice(5, 7), 16)
   };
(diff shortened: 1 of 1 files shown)


pkg/interface/src/logic/lib/sigil.js
```diff
@@ -8,46 +8,58 @@ export const foregroundFromBackground = (background) => {
...
+    padded = false
+  }) => {
+    const padding = icon && padded ? '2px' : '0px';
+    const innerSize = icon && padded ? Number(size) - 4 : size;
+    const foregroundColor = foreground
+      ? foreground
+      : foregroundFromBackground(color);
+    return ship.length > 14 ? (
+      <Box
+        backgroundColor={color}
```
jamesacklin, line 35: This is the operative change.

liam-fitzgerald (approved):
LGTM!

PR #4345, opened 2021-01-29
profile: more stubs

matildepark:
- We renamed contacts to groups in associations, so I just replaced that across the interface.
- An attempt at making GroupSearch multi-group select, adjusting its behaviour in Groupify to compensate.
- Incorporates GroupSearch on the profile page.
- Stubbed API calls for setting groups.

Can't test this much since we can't make groups and metadata-store crashes but it's 90% of the way.


tacryt-socryp (approved):
LGTM

PR #4347, opened 2021-01-30
language-server: fixes & improvements

liam-fitzgerald:
- Fixes parsing of identifiers for hover, autocomplete etc. 
- Improves formatting of hover items
- Ignores /sys files due to their length and nonstandard nature
- Prevent %play-open on %zpgl
- Retains state on reload


philipcmonk (approved):
LGTM, sorry for the delay

PR #4350, opened 2021-01-31
interface: clears 'copied' message on shipname click

jamesacklin:
Sets and clears the chat message author's name on click with hooks, rather than with a ref. Sets this message in Inter regardless if the author has set a nickname or not.

Fixes (urbit/landscape#331)[https://github.com/urbit/landscape/issues/331]


liam-fitzgerald (approved):
LGTM! I'm a sucker for abstract hooks, and we actually use this elsewhere (in the base hash indicator). I suppose the rule of three applies.

PR #4351, opened 2021-02-01
chat: set initialisation time to 100ms

matildepark:
Eliminates the friction quite a bit! See urbit/landscape#332.

cc @tylershuster


tacryt-socryp:
What does the initialization time constant *do*?

tylershuster:
￼Gives the renderer time to calculate the size of the chat items￼. Includes loading images. ￼

matildepark:
Right now each time you click a chat you wait 1.5s with the placeholder. But we don’t really need it. And the 1.5s itself adds up...

liam-fitzgerald (approved):
Looks good, plays ok on livenet

matildepark:
We'll see how next-js reacts and adjust if need be.

PR #4352, opened 2021-02-01
interface: changes wrapping behavior

tylershuster:
Fixes https://github.com/urbit/landscape/issues/325

Slightly different behavior — sometimes wraps on small screens.

image


liam-fitzgerald (approved):
LGTM!

PR #4353, opened 2021-02-01
publish: fixes codeblock wrapping behavior

tylershuster:
fixes https://github.com/urbit/landscape/issues/324

Seems fine to `overflow: hidden` `Note`s, right?


liam-fitzgerald (approved):
yeah that's fine. LGTM!

PR #4354, opened 2021-02-01
chat: fixes banner pluralization

tylershuster:
Fixes https://github.com/urbit/landscape/issues/322


liam-fitzgerald (approved):
LGTM, refactoring it into a `pluralize` function could be nice, but obviously the vagaries of language tend to resist abstraction (irregular pluralization and such).

PR #4356, opened 2021-02-01
contact: networking updated to send out data with the appropriate resources

tacryt-socryp:
We are now federating all contact data sent out via groups, whether via DMs or otherwise. We join contacts upon joining a group, leave it upon leaving a group, and stop serving data from the push-hook when we delete the group.

Needs testing, but is a logically complete implementation of the core networking functionality.


pkg/arvo/app/contact-pull-hook.hoon
```diff
@@ -32,14 +33,11 @@
...
+++  resource-for-update  resource-for-update:con
 ++  on-pull-nack
   |=  [=resource =tang]
   ^-  (quip card _this)
-  :_  this
-  ?~  (get-contact:con entity.resource)  ~
-  =-  [%pass /pl-nack %agent [our.bowl %contact-store] %poke %contact-update -]~
-  !>  ^-  update:store
-  [%remove entity.resource]
+  [~ this]
```
liam-fitzgerald, line 40: Why has this nack behaviour been removed?

pkg/arvo/app/group-view.hoon
```diff
@@ -152,43 +152,60 @@
...
-      ::
+    ::
         %pull-md
       ?>  ?=(%poke-ack -.sign)
       (ack +.sign)
-      ::
+    ::
+        %pull-co
+      ?>  ?=(%poke-ack -.sign)
+      (ack +.sign)
```
liam-fitzgerald, line 173: I don't think these contact pokes should use the +ack gate, because it'll report the join transaction as failed if the poke nacks, which isn't necessarily true. Maybe just +ignore-ack where
```hoon
++  ignore-ack
  |=  err=(unit tang)
  ?~  err  jn-core
  %-  (slog u.err)
  jn-core
```

liam-fitzgerald (approved):
Looks good, have some questions, but nothing really blocking

pkg/arvo/app/group-view.hoon
tacryt-socryp: I think the transaction should fail tho. Talking about this on call with you rn, but documenting discussion lol

pkg/arvo/app/contact-pull-hook.hoon
tacryt-socryp: no longer relevant given implementation details changing

pkg/arvo/mar/contact/share.hoon
```diff
@@ -0,0 +1,12 @@
+|_  share=[%share =ship]
+++  grad  %noun
+++  grow
+  |%
+  ++  noun  share
+  --
+::
+++  grab
+  |%
+  ++  noun  share
```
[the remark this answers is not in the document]
tacryt-socryp: fixing

pkg/interface/src/views/apps/chat/components/ShareProfile.js
```diff
@@ -1,16 +1,100 @@
...
+    if (!res) { return; }
+    if (!group) { return; }
+    console.log(group);
+    if (group.hidden) {
+      // TODO:
+      // take the union of the pending set and the members set,
+      // subtract ourselves, then if *anyone* has not already been shared with,
+      // show the banner
+      // Promise.all all the members of the set
+      let check =
```
tacryt-socryp, line 42: TODO

pkg/interface/src/views/apps/chat/components/ShareProfile.js
```diff
@@ -1,16 +1,100 @@
...
+      <Sigil
+        ship={window.ship}
+        size={16}
+        color={`#${uxToHex(props?.our?.color)}` || "#000000"}
+        icon />
+    </Row>
+  );
+
+  const onClick = () => {
+    api.contacts.allow(recipient).then(() => {
```
tacryt-socryp, line 91: TODO

PR #4357, opened 2021-02-01
chat: add stubbed share banner

matildepark:
image

Needs conditional render logic ("have we already shared?") and hooking into API call to share.


tacryt-socryp (approved):
LGTM

PR #4358, opened 2021-02-02
Contacts: cleanup GroupSearch, allow add/remove of pinned groups from profile

liam-fitzgerald:
- Cleanups GroupSearch to prevent funky behaviour cause by formik and its field comparisons
- Allows for add/remove of pinned profiles from groups


tacryt-socryp (approved):
LGTM, much needed!

PR #4359, opened 2021-02-02
contact-hooks: some improvments 

liam-fitzgerald:
- Adds %no-validate flag, to prevent checking the wire/src.bowl/resource all match-up to enable P2P connections. 
- Moves the resource-for-update back into the inner door, to match the push-hook.
- Fixes `+get-contact` and `+is-allowed`

This is about as much as I can do given what I know. Unsure how you intend to share contacts to the hosts, but contacts that are on the host are propagated fine to members of the group.


tacryt-socryp (approved):
LGTM, great improvements!

PR #4363, opened 2021-02-02
landscape: direct message workspace

matildepark:
Screen Shot 2021-02-02 at 6 44 52 PM

- Removes a ton of deprecated chat-store related FE logic.
- Adds a "messages" workspace. Conditional renders are patched all over the place. URLs are amended per module type in Leap, etc. to match the new URL format.
- New DM flow uses the New Channel form inside a dropdown now:

image

You can't name it until you make it. If you list only one ship it does the usual redirect flow; otherwise it creates an unnamed group DM.

- New Channel now doesn't require you to name the channel, falling back to the module type name if blank. Only DMs allow for channel name clobbers; home and group workspaces both use random number namespacing (`channelpath-2233`).

- "DMs + Drafts" is now "My Channels". It includes all unmanaged channels except chats. The new channel flow does not include chats in that workspace.


matildepark:
Oh, might've missed link construction in hark.

liam-fitzgerald (approved):
Tested, LGTM!

PR #4364, opened 2021-02-03
king: Fix lane serialization problem in king v1.1

### LINKED ISSUE #4346: Ship won't boot after v1.1 urbit-king. Assertion 'c3y == u3r_safe_chub(lan, &lan_d)'
**Describe the bug**
After briefly trying out the new v1.1 haskell urbit-king binary, I'm now unable to boot my ship with either ./urbit-king or or ./urbit.

My first run on urbit-king seemed to go fine, but I shut it down after noticing the web ui wasn't running on the port I wanted (it was on 8080 instead of 80). 

When I try to boot again with urbit-king, I now receive this error and am spit back out to the terminal prompt. 
`libnuma: Warning: Unable to determine max cpu (sched_getaffinity: Function not implemented); guessing...`

And when I try to boot with the normal urbit binary, I get this after `playback complete`
```
vere: checking version compatibility
ames: live on 59785
http: web interface live on http://localhost:8080
pier (946502): live
ames: czar ten.urbit.org: ip .104.196.239.18
Assertion 'c3y == u3r_safe_chub(lan, &lan_d)' failed in vere/io/ames.c:408

bail: oops
bailing out
serf: pier unexpectedly shut down
Aborted (core dumped)
```

 - Ubuntu 20.04.2 (Digitalocean droplet) 
 - Urbit v1.1
 - ~mirmus-dozryx

I'm not very educated on the inner-workings of Urbit yet, but I'm happy to provide any other info needed. Thanks very much!

### LINKED ISSUE #4361: urbit-king fails silently when `.vere.lock` already exists, even when no process is running
When trying to boot my ship with `urbit-king`, the program fails silently and exits, printing only `libnuma: Warning: Unable to determine max cpu (sched_getaffinity: Function not implemented); guessing...`

I ran the process with all the debug flags I could find (`--verbose --log-to-stderr --log-debug --trace`). Here's the output:

```bash
libnuma: Warning: Unable to determine max cpu (sched_getaffinity: Function not implemented); guessing...
2021-02-02 10:09:50.029145: [info] ("EYRE","MULTI",MultiEyreConf {mecHttpsPort = Nothing, mecHttpPort = Nothing, mecLocalhostOnly = False})
2021-02-02 10:09:50.039280: [info] Taking lock file: piers/sipfyn-pidmex/.vere.lock
2021-02-02 10:09:50.039714: [error] Unable to acquire lock file: "piers/sipfyn-pidmex/.vere.lock": piers/sipfyn-pidmex
```

The culprit appears to be a leftover vere lockfile, as I had no previous urbit process running.

File permissions don't seem to have any effect (i.e. running `chmod a+w .vere.lock` does not solve the problem).

Removing / renaming the lock file works as a band-aid fix.

**System**
 - OS: Ubuntu 20.04
 - Vere: Urbit v1.1

pilfer-pandex:
I accidentally kept on using the old format for lanes in king v1.1, with the result that vere would error out when encountering these malformed lanes.

Changes:
- serialize and deserialize lanes properly using the new 6-byte format
- change vere to ignore %sends with uninterpretable lanes, rather than crashing. We need to do this because the people who ran kh 1.1 have those lanes in their state. They'll get flushed out over time. KH already handles errors this way.
- Also add an error message if the king is unable to acquire the lock file.

Fixes #4361.
Fixes #4346.

Test plan:
- Having run ~pp on kh 1.1, I verified that starting vere 1.1 would result in the reported crash, but staring my modified vere wouldn't.
- While running ~pp elsewhere, I tried to start it anyway and observed the new error message
- (Pending.) Rerun my stateless forwarding ames test procedure as described in #3595


pkg/urbit/version
```diff
@@ -1 +1 @@
-1.1
\ No newline at end of file
+1.1.1
```
joemfb, line 1: We're not using semver; this'll be `1.2`.

pilfer-pandex:
@belisarius222 registers his approval to me by text message and says he has temporary issues authenticating. I'm satisfied with this (and am confident the texter isn't an impostor), and will regard it the same as if he had used the GitHub ui.

PR #4366, opened 2021-02-03
wip GCP storage in hoon

mrdomino:
@Fang- @vvisigoth am I barking up the right tree here?

### DIFF
--- a/pkg/arvo/app/gcp-store.hoon
+++ b/pkg/arvo/app/gcp-store.hoon
@@ -1,0 +1,117 @@
+::  gcp-store [landscape]:
+::
+::  stores GCP credentials for object storage. This is
+::  "heavily influenced by" s3-store.hoon, and probably
+::  should have the common logic extracted some day.
+::
+/-  *gcp
+::  TODO?  /+  gcp-json
+/+  default-agent, verb, dbug
+~%  %gcp-top  ..part  ~
+|%
++$  card  card:agent:gall
++$  versioned-state
+  $%  state-zero
+  ==
+::
++$  state-zero  [%0 =credentials =configuration]
+--
+::
+=|  state-zero
+=*  state  -
+::
+%-  agent:dbug
+%+  verb  &  :: TODO: |
+^-  agent:gall
+~%  %gcp-agent-core  ..card  ~
+|_  =bowl:gall
++*  this       .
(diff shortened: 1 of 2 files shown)


tacryt-socryp:
What does the refresh-token do? Is there a need to periodically perform an http request to get a new token or something?

mrdomino:
If I'm not mistaken it's a bearer token, sent with the `Authorization` header on requests or somesuch. And it does indeed expire periodically and need to be refreshed.

I think per @brendanhay that I should ultimately store it in the browser along with the expiry, and refresh it from the oauth grant stored in the ship as needed.

tylershuster:
I think we're probably putting settings stuff in `settings-store` (https://github.com/urbit/urbit/pull/4276) from now on but I don't think S3 has been changed

matildepark:
Yeah we’ll get to s3. His days are numbered.

mrdomino:
Sweet. I am tempted to stick in a bucket called `gcp-oauth`. Do I have the right idea?

For the storage configuration itself, `s3-store` seems to want there to be a `(set @t)` of (storage) buckets. At a first glance it doesn't seem like `settings-store` handles this straightforwardly. What do we want to do about that? (Thoughts that come to mind: make `settings-store` handle sets, pretend that we only ever needed to know the current bucket and get rid of the set, or store a comma-separated list of buckets.)

At the risk of incurring scope creep on this interim project, it seems like it'd be fine to just genericize the storage configuration between s3 and gcp while we're at it, since the former has to change anyway. Thoughts?

tacryt-socryp:
Storing the current bucket only is the best approach, I think.

mrdomino:
oh huh, I guess perhaps I want to be on release/next-userspace?

mrdomino:
Hey @ixv — how do I use `settings-store`?

ixv:
Settings-store state is `(map key (map key value))` the idea being that the we have seperate buckets of key-value pairs. A key is just a @tas, and a value is basically a hoon json type without arrays or objects.

To write, poke settings-store with a %settings-event mark, the type for which is defined in sur/settings.hoon as +event. To read, you can scry or subscribe on the paths `/all`, `/bucket/<bucket-key>`, and `/entry/<bucket-key>/<entry-key>`

PR #4367, opened 2021-02-03
interface: fix profile overlay positioning in chat

jamesacklin:
Restores the intended dynamic positioning of the profile overlay in chat messages. Calculates the space between a chat message and the vertical bounds of the chat window using client geometry, rather than scroll offset, which would sometimes throw a false positive and position the profile overlay contents below the chat window.

Fixes [urbit/landscape#303](https://github.com/urbit/landscape/issues/303)


matildepark (approved):
Overdue on our part. Thanks so much. Remedies issue on common cases in my tests.

PR #4369, opened 2021-02-03
meta: update janeway dep in glob action

jtobin:
This should unbreak the glob action.  I recently pushed an image for the latest janeway version, but am not yet sure why the old one became unavailable..


matildepark (approved):

jtobin:
Let me know if this still continues to be an issue.  The old images should still be available on dockerhub, so the error seen in CI might have just been a random one-off networking problem.  The janeway version can stand to be bumped regardless.

PR #4370, opened 2021-02-03
Fix joining DMs and a few visual issues

tacryt-socryp:


matildepark (approved):
lgtm

pkg/interface/src/views/apps/chat/components/ShareProfile.js
```diff
@@ -83,7 +83,7 @@ export const ShareProfile = (props) => {
       p={1}
       alignItems="center"
       borderRadius={2}
-      backgroundColor={!props.our ? `#${uxToHex(props.our.color)}` : "#000000"}>
+      backgroundColor={!!props.our ? `#${uxToHex(props.our.color)}` : "#000000"}>
```
matildepark, line 86: Note that you don't have to explicitly cast to a boolean, you can just write `props.our`. Our linter points this out a lot.

tacryt-socryp: aww

PR #4372, opened 2021-02-03
interface: popover settings style fixes

tacryt-socryp:
Fixed flex things bunching up into each other and added "View profile" option to Participants component.

### DIFF
--- a/pkg/interface/src/views/landscape/components/ChannelPopoverRoutes/Notifications.tsx
+++ b/pkg/interface/src/views/landscape/components/ChannelPopoverRoutes/Notifications.tsx
@@ -25,7 +25,7 @@ export function ChannelNotifications(props: ChannelNotificationsProps) {
   };
 
   return (
-    <Col mb="6" gapY="4">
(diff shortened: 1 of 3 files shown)


pkg/interface/src/views/landscape/components/ChannelPopoverRoutes/Notifications.tsx
```diff
@@ -25,7 +25,7 @@ export function ChannelNotifications(props: ChannelNotificationsProps) {
   };
 
   return (
-    <Col mb="6" gapY="4">
+    <Col mb="6" gapY="4" display="block">
```
matildepark, line 28: `Col`'s whole function is that it's shorthand for `display: flex; flex-direction: column` — is there a reason `flexShrink={0}` won't suffice here?

tacryt-socryp: It did suffice. Changed it.

matildepark (approved):

PR #4375, opened 2021-02-04
interface: allow setting of status from ProfileOverlay

tacryt-socryp:


pkg/interface/src/views/components/ProfileOverlay.tsx
```diff
@@ -138,7 +143,7 @@ class ProfileOverlay extends PureComponent<ProfileOverlayProps, {}> {
                   color='black'
                   cursor='pointer'
                   fontSize={0}
-                  onClick={() => history.push('/~profile/~' + window.ship)}>
+                  onClick={() => history.push(`/~profile/~${ship}`)}>
```
tacryt-socryp, line 146: fixed bug of always going to our own profile

matildepark (approved):

PR #4378, opened 2021-02-04
NewChannel: unconditionally add uid to channel name

liam-fitzgerald:
Fixes urbit/landscape#273


matildepark:
Pretty sure I fixed this in next userspace — everything but DMs add uids so we can close this.

PR #4380, opened 2021-02-04
VirtualScroller: performance improvements

liam-fitzgerald:
curry and memoize the measure function, to prevent needless rerenders.
Mark ChatMessage as a PureComponent, to capitalize on prop equality.
Bind the onScroll handler in didMount so we can mark it as passive,
easing render times.

This causes a local copy of this branch (i.e. development, no optimisations etc.) to be about as fast, if not faster than the current mainnet production build. 
cc: @tylershuster have I done anything stupid?

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -50,7 +50,7 @@ interface ChatMessageProps {
   highlighted?: boolean;
 }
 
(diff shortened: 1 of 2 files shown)


matildepark:
Change the word ‘development’ to ‘production’ in the dev server config (webpack.dev etc) and try it there

liam-fitzgerald:
going to rework a little bit more

tylershuster:
Tag me again when you re-request this. I have been out today, but I am happy to take a look￼

PR #4384, opened 2021-02-05
contact-store: prevent hook forwarding loop

liam-fitzgerald:
This is a possible fix for the forwarding loop problem. I'm unsure what the best way to address this is. Maybe a `+validate-update` arm in the pull-hook? 

The forwarding loop occurs like so.

1 host subscribes to a members profile.
2 member subscribe to the groups contact information.
3 member updates their profile, causing a fact to be sent to their push-hook
4 this causes the host to receive the same fact, which they then forward onto the groups contact subscription.
5 the member then receives the fact on the groups subscription, which goes to the store and then the push hook
6 GOTO 4

### DIFF
--- a/pkg/arvo/app/contact-store.hoon
+++ b/pkg/arvo/app/contact-store.hoon
@@ -137,6 +137,9 @@
       ++  edit-contact
         |=  [=contact:store edit=edit-field:store]
         ^-  contact:store
+        ::  ensure difference
+        =;  new=contact:store
+          ?<(=(contact new) new)
         ?-  -.edit
           %nickname  contact(nickname nickname.edit)
           %bio       contact(bio bio.edit)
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
One simple option is to make the updates to contact-store idempotent.

liam-fitzgerald:
I'm actually convinced this is the best solution, after thinking about it some more. This is because you can have a forwarding loop with any number of group hosts, not just one. Crashing (or noopping) if it's identical stops these duplicate facts, regardless of the sender.

tacryt-socryp:
Yes, your solution makes this idempotent. But are the other action types idempotent as well?

tacryt-socryp:
Update: the %remove action is idempotent, and I have now made the %add action idempotent. All potential forwarding loops have been removed.

matildepark (approved):
As per @philipcmonk:

> That looks like exactly the right solution

So I am comfortable getting this into staging. Please hold.

PR #4387, opened 2021-02-07
chat: fix OverlaySigil positioning

jamesacklin:
Makes a few changes to the `OverlaySigil`, `MessageWithSigil`, `MentionText`, and `Author` components to consistently and predictably display the sigil overlay.

Refactors `OverlaySigil` as a functional component with hooks:

1. Makes `scrollContainer` optional so we can extend this clip-prevention method to other scrolling panes (such as Notebooks or Collections).
2. Makes the parent element responsible for showing and hiding the overlay with state, but still handles the event with `onDismiss()`.
3. Adds a `visibility` state to measure the rendered-but-invisible overlay before relocating it, then showing it when the relocation is finished.
4. Divorces the mini-sigil/avatar display from this component to allow for its use in a variety of contexts (text triggers and the like).

- In keeping with point 2 above, `Author`, `MessageWithSigil`, and `MentionText` are now all responsible for handling the `OverlaySigil`'s visibility on their own. This was to keep the click handler and state as close to the interactive element as possible; totally happy to go back on this.
- `MessageWithSigil` and `Author` now display their own mini-sigils (see point 4 above). I did this to simplify the `OverlaySigil` component a bit and make way for non-image click triggers (like the `MentionText`), but could easily bring this back in.

Feel free to squash this on merge; I'm not sure how many commits constitutes a faux pas.

Fixes urbit/landscape#303


matildepark:
I'm gonna target release/next-userspace here specifically because the component is changing a little in the upcoming release.

matildepark:
Fairly confident in that merge conflict resolution but it may be worth cracking open a fakezod and looking. Unsure if you know yet; but you'd build the Urbit binary with Nix (see root install instructions) or just download the latest from urbit.org/install inside the root of the repo here and

```
./urbit -F zod -B "bin/solid.pill" -A "pkg/arvo"
```

while next userspace is checked out. Then you'd point urbitrc to localhost:80.

jamesacklin:
@matildepark Sounds good, I will check it out shortly. Thank you for pointing the changes to the right place.

jamesacklin:
Works as intended on `next-userspace`, thank you for the guidance.

pkg/interface/src/views/components/OverlaySigil.tsx
```diff
@@ -1,134 +1,103 @@
...
+
+  useEffect(() => {
+    updateContainerOffset();
+    if (scrollWindow)
+      scrollWindow.addEventListener('scroll', updateContainerOffset);
+    return () => {
+      if (scrollWindow)
+        scrollWindow.removeEventListener('scroll', updateContainerOffset, true);
+    };
+  }, []);
```
liam-fitzgerald, line 75: kinda wondering if this should have `scrollWindow` as a dependency, given that it looks like `this.window` can change in virtualscroller. cc: @tylershuster

pkg/interface/src/views/components/OverlaySigil.tsx
```diff
@@ -1,134 +1,103 @@
...
-      </Box>
-    );
-  }
-}
-
-export default withLocalState(OverlaySigil, ['hideAvatars']);
+        contact={contact}
+        group={group}
+        history={history}
+        onDismiss={() => onDismiss()}
```
liam-fitzgerald, line 94: any reason for the unnecessary arrow functioN?

pkg/interface/src/views/components/MentionText.tsx
```diff
@@ -36,57 +31,49 @@ export function MentionText(props: MentionTextProps) {
...
-    setShowOverlay(true);
-  }, [setShowOverlay]);
-
   const group = props.group ?? { hidden: true };
+  const [showOverlay, setShowOverlay] = useState(false);
 
-  const history = useHistory();
+  const toggleOverlay = () => {
+    setShowOverlay((value) => !value);
+  };
```
liam-fitzgerald, line 51: wrapping this in `useCallback` might be nice, just to dodge unnecessary rerenders

liam-fitzgerald (approved):
Works in testing, have a few non-blocking comments but LGTM!

pkg/interface/src/views/components/OverlaySigil.tsx
jamesacklin: Sure, that makes sense. I made it optional because we also use this in notebooks and posts, where I don't think we virtualize the content and thus don't drill the scrollWindow ref all the way down. Totally open to revising.

pkg/interface/src/views/components/MentionText.tsx
jamesacklin: Thanks, will address

pkg/interface/src/views/components/OverlaySigil.tsx
jamesacklin: None, will fix

pkg/interface/src/views/components/OverlaySigil.tsx
tylershuster: Yep, it should depend on that. There's already some weird behaviors (some resolved, others outstanding) when switching between chats due to this

pkg/interface/src/views/components/OverlaySigil.tsx
jamesacklin: My mistake -- this is calling the parent function that's passed in, which sets that `showOverlay` flag one component up. Is there a better way to handle this?

pkg/interface/src/views/components/OverlaySigil.tsx
jamesacklin: Just for clarity, should I leave as-is and make a separate task to pass the window ref down in other apps, or do that in this PR?

pkg/interface/src/views/components/OverlaySigil.tsx
tylershuster: Just do `onDismiss={onDismiss}`

pkg/interface/src/views/components/OverlaySigil.tsx
tylershuster: The empty array `[]` at the end of the `useEffect` declaration should be `[scrollWindow]` instead — you're declaring that the effect declared depends on the identity of `scrollWindow` — if it changes, we want to recalculate the listeners because the element has changed.

jamesacklin: Understood, I realize now @liam-fitzgerald meant `this` as `this effect`, not `this component`. Thank you!

pkg/interface/src/views/components/OverlaySigil.tsx
tylershuster: Actually you could just remove `api`, `color`, `contact`, `group`, and `history` from the destructuring of props and they'd get passed in with `...rest`

PR #4388, opened 2021-02-08
interface: add overlay tutorial

liam-fitzgerald:
This adds a simple overlay tutorial to landscape. The architecture is a bit funky, so i'll explain it a bit.
We use `useTutorialModal` to mark a component as part of the tutorial, passing it the step that it's a part of, whether this particular instance is part of the tutorial and a ref for the popover to position itself against. This ref is then used to position the popover. 

This works fine as is, and i'm hosting the beginners island on ropsten. Some questions though:
- The way this is currently setup, we need to join the group before we can initiate the tutorial. Is this an acceptable tradeoff?
- We almost certainly want the beginner island group join to automatically join all the channels. How should this work? I've been kicking around the idea of a pinned channel also being joined automatically (i.e. a 'default loadout') from the group owner. If so, we could just pin all the channels.
- How should we handle error cases? Group failed to join etc. 
- The designs call for `x days remaining` indication on the tutorial tile. Is it worth writing another policy for this? Else how can we implement?

Capture d’écran 2021-02-08 à 11 05 56 AM
Capture d’écran 2021-02-08 à 11 06 12 AM
cc:
@jyng 
@tylershuster I'm working with the new zustand stuff, anything I missed? Is it worth refactoring the tutorial state into a seperate store?
@ixv I'm adding a setting in the +on-load of settings-store, partly because I can't tell the difference between a setting that hasn't been created and one that has, but hasn't loaded yet. Is this ok?


matildepark:
Oh excellent! Will review in the morning.

pkg/interface/src/logic/state/local.tsx
```diff
@@ -41,7 +72,10 @@ const useLocalState = create<LocalState>(persist((set, get) => ({
   })),
   set: fn => set(produce(fn))
   }), {
-  blacklist: ['suspendedFocus', 'toggleOmnibox', 'omniboxShown'],
+    blacklist: [
```
tylershuster, line 75: Do we want to not keep track of tutorial progress across sessions?

matildepark:
> The designs call for x days remaining indication on the tutorial tile. Is it worth writing another policy for this? Else how can we implement?

behn it after joining? Or handle it in fe. I’d prefer if we did this in back end for multiple session support. Certainly there’s some observe hook thing we could do?

tylershuster:
Seems fine to me (didn't run, just reviewed code) except the noted question. I think this is a fine place to put the state. Soon we'll transition it to settings-store so we don't get prompts across devices.

ixv:
I think its fine to put add a setting in the on-load of settings-store, since there doesn't seem to be any other more appropriate place to put it.

matildepark:
@urcades says we can triage out the timer if need be and let people leave when they're done or hang out.

pkg/interface/src/views/apps/launch/app.js
```diff
@@ -51,12 +65,60 @@ export default function LaunchApp(props) {
...
+  }, [props.settings]);
+
   return (
     <>
       <Helmet defer={false}>
         <title>{ props.notificationsCount ? `(${String(props.notificationsCount) }) `: '' }Landscape</title>
       </Helmet>
       <ScrollbarLessBox height='100%' overflowY='scroll' display="flex" flexDirection="column">
+        {modal}
         <Welcome firstTime={props.launch.firstTime} api={props.api} />
```
matildepark, line 122: This can die now by the way.

urcades:
Yeah — the main POV here is that if the development overhead for
implementing a timer is vast, the experiential 'benefit' of automatically
removing someone from a group seems low-value.

—
~fabled-faster
https://urbit.org/

matildepark:
That and I can imagine there being people who want to voluntarily stay as guides.

pkg/arvo/app/settings-store.hoon
```diff
@@ -31,7 +31,10 @@
     ^-  (quip card _this)
     =/  old  !<(versioned-state old-vase)
     ?-  -.old
-      %0  [~ this(state old)]
+        %0  
+      :_  this(state old)
+      =-  [%pass / %agent [our dap]:bowl %poke -]~
+      settings-event+!>([%put-entry %tutorial %seen b+%|])
```
matildepark, line 37: What's `b+%|` as a default value here?

matildepark:
> We almost certainly want the beginner island group join to automatically join all the channels. How should this work? I've been kicking around the idea of a pinned channel also being joined automatically (i.e. a 'default loadout') from the group owner. If so, we could just pin all the channels.

For now I'm more apt to recommend an autojoin in this one-off case; the trouble here is that your suggestion redefines what "pinning" is. You could imagine a group wanting to pin its rules so that they're autojoined when you join the group, but then the rules show up as a "featured channel" for some reason. The featured channels and the autojoin functionality are not necessarily correlated.

jyng:
Some background on Beginner Island membership:
- The ephemeral nature is to keep the user level relatively the same to promote friendliness. I think joining something as a newcomer and bonding over that could help newcomers feel a lot more cozy.

Regarding timer concerns:
- Another concern on the flipside is the participant list. If it's not ephemeral, the list could potentially be infinite. (I hear Urbit Community already has this problem?)
- Another issue is just general moderation, just want to reduce overhead for Comms

pkg/arvo/app/settings-store.hoon
liam-fitzgerald: JSON false

pkg/interface/src/views/apps/launch/app.js
```diff
@@ -51,12 +65,60 @@ export default function LaunchApp(props) {
...
+          <StatelessAsyncButton primary onClick={onContinue}>
+            Yes
+          </StatelessAsyncButton>
+        </Row>
+      </Col>
+
+    )}
+  });
+  useEffect(() => {
+    const seenTutorial = _.get(props.settings, ['tutorial', 'seen'], true);
```
matildepark, line 109: I don't think lodash is imported here.

matildepark (approved):

PR #4391, opened 2021-02-08
ViewProfile: fix incorrect loading conditional

liam-fitzgerald:
Also cleans up error handling, to account for the case that someone's pinned a group that are banned from / can't join.


matildepark:
Try this out:

1. Load a profile with a pinned group (eg. ~haddef-sigwen)
2. Leap to a profile with a different pinned group (eg. ~tonlur-hidlyn)

It'll show profile 1's pinned group on profile 2. Then it'll suddenly time out and show a spinner forever. Any way to trigger a rerender here?

liam-fitzgerald:
@tacryt-socryp this OTA logic look ok?

matildepark (approved):
Front-end changes look fine. The OTA logic should be looked at if @tacryt-socryp comes around, otherwise can be merged before we deploy Ropsten.

PR #4393, opened 2021-02-08
Inbox: make useLazyScroll account for archived notifications

liam-fitzgerald:
useLazyScroll was not accounting for the case where a notification would
be archived. This would mean that archiving a lot of notifications
without scrolling would erroneously not load more. Accounts for this
case by passing a count argument to useLazyScroll, which is then
compared against its previous value. If the previous value is greater
than the new value, (i.e. a notification was archived) then check if we
should load more.


matildepark:
Targeting next userspace since it’s the next immediate release and it keeps the merge flow unidirectional. Conflicts now though.

tacryt-socryp (approved):
LGTM

PR #4395, opened 2021-02-08
interface: more visual fixes, QA

tacryt-socryp:
- https://github.com/urbit/landscape/issues/359
- https://github.com/urbit/landscape/issues/351


matildepark:
You missed half of urbit/landscape#351 — it always says "0 channels".

tacryt-socryp:
Oh, didn't even see that part. kk, will fix

matildepark:
That's not what I mean. Here's an example:

image

matildepark:
The issue is L200 in GroupsPane.tsx, we just stub it out with "0" instead of passing in the channels count.

matildepark (approved):

PR #4396, opened 2021-02-08
links-fe: correct padding

liam-fitzgerald:
Corrects the padding on comments in links. Is this the alignment you're looking for @urcades ?

Capture d’écran 2021-02-09 à 8 05 44 AM

Fixes urbit/landscape#364


urcades:
Yes, this is perfect!

—
~fabled-faster
https://urbit.org/

PR #4397, opened 2021-02-09
TextContent: handle linebreaks before feeding parsing

liam-fitzgerald:
Handling linebreaks outside of the parser allows us to preserve
linebreaks, which goes against the markdown spec and is poorly
supported by the parser library.


matildepark:
This seems to work fine except with line breaks in blockquotes, which should just collapse to one blockquote.

image

matildepark (approved):
Honestly, I'll take it, because it's a strict improvement regardless.

PR #4398, opened 2021-02-09
ted: add -code thread

liam-fitzgerald:
@narner callable by POSTing `{}` to `/spider/json/code/tape.json` (requires that you're already authenticated)

### DIFF
--- a/pkg/arvo/mar/tape.hoon
+++ b/pkg/arvo/mar/tape.hoon
@@ -1,0 +1,12 @@
+|_  tap=tape
+++  grad  %noun 
+++  grow
+  |%
+  ++  noun  tap
+  ++  json  s+(crip tap)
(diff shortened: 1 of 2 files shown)


tacryt-socryp (approved):
Is code a @p or @q? Other than that question, looks good.

liam-fitzgerald:
Check `/gen/code.hoon`, i ported the code from there. It really should be @q but I suspect +code predates the introduction of @q

Fang-:
I guess we can't just bind `/gen/code` somewhere (as a `%gen` binding) because eyre requires results from those to be actual http response payloads. Wondering if we want to try doing mark conversion on the result we get, similar to what we do for the scry endpoint...

Authentication code does indeed predate `@q`.

PR #4399, opened 2021-02-09
Profile: tweak sigil size

liam-fitzgerald:
Fixes urbit/landscape#370


matildepark (approved):
The sigil in the statusbar could be rewritten to take advantage of this but this looks fine as is

PR #4400, opened 2021-02-09
contact-hooks: prevent infinite hook loop

liam-fitzgerald:
- Prevents infinite loop in proxy a %share action
- Does not send a %share if we are the recipient


matildepark (approved):
One line assertion looks sane to me.

PR #4410, opened 2021-02-10
interface: fixed conditional to only show avatar in chat when we have…

tacryt-socryp:
… shared it

Fixes https://github.com/urbit/landscape/issues/363


matildepark (approved):
lgtm, works, though yes ChatResource is getting a little big by punting the logic up

PR #4411, opened 2021-02-10
interface: don't show link submit when empty collection and we don't have permission

tacryt-socryp:
Fixes https://github.com/urbit/landscape/issues/375


matildepark (approved):
Tested, works, needs one line of body copy for the now-blank collection appearance.

PR #4412, opened 2021-02-10
interface: redirect to base url when unsubscribing, removing, or arch…

tacryt-socryp:
…iving

Fixes https://github.com/urbit/landscape/issues/385


matildepark (changes requested):
This all drops us on an infinite spinner because it redirects to the "view graph" URL not the "join graph" URL. Pushing us to the group's landing page would be preferable here.

matildepark (approved):

matildepark:
Got a conflict here, can you resolve.

PR #4413, opened 2021-02-10
La/preview render

tacryt-socryp:
Fix https://github.com/urbit/landscape/issues/388


matildepark:
Gonna rebase this to pull in Liam's fix but retain your re-render fix.

matildepark:
Right, I don't think we have a useEffect on ViewProfile anymore because we use GroupLink not GroupSummary.

PR #4415, opened 2021-02-11
Tutorial: bugfixes

liam-fitzgerald:
Cleans up the tutorial to match designs
cc: @jyng


matildepark (approved):
There's a lil more going on here than just tutorial design fixes but it seems like "inadvertently finding and catching bugs" which is also fine. This is unglamorous work, thank you for doing it.

PR #4416, opened 2021-02-11
interface: mini profile does not dismiss improperly on click, and shows success indicator when status is set

tacryt-socryp:
Fixes https://github.com/urbit/landscape/issues/389
Fixes https://github.com/urbit/landscape/issues/390

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -294,7 +294,9 @@ export const MessageWithSigil = (props) => {
   return (
     <>
       <Box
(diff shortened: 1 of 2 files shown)


tacryt-socryp:
@matildepark that will make the whole card wiggle, won't it? that's why I put whitespace 'pre' on it

matildepark:
It makes the enter status field jump up for a second after setting, which is still preferable to that permanent padding gap. Design will want it to be some sort of checkmark iconography or field border colour, which we can get to stepwise in the next week or so.

tacryt-socryp:
ok *shrugs*

matildepark (approved):

PR #4419, opened 2021-02-11
contact-hooks: do not serve contact data at DM resource

tacryt-socryp:


pkg/arvo/app/contact-push-hook.hoon
```diff
@@ -39,6 +39,7 @@
   %+  murn  ~(tap in scry-groups:grp)
   |=  rid=res
```
liam-fitzgerald, line 40: This should probably be moved to a self-poke, so we can always scry, even if gall is molting

liam-fitzgerald (approved):
LGTM!

ixv (approved):

PR #4427, opened 2021-02-12
StatusBar: prevent sigil flexing offscreen

matildepark:
For those of us on BigYellowButton.


liam-fitzgerald (approved):

tylershuster:
Weird flex but ok

PR #4429, opened 2021-02-12
metadata-pull-hook: keep track of pending requests

liam-fitzgerald:
Backstop the overflow of requests that is occurring on livenet right now because of the slow OTA processing time.

### DIFF
--- a/pkg/arvo/app/metadata-pull-hook.hoon
+++ b/pkg/arvo/app/metadata-pull-hook.hoon
@@ -20,6 +20,17 @@
 +$  state-zero
   [%0 previews=(map resource group-preview:metadata)]
 ::
++$  state-one
+  $:  %1
+      pending=(set resource)
+      previews=(map resource group-preview:metadata)
+  ==
+::
++$  versioned-state
+  $%  state-zero
+      state-one
+  ==
+::
 --
 ::
 ::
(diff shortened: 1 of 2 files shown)


pkg/arvo/app/metadata-pull-hook.hoon
```diff
@@ -149,7 +195,11 @@
   =+  !<(=hook-update:metadata vase)
   ?.  ?=(%preview -.hook-update)
     (on-poke:def mark vase)
-  :_  this(previews (~(put by previews) group.hook-update +.hook-update))
+  =.  previews
```
tacryt-socryp, line 198: =: is more appropriate for these two modifications than =.

tacryt-socryp (approved):
Extremely straightforward, does what it says on the box. Make sure it compiles, otherwise LGTM.

pkg/arvo/app/metadata-pull-hook.hoon
liam-fitzgerald: good catch!

liam-fitzgerald:
tested on ~wicdev-wisryt, appears to resolve the issue

philipcmonk (approved):

tacryt-socryp:
Dedent 199-201 please

PR #4432, opened 2021-02-12
interface: profiles would intermittently show contact information when they should not, fixes this

tacryt-socryp:
Perhaps we can release to BYB first so Anthony can test if it fixes his case?

### DIFF
--- a/pkg/interface/src/logic/reducers/contact-update.ts
+++ b/pkg/interface/src/logic/reducers/contact-update.ts
@@ -59,18 +59,16 @@ const edit = (json: ContactUpdate, state: S) => {
     if (!field) {
       return;
     }
-    const contact = state.contacts?.[ship];
+
     const value = data['edit-field'][field];
-    if(!contact) {
-      return; 
-    }
     
     if(field === 'add-group') {
-      contact.groups.push(value);
+      state.contacts[ship].groups.push(value);
     } else if (field === 'remove-group') {
-      contact.groups = contact.groups.filter(g => g !== value);
+      state.contacts[ship].groups =
+        state.contacts[ship].groups.filter(g => g !== value);
     } else {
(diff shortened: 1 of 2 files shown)


matildepark:
Well, yeah.

pkg/interface/src/logic/reducers/contact-update.ts
```diff
@@ -59,18 +59,16 @@ const edit = (json: ContactUpdate, state: S) => {
...
     if(field === 'add-group') {
-      contact.groups.push(value);
+      state.contacts[ship].groups.push(value);
     } else if (field === 'remove-group') {
-      contact.groups = contact.groups.filter(g => g !== value);
+      state.contacts[ship].groups =
+        state.contacts[ship].groups.filter(g => g !== value);
     } else {
-      contact[field] = value;
+      state.contacts[ship][field] = value;
```
matildepark, line 71: Why would stating it directly instead of using the assignment remedy this?

tacryt-socryp: This is a simplification, not a part of the bug fix. I did remove the unnecessary !contact check from the logic, though. The meat of this fix is in the "hasLoadedAllowed" state hook. Happy to chat through it if you want to call.

matildepark: I'm just gonna try AA's specified repro with this branch

matildepark:
Good news: This does fix that specific repro.

Bad news: my nickname shows ... never. It's definitely set. If I share profile, my avatar updates, but nickname does not. It also does not if I refresh and come back to the chat. It never ever displays.

matildepark (approved):
As per edited comment ... appears to work. Just a slight lag on user info perpetuating.

PR #4438, opened 2021-02-13
contact-store: enforce timestamp ordering on %initial

liam-fitzgerald:
Will stop flickering due to an excess of `%kick`s but does nothing for the network load causing the kicks in the first place.


tacryt-socryp (approved):

tacryt-socryp:
Note: should only send initial diff if we actually modify the rolodex.

liam-fitzgerald:
we aren't sending %initial on ames other than on the first watch, so does it really matter?

tacryt-socryp:
Not too much, no, but semantically it is preferable. Less noisy.

liam-fitzgerald:
done!

tacryt-socryp:
Perfect, good to go

PR #4441, opened 2021-02-14
chat: ChatMessage spacing refactor

jamesacklin:
Provides consistent padding, margin, and other spacing rules between chat messages in the Chat window. 

Following the 1.6 spec from here: https://www.figma.com/file/ovD1mlsYDa0agyYTdvCmGr/Landscape?node-id=1973%3A0

Because each message is itself an item in a list, and not bound by the same box model as implemented in the UI reference diagram with the Auto Layout feature, there are some interesting sleights of hand at work to get everything _visually_ aligned.

spacing

I've attached a marked-up diagram that will helpfully explain the proposed changes a bit more visually. Generally speaking, this tightens the message view up considerably, but I'll leave the finer details to @urcades and @jyng 

Fixes urbit/landscape#357
Fixes urbit/landscape#355


urcades:
From the visual comparisons offered, this is looking incredibly good — I'll find time to run it so I can do the due diligence of seeing how these changes look live/applied to irl situations

jamesacklin:
Sounds good, @urcades. I'm prepared for a fair bit of iteration on this, so I appreciate your willingness to See It To Believe

matildepark:
♫ wanna squash this into one commit?

jamesacklin:
Sure, no problem. (θ‿θ)

matildepark:
I think links have bonus margins? Is this on purpose?

image

Screen Shot 2021-02-16 at 2 28 08 PM

matildepark (commented):
That's actually it. It's just the bonus margin on links. Everything else looks great!

matildepark:
Merge conflict now, but it just looks like you linted the conflicted file in question.

urcades:
status update: I have the branch synced and will be forming some responses soon!

urcades:
I intentionally parsed this work optically, rather than actually inspect its elements via the browser, I noticed a lot of very nice subtle improvements, but a nice pile of spacings we could continue to improve/tweak.

In all of the following images: 

1. **LIVENET** chat spacing is on the **LEFT**
2. **PR/NEW** chat spacing is on the **RIGHT**

—

Screen Shot 2021-02-17 at 2 15 18 AM

Overall, the "block" i've highlighted here is looking really good compared to its livenet state, I think it simply needs one less unit of spacing on its bottom (in red), to 'match' the referenced spacing indicated (in green) between two separate text posts by two different authors.

Screen Shot 2021-02-17 at 2 21 59 AM

This change is looking really, really good. The smaller line height of the 'paragraph/multi-line' text does a great job of emphasizing its 'togetherness' compared to single-line text.

Screen Shot 2021-02-17 at 2 25 34 AM

In the new chat spacing scheme, we need a lot more padding around these time/day demarcators — In this particular case's example, the time 'split' appears nested *within* a single author's multiple lines of input, the over-time-element being a multiline text row, and the under-time-element being an embed.

Screen Shot 2021-02-17 at 2 29 30 AM

Some of this spacing feels good (indicated in green), but the padding beneath embeds appears to suffer from the same aforementioned 'one unit of spacing' that the area below images seems to be affected by. Image + embed over-spacing is highlighted in red.

Screen Shot 2021-02-17 at 11 51 10 AM

Huge improvements here on the right side w/ the PR change — we need to retain the work you've done here very specifically.

Last but not least:

Screen Shot 2021-02-17 at 11 54 26 AM

It was a good intuition here to 'fix' the spacing of the left-hand margin such that it aligns nicely with the message input field (and the header channel title, which isn't shown), but I think we need to do some rethinking here of the element sizing, including the aforementioned bump of pilot avatars/sigils in chat to 24px. This would likely require us to actually keep the old padding, which should look a little bit better when the input is fixed in parallel as so:

Screen Shot 2021-02-17 at 11 58 38 AM
   
—

Echoing Matilde's comment that links and embeds seem to have a touch more padding than they need to. A lot of this is looking quite good though.

jamesacklin:
Awesome, all of this is extremely helpful @urcades. I'll address these comments and leave the 24px sigil consideration open for later action.

matildepark:
@jamesacklin you'll want to pull next-js back in post-timestamp merge

matildepark:
Ping me for review? Right now it errors:

`ERROR in ./src/views/apps/chat/components/ChatWindow.tsx
Module build failed (from ./node_modules/babel-loader/lib/index.js):
SyntaxError: /Users/matilde/git/tlon/urbit/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx: Identifier 'Patp' has already been declared (10:9)`

jamesacklin:
Breaking out urbit/landscape#460

matildepark:
image

Too much top margin on that link! Also this changes the timestamps to use 12 hour time which matches daybreak so seems fine? Would invite @urcades to check out the branch again

matildepark (approved):

PR #4443, opened 2021-02-15
landscape: button label nowrap

jamesacklin:
Sets the text label of the "Join a Group" and "Create Group" buttons on the home screen to never wrap to two lines.

image

Fixes urbit/landscape#435


pkg/interface/src/views/apps/launch/components/ModalButton.tsx
```diff
@@ -1,38 +1,33 @@
...
         borderRadius={2}
-        boxShadow="0 0 0px 1px inset"
-        color="scales.black20"
+        boxShadow='0 0 0px 1px inset'
+        color='scales.black20'
         {...rest}
       >
-        <Icon icon={props.icon} mr={2} color={color}></Icon><Text color={color}>{props.text}</Text>
+        <Icon icon={props.icon} mr={2} color={color}></Icon>
+        <Text color={color} className='nowrap'>
```
matildepark, line 25: Basically never use indigo-static classes because they're slowly phasing out of code, you can do `flexWrap="nowrap"` here

matildepark (changes requested):

matildepark (approved):

PR #4451, opened 2021-02-17
interface: upgraded SourceCodePro

tylershuster:
Fixes https://github.com/urbit/landscape/issues/404

Also there was a typo in the bold version


matildepark (approved):

matildepark:
Wait — should these be empty?

image

tylershuster:
That looks like a github bug...if you view the file like https://github.com/urbit/urbit/blob/24f0ae5b792a7536ac3004ac1e31c37e70b45083/pkg/arvo/app/landscape/fonts/sourcecodepro-bold.woff2 you can download it

PR #4452, opened 2021-02-17
links: adds titles to remote content

tylershuster:
Fixes https://github.com/urbit/landscape/issues/260


pkg/interface/src/views/apps/links/components/LinkItem.tsx
```diff
@@ -119,8 +119,10 @@ export const LinkItem = (props: LinkItemProps) => {
         overflow="hidden"
         onClick={markRead}
       >
+        <Text p={2}>{contents[0].text}</Text>
         <RemoteContent
-          ref={r => { remoteRef.current = r}}
+          ref={r => { remoteRef.current = r }}
+          renderUrl={false}
```
matildepark, line 125: So we’re not overwriting the submitted title, just hiding it and showing the embedded one?

tylershuster: No - the `RemoteContent` component was designed to intelligently display either text or the content (or both when manually set), which is fine when it's not also in the context of the links app. Like, there is no explicit text in chat. So this PR says "forget about the tricks you use to render text" to `RemoteContent` and explicitly inserts a `Text` in `LinkItem`

matildepark (approved):

Issue #2851, opened 2020-05-05
gall: failed mark translation causes %done to be given to an agent


liam-fitzgerald:
**Describe the bug**
If ford fails to translate a mark (in `+mo-handle-sys-rep`), then an internal `%done` gift is given, which calls `ap-generic-take` by way of `mo-handle-use` and `take`. This causes the `%done` gift to be given to the agent in +on-arvo.

**To Reproduce**
Steps to reproduce the behaviour:
1. Boot two fake ships
2. create a chat between the two fake ships
3. Change a mark sent over ames to be incorrect on subscriber-side. e.g. change `noun:grab` in `mar/chat/update.hoon` to `@` for the ship that isn't the chat host
4.  Cause the bad mark to be built via an ames messsage. e.g. try and join a chat. 

**Expected behaviour**
`%done` should be given to ames, so it can acknowledge that the message handling failed. 

cc: @philipcmonk

belisarius222:
What vane does the %done come from?  I don't think Ford knows about %done, which suggests to me that this is Gall giving back to itself.

—
~rovnys-ricfer
https://urbit.org

liam-fitzgerald:
Yeah `%done` comes from Gall when it handles an incomplete or errorred build from ford inside `mo-handle-sys-rep`. Apologies looks like the description is unintentionally ambiguous

matildepark:
@ixv context on why I poked you about this, from our Watercooler chat over weekend...

@belisarius222:

```
hm, this 2nd watch ack is getting bad on my shiop
*ship
https://pastebin.com/wZVHgu8k
something like one per second
brief |verb output:
https://pastebin.com/t3vK7zRG
|ames-verb is more illuminating:
https://pastebin.com/vX8pVUBC
some +verb output for :group-pull-hook:
https://pastebin.com/nEsqstDV
here's some output with both app debug and ames debug traces on:
https://pastebin.com/ZLzfzQ69
tried that again, but added %msg printing to ames, which gives a lot more info
https://pastebin.com/qL3BGCic
looks like error-handling is implicated here
for some reason, :group-push-hook is crashing repeatedly on a %boon
this is probably a bug, although it shouldn't be able to put ames into this loop
which means there's also an error-handling bug in ames or gall
actually it looks like there's another possibility: if ames itself is crashing on the boon, it delivers a %lost notice to the vane
yeah, if gall gets a %lost, it delivers a %kick to the app
[...]

PR #4454, opened 2021-02-17
gall: fixes to %boon handling

belisarius222:
When receiving a `%lost` from Ames, Gall was sending a `%u` message to leave the subscription.  However, Gall wasn't calling the normal routine for this, so it wasn't adding the leave request to its request queue -- this was causing the request queue to become desynchronized, which at least partially contributed to the repeating "2nd watch ack" printfs.

This fix does not recover from a broken queue (at least not on my ship), but it should prevent a working queue from desynchronizing.  I'm not yet sure how we should try to recover from this condition, but it might be good to get this onto the network ASAP to prevent more ships from entering the error condition.

EDIT:
The first commit only fixed one of a few problems going on simultaneously.  https://github.com/urbit/urbit/pull/4454/commits/d317a0847b88e209b5c0e29dfb791aabf33b3889 fixed another one: Gall had been trying to %give a %done to Ames if it was unable to process a %boon from Ames.  This was the wrong thing to do, because Ames clients should not ack %boon's, and because Gall's attempt at acking actually delivered the %done back to itself, causing a later crash in wire parsing.  Now Gall crashes deliberately, Ames prints the stack trace, and then Gall (mostly) correctly enqueues a %leave to Ames in response.

This PR does not fix a serious underlying issue in Gall's handling of %boon's and %done's from Ames.  Gall does not properly handle acks to outgoing sequences of repeating %leave and %watch moves over Ames, leading to the "2nd-watch-ack" error message and potentially to more serious inconsistencies.

The solution to this problem is out of scope for this PR, but it will probably involve having Gall assign a nonce to each %watch message, which it will send to the publisher Gall, who will include the nonce in all %fact and %kick messages while that instance of the subscription is active.  When the client resubscribes, it will send a %watch with a new nonce.  This will allow the subscriber to ignore responses (%done's, %fact's, and %kick's) on an already-defunct nonce and prevent queue desynchronization.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1909,6 +1909,9 @@
       =/  =bone  bone.shut-packet
       ::
       ?:  ?=(%& -.meat.shut-packet)
+        =+  ?~  dud  ~
+            %.  ~
+            (slog leaf+"ames: crashed on fragment" >mote.u.dud< tang.u.dud)
         (run-message-sink bone %hear lane shut-packet ?=(~ dud))
       ::  Just try again on error, printing trace
       ::
@@ -1917,7 +1920,7 @@
       ::
       =+  ?~  dud  ~
           %.  ~
-          (slog leaf+"ames: crashed on message ack" >mote.u.dud< tang.u.dud)
+          (slog leaf+"ames: crashed on ack" >mote.u.dud< tang.u.dud)
       (run-message-pump bone %hear [message-num +.meat]:shut-packet)
     ::  +on-memo: handle request to send message
     ::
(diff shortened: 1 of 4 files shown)


belisarius222:
cc @philipcmonk

joemfb (approved):
LGTM.

FWIW, I'm not sure whether it should be released before the rest of the "2nd watch ack" phenomenon is understood.

belisarius222:
I agree.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
Great catch!  When we receive a message with an empty `outstanding` queue, we go into the `%missing` branch here: https://github.com/urbit/urbit/blob/081c4579b268b8844feb6428a0ce28ccb8e0f516/pkg/arvo/sys/vane/gall.hoon#L663-L669

Basically, if we don't know what type of request it was, we acknowledge it both as a watch and a poke.  That could definitely cause "2nd watch ack".  You would see this in |verb because they'd both show up.

belisarius222:
ready for re-review @joemfb

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1909,6 +1909,9 @@
       =/  =bone  bone.shut-packet
       ::
       ?:  ?=(%& -.meat.shut-packet)
+        =+  ?~  dud  ~
+            %.  ~
+            (slog leaf+"ames: crashed on fragment" >mote.u.dud< tang.u.dud)
```
joemfb, line 1914: I though we were going to print one line by default, and put the whole trace behind I flag -- do you think this happens rarely enough to print the full trace unconditionally?

joemfb (commented):
The %gall changes all LGTM.

One question about %ames' output:

joemfb (approved):
LGTM

philipcmonk:
Additional context on the `%done` misgiving and 2nd-watch-ack is here: #2851

Issue #3598, opened 2020-09-27
test: build failures in -test not identified in CI


joemfb:
This came up in #3576: I added a core to the stdlib, mistakenly pushed a pill without said core, then added tests exercising the new functionality. The tests fail to build, but this fact is not observed by CI. Presumably this problem dates to the addition of `-test` with ford-fusion.

joemfb:
This applies to the `:test` app too, not just the `-test` thread (which is more egregious, since checking for build failures is the only purpose of the `:test` app).

`release/next-sys` currently has a build failure in the `%graph-update` mark (see https://travis-ci.org/github/urbit/urbit/jobs/732403775#L2422).

/cc @tacryt-socryp

joemfb:
@belisarius222, bumping this. We've had repeated issues with bad merges due to spuriously successful CI results.

PR #4455, opened 2021-02-17
Vector Library (Hoon)

sigilante:
This PR reflects a subset of the changes in #3536, which has become rather unwieldy.  This is the Hoon-only vector-only code for `@lvs` and `@lvd`.

Right now, I believe that if there are any bugs latent, they are minor.  A test suite is included.

@joemfb proposed that the library pin a single bit at the head of the `@lv` atom and use `++met` to get the atom length.  This is acceptable (but deviates from expected `@lm` behavior) but has not been implemented as of initial creation of this PR.


ohAitch:
I have also proposed this, except storage-wise it doesn't really matter if
the MSB is a bit or a byte, so may as well store a type tag that it's a
vector and of single-precision and so on ;)

(And for a matrix, you still need the row width in the lsb-trailer, but you
no longer need the number of rows)

sigilante:
Yeah, we're all agreed it's a metadata bit/byte.

- `1` single bit as head tag, use `++met` in `++length`; advantage:  no 2^31 length limit
- vector length as head tag, no real issue with this but runs risk of redundancy and being wrong
- vector identifier (think like NaN-hacking)

I think, looking at the Hoon, that the `1` pinned bit has the minor advantage that almost nothing need change to implement it since `++length` does all the work for the other operators.

ohAitch:
Oh yeah I definitely think you shouldn't have the *total length* in there, just the dimensionality(vector vs matrix vs cube vs higher-order) and maybe bit width

sigilante:
Head pin of `1` is dimensionality for the vector library, it's true.

ohAitch:
:P yes, if we're keeping complex vs real and double vs single vs etc precision purely in the type system, they are equivalent, and my proposal boils down to "use a consistent similar tagging system for higher-rank arrays"

sigilante:
The change was trivial, only in `++zeroes` and `++length`.

Since nothing is done with that pinned value (or even multiple values at the head), setting up a more sophisticated type tag later is straightforward.

sigilante:
@joemfb accounting for jet matching is it likely that a jetted C version of the get/set routines would actually be faster?  (Currently it's done with bit masks etc. in Hoon.)

sigilante:
I added the `++cumsum` function and a `++subvector` tool as well.

joemfb:
The jet-matching overhead is mostly just call (ie, nock `%9`) overhead. `+get:lvs` would see basically no improvement; `+get:lvd` would see minor improvement (should those be so different?); both `+set` arms could see significant improvement, as most if not all intermediate allocations could be elided.

BTW, the CI success is spurious here (due to #3598), there's a build failure in the tests:`

```
syntax error at [36 23] in /tests/lib/lv/hoon
      !>  `@lvs`0x100000000
----------------------^
[%error-building /tests/lib/lv/hoon]
```

sigilante:
This should synchronize @lvs and @lvd entirely, no equivalent forms.

pkg/arvo/tests/lib/lv.hoon
```diff
@@ -0,0 +1,508 @@
...
+::  Tests for utility functions
+::
+++  test-length  ^-  tang
+  ;:  weld
+    %+  expect-eq
+      !>  0
+      !>  (length:lvs 0x0)
+    %+  expect-eq
+      !>  1
+      !>  (length:lvs 0x13ff0000)
```
joemfb, line 222: s/b `0x13ff.0000`

pkg/arvo/lib/lv.hoon
```diff
@@ -0,0 +1,433 @@
...
+  ::
+  ::    Scalar division
+  ++  divs
+    |=  [u=@lvs s=@rs]  ^-  @lvs
+    ::~/  %divs
+    =/  ss  (fill (length u) s)
+    (divv u ss)
+  ::
+  ::    Turn on a gate of two variables.
+  ++  turn2  :: I guess not in hoon.hoon
```
tacryt-socryp, line 157: I've seen this method referred to as "zip" in ReactiveX.

sigilante:
There are five issues to resolve:

- [ ] `++expect-eq` isn't adequate for floating-point maths so I'll need to write an `++expect-close`
- [ ] The empty vector is `0x1` not `0x0` because it has a dimensionality of one.  I can change this but it means special handling.  I'm inclined to leave it alone for now.
- [ ] `++mul:rs` has the wrong default sample for a product with `++roll`.  Do you remember how to supply a different default sample to the gate?
- [ ] The `@lvd` tests should be straightforward but aren't written pending resolving these three issues.
- [ ] Currently the error raised is tagged with `~_  leaf+"lagoon-fail"`, but we can decide if we want to stick with that error type or something different.

sigilante:
The current test status:

```
OK      /lib/lv/test-zeros
OK      /lib/lv/test-unmake
OK      /lib/lv/test-sum
OK      /lib/lv/test-subv
OK      /lib/lv/test-subs
OK      /lib/lv/test-set
FAILED  /lib/lv/test-product
expected: .1.2e2
actual: .0
OK      /lib/lv/test-ones
OK      /lib/lv/test-mulv
OK      /lib/lv/test-muls
OK      /lib/lv/test-max
OK      /lib/lv/test-make-unmake
OK      /lib/lv/test-make-nat
OK      /lib/lv/test-make
CRASHED /lib/lv/test-length
/ted/test/hoon:<[69 12].[69 37]>
/ted/test/hoon:<[69 20].[69 36]>
/tests/lib/lv/hoon:<[203 18].[214 5]>
/tests/lib/lv/hoon:<[204 3].[214 5]>
/tests/lib/lv/hoon:<[205 5].[207 33]>
/tests/lib/lv/hoon:<[207 7].[207 33]>
/tests/lib/lv/hoon:<[207 11].[207 33]>
/lib/lv/hoon:<[39 17].[41 20]>
/lib/lv/hoon:<[40 5].[41 20]>
lagoon-fail
/lib/lv/hoon:<[41 5].[41 20]>
decrement-underflow
FAILED  /lib/lv/test-isclose
failed to nest: -actual.%.y: -expected.?(%.y %.n)
failed to nest: -actual.%.n: -expected.?(%.y %.n)
OK      /lib/lv/test-iota
OK      /lib/lv/test-inner
OK      /lib/lv/test-get
OK      /lib/lv/test-fill
FAILED  /lib/lv/test-divv
expected: 0x13e4ccccd3e8000003eaaaaaa3f0000003f800000
actual: 0x13e4ccccc3e8000003eaaaaaa3f0000003f800000
FAILED  /lib/lv/test-divs
expected: 0x1412000004100000040c000004080000040000000
actual: 0x140200000400000003fc000003f8000003f000000
expected: 0x13f0000003ecccccd3e99999a3e4ccccd3dcccccd
actual: 0x14247ffff421fffff41efffff419fffff411fffff
OK      /lib/lv/test-catenate
OK      /lib/lv/test-append
OK      /lib/lv/test-addv
OK      /lib/lv/test-adds
```

- `++test-product` is because of the `++mul:rs` default.
- `++test-length` unknown, I think it's because I changed the atom structure, will investigate
- `++test-isclose` has something to do with what I expect `++expect-eq` to do with loobeans
- `++test-divv` is because FP math is path-dependent
- `++test-divs` looks like a math error on my part, will fix

sigilante:
Currently getting a build error around something that works locally...

`++all-close` checks all members of two vectors for `++isclose`, but the testing suite blocks the build because of an atom mismatch.

pkg/arvo/lib/lv.hoon
```diff
@@ -0,0 +1,451 @@
...
+    |=  s=@rd
+    (isclose s .~0 .~1e-6)
+  ++  all-close
+    |=  [u=@lvd v=@lvd]
+      =/  n  `@ud`(length u)
+      =/  count  1
+      =/  off  0
+      |-
+        ?:  (gth count n)  =(0 off)
+      $(count +(count), sum (add off `@`(isclose (get u +(count)) (get v +(count)) .~1e-6)))
```
joemfb, line 311: The test failure is pointing to this line: you're replacing `sum` and recursing, but that's not a local face (and is presumably a core in the stdlib). Is it supposed to be `off (add off ...)`?

sigilante:
I wrote an `++expect-near` to handle closeness of floats.

```hoon
++  expect-near
  |=  [expected=vase actual=vase]
  ^-  tang
  ::
  =|  result=tang
  ::
  =?  result  !(all-close:lvs `@lvs`q.expected `@lvs`q.actual .1e-6)
    %+  weld  result
    ^-  tang
    :~  [%palm [": " ~ ~ ~] [leaf+"expected" (sell expected) ~]]
        [%palm [": " ~ ~ ~] [leaf+"actual" (sell actual) ~]]
    ==
  ::
  =?  result  !(~(nest ut p.actual) | p.expected)
    %+  weld  result
    ^-  tang
    :~  :+  %palm  [": " ~ ~ ~]
        :~  [%leaf "failed to nest"]
            (~(dunk ut p.actual) %actual)
            (~(dunk ut p.expected) %expected)
    ==  ==
  result
```

The trouble is that `q.expected` etc. from the vases are `*` not `@` so I can't cast them properly.  Can you think of a workaround?

joemfb:
I think you should just write it for the aura's you expect:

```
++  expect-near
  |=  [expected=@lvs actual=@lvs]
  ^-  tang
  ?:  (all-close:lvs `@lvs`q.expected `@lvs`q.actual .1e-6)
    ~
  :~  [%palm [": " ~ ~ ~] [leaf+"expected" leaf+(pprint:lvs expected) ~]]
      [%palm [": " ~ ~ ~] [leaf+"actual" leaf+(pprint:lvs actual) ~]]
  ==
```

`+expect-eq` takes vases primarily so it can print arbitrary data. The nest-check is debatable. FWIW, I find the vase-based test framework frustrating to use, and would like to play around with more language-integrated testing approaches.

sigilante:
Shunting forwards discussion:

- [ ] Resolve loobean match (previous comment).
- [ ] Write the `@lvd` tests, which are straightforward, pending loobean test fix.

On that basis, this Hoon is ready to go.  I'm going to put the jets as a separate PR for developer hygiene.

tagging @joemfb

pkg/arvo/lib/lv.hoon
```diff
@@ -0,0 +1,453 @@
...
+  ::  |x-y| <= tol
+  ++  isclose
+    |=  [s=@rs t=@rs tol=@rs]
+    (lth:rs (abs (sub:rs s t)) tol)
+  ++  near0
+    |=  s=@rs
+    (isclose s .0 .1e-6)
+  ++  all-close
+    |=  [u=@lvs v=@lvs tol=@rs]
+      =/  n  `@ud`(length u)
```
joemfb, line 85: This block should be dedented to the same level as the `|=`

pkg/arvo/lib/lv.hoon
```diff
@@ -0,0 +1,453 @@
...
+      =/  count  1
+      =/  off  0
+      |-
```
joemfb, line 88: It's idiomatic to cast loops:

```
|-  ^-  ?
```

pkg/arvo/lib/lv.hoon
```diff
@@ -0,0 +1,453 @@
...
+        ?:  (gth off 0)  %.n
+        ?:  (gth count n)  %.y
```
joemfb, line 90: `%.y` and `%.n` are constant atoms -- their type is their value. `?` is `?(%.y %.n)`. The type of `&` or `|` is `?` (you can also use `.y` and `.n`, but idk if I've ever seen those). This is an awkward corner of the language, and also the problem you're hitting in your tests.

pkg/arvo/tests/lib/lv.hoon
```diff
@@ -0,0 +1,509 @@
...
+  ;:  weld
+    %+  expect-eq
+      !>  `@lvs`0x1.40a0.0000.4080.0000.4040.0000.4000.0000.3f80.0000
+      !>  (iota:lvs 5)
+  ==
+++  test-isclose  ^-  tang
+  ;:  weld
+    %+  expect-eq
+      !>  (isclose:lvs .1 .1.00001 .0.0001)
+      !>  %.y
```
joemfb, line 239: As noted above, you're running into a problem here because the type of `%.y` is `%.y`, while the product of `+isclose` is `?`, which is `?(%.y %.n)`. You're hitting the nest-check in `+expect-eq`, which I earlier characterized as questionable.

You could fix this by changing using `&` instead of `%.y`. But your test sample is in the wrong order, `(expect-eq expected actual)` is the design here. If you reverse the arguments, the nest check will succeed.

There's also a bare `+expect` arm that would probably be best here, and will also avoid the argument order (and nest direction) issue.

joemfb (changes requested):

sigilante:
With that resolved, I need to bring in the `@lvd` tests, then this PR is "done" or at least ready for the next phase.

sigilante:
Done with `@lvd`, and that's a Highday night.

HMU when you're ready to take this to the next step, figuring out where in `%zuse` or `%lull` to house it or whatever.  I'll be working on the jet PR next.

ohAitch:
My vote is, split floats across zuse and hoon.hoon leaving the minimum you
need for parsing/printing, lagoon should have a hoon language syntax and
similarly have code to pack/unpack it in hoon.hoon with all math in zuse.
Jets work properly in zuse nowadays right?

Throwing in round-to-even printing while we're at it
https://youtu.be/kw-U6smcLzk

sigilante:
@joemfb I'm going to pull the jets together.  They're pretty much all written---actually, I'm going to cut some out---but should I work on a %zuse-compatible refactor of this so the jets only have to be set up once or just go ahead with the jets as-is?

joemfb:
As-is should be fine -- they don't need to be in %zuse unless we want to require that they be present.

joemfb:
(Or if we change to out-of-band jet registrations, but it's still too early to plan around that)

sigilante:
@joemfb this one is still flagging a requested change but I can't figure out which it would be. (I think I outdated it but not sure.)

tacryt-socryp:
Ping.

PR #4456, opened 2021-02-18
settings-store: add existence scry

ixv:
cc @mrdomino I know the resolution we came to for your use case was to just use subscriptions, but I figured this is a simple enough change and it will probably be helpful to have anyway.


mrdomino:
Wrong merge base?

mrdomino:
Thanks!

PR #4459, opened 2021-02-18
interface: ensure that we display the correct ship name and do not retain stale state

tacryt-socryp:
Fixes https://github.com/urbit/landscape/issues/434

### DIFF
--- a/pkg/interface/src/views/apps/chat/ChatResource.tsx
+++ b/pkg/interface/src/views/apps/chat/ChatResource.tsx
@@ -114,7 +114,6 @@ export function ChatResource(props: ChatResourceProps) {
         } else {
           setShowBanner(false);
         }
-
       } else {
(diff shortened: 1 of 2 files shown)


matildepark (changes requested):
Still at large!

Try this repro:

1. Go to a direct message.
2. Refresh the page.

Your nickname will never appear. Also note that ChatInput also needs this fix too.

tacryt-socryp:
Huh, I couldn't reproduce that on the DMs page in my testing. Will try again.

tacryt-socryp:
Are you sure that this doesn't work? I've created a test case in which it takes an additional 5 seconds to set the "hasLoadedAllowed" boolean to true, and the contact information comes in properly.

matildepark:
Will film and show. Standby

matildepark (approved):
This is good to go if you remove the console.logs!

PR #4460, opened 2021-02-18
chat: only show avatars if we have them

matildepark:
Also respects CalmEngine, as (unfiled) we weren't checking `hideAvatars` before rendering inside chat messages.

Fixes urbit/landscape#315


Fang-:
> Also respects CalmEngine

Yess, good fix. Note that this issue also exists in the "messages" view sidebar, where it currently shows avatars (and nicknames) despite CalmEngine settings. Can file separately, if you think that's out of scope here.

matildepark:
Thought I caught that there ... I can add that here, it's no problem.

jamesacklin (approved):
LGTM

PR #4462, opened 2021-02-19
metadata: do not generate previews for hidden groups

liam-fitzgerald:
Fixes an issues where groups that were originally hidden but got groupified will be unable to be joined. We no longer render previews for hidden groups anyway, so just discard. 
cc: @yosoyubik


tacryt-socryp (approved):
LGTM, simple. Would we have any way to clear previews for already fetched hidden groups?

liam-fitzgerald:
Unlikely to be an issue, as most of these previews will disappear when the invite is accepted

PR #4463, opened 2021-02-19
term: next gen, phase one

### LINKED ISSUE #353: prompt doesn't always update
upon the pier first starting, after ~&s and idle ^C, etc.

### LINKED ISSUE #2646: unused Dill tasks
There are four `task`s in Dill that appear to be unused.

`$flow` and `$harm` appear in `+task:able:dill` in `zuse.hoon`, but their entry in `+call` in `dill.hoon` is just a lark expression that does nothing.

`$hook` and `$talk` also appear in `+task:able:dill` in `zuse.hoon` but have no entry in `+call` in `dill.hoon`.

Talking to @philipcmonk, these might be a mixture of `task`s that once did something but have been deprecated, or unfinished `task`s that someone intends to get to eventually, so I don't think they necessarily all need to be scrubbed.

Fang-:
Bunch of small goodies in here, but the majority of this revolves around decoupling term.c, dill and drum.

Previously, they all knew about and played a role in drum's "prompt at the bottom" semantics.  
With these changes, only drum knows about the prompt. term.c just remembers whatever the line at the bottom of the screen contains to aid in spinner drawing. It also whimsically inserts printfs between the second-to-last and bottom lines of the screen, but those only show up in the default session, which is always linked to drum, so it works out fine.

Other fixes and improvements include:
- Support for 2D cursor movement, and by extension 2D screen printing,
- Better (but still incomplete) support for multiple sessions in dill,
- Minimal X10-style mouse-click support,
- A cleaned up dill api,
- General cleanup and smaller improvements.

There's some remaining work here, but this is probably good to go for a first review while I work on that. While the _order_ of these commits is a bit all over the place, the commits _themselves_ should be well-contained changes, often paired with additional commentary, and as such probably the best way to review all of this.  
Please also keep an eye out for `//REVIEW` comments strewn about, I'd love some input on those.

Some remaining work:
- [x] Fix/improve `-t` mode behavior. See also CI output, it's kinda not good in some places.
- [x] ~~Update webterm and its dependencies to reflect these changes. (The blit mark doesn't even build rn, _but CI passes anyway!_ [distant screaming])~~ Updated, but not yet perfect. Remaining work includes:
  - [x] Restructuring and cleanup of codebase, follow best practices.
  - [x] Improved subscription management. (Pending broader landscape subscription reform.)
  - [x] Respond to window resizing.
  - [x] Respond to landscape theme (light vs dark).
- [x] Update KH to reflect these changes.
- [x] Make clay and gall send proper printout requests to dill, instead of having dill handle their updates locally.
- [x] ~~Experiment with multi-session dill in webterm, settle on an interface. May or may not get~~ Settled on an (as of yet unreviewed) interface. Further drum & webterm work punted to a separate PR. Fixes #2646.

Currently targeting ~`release/next-sys`~ `release/next-dill`, which is the merge of `next-sys` and `next-js`, so we have a cleaner diff here.  
The arvo-side changes depend on the runtime-side changes. So maybe we want `release/next-step` or w/e for that? Also depends on urbit/landscape#188 if we want users to actually know about the strict need for a runtime upgrade.

Might fix #353, if that wasn't already the case.


Fang-:
Noting for my own records that there's currently a bug in this, where sometimes when printing large amounts of output (ie, printing many lines), the last handful of lines have double newlines between them, where there should be none. Suspecting that this relates to the spinner logic, because I hadn't seen this prior to those changes, but that's really just a hunch...

joemfb (changes requested):
Sorry for the delay! I've noted a couple minor issues:

Fang-:
Alright, nu-webterm is in. xterm.js is pretty nice to use, we get to basically just port the simple parts of `term.c` and call it a day. Still needs some work though. It doesn't handle window resizing at all nicely, and I feel like there's general structural improvements to be made.

@liam-fitzgerald please take a look. I might need your help with cleaning up/restructuring the single-file mess I've made, maybe we can pair a little bit if you have time.

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -548,6 +549,19 @@
     leaf+(weld (scag (sub edg 3) tape) "...")
   leaf+tape
 ::
+++  se-blin                                           ::  print and newline
+  |=  lin=dill-blit:dill
```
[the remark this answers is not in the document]
Fang-: Good, and done, but noting that this turns out kinda ugly due to #2688.

Fang-:
Haskell changes are in, paging @pilfer-pandex for review. I haven't put the re/connecting to a running ship to the test yet, so don't be surprised if anything about that is wrong.

But also, if I understand correctly, it's storing history and such so that it can redraw the screen when you connect. But there's a dill move for that, which will go into the terminal handler and just redraw the screen for you, no further runtime logic required.  
(This doesn't account for slogs and logging output and such, and for current drum only redraws the prompt, but that will improve as more output gets reified and terminal handler technology evolves.)

ohAitch:
Any roadmap on sixel support? :)

Fang-:
@ohAitch I've been keeping an eye on xtermjs/xterm.js#2503 for laughs, since webterm will likely be the easiest place to experiment with these things, but it's not on the roadmap per se. It's certainly a nice-to-have, but there's plenty of other work in this area that's more profitable.

fwiw, I'm thinking the order of operations here is going to be something like:
1. this pr,
2. full session support in drum, webterm, eventually the runtimes,
3. drum-alternative with per-application buffers,
4. native scroll/back support & copy handling in drum-style buffers, enabling full mouse integration,
5. terminal-based experiments in the direction of "ui protocol"

5 is pretty underdefined, but sixels would happen somewhere around there. All this is of course somewhat tentative, and notably lower-priority work than many other things going on, so the best answer I can give you is "likely, eventually".

pkg/hs/urbit-king/lib/Urbit/Vere/Term/Logic.hs
```diff
@@ -83,15 +83,17 @@ step st@St{..} = \case
     EvLine t -> st & recordText t
     EvSlog s -> st & recordSlog s
     EvSpin s -> st { sSpinner = s }
-    EvMove w -> st { sCurPos = min w (word $ length sLine) }
-    EvEdit t -> st { sLine = t, sCurPos = word (length t) }
-    EvMore   -> st { sLine = "", sCurPos = 0 } & recordText (sLine <> "\n")
+    EvMove p -> st { sCurPos = p }
+    EvEdit t -> if fst sCurPos == 0 then st { sLine = t }
+                else st
```
pilfer-pandex, line 88:
```suggestion
    EvEdit t | (0, _) <- sCurPos -> st { sLine = t }
             | otherwise         -> st
```
and likewise below. This is called a "pattern guard." In general the existence of "guards" means you shouldn't ever need to put an if expression to the right hand side of a pattern match

pkg/hs/urbit-king/lib/Urbit/Vere/Term/Logic.hs
```diff
@@ -111,8 +113,8 @@ drawState :: St -> [Ev]
 drawState St{..} = hist <> out <> cur <> spin
   where
     hist = drawHistory <$> toList sHistory
-    out  = if null sLine   then [] else [EvEdit sLine]
-    cur  = if 0 == sCurPos then [] else [EvMove $ fromIntegral $ sCurPos]
+    out  = if null sLine       then [] else [EvEdit sLine]
```
pilfer-pandex, line 116: isn't this just `map EvEdit sLine`?

pkg/hs/urbit-king/lib/Urbit/Vere/Term/Logic.hs
```diff
@@ -148,12 +151,12 @@ fromTermEv = \case
...
-    EvMore    -> Term.Blits [Arvo.Mor ()]
+    EvLine ""  -> Term.Blank
+    EvLine t   -> Term.Trace (Cord t)
+    EvSlog s   -> Term.Slog s
+    EvSpin s   -> Term.Spinr (fromCause <$> s)
+    EvMove (r, c) -> Term.Blits [Arvo.Hop $ Arvo.Roc (fromIntegral r) (fromIntegral c)]
+    EvBell     -> Term.Blits [Arvo.Bel ()]
+    EvDraw     -> Term.Blits [Arvo.Clr ()]
+    EvEdit t   -> Term.Blits [Arvo.Put $ unpack t]
+    EvNewl     -> Term.Blits [Arvo.Nel ()]
```
pilfer-pandex, line 162: Please line these up (or don't)

pkg/hs/urbit-king/lib/Urbit/Vere/Term/Render.hs
```diff
@@ -25,8 +29,31 @@ clearLine = liftIO $ ANSI.clearLine
...
-cursorLeft :: MonadIO m => Int -> m ()
-cursorLeft = liftIO . ANSI.cursorBackward
+--NOTE  top-left-0-based coordinates
+cursorMove :: MonadIO m => Int -> Int -> m ()
+cursorMove r c = liftIO $ ANSI.setCursorPosition r c
 
-cursorRight :: MonadIO m => Int -> m ()
-cursorRight = liftIO . ANSI.cursorForward
+cursorSave :: MonadIO m => m ()
+cursorSave = liftIO $ ANSI.saveCursor
```
pilfer-pandex, line 37: no need for `$` here

pkg/hs/urbit-noun-core/lib/Urbit/Noun/TH.hs
```diff
@@ -101,6 +97,15 @@ deriveToNoun tyName = do
 
     pure [InstanceD overlap ctx inst [ValD (VarP 'toNoun) body []]]
 
+deriveToNounFunc :: Name -> Q Exp
+deriveToNounFunc tyName = do
```
pilfer-pandex, line 101: Awesome!

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
```diff
@@ -277,19 +278,56 @@ data LegacyBootEvent
...
+
+instance FromNoun Bolt where
+  parseNoun = \case
+    A c -> pure $ Key $ C.chr $ fromIntegral c
+    n   -> $(deriveFromNounFunc ''Bolt) n
+
+instance FromNoun Belt where
+  parseNoun = \case
+    A c       -> pure $ Bol $ Key $ C.chr $ fromIntegral c
+    n         -> runParser ($(deriveFromNounFunc ''Bolt) n) [] belt bolt
```
pilfer-pandex, line 314: man dat `($(`

it's like `(~(` in hoon

I'll see myself out

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
```diff
@@ -277,19 +278,56 @@ data LegacyBootEvent
...
+instance ToNoun Bolt where
+  toNoun = \case
+    Key c -> A $ fromIntegral $ C.ord c
+    n     -> $(deriveToNounFunc ''Bolt) n
+
+instance ToNoun Belt where
+  toNoun = \case
+    Bol b -> case b of
+               Key c -> A $ fromIntegral $ C.ord c
+               b     -> $(deriveToNounFunc ''Bolt) b
```
pilfer-pandex, line 328:
```suggestion
    Bol b -> toNoun b
```

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
```diff
@@ -277,19 +278,56 @@ data LegacyBootEvent
...
+    A c -> pure $ Key $ C.chr $ fromIntegral c
+    n   -> $(deriveFromNounFunc ''Bolt) n
+
+instance FromNoun Belt where
+  parseNoun = \case
+    A c       -> pure $ Bol $ Key $ C.chr $ fromIntegral c
+    n         -> runParser ($(deriveFromNounFunc ''Bolt) n) [] belt bolt
+                   where
+                    belt p m = $(deriveFromNounFunc ''Belt) n
+                    bolt !b  = pure $ Bol b
```
pilfer-pandex, line 317: Suggested code:

- parse the noun coming in into a Maybe Bolt
- if it's just (use `maybe`, possibly), wrap in Bol
- otherwise, call the derived Belt parser

pkg/hs/urbit-king/lib/Urbit/Vere/Term/Logic.hs
```diff
@@ -111,8 +113,8 @@ drawState :: St -> [Ev]
 drawState St{..} = hist <> out <> cur <> spin
   where
     hist = drawHistory <$> toList sHistory
-    out  = if null sLine   then [] else [EvEdit sLine]
-    cur  = if 0 == sCurPos then [] else [EvMove $ fromIntegral $ sCurPos]
+    out  = if null sLine       then [] else [EvEdit sLine]
+    cur  = if 0 == fst sCurPos then [] else [EvMove sCurPos]
```
pilfer-pandex, line 117: Some people would say that you should use a guard here instead of an if expression, but I'll leave it up to you

pilfer-pandex (commented):
Looking mostly good on the haskell side, with a few suggested changes. Unfortunately, you now know more than me about the king's terminal driver, so my ability to, e.g., comment on the ANSI stuff is limitied

might be a good idea for you to teach me when I'm a bit less busy, if you want :)

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
pilfer-pandex: Actually, you ought to be able to accomplish the above using the Alternative instance of parsers. It should look something like

```
parseNoun n = Bol <$> parseNoun n <|> $(deriveFromNounFunc ''Belt) n
```

which also very clearly and cleanly expresses what is going on

pkg/hs/urbit-king/lib/Urbit/Vere/Term/Logic.hs
Fang-: No? `sLine` isn't a list, so it doesn't let me `map` over it. Also, we want to explicitly produce `[]` for the null/empty string case, not `[EvEdit null]` or w/e.

pkg/interface/src/logic/api/term.ts
```diff
@@ -0,0 +1,21 @@
+import BaseApi from './base';
+import { StoreState } from '../store/type';
+
+export type Bolt =
+  | string
+  | { aro: 'd' | 'l' | 'r' | 'u' }
```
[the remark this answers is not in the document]
Fang-: Just an `if` check or similar on the type/properties of the value.

(Of course in practice we only care about _constructing_ this input type, we don't get it sent to us.)

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
Fang-: > which also very clearly and cleanly expresses what is going on

image

Fang-:
Alright, afaic this is good to go. Paging @pilfer-pandex, @liam-fitzgerald and @joemfb for final review on the Haskell side, webterm, and vere/dill/drum respectively.

Sessions _support_ is partially here, and will be completed in a subsequent PR. Webterm subscription management still does the old thing because the new thing isn't yet in. Any additional changes that need to be made to accommodate the release of this work will go in a separate PR also.

I've been running this on my local dev ships for the entire time I've worked on this. As far as I've been able to discern, no serious bugs remain.

ryjm:
I've been running this on a fake zod for the past couple days, also not seeing anything strange. Anything I should be on the lookout for?

Fang-:
@ryjm neat, thank you! Things I'm most concerned about are the refactored pre- and post-spinner logic, and interactions between arvo-level and runtime-level output. ie printfs overwriting parts of the screen, stuff like that.

Fang-:
I don't think any of the typescript errors here are about webterm code, fwiw.

matildepark:
> I don't think any of the typescript errors here are about webterm code, fwiw.

They aren't, there's about 127 TS errors on master and the work to clear the test (https://github.com/urbit/urbit/pull/4899) hasn't hit master. You could always pull in next-userspace to resolve that, unless this is going out sometime sooner than the next week or so.

pkg/arvo/sys/lull.hoon
```diff
@@ -1026,90 +1025,82 @@
...
+        [%view ~]                                       ::  watch session blits
     ==                                                  ::
   ::
   ::::                                                  ::  (1d2)
     ::
   +$  blew  [p=@ud q=@ud]                               ::  columns rows
-  +$  belt                                              ::  old belt
+  +$  belt                                              ::  client input
+    $?  bolt                                            ::  simple input
+        [%mod mod=?(%ctl %met %hyp) key=bolt]           ::  w/ modifier
```
philipcmonk, line 1069: Are we supposed to not be able to represent eg "control alt C"?  Feels like this should be `key=belt`

philipcmonk: Or maybe just `mod=(set ?(%ctl %met %hyp))`

Fang-: `%hyp` here represents the simultaneous pressing of `%ctl` and `%met`, not a single "hyper key".

philipcmonk: Ah, okay

pilfer-pandex (changes requested):

pkg/interface/src/views/apps/term/app.tsx
```diff
@@ -1,100 +1,458 @@
...
+  const session = useTermState(useCallback(
+    state => state.sessions[state.selected],
+    [selected, sessions]
+  ));
+
+  const osDark = useLocalState((state) => state.dark);
+  const theme = useSettingsState(s => s.display.theme);
+  const dark = theme === 'dark' || (theme === 'auto' && osDark);
+
+  const onSlog = (slog) => {
```
liam-fitzgerald, line 142: This can (and should) be defined outside the scope of this function

pkg/interface/src/views/apps/term/app.tsx
```diff
@@ -1,100 +1,458 @@
...
+             + csi('S', 1)
+             + csi('H', term.rows - 1, 1)
+             + csi('m', 90)
+             + slog
+             + csi('m', 0)
+             + csi('r')
+             + csi('u'));
+  };
+
+  const onBlit = (sesId: string, blit: Blit) => {
```
liam-fitzgerald, line 168: again, doesn't need to be defined in-component

pkg/interface/src/views/apps/term/app.tsx
```diff
@@ -1,100 +1,458 @@
...
+        }, 10000);
+      }
+    }
+
+    set(state => { state.slogstream = slog });
+  }, [sessions]);
+
+  const onInput = useCallback((ses: string, e: string) => {
+    const term = useTermState.getState().sessions[ses].term;
+    let belts: Array<Belt> = [];
```
liam-fitzgerald, line 261: `Belt[]`

pkg/interface/src/views/apps/term/app.tsx
```diff
@@ -1,100 +1,458 @@
...
+      e = e.slice(1);
+    }
+    if ('' !== strap) {
+      belts.push({ txt: strap.split('') });
+      strap = '';
+    }
+    belts.map(b => {  //NOTE  passing api.term.sendBelt makes `this` undefined!
+      api.term.sendBelt(ses, b);
+    });
+  }, [sessions, api.term]);
```
liam-fitzgerald, line 342: consider refactoring into `inputToBelts` (where the meat of the logic is and can be outside component scope) and then `onInput` would just call that and forward the results to the api.

liam-fitzgerald (changes requested):
some style and performance nits and will need a slight refactor to fit into the new api model. Else looks good!

pkg/interface/src/types/term-update.ts
```diff
@@ -0,0 +1,33 @@
...
+  | Blit;
+
+export type Tint =
+  | null
+  | 'r' | 'g' | 'b' | 'c' | 'm' | 'y' | 'k' | 'w'
+  | { r: number, g: number, b: number };
+
+export type Deco = null | 'br' | 'un' | 'bl';
+
+export type Stye = {
```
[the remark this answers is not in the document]
Fang-: I didn't know about the distinction, so I read about these a little bit. It seems `interface` is _slightly_ fancier than `type`? The former seems to carry more OOP semantics with it than the latter, certainly.

I'm fine going either way, but I'm just defining some data's shape here, so it's not obvious to my why I'd want `interface` for that.

pilfer-pandex (approved):

pkg/interface/src/views/apps/term/app.tsx
```diff
@@ -1,101 +1,452 @@
...
+  }
+  return belts;
+};
+
+export default function TermApp(props: TermAppProps) {
+  const container = useRef<HTMLDivElement>(null);
+  // TODO  allow switching of selected
+  const { sessions, selected, slogstream, set } = useTermState();
+
+  const session = useTermState(useCallback(
```
liam-fitzgerald, line 282: No need to use the `useTermState` hook you already have all these variables

pkg/interface/src/views/apps/term/app.tsx
```diff
@@ -1,101 +1,452 @@
...
+    state => state.sessions[state.selected],
+    [selected, sessions]
+  ));
+
+  const osDark = useLocalState(state => state.dark);
+  const theme = useSettingsState(s => s.display.theme);
```
liam-fitzgerald, line 288: We now have `useDark()` in `state/join.tsx` that does this logic for you

liam-fitzgerald (approved):
LGTM! Minor nits, nothing severe

pkg/urbit/include/vere/vere.h
```diff
@@ -145,16 +145,17 @@
         } siz;
 
         struct {
-          c3_y* lin_y;                      //  current line (utf8)
-          c3_w  byt_w;                      //  utf8 line-length
-          c3_w  wor_w;                      //  utf32 line-length
-          c3_w  sap_w;                      //  escape chars in line
-          c3_w  cus_w;                      //  cursor position
+          u3_noun lin;                      //  bottom line (stub)
```
joemfb, line 148: I believe we've discussed this before, but I've forgotten: why is the bottom line stored in the driver state as a noun instead of a processed `uv_buf_t` or whatever?

joemfb (commented):
Some requests for clarification and minor nits:

pkg/urbit/include/vere/vere.h
Fang-: Because a blit might _partially_ overwrite the line. Using a `uv_buf_t` that would of course still be trivial for the plaintext case, but keeping _styling_ correct when that's in play would require parsing the `uv_buf_t` for escape sequences etc etc.

joemfb: I see, I hadn't realized that `mir.lin` isn't just the last blit, but has been processed with `+wail:klr:format` and `+pact:klr:format`.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -266,12 +266,22 @@
...
   ++  mo-give  |=(g=gift mo-core(moves [[hen give+g] moves]))
+  ++  mo-talk
+    |=  rup=(each suss tang)
+    ^-  [wire note-arvo]
+    :+  /sys/say  %d
+    ^-  task:dill
+    ?-  -.rup
+      %&  [%text "gall: {(t q)}ed %{(t p)}":[t=trip p.rup]]
+      %|  [%talk leaf+"gall: failed" (flop p.rup)]
+    ==
```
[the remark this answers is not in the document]
Fang-: > can you write a little here about the reason for this %gall change?

I interpreted `%onto` as something gall was using solely to message dill about app load/load failures it wanted to be printed, and figured that should be replaced with gall asking dill explicitly to print some text.

Some more commentary in 5ef68d3.

> looks like drum changes state on `%onto`

Oh, I thought it was just partaking in upgrade notification printing, but on closer inspection it _appears_ to be doing some internal state song-and-dance too... Guessing this is the mechanism that auto-`|link`s started apps? But will have to play with this a little bit to see what exactly is affected.

> @belisarius222 was planning some overlapping changes

We already dm'ed about this a little bit, and the sentence "gall could deliver this info to dill with an %onto, or whatever" was used. (^: So, we'll need to talk about this a bit more I guess. I'm amenable to putting %onto back in if it turns out gall wants update status notifications for non-printing purposes.

pkg/urbit/include/vere/vere.h
joemfb: On second thought, I'm slightly less convinced. `+wail` as called is basically for line-wrapping, but it assumes that each `@c` is an individual glyph. This assumption is usually correct, but not always (see "grapheme clusters" ...). I'm not sure how much of a problem this is, but I wonder if we should bite the bullet and pin the spinner to the bottom right corner, or even a temporary new line.

Fang-: > or even a temporary new line

(Not sure that would work in the first place, but) honestly if we're going there we should just add a persistent status bar... and write a handful of `//TODO`s for things to put in it.

pkg/urbit/vere/io/term.c
```diff
@@ -1424,20 +1430,49 @@ _term_ef_blit(u3_utty* uty_u,
     } break;
 
     case c3__url: {
-      u3_noun txt = u3t(blt);
+      //  platform-agnostically opening the default web browser from within a
+      //  c program is an unsolved problem.
+    } break;
```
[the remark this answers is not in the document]
Fang-: This is actually a little bit complicated. Printing it would give it the same semantics that `%put` has, requiring the sender to be aware of where the cursor is, make sure the url isn't overwriting anything important on-screen, and/or print a newline to make space for it.

Maybe that's fine, and we want to treat it that way because we can't guarantee the url will be opened. Thinking about the ideal future where all clients support all the blits fully, that's not what we want `%url` to do though.

The alternative is pushing responsibility for visually printing the url to whoever is emitting the `%url`.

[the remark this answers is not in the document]
Fang-: Looking into this again, perhaps we can get away with a simple `system()` depending on which OS we're building for. `open $url` on MacOS, `xdg-open $url` on Linux, `start $url` (or similar) on Windows.

But maybe there's some security concern there around apps being able to both write to disk (`%sav`) and open arbitrary strings.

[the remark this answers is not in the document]
Fang-: @philipcmonk the thing I mentioned during our last call, we actually already do:

https://github.com/urbit/urbit/blob/d138ab254b9985a1fdf50cdbdcf78e6f7fd1479e/pkg/arvo/lib/hood/drum.hoon#L916

That is, I was going to update drum to always print the url, in addition to sending a `%url` blit. Turns out I already did this five(!) months ago in 0d8288e.

pkg/arvo/sys/vane/gall.hoon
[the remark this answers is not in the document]
Fang-: Looked into drum's `%onto` handling, and I can't quite make out what the `%&` case was doing. Auto-linking after `|start`, `|link` and `|unlink`ing, and `|fade`ing all work fine, and hit the same code paths they did before (with the exception of `+take-onto` itself, of course).

With that, I'd be reasonably comfortable calling the removal of `%onto` handling in drum a cosmetic change.

pkg/urbit/include/vere/vere.h
joemfb: Having thought about this some more, I'm comfortable with the status quo. The spinner as designed simply conflicts with a 2-dimensional output array, and that will have to be solved at some point. Since it's purely a feature of the runtime, it can be changed without impact to the %dill protocol.

joemfb (changes requested):
@Fang- can you merge master into `release/next-dill`, and then update this PR to resolve the conflicts?

joemfb:
This PR is now ready, the only thing still blocking it is an issue with the mingw windows build.

We're hitting an assertion in libuv, but only when running with `-t` (or -d`):

```
Assertion failed: req->write_buffer.base, file src/win/pipe.c, line 1063
```

Notably, this is the same assertion that we would hit if we were building against an unpatched libuv. @locpyl-tidnyd, do you have any insight into what might be going wrong here?

pkg/urbit/vere/io/term.c
```diff
@@ -664,10 +664,15 @@ _term_io_suck_char(u3_utty* uty_u, c3_y cay_y)
     else {
       if ( (cay_y >= 'a') && (cay_y <= 'z') ) {
         tat_u->esc.ape = c3n;
+        //  XX for backwards compatibility, check kelvin version
+        //  and fallback to [%met @c]
+        //
```
Fang-, line 669: @joemfb I don't think these are necessary? Current dill, as it exists on the network, is already forward-compatible with the inputs from the new runtime. See also #4942.

joemfb: But previous dill isn't, so people with old piers could end up in a situation where they have to juggle binaries to get upgraded (or even just poke around). (Since non-modified `%txt` input will still work, they can probably get unstuck by just typing `|merge ...` -- these concerns are a little more theoretical.)

In general, dropping support for old protocols should happen after some longer period time, and it should be straightforward to continue supporting multiple versions. But those patterns are yet to be established, which is why I'm settling for comments.

Fang-: Fair game. I'm relatively comfortable (and relieved (^: ) to merge this as is, but if any poor sap happens to get hurt by this: blame me.

tacryt-socryp:
So how do you use any of the new terminal features such as session support?

Fang-:
@tacryt-socryp Depends on what context you're talking about.

Drum/sole/shoe agents can distinguish between connecting sessions from the subscription path. For `+on-watch` calls where the path starts with `/sole`, you can call `+path-to-id` from the `/lib/sole` to receive a `sole-id` identifying the session. Use this to register it in your state. Inputs for that session will have that same id in their `%sole-action` pokes. If you use `/lib/shoe`, these ids just come in as arguments on the shoe-specific arms.  
The rest of the sole protocol has not changed. You do not get any new output affordances.

If you're developing a standalone application, it's possible for you to speak the dill protocol directly, like drum does. If a session connects to an agent directly, it gets watched on a `/dill/[session-id]` path. For inputs, it will receive pokes with a `%dill-poke` mark, containing a `[session=@ta belt=dill-belt:dill]`. To send outputs, produce `%dill-blit` facts containing a single `blit:dill`.  
The `blit:dill` includes new affordances, like moving the cursor around on the screen, to draw at arbitrary locations. It's also very primitive, in practice this means you'll write a lot of dumb rendering logic.

The web terminal can create and connect to different sessions. To connect directly to a custom agent, use a session name with the following format: `agent!session-name`. If the target agent does not support the dill protocol, the session will be unusable.  
I have ideas on how to let apps bring you into specific webterm sessions like this in a more user-friendly way, stay tuned.

The runtime currently does not support connecting to anything but the default session, as it always has.

PR #4465, opened 2021-02-19
chatMessage: pass history to overlays

matildepark:
Fixes unfiled bug where window.History was passed instead of react-router's history, so profile overlay redirects didn't work on development stream.


matildepark:
Superseded by #4470

PR #4467, opened 2021-02-20
ci: fix test harness output, report build failures

### LINKED ISSUE #3598: test: build failures in -test not identified in CI

joemfb:
This fixes #3598 (or at least all known causes -- the whole herb/lens/test harness is still pretty fragile). It also fixes a couple current build/test failures. I'm targeting master to fix CI quickly and broadly.

### DIFF
--- a/pkg/arvo/app/test.hoon
+++ b/pkg/arvo/app/test.hoon
@@ -106,11 +106,18 @@
 ++  on-peek   on-peek:def
 ++  on-agent  on-agent:def
 ++  on-arvo
+  =>  |%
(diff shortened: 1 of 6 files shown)


Fang- (approved):
Thank you! All the changes here seem fine. ~zod's work, Joe.

PR #4469, opened 2021-02-20
chat: correctly position mentions

jamesacklin:
Horizontally aligns a mention in a chat message if it is the first thing in the message; adds a margin if it is inline with surrounding text.

Fixes urbit/landscape#468

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -448,7 +448,7 @@ export const Message = ({
               const first = (i) => (i === 0);
               return (
                 <Mention
-                  marginLeft={first(i) ? "-4px" : "0"}
+                  first={first(i)}
                   group={group}
                   scrollWindow={scrollWindow}
                   ship={content.mention}
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -445,8 +445,10 @@ export const Message = ({
                 </Box>
               );
             case 'mention':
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
```diff
@@ -445,8 +445,10 @@ export const Message = ({
                 </Box>
               );
             case 'mention':
+              const first = (i) => (i === 0);
               return (
                 <Mention
+                  marginLeft={first(i) ? "-4px" : "0"}
```
matildepark, line 451: Indigo would do this as just "1" I think. We could stick to theme variables.

(Alternatively — don't margins collapse? I wonder if we could avoid all this logic with clever margin stuff.)

jamesacklin: Unfortunately, margins only collapse vertically. From what I can discern, we want margins around mentions if they're surrounded by other content, and only on the right side if they're not. I will reverse the logic here and thread  to only positively apply margins.

pkg/interface/src/views/components/MentionText.tsx
```diff
@@ -53,10 +54,11 @@ export function Mention(props: {
   }, [showOverlay]);
 
   return (
-    <Box position='relative' display='inline-block' cursor='pointer'>
+    <Box position='relative' display='inline-block' cursor='pointer' {...rest}>
       <Text
         onClick={() => toggleOverlay()}
-        mx={1}
+        marginLeft={first? 0 : 1}
```
matildepark, line 60: I don't know if the ternary should be right up in there, but it does seem to work.

matildepark (approved):

PR #4470, opened 2021-02-20
interface: fixes miniprofile layout, restores actions

jamesacklin:
Conforms the mini-profile layout to design spec. Also normalizes the cursor appearance within the mini-profile and restores the `history.push` actions (whoops).

Here are a few instances of mini-profiles in ascending order of complexity.

A mini-profile with nothing set:
image

A mini-profile with display name set:
image

A mini-profile with display name and avatar set:
image

A mini-profile with display name, avatar, and status set:
image

My own mini-profile:
image

Fixes urbit/landscape#461

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -226,6 +226,7 @@ export const MessageAuthor = ({
   api,
   associations,
   groups,
+  history,
   scrollWindow,
   ...rest
 }) => {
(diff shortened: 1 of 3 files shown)


matildepark:
@urcades did we want header images in these cards?

urcades:
@matildepark we did not want them in cards — for now! 

It's a question worth asking, but one we are intentionally pushing back for the time being. Trying to keep these cards rather bare for the time being as their action set expands (first with messaging/status, soon with payments).

matildepark (approved):

PR #4471, opened 2021-02-22
notifications: fix negative unread count

liam-fitzgerald:
notifications: reduce group stats properly

hark-store: give stats facts for group notifications

Group notifications were being left out of the initial %unreads fact,
which was causing faulty calculation of notification counts  on the
frontend

Fixes urbit/landscape#276


matildepark:
Oh interesting. I didn’t have any group activity harks so makes sense I never saw it.

tacryt-socryp (approved):
Changes look sane. Is there any reason we're storing / processing these group-unreads with separate logic than the since-unreads? In other words: are these stats a separate type specific to groups?

PR #4472, opened 2021-02-22
interface: adds serviceworker support

liam-fitzgerald:
- Adds a serviceworker to the build process
- Adds crossOrigin=anonymous to remote content to allow caching
- Adds headers to ensure that the serviceworker can register itself properly
- Caches all landscape static artifacts and linked images, if possible
- Does not interfere with regular update semantics (i.e. refreshing your browser will pull in the latest interface like normal)
- Enables offline support! It'll still hang on acquiring the subscriptions, but this will be fixed with localstorage backup in the new state management


matildepark:
Last we spoke this would force a refresh on serviceworker hash update. On dev stream this would be pretty often. Any update on how that works here?

matildepark:
@tacryt-socryp Strong opinions against this targeting next-js since it has two small changes to file-server and glob? 

Any divergent deploy to master would get reglobbed with these changes before hitting dev stream ships so I’m inclined to see it play out on the QA crowd.

tacryt-socryp:
Go for it

pkg/interface/config/webpack.prod.js
```diff
@@ -56,20 +60,22 @@ module.exports = {
     new CleanWebpackPlugin(),
     new webpack.DefinePlugin({
       'process.env.LANDSCAPE_STREAM': JSON.stringify(process.env.LANDSCAPE_STREAM),
-      'process.env.LANDSCAPE_SHORTHASH': JSON.stringify(process.env.LANDSCAPE_SHORTHASH),
+      'process.env.LANDSCAPE_SHORTHASH': JSON.stringify(GIT_DESC),
```
matildepark, line 63: Why change this? We pull the hash from the bash env on deploy, what’s this get us?

pkg/interface/src/views/components/RemoteContent.tsx
```diff
@@ -111,6 +111,7 @@ return;
     if (isImage && remoteContentPolicy.imageShown) {
       return this.wrapInLink(
         <BaseImage
+          crossOrigin="anonymous"
```
matildepark, line 114: A few other places to do this! For example, avatars use BaseImage and the like.

matildepark:
> Last we spoke this would force a refresh on serviceworker hash update. On dev stream this would be pretty often. Any update on how that works here?

Oh, I see we do that in JS in https://github.com/urbit/urbit/pull/4472/commits/1d69778225840c635f2b26521ae3a4c6fd5c3418

pkg/interface/config/webpack.prod.js
liam-fitzgerald: For one, I think using the git description is better, because you can see the version number. For two, I wanted to inject it at build time, instead of using an env var.

matildepark: image

👀 this is at least a little preferable to our base hash indicator honestly. @urcades for further contemplation

matildepark (approved):
Content to let this loose on the devnet. Nothing explodes, it persists across refreshes, the instant load feels nice.

pkg/interface/src/serviceworker.js
```diff
@@ -0,0 +1,82 @@
...
+    // Put all cached files in a cache named 'images'
+    cacheName: 'images',
+    plugins: [
+      // Ensure that only requests that result in a 200 status are cached
+      new CacheableResponsePlugin({
+        statuses: [200],
+      }),
+      // Don't cache more than 50 items, and expire them after 30 days
+      new ExpirationPlugin({
+        maxEntries: 50,
```
tacryt-socryp, line 77: Number seems low. Surely we can make use of more space than this. Users clear browser caches regularly and disk space is freely available, so we shouldn't worry about space usage. I think we should ask for 200-300 here.

pkg/interface/src/serviceworker.js
```diff
@@ -0,0 +1,82 @@
...
+        maxAgeSeconds: 60 * 60 * 24 * 30, // 30 Days
```
tacryt-socryp, line 78: Max age seems high. Perhaps 7 days? We don't want to take up space with no longer relevant entries.

pkg/interface/src/serviceworker.js
tacryt-socryp: @matildepark

pkg/interface/src/serviceworker.js
tacryt-socryp: @matildepark

matildepark: Yeah no worries I'm happy to keep tweaking it as we go

PR #4473, opened 2021-02-22
RemoteContent: remove crossOrigin attribute if CORS unsupported

liam-fitzgerald:
Some filehosts do not support CORS requests from third party domains. This breaks image loading from these hosts on current release branch. Fixes this by falling back to not enabling CORS on these requests if we catch an error from the image element. 

cc: @tylershuster


tylershuster:
It works. Some flickering, but I think it's acceptable.

matildepark (approved):

PR #4474, opened 2021-02-22
zuse: time conversion fixes

mrdomino:
State before: in `chrono:userlib`, there were second-resolution
`@da`-to-unix and unix-to-`@da` functions. In `en`/`dejs:format`, there were
millisecond-resolution `@da`-to-unix and unix-to-`@da` functions. The
`@da`-to-unix path in `time:enjs` confusingly rounded to the nearest
millisecond, meaning millisecond *n* was a label for [*n*-0.5, *n*+0.5) rather
than [*n*, *n*+1).

This adds a millisecond-resolution `@da`-to-unix and unix-to-`@da` to
`chrono:userlib`, and a second-resolution conversion to `en`/`dejs:format`.
It makes use of the `chrono:userlib` functions in `en`/`dejs`, and doesn't do
any rounding.

Backwards-incompatible changes:

- made `unt:chrono:userlib` take `@da` rather than `@`.

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -3368,16 +3368,10 @@
       (poq (wit jon))
     ::                                                  ::  ++di:dejs:format
     ++  di                                              ::  millisecond date
-      %+  cu
-        |=  a=@u  ^-  @da
-        (from-unix-ms:chrono:userlib a)
-      ni
+      (cu from-unix-ms:chrono:userlib ni)
     ::                                                  ::  ++du:dejs:format
     ++  du                                              ::  second date
-      %+  cu
-        |=  a=@u  ^-  @da
-        (from-unix:chrono:userlib a)
-      ni
+      (cu from-unix:chrono:userlib ni)
     ::                                                  ::  ++mu:dejs:format
     ++  mu                                              ::  true unit
       |*  wit=fist
@@ -3587,10 +3581,7 @@
       (bind (stud:chrono:userlib p.jon) |=(a=date (year a)))
     ::
     ++  di                                              ::  millisecond date
-      %+  cu
-        |=  a=@u  ^-  @da
-        (from-unix-ms:chrono:userlib a)
-      ni
+      (cu from-unix-ms:chrono:userlib ni)
     ::
     ++  mu                                              ::  true unit
       |*  wit=fist
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -3297,11 +3297,14 @@
       %-  flop
       |-  ^-  ^tape
       ?:(=(0 a) ~ [(add '0' (mod a 10)) $(a (div a 10))])
+    ::                                                  ::  ++sect:enjs:format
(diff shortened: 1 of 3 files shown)


pkg/arvo/sys/zuse.hoon
```diff
@@ -5408,13 +5417,20 @@
     ::                                                  ::::
   ++  chrono  ^?
     |%
-    ::  +from-unix: unix timestamp to @da
+    ::  +from-unix-s: unix seconds to @da
     ::
-    ++  from-unix
+    ++  from-unix-s
```
Fang-, line 5422: idk if this renaming is worth it, or even necessary. "unix timestamp" generally means second resolution. At the very least, consider adding `++  from-unix  from-unix-s` to preserve backwards compatibility.

tacryt-socryp: I like the addition of a distinction between `+from-unix` and `+from-unix-ms`. I don't see a need for `+from-unix-s`, though.

mrdomino: Yeah, I wondered about that. I'll fix it.

mrdomino:
Looking into the test failure, hopefully it's a simple fix and I don't have to add half a millisecond to every timestamp…

mrdomino:
Okay, learned my lesson. Always multiply before divide.

Should work now.

pkg/arvo/sys/zuse.hoon
```diff
@@ -3367,7 +3370,13 @@
...
         |=  a=@u  ^-  @da
-        (add ~1970.1.1 (div (mul ~s1 a) 1.000))
+        (from-unix-ms:chrono:userlib a)
+      ni
+    ::                                                  ::  ++du:dejs:format
+    ++  du                                              ::  second date
+      %+  cu
+        |=  a=@u  ^-  @da
+        (from-unix:chrono:userlib a)
       ni
```
Fang-, line 3380:
```suggestion
      (cu from-unix:chrono:userlib ni)
```

Fang-: Same change for `+di` I guess.

Fang- (commented):
As discussed out of band, "I like your new implementation better even if just because it's more obvious about what it does".

While not strictly necessary, this seems like a fine QoL change.

pkg/arvo/sys/zuse.hoon
mrdomino: Done, thanks.

mrdomino:
As discussed out of band, I fell down this rabbit hole after I stared at the old version too long, thought "that looks funny" eventually decided "it doesn't seem like it needs to do it that way", and then tried to make the world a better place.

There's now a man in a top hat trying to serve me some kind of liquid beverage that he claims is "tea", but I think I'm almost through.

Fang- (approved):
Alright, changes here seem fine. I can't imagine anything depending on the millisecond rounding behavior.

You'll want to phrase the pill commit as `pill: solid` or similar, to match the commit format, and target this against `release/next-sys` instead of master.

PR #4475, opened 2021-02-22
settings: move onto settings store

liam-fitzgerald:
Done:
- Updates the design of the settings page to (mostly) match spec.
- Display and CalmEngine now use settings-store
- Settings state is now managed with zustand

Spec mismatch:
- Stubbed routes for Keyboard, Hosting and Leap as those haven't been implemented yet
- Uses the multipage layout, but I'm unsure whether it should be the long list (cc: @urcades )
- Implements leap category blacklisting, but stubbed because we can't store arrays in settings-store
- S3 is a toplevel settings category, because none of its siblings have been implemented yet (Chat, BTC etc.)
- No images in the navigation page I was unsure how we're handling assets these days, and we might not even be using them if we do the long list design


matildepark:
> stubbed because we can't store arrays in settings-store

We can store strings and JSON.stringify them out of there though

matildepark:
Wait, cool, this is entirely FE. We can target next-js.

liam-fitzgerald:
update: a first pass at leap config is now in!

tylershuster:
I'll race you to the zustand implementation

tacryt-socryp:
I would fully support just... adding arrays (and sets) to settings store: https://github.com/urbit/urbit/pull/4475#issuecomment-783108690

tacryt-socryp (commented):
If it matches the design spec and works, the code itself all seems sane. The Zustand stuff is way prettier than the way we were handling localStorage stuff before.

matildepark:
This is just waiting some localstorage -> settings-store migration, then we can deploy to devnet separately once I test

matildepark (approved):
Made appropriate changes:

- changed the layout to multi-page
- changed leap category storage to an array, not a blob, ripped out the parsing and stringifying
- implemented hideAvatars where appropriate (some spaces were still getting from localstorage)
- gave calmengine a temp icon

tacryt-socryp (approved):
LGTM

PR #4477, opened 2021-02-22
chat: blockquote spacing

jamesacklin:
Addresses the `blockquote` spacing inconsistency as outlined in [urbit/landscape#473](https://github.com/urbit/landscape/issues/473) and adds an even unit of padding to the last chat message in the window.

image

Fixes urbit/landscape#473

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -188,7 +188,8 @@ export default class ChatMessage extends Component<ChatMessageProps> {
       <Box
         ref={this.divRef}
         pt={renderSigil ? 2 : 0}
-        pb={2}
+        pb={isLastMessage ? 5 : 2}
+        pr={5}
         className={containerClass}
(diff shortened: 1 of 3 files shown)


matildepark:
`5` is slightly too high on that last message, I think. `3` seems even? @urcades 

Before / after:

image

image

pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
```diff
@@ -188,7 +188,8 @@ export default class ChatMessage extends Component<ChatMessageProps> {
       <Box
         ref={this.divRef}
         pt={renderSigil ? 2 : 0}
-        pb={2}
+        pb={isLastMessage ? 5 : 2}
```
urcades, line 191: Running this now and yeah, let's change `5` to `4` here — it looks very good

urcades (approved):
The "Last Message" being set to `pb={isLastMessage ? 4 : 2}` seems perfect. Just ran this and it's looking very, very good.

PR #4478, opened 2021-02-22
push-hook: replace +should-proxy-update with +transform-proxy-update

tacryt-socryp:
This is a prerequisite for us to stop trusting client timestamps in `%graph-store`. Note that this retains all of the semantics and behavior of the `+should-proxy-update` arm, while adding the option of transforming the vase that the client supplies to the `%push-hook`.

Fixes: https://github.com/urbit/landscape/issues/270

Note: Needs close review and Ropsten testing before it is ready to go out. It is ready for review though!


ixv:
I get this when trying to boot a fakezod: 
image

tacryt-socryp:
Will boot up a new fakezod and try again.

tacryt-socryp:
@ixv Resolved the issue. It was as simple as `+welp` instead of `+weld`.

urcades:
stamp

ixv:
ropsten net is up

pkg/arvo/lib/push-hook.hoon
```diff
@@ -301,20 +301,20 @@
...
   ::
   ++  poke-update
-    |=  =vase
+    |=  vas=vase
     ^-  (quip card:agent:gall _state)
-    ?>  (should-proxy-update:og vase)
-    =/  wire
-      (make-wire /store)
+    =/  vax=(unit vase)  (transform-proxy-update:og vas)
+    ?>  ?=(^ vax)
```
liam-fitzgerald, line 307: `=/  =vase  (need (tranform-proxy-update:org vas))`

pkg/arvo/app/graph-push-hook.hoon
```diff
@@ -66,28 +66,99 @@
...
-      %tags               %.n
-      %tag-queries        %.n
-      %run-updates        %.n
+      %add-nodes
+    ?.  (is-allowed-add:hc rid nodes.q.update)
+      ~
+    =/  mark  (get-mark:gra rid)
+    ?~  mark  `vas
+    |^
+    =/  transform=$-([index:store post:store atom ?] [index:store post:store])
```
liam-fitzgerald, line 82: This would be way cleaner as just a `=+` and `!<`

liam-fitzgerald (approved):
Couple style nits, nothing major. If this works well on testnet then LGTM!

tacryt-socryp:
@ixv found an issue on the testnet, but I've pushed a fix for it here, and also applied it to `~tacryt`. It's unnecessary for it to be applied to the other ships, since `~tacryt` is the group host. Ready for more testing!

tacryt-socryp:
If this is merged as is, it will regress the "pending message display" feature, so I'm putting this back to draft status and continuing a rewrite to the pending message system to use `node hash` as its unique identifier. As such, I'm building out a thread that we will send all `%add-nodes` events from the frontend to, which will return us a map from index to hash that we can use to generate "unique pending messages" on the frontend that get cleared as they currently do.

tacryt-socryp:
@liam-fitzgerald Can you do a review pass on the %graph-add-nodes thread I just added? It compiles, but have yet to integrate it into the frontend.

liam-fitzgerald:
Yeah `-graph-add-nodes` looks fine for that particular case

tacryt-socryp:
Okay, as of now, this is ready for full-on testing, as it is feature-complete. @matildepark @ixv

matildepark:
Will review the API and store logic but could you merge master back in? The store has a conflict, may affect the review.

tacryt-socryp:
I've discovered issues today and been fixing them. Now, all of the backend actions work properly, but I still need to fix some interface display issues.

liam-fitzgerald (approved):

PR #4481, opened 2021-02-23
chat: correctly pad daybreak divider

jamesacklin:
Pads the message content itself, not the daybreak or unread dividers.

image

Fixes urbit/landscape#475.


matildepark (approved):
lgtm, straightforward

PR #4482, opened 2021-02-23
interface: Changed global state to hooks, replaced global api, linted

tylershuster:
Alright, folks here it begins.

Fixes https://github.com/urbit/landscape/issues/172

**Note** I do not recommend running this on a real ship. It is not yet complete.

However, I would like to begin the review process of this absolutely ridiculous undertaking. This PR finishes the work of converting the threaded global state into state hooks using Zustand. If you know, you know.

It also makes some changes to the `npm` packages, notably cleaning up stale type declarations.

It cleans up the imports in file heads consistently into three different sections: generic npm, urbit/tlon packages, and internal components.

It converts almost all of the rest of the interface into Typescript.

## Known Todos

- [ ] Make the API resume correctly (starts to get `text/html` responses)
- [ ] Settings Store / API
- [ ] S3 Store / API
- [ ] `GraphUpdate` types
- [ ] All the other `TODO`s in the code

I will try to tag the appropriate person when I have a question, but it will probably be @liam-fitzgerald or @tacryt-socryp.


pkg/interface/src/logic/lib/util.ts
```diff
@@ -408,3 +209,8 @@ export function getItemTitle(association: Association) {
   }
   return association.metadata.title || association.resource;
 }
+
+export const stateSetter = <StateType>(fn: (state: StateType) => void, set): void => {
+  // TODO this is a stub for the store debugging
```
tylershuster, line 214: @liam-fitzgerald

pkg/interface/src/views/apps/profile/components/SetStatus.tsx
```diff
@@ -26,8 +29,8 @@ export function SetStatus(props: any) {
   }, [contact]);
 
   const editStatus = () => {
-    api.contacts.edit(ship, { status: _status });
-
+    api.poke(editContact(ship, { status: _status }));
+    // TODO should await promise?
```
tylershuster, line 33: Whoever, should this be `api.poke(...).then(callback)`?

pkg/interface/src/views/apps/publish/components/MetadataForm.tsx
```diff
@@ -47,30 +39,32 @@ const ResetOnPropsChange = (props: { init: FormSchema; book: string }) => {
...
     actions: FormikHelpers<FormSchema>
   ) => {
     try {
       const { name, description } = values;
-      await api.metadata.metadataAdd(
+      await api.poke(metadataAdd(
+        window.ship,
         "publish",
         props.association.resource,
+        // TODO there should be a path here
```
tylershuster, line 61: @liam-fitzgerald or @tacryt-socryp the `metadataAdd` poke requires a path. Which do I use?

matildepark:
Man oh man. There is just too much going on here, @tylershuster. Changing the API calls, *and* state hooks, *and* TypeScript rewrites, and a lint, across the whole interface, in one commit? This should be like 5 different PRs if this is going to merge quickly. I'd recommend taking parts of this work and splitting it out.

tylershuster:
Yeah, I get that.

Here's my thought:

API calls and state hooks have to be in the same, because those are intertwined.
Now that state isn't threaded, you have to go through the whole interface *anyway* to remove all the extraneous props and get the data from a different source.
To do that, the vanilla JS components don't give you any help showing you which props are not useful, or have incorrect types.
The lint isn't really a lint, because it's not thorough — it's just splitting up the import declarations which you have to do *anyway*.

Anyway I get that this is huge. But the biggest part needs to be done together. I'd like to know what the others think. I can split some of it out, but I honestly don't think it will be any faster.

tylershuster:
My biggest motivation in putting this out in this state is just because @liam-fitzgerald wants to see it so badly so take that as you will

matildepark:
My main hesitation is that with the amount of work remaining to do, the longer this open and conflicts (as it will constantly, hitting and linting every single file), the more likely we will have regressions resolving those conflicts. I also don't know how reviewable this code is — it's a 5000 line diff in a single commit with some cruft (eg. why target localhost:9001? That seems to be based on your development).

pkg/interface/config/webpack.dev.js
```diff
@@ -45,21 +45,21 @@ class UrbitShipPlugin {
 let devServer = {
   contentBase: path.join(__dirname, '../dist'),
   hot: true,
-  port: 9000,
+  port: 9001,
```
tylershuster, line 48: @liam-fitzgerald @matildepark this is not permanent — I accidentally left it in, but I think the ServiceWorker cached something that left 9000 in a broken state on my machine

matildepark:
We can help you take this home if you've enabled [allow edits from maintainers](https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork), though — then we don't have to make separate PRs, just add commits to your fork's branch as normal.

tylershuster:
image

tylershuster:
Following a suggestion from @liam-fitzgerald and the work I've done today, I'll take a new approach incrementally incorporating the code I've already written here.

1. Update API library with new definitions
2. Update HTTP API library with small change
3. Replace stores with Zustand stores and adapt reducers to use zustand stores
4. Replace API with http-library implementation, changing subscriptions model in the process
5. Clean up, stripping prop drilling and making code quality changes

PR #4484, opened 2021-02-23
GCP storage support

mrdomino:
This PR achieves:

- [x] poke relevant json fields from the GCP service account JSON into a pier
- [x] make a thread that makes a JWT out of these fields
- [x] send this local JWT to Google in return for an access token
- [x] the thread fails gracefully when the service account JSON hasn't been poked
- [x] Landscape calls this thread whenever it needs to to get a new token
- [x] Landscape can upload files to GCP Storage using the token
- [x] Landscape intelligently switches between GCP Storage and S3 depending on what's available.
- [x] Verify that S3 still works.
- [x] Don't pollute the console with error output if GCP store is not configured.

This does not yet touch the settings UI. Perhaps we should show a checkbox that says "GCP Storage is configured" and hide the S3 settings in that case. Perhaps one day we could even have an upload form for a service account JSON file.

Continues from #4386 (which was accidentally closed, and seems to be un-reopenable, when its merge base was deleted — TIL).


tacryt-socryp (commented):
All seems sane. Looking forward to Landscape integration.

mrdomino:
As of now I am getting actual access tokens that actually work with actual curls to the GCP storage API.

mrdomino:
@tacryt-socryp let me know what you think — this just gets GCP support into the Links app as a proof-of-concept.

Next I will probably write a `useStorage` that uses GCP if a token is set, else S3 if access / secret keys are set, else neither, and replace all calls to `useS3` with `useStorage`. `useGcp` will most likely get removed.

pkg/interface/src/logic/api/gcp.ts
```diff
@@ -4,40 +4,20 @@ import {GcpToken} from '../types/gcp-state';
...
-
-  private async refreshToken() {
-    const token = await this.spider('noun', 'gcp-token', 'get-gcp-token', {});
-    return token['gcp-token'];
+        if (token['gcp-token'] !== undefined &&
+            token['gcp-token']['expiresIn'] !== undefined &&
+            typeof(token['gcp-token']['expiresIn']) === 'number') {
+          return Promise.resolve(token['gcp-token']['expiresIn']);
+        }
+        return Promise.reject({reason: 'invalid token'});
```
mrdomino, line 20: @tacryt-socryp is this still bad? If so, how can I get `GcpManager` to listen to storage updates, or what should I do instead?

mrdomino:
@tacryt-socryp actually now it's ready for review

pkg/interface/src/logic/api/gcp.ts
mrdomino: The API no longer does this.

pkg/interface/src/logic/lib/GcpClient.ts
```diff
@@ -0,0 +1,65 @@
...
+    const headers = new Headers();
+    headers.append('Authorization', `Bearer ${this.#accessKey}`);
+    headers.append('Content-Type', ContentType);
+    const response = await fetch(url, {
+      method: 'POST',
+      mode: 'cors',
+      cache: 'default',
+      headers,
+      referrerPolicy: 'no-referrer',
+      body: Body
```
tacryt-socryp, line 47: This does not appear to support streaming of large files into the bucket, which I believe our current implementation does. Any plans?

pkg/interface/src/logic/lib/gcpManager.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+    this.#running = false;
+    if (this.#timeoutId !== null) {
+      clearTimeout(this.#timeoutId);
+      this.#timeoutId = null;
+    }
+  }
+
+  restart() {
+    if (this.#running)
```
tacryt-socryp, line 58: Always include {} after if statements, even though they are not required.

pkg/interface/src/logic/lib/gcpManager.ts
```diff
@@ -0,0 +1,139 @@
...
+      .catch((reason) => {
+        this.#consecutiveFailures++;
+        console.warn('GcpManager token refresh failed; retrying with backoff');
+        this.refreshAfter(this.backoffInterval());
+      });
+  }
+
+  private refreshAfter(durationMs) {
+    if (!this.#running)
+      return;
```
tacryt-socryp, line 109: if statements should have {}

pkg/interface/src/logic/lib/useStorage.ts
```diff
@@ -1,41 +1,57 @@
...
+  const client = useRef<StorageClient | null>(null);
 
   useEffect(() => {
-    if (!s3.credentials) {
-      return;
+    // prefer GCP if available, else use S3.
+    if (gcp.token !== undefined) {
+      client.current = new GcpClient(gcp.token.accessKey);
+    } else {
+      // XX ships currently always have S3 credentials, but the fields are all
```
tacryt-socryp, line 29: I see. Well, empty strings do evaluate to false!

pkg/interface/src/views/apps/chat/ChatResource.tsx
```diff
@@ -180,6 +179,7 @@ export function ChatResource(props: ChatResourceProps) {
           (!showBanner && hasLoadedAllowed) ? contacts : modifiedContacts
         }
         onUnmount={appendUnsent}
+        gcp={props.gcp}
```
tacryt-socryp, line 182: Instead of passing down gcp and s3 separately, please put them into the same prop and call it `storage`.

tacryt-socryp (commented):
Sane, LGTM. A few comments.

pkg/interface/src/logic/lib/GcpClient.ts
matildepark: I don't believe we do - https://github.com/urbit/landscape/issues/294

pkg/interface/src/views/apps/chat/ChatResource.tsx
mrdomino: Oh man, that is definitely sane and a mf'er to implement. On it...

tacryt-socryp: Sounds good.

pkg/interface/src/logic/lib/GcpClient.ts
tacryt-socryp: Sounds good.

tacryt-socryp (approved):

pkg/interface/src/logic/lib/useStorage.ts
mrdomino: JavaScript!

pkg/interface/src/views/apps/chat/ChatResource.tsx
mrdomino: Hmm, I suppose this is most parsimoniously supported by starting from the store state itself – creating a top-level storage object and changing the reducers to match? Starting down that path assuming it is the way...

matildepark: Except all of the stores are changing (https://github.com/urbit/urbit/pull/4510) — I would continue to recommend making this PR as simple as possible and deploying it sooner than later to prevent colliding with that stuff.

mrdomino: Done.

matildepark: Just merge — if you can push to get next-sys on master within the week, may be moot commentary.

PR #4485, opened 2021-02-23
landscape: bring placeholder group joins to parity

matildepark:
cc @urcades 

image

Fixes urbit/landscape#478

### DIFF
--- a/pkg/interface/src/views/components/GroupLink.tsx
+++ b/pkg/interface/src/views/components/GroupLink.tsx
@@ -71,20 +71,15 @@ export function GroupLink(
         pr="2"
         onClick={showModal}
         cursor='pointer'
+        opacity={preview ? '1' : '0.6'}
       >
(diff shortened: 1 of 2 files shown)


urcades:
Looking wonderful — two very minor requests: 

1. Let us be sure to apply the "small" border radius style to the group tiles (this is marked in figma as 4pt/px, but if there is a variable that represents this, let's use that of course.
2. Rather than use the language of "Unknown members", I'd prefer if we pointed to the fact that the number was awaiting an async update — something like, "Member/peer count pending" or "Fetching peer count" would be preferable — anything that communicates the "live"ness of the state would be good here.

matildepark:
image

urcades:
beautiful — spiritual stamp

—
~fabled-faster
https://urbit.org/

jamesacklin (approved):
Looks wonderful.

PR #4486, opened 2021-02-23
links, publish, notifications: work with flattened contacts

matildepark:
Both Links and Publish were still accessing group paths in contacts; additionally some sig mismatching prevented the Author component from rendering, anyhow. We desig aggressively to resig for the contact in the component; we pass the whole contacts object now.

Fixes urbit/landscape#457, fixes urbit/landscape#458


matildepark:
Also restores contacts inside notifications. I just pipe the author component to prevent duplicate code but it produces a slight design change. @urcades?

image

liam-fitzgerald (approved):

PR #4490, opened 2021-02-24
clay: static mark building and casting

belisarius222:
Refactors Clay to build (vases of) statically typed mark cores and mark conversion gates, with corresponding Ford runes.  The dynamically typed mark core and mark conversion builders now call the statically typed ones internally.

The new Ford runes can be used like this:
```hoon
::  load a fully resolved mark core for the %mime mark and name it .mim
::
/%  mim  %mime
::  load a statically typed gate to convert from %mime to %html and name it .gat
::
/$  gat  %mime  %html
::
=/  default-mime  bunt:mim
=/  converted-html  (gat default-mime)
...
```
A file can include multiple `/%` and `/$` runes, just like with `/=` and `/*` runes.  For now, the runes must be included in the same order as in the `$pile:clay` datastructure, as defined in `/sys/lull/hoon`.  There are now enough different kinds of Ford runes that I suspect it's worth letting the parser accept them in any order, but that's out of scope for this PR.

The `+load` implementation clears the old Ford caches, since new mark cores and conversion gates will not be noun-equal to ones.


belisarius222:
Closing; superseded by https://github.com/urbit/urbit/pull/4491, which should fix a git snafu.

PR #4491, opened 2021-02-24
clay: static mark building and casting

belisarius222:
Refactors Clay to build (vases of) statically typed mark cores and mark conversion gates, with corresponding Ford runes.  The dynamically typed mark core and mark conversion builders now call the statically typed ones internally.

The new Ford runes can be used like this:
```hoon
::  load a fully resolved mark core for the %mime mark and name it .mim
::
/%  mim  %mime
::  load a statically typed gate to convert from %mime to %html and name it .gat
::
/$  gat  %mime  %html
::
=/  default-mime  bunt:mim
=/  converted-html  (gat default-mime)
...
```
A file can include multiple `/%` and `/$` runes, just like with `/=` and `/*` runes.  For now, the runes must be included in the same order as in the `$pile:clay` datastructure, as defined in `/sys/lull/hoon`.  There are now enough different kinds of Ford runes that I suspect it's worth letting the parser accept them in any order, but that's out of scope for this PR.

The `+load` implementation clears the old Ford caches, since new mark cores and conversion gates will not be noun-equal to ones.


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1614,24 +1686,26 @@
     ::
     ++  checkout-changes
       |=  [=ford=args:ford:fusion changes=(map path (each page lobe))]
-      =/  cans=(list [=path change=(each page lobe)])  ~(tap by changes)
-      |-  ^-  [(map path [=lobe =cage]) ford-cache]
-      ?~  cans
-        [~ ford-cache.ford-args]
+      ^-  [(map path [=lobe =cage]) ford-cache]
+      %+  roll  `(list [path (each page lobe)])`~(tap by changes)
```
philipcmonk, line 1690: The recursion->roll change is cosmetic, right?  The naked recursion is clearer to me, but I don't care much either way.

philipcmonk (approved):
The code comes out looking real good.  I'll release this tonight.

philipcmonk:
Was this ever run against a ship built on master?  The +load doesn't seem to punch out the ford-cache in `hoy`.  I'll dig that out.

Also, it would be nice to be able to read these through a scry, though I guess that's rarely going to be needed.  Threads like `-build-file/mark/cast` would be good though.

belisarius222:
A stack trace I saw in development included many hundreds of calls to this function, so I realized it wasn't tail-recursive.  It can be written tail-recursively, though, and it fits pretty neatly into +roll, whose jet is faster and uses less memory than hand-rolled recursion.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
Ah, that makes sense.  It will be called quite a bit, yeah

belisarius222:
Yes, I merged this code into a fakezod that I'd booted off master, and checked that I could still build files and make commits.  It's quite possible I missed some ford caches, though, so good catch.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
Hm, well, it doesn't run on my machine, and when I punch out the ford cache in `hoy`, it... crashes on the assertion that it's at version `7` because apparently it's at version `~2000.1.1`, which I can only see by casting to `*` because that's clearly contrary to its type.  I think there must be a runtime bug here?

philipcmonk:
https://github.com/urbit/urbit/commit/4826c83f14d519ba3e8d935c08af96c41bb266f4

This is my diff on top of release/next-sys, this is on a fresh fakezod booted from v1.2 with... some random pill.  I'll try the mainnet pill, but I doubt that'll make a difference.

PR #4493, opened 2021-02-24
launch: fix modal backgrounds

matildepark:
I used `washedGray` and forgot it was an opacity colour — falling back to the hex here.

### DIFF
--- a/pkg/interface/src/views/apps/launch/app.js
+++ b/pkg/interface/src/views/apps/launch/app.js
@@ -48,12 +48,12 @@ export default function LaunchApp(props) {
       position={["relative", "absolute"]}
       left="0"
       bottom="0"
-      bg="#f2f2f2"
+      backgroundColor="white"
       ml={3}
       mb={3}
       borderRadius={2}
+      overflow='hidden'
       fontSize={0}
-      p={2}
       cursor="pointer"
       onClick={() => {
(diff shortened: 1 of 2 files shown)


jamesacklin (approved):
image

Looking much better.

matildepark:
Let me see if I can make it match better in dark mode, it sticks out now.

matildepark:
Screenshot 2021-02-24 at 5 11 01 PM

urcades:
The buttons themselves looks beautiful with a background, and are looking good on the 'default' dark background, in isolation. In conjunction with the way we render tiles in dark mode, I do have a lurking feeling that this isn't as visually 'put together' as it could be, mostly due to how we actually render the tiles (not the buttons):

Screen Shot 2021-02-24 at 5 17 13 PM

...The buttons themselves (the subject of this issue) feel fine, but there's something I can't put my finger on with how the surrounding elements are rendered. Maybe it's the extreme contrast of the titles (white) with the very light borders that makes this a little uncomfortable to parse.

jamesacklin:
The buttons themselves look good to me; I can rationalize the tile borders for the time being or resolve seperately.

urcades (approved):
Stamping for the button change in particular, but feel like we need to render tiles better in dark mode, maybe via a separate PR

PR #4494, opened 2021-02-25
settings: add array type

ixv:


ixv:
Might make sense to just cherry-pick this onto your branch and close this PR @liam-fitzgerald. I don't know if we need to do a separate release for this.

matildepark:
This should target master, go out first. Then Liam’s work hits dev stream for a week, then it goes out.

liam-fitzgerald (approved):
LGTM! Pending @tacryt-socryp review, will cherry-pick

tacryt-socryp:
I think the array type should be a list of either %n, %b, or %s. Do we really need arrays of arrays?

ixv:
@matildepark whats the reasoning behind that? As far as I know theres no urgent need for this other than Liam's PR. Happy to do it either way, just wondering.

@tacryt-socryp I prefer it just because its a simpler implementation, and matches the json type more closely. I don't see any obvious use cases for it, but then we said that about arrays at first too.

tacryt-socryp:
Sounds good

ohAitch:
Does this need a state adapter/revision? Shouldn't it nest.

matildepark:
> @matildepark whats the reasoning behind that? As far as I know theres no urgent need for this other than Liam's PR. Happy to do it either way, just wondering.

If you merge this with Liam’s PR it becomes a full stack feature, which means we either release it with no dev stream testing (bad), hold it for next userspace merge (bad), or merge it into next-js (breach of discipline) for testing. 

If you merge this into the next hotfix then it’s primed Liam’s PR for dev stream testing without impacting anyone. It provides the smoothest deploy experience. Incremental deploys like that are generally preferable.

ixv:
@ohAitch it satisfies my ocd to do it this way

ixv:
@matildepark alright, thanks for explaining

PR #4495, opened 2021-02-25
virtualScroller: rework

liam-fitzgerald:
- We now virtualize 'treadmill' style i.e. by rendering a subset of the
list into a window. When the scroll position gets close to an edge, we
save our scroll position, adjust the subset and then restore our scroll
position
- Instead of propdrilling the `measure` prop everywhere, we now use a Context to manipulate the scroll when the size of chlidren changes
- Available @ `|ota ~hastuc-dibtux %virt`
TODO:
- Good unread behaviour 
- Cross browser support (tested on chrome only rn)
- Unbreak links virtualisation


pkg/interface/src/views/components/VirtualScroller.tsx
```diff
@@ -303,8 +303,10 @@ export default class VirtualScroller<T> extends Component<VirtualScrollerProps<T
     const ref = this.childRefs.get(this.savedIndex)!;
     //ref.scrollIntoView();
     const newScrollTop = this.window.scrollHeight - ref.offsetTop - this.savedDistance;
-    
+
+    this.window.style['-webkit-overflow-scrolling'] = 'auto';
     this.window.scrollTop = newScrollTop;
+    this.window.style['-webkit-overflow-scrolling'] = 'touch';
```
tylershuster, line 309: What's this about?

matildepark: I remember patching this on Urbit.org once where touch scroll would feel really janky and unnatural. It wouldn’t glide.

matildepark:
Things that are weird so far:

- The scrollbar jumping around but you know this. Prescribing "at 25% of scroll, load in previous page of messages"
- Scrolling down can stick if we're waiting on backend — and why? Don't we _have_ these messages? Scrolling down shouldn't leave you trapped in scroll limbo potentially

liam-fitzgerald:
> Things that are weird so far:
> 
> * The scrollbar jumping around but you know this. Prescribing "at 25% of scroll, load in previous page of messages"

Ok but we also want to unload the previous page, otherwise theres not point to virtualisation

> * Scrolling down can stick if we're waiting on backend — and why? Don't we _have_ these messages? Scrolling down shouldn't leave you trapped in scroll limbo potentially

What browser was this on?

matildepark:
> Ok but we also want to unload the previous page, otherwise theres not point to virtualisation

I just mean to fake the scrollbar in this way. Like, "if it looks like I'm near the top, let's prepare the page above us"

> What browser was this on?

Chrome. I scroll up and eventually hit a wall, scroll down and also hit a wall.

urcades:
@liam-fitzgerald I'm noticing the scrolling getting stuck in safari!

Have some videos of a "current" scrolling vs. the new scrolling:

Current:

https://user-images.githubusercontent.com/1195363/109236099-ed9c9e80-779c-11eb-8c93-c97d4e5fd4c3.mp4

Note how long the group preview takes to load and the 'whiteout' of the scroll before it loads things in at the top edge

The new implementation appears to improve on both of these fronts:

https://user-images.githubusercontent.com/1195363/109236188-1ae94c80-779d-11eb-8479-59f8bca2ccc1.mp4

I do hit the wall @matildepark mentions though. And I mentioned this in Interface, but the 'treadmill' area seems quite short for my tastes

lycurgus:
I've had a play with it too - a couple of quick thoughts:

1. 1. I wonder that the scrollbar needs to be divorced from actually representing our position in what's _loaded_ and instead just represent our position in history altogether (approximately is fine, I expect). That solves the jumping as well as the short-feeling treadmill that Éd mentions.
   2. Not sure if there should be a transition point (e.g. scroll within recent history while we're near "now", and switch to scrolling in full history once we load some) or just always have the long scroll. Something like U-C general would probably have an imperceptible scroll handle, though...
2. There needs to be a way to get back to "now" other than just scrolling all the way back down - something like Discord's "jump to present" bar
3. Might also be nice to have a "go back" feature to mirror no. 2 above? Probably the most-used would be "jump to last read message" but I could also see "go back a week" or "go back x00 messages"

urcades:
Note: Checking out the latest updates to the scroller, noticing a few bugs or otherwise undesirable behavior, will slot each observation in as a single comment to this PR

~

Issue: "New Messages" banner disappears instantly when visiting a chat

Check out this video:

https://user-images.githubusercontent.com/1195363/109457408-d1ad2d00-7a28-11eb-9b6c-c2f3395a59b0.mp4

In the first channel I click into, note that there's no visible blue divider specifying "new messages below". Because the "New Messages" banner disappears instantly, we really have no idea how many messages we have to scroll up to see. This is really the core experiential issue here.

Interestingly, in the second channel I click into, you can see the blue line actually visible in the viewport — there are only two new messages, we can see them under the blue line, and it's clear there's no back-scroll to view — this is actually **correct** behavior! The "New Messages" banner still flashes for a moment, and we should do what we can (perhaps in a separate PR?) to ensure that no "New Messages" banner pops up, and only a visual blue line appears if there are _only_ new messages in the initially-loaded viewport.

urcades:
Issue: Scrollbar has very jerky moments when it hits a treadmill loading seam, or when it hits a point where it needs to actually use a circle-loader to load another treadmill.

Pay attention to the right side of the viewport in this video:

https://user-images.githubusercontent.com/1195363/109458919-c0195480-7a2b-11eb-8f2a-77721aec0754.mp4

Basically it seems as though whenever the treadmill hits a 'seam', we basically get the scrollbar hopping around. The most egregious displacement appears to occur when a spinner is prompted at what I assume to be the treadmill's edge.

urcades:
issue: the bottom of the scroll has particularly bad 'scrolling stickiness' — initial launch is slow, while scrolling above seems good

Please watch this video: 

https://user-images.githubusercontent.com/1195363/109459661-42eedf00-7a2d-11eb-8fff-23152364fec4.mp4

For context, it's basically me at the bottom of a chat's scroll, gently scrolling upwards, and then scrolling back downwards, to hit the bottom again. There's a highly perceptible 'stickiness' or 'catching' action that happens when you either 1) scroll away from the bottom of the scroll or 2) scroll back down and hit the bottom of the scroll.

urcades:
The above issues are really the only one's i'm running into — I'd like to note that in-between "seams", and while anywhere BUT the bottom of the scroll, the scrolling is essentially perfect and very smooth. Group previews load their metadata close to instantly for the most part, and i have yet to see an image load into place.

These 'seams' in between the treadmills/rendered areas are rough, idk if there's a sane way to smooth them over, but it'd make this all feel very nice.

liam-fitzgerald:
@urcades If you're running the dev env then running a prod copy should help with the lagginess. Just merge from the %virt desk on my ship

urcades:
Leaving a comment here to indicate a few new notes after running the prod copy:

1. The mobile version of the virtual scroller is perfect. It needs absolutely no changes. 

2. The final/remaining issue with the desktop scrolling behavior is a small moment of lag that seemingly occurs in the bottom margin of the chat view, see image below:

Screen Shot 2021-03-01 at 10 13 25 PM

3. A rare issue I've been encountering across both mobile + desktop is a strange bug that occurs while scrolling upward. I'm not sure what behavior triggers it, but I sometimes end up with a few rows of content sandwiched between two loaders:

Screen Shot 2021-03-01 at 10 52 22 PM

Screen Shot 2021-03-01 at 10 52 27 PM

urcades:
I'm affixing a spiritual stamp to this project, assuming the above issues are addressed.

matildepark:
Is this just waiting on the unread banner now? This is still marked as draft.

urcades:
> Is this just waiting on the unread banner now? This is still marked as draft.

Related to this, noting a small regression where clicking on the banner no longer auto-scrolls to the point in time where it specifies.

matildepark (approved):
Hashed out some jank out of band. I am content.

As per @urcades, drag to scroll is good to get to, but non-blocking.

PR #4496, opened 2021-02-25
api: brought to parity with livenet

tylershuster:
Integrates new changes from livenet, fixes build tool.

@liam-fitzgerald


liam-fitzgerald (approved):
LGTM, if this builds properly

matildepark:
> LGTM, if this builds properly

This is something to establish _before_ stamping!

tylershuster:
Most of these don't get used in building anyway; it's just prep for the next step. But it builds at least for development

PR #4497, opened 2021-02-25
http-api: changed eventsource tool

tylershuster:
Changes the build system for the `http-api` package,  exports the types in an importable manner, and changes from native eventsource to `@microsoft/fetch-event-source`


tacryt-socryp:
What is the advantage of using the Microsoft event source library?

tylershuster:
It uses `fetch` in streaming mode which supports resuming a connection. This should improve time-to-wake performance. It also precludes the necessity of using a polyfill for both fetch and eventsource on the serverside, because it can take an injected fetch object. Because fetch is being actively developed while eventsource has generally stagnated, @liam-fitzgerald felt this was the best course of action. Merely judging by the difficulty I had getting the eventsource polyfill to compile, I'm inclined to agree, plus the theoretical arguments outlined above.

Fang-:
For prior discussion: #4004 and #4157. The former can probably be closed once this is in.

tylershuster:
Yes, assuming this works as intended, this being implemented is 2 PRs away (after state conversion). I had some problems but I'm realizing that was probably due to trying my to resume a connection that hadn't died!

tacryt-socryp:
@tylershuster Is this implementation of Channel.js going to live alongside the existing one, or is it replacing it as the "official JS way of talking to Urbit"? I'm now convinced that this is good and proper for Landscape, but still like how minimal the other Channel.js is for "official implementation spec" (with few dependencies).

tylershuster:
I think "live alongside the existing one" is the approach we're taking. I like channel.js as a quick and dirty way of connecting to Urbit — the key words being both quick (no dependencies) and dirty (no extra utility). It's a nice single-file solution in a pinch, but I think most people will want to use this library.

pkg/npm/http-api/example/browser.js
```diff
@@ -1,17 +1,32 @@
...
 /*!********************************!*\
   !*** ./src/example/browser.js ***!
   \********************************/
-/*! unknown exports (runtime-defined) */
-/*! runtime requirements:  */
+/***/ (() => {
+
 eval("// import Urbit from '../../dist/browser';\n// window.Urbit = Urbit;\n\n//# sourceURL=webpack://@urbit/http-api/./src/example/browser.js?");
+
+/***/ })
```
liam-fitzgerald, line 20: Can you add this to the `.gitignore`?

pkg/npm/http-api/example/node.js
```diff
@@ -1,17 +1,32 @@
...
   !*** ./src/example/node.js ***!
   \*****************************/
-/*! unknown exports (runtime-defined) */
-/*! runtime requirements:  */
+/***/ (() => {
+
 eval("// import Urbit from '../../dist/index';\n// async function blastOff() {\n//   const airlock = await Urbit.authenticate({\n//     ship: 'zod',\n//     url: 'localhost:8080',\n//     code: 'lidlut-tabwed-pillex-ridrup',\n//     verbose: true\n//   });\n//   airlock.subscribe('chat-view', '/primary');\n// }\n// blastOff();\n\n//# sourceURL=webpack://@urbit/http-api/./src/example/node.js?");
+
+/***/ })
+
```
liam-fitzgerald, line 21: likewise with `.gitignore`

liam-fitzgerald (approved):
LGTM, getting a little bugged by the built examples, but it shouldn't block this

tylershuster:
Understood. This is something I'd like to discuss with @jtobin when we auto-build releases

PR #4498, opened 2021-02-25
ResourceSkeleton: flex and truncate description

matildepark:
It was driving me a little crazy.

Before:

image

After:

image

### DIFF
--- a/pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
+++ b/pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
@@ -1,5 +1,5 @@
 import React, { ReactElement, ReactNode } from 'react';
-import { Icon, Box, Col, Text } from '@tlon/indigo-react';
+import { Icon, Box, Col, Row, Text } from '@tlon/indigo-react';
 import styled from 'styled-components';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
```diff
@@ -12,7 +12,7 @@ import GlobalApi from '~/logic/api/global';
 import { isWriter } from '~/logic/lib/group';
 import { getItemTitle } from '~/logic/lib/util';
 
-const TruncatedBox = styled(Box)`
+const TruncatedText = styled(RichText)`
   white-space: pre;
   text-overflow: ellipsis;
```
liam-fitzgerald, line 17: wonder if setting
```css
a {
  max-width: 100px;
}
```
would be worthwhile?

liam-fitzgerald (approved):
LGTM!

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
matildepark: ? Why? That seems trickier than it’s worth. The thing here is that truncating the box won’t properly flex the text dynamically. If we let the flex go up each container then we don’t need fixed pixel widths. It’s much more fluid.

PR #4506, opened 2021-02-26
fix Ames comet handshake

belisarius222:
This fixes a bug where if a ship tried to send an Ames message to a comet before the comet had contacted it, it would ignore all packets from that comet without ever applying any self-attestation packets the comet sent it.  Confirmed on a live repro with @ixv that previously stuck `|hi`'s go through after applying this change.

This also changes the behavior on hearing a "shut" encrypted packet from a comet whose keys we don't know yet but to whom we have messages enqueued.  Instead of no-op'ing, we now crash, expecting an "open" unencrypted self-attestation packet instead.  It's difficult to imagine this working correctly any other way without adding an "open or shut" bit back into the packet format.

If we tried to enqueue them, for example, then we wouldn't know whether they decrypt properly until we've heard the self-attestation packet, which could be used in a resource exhaustion attack by getting us to hold onto and then try to decrypt a bunch of bogus packets.  In general with comets I think it's probably safer to err on the side of as little queueing as possible.

I think crashing is fine, though.  It arguably fits better into our general paradigm of crashing on invalid packets, and it shouldn't be harmful.  If a few shut packets arrive out of order and crash before the attestation packet arrives, they should get processed successfully after the attestation packet arrives and the comet sends them again.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1217,7 +1217,7 @@
       on-hear-forward
     ::
     ?:  ?&  ?=(%pawn (clan:title sndr.packet))
-            !(~(has by peers.ames-state) sndr.packet)
+            !?=([~ %known *] (~(get by peers.ames-state) sndr.packet))
         ==
       on-hear-open
     on-hear-shut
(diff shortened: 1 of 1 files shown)


joemfb (approved):
This makes sense -- LGTM

matildepark:
@ixv asked if we could include in our next minor deploy tonight; @philipcmonk signed off, so just heads up; if this is based on next-sys we may cherry-pick

belisarius222:
go for it

—
~rovnys-ricfer
https://urbit.org

liam-fitzgerald:
This went out in the OTA on thursday, however I had to rebase the branch onto master, so it didn't get automerged. Closing

PR #4508, opened 2021-02-26
meta: automatically deploy group-timer

liam-fitzgerald:
- Auto globs and deploys to ~difmex-passed
- Automerges master into ops/group-timer
- Need to verify that CI has access to ~difmex-passed


jtobin (approved):
LGTM.  As you guessed, I need to grant the service account permissions to touch ~difmex-passed before this will work.  Will merge this after doing that.

PR #4510, opened 2021-02-26
interface: convert stores to zustand

tylershuster:
This PR moves all the stores to zustand. I will try to make a separate commit for each store, which should implement the store, convert the reducer, and strip all the drilled props belonging to that store. Each commit will build on the last and should compile completely.

- [x] contact
- [x] graph
- [x] groups
- [x] hark
- [x] invite
- [x] launch
- [x] metadata
- [x] ~s3~ storage
- [x] settings


pkg/interface/src/logic/state/contacts.tsx
```diff
@@ -0,0 +1,51 @@
...
+  //   ship,
+  //   personal
+  // ): Promise<boolean> => {
+  //   const isPersonal = personal ? 'true' : 'false';
+  //   const api = useApi();
+  //   return api.scry({
+  //     app: 'contact-store',
+  //     path: `/is-allowed/${entity}/${name}/${ship}/${isPersonal}`
+  //   });
+  // },
```
tylershuster, line 34: trust the plan

tylershuster:
feelsgoodman.jpg

tylershuster:
**Note** (this should be its own issue) — non-serializable types such as `Set<string>` such as `graphKeys` cannot be persisted in localStorage and will therefore be ignored **for persistence** in this pass. We should write a custom storage handler for each store to properly dehydrate these from localStorage. The current solution will not offer any performance benefits for those stores, but it will also not inflict any performance hits

tylershuster:
Alright, this has some issues with immutability that we need to address but I think generally it's ready to review

tylershuster:
Ok, there's definitely something weird going on after the initial population of the store, likely related to the use of immer in the setter. I'm working on a reusable (and also usable) pattern for reducing state.

tylershuster:
Now the stores actually work, but at the cost of losing immutability. I don't think we're ever actually using the benefits of `immer`, so I'm not sad but I would like to figure out how to do this properly. The relevant method is in `lib/util/reduceState`

tylershuster:
@liam-fitzgerald We should talk at some point about the methods I'm using to weed out any potential problems before I convert `settings` to use the same pattern

liam-fitzgerald (approved):

tylershuster:
Copying from out of band:

- [ ] Converting settings store
- [ ] I forgot to do GroupViewReducer
- [ ] localreducer?
- [ ] the connection handling I don't quite understand
- [ ] implementing the debugStore
- [ ] deleting all the old references to the global store members

matildepark:
All these todos and somehow this got an approve stamp.

tylershuster:
He talked to me after the stamp. It won't break anything if it goes in, just needs some cleaning up

tylershuster:
So sassy

matildepark:
Hmm. Merge next-js back in?

tylershuster:
Ok, but don't merge this until you've built it. There's a weird error that *shouldn't* be happening, but is.

tylershuster:
To be exceedingly clear, there are a few niggling things in here which should be cleaned up (like the base hash and connection state) but are better addressed once the API is switched out.

pkg/interface/src/logic/lib/withState.tsx
```diff
@@ -20,4 +20,27 @@ const withState = <
...
+          keys.reduce(
+            (object, key) => ({ ...object, [key]: state[key] }), {}
+          )
+        })
+        : store();
+      Object.assign(stateProps, storeProps);
+    })
+    return <Component ref={ref} {...stateProps} {...props} />
+  });
+}
```
tylershuster, line 44: @liam-fitzgerald Clear as mud?

tylershuster:
I will resolve these conflicts once stamped

pkg/interface/src/logic/lib/withState.tsx
liam-fitzgerald: Nah that's not that bad

pkg/interface/src/views/apps/graph/app.js
```diff
@@ -2,58 +2,53 @@ import React, { PureComponent } from 'react';
...
-            } else if(!!association) {
-              props.history.push(`/~landscape/home/resource/${association.metadata.module}${path}`);
+import useGraphState from '~/logic/state/graph';
+import useMetadataState from '~/logic/state/metadata';
+
+const GraphApp = (props) => {
+  const associations= useMetadataState(state => state.associations);
+  const graphKeys = useGraphState(state => state.graphKeys);
+
+  const { api } = this.props;
```
liam-fitzgerald, line 12: hm??

pkg/interface/src/views/landscape/components/JoinGroup.tsx
```diff
@@ -75,12 +77,12 @@ export function JoinGroup(props: JoinGroupProps): ReactElement {
     await api.groups.join(ship, name);
     try {
       await waiter((p: JoinGroupProps) => {
```
liam-fitzgerald, line 79: You're going to want to pass the hook values into `useWaitForProps` and then check there

liam-fitzgerald (changes requested):
Did a full read of this. Looks good. Obviously needs an update for the new GCP storage stuff, merge conflicts and comments need to be resolved.

pkg/interface/src/views/apps/graph/app.js
tylershuster: hm hm? All of these are used in the file.

tylershuster:
@liam-fitzgerald  Can you give this a run? I'm getting this weird error that seems to be a result of something failing to import correctly, but I see no reason why it wouldn't.

liam-fitzgerald:
don't see any issues when building this, other than a missing quote on L110 of `lib/tutorialModal.ts`

tylershuster:
Alright, I will wipe and reload. Once I've built it and checked I'll make a final commit wrapping this up (God willing).

tylershuster:
@liam-fitzgerald discovered a circular dependency that was causing the build to fail. Once I fixed that I discovered a few other problems clicking around the interface. Now it works and in my limited testing I haven't been able to break it. I think this is ready for testnet

tylershuster:
one day

matildepark:
Conflicts again. Otherwise we can start shoving it into next-js once it has an all clear and we reasonably expect it to work.

tylershuster:
I'm under the impression it has an all-clear, I just don't want to keep cleaning up conflicts. @liam-fitzgerald ?

liam-fitzgerald:
merge, and then i'll test. If works, i'll merge

matildepark:
We were basically holding bc next-js was going into the next release (today) and this needs some dev stream time first. But this blocks a lot of big ticket so it should be expedited if we have baseline confidence in its stability.

tylershuster:
aye aye. I would not be comfortable with this going out today; needs a little bit of field experience first.

tacryt-socryp:
I'd love to get this in release/next-js so we can test ASAP

tacryt-socryp:
Are we planning another release today, or are you referring to the earlier one @matildepark ?

liam-fitzgerald:
posting a link comment crashes

matildepark:
Earlier. If we release today it will be hotfix based on the remaining p1s on Liam’s plate. 

— 
~haddef-sigwen
https://urbit.org

liam-fitzgerald (approved):
LGTM!

PR #4512, opened 2021-02-26
dbug: virtualize ames lane cueing

Fang-:
Lanes aren't strictly guaranteed to be jams, and we don't want to crash when
processing them.

### DIFF
--- a/pkg/arvo/app/dbug.hoon
+++ b/pkg/arvo/app/dbug.hoon
@@ -593,10 +593,10 @@
                 %&  (ship p.lane)
               ::
                   %|
-                ?~  l=((soft ,[=@tas =@if =@ud]) (cue p.lane))
-                  s+(scot %x p.lane)
-                =,  u.l
-                (tape "%{(trip tas)}, {(scow %if if)}, {(scow %ud ud)}")
+                %-  tape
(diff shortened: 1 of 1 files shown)


belisarius222:
As of the latest breach, a lane is now encoded as a 48-bit bitstring, so it will never cue.  To get the debug pane to show lanes, it should decompose the lane into a 32-bit IPv4 address and 16-bit port number.  I don't think we need to virtualize that, since the packet format now specifies it as 48 bits, so I don't think it could fail to parse as IP and port.

philipcmonk (approved):
> I don't think we need to virtualize that

Considering the number of lane-related issues we've run into where :dbug would have been really helpful it wasn't crashing, I think it makes sense to virtualize anyway

belisarius222:
Sure, that makes sense

—
~rovnys-ricfer
https://urbit.org

Fang-:
Alright, we assume the new lane format now, interpreting the lane as @belisarius222 described.

image

The decoding here definitionally doesn't crash, so we don't virtualize. But since we don't (can't?) sanity-check the resulting IP and port we get, we also always show the raw lane as hex.

philipcmonk (approved):

Issue #4424, opened 2021-02-12
ames: temporary %strange-current crash


philipcmonk:
Several ships exhibited this crash recently.  In each case, it only happened once.  Also, I verified afterward that no flows still had that `current` and none appeared to be stuck.  I expect this means the crash is triggered on an out-of-order acknowledgment, and resolved when the in-order ack arrives.

Fairly recently we allowed Ames to print stack traces in this sort of case, so it's likely this bug has existed for a while, possibly since New Ames.

```
ames: crashed on message ack
%exit
/sys/vane/ames/hoon:<[2.491 7].[2.491 9]>
[%strange-current current=11 message-num=10 fragment-num=35]
/sys/vane/ames/hoon:<[2.490 7].[2.491 9]>
/sys/vane/ames/hoon:<[2.489 5].[2.492 17]>
/sys/vane/ames/hoon:<[2.487 5].[2.492 17]>
/sys/vane/ames/hoon:<[2.486 5].[2.492 17]>
/sys/vane/ames/hoon:<[2.299 9].[2.299 15]>
/sys/vane/ames/hoon:<[2.296 5].[2.301 7]>
/sys/vane/ames/hoon:<[2.294 5].[2.301 7]>
/sys/vane/ames/hoon:<[2.083 47].[2.083 71]>
/sys/vane/ames/hoon:<[2.083 7].[2.140 9]>
/sys/vane/ames/hoon:<[2.082 7].[2.140 9]>
/sys/vane/ames/hoon:<[2.079 7].[2.140 9]>
/sys/vane/ames/hoon:<[2.076 7].[2.140 9]>
/sys/vane/ames/hoon:<[1.921 7].[1.921 69]>
/sys/vane/ames/hoon:<[1.918 7].[1.921 69]>
/sys/vane/ames/hoon:<[1.911 7].[1.921 69]>
/sys/vane/ames/hoon:<[1.909 7].[1.921 69]>
/sys/vane/ames/hoon:<[1.907 7].[1.921 69]>
/sys/vane/ames/hoon:<[1.904 7].[1.921 69]>
/sys/vane/ames/hoon:<[1.344 10].[1.344 62]>
/sys/vane/ames/hoon:<[1.344 5].[1.344 62]>
/sys/vane/ames/hoon:<[1.343 5].[1.344 62]>
/sys/vane/ames/hoon:<[1.314 5].[1.344 62]>
/sys/vane/ames/hoon:<[1.310 5].[1.344 62]>
[...]

PR #4515, opened 2021-02-27
zuse: fix ordered-map delete many

belisarius222:
The ordered map `+traverse` function failed to delete all items because it could skip a node under certain deletion conditions.  This fixes that, confirmed by a new test that failed with the old implementation.

This should fix the [ames: strange current issue](http://github.com/urbit/urbit/issues/4424).


philipcmonk (approved):
Tested on a live ship, and verified to fix its recurring strange-current crashes.

PR #4517, opened 2021-02-27
Add objectFit='cover' to various Landscape avatars to avoid aspect ratio squishing

0x70b1a5:
**PLEASE NOTE** I was unable to test this PR locally as I had issues cloning the repository (something with `con.c`, notable that I initially attempted to clone via Windows, fool that I am!).

Anyway, shout out to @matildepark for telling me where to change the things that needed to change. You may also wish to squash these commits as they were made via GitHub web (due to aforesaid git clone issue).

### DIFF
--- a/pkg/interface/src/views/landscape/components/Participants.tsx
+++ b/pkg/interface/src/views/landscape/components/Participants.tsx
@@ -10,6 +10,7 @@ import {
   Row,
   Text,
   Icon,
+  Image,
   Action,
   StatelessTextInput as Input
 } from '@tlon/indigo-react';
@@ -297,7 +298,7 @@ function Participant(props: {
 
   const avatar =
     contact?.avatar !== null && !hideAvatars ? (
-      <img 
+      <Image 
         src={contact.avatar} 
         height={32} 
         width={32} 
--- a/pkg/interface/src/views/landscape/components/Participants.tsx
+++ b/pkg/interface/src/views/landscape/components/Participants.tsx
@@ -297,7 +297,13 @@ function Participant(props: {
 
   const avatar =
     contact?.avatar !== null && !hideAvatars ? (
-      <img src={contact.avatar} height={32} width={32} className="dib" />
+      <img 
+        src={contact.avatar} 
+        height={32} 
+        width={32} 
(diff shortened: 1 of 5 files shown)


matildepark (changes requested):
This doesn't work — you'll need `style={{ objectFit: 'cover' }}`; also use `display="inline-block"` not dib class.

matildepark:
So yeah, in sum -

- the style change previously specified
- in one commit
- labelled something like `landscape: correct avatar aspect ratio`

You did this work off your fork's master, which might make this difficult, though.

0x70b1a5:
Thanks! I'll try booting up my Linux machine and working on this there, to catch anything else that crops up.

matildepark:
You'll want to squash this to one commit.

```
git rebase -i 6bcbbf8f1a4756c195a324efcf9515b6f288f700
```

Then change all the commits in the list to `squash`, but the first commit to `reword`. Reword it to `landscape: correct avatar aspect ratio` and then save, `git push -f`.

0x70b1a5:
Oh gotcha, I had misunderstood what "in one commit" meant. Apologies.

pkg/interface/src/views/landscape/components/Participants.tsx
```diff
@@ -297,7 +297,13 @@ function Participant(props: {
 
   const avatar =
     contact?.avatar !== null && !hideAvatars ? (
-      <img src={contact.avatar} height={32} width={32} className="dib" />
+      <img
```
matildepark, line 300: `display="inline-block"` actually won't work with the unstyled `img` element. What you'll want is to add an `Image` import from indigo-react (you can find the indigo elements we import at the top), then just change this to `<Image`. That's all. You can append it as a review comment — `landscape: address #4517 feedback`

matildepark (changes requested):

pkg/interface/src/views/landscape/components/Participants.tsx
0x70b1a5: Thank you for your patience.

matildepark (approved):

PR #4518, opened 2021-03-01
meta: adjust merge operations

jtobin:
Adds a 'merge-release' workflow that runs on any push to a release branch.  At present the only job merges the triggering branch to 'ops-tlon'.

Also renames the 'merge' workflow to 'merge-master' to distinguish it appropriately.


Fang- (approved):
Neat, I think this is what I wanted. Thank you!

Now to fix the CI failure on `ops-tlon` that was _already there_...

jtobin:
> I think this is what I wanted.

Yeah I think so.  For the benefit of any onlookers: the merge commit on the ops branch will trigger the CI build, so this workflow will check that the ops branch is always building against the latest release branches.

PR #4519, opened 2021-03-01
MetadataIcon: use box-shadow

matildepark:
Group avatars use inset box shadows instead of borders. This sort of has the effect of not showing any border at all. Is this on purpose? I don't know. @g-a-v-i-n Please confirm!

image

Fixes urbit/landscape#492.


g-a-v-i-n:
Ideally, they should not have borders. But it's nice to show a border when the icon background is white, so what's why I want to use shadows here.

tacryt-socryp (approved):

PR #4523, opened 2021-03-02
settings: add 'hide utility tiles'

matildepark:
Adds a setting within the CalmEngine to hide the top-level tiles separately from group tiles (so you can choose both, neither, one or the other, with and without unread counts).

Requested by @urcades, presuming design stamp.

Screen Shot 2021-03-02 at 12 58 00 AM
Screen Shot 2021-03-02 at 12 58 05 AM


urcades:
insta-stamp

tacryt-socryp (approved):

PR #4524, opened 2021-03-02
Update the home screen icon for iOS

sirburpalot:
The current home screen icon is a blank white image, which doesn't look good in my opinion. I have created a replacement icon, as shown in the screenshot below:

landscape_icon


tacryt-socryp:
@urcades ping

urcades (changes requested):
Urbit isn't Landscape, and we need to continue to reinforce this wherever possible.

The sig "logo" is representative of Urbit as a project — using it to represent Landscape would be like using the Apple logo for any one of their sub-apps, like Messenger.

Attaching home screen icons for iOS, Android, MacOS, and one svg for stuff like web, in a moment.

urcades:
Here are a set of icons I'd support folding into the product, to be used whenever the SPA is turned into a "home" icon on mobile platforms:

[landscape-icons.zip](https://github.com/urbit/urbit/files/6347431/landscape-icons.zip)

It looks like this:

Screen Shot 2021-04-20 at 9 19 04 PM

Screen Shot 2021-04-20 at 9 19 08 PM

It's an abstraction of our current Indigo icon representation of "Groups" — the meaning is derived from some older illustrations we had provided to refer to groups:

Screen Shot 2021-04-20 at 9 19 47 PM

Which is derived from even older visual languages that were created to establish a "visual logic" to describe the project:

Screen Shot 2021-04-20 at 9 20 42 PM

I feel pretty good about this direction, as a base representation of "Landscape" — another cool direction to take with this would be to utilize sigils.js to make these icons representative of one's ship, but we need a baseline, and this is it.

tacryt-socryp:
Closed as no progress has been made.

PR #4525, opened 2021-03-02
meta: publish interface packages when available

jtobin:
Adds a workflow that, on every push to master, will publish new versions of interface subpackages to the NPM registry if any are available (i.e., if version numbers have been incremented in any of the respective package.json files).

cc @tylershuster


matildepark (approved):
Kickass, thanks.

PR #4527, opened 2021-03-02
Change PNG aura from @t to @

juped:


Fang- (approved):
Similar to the change for `octs` some time ago, this is definitely the correct aura here. Thanks!

juped:
Just make sure future binary marks are `octs` - this mark won't corrupt PNGs that aren't already corrupt, but not every binary file is that convenient!

PR #4531, opened 2021-03-03
profile: display and interaction tweaks

jamesacklin:
This PR introduces several visual and interaction changes to the profile edit and view interfaces.

https://user-images.githubusercontent.com/748181/109738767-68e3c300-7b96-11eb-99e0-509272e93d6b.mp4

- Breaks the profile interface elements out into components with varying degrees of child-rendering and prop-passing in order to ensure the display is consistent between modes.
- Makes changes to the profile header actions to reflect the profile public/private status and let pilots submit* the form.
- Wraps the entire edit view in the Formix form in order to relocate the cover image field inside the cover itself.
- Moves the sigil hex* and avatar image controls up in the form and closer to the elements they modify.
- Moves the cover image control into the cover area itself.
- Introduces "Replace" and "Remove" actions for the cover.

`*` Introducing this as a draft PR for now, since I have a few technical questions.


pkg/interface/src/views/apps/profile/components/EditProfile.tsx
```diff
@@ -103,28 +144,73 @@ export function EditProfile(props: any): ReactElement {
...
-    </Formik>
-  </>
+        {({ setFieldValue }) => (
+          <Form width='100%' height='100%'>
+            <ProfileHeader>
+              <ProfileControls>
+                <Row>
+                  <button
+                    type='submit'
+                    style={{ appearance: 'none', background: 'transparent' }}
```
jamesacklin, line 154: @matildepark This empty, unstyled submit button with an Indigo Text element inside of it gets us what we want visually and interactively, but is probably bad form. Do you have suggestions on how to submit a Formik form with two submit buttons? I tried calling onSubmit directly but to no avail.

pkg/interface/src/views/apps/profile/components/EditProfile.tsx
```diff
@@ -103,28 +144,73 @@ export function EditProfile(props: any): ReactElement {
...
+                  contact={contact}
+                  s3={props.s3}
+                  setFieldValue={setFieldValue}
+                  handleHideCover={handleHideCover}
+                />
+              </ProfileImages>
+            </ProfileHeader>
+            <Row mb={3} pt={5} width='100%'>
+              <Col pr={2} width='25%'>
+                <ColorInput id='color' label='Sigil Color' />
```
jamesacklin, line 185: This throws an error in the console, complaining about a `value` prop being set without an `onChange` handler, but both are present in the `ColorInput` component itself. Has anyone else fought this fight? It doesn't block interaction and doesn't introduce a regression, but adds some unwanted noise in the console.

pkg/interface/src/views/apps/profile/components/EditProfile.tsx
matildepark: This is a confusing question, because [button elements submit forms by default](https://stackoverflow.com/questions/3314989/can-i-make-a-button-not-submit-a-form). Both buttons should!

matildepark:
image

The corners of the header borders are cut off — you'll want `overflow="hidden"` on the header container here.

pkg/interface/src/views/apps/profile/components/EditProfile.tsx
matildepark:
```
                  <Button
                    type='submit'
                    display='inline'
                    cursor='pointer' fontWeight='500' color='blue' pl='0' pr='0'
                    border='0'
                    style={{ appearance: 'none', background: 'transparent' }}
                  >
                    Save Edits
                  </Button>
```

image

matildepark:
You'll want to merge next-js back in too.

pkg/interface/src/views/apps/profile/components/EditProfile.tsx
jamesacklin: Thanks, wasn't sure what the policy was on overriding buttons vs using text with click handlers.

pkg/interface/src/views/apps/profile/components/EditProfile.tsx
jamesacklin: Thanks to @liam-fitzgerald

matildepark:
Avatar moves as you scroll hehe

image

matildepark:
Still has the cut off corners on the header, I'm thinking about how to remedy...

matildepark:
It seems like you might be able to fix header clipping if you put the header as the background image of its sibling div — this one:

image

and use `background-size: cover; background-position: center;`

jamesacklin:
OK, I will try that. Which browser / OS are you seeing this in? "Looks fine for me"

image

matildepark:
Safari™

matildepark (approved):
Had a stale build again for the header thing, very annoying, lgtm

PR #4533, opened 2021-03-03
tally: be safer wrt mangled metadata

Fang-:
Previously, if metadata-store said a graph existed, we'd unconditionally scry
for it. Now, we make sure the graph _actually_ exists, to avoid risking a crash.

I think `/app/sane` can catch/fix such cases, but it's not guaranteed to have run on a user's ship, so we need to do a sanity-check here too.

### DIFF
--- a/pkg/arvo/gen/tally.hoon
+++ b/pkg/arvo/gen/tally.hoon
@@ -61,14 +61,24 @@
   ::NOTE  we only count graphs for now
   ?.  &(=(%graph app-name.m) =(our creator.metadatum))  ~
   `[module.metadatum resource.m]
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
We allow you to have metadata for graphs that you have not joined yet. Not a bug that sane will fix, just a part of the specified behavior! This is a good fix for tally.

tacryt-socryp (approved):

matildepark:
Yeah, Landscape gets the graphs we *have locally* by subscribing to `/keys`.

Fang-:
Oh, I guess that makes sense. Thanks for clarifying!

PR #4535, opened 2021-03-04
VirtualScroller: prevent crash on iOS

liam-fitzgerald:
Fixes urbit/landscape#515


matildepark (approved):
Confirmed to remedy the issue on iOS.

tacryt-socryp (approved):

PR #4537, opened 2021-03-04
invites: don't pin to top

liam-fitzgerald:
- Removes pinning of invites to top
- Adds ability to hide join requests

### DIFF
--- a/pkg/arvo/app/group-view.hoon
+++ b/pkg/arvo/app/group-view.hoon
@@ -4,23 +4,30 @@
 |%
 ++  card  card:agent:gall
 ::
-+$   base-state
++$   base-state-0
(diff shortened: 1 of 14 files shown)


matildepark:
We can probably squeeze this into the la/release-2021-3-5 branch if @tacryt-socryp reviews and finds it innocent enough to test alongside the graph work.

tacryt-socryp:
Plz no

matildepark:
These are pretty innocent changes that we’d be testing on ropsten anyway.

matildepark:
Targetting next-userspace, we'll just test in the QA period.

tacryt-socryp (approved):
LGTM

PR #4539, opened 2021-03-04
LinkResource: fix changing channels

liam-fitzgerald:
Forcibly destroys and recreates a linkwindow when the channel changes. This is necessary to prevent the virtualscroller from acting upon stale state.

Fixes urbit/landscape#508


matildepark (approved):
lgtm

PR #4540, opened 2021-03-04
Participants: remove grid

liam-fitzgerald:
running into https://bugs.chromium.org/p/chromium/issues/detail?id=688640
on large groups

Fixes urbit/landscape#513


matildepark (approved):
Tested on Safari and Chrome before/after, confirm replication, code looks fine

PR #4541, opened 2021-03-04
Rename ++html to ++encoding

juped:
Zuse has a core named ++html implementing various encodings, which are largely unrelated to HTML.  Rectify this name by renaming the core to ++encoding, leaving a compatibility shim in place for userspace.

Anton (in 2016) suggests ++web-txt, but these arms are not really web or text related either.


zalberico:
@joemfb @mcevoypeter if this is something we want let me know.

joemfb:
This API should definitely be changed. I'm not sure when, or to what exactly.

PR #4545, opened 2021-03-04
landscape: avatars use borderRadius '1'

matildepark:
As per @urcades, we style avatars to match sigils' rounding.


tacryt-socryp:
Perhaps we should just have an Avatar component ☺️

tacryt-socryp (approved):

PR #4550, opened 2021-03-05
graph-threads: make %groupify and %restore compile

tacryt-socryp:
These threads had stopped compiling when we shipped the upgrade from `%metadata-hook` to `%metadata-push-hook`.


philipcmonk (approved):
LGTM

philipcmonk:
(will merge out-of-band)

PR #4554, opened 2021-03-05
clay: remove +bunt from marks

belisarius222:
- Removed the `+bunt` arm from both `$nave` and `$dais`.  This should also fix the bug where extra faces showed up in the result of `/*`'ing a file.
- Replaced Dojo's mark bunting implementation with `*vale:dais`.
- Removed one layer of vase from `%ce` scrys.
- Fixed assertion bug in `-build-nave` thread.
- Updated `/sys` tests.


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -649,7 +646,7 @@
         |=  diff=vase
         (slam (slap nav limb/%pact) (slop sam diff))
       ++  vale
-        |=  =noun
+        |:  noun=q:(slap nav !,(*hoon *vale))
```
philipcmonk, line 649: Clever

philipcmonk (approved):
Sorry I missed this earlier.  Looks fine, as long as everything in fact works (eg |merge).

tacryt-socryp:
Is this change still slated?

philipcmonk:
It is, I just want to do a final testing round, probably at the same time as the usual next-sys release testing.

belisarius222:
It's worth making sure this PR updates Clay properly before merging it -- it's been long enough since I wrote this that I don't have a clear memory of whether I tested upgrade properly, and there's no explicit code for it.  Even if it does work like this, I wonder if we'd be better off incrementing the Clay state version number; it did change, if only because the Lull types changed.

PR #4557, opened 2021-03-06
term: flex history container width

matildepark:
Try running `.` in webterm at half your screen width — the history container overflows off screen. This is because you need `min-width: 0` to flex child containers when they try to overflow. We apply that here.


Fang- (approved):
> This is because you need `max-width: 0` to flex child containers when they try to overflow.

The change here does `minWidth` but I assume that's correct then. Thanks!

matildepark:
Min, yeah. I'm sleepy

PR #4558, opened 2021-03-06
king: fix two space leaks leading to oom on ~def.

pilfer-pandex:
Before:
image
After:
image


jtobin (approved):
Heap / Haskell are both great.  Nice job finding that *nasty* IPC leak in particular. :+1: :+1:

vvisigoth:
nice!
—
~poldec-tonteg
http://urbit.org

PR #4559, opened 2021-03-06
profile: surface "message" action on non-owned profiles

jamesacklin:
"On the tin," indeed!

image

Fixes urbit/landscape#532


matildepark (changes requested):
Sig mismatch. The new DM route uses non-prepended patps, eg. `/dm/mex` but our `ship` prop here will provide `~mex`, which will result in a failed graph-create, spinning the creation screen forever.

urcades (approved):
Stamped wrt design

matildepark (approved):

PR #4561, opened 2021-03-07
top bar: dropdown styling tweaks

jamesacklin:
- Pads the profile dropdown menu items out a bit and adds "interactive" 500-weight type.
- Rearranges the items and renames "System Settings" to "System Preferences" as per the spec.
- Allows users to set their status from an inline form element and pressing ENTER.

Expectation:
image 2

Reality:
image

Fixes urbit/landscape#531


matildepark (changes requested):
You'll want it to set it either onBlur *or* on enter.

urcades (approved):
+1 @matildepark 's remark, but otherwise, this is good to go

pkg/interface/src/views/apps/profile/components/SetStatus.tsx
```diff
@@ -50,6 +50,9 @@ export function SetStatus(props: any) {
             value={_status}
             autocomplete='off'
             width='100%'
+            onBlur={() => {
```
jamesacklin, line 53: @liam-fitzgerald This _almost_ does what I want it to, but I need to hook into the `Dropdown` above it to set the status when the user clicks outside the dropdown (I open the dropdown, I set my status, I click outside, it is set). Any suggestions would be extremely welcome

liam-fitzgerald: In the `useEffect` that you use to set the status, you might consider 
```typescript
return smallControl ? editStatus : () => {}
```

jamesacklin:
Instead of reinventing the wheel, I decided to use the status control we already have in the profile overlay, so whatever behavior is there also exists here (blur, enter, etc). No visual change while data is at rest, added bonus of status-set-confirmation. 

G2G

matildepark (approved):

PR #4563, opened 2021-03-08
collections: virtualisation flicker fix

liam-fitzgerald:
- Fixes weird flicker in the virtualisation of flicker, due to the `renderer` not being referentially stable
- Documents the invariants that VirtualScroller needs from its parent components


matildepark (approved):
Does what it says

PR #4565, opened 2021-03-08
interface: grabbag of fixes

liam-fitzgerald:
Leap: fix double section

Fixes urbit/landscape#504

Participants: kick pending contacts

Fixes urbit/landscape#269

Profile: fix stale headers

Fixes urbit/landscape#419

StatusBar: fix dismiss

Fixes urbit/landscape#476

VirtualScroller: correct lazy loading with small chats

Fixes urbit/landscape#523

Comments: validate on submit only

Fixes urbit/landscape#530

### DIFF
--- a/pkg/interface/src/logic/reducers/settings-update.ts
+++ b/pkg/interface/src/logic/reducers/settings-update.ts
@@ -60,7 +60,7 @@ export default class SettingsStateZusettingsReducer{
   getAll(json: any, state: SettingsStateZus) {
     const data = _.get(json, 'all');
     if(data) {
-      _.merge(state, data);
+      _.mergeWith(state, data, (obj, src) => _.isArray(src) ? src : undefined)
     }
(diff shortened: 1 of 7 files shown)


matildepark:
Regarding 456 — how does this behave if you’re scrolled up and disconnect?

— 
~haddef-sigwen
https://urbit.org

liam-fitzgerald:
you lose the scroll pos, sadly. We can fix this when connection resuming lands in `@urbit/http-api`

matildepark:
I think that's actually worse UX than our "temporarily, silently dropping a message" — it means people with multiple tabs open or the like won't be able to read backlog like, at all. I'd punt the commit out and reapproach after http-api.

liam-fitzgerald:
commit dropped

matildepark (approved):
Rest of these read sanely.

PR #4566, opened 2021-03-08
Invites: accept already joined group

liam-fitzgerald:
Fixes urbit/landscape#503


tacryt-socryp:
Are you sure we don't want to "reject"? We may desire not to re-attempt any invite-related effects. Joining a group *should* be idempotent, but I'm unsure if it actually is.

liam-fitzgerald:
Joining a group is indeed idempotent, but changed anyway because it's the right thing to do

matildepark (approved):

PR #4567, opened 2021-03-09
tutorial: Chat copy edit to match design

jyng:


jyng:
https://github.com/urbit/landscape/issues/519

liam-fitzgerald (approved):

pkg/interface/src/logic/lib/tutorialModal.ts
```diff
@@ -107,7 +107,7 @@ export const progressDetails: Record<TutorialProgress, StepDetail> = {
   chat: {
     title: 'Chat',
     description:
-      'Chat channels are for messaging within your group. Direct Messages are also supported, and are accessible from the “DMs” tile on the homescreen',
+      'Chat channels are for messaging within your group. Direct Messages can be accessed from Messages in the top right,
```
matildepark, line 110: This broke the build on dev stream because it's an unterminated string. Patched in https://github.com/urbit/urbit/commit/8f8786ff2667cf2f802a67963f12ac100f0a8a8b, please be careful.

PR #4568, opened 2021-03-09
Tutorial: polish for release

liam-fitzgerald:
- Fixes broken profile step
- Hide caret on mobile

### DIFF
--- a/pkg/interface/src/views/apps/profile/components/Profile.tsx
+++ b/pkg/interface/src/views/apps/profile/components/Profile.tsx
@@ -161,9 +161,6 @@ export function Profile(props: any): ReactElement {
 
   const anchorRef = useRef<HTMLElement | null>(null);
 
-  console.log(ship);
-
-
   if (!props.ship) {
     return null;
   }
--- a/pkg/interface/src/views/apps/profile/components/Profile.tsx
+++ b/pkg/interface/src/views/apps/profile/components/Profile.tsx
@@ -29,6 +29,10 @@ export function ProfileImages(props: any): ReactElement {
   const { contact, hideCover, ship } = { ...props };
   const hexColor = contact?.color ? `#${uxToHex(contact.color)}` : '#000000';
 
+  const anchorRef = useRef<HTMLElement | null>(null)
+
+  useTutorialModal('profile', ship === `~${window.ship}`, anchorRef);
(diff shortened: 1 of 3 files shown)


jyng:
I seem to have found some new bugs.

### Tutorial is stuck if I had exited via the 'X'
Repro:
1) Start the tutorial
2) Exit at any step. 
3) Confirm `End Tutorial `
4) Start tutorial again through Leap. I get 'End Tutorial Now?' again.



### If I do a hard refresh(clear my cookies) and I get prompted the initial 'Welcome to Landscape..'
Ideally, we should be prompted this screen only on the user's first entry. Do you think that would be reasonably possible? Checking to see if it's the user's first time booting and starting the ship?

### Tutorial restores state of where it left off
This was in issue about tutorial stopping on profile. But @urcades notes a different issue. I wasn't able to repro due the first issue I listed here.
https://github.com/urbit/landscape/issues/520#issuecomment-790891867

matildepark:
> Ideally, we should be prompting this screen only on the user's first entry. Do you think that would be reasonably possible? Checking to see if it is the user's first time booting?

I mean, that's why we use `settings-store` here, it's meant to be once per boot. If you didn't wrap up the tutorial and we didn't mark it as 'seen' it would show this modal next time presumably.

liam-fitzgerald:
Yeah it'll only mark as done when you've either a) dismissed the tutorial or b) finished it

pkg/interface/src/views/apps/profile/components/Profile.tsx
```diff
@@ -160,43 +161,38 @@ export function Profile(props: any): ReactElement {
 
   const anchorRef = useRef<HTMLElement | null>(null);
 
-  useTutorialModal('profile', ship === `~${window.ship}`, anchorRef);
+  console.log(ship);
```
matildepark, line 164: *whispers* console.log...

liam-fitzgerald: sry!

jyng:
> Yeah it'll only mark as done when you've either a) dismissed the tutorial or b) finished it

Hm, interesting. I was running on incognito mode and it prompted up even when I had finished the tutorial previously.

PR #4569, opened 2021-03-09
Sigils: fix default background

liam-fitzgerald:
Fixes default background on sigils when the theme does not match the OS darkmode


matildepark (approved):
lgtm

PR #4570, opened 2021-03-09
gall: add $foot to kick sign

liam-fitzgerald:
Changes the type of %kick to include `p=(unit foot)`, where foot is a
union of symbols describing the reason for the kick.
%gone: Publisher breached
%mark: Mark validation or conversion failed
%fail: Subscriber failed to process fact

If there is no foot for the kick, then it's either a %clog-triggered
kick or manually by the publisher. We do not distinguish between these
as that would require a new version of the ames protocol, which is out
of scope for now, given that this work is largely motiviated by
versioned subscriptions.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:ec80a42446a1d80974f32bf87283435547441cb7ea3fcd340711df2ce6cbec81
-size 9146390
+oid sha256:a59df8c0a721f5daae01967db29a3d61b26a320bae2513d193c85a2fc1b917cc
+size 10135909
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1653,6 +1653,7 @@
                   byk=beak                              ::  load source
           ==  ==                                        ::
   +$  dude  term                                        ::  server identity
+  +$  foot  ?(%mark %gone %fail)                        ::  reason for kick  
   +$  gill  (pair ship term)                            ::  general contact
   +$  scar                                              ::  opaque duct
     $:  p=@ud                                           ::  bone sequence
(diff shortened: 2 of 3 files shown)


liam-fitzgerald:
I've uh realised that the mark kick will usually come over the network which will obviously strip it before going over ames. Which removes the entire motivation behind this PR. Closing as a result

joemfb:
I don't follow -- why isn't the mark mold crashing locally ("vale fail") sufficient information?

belisarius222:
You should be able to modify the protocol that Gall uses over Ames to include extra data with the `%kick`.  See the [`%x` case](https://github.com/urbit/urbit/blob/fb1da2639da7601405174818b9b870e988e56493/pkg/arvo/sys/vane/gall.hoon#L82) of the `$ames-response` datatype in Gall.

Just be careful to make sure upgrading the remote Gall protocol over the wire has acceptable behavior in all the possible intermediate states (e.g. one ship subscribes, the publisher acks, upgrades, then kicks).  Should be doable, though, because the requests (although not the responses) include a version number.

joemfb:
I don't think it's a good idea to update %gall's %ames messages until we fix the subscription state machine.

belisarius222:
That's a good point.

—
~rovnys-ricfer
https://urbit.org

philipcmonk:
Why? Aren't they orthogonal changes?

belisarius222:
They should be orthogonal, but right now that state machine is broken, so trying an update without fixing it might be risky -- I wonder if we should fix the state machine and add the extra info to kick as part of the same PR to minimize the amount of migration work we need to do.

—
~rovnys-ricfer
https://urbit.org

liam-fitzgerald:
My original intention was to use `%watch-as` and then use the foot to check if the conversion failed, however because the conversion is always done on the publishers side, the foot will always be blank because it comes over the network. If the changes are too wide-reaching for right now, I'm happy to let this lie, as I have designs that don't depend on this behaviour. However, I would really like this to become a part of gall at some point.

PR #4573, opened 2021-03-10
interface: p1 hotfixes

liam-fitzgerald:
LinkItem: safely access size

Fixes urbit/landscape#542

RemoteContent: use settings instead of local state

Fixes urbit/landscape#544


matildepark (approved):
These both read sanely and straightforwardly. If you’ve confirmed both original repro and that these fix, you can deploy

PR #4574, opened 2021-03-10
interface: zustand withState hotfix

tylershuster:
fixes crash on devnet


matildepark:
We did it directly to act as fast as possible. https://github.com/urbit/urbit/commit/ca682f7bcffc4e3c921935775fba38515eec36cf

PR #4576, opened 2021-03-10
hooks: versioned subscriptions

liam-fitzgerald:
A draft of versioned subscriptions.

There are four cases that we care about with versioning

#### Publisher ahead, compatible
If the publisher is ahead and the requested version is convertible to the publisher's current version, then publisher does mark conversion to the requested version and gives it on the subscription

#### Subscriber ahead, compatible
If the subscriber is ahead and the received version is convertible to the subscriber's current version, then the subscriber does a mark conversion to the requested version before proxying it to the store

##### Subscriber ahead, incompatible
If the subscriber is ahead and the received version is not convertible, then we suspend the connection and subscribe to the /versions path. We then resume the connection when the publisher gives us a version that is convertible

##### Publisher ahead, incompatible
If the publisher is ahead and the requested version is not convertible, then we suspend the connection and wait to learn about the version that the publisher is on. 

I'd note that I'm concerned about the publisher ahead & compatible case. We need to ensure that there's no possible loss of information in this conversion

TODO:

- [x] Proxied pokes should be converted
- [x] Restart subscription logic
- [x] A functioning +on-load
- [x] Groups
- [x] Graph
- [x] Metadata
- [x] Contacts

### DIFF
--- a/pkg/arvo/lib/pull-hook-virt.hoon
+++ b/pkg/arvo/lib/pull-hook-virt.hoon
@@ -29,10 +29,10 @@
   ``.^(* u.pax)
 ::
 ++  kick-mule
-  |=  [rid=resource =trap]
+  |=  [rid=resource trp=(trap *)]
   ^-  (unit (unit path))
   =/  res=toon
-    (mock [trap %9 2 %0 1] mule-scry)
+    (mock [trp %9 2 %0 1] mule-scry)
   =/  pax=(unit path)
     !<  (unit path) 
     :-  -:!>(*(unit path)) 
--- a/pkg/arvo/lib/pull-hook-virt.hoon
+++ b/pkg/arvo/lib/pull-hook-virt.hoon
@@ -1,0 +1,52 @@
+::  pull-hook-virt: virtualisation for pull-hook
+/-  *resource
+|_  =bowl:gall
+++  mule-scry
+  |=  [ref=* raw=*]
+  =/  pax=(unit path)
+    ((soft path) raw)
+  ?~  pax  ~
+  ?.  ?=([@ @ @ @ *] u.pax)  ~
+  =/  ship
+    (slaw %p i.t.u.pax)
+  =/  ved
+    (slay i.t.t.t.u.pax)
+  =/  dat
+    ?~  ved  now.bowl
+    =/  cas=(unit case)
+      ((soft case) p.u.ved)
+    ?~  cas  now.bowl
+    ?:  ?=(%da -.u.cas)
+      p.u.cas
+    now.bowl
+  ::  catch bad gall scries early
+  ?:  ?&  =((end 3 i.u.pax) %g)
+          ?|  !=(`our.bowl ship)
+              !=(dat now.bowl)
+          ==
+      ==
+    ~
+  ``.^(* u.pax)
+::
+++  kick-mule
+  |=  [rid=resource =trap]
+  ^-  (unit (unit path))
+  =/  res=toon
+    (mock [trap %9 2 %0 1] mule-scry)
+  =/  pax=(unit path)
+    !<  (unit path) 
+    :-  -:!>(*(unit path)) 
+    ?:(?=(%0 -.res) p.res ~)
+  ?:  !?=(%0 -.res)
+    =/  =tang
+      :+  leaf+"failed kick handler, please report" 
+        leaf+"{<rid>} in {(trip dap.bowl)}"
+      ?:  ?=(%2 -.res)
+        p.res
+      ?>  ?=(%1 -.res)
+      =/  maybe-path=(unit path)  ((soft path) p.res)
+      ?~  maybe-path  ~
+      [(smyt u.maybe-path) ~]
+    ((slog tang) ~)
+  `pax
+--
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -423,7 +423,7 @@
     ::
     ++  tr-fact
       |=  =cage
-      ?:  ?=(%hook-meta-update p.cage)
+      ?:  ?=(%version p.cage)
         (tr-suspend-sub-ver !<(@ud q.cage))
       ?>  (is-root:ver p.cage)
       =/  fact-ver=@ud
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -317,12 +317,14 @@
           (take-version:hc src.bowl sign)
         =.  pull-hook  hook
         [cards this]
-      ?.  ?=([%pull %resource *] t.t.wire)
+      ?.  ?=([%pull ?(%unver-resource %resource) *] t.t.wire)
         (on-agent:def wire sign)
       =/  rid=resource
         (de-path:resource t.t.t.t.wire)
+      =/  versioned=?
+        ?=(%resource i.t.t.t.wire)
       =^  [cards=(list card) hook=_pull-hook]  state
-        tr-abet:(tr-sign:(tr-abed:track-engine:hc rid) sign)
+        tr-abet:(tr-sign:(tr-abed:track-engine:hc rid) sign versioned)
       =.  pull-hook  hook
       [cards this]
     ::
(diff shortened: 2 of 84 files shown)


liam-fitzgerald:
This is basically finished, just needs thorough testing. Review would be nice

Fang- (commented):
Neat! Just some drive-by golfing below.

pkg/arvo/lib/versioning.hoon
```diff
@@ -0,0 +1,48 @@
+/+  agentio
+|_  [=bowl:gall root=mark version=@ud]
++*  io  ~(. agentio bowl)
+++  is-root
+  |=  =mark
+  =/  tape-mark
+    (trip mark)
+  =/  tape-version
+    (trip root)
+  =((scag (lent tape-version) tape-mark) tape-version)
```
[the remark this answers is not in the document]
joemfb: we need an atomic +find, these allocations kill me ...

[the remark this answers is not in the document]
Fang-: Actually, forget `+find`, for this specific case you can just `=(root (end 3^(met 3 root) mark))`.

This has the same problem all of these snippets have though, that a root of `%mark` will match both `%mark-3` and `%mark-other-2` or similar. Feeling like you might want to roll this up into version parsing, so you'd just say `;~(pfix (jest root) hep dum:ag)`, and know it's invalid (either unparseable or not the expected root) if that fails.

pkg/arvo/lib/versioning.hoon
```diff
@@ -0,0 +1,52 @@
...
+  =/  ver
+    (read-version mark)
+  &((gte ver min) (lte ver version))
+::
+++  convert-to
+  |=  =cage
+  ^-  vase
+  ?:  =(p.cage current-version)
+    q.cage
+  ((tube-to p.cage) q.cage)
```
tacryt-socryp, line 35: Are we planning on building out any sort of caching utility for this? (Honestly, starting to think we should build a "ford-hook" (forgive the horrible name) that lets one start a subscription to a file or tube with a poke to the app and optionally subscribe to the results. That way all the caching logic for mark conversions can live in one app, and all of the state for "cached marks" can exist together.

Fang-: As of #4491 you can achieve the same by importing the relevant marks into your app I believe. You can't get _as_ dynamic with it, but I don't think you want/need to.

tacryt-socryp: I think we'd like to retain our current dynamic behavior with `%graph-validator` marks, at least. For normal "update versions" perhaps it would make the most sense to do it statically.

tacryt-socryp (changes requested):
Code doesn't compile upon boot: 
Screen Shot 2021-03-19 at 11 52 47 AM

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -295,72 +312,310 @@
...
+      ^-  [(list card:agent:gall) agent:gall]
+      =^  cards  pull-hook
+        (on-fail:og term tang)
+      [cards this]
+    ::
+    ++  on-peek   
+      |=  =path
+      ^-  (unit (unit cage))
+      ?:  =(/x/dbug/state path)
+        ``noun+(slop !>(state(inner-state *vase)) on-save:og)
```
tacryt-socryp, line 356: Why is inner-state being replaced for this output?

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -295,72 +312,310 @@
...
+      |=  ap=_^?(|.(*(quip card _pull-hook)))
+      =^  caz  pull-hook
+        (ap)
+      (tr-emis caz)
+    ::  +|  %sign: sign handling
+    ::
+    ::
+    ++  tr-sign
+      |=  [=sign:agent:gall versioned=?]
+      ?+   -.sign  !!
```
tacryt-socryp, line 405: Seems best to put tr-wack and tr-fact into a |^ here, but since tr-kick is used elsewhere seems best to keep that arm where it is.

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -295,72 +312,310 @@
...
+      (tr-watch u.u.pax)
+    ::
+    ++  tr-fact
+      |=  =cage
+      ?:  ?=(%version p.cage)
+        =/  req-ver=@ud
+          !<(@ud q.cage)
+        ?:  (lth req-ver min-version.config)
+          (tr-suspend-pub-ver min-version.config)
+        (tr-suspend-sub-ver req-ver)
```
tacryt-socryp, line 438: Why do we suspend on sub-ver here? Doesn't that mean we suspend in all cases upon receiving a version fact?

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -295,72 +312,310 @@
...
+      [(watch-version who)^~ pull-hook]
+      ::
+        %fact
+      ?.  =(%version p.cage.sign)
+        [~^pull-hook state]
+      =+  !<(version=@ud q.cage.sign)
+      =/  tracks=(list [rid=resource =track])
+        ~(tap by tracking)
+      =|  cards=(list card)
+      =|  leave=_&
```
tacryt-socryp, line 581: Doesn't `?` already bunt to `%.y`???

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -295,72 +312,310 @@
...
+  ::
+  ++  leave-version
+    |=  =ship
+    (~(leave pass version-wir) [ship push-hook-name.config])
+  ::
   ++  poke-sane
     ^-  (quip card:agent:gall _state)
     =/  cards
-      restart-subscriptions
+      ~ :: restart-subscriptions
```
tacryt-socryp, line 614: perhaps a TODO?

pkg/arvo/lib/pull-hook.hoon
```diff
@@ -295,72 +312,310 @@
...
     ~?  >  ?=(^ cards)
       "Fixed subscriptions in {<dap.bowl>}"
     :_  state
-    restart-subscriptions
+    ~  :: restart-subscriptions
```
tacryt-socryp, line 618: also perhaps a TODO?

pkg/arvo/lib/push-hook.hoon
```diff
@@ -378,26 +412,51 @@
     [%pass wire %agent [our.bowl store-name.config] %watch store-path.config]
   ::
   ++  push-updates
-    |=  =vase
+    |=  =cage
     ^-  (list card:agent:gall)
-    =/  rids=(list resource)  (resource-for-update vase)
+    =/  rids=(list resource)  (resource-for-update q.cage)
     =|  cards=(list card:agent:gall)
     |-
```
tacryt-socryp, line 419: I believe I changed this out to be a `+murn` in my PR, and that still seems prudent, given that it's jetted, whereas manual loops are not, and are more bespoke than a simple `+turn` with a `+zing` at the end or a `+murn`.

pkg/arvo/lib/push-hook.hoon
```diff
@@ -378,26 +412,51 @@
...
+      cards  (welp cards caz)
     ==
   ::
+  ++  ver-from-path
+    |=  =path
+    =/  extra=^path
+      (slag 5 path)
+    ?>  ?=(^ extra)
+    (slav %ud i.extra)
+
```
tacryt-socryp, line 458: blank line

pkg/arvo/lib/pull-hook.hoon
liam-fitzgerald: Yes, but I find this behaviour confusing, and prefer to make it explicit

pkg/arvo/lib/pull-hook.hoon
liam-fitzgerald: Because the prettyprinter will print `inner-state` as a vase otherwise, so we bunt inner-state and slop on the actual inner state.

pkg/arvo/lib/pull-hook-virt.hoon
```diff
@@ -0,0 +1,52 @@
+::  pull-hook-virt: virtualisation for pull-hook
+/-  *resource
+|_  =bowl:gall
+++  mule-scry
+  |=  [ref=* raw=*]
```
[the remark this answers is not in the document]
liam-fitzgerald: Just added unit tests

pkg/arvo/lib/pull-hook.hoon
liam-fitzgerald: Yes, because a `version` fact is only sent if the publisher cannot convert it's facts to the requested version. It's always followed by a kick.

tacryt-socryp (approved):

matildepark:
We're on hold for a Ropsten merge because of this appetiser which should not have been merged with a visible CI failure on all the marks — this doesn't compile, and should not have been stamped, either.

Standby for staging session after the midweek...

PR #4578, opened 2021-03-10
interface: reduce group-view messages

tylershuster:
should fix https://github.com/urbit/landscape/issues/554

...might help to actually reduce the message


matildepark:
It'll still fail to accept the invite because of a non-existent `props.groups` on L47 of pkg/interface/src/views/components/invite/index — can you amend that in this PR too?

tylershuster:
I was able to confirm that this fixes the problem for the group to which you invited me.

matildepark (approved):

PR #4579, opened 2021-03-10
settings: sidebar same width as groups

jamesacklin:
Adjusts the System Preferences layout to be the same (CSS grid-based) as Groups and Messages.

The sidebar in both instances is flexible. The proportions match, therefore so do the pixel measurements at various window widths.

Frame 1 (1)

An easy way to test is to visit the System Preferences, Leap to a Group, then use browser keyboard shortcuts to quickly flip back and forth between pages.

Fixes urbit/landscape#541

### DIFF
--- a/pkg/interface/src/views/apps/settings/components/lib/BackButton.tsx
+++ b/pkg/interface/src/views/apps/settings/components/lib/BackButton.tsx
@@ -4,8 +4,16 @@ import { Text } from '@tlon/indigo-react';
 
 export function BackButton(props: {}) {
   return (
-    <Link to="/~settings">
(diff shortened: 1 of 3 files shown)


urcades (approved):
I meant to write what I wrote here:

https://github.com/urbit/landscape/issues/541#issuecomment-796189061

In this PR, but the sentiment still holds — it's looking great, but if you have a better sense/thought around how to modify the width, go right ahead!

matildepark (changes requested):
This seems not desired

image

jamesacklin:
> This seems not desired

Fixed by yanking the bonus Row container. 

image

Also appropriately padded the Back button on mobile.

image

matildepark (approved):

PR #4582, opened 2021-03-11
interface: changed ProfileOverlay to functional component

tylershuster:


pkg/interface/src/views/components/ProfileOverlay.tsx
```diff
@@ -14,7 +14,7 @@ import {
   Icon
 } from '@tlon/indigo-react';
 import RichText from './RichText';
-import { withLocalState } from '~/logic/state/local';
+import useLocalState from '~/logic/state/local';
```
matildepark, line 17: shouldn't we be using settings state here? this isn't in localstorage anymore

tylershuster: image

uh should we

matildepark: `state.dark` is changed locally using an eventlistener, so it makes sense in Author. same with toggleOmnibox. but for ViewProfile and this ProfileOverlay it seems worth changing to use `settings.calm` selector stuff.

tylershuster: Ok I changed this component. Talking with @liam-fitzgerald about the other stuff

matildepark (approved):
Scope of this PR works, looks fine

PR #4583, opened 2021-03-11
interface: added tool to track unneeded renders

tylershuster:
As per a discussion with @liam-fitzgerald a while back, adds `why-did-you-render` during development to check for unneeded renders. Seems to be pretty clean already, but it looks like it doesn't track functional components by default (only PureComponents), but might aid in future debugging.

Also removes some unnecessary `console.log`s

### DIFF
--- a/pkg/interface/package.json
+++ b/pkg/interface/package.json
@@ -71,6 +71,7 @@
     "@types/yup": "^0.29.11",
     "@typescript-eslint/eslint-plugin": "^4.15.0",
     "@urbit/eslint-config": "file:../npm/eslint-config",
+    "@welldone-software/why-did-you-render": "^6.1.0",
     "babel-eslint": "^10.1.0",
     "babel-loader": "^8.2.2",
     "babel-plugin-lodash": "^3.3.4",
--- a/pkg/interface/config/webpack.dev.js
+++ b/pkg/interface/config/webpack.dev.js
@@ -94,7 +94,11 @@ module.exports = {
         use: {
           loader: 'babel-loader',
           options: {
-            presets: ['@babel/preset-env', '@babel/typescript', '@babel/preset-react'],
+            presets: ['@babel/preset-env', '@babel/typescript', ['@babel/preset-react', {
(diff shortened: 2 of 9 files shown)


matildepark:
Seems innocent enough, just wondering what this actually looks like in development now.

tylershuster:
Give it a shot. Pretty clean. The clock apparently rerenders once due to a strictly equal object. Should be more useful by doing:
```jsx
const FunctionalComponent = (props) => <div></div>;
FunctionalComponent.whyDidYouRender = true;
export default FunctionalComponent;
```

matildepark:
I have literally no idea how this changes the experience of developing or where to find this information. I boot up dev environment and it's just Landscape.

tylershuster:
It won't change visually. Open the console. There will be one exemplary log message on Launch.

Add the above code to a Functional Component which you believe is rerendering unnecessarily.

matildepark:
image

?

tylershuster:
show the clock

pkg/interface/package.json
```diff
@@ -13,6 +13,7 @@
     "@tlon/indigo-react": "^1.2.19",
     "@tlon/sigil-js": "^1.4.3",
     "@urbit/api": "file:../npm/api",
+    "@welldone-software/why-did-you-render": "^6.1.0",
```
matildepark, line 16: This should specifically be a dev dependency. This sort of utility seems like it shouldn't be in the bundle at all, in the same way I don't bundle VSCode with Landscape, but I don't think React has such a concept, so that's moot.

matildepark (commented):
Fine to add, given the change here — though I note we don't seem to use pure components, I never see this in console in Chat, even though ChatMessage ostensibly *is* one? Maybe it's a perfect render.

tylershuster:
Ok, I updated the dev dependency. I added `LinkResource.whyDidYouRender = true` and confirmed that you need to add that to functional components, and also that there are some other rerenders.

matildepark (approved):

PR #4584, opened 2021-03-11
push-hook: transform all messages, even the host's

tacryt-socryp:
Fixes https://github.com/urbit/landscape/issues/270 for good, by ensuring that *all messages* passed through the hooks are transformed and applied in the order they're received.

### DIFF
--- a/pkg/arvo/lib/push-hook.hoon
+++ b/pkg/arvo/lib/push-hook.hoon
@@ -396,20 +396,38 @@
   ++  forward-update
     |=  vas=vase
     ^-  (list card:agent:gall)
-    %+  murn  (resource-for-update vas)
-    |=  rid=resource
-    ^-  (unit card:agent:gall)
-    =/  =path  resource+(en-path:resource rid)
+    =-  lis
+    %+  roll  (resource-for-update vas)
+    |=  [rid=resource [lis=(list card:agent:gall) tf-vas=(unit vase)]]
+    ^-  [(list card:agent:gall) (unit vase)]
+    =/  =path
+      resource+(en-path:resource rid)
     =/  =wire  (make-wire path)
     =*  ship   entity.rid
-    =/  [=dock vax=(unit vase)]
-      ?.  =(our.bowl entity.rid)
-        :_  `vas
-        [ship dap.bowl]
-      :-  [ship store-name.config]
+    =.  tf-vas
+      ?.  =(our.bowl ship)
+        ::  do not transform before forwarding
+        ::
+        `vas
+      ::  use cached transform
+      ::
+      ?^  tf-vas  tf-vas
+      ::  transform before poking store
+      ::
       (transform-proxy-update:og vas)
-    ?~  vax  ~
-    `[%pass wire %agent dock %poke update-mark.config u.vax]
+    ~|  "forwarding failed during transform. mark: {<p.vas>} resource: {<rid>}"
+    ?>  ?=(^ tf-vas)
+    =/  =dock
+      :-  ship
+      ?.  =(our.bowl ship)
+        ::  forward to host
+        ::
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (commented):
hm, this functionality is also used to proxy our own pokes to foreign ships? How are we handling this now?

tacryt-socryp:
Ah, you're right. I'll make the necessary changes to preserve the current forwarding behavior for sending things to foreign push-hooks.

tacryt-socryp:
@liam-fitzgerald Changed the implementation, am going to test on `~tacryt` now. The logic is really straightforward and handles the "send to foreign" case properly now.

pkg/arvo/lib/push-hook.hoon
```diff
@@ -380,44 +380,36 @@
   ++  push-updates
     |=  =vase
     ^-  (list card:agent:gall)
-    =/  rids=(list resource)  (resource-for-update vase)
-    =|  cards=(list card:agent:gall)
-    |-
-    ?~  rids  cards
-    =/  prefix=path
-      resource+(en-path:resource i.rids)
+    %+  murn  (resource-for-update vase)
```
liam-fitzgerald, line 383: Isn't this missing an `:og`

pkg/arvo/lib/push-hook.hoon
```diff
@@ -380,44 +380,36 @@
...
+    |=  rid=resource
+    ^-  (unit card:agent:gall)
+    =/  prefix=path  resource+(en-path:resource rid)
     =/  paths=(list path)
       %~  tap  in
       %-  silt
       %+  turn
```
liam-fitzgerald, line 390: should be wide form

pkg/arvo/lib/push-hook.hoon
```diff
@@ -380,44 +380,36 @@
...
+    ^-  (unit card:agent:gall)
+    =/  =path  resource+(en-path:resource rid)
+    =/  =wire  (make-wire path)
+    =*  ship   entity.rid
+    =/  [=dock vax=(unit vase)]
+      ?.  =(our.bowl entity.rid)
+        :_  `vas
+        [ship dap.bowl]
+      :-  [ship store-name.config]
+      (transform-proxy-update:og vas)
```
liam-fitzgerald, line 410: This should be done outside of the iteration, for performance

pkg/arvo/lib/push-hook.hoon
```diff
@@ -380,44 +380,36 @@
...
+    ?~  vax  ~
```
liam-fitzgerald, line 411: I think this poke should nack if the transform fails.

liam-fitzgerald (changes requested):
Looks good overall. Have highlighted some changes though, think they should go in before merge

pkg/arvo/lib/push-hook.hoon
tacryt-socryp: No, because we're calling the arm in the hook core that wraps `resource-for-update:og`.

pkg/arvo/lib/push-hook.hoon
tacryt-socryp: Yes, that seems proper.

pkg/arvo/lib/push-hook.hoon
tacryt-socryp: Done, using `+roll`

tacryt-socryp:
@liam-fitzgerald This is ready to go, review responded to. Thanks for being thorough!

matildepark:
Switched target to next-userspace to consolidate the testing here, especially since this is a bit of an edge case.

liam-fitzgerald (approved):

PR #4588, opened 2021-03-11
interface: small grab-bag

jamesacklin:
- Correctly references `contacts` from the Zustand store in the header dropdown status-setter (the old one came from app root, which never refreshed and would result in out-of-sync statuses).
- Correctly aligns the "Save Profile" button to the text baseline of the other profile actions.

image

### DIFF
--- a/pkg/interface/src/views/components/StatusBar.js
+++ b/pkg/interface/src/views/components/StatusBar.js
@@ -9,6 +9,7 @@ import {
   Button,
   BaseImage
 } from '@tlon/indigo-react';
+
 import ReconnectButton from './ReconnectButton';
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/components/StatusBar.js
```diff
@@ -61,6 +63,8 @@ const StatusBar = (props) => {
   const floatLeap =
     leapHighlight && window.matchMedia('(max-width: 550px)').matches;
 
+  const contact = useContactState((state) => state.contacts[`~${ship}`]);
```
matildepark, line 66: Seems like this should be at very top and replace `ourContact` variable?

matildepark (approved):

PR #4589, opened 2021-03-12
interface: fixes base state typo and improves notification initial st…

tylershuster:
@liam-fitzgerald


liam-fitzgerald (approved):
We paired over this, approved

PR #4590, opened 2021-03-12
ChatWindow: fix unread mark and gaps in scrollback

liam-fitzgerald:
- Instead of marking graphs as loaded when we get a empty response from /node-siblings, mark them as read when the response is not as big as it should be
- Only calculate unread index when we've loaded all the way to the bottom edge

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
@@ -60,9 +60,6 @@ class ChatWindow extends Component<
   private virtualList: VirtualScroller | null;
   private unreadMarkerRef: React.RefObject<HTMLDivElement>;
   private prevSize = 0;
-  private loadedNewest = false;
-  private loadedOldest = false;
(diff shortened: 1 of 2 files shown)


matildepark:
What case does this fix? Wondering how to repro and sanely test this against what we have.

liam-fitzgerald:
If the chat has not been properly loaded, but we've received a handful of updates it will be incorrectly marked as loaded when it shouldn't be.

matildepark:
So, I got into a situation where I send a pending, scrolled up, then had an infinite spinner at the bottom that cleared up when another message arrived. But I can't even firmly repro that now.

matildepark:
On devnet if I add a message to a chat and in another tab load the graph for the first time, after that update already arrived, it does in fact grab the backlog. I'm hesitant to just stamp this without seeing a firm before/after I can confirm remedy for, since the behaviour here is so complex.

matildepark (approved):
I suspect this is the remedy for https://github.com/urbit/landscape/issues/536, so merging ...

PR #4592, opened 2021-03-12
interface: better error boundary reporting tools and formatting

tylershuster:
This should provide slightly better stack traces, plus it updates the "submit an issue" link to link to Landscape with the stack trace and title already filled in.

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -3989,6 +3989,14 @@
         "prr": "~1.0.1"
       }
     },
+    "error-stack-parser": {
+      "version": "2.0.6",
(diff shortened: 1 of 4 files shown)


matildepark:
It seems to work fine enough, but why is its unpacked size 1.46mb? It seems potentially heavier than lodash.

tylershuster:
That seems...wrong:

image

matildepark:
Just going off npm — 

https://www.npmjs.com/package/stacktrace-js
https://www.npmjs.com/package/lodash

tylershuster:
The `build:prod` version isn't significantly bigger than before, not sure what NPM is on about

matildepark:
Oh, this actually didn't hit next-js. Well, fine

matildepark (approved):

PR #4593, opened 2021-03-12
graph-update: sort indices correctly

liam-fitzgerald:
Correctly sorts indices (smallest to largest)


tacryt-socryp (approved):
Sorting function is correct. Why is the logging still here? No repro?

matildepark:
Informally can attest it seems to remedy a reproduction of this stack trace - https://github.com/urbit/landscape/issues/555#issue-828163831 - that you get when making first comment on a notebook, but that the others are still awaiting a firm repro

PR #4594, opened 2021-03-12
interface: added debug menu

tylershuster:
Watch this video:

https://user-images.githubusercontent.com/3999320/110891389-be128980-82a6-11eb-9cfa-ec361cb46c56.mp4

In the absence of an actual "Debug" menu item spec from design, this can be invoked by typing `~` on your keyboard while looking at the settings screen.

The state can be filtered using lodash's `_.get(object, path, default)` syntax, and the stores will update as data changes. Each time the filter correctly matches a part of the store, that result is logged in the console.

Useful for development and also for working through issues with Landscape with people experiencing issues. You could tell someone to go to this page and then enter `unreads.graph[/ship/some-long-string-here][/02384038438543]` to get a consistent track.

### DIFF
--- a/pkg/interface/src/views/apps/settings/components/lib/CalmPref.tsx
+++ b/pkg/interface/src/views/apps/settings/components/lib/CalmPref.tsx
@@ -80,7 +80,7 @@ export function CalmPrefs(props: {
       <Form>
         <BackButton/>
         <Col borderBottom="1" borderBottomColor="washedGray" p="5" pt="4" gapY="5">
-            <Col gapY="1" mt="0">
+          <Col gapY="1" mt="0">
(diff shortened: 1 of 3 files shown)


urcades:
>  this can be invoked by typing ~ on your keyboard while looking at the settings screen.

YES!

That said, was happy to draft up a minimal set of affordances for this:

I imagine accessing the menu would live in the Developer section, as an explicit button:

Screen Shot 2021-03-12 at 12 30 23 AM

The resulting page would essentially be what you've already built:

Screen Shot 2021-03-12 at 12 37 32 AM

matildepark:
Very cool

matildepark (approved):
Small, useful

PR #4595, opened 2021-03-12
tutorial: update for zustand

liam-fitzgerald:
Fixes urbit/landscape#549


matildepark (approved):
Works for everyone but me it seems? I think I got my tutorial join in a bad state while I was developing on it

PR #4596, opened 2021-03-12
graph-store signatures: turn them on and correctly validate them

tacryt-socryp:
As per @matildepark, waiting to put this in until closer to end of sprint. Can confirm that this works. I tested it on `~bolbex-fogdys` and `~tacryt-socryp` on livenet.


liam-fitzgerald (approved):
LGTM!

PR #4597, opened 2021-03-12
chat: 24px sigils

jamesacklin:
Resizes chat sigils and avatars to 24px square and slightly respaces some elements to keep everything in line.

- Keeps sigils themselves to 16px square, but pads enough around them to appear as 24px boxes. This is to keep them aligned to the author's name.
- Message author name type size is now uniform to the message content.
- Sigil, author, and timestamp are padded over to line up with the channel title in the header.
- Subsequent (non-sigil'd) messages are padded to align with author name.

image

image

Keeping as a draft right now to iterate with design.

Fixes urbit/landscape#460

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
@@ -138,14 +138,24 @@ class ChatInput extends Component<ChatInputProps, ChatInputState> {
           display='inline-block'
         />
       ) : (
-        <Sigil
-          ship={window.ship}
-          size={24}
-          color={`#${color}`}
-          classes={sigilClass}
-          icon
-          padding={2}
-        />
+        <Box
+          width={24}
+          height={24}
+          display='flex'
+          justifyContent='center'
+          alignItems='center'
+          backgroundColor={`#${color}`}
+          borderRadius={1}
+        >
+          <Sigil
+            ship={window.ship}
+            size={16}
+            color={`#${color}`}
+            classes={sigilClass}
+            icon
+            padding={2}
+          />
+        </Box>
       );
 
     return (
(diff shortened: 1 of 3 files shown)


pkg/interface/src/views/apps/chat/components/ChatInput.tsx
```diff
@@ -158,7 +170,7 @@ class ChatInput extends Component<ChatInputProps, ChatInputState> {
         className='cf'
         zIndex={0}
       >
-        <Row p='2' alignItems='center'>
+        <Row p='12px 8px 12px 12px' alignItems='center'>
```
matildepark, line 173: isnt this just '3 2 3 3'? we should opt for indigo values before hard coding

jamesacklin: Unfortunately not -- https://github.com/urbit/indigo-react/blob/1b93c5e502e28fa9e459f5edcb2bd428eeb82585/src/themes/light.ts#L150

matildepark: So our mocks are divergent from Indigo's spacing? Seems weird

urcades (approved):
I Truly Love This — We should "Ship It" asap

matildepark (approved):
The code is straightforward and it doesn't break other renders of sigils, works on mobile, works on desktop. Seems fine to me

pkg/interface/src/views/apps/chat/components/ChatInput.tsx
urcades: The mocks are derived from gav's actual indigo chat components, with the exception of the left-hand margin, which is a bit larger for ensuring highlighting will work sanely when that ships, and that timestamps fit in without being too squashed

matildepark: The issue here is that 12px doesn't appear to be an indigo spacing value

PR #4599, opened 2021-03-12
zuse: fix time fix regression

### LINKED ISSUE #4598: unm:chrono:userlib and from-unix-ms:chrono:userlib are sometimes not invertible
```
(unm:chrono:userlib (from-unix-ms:chrono:userlib 1.615.573.389.789))
1.615.573.389.788
```

```
(unm:chrono:userlib (from-unix-ms:chrono:userlib 1.615.573.389.000))
1.615.573.389.000
``` 

Seems to have been introduced here, but his logic seems correct at a glance. Very confused.

https://github.com/urbit/urbit/pull/4474

@joemfb @philipcmonk @belisarius222

mrdomino:
Fixes #4598.

#4474 made the JSON time conversion no longer invertible, which caused
problems for chat, which uses message timestamp in milliseconds as a key
— so chat would send a message with ms timestamp x, it would get
encoded as @da x, but then when it went back through the conversion to
milliseconds, it would often (not always) get encoded as x-1.

I still do not fully understand why this is — and why it doesn't seem
to be a problem with seconds based on cursory testing — but integer
multiplication and division generally do not invert. And adding a half a
millisecond to the input date before converting it resolves the issue
and makes the functions invertible.

I added a regression test, so hopefully the next courageous adventurer
who winds up here after wondering why `+unm` looks funny will have a
safeguard against some of the mistakes I made.

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -5535,7 +5535,8 @@
     ::                                                  ::  ++unm:chrono:userlib
     ++  unm                                             ::  Urbit to Unix ms
       |=  a=@da
-      (div (mul (sub a ~1970.1.1) 1.000) ~s1)
+      =-  (div (mul - 1.000) ~s1)
+      (sub (add a (div ~s1 2.000)) ~1970.1.1)
     ::                                                  ::  ++unt:chrono:userlib
     ++  unt                                             ::  Urbit to Unix time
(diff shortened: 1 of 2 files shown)


joemfb (approved):
LGTM

mrdomino:
Do we want this on master or next-sys?

joemfb:
`release/next-sys`

PR #4600, opened 2021-03-12
gen: delete |wash-gall

Fang-:
Straggler from days long gone. The functionality this would trigger no longer exists.


belisarius222 (approved):
LGTM

PR #4602, opened 2021-03-12
landscape profiles: tiny tweaks to visual elements

urcades:
Made a few tweaks to the current interface, some closer to designs, some just "better":

Screen Shot 2021-03-12 at 6 40 14 PM

List of Tweaks:

1. Increased Height of Header
2. Larger radii on header + avatar/sigil
3. UI-weight font weight for user-set name
4. Spacing between patp and bio matches space between username + avatar/sigil

Caveats:

I need to check out how these end up being applied to the mobile version of the interface. For example, I want to eliminate the border radii and border stroke on small viewports so the header is edge-to-edge.


matildepark (approved):
straightforward

PR #4603, opened 2021-03-13
chat: row hover

jamesacklin:
Adds a light-gray background to chat messages on hover while preserving the existing chat message spacing.

111016218-90971000-837a-11eb-8c90-20bb1ba189e0

Also adds commented-out dummy message actions that match the mockups and discussion found here: https://github.com/urbit/landscape/issues/460#issuecomment-797258413

Fixes urbit/landscape#460


matildepark (approved):
Looks great!

PR #4606, opened 2021-03-15
ed: sidebar group avatar fix

urcades:
This is a one character/variable change — editing the group avatar radius from `2` to `1`.

Whenever we have two radii'd elements near one another, in a nested manner, we want to always attempt to "drill down" from largest corner radius on the outside, with increasingly smaller radii applied to inner elements.


urcades:
Oh shit, it looks like i uh, included a ton of other next-js commits to this PR somehow ... did not mean to do that

matildepark (approved):

PR #4609, opened 2021-03-15
interface: only display our contact when appropriate

tacryt-socryp:
Fix https://github.com/urbit/landscape/issues/572

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
@@ -209,7 +209,7 @@ class ChatWindow extends Component<
       api,
       association,
       group,
-      contacts,
+      showOurContact,
       graph,
       history,
       groups,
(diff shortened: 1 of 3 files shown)


pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
```diff
@@ -317,7 +323,7 @@ class ChatWindow extends Component<
           onScroll={this.onScroll}
           data={graph}
           size={graph.size}
-          pendingSize={pendingSize}
+          pendingSize={pendingSize + contactsModified}
```
matildepark, line 326: Why are we adding 100 if we hide our contact?

matildepark (commented):
Mostly what I expect.

pkg/interface/src/views/apps/chat/components/ChatWindow.tsx
tacryt-socryp: To force a re-render of the virtual scroller

matildepark (approved):
We only seem to use pendingSize for triggering rerenders and not for any render-specific logic so I guess this is okay (albeit obscure and a little hacky). I'd put a comment in here to flag this

tacryt-socryp:
Thanks, added a comment.

PR #4610, opened 2021-03-15
build: pkg/interface builds pkg/npm preinstall

matildepark:
cc @tylershuster 

Instead of our awful hack in build:prod to grab npm/api and install its dependencies, we add a preinstall script that iterates through each folder in pkg/npm and installs all their dependencies, preventing any need for additional developer intervention before working on Landscape development.

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -1783,36 +1783,30 @@
       "dependencies": {
         "@babel/runtime": {
           "version": "7.12.5",
-          "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.12.5.tgz",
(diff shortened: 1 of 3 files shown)


tylershuster:
Lovely

ohAitch:
I can't shake the feeling nix aspires to be [somewhat more declarative](https://www.npmjs.com/package/node2nix) than "install whatever off npm hopefully we have a lockfile"

jtobin (approved):
Per @ohAitch's comment it's probably not too hard to rig this up using Nix.  node2nix can be used to generate derivations for the packages in `pkg/npm/*` and `pkg/interface`, and then a derivation for the Landscape build (i.e. involving webpack) can be written [something like as is described here in the node2nix docs](https://github.com/svanderburg/node2nix#using-the-nodejs-environment-in-other-nix-derivations).

That aside, I think including these steps in `preinstall` is strictly better than gluing them to the start of `install`, so +1.  As a side note, I think janeway should just be able to continue globbing happily in CI without any changes.

PR #4611, opened 2021-03-15
interface: convert api to http-api package

tylershuster:
This PR converts the global, prop-drilled API to a pseudo-hooked api that uses the `npm/http-api` package to send pokes, threads, and scrys as defined by the `npm/api` helper functions.

This is still under heavy development but works with a bit of finagling. Notable problems include not getting new messages after sending one in chat...? and initial connection issues. However, you can see the general patterns. Lots going on here, but I think this really all has to go in one go.

It also adds some functionality into `npm/http-api` itself (need to bump version) that I'm still working out to bring to parity with `channel.js`.

For ergonomics' sake, I've also updated the exports in `npm/api` so as to use the following pattern:
```ts
import { settings } from '@urbit/api';
api.poke(settings.putEntry(...));
```

cc @matildepark @tacryt-socryp @liam-fitzgerald


tylershuster:
Remaining items, but this seems to work generally:

- [ ] Initial hark-update seems to fail? (no notification counts)
- [ ] New messages trigger a spinner in chat
- [ ] Make sure GCP works

tylershuster:
Apparently the order of subscriptions matters

tylershuster:
Ready for review with the exception of the endless stream of merge conflicts that will occur and the issue with pending message resolution in chat

tylershuster:
Ok, there's some strange connection loss but functionally it's all g2g.

matildepark:
So what I’m thinking is pending last hark fixes we’ll stage next-js to livenet and next-userspace early next week and stage this on devnet. 

Though I’m afraid of the merge conflict that may result when next userspace hits master, getting merged into next-js. I guess we deal with that then.

tylershuster:
Is now the time?

matildepark:
I'm aiming for a Friday stage, there's a regression we want to stage overnight for a Thurs release ... sorry.

tylershuster:
Latest commit addresses the new fetch API throwing a "channel already bound" error when leaving and returning to the window where it is being used.

matildepark:
@tylershuster Should we get ready to stage and test this, then?

tylershuster:
Well well well if it isn't the consequences of my own actions

tacryt-socryp (approved):
Looks beautiful, let's see how it feels on release/next-js

pkg/interface/src/logic/reducers/contact-update.ts
```diff
@@ -5,25 +5,29 @@ import { ContactUpdate } from '@urbit/api';
...
   if (data) {
-    reduceState<ContactState, ContactUpdate>(useContactState, data, [
-      initial,
-      add,
-      remove,
-      edit,
-      setPublic
-    ]);
+    useContactState.getState().set(state => {
+      state = reduceState<ContactState, ContactUpdate>(useContactState, data, [
```
liam-fitzgerald, line 16: !! please don't assign to immer generated variables, this breaks it

pkg/interface/src/logic/reducers/gcp-reducer.ts
```diff
@@ -7,10 +7,12 @@ import { reduceState } from '../state/base';
 
 export default class GcpReducer {
   reduce(json: Cage) {
-    reduceState<StorageState, any>(useStorageState, json, [
-      reduceConfigured,
-      reduceToken
-    ]);
+    useStorageState.getState().set(state => {
+      state = reduceState<StorageState, any>(useStorageState, json, [
```
liam-fitzgerald, line 11: same

pkg/interface/src/logic/state/base.ts
```diff
@@ -24,11 +24,13 @@ export const reduceState = <
   state: UseStore<StateType>,
   data: UpdateType,
   reducers: ((data: UpdateType, state: StateType) => StateType)[]
-): void => {
+): StateType=> {
   const oldState = state.getState();
   const reducer = compose(reducers.map(reducer => reducer.bind(reducer, data)));
   const newState = reducer(oldState);
-  state.getState().set(state => state = newState);
```
liam-fitzgerald, line 31: again.
consider `Object.assign(state, newState)`

liam-fitzgerald (changes requested):
Please fix assigning to immer top level variables. Other than that this is a good and righteous change.

pkg/interface/src/logic/reducers/contact-update.ts
tylershuster: Yep, these are all marked as merge conflicts anyway. Will use new pattern

tylershuster:
This has gotten pretty bonkers with all the new work that's gone in. I think I should probably just publish the API/HTTP-API changes I've made while learning this and then work on rolling it in in yet another PR.

PR #4612, opened 2021-03-15
chat: minor visual fixes

jamesacklin:
Wraps message author names in chat and top-aligns the sigil to the resultant tall author box (offsets sigil by 1px for 1-to-1 visual translation (author type is 14px and sigil shapes are 16px square, this splits the difference and center-aligns the two). 

Also fixes a regression with the unread marker spacing.

A gratuitously long display name, which now does not wrap the window:
corto_9000_(iPhone 5_SE)

Unread rule between messages from different pilots:
image

Unread rule between messages from the same pilot:
Screenshot 2021-03-15 162803

Demonstrating parity with daybreak marker:
Frame 1316


urcades (approved):
Stamp!

Issue #4605, opened 2021-03-13
Ship won't start after successful replay and pier shutdown


sethfork:
**Description**
Ship won't start. Tried meld & pack, which didn't work. Then tried replaying events. No dice. When I replay, I get this after an hour or so

<img width="516" alt="Screen Shot 2021-03-13 at 2 54 05 PM" src="https://user-images.githubusercontent.com/7088679/111042565-f5e71180-840b-11eb-91de-726581767abd.png">

**System**
 - macOS 10.15.7

joemfb:
The replay completed successfully, at which point a snapshot is taken, so there shouldn't be any replay needed if you restart. Does this crash reoccur if you try? And is it the same output when it crashes (`address 0x28dc2400a out of loom!`)?

What platform are you running on?

sethfork:
Thanks for the reply @joemfb - yes, I get the same thing - `address 0x28dc2400a out of loom!`

I'm running this locally on macOS 10.15.7 - iTerm2 in a zsh shell

<img width="636" alt="Screen Shot 2021-03-13 at 7 31 00 PM" src="https://user-images.githubusercontent.com/7088679/111053547-a7993900-8432-11eb-8f01-5d9cbdce672c.png">

joemfb:
@sethfork this is a segfault. Normally you'd be able to get some crash details from console.app, but we have a SIGSEGV handler installed for the userspace paging implemented for our memory arena (the "loom"). When we get a segfault we can't handle (the memory address is question is outside of the "loom")  we just `abort()`, so crash reports won't be helpful.

This segfault is occurring early in the startup process, at a deterministic address. I've only seen a couple other crash reports like this, and only been able to debug one pier in this state: that one had a corrupted LMDB event log. In that case, the database metadata reported more pages of data than the file contained, suggesting either truncation or an unreported failure to sync the database to disk.

We don't (yet) have any LMDB diagnostic tooling in the urbit binary. But if you happen to have homebrew installed, it's pretty easy to check:

```
brew install lmdb
mdb_stat -a -e path/to/pier/.urb/log/
```

If you're able to run that, please post the output here.

---

It's possible that this is not log corruption -- you were able to read database metadata and events during replay. If that's the case, I don't know why we're crashing and will need to help you get a more detailed crash report. (Long story short, `lldb -- ./urbit -t -L ./myBackupStuffThing`, then `proc run`, the `bt all` once it crashes.)

[...]

PR #4615, opened 2021-03-16
build: patch lmdb to use F_FULLFSYNC on darwin

joemfb:
This is a (presumed) fix for the (presumed) event log corruption observed in #4605 (see the [fsync man page for MacOS](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fsync.2.html)).

@brendanhay, plz review for general nix sanity.

Also, do you have any idea how i would apply a platform-specific patch to  `pkg/hs/lmdb-static/cbits/`? (Alternately, we could just warn people off of `urbit-king` on MacOS until the log is moved.)

### DIFF
--- a/nix/overlays/native.nix
+++ b/nix/overlays/native.nix
@@ -25,4 +25,15 @@ in {
     ldapSupport = false;
     brotliSupport = false;
   };
+
+  lmdb = prev.lmdb.overrideAttrs (attrs: {
+    patches =
+      if builtins.currentSystem != "x86_64-darwin"
+      then
+        attrs.patches
+      else
+        optionalList attrs.patches ++ [
+          ../pkgs/lmdb/darwin-fsync.patch
+        ];
+  });
 }
--- a/nix/overlays/native.nix
+++ b/nix/overlays/native.nix
@@ -28,12 +28,8 @@ in {
 
   lmdb = prev.lmdb.overrideAttrs (attrs: {
     patches =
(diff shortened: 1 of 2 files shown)


brendanhay (approved):
LGTM - probably should use `stdenv.isDarwin` rather than a string comparison:

```
lmdb = prev.lmdb.overrideAttrs (attrs: {
  patches =
    optionalList attrs.patches ++ prev.lib.optional prev.stdenv.isDarwin [
      ../pkgs/lmdb/darwin-fsync.patch
    ];
});
```
It looks like `lmdb-static/cbits` is only vendored to prevent dynamic linking? If so it shouldn't exist at all, unless there are other custom changes there. It should be sufficient to ensure it's tracked via https://github.com/urbit/urbit/blob/master/nix/pkgs/hs/default.nix#L67) with the rest of the KH static dependencies.

Otherwise we can probably add [patches](https://input-output-hk.github.io/haskell.nix/reference/modules/#packagespatches) to the `haskell.nix` module options in `nix/pkgs/hs/default.nix`.

Something like: 

```
packages = {
  urbit-king.patches = [ ... ];
  ...
};
```

I'll take a quick look at the former today, unless someone has additional flavour text about the vendored lmdb.

joemfb:
Thanks, @brendanhay, I've updated the overlay.

@pilfer-pandex, can you look into updating the lmdb dependency in king haskell and remove the vendored copy?

Issue #4530, opened 2021-03-03
eyre falls over with Network.Socket.accept: resource exhausted (No file descriptors available)


lycurgus:
**Describe the bug**
My ship stopped responding to web requests, and printed `Network.Socket.accept: resource exhausted (No file descriptors available)` in the dojo. I had to restart the ship to get it to respond again.

I'm not sure what the reproduction steps look like; my home internet connection is fairly unreliable so there might have been a lot of disconnects/reconnects from my machine (local) to the ship (cloud). Other than that I've been running `urbit-king` since 1.1 was available, and then switched to 1.2 when it came out. Haven't had any problems with it previously.

**Expected behaviour**
The ship should keep accepting web connections.

**System (please supply the following information, if relevant):**
 - OS: linux (Debian 10)
 - Vere and Urbit OS versions: `urbit-king v1.2` and `urbit-os v2.34`
 - Your ship's `%base` hash (use `+trouble` to check): 0vu.kam7n.dan2e.sqscn.9iiuf.pkvdt.m0gea.8r3d2.kddjt.7n2d6.gvho5

joemfb:
@pilfer-pandex, this resource leak may be related to the memory issues you're investigating.

sorbel:
Interesting. Lycurgus, what class of ship are you sailing? If a higher class, how many sponsees are you supporting?

pilfer-pandex:
Interesting. Lycurgus, what class of ship are you sailing? If a higher class, how many sponsees are you supporting?

lycurgus:
@pilfer-pandex just a planet (~monted-tallex). Not hosting any groups or anything like that, either..

pilfer-pandex:
@joemfb, I don't think this is related. I checked the file descriptors for ~def using /proc, and found only 30 open. Surely this is not a lot.

I'm a bit stumped by this issue and unsure how to proceed.

lycurgus:
Looking at /proc on my ship's machine, it seems like I've got around 110 sockets open for the `urbit-king` process? Not sure if I'm misreading that. I also noticed quite a few shown by the `ss` utility, though I think it also shows a similar number held by my `minio` instance - so wondering if it's maybe a leak in minio that just spilled over...

Happy to check anything else that might be of interest. Not sure if there will be much to see given that I restarted the process but let me know :)

lycurgus:
[...]

Issue #4613, opened 2021-03-16
Strange TLS errors when running urbit-king


ryjm:
**Describe the bug**
When running urbit-king, ship spits out strange, serpent-like messages. Posting on behalf of ~sigryn-habrex.

sigryn-habrex: well this appears to be someone spamming my https for TLS weaknesses
sigryn-habrex: i like how it mangles the hell out of the messages though
littel-wolfur: yeah some slytherin shit going on there
sigryn-habrex: who among us hasn't experienced a HandshaHHHkaaaennnFdddasssihhhlaaaekkkdeee FFF(aaaEiiirlllreeeodddr

**Expected behaviour**
No messages from the deep.

**Screenshots**
![](https://static-sigryn-habrex.nyc3.digitaloceanspaces.com/sigryn-habrex/2021.3.15..21.03.07-Screen%20Shot%202021-03-15%20at%205%2C02%2C58%20PM.png)

**System (please supply the following information, if relevant):**
- urbit-king 1.2

**Additional context**
ship URL was shared on UC

philipcmonk:
I've also experienced this, though I don't have any more info.  I too use https.

pilfer-pandex:
That's pretty funny. The reason you're seeing interlacing like that is that
the standard haskell logger we use, RIO.Logger, just writes to the
underlying handle, without thread-safing it in any way. When I figured this
out last year (by reading the logger code, and, ultimately, the internals
of the GHC file handle implementation, just to be sure), I immediately
asked @benjamin-tlon about it, and he said "Yeah, that's the case. I've
implemented locking loggers in Haskell before in the past and they're
difficult and have performance tradeoffs, which is why snoyberg probably
didn't try in RIO. I think it's best to just keep it like this."
Accordingly, I haven't changed it.

Now as you know, each request to a warp server spawns a new (Haskell,
green) thread. So if you have two inbound invalid requests, it's quite
likely, based on the runtime's thread fairness algorithm, you'll get a
hhaakkeeFFaaiilleedd, or some variant thereof.

Here are some action items:
- actually figure out what causes a "handshake failed" (can you guys
confirm that you weren't trying to make legit requests? are you like on a
cloud provider or something, getting pummeled by roving gangs of bandits?)
- figure out why these get printed to stderr, rather than logged at a lower
priority, if this is an ordinary occurrence for cloudy ships, and change
that (you guys aren't running with --stderr, are you?)
- actually bite the bullet and investigate thread-safing the logger. The
[...]

PR #4616, opened 2021-03-16
king: commit seppuku when server is out of fds, also setrlimit allowing for more fds

pilfer-pandex:
Following joe's suggested response to #4530, we shut down the king when the http server is saturated with requests, as an alternative to getting the server permanently stuck in a state where it cannot accept new connections.

To make this condition far less likely, we increase the resource cap on file descriptors from 256 to 10240 (the value of OPEN_MAX, at least on darwin).

In the long run, we should probably try to restart the http server rather than shut down the king, or selectively go through and terminate specific requests.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
@@ -99,6 +99,7 @@ import qualified Data.Set                as Set
 import qualified Data.Text               as T
 import qualified Network.HTTP.Client     as C
 import qualified System.Posix.Signals    as Sys
+import qualified System.Posix.Resource   as Sys
 import qualified System.ProgressBar      as PB
 import qualified System.Random           as Sys
 import qualified Urbit.EventLog.LMDB     as Log
@@ -460,6 +461,13 @@ pillFrom = \case
     noun <- cueBS body & either throwIO pure
     fromNounErr noun & either (throwIO . uncurry ParseErr) pure
 
+multiOnFatal :: HasKingEnv e => e -> IO ()
+multiOnFatal env = runRIO env $ do
+  (view stderrLogFuncL >>=) $ flip runRIO $ logError
+    ("Urbit is shutting down because of a problem with the HTTP server.\n"
+    <> "Please restart it at your leisure.")
+  view killKingActionL >>= atomically
+
 newShip :: CLI.New -> CLI.Opts -> RIO KingEnv ()
 newShip CLI.New{..} opts = do
   {-
@@ -472,7 +480,8 @@ newShip CLI.New{..} opts = do
     "run ship" flow, and possibly sequence them from the outside if
     that's really needed.
   -}
-  multi <- multiEyre (MultiEyreConf Nothing Nothing True)
+  env <- ask
+  multi <- multiEyre (multiOnFatal env) (MultiEyreConf Nothing Nothing True)
 
   -- TODO: We hit the same problem as above: we need a host env to boot a ship
   -- because it may autostart the ship, so build an inactive port configuration.
@@ -660,6 +669,7 @@ main = do
 
   hSetBuffering stdout NoBuffering
   setupSignalHandlers
+  setRLimits
 
   runKingEnv args log $ case args of
     CLI.CmdRun ko ships                       -> runShips ko ships
@@ -693,6 +703,11 @@ main = do
     for_ [Sys.sigTERM, Sys.sigINT] $ \sig -> do
       Sys.installHandler sig (Sys.Catch onKillSig) Nothing
 
+  setRLimits = do
+    nofiles <- Sys.getResourceLimit Sys.ResourceOpenFiles
+    Sys.setResourceLimit Sys.ResourceOpenFiles
+      nofiles { Sys.softLimit = Sys.ResourceLimit 10240 }
+
   verboseLogging :: CLI.Cmd -> Bool
   verboseLogging = \case
     CLI.CmdRun ko ships -> any CLI.oVerbose (ships <&> \(_, o, _) -> o)
@@ -715,7 +730,6 @@ main = do
                                    | otherwise -> CLI.LogFile Nothing
       CLI.CmdRun ko _                          -> CLI.LogStderr
       _                                        -> CLI.LogStderr
-
 
 {-
   Runs a ship but restarts it if it crashes or shuts down on it's own.
@@ -792,7 +806,8 @@ runShips CLI.Host {..} ships = do
                                    -- a king-wide option.
         }
 
-  multi <- multiEyre meConf
+  env <- ask
+  multi <- multiEyre (multiOnFatal env) meConf
 
   ports <- buildPortHandler hUseNatPmp
 
--- a/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/King/Main.hs
@@ -704,9 +704,13 @@ main = do
       Sys.installHandler sig (Sys.Catch onKillSig) Nothing
 
   setRLimits = do
-    nofiles <- Sys.getResourceLimit Sys.ResourceOpenFiles
+    openFiles <- Sys.getResourceLimit Sys.ResourceOpenFiles
+    let soft = case Sys.hardLimit openFiles of
+          Sys.ResourceLimit lim     -> Sys.ResourceLimit lim
+          Sys.ResourceLimitInfinity -> Sys.ResourceLimit 10240  -- macOS
+          Sys.ResourceLimitUnknown  -> Sys.ResourceLimit 10240
     Sys.setResourceLimit Sys.ResourceOpenFiles
-      nofiles { Sys.softLimit = Sys.ResourceLimit 10240 }
+      openFiles { Sys.softLimit = soft }
 
   verboseLogging :: CLI.Cmd -> Bool
   verboseLogging = \case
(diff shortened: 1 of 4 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Eyre.hs
```diff
@@ -356,7 +361,11 @@ eyre env who plan isFake stderr sub = (initialEvents, runHttpServer)
   restart :: Drv -> HttpServerConf -> RIO e Serv
   restart (Drv var) conf = do
     logInfo "Restarting http server"
-    let startAct = startServ who isFake conf plan stderr sub
+    let onFatal = runRIO env $ do
+          -- XX instead maybe restart following logic under HSESetConfig below
+          stderr "You've been DDoSed. Please restart your ship."
```
joemfb, line 366: [citation needed]

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
```diff
@@ -254,19 +255,28 @@ startServer
...
   let host = case hos of
         SHLocalhost -> "127.0.0.1"
         SHAnyHostOk -> "*"
 
+  let handler r e
+        | Just (IOError {ioe_type = ResourceExhausted}) <- fromException e
+        = runRIO envir $ do
+            logError (displayShow e)
+            io onFatal
+        | otherwise = W.defaultOnException r e
```
joemfb, line 272: What does the default exception do? Is that what's currently printing and then hanging?

joemfb (changes requested):
I think we should print the actual error before shutting down, and I'm not sure we want to be deferring to the default exception handler.

Also, there's a build issue with this PR. Look like maybe you missed an Eyre call site?

joemfb:
As for `OPEN_MAX` it's a constant `10240`, and has been for quite awhile. FFI seems like serious overkill; it'd probably be fine to just hardcode it with a comment.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre.hs
pilfer-pandex: [S. T. Zargar, J. Joshi and D. Tipper, "A Survey of Defense Mechanisms Against Distributed Denial of Service (DDoS) Flooding Attacks," in IEEE Communications Surveys & Tutorials, vol. 15, no. 4, pp. 2046-2069, Fourth Quarter 2013, doi: 10.1109/SURV.2013.031413.00127.](https://sci-hub.se/https://ieeexplore.ieee.org/abstract/document/6489876)

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
pilfer-pandex: The default handler is found at https://hackage.haskell.org/package/warp-3.3.14/docs/src/Network.Wai.Handler.Warp.Settings.html#defaultOnException. It either prints or does not print the exception depending on the type, then discards it. Accordingly, it's the origin of the printout @lycurgus reported. It's not the source of the hang; that's because warp gets into a bad state when it runs out of file descriptors; it's not a generic thing that happens on any exception we might encounter here, nor does "the exception handler hang" as you describe it.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
```diff
@@ -254,19 +255,28 @@ startServer
...
   envir <- ask
 
   let host = case hos of
         SHLocalhost -> "127.0.0.1"
         SHAnyHostOk -> "*"
 
+  let handler r e
+        | Just (IOError {ioe_type = ResourceExhausted}) <- fromException e
+        = runRIO envir $ do
+            logError (displayShow e)
```
pilfer-pandex, line 270: Replying to your top-level comment: We do print the error before shutting down, and do so right here. Do you want this to be to stderr in addition to (or instead of) the logError?

pilfer-pandex (commented):
> it's a constant 10240

Strange. I got what I think was EINVAL when I tried even 1023. (I was trying to do binary search to discover it lol.)

Let me try again.

pilfer-pandex:
Holy shit my Stack cache was dirty. That's why the build error got through. It's holding onto previous copies of compiled modules as though they were the current ones.

Also your suggestion of 10240 worked. I suspect it was just too late last night and I forgot to copy the executable when I did my test.

pilfer-pandex:
To confirm, do we also want 10240 on linux?

pilfer-pandex:
@joemfb, build problem this time seems to be in nix itself; there's a printout inviting me to report a bug to nixos, to wit:

```

installing 'nss-cacert-3.49.2'
error: unexpected EOF reading a line

---- oh no! --------------------------------------------------------------------
Jeeze, something went wrong. If you can take all the output and open
an issue, we'd love to fix the problem so nobody else has this issue.
```

Accordingly I think you can proceed with the review. @brendanhay have you seen this problem before?

pilfer-pandex:
Seems to be a nondeterministic failure that you just get occasionally if you use macos?? https://github.com/NixOS/nix/issues/3605. "Jeeze" indeed.

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
```diff
@@ -693,6 +703,11 @@ main = do
     for_ [Sys.sigTERM, Sys.sigINT] $ \sig -> do
       Sys.installHandler sig (Sys.Catch onKillSig) Nothing
 
+  setRLimits = do
+    nofiles <- Sys.getResourceLimit Sys.ResourceOpenFiles
+    Sys.setResourceLimit Sys.ResourceOpenFiles
+      nofiles { Sys.softLimit = Sys.ResourceLimit 10240 }
```
joemfb, line 709: We still want the max number of files on linux, which should be accurately reported by `getResourceLimit` and can be higher than this. It just needs to be hardcoded on MacOS.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre.hs
joemfb: I don't think this error message is appropriate. If this is to be a generic error handler, what it print should be generic. If it's specifically meant to handle the ResourceLimit exception caused by leaking sockets, it shouldn't make grandiose claims about cyberwarfare.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
joemfb: Do you have any indication that printing/ignoring exceptions is the right default behavior here? It seems like it'd be better to whitelist non-fatal exceptions and otherwise `abort()`.

joemfb (changes requested):

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
pilfer-pandex: Yeah, I do. I looked for examples of exception handling behavior in major open source codebases using warp. Take a look at:

- https://github.com/hasura/graphql-engine/blob/master/server/src-lib/Hasura/App.hs#L583
- https://github.com/PostgREST/postgrest/blob/main/main/Main.hs#L89

Both of these major, legitimate applications *fail to modify the default exception handler at all*, so they definitely don't `abort()` on any exception. 

Now the comment at https://hackage.haskell.org/package/warp-3.3.14/docs/src/Network.Wai.Handler.Warp.Settings.html#defaultShouldDisplayException says

> The goal is to hide exceptions which occur under the normal course of the web server running.

So perhaps the list found in that function will serve as a suitable whitelist? But in fact, the stderr printouts reported in https://github.com/urbit/urbit/issues/4613 are (I've concluded) coming out of this exception handler, and I think we can agree that these should not shut down the server.

Because of the (unfortunate) nature of the Haskell exception mechanism, it's not possible to confidently get a stable list of all the exceptions that might trip this handler, let alone fix a subset of those such that we can be sure its complement contains only things worth shutting down the server over. I'm just not comfortable implementing a policy like you propose and putting it on zod. I'd have to read all of the upstream libraries to find all places where exceptions are thrown, and I'd have to be confident that that doesn't change over time.

I'm going to issue a change here which uses the logger instead of stderr for printing exceptions, and maybe adds more info. (That's a good fix for the linked issue.)

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
```diff
@@ -254,19 +255,38 @@ startServer
...
+        Just r  -> "eyre: failed request from " <> (tshow $ W.remoteHost r)
+                <> " for " <> (tshow $ W.rawPathInfo r) <> ": " <> tshow e
+        Nothing -> "eyre: server exception: " <> tshow e
+
   let opts =
         W.defaultSettings
           & W.setHost host
           & W.setPort (fromIntegral por)
-          & W.setTimeout (5 * 60)
+          & W.setTimeout 30
```
joemfb, line 288: This timeout might be a little too aggressive, but it'll be easy to tune it later.

joemfb (approved):
LGTM

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
pilfer-pandex: IIRC it's the timeout vere uses.

PR #4619, opened 2021-03-16
interface: remove referrer headers from img tags

matildepark:
In response to an unfiled bug on display in UC last weekend.

image

### DIFF
--- a/pkg/interface/src/views/components/RichText.js
+++ b/pkg/interface/src/views/components/RichText.js
@@ -37,7 +37,7 @@ const RichText = React.memo(({ disableRemoteContent, ...props }) => (
           return <RemoteContent className="mw-100" url={linkProps.href} />;
         }
 
-        return <Anchor display="inline" target='_blank' rel='noreferrer noopener' borderBottom='1px solid' remoteContentPolicy={remoteContentPolicy} {...linkProps}>{linkProps.children}</Anchor>;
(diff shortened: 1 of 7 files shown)


liam-fitzgerald (approved):

pkg/interface/src/views/components/RichText.js
```diff
@@ -37,7 +37,7 @@ const RichText = React.memo(({ disableRemoteContent, ...props }) => (
           return <RemoteContent className="mw-100" url={linkProps.href} />;
         }
 
-        return <Anchor display="inline" target='_blank' rel='noreferrer noopener' borderBottom='1px solid' remoteContentPolicy={remoteContentPolicy} {...linkProps}>{linkProps.children}</Anchor>;
+        return <Anchor display="inline" target='_blank' referrerPolicy='noreferrer noopener' borderBottom='1px solid' remoteContentPolicy={remoteContentPolicy} {...linkProps}>{linkProps.children}</Anchor>;
```
tacryt-socryp, line 40: Is this right, since it doesn't have dashes like all the others?

matildepark: [Yeah](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel). The web's a weird place.

matildepark: Oh wait, I wonder if there's a referrer policy on this. I shouldn't have changed anchors lol

PR #4621, opened 2021-03-17
landscape/RichText: revert anchors to use rel= for referrer

matildepark:


liam-fitzgerald (approved):

Fang-:
Have we considered a single `<meta name="referrer" content="no-referrer">` in the `index.html` instead?

PR #4623, opened 2021-03-17
hoon: update seminoun bunt to be fully blocked

joemfb:
The current default `seminoun` is wrong, given the pattern of bunting batteries in `+play:ut`.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -6627,7 +6627,7 @@
 +$  seminoun
   ::  partial noun; blocked subtrees are ~
   ::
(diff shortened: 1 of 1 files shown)


belisarius222:
Can you remind me what was broken because of this?  I understand that bunting to full is wrong, since we should assume a piece of data is unknown until proven otherwise, but I don't recall what problems this caused that I should expect this change to fix.

joemfb:
The incorrect default leads us thinking the battery of `add` is `0` here:

```
(fragment:musk 2 ~(bran ut -:!>(add)))
[mask=[%full blocks={}] data=0]
```

(but this doesn't affect current compilation patterns).

PR #4624, opened 2021-03-17
hoon: strip faces in =^

joemfb:
This PR is a proposal for a minor behavior change in the `=^` rune.

The current macro expansion has something like:

```
=>  a=foo
=^  b  a  bar
baz
```

```
=>  a=foo
=/  x=bar
=/  b  -.x
=.  a  +.x
baz
=
```

The use of `=.` means that any outer face in the tail of (the product of) `bar` is stripped -- `=.` expands to `%_` which uses a cast-by-example (`^+`). But the outer face  of the head of (the product of) `bar` is preserved inside `b`.

With this change, both faces are stripped (if present). This PR is a draft, as the proposal has been more controversial than anticipated.

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -944,8 +944,9 @@
           =^  lab  sac  (~(spot wa sac) 7 vax)
           $(vax lab)
         ::
-        ?>  ?=(?(%pass %give) -.p.bal)
-        [[%hurl goof p.bal] sac]
+        ~!  bal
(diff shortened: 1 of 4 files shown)


belisarius222:
LGTM.  What's the controversy?

joemfb:
See urbit-community/development.

joemfb:
@juped long ago convinced me that this is a bad idea -- structural deconstruction should leave faces intact, and types (faces in this case) should not be changed except through casts or pattern matches.

PR #4626, opened 2021-03-18
landscape: convert non-latin group and channel names to ascii

mirtyl-wacdec:
Fixes urbit/landscape#498 problem where groups wouldn't be created as non-ascii characters get ignored by the parser.

Some cosmetic changes done automatically by my overzealous VSCode setup, sorry.


pkg/interface/src/logic/lib/util.ts
```diff
@@ -319,11 +320,13 @@ export function getContactDetails(contact: any) {
 }
 
 export function stringToSymbol(str: string) {
+  const ascii = anyAscii(str);
+  console.log(ascii);
```
liam-fitzgerald, line 324: Please remove console.log

liam-fitzgerald (commented):
Thanks for your contribution!

This looks super reasonable. I'm not overly familiar with internationalisation, but this appears to just romanize the string before feeding it into the symbol converter? How does it deal with emojis, or other codepoints that don't have obvious romanisations?

liam-fitzgerald:
You'll also need to fix the conflicts from changing the base. All JS-only PRs should be targeted to `release/next-js`

mirtyl-wacdec:
> This looks super reasonable. I'm not overly familiar with internationalisation, but this appears to just romanize the string before feeding it into the symbol converter? How does it deal with emojis, or other codepoints that don't have obvious romanisations?

Yes, it romanizes the string according to a set of rules. The library has a large list of language it targets. Emojis I just checked and are transcribed to their descriptions, e.g. `:joy:` becomes `joy`.
I think this would target 99%+ of languages we're likely to encounter on Urbit

matildepark (approved):
Was worried ascii library would convert some chars to uppercase, but doesn't seem to in practice (tried with CN chars) — can't think of a better way to duplicate what this offers and it's forward thinking

PR #4627, opened 2021-03-19
ames: silence "fragment crashed" printf by default

joemfb:
It can still be enabled with `|ames-verb %msg`, but it's too loud to be on by default.


belisarius222 (approved):
LGTM.  Thanks for doing this; I should have fixed it when we discussed it a while ago, but it slipped my mind.

PR #4628, opened 2021-03-19
hark: eliminate chatmessage hover backgrounds

jamesacklin:
Adds an optional `hoverMarker` prop to `ChatMessage`. When set, this hides the washedGray hover background. We then use this in the Notifications view, where we want to prevent the gray background from appearing until we get through the detailed refactor coming from @g-a-v-i-n.

Pointed out by @matildepark , no corresponding Landscape issue.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -207,7 +207,7 @@ const MessageWrapper = (props) => {
     <Box
       py='1'
       backgroundColor={
-        hovering && props.hoverMarker !== false ? 'washedGray' : 'transparent'
+        hovering && !props.hideHover ? 'washedGray' : 'transparent'
       }
       position='relative'
       {...bind}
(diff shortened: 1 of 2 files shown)


urcades:
Yes.jpg

pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
```diff
@@ -206,7 +206,9 @@ const MessageWrapper = (props) => {
   return (
     <Box
       py='1'
-      backgroundColor={hovering ? 'washedGray' : 'transparent'}
+      backgroundColor={
+        hovering && props.hoverMarker !== false ? 'washedGray' : 'transparent'
```
matildepark, line 210: `false` checks rarely play nice being passed around. Could we rewrite this flipped to check for true?

jamesacklin: Sure, I'll just invert the flag. No problem

matildepark (approved):

PR #4629, opened 2021-03-19
hark: kill incorrect count reporting

liam-fitzgerald:
When we receive a notification, it might not affect our count because
it’ll get merged on the server side, but we unconditionally increment
our count on the client if the notification hasn’t been loaded yet.
Addresses this by storing a set of time, index so we can compute whether
or not the notification is merged or unloaded and change the count
accordingly. Calculates the notificationsCount as derived state, instead of independently updated state

Fixes urbit/landscape#276


tacryt-socryp (approved):
Principled as a solutin and good. Still need to test.

PR #4631, opened 2021-03-19
chat: prevent marker text from wrapping

jamesacklin:
On the tin. Applies to daybreak and unread markers.

Fixes urbit/landscape#589


matildepark (approved):
lgtm

PR #4633, opened 2021-03-20
sidebar: updated channel state design

tylershuster:
Fixes https://github.com/urbit/landscape/issues/122

Hashed the final bits out with @urcades in uc/design. Dots indicate *notifications* in a channel where bright indicates unread. 60% gray indicates read but subscribed and 30% gray indicates unsubscribed. Selected/hovered for both subscribed and unsubscribed channels shown. (There isn't a dot in the light version on accident, but you can verify it does show up)

2021 3 20 03 53 06-image

### DIFF
--- a/pkg/interface/src/views/components/Dot.tsx
+++ b/pkg/interface/src/views/components/Dot.tsx
@@ -1,0 +1,22 @@
+import { Box, BoxProps } from '@tlon/indigo-react';
+import React, { ReactElement } from 'react';
+
+type DotProps = BoxProps & {
+  color?: string
+};
+
+const Dot = ({ color, ...rest }: DotProps): ReactElement => {
+  return (
+    <Box
+      style={{
+        backgroundColor: color || 'currentColor',
+        width: '4px',
+        height: '4px',
+        borderRadius: '50%'
+      }}
+      {...rest}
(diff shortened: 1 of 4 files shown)


pkg/interface/src/views/components/Dot.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const Dot = (props) => {
+  return (<span style={{
+    backgroundColor: props.color || 'currentColor',
+    width: '4px',
+    height: '4px',
+    borderRadius: '50%'
+  }}></span>);
```
liam-fitzgerald, line 9: This should use `Box` and the associated styled system props

liam-fitzgerald (approved):
LGTM, other than the nit about the bare `span`

pkg/interface/src/views/components/Dot.tsx
matildepark: Seems like you could use `Icon` and `Bullet` or etc.

tylershuster:
Ok, I used `Box` and typed it correctly. I'd like to use `Icon` but the padding is kind of funky and the scale doesn't allow specifying exactly 4px.

PR #4640, opened 2021-03-22
hs: remove lmdb-static; just use lmdb

pilfer-pandex:
Fulfilling @joemfb's request in #4615, we remove lmdb-static, which appears to be a verbatim copy of hackage's lmdb, plus a verbatim copy of the C implementation of lmdb. With this change, we move to lmdb, which expects the C implementation to sit in the linking path, which nix now ensures for us.


joemfb:
Can you point me towards the nix config that provides our (possibly overlayed) lmdb derivation to these haskell builds? I don't see anything explicitly handling that.

Also, this PR should target `release/next-vere` (which has the patched lmdb already).

pilfer-pandex:
Great question. My understanding from "talking" with Bren today is that it *is* #4615 which provides this overlay. The lmdb provided by nix, which is the lmdb that will be available to stack/cabal for build, is precisely the lmdb built by the relevant nix derivation, which will include the patch because we said so there. @brendanhay, please please tell me if I have this wrong.

joemfb:
I was expecting to see something like `"--ghc-option=-optl=-L${lmdb}/lib"` in the `configureFlags` in `pkgs/hs/default.nix`, but maybe I misunderstood his comment in #4615.

pilfer-pandex:
No, it doesn't look like you misunderstood his comment; I just missed it. Mind you, I have no idea what I'm doing here, but I put something in the spot he asked for, based on what you put in your patch, which seemed to me to make sense, and which was successfully loaded by nix-shell. It would be nice if there were some action I could take from within nix-shell to verify that the patched llvm is entering into the haskell build, but I have no idea what that would be.

pilfer-pandex:
And adding 

```
"--ghc-option=-optl=-L${lmdb}/lib"
```

as you suggest results in


```
error: undefined variable 'llvm' at /Users/harrison/Documents/urbit-ipc/nix/pkgs/hs/default.nix:71:35
```

/tosses hands up in air and shrugs

joemfb:
It was `error: undefined variable 'llvm'`, not `'lmdb'`? If it was the latter that makes sense, you'd have to pass `lmdb` to that derivation.

To check that my patch was being used, I monitored the syscalls while running vere:

```
sudo dtrace -n syscall:::entry'/pid == $PID/{ @syscalls[probefunc] = count(); }'
... do some stuff
... ctrl-c to exit and print output
```

Without the patch, you'll see a number of `fsync()` calls corresponding to the number of event-batches committed to disk. With the patch, those will be `fcntl()` calls. (All of this is only on MacOS.)

pilfer-pandex:
It was lmdb in the error message, sorry. By "pass" do you mean "add to the listing in the `{}` at the top of the file?

pilfer-pandex:
Okay, with nix-build -A hs.urbit-king, we (eventually) get

```
source root is urbit-king
patching sources
urbit-king.cabal is up-to-date
applying patch /nix/store/zk6vd37pvl08qdbnmnif75qvqfyp7xkz-darwin-fsync.patch
can't find file to patch at input line 5
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
|diff --git a/libraries/liblmdb/mdb.c b/libraries/liblmdb/mdb.c
|index fe65e30..0070215 100644
|--- a/libraries/liblmdb/mdb.c
|+++ b/libraries/liblmdb/mdb.c
--------------------------
File to patch:
Skip this patch? [y]
Skipping patch.
1 out of 1 hunk ignored
builder for '/nix/store/ia1gfmc1956h6jgnw622nzxj8vakw5m9-urbit-king-lib-urbit-king-1.2.drv' failed with exit code 1
cannot build derivation '/nix/store/gcwxj1cfxvrnyzl15h6jpm57fjdls3j0-urbit-king-1.2-coverage-report.drv': 1 dependencies couldn't be built
error: build of '/nix/store/gcwxj1cfxvrnyzl15h6jpm57fjdls3j0-urbit-king-1.2-coverage-report.drv', '/nix/store/vhh15z5qim6laisvpcdlfidq4clia7rl-urbit-king-test-urbit-king-tests-1.2-check.drv' failed
```

So I've done something wrong, again.

pilfer-pandex:
With the latest change, nix-build -A hs.urbit-king works. I misread bren's comment; it seems he was saying we needed to add a `patches` field *only if* we were still using `lmdb-static`, which we're not. That's why there's no file to patch.

I still want @brendanhay to review this, though.

joemfb:
Can you run the `dtrace` test I suggested?

pilfer-pandex:
It seems that the the binary built by stack from within nix-shell does not have the correct version of lmdb, based on this dtrace probe. I don't know how to fix this. I'm trying to figure out how to get a binary out of nix-build, but it seems it makes only the library and the tests??

pilfer-pandex:
Okay the command to run is actually

```
nix-build -A hs.urbit-king.components.exes.urbit-king --arg enableStatic true
```

With the binary coming out of that command, we see

```
bsdthread_create                                                  1
  thread_selfid                                                     1
  madvise                                                           5
  psynch_cvclrprepost                                              42
  psynch_mutexdrop                                                 50
  psynch_mutexwait                                                 50
  lseek                                                            52
  fcntl                                                            59
  sem_post                                                         59
  sem_wait                                                         59
  getrusage                                                        67
  close_nocancel                                                   89
  fstatfs64                                                        89
  open_nocancel                                                    89
  writev                                                          111
  sigprocmask                                                     134
  getdirentries64                                                 178
  sendto                                                          246
  pwrite                                                          251
  recvfrom                                                        481
  close                                                           578
  open                                                            578
  lstat64                                                         666
  fstat64                                                        1156
  ioctl                                                          1156
  __semwait_signal                                               1157
  kevent                                                         1180
  read                                                           1462
  psynch_cvsignal                                                2648
  psynch_cvwait                                                  2649
  poll                                                           3240
  write                                                          3296
```

which looks good. So on that basis, your proposed test passes. But we still need to get stack working properly. @brendanhay?

joemfb (approved):
Thanks for checking. Since the binary we release is from `nix-build`, I'm in favor of merging this as is (so we can cut the release ASAP) and figuring out stack separately.

LGTM

joemfb:
It's not exactly as if all previous releases were *wrong* (let any other piece of software that uses lmdb on macos), it's just that they were not completely crash-safe in the way they should be.

PR #4641, opened 2021-03-22
publish: call all hooks before returning early

matildepark:
Fixes urbit/landscape#597


tylershuster:
Yep, I caught a few of these but obviously not all.

tacryt-socryp (approved):

PR #4643, opened 2021-03-22
interface: wraps zustand setter in batchUpdate

tylershuster:
It works, [is correct](https://github.com/pmndrs/zustand#calling-actions-outside-a-react-event-handler) — hard to reproduce https://github.com/urbit/landscape/issues/536 but fixes in in theory

ht @liam-fitzgerald

### DIFF
--- a/pkg/interface/src/logic/store/store.ts
+++ b/pkg/interface/src/logic/store/store.ts
@@ -20,6 +20,7 @@ import GcpReducer from '../reducers/gcp-reducer';
 import { OrderedMap } from '../lib/OrderedMap';
 import { BigIntOrderedMap } from '../lib/BigIntOrderedMap';
 import { GroupViewReducer } from '../reducers/group-view';
+import { unstable_batchedUpdates } from 'react-dom';
 
 export default class GlobalStore extends BaseStore<StoreState> {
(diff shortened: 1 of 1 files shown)


pkg/interface/src/logic/state/base.ts
```diff
@@ -25,10 +26,12 @@ export const reduceState = <
   data: UpdateType,
   reducers: ((data: UpdateType, state: StateType) => StateType)[]
 ): void => {
-  const oldState = state.getState();
-  const reducer = compose(reducers.map(reducer => reducer.bind(reducer, data)));
-  const newState = reducer(oldState);
-  state.getState().set(state => state = newState);
+  unstable_batchedUpdates(() => {
```
liam-fitzgerald, line 29: This should actually be in store/store.ts so that all the updates from a subscription message should be batched together. Also, can you remove the global setState while you're there

liam-fitzgerald (changes requested):

tylershuster:
Changed to global store reducer — global store isn't quite dead yet, but this should help twist its neck before the API PR finishes it off

liam-fitzgerald (approved):
righteous

PR #4644, opened 2021-03-23
chat: new unread message style

jamesacklin:
- Restyles `foo messages since bar` banner in Chat as per spec. Prevents chat contents from appearing behind our semi-transparent colors by mocking in a solid-color background. Respects dark mode.
- Replaces left-padded date in `Timestamp` with a non-breaking space to retain typographic purity.

Light mode:
image

Dark mode:
image

Light mode with a very wide viewport:
image

Fixes urbit/landscape#563

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/unread-notice.js
+++ b/pkg/interface/src/views/apps/chat/components/unread-notice.js
@@ -55,8 +55,7 @@ export const UnreadNotice = (props) => {
               cursor='pointer'
               onClick={onClick}
             >
-              {unreadCount} new message{unreadCount > 1 ? 's' : ''}{' '}
-              <Text color='lightGray'>since </Text>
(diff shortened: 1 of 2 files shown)


matildepark (approved):
Looks great, just getting an informal stamp from @urcades out of band, will merge if I get a clear

tacryt-socryp:
Why is the word 'since' so light?

matildepark:
> Why is the word 'since' so light?

This is why I'm looking for a design clear, lol

urcades:
This is how it was originally designed, but now that y'all mention it, we should just keep it the default text color.

Let's make it black/one-color, and I will provide a stamp.

jamesacklin:
Uniform colors abound.

image

urcades:
Spirit Stamp

PR #4646, opened 2021-03-23
group feed: interface improvements

liam-fitzgerald:
- Refactors profileoverlay to have fixed position, so it can escape it's bounding container
- Adds code, links input to adding posts (code is stubbed, given we cannot render them now)
- Notifications for post, no unreads, only notify if it's a reply
- VirtualScroller style fixes


tacryt-socryp (approved):
LGTM, thank you!

PR #4649, opened 2021-03-23
meta: only build ci if changing arvo, vere, king

matildepark:
Part 1 of a "CI specificity" series — 

In userspace we often have JS-specific changes that are just hanging on CI for a long period of time. They don't need to be in CI, and they clog up the queue. This only runs the build CI if we're changing files in pkg/arvo, pkg/hs or pkg/urbit in the push.

Could use a looksee because I am unsure how to test the change, though I noticed the first time I pushed this change there was a fail because of a typo, and I fixed the typo and it just didn't run on the force push, so presumably it worked.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -32,7 +32,17 @@
 
 name: build
 
-on: [push, pull_request]
+on:
+  push:
+    paths:
+      - 'pkg/arvo'
+      - 'pkg/hs'
+      - 'pkg/urbit'
+  pull_request:
+    paths:
+      - 'pkg/arvo'
+      - 'pkg/hs'
+      - 'pkg/urbit'
 
 jobs:
   urbit:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,11 +38,15 @@ on:
       - 'pkg/arvo'
       - 'pkg/hs'
       - 'pkg/urbit'
+      - 'bin'
+      - 'nix'
   pull_request:
     paths:
       - 'pkg/arvo'
       - 'pkg/hs'
       - 'pkg/urbit'
+      - 'bin'
+      - 'nix'
 
 jobs:
   urbit:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,14 +36,22 @@ on:
   push:
     paths:
       - 'pkg/arvo'
+      - 'pkg/docker-image'
+      - 'pkg/ent'
+      - 'pkg/ge-additions'
       - 'pkg/hs'
+      - 'pkg/libaes_siv'
       - 'pkg/urbit'
       - 'bin'
       - 'nix'
   pull_request:
     paths:
       - 'pkg/arvo'
+      - 'pkg/docker-image'
+      - 'pkg/ent'
+      - 'pkg/ge-additions'
       - 'pkg/hs'
+      - 'pkg/libaes_siv'
       - 'pkg/urbit'
       - 'bin'
       - 'nix'


ohAitch:
The "correct" answer here is probably cachix fetching `build succeeded` results that don't have `pkg/interface` as an _input_, but this is a step in the right direction.

matildepark:
> The "correct" answer here is probably cachix fetching `build succeeded` results that don't have `pkg/interface` as an _input_, but this is a step in the right direction.

I don't know what this means but I intend to write some TS + linter checks for JS changes next

ohAitch:
Idk if the fix is literally "write up a pkg/*/[flake.nix](https://www.tweag.io/blog/2020-05-25-flakes/)" but
- if pkg/urbit hasn't changed since the last time we built it, we don't want to rebuild it
- iiuc tlon CI uses cachix to reuse nix expression evaluations across CI runs
- if you're seeing a full build anyway, this means the js changes are ending up upstream of the vere/king compile somehow, such that the nix formula to rerun the build is not identical
- (e.g. if the entirety of `urbit/urbit` is pulled in as a dependency regardless of `pkg/`)
- and if this were set up to more accurately specify dependencies, you'd neither have to *skip* nor *run* the build, and could just get "oh yeah we already built this hash of the `pkg/urbit` source, here's the test results" transparently and ~instantly
- I may of course be wrong on any of those points

jtobin (commented):
Looks good, though I think you could use `paths_ignore` and specify the interface/npm stuff instead?  We may still want to build if tweaking `.github`, `nix`, pills, other `pkg` contents, and so on.

matildepark:
> Looks good, though I think you could use `paths_ignore` and specify the interface/npm stuff instead? We may still want to build if tweaking `.github`, `nix`, pills, other `pkg` contents, and so on.

There are userspace changes that touch interface and arvo that should be compiled with this ci. We want inclusive paths so that per module ci is the end goal.

jtobin (approved):
Aha, gotcha.  LGTM then.

matildepark:
Are there other paths that necessitate a build that I'm not including?

jtobin:
I would probably include /nix (if e.g. the nixpkgs pin changes then perhaps one wants to perform the build) and /bin to catch pill-only changes.  On second thought it's probably not necessary to include /.github -- that can just be enabled when needed, e.g. if testing changes to the build workflow.

matildepark:
Just holding on @brendanhay input before I merge this...

brendanhay (approved):
We'll need any inputs to `pkg/urbit` as well:

- pkg/libaes_siv # pkg/urbit dependency
- pkg/ent # pkg/urbit dependency
- pkg/ge-additions # pkg/urbit dependency

And probably the docker image build code should be rebuilt if it changes:

- pkg/docker-image

Most of the things built/defined in [default.nix](https://github.com/urbit/urbit/blob/master/default.nix#L87-L124) probably has a case for being added to paths.

ohAitch:
If we're going that detailed, I think it *would* be much more robust to resolve this in nix, such that we have an ironclad hash-based guarantee that nothing the build sees changed, and nothing that changed can be seen by the build.

matildepark:
1. Is someone who understands the Nix build system willing to do the work to incorporate TypeScript checks and lints for different submodules?
2. Is there any drawback for the redundancy this solution introduces?

matildepark:
I guess additionally:

3. Do we have the capacity to investigate why we are not caching Nix derivations for changes that don't touch Vere nor Arvo, and presently aren't even being built?
4. Are we OK with attempting to build CI on changing markdown files, falling back to Nix to check for the guarantee?

ohAitch:
If I were doing the build as a docker file, I'd run "delete all the markdown files, build, (maybe) copy them back in again" to take advantage of layer caching; not sure what the nix equivalent is

brendanhay:
The problem with doing this in nix is we still pay for the time it takes for the instance/container spin up + derivation evaluation + cachix downloads. Ignoring the paths for now is simpler + quicker, the event which would triggered all aforementioned nonsense is ignored.

ohAitch:
Is it not possible to download *just* a log of "the test result for this set of input files was X", without the actual binaries? Is container startup taking over ten seconds, and if so, what went wrong there?

ohAitch:
Ah I see looks like *installing nix* takes sometimes 1 minute, oof. And sometimes 5 seconds, do we know why this is highly variable? Do GH actions support baking a macOS image that's faster to retrieve than "run official nixos.org installer"?

brendanhay:
You can use custom [https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer](container images) I believe, not sure how far that extends torwards macOS support, though.

PR #4650, opened 2021-03-23
landscape: add Jest for testing

tomholford:
Felt a little dirty opening PRs without tests, so here's an effort to add Jest tests to Landscape. 

A trivial test case is included as a demo. While adding it, turns out there were some unrelated minor `tsc` bugs in the `utils` file, so those changes are also included.


tylershuster:
This would be nice. Target `release/next-js` though

tomholford:
@tylershuster Thanks for the feedback, done 👍

tomholford:
After rebasing, unable to run tests due to this type error:

image

There's a comment right above it ("// TODO why does this typing break?"), and I agree, no idea what's going on :)

tylershuster:
@liam-fitzgerald any thoughts on this typing?

tomholford:
Sharing stack trace since it appears it could be an upstream `zustand` issue:

image

liam-fitzgerald:
If you could use npm v6 to regenerate the lockfile such that we don't have a 30k line diff for this PR, that'd be fantastic

liam-fitzgerald:
bump @tomholford

tomholford:
> If you could use npm v6 to regenerate the lockfile such that we don't have a 30k line diff for this PR, that'd be fantastic

@liam-fitzgerald Just regenerated the lockfile with 6.14.13, now it is down to ~15K line diff. Thanks `npm` for letting me pad my contributor stats 😆

matildepark:
Superseded by https://github.com/urbit/urbit/pull/4933, feel free to PR unit tests separately.

PR #4651, opened 2021-03-24
notifications: fix DM redirect

liam-fitzgerald:
Weird zombie children issues, refactored to use props from parent
instead

Fixes urbit/landscape#595


matildepark (approved):
paired on a fix

PR #4652, opened 2021-03-24
chat: fix unread line setting

liam-fitzgerald:
- Fixes unread line setting
- Fixes loading indicator not disappearing

### DIFF
--- a/pkg/interface/config/webpack.dev.js
+++ b/pkg/interface/config/webpack.dev.js
@@ -82,7 +82,7 @@ if(urbitrc.URL) {
 }
 
 module.exports = {
-  mode: 'development',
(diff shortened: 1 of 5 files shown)


matildepark (approved):

pkg/interface/config/webpack.dev.js
```diff
@@ -82,7 +82,7 @@ if(urbitrc.URL) {
 }
 
 module.exports = {
```
tylershuster, line 84: Are we always running 'production' in dev now?

matildepark: -_-; I'll hotfix

PR #4654, opened 2021-03-24
ChatWindow: stop blue line moving

liam-fitzgerald:


matildepark:
Fixes the regression in https://github.com/urbit/landscape/issues/601

matildepark (approved):

PR #4655, opened 2021-03-24
interface: type annotations, imports

mrdomino:
cc @tylershuster


liam-fitzgerald (approved):
LGTM! Thanks for the types!

PR #4656, opened 2021-03-24
eyre: clean up connection in ++discard-channel

### LINKED ISSUE #4617: Channel connections linger in Eyre state
Happened to look at debug dashboard just now, and noticed I had 1216 connections open in eyre. Notably I was not running `urbit-king` (see also #4530), instead still on the `urbit` rc-1 that shipped with flagday.

I have since updated to latest vere, but can still reproduce connection creep by simply reloading my landscape page. The channel (in Eyre state) itself gets properly cleaned up, but its connection lingers.

Arguably Eyre should remove the connection from state, as it does for completion or error cases. But I think we managed to get `%cancel-request` tasks for these cases in the past, so I wonder what changed...

pkova:
Fixes #4617.

The next step is to do some `+load` magic in eyre to get rid of all the outstanding connections people have because of this bug.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1706,7 +1706,8 @@
       (snoc http-moves (set-heartbeat-move channel-id heartbeat-time))
     ::  +discard-channel: remove a channel from state
     ::
-    ::    cleans up state, timers, and gall subscriptions of the channel
+    ::    cleans up state, timers, gall subscriptions, and connection
+    ::    of the channel
     ::
     ++  discard-channel
       |=  [channel-id=@t expired=?]
@@ -1725,6 +1726,9 @@
               duct-to-key.channel-state
             ?.  ?=(%| -.state.session)  duct-to-key.channel-state.state
             (~(del by duct-to-key.channel-state.state) p.state.session)
+          ::
+              connections
+            (~(del by connections.state) p.state.session)
           ==
       =/  heartbeat-cancel=(list move)
         ?~  heartbeat.session  ~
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1725,6 +1726,9 @@
               duct-to-key.channel-state
             ?.  ?=(%| -.state.session)  duct-to-key.channel-state.state
             (~(del by duct-to-key.channel-state.state) p.state.session)
+          ::
+              connections
+            (~(del by connections.state) p.state.session)
```
Fang-, line 1731: I suspect we should also produce a `[p.state.session %give %response %cancel ~]` move to make sure the request is known as "cancelled by arvo" in the runtime as well.

Fang- (commented):
Sweet!

> The next step is to do some +load magic in eyre to get rid of all the outstanding connections 

I wouldn't say this is _strictly_ necessary, considering you can also clear these by restarting the urbit process.

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: Exciting! Perhaps we won't run out of file descriptors if we do this.

joemfb: vere no-ops on `%cancel`, and king-haskell doesn't even acknowledge the existence of the effect. We should have timeouts/process-limits such that the sockets and file descriptors are completely and correctly managed without Arvo's involvement. But implementing the effect could also spur such management.

/cc @pilfer-pandex

joemfb: In general, a runtime has to explicitly manage its resource utilization -- it can't depend on arvo to do that for it.

philipcmonk:
(didn't mean to close this, apparently github deletes PR branches even if other PRs are targeted to them and doesn't retarget the open PRs?)

PR #4658, opened 2021-03-25
ChatWindow: stop unread notice disappearing

liam-fitzgerald:


matildepark (approved):
Behaviour seems an improvement in a few rooms, about what is desired, let's try it over a longer period ...

PR #4662, opened 2021-03-25
interface: borders changed from washedGray to lightGray to match Figma mocks

tacryt-socryp:
I noticed an inconsistency between the Landscape JS and the Figma mocks. This may fix https://github.com/urbit/landscape/issues/633

Screen Shot 2021-03-25 at 4 19 05 PM
Screen Shot 2021-03-25 at 4 19 25 PM


urcades (approved):
Thank you for reading my mind, was going to produce an issue about this!

matildepark:
Should be next-js but whatev

PR #4663, opened 2021-03-25
interface: profile and settings light gray

tacryt-socryp:
Finalizing washedGray to lightGray change, realized I had missed reflecting the change across the profile and settings pages and the terminal / permalink embed.


urcades (approved):
Big Yes

PR #4668, opened 2021-03-26
push-hook: add version scries

liam-fitzgerald:
Allows airlock libraries and other local stores to negotiate versioning better


tacryt-socryp (approved):

matildepark:
(I merged this into release/next-userspace as well to keep them at parity.)

PR #4669, opened 2021-03-26
vere: add -b to specify http server bind ip

### LINKED ISSUE #4638: Can't specify an IP address to bind to
Urbit always binds its http ports promiscuously to 0.0.0.0. An address should be specifiable.

pkova:
Fixes #4638

Note that this prints `http: listen: address not available` when we get `UV_EADDRNOTAVAIL` and continues with no http server, should we crash instead?


joemfb:
If we can't bind the address we should crash. You can force exit anytime with `u3_king_bail()` (calling `exit()` directly leaves the terminal in raw mode).

pkova:
Done.

joemfb (approved):
LGTM

PR #4671, opened 2021-03-26
interface: unify back button

tylershuster:
fixes urbit/landscape#647

As you can see, it's not exactly like the mockups. The size of the caret is kind of small. @g-a-v-i-n any ideas on how to make this match the correct size? `size={2}` is too big

image

### DIFF
--- a/pkg/interface/src/views/apps/links/LinkResource.tsx
+++ b/pkg/interface/src/views/apps/links/LinkResource.tsx
@@ -16,6 +16,7 @@ import { Association } from '@urbit/api/metadata';
 import useGraphState from '~/logic/state/graph';
 import useMetadataState from '~/logic/state/metadata';
 import useGroupState from '../../../logic/state/group';
+import BackButton from '~/views/components/BackButton';
 
 const emptyMeasure = () => {};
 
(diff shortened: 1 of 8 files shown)


matildepark:
You can just give it a specific `px` value if you wanna fine tune it.

g-a-v-i-n:
Try giving the icon width="4" height="4" with no padding on the container.
That way the icon will be 24px and appear bolder because the icon scales
with size dimensions.
—
~ridlur-figbud
https://urbit.org

tylershuster:
image

urcades:
I had to run the PR locally to confirm why the sizing of the caret/icon seemed off with respect to the as-designed back buttons. The tl;dr here is that in each of the edits made, the back button is still being located within the body of the viewport itself, rather than within the header, which is where we'd like back buttons to live, when 'nested' in a channel object: 

Screen Shot 2021-03-29 at 5 22 06 PM

Screen Shot 2021-03-29 at 5 22 29 PM

Screen Shot 2021-03-29 at 5 21 58 PM

You can see that in each of the screens above, the icons are appearing thinner than our icons normally present because they are being based on the size of the text we've been using as back buttons in the view. If these carat-icon-buttons are located in the header, as they are designed at present, they should present themselves as-designed by Gav:

Screen Shot 2021-03-29 at 5 27 27 PM

urcades:
You can see me emphsize it in the second image, but eliminating the entire "back" element from the primary view of the channel object, and relocating the ability to "navigate back" as a header element, is more or less the ask here.

tylershuster:
Ah, this is definitely a sane pattern visually but not particularly sane in terms of code implementation. I will have to consider this more

urcades:
how is this item doing btw

PR #4672, opened 2021-03-26
launch: get tutorial state from zustand

matildepark:
The check for rendering the tutorial modal was relying on a passed prop that does not exist now.

(I also move all the hooks to the top of the function as per best practice.)


tacryt-socryp (approved):
looks fine

matildepark:
Caught a bug where `hasLoaded` is always `true` because we keep contacts in state, rendering the tutorial every load. I passed `state.connection` as a better heuristic.

PR #4673, opened 2021-03-26
group-feed: improve performance and make mobile load

tacryt-socryp:
Drastically cut down on renders by no longer passing contacts, associations, or graphResource down. Instead, uses useContactState lower down in component tree and passes graphPath, a string that has the graphResource data in it. Also adds a key to VirtualScroller that has the history.location.pathname so as to not have weird broken states.


matildepark:
RIP withMemo

tacryt-socryp:
Why did you render is dope. @jyng I fixed the mobile issue, so we can move forward with design QA stuff on Monday

matildepark:
New posts do not come in until next visit.

matildepark (approved):

PR #4674, opened 2021-03-27
king: scry on GET requests (WIP)

ryjm:
Preliminary work on handling GET requests to eyre through scrying. This PR contains an initial implementation in king and a dummy implementation on the hoon side. Would like some feedback on the haskell side, and any suggestions on what could be improved, both syntax wise and on the correctness of the approach. my knowledge of Wai and the subsite functionality had me a bit confounded.

I also made a first pass at moving the Eyre site intitialization in `Pier.hs`, which was mentioned a TODO. Don't think this is an ideal approach so open to critique. Keep in my mind this is a WIP - there are impacts to performance that haven't been succintly addressed yet.


ryjm:
@pilfer-pandex @Fang-

ryjm:
oops lemme squash these commits to make it a bit more readable

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2528,6 +2528,26 @@
         %rejected  ``noun+!>((~(has in rejected.cors-registry) u.origin))
       ==
     ::
+        [%scry @ @ ~]
+      =/  host  `i.t.tyl
+      =/  url  i.t.t.tyl
+      =/  server  (per-server-event [eny *duct now rof] server-state.ax)
+      =/  [=action suburl=@t]
+        (get-action-for-binding:server host url)
+      =/  invalid  (error-page 400 %.y url "invalid scry: {<action>} {<suburl>}")
```
Fang-, line 2537: I don't think you want to actually produce an error page anywhere. Instead you want to produce `~` or `[~ ~]` for the failure cases ("unknown/invalid scry path" and "no result" respectively). Giving a proper http-event result would make the runtime serve that up, but if the scry can't handle it then we just want to fall through to the eventful handling, right?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2528,6 +2528,26 @@
...
+      :^  ~  ~  %noun
+      !>  ^-  octs
```
Fang-, line 2539: At the very least you'll also want a `response-header` here. Might be best to just produce a `$>(%start http-event:http)` where `complete=&`, the runtimes already have logic for handling such nouns.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2528,6 +2528,26 @@
...
+      ?+    -.action  !!
+        %gen
+      invalid
+        %scry
+      (handle-scry %.y (need host) url)
```
Fang-, line 2544:
```suggestion
      ?+    -.action  !!
          %gen
        invalid
      ::
          %scry
        (handle-scry %.y (need host) url)
      ==
```

Alternatively, if they're gonna stay this short:

```suggestion
      ?+  -.action  !!
        %gen   invalid
        %scry  (handle-scry %.y (need host) url)
      ==
```

(With the caveat about not crashing of course, but this looks WIP anyway.)

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2567,4 +2587,72 @@
       ``[secure port host]
     ==
   ==
+++  return-static-data
```
Fang-, line 2590: Probably rename as `+build-http-event` and make `+return-static-data-on-duct` use it.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2567,4 +2587,72 @@
...
+    :*  %start
+        :-  status-code=code
+        ^=  headers
+          :~  ['content-type' content-type]
+              ['content-length' (crip (format-ud-as-integer p.data))]
+          ==
+        data=[~ data]
+        complete=%.y
+    ==
+++  handle-scry
```
Fang-, line 2602: We've already talked about deduplicating this with the `+request:per-server-event` logic, just leaving a note here so we don't forget!

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -1,71 +1,77 @@
...
-
-  - Requests come in, are given an identifier and are passed to a callback.
-
-  - When requests timeout, the identifier is passed to anothing callback.
-
-  - The server pulls response actions, and passes them to the associated
-    request.
--}
-
+{-# LANGUAGE LambdaCase #-}
```
Fang-, line 1: Looks like a linter mangled this file? Busted coms, indentation and other stylistic changes all over the place.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -159,76 +166,139 @@ dupHead = do
...
+  where
+    scryPath path = do
+      logDebug $ display ("scrying for " <> tshow path)
+      s <- scryOcts path
+      case s of
+        Just octs -> do
+          logDebug $ display ("scry result: " <> tshow octs)
+          pure $ unOcts octs
+        Nothing -> do 
+          logError $ display ("exception during scry on path: " <> tshow path)
```
Fang-, line 285: I'd probably log this as `logInfo` or even `logDebug`? If we're gonna try this for all requests, failure here is certainly common and expected. And we fall back, so it's not a hard error, right?

Fang- (commented):
You probably want to target `next-vere` for this. Also, GitHub lets you explicitly mark new PRs as WIP when you open them.

> there are impacts to performance that haven't been succintly addressed yet.

Yeah, I'd be hesitant about merging this into a release branch until we've investigated the performance side of this, as I wrote in the scholarship description.

I know you said it's a dummy implementation, but I left some comments on the Eyre side of this change to nudge you in the right direction. Mostly feedback around _what_ this should/n't be producing, and some written to-do's for refactoring we had already talked about previously.

As for the Haskell side, I'll mostly leave that up to @pilfer-pandex.

pkg/arvo/sys/vane/eyre.hoon
ryjm: yes, all of this was just to make sure it was getting to the eyre vane. if you `!!` here it will take the normal flow.

pkg/arvo/sys/vane/eyre.hoon
ryjm: yeah i'm in the process of figuring out why i'm getting empty scry results when i scry into other vanes through `per-server-event`. still crossing my fingers i can get my HD back because i can't quite remember how i fixed that.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
ryjm: ah i thought cleared those out. will fix.

ryjm:
> You probably want to target `next-vere` for this. Also, GitHub lets you explicitly mark new PRs as WIP when you open them.
> 
hm i didn't see that option. is it a label? can i set it after the fact? 

> > there are impacts to performance that haven't been succintly addressed yet.
> 
> Yeah, I'd be hesitant about merging this into a release branch until we've investigated the performance side of this, as I wrote in the scholarship description.
> 
> I know you said it's a dummy implementation, but I left some comments on the Eyre side of this change to nudge you in the right direction. Mostly feedback around _what_ this should/n't be producing, and some written to-do's for refactoring we had already talked about previously.
> 
yeah it's quite dumb, just wanted to show where this is coming in. 

> As for the Haskell side, I'll mostly leave that up to @pilfer-pandex.

ryjm:
tried this on a live ship, not noticeably slower afaict, even without explicitly handling failed scry's intelligently.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -199,36 +209,99 @@ sendResponse cb waitAct = do
...
-        atomically $ inform reqId $ ReqInfo adr met url hdr bod
-
-        try (sendResponse respond respApi) >>= \case
-          Right rr  -> pure rr
-          Left  exn -> do
-            atomically (cancel reqId)
-            logError $ display ("Exception during request" <> tshow exn)
-            throwIO (exn :: SomeException)
+    ("~_~" : _) -> runKingSubsite sub req respond
+    _ -> if getp then scryFlow else normalFlow
```
pilfer-pandex, line 240:
```suggestion
    ("~_~" : _) -> runKingSubsite sub req respond
    _ | met == H.GET -> scryFlow `onException` normalFlow
    _ -> normalFlow
  where
    normalFlow = runRIO env $
      rwith (liveReq who liv) $ \(reqId, respApi) -> do {- rest of normalSite body here... -}
    scryFlow = do
      env <- ask
      pure \req respond -> do
        let url = {- rest of scrySite body... -}
```

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -199,36 +209,99 @@ sendResponse cb waitAct = do
...
+  let getp = met == H.GET
+      scryFlow =
+        onException
+          (runEyreSite site req respond)
+          normalFlow
+      normalFlow =
+        runEyreSite
+          (normalSite env who liv inform cancel)
+          req
+          respond
```
pilfer-pandex, line 237:
```suggestion
```

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -199,36 +209,99 @@ sendResponse cb waitAct = do
...
+  e ->
+  Ship ->
+  TVar LiveReqs ->
+  (Word64 -> ReqInfo -> STM ()) ->
+  (Word64 -> STM ()) ->
+  EyreSite ->
+  KingSubsite ->
+  W.Application
+app env who liv inform cancel site sub req respond = do
+  met <- maybe (error "bad method") pure (cookMeth req)
```
pilfer-pandex, line 227: This doesn't have to be monadic

```suggestion
  let met <- maybe (error "bad method") id (cookMeth req)
```

and in fact you can even move it into the where block I suggest below, if you feel like it (or not)

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -199,36 +209,99 @@ sendResponse cb waitAct = do
...
+
+normalSite ::
+  HasLogFunc e =>
+  e ->
+  Ship ->
+  TVar LiveReqs ->
+  (Word64 -> ReqInfo -> STM ()) ->
+  (Word64 -> STM ()) ->
+  EyreSite
+normalSite env who liv inform cancel = ES $ \req respond ->
```
pilfer-pandex, line 250: As proposed above, move these into the where block in `app` so you don't have to take in all these arguments. Also get rid of the EyreSite type. Since we no longer need to pass these whole "sites" though the stack, we don't need a newtype to document them.

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Wai.hs
```diff
@@ -21,6 +21,9 @@ module Urbit.Vere.Eyre.Wai
   , rmLiveReq
   , newLiveReq
   , app
+  , EyreSite
+  , fourOhFourSite
+  , scrySite
```
pilfer-pandex, line 26:
```suggestion
```

pilfer-pandex (changes requested):

pkg/hs/urbit-king/lib/Urbit/Vere/Eyre/Serv.hs
```diff
@@ -76,17 +77,17 @@ data ReqApi = ReqApi
...
 
 instance Show ServType where
   show = \case
-    STHttp  who _ _      -> "STHttp "  <> show who
-    STHttps who tls _ _  -> "STHttps " <> show who <> " " <> show tls
-    STMultiHttp _ _      -> "STMultiHttp"
-    STMultiHttps tls _ _ -> "STMultiHttps"
+    STHttp  who _ _ _      -> "STHttp "  <> show who
+    STHttps who tls _ _ _  -> "STHttps " <> show who <> " " <> show tls
+    STMultiHttp {}      -> "STMultiHttp"
```
ryjm, line 89: #
```suggestion
    STMultiHttp {}      -> STMultiHttp _ _
```

ryjm:
out of curiousity - where are the style guidelines? i just used the stylish \ haskell thing in our repo, so i'm assuming that's wrong? is it sh/fmt-haskell?

zalberico:
As it is, I'm going to close this. If it becomes something we want to focus on and we need to reopen it we can.

PR #4677, opened 2021-03-28
sys: Speed up things used in Landscape

philipcmonk:
Four separate changes aimed at improving landscape initial load time, with notes in the individual commit messages.  I also determined that we're compiling marks on every refresh.  In my opinion, we should keep all marks "warm", but in the meantime you can run `-test %/mar ~` to warm them up until they're invalidated by a dependency changing (for example zuse).

Combined, these changes take my initial landscape load from 28 seconds to 6.3 seconds (under profiling; a bit faster in real-world).  Still far too long!

With marks warmed up (by running `-test %/mar ~`) it takes 12 seconds even without any other changes, so that's the most important thing.  The changes in this PR cut that time in half again.

The majority of the remaining time (3.5 seconds) is spent in mark conversion to json (largely in `(scot %p)`), and a significant other portion (1.8 seconds) is spent in +en-json.  See:
Screen Shot 2021-03-28 at 1 14 59 AM

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -5455,9 +5455,10 @@
   ~%  %co  ..co  ~
   =<  |_  lot=coin
       ++  rear  |=(rom=tape rend(rep rom))
-      ++  rent  `@ta`(rap 3 rend)
+      ++  rent  ~+  `@ta`(rap 3 rend)
       ++  rend
         ^-  tape
+        ~+
         ?:  ?=(%blob -.lot)
           ['~' '0' ((v-co 1) (jam p.lot))]
         ?:  ?=(%many -.lot)
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1597,6 +1597,7 @@
     ::  +channel-event-to-sign: attempt to recover a sign from a channel-event
     ::
     ++  channel-event-to-sign
+      ~%  %eyre-channel-event-to-sign  ..part  ~
       |=  event=channel-event
       ^-  (unit sign:agent:gall)
       ?.  ?=(%fact -.event)  `event
@@ -1677,6 +1678,7 @@
       ==
     ::
     ++  event-json-to-wall
+      ~%  %eyre-json-to-wall  ..part  ~
       |=  [event-id=@ud =json]
       ^-  wall
       :~  (weld "id: " (format-ud-as-integer event-id))
@@ -2094,6 +2096,7 @@
 ~%  %http-server  ..part  ~
 |%
 ++  call
+  ~/  %eyre-call
   |=  [=duct dud=(unit goof) wrapped-task=(hobo task)]
   ^-  [(list move) _http-server-gate]
   ::
@@ -2296,6 +2299,7 @@
   ==
 ::
 ++  take
+  ~/  %eyre-take
   |=  [=wire =duct dud=(unit goof) =sign]
   ^-  [(list move) _http-server-gate]
   ?^  dud
@@ -2483,6 +2487,7 @@
 ::  +scry: request a path in the urbit namespace
 ::
 ++  scry
+  ~/  %eyre-scry
   ^-  roon
   |=  [lyc=gang car=term bem=beam]
   ^-  (unit (unit cage))
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -425,10 +425,12 @@
   :-  ~
   %-  as-octs:mimes:html
   %-  crip
-  %-  zing
+  %-  zing  ^-  ^wall
+  %-  zing  ^-  (list ^wall)
   %+  turn  wall
   |=  t=tape
-  "{t}\0a"
+  ^-  ^wall
+  ~[t "\0a"]
 ::  +internal-server-error: 500 page, with a tang
 ::
 ++  internal-server-error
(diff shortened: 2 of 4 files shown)


pkg/arvo/sys/zuse.hoon
```diff
@@ -3286,7 +3286,7 @@
     ++  ship                                            ::  string from ship
       |=  a=^ship
       ^-  json
-      (tape (slag 1 (scow %p a)))
+      [%n (rap 3 '"' (scot %p a) '"' ~)]
```
Fang-: json-string nouns are `[%s @t]`, so couldn't this just be `[%s (scot %p a)]`?

But this changes the result, the original discards the leading `~` for whatever reason. So, `[%s (rsh 3^1 (scot %p a))]` perhaps?

Fang-: The follow-up commit (which I should've seen before commenting) addresses the `~`, but the comment about `%s` remains!

philipcmonk: +en-json's performance is dominated by +jesc -- escaping strings -- and landscape's initial load is mostly ship names, which never need to be escaped.  Using %n lets us skip that while printing.  I got about 1 second better load time with this change.

tacryt-socryp:
Interesting, so the vast majority of the improvement came from the marks. Are there plans to keep all marks warm? If not, could Eyre warm them up (by sending a %sing the first time it sends a mark conversion scry for any particular conversion) in the same way graph-store does?

tacryt-socryp:
Worth noting that `+of-wall:format` uses the same slow rendering pattern as the Eyre method (but for performing the inverse operation), though I'm not sure how widely it's used.

matildepark:
Your landscape load has taken 28 seconds this whole time? I’m intrigued by this since our time to paint is instant and the remainder of my own subscriptions come in within a few seconds.

pkg/arvo/sys/zuse.hoon
Fang-: Oh my bad, I missed that in the commit message. Still feels a bit weird, but it does the job for now. Carry on!

philipcmonk:
> Your landscape load has taken 28 seconds this whole time?  I’m intrigued by this since our time to paint is instant and the remainder of my own subscriptions come in within a few seconds.

Yes.  Well, more like 20 seconds when not under the profiler.  Interesting, I assumed everyone was living with my pain.  I'm in 69 groups and a bunch of DMs, so my guess is that I just have a lot of data being downloaded (>1 MB of json on refresh)

> Interesting, so the vast majority of the improvement came from the marks. Are there plans to keep all marks warm? If not, could Eyre warm them up (by sending a %sing the first time it sends a mark conversion scry for any particular conversion) in the same way graph-store does?

The way to keep marks warm is to have or fake an open subscription to them on `%ca`.  You could do this in clay, but honestly it would cleaner to just use an app for it, either an existing one or a new one.  It could have a hardcoded list of marks or just traverse and get all of them every time `%cz` on /mar changes.  This is no more than a couple hours of work, but I doubt I'll get to it soon.

tacryt-socryp (approved):

tacryt-socryp:
Would be really nice not to hardcode it and have it do a traversal... Maybe a nice contributor task? https://github.com/urbit/urbit/pull/4677#issuecomment-809566863

pkg/arvo/sys/hoon.hoon
```diff
@@ -5455,9 +5455,10 @@
   ~%  %co  ..co  ~
   =<  |_  lot=coin
       ++  rear  |=(rom=tape rend(rep rom))
-      ++  rent  `@ta`(rap 3 rend)
+      ++  rent  ~+  `@ta`(rap 3 rend)
       ++  rend
         ^-  tape
+        ~+
```
joemfb, line 5461: I'm a little hesitant to have both of these arms memoized; it seem like it could be a lot of churn in the cache. I'm also curious to see the impact of #4692 on these numbers. (Also, we have a space leak on arvo reset, so I really want to avoid frequent hoon/arvo changes. But that has to be fixed regardless.)

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -425,10 +425,12 @@
...
   %-  as-octs:mimes:html
   %-  crip
-  %-  zing
+  %-  zing  ^-  ^wall
+  %-  zing  ^-  (list ^wall)
   %+  turn  wall
   |=  t=tape
-  "{t}\0a"
+  ^-  ^wall
+  ~[t "\0a"]
```
joemfb, line 433: It seems like a `+weld` here might be more efficient that a `+zing` later. But it probably doesn't matter compared to the impact of removing the tape interpolation.

joemfb (commented):
Sorry for the delay. The changes themselves all look fine, but I have a couple hesitations about releasing them. I'll follow up out-of-band.

joemfb (approved):
LGTM (with the caveat that we may want to hold off on deploying the `hoon.hoon` changes until there's other stuff that needs to go too).

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: Can you explain the heuristic you're using from a performance perspective of when it's better to use `+zing` vs `+weld` ?

pkg/arvo/sys/hoon.hoon
ohAitch: In particular, these are frequently called on dates, which are new per event. Not the ideal cache candidate.

pkg/arvo/sys/vane/eyre.hoon
ohAitch: Isn't most of the impact of the tape interpolation the weld? The zing seems better because it'll traverse the `t` list only once.

tacryt-socryp:
is this ready to hit next-sys?

PR #4679, opened 2021-03-29
next-userspace grabbag vol. 2

liam-fitzgerald:
post: do not notify on top level content

interface: support %started update

group-view: add %started update

Invites: fix prop passing

Fixes urbit/landscape#654

interface: version serialized state

collections: update link copy

Fixes urbit/landscape#615

### DIFF
--- a/pkg/arvo/app/group-view.hoon
+++ b/pkg/arvo/app/group-view.hoon
@@ -193,6 +193,11 @@
       (~(put by joining) rid [%.n now.bowl ship %start])
     =.  jn-core
       (jn-abed rid)
+    =.  jn-core
+      %-  emit
+      %+  fact:io
+        group-view-update+!>([%started rid (~(got by joining) rid)])
+      ~[/all]
     ?<  ~|("already joined {<rid>}" (has-joined rid)) 
     =.  jn-core
       %-  emit
--- a/pkg/arvo/lib/group-view.hoon
+++ b/pkg/arvo/lib/group-view.hoon
@@ -54,7 +54,15 @@
     ?-  -.upd
       %initial  (initial +.upd)
       %progress  (progress +.upd)
+      %started   (started +.upd)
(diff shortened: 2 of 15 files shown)


pkg/interface/src/logic/state/metadata.ts
```diff
@@ -51,7 +51,7 @@ const useMetadataState = createState<MetadataState>('Metadata', {
   //     });
   //   });
   // },
-});
+}, [], 2);
```
matildepark, line 54: A lot of potential deploy overhead on this approach! Having to increment the version each time we edit. Why can’t we just inject something each build? On livenet this is rare, on devnet it will just mean inconsistent time to paint.

matildepark (approved):
Reverted the dev cruft, js seems fine

tacryt-socryp (approved):
backend looks good. Please change the version stuff in the JS to wipe local storage at every version hash change.

matildepark:
Dropped the commit so we can come back again and re-approach JS state

liam-fitzgerald:
@tacryt-socryp this requires us to block on deserialisation until we've loaded the new base hash. the primary reason for state serialisation is to improve load times. Given that we're already using version numbers on the frontend in the API calls, how do you feel about piggybacking those instead?

tacryt-socryp:
I think we should be more aggressive with clearing this data than just clearing it upon the mark version changing. For instance, many people still have the phantom "message from the future" in the UC Design chat because it's in their localStorage even though it isn't on the server. Perhaps if we both cleared localStorage upon the mark being upgraded AND cleared it once a week I'd like it more, but even then, is it an acceptable UX to continue displaying deleted data for a week just because it's cached? I'm not sure that it is. Maybe if we had a good solution to reflecting deletes into the localStorage we could solely clear it upon mark version upgrades.

PR #4680, opened 2021-03-29
collections: allow %reference as link

liam-fitzgerald:
Allows permalinks to be added to collections.

Capture d’écran 2021-03-29 à 12 41 49 PM

Fixes urbit/landscape#661

### DIFF
--- a/pkg/arvo/mar/graph/validator/link.hoon
+++ b/pkg/arvo/mar/graph/validator/link.hoon
@@ -52,7 +52,7 @@
         ::  top-level link post; title and url
         ::
         [@ ~]
-      ?>  ?=([[%text @] [%url @] ~] contents.p.ip)
+      ?>  ?=([[%text @] $%([%url @] [%reference *]) ~] contents.p.ip)
       ip
     ::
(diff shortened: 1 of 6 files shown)


matildepark:
Why does this require a validator change? You could just check for the URL like any other and convert it to a render. Collections are links. Let’s not change that.

tacryt-socryp:
No, we absolutely must use the type system here.

matildepark:
If you insist ... will this produce backwards incompatibility issues if someone adds a permalink and the subscriber doesn’t have this change? Presumably so. Not ideal

tacryt-socryp:
Yes, it would produce backward incompatibility, but the worst case is just a missing link on someone's old / incompatible ship. The missing link issue could be resolved simply by incrementing the versions in graph-push/pull hook, which will prompt a resubscribe.

matildepark:
Well as long as the version bump we're already doing for adding references is here, this is considered to be part of that new version, and then we're already covered.

matildepark (approved):
Reads sanely enough

PR #4683, opened 2021-03-29
build: match on path subdirectories

### LINKED ISSUE #4682: build: CI now never runs at all
My own instigation, I imagine — PRs and pushes that touch pkg/arvo aren't processing CI, when they should be.

Are we supposed to follow paths with a more loose `/*` syntax?

https://github.com/urbit/urbit/blob/a74bdfea20f9cee4e4e99afa0c9092c4f54f4256/.github/workflows/build.yml#L35-L57

matildepark:
Fixes #4682 though unsure how to confirm this.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,26 +35,26 @@ name: build
 on:
   push:
     paths:
-      - 'pkg/arvo'
-      - 'pkg/docker-image'
-      - 'pkg/ent'
-      - 'pkg/ge-additions'
-      - 'pkg/hs'
-      - 'pkg/libaes_siv'
-      - 'pkg/urbit'
-      - 'bin'
-      - 'nix'
+      - 'pkg/arvo/**'
+      - 'pkg/docker-image/**'
+      - 'pkg/ent/**'
+      - 'pkg/ge-additions/**'
+      - 'pkg/hs/**'
+      - 'pkg/libaes_siv/**'
+      - 'pkg/urbit/**'
+      - 'bin/**'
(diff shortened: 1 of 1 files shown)


jtobin (approved):
LGTM.  Sorry for not catching this previously; I don't really "know" GitHub Actions syntax per se.. 😄 

To test this sort of thing you should be able to just stage an intermediate commit that includes (I guess) `.github/**` in order to trigger the action, then just add another one that deletes it.

jtobin:
A note: I think `paths-ignore` might actually work better here.  From the docs:

> When using the push and pull_request events, you can configure a workflow to run when **at least one file does not match paths-ignore** or at least one modified file matches the configured paths. Path filters are not evaluated for pushes to tags.

and from the example ignoring paths:

> When **all** the path names match patterns in paths-ignore, the workflow will not run.

So it seems like pushes/PRs that touch e.g. pkg/arvo *and* pkg/interface will still be built, if pkg/interface is ignored, but those that exclusively touch pkg/interface will not be.  Which is IIRC what's desired?

matildepark:
Depends what you think the monorepo will become; we'll have to ignore pkg/npm, too. It seemed preferable to me to be specific and _positive_ about conditions that prompt CI for building vere and arvo than saying "everything _but_" and having to continually change it should more packages enter the repo that don't need to prompt a vere / arvo build needlessly. Having submodules with specific positive conditions requires less maintenance and is simpler to read.

jtobin:
Yeah you make a good case for it.  Positivity ho!

PR #4684, opened 2021-03-29
landscape: don't build serviceworker in dev

matildepark:
This is tripping us up like the old days — serviceworker caches the JS so that changes stick across hot reloads, making it hard to tell whether changes are being reflected as we work. We just don't build the serviceworker for dev.

If you are developing on Safari, you will have to run the following after this is on next-userspace:

```
navigator.serviceWorker.getRegistrations().then(function(registrations) {
 for(let registration of registrations) {
  registration.unregister()
} })
```

cc @tylershuster

### DIFF
--- a/pkg/interface/src/register-sw.js
+++ b/pkg/interface/src/register-sw.js
@@ -1,6 +1,4 @@
-
-
-if ("serviceWorker" in navigator) {
+if ("serviceWorker" in navigator && process.env.NODE_ENV !== 'development') {
   window.addEventListener("load", () => {
     navigator.serviceWorker.register("/~landscape/js/bundle/serviceworker.js", {
       scope: "/",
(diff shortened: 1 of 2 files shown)


tacryt-socryp (approved):

pkg/interface/src/register-sw.js
```diff
@@ -1,6 +1,5 @@
 
-
-if ("serviceWorker" in navigator) {
+if ("serviceWorker" in navigator && process.env.NODE_ENV === 'production') {
```
matildepark, line 2: Actually, I wonder — will this proc on a build?

matildepark: If I console.log NODE_ENV I do get `development` from within the file, but in prod we aren't in a node session, are we?

PR #4688, opened 2021-03-30
Post: add relative timestamps, refactor Timestamp component

liam-fitzgerald:
Fixes urbit/landscape#671


matildepark (changes requested):
This breaks timestamps in Publish and Links comments and posts.

image

image

PR #4689, opened 2021-03-30
ChannelSettings: unify form handling

liam-fitzgerald:
- Adds a formgroup component for the manipulation of several forms at once.
- Uses this to unify the disparate forms in channelsettings
- Worth noting that forms that are part of a formgroup retain their composability, i.e. you can still use them outside of the formgroup context if you'd like

Fixes urbit/landscape#534


matildepark (changes requested):
The original mock has a disabled button when we haven't made changes:



But instead we just don't render any of the buttons until we make a change. We should match mock.

image

matildepark (changes requested):
From @urcades: 

> The "Cancel" button would effectively be a way to close the window while keeping all forms unedited + not submitted.

The default open state here has both disabled:

image

matildepark (approved):

PR #4691, opened 2021-03-30
group-store: remove empty tags

liam-fitzgerald:
Fixes urbit/landscape#568

### DIFF
--- a/pkg/arvo/app/group-store.hoon
+++ b/pkg/arvo/app/group-store.hoon
@@ -414,11 +414,8 @@
       ?>  ?&  ?=(~ (~(dif in ships) members))
               (~(has by tags) tag)
           ==
-      %=    +<
-        ::
-          tags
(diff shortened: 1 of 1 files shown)


matildepark:
@tacryt-socryp Able to test this today? Might be good to sneak this out tonight since it's basic functionality (graph permissions) not functioning if you make a room "host only", but then change it to "all group members" later.

matildepark:
Moving this to pre-release/next-userspace, please review tomorrow @tacryt-socryp

tacryt-socryp (changes requested):
This manual list conversion + iteration on a jug seems really bespoke and iffy. Why not just use a map method to get the set directly and dif it?

liam-fitzgerald:
because that would leave an empty set, which is what this is trying to address

——
~hastuc-dibtux

https://urbit.org

tacryt-socryp (approved):

PR #4693, opened 2021-03-30
graph-push-hook: add hc where appropriate and enforce type-checking to prevent issue in future

tacryt-socryp:
Documenting the work done with Phil to both fix the `read-at-aeon` error occurring in `%graph-push-hook` and prevent it in the future by more rigorous type-checking. This works as is and the pattern should be replicated across all apps that use `=<`

### DIFF
--- a/pkg/arvo/app/graph-push-hook.hoon
+++ b/pkg/arvo/app/graph-push-hook.hoon
@@ -24,20 +24,20 @@
   state-zero
 --
 ::
-=|  state-zero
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (approved):
Oh man this pattern is gonna save me some annoyances. LGTM!

matildepark:
Cherrypicked this into pre-release as well as per @liam-fitzgerald — cc @tacryt-socryp

PR #4694, opened 2021-03-30
graph-validators: speed up validators by adding type casts

tacryt-socryp:
All validators (and uses of `!>` and `!<`) should have type-casts to avoid expensive nest-checking.

Please put into `prerelease` @matildepark


liam-fitzgerald (approved):
LGTM! Any ideas on the speedup numbers here?

matildepark:
Cherry-picked into pre-release.

PR #4698, opened 2021-03-31
group-feed: tons of post item touch-ups

tacryt-socryp:
Screen Shot 2021-03-30 at 9 33 52 PM

### DIFF
--- a/pkg/interface/src/views/components/Timestamp.tsx
+++ b/pkg/interface/src/views/components/Timestamp.tsx
@@ -11,10 +11,20 @@ export type TimestampProps = BoxProps & {
   stamp: MomentType;
   date?: boolean;
   time?: boolean;
+  relative?: boolean;
 };
 
 const Timestamp = (props: TimestampProps): ReactElement | null => {
-  const { stamp, date, time, color, fontSize, ...rest } = {
+  const {
+    stamp,
+    date,
+    time,
+    color,
+    relative,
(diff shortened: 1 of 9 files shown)


pkg/interface/src/views/components/Timestamp.tsx
```diff
@@ -24,14 +34,24 @@ const Timestamp = (props: TimestampProps): ReactElement | null => {
   const { hovering, bind } =
     date === true ? { hovering: true, bind: {} } : useHovering();
   let datestamp = stamp.format(DateFormat);
-  if (stamp.format(DateFormat) === moment().format(DateFormat)) {
-    datestamp = 'Today';
-  } else if (
-    stamp.format(DateFormat) === moment().subtract(1, 'day').format(DateFormat)
-  ) {
-    datestamp = 'Yesterday';
+  if (!dateNotRelative) {
```
liam-fitzgerald, line 37: This is probably better handled with `moment.calendar`, instead of special casing the format string like this

liam-fitzgerald (approved):
Nit about moment formatting, else LGTM!

pkg/interface/src/views/components/Timestamp.tsx
matildepark: Does this conflict with your time stamp pr? I like its style because it lets us expose relative time as a setting ppl can choose later.

PR #4699, opened 2021-03-31
interface: performance enhancement

liam-fitzgerald:
- Adjusts reducers to be compatible with immer.js
- Removes redundant unconditional this.setState everytime we get a subscription update 
- Makes interface go fast

cc: @tylershuster


matildepark:
Ooo I’m excited

PR #4702, opened 2021-03-31
ViewProfile: clone contact groups array

matildepark:
Fixes urbit/landscape#682.

For whatever reason we can't mutate these arrays in place when using immer. See [this StackOverflow thread](https://stackoverflow.com/questions/53420055/error-while-sorting-array-of-objects-cannot-assign-to-read-only-property-2-of/53420326), this fixes the crash by copying the array we're rendering.


matildepark:
This is also the only place in the codebase that we were actually doing this (all other `.sort` is doing it on `Object.keys` or etc, making a new copy). But this is also sorta troubling about the previous behaviour — we were _mutating state on render_?

tylershuster:
image

tacryt-socryp (approved):
LGTM

PR #4704, opened 2021-03-31
build: only run 'push' build on release branches

### LINKED ISSUE #4657: build: CI triggers overlap, and run twice
If you push a change on a PR, it will run the PR and the push change, spawning 8 instances when we only need 4, clogging the queue.

matildepark:
Part of the increasingly inaccurately named "CI specificity" trilogy.

If someone pushes a commit onto an open PR, the CI runs twice: once for the push trigger, once for the pull request trigger. To remedy this, we now only run the push CI when pushing to a release/, na-release/, or pre-release/ branch.

cc @joemfb — how often do we want builds to process? Pre-existing behaviour was "any commit on any branch," which may have been intentional, but kept the queue busy.

Fixes #4657

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,6 +34,10 @@ name: build
 
 on:
   push:
+    branches:
+      - 'release/**'
+      - 'na-release/**'
+      - 'pre-release/**'
     paths:
       - 'pkg/arvo/**'
       - 'pkg/docker-image/**'


matildepark:
Something I don't understand: are the branches and paths declarations here an "and" or an "or" check? Will _any_ of them trigger? Presumably yes given pre-existing behaviour triggering twice, in which case, pushes to release/next-js will trigger, and so we should be _more specific_ about the branches here?

Should I add release/next-js as an ignored branch?

joemfb:
I'd prefer to keep pushes, outside of specifically ignored branches. Lots of work gets done without an open PR, and, at least for arvo and runtime changes, it's good to flag issues early.

Is the build queue backing up to the point where it's blocking active work? Is that related to npm deploys, glob, or something else?

matildepark:
Not unless we are really up against the wall, but I'd like to add CI for interface next with type checking and all that fun stuff, so I'm trying to prevent unnecessary or redundant builds as I go. The case that specifically I want to avoid is the "pushing to a PR causes duplicate builds" mentioned here because commits often come in fast there and clog things up that way.

We can specifically ignore branches if you like — but the paths changes from before clean up the most egregious backup in our queue, so it may be moot if you want on-push retained.

ohAitch:
The [thread](https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662/13) about this suggests [`actions/skip-duplicate-actions`](https://github.com/marketplace/actions/skip-duplicate-actions#skip-concurrent-workflow-runs).

jtobin:
> Something I don't understand: are the branches and paths declarations here an "and" or an "or" check? Will any of them trigger?

Yeah, IIRC they're "or," in that a change to any of those paths will trigger the workflow.  For `paths_ignore`, workflows won't trigger only when *all* of the changes are on the listed paths.

> Should I add release/next-js as an ignored branch? Wait I can't.

I *think* you can still do this via [positive/negative patterns](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-using-positive-and-negative-patterns).  À la:

```
on:
  push:
    branches:    
    - 'release/**'
    - '!release/**-js'
    - 'na-release/**'
    - 'pre-release/**'
```

That's how I understand the description given there anyway.

jtobin:
As a tangential note, GitHub seems to have [recently added support](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/) for Travis-like skipping of workflows.  Worth remembering to tack a `[skip ci]` to one's commit message when touching READMEs and the like.

joemfb:
@ohAitch, thanks! I remember seeing that and meaning to investigate (and then not doing so). If we can transparently remove duplicate test runs, that would seem to be best. (Note that push and pr are often not the same, the latter is the result of a merge).

matildepark:
Closing in favour of another swing with @ohAitch's suggestion.

PR #4705, opened 2021-04-01
graph-push-hook: don't crash on multipart add

liam-fitzgerald:


tacryt-socryp (approved):
This looks good to me. Strange that it came up when it did. Have you tested it yet?

tacryt-socryp:
... I think this won't work if it's nested deeper than one

liam-fitzgerald:
It will work because the nested would require each immediate parent to be in the poke

tacryt-socryp:
But it won't be in the top-level nodes map that you're doing the ~(has by) on

liam-fitzgerald:
But we don't recurse, permissions are only checked on the root

liam-fitzgerald:
Honestly, I would just declare that behaviour as unsupported, given we don't use it. I'd like to revisit the semantics of this flat vs nested add-nodes poke in future though.

tacryt-socryp:
No, permissions are checked on any node that is included in the map as a top-level node.

You could have any of the following entries in the map and it would be valid:

/a containing /a/b

/a containing /a/b containing /a/b/c

/a
/a/b containing /a/b/c


and even

/a
/a/b
/a/b/c

as separate top-level items in the map

liam-fitzgerald:
It would work for the third? 
Check a -> fine
check b containing c -> skip because we have a

PR #4709, opened 2021-04-01
group-feed: implement permissions properly

tacryt-socryp:
Have done API and API actions. Also did prop-drilling. Now to implement the different render states.

EDIT: Ready to go.


matildepark (approved):
Reads to me. We'll sanity check on testnet.

PR #4715, opened 2021-04-02
push-hook: nack unversioned subscriptions when unsupported

liam-fitzgerald:
Mainnet is currently experiencing undue load due to a watch-kick loop.
If the watch is unsupported, it will be kicked due to a mark mismatch,
which will be interpreted by the unversioned hooks as a network pressure
kick, prompting a resub and continuing the loop. Instead, we now no-op
on unversioned, unsupported watches, waiting for the subscriber to
rewatch after it processes the versioning OTA. This will silently break
groups for the subscriber until the reach the latest OTA, but is
preferable to nacking, and kicking them permanently

### DIFF
--- a/pkg/interface/src/views/apps/permalinks/embed.tsx
+++ b/pkg/interface/src/views/apps/permalinks/embed.tsx
@@ -23,6 +23,7 @@ import { Link } from "react-router-dom";
 import useGraphState from "~/logic/state/graph";
 import { GraphNodeContent } from "../notifications/graph";
 import { TranscludedNode } from "./TranscludedNode";
+import {useVirtualResizeProp} from "~/logic/lib/virtualContext";
 
 function GroupPermalink(props: { group: string; api: GlobalApi }) {
   const { group, api } = props;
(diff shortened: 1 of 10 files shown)


pkg/interface/src/views/landscape/components/GroupSettings/GroupFeed.tsx
```diff
@@ -66,7 +66,7 @@ export function GroupFeedSettings(props: {
               </Label>
             </Col>
           </BaseLabel>
-          {isEnabled && false && (
+          {isEnabled && (
```
matildepark, line 69: We put this in last night because it doesn’t handle having permissions changed. Does this work now? I’ll try, otherwise this should drop

liam-fitzgerald: It does work now

pkg/interface/src/views/apps/permalinks/embed.tsx
```diff
@@ -61,13 +62,15 @@ function GraphPermalink(
...
     ])
   );
+
+  useVirtualResizeProp(node)
   useEffect(() => {
     (async () => {
       if (pending || !index) {
         return;
       }
       try {
```
matildepark, line 72: Why drop the await?

liam-fitzgerald: good catch!

tacryt-socryp (approved):
Nooping instead of nacking in this case is righteous

PR #4716, opened 2021-04-02
extras: emacs syntax highlighting

dasrun-fadben:
From the style guide:

> Modern Hoon naming is *verbose*.

As such, hoon tends to have hyphenated names, but these names are only partially matched by the current font-lock regex.


drbeefsupreme:
I think this belongs in https://github.com/urbit/hoon-mode.el

dasrun-fadben:
Ah, it probably would, but it looks like that version of hoon mode has progressed beyond what's here. Thanks for pointing it out.

PR #4720, opened 2021-04-02
metadata-store lib: make config encode/decode invertible

tacryt-socryp:


matildepark:
This doesn't handle the "I dismissed the group feed notice" case, metadata editing still crashes.

matildepark (approved):
Tested on various cases and the logic change is simple and fine

PR #4722, opened 2021-04-03
pull-hook: noop on unsupported, unversioned facts

liam-fitzgerald:
Pull-hook side of the previous change to push-hook
(479fbfd). Noop on unsupported
unversioned marks, waiting for the publisher to upgrade and kick. Same
tradeoffs as previous change apply. Restores the historical marks to their correct types so that clamming will no longer fail


philipcmonk (approved):
Tried this on my ship that had this issue.  The crashes stopped, and the subscriptions are still present, so it looks like this worked.

PR #4724, opened 2021-04-03
groups: resize group summary

jamesacklin:
Adds a manual but indigo-appropriate (indigo "8" x 3 = 288px) max-width to the group message such that it wraps nicely and is centered in the main pane.

image

fixes urbit/landscape#728


liam-fitzgerald (approved):
Looks good, wondering if the groupsummary should always have this max-width cc: @urcades

urcades:
@liam-fitzgerald this seems like a reasonable size to maintain for all viewports besides mobile ones — we can probably opt for a `width: 100%` (or something)  pending proper padding being applied to the left and right edges of the element.

liam-fitzgerald (changes requested):
Please apply max-width inside the groupsummary component

tacryt-socryp (approved):

jamesacklin:
Totally makes sense to do -- just wondering how to handle this in the downstream case of the group-join modal? Happy to set a max-width on this modal container for large viewports, but this affects a host of other things as well.

image

liam-fitzgerald (approved):

PR #4725, opened 2021-04-03
interface: redesign ImageInput

jamesacklin:
Redesigns the vanilla ImageInput component a bit by adding a bit of clickable text in the placeholder, surfacing an uploading status, removing the "Upload" button, and allowing the user to recover from upload failures.

Default appearance (if storage is configured):
image

The user can click the blue text, is prompted to upload an image, and when the selected image is uploading:
image

Error state:
image

The user can click the blue text to select another image or try again.

Completed upload or URL pasted:
image

This work was intended to address the header image in the Profile editor, but has knock-on effects elsewhere (Group image, Landscape background image, Profile avatar).

This also addresses the unintentional truncation of this placeholder text in the Landscape background Settings.

Fixes urbit/landscape#695, urbit/landscape#581


tylershuster:
Does this maintain the ability to paste an image from the user's clipboard?￼

jamesacklin:
@tylershuster I'm not too tremendously sure -- I just tried to paste image binary data into an ImageInput field on livenet; is there a secret to invoking that? AFAICT that only works in the chat Message field...

If it's not a regression, that's certainly an area for improvement (along with actual drag-and-drop).

pkg/interface/src/views/components/ImageInput.tsx
```diff
@@ -55,25 +97,21 @@ export function ImageInput(props: ImageInputProps): ReactElement {
           {caption}
         </Label>
       ) : null}
-      <Row mt="2" alignItems="flex-end">
+      <Row mt="2" alignItems="flex-end" position='relative'>
+        {!field.value && !uploading && meta.error === undefined ? (<Prompt />) : null}
+        {uploading && meta.error === undefined ? (<Uploading />) : null}
```
liam-fitzgerald, line 102: Calling closures as components is bad, because components are ideally referentially stable. Because referential equality is used in react's diffing algorithm, this causes the DOM elements to be removed and readded *on every render*. It's better to just call them as a regular function, as they won't be inserted into the component tree and will have no effect on the diffing. e.g. `{errorRetry()}`

pkg/interface/src/views/components/ImageInput.tsx
```diff
@@ -43,10 +40,55 @@ export function ImageInput(props: ImageInputProps): ReactElement {
...
+      left={2}
+      top={2}
+      style={{ pointerEvents: 'none' }}
+    >
+      Paste a link here, or{' '}
+      <Text
+        fontWeight='500'
+        cursor='pointer'
+        color='blue'
+        style={{ pointerEvents: 'all' }}
```
liam-fitzgerald, line 61: just curious, wat mean?

liam-fitzgerald (changes requested):
Please remove the use of closures as components. Else LGTM, looks really nice!

tylershuster (approved):
I see now that this is a different input than the one that uses the drag logic. 

If you feel like you're bursting with chi and you don't want to let it stagnate, the logic to add dragging is pretty simple:

```jsx
import { useFileDrag } from '~/logic/lib/useDrag';

export function ImageInput(props) {
  const { uploadDefault, canUpload, uploading } = useStorage();
  ...
  const dragCallback = useCallback((files: FileList | File[]) => {
    try {
        const url = await uploadDefault(Array.from(files)[0]);
        setValue(url);
      } catch (e) {
        setError(e.message);
      }
  }, [uploadDefault, canUpload, uploading])
  const { bind, dragging } = useFileDrag(dragCallback);
  return <Box {...bind} style={{ background: dragging ? 'whatever' : 'default' }}>...</Box>
}
```

For more examples see the other 3 uses of `useDrag` in the codebase.

tylershuster:
Furthermore, here's a succinct example of handling paste ripped directly from `LinkSubmit`

```jsx
  const onPaste = useCallback(
    (event: ClipboardEvent) => {
      if (!event.clipboardData || !event.clipboardData.files.length) {
        return;
      }
      event.preventDefault();
      event.stopPropagation();
      uploadDefault(event.clipboardData.files[0]).then(setLinkValue);
    }, [setLinkValue, uploadDefault]
  );
```

pkg/interface/src/views/components/ImageInput.tsx
jamesacklin: Thanks so much for this explanation, will refactor into regular functions.

pkg/interface/src/views/components/ImageInput.tsx
jamesacklin: This lets the user click on the link itself instead of focusing the field beneath it. A bit of trickery, can see if it will still work without.

jamesacklin:
@liam-fitzgerald Feel free to take another look -- I'm fairly sure I eliminated the closures-as-components issue, but feel free to guide me where I may have misinterpreted what you asked.

In other news, I was able to move the upload error inline with the rest of the control, eliminating the line of text beneath the field on error:
image

@tylershuster I will look into adding the drag-and-drop functionality in a bit.

pkg/interface/src/views/components/ImageInput.tsx
liam-fitzgerald: This is still an issue, because even though it looks like we aren't capturing any variables, you're still defining these components inside the scope of the component. You can either pull these out so that they're defined in the top level of the file, or just call them like `{errorRetry()}`

pkg/interface/src/views/components/ImageInput.tsx
liam-fitzgerald: oh, cool

liam-fitzgerald:
@jamesacklin closures-as-components still outstanding as an issue, lmk if you'd like to pair on it at some point or have questions. See my comment in review above

liam-fitzgerald (approved):
Stamped!

PR #4727, opened 2021-04-05
vere: enable HTTP2

liam-fitzgerald:
Enables HTTP2 for the C king. This greatly reduces head of line blocking due to request contention and improves landscape performance across the board, due to far lower TCP overhead. As HTTP scries are still stateful and cannot be run in parallel, the request still blocks, only on the arvo-side rather than on the client. This is also likely to reduce the frequency of reconnects on mobile. I've been running this on my ship over the weekend to no apparent ill effect. Worth noting that all modern browsers only support HTTP2 iff served over HTTPS. 

Questions:
- How will this impact proxying in the hosting environment? @brendanhay 
- Should this also be enabled for KH?


liam-fitzgerald:
update: SSE appears to not work in safari, investigating further

tylershuster:
Is that something we can fix in https://github.com/urbit/urbit/pull/4611 ?

liam-fitzgerald:
No, I suspect it's a header issue (HTTP2 doesn't support `transfer-encoding: chunked`)

tacryt-socryp:
Interesting, I would have expected a seamless upgrade for HTTP/2 SSE, since they're primarily meant to be used with HTTP/2.

liam-fitzgerald:
Updated to strip the `connection` header from outgoing HTTP2 responses, which fixes it for safari. Given that `connection: keep-alive` is the default in HTTP/1.1, I question its necessity, because otherwise you end up with the uncomfortable hack in the latest commit.

brendanhay:
No effect on hosting, everything is HTTP/2 capable by default.

Server support aside; I'll highlight that the [curl](https://github.com/urbit/urbit/blob/fbd85abf4e41d580654606a6defb764f6a97256d/nix/overlays/native.nix#L21-L27) we build [vere](https://github.com/urbit/urbit/blob/fbd85abf4e41d580654606a6defb764f6a97256d/nix/pkgs/urbit/default.nix#L25) against has HTTP/2 disabled due to static linking issues with the upstream dependencies - I assume this means`vere` is unable to originate HTTP/2 requests.

liam-fitzgerald:
You're right, no gesture has been made towards `%iris` HTTP2 support, but `%eyre` HTTP2 solves real problems in userspace right now, whereas I can't think of much that HTTP2 support in iris would solve.

locpyl-tidnyd:
Correct me if I'm wrong but libcurl is only used in C-urbit for two things: downloading pills (vere/king.c) and communicating with Ethereum (vere/dawn.c). The HTTP client in vere/io/cttp.c which I assume is used as the backend for `%iris` relies on `h2o_http1client_t`.

Fang-:
Worth noting that there's to-do's like this one in both the runtime and eyre:

https://github.com/urbit/urbit/blob/f18c607186eb874ca591dcc6784a979bfd638671/pkg/urbit/vere/io/http.c#L890

So as-is this might run into "not seeing the relevant cookie" failure cases.

tacryt-socryp:
Any word on ways to fix this / if it even affects anything on Urbit currently? @liam-fitzgerald Definitely think this should go out ASAP (did I hear word of a Vere 1.4 release soon @joemfb ?)

pkg/urbit/vere/io/http.c
```diff
@@ -677,6 +677,8 @@ _http_start_respond(u3_hreq* req_u,
     if ( 0 == strncmp(hed_u->nam_c, "content-length", 14) ) {
       has_len_i = 1;
     }
+    // ignore connection specific headers for HTTP2
+    else if ( 0 == strncmp(hed_u->nam_c, "connection", 10) ) {}
```
joemfb, line 681: I think this header should only be ignored if the request used http2. You can check that with `0x200 <= rec_u-version`.

joemfb (changes requested):
Sorry, I missed your update and thought that there was still an issue with Safari.

joemfb:
@locpyl-tidnyd, that's correct. We've talked about switching the client implementation to curl (and @tacryt-socryp did most of the work on an old branch), but there's hasn't been a pressing need. I believe libh2o has changed significantly, so it may be worth revisiting when we go to upgrade.

joemfb (approved):
LGTM

FWIW. I'll probably run this through ASAN before we cut the next release, just to ease my mind as to why it was disabled.

Also, you may want to pull the `file-server.hoon` changes out of this PR so that they can get onto the network in a timely fashion.

And we should definitely look into HTTP2 in KH.

joemfb:
@liam-fitzgerald, i haven't had a chance to test these changes yet. But @tacryt-socryp did, and there were still issues in safari. So my suggestion wrt the version flag might be wrong in some way.

liam-fitzgerald:
Screen Shot 2021-04-13 at 4 58 17 pm
I'm successfully using HTTP/2 with safari, @loganallenc can you provide a more specific repro?

tacryt-socryp:
I have no clue. It just wouldn't work when I was testing Friday no matter what I did

tacryt-socryp:
@joemfb @liam-fitzgerald There is a strange and persistent bug that is occurring currently on my ship running the HTTP/2 branch. Easy to recreate.

```
curl https://tacryt-socryp.arvo.network/~landscape/js/bundle/index.fbd802ee7176603abe17.js
curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
```

joemfb:
Can you try again with `-v` or `-trace`? Per h2o/h2o#778 this may be an issue with the content-length header.

tacryt-socryp:
```
*   Trying 35.193.202.187...
* TCP_NODELAY set
* Connected to tacryt-socryp.arvo.network (35.193.202.187) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/cert.pem
  CApath: none
* TLSv1.2 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS handshake, Server key exchange (12):
* TLSv1.2 (IN), TLS handshake, Server finished (14):
* TLSv1.2 (OUT), TLS handshake, Client key exchange (16):
* TLSv1.2 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS handshake, Finished (20):
* TLSv1.2 (IN), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (IN), TLS handshake, Finished (20):
* SSL connection using TLSv1.2 / ECDHE-RSA-AES256-GCM-SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=tacryt-socryp.arvo.network
*  start date: Apr  7 17:32:37 2021 GMT
*  expire date: Jul  6 17:32:37 2021 GMT
*  subjectAltName: host "tacryt-socryp.arvo.network" matched cert's "tacryt-socryp.arvo.network"
*  issuer: C=US; O=Let's Encrypt; CN=R3
*  SSL certificate verify ok.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x7fe47a00c600)
> GET /~landscape/js/bundle/index.fbd802ee7176603abe17.js HTTP/2
> Host: tacryt-socryp.arvo.network
> User-Agent: curl/7.64.1
> Accept: */*
> 
* Connection state changed (MAX_CONCURRENT_STREAMS == 100)!
< HTTP/2 307 
< server: urbit/vere-1.3
< date: Fri, 16 Apr 2021 18:04:45 GMT
< location: /~/login?redirect=/~landscape/js/bundle/index.fbd802ee7176603abe17.js
< 
* Connection #0 to host tacryt-socryp.arvo.network left intact
* Closing connection 0
```

Note that "Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0"

pkg/urbit/vere/io/http.c
```diff
@@ -677,6 +677,8 @@ _http_start_respond(u3_hreq* req_u,
     if ( 0 == strncmp(hed_u->nam_c, "content-length", 14) ) {
       has_len_i = 1;
     }
+    // ignore connection specific headers for HTTP2
+    else if ( 0x200 <= rec_u->version && 0 == strncmp(hed_u->nam_c, "connection", 10) ) {}
```
joemfb, line 681: We should downcase all headers for http2 responses. I'd switch to something like this:

```
// ignore connection specific headers for HTTP2
//
if ( 0x200 <= rec_u->version ) {
  h2o_strtolower(hed_u->nam_c, hed_u->nam_w);
  
  if ( 0 == strncmp(hed_u->nam_c, "connection", 10) ) {
    continue;
  }
}

if ( 0 == strncmp(hed_u->nam_c, "content-length", 14) ) {
  has_len_i = 1;
}
```

joemfb (changes requested):

pkg/urbit/vere/io/http.c
joemfb: We might be able to get away with this for HTTP1 as well, but I've seen comments to the effect that title-case is maximally compatible, at least for key headers ...

tacryt-socryp:
It could be the case that libh2o has the same issue that Apache 9.8 had, though I'm looking closer to verify: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=915103

It does seem to present with the same error, and in similar conditions.

tacryt-socryp:
I'm curious if HTTP/2 will "just work" without any further changes in this branch if we get https://github.com/urbit/urbit/pull/5003 merged in.

tacryt-socryp:
Can confirm that when using my v2.2.6 libh2o branch alongside this branch that in Safari we connect to Urbit via http2. However, this breaks the event source used in server-side events. It seems we need to wait until we finish the frontend Hook-based Fetch EventSource API switch before pursuing this further.

tacryt-socryp:
Ping.

joemfb:
I've been assured that this is ready to go. I've also smoke-tested locally: on a fake ship, booted from the livenet pill, using a self-signed cert, chat works over http2 in chrome and safari on my mac.

joemfb (approved):
LGTM

PR #4729, opened 2021-04-06
subscription: clear opened subs before restart

liam-fitzgerald:
Fixes subscriptions not restarting properly

### DIFF
--- a/pkg/interface/src/logic/subscription/global.ts
+++ b/pkg/interface/src/logic/subscription/global.ts
@@ -49,6 +49,7 @@ export default class GlobalSubscription extends BaseSubscription<StoreState> {
   }
 
   restart() {
-    this.start();
+    this.openSubscriptions = {};
+    super.restart();
   }
 }


tacryt-socryp (approved):

tacryt-socryp:
Could this cause us to subscribe twice to the same endpoint? I'm not sure, since I think it would depend on whether Eyre ensures a lack of duplicate subscriptions within a channel object, since this shouldn't switch the channel we're using.

liam-fitzgerald:
When we restart, it's against a totally new channel object. channel.js makes an attempt to cleanup the connection, but it's not always successfully. Even if it's not cleaned up properly, it should get caught by the clog logic. It's more than likely that the old subscriptions have clogged by the time we restart our subscriptions as well.

tacryt-socryp:
Surely we should try to use the same channel in case we didn't clog? Because I'd imagine the majority of subscriptions wouldn't clog (how long does it take to get 50 group-store updates or metadata-store updates, right?)

liam-fitzgerald:
We can't use the same channel until @tylershuster s fetch based event source work lands, see urbit/urbit#4004

PR #4730, opened 2021-04-06
Invite: fix stale props in inviteAccept callback

liam-fitzgerald:
Now that the store is immutable, callbacks that close over props will
see stale props because the reference is captured at instantiation. This
was not an issue previously, because the references would be directly
mutated. To remedy this, useRunIO has been added, which takes an async
function to run for the actual network processing, and a second
callback, which is able to see fresh props because it is not
instantiated until the async callback has resolved successfully.
Importantly, useRunIO does not resolve until the second callback has
finished, preserving the semantics of the callback for Formik handlers
and the like.

Fixes urbit/landscape#691

### DIFF
--- a/pkg/interface/src/views/components/Invite/index.tsx
+++ b/pkg/interface/src/views/components/Invite/index.tsx
@@ -1,27 +1,28 @@
-import React, { useState, useEffect, useCallback } from 'react';
-import { useHistory } from 'react-router-dom';
+import React, { useState, useEffect, useCallback } from "react";
+import { useHistory } from "react-router-dom";
 
 import {
   MetadataUpdatePreview,
   Contacts,
   JoinRequests,
   Groups,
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/components/Invite/index.tsx
```diff
@@ -32,62 +33,81 @@ interface InviteItemProps {
...
-    } else {
-      history.push(`/~landscape${resource}`);
-    }
-  }, [app, history, waiter, invite, uid, resource, groups, associations]);
+    },
+    resource
+  );
+}
+
+export function InviteItem(props: InviteItemProps) {
```
tacryt-socryp, line 99: Please break out this component to its own file

tacryt-socryp (approved):
useRunIO is really well written, good job

PR #4732, opened 2021-04-06
sane: don't kick unsubscribed

liam-fitzgerald:
Fixes urbit/landscape#735


tacryt-socryp (approved):

tacryt-socryp:
Why did we do this to begin with?

liam-fitzgerald:
Post-breach cleanup essentially. People who don't have subscriptions to the group have likely breached, which could affect group joining flows, so we remove them. I don't think it's a great idea though, because it can be *very* destructive

matildepark:
This probably should've just targetted master, it should go out with a few other hotfixes today.

PR #4734, opened 2021-04-06
interface: minor tsc fixes

mrdomino:
Prior to these fixes, running `npm run tsc` in `pkg/interface` would hard-fail with only those two errors. Fixing this makes the `npm run tsc` output more usable, in case we ever desire a CI check to e.g. make sure the count of type errors does not increase.


liam-fitzgerald (approved):
Righteous

PR #4735, opened 2021-04-06
gcp: isConfigured doesn't touch store

mrdomino:
Makes the thread return a simple boolean; almost as simple as just
returning `%.y` or `%.n`.

`gcpManager` now stores whether GCP is configured as a private variable.

Also fixes some type errors in gcp.


liam-fitzgerald:
What's the motivation for this not touching store? You can generally end up with nasty stale render bugs, it's not recommended

mrdomino:
Basically just not wanting the thread to return an object. Also thinking that Landscape doesn't need to know whether GCP store is "configured" or not — i.e. that it's only interested in the state "has a token" and "doesn't have a token." So the thing to look at is gcp.token anyway, which is what useStorage does.

If Landscape in fact wants to show the state where GCP has been configured but doesn't have a token yet, this PR probably doesn't make sense.

liam-fitzgerald (approved):
Sounds good. Wondering if we should maybe have a bool mark so you don't have to manually return JSON here

mrdomino:
Maybe. I have the feeling of an overall “something here is out of place / can be simplified, and I don’t know what it is.” A bool mark feels like maybe a step in the wrong direction — or maybe it’d be perfect. It more just doesn’t seem like `-gcp-is-configured` deserves to be a thread, but it’s not clear what else it would be (aside from a gall app, which is probably what `gcp-store` really ought to be anyway.)

PR #4736, opened 2021-04-06
observe-hook: upgrade to keep marks warm

tacryt-socryp:
Keep all marks warm so as to never need to build a mark on demand. Can confirm that on a livenet comet, this stops all marks from rebuilding, whereas some marks (ex: `%hark-update`) would rebuild continually before this.

### DIFF
--- a/pkg/arvo/app/observe-hook.hoon
+++ b/pkg/arvo/app/observe-hook.hoon
@@ -162,8 +162,7 @@
   ++  warm-cache-all
     ?:  warm-cache
       ~|('cannot warm up cache that is already warm' !!)
-    :::_  this(warm-cache %.y)
-    :_  this
+    :_  this(warm-cache %.y)
     =/  =mood:clay  [%t da+now.bowl /mar]
     [%pass /warm-cache %arvo %c %warp our.bowl %home `[%sing mood]]~
   ::
@@ -325,7 +324,6 @@
     =*  mark  t.wire
     ?~  riot
       ~|('mark build failed for {<mark>}' !!)
-    ~&  mark
     :_  this
     =/  =rave:clay  [%next %b [%da now.bowl] mark]
     [%pass wire %arvo %c %warp our.bowl [%home `rave]]~
--- a/pkg/arvo/app/observe-hook.hoon
+++ b/pkg/arvo/app/observe-hook.hoon
@@ -163,8 +163,8 @@
     ?:  warm-cache
       ~|('cannot warm up cache that is already warm' !!)
     :_  this(warm-cache %.y)
-    =/  =mood:clay  [%t da+now.bowl /mar]
-    [%pass /warm-cache %arvo %c %warp our.bowl %home `[%sing mood]]~
+    =/  =rave:clay  [%sing [%t da+now.bowl /mar]]
+    [%pass /warm-cache %arvo %c %warp our.bowl %home `rave]~
   ::
   ++  cool-cache-all
     ?.  warm-cache
@@ -291,17 +291,18 @@
 ++  on-arvo
   |=  [=wire =sign-arvo]
   ^-  (quip card _this)
+  :_  this
   ?+    wire  (on-arvo:def wire sign-arvo)
       [%warm-cache ~]
     ?.  warm-cache
-      [~ this]
+      ~
     ?>  ?=([%clay %writ *] sign-arvo)
-    :_  this
-    :-  =/  =rave:clay  [%next [%t da+now.bowl /mar]]
-        [%pass /warm-cache %arvo %c %warp our.bowl %home `rave]
     =*  riot  p.sign-arvo
     ?~  riot
-      ~
+      =/  =rave:clay  [%next [%t da+now.bowl /mar]]
+      [%pass /warm-cache %arvo %c %warp our.bowl %home `rave]~
+    :-  =/  =rave:clay  [%next [%t q.p.u.riot /mar]]
+        [%pass /warm-cache %arvo %c %warp our.bowl %home `rave]
     %+  turn  !<((list path) q.r.u.riot)
     |=  pax=path
     ^-  card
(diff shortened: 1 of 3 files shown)


joemfb:
Does this need to account for changes (additions, deletions) to the set of marks on a ship? And does the current %clay subscription system support such a request?

tacryt-socryp:
@joemfb Upon a `%next` request receiving `~`, we crash and print that the mark build failed for that particular mark. This will prevent attempts to rebuild that mark. So if we deleted a mark then re-added it later, you'd need to run the `[%warm-cache-all ~]` action again, but other than that, it should handle deletions fine.

tacryt-socryp:
@joemfb I have now tested that this handles marks being added properly.

philipcmonk:
You can tell whether a directory's contents have changed with a `%z` request, then scan for any new marks.

philipcmonk:
Oh yeah I see you used `%t`, that should work too.  It produces a deep directory listing, right?

tacryt-socryp:
@philipcmonk Yes, it produces a deep directory listing, which is ideal for our use case.

pkg/arvo/app/observe-hook.hoon
```diff
@@ -260,9 +288,53 @@
...
+++  on-arvo
+  |=  [=wire =sign-arvo]
+  ^-  (quip card _this)
+  ?+    wire  (on-arvo:def wire sign-arvo)
+      [%warm-cache ~]
+    ?.  warm-cache
+      [~ this]
+    ?>  ?=([%clay %writ *] sign-arvo)
+    :_  this
+    :-  =/  =rave:clay  [%next [%t da+now.bowl /mar]]
```
philipcmonk, line 300: It would be slightly more conceptually coherent to get the case from `q.p.u.riot.sign-arvo`, which will guarantee you get an update for every relevant revision with no gaps.  I'm not sure there's a case in the system as-is where this could break, but it relies on certain properties that we haven't always had and might not in the future.  A problematic case would be something along the lines of: another subscriber to clay hears this update first and makes another commit before this resubscription happens, so you don't hear about the second commit.  This wouldn't be a problem now because you're warming up the most recent commit no matter what (which is good), but it could be different if e.g. we adjust the meaning of a clay commit at `now.bowl`.

pkg/arvo/app/observe-hook.hoon
```diff
@@ -260,9 +288,53 @@
...
+    [%pass [%mar mark ~] %arvo %c %warp our.bowl %home `rave]
+  ::
+      [%mar ^]
+    ?.  warm-cache
+      [~ this]
+    ?>  ?=([%clay %writ *] sign-arvo)
+    =*  riot  p.sign-arvo
+    =*  mark  t.wire
+    ?~  riot
+      ~|('mark build failed for {<mark>}' !!)
```
philipcmonk, line 330: I'm a little surprised this doesn't crash when you delete a mark.  In that case I would expect a `%next` subscription to give a null result.

tacryt-socryp: It does crash upon a mark being deleted, but afterward doesn't rebuild the mark or continue to listen to that mark. Would you prefer we just return ~?

liam-fitzgerald (approved):
A nit, else LGTM

pkg/arvo/app/observe-hook.hoon
tacryt-socryp: Do you think I should stop using `da+now.bowl` in all of the scries, or just this one? I don't understand the semantics of the `case` and how it should be different well enough to be confident that this change will preserve the appropriate semantics.

pkg/arvo/app/observe-hook.hoon
philipcmonk: Yeah probably.  We shouldn't print stack traces when it's probably not an error.

pkg/arvo/app/observe-hook.hoon
philipcmonk: Just this one.  The others are properly thought of as "most recent", while this one is "next time this changes after the notification i got"

tacryt-socryp: What about the one on line 327 where we also send a `%next` request?

philipcmonk (approved):
Looks good!

pkg/arvo/app/observe-hook.hoon
```diff
@@ -260,9 +288,48 @@
...
+    =*  riot  p.sign-arvo
+    ?~  riot
+      =/  =rave:clay  [%next [%t da+now.bowl /mar]]
+      [%pass /warm-cache %arvo %c %warp our.bowl %home `rave]~
+    :-  =/  =rave:clay  [%next [%t q.p.u.riot /mar]]
+        [%pass /warm-cache %arvo %c %warp our.bowl %home `rave]
+    %+  turn  !<((list path) q.r.u.riot)
+    |=  pax=path
+    ^-  card
+    =.  pax  `path`(snip (slag 1 `(list @)`pax))
```
joemfb, line 309: These casts look unnecessary: `=/(a=path /foo/bar/baz/bud =.(a (snip (slag 1 a)) a))`.

pkg/arvo/app/observe-hook.hoon
```diff
@@ -260,9 +288,48 @@
...
+    =/  mark=@ta
+      %+  roll  pax
+      |=  [=term mark=term]
+      ?:  =(0 (lent (trip mark)))
```
joemfb, line 313: This could just be `?=(%$ mark)`.

joemfb (approved):
This makes sense and looks good.

One overall note:

It looks like this emits duplicate subscriptions anytime the `%t` read comes back, and therefore depends on %clay to deduplicate them (which should work, since %gall will alway use the same duct). That's much simpler than tracking the current subscriptions and deduplicating them eagerly, I just want to call out the behavior (and assumptions) explicitly.

I couple style nits:

pkg/arvo/app/observe-hook.hoon
```diff
@@ -260,9 +288,49 @@
...
+        term
+      %-  crip
+      %-  zing
+      :~  (trip mark)
+          "-"
+          (trip term)
```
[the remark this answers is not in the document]
ohAitch: `(rap 3 mark '-' term ~)`
(or `(rap 3 (join '-' ~[mark term]))` but that's arguably less readable)

pkg/arvo/app/observe-hook.hoon
ohAitch: Also you probably wanted double quotes, though if the message is going away the point is moot.

pkg/arvo/app/observe-hook.hoon
tacryt-socryp: Thanks, I've included both of your recommendations in the latest commit.

PR #4743, opened 2021-04-08
groups: forward to already-joined group from GroupLink

jamesacklin:
Forwards users directly to groups they have already joined when they click a GroupLink.

Video attached; I belong to the "Clack" group and not the "Areological Fermentation" group:

https://user-images.githubusercontent.com/748181/114098922-087c2b80-9890-11eb-8a5a-a0de17f8cc08.mp4

Fixes urbit/landscape#593


urcades (approved):
This is perfect — thanks much for including a video with PRs like this, we should start making this into A Thing.

liam-fitzgerald (approved):
Very nice!

PR #4744, opened 2021-04-08
king: fix lanes yet again, also reuse udp port

### LINKED ISSUE #4707: frequent "ames: send fail: invalid argument" on (our) host ships
Filing for posterity. Last night during deploy we started the deploy on ~bitbet-bolbel, Urbit Community's host ship, before other ships, to get into an ideal networking case for that group. We noticed inconsistent `ames: send fail: invalid argument` printfs, so @belisarius222 jumped in call and @philipcmonk + @joemfb provided help as well.

Cause is still indeterminate, but suspicion lies on potential different endianness between KH+Vere's IPs / lanes (see also #4364) creating invalid IPs for either. As per @philipcmonk, "we need to print out what's actually getting produced from both kh and c vere and just see if they're the same."

pilfer-pandex:
In king haskell, it seems, the "lane" is more of a poorly lit alley, an alley I hope I don't have to walk through ever again.

This time around, it seems I got the byte order of the ip address wrong. According to https://hackage.haskell.org/package/network-3.1.2.1/docs/Network-Socket.html#t:HostAddress, the ip address in haskell is represented as a Word32, of which it is said, confusingly, that "[t]he raw network byte order number is read using host byte order." On real computers, this means it's reversed, which, apparently, is not what the corresponding c function does. So we fix that here. Hopefully.

We also add code to guarantee that when your UDP port is closed and reopened due to a problem, it is opened at the same port. Previously, if you hadn't explicitly specified a port, we'd ask the OS for the next available one, which would be different each time. Because not explicitly specifying a port is represented as "port 0" in the code, however, a naive read-through of the code suggests that this was always the behavior, which it was not.

Fixes #4707

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
@@ -25,7 +25,6 @@ module Urbit.Arvo.Common
 import Urbit.Prelude
 
 import Control.Monad.Fail (fail)
-import Data.Bits
 import Data.Serialize
 
 import qualified Network.HTTP.Types.Method as H
@@ -185,9 +184,9 @@ newtype Ipv4 = Ipv4 { unIpv4 :: N.HostAddress }
   deriving newtype (Eq, Ord, Enum)
 
 instance Serialize Ipv4 where
-  get = Ipv4 <$> N.tupleToHostAddress
-    <$> ((,,,) <$> getWord8 <*> getWord8 <*> getWord8 <*> getWord8)
-  put (Ipv4 (N.hostAddressToTuple -> (a, b, c, d))) = for_ [a, b, c, d] putWord8
+  get = (\a b c d -> Ipv4 $ N.tupleToHostAddress $ (d, c, b, a))
+    <$> getWord8 <*> getWord8 <*> getWord8 <*> getWord8
+  put (Ipv4 (N.hostAddressToTuple -> (a, b, c, d))) = for_ [d, c, b, a] putWord8
 
 instance ToNoun Ipv4 where
   toNoun = serializeToNoun
@@ -196,11 +195,11 @@ instance FromNoun Ipv4 where
   parseNoun = serializeParseNoun "Ipv4" 4
 
 instance Show Ipv4 where
-  show (Ipv4 i) =
-    show ((shiftR i 24) .&. 0xff) ++ "." ++
-    show ((shiftR i 16) .&. 0xff) ++ "." ++
-    show ((shiftR i  8) .&. 0xff) ++ "." ++
-    show (i .&. 0xff)
+  show (Ipv4 (N.hostAddressToTuple -> (a, b, c, d))) =
+    show a ++ "." ++
+    show b ++ "." ++
+    show c ++ "." ++
+    show d
 
 -- @is
 -- should probably use hostAddress6ToTuple here, but no one uses it right now
--- a/pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
@@ -29,6 +29,7 @@ import Data.Bits
 import Data.Serialize
 
 import qualified Network.HTTP.Types.Method as H
+import qualified Network.Socket            as N
 import qualified Urbit.Ob                  as Ob
 
 
@@ -159,6 +160,19 @@ deriveNoun ''JsonNode
 
 -- Ames Destinations -------------------------------------------------
 
+serializeToNoun :: Serialize a => a -> Noun
+serializeToNoun = A . bytesAtom . encode
+
+serializeParseNoun :: Serialize a => String -> Int -> Noun -> Parser a
+serializeParseNoun desc len = named (pack desc) . \case
+    A (atomBytes -> bs)
+      -- Atoms lose leading 0s, but since lsb, these become trailing NULs
+      | length bs <= len -> case decode $ bs <> replicate (len - length bs) 0 of
+        Right aa -> pure aa
+        Left msg -> fail msg
+      | otherwise -> fail ("putative " <> desc <> " " <> show bs <> " too long")
+    C{} -> fail ("unexpected cell in " <> desc)
+
 newtype Patp a = Patp { unPatp :: a }
   deriving newtype (Eq, Ord, Enum, Real, Integral, Num, ToNoun, FromNoun)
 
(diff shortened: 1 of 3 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Ames/UDP.hs
```diff
@@ -5,7 +5,8 @@
 
     - If can't open the port, wait and try again repeatedly.
     - If there is an error reading or writting from the open socket,
-      close it and open another.
+      close it and open another, making sure, however, to reuse the
+      same port
```
joemfb, line 9: What kinds of errors make it necessary to rebind the socket?

pkg/hs/urbit-king/lib/Urbit/Vere/Ames/UDP.hs
```diff
@@ -5,7 +5,8 @@
 
     - If can't open the port, wait and try again repeatedly.
     - If there is an error reading or writting from the open socket,
```
philipcmonk, line 7: s/writting/writing

philipcmonk (commented):
I don't see the byte order reversal, is that commit here?

pilfer-pandex:
@philipcmonk, sorry about that; pushed

pilfer-pandex:
please disregard the first commit of this pr for review purposes. It is properly part of another (in-progress) pr, but is needed for testing this functionality on ~def

pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
```diff
@@ -167,8 +181,19 @@ newtype Port = Port { unPort :: Word16 }
...
 -- @if
-newtype Ipv4 = Ipv4 { unIpv4 :: Word32 }
-  deriving newtype (Eq, Ord, Enum, Real, Integral, Num, ToNoun, FromNoun)
+newtype Ipv4 = Ipv4 { unIpv4 :: N.HostAddress }
+  deriving newtype (Eq, Ord, Enum)
+
+instance Serialize Ipv4 where
+  get = Ipv4 <$> N.tupleToHostAddress
+    <$> ((,,,) <$> getWord8 <*> getWord8 <*> getWord8 <*> getWord8)
+  put (Ipv4 (N.hostAddressToTuple -> (a, b, c, d))) = for_ [a, b, c, d] putWord8
```
philipcmonk, line 190: Let me see if I understand this.  `<$>` is infix fmap with the function on the left and the value on the right, so the flow of control sort of goes right-to-left.  So the second line says "get 4 word8s and put them in a tuple".  I assume this happens "in order", so that if the original string was `0x7f000001`, then the tuple will be `[0x7f, 0x0, 0x0, 0x1]`.  Is that correct?

Then the first line converts this tuple into a `N.HostAddress` and then makes it of type `Ipv4`.

And then for `put`, we reverse `N.hostAddressToTuple` to get the word8s, which we put back in.  This seems to add them in order from "left to right", so I guess putWord8 adds to the least significant side.

So this looks to me like it leaves the everything in big-endian (i.e. network) order instead of reversing at all.  Is that correct?

philipcmonk (commented):
How did you test this does the right thing now?

pkg/hs/urbit-king/lib/Urbit/Arvo/Common.hs
joemfb: The key to note is that `hostAddressToTuple` calls `ntohl()` in its implementation, and `tupleToHostAddress` calls `htonl()` in its. I only know this 'cause I read the source ... I can't say I'm much of a fan of `Network.Socket` after this.

pilfer-pandex: Joe is right in pointing out that the "tuple conversion" functions in Network.Socket do "reversals" on little-endian machines. Philip's read of the applicative notation is right, in that it preserves order.

The put and get functions operate in monads which statefully produce / consume a ByteString in the order it would go on the network. So putWord8 0x01 >> putWord8 0x02 is the same as putWord16be 0x0102, where "be" means "big endian." Now <*> and <$> are applicative notation; <$> is fmap, and <*> basically lets you curry additional monadic arguments to your function. You'll note that if you just have fmap, and you have a binary function a -> b -> c, and an f a and an f b, you don't have any way of getting an f c; you need an extra capability for that.

The main problem with Network.Socket was that it made HostAddress a type alias, rather than an opaque newtype. We should never have had direct access to the underlying Word32, and that's how we got into this mess.

That, or the Word32 should have been in network order on all machines (unless that would mean calling htonl at each syscall, in which case the first seems to be the superor option).

joemfb: > So this looks to me like it leaves the everything in big-endian (i.e. network) order instead of reversing at all. Is that correct?

@philipcmonk you are right, and this code is wrong. An example:

The ip address for localhost is `.127.0.0.1`. In network byte order, that is 0x1000007f; in (little-endian) host byte order, that is 0x7f000001. As returned by `hostAddressToTuple`, it is `(127, 0, 0, 1)`. `Network.Socket.HostAddress` is in network byte order, the tuple calls convert to/from host byte order.

This line specifically is wrong:

```
put (Ipv4 (N.hostAddressToTuple -> (a, b, c, d))) = for_ [a, b, c, d] putWord8
```

In our example, that becomes `for_ [127, 0, 0, 1] putWord8`, which evaluates to `0x1000007f`. The correct result would be 0x7f000001.

---

> ... the Word32 should have been in network order on all machines ...

@pilfer-pandex it is.

pkg/hs/urbit-king/lib/Urbit/Vere/Ames/UDP.hs
joemfb: @pilfer-pandex bump.

pilfer-pandex:
Reversal now in place. I tested it on def and dev, as follows, scrying for my lane using

```
=/(ss .^(ship-state:ames ax+/=//=/peers/~pilfer-pandex) ?>(?=(%known -.ss) route.ss))
```

in dojo, after having |hi'd from ~pilfer-pandex:

- On ~def, running old king binary, result was `[~ [direct=%.y lane=[%.n p=0xaabb.ccdd.eeff]]]`
- On ~dev, running vere, result was `[~ [direct=%.y lane=[%.n p=0xaabb.ffee.ddcc]]]`
- On ~def, running king binary built with this pr, result was `[~ [direct=%.y lane=[%.n p=0xaabb.ffee.ddcc]]]`

pkg/hs/urbit-king/lib/Urbit/Vere/Ames/UDP.hs
pilfer-pandex: Any failure to send along the socket, whatsoever.

joemfb: I see that's what it does, but i'm trying to understand why. I'm not aware of any issues with UDP itself that would require you to rebind the socket (unless we were binding a specific ip address and that address changed, but we always use localhost or "any address").

joemfb: @pilfer-pandex rebump.

pilfer-pandex: @joemfb I understand what you're saying here: The udp socket, we think, is a "stateless" object, such that destroying it and creating it anew is exactly the same thing as using the existing one. That makes sense as a hypothesis, but on the other hand, I can't believe that benjamin didn't know about it. The code here was very deliberately written to do this cumbersome thing, and the comment at the top of the file calls out how important it is to do this (but not why). I don't feel comfortable removing it until I understand what problem it solves. If you could conclusively argue that it definitely has no effect, then I'd be less concerned. Right now, any "IOError" that can occur on send will trigger this error handling. I think (?) that all these IOErrors correspond to potential errnos. You think (?) that all of those possible errnos are independent of any "state" of the socket. These (?)s need to be spelled out in full.

tacryt-socryp:
is this ready?

pilfer-pandex:
Yes, it's ready. @joemfb @philipcmonk

philipcmonk (approved):
This looks correct to me, so if it's good in tests then let's release it.

joemfb (changes requested):
Please add a comment noting that the UDP socket rebinding on error appears to be unnecessary.

And I think the LMDB revert should be dropped from this PR.

The ip address handling LGTM.

joemfb:
@pilfer-pandex bump

pkg/hs/urbit-king/lib/Urbit/Vere/Ames/UDP.hs
joemfb: @pilfer-pandex this typo is still present.

joemfb (approved):
LGTM

(There's till a typo in an old comment, noted above)

PR #4745, opened 2021-04-09
clay: add /~ ford rune, as /= for directories

Fang-:
Hoon files may want to import nouns from all files in a given directory.
`/~` lets you do so, importing as a noun structured like a `(map @ta *)`.

Note the description as "directories" here, instead of "path prefix".
The behavior, as implemented, will not include `/path/hoon` for `/~  /path`, instead only including `/path/more/hoon` and more deeply nested files.
This seems to be, generally, the behavior you want, for example when importing from `/app/myapp/*` for use in `/app/myapp/hoon`.
It's a bit out of line compared to general clay semantics though, so RFC.

Also, the map keying by `@ta` seems more ergonomic than `path` for most cases, but does run into the usual hyphenation-ambiguity problem, where `/path/a/b/hoon` and `/path/a-b/hoon` would conflict with each other. Again, RFC.

From my testing, using the resulting map requires manual casting, which is not ideal. Not sure how to solve for this besides adding `/^` back in as well, but perhaps there's some wizardry to be done here. RFC the third!

Internally, in addition to implementing the logic for `/~` by itself, we update the `ford-cache` to be able to tell whether a dependency is a single file, or the path of a directory whose contents we depend on. Without this, we would not rebuild for file additions.

This code isn't gonna win any beauty contests yet (`+build-directory` and `+invalidate` in particular need some love), but I wanted to go ahead and start getting eyes on this. Maybe the ideal refactoring will come to me in a dream...

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -932,7 +932,7 @@
         ;~(plug sym ;~(pfix gap stap))
       ::
         %+  rune  sig
-        ;~(plug sym ;~(pfix gap stap))
+        ;~((glue gap) sym wyde:vast stap)
       ::
         %+  rune  cen
         ;~(plug sym ;~(pfix gap ;~(pfix cen sym)))
@@ -995,7 +995,7 @@
       $(sut (slop pin sut), raw t.raw)
     ::
     ++  run-raz
-      |=  [sut=vase raz=(list [face=term =path])]
+      |=  [sut=vase raz=(list [face=term =spec =path])]
       ^-  [vase state]
       ?~  raz  [sut nub]
       =^  res=(map path vase)  nub
@@ -1003,20 +1003,23 @@
       =;  pin=vase
         =.  p.pin  [%face face.i.raz p.pin]
         $(sut (slop pin sut), raz t.raz)
-      ::  convert the (map path vase) into a vase of (map @ta *),
-      ::  with paths hyphenated into names and the .hoon cut off
       ::
       =/  lap=@ud  (lent path.i.raz)
-      |-  ^-  vase
-      ?~  res  [[%atom %n `0] 0]
-      %+  slop
-        %+  slop
-          :-  [%atom %ta ~]
-          %+  rap  3
-          %+  join  '-'
-          (snip (slag lap p.n.res))
-        q.n.res
-      (slop $(res l.res) $(res r.res))
+      =/  =type  (~(play ut p.sut) [%kttr spec.i.raz])
+      ::  ensure results nest in the specified type,
+      ::  and produce a homogenous map containing that type.
+      ::
+      :-  %-  ~(play ut p.sut)
+          [%kttr %make [%wing ~[%map]] ~[[%base %atom %ta] spec.i.raz]]
+      |-
+      ?~  res  ~
+      ~|  [%nest-fail path.i.raz p.n.res]
+      ?>  (~(nest ut type) | p.q.n.res)
+      :_  [$(res l.res) $(res r.res)]
+      :_  q.q.n.res
+      %+  rap  3
+      %+  join  '-'
+      (snip (slag lap p.n.res))
     ::
     ++  run-maz
       |=  [sut=vase maz=(list [face=term =mark])]
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -529,29 +529,6 @@
       ?<  (~(has in deletes) path)
       ~|  %file-not-found^path
       :_(nub (need (~(get an ankh) path)))
-    ::  +list-directory: retrieve directory listing
-    ::
-    ::    this excludes files directly at /path/mark,
-    ::    instead only including files in the unix-style directory at /path,
-    ::    such as /path/more/mark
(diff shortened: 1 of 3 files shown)


belisarius222:
I think treating a path as a directory and ignoring files directly under that path is fine here.

Note that dependency tracking with path -> `@ta` conversion is tricky.  You have to handle at least the following contingencies:
- /foo/bar was already there, then someone adds /foo-bar
- /foo-bar was already there, then someone adds /foo/bar
- /foo/bar and /foo-bar were already there, then someone deletes /foo/bar
- /foo/bar and /foo-bar were already there, then someone deletes /foo-bar
- /foo/bar gets renamed to /foo-bar
- /foo-bar gets renamed to /foo/bar

Another complication with path -> `@ta` conversion is that files in /lib, /sur, and /mar already have special behavior in this regard, so anything you do here will need to be compatible with that.  One option would be to turn off `/~` for any subdirectories of those folders, to prevent colliding semantics, although that would arguably be layering even more special cases on top of the special cases we already have.

> using the resulting map requires manual casting

This feels wrong to me.  What's the use case here?  We could handle this in a few different ways, but I don't have a sense of how this result is expected to be used.

The main use case I can think of this is to build tests.  For that, I'd be inclined to require a static type for all imported files.  We're used to thinking of test files as cores, which is cute, but they'd be easier to work with if every test file built to a `(map path (trap tang))`.

A test file would end up looking like:
```hoon
=>  |%
    ++  helper-foo  !!
    ++  helper-bar  !!
    --
%-  ~(gas by *(map path (trap tang)))
:~  :-  /ordered-map/put/simple
    |.  ^-  tang
    !!
::
    :-  /ordered-map/put/tough
    |.  ^-  tang
    !!
==
```

Then again, I'd rather not build tests using this rune, since we would lose the ability to build only a subset of test files, which is important for cycle time in development.  So I'm a bit at a loss for what we'd want to do with this new rune.

Fang-:
> Note that dependency tracking with path -> `@ta` conversion is tricky.

We don't use `@ta` in dependency tracking at all though! The only place that conversion happens is in the `map` that results from building the `/~`, that goes into the build target's subject. Dependency tracking here is just `[dir=%.y path=/foo]`.

> What's the use case here?

The immediate use case I have, for both professional and private projects, is the importing of "programmable content". For example, [in the chatbot](https://github.com/urbit/urbit/blob/71878ce109a8ffd199bb5157310836de188d886f/pkg/arvo/app/eliza.hoon#L633-L644), or [a choose-your-own-adventure engine](https://github.com/Fang-/suite/blob/4f9e2dfe69f1c76a54e0247281332f12c8240ba4/app/bard.hoon#L562-L574) there's "scripts" that the application can present the user with. Defining those in app source works, but it really shouldn't be that coupled. If it loaded that from a directory instead, you could trivially merge in new content without affecting app source at all.

> I'd be inclined to require a static type for all imported files

As in, requiring a consistent type across the imported files? This seems sane to me, I'm having trouble imagining a use case for "import all of these things, but we don't know what they might be".

But also, thinking on this again, maybe the only thing I was really missing were faces on the map? [Old ford](https://github.com/urbit/urbit/blob/4dde2081f12bb0f6900c49ee32894cc7c868913b/pkg/arvo/sys/vane/ford.hoon#L4232-L4248) didn't add those though, so, wondering how that worked... (The current map generation code is borked anyway, considering it changes keys without changing map ordering.) Will experiment!

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -931,6 +974,29 @@
...
+      ::
+      =/  =type  (~(play ut p.sut) [%kttr spec.i.raz])
+      ::  ensure results nest in the specified type,
+      ::  and produce a homogenous map containing that type.
+      ::
+      :-  %-  ~(play ut p.sut)
+          [%kttr %make [%wing ~[%map]] ~[[%base %atom %ta] spec.i.raz]]
+      |-
+      ?~  res  ~
+      ~|  [%nest-fail path.i.raz p.n.res]
```
belisarius222, line 995: I think this will print a whole list of failures, instead of just the single nest failure.  Maybe put the nest check inside the head of the `:-` expression so that further recursive cases aren't subexpressions of the `~|`.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1582,12 +1648,19 @@
...
-      |=  [cache=(map key [value dez=(set path)]) invalid=(set path)]
-      =/  builds=(list [key value dez=(set path)])  ~(tap by cache)
+      |=  [cache=(map key [value dez=(set [dir=? =path])]) invalid=(set path)]
+      =/  builds=(list [key value dez=(set [dir=? =path])])
+        ~(tap by cache)
       |-  ^+  cache
       ?~  builds
         ~
-      ?:  ?=(^ (~(int in dez.i.builds) invalid))
+      ?:  %+  lien  ~(tap in dez.i.builds)
```
belisarius222, line 1657: I think this and the next `+lien` call could both be replaced with `~(any in ...)`.

belisarius222 (changes requested):
A couple nitpicks, but LGTM other than that.

belisarius222 (approved):
LGTM

PR #4747, opened 2021-04-09
eyre + interface: use font-display: swap; to reduce time to paint text

tacryt-socryp:
https://web.dev/font-display/?utm_source=lighthouse&utm_medium=unknown

Fixes https://github.com/urbit/landscape/issues/762

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -249,6 +249,7 @@
                  font-family: "Source Code Pro";
                  src: url("https://storage.googleapis.com/media.urbit.org/fonts/scp-regular.woff");
                  font-weight: 400;
+                 font-display: swap;
              }
              :root {
                --red05: rgba(255,65,54,0.05);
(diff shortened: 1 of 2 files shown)


tylershuster:
Why are we still serving from media.urbit.org? I thought we moved these to local host

matildepark:
> Why are we still serving from media.urbit.org? I thought we moved these to local host

Those two fonts don't actually exist on localhost, so they were constantly hitting 404 in console, so I removed the localhost lines. The others do exist. Note that we should serve from urbit.org as a fallback.

matildepark (approved):
lgtm, can probably just go out with https://github.com/urbit/urbit/pull/4748 since it changes eyre (however trivially).

liam-fitzgerald (approved):
LGTM! Wondering if this should be split into two commits and deployed seperately, given that the majority of the improvement here is on userspace side

tacryt-socryp:
Seems safe to me regardless. We primarily care about separating sys and userspace deploys so that things affecting vane logic can be properly debugged etc

tacryt-socryp:
is this ready?

PR #4752, opened 2021-04-12
leap: deduplicate selection with mouse+keys

jamesacklin:
This PR proposes we unify the selection in Leap between hovered (mouse) and keyboard selections by doing the following:
- Set the active selection, not just hover treatment, by mouse among leap results
- Determine an item's selection-eligibility by mouseMove event (not hover) to account for items scrolling into view under the cursor
- Hide the mouse cursor while navigating results by keyboard; restore the cursor on mouse move

Reformatting handled in a separate commit.

Video demonstrating behavior attached.

https://user-images.githubusercontent.com/748181/114450025-7d5aa880-9ba3-11eb-9cf8-5deb200cabb9.mp4

Fixes urbit/landscape#598

### DIFF
--- a/pkg/interface/src/views/components/leap/Omnibox.tsx
+++ b/pkg/interface/src/views/components/leap/Omnibox.tsx
@@ -62,7 +62,7 @@ export function Omnibox(props: OmniboxProps): ReactElement {
   const contacts = useMemo(() => {
     const maybeShip = `~${deSig(query)}`;
     const selflessContactState = omit(contactState, `~${window.ship}`);
-    return ob.isValidPatp(maybeShip)
+    return ob.isValidPatp(maybeShip) && maybeShip !== `~${window.ship}`
       ? { ...selflessContactState, [maybeShip]: {} }
       : selflessContactState;
   }, [contactState, query]);
--- a/pkg/interface/src/views/components/leap/Omnibox.tsx
+++ b/pkg/interface/src/views/components/leap/Omnibox.tsx
@@ -210,6 +210,10 @@ export function Omnibox(props: OmniboxProps): ReactElement {
     }
   }, [selected, results]);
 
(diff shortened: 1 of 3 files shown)


tylershuster (approved):
Does what it says on the tin

matildepark:
This actually doesn't solve the issue in question.

image

liam-fitzgerald (approved):
Changes read well, formatting looks great! LGTM!

urcades (approved):

jamesacklin:
Results that are `ships` with titles that directly match `window.ship` are now removed, which enable users to move past the "Profile" result, which was blocking advancement.

image

matildepark:
I think it would be better to remove it from here, where we generate it per search query and inject valid patps as results: https://github.com/urbit/urbit/blob/e80101ab0a316d64f4bd9ea4143821a8fdca8985/pkg/interface/src/views/components/leap/Omnibox.tsx#L63

pkg/interface/src/views/components/leap/Omnibox.tsx
```diff
@@ -60,9 +61,10 @@ export function Omnibox(props: OmniboxProps): ReactElement {
 
   const contacts = useMemo(() => {
     const maybeShip = `~${deSig(query)}`;
+    const selflessContactState = omit(contactState, `~${window.ship}`);
     return ob.isValidPatp(maybeShip)
-      ? { ...contactState, [maybeShip]: {} }
-      : contactState;
+      ? { ...selflessContactState, [maybeShip]: {} }
```
matildepark, line 66: This pushes the maybeship into the contact state if it's a valid patp, so we'll still show up. 

You want it to be like "if this is a valid patp *and* it's not us, push it in", not removing us as a contact in the state, because we push any valid patp into the list, not just our contacts as an index.

PR #4762, opened 2021-04-14
docker: add exec to final urbit invocation

wexpertsystems:
This lets docker's SIGTERM pass through to urbit so it stops cleanly.

Previously it was taking 7-8 seconds to stop and occasionally left a `.vere.lock` file behind when it did. After adding `exec` it stops in <1 sec and seems to clear the lock properly. I also think the lack of this might have been truncating the error messages during some crashes I was having in the past week.


eamsden (approved):
This is the correct semantics and should be regarded as a bugfix.

jtobin (approved):

PR #4764, opened 2021-04-14
graph-store: make it referentially transparent

tacryt-socryp:
Everything touched compiles and every file in the system has been tested to build with `-test % ~`. Thus, we didn't miss any things that used `%graph-store`.

I still need to integrate this with the frontend.

Provides the prerequisite backend code for: https://github.com/urbit/landscape/issues/778

Fixes https://github.com/urbit/landscape/issues/605

Almost completely solves https://github.com/urbit/landscape/issues/604 (note that until the graph namespace itself is fully referentially transparent and `%gall` allows apps to publish referentially transparent data this won't be 100%)


pkg/arvo/app/graph-pull-hook.hoon
```diff
@@ -9,7 +9,7 @@
       update:store
       %graph-update
       %graph-push-hook
-      1  1
+      2  2
```
liam-fitzgerald, line 12: Pull-hook could probably be backwards compatible with old publishers. Doesn't have to be though.

pkg/arvo/app/graph-push-hook.hoon
```diff
@@ -281,36 +283,46 @@
...
-    (get-permission permissions is-admin writers)
-  ?-  permission-level
-      %yes  %.y
-      %no   %.n
-    ::
-        %self
-      =/  parent-node=node:store
-        (got-node:gra resource parent-index)
-      =(author.post.parent-node src.bowl)
+  ?-  -.post.node
```
liam-fitzgerald, line 303: Feels weird as a `?-`, wouldn't a `?:` or `?.` be better, given there are only two paths here

pkg/arvo/app/graph-push-hook.hoon
```diff
@@ -322,17 +334,23 @@
...
-  =/  =permissions:store
-    %^  remove-mark  resource  vip
-    (node-to-indexed-post node)
-  =/  =permission-level:store
-    (get-permission permissions is-admin writers)
-  ?-  permission-level
-    %yes   %.y
-    %no    %.n
-    %self  =(author.post.node src.bowl)
+  ?-  -.post.node
```
liam-fitzgerald, line 340: same q

pkg/arvo/app/graph-store.hoon
```diff
@@ -412,90 +316,91 @@
...
-            (get-descendants p.children.u.rm-node)
-          [(~(uni in indices) new-indices) graph]
-        =/  =node:store
+          =/  =node:store
+            ~|  "cannot remove index that does not exist {<index>}"
+            (need (get:orm graph atom))
+          ~|  "cannot remove post that has already been removed"
+          ?>  ?=(%& -.post.node)
+          =*  p  p.post.node
+          %=    node
```
liam-fitzgerald, line 374: `%_` is preferable here

liam-fitzgerald (changes requested):
Some style nits, a missing migration, looks good otherwise

pkg/arvo/app/hark-store.hoon
```diff
@@ -202,7 +202,7 @@
   ++  convert-graph-contents-4
     |=  con=(list post:post-zero:post)
     ^-  (list post:post)
-    (turn con post-to-one:graph-store)
```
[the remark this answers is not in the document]
liam-fitzgerald: Wait that's a lie, because hark only store the `post` type, carry on

pkg/arvo/app/graph-store.hoon
tacryt-socryp: Not possible since it changes the type of maybe-post from the %y branch to the %n branch of the mint

pkg/arvo/app/graph-push-hook.hoon
tacryt-socryp: Using the ?- feels the most explicit to me, but I'm happy to be argued otherwise.

pkg/arvo/app/chat-store.hoon
```diff
@@ -1,334 +1,28 @@
...
-  ^-  graph:graph-store
-  %+  gas:orm:graph-store  *graph:graph-store
-  (turn envelopes.mailbox envelope-to-node)
+|_  =bowl:gall
++*  this       .
+    def        ~(. (default-agent this %|) bowl)
+::
+++  on-init   on-init:def
+++  on-save   !>(~)
+++  on-load
```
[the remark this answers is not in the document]
tacryt-socryp: This is an @ixv question. I'll do whatever he says

pkg/arvo/app/graph-store.hoon
joemfb: I have no object level opinion here, but it's worth noting that `&` or `|` don't have this problem -- their type is `?(%.y %.n)` (see also `.y` or `.n`, which i've never actually seen used ...).

tacryt-socryp: Oh you may have misunderstood.. this is just an each, so the actual values in question are %| and %&. I just colloquially referred to them as %y and %n (imprecise!)

joemfb: My bad, I was just skimming the comments. (If you *want* to use `%_`, you'll probably need to move `?> ?=(%& -.post.node)` inside it ...)

pkg/arvo/app/graph-push-hook.hoon
liam-fitzgerald: I just dislike the extra indentation level stylistically, if it can be backstepped, it probably should be. Especially because one hoon is fair lighter than the other. Up to you though.

pkg/arvo/app/graph-store.hoon
liam-fitzgerald: Oh right, didn't catch the ?>

pkg/arvo/app/graph-push-hook.hoon
belisarius222: I'd probably also do this as
```
?:  ?=(%| -.post.node)  %.n
```
and then an unindented remainder, but if you do keep the `?-`, standard style is:
```
?-    -.post.node
    %|  %.n
    %&
  ...
==
```

pkg/arvo/app/chat-store.hoon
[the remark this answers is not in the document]
matildepark: I sorta figured flagday imported at a specific OTA, not at whatever latest is.

pkg/arvo/app/graph-push-hook.hoon
tacryt-socryp: Resolved to use `?:` as requested in all places it is used (that I could find)

pkg/arvo/app/graph-store.hoon
tacryt-socryp: Resolved to use the `?>` inside of a `%_` instead of using the `%=`. Thanks all!

tacryt-socryp:
@liam-fitzgerald Now targeting a pre-release branch so that I can organize my frontend changes in a separate PR for legibility. I'll merge this into the pre-release branch as soon as I get a stamp. I've addressed all requested changes.

pkg/arvo/app/graph-store.hoon
```diff
@@ -412,90 +318,91 @@
...
-          ?~  rm-node  `graph
-          ?.  ?=(%graph -.children.u.rm-node)
-            `graph
-          =/  new-indices
-            (get-descendants p.children.u.rm-node)
-          [(~(uni in indices) new-indices) graph]
-        =/  =node:store
+          =/  =node:store
+            ~|  "cannot remove index that does not exist {<index>}"
+            (need (get:orm graph atom))
```
liam-fitzgerald, line 372: Do we not have `got:orm`?

pkg/arvo/app/graph-store.hoon
```diff
@@ -41,19 +37,18 @@
   ^-  (quip card _this)
   =+  !<(old=versioned-state old-vase)
   =|  cards=(list card)
-  |^
+  |-
   ?-    -.old
-      %0  
+      %0
+    =*  zro  zero-load:upgrade:store
```
liam-fitzgerald, line 43: Can't we nix state-zero? iirc it was before the breach

liam-fitzgerald (approved):
Looks good! Some questions:
- Has this migration been run on a `-L` ?
- What else is supposed to go into the `pre/ref-transparency` branch?
- When is this getting deployed?

matildepark:
This will go out either this week or next depending on our confidence. We’ll be doing migrations and testing them today and onward. 

— 
~haddef-sigwen
https://urbit.org

pkg/arvo/app/graph-store.hoon
tacryt-socryp: We do not have it until I merge #4788

pkg/arvo/app/graph-store.hoon
tacryt-socryp: I'll look at the history and see if that is the case. If so, I'll get rid of it.

tacryt-socryp:
@liam-fitzgerald yes this has been migrated on a -L

matildepark:
Tested on `-L`s on Ropsten. Data successfully migrated; there was one issue with deleting nodes predating this change that was patched in https://github.com/urbit/urbit/pull/4764/commits/42a1a68913a390d87bab61457c0779f66420317b.

@tacryt-socryp — CI still fails, presumably we need to amend tests.

PR #4765, opened 2021-04-14
sidebar: typesafe notifications check

tylershuster:
Fixes a bug where the "this channel has notifications" dot would not appear when it should have


matildepark:
Why ... is it either one?

tylershuster:
Yup. `NotifRef[] | number`

Maybe @liam-fitzgerald had some plans for unifying these but for now it can be either. Maybe it's only `NotifRef[]` now but it has been number in the past.

matildepark (approved):

PR #4748, opened 2021-04-09
metadata-store: subscribe to all feeds that got missed in the upgrade

tacryt-socryp:
Group Pull Hook `%add` is idempotent. This is a very calm upgrade TM.

PR #4770, opened 2021-04-15
metadata-store: actually fix it, strongly type the action sent to %pull-hook to prevent moronic errors

tacryt-socryp:
I should have cast the action I passed into the !> in #4748. Apologies, I messed up.

### DIFF
--- a/pkg/arvo/app/metadata-store.hoon
+++ b/pkg/arvo/app/metadata-store.hoon
@@ -23,7 +23,7 @@
 ::  /app-name/%app-name                            associations for app
 ::  /group/%path                             associations for group
 ::
-/-  store=metadata-store
+/-  store=metadata-store, pull-hook
 /+  default-agent, verb, dbug, resource, *migrate
 |%
 +$  card  card:agent:gall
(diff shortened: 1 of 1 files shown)


joemfb:
Is there anything useful that the metadata-hook could've been with the negative poke-ack it was getting here?

tacryt-socryp:
@joemfb I was just irresponsible with translating the work I had done in `+poke-noun` into `+on-load`. This was a deterministic failure, not really something that we would want to meaningfully recover from or use the poke nack on.

tacryt-socryp:
@matildepark I've tested this on ~littel-wolfur's fiance's livenet ship and the updated version does indeed fix all group feeds.

liam-fitzgerald (approved):
LGTM!

liam-fitzgerald:
This illustrates a pressing need to abstract the creation of different cages in a typesafe and concise manner.
```hoon
::  lib/pull-hook.hoon
|%
++  cg
  |%
  ++  action
     |=  =update:store
     :-  %pull-hook-action
     !>(update)
  --
--
::  at callsite
(poke-our:pass %metadata-pull-hook (action:cg:pull-hook %add [entity .]:rid))
```
Something like this pattern, where we have a `cg` core in the helper lib that exposes cage-creation functions as its arms would go a long way towards preventing these kinds of silly mistakes, and also help reduce churn due to new mark versions

matildepark:
@liam-fitzgerald tested this last night on a -L and also confirmed it sends the adds. Still, I'm gonna opt to hold for Monday since this is an edge case, not a universal issue.

tacryt-socryp:
can we push this out today?

matildepark:
I intend to slate it, yes.

—
~haddef-sigwen
https://urbit.org

PR #4771, opened 2021-04-15
docker: allow setting Ames port via docker run argument

wexpertsystems:
This allows the docker container Ames port to be optionally overridden by passing a `--port=xxx` argument along with the `docker run` command. If no port argument is passed the default 34343 is still used instead.

Example: 
```
docker run -p 80:80 -p 13436:13436/udp \
     --mount type=volume,source=/home/user/urbit/galaxy,destination=/urbit \
     tloncorp/urbit --port=13436
```

This is fairly edge case and probably not needed by many users but adds a little flexibility and it can suppress the warnings galaxies get:
```
ames: czar: overriding port 13436 with -p 34343
ames: czar: WARNING: 13436 required for discoverability
```

### DIFF
--- a/nix/pkgs/docker-image/default.nix
+++ b/nix/pkgs/docker-image/default.nix
@@ -13,7 +13,7 @@ let
     do
     case $i in
         -p=*|--port=*)
-            amesPort="${i#*=}"
+            amesPort="''${i#*=}"
             shift
             ;;
     esac
--- a/nix/pkgs/docker-image/default.nix
+++ b/nix/pkgs/docker-image/default.nix
@@ -5,6 +5,20 @@ let
 
     set -eu
 
+    # set defaults
+    amesPort=${toString amesPort}
+
+    # check args
+    for i in "$@"
+    do
+    case $i in
+        -p=*|--port=*)
+            amesPort="${i#*=}"
+            shift
+            ;;
+    esac
+    done
+
     # If the container is not started with the `-i` flag
     # then STDIN will be closed and we need to start
     # Urbit/vere with the `-t` flag.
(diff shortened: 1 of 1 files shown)


eamsden (approved):

nix/pkgs/docker-image/default.nix
```diff
@@ -5,6 +5,20 @@ let
...
+    amesPort=${toString amesPort}
+
+    # check args
+    for i in "$@"
+    do
+    case $i in
+        -p=*|--port=*)
+            amesPort="${i#*=}"
+            shift
+            ;;
```
joemfb, line 18: Per CI, this is a *nix* syntax error. Not sure how it should be escaped.

joemfb (changes requested):

nix/pkgs/docker-image/default.nix
wexpertsystems: Not sure, I tested this by updating the script directly in a container. Maybe @eamsden knows how?

nix/pkgs/docker-image/default.nix
```diff
@@ -5,6 +5,20 @@ let
...
 
+    # set defaults
+    amesPort=${toString amesPort}
+
+    # check args
+    for i in "$@"
+    do
+    case $i in
+        -p=*|--port=*)
+            amesPort="${i#*=}"
```
eamsden, line 16: The error is actually here,  you need to escape the nix antiquotation by prefixing `$` with two single quotes per
https://nixos.org/manual/nix/stable/#idm140737322057376
@wexpertsystems @joemfb 
```suggestion
            amesPort="''${i#*=}"
```

joemfb (approved):
LGTM

PR #4772, opened 2021-04-15
meta/interface: re-integrate preinstall script with clean install

matildepark:
The original commit was reverted hot on next-js because the auto-deploy script was crashing on some uncompiled `../npm/api` files, in the wake of package-lock requiring it to actually grab the `../npm/api` folder. 

This is because the npm in the docker instance we use to auto-deploy can't leave its pwd in an automated mode, we will need to amend docker or the package.json to let npm do that for installing these local dependencies. [See here](https://stackoverflow.com/questions/18136746/npm-install-failed-with-cannot-run-in-wd).

However with and without this preinstall script there's an ongoing issue in developing on Landscape where the npm cache across these four packages can get into some wild desynchronisation, requiring some [truly silly incantations](https://github.com/urbit/indigo-react/issues/101) to clear those caches afterward and try again. I believe this is because all the node_modules in the script are changing on the fly.

By using `npm ci` instead of `npm i` we prevent touching those package-locks or package.json files, which has, at least in some brief testing, resolved those errors, but we'll have to see after this. I also amend contributing docs to also advise toward `npm ci` unless changing dependencies.

### DIFF
--- a/pkg/interface/CONTRIBUTING.md
+++ b/pkg/interface/CONTRIBUTING.md
@@ -32,7 +32,7 @@ same (if [developing on a local development ship][local]). Then, from
 'pkg/interface':
 
 ```
-npm install
+npm ci
 npm run start
 ```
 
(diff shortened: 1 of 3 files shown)


matildepark:
Also, FYI, I leave the `build:prod` command with the bad hack in here for now so that janeway's auto-deploy can still run until it gets patched.

tacryt-socryp (approved):

PR #4778, opened 2021-04-18
ed: visual tweaks to notifications and sidebar 

urcades:
contained in this PR:

Purely visual fixes to two small issues:

1. Incorrect border radius in notification items (the little "reference elbow" that shows were an invite is coming from)
2. Sidebar coloration: After actually trying out `black80`, it's still too subtle. I think what we had before was good, in that no one was complaining about sidebar contrast. It looks a little dull/inactive at first, but you quickly get acclimated.

Reference Elbow Before:

Screen Shot 2021-04-16 at 6 51 53 PM

Screen Shot 2021-04-16 at 6 50 15 PM

Reference Elbow After:

Screen Shot 2021-04-18 at 1 35 19 PM

~

Sidebar Before:

Screen Shot 2021-04-18 at 1 35 02 PM

Sidebar After:

Screen Shot 2021-04-18 at 1 34 55 PM

### DIFF
--- a/pkg/interface/src/views/components/Invite/Group.tsx
+++ b/pkg/interface/src/views/components/Invite/Group.tsx
@@ -64,7 +64,7 @@ function Elbow(
     >
       <Box
         border="2px solid"
-        borderRadius={2}
+        borderRadius={3}
         borderColor={color}
         position="absolute"
         left="0px"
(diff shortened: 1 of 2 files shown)


urcades:
I wanted to catch + fix more visual bugs, but this is fine.

matildepark (approved):

PR #4780, opened 2021-04-19
messages: surface invite action in group DMs

jamesacklin:
This surfaces an invite action in group DMs such that anyone in the group DM can invite another ship to the resource. In 1:1 DMs, the same action appears, but pre-populates the `ShipSeach` component with the other ship and creates a new group DM from the resulting list of ships.

Opening this as a draft as there are still some active discussions on how to handle the ghost ship case.

Fixes urbit/landscape#497


pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
```diff
@@ -172,7 +218,7 @@ export function ResourceSkeleton(props: ResourceSkeletonProps): ReactElement {
           flexShrink='0'
           ref={actionsRef}
         >
-          {canWrite && <WriterControls />}
+          <ExtraControls />
```
liam-fitzgerald, line 221: `{ExtraControls()}`

liam-fitzgerald, line 221: `{ExtraControls()}`

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
```diff
@@ -121,26 +124,69 @@ export function ResourceSkeleton(props: ResourceSkeletonProps): ReactElement {
...
-  let actionsWidth = 0;
-
-  if (actRef) {
-    actionsWidth = actRef.clientWidth;
-  }
+  useEffect(() => {
+    if (actionsRef.current) {
+      setActionsWidth(actionsRef.current.clientWidth);
+    }
+  }, [actionsRef]);
```
liam-fitzgerald, line 189: Wouldn't it be easier to just use a callback ref to call the setter directly?

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
liam-fitzgerald, line 221: `{ExtraControls()}`

liam-fitzgerald (changes requested):

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
jamesacklin: Far easier, thank you

jamesacklin:
This is G2G if folks want to comment on the behavior, or shelve the 1:1 invitation until we figure something better out.

matildepark:
Shelve the 1:1 implementation, but inviting to group DMs is critical. Is that already how it's implemented here?

tacryt-socryp (commented):
Will be ready to go once it no longer adds this option to 1:1 DMs.

pkg/interface/src/views/landscape/components/NewChannel.tsx
```diff
@@ -31,33 +35,38 @@ type FormSchema = {
...
+    writers: members ? shipSearchSchemaInGroup(members) : shipSearchSchema,
+    writePerms: Yup.string()
+  });
 
 interface NewChannelProps {
   api: GlobalApi;
   group?: string;
   workspace: Workspace;
+  borderRadius?: number;
+  existingMembers?: Array;
```
liam-fitzgerald, line 53: Array of what? Also fwiw, I'd much prefer the style of
```typescript
type NewChannelProps = {
  api: GlobalApi;
  group?: string;
  workspace: Workspace;
  existingMembers: string[];
} & PropFunc<typeof Col>;
// in component
const { api, group, workspace, existingMembers, ...rest } = props;
// in JSX
return (
  <Col
      {...rest}
   //.... rest of components
```

because it's cleaner and allows for styled-systems affordances (short prop names, responsive arrays). PropFunc is a type helper to get the props of a functional component, it's in `~/types`

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
```diff
@@ -121,26 +124,59 @@ export function ResourceSkeleton(props: ResourceSkeletonProps): ReactElement {
...
   );
 
-  const actRef = actionsRef.current;
-  let actionsWidth = 0;
-
-  if (actRef) {
-    actionsWidth = actRef.clientWidth;
-  }
+  const actionsRef = useCallback((actionsRef) => {
+    setActionsWidth(actionsRef?.getBoundingClientRect().width);
```
liam-fitzgerald, line 178: +1 for callback refs if you don't need persistence

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
```diff
@@ -172,7 +208,7 @@ export function ResourceSkeleton(props: ResourceSkeletonProps): ReactElement {
           flexShrink={0}
           ref={actionsRef}
         >
-          {canWrite && <WriterControls />}
+          {ExtraControls()}
```
liam-fitzgerald, line 211: +1 for no closures as components

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
```diff
@@ -121,26 +124,59 @@ export function ResourceSkeleton(props: ResourceSkeletonProps): ReactElement {
...
+    if (canWrite) {
+      return (
+        <Link to={resourcePath('/new')}>
+          <Text bold pr='3' color='blue'>
+            + New Post
+          </Text>
+        </Link>
+      );
+    }
+    return <></>;
```
liam-fitzgerald, line 168: `return null` is more idiomatic when producing nothing to render

liam-fitzgerald (approved):
Looks good, some comments about types

pkg/interface/src/views/landscape/components/ResourceSkeleton.tsx
jamesacklin: Thank you for reading the React source code to uncover this wisdom

pkg/interface/src/views/landscape/components/NewChannel.tsx
jamesacklin: Makes sense, thank you

jamesacklin:
Pretty sure this solves the mystery hamburger menu bounce, would appreciate a checkout and verify.

matildepark (changes requested):
It looks good _but_ `+Add Ship` shows in group DMs where we are not the admin, so we can "send an invite" that errors out. You'll want to check and see if we're an admin of the hidden group.

matildepark:
I was gonna review but ... this has a conflict again. Sorry.

PR #4781, opened 2021-04-19
interface: correctly scroll lock document

liam-fitzgerald:
Even if overflow is hidden, certain mobile browsers may interpret a
scroll event to be on the document root, leading to strange and
inconsistent scrolling behaviours. Addresses this by adding
touch-action: none, which explicitly instructs browers to ignore this
element in scroll calculations.

### DIFF
--- a/pkg/interface/src/views/App.js
+++ b/pkg/interface/src/views/App.js
@@ -53,6 +53,7 @@ const Root = withState(styled.div`
   }
   display: flex;
   flex-flow: column nowrap;
+  touch-action: none;
 
   * {
     scrollbar-width: thin;


matildepark:
What’s this in reference to? Any example?

tacryt-socryp (approved):
I imagine this could solve some of the scroll jumping on mobile.

liam-fitzgerald:
It's most obvious in PWA mode, where if you try and scroll, safari will occasionally think you're trying to scroll the whole app, instead of the content. Image for example:
image

matildepark:
Ah ok. Understand, seems sane.

matildepark (approved):

PR #4783, opened 2021-04-19
launch: fix reversed destructure

liam-fitzgerald:
This was causing the tutorial to never be shown. This change should be
deployed with a new pill, to ensure comets show the tutorial.


matildepark (approved):
Damn yeah. This should go out with next hotfix (after weekly probably).

PR #4785, opened 2021-04-19
interface: fix #591 message display by time-sent in chat

tacryt-socryp:
https://github.com/urbit/landscape/issues/591


matildepark (approved):
I went back to see the original incident in urbit/landscape#591 and it looks like this as of this PR:

Screenshot 2021-04-19 at 5 26 30 PM

That means that we are both sorting and representing messages as per the host's timestamp. That is sane to me.

PR #4786, opened 2021-04-19
interface: group feed no longer shows in omnibox

tacryt-socryp:
Fixes https://github.com/urbit/landscape/issues/721

### DIFF
--- a/pkg/interface/src/logic/lib/omnibox.js
+++ b/pkg/interface/src/logic/lib/omnibox.js
@@ -102,7 +102,7 @@ export default function index(contacts, associations, apps, currentGroup, groups
   }).map((e) => {
     // iterate through each app's metadata object
     Object.keys(associations[e])
-      .filter((association) => !associations[e][association].metadata.hidden)
+      .filter((association) => !associations?.[e]?.[association]?.metadata?.hidden)
       .map((association) => {
         const each = associations[e][association];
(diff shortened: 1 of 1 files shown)


pkg/interface/src/logic/lib/omnibox.js
```diff
@@ -99,9 +99,11 @@ export default function index(contacts, associations, apps, currentGroup, groups
   Object.keys(associations).filter((e) => {
     // skip apps with no metadata
     return Object.keys(associations[e]).length > 0;
-    }).map((e) => {
-      // iterate through each app's metadata object
-      Object.keys(associations[e]).map((association) => {
+  }).map((e) => {
+    // iterate through each app's metadata object
+    Object.keys(associations[e])
+      .filter((association) => !associations[e][association].metadata.hidden)
```
matildepark, line 105: This should probably be a bit safer I think, it's quite nested

tacryt-socryp: Accessing using the same pattern found in the following map? Not sure what else to do.

matildepark:
poke @tacryt-socryp

tacryt-socryp:
Totally forgot about this, will do

tacryt-socryp:
okey doke @matildepark

matildepark (approved):

PR #4787, opened 2021-04-19
contact-store: disallow data: urls from being added

tacryt-socryp:
Fix https://github.com/urbit/landscape/issues/675 for all new data, no migration needed.

### DIFF
--- a/pkg/arvo/app/contact-store.hoon
+++ b/pkg/arvo/app/contact-store.hoon
@@ -126,6 +126,14 @@
               !=(contact(last-updated *@da) u.old(last-updated *@da))
           ==
         [~ state]
+      ~|  "cannot add a data url to cover!"
+      ?>  ?|  ?=(~ cover.contact)
+              !=('data:' (cut 3 [0 5] u.cover.contact))
+          ==
+      ~|  "cannot add a data url to avatar!"
+      ?>  ?|  ?=(~ avatar.contact)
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/contact-store.hoon
```diff
@@ -126,6 +126,14 @@
               !=(contact(last-updated *@da) u.old(last-updated *@da))
           ==
         [~ state]
+      ~|  "cannot add a data url to cover!"
+      ?>  ?|  ?=(~ cover.contact)
```
liam-fitzgerald, line 130: What's the point of this check?

tacryt-socryp: Either there is no cover image OR we check to make sure the cover image does not begin with the data url schema

liam-fitzgerald (approved):

ritfed-tabryg:
I imagine the rationale here was to prevent someone from setting a megabyte-size base64-encoded data URL as their profile pic, but it's unfortunate this blocks SVG profile images too, which can be pretty small. Would the Landscape team be open to allowing data URLs of sizes under, say, 4096 bytes?

liam-fitzgerald:
@ritfed-tabryg there's nothing stopping you from setting an externally hosted SVG as a profile pic. The reason this was pulled out was due to concerns about memory pressure, so I'd defer to @joemfb about what sane default for potential image size should be.

PR #4788, opened 2021-04-19
zuse: add +all, +any, +got, +has, +run, and +tab to ++ordered-map, jet hints, and reorganize alphabetically

tacryt-socryp:
I looked through the usage of `++ordered-map` in `%graph-store` and determined that I could cut out about half of all traversals by using methods akin to `+all` and `+run`, so I am now adding the following to `%zuse`'s `++  ordered-map` implementation.

Fixes https://github.com/urbit/landscape/issues/674

The `+tab` arm paves the way for fixing a long-standing TODO in `%graph-store` on the time complexity of `%node-siblings`, which is in the hot path of how chat fetches messages and also improves the time complexity of the upcoming deep subset scry in `%graph-store`.


pkg/arvo/sys/zuse.hoon
```diff
@@ -5296,6 +5296,23 @@
...
+    [n=[key.n.a (b val.n.a)] l=$(a l.a) r=$(a r.a)]
+  ::  +all: apply logical AND boolean test on all values
+  ::
+  ++  all
+    |=  [a=(tree item) b=$-(item ?)]
+    ^-  ?
+    |-
+    ?~  a
+      &
+    ?&((b n.a) $(a l.a) $(a r.a))
```
philipcmonk, line 5315: Would be pretty easy to make this short-circuit

belisarius222 (changes requested):
These arms should have jet hints.  They're decent candidates for jetting.  LGTM otherwise.

pkg/arvo/sys/zuse.hoon
tacryt-socryp: What do you mean? It will end early if any check returns false.

philipcmonk: It'll not look at its children, but if the top node is true, then even if the left branch is false, the right branch still gets run.

philipcmonk: wait no i'm wrong aren't it

philipcmonk: Yeah no this is correct

tacryt-socryp: Interesting, okay. Worth noting that I took this directly from +in in hoon.hoon, so that should short-circuit earlier too. Likely also the one in +by

joemfb:
```
> &(& ~&(%foo |))
%foo
%.n
> 
> &(| ~&(%foo |))
%.n
```

joemfb (commented):
As @belisarius222 says, these should hints.

While you're adding arms, you might want to consider `+any` and `+rut`.

pkg/arvo/sys/zuse.hoon
joemfb: If it was not short-circuiting, that second example would also slog.

philipcmonk: > Yeah no this is correct
By this I meant your code is correctly short-circuitying

pkg/arvo/sys/zuse.hoon
```diff
@@ -5174,39 +5239,104 @@
...
+    ?:  (mor key.n.a key.n.r)
+      a(r r)
+    r(l a(r l.r))
+  ::  +run: apply gate to transform all values in place
   ::
-  ++  nip
-    |=  a=(tree item)
+  ++  run
+    ~/  %run
+    |=  [a=(tree item) b=$-(val *)]
```
belisarius222, line 5273: This will lose the product type unless it's wet.

tacryt-socryp: Does this suffice?

`|*  [a=(tree item) b=$-(val *)]`

pkg/arvo/sys/zuse.hoon
```diff
@@ -5094,9 +5094,29 @@
   ::
   |=  compare=$-([key key] ?)
   |%
```
joemfb, line 5096: You'll need to put a fast hint on the whole core as well. This should work:

```
~%  %ordered-map  ..part  ~
|%
```

pkg/arvo/sys/zuse.hoon
```diff
@@ -5070,7 +5070,7 @@
   |=  ord=$-([key key] ?)
   |=  a=*
   =/  b  ;;((tree [key=key val=value]) a)
-  ?>  (check-balance:((ordered-map key value) ord) b)
+  ?>  (apt:((ordered-map key value) ord) b)
   b
 ::
 ::  $mk-item: constructor for +ordered-map item type
```
joemfb, line 5076: While you're cleaning things up generally, I don't see any reason for `+mk-item` to be a top-level name. I'd just replace `$item` below:

```
+$   item  [=key =val]
```

joemfb (changes requested):
This general cleanup seems well worth doing. I'll read more closely, but two things jump out:

pkg/arvo/sys/zuse.hoon
tacryt-socryp: I still get parent not found, but I'm not sure if that's expected or not.

joemfb: For which label; what's an example of the output? (Also, you might to `:goad %force` or something to ensure everything's rebuilt.)

joemfb: Once you've been messing around with the stuff, you really need to make a pill and boot from that to get a clean test.

I forgot that the %zuse core has a hint now, so we could try going straight to the right way. I think that would look like this:

```

++  ordered-map
  ~/  %ordered-map
  |*  [key=mold val=mold]
  =>  |%
      +$  item  [key=key val=val]
      --
  ::  +compare: item comparator for horizontal order
  ::
  ~%  %comp  +>+  ~
  |=  compare=$-([key key] ?)
  ~%  %core    +  ~
  |%
```

/cc @frodwith

tacryt-socryp: Booting from a solid pill formed with the changes you just specified, I still receive the parent not found:
Screen Shot 2021-04-20 at 3 01 20 PM

pkg/arvo/sys/zuse.hoon
```diff
@@ -5085,18 +5082,41 @@
...
+    |-  ^-  ?
+    ?~  a
+      |
+    ?|((b n.a) $(a l.a) $(a r.a))
+  ::  +apt: verify horizontal and vertical orderings
   ::
-  ++  check-balance
+  ++  apt
+    ~/  %apt
     =|  [l=(unit key) r=(unit key)]
```
joemfb, line 5120: This hint is wrong. `~/  %foo` is `~%  %foo  +>  ~`, but, in this case, the parent core is at `+>+`, not `+>`.

You could fix the hint, or move `l` and `r` inside the gate and add an inner `|-`. That's probably the right thing to do for jetting anyway, since`l` and `r` should be controlled by the caller.

joemfb (changes requested):

pkg/arvo/sys/zuse.hoon
tacryt-socryp: Fixed, thanks.

pkg/arvo/sys/zuse.hoon
tacryt-socryp: @belisarius222 ???

joemfb: That's right.

tacryt-socryp:
@belisarius222 @joemfb I've now applied all of the suggestions in the review comments, so this is ready to go.

belisarius222 (approved):
LGTM

tacryt-socryp:
@joemfb would love a final pass on this so I can get it merged

pkg/arvo/sys/zuse.hoon
```diff
@@ -5339,11 +5263,103 @@
...
+    ::
+    |-  ^-  (tree item)
+    ?~  l.a  r.a
+    ?~  r.a  l.a
+    ?:  (mor key.n.l.a key.n.r.a)
+      l.a(r $(l.a r.l.a))
+    r.a(l $(r.a l.r.a))
+  ::  +peek: produce head (smallest item) or null
+  ::
+  ++  peek
```
joemfb, line 5297: They're not not cvc,  but `+pry` or `+see` would at least by three letters.

pkg/arvo/sys/zuse.hoon
```diff
@@ -5339,11 +5263,103 @@
...
+    ~/  %run
+    |*  [a=(tree item) b=$-(val *)]
+    |-
+    ?~  a  a
+    [n=[key.n.a (b val.n.a)] l=$(a l.a) r=$(a r.a)]
   ::  +subset: take a range excluding start and/or end and all elements
   ::  outside the range
   ::
   ++  subset
+    ~/  %subset
```
joemfb, line 5362: `+lot`?

pkg/arvo/sys/zuse.hoon
```diff
@@ -5122,61 +5144,22 @@
...
-    ?~  l.a  [n.a r.a]
+    ^-  (list item)
     ::
-    =/  l  $(a l.a)
-    :-  head.l
-    ::  load .rest.l back into .a and rebalance
+    =|  b=(list item)
+    |-  ^+  b
+    ?~  a  b
     ::
```
joemfb, line 5157: The blank lines here are little unusual.

joemfb (approved):
LGTM; substantial improvements IMO.

I've left some style nits below, nothing pressing.

I think a change of this magnitude merits a kelvin, but we should discuss that elsewhere.

PR #4790, opened 2021-04-20
metadata-store: prevent infinite loop

liam-fitzgerald:


philipcmonk (approved):
That would cause the bug we saw.  Should `%graph-pull-hook` or something else also have this check?

liam-fitzgerald:
Yeah, i'm just cautious of reloading the hooks until the kick behaviour is fixed (if you upgrade the libraries themselves, there's a bit too much network pressure)

PR #4791, opened 2021-04-20
notifications: slide to dismiss

liam-fitzgerald:
This PR introduces react-spring and react-use-gesture, an animation library and gesture recognition library. It uses those libraries to build a SwipeMenu, a component which is able to be left swiped on to reveal a secondary menu. This component is then used in the notifications inbox, as per the design spec. Feel free to bikeshed the specific physics of the slider, most of the configuration is defaults.

cc: @g-a-v-i-n


urcades:
Swiping to the left produces a visual artifact indicating the presence of the dismiss option - not sure if this is something we find desirable (it’s sort of like an indirectswipe hint, I suppose), but flagging it regardless as something to possibly remove, as it was not specified in the designs:


https://user-images.githubusercontent.com/1195363/115454688-10708f80-a1ef-11eb-9696-22d494e7656e.mov

urcades:
A broad observation I'm making in the current interaction/animation speed is that it's feeling a little sluggish - could we bump up the speed of the element in a straightforward manner? Perhaps increase the speed of animation by 50% or so?

urcades:
Also noting it's rather easy to scroll vertically while attempting to swipe horizontally, if there's any way we can reduce this, that would be wonderful.

matildepark:
> A broad observation I'm making in the current interaction/animation speed is that it's feeling a little sluggish - could we bump up the speed of the element in a straightforward manner? Perhaps increase the speed of animation by 50% or so?

This is the main blocker on this.

matildepark (approved):

PR #4792, opened 2021-04-20
king: do not commit files starting with a dot

liam-fitzgerald:
Brings the haskell king inline with the C king (see: https://github.com/urbit/urbit/blob/268677e74412d80eb5ebc9eed4c3add20a274f91/pkg/urbit/vere/io/unix.c#L845)
This discrepancy in behaviour breaks janeway. 

(Also, is this against the right branch?)

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Clay.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Clay.hs
@@ -37,12 +37,12 @@ textPlain = Path [(MkKnot "text"), (MkKnot "plain")]
 
 -- | Filter for dotfiles, tempfiles and backup files.
 validClaySyncPath :: FilePath -> Bool
-validClaySyncPath fp = hasPeriod && notTildeFile && notDotHash && notDoubleHash
+validClaySyncPath fp = hasPeriod && notTildeFile && notDotFile && notDoubleHash
   where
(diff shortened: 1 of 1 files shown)


pkg/hs/urbit-king/lib/Urbit/Vere/Clay.hs
```diff
@@ -42,7 +42,7 @@ validClaySyncPath fp = hasPeriod && notTildeFile && notDotHash && notDoubleHash
     fileName = takeFileName fp
     hasPeriod = elem '.' fileName
     notTildeFile = not $ "~" `isSuffixOf` fileName
-    notDotHash = not $ ".#" `isPrefixOf` fileName
+    notDotHash = not $ "." `isPrefixOf` fileName
```
pilfer-pandex, line 45:
```suggestion
    notDotFile = not $ "." `isPrefixOf` fileName
```

pilfer-pandex (approved):
Looks good modulo my one suggestion, which you should apply.

Thanks for bringing this to my attention, and thanks doubly for fixing it.

pilfer-pandex:
Btw how does this break *janeway*?

matildepark:
Janeway boots ships, commits and globs the JS blob for deployment. Committing dots broke a former ota.

PR #4793, opened 2021-04-20
hooks: don't kick every reload

liam-fitzgerald:
Prevents undue network pressure due to spurious kicking. Also prevents the looped pull-hook add issue at the source.

### DIFF
--- a/pkg/arvo/lib/push-hook.hoon
+++ b/pkg/arvo/lib/push-hook.hoon
@@ -249,12 +249,10 @@
       --
     ::
     ++  on-save
-      =.  inner-state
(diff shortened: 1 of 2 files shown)


tacryt-socryp (approved):
LGTM

pkg/arvo/lib/push-hook.hoon
```diff
@@ -232,6 +251,10 @@
     ++  on-save
       =.  inner-state
```
tacryt-socryp, line 252: Better to use `=:` here to write this all as one operation. Please do so.

PR #4794, opened 2021-04-20
interface: fast BigIntOrderedMap

liam-fitzgerald:
New ordered map implementation that is faster and more space
efficient. Stores the map and keys as a flat POJO, sorting them upon
iteration. The sorting cost is only typically paid once per render
however as the sorted array is cached until the map is mutated. This
implementation appears to play nicer with immer's structural sharing,
reducing the incidence of 'props are equal by value, but not by
reference'.


pkg/npm/api/lib/BigIntOrderedMap.ts
```diff
@@ -1,227 +1,63 @@
...
-  private root: MapNode<V> = null;
+  private root: Record<string, V> = {}
+  private cachedIter: [BigInteger, V][] | null = null;
   [immerable] = true;
-  size: number = 0;
 
-  constructor(initial: [BigInteger, V][] = []) {
-    initial.forEach(([key, val]) => {
+  constructor(items: [BigInteger, V][] = []) {
+    _.forEach(items, ([key, val]) => {
```
tylershuster, line 20: Couldn't you just do `items.forEach` and lose the lodash dependency?

tylershuster (commented):
Certainly easier to understand, although I never quite understood all the nesting in the original version. Seems easier to traverse. Was the old version just overbuilt?

pkg/npm/api/lib/BigIntOrderedMap.ts
tacryt-socryp: That would be nice.

tacryt-socryp (approved):
LGTM, if you can get rid of lodash, please do. Would love to see this out on next-js soon.

liam-fitzgerald:
Autofreeze prevents the memoisation technique, so we turn it off.

PR #4795, opened 2021-04-20
graph-store, metadata-store lib: skip trip and crip, use rap

tacryt-socryp:
Attempting to remove all uses of `+trip` and `+crip` from the hotpath that Landscape data moves through. These arms make quite a few allocations and can be replaced with `+rap` for better performance.

The change to the `%graph-store` lib is likely to have a larger effect.

### DIFF
--- a/pkg/arvo/lib/graph-store.hoon
+++ b/pkg/arvo/lib/graph-store.hoon
@@ -145,18 +145,17 @@
     ==
   ::
   ++  index
-    |=  i=^index
+    |=  ind=^index
     ^-  json
-    ?:  =(~ i)  s+'/'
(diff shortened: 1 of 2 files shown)


liam-fitzgerald (approved):
LGTM!

tacryt-socryp:
Confirmed on ~tacryt-socryp on livenet that this does not cause any problems and that the returned strings are the exact same as they were before.

matildepark:
Does this have to hit network or can we just put it in the eventual release?

tacryt-socryp:
Shouldn't matter which it targets, but it will make Landscape load times snappier

PR #4796, opened 2021-04-20
u3: propagate bail:meme

joemfb:
This PR updates the u3 exception mechanism (bail) to no longer treat `bail:meme` (OOM on the loom) as a fatal error.

I've been hesitant to make this change, as I'm not confident that I completely understand all scenarios. So I've additionally made all top-level bails fatal. This was already explicitly the case for `bail:exit` (deterministic nock reduction failure), and appeared to be implicitly the case for most other errors. There is still code that tries to recover from some top-level bails (see `recover: top`), but I don't see how it could maintain our invariants with regards to memory management. I'm leaving it for now until this implementation can be more thoroughly cleaned up.

Additionally, I've cherry-picked a cleanup commit from #4675, and removed assertions from the emergency printf function. Those were added to suppress compiler warnings, but they're counterproductive.

### DIFF
--- a/pkg/urbit/noun/manage.c
+++ b/pkg/urbit/noun/manage.c
@@ -98,28 +98,22 @@ _cm_punt(u3_noun tax)
 }
 #endif
 
-static void _write(int  fd,  const  void  *buf,  size_t count)
-{
-  if (count != write(fd,  buf, count)){
-    u3l_log("write failed\r\n");
-    c3_assert(0);
-  }
-}
-
-
 /* _cm_emergency(): write emergency text to stderr, never failing.
 */
 static void
 _cm_emergency(c3_c* cap_c, c3_l sig_l)
 {
-  _write(2, "\r\n", 2);
-  _write(2, cap_c, strlen(cap_c));
+  c3_i ret_i;
+
+  ret_i = write(2, "\r\n", 2);
+  ret_i = write(2, cap_c, strlen(cap_c));
 
   if ( sig_l ) {
-    _write(2, ": ", 2);
-    _write(2, &sig_l, 4);
+    ret_i = write(2, ": ", 2);
+    ret_i = write(2, &sig_l, 4);
   }
-  _write(2, "\r\n", 2);
+
+  ret_i = write(2, "\r\n", 2);
 }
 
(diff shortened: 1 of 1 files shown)


pkg/urbit/noun/manage.c
```diff
@@ -84,50 +84,30 @@ static sigjmp_buf u3_Signal;
 static uint8_t Sigstk[SIGSTKSZ];
 #endif
 
-#if 0
-/* _cm_punt(): crudely print trace.
```
philipcmonk, line 88: Do we have something like this for use in gdb?

philipcmonk (approved):
A little worried about other top-level errors, but I suppose there can't be many of them.

pkg/urbit/noun/manage.c
joemfb: `u3m_p()`/`u3m_pretty()` usually work in gdb. I've never used this, but I'd be happy to leave it if it looks useful to you.

joemfb:
Wrt "top-level" exceptions, we probably need more precise terminology. What I've disabled here are "explicitly initiated home-road bails", ie, any call to `u3m_bail()` on the home road. Before disabling them, i tested with `bail:fail` on the home road; there was a segfault (presumably due to a `_longjmp()` with a stale buffer), which was caught and turned into `bail:oops`, which aborted the process. Also the convention throughout u3 has been to not worry about memory management at `u3m_bail()` call sites, since the entire road will be discarded.

Separately, there are "external exceptions" that terminate inner roads and produce a trace at the call site. Of these, I've tested ctrl-c (SIGINT), event timeouts (SIGALRM), and stack overflows; all are still working as expected

pkg/urbit/noun/manage.c
philipcmonk: If this works we should just uncomment it.  I often want to attach a debugger and see what it's working on in a hoon stack trace, but haven't been able to.

philipcmonk:
That makes sense

pkg/urbit/noun/manage.c
joemfb: I put it back, ifdef'd out, as a reminder to get trace printing working.

PR #4801, opened 2021-04-21
chat: jimmy's visualfix omnibus

jamesacklin:
Some very fun visual updates to chat, courtesy of @jyng.

- Introduces click-to-copy a ship's name in the mini-profile overlay.
- Swaps to the new Messages icon in the status bar and in Leap.
- Switches the cursor to `pointer` for Attachments and Dojo in the Chat input field.
- Swaps the message loading placeholder background to `washedGray` and rounds the box corners by 4px.
- Presents external images in chat with no gray border or padding, and introduces a 4px border-radius.
- When hovering over an external image, surfaces a small visual cue in the upper-right of the image.
- When images are in a normal chat flow context, makes the entire image, but when an image is transcluded, limits interactivity to this small visual cue. (This is in anticipation of clicking the entire transuded block to jump to its context, but allowing some external actions.)

Non-transcluded image appearance, hovered:
image

Transcluded image appearance, hovered:
image

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
@@ -143,6 +143,7 @@ class ChatInput extends Component<ChatInputProps, ChatInputState> {
     const avatar =
       props.ourContact && props.ourContact?.avatar && !props.hideAvatars ? (
         <BaseImage
+          flexShrink={0}
           src={props.ourContact.avatar}
           height={24}
           width={24}
@@ -183,7 +184,7 @@ class ChatInput extends Component<ChatInputProps, ChatInputState> {
         className='cf'
         zIndex={0}
       >
-        <Row p='12px 4px 12px 12px' alignItems='center'>
+        <Row p='12px 4px 12px 12px' flexShrink={0} alignItems='center'>
           {avatar}
         </Row>
         <ChatEditor
--- a/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatInput.tsx
@@ -189,6 +189,7 @@ class ChatInput extends Component<ChatInputProps, ChatInputState> {
             ) : (
               <Icon
                 icon='Attachment'
+                cursor='pointer'
                 width='16'
                 height='16'
                 onClick={() =>
@@ -201,6 +202,7 @@ class ChatInput extends Component<ChatInputProps, ChatInputState> {
         <Box mr={2} flexShrink={0} height='16px' width='16px' flexBasis='16px'>
           <Icon
             icon='Dojo'
+            cursor='pointer'
             onClick={this.toggleCode}
             color={state.inCodeMode ? 'blue' : 'black'}
           />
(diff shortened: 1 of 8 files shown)


matildepark:
@jyng @g-a-v-i-n note the target branch here is itself blocked on https://github.com/urbit/indigo-react/issues/101

urcades (approved):
Adding small clarification to the last bullet for anyone reading:

> When images are in a normal chat flow context, makes the entire image **clickable**, but when an image is transcluded, limits interactivity to this small visual cue. (This is in anticipation of clicking the entire transuded block to jump to its context, but allowing some external actions.)

This is feeling really good, but did catch the messed up spinner @matildepark mentioned, which is blocking this PR.

urcades:
Actually, I take it back there's ONE MORE THING related to this item:

> Switches the cursor to pointer for Attachments and Dojo in the Chat input field.

Since this is already being tweaked, I'd like to humbly request that we perform a very simple restyling of how these icons are placed within the input field. Right now they are a bit cramped:

Screen Shot 2021-04-21 at 6 27 11 PM

And the placement of the "Execute Hoon" button is on the right edge, which confuses people often (because it vaguely looks like a "send" button).

Rather, I think we should move the "Attach" to the rightmost slot, and pad out the items as indicated in the following video:

https://user-images.githubusercontent.com/1195363/115629365-36ba2c00-a2d0-11eb-9318-8e8cbdd1550d.mp4

If we could make these small edits, I think chat would feel increasingly put together.

matildepark:
See also https://github.com/urbit/landscape/issues/777

matildepark:
@urcades We could probably slip in a send button on mobile devices, if you want

urcades:
This would be excellent — we don't need to overthink this! It could be as simple as a text button added to the right side of the icons, given the same approximate amount of tap space.

Screen Shot 2021-04-22 at 5 38 24 PM

On the other hand, if y'all are OK with pushing this a bit, into a much better space, I have something of a more involved spec here:

Screen Shot 2021-04-22 at 5 39 23 PM

g-a-v-i-n:
Should be resolved now

matildepark:
Unblocked, thanks @g-a-v-i-n. I’ll review but @urcades has some feedback to address first I think.

jamesacklin:
@urcades I was happy to push the send button in the ChatInput a bit, for your review. Forgive the Chrome device simulator here, but it only appears on mobile as desired, 

No input, not focused:
image

No input, field focused (cursor blinks here, you'll have to trust me):
image

Some input, field focused:
image

Some input, field blur:
image

matildepark (changes requested):
We need to prevent the avatar shrinking.

image


Kinda loud — 
image

The arrow pops in and out in Chrome dev tools for me, presumably on hot reload, how is it being prompted?

jamesacklin:
> We need to prevent the avatar shrinking.
Done.

> Kinda loud —
"I found it like that, I swear." Not sure how to fix this, since we save the input to local storage if you navigate away and come back to that channel, so the unmount function is kind of critical here. @liam-fitzgerald know how this works?

> The arrow pops in and out in Chrome dev tools for me, presumably on hot reload, how is it being prompted?
Currently shown and hidden in tandem with the `chat-editor` focus; there's no real good way to detect if the soft keyboard is visible or not, so I figured :focus was a good-enough proxy.

matildepark (approved):
Fair enough, out of scope.

tacryt-socryp:
Feeling like my text input is really squeezed by these icons on iOS now

matildepark:
https://github.com/urbit/landscape/issues/240 is probably exacerbated, it's likely calcing `100% - 25px` or whatever where 100% isn't the full viewport width.

PR #4802, opened 2021-04-21
landscape/logic: type safety adjustments

matildepark:
I want to work on doing some interface CI, but first that means I need to actually make our TypeScript check not have 300+ errors. I started by adjusting the `pkg/interface/src/logic` folder, here, to fix all the errors when type checking by making them safer or otherwise reflect what we're actually doing.

I still have to figure out the Zustand types failing for some strange reason. I would rather just not use them at this rate.

I also upgrade TypeScript to 4.2 in this PR. I don't think the breaking changes in 4.0 affect us.

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -9495,9 +9495,9 @@
       "dev": true
     },
     "typescript": {
-      "version": "3.9.7",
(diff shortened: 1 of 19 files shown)


liam-fitzgerald (approved):
LGTM! How many TS errors do we have now?

matildepark:
Idk, 305. A lot to do.

PR #4803, opened 2021-04-22
GraphNotification: safe property access

liam-fitzgerald:
Fixes crash in notifications screen. Don't have repro, @matildepark can you confirm fix?

### DIFF
--- a/pkg/interface/src/views/apps/notifications/graph.tsx
+++ b/pkg/interface/src/views/apps/notifications/graph.tsx
@@ -145,8 +145,7 @@ function ContentSummary({ icon, name, author, to }) {
 export const GraphNodeContent = ({ post, mod, index, hidden, association }) => {
   const { contents } = post;
   const idx = index.slice(1).split("/");
-  const { group, resource } = association;
-  const url = getNodeUrl(mod, hidden, group, resource, index);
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/apps/notifications/graph.tsx
```diff
@@ -296,7 +295,7 @@ export function GraphNotification(props: {
     dm,
     singleAuthor
   );
-  const groupAssociation = useAssocForGroup(association.group);
+  const groupAssociation = useAssocForGroup(association?.group);
```
matildepark, line 298: I don’t have repro either. But do these functions handle undefined well?

liam-fitzgerald: yes

matildepark: Well as per https://github.com/urbit/landscape/issues/799 it does appear to be this line in question. I just wonder when we are passing undefined as an association.

liam-fitzgerald: I'm not sure, perhaps notifications aren't being cleaned up correctly on graph remove?

matildepark (approved):
Well since neither of us can repro and these are harmless nice-to-dos anyway, it seems fine to try this?

PR #4804, opened 2021-04-22
u3: rewrites %ud parser

### LINKED ISSUE #4767: (slav %ud @t) is too lenient when parsing
Allows the first chunk to have 4 digits in a row without a separator. 4 digits in any other chunk is rejected. 

Allows redundant separators.

```
:: 4 in first chunk is okay
(slav %ud '1234')   -> 1.234   
(slav %ud '12345')  -> crash
(slav %ud '1234.5') -> 12.345
(slav %ud '1.234.5678')   -> crash

:: redundant separators are okay
(slav %ud '1234.567.8')   -> 12.345.678
(slav %ud '1234.56..78.') -> 12.345.678
```

joemfb:
Fixes #4767, and improves performance.

There was a jet mismatch in `+slaw`; the `@ud` parser mistakenly allowed leading zeros, and less-than-3-digit blocks between `.` separators. Those bugs have been corrected in a new implementation.

To test the performance, I parsed a big `@ud` in a loop.

```
> =do |*([max=@ud arg=* fun=gate] =|(i=@ud |-(?:(=(max i) ~ =+((fun arg) $(i +(i)))))))
> |start %time
```

The raw parser combinator:

```
> :time (do 1.000 (rap 3 ['123.456.789' (reap 200 '.123.456.789')]) |=(a=@t (rush a dem:ag)))
[%took ~s21..438a]
```

The old, overly-lenient jet:

```
> :time (do 1.000 (rap 3 ['123.456.789' (reap 200 '.123.456.789')]) |=(a=@t (slaw %ud a)))
[%took ~s5..697e]
```

The new jet:

```
> :time (do 1.000 (rap 3 ['123.456.789' (reap 200 '.123.456.789')]) |=(a=@t (slaw %ud a)))
[%took ~s0..226a]
> :time (do 10.000 (rap 3 ['123.456.789' (reap 200 '.123.456.789')]) |=(a=@t (slaw %ud a)))
[%took ~s0..769a]
> :time (do 100.000 (rap 3 ['123.456.789' (reap 200 '.123.456.789')]) |=(a=@t (slaw %ud a)))
[%took ~s3..dd65]
```

### DIFF
--- a/pkg/urbit/noun/imprison.c
+++ b/pkg/urbit/noun/imprison.c
@@ -313,12 +313,17 @@ u3i_word(c3_w dat_w)
 u3_atom
 u3i_chub(c3_d dat_d)
 {
-  c3_w dat_w[2] = {
-    dat_d & 0xffffffffULL,
-    dat_d >> 32
-  };
+  if ( c3y == u3a_is_cat(dat_d) ) {
+    return (u3_atom)dat_d;
+  }
+  else {
+    c3_w dat_w[2] = {
+      dat_d & 0xffffffffULL,
+      dat_d >> 32
+    };
 
-  return u3i_words(2, dat_w);
+    return u3i_words(2, dat_w);
+  }
 }
 
 /* u3i_bytes(): Copy [a] bytes from [b] to an LSB first atom.
--- a/pkg/urbit/noun/serial.c
+++ b/pkg/urbit/noun/serial.c
@@ -856,3 +856,111 @@ u3s_cue_atom(u3_atom a)
 
   return u3s_cue_bytes((c3_d)len_w, byt_y);
 }
+
+#define NONZERO(a) ( ((a) >= '1') && ((a) <= '9') )
+#define DIGIT(a)   ( ((a) >= '0') && ((a) <= '9') )
+
+/* u3s_sift_ud_bytes: parse @ud
+*/
+u3_weak
+u3s_sift_ud_bytes(c3_w len_w, c3_y* byt_y)
+{
+  //  +ape:ag: just 0 or
+  //
+  if ( (1 == len_w) && ('0' == *byt_y) ) {
+    return (u3_noun)0;
+  }
+
+  //  +ted:ab: leading nonzero, 0-2 digits
+  //
+  if ( len_w && NONZERO(*byt_y) ) {
+    c3_s val_s = *byt_y++ - '0';
+
+    //  0 digits
+    //
+    if ( 0 == --len_w ) return (u3_noun)val_s;
+
+    //  1 digit
+    //
+    if ( !DIGIT(*byt_y) ) goto fail;
+
+    val_s *= 10;
+    val_s += *byt_y++ - '0';
+
+    if ( 0 == --len_w ) return (u3_noun)val_s;
+
+    //  2  digits
+    //
+    if ( !DIGIT(*byt_y) ) goto fail;
+
+    val_s *= 10;
+    val_s += *byt_y++ - '0';
+
+    if ( 0 == --len_w ) return (u3_noun)val_s;
+
+    //  +tid:ab: dot-prefixed 3-digit blocks
+    //
+    if ( 0 != (len_w % 4) ) goto fail;
+
+    {
+      //  XX estimate size, allocate once
+      //
+      mpz_t a_mp;
+      mpz_init_set_ui(a_mp, val_s);
+
+      while ( len_w ) {
+        if (  ('.' != byt_y[0])
+           || !DIGIT(byt_y[1])
+           || !DIGIT(byt_y[2])
+           || !DIGIT(byt_y[1]) )
+        {
+          mpz_clear(a_mp);
+          goto fail;
+        }
+
+        byt_y++;
+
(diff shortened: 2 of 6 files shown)


tacryt-socryp:
God bless

ohAitch:
Sanity check, how often are we parsing indirect `@ud`? The comparison for like `'12.345'` will probably be more relevant(and further favor the C parser), and unless the answer is "often" it seems wise to rip out all bigint handling and defer to the hoon implementation for input strings length >11(making the C logic even faster)

ohAitch:
("or input strengths of length 12 that start with '1' or '2.0'" if you want to cover almost all 0..2^31, but if we're not seeing @ud around 0x7fff.ffff we're probably not seeing anything above 0x1fff.ffff either)

pkg/urbit/noun/serial.c
```diff
@@ -856,3 +856,111 @@ u3s_cue_atom(u3_atom a)
...
+
+        len_w -= 4;
+      }
+
+      return u3i_mp(a_mp);
+    }
+  }
+
+fail:
+  return u3_none;
```
ohAitch, line 940: While the reference to the Apple SSL parsing bug is clever, given this is just a return I would favor inlining the gotos unless that would have significant performance implications.

joemfb: Good point.

joemfb:
I don't think there's much to be gained in clarity or simplicity with special-casing as you suggest. But avoiding indirect atom allocation where possible is a win.

hoon:

```
> :time (do 10.000 '123.456.789' |=(a=@t (rush a dem:ag)))
>=
[%took ~s1..5402]
```

old jet:

```
> :time (do 10.000 '123.456.789' |=(a=@t (slaw %ud a)))
[%took ~s0..15c9]
> :time (do 1.000.000 '123.456.789' |=(a=@t (slaw %ud a)))
[%took ~s1..4747]
```

new jet:

```
> :time (do 1.000.000 '123.456.789' |=(a=@t (slaw %ud a)))
[%took ~s0..ecf0]
> :time (do 10.000.000 '123.456.789' |=(a=@t (slaw %ud a)))
[%took ~s8..47b8]
```

(Yes, I had stacked the deck by comparing really big atoms ...)

Fang- (approved):
Very nice!

pkg/urbit/noun/imprison.c
```diff
@@ -313,12 +313,17 @@ u3i_word(c3_w dat_w)
 u3_atom
 u3i_chub(c3_d dat_d)
 {
-  c3_w dat_w[2] = {
-    dat_d & 0xffffffffULL,
-    dat_d >> 32
-  };
+  if ( c3y == u3a_is_cat(dat_d) ) {
```
ohAitch, line 316: Nice. This works on doubles bc it's testing *any* set bit above 31 right?

pkg/urbit/tests/jet_tests.c
```diff
@@ -9,6 +9,108 @@ _setup(void)
...
+      fprintf(stderr, "sift_ud: (bex 32) wrong\r\n");
+      ret_i = 0;
+    }
+
+    u3z(out); u3z(pro);
+  }
+
+
+  {
+    c3_c* num_c = "340.282.366.920.938.463.463.374.607.431.768.211.456";
```
ohAitch, line 93: And tests including for the indirect handling, A+

pkg/urbit/noun/serial.c
```diff
@@ -856,3 +856,124 @@ u3s_cue_atom(u3_atom a)
...
+      val_d += *byt_y++ - '0';
+
+      len_w -= 4;
+    }
+
+    return u3i_chub(val_d);
+  }
+
+  {
+    //  XX estimate size, allocate once
```
ohAitch, line 928: Every 4 bytes of input is a couple percent under 10 bits of output right?

ohAitch (approved):
We're down iirc 1/3 of the lines, the gotos have been turned into a normal if/else, the while len *p++ business is suspicious but I can follow it. Given we're not getting tapes below 24 bytes per character any time soon(which I expect is what hamstrings the hoon on large input), I grudgingly formally concede that this being in is better than not 👍

pkg/urbit/noun/serial.c
ohAitch: (More inaccurately, as noted above 24 characters is a double / 12 characters is a word, being 10% over isn't a huge deal given u3i_mp will reallocate to exact size anyway)

pkg/urbit/noun/imprison.c
joemfb: That's right:

```
#     define u3a_is_cat(som)    (((som) >> 31) ? c3n : c3y)
```

pkg/urbit/noun/serial.c
joemfb: Yeah, this worth doing (was previously `~s3`+):

```
> :time (do 100.000 (rap 3 ['123.456.789' (reap 200 '.123.456.789')]) |=(a=@t (slaw %ud a)))
[%took ~s2..dde3]
```

PR #4805, opened 2021-04-22
group-timer: fix mark assertion

liam-fitzgerald:
Fixes an infinite loop of kick and resubscribes. Already deployed to ~difmex-passed


tacryt-socryp (approved):
Gee whiz

Fang-:
fwiw I already committed ec2b9be9e to the `ops-tlon` branch. Guess I should've deleted all the branches that got merged into that? It's the CI target, at least.

PR #4809, opened 2021-04-22
publish: forestry

liam-fitzgerald:
Instead of cutting up lines to get markdown and graph contents to flow correctly, we render graph contents as a list of markdown trees and then mangle them horribly to form one tree. We then walk the tree, rendering as per usual. Enables mentions, references and remote content for publish. 
Screen Shot 2021-04-22 at 7 02 38 pm


tacryt-socryp:
This looks really good as an approach. It's correct and not too weighty.

matildepark:
I'm curious why this is still a draft

liam-fitzgerald:
This is in theory ready to go. I have minor reservations about pushing this to next-js, as it will cause publish notes by users of next-js to be unreadable by mainnet

matildepark:
My main reservation is that if you start a note with a reference or a URL it will probably crash.

liam-fitzgerald:
Proposal: have publish disable the new mentions and other content types to maintain backwards compatibility, port GraphContentWide to the new technique and then ship it to next-js to validate correctness. Then we can turn the new publish content on when we ship next to master

matildepark:
I mean, go for it, let's just get this out already

matildepark (changes requested):
This retires the auto-group linking currently in place interface-wide.

image

Until we have a complete user story around getting group links in permalink format, this is just not advisable. Even parsing "~bitbet-bolbel/urbit-community" and transforming it into a reference during tokenisation is better than mandating users construct, eg., `web+urbitgraph://group/~bitbet-bolbel/urbit-community` by hand, to link to a group. cc @urcades 

Second, this doubles the tokenised message in Publish if I uncomment the lines:

image

matildepark (approved):
OK, this works. Also fixes the issue where chat only submits group links on their own lines.

PR #4820, opened 2021-04-25
Clay virtual desks

taalhavras:
This PR adds support for %fuse requests to clay. %fuse combines many desks, replacing the contents of the target with the merge of all sources (specified by the germs in the request). 

tests:
running |fuse and |pass on ships booted from pill
fusing remote desks
upgrading an old ship while it's running (and testing that subsequent clay changes also work)
running |fuse and |pass on upgraded ships
verifying that previous versions of remote desks are correctly pulled in
verify previous contents of a desk are overwritten by fuse

cc @philipcmonk

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -60,7 +60,7 @@
 +$  cult  (jug wove duct)
 ::  State for ongoing %fuse merges - the list maintains the ordering
 ::  and the map stores the data needed to merge
-+$  melt  [(list [beak germ]) (map [beak germ] (unit dome:clay))]
++$  melt  [bas=beak con=(list [beak germ]) sto=(map beak (unit dome:clay))]
 ::
 ::  Domestic desk state.
 ::
@@ -1049,7 +1049,7 @@
           ~
         =/  rus  rus:(~(gut by hoy.ruf) her *rung)
         %+  ~(gut by rus)  syd
-        [lim=~2000.1.1 ref=`*rind qyx=~ dom=*dome per=~ pew=~ fiz=[~ ~]]
+        [lim=~2000.1.1 ref=`*rind qyx=~ dom=*dome per=~ pew=~ fiz=*melt]
       ::  administrative duct, domestic +rede
       ::
       :+  ~  `hun.rom.ruf
@@ -1969,39 +1969,54 @@
     (emit hen %pass wire %c %warp ali-ship ali-desk `[%sing %v case /])
   ::
   ++  make-melt
-    |=  srcs=(list [beak germ])
+    |=  [bas=beak con=(list [beak germ])]
     ^-  melt
-    :-  srcs
-    %-  ~(gas by *(map [beak germ] (unit dome:clay)))
-    (turn srcs |=(a=[beak germ] [a *(unit dome:clay)]))
+    :+  bas  con
+    %-  ~(gas by (~(put by *(map beak (unit dome:clay))) bas *(unit dome:clay)))
+    (turn con |=(a=[beak germ] [-.a *(unit dome:clay)]))
   ::
   ++  start-fuse
-    |=  [srcs=(list [beak germ])]
+    |=  [ful=(list beak) gim=(list germ)]
     ^+  ..start-fuse
-    ::  use emil here to queue a list of moves
-    ::  syd is current desk (target desk)
-    ::  what is hen? it's the current duct... can this be reused?
+    ::  must have n sources and n-1 gems
+    ?>  =((lent ful) +((lent gim)))
     =/  moves=(list move)
         %+  turn
-          srcs
-        |=  [bec=beak g=germ]
+          ful
+        |=  bec=beak
         ^-  move
-        =/  wir=wire  /fuse/[syd]/(scot %p p.bec)/[q.bec]/(scot r.bec)/[g]
+        =/  wir=wire  /fuse/[syd]/(scot %p p.bec)/[q.bec]/(scot r.bec)
         [hen %pass wir %c %warp p.bec q.bec `[%sing %v r.bec /]]
+    ?~  ful
+      !!
+    =/  base=beak  i.ful
+    =/  ful=(list beak)  t.ful
+    =/  con=(list [beak germ])
+        =|  acc=(list [beak germ])
+        |-
+        ?~  ful
+          acc
+        ?~  gim
+          !!
+        $(acc [[i.ful i.gim] acc], ful t.ful, gim t.gim)
     ::  we also want to clear the state (fiz) associated with this
     ::  merge and print a warning if it's non trivial i.e. we're
     ::  starting a new fuse before the previous one terminated.
-    ~&  ?~  -.fiz
-          [%starting-fuse srcs]
-        :^  %starting-fuse  srcs  %discarding-state
-        ::  we don't want to ~& an entire dome
-        %-  ~(run by +.fiz)
-        |=(v=(unit dome:clay) ?~(v %not-received %recieved))
-    =.  fiz  (make-melt srcs)
+    ~&  ?~  con.fiz
+          [%starting-fuse base con]
+        :*  %starting-fuse
+            base
+            con
+            %discarding-state
+            ::  we don't want to ~& an entire dome
+            %-  ~(run by sto.fiz)
+            |=(v=(unit dome:clay) ?~(v %not-received %recieved))
+        ==
+    =.  fiz  (make-melt base con)
     (emil moves)
   ::
   ++  take-fuse
-    |=  [[bec=beak g=germ] =riot]
+    |=  [bec=beak =riot]
     ^+  ..take-fuse
     ?~  riot
       ~&  [%fuse-for syd %missing bec]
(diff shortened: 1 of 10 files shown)


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1962,32 +1968,163 @@
...
+    =/  moves=(list move)
+        %+  turn
+          [[bas *germ] con]
+        |=  [bec=beak germ]
+        ^-  move
+        =/  wir=wire  /fuse/[syd]/(scot %p p.bec)/[q.bec]/(scot r.bec)
+        [hen %pass wir %c %warp p.bec q.bec `[%sing %v r.bec /]]
+    ::  we also want to clear the state (fiz) associated with this
+    ::  merge and print a warning if it's non trivial i.e. we're
+    ::  starting a new fuse before the previous one terminated.
```
philipcmonk, line 1990: Kernel comment style is to use "breathing" comments, which means they have a "blank" line after each comment, and optionally one before.  This should be fixed throughout.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1962,32 +1968,163 @@
...
+      ::  merge failed
+      (mean %fuse-merge-failed p.result)
+    ::
+        %&
+      ::  now we have a unit merge-result
+      =/  merge-result=(unit merge-result)  +.result
+      ?~  merge-result
+        ::  this merge was a no-op, just continue
+        $(merges t.merges)
+      ?~  conflicts.u.merge-result
```
philipcmonk, line 2053: I'd make this a `?^` so that the two-line exceptional case can be handled first, reducing how much stack you have to keep in your head while reading.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1962,32 +1968,163 @@
...
+    =/  parents=(list tako)  ~[(~(got by hit.initial-dome) let.initial-dome)]
+    =/  merges  con.fiz
+    |-
+    ^+  ..take-fuse
+    ?~  merges
+      ::  reset fiz before potentially parking - we hope to
+      ::  succeed so we don't need this state anymore.
+      =.  fiz  *melt
+      =.  hut.ran  clean-hut-ran
+      (park | [%| continuation-yaki(p (flop parents))] rag)
```
philipcmonk, line 2038: Since we give `%mere` in the failure case, we should give it here as well.  I think it should come *after* the `park`, call (so the `%mere` arrives after any moves caused by the commit), so something like

```
=.  ..take-fuse  (park ...)
(done ...)
```

philipcmonk: Although I note that `|merge` sends the `%mere` before the `park`, so maybe that's fine too.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1962,32 +1968,163 @@
...
+        %=  $
+          continuation-yaki  merged-yaki
+          merges  t.merges
+          hut.ran  (~(put by hut.ran) r.merged-yaki merged-yaki)
+          lat.rag  (~(uni by lat.rag) lat.u.merge-result)
+          parents  [(~(got by hit.ali-dom) let.ali-dom) parents]
+        ==
+      :: if there are merge conflicts send the error and abort the merge
+      =.  fiz  *melt
+      (done %& conflicts.u.merge-result)
```
philipcmonk, line 2082: Need to restore `hut.ran` at least so we don't leave garbage around.  Though, I don't like the idea of having to specifically restore changed state, since it would be easy to later change more state and not realize it needs to be restored.

Maybe next to `clean-hut-ran` we should pin a clean `..take-fuse` as well and just restore that.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1962,32 +1968,163 @@
...
+      =.  fiz  *melt
+      =.  hut.ran  clean-hut-ran
+      (park | [%| continuation-yaki(p (flop parents))] rag)
+    =/  [bec=beak g=germ]  i.merges
+    =/  ali-dom=dome:clay  (need (~(got by sto.fiz) bec))
+    =/  result  (merge-helper p.bec q.bec g ali-dom `continuation-yaki)
+    ?-  -.result
+        %|
+      ::  merge failed
+      (mean %fuse-merge-failed p.result)
```
philipcmonk, line 2045: Since we're using `%mere` in the other case to give in-band errors (which I think is nicer ergonomically), we should probably do that here as well.  Which also means restoring `hut.ran`.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3384,19 +3516,37 @@
     ::
     ::  Creates a nako of all the changes between a and b.
     ::
+    ::
```
philipcmonk, line 3519: Errant blank line

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3429,15 +3579,9 @@
       |=  [plops=? a=(unit tako) b=tako]
```
philipcmonk, line 3579: This whole function can be removed now.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4320,6 +4525,7 @@
       dos.rom
     %-  ~(run by dos.rom.ruf)
     |=  =dojo
+    ::dojo
```
philipcmonk, line 4528: can delete now

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4329,6 +4535,7 @@
         rus
       %-  ~(run by rus.rung)
       |=  =rede
+      ::rede
```
philipcmonk, line 4538: delete

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4202,11 +4354,43 @@
...
+              mon=(map term beam)
+              hez=(unit duct)
+              cez=(map @ta crew)
+              pud=(unit [=desk =yoki])
+          ==
+      +$  room-7
+          $:  hun=duct
+              dos=(map desk dojo-7)
+          ==
+
```
philipcmonk, line 4375: delete blank line

philipcmonk (changes requested):
Looks pretty good.  Mostly style nits, plus some error-handling edge cases.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1962,32 +1968,163 @@
...
+      ..take-fuse
+    ?>  (~(has by sto.fiz) bec)
+    =.  fiz
+        :+  bas.fiz  con.fiz
+        (~(put by sto.fiz) bec `!<(dome:clay q.r.u.riot))
+    =/  all-done=flag
+        %-  ~(all by sto.fiz)
+        |=  res=(unit dome:clay)
+        ^-  flag
+        !=(res ~)
```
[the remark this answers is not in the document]
taalhavras: yeah I'll keep my bastardized style to ucal exclusively :P 

fwiw the runes that "introduce new variables" just make more sense to me if the name and the value have different indentation levels because I can more easily see what the variable is instead of trying to read the entire sub expression. not trying to advocate for this to be generally used but would be interested to hear if other people have the same opinion. perhaps a question for the forge..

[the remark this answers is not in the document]
ohAitch: I'd say consistency is the most important thing for readability, no matter the rune.

pkg/arvo/sys/vane/clay.hoon
taalhavras: In all cases I think we want to restore `hut.ran` and set `fiz` to `*melt`. I guess we could pin `..take-fuse` and then change `fiz` but if we're doing that I think we might as well be explicit.

IMO since we don't know what state is changing/how it should be restored I prefer explicitly restoring those parts. that should be restored and changing those parts that should be changed.

As an aside - I don't think there's a simple way to define a "cleanup handler" (closest analog I can think of) here to execute these state changes but if there is that would also be a way to reduce errors of this kind.

ohAitch: I'm definitely fond of "copy-on-write root as contiguous a chunk of per-fallible-operation state as possible", not sure how feasible this is for merging maps or w/e.

pkg/arvo/sys/vane/clay.hoon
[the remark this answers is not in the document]
philipcmonk: Perhaps should be more clear.  This is the correct style:

```
=/  new-name=type
  ?~  some
    long
  value
continuation
```

Some of the changes you made dedented too far, to

```
=/  new-name=type
?~  some
  long
value
continuation
```

Which I agree is hard to read because it mixes up the value with the remaining code.  In the principle above, I count the remaining code as the (first-to-) last child and the value of the variable as the 2nd-to-last child, which is why it's indented one level, but not two.

pkg/arvo/sys/vane/clay.hoon
taalhavras: I think I've caught all of these.

pkg/arvo/sys/vane/clay.hoon
taalhavras: ok took a shot at this. Given how I implemented this change I had to call park after calling done, which does line up with merge but is different from @philipcmonk's earlier suggestion.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -2031,33 +2033,24 @@
...
-    ::
     |-
     ^+  ..take-fuse
     ?~  merges
       =/  t=tang  [leaf+"{<syd>} fused from {<bas.fiz>} {<con.fiz>}" ~]
-      =.  fiz  *melt
-      =.  hut.ran  clean-hut-ran
-      =.  ..take-fuse  (park | [%| continuation-yaki(p (flop parents))] rag)
-      (done %| %fuse-succeded t)
+      =.  ..take-fuse  (done-fuse clean-state %| %fuse-succeded t)
```
philipcmonk, line 2046: This should be `%& ~`.  The term/tang case is for failure traces

PR #4821, opened 2021-04-25
landscape: highlight content mentioning our ship

matildepark:
Was using Landscape and was just thinking "we should probably have this"...

cc @urcades — how's this look (it's just using our already-defined highlight styles) and do you approve of the feature?
comparison

@liam-fitzgerald @tacryt-socryp I don't think this is that expensive, is it?


urcades:
This is a great idea, let's do it

tacryt-socryp (approved):

PR #4822, opened 2021-04-26
stying edits: small collection of various visual fixes

urcades:
All edits were constrained within the `/components` directory of the wallet src. Most involve simple value switches and other similarly simple fixes, towards matching designs.

Will screenshoot and capture other misc. feedback that I can't edit easily.

### DIFF
--- a/pkg/btc-wallet/src/js/components/lib/balance.js
+++ b/pkg/btc-wallet/src/js/components/lib/balance.js
@@ -76,14 +76,14 @@ export default class Balance extends Component {
            height="400px"
            width='100%'
            backgroundColor="white"
-           borderRadius="32px"
+           borderRadius="48px"
            justifyContent="space-between"
            mb={5}
            p={5}
          >
            <Row justifyContent="space-between">
              <Text color="orange" fontSize={1}>Balance</Text>
-             <Text color="lighterGray" fontSize="14px" mono>{addressText}</Text>
+             <Text color="lightGray" fontSize="14px" mono>{addressText}</Text>
              <CurrencyPicker
                denomination={denomination}
                currencies={this.props.state.currencyRates}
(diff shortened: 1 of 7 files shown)


urcades:
Observation: Filtering on valid ship names is too strict, should work without someone needing to type `~` — I know this is a broader issue we run into across urbit interfaces, but requiring typed input for valid patP filtering is unfriendly.

Screen Shot 2021-04-26 at 1 23 51 AM

urcades:
Observation: Overfill on Invoicing interface when a long string is passed as a recipient (BTC addresses) should be properly truncated.

Screen Shot 2021-04-26 at 1 41 49 AM

urcades:
For any truncated address we display, one should be able to click the string to copy the "full" version of the address. This should effectively mimic the pattern we use in Landscape, on patPs.

Screen Shot 2021-04-26 at 2 23 48 AM

urcades:
The current way we display the "BTC" view of one's wallet needs some tweaking:

Screen Shot 2021-04-26 at 2 25 56 AM

Basically, let's remove the 'B' prefix, truncate the trailing decimal numbers to 6 digits, and ensure the count ends in a BTC suffix. The total currency should more or less appear as so:

Screen Shot 2021-04-26 at 2 36 45 AM

urcades:
Observation: Due to the uneven horizontal sizing of the components that make up the "Send" header info, they appear unevenly aligned:

Screen Shot 2021-04-26 at 2 25 42 AM

The simple observation here is that these elements should be treated such that the title "Send BTC" should appear pinned to the top left, the "X" to go back should be pinned to the top right, and the currency should be perfectly centered to the entire column, regardless of the other two elements.

urcades:
The formatting of this invoice/send state is a bit weird:

Screen Shot 2021-04-26 at 1 43 52 AM

An improved design allows the top messaging to remain "whole" and balanced-feeling, and allows one to use the normative "primary button area" to allow a pilot to go back to their balance screen:

Screen Shot 2021-04-26 at 2 47 01 AM

ixv:
Thanks @urcades, I'm going to close this and cherrypick your commit onto another branch, because this isn't really the right branch to PR to, and I'll look into your requested changes.

PR #4823, opened 2021-04-26
VirtualScroller: performance enhancements

liam-fitzgerald:
- Removes unused props from a variety of chat components
- Prevents scroll adjustments from transcluded nodes from being render-blocking (useEffect vs useLayoutEffect)
- Renders less chatmessages at once
- Makes BigIntOrderedMap immutable, to correctly leverage React memoisation
- GraphContentWide now correctly memoizes
- Prevents refs from being reset every render

ChatMessage: remove dead props

ChatResource: remove unnecessary props

ChatWindow: remove unnecessary props

interface: remove dead pendings

virtualContext: drop useLayoutEffect

VirtualScroller: rework for less memory use, faster speeds

interface: make BigIntOrderedMap immutable

interface: remove unnecessary routing props


matildepark (approved):

matildepark:
Appears so, yes.

matildepark:
There is one weird thing, though, I would suggest looking at: if you use this to proxy onto dabben-larbet, and you go to Interface chat in the Tlon group, it uh, crashes the tab.

tacryt-socryp:
Well, let's throw it onto next-js then lmao

tacryt-socryp (approved):
Idk the difference between useLayoutEffect and useEffect but this is good to go.

matildepark:
Wait wait wait don't merge, crashing the tab is sort of a big deal.

matildepark (changes requested):
What's going on with that case ...

pkg/interface/src/views/components/VirtualScroller.tsx
```diff
@@ -575,8 +574,10 @@ function VirtualChild(props: VirtualChildProps) {
 
   const ref = useCallback((el: HTMLElement | null) => {
     setRef(el, props.index);
-  }, [setRef, props.index])
+  //  VirtualChild should always be keyed on the index, so the index should be
+  //  valid for the entire lifecycle of the component, hence no dependencies
+  }, []);
```
tacryt-socryp, line 579: I believe this makes it update on every tick

liam-fitzgerald: No, that's if the dependencies are not provided. `[]` will cause the ref callback to be constant throughout the entire lifecycle.

matildepark:
The unread banner straight up doesn't appear in these "big unread backlog" cases, which is a regression.

matildepark (approved):

PR #4825, opened 2021-04-26
tests: update to reflect recent changes

Fang-:
idk why we don't get CI on next-sys anymore, but it let these slip by. Could've been worse!


matildepark:
I don't think next-sys has https://github.com/urbit/urbit/commit/5e806fef1300c8a22fc5991e32255fa2e2a8366c in it, might be worth merging master back in.

Fang-:
Good call! Just merged master into `release/next-sys`.

philipcmonk (approved):
Great, thanks

PR #4766, opened 2021-04-14
vere: fix cttp crash by accounting for synchronous error callback

joemfb:
The PR fixes a long-standing occasional crash in vere's http client i/o driver. The issue was my misuse of the `h2o_http1client_connect()` -- while usually asynchronous regardless of success, a failure to create a sock results in a synchronous invocation of the error callback, which causes a use-after-free.

Fixes #4601 and #2006

PR #4828, opened 2021-04-27
vere: fix cttp tls crash from (bad) fix for synchronous error callback 

joemfb:
This PR fixes a crash introduced by #4766. That PR fixed a longstanding, rare crash in the http-client driver, caused by a surprising synchronous call to a normally async callback, which would result in a use-after-free. The fix introduced a new, deterministic crash for every TLS request, due to my misunderstanding of the code (which I had written).

`cttp.c` always connects by ip/port, to avoid synchronous `getaddrinfo()` calls in the library (we use libuv's threadpool at call sites). In order to correctly negotiate TLS sessions, we must also synchronously patch the SNI data, but only after successful `connect()` calls. This PR extends the client request state machine to correctly cover these edge cases.

### DIFF
--- a/pkg/urbit/vere/io/cttp.c
+++ b/pkg/urbit/vere/io/cttp.c
@@ -36,26 +36,26 @@
 /* u3_creq: outgoing http request.
 */
   typedef struct _u3_creq {             //  client request
-    c3_l             num_l;             //  request number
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
looks good, and works on my machine

PR #4829, opened 2021-04-27
king: ensure lmdb is statically linked on osx

brendanhay:
By way of ensuring the lmdb package's cabal extra-libraries field is ignored and our lmdb overlay is used. 🤷🏽

### DIFF
--- a/nix/pkgs/hs/default.nix
+++ b/nix/pkgs/hs/default.nix
@@ -82,6 +82,8 @@ haskell-nix.stackProject {
 
       urbit-king.components.tests.urbit-king-tests.testFlags =
         [ "--brass-pill=${brass.lfs}" ];
+
+      lmdb.components.library.libs = lib.mkForce [ lmdb ];
     };
   }];
 }


joemfb (approved):
Looks Obvious To Me

brendanhay:
Waiting for CI to finish then I'll download the artefacts to confirm.

brendanhay:
Note for the future - it looks like the `configureFlags` are setting the wrong [output derivation](https://github.com/NixOS/nixpkgs/blob/d06818e5a4d34acc9235a7f1e32a34d10e6cd4a4/pkgs/development/libraries/lmdb/default.nix#L18), possibly `.bin`, since it doesn't contain the `lib` directory, from the log output:

```
ld: warning: directory not found for option '-L/nix/store/n15sh5piqvl5bgs06bb375b2x7d3imxn-lmdb-0.9.25-bin/lib
```

An alternative to the fix in this PR would be to patch the path in `nix/pkgs/hs/default.nix`:

```
configureFlags = lib.optionals enableStatic [
  "--ghc-option=-optl=-L${lmdb}/lib"
  ...
]
```

pilfer-pandex:
> Looks Obvious To Me

I assume you're being sarcastic?

PR #4830, opened 2021-04-28
Chat: refactor to strip associations prop

liam-fitzgerald:
Paves the way for DMs, by stripping the association prop everywhere and refactoring the Chat functionality into ChatPane, which makes no assumptions about the presence of an association or group by lifting that functionality into prop callbacks.


matildepark:
image

Memoises my contact a little aggressively.

matildepark:
Oh wait, it's because we're passing the inverted boolean we should be.

matildepark (approved):

PR #4831, opened 2021-04-28
interface, npm/api: type checking, round 2

matildepark:
Only npm change is making `value` in a `settings-update` interface optional, since we don't require it for deleting a key.

Mostly correcting imports.


matildepark:
We're at 254 errors now. Quite a handful should be fixed by #4810. Some weird ones I couldn't stomp out is guarding that `_.capitalize(mod)` will output something in `IconRef` in `getModuleIcon`; and I need to revisit type checking content types by keys in GraphContentWide.

liam-fitzgerald (approved):

PR #4833, opened 2021-04-29
file-server: Added support for .ico files:

rabsef-bicrym:
Adding support for .ico files makes hosting prebuilt websites on urbit easier.

This commit utilizes the following changes to effectuate .ico support:
-/mar/ico/hoon
 * Utilizes similar structure to /mar/png/hoon, w/ changed mimetype
-/lib/server
 * Added ico-response, again similar to png w/ same caching
-/app/file-server
 * Added reference to ico-response:gen

This will also rectify one complexity glossed over in [this docs addition](https://github.com/urbit/docs/pull/1067) and allow for minor simplification of the guide.


liam-fitzgerald (approved):
LGTM! Thanks for the contribution

Issue #4307, opened 2021-01-19
Running +ls after a |commit results in dojo becoming unusable until after pressing backspace


hathec-pilsyx:
**Describe the bug**
After creating a text file in Linux  within my mounted %sandbox desk, if I then run "+ls %" to check the results of the commit, I wind up with a situation similar to below, where I get some error messages, followed by "%dy-edit-busy" resulting from any keypress until I hit backspace:

`
> =dir /=sandbox=
=% /~hathec-pilsyx/sandbox/~2021.1.19..01.31.32..d654
[ %clay-sing-indirect-data
  desk=%sandbox
  mood=[care=%a case=[%da p=~2021.1.19..01.31.32..d654] path=/gen/ls/hoon]
  aeon=3
]
> +ls %
[ %clay-sing-indirect-data
  desk=%sandbox
  mood=[care=%a case=[%da p=~2021.1.19..01.31.32..d654] path=/gen/ls/hoon]
  aeon=3
]
%dy-no-prompt
%dy-edit-busy
%dy-edit-busy
%dy-edit-busy
~hathec-pilsyx:dojo/=/sandbox/~2021.1.19..01.31.32..d654>
`

**To Reproduce**
Steps to reproduce the behaviour:
1. Create a textfile in Linux within a mounted desk
2. In dojo, commit the changes to the desk
3. Use +ls to examine desk contents

**Expected behaviour**
The current directory contents are printed to terminal

**System (please supply the following information, if relevant):**
 - OS: [e.g. macOS, linux64, FreeBSD] Ubuntu 20.04.1 x64
 - Vere and Urbit OS versions - 1.0
 - Your ship's `%base` hash (use `+trouble` to check) - [0v1.1gbvo.e03d1.nk94q.lmsv5.lasnk.7s08i.jmgvq.7goos.3uqij.fienr]

**Additional context**
Spoke to some users in ~bitbet-bolbel/urbit-community/help, and was told my issue sounded similar to  #4037

philipcmonk:
I can confirm this issue, it appears to be the `=dir /=sandbox=` line that's the main issue.  It's pinning the version to a particular time (which is why it appears in the prompt), and we apparently can't build files against old versions.  The two fixes we need are:

- Make `=dir /=sandbox=` track the head of the desk instead of pinning to a time
- Allow building old stuff -- you should be able to cd into the past

[...]

PR #4835, opened 2021-04-29
clay: fix requests for strange cares over ames

### LINKED ISSUE #4834: clay: Handle foreign requests for other cares
If you make a request to a foreign desk on a care other than wxyz, the request goes through, but we fail on validating the response (in +validate).  This is incorrect, and also we don't handle some cares that we should be able to handle.  This suggests the following tasks:

- In +duce, refuse to make a foreign request which we won't be able to validate the answer to
- In +start-request, refuse to answer a foreign request which they won't be able to validate
- In +validate, we should be able to trivially validate `%u`, by copying +validate-z and changing `@uvI` to `?`
- In +validate-x, we should build the foreign stdlib if we need to (and cache it), instead of crashing on `(need fer.dom)`.  Or else we should try to build it against our own stdlib/marks.

philipcmonk:
Four changes:

- implement +validate-u to allow %u requests over the network
- make +validate-x use our local marks to make %x requests generally
  work over the network
- in +start-request, if a foreign ship is making a request that we
  shouldn't send over the network, ignore it.  This closes a DOS vector.
- in +duce, if we're about to make a request to a foreign ship which
  they won't be able to answer, crash the event.

Combined, these fix many of the common cases of weirdness around foreign
clay requests.  Notably absent is a fix for reading `%a` across the
network, which I still maintain should happen against the foreign
hoon/zuse.

fixes #4834
see also #4307

Thanks to @tinnus-napbus for providing a repro for these

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -1256,6 +1256,24 @@
     =/  =path  [%question desk (scot %ud index) ~]
     (emit duct %pass wire %a %plea ship %c path `riff-any`[%1 riff])
   ::
+  ++  foreign-capable
+    |=  =rave
+    |^
+    ?-    -.rave
+        %many  &
+        %sing  (good-care care.mood.rave)
+        %next  (good-care care.mood.rave)
+        %mult
+      %-  ~(all in paths.mool.rave)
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
LGTM

PR #4836, opened 2021-04-29
chat: input behavior quirks

jamesacklin:
- Clears state on submit such that the button does not stay blue
- Refocuses field on submit such that the mobile keyboard doesn't disappear
- Removes submit on enter; retains enter for linebreak insertion


urcades (approved):
Yes

liam-fitzgerald (approved):
LGTM!

PR #4837, opened 2021-04-30
interface: fix omnibus

liam-fitzgerald:
VirtualScroller: fix pendings and scroll after jump

hark: fix notification archival

interface: include iPad in UA check


matildepark (approved):
Seems fine but doubtful about iPad. iPad reports itself as a Mac I think.

liam-fitzgerald:
> Seems fine but doubtful about iPad. iPad reports itself as a Mac I think. 

Only on iPad Pro iirc

PR #4838, opened 2021-04-30
ames: add packet filter for rescue operations

philipcmonk:
In some situations it's useful to restrict the Ames packets that arrive
to be from a certain set of ships.  We've needed this a few times in the
last few years, so this creates an ifdef-ed out function for doing this.

It looks like `lan_u` does not get freed in the normal case, but I'm not really sure why not.

### DIFF
--- a/pkg/urbit/vere/io/ames.c
+++ b/pkg/urbit/vere/io/ames.c
@@ -1047,6 +1047,25 @@ _ames_try_forward(u3_ames* sam_u,
   }
 }
 
+#undef AMES_SKIP
+#ifdef AMES_SKIP
+/* _ames_skip(): decide whether to skip this packet, for rescue
+*/
+static c3_o
+_ames_skip(u3_body* bod_u) {
+  if ( bod_u->sen_d[1] == 0 &&
+       ( bod_u->sen_d[0] == 0x743a17a6
+         || bod_u->sen_d[0] == 0xea99acb6
+         || bod_u->sen_d[0] == 0x10100
+     ) ) {
+    return c3n;
+  }
+  else {
+    return c3y;
+  }
+}
+#endif
+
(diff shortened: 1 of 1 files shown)


belisarius222:
The old Ames vane used to have this kind of filter. What's the reason to put this in the runtime instead of in hoon?

—
~rovnys-ricfer
https://urbit.org

joemfb:
`lan_u` is on the stack in `_ames_recv_cb()`. if we need to keep it (via `_ames_put_packet()`), we copy it.

philipcmonk:
> The old Ames vane used to have this kind of filter. What's the reason to put this in the runtime instead of in hoon?

The ship didn't have enough memory to compile Ames.  It's probably cleaner this way anyway.

belisarius222:
Ok, that makes sense.

pkg/urbit/vere/io/ames.c
```diff
@@ -1149,7 +1168,16 @@ _ames_hear(u3_ames* sam_u,
...
+#ifdef AMES_SKIP
+    if (_ames_skip(&bod_u) == c3y ) {
+      u3z(msg);
+    }
+    else {
+#endif
+      _ames_put_packet(sam_u, msg, *lan_u);
+#ifdef AMES_SKIP
+    }
+#endif
```
joemfb, line 1180: I'd be tempted to refactor this block so that we only allocate `msg` if we're going to use it. But it's hard to imagine that actually mattering very much ...

joemfb (approved):
LGTM

pkg/urbit/vere/io/ames.c
philipcmonk: Yeah, arguably it should all move up to right after `bod_u` is defined but with early return you want to make it obvious you're not leaking memory, so I did it at the last possible second.

PR #4840, opened 2021-04-30
graph-store: update log hotfixes

liam-fitzgerald:
Hotfixes the lack of update log for graphs which causes frequent resetting of unreads and notifications.
The varying timestamp in the %add-graph fact from graph-push-hook was
preventing deduplication of the jammed noun in ames. Prevents
unnecessary bail: memes upon sending many %add-graph facts.

### DIFF
--- a/pkg/arvo/lib/graph.hoon
+++ b/pkg/arvo/lib/graph.hoon
@@ -76,6 +76,7 @@
 ++  get-graph
   |=  res=resource
   ^-  update:store
+  =-  -(p *time)
   %+  scry-for  update:store
   /graph/(scot %p entity.res)/[name.res]
 ::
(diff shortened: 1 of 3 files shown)


pkg/arvo/lib/graph.hoon
```diff
@@ -76,6 +76,7 @@
 ++  get-graph
   |=  res=resource
   ^-  update:store
+  =-  -(p *time)
```
Fang-, line 79: So is this timestamp not used for anything? If it is, then we have to be careful about use of it not overlapping with results coming from this function. How would it having an ancient/bogus value affect behavior, if it's used?

tacryt-socryp (approved):
Looks like a safe way to prevent this type of OOM. It's surprising that we didn't see this with any of our previous OTAs, but I guess this is due to increased usage + the simultaneous kick of everyone in the versioning stuff

PR #4841, opened 2021-04-30
meta: rescue flow

liam-fitzgerald:
PRing to document the rescue flow. While the parity is not evenly distributed, it should be enough to get memory under control and doesn't require the host to hex-convert ship names and custom build a binary. A derivative of urbit/urbit#4838 that is designed so that the user doesn't have to control, and also includes fixes to prevent the backlog from getting larger. To fix a ship that's bail: memeing
- Grab the binary for this branch
- Run the bail: memeing ship on this branch and run `|pack` in the dojo
- Commit the arvo changes to the ship
- Allow the ship to clear its queue, you may have to sporadically run `|pack` to avoid additional bail:memes
- Once the result of `|pack` is under 1.6GB, return to the unmodified binary
- `|merge %home ~sponsor %kids, =gem %take-that` to clear the temporary arvo changes

### DIFF
--- a/pkg/arvo/app/graph-push-hook.hoon
+++ b/pkg/arvo/app/graph-push-hook.hoon
@@ -74,6 +74,7 @@
 ++  transform-proxy-update
   |=  vas=vase
   ^-  (unit vase)
+  ?:  =(1 1)  `this
   =/  =update:store  !<(update:store vas)
   =*  rid  resource.q.update
   =.  p.update  now.bowl
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -10,16 +10,17 @@
       [%1 network:zero:store]
       [%2 network:zero:store]
       [%3 network:one:store]
-      state-4
+      [%4 network:store]
+      state-5
   ==
 ::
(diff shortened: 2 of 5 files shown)


philipcmonk:
This is what I was using in the dojo to find which ships are waiting on >1MB messages, so we can filter exactly some subset of these:

```
=/  peers  (murn ~(tap by .^((map ship ?(%alien %known)) %ax /=//=/peers)) |=([=ship al=?(%alien %known)] ?:(=(al %known) `u=ship ~)))
%+  murn  peers
|=  =ship
=-  ?~(- ~ `u=[ship -])
=/  ss  .^(ship-state:ames %ax /=//=/peers/(scot %p ship))
?>  ?=(%known -.ss)
%+  murn  ~(tap by snd.ss)
|=  [bone=@ud mps=message-pump-state:ames]
=-  ?~(- ~ `u=[bone -])
%+  murn  ~(tap in unsent-messages.mps)
|=  blob=@
=/  len  (met 3 blob)
?:  (lth len 1.000.000)  ~
`u=len
```

philipcmonk:
Also the parity check here won't work well because it checks the low bits of the galaxy, and planets are not uniformly distributed across galaxies.

PR #4846, opened 2021-05-01
hark-graph-hook: remove unnecessary scries where possible for perf

tacryt-socryp:
Marginal performance gains that pave the way for larger ones. (Really tiny)

Should do two fewer scries when adding a graph, and one less per %add-nodes update.

Ideally, we'll move to using %watch-as for converting %add-nodes into $notif-kind


liam-fitzgerald (approved):
Righteous

liam-fitzgerald:
(fyi this could probably target next-js)

matildepark:
Hit the button if you dare

PR #4847, opened 2021-05-03
interface: add the ability to delete chat messages

tacryt-socryp:
Careful! Once this is on release/next-js, don't delete messages in prod channels until this is on master, as it will break rendering for non BYB users.

Note that this does not have a renderable state for "deleted messages", it just produces null for them. We can and should render "This message has been deleted." once we have designs @urcades.

Note: this PR now also allows the deletion of group feed messages. Same warning applies about not deleting them on any prod feeds yet.

I went ahead and implemented a "deleted message view" for group feed, as it would be clunky in the interface otherwise.

Screen Shot 2021-05-03 at 3 42 07 PM


urcades:
@tacryt-socryp very rad, will provide a very simple design and drop that into this issue soon

pkg/interface/src/views/apps/chat/ChatResource.tsx
```diff
@@ -134,6 +134,11 @@ function ChatResource(props: ChatResourceProps) {
     api.graph.addPost(ship, name, createPost(window.ship, contents))
   }, [resource]);
 
+  const onDelete = useCallback((msg: Post) => {
+    const { ship, name } = resourceFromPath(resource);
+    api.graph.removePosts(ship, name, [msg.index]);
+  }, [resource]);
```
liam-fitzgerald, line 140: missing `msg.index` as a dependency

liam-fitzgerald (changes requested):
Nit about callback deps, else LGTM

urcades:
Enclosing a design for a very minimal handling of chat deletion:

Message Deletion

In a word: When a message is deleted, it's simply replaced with a generic `black60` replacement text that states `Message deleted by author` or `Message deleted by admin` — if the distinction between admin/author is possible and easy/cheap, we should make it. 

Note that the message is retained and still able to be hovered-over, which should allow one to see the fact that the message was signed and "real" — wondering if we need to afford for signing of a deletion action?

Also note that I've drawn the ability to "Undo" the deletion, if one has the power to do so. I would imagine this wouldn't be a true technical "undo", but a third revision that simply re-exposes the message again.

LMK how this sounds, or if there's anything I'm overlooking or need to account for.

cc @tacryt-socryp

tacryt-socryp:
Well, there is no way to undo deletion, or to know who originally sent the message, or to know who deleted it.

urcades:
Rad — then we can likely roll with the above designs but simply offer up the copy `Message has been deleted` as indicated above, no other bits of functionality that were gestured towards. This should suffice

pkg/interface/src/views/apps/chat/ChatResource.tsx
tacryt-socryp: Why is msg.index a dependency? We pass msg into the callback.

liam-fitzgerald: nvm!

liam-fitzgerald (approved):

matildepark (approved):

PR #4848, opened 2021-05-03
Ames: time out direct lanes faster

belisarius222:
Time out direct lanes much faster after inactivity.  Removes logic in +on-memo that reset the `last-contact` time to `now` after a period of inactivity; this caused the ship to restart the lane timeout logic from scratch on a new message after a period of inactivity.  The problem with this was that if the lane had been held open by the receiver's firewall as a pinhole that has since closed, the firewall would drop any new packets, and it would take on the order of minutes to resume connectivity between the two ships.

After this change, if the first packet times out on a direct lane that has been inactive for thirty seconds or more, the lane will be marked as indirect, causing retries to hit the peer's sponsor.  This puts marginally more relaying load on sponsors, but should improve reconnection time dramatically.  Relaying is now faster than most other parts of the system, so the extra load shouldn't be too onerous.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1967,20 +1967,12 @@
     ++  on-memo
       |=  [=bone payload=* valence=?(%plea %boon)]
       ^+  peer-core
-      ::  if we haven't been trying to talk to %live, reset timer
-      ::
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
We tested this together and it seems to fix the issue correctly.

PR #4850, opened 2021-05-04
hooks: refactor, fix nacking, slog ghosts

liam-fitzgerald:
hooks: add diplomacy flag

pull-hook: slog tank on versioned nack

Should hopefully provide more debugging info for ghost subscriptions

push-hook: hold subscription open after validating access

When we receive a watch-nack on a versioned subscription, we attempt to
resubscribe on an unversioned path. However, if we receive an unversioned
subscription request for which we are not compatible, we hold the
subscription open and never send any facts on unversioned subscriptions.
As the check for unversioned compatiblility occurs before retrieving the
initial-watch, subscriptions which are supposed to nack never do. Fixed
by moving the conditional.

Fixes urbit/landscape#795

versioning: convert gates return a cage

### DIFF
--- a/pkg/arvo/lib/pull-hook.hoon
+++ b/pkg/arvo/lib/pull-hook.hoon
@@ -463,18 +463,18 @@
             ::  subscription
             tr-core
           (tr-suspend-pub-ver min-version.config)
-        =/  =vase
+        =/  =^cage
           (convert-to:ver cage)
         =/  =wire
           (make-wire /store)
-        =+  resources=(~(gas in *(set resource)) (resource-for-update:og vase))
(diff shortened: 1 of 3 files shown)


tacryt-socryp (commented):
LGTM, but I think we should put some thought into how to release this with minimal network churn before we go about actually pushing this out.

liam-fitzgerald:
This is backwards compatible so we can canary it on infra ships, then release/next-js, then mainnet

tacryt-socryp (approved):
Sg. We should follow up fairly quickly with an OTA to set diplomatic to `%.y`

liam-fitzgerald:
We can just do that on the next OTA that touches a hook, no?

PR #4856, opened 2021-05-05
interface: add basic keybinds

liam-fitzgerald:
included defaults (all customisable):
- move forward/backward in history (ctrl+b, ctrl+f)
- cycle through items in sidebar (ctrl+n, ctrl+p)
- hide sidebar (ctrl+h)


pkg/interface/src/wdyr.js
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 
-if (process.env.NODE_ENV === 'development') {
+if (false && process.env.NODE_ENV === 'development') {
```
matildepark, line 3: Cruft

pkg/interface/src/logic/state/settings.ts
```diff
@@ -60,7 +71,19 @@ const useSettingsState = createState<SettingsState>('Settings', {
   tutorial: {
     seen: true,
     joined: undefined
+  },
+  keyboard: {
```
jamesacklin, line 75: Suggestions:

History back: `ctrl + [`
History forward: `ctrl + ]`
Previous channel: `ctrl + ;`
Next channel: `ctrl + '`
Toggle sidebar: `ctrl + \`

Ergonomically:
image

urcades: I actually love these — @liam-fitzgerald whaddya think?

liam-fitzgerald: Sounds good although I haven't used a normal keyboard in a minute so not sure how those would feel in practice

urcades: What kind of keyboard ya using? 

I tested this on a thinkpad and macbook and a 70% mech keyboard, it's pretty cozy

jamesacklin: `~mirmus-dozryx/clack` is calling you @liam-fitzgerald

liam-fitzgerald: Kinesis Advantage

urcades:
Unsure how trivial it is to extend this PR (or if extending it is welcome), but for record's sake, I've added a deeply desired keybinding set for notifications:

https://github.com/urbit/landscape/issues/871

matildepark (approved):
It's the basics. Feels nice, though.

PR #4857, opened 2021-05-05
permalinks: appearance uniformity tweaks

jamesacklin:
This PR brings the appearance of transcluded permalinks into line with what Design has provided and asked for.

For all of these, the default click action of the entire block is to navigate the user to the permalink source. Clicking on a URL navigates the user to that URL; clicking the embed caret unfolds the content; clicking the external arrow in images opens the image in a new tab. Embedded permalinks use the `pointer` cursor. 

Fixes urbit/landscape#759

Also fixes the `GraphContentWide` padding; fixes urbit/landscape#823

_Does not include_ fixing URL truncation from urbit/landscape#777 since that change might be difficult to read in this context.

Looking for comments on how to determine the reference's context (handled with location and some dumb equalities now, would like to make that much cleaner).

## Chat permalinks in chat channels

A chat message from the same chat channel ("reply"):
image

A single-line chat message from another chat channel:
image

A multi-line chat message from another chat channel:
image

A chat message from another chat channel containing a picture:
image

A chat message from another chat channel containing an external link:
image

A chat message from another chat channel containing embedded content:
image
image

A chat message from another channel containing a code snippet:
image

## Collection permalinks in chat channels

A collection item from another channel:
image

## Collection permalinks in collection channels

A collection item from the same channel:
image

Collection items from other channels is unfortunately still unsupported:
image

## Publish permalinks in chat channels

A note preview:
image

A comment on a note:
image

## Group feed permalinks in chat channels

Group feed item with no replies: 
image

Group feed item with replies:
image

## Group permalinks in chat channels

A joined group:
image


urcades:
YOU ARE A KING. 

Just leaving this here while I actually set out time to check it all out 🥇

matildepark:
> Collection items from other channels is unfortunately still unsupported

Huh, TIL. This just seems like an oversight, there can't possibly be a hard limitation here?

pkg/interface/src/views/apps/chat/ChatResource.tsx
```diff
@@ -89,7 +89,7 @@ function ChatResource(props: ChatResourceProps) {
         props.association.resource,
         msg.index
       );
-      return `${url}\n~${msg.author} : `;
+      return `${url}\n`;
```
liam-fitzgerald, line 92: we aren't mentioning people in replies anymore?

liam-fitzgerald: cc: @urcades

liam-fitzgerald (commented):
need design approval to remove mentions from replies, else lgtm

pkg/interface/src/views/apps/chat/ChatResource.tsx
urcades: Yes, we are removing them — I wanted to make sure though, will people still receive notifications whenever their messages are replied to or referenced?

urcades: To be clear:

If notifications (for people using your posts as references) are reliant on **explicit** mentions, we need to generally engineer these references such that notifications are generated whenever one of your nodes are referenced.

If this isn't easy for whatever reason, or out of scope, we should continue to include them in the reply text

cc @matildepark — let me know if this ask (generating hark notifications whenever one's node is referenced) is out of scope for this work

matildepark: image

This did not notify me

matildepark: That is definitely 100% out of scope of this PR @urcades, let's just retain mentions

urcades: damn it, ok — will cover the mechanics of this UX in another issue

matildepark:
Merge conflict!

matildepark:
image

This needs a tweak, otherwise good to go.

jamesacklin:
image

matildepark (approved):

PR #4858, opened 2021-05-05
gen: warn and prompt for confirmation in |code %reset

joemfb:
Because `|code %reset` logs out all active web sessions (and does not produce or print the new code), it causes problems if run through web terminal, without access to the actual terminal. This feature should be redesigned for compatibility without hosting.

In the meantime, this PR adds a warning and requires for confirmation before reset. Wording improvements welcome.

/cc @brendanhay

### DIFF
--- a/pkg/arvo/gen/hood/code.hoon
+++ b/pkg/arvo/gen/hood/code.hoon
@@ -32,8 +32,8 @@
   [%& %project "y/n: "]
 %+  parse
   ;~  pose
-    (cold %.y ;~(pose (just 'y') (just 'Y')))
-    (cold %.n ;~(pose (just 'n') (just 'N')))
+    (cold %.y (mask "yY"))
+    (cold %.n (mask "nN"))
   ==
 |=  reset=?
 ?.  reset
--- a/pkg/arvo/gen/hood/code.hoon
+++ b/pkg/arvo/gen/hood/code.hoon
@@ -3,15 +3,14 @@
 ::::  /hoon/code/hood/gen
   ::
 /?    310
-::
-::::
-  ::
-:-  %say
+/-  *sole
+/+  *generators
+:-  %ask
 |=  $:  [now=@da eny=@uvJ bec=beak]
         [arg=?(~ [%reset ~]) ~]
     ==
 =*  our  p.bec
-:-  %helm-code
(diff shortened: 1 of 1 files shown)


joemfb:
see also urbit/urbit.org#618

matildepark:
May be preferable to target master, next-userspace doesn't land for a while.

pkg/arvo/gen/hood/code.hoon
```diff
@@ -20,11 +19,23 @@
...
 ?>  =(%reset -.arg)
-%reset
+%+  print  'continue?'
+%+  print  'warning: resetting your code closes all web sessions'
+%+  prompt
+  [%& %project "y/n: "]
+%+  parse
+  ;~  pose
+    (cold %.y ;~(pose (just 'y') (just 'Y')))
+    (cold %.n ;~(pose (just 'n') (just 'N')))
```
Fang-, line 36:
```suggestion
    (cold %.y (mask "yY"))
    (cold %.n (mask "nN"))
```

Fang- (approved):
Neat! Parser code could be a bit tighter, but looks good otherwise. No strong opinions about the copy here.

joemfb:
@matildepark i just want it go out with the next OTA, is there a better branch for that?

matildepark:
Arguably next-js is most likely to hit soonest. What we normally do is hold "Arvo-side userspace fixes" with a target of master and ship them as they come in, but that defeats the purpose of _next_ userspace, so I'll have to change the process here, clearly.

joemfb:
Sounds good, i've rebased onto master.

vcavallo:
Does this make it impossible to reset the code over the JSON API? Or is there a way to send along a preemptive "YES"? Otherwise, the request returns the prompt.

Simple example:

```
+$ curl -s -X POST -H "Content-Type: application/json" \
+>       -d '{ "source": { "dojo": "+hood/code %reset" }, "sink": { "app": "hood" } }' \
+>       http://127.0.0.1:12322
"warning: resetting your code closes all web sessions\ncontinue?\n"
```

Fang-:
@vcavallo you could `:hood &helm-code %reset` to do the hood poke that the generator does directly.

PR #4860, opened 2021-05-05
interface: Converting all to typescript

tylershuster:
As part of the http-api replacement I need to have better TS checking everywhere the API is used — I've already done this twice (https://github.com/urbit/urbit/pull/4482 and https://github.com/urbit/urbit/pull/4611) and I don't want to do it again, so this is an interim PR that converts everything to Typescript and in doing so cleans up some linting errors.

Most of the linting problems fixed here are just autofixing, but there's more typing that needs to be done. This all builds and everything, but I plan to keep adding to this until it's more or less comprehensive.


matildepark:
I'm fine doing a once-over functionality test and merging this if it makes your work easier. Did you _want_ us to hold this open? Also #4859 may cause conflict here, maybe this should merge first.

matildepark (approved):

PR #4868, opened 2021-05-06
permalinks: fix crashing slice error

jamesacklin:
Doesn't calculate the index of the transcluded node if the node prop is undefined because of a timing issue in the parent component or any other reason.

fixes urbit/landscape#841

### DIFF
--- a/pkg/interface/src/views/apps/permalinks/TranscludedNode.tsx
+++ b/pkg/interface/src/views/apps/permalinks/TranscludedNode.tsx
@@ -20,7 +20,7 @@ function TranscludedLinkNode(props: {
   api: GlobalApi;
 }) {
   const { node, api, assoc, transcluded } = props;
-  const idx = node.post.index.slice(1).split('/');
+  const idx = node?.post.index.slice(1).split('/');
 
   switch (idx.length) {
     case 1:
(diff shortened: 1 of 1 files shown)


matildepark:
I mean, this will still crash, you'll want `?.` on every single access level here and then a nullish coalesce (`??`) with a sane default

matildepark:
Well, neither of us can repro this, and @urcades is on hosting ... I guess we can let it hit next-js and ask Ed.

pkg/interface/src/views/apps/permalinks/TranscludedNode.tsx
```diff
@@ -20,7 +20,7 @@ function TranscludedLinkNode(props: {
   api: GlobalApi;
 }) {
   const { node, api, assoc, transcluded } = props;
-  const idx = node.post.index.slice(1).split('/');
+  const idx = node?.post?.index?.slice(1).split('/') ?? [];
```
matildepark, line 23: `.slice(1).split()` will also error because .slice will return undefined and we'll attempt to access split on it.

matildepark: You want every single access to be forgiving of `undefined` and then let `??` catch

matildepark (approved):

PR #4869, opened 2021-05-06
publish: amazon

matildepark:
- Enables forestry features for all users now that the renderer is on livenet. Mentions, group links, transclusions, etc.
- Makes headers blocks.
- Fixes links with proper href now.
- Adds break elements to renderers.
- Passes `tall` to the renderer so it doesn't treat all notes as inline. Changes the container for tall content into a grid so that `row-gap` works and separates content.


matildepark:
Fixes https://github.com/urbit/landscape/issues/843

tacryt-socryp (approved):
All changes look exceedingly sane. I'm surprised there's any difference between `tall` and `tall={true}` though

matildepark:
> All changes look exceedingly sane. I'm surprised there's any difference between `tall` and `tall={true}` though

There isn't, I just retained an artifact of dev

PR #4870, opened 2021-05-06
ResourceSkeleton: always flex-shrink

matildepark:
Fixes urbit/landscape#842

She said, "it's literally always better to flex truncate a title than to push UI out of the container. lol. lmao"

Screen Shot 2021-05-06 at 1 13 03 PM


matildepark:
Oh, wait, urbit/landscape#842 as specified really _was_ actually fixed by #4780! But overlapping New Post is still an issue on long titles, so this stands.

tacryt-socryp (approved):
I believe in it

PR #4872, opened 2021-05-06
more typescript cleanup

tylershuster:
Another TS cleanup, pushing this for further commits


matildepark:
Flatten into one commit?

matildepark (changes requested):
This doesn't actually compile — PostItem.tsx crashes. Can you try building and doing a runthrough and tag me again if you have reasonable confidence?

tylershuster:
Yep, I'm sure I know why it crashed before and am confident it won't again.

matildepark (approved):
Human selenium'd through each route and corrected one prop that really should be a string.

matildepark:
123 errors remaining.

PR #4873, opened 2021-05-06
interface: adding basic notifications

tylershuster:
dumb, works.

image

image

Button only shows if they're not allowed. Use the browser's UI to disable them


tylershuster:
@liam-fitzgerald Can you tell me offhand to get the reference of the notification? All the hark updates are...ill-documented

pkg/interface/src/logic/reducers/hark-update.ts
```diff
@@ -187,7 +187,14 @@ function readSince(json: any, state: HarkState): HarkState {
 
 function unreadSince(json: any, state: HarkState): HarkState {
   const data = _.get(json, 'unread-count');
-  if(data) {
+  if (data) {
+    if (!useHarkState.getState().doNotDisturb && !document.hasFocus()) {
+      new Notification('New Landscape Notification', {
+        tag: 'landscape',
```
liam-fitzgerald, line 193: ? This will notify on literally every message lol. Move this logic to 'add-note'

liam-fitzgerald (changes requested):

matildepark (approved):
Sure is basic notifications!

PR #4877, opened 2021-05-07
graph-push-hook: add better hints for profiling

tacryt-socryp:


matildepark:
Bunch of jet registrations. Is there a reason this has to hit next-js instead of just running it on an infra ship?

tacryt-socryp:
I like the idea of putting it on next-js because it introduces no changes to the API, and allows for any ship on our dev channel to have more accurate performance profiling sooner. There's no real reason to deploy this to an infrastructure ship specifically.

tacryt-socryp:
This isn't debugging information, as it has no effect unless you turn on tracing with `-j`.

tacryt-socryp:
Screen Shot 2021-05-07 at 12 24 02 PM

joemfb:
It does add registrations, and those persist (cold jet state). We don't currently have any way to reclaim them, although I have plans. These cores are pretty small, so I'm not sure how much it matters. On my ship, I have ~24 MB in cold jet state after `|meld`; last era, that go into the hundreds.

liam-fitzgerald (approved):

PR #4879, opened 2021-05-07
permalinks: present error for unavailable references

jamesacklin:
Presents a message when a reference is unavailable -- in this case, it's referring to a chat message in a channel in a private group I have not joined.

image


matildepark:
image

So: note that I'm _in_ that chat. If I scroll up, the reference is visible. So this says a reference is unavailable if it's nested a layer, which is not accurate to the user.

matildepark (approved):

PR #4880, opened 2021-05-07
graph-push-hook + store: 30% speed up of graph messages by using persistent caching

tacryt-socryp:
Just PRing it against this branch so you can review it easier, not sure where to put this yet. Speed up processing time of a single chat message / graph-store message event from 500ms to 350ms. All the gains come from %graph-push-hook, which became significantly slower earlier this year when we implemented a permissioning model and started using host timestamps. 30% faster on our most used hot path!

Before:
Screen Shot 2021-05-07 at 3 45 34 PM

After:
Screen Shot 2021-05-07 at 3 48 35 PM

After adding graph-store dais caches:
Screen Shot 2021-05-10 at 12 26 27 PM

### DIFF
--- a/pkg/arvo/app/graph-push-hook.hoon
+++ b/pkg/arvo/app/graph-push-hook.hoon
@@ -26,6 +26,8 @@
       state-one
   ==
 ::
+::  TODO: come back to this and potentially use send a %t
+::  to be notified of validator changes
 +$  cache
   $:  graph-to-mark=(map resource:res (unit mark))
       perm-marks=(map [mark @tas] tube:clay)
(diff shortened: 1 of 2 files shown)


joemfb (approved):
As we've discussed out-of-band, i think this could be almost as effective (an order of magnitude simpler) if it simply imported the hot-path marks and conversion gates, falling back to uncached scries for all others. And for the longer term, I'd like the system to just provide the equivalent functionality in a performant way.

But for now, if you're comfortable with the maintenance burden, this seems reasonable.

LGTM

liam-fitzgerald (approved):
LGTM, pending comments

PR #4882, opened 2021-05-09
jets: fix malloc check

l2dy:
`output` would never be NULL.


joemfb (changes requested):
This is a great catch! But, hilariously enough, the current implementation is "accidentally correct".

The u3 allocator will never return `NULL` on OOM -- instead it will "bail" (exit the virtual context / throw an exception). So this function should simply `return 1;`.

l2dy:
> The u3 allocator will never return `NULL` on OOM -- instead it will "bail" (exit the virtual context / throw an exception). So this function should simply `return 1;`.

PR updated.

joemfb (approved):
LGTM, thanks!

PR #4884, opened 2021-05-09
groups: World Privacy icon fix

jamesacklin:
Although the bug was poetic, this makes the icon in GroupLink aware of the group's hidden status and surfaces a lock icon in those cases instead of the "public" globe icon.


urcades:
Yes

matildepark (approved):

PR #4886, opened 2021-05-10
GraphContent: rendering fixes 

liam-fitzgerald:
- Preserve newlines
- Fix lineheight
- Disable nested blockquotes
- Disable inline html

### DIFF
--- a/pkg/interface/package-lock.json
+++ b/pkg/interface/package-lock.json
@@ -9133,9 +9133,38 @@
       }
     },
     "remark-breaks": {
-      "version": "2.0.1",
-      "resolved": "https://registry.npmjs.org/remark-breaks/-/remark-breaks-2.0.1.tgz",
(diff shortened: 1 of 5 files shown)


matildepark:
Mind making links inline instead of inline blocks?

matildepark:
For future reference, can you start committing lockfiles when you change dependencies? It wreaks a little bit of havoc otherwise.

matildepark:
The line-height was better at 20px I think, now the chat interface doesn't line up.

liam-fitzgerald:
Messages with newlines look a little squashed to me with the old line-height

matildepark:
> Messages with newlines look a little squashed to me with the old line-height

Behold my neurosis! This before/after with the timestamp off with the line-height! @urcades 

Screen Shot 2021-05-10 at 12 02 20 AM

matildepark (approved):

PR #4892, opened 2021-05-11
interface: added component class names

tylershuster:
Where a component returned a single, sane, top-level classable element, added the name of the component as a kebab-cased class name

Frees up development time by preventing @Fang- from writing new CSS every OTA.


matildepark:
I feel like this would be better solved by having each component in the architecture automatically generate and append an eponymous classname to its output's container during the build process than by doing this by hand forever onward.

tylershuster:
I agree, but what about `<><X /><Y /></>` components, or components that just return another component? Or components that return `<Formik whichIsFake><Form className='class-name'>...</Form></Formik>`. There's all sorts of similar examples in this PR.

PR #4893, opened 2021-05-11
meta: auto-deploy release/next-userspace

matildepark:
This PR retires next-js in favour of next-userspace, auto-deploying next-userspace to the dev stream. 

Going forward _all possible changes_ should target next-userspace for staging deployment; irrevocably breaking changes will target the next dated release branch (eg. 1.8's branch will be `release/2021-5-27`).

### DIFF
--- a/.github/workflows/merge-master.yml
+++ b/.github/workflows/merge-master.yml
@@ -6,13 +6,13 @@ on:
 jobs:
   merge-to-next-js:
     runs-on: ubuntu-latest
-    name: "Merge master to release/next-js"
+    name: "Merge master to release/next-userspace"
     steps:
       - uses: actions/checkout@v2
       - uses: devmasx/merge-branch@v1.3.1
         with:
           type: now
-          target_branch: release/next-js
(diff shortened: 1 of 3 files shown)


matildepark:
And yes, @tacryt-socryp would prefer `release/next-ui` but I would rather not introduce new branch names, and I don't see the point in calling it "user interface" compared to userspace — I would want generators and the like in next-userspace, full stop. 

Until Landscape properly splits out, the branch is exactly what it says — the next userspace release.

philipcmonk (approved):
Sounds good to me.  For brevity, we can always use "newserspace"...

tacryt-socryp (approved):

.github/workflows/merge-master.yml
```diff
@@ -6,13 +6,13 @@ on:
 jobs:
   merge-to-next-js:
```
jtobin, line 7: I'm a little too late, but this job has retained an unfortunate name.. 😄

PR #4897, opened 2021-05-13
interface: optimistic updating

liam-fitzgerald:
- Optimistic contacts and notifications updates
- Scrap submit buttons in preferences, save on change

### DIFF
--- a/pkg/interface/src/views/apps/settings/components/lib/CalmPref.tsx
+++ b/pkg/interface/src/views/apps/settings/components/lib/CalmPref.tsx
@@ -6,9 +6,11 @@ import {
 import { Form, Formik, FormikHelpers } from 'formik';
 import React, { useCallback } from 'react';
 import GlobalApi from '~/logic/api/global';
-import useSettingsState, { selectSettingsState } from '~/logic/state/settings';
+import useSettingsState, { selectSettingsState, SettingsState } from '~/logic/state/settings';
 import { AsyncButton } from '~/views/components/AsyncButton';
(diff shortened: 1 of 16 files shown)


pkg/interface/src/views/apps/settings/components/lib/CalmPref.tsx
```diff
@@ -22,57 +24,40 @@ interface FormSchema {
...
+        promises.push(api.settings.putEntry(bucket, key, bool));
+      }
+    })
+    await Promise.all(promises);
     actions.setStatus({ success: null });
   }, [api]);
 
   return (
-    <Formik initialValues={initialValues} onSubmit={onSubmit}>
+    <FormikOnBlur initialValues={initialValues} onSubmit={onSubmit}>
```
matildepark, line 60: The issue with FormikOnBlur is that it requires you to click the page to apply the values. Is there a way to apply values on nav as well?

liam-fitzgerald: It's been rework to work on the `dirty` flag instead of the `touched` property, so this should no longer be an issue

matildepark: Oki

matildepark (approved):
This is a fine start in the areas it's in, so I'm not averse to merging this as is, but it won't truly be felt by the user until we do `markCountAsRead` optimistically too, so let's talk about that tonight.

PR #4898, opened 2021-05-13
interface: changed alphabeticalOrder to ignore non-word characters

tylershuster:
~~before/after:~~ after/before:
Screen Shot 2021-05-13 at 2 27 00 PM

Also changes the sidebar sort algorithm — doesn't seem to break anything but I don't think I'm in any groups with emoji channel names.

Fixes https://github.com/urbit/landscape/issues/875


matildepark (approved):
lgtm

PR #4899, opened 2021-05-13
landscape: zero typescript errors

matildepark:
Screen Shot 2021-05-13 at 6 34 28 PM

That's it? Nothing? Silence?

- Lots of `@ts-ignore` flags in a few camps:
   - We ignore zustand types entirely
   - A handful of weird VirtualScroller type issues, flagged @liam-fitzgerald in code
   - forwardRef, memo, and `with(whatever)` all don't seem to pass and typecheck props in a way we can detect, so they always fail
- Otherwise, general TS linting and type guarding
-  Adds `ts-mdast` for type checking our GraphContent properly


liam-fitzgerald (approved):
LGTM, Zustand recently had typing changes that should make not breaking everything a lot easier, will investigate

PR #4903, opened 2021-05-14
landscape: grabbag

matildepark:
- Fixes the null collection icon in Leap (urbit/landscape#876)
- Removes private/public labels from group links that don't actually reflect that information (urbit/landscape#874)
- Group Summary now displays a padlock for closed groups, and a copy icon for public groups (urbit/landscape#411). I took some liberties with the  design, please stamp @urcades:

Public

May-14-2021 12-40-31

Private

Screen Shot 2021-05-14 at 12 41 23 PM

It copies the web+urbitgraph link for the group to clipboard.

### DIFF
--- a/pkg/interface/src/logic/lib/useCopy.ts
+++ b/pkg/interface/src/logic/lib/useCopy.ts
@@ -1,7 +1,7 @@
 import { useCallback, useMemo, useState } from 'react';
 import { writeText } from './util';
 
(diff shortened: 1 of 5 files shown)


urcades:
Design liberties are very excellent, stamp stamp

jamesacklin (approved):

PR #4904, opened 2021-05-14
group feed renovation: social network 1.0

tacryt-socryp:
lmao added threading and a flat timeline view

TODO:
- [x] add methods to ordered-map
- [x] test methods for ordered-map
- [x] write scries that use new ordered-map methods
- [x] modified ordered map
- [x] modified virtual scroller
- [x] timeline view
- [x] thread view
- [x] back button behavior
- [x] pending messages on timeline view
- [x] pending messages on thread view
- [x] pending messages on reply view
- [ ]  Permalinks behavior
- [ ]  Testing admin feed

Not planned for this release, but worth noting:
- We still need to figure out how to display the number of top-level replies to a message without actually fetching all of the replies.


pkg/interface/src/views/landscape/components/Home/GroupFeedHeader.tsx
```diff
@@ -14,9 +14,17 @@ export function GroupFeedHeader(props) {
 
   const locationUrl =
     history.location.pathname.replace(`${baseUrl}/feed`, '');
-  const nodeIndex = locationUrl.split('/').slice(1).map((ind) => {
-    return bigInt(ind);
-  });
+  console.log(locationUrl);
```
liam-fitzgerald, line 17: strip `console.log`

pkg/interface/src/views/landscape/components/Home/Post/PostThread.tsx
```diff
@@ -0,0 +1,124 @@
...
+    graphPath,
+    group,
+    vip,
+    pendingSize
+  } = props;
+
+  const graphResource =
+    graphPath ? resourceFromPath(graphPath) : resourceFromPath('/ship/~zod/null');
+
+  const locationUrl = props.locationUrl.replace(`${baseUrl}/feed/thread`, '');
```
liam-fitzgerald, line 28: It's not a blocker, given that it's just matching the other Post code, but I'd like to voice how much i dislike this pattern. We should use `react-router-dom` instead, which would allow us to only construct the bigints when they change, rather than every render.

pkg/arvo/app/graph-store.hoon
```diff
@@ -826,11 +831,79 @@
...
         [ship term]
       %-  ~(gas by *(map index:store node:store))
-      %+  turn  (tap:orm `graph:store`(subset:orm p.children.u.node end start))
+      %+  turn  (tap:orm `graph:store`(lot:orm p.children.u.node end start))
       |=  [=atom =node:store]
       ^-  [index:store node:store]
       [(snoc index atom) node]
     ==
+  ::
+      [%x %deep-nodes-older-than @ @ @ @ ~]
```
liam-fitzgerald, line 840: Flagging this ship parsing code as something that should be refactored next sprint

liam-fitzgerald (commented):
LGTM! Just went ahead and highlighted some things that should be addressed next sprint, but given that they're all building on the existing style, they shouldn't block this from release. I didn't read the /sys changes, assuming they're all in next-sys. I'm curious, do you have any idea of the performance characteristics of the FE changes? 

(You should treat this as an approval, just holding off on actually stamping until it's finished)

pkg/arvo/app/graph-store.hoon
tacryt-socryp: What about the ship parsing code needs to be refactored?

pkg/interface/src/views/landscape/components/Home/GroupFlatFeed.tsx
```diff
@@ -0,0 +1,140 @@
...
+function GroupFlatFeed(props) {
+  const {
+    baseUrl,
+    api,
+    graphPath,
+    groupPath,
+    vip
+  } = props;
+
+  const groups = useGroupState(state => state.groups);
```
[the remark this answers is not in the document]
tacryt-socryp: Thanks, this is good advice. I definitely need to do a perf pass on this. It's been good so far, but still

pkg/interface/src/logic/reducers/graph-update.ts
```diff
@@ -22,6 +25,16 @@ export const GraphReducer = (json) => {
   if(loose) {
     reduceState<GraphState, any>(useGraphState, loose, [addNodesLoose]);
   }
+
+  const flat = _.get(json, 'graph-update-flat', false);
```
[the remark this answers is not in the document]
tacryt-socryp: Absolutely, a lot of this reducer code has gotten way too unwieldy

tacryt-socryp:
Performance has been really fast so far, but I haven't stress tested it yet. Memory usage for group feed data will likely be much higher than other graphs given our need to store them in different formats for different rendering styles, but I've worked on ensuring that we only store them multiple times in that fashion when that data has actually been requested to be shown.

pkg/arvo/app/graph-store.hoon
liam-fitzgerald: I'd consider moving the resource parsing out to a `|^`
```hoon
[%graph @ @ *]
=/  =ship  (slav %p i.t.path)
=/  =term  i.t.t.path
|^
?+  t.t.t.path  [~ ~]
   ~   (get-graph ship term)
   [%subset @ @]
  =/  start=@ud  i.t.path
  =/  end=@ud   i.t.t.path
  :: subset
   [%node *]
  ::  node
```

etc. etc.
Mainly for conciseness and readability. (You don't have overly long t.t.t.t.path accesses)

tacryt-socryp: Oh I see. I hadn't considered using that pattern, but I like it! Happy to refactor in that way

pkg/arvo/sys/hoon.hoon
```diff
@@ -5452,6 +5452,7 @@
 ::::  4k: atom printing
   ::
 ++  co
+  !:
```
liam-fitzgerald: drop this?

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1967,20 +1967,12 @@
...
-              |=  =message-pump-state
-              =(~ live.packet-pump-state.message-pump-state)
-          ==
-        now
-      ::
       =/  =message-blob  (dedup-message (jim payload))
       =.  peer-core  (run-message-pump bone %memo message-blob)
       ::
-      ?:  &(=(%boon valence) ?=(?(%dead %unborn) -.qos.peer-state))
+      ?:  ?&  =(%boon valence)
```
liam-fitzgerald: ??? Is this cruft from the next-sys merge?

pkg/interface/src/logic/api/graph.ts
```diff
@@ -361,15 +361,18 @@ export default class GraphApi extends BaseApi<StoreState> {
...
     );
     const node = data['graph-update'];
     this.store.handleEvent({
       data: {
-        'graph-update-thread': node,
+        'graph-update-thread': {
+          index,
+          ...node
+        },
         'graph-update': node
```
liam-fitzgerald, line 377: I think that the store should only handle `graph-update-thread` and not `graph-update` because we might potentially lose some of our ordering invariants with respect to missing backlog

pkg/interface/src/views/landscape/components/Home/GroupFeedHeader.tsx
```diff
@@ -1,20 +1,29 @@
...
   const { baseUrl, history, graphResource, vip } = props;
+
   let graph = props.graph;
   const historyLocation = history.location.pathname;
   const graphId = `${graphResource.ship.slice(1)}/${graphResource.name}`;
 
+  const [locationList, setLocationList] = useState([]);
+  useEffect(() => {
+    locationList.push(history.location.pathname);
+    setLocationList(locationList);
```
liam-fitzgerald, line 18:
```typescript
setLocationList(ls => [...ls, history.location.pathname]);
```
please do not mutate state

pkg/interface/src/views/landscape/components/Home/Post/PostItem/PostItem.tsx
```diff
@@ -80,21 +80,12 @@ class PostItem extends React.Component<PostItemProps, PostItemState> {
...
     });
 
-    history.push(`${baseUrl}/feed/replies${indexString}`);
-  }
-
-  navigateToThread() {
-    const { history, baseUrl, index, isParent } = this.props;
-    if (isParent) {
-      return;
+    //  TODO: ensure that the logic here works properly
```
liam-fitzgerald, line 83: is this TODO addressed?

pkg/interface/src/logic/state/graph.ts
```diff
@@ -148,6 +148,28 @@ export function useGraph(ship: string, name: string) {
   );
 }
 
+export function useFlatGraph(ship: string, name: string) {
+  return useGraphState(
+    useCallback(s => s.flatGraphs[`${deSig(ship)}/${name}`], [ship, name])
+  );
+}
+
+export function useThreadGraph(ship: string, name: string, index: string) {
```
liam-fitzgerald, line 157: +1 for these accessors, very nice

liam-fitzgerald (commented):
LGTM, some outstanding questions though

pkg/arvo/sys/hoon.hoon
liam-fitzgerald: nvm

pkg/arvo/sys/vane/ames.hoon
liam-fitzgerald: nvm

PR #4905, opened 2021-05-17
GraphContent: improve rendering

liam-fitzgerald:
- Uses actual `<li><ol><ul>` elements from indigo
- Emphasis and strong compose properly
- Blockquote parsing fixed to break after one newline e.g.
```
> line
break
```
produces
>line 

break
- Newlines fixed


matildepark:
Failed tsc, mind patching?

pkg/interface/src/views/landscape/components/Graph/blockquote.js
```diff
@@ -0,0 +1,124 @@
+/** pulled from remark-parse
+ * 
+ *  critical change is that blockquotes require a newline to be continued, see
+ *  the `if(!prefixed) conditional
+ */
+'use strict'
+
+var trim = require('trim')
+var interrupt = require('remark-parse/lib/util/interrupt')
+
```
tacryt-socryp, line 10: Truly horrifying, but looks good to me. Let's get this out onto dev stream.

tacryt-socryp (approved):

PR #4906, opened 2021-05-17
hark: optimistic read

liam-fitzgerald:
- Reads chats and comments optimistically, to avoid unnecessary lag


matildepark (approved):
People are gonna notice this one.

PR #4907, opened 2021-05-17
dm-hook: boot fixes

liam-fitzgerald:
- Autoboots dm-hook on upgrade and initialisation
- Fixes the +on-init logic scry


tacryt-socryp (approved):
LGTM. How did the `%gu` scry work before if it was using the wrong args?

liam-fitzgerald:
I didn't :/

PR #4908, opened 2021-05-17
Matrix Library (Hoon)

sigilante:
This PR reflects a subset of the changes in #3536, which has become rather unwieldy. This is the Hoon-only matrix-only code for @lms and @lmd.

Compare #4455, the vector library components.

I have some refactoring to do in this branch, I will ping you when ready.


sigilante:
OK, this one is quite close to done.

- [ ] Beef up test cases.
- [ ] Prune out unused inversion functions.

sigilante:
Update:  all legacy unit tests work with the current atom format now.  I need to add a few more categories of tests (currently marked with `!!`), then prune out unused auxiliary functions for matrix inversion.

sigilante:
feelsgoodman.jpeg

```
built   /tests/lib/lm/hoon
execute: run: MB/6.797.416
execute: run: MB/8.270.404
built   /tests/lib/lv/hoon
execute: run: MB/7.955.580
```

sigilante:
OK, I'm actually using all of the arms still in there.

What I _don't_ like is that the matrix functions fail in a relatively opaque way for singular matrices.  They _should_ issue a warning but instead they just return an array of `nan`s or `inf`s.

I'd also like to come back and add a proper `++expect-close` testing arm here.

OTHER THAN THAT @joemfb this is ready for review.

sigilante:
Hmm, looks like it's choking a bit on floating-point equality comparison.  Guess I'll need to write an `++expect-close:lms` now instead.  So pend the review until I set that up.

sigilante:
Está hecho.  `++expect-close` operates on `@lms`; I should add a similar one for `@lvs`.  It doesn't pattern-match exactly with the other `++expect` arms but whatevs.

PR #4910, opened 2021-05-18
hark: new boxing model

liam-fitzgerald:
- Implements the new boxing model as specified by @g-a-v-i-n 
Screen Shot 2021-05-18 at 11 02 10 am


pkg/interface/src/logic/reducers/hark-update.ts
```diff
@@ -414,6 +389,22 @@ function notifIdxEqual(a: NotifIndex, b: NotifIndex) {
...
+    };
+  } else if ('group' in a && 'group' in b) {
+    return {
+      group: [...a.group, ...b.group]
+    };
+  }
+  return a;
+}
+
+
```
tacryt-socryp, line 405: do you need this many new lines? :)

tacryt-socryp (approved):
LGTM, this is a major rewrite, so it's going to need a lot of testing. Why do you no longer need the inflated-cache?

liam-fitzgerald:
Don't need caching anymore because unread notifications are now always in
unread-notes, so it's not hard to look them up

——
~hastuc-dibtux

https://urbit.org

PR #4911, opened 2021-05-18
interface: misc grabbag

liam-fitzgerald:
Groups: fix keying

GroupSwitcher: update recent groups on unmount

Fixes an unfiled bug where switching to a non-group workspace would not
add the most recent group to the group switcher

graph-reducer: ignore superfluous updates

Reduces RAM usage drastically, especially in long-lived sessions

### DIFF
--- a/pkg/interface/src/logic/api/graph.ts
+++ b/pkg/interface/src/logic/api/graph.ts
@@ -328,6 +328,7 @@ export default class GraphApi extends BaseApi<StoreState> {
 
   async getNewest(ship: string, resource: string, count: number, index = '') {
     const data = await this.scry<any>('graph-store', `/newest/${ship}/${resource}/${count}${index}`);
+    data['graph-update'].fetch = true;
     this.store.handleEvent({ data });
   }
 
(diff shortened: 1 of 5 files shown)


pkg/interface/src/logic/api/graph.ts
```diff
@@ -328,14 +328,15 @@ export default class GraphApi extends BaseApi<StoreState> {
 
   async getNewest(ship: string, resource: string, count: number, index = '') {
     const data = await this.scry<any>('graph-store', `/newest/${ship}/${resource}/${count}${index}`);
+    data['graph-update'].fetch = true;
```
tacryt-socryp, line 331: I assume we currently always use this method to fetch the first page of results in our various graph apps?

pkg/interface/src/logic/reducers/graph-update.ts
```diff
@@ -180,9 +180,15 @@ const addNodes = (json, state) => {
 
     const resource = data.resource.ship + '/' + data.resource.name;
     if (!(resource in state.graphs)) {
-      state.graphs[resource] = new BigIntOrderedMap();
+      if(json.fetch) {
+        state.graphs[resource] = new BigIntOrderedMap();
+      } else {
+        //  ignore updates until we load backlog deliberately, to avoid
```
tacryt-socryp, line 186: Clear, makes sense

tacryt-socryp (approved):
LGTM, one question that I'd like you to check on just to make sure, but this should go in soon, it's much needed.

pkg/interface/src/logic/api/graph.ts
liam-fitzgerald: Yeah, when we aren't loading the entire graph at once.

PR #4912, opened 2021-05-18
permalinks: graph node placeholders

jamesacklin:
Adds an sigil/author skeleton loader to transcluded blocks such that the chat view doesn't dramatically jump with unloaded content. Sets a 'pending' flag in state in the async call, as it turns out this was much simpler and reliable than checking for undefined props.

https://user-images.githubusercontent.com/748181/118661351-070e1f00-b7bd-11eb-85b5-32eb5bd922a6.mp4

Also brings the author/content lines into alignment with Chat references due to the new ChatMessage line-height.

### DIFF
--- a/pkg/interface/src/views/apps/permalinks/embed.tsx
+++ b/pkg/interface/src/views/apps/permalinks/embed.tsx
@@ -15,11 +15,21 @@ import { TranscludedNode } from "./TranscludedNode";
 
 function Placeholder() {
   return (
-    <Row margin='12px' marginBottom='0' height='4'>
-      <Box backgroundColor='washedGray' size='4' marginRight='2' borderRadius='2' />
-      <Box backgroundColor='washedGray' heigh='4' width='75%' borderRadius='2' />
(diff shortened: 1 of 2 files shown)


matildepark:
The popped in content is twice as big as the placeholder. Might be worth adding another placeholder line?

matildepark:
Likewise I would probably suggest each transcluded node type have its own placeholder with its general size (notes generally look the same, links, etc)

matildepark (approved):
Excellent, thank you

pkg/interface/src/views/apps/permalinks/embed.tsx
```diff
@@ -131,8 +154,8 @@ function GraphPermalink(
         navigate(e);
       }}
     >
-      {loading && Placeholder()}
-      {showTransclusion && index && (
+      {loading && association && Placeholder(association.metadata.config.graph)}
```
matildepark, line 157: You failed tsc because of this though! Try `(association.metadata.config as GraphConfig).graph` and import `GraphConfig` type from `@urbit/api`

PR #4913, opened 2021-05-18
chat: fix black-on-black text

jamesacklin:
Before:
image

After:
image

### DIFF
--- a/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
+++ b/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
@@ -283,9 +283,9 @@ const renderers = {
   },
   paragraph: ({ children }) => {
     return (
-      <Box color="black" fontSize={1} lineHeight="tall">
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
```diff
@@ -283,7 +283,7 @@ const renderers = {
   },
   paragraph: ({ children }) => {
     return (
-      <Box fontSize={1} lineHeight="tall">
+      <Box color="black" fontSize={1} lineHeight="tall">
```
matildepark, line 286: Someone has to stop Liam from using Box like this. Can you wrap this in a `<Text>` child?

jamesacklin: He Simply Must Be Stopped

matildepark:
One sec, I'm just gonna handle these all in one PR.

PR #4919, opened 2021-05-19
interface: another markdown pass

liam-fitzgerald:
### Overview

This PR adds storybook to the landscape SPA, and includes two stories for our most heavily reused and buggy component `GraphContent`, in an effort to slow the pace of accidental breakages to the rendering, and cover all edge cases. This should probably be paired with [Chromatic](https://www.chromatic.com/) or something similar, to allow for comparisons and to integrate it through CI. Additionally, the fixtures for this being static will help bisect whether rendering breakage is coming from `tokenizeMessage` or the renderer. Includes bugfixes for most of the outstanding issues against `GraphContent` rendering, however list indicators are broken on darkmode and an indigo patch is out to fix it. https://github.com/urbit/indigo-react/pull/109

#### Running storybook
Simply `npm run storybook` in the pkg/interface folder. However, both indigo themes are currently broken as storybook doesn't support importing typescript directly from npm packages. I've pinged @g-a-v-i-n to fix this, as we really should distribute TS libraries with JS and declaration maps. To work around this, you'll need to `cd node_modules/@tlon/indigo-dark && npm i && npm run build`, for both indigo-dark and indigo-light. 

##### Demo
tall:
Screen Shot 2021-05-19 at 3 02 06 pm
Screen Shot 2021-05-19 at 3 02 14 pm

wide:
Screen Shot 2021-05-19 at 3 03 03 pm

---
### Changelog

RemoteContent: expand to fill if tall

interface: reset UA css for lists

GraphContent: fix leading inline content breaking stitching

GraphContent: add table support

GraphContent: support strikethrough

GraphContent: don't linebreak on empty text nodes

GraphContent: add stories to document edge cases

interface: add storybook

### DIFF
--- a/pkg/interface/.storybook/main.js
+++ b/pkg/interface/.storybook/main.js
@@ -1,0 +1,10 @@
+module.exports = {
+  "stories": [
+    "../src/**/*.stories.mdx",
+    "../src/**/*.stories.@(js|jsx|ts|tsx)"
+  ],
+  "addons": [
+    "@storybook/addon-links",
+    "@storybook/addon-essentials"
+  ]
+}
--- a/pkg/interface/.storybook/preview.js
+++ b/pkg/interface/.storybook/preview.js
@@ -1,0 +1,133 @@
+import dark from '@tlon/indigo-dark';
+import light from '@tlon/indigo-light';
+import { Reset } from '@tlon/indigo-react';
+import { BrowserRouter } from 'react-router-dom';
+import { ThemeProvider } from 'styled-components';
+import useGraphState from '~/logic/state/graph';
+import useMetadataState from '~/logic/state/metadata';
+import '~/views/landscape/css/custom.css';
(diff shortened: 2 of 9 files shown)


matildepark:
As per https://github.com/urbit/landscape/issues/867 — 

image

jamesacklin:
Got Storybook running. This is honestly a huge net gain and I'm a big fan of it. Essentially this means we have an isolated environment to build and test UI components in concert with design, either as a first pass before integration or after, when we're doing refinement passes.

As far as the content itself:

image

These decimal list items are browser-default styled; I can see that we're rendering a block-level element as the first child of the `li`, but in order for everything to use the same font, `li` itself needs a `font-face` declaration.

liam-fitzgerald:
This is addressed in the indigo-react PR that I linked in the description

——
~hastuc-dibtux

https://urbit.org

matildepark (approved):

PR #4920, opened 2021-05-19
interface: all visible 'members' or 'participants' to 'peers'

jamesacklin:
In order to reinforce that Urbit is, in fact, p2p, this change migrates all instances of 'members' or 'participants' in group contexts to 'peers,' which is harmonious with the new label in GroupLink.

image
image
image
image
image

Drafting this for now as I'm sure I'm going to find more.

### DIFF
--- a/pkg/interface/src/logic/lib/tutorialModal.ts
+++ b/pkg/interface/src/logic/lib/tutorialModal.ts
@@ -86,7 +86,7 @@ export const progressDetails: Record<TutorialProgress, StepDetail> = {
   'group-desc': {
     title: 'What\'s a group',
     description:
-      'A group contains members and tends to be centered around a topic or multiple topics.',
+      'A group contains peers and tends to be centered around a topic or multiple topics.',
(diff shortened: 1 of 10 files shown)


tacryt-socryp:
This seems strange to me, as a group genuinely does have members. Peer is a term generally used with the connotation that your server has connected to theirs, which is not necessarily the case if they are just in your group.

matildepark:
> This seems strange to me, as a group genuinely does have members. Peer is a term generally used with the connotation that your server has connected to theirs, which is not necessarily the case if they are just in your group.

The issue is that we've set up a precedent where group members are labelled peers:

image

This is likely as per design conversations, in which case @jyng @urcades do we have a distinct philosophy over the terminology? Personally, I see flattening everything to 'peer' as being pretty intuitive, but I wonder about your own thinking.

tacryt-socryp:
I'd rather flattening everything to "members" since that seems very literal

jamesacklin:
I'm gonna close this. Git means never having to say you're sorry, but this seems like a stupid one to leave hanging open.

PR #4921, opened 2021-05-19
graph-validators: hotfix broken vale:dais caused by removal of +bunt from clay

tacryt-socryp:
Introduced by lack of full mark testing on: https://github.com/urbit/urbit/pull/4554

Fixes urbit/landscape#889

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -1,6 +1,5 @@
 ::  graph-store [landscape]
 ::
-::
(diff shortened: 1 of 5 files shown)


philipcmonk (approved):
@belisarius222 Interesting issue.  We assumed +noun:grab always had a bunt, but this wasn't true.  Now every mark has a bunt which actually passes its +vale, which is probably good.

belisarius222:
Good catch.  I had in fact assumed `+noun:grab` always had a bunt.

PR #4922, opened 2021-05-19
eyre: send warp to warm %f cache for any mark that gets used and switch to %f from %c

tacryt-socryp:
A draft of work to prime caches for any `%c` conversion that `%eyre` does to json.

Note: I don't know much about ducts or how to handle the response I get back from clay (or if I need to), but this compiles so I figured I'd push it up and get somebody to help me figure out the rest.

Edit: Can confirm that this works in practice on livenet and shaves 70ms off of a `%graph-update` event.

Given that I've tested this on my livenet ship, I'm marking this as ready, but I'd still like some comments on the duct stuff / ways to heat these caches more efficiently (getting rid of the time spent in nest checks would be ideal).

Edit: Latest version takes about 300ms per event, so this overall gives us a 40% speedup.

Screen Shot 2021-05-20 at 2 32 57 PM

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1642,17 +1642,16 @@
         ::
         =*  have=mark  p.cage.sign
         =*  desc=tape  "from {(trip have)} to json"
-        =/  tube=(unit tube:clay)
-          =/  tuc=(unit (unit cage))
-            (rof ~ %cc [our %home da+now] /[have]/json)
-          ?.  ?=([~ ~ *] tuc)  ~
-          `!<(tube:clay q.u.u.tuc)
-        ?~  tube
-          ((slog leaf+"eyre: no tube {desc}" ~) [~ ~])
-        ::
-        =/  res  (mule |.((u.tube q.cage.sign)))
+        =/  convert=(unit $-(* json))
+          =/  cag=(unit (unit cage))
+            (rof ~ %cf [our %home da+now] /[have]/json)
+          ?.  ?=([~ ~ *] cag)  ~
+          `!<($-(* json) q.u.u.cag)
+        ?~  convert
+          ((slog leaf+"eyre: no convert {desc}" ~) [~ ~])
+        =/  res  (mule |.((u.convert q.cage.sign)))
         ?:  ?=(%& -.res)
-          [`have `[%fact %json p.res]]
+          [`have `[%fact %json !>(p.res)]]
         ((slog leaf+"eyre: failed tube {desc}" ~) [~ ~])
       ::
       ?~  q.jsyn  ~
@@ -1660,7 +1659,7 @@
       :_  ?~  p.jsyn  ~
           :_  ~
           :^  duct  %pass  /pass/(scot %ud request-id)
-          [%c %warp our %home `[%sing %c da+now /[u.p.jsyn]/json]]
+          [%c %warp our %home `[%sing %f da+now /[u.p.jsyn]/json]]
       =*  sign  u.q.jsyn
       =,  enjs:format
       %-  pairs
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -59,7 +59,9 @@
           ::
       ==
       $:  %clay
-          $>(%writ gift:clay)
+          gift:clay
+          ::  $>(%writ gift:clay)
+          ::
   ==  ==
 --
 ::  more structures
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1238,9 +1238,9 @@
         ::NOTE  these will only fail if the mark and/or json types changed,
         ::      since conversion failure also gets caught during first receive.
         ::      we can't do anything about this, so consider it unsupported.
-        ?~  sign=(channel-event-to-sign channel-event)  $
-        ?~  json-with-moves=(sign-to-json request-id u.sign)       $
-        $(events [(event-json-to-wall id p.u.json-with-moves) events])
+        ?~  sign=(channel-event-to-sign channel-event)        $
+        ?~  json-with-moves=(sign-to-json request-id u.sign)  $
+        $(events [(event-json-to-wall id +.u.json-with-moves) events])
       ::  send the start event to the client
       ::
       =^  http-moves  state
@@ -1506,13 +1506,13 @@
       ::  if conversion succeeds, we *can* send it. if the client is actually
       ::  connected, we *will* send it immediately.
       ::
-      =/  json-with-moves=(unit (pair json (list move)))
+      =/  json-with-moves=(unit (quip move json))
         (sign-to-json request-id sign)
       =/  json=(unit json)
         ?~  json-with-moves  ~
-        `p.u.json-with-moves
+        `+.u.json-with-moves
       =?  moves  ?=(^ json-with-moves)
-        (weld moves q.u.json-with-moves)
+        (weld moves -.u.json-with-moves)
       =*  sending  &(?=([%| *] state.u.channel) ?=(^ json))
       ::
       =/  next-id  next-id.u.channel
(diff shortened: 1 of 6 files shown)


Fang- (changes requested):
I imagine we'll eventually want to replace this with a broader-scope mechanism for warming the mark conversion cache, right?

For the time being I thought we would `/$` in individual apps but as per infra chat that doesn't quite work for graph-store's purposes. The approach here takes care of most cases without additional administrative burden on userspace devs, so it's a fine stopgap.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1642,17 +1642,14 @@
...
           ?.  ?=([~ ~ *] cag)  ~
-          `!<($-(* json) q.u.u.cag)
+          `q.u.u.cag
         ?~  convert
           ((slog leaf+"eyre: no convert {desc}" ~) [~ ~])
-        =/  res  (mule |.((u.convert q.cage.sign)))
-        ?:  ?=(%& -.res)
-          [`have `[%fact %json !>(p.res)]]
-        ((slog leaf+"eyre: failed tube {desc}" ~) [~ ~])
+        [`have `[%fact %json (slam u.convert q.cage.sign)]]
```
Fang-, line 1652: Hmm I thought this (the `mule`) was protecting against conversion failing due to `!!` in the implementation, not necessarily clay giving us bogus or w/e. Not sure we're fine with crashing in those cases.

belisarius222:
> I imagine we'll eventually want to replace this with a broader-scope mechanism for warming the mark conversion cache, right?

I'd prefer not to add another cache, especially since Clay already has decent caching semantics for mark conversion gates if we interact with it appropriately.

The main alternative I know of would be to remove the scry and just send a %warp to Clay each time.  That would prime Clay's cache the first time, and it would also mean there would be no more duplication between a scry and a %warp on each Eyre +take -- there would just be one %warp each time, which Clay could serve out of its cache probably even faster than handling a scry since I think there's less path parsing and serialization involved.

The downside of that approach is that Eyre would need to maintain state about what it was trying to process, since it couldn't do the work synchronously in one activation.  This isn't really a problem; it would just be a bit more work to implement.  I think long-term that's the way to go, unless someone has a better idea.

ohAitch:
If it's a caching not correctness issue, you could always do both the scry and the warp, throw away the warp result, an have the next event's scry hit cache?

tacryt-socryp:
Thanks for the style comments @Fang-! I'll get to them, I think your suggestions will make this feel way cleaner. As for whether this is a stop-gap measure, I don't think that it is. I think we want Eyre to warm all caches that it sends data through without needing to hard-code them anywhere. As for the warp overhead, it's minimal from my profiling. (1ms or so)

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: @joemfb @belisarius222 Can I have some input here? From profiling with mule vs without, using it with mule wipes out the majority of the performance gain we received from switching to `+slam`.

joemfb: We should think about whether virtualization is really necessary here. Is it better for %eyre to drop facts, or crash the event?

Regardless of the +mule, you can use `+slym` instead of `+slam` to skip nest checks (since we trust %clay).

belisarius222: It looks like the current "protection" is just a no-op, which seems like a footgun to me.  If we're going to catch errors, I'd be inclined to let the app know that its attempt to deliver an effect to the browser failed, rather than pretending everything is ok.

Also, try +slym instead of +slam, as @joemfb suggested.

belisarius222: @joemfb beat me to it :)

joemfb: For even more juice, you could make this codepath stateful and use `+slym:wa`.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1619,32 +1631,32 @@
...
-        ((slog leaf+"eyre: failed tube {desc}" ~) ~)
-      ::
+        =/  convert=(unit vase)
+          =/  cag=(unit (unit cage))
+            (rof ~ %cf [our %home da+now] /[have]/json)
+          ?.  ?=([~ ~ *] cag)  ~
+          `q.u.u.cag
+        ?~  convert
+          ((slog leaf+"eyre: no convert {desc}" ~) [~ ~])
+        [`have `[%fact %json (slym u.convert q.q.cage.sign)]]
```
joemfb, line 1653: This line should have an appropriate `~|` or `~_`, so that it's obvious if it crashes.

joemfb (changes requested):

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: done

pkg/arvo/sys/vane/eyre.hoon
Fang-: I believe crashing here will "just" result in the subscription getting killed, eyre getting the `%kick`, and proceeding from there as normal. I vaguely remember some gotcha in this area, but might have just been aggressive resubscribe behavior on the client side.  
@tacryt-socryp this is probably fine (and better than swallowing conversion errors), but worth testing the "conversion logic crashes" case to make sure nothing borks.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1665,7 +1678,7 @@
             :-  'json'
             ~|  [%unexpected-fact-mark p.cage.sign]
             ?>  =(%json p.cage.sign)
-            ;;(json q.q.cage.sign)
+            !<(json q.cage.sign)
```
Fang-, line 1681: I'm unsure what the practical different between the `;;` and `!<` here is. I'd naively expect the latter to be slower, and the former to be safe in this context? But that might be a false sense of safety.

Fang- (commented):
The comment about the no-op in `+take` still stands, as well as my question about testing the crash case in that comment thread above.  
Other than that, looks good. A small style comment and a question below.

pkg/arvo/sys/vane/eyre.hoon
Fang-: Nevermind, I think I missed this one: https://github.com/urbit/urbit/commit/306dbbaead83ca19fe95c1b60c22820f3591f6dd#r51086544

tacryt-socryp:
Thanks a ton @Fang- ! I've now made all of your requested style changes. I also tested putting a `!!` in place of the call to `+slym` to see if anything bad occurred. As expected, it crashed and printed the error saying that the conversion had failed. No other negative effects occurred, and once I changed the line back to not crash, Landscape reconnected and everything worked great once more. This seems ready to go to me.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -53,6 +56,10 @@
       $:  %gall
           gift:gall
           ::  $>(%unto gift:gall)
+          ::
+      ==
+      $:  %clay
+          gift:clay
```
joemfb, line 62: I think this should be `$>(%writ gift:clay)` -- using more specific types has a useful documentary effect, if nothing else.

joemfb (changes requested):
One nit, otherwise LGTM

joemfb (approved):
LGTM

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: I added that, but then all HTTP events crashed. Unsure how to proceed, so I commented it out next to the type as it's done in the `%gall` option directly above it.

joemfb: The `$%` in `$sign` needs to be wrapped in an appropriate `$~`, then you can use `$>` for all these entries.

joemfb: And `$note` should get the same treatment, it only happens to work currently due to bunt order.

PR #4924, opened 2021-05-20
DMs: migration + interface fixes 

liam-fitzgerald:
- Fixes sys breakage from lack of bunt
- Fixes migrations for DMs
- Drops metadata for old DMs, but not the graphs themselves

Tested on a `-L`


tacryt-socryp (approved):
LGTM

PR #4929, opened 2021-05-21
[wip] interface: add math support

liam-fitzgerald:
Adds math notation (LaTeX subset) support to GraphContent. Wide format graph content is restricted to only inline math e.g. `$x$`, where as tall format content may use display mode and environments. A demo:
Screen Shot 2021-05-21 at 2 30 31 pm
Screen Shot 2021-05-21 at 2 42 57 pm

Open questions before this gets merged:
- What does design think of this?
- Usability, unless you're a madman at latex is kinda poor for longer equations. Could potentially use a preview? Maybe if we check that it parses correctly and prevent sending otherwise. 
- Absolutely ruins our bundle size with fonts and css, partly due to the (ab)use of url-loader. Should be addressed before merge.

### DIFF
--- a/pkg/interface/config/webpack.dev.js
+++ b/pkg/interface/config/webpack.dev.js
@@ -3,18 +3,10 @@ const webpack = require('webpack');
 // const HtmlWebpackPlugin = require('html-webpack-plugin');
 // const { CleanWebpackPlugin } = require('clean-webpack-plugin');
 const urbitrc = require('./urbitrc');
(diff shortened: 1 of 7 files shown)


ohAitch:
MathJax also supports http://asciimath.org/ right?

zalberico:
@arthyn @liam-fitzgerald closing this one as stale since it looks like it's not likely to merge, let me know if I'm wrong and we can reopen.

PR #4930, opened 2021-05-23
fuse: add support for tracking sources

taalhavras:
This PR adds support for tracked sources to |fuse. When tracked sources are updated, kiln will resend a fuse request to clay. 

testing:
[x] fuses with no tracking
[x] fuses with tracked sources update
[x] fuses with tracked and untracked sources don't update when untracked sources are changed
[x] cancelling works
[x] cancelling and restarting works
[x] upgrade a running ship
[x] boot from pill

I still need to to a style/formatting pass but I think the code's worth getting some more eyes on. 

cc @philipcmonk


philipcmonk:
It would also be really nice to have a way of printing the existing fuses, and maybe even printing them as a command to copy into the dojo and modify.

philipcmonk (changes requested):
A few requested changes, but it's working great!

It would also be really nice to have a way of printing the existing fuses, and maybe even printing them as a command to copy into the dojo and modify.

You should merge master and 9bc361379c3, which updates hood to handle the new state, since kiln's state is in hood's state.

philipcmonk:
Looks like %trak is not working across the network.  Apparently %v doesn't work with %next, only %sing.  This could maybe be fixed in clay, but it's not immediately obvious how.  Maybe the easiest solution is to do a `%next %z` and then when that comes back do a `%sing %v` along with another `%next %z`.

pkg/arvo/gen/hood/fuse/help.txt
```diff
@@ -1,8 +1,21 @@
 Usage:
 
-  |fuse %destination-desk base-beak ~[[source-beak %some-germ] [another-beak %another-germ]]
+  |fuse %dest base-beak mate//~nel/home= meet//~zod/kids/track
```
philipcmonk, line 3:
```suggestion
  |fuse %dest /=kids= mate//~nel/home= meet//~zod/kids/track
```

Should show the entire command in a format where you can just change words

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -2114,7 +2114,8 @@
       ::
       =.  fiz  *melt
       ((slog [leaf+"clay: fuse failed, missing {<bec>}"]~) ..take-fuse)
-    ?>  (~(has by sto.fiz) bec)
+    ?.  (~(has by sto.fiz) bec)
+      ((slog [leaf+"clay: got strange fuse response {<bec>}"]~) ..take-fuse)
```
philipcmonk, line 2118: Here and above on 2116 and 2099, should be `<(en-beam bec ~)>` so that it prints nicer.

philipcmonk (changes requested):
Couple of small changes, but I've tested this on fake ships and it seems to work.

philipcmonk (approved):
LGTM

PR #4931, opened 2021-05-23
chat: no s3, no blank space

jamesacklin:
Gets rid of a blank space in ChatInput if S3 isn't configured.

Frame 1

fixes urbit/landscape#892


matildepark (approved):
The double ternary is confusing, but we use them a lot anyhow — I need to look into other ways of writing that for our style ... otherwise, does what it says.

PR #4933, opened 2021-05-25
interface: fix tokenisation

liam-fitzgerald:
Fixes basically all outstanding tokenisation related issue, adds jest to verify this. Enables mentions that are immediately surrounded by syntax.


matildepark (approved):
Strictly speaking an improvement but still leaves a very eager group regex -

image

PR #4934, opened 2021-05-25
interface: add more stories, integrate Chromatic CI

liam-fitzgerald:
Adds some more stories and integrates chromatic through CI


matildepark:
Looks like we need to define baselines for these components before this helps us test? Also unclear how Tlon people could log into Chromatic.

liam-fitzgerald:
The login is handled through github, if you are a member of the urbit
organisation then you should have access

——
~hastuc-dibtux

https://urbit.org

matildepark (approved):
lgtm

PR #4935, opened 2021-05-25
tokenizeMessage: revenge

matildepark:
- Patp regex is less flexible, only accepts one "any" character before a mention.
- Links regex parses _before_ group reference parser to prevent breaking references.
- Groups regex will only match on preceding spaces, there will never be a case where people are mingling it mid-word like `Here's a message~haddef-sigwen/mygroup`.
- All regex looks for `\s\S` now, not `.`, because `.` doesn't match on whitespace characters eg. line breaks, thus eating the remainder of a message that has line breaks after, well, any mention or URL.

You can give this a shot with a notebook's comment field, if you segregate links to groups, mentions, URLs with line breaks.

### DIFF
--- a/pkg/interface/src/logic/lib/tokenizeMessage.js
+++ b/pkg/interface/src/logic/lib/tokenizeMessage.js
@@ -1,11 +1,11 @@
 import urbitOb from 'urbit-ob';
 import { parsePermalink, permalinkToReference } from '~/logic/lib/permalinks';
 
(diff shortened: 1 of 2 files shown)


matildepark:
It's possible even *prefixes* should be using `[\s\S]`, you can see differences between using a newline or not here —

image

liam-fitzgerald:
These changes look normal, can you add tests to capture the previously broken cases?

liam-fitzgerald (approved):

PR #4936, opened 2021-05-26
hark: read all in group, graph

liam-fitzgerald:
- Adds %read-graph, to dismiss all unreads and notifications for a graph.
- Adds %read-group, which calls %read-graph on all the graphs in a specified group
- Binds %read-group to `Shift+Escape`
- Adds a 'Mark all as Read' button to publish

This is intended to backstop and mitigate wonky unread counts due to note deletion. An actual fix is otw, but requires more testing as it's a mark upgrade. 
Screen Shot 2021-05-26 at 2 12 40 pm


matildepark:
@liam-fitzgerald conflicts on hark-store

liam-fitzgerald:
ping @tacryt-socryp review?

pkg/arvo/app/hark-store.hoon
```diff
@@ -566,10 +570,53 @@
       (~(put by last-seen) stats-index new-time)
     (give %seen-index new-time stats-index)
   ::
+  ++  get-stats-indices
+    |=  rid=resource
+    %-  ~(gas ^in *(set stats-index:store))
```
tacryt-socryp, line 575: Needing to constantly remember to do ^in is... unfortunate. I'd consider a different variable name so as not to shadow here.

tacryt-socryp (approved):
Sane. LGTM

pkg/arvo/app/hark-store.hoon
liam-fitzgerald: yeah ik, hopefully can get to a style refactor soonish

PR #4938, opened 2021-05-26
jael, dawn: support multi-keyfile during boot

Fang-:
Allows booting with a keyfile containing multiple keys, as long as one of them matches current PKI pubkeys for the specified ship.

All relevant keys are loaded into jael and will automatically be put to use when they match PKI state.

Motivation for this is a desired L2 user flow: taking ownership of a prefabricated planet that has already had its keys set, and then setting your own. Since broadcast/confirmation of the key change may be slow, it'd be nice to boot with _both_ old and new keys, so you can get onto the network right away, while also switching to new keys automagically when the change does eventually confirm.

This change is backwards compatible with existing keyfiles and isn't strictly tied to L2 work, so I'm putting this out as a standalone PR.

`(list seed)` instead of `[=ship (list [=life =ring])]` for maximal flexibility, though this does make key validation a little bit more complicated.

Contains both arvo and runtime changes, but their release order doesn't matter. lmk what target branch would be best to use.


philipcmonk (approved):
Looks good, but don't the changes to +dawn-event mean that we need to release the runtime with a new pill?  IOW:

- existing ships can update either runtime or arvo independently
- new ships with the current pill need to be run with the current runtime
- new ships with a new pill will need to be run with the new runtime

Is that right?

Fang-:
Yeah I realized after the fact that my phrasing around release order isn't quite accurate. You're right that pill the runtime and pill need to be updated simultaneously for this.

Also realized that I need to make the Haskell king compatible with this, so, some changes pending.

Fang-:
Tagging @pilfer-pandex for the Haskell review. Note that I cherry-picked 2e2361e from #4463, since this does similar half-custom noun conversions.

joemfb:
Sorry for the delay. I have a couple requests, not sure if they're too disruptive:

I deeply regret not versioning the keyfile structure. It's a broad integration point, involves many external projects, creates long-lived artifacts, &c. I'm not sure if there's any better way than the versioned-cell pattern you've used here.

I don't like the repeated ship in this structure; we can validate that it's the same in all cases, but it'd be nicer to not need to. (This is also somewhat selfish, as I've designed the upcoming mars/urth IPC architecture to extract the identity from the dawn event).

We can decouple the injection of additional private keys from the "legacy boot event" (`%dawn`). It seems straightforward to add a new event that simply stores new life/ring pairs in %jael. That would let us preserve the existing `%dawn` type, and avoid changes to the pre-boot validation logic.

The new keyfile type would be 

```
$^  $?  [[%1 ~] who=ship kyz=(list (pair life ring))]
    ::  future versions [%2 ~], [%3 ~], &c
    ==
seed
```

%jael would get a new-task (name TBD):

```
[%prekey kyz=(list (pair life ring))]
```

And that new task could be injected during the boot sequence after %dawn, if we want to make require it. Or it could be injected after the boot sequence succeeds, if we want it to be optional.

Thoughts?

joemfb:
This is arguably a nicer type if we anticipate future changes:

```
$^  $:  [~ ~]
        $%  [%1 who=ship kyz=(list (pair life ring))]
    ==  ==
seed
```

Fang-:
I'm not opposed to changing the type/format here. The current thing admittedly got a little awkward.

> It seems straightforward to add a new event that simply stores new life/ring pairs in %jael.

`%rekey` already exists, and I think we can use that for this without issue. It does send a `%private-keys` update out to subscribers, but (at least for ames) that no-ops if the active key hasn't changed.

> We can decouple the injection of additional private keys from the "legacy boot event" (%dawn). ... That would let us preserve the existing %dawn type, and avoid changes to the pre-boot validation logic.

I think we can't get around updating the preboot validation logic, since you'll still want to make sure one of the provided keys is valid, and put that key specifically into the dawn event.

Separating it from the `%dawn` event proper seems fine to me. Perhaps even a little bit cleaner? You do have to deal with the keyfile in two places in that case (dawn and post-dawn), but we'll see if that's actually ugly in practice.

Thanks for taking a look, I'll go ahead and, uh, do this over. (^:

Fang-:
Changes are in, PTAL.

I think there's some issue with my noun conversions in King Haskell. It gives me the following during boot (after verifying the keyfile, notably), but helpfully doesn't tell me what it was trying to convert from/to. To be continued...

```
urbit-king: (["(,,,,)"],"record too short, only 3 cells")
CallStack (from HasCallStack):
  error, called at lib/Urbit/Arvo/Common.hs:309:21
```

philipcmonk:
> To be continued...

Is this working now?  Wanting to know if I should include in the next-sys release I'm putting together

Fang-:
Oh oops, no, haven't fixed the Haskell-side issue with this yet. Will pick this back up. No particular rush to get this out though!

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
```diff
@@ -117,6 +117,36 @@ data Seed = Seed
     }
   deriving (Eq, Show)
 
+data Seeds = Seeds
+    { gShip :: Ship
+    , gFeed :: [Germ]
```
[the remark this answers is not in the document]
Fang-: We can't actually do that, it clashes with the `sShip` and `sFeed` defined for `Seed`.

ryjm:
The `JaelEv` just needed some love. This should work:

image

ryjm:
There might be more changes needed on the king side - it didn't like multiple keys in the same file.

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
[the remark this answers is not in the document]
ryjm: ... `sMilt`? hehehe

Fang-:
@ryjm's a hero. Haskell side should be all good now!

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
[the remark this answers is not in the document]
pilfer-pandex: use `seed` or `se` prefix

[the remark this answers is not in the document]
pilfer-pandex: or rename Seeds -> Germs

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,67 @@ retrievePoint endpoint block ship =
...
-    Ob.Comet -> validateComet
-    Ob.Moon  -> validateMoon
-    _        -> validateRest
+validateFeedAndGetSponsor :: String
+                          -> TextBlockNum
+                          -> Feed
+                          -> RIO e (Seed, Ship)
+validateFeedAndGetSponsor endpoint block = \case
+    Feed0 s -> do
+                 r <- (validateSeed s)
```
pilfer-pandex, line 344: just indent 2 spaces from Feed0, no parens

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,67 @@ retrievePoint endpoint block ship =
...
-                    show life <> ", but Azimuth claims life " <>
-                    show netLife)
-          when ((ringToPass ring) /= pass) $
-              error "keyfile does not match blockchain"
-          -- TODO: The hoon code does a breach check, but the C code never
-          -- supplies the data necessary for it to function.
-          pure who
+    validateSeeds Seeds{..} =
+      case gFeed of
+        []    -> error ("no usable keys in keyfile")
```
pilfer-pandex, line 351: no parens

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,67 @@ retrievePoint endpoint block ship =
...
+        (Germ{..}:f) -> do
+                  let seed = Seed gShip gLife gRing Nothing
+                  r :: Either SomeException Ship
+                    <- try do validateSeed seed
+                  case r of
+                    Left _  -> validateSeeds $ Seeds gShip f
+                    Right r -> pure (seed, r)
+
+    validateSeed (Seed ship life ring oaf) =
```
pilfer-pandex, line 360: Change this to be Either String Whatever

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,67 @@ retrievePoint endpoint block ship =
...
+      case clanFromShip ship of
+        Ob.Comet -> validateComet
+        Ob.Moon  -> validateMoon
+        _        -> validateRest
+      where
+        validateComet = do
+          -- A comet address is the fingerprint of the keypair
+          let shipFromPass = cometFingerprint $ ringToPass ring
+          when (ship /= shipFromPass) $
+            error ("comet name doesn't match fingerprint " <> show ship <> " vs " <>
```
pilfer-pandex, line 370: Change to fail or Left

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,67 @@ retrievePoint endpoint block ship =
...
-          -- TODO: The hoon code does a breach check, but the C code never
-          -- supplies the data necessary for it to function.
-          pure who
+    validateSeeds Seeds{..} =
+      case gFeed of
+        []    -> error ("no usable keys in keyfile")
+        (Germ{..}:f) -> do
+                  let seed = Seed gShip gLife gRing Nothing
+                  r :: Either SomeException Ship
+                    <- try do validateSeed seed
```
pilfer-pandex, line 355: Too broad. If user does ^C you'll skip the seed.

pilfer-pandex (changes requested):

joemfb (approved):
LGTM

joemfb:
This PR should target `release/next-vere`, and include an updated ivory pill.

pkg/arvo/sys/lull.hoon
```diff
@@ -1870,7 +1870,12 @@
         [%public-keys =public-keys-result]            ::  ethereum changes
         [%turf turf=(list turf)]                      ::  domains
     ==                                                ::
-  ::  +seed: private boot parameters
+  ::  +feed: potential boot parameters
```
philipcmonk, line 1873: Is feed used?

joemfb: In `lib/vere` (and associated tests), which is exclusively used by the runtime via the ivory pill.

joemfb:
I'm merging this PR out-of-band due to the conflicts.

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
```diff
@@ -395,7 +396,12 @@ testPill pax showPil showSeq = do
   pill <- fromNounErr pillNoun & either (throwIO . uncurry ParseErr) pure
 
   logInfo "Using pill to generate boot sequence."
-  bootSeq <- genBootSeq (Ship 0) pill False (Fake (Ship 0))
+  bootSeq <- genBootSeq
```
pilfer-pandex, line 399: Questions for @joemfb re urth/mars:
- Given that the boot sequence has changed in king/serf, what (if any) change needs to be made in the urth/mars "boot message"
- does new mars implement this new boot sequence correctly?

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
```diff
@@ -541,30 +547,31 @@ newShip CLI.New{..} opts = do
           Nothing -> error "Urbit.ob didn't produce string with ~"
           Just x  -> pure x
 
-    bootFromSeed :: Pill -> Seed -> RIO HostEnv ()
-    bootFromSeed pill seed = do
-      ethReturn <- dawnVent nEthNode seed
+    bootFromSeed :: Pill -> Feed -> RIO HostEnv ()
```
pilfer-pandex, line 550: Arguably we should rename this `bootFromFeed`. Although may be moot with urth/mars

pkg/hs/urbit-king/lib/Urbit/Arvo/Event.hs
```diff
@@ -117,6 +117,36 @@ data Seed = Seed
...
+    , gRing :: Ring
+    }
+  deriving (Eq, Ord, Show)
+
+data Feed
+  = Feed0 Seed
+  | Feed1 Germs
+  deriving (Eq, Show)
+
+--NOTE  reify type environment
```
pilfer-pandex, line 137: minor nit: please put a space after `--`

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,72 @@ retrievePoint endpoint block ship =
...
+                    Left _  -> validateGerms $ Germs gShip f
+                    Right r -> pure (seed, r)
+
+    validateSeed (Seed ship life ring oaf) =
+      case clanFromShip ship of
+        Ob.Comet -> validateComet
+        Ob.Moon  -> validateMoon
+        _        -> validateRest
+      where
+        validateComet = do
```
pilfer-pandex, line 368: so it seems that both validateComet and validateMoon are both pure, and only validateRest actually needs to be in the IO monad; do I have that right?

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
```diff
@@ -335,45 +335,72 @@ retrievePoint endpoint block ship =
...
-          pure who
+    validateGerms Germs{..} =
+      case gFeed of
+        []    -> error "no usable keys in keyfile"
+        (Germ{..}:f) -> do
+                  let seed = Seed gShip gLife gRing Nothing
+                  r :: Either String Ship
+                    <- validateSeed seed
+                  case r of
+                    Left _  -> validateGerms $ Germs gShip f
```
pilfer-pandex, line 359: cool

pilfer-pandex (commented):
There are some minor points here which you should fix, as well as some questions for @joemfb that I'd like answered before I stamp.

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
Fang-: I believe so, yes.

pkg/hs/urbit-king/lib/Urbit/King/Main.hs
joemfb: The mars/urth interface actually has affordances for this kind of pattern. It should be straightforward to port these changes.

pkg/hs/urbit-king/lib/Urbit/Vere/Dawn.hs
pilfer-pandex: I was being too passive. I think it's good practice to make things as pure as possible. The only reason it'd make sense to leave something that could be pure in IO is if you anticipate needing to add more IO operations to it later.

PR #4939, opened 2021-05-26
clay + spider: make clay hand us double-vased %f and %e results, speed up spider with static conversions

tacryt-socryp:


pkg/arvo/app/spider.hoon
```diff
@@ -284,21 +285,30 @@
...
-      =tube:clay
-      %cc 
-      /(scot %p our.bowl)/[q.byk.bowl]/(scot %da now.bowl)/json/[input-mark]
-    ==
+  =/  convert
+    .^  vase
+        %cf 
+        /(scot %p our.bowl)/[q.byk.bowl]/(scot %da now.bowl)/json/[input-mark]
+      ==
+  ~|  "body cannot be null"
```
belisarius222, line 293: I'd preface these error messages with "spider: " for clarity.

pkg/arvo/app/spider.hoon
```diff
@@ -249,7 +249,8 @@
       ?+  wire  (on-arvo:def wire sign-arvo)
         [%thread @ *]  (handle-sign:sc i.t.wire t.t.wire sign-arvo)
         [%build @ ~]   (handle-build:sc i.t.wire sign-arvo)
-        [%bind ~]      `state
+        [%bind ~]              `state
+        [%conversion-cache *]  `state
```
belisarius222, line 253: this looks unused

belisarius222 (approved):
left a couple nits, but looks good otherwise

pkg/arvo/app/spider.hoon
tacryt-socryp: The conversion-cache? Yes, we throw away the result, but we no-op so as to avoid a crash message in the default case of `?+`

belisarius222: nevermind, I misread this

tacryt-socryp:
Disappointing performance results, more or less the same.

PR #4940, opened 2021-05-26
Bitcoin Wallet

ixv:
Here is the bitcoin wallet

Beyond being just a bitcoin wallet, its also Tlon's first attempt at distributing multiple frontend packages. Includes 2 new hoon apps, a new pkg/btc-wallet, some changes to pkg/interface to make custom tiles work with images, and changes to glob to make it work with multiple packages.

Thanks to @timlucmiptev for writing the bulk of the backend code, and thanks to @pkova for contributing to both frontend and backend.

This is still somewhat of an alpha, so we are releasing this in an unstarted state. People who feel savvy enough to use this can `|start %btc-wallet`


liam-fitzgerald:
I'd argue that this can and should go out to next-userspace before hitting master

matildepark:
> I'd argue that this can and should go out to next-userspace before hitting master

Idk if it can with the glob changes — janeway isn't updated to handle the new glob logic. But I like the intention.

tacryt-socryp:
Could the glob changes be separately targeted to hit master?

tacryt-socryp (approved):
Approving the `%glob` and `%file-server` changes.

liam-fitzgerald (approved):
(I'm only reviewing the build process and integration with the rest of landscape)
Discussed OOB, everything LGTM!

jtobin:
> > I'd argue that this can and should go out to next-userspace before hitting master
> 
> Idk if it can with the glob changes — janeway isn't updated to handle the new glob logic. But I like the intention.

Anything I need to do here?  Can't remember if I was supposed to update something and forgot.. 😝

jtobin:
> changes to glob to make it work with multiple packages.

(ah, nm)

liam-fitzgerald:
I've done the glob changes, PRing now

jalehman (approved):

PR #4942, opened 2021-05-27
dill: support next runtime's inputs

Fang-:
Quick and dirty forward-compatibility by pulling the new belts (from #4463) into the current dill. Intent is to send this out prior to the new dill/runtime combo as an appetizer OTA.

This, along with the new runtime doing best-effort handling of current dill's blits, will ensure that the new runtime is reasonably usable even with the old dill.

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1070,15 +1070,15 @@
     ::
   +$  blew  [p=@ud q=@ud]                               ::  columns rows
   +$  belt                                              ::  client input
-    $%  bolt                                            ::  simple input
-        [%mod mod=?(%ctl %met %hyp) key=bolt]           ::  w/ modifier
+    $?  bolt                                            ::  simple input
+    $%  [%mod mod=?(%ctl %met %hyp) key=bolt]           ::  w/ modifier
         [%txt p=(list @c)]                              ::  utf32 text
         ::TODO  consider moving %hey, %rez, %yow here   ::
-        ::TMP  forward backwards-compatibility
-        ::
-        [%ctl p=@c]
-        [%met p=@c]
-    ==                                                  ::
+        ::TMP  forward backwards-compatibility          ::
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/lull.hoon
```diff
@@ -1069,14 +1069,23 @@
   ::::                                                  ::  (1d2)
     ::
   +$  blew  [p=@ud q=@ud]                               ::  columns rows
-  +$  belt                                              ::  old belt
+  +$  belt                                              ::  client input
+    $%  bolt                                            ::  simple input
```
Fang-, line 1073: Curiously, `$?` works fine in #4463, but gives `fish-loop` here. `$%` works somewhat unexpectedly, considering the `@c` case in `bolt`. Probably benign shenanigans afoot.

Fang-: ...and it turns out this actually explodes over as soon as you try to actually use a `@c` `bolt`.

philipcmonk: The fish-loop trace points right here?

Fang-: The entire `$?` (up to the `==`), yes.

pkg/arvo/sys/vane/dill.hoon
```diff
@@ -116,6 +115,18 @@
...
           %verb  (pass /verb %$ kyz)
+        ::
+            %belt
+          %-  send
+          ::TMP  forwards compatibility with next-dill
+          ::
+          ?@  p.kyz  [%txt p.kyz ~]
+          ?:  ?=(%hit -.p.kyz)  [%txt ~]
+          ?.  ?=(%mod -.p.kyz)  p.kyz
+          =/  =@c
```
philipcmonk, line 126: big fan of exotic autoname

philipcmonk (approved):
LGTM

PR #4944, opened 2021-05-27
arvo: add %whiz option to  that allows one to prime the compiler cache

tacryt-socryp:
This PR saves us an additional 15-20ms per event by reducing arvo scry overhead.

Screen Shot 2021-05-27 at 1 26 28 PM
Screen Shot 2021-05-27 at 1 26 46 PM


joemfb (approved):
LGTM

I think we might want to include vane labels/abbreviations in the card and be able to prime them individually. But that could be added later, with backwards compatibility (`[%whiz @tas]`, where `%$` has the current behavior).

belisarius222 (approved):
LGTM.  I think Ames should also send a %whiz after it exits the larval stage, but that could be added in a later PR.

PR #4947, opened 2021-05-27
btc: fix warning link style

ixv:


matildepark (approved):
Anchor doesn't inherit our theme colours? Well okay.

Issue #4941, opened 2021-05-26
http-api: Can't be used from NPM due to local dependency


tylershuster:
See this line:

https://github.com/urbit/urbit/blob/61a3ea8028ff18a2f4bf5ed6618b921a8e148bab/pkg/npm/http-api/package.json#L55

This allows us to do development of the Urbit API's Javascript representation alongside its core by requiring the local version of the API. Unfortunately this means that when the package is deployed to NPM, it can't find the correct package. @jtobin any ideas on how to handle this in CI?

jtobin:
@tylershuster I think this might be able to be handled by adding the local package to [`bundledDependencies`](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#bundleddependencies) in package.json.  That's [used](https://github.com/urbit/azimuth/blob/master/package.json#L21) in azimuth-solidity to bundle openzeppelin.

PR #4949, opened 2021-05-27
npm/http-api: changed relative dependency

tylershuster:
Targeting master so it gets pushed to npm faster. LMK if there's a problem with that. Fixes https://github.com/urbit/urbit/issues/4941


matildepark (approved):
lgtm

PR #4960, opened 2021-05-28
interface: catch contact-update type mismatch

matildepark:
When applying add-group or remove-group updates optimistically (https://github.com/urbit/urbit/commit/a11a1e42ce402e0433a63ec45ee8b964867d38d5), we send
our ship resourceFromPath objects, and get back `/ship/~name/resource`
strings. When we apply the update optimistically, we use the former,
storing the object in state. This causes interface errors, as it expects
the string. Therefore, on both cases in the reducer we construct the string and 
dedupe if it comes in after application.

Fixes urbit/landscape#917


liam-fitzgerald (approved):
LGTM, cause of mismatch should be fixed in engineering sprint

PR #4962, opened 2021-05-28
leap: render custom app tiles

matildepark:
Fixes urbit/landscape#915

image

- Iterates through all apps regardless of tile type when indexing; instead just filters the two tiles we don't want in the index
- Provides appropriate fallback values for custom tile cases

Leap result logic is getting a little long in the tooth, but will probably amend its index behaviour quite soon anyhow. BTC Wallet's entry has no title; it feels weirder to special case it in FE, probably should add a title to it on back-end?


ixv (approved):
LGTM

Issue #4959, opened 2021-05-28
Sending malformed dojo requests via herb breaks the ability to send further commands over herb


baudtack:
**Describe the bug**
Sending malformed dojo commands via herb results in eyre always responding with 500 to any subsequent request.

**To Reproduce**
Steps to reproduce the behaviour:
1. `herb -d  "(add 2)"
2. Herb will hang, dojo will show a syntax error
3. Any subsequent request gets  
```
<html><head><title>500 Internal Server Error</title></head><body><h1>Internal Server Error</h1><p>There was an error while handling the request for /.</p><code>/app/lens/hoon:&lt;[59 3].[118 5]&gt;<br />/app/lens/hoon:&lt;[58 3].[118 5]&gt;<br />/app/lens/hoon:&lt;[56 3].[118 5]&gt;<br />/app/lens/hoon:&lt;[52 3].[118 5]&gt;<br />/app/lens/hoon:&lt;[50 3].[118 5]&gt;<br />/sys/vane/gall/hoon:&lt;[1.370 9].[1.370 37]&gt;<br /></code></body></html>
```
**Expected behaviour**
Herb should get a response on bad syntax and not break.

ixv:
I'm pretty sure you can do `:lens %cancel` to fix it.

ixv:
I guess that doesn't help if you're doing this to a ship you have no normal dojo access to though

baudtack:
It's good to know for sure, but yeah in this case I'm struggling with ships I don't otherwise have access to.

jfranklin9000:
related: https://github.com/urbit/urbit/pull/1888

PR #4963, opened 2021-05-29
herb: add cancel command

ixv:
Lets you run `herb --cancel` to cancel any ongoing command. Addresses this issue https://github.com/urbit/urbit/issues/4959


baudtack (approved):
You're my hero.

PR #4964, opened 2021-06-01
messages: better group DM titles

jamesacklin:
When creating a group DM, uses the invited ships to name the resource, then conditionally displays either a patp (if set) or a nickname (if allowed by CalmEngine) for each ship in the DM title. Changes appear in the DM sidebar and the current DM ResourceSkeleton header.

Also fixes a few inconsistencies:
- Respects hidden-nicknames in 1:1 DM titles
- Consistently displays papts as mono, all nicknames as Inter
- Actually surfaces the DM description if set, falls back to monospaced participants (either single or list)

Does not set the group DM to a random color (as specified in urbit/landscape#870) because we don't handle setting metadata before the channel is created; hoping to tackle this with @liam-fitzgerald.

With nicknames turned on: 
image

With nicknames turned off:
image

With description set:
image

1:1 DM with nicknames turned on: 
image

1:1 DM with nicknames turned off:
image

1:1 DM with no nickname set (might be stupid, but is at least consistent):
image

Fixes urbit/landscape#623


liam-fitzgerald (approved):
LGTM!

liam-fitzgerald:
Can you change target to release/next-userspace?

PR #4965, opened 2021-06-01
imageinput: disable placeholder selection, hide placeholder on focus

jamesacklin:
Default:
image

Click into field:
image

Click "upload" blue text:
image
(will vary from OS to OS)

Successful upload:
image

Failed upload:
image

fixes urbit/landscape#893

### DIFF
--- a/pkg/interface/src/views/components/ImageInput.tsx
+++ b/pkg/interface/src/views/components/ImageInput.tsx
@@ -34,7 +34,7 @@ const prompt = (
         fontWeight='500'
         position='absolute'
         left={2}
-        top='6px'
+        top='6.5px'
         style={{ pointerEvents: 'none' }}
         onSelect={e => e.preventDefault}
       >
@@ -66,7 +66,7 @@ const uploadingStatus = (uploading, meta) => {
       <Text
         position="absolute"
         left={2}
-        top={2}
+        top='6.5px'
         gray
         onSelect={e => e.preventDefault}
       >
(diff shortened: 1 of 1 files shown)


liam-fitzgerald:
@urcades chromatic caught some spacing changes, I've assigned you for review

matildepark:
It shouldn’t show the upload if s3 isn’t set up. Have you fixed that?

urcades:
> @urcades chromatic caught some spacing changes, I've assigned you for review

I'm still getting the hang of how chromatic interacts with existing PRs/issues/comments, but figured this might be useful to post here, to this issue:

Screen Shot 2021-06-01 at 1 51 09 AM

These are the two branches in the UI which appear to be 1px too high and low, as noted in chromatic. We basically should be editing this interface such that the text is perfectly centered in this 32px high container.

What I wrote in chromatic:

> There's a funny bit of weirdness in this diff as the present UI in release/next-userspace seems like it's 1px too low, and james' fix in /imageinput-fixes seems 1px too high.
> 
> The container/field itself appears to be 32px high, so this is signaling that we have some weirdness going on in how this text is being styled.

jamesacklin:
> It shouldn’t show the upload if s3 isn’t set up. Have you fixed that?

I didn't explicitly see it in scope for that issue, but I just fixed it. I realize it may have been a bit hidden behind the concern around actually showing the upload dialog.

@urcades I'm happy to take another look at this; I had bumped the text alignment _up_ by 1px because it seemed too low; let me play with it a bit more and I'll review with you.

jamesacklin:
As reviewed and stamped spiritually:

image

jamesacklin:
> Is it possible this is a line height issue? Like flattening it to `lineHeight={1}` or something would prevent subpixel wrangling?

It's a line-height issue within Inter itself, not necessarily a bounding-box mathematical one, but I made some changes such that we don't have to think about it as hard.

matildepark (approved):

PR #4966, opened 2021-06-01
interface: fix issues with dm-hook release

liam-fitzgerald:
After thorough testing, it would appear we cannot just kick subs and cleanup after ourselves immediately, due to potential latency in the DM counterparty processing the OTA. Instead, we set the hidden flag on metadata, with the intention of cleaning up the stale state at a later date instead. We could potentially cleanup old DMs as we establish their new counterparts? Also includes a merge of master, so that release is up to date.


tacryt-socryp:
Could you merge master into the release branch then retarget this to it? It will make the diff easier to read.

tacryt-socryp (approved):
Only reviewing the changes you've made since merging in master. LGTM.

PR #4970, opened 2021-06-01
graph-push-hook: disable proxying of DM updates

liam-fitzgerald:
Prevents messages from being dropped during the OTA, when there's a mismatch in deployed versions between host and subscriber.


tacryt-socryp (approved):
LGTM

PR #4973, opened 2021-06-02
interface: deploy omnibus

liam-fitzgerald:
Anything I'm missing?

### DIFF
--- a/pkg/btc-wallet/src/index.js
+++ b/pkg/btc-wallet/src/index.js
@@ -8,7 +8,7 @@ import './css/indigo-static.css';
 import './css/fonts.css';
 import './css/custom.css';
 
-// rebuild x2
+// rebuild x3
 
 window.NETWORK = 'testnet'; // 'bitcoin'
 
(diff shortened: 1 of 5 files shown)


matildepark:
I believe the sidebar colours were more graduated before our merge conflict. You may need to look at the repo state before we resolved and see how we did that? I’m unsure how it regressed.

matildepark:
Also your rebuild commit’s contents are already on next userspace so it’ll choke the glob build.

pkg/btc-wallet/src/index.js
```diff
@@ -8,7 +8,7 @@ import './css/indigo-static.css';
 import './css/fonts.css';
 import './css/custom.css';
 
-// rebuild x2
+// rebuild x3
```
liam-fitzgerald, line 11: @matildepark it's changed, even if it doesn't make it to the built JS, it'll end up in the source map, which means the glob hash will be new.

matildepark (approved):

PR #4976, opened 2021-06-03
messages: synced/unread treatments for group DMs in sidebar

jamesacklin:
Fixes this confusing non-state:
image

Now reflective of unread/unsynced/hasNotification states (shown below with nicknames turned on and the DM at rest).

image

fixes urbit/landscape#623

### DIFF
--- a/pkg/interface/src/views/landscape/components/Sidebar/SidebarItem.tsx
+++ b/pkg/interface/src/views/landscape/components/Sidebar/SidebarItem.tsx
@@ -1,10 +1,8 @@
 import _ from 'lodash';
 import React, { useRef, ReactNode } from 'react';
 import urbitOb from 'urbit-ob';
-
 import { Icon, Row, Box, Text, BaseImage } from '@tlon/indigo-react';
 import { Association, cite } from '@urbit/api';
-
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/landscape/components/Sidebar/SidebarItem.tsx
```diff
@@ -199,22 +191,25 @@ export function SidebarAssociationItem(props: {
     return null;
   }
 
-  const participantNames = (str: string, color: string) => {
+  const participantNames = (str: string) => {
+    const color = isSynced ? (hasUnread || hasNotification) ? 'black' : 'gray' : 'lightGray';
```
matildepark, line 195: I don't think the linter allows for nested ternaries like this?

jamesacklin: Doesn't complain about it here: https://github.com/urbit/urbit/blob/460982d74b72d510bc606477c27764cf7fbc7159/pkg/interface/src/views/landscape/components/Sidebar/SidebarItem.tsx#L39

matildepark (approved):

PR #4977, opened 2021-06-03
graph-push-hook/thread: shave off 20ms from nest checks by getting ri…

tacryt-socryp:
…d of tmi problem with types


philipcmonk (approved):
Looks good, though you could do them in irregular form

For future reference, the `?>`s above specialize the type so that it's no longer exactly equal to what we're expecting on the other side.  The cast makes them exactly equal.

PR #4978, opened 2021-06-03
graph-store + validators: stop using +grab:noun and switch to using a static conversion to %graph-indexed-post

tacryt-socryp:
@belisarius222  recommended I do this recently, and I tested it and it is indeed significantly faster (by about ~20ms for a single chat / publish / link / feed message). Tested to work on Chat, Publish, Links, and Feed.

### DIFF
--- a/pkg/arvo/app/graph-store.hoon
+++ b/pkg/arvo/app/graph-store.hoon
@@ -19,7 +19,7 @@
 +$  debug-input  [%validate-graph =resource:store]
 ::
 +$  cache
-  $:  validators=(map mark dais:clay)
+  $:  validators=(map mark $-(indexed-post:store indexed-post:store))
   ==
 ::
 ::  TODO: come back to this and potentially use ford runes or otherwise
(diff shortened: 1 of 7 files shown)


pkg/arvo/app/graph-store.hoon
```diff
@@ -627,22 +627,27 @@
...
-    :_  state(validators (~(put by validators) u.mark dais))
+    :_  state(validators (~(put by validators) u.mark validate))
     |-  ^-  ?
     ?~  graph  %.y
     %+  roll  (tap:orm graph)
     |=  [[=atom =node:store] out=?]
     ^-  ?
     ?&  ?|  ?=(%| -.post.node)
-            ?=(^ (vale:dais [atom p.post.node]))
+            ?=(^ (validate [atom p.post.node]))
```
philipcmonk, line 650: when could this be an atom?

tacryt-socryp: It cannot be. This is a trivially true assertion, since any of these validators will crash if given invalid input.

tacryt-socryp:
What do you think @liam-fitzgerald ?

pkg/arvo/app/graph-store.hoon
```diff
@@ -627,22 +627,27 @@
     |=  [=graph:store mark=(unit mark:store)]
     ^-  [? _state]
     ?~  mark   [%.y state]
-    =/  has-dais  (~(has by validators) u.mark)
-    =/  =dais:clay
-      ?:  has-dais
+    =/  has-validator  (~(has by validators) u.mark)
+    =/  validate=$-(indexed-post:store indexed-post:store)
+      ?:  has-validator
```
liam-fitzgerald, line 632: We have to use `?:` instead of `fall` because `fall`s arguments will be evaluated eagerly?

liam-fitzgerald (approved):
Looks great. My general concern is the semantics of graph marks are becoming increasingly messy, but I guess it's the price we pay for hacking in typeclasses.

pkg/arvo/app/graph-store.hoon
tacryt-socryp: I'm doing a style pass separately (and all at once) as a part of this sprint. Not sure if they're eagerly evaluated! Will evaluate using that then

tacryt-socryp:
Just need to keep making the documentation better I guess! I'll do a graph docs update to make sure it's up to date as well during this sprint.

ohAitch:
Surely the primary cost of hacking in typeclasses is having typeclassess.

tacryt-socryp:
We could potentially try to add some sort of constraint where we check whether the relevant arm (that converts indexed-post to indexed-post and is called `+graph-indexed-post` inside of `+grab`) is present when someone adds a graph that has a new validator mark. I imagine this is possible, though it might be some sort of crazy `+nest` partial reimplementation.

pkg/arvo/app/graph-store.hoon
joemfb: If you need a lazily-evaluated default, use `+bond`.

PR #4980, opened 2021-06-03
graph-push-hook: speed up by getting rid of nest checks, using %cf, a…

tacryt-socryp:
…nd storing gates instead of tubes

We now spend only 4ms (!!!) in `%graph-push-hook`.


liam-fitzgerald (approved):
LGTM!

PR #4983, opened 2021-06-04
leap: display chording options

liam-fitzgerald:
Screen Shot 2021-06-04 at 12 24 52 pm

### DIFF
--- a/pkg/interface/src/logic/lib/omnibox.js
+++ b/pkg/interface/src/logic/lib/omnibox.js
@@ -11,19 +11,23 @@ const makeIndexes = () => new Map([
   ]);
 
 // result schematic
-const result = function(title, link, app, host) {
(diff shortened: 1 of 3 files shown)


tacryt-socryp:
How does this look / function on mobile? Is it hidden?

tacryt-socryp:
Dope

matildepark:
You gotta fix tsc. I pinged Ed for design look

matildepark (approved):
This hard restricts us to shift chords, but is a bit more flexible than my idea ("pairing all applicable chords to search categories"). We really need to refactor Leap, but it may just end up being rewritten from scratch whenever we do the next iteration of it — it'll have different product requirements entirely.

matildepark:
Merging because this mock variation was given by Ed OOB, it seems.

PR #4986, opened 2021-06-04
hoon: add +sink, a fast(er than dunk) type printer

tacryt-socryp:


ohAitch:
As opposed to dunk?

ohAitch:
Also surely this shouldn't be _unconditional_, if it's being proposed for merge into next-sys?

tacryt-socryp:
I need to comment out the slogs, yeah. I was just PRing what I had.

philipcmonk:
Yes, sink is meant as a fast and short (non-recursive) alternative to dunk, suitable for tracing through +nest.  The original use was for determining why two types that we thought should be equal were in fact non-equal.  The trace with all the mugs makes it clear which branch is non-equal at any level.

I actually think this has enough general applicability that we should put it at the top level of +ut next to +dunk and with a similar call pattern.  So instead of accepting a type, it operates on `sut`, and you call `sink(sut ref)` to print `ref`.  From the dojo, then you could run `~(sink ut %noun)`.  The next level would be to make it produce complex tanks instead of tapes, but I don't think that's important.

tacryt-socryp:
@philipcmonk done

philipcmonk (approved):
Looks good.

Making `mup` be `(scot %p (mug +<))` is probably a little cleaner.

tacryt-socryp:
Good feedback. Implemented. Will merge once tests pass.

pkg/arvo/sys/hoon.hoon
```diff
@@ -8764,6 +8764,7 @@
       %peek   peek
       %repo   repo
       %rest   rest
+      %sink   sink
```
joemfb, line 8767: For future reference, this hook should not have been added in this PR. It changes the jet-matching interface (per hashboard at least), and so requires staging on both sides.

PR #4987, opened 2021-06-04
zuse: comb through all notes in libraries, and move arms to zuse

tacryt-socryp:
We've had notes to move various parsing functions to zuse for literally years, so I figured I'd go ahead and do it.

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -3391,6 +3391,11 @@
     ::                                                  ::  ++no:dejs:format
     ++  no                                              ::  number as cord
       |=(jon=json ?>(?=([%n *] jon) p.jon))
+    ::                                                  ::  ++nu:dejs:format
+    ++  nu                                              ::  parse number as hex
+      |=  jon=json
+      ?>  ?=([%s *] jon)
+      (rash p.jon hex)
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/zuse.hoon
```diff
@@ -3653,6 +3678,13 @@
       |*  [pre=* wit=fist]
       (cu |*(* [pre +<]) wit)
     ::
+    ++  re                                              ::  recursive reparsers
+      |*  [gar=* sef=_|.(fist)]
+      |=  jon=json
+      ^-  (unit _gar)
+      =-  ~!  gar  ~!  (need -)  -
```
Fang-, line 3685: Probably just nuke this line.

Fang- (commented):
Neat! Can you replace the inline implementations you found, with calls to their zuse implementations? Might also be worth searching the codebase more broadly for similar implementations, but probably pretty hard to catch those. We can close those up as we run into them.

pkg/arvo/sys/zuse.hoon
tacryt-socryp: done

Fang- (approved):
As discussed out of band, under the assumption that removal of their previous location will follow soon, approved!

tacryt-socryp:
yay thanks!

PR #4990, opened 2021-06-07
SidebarList: fix shortcuts

liam-fitzgerald:
The SidebarList refactor that occurred to accomodate new-style DMs
inadvertently broke the keyboard shortcuts to switch between channels.
Revives this, and adds support for switching between DMs as well.

### DIFF
--- a/pkg/interface/src/views/landscape/components/Sidebar/SidebarList.tsx
+++ b/pkg/interface/src/views/landscape/components/Sidebar/SidebarList.tsx
@@ -108,14 +108,19 @@ export function SidebarList(props: {
     const offset = backward ? -1 : 1;
 
     const newIdx = modulo(idx+offset, ordered.length - 1);
-    const { metadata, resource } = associations[ordered[newIdx]];
-    const joined = graphKeys.has(resource.slice(7));
-    let path = '/~landscape/home';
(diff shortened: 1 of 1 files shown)


tacryt-socryp (approved):
LGTM, but what does putting history.push in the dependencies of that hook even do?

jamesacklin (approved):

jamesacklin:
Also interested in why, but only academically.

liam-fitzgerald:
ok, imagine useHistory was defined like this (yes i know that the browser keeps it's own history stack, it's purely illustrative)

```typescript
function useHistory() {
  const [stack, setStack] = useState([] as string[]); // history stack

  const push = useCallback((url: string) => {
    setStack([...stack,  url]);
    window.history.pushState(url);

  
  }, [stack]);

   return { push }
}
```
This means that the value of push changes everytime stack changes. Then if we omit the dependency, we can end up with stale props
```typescript
const navigate = useCallback(() => {
  history.push(url);
}, [url]); 
```
If the value of `stack` in useHistory hook changes, then the value of push will change, but the reference to push inside navigate will not be updated, thus the stack will be reset to it's value on the first render, when the reference to `history.push` was captured.

PR #4991, opened 2021-06-07
graph-store: clean up old migrations, use some of the new +on methods

tacryt-socryp:


pkg/arvo/app/graph-store.hoon
```diff
@@ -5,9 +5,9 @@
 |%
 +$  card  card:agent:gall
 +$  versioned-state
-  $%  [%0 network:zero:store]
-      [%1 network:zero:store]
-      [%2 network:zero:store]
+  $%  [%0 *]
+      [%1 *]
+      [%2 *]
```
liam-fitzgerald, line 10: These are all pre breach, yeah?

pkg/arvo/mar/graph/update-0.hoon
```diff
@@ -1,18 +0,0 @@
-/+  *graph-store
```
liam-fitzgerald, line 1: These marks are definitely not pre breach

liam-fitzgerald (commented):
Some questions about migration and perf regressions. Else looks good so far

pkg/arvo/app/graph-push-hook.hoon
```diff
@@ -316,28 +314,27 @@
...
+      (~(get by graph-to-mark.cache) resource)
     (get-mark:gra resource)
   ?~  mark
     [[%no %no %no] ~]
   =/  key  [u.mark (perm-mark-name perm)]
-  =/  perms-cached  (~(has by perm-marks.cache) key)
   =/  convert
-    ?:  perms-cached
-      (~(got by perm-marks.cache) key)
+    %+  fall
```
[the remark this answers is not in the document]
tacryt-socryp: I looked at the code. b is not evaluated unless a is sig.

pkg/arvo/app/graph-store.hoon
tacryt-socryp: %2 is not, but %0 and %1 are. Would you like me to add back %2?

liam-fitzgerald:
Yeah please, I'm hesitant to drop support for anything that's not pre-breach

——
~hastuc-dibtux

https://urbit.org

tacryt-socryp:
@liam-fitzgerald Only removed `%0` and `%1` migrations now. Reverted the controversial bits. This should be good to go.

liam-fitzgerald (approved):

PR #4993, opened 2021-06-07
groups: prevents group names starting with digits

jamesacklin:
Implements some very simple form validation preventing group names that begin with digits. Emoji and symbols are still permitted.

fixes urbit/landscape#906

### DIFF
--- a/pkg/interface/src/views/landscape/components/NewGroup.tsx
+++ b/pkg/interface/src/views/landscape/components/NewGroup.tsx
@@ -17,7 +17,7 @@ import { AsyncButton } from '~/views/components/AsyncButton';
 
 const formSchema = Yup.object({
   title: Yup.string()
-    .matches(/^(\D).+$/, 'Group names may not start with a number')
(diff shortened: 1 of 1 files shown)


pkg/interface/src/views/landscape/components/NewGroup.tsx
```diff
@@ -16,7 +16,9 @@ import useMetadataState from '~/logic/state/metadata';
 import { AsyncButton } from '~/views/components/AsyncButton';
 
 const formSchema = Yup.object({
-  title: Yup.string().required('Group must have a name'),
+  title: Yup.string()
+    .matches(/^(\D).+$/, 'Group names may not start with a number')
```
matildepark, line 20: The `+` should really be a `*` I think — single character group names should be allowed, and not error confusingly.

image

matildepark (changes requested):

matildepark (changes requested):
`%` leaves a group spinning forever.

image

Maybe `\W` or whatever word is?

matildepark (approved):

PR #4995, opened 2021-06-07
interface: consistent  titles for mentions, authors with nicknames

jamesacklin:
My understanding of urbit/landscape#65 is that the tooltip over any sort of ship text (mention, author, chat byline) should display the inverse of whatever logic is showing it.

- If I have nicknames set to show, I should naturally see ships' nicknames (if set), and see their ~patps on hover
- If I have nicknames set to not show, I should naturally see ships' ~patps, and their nicknames on hover (if set)
- Regardless of my settings, if others have not set nicknames, they should not show and I should naturally see their ~patps with no hover text otherwise.

I may have misinterpreted the issue, since this seems inconsistent with our rule about never showing nicknames if the pilot truly does not want to see them. It's an easy enough fix to back out the second behavior above, but just need some clarity.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatMessage.tsx
@@ -159,7 +159,7 @@ export const MessageAuthor = ({
             fontWeight={nameMono ? '400' : '500'}
             cursor='pointer'
             onClick={doCopy}
-            title={`~${msg.author}`}
+            title={showNickname ? `~${msg.author}` : contact?.nickname}
           >
             {copyDisplay}
           </Text>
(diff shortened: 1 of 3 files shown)


Fang-:
> It's an easy enough fix to back out the second behavior above

The second case is exactly what I wanted though! Thank you. (:

The third case in that list feels a _little_ bit weird, in that it's hard (in the "nicks disabled" case) to tell the difference between "no nick set" and "tooltip not showing for user error reasons" or w/e, but I doubt that's a big deal in practice.

liam-fitzgerald (approved):

jamesacklin:
Awesome, thank you for the input @Fang-. I will fix the TSC error.

liam-fitzgerald:
bump @jamesacklin

jamesacklin:
All good, just had to re-run the checks.

matildepark (approved):

PR #4997, opened 2021-06-08
tokenizeMessage: fix urls that end in punctuation

liam-fitzgerald:
- Also, respect calmengine settings in Messages sidebar (hideNicknames, hideAvatars)


tacryt-socryp (approved):
LGTM

tacryt-socryp:
Could we remove periods and commas from being included in this?

PR #4998, opened 2021-06-08
interface: hook-based api

liam-fitzgerald:
### Architecture

We define a singleton `Airlock` instance (from @urbit/http-api npm package) and simply import it where it needs to be used. Stores define their own subscriptions, allowing far better modularity. Scries must be done as functions inside the store, however pokes and subscriptions live entirely outside of the store, allowing for better performance due to improved memoisation. Seeking feedback on the general design of this, as it has pretty far reaching implications.

This cannot be merged until we totally deprecate the old api/subscription pattern, as we cannot have two seperate channels open without running into networking issues. Changed functionality appears to work perfectly other than this. Also upgrades types, to bring the `@urbit/api` package inline with livenet datastructures.

Stores completed:
- [x] Hark
- [x] Graph
- [x] Groups
- [x] Metadata
- [x] Contacts
- [x] Settings


tylershuster:
This is starting to look good. You've solved some problems that I ran into with the pattern of putting subscriptions directly on the state objects. I'll review further as things get added.

liam-fitzgerald:
Status update: largely done!

Still todo:
- Optimistic store updating
- Display subscription status
- Investigate subscription reconnect behaviour

liam-fitzgerald:
This is ready to go! Optimistic state updates are back, and we're tracking/displaying the current subscription connection status.

matildepark:
Needs to pass tsc.

matildepark:
> This cannot be merged until we totally deprecate the old api/subscription pattern, as we cannot have two seperate channels open without running into networking issues.

Is this still the case? It seems clear from my breezing through it.

matildepark:
Also curious about the net +5,162 LoC from a refactor. Is there something still left to clean up here? It all looks like what I would expect from this migration.

liam-fitzgerald:
> Also curious about the net +5,162 LoC from a refactor. Is there something still left to clean up here? It all looks like what I would expect from this migration.

Just did a final sweep of remaining cruft, think that's it. Some of the LoC difference is in package-locks because I had to tweak the build process a little to get things working. Also endeavoured to write descriptive doccoments for the `@urbit/api` package, so would be contributing to diff size also.

liam-fitzgerald:
CI failures seem spurious? They're both complaining about `rest` not being defined in the `Mention` component, but there's no reference to `rest` inside that component at all?

tacryt-socryp:
This seems good to go to me. Once we've done tonight's release, I'd like to merge this into next-userspace.

matildepark (approved):

PR #5000, opened 2021-06-09
king: don't kill on C-d

ryjm:
Have mercy.

### DIFF
--- a/pkg/hs/urbit-king/lib/Urbit/Vere/Term.hs
+++ b/pkg/hs/urbit-king/lib/Urbit/Vere/Term.hs
@@ -566,7 +566,6 @@ localClient doneSignal = fst <$> mkRAcquire start stop
                 loop rd
               else if w <= 26 then do
                 case pack [BS.w2c (w + 97 - 1)] of
-                    "d" -> atomically doneSignal
                     c   -> do sendBelt $ Ctl $ Cord c
                               loop rd
               else if w == 27 then do


pilfer-pandex:
Why is this desirable?

ohAitch:
Yeah isn't this kinda the standard way to quit urbit on the C runtime? Or does "kill" mean something weird here

ryjm:
C-d should only kill when the line is blank, otherwise it's delete forward
one character. I use C-d all the time so I would accidentally quit dojo all
the time.

joemfb:
There's no reason to implement the behavior at this level, exiting is implemented inside `lib/hood/drum`. (Also, the current implementation is wrong...)

ryjm:
Yeah after deleting this line the behavior is correct - it quits on C-d only when the dojo is blank.

pilfer-pandex:
Okay in that case I do want to push back against the proposed behavior change. I don't like that my line has to be empty in order for me to quit. I know that this is a difference from vere, but I prefer our way better. A large man once said that "wrong" is always and everywhere a matter of taste.

Bikeshedding: Arguably, ^D also sucks in that if you press it twice (for example, because there is lag and you forgot that you pressed it), you wind up losing your shell. This happens to me on a regular basis. The best of all worlds would then be to come up with a different control character which would take the place of ^D in the haskell line which @ryjm proposes to delete. Then ^D can proceed to work only on empty lines, "deleting" otherwise, and everyone can be happy. Is there a standard alternative character for this? I don't know one off of the top of my head.

ohAitch:
`^Z` and `^\` should both quit instantly regardless of line state right? (The latter recording a core dump for inspection)

pilfer-pandex:
[You're wrong.](https://www.youtube.com/watch?v=GM-e46xdcUo) I just tried it on macOS and:
- ^Z did nothing
- ^\ caused the terminal/king to become completely unresponsive

philipcmonk:
Well those two should be fixed to work like in vere, they're the normal ways to exit without going through event loop.  Note that they're handled as SIGTSTP and SIGQUIT rather than keystrokes directly.  ^D has the same "only exit if the line is empty" behavior in bash, fish, etc, so it'd be nice to preserve that behavior.

ohAitch:
Can one really be "right" or "wrong" on matters of "should"? ;)
(But yes, I was referring to the reference implementation behavior)

ohAitch:
I suppose I *can* confidently claim "caused the terminal/king to become completely unresponsive" is not what any button should do, UI considerations aside. (`^S` maybe? I think that one's built in.)
Is this a Haskell core dump difference, something about the connection to the worker and/or terminal not being cleaned up, some performance edge case?

joemfb:
Merging to ensure this is not lost.

PR #5001, opened 2021-06-10
hark-graph-hook: speed it up with static conversions

tacryt-socryp:
Given that we know the types of the conversions used in `%hark-graph-hook` at compile time, there's no need to use dynamic conversions. Replaces all tubes with static conversions.


liam-fitzgerald (approved):
Based and staticpilled. Got any ideas on performance numbers?

tacryt-socryp:
No clue, can't save more than 30ms or so, but I bet it only helps by 5-10ms so I didn't profile it.

liam-fitzgerald:
please hold the merge on this until i get another OTA out the door

PR #5004, opened 2021-06-11
Fix mispelling in chat-cli help message

finned-palmer:
Noticed this misspelling in the output of the `;help` command.


Fang- (approved):
Nice catch, thank you!

PR #5005, opened 2021-06-11
btc: use blockfilter when requesting address info

ixv:
Now when we get updated about a new block, only request address info for addresses that match that block's blockfilter. If there's a gap of multiple blocks, we need to go request blockfilter information for those blocks specificially. 

This depends on this PR on the urbit-bitcoin-rpc repo, which allows you to make those requests for specific block information: https://github.com/timlucmiptev/urbit-bitcoin-rpc/pull/26

### DIFF
--- a/pkg/arvo/app/btc-wallet.hoon
+++ b/pkg/arvo/app/btc-wallet.hoon
@@ -167,6 +167,11 @@
   |=  [=wire =sign:agent:gall]
   ^-  (quip card _this)
   ?+  -.sign  (on-agent:def wire sign)
+      %watch-ack
+    ?~  p.sign  `this
+    %-  (slog leaf+"connection rejected by provider ({<src.bowl>})" u.p.sign)
+    `this
+    ::
       %kick
     ?~  prov  `this
     ?:  ?&  ?=(%set-provider -.wire)
@@ -724,13 +729,9 @@
       (retry-scans network)
     ==
   =?  cards  ?&(?=(^ blockhash) ?=(^ blockfilter) (gth gap 0))
-    ;:  weld  cards
-      (retry-filtered-addrs network u.blockhash u.blockfilter)
-    ==
+    (weld cards (retry-filtered-addrs network u.blockhash u.blockfilter))
   =?  cards  (gth gap 50)
-    ;:  weld  cards
-      (retry-addrs network)
-    ==
+    (weld cards (retry-addrs network))
   :-  cards
   %_  state
     prov  `p(connected %.y)
@@ -755,7 +756,9 @@
   %-  zing
   %+  murn  ~(tap by scans)
   |=  [[=xpub:bc =chyg] =batch]
-  ?.  =(network network:(~(got by walts) xpub))  ~
+  =/  w  (~(get by walts) xpub)
+  ?~  w  ~
+  ?.  =(network network.u.w)     ~
   `-:(req-scan batch xpub chyg)
 ::  +retry-addrs: get info on addresses with unconfirmed UTXOs
 ::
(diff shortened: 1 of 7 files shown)


timlucmiptev:
Looks good, just two comments/questions:

1. Do we want to expose bip158 via bitcoin.hoon? I split up the various BIP functionality just for modularity, but it doesn't seem like a relevant detail to app developers.

2. Do you want to have a `gap` size past which you just check addresses in the normal way? If a wallet was offline a long time and missed tons of blocks, the fetching of blockfilters would take longer than just checking addresses, since blockfilters aren't *that* small (~10% of total block size iirc).

ixv:
1. Not sure I understand, do you mean doing `/+  bip-b158` in `/lib/bitcoin.hoon` rather than in  `/app/btc-wallet.hoon`, or did you have something else in mind?

2. Yeah that's a good idea, what would you suggest is a good gap size?

timlucmiptev:
1. Yeah, include it in `lib/bitcoin.hoon`. If we only use 1 or 2 of its functions externally (I think that's the case?), you could even wrap just those in arms, which seems best now that I think about it.

2. I just checked the blockfilter for the latest block: it was ~32K bytes on a 1.5M block, so about 2% of the size. In terms of what we're good sending over Ames, I think about 50 blocks would be good here? 1.5M on a reconnect seems reasonable, but if you feel it's too much, use that ~30KB number to adjust as appropriate.

timlucmiptev:
If `all-match` is the only thing we use externally from BIP158, I would only expose that and `match` as public arms through `bitcoin.hoon`.

Other than that, looks good. Might want a review from @tacryt-socryp  on the provider kick/re-sub.

pkg/arvo/app/btc-wallet.hoon
```diff
@@ -704,102 +690,148 @@
...
-    `(poke-provider [%tx-info txid])
-  --
+  =?  cards  ?|(!connected.p (gth gap 0))
+    ;:  weld  cards
+      (retry-poym network)
+      (retry-txs network)
+      (retry-scans network)
+    ==
+  =?  cards  ?&(?=(^ blockhash) ?=(^ blockfilter) (gth gap 0))
+    ;:  weld  cards
```
tacryt-socryp, line 727: Use regular form with %+ when only using two items.

pkg/arvo/app/btc-wallet.hoon
```diff
@@ -704,102 +690,148 @@
...
+      (retry-filtered-addrs network u.blockhash u.blockfilter)
+    ==
+  =?  cards  (gth gap 50)
+    ;:  weld  cards
```
tacryt-socryp, line 731: Regular form

pkg/arvo/app/btc-wallet.hoon
```diff
@@ -704,102 +690,148 @@
...
+  |=  =txid
+  (poke-provider [%tx-info txid])
+::
+++  retry-scans
+  |=  =network
+  ^-  (list card)
+  %-  zing
+  %+  murn  ~(tap by scans)
+  |=  [[=xpub:bc =chyg] =batch]
+  ?.  =(network network:(~(got by walts) xpub))  ~
```
tacryt-socryp, line 758: Crashing here in the got by seems like a bad idea. Let's make sure we produce ~ instead

tacryt-socryp (changes requested):
I don't see anything here to handle a negative watch ack. We want to ensure we don't cause a resubscribe loop in that case. Otherwise, small style nits

ixv:
Not sure I follow, if we don't want to do anything in the negative watch-ack case, why isn't the default-agent handling of this sufficient?

tacryt-socryp:
Line 729 won't compile due to using a gap inside of the (weld ... ) statement. As for the watch nack, I think you're right, but we need a more semantic error message for the crash, likely with a `~|`

ixv:
Alright done, also found and fixed another issue with it

tacryt-socryp (approved):

PR #5007, opened 2021-06-12
Library matrix jets

sigilante:
This PR reflects a subset of the changes in #3536, which has become rather unwieldy. This includes the jets for #4908, the matrix library components.

I will ping when ready.

Besides the scope of this PR, some work will not be possible until an `@z` complex aura is introduced. Furthermore, some `@lv` work should be done once the matrix work is all integrated, in particular `++outer`.


sigilante:
Update on current status:

There are some persistent and pesky memory allocation issues.  These are particularly disturbing because the amounts of contiguous memory we're talking about are quite small.  I can avoid them for now by just letting the memory leak, but when I free the array it `bail:meme`s the system.  Based on what I know now, this may just have hit a fundamental limitation of the loom system for unpacked arrays, and may militate in favor of direct in-atom manipulations, which are more complicated to get right.  I'll counsel with @joemfb @ohAitch and others on this soon.  (They're present in the current uploaded commit but not highlighted; you can probably see them in the current CI tests.)

The jets mostly work, with the exception of a couple of legacy atom format pieces that I know about and need to fix (also visible in the CI tests, but I know about these and how to fix them quickly).  I'm also deciding "how much" to optimize the matrix inversion procedure.

sigilante:
The jets work and are complete (to the unit test coverage) with the exception of lingering memory allocation issues.

Right now several jets paste over this by not freeing memory with `u3a_free`.  This is wrong but turning on the memory freeing causes a hard crash (of Vere) rather than a soft crash (of the test thread).

What should I do to diagnose and solve the memory issues?  How useful is Valgrind with the Urbit loom?  `gdb`?  Debugging these problems is going to require some fresh thinking or procedures for me to try out.

pkg/urbit/jets/e/lms.c
```diff
@@ -0,0 +1,1492 @@
...
+    c3_w ii, jj;
+    union trip a;
+    fprintf(stderr, "[%u x %u] matrix:\n\r", m_u, n_u);
+    for ( ii = 1; ii <= m_u; ii++ )
+    {
+      fprintf(stderr, "[ ");
+      for ( jj = 1; jj <= n_u; jj++ )
+      {
+        a.c = u[(ii-1)*n_u+jj-1];
+        fprintf(stderr, "%f ", a.b);
```
ohAitch, line 113: This may be fine for tests - I'm failing to track down what's calling this actually - but for general use, note that printing rows individually can significantly disrupt vertical alignment. Compare:
18513121-F28A-48A4-A2C0-7B9336E62848

sigilante: Yeah, it's terrible but it's basically a debug printf.  Not currently in use but frequently helpful.

sigilante:
I've made no progress on resolving the memory overflows to date.

Some other things to think about:

- switching to native FP rather than SoftFloat
- implementing the outer product back into `@lv` once `@lm` is happy
- `@lmd` version once memory issues are solved
- complex aura, I've started to sketch this out as `@z`; this is necessary for eigenvectors etc.

@joemfb bumping this to glean your insight on solving memory issues in the loom.

sigilante:
Status update:

I've been doing bisection tests on memory issues to figure out what's going wrong.  Esp. since it only shows up unpredictably.

What I've been able to narrow it down to is a call to `u3r_met` in `shape`, which sometimes throws an out-of-loom access error.  Since the `u3_atom`s passed in are always declared normally, I admit myself at a loss for why this is happening.  **This seems to be the outstanding issue.**

The shape function that causes this looks like this:

```c
  struct
  dims
  {
    c3_w rows, cols;
  };

  struct dims
  shape(u3_atom p)  /* @ud */
  {
    c3_w mn = u3r_met(3, p)/4 - 1;  // total elements    <-- PROBLEM CALL
    c3_w d = u3r_word(mn+1, p);     // dimensionality, for now presumptively 2
    c3_w m = u3r_word(mn, p);       // number of rows

    struct dims shp;
    shp.rows = m;
    shp.cols = (c3_w)(mn / m);              // number of columns

    return shp;
  }
```

You can observe this behavior (assuming your memory behaves like mine!) with a generator like the following with the current commit.  (The unit tests do not yield this problem on my system but do on the CI system.)

```
/+  *lm, *lv
!:
:-  %say
|=  [[* eny=@uv *] [n=@ud ~] ~]
=/  vr  (make:lvs `(list @rs)`~[.-1 .-3 .-5 .-7 .-9])
=/  vc  (make:lvs `(list @rs)`~[.-10 .-20 .-30 .-40])
=/  mr  `@lms`(make:lms ~[~[.1 .2 .3] ~[.4 .5 .6] ~[.7 .8 .9]])
~&  `@ux`mr
~&  (unmake:lms mr)
~&  (unmake:lms `@lms`0x2.0000.0003.c040.0000.c000.0000.bf80.0000.0000.0000.3f80.0000.4000.0000.4040.0000.4080.0000.c080.0000.c040.0000.c000.0000.bf80.0000.0000.0000.3f80.0000.4000.0000.4040.0000.c0a0.0000.c080.0000.c040.0000.c000.0000.bf80.0000.0000.0000.3f80.0000.4000.0000)
~&  `@ux`(addm:lms mr mr)
[%verb ~]
```

This yields

```
address 0x67fcab7c0 out of loom!
Assertion '0' failed in noun/events.c:129

bail: oops
bailing out
```
typically at the same bad address.

sigilante:
A ha!  I'm pretty sure that what's happening is `len_w` is `0` sometimes and this is underflowing in `u3r_met` for the `dog` case.

So the next question is why an atom has length zero.  What path did I take that sometimes produces that condition?

sigilante:
It's a refcount I'm dropping somewhere.

_Narrator:  It's always a refcount._

sigilante:
Am I missing a refcount?  The following code bails on a trivial example and I can't for the life of me figure out what memory is leaking.  It fails at the `u3a_free` call, which matters.

```
/* getter from @lms, 1-indexed
*/
  u3_noun
  u3qelms_getr(u3_atom u,  /* @lms */
               u3_atom i)  /* @ud */
  {
    c3_w i_    = u3r_word(0, i);  // 1-indexed

    struct dims mn = shape(u);
    c3_w m_u = mn.rows;
    c3_w n_u = mn.cols;

    if ((i_ < 1) || (i_ > m_u))
    {
      return u3m_bail(c3__exit);
    }

    c3_w* v_  = (c3_w*)u3a_malloc((n_u+1)*sizeof(float32_t));
    u3r_words((i_-1)*n_u, i_*n_u, v_, u);
    v_[n_u] = 1;

    u3_atom v = u3i_words(n_u+1, v_);  /* @lvd */
    u3a_free(v_);
    return v;
  }

  u3_noun
  u3welms_getr(u3_noun cor)
  {
    u3_noun u, i;

    if ( c3n == u3r_mean(cor, u3x_sam_2, &u,
                              u3x_sam_3, &i, 0) ||
         c3n == u3ud(u) ||
         c3n == u3ud(i) )
    {
      return u3m_bail(c3__exit);
    }
    else {
      return u3qelms_getr(u, i);
    }
  }
```

Jbollenbacher:
any update on merging this?  matrix functionality is pretty critical for the future of hoon and urbit.  it'll open a lot of doors in app development.

PR #5010, opened 2021-06-14
eyre: cast within zapgar to avoid nest check

tacryt-socryp:
Surprisingly enough, this gets rid of a small nest check within spider. ~4-8ms speed-up to sending a message. I really can't seem to speed up spider any more at all, despite banging my head on it. Vases.


Fang- (approved):
Huh, kinda surprised this does anything, since the nouns in the cell here should already have exactly the types from the cell type being cast to. Wondering what's different...

The change here can't hurt though, we'll take the magic perf improvements.

PR #5012, opened 2021-06-14
gitattributes: treat package-lock.json files as binary

tacryt-socryp:


matildepark (commented):
Unsure if this is the best call, though we are annoyed at the LoC changes.

First, you'll want `merge=theirs` beside that additional line [as it'll default to ours if there is a merge conflict, which is basically what we never want](https://martin.beryllium.net/2019/08/02/automatically-resolve-npm-package-lock-conflicts-using-git/).

Second, we have a lot of package-locks. Why be specific about pkg/interface's?

Tagging @liam-fitzgerald for second opinions.

liam-fitzgerald:
I'm a fan, but you definitely should change the merge driver. Also, should apply to all package-lock.json files

tacryt-socryp:
Done.

matildepark (approved):

Issue #4958, opened 2021-05-28
glob stuck on %waiting, never downloads updated JS blob, leaving 404 page


matildepark:
This was always an issue for maybe <1% of users post-OTA in hte past, requiring a `:glob %kick`. Now that glob has multiple apps (implementation discussion in https://github.com/urbit/urbit/issues/4874, deploy in https://github.com/urbit/urbit/pull/4940) this seems to be happening much more frequently. The bandaid fix proposed in https://github.com/urbit/urbit/issues/4957 would hide the issue, but an updated JS blob simply _never downloading_ is not a sustainable solution.

Because this concerns Tlon hosting customers, and requires manual devops intervention, this is high priority to fix.

For anyone facing a 404'd Landscape or BTC Wallet, you'll want to use these commands now if either app won't load:

```
:glob [%kick /'~landscape'/js/bundle]
:glob [%kick /'~btc'/js/bundle]
```

PR #5013, opened 2021-06-14
ted/glob: add exponential retry

liam-fitzgerald:
The GCP bucket will occasionally fail requests for no good reason. Adds exponential retry to the glob thread to prevent this from affecting UX. 

Fixes urbit/urbit#4958


tacryt-socryp (approved):
LGTM

philipcmonk:
Perfect, it's factored as a thread exactly to make this sort of thing easy

PR #5016, opened 2021-06-15
observe-hook: add ability to warm static mark conversions

tacryt-socryp:
Simplified our other apps' caching strategies and consolidated them into a general purpose utility in `%observe-hook`.  The previous caching strategy was fragile when marks were changed. Performance is approximately the same.

### DIFF
--- a/pkg/arvo/app/graph-push-hook.hoon
+++ b/pkg/arvo/app/graph-push-hook.hoon
@@ -26,18 +26,15 @@
       state-one
   ==
 ::
-+$  cached-transform
++$  post-transform
+  $-  indexed-post:store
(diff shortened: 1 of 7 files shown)


liam-fitzgerald (approved):
LGTM! Aren't we reincurring the scry boundary by not saving the mark conversion in the state of the app?

tacryt-socryp:
The scry boundary is 0.5ms now, whereas when we originally were trying to avoid the scry boundary, it was 7-10ms.

tacryt-socryp:
I need to fix one thing in the `+on-load` of this, so please hold. Will merge when ready.

tacryt-socryp:
Tested on-load. All good.

PR #5020, opened 2021-06-16
dm-hook: fix updating pendings

liam-fitzgerald:
We weren't emitting facts when pendings got updated, so new pending messages would only show on refresh


tacryt-socryp (dismissed):
Cast any types before putting them into vases: for correctness guarantees and performance.

tacryt-socryp (approved):

PR #5025, opened 2021-06-16
npm: build ts -> js files before publishing

gregorytoprak:
When publishing to npm, the `dist/*.js` files weren't being built, so the packages didn't work.

cc: @tylershuster

### DIFF
--- a/pkg/npm/api/tsconfig.json
+++ b/pkg/npm/api/tsconfig.json
@@ -5,7 +5,7 @@
 	  "outDir": "./dist",
     "module": "ES2020",
     "noImplicitAny": true,
-    "target": "ES2020",
+    "target": "ES2017",
     "pretty": true,
     "moduleResolution": "node",
     "esModuleInterop": true,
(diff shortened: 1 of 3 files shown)


liam-fitzgerald:
Is it emitting `*.d.ts` files?

gregorytoprak:
When the `build` script is run, it emits all the right files, including *.d.ts. The problem was just the script wasn't being run so the npm-hosted package didn't have them.

pkg/npm/api/tsconfig.json
```diff
@@ -5,7 +5,7 @@
 	  "outDir": "./dist",
     "module": "ES2020",
     "noImplicitAny": true,
-    "target": "ES2020",
+    "target": "ES2017",
```
tylershuster, line 8: Any particular reason for this? Not complaining just curious

tylershuster (approved):
This looks good to me. Strictly by coincidence it's worked so far but this is good sanity henceforth

pkg/npm/api/tsconfig.json
gregorytoprak: I was getting incompatibility with the create-urbit-app defaults. Maybe it's my setup that's the problem, but all of this (incl the pull here: https://github.com/tylershuster/cra-template-urbit/pull/1) was just what it took to get the fresh CRA-urbit app running for me.

PR #5029, opened 2021-06-17
aqua: revive

liam-fitzgerald:
- Introduces a %boot-pill mark and associated types, to distinguish pills
that can be used to boot a ship and the ivory pill. Changes aqua to
receive this new %boot-pill mark
- Upgrades threads, generators and boot sequence for the new wire format introduced in arvo 240K
- Introduces no-op %wyrd negotiation into the boot sequence so that arvo can metamorphose successfully
- New limbs for +peek, +wish and +scry

Still todo:
- [x] Cleanup ph/io.hoon still has some cruft left in from when azimuth was handled by a thread
- [x] Make all ph tests run and pass
- [x] Boot sequence currently only works with a +brass pill. Should I switch on `nam.boot-pill` and get the compiler and arvo source differently for each pill?
- [x] What is `.non` and `.rev` in `$vere`


belisarius222:
This is a nit, but if we do need to distinguish different kinds of pills, I think I'd prefer making a new pill mark (%pill-ivory, maybe) for ivory pills instead of reassigning the %pill mark to refer to non-bootable pills, since there are already a lot of bootable .pill files sitting around from development.

—
~rovnys-ricfer
https://urbit.org

pkg/arvo/app/aqua.hoon
```diff
@@ -511,24 +508,37 @@
...
+          ^-  (list (pair term @))
+          :~  zuse+zuse
+              lull+lull
+              arvo+arvo
+              hoon+hoon-version
+              nock+4
+          ==
+          [/ %whom who.ae]  ::  who
+          -.kernel-ova.pil  :: load compiler
+          +<:kernel-ova.pil :: load arvo XX how to work with solid?
```
philipcmonk, line 531: I think we can just weld in `kernel-ova.pil` instead of implicitly asserting it has two events.  I think that should be all that's needed to support solid pills.  `userspace.ova` below should also be welded.

liam-fitzgerald:
> This is a nit, but if we do need to distinguish different kinds of pills, I think I'd prefer making a new pill mark (%pill-ivory, maybe) for ivory pills instead of reassigning the %pill mark to refer to non-bootable pills, since there are already a lot of bootable .pill files sitting around from development. — ~rovnys-ricfer https://urbit.org

The pill mark still refers to both ivory and bootable pills. Only need the boot-pill mark to prevent type erasure when poking aqua in the dojo

liam-fitzgerald:
ready to go, the question about some unused fields in `$vere` notwithstanding

philipcmonk:
I don't know what non is (probably a nonce identifying each time you start up the runtime?), but `rev` encodes the runtime version, for purposes of recognizing when to set `lag`.  When an OTA event crashes with `%wyrd` on its stack, we set `lag` to indicate the OTA is frozen until the binary is updated (and this shows up as a yellow notification in landscape).  When `rev` changes, we unset this (hoping the new binary works).

matildepark:
@joemfb @philipcmonk Ready for review.

philipcmonk (approved):
Sorry for the delay, this looks good

joemfb:
> The pill mark still refers to both ivory and bootable pills. Only need the boot-pill mark to prevent type erasure when poking aqua in the dojo

Why not just use `&pill` and `?<  ?=(%ivory -.pill)`?

pkg/arvo/app/aqua.hoon
```diff
@@ -507,28 +505,42 @@
...
-          [%dawn keys]
-          -.userspace-ova.pil
-          [//http-client/0v1n.2m9vh %born ~]
-          [//http-server/0v1n.2m9vh %born ~]
-          [//http-server/0v1n.2m9vh %live 8.080 `8.445]
+      %-  zing
+      :~
+        :~  [/ %wack 0]  ::  eny
+            :: [/ %verb `|]  :: possible verb
+            :^  /  %wyrd  [%nonce /a-version] :: dummy runtime version + nonce
```
joemfb, line 517: This should probably at least be `[~.nonce /aqua]`. To match vere's behavior, the nonce would be `(scot %ta (mug now.bowl))`.

pkg/arvo/lib/ph/io.hoon
```diff
@@ -186,27 +138,11 @@
...
-++  real-ship
-  |=  [=tid:spider =ship]
-  ~&  >  "booting real {<ship>}"
-  =/  m  (strand ,~)
-  =/  =vase  !>([tid %azimuth-command !>([%create-ship ship])])
-  ;<  ~  bind:m  (poke-our %spider %spider-input vase)
-  (check-ship-booted ship)
-::
-++  raw-ship
-  |=  [=ship keys=(unit dawn-event:jael)]
```
joemfb, line 205: Seems like it'd be worth keeping around the `+fake-ship` / `+real-ship` api.

joemfb (commented):
I don't like the `%boot-pill` mark. `:aqua` is the only reason to have a `%pill` mark in the first place, and the additional mark doesn't seem to accomplish anything that an assertion couldn't handle.

Otherwise this looks good. I've picked a couple nits:

pkg/arvo/lib/ph/io.hoon
liam-fitzgerald: It's still around, just by a different name. `++init-ship` takes a flag indicating whether it's a fakeship. It's worth noting that some of these functions were never updated to use the new azimuth model, where the state is persistent in the aqua agent.

ixv:
I can't comment directly on this line because its not touched by this PR, but on line 33 I think you want `pil=$>(%pill pill)`

PR #5030, opened 2021-06-17
graph-store: rework scry format to be more succinct and consistent

tacryt-socryp:
I'm refactoring the scry API to be less of a hodge-podge of use cases. I retain all functionality of the old scry API, and have changed only implementation, not any logical specifics. I'll document the scry API in this PR and in the file itself. I'd like a review mainly of the API definition itself (wording, path structure, etc).


pkg/arvo/app/graph-store.hoon
```diff
@@ -618,392 +615,272 @@
...
-      %_  $
-        b  t.b
-        e  +(e.f)
-        d  (~(put by d.f) c node(children [%empty ~]))
-        c  (snip c)
+        [%node *]
+      |^
+      =*    pax  t.t.t.t.t.path
+      ?+    pax  (on-peek:def path)
+          [%exists ^]
```
Fang-, line 707: Have you considered using the `%u` care for existence checks? That's more generically legible, think namespace exploration and such.

Fang- (commented):
Can we please get a comment in this file along the lines of:

https://github.com/urbit/urbit/blob/923e6683a6928764d3e16fab86f4418abd85b655/pkg/arvo/sys/vane/ames.hoon#L999-L1004

or whatever format/level of detail is most appropriate here.

It's rather difficult getting a good overview of the namespace here by chasing nested `?+`s and reading the atom parsing code. Also things like "firstborn" may or may not want a sentence of elaboration. (I'd guess it's the first/oldest child node?)

tylershuster:
heir

pkg/arvo/app/graph-store.hoon
liam-fitzgerald: Would be neat if `%u` scries were `HEAD` requests

matildepark:
@liam-fitzgerald This is ready for review and 
- would target our dated release, 
- requires some detail about any breaking changes, 
- and be flagged on urbit-dev

liam-fitzgerald (approved):
LGTM!

PR #5031, opened 2021-06-18
-graph-leave: don't crash on inconsistent state

liam-fitzgerald:
- Don't wait for an ack for pokes to `graph-pull-hook` and `graph-store`
- Return early, instead of crashing, if groups or metadata are missing

Fixes urbit/landscape#921


tacryt-socryp (changes requested):

tacryt-socryp (approved):
yah

PR #5033, opened 2021-06-18
bip158: change swp to rev in +to-range

tinnus-napbus:
change `swp` to `flim:sha` in +to-range of the bip158 lib to ensure leading 0's are preserved

### DIFF
--- a/pkg/arvo/lib/bip/b158.hoon
+++ b/pkg/arvo/lib/bip/b158.hoon
@@ -157,7 +157,7 @@
   ++  to-range
     |=  [item=byts f=@ k=byts]
     ^-  @
-    (rsh [0 64] (mul f (swp 3 dat:(siphash k item))))
+    (rsh [0 64] (mul f +:(flim:sha (siphash k item))))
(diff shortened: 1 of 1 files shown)


tinnus-napbus:
@timlucmiptev

pkg/arvo/lib/bip/b158.hoon
```diff
@@ -157,7 +157,7 @@
   ++  to-range
     |=  [item=byts f=@ k=byts]
     ^-  @
-    (rsh [0 64] (mul f (swp 3 dat:(siphash k item))))
+    (rsh [0 64] (mul f +:(flim:sha (siphash k item))))
```
Fang-, line 160: This should probably use `+rev` directly, instead of passing through some `sha`-related helper. `+rev` is explicitly the leading-zero-safe version of `+swp`.

```suggestion
    (rsh [0 64] (mul f (rev 3 (siphash k item))))
```

tinnus-napbus:
changed to `rev` rather than `flim:sha` as per @Fang- suggestion

Fang- (approved):
Thanks, this looks good to me!

PR #5034, opened 2021-06-19
gall: allow apps to run from any desk

philipcmonk:
(draft because this is the product of a single long pair programming session with @belisarius222, and this needs to be evaluated next week with a clear mind.  It seems to work fine, though)

Start with `|start %desk %app-name`.  `|start %app` still defaults to `%home`.

Everywhere in the kernel that we deal with marks, we infer the app it's
connected to and use the marks from that desk.

Also some light renaming in gall, especially path->wire and
current-agent->yoke.

Subsequent tasks:

- Dojo needs a syntax to run generators and threads from other desks
- The home desk should be split into at least a minimal base desk and
  big "userspace" desk.  Dill's initialization logic should be updated
  to handle
- |show-package, |install, and |uninstall should to be written
- Clay should have smarter handling of system versions instead of just
  ignoring what's on each desk.  It's not clear that this will work
  correctly when sys updates right now.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -64,7 +64,7 @@
   ==
 ::  $blocked-move: enqueued move to an agent
 ::
-+$  blocked-move  [=duct =routes move=(each deal sign:agent)]
++$  blocked-move  [=duct =routes move=(each deal unto)]
 ::  $stats: statistics
 ::
 ::    change: how many moves this agent has processed
@@ -272,6 +272,13 @@
   ++  mo-abet  [(flop moves) gall-payload]
   ++  mo-pass  |=(p=[wire note-arvo] mo-core(moves [[hen pass+p] moves]))
   ++  mo-give  |=(g=gift mo-core(moves [[hen give+g] moves]))
+  ++  mo-past
+    |=  =(list [wire note-arvo])
+    ?~  list
+      mo-core
+    =.  mo-core  (mo-pass i.list)
+    $(list t.list)
+  ::
   ::  +mo-boot: ask %ford to build us a core for the specified agent.
   ::
   ++  mo-boot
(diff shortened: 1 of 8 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -365,30 +372,39 @@
...
-    =/  =wire  /sys/lyv
-    =.  mo-core  (mo-pass /sys/lyv %c %warp our %home ~)
-    =/  =mool:clay
-      :-  da+date
-      %-  ~(gas in *(set [care:clay path]))
-      :*  [%z /sys/hoon/hoon]
+    ::
+    =/  sources=(jug desk [care:clay path])
+      %+  ~(put by *(jug desk [care:clay path]))  %home
+      %-  sy
```
tacryt-socryp, line 378: Likely better to use `+silt` for faster compile time or `~(gas in ...)`

tacryt-socryp (commented):
Overall approach looks good.

philipcmonk:
silt makes me cast it, and I got tired of writing the type out so many
times. Probably a sign that it should be a named type.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1753,6 +1765,16 @@
...
   ::
+  ?:  ?&  =(%d care)
+          =(~ path)
+          =([%$ %da now] coin)
+          =(our ship)
+      ==
+    =/  yok=(unit yoke)  (~(get by yokes.state) dap)
+    ?~  yok
+      [~ ~]
+    [~ ~ desk+!>(q.beak.u.yok)]
```
Fang-, line 1776: Worth noting that clay implements the `%d` care as "set of all desks". If gall is going to continue using the desk field for app names, we might want to have `%gd` produce a set of all running apps instead.

At this point clay also implements a handful of less-generic cares though, certainly no other vane is going to implement a care for mark conversions. Wondering how to best delineate between generic and special-purpose cares, and whether the latter should exist in the first place...

PR #5035, opened 2021-06-19
btc-wallet: Add eslint, prettier, husky. Fix linting errors

finned-palmer:
We'll change over to the urbit eslint config after we change btc-wallet over to typescript, but this should work for now.
Note that the `react/prop-types` rule is currently set to 0, this is in anticipation of switching over to TS (not worth defining prop types until we make the switch IMO). 

Husky will run lint-staged for all staged files automatically pre-commit. Other hooks can be added later (`npm test` pre-push, for instance).


matildepark:
This might conflict with pkg/interface’s install of husky, moving the git hook folder to a variety of places depending on what project you install. @tylershuster @liam-fitzgerald ideas on using husky in a monorepo?

finned-palmer:
I think as long as the npm prepare script correctly points to the pkg/btc-wallet it should be ok, right? I.e., : ` cd ../.. && husky install pkg/btc-wallet/.husky` vs `cd ../.. && husky install pkg/interface/.husky`

finned-palmer:
Hey @tylershuster or @liam-fitzgerald are we good to merge this, you think? So long as the prepare script points at the right dir I think we should be good to go. If you wanna test, pull down this branch and make some commit in the relevant package and watch what happens with hooks. Seems to work fine on my machine.

matildepark:
Neither really maintain this; you'll want @ixv

finned-palmer:
I think the current plan is to start using a separate btc-wallet release branch, I'll open the PR up against it when @timlucmiptev has it setup.

ixv (approved):
I went ahead and made the release branch and retargeted this PR. This all LGTM

matildepark:
Just want to note that if you work on both Landscape and the BTC wallet then husky will be in one of two locations, relocating Git hooks, and it might affect whether precommit hooks even run depending on which one you most recently installed. But I am not sure. It might be a case for Lerna? @liam-fitzgerald

PR #5037, opened 2021-06-19
settings: my channel -> my channels

Fang-:
This matches the phrasing used in the omnibox and on the tile.


matildepark:
Kinda curious, @liam-fitzgerald will this break state if we change the value here? We'll have stuff in localstorage / settings-store that don't correspond to this

liam-fitzgerald:
I don't think any references to my channels live in persisted state anywhere

matildepark (approved):
I guess it's the underlying `LeapCategories` here that makes the difference. LGTM

PR #5043, opened 2021-06-23
interface: componentisation sweep

liam-fitzgerald:
interface: unify ship image and names rendering
Unifies the rendering of ship names and images everywhere in landscape. 

interface: refactor to use anchors instead of onclick

Introduces a variety of links components and uses them instead of
listening for a click event on a regular DOM node. This is both advantageous
for accessibility, as well as being more performant because we don't have to
create unnecessary link nodes. It's also less code

interface: add Container component
Just a component for big white boxes with borders, useful for `<Dropdown />` and main body content


matildepark:
As per UI tests, nicknames have lost their semi-bold, needs to be restored.

zalberico:
@arthyn @liam-fitzgerald closing this one as stale since it looks like it's not likely to merge, let me know if I'm wrong and we can reopen.

PR #5045, opened 2021-06-23
btc-provider: refactor and improve performance

tacryt-socryp:
Big props to @ixv for helping track through all of this. Used to take 40s, now takes 0.5s to process a `%block-info` request.

### DIFF
--- a/pkg/arvo/app/btc-provider.hoon
+++ b/pkg/arvo/app/btc-provider.hoon
@@ -74,27 +74,25 @@
     ^-  (quip card _state)
     ?-  -.comm
         %set-credentials
-      :-  :~  do-ping:hc
-              (start-ping-timer:hc ~s30)
-          ==
-      %_  state
-          host-info
-        [api-url.comm connected=%.n network.comm block=0 clients=*(set ship)]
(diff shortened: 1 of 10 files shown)


ixv:
You want to keep all those jet hints in there?

tacryt-socryp:
Yes, jet hints are good to have to track any performance regressions over time.

ixv (approved):

timlucmiptev:
What's the rationale for not exposing `bitcoin-utils` via `bitcoin`? Those are pieces of logic that really should just be part of the `bitcoin` lib, but had to be put in their own so that the BIP libs could depend on them too without creating a circular dependency.

tacryt-socryp:
Easier to do jet hints / profiling when there isn't as much subject modification going on. ie =, multiple =< of other libraries, etc

timlucmiptev:
Gotcha--is there another simple way to keep the semantics of exposing via
bitcoin.hoon?

Пт, 25 июня 2021 г. в 08:45, L ***@***.***>:

tacryt-socryp:
We could just access them through bcu:bc within agents. In other words, there isn't really an obvious way

PR #5048, opened 2021-06-24
eyre: replace channel upon new GET request

liam-fitzgerald:
As SSE are unidirectional, the client always realises that the
connection has failed faster than the server does. Hence, resuming a
subscription is useless, because channels can only be bound to one duct
at a time. Now, instead of failing a request for a channel
that is already bound to a duct, we replace the duct and continue
normally.


liam-fitzgerald:
(This gives frequent `http: %cancel not handled yet` printfs, but I'd rather leave the move in for correctness)

tacryt-socryp (commented):
This logically makes sense. Will test tomorrow and approve.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1206,15 +1206,24 @@
...
+        =/  cancel-heartbeat  
+          ?~  heartbeat.u.maybe-channel  ~
+          :_  ~
+          %+  cancel-heartbeat-move  channel-id 
+          [date duct]:u.heartbeat.u.maybe-channel
+        =-  [(weld cancel-heartbeat -<) ->]
+        %.  [%cancel ~]
+        %*  .  handle-response
+          duct  p.state.u.maybe-channel
+        ==
```
Fang-, line 1226: Does `handle-response(duct p.state.u.maybe-channel)` (or `%_` if you must go tall) not work?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1206,15 +1206,24 @@
...
+      ::  and replace it
+      ::
+      =^  cancel-moves  state
+        ?.  ?=([%| *] state.u.maybe-channel)  
+          :_  state
+          (cancel-timeout-move channel-id p.state.u.maybe-channel)^~
+        =/  cancel-heartbeat  
+          ?~  heartbeat.u.maybe-channel  ~
+          :_  ~
+          %+  cancel-heartbeat-move  channel-id
```
Fang-, line 1220:
```suggestion
        ?.  ?=([%| *] state.u.maybe-channel)
          :_  state
          (cancel-timeout-move channel-id p.state.u.maybe-channel)^~
        =/  cancel-heartbeat
          ?~  heartbeat.u.maybe-channel  ~
          :_  ~
          %+  cancel-heartbeat-move  channel-id
```

Fang- (approved):
Sweet! Some nits, but this looks like it should be fine otherwise.

pkg/arvo/sys/vane/eyre.hoon
liam-fitzgerald: It does, but i've had some issue doing this in the past for whatever reason, updated!

Fang- (approved):
> send comments as heartbeats, not empty events 

Oh right, I guess naive implementations would just end up dispatching empty events, given a stream of `\n`. Curious to hear how this broke reconnect behavior, but the change here should be fine.

liam-fitzgerald:
Empty events reset the last event ID according to the spec, and if the last
event ID is not set, then the client should not attempt a reconnect

——
~hastuc-dibtux

https://urbit.org

philipcmonk:
> This logically makes sense. Will test tomorrow and approve.

@tacryt-socryp Is this good from your end?

tacryt-socryp (approved):

tacryt-socryp:
Good on my end

PR #5051, opened 2021-06-24
btc-wallet: refactor and improve performance

tacryt-socryp:
Used to take 28s to apply the blockfilter, now takes 8s. 70% improvement.

### DIFF
--- a/pkg/arvo/app/btc-wallet.hoon
+++ b/pkg/arvo/app/btc-wallet.hoon
@@ -149,10 +149,10 @@
   ++  handle-command
     |=  comm=command
     ^-  (quip card _state)
-    |^
     ?>  (team:title our.bowl src.bowl)
     ?-  -.comm
         %set-provider
+      |^
       ?~  provider.comm
         ?~  prov  `state
         :_  state(prov ~)
@@ -166,6 +166,12 @@
           (watch-provider:hc u.provider.comm)
           (give-update:hc %change-provider `[u.provider.comm %.n])
       ==
+      ::
+      ++  leave-provider
+        |=  who=@p
+        ^-  card
+        [%pass /set-provider/[(scot %p who)] %agent who^%btc-provider %leave ~]
+      --
     ::
         %check-provider
       =/  pax  /permitted/(scot %p provider.comm)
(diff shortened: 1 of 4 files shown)


pkg/arvo/app/btc-wallet.hoon
```diff
@@ -125,73 +126,536 @@
...
-    ::
-      %btc-wallet-action
-    ?<  =(our.bowl src.bowl)
-    (handle-action:hc !<(action vase))
-    ::
-      %btc-wallet-internal
-    ?>  =(our.bowl src.bowl)
-    (handle-internal:hc !<(internal vase))
-  ==
+    ?+  mark  (on-poke:def mark vase)
```
ixv, line 134: Large transfers of code between the cores make me nervous, can be hard to detect a missing `:hc`. Needs to be tested thoroughly.

pkg/arvo/lib/bip/b158.hoon
```diff
@@ -86,39 +88,44 @@
 ::   write appends to the back
 ::
 ++  str
+  ~%  %str  ..params  ~
   |%
   ++  read-bit
+    ~/  %read-bit
```
ixv, line 94: Since this change can be tested in isolation, I would be happy to approve this in its own PR, if we don't want to delay this for testing all the style changes.

pkg/arvo/app/btc-wallet.hoon
tacryt-socryp: Yes absolutely. For the record I've gone through and manually searched every arm call to ensure the hc is there where appropriate. I'm currently running this on my ship and it's working good. What would you like me to test?

ixv: I mean, basically all of it. Delete provider, set provider again, delete wallet, set wallet again, send btc to a ship, send btc to a non-ship address, receive btc from a ship, receive btc from a non-ship address

ixv: happy to show you how to set up testnet if you want to do it that way.

tacryt-socryp: Yeah, can we spin up a testnet tomorrow and go through it?

ixv: Sure

ixv (approved):

ixv:
We tested it all and it looks good

PR #5053, opened 2021-06-24
interface: general omnibus

liam-fitzgerald:
- Batches updates in graph-reducer, preventing chat update weirdness
- Clears toplevel unreads upon visiting collections
- Refactors subscription handling to account for the current brokenness of resume in eyre. (see #5048  )
- Fixes an unfiled bug where you could not unfold rich embeds in a PostItem

Fixes urbit/landscape#995


tacryt-socryp (approved):

matildepark:
Does this include https://github.com/urbit/landscape/issues/993 ?

liam-fitzgerald:
Yes, as part of batching updates

PR #5055, opened 2021-06-25
graph-fe: clone requests, to ensure identical timestamps

liam-fitzgerald:
Fixes strange behaviour in sending messages.


matildepark (changes requested):
Not quite what it says on the tin.

Jun-25-2021 16-06-56

matildepark (approved):

PR #5056, opened 2021-06-25
file-server: scry for identity

ixv:
Need the iphone app to get its own identity in a way that doesn't rely on the domain name.

I don't know if this is the best place to put this, but I don't know any better place. I figure since this is the place where`window.ship` is defined its somewhat related.


tacryt-socryp (approved):
LGTM

PR #5060, opened 2021-06-28
Gall: fix subscription synchronization

belisarius222:
Adds a unique "nonce" (scare quotes to escape the ire of irascible cryptographers) to each subscription wire in Gall 
so that responses before a resubscribe can be distinguished from those after a resubscribe.  Responses on anything but the current nonce are printed and ignored -- this might get loud, but I thought it would be better to leave the printing in for now.  

Only `%watch` and `%watch-as` wires have nonces added to them.  Old pre-nonce wires are treated as nonce 0 for backward compatibility.  Each app begins with a nonce of 1, which is incremented on each new subscription the app makes.

Both fresh boot and upgrade have been successfully tested on a fakezod, but I haven't tested how this interacts with the %clog resubscribe machinery.  This probably shouldn't be merged until we also have automatic Ames flow closing, to prevent a linearly growing space leak.

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1652,7 +1652,7 @@
   +$  bitt  (map duct (pair ship path))                 ::  incoming subs
   +$  boat                                              ::  outgoing subs
     %+  map  [=wire =ship =term]                        ::
-    [acked=? =path]                                     ::
+    [acked=? =path nonce=@]                             ::
   +$  bowl                                              ::  standard app state
           $:  $:  our=ship                              ::  host
                   src=ship                              ::  guest
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -45,6 +45,8 @@
 ::  $yoke: agent runner state
 ::
 ::    control-duct: TODO document
+::    run-nonce: unique for each rebuild
+::    sub-nonce: app-wide global %watch nonce
 ::    live: is this agent running? TODO document better
 ::    stats: TODO document
 ::    watches: incoming and outgoing subscription state
@@ -54,7 +56,8 @@
 ::
 +$  yoke
   $:  control-duct=duct
-      nonce=@t
+      run-nonce=@t
+      sub-nonce=@
       live=?
       =stats
       =watches
@@ -128,7 +131,7 @@
 ::
 +$  egg
   $:  control-duct=duct
-      nonce=@t
+      run-nonce=@t
       live=?
       =stats
       =watches
@@ -336,7 +339,7 @@
         control-duct  hen
         beak          bek
         agent         &+agent
-        nonce         (scot %uw (end 5 (shas %yoke-nonce eny)))
+        run-nonce         (scot %uw (end 5 (shas %yoke-nonce eny)))
       ==
     ::
     =/  old  mo-core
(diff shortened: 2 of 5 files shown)


philipcmonk:
Did you run `-test % ~` or similar to make sure nothing else is affected by the bowl definition change?

> This probably shouldn't be merged until we also have automatic Ames flow closing, to prevent a linearly growing space leak.

IMO this isn't necessary.  It's not that big of a leak.

pkg/arvo/sys/lull.hoon
```diff
@@ -1652,7 +1652,7 @@
   +$  bitt  (map duct (pair ship path))                 ::  incoming subs
   +$  boat                                              ::  outgoing subs
     %+  map  [=wire =ship =term]                        ::
-    [acked=? =path]                                     ::
+    [acked=? =path nonce=@]                             ::
```
tacryt-socryp, line 1655: Is there a reason for the user-developer to need access to this nonce?

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1648,25 +1736,32 @@
...
           %-  (slog >out=have< tang)
           (ap-error %watch-not-unique tang)  ::  reentrant, maybe bad?
         $(moves t.moves)
-      =.  outbound.watches.current-agent
-        (~(put by outbound.watches.current-agent) [short-wire dock] [| path])
-      $(moves t.moves, new-moves [move new-moves])
+      ::
+      =.  p.move.move
+        (weld sys-wire [(scot %ud sub-nonce.current-agent) sub-wire])
+      %=    $
```
tacryt-socryp, line 1755: Surprised this doesn't work as a `%_`

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1130,23 +1189,22 @@
         ==
       ::
           %pass
-        =/  =duct  system-duct.state
-        =/  =wire  p.card
-        =/  =neet  q.card
+        =/  [=wire =neet]  [p q]:card
```
tacryt-socryp, line 1192: Since p and q are already in the subject, is there a reason not to use `=*` here? We're no longer modifying `wire` in place

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -336,7 +396,7 @@
         control-duct  hen
         beak          bek
         agent         &+agent
-        nonce         (scot %uw (end 5 (shas %yoke-nonce eny)))
+        run-nonce         (scot %uw (end 5 (shas %yoke-nonce eny)))
```
tacryt-socryp, line 399: Can this be inline with the others?

pkg/arvo/sys/lull.hoon
belisarius222: No, not really, but it also doesn't seem problematic for them to know it, so I opted for just leaving it in the bowl, without needing to transform the `$boat` on every app activation to remove the nonces.

An alternative design would have two parallel maps in Gall: one to `[acked=? =path]`, and another to `nonce=@`.  Then only the first would be exposed in the bowl.  It's an option if we decide it's important to hide nonces from agents, but it would be a bit more unwieldy.

pkg/arvo/sys/vane/gall.hoon
belisarius222: well, there is a `=.` a couple lines later, but yeah, using a `=*` is probably better here

pkg/arvo/sys/vane/gall.hoon
belisarius222: it might, I could try it

pkg/arvo/sys/lull.hoon
tacryt-socryp: I see. Seems harmless, but I hope developers don't make use of it. Just a little strange to expose this sort of implementation detail

PR #5061, opened 2021-06-28
btc-provider: ping _only_ once every 30 seconds

tacryt-socryp:
We currently ping once every 30 seconds *and* every time a peer connects. This can lead to the provider hanging, particularly in cases when someone reconnects repeatedly. The solution is to ping only on our schedule of once every 30 seconds.

### DIFF
--- a/pkg/arvo/app/btc-provider.hoon
+++ b/pkg/arvo/app/btc-provider.hoon
@@ -72,8 +72,8 @@
     ^-  (quip card _state)
     ?-  -.comm
         %set-credentials
-      :_  state(host-info [api-url.comm %.n network.comm 0 *(set ship)])
-      ~[(start-ping-timer:hc ~s30)]
+      :-  ~[(start-ping-timer:hc ~s0)]
+      state(host-info [api-url.comm %.n network.comm 0 *(set ship)])
       ::
         %add-whitelist
       :-  ~
--- a/pkg/arvo/app/btc-provider.hoon
+++ b/pkg/arvo/app/btc-provider.hoon
@@ -16,12 +16,14 @@
 +$  card  card:agent:gall
 +$  versioned-state
     $%  state-0
+        state-1
     ==
 ::
 +$  state-0  [%0 =host-info =whitelist]
++$  state-1  [%1 =host-info =whitelist timer=(unit @da)]
 --
 %-  agent:dbug
-=|  state-0
+=|  state-1
 =*  state  -
 ^-  agent:gall
 =<
@@ -36,9 +38,9 @@
   =|  wl=^whitelist
   :-  ~
   %_  this
-      host-info
-    ['' connected=%.n %main block=0 clients=*(set ship)]
+      host-info  ['' connected=%.n %main block=0 clients=*(set ship)]
       whitelist  wl(public %.n, kids %.n)
+      timer      ~
   ==
 ::
 ++  on-save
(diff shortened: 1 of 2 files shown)


pkg/arvo/app/btc-provider.hoon
```diff
@@ -75,9 +75,7 @@
     ?-  -.comm
         %set-credentials
       :_  state(host-info [api-url.comm %.n network.comm 0 *(set ship)])
-      :~  do-ping:hc
-          (start-ping-timer:hc ~s30)
-      ==
+      ~[(start-ping-timer:hc ~s30)]
```
ixv, line 78: This is the command the provider enters to connect to the rpc server, no reason to delay 30 seconds here.

tacryt-socryp: Got it, I'll do a delay 0s. Feels good to do it through `behn` so only one code path is hitting the timers.

tacryt-socryp: Done!

ixv (approved):

pkg/arvo/app/btc-provider.hoon
```diff
@@ -74,9 +91,13 @@
...
-      :_  state(host-info [api-url.comm %.n network.comm 0 *(set ship)])
-      :~  do-ping:hc
-          (start-ping-timer:hc ~s30)
+      :_  %_  state
+              host-info  [api-url.comm %.n network.comm 0 *(set ship)]
+              timer      `now.bowl
+          ==
+      :*  (start-ping-timer:hc ~s0)
+          ?~  timer  ~
+          [[%pass /ping-time %arvo %b %rest u.timer] ~]
```
tacryt-socryp, line 100: Should be /block-time, no?

pkg/arvo/app/btc-provider.hoon
```diff
@@ -66,6 +75,14 @@
     ::
         %btc-provider-action
       (handle-action !<(action vase))
+    ::
+        %noun
+      ?.  =(q.vase %kick-timer)  `state
+      :_  state(timer `now.bowl)
+      :*  (start-ping-timer ~s0)
+          ?~  timer  ~
+          [[%pass /ping-time %arvo %b %rest u.timer] ~]
```
tacryt-socryp, line 84: ?

pkg/arvo/app/btc-provider.hoon
ixv: good catch

tacryt-socryp:
@ixv can you explain how changing that conditional fixes the bug?

ixv:
The `blocks` variable is a list of blocks we need to request block info for, the condition was formerly that if we hear about several new blocks we need to request info for those, and if the latest block didn't come back with blockhash or blockfilter data we need to request for that one as well. However if we're hearing about a block we already know about, we don't need to request data for it, regardless of whether it came in with blockhash and blockfilter.

pkg/arvo/app/btc-wallet.hoon
```diff
@@ -728,7 +728,7 @@
     =/  gap  (sub block block.btc-state)
     =?  blocks  (gth gap 1)
       (gulf +(block.btc-state) (dec block))
-    =?  blocks  ?|(?=(~ blockhash) ?=(~ blockfilter))
+    =?  blocks  ?&((gth gap 0) ?|(?=(~ blockhash) ?=(~ blockfilter)))
```
tacryt-socryp, line 731: Pretty long expression. It'd be easier to read if the ?& was in tall-form

PR #5063, opened 2021-06-29
interface: collections fixes 

liam-fitzgerald:
collections: key VirtualScroller correctly

Fixes an unfiled bug where navigating to two different grids
consecutively would crash the render

Skeleton: fix grid styling on hidden sidebar

Fixes urbit/landscape#1006

RemoteContent: center iframes

Fixes urbit/landscape#1005

RemoteContent: suspend w/ fallback on err, loading

Fixes an unfiled bug where embeds with no thumbnail would not display a fallback in the block view and would be blank instead

interface: add suspense utils

settings: fix notification optimism

Fixes urbit/landscape#1009


matildepark (changes requested):
RemoteContent needs to be flexed down, it regularly breaks outside the container and overlaps UI.

image

matildepark:
@arthyn

liam-fitzgerald:
Remote content flexing is solved for all browsers and screensizes except safari on a small screensize. Permalinks have been fixed

matildepark (approved):
Strict improvement, we likely need a mobile pass PR after this.

PR #5067, opened 2021-06-30
btc-provider: do not allow comets to connect unless we have explicitly whitelisted them

tacryt-socryp:
We had an errant comet that was slowing down `~tirrel` because it was still running the old `%btc-wallet` code that made repeated requests. I updated the conditions to ban all comets unless the one in particular is explicitly allowed. Let me know what you think.

### DIFF
--- a/pkg/arvo/app/btc-provider.hoon
+++ b/pkg/arvo/app/btc-provider.hoon
@@ -348,12 +348,16 @@
   |=  user=ship
   ^-  ?
   |^
-  ?|  public.whitelist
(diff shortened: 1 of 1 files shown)


ixv:
I already dislike the duplicated functionality between this permissions model and the groups permissions model, adding more complications to this doesn't seem like the right thing. It should really just point to a group.

I also dislike the idea of default banning comets from things. We should certainly have the ability to ban all comets, but there's no reason to make it a default.

PR #5068, opened 2021-06-30
btc-wallet: Switch to hooks and functions

finned-palmer:
This should be good to go based on my testing. Would be good to get others to poke around and test it.

For https://github.com/urbit/bitcoin-wallet/issues/34


timlucmiptev:
OK, I'm probably less qualified than you to review this, so I'll merge it once you tell me it's fully tested.

finned-palmer:
Ok, I think we're good to go. We really need to build out some unit tests eventually.

finned-palmer:
merge conflicts resolved

finned-palmer:
I derived a vpub from my master ticket's xpub, been using that to test. Doesn't look like we can actually sign TXs in bridge that were generated using a vpub, though (in either my branch or master). Tested signing with mainnet though and it works. Everything else is working in testnet, so this branch should be good to go.

PR #5069, opened 2021-07-01
interface: fixing chat reply only firing once

arthyn:
Fixes https://github.com/urbit/landscape/issues/1002, also includes moving message state higher and a good bit of refactoring the Chat components involved.

### DIFF
--- a/pkg/interface/src/views/apps/chat/components/ChatPane.tsx
+++ b/pkg/interface/src/views/apps/chat/components/ChatPane.tsx
@@ -1,38 +1,39 @@
 import { Col } from '@tlon/indigo-react';
 import { Content, Graph, Post } from '@urbit/api';
 import bigInt, { BigInteger } from 'big-integer';
-import _ from 'lodash';
 import React, { ReactElement, useCallback, useEffect, useRef, useState } from 'react';
 import create from 'zustand';
 import { useFileDrag } from '~/logic/lib/useDrag';
-import { useLocalStorageState } from '~/logic/lib/useLocalStorageState';
+import { retrieve } from '~/logic/lib/useLocalStorageState';
 import { useOurContact } from '~/logic/state/contact';
 import { useGraphTimesent } from '~/logic/state/graph';
 import ShareProfile from '~/views/apps/chat/components/ShareProfile';
 import { Loading } from '~/views/components/Loading';
 import SubmitDragger from '~/views/components/SubmitDragger';
-import ChatInput, { ChatInput as NakedChatInput } from './ChatInput';
+import ChatInput from './ChatInput';
 import ChatWindow from './ChatWindow';
 
 interface useChatStoreType {
+  id: string;
   message: string;
+  messageStore: Record<string, string>;
   setMessage: (message: string) => void;
 }
 
-export const useChatStore = create<useChatStoreType>(set => ({
+const unsentKey = 'chat-unsent';
+
+export const useChatStore = create<useChatStoreType>((set, get) => ({
+  id: '',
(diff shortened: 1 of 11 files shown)


matildepark (changes requested):
First PR! Hooray.

1. Fails tsc, it looks like?
2. You'll want your commits to match [our hygiene spec](https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md#hygiene).
3. Very curious about this 41k line diff in package-lock, what changed? package.json didn't bump a version or anything, so maybe just don't commit the package-lock.

matildepark:
Oh, it's failing Chromatic and tsc because the package-lock is too new, I see. Weird.

matildepark:
In general, you'll want to rebase and force push to keep your commits atomic, as well.

arthyn:
I'm on the newer npm so that's the reason for the package-lock blowing up in size. I can switch to the older one and push that so it won't be changed.

matildepark:
i'd just drop the package-lock from the commits outright, really; it doesn't need to be committed unless a dependency changes

pkg/interface/src/views/apps/chat/components/ChatPane.tsx
```diff
@@ -79,36 +101,27 @@ export function ChatPane(props: ChatPaneProps): ReactElement {
...
+  const chatInput = useRef<{ uploadFiles: (files: FileList | File[]) => void }>();
+  const setMessage = useChatStore(s => s.setMessage);
 
   const onFileDrag = useCallback(
     (files: FileList | File[]) => {
       if (!chatInput.current) {
         return;
       }
-      (chatInput.current as NakedChatInput)?.uploadFiles(files);
+      chatInput.current?.uploadFiles(files);
```
liam-fitzgerald, line 112: This ref technique is fine for now, but it's pervasive bc you want the drop target to be quite close to the root of the component tree but the receiver is usually quite deep in the tree. We should consider refactoring it.

liam-fitzgerald (commented):
Functional style looks v nice. Only concern is hacking in localStorage, we we can get that for free

pkg/interface/src/views/apps/chat/components/ChatPane.tsx
```diff
@@ -1,18 +1,40 @@
...
+  message: '',
+  messageStore: retrieve(unsentKey, {}),
+  setMessage: (message: string) => {
+    const store = get().messageStore;
+    store[get().id] = message;
+
+    localStorage.setItem(unsentKey, JSON.stringify(store));
+    set({ message });
+  }
+}));
```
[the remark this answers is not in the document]
arthyn: omg I saw that persist function a few times when looking at uselocalstate and it never clicked that I was doing the same thing here, good call

pkg/interface/src/views/apps/chat/components/ChatPane.tsx
arthyn: Yeah I didn't like this either was just trying to do the conversion.  Now that I'm looking at it again, I think it would be quick to refactor if y'all think we should just do it in this change set.

liam-fitzgerald: Go nuts

liam-fitzgerald: It'd be nice if this solution worked for our other drag and drop inputs as well

liam-fitzgerald (approved):
LGTM!

matildepark (approved):

PR #5070, opened 2021-07-02
metadata-store: do not hide groups on creation, unhide all current groups

liam-fitzgerald:
The `.hidden` flag was being incorrectly set upon new group creation, so any groups created since it's introduction would not show up in the omnibox.

Fixes urbit/landscape#1017

### DIFF
--- a/pkg/arvo/app/metadata-store.hoon
+++ b/pkg/arvo/app/metadata-store.hoon
@@ -307,7 +307,6 @@
     =?  hidden.met  ?=(%groups app-name.m)
       %.n
     [m [g met]]
-
   ::
   ++  hide-dm-assoc
     |=  assoc=associations:store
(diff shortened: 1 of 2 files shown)


pkg/arvo/app/metadata-store.hoon
```diff
@@ -292,6 +296,18 @@
...
+    |=  assoc=associations:store
+    ^-  associations:store
+    %-  ~(gas by *associations:store)
+    %+  turn  ~(tap by assoc)
+    |=  [m=md-resource:store [g=resource met=metadatum:store]]
+    ^-  [md-resource:store association:store]
+    =?  hidden.met  ?=(%groups app-name.m)
+      %.n
+    [m [g met]]
+
```
tacryt-socryp, line 310: you've got to stop introducing errant newlines

ixv (approved):

PR #5071, opened 2021-07-02
interface: handle malformed group links, fix mixed content errors on embeds

liam-fitzgerald:
Fixes urbit/landscape#1020


arthyn (approved):

arthyn:
Looks good, tsc failed though, maybe check that out

pkg/interface/src/stories/LinkDetail.stories.tsx
```diff
@@ -35,20 +36,57 @@ const node = {
...
+    'time-sent': 1609969377513,
+    signatures: [],
+    contents: [
+      { text: 'LindyMan' },
+      {
+        url: 'https://twitter.com/PaulSkallas/status/1388896550198317056'
+      }
+    ],
+    hash: null
+  },
```
matildepark, line 72: This is not a YouTube node

PR #5073, opened 2021-07-02
interface, btc-wallet: move to lerna to manage packages

liam-fitzgerald:
Moves to lerna to manage our dependencies.
Benefits:
- Global precommit linter, each package defines their own eslint rules
- Removes hacky preinstall scripts
- CI typescript checking has been replaced with a call to `npm run test` in each package that has been changed. `test` in interface now runs both typescript and jest, ensuring that we don't lose functionality
- Fixes build process for @urbit/api, so it doesn't need to wipe `dist/` every time you rebuild
- Introduces some convenience scripts for building the libraries

This should ***not*** be merged until the PR for janeway has been merged and published as well


matildepark:
cc @finned-palmer @ixv

jtobin:
(A new janeway version is available, so merge at will.)

arthyn (approved):
Damn this is awesome, didn't realize it was that easy to setup 👍🏼 

Ran `npm run bootstrap` at root and then the dev server in pkg/interface, poked around, everything looked normal, is there anything else that should be tested?

PR #5076, opened 2021-07-02
scan-progress added

timlucmiptev:
@ixv added the `%scan-progress` update type. It is sent in `+init-batches` and `+check-scan`.

`main` and `change` are the 2 sub-wallets. `~` means fully scanned; an `idx` value is the lowest index of the current scanning batch.


ixv:
What happens if it hits the `end-scan` case in `check-scan`? I'm not super familiar with this logic but it seems like it'll never actually set `main` and `change` to `~`.

timlucmiptev:
You're correct; I added that now.

ixv (approved):
Looks good

PR #5078, opened 2021-07-02
btc-prov/wallet: btc-provider only publishes information to the speci…

tacryt-socryp:
…fic client that is asking for it now, except for new blocks and connection info.

Thanks for finding this issue @ixv!

### DIFF
--- a/pkg/arvo/app/btc-provider.hoon
+++ b/pkg/arvo/app/btc-provider.hoon
@@ -155,7 +155,7 @@
     ^-  (quip card _state)
     :_  state
     ?.  ?|(connected.host-info ?=(%ping -.act))
-      ~[(send-update:hc [%| %not-connected 500])]
+      ~[(send-update:hc [%| %not-connected 500] ~)]
     :_  ~
     %+  req-card  act
     ^-  action:rpc-types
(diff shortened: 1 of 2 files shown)


timlucmiptev:
Wow, that's brutal. The only thing I can say is that this is some of the very first logic written in the whole app, and my intuitions around subscriptions were a lot worse than. But still.....

Anyway, the fix to put ship name in subscription path looks good.

Is there a case to be made that this should be a poke back from the provider? I wouldn't let it hold up merging this, just been seeing this a lot lately.

tacryt-socryp:
No, this is definitely better as a give than a poke, as it is logically a subscription.

pkg/arvo/app/btc-provider.hoon
```diff
@@ -327,21 +334,24 @@
...
 ++  send-status
-  |=  =status
+  |=  [=status ship=(unit ship)]
   ^-  card
   %-  ?:  ?=(%new-block -.status)
         ~&(>> "%new-block: {<block.status>}" same)
       same
-  [%give %fact ~[/clients] %btc-provider-status !>(status)]
+  =-  [%give %fact ~[-] %btc-provider-status !>(status)]
+  ?~  ship  /clients
```
ixv, line 343: Do wires work that way? I've never seen this before.

tacryt-socryp: Yep! You can look at `/lib/push-hook` for more examples.

ixv (approved):

PR #5080, opened 2021-07-06
Invites: fix dismissal, style pass

liam-fitzgerald:
Reworks part of the invite flow to be clearer, and also to prevent brokeness
- Errored invites now have to be dismissed manually, to ensure that the user is actually notified of the join failure.
- Invites reflow better on mobile screensizes
- Fixes invites dismissal failure https://github.com/urbit/landscape/issues/949

### DIFF
--- a/pkg/interface/src/views/components/Invite/Group.tsx
+++ b/pkg/interface/src/views/components/Invite/Group.tsx
@@ -1,16 +1,16 @@
 import { css } from '@styled-system/css';
-import {
-  Box,
-  Icon,
-  LoadingSpinner, Row, Text
-} from '@tlon/indigo-react';
+import { Box, Icon, LoadingSpinner, Row, Text } from '@tlon/indigo-react';
 import {
   accept,
   decline,
   hideGroup,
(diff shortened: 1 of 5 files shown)


liam-fitzgerald:
Also fixes https://github.com/urbit/landscape/issues/897

pkg/interface/src/views/components/Invite/Group.tsx
```diff
@@ -234,7 +231,9 @@ function InviteActions(props: {
           backgroundColor="white"
           onClick={hideJoin}
         >
-          {status?.progress === 'done' ? 'Dismiss' : 'Cancel'}
+          {[...joinResult].includes(status?.progress as any)
```
arthyn, line 234: Curious about the spread into array, seems like it's an array already?

arthyn (approved):
No idea about the hoon hah, but the rest looks good

matildepark (approved):
Skimming hoon looks ok; builds fine; would like to do a brief test with you on devnet

PR #5082, opened 2021-07-06
fixed address bug for legacy addresses

timlucmiptev:
Fixed `to-cord:adr` in `bitcoin.hoon`: it was printing old-style addresses with their Urbit `0c` prefix for the `@uc` aura.

### DIFF
--- a/pkg/arvo/lib/bitcoin.hoon
+++ b/pkg/arvo/lib/bitcoin.hoon
@@ -57,7 +57,9 @@
     ~/  %to-cord
     |=  a=address  ^-  cord
     ?:  ?=([%base58 *] a)
-      (scot %uc +.a)
+      %-  crip
+      %+  slag  2
+      (trip (scot %uc +.a))
     +.a
   ::
   ++  from-pubkey
--- a/pkg/arvo/lib/bitcoin.hoon
+++ b/pkg/arvo/lib/bitcoin.hoon
@@ -59,7 +59,7 @@
     ?:  ?=([%base58 *] a)
       %-  crip
       %+  slag  2
-      (trip (scot %uc +.a))
+      (scow %uc +.a)
     +.a
   ::
   ++  from-pubkey


timlucmiptev:
If there's a much faster way to do this operation just by manipulating the `@ta` directly, fine to switch to that.

pkg/arvo/lib/bitcoin.hoon
```diff
@@ -57,7 +57,9 @@
     ~/  %to-cord
     |=  a=address  ^-  cord
     ?:  ?=([%base58 *] a)
-      (scot %uc +.a)
+      %-  crip
+      %+  slag  2
+      (trip (scot %uc +.a))
```
Fang-, line 62: > If there's a much faster way to do this operation just by manipulating the `@ta` directly, fine to switch to that.

```suggestion
      (rsh 3^2 (scot %uc +.a))
```

(Also for future reference, you might use `+scow` in place of `(trip (scot ...))`.)

tacryt-socryp (approved):
LGTM

timlucmiptev:
So the suggested change doesn't work, unfortunately:
```
> (to-cord:adr:bc [%base58 0c2MubYjvhiDujsRHkw1BHtERPqbdBpxe64uP])
'2MubYjvhiDujsRHkw1BHt'
```

Original, with `scow` instead of trip scot:
```
> (to-cord:adr:bc [%base58 0c2MubYjvhiDujsRHkw1BHtERPqbdBpxe64uP])
'2MubYjvhiDujsRHkw1BHtERPqbdBpxe64uP'
```

ixv (approved):
Looks good. This should solve the provider disconnection issue right?

Fang-:
Yeah, that very strange indeed. Certainly seems to work fine in dojo:

```
> =v 0c2MubYjvhiDujsRHkw1BHtERPqbdBpxe64uP
> (scot %uc v)
~.0c2MubYjvhiDujsRHkw1BHtERPqbdBpxe64uP
> `@t`(rsh 3^2 (scot %uc v))
'2MubYjvhiDujsRHkw1BHtERPqbdBpxe64uP'
```

Wonder what shenanigans are afoot here...

timlucmiptev:
Yeah not sure--I tested by compiling the lib I just pushed and then running that code.

ixv:
I just tested this @Fang-'s way and didn't get the same issue as you @timlucmiptev

PR #5083, opened 2021-07-06
btc-provider: made interval ~m1, and allow provider to set interval

tacryt-socryp:
Pinging the server every 30 seconds spikes the CPU too frequently for most providers. I made it ping every 60 seconds instead as that seems to be less load. I also made the interval for ping frequency configurable so that people can play with it based on their own machine's performance characteristics.


ixv:
Make sure you target release/bitcoin-wallet from now on. Also please explain why you want this change.

ixv (approved):

PR #5085, opened 2021-07-07
npm/api: correct metadata resource interface

matildepark:
Fixes urbit/landscape#1035
Fixes urbit/landscape#1024

Our associations in Airlock referred to an `app` propertty in the association resource when all associations use [`app-name`](https://github.com/urbit/urbit/blob/938ce9d5576ec11a912846aa4e801bed8306bfea/pkg/interface/src/logic/api/metadata.ts#L39).

Incidentally when committing ran into some issues with the pre-commit hook; I added some missing dependencies to `pkg/npm/api` to resolve that and updated all the package-locks.


matildepark:
The failed tsc doesn't seem related to what I did, but I'll look ...

tacryt-socryp (approved):

PR #5086, opened 2021-07-07
vere: have king chdir to pier directory on boot

mrdomino:
The original impetus for this was creating the `%khan` socket: Unix domain
socket paths are limited to 108 characters since they have to fit into a
`struct sockaddr`; we want the `%khan` socket to be relative to the pier;
hence it's most expedient if the socket is itself a relative path of
known length, and is created with cwd located at the pier directory.

There has also been talk about having the binary chroot itself to the
pier directory for ~security reasons. The code to do so existed, but was
`ifdef`'d out (and would require further work, e.g. patching in
`/dev/urandom`.)

This patch merely calls `chdir` in `main` and sets `u3_Host.dir_c` to `"."`. It
seems to work; however, `u3_Host.dir_c` (and `u3_Local`) is now largely
redundant, and is used in a lot of places to construct paths that look
like "./foo". Hence, references to it should be cleaned up in a future
change.

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -676,12 +676,19 @@ main(c3_i   argc,
     u3_Host.bot_f = _stop_on_boot_completed_cb;
   }
 
-#if 0
-  if ( 0 == getuid() ) {
-    chroot(u3_Host.dir_c);
-    u3_Host.dir_c = "/";
+  {
+    mkdir(u3_Host.dir_c, 0700);
+
+    //  Some day, it might be nice to chroot() here or something.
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
LGTM

mrdomino:
Cherry-picking onto release/next-sys, changing merge base.

philipcmonk:
For reference, this should have gone into release/next-vere. Next-sys is
for arvo /sys stuff

mrdomino:
Is it worth reverting and redoing against next-vere @philipcmonk?

joemfb:
@mrdomino, yes please move these changes to `release/next-vere`.

PR #5092, opened 2021-07-09
CommentInput: stop recreating input

liam-fitzgerald:
Fixes urbit/landscape#1042


arthyn (approved):
Seems fine, I thought Formik had a way to disable this, but useMemo works.

PR #5094, opened 2021-07-09
hoon: add +pry, +up

jtobin:
Adds the following to layer two of hoon.hoon:

* a priority search queue data structure, `+pri`, and its validated form `+pry`
* a PSQ logic core, `+up`, for working with `+pri`

The implementation is more or less a clone of parts of the [psqueues](https://hackage.haskell.org/package/psqueues) library for Haskell, which is itself based on [a 2001 ICFP paper](http://www.cs.ox.ac.uk/ralf.hinze/publications/ICFP01.pdf).  The motivation for adding this to hoon.hoon proper, rather than e.g. zuse, is that it's expected we'll use this in layer five of hoon.hoon itself in order to implement/employ an LRU cache.

N.b., this is presently marked as draft because the in-progress test suite, not yet included, has picked up at least one logic error in `+jab`, the arm for altering the PSQ.  I'll tack on the test suite when I get it finished, but it seems worth submitting a draft PR now to gage opinions on overall style, etc.

cc @tacryt-socryp @jaspervdj


tacryt-socryp:
A naive question: what is the conceptual difference between an ordered map and a priority search queue?

philipcmonk:
In a priority search queue, the key you use to look up a value is different from the key you use to order them.  For example, to implement LRU cache reclamation, you want to order entries by how recently they were accessed (so you can pop the oldest one off), but you look up values by the cache key.

I think (?) a priority search queue where the ordering key and look up key are the same would be equivalent to an ordered map.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1897,6 +1938,488 @@
...
+    ?~  q.bee
+      [p.bee a]
+    [p.bee (put k p.u.q.bee q.u.q.bee)]
+  ::
+  ++  jib                                               ::  update at min-prio
+    =/  typ
+      (unit (trel _?>(?=(^ a) k.n.a) _?>(?=(^ a) p.n.a) _?>(?=(^ a) v.n.a)))
+    |*  f=$-(typ (pair * typ))
+    =/  vue  bot
+    =/  may
```
belisarius222, line 2241: .may looks unused

jtobin: Nice catch.  A vestigial helper gate, axed.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1897,6 +1938,483 @@
...
+    ?:  ?=(~ t.a)
+      ?:  =(k m.a)
+        (one k p v)
+      ?:  (gor k m.a)
+        (toy (one k p v) (one n.a))
+      (toy (one n.a) (one k p v))
+    ?-    -.t.a
+        %rlos
+      ?:  (gor k m.p.t.a)
+        (toy $(a [n.a l.p.t.a m.p.t.a]) [n.p.t.a r.p.t.a m.a])
```
tacryt-socryp, line 2184: Maybe just me, but these long `toy` statements seem like they'd be a lot clearer in tall-form.

jtobin:
I think this one is ready for review.  We encountered compilation time issues due to heavy wetness use in the previous implementation, so I've since rewritten things to get rid of most of that.

The PR now also includes a (passing) test suite that gives +up some reasonable exercise.

jtobin:
Some benchmarks available [here](https://gist.github.com/jtobin/4c05c17d5423810ed273ca7f64174af8).

jtobin:
Re-marking this as draft as I intend to add a fair bit to it over the next few days.

pkg/arvo/sys/hoon.hoon
jtobin: I think these are ok here.  I feel like this code, with all its gory balancing internals and such, is necessarily pretty terse anyway, so making these expressions tall would sort of stretch things out without adding much real clarity.

I'll mark this as resolved for now, but happy to hear additional/other opinions along these lines.

jtobin:
Collecting benchmark details here:

**ord**:

```
put:  [%took ~s2..f908]
get:  [%took ~s1..2d3e]
bot:  [%took ~s2..ae34]
del:  [%took ~s2..a74a]
```

**nat**:

```
put:  [%took ~s2..2945]
get:  [%took ~s1..e019]
bot:  [%took ~s1..2591]
del:  [%took ~s1..4cd9]

put (w/each prio greater than resident max):  [%took ~s1..a0d1]
```

pkg/arvo/sys/hoon.hoon
```diff
@@ -1897,6 +1898,1491 @@
...
+++  up
+  ~%  %up  ..up  ~
+  |*  [k=mold v=mold]
+  =>
+  ::  molds and mold builders
+  ::
+  |%
+  +$  elem  [=k p=@ =v]                                 ::  pri element
+  ::
+  +$  lnode  [n=elem l=ltree m=k r=ltree]               ::  loser tree node
```
pilfer-pandex, line 1914: what is the difference between `k.n` and `m` here?

pkg/arvo/sys/hoon.hoon
```diff
@@ -1897,6 +1898,1491 @@
...
+    |=  a=lnode
+    ^-  ltree
+    ?>  ?=(^ r.a)
+    ?-  -.r.a
+        %llos
+      ?:  (win [p.n.a k.n.a] [p.n.p.r.a k.n.p.r.a])
+        (llos n.a (rlos n.p.r.a l.a m.a l.p.r.a) m.p.r.a r.p.r.a)
+      (llos n.p.r.a (llos n.a l.a m.a l.p.r.a) m.p.r.a r.p.r.a)
+        %rlos
+      (rlos n.p.r.a (llos n.a l.a m.a l.p.r.a) m.p.r.a r.p.r.a)
```
pilfer-pandex, line 2042: I hate to say it, but I find this almost completely unreadable, along with a bunch of the following functions.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1897,6 +1898,1491 @@
     ?~  a  ~
     ?~(r.a [~ n.a] $(a r.a))
   --
+::                                                      ::
+::::  2r: psq logic                                     ::
```
pilfer-pandex, line 1902: it might be nice to expand "psq" to "priority search queue," at least in the documentation comments, or otherwise find a nice short english name for it

pkg/arvo/sys/hoon.hoon
jtobin: Yes, @tacryt-socryp mentioned the same.  My rationale for the terseness here is that, as data structure internals, these are sort of bound to be a little inscrutable anyway.  But given there are now two "this could be clearer" comments, I'll look at cleaning them up a little.

pkg/arvo/sys/hoon.hoon
jtobin: :+1: I'll touch up the comments in general.

pkg/arvo/sys/hoon.hoon
jtobin: `k.n` is the actual key corresponding to a `[priority value]` pair, whereas `m` is a 'split key' used to enforce so-called semi-heap & search tree structural properties.  You can see comparisons to `m` being used to determine balancing in `+put` and `+del`, for example.

pkg/arvo/sys/hoon.hoon
jtobin: Some comments added in 58d0ec8.

jtobin:
Closing as subsumed by the linked PR.

PR #5098, opened 2021-07-10
btc-wallet: Add sig to valid patp in send component

finned-palmer:
Fixes: https://github.com/urbit/bitcoin-wallet/issues/11

### DIFF
--- a/pkg/btc-wallet/src/js/components/lib/send.js
+++ b/pkg/btc-wallet/src/js/components/lib/send.js
@@ -81,10 +81,10 @@ const Send = ({ stopSending, value, conversion }) => {
     };
 
     let payeeReceived = e.target.value;
-    let isPatp = ob.isValidPatp(payeeReceived);
+    let isPatp = ob.isValidPatp(`~${deSig(payeeReceived)}`);
     let isAddress = validate(payeeReceived);
     if (isPatp) {
-      validPatPCommand(payeeReceived);
+      validPatPCommand(`~${deSig(payeeReceived)}`);
     } else if (isAddress) {
       setPayee(payeeReceived);
       setReady(true);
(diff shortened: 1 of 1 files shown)


pkg/btc-wallet/src/js/components/lib/send.js
```diff
@@ -68,30 +68,39 @@ const Send = ({ stopSending, value, conversion }) => {
...
       setCheckingPatp(false);
       setPayeeType('address');
       setValidPayee(true);
     } else {
-      setPayee(payeeReceived);
-      setReady(false);
-      setCheckingPatp(false);
-      setPayeeType('');
-      setValidPayee(false);
+      const possibleValidPatPMissingSig = '~'.concat(payeeReceived);
```
tacryt-socryp, line 94: Should just use deSig method (can find it in Landscape) to simplify this flow.

tacryt-socryp (changes requested):

pkg/btc-wallet/src/js/components/lib/send.js
finned-palmer: I might be misunderstanding what you meant here. `deSig` appears to remove a sig from a string, here we're adding a sig to a string and then testing whether or not that string is a valid patp.

tacryt-socryp: Use deSig as follows on the input before testing it:

`~${deSig(input)}`

This will guarantee that input has a leading ~, as deSig has no effect on input without a leading ~.

finned-palmer: Ah, ok. Updating now.

pkg/btc-wallet/src/js/components/lib/send.js
```diff
@@ -68,30 +69,39 @@ const Send = ({ stopSending, value, conversion }) => {
...
         setCheckingPatp(false);
       }, 5000);
       setCheckingPatp(true);
       setPayeeType('ship');
-      setPayee(payeeReceived);
+      setPayee(validPatP);
+    };
+
+    let payeeReceived = e.target.value;
+    let isPatp = ob.isValidPatp(payeeReceived);
```
tacryt-socryp, line 84: This isPatp variable should be called on the deSigged payeeReceived so that we don't need the unnecessary if/else branch.

tacryt-socryp (changes requested):

pkg/btc-wallet/src/js/components/lib/send.js
finned-palmer: ...that makes more sense, ha. Just pushed the fix.

tacryt-socryp (approved):

PR #5099, opened 2021-07-10
btc-wallet: Show wallet scan progress

finned-palmer:
Fixes https://github.com/urbit/bitcoin-wallet/issues/17

@urcades this matches your design here: 
image

### DIFF
--- a/pkg/btc-wallet/src/js/components/lib/balance.js
+++ b/pkg/btc-wallet/src/js/components/lib/balance.js
@@ -16,10 +16,12 @@ const Balance = () => {
     setPsbt,
     setFee,
     setError,
+    scanProgress,
   } = useSettings();
   const [sending, setSending] = useState(false);
   const [copiedButton, setCopiedButton] = useState(false);
   const [copiedString, setCopiedString] = useState(false);
+  const scanning = scanProgress?.main !== null || scanProgress?.change !== null;
 
   const copyAddress = (arg) => {
     navigator.clipboard.writeText(address);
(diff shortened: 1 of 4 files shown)


pkg/btc-wallet/src/js/components/lib/balance.js
```diff
@@ -93,10 +95,25 @@ const Balance = () => {
...
+            {scanning ? (
+              <Col alignItems="center">
+                <Row>
+                  <Text fontSize={1} color="orange">
+                    Balance will be updated shortly:
+                  </Text>
+                </Row>
+                <Row>
+                  <Text fontSize={1} color="orange">
+                    {scanProgress.main}/24 addresses scanned
```
tacryt-socryp, line 107: Why hardcoded 24?

tacryt-socryp (approved):

pkg/btc-wallet/src/js/components/lib/balance.js
finned-palmer: It seems to be the max number that comes back when the scan completes. Do you know if it will ever return higher than that number?

finned-palmer: @timlucmiptev may know better though

timlucmiptev:
Ah, so it's a param that's set--I have to check the default.

I think the simpler/best approach here is to just output the total number scanned so far, since we don't know how far we'll eventually go in checking the wallet until we've checked.

finned-palmer:
> Ah, so it's a param that's set--I have to check the default.
> 
> I think the simpler/best approach here is to just output the total number scanned so far, since we don't know how far we'll eventually go in checking the wallet until we've checked.
> 
> Also remember that we should be summing total addresses scanned in main wallet and change wallet, or indicating that in some way.

Ok, I can just show total "scanned so far" for each.

timlucmiptev:
Exactly.

Вс, 11 июля 2021 г. в 17:42, finned-palmer ***@***.***>:

finned-palmer:
Now just showing total main and change addresses scanned.

timlucmiptev:
LGTM

PR #5100, opened 2021-07-12
landscape: allow setting shortcut for leap

matildepark:
- Adds fallback for Leap shortcut in initial state (`meta+/`)
- Moves Leap from the App.js component to the other shortcut contexts in Content
- Allows the Leap shortcut to be set in the settings menu
- Shows the current shortcut on the Leap button

### DIFF
--- a/pkg/interface/src/views/components/StatusBar.tsx
+++ b/pkg/interface/src/views/components/StatusBar.tsx
@@ -7,7 +7,7 @@ import {
   Row,
   Text
 } from '@tlon/indigo-react';
-import React, { useRef } from 'react';
+import React, { useRef, useCallback } from 'react';
 import { useHistory } from 'react-router-dom';
 import { Sigil } from '~/logic/lib/sigil';
 import { uxToHex } from '~/logic/lib/util';
@@ -36,7 +36,9 @@ const StatusBar = (props) => {
   const invites = [].concat(
     ...Object.values(inviteState).map(obj => Object.values(obj))
   );
-  const metaKey = window.navigator.platform.includes('Mac') ? '⌘' : 'Ctrl+';
+  const leapKey: string = useSettingsState(
+    useCallback(s => s.keyboard['leap'], ['leap'])
+    )?.replace('meta+', '⌘') || '';
   const { toggleOmnibox } = useLocalState(localSel);
   const { hideAvatars } = useSettingsState(selectCalmState);
 
@@ -100,7 +102,7 @@ const StatusBar = (props) => {
             Leap
           </Text>
           <Text display={['none', 'inline']} ml={2} color='gray'>
-            {metaKey}/
+            {leapKey}
           </Text>
         </StatusBarItem>
         <ReconnectButton />
--- a/pkg/interface/src/views/components/StatusBar.tsx
+++ b/pkg/interface/src/views/components/StatusBar.tsx
@@ -7,7 +7,7 @@ import {
   Row,
   Text
 } from '@tlon/indigo-react';
-import React, { useRef, useCallback } from 'react';
+import React, { useRef } from 'react';
 import { Link } from 'react-router-dom';
 import { Sigil } from '~/logic/lib/sigil';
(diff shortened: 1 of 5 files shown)


pkg/interface/src/views/components/StatusBar.tsx
```diff
@@ -36,7 +36,9 @@ const StatusBar = (props) => {
   const invites = [].concat(
     ...Object.values(inviteState).map(obj => Object.values(obj))
   );
-  const metaKey = window.navigator.platform.includes('Mac') ? '⌘' : 'Ctrl+';
+  const leapKey: string = useSettingsState(
+    useCallback(s => s.keyboard['leap'], ['leap'])
+    )?.replace('meta+', '⌘') || '';
```
arthyn, line 41: Did you want to always show `⌘` here?

matildepark: If it's a meta shortcut, probably. Do you think it should be agnostic for Linux or the like?

arthyn: Yeah I think so, it seems like y'all were using `window.navigator.platform.includes('Mac') ? '⌘' : 'Ctrl+'` before.

matildepark: Yes, but there's no Linux analogue for the Meta key. Previously, we bound both Ctrl and Meta, and showed Ctrl for other platforms. If you set your key to Meta, what would you refer to it as in a Linux context? Just `Super+`?

arthyn: Oh I see, yeah I guess so, `⌘`, `Win` or `Super`

pkg/interface/src/views/landscape/components/Content.tsx
```diff
@@ -37,6 +37,12 @@ export const Content = (props) => {
     history.goBack();
   }, [history.goBack]));
 
+  useShortcut('leap', useCallback((e) => {
+    e.preventDefault();
+    e.stopImmediatePropagation();
+    props.toggleOmnibox();
+  }, []));
```
liam-fitzgerald, line 44: Needs props.toggleomnibox as a dependency

pkg/interface/src/views/components/StatusBar.tsx
```diff
@@ -36,7 +36,17 @@ const StatusBar = (props) => {
...
+  let metaKey = 'Super+';
+  if (window.navigator.platform.includes('Mac')) {
+    metaKey = '⌘';
+  } else if (window.navigator.platform.includes('Win')) {
+    metaKey = '⊞ Win+';
+  }
+
+  const leapKey: string = useSettingsState(
+    useCallback(s => s.keyboard['leap'], ['leap'])
+    )?.replace('meta+', metaKey) || '';
```
liam-fitzgerald, line 49: Most of this should be defined outside the scope of the component and then just pass the function to useSettingsState

liam-fitzgerald (changes requested):
Some nits, else LGTM

pkg/interface/src/views/components/StatusBar.tsx
matildepark: I never really understood how that worked with CalmEngine, can you give an example?

liam-fitzgerald:
```typescript
// outside component
  let metaKey = 'Super+';
  if (window.navigator.platform.includes('Mac')) {
    metaKey = '⌘';
  } else if (window.navigator.platform.includes('Win')) {
    metaKey = '⊞ Win+';
  }
const selKey = (s: SettingsState) => s.keyboard.leap.replace('meta+', metaKey);

// inside component

const leapKey = useSettingsState(selKey);
```

matildepark:
All nits addressed, good to stamp and merge

arthyn (approved):

matildepark:
Going through my dashboard and closing stale PRs ...

PR #5101, opened 2021-07-13
interface: prototype new graph state

liam-fitzgerald:
Unifies node storage, deprecating `graphs` `threadGraphs` and `flatGraphs`, replacing them with `views` and `nodes`. 
`nodes` is a map of graph + index -> post.
`views` is a map of graph + index + view name -> a list of indices[]

The intention here is that we can have multiple ways of viewing an arrangement of nodes that doesn't require the nodes to be contiguous or even at the same depth. To start, we use `"latest"` as a synonym for how we currently fetch nodes newest to oldest. `"thread"` and `"graph"` will be added in soon, mimicking the current `threadGraphs` and `flatGraphs` props. This paves the way for jumping to a message outside of backlog, by instantiating a `surrounding-${index}` view so we can load historical backlog without affecting the consistency of our `"latest"` view. Embeds of messages in the same channel, (i.e. replies) should no longer require an additional fetch, as we can look it up in the state. Making the graph storage flatter also makes the reducer much simply and also faster (2x speedup). 

Seeking feedback on the general architecture. I'm aware that separating the structure and content will make some of our rendering logic a bit weird, so ideas there are very much appreciated.

### DIFF
--- a/pkg/interface/src/logic/reducers/graph-update.ts
+++ b/pkg/interface/src/logic/reducers/graph-update.ts
@@ -1,4 +1,4 @@
-import { GraphNode } from '@urbit/api';
+import { GraphNode, Post } from '@urbit/api';
 import BigIntOrderedMap from '@urbit/api/lib/BigIntOrderedMap';
 import BigIntArrayOrderedMap, {
   arrToString,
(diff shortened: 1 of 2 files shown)


zalberico:
Closing this one since it looks stale and not likely to see any action soon. If I'm wrong and this should be prioritized we can reopen.

PR #5106, opened 2021-07-14
Revert "Merge pull request #5086 from urbit/jo/chdir"

mrdomino:
This reverts commit efd123bfc7ef017d99ddade80040e20eb9de2e0a, reversing
changes made to f0e11a84562c84f290590b69cb1d970bec348f48.

This reverts #5086, which should have been applied to release/next-vere,
not release/next-sys.

PR #5107, opened 2021-07-14
vere: have king chdir to pier directory on boot

mrdomino:
This was originally reviewed and applied in #5086, but that was based on release/next-sys, whereas it should have been based on release/next-vere. This PR has the correct merge base. See also #5106, which reverts the change on release/next-sys.

The original impetus for this was creating the %khan socket: Unix domain
socket paths are limited to 108 characters since they have to fit into a
struct sockaddr; we want the %khan socket to be relative to the pier;
hence it's most expedient if the socket is itself a relative path of
known length, and is created with cwd located at the pier directory.

There has also been talk about having the binary chroot itself to the
pier directory for ~security reasons. The code to do so existed, but was
ifdef'd out (and would require further work, e.g. patching in
/dev/urandom.)

This patch merely calls chdir in main and sets u3_Host.dir_c to ".". It
seems to work; however, u3_Host.dir_c (and u3_Local) is now largely
redundant, and is used in a lot of places to construct paths that look
like "./foo". Hence, references to it should be cleaned up in a future
change.

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -676,12 +676,19 @@ main(c3_i   argc,
     u3_Host.bot_f = _stop_on_boot_completed_cb;
   }
 
-#if 0
-  if ( 0 == getuid() ) {
-    chroot(u3_Host.dir_c);
-    u3_Host.dir_c = "/";
+  {
+    mkdir(u3_Host.dir_c, 0700);
+
+    //  Some day, it might be nice to chroot() here or something.
+    if ( 0 != chdir(u3_Host.dir_c) ) {
+      u3l_log("boot: chdir to %s failed: %s\r\n", u3_Host.dir_c,
+              strerror(errno));
+      exit(1);
+    }
+    c3_free(u3_Host.dir_c);
+    u3_Host.dir_c = strdup(".");
   }
-#endif
+
   u3_ve_sysopt();
 
   //  Block profiling signal, which should be delivered to exactly one thread.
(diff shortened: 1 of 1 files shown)


joemfb (approved):
LGTM

joemfb:
It turns out that this works well when running `urbit` out of `$PATH` (as we do in CI), but fails when invoking `urbit` via a relative path. This is pretty obvious in retrospect, and I should've caught it earlier.

Running with a relative path to the binary is pretty important, at least in development. To restore that functionality we could:

- only `chdir()` when binding the network socket
- save the old working dir, and chdir to it and back when we spawn a subprocess

Thoughts? Other ideas?

mrdomino:
Ouch. Duh. Hmm. 

I think it's mildly desirable to have cwd be the pier, as a baseline assumption at all times. So that suggests chdir'ing on spawn. But both are kind of gross.

Unless we want to do something insane like link in urbit-worker in the pier directory (which we might do in order to get chroot jails working), I think option 2 is our best bet. I can do this Saturday or so.

mrdomino:
Looking into this a bit more.

It seems like the only spawn we do is the one for the urbit-worker process. This suggests two more options:

1. Move the chdir to after `u3_lord_init`.
2. Resolve the path to `urbit-worker` when `wrk_c` is initialized at `pkg/urbit/daemon/main.c:663`. (On next-vere, the check for whether to daemonize is currently on line 666, about which I can offer no further comment.)

1 would probably be better for eventual chroot-readying. 2 seems more expedient. I'm inclined to go with 2.

mrdomino:
CI output is kind of inscrutable, but logically it seems like the issue is with the `-B` argument and similar.

Trying to play whack-a-mole with absolute-izing path arguments like that feels kinda gross, like I'm doing the wrong thing.

This strongly suggests that the correct approach is not option 2.

Between option 1 and "only `chdir()` when binding the network socket", I'm inclined to go with option 1 if we want to earnestly try the chroot jail thing, and `chdir()`-on-bind otherwise.

mrdomino:
Yep, sounds good. Simplest thing is probably "chdir() on socket open, then
chdir() back".

PR #5108, opened 2021-07-14
ChatEditor: removing unused code preventing slashes

arthyn:
Fixes urbit/landscape#1059, @liam-fitzgerald not sure what that code was used for do you remember?


liam-fitzgerald (approved):
I think this may have been part of autocomplete stuff, should be fine to be removed

PR #5111, opened 2021-07-14
App: filtering props from html

arthyn:
Fixes urbit/landscape#975


liam-fitzgerald (commented):
Does this fix it globally, or just for this component?

arthyn:
Does not fix globally, we'd have to probably make our own wrapper for `styled` and change all instances of styled or maybe switch to using "transient" props as they describe here: https://github.com/styled-components/styled-components/releases/tag/v5.1.0

matildepark:
> Does this fix it globally, or just for this component?

Do we even see this in other components?

zalberico:
@arthyn same question here - is this little change good to clean up and merge or should we close it?

PR #5112, opened 2021-07-16
embeds: convert embed links for better usability

arthyn:
Fixes urbit/landscape#987 and adds same functionality to graph embeds.


matildepark:
I don't think it's 897?

For future reference — the best place for "Fixes urbit/landscape###" is generally in the commit description. If you do it in the PR, then when the PR merges, it should close the issue, but it doesn't if you don't merge it into master. If you do it in the commit, then when the commit eventually hits master, the issue closes, which is best for our release branch system.

arthyn:
Ah ok didn't know that, that makes sense. I can do that next time.

liam-fitzgerald (approved):
LGTM! Wondering whether we should include this in the tab order, but probably out of scope for here

PR #5113, opened 2021-07-17
support third-party software distribution

belisarius222:
Working PR, for transparency and ease of review.  Not ready for primetime yet, but the latest solid pill boots.

Some notes:
- I switched solid pills to just use clay `page`s, without converting to mime
- ford should share caches across desks
- gall should send %warp's to clay instead of scrying agent builds; rebuilding agents is way too slow
- I went ahead and tried the dojo yolo mode for shadow marks (I wrote it; actually haven't tried it yet)
- clay should get (at least two) verbosity levels; it's really nice to see all the files get built in ford, but not under all circumstances
- I cleaned up hood a bit; now it strips the sub-app's name off of every incoming wire, not just some of them
- I removed :goad, since the %goad move isn't part of gall's interface anymore -- if we need a goad, it could be rebuilt by writing something that sets an agent to %idle and then %jolt's it again
- I need to rewrite the |fade generator to have it ask kiln to suspend the agent and store that state in its $rein
- other than boot and some generators, this code hasn't been tested; next step is to make a landscape desk and test installing that, then test messing with kelvin versions in various desks and make sure the system code does the right thing

### DIFF
--- a/pkg/arvo/app/hood.hoon
+++ b/pkg/arvo/app/hood.hoon
@@ -14,8 +14,8 @@
       [%11 drum=state-2:drum helm=state:helm kiln=state-0:kiln]
       [%12 drum=state-2:drum helm=state:helm kiln=state-0:kiln]
       [%13 drum=state-2:drum helm=state:helm kiln=state-1:kiln]
-      [%14 drum=state-3:drum helm=state:helm kiln=state-1:kiln]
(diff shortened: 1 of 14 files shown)


joemfb:
@belisarius222 is there anything here that wasn't already released?

philipcmonk:
Released

PR #5114, opened 2021-07-17
vere: use realpath() of urbit-worker binary

mrdomino:
This resolves an issue introduced in #5107: when the path to the urbit
binary is relative, chdir() prevented vere from finding the urbit-worker
executable (which is supposed to be in the same directory as the urbit
binary.)


joemfb:
@mrdomino this is already being handled in your %kahn PRs? Can we close this PR?

mrdomino:
Yes.

PR #5116, opened 2021-07-19
interface: prevent crash in reducers on new group DMs

liam-fitzgerald:
Fixes urbit/landscape#1044 by preventing a crash in the reducer that would kill the subscription. FWIW, i strongly dislike that we receive the cards backwards on the channel (which is the cause of this bug), i'd be interested in ways to make this more usable. Wondering whether we should catch exceptions from callbacks in @urbit/http-api ?


matildepark (approved):
I guess it makes sense to skip it, since we'll get the members in the initial.

PR #5117, opened 2021-07-19
|mount: support simple @tas arg

belisarius222:
I got sick of running `|mount /=landscape=`, or forgetting that the syntax is different from `|commit %landscape`.  This preserves all existing behavior, but now you can also just type `|mount %landscape`, which will expand out to `/=landscape=`.

### DIFF
--- a/pkg/arvo/gen/hood/mount.hoon
+++ b/pkg/arvo/gen/hood/mount.hoon
@@ -7,11 +7,21 @@
 ::::
   ::
 :-  %say
+=>  |%
+    +$  bath
+      $@  desk
+      (lest knot)
+    --
 |=  $:  [now=@da eny=@uvJ bec=beak]
-        [[pax=path pot=$@(~ [v=@tas ~])] ~]
+        [=bath pot=$@(~ [v=@tas ~])]
+        ~
     ==
-?~  pot
-  =+  bem=(need (de-beam pax))
-  $(pot ~[?^(s.bem (rear s.bem) q.bem)])
-:-  %kiln-mount
-[pax v.pot]
+=/  bem=beam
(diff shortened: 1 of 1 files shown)


pkg/arvo/gen/hood/mount.hoon
```diff
@@ -7,11 +7,21 @@
...
+        [=bath pot=$@(~ [v=@tas ~])]
+        ~
     ==
-?~  pot
-  =+  bem=(need (de-beam pax))
-  $(pot ~[?^(s.bem (rear s.bem) q.bem)])
-:-  %kiln-mount
-[pax v.pot]
+=/  bem=beam
+  ?@  bath  [bec(q bath) /]
```
Fang-, line 20: Shouldn't this expand to `[bec(q bath, r da+now) /]` to cover the "ran the generator while `|cd`'d into a not-latest revision case? I imagine we only want to mount non-live desks explicitly (ie `|mount %` in that case).

belisarius222: I'm not entirely sure what you mean by that last sentence, but I think if run from a non-live checkout, you should still be able to `|mount %foo` and have that mean the latest revision of the `%foo` desk, which would imply that I should indeed make the change you're suggesting.

Fang-: We already support mounting older revisions of desks. I was just reiterating the proposed change by saying "we don't want to mount older revisions if the user doesn't explicitly say so in the command arguments".

belisarius222:
@Fang- addressed your comment, PTAL

Fang- (approved):
Good, thank you!

Fang-:
This change has since been released. Closing. (:

PR #5122, opened 2021-07-20
PostContent: ensure group feed spacing correct

arthyn:
Refactors post content height detection to use resize observer so we always get correct truncation and adds small margin when doing so.

### DIFF
--- a/pkg/interface/src/views/landscape/components/Home/Post/PostItem/PostContent.tsx
+++ b/pkg/interface/src/views/landscape/components/Home/Post/PostItem/PostContent.tsx
@@ -2,6 +2,7 @@ import { Col, ColProps } from '@tlon/indigo-react';
 import { Post } from '@urbit/api';
 import React, { ReactElement, useCallback, useState } from 'react';
 import styled, { css } from 'styled-components';
+import { useResize } from '~/logic/lib/useResize';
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (commented):
Have you checked this on safari? It's a total madhouse vis a vis resize observer. Else, LGTM!

arthyn:
Checked Safari, get this 
image

but it's still working, and from what I can tell by googling that issue, doesn't seem to affect anything.

pkg/interface/src/views/landscape/components/Home/Post/PostItem/PostContent.tsx
```diff
@@ -27,15 +29,13 @@ const PostContent = ({ post, isParent }: PostContentProps): ReactElement => {
...
   const showFade = !isParent && height >= 300;
 
-  const measuredRef = useCallback((node) => {
-    if (node !== null) {
-      setHeight(node.getBoundingClientRect().height);
-    }
-  }, []);
+  const resizer = useResize<HTMLDivElement>(useCallback((entry) => {
+    setHeight(entry.target.getBoundingClientRect().height);
+  }, [setHeight]));
```
liam-fitzgerald, line 34: setters unnecessary in dependency lists

liam-fitzgerald (approved):
Nit, else LGTM

PR #5127, opened 2021-07-21
btc-wallet: Feature/surface psbt on frontend

finned-palmer:
For https://github.com/urbit/bitcoin-wallet/issues/39

This does allow a user to either copy a psbt to clipboard or download a psbt file, but I think we should still get @urcades's opinion on how it looks (and how best to add some explainer copy).

Here's how the send screen looks now (after popping open the existing ellipsis menu):
send

And here's how the external transaction signing screen itself looks:
sign

I think this screen probably needs the most work from a UX perspective. You can either click on the PSBT string directly to copy it to clipboard or click the "Copy" button (similar to addresses on the balance screen), or you can download the PSBT file (probably what most people will do).

I can envision what the helper text would say here pretty easily, just not sure where you might want to stick it. Maybe like a question mark icon that a user can hover over right next to "Partially-signed Bitcoin Transaction"? Maybe just a a "What's this?" link somewhere (which could link to an explainer article somewhere external, or maybe a modal within the app itself)?

One other note: at the moment, we don't have a way to know if the xpub the user gave us is the one provided by bridge or if it's an external xpub, so they'll need to know that they have to select "Sign Externally" when sending a transaction. Should we include some copy on that initial xpub input modal to that effect?

cc: @urcades @timlucmiptev


finned-palmer:
Oh, I also know that this doesn't match the new figma screens for the wallet, I figure we'll work through the UI/UX updates from the latest figma screens as separate feature work.

finned-palmer:
Fixed merge conflicts.

urcades:
yo @finned-palmer ! This is looking pretty solid.

I've given this a bit of thought, and without offering too much of a ux edit, I've simply translated your above interfaces into something that feels a little easier to reason about, with more explicit visual indicators for interactivity:

Frame 1447

The primary differences are as follows:

1. In the first screens, prior to the actual signing step, you'll not that some of the forms and buttons are given backgrounds that make the points of interactivity on the card a little easier to parse. None of these are particularly functionality-critical, but they're a part of the more recent "visual language" for the BTC wallet, and I'd highly vouch for their inclusion across the app.
2. In the final signing step, you'll note that I've save a bit more space by situating the buttons associated with the PSBT text closer in look and feel, using icons instead of text to indicate the actions that one can take on the text.

finned-palmer:
Updated to match @urcades's designs (except for a few things that will be addressed in a later PR all around updating styles for the whole app, mostly around "card open/closed" state).

finned-palmer:
Should be ready to go.

timlucmiptev:
lgtm, just need @ixv

finned-palmer:
@ixv have you seen this yet?

ixv (approved):
Looks good, do we want to merge this one first?

finned-palmer:
Yep, thanks!

PR #5128, opened 2021-07-22
interface: actually resume channels

liam-fitzgerald:
- Reenables subscription resume after the fix blocking it was releasd.
- If kicked, restart channel
- If we are restarting a channel totally, then wipe graph-store state to avoid data inconsistency

Running this on my phone produces a roughly 100ms reconnect time from launch. Merging this should probably be held until the dated release, as changing the channel interface is notoriously prone to breakage.

### DIFF
--- a/pkg/interface/src/logic/state/base.ts
+++ b/pkg/interface/src/logic/state/base.ts
@@ -1,7 +1,7 @@
 import { applyPatches, Patch, produceWithPatches, setAutoFreeze, enablePatches } from 'immer';
 import { compose } from 'lodash/fp';
 import _ from 'lodash';
-import create, { GetState, SetState, UseStore } from 'zustand';
+import create, { GetState, SetState, UseStore, PartialState } from 'zustand';
 import { persist } from 'zustand/middleware';
 import Urbit, { SubscriptionRequestInterface, FatalError } from '@urbit/http-api';
 import { Poke } from '@urbit/api';
@@ -94,6 +94,7 @@ export interface BaseState<StateType extends {}> {
   removePatch: (id: string) => void;
   optSet: (fn: (state: StateType & BaseState<StateType>) => void) => string;
   initialize: (api: Urbit) => Promise<void>;
+  clear: () => void;
 }
 
 export function createSubscription(app: string, path: string, e: (data: any) => void): SubscriptionRequestInterface {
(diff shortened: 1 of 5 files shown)


pkg/interface/src/logic/state/base.ts
```diff
@@ -112,8 +115,12 @@ export const createState = <T extends {}>(
   name: string,
   properties: T | ((set: SetState<T & BaseState<T>>, get: GetState<T & BaseState<T>>) => T),
   blacklist: (keyof BaseState<T> | keyof T)[] = [],
-  subscriptions: ((set: SetState<T & BaseState<T>>, get: GetState<T & BaseState<T>>) => SubscriptionRequestInterface)[] = []
+  subscriptions: ((set: SetState<T & BaseState<T>>, get: GetState<T & BaseState<T>>) => SubscriptionRequestInterface)[] = [],
+  clear?: Partial<T>
```
arthyn, line 119: I think it's a little confusing that the empty fields themselves are called clear and the function to use them is called clear.

arthyn (changes requested):
I think I understand this correctly. Might need you to explain later why the shallow children call immediately in bootstrap though. 

Code itself looks good, except for that one naming thing.

tacryt-socryp:
Shallow children call is for grabbing the list of people we have a DM open with.

liam-fitzgerald:
Been running this on my phone for the last couple of days, running into some issues, will discuss in weekly

arthyn (approved):
I'll approve as is, but I'm curious what the issues are

tacryt-socryp:
What are the issues with this @liam-fitzgerald ?

liam-fitzgerald:
Oh right, we never got to this. The long and short of it is that clogs are fairly pervasive on mobile, which is deeply annoying, but it's not necessarily a step back from the current UX. Increasing the clog limit doesn't actually make this any better, because of the performance issues fixed by https://github.com/urbit/urbit/pull/5132 I'm wondering if we should make subscriptions more granular w.r.t. graphs and be more aggressive about stopping them etc.

liam-fitzgerald:
Depending on the activity, a clog can occur in a couple of minutes of inactivity on a phone

tacryt-socryp:
We should determine an acceptable average clog time for mobile, say 15 minutes. Then we should tweak our approach until we hit that. I'm open to more granular graph subscriptions.

PR #5129, opened 2021-07-22
notify: push notifications for mobile app

ixv:
We use a single app as both provider and client for push notifications. The provider can add or remove a service, and a client can connect or disconnect to a provider.

When a client connects to a provider, the provider sets up a binding for them, and then subscribes to the client. The client has a subscription to their own hark-store, which they filter for dms, turn into `notify` specific notifications, and forward to the provider. The provider then sends the push notification.

This is written in a somewhat general way, the idea being that many providers could provider many services, but we're somewhat constrained in specificity to dms for and the specific twilio api we're targeting, for the time being. It can be made more general in the future.

### DIFF
--- a/pkg/arvo/app/notify.hoon
+++ b/pkg/arvo/app/notify.hoon
@@ -1,0 +1,449 @@
+::
+/-  *notify, resource, hark-store, post
+/+  default-agent, verb, dbug, group
+::
+|%
++$  card  card:agent:gall
+::
++$  provider-state  (map term provider-entry)
++$  provider-entry
+  $:  notify-endpoint=@t
+      binding-endpoint=@t
+      auth-token=@t
+      clients=(map ship (unit @t))
+      =whitelist
+  ==
+::
++$  client-state
+  $:  providers=(jug @p term)
+  ==
+::
++$  state-0
+  $:  %0
+      =provider-state
+      =client-state
+  ==
+::
++$  versioned-state
+  $%  state-0
+  ==
+::
+--
+::
+=|  state-0
+=*  state  -
+::
+%-  agent:dbug
+%+  verb  |
+^-  agent:gall
+::
+=<
+  |_  =bowl:gall
+  +*  this  .
+      def  ~(. (default-agent this %|) bowl)
+      do   ~(. +> bowl)
+  ::
+  ++  on-init
+    :_  this
+    [%pass /hark %agent [our.bowl %hark-store] %watch /updates]~
+  ::
+  ++  on-save   !>(state)
+  ++  on-load
+    |=  =old=vase
+    ^-  (quip card _this)
+::    `this(state *state-0)
+    =/  old  !<(versioned-state old-vase)
+    ?-  -.old
+        %0
+      `this(state old)
+    ==
+  ::
+  ++  on-poke
+    |=  [=mark =vase]
+    ^-  (quip card _this)
+    |^
+    =^  cards  state
+      ?+  mark  (on-poke:def mark vase)
+        %notify-provider-action  (handle-provider-action !<(provider-action vase))
+        %notify-client-action    (handle-client-action !<(client-action vase))
+      ==
+    [cards this]
+    ::
+    ++  handle-provider-action
+      |=  act=provider-action
+      ^-  (quip card _state)
+      ?-  -.act
+          %add
+        ?>  (team:title our.bowl src.bowl)
+        =/  new-entry=provider-entry
+          :*  notify.act
+              binding.act
+              auth-token.act
+              ~
+              whitelist.act
+          ==
+        [~ state(provider-state (~(put by provider-state) service.act new-entry))]
+      ::
+          %remove
+        ?>  (team:title our.bowl src.bowl)
+        =/  entry=(unit provider-entry)  (~(get by provider-state) service.act)
+        ?~  entry
+          ~|("no such service: {<service.act>}" !!)
+        :_  state(provider-state (~(del by provider-state) service.act))
+        %+  turn  ~(tap by clients.u.entry)
+        |=  [who=@p *]
+        ^-  card
+        (leave-client who service.act)
+      ::
+          %client-join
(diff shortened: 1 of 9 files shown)


pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+  ::
+  ++  on-init
+    :_  this
+    [%pass /hark %agent [our.bowl %hark-store] %watch /updates]~
+  ::
+  ++  on-save   !>(state)
+  ++  on-load
+    |=  =old=vase
+    ^-  (quip card _this)
+::    `this(state *state-0)
```
liam-fitzgerald, line 54: remove

liam-fitzgerald (changes requested):
looks good generally, some feedback. Have you considered using `lib/agentio.hoon` could help cleanup some of your card construction

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+::
+|%
++$  card  card:agent:gall
+::
++$  provider-state  (map term provider-entry)
++$  provider-entry
+  $:  notify-endpoint=@t
+      binding-endpoint=@t
+      auth-token=@t
+      clients=(map ship (unit @t))
```
tacryt-socryp, line 13: Can you explain what the (unit @t) is? Having a hard time getting that from the code so far.

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+    |=  [=wire =sign:agent:gall]
+    ^-  (quip card _this)
+    ?+  wire  (on-agent:def wire sign)
+    ::
+    ::  subscription from client to their own hark-store
+    ::
+        [%hark ~]
+      ?+  -.sign  (on-agent:def wire sign)
+          %fact
+        ?.  ?=(%hark-update p.cage.sign)
```
tacryt-socryp, line 187: You can factor out `this` into `:_  this` as it isn't modified in any of the cases.

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+      ==
+    ::
+    ::  subscription from provider to client
+    ::
+        [%notify @ @ ~]
+      =/  who      (slav %p i.t.wire)
+      =*  service  i.t.t.wire
+      ?+  -.sign  (on-agent:def wire sign)
+          %fact
+        ?>  ?=(%notify-update p.cage.sign)
```
tacryt-socryp, line 210: Same deal with this here.

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+      ==
+    ==
+  ::
+  ++  on-arvo
+    |=  [=wire =sign-arvo]
+    ^-  (quip card _this)
+    ?+  wire  (on-arvo:def wire sign-arvo)
+        [%register-binding @ @ @ ~]
+      =/  who=@p   (slav %p i.t.wire)
+      =/  service  i.t.t.wire
```
tacryt-socryp, line 239: Better as `=*`

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+        (fall (rush (@t q.data.mime-data) apex:de-json:html) *json)
+      =/  [sid=@t message=@t]
+        %.  jon
+        %-  ot:dejs:format
+        :~  sid+so:dejs:format
+            message+so:dejs:format
+        ==
+      ::
+      =/  entry=(unit provider-entry)  (~(get by provider-state) service)
+      ?~  entry
```
tacryt-socryp, line 257: You never send any moves, so you can factor out the `:-  ~` to be further up.

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+  [%pass wire %arvo %i %request request *outbound-config:iris]
+::
+++  build-form-data
+  |=  data=(list [@t @t])
+  ^-  @t
+  %+  roll  data
+  |=  [[p=@t q=@t] out=@t]
+  ?:  =(out '')
+    (cat 3 p (cat 3 '=' q))
+  (cat 3 out (cat 3 '&' (cat 3 p (cat 3 '=' q))))
```
tacryt-socryp, line 372: Better to use 

```
%+  rap  3
:~  ...
==
```

than multiple cat calls.

pkg/arvo/app/notify.hoon
```diff
@@ -0,0 +1,449 @@
...
+  [%pass wire %agent [who %notify] %poke %notify-provider-action !>(act)]
+::
+++  give-update
+  |=  =update
+  ^-  card
+  =/  paths=(list path)
+    %+  turn  ~(tap by sup.bowl)
+    |=  [duct =ship =path]
+    path
+  [%give %fact paths %notify-update !>(update)]
```
tacryt-socryp, line 447: Does this unconditionally give every update to every path? Seems questionable, but I haven't looked closely at how you're using it.

pkg/arvo/app/notify.hoon
ixv: A binding token, we set it when the call to register-binding returns, and it is used when we call remove-binding

liam-fitzgerald (approved):

pkg/arvo/lib/agentio.hoon
```diff
@@ -115,6 +115,15 @@
...
   [%give %fact paths cage]
 ::
+++  fact-all
+  |=  =cage
+  ^-  card
+  =/  paths=(list path)
+    %+  turn  ~(tap by sup.bowl)
+    |=  [duct ship =path]
+    path
+  [%give %fact paths cage]
```
Fang-, line 125: If there are no subscribers, the `paths` here is `~`, but that has special semantics in gall, generally used for sending facts just to the subscription that was just opened. Suspect this causes the symptom described in #5138.

PR #5131, opened 2021-07-25
tokenizeMessage: fix urls after inline urls

liam-fitzgerald:


arthyn (approved):
Took me a while to see what was going on here

PR #5133, opened 2021-07-26
slightly smarter bytecode cache clearing

taalhavras:
Previously the entire contents of the bytecode cache were tossed every 1k events (among other reasons). This is an artificial limit, we can make a slightly smarter policy for this case. To achieve this I modified the hashtable so you can specify a freeing function (either when freeing the whole thing or just when trimming). 

notes:
I thought about putting the freeing function in the hashtable itself (so it's specified on construction and we don't have to pass it around everywhere). If people have opinions on this please weigh in - I think I prefer the current approach but I'll defer to those with more experience here.

Currently every 1k events we cut the cache in half - it might be better to just trim it to some constant size.

Testing:
- [x] boot a ship
- [x] run |mass a whole lot


ohAitch:
How long does it take to generate the bytecodes, is it approximately linear
in the size of the result? Don't suppose we like, have hit counters for
what percent of events use each entry?

zalberico:
@joemfb @mcevoypeter if this is something we want to merge I think you're the right ones to review it, if not I can close.

frodwith:
I don't think it's done. Freeing programs isn't simple, particularly because of call site caches (any reference to u3n_progs, really). Unless you do something (probably clear those references), programs will have stale references to programs that are now gone - an issue you don't run into if you just delete all the programs at once. There are probably some other gotchas too, the memory management for programs is particularly tricky (and probably not well designed). The changes need to be reviewed carefully with respect to that rat's nest, and hammered with the memory checker on by someone who is trying to break them.

mcevoypeter:
Thanks for the input, Paul. In that case, I'd be in favor of either leaving the PR open until someone has the time to take it across the finish line or simply close it. @joemfb, thoughts?

zalberico:
Thanks - in the interest of not having open PRs that aren't being worked on that have no clear path to getting released I'm going to close this. If this *is* something we want to scope then I can create an issue for the underlying goal and link to this PR for historical context.

tacryt-socryp:
Extraordinarily depressing that the lack of resourcing of the runtime leads to results like this.

PR #5137, opened 2021-07-28
graph lib: make subset interface work the same as old one did

tacryt-socryp:
Should fix weird notification bug.


tacryt-socryp:
This needs to be pushed out as a hotfix to prevent excessive network traffic @liam-fitzgerald

liam-fitzgerald:
Has it been tested?

liam-fitzgerald (approved):
LGTM, if tested

tacryt-socryp:
Yes, tested

Issue #5138, opened 2021-07-28
Cannot receive DMs


Fang-:
When people try to send me messages over DM, I see the following in terminal scrollback:

```
[%agent-giving-on-system-duct %notify %fact]
```

One of those per message they try to send. The messages they send do not show up in my messages list, or notifications, or anywhere else afaict.        

Looking at scrollback history, I do not see any of these printfs prior to receiving today's OTA (v2.96). Potentially related to #5129 cc @ixv?

```
> +trouble
[ [%base-hash ~[0v1t.n8ehb.qkqcg.eoltg.afj2c.pba9h.rrj3l.r7f3b.l86fh.6elv9.uhale]]
  [%sour-hash 0v1t.n8ehb.qkqcg.eoltg.afj2c.pba9h.rrj3l.r7f3b.l86fh.6elv9.uhale]
  [%home-hash 0v1i.fcp4v.8qcp3.e4pn4.okfkl.al706.r147t.7788d.7cing.fdpd4.8ck7n]
  [%kids-hash 0vr.m11c5.hd54q.68nql.qsq2c.pruo9.no8e4.cr1pl.fkspl.jbk8m.fbek2]
  [ %glob-hash
    ~[
      [/~btc/js/bundle 0v2.ifoe4.fbv35.aigir.66su4.fbspu %done]
      [/~landscape/js/bundle 0v5.48552.hdu43.jrjb7.jd19b.74vpc %done]
    ]
  ]
  [%our ship=~palfun-foslup point='1245952' life=[~ 1] rift=[~ 0]]
  [%sponsor ship=~wanzod point='768' life=[~ 3] rift=[~ 0]]
  [%dopzod ship=~dopzod point='4608' life=[~ 3] rift=[~ 2]]
  "Compare lifes and rifts to values here:"
  "https://etherscan.io/address/azimuth.eth#readContract"
  "  life - getKeyRevisionNumber"
  "  rift - getContinuityNumber"
  ~
]
```

ixv:
I added this code to lib/agentio.hoon that sends a fact on all paths, which seems like it could be related. I have no idea how the system duct could get hooked up to that though.
```
++  fact-all
  |=  =cage
  ^-  card
  =/  paths=(list path)
    %+  turn  ~(tap by sup.bowl)
    |=  [duct ship =path]
    path
  [%give %fact paths cage]
```

can you run `:notify +dbug %bowl`

Fang-:
[...]

PR #5140, opened 2021-07-29
agentio: fix bug with fact-all

ixv:
Should resolve this issue: https://github.com/urbit/urbit/issues/5138

May also require running `|fade %notify` followed `|start %notify`, maybe worth doing that automatically in hood if that's an option.


liam-fitzgerald (approved):
(drop card) is a little cleaner then `?~  card   ~  card.. ` etc. else LGTM

PR #5145, opened 2021-07-29
jets: updates +rip and +rep stub jets to defer to nock

joemfb:
The jets for `+rip` and `+rep` are embarrassingly partial; see #4049 and #1210 for context. This PR updates them to defer to nock, rather than bail non-deterministically in unimplemented cases. Finishing their implementations remains a priority.

/cc @ohAitch


philipcmonk (approved):
LGTM, leaving the logs in place is probably good because these functions are often used explicitly to get the jetted performance

PR #5150, opened 2021-08-04
hoon: remove +ad

philipcmonk:
This reverts d96d50199 because +ad is incredibly opaque, and +ergo's
sitting right there anyway.  It looks like it was intended to abstract
over +endo, +elbo, and +ergo, but only +ergo was every implemented.  I
don't doubt the others could be as well, but then they would be just as
inscrutable.

Also remove a dumb double-compile in mint's `.=` case.

Neither of these should change the semantics of the compiler.


joemfb (approved):
Amazing. LGTM

belisarius222 (approved):
LGTM, lol

PR #5151, opened 2021-08-05
npm: add L2 Roller typescript client

yosoyubik:
Adds a typescript client to interact with the future L2 Roller (see [here](https://documenter.getpostman.com/view/16338962/Tzm3nx7x) for some WIP documentation)

@philipcmonk @liam-fitzgerald

PS: I'll be happy to publish this to npm, but I need to be added to the urbit org there—or someone with access can also do it :)


pkg/npm/roller-api/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "roller-api",
+  "version": "0.1.0",
```
liam-fitzgerald, line 3: `"private": true` to prevent this from automatically being published

liam-fitzgerald:
Is this ready? 
Before merge:
- Please strip out built docs. I'm happy to help setup CI so that we can publish these docs somewhere, I just don't think they should be committed to the repo
- cc: @arthyn this should probably use the new rollup configs, no?
- This should probably straight up target master, given that it doesn't fit cleanly into any of the current `next/*` branches

zalberico:
@liam-fitzgerald is this something still relevant that we want to merge or should I close this?

zalberico:
@tomholford and @yosoyubik - I'm going through issues and this looks like something that can merge, what's blocking it?

yosoyubik:
@zalberico there are a couple of things around this that are still missing. I removed the autogenerated docs, but there should be a CI process to generate them and publish them somewhere—currently they live [here](https://github.com/urbit/roller-rpc-client) and can be accessed on this URL: https://urbit.github.io/roller-rpc-client/index.html but they are manually updated.

I've also published the latest version to NPM— @tomholford, I think Bridge still uses 0.1.15; the latest version (https://www.npmjs.com/package/@urbit/roller-api/v/0.1.16) fixed a typo (https://github.com/urbit/urbit/pull/5151/commits/beefa6a3a87a0d539c3a3e7f44f7161fed2ac519) in the L2Point type.

yosoyubik:
As far as I'm aware, only Bridge uses this package so maybe it should belong in that repository. As of the remaining this, the list of things listed here https://github.com/urbit/urbit/pull/5151#issuecomment-1356726045 still applies.

PR #5156, opened 2021-08-12
graph-store: Handle missing update-logs gracefully in scries

liam-fitzgerald:
If a subscription was kicked before receiving any graph-updates on it,
then the attempt to construct a valid resubscribe path would fail, as
the scry crashes. Fixes this by handling the missing update-log case
correctly.


tacryt-socryp (approved):
Looks good to me.

PR #5159, opened 2021-08-12
khan: control plane

ryjm:
Send tasks as jammed nouns to arvo through a unix socket. For now, these tasks are specific, pre-defined ones (i.e to request a code request and to retrieve a code). Could be more generic in theory. 

To try this out, jam and save a task (can define your own in lull.hoon and khan.hoon) in the dojo , and then send it through the socket with `nc`. 

Some initial WIP documentation, with TODOs (in addition to a couple scattered around the code):

- Spec
	- Vere/Khan
		- TODO Refactor to remove socket references in type names where they are not helpful
		- a `SockAPI` created in `startServ`
			- [[SockApi]]
			- Created with `watch` , which takes an `IO` and a `SockConf`
				- [[SockConf]]
					- `FilePath` and `SockType`
					- [[SockType]]
						- [[STGeneric]]
							- No effect handler, just return on the socket.
						- [[STReq]]
							- [[SReqApi]]
								-
								  ``` javascript
								  data SReqApi = SReqApi
								    { sReq :: Ship -> Word64 -> STM ()
								    -- , sKil :: Ship -> Word64 -> STM ()
								    }
								  								  
								  ```
								- Holds a function `sReq` used to handle effects. This is used to send a [[khan-task]] to arvo.
								- TODO Figure out `sKil`
				- Delegates to [[watchSocket]]
					- The main loop handling incoming messages on a [[Socket]] . This is done through the [[Net.Socket/accept]] function, which returns a new socket on which to send data back.
					- Takes a [[Net.Socket]] and the [[SockType]]
		- The [[SockApi]] is then used to create a `Sock` holding the socket id, the api and the conf corresponding to the socket.
			- DONE Do [[Sock]] , [[SockConf]] , and [[SockApi]] need to the same FilePath reference?
				- Probably only needs to be set in one place.
				- The one in [[Sock]] is used to remove the file, and the one in [[SockConf]] is used to open the socket.
					- Changed to only use the one in [[SockConf]]
		- Driver
			- Invoked in the Pier module.
			- `khan'` : Creates a `DriverApi` with a `SocketEf` along with a set of events.
				- [[DriverApi]]
					- `DriverApi` consists of an [[STM]] of a [[RunReq]] and a function that runs when an event is received. If the [[SockType]] is [[STReq]] then this will be in [[SReqApi]] .
						- The first argument is the Event Source and is created by applying the `RRWork` [[RunReq]] to the read queue.
			- `khan` : Creates the initial events and the function that runs on every event.
				- Takes a `plan` and a `stderr`
					- `plan` : [[STM]]
						- `writeTQueue` - writes to a FIFO queue.
					- `stderr` : Text to [[RIO]]
				- Returns a function that wrapped in [[RAcquire]] that responds to a [[SocketEf]] .
					- The [[Net.Socket]] isn't opened until a [[SESetConfig]] effect is seen.
					- A response to a [[SocketEvFyrd]] is expected to come in as a [[SocketResponse]] .
						- TODO Better naming
	- Effects
		- [[SocketEf]]
		- [[router]] loop manages incoming effects.

### DIFF
--- a/nix/default.nix
+++ b/nix/default.nix
@@ -36,7 +36,7 @@ let
         };
       };
     })
-
+    (import ./overlays/haskell.nix)
     # General unguarded (native) overrides for nixpkgs.
     (import ./overlays/native.nix)
 
--- a/nix/overlays/haskell.nix
+++ b/nix/overlays/haskell.nix
@@ -1,0 +1,111 @@
+final: prev:
+let
+  compiler-nix-name = "ghc884";
+  project = prev.haskell-nix.stackProject {
+    compiler-nix-name = "ghc884";
+    index-state = "2020-09-24T00:00:00Z";
+
+    # This is incredibly difficult to get right, almost everything goes wrong.
+    # See: https://github.com/input-output-hk/haskell.nix/issues/496
+    src = prev.haskell-nix.haskellLib.cleanSourceWith {
+      # Otherwise this depends on the name in the parent directory, which
+      # reduces caching, and is particularly bad on Hercules.
+      # See: https://github.com/hercules-ci/support/issues/40
+      name = "urbit-hs";
+      src = ../../pkg/hs;
+    };
+
+    modules = [{
+      # This corresponds to the set of packages (boot libs) that ship with GHC.
+      # We declare them yere to ensure any dependency gets them from GHC itself
+      # rather than trying to re-install them into the package database.
+      nonReinstallablePkgs = [
+        "Cabal"
+        "Win32"
+        "array"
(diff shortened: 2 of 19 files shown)


zalberico:
Closing this since I don't think there's a path to shipping it and nobody is working on it.

ryjm:
getting ptsd from looking at all this haskell.

just for future reference, this was the implementation of the socket side of khan with the corresponding vane. haskell runtime was discontinued and the rest was pulled into #5420.

PR #5160, opened 2021-08-12
landscape, btc-wallet: upgrade build process 

liam-fitzgerald:


liam-fitzgerald:
cc: @finned-palmer

pkg/btc-wallet/src/channel.js
```diff
@@ -0,0 +1,290 @@
+export default class Channel {
```
arthyn, line 1: Why does the BTC wallet have it's own implementation of this instead of using http-api?

arthyn (approved):

liam-fitzgerald:
Historical reasons mainly

Issue #5161, opened 2021-08-12
Debugging scot/scow and solving memory leaks


sigilante:
I have taken the `++slaw`/`++scot` memory leak jet bounty.

I have a couple questions about the current code, which is commented out by `tree.c`.

---

First, the Hoon, the jets, and the actual usage all deviate from each other:

| Mode | `%da` | `%p` | `%t` | `%ta` | `%tas` | `%ud` | `%uv` | `%uw` | `%ux` |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| Hoon `++slaw` | √ | √ | (default) | √ | √ | √ | √ | (default) | √ |
| Jetted `u3qe_slaw` | √ | √ | `~` | `~` | √ | √ | `~` | `~` | `~` |
| Slaws in use |  √ | √ | √ | - | √ | √ | - | √ | - |

So we have a bit of an odd situation with mismatch, particularly as the jet returns null `~` rather than a default slay.  What's the right way to synchronize these?

---

Secondly, no combination of `-g` and Valgrind seem to work on contemporary Urbit for me.  That is, `-g` by itself hangs before the Dojo prompt, Valgrind causes the LMDB to fail to load, and both together still hang.  Is this a known bug or do I have something going on?  I have used 1.0, 1.2, and 1.5 binaries with the same results.

---

The jet code for `++scow` has a lot of magic numbers in it, like this:

```c
static
c3_y to_w_digit(c3_y tig)
{
  if (tig == 63) {
    return '~';
  } else if (tig == 62) {
    return '-';
  } else if (tig >= 36) {
    return 29 + tig;
  } else if (tig >= 10) {
    return 87 + tig;
  } else {
    return '0' + tig;
  }
}
```

ASCII 87 is `W` so I don't think that's what's going on.  Anyone remember what these were for?

Also referencing #4804 and #4767.

sigilante:
The particular error I get is:

```
lmdb: failed to open event log: Invalid argument
disk: failed to initialize database
pier: stay: init fail
```

when using Valgrind, which is delightfully ambiguous.  `strace` on `valgrind urbit` etc. hasn't given me any insight either.  (I've been digging around in LMDB docs to see what it thinks is invalid without yielding fruit yet.)

The line that yields the error (for Valgrind) is this one, `lmdb.c` line 67:

```c
ret_w = mdb_env_open(env_u, pax_c, 0, 0664)
```

[...]

PR #5163, opened 2021-08-16
++scot/++slaw jets redux

sigilante:
Related issue:  #5161 

I am rewriting these to not use `tape`s internally but to directly operate on `cord` atoms.

Status:
- [x] `++scot` `@da` as cord
- [x] `++scot` `@p` as cord
- [x] `++scot` `@ud` as cord
- [x] `++scot` `@uv` as cord
- [x] `++scot` `@uw` as cord
- [ ] `++slaw` `@da` as cord
- [ ] `++slaw` `@p` as cord
- [ ] `++slaw` `@ud` as cord
- [ ] `++slaw` `@tas` as cord

### DIFF
--- a/pkg/urbit/jets/e/scow.c
+++ b/pkg/urbit/jets/e/scow.c
@@ -1100,11 +1100,13 @@ u3we_scot(u3_noun cor)
   switch (mod) {
     case c3__da:
       //return __print_da(cor, atom);
-    return u3_none;
+      mpz_clear(u_mp);
+      return u3_none;
       tape = _print_da(cor, atom);
       break;
 
     case 'p':
+      mpz_clear(u_mp);
       return u3_none;
       u3r_mp(u_mp, atom);
 
--- a/pkg/urbit/jets/e/scow.c
+++ b/pkg/urbit/jets/e/scow.c
@@ -4,6 +4,8 @@
 #include "all.h"
 
 #include <ctype.h>
+#include <stdio.h>  // for fprintf
+#include <math.h>   // for ceil/floor
 
 static
 c3_y to_digit(c3_y tig)
@@ -54,9 +56,18 @@ _y_co_two(c3_w src, c3_y* a, c3_y* b)
   *a = to_digit(src / 10);
 }
 
-#define DOT 0x2e
-#define HEP 0x2d
-#define SIG 0x7e
+//
+static
+u3_noun
+_add_year(c3_w year, u3_noun out)
+{
+  while (year > 0) {
+    out = u3nc(to_digit(year % 10), out);
+    year = year / 10;
+  }
+
+  return out;
+}
 
 static
 u3_noun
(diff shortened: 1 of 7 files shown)


sigilante:
As of this commit, ++scot and ++scow have been revised to use a cord-based representation internally rather than tapes.

++slaw needs some work still.

I will ping when ready for review.

sigilante:
> the http driver was unable to parse the wire on that %set-config effect

Current CI log indicates failure on strange mole HTTP issue.  _Most likely_ this means scot is parsing some case wrong, so I need to compose a bunch of test cases...

joemfb:
@sigilante, those calls to `+slaw` in http.c use `@ud` and `@uv`. And it should be pretty easy to get some test cases by logging the arguments in `_reck_mole()` (just use `u3m_p("san", san);`)

sigilante:
The scot jets seem okay.

The modified scow jets have some subtle memory leak or something going wrong.

```
built   /tests/sys/zuse/ordered-map/hoon
built   /tests/sys/zuse/title/hoon
>   test-cite: took 12644µs
leak: 0x2168fcdec mug=0 (marked=0 swept=1)
code: %scow
    size: B/44
    data: (null)
leak: 0x216902858 mug=0 (marked=0 swept=1)
code: %scow
    size: B/64
    data: (null)
leak: 0x216903ecc mug=0 (marked=0 swept=1)
...
Assertion '(0 == leq_w) && (0 == weq_w)' failed in noun/allocate.c:2263

bail: oops
bailing out
pier: serf unexpectedly shut down
```

Can someone put another pair of eyes on these to see if they have an idea why this is leaking?  The previous logic was tape-first; i.e. scot converted cords from tapes.  The current logic reverses that, and scow converts tapes from cords while scot is cord-only.

(Slaw set aside for now, it's not coming from that.)

pkg/urbit/jets/e/scow.c
```diff
@@ -188,202 +753,260 @@ _print_p(u3_atom cor, u3_atom p)
...
       return u3_none;
   }
+
+  if (cord == u3_none) {
+    return cord;
+  }
+  u3_noun ret = u3qe_trip(cord);
+  //u3z(mod);
+  //u3z(atom);
+  return ret;
```
joemfb, line 974: This leaks `cord` -- `u3qe_trip()` has `RETAIN` refcount semantics (we this because it's prefixed with `u3q`), but `cord` is an allocation from this function, so this function is still responsible for it.

sigilante:
Looks like another part of the trouble was ++cite being changed and thus the test suite being different here.

sigilante:
Status update:  the memory corruption I'm seeing is the same as the original scot/slaw jets, so I think I'm going to have to back all the way out and implement/copy in from scratch to see if I can localize.  These don't show up in regular operation but do in the test suite, so it's a rather slow loop to change and iterate.

sigilante:
Is there a way to list everything that's being refcounted in u3 and its current state?  Bisection tests and one-line-at-a-time aren't really getting me anywhere -_-

ohAitch:
`|mass` does some kind of gc sweep, you might be able to hook into that? Possibly under a time-travel debugger so you can run it at interactively specified points

joemfb:
@sigilante I don't see any memory corruption in CI, just some leaks. If you have other errors/crashes/&c, I'd like to see those.

There is a mark and sweep garbage collector that can help track down leaks. We run it automatically in CI, and programmatically in the vere unit tests -- see `u3m_grab(u3_none)` in `pkg/urbit/tests/jet_tests.c`.

`u3m_grab()` takes a varargs lists of additional gc roots (global state is already tracked), and asserts if any leaks are found. When build with memory debugging enabled, allocations include a second refcount and a "code" slot for tracking provenance (change/restore the code with `u3a_lush()`/`u3a_lop()`).

You could isolate a single test case to be run repeatedly, and even gc multiple times inside a single function (as long as you track all your roots). And you could use different codes for different blocks/sections of the function.

All that being said, I'd like to reiterate my advice in https://github.com/urbit/urbit/issues/5161#issuecomment-901426979. These are long, fairly complicated printing/parsing routines, and there's no need for their unobservable intermediate states to use nouns. That approach will involve much less refcounting.

sigilante:
The main reason to stick close to the noun functions is if I end up doing something that inadvertently overflows 31 bits or the like.  I wrote up an approach just now that _shouldn't_.

At this point I am adding `math.h` and `-lm` for `ceil`/`floor`.  Is there a reason these haven't been linked before?

joemfb:
Outside of the floating point jets, there's hardly any use of floats in the codebase. It shouldn't be a problem to add `-lm`.

My recommendation is to just use normal C types for these internals (and maybe GMP directly), avoiding those kinds of representation concerns except at the edges. `u3s_sift_ud_bytes()` is an example of this for parsing `@ud`, and `u3qe_{en,de}_base16()` are similar examples for vanilla hex.

sigilante:
Minor update:  I have versions of ud, uv, and uw which work without leaking.  Now to copy that approach to da and p.

sigilante:
```
>   test-slaw-patuw: took 23699µs
OK      /sys/hoon/slaw/test-slaw-patuw
>   test-slaw-patuv: took 15873µs
OK      /sys/hoon/slaw/test-slaw-patuv
>   test-slaw-patud: took 4027µs
OK      /sys/hoon/slaw/test-slaw-patud
>   test-slaw-patp-star: took 2570µs
OK      /sys/hoon/slaw/test-slaw-patp-star
>   test-slaw-patp-planet: took 2590µs
leak: 0x21690339c mug=0 (marked=0 swept=1)
code: %slaw
    size: B/32
address 0x2ee937004 out of loom!
Assertion '0' failed in noun/events.c:128
```

I'm not even _using_ the `@p` jet; it just returns `u3_none`.  The test in question which starts then fails:

```
++  test-scot-patp-planet
  ;:  weld
    %+  expect-eq
      !>  ~.~dapnep-ronmyl
      !>  (scot %p ~dapnep-ronmyl)
  ::
    %+  expect-eq
      !>  ~.~dapnep-ronmyl
      !>  (scot %p 0x1.0000)
  ::
    %+  expect-eq
      !>  ~.~dostec-risfen
      !>  (scot %p ~dostec-risfen)
  ::
    %+  expect-eq
      !>  ~.~dostec-risfen
      !>  (scot %p 0xffff.ffff)
  ==
```

I don't even have a question at this point.. for a given code attempt the behavior on my system may be "normal" (no noted memory leaks), and then the CI crashes here; or vice versa.

jalehman:
This is being redone, closing per @sigilante

PR #5164, opened 2021-08-17
Leap: improving search capability

arthyn:
This lets you navigate the leap bar like a real combobox, you can also now type in an app link such as: `~nocsyx-lassul/groups` to navigate directly to a treaty.


pkg/grid/src/nav/search/Providers.tsx
```diff
@@ -29,18 +47,25 @@ export const Providers = ({ match, history }: ProvidersProps) => {
...
   useEffect(() => {
-    if (searchInput) {
-      handleSearch(searchInput);
+    if (results) {
+      useLeapStore.setState({
+        matches: results.map((p) => ({ value: p.shipName, display: p.nickname }))
+      });
     }
-  }, [searchInput]);
+  }, [JSON.stringify(results)]);
```
liam-fitzgerald, line 56: yeesh, is there no way to ensure that results isn't recreated every render?

liam-fitzgerald (approved):
LGTM! some minor nits

pkg/grid/src/nav/search/Providers.tsx
arthyn: this might have been before I had `useMemo` on results. I tried it just now without the stringify and it seems fine.

PR #5157, opened 2021-08-12
urcrypt: integrate

joemfb:
This PR integrates `urcrypt`: @frodwith's new urbit crypto library. This project encapsulates urbit's standard crypto dependencies (openssl, secp256k1, ed25519, scrypt, libaes_siv) and custom variants (argon2u, ge-additions) into a singular library, suitable for use in other projects. For more details, see the [urcrypt README](https://github.com/urbit/urbit/blob/f5676a9b0fe77d1bd863a5d47ac2633abe6ccec0/pkg/urcrypt/README.md).

The development of urcrypt necessarily entailed significant refactoring and of vere's crypto jets; notable benefits include normalized logging and error handling, with new u3 apis for safe noun deconstruction. Many bugs were fixed, missing jet configuration was rectified (scrypt), and the secp256k1 implementation was significantly improved in both the jet and the hoon. All credit to @frodwith -- I am but a humble integrator.

------

Since `urcrypt` was developed on a branch predating #3806, and this PR is targeting a release-candidate branch that includes #4675, most of my integration work has revolved around the build system. Statically-linking vere with urcrypt involves resolving transitive library dependencies, which lead me to revise our use of `pkg-config`.

- @brendanhay, please review the `nix` and `pkg/urbit/configure` changes for sanity
- @locpyl-tidnyd, please take a look at `nix/sources*`, `pkg/urbit/compat/` and `pkg/urbit/configure`
- @frodwith, please review for completeness and intent

I owe a review of the `u3` changes.

------

TODO:

- [x] this branch includes binaries in its history (`ed25519.dll`, non-lfs pill)
- [x] `pkg/urcrypt` needs an autotools appropriate .gitignore (perhaps https://github.com/github/gitignore/blob/master/Autotools.gitignore)
- [ ] urcrypt could use a linkable version string/symbol (for `urbit -R`)
- [x] consider re-including urcrypt's vendored dependencies with `git sub-tree`
- [ ] the mingw build downloads the ivory pill as html (githubs raw download behavior seems to have a recently changed)
- [ ] the mingw build cache key for local builds isn't precise enough (doesn't affect CI, maybe out of scope)

PR #5170, opened 2021-08-19
build: update secp256k1, upstream libaes_siv

joemfb:
This PR is updates dependencies as prep for #5157 (which I'm splitting up for review/posterity).

Notes:

- the libaes_siv cmakefiles.txt patch should be upstreamed later, if possible
- the static/cross build of libaes_siv requires `inherit (pkgsNative) cmake;`

I'd like to know why putting `cmake` in `nativeBuildInputs` wasn't enough, but it kept trying to cross-compile `cmake` itself, which interacted poorly with the static stdenv (??) putting `--enable-static` and `--disable-shared` all over the place, often duplicated. The nixpkgs cmake derivation has a `--` in its `CFLAGS`, and one of those args was interpreted as a relative path ... Of course, If the answer is not simple/obvious, it doesn't need to block this PR.


brendanhay (approved):

brendanhay:
> I'd like to know why putting cmake in nativeBuildInputs wasn't enough

Without investigating myself - I assume since we use `<overlay>.callPackage` that the `cmake` argument to your derivation will be supplied from said `<overlay>`, which in the case of `pkgsStatic` will result in passing a yet-to-be-cross-compiled `cmake` to  `nativeBuildInput`s?

PR #5171, opened 2021-08-19
urcrypt: merge

joemfb:
This PR includes a merge of @frodwith's urcrypt library onto #5170. Source-level conflicts have been resolved, but the build is broken. Patches, review resolutions, &c will be handled in #5157 (which is being split up and will target this branch).


joemfb:
I've redone this merge, as I had accidentally checked in build results from local testing.

botter-nidnul:
Hmm, getting an error when trying to build urbit-v1.6-rc1 for aarch64:

```
libtool: compile:  aarch64-unknown-linux-musl-gcc -DHAVE_CONFIG_H -I. -I./argon2/include -DARGON2_NO_THREADS -Wno-unused-value -Wno-unused-function -g -O2 -c argon2/src/opt.c -o argon2/src/libargon2_la-opt.o
In file included from argon2/src/opt.c:26:
argon2/src/blake2/blamka-round-opt.h:23:10: fatal error: emmintrin.h: No such file or directory
   23 | #include <emmintrin.h>
      |          ^~~~~~~~~~~~~
compilation terminated.
make[1]: *** [Makefile:945: argon2/src/libargon2_la-opt.lo] Error 1
make[1]: Leaving directory '/build/urcrypt'
```

I'm wondering if it's just pulling in extraneous dependencies (which I should be able to turn off in configuration, I think), or if urbit is actually trying to use some SSE instructions that it's never used before? Might need a translation layer from SSE to NEON if that's the case.

I'll look into the underlying cause further, later today or tomorrow. Just wondering if these changes are deep or shallow, and might be interacting with architecture extensions.

joemfb:
@botter-nidnul, I think this might be as simple as updating the following line to point to `ref.c` instead of `opt.c`:

https://github.com/urbit/urbit/blob/d3e7dc1aa8bbe006a9754be82f4c475faa268daf/pkg/urcrypt/Makefile.am#L93

If so, that should be fine for now (and please PR it against `release/next-vere`). At some point, it'd be nice to do that in a platform-specific way so as to use the optimized version on x86_64, but I don't think that's pressing at all.

botter-nidnul:
Yeah, that works. Doesn't look hard to do that conditionally with a `if ARCH_X86_64, libargon2_la_SOURCES += argon2/src/opt.c` in Makefile.am. Will put in a PR doing something like that in a bit...

botter-nidnul:
submitted PR: #5248

PR #5173, opened 2021-08-19
grid: Help and Support section

arthyn:
Implements the help section, not sure about the `tlon.io/help` link that doesn't seem to exist, just replaced it with a link to our support email for now.


pkg/grid/src/nav/Help.tsx
```diff
@@ -9,12 +10,65 @@ export const Help = () => {
...
+          <p>
+            <a
+              className="text-blue-400"
+              href="web+urbitgraph://group/~bitpyx-dildus/tlon-support"
+              target="_blank"
+              rel="noreferrer"
+            >
+              Join Tlon Support
+            </a>
+            <span className="mx-2 text-black">or email</span>
```
liam-fitzgerald, line 49: Cc: @urcades tlon support is only for hosting customers, yes?

liam-fitzgerald (changes requested):
Nit about visibility of Tlon support group for non hosting customers, else LGTM

pkg/grid/src/nav/Help.tsx
arthyn: Hmm if it is, how would we be able to differentiate people here to show/hide?

PR #5176, opened 2021-08-23
docket: final pass

liam-fitzgerald:
This PR essentially finalises a majority of the userspace backend for software distribution

## Architecture
- `%docket` has now been split into `%docket` and `%treaty`. `%docket` handles the lifecycle of a local desk. `%treaty` handles discovering and fetching descriptions of foreign desks. 
- `%treaty` subscribes to provider ships to discover packages. Provider ships do not have to host the packages that they expose for discovery, allowing for a bare minimum app-store like experience. By default, we subscribe to our sponsor. 
- Significant refactor and cleanup
- Dockets react to suspension and revival
- Adds support for non-SPA web bundles. If a `site` path is specified instead of the usual `glob` & `base`, then the docket links directly to `site`
- Parameterises the transport for the retrieval of glob bundles( TODO: actually use this)
- Adds images!
- Renames pkg/grid to pkg/garden (TODO: better name)

## Other changes
- Fixes some latent bugs in nu-kiln, including:
    -  Subscriptions not being given on the correct path
- Adds marks so that kiln is usable through eyre channels & scries


Fang-:
This has since been released. Closing. (:

PR #5180, opened 2021-08-24
landscape: altered RemoteContent regexes for images, video, and audio to include period literal.

litmus-ritten:
This prevents URLs ending in 'mov', 'ogg', etc. from rendering as empty video/audio, allowing people to learn about Isaac Asimov and William Rees-Mogg. Tested locally, seems to work.

Guessing @urcades is a reasonable maintainer to tag?

### DIFF
--- a/pkg/interface/src/views/components/RemoteContent/index.tsx
+++ b/pkg/interface/src/views/components/RemoteContent/index.tsx
@@ -38,10 +38,10 @@ export interface RemoteContentProps {
 }
 
 export const IMAGE_REGEX = new RegExp(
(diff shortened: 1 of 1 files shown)


liam-fitzgerald (changes requested):
LGTM! Can you squash your commits so that the only commit in this PR is the one that actually performs the change? 
Thanks for your contribution!

litmus-ritten:
@liam-fitzgerald Done, I think, sorry about that.

liam-fitzgerald (approved):

PR #5181, opened 2021-08-24
metadata-store: Handle removes on kick correctly

liam-fitzgerald:
Previously, if an %initial-group fact had less associations than the agent did, we ignored those missing facts. This causes channels to not remove their metadata correctly. Addresses this issue both in %metadata-store and also in the associated reducer


ryjm (approved):
LGTM

PR #5184, opened 2021-08-25
docket: add web interface for glob uploading

Fang-:
As discussed with @belisarius222 and @ryjm yesterday, we want to make it easy to add glob data into ships without going through clay.

Other options include shipping an external script for sending the files into the ship (either eyre or kahn), and/or reviving `.urb/get`. The web interface here seems like the cheapest, least issue-prone, short-term option.  

@liam-fitzgerald please take a look at the docket behavior this triggers,

https://github.com/urbit/urbit/blob/72955838f214d00bb7b85a62bf14fdec619e0710/pkg/garden/app/docket.hoon#L391-L396

as well as the error cases that it tries to catch.

https://github.com/urbit/urbit/blob/72955838f214d00bb7b85a62bf14fdec619e0710/pkg/garden/app/docket.hoon#L380-L384

I have tested this in the sense that uploading works and error cases are caught, but I haven't tried using or sharing the resulting glob yet.

Would also love to get feedback on the terminology here ("globulator" is obviously temporary), as well as what instructions/direction (if any) should be displayed on the upload page.

(Note that the `/lib/multipart` here is copied straight from a personal project. It's been tested in the wild already, and seems to handle the multi-file upload case here without issue, even when including images and other binary files.)

If we think the upload page's logic is too big, we can probably factor it out into a separate file.

### DIFF
--- a/pkg/garden/app/docket.hoon
+++ b/pkg/garden/app/docket.hoon
@@ -413,15 +413,20 @@
     :-  desk
     ::  all submitted files must be complete
     ::
-    ::TODO  do we want to ignore .DS_Store and other such files?
     ?.  =('glob' name)  [glob (cat 3 'weird part: ' name) err]
     ?~  file            [glob 'file without filename' err]
     ?~  type            [glob (cat 3 'file without type: ' u.file) err]
     ?^  code            [glob (cat 3 'strange encoding: ' u.code) err]
     =/  filp            (rush u.file fip)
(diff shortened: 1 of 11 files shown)


ryjm:
I somehow completely missed that you had already done this. I'll see if I can get this working with my changes, it will make testing a lot easier.

Fang-:
Merged out of band.

PR #5186, opened 2021-08-27
clay: fetch blobs from sponsor

liam-fitzgerald:
Attempts to fetch clay blobs from sponsor, failing over to the actual host.


philipcmonk (changes requested):
One more thing I realized is that now if your sponsor is offline, you can't download anything, which doesn't feel great.  Anything that threatens OTAs is a significant risk.

Ideally it would time out, but maybe the quickest solution is a little card that just resets all `src`s in active update-states to be the source ship.  That way if something goes wrong we can say "run `|pass ...` if your sponsor is offline".

philipcmonk:
This has turned out to be more difficult that we thought to make robust, and it hasn't proven necessary yet.  Content distribution should resolve this more generally.

PR #5188, opened 2021-08-27
notify: cleanup %watch-not-unique cases

ixv:
Stops some noisy errors from printing, doesn't change any behavior beyond that


liam-fitzgerald (approved):
LGTM, should target release/next-userspace though

PR #5189, opened 2021-08-28
|ames-prod reset congestion control

belisarius222:
Sometimes Ames waits for a couple minutes before re-sending a packet.  This is annoying, especially when debugging, so I added an `|ames-prod` generator that resets congestion control for all flows on the specified list of ships, or all ships if no argument.  This causes at least one packet to be re-sent on each open flow to each peer, and resets congestion control to initial values.

We should also fix Ames so it no longer has this problem, but sometimes you just want to tell the computer to send the packet already.

cc @philipcmonk, you might want to take quick lookover here too.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2615,7 +2615,7 @@
     |-  ^+  packet-pump
     ?:  =(0 sot)  packet-pump
     ?:  =(~ liv)  packet-pump
-    =^  hed  liv  (pop:packet-queue live.state)
+    =^  hed  liv  (pop:packet-queue liv)
     =.  packet-pump  (give %send (to-static-fragment hed))
     $(sot (dec sot))
   ::  +on-wake: handle packet timeout
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -654,12 +654,14 @@
 ::    %memo: packetize and send application-level message
 ::    %hear: handle receipt of ack on fragment or message
 ::    %near: handle receipt of naxplanation
+::    $prod: reset congestion control
 ::    %wake: handle timer firing
 ::
 +$  message-pump-task
   $%  [%memo =message-blob]
       [%hear =message-num =ack-meat]
       [%near =naxplanation]
+      [%prod ~]
       [%wake ~]
   ==
 ::  $message-pump-gift: effect from |message-pump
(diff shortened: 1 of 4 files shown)


philipcmonk:
Nice.  FWIW, `%stir` in theory is meant for these kinds of commands.  It dates back to a time when adding tasks was harder (bc explicit agent upgrades in gall), but it's not unreasonable to have a generic "debug/development" task.  Its argument should probably be a noun, though.

belisarius222:
Note: this should be tested more before merging. I got some crashes on my test moon. I think they're from an earlier version of this code that broke the contract with Behn, but better to be sure.

—
~rovnys-ricfer
https://urbit.org

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2566,8 +2595,29 @@
...
       %done  (on-done message-num.task)
       %wake  (on-wake current.task)
+      %prod  on-prod
       %halt  set-wake
     ==
+  ::  +on-prod: reset congestion control, re-send packets
+  ::
+  ++  on-prod
+    ^+  packet-pump
+    ?:  =(~ next-wake.state)
```
eamsden, line 2605: Why is this not just `?~  next-wake.state`

belisarius222: because of hoon's TMI problem -- if you do `?~`, then within the expressions that comes after it, the type of the argument is ~ in one branch, and the non-null type in the other branch (in this case, `_[~ u=*@da]`) .  This is fine unless you want to "reassign" a new value of the other subtype to .next-state later on, which will nest-fail.  This function does this by calling the +give function and overwriting the whole |packet-pump core with its result, which might contain either a null or non-null .next-wake.

So instead, we can use `=(~ next-wake)`, which does not specialize the type and thereby avoids the TMI problem.

There's been discussion of adding a rune that would be like `=.` but would allow an LHS field's type to be expanded, which would obviate this kind of TMI, since assignment would just expand the .next-wake in the subject back to either null or non-null instead of nest-fail'ing.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2576,7 +2626,6 @@
     ::  assert temporal coherence
     ::
     ?<  =(~ next-wake.state)
-    ?>  (gte now.channel (need next-wake.state))
```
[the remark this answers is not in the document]
belisarius222: That was the idea, although since I made some changes, I wonder if I could reinstate this assertion -- +on-prod should actually lawfully reset the .next-wake now, instead of the original behavior, which updated state but then threw it away.  Unsurprisingly, that was easy to get wrong, so I added a new codepath in the packet-pump for +on-prod that doesn't rely on hacking +on-wake the way the original one did.

I'll try reinstating the assertion and seeing if it still works.

PR #5199, opened 2021-09-02
spider: store desk for http response

liam-fitzgerald:
Fixes spider to scry the mark conversion from the correct desk

### DIFF
--- a/pkg/arvo/app/spider.hoon
+++ b/pkg/arvo/app/spider.hoon
@@ -18,17 +18,26 @@
   $:  starting=(map yarn [=trying =vase])
       running=trie
       tid=(map tid yarn)
-      serving=(map tid [@ta =mark])
+      serving=(map tid [@ta =mark =desk])
   ==
 ::
 +$  clean-slate-any
   $^  clean-slate-ket
   $%  clean-slate-sig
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/spider.hoon
```diff
@@ -193,6 +204,23 @@
...
+      ?:  ?=(%3 -.old)
+        `old
+      :-  ~
+      :*  %3
+        starting.old
+        running.old
+        tid.old
+        %-  ~(gas by *(map tid [@ta =mark =desk]))
+        %+  turn  ~(tap by serving.old)
+        |=([=tid id=@ta =mark] [tid id mark q.byk.bowl])
```
belisarius222, line 221: We should remember to migrate HTTP bindings for threads to their new desks during the dist OTA.

belisarius222 (approved):
LGTM

Fang- (approved):
This checks out.

pkg/arvo/app/spider.hoon
liam-fitzgerald: We don't necessarily need to do this, as all threads called via HTTP resolve in a single event.

Issue #5198, opened 2021-09-01
http-api: allow authentication with existing browser cookies and redirect to login page


eamsden:
**Describe the bug**

Presently the `http-api` package cannot authenticate using an existing browser session cookie. This means, for instance,
that an application using http-api must provide its own login form, and must require authentication every time it was opened,
even if it was opened in a new tab from an existing Landscape session. Further, the user must supply the ship name.

**Expected behaviour**
http-api should provide a method to load an existing session cookie, and redirect to a ship-hosted login page if one is not
present or it is invalid.

**Additional context**
This is presently a problem for the WebRTC work, which will have a backlink to this issue.

**Notify maintainers**
If you happen to know who the appropriate maintainers are, consider mentioning them with an @ here.  You may want to use `git blame` to see who has last touched any relevant code.
@liam-fitzgerald @tylershuster

eamsden:
HTTP cookies don't respect port numbers, so if multiple ships are running on the same host it's possible to have cookies for multiple ships with no way to tell which one is authenticating for our ship.

Landscape figures out the ship by loading a special case path from file-server:

https://github.com/urbit/urbit/blob/fa894b9690deae9e2334ccec5492ba90cb0b38f9/pkg/arvo/app/file-server.hoon#L202-L209

If this JS is loaded and evaluated then the `window.ship` property is set to the ship name. It is almost certainly better to use this
approach in-browser rather than trying to parse cookies, as nothing disambiguates which session cookie is for the current ship.

PR #5202, opened 2021-09-04
http-api: add method to try to initialize from existing cookie

eamsden:
(and session.js context)

(draft because not yet tested, would appreciate style/architecture comments)

Implements #5198

Provides a static method to initialize an `Urbit` airlock object based on the existing session cookies in a browser, and the `window.ship` property set by the dynamically generated `session.js`, thus allowing applications using the HTTP api to avoid
requiring a separate login step when loaded from a ship in an existing browser session.

It is intended that browser applications using react will add a static script tag

```html
<script src="/~landscape/js/session.js"></script>
```

This mechanism is currently used by the bitcoin wallet and Landscape to dynamically determine the ship name. 
An application which includes this script tag will then be able to invoke `await Urbit.tryExistingSession()` to receive an airlock object, which will connect to the `window.location.origin` as the base URL for the airlock. If either of the necessary properties (`window.location.origin` and `window.ship`) are undefined, or the context is not a browser, or if the initialization of the event source throws an exception, then `Urbit.tryExistingSession()` will throw an exception.

### DIFF
--- a/pkg/npm/http-api/src/Urbit.ts
+++ b/pkg/npm/http-api/src/Urbit.ts
@@ -138,7 +138,7 @@ export class Urbit {
    * Try to use an existing session, for browser use only.
    *
    * This method will check that `window.ship` is set, and try to connect to the airlock
-   * at window.location.origin as window.ship.
+   * as window.ship.
    * (Generally one should include `<script src="/~landscape/js/session.js"></script>`
    * in e.g. the index.html file of a React app, in order to set window.ship).
    * If so, it will attempt to poke and initialize the event source in the same manner as 'authenticate'.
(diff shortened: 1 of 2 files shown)


pkg/npm/http-api/src/Urbit.ts
```diff
@@ -134,6 +134,36 @@ export class Urbit {
...
+    if(!isBrowser) {
+      throw('tryExistingSession is for browser use only');
+    }
+    if(typeof window.location.origin === 'undefined') {
+      throw 'Cannot try existing session, window.location.origin is undefined';
+    }
+    if(typeof (window as any).ship === 'undefined') {
+      throw 'Cannot try existing session, window.ship is undefined';
+    }
+    const airlock = new Urbit(window.location.origin);
```
liam-fitzgerald, line 157: can just be `const airlock = new Urbit('');` and also remove the origin check

pkg/npm/http-api/src/Urbit.ts
```diff
@@ -78,7 +78,7 @@ export class Urbit {
 
   onOpen?: () => void = null;
 
-  /** This is basic interpolation to get the channel URL of an instantiated Urbit connection. */
+  /** This is basic interpolation to get the cpokehannel URL of an instantiated Urbit connection. */
```
liam-fitzgerald, line 81: ????

liam-fitzgerald (changes requested):
Looks good other than typo and URL injection

pkg/npm/http-api/src/Urbit.ts
eamsden: woops, accidental vim

eamsden:
Closing in favor of some documentation that points out:
- no need to start the event source explicitly, it will be done by the first poke or subscribe
- an empty string for the URL results in channel URLs being host-relative.
- you can simply set ship from window.ship

So this is sufficient initialization code:
```javascript
const airlock = new Urbit('', '');
airlock.ship = window.ship;
```

PR #5207, opened 2021-09-07
dist: fix update ordering in kiln, wire to UI

liam-fitzgerald:
When we receive the %mere gift from clay, the kernel has not yet been
reloaded. This means any attempts to bump desks will fail, as they will
be bumped against the old kernel. Rectifies this by continuing the %base
desk update flow in +on-load, instead of +take-merge-main. Also adds a
wef=(unit weft) to the state in order to discriminate whether or not the
kernel has just been reloaded

Also allows bumps to be triggered from the UI

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -5,7 +5,8 @@
 =,  format
 =*  dude  dude:gall
 |%
-+$  state    state-5
++$  state    state-6
++$  state-6  [%6 pith-6]
 +$  state-5  [%5 pith-5]
 +$  state-4  [%4 pith-4]
 +$  state-3  [%3 pith-3]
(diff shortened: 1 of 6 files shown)


pkg/arvo/lib/hood/kiln.hoon
```diff
@@ -598,9 +625,19 @@
         (emil sync-ud:pass (diff:give diff) ~)
       ?:  (lth num.new-weft num.old-weft)
         ~>  %slog.0^leaf/"kiln: future version {<new-weft>}, enqueueing"
+        ::  retry upgrade if not blocked anymore
+        =/  base=arak  (~(got by ark) %base)
         =.  next.rak  (snoc next.rak [(dec aeon.rail.rak) new-weft])
+        =.  ark  (~(put by ark) loc rak)
         =/  =diff  [%block loc rak new-weft blockers=(sy %base ~)]
-        (emil sync-ud:pass (diff:give diff) ~)
+        =.  vats  (emil sync-ud:pass (diff:give diff) ~)
```
belisarius222, line 633: I'm not sure we still want to report a block if we're going to update the kernel.  Maybe we should move this line down to the first branch of the `?.` so it only notifies if we can't upgrade the kernel yet.

belisarius222: A counterargument is that the kernel update might happen in a later event, and until then, this desk is blocked, in which case we should notify about that intermediate state.

belisarius222 (approved):
LGTM.  Thank you, this is a significant improvement.

PR #5209, opened 2021-09-08
grid/treaty: link to apps

liam-fitzgerald:
- Can link directly to apps with the url `web+urbitgraph://~host/desk/rest/of/link`
- If the rest of link is empty, then url links directly to app for install
- Special cases existing landscape links
- Can now request treaties outside of alliance via subscription

Coming in a seperate PR:
- Integrate into landscape calling scheme
- UI designs for invalid links
- UI designs for settings page
- (maybe) send links to apps as a graph-store content

cc: @urcades


arthyn (approved):
Happened to pop in here this morning haha, LGTM

PR #5210, opened 2021-09-08
interface: rework group joining interface

liam-fitzgerald:
Does what it says on the tin. Seeking UI/design feedback cc: @urcades also, this doesn't include a story around direct message/group chat invites/joins but perhaps they should go elsewhere
Screen Shot 2021-09-08 at 2 14 26 pm
Screen Shot 2021-09-08 at 2 14 40 pm


zalberico:
I don't think anything is going out for groups 1 at this point.

PR #5211, opened 2021-09-09
settings-store: move to garden, namespace

liam-fitzgerald:
- Namespaces all setting buckets by desk
- Moves settings-store to garden
- Integrates settings-store with grid UI


pkg/garden/lib/mip.hoon
```diff
@@ -0,0 +1,55 @@
+|%
```
Fang-, line 1: ur welcome (^:

Fang- (approved):
Looks good as far as I can tell!

PR #5214, opened 2021-09-11
btc-wallet: Add home button

finned-palmer:
Fixes: https://github.com/urbit/bitcoin-wallet/issues/43

Adds a home button (moves the header up to where it is in the new designs, too). Adds an "X" icon to close out of settings.

cc: @timlucmiptev @jalehman


finned-palmer:
@ridoy I'm not sure to be honest. Whenever tlon cuts a new release this will probably get included. I'm not sure when that will be, however.

PR #5216, opened 2021-09-12
kiln, gall: Fix removing agents from desk

liam-fitzgerald:
A two-parter, designed to cleanup the flow of removing an agent from a desk

notes:
- |commiting straight into an installed desk still breaks.
- L583 returned `mo-cor` instead of `cor`. My understanding is that this will throw away all the previously executed iterations of the `roll`, but continue the iteration regardless. I believe this constitutes a bug and so changed it to return `cor`

kiln: kill agents before merge

If a remote commit is downloaded that simultaneously removes an agent from
desk.bill but also removes the associated source files, then the commit
will fail as gall will not have received the card to kill the agent yet.
Instead, we read our foreign copy of the bill in +take-download, and
kill any necessary agents there, preventing a reload of the deleted
agent from occurring.

gall: ignore reload for dead agents

Fixes a bug where even if an agent was dead, if its source was modified,
then gall would still attempt a scry for its cage.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -580,7 +580,10 @@
     =/  yok=(unit yoke)  (~(get by yokes.state) dap)
     ?~  yok
       ~>  %slog.[0 leaf+"gall: no agent to reload: {<dap>}"]
-      mo-core
+      cor
+    ?:  ?=(%| -.agent)
+      ~>  %slog.[0 leaf+"gall: dead agent reload: {<dap>}"]
+      cor
     =/  bek=beak  [our q.beak.u.yok p.sign-arvo]
     =/  rag  (mo-scry-agent-cage dap q.bek p.sign-arvo)
     ?:  ?=(%| -.rag)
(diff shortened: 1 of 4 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -580,7 +580,10 @@
     =/  yok=(unit yoke)  (~(get by yokes.state) dap)
     ?~  yok
       ~>  %slog.[0 leaf+"gall: no agent to reload: {<dap>}"]
-      mo-core
+      cor
```
belisarius222, line 583: oof, good catch

belisarius222 (approved):
LGTM

belisarius222:
This problem where committing directly into a kiln-installed desk breaks things is one of a whole class of problems that come from kiln trying to manage desks and agents over which it doesn't have complete control.  This means there are moves that bypass kiln and change the state of the system out from under it (such as committing to a desk, or turning an agent on or off) where kiln has no way of intercepting that move.

I'm not sure what's the best way to deal with this class of problems.  One could imagine moving more logic into the kernel itself, which could eliminate the problem of multiple loci of control.  Another alternative would be to have Kiln listen more aggressively for attempts to change Clay or Gall state, by subscribing to notifications from Clay and Gall, and either adapt to the externally sourced change, or kill the event with a %pyre to prevent the change from taking hold.

Maybe there are other options too.  I'm open to ideas.

PR #5217, opened 2021-09-13
btc-wallet, bitcoin: update for distribution

liam-fitzgerald:
Adjusts routes, build, pokes to allow the SPA to be served from a non-root desk.


belisarius222 (approved):
LGTM

PR #5218, opened 2021-09-13
vere: use memset to wipe hashmaps

eamsden:
the hashcons used for cueing input events grows large on initial boot due to the need to store the entire pill. This exposed the slowness of iterating over all buckets in the hashcons to reset fill counts to 0 and wipe (~ 5*10^6 buckets on a run which loads solid.pill).

By factoring out the `fill` field of the `pail` types into a `fills` array field parallel with the `buckets` array in the `dict` types, we can use `memset` to 0 all of the pail fill fields.

NB it is semantically valid to simply memset the `buckets` array but the need to 0 all of the ref and val entries outweights any performance gain from using `memset` in this case, per my benchmarking.

### DIFF
--- a/pkg/urbit/include/ur/hashcons.h
+++ b/pkg/urbit/include/ur/hashcons.h
@@ -40,7 +40,6 @@ typedef uint32_t ur_mug;
 #define ur_pail_max            10
 
 typedef struct ur_pail32_s {
-  uint8_t  fill;
   ur_nref  refs[ur_pail_max];
   uint32_t vals[ur_pail_max];
 } ur_pail32_t;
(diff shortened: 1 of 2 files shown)


joemfb (commented):
These changes look good.

I would like to try to estimate the performance impact on get/put. There are some limited de/serialization benchmarks you can run with `make bench`. They only test small inputs (an ames packet event), use measure with the wall clock, but they're better than nothing. For a big input, it'd be good to try de/serializing a pill, or reallocating the persistent state with `|meld`.

(While we're trending towards data-oriented design, it'd be interesting to see the impact of further deconstructing the keys/vals structs. But that's probably an experiment for another time.)

joemfb (approved):
I run some basic before/after benchmarking (`make bench` 3x each time), and the difference in timing was around the normal variation. (Note, we should find a format or the benchmark output that makes it easy to compare and average runs.)

I also tried run `time build/urbit-work meld path/to/pier` a couple times before and after (ignoring the first run due to it's extra deduplication work), and this change made it close to a second faster.

So I'm comfortable with these changes as is. A couple more things worth trying (now or later) are to combine the allocations of the `buckets` and `fills` array, and to complete the data-oriented design by removes the `ur_pail*_t` structs and simply adding arrays of `keys`/`vals` to the relevant `ur_dict*_t` structs.

Also, please target this PR against `release/next-vere`.

PR #5219, opened 2021-09-14
dist: webterm: standalone package

Fang-:
Splits ~~herm and~~ the webterm interface out into their own directories for distribution separate from landscape/groups. Package.json, webpack configs, and base page code adapted from landscape.

- [ ] Webterm glob pending. What's the latest on generating globs for upload to bootstrap?

Also unsure what the right pattern is for web interface sources. I just put it under pkg/interface, because we obviously can't have it in pkg/webterm itself.

(Notably: this is still old webterm. Expect a nasty full-manual merge here later on.)


Fang-:
Totally forgot that @belisarius222 and I had previously talked about `/app/herm` and decided it should stay in base. So `pkg/webterm` now contains just the config files...

belisarius222 (approved):
LGTM

Fang-:
Oops almost forgot about the glob. Should be ready for review now @liam-fitzgerald.

liam-fitzgerald (approved):

PR #5220, opened 2021-09-14
dist: docket: improve upload page

Fang-:
Most notably, we further restrict the desks eligible for glob upload
to just the ones whose glob is said to be hosted on the local ship.
I'm assuming this is correct and sane, because I'm having trouble imagining the flow/use of uploading a glob in the "uploaded glob for desk that says it gets glob from a remote location" cases.

On the upload page, we replace the desk input field with a dropdown
from which eligible desks can be selected.
Additionally we now display some instructions, instead of leaving
developers entirely on their own.

Looks like so:

image

Or like so if there's no local-glob desks:

image

### DIFF
--- a/pkg/garden/app/docket.hoon
+++ b/pkg/garden/app/docket.hoon
@@ -337,31 +337,66 @@
     %-  as-octt:mimes:html
     %-  en-xml:html
     ^-  manx
+    ::  desks: with local globs, eligible for upload
+    ::
+    =/  desks=(list desk)
+      %+  murn  ~(tap by charges)
+      |=  [d=desk [docket *]]
+      ^-  (unit desk)
+      =-  ?:(- `d ~)
+      ?&  ?=(%glob -.href)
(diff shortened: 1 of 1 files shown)


belisarius222:
> we further restrict the desks eligible for glob upload to just the ones whose glob is said to be hosted on the local ship.

What do we gain from restricting this?  I can imagine wanting to tell a user in dire straits to manually load in a new glob, but being unable to do so because of this restriction.  If the globulator is a security vulnerability, then that's still a problem with this restriction, since the ship intending to host the app could get pwned and push out bad apps even if other ships are restricted.

Fang-:
> I can imagine wanting to tell a user in dire straits to manually load in a new glob

Oh, yeah that's totally fair game actually. I was only coming at this from a "distributor uploading canonical globs" case, but I guess users may want to fiddle with this themselves in certain cases. I'll loosen this again.

ryjm (commented):
LGTM

PR #5221, opened 2021-09-14
vere: windows fixes, refactoring for new terminal

joemfb:
This PR fixes a crash on windows (under `-t`) by applying libuv/libuv#3302 (allowing null writes to `uv_pipe_t`). Additionally, it suppresses (most) null writes in `term.c`, and includes a little refactoring.

/cc @locpyl-tidnyd

### DIFF
--- a/pkg/urbit/compat/mingw/uv.patch
+++ b/pkg/urbit/compat/mingw/uv.patch
@@ -15,6 +15,14 @@ index 0f2bb869b..f81245ec6 100644
      if (err == ERROR_ACCESS_DENIED) {
        /*
         * SetNamedPipeHandleState can fail if the handle doesn't have either
+@@ -1054,7 +1054,6 @@ static DWORD WINAPI uv_pipe_writefile_thread_proc(void* parameter) {
+   assert(req != NULL);
+   assert(req->type == UV_WRITE);
+   assert(handle->type == UV_NAMED_PIPE);
+-  assert(req->write_buffer.base);
+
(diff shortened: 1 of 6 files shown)


pkg/urbit/compat/mingw/uv.patch
```diff
@@ -15,6 +15,15 @@ index 0f2bb869b..f81245ec6 100644
      if (err == ERROR_ACCESS_DENIED) {
        /*
         * SetNamedPipeHandleState can fail if the handle doesn't have either
+@@ -1054,7 +1054,7 @@ static DWORD WINAPI uv_pipe_writefile_thread_proc(void* parameter) {
+   assert(req != NULL);
+   assert(req->type == UV_WRITE);
+   assert(handle->type == UV_NAMED_PIPE);
+-  assert(req->write_buffer.base);
++  assert(req->write_buffer.base || req->write_buffer.len == 0);
```
Fang-, line 23: The linked PR simply removes the assert entirely, instead of making it pass under more circumstances.

I understand the patch here resolves our issue as-is, and assume this patch will be removed once we can use the new libuv version, but don't we want to match that new version as closely as possible until then?

Fang- (approved):
Question about the libuv patch herein, but other than that these are all good changes!

pkg/urbit/compat/mingw/uv.patch
joemfb: good point

PR #5224, opened 2021-09-15
naive: tx parser performance improvements

joemfb:
This PR updates the naive l2 tx parser to use a bitwise cursor in place of reallocating right shifts. Parsing a batch of 1k txs improved from 3 seconds to .5 seconds, while 5k went from 77 seconds to 8 (all numbers approximate, and from my machine).

Additionally, this PR replaces separate `+div`/`+mod` calls with `+dvr`, rewrites `+ud-to-ascii`, and includes a bunch of cleanup and new tests (from @philipcmonk and @drbeefsupreme) that were already in the base branch.


joemfb:
It looks like CI has been broken on this branch for some time. I can look into that tomorrow if nobody knows why ...

pkg/arvo/lib/naive.hoon
```diff
@@ -402,10 +382,12 @@
 ::
 ++  ud-to-ascii
   |=  n=@ud
-  ^-  @t
-  ?~  n
-    *@t
-  (cat 3 $(n (div n 10)) (add '0' (mod n 10)))
+  ?~  n  '0'
```
joemfb, line 385: The previous implementation didn't handle `0`. If all argument are known to be non-zero, this line can be deleted.

pkg/arvo/tests/lib/naive.hoon
```diff
@@ -3692,13 +3693,19 @@
         [[c b %marbud-key-0] +(c)]
       [[c b %marbud-key-1] +(c)]
   ::
+  =/  txs
+    (cad:naive 3 (gen-tx 0 tx-1 %marbud-key-0) (gen-tx 1 tx-2 %marbud-key-1) ~)
+  =/  bat=octs  (cad:naive 3 (reap (add 1 (div batch-size 2)) txs))
```
drbeefsupreme, line 3698: This gives the wrong nonce after the first two transactions. I suppose it doesn't matter if all we're testing is parsing speed since parsing happens before nonce is checked, but it would also be good to know that there are no issues when a batch consists of a large number of successful transactions too.

philipcmonk: Yeah.  The issue right now is that processing transactions takes much longer than parsing, because verifying signatures requires two (unjetted) keccak calls (one before because we sign a hash and one after because we hash to turn the recovered point into an ethereum address), so the test takes too long to run at the batch sizes we need to test (5k is probably sufficient).

Now that I realize keccak is used at runtime, it's a priority to jet it, so we'll get that done soon.  When that happens, I expect we'll be able to do your full version with 5k txs and it'll take around 15 seconds, which is worth doing.  The rest of the test suite will be a lot faster too, because it spends virtually all its time in keccak.

Until then, we should make sure this test (1) passes, (2) is ready to be used in its full form to make sure it processes all txs (like you had it), and (3) is ready to be used to test only parsing speed.  For (3), generating valid txs is time-consuming and unnecessary, and in fact we should probably just call parse-roll directly.  Probably just having two separate tests is good.

joemfb: I split the test into two and tweaked the output.

joemfb:
@philipcmonk spotted a bug where i was using an absolute step as if it were relative. This basically recapitulated the previous pattern of allocating many, large intermediate atoms from the batch. With that resolved, the memory usage of the parser falls of a cliff (and is now linear in the size of the batch, as opposed to quadratic or worse). Parsing 100k transactions takes `~s3` on my machine (per a stopwatch), smaller numbers are too fast to time in that way.

joemfb:
batch size, overall memory used, cells allocated:

before (with the cursor bug):

```
500:  MB/6.057.984,     418.013 cells
1k:   MB/23.476.496,    835.013 cells
2.5k: MB/143.981.984, 2.086.013 cells
5k:   MB/572.324.496, 4.171.013 cells
```

after:

```
10K:  MB/6.814.592,   8.431.022 cells
100K: MB/68.014.592, 84.301.022 cells
```

philipcmonk (approved):
This is great, thanks!

PR #5226, opened 2021-09-15
dist: install software from multi-desk pills on boot, track local desk changes

Fang-:
a29c341 updates the `+solid` generator to support including any number of secondary desks in addition to `%base`. This works fine for including additional filesystem data. If I try to make the equivalent change to `+brass`, the resulting pill fails to boot with `arvo: bad wisp`, even though I don't think I included any new kinds of tasks. Unsure why that is, or whether we care about this for brass pills in the first place.

22ad251 updates kiln to install software from any desks that might already exist besides `%base` and `%kids` during `+on-init`. This is why the above change puts the base desk _last_. Additionally, after installing a desk we immediately initiate another install flow, this time sourcing from our sponsor, so that we may get updates for the desk. However...

d7afe32 is required to support that behavior as it's currently implemented. If we continue to do installs from local desks "asychronously" by passing request cards to clay, then the two requests for a desk (local and sponsor installs) will clobber each other, and only the second request will return a response, presumably because they all go over the same wire.* So instead, we make local installs "synchronous" by getting the requisite** data through scries.

\* I just went and assumed this is intended behavior. ...right?

** The `sign` that the relevant gates take only gets used in `+take-sync`'s case, so perhaps we can get away with just giving a `*rant:clay` instead? I'm guessing the sign is merely treated as an indicator that clay now has the data we care about.

~~Asterisks aside, I'm not entirely sure this is the correct approach in the first place, so I'd like some feedback here. My general intuition is that doing these install steps sequentially (local, then enabling remote) is the right (and perhaps only) way to sanely & safely do this, and doing special logic for "completion of installs triggered by `+on-init`" would add more complexity to these flows.~~


belisarius222 (approved):
LGTM, except that docket needs to be listening for %commit rather than %merge.  Once that's fixed, let's merge this.

PR #5228, opened 2021-09-16
interface: app permalink embeds

arthyn:
Adds rich embeds for app links in Landscape for easy sharing

Screen Shot 2021-09-16 at 4 25 17 PM


pkg/interface/src/logic/lib/permalinks.ts
```diff
@@ -80,6 +81,14 @@ export function referenceToPermalink({ reference }: ReferenceContent): Permalink
       link,
       ...reference.graph
     };
+  } else if ('app' in reference) {
+    const { ship, desk } = reference.app;
+    return {
+      type: 'app',
+      link: `web+urbitgraph://app/${ship}/${desk}`,
```
liam-fitzgerald, line 88: New link is actually 
```typescript
`web+urbitgraph://${ship}/${desk}`
```

liam-fitzgerald (changes requested):
Nit about link structure, plus tests are borked

tacryt-socryp:
Ideally these links could have additional path items that would be passed on to the app in question if it were opened.

pkg/landscape/sur/post.hoon
```diff
@@ -29,6 +29,7 @@
 +$  reference
   $%  [%graph group=resource =uid]
       [%group group=resource]
+      [%app =ship =desk path=@ta]
```
tacryt-socryp, line 32: Makes more sense as `[%app =ship =desk =path]` than throwing a `@ta` in there, no? @liam-fitzgerald @arthyn

liam-fitzgerald: Yeah absolutely

arthyn: oh cool, yeah that makes sense, didn't know we had a type specifically for that

PR #5229, opened 2021-09-17
serf: fix incorrect conditional (integer precision) in event tracing

joemfb:
I had broken event labels in `-j` trace output in a previous refactoring.


eamsden (approved):
it works

Verified that this works while tracing Ames large-noun sending.

PR #5230, opened 2021-09-17
u3: checks memory on all road transitions (fall), fixes memory leaks

joemfb:
This PR updates u3's memory-debugging gc to trace on every inner road exit, nor merely at the outermost layer of virtualization. The hoon test suite is already run under gc in CI, but this change makes that much more effective (as every individual test will be checked for leaks, not merely changes to the persistent state). Additionally, this PR fixes existing memory leaks uncovered by this new, more rigorous testing discipline.

All fixes are from @frodwith and @philipcmonk, I'm merely integrating their work..


eamsden (approved):
@joemfb went over this per-commit this afternoon.

A few rather involved changes but overall looks good.

PR #5231, opened 2021-09-17
hood: respect ordering of agents in bill

liam-fitzgerald:
This feels inconsequential, but there are a surprising number of edge cases that require this,  for example, 
- scrying in `+on-init`
- Moving an agent from one desk to another where the moved agent is also poked by an agent on the freshly started desk. 

In general, it feels nice that we control the order of the agents starting


belisarius222 (approved):
LGTM.  I think this would be a bit cleaner to just leave everything as lists and +skim for the existing entries, but that's a nitpick.

It's good for Kiln to honor the initialization order specified in the bill.  Otherwise, app developers need to write their agents so that they're robust to changes in initialization order, which is more work.  Better to just let the dev specify and rely on the order.

PR #5232, opened 2021-09-18
khan C driver

mrdomino:
Tracks work on the C IO driver for the control plane. Do not depend on this branch for anything please; it will probably be rebased somewhat aggressively before it becomes an actual PR.


ryjm:
Will be interesting to square the haskell version with this, a cursory glance is definitely throwing me for a loop. Might need a walkthrough.

mrdomino:
superseded by #5296

PR #5235, opened 2021-09-20
azimuth: update for latest ecliptic address

Fang-:
The upgrade hasn't triggered yet, but opening this so we don't forget.

Notably, the previous address change hasn't even made it to master yet. (^:


ritfed-tabryg:
> Notably, the previous address change hasn't even made it to master yet. (^:
This bit me when trying to use `claz` to rescue my planet, but then I forgot to make a PR. Thanks for making sure it happens *this* upgrade

Fang-:
Contract upgrade has happened, merging now.

PR #5237, opened 2021-09-21
treaty: default to hardcoded ship instead of sponsor

liam-fitzgerald:
- Our handling of 'allies' (treaty's term for a provider that allows app discovery), was inconsistent and incomplete. We previously used the ship sponsor, however this presents usability issues if the user's sponsor is offline, or doesn't maintain a useful list of apps. Additionally, we previously prompted to view ~zod's desks in the UI, if we didn't have any other allies to look at, which was inconsistent, because it should prompt for the sponsor, as they're the only ship that's guaranteed to be allied.
- No longer defaults to the sponsor, instead using a hardcoded ship name.
- Adjusts frontend to respect this hardcoded ship name

cc: @urcades

### DIFF
--- a/pkg/arvo/mar/kiln/vats-diff.hoon
+++ b/pkg/arvo/mar/kiln/vats-diff.hoon
@@ -11,7 +11,7 @@
     ?-  -.diff
       %block  (block +.diff)
       ?(%merge-sunk %merge-fail)  (desk-arak-err +.diff)
-      ?(%reset %merge %suspend %revive)  (desk-arak +.diff)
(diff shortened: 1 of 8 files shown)


Fang-:
We support multiple allies, right? Can we not have _both_ the sponsor and the hardcoded one be the default?

arthyn (approved):

urcades:
> Imo networking costs would look pretty similar to "how many different chatrooms can you be in" or in other words, would already scale up to having hundreds, if not thousands.

nice, very great to hear

PR #5238, opened 2021-09-21
docket: hash-based glob-ames, handled within app

Fang-:
This works around the mark definition issue that doing the same in spider would face, and makes versioning of globs explicit in the ames case as well, letting the docket file trigger new glob fetching.

Still missing:
- [x] Auto-updating the .docket file for a desk when a new glob is uploaded from the upload page.
- [ ] Treaty `+on-load` logic. Do we care about this or do we just nuke the testnet again?

Opening the PR now for review expedience.


belisarius222 (commented):
Hm, intuitively, I would expect the hash to be there as a sanity check no matter whether it's downloaded over Ames or HTTP.  Would be it too much of a pain to require devs to put the hash in there if they're hosting globs over HTTP?

Fang-:
We could probably do that. `-make-glob` just creates a file whose name is the hash, so shouldn't be hard on devs at all.

pkg/garden/ted/glob.hoon
```diff
@@ -9,7 +9,7 @@
 |^
 ?-  -.loc
   %http  (fetch-http url.loc)
-  %ames  (fetch-ames ship.loc base)
+  %ames  !!  ::NOTE  done within docket itself
```
ryjm, line 12: should we add a reminder to move this logic back here when the desk routing in gall is done?

Fang-:
Alright I haven't tested the remote distribution path here very extensively yet, but at least locally this should behave a lot better now. ptal

pkg/arvo/lib/hood/kiln.hoon
```diff
@@ -485,9 +485,9 @@
       ~>  %slog.0^leaf/"kiln: already tracking {here:(abed lac)}, ignoring"
       vats
     =:  loc  lac
-        rak  [~ *rein]
+        rak  [`[| our lac *aeon ~] *rein]
```
Fang-, line 488: This should probably be reverted in favor of 8d91e8a ?

pkg/garden/app/docket.hoon
```diff
@@ -235,7 +271,35 @@
...
+        [%glob-ames ~]
+      ?-  -.sign
+        %kick      `state
+        %poke-ack  ~&([dap.bowl %unexpected-poke-ack] `state)
+      ::
+          %watch-ack
+        ?~  p.sign  `state
+        %-  %-  slog
+            leaf+"docket: failed to fetch glob over ames for {(trip desk)}"
+        ::TODO  maybe we want to retry, but with delay?
```
belisarius222, line 283: I think this should cancel the installation.  Docket should display a notification that the app failed to install, and stop trying to install it.  The user could ask around and try again or something, but the system shouldn't keep trying indefinitely if something is broken on the remote.

pkg/garden/ted/glob.hoon
belisarius222: Sure.  If we later fetch globs file by file, then doing that in a thread would probably be better.

belisarius222 (changes requested):
We should:
- pull the glob hash out of the `$%`, since we'll want that there no matter how we fetch it
- write a migration, so we don't have to breach the testnet

PR #5243, opened 2021-09-22
Keccak

frodwith:
vere, urcrypt, jets: implementation of keccak crypto jets

### DIFF
--- a/pkg/urbit/jets/e/keccak.c
+++ b/pkg/urbit/jets/e/keccak.c
@@ -1,0 +1,39 @@
+/* jets/e/keccak.c
+*/
+#include "all.h"
+#include <urcrypt.h>
+
+#define defw(bits) \
+  u3_atom \
+  _kecc_##bits(c3_w len_w, u3_atom a) \
+  { \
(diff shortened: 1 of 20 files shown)


pkg/urbit/jets/e/keccak.c
```diff
@@ -0,0 +1,39 @@
...
+
+#define defw(bits,byts) \
+  u3_atom \
+  _kecc_##bits(c3_w len_w, u3_atom a) \
+  { \
+    c3_y  out[byts]; \
+    c3_y* buf_y = u3r_bytes_alloc(0, len_w, a); \
+    if ( 0 != urcrypt_keccak_##bits(buf_y, len_w, out) ) { \
+      /* urcrypt_keccac_##bits always succeeds when called correctly */ \
+      return u3m_bail(c3__oops); \
```
joemfb, line 14: I slightly favor `c3_assert()`, as it's more obvious that it results in a fatal error, and the macro includes a file/lineno printf. But that's pretty academic, we'll never hit it in practice.

joemfb (approved):
LGTM

PR #5244, opened 2021-09-23
dist: state adapters for glob hash changes

Fang-:
Not the prettiest (they could be deduplicated, should maybe have made use of `$docket`'s version tag), but this should be sufficient for testnet deployment purposes.


Fang-:
Apparently we just breached the testnet instead. ez pz

PR #5245, opened 2021-09-23
landscape: faster grid redirects

arthyn:
This gets the actual redirect happening faster, but subscriptions are still slow. It feels faster though.


zalberico:
@arthyn is this something we want to clean up and merge or should it be closed?

arthyn:
Outdated

PR #5246, opened 2021-09-23
dojo: support running other-desk generators

Fang-:
Syntax is `+desk#gen`. The `#` is just a placeholder (as is tradition), and I'd be happy to take suggestions/votes on the real thing. I was thinking `+desk>gen` maybe?

`:app|gen` de-sugars to `:app +desk#gen`, where the desk is the desk the app is running on.

`:app|desk#gen` is currently broken, in the sense that the above de-sugaring takes precedence, overwriting the manually specified desk with the desk where the app is running.

The `+on-load` here just aborts any running user action to avoid having a monstrously large state adapter.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -274,7 +274,7 @@
   ++  parse-server
     %+  stag  0
     ;~  plug
-      ;~(pose ;~(sfix sym hax) (easy q.dir))
+      ;~(pose ;~(sfix sym zap) (easy q.dir))
       (most fas sym)
     ==
   ::
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -12,7 +12,7 @@
 =>  |%                                                  ::  external structures
     +$  id  @tasession                                  ::  session id
     +$  house                                           ::  all state
-      $:  %6
+      $:  %7
           egg=@u                                        ::  command count
           hoc=(map id session)                          ::  conversations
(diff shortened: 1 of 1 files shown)


ohAitch:
What does +term/gen currently do?

ohAitch:
(For example `+/=desk==name` would be more compositional, but kind of a mouthful)

Fang-:
> What does +term/gen currently do?

It runs the generator in `/gen/term/gen.hoon`.

ohAitch:
Actually hm, `+desk%foo` as a shortening of `/=desk%/foo`(again wrt `/===gen` as %) would be not completely terrible

ryjm:
`+desk!gen` ala `Dark!Harry`

desks are just urbit fanfics

joemfb:
Is there a need for this separate from changing the working directory? Ie, would the following be sufficient?

```
=dir /=desk=
+gen
```

belisarius222:
maybe we could do like %= syntax?

+gen(dir /=desk) arg

—
~rovnys-ricfer
https://urbit.org

Fang-:
> Is there a need for this separate from changing the working directory?

@joemfb This is precisely so you don't _have_ to continuously change the working directory. Certainly needing to do so for the `:app|gen` case is dumb.

belisarius222:
I'll confirm that needing to change dir to run generators for agents on other desks is annoying and error-prone.

—
~rovnys-ricfer
https://urbit.org

Fang-:
CI errors should be fixed by pills on latest dist. @belisarius222 gave his approval out of band.

PR #5248, opened 2021-09-24
urcrypt: argon2 only optimize for x86_64 on x86_64

botter-nidnul:
fixes urcrypt's automake/autoconf files so that libargon2 will compile on architectures without SSE instructions, but x86_64 will still use the optimizations


joemfb (approved):
LGTM, thanks!

PR #5249, opened 2021-09-24
vere: fixes crash in comet booting

joemfb:
This fixes a bug in #4938 -- `who_c` was set in all cases except comet boot, in which case it was null -- by simply retrieving the ship name from the `$feed`. Also fixes the `c3_assert()` printf and bumps the vere version.

 (Note to self: the dataflow for all of pre-boot has become very convoluted, and these modules should be refactored.)


philipcmonk (approved):
Seems fine

PR #5250, opened 2021-09-24
vere: revert next-gen terminal

joemfb:
This PR reverts #4463, for the following reasons:

- the implementation includes a couple minor bugs: `|mass` output is broken, and previous terminal state is not always properly restored
- the implementation includes a mandatory binary upgrade (via arvo/vere version negotiation), and this will intersect painfully with software distribution's new arvo/userspace version negotiation (at the very least, it could involve failed upgrade attempts, and more than a little pausing of actually-unaffected agents)
- at this time, it's high risk and low reward (since no *released* agents make use of the new functionality)

I've restored the terminal branch at `m/next-gen-term`. This terminal is important and exciting work, and I'm committed to integrating and releasing it as soon as practical.

(This PR includes a revert of the original merge commit, and cherry-picks of the build fixes and cleanup that was added at the end of the original PR.)


philipcmonk (approved):
Agree that we should delay this until later.

Issue #1468, opened 2016-12-04
Failing first compile leaves app in a half-initialized state


ponnys-podfer:
Create an app which apparently is compiles fine but that gall will reject. Here's my minimal bad `app/mood2.hoon`:

    !:
    |_  bowl
    ++  poke-noun
      |=  arg/*
      ^-  {(list) _+>.$}
      ~&  [%argument arg]
      [~ +>.$]
    --

Try to start it:

    + /~picdyl-mipdec-ponnys-podfer/mood/8/app/mood2/hoon
    activated app mood/mood2
    %mood2: bogus core
    >> [%mo-not-running %mood2 %peer]
    > |start %mood2
    >=
    : /~picdyl-mipdec-ponnys-podfer/mood/9/app/mood2/hoon
    [already running %mood/%mood2]
    > |start %mood2
    >=
    >> [%mo-not-running %mood2 %poke]
    > :mood2 5

Gall becomes confused about whether the app is running or not; it will claim it's not on each poke, while claim it's already running when attempting to `|start` it.

(Also, this bogus core message should be more diagnostic about what exactly is wrong about the core.)

zalberico:
Looks like the changes in #5251 got cherry picked and merged, which from what I can tell resolves this.

PR #5251, opened 2021-09-25
clay: fix failure to load apps

philipcmonk:
When you loaded an app with an error, then fixed the error, it would
create the main gall %mult subscription at a time in the past.  Then,
clay would never fill the subscription since it couldn't get the old %a
entries for the apps.

This fixes the issue in two ways:  first, don't subscribe in the past.
Second, if clay can't get the old versions, just fire the subscription
anyway.

@belisarius222 please merge or cherry-pick this into dist upon approval, so everything gets tested there.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -3404,6 +3404,19 @@
           [new-cach.rov fod.dom]
         (read-unknown mool.rov(case [%ud u.aeon.rov]) new-cach.rov)
       =.  new-cach.rov  n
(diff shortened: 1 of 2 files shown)


belisarius222 (approved):
LGTM.  I don't recall why the date is in the wire in the first place, but I looked through git blame, and that's been the case since before ford fusion, so considering that this fixes the issue at hand, I'll assume that the date was relevant in an older context that was quite different from the current one.

Fang-:
This seems like it might finally fix #1468 and/or #1481.

belisarius222:
Merged into the dist branch.

PR #5253, opened 2021-09-25
kiln, docket: fix local installation

Fang-:
I noticed docket wasn't getting notified when I ran `|install our %something`. Turns out kiln just wasn't sending the `%commit` diffs that docket relied on in the local case.

Patched kiln to send a `%commit` during local install. Not entirely sure if this is the semantically correct change, please lmk.

Once kiln was sending the notifications, docket was crashing when trying to add the new docket to state, so patched that as well.


ryjm:
https://github.com/urbit/landscape/issues/1200#issuecomment-926357699

Possibly related? I also noticed this but thought it was by design.
`:docket|install` didn't work for local installs either though (see linked
issue).

belisarius222 (approved):
LGTM.  I'm not completely sure if docket should also be doing other things to react to a newly installed desk in kiln than it does now, but if so, this change should still get us closer to the desired behavior.

PR #5255, opened 2021-09-26
garden spider: support globs over ames

ryjm:
Moves mark validation for spider out of clay (for now). Spider now performs that validation itself, since it knows what desk the thread originates from. This keeps threads scoped to the app desk where they were defined. 

This is a workaround - gall should take care of this in the future.


pkg/arvo/app/spider.hoon
```diff
@@ -167,15 +177,17 @@
...
     =.  any  (old-to-3 any)
-    ?>  ?=(%3 -.any)
+    =.  any  (old-to-4 any)
+    ?>  ?=(%4 -.any)
     ::
     =.  tid.state  tid.any
     =/  yarns=(list yarn)
       %+  welp  running.any
       ~(tap in ~(key by starting.any))
+
```
Fang-, line 190:
```suggestion
```

philipcmonk (commented):
This looks correct to me, pending sigpams cleanup

pkg/arvo/app/spider.hoon
```diff
@@ -206,16 +218,28 @@
     ::
     ++  old-to-3
       |=  old=clean-slate-any
-      ^-  clean-slate
-      ?>  ?=(?(%2 %3) -.old)
-      ?:  ?=(%3 -.old)
+      ^-  clean-slate-any
+      ?>  ?=(?(%2 %3 %4) -.old)
+      ?:  ?=(?(%3 %4) -.old)
         old
```
Fang-, line 224: Might be nicer for this to produce `clean-slate-3` as per the arm name, but it might not actually matter.

pkg/arvo/app/spider.hoon
```diff
@@ -511,11 +534,11 @@
...
   ^-  (quip card ^state)
   =-  (fall - `state)
-  %+  bind  
+  %+  bind
     (~(get by serving.state) tid)
-  |=  [eyre-id=@ta output=mark =desk]
+  |=  [eyre-id=(unit @ta) output=mark =desk]
   :_  state(serving (~(del by serving.state) tid))
-  %+  give-simple-payload:app:server  eyre-id
+  %+  give-simple-payload:app:server  (need eyre-id)
```
Fang-, line 541: What happens if this crashes? Same q for the identical change below.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1590,7 +1590,7 @@
     =.  fod.dom  ford-cache.args
     =.  ..park  (emil (print q.old-yaki data))
     ::
-    =?  ..park  &(updated !dist-upgraded.ruf)  migrate-dist
+    ::  =?  ..park  &(updated !dist-upgraded.ruf)  migrate-dist
```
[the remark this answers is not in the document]
Fang-: Please revert!!

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1408,6 +1408,12 @@
         ?.  ?=(%raw-fact -.unto)
           [unto ap-core]
         =/  =case:clay  da+now
+        ~&  >  take+beak.yoke
+        ~&  >  take+mark.unto
+        ~&  >  take+[agent-name agent-routes agent-duct]
```
Fang-, line 1413: Probably take these out.

pkg/base-dev/lib/strand.hoon
```diff
@@ -128,6 +128,18 @@
           [%done value=a]
       ==
     ::
+    ++  validate-mark
+      |=  [in=* =mark =bowl]
+      ^-  cage
+      ~&  validate+[mark q.byk.bowl]
```
Fang-, line 134: Probably take this out.

pkg/garden/app/docket.hoon
```diff
@@ -275,53 +275,17 @@
...
         %kick   `state
       ::
           ?(%poke-ack %watch-ack)
         ?~  p.sign  `state
+        =/  err
         =/  act=tape  ?:(?=(%poke-ack -.sign) "start" "listen")
-        =.  charges  (new-chad:cha hung+'glob-failed')
+        =.  charges  %-  new-chad:cha
+          ?:  ?=(i.t.t.wire %http)  hung+'glob-failed'
+            hung+'failed to fetch glob via ames'
```
Fang-, line 288:
```suggestion
        =.  charges
          %-  new-chad:cha
          ?:  ?=(i.t.t.wire %http)
            hung+'failed to fetch glob via http'
          hung+'failed to fetch glob via ames'
```

pkg/garden/app/docket.hoon
```diff
@@ -275,53 +275,17 @@
...
-    ::
-        [%glob @ %http @ ~]
+        ?([%glob @ %http @ ~] [%glob @ %ames @ ~])
       ?-  -.sign
         %kick   `state
       ::
           ?(%poke-ack %watch-ack)
         ?~  p.sign  `state
+        =/  err
         =/  act=tape  ?:(?=(%poke-ack -.sign) "start" "listen")
```
Fang-, line 285: Syntax error? `=/  err` should either be completed or removed.

pkg/garden/ted/glob.hoon
```diff
@@ -25,4 +25,14 @@
     =-  ?>(=(hash (sham -)) -)
```
Fang-, line 25: We might want to remove this hash check assertion while we're at it, `/app/docket` already handles this, and much more gracefully.

Fang- (changes requested):
Some nits, a question, and some change requests.

pkg/arvo/app/spider.hoon
ryjm: i did it like this because old-to-2 had a similar thing. and it looks like it can produce a clean-slate-4 too... i'm not sure i like the pattern but probably best to keep it consistent.

pkg/garden/ted/glob.hoon
ryjm: do we still need the `|.` there?

Fang-: Yes, the `;;` might still crash.

pkg/arvo/app/spider.hoon
```diff
@@ -18,18 +18,27 @@
...
+      clean-slate-3
       clean-slate
   ==
 ::
 +$  clean-slate
+  $:  %4
+      starting=(map yarn [=trying =vase])
+      running=(list yarn)
+      tid=(map tid yarn)
+      serving=(map tid [(unit @ta) =mark =desk])
```
liam-fitzgerald, line 38: IMO, given that we're serious about multi-desk support for spider, `serving` should probably be just `(map tid [eyre-id=@ta =mark])` and we should store the desk of the thread alongside the thread itself. Defaulting to `%noun` as a mark feels kinda wrong tbh.

liam-fitzgerald (commented):
LGTM, but interesting on the `serving` note

ryjm:
i don't think this code is going to live in spider for very long, so
putting it in serving means we don't need to update state when we remove
it. this really should  be fixed in gall.

liam-fitzgerald (approved):

belisarius222:
> i don't think this code is going to live in spider for very long
> […](#)

It could be there for quite a while.  Moving thread execution to kernelspace is nontrivial and is unlikely to land before a few months from now.  We should consider this hack to be basically permanent for now.

Fang-:
Note that this was merged into `dist`, but we might still want to address @belisarius222's comment. Whether that's worth addressing _prior_ to release...

PR #5259, opened 2021-09-28
dist: version necessary marks

liam-fitzgerald:
- Bump %graph-update to v3
- Version glob, treaty, docket, kiln-vats-diff marks


pkg/landscape/app/graph-store.hoon
```diff
@@ -111,7 +112,7 @@
   ?>  (team:title our.bowl src.bowl)
   =^  cards  state
     ?+  mark           (on-poke:def mark vase)
-      %graph-update-2  (graph-update !<(update:store vase))
+      %graph-update-3  (graph-update !<(update:store vase))
```
Fang-, line 115: Are these not backwards compatible in some way? What changed?

belisarius222: references to other apps were added

Fang-: We can and should continue accepting `%graph-update-2` then, right?

Fang-:
Also what about `/mar/treaties`?

PR #5261, opened 2021-09-28
Jb/dynamic hint template

DaneBettis:
vere: adds %bout function, reports nanoseconds spent on the given hoon expression

This PR adds the %bout hoon function, which vere reads as a hint to run the enclosed hoon code capturing nanoseconds since the epoc (whichever epoc each host OS uses) before and after the run, which then prints the time delta at the end of the run, in nanoseconds.

This is my first attempt at a PR to urbit, I have attempted to follow the guidelines, and welcome all feedback.


ohAitch:
Nice, definitely much better than `:time`. As long as it's referentially transparent ;)

PR #5269, opened 2021-09-30
graph-pull-hook: restore network connectivity

liam-fitzgerald:
- The pull hook renegotiation poke may have crashed during the software distribution OTA. This PR fixes this by rerunning the poke.
- Additionally adds a flag to the kick poke to specify the aggressiveness of renegotiations.


Fang- (approved):
Tested on ~palfun-foslup, it works!

PR #5283, opened 2021-10-01
jael: add %ruin task for pretend-breaching others

Fang-:
The initial use case here is clearing state relating to long-gone
comets, but this could prove to be a useful general-purpose rescue tool.

@philipcmonk you used this successfully during rescue, right? Should we ship it as-is, or provide some integration/generator/whatever?


philipcmonk (approved):
Yes, this worked.  I think relying on |pass is fine for a debugging command like this.

PR #5284, opened 2021-10-01
hood: report %kids desk hash in +report-vats

eamsden:
`+vats` / `+trouble` are generators that simply print the output of the `+report-vats` arm of sur/hood/hoon.

@philipcmonk asked for the kids hash to be printed as well.

### DIFF
--- a/pkg/base-dev/sur/hood.hoon
+++ b/pkg/base-dev/sur/hood.hoon
@@ -65,8 +65,7 @@
   =+  .^  raz=(list vat)
           %gx  /(scot %p our)/hood/(scot %da now)/kiln/vats/noun
       ==
-  %+  weld
-    [(report-kids our now) ~]
+  :-  (report-kids our now)
   (turn raz |=(v=vat (report-vat our now v)))
 ::  +report-vat: report on a single desk installation
 ::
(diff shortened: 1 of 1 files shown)


pkg/base-dev/sur/hood.hoon
```diff
@@ -65,6 +65,8 @@
   =+  .^  raz=(list vat)
           %gx  /(scot %p our)/hood/(scot %da now)/kiln/vats/noun
       ==
+  %+  weld
```
eamsden, line 68: originally I had this the other way so it needed to be a weld but now it can just be a `:-`

pkg/base-dev/sur/hood.hoon
```diff
@@ -103,6 +104,18 @@
...
   ==
+::  +report-kids: non-vat cz hash report for kids desk
+::
+++  report-kids
+  |=  [our=ship now=@da]
+  ^-  tank
+  =/  dek  %kids
+  =/  ego  (scot %p our)
+  =/  wen  (scot %da now)
+  ?.  (~(has in .^((set desk) %cd /[ego]/base/[wen])) dek)
```
Fang-, line 115: You could consider leaving out the desk here because it doesn't affect the scry.

```suggestion
  ?.  (~(has in .^((set desk) %cd /[ego]//[wen])) dek)
```

Fang- (approved):
As discussed out of band, `%kids` is not an installed desk, but it's an acceptable exception. Looks fine to me!

PR #5292, opened 2021-10-02
docket: correctly en-vase a docket definition

Fang-:
This was not so clearly correct after all. (^:


ryjm:
hah i was just about to submit a PR for this.swiped.

belisarius222 (approved):
LGTM

PR #5294, opened 2021-10-02
shoe: allow scrying into the underlying shoe app

Fang-:
For some reason we were never passing +on-peek calls on to the underlying agent.


philipcmonk (approved):
Missed this, LGTM

PR #5300, opened 2021-10-04
kiln: refresh dudes on local reinstall

belisarius222:
This should fix a bug where after uninstalling and reinstalling a desk locally, the agents would not be jolted back to life.


Fang- (approved):
Good change!

PR #5303, opened 2021-10-05
grid: auto-ally on leaping to a ship

liam-fitzgerald:
Automatically allies with a ship if it is leapt to.

### DIFF
--- a/pkg/garden/app/treaty.hoon
+++ b/pkg/garden/app/treaty.hoon
@@ -46,7 +46,7 @@
     =+  ;;([%add =desk] q.vase)
     =/  =docket:docket  ~(get-docket so:cc desk)
     =/  =treaty  (treaty-from-docket:cc desk docket)
-    =.  sovereign  (~(put by sovereign) desk treaty)
+    =.  sovereign  (~(del by sovereign) desk treaty)
     `this
   ==
   ::
@@ -78,7 +78,7 @@
       =/  =treaty  (treaty-from-docket:cc desk docket)
       =.  sovereign  (~(put by sovereign) desk treaty)
       :_  this
-      ~[publish warp give]:so
+      [publish warp give]:so
     ::
         %del
       =,  update
@@ -105,6 +105,9 @@
     ?>  =(our.bowl src.bowl)
     =.  direct  (~(put in direct) [ship desk])
     :_(this (drop ~(safe-watch tr:cc [ship desk])))
+    ::
+      [%treaties ~]
+    `this
     ::
       [%alliance ~]
     :_  this
@@ -195,6 +198,7 @@
   ::
   ++  take-treaty
     |=  =desk
+    =*  tr   ~(. tr:cc ship desk)
     ?+  -.sign  (on-agent:def wire sign)
     ::
     :: rewatch only if we aren't source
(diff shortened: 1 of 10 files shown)


pkg/garden/app/treaty.hoon
```diff
@@ -46,7 +46,7 @@
     =+  ;;([%add =desk] q.vase)
     =/  =docket:docket  ~(get-docket so:cc desk)
     =/  =treaty  (treaty-from-docket:cc desk docket)
-    =.  sovereign  (~(put by sovereign) desk treaty)
+    =.  sovereign  (~(del by sovereign) desk treaty)
```
Fang-, line 49: But is this not the `%add` case? This looks wrong.

pkg/garden/app/treaty.hoon
```diff
@@ -298,9 +302,17 @@
   ++  watching  (~(has by wex.bowl) [path dock])
   ++  safe-watch  `(unit card)`?:(|(watching =(our.bowl ship)) ~ `watch)
   ++  leave  (leave:pass dock)
+  ++  gone   
+    ^-  (list card)
+    :~  (fact:io (treaty-update:cg %del ship desk) /treaties ~)
+        (kick-only:io our.bowl path ~)
+    ==
```
Fang-, line 309: These arms feel like they might be safer to use if they do both the state modification and effects at once (ie `^-  (quip card treaty)` or w/e), but that might be out of scope for this pr.

pkg/garden/app/treaty.hoon
```diff
@@ -105,6 +105,9 @@
     ?>  =(our.bowl src.bowl)
     =.  direct  (~(put in direct) [ship desk])
     :_(this (drop ~(safe-watch tr:cc [ship desk])))
+    ::
+      [%treaties ~]
+    `this
```
Fang-, line 110: If we give new/removed treaty updates on this path, we should also give an initial result here, right?

pkg/garden/app/treaty.hoon
```diff
@@ -314,8 +326,11 @@
   ++  kick
     (kick:io path ~)
   ++  give
+    ^-  (list card)
     =/  t=treaty  (~(got by sovereign) desk)
-    (fact:io (treaty:cg t) /sovereign path ~)
+    :~  (fact:io (treaty-update:cg %add t) /treaties ~)
```
Fang-, line 331: Looking at `+on-watch`, the `/sovereign` path here seems like it was an erroneous leftover. We don't distinguish between "subscribing to treaties published by the host" and "all treaties known to the host" anymore, do we? Just making sure I understand this correctly.

Fang- (changes requested):
I think this isn't quite right, ptal at the comments below. Could be that I just need to update my understanding of how this is intended to be used.

ryjm:
@arthyn think you could take a quick pass over these frontend changes too?

pkg/grid/src/nav/search/Providers.tsx
```diff
@@ -87,11 +103,22 @@ export const Providers = ({ match }: ProvidersProps) => {
...
+          ]
+        : [];
+
       useLeapStore.setState({
-        matches: ([] as MatchItem[]).concat(appMatches, providerMatches)
+        matches: ([] as MatchItem[]).concat(appMatches, providerMatches, newProviderMatches)
       });
     }
-  }, [results]);
+  }, [results, patp]);
```
arthyn, line 121:
```suggestion
  }, [results, patp, isValidPatp]);
```

arthyn (approved):
Looks good front-end wise if we can just commit that suggestion

PR #5307, opened 2021-10-06
dojo: support running threads from other desks

Fang-:
Similar to the recently added syntax for generators, you can now `-desk!thread` to run the thread from the specified desk.


Fang-:
Can confirm that works fine. Merging!

PR #5310, opened 2021-10-07
acme: load custom cert from base desk, not home

Fang-:
Small patch for the ancient workaround.

Arguably this should be `q.byk.bowl` instead, but then if this ever goes onto a non-base desk, you risk pushing the cert out to all who installed from the desk.


joemfb (approved):
LGTM

Issue #44, opened 2014-08-29
[CLOSED] Initial checkin of emacs mode hoon-mode.el


philipcmonk:
<a href="https://github.com/abliss"><img src="https://avatars.githubusercontent.com/u/215283?v=2" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [abliss](https://github.com/abliss)**
_Wednesday Oct 02, 2013 at 08:28 GMT_
_Originally opened as https://github.com/urbit/archaeology/pull/44_

---

No indentation support, but somewhat-accurate syntax highlighting.

---

_**[abliss](https://github.com/abliss)** included the following code: https://github.com/urbit/archaeology/pull/44/commits_

PR #5311, opened 2021-10-07
btc-wallet: import pre-dist settings as needed

Fang-:
If we have btc-wallet settings stored under the landscape desk,
reimport them if we haven't written it to the bitcoin desk yet.

Additionally, removes the bitcoin settings from the landscape desk.

This should resolve the error reported in #44, though certainly the frontend could stand to handle such failures more gracefully too.

Not entirely sure who to tag for review on, @timlucmiptev feel free to delegate as needed.

### DIFF
--- a/pkg/bitcoin/app/btc-wallet.hoon
+++ b/pkg/bitcoin/app/btc-wallet.hoon
@@ -24,6 +24,7 @@
     $%  state-0
         state-1
         state-2
+        state-3
     ==
 ::
 +$  state-0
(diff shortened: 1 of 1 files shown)


Fang-:
> This is a cross-desk interaction, meaning that it's worthwhile thinking about what happens if breaking changes are made to settings-store

May or may not be worth biting the bullet and having gall auto-generate `%u` responses based on whether a `%x` to the same path gives a result or not. Then we can (more) safely do the individual scries.

joemfb:
@belisarius222, @Fang-, this can't be safe in general -- if the kernel reloads as part of the same update, the entire agent namespace will block. It can only be safe if we know the version from which we'll be upgrading. And we can only know that (ie, the nature of "this specific update") *on average*. Consider what happens if previous updates failed on some ship, or if they turned upgrades off then on, or switched their OTA provider, &c. It seems like it would've been worth applying this migration on a timer and/or adding some manual command that could be used to force the settings migration.

PR #5336, opened 2021-10-19
landscape: fix failing scry causing messages to be improperly validated

liam-fitzgerald:
If we received a graph-node with that was signed by a past or future
life of a ship, then we would potentially crash scrying for the public
keys. Now, if there is a mismatch between lives, we simply no-op and
treat the signature as valid.


ryjm:
yeah i agree i just think we should have *something* here saying it's kinda sucky

PR #5337, opened 2021-10-19
docket: glob with site

ryjm:
This PR provides support for specifying both `glob` and `site` entries in a docket file. Previously, you could only use one or the other, effectively preventing developers from serving hoon based frontend (like sail) while also including frontend assets in a glob.

Apps with older docket files will still work, as the older `docket-0` marks are converted to `docket-1`. 

Since this requires a mark update for `docket`, it is a breaking change in that ships running an older version of the docket agent will be unable to install apps from ships running a newer version.


pkg/garden/sur/treaty.hoon
```diff
@@ -18,6 +18,20 @@
...
+::
+++  treaty-0
+  =<  treaty
+  |%
+  +$  treaty
+    [=ship =desk =case hash=@uv =docket-0:docket]
+  ::
+  +$  update
+    $%  [%ini init=(map [=ship =desk] =treaty)]
+        [%add =treaty]
```
Fang-, line 31: Hmm good q. I guess you can't `treaty(docket:docket docket-0)`... Pretty sure I've seen something weird be done for cases like this, but can't remember what it was.

pkg/garden-dev/lib/docket.hoon
```diff
@@ -17,23 +17,56 @@
...
+        u.info
+        u.color
+        u.href-0
+        image
+        u.version
+        u.website
+        u.license
+    ==
+  ::
+  ++  finalize
```
Fang-, line 51: Wondering if we just want a parameterized `+finalize`. You hand is the version, and it branches where it needs to. Instead of having two separate but largely-duplicate implementations.

pkg/garden/app/docket.hoon
```diff
@@ -50,20 +58,43 @@
 ++  on-load
   |=  =vase
   ^-  (quip card _this)
-  =+  !<(old=state-0 vase)
-  =*  cha  ~(. ch q.byk.bowl)
+  =/  maybe-old  (mule |.(!<(state-0 vase)))
+  ~!  +.maybe-old
+  =/  old=versioned-state
+    ?:  ?=(%& -.maybe-old)  [%0 +.maybe-old]
+  [%1 !<(state-1 vase)]
```
Fang-, line 65:
```suggestion
  =/  old=versioned-state
    ?:  ?=(%& -.maybe-old)  [%0 +.maybe-old]
    [%1 !<(state-1 vase)]
```

And, the comment about trying latest state version applies here too.

pkg/garden/app/docket.hoon
```diff
@@ -50,20 +58,43 @@
...
-  =.  -.state  old
-  =.  +.state  inflate-cache
-  `this
+  |-
+  ?-  -.old
+    %0  $(old [%1 (state-0-to-1 +.old)])
+    %1
+       =.  -.state  +.old
+       =.  +.state  inflate-cache
+       `this
```
Fang-, line 73:
```suggestion
    %1  `this(state [+.old inflate-cache])
```

pkg/garden/app/docket.hoon
```diff
@@ -50,20 +58,43 @@
...
-    ?.  ?=(%glob -.href.docket.charge)  ~
-    `:_(desk base.href.docket.charge)
+    ?.  ?=(^ spot.href.docket.charge)  ~
+    `:_(desk base.u.spot.href.docket.charge)
+  ::
+  ++  state-0-to-1
+    |=  sta=state-0
+    ^-  state-1
+    =/  garden=path  /(scot %p our.bowl)/garden/(scot %da now.bowl)
+    =+  .^(=dock=tube:clay %cc (weld garden /docket-0/docket-1))
```
Fang-, line 88: Here, too, we shouldn't be scrying during `+on-load`.

Fang- (changes requested):
You asked for it!

Mainly: the new `href` type can be better, and we shouldn't rely on marks for agent state conversion. Also a handful of style things.

`garden/ted/make-glob.hoon` is listed by gh as a binary file with just 1 byte? Maybe something went wrong during commit.

pkg/garden-dev/sur/docket.hoon
```diff
@@ -15,9 +31,19 @@
...
+::  $spot: full glob reference including the base desk
+::
++$  spot  (unit [base=term =glob-reference])
+::  $href: app links
 ::
 +$  href
+  $:  =spot
+  $%  [%glob ~]
+      [%site =path]
+  ==  ==
```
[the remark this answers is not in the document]
ryjm: I actually started with your suggestion, but couldn't get around some gnarly type errors - certainly could've been a brain fart or something. Will give it another shot.

pkg/garden-dev/sur/docket.hoon
```diff
@@ -1,12 +1,28 @@
 |%
 ::
+++  docket-0
+  =<  docket
+  |%
+  +$  docket
+    $:  %1
```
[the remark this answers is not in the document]
ryjm: The original docket type has a `%1` so I kept it... but yeah it should be a `%0`. Guess I can do that conversion at the mark level.

zalberico:
Closing this as stale with no clear path to getting it merged. @ryjm if I'm wrong about this happy to reopen.

ryjm:
the clear path is me actually doing it ;) it's just not worth the hassle of a mark version upgrade imho.

PR #5338, opened 2021-10-20
gall: add +on-rift arm to agents

Fang-:
Instead of subscribing only to breach notifications for ships that its
agents have interacted with, gall now listens to all breach
notifications across the network.

These notifications now get passed into all running agents through
+on-rift, taking the involved ship as an argument.

Notably, +on-rift is called _before_ an agent's subscriptions get
cleaned up. This ensures agents have full information for deciding how
to handle the breach. But because +on-leave and +on-agent (%kick) may
still be called afterwards, the agent is advised to not clean up
subscription state during +on-rift handling.

Developed together with @philipcmonk after getting bitten by an "app didn't know about breach" situation. This has been a long time coming!

We should hold off on merging & releasing this for a month or two, and do outreach to developers to let them know this is coming so they can prepare. We should also upgrade the apps in the other desks as part of this PR before it gets merged.

@joemfb tagging you for kelvin bump ops.

### DIFF
--- a/pkg/arvo/app/acme.hoon
+++ b/pkg/arvo/app/acme.hoon
@@ -383,6 +383,7 @@
       ==
     [cards this]
   ::
+  ++  on-rift   on-rift:def
   ++  on-fail   on-fail:def
   --
 ::
--- a/pkg/arvo/app/aqua.hoon
+++ b/pkg/arvo/app/aqua.hoon
@@ -119,6 +119,7 @@
         (handle-wake:ac wen)
       [cards this]
     ==
+  ++  on-rift   on-rift:def
   ++  on-fail   on-fail:def
   --
 ::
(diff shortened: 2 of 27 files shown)


belisarius222 (commented):
My only concern with this is race conditions: Jael learns a ship breached, then tells Ames, which tells Gall -- I think then Jael tells Clay.  This means that agents could be running their +on-rift arms before Clay has learned about the breach.

Maybe that's ok in practice, but it's something we should be careful about, especially for agents like Kiln -- what should it do if an upstream breaches, for instance?

zalberico:
@Fang- and @belisarius222 is this something we still want to clean up and merge? Not clear to me from the last couple of comments (otherwise we should close it).

belisarius222:
Last I heard, we're not sure this is the approach we want to take for improving userspace breach handling.

—
~rovnys-ricfer
https://tlon.io

PR #5343, opened 2021-10-21
Release/bitcoin wallet

finned-palmer:
Some bitcoin wallet updates that got left behind post app distribution. The invoice fixes are pretty critical to making the app not annoying to use.

cc @timlucmiptev @liam-fitzgerald @arthyn @jalehman


liam-fitzgerald:
@finned-palmer adhering to the usual commit message format would be appreciated

PR #5345, opened 2021-10-21
conn C driver

mrdomino:
This is just the C driver itself, based off of `next/vere`.

It will work fine without the `%khan` vane, except that threads (`%fyrd`) will not work.


pkg/urbit/vere/io/khan.c
```diff
@@ -0,0 +1,477 @@
...
+  can_u->mor_u.nex_u = (u3_moor*)san_u->can_u;
+  san_u->can_u = can_u;
+}
+
+/* _khan_sock_init(): initialize socket device.
+*/
+static void
+_khan_sock_init(u3_shan* san_u)
+{
+  // The full socket path is limited to about 108 characters, and we want it to
```
locpyl-tidnyd, line 194: I see this protocol is intended for control backplane in hosted scenarios, so idk if it should exist on Windows at all, but just in case - the closest equivalent of Unix sockets on Windows, named pipes, do not share a namespace with the filesystem and cannot be linked/mounted into it relative to anything. Easiest way is probably to create a pipe name based on ship name, something like `\.\pipe\%khan-sampel-palnet`.

ryjm: it's first use case is for that, but it will definitely not be only for that. it should certainly exist on windows.

mrdomino: for now i'm no-opping on windows; happy to come back to this and make pipes work later.

mrdomino:
mingw failure is not erroneous; i just forgot about windows with the ivory pill change. Downloading logs to look at the mac failure.

mrdomino:
mac failure looks erroneous unless i'm mistaken; deep in the weeds of trying to build clang.

mrdomino:
@joemfb I think we're basically there. I'd like to get this reviewed and merged before the 1.9 release. The vane tests are not quite finished yet, but I believe @philipcmonk said he was fine with the driver getting merged ahead of the vane assuming no driver interfaces would need to change (which I think is the case.)

joemfb (changes requested):

pkg/urbit/vere/ivory.c
```diff
@@ -0,0 +1,3 @@
+#include "vere/ivory.h"
+
+#include "ivory_impl.h"
```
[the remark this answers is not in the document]
mrdomino: yes. it seems slightly better to expose a header file that is included into an in-project C file, than to expose a C file that is added to the sources or something.

pkg/urbit/vere/io/conn.c
```diff
@@ -0,0 +1,963 @@
...
+_conn_moor_poke(void* ptr_v, c3_d len_d, c3_y* byt_y)
+{
+  u3_weak             jar;
+  u3_noun             can, rid, tag, dat;
+  u3_chan*            can_u = (u3_chan*)ptr_v;
+  u3_conn*            con_u = can_u->san_u->con_u;
+
+  jar = u3s_cue_xeno_with(con_u->sil_u, len_d, byt_y);
+  if ( u3_none == jar ) {
+    can_u->mor_u.bal_f(can_u, -1, "cue-none");
```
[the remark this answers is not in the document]
mrdomino: switched to using `goto _moor_poke_out` at call sites.

pkg/urbit/tests/boot_tests.c
```diff
@@ -0,0 +1,68 @@
...
+       (0x7fffffff != lit_l) ) {
+    printf("*** fail _test_lily-2a\n");
+    exit(1);
+  }
+  cod = u3dc("scot", c3__ux, u3i_word(0x80000000));
+  if ( c3y == u3v_lily(c3__ux, cod, &lit_l) ) {
+    printf("*** fail _test_lily-2b\n");
+    exit(1);
+  }
+}
```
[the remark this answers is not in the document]
mrdomino: this would require moving the boot sequence from `_setup` here into `noun_tests`. assuming that's ok, can do.

pkg/urbit/vere/io/conn.c
```diff
@@ -0,0 +1,918 @@
...
+        while ( car_u ) {
+          if ( car_u->io.info_f ) {
+            res = u3nc(
+              u3nc(car_u->nam_m, car_u->io.info_f(car_u)),
+              res);
+          }
+          car_u = car_u->nex_u;
+        }
+        res = u3nc(u3_nul, res);
+      } break;
```
joemfb, line 492: I think we want to move this into `u3_auto_info()`, and call that from a `u3_pier_info()`, which should do the check for `u3_psat_work`.

joemfb (changes requested):
A refactoring request for the new `info` functionality. And a bump for the unit tests changes requested above.

pkg/urbit/vere/io/conn.c
mrdomino: done. `u3_pier_slog` (previously named `…_info`) does some extra work, which I mostly replicated in the new `u3_pier_info`. i skipped disk and lord info for now for expedience and since we may still change the output type.

mrdomino:
@joemfb I switched the type for the info functions to `$mass`; seems way cleaner. Any other feedback?

pkg/urbit/vere/auto.c
```diff
@@ -353,10 +353,26 @@ u3_auto_exit(u3_auto* car_u)
...
+{
+  u3_noun lit = u3_nul;
+
+  while ( car_u ) {
+    if ( car_u->io.info_f ) {
+      lit = u3nc(u3_pier_mass(car_u->nam_m, car_u->io.info_f(car_u)), lit);
+    }
+    car_u = car_u->nex_u;
+  }
+  return lit;
```
joemfb, line 369: I think we should either flop this so it's produced in driver link-order, or alphabetize by driver name. The former is easier, so I'd just do that.

pkg/urbit/vere/disk.c
```diff
@@ -669,7 +669,7 @@ u3_disk_info(u3_disk* log_u)
   while ( red_u ) {
     lit = u3nc(
       u3_pier_mass(
-        c3__read,
+        u3dt("cat", 3, c3__read, u3dc("scot", c3__ux, u3i_chub((c3_d)red_u))),
```
joemfb, line 672: I don't think this kind of naming convention will be useful to consumers. But I expect we'll need to iterate on this interface quite a bit. The module-specific `%info` implementations will also have to change quite a bit due to the mars/urth architecture. So these should be considered first drafts. (Nothing to do here, just noting ...)

joemfb (changes requested):
This looks about ready to go. There's a header conflict, and a couple minor requests below:

pkg/urbit/vere/disk.c
mrdomino: yeah, agreed. no obvious way to change this while preserving the constraint that flattened variable names are unique, so i'll leave it as-is for now.

mrdomino: (another option would have been to split these out into multiple levels - `c3__read` at top level, and the address below. that would mildly help clients who were depending on the hierarchical structure, but not clients who were depending on the flattened variable name. can do it if desired.)

mrdomino: (went ahead and did this)

joemfb (approved):
LGTM

PR #5351, opened 2021-10-22
%khan vane

mrdomino:
This is the hoon side of the work in #5345. It is still a very rough cut and needs some love.


mrdomino:
Superseded by the auspiciously numbered #5420 on the hoon side. Haskell driver branch TODO.

PR #5360, opened 2021-10-25
gen: add |new-desk for creating minimal desks

Fang-:
The created desk includes the bare minimum of necessary files, sourced from base (by default), or some desk of the user's choosing.


belisarius222:
Not all Clay commits are diffs, which might be why the missing codepath isn't always triggered.  I think we do need %txt in there for now, though.

rabsef-bicrym:
I beg of thee, bring this into core.

philipcmonk (approved):

rabsef-bicrym:
Yay!

matildepark:
Thank you Phil. I've never seen the boy so happy like this.

Issue #5329, opened 2021-10-13
@urbit/http-api should have a simple means of building it into a single file


tacryt-socryp:
I get that so far only React apps have used @urbit/http-api and all have already had complicated JS build tooling in place from the get-go, but this is a non-starter for people trying to build vanilla JS webpages.

tacryt-socryp:
Here's a sample script that I got working with some modifications to the files to ensure Urbit is exported: https://gist.github.com/tacryt-socryp/c24f60c851b4fdfd3e28a351f7bd0a5f

PR #5371, opened 2021-10-26
npm: fix builds for npm packages

finned-palmer:
The npm packages were previously targeting ESNext and ES2020 on build, and they were also using esmodules rather than commonjs.

See this issue: https://github.com/urbit/urbit/issues/5329

Because of that setup you actually couldn't even run `npm start` in the `pkg/interface` directory to run landscape locally.

probably best for @arthyn to review this one.

### DIFF
--- a/pkg/npm/api/tsconfig.json
+++ b/pkg/npm/api/tsconfig.json
@@ -1,10 +1,11 @@
 {
   "compilerOptions": {
     "baseUrl": ".",
-	  "outDir": "./dist",
-    "module": "ES2020",
+    "outDir": "./dist",
+    "module": "CommonJS",
     "noImplicitAny": true,
-    "target": "ES2017",
+    "target": "ES5",
+    "downlevelIteration": true,
     "pretty": true,
     "moduleResolution": "node",
     "esModuleInterop": true,
--- a/pkg/npm/http-api/tsconfig.json
+++ b/pkg/npm/http-api/tsconfig.json
@@ -2,10 +2,11 @@
   "include": ["src/**/*"],
   "exclude": ["node_modules", "dist", "@types"],
   "compilerOptions": {
-	  "outDir": "./dist",
-    "module": "ESNext",
+    "outDir": "./dist",
+    "module": "CommonJS",
     "noImplicitAny": true,
-    "target": "ESNext",
+    "target": "ES5",
+    "downlevelIteration": true,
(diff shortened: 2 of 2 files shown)


arthyn:
@finned-palmer can you try this now? We just published the new versions, though you may have to manually bump the versions to 2.0.0 in package.json until the version bumps get merged to master.

finned-palmer:
Same issue, you can't use these packages outside of a TS project because they don't compile down to javascript.

tacryt-socryp:
I had this exact issue. There needs to be better support for vanilla JS with and without your favorite modern web framework.

arthyn:
Alright I've got a PR for this going here: https://github.com/urbit/urbit/pull/5411

If y'all want to play with the packages you should be able to install them like `npm i @urbit/api@dev` or alternatively include as a script tag like `<script src="https://unpkg.com/@urbit/api@2.1.0-dev/dist/urbit-api.min.js">...` to get the dev release of what's in the PR above. Still need to document what the globals look like if you use the script tag, but they're available as `UrbitHttpApi` and `UrbitAPI`. Which I'm realizing now is inconsistent lol. I'll fix that before the official release and provide some documentation.

tacryt-socryp:
Thank you!

finned-palmer:
Thanks so much for working on this, @arthyn. I'm on the 2.1.0-dev packagers, still seeing loader issues around TS files. If you want to try to reproduce this just run npm start on the landscape interface as a separate package.

arthyn:
So I just looked deeper into this and it's a multi-layer issue. Because right now we ship all the files of @urbit/api including source with the package, that gets picked up in auto-complete when adding imports. So when we have things like `import from '@urbit/api/lib...` etc. it's actually literally picking up typescript files instead of the actual package's JS for distribution. In the middle of fixing right now, I'll keep you posted.

finned-palmer:
Awesome, much appreciated!

arthyn:
Ok so I dug really deep today. If you go to the PR you can see all the changes that were necessary starting with this one https://github.com/urbit/urbit/pull/5411/commits/c07512b9de5f31de1f19144d7964fd35cbd73a78. 

This should be able to be independently pulled out from the repo now. I tested on a fresh folder outside of the repo just now and it built fine.

finned-palmer:
It works! That's great.

PR #5385, opened 2021-10-28
into: %khan client

mrdomino:
This is to be a library that hosting can use, and also an executable that can be used by anyone for one-shot commands.

cc @ryjm @brendanhay @jalehman


mrdomino:
As of the last commit pushed, `into` works. Example usage (assuming you're running a pier with `urbit` from #5345 and `arvo` from #5351 and you've booted the pier with `-h` and you're in the `pkg/hs` directory):

```
% stack run into ~/zod reset
+code reset.
% stack run into ~/zod code
~hoprus-ritpem-haldur-fannep
```

I suppose it's desirable to extract a `runInto` or something that takes a command/response and sends it over a connection. That's still TODO for now.

PR #5387, opened 2021-10-28
hark-graph-hook: correctly get rear of index

liam-fitzgerald:
- Correctly get ID from index, was incorrectly fetching version numbers from notes instead of the note ID
- Works around malformed unreads by dismissing them


ryjm (approved):
LGTM, though i don't know what that failing test is? does that need to be addressed?

liam-fitzgerald:
no, but we ought to fix the frontend CI

PR #5389, opened 2021-10-28
autopill: make pills when desk contents change

### LINKED ISSUE #5364: Fix the pill situation
Instead of building pills ad hoc and uploading them to boostrap.urbit.org, build them directly on ~zod and download them directly from there.

This would mean that pills would more or less always be up to date.

Fang-:
As discussed previously:  
Automatically builds pills and writes them to `.urb/put` when relevant desks change. Intended for deployment onto a livenet ship from which pills may then be sourced.  
Closes #5364.

~~This PR brought to you by overengineering incorporated.~~ This implements the fully general solution to the problem of "I want to auto-export fresh pills". In practice we'll likely only want this for one or two kinds of pills, but the sole interface is mandatory to be able to output `%sag`s under current dill, and it seems easier to operate this way.

Original intent was to deploy this to ~zod and put a thin server in front of it to expose the `.urb/put/pills/`. If we're concerned about compilation times keeping it too busy, we can always spin up a moon or other such ship and go from there, the delay between ~zod's kids and the "~piller" ship seems unlikely to bite us, and might even be better than no delay at all.


gusmacaulay:
does this solve the problem of freshly created fakezods being out of date?

Fang-:
The best way to boot fakezods for development is to just have a checkout of the urbit/urbit repo, and then `urbit -F zod -B bin/solid.pill -A pkg/arvo` in order to make sure it starts up with the latest files included.

I should go ahead and close this PR for now. We'll manually update the pill after the next OTA as a smaller, simpler band-aid. Afterwards, we'll work towards improving the boot sequence, with the hope of delivering something akin to a "pick your pill" situation. /app/autopill will likely be repurposed to support that.

PR #5391, opened 2021-10-29
vercel: remove .vercel from root

### LINKED ISSUE #5378: `.vercel` is in file tree when maybe it shouldn't?
Stumbled across this
```
> Should I commit the ".vercel" folder?
No, you should not share the ".vercel" folder with anyone.
Upon creation, it will be automatically added to your ".gitignore" file.
```

¯\\\_(ツ)_/¯

Fang-:
It's in the .gitignore and was never supposed to be included. Closes #5378.

@arthyn does this break vercel-related CI? If yes, shouldn't these variables be in some gh actions field or something?


arthyn (approved):

arthyn:
Never meant to put these in, thanks!

PR #5397, opened 2021-11-02
jets: add lib/tiny tree for L2

philipcmonk:
Layer 2 "Naive Rollups" uses lib/tiny as its standard library.  This
duplicates the jet hooks from the standard library to a new tree with
only those jets.

We decrement the permitted kelvin zuse version so that when L2 is
released, it can decrement zuse's kelvin, forcing users to upgrade to
this version.  Otherwise, users would fail to progress because all L2
computation would be unjetted.

The proposed lib/tiny can be found in urbit/azimuth#49.


pkg/urbit/jets/tree.c
```diff
@@ -2114,8 +2114,75 @@ static c3_c* _k140_ha[] = {
...
+  { "dis", 7, _140_two_dis_a, 0, _140_two_dis_ha },
+  { "dor", 7, _140_two_dor_a, 0, _140_two_dor_ha },
+  { "end", 7, _140_two_end_a, 0, _140_two_end_ha },
+  { "gor", 7, _140_two_gor_a, 0, _140_two_gor_ha },
+  { "lsh", 7, _140_two_lsh_a, 0, _140_two_lsh_ha },
+  { "met", 7, _140_two_met_a, 0, _140_two_met_ha },
+  { "mix", 7, _140_two_mix_a, 0, _140_two_mix_ha },
+  { "mor", 7, _140_two_mor_a, 0, _140_two_mor_ha },
+  { "mug", 7, _140_two_mug_a, 0, _140_two_mug_ha },
+  { "muk", 59, _140_two_muk_a, 0, _140_two_muk_ha },  //  TODO: why 59?
```
joemfb, line 2163: The short answer is that the `+muk` pins the `|fe` core *outside* of itself. The long answer is that but a lot funnier ...

pkg/urbit/jets/tree.c
```diff
@@ -2114,8 +2114,75 @@ static c3_c* _k140_ha[] = {
   0
 };
 
+
+//  TODO: probably need different ha hashes
```
joemfb, line 2118: Yes, but not pressing.

joemfb (approved):
LGTM (modulo ci)

PR #5398, opened 2021-11-02
vere: abort on checkpoint sync errors

joemfb:
And discard invalid checkpoint patches on restart. (That was needlessly treated as a fatal error, dropping the patch just means a short event log replay.)


philipcmonk (approved):
LGTM

joemfb:
Merging, as the CI failures are fixed in #5397

PR #5400, opened 2021-11-02
SidebarItem: fix status priority, if graph is not cleaned up correctly

liam-fitzgerald:
If the graph is not cleaned up properly (e.g. directly poking with %remove-graph), then we may still have unreads for it. To remedy this, the unsubscribed state should take priority over any potential notification state


liam-fitzgerald:
cc: @baudtack

pkg/interface/src/views/landscape/components/Sidebar/SidebarItem.tsx
```diff
@@ -24,13 +24,13 @@ function useAssociationStatus(resource: string) {
...
+  return !isSubscribed
+    ? 'unsubscribed'
+    : hasNotifications
     ? 'notification'
     : hasUnread
     ? 'unread'
-    : isSubscribed
-    ? undefined
-    : 'unsubscribed';
+    : undefined;
```
arthyn, line 33: Can we possibly break this up into if statements so it's readable? 🤣

arthyn (changes requested):
Hard to read, but logic seems correct

PR #5403, opened 2021-11-03
useResize: fix destructor

liam-fitzgerald:
Interior mutability bites again. This was causing a memory leak, as the reference to the element was never let go


arthyn (approved):
Woah that's weird

PR #5405, opened 2021-11-04
chat-cli: link to drum on-init

liam-fitzgerald:
Allows freshly booted ships to ctrl-x into chat-cli. Even though we unconditionally `|link` in the +on-load, you can disable %chat-cli with `|rein %landscape [| %chat-cli]` so no functionality is lost. 

cc: @belisarius222 I know that linking agents was something that early prototypes of software-dist kiln did, do you think this is the right pattern

### DIFF
--- a/pkg/landscape/app/chat-cli.hoon
+++ b/pkg/landscape/app/chat-cli.hoon
@@ -213,7 +213,7 @@
   |=  old=(unit versioned-state)
   ^-  (quip card _state)
   ?~  old
-    [~[connect] state(width 80)]
+    [~[connect link] state(width 80)]
   ::
   =?  u.old  ?=(?(~ ^) -.u.old)
     ^-  state-1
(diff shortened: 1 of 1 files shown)


joemfb:
drum auto-linking was removed?

liam-fitzgerald:
Yeah earlier versions of dist kiln had the bill file specify a list of
agents to automatically link to. This is because app starting and stopping
now occurs in kiln, not drum

Fang-:
> Allows freshly booted ships to ctrl-x into chat-cli

I'm inclined to oppose this until the (chat-)cli experience becomes good again. Not sure what we gain from letting freshly booted ships trivially access it. Surely `|link %chat-cli` is a reasonable barrier to entry here?

philipcmonk:
Fact becomes policy, so closing this unless we actually want to change this behavior

PR #5406, opened 2021-11-06
jael, azimuth: reconfigure for L2

Fang-:
Jael needs to be reconfigured to listen to the new aagent for azimuth
events, and the old app needs to be shut down. We do this in
/app/azimuth's +on-init.

Additionally, we make sure that jael doesn't crash when it (as expected)
loses its subscription to the old agent.

Only tested on a fakenet ship so far, which might not hit the full/relevant case here.

### DIFF
--- a/pkg/arvo/app/azimuth.hoon
+++ b/pkg/arvo/app/azimuth.hoon
@@ -59,7 +59,10 @@
     ^-  (quip card _this)
     =.  net.state  %local
     :_  this
-    [%pass /eth-watcher %agent [our.bowl %eth-watcher] %watch /logs/[dap.bowl]]~
+    :~  [%pass /eth-watcher %agent [our.bowl %eth-watcher] %watch /logs/[dap.bowl]]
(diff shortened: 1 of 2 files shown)


pkg/arvo/app/azimuth.hoon
```diff
@@ -59,7 +59,10 @@
     ^-  (quip card _this)
     =.  net.state  %local
     :_  this
-    [%pass /eth-watcher %agent [our.bowl %eth-watcher] %watch /logs/[dap.bowl]]~
+    :~  [%pass /eth-watcher %agent [our.bowl %eth-watcher] %watch /logs/[dap.bowl]]
+        [%pass /old-tracker %agent [our.bowl %azimuth-tracker] %poke %kiln-nuke !>([%azimuth-tracker %base])]
```
philipcmonk, line 63: Shouldn't this be aimed at %hood?

Fang-: Most definitely! Not sure how this slipped in.

PR #5407, opened 2021-11-06
hood: fix |rein when starting/stopping multiple agents

yosoyubik:
Realized that doing something like this `|rein %base [& %roller] [& %roller-rpc] [& %azimuth-rpc]` would only update the last agent in the list.


philipcmonk (approved):
LGTM

PR #5408, opened 2021-11-08
clay: Add support for representing tombstones

philipcmonk:
This adds support for tombstoned files to clay.  It does not include any
way to actually tombstone them; that is left for later.

This allows tombstoning at the level of a file.  Precisely, this expands
+blob:clay by adding a %dead case:

    +$  blob                                              ::  fs blob
      $%  [%delta p=lobe q=[p=mark q=lobe] r=page]        ::  delta on q
          [%direct p=lobe q=page]                         ::  immediate
          [%dead p=lobe ~]                                ::  tombstone
      ==                                                  ::

Thus, we maintain the invariant that every lobe corresponds to a blob,
but now a blob may be an explicit tombstone.

Details:
- This has not been tested at all, except that it compiles and boots.
- This does not have a state adapter from master.  The only state change
  is the definition of +cach.
- Additionally, out-of-date ships may unexpectedly receive a %dead blob
  from a foreign clay which would interfere with their ability to download
  that desk.  No code changes necessary, but sponsors should avoid
  tombstoning files in %base for a while so their children can get the
  update.
- A merge will only fail if the tombstoned file conflicts with another
  change.  Note that as written, merging from a past desk *can* bring a
  tombstoned file to the head of a desk.  Possibly this shouldn't be
  allowed.

This also includes several refactors that were made possible by ford
fusion (since everything is synchronous now) but never got done.  In
both cases we get to remove a monad, which simplifies the code
considerably.

- refactor +merge's error handling to use !!/mule instead of threading
  through errors
- refactor all +read-* functions and related parts of +try-fill-sub to
  eagerly convert lobes to cages.
- refactor +wake to remove several abstractions that are no longer
  necessary

We also add support reading %a/b/c/e/f/r/x from past and foreign desks,
when possible.  This was left underimplemented since ford fusion.

Apologies that all of these are in one commit, it was
all a single chunk of work.

This is a draft until we have a way to tombstone.  I suspect we'll want
to have a mechanism of keeping track of gc roots and trace to remove,
but this PR doesn't suggest any particular strategy.


ohAitch:
> Note that as written, merging from a past desk *can* bring a tombstoned file to the head of a desk.

Or to the head of an entirely unrelated desk perhaps?

Straight-up deleting blobs(after doing the trace pass, or perhaps some sort of reference counting) and maintaining a parallel dead-hash set seems much safer for like, `~(has by file-store)` doing what it says on the tin, if you pull in an indirect it not mistakenly believing the base is still available, etc. Tombstoning should imo be an operation on *sets of revisions*(invoking gc), perhaps with a separate "I don't want this file reject future updates that contain it" that *eagerly* faults instead of leaving your content-addressed store in an inconsistent state.

ohAitch:
To expand on "sets of revisions": tombstoning is not an operation on the present, but rather on the past. If you delete a file normally, you get a revision without it; if you merge that revision with upstream, this cleanly maintains one where it's gone. (I forget if clay has the concept of moves, if those are just del+add and del+del don't conflict this might cause problems, tombstone or no)

The question then is getting the file out of some finite list of beaks you've already seen, fixed at the time you run the tombstone command. *your* beaks primarily, anything referenced only by foreign beaks you can safely delete anyway, at the cost of later re-download. Perhaps you point the generator at one beam, it rattles off every known alias to it(compressing ranges, `/=/base/[29-42]/some/file`), you confirm, clay stores per-desk "here lies hash x, deceased as of revision y".

imo an OTA involving tombstones should outright fail, the child is asking for the version of the desk that was deployed, not your current opinion on what parts of that version of the desk are still relevant - either you can satisfy this request or you cannot. Otherwise you get the remote retroactive deletion situation you described.

philipcmonk:
> Straight-up deleting blobs(after doing the trace pass, or perhaps some sort of reference counting) and maintaining a parallel dead-hash set seems much safer for like, ~(has by file-store) doing what it says on the tin

Many, even most, operations work fine on tombstoned blobs.  Even merges only fail if you *actually* need to inspect the contents.

I agree that tombstoning should follow a gc pattern, so "delete" should mean "mark this as a weak reference" so that a gc pass will throw it away.

Useful commands:

- delete a set of paths (recursively) across a set of revision ranges
- dry-run of the above, estimating savings and specifying which files will *not* be deleted because they're referenced elsewhere
- set a policy for a desk to always tombstone everything except the most recent revision.  Another policy might be to always tombstone old js files.
- given a file, list all the revisions/files that refer to it (i.e. list all backreferences)

> imo an OTA involving tombstones should outright fail,

I really think everything should succeed to the best of its ability.  You don't have the data, but you have metadata -- might as well give that to me.  Maybe I have the data, or maybe I don't care I just want a hash anyway.  Almost any artificial constraint in this area has useful counterexamples.

> (I forget if clay has the concept of moves, if those are just del+add and del+del don't conflict this might cause problems, tombstone or no)

(It does not have moves)

ohAitch:
> > imo an OTA involving tombstones should outright fail,

> I really think everything should succeed to the best of its ability. You don't have the data, but you have metadata -- might as well give that to me. Maybe I have the data, or maybe I don't care I just want a hash anyway. Almost any artificial constraint in this area has useful counterexamples.

> Many, even most, operations work fine on tombstoned blobs. Even merges only fail if you actually need to inspect the contents.

I think "inspect" is understating it: most operations on files, such as mounting them to unix, running the contained JavaScript, determining their *size*, will crash in the presence of tombstones. If you have a revision of landscape that currently runs, and you attempt to merge a newer version whose contents the sender has forgotten, the way for that operation to succeed to the best of its ability is to conflict the merge, and not give you a checkout that actively contains a tombstone.

If the upstream has tombstoned some JavaScript you're still using while you have modified it, you keep the modified version, but that does not entail actually *sending* a `%dead` over the network: you already had the real contents! And of course if you have *modified* it you probably now have an `%indirect`, and that `%indirect` should certainly keep working.

ohAitch:
I suppose that might be my core question: what does it mean for the current version of a desk to contain a tombstone instead of a deletion? Past versions of a desk getting tombstones OTA'd in is not a problem, because you don't merge into the past. (And a transmitted `%dead` won't do anything if it's rooted in previous local beaks already, and won't be transmitted if merge determines the reference to it is gone in the first place, so seems only relevant if you're actively constructing a new revision with paths you know ahead of time are unfulfillable)

philipcmonk:
> I think "inspect" is understating it: most operations on files, such as mounting them to unix, running the contained JavaScript, determining their size, will crash in the presence of tombstones.

Yeah but not +ls, -diff, most merges (where the mergebase is tombstoned), hashing to check versions, etc.  That's stuff you might want to do with old versions and should work fine.

> If the upstream has tombstoned some JavaScript you're still using while you have modified it, you keep the modified version, but that does not entail actually sending a %dead over the network: you already had the real contents! And of course if you have modified it you probably now have an %indirect, and that %indirect should certainly keep working.

Whatever we get over the network is added to our local blob store without overwriting anything that currently exists: `(~(uni by new-blobs) old-blobs)`.  This is what caused `%delta` loops a couple years ago, so we've been disciplined about making sure that always happens.

> I suppose that might be my core question: what does it mean for the current version of a desk to contain a tombstone instead of a deletion?

This is why I lean toward saying the current version of a desk can never have a tombstone.  If you do something that would cause this, +park (which handles all commits ultimately) should catch it in +sane-changes and bail the commit.

belisarius222:
I'm curious why you picked tombstoning in the store, vs. in the domes that point to the store. The data structures used here imply a user would delete a file *by hash*, as opposed to by path. I think this would make it harder to enforce data retention policies.

Let's say desks foo and bar contain files whose values hash to 0xdead.beef. The way this is written, I could tombstone 0xdead.beef, effectively deleting both files, but there's no way to tombstone just one file.

From one perspective, this is nice, since a tombstoning operation corresponds directly to freeing memory. From another, though, it makes memory management more difficult.

Let's say the user has a policy that desks foo and bar don't data from old revisions. If so, then a garbage collection routine will have to decide for each file file whether any desks point to that file's hash before tombstoning it.

Maybe this is workable, although it's a bit hard for me to picture. It seems more difficult than the alternative, which we had discussed before: each "GC root" (file listing in a dome) pointing into the hash-addressed store would gain a "dead" bit. If the bit is true, then it's only a weak pointer and will not prevent that data from being deleted during a GC pass.

This would allow Clay to store *per-file* deletion state naturally, and then it's up to Clay to trim its hash-addressed store at some point, probably with a copying GC pass. Representing the fact that the user doesn't want to hold a file's value seems more helpful in managing memory, since instead of directly manipulating the shared hash-addressed store, a tombstoning operation could straightforwardly correspond to user-configured per-desk policy about data retention.

philipcmonk:
Yes, a bit on the dome per-file is still the intention.  That's a clean layer on top of this, or rather in parallel to the existing `%tomb` operations.  Something like this: https://github.com/urbit/plan/blob/4850b53e4405b079592e86f128e79b683c12a94c/oort/clay-tombstoning.hoon

The `%tomb` options here are just the dumbest, simplest interface I could build.  I think they're useful to keep, where `[%tomb %lobe =lobe]` is useful for surgical rescue operations and `[%tomb %all ~]` is "save as much memory as we safely can right now".  But those are both for exceptional situations.

PR #5409, opened 2021-11-08
Azimuth L2 / Naive Rollup

yosoyubik:
Includes all code necessary to support naive rollups, the Layer 2 solution for the reduction of Urbit ID gas prices.

The main addition is `/lib/naive.hoon` (the first "Hoon smart contract")

The relevant agents are:

- %azimuth - obtains and holds PKI state.
- %azimuth-rpc - JSON RPC-API for %azimuth.
- %eth-watcher - Ethereum event log collector.
- %roller - submits batches of L2 transactions to Ethereum.
- %roller-rpc - JSON RPC-API for %roller.

([Here](https://urbit.org/docs/azimuth/flow#gall-agents) is a more detailed description on these agents)

%azimuth is required for all live ships, and will start upon booting a new ship or receiving the corresponding OTA, since its purpose is to receive updates about Urbit's PKI, for example key changes.

The remaining agents (%roller, %roller-rpc, and %azimuth-rpc) are meant to be run by a Roller operator that will submit batches of L2 transaction to Ethereum, and therefore will be paying fees to include those in the blockchain. It's also possible to run the roller in a passive mode (without a private key loaded) and use it as a very simple index of Azimuth data—right now it tracks ownership and sponsorship of ships.

Some useful commands to interact with the roller agents:

- `|rein %base [& %roller] [& %roller-rpc] [& %azimuth-rpc]` : starts all roller agents
- `:roller|local`: connects to a local ethereum node (needs to be run manually by the roller operator)
- `:roller|ropsten`: connects to a ropsten infura node
- `:roller|endpoint 'URL' %local`: connects to an eth node, for the given type of network (can be %local, %ropsten, or %mainnet)
- `:roller|setkey 'abcd1234'`: loads a hex-encoded private key
- `:roller|config ...`: general configuration for the roller
- `:roller|commit`: commit pending  transactions
- `:roller|frequency ~h1`: changes how often the roller submits L2 batches
- `:roller|quota 32`: updates the max transaction quota per ship, per slice
- `:roller|slice ~d7`: update the time range (e.g. ~d1, ~h1...) where transactions are allowed to be sent

To interact with the roller outside of urbit, there's a [WIP TypeScript library](https://www.npmjs.com/package/@urbit/roller-api) that sends commands via HTTP. Documentation for this library is still being worked on, but a draft document of how to manually construct these calls can be found [here](https://documenter.getpostman.com/view/16338962/Tzm3nx7x)


pkg/arvo/ted/eth-watcher.hoon
```diff
@@ -80,26 +82,94 @@
...
     (pure:m pup(blocks ~))
   =/  to-number=number:block
-    (min up-to-number (add number.pup zoom-step))
+    =;  step
+      (min up-to-number (add number.pup step))
+    ::  Between "launch" (6.784.800) and "public" (7.033.765) blocks,
+    ::  there are a lot events belonging to all the pre-ethereum ships
+    ::  being established on-chain. By reducing the step, we avoid crashing.
+    ::
+    ?.  =(contracts:azimuth mainnet-contracts:azimuth)
```
yosoyubik, line 104: @philipcmonk I had to comment this line when getting all blocks on mainnet so it wouldn't use 100.000 blocks all the time (it came from this commit: https://github.com/urbit/urbit//commit/781cdf39606dcad0dcdbd4ae71764ec0aca13aea)

philipcmonk: I don't understand, isn't the point of this line to make sure that on mainnet it uses smaller block batches?  On mainnet shouldn't this condition evaluate to true?

yosoyubik: I think i found the issue, we need to change this https://github.com/urbit/urbit//commit/781cdf39606dcad0dcdbd4ae71764ec0aca13aea#diff-f78b5c5642b98a402f139a465bdfcb91309f95ce3c5574b906d12258335ad692R81 to `mainnet-contracts`

PR #5410, opened 2021-11-08
kiln: fix crashing when reloading kiln after a new kelvin

liam-fitzgerald:
kiln: prevent unnecessary bumps when reloading 

If a desk is suspended/paused/local, then we should exclude it when
bumping in the +on-load, because a suspended/paused/local desk cannot
have a pending commit that would allow it to bump successfully.

kiln: respect .liv flag in +get-apps-want

Accounts for the value of .liv in +get-apps-want, such that if a desk
is suspended then +get-apps-want should always return the empty list.

h/t @Fang- for pairing on this


philipcmonk (approved):
looks correct

PR #5411, opened 2021-11-09
npm: change build tool for wider distribution

arthyn:
This adds rollup as a build tool for both @urbit/api and @urbit/http-api. This gives us better capabilities to bundle and create package distributions for various formats. This PR also removes lodash as a dependency from @urbit/api because it is extremely heavy, and if it was made a peer dependency instead, it would make this library more difficult to consume.

I'm not exactly sure how we should version this, possibly both 2.1.0. I don't think we broke any backwards compatibility, but if you could watch out for that it would be appreciated. I'm going to publish a dev version before these officially go out so that people can start trying them out.

Something that is kind of awkward is that the old school "script tag output" for http-api has to be used like `UrbitHttpApi.Urbit` because we're exporting some error classes instead of just the single class. It would be nice to maybe refactor those into the Urbit instance itself. Maybe some static class method like  `Urbit.createFatalError(msg)` or similar so that the global that gets create is just the `Urbit` class itself.


pkg/npm/http-api/example/index.html
```diff
@@ -1,98 +1,114 @@
...
+      </p>
+      <pre>
+window.airlock = await UrbitHttpApi.Urbit.authenticate({
     ship: 'zod',
     url: 'localhost:8080',
     code: 'lidlut-tabwed-pillex-ridrup',
     verbose: true
 });
-window.airlock.subscribe('chat-view', '/primary', { event: console.log });</pre>
+window.airlock.subscribe('chat-view', '/primary', { event: console.log });</pre
```
liam-fitzgerald, line 54: Should be changed to `('graph-store', '/updates', { event: console.log })`, also I'm curious as to whether this actually works bc this might not be set correctly

liam-fitzgerald: EDIT: nvm, saw the duck punch

pkg/npm/http-api/example/index.html
```diff
@@ -1,98 +1,114 @@
...
-        });
-        window.airlock.subscribe('chat-view', '/primary', { event: console.log });
-        document.body.removeChild(document.getElementById('blastoff'))
+      window.airlock = await UrbitHttpApi.Urbit.authenticate({
+        ship: 'zod',
+        url: 'localhost:8080',
+        code: 'lidlut-tabwed-pillex-ridrup',
+        verbose: true,
+      });
+      window.airlock.subscribe('chat-view', '/primary', { event: console.log });
```
liam-fitzgerald, line 110: same deal

liam-fitzgerald (commented):
few nits, LGTM otherwise!

arthyn:
So trying to get this to build outside of our repository showed quite a few issues. Lerna seems to be linking the entire folders and not really respecting the "files" field in package.json, which means we were actually relying on importing the direct TS files not the output of the package.

finned-palmer:
This is great, thanks so much for your work on this @arthyn. This will open up a lot of opportunities for future landscape forks or other JS apps/sites for urbit.

liam-fitzgerald (approved):

PR #5412, opened 2021-11-09
dojo: improve safety of mark conversion

joemfb:
Since ford fusion, `:dojo` crashes on failed mark conversion or bunting a non-existent mark. Such crashes unlink dojo from drum, which discards the session and its history. This PR appropriately protects and virtualizes those operations.

Also, has dojo's mark conversion always been a mold invocation? That seems unnecessarily expensive -- it'd be nice to just cast if the value already nests.


ohAitch:
I'm guessing it goes through normal ++grab, and the default mark of a dojo result is %noun - with multiple stacked mark casts doing actual conversion, `&mime &txt /line1/line2` or w/e. Makes sense to special-case though.

philipcmonk (approved):
LGTM, and I think @ohAitch is correct

PR #5415, opened 2021-11-11
Minimal kelvin bump

philipcmonk:
This is what is currently deployed on ~latter-fordev-dozzod-dozzod and is a release candidate for the kelvin upgrade OTA.

Each of these changes fixes another case of upgrade.  Tested:

- Booting a new ship with the current pill (unfederated distribution) and receiving this update
- Receiving this update on a ship created to simulate federated distribution, ie receiving %base first and then the apps.
- Same, but running `|bump, =force &, =except (sy all-correctly-updated-desks ~)`
- A ship with 3rd party apps that are and are not ready for the upgrade (used %pals and %urchatfm)

The biggest thing we're missing is a button in landscape to run `|bump, =force &, =except (sy all-correctly-updated-desks ~)`.  Without that, anyone who's installed apps which haven't upgraded will be stuck until the app upgrades (then the OTA will automatically install) or until they run that from the cli.  Arguably we should fix `|bump` so it doesn't need the `=except` argument too.  Unfortunately both of those would need to go out before this update to be able to use them.


Fang- (approved):
This looks like the changes we paired on and tested, yes.

philipcmonk:
As of the last release, all known issues with the kelvin bump are fixed.

PR #5418, opened 2021-11-11
Add -G to Available Argument List

rabsef-bicrym:
-G is the way I primarily boot my ships and yet that flag has not been in the usage notes for some time (ever?). Adding it should assist new pilots in getting running and avoid questions of "I booted my ship but it took an hour and now I'm a comet" which sometimes (used to, at least) come w/ new pilots attempting to `-K kEy-S741nG`.


pkg/urbit/daemon/main.c
```diff
@@ -446,6 +446,7 @@ u3_ve_usage(c3_i argc, c3_c** argv)
     "-d            Daemon mode; implies -t\n",
     "-e url        Ethereum gateway\n",
     "-F ship       Fake keys; also disables networking\n",
+    "-G            Boot using key-string from key file\n",
```
Fang-, line 449: Sweet! "key-string from key file" is a bit confusing though. Do we pass the string or the file?

Might be more appropriately worded as "Private key string", to mirror the "Private key file" from `-k`. Also want to make clear you pass something in. ie:

```suggestion
    "-G 0wEtc      Private key string\n",
```

Arguably this should specify that the string is a `@uw`-encoded jamfile, but probably not worth the complexity in this context.

rabsef-bicrym: Let me know if the changes I just committed make sense to address the concerns you have above. Very good points, all around.

Fang- (approved):
Looks good to me!

PR #5419, opened 2021-11-12
vere: update event injection for four letter vanes

liam-fitzgerald:
Event injection validation was never updated to support routing on a
full vane name. Upgrades the check, so that wires with a first element
longer than 4 characters fail validation, instead of the previous limit
of 1 character


joemfb (approved):
LGTM

PR #5424, opened 2021-11-13
kiln: Fix for kelvin updates

philipcmonk:
This includes several fixes to kiln's handling of kelvin version updates.  These need to be released *before* the actual kelvin change.

For releasing the %garden changes, we need to downgrade ~mister-dister-dozzod-dozzod's %base +zuse and sys.kelvin, then commit these changes to %garden.


liam-fitzgerald (approved):
Kiln changes look very nice, feels a lot clearer now.

PR #5427, opened 2021-11-15
meta: enforce LF line endings for hoon files

ynx0:
This is a simple change which forces all hoon files to be checked in and checked out using LF newlines, regardless of the user's `core.autocrlf` setting. Without this, developers on Windows can inadvertently try to create a fakezod with illegal line endings, all while receiving unrelated error messages.


philipcmonk (approved):
LGTM

PR #5428, opened 2021-11-15
vere: support long-form command line arguments

Fang-:
Uses `getopts_long` to support long-form versions of every existing
command line argument, and adds them all to the help text, adding some
missing entries along the way.

This frees us from the confines of the alphabet, giving us the breathing room we need to move forward confidently.

I just off-the-cuff'd most of these, feel free to suggest alternatives etc.


joemfb (approved):
Thanks for doing this, @Fang-!

In the long run, I'm not sure that we want to enforce the variant that all short options have long options. There's more than a couple short options that are just gestures towards features that need to be (re-)implemented. And I'd like to make a pass over some of these names. But I haven't been able to get to those things, and this is much better than what we currently have.

LGTM

PR #5434, opened 2021-11-17
azimuth: update state on-watch

yosoyubik:
We were not updating the azimuth state (`nas`, `own` and `spo`) after running the logs from the snapshot.

Doing `:azimuth %rerun` gets us up to date, so this could be run on-load after merging, but since this is an expensive operation it might be fine leaving it as is?—people who want to run rollers or interested in inspecting azimuth data can be informed about the %rerun step, prior to `|rein %base [& %roller] [& %roller-rpc]`)

Update: i booted a moon from a pill that had this change and looks like azimuth has the correct state


philipcmonk:
The idea is that we update the state when poked with `%watch`, instead of in +on-watch.  Is that not working?

philipcmonk:
ah, that only happens on boot, not on upgrade

philipcmonk:
Then maybe we should remove running the logs in `%watch`.  The rule will be then that logs get (re)calculated in +on-watch and that's it.

(sorry about %watch vs +on-watch, back then subscriptions were handled in +peer)

philipcmonk:
And I guess we should reload in +on-load to make sure everyone on the network has the right state too

PR #5435, opened 2021-11-17
azimuth: stop processing state in %watch

philipcmonk:
Also reprocess state in +on-load for ships which currently have unprocessed state.


yosoyubik (approved):
LGTM

PR #5436, opened 2021-11-18
claz: better linear lockup ops support

Fang-:
I keep making these changes locally, so probably good to just commit them instead. 0a5bcf7 is pretty ugly, but gets the job done. We're not doing spawn-based lockups anymore anyway.

Arguably this should go into its own `pkg/` folder too, but the easier access is nice for now.


philipcmonk (approved):
Yeah eventually this should be split out, but these changes look good

PR #5437, opened 2021-11-18
u3: backup checkpoint on save (if no backup exists)

joemfb:
This PR makes vere automatically backup it's checkpoint on save, if no backup already exists. If any error occurs while attempting to create a backup, the `bhk` directory and its contents are entirely removed, so a subsequent save will reattempt to create a backup. For now, there is no automatic backup restoration; this functionality is being added to help prevent the most onerous event log replay situations.

### DIFF
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -768,6 +768,99 @@ _ce_image_fine(u3e_image* img_u,
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
LGTM

joemfb:
I've tested this on mac, linux, and windows; everything is working as expected.

PR #5439, opened 2021-11-19
meta: fix various issues in file tree for development

ynx0:
- remove outdated `sys.kelvin` from base-dev/
- remove outdated `lib/azimuth.hoon` from base-dev
- properly move updated `lib/azimuth.hoon` into base-dev/, then symlink back into arvo/


philipcmonk (approved):
These look good

PR #5445, opened 2021-11-23
jets: cache hmac hashing gate

Fang-:
As suggested by @joemfb. Probably not _that_ big a deal here, but figured it's worth doing anyway.


joemfb (approved):
LGTM, and sorry for the delay.

Please merge this into `next/vere`

PR #5448, opened 2021-11-23
nix: bump secp256k1 to latest version

mrdomino:
Needed for Schnorr signatures.


mrdomino:
This turned out to be more involved than I expected; closing.

PR #5449, opened 2021-11-23
groups: refactor joining process

liam-fitzgerald:
- Refactors the flow of group joining 
- Now handles breaches correctly
- Allows requests to be properly aborted
- Cleans up error handling
- Adds auto joining all graphs functionality
- Adds automatic sharing of contacts
- Updates UI to match design spec


pkg/interface/src/views/apps/launch/App.tsx
```diff
@@ -28,73 +27,85 @@ interface LaunchAppProps {
...
+                icon="Plus"
+                bg="washedGray"
+                color="black"
+                text="New Group"
+                style={{ gridColumnStart: 1 }}
+              >
+                <NewGroup />
+              </ModalButton>
+              <Button
+                bg="washedGray"
```
arthyn, line 94:
```suggestion
                backgroundColor="washedGray"
```

This button was invisible. I think the `bg` prop is special to `ModalButton` maybe?

arthyn (changes requested):
Frontend changes look good, a few small things to consider

pkg/landscape/app/metadata-pull-hook.hoon
```diff
@@ -322,6 +325,7 @@
   =/  =associations:metadata
     (metadata-for-group:met resource)
   :_  this
+  :-  `card`[%give %fact ~[/nack] resource+!>(resource)]
```
Fang-, line 328: Do we ever not give the resource on `/nack` here? Seems like this could just be in the /lib/pull-hook.

pkg/landscape/app/group-view.hoon
```diff
@@ -205,14 +232,12 @@
...
-        [ship %group-push-hook]
-      group-update-0+!>([%add-members rid (silt our.bowl ~)])
+    =.  jn-core  (emit add-us:pass)
     =.  jn-core  (tx-progress %start)
-    =>  watch-md
-    watch-groups
+    =>  (emit watch-md:pass)
+    =>  (emit watch-groups:pass)
+    =>  (emit watch-grp-nacks:pass)
+    (emit watch-md-nacks:pass)
```
Fang-: Consider `=~`. (Alternatively, `+emil`.)

pkg/landscape/app/group-view.hoon
```diff
@@ -239,6 +244,78 @@
...
+        =/  =metadatum:metadata  (need (peek-metadatum:met %groups rid))
+        :*  ~[(title title.metadatum 'succeeded')]
+            ~
+            now.bowl
+            /
+            /groups/(scot %p entity.rid)/[name.rid]
+        ==
+      ::
+          ?(%strange %retry)
+        :*  ~[(title name 'errored unexpectedly')]
```
Fang-: Should probably mention it will (or can?) retry?

pkg/landscape/sur/group-view.hoon
```diff
@@ -1,34 +1,66 @@
...
 |%
++$  app  ?(%graph %groups)
++$  uid  @uvH
+::
+::  $request: State of a join request
+::
+::    .started: Time request first sent
+::    .ship: Host of group
+::    .progress: Progress of request
+::    .share-co: Automatically share contact?
```
Fang-, line 12: Gotta say I continue not being a fan of abbreviations. the `%join` action gets the full thing, this probably should too.  
(Of course, the full thing feels large and unwieldy, but we have no cute metaphor to fall back on.)

Fang-: Also might be nice to match the ordering of items between this and `%join`.

pkg/landscape/app/group-view.hoon
```diff
@@ -135,7 +146,28 @@
...
-  state(joining (~(put by joining) rid request(hidden %.y)))
+  state(joining (~(put by joining) rid request))
+::
+++  is-tracking
+  |=  her=ship
+  ^-  ?
+  %+  lien  ~(tap in ~(key by joining))
+  |=([him=ship name=term] =(her him))
+::
+++  breach
```
Fang-, line 139: Feels kind of strange to me to have all this logic in group-view. The *-view agents have always held the role in my mind of thin in/output aggregation layer tailored towards frontends. Breach handling, request tracking etc, don't feel like they fit that very well, might be more at home in the pull-hook. But then, maybe all these things are too entangled for that to be feasible?

pkg/landscape/app/group-view.hoon
```diff
@@ -436,13 +443,14 @@
...
       |^  ^+  jn-core
+      =/  =request:view  (~(got by joining) rid)
       =/  feed  feed-rid
       =.  jn-core  (cleanup %done)
       =/  hidden  hidden:(need (scry-group:grp rid))
       =?  jn-core  ?&(!hidden ?=(^ feed))
         %-  emit
         (pull-gra:pass (need feed))
-      =?  jn-core  !hidden
+      =?  jn-core  |(hidden autojoin.request)
```
Fang-, line 434: Previously, we would pull if it wasn't hidden. Now, we pull when it's hidden or we want to autojoin. Feels like this should be `&(!hidden autojoin.request)` maybe?

Fang- (commented):
Biggest question for me here is why this logic goes into group-view rather than the pull-hook. Presumably for aggregation/entanglement/convenience reasons?

pkg/landscape/app/group-view.hoon
liam-fitzgerald: > more at home in the pull-hook
Right, but the motivating use case for tracking breaches is because poke handling is funny across a breach, and if you send a poke to an old life because you aren't up-to-date on chain state yet, that poke will never get (n)acked, permanently halting the group join flow. More general breach tracking can wait until there's better tooling for this (+on-rift).
> thin IO aggregation layer
This actually isn't how it works in practice because any long term UI state that is per-ship, must live in -view agents, under the store/hook/view agent model. in my mind, the hooks are the most principled part of the trichotomy and should serve as shims until subscription reform becomes real. When/if subscription reform/shrub/??? lands I'd like to seriously look at making the landscape codebase more monolithic

pkg/landscape/sur/group-view.hoon
liam-fitzgerald: good points!

pkg/landscape/app/metadata-pull-hook.hoon
liam-fitzgerald: good point!

pkg/landscape/app/group-view.hoon
liam-fitzgerald: Hm? Either it's hidden (i.e is an unmanaged group) or the user has manually selected that they wish to join all graphs

Fang-: Ok, but this is clearly different behavior from the previous version, where we only emitted these cards if it was _not_ hidden. Was that just a faulty implementation? Just making sure that checks out with the indented behavior here (which I admittedly do not fully understand, in either the old or new context).

Fang- (approved):
Changes look good! Aside from my confusion in that one thread above, this should be good to go into live qa afaict.

arthyn (approved):

PR #5451, opened 2021-11-23
king: bump zuse version to 419

### LINKED ISSUE #5450: King Haskell still has 420 as its vere version, blocking upgrade
Even in v1.17 package, the urbit-king binary has vere version 420 hardcoded instead of the newer 419, and this is blocking the required network upgrade. Here:

https://github.com/urbit/urbit/blob/1e70a1be2a03a1ba60b7d06ce50f12305be62eb0/pkg/hs/urbit-king/lib/Urbit/Vere/Pier.hs#L441-L442

@pilfer-pandex

pilfer-pandex:
Fixes #5450; thanks for the report @eglaysher.

@joemfb, @philipcmonk, what's the best way to make sure this doesn't recur?


joemfb:
This was already fixed on `next/vere`, which was just released as `v1.8`.

PR #5454, opened 2021-11-25
webterm: backport

Fang-:
Backports the webterm changes from #4463.

Draft because this doesn't quite compile, and is untested.

@liam-fitzgerald I can run `npm run start` from `pkg/interface` just fine, but it doesn't pick up on webterm. Running it from `pkg/interface/webterm` complains about not being able to resolve imports, particularly the `~/etc`-style imports in non-webterm files.  
Whenever you have time, would love to get some advice and/or pair on setting this up right again.


Fang-:
Thanks for your help @liam-fitzgerald! Small styling tweaks aside, this should be ready for review now.

I could test this just fine against a fakezod. Connecting it to my live ship made eyre complain about the format of the json sent into the channel. "It's probably nothing", but, curious to hear if you can replicate or not.

liam-fitzgerald (changes requested):
Changes LGTM in general, however:

looks like this is blocked on deploying a fix to `+stap`:
On a fakezod:
```
> `@t`(scan "/test/" stap)
/test/
```
On my livenet ship:
```
> `@t`(scan "/test/" stap)
syntax error
```

Looks like it doesn't like empty path segments, is there any way to work around this?

Fang-:
Oh, that's actually #5227, which our livenet ships have but our fakezods somehow don't. It makes `/test/` invalid, because trailing `/`.

Herm hasn't been updated to account for this, it expects subscription paths of the shape `/session/[id]`, where the id for the default session is `%$` aka `~` aka `/session/`.

I suppose there's some case to be made that, if we accept empty segments in the middle of a path, we should accept them at the tail of a path as well. (Not to mention being able to `` `path`[%a %$ ~]``...)

Whatever the conclusion there though, it seems out of scope for this. Considering herm currently produces facts only on the `[%session %$ ~]` path, we'll need to ship a herm (base desk) change alongside this.  
I'll commit something in a bit...

Fang-:
Since /app/herm is distributed with `%base`, we'll want to just release both it and `%webterm` simultaneously. I don't think we need to be too concerned about the synchronicity of these, considering webterm's current state is broken anyway.

Fang-:
(fwiw tested on my live ship and Works Fine™)

urcades (approved):
Here we go

PR #5455, opened 2021-11-25
vere: alphabetize getopt short options

mrdomino:
Tested by building vere and trying it with `-B`, `-p`, and `-F`.


Fang- (approved):
Cute Enough™

PR #5457, opened 2021-11-28
improve de-xml parser

h5gq3:
de-xml parser fails when xml content node contains doublequotes (`doq` rule), this PR proposes to remove this restriction as high-level javascript APIs that operate on DOM don't entitize/encode doublequotes by default.


Fang- (approved):
Kind of weird that it was being excluded in `+chrd` rather than `+escp` in the first place, the latter being where the exclusion for `<`, `>` and `&` happens.

I don't think loosening the parser in this way will give us any trouble. Approved!

In the future, please respect our commit message [guidelines](https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md#hygiene).

PR #5459, opened 2021-11-28
roller-cli: add cli to interact with azimuth rollers

yosoyubik:
Marked as draft for now, functionality is pretty much the same as this [demo](https://www.youtube.com/watch?v=iEUxoBuTvqY) but the CLI gets, in addition to transaction status, point updates (for example if the nonce is incremented, we receive an update with the new one, so it can be retrieved locally for signing transactions).

I'd be interested to hear if this is something we could add together with the roller, or can be extracted out at a later time into a separate indexer/roller desk.

It also lacks a nice way of choosing the type of L2 transaction that you want to do. I can imagine a sort of dialog that retrieved users' data and uses it based on the entered actions (e.g. "Do you want to escape?" "Enter your new parent...")

### DIFF
--- a/pkg/arvo/app/azimuth.hoon
+++ b/pkg/arvo/app/azimuth.hoon
@@ -130,6 +130,7 @@
         ~&  [%rerunning (lent logs.state)]
         =.  points.nas.state  ~
         =.  own.state  ~
+        =.  spo.state  ~
         =^  *  state  (run-logs:do logs.state)
         `this
       ::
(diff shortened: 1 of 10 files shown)


zalberico:
Leaving this open for now since it's a WIP and in a decent state (and @yosoyubik is out still), but if we don't have a clear path/goal for merging this when @yosoyubik returns we should probably close as not planned for now.

zalberico:
On second thought - in the interest of trying to bias to cleaning up stale state, I'll close this and we can reopen if it's something we decide to move forward with at that time.

PR #5461, opened 2021-11-30
notify: upgrade for new hark-store, miscellaneous improvements

liam-fitzgerald:
- %notify now leaks 0 (zero) private information over APNS
- Sends a silent push notification whenever either a) A new notification ought to be generated or b) An existing notification should be dismissed
- Exposes scry endpoint for downloading the contents of a push notification
- Infrastructure is now far more generic, could be used for any notifications from hark-store
- Moves `/+ hark-store` into garden-dev

cc: @ixv no pressure to review, just thought you might want to


pkg/landscape/app/notify.hoon
```diff
@@ -63,18 +71,35 @@
...
-            (~(watch-our pass:io /hark) %hark-store /notes)
-        ==
+        %2
+      =/  upd=wire  /hark/updates
+      =/  not=wire  /hark/notes
+      =/  =dock  [our.bowl %hark-store]
+      =?  cards  !(~(has by wex.bowl) [upd dock])  :: rewatch notes
+        :_(cards [%pass upd %agent dock %watch /updates])
+      =?  cards  !(~(has by wex.bowl) [not dock])  ::  rewatch updates
+        :_(cards [%pass not %agent dock %watch /notes])
```
Fang-, line 82: I think the comments here are the wrong way around.

pkg/landscape/app/notify.hoon
```diff
@@ -22,21 +22,29 @@
...
 ::
 +$  state-0
   [%0 base-state-0]
 ::
 +$  state-1
   [%1 base-state-0]
 ::
++$  state-2
+  [%2 base-state-1]
+::
```
Fang-, line 38: I think it's more common to just bump directly to `base-state-2`, as in "the `base-state` that was introduced during state version `%2`. Seems better than getting versioning for parts of the state completely out of sync.

pkg/landscape/app/notify.hoon
```diff
@@ -63,18 +71,35 @@
...
+::        %0
+::      %_    $
+::          -.old  %1
+      ::
+::          cards
+::        %+  welp  cards
+::        :~  (~(leave-our pass:io /hark) %hark-store)
+::            (~(watch-our pass:io /hark) %hark-store /notes)
+::        ==
+::      ==
```
Fang-, line 102: This can just be removed... right?

pkg/landscape/app/notify.hoon
```diff
@@ -289,10 +319,12 @@
...
     ::
-        [%remove-binding *]
-      `this
+        [%remove-binding *]  `this
     ::
         [%send-notification *]
+      ?>  ?=(%iris -.sign-arvo)
+      ?>  ?=(%http-response +<.sign-arvo)
+      ?>  ?=(%finished -.client-response.sign-arvo)
       `this
```
Fang-, line 328: Do we not care about handling non-200 responses here? (Or `%cancel`, though iirc those never happen in practice rn. Might still be nice to prep for them existing though.)

Fang- (commented):
I think the logic here checks out. Some small cleanliness nits and a question about error handling below.

pkg/landscape/app/notify.hoon
liam-fitzgerald: This didn't actually cause non-200 response to break in my testing, but idk, happy to change

Fang-: Well they won't break, because they will come back as `%finished`, but you might want to recognize, say, a `500`, or some legible api-level error message, and retry submitting the notification, or things like that?

liam-fitzgerald: Given the way that notifications works, I'm not sure retrying makes sense, but yeah, we should probably at least log this

Fang- (approved):
Changes look good, thanks!

PR #5462, opened 2021-11-30
invite-hook: use equality instead of team:title

liam-fitzgerald:
Fixes a bug that prevented moons from being invited


Fang- (approved):
Moon liberation squad rolling out huh. Changes here look good!

ftr, `=(our src):bowl` would also work fine.

PR #5464, opened 2021-12-01
eyre: include mark of fact in channel json

liam-fitzgerald:
This change greatly improves the ergonomics of working with channel JSON
in statically typed languages, as the polymorphism is moved out of the
actual diff and into the event framing.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1705,12 +1705,16 @@
         ==
       ::
           %fact
-        :~  ['response' [%s 'diff']]
+        :+  ['response' [%s 'diff']]
           ::
-            :-  'json'
-            ~|  [%unexpected-fact-mark p.cage.sign]
-            ?>  =(%json p.cage.sign)
-            !<(json q.cage.sign)
+          :-  'json'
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1705,12 +1705,16 @@
...
-            :-  'json'
-            ~|  [%unexpected-fact-mark p.cage.sign]
-            ?>  =(%json p.cage.sign)
-            !<(json q.cage.sign)
+          :-  'json'
+          ~|  [%unexpected-fact-mark p.cage.sign]
+          ?>  =(%json p.cage.sign)
+          !<(json q.cage.sign)
+        ::
+        ?~  from  ~
```
Fang-, line 1715: You're fine to remove the `::` now imo.

Fang- (commented):
I can understand adding the mark, seems like a good & fair change. We'd probably do something similar for jam-channels.

Not sure about the desk though. Don't you implicitly know it based on the agent your subscription is pointing at? Even if you don't, doesn't the agent case have the same problem then, making the desk _here_ kind of "too little too late"?

A little bit surprised Eyre tests are passing, but I guess they hit the `?=(~ from)` case here.

Fang- (approved):
Looks good to me!

PR #5467, opened 2021-12-03
azimuth: load azimuth data from state snapshot

yosoyubik:
Previously Azimuth data was initialized into `/app/azimuth` by loading a snapshot with a list of ethereum logs, and applying `/lib/naive.hoon` to them, recreating the Azimuth state in return.

This PR gets the data from a state snapshot (`/app/azimuth/state/naive`) of shape: `[%0 =id:block:jael nas=^state:naive =owners =sponsors]`—`id` points to the `id:block:jael` from the latest ethereum log at which the state was produced.

Calling [+run-state](https://github.com/urbit/urbit//commit/ad16155c77b3f1ddc325cfa7b9168bae29f6ef62#diff-3c22b74a1f92a7f9793890783845c9c517d571f8b91d2541774a8a0167d9c832R299) for sending `%jael` udiffs, from `77.319 points` required around [35MB](https://gist.github.com/yosoyubik/b1c18ede4ef5882fb6a1e2e267a82411) of memory—[here](https://github.com/urbit/urbit/pull/5432#issue-1055681814) there's a comparison with different versions of running the naive function over a list of logs.

The PR also includes two threads (`-azimuth-snap-state` and `-azimuth-snap-logs`) that create the snapshot from the current `/app/azimuth` state or from a list of ethereum logs.

Because we are sending udiffs with points that have >1 lives and rifts, booting a new ship prints out a lot of these: `%udiff-to-diff-skipped-rift` from [here](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/lull.hoon#L1998)

Maybe it could be posible to add a new udiff (`%boot`) to account for this case?:

```
+$  udiff
  $:  =id:block
  $%  [%boot =point]
      [%rift =rift]
      [%keys key-update]
      [%spon sponsor=(unit @p)]
      [%disavow ~]
  ==  ==
```

PS: `/app/azimuth/logs/eth-logs` has not been removed in this PR


yosoyubik:
I tested this  (adding `boot=?` to both rift/keys udiffs), instead of the changed I proposed earlier, and then checking for it when showing the printf:

```
      $:  =id:block
      $%  [%rift =rift boot=?]
          [%keys key-update boot=?]
          [%spon sponsor=(unit @p)]
          [%disavow ~]
      ==  ==
```

Booting a comet from a pill with the changes showed that the memory for the serf was around 1.5GB during the whole booting process, going up to 1.8GB; and 890MB for the king.

philipcmonk:
Do you have a sense of how long loading the snapshot takes now?

I also think we should make sure the snapshot is put in the pill as a %mime if we can, which should reduce the king's memory usage by quit a bit more, since it won't need to cue a 77k-long list.  mars/urth will also fix this, but it's not clear when that will land.  Do you know how to change that in lib/pill?

yosoyubik:
> Do you have a sense of how long loading the snapshot takes now?

I'll get some measurements

> Do you know how to change that in lib/pill?

Yeah I think so (just by quickly looking at it now)—will ping you if something comes up.

yosoyubik:
> Do you have a sense of how long loading the snapshot takes now?

Just did some manual measurements and it looks like it takes between 15-30 s. (less the firs time, and a bit more the second time +on-watch is called) so around ~s45 in total. for turning over all 77K points.

I just [added the snapshot to the pill](https://github.com/urbit/urbit/pull/5467/commits/81eb6ae247aec77be793ad135b45812cdf261fdf) and tested it booting a comet with it, and all looks good.

philipcmonk:
These changes look good and should make bootup on the live net significantly better.  This should target next/arvo.

CI is broken on macos for unrelated nix reasons, but the ubuntu breakage is real.  Looks like `|moon-breach` was using the old udiff type.

CI takes forever to run of course, so once it recognizes a test failure I always run the compilation and unit tests locally with:
```
-test %/tests ~
|start %test
:test %agents
:test %generators
:test %marks
```

First line runs the unit tests, the others makes sure all our agents/generators/marks compile.

> PS: /app/azimuth prints out some logs that break the 4-letter word symmetry:

Not thinking of a good 4-letter name, maybe `id:` with three spaces?

yosoyubik:
Done! all tests (not counting mac) pass now—thanks for the snippets to run them locally!

yosoyubik:
@philipcmonk Just pushed [the update](https://github.com/urbit/urbit/pull/5467/commits/fb2c078fde65c583df4f6ca43fe20597b5cdcc8d) for including the snapshot correctly in the pill as mime. Booting from it takes less memory that before (around 1.5-1.6GB/700MB)

PS: I tested several versions of how it'd look to print out "id:   XX", "  id: X", but went with "ship: loading XX"

yosoyubik:
There seems to be an issue with creating a pill from this branch, that causes a ship to not be able to boot from it—here's the error in [CI](https://github.com/urbit/urbit/runs/4483265622?check_suite_focus=true#step:5:25)

(If we boot with a directory for initial clay sync, it works)

It seems like it's failing in the `%0` tag here: `+$  snap-state  [%0 =id:block:jael nas=^state:naive =owners =sponsors]` but [this is cast](https://github.com/urbit/urbit/blob/8601b13e3fe35e138089032e1f390ead85f32244/pkg/arvo/ted/azimuth/snap-logs.hoon#L64-L67) when the snapshot is created...

yosoyubik:
Still trying to figure out what's wrong with [this](https://github.com/urbit/urbit/pull/5467#issuecomment-991105800).

The story goes like this: 

- [This thread](https://github.com/urbit/urbit/blob/8601b13e3fe35e138089032e1f390ead85f32244/pkg/arvo/ted/azimuth/snap-state.hoon) scries for the relevant data in /app/azimuth and writes it to a file (e.g. `version-0.azimuth-snapshot`) using [this mark](https://github.com/urbit/urbit/blob/8601b13e3fe35e138089032e1f390ead85f32244/pkg/arvo/mar/azimuth/snapshot.hoon)

- The urbit noun (`snap-state`) that represents the data is [declared here](https://github.com/urbit/urbit/blob/8601b13e3fe35e138089032e1f390ead85f32244/pkg/arvo/sur/dice.hoon) and is of shape: `[%0 =id:block:jael nas=^state:naive =owners =sponsors]`

- The mark has `+mime` arms to convert to and from it (using `jam`/`cue` internally)

- When creating a solid pill, if the file is a `azimuth-snapshot`, we use a [mark conversion gate](https://github.com/urbit/urbit/blob/8601b13e3fe35e138089032e1f390ead85f32244/pkg/base-dev/lib/pill.hoon#L89-L91) from `snap-state` to `mime`, and store it as such.

- Now, booting from a solid pill produces [this error](https://gist.github.com/yosoyubik/9384024780d7f4fdb00194cccee70c6f):

```
tone: 2
tax: [[%sur %dice %hoon 0] [11 18] 11 20]
tax: [[%sur %dice %hoon 0] [11 17] 11 71]
mot: %mean
```

- Uncommenting all slogs in %clay, this comes from [here](https://github.com/urbit/urbit/blob/81eb6ae247aec77be793ad135b45812cdf261fdf/pkg/arvo/sys/vane/clay.hoon#L766)—as a validation check, I tested manually creating a dais and calling it in the thread that produces the snapshot file, but that works:


```
;<  =dais:clay  bind:m  (build-mark:strandio -.u.bem %azimuth-snapshot)
::
=+  (vale:dais [%0 id nas owners sponsors])
```

- Some other tests to check that the mark works as expected:

```
> =n -build-file %/lib/naive/hoon
> =s -build-file %/sur/dice/hoon
> =snap [%0 *id:block:jael *^state:n *owners:s *sponsors:s]
ford: make cast %noun -> %azimuth-snapshot
ford: make tube %noun -> %azimuth-snapshot
> &azimuth-snapshot &mime &azimuth-snapshot snap
[ %0
  id=[hash=0x0 number=0]
  nas=[%0 points={} operators={} dns=<||>]
  owners={}
  sponsors={}
]
```

- Booting from a clean solid pill (from master) and then using -A to a version of pkg/arvo that has all files up to date (snapshot included) works fine. But then, mounting the files and creating a pill, that's used to boot a ship from fails with the error in `/sur/dice`.

yosoyubik:
Happy to report that as of [this commit](https://github.com/urbit/urbit/pull/5467/commits/4724c32d3dbeb37a69f5c17657bfae3562029c4f) botting from the solid pill included in this branch works

philipcmonk (approved):

PR #5469, opened 2021-12-03
interface: minor group join fixes

liam-fitzgerald:
- Fixes graph redirects of unjoined groups
- Stops the group summary from reflowing the modal when loaded
- Adds invite declination
- Fixes clicking on an invite notification
- Fixes group navigation


pkg/interface/src/views/landscape/components/Join/Skeleton.tsx
```diff
@@ -18,6 +18,9 @@ import { resourceFromPath } from "~/logic/lib/group";
 import useMetadataState, { usePreview } from "~/logic/state/metadata";
 import useInviteState, { useInviteForResource } from "~/logic/state/invite";
 
+const SUMMARY_HEIGHT = "96px";
+const SUMMARY_WIDTH = ["90vw", "512px"];
```
arthyn, line 22: This doesn't seem to be used
```suggestion
```

arthyn (approved):
Looks good, I was wondering where that decline was but I forgot to ask haha

PR #5470, opened 2021-12-03
boot: bring your own boot sequence

Fang-:
Technically more like, "add your own boot sequence events".

This implements a set of boot arguments starting with `--prop-`, used for loading in "props", jamfiles that contain additional events for inclusion in the boot sequence. (The dream is to make this just be a single `--prop` or w/e that auto-distinguishes between filepaths, urls and names, but we're grounded in reality for now.)

Intended use cases here include specifying desks to install during boot, bootstrapping Azimuth state, custom userspace configuration, etc. Being able to do these separate from the pill brings some much-needed flexibility and lets us eradicate "multi-pills" from existence.

Main thing to look at is the definition of `$prop` in base-dev's `/lib/pill` and the way those nouns get handled in `mars.c`.

Based on and targeting `jb/wip-urth` for @joemfb's convenience.  
Includes #5428 because we want long opts for this.  
Includes #5389 because that should be updated to make use of the affordances this provides. ~~Draft mode until `/app/autopill` improvements make their way in, but the vere part here should be ready for its first review.~~

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -398,7 +398,7 @@ _main_getopt(c3_i argc, c3_c** argv)
            && u3_Host.ops_u.url_c == 0
            && u3_Host.ops_u.git == c3n ) {
     u3_Host.ops_u.url_c =
-      "https://bootstrap.urbit.org/props/brass.pill";
+      "https://bootstrap.urbit.org/props/" URBIT_VERSION "/brass.pill";
   }
   else if ( u3_Host.ops_u.nuu == c3y
            && u3_Host.ops_u.url_c == 0
--- a/pkg/urbit/vere/king.c
+++ b/pkg/urbit/vere/king.c
@@ -189,10 +189,10 @@ _king_prop()
       } break;
 
       case 3: {  //  name
-        //NOTE  this implementation limits us to max 20 char prop names
+        //NOTE  this implementation limits us to max 38 char prop names
         c3_c url_c[80];
         sprintf(url_c,
-                "https://bootstrap.urbit.org/props/%s.jam",
+                "https://bootstrap.urbit.org/props/" URBIT_VERSION "/%s.jam",
                 vex_u->loc_c);
         u3l_log("boot: downloading prop %s\n", url_c);
         u3_atom jam = _king_get_atom(url_c);


Fang-:
This now includes an updated `pkg/autoprop` for auto-generating pill and prop files. Thanks to both @joemfb and @philipcmonk for the clay pointers. Please take a look, this should be ready for full review now.

pkg/urbit/worker/mars.c
```diff
@@ -1004,6 +1004,9 @@ _mars_boot_make(u3_boot_opts* inp_u,
         if ( c3n == u3r_cell(dat, &who, 0) ) {
           return c3n;
         }
+        if ( c3n == u3r_cell(who, &who, 0) ) {
+          return c3n;
+        }
```
joemfb, line 1009: This seems redundant and wrong, did the format change somehow?

pkg/urbit/worker/mars.c
```diff
@@ -1063,6 +1066,40 @@ _mars_boot_make(u3_boot_opts* inp_u,
...
+            break;
+          }
+
+          if ( c3y == u3r_sing_c("pre-userspace", ter) ) {
+            u3m_p("prop: pre-userspace", met);
+            pre = u3kb_weld(pre, ves);
+          }
+          else if ( c3y == u3r_sing_c("post-userspace", ter) ) {
+            u3m_p("prop: post-userspace", met);
+            aft = u3kb_weld(aft, ves);
```
joemfb, line 1090: `u3k(ves)`

pkg/urbit/worker/mars.c
```diff
@@ -1063,6 +1066,40 @@ _mars_boot_make(u3_boot_opts* inp_u,
...
+            u3m_p("unrecognized prop tier", ter);
+          }
+        } break;
+
+        default: u3m_p("unrecognized boot sequence enhancement", u3h(mot));
+      }
+
+      mos = u3t(mos);
+    }
+
```
joemfb, line 1102: this should all be in its own block (starting with `u3_noun mos = mor;`), and it should end with `u3z(mor)` (which is being leaked).

pkg/base-dev/lib/pill.hoon
```diff
@@ -128,4 +142,225 @@
...
+    |.  ~+
+    =/  pro  q:$:tap
+    [[%cell p.gun p:$:tap] [.*(pro q.gun) pro]]
+  --
+::
+++  events
+  |%
+  +$  prop  [%prop meta tier (list ovum)]
+  +$  meta  term
+  +$  tier  ?(%pre-userspace %post-userspace)
```
joemfb, line 349: I'm tempted to say that these types should be in the kernel somewhere. They are being referenced in the ipc boot message specification. But I'm not sure where.

I don't like the aesthetic of `$tier`, I'd rather it be something like `?(%fore %hind)`. And I'm having trouble coming up with a use-case for other tiers ("pre-kernelspace"?). Thoughts?

joemfb (changes requested):
Sorry for the absurd delay @Fang-!

This all looks reasonable. I've picked some nits the C changes, and noted a couple refcounting issues. I've only minimally reviewed the hoon so far, mostly to highlight some questions about the new move/message types that have been introduced.

We'll need to go over which ovums are included in the pill by default, and which (if any) props are downloaded by default. I suspect will want a set of default URL props, and a commandline flag to turn them off.

pkg/urbit/vere/king.c
```diff
@@ -222,22 +224,65 @@ _king_dawn(u3_noun feed, u3_noun pill, u3_noun path)
...
+
+        case 3: {  //  name
+          c3_c* url_c = c3_malloc(80);
+          sprintf(url_c,
+                  "https://bootstrap.urbit.org/props/%s.jam",
+                  vex_u->loc_c);
+          u3_atom jam = _king_get_atom(url_c);
+          mor = u3nc(u3ke_cue(jam), mor);
+          c3_free(url_c);
+        }
```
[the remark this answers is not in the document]
Fang-: I added some dumb retry logic. Not sure what else you were imagining with better error handling here. Probably not worth following redirects or w/e?

pkg/arvo/sys/lull.hoon
```diff
@@ -773,6 +773,7 @@
         [%park des=desk yok=yoki ran=rang]              ::  synchronous commit
         [%perm des=desk pax=path rit=rite]              ::  change permissions
         [%pork ~]                                       ::  resume commit
+        [%prep lat=(map lobe blob)]                     ::  prime blob store
```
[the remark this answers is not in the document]
Fang-: Discussed this with @philipcmonk, just leaving it as a `map` for now.

pkg/urbit/daemon/main.c
```diff
@@ -389,7 +398,7 @@ _main_getopt(c3_i argc, c3_c** argv)
            && u3_Host.ops_u.url_c == 0
            && u3_Host.ops_u.git == c3n ) {
     u3_Host.ops_u.url_c =
-      "https://bootstrap.urbit.org/urbit-v" URBIT_VERSION ".pill";
+      "https://bootstrap.urbit.org/props/brass.pill";
```
Fang-, line 401: @joemfb This change probably won't pass muster. Not sure what the correct solution here is though.

I have deployed the version of `/app/autoprop` from this PR to a ~zod moon, running next to a little script that uploads the .pill and .jam files it generates to bootstrap.urbit.org, using static per-prop filenames.

We could have the autoprop build pills with the old versioned naming scheme, but then need to be very good about remembering to update the autoprop config whenever necessary. The point of this side of this work is to have automated nearly-latest pills... Arguably if you find yourself trying to use a pill you can't run, you shouldn't be using this old a binary anyway?

joemfb (approved):
There's a couple things here I may still want to tweak, especially the behavior around default props. But I'm going to merge this out of band (don't worry about the conflicts), so I'll address those elsewhere.

Thanks for pulling this together!

PR #5472, opened 2021-12-05
landscape: correct mark

mrdomino:
This was crashing before due to a `[%no-cast-from %graph-update-3 %graph-update]`. With this change it produces output.


liam-fitzgerald (approved):
Ugh, code rot. LGTM!

PR #5473, opened 2021-12-05
pkg: include atom mark in base-dev

Fang-:
Found out the hard way that the `%atom` mark was actually not included in base-dev. Proposing here that it should be in there, considering it's not too uncommon to produce data with that mark.


belisarius222 (approved):
lgtm

PR #5475, opened 2021-12-07
bitcoin-utils: Fix parsing of of CompactSize integers

dasrun-fadben:
When the first byte is greater or equal to 0xfd, (bex len) bytes are consumed to
form the csiz atom, but only one byte is dropped from the 'rest' of the input.

The parser should consume all bytes of the CompactSize.

### DIFF
--- a/pkg/base-dev/lib/bitcoin-utils.hoon
+++ b/pkg/base-dev/lib/bitcoin-utils.hoon
@@ -143,6 +143,7 @@
     |=  a=@
     ^-  hexb
     =/  l=@  (met 3 a)
+    ?:  =(l 0)  1^a
     ?:  =(l 1)  1^a
     ?:  =(l 2)  (cat:byt ~[1^0xfd (flip:byt 2^a)])
     ?:  (lte l 4)  (cat:byt ~[1^0xfe (flip:byt 4^a)])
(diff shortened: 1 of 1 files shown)


zalberico:
@belisarius222 assigning this one to you for next steps, either way we should not be merging it directly into master

cyclomancer:
@belisarius222 asked me to take a look at this and I think it's good to merge. The first commit covers the missing 0-value case for encoding, and the second fixes a straightforward bug in parsing that treated a `bloq`/exponent "length" value as the actual bytewidth.

Something else to fix in the future here (and elsewhere in the bitcoin libraries?) is the use of `bloq` as a bytewidth value rather than strictly bitwidth. That doesn't impact this PR's fix however.

belisarius222:
Thanks @cyclomancer.  Let's get this merged, then.  @zalberico should this go into next/arvo?  Please advise.

zalberico:
Yep - just needs approval first and be up to date with next/arvo prior to merge. Then if things look good and tests are passing you can merge it.

We're not deploying next/arvo to the canary moon on merge yet because Philip is testing his upgrade work there, but his upgrade work is moving to the tlon internal moon shortly at which point the tip of next/arvo will move to the canary.

belisarius222 (approved):
rubber stamp lgtm

PR #5477, opened 2021-12-07
u3: disable page tracking on inner roads

mcevoypeter:
Disable page tracking on all non-home roads. Two changes were required to accomplish this:

(1) In `u3m_leap`, write protections are removed from all inner road pages when jumping into an inner road from the home road. Because `mprotect` operates on native pages, not Urbit pages, the following macros have been added to more clearly express the rounding logic of the home road hat and cap:
- `U3_OS_NativePageBytes`: number of bytes in a virtual memory page on the given platform (i.e. `sysconf(_SC_PAGESIZE)`).
- `U3_OS_NativePageWords`: `U3_OS_NativePageBytes / 4`.
- `c3_rup`: round an arbitrary natural number up to the specified factor, which must be a power of 2.
- `c3_rud`: round an arbitrary natural number down to the specified factor, which must be a power of 2.

(2) In `u3m_fall`, write protections are reinstated on *all clean pages* (including clean home road pages) when returning to the home road. Note that the seemingly more efficient solution of issuing a single `mprotect` call to reinstate write protections on all inner road pages based on the current home road hat and cap is more nuanced than it appears, and so the obvious solution of handling each page in the loom individually is more appropriate here.

### DIFF
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -471,25 +471,6 @@ _ce_patch_save_page(u3_ce_patch* pat_u,
   return pgc_w;
 }
 
-/* _ce_patch_junk_page(): mark a page as junk.
-*/
-static void
-_ce_patch_junk_page(u3_ce_patch* pat_u,
-                    c3_w         pag_w)
-{
-  c3_w blk_w = (pag_w >> 5);
-  c3_w bit_w = (pag_w & 31);
-
-  // u3l_log("protect b: page %d\r\n", pag_w);
-  if ( -1 == mprotect(u3_Loom + (pag_w << u3a_page),
-                      (1 << (u3a_page + 2)),
-                      PROT_READ) )
-  {
-    c3_assert(0);
-  }
-  u3P.dit_w[blk_w] &= ~(1 << bit_w);
-}
-
 /* _ce_patch_compose(): make and write current patch.
 */
 static u3_ce_patch*
@@ -545,9 +526,6 @@ _ce_patch_compose(void)
     }
     for ( i_w = 0; i_w < sou_w; i_w++ ) {
       pgc_w = _ce_patch_save_page(pat_u, (u3a_pages - (i_w + 1)), pgc_w);
-    }
-    for ( i_w = nor_w; i_w < (u3a_pages - sou_w); i_w++ ) {
-      _ce_patch_junk_page(pat_u, i_w);
     }
 
     pat_u->con_u->nor_w = nor_w;
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -490,43 +490,6 @@ _ce_patch_junk_page(u3_ce_patch* pat_u,
   u3P.dit_w[blk_w] &= ~(1 << bit_w);
 }
 
-/* u3e_dirty(): count dirty pages.
-*/
-c3_w
-u3e_dirty(void)
-{
-  c3_w pgs_w = 0;
-  c3_w nor_w = 0;
-  c3_w sou_w = 0;
-
-  /* Calculate number of saved pages, north and south.
-  */
-  {
-    c3_w nwr_w, swu_w;
-
-    u3m_water(&nwr_w, &swu_w);
-
-    nor_w = (nwr_w + ((1 << u3a_page) - 1)) >> u3a_page;
-    sou_w = (swu_w + ((1 << u3a_page) - 1)) >> u3a_page;
-  }
-  //  u3K.nor_w = nor_w;
-  //  u3K.sou_w = sou_w;
-
-  /* Count dirty pages.
-  */
-  {
-    c3_w i_w;
-
-    for ( i_w = 0; i_w < nor_w; i_w++ ) {
-      pgs_w = _ce_patch_count_page(i_w, pgs_w);
-    }
-    for ( i_w = 0; i_w < sou_w; i_w++ ) {
-      pgs_w = _ce_patch_count_page((u3a_pages - (i_w + 1)), pgs_w);
-    }
-  }
-  return pgs_w;
-}
-
 /* _ce_patch_compose(): make and write current patch.
 */
 static u3_ce_patch*
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -677,27 +677,25 @@ _ce_image_blit(u3e_image* img_u,
     return;
   }
 
-  c3_w siz_w = img_u->pgs_w * (1 << (u3a_page + 2));
+  c3_w i_w;
+  c3_w siz_w = 1 << (u3a_page + 2);
 
   lseek(img_u->fid_i, 0, SEEK_SET);
-  if ( -1 == read(img_u->fid_i, ptr_w, siz_w) ) {
-    fprintf(stderr, "loom: image blit read: %s\r\n", strerror(errno));
-    c3_assert(0);
-  }
-
-  {
-    c3_w i_w;
-    for ( i_w = 0; i_w < img_u->pgs_w; i_w++) {
-      c3_w pag_w = (u3a_outa(ptr_w) >> u3a_page) + i_w;
-      c3_w blk_w = pag_w >> 5;
-      c3_w bit_w = pag_w & 31;
-      u3P.dit_w[blk_w] &= ~(1 << bit_w);
+  for ( i_w = 0; i_w < img_u->pgs_w; i_w++ ) {
+    if ( -1 == read(img_u->fid_i, ptr_w, siz_w) ) {
+      fprintf(stderr, "loom: image blit read: %s\r\n", strerror(errno));
+      c3_assert(0);
     }
-  }
 
-  if ( 0 != mprotect(ptr_w, siz_w, PROT_READ) ) {
-    fprintf(stderr, "loom: live mprotect: %s\r\n", strerror(errno));
-    c3_assert(0);
+    if ( 0 != mprotect(ptr_w, siz_w, PROT_READ) ) {
+      fprintf(stderr, "loom: live mprotect: %s\r\n", strerror(errno));
+      c3_assert(0);
+    }
+
+    c3_w pag_w = u3a_outa(ptr_w) >> u3a_page;
+    c3_w blk_w = pag_w >> 5;
+    c3_w bit_w = pag_w & 31;
+    u3P.dit_w[blk_w] &= ~(1 << bit_w);
   }
 }
 
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -981,5 +981,5 @@ u3e_yolo(void)
 void
 u3e_foul(void)
 {
-  memset((void*)u3P.dit_w, 0xff, u3a_pages >> 3);
+  memset((void*)u3P.dit_w, 0xff, sizeof(u3P.dit_w));
 }
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -674,6 +674,8 @@ _ce_image_blit(u3e_image* img_u,
     c3_w blk_w = pag_w >> 5;
     c3_w bit_w = pag_w & 31;
     u3P.dit_w[blk_w] &= ~(1 << bit_w);
+
+    ptr_w += stp_ws;
   }
 }
 
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -123,13 +123,13 @@ u3e_fault(void* adr_v, c3_i ser_i)
   c3_w* adr_w = (c3_w*) adr_v;
 
   if ( (adr_w < u3_Loom) || (adr_w >= (u3_Loom + u3a_words)) ) {
-    fprintf(stderr, "address %p out of loom!\r\n", adr_v);
+    fprintf(stderr, "address %p out of loom!\r\n", adr_w);
     fprintf(stderr, "loom: [%p : %p)\r\n", u3_Loom, u3_Loom + u3a_words);
     c3_assert(0);
     return 0;
   }
   else {
-    c3_w off_w = (adr_w - u3_Loom);
+    c3_w off_w = u3a_outa(adr_w);
     c3_w pag_w = off_w >> u3a_page;
     c3_w blk_w = (pag_w >> 5);
     c3_w bit_w = (pag_w & 31);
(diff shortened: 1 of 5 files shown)


philipcmonk:
Very nice!  Does this make a noticeable performance difference on your system?

mcevoypeter:
> Very nice! Does this make a noticeable performance difference on your system?

I assume so, but I'm still working on testing the performance. My current idea is to time how long it takes to count from 0 to 32. Let me know if you have a different suggestion though!

mcevoypeter:
It actually appears that disabling page tracking on inner roads leads to a slight performance regression, at least in the test cases that @joemfb recommended.

Here's the comparison of the performance of the `urbit` binary built from the current tip of `master` (SHA `ac2337a6a7a0821530299cf5d49589a5ed229da0`) with the `urbit` binary built from the tip of `peter/page-tracking` (SHA `4a781339914b2c47772107d8654a27c40aef7ddf`):

|        | Dojo Command                                         |
|--------|------------------------------------------------------|
| Test 1 | `:time =\|(i=@ud \|-(?:(=(i 0xff.ffff) ~ $(i +(i)))))` |
| Test 2 | `:time (gulf 1 0xf.ffff)`                            |
| Test 3 | `:time (jam (gulf 1 0xf.ffff))`                      |
| Test 4 | `:time (cue (jam (gulf 1 0xf.ffff)))`                |



| Branch                | Test 1      | Test 2      | Test 3      | Test 4      |
|-----------------------|-------------|-------------|-------------|-------------|
| `master`              | `~s1..7469` | `~s0..4f88` | `~s1..241c` | `~s2..8f67` |
| `peter/page-tracking` | `~s1..9203` | `~s0..5f9d` | `~s1..3ff1` | `~s2..b464` |


`master` console log:
```console
> |start %time
>=
> :time =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
>=
[%took ~s1..7469]
> :time (gulf 1 0xf.ffff)
>=
[%took ~s0..4f88]
> :time (jam (gulf 1 0xf.ffff))
>=
[%took ~s1..241c]
> :time (cue (jam (gulf 1 0xf.ffff)))
>=
[%took ~s2..8f67]
> :time =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
>=
[%took ~s1..5409]
> :time (gulf 1 0xf.ffff)
>=
[%took ~s0..42ea]
> :time (jam (gulf 1 0xf.ffff))
>=
[%took ~s1..1fdb]
> :time (cue (jam (gulf 1 0xf.ffff)))
>=
[%took ~s2..7a6e]
```

`peter/page-tracking` console log:
```console
> |start %time
>=
> :time =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
>=
[%took ~s1..9203]
> :time (gulf 1 0xf.ffff)
>=
[%took ~s0..5f9d]
> :time (jam (gulf 1 0xf.ffff))
>=
[%took ~s1..3ff1]
> :time (cue (jam (gulf 1 0xf.ffff)))
>=
[%took ~s2..b464]
> :time =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
>=
[%took ~s1..6c2f]
> :time (gulf 1 0xf.ffff)
>=
[%took ~s0..5dab]
> :time (jam (gulf 1 0xf.ffff))
>=
[%took ~s1..3fb6]
> :time (cue (jam (gulf 1 0xf.ffff)))
>=
[%took ~s2..787e]
```

mcevoypeter:
@joemfb I was unable to use the `%bout` hint successfully. Using the binary from `master`, I hit the following:
```console
~bus:dojo> ~>  %bout  =|(i=@ud |-(?:(=(i 0xff) ~ $(i +(i)))))address 0x230000000a out of loom!
Assertion '0' failed in noun/events.c:128

bail: oops
bailing out
pier: serf unexpectedly shut down
```

joemfb:
@mcevoypeter that `%bout` crash is surprising, but I can reproduce it locally. It looks like it fails when it's wrapped around an expression that introduces new core:

```
> ~>  %bout  (add 2 2)
took 5µs
4
> ~>  %bout  (|=([p=@ q=@] (add p q)) 2 2)
address 0x2300000288 out of loom!
loom: [0x28000000000 : 0x28080000000)
Assertion '0' failed in noun/events.c:128
```

So there's an interpreter bug at the intersection of new cores and the dynamic hint protocol.

/cc @frodwith

joemfb:
I'm also surprised by these timing results, I expected this to be a straightforward win.

It would be really nice to know the cost of the new bookkeeping on leaping-from/falling-to the home road, I suspect that's the overhead we're seeing (but I'm not sure how to easily isolate and measure that). Restarting the process between each of these measurements would give us maximally-optimistic measurements (as we'd be starting out with full memory protection each time), but that might not be the right way to assess these changes.

In practice, the cost of formally-spurious page faults may be mostly amortized across the events run between snapshot updates, in which case decreasing the snapshot frequency could be an easy optimization.

philipcmonk:
It's not clear how much memory each of those use.  For the counting ones at least, might be good to add a `!:` to break the tail recursion.  And yeah, it would be good to run right after a sync, when all the pages should be clean.

mcevoypeter:
> I'm also surprised by these timing results, I expected this to be a straightforward win.

I was too. As I was thinking about this, two things occurred to me:

(1) We issue one `mprotect()` call for each clean page in `u3m_fall()`. Optimizing to call `mprotect()` once for each contiguous block of clean pages should lead to a performance boost.

(2) `U3_OS_NativePageBytes` is a macro that expands to `sysconf(_SC_PAGESIZE)` and is used twice in each invocation of `u3m_leap()` (by way of `U3_OS_NativePageWords`). Changing `U3_OS_NativePageBytes` to a static global variable that is instantiated only once in `u3m_init` should also lead to a performance boost.

After making both changes, I see small performance improvements over `master` on three of the four tests. Here are the updated results:

|        | Dojo Command                                         |
|--------|------------------------------------------------------|
| Test 1 | `:time =\|(i=@ud \|-(?:(=(i 0xff.ffff) ~ $(i +(i)))))` |
| Test 2 | `:time (gulf 1 0xf.ffff)`                            |
| Test 3 | `:time (jam (gulf 1 0xf.ffff))`                      |
| Test 4 | `:time (cue (jam (gulf 1 0xf.ffff)))`                |



| Branch                              | SHA         | Test 1      | Test 2      | Test 3      | Test 4      |
|-------------------------------------|-------------|-------------|-------------|-------------|-------------|
| `master`                            | `ac2337a6a` | `~s1..5bc2` | `~s0..472a` | `~s1..2798` | `~s2..7dce` |
| `peter/page-tracking` (optimized)   | `16c52a4b7` | `~s1..596c` | `~s0..4747` | `~s1..262b` | `~s2..5be8` |
| `peter/page-tracking` (unoptimized) | `4a7813399` | `~s1..9203` | `~s0..5f9d` | `~s1..3ff1` | `~s2..b464` |


`master` (`ac2337a6a`) console log:
```console
> |start %time
>=
> :time =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
>=
[%took ~s1..5bc2]
> :time (gulf 1 0xf.ffff)
>=
[%took ~s0..472a]
> :time (jam (gulf 1 0xf.ffff))
>=
[%took ~s1..2798]
> :time (cue (jam (gulf 1 0xf.ffff)))
>=
[%took ~s2..7dce]
```

`peter/page-tracking` (`16c52a4b7`) console log:
```console
> |start %time
>=
> :time =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
>=
[%took ~s1..596c]
> :time (gulf 1 0xf.ffff)
>=
[%took ~s0..4747]
> :time (jam (gulf 1 0xf.ffff))
>=
[%took ~s1..262b]
> :time (cue (jam (gulf 1 0xf.ffff)))
>=
```

mcevoypeter:
> It's not clear how much memory each of those use. For the counting ones at least, might be good to add a `!:` to break the tail recursion.

Maybe an obvious question, but where does the `!:` slot into the counting expression? Based off of the [`!:` section of the docs](https://urbit.org/docs/hoon/reference/rune/zap#-zapcol), I would've guessed that it would be
```console
dojo> :time !: =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
```
but the command line parser doesn't accept this as vaild.

frodwith:
> dojo> :time !: =|(i=@ud |-(?:(=(i 0xff.ffff) ~ $(i +(i)))))
You need either (for tall form) a gap (2 or more whitespace characters) between !: and the rest of the expression, or (for wide form) !:(=|(...))

mcevoypeter:
I moved the declaration of `U3_OS_NativePageWords` into [`portable.h`](https://github.com/urbit/urbit/tree/peter/page-tracking/pkg/urbit/include/c/portable.h) where it belongs. @joemfb: I was mistaken in thinking that doing so would require header guards. It turns out that we can safely declare global variables in [`portable.h`](https://github.com/urbit/urbit/tree/peter/page-tracking/pkg/urbit/include/c/portable.h), but we cannot define them there, so we just let the compiler default initialize `U3_OS_NativePageWords` to `0` and re-assign its value in `u3m_init()`.

joemfb:
@mcevoypeter I've thought about this some more, and our approach is precisely backwards. We want to avoid the cost of tracking unusable page metadata, but we're adding a lot of bookkeeping to do so. Instead of ensuring that the pages of contiguous free space inside the loom are write-protected, we should simply always keep them dirty.

The only pages in the snapshot are the in-use portions of the heap (`north.bin`) and stack (`south.bin`) of the home road. When we patch the snapshot, we only write the dirty pages within the heap or stack, ignoring any in the space between. If we preemptively dirty all pages in that contiguous space, we'll automatically capture all growth of the heap or stack. The only page faults we need to handle are the initial mutations of pages that are already in the snapshot.

In short, we should dirty/unprotect all pages on first boot. When we load or save a snapshot, we should write-protect and mark as clean all pages of the heap and stack, and dirty/unprotect all pages between. (As an additional optimization, we could also dramatically reduce the number `mprotect()` calls we make.) Sorry I didn't realize this sooner!

mcevoypeter:
@joemfb this is a great point. I implemented the change, the high points of which are:
- Revert all previous page tracking changes.
- Mark all loom pages as dirty and omit write protections in `u3e_live()` if the images were empty (i.e. on first boot).
- Mark all loom pages that were mapped in from the image as clean and enable write protections on them in `_ce_image_blit`. 

Assuming that I understood what you were saying correctly, we're now seeing a further performance bump, especially on Tests 3 and 4.

|        | Dojo Command                                         |
|--------|------------------------------------------------------|
| Test 1 | `:time =\|(i=@ud \|-(?:(=(i 0xff.ffff) ~ $(i +(i)))))` |
| Test 2 | `:time (gulf 1 0xf.ffff)`                            |
| Test 3 | `:time (jam (gulf 1 0xf.ffff))`                      |
| Test 4 | `:time (cue (jam (gulf 1 0xf.ffff)))`                |



| Branch                                             | SHA         | Test 1      | Test 2      | Test 3      | Test 4      |
|----------------------------------------------------|-------------|-------------|-------------|-------------|-------------|
| `master`                                           | `ac2337a6a` | `~s1..5bc2` | `~s0..472a` | `~s1..2798` | `~s2..7dce` |
| `peter/page-tracking` (default dirty)              | `174b06577` | `~s1..5739` | `~s0..4853` | `~s1..0d24` | `~s2..236f` |
| `peter/page-tracking` (default clean, optimized)   | `16c52a4b7` | `~s1..596c` | `~s0..4747` | `~s1..262b` | `~s2..5be8` |
| `peter/page-tracking` (default clean, unoptimized) | `4a7813399` | `~s1..9203` | `~s0..5f9d` | `~s1..3ff1` | `~s2..b464` |

pkg/urbit/noun/events.c
```diff
@@ -710,24 +673,31 @@ _ce_image_blit(u3e_image* img_u,
...
+      c3_w bit_w = pag_w & 31;
+      u3P.dit_w[blk_w] &= ~(1 << bit_w);
     }
-#endif
-    ptr_w += stp_ws;
+  }
+
+  if ( 0 != mprotect(ptr_w, siz_w, PROT_READ) ) {
+    fprintf(stderr, "loom: live mprotect: %s\r\n", strerror(errno));
+    c3_assert(0);
```
joemfb, line 700: This change doesn't look right -- I suspect it only works because `south.bin` is alway a single page.

The old loop used to work a page at a time, iterating in opposite directions for north/south segments (`ptr_w += stp_ws;`). If I'm reading things correctly, `south.bin` is actually paginated left-to-right on disk, while it's right-to-left in memory. So I think you'll need to restore the loop and dirty/mprotect a page at a time.

joemfb (changes requested):
Thanks for doing this!

You should also change `_ce_patch_junk_page()` to preserve the property that unused inner free pages are always dirty and read+write.

pkg/urbit/noun/events.c
mcevoypeter: I overlooked this fact, so thanks for pointing it out. Out of curiosity, will it ever be possible to migrate to an approach where we iterate over the north and south segments in the same direction? From a performance standpoint, it would be ideal if we could issue a `read()` call and an  `mprotect()` call per segment rather than per page, which would require us to iterate from low addresses to high addresses. That said, my suspicion is that the current format of `south.bin` (pages at higher addresses coming first on disk) will make this infeasible if not impossible to switch.

mcevoypeter:
> You should also change _ce_patch_junk_page() to preserve the property that unused inner free pages are always dirty and read+write.

Sanity check me on this, but I think we can remove `_ce_patch_junk_page()`. It's only used on pages between the north and south watermarks (i.e. inner road pages) in `_ce_patch_compose()`, so if those pages are dirty and writable by default, then we should be able to remove it entirely and simply do nothing for the inner road pages on `_ce_patch_compose()`.

pkg/urbit/noun/events.c
joemfb: Yeah, it seems like that would be better. I'd be inclined to go with whatever approach will work best for demand paging. We'll have to migrate old snapshots if we change the layout, but that shouldn't be too difficult.

mcevoypeter: I'm in favor of iterating over both segments in the same direction since it'll be easier to reason about, but I think the change should be put into a separate PR. For the sake of landing the inner road page tracking changes, I've switched back to a per-page approach to `read()` and `mprotect()` in `_ce_image_blit()`.

joemfb (approved):
LGTM

You're right about `_ce_patch_junk_page()`. The only way the inner space grows is by reclaiming pages from the snapshot (and, in practice only from the heap `north.bin`), and those pages will have been dirtied by the fault handler.

Please target the `next/vere` branch, and try these changes on a pre-existing ship as a sanity check before merging. (I don't see any potential issues, but it's good to be paranoid about this area.)

mcevoypeter:
I ran my ship using the binary built from 2af12f1ee3 (the tip of this branch) today, and I had no issues sending DMs or receiving messages in groups, which should be sufficient as a sanity check. Merging into `next/vere`.

mrdomino:
Ah, `|mount`ing `%base` led to memory corruption on a fakezod for me just now; git bisect showed this commit to be the culprit.

philipcmonk:
Mounting %base is most notable for sending a much larger effect (several
MBs) across the IPC pipe than almost anything else. Allocation for that
happens in the home road. No obvious problem but it's suggestive.

PR #5481, opened 2021-12-08
fmt: auto format whitespace in C code using clang-format

mcevoypeter:
Use `clang-format` to format whitespace in C source and header files. The current configuration conforms as closely to the existing style of C code as possible, including:
- Align pointers to the left.
```c
void* ptr_v;
```
- Put the return type of a function on its own line.
```c
c3_w
fun(void)
{

}
```
- Include the open curly brace on same line as control statements, struct declarations, and enum declarations.
```c
if ( c3n == var_o ) {

}

for ( var_w = 0; var_w < 137; var_w ++ ) {

}

while ( 137 > var_w ) {

}

enum bar {

};


struct foo {

};

```
- Add the open curly brace to a new line for function declarations and before else statements.
```c
if ( c3n == var_o ) {

}
else {

}

c3_w
fun(void)
{

}
```
- Add leading and trailing space inside of parentheses.
```c
if ( c3n == var_o && c3y == rav_o ) {

}
```
- Align consecutive variable declarations and assignments.
- Align macro definitions.
- Align inline comments.

Note that the above is a non-exhaustive list of the formatting options selected. For a comprehensive view, take a look at `.clang-format` and the [clang-format options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html).

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,131 @@
+# https://clang.llvm.org/docs/ClangFormatStyleOptions.html
+---
+Language: Cpp
+AlignAfterOpenBracket: Align
+#AlignArrayOfStructures: Left # >= 13.0
+AlignConsecutiveAssignments: Consecutive
+AlignConsecutiveBitFields: Consecutive
+AlignConsecutiveDeclarations: Consecutive
+AlignConsecutiveMacros: AcrossEmptyLinesAndComments
+AlignEscapedNewlines: Left
+AlignOperands: Align
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: Empty
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortEnumsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterReturnType: All
+AlwaysBreakBeforeMultilineStrings: true
+#AttributeMacros:
+  #- __capability
+BinPackArguments: false
+BinPackParameters: false
+BitFieldColonSpacing: None
+BraceWrapping:
+  AfterCaseLabel: false
+  AfterControlStatement: MultiLine
+  AfterEnum: false
+  AfterFunction: true
+  AfterStruct: false
+  AfterUnion: false
+  BeforeElse: true
+  BeforeWhile: false
+  IndentBraces: false
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Custom
+BreakBeforeTernaryOperators: true
+BreakStringLiterals: true
+ColumnLimit: 80
+#CommentPragmas:  '^ IWYU pragma:'
+ContinuationIndentWidth: 2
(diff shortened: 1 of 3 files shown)


brendanhay:
I'd suggest adding [clang-tools](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/tools/clang-tools/default.nix) to [shell.nix](https://github.com/urbit/urbit/blob/ac2337a6a7a0821530299cf5d49589a5ed229da0/shell.nix#L54-L60) to make `clang-format` available in the shell.

joemfb:
How big is the diff when you run this?

mcevoypeter:
> How big is the diff when you run this?

It's large. Of the 343 files in [pkg/](https://github.com/urbit/urbit/tree/peter/clang-format/pkg), 334 of them are modified when running [sh/fmt-c](https://github.com/urbit/urbit/blob/peter/clang-format/sh/fmt-c). I could likely tweak [.clang-format](https://github.com/urbit/urbit/blob/peter/clang-format/.clang-format) so that the diff shrinks, but I'm skeptical that doing so will make much of a dent. There are a variety of whitespace conventions sprinkled throughout [pkg/](https://github.com/urbit/urbit/tree/peter/clang-format); a large diff is likely inevitable if we want to settle on a standard.

mcevoypeter:
I've noticed that the build fails with `error: unknown type name 'u3_noun` after running [sh/fmt-c](https://github.com/urbit/urbit/blob/peter/clang-format/sh/fmt-c), which I suspect is caused by the fact that (1) `clang-format` reorders the inclusion of header files to be alphabetical and (2) we lack header guards in some of our header files, most notably those in [pkg/urbit/include](https://github.com/urbit/urbit/blob/peter/clang-format/pkg/urbit/include).

philipcmonk:
Would it be possible to see the diff?  Eg if you commit that to another branch and push, then we can see the diff here.

ohAitch:
How many of the `pkg/` directories contain c? Standardizing things towards "conventional vere style" is probably good, since that should be most of it. (Though tbh I'm skeptical of automated tools' ability to do this as opposed to putting inappropriate newlines everywhere.)

mcevoypeter:
> Would it be possible to see the diff? Eg if you commit that to another branch and push, then we can see the diff here.

Definitely. I ran `sh/fmt-c` and pushed the changes to [`peter/clang-format-diff`](https://github.com/urbit/urbit/tree/peter/clang-format-diff).

> How many of the `pkg/` directories contain c? Standardizing things towards "conventional vere style" is probably good, since that should be most of it. (Though tbh I'm skeptical of automated tools' ability to do this as opposed to putting inappropriate newlines everywhere.)

There are four directories in `pkg/` that contain C: `pkg/ent`, `pkg/hs`, `pkg/urcrypt`, and `pkg/urbit`. `clang-format` works pretty well in my experience, although there are some things specific to our codebase that it's unable to enforce. For example, the extra indentation that is currently used in many of the header files (at least those in `pkg/urbit/include`) would be removed, and multi-line comments would change from
```c
/* This is a comment.
*/
```
to 
```c
/* This is also a comment.
 */
```
The benefit of automating code formatting is that once we properly configure the tool, we no longer have to worry about manually enforcing a style. In my opinion, removing that overhead from the development process makes it worth it, even if the code style has to change slightly in certain cases.

zalberico:
@philipcmonk putting this on your radar for a decision about auto formatting whitespace - I tend to think this reduces overhead by using tools to standardize this, but leave the decision to you to make the call. We should make a decision either way though.

PR #5484, opened 2021-12-09
grid / landscape: `theme-color` updates

mwvd:
https://user-images.githubusercontent.com/4127151/145456964-7ba6c3f6-16ae-4afa-a089-81b25644f70a.mp4

Notes:
- Correct order for `theme_color` meta tags seems to be:
	```html
	<link rel="manifest" href="..." />
    <!--
        After the manifest's `theme_color` value, theme color can be overwritten with
		"theme-color" meta tags
		Ref for this behaviour (manifest vs meta) is here:
		https://w3c.github.io/manifest/#theme_color-member

		From here, the browser selects the first possible `theme-color` meta tag that matches,
		so we add our media queries first, and then if the browser doesn't respect either of
		those, it uses the fallback
	-->
    <meta name="theme-color" content="#ffffff" media="(prefers-color-scheme: light)" />
    <meta name="theme-color" content="#000000" media="(prefers-color-scheme: dark)" />
    <meta name="theme-color" content="#ffffff" />
	```


urcades:
@arthyn + @liam-fitzgerald + @Fang- 

I basically support us merging this asap if possible — it's a simple change, would make everything feel a lot more pleasant.

arthyn (approved):

mwvd:
thanks @arthyn @urcades 💙

PR #5487, opened 2021-12-10
roller: add utilities for handling low-gas batches

yosoyubik:
This adds support for handling cases when the send-batch thread failed [[1](https://github.com/urbit/urbit/commit/f6ff4f108bfe9b924f01e25bc63c679df7c5e20c#diff-93b40677881b0eba00b68e301f530361e25ba8762a0141a6a96f95f078c88e1fR1255)]-[[2](https://github.com/urbit/urbit/commit/f6ff4f108bfe9b924f01e25bc63c679df7c5e20c#diff-93b40677881b0eba00b68e301f530361e25ba8762a0141a6a96f95f078c88e1fR1247)],
mainly among them, a thread crash. One of the events that causes this
behavior is a ver low gas price for this L1 transaction.

Here we add support for manually bumping the price for such transaction,
and for increasing the default fallback gas-price, together with discarding
any malformed batch from the sending queue.


philipcmonk (approved):
These will be useful

PR #5489, opened 2021-12-13
groups: fix permission handling

liam-fitzgerald:
Aggressively refactors the permissioning flow for groups, so as to make more reliable

On leaving a group:
- Leaving a group is triggered through the `-group-leave` thread
- This pokes group-push-hook, to remove ourselves from the groups, group-store to remove our copy of the group, and group-pull-hook, to cancel syncing of group
- When hosts hear of a %remove-members fact in the metadata-push-hook, it revokes the ships included in that fact. This should trigger the usual nacking and cleanup flow, which has been thoroughly tested.

### DIFF
--- a/pkg/landscape/app/metadata-push-hook.hoon
+++ b/pkg/landscape/app/metadata-push-hook.hoon
@@ -1,6 +1,6 @@
 ::  metadata-push-hook [landscape]:
 ::
-/-  *group, *invite-store, store=metadata-store
+/-  *group, *invite-store, store=metadata-store, group-store
 /+  default-agent, verb, dbug, grpl=group, push-hook,
     resource, mdl=metadata, gral=graph, agentio
 ~%  %group-hook-top  ..part  ~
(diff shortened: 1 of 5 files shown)


pkg/landscape/app/metadata-push-hook.hoon
```diff
@@ -82,7 +101,44 @@
...
+    ++  hosting-graphs
+      |=  rid=resource
+      ^-  (set resource)
+      =/  graphs=associations:store
+        (app-metadata-for-group:met resource.update %graph)
+      %-  ~(gas in *(set resource))
+      %+  murn  ~(tap in ~(key by graphs))
+      |=  [app=term graph=resource]
+      ?.  =(our.bowl entity.graph)  ~
+      `graph
```
Fang-, line 139: Using `+skim` (or `+skip`) might be semantically cleaner.

Fang- (changes requested):
Not sure the `+on-load` here checks out, but looks reasonable otherwise.

pkg/landscape/app/metadata-push-hook.hoon
liam-fitzgerald: using the filtering functions would necessitate another transform, which imo kinda negates any improvement in clarity

pkg/landscape/app/metadata-push-hook.hoon
```diff
@@ -43,11 +53,20 @@
...
+  :_  this
+  ~[watch-groups:hc]
+::
+++  on-save  !>(state)
+++  on-load
+  |=  =vase
+  =+  !<(old=versioned-state vase)
+  ?:  ?=([%0 ~] old)  `this
+  :_  this
+  ~[watch-groups:hc]
```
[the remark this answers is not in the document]
liam-fitzgerald: Right, but the trick here is that the `state-zero` case has never been used (check the diff) I realised this halfway through, but forgot to excise the extra case (this has now been addressed)

Fang- (approved):
You are right on both counts, oversights on my part. Approved!

PR #5491, opened 2021-12-14
u3: add header guards to header files in pkg/urbit/include

mcevoypeter:
Add header guards to the header files in [`pkg/urbit/include`](https://github.com/urbit/urbit/tree/peter/header-guards/pkg/urbit/include), which removes the need for the forward declaration of functions referred to in header files (see line 359 in [`allocate.h`](https://github.com/urbit/urbit/blob/master/pkg/urbit/include/noun/allocate.h)).


pkg/urbit/include/ur/ur.h
```diff
@@ -1,9 +1,13 @@
-#ifndef UR_UR_H
-#define UR_UR_H
+/* pkg/urbit/include/ur/ur.h
+**
+** This file is in the public domain.
+*/
+#ifndef URBIT_UR_UR_H
+#define URBIT_UR_UR_H
```
joemfb, line 6: This should just be `UR_H` (idk what I was thinking)

pkg/urbit/include/vere/db/lmdb.h
```diff
@@ -1,5 +1,9 @@
-/* include/vere/db/lmdb-impl.h
+/* pkg/urbit/include/vere/db/lmdb-impl.h
```
joemfb, line 1: IDK how `-impl` got in here, but that should be removed.

pkg/urbit/include/vere/vere.h
```diff
@@ -1,5 +1,9 @@
-/* include/vere/vere.h
+/* pkg/urbit/include/vere/vere.h
+**
+** This file is in the public domain.
 */
+#ifndef URBIT_VERE_VERE_H
+#define URBIT_VERE_VERE_H
```
joemfb, line 6: This should just be `U3_VERE_H`.

joemfb (changes requested):
Cross-cutting changes turn up all sorts of interesting things!

- `include/jets/{g,l,y}.h` are left over from an old staging process (c20c66901d15f5a0bda3d113088bfccfd31cb02d), and should simply be deleted.
- the public domain notices are vestigial and should be removed -- the entire repository is MIT licensed.
- there's unnecessary namespacing in these symbols and comments

To the last point, some extensive stylistic nits:

The contents of `pkg/urbit` don't really "know" they're in a monorepo, and I don't think they should. So I'd remove `pkg/urbit/` from the paths in the header comments. Those paths are kind of a questionable convention anyway, so I wouldn't extend them to the `ur` headers (ur is basically a separate library, inlined for build simplicity / out of laziness).

As for the symbols themselves, I think we can just use the existing namespace conventions:

- `include/c/defs.` s/b `C3_DEFS_H`
- `include/jets/k.h` s/b `U3_JETS_K_h`
- the core implemention of `u3` is in `noun/`, so `include/noun/allocate.h` s/b `U3_ALLOCATE_H` (`u3` is, or should be `libnoun`; the allocation functions are namespaced under `u3a_`, not `u3_noun_a_` or whatever)
- `include/ur/defs.h` s/b `UR_DEFS_H` (the ur change should be reverted)

A couple more specific tweaks:

mcevoypeter:
> The contents of `pkg/urbit` don't really "know" they're in a monorepo, and I don't think they should. So I'd remove `pkg/urbit/` from the paths in the header comments. Those paths are kind of a questionable convention anyway, so I wouldn't extend them to the `ur` headers (ur is basically a separate library, inlined for build simplicity / out of laziness).

Is it worth just removing the header comments entirely? I don't think they add much, and they require updating if files ever move around.

pkg/urbit/include/vere/vere.h
mcevoypeter: On a similar note, should the header guard in [`serf.h`](https://github.com/urbit/urbit/tree/peter/header-guards/pkg/urbit/include/vere/serf.h) be `U3_VERE_SERF_H` or `U3_SERF_H`?

joemfb (approved):
LGTM

Feel free to kill the head path-comments. And please target the `next/vere` branch.

pkg/urbit/include/vere/vere.h
joemfb: Either one is fine, I'm gonna remove that file soon.

mcevoypeter: Went with `U3_VERE_SERF_H`.

pkg/urbit/include/vere/db/lmdb.h
mcevoypeter: Removed the header path comments.

PR #5492, opened 2021-12-14
Adding tag == c3_doze check to _behn_io_kick in Behn io driver

laudiacay:
@joemfb and I (really just Joe) noticed this bug in the Behn io driver when he was walking me through how IO drivers work. it wasn't checking that the tag on the kick was c3__doze, and now it does.

Yay! first PR!


joemfb (approved):
LGTM, thanks!

(The CI failures are spurious, and fixed on another branch.)

PR #5493, opened 2021-12-14
u3: fix compilation of dynamic hints in tail position

joemfb:
This PR fixes a miscompilation of dynamic hints in the bytecode interpreter. Calls (nock 9) in hinted formulas in tail position were being compiled into tail calls, but hinted formulas (in the dynamic hint protocol) are never in tail position, as there is a post-evaluation callback.

This bug caused a crash, which was first observed in #5477. Dynamic hints were added in #3979, and first used in #5261.

/cc @frodwith


joemfb:
This PR doesn't require any special release staging, as the only deployed use-site is in the test suite (and also doesn't hit the bug).

mcevoypeter:
Makes sense to me.

mcevoypeter (approved):

PR #5495, opened 2021-12-14
helm: bind /who.json

Fang-:
Before the software distribution era, file-server used to bind this. Ships that existed pre-dist still have that bound, pointing to the generator in `%home`. Ships that booted directly into the softdist world though, never got this endpoint bound.

Apparently some things still (want to) depend on this. This PR simply restores the endpoint and makes sure it's pointing to the generator on base. Some slightly odd coupling here, with the generator on base, and docket (who now sets up the binding) living in garden, but unless we want to have kiln set this up I'm not sure there's a better place to do this right now.

### DIFF
--- a/pkg/arvo/app/hood.hoon
+++ b/pkg/arvo/app/hood.hoon
@@ -2,27 +2,28 @@
 /+  drum=hood-drum, helm=hood-helm, kiln=hood-kiln
 |%
 +$  state
-  $~  [%22 *state:drum *state:helm *state:kiln]
-  $>(%22 any-state)
+  $~  [%23 *state:drum *state:helm *state:kiln]
+  $>(%23 any-state)
 ::
 +$  any-state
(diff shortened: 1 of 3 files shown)


lukestiles:
Is there a better way to perform this function without reintroducing the coupling?

Fang-:
If we want it to bind automatically we're gonna have to couple it to _some_ app. The argument for doing this in hood is that that's where all the miscellaneous kernel interaction happens already, but also imo it doesn't cleanly fit into any of its subsystems.

lukestiles:
Got it. Sounds like this is good for now then. Thanks.

joemfb:
It seems like this would fit reasonably well in helm.

Fang-:
I had completely forgotten about helm, good call @joemfb!

arthyn:
@liam-fitzgerald anything blocking your approval? we should get this merged for pocket folks

philipcmonk:
@liam-fitzgerald approved out-of-band

PR #5500, opened 2021-12-17
u3: remove obfuscating alias macros

mcevoypeter:
Replace the macro aliases from [`pkg/urbit/include/noun/aliases.h`](https://github.com/urbit/urbit/tree/next/vere/pkg/urbit/include/noun/aliases.h) in the first and second columns of the table below with the macro/function that they resolve to (the third column) except for `u3v_d[ctq]`, which are not pure aliases. Functionality was verified by invoking `make && urbit bus` at the Nix shell.
| Alias                     | Intermediate alias          | Function                              |
|---------------------------|-----------------------------|---------------------------------------|
| `u3h(som)`                |                             | `u3x_h(som)`                          |
| `u3t(som)`                |                             | `u3x_t(som)`                          |
| `u3at(axe, som)`          |                             | `u3x_at(axe, som)`                    |
| `u3nc(a, b)`              |                             | `u3i_cell(a, b)`                      |
| `u3nt(a, b, c)`           |                             | `u3i_trel(a, b, c)`                   |
| `u3nq(a, b, c, d)`        |                             | `u3i_qual(a, b, c, d)`                |
| `u3nl`                    |                             | `u3i_list`                            |
| `u3du(som)`               | `u3r_du(som)`               | `u3a_is_cell(som)`                    |
| `u3ud(som)`               | `u3r_ud(som)`               | `u3a_is_atom(som)`                    |
| `u3k(som)`                |                             | `u3a_gain(som)`                       |
| `u3z(som)`                |                             | `u3a_lose(som)`                       |
| `u3do(txt_c, arg)`        |                             | `u3v_do(txt_c, arg)`                  |
| `u3dc(txt_c, a, b)`       | `u3v_dc(txt_c, a, b)`       | `u3v_do(txt_c, u3i_cell(a, b))`       |
| `u3dt(txt_c, a, b, c)`    | `u3v_dt(txt_c, a, b, c)`    | `u3v_do(txt_c, u3i_trel(a, b, c))`    |
| `u3dq(txt_c, a, b, c, d)` | `u3v_dq(txt_c, a, b, c, d)` | `u3v_do(txt_c, u3i_qual(a, b, c, d))` |

For a developer who is getting up to speed with `u3`, the aliases make it difficult to understand what any given chunk of code is doing without looking up the alias definition at least once. Many of the alias names are not intuitive, which means that more often than not, the alias definition has to be looked up multiple times before it sticks temporarily in short-term memory. And just because the alias sticks in short-term memory does not mean that it will stick in long-term memory, so the developer may have to repeat the process a few days or weeks later when they revisit the same chunk of code. As an example, the following if statement changes from being cryptic to intuitive by removing the alias:
```c
// With alias
if ( c3n == u3du(b) ) {
    return u3_nul;
}

// Without alias
if ( c3n == u3a_is_cell(b) ) {
    return u3_nul;
}
```
The aliases `u3h`, `u3t`, and `u3at` are very close in name to the actual macro/function that they resolve to. Even in this case, the `u3x_` prefix adds valuable information for a developer, namely that the macro definition can be found in [`pkg/urbit/include/noun/xtract.h`](https://github.com/urbit/urbit/tree/next/vere/pkg/urbit/include/noun/xtract.h). If, for example, `u3h` is used instead of `u3x_h` and the developer does not recognize `u3h`, they have to look up the definition in [`pkg/urbit/include/noun/aliases.h`](https://github.com/urbit/urbit/tree/next/vere/pkg/urbit/include/noun/aliases.h), which will show that `u3h` resolves to `u3x_h`. In other words, the alias is only serving as an unnecessary level of indirection. Undoubtably, the aliases save the experienced developer a few keystrokes while writing code, but as Urbit continues to mature, the ratio of time spent reading code to writing code will only increase. We should optimize for the reading case rather than the writing case, especially when the number of keystrokes saved is small. And perhaps most importantly, **two names are not necessary when one will do**. 

One major objection to this PR is the size of the diff, which is massive. As a result of the number of lines of code that this PR touches, it will unfortunately add another layer of indirection for the developer using `git blame` to navigate. However, any given chunk of code within `u3` is likely read an order of magnitude more times than `git blame` is used on it. So even by increasing the overhead of using `git blame`, we'll still win because we're making the common case faster. Also, to alleviate concern surrounding *how* these changes were made, here's the shell script that was used:
```bash
#!/usr/bin/env bash

# Replace an original macro with a replacement macro in all C files and delete the original macro 
# pattern from the header file in which it was defined.
#
# @param $1   the original macro pattern.
# @param $2   the macro pattern to replace the original with.
# @param $3   the header file in pkg/urbit/include/noun in which the original
#             macro pattern is defined.
replace() {
  sed -i='' "/define *$1(/d" "pkg/urbit/include/noun/$3"
  rg -l -t=c "$1\(" \
    | xargs sed -i='' "s/$1(/$2(/g"
}

#       Original Replacement   Source of original definition
replace "u3h"    "u3x_h"       "aliases.h"
replace "u3t"    "u3x_t"       "aliases.h"
replace "u3at"   "u3x_at"      "aliases.h"

replace "u3nc"   "u3i_cell"    "aliases.h"
replace "u3nt"   "u3i_trel"    "aliases.h"
replace "u3nq"   "u3i_qual"    "aliases.h"

replace "u3nl"   "u3i_list"    "aliases.h"
replace "u3du"   "u3r_du"      "aliases.h"
replace "u3ud"   "u3r_ud"      "aliases.h"

replace "u3k"    "u3a_gain"    "aliases.h"
replace "u3z"    "u3a_lose"    "aliases.h"

replace "u3do"   "u3v_do"      "aliases.h"
replace "u3dc"   "u3v_dc"      "aliases.h"
replace "u3dt"   "u3v_dt"      "aliases.h"
replace "u3dq"   "u3v_dq"      "aliases.h"

replace "u3r_du" "u3a_is_cell" "retrieve.h"
replace "u3r_ud" "u3a_is_atom" "retrieve.h"
```
Note that a few small post clean up tasks were required after running the script above, namely:
- delete stale comments in [`pkg/urbit/include/noun/aliases.h`](https://github.com/urbit/urbit/tree/next/vere/pkg/urbit/include/noun/aliases.h);
- define the following macros in [`pkg/urbit/include/noun/vortex.h`](https://github.com/urbit/urbit/tree/next/vere/pkg/urbit/include/noun/vortex.h);
```c
#     define u3v_dc(txt_c, a, b)       u3v_do(txt_c, u3i_cell(a, b))
#     define u3v_dt(txt_c, a, b, c)    u3v_do(txt_c, u3i_trel(a, b, c))
#     define u3v_dq(txt_c, a, b, c, d) u3v_do(txt_c, u3i_qual(a, b, c, d))
```
- and delete the `#if 1 #else` block related to the `u3r_du` and `u3r_ud` aliases at the top of [`pkg/urbit/include/noun/retrieve.h`](https://github.com/urbit/urbit/tree/next/vere/pkg/include/noun/retrieve.h).

### DIFF
--- a/pkg/urbit/include/noun/imprison.h
+++ b/pkg/urbit/include/noun/imprison.h
@@ -148,7 +148,6 @@
         */
           u3_noun
           u3i_list(u3_weak som, ...);
-#         define u3nl u3i_list
 
         /* u3i_edit():
         **
--- a/pkg/urbit/include/noun/retrieve.h
+++ b/pkg/urbit/include/noun/retrieve.h
@@ -9,7 +9,7 @@
 
     /** u3r_*: read without ever crashing.
     **/
-#       define u3du(som)  (u3a_is_cell(som))
+#       define u3ic(som)  (u3a_is_cell(som))
 #       define u3ud(som)  (u3a_is_atom(som))
 
       /* u3r_at(): fragment `a` of `b`, or u3_none.
--- a/pkg/urbit/include/noun/retrieve.h
+++ b/pkg/urbit/include/noun/retrieve.h
@@ -9,23 +9,8 @@
 
     /** u3r_*: read without ever crashing.
     **/
-#if 1
-#       define u3r_du(a)  u3a_is_cell(a)
-#       define u3du(som)  (u3r_du(som))
-
-#       define u3r_ud(a)  u3a_is_atom(a)
-#       define u3ud(som)  (u3r_ud(som))
-#else
-      /* u3r_du(): c3y iff `a` is cell.
-      */
-        c3_o
-        u3r_du(u3_noun a);
-
-      /* u3r_ud(): c3n iff `a` is cell.
-      */
-        c3_o
-        u3r_ud(u3_noun a);
-#endif
+#       define u3du(som)  (u3a_is_cell(som))
+#       define u3ud(som)  (u3a_is_atom(som))
 
       /* u3r_at(): fragment `a` of `b`, or u3_none.
       */
--- a/pkg/urbit/include/noun/retrieve.h
+++ b/pkg/urbit/include/noun/retrieve.h
@@ -9,7 +9,7 @@
 
     /** u3r_*: read without ever crashing.
     **/
-#       define u3ic(som)  (u3a_is_cell(som))
+#       define u3du(som)  (u3a_is_cell(som))
 #       define u3ud(som)  (u3a_is_atom(som))
 
       /* u3r_at(): fragment `a` of `b`, or u3_none.
--- a/pkg/urbit/include/noun/retrieve.h
+++ b/pkg/urbit/include/noun/retrieve.h
@@ -10,7 +10,7 @@
     /** u3r_*: read without ever crashing.
     **/
 #       define u3ic(som)  (u3a_is_cell(som))
-#       define u3ia(som)  (u3a_is_atom(som))
+#       define u3ud(som)  (u3a_is_atom(som))
 
       /* u3r_at(): fragment `a` of `b`, or u3_none.
       */
--- a/pkg/urbit/include/noun/retrieve.h
+++ b/pkg/urbit/include/noun/retrieve.h
@@ -10,7 +10,7 @@
     /** u3r_*: read without ever crashing.
     **/
 #       define u3ic(som)  (u3a_is_cell(som))
-#       define u3ud(som)  (u3a_is_atom(som))
+#       define u3ia(som)  (u3a_is_atom(som))
 
       /* u3r_at(): fragment `a` of `b`, or u3_none.
       */
--- a/pkg/urbit/include/noun/vortex.h
+++ b/pkg/urbit/include/noun/vortex.h
@@ -65,7 +65,7 @@
       u3v_do(const c3_c* txt_c, u3_noun arg);
 #     define  u3v_dc(txt_c, a, b)        u3v_do(txt_c, u3nc(a, b))
 #     define  u3v_dt(txt_c, a, b, c)     u3v_do(txt_c, u3nt(a, b, c))
-#     define  u3dq(txt_c, a, b, c, d)  u3v_do(txt_c, u3nq(a, b, c, d))
+#     define  u3v_dq(txt_c, a, b, c, d)  u3v_do(txt_c, u3nq(a, b, c, d))
 
     /* u3v_wish(): text expression with cache.
     */
--- a/pkg/urbit/include/noun/vortex.h
+++ b/pkg/urbit/include/noun/vortex.h
@@ -63,7 +63,7 @@
     */
       u3_noun
       u3v_do(const c3_c* txt_c, u3_noun arg);
-#     define  u3dc(txt_c, a, b)        u3v_do(txt_c, u3nc(a, b))
+#     define  u3v_dc(txt_c, a, b)        u3v_do(txt_c, u3nc(a, b))
 #     define  u3dt(txt_c, a, b, c)     u3v_do(txt_c, u3nt(a, b, c))
 #     define  u3dq(txt_c, a, b, c, d)  u3v_do(txt_c, u3nq(a, b, c, d))
 
--- a/pkg/urbit/include/noun/vortex.h
+++ b/pkg/urbit/include/noun/vortex.h
@@ -64,7 +64,7 @@
       u3_noun
       u3v_do(const c3_c* txt_c, u3_noun arg);
 #     define  u3v_dc(txt_c, a, b)        u3v_do(txt_c, u3nc(a, b))
-#     define  u3dt(txt_c, a, b, c)     u3v_do(txt_c, u3nt(a, b, c))
+#     define  u3v_dt(txt_c, a, b, c)     u3v_do(txt_c, u3nt(a, b, c))
 #     define  u3dq(txt_c, a, b, c, d)  u3v_do(txt_c, u3nq(a, b, c, d))
 
     /* u3v_wish(): text expression with cache.
--- a/pkg/urbit/include/noun/vortex.h
+++ b/pkg/urbit/include/noun/vortex.h
@@ -63,7 +63,6 @@
     */
       u3_noun
       u3v_do(const c3_c* txt_c, u3_noun arg);
-#     define  u3do(txt_c, arg)         u3v_do(txt_c, arg)
 #     define  u3dc(txt_c, a, b)        u3v_do(txt_c, u3nc(a, b))
 #     define  u3dt(txt_c, a, b, c)     u3v_do(txt_c, u3nt(a, b, c))
 #     define  u3dq(txt_c, a, b, c, d)  u3v_do(txt_c, u3nq(a, b, c, d))
--- a/pkg/urbit/include/noun/xtract.h
+++ b/pkg/urbit/include/noun/xtract.h
@@ -93,11 +93,10 @@
         u3_noun
         u3x_good(u3_weak som);
 
-      /* u3x_at (u3at): fragment.
+      /* u3x_at: fragment.
       */
         u3_noun
         u3x_at(u3_noun axe, u3_noun som);
-#     define u3at(axe, som)    u3x_at(axe, som)
 
       /* u3x_mean():
       **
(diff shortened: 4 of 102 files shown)


Fang-:
> obfuscating

I don't know about this one. Of course, I'm gonna be biased here; not only did I _learn_ the aliases, I _like_ them.

I can see how the initial "huh what's this thing" followed by alias chasing is not a Good experience, but most of these are common enough, and their mnemonic mappings easy enough* that they'll stick fairly quickly. I'm not convinced the readability gains for newcomers here outweigh the readability hit for people who have seen these things more than twice.

(* the only not-straightforward ones are `u3ud` and `u3du`, which I remember as "`@ud`, so atom" and "du for double for tuple", and `u3k` and `u3z`, which I just remember by their ubiquity (and "k for keep").)

Removing aliases for the commonly in-lined functions here is going to actually hurt readability in many cases. `u3k` is the worst offender (and makes the least sense to not in-line), cell construction ones coming in at a close second. The `_do` ones can honestly go either way afaic, they don't come up that often, and I've rarely ever seen/written them in-lined.

(Also hazing is good actually.)

tacryt-socryp:
I agree wholeheartedly with the rationale behind this PR and think any objection should be ignored.

ohAitch:
Yeah I would not say the problem with this diff is it being large, the problem is it makes the codebase worse, by trading off skimability by someone familiar with the project for skimability for anyone who isn't. Jets in particular are heavily affected, by making the already verbose mental translation out of hoon even more verbose.

Like, the u3 dialect of c is not standard c, there's a lot of greenspinning insofar as nock+jets is in a fairly literal sense a lisp. Primitives like recounts and cell construction/destructuring and `.?` and function calls, staying out of your way as much as possible, matters. If people don't want to learn a lisp, this is fine for them, but going to cause problems if the hope is to understand and propose changes to a lisp runtime.

philipcmonk:
> staying out of your way as much as possible

Yes, this is the goal.  You want small things to feel small, and you want big things to feel big.

The ones I'd keep are:
- `u3z`, `u3k` commonly used inline, very important they don't make lines too big
- `u3h,` `u3t` commonly used, and some of them are used several times on the same line.  Also easy to remember
- `u3nc`, `u3nt`, `u3nq` commonly used, and when you're building nouns, you often use several of these.  You don't want constructing a tree to be visually expensive -- that's one of the main things we do.

The ones I'd remove are:
- `u3at` didn't know this existed, and the prefix is very important for the difference between `u3r_at` vs `u3x_at`
- `u3nl` rare, hard to remember
- `u3du`, `u3ud` hard to visually recognize.  Never nested, and they're usually in if statements, so line length doesn't matter much
- `u3d[octq]` could go either way, but invoking nock is not a small thing, and it's not super common.  The `v` prefix helps to remind that we're doing nock too.

jtobin:
> Yeah I would not say the problem with this diff is it being large, the problem is it makes the codebase worse, by trading off skimability by someone familiar with the project for skimability for anyone who isn't. Jets in particular are heavily affected, by making the already verbose mental translation out of hoon even more verbose.

I agree with this, and FWIW found the aliases *qua* aliases gave me no trouble when learning to work with u3 -- I found them to be pretty immediately intuitive, and easy to remember. 

If some must be pruned, I vote to keep all the ones @philipcmonk listed, as well as `u3ud` and `u3du`.

mcevoypeter:
Thanks everyone for the feedback on what I know is a contentious change. Based off of the comments above, @philipcmonk's suggestion to remove only some of the aliases sounds like the best balance of first-time readability and repeat skimmability. I compiled the number of occurrences of each of the aliases that Phil mentioned. With the exception of `u3ud` and `u3du`, all of the aliases that have been removed occur fewer than 50 times each across the entire codebase, and a lot of them occur fewer than 10 times each. 
| Alias                     | Number of Occurrences                 |
|---------------------------|---------------------------------------|
| `u3at(axe, som)`          | `2`                                   |
| `u3nl`                    | `2`                                   |
| `u3du(som)`               | `113`                                 |
| `u3ud(som)`               | `245`                                 |
| `u3r_du(som)`             | `7`                                   |
| `u3r_ud(som)`             | `2`                                   |
| `u3do(txt_c, arg)`        | `19`                                  |
| `u3dc(txt_c, a, b)`       | `47`                                  |
| `u3dt(txt_c, a, b, c)`    | `2`                                   |
| `u3dq(txt_c, a, b, c, d)` | `1`                                   |

mrdomino:
fwiw I find the `u3do` family of macros useful and intuitive — they reuse the `c`/`t`/`q` pattern from `u3nc` etc. The rest I don’t have strong feelings about.

mrdomino:
Oh — if it’s just renaming to `u3v_dc` etc, I’m fine with that.

joemfb (changes requested):
Based merely on the occurrence counts, this PR will be much easier to merge if you drop `u3du()` and `u3ud()` from it. Everyone (but me) seems to have independently arrived at the same mnemonic, so I'm also inclined to retract my earlier criticism of those macros.

belisarius222:
I'd be ok with removing a few of these, but things like u3k() need short aliases.

u3du() and u3ud() seem easy enough to make them a footgun, so picking names that aren't a trap for dyslexics might be a good idea.

This whole exercise also runs the risk of sucking in a lot of time for marginal and non-semantic gains, so it would be good to resolve it quickly.

—
~rovnys-ricfer
https://urbit.org

belisarius222:
*easy enough to mix up

—
~rovnys-ricfer
https://urbit.org

mcevoypeter:
Thanks @joemfb and @belisarius222 for the feedback. Based on your thoughts, I've done the following:
I renamed the following aliases:
- `u3do` -> `u3v_do`
- `u3dc` -> `u3v_dc`
- `u3dq` -> `u3v_dq`
- `u3du` -> `u3ic` (short for "is cell")
- `u3ud` -> `u3ia` (short for "is atom")
and removed the following aliases:
- `u3at`
- `u3nl`
- `u3r_du`
- `u3r_ud`

mcevoypeter:
Also note that I decided to recreate the branch off the tip of `peter/all-header`--which removed `noun/aliases.h`--and force push.

mrdomino:
@belisarius222 and i mentioned `u3i[ac]` in person, the only problem noted was the semi-overlap with the `imprison` namespace. i suggest `u3ua` / `u3uc` or — perversely, at cost to existing maintainers but benefit to newcomers, `u3da` / `u3dc`. the latter has the mnemonic `u3 describes {atom, cell}`.

mcevoypeter:
> @belisarius222 and i mentioned `u3i[ac]` in person, the only problem noted was the semi-overlap with the `imprison` namespace. i suggest `u3ua` / `u3uc` or — perversely, at cost to existing maintainers but benefit to newcomers, `u3da` / `u3dc`. the latter has the mnemonic `u3 describes {atom, cell}`.

Using this definition of definition of namespace conflict, `u3ud` conflicts with the `u3u_` (`noun/urth.c`) namespace. I've been considering the namespace of a module to include the underscore immediately following `u3[a-z]`; anything without an `_` in its name belongs to the global namespace. So `u3ia` and `u3ic` don't feel problematic, at least to me.

mrdomino:
hmm ok. i think `u3i` is more commonly used than `u3u`, so the semi-conflict is a bit more pronounced. but it's nonetheless only a semi-conflict; not a formal conflict, just a slightly annoying overlap.

at the end of the day i wouldn't object to `u3ia` / `u3ic`, though i do think `u3da` / `u3dc` are probably superior.

mrdomino:
the cost might actually be a benefit — the handle `u3dc` is already familiar in people’s minds, and the usage patterns are distinct enough (1 arg vs 2 where arg 1 is ~always a string literal) that there's little danger of ambiguity. seems like it'd reuse some wrinkles in people's brains.

mcevoypeter:
Remapping an existing name--an annoying overlap in its own right--sounds more difficult for people than simply introducing a new name with a slightly easier to remember mnemonic. We're probably thinking too hard about this though.

joemfb:
I'd like to reiterate my request to leave `u3du()/u3ud()` out of this PR. It's not controversial without those changes, and clearly controversial with them. It's also harder to merge, as it will create more conflicts due to various pieces of outstanding work.

Also, since you've rebased onto it, can you target this PR against `peter/all-header`?

joemfb (approved):
LGTM

PR #5509, opened 2021-12-30
zuse: add schnorr signatures

mrdomino:
Jets here: #5510.


joemfb (commented):
This should have some unit tests -- there are official vectors here: https://github.com/bitcoin/bips/tree/master/bip-0340.

I'm not familiar with the domain, but would it be reasonable to refactor this into a generic implementation, parameterized on hash functions, lengths, tag constants, &c?

mrdomino:
I’ll add the test vectors; I tested against them manually from dojo but obviously we want them in the repo.

I’ll ask about the generic implementation as well.

mrdomino:
Re generic implementation: Tim says if it’s good enough for Bitcoin to add in a major update, it’s enough of an update for us for now.

(The C side’s [`secp256k1_schnorrsig_sign`](https://github.com/bitcoin-core/secp256k1/blob/master/src/modules/schnorrsig/main_impl.h#L195) only implements the default parameters; other stuff is supported only in a separate `secp256k1_schnorrsig_sign_custom`. Seems reasonable to take a similar approach with the hoon if in future anyone ever wants to parameterize further.)

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,107 @@
         =/  pub  (from.j qj)
         ?<  =([0 0] pub)
         pub
+      ++  schnorr
+        ~%  %schnorr  ..schnorr  ~
+        =>  |%  ++  tagged-hash
```
joemfb, line 2240: Style nit, this indentation should be:

```
=>  |%
    ++  tagged-hash
```

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,107 @@
...
+                  |=  [tag=@ [l=@ x=@]]
+                  =+  hat=(sha-256:sha (swp 3 tag))
```
joemfb, line 2242: The levels of swapping here are frustrating. `+sha-256:sha` swaps it's input and output, so this swaps thrice. It's equivalent to `(rev 3 32 (shax tag))`, but idk if it's worth changing. Just frustrating ...

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,107 @@
...
+            ?:  =(0 (mod y.pp 2))
+              sk
+            (sub n.domain.c sk)
+          =/  t
+            %+  mix  d
+            (tagged-hash 'BIP0340/aux' [32 a])
+          =/  rand
+            %+  tagged-hash  'BIP0340/nonce'
+            :-  96
+            (can 8 ~[[1 m] [1 x.pp] [1 t]])
```
joemfb, line 2284: This is the same as `(rep 8 [m x.pp t ~])`.

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,107 @@
...
+          =/  k
+            ?:  =(0 (mod y.rr 2))
+              kp
+            (sub n.domain.c kp)
+          =/  e
+            %-  mod
+            :_  n.domain.c
+            %+  tagged-hash  'BIP0340/challenge'
+            :-  96
+            (can 8 ~[[1 m] [1 x.pp] [1 x.rr]])
```
joemfb, line 2296: Also `(rep 8 ...)`

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,107 @@
...
+          ?>  (gte 64 (met 3 sig))
+          =/  c  curve
+          =/  pup  (lift-x pk)
+          ?~  pup
+            %.n
+          =/  pp  u.pup
+          =/  r  (cut 8 [1 1] sig)
+          ?:  (gte r p.domain.c)
+            %.n
+          =/  s  (cut 8 [0 1] sig)
```
joemfb, line 2319: This is `(end 8 sig)`

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,107 @@
...
+          ?:  (gte s n.domain.c)
+            %.n
+          =/  e
+            %-  mod
+            :_  n.domain.c
+            %+  tagged-hash  'BIP0340/challenge'
+            :-  96
+            (can 8 ~[[1 m] [1 x.pp] [1 r]])
```
joemfb, line 2327: Also `(rep 8 ....)`

pkg/arvo/tests/sys/zuse/crypto/secp256k1.hoon
```diff
@@ -116,4 +116,237 @@
...
+      ==
+    ==
+  :_  .
+  ^=  t3
+    %+  turn  cases-pub
+    |=  case-pub
+    ^-  tang
+    %+  expect-eq
+      !>  res
+    !>  (verify:schnorr:ecc pub mes sig)
```
joemfb, line 318: In a departure from standard, 3-fixed backstep indentation, test-case vase literals are usually both indented:

```
%+  expect-eq
  !>  expected
  !>  actual
```

joemfb (changes requested):
I've noted some style nits and simpler atom ops. I've glanced at the spec, but haven't closely reviewed for edge cases.

pkg/arvo/sys/zuse.hoon
mrdomino: Yep. Root cause is that BIP 340 specifies things the wrong endianness so something is gonna be backwards no matter what. I opted for consistency with the `+sha` usage below.

joemfb (approved):
Changes look good.

Please get another review for spec sanity check, and hold off on merging 'till the jet is sorted out.

pkg/arvo/sys/zuse.hoon
```diff
@@ -2265,7 +2265,6 @@
           ~/  %sosi
           |=  [sk=@I m=@I a=@I]
           ^-  @J
-          ?>  (gte 32 (met 3 sk))
```
joemfb: It's important to be straightforward to verify that assertions/crashes in the jet match the hoon/nock. Can you add a comment to the source in addition to the commit msg that removed this?

tacryt-socryp:
I have manually reviewed every value in every test vector, and determined that they are identical to those found in the BIP.

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,109 @@
...
+              sk
+            (sub n.domain.c sk)
+          =/  t
+            %+  mix  d
+            (tagged-hash 'BIP0340/aux' [32 a])
+          =/  rand
+            %+  tagged-hash  'BIP0340/nonce'
+            :-  96
+            (rep 8 ~[m x.pp t])
+          =/  kp  (mod rand n.domain.c)
```
tacryt-socryp, line 2287: directly after 2287, insert an assertion that kp must not be 0

pkg/arvo/sys/zuse.hoon
```diff
@@ -2235,6 +2235,109 @@
...
+            ++  tagged-hash
+              |=  [tag=@ [l=@ x=@]]
+              =+  hat=(sha-256:sha (swp 3 tag))
+              %-  sha-256l:sha
+              :-  (add 64 l)
+              (can 3 ~[[l x] [32 hat] [32 hat]])
+            ++  lift-x
+              |=  x=@I
+              ^-  (unit point)
+              =/  c  curve
```
tacryt-socryp, line 2250: Must assert that input x is in the range 0...p-1

tacryt-socryp:
Aside from the comments that I read, I went over the pseudocode of every line alongside the Hoon and verified that it is all represented here, even the pesky "is infinite check" in point addition which is mirrored here: https://github.com/bitcoin/bips/blob/97e02b2223b21753acefa813a4e59dbb6e849e77/bip-0340/reference.py#L45

tacryt-socryp (approved):

mrdomino:
added assertions and verified that unit tests still pass with jets disabled

PR #5510, opened 2021-12-30
vere: schnorr signature jets

mrdomino:
Arvo side: #5509.


mrdomino:
CI failure looks unrelated.

joemfb:
@frodwith PTAL

pkg/urbit/jets/e/secp.c
```diff
@@ -208,3 +208,90 @@ u3we_make(u3_noun cor)
...
+/* verify a schnorr signature
+*/
+static u3_atom
+_cqes_sove(u3_atom pk, u3_atom m, u3_atom sig)
+{
+  c3_y pub_y[32];
+  c3_y mes_y[32];
+  c3_y sig_y[64];
+
+  if ( (0 != u3r_bytes_fit(32, pub_y, pk)) ||
```
frodwith, line 267: In a lot of our crypto, the hoon-as-written can handle larger arguments - sometimes by truncating, sometimes by just returning a different answer. Returning `u3_none` from the jet would let the hoon cover that case, but that's not ideal for crypto because of timing attacks. Since this is new hoon code, it should explicitly assert the widths of these arguments at the top of the gate, with a series of i.e.
    ?>  (gte 32 (met 0 pk))

It might be that the hoon already exits when given mis-sized arguments (worth including tests for). In that case, it would still be clarifying to add the asserts.

pkg/urbit/jets/e/secp.c
```diff
@@ -208,3 +208,90 @@ u3we_make(u3_noun cor)
...
+/* create a schnorr signature
+*/
+static u3_weak
+_cqes_sosi(u3_atom sk, u3_atom m, u3_atom a)
+{
+  c3_y key_y[32];
+  c3_y mes_y[32];
+  c3_y aux_y[32];
+
+  if ( (0 != u3r_bytes_fit(32, key_y, sk)) ||
```
frodwith, line 221: see my comment below in _cqes_sove

frodwith: Maybe you have tests and I'm missing them, but I would have expected to see them in pkg/arvo/tests/sys/zuse/crypto/secp256k1.hoon? Just a few cases (from an RFC if possible) is fine, please try things that are zero and things that are outside the size range for the inputs, too. The mis-sized values should crash, and the tests should pass with the jets enabled and disabled.

pkg/urbit/jets/e/secp.c
mrdomino: Added boundary assertions and test cases.

pkg/urbit/jets/e/secp.c
mrdomino: Added test cases from BIP-340; I had tested against those manually before.

mrdomino:
Test cases on #5509 pass for me locally with and without jets enabled.

mrdomino:
@frodwith mind taking another look? Relevant changes added on the other branch.

mrdomino:
The tests still pass with #5581 and equivalence testing turned on. @joemfb PTAL

mrdomino:
Given that #5581 is a wontmerge, we need some way to test this. Options seem to be:

1. You read [the BIP](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki), the Hoon, and the C, and become confident that the Hoon implements the BIP and the C jets the Hoon.
2. I do some extensive local fuzz testing with randomized inputs and make sure the jet seems to match the hoon on all of them.
3. We add a flag to `u3j_harm` that does more robust non-recursion-aware livenet equivalence testing.
4. We fix #5580 the right way, which sounds like it’s to virtualize the jetted and non-jetted calls separately.

Option 4 is clearly the best long-term option. 3 is probably the fastest short-term option. Depending on when we want the next release to go out, I recommend one or the other of them.

mrdomino:
I thought it might be trivial to descend into a virtualization context for the jet call, but it seems like I don’t understand what’s going on — `u3m_soft_run` takes a `u3_funq`, not a `u3_funk`, and I lack enough knowledge of the system to confidently answer even the question of whether we ever expect jet code to scry. (Seems like no? But I can't say for sure.)

So I propose either 1 or 3 depending on your availability @joemfb, unless someone wants to hold my hand through 4.

pkg/urbit/jets/tree.c
```diff
@@ -403,6 +403,24 @@ static u3j_harm _140_hex_secp_secp256k1_sign_a[] = {{".2", u3we_sign, c3y}, {}};
 static c3_c* _140_hex_secp_secp256k1_reco_ha[] = { 0 };
 static u3j_harm _140_hex_secp_secp256k1_reco_a[] = {{".2", u3we_reco, c3y}, {}};
 
+static c3_c* _140_hex_secp_secp256k1_schnorr_sosi_ha[] = { 0 };
+static u3j_harm _140_hex_secp_secp256k1_schnorr_sosi_a[] =
+  {{".2", u3we_sosi, c3n, c3n}, {}};
+static c3_c* _140_hex_secp_secp256k1_schnorr_sove_ha[] = { 0 };
+static u3j_harm _140_hex_secp_secp256k1_schnorr_sove_a[] =
+  {{".2", u3we_sove, c3n}, {}};
```
joemfb, line 411: As discussed elsewhere, please remove these testing flags.

pkg/urbit/jets/e/secp.c
```diff
@@ -208,3 +208,90 @@ u3we_make(u3_noun cor)
...
+  c3_y sig_y[64];
+
+  if ( (0 != u3r_bytes_fit(32, pub_y, pk)) ||
+       (0 != u3r_bytes_fit(32, mes_y, m)) ||
+       (0 != u3r_bytes_fit(64, sig_y, sig)) )
+  {
+    return u3m_bail(c3__exit);
+  }
+  else {
+    return __(urcrypt_secp_schnorr_veri(sec_u, sig_y, mes_y, pub_y));
```
joemfb, line 274: This return value likely includes non-deterministic results, at least allocation failure and invalid context (`sec_u`). I'm not sure if there's enough information in the secp api to distinguish deterministic verification failure from other cases. Ideally there is, and the urcrypt api can reflect it. If there is not, you'll have to "punt" (return `u3_none` and defer to nock).

joemfb (changes requested):
This all looks reasonable. Crypto jets are unusual, in that we basically matching the hoon to a reference impl jet. Normally we're doing the opposite.

You'll need to drop the testing flags from this PR, and sort out the failure/error-handling case:

pkg/urbit/jets/e/secp.c
mrdomino: Looks like nondeterministic failures are done via `abort()` in secp256k1. This should only return deterministic results unless I'm mistaken.

joemfb: What is the `illegal_callback`? https://github.com/bitcoin-core/secp256k1/blob/c8aa516b57ab951e26a1ff5fa283d6f5612ca809/src/secp256k1.c#L34-L39

mrdomino: Winds up getting assigned to [`secp256k1_default_illegal_callback_fn`](https://github.com/bitcoin-core/secp256k1/blob/c8aa516b57ab951e26a1ff5fa283d6f5612ca809/src/util.h#L29).

joemfb: And the `VERIFY_CHECK`s are compiled out by default?

mrdomino: correct

joemfb (approved):
LGTM

PR #5515, opened 2022-01-03
Dane.bettis/stacktrace hints

DaneBettis:
vere: add %nara %hela hints, add a hoon test module for hints

nara: swedish for near, provides a trace from the start of the nearest virtualization to the statement wrapped by the nara hint.

hela: swedish for whole, provides a trace from the start of the event  to the statement wrapped by the hela hint.

The eventual goal is to present these traces in a way that is most salient & useful to developers and in a way that is aesthetically pleasing as the bout hint is useful and nice looking.

For this pull request the goal is to provide the actual functionality in code that we can feel safe and comfortable adding to the master branch. To this end, I've added a trivial set of tests for nara and hela in pkg/arvo/tests/run/hints.hoon - they only prove that invoking these hints will not crash the runtime, but I'm open to ideas for how I can test these hints further.


DaneBettis:
I have spoken with ~ritpub-sipsyl who tipped me off about the Racket debugging tools, esp DrRacket and its debugger. As I have dug into this, I've also seen the output of its tracer, and I think from an aesthetic standpoint these tools are at least one very good inspiration for how we can format and present data from our stacktrace hints.

pkg/urbit/noun/nock.c
```diff
@@ -1742,13 +1752,46 @@ _n_hilt_hind(u3_noun tok, u3_noun pro)
...
+    } break;
+
+    case c3__hela: {
+      u3_noun pri, cap;
+      c3_assert(c3y == u3r_cell(*clu, &pri, &cap));
+      c3_l pri_l = c3y == u3a_is_cat(pri) ? pri : 0;
+      u3t_slog_cap(pri_l, u3i_string("trace of"), u3k(cap));
+      u3t_slog_hela(pri_l);
+      u3z(*clu);
+      *clu = u3_nul;
```
joemfb, line 1788: The assertion isn't appropriate here, since `*clu` contains arbitrary, user/dev controlled values. So it should be replaced with `if ( c3y == u3r_cell ...`. And `cap` is the wrong name for the tail -- `u3i_string("trace of")` is the caption). (both of these are my mistakes ...) 

The same goes for `%nara` above, *mutatis mutandis*.

pkg/urbit/noun/nock.c
```diff
@@ -1700,6 +1702,14 @@ _n_hilt_fore(u3_noun hin, u3_noun bus, u3_noun* out)
...
     *out = u3i_cell(u3h(hin), now);
   }
+  else if ( c3__nara == u3h(hin) ) {
+    u3t_slog_nara(0);
+    *out = u3_nul;
+  }
+  else if ( c3__hela == u3h(hin) ) {
+    u3t_slog_hela(0);
+    *out = u3_nul;
+  }
```
joemfb, line 1712: With the two new hints here, I think it's worth converting this to a `switch` statement, as we did in `_n_hint_fore()`.

joemfb (changes requested):
This looks good. I've got a couple minor tweaks:

joemfb (changes requested):

joemfb (approved):
LGTM

belisarius222:
These look useful.  Can I see some examples on how to use them?

joemfb:
@belisarius222 the unit tests in this PR have examples

Issue #5516, opened 2022-01-03
|rekey doesn't work with new multikey format


drbeefsupreme:
It expects `seed:jael` but it needs to be switched to `feed:jael`.

Since I already spent some time looking at this before the break, I'll take a go at fixing this this week.

drbeefsupreme:
Closed by #5522

PR #5521, opened 2022-01-05
Fix |rekey to work with multikey format

drbeefsupreme:
Addresses #5516

Should be backwards compatible with the old key format as well.

I think the style could be improved somewhat but I'm not sure how. The following snippet appears twice:

```
    ?.  =(our.bowl who.fed)
      ~&  [%wrong-private-key-ship who.fed]
```

The issue is that until the compiler checks whether it has a `feed:jael` or `seed:jael` with `?@  -.fed`, it cannot access `who.fed`. So I seem to need to do the check inside of both branches of `?@`, which looks ugly and feels wrong. What's a better way to do this?


drbeefsupreme:
Hmm, I think I did this wrong. Shouldn't have all those other commits. I will retry.

philipcmonk:
Looks like this was branched off release/next-userspace, which has those janeway merge commits that we don't want to hit master.

You should checkout master, cherry-pick 739f376 onto a new branch, then PR that against next/userspace.

drbeefsupreme:
New PR #5522

PR #5523, opened 2022-01-05
landscape: serviceworker eventstream multiplexing

arthyn:
This PR adds a base level serviceworker at the root of your Urbit domain through grid. That serviceworker then has the capability to intercept new event stream connections and reuse an existing connection if one already exists. This adds some new options to @urbit/http-api and also enables this behavior by default.

For those apps using their own serviceworkers who want to partake in this behavior, a small proxy utility was added to @urbit/http-api to allow similar behavior since their serviceworker will take priority over the base. It just uses a simple call to setup the listeners needed to proxy the same requests back to the existing base serviceworker.

This still needs to be tested in a deployed environment and we need to make sure that should something catastrophic happen, event streams just revert to the default behavior and don't get caught in limbo. However, from what I could tell in local testing it seems that this may already be handled.


zalberico:
@arthyn is this still relevant/is it something we could merge and test for release?

arthyn:
the problem is still relevant, but I deemed this too experimental to release. we should just rely upon http2 increase in active connections or find some other solution.

PR #5526, opened 2022-01-06
[WIP] clay: unify request & response types

liam-fitzgerald:
Refactors clay to use unified request and response types

Open questions:
- Probably need to pair on making this work over the network
- Backward compatibility locally and over the network. This has been designed so that local backwards compat is *possible*, although not implemented. Network backward compatibility is probably necessary to OTA this properly, right?
- There's a `$riff-any`, but no versioned response type?
- TODO: state adapters


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -412,12 +425,7 @@
 ++  print-wove
   |=  =wove
   :-  for.wove
-  ?-  -.rove.wove
-    %sing  [%sing mood.rove.wove]
-    %next  [%next [mood aeon]:rove.wove]
-    %mult  [%mult [mool aeon]:rove.wove]
-    %many  [%many [track moat]:rove.wove]
-  ==
+  ~
```
philipcmonk, line 428: +print-wove is constantly useful for `~&`s because requests tend to be too big to print.  I never leave home without it.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1142,9 +1150,10 @@
   ++  aver
     |=  [for=(unit ship) mun=mood]
     ^-  [(unit (unit cage)) ford-cache]
-    =+  ezy=?~(ref ~ (~(get by haw.u.ref) mun))
-    ?^  ezy
-      [`u.ezy fod.dom.red]
+    ::  XX: wat do
```
philipcmonk, line 1153: You are going to want to look up the request in `haw`.  IIRC this is how non-full-sync foreign requests are filled.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1575,12 +1555,13 @@
         |-  ^+  ..park
         ?~  wux  ..park
         =/  rov  rove.wove.i.wux
-        =?    qyx
-            ?+  -.rov  |
-              %sing  ?=([%a * %app %publish %hoon ~] mood.rov)
-              %next  ?=([%a * %app %publish %hoon ~] mood.rov)
-            ==
-          (~(del by qyx) wove.i.wux)
+        ::  XX: revive
```
philipcmonk, line 1558: OTOH, the upgrade's already happened and has to be done stepwise anyway, so we can probably just get rid of the whole +migrate-dist.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1307,21 +1333,18 @@
     ?~  ref
       (run-if-future rove.wov |=(@da (bait hen +<)))
     |-  ^+  +>+.$
-    =/  =rave  (rove-to-rave rove.wov)
-    =.  rave
-      ?.  ?=([%sing %v *] rave)  rave
-      [%many %| [%ud let.dom] case.mood.rave path.mood.rave]
```
philipcmonk, line 1313: This transmutation is load-bearing and matches with the TODO in +apply-foreign-update.  Worth paring on this, but the gist is that `%sing %v` ends up producing a response `+dome:clay` (similar to the +dome defined in clay.hoon but without some of the caches).  But if that request goes over the network, it actually gets a +nako over the network instead of a +dome.

You can think of a +nako as "+dome without even the ankh cache because that's too big, plus the commits referred to in the dome", and when you receive one you need to scan the commits for unknown blobs and request those (the +foreign-update state machine, starting from +apex).

I suspect we can do this without a transmute-to-%many step by simply recognizing at the edges that we're about to send a %v request over the wire (so convert +dome to +nako) or being received over the wire (so convert +dome to +nako and fetch blobs).  Would be much simpler.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3343,256 +3331,88 @@
...
-        ::  we don't have the data directly.  how can we fetch it?
-        ::
-        ?:  =(0 u.aeon)
-          ~&  [%clay-sing-indirect-data-0 `path`[syd '0' path.mood.rov]]
-          [[~ ~] fod.dom]
-        ~&  [%clay-sing-indirect-data desk=syd mood=mood.rov aeon=u.aeon]
-        [[`rov ~] fod.dom]
-      ::  we have the data, so produce the results
-      ::
+    =/  cache-value=(unit (map [care path] (unit cage)))
```
philipcmonk, line 3334: incredibly powerful diff, this whole function

philipcmonk (commented):
Down to pair on this, for sure.  Find a time on my calendar after 9am PT.

> There's a `$riff-any`, but no versioned response type?

The previous upgrade made %many requests change from producing `[(map aeon tako) aeon (set yaki) (set blob)]` (ie +nako) to `[(map aeon tako) aeon (set yaki) ~]`, which nests.  If you requested with the new version, you wouldn't get the blobs because you know how to request them individually.

That upgrade strategy works for some types of upgrades, potentially including this one.  If you made a `%rage` request you know how to handle a `%wriz`, so you don't actually need a number on it.  Could be reasonable to put one inside the `%wriz` though.  Then we wouldn't need to make up more `%wri` words going forward.

Notably, for network compatibility we don't need "4-way compatibility".  Old client's request to a new server needs to receive an old response, but a new client's request to an old server can just crash until the server gets up to date.

zalberico:
@liam-fitzgerald @philipcmonk - is this work still relevant/something we want to merge?

philipcmonk:
In general yes, it's a really good cleanup.  It is not complete, and it's not clear when it will become a priority.

PR #5527, opened 2022-01-07
u3: consolidate functions in events.c

mcevoypeter:
In preparation for iterating over the north and south segments of the snapshot in the same direction as mentioned [here](https://github.com/urbit/urbit/pull/5477#discussion_r779236258) in #5477, clean up the functions in [`events.c`](https://github.com/urbit/urbit/tree/peter/consolidate-event-functions/pkg/urbit/noun/events.c) by;
- deleting functions that are not called from anywhere in the system (`u3e_check()`, `_ce_check_page()`, `_ce_maplloc()`, `_ce_mapfree()`);
- deleting short static functions in [`events.c`](https://github.com/urbit/urbit/tree/peter/consolidate-event-functions/pkg/urbit/noun/events.c) with only a single call site (`_ce_patch_write_page()`, `_ce_patch_create()`, `_ce_patch_read_control()`, `_ce_patch_write_control()`) and instead inling the functionality at the call site;
- and folding the functionality of `_ce_patch_free()` into `_ce_patch_delete()`.

### DIFF
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -15,17 +15,6 @@ struct {
   c3_w mug_w[u3a_pages];
 } u3K;
 
-/* _ce_check_page(): checksum page.
-*/
-static c3_w
-_ce_check_page(c3_w pag_w)
-{
-  c3_w* mem_w = u3_Loom + (pag_w << u3a_page);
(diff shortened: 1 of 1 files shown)


mcevoypeter:
Closing without merging because this change has been folded into the scope of event log truncation. The new PR within the scope of the event log truncation effort can be found at #5546.

Issue #5507, opened 2021-12-28
Cannot select or copy text in webterm


litmus-ritten:
**Description**
Text in the webterm interface cannot be selected, with the consequence that useful webterm output (+vats, moon keys, etc.) cannot be copied.

**To Reproduce**
- Open webterm.
- Wait for activity or type a command which elicits some text output.
- Observe that the text cannot be selected.

**Expected behaviour**
Webterm should be selectable.

**System:**
 - OS: raspbian arm64
 - Urbit version: 1.8 C-king
 - %webterm hash: `6jq91`
 - %base hash: `9m7kg`
 - Browser: Vivaldi 5.0.2497.32 64-bit (Chromium-based)

**Additional context**
~tinnus-napbus reports analogous issues on Port (Chromium-based) and Qutebrowser (Webkit-based).
~nocsyx-lassul reports: *"right clicking the text seems to highlight it, but [...] normal highlight doesn't seem to work"* and *"when I right click a moonkey it seems to highlight the whole key and gives me the option to copy"* using Port.
I cannot select or copy a moonkey under any circumstances.

Thank you to tinnus and nocsyx for their additional input.

Fang-:
You can hold down the alt key, text selection will work as normal then. There's some tweaks to be made there, primarily how undiscoverable this is, but should help you out for the time being.

See also https://github.com/urbit/landscape/issues/1270#issuecomment-994152460.

litmus-ritten:
Ah, I found that alt did not work, however I could shift-select text in the webterm, which represents a workable stopgap. Nevertheless the selection feedback is virtually invisible - changing the text to a minutely different shade in both the case of grey and black text:

![Screenshot from 2021-12-30 01-54-31](https://user-images.githubusercontent.com/62781197/147674855-f319dbbe-45ed-41b9-96c6-dde5ff0b6fdd.png)

Here, the words 'dozing' → 'pintem' are highlighted.

Thanks for the assist.

PR #5529, opened 2022-01-07
webterm: improve selection look & feel

Fang-:
Specifies a selection highlight color to make sure it's legible within the active theme. Fixes the visual aspect of urbit/landscape#1270 and #5507, but the selection functionality itself is still hard to discover.

Also disables some xterm.js behavior that would cause alt-clicking to insert arrow keys, with the intent to move the cursor to that location. We (will fully) support clicks natively, so there's no need for this.

Selection highlight color is just whatever the current foreground color is, but notably still gets rendered at the usual 50(?)% opacity.

image

image


urcades (approved):
so excited for this

Issue #5518, opened 2022-01-04
/app/azimuth doesn't have its naive state up to date


yosoyubik:
This has been seen in a couple of ships (moons and stars so far). Its more direct consequence is that a newly L2 ship booted with the "old" private key (i.e. not going through the invite-acceptance flow) will have connectivity issues with these ships until their naive state is again up to date.

A couple of different versions of this issue:

- A ship has a few thousand of points in `points.nas`—this could be caused by the issue fixed here https://github.com/urbit/urbit/pull/5434, and later pushed also to a new pill here https://github.com/urbit/urbit/pull/5503, so some of the ships (`~sordem`) that are experiencing this probably booted before the updated pill was pushed.

- Some ships (confirmed on `~samzod` but possibly others) did not apply the L2 transaction that updated `~sordem`'s spawn-proxy, as seen here:
<img width="899" alt="Screen Shot 2022-01-04 at 1 50 35 PM" src="https://user-images.githubusercontent.com/1033194/148087628-7481bcee-2976-4565-9ac8-219c75508d85.png">

<img width="940" alt="Screen Shot 2022-01-04 at 5 02 59 PM" src="https://user-images.githubusercontent.com/1033194/148087623-d1d250df-add1-48f1-923e-2cf65f24fdef.png">

- This causes any subsequent L2 spawn from `~sordem` to fail on /app/azimuth, which would cause communication issues with that ship, since it doesn't have their keys.

- A solution to both problems could be to push an OTA that does a rerun of the logs on-load, since, at least the mentioned ships that have problems seem to have all eth-logs up to date—this could be a good moment to include this PR https://github.com/urbit/urbit/pull/5496 that replaces the eth-logs snapshot with the azimuth-state snapshot

- The remaining question here is why a specific L2 transaction was not applied, which would need to also be addressed because that can certainly happen again.

  - A possible explanation for this could be the issue experienced when the L2 batch that contained the transaction that updated `~sordem`'s spawn-proxy was sent out, in which Tlon's eth-gateway had not whitelisted a specific method used by eth-watcher, as seen here on `~samzod`:

<img width="1436" alt="Screen Shot 2022-01-04 at 1 47 04 PM 2" src="https://user-images.githubusercontent.com/1033194/148093395-c2356670-598a-4422-9b49-39eae04bc55c.png">

 - There seems to be a more general error with `http://eth-mainnet.urbit.org:8545` as these errors  has shown up in some ships:
[...]

Issue #5517, opened 2022-01-04
L2 planets has connectivity issues if booting while acceptance txs are confirmed on L1


yosoyubik:
Booting an L2 planet with keys produced after going through the invite-acceptance flow (i.e. multi-key file with old and new keys) during the time when the L2 batch that contains the transaction with the new keys is confirmed on L1, and presuming that all ships update their state with the new keys before the L2 planet itself knows that it can start using the new keys, results in the planet having connectivity issues.

- Looking at the planet's keys, it has the second private key that was used when booting (as seen running `;;(seed:jael (cue .^(@ %j /=vile=)))` with life=2
- Pubkey at life=2 (`.^([life @uw (unit @ux)] %j /=deed=/(scot %p our)/2)`) matches what peers (`.^(ship-state:ames %ax /=//=/peers/~doltyc-tobfyl)`) expect
- Number of eth-logs received by %azimuth are also up to date
- Ames log:

```
ames: ~zod: plea [[~doltyc-tobfyl 1] [~zod 6] bone=[8 %g /ge/hood]]
> |hi ~zod
>=
ames: ~zod: timeout update [rto=120.000 rtt=1.000 rttvar=1.000 ssthresh=1 cwnd=1 num-live=1 counter=0]
ames: ~zod: dead [[1 0] rto=120.000 rtt=1.000 rttvar=1.000 ssthresh=1 cwnd=1 num-live=1 counter=0]
ames: ~zod: send-blob: to ~zod
ames: ~zod: trying route: ~zod
```

<details>
<summary>- Boot log:</summary>

```
❯ urbit -w doltyc-tobfyl -k doltyc-tobfyl-1.key
~
urbit 1.8
boot: home is ~/doltyc-tobfyl
loom: mapped 2048MB
lite: arvo formula 27e494c5
lite: core 7b144622
lite: final state 7b144622
boot: downloading pill https://bootstrap.urbit.org/urbit-v1.8.pill
boot: retrieving ~doltyc-tobfyl's public keys
boot: verifying keys
boot: getting sponsor
boot: retrieving galaxy table
boot: retrieving network domains
boot: retrieving keys for sponsor ~sordem
boot: retrieving keys for sponsor ~dem
boot: parsing %brass pill
loom: mapped 2048MB
boot: protected loom
live: logical boot
boot: installed 348 jets
---------------- playback starting ----------------
pier: replaying events 1-19
1-b
1-c (compiling compiler, wait a few minutes)
ride: parsing
[...]

PR #5496, opened 2021-12-15
azimuth: update /app/azimuth snapshot

yosoyubik:
This PR is built on top the other roller/azimuth PRs (#5467 #5458 #5487) and is currently deployed to the mainnet L2 roller.

The contents of the Azimuth snapshot (also included in the solid pill) up to the latest L2 batches:
```
points: 77.447
sponsors: 460
owners: 19.132
block-number: 13.807.777
block-hash: 0x3cb6.ecac.64b7.bd3d.8155.cd4b.d741.f1b1.387f.13aa.aa27.2e17.cc71.c22a.5c90.6264
```

PR #5532, opened 2022-01-08
jael: give new private key to subscribers on %keys new-event

yosoyubik:
This is built on top of https://github.com/urbit/urbit/pull/5496 using the latest snapshot. A solid pill is also included—a new pill should be pushed to `bootstrap.urbit.org` before any L2 ships are booted, otherwise they might end up in a situation where the eth-log that includes the new private key comes before the OTA that has this fix, which will make them disconnected from the network, before the fix for https://github.com/urbit/urbit/issues/5517 is applied, and a manual intervention would be needed.

This also (from https://github.com/urbit/urbit/pull/5496) switches /app/azimuth to use a snapshot of the Azimuth state, instead of relying on eth-logs, which will be applied on-load, and will bring all ships of the network up to date, fixing the incomplete state found in https://github.com/urbit/urbit/issues/5518

Couple of tests performed to validate that this works:

- Boot from a pill that contains this fix, after going through the invite flow on Bridge
- Boot from the current pill, and downloading an OTA that contains this fix.
- Applied this PR to ships that lost connectivity after switching to the new private key, when the L2 batch that contained the %keys event was processed by /app/azimuth

This PR has been tested and deployed to ~roller-dozzod-dozzod and ~sordem (both online) and several L2 ships.


yosoyubik:
I also booted a comet from the current pill and `|OTA`ed from `~sordem` (that has this PR deployed) and it got merged successfully—I could `|hi` several ships, both L1 and L2

yosoyubik:
I missed this looking at the backlog, but noticed this after receiving the OTA:

```
ping: strange state [~marzod ~ u=[rift=4 ship-state=[%waiting until=~2022.1.10..15.10.01..15bd]]]
ping: strange state [~zod ~ u=[rift=2 ship-state=[%waiting until=~2022.1.10..15.10.01..1ab9]]]
```

pkg/arvo/sys/vane/jael.hoon
```diff
@@ -733,15 +740,18 @@
...
           ==
-        life.to.u.a-diff
-      (public-keys:feel original-pos %diff ship.i.udiffs u.a-diff)
+        this-su
+      ::  if this about our keys, and we already know these, start using them
+      ::
+      =.  lyf.own  life.to.u.a-diff
+      ::  notify subscribers (ames) to start using our new private keys
+      ::
+      (exec yen.own [%give %private-keys [lyf jaw]:own])
```
philipcmonk, line 754: Does this work?  I think so, but the pattern feels weird since it puts two conflicting copies of `this-su` in our subject, like as though it needs a `public-keys:feel:-` to make sure we're using the right one.

I'd probably do something like:

```
      =?    this-su
          ?&  =(our ship.i.udiffs)
              ?=(%keys -.u.a-diff)
              (~(has by jaw.own) life.to.u.a-diff)
          ==
        ::  if this about our keys, and we already know these, start using them
        ::
        =.  lyf.own  life.to.u.a-diff
        ::  notify subscribers (ames) to start using our new private keys
        ::
        (exec yen.own [%give %private-keys [lyf jaw]:own])
      ::
      (public-keys:feel original-pos %diff ship.i.udiffs u.a-diff)
```

philipcmonk (approved):
This looks like it should work

pkg/arvo/sys/vane/jael.hoon
yosoyubik: Yeah, I tested it with a couple of L2 ships, but agree that your suggestion looks better—I deployed it https://github.com/urbit/urbit/pull/5532/commits/432d9674f5cb780f1de2a33f88e3b07404b632b4 on `~sordem` and also booted a L2 ship from a the pill and checking that communication with other ships works fine before and after switching to the new private keys.

PR #5535, opened 2022-01-11
Removing delete option from DMs

arthyn:
This PR addresses the misconstrued nature of "deleting" dms as stated here urbit/landscape#1110 by removing the option entirely.

A few other minor fixes:
- fixes DM loading from #5525 
- fixes an issue in dev which prevents you from refreshing the page at any url besides the root
- updates bad dependency routes and types in current next/groups


liam-fitzgerald (approved):
LGTM!
Good catch of the DM reloading

PR #5536, opened 2022-01-12
group-chat: fix joins

arthyn:
This fixes a few issues with the way `JoinRoute` was processing `kind` so we were never getting correct differentiation between groups and group chats which caused issues when clicking 'View Group'. 

- ensures that we retain join request state if the modal is open, and ensure that group is in state before allowing them to navigate to it
- notification link handling to redirect to the already joined chat, if already joined.
- fixes an issue causing notifications to not show at all because the check for desk was slightly off.
- stops DMs from trying to load if we haven't accepted it yet
- fixes potential undefined ref in `SidebarListHeader` urbit/landscape#1268

The original "joins" refactor addressed urbit/landscape#1308, urbit/landscape#1342, but this fixes the final part of the flow and a few other miscellaneous bugs.

Fixes urbit/landscape#1342

### DIFF
--- a/pkg/garden/desk.docket-0
+++ b/pkg/garden/desk.docket-0
@@ -1,10 +1,10 @@
 :~  title+'System'
     info+'An app launcher for Urbit.'
     color+0xee.5432
-    glob-http+['https://bootstrap.urbit.org/glob-0v4.64ana.19ug9.ik7l6.og080.68ce4.glob' 0v4.64ana.19ug9.ik7l6.og080.68ce4]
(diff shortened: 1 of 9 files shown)


lycurgus:
Meta nit: I suspect that issue link should be urbit/landscape#1308 - currently it links to an old PR :)

arthyn:
Oops yep thanks!

nerveharp:
@liam-fitzgerald Can you take a look at this? Thank you :)

PR #5539, opened 2022-01-13
hoon: caching compiler data structures and generator

### LINKED ISSUE #1732: Compiler jets should be disabled
Making changes, especially OTA changes, to hoon.hoon has historically been complicated by the presence of various compiler jets (i.e. jets to arms of `+ut`).

#1422 "decapitated" all the caching compiler jets and removed the others that didn't rely on a cache.  Per the linked PR, the caching logic in the remaining jets should be moved into Hoon proper.

(It would also be desirable to formally characterise the performance impact of removing the jets.)

jtobin:
## Update ~2022.10.12

Per discussion in the comments. this PR now adds:

* general-purpose LRU cache and priority search queue data structures to layer two of hoon.hoon,
* a suite of jets for the same, and
* a +hoon-caching generator, consisting of a hoon compiler that doesn't rely on jets.

The aim is to punt on upgrading the compiler proper until the performance of its jetless replacement (currently about 2x slower) can be improved.  But, the data structures supporting it are useful more generally.

(The original summary follows below.)

----

## Summary

* Replaces the "jetted" Hoon compiler, i.e. the +ut core in layer five of hoon.hoon that relies on a Vere-level cache implemented via jets, with a "caching" Hoon compiler, i.e. an +ut that relies on a Hoon-level cache for operation instead.  The old compiler jets (for +mint, +rest, +nest, etc.) are removed.

  With these changes hoon.hoon becomes independent of the implementation of Vere, and thus becomes more tractably updateable over-the-air.

* Adds fast (jetted) generic LRU cache and priority search queue implementations to layer two of hoon.hoon.

Resolves #1732, subsumes and replaces #5094.

### Dejetting the compiler

The compiler jets, i.e. jets for various arms of +ut, are "the last of Hoon's unnatural ties to Vere."  The present compiler relies on them (and specifically, as of #1422, a Vere or runtime-level cache implemented by them) in order to be practically efficient, resulting in the language & compiler being tightly and unnecessarily coupled with the particular runtime of Vere.  Notably at present, this coupling makes it difficult to perform compiler upgrades OTA.

The solution to this problem is to replace the runtime-level cache with a compiler-level cache, making the cache directly accessible to the compiler itself and thus obviating the need for the runtime to manage it.

This is accomplished first by the introduction of generic compiler state, $grub, that is threaded monadically through the entirety of +ut, and that can be accessed and modified locally as desired.  All arms in +ut that previously returned values at type `foo` now return values at type `[foo grub]`, and all +ut internals have been modified to make heavy use of tisket (`=^`) accordingly.  The state can be preserved and made accessible to other invocations of +ut.

$grub contains, at present, LRU caches for each of +nest, +rest, +mint, +mull, +fish, +crop, and +fuse -- a collection of arms found empirically to constitute a sort of minimal sweet spot for performance -- as well as several one-off internal caches intended to live only for the duration of a given arm invocation.  However, $grub can (and is likely to) be extended arbitrarily.  For example, it's suspected that +fond -- presently a major performance hot spot -- can be cached somewhat more intelligently in order to yield a sizeable performance win, so a separate cache is likely to be added for it at some point in the near future.

(N.b., some work has also been done on +fond to make it more amenable to caching, so its structure has changed quite a bit beyond the standard tiskettery that can be seen elsewhere.  It is not cached in these changes, however.)

### Adding supporting data structures

A generic LRU cache implementation used for the caches in $grub is also added to layer two of hoon.hoon under the +lu arm, and this LRU cache is backed by a new priority search queue implementation similarly added to layer two under +up. 
 Both implementations come with a full suite of jets, and with them should be roughly comparable to Hoon's +map in terms of performance.

(N.b., the PSQ implementation is more specialised and exposes a smaller API than the larger-scale one previously included in #5094.  It can be further fleshed out as needed, the motivation to make it smaller merely being to have to write fewer jets for the time being.)

## Etc.

The scope of the changes mean that the diffs are probably not exactly reviewable in any sane form, so aside from general takes on implementation/structure/style and localised spot checks, most of the review onus here should probably be on the changes passing relevant tests and meeting appropriate benchmarks.  Eager to receive comments on all of this; in particular, I'll update benchmarks I've made previously and post them here in short order.

I'll be running over a final checklist before unmarking this as draft and requesting the changes proper; there are a lot of moving parts that have been assembled over a long period of time, so there are a few final sanity checks and tweaks to be made in order to glue everything together in satisfactory fashion.

Off the top of my head, these are in order:

- [x] jets for the LRU cache/PSQ implementations are confirmed to be free of memory leaks
- [x] Nock confirmed to match between the jetted/caching compilers on standard test cases (e.g. compiling compiler/kernel)
- [x] performance benchmarks (in both time & memory) are updated and the differential is characterised rigorously
- [x] old +ut arms are deleted, new +caching-{nest, mint, etc.} arms are swapped in and renamed
- [x] +ut takes $grub as sample, invocations elsewhere in codebase are updated accordingly
- [x] old compiler jets are removed


ohAitch:
What sort of performance impact is the monadic cache having so far, when you self-compile hoon.hoon?

jtobin:
> What sort of performance impact is the monadic cache having so far, when you self-compile hoon.hoon?

It's currently a little under 2x slower than the jetted compiler for hoon.hoon.  I'm optimistic that this can be improved and will spend some time on performance measurement & tweaking in short order; presently chasing down the source of a stubborn Nock mismatch.

jtobin:
(N.b., the aforementioned Nock mismatch -- and really type + Nock mismatch, more accurately 'product of +mint' mismatch -- has proven to be very difficult to characterise, having withstood a fair bit of investigation from both myself and @philipcmonk.  Increasingly I'm of the opinion that it's benign; it only seems to appear when compiling layer 5c of hoon.hoon, i.e. the one containing +ut and +us, and even then only the 5c found in particular versions of hoon.hoon.  It doesn't appear to *do* anything, or at least anything obvious, in that at least booting a ship with the caching compiler appears to result in normal operation, and e.g. running the entire test suite produces no failures.)

jtobin:
Pulling in #5623 eliminates the Nock mismatch.

jtobin:
> Next steps?

Apologies for the delay here, resolving the Nock mismatch unfortunately burned *way* more time than I'd expected and I'm just back from a couple weeks of vacation.

Currently scouring everything to make sure the performance characteristics are completely understood and that there's no other low-hanging fruit lying around.  Expect this to be ready to go next week.

jtobin:
> Expect this to be ready to go next week.

Well, almost -- I found some of that aforementioned low-hanging fruit.  The change to the +feud jets about doubles the performance of PSQ inserts/lookups/deletes, and I suspect I can shave down the special-case lookup + priority bump operation used by the LRU cache as well.

Some data structure benchmarks collected below: put/raw/vip are inserts, get/see are lookups, del is delete, and bot is min-priority view, each performed on 100k elements.  In general 'qah' refers to the PSQ of interest (the one used to power the LRU cache).

```
                        unjetted       jetted

[%map-put n=100.000]   3076869µs     701436µs
[%qor-put n=100.000]  53166340µs    1331361µs
[%qat-put n=100.000]  18776979µs     987986µs
[%qat-raw n=100.000]   9731359µs     723338µs
[%qat-vip n=100.000]   8563671µs     719774µs
[%qah-put n=100.000]  52643640µs    1457769µs
[%qah-raw n=100.000]  25926399µs    1037767µs
[%qah-vip n=100.000]  21230061µs    1019978µs

[%map-get n=100.000]   2013671µs     530186µs
[%qor-get n=100.000]   9629062µs     963429µs
[%qat-get n=100.000]  15654980µs     701033µs
[%qah-get n=100.000]  22928727µs     895456µs
[%qat-see n=100.000]  18830281µs     886385µs
[%qah-see n=100.000]  25557845µs    1287261µs

[%map-del n=100.000]   2079493µs     543489µs
[%qor-del n=100.000]  59909723µs    1201152µs
[%qat-del n=100.000]   3746781µs     299017µs
[%qah-del n=100.000]  10553658µs     804973µs

[%qor-bot n=100.000]  54747735µs     978562µs
[%qat-bot n=100.000]   2352517µs     247852µs
[%qah-bot n=100.000]   7689472µs     684823µs
```

tacryt-socryp:
Ping. Progress?

jtobin:
> Ping. Progress?

Still chugging.  There is an issue with memory usage; while the time differential is not so so bad, the caching compiler uses *way* too much memory as compared with the jetted compiler (think ~1G vs 40MB to compile hoon.hoon).  After exhaustively testing the data structure jets for memory leaks, my current suspicion is that the hoon-level cache is being leaked undesirably when dealing with seminouns via `+laze`; i.e., that the thunk that is created for a %lazy seminoun winds up having in its subject the hoon-level cache, and that these are not being shared well structurally (incorrect seminoun handling was also involved in the aforementioned Nock mismatch that took months to track down).

What I'm going to try tomorrow is to make the +musk core accept the cache as a sample, and then supply it with the current cache whenever calling +complete:musk, which is the only entry point to processing %lazy seminouns.  At present, whatever cache present when calling +laze:ut is used to construct the thunk, and is then just thrown away afterwards, which seems really suspect.

jtobin:
(There was a similar issue with the hoon-level cache leaking into the *vere*-level memo cache, which is really pathological.  The solution here was to manually memoize +burp, rather than using sigluses.)

tacryt-socryp:
The `..^$` being passed into `+caching-hemped` inside of `+caching-laze` during stencil construction seems like it could be a leak. Thoughts?

jtobin:
Yeah, that's precisely it -- `|=(axe=@ud ..)` is the dodgy thunk in question.  Will update after some probing!

jtobin:
Some additional simple benchmarks:

```
  +mint:ut  (hoon.hoon, layer five)
            took s/42.069.090          (jetted)
            took s/90.689.227          (caching)

            max cache size:
            total: MB/71.655.480

            (hoon.hoon, layer four)
            took s/7.168.708           (jetted)
            took s/16.708.800          (caching)

            max cache size:
            total: MB/21.049.248

            (hoon.hoon, layer three)
            took s/1.630.651           (jetted)
            took s/2.766.979           (caching)

  +solid    2:11:32                    (jetted)
            4:16:36                    (caching)
```

Compiler-heavy operations are about 2x slower with the current settings.  There is a decent bit of flexibility here; memory usage can be further suppressed (probably to about 20-30MB for hoon.hoon), but of course at the cost of higher compile times.

jtobin:
Ok, I've opened this up.  One note re: the remark on +fond in the OP:

> N.b., some work has also been done on +fond to make it more amenable to caching, so its structure has changed quite a bit beyond the standard tiskettery that can be seen elsewhere. It is not cached in these changes, however.

I've since reverted these changes for the time being, but will preserve them in the git history for when I'm able to explore better-caching +fond again.

Regarding these changes more generally: I'm presently running this compiler on a mainnet moon and the ship is running normally and without issue.  It's likely we don't want to actually release these changes onto the network yet, as the only thing users will notice initially is that their compiler has slowed down somewhat when performing heavier tasks, but it's worth continuing to move the work ahead.

tacryt-socryp:
Tbh merge it 😂 The compiler is not in the hot path for commonly performed tasks, and it is unclean to have long-running PRs!

philipcmonk:
> Tbh merge it

I tentatively agree with this.

For testing and deployment, we'll need a binary that has both the +ut and +up caches enabled

philipcmonk:
I notice that this line takes way longer to run:

```
!,  *hoon  a=1
```

Specifically, dojo +sell's the result, and that takes a long time.  The trace is attached, which suggests that it's spending 20 seconds in nest, which is being called 2000 times.  My guess is that nest call used to "happen" to be cached because it was in the same road, and now we need to explicitly thread around the  cache in +dole.  Specifically it calls +meet and +repo, and those should probably save their caches.

[trace.tar.gz](https://github.com/urbit/urbit/files/9124986/trace.tar.gz)

joemfb:
> The compiler is not in the hot path for commonly performed tasks

lol. lmao.

> we'll need a binary that has both the +ut and +up caches enabled

That's right, we can't delete the old compiler jets here, or anytime soon.

tacryt-socryp:
At the least, when I left Tlon, after spending months combing through Eyre and Gall performance traces, the compiler was no longer in the hot path (at least for Groups)! I suppose I can't speak for now.

philipcmonk:
Opened #5903 against this branch to resolve the slowness in prettyprinting.

I will say that subjectively it feels like it takes significantly longer to compile desks with these changes.

jtobin:
> Tbh merge it 😂 The compiler is not in the hot path for commonly performed tasks, and it is unclean to have long-running PRs! Plus the LRU stuff is independently useful. Exciting stuff!

> I tentatively agree with this.

Not like I don't want to have this work officially in the main tree, but I disagree that we should just merge it in and push it out because it's "done" (by some definition).  These are *big freaking changes*, and ones that have a known serious negative effect on end users -- namely, roughly 2x slower compiles.  Moreover it's likely that there's some weirdness that isn't well-captured by simple direct compilation tests and ship operation monitoring that I've done, e.g. the issue that @philipcmonk found and fixed above.

In my opinion these changes shouldn't be pushed out to the network until things are at least at rough parity with the present (jetted) compiler's performance.  While the branch is, uh, *very*, long running, and long-running branches are not the nicest things in the world, it isn't going to bit-rot, as I'm still minding it.

(The PSQ/LRU stuff I agree could go out separately if there's demand for it, though at present the API's are specialised to what the compiler strictly needs.  They can be expanded very easily, though.)

philipcmonk:
What do you think about merging this just as a generator plus the psq jets?
That way we can easily use either compiler, and it's guaranteed to stay
working because we regularly ensure all generators compile.

tacryt-socryp:
Note: turns out we have a use case for an LRU in the Uqbar transaction execution engine! Just flagging this so that it's noted that use of the LRU elsewhere is no longer theoretical.

jtobin:
Ok, well what might be ideal for the time being would be to:

1) Flesh out the PSQ / LRU cache API's so that they're useful more generally, i.e. not just to the compiler.  There are a lot of basic, desirable arms that are just missing.  This isn't a big job.

2) Release the PSQ / LRU stuff as a library or two so that it can just be imported trivially in whatever application.  When we eventually need to move it into the compiler, we can do that, and consumers can just update their code accordingly.  (We should also patch tree.c accordingly and release the jets with Vere.)

3) Add the caching compiler as a generator, per @philipcmonk's suggestion, and just have it import the LRU library for now.  Then it's easy to pick up the work later.

Thoughts?

jtobin:
> Release the PSQ / LRU stuff as a library or two so that it can just be imported trivially in whatever application. When we eventually need to move it into the compiler, we can do that, and consumers can just update their code accordingly. (We should also patch tree.c accordingly and release the jets with Vere.)

Alternatively the PSQ / LRU stuff could also be just added to hoon.hoon outright, which would be simpler, and odds are it's going to go in there eventually anyway..

jtobin:
Just a note: we resolved out-of-band to go ahead with the above plan.  I'll flesh out the API for the data structures and then PR the changes adding them to hoon.hoon (as well as their associated jets to Vere), and we'll add the new compiler as a generator for the time being.  I'm involved with some other work at the moment, but should be able to come back to this and finish it off in a week or so.

jtobin:
> Just a note: we resolved out-of-band to go ahead with the above plan. I'll flesh out the API for the data structures and then PR the changes adding them to hoon.hoon (as well as their associated jets to Vere), and we'll add the new compiler as a generator for the time being.

General PSA that this is complete (I've updated the PR title and description accordingly).

jtobin:
I'm inclined to rip the generator out of this PR.  The data structures are a much smaller & less controversial add than the giant generator, which is more annoying to keep up to date with upstream changes to hoon.hoon, and which would probably see limited use were it to be merged in anyway.

Maybe better to just kill the generator and create another branch & (draft) PR for the +ut changes to hoon.hoon for whenever this work can be picked up again?

jtobin:
(Years later this old PR popped into my head and I wound up thinking about it. There was no need to strictly "get rid" of the compiler jets, as we originally aimed to do here; this work eliminated the *necessity* of the compiler jets, which is the more important item, but it's still fine to *use* the things in order to keep the compiler fast.

If anyone ever goes about resurrecting this stuff, I suggest using (something like) the compiler included here, while also keeping the existing compiler jets enabled. That should be a strict improvement.)

PR #5542, opened 2022-01-14
main: realpath uses straight malloc

mrdomino:


mrdomino:
This doesn't matter since `c3_free` is just a wrapper around `free`.

PR #5545, opened 2022-01-18
ames: ignore missing peer-state on-publ-sponsor

yosoyubik:
This adds a fix that crashed the [latest OTA](https://github.com/urbit/urbit/releases/tag/urbit-os-v2.119) due to a [`%freaky-alien`](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/ames.hoon#L1835) error. The solution is to use `+get-peer-state` instead, and ignore the event.


philipcmonk (approved):
LGTM, thanks!

PR #5546, opened 2022-01-18
u3: consolidate functions in events.c

mcevoypeter:
Clean up [`events.c`](https://github.com/urbit/urbit/tree/peter/truncate-event-log/consolidate/pkg/urbit/noun/events.c) by;
- deleting functions that are not called from anywhere in the system (`u3e_dirty`, `u3e_check()`, `_ce_check_page()`, `_ce_maplloc()`, `_ce_mapfree()`);
- deleting short static functions in [`events.c`](https://github.com/urbit/urbit/tree/peter/consolidate-event-functions/pkg/urbit/noun/events.c) with only a single call site (`_ce_patch_write_page()`, `_ce_patch_create()`, `_ce_patch_read_control()`, `_ce_patch_write_control()`) and instead inling the functionality at the call site;
- and folding the functionality of `_ce_patch_free()` into `_ce_patch_delete()`.

### DIFF
--- a/pkg/urbit/include/noun/events.h
+++ b/pkg/urbit/include/noun/events.h
@@ -75,11 +75,6 @@
       c3_o
       u3e_live(c3_o nuu_o, c3_c* dir_c);
 
-    /* u3e_dirty(): count dirty pages.
-    */
-      c3_w
-      u3e_dirty(void);
-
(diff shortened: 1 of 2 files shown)


mcevoypeter:
Merging this to avoid creating a backlog of changes on the event log project, but conversation on whether this is a good idea or not can continue and changes can be reverted if necessary.

PR #5548, opened 2022-01-19
eth-watcher: fix zoom-margin to account for reorgs

yosoyubik:
The previous value—used for testing—didn't consider block reorgs, which meant that if we zoom to the latest block that has no transactions, but that gets later replaced by a 1-block reorg that does have a transaction, we'll miss it, making our Azimuth state incomplete.

To fix it, we rewind the Azimuth state to the contents of the snapshot, and then start retrieving logs from the latest one we have.


philipcmonk (approved):
Looks good

PR #5555, opened 2022-01-21
u3, vere: create snapshot module

mcevoypeter:
The motivation behind this PR is to draw a clear boundary around the snapshot system.

- Move [`events.c`](https://github.com/urbit/urbit/tree/peter/truncate-event-log/integrate/pkg/urbit/noun/events.c) out of [`pkg/urbit/noun`](https://github.com/urbit/urbit/tree/peter/truncate-event-log/integrate/pkg/urbit/noun) and into [`pkg/urbit/vere`](https://github.com/urbit/urbit/tree/peter/truncate-event-log/integrate/pkg/urbit/vere) because the `events` module is not concerned with nouns or Nock execution. The same, of course, goes for [`events.h`](https://github.com/urbit/urbit/tree/peter/truncate-event-log/integrate/pkg/urbit/include/noun/events.h).

- Rename the `events` module as the `snapshot` module, which is more concrete. 

- Remove the `u3_Pool`/`u3P` global variable in favor of a static global variable that is only accessible within the `events`/`snapshot` module.

- Begin improving documentation of the `events`/`snapshot` module by introducing a basic Doxygen configuration file and adopting a commenting style that aims to accurately describe parts of the code that are not self evident while simultaneously minimizing the amount of boilerplate comments.


mcevoypeter:
Merging this to avoid creating a backlog of changes on the event log project, but conversation on whether this is a good idea or not can continue and changes can be reverted if necessary.

PR #5557, opened 2022-01-23
lib/ethereum: +address-to-checksum gate to ethereum lib as per EIP-55

xiphiness:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-55.md

I needed it because some api's i'm hitting fail if the address isn't checksummed.

Could maybe be written better, haven't done much byte munging in hoon. Feel free to improve.

### DIFF
--- a/pkg/base-dev/lib/ethereum.hoon
+++ b/pkg/base-dev/lib/ethereum.hoon
@@ -948,6 +948,21 @@
   %-  prefix-hex
   (render-hex-bytes 20 `@`a)
 ::
+++  address-to-checksum
+  |=  =address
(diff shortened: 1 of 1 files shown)


Fang- (commented):
Sweet! Very nice to have this. For completeness, could you add the [test cases](https://eips.ethereum.org/EIPS/eip-55#test-cases) to `/tests/lib/ethereum/encoding`?

Probably nice to have a checksum validation at some point, but not in scope for here.

xiphiness:
will do. re: checksum validation, can just test for equality

PR #5558, opened 2022-01-24
build: arm64 architecture

mrdomino:
This branch includes support for building on M1 Mac with stock nix.

### DIFF
--- a/nix/pkgs/urbit/default.nix
+++ b/nix/pkgs/urbit/default.nix
@@ -59,7 +59,7 @@ in stdenv.mkDerivation {
     then [ "--disable-shared" "--enable-static" ]
     else [];
 
-  CFLAGS = [ (if enableDebug then "-O0" else "-O3") "-g" ]
+  CFLAGS = (if enableDebug then [ "-O0" "-g" ] else [ "-O3" ])
     ++ lib.optionals (!enableDebug) [ "-Werror" ];
 
   MEMORY_DEBUG = enableDebug;
--- a/nix/pkgs/urbit/default.nix
+++ b/nix/pkgs/urbit/default.nix
@@ -14,6 +14,12 @@ let
 
   version = builtins.readFile "${src}/version";
 
+  # See https://github.com/urbit/urbit/issues/5561
+  oFlags =
+    if stdenv.hostPlatform.system == "aarch64-darwin"
+    then (if enableDebug then [ "-O0" "-g" ] else [ "-O3" ])
+    else [ (if enableDebug then "-O0" else "-O3") "-g" ];
+
 in stdenv.mkDerivation {
   inherit src version;
 
@@ -59,8 +65,7 @@ in stdenv.mkDerivation {
     then [ "--disable-shared" "--enable-static" ]
     else [];
 
-  CFLAGS = (if enableDebug then [ "-O0" "-g" ] else [ "-O3" ])
-    ++ lib.optionals (!enableDebug) [ "-Werror" ];
+  CFLAGS = oFlags ++ lib.optionals (!enableDebug) [ "-Werror" ];
 
   MEMORY_DEBUG = enableDebug;
   CPU_DEBUG = enableDebug;
--- a/pkg/urbit/configure
+++ b/pkg/urbit/configure
@@ -93,7 +93,7 @@ case $(tr A-Z a-z <<< $cpu) in
   x86_64)
      defmacro U3_OS_ENDIAN_little 1
      ;;
-  aarch64)
+  arm64|aarch64)
      defmacro U3_OS_ENDIAN_little 1
      defmacro U3_CPU_aarch64 1
      ;;


nix/pkgs/urbit/default.nix
```diff
@@ -59,7 +59,7 @@ in stdenv.mkDerivation {
     then [ "--disable-shared" "--enable-static" ]
     else [];
 
-  CFLAGS = [ (if enableDebug then "-O0" else "-O3") "-g" ]
+  CFLAGS = (if enableDebug then [ "-O0" "-g" ] else [ "-O3" ])
```
eamsden, line 62: could this be made specific to arm64 since it doesn't appear to be an issue anywhere else? I assume we want debugging symbols in prod builds cc @joemfb @brendanhay

joemfb: Yeah, those symbols have been pretty handy on several occasions. I'd like to keep them around where possible.

pkg/urbit/configure
```diff
@@ -93,7 +93,7 @@ case $(tr A-Z a-z <<< $cpu) in
   x86_64)
      defmacro U3_OS_ENDIAN_little 1
      ;;
-  aarch64)
+  arm64|aarch64)
```
joemfb, line 96: `aarch64` is handled differently in a couple places (smaller event log mmap size), so it'd be best to handle `arm64` separately:

```
      ;;
+  arm64)
+     defmacro U3_OS_ENDIAN_little 1
+     defmacro U3_CPU_aarch64 1
+     ;;
   *)
```

I had done this locally, but forgot to PR it. Thanks!

joemfb (changes requested):

pkg/urbit/configure
mrdomino: Done, though this seems like it's structurally identical?

nix/pkgs/urbit/default.nix
mrdomino: Done

nix/pkgs/urbit/default.nix
```diff
@@ -14,6 +14,12 @@ let
 
   version = builtins.readFile "${src}/version";
 
+  # See https://github.com/urbit/urbit/issues/5561
+  oFlags =
+    if stdenv.hostPlatform.system == "aarch64-darwin"
```
mrdomino, line 19: i don't know why this isn't `"arm64-darwin"`, but it isn't.

mrdomino:
un-requested review from @brendanhay since we're no longer getting rid of debug symbols everywhere.

pkg/urbit/configure
joemfb: They are, and that's my bad. I meant to say `defmacro U3_CPU_arm64 1` (the whole point is to avoid the ARM special-casing that already existed for raspberry pi).

mrdomino: fixed

mrdomino: hmm, however, glancing at the code, it looks like at all the call sites they check for `defined(U3_CPU_aarch64) && defined(U3_OS_linux)`.

There seems to be some ambiguity around the architecture; see also [nix/pkgs/urbit/default.nix:19](https://github.com/urbit/urbit/pull/5558/files/a280a9dd8b7689c2943ab8f5a0fb7c35217a2004#diff-54800b555b0f35695547d8178e90a1f0e3ab5e7c1abcbeaffeb1d6d9464e57b3R19).

I suppose if nix is going to standardize on `arm64` eventually and stuff like that is a bug, then I'm fine with the different macro. But if not, I'm inclined to keep `aarch64` overloaded for now to match the ill-specified environment.

joemfb: Looks like you're right, `arm64` and `aarch64` are apparently synonyms: https://www.phoronix.com/scan.php?page=news_item&px=MTY5ODk. Request retracted.

mrdomino: \> article is from 2014
\> naming still confusing

good times.

joemfb (approved):
LGTM

pkg/urbit/configure
botter-nidnul: > hmm, however, glancing at the code, it looks like at all the call sites they check for `defined(U3_CPU_aarch64) && defined(U3_OS_linux)`.

glad to see my forward planning with `&&` prevented a problem here

> I suppose if nix is going to standardize on `arm64` eventually and stuff like that is a bug, then I'm fine with the different macro. But if not, I'm inclined to keep `aarch64` overloaded for now to match the ill-specified environment.

Afraid this is going to be a lasting problem. I think `aarch64-darwin` is what nix has [adopted internally](https://github.com/NixOS/nix/commit/addf9f4edeab8e8e2cb115dd98dc764cc1f19fcf), and is what nix previously used to refer just to iOS (which causes its own problems).

But external packages still might use the `arm64` name in places, and `uname` will return different results depending if it's the gnu version (`aarch64`) or apple's version (`arm64`).

mrdomino:
Yeah, imagine the fun when someone tries to get urbit running on Ubuntu on
M1.

PR #5562, opened 2022-01-24
dm-hook: mutate screened before giving fact

liam-fitzgerald:
Fixes an issue where the subscription update would contain stale data


liam-fitzgerald:
cc: @arthyn

arthyn (approved):

PR #5563, opened 2022-01-25
groups: perf optimization

arthyn:
This PR addresses a few issues with how we load things in Groups.

- it removes using `Set` for group data so we can store it safely in local storage so your groups are available immediately after first launch
- adds in a good amount of code-splitting so our initial bundle is much smaller
- moves the redirecting action from outside protocol links to be much earlier
- removes random unused code


patosullivan (approved):
Lots of good looking changes here. No outstanding issues (there are a few files that were touched here that happen to have other eslint errors, but that is beyond the scope of this PR I'm sure). Runs great locally, all basic functionality works and it does feel snappier.

tomholford (approved):
Well done, many welcome changes :muscle: LGTM

PR #5568, opened 2022-01-26
vere: do not use eth-mainnet for galaxy booting

Fang-:
Fallback to the default happens in dawn.c, which correctly points to
roller.urbit.org, an endpoint that matches its request/response logic.
Continuing to use an Ethereum endpoint instead of an L2 one will just result
in 400s, since they don't speak the same language.

Arguably `ops_u.eth_c` should be renamed, but out of scope for here.

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -277,9 +277,6 @@ _main_getopt(c3_i argc, c3_c** argv)
     }
   }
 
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):

timlucmiptev:
For anyone with this issue before the release is out: adding option `-e https://roller.urbit.org/v1/azimuth` got past this.

PR #5569, opened 2022-01-26
Fix rekey multikey

drbeefsupreme:
This is identical to #5522, but I was told I should have merged into `next/arvo` instead of `release/next-userspace`.


drbeefsupreme:
Nevermind lemme clean up the git history

PR #5570, opened 2022-01-26
helm: fix |rekey to work with multikey files

drbeefsupreme:
This is identical to #5522, I was just asked to merge it into `next/arvo` instead of `release/next-userspace`

### DIFF
--- a/pkg/arvo/lib/hood/helm.hoon
+++ b/pkg/arvo/lib/hood/helm.hoon
@@ -68,12 +68,10 @@
   =|  caz=(list card)
   %-  emil
   |-
-  ?~  kyz.fed
-    caz
-  =/  sed=seed:jael  [who [lyf key ~]:i.kyz]:fed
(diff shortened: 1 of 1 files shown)


Fang-:
I actually like the code as it exists in #5522 a little bit better. This does an unnecessary assignment to `sed` and does repeated `+snoc`s instead of a single `+flop`.

Any particular reason for these changes? (Seems functionally equivalent though.)

drbeefsupreme:
Whoops, that's what I get for copying and pasting on a new branch without double checking instead of doing things the right way. Thanks for catching that. It was supposed to be identical to #5522.

drbeefsupreme:
@Fang- OK now it should actually be identical.

Fang- (approved):
> copying and pasting on a new branch

Just `git cherry-pick` the commit over. (;

This looks good, thanks!

PR #5571, opened 2022-01-27
khan: thread runner vane

mrdomino:


mrdomino:
Related: #5345 implements the `conn` C driver, which makes use of `%khan` to run threads.

mrdomino:
reviewed with @philipcmonk; we decided to add unit tests, so i’ll mark this a draft until that is done.

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,148 @@
...
+  ?~  ret=(rof ~ %cb beak /[mark])
+    ~|(%mark-unknown !!)
+  ?~  u.ret
+    ~|(%mark-invalid !!)
+  ?>  =(%dais p.u.u.ret)
+  !<(dais:clay q.u.u.ret)
+++  start-spider
+  |=  [our=@p =cage]
+  ^-  note
+  [%g %deal [our our] %spider %poke cage]
```
mrdomino, line 43: @philipcmonk should either end of the sock change if we request a thread from an external `+beak`?

mrdomino:
This ought to be orthogonal to the 1.9 vere release; `conn` and `%khan` should each work fine without the other.

pkg/arvo/sys/vane/khan.hoon
mrdomino: @Fang- says `[our our]` is correct here

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,214 @@
...
+  ^-  (pair beak mark)
+  :_  (snag 5 wire)
+  :+  (slav %p (snag 1 wire))
+    (snag 2 wire)
+  =/  ras  (snag 4 wire)
+  ?+  (snag 3 wire)  !!
+    %da   [%da (slav %da ras)]
+    %tas  [%tas ras]
+    %ud   [%ud (slav %ud ras)]
+  ==
```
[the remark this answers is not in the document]
philipcmonk: Also, I don't really believe in `snag`, at least not with a constant index, which is why it's essentially unused in kernel space.  The usual pattern is more like:

```
?>  ?=([%fyrd @ @ @ @ @ ~] wire)
:_  i.t.t.t.t.t.wire
:+  (slav %p i.t.wire)
  i.t.t.wire
(slav i.t.t.t.wire i.t.t.t.t.wire)
```
or
```
?>  ?=([%fyrd @ @ @ @ @ ~] wire)
=*  =ship  (slav %p i.t.wire)
=*  =desk  i.t.t.wire
=/  =case  [i i.t]:t.t.t.wire
=*  =mark  i.t.t.t.t.t.wire
[[ship desk (slav case)] mark]
```

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,214 @@
...
+    |=(=note ^-(move [hen %pass //g note]))
+  ::
+      %fyrd
+    =*  fyd     p.task
+    =/  =beak   (get-beak bear.fyd now)
+    =/  =wire   (make-wire beak p.args.fyd)
+    =/  =dais:clay
+      (get-dais beak p.q.args.fyd rof)
+    =/  =vase
+      (slam !>(some) (vale.dais q.q.args.fyd))
```
philipcmonk, line 147: This works, but feels maybe a little overpowered.  `(slap (vale.dais q.q.args.fyd) !,(*hoon [~ u=.]))` feels simpler in some sense, because you're not calling a function in vase mode.

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,214 @@
...
+      ==  ==                                            ::
+    +$  sign                                            ::    in response $<-
+      $%  $:  %gall                                     ::    from %gall
+              $>(%unto gift:gall)                       ::  update
+          ==                                            ::
+          $:  %khan                                     ::    from self
+              $>  ?(%arow %avow)  gift                  ::  thread result
+      ==  ==                                            ::
+    +$  khan-state                                      ::
+      [%2 hey=duct]                                     ::  current unix duct
```
philipcmonk, line 54: Can set version to `%0` for initial release

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,214 @@
...
+      old
+    [%2 hey.old]
+  khan-gate(state new)
+::  +scry: nothing to see as yet
+::
+++  scry
+  ^-  roon
+  |=  [lyc=gang car=term bem=beam]
+  ^-  (unit (unit cage))
+  ~
```
philipcmonk, line 170: Does `|mass` work without implementing `/whey`?  I don't remember.

pkg/arvo/tests/sys/vane/khan.hoon
```diff
@@ -0,0 +1,414 @@
...
+::
+/+  *test
+/=  khan-raw  /sys/vane/khan
+=/  khan-gate  (khan-raw ~nul)
+|%
+++  test-khan-fyrd-start-args
+  =^  born-moves  khan-gate
+    %-  khan-call  :*
+      khan-gate
+      now=~1111.1.1
```
philipcmonk, line 14: recommend ~1162.1.1

philipcmonk (commented):
Looks great, mostly just stylistic nits.

pkg/arvo/sys/vane/khan.hoon
mrdomino: Maybe? I see:

```
    %khan:
      %dot: B/344
      %typ: KB/92.540
      %sac: KB/8.088
    --KB/100.972
```

mrdomino:
@philipcmonk PTAL

pkg/arvo/sys/vane/khan.hoon
joemfb: I made it optional recently.

pkg/arvo/sys/lull.hoon
```diff
@@ -2092,6 +2092,30 @@
...
+  |%
+  +$  gift                                              ::  out result <-$
+    $%  [%arow p=(avow vase)]                           ::  in-arvo result
+        [%avow p=(avow crag)]                           ::  external result
+    ==                                                  ::
+  +$  task                                              ::  in request ->$
+    $~  [%vega ~]                                       ::
+    $%  $>(%born vane-task)                             ::  new unix process
+        [%done ~]                                       ::  socket closed
+        [%fard p=(fyrd vase)]                           ::  in-arvo thread
```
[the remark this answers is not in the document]
joemfb: It's possible to change the type, but it's much easier to preserve the type and just change behavior.

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,230 @@
...
+  (read-case [i i.t]:t.t.t.wire)
+++  start-spider
+  |=  [our=@p =vase]
+  ^-  note
+  [%g %deal [our our] %spider %poke %spider-start vase]
+++  watch-spider
+  |=  [our=@p =path]
+  ^-  note
+  [%g %deal [our our] %spider %watch path]
+--
```
joemfb, line 125: Style nit: all the arms in this core should be separated by blank comments.

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,230 @@
...
+  =/  =task  ((harden task) wrapped-task)
+  ?^  dud
+    ~|(%khan-call-dud (mean tang.u.dud))
+  ?+    -.task  [~ khan-gate]
+      %born
+    [~ khan-gate(hey hen)]
+  ::
+      %fard
+    =/  tid=@ta
+      (cat 3 'khan-fyrd--' (scot %uv (sham eny)))
```
joemfb, line 150: Arvo does not currently rotate entropy during an event, so thread IDs will collide if multiple `%fard`'s are passed in a single event. So ID generation will need to be stateful within `%kahn`.

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,230 @@
...
+  ==
+++  read-wire
+  |=  =wire
+  ^-  (pair beak mark)
+  ~|  khan-read-wire+wire
+  ?>  ?=([%fyrd @ @ @ @ @ ~] wire)
+  :_  i.t.t.t.t.t.wire
+  :+  (slav %p i.t.wire)
+    i.t.t.wire
+  (read-case [i i.t]:t.t.t.wire)
```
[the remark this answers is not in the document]
joemfb: And for `+make-wire` above: `[%fyrd (en-beam [beak mark ~])]`.

joemfb (changes requested):

pkg/arvo/sys/arvo.hoon
```diff
@@ -44,6 +45,7 @@
   ==
 +$  cage  (cask vase)
 ++  cask  |$  [a]  (pair mark a)
++$  crag  (cask *)
```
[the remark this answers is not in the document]
mrdomino: disagree re reusing `$card` fwiw, that has the connotation of being an event. we use `$mark` and `$desk` instead of `@tas`. but will remove `*`.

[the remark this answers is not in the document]
mrdomino: OTOH maybe the desire here is that threads should take `$card`s?

pkg/arvo/sys/vane/khan.hoon
mrdomino: resolved this, but not sure yet how to test it — just call `+call` multiple times against the same core?

mrdomino: ah, you just call multiple times against the same `eny` argument. ok, working on this

pkg/arvo/sys/vane/khan.hoon
```diff
@@ -0,0 +1,222 @@
...
+  ?+    -.task  [~ khan-gate]
+      %born
+    [~ khan-gate(hey hen)]
+  ::
+      %fard
+    =/  tid=@ta
+      %^  cat  3
+        'khan-fyrd--'
+      (scot %uv (sham (mix tic eny)))
+    =.  tic     (mod +(tic) (bex 128))
```
joemfb, line 142: This seems like overkill. Not only is it hard to imagine this call `+mod` ever affecting the value of `tic`, but the actual space of collisions that we want to avoid is only within a single event. If you're worried about `tic` growing unbounded, you could just reset it to `0` on `%born`.

(You could alternately persist the entropy in your state on `%born`, and then stretch/iterate it yourself a la `|og`. But that might be overkill for generating ids.)

pkg/arvo/sys/arvo.hoon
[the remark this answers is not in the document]
mrdomino: talked about this with @philipcmonk — we decided that `$card` is not a real `(cask)` and should instead be `(pair @tas *)` (since the type is not a mark). we also noticed that `$page` in clay is already what we want, and decided to promote it into arvo and use it instead of `$crag`.

joemfb (changes requested):

philipcmonk (approved):
Pushed a new pill because the old one had `%2` and wouldn't boot with `-A pkg/arvo`.  Tested, was able to invoke threads with `|pass`.  Looks good!

mrdomino:
cool, about to update all pills in this branch.

joemfb:
@mrdomino I was actively engaged in reviewing this PR

pkg/arvo/sys/arvo.hoon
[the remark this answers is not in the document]
joemfb: I disagree with this, at least in part. In the broader sense of `mark` as "symbolic content type", `card` is certainly a `(cask)`. Events, effects, userspace inputs (`%poke` payloads) and outputs (subscription updates), and successful scry responses all have the same shape -- a noun tagged with a symbolic name/description/type. The name used at all these different levels are by no means all backed by mark files, but that's already true of mark usage patterns in userspace.

I say "at least in part" because we might want to restrict `card` to mean "untyped event/effect" (although it already means more than that in userspace).

philipcmonk:
That's my bad, I didn't look closely enough at where you were at in the review

mrdomino:
yeah, apologies. happy to submit more work in another PR or do whatever else you prefer.

pkg/arvo/sys/arvo.hoon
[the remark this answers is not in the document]
mrdomino: Based on all this, I'm leaning:

- `$card` is a `(cask)` (I have no dog in the fight between it being that or `(pair @tas *)`).
- `$card` and `$page` are separate names. `$card` connotes an untyped arvo event, and `$page` connotes an untyped `$cage`.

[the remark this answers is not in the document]
mrdomino: (or even `(pair event-type *)`)

PR #5574, opened 2022-01-29
hotfix/groups: fix join links

arthyn:
- Handles join kind more safely
- Handles web+urbitgraph in join prompt


nerveharp (approved):
String matching a-go-go

PR #5575, opened 2022-01-31
vere/unix: sane(r) path handling

mrdomino:
This tries to prevent vere from ever handling non-canonical paths at
the C level. It introduces wrapped fs calls that crash-fail if they
encounter a path containing `''` (empty filename), `'.'`, or `'..'`,
replaces ordinary fs calls with their wrapped counterparts, and adds
assertions before other fs calls (e.g. opening the lmdb log file.)

In cases where we want to handle non-canonical paths (e.g. command-line
arguments), we canonicalize the paths prior to passing them to the fs
calls. To do this, we first try calling `realpath`. If that fails, we
assume it’s because at most one layer of path has not yet been created,
and try to do `realpath` on `dirname` of the passed path, then append
`basename` to the result.

In the unix IO driver, we introduce a simple mapping between `+knot`
and C filenames that prepends a `~.` to the knots `~`, `'.'`, or `'..'`
as well as any knots starting with `'~.'`. We also enforce stricter
filename validation, ignoring any paths containing non-`@ta` characters
when processing commits from the filesystem.

There is also some minor reorganization here; `u3_unix_acquire` /
`release` made reasoning about `unix.c` difficult and had comments
suggesting that it be relocated to `vere/disk.c`, so I did so. I also
removed `vere/walk.c` except for `u3_walk_save`, which is now in
`unix.c` as `u3_unix_save` with a substantially simpler (but not yet
ideal) API. Some low-hanging fruit would be to make `u3_unix_save`
take its path as a `u3_noun` instead and do knot conversion. A future
refactor would probably share logic between `u3_unix_save` and
`_unix_sync_change`, e.g. by making `unx_u` optional and/or adding a
“one-time sync” flag.


Fang- (commented):
@mrdomino and I went over this together on-call. This all seems fine afaict (after latest commits). Would be good if someone with "authority over vere" could do another pass & formally approve, unless they want to put their trust in us.

mrdomino:
@joemfb and i discussed this in our 1-on-1 today. sounds like as long as this is understood as temporary, and there is an intention to rewrite unix.c “soon”, it’s fine to merge as-is.

mrdomino:
@joemfb PTAL at the comment i just pushed

mrdomino:
@joemfb i will merge this soon unless any objections

joemfb (approved):
LGTM

PR #5578, opened 2022-02-02
aqua: rerevive

joemfb:
Building on #5029 (which seems to have been automatically merged?), here's a couple of changes to get `:aqua` closer to working. The `%pill`/`%boot-pill` mark changes had never been fully reverted. And the `%aqua-effect` conversion to `%mime` is stubbed out, as I don't see how it should be necessary (but its absence was causing crashes).

It seems like `%ames` packets are not being delivered. And some timers may also be being dropped -- the basic ph tests hang after booting.

/cc @lukechampine

### DIFF
--- a/pkg/arvo/gen/brass.hoon
+++ b/pkg/arvo/gen/brass.hoon
@@ -26,7 +26,7 @@
       ::
         ~
     ==
-:-  %boot-pill
+:-  %pill
 ^-  pill:pill
 ::  sys: root path to boot system, `/~me/[desk]/now/sys`
 ::  bas: root path to boot system' desk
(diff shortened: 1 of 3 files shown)


joemfb:
See urbit/urbit.org#1272

philipcmonk:
I wonder if it ever got updated after the last round of Ames protocol
changes. It only really needs to parse out the target ship, so that should
be an easy fix.

joemfb:
It looks like it just overwrites lanes, and doesn't seem like that would've changed.

belisarius222:
Hm, we did change the origin encoding in the last round of ames protocol changes. Maybe that broke.

—
~rovnys-ricfer
https://urbit.org

PR #5579, opened 2022-02-02
noun: go ahead and enforce that invariant

mrdomino:


mrdomino:
closed because i forgot about windows and thought it needed more work there, but it seems fine now actually

joemfb (approved):
TYFYS

Issue #5580, opened 2022-02-03
jets that bail with %exit get `ice` reset


mrdomino:
**Describe the bug**
Equivalence testing is currently implemented as follows:
1. Check `_(ham_u->ice)`.
2. If `c3n`, then set `ham_u->ice = c3y`.
3. Run `ham_u->fun_f`.
4. Set `ham_u->ice = c3n`.

However, if `ham_u->fun_f` returns `u3m_bail(c3__exit)` (as commonly happens in jet code), then:
1. no diagnostic is printed indicating that the call bailed.
2. `ham_u->ice` is never reset to `c3n`.
3. Future calls will therefore treat the jet as though it was perfect.

**To Reproduce**
Steps to reproduce the behaviour:
1. Set a jet’s `.ice` to `c3n`.
2. Have the jet body `return u3m_bail(c3__exit)`.
3. Observe (e.g. through printfs / `~&`) that the hoon code is not called after the first bail, and the C code is called.

**Screenshots**

**System (please supply the following information, if relevant):**
 - Vere 1.9rc1

**Notify maintainers**
@joemfb @frodwith

joemfb:
This behavior is fine for equivalence testing in local development, but obviously not suitable for livenet-deployed testing. This was known, just not to me -- moar and better testing remains highly desirable.

PR #5581, opened 2022-02-03
jets: disable equivalence test ice toggle (#5580)

mrdomino:


joemfb:
The design here (longstanding, I believe) is to allow for outer-level equivalence testing, while still jetting internal recursion or reentrant calls. That obviously doesn't play well with bails, and may require separately virtualizing the nock and jet invocations to be generally useful. Simply disabling this approach to recursion may be too costly ...

/cc @frodwith

mrdomino:
This is a natural use case for something like go’s `defer`. Alas, I don't know of a way to model that easily in C…

mrdomino:
So the practical issue here is that `sosi` (the schnorr signature jet) bails if you call it with wrong-sized arguments, as does the hoon. So equivalence testing for it gets disabled right after the first unit test for bounds checking within it.

If possible, it'd be nice to push this change with the next release (since we don't seem to be doing any equivalence testing aside from the schnorr jets, which don't recurse), and then ship an actual fix later.

joemfb (commented):
This change is not practical, as it will redo the equivalence testing for every iteration of a jetted recursive gate. So this equivalence testing regime is not suitable for long-running, livenet usage.

It was my mistake to suggest it as a strategy for #5510. Please back out the testing change from that PR, we can discuss further testing there. This PR is a WONTMERGE.

Issue #5586, opened 2022-02-08
%depth-first scry for %graph-store is impossible to resolve successfully.


rabsef-bicrym:
[Graph-Store 's `%depth-first` scry is impossible to complete as written](https://github.com/urbit/urbit/blob/874d380ad5d5e1bca0c22c498294572c6d160f51/pkg/landscape/app/graph-store.hoon#L860).

The right way of framing this scry, however, is unclear - either it should nest under `node` or be its own thing - but currently it doesn't function either way.

According to the documentation, the path expected would result in a scry looking like this:
`^(update:store %gx /=graph-store=/graph/~zod/test-chat/node/depth-first/<count>/<start>/graph-update-3)`
  *  Incidentally, the docs should call out that this is a non-`noun` `%gx` scry

This would, to my understanding, result in a path of `/x/graph/~zod/chat/node/depth-first/<count>/<start>/graph-store-3`.  This makes sense for other `node` path-ed scries.  See lns 683, 706 where `t.t.t.t.path` is checked for `[%node *]`.
```
> =a /x/graph/~zod/res-name/node/some-other-data/a-mark
> +>+>.a
[%node %some-other-data %a-mark ~]
```

For the proposed structure of a `%depth-first` scry, we'd have:
```
> =a /x/graph/~zod/res-name/node/depth-first/1/0/graph-update-3
> +>+>.a
[%node %depth-first ~.1 ~.0 %graph-update-3 ~]
```

However, at line 859, we're checking `t.t.t.t.path` to be `[%depth-store @ @ ~]` which, as we see above, would in fact be `%node` if constructed as per documentation. If, however, we take the documentation to be mistaken about lumping this path in under `node`, we still fail because ln 860 checks `i.t.t.t.t.path` for `count` (`(rush count dem:ag)`) and gives a `[~ ~]` if `count` is null, which it absolutely must here.

I propose that we change the documentation to not lump this under `node` and add a `t` to that check at line `860` thereby resolving the problem - consider the following edit made to line 859 and the result, described below (just to be clear, this still breaks - but it would be fixed w/ extra `t`s):
**edit**
```
        [%depth-first @ @ ~]
      ~&  >>  "I'm in depth first"
      =/  count=(unit atom)  (rush i.t.t.t.t.path dem:ag)
      =/  start=(unit atom)  (rush i.t.t.t.t.t.path dem:ag)
      ?:  ?=(~ count)
        ~&  >>>  [t.t.t.t.path count]
[...]

PR #5587, opened 2022-02-08
%depth-first has pattern-matching issue

rabsef-bicrym:
support a `%depth-first` scry constructed as follows `.^(* %gx /=graph-store=/graph/<ship of resource>/<name of resource>/depth-first/<count>/<start>/graph-update-3)`

see #5586

### DIFF
--- a/pkg/landscape/app/graph-store.hoon
+++ b/pkg/landscape/app/graph-store.hoon
@@ -857,8 +857,8 @@
       --
     ::
         [%depth-first @ @ ~]
-      =/  count=(unit atom)  (rush i.t.t.t.t.path dem:ag)
-      =/  start=(unit atom)  (rush i.t.t.t.t.t.path dem:ag)
+      =/  count=(unit atom)  (rush i.t.t.t.t.t.path dem:ag)
+      =/  start=(unit atom)  (rush i.t.t.t.t.t.t.path dem:ag)
       ?:  ?=(~ count)
         [~ ~]
       :-  ~  :-  ~  :-  mar


pkg/landscape/app/graph-store.hoon
```diff
@@ -857,8 +857,8 @@
       --
     ::
         [%depth-first @ @ ~]
-      =/  count=(unit atom)  (rush i.t.t.t.t.path dem:ag)
-      =/  start=(unit atom)  (rush i.t.t.t.t.t.path dem:ag)
+      =/  count=(unit atom)  (rush i.t.t.t.t.t.path dem:ag)
+      =/  start=(unit atom)  (rush i.t.t.t.t.t.t.path dem:ag)
```
joemfb, line 861: This isn't actionable right now, but it's unfortunate that this interface doesn't/can't use `(slaw %ud ...)`, which is jetted.

liam-fitzgerald:
cc: @tacryt-socryp This looks ok to me, but I'm uncertain where it's used or what it's supposed to mean

rabsef-bicrym:
If it is helpful, I intended to use it to paginate a scry into graph-store. I'd use sail to render `count` documents from `start` and get the index of the last post to use in the button to then get the `start` for the next page (or something like that).

tacryt-socryp:
It was intended to be used in group feed to show a timeline view, but I don't believe that was ever merged in. Maybe it was? If the code doesn't work, it wasn't

liam-fitzgerald (approved):

PR #5588, opened 2022-02-08
c3: implement generic list

mcevoypeter:
Implements and tests a doubly-linked-list-backed queue abstraction with iterator support. The assumption is that this will be a useful primitive in higher layers of the runtime. At the very least, it'll be useful in upcoming event log truncation work.

### DIFF
--- a/pkg/urbit/Makefile
+++ b/pkg/urbit/Makefile
@@ -1,6 +1,7 @@
 include config.mk
 include $(foreach dir,$(compat),$(wildcard compat/$(dir)/*.mk))
 
+c 	 = $(wildcard c/*.c)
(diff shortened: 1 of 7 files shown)


mcevoypeter:
The functionality required is leaning more towards a list rather than a strict queue, so I've renamed the module `list`.

mcevoypeter:
A few updates:
- I cut the number of allocations required to allocate a new node from two to one by using a flexible array struct member. 
- I rewrote the simpler functions as macros, which will remove any overhead (however slight) associated with function calls.
- The `node` struct is now exposed in the interface, but all of its members should be accessed through an appropriate macro, so we'll be able to change the implementation without changing the interface.

PR #5589, opened 2022-02-08
vere: define abstract event log interface

mcevoypeter:
Defines a single abstract interface behind which all event log functionality can be pulled. This will make the event log easier to reason about. It will also enable us to develop multiple event log implementations--different databases, epochs vs. no epochs, local storage vs. distributed/network-based storage--while minimizing the impact on other areas of the runtime, namely the implementation of the state machine.


mcevoypeter:
After talking in-depth with @joemfb about this, the move to an abstract event log interface isn't appropriate now when we can't anticipate the exact changes to the persistence layer that lurk on the horizon.

Issue #5590, opened 2022-02-09
Groups impossible to use from iPhone 12 mini


tacryt-socryp:
It takes 30s+ to load a channel over wifi on a 300mb/s fiber connection. I have no clue what happened, but the speed of the network connection from the client to the Urbit has gotten steadily worse over the past few months, and has now hit the point of being entirely unusable.

tacryt-socryp:
Can confirm running `:observe-hook &observe-action [%warm-cache-all ~]` fixes it entirely. I recommend releasing an OTA that runs this action on all ships. @liam-fitzgerald

mirtyl-wacdec:
Can someone explain what that command does and why would `observe-hook` be slowing down airlock requests? We've also had some issues with Urbit Visor being slower than usual lately.

liam-fitzgerald:
Good catch @tacryt-socryp must've died in the distribution OTA

@mirtyl-wacdec that command ensures that all the data structure definitions are already cached so they don't require compilation at run time

cc: @belisarius222 this should just be the default behaviour of %clay, as we've discussed

eamsden:
I tried the above command:

```
> :observe-hook &observe-action [%warm-cache-all ~]
/sys/vane/gall/hoon:<[1.372 9].[1.372 37]>
/app/observe-hook/hoon:<[211 3].[287 5]>
/app/observe-hook/hoon:<[212 3].[287 5]>
/app/observe-hook/hoon:<[213 3].[287 5]>
/app/observe-hook/hoon:<[215 3].[287 5]>
/app/observe-hook/hoon:<[216 3].[226 5]>
/app/observe-hook/hoon:<[217 3].[226 5]>
/app/observe-hook/hoon:<[218 3].[226 5]>
/app/observe-hook/hoon:<[219 3].[226 5]>
/app/observe-hook/hoon:<[222 30].[222 44]>
/app/observe-hook/hoon:<[260 5].[264 60]>
/app/observe-hook/hoon:<[261 7].[261 57]>
'cannot warm up cache that is already warm'
/app/observe-hook/hoon:<[261 54].[261 56]>
dojo: app poke failed
```

tacryt-socryp:
Your cache is already warm then. That's good!

PR #5591, opened 2022-02-09
landscape: cache marks again

liam-fitzgerald:
Fixes urbit/urbit#5590


tacryt-socryp (approved):

tacryt-socryp:
Note: the code may need to be tweaked to support multiple desks in order to truly cache *all* marks, but in practice made my landscape faster as is.

liam-fitzgerald:
That should probably come when clay caches all marks by default imo.

PR #5593, opened 2022-02-09
zuse: add ++ns:dejs:format

liam-fitzgerald:
Adds a signed integer JSON parser, i.e. a gate of type $-(json @s)


Fang- (approved):
This seems fine. Thanks for adding tests!

#5566 may want to be aware of this addition, since it touches the tests also.

PR #5594, opened 2022-02-09
vere: show git hash in vere output

mrdomino:
e.g.:

```console
% urbit ~/zod
~
urbit 1.8 (git 2de3581ae0)
```

### DIFF
--- a/pkg/urbit/configure
+++ b/pkg/urbit/configure
@@ -3,6 +3,7 @@
 set -euo pipefail
 
 URBIT_VERSION="$(cat ./version)"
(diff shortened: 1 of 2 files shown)


mrdomino:
nvm, nix build env doesn't have git i guess

ohAitch:
nix should have a `self.shortRev` looks like, can you pass that as the env variable? https://discourse.nixos.org/t/passing-git-commit-hash-and-tag-to-build-with-flakes/11355

mrdomino:
don't see any straightforward way to get at shortRev from the derivation.

PR #5595, opened 2022-02-09
ames: comet communication fixes

lukechampine:
This is a set of changes to ames that affect comet communication, loosely following @philipcmonk's design. Specifically:

- Self-attestation packets are now processed via `+on-publ-full`. This means that comets will respond to a self-attestation with their own self-attestation. It also fixes a bug where receiving a self-attestation would clobber existing peer state.
- Non-comets now send a special "sendkeys" packet to comets that they wish to communicate with. Technically, this is an empty `shut-packet`, encrypted with dummy key. The comet will drop this packet on the floor, but will still respond with a self-attestation.
  - When a comet wants to talk to another comet, it sends its own self-attestation rather than a sendkeys packet. The effect is the same (the receiver replies with their self-attestation), but we save a roundtrip.
- Messages addressed to unknown comets are not enqueued; instead, they are immediately sent to the comet's sponsor. (Without this, we can't talk to a comet until they talk to us first.)

So the packet exchange for planet->comet is:
-> "plz send key"
<- "here's my key"
-> "message 1 packet 1"
<- "ack 1:1"
 
and for comet->comet:
-> "here's my key"
<- "here's my key"
-> "message 1 packet 1" and "here's my key"
<- "ack 1:1"

I added some tests for galaxy-to-comet and comet-to-comet communication. The test setup is a bit bloated now; could probably use some cleaning up. It would be nice to have a pH test too, but I think that's blocked on https://github.com/urbit/urbit/pull/5578.

I think the only thing missing here is timers (and maybe acks?) for resending self-attestation/sendkeys packets. I'm not familiar with how timers work in ames, so I could use some direction there. I'd also like to test what happens if a comet receives a sendkeys packet from a known peer -- will the dummy key cause a crash? Is that okay? If not, we'll need to explicitly check for sendkey packets somehow.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1757,6 +1757,8 @@
         =/  ship-state  (~(get by peers.ames-state) ship)
         ::
         ?.  ?=([~ %known *] ship-state)
+          ?:  ?=(%pawn (clan:title ship))
+            (try-next-sponsor (^sein:title ship))
           %+  enqueue-alien-todo  ship
           |=  todos=alien-agenda
           todos(packets (~(put in packets.todos) blob))
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1301,25 +1301,27 @@
       event-core
     ::
     =/  =open-packet  (decode-open-packet packet our life.ames-state)
-    ::  store comet as peer in our state
+    ::  add comet as an %alien if we haven't already
+    ::
+    =?  peers.ames-state  ?=(~ ship-state)
+      (~(put by peers.ames-state) sndr.packet %alien *alien-agenda)
+    ::  upgrade comet to %known via on-publ-full
+    ::
+    =.  event-core
+      =/  crypto-suite=@ud  1
+      =/  =point
+        :*  ^=     rift  0
+            ^=     life  sndr-life.open-packet
+            ^=     keys  (my [sndr-life.open-packet crypto-suite public-key.open-packet]~)
+            ^=  sponsor  `(^sein:title sndr.packet)
+        ==
+      (on-publ / [%full (my [sndr.packet point]~)])
+    ::  manually add the lane to the peer state
     ::
     =.  peers.ames-state
-      %+  ~(put by peers.ames-state)  sndr.packet
-      ^-  ^ship-state
-      :-  %known
-      =|  =peer-state
-      =/  our-private-key  sec:ex:crypto-core.ames-state
-      =/  =symmetric-key
-        (derive-symmetric-key public-key.open-packet our-private-key)
-      ::
-      %_  peer-state
-        qos            [%unborn now]
-        symmetric-key  symmetric-key
-        life           sndr-life.open-packet
-        public-key     public-key.open-packet
-        sponsor        (^sein:title sndr.packet)
-        route          `[direct=%.n lane]
-      ==
+      =/  =peer-state  (gut-peer-state sndr.packet)
+      =.  route.peer-state  `[direct=%.n lane]
+      (~(put by peers.ames-state) sndr.packet %known peer-state)
     ::
     event-core
   ::  +on-hear-shut: handle receipt of encrypted packet
@@ -1712,7 +1714,8 @@
   ::  +enqueue-alien-todo: helper to enqueue a pending request
   ::
   ::    Also requests key and life from Jael on first request.
-  ::    On a comet, enqueues self-attestation packet on first request.
+  ::    If we're a comet talking to another comet, sends our
+  ::    self-attestation packet.
   ::
   ++  enqueue-alien-todo
     |=  [=ship mutate=$-(alien-agenda alien-agenda)]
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1820,6 +1831,23 @@
...
+    ^-  blob
+    ?>  ?=(%pawn (clan:title her))
+    %-  encode-packet
+    %-  encode-shut-packet
+    :*  ^=    shut-packet  *shut-packet
+        ^=  symmetric-key  *symmetric-key
+        ^=           sndr  our
+        ^=           rcvr  her
+        ^=      sndr-life  life.ames-state
+        ^=      rcvr-life  1
```
joemfb, line 1849: I wonder if these lives should both be `0`, since normal packets will never have that.

joemfb (commented):
Both attestation and sendkeys packets are currently informal (no flow state, no acks, &c) -- that's probably not ideal in some sense, but it's much simpler than the alternative.

For timers, I think you'll want to emit `[%wait (add now ~s30)` or somesuch on an appropriate wire (`/alien`?) and then handle it in `+on-wake:event-core` by checking for that wire and dispatching to the send logic in `+enqueue-alien-todo` (which could ignore `already-pending` in the comet case).

I'm curious to hear what @philipcmonk and @belisarius222 think of this.

joemfb:
Receiving a sendkeys packet from a known pier will cause a crash. I think that's fine:

- sendkeys are only sent from non-comets to comets
- comets will usually already have the keys in %jael (except for moons), so the pier state can switch to %known before the sendkeys sender receives the attestation
- unless the comet sends a %plea, we won't hit the current resend-attestation case in `+on-wake:event-core`
- but if we had a retry timer as i've suggested, the attestation packet will eventually get through

pkg/arvo/sys/vane/ames.hoon
lukechampine: sure -- it's moot since the packet should never be decoded in the first place, but no harm in making it extra clear that this is not a valid packet

lukechampine:
ok, took a stab at implementing a retry timer

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1729,14 +1739,33 @@
...
+  ::
+  ::    Also sets a timer to resend the request every 30s.
+  ::
+  ++  request-attestation
+    |=  [=ship]
+    ^+  event-core
+    =/  packet  ?.  =(%pawn (clan:title our))
+                  (sendkeys-packet ship)
+                (attestation-packet ship 1)
+    =.  event-core  (send-blob | ship packet)
```
lukechampine, line 1766: is there a cleaner way to write this?

joemfb: I don't think so. Both "actions" (`+send-blob` and `+emit`) produce the whole event-core, and dispatching each requires local state, so you can't easily sequence them with `=>`. You could build some kind of helper core hear, but this seems fine.

joemfb (changes requested):
A couple state management points, and some style nits:

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1301,25 +1301,25 @@
...
-      ::
-      %_  peer-state
-        qos            [%unborn now]
-        symmetric-key  symmetric-key
-        life           sndr-life.open-packet
-        public-key     public-key.open-packet
-        sponsor        (^sein:title sndr.packet)
-        route          `[direct=%.n lane]
-      ==
+      (~(put by peers.ames-state) sndr.packet %known peer-state)
```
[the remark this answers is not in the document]
lukechampine: 100% agree!

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1301,25 +1301,27 @@
...
     ::
     =/  =open-packet  (decode-open-packet packet our life.ames-state)
-    ::  store comet as peer in our state
+    ::  add comet as an %alien if we haven't already
+    ::
+    =?  peers.ames-state  ?=(~ ship-state)
+      (~(put by peers.ames-state) sndr.packet %alien *alien-agenda)
+    ::  upgrade comet to %known via on-publ-full
+    ::
+    =.  event-core
```
lukechampine, line 1310: tbh I'm not sure I understand the distinction between `event-core` and `..event-core` here. How can I check which one is correct?

belisarius222:
If used in the left-hand side of a =. then event-core and ..event-core are equivalent. I use just event-core since it's shorter and more evocative imo.

—
~rovnys-ricfer
https://urbit.org

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1420,6 +1424,17 @@
     |=  [=wire error=(unit tang)]
     ^+  event-core
     ::
+    ?:  ?=([%alien @ ~] wire)
```
belisarius222, line 1427: Unless I'm missing something, as written, this will keep re-sending the sendkeys packet forever.  To end it, we should check whether our ship-state for this peer has gone from %alien to %known, and only re-send if not.

belisarius222 (changes requested):
LGTM other than what appear to be infinitely recurring requests for attestations.

pkg/arvo/sys/vane/ames.hoon
lukechampine: good catch, fixed

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1420,6 +1424,20 @@
...
+      ::  if we haven't received an attestation, ask again
+      ::
+      ?^  error
+        %-  (slog leaf+"ames: attestation timer failed: {<u.error>}" ~)
+        event-core
+      ?~  ship=`(unit @p)`(slaw %p i.t.wire)
+        %-  (slog leaf+"ames: got timer for strange wire: {<wire>}" ~)
+        event-core
+      =/  state=(unit peer-state)  (get-peer-state u.ship)
+      ?:  ?=([~ %known *] ship-state)
```
belisarius222, line 1437: ship-state -> state

lukechampine: huh. I can see why this didn't trigger a compilation error, but it'd be nice if it could (either by flagging `state` as an "unused face" or the comparison as always false).

Anyway, `=/  ship-state  (~(get by peers.ames-state) u.ship)` seems like the more common pattern in ames, so I went with that.

belisarius222 (approved):
LGTM now.

lukechampine:
bump 😬

belisarius222:
We're going to try to start testing this on some livenet ships soon, in preparation for a wider deployment.  Sorry for the wait.

PR #5602, opened 2022-02-18
u3: expose snapshot duplication in events.h

mcevoypeter:
Enable copying the snapshot to a caller-provided directory.
- Remove `_ce_backup()`, which copies the snapshot to a hard-coded directory (`<pier>/.urb/bhk`) that is created if it doesn't already exist.
- Add `u3e_copy()`, which copies the snapshot to a caller-provided directory that must already exist. 

The following sequence works as expected (ship boots without replaying after restoring the snapshot from the `bhk` directory):
```console
$ rm -r bus/.urb/bhk
$ urbit bus
$ rm bus/.urb/chk/*
$ cp bus/.urb/bhk/* bus/.urb/chk
$ urbit bus
```


pkg/urbit/noun/events.c
```diff
@@ -71,6 +71,18 @@
...
+
+//! South segment image file name.
+static const c3_c sou_nam_c[] = "south.bin";
+
+//! Control patch file name.
+static const c3_c ctl_nam_c[] = "control.bin";
+
+//! Memory patch file name.
+static const c3_c mem_nam_c[] = "memory.bin";
+
```
joemfb, line 85: `_nam` is redundant in each of these, s/b just `nor_c`, `sou_c`, &c.

pkg/urbit/noun/events.c
```diff
@@ -937,7 +900,96 @@ u3e_save(void)
...
+  snprintf(pax_c, sizeof(pax_c), "%s/.urb/bhk", u3P.dir_c);
+  if ( c3n == u3e_copy(pax_c) ) {
+    fprintf(stderr, "loom: failed to copy snapshot to %s\r\n", pax_c);
+  }
+}
+
+//! @n (1) Attempt to `dir_c`.
+//! @n (2) Attempt to create north image file in `dir_c`.
+//! @n (3) Attempt to create south image file in `dir_c`.
+//! @n (4) Copy north and south image files to `dir_c` from `u3P.dir_c`.
```
joemfb, line 913: Same question as elsewhere with regards to separating/indexing these comments.

pkg/urbit/noun/events.c
```diff
@@ -937,7 +900,96 @@ u3e_save(void)
...
+close_north:
+  close(nop_u.fid_i);
+  if ( c3n == suc_o ) {
+    unlink(pan_c);
+  }
+exit:
+  return suc_o;
+}
+
+//! @n (1)
```
joemfb, line 967: And moreso here ;)

pkg/urbit/noun/events.c
```diff
@@ -937,7 +900,96 @@ u3e_save(void)
...
+  }
+
+  u3e_image nor_u, sou_u;
+
+  if ( (c3n == _ce_image_open(dir_c, &nor_u)) ||
+       (c3n == _ce_image_open(dir_c, &sou_u)) )
+  {
+    fprintf(stderr, "boot: failed to load snapshot in %s\r\n", dir_c);
+    exit(1);
+  }
```
joemfb, line 982: The error handling/return behavior should be reviewed (or revisited later, it's not the only such issue in this file).

pkg/urbit/noun/events.c
```diff
@@ -937,7 +900,96 @@ u3e_save(void)
...
+
+  u3e_foul();  // (1)
```
joemfb, line 984: If the behavior of this function is to load a snapshot from out-of-band, uncorrelated with the page-tracking and increment updates that drive updates to the live, writeable snapshot, then i think you'll want to call `u3e_foul()` after all the pages have been loaded into memory. (As an optimization, we might later want to parameterize  `_ce_image_blit()`'s page tracking and memory protection behavior, to avoid thrashing that state.)

Also worth noting, this function is necessary but not sufficient to restore an out-of-band snapshot. See `u3m_boot()`, or `_cu_realloc()` (called by `u3u_meld()`) for other, necessary loom reinitialization.

joemfb (changes requested):

pkg/urbit/noun/events.c
mcevoypeter: True, but it's nice for global variable names to consist of six characters, which makes them easily identifiable amidst the three-character names that are currently the standard for local variable names. I think the ease of identification is worth the few extra keystrokes.

pkg/urbit/noun/events.c
mcevoypeter: The thought is to mimic the use of footnotes in prose: further elaboration of a point does not belong at the site the point is made, which applies to comments in code just as much as it does for natural language. This way, comments don't break up the flow of the code because we're able to inline trivial `// ([0-9]+)` comments and point the reader elsewhere *should they need further explanation* of a particular line/chunk of code. The indexed list of "footnotes" (headnotes?) also often provides a quick high level overview of the steps that a given function is performing.

pkg/urbit/noun/events.c
mcevoypeter: > Also worth noting, this function is necessary but not sufficient to restore an out-of-band snapshot. See `u3m_boot()`, or `_cu_realloc()` (called by `u3u_meld()`) for other, necessary loom reinitialization.

In this case, then we should probably refactor `u3e_live()` to load a snapshot from an arbitrary directory so that we can simply call `u3m_boot()` when loading an epoch's snapshot.

joemfb: I don't think that's quite right; they're very different operations. `u3e_live()` restores a snapshot in a mutable context: later calls to `u3e_save()` will update that snapshot incrementally with any intervening, persistent state changes. `u3e_load()` restores an immutable snapshot: subsequent calls to `u3e_save()` won't affect it, but a different, mutable snapshot (optionally?).

I think it's reasonable for that distinction to exist in this API, I'm just noting that it will also need to exist at higher levels as well. (And once those higher levels exist, we may find that this new function needs to change, too.)

pkg/urbit/noun/events.c
joemfb: I agree that comments can be overused and disruptive. I've made efforts to group them together into reasonably sized blocks (more so in hoon, see the types in `sys/arvo.hoon`). But I find this style to be disruptive and ugly; `//! @n (1)` is clearly something that is written for a machine, not for me. What does doxygen do with it? I can't seem to find it in the manual.

mcevoypeter: > I find this style to be disruptive and ugly

I agree it's unconventional and can trigger a negative knee-jerk reaction like this, but the more I've used the convention, the more pleasant I've found it to use. It nicely separates my development process into "code" and "explain"; whenever I write a line/chunk of code that warrants some deeper explanation, I just drop a `([0-9]+)` comment in line and come back after I've finished with my "code thought" to explain. The division between code and the justification for the code is also perfectly separated in this style. If you just want to look at code, you can. If you just want to look at documentation, you can. If you want to look at both, you easily can because the two are clearly connected. The one drawback that I can think of is that it's a new style and so will take getting used to.

> `//! @n (1)` clearly something that is written for a machine, not for me.

Would `//! (1)` be more palatable? `@n` (an alternate form of [`\n`](https://www.doxygen.nl/manual/commands.html#cmdn)) forces a newline in the generated documentation and so is really just a nice-to-have. I'm happy to remove its use if it's distracting in the source code files.

mcevoypeter:

PR #5605, opened 2022-02-22
graph-store: fix scrying into archive

ryjm:
Also adds a /keys equivalent for the archive.


liam-fitzgerald (changes requested):
target `next/groups` please looks good otherwise

PR #5607, opened 2022-02-23
%gall: integrate agents with |mass

joemfb:
This PR updates %gall to allow its agents to optionally integrate with `|mass`. %gall's `/whey` scry handler peeks into each agent , and incorporates their labels if successful/valid (with error output suppressed).

Additionally, this PR includes a basic integration with `:graph-store`. Only the resource name is displayed, which might not be sufficient. It may also be worth grouping graphs and logs instead of listing the separately. This new output should be coupled with simple instructions for listing graphs/archives, measuring their memory usage, and removing them if it becomes necessary. (All the graph-store stuff could easily be a separate PR, can easily be released piecemeal, &c.)

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -840,11 +840,11 @@
   ::
   ++  mo-peek
     ~/  %mo-peek
-    |=  [dap=term =routes care=term =path]
+    |=  [veb=? dap=term =routes care=term =path]
     ^-  (unit (unit cage))
     ::
     =/  app  (ap-abed:ap dap routes)
-    (ap-peek:app care path)
+    (ap-peek:app veb care path)
   ::
   ++  mo-apply
     |=  [dap=term =routes =deal]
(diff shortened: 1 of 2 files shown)


belisarius222:
Sure, we'll want to make the graph-store integration more substantive, but this is a good start and worth deploying on its own.

Fang-:
Maybe there's a case to be made for this to scry at `/~/whey` or some other "this is clearly a reserved/system thing" path as opposed to plain `/whey`. Glad to get this either way!

liam-fitzgerald:
> Maybe there's a case to be made for this to scry at `/~/whey` or some other "this is clearly a reserved/system thing" path as opposed to plain `/whey`. Glad to get this either way!

Perhaps a care for system scries?

liam-fitzgerald (approved):
LGTM!
Graph-store event log pruning should probably arrive at some point, in which case it makes sense to measure logs and graphs separately

joemfb:
I agree with @Fang- and @liam-fitzgerald that it'd be best for this kind of thing to not be a reserved path in the agent's namespace. But I don't think a prefix helps much, and I'm broadly opposed to ad-hoc care proliferation ("talk to your kids about care control"). I expect we'll want to iterate on this sort of thing in the future.

PR #5608, opened 2022-02-23
roller: fix handling of forced transactions

yosoyubik:
This PR fixes the way the roller handled "forced" transactions (i.e. transactions that had a force flag indicating that they should be included in the next batch even if they fail to apply to the roller's predicted state), specially when it relates to the roller's transaction history. This PR also modifies the roller's state to keep track of this forced transactions when the are included in the sending map, so when they fail, they are still considered as part of the batch.

%roller-rpc and it's TS client [v0.1.15](https://www.npmjs.com/package/@urbit/roller-api/v/0.1.15) have been updated to support this.

The contents of the PR are currently deployed to ~roller-dozzod-dozzod and used with the latest version of Bridge.

(Note: this PR is built on top of the [roller/quota-fix](https://github.com/urbit/urbit/compare/roller/quota-fix) branch that has a small fix for retrieving the quota of a ship)

### DIFF
--- a/pkg/arvo/app/roller.hoon
+++ b/pkg/arvo/app/roller.hoon
@@ -1440,7 +1440,7 @@
     ?~  sen=(get:ors:dice sending [address nonce])
       ~?  lverb  [dap.bowl %weird-double-remove nonce+nonce]
       sending
-    ?~  nin=(find [raw-tx.diff]~ (turn txs.u.sen tail))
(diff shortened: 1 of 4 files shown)


tacryt-socryp:
What exactly is a forced transaction?

tacryt-socryp:
Can you go ahead and merge this? I don't get why it's laying fallow given that it works better than what's on master.

PR #5609, opened 2022-02-24
devex: peg node versions

tomholford:
This pegs the `interface` package to the current LTS version of Node, as well as the project root for `urbit/urbit`.

See: https://nodejs.org/en/about/releases/


arthyn (approved):
Looks good thanks for doing this!

PR #5610, opened 2022-02-25
interface: use single sig in NotificationText

tomholford:
# Changes

The `ProfileOverlay` component expects a `ship` arg as a patp without the sig. This deSigs the ship value used in `NotificationText`.

This closes urbit/landscape#1376

# Preview

image


liam-fitzgerald (approved):
This should (confusingly) target `next/groups`, else LGTM

PR #5612, opened 2022-02-26
clay: add `story`, a set of tools to support commit messages 

ynx0:
## Summary

This PR adds support in userspace for reading/writing/updating commit messages to clay.

- Added mark `story`
	- This mark is used to store commit messages in story files
	- story files are always expected to be at the root of a given desk, i.e. `base/story`. (N.B. this results in a file `base.story` outside of the mounted pier dir `base/`)
	- The mark has a proper unix representation, and handles diffing and merging of story files properly
- added a lib file
	- Contains arms to parse in from / write out to the unix plaintext representation
	- Contains two jug helper methods, `uni:ju` and `dif:ju`, which should ideally be lifted into the stdlib at some point
- Added a sur file that contains the core `story` datastructures
	- `+$  prose  [title=@t body=@t]` represents a commit message
    - `+$  story  (jug tako:clay prose)` is a map of commit hashes to a set of commit messages, where more than one message per hash represents conflicting messages assigned to the same hash.
	- `+$  proses  (set prose)` is the value type of the `story` jug.
    - `+$  chapter  [tako:clay proses]` represents a `story` entry 
    - `+$  cash  $%([%tako p=tako:clay] case)` is used in the various generators to accept a `%tako` directly in addition to a `case`
- Added generators to interact with story files
	- `|story-init, =desk <desk> =overwrite <flag>` creates a blank story file in the current desk if `=desk` is not given, and overwrites the story file if `=overwrite` is `%.y`.
	- `|story-write <title=@t> <body?=@t>, =desk <desk>, =cas <cash>` write the commit message to the current desk and case, preferring `=desk` and `=cas` if provided
	- `|story-remove <cash>, =desk <desk>, =prz <prose>` remove all commit messages from the given `cash` in the current desk, preferring `=desk`if provided. If `=prz` is provided, only that commit message will be removed from the `chapter` instead of all of them.
	- `+story-read, =desk <desk>` is the equivalent of `git log`, and logs the commits in order from most-recent to least-recent, and takes care to show correct output in complex cases such as a merge commit, a commit with two ancestors and no mergebase, and a commit with 3 ancestors (e.g. a commit from `|fuse`), and tries to avoid printing duplicates.
	- `+story-list, =desk <desk>` is an internal tool that simply `+turn`s over the `story` file in the current desk (or `=desk` if provided) and prints out the commit messages in an arbitrary order. It's main utility is that it can print out commit messages that are assigned to "orphan" commits and would otherwise never show up in the output of `+story-read`. This can occur because we allow the user to provide any commit hash when using `|story-write`, even hashes for commits that don't necessarily exist.

You can find an example of what usage looks like at [this](https://gist.github.com/ynx0/c485b9a5c1f61d11c4c4eefdbeb5ce73) gist.

## List of changes to clay

This effort motivated some extra scries being added to clay, and unearthed a bug in the way `|fuse` was written

- Added `%cs %tako` scry
	- Returns the commit hash associated with the given beak 
	- Example: `.^(tako:clay %cs /(scot %p our)/[desk]/(scot cas)/tako/~)`
- Added `%cs %base-tako` scry
	- Same as `%base` scry except it takes 2 `tako`s instead of 2 `desk`s
	- **ignores given beak**
	- Example: `.^((list tako:clay) %cs /(scot %p our)/$/(scot %da now)/base-tako/(scot %uv tako-a)/(scot %uv tako-b))`
- Added a `%cs %case` scry
	- Returns flag, `%.y` if provided case exists on given ship and desk, `%.n` otherwise (one issue left to iron out)
	- Example: `.^(? %cs /(scot %p our)/base/(scot %ud 1)/case)`, `.^(? %cs /(scot %p our)/base/(scot %ud 999)/case)`
- Fixed bug in `|fuse` which would prevent a normal invocation from working

Finally, I'd like to express my sincere gratitude to @philipcmonk for his mentorship and support during this project.


pkg/arvo/gen/story-read.hoon
```diff
@@ -0,0 +1,122 @@
...
+    ::
+    ::  From the context of e, commit c is on its relative mainline, or nowline, while commit d is on its sideline
+    ::  %base  a--b-------------X  :: mainline
+    ::  %new      \--c------e--/   :: nowline
+    ::  %new2        \--d--/       :: sideline 
+    ::
+    ::  
+    =/  sideline=tako:clay               i.reverse-ancestors
+    =/  mainline=tako:clay               i.t.reverse-ancestors
+    =/  mergebases                       .^((list tako:clay) %cs /(scot %p our)/[syd]/(scot cas)/base-tako/(scot %uv mainline)/(scot %uv sideline))  :: XX base-tako ignores beak
```
philipcmonk, line 78: We have a strict 80 columns rule, which applies in a few files here.

"80 columns is not just a good idea, it's the law!"

pkg/arvo/gen/story-read.hoon
```diff
@@ -0,0 +1,122 @@
...
+    =/  msg=(list cord)                  (msg-from-commit current-commit tale)
+    ::
+    ::  1 - process current commit
+    ::  2 - recur and queue processing on all commits on the sideline
+    ::  3 - recur and queue processing on all commits on the mainline
+    ::
+    ::  Because mainline messages are cons'd to result last, they are 
+    ::  (by definition) towards the less recent side of the resulting flopped list
+    ::
+    =.  state  [result=(weld msg result.state) mergebase=next-mergebase]  :: 1
```
philipcmonk, line 92: end comments should be aligned

pkg/base-dev/mar/story.hoon
```diff
@@ -0,0 +1,70 @@
+/-  *story
+/+  *story
+!:
+|_  tale=story
++$  story-diff       [additions=story deletions=story]
+::
+++  grad
+  |%
+  ++  form  %story-diff
```
philipcmonk, line 9: I think we're going to want a simple mark file for %story-diff so that they can be sent over the wire.  It can be very simple, just needs a conversion to/from noun.

The case where it might be needed is when you do a nontrivial merge of a story file and then try to sync that desk over the network.

pkg/base-dev/lib/story.hoon
```diff
@@ -0,0 +1,109 @@
+/-  *story
```
philipcmonk, line 1: These base-dev files need to be symlinked into pkg/arvo

pkg/arvo/gen/hood/story-write.hoon
```diff
@@ -0,0 +1,31 @@
+::  story: Attach a commit message to the commit associated with a given case on a given desk
+::
+::::
+  ::
+/-  *story
+:-  %say
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        [[=prose ~] =desk cas=cash ~]
```
philipcmonk, line 8: I would explode this to `[title=@t body=@t ~]` so that you can just run `|story-write 'title' 'body'` instead of needing brackets

philipcmonk (changes requested):
A few incidental changes need to be made, but looks good!

Very excited to have this, it's critical for eventually developing everything within urbit

pkg/arvo/gen/hood/story-write.hoon
ohAitch: And then to `[title=@t body=$@(~ [p=@t ~])]` so you can have just a title \:P
(Also I wonder what a snappier for this, most common, command would be - possibly just `|story`?)

ynx0: > And then to `[title=@t body=$@(~ [p=@t ~])]` so you can have just a title :P

what’s the advantage of using that type vs. a unit?

tacryt-socryp: In the "body is present" case you don't need a sig prior to the body

ohAitch: Yeah it's a "list with one or two elements" type
which maps more cleanly to dojo generator syntax

ynx0: that clears it up, thx

pkg/arvo/gen/story-read.hoon
ynx0: done in https://github.com/urbit/urbit/pull/5612/commits/a43dafdf4415297c800d66357d1b9546d108c1f7

pkg/base-dev/lib/story.hoon
ynx0: done in https://github.com/urbit/urbit/pull/5612/commits/add4452589f40146fe2e93ded4785b3f86eb50cc

pkg/arvo/gen/hood/story-write.hoon
ynx0: done in https://github.com/urbit/urbit/pull/5612/commits/18537e0ae0820d2f1a93bacb25f8ee6fa28ad59f

pkg/base-dev/mar/story.hoon
ynx0: done in https://github.com/urbit/urbit/pull/5612/commits/b43b9f195e12e6313d5eb68e401cfe85c324cc93

pkg/arvo/gen/story-read.hoon
ynx0: done in [`341e27b` (#5612)](https://github.com/urbit/urbit/pull/5612/commits/341e27b5c36ce8f6b20c138e606df84963891798)

ynx0:
@philipcmonk PTAL

philipcmonk (approved):
Looks good!

CI passes on Ubuntu and fails on Macos and Mingw, same as on next/arvo, so there's no regression there, and the tests are passing.

PR #5616, opened 2022-03-01
vere: sort motes

mrdomino:
just piped the whole body through !sort in vim.


pkg/urbit/include/c/motes.h
```diff
@@ -1274,10 +1274,10 @@
 #   define c3__wtts   c3_s4('w','t','t','s')
 #   define c3__wtzp   c3_s4('w','t','z','p')
 #   define c3__wyrd   c3_s4('w','y','r','d')
-#   define c3__yew    c3_s3('y','a','w')
 #   define c3__yell   c3_s4('y','e','l','l')
-#   define c3__z      c3_s1('z')
 #   define c3__yelp   c3_s4('y','e','l','p')
+#   define c3__yew    c3_s3('y','a','w')
```
lycurgus, line 1279: Apologies for the drive-by review but I noticed this one doesn't line up - `yew` vs `'y', 'a', 'w'`. Not sure if it needs to change? Or safer to leave it as-is...

mrdomino: > Apologies for the drive-by review

not at all!

and, scary. maybe we should write an automated check for these. anyway, `c3__yew` was not referenced anywhere so i just removed it.

belisarius222 (commented):
lgtm

belisarius222 (approved):
lgtm, second try

mrdomino:
fyi we looked into changing this so you could just do something like:

```c
c3_m4(tsst);
```

but this did not work. you can't `#define` in a `#define` (afaik) and if you change it to variable declarations then you break the use of motes in e.g. switch case statements.

PR #5617, opened 2022-03-02
c3: add binary file module for easy persistent storage of arbitrary data

mcevoypeter:
Write/read arbitrary byte arrays to/ from a file. Features:
- Simple interface
- Handles retries of partial reads and writes, which is a risk associated with the `read()` and `write()` system calls (respectively).

This interface has at least two cases (that I can think of) in the runtime:
(1) Event log metadata storage in the new epoch-based event log. Storing event log metadata in each epoch's LMDB instance is unnecessarily redundant, and having a full-blown LMDB instance specifically for four pieces of metadata (ship name, lifecycle length, whether the ship is fake, and version number) is overkill. One or more binary files containing the metadata is simpler and easier to interact with.
(2) Snapshot file management. The current snapshot implementation in [events.c](https://github.com/urbit/urbit/tree/peter/file/pkg/urbit/noun/events.c) uses the raw `read()` and `write()` system calls to read/write snapshot patch and image files, which are already binary files. Using the `c3_bile_*` interface in the snapshot module will reduce the complexity of a vital module.

More generally, it's a useful primitive that I suspect will come in handy in the future.

Tests:
See [bile_tests.c](https://github.com/urbit/urbit/tree/peter/file/pkg/urbit/tests/bile_tests.c).

### DIFF
--- a/pkg/urbit/c/bile.c
+++ b/pkg/urbit/c/bile.c
@@ -1,0 +1,192 @@
+//! @file bile.c
+
+#include "c/bile.h"
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stddef.h>
+#include <sys/stat.h>
+
+#include "c/defs.h"
+
+//==============================================================================
+// Macros
+//==============================================================================
+
+//! Error-handling wrapper for system calls that return >=0 on success
+//! and <0 on failure.
+//!
+//! It's unclear whether `strerror(errno)` is a portable solution for reporting
+//! error messages.
+//!
+//! @param[in] system_call     System call to execute.
+//! @param[in] failure_action  Statement to execute after logging failure.
+#define try_syscall(system_call, failure_action, ...)                          \
+  do {                                                                         \
+    if ( 0 > (system_call) ) {                                                 \
+      fprintf(stderr, "syscall: " __VA_ARGS__);                                \
+      c3_c* msg_c = strerror(errno);                                           \
+      if ( !msg_c ) {                                                          \
+        msg_c = "Unknown error";                                               \
+      };                                                                       \
+      fprintf(stderr,                                                          \
+              " (%s) [%s:%s():%d]\r\n",                                        \
+              msg_c,                                                           \
+              __FILE__,                                                        \
+              __func__,                                                        \
+              __LINE__);                                                       \
+      failure_action;                                                          \
+    }                                                                          \
+  } while ( 0 )
+
+//==============================================================================
+// Functions
+//==============================================================================
+
+c3_bile*
+c3_bile_open(const c3_path* const pax_u)
+{
+  if ( !pax_u ) {
+    goto fail;
+  }
+
+  c3_bile* bil_u = c3_calloc(sizeof(*bil_u));
+
+  bil_u->pax_u = c3_path_fp(pax_u);
+
+  try_syscall(bil_u->fid_i = open(bil_u->pax_u->str_c, O_RDWR | O_CREAT, 0666),
+              goto free_file,
+              "failed to open %s",
+              bil_u->pax_u->str_c);
+
+  struct stat buf_u;
+  try_syscall(fstat(bil_u->fid_i, &buf_u),
+              goto close_file,
+              "failed to get %s stats",
+              bil_u->pax_u->str_c);
+
+  bil_u->len_i = buf_u.st_size;
+  goto succeed;
+
+close_file:
+  close(bil_u->fid_i);
+free_file:
+  c3_free(bil_u->pax_u);
+  c3_free(bil_u);
+fail:
+  return NULL;
+
+succeed:
+  return bil_u;
+}
+
+//! @n (1) Seek to the end of the file to ensure append-only behavior.
+//! @n (2) Attempt the write again if we were interrupted by a signal.
+//! @n (3) Writing failed partway through, so clear clear the file of any data
+//!        from the partial write.
+c3_t
+c3_bile_put_raw(c3_bile* const    bil_u,
+                const void* const dat_v,
+                const size_t      len_i)
+{
+  if ( !dat_v ) {
+    goto fail;
+  }
+
+  if ( bil_u->off_i != bil_u->len_i ) { // (1)
+    try_syscall(lseek(bil_u->fid_i, bil_u->len_i, SEEK_SET), goto fail);
+    bil_u->off_i = bil_u->len_i;
+  }
+
+  const c3_y* dat_y = dat_v;
+  ssize_t     lef_i = len_i;
+  ssize_t     wri_i;
+  while ( 0 < lef_i ) {
+    if ( -1 == (wri_i = write(bil_u->fid_i, dat_y, lef_i)) ) {
+      if ( EINTR == errno ) { // (2)
+        continue;
+      }
+      else {
+        goto abandon_write;
+      }
+    }
+    else {
+      dat_y += wri_i;
+      lef_i -= wri_i;
+    }
+  }
+  bil_u->len_i += len_i;
+  bil_u->off_i = bil_u->len_i;
+  goto succeed;
+
+abandon_write: // (3)
+  try_syscall(truncate(bil_u->pax_u->str_c, bil_u->len_i), goto fail);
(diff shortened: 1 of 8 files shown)


joemfb:
Factoring this out as shared/reusable functionality sounds good. Some notes/questions:

Have you seen `u3_write_fd()` in `term.c`? That also retries on certain errors, and handles signal-interrupted syscalls. Not sure if all of that is universally applicable ...

I'm not sure that framing multiple buffers should happen at this layer. Snapshot patches in particular apply their own headers/framing, and I don't see any wins from nesting that. If we have use-cases for multiple framed entries, it seems like that framing could happen at a higher layer. I'm also not sure that host/network byte order swapping is right here, i think we should just use a consistent byte order and XOR/shift appropriately.

For pier metadata specifically, I'm tempted to just use strings for easy legibility via external tooling.

And I think this will be more general if we factor allocations out. We can often put the handler/control struct on the stack in synchronous code, and callers can employ their own allocation strategy (they are passing a length, after all).

mcevoypeter:
> Have you seen `u3_write_fd()` in `term.c`? That also retries on certain errors, and handles signal-interrupted syscalls. Not sure if all of that is universally applicable ...

Not until just now. I took a look, and it looks like the only relevant error for a blocking write to a file that is handled in `u3_write_fd()` but not in `try_write()` is `EINTR == errno`. The other two--`EAGAIN` and `EWOULDBLOCK`--are only relevant for file descriptors that have been marked as nonblocking.

> I'm not sure that framing multiple buffers should happen at this layer. Snapshot patches in particular apply their own headers/framing, and I don't see any wins from nesting that. If we have use-cases for multiple framed entries, it seems like that framing could happen at a higher layer.

I buy this. The service changes from "store an arbitrary number of binary blobs in this file" to "store a single binary blob in this file." If a caller wants to store multiple binary blobs, it's up to them to coalesce them into a single blob.

> I'm also not sure that host/network byte order swapping is right here, i think we should just use a consistent byte order and XOR/shift appropriately.

Converting to network byte order is using a consistent byte order. It's just that the byte twiddling is done for us. If we don't use `ntohl()` and `htonl()`, we'll need an endianness check to determine if we actually need to convert the byte order. `ntohl()` and `htonl()` seem like clear wins to me. That said, I think the point is moot because storing a single binary blob removes the need for length encoding.

> For pier metadata specifically, I'm tempted to just use strings for easy legibility via external tooling.

The single binary blob approach will also address this, at least when the binary blob is a C string (although the null terminator will show up at the end of the file).

> And I think this will be more general if we factor allocations out. We can often put the handler/control struct on the stack in synchronous code, and callers can employ their own allocation strategy (they are passing a length, after all).

Agreed.

I incorporated your above feedback so that the module only stores a single blob per file, retries reads/writes when interrupted by signals, and does no allocations.

pkg/urbit/c/bile.c
```diff
@@ -0,0 +1,162 @@
...
+  return NULL;
+
+succeed:
+  return bil_u;
+}
+
+//! @n (1) Overwrite any existing data.
+//! @n (2) Attempt the write again if we were interrupted by a signal.
+//! @n (3) Chop off any data that remains from a previous call.
+//! @n (4) Writing failed partway through, so clear the file.
```
joemfb, line 71: What's the advantage of placing these comments here and associating them with target code via indexing instead of just placing them over the target code?

pkg/urbit/include/c/bile.h
```diff
@@ -0,0 +1,134 @@
...
+//! field representations or names don't require changes at call sites.
+//!
+//! @param[in] bil_u  Binary file handle. Must not be NULL.
+//!
+//! @return  Path of binary file.
+static inline c3_c*
+c3_bile_path(c3_bile* const bil_u)
+{
+  return bil_u->pax_c;
+}
```
joemfb, line 61: As long as this library and its consumers are built together, this benefit seems pretty minimal. I don't object, but I'd just use the struct members.

Elsewhere, i've put `inline` functions in headers (`allocate.h`), and then instantiated (?) the functions in the implementation. The precise implications of `static inline` in a header is not clear to me -- is a copy of the function generated for every module that includes the header, or every module that includes a call site?

pkg/urbit/include/c/bile.h
```diff
@@ -0,0 +1,134 @@
...
+//! @param[out] dat_c  Destination address to write string to. Must have enough
+//!                    space for `c3_bile_len(bil_u)`  bytes.
+//!
+//! @return 0  String could not be read from file.
+//! @return 1  Success.
+static inline c3_t
+c3_bile_get_str(c3_bile* const bil_u, c3_c* const dat_c)
+{
+  return c3_bile_get_raw(bil_u, dat_c);
+}
```
joemfb, line 126: It'd be nice if we could ensure that the output is null-terminated here, but I think that would require a different api. In the meantime, can you add a comment noting that whether the string is null-terminated will depend on how the file was written?

pkg/urbit/include/c/defs.h
```diff
@@ -122,4 +122,30 @@
...
+      fprintf(stderr,                                                          \
+              " (%s) [%s:%s():%d]\r\n",                                        \
+              msg_c,                                                           \
+              __FILE__,                                                        \
+              __func__,                                                        \
+              __LINE__);                                                       \
+      failure_action;                                                          \
+    }                                                                          \
+  } while ( 0 )
+
```
joemfb, line 150: This is a pretty heavyweight macro, I'm not sure how much we want to encourage its use. I'm inclined to put it in `bile.c`, and maybe add some kind of internal/impl header if we need to share it across modules. But this gets to the broader question of a "public api" ...

joemfb (changes requested):
If we want to use this module for snapshots, we'll need to have partial read/write apis (maybe through a cursor). We may also need to track the length with `size_t`. But those should be straightforward to do later.

Some nits and questions:

pkg/urbit/c/bile.c
mcevoypeter: See the explanation in https://github.com/urbit/urbit/pull/5602#discussion_r824237092.

pkg/urbit/include/c/bile.h
mcevoypeter: Agreed that the benefit is minimal at the moment, but the cost is essentially zero, and so any benefit, no matter how small, feels worth it, especially because it will make refactoring easier in the future. Whether there *will be* any refactoring of this module in the future is besides the point.

If [this Stack Overflow answer](https://stackoverflow.com/questions/47819719/static-inline-functions-in-a-header-file) is to be believed, then declaring functions as `static inline` in a header file makes it more likely that the compiler inlines the function definitions at compile-time. If the compiler doesn't inline, then the `static` keyword prevents a multiple-definition error at link time.

pkg/urbit/include/c/defs.h
mcevoypeter: Yeah, I still haven't decided on how I feel about this one. I wrote `try_epoc() `and `try_lmdb()`, which are similar macros for calling `u3_epoc_*` and `mdb_*` functions (respectively), and they've turned out to be super handy, but syscall/libc calls may just be too heterogenenous to paper over with a single macro like this.

pkg/urbit/include/c/bile.h
mcevoypeter: Done!

pkg/urbit/include/c/bile.h
joemfb: The cost is precisely 13 lines! (It's a new module, do what you want ...)

That explanation of `static inline` header functions looks convincing enough for me.

pkg/urbit/include/c/defs.h
mcevoypeter: Moved into `bile.c`.

joemfb:
@mcevoypeter sounds like you've got some changes to this that haven't been pushed yet.

mcevoypeter:
Excuse all the force pushing. The most recent (and now only) commit simplifies the interface to `open`, `put_raw`, `get_raw`, and `close` and also adds support for partial reads and writes. The offset and length fields of `c3_bile` have also been converted to `size_t`.

mcevoypeter:

PR #5619, opened 2022-03-02
landscape: fix issue with adding duplicate ships to shipsearch pills

patosullivan:
Fixes https://github.com/urbit/landscape/issues/1139

Checks if pills array in ShipSearch includes the submitted ship name


arthyn (approved):
Looks good, although confusingly enough this should target next/groups haha. Tom actually had a good guide that he wrote because of that https://gist.github.com/tomholford/dcc55f19aff0b92f395e80555a340452

Just need to find a home for it

PR #5622, opened 2022-03-04
groups: Fix markdown styling around ship names in chat messages

patosullivan:
Fixes https://github.com/urbit/landscape/issues/1134

Note that this bypasses remark, which seems necessary for our purposes (and we can just style `<Mention />` directly), but I could be wrong.

To test: send messages like these in any chat:
```
yo *~zod*, what's up?
yo *~zod*
hey **~zod**
hey **~zod**, how've you been?
```

### DIFF
--- a/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
+++ b/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
@@ -34,6 +34,72 @@ interface GraphMentionNode {
   ship: string;
 }
 
+const addEmphasisToMention = (contents: Content[], content: Content, index: number) => {
+  const prevContent = contents[index - 1];
+  const nextContent = contents[index + 1];
+
+  if (
+    'text' in content &&
+    (content.text.trim() === '**' || content.text.trim() === '*' )
+  )  {
+    return {
+      text: ''
+    };
+  }
+  if(
+    'text' in content &&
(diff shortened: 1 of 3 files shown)


pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
```diff
@@ -34,6 +34,72 @@ interface GraphMentionNode {
   ship: string;
 }
 
+const addEmphasisToMention = (contents: Content[], content: Content, index: number) => {
+  const prevContent = contents[index - 1];
```
tomholford, line 38: Will this ever go out of range?

patosullivan: prevContent/nextContent being undefined is handled in the conditions on 63, 65, 75, and 79.

tomholford (approved):
@patosullivan and I just discussed this out band, now documenting here for the rest of the team:

Agreed that this function is janky but probably "fine" for now given the constraints (e.g., different implementations on mobile vs larger devices, an open bug in Remark that affects this behavior, and the eventual refactor of how Groups handle markdown).

We also discussed looping in @liam-fitzgerald since he mentioned a potential Markdown refactor previously.

liam-fitzgerald (approved):
LGTM!

PR #5626, opened 2022-03-06
%home -> %base

ryjm:
Mostly cosmetic, but also fixes `language-server` and parts of sundry neglected components (the nix build for solid pills, `aqua`, `gaze` exports, `+help`)


pkg/arvo/sys/arvo.hoon
```diff
@@ -1310,7 +1310,7 @@
       ::
       %+  turn
         (sort ~(tap by van.mod) |=([[a=@tas *] [b=@tas *]] (aor a b)))
-      =/  bem=beam  [[our %home da+now] /whey]  ::TODO  %base?
+      =/  bem=beam  [[our %base da+now] /whey]  ::TODO  %base?
```
ryjm, line 1313: The only part I'm concerned about. @mrdomino maybe you know, how does this work? If the beam mattered here I would expect `mass` to be broken, so it must not?

joemfb: It doesn't matter, and shouldn't be changed here. (It should be made to matter, and probably be `%$`, but that's a separate problem ...)

PR #5628, opened 2022-03-06
arvo: language-server desk support

ryjm:
Looks in all desks when attempting to build a file and picks the first match.

### DIFF
--- a/pkg/arvo/app/language-server.hoon
+++ b/pkg/arvo/app/language-server.hoon
@@ -5,7 +5,7 @@
     easy-print=language-server-easy-print,
     rune-snippet=language-server-rune-snippet,
     build=language-server-build,
-    default-agent, verb
+    default-agent, verb, dbug
 |%
 +$  card  card:agent:gall
 +$  lsp-req
@@ -44,6 +44,7 @@
   ==
 --
 ^-  agent:gall
+%-  agent:dbug
 %+  verb  |
 =|  state-zero
 =*  state  -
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/language-server.hoon
```diff
@@ -217,7 +218,7 @@
   :_  (give-rpc-notification (get-diagnostics uri))
   :*
     %pass
-    /commit
+    /commit˘
```
liam-fitzgerald, line 221: ???

pkg/arvo/app/language-server.hoon
```diff
@@ -240,18 +241,29 @@
...
   ?~  p.gift
-    [~ state]
+    [[~ state] %base]
   =.  builds
     (~(put by builds) uri q.r.u.p.gift)
   =.  ford-diagnostics
     (~(del by ford-diagnostics) uri)
-  =+  .^(=open:clay %cs /(scot %p our.bow)/home/(scot %da now.bow)/open/foo)
-  =/  =type  -:(open (uri-to-path:build uri))
+  =/  bek  byk.bow
```
liam-fitzgerald, line 253: This will have the wrong case, you'll want
```hoon
=/  bek  byk.bow(r da+now.bow)
```

liam-fitzgerald (changes requested):
Some nits, general idea looks good. Wondering if you have any ideas for the direction of cross-desk support in the LSP?

pkg/arvo/app/language-server.hoon
ryjm: what it's cute

liam-fitzgerald (approved):

PR #5629, opened 2022-03-08
arvo: kelvin

mrdomino:
can also use this as a grab bag for “stuff that should have gone in with `%khan` if you like, or happy to revert the merge and resubmit that PR @joemfb.


joemfb:
I agree that this is the right thing to do. If we decrement %lull, we must also decrement %zuse. Changing kelvin versions will force binary upgrades globally, and software distributors will have to update their distribution metadata (or their users will have to suspend the distribution and force the upgrade). So we'll want to schedule such a release carefully, and include any other appropriate pending changes to avoid unnecessary churn.

mrdomino:
added a couple minor fixes here

PR #5631, opened 2022-03-08
notify: disable notifications dismissal

liam-fitzgerald:
Disables automatic dismissal of push notifications in an attempt to
tighten the debugging space. This should prevent unintentional 404s. To
prevent a space leak we clear notifications once a week on a timer.

### DIFF
--- a/pkg/landscape/app/notify.hoon
+++ b/pkg/landscape/app/notify.hoon
@@ -13,6 +13,7 @@
       clients=(map ship binding=(unit @t))
       =whitelist
   ==
+++  clear-interval  ~d7
 ::
 +$  client-state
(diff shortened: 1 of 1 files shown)


Fang- (commented):
I wouldn't use a priority printf for this (plain `~&` is fine), and would just comment out the problematic code instead of doing a weird `?:`. Other than that, this looks fine.

liam-fitzgerald:
addressed!

Fang- (approved):
Sorry for the wait, this looks good!

(Realizing now that doing the `?:` way ensures the code remains compiling in the mean time. But either way is probably fine, I don't think we're expecting big refactors here anyway.)

PR #5632, opened 2022-03-10
groups: handle unsigged group links in join

patosullivan:
Fixes: https://github.com/urbit/landscape/issues/1404

Adds a new func to utils, "preSig", which takes a string and checks for a sig, returning that string if the sig already exists, otherwise returning a sigged string.

### DIFF
--- a/pkg/interface/src/logic/lib/util.tsx
+++ b/pkg/interface/src/logic/lib/util.tsx
@@ -229,6 +229,18 @@ export function deSig(ship: string): string {
   return ship.replace('~', '');
 }
 
+export function preSig(ship: string): string {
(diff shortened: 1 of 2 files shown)


arthyn (approved):
This is great, we could replace a lot of `~${deSig(...` with this. It should probably also go into the pkg/npm/api/lib/lib file which is the kinda replacement for this.

patosullivan:
> This is great, we could replace a lot of `~${deSig(...` with this. It should probably also go into the pkg/npm/api/lib/lib file which is the kinda replacement for this.

Cool, I'll open a separate PR adding it there.

Issue #2359, opened 2020-02-26
Ames: %dangling-bone error


belisarius222:
~litzod encountered this on the network.  This error should never happen, so I suspect either there's an internal Ames invariant that isn't being maintained, or something funny happened across a breach.
https://pastebin.com/RsG1H2M5

ixv:
I also have this issue on a ship I run. If anyone is interested in doing a debugging session about this, let me know.

philipcmonk:
For future reference one version of this I've seen several times is when a breach wasn't properly performed -- eg they didn't realize they couldn't just rm -rf, or they tried to breach but didn't have the "breach/discontinuous" box checked.

There may be another issue here, but the first thing is to check (ideally on both sides) what are the range of bones displayed in the debug dashboard.  If the sender has more (and higher) bones than the receiver, then chances are the receiver didn't properly breach.

philipcmonk:
This is likely the same issue as #3069 and #4737.

We've identified an instance of this that you can see here, along with discussion (cc @yosoyubik):

<img width="1219" alt="Screen Shot 2022-01-12 at 10 11 08 AM" src="https://user-images.githubusercontent.com/5553176/149195426-e0a1bfb7-d4e6-4d10-ab98-546364e4810c.png">
<img width="707" alt="Screen Shot 2022-01-12 at 10 11 42 AM" src="https://user-images.githubusercontent.com/5553176/149195436-8d76ccff-a43b-4d86-8385-1d4aaefa36fd.png">
<img width="964" alt="Screen Shot 2022-01-12 at 10 56 44 AM" src="https://user-images.githubusercontent.com/5553176/149195699-c1c31972-d561-48a5-bf56-b6556f690c18.png">
<img width="914" alt="Screen Shot 2022-01-12 at 10 56 54 AM" src="https://user-images.githubusercontent.com/5553176/149195706-7b0263ee-25e4-4131-a522-e2bd87b770e3.png">

zalberico:
We're tracking a fix for this dangling bone issue here: https://github.com/urbit/urbit/pull/5634

Three issues exist for this same problem:
- https://github.com/urbit/urbit/issues/2359
- https://github.com/urbit/urbit/issues/4564
- https://github.com/urbit/urbit/issues/4737

I'll keep the first open until the fix is deployed, and will close the newer two as duplicates.

zalberico:
We shipped the Gall Request Queue Fix release

yosoyubik:
[...]

PR #5634, opened 2022-03-11
ames: fix %dangling-bone error

yosoyubik:
This PR modifies ames' state to track the rift of a peer, including it in the wire that is used to interact with client vanes.

New wire format: `^-([@t ship=@ rift=@ bone=@ ~] /bone/~zod/0/1)`

Old wire formats are still parsed, but ames no-ops when dealing with them. In order to update the state properly, scrying the rifts out of `%jael`, ames regresses into the larval state, and upon receiving a `+call` or a `+take`, and if there are not more pending events to process, scries the rifts and updates the state, evolving then into adult ames.

This has been tested using the included unit tests and deploying it in a live ship, and two of its moons, that successfully apply the OTA and can continue messaging each other.

Next would be to test this in a ship that has the issue described here https://github.com/urbit/urbit/issues/2359. I've confirmed that this happens in the L2 `~roller`, but since that's an infrastructure-critical ship, it's probably better to test it somewhere else (although it's not really connected to the rest of the ames network, so maybe it's fine?)

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1220,8 +1220,8 @@
     ?~  parsed=(parse-bone-wire wire)
       ::  no-op
       ::
-      =/  =tape  "ames dropping malformed wire: {(spud wire)}"
-      (emit duct %pass /malformed-wire %d %flog %text tape)
+      ~>  %slog.0^leaf/"ames: dropping malformed wire: {(spud wire)}"
+      event-core
     ?>  ?=([@ her=ship *] u.parsed)
     =*  her          her.u.parsed
     =/  =peer-state  (got-peer-state her)
@@ -1233,20 +1233,19 @@
         ==
       ::  ignore events from an old rift
       ::
-      ?.  odd.veb  event-core
-      (log "ames dropping old rift wire: {(spud wire)}")
+      %-  %^  trace  odd.veb  her
+          |.("dropping old rift wire: {(spud wire)}")
+      event-core
     =/  =bone
       ?-(u.parsed [%new *] bone.u.parsed, [%old *] bone.u.parsed)
-    =?  event-core.peer-core  &(odd.veb ?=([%old *] u.parsed))
-      (log "ames parsing old wire format: {(spud wire)}")
+    =?  peer-core  ?=([%old *] u.parsed)
+      %-  %^  trace  odd.veb  her
+          |.("parsing old wire: {(spud wire)}")
+      peer-core
     ?~  error
       (send-ack bone)
     (send-nack bone u.error)
     ::
-    ++  log
-      |=  =tape
-      ^+  event-core
-      (emit duct %pass /on-take-done-parse-wire %d %flog %text tape)
     ::  if processing succeded, send positive ack packet and exit
     ::
     ++  send-ack
@@ -1525,9 +1524,9 @@
   ++  on-take-boon
     |=  [=wire payload=*]
     ^+  event-core
-    |^
     ?~  parsed=(parse-bone-wire wire)
-      (log "ames dropping malformed wire: {(spud wire)}")
+      ~>  %slog.0^leaf/"ames: dropping malformed wire: {(spud wire)}"
+      event-core
     ::
     ?>  ?=([@ her=ship *] u.parsed)
     =*  her          her.u.parsed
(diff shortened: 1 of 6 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -728,6 +782,7 @@
 ::
 =<  =*  adult-gate  .
     =|  queued-events=(qeu queued-event)
+    =|  cached-state=[?(%5 %~) ames-state-5]
```
belisarius222, line 785: Is there a reason this isn't 
```
=|  cached-state=(unit [%5 ames-state-5])
```
?

yosoyubik: None, I like yours more—I remember the type system yelling at me when I tried that, but just now I just added `?>  ?=(^ cached-state)` prior to the `state-5-to-6` update and it doesn't complain any more :)

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -965,6 +1057,28 @@
...
+    %-  ~(rut by peers.ames-state)
+    |=  [=ship ship-state=ship-state-5]
+    ^-  ^ship-state
+    ?.  ?=(%known -.ship-state)
+      ship-state
+    =/  peer-state=peer-state-5  +.ship-state
+    =|  =rift
+    =/  scry=(unit (unit cage))
+      (rof ~ %j `beam`[[our %rift %da now] /(scot %p ship)])
+    =?  rift  ?=([~ ~ ^] scry)
```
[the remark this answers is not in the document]
yosoyubik: I'm failing to see how this would crash... Looking in `%jael` the scry gives us the rift if it succeeds, `~` if it can't find the ship, or `[~ ~]` if we called it with a wrong beam.

yosoyubik:
@belisarius222 I've just pushed a [commit](https://github.com/urbit/urbit/pull/5634/commits/5406c039475a7d4d073569e1f474d6dfdf9dac51) that I think fixes a problem with the previous logic of dropping certain events based on the wire. From the commit:
```
Previously we were dropping events that used old
wires that lacked a rift in them. This seems a
bad behavior because we don't want to destroy a
flow that has not been processed by both ends.

Note: pending a fix to test-old-ames-wire
```
This makes sense to me, but if we have a flow with the dangling-bone, the event with the old wire will be then handled, and then crash again? Curious to see if @Fang-'s https://github.com/urbit/urbit/pull/5681 PR still passes the -breach-hi test with this latest commit.

belisarius222:
Good catch.  I don't have a good sense of what specifically could have failed given the old behavior, but it was indeed incorrect -- Ames should not drop signs unless they're from a ship at a stale rift.

yosoyubik:
Test performed here so far:

- Applying this PR to live ships that had previously communicate with each other works—they can still `|hi` each other
- Same for ships (e.g. moons) that boot directly with the PR synced.
- Breaching moons and L2 planets also works—the newly booted ships are able to restore communication.
- Breaching a moon, with `|ames-verb %odd` in the parent planet shows the wires with the old rift being dropped:

```
; ~navfex-libmun-hoppur-fastun has sunk
ames dropping old rift wire: /bone/~navfex-libmun-hoppur-fastun/2/1
ames dropping old rift wire: /bone/~navfex-libmun-hoppur-fastun/22/1
ames dropping old rift wire: /bone/~navfex-libmun-hoppur-fastun/16/1
ames dropping old rift wire: /bone/~tocsec-taptep-ravdul-dalnet/0/21
ames dropping old rift wire: /bone/~navfex-libmun-hoppur-fastun/0/1
ames dropping old rift wire: /bone/~rondun-lardyr/0/1
```
- These subscriptions (`/jael/public-keys /gall/use/ping/0w1.qOoqk/~navfex-libmun-hoppur-fastun/jael/~navfex-libmun-hoppur-fastun`) are not cleared by %jael, but as the log shows, are ignored and [dropped](https://github.com/urbit/urbit/blob/ff68ef06f7a07894757663d156cb7003e47c7f67/pkg/arvo/sys/vane/ames.hoon#L1240) by ames.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1105,31 +1217,62 @@
...
+    =?  event-core.peer-core  &(odd.veb ?=([%old *] u.parsed))
+      (log "ames parsing old wire format: {(spud wire)}")
+    ?~  error
+      (send-ack bone)
+    (send-nack bone u.error)
+    ::
+    ++  log
+      |=  =tape
+      ^+  event-core
+      (emit duct %pass /on-take-done-parse-wire %d %flog %text tape)
```
philipcmonk, line 1249: I think it's better to use `+trace`, because then it'll print even if the event crashes, and it'll print in order with other things from ames-verb or from `|verb`

philipcmonk (approved):
This looks pretty good, just one small change

pkg/arvo/sys/vane/ames.hoon
yosoyubik: Sounds good—I mostly added it so I could then capture that in the tests, but I guess I could look instead at how many events are emitted

yosoyubik:
closing in favor of https://github.com/urbit/urbit/pull/5732

PR #5635, opened 2022-03-11
Update /lib/ring/hoon for %jael scries.

rabsef-bicrym:
Also I took the liberty of making `build-signing-participants` use a `%lyfe` scry and remove unknowns from the set rather than crashing. This is a choice - happy to change it back the other way.


Fang- (approved):
This seems straightforward enough. Thanks!

I don't think anything depends on this, so we can theoretically merge this when/wherever. Trying to figure out the best path from the perspective of being good about releases though.

PR #5638, opened 2022-03-14
groups: do not notify user if mentioned in DM

patosullivan:
Fixes https://github.com/urbit/landscape/issues/1302

Note that the actual change is on line 474. For whatever reason, hoon.vim (I think?) decided to strip out trailing spaces on some lines on save.

### DIFF
--- a/pkg/landscape/app/hark-graph-hook.hoon
+++ b/pkg/landscape/app/hark-graph-hook.hoon
@@ -127,7 +127,7 @@
   ++  hark-graph-migrate
     |=  old=state-7:hist
     =|  cards=(list card)
-    |^ 
+    |^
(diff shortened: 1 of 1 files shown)


pkg/landscape/app/hark-graph-hook.hoon
```diff
@@ -471,6 +471,7 @@
       ^+  post-core
       ?.  should-notify  post-core
       =/  title=(list content:store)
+        ?:  =(title (crip "{(scow %p our.bowl)}/dm-inbox"))  title.kind
```
patosullivan, line 474: This is the only real change.

arthyn (approved):

PR #5640, opened 2022-03-15
grid: better searching software progress

arthyn:
Addresses urbit/landscape#1360 by making the state transitions more explicit, act across the multiple requests, and introduces a timeout so that people aren't waiting for more than 30s.


patosullivan (approved):
Nice! Much cleaner, and the timeout worked as expected when I searched for a host that I know had no software available.

PR #5641, opened 2022-03-16
hoon: hook up $+ for shorthand type rendering

Fang-:
Too often when dealing with big types the compiler traces and other such outputs become hard to read. Wrapping a type as $+(shorthand big-type) will now print #shorthand in place of the type.

To be clear, the changes here are simply connecting the last few dots of a system that's already implemented. I've been using this for a little bit in development, and it behaves fine.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -7781,7 +7781,7 @@
       [%bcgl *]  $(mod q.mod)
       [%bcgr *]  $(mod q.mod)
       [%bckt *]  $(mod q.mod)
-      [%bcls *]  $(mod q.mod)
+      [%bcls *]  [%note [%know p.mod] $(mod q.mod)]
       [%bcmc *]  ::  borrow sample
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -10996,6 +10997,11 @@
           [%unit *]
         =^  cox  gid  $(q.ham q.q.ham)
         :_(gid [%rose [" " (weld (trip p.q.ham) "(") ")"] cox ~])
+      ::
+          [%name *]
+        :_  gid
+        ?@  p.q.ham  (cat 3 '#' mark.p.q.ham)
+        (rap 3 '#' auth.p.q.ham (spat type.p.q.ham) ~)
```
joemfb, line 11004: Might want to have a separator between the name and path in the `^` case" something like `#name+/some/path` or `#name:/some/path` instead of `#name/some/path`.

pilfer-pandex (approved):

PR #5642, opened 2022-03-16
c3: add file path module for easy path manipulation

mcevoypeter:
Add `c3_path` type for easily creating and manipulating file paths. We currently use `snprintf` in many places to build up file paths, which is at best a hassle and at worst prone to buffer overflow bugs. The `c3_path` module provides a path abstraction that is lightweight and easy to use. Only a single heap allocation is required in many cases, and the module attempts to minimize the overhead of reallocations by doubling the newly needed allocation size upon reallocation. The `c3_path` type will be used in the new epoch-based event log system and also in an upcoming redesign of `unix.c`. It can also be used to simplify to `events.c`. 

Interface:
- `c3_path* c3_path_fa(const c3_c** toks_c, const size_t tok_i)`: create a path from an array of strings.
- `c3_path* c3_path_fv(const size_t tok_i, ...)`: create a path from a variadic list of string arguments.
- `void c3_path_push(c3_path* pax_u, const c3_c* const tok_c)`: push a new component onto the end of a path.
- `void c3_path_pop(c3_path* const pax_u)`: pop a component from the end of a path.

Tests:
See [path_tests.c](https://github.com/urbit/urbit/tree/peter/path/pkg/urbit/tests/path_tests.c).

### DIFF
--- a/pkg/urbit/c/path.c
+++ b/pkg/urbit/c/path.c
@@ -1,0 +1,132 @@
+//! @file path.c
+
+#include "c/path.h"
+
+#include "c/defs.h"
+
+//==============================================================================
+// Constants
+//==============================================================================
+
+//! Path separator.
+static const c3_c pax_sep_c[] = "/";
+
+//==============================================================================
+// Static functions
+//==============================================================================
+
+static inline c3_t
+_is_empty_token(const c3_c* const tok_c)
+{
+  return !tok_c || 0 == strcmp("", tok_c) || 0 == strcmp(".", tok_c);
+}
+
+static inline c3_t
+_is_root(const c3_c* const tok_c)
+{
+  return 0 == strcmp(pax_sep_c, tok_c);
+}
+
+static inline c3_t
+_is_parent(const c3_c* const tok_c)
+{
+  return 0 == strcmp("..", tok_c);
+}
+
+//==============================================================================
+// Functions
+//==============================================================================
+
+//! @n (1) Allocate 32 bytes for the path string as a default.
+c3_path*
+c3_path_fa(const c3_c** tok_c, const size_t tok_i)
+{
+  static const size_t cap_i = 32; // (1)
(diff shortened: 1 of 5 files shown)


pkg/urbit/c/path.c
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  size_t tok_i = sizeof(pax_sep_c) + strlen(tok_c);
+  if ( pax_u->siz_i - pax_u->len_i < tok_i ) {
+    size_t siz_i = 2 * (pax_u->siz_i + tok_i); // (2)
+    pax_u        = c3_realloc(pax_u, sizeof(*pax_u) + siz_i);
+    pax_u->siz_i = siz_i;
+  }
+
+  if ( 0 < pax_u->len_i && !_is_root(pax_u->str_c) ) { // (3)
```
joemfb, line 94: there's no `@n 3` corresponding comment

joemfb (changes requested):
This all looks reasonable.

We're going to need to review and come to terms with some of the new comment/header style conventions you're been establishing. I guess we can let that play out for some time before attempting to fully resolve those questions.

In a similar vein, I don't really like naming the c3 "library" directory `c/`. The conceit of the `c3` types is that they're "extending" the C language. So maybe this library is "extending" libc?

style nit:

mcevoypeter:
> We're going to need to review and come to terms with some of the new comment/header style conventions you're been establishing. I guess we can let that play out for some time before attempting to fully resolve those questions.

Definitely. No convention will be effective unless we're all on the same page. 

> In a similar vein, I don't really like naming the c3 "library" directory `c/`. The conceit of the `c3` types is that they're "extending" the C language. So maybe this library is "extending" libc?

So you're suggesting that modules like `path` and `bile` don't belong in `c3` proper but in a separate directory (and therefore library)? I've been thinking of `c3` as our C programming environment, and so it's felt permissible to put these primitive-ish types in `c3`.

mcevoypeter:

PR #5191, opened 2021-08-29
ames-flow-kill

h5gq3:
close and delete state from ames flow after %leave task
web+urbitgraph://group/~middev/the-forge/graph/~littel-wolfur/writs-7082/170141184505118342639209431323694333952

PR #5644, opened 2022-03-17
gall, ames: fix subscription desync (stage i)

yosoyubik:
This PR combines https://github.com/urbit/urbit/pull/5060 (thanks @belisarius222!) and https://github.com/urbit/urbit/pull/5191 (thanks @h5gq3!), modernizing both to fit the new post software-distribution world.

It compiles successfully on a fakezod (also on a live moon, but after suspending all other desks, committing the changes to [%/lib/dbug/hoon](https://github.com/urbit/urbit/commit/9cda1c660a8c2c104d5531b1edf3390db0873f30), and only reviving them once the `%base` desks has been updated, otherwise the commit fails to apply; not sure how this would be released to the network without suspending all non-base desks?), but it seems to have some networking issues:

- High-level tests:
  - Can't join new groups—you get to see the preview (number of participants... etc) but can't finish joining
  - Can message a previously known peer
    - But can't join its groups
  - Can join chats/notebooks from a previously joined group
  - ~~OTAs seem to work~~—the OTA starts, but the sponsee's acks are not acked
  
- Low-level tests
  - Can send/receive `-hi`—new flows would seem to work fine?
  - Can't send/receive `|hi`—from previously known peers, so old flows seem to not work, but this also happens with peers we haven't contacted before
  - Most lanes become indirect at some point—I've seen some direct with ~zod, the sponsor of the moon, but those then become indirect eventually until `~zod is ok`

Looking at ~debug, the forward flow for the `|hi` seems to be retaining all the `%plea`s. `num-slots:gauge` is always `0`, so there seems to be something wrong with the congestion control pump.

Screen Shot 2022-03-17 at 5 12 32 PM

Doing `-hi` works, both ways, but ~debug shows a high number of retries:

Screen Shot 2022-03-17 at 5 41 50 PM

<details>
  <summary>Seeing also some ames crashes in the parent:</summary>

  ```
  [%poke %crud]
  bar-stack=~[~[//ames]]
  call: failed
  /sys/vane/ames/hoon:<[880 3].[909 20]>
  /sys/vane/ames/hoon:<[882 3].[909 20]>
  /sys/vane/ames/hoon:<[883 3].[909 20]>
  /sys/vane/ames/hoon:<[885 3].[909 20]>
  /sys/vane/ames/hoon:<[886 5].[907 7]>
  /sys/vane/ames/hoon:<[889 5].[907 7]>
  /sys/vane/ames/hoon:<[890 7].[893 9]>
  /sys/vane/ames/hoon:<[892 16].[892 60]>
  /sys/vane/ames/hoon:<[1.243 5].[1.243 43]>
  /sys/vane/ames/hoon:<[1.249 5].[1.263 17]>
  /sys/vane/ames/hoon:<[1.251 5].[1.263 17]>
  /sys/vane/ames/hoon:<[1.254 5].[1.263 17]>
  /sys/vane/ames/hoon:<[1.330 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.331 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.334 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.342 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.343 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.344 5].[1.379 62]>
  [%ames-crash-on-packet-from ~banpec-holren-norsyr-torryn]
  /sys/vane/ames/hoon:<[1.345 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.346 7].[1.346 76]>
  /sys/vane/ames/hoon:<[428 3].[439 63]>
  /sys/vane/ames/hoon:<[429 3].[439 63]>
  /sys/vane/ames/hoon:<[431 3].[439 63]>
  /sys/vane/ames/hoon:<[433 3].[439 63]>
  /sys/vane/ames/hoon:<[434 3].[439 63]>
  /sys/vane/ames/hoon:<[435 3].[439 63]>
  /sys/vane/ames/hoon:<[436 3].[439 63]>
  [%ames-decrypt [~banpec-holren-norsyr-torryn ~norsyr-torryn ~] 5 170.005.604.959.354.425.787.202.162.006.145.340.004]
  /sys/vane/ames/hoon:<[437 3].[439 63]>
  /sys/vane/ames/hoon:<[438 3].[439 63]>
  /sys/vane/ames/hoon:<[542 3].[545 5]>
  /sys/vane/ames/hoon:<[542 3].[545 5]>
  /sys/vane/ames/hoon:<[542 3].[545 5]>
  /sys/vane/ames/hoon:<[544 12].[544 41]>
  /sys/vane/ames/hoon:<[544 7].[544 41]>
  /sys/vane/ames/hoon:<[563 15].[563 49]>
  /sys/vane/ames/hoon:<[563 34].[563 48]>
  /sys/vane/ames/hoon:<[563 34].[563 48]>
  /sys/vane/ames/hoon:<[563 44].[563 47]>
  /sys/vane/ames/hoon:<[563 40].[563 47]>
  hear: bail: %exit
  
  [%poke %hear]
  bar-stack=~[~[//ames]]
  call: failed
  /sys/vane/ames/hoon:<[880 3].[909 20]>
  /sys/vane/ames/hoon:<[882 3].[909 20]>
  /sys/vane/ames/hoon:<[883 3].[909 20]>
  /sys/vane/ames/hoon:<[885 3].[909 20]>
  /sys/vane/ames/hoon:<[886 5].[907 7]>
  /sys/vane/ames/hoon:<[889 5].[907 7]>
  /sys/vane/ames/hoon:<[895 5].[907 7]>
  /sys/vane/ames/hoon:<[897 14].[897 53]>
  /sys/vane/ames/hoon:<[1.243 5].[1.243 43]>
  /sys/vane/ames/hoon:<[1.249 5].[1.263 17]>
  /sys/vane/ames/hoon:<[1.251 5].[1.263 17]>
  /sys/vane/ames/hoon:<[1.254 5].[1.263 17]>
  /sys/vane/ames/hoon:<[1.330 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.331 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.334 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.342 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.343 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.344 5].[1.379 62]>
  [%ames-crash-on-packet-from ~banpec-holren-norsyr-torryn]
  /sys/vane/ames/hoon:<[1.345 5].[1.379 62]>
  /sys/vane/ames/hoon:<[1.346 7].[1.346 76]>
  /sys/vane/ames/hoon:<[428 3].[439 63]>
  /sys/vane/ames/hoon:<[429 3].[439 63]>
  /sys/vane/ames/hoon:<[431 3].[439 63]>
  /sys/vane/ames/hoon:<[433 3].[439 63]>
  /sys/vane/ames/hoon:<[434 3].[439 63]>
  /sys/vane/ames/hoon:<[435 3].[439 63]>
  /sys/vane/ames/hoon:<[436 3].[439 63]>
  [%ames-decrypt [~banpec-holren-norsyr-torryn ~norsyr-torryn ~] 5 170.005.604.959.354.425.787.202.162.006.145.340.004]
  /sys/vane/ames/hoon:<[437 3].[439 63]>
  /sys/vane/ames/hoon:<[438 3].[439 63]>
  /sys/vane/ames/hoon:<[542 3].[545 5]>
  /sys/vane/ames/hoon:<[542 3].[545 5]>
  /sys/vane/ames/hoon:<[542 3].[545 5]>
  /sys/vane/ames/hoon:<[544 12].[544 41]>
  /sys/vane/ames/hoon:<[544 7].[544 41]>
  /sys/vane/ames/hoon:<[563 15].[563 49]>
  /sys/vane/ames/hoon:<[563 34].[563 48]>
  /sys/vane/ames/hoon:<[563 34].[563 48]>
  /sys/vane/ames/hoon:<[563 44].[563 47]>
  /sys/vane/ames/hoon:<[563 40].[563 47]>
  crud: bail: %exit
  ames: packet failed (1000 total)
  ```
</details>


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1113,10 +1212,12 @@
     ::  if processing succeded, send positive ack packet and exit
     ::
     ?~  error
-      abet:(run-message-sink:peer-core bone %done ok=%.y)
+    ?:  (~(has in closing.peer-state) bone)
```
belisarius222, line 1215: indent

also might want to refactor to
```
=/  cork=?  (~(has in closing.peer-state) bone)
abet:(run-message-sink:peer-core bone %done ok=%.y cork)
```

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2144,10 +2270,24 @@
         ::  not a nack-trace bone; relay ack to client vane
         ::
         (emit (got-duct bone) %give %done error)
+      ::  +on-pump-cork: kill flow on cork sender side
+      ::
+      ++  on-pump-cork
+        ^+  peer-core
+        =.  by-duct.ossuary.peer-state
```
belisarius222, line 2277: These should all be one big `=:` so it's one atomic state update.

arguser:
I don't mean to add noise to this PR but aside from the mentioned issue in #5060 I'm also experiencing what I've commented on https://github.com/urbit/landscape/issues/1372#issuecomment-1024769781

yosoyubik:
Looking a bit closer, it seems that intra-moon communication (two live moons, but also two fake galaxies) works fine, but parent-child doesn't—this is probably contingent because as described above, there are also issues with live non-related ships. When sending a `|hi` (after applying the changes here) the parent planet receives it and prints it in the console, but the ack doesn't reach the moon, so it keeps it in its `live` queue waiting to be acked, and subsequent `|hi`s are added to `queued-message-acks`:

Screen Shot 2022-03-18 at 3 26 07 PM

and something similar happens with `-hi`, so acking the first `%plea`s is creating the issue:
Screen Shot 2022-03-18 at 3 36 15 PM

belisarius222:
Huh, do we really need to modify the message ack data format?  If we send a %cork and get an ack, we can know by sequence number that the ack is for the %cork message.  Then we should be able to use the 'ok' bit in the ack to determine whether the remote performed the flow closing operation as requested.

As an aside, we should think through the error handling if the remote nacks the flow closing -- my first hunch is just do a local delete, ignore all future packets we receive on that flow, and track the fact that we closed this but the remote didn't.  That could lead to space leaks on old publishers, though.

belisarius222:
These latest commits all look good to me.

belisarius222:
> The state migration from ames-4 was incorrect, but we still need to support this if there's been a ship that has been offline for a while—note: I don't have any ships that are that far behind.

Is this still incorrect after these changes?  I'm also curious what's incorrect about it.  We don't have an easy way to guarantee that a ship will do two updates in a row, although one thing we could do would be to crash if trying to upgrade from 4 directly to 6, ideally with a message that says "go download version 5 first".  We could have a moon set up with a desk to download from once, so a user could download that source once, then go back to downloading from their sponsor.

belisarius222:
> this doesn't cause any problems later on when you are re-subscribing again, so the only thing is those messages making %dojo a bit loud.

We should hide this behind a debug flag so users can turn it on and off.  People hate having their dojo's cluttered up with frequent messages.

yosoyubik:
@belisarius222 thanks for the fedback!

> The state migration from ames-4 was incorrect,
> Is this still incorrect after these changes?

I think it should be fine now. The main issue was that ames-5-to-6 introduced a hole in the adult load: the migration was performed in the larval call, so if you state was ames-5 you would skip the adult load. If you had ames-4, you would do the state migration in the adult load that was previously there... and then crash, because there is no 5-to-6 update. Instead the last change here does 4-to-5 in the larval load, and then calls itself, which will do 5-to-6 in the larval call, and the directly metamorphose to the adult gate.

> We should hide this behind a debug flag

I like that, are you thinking something similar to what %ames does with `|ames-verb / |ames-sift`—instead of per ship, verbosity could be per agent.

This PR itself introduces some extra printfs, but there's also some patterns in landscape that might not be necessary, that also make %dojo's output very loud: for example %spider [%kicks](https://github.com/urbit/urbit/blob/master/pkg/arvo/app/spider.hoon#L585) a thread when it's done, but observe-hook also [%leaves](https://github.com/urbit/urbit/blob/master/pkg/landscape/app/observe-hook.hoon#L360) it, and because we use a lot of threads this prints a ton of: `got %kick for nonexistent subscription`

> Doesn't this change the type of $bowl:gall?

Yes, that's right; the boat is updated to include a nonce in the value of the map.

> have enough context to know this %done is for the %cork

Yeah I think that's possible, we can check that the bone is in `closing.peer-state` and that we run the pump with a `%near` task, so coming from a `%naxplanation`, and in that case, just produce `$peer-core` `on-pump-done`

zalberico:
@yosoyubik is going to kick off the dev stream process for this. Getting this live for tlon ships for testing prior to release. Jose to follow up with Philip about specifics for this.

zalberico:
There are some group hosts behind getting the previous OTA.
- The previous OTA has an PR required for this.
- If you have this fix, but the group host does not have it - you can get into a situation where the host gets into a loop of resubscribe subscriptions.

Are there specific large group hosts we can mitigate this with? @yosoyubik to try to handle the case where the group host is a version behind - fixing that is the most desirable way to handle this.

yosoyubik:
This PR has been updated to handle stage I of the deployment of the gall-rq-fix—see https://github.com/urbit/urbit/pull/5886 for state II. The reasoning behind splitting this in two is to mitigate any scenarios where due to resubscribing (e.g. getting `%kick`ed because of network congestion), the number of ames flows increases, causing memory issues. This PR allows any ship to understand %corks (ames tasks for killing ames flows) but doesn't actually send any cork at this point (this is a task for stage II) when leaving a subscription—all the cork handle logic and new flow creation is there but we bypass it by manually using a `=. 
 sub-nonce.yoke. 0`. This makes it so that most ships will be able to handle `%corks` when the subscribers start sending them.
 
Worth noticing that we can't control what ships get this OTA, so some could still experience memory issues if they are not able to close ames flows, but this will hopefully minimize the number of affected (if any) ships.

PS: tested locally on fakezods using toy pub/sub agents and Group join/leaving, next step is to put this on live moons and monitor for any issues.

zalberico:
@zalberico to follow up with @belisarius222 - I think we're in a place where we probably want to deploy this to the tlon internal pre-release moon to catch remaining hard to reproduce bugs.

yosoyubik:
The (extra) recorks are not succeeding because they already did (on the publisher side), so if we get a message for a bone that's already corked (in `+run-message-sink`) we just drop it and the subscriber is going to retry again.

belisarius222:
Sounds like even current subscribers end up getting nacks when leaving a group. That must be a bug. Also, it looks like this is sending two %leave’s, one on bone 88 and one on 92. I thought leaving a group was only supposed to send one %leave — also, there’s a %spider error in the second log, potentially implicating a thread somewhere in here.

Just to be sure the stage 1 changes aren’t inducing this somehow, it might be worth double-checking whether the same thing happens for old publishers on next/arvo.

Even if that is a bug, though, we shouldn’t be mixing up nacks on %leave’s with nacks on %cork’s. I’m not sure if that’s happening here or not (although Jose, maybe you know this already). One way to check would be to print the message sequence number for each cork and nack and see exactly what the publisher is nacking.

The “unsent message from the future” error is important to track down precisely. My guess is sequence numbers are getting messed up from the combination of corking and nacking, but that’s just a hunch.

—
~rovnys-ricfer
https://urbit.org

zalberico:
The gall request queue fix is working, but the killing (and subsequent automatic restarting) of subscription fails. We’re splitting this so we’ll ship the gall request queue fix without killing all subscriptions. The leaving a group issue was a user space bug that was fixed independently.

The gall request queue fix release is being deployed to the Canary release ship today and to the Tlon release ship when Ted returns EOW.

There will be a separate issue for handling killing subscriptions by agent or globally (possibly with a command).

Fang-:
Included in #5886, which got merged. Closing as such.

PR #5646, opened 2022-03-18
vere: bytecode hints & memory hints

DaneBettis:
This branch is based on /next/vere

I am adding a new hint %xray.  %meme will be added in my next PR, after this one makes it into master.

%xray reports the c-vere bytecode of the hoon expression it wraps.

%meme will report the current memory statistics at the moment it is encountered.
In other words, %meme doesn't wrap any computation, so much as it triggers a memory report be printed when it itself is evaluated. To see changes in memory stats over time, just wrap various hoon expression in %meme calls, and they will report in order as they are encountered.

Like %bout, %nara and %hela, %meme and %xray will optionally be able to print user-supplied data.


pkg/urbit/noun/nock.c
```diff
@@ -1665,6 +1672,176 @@ u3n_find(u3_noun key, u3_noun fol)
...
+  case KITB: case KITB+1:
+  case MITB: case MITB+1:
+  case HILB: case HILB+1:
+  case HINB: case HINB+1:
+    return 1;
+  default:
+    return 0;
+  }
+}
+
```
DaneBettis, line 1754: @joemfb I rewrote _is_indexed() as you suggested - what do you think?
Thumbs up?
Otherwise what's the first thing you'd suggest I change?

pkg/urbit/noun/nock.c
```diff
@@ -1665,6 +1672,176 @@ u3n_find(u3_noun key, u3_noun fol)
...
+/* _num_from_pog(): returns an unsigned int of variable size
+ *                  with a side effect of correctly incrementing ip_w
+ *                  defined as a macro to avoid further pointer manipulations
+ */
+#define _num_from_pog(go, pog, ip_w) (\
+  go == 4 ? _n_rewo(pog, &ip_w):      \
+  go == 2 ? _n_resh(pog, &ip_w):      \
+  pog[ip_w++])
+
```
DaneBettis, line 1763: @joemfb The goal here was to improve readability and code reuse without having to reason any further with pointer/reference passing or transformation. I also explain this in the in-line docs. 

It was originally just to aid my own reading of _slog_byecode(), and I am happy to rewrite it if a macro here isn't the right way, but I've come to like it.

What do you think?

pkg/urbit/noun/nock.c
```diff
@@ -1665,6 +1672,176 @@ u3n_find(u3_noun key, u3_noun fol)
...
+ *          and a noun fol to complie into a u3n_prog bytecode program,
+ *          slog the program as a string of bytecodes,
+ *          and free the temp program afterwards.
+ */
+void
+_xray(c3_l pri_l, u3_noun fol) {
+  u3n_prog* pog_u = _n_bite(fol);
+  _slog_bytecode(pri_l, pog_u);
+  _n_prog_free(pog_u);
+}
```
DaneBettis, line 1843: @joemfb I think this one is also pretty quick - any feedback?

pkg/urbit/noun/nock.c
```diff
@@ -1665,6 +1672,176 @@ u3n_find(u3_noun key, u3_noun fol)
...
+        }                                    //
+      }                                      //
+      strcat(str_c, "]");                    // add the closing brace
+    }
+  }
+  // replace the first leading space and append the last char to the string
+  str_c[0] = '{';
+  strcat(str_c, "}");
+  u3t_slog_cap(pri_l, u3i_string("bytecode"), u3i_string(str_c));
+}
```
DaneBettis, line 1831: _slog_bytecode()

pkg/urbit/noun/nock.c
```diff
@@ -1769,6 +1951,17 @@ _n_hint_fore(u3_cell hin, u3_noun bus, u3_noun* clu)
       *clu = u3_nul;
     } break;
 
+    case c3__xray : {
+      u3_noun pri, tan;
+      if ( c3y == u3r_cell(*clu, &pri, &tan) ) {
+        c3_l pri_l = c3y == u3a_is_cat(pri) ? pri : 0;
+        u3t_slog_cap(pri_l, u3i_string("bytecode of"), u3k(tan));
+        _xray(pri_l, fol);
```
joemfb, line 1959: This prints "bytecode" twice. The caption here should just be `u3k(tan)`.

pkg/urbit/noun/nock.c
```diff
@@ -1695,6 +1872,11 @@ _n_hilt_fore(u3_noun hin, u3_noun bus, u3_noun* out)
       *out = u3_nul;
     } break;
 
+    case c3__xray : {
+      _xray(0, fol);
+      *out = u3_nul;
+    } break;
```
joemfb, line 1878: This version of the hint shouldn't have any caption at all. I think that `_slog_bytecode()` just just produce `u3i_string()` (a trivial tank), and the different hint implementations should `u3t_slog()` or `u3t_slog_cap()` as appropriate.

pkg/urbit/noun/nock.c
```diff
@@ -495,14 +500,14 @@ _n_nock_on(u3_noun bus, u3_noun fol)
...
-  X(KITB, "kitb", &&do_kitb),  /* 93 */                                        \
-  X(KITS, "kits", &&do_kits),  /* 94 */                                        \
+  X(MUTB, "mutb", &&do_mutb),  /* 87: c3_b */                                  \
+  X(MUTS, "muts", &&do_muts),  /* 88: c3_s */                                  \
+  X(MITB, "mitb", &&do_mitb),  /* 89: c3_b */                                  \
+  X(MITS, "mits", &&do_mits),  /* 90: c3_s, seems unused in codebase */        \
+  X(KUTB, "kutb", &&do_kutb),  /* 91: c3_b */                                  \
+  X(KUTS, "kuts", &&do_kuts),  /* 92: c3_s */                                  \
+  X(KITB, "kitb", &&do_kitb),  /* 93: c3_b */                                  \
+  X(KITS, "kits", &&do_kits),  /* 94: c3_s, seems unused in codebase */        \
```
joemfb, line 510: Some of these appear to be unused, but they are actually used through the index-size overflow mechanism.

joemfb (changes requested):
I see a couple things to tweak here:

On the code style front: function names (`_cn_X`), block comments, variable names (`cvc_TYPE_LETTER`). when in doubt, try to match the convention of surrounding code.

jfranklin9000:
litteral -> literal

pkg/urbit/noun/nock.c
DaneBettis: rewrote so that _slog_bytecode just slogs, no slog with caption, thus the two call sites will render as expected and either render one caption or none at all.

pkg/urbit/noun/nock.c
DaneBettis: rewrote so that _slog_bytecode just slogs, no slog with caption, thus the two call sites will render as expected and either render one caption or none at all.

joemfb: I think we should go one step further, as I described.

- `_cn_slog_bytecode` should be renamed to `_cn_etch_bytecode()`, and simply return `u3i_string(str_c)`
- the case block here should call `u3t_slog(u3nc(0, _cn_etch_bytecode()))`
- the case block below should call `u3t_slog_cap()`

pkg/urbit/noun/nock.c
```diff
@@ -1769,6 +1927,17 @@ _n_hint_fore(u3_cell hin, u3_noun bus, u3_noun* clu)
       *clu = u3_nul;
     } break;
 
+    case c3__xray : {
+      u3_noun pri, tan;
+      if ( c3y == u3r_cell(*clu, &pri, &tan) ) {
+        c3_l pri_l = c3y == u3a_is_cat(pri) ? pri : 0;
+        u3t_slog_cap(pri_l, u3i_string("bytecode of"), u3k(tan));
+        _cn_xray(pri_l, fol);
+      }
```
joemfb, line 1936: The output here isn't quite right; `tan` should be the caption of the bytecode, not it's own output with a constant caption (`"bytecode of"`). Given the changes described above, this should call `u3t_slog_cap(pri_l, u3k(tan), _cn_etch_bytecode())`.

joemfb (changes requested):
This looks good, but we need to finish tweaking the caption implementation:

pkg/urbit/noun/nock.c
DaneBettis: I've fixed the code per your notes.
I've tested it and it seems to be working correctly.

joemfb (approved):
LGTM

PR #5647, opened 2022-03-20
roller: add batch-trimming functionality

yosoyubik:
If a batch gets bigger than a max size defined by the ethereum node
the raw transaction is sent to, the /ted/roller/send thread will crash and
the batch will be blocked, stopping any subsequent batches to be sent.

[This](https://github.com/urbit/urbit/pull/5647/commits/cf838fd1d7491cfa176e70146b709814377ce044) detects when the current batch reaches a certain threshold and only
includes transactions up to that point, moving the ones that are not sent
back to the pending queue, adjusting their history and finding status.

Currently the max size allow for a batch is 128 Kb, which seems to be the max
that infura expects—haven't been able to find anything official besides [this](https://github.com/ethjs/ethjs-query/issues/16#issuecomment-751398800)

PS: It looks like it's a [geth option](https://github.com/ethereum/go-ethereum/blob/master/core/tx_pool.go#L53)


yosoyubik:
This should be ready to go. It builds on top of https://github.com/urbit/urbit/pull/5552 and https://github.com/urbit/urbit/pull/5608. All three PRs are deployed to the live L2 roller.

The max size of 128 Kb per batch seems to be correct—we've hit that case a couple of times, and also went over the limit, which has worked fine—but @philipcmonk let me know if you have more info about that, beyond what [I found](https://github.com/ethereum/go-ethereum/blob/master/core/tx_pool.go#L53)

philipcmonk (approved):

PR #5650, opened 2022-03-21
tile-grid: prevent uninstalls from breaking view

arthyn:
This cleans up the keys stored for ordering so that if an app is uninstalled it gets removed from the list and also puts preventative measures in.


patosullivan (approved):
LGTM

PR #5651, opened 2022-03-21
vere: add http/s port flags

mrdomino:
Based off of `jo/khan-c3`, which already added `per_s` / `pes_s` to `u3_pier` (though on the write end.)


joemfb (changes requested):
Thanks, this looks good.

Now that we have long opts, I'm inclined to think we don't want short opts for everything. These seem like a good candidate for long-only opts. Maybe we can use motes for the int value in the `struct option` array?

joemfb (approved):
LGTM

PR #5652, opened 2022-03-21
clay, ford: Reduce memory usage

liam-fitzgerald:
This fixes a space leak where the entire ford/clay core would be
included in the ford cache. Heavily reduces memory usage by clay,
reducing total usage from 100-200MB to around 10MB per-desk.

Also heavily reduces the size of clay results, they now jam to about 1MB instead of ~20MB

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:c29577dc949ac0689ba3c97ad13b812ea6c3c6cc9d255b770d2da95fd9af84b9
-size 23121802
+oid sha256:5c963ce1b61ac0948971e34f963fec6afa960d7c74f32392306b411d8bffd4fd
+size 23236383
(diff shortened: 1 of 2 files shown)


Fang- (commented):
To be clear: reduces memory usage because the type from the vase no longer comes from(?) the local subject, but instead the smaller subject of `[..zuse the-noun]`? Does this mean other places storing vases are getting bitten by this as well?

The new `|mass` output for clay feels _very_ verbose, and with the way memory measurement works (not counting deduplicated nouns already accounted for elsewhere) it's not clear to me this is useful beyond "find the first thing that's bigger than expected". Perhaps mass rendering should be configurable to collapse things beyond a certain depth or w/e. Not sure I like this as-is, but also wouldn't hold the fix back because of this, we can always patch it later.

(Unrelated to this pr, also surprised at how mass selectively prints `%strings: B/0` vs `'strings': B/0` depending on `@tas` sanity. (And even `29546` for `%js`.) Could be nice to make this show consistently as `strings: B/0`.)

Also consider a more polite branch name next time. (;

belisarius222:
🤦‍♂️sorry I missed this in the original ford fusion work

good catch @liam-fitzgerald

philipcmonk:
Very nice!  `|mass` is probably too verbose for regular usage, maybe we can put these behind a flag in the code for now, and just change the source to see it (and eventually have a verbose setting in the scry path, controllable by the user per-vane, etc...)

PR #5653, opened 2022-03-21
grid: actually use settings and improve ordering correction

arthyn:
This PR addresses issues seen with persisting tile ordering. We were not actually using the settings store correctly so this addresses that which enables the ordering to persist. It also fixes issues where we were overriding previously set ordering due to not waiting for the settings to be loaded.

### DIFF
--- a/pkg/grid/src/app.tsx
+++ b/pkg/grid/src/app.tsx
@@ -10,7 +10,7 @@ import useContactState from './state/contact';
 import api from './state/api';
 import { useMedia } from './logic/useMedia';
 import { useHarkStore } from './state/hark';
-import { useTheme } from './state/settings';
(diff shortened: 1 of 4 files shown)


patosullivan:
I'm seeing a loop when I uninstall an app. I'm guessing it's something going on inside that useEffect in TileGrid, but I'm not exactly sure what it is. To reproduce, try installing %hodl from ~dister-datder-sonnet and uninstalling it

patosullivan:
The loop appears to occur on the deployed grid as well. Guessing this was actually introduced by the PR earlier this morning.

patosullivan:
Just pushed a fix PR. Sorry that I did not catch this in reviewing the earlier PR!

patosullivan (approved):
Ok, this all works as expected now.

PR #5623, opened 2022-03-04
hoon: move $cape, $wine definitions to layer four

jtobin:
The modernisation of syntax that occurred in a539d986 evidently had the perverse effect of ballooning the size of the type/nock pair produced by +mint in certain cases (namely, when compiling the compiler itself), entirely due to the presence of these definitions in the subject of +us, the prettyprinter core.

Some benchmarks:

```
edf82f compiler compiling edf82f compiler:
> (met 3 (jam jnock))
1.643.757

edf82f compiler compiling edf82f compiler w/L4 cape/wine:
> (met 3 (jam jnock))
508.725

edf82f compiler w/L4 cape/wine compiling edf82f compiler:
> (met 3 (jam jnock))
631.855

edf82f compiler w/L4 cape/wine compiling edf82f compiler w/L4 cape/wine:
> (met 3 (jam jnock))
508.707
```

The ballooning in the size of the produced type/Nock seems to be unnecessary, as ship operation using the modified compiler appears to be unaffected, and the test suite continues to pass.

PR #5659, opened 2022-03-23
pill: latest from next/arvo

mrdomino:
Pulls in next/arvo and #5623; may be undesirable or premature.


mrdomino:
(this does result in about 1MB smaller binaries, unsurprisingly.)

mrdomino:
see also #5658.

joemfb:
We'll want to handle this through the release process (or explicit preparation for it) -- currently unclear how much will go into the next kernel release.

PR #5656, opened 2022-03-22
hark: add %add-note JSON decoder

liam-fitzgerald:
Allows notifications to be created from the frontend, and adds api bindings for the same 
cc: @patosullivan

PR #5660, opened 2022-03-24
grid: new browser api toggles/settings

patosullivan:
Addresses: https://github.com/urbit/landscape/issues/1282

This PR changes Interface Settings in System Preferences to reflect the per-browser (via browser fingerprint) preferences of the user. 

Upon login the browser is fingerprinted and assigned an ID, that ID is stored in local storage. The settings-store for the user's ship is then checked to see if any browser settings preferences have been stored for that ID. If the user has not yet set preferences for protocol handling and notifications in that particular browser, a notification is sent prompting them to do so. Clicking on the notification opens up the Interface Settings tab in System Preferences. Toggling either one of those settings in that tab both makes the permissions request to the browser and persists the browserId and settings preferences to the user's ship.

In the future, we could list out preferences for multiple browsers.

The Do Not Disturb toggle in the Notifications section is also decoupled from the browser notifications setting. 

Note, this is dependent upon this PR for notifying the user: https://github.com/urbit/urbit/pull/5656


patosullivan:
> Still need to handle non-protocol links for HTTPS users. That work is incoming.

Actually, I think this should probably be a separate PR.

tomholford:
Overall this LGTM. When you have a moment, can you provide some additional detail on why fingerprinting + localStorage is needed (as opposed to just setting a value in localStorage)?

patosullivan:
> Overall this LGTM. When you have a moment, can you provide some additional detail on why fingerprinting + localStorage is needed (as opposed to just setting a value in localStorage)?

Right now it's fingerprinting + storing it the ship. So we're persisting the browser preferences for each browser in the user's ship. If they use a new browser we'll bug them to let them know they haven't set the preferences yet. Make sense?

pkg/grid/src/app.tsx
```diff
@@ -10,23 +12,79 @@ import useContactState from './state/contact';
...
+          addNote(
+            {
+              path: '/browser-settings',
+              place: { desk: 'garden', path: '/desk/garden' }
+            },
+            {
+              title: [{ text: 'Browser Preferences' }],
+              time: Date.now(),
+              content: [
+                { text: "You haven't set your preferences for this browser, set them now?" }
```
arthyn, line 74: We should get @urcades / @e-shee to review this language

arthyn (approved):
LGTM, let's make sure copy is reviewed before merge

tomholford (approved):
LGTM :+1:

pkg/grid/src/app.tsx
e-shee: Looks good.

PR #5661, opened 2022-03-24
graph-store: drop unnecessary logs

liam-fitzgerald:
Drops unnecessary logs for subscribers, to reduce memory usage. In local
testing, this incurred a 20% improvement in graph-store memory usage.
Note that publishers still retain the entire history, for backlog and
resubscribe purposes. Subscribers still retain a handful of updates so
that they can resubscribe upon kick correctly.

### DIFF
--- a/pkg/landscape/app/graph-store.hoon
+++ b/pkg/landscape/app/graph-store.hoon
@@ -10,16 +10,17 @@
       [%2 network:zero:store]
       [%3 network:one:store]
       [%4 network:store]
-      state-5
+      [%5 network:store]
+      state-6
   ==
-::
-+$  state-5  [%5 network:store]
+::-
++$  state-6  [%6 network:store]
 ++  orm      orm:store
 ++  orm-log  orm-log:store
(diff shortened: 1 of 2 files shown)


Fang-:
This looks like a one-time band-aid fix? Don't we want to keep the log size down consistently for non-hosts?

(Also, the unnecessity of the logs assumes you're never gonna do republishing, right? This is probably fine, considering we aren't doing that, but worth being aware of and thinking through.)

liam-fitzgerald:
Yeah it does assume that you aren't republishing. How would you go about making this recurring?

Fang-:
I was thinking of just limiting the size of the log whenever we add something to it, in places such as this:

https://github.com/urbit/urbit/blob/1a2a9ee8997190876f5287f1ce151139f2419658/pkg/landscape/app/graph-store.hoon#L211-L212

Probably factored into a `+put-logs` or w/e.

Fang- (approved):
This looks better, thanks!

PR #5663, opened 2022-03-25
term: extended session support, other improvements

Fang-:
Couple things included. Primarily:

- Drum and it's protocol (sole) have been updated to support separate sessions connecting to the same application.
  - Notably, this is a breaking change, even for agents that are wrapped by `/lib/shoe`. In most cases, simply accounting for the new type of `sole-id` is sufficient, semantics haven't changed.
- Related, in dojo, if a generator has a `drum-session=@ta` named argument, and the user does not supply a value manually, it gets auto-filled to be whatever session the generator is run from. This lets `|link` and `|unlink` behave as desired: for the current session only.
- Webterm has been updated to support using, creating and deleting sessions. Special thanks to @tomholford for making this happen!
- Touches up the `%belt` and `%blit` types (specifically, `%hit` and `%hop`) to use coordinate-ordering that's less confusing, and consistent with `%blew`. We now use `[x=@ y=@]` everywhere, instead of variations on `[row=@ column=@]` which was easy to misremember.

webterm-demo

Missing from this is support for non-default sessions in the runtime. I'm fine doing that in a separate PR at a later point in time, but could be convinced to roll it into this if desired.

Other work that could or should be done before merging this:
- [x] Do another full run-through of the upgrade from master to this. I have already seen that the new binary continues to work fine with the old dill, but should do due diligence.
- [x] Resizing in webterm is either finicky or doesn't work at all. Not super important, a page refresh will do, 
- [x] Now that we have some non-buffer real estate in webterm, we could consider adding a little (i) to fix the keybind discoverability issue highlighted by #5507.
- [x] See if we can make the text selection key in xterm.js consistent cross-platform.

Since none of the above is particularly invasive, this should be ready for review already.


Fang-:
Small note about the latest commit. While it makes pasting chunks of text a _lot_ faster[1], it does interact with dojo's input validation in a funny way[2]. Specifically, when pasting text with syntax errors in it, for every new chunk of pasted input, the dojo will complain about the syntax error, moving the cursor to its location, and causing the remainder of the text to be inserted in that position. This may result in garbled-looking input in some cases.

This ux problem should be resolved on dojo's end. May or may not take a stab at that in the context of this PR. (Scope is bloating again, I know...)


<details><summary>[1] fast-paste action sequence</summary>

term-speed-paste
</details>

<details><summary>[2] pasting random chunk of c code</summary>

term-speed-haste
</details>

pkg/interface/webterm/App.tsx
```diff
@@ -0,0 +1,94 @@
...
+import Buffer from './Buffer';
+import { DEFAULT_SESSION } from './constants';
+import { showSlog } from './lib/blit';
+import { InfoButton } from './InfoButton';
+
+export default function TermApp() {
+  const { names, selected } = useTermState();
+  const dark = useDark();
+
+  const initSessions = useCallback(async () => {
```
liam-fitzgerald, line 28: this doesn't need to be a closure i.e. you can pull it's definition outside the component

pkg/interface/webterm/lib/useAddSession.ts
```diff
@@ -0,0 +1,68 @@
...
+        state.selected = sessionName;
+        state.sessions[sessionName] = null;
+      });
+    } catch (error) {
+      console.log('unable to create session:', error);
+    }
+  }, [names]);
+
+  return {
+    addSession
```
liam-fitzgerald, line 66: return just this, instead of an object containing it, because it's nicer to performance

liam-fitzgerald (approved):
Minor nits, else LGTM

tomholford:
Before merging this, we will need to merge and publish #5904, and update the `package.json` in this project to use the latest `@urbit/api` package.

zalberico:
I want this to be one of the first things we release in the new year (probably right after #6144).

Some questions:
 - I think it requires a kelvin version change (what type and what do we have to communicate to app devs about it)
 - It currently has a lot failing tests - can we get the PR into a ready to merge state?
 - Is there anything else I should know that might block us merging this when we get back?

I want #6144 done when we get back, @philipcmonk and I should be able to team up on that and get it merged and ready in a single pairing session. If we get that out then we can be more confident about the UX around kelvin releases generally.

Fang-:
So I merged next/arvo back in... but forgot to do so into next/dill first, so now the diff is even bigger than it was...

@zalberico Answering your questions from the perspective of the entirety of next/dill + this pr:

> I think it requires a kelvin version change (what type and what do we have to communicate to app devs about it)

Yes. These changes require a new runtime. The new runtime is largely backwards compatible with the old arvo, but compatibility between the _old_ runtime and the _new_ arvo is Bad at best. So: we use a kelvin bump to prevent that.

> It currently has a lot failing tests - can we get the PR into a ready to merge state?

I'll make sure things are building and tests passing. idk about frontend tests though, last I heard I could just ignore those?

> Is there anything else I should know that might block us merging this when we get back?

Chat-cli was culled for Groups 2's sake. RIP. We need to bring it back. Shouldn't be too difficult. Not a hard blocker but certainly very nice to have.

zalberico:
> So I merged next/arvo back in... but forgot to do so into next/dill first, so now the diff is even bigger than it was...

@philipcmonk ^ would rebase have prevented this from being as hard to read?

@Fang- thanks for the update, I just sent out an email to the kernel team with a general overview of the next couple of weeks (writing one for runtime next). What do I/we need to do for the new runtime, is it primarily bumping the syskelvin? I think that's something we'll need to do prior to merging this right (so we can run it on binnec)?

Fang-:
next/dill is up to speed now, merged that back in, diff looking a lot cleaner.

I personally haven't touched the vere side of kelvin releases yet at all, but I think that's right. Runtime needs to signal support for the newer kelvin, so that arvo lets itself be updated to that kelvin.

Seeing some builds fail for nix cache access reasons. I thought we were past this, but apparently not. Will re-run, poke around a bit as needed.

Fang-:
Presently unclear to me where the CI failures are coming from, or how to effectively debug them. I don't think this or `next/dill` touch any of the CI logic itself. Some things about the way output gets rendered were changed. Maybe we need a `-t` flag when running urbit in CI? Doesn't seem always present rn. But also unclear where exactly to add that. Will look around further tomorrow.

Fang-:
Okay, I think CI is good now. "But failures!" Yes, but mingw flow is getting removed in #6190, and the "frontend-test" flow can be ignored and/or ripped out according to @arthyn, so this isn't hard-blocking anymore. (Obviously we'll want to get green checkmarks when this hits next/arvo, but I'm not too concerned about next/dill.)

zalberico:
Thanks - @Fang- , if @arthyn says 'front-end' test can be ripped out, let's rip it out.

zalberico:
@Fang- I'm still seeing that failing FE job here, is that because it needs to have the change merged in from master or something else?

Fang-:
Had to resolve conflicts on both `next/dill` and this branch _again_ for whatever reasons. Guessing it's because this just touched _a lot_ due to the `u3l_log` semantics change.

But I think we should be golden now. Let's see those green checkmarks in ~30 minutes. (:<

philipcmonk:
This works on my machine as well.  We still need to glob and update the kelvins.

zalberico:
If the new runtime repo is ready we'll do the runtime kelvin update there. Otherwise we'll use the current repo.

Then on Monday we can have internal people install the runtime with the kelvin bump and then merge and deploy the terminal changes internally.

What's needed to glob?

Fang-:
> What's needed to glob?

I can do this tomorrow.

lmk what the situation with the new runtime repo is. Since it's forked off this one I'm assuming I can just bring my branch and obliterate the non-vere things from it? (May also have to make sure the CI changes land there.)

Fang-:
Glob is in. Webterm should work out the box now, no more local `npm run start` shenanigans.

I threw kelvin bumps for everything into this PR. We can reapply on other repos as needed. It's maybe a little out of scope for this PR, so apologies if this muddies the water a bit.

@philipcmonk did you want to stamp approval on this? Then we can get this into next/dill, and get next/dill into... next/arvo? next/vere? master? All of the above??

Remaining work/ops is cutting the new binary, getting that out the door first, then doing comms around upcoming kelvin bump to devs, getting our own stuff updated (pending) with that, and then lastly releasing the arvo changes herein. (Of course sprinkling in pre-release steps as appropriate.)

PR #5665, opened 2022-03-25
groups: improve key load time and general load performance

arthyn:
This fixes urbit/landscape#1364 by using a scry instead of the subscription. It seems that for whatever reason sometimes the subscription gets hung up whereas the scry does not.

This also reworks some of the code-splitting done previously which ended up not being as much of a performance gain as I thought. Because we're starved for TCP connections in the initial load it actually ends up making rendering wait for those to become available which is not ideal. Instead, I've split out two of our heaviest libraries.

Because of the scry for graph keys, I was inspired to also make a new scry for unreads to eliminate some of the roundtrip time so that being in a group or DM will actually show you the correct order faster.


liam-fitzgerald (approved):
This is the best thing I've ever approved! LGTM!

PR #5668, opened 2022-03-25
repo: fix build on mingw

mrdomino:
Apparently at some point `make` went away on the mingw system. `mingw32-make` still exists and does the trick though.


joemfb (approved):
LGTM, thanks!

Issue #5442, opened 2021-11-22
spider: `+validate-mark` is run on local messages eg `%thread-done`


tinnus-napbus:
**Describe the bug**
`+validate-mark` is run on messages from the local ship as well as foreign ones. This results in Clay trying to build `%thread-done` and `%thread-fail` marks when you run child threads, and since they don't exist it fails.

**To Reproduce**
Steps to reproduce the behaviour:

parent.hoon:

```
/-  spider
/+  *strandio
=,  strand=strand:spider
^-  thread:spider
|=  arg=vase
=/  m  (strand ,vase)
^-  form:m
;<  =bowl:spider  bind:m  get-bowl
=/  tid  `@ta`(cat 3 'strand_' (scot %uv (sham %child eny.bowl)))
;<  ~             bind:m  (watch-our /awaiting/[tid] %spider /thread-result/[tid])
;<  ~             bind:m  %-  poke-our
                          :*  %spider
                              %spider-start
                              !>([`tid.bowl `tid byk.bowl %child !>(~)])
                          ==
;<  =cage         bind:m  (take-fact /awaiting/[tid])
;<  ~             bind:m  (take-kick /awaiting/[tid])
?+  p.cage  ~|([%strange-thread-result p.cage %child tid] !!)
  %thread-done  (pure:m q.cage)
  %thread-fail  (strand-fail !<([term tang] q.cage))
==
```

child.hoon:

```
/-  spider
/+  *strandio
=,  strand=strand:spider
^-  thread:spider
|=  arg=vase
=/  m  (strand ,vase)
^-  form:m
;<  hash=@uv  bind:m  (scry @uv /cz/base)
(pure:m !>(hash))
```

install both in `%base` and run `-parent`, observe it crashes due to not being able to build `%thread-done`:

```
clay: read-at-aeon fail [desk=%base care=%b case=[%da p=~2021.11.21..10.47.48..8f37] path=/thread-done]
clay: no files match /mar/thread-done/hoon
[%error-building-mark %thread-done]
[%error-building-dais %thread-done]
/sys/vane/clay/hoon:<[4.257 13].[4.257 34]>
/sys/vane/clay/hoon:<[4.245 7].[4.271 9]>
~zod:dojo> -parent
bail: 4

bail: 2
spider crashed, killing all strands: %fact
! closing subscription
! /cb/~zod/base/~2021.11.21..10.47.48..8f37/thread-done
[...]

PR #5670, opened 2022-03-25
spider: fix child threads

drbeefsupreme:
this is a band-aid for #5442.

adds `%thread-done` and `%thread-fail` marks so that child threads work.
also fixes `await-thread:strandio` and removes some blank lines from
`app/spider.hoon`

`%thread-done` loses the type of the result, so you'll need to use `;;` to
get it back. the right way to fix this is to have threads produce cages
instead of vases according to @philipcmonk

### DIFF
--- a/pkg/arvo/app/spider.hoon
+++ b/pkg/arvo/app/spider.hoon
@@ -101,7 +101,6 @@
     ~
   $(trie u.son, yarn t.yarn)
 ::
-::
 ++  has-yarn
   |=  [=trie =yarn]
   !=(~ (get-yarn trie yarn))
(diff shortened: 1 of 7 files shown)


drbeefsupreme:
I should point out that the new marks are in `pkg/arvo/mar` rather than `base-dev/mar`, with symlinks to it in `pkg/arvo/mar`. Not sure which is preferred.

belisarius222:
The marks should be in base-dev if other desks will want to import them. I haven't looked at this PR closely, but most spider-related code should probably be in base-dev, since it has lots of clients in other desks.

—
~rovnys-ricfer
https://urbit.org

philipcmonk (approved):
Looks good, thanks!

PR #5672, opened 2022-03-26
clay: updates %prep to work with raw pages

joemfb:
This PR tweaks the %clay blob-store priming in #5470, working with raw pages and recalculating lobes to prevent the possibility (however remote) of poisoning the blob-store.


Fang- (approved):
Yeah, this is probably safer. Looks good to me!

philipcmonk (approved):
Yeah this is good

PR #5673, opened 2022-03-26
hoon: fix +uno:by when map is null

philipcmonk:
This should be merged before next/arvo is released, because #5612 uses +uno.

This crashed at runtime when `a` is null because it tried to instantiate
`meg`, and that means bunting its argument, and that calls `node`, where
the assert failed.

Wet gates use the bunt of their formal argument, so we use that.


joemfb (approved):
LGTM

Issue #5674, opened 2022-03-26
cue: gives 0 on large atoms


philipcmonk:
There appears to be some threshold of size above which +cue overflows silently.  It should at least crash.

```hoon
> (cue (jam (bex (bex 30))))
0
```

I tracked it into `ur_bsr_rub_len`, which is definitely getting ` len` of 0 back from `ur_bsr_log` with no error, which is incorrect.

joemfb:
This bug is an overflow in the `+jam` jet:

```
> `@ux`(end 6 (jam (bex (bex 30))))
0x2.0000.0002
```

should be

```
> `@ux`(end 6 =/(a (bex (bex 30)) =/(b (met 0 a) =/(c (met 0 b) (lsh 0^1 (cat 0 (bex c) (end 0^(dec c) b)))))))
0x3.0000.0000
```

PR #5676, opened 2022-03-27
u3: fix integer overflow in +jam jet atom encoding

joemfb:
This is a rather unsatisfying fix for #5674, but it buys time for a refactoring of the `+jam` jet implementation to support larger atoms. It just makes `(jam (bex (bex 30)))` a `bail:fail`.

### DIFF
--- a/pkg/urbit/noun/serial.c
+++ b/pkg/urbit/noun/serial.c
@@ -19,7 +19,7 @@ struct _cs_jam_fib {
 
 /* _cs_jam_fib_grow(): reallocate buffer with fibonacci growth
 */
-static void
+static inline void
 _cs_jam_fib_grow(struct _cs_jam_fib* fib_u, c3_w mor_w)
 {
   c3_w wan_w = fib_u->bit_w + mor_w;
@@ -28,6 +28,7 @@ _cs_jam_fib_grow(struct _cs_jam_fib* fib_u, c3_w mor_w)
   //
   if ( wan_w < mor_w ) {
     u3m_bail(c3__fail);
+    return;
   }
 
   if ( wan_w > fib_u->a_w ) {
@@ -48,7 +49,7 @@ _cs_jam_fib_grow(struct _cs_jam_fib* fib_u, c3_w mor_w)
 
 /* _cs_jam_fib_chop(): chop [met_w] bits of [a] into [fib_u]
 */
-static void
+static inline void
 _cs_jam_fib_chop(struct _cs_jam_fib* fib_u, c3_w met_w, u3_noun a)
 {
   c3_w bit_w = fib_u->bit_w;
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
Yeah this is good for now

joemfb:
I've updated this to target #6007. `(jam (bex (bex 30)))` no longer bails. `(bex (bex 31))` already bails, but `=+((bex (bex 30)) (mul - -)` does not (very slowly), and `(jam =+((bex (bex 30)) (mul - -)))` also succeeds.

pkg/urbit/noun/serial.c
```diff
@@ -70,12 +71,51 @@ _cs_jam_fib_mat(struct _cs_jam_fib* fib_u, u3_noun a)
...
-    c3_w a_w = u3r_met(0, a);
-    c3_w b_w = c3_bits_word(a_w);
+    c3_w   a_w = u3r_met(0, a);
+    c3_w   b_w = c3_bits_word(a_w);
+    c3_w bit_w = fib_u->bit_w;
+
+    //  amortize overflow checks and reallocation
+    //
+    {
+      c3_w met_w = a_w + (2 * b_w);
```
mcevoypeter, line 81: Why not defer this variable assignment until after the overflow check immediately below?

pkg/urbit/noun/serial.c
```diff
@@ -70,12 +71,51 @@ _cs_jam_fib_mat(struct _cs_jam_fib* fib_u, u3_noun a)
...
+
+      if ( a_w > (UINT32_MAX - 64) ) {
+        u3m_bail(c3__fail);
+        return;
+      }
```
mcevoypeter, line 86: Can you add a comment explaining that `a_w` must be less than or equal to `UINT32_MAX - 64` so that there's space for `2 * b_w`, which is at most 64?

pkg/urbit/noun/serial.c
```diff
@@ -70,12 +71,51 @@ _cs_jam_fib_mat(struct _cs_jam_fib* fib_u, u3_noun a)
...
+      c3_w* buf_w = fib_u->sab_u->buf_w;
+
+      //  _cs_jam_fib_chop(fib_u, b_w+1, 1 << b_w);
+      //
+      {
+        c3_d dat_d = (c3_d)1 << b_w;
+        src_w[0]   = (c3_w)dat_d;
+        src_w[1]   = dat_d >> 32;
+
+        u3r_chop_words(0, 0, b_w + 1, bit_w, buf_w, 2, src_w);
```
mcevoypeter, line 103: This would be more clearly expressed as
```c
c3_w src_w[] = {
  (c3_w)dat_d,
  dat_d >> 32,
};
u3r_chop_words(0, 0, b_w + 1, bit_w, buf_w, sizeof(src_w) / sizeof(*src_w), src_w);
```
without any loss in performance because `sizeof` is evaluated at compile time.

pkg/urbit/noun/serial.c
```diff
@@ -70,12 +71,51 @@ _cs_jam_fib_mat(struct _cs_jam_fib* fib_u, u3_noun a)
...
+        bit_w += b_w + 1;
+      }
+
+      //  _cs_jam_fib_chop(fib_u, b_w-1, a_w);
+      //
+      {
+        src_w[0] = a_w;
+        u3r_chop_words(0, 0, b_w - 1, bit_w, buf_w, 1, src_w);
```
mcevoypeter, line 111:
```c
c3_w src_w[] = { a_w };
u3r_chop_words(0, 0, b_w - 1, bit_w, buf_w, sizeof(src_w) / sizeof(*src_w), src_w);
```

pkg/urbit/noun/serial.c
mcevoypeter: Or, if you're opposed to inline use of `sizeof`, add a `c3_arr_len` macro to `c/defs.h`:
```c
#define c3_arr_len(arr) (sizeof(arr) / sizeof(*(arr)))
```

mcevoypeter (changes requested):
Aside from the usual style nits, this looks good to me.

pkg/urbit/noun/serial.c
```diff
@@ -70,12 +71,51 @@ _cs_jam_fib_mat(struct _cs_jam_fib* fib_u, u3_noun a)
...
-    _cs_jam_fib_chop(fib_u, b_w-1, a_w & ((1 << (b_w-1)) - 1));
-    _cs_jam_fib_chop(fib_u, a_w, a);
+    {
+      c3_w  src_w[2];
+      c3_w* buf_w = fib_u->sab_u->buf_w;
+
+      //  _cs_jam_fib_chop(fib_u, b_w+1, 1 << b_w);
+      //
+      {
+        c3_d dat_d = (c3_d)1 << b_w;
```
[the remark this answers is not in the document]
joemfb: This is too much. The purpose of "style nits" in a PR review is to enforce the conventions of the codebase, not to advocate for arbitrary personal aesthetics.

pkg/urbit/noun/serial.c
joemfb: The line immediately prior to this calculates `a_w + (2 * b_w)`, `b_w` is the result of `c3_bits_word()`, and a word is 32 bits. There's no end to the amount of documentation required if all such relationships need be spelled out in prose.

pkg/urbit/noun/serial.c
joemfb: It's impossible to argue that `sizeof(src_w) / sizeof(*src_w)` is clearer than `2`, when the length in question is the length of a "double word" in words.

mcevoypeter: I disagree. That `src_w` has two elements and and that `2` is  an argument of `u3r_chop_words()` could be coincidental, and by using `sizeof()` in this way in place of `2` when calling `u3r_chop_words()`, you make that relationship explicit. 

I don't understand this opposition to what feels like plain common sense. There is no performance overhead, no added complexity, only a more explicitly communicated relationship between two different lines of code. I understand that magic numbers are used liberally throughout this code base, but using past behavior as a justification for present and future behavior is a tenuous argument at best.

ohAitch: A consistent good style is better than a consistent flawed style is better than no consistent style.

pkg/urbit/noun/serial.c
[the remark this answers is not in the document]
mcevoypeter: This was meant purely as a suggestion in case it was an option you hadn't considered. Note that I didn't say, "please use `1ULL` instead of `(c3_d)1`, but merely that "`1ULL` *might* be a better option". I don't have a preference one way or another but was merely providing an alternative perspective, which I understand to be one of the responsibilities of a PR reviewer.

pkg/urbit/noun/serial.c
mcevoypeter: It takes one line, literally 7 tokens. Put it inline for all I care. Again, the relationship between some of these things is obvious in the head of whoever is writing the code, but it's not immediately obvious to someone reading the code.
```c
// 2 * b_w is at most 64
```

jalehman:
This seems ready, but needs to retarget urbit/vere. @matthew-levan , want to bring this over the line?

joemfb:
This fix still can't be shipped without ensuring that pilots will not need to replay past the history of the jet mismatch in question. The planned epoch system accomplishes this.

PR #5681, opened 2022-03-30
aqua: make tests pass, faster

Fang-:
Updates aqua and its tests to run & pass again, slightly faster than before.

Aside from the expected small tweaks and changes to bring it up to speed with recent changes, we added two optimizations to help tests run faster:
- For "real" ships, aqua configures their Azimuth handler to update faster. This lowers the amount of time spent waiting for new blocks to process in pki-related tests.
- For both "real" and fake ships separately, we cache the results of the boot sequence. This way, when we start up the same ship afresh for later tests, we can skip processing the boot sequence entirely.

For `-ph-breach-hi`, this reduced the testing time from around 20 minutes down to around 7.

New tests and improvements to aqua's interface & testing utilities may follow separately.

(Notably, the breach-related tests only pass with #5634 merged in.)


joemfb:
This supersedes #5578.

pkg/arvo/ted/ph/moon-az.hoon
```diff
@@ -12,6 +12,7 @@
 ;<  ~  bind:m  (init-ship ~bud |)
 ;<  ~  bind:m  (init-ship ~marbud |)
 ;<  ~  bind:m  (init-ship ~linnup-torsyx |)
+::TODO  not supported!
```
philipcmonk, line 15: What's not supported, moons?

philipcmonk (approved):
Looks great

pkg/arvo/ted/ph/moon-az.hoon
Fang-: That's right. Went ahead and added (the dumbest possible) moon support. Seems to work, except that moons can't talk to non-sponsor galaxies.

PR #5682, opened 2022-03-30
groups: fix silent s3 failures

patosullivan:
This fixes https://github.com/urbit/landscape/issues/1411

To test: provide a bad s3 config.

Shows an exclamation point icon (and some helpful copy) when s3 failures occur in collections and chat:

image
image
image

Tooltip in chat on mobile:
localhost_9000_apps_landscape__landscape_ship__zod_test-group_resource_publish_ship__zod_test-chat-6902_new(iPhone 12 Pro)


patosullivan:
fyi, the test failure can be ignored, is related to this: https://github.com/urbit/urbit/pull/5680

arthyn (changes requested):
Just a few tweaks and it'll be ready I think

arthyn (approved):

PR #5686, opened 2022-03-31
interface: fix reconnect bugs

liam-fitzgerald:
- Properly handles and restarts connection upon receiving kicks
- Restarted (not resumed) connection now works correctly, I suspect due to closing over stale state


arthyn (approved):
LGTM!

PR #5688, opened 2022-04-01
groups: fix issue with wrapping urls in notes

patosullivan:
Fixes this: https://github.com/urbit/landscape/issues/1408

Now looks like this: 
image


tomholford (approved):
LGTM

PR #5693, opened 2022-04-06
grid: browser setting and notification prefs cleanup

arthyn:
- Fixes an issue where upon opening another browser your previous browser's settings would be cleared
- Makes some of the setting handling more consistent and added types
- Restores Do Not Disturb with its original intention I think, but updates the language
- Detects Notification feature to warn browsers that can't use it
- Fixes Lock icon sizing
- Fixes weird bug in Brave causing white flashes on semi-opaque white background in Dialog


patosullivan (approved):
This looks and works great for me.

PR #5694, opened 2022-04-06
clay: unify all trie implementations

philipcmonk:
We had trie operations independently implemented in +de in arvo,
+an:cloy in zuse, +zu in clay, lib/trie, and app/spider.  This unifies
them all into +de in arvo, aggregating the used operations.

This PR is on top of #5408, since it conflicts with other cleanup on that branch.


pkg/arvo/sys/arvo.hoon
```diff
@@ -417,29 +417,67 @@
...
+  ++  get
+    |=  pax=path
+    ^+  [path fat]
+    ?~  pax  [~ fil.fat]
+    =/  kid  (~(get by dir.fat) i.pax)
+    ?~  kid  [pax fil.fat]
+    =/  low  $(fat u.kid, pax t.pax)
+    ?~  +.low
+      [pax fil.fat]
+    low
```
joemfb, line 452: This may be *the iconic trie operation*, but I think it's worth considering that all our call sites use `+dip` instead. I'd be inclined to swap the names based on that alone (I realize this was hashed out elsewhere).

pkg/arvo/sys/arvo.hoon
```diff
@@ -417,29 +417,67 @@
 ++  de
   =|  fat=(axal)
   |@
+  ++  del
+    |=  pax=path
+    ?~  pax  [~ dir.fat]
+    =/  kid  (~(get by dir.fat) i.pax)
+    ?~  kid  fat
+    $(fat u.kid, pax t.pax)
```
joemfb, line 425: It looks like this is (almost) just `=+  (~(dip de fat) pax)  -(fil ~)`. Doesn't this need to rebuild the `axal` on the way up the stack?

pkg/arvo/sys/arvo.hoon
```diff
@@ -417,29 +417,67 @@
...
+  ::  Serialize subppath to map
+  ::
+  ++  mup
+    ^-  (map path _?>(?=(^ fil.fat) u.fil.fat))
+    (~(gas by *(map path _?>(?=(^ fil.fat) u.fil.fat))) tap)
+  ::
+  ++  put
+    |*  [pax=path dat=*]
+    =>  .(dat `_?>(?=(^ fil.fat) u.fil.fat)`dat)
+    ^+  fat
```
joemfb, line 477: Looks like I wrote this with wet recursion, I think it'd be better to have a `|-` here so it's clear that `dat` is only cast/inferred once.

joemfb (changes requested):
It's great to unify all this stuff. A couple nits:

pkg/arvo/sys/arvo.hoon
belisarius222: Yeah, this looks like it will produce just the child axal, not the mutated root axal.

Replacing the last line with this should produce the mutated root axal:

```hoon
[fil.fat (~(put by dir.fat) i.pax $(fat u.kid, pax t.pax))]
```

belisarius222: Actually, I think this line, taken from +lop, would allocate less:

```hoon
fat(dir (~(put by dir.fat) i.pax $(fat u.kid, pax t.pax)))
```

pkg/arvo/sys/arvo.hoon
```diff
@@ -417,29 +417,67 @@
...
+  ++  lop
+    |=  pax=path
+    ^+  fat
+    ?~  pax  fat
+    |-
+    ?~  t.pax  fat(dir (~(del by dir.fat) i.pax))
+    =/  kid  (~(get by dir.fat) i.pax)
+    ?~  kid  fat
+    fat(dir (~(put by dir.fat) i.pax $(fat u.kid, pax t.pax)))
+  ::  Serialize subppath to map
```
belisarius222, line 468: subppath -> subpath

belisarius222: I don't really understand the description, though.  Isn't this more like "convert trie to map from path to file"?  There's no subpath specified.

pkg/arvo/sys/arvo.hoon
philipcmonk: I think of `get` as getting an in item in the container, not returning an instance of the container itself, which is why `+an:cloy` and lib/trie used that for `+get`.  Something like `(~(get xx (~(put xx foo) bar baz)))` should generally be `baz`.  I guess `+get:to` is a partial counterexample, but it's still focused on the item, not the container.

Our +dip call-sites are mostly (5 out of the 8) referring to the item anyway, which is why +get feels natural for them.  I agree that it's strange to give the standard retrieval name to the "loose" get.  So maybe we really should have three:

- `+dip` descends into the axal, producing a new axal
- `+get` produces the item at that path, producing a unit item
- `+fit` produces the item at the longest existing prefix to the path, producing `[remaining-path (unit item)]`

I think that will be the least surprising interface

pkg/arvo/sys/arvo.hoon
philipcmonk: Weirdly, to make this work I have to add ``, pax `path`pax`` to the `=>` to make this work, else it chokes on the callsites with literal paths like `/sys/zuse`.  Probably still worth it

philipcmonk:
New commit should address this feedback

pkg/arvo/sys/arvo.hoon
```diff
@@ -417,30 +417,73 @@
...
+    =/  kid  (~(get by dir.fat) i.pax)
+    ?~  kid  [pax fil.fat]
+    =/  low  $(fat u.kid, pax t.pax)
+    ?~  +.low
+      [pax fil.fat]
+    low
+  ::
+  ++  has
+    |=  pax=path
+    !=(~ fil:(dip pax))
```
joemfb, line 461: This should use `+get`.

pkg/arvo/sys/arvo.hoon
```diff
@@ -417,30 +417,73 @@
...
+    |-
+    ?~  t.pax  fat(dir (~(del by dir.fat) i.pax))
+    =/  kid  (~(get by dir.fat) i.pax)
+    ?~  kid  fat
+    fat(dir (~(put by dir.fat) i.pax $(fat u.kid, pax t.pax)))
+  ::  Serialize to map
+  ::
+  ++  tar
+    ^-  (map path _?>(?=(^ fil.fat) u.fil.fat))
+    (~(gas by *(map path _?>(?=(^ fil.fat) u.fil.fat))) tap)
```
joemfb, line 477: This should be after `+tap`, and I think the `+gas:by` makes the cast unnecessary.

pkg/arvo/sys/arvo.hoon
```diff
@@ -25,7 +25,7 @@
 +$  arch  (axil @uvI)
 ++  axal
   |$  [item]
-  [fil=(unit item) dir=(map @ta $)] ::
+  [fil=(unit item) dir=(map @ta $)]
```
joemfb, line 28: Since you've already changed this line ... I think I was too cute with this, and the recursion should just be `(axal item)` (referencing `$` there depends on `|$`not having an hygienic expansion).

joemfb (changes requested):
scraping the bottom of the barrel for a few more nits:

pkg/arvo/sys/arvo.hoon
philipcmonk: That actually blows up the compiler, overflowing the stack.  I assume it's because list/lest/pole/tree have an explicit non-recursive case (`~`), and this doesn't.

joemfb: My bad, I only tested it in the dojo. It's obviously not pressing -- spec hygiene will be its own effort.

joemfb (approved):
LGTM

PR #5695, opened 2022-04-06
eth-watcher: implement optional to-block parameter

pkova:
As discussed out of band with @yosoyubik and @Fang-, I need this feature because my custom `/app/gaze` that produces events for the [network explorer](https://network.urbit.org/)  runs out of memory when processing a massive batch of logs.

### DIFF
--- a/pkg/arvo/app/eth-watcher.hoon
+++ b/pkg/arvo/app/eth-watcher.hoon
@@ -8,7 +8,7 @@
 =>  |%
     +$  card  card:agent:gall
     +$  app-state
-      $:  %5
+      $:  %6
           dogs=(map path watchdog)
       ==
     ::
@@ -133,14 +133,16 @@
   ::
   =?  old-state  ?=(%4 -.old-state)
     %-  (slog leaf+"upgrading eth-watcher from %4" ~)
-    ^-  app-state
+    ^-  app-state-5
     %=    old-state
         -  %5
         dogs
       %-  ~(run by dogs.old-state)
       |=  dog=watchdog-4
+      ^-  watchdog-5
       %=  dog
           -
+        ^-  config-5
         =,  -.dog
         [url eager refresh-rate timeout-time from contracts ~ topics]
       ::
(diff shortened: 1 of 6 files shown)


pkg/arvo/ted/eth-watcher.hoon
```diff
@@ -13,7 +13,7 @@
 =/  m  (strand:strandio ,vase)
 ^-  form:m
 ;<  =latest=block  bind:m  (get-latest-block:ethio url.pup)
-;<  pup=watchpup   bind:m  (zoom pup number.id.latest-block)
+;<  pup=watchpup   bind:m  (zoom pup ?~(to.pup number.id.latest-block u.to.pup))
```
Fang-, line 16: Not sure this is quite right. Two things:
- Since we cannot look at blocks beyond the latest existing block, I think you want this to be `=+(last=number.id.latest-block (min last (fall to.pup last))`.
- `+zoom` notably _approaches_ the given block, but doesn't actually reach it. This means that, as currently written, we will only get logs from up to `(sub to.pup zoom-margin)`.

Might be cleanest to lightly rewrite `+zoom` to take an additional `up-to=number:block` argument, including that in its `up-to-block` calculation, to ensure that it stays the margin away from the network's latest block, but does eventually reach up to exactly the block we specify.

Fang-: Also wondering if, once we fetched the `to.pup` block, we should signal "completion" of this watchpup somehow. Right now it'll probably stick around and continue checking the latest block number, but once it's out of range, it'll be out of range forever. Not a super heavy load, but would be nice to clean up if we can.

pkova: Fixed, and decided to check for completion on behn `%wake`, where we kill and decide not to restart the thread if the pup is complete.

pkg/arvo/app/eth-watcher.hoon
```diff
@@ -509,12 +564,16 @@
...
       ?.  ?&  ?=(^ running.dog)
             ::
-              %+  gth  now.bowl
-              (add since.u.running.dog timeout-time.dog)
+            ?|  %+  gth  now.bowl
+                (add since.u.running.dog timeout-time.dog)
+                ::
+                !(has-reached-to-block dog)
+            ==
+            ::
```
Fang-, line 576: I think the change below, which prevents us from starting threads that have already "finished", is sufficient. The `number.watchdog` isn't going to change until we hear a result from the thread, which happens once the "last" invocation finishes, so we don't ever be in a position where we still have a thread running that's doing extraneous work.

tldr: pretty sure this change is not needed.

pkg/arvo/ted/eth-watcher.hoon
```diff
@@ -79,15 +80,15 @@
...
+  |=  [pup=watchpup =latest=number:block up-to=number:block]
   =/  m  (strand:strandio ,watchpup)
   ^-  form:m
   =/  zoom-margin=number:block  30
   =/  zoom-step=number:block    100.000
   ?:  (lth latest-number (add number.pup zoom-margin))
     (pure:m pup)
   =/  up-to-number=number:block
-    (min (add 10.000.000 number.pup) (sub latest-number zoom-margin))
+    (min (add 10.000.000 number.pup) (min up-to (sub latest-number zoom-margin)))
```
Fang-, line 91: Consider the following:

```suggestion
    ;:  min
      (add 10.000.000 number.pup)
      (sub latest-number zoom-margin)
      up-to
    ==
```

Fang- (approved):
This looks good, thank you!

I think the thread _killing_ you added isn't needed, since the other check precludes threads matching that condition to be running in the first place.  
Also have a suggestion for slightly improving legibility.

This should be good to go regardless, will merge by end of week. (:

pkg/arvo/ted/eth-watcher.hoon
pkova: What a sick rune, fixed.

pkg/arvo/app/eth-watcher.hoon
pkova: Fair enough, the helper arm can also go then.

PR #5696, opened 2022-04-06
interface: remove share profile chat button

dnmfarrell:
Removes the "Share private profile?" text and button that appear
at the top of the chat window when the ship has not opted in to sharing
their profile.

fixes [urbit/landscape#1195]([1195](https://github.com/urbit/landscape/issues/1195))

cc @nerveharp @urcades


dnmfarrell:
Maybe I should branch off next/landscape instead of master?

patosullivan (approved):
LGTM

patosullivan:
Thanks for the contribution!

arthyn:
For future reference this goes in next/groups confusingly enough, we hope to have this sorted better in the future.

PR #5700, opened 2022-04-07
groups: show group settings menu when pressing settings cog on mobile

patosullivan:
Fixes: https://github.com/urbit/landscape/issues/1426

Added the useMedia hook that we use in grid to groups, added an isMobile check to the GroupSwitcher in order to navigate to just the popover route on mobile so we can see the menu rather than going straight to the notifications section of the settings screen.


arthyn (changes requested):
It seems like we could simplify this a lot if we just defined the group settings path at the top switching on `isMobile` and then we could just set `navTo(groupSettingsPath)` and avoid having to have big ternaries.

arthyn (approved):
LGTM

PR #5702, opened 2022-04-08
build: wip depend on libecc in urcrypt

mrdomino:
libecc doesn't have an install target; in particular there's no
provision for making the header files available. Just hand-roll an rsync
to grab them.

TODO: pmnsh, probably need a way to depend on rsync (maybe also bc?)
from Windows.


mrdomino:
created by mistake

PR #5703, opened 2022-04-08
Po/remove prompt for browser settings notification

patosullivan:
Should wrap up https://github.com/urbit/landscape/issues/1282

We had decided (out of band) to remove the notification prompting users to set their notification/protocol handler settings on new browsers.


nerveharp:
Looks good, merging.

PR #5704, opened 2022-04-08
groups: fix join/leave notification toggle

nerveharp:
Adjusts the "Group Notifications" toggle copy to reflect reality.

image

Also makes some minor refinements to the StatelessAsyncToggle such that state changes don't cause layout jumps and/or rely on words like "Saved" or "Done," which may be confusing when used in non-matching contexts.

Off:
image

Loading:
image

Success (tooltip shown):
image

Error (tooltip shown):
image

On:
image

fixes urbit/landscape#1012


patosullivan (approved):
Nice, LGTM.

PR #5708, opened 2022-04-10
u3: improve %bout time printing

Fang-:
image

image

If we ever come up with a good `@dr` sub-second notation, we could consider using that instead. Until that time, this should be more readable than what we have now.

(Not very excited about how the interface here is subtly inconsistent with `u3a_print_memory`, but didn't want to make a bigger change. Maybe there's C magic I don't have that can bridge the gap, idk!)


joemfb (approved):
LGTM, TYFYS

/cc @DaneBettis

PR #5710, opened 2022-04-11
eth-watcher: remove obsolete zoom-margin handling

pkova:
The zoom margin is handled in [`/ted/eth-watcher`](https://github.com/urbit/urbit/blob/132592e32eea16eafdcc275bbd4652421d51b53e/pkg/arvo/ted/eth-watcher.hoon#L86), so this code is unnecessary. My work on #5695 revealed this problem since if you specify a `to-block` when subscribing to `/app/eth-watcher` the last batch of logs never gets released.


Fang- (approved):
Sorry for the wait! Was just double-checking with @philipcmonk. This looks good to both of us!

PR #5711, opened 2022-04-12
Interview eval: fix the long link layout issue [DNM]

yungcalibri:
Howdy. I'm ~normul-postem. This is a demo PR for the frontend dev position at Tlön. 

This is meant to address [Landscape issue 1337](https://github.com/urbit/landscape/issues/1337). 

# Warning
I wouldn't recommend merging this. I wanted to demonstrate my approach, first and foremost, and I have low context on the codebase. Might've broken something. 

# Fix
In specific implementation terms, the issue was caused by the use of `height` and `width` to control the dimensions of a flex child. They were tied up with the more appropriate flex dimensions, fighting one another. 

In a more philosophical sense, the issue is that this `<LinkBlockItem>` wants to control its own dimensions. For it to be flexible enough to appear in any context, it must not control its own dimensions with this kind of specificity. Instead, it should accommodate itself to whatever dimensions are dictated by the containing context. (I can specify height and width, but then I can't be relied upon to accord to the dimensions my flex container wants to give me; or vice versa.) So, to achieve that flexibility, I removed the `size` prop. I also removed the `<Center>` from this component's rendered output, and added it in the containing context instead. 

Now the `<LinkBlockItem>` isn't making demands about what kind of container it appears in, so we can give full flexibility back to the context in which it's used. 

# Bugs
- There's a double border on the internal `<Box>`. It's already pretty close to the deadline here, and I think this would be a two-second fix for someone who's not completely ignorant of the styling system as I am. 
- When resizing the page across breakpoints, there are occasional crashes with the error: `Cannot call a class as a function`. I doubt this is something I introduced, but felt it worth mentioning just in case.

# Insanity 
I'm gonna let my freak flag fly a little bit here, partly by way of introduction, and partly to draw attention to what I think are Urbit's unique strengths in Web UI. There are some choices made by frontend teams at almost every major SaaS company which I disagree with, and I see some of those repeated here. I wanted to lead with my spiel, so I wrote up an overlong document with the points I thought were important. I hope this can give all concerned some idea of my engineering principles, and maybe spark some discussion. 

[Here goes nothin'. 😄](https://gist.github.com/yungcalibri/b64dfe7a0e0e46135d9516a957b1ff9e)

Thanks for your time!


pkg/interface/src/views/apps/links/components/LinkDetail.tsx
```diff
@@ -27,12 +27,13 @@ export function LinkDetail(props: LinkDetailProps) {
...
+        <LinkBlockItem maxHeight="100%" node={node} />
+      </Center>
       <Col
-        minHeight="0"
-        flexShrink={1}
-        width={['100%', '100%', '350px']}
-        flexGrow={0}
+        flexBasis={['25%', '25%', '350px']}
+        flexShrink={0}
+        flexGrow={1}
```
arthyn, line 36: This seems to grow too much with enough content:

image

pkg/interface/src/views/apps/links/components/LinkDetail.tsx
```diff
@@ -27,12 +27,13 @@ export function LinkDetail(props: LinkDetailProps) {
   return (
     /*  @ts-ignore indio props?? */
     <Row height="100%" width="100%" flexDirection={['column', 'column', 'row']} {...rest}>
-      <LinkBlockItem minWidth="0" minHeight="0" height={['50%', '50%', '100%']} width={['100%', '100%', 'calc(100% - 350px)']} flexGrow={0} border={0} node={node} />
+      <Center flexBasis={['100%', '100%', '75%']} flexShrink={1} overflowY="scroll" >
```
arthyn, line 30: The overflow here is good for wider desktop views, but collapses too much on mobile, may need to set an explicit height or adjust the flex styles between this and the sidebar to prevent.

Screen Shot 2022-04-14 at 3 54 11 PM

arthyn, line 30: You added a `Center` here, but there's another place where `LinkBlockItem` is used that needs it as well.

arthyn (commented):
Thanks for the writeup and essay! I'm going to give my feedback here on the PR first, but I do want to write a response for your essay as well.

This PR does seem to solve the big issue here, but does fall short in a few ways. If you could address those issues, then we'd likely be able to merge it.

A good group for testing this is web+urbitgraph://group/~fabled-faster/interface-testing-facility specifically Test Collection A

yungcalibri:
Thanks for the feedback, it's extremely helpful. I'll be happy to work out these kinks so this can be merged.

It may take me a day or two as the wife and I have come down with something. I'll try to get these up before Monday morning.

pkg/interface/src/views/apps/links/components/LinkDetail.tsx
yungcalibri: The summary view is an interesting case; I'm not immediately sure what the best approach is. For now, I was able to address it by adding a flex container inside the box: see 72efd56b0b and 2f08d07527. 

image

pkg/interface/src/views/apps/links/components/LinkBlockItem.tsx
```diff
@@ -64,16 +63,11 @@ export function LinkBlockItem(props: LinkBlockItemProps & CenterProps) {
     history.push(`${pathname}/index${index}${search}`);
   };
   return (
-    <Center
+    <Box
       onClick={onClick}
-      border={border}
-      borderColor="lightGray"
       position="relative"
-      borderRadius="1"
```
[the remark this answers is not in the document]
yungcalibri: Re-added: f0a1724d97]

pkg/interface/src/views/apps/links/components/LinkDetail.tsx
```diff
@@ -27,12 +27,13 @@ export function LinkDetail(props: LinkDetailProps) {
...
+        <LinkBlockItem maxHeight="100%" node={node} />
```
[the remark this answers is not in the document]
yungcalibri: 99396e8133

yungcalibri:
Incidentally discovered a related issue in the link block view on the narrowest breakpoint. Comment counter bubble floats up above the title of the link when hovered. Not likely to be experienced often in production as devices with screens this narrow usually have no hover capability, but nonetheless. 

image

pkg/interface/src/views/apps/links/components/LinkDetail.tsx
yungcalibri: Tweaked the flex distribution in 4481660842 and c70ca0ad10. Still a bit quirky, but not broken. See gifs below for iphone and ipad screen dimensions, respectively.

It might be worth considering switching from the side-by-side layout here to a more conventional one with comments inline with the document itself, particularly for mobile web: two adjacent vertical scroll boxes is pretty confusing as compared to the old school blogspot-style thing.

pkg/interface/src/views/apps/links/components/LinkDetail.tsx
yungcalibri: Same issue as mentioned at line 30, see discussion there

PR #5714, opened 2022-04-13
clay: convert all blobs to pages

philipcmonk:
This converts the blob store from having deltas, directs, and
tombstones, to just having direct pages.  This simplifies a lot of code,
since we don't have to constantly ensure that deltas always have their
parent available.

This removes the hardcoded text diff logic from clay, which was
previously required for bootstrapping.

Over the wire, we handle both old and new requests and responses
transparently, so communication is normal in both directions across
ships which do or do not have this change.

Built on top of #5408 to take advantage of the cleanup there.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -144,10 +144,17 @@
       mut=(list (trel path lobe cage))                  ::  mutations
   ==                                                    ::
 ::
-::  Over-the-wire backfill request
+::  Over-the-wire backfill request/response
 ::
 +$  fill
   $%  [%0 =desk =lobe]
+      [%1 =desk =lobe]
+  ==
++$  fell
+  $%  [%direct p=lobe q=page]
+      [%delta p=lobe q=[p=mark q=lobe] r=page]
+      [%dead p=lobe ~]
+      [%1 peg=(unit page)]
   ==
 ::
 ::  Ford cache
@@ -160,10 +167,6 @@
       tubes=(map mars [res=tube dez=(set [dir=? =path])])
   ==
 ::
-::  Hash of a blob, for lookup in the object store (lat.ran)
-::
-+$  lobe  @uvI                                          ::  blob ref
-::
 ::  New desk data.
 ::
 ::  Sent to other ships to update them about a particular desk.  Includes a map
@@ -174,7 +177,7 @@
   $:  gar=(map aeon tako)                               ::  new ids
       let=aeon                                          ::  next id
       lar=(set yaki)                                    ::  new commits
-      bar=(set plop)                                    ::  new content
+      bar=~                                             ::  new content
   ==                                                    ::
 ::
 ::
@@ -198,15 +201,6 @@
       hez=(unit duct)                                   ::  sync duct
       cez=(map @ta crew)                                ::  permission groups
       pud=(unit [=desk =yoki])                          ::  pending update
-  ==                                                    ::
-::
-::  Object store.
-::
-::  Maps of commit hashes to commits and content hashes to content.
-::
-+$  rang                                                ::
-  $:  hut=(map tako yaki)                               ::
-      lat=(map lobe blob)                               ::
   ==                                                    ::
 ::
 ::  Unvalidated response to a request.
@@ -263,8 +257,7 @@
 +$  update-state
   $:  =duct
       =rave
-      have=(map lobe blob)
-      need=[let=(list lobe) old=(list lobe)]
+      need=(list lobe)
       nako=(qeu (unit nako))
       busy=_|
   ==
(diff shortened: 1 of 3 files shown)


ohAitch:
I suspect I am missing something here: does this mean if you edit hoon.hoon 20 times, this now makes your loom permanently bigger by 10MB of indirect atoms, i.e. 0.5% of its total lifetime capacity?

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3120,123 +3037,75 @@
...
+        missing
       =/  =norm  (~(gut by tom.dom) r.i.yakis nor.dom)
       =/  lobes=(list [=path =lobe])  ~(tap by q.i.yakis)
-      |-  ^-  [(set lobe) (set lobe) (map lobe blob)]
-      =*  blob-loop  $
+      |-  ^-  (set lobe)
+      =*  lobe-loop  $
       ?~  lobes
         yaki-loop(yakis t.yakis)
+      ?:  (~(has by lat.ran) lobe.i.lobes)
```
belisarius222, line 3060: This might be clearer as a `=?`, to consolidate the conditional and only recurse in one place instead of three.

```hoon
=?    missing
    ?&  !(~(has by lat.ran) lobe.i.lobes)
        !=([~ %|] (~(fit ^de norm) path.i.lobes))
    ==
  (~(put in missing) lobe.i.lobes)
lobe-loop(lobes t.lobes)
```

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -3935,7 +3783,7 @@
       ?~  x    ~
       ?~  u.x  [~ ~]
       ``[p.u.u.x !>(q.u.u.x)]
-    ::  +read-s: produce yaki or blob for given tako or lobe
+    ::  +read-s: produce miscellaneous
```
belisarius222, line 3786: lol yeah, this is a more accurate description

belisarius222:
@philipcmonk The code generally looks good to me, with a couple of nits that I left in comments.

ohAitch:
I mean my concern is not so much half a percent after a couple hours hacking on hoon.hoon, as most of the pier after a year of intermittently hacking on hoon.hoon. Urbit is supposed to be a forever computer, and this is moving in the opposite direction.

And like, one obvious goal with all of the recent commit message work would be to import the entire pkg/arvo history into your urbit, which I think this PR renders infeasible? I thought the goal of tombstones was always "accidentally added something giant" / "legal requirements", not "our revision-controlled file system crashes your ship if you actually use it that way".

It is additionally unclear to me why you'd keep diffs in the network protocol / marks, if the space savings is really insignificant. hoon.hoon is somewhat of an edge case, most of the files do not get 1000x+ write amplification from storing the whole thing(and by the same logic, how often do you update sys anyway). You could similarly use yet unwritten sufficiently-clever tricks to get it down to merely 250x write amplification.
(More seriously, something like an rsync rolling checksum partition scheme for indirect atoms could obviate the need for explicit diffs, either in the runtime or explicitly in clay if you want to reuse it for transmission efficiency. It seems advisable to implement performance optimizations before relying on them, however, instead of relying on them before implementing them)

What is the relation between build results and tombstones? Was the agent checkpointing a copy of itself to clay? Did the ford/clay integration remove the ford algorithm for invalidating the cache of stale build artefacts, for similar elegance reasons, and stick them in the blob store, so you can't tell them apart from normal files intended to be stored indefinitely?

ohAitch:
One obvious idea to float here is "clay update: remove the ability to access files with cases other than `now`, if you want version control build it in userspace", which to me seems like it has significantly less *user* mental overhead than  
"when using the urbit system, you will need to occasionally manually use tombstones to blast holes in the history, such that in normal use it is not reliably present, but we're still keeping it around for sufficiently small files kind of"

We could even remove the case from `.^`, it's not like anything else is using repeatable non-`now` ones. You ask clay for a dependency version and it returns the closest match it hasn't deleted yet.

philipcmonk:
The answer there is that %hoon shouldn't be a cord, for obvious reasons. A list of lines is also suboptimal, but a tree of lines (eg ordered map, queue, rope or whatever) will do quite well at structurally sharing.  This doesn't solve the problem of network use without a fairly advanced grain system, but it's a much better solution for local memory savings than explicit diffs.  In practice, solving memory use over the network should be addressed specifically instead of by a naive solution that was built long ago for a future that still hasn't come (it's been a couple years since anyone's created a %delta).

> What is the relation between build results and tombstones? Was the agent checkpointing a copy of itself to clay? Did the ford/clay integration remove the ford algorithm for invalidating the cache of stale build artefacts, for similar elegance reasons, and stick them in the blob store, so you can't tell them apart from normal files intended to be stored indefinitely?

No, the ford state was accidentally ending up in the ford cache because we used the current subject in some virtual environments instead of `..zuse` explicitly

ohAitch:
> it's been a couple years since anyone's created a %delta

Hm okay, if this PR is merely about restricting the types of values to the types they in practice have when deployed, that makes a lot more sense, though I'm baffled how we quite got to that point. Did the diff system ever work?
> the ford state was accidentally ending up in the ford cache

right so if there *is* a ford cache, with its own tombstoning logic("no tombstones required, you can just delete things, it's a cache"), this seems like it's a memory issue with ford separate from memory issues with the clay part of clay. which as opposed to being novel were introduced years ago, I suppose.

ohAitch:
> A list of lines is also suboptimal, but a tree of lines (eg ordered map, queue, rope or whatever) will do quite well at structurally sharing. 

A hoon source file, or dojo session, is a `qeu` of lines: you push new ones onto the end, the compiler takes a shallow snapshot and pops them from the front, when you've pushed in enough lines you have a functional urbit system :^)

pkg/arvo/sys/vane/clay.hoon
philipcmonk: good point

belisarius222:
@ohAitch I think I misled you about the ford memory use issue.

Ford does have a cache, and it does "promote" values of that cache from one revision to the next if its dependencies are unchanged.  The values in that cache are not stored in the blob store.  The Ford memory use bug, a fix for which can be found in https://github.com/urbit/urbit/pull/5652/files, was that Ford state was ending up inside certain build results, such as mark conversion gates.

belisarius222 (approved):
LGTM

ohAitch:
Yeah I think I missed the crucial bit about "not even a Clay problem, *and if it were*, we'd want tombstoning"

My revised understanding of this PR is the thing I'm objecting to is - the Ford Fusion merge? - somehow disabling the diff machinery, and not the PR itself which is fundamentally about, admitting that the machinery is not in use. Which does seem legitimate in that context, especially if it hasn't caused caused issues in practice in the intervening time.

PR #5715, opened 2022-04-13
Fix proposal for Groups: embeds break formatting context

afincato:
This PR proposes a fix and some design suggestions for https://github.com/urbit/landscape/issues/1273. So far *it focuses only on the chat portion* of the bug.

There are four commits:

- one to set URLs to plain text instead of the rich embed format
- one to correctly render blockquote content:  right now a blockquote with a text, a URL and some more text only displays the first text and the URL, and puts the rest of the content onto a new block; this commit joins everything together

The last two commits are design suggestions:

- one for re-integrating rich embed blocks in the chat / notebook, and to display them just below a block of text containing a URL; in case there are two or more URLs, all the rich embed blocks are appended in order after each text block
- one for not displaying "normal" URLs as rich embed blocks: eg, it only displays URLs to images, sound, video etc in this format because they have a useful preview content. This should be tested more in case of website with OG tags...

While reviewing the work and naively testing it in the notebook, I realised my PR butcher heavily how content is displayed in a notebook post. I figured the [culprit](https://github.com/afincato/urbit/blob/48c78298cea0bc734996c59f65bd7ada94c658f3/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx#L247-L255) and began to work on a patch, but time is running out and I prefer to share the work done so far than to shoot on what I've done!


pkg/interface/src/views/components/RemoteContent/index.tsx
```diff
@@ -65,9 +65,7 @@ export function RemoteContent(props: RemoteContentProps) {
   };
 
   const fallback = !renderUrl ? null : (
```
arthyn, line 67: Seems like this is unnecessary now

arthyn (commented):
Thanks for the writeup! Everything looked good in chats, notebooks were actually improved IMO, but there are two small issues that are hopefully easy to resolve:

Links on their own line seem to have an extra child causing more spacing than normal
image

Links by themselves in a in a list seem to be getting a smaller font size:
image

arthyn:
Going to close this one since you opened the other one.

PR #5716, opened 2022-04-14
remove false positive embed rendering in chat

rcrdlbl:
Fixes false positive rendering issue described in urbit/landscape#1363.

The current oembed provider we use, `noembed` hasn't been maintained in a while and is out of sync with the list in the `oembed-parser` library, from which the `hasProvider()` function comes from, which causes requests made to `noembed` to return a `provider not found` error. 

[noembed's own endpoint of supported sites](https://noembed.com/providers) also lists providers that aren't actually supported by noembed, so simply searching this endpoint instead of using `hasProvider()` would lead to the same false positive issues. Instead, I created `validOembedCheck`, which first tests to see if a provider exists and the URL is not from facebook or instagram (see note), then asks for the embed and checks to see if the object returned contains an `error` property. 

Previously, the only check that existed was a check for a `url` property on the returned embed object in `useEmbedState()`, but if noembed encounters an error, it returns an object containing an `error` property as well as the `url` that it failed to retrieve an embed for, which we were passing along to `LinkBlockItem` as if it were a valid embed.

### Notes on facebook/instagram embeds
Facebook and instagram now require people hitting their oEmbed endpoint to create a Facebook app that has access to this endpoint, and append said facebook app's authentication keys to their requests [(docs)](https://developers.facebook.com/docs/features-reference/oembed-read). Landscape by design is likely to be self-hosted, which presents technical challenges with regards to safely handling these API keys – we'd either have to create a store in the settings menu for self-hosted users to paste in keys from a Facebook app they created on their own, or we'd have to create our own oAuth endpoint.

This PR includes code that checks for the strings `facebook.com` or `instagram.com` and stops those URLs from being processed as embeds, and while we're still using `noembed` any edge cases should return an error. The latest release of `oembed-parser` (3.0.0) contains a function called `findProvider()` which returns an object containing a plaintext `providerName`. This function would allow us to include an easily readable array of "embed endpoints which cause trouble" in our code, and simply run `findProvider()` on any incoming URL and cross-reference the result in our checks. Unfortunately, the `oembed-parser` package doesn't include this function in its type definitions, meaning we can't import it into a typescript file.

### Potential routes for further improvement of embed handling

**Paid oAuth providers (embed.ly, microlink, etc)**

These seem to be more reliable and feature-rich than noembed. The issue lies with the fact that these are paid services that require authentication with each request, which leads to similar issues with API key safety as handling FB/IG embeds natively.

**Further utilizing oembed-parser**

`oembed-parser` includes a function called `extract()` which is intended to hit external oEmbed endpoints and return embed data. In the version of this library we currently use (1.4.5), this function seems to no longer work properly for every provider I tested it with. In version 3.0.0, it seems to be more reliable, however it still hits snags with certain oEmbed providers. It also uses oEmbed's main provider list, which seems to be scraped off the public web. We would likely have to create a whitelist of providers to avoid the chance of a malicious URL being picked up by oEmbed's provider list, as a malicious user could for example, paste said URL into a public chat and expose everyone who expands the embed to an XSS attack.

These caveats with along with the type definition issue leads me to have doubts in this package's overall reliability, and if we fully utilize it there's a big chance we'll have to take on maintenance tasks for an external package.

**Our own oEmbed handling solution**
I feel like this is the best option going forward – there are many ways this could be accomplished and a full write-up is out of scope of this PR message, but it could even be as simple as hosting a server that sends requests to one of the paid oEmbed endpoints.


rcrdlbl:
Requesting a review from @arthyn.

arthyn:
Will do, going to switch it's bases back and forth so github realizes all those commits are already in next/groups

arthyn (commented):
Thanks for the comprehensive write-up! This looks good. There is one edge case that we need to account for though, it seems that now on pages with lots of embeds we seem to get an infinite loop of fetching. Check the faster harder group sets collection for a good example.

rcrdlbl:
This commit should fix it – dropped in a `useRef()` to keep `useEmbed()` from getting called constantly!

arthyn:
I'm still seeing it constantly make thousands of requests in ~lidreg-dillut/faster-harder sets channel. Seems like both calls to `useEmbed` in `<RemoteContent>` and `<RemoteContentOembed>` is kinda ping-ponging the useEmbedState embeds field state so it gets emptied of the previous result.

rcrdlbl:
Alright, I've simplified the embed state logic, moving state modification out of fetch() and filling the embed state with suspender objects – in my testing this leads to faster load times and fewer requests made than even the master branch in a collection, while chat embeds remain unaffected. If you see any further potential edge cases come up let me know.

arthyn:
This looks great and so much faster! I found something with a link that's currently borked. Our suspend fn and embed needed to be changed to support it, but I don't think this was in scope for your bounty. 

I made this patch https://gist.github.com/arthyn/f4c68e2881dc7fe147000d65085b92b8

You can just apply it using: `git apply embed-updates.patch` from inside the repo. Also it looks like there's a small conflict, let me know if you need help resolving it.

After that I think this is done! Excited to see it live. Thank you!

rcrdlbl:
I think I may have messed up while resolving that conflict – in collections now, the expanded view of youtube links seems to show nothing, while in the HTML inspector the youtube embed iframe is clearly present, could you help me out with that?

arthyn:
So I was looking through this again, and started experiencing some weird thrashing when scrolling in chats. I didn't realize this suspender thing was actually specifically to use this pattern: https://17.reactjs.org/docs/concurrent-mode-suspense.html#approach-3-render-as-you-fetch-using-suspense

So since we're now calling `embed.read()` a level higher than before it's triggering a suspense throw, and then what's happening is it's falling back to the closest Suspense boundary and rerendering, and the closest boundary is way higher up causing the whole chatpane to rerender.

So we'll need to wrap a `<Suspense>` around the components calling `.read()`. It seems like we already kinda had a wrapper for this, `<AsyncFallback>`, so if we moved that up one level it should work. I just don't want people using `<LinkBlockItem>` or `<RemoteContent>` to have to know they need a Suspense around those.

rcrdlbl:
I wasn't able to notice any "weird thrashing", however I've encased RemoteContent and LinkBlockItem in an AsyncFallback. The chatpane seems to rerender less often when I run the profiler and open every single embed with this applied.

pkg/interface/src/views/components/RemoteContent/index.tsx
```diff
@@ -125,9 +128,11 @@ function RemoteContentInner(props: RemoteContentProps) {
     );
   } else if (isOembed && remoteContentPolicy.oembedShown) {
     return (
-      <AsyncFallback fallback={fallback}>
-        <RemoteContentOembed ref={embedRef} url={url} renderUrl={renderUrl} oembed={oembed} />
-      </AsyncFallback>
+      <Box mt={1} mb={2} flexShrink={0}>
+        <AsyncFallback fallback={fallback}>
+          <RemoteContentOembed ref={embedRef} url={url} renderUrl={renderUrl} />
```
arthyn, line 133: @rcrdlbl I think you missed `oembed={oembed}` in the props for this merge, but after you commit that this is good to merge!

rcrdlbl: Just pushed this commit!

arthyn (approved):
Looks good, thank you!

PR #5717, opened 2022-04-14
vere: removes urbit-worker

joemfb:
This PR adds all functionality from `urbit-worker` to the main `urbit` binary. It also updates all stateful utility commands (`pack`, `meld`, &c) to open the event log (and thereby aquire the pier lock) before changing state.

The release build still produces a tarball (which is probably worth preserving for a release or two, since people have scripts and automation around it). More to come soon.


pkg/urbit/daemon/main.c
```diff
@@ -1196,10 +1196,32 @@ _cw_utils(c3_i argc, c3_c* argv[])
...
 
+/* _main_bin_path(): copy binary path, ensuring ".exe" ending on windows
+*/
+static c3_c*
+_main_bin_path(c3_c* bin_c)
+{
+#ifdef U3_OS_mingw
+  c3_w len_w = strlen(bin_c);
+
+  if ( (4 <= len_w) && !strcmp(bin_c + len_w - 4, ".exe") ) {
```
mcevoypeter, line 1207: The `strcmp` return value check appears to be wrong, which seems to be causing the CI failure on MinGW. It should be `0 != strcmp(bin_c + len_w - 4), ".exe")`.

pkg/urbit/daemon/main.c
```diff
@@ -1196,10 +1196,32 @@ _cw_utils(c3_i argc, c3_c* argv[])
...
+    c3_c* pat_c = c3_malloc(5 + len_w);
+    sprintf(pat_c, "%s.exe", bin_c);
+    return pat_c;
+  }
+#endif
+
+  return strdup(bin_c);
+}
```
mcevoypeter, line 1215: This doesn't seem to work if `strlen(bin_c) < 4`. I know it's unlikely that we have binary names shorter than four characters, but I think we should still get this right for all cases. Based off preliminary testing, the following should work for all possible lengths of `bin_c`.
```c
#ifdef U3_OS_mingw
  static const c3_c exe_c[] = ".exe";
  size_t len_i = strlen(bin_c);

  c3_c* dot_c = strrchr(bin_c, '.');
  if ( !dot_c || 0 != strcmp(dot_c, exe_c) ) {
    len_i += sizeof(exe_c);
    c3_c* pat_c = c3_malloc(len_i);
    snprintf(pat_c, len_i, "%s%s", bin_c, exe_c);
    return pat_c;
  }
#endif
  return strdup(bin_c);
}
```

mcevoypeter (changes requested):
Aside from the two comments on `daemon/main.c:_main_bin_path()`, everything LGTM.

joemfb:
@mcevoypeter you're absolutely right about the binary file-extension handling being wrong. But that wasn't the issue in CI: the worker process was dying in `RegisterWaitForSingleObject()`, trying to setup ctrl-c handling. I've (temporarily) disabled that functionality in the windows build, and removed the file-extension munging. PTAL.

joemfb:
The windows crashes were a (latent) integer precision bug in `HANDLE` serialization and parsing -- thanks to @locpyl-tidnyd for tracking that down.

mcevoypeter (commented):
LGTM.

PR #5718, opened 2022-04-14
s3: adding v4 signatures for updated compatibility

arthyn:
This PR adds the latest signature version used in most s3 providers, many of which only support v4. Notably this allows people to use [Filebase](https://filebase.com) which has a generous free tier.


patosullivan (approved):
Not sure what wasn't working with filebase before (I had been using filebase for my s3 storage) but I can confirm that everything I expect to work is working here.

PR #5719, opened 2022-04-14
grid: fix leap ship autocomplete

patosullivan:
Fixes: https://github.com/urbit/landscape/issues/1421

also makes the display name visible separately, next to the shipname: 
image


arthyn (approved):
LGTM, what happens if the nickname is super long, especially on mobile?

patosullivan:
> LGTM, what happens if the nickname is super long, especially on mobile?

it wraps:
localhost_3000_apps_grid_(iPhone 12 Pro)

PR #5712, opened 2022-04-12
Images should use -webkit-optimize-contrast.

0x70b1a5:
The gains from this unobtrusive change are fairly straightforward.

before

after

PR #5720, opened 2022-04-14
Images should use -webkit-optimize-contrast.

0x70b1a5:
(Replicated from #5712, minus the extraneous changes.) 

The gains from this unobtrusive change are fairly straightforward.

before

after


liam-fitzgerald (approved):

nerveharp:
Very nice, thank you!

PR #5723, opened 2022-04-15
landscape: fix automatically watching group chats

liam-fitzgerald:
Adds functionality in %group-view and -group-create to automatically
watch group chats when created. This appears to have gotten lost around
the time of the hark-graph-hook and group-view rewrite


Fang- (approved):
Oh yeah that seems kinda important. This looks like it checks out.

PR #5724, opened 2022-04-15
landscape: remove new group from group-view on leave

liam-fitzgerald:
This fixes "ghost groups" if you leave a group soon after joining


Fang- (approved):
iow, just remove any "we are joining this rn" flag when we leave. Makes sense!

PR #5728, opened 2022-04-18
groups: fix edit profile

dillon:
Adds async/await to prevent pokes from being interrupted by a premature route change in the profile page.

The airlock pokes responsible for updating user profile values in `/~profile/[~zod]/edit` were not being awaited, and would therefore occasionally be interrupted by history.push(). I added async/await, switched to for loops for their simple async/await readability, and removed some redundant if/else logic for the `isPublic` key.

Fixes urbit/landscape#1199


arthyn (approved):
This looks great, thank you!

PR #5729, opened 2022-04-19
Fix proposal for Groups: embeds break formatting context and more

afincato:
This is a follow up to https://github.com/urbit/urbit/pull/5715. The commits in the previous PR were based off of the wrong branch (`next/landscape` instead of `next/groups`). I re-applied them to this branch in order to avoid excessive diff.

To summarise, this PR:

- change links to be plain URLs
- add rich-embed URL below each plain URL link when appropriate (eg when the link is audio, video, oEmbed, etc)
- allow links to be part of a blockquote without breaking out from the syntax

Furthermore, it addresses a few more small details reported in the previous PR convo (https://github.com/urbit/urbit/pull/5715):

- set URLs in a list item to keep the same text size as the rest of the text
- avoid to display extra nodes below single URLs

This last point, while implemented "visually", is not fully implemented: the extra space is caused by this bit of code (https://github.com/afincato/urbit/blob/db6130146bb9949e249a0b668f8836297340b55a/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx#L302-L321) searching for links in the graph AST, and given that at that point we don't know yet if the link will qualify as a rich-embed, we are adding already a root node. A more comprehensive solution would be to work in tandem with the RemoteContent component.

### DIFF
--- a/pkg/interface/src/views/components/RemoteContent/index.tsx
+++ b/pkg/interface/src/views/components/RemoteContent/index.tsx
@@ -65,9 +65,7 @@ export function RemoteContent(props: RemoteContentProps) {
   };
 
   const fallback = !renderUrl ? null : (
-    <RemoteContentWrapper {...wrapperProps}>
-      <TruncatedText>{url}</TruncatedText>
-    </RemoteContentWrapper>
+    ``
   );
 
   if (isImage && remoteContentPolicy.imageShown) {
--- a/pkg/interface/src/views/components/RemoteContent/index.tsx
+++ b/pkg/interface/src/views/components/RemoteContent/index.tsx
@@ -1,3 +1,6 @@
+import {
+  Box,
+} from '@tlon/indigo-react';
 import { hasProvider } from 'oembed-parser';
 import React from 'react';
 import useSettingsState from '~/logic/state/settings';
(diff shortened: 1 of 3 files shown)


afincato:
Commit 2f7bc1e makes the new blockquote design to work correctly also in notebooks.

The last problem left in notebooks has to do with links: in order to convert links to plain URLs, I switched URL elements from being `block` based to `inline` based (https://github.com/afincato/urbit/blob/db6130146bb9949e249a0b668f8836297340b55a/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx#L148-L150).

This works fine in the chat, but in a notebook the result is this

Screenshot 2022-04-19 at 15 22 41

eg, they are (correctly) rendered as inline nodes.

A first attempt was trying to wrap them inside a parent paragraph node, but that would add an extra node and it was not always working. What I think could work better, is a function that returns if the `url` should be set to `inline` or `block` on a url node basis. At the moment this seems out of my capacity, as it should probably be discussed with the team first? I'd be down to work on it in case.

arthyn:
I wasn't seeing the behavior above where it didn't add spacing around links so I think that is fine. Here's the only thing I saw lingering:

image


Seems like nested lists where the line is just a link, breaks on the first level of nesting. You can check it out here web+urbitgraph://group/~fabled-faster/interface-testing-facility/graph/~hastuc-dibtux/notebook-a-4589/170141184505596408090556188626318786560

Let me know if this is a trivial fix or not, either way the behavior is still better than what's currently in production

afincato:
OK, I will have a look! I missed that combination while testing.

afincato:
@arthyn I took a look at it this morning:

- your example with broken sub-list items is due to using tab instead of 2 spaces when indenting — the parser reads 4 spaces as a code block
- a sub-list item whose content is a link (or contains a link) breaks any other following sub-list item coming after it

For the second point, it seems to be very similar to the blockquote problem (for which I proposed a fix). In this case though, when logging what's going in `stitchInline`, I see the parse reads any sub-list item after a sub-list item which contains a link, *not* at the same indent level of the previous item anymore. I imagine then the problem lies at the parser level / when regexing and breaking text up in strings... Fundamentally changing how the link behaves brought the need to change some structural steps through the Graph code.

I also noticed currently a link in a list item is not wrapped inside a paragraph. I quick and dirty tried to do that and it adds a slightly more consistent vertical spacing. I can improve the code into a better shape and make a commit out of it.

LMK

afincato:
@arthyn I added a commit for wrapping a link in a list-item inside a paragraph tag, plus one clean up commit checking for linting errors.

I think now it should be all good!

arthyn (approved):
Looks good, thank you!

PR #5732, opened 2022-04-22
ames: %dangling-bone fix release

yosoyubik:
This PR supersedes https://github.com/urbit/urbit/pull/5634 —it's built on top of master instead of `next/arvo` so it's ready to be released.

It also adds an extra commit that removes the option for syncing `%plea`s targeted to `%ames`—it looks like that was added [here](https://github.com/urbit/urbit/commit/ea2a3b0f678fea0700af25e41e8eb84f0de9a6a0#diff-170aa64feef9c085358aed454b0f366f7654d665439d028585e299c693d5d0a2R1888), and was meant to be used as part of the ping-sponsor flow, which now uses the `:ping` app—keeping this causes a problem for deploying https://github.com/urbit/urbit/pull/5644 (publishers behind receiving the OTA will just send `%plea`s back to the subscriber when receiving a `%cork`—close ames-flow move—instead of `%nack`ing it)

(I tested just that commit on a live moon and haven't seen any weird behavior)

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:bcab0698de6efda1bbac54b0833da5e853bca058919110aa5668aa63fb40626e
-size 9392699
+oid sha256:564bc5acb0fede7612350b0e3b4c3d7597d9cbc9cfde20144550f15a450fb167
+size 9533002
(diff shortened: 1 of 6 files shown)


yosoyubik:
@belisarius222 @philipcmonk how are we feeling about this?

belisarius222 (approved):
lgtm

philipcmonk (approved):
Looks good!

PR #5738, opened 2022-04-26
grid: pasting shortcodes redirects to app page

matildepark:
Current behaviour when pasting, eg. `~paldev/pals` is to show the list
of all apps by ~paldev, even though the URL knows we're looking for pals
(eg. `/~paldev/apps/pals`). By reconstructing the URL to include the
host and desk twice, we can redirect straight to the app prompt, as
that's where that route exists.


matildepark:
(Retarget to branch as necessary — I don't know what the current process is...)

arthyn (approved):
Looks good, thanks!

PR #5739, opened 2022-04-27
groups: don't double sig the planet name on copy

dnmfarrell:
The copy function copies ~~zod instead of ~zod to the clipboard. This
removes the extra sig from the copy routine via deSig() incase the ship
string has no leading sig.

Fixes
[urbit/landscape#1434](https://github.com/urbit/landscape/issues/1434).


patosullivan (approved):

patosullivan:
Thank you!

PR #5741, opened 2022-04-27
npm: fix libraries module setup

arthyn:
This fixes an issue which causes the @urbit/api and @urbit/http-api modules to not be able to be used in CommonJS environments, because it treats the CJS builds as ESM.


tomholford (approved):
LGTM

PR #5744, opened 2022-05-03
u3: unify interpreter opcode information using X macro

mcevoypeter:
Use an X macro to unify the definition of opcodes, their names, and their indirect jumps in the bytecode interpreter into a single table. The likelihood of these bits of information changing is admittedly low, but I'm dipping my toes into the interpreter waters, and a simple hygiene task like this is a good first step. Plus, it eliminates 50 lines of code :)

Tested:
```console
$ urbit -F bus
<snip>
$ urbit bus
<snip>
```


mcevoypeter:
For context, X macros are a known pre-processor construct with a long history: https://en.wikipedia.org/wiki/X_Macro.

frodwith:
I like it.

PR #5745, opened 2022-05-04
ford: rewrite cache to share more builds

philipcmonk:
The current Ford cache is per-desk, keyed on the name of the build (eg a file at a certain path).  It's impossible to share a cache like this between desks because that name may refer to different things on different desks or at different revisions.

Since these caches aren't shared, they commonly hold exactly the same data, just generated independently.  For example, the same library used on different desks is built repeatedly, and the memory is not shared unless the user manually runs `|meld`.  For users with many apps installed, this adds significant memory pressure.

@belisarius222 and I designed a new cache which is keyed on the name of the build plus its dependencies.  This is all the input to a build except the standard library, so we when we match a key we don't need to know which desk it was on, so we can share this across desks.  When the standard library changes, we already cleared all the caches, and we continue to do that.

An important issue is reclaiming space from this cache.  Since this is a "true" cache, it's never incorrect to keep data in it, so you could use a heuristic such as "clear the whole thing every hour" or an LRU.  However, Ford has a long history of trying to use such heuristics and using exorbitant amounts of memory.  The primary innovation of the current ford cache is that its size is deterministic, and it stores no builds unless they could be generated from the head of its desk.

We extend these properties to the global cache by counting references and maintaining a per-desk set of references to builds which are still relevant to the head of that desk.  On every commit we inspect the per-desk set of references to determine which have been invalidated, and these are freed in the usual manner -- i.e. their refcount is decremented, and if it's now zero, then it's deleted from the cache and all its immediate dependencies are freed.

An alternative would be to garbage collect, which could be done on every commit to maintain determinism.  However, this scales with the size of the cache (thus, with the number of desks installed), whereas refcounting scales only with the desk in question.  Additionally, the cache is acyclic, and no manual refcounting is required -- there is precisely one place where references are gained and one where references are lost.

As a result, this cache has a "least upper bound" property: first, we minimize the number of rebuilds required; given that, we minimize the amount of memory required.  IOW, we throw away cache entries only when they become irrelevant.  An alternate approach would be a "greatest lower bound" -- throw away any cache entries that we're not certain to need.  This uses a bit less memory, but it results in more rebuilds, and it's a little more complex to implement, since it requires clients to "register" the builds they want to keep warm in the cache, even if those builds didn't become invalid.

Generating a cache key for this cache can be slow -- hundreds of milliseconds is not uncommon, and it scales with the number of transitive dependencies.  So we include a per-desk cache isomorphic to the one currently on mainnet.  This has sub-millisecond lookup speed, and is well understood.

So, to perform a build we:
- check if it's in our per-desk cache; if so, return it
- generate its dependencies
- check if it with these dependencies are in our global cache; if so, return it
- else build it

In every case, we add it to either cache if it wasn't already there, and if it wasn't in our per-desk set of references, we also add it there.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -1220,9 +1220,9 @@
       =^  =soak  nub
         ?^  got=(~(get by cache.nub) leak)
           =/  refs   ?:(spilt 0 1)
-          =/  tape-1  "ford: cache {<pour.leak>}: adding {<refs>}"
-          =/  tape-2  ", giving {<(add refs refs.u.got)>}"
-          %-  (slog leaf+(welp tape-1 tape-2) ~)
+          ::  %-  =/  tape-1  "ford: cache {<pour.leak>}: adding {<refs>}"
+          ::      =/  tape-2  ", giving {<(add refs refs.u.got)>}"
+          ::      (slog leaf+(welp tape-1 tape-2) ~)
           =?  cache.nub  !=(0 refs)
             (~(put by cache.nub) leak [(add refs refs.u.got) soak.u.got])
           [soak.u.got nub]
@@ -1236,9 +1236,9 @@
         ?~  deps
           [soak nub]
         =/  got  (~(got by cache.nub) i.deps)
-        %-  =/  tape-1  "ford: cache {<pour.leak>} for {<pour.i.deps>}"
-            =/  tape-2  ": bumping to ref {<refs.got>}"
-            (slog leaf+(welp tape-1 tape-2) ~)
+        ::  %-  =/  tape-1  "ford: cache {<pour.leak>} for {<pour.i.deps>}"
+        ::      =/  tape-2  ": bumping to ref {<refs.got>}"
+        ::      (slog leaf+(welp tape-1 tape-2) ~)
         =.  cache.nub  (~(put by cache.nub) i.deps got(refs +(refs.got)))
         $(deps t.deps)
       ?:  spilt
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -123,7 +123,7 @@
       tom=(map tako norm)                               ::  tomb policies
       nor=norm                                          ::  default policy
       mim=(map path mime)                               ::  mime cache
-      fod=ford-cache                                    ::  ford cache
+      fod=(set leak)                                    ::  ford cache
   ==                                                    ::
 ::
 ::  Commit state.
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -863,8 +923,14 @@
     ::    such as /path/file/hoon, but not /path/more/file/hoon.
     ::
     ++  build-directory
+      =/  gain  &
```
belisarius222, line 926: unused

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -149,21 +149,84 @@
...
+::
++$  pour
+  $%  [%file =path]
+      [%nave =mark]
+      [%dais =mark]
+      [%cast =mars]
+      [%tube =mars]
+      ::  leafs
+      ::
+      [%vale =path lob=(unit lobe)]
```
belisarius222, line 207: why is `lob` a unit?  I would have expected to always have a value for the file contents.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1066,7 +1131,161 @@
...
+        |=  file=(each page lobe)
+        ?-  -.file
+          %&  (page-to-lobe p.file)
+          %|  p.file
+        ==
+      ==
+    ::
+    ++  soak-cage
+      |=  [=soak nub=state]
+      ?>  ?=(%cage -.soak)
```
belisarius222, line 1185: What do you think of this?

```hoon
++  soak-cage  |=([s=soak n=state] [?>(?=(%cage -.s) cage.s) n])
++  soak-vase  |=([s=soak n=state] [?>(?=(%vase -.s) vase.s) n])
++  soak-dais  |=([s=soak n=state] [?>(?=(%dais -.s) dais.s) n])
++  soak-tube  |=([s=soak n=state] [?>(?=(%tube -.s) tube.s) n])
++  soak-arch  |=([s=soak n=state] [?>(?=(%arch -.s) dir.s) n])
```

philipcmonk: definitely better

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -1066,7 +1131,161 @@
...
       $(paz t.paz)
+    ::
+    ++  all-fits
+      |=  [=term suf=term]
+      ^-  (list path)
+      %+  turn  (segments suf)
+      |=  seg=path
+      [term (snoc seg %hoon)]
+    ::
+    ::  Gets a map of the data at the given path and all children of it.
```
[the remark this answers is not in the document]
philipcmonk: yeah there's a bunch of similar-but-not-identical instances of this pattern, which can be summed up as the +of operations

pkg/arvo/sys/vane/clay.hoon
philipcmonk: there was a version where it could depend on a file not existing, but that's gone now, so i'll change it

belisarius222 (approved):
LGTM

PR #5754, opened 2022-05-08
garden: add symbolic link to css.hoon

matildepark:
[create-landscape-app](https://github.com/urbit/create-landscape-app) ships .css in its bundle; this mark file being missing from a `%garden` source causes the build to crash. 

(The fact that my ship served it with `text/html` headers anyway after adding this seems like a separate issue or a fluke, so let's put this in place first...)


liam-fitzgerald (changes requested):
Dependencies should go in `*-dev` desks, so this PR should 
- move %css to `base-dev`
- create a symlink from copied file in both `arvo` and garden`

Please target `next/landscape` as well.

good catch, ty for the PR

matildepark:
Looks like next/landscape is behind master. Mind merging master into next/landscape @liam-fitzgerald?

liam-fitzgerald:
done

liam-fitzgerald (approved):

PR #5755, opened 2022-05-09
azimuth:  use %kick poke to restart subscriptions

yosoyubik:
Repurposes %kick to check for a possible missing subscription to /app/eth-watcher, notify %jael to resubscribe again to /app/azimuth and re-start downloading eth-logs from the last one we have previously downloaded, instead of going all the way back to the last processed block in the azimuth snapshot.

This replaces the previous flow of running this command: `:azimuth|watch 'INFURA_NODE' %default` to get up to date with the ethereum logs with `:azimuth|kick`

### DIFF
--- a/pkg/arvo/app/azimuth.hoon
+++ b/pkg/arvo/app/azimuth.hoon
@@ -39,6 +39,9 @@
       ::  %watch: configure node url and network
       ::
       [%watch url=@ta =net]
+      ::  %kick: re-start %azimuth subscriptions
+      ::
+      [%kick ~]
   ==
 ::
 +$  tagged-diff  [=id:block diff:naive]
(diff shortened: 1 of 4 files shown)


philipcmonk:
Maybe %kick should also check to see if eth-watcher is running, and if not then start it?

yosoyubik:
> Maybe %kick should also check to see if eth-watcher is running, and if not then start it?

Ah right, yes—will add it asap.

philipcmonk (approved):
looks good!

vvisigoth:
Hey @philipcmonk any idea when this can go out? Seeing a lot log processing issues on my ship and (I think) others.

vvisigoth:
ETA on this going out?

PR #5761, opened 2022-05-13
landscape/grid: Prevent cen from breaking react router in grid's leap

patosullivan:
fixes https://github.com/urbit/landscape/issues/1445


arthyn (approved):
Looks good I didn't even think about this lol

liam-fitzgerald (approved):

PR #5765, opened 2022-05-20
vere: adds binary upgrade tooling

joemfb:
vere: adds binary upgrade tooling

This PR integrates binary upgrade tooling into vere. It establishes a new model for running urbit, linking binaries into the pier on boot and upgrade. It presumes a new release model (not yet automated): a single (mostly) static binary for each architecture, organized into version/architecture specific directories on bootstrap.urbit.org, and grouped into separate "paces" (release channels). The experience is as follows:

##### boot

After a successful boot (performed as before), the binary is copied into the pier (`$pier/.bin/[$pace]/vere-v[$version][architecture]). If the platform/fs supports it, this copy will be structurally shared with the original, "reflink" style. Additionally, a hardlink is established at `$pier/.run` (`.exe`, on windows), pointing at that copy of the binary.

##### run

The pier can subsequently be run through the hardlink at `$pier/.run`, without requiring additional arguments. All the usual arguments are supported (XX except specifying the pier?).

##### upgrade

The binary can be upgraded -- together with a specified pier -- via `path/to/urbit --next path/to/pier` (or `$pier/.run --next`). This ensures that the snapshot is up-to-date, then checks for the next/latest binary (see below). If a newer release exists, it's downloaded into `$pier/.bin` and linked via `$pier/.run`. Combining this process into a single command enables us to strictly enforce invariants during migration; this will become critical soon with the addition of event log truncation, and enable simpler migrations for persistent state.

##### release structure

Releases are to be made to `bootstrap.urbit.org/vere/[$pace]/[$version]/vere-v[$version]-[$architecture]`. `$version` is the current `major.minor` release number, and `$architecture` is one of `X86_64-linux`, `X86_64-darwin`, or `X86_64-windows`. Obvious near-term additions include `aarch64-linux` (currently built by `~botter-nidnul`) and `aarch64-darwin` (currently supported via rosetta).

There is a latest release number in a text file `bootstrap.urbit.org/vere/[$pace]/last`. Additionally, each release can specify a version to which it should be specifically upgraded via `bootstrap.urbit.org/vere/[$pace]/[$version]/next`. This will allow us to stage migrations, if that becomes necessary.

The default pace is `live`. Users can opt for alternate paces by creating a text file at `$pier/.bin/pace` (anticipated paces include `soon`, for prerelease testing, and `often` for unstable/nightly-ish builds). For the smoothest user experience, it should be possible to "end" or "redirect" a pace with specific, non-version strings in the `/last` or `/next` version files.

##### next steps

- implement "pace" redirects
- trim whitespace when reading version/pace files
- implement additional CLI arguments
  - to disable automatic binary copying/linking
  - to expose the individual bootstrap/upgrade operations for admins
    - copy/link a specific binary in a pier
    - prep pier for upgrade
    - download next/specified binary version to specified directory
- automate creation of release artifacts
  - hosted runners should be added to build additional architectures via scheduled jobs
- test exhaustively


pkg/urbit/vere/king.c
```diff
@@ -231,54 +236,161 @@ _king_curl_alloc(void* dat_v, size_t uni_t, size_t mem_t, void* buf_v)
...
+    u3l_log("curl: error %s: HTTP %ld\n", url_c, cod_i);
+    ret_i = -2;
   }
-  if ( 300 <= cod_l ) {
-    u3l_log("error fetching %s: HTTP %ld\n", url_c, cod_l);
+
+  curl_easy_cleanup(cul_u);
+
+  *len_w = buf_u.len;
+  *hun_y = (c3_y*)buf_u.base;
```
mcevoypeter, line 278: Can we check that these aren't NULL before writing them? This isn't a hot codepath, and so the extra peace of mind is worth it in my opinion.

pkg/urbit/vere/king.c
```diff
@@ -231,54 +236,161 @@ _king_curl_alloc(void* dat_v, size_t uni_t, size_t mem_t, void* buf_v)
...
+
+  fid_i = c3_open(pat_c, O_RDONLY, 0644);
+
+  if ( (fid_i < 0) || (fstat(fid_i, &buf_u) < 0) ) {
+    c3_free(pat_c);
+    return strdup("live");
+  }
+
+  len_w = buf_u.st_size;
+  hun_y = c3_malloc(len_w);
```
mcevoypeter, line 324: Why not stack allocate here since it's freed before the function returns?

pkg/urbit/vere/king.c
```diff
@@ -231,54 +236,161 @@ _king_curl_alloc(void* dat_v, size_t uni_t, size_t mem_t, void* buf_v)
...
+
+  //  XX trim pat_c ?
+  //
+  return pat_c;
+}
+
+/* _king_get_next(): get next vere version string, if it exists.
+*/
+static c3_i
+_king_get_next(c3_c* pac_c, c3_c** out_c)
```
mcevoypeter, line 349: Can you document the return values in a top-level comment? It's easier to read a comment than to go hunting for the call site to figure out what the return values mean.

pkg/urbit/vere/king.c
```diff
@@ -231,54 +236,161 @@ _king_curl_alloc(void* dat_v, size_t uni_t, size_t mem_t, void* buf_v)
...
+  c3_c* ver_c;
+  c3_c* url_c;
+  c3_w  len_w;
+  c3_y* hun_y;
+  c3_i  ret_i;
+
+  ret_i = asprintf(&url_c, "%s/%s/%s/next", ver_hos_c, pac_c, URBIT_VERSION);
+  c3_assert( ret_i > 0 );
+
+  if ( !_king_curl_bytes(url_c, &len_w, &hun_y) ) {
```
mcevoypeter, line 360: Can you use the explicit numerical check instead?
```
  if ( 0 == _king_curl_bytes(url_c, &len_w, &hun_y) ) {
```

pkg/urbit/vere/king.c
```diff
@@ -231,54 +236,161 @@ _king_curl_alloc(void* dat_v, size_t uni_t, size_t mem_t, void* buf_v)
...
+  }
+  else {
+    c3_free(url_c);
+
+    //  XX support channel redirections
+    //
+    ret_i = asprintf(&url_c, "%s/%s/last", ver_hos_c, pac_c);
+    c3_assert( ret_i > 0 );
+
+    if ( _king_curl_bytes(url_c, &len_w, &hun_y) )
```
mcevoypeter, line 372: Same explicit numerical check nit here too.

pkg/urbit/vere/king.c
```diff
@@ -231,54 +236,161 @@ _king_curl_alloc(void* dat_v, size_t uni_t, size_t mem_t, void* buf_v)
...
+  c3_free(url_c);
+
+  //  XX trim ver_c ?
+  //
+  if ( 0 == strcmp(ver_c, URBIT_VERSION) ) {
+    c3_free(ver_c);
+    return -1;
   }
+
+  *out_c = ver_c;
```
mcevoypeter, line 392: NULL check.

pkg/urbit/vere/king.c
```diff
@@ -864,13 +976,490 @@ _king_done_cb(uv_handle_t* han_u)
...
+    u3l_log("unable to open %s: %s\r\n", bin_c, strerror(errno));
+    c3_free(bin_c);
+    return -1;
+  }
+
+  ret_i = asprintf(&url_c, "%s/%s/%s/vere-v%s-%s",
+                   ver_hos_c, pac_c, ver_c, ver_c, arc_c);
+  c3_assert( ret_i > 0 );
+
+  if ( (ret_i = _king_save_file(url_c, fil_u)) ) {
```
mcevoypeter, line 1116: Explicit numerical check nit here too.

mcevoypeter (commented):
Some comments so far. I'm traveling today, and so I'll take a second look tomorrow morning with fresh eyes to see if I have any other points of feedback.

mcevoypeter:
What's the advantage of using a reflink for the copy of the binary in the pier i.e. at `<pier>/.bin`? The fact that reflink support is filesystem dependent is concerning and adds needless complexity, especially because this upgrade process will not be run frequently enough to justify the complexity on the grounds of improved performance. `_king_copy_file()` is 100+ lines of gross platform-dependent code, which could be almost entirely eliminated (with the exception of the MinGW support) in favor of just using `_king_copy_raw()`. 

I would also always include `<pier>/.bin/pace` with a default value of `live`. That way, users don't have to remember that they're supposed to create a file with a particular name, just edit one with a particular name, which seems less error prone. 

Also, how can I test this locally? When I attempt `urbit --next <pier>` with a fake ship, I'm getting HTTP 403 from the bootstrap.urbit.org URL (which I uncommented at the top of `king.c`, which I'm assuming means that the binaries and version files are not up yet.

joemfb:
This PR now includes:

build:

- properly mostly-static builds on macos (thanks @eamsden!)
- assertions of mostly-static builds on macos and windows in ci
- automatic deployment of binaries to the "often" `pace` (release-channel)
  - triggered by pushes to `next/vere`, or configured on manually triggered builds
  - the truncated commit hash is used as the version number in this `pace`

upgrade:

- vere has a `next` subcommand which
  - prepares a pier for upgrade
  - checks for a new version (in the configured or default `pace`)
  - saves the new binary in `$pier/.bin/[pace]/`
  - and hardlinks that new binary in `$pier/.run`
- vere has a `prep` subcommand which prepares the pier for upgrade

boot:

- vere copies itself into a new pier, after a successful boot, on exit
  - this behavior can be disabled with `--no-dock`
  - problem: vere names itself with its hardcoded version string, and the default pace


These changes have been tested on linux, macos, and windows. To try for yourself,
build this branch, boot a fake ship (at `$pier`) and run the following:

```
mkdir $pier/.bin/
echo -n "often" > $pier/.bin/pace
path/to/urbit next $pier
```

problems and next steps:

The current boot behavior is correct for released binaries, but somewhat wrong for dev or prelease builds, as the binary will always identify itself as being the `$version` release in the "live" `pace`. The root of the problem is that the release channel (and version, in the case of automatic ci build) is not part of the binary, but inferred from where the binary was retrieved (the parameters of the url, specifically). It would be possible to work around this ambiguity
by always attempting to upgrade the binary on boot, at the cost of downloading vere twice (first in order to boot, then again in order to link it to the pier).

The `prep` command is intended to be part of manual upgrade. A related command to download a binary to a specified directory would be very useful.

Migrating a pier to a new architecture with old-architecture binaries copied/linked in it is a little strange. They don't hurt anything per se, but it's confusing, and `$pier/.run` won't work until it's replaced. A command to explicitly prepare for migration would be very useful.

joemfb:
To address the ambiguity around binary copying on boot, I've made `pace` a compile-time value in vere. I had wanted to avoid this kind of step, in the hopes that promoting artifacts from nightly -> prerelease -> release could be done without rebuilding. But our version strings for prerelease binaries (`v1.9-rc4`) already prevent this pattern.

There's still a minor ambiguity with copying on boot. Nightly binaries are uploaded with a "version" set to the truncated git commit hash, but they'll name themselves with an actual version string (ie, `pkg/urbit/version`) when copied into a pier. That's a little weird, but I can't see how it could hurt anything.

pkg/urbit/daemon/main.c
```diff
@@ -347,8 +355,22 @@ _main_getopt(c3_i argc, c3_c** argv)
...
+      c3_w len_w = strlen(argv[0]);
+
+      //  no args, argv[0] == $pier/.run
+      //
+      if (  (U3_BIN_ALIAS_LEN <= len_w)
+         && (0 == strcmp(argv[0] + (len_w - U3_BIN_ALIAS_LEN), U3_BIN_ALIAS)) )
+      {
+        c3_c* bin_c = strdup(argv[0]);
+        u3_Host.dir_c = _main_repath(dirname(bin_c));
+        c3_free(bin_c);
```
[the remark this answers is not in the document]
joemfb: How big should the target buffer be?

pkg/urbit/daemon/main.c
```diff
@@ -1239,6 +1266,13 @@ c3_i
 main(c3_i   argc,
      c3_c** argv)
 {
+  if ( argc <= 0 ) {
+    fprintf(stderr, "nice try, fbi\r\n");
+    exit(1);
+  }
```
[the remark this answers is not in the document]
joemfb: `argc <= 0` would be a very bizarre situation -- I don't think it's except through `exec()` and friends.

pkg/urbit/include/c/portable.h
```diff
@@ -101,6 +104,38 @@
...
+
+  /** Binary alias.
+  **/
+#   ifdef U3_OS_mingw
+#     define U3_BIN_ALIAS ".run.exe"
+#   else
+#     define U3_BIN_ALIAS ".run"
+#   endif
+
+#   define U3_BIN_ALIAS_LEN (sizeof(U3_BIN_ALIAS) - 1)
```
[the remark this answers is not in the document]
joemfb: It is guaranteed, the type of a string literal includes the size (`"foo"` is `const char[4]`). I wrote it this way to "run `strlen()` at compile time". But it's not exactly a hot loop, so I've removed the second macro.

But not the first -- if I can't put portability-oriented string literals in macros, what can I put in them?

pkg/urbit/vere/king.c
joemfb: The purpose of this function is get an HTTP response body as a byte buffer. Why should either the base pointer or length out-params be optional?

pkg/urbit/vere/king.c
joemfb: The subsequent asprintf() was only null-terminating this buffer (and copying it to a string), so I've removed it. But I have the usual question for stack-allocating strings/buffers: what size should it be?

pkg/urbit/vere/king.c
joemfb: Same reply: the purpose of this function is to (try to) get the next version string? Why make receiving that string optional?

pkg/urbit/vere/king.c
joemfb: done.

pkg/urbit/vere/king.c
joemfb: In the past I might've agreed with you, but, for better or worse, this is idiomatic C for apis with return codes where `0` signals success.

joemfb:
@mcevoypeter the reason for the reflinks is to try to avoid proliferating disk-space-using copies of binaries when booting multiple ships. But that's not the only reason for the 100+ lines of platform-dependant api calls -- I'm also trying to use the most efficient mechanism for copying the binary. That's not essential here, but this is not the only place where we (will be) routinely copying large files (see epoch creation). So I think it's worth figuring out and using these platform-dependant apis. I had first looked at re-using the existing copying code for snapshot backup, but that handles fewer error cases, and appears to mishandle partial reads. (And that's not the only snapshot code with those deficiencies!)

Automatically creating the `pace` file is an interesting idea. It doesn't simplify the implementation (we still have to do something sane if the file has been deleted), but it might smooth out the UX. Even more important is trimming whitespace (at least trailing), it's irritating to have to create text files without a trailing newline. I've also been wondering if it'd be better to put the file somewhere other than `.bin/` -- maybe just `$pier/.pace`?

joemfb:
Also worth noting re: code review -- most of these additions to `vere/king.c` would probably be better in a separate module (`vere/dock.c`?). And the "vendoring" of libwhereami is pretty gross. Both of these are concessions to deadlines, it'd be great to clean things up further ASAP.

joemfb:
Here are scenarios enabled by this PR:

boot:

- boot normally, get a binary copied into `$pier/.bin/[pace]` and linked at `$pier/.run`
- pass `--no-dock`, skip copying/linking

run:

- `path/to/urbit ARGS $pier` works as normal
- `$pier/.run ARGS` works as well

subcommands

- `path/to/urbit CMD $pier` or `$pier/.run CMD`
  - except for commands where the pier is not relevant
- new commands include
  - `next` (or `upgrade`), full upgrade
  - `prep`, prepare pier for upgrade
  - `dock`, copy/link binary into pier
  - `vere` (or `fetch-vere`), download binary into specified directory

full upgrade:

- `path/to/urbit next $pier` or `$pier/.run next` (`upgrade` also works)

deconstructed upgrade:

- `path/to/urbit prep $pier` or `$pier/.run prep`
- `path/to/urbit vere $output-dir` (`fetch-vere` also works)
  - arguments include `-a`/`--arch`, `-v`/`--version`, and `-p`/`--pace`
- run with `$output-dir/vere-v$version-$arch $pier`
- or link it back into the pier with `$output-dir/vere-v$version-$arch dock $pier`

migration

- migration is somewhat supported via the optional `-a`/`--arch` argument to `next`
  - supported architectures are currently `x86_64-linux`, `x86_64-darwin`, and `x86_64-windows`

Fang-:
> `dock`, copy/link binary into pier

Feel like, in line with `upgrade` and `fetch-vere`, we should probably have a `copy` alias. And likewise `--no-copy` in addition to (or instead of?) `--no-dock`.

Wondering if there's a verb that more accurately describes "copy into ourselves". "Install" is too overloaded, and not _quite_ right, but does feel closer...

(Also reiterating my suggestion for `fast` in place of `often`, but we can probably swap that out less disruptively.)

mcevoypeter:
> That's not essential here, but this is not the only place where we (will be) routinely copying large files (see epoch creation). So I think it's worth figuring out and using these platform-dependant apis. 

This makes a lot of sense. Given that it's something that we'll be using in other places, I think we should factor it out into a separate `c3` level module called `fs` or similar. Mixing it in at the site of its use is not ideal, and pulling it out into its own module should be simple.

> Automatically creating the `pace` file is an interesting idea. It doesn't simplify the implementation (we still have to do something sane if the file has been deleted), but it might smooth out the UX. 

Yeah, even if it's just a UX improvement, I think it's worth doing. Plus, the sane thing to do is probably either complain and exit or default to `live`, the latter of which I assume we'd have to do anyway.

> Even more important is trimming whitespace (at least trailing), it's irritating to have to create text files without a trailing newline. 

```c
static inline void
trim_whitespace(c3_c* const str_c)
{
  if ( !str_c ) {
    return;
  }
  
  // Trim leading whitespace.
  {
    c3_c* ptr_c = str_c + strspn(str_c, " \n\t");
    if ( str_c < ptr_c ) {
      memmove(str_c, ptr_c, strlen(ptr_c) + 1);
    }
  }

  // Trim trailing whitespace.
  {
    c3_c* ptr_c = str_c + strlen(str_c) - 1;
    for ( ;; ) {
      if ( ptr_c < str_c ) {
        break;
      }

      c3_c char_c = *ptr_c;
      if ( ' ' != char_c && '\n' != char_c && '\t' != char_c ) {
        break; 
      }

      *ptr_c-- = '\0';
    }
  }
}
```
:) (untested)

> I've also been wondering if it'd be better to put the file somewhere other than `.bin/` -- maybe just `$pier/.pace`?

I like this idea.

joemfb:
This PR is now feature complete.

- `pace` redirects have not been implemented. these can be added in the future, with backward compatibility, through staged upgrades (ie, `$version/next`)
- full release automation is not quite complete, due to some polish and permissioning issues
- further modularization, code cleanup, &c can be deferred to subsequent PRs

Also, user-facing names/interfaces need a final round of bikeshedding. There are subcommmand and argument names to consider (`vere`, `--no-dock`), configuration file locations (`$pier/.bin/pace` vs `$pier/.pace`), and `pace` names:

- "once": ad-hoc or development builds. "not a pace"
- "often": "nightly", intended to built from `next/vere` very frequently (@Fang- has suggested "fast")
- "soon": pre-release, for final testing, or staging kernel updates with 3rd-party developers
- "live": release, for everyone

pkg/urbit/daemon/main.c
```diff
@@ -530,23 +594,37 @@ _setup_ssl_curl(void* arg)
...
+    "  %s cram %.*s              jam state:\n",
+    "  %s dock %.*s              copy binary:\n",
+    "  %s grab %.*s              measure memory usage:\n",
+    "  %s info %.*s              print pier info:\n",
+    "  %s meld %.*s              deduplicate snapshot:\n",
+    "  %s pack %.*s              defragment snapshot:\n",
+    "  %s prep %.*s              prepare for upgrade:\n",
+    "  %s next %.*s              request upgrade:\n",
+    "  %s queu %.*s<at-event>    cue state:\n",
+    "  %s vere ARGS <output dir>    download binary:\n",
```
Fang-, line 610: I think it makes some sense to only print the short versions here given the context (cli-level usage docs, instead of a more handhold-y guide/manual).

Should we mention the options/flags too? We have some for `vere` and `next`, right?

Also unsure why there's `:`s at the end of all of these.

Some suggestions for improving clarity below.

```suggestion
    "utilities:\n",
    "  %s cram %.*s              export snapshot:\n",
    "  %s dock %.*s              install binary into pier:\n",
    "  %s grab %.*s              measure memory usage:\n",
    "  %s info %.*s              print pier info:\n",
    "  %s meld %.*s              deduplicate snapshot:\n",
    "  %s pack %.*s              defragment snapshot:\n",
    "  %s prep %.*s              prepare for upgrade:\n",
    "  %s next %.*s              request upgrade:\n",
    "  %s cueu %.*s<at-event>    import snapshot:\n",
    "  %s vere ARGS <output dir>    download binary:\n",
```

("queu" is not a word, so if we're making things up anyway, "cueu" more closely matches the "cue" of "cram"'s "jam". Alternatively, "clue" could be worth considering. Happens to fit nicely with "cramming" in the context of studying, getting a clue to recall things by.)

joemfb:
This PR is a *fait accompli* due to the urgency around v1.9. But all subsequent feedback is greatly appreciated (@Fang-, @mcevoypeter) -- I'll have a follow up PR soon.

PR #5766, opened 2022-05-24
arvo: fix scries when jamming desks

ryjm:


zalberico:
@belisarius222 - assigning this to you to review and merge or close.

@ryjm - reminder that when you create PRs if you can put a basic summary of what it is in the description that'll make it easier for people to evaluate and act on. For example: "The current behavior X happens when doing Y. This fix changes Z so that correct behavior A happens." or something like that.

belisarius222 (approved):
I'm not sure what exactly this is for, but the scrys were clearly wrong before this change.  @ryjm if you've tested this, it should be fine to merge.

Issue #5803, opened 2022-05-26
-test silently skips test arms not under %/tests


dnmfarrell:
The `-test` command evaluates broken tests as "ok" because it filters test paths that don't begin with `%test`:

```
:: /test/yeqn/hoon
/+  *test
|%
++  test-fail
  ;:  weld
  %+  expect-eq
    !>  %.y
    !>  %.n
  ==
--
```
```
> -test %/test/yeqn/hoon ~
ok=%.y
```

This is caused by the `resolve-test-paths`  arm `pkg/arvo/ted/test.hoon` converting paths not starting with %test to ~ and then zinging them to nothing. Seems unintentional?

```
  %+  turn  ~(tap by paths-to-tests)
  |=  [=path test-arms=(list test-arm)]
  ^-  (list test)
  ::  strip off leading 'tests' from :path
  ::
  ?.  ?=([%tests *] path)  ~
  =/  path  t.path
```

As the test command takes a path argument, it should be able to run tests in any path, rather than just those under `%tests`.

belisarius222:
That's intentional, but you might be right that it would be better to be able to point `-test` at any file and have it run all the `+test-*` arms in that file.  I think if we did that, we might want `-test` with no arguments to run all the tests in the /tests directory.

dnmfarrell:
An alternative solution would be to crash if the path is not under `%tests` - and/or accept %test relative paths only. That leads to less flexible/useful code (imo). But it would solve the issue of marking tests with failing test arms as `ok`.

I guessed the reason %tests was stripped was to show the test coverage/correspondence for /lib more conveniently - if that's the case, and it's safe to do so, a better solution altogether would be to not strip %tests at all and let consuming code strip it if needed to (like a report builder or something). But I wasn't sure what the intention was so my PR preserves the behavior.

PR #5804, opened 2022-05-26
pkg: don't silently skip non %tests/ test arms

dnmfarrell:
Preserves the behavior of stripping "%tests" from test paths but keeps
tests under other directories so their test- arms are still evaluated.

Fixes [#5803](https://github.com/urbit/urbit/issues/5803)

### DIFF
--- a/pkg/arvo/ted/test.hoon
+++ b/pkg/arvo/ted/test.hoon
@@ -44,10 +44,6 @@
   %+  turn  ~(tap by paths-to-tests)
   |=  [=path test-arms=(list test-arm)]
   ^-  (list test)
-  ::  strip off leading 'tests' from :path
-  ::
-  ?.  ?=([%tests *] path)  ~
-  =/  path  t.path  ::NOTE  TMI
   ::  for each test, add the test's name to :path
   ::
   %+  turn  test-arms


belisarius222:
Let's go ahead with your idea and stop requiring all tests to be in the `/tests` directory, but in that case I think we'd be better off including the full filename in the test name, and just not strip at all.

Thank you for looking into this.

dnmfarrell:
Hah, I had the same thought.
https://github.com/urbit/urbit/issues/5803#issuecomment-1138649500

Ok I'll revise to remove the stripping behavior.

belisarius222 (approved):
LGTM

belisarius222:
@dnmfarrell Thanks again for doing this.  If you're interested in further improvements, it would also be nice to have `-test` with no args run on the whole `/tests` directory, and especially nice to be able to pass in just a single path to `-test`, instead of a list of paths.

dnmfarrell:
Sure I'll take a look at those next

philipcmonk:
This change unfortunately prevents you from running `-test % ~` because some files (at least lib/test.hoon) include arms which start with `test-` but are not valid tests.

https://github.com/urbit/urbit/blob/361384d15c6a22ffc9424deb4f23848c36123906/pkg/base-dev/lib/test.hoon#L55

philipcmonk:
Sorry, that issue is actually a regression in #5972.  This change looks fine.

PR #5805, opened 2022-05-26
@urbit/http-api: surface mark of updates

liam-fitzgerald:
Surfaces the mark of the subscription updates.
For this to not be `json` requires that the ship be running the upcoming kernelspace update. However this upgrade is backwards and forwards compatible, i.e. current `http-api` and these changes are compatible with both current kernel space and upcoming kernel space. How to proceed with versioning changes? Suspect a minor version bump, because type signatures have changed, albeit not very meaningfully.

### DIFF
--- a/pkg/npm/http-api/src/Urbit.ts
+++ b/pkg/npm/http-api/src/Urbit.ts
@@ -273,7 +273,7 @@ export class Urbit {
             ) {
               const funcs = this.outstandingSubscriptions.get(data.id);
               try {
-                funcs.event(data.json);
+                funcs.event(data.json, data.mark ?? 'json');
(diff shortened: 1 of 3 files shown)


liam-fitzgerald:
no, as in for the `mark` argument to the event callback to not be the string `"json"`

matildepark:
How breaking is this for non-Landscape Urbit API consumers? I recall Landscape used to key off that `json`, unsure if others do.

liam-fitzgerald:
This is not breaking whatsoever, we retain full forward and backward compat

arthyn (approved):
Seems fine to me

PR #5808, opened 2022-05-27
pkg: -test shouldn't require a list of paths

dnmfarrell:
Changes the test command to check if args contains a single path and
wraps it in a list. Now a test thread can be started without providing
a list:

    -test %/tests/lib

And passing a list still works:

    -test %/tests/lib ~

If no arg is given, `test` will run all tests under `/=base=/tests`.


dnmfarrell:
Hey @belisarius222 see what you think of this. I tried handling the `~` case but as clay paths are relative I wasn't sure how to construct a root path to /=base=/tests. Maybe I can build it from the thread bowl?

PR #5810, opened 2022-05-29
Symmetric routing draft

leowini:
@belisarius222 
Submitting for feedback:
* replaced origin field with the next field
* added two bits to the packet header for scry notifications packet type
* added the relay-state data structure to the ames-state
* changed send-blob to send-packet
* +on-hear-packet:
  1. Add the ship to the sponsees map if the packet is from a sponsee and not forwarded.
  2. Add the lane to the pending map.
  3. Update the pending map next lane if the packet has a next field.
* +send-packet
  1. If we are the sponsor of rcvr.packet, send to the lane from the sponsees map.
  2. If we are not the sponsor, try to get the lane from the pending map.
  3. If we don't know rcvr.packet, send to its sponsoring galaxy.
  4. Write the next field if forwarding
TODOs
- [ ] Update the ames-state to remove routes completely
- [ ] Update the scry paths since previous routing system is deprecated - need guidance on specifics
- [ ] testing
- [ ] handle upgrading


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
...
+        =(~ next.packet)
+        =/  val  (~(get by sponsees.relay-state) sndr.packet)
+        =(~ val)
+        =(our (sein:title our now sndr.packet))
+      ==
+      ::  send 30s expiration timer to behn
+      ::
+      =/  duct  ~[/ames]
+      =/  wire  (make-sponsee-timer-wire sndr.packet)
+      =.  event-core
```
belisarius222, line 1533: This mutation won't escape the enclosing `=?`.  You'll need to `=?  event-core`, I think.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -248,6 +248,39 @@
...
+  /pending/(scot %p sndr)/(scot %p rcvr)
+::  +parse-pending-timer-wire
+::
+++  parse-pending-timer-wire
+  |=  =wire
+  ^-  dyad
+  ::
+  ~|  %ames-wire-timer^wire
+  ?.  ?=([%pending @ @ ~] wire)  !!
+  ?~  sndr=`(unit @p)`(slaw %p i.t.wire)  !!
```
belisarius222, line 265: You can use `+slav` instead of `+slaw` to do the assertions for you.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -283,17 +316,18 @@
...
     ==
   =/  checksum  (end [0 20] (mug body))
-  =?  body  ?=(^ origin)  (mix u.origin (lsh [3 6] body))
+  =?  body  ?=(^ next)  (mix u.next (lsh [3 6] body))
   ::
   =/  header=@
     %+  can  0
-    :~  [3 reserved=0]
+    :~  [1 reserved=0]
+        [2 packet-type=0]
```
belisarius222, line 324: We'll need at a bit to represent "is this a request?", i.e. not a response.  The `is-ames` bit should also be consolidated into the 2-bit `packet-type` field to save space, and of course this function should return these new values -- my inclination would be to add the fields to the `$packet` type.

This is a note as much to myself as to you:

Something to keep in mind with this sort of change to the packet format is that we should ensure packets at this version are discriminable from older packets.  In this case it's good enough just to look at the protocol version, but I was hoping each protocol could develop its own version to enable per-protocol updates later.  We could probably still do that, though, by looking at both the protocol version bits and the packet-type bits.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
     ^+  event-core
     %-  %^  trace  odd.veb  sndr.packet
         |.("received packet")
+    :: add the ship to our sponsees map if we are the sponsor
+    ::
+    =*  relay-state  relay-state.ames-state
+    =?  sponsees.relay-state
+      ?&
```
belisarius222, line 1523: style nit:
```hoon
?&  foo
    bar
==
```
is preferred over
```hoon
?&
  foo
  bar
==
```
```

File: pkg/arvo/sys/vane/ames.hoon [hunk 5]
```diff
@@ -1434,6 +1516,57 @@
...
+        =(our (sein:title our now sndr.packet))
+      ==
+      ::  send 30s expiration timer to behn
+      ::
+      =/  duct  ~[/ames]
+      =/  wire  (make-sponsee-timer-wire sndr.packet)
+      =.  event-core
+        (emit duct %pass wire %b %wait (add ~s30 now))
+      ::
+      ^+  sponsees.relay-state
```
belisarius222, line 1536: don't need this cast

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
...
+      (~(put by sponsees.relay-state) [sndr.packet [lane (add ~s30 now)]])
+    :: add the lane to our pending map
+    ::
+    =*  inv-req  [%ames `dyad`[rcvr.packet sndr.packet]]
+    =?  pending.relay-state
+      =/  val  (~(get by pending.relay-state) inv-req)
+      ?~  val  %.y
+        =/  next  q.-.u.val
```
belisarius222, line 1544: style: dedent these four lines, since it's the second option of the `?~  val` and should therefore be at the same indentation level as the `?~` to maintain backstep

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
...
+      ::
+      =/  val  (~(get by pending.relay-state) inv-req)
+      ?~  val  !!
+      =/  next  q.-.u.val
+      &(=(~ next) !=(~ next.packet))
+      ::
+      ^+  pending.relay-state
+      ?~  next.packet  !!
+      %-  ~(put by pending.relay-state)
+        [inv-req [[lane `[%.n u.next.packet]] expiry=(add ~s30 now)]]
```
belisarius222, line 1569: dedent once to match the `%-`

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
...
+      ?~  val  %.y
+        =/  next  q.-.u.val
+        ?~  next
+          %.n
+        =(lane u.next)
+      ::  send 30s expiration timer to behn
+      ::
+      =/  duct  ~[/ames]
+      =/  wire  (make-pending-timer-wire [rcvr.packet sndr.packet])
+      =.  event-core
```
belisarius222, line 1552: This mutation also won't escape its enclosing `=?`.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
...
+        (emit duct %pass wire %b %wait (add ~s30 now))
+      ::
+      ^+  pending.relay-state
+      (~(put by pending.relay-state) [inv-req [[lane ~] (add ~s30 now)]])
+    ::  update the pending map 'next' lane if necessary
+    ::
+    =?  pending.relay-state
+      ::
+      =/  val  (~(get by pending.relay-state) inv-req)
```
belisarius222, line 1561: could use `(~(got by ...` here for the null assertion

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1434,6 +1516,57 @@
...
+      ?~  val  !!
+      =/  next  q.-.u.val
+      &(=(~ next) !=(~ next.packet))
```
belisarius222, line 1564: this line looks like it'll cause an early return from the `=?`, so the code on the next few lines is orphaned

belisarius222 (commented):
Did a cursory initial review.  Some nits, but this looks very promising.

pkg/arvo/sys/vane/ames.hoon
leowini: I think this is fine; this block returns the loobean for the conditional.

pkg/arvo/sys/vane/ames.hoon
leowini: What would the request/response bit be used for? My current implementation is request/response agnostic. I'm setting the next lane on every forwarded packet. Is this a bad idea?

pkg/arvo/sys/vane/ames.hoon
belisarius222: ah, I see -- it should be indented, then, along with the first arg to the `=?`, like this:
```hoon
=?    pending.relay-state
    ::
    =/  val ...
    &(...)
  ::
  ^+  pending.relay-state
  ...
```

pkg/arvo/sys/vane/ames.hoon
leowini: This makes sense to me. I can change the logic to treat requests and responses differently, but I'm not sure it's necessary.
My current relay logic works like this:
1. If we have a lane for sndr to rcvr in the pending map, use that.
2. If we are the direct sponsor of the rcvr, send to rcvr. (rcvr is in our sponsees map)
  --> otherwise, +try-next-sponsor up the sponsorship chain in case rcvr is our transitive sponsee
3. If we don't have an entry for the request in the pending map AND the rcvr isn't our direct or transitive sponsee, send to the rcvr's galaxy.

Regardless of request/response, +on-hear-packet stores [[rcvr sndr]] [lane (unit next-lane)] in the sponsorship map. I think this handles the routes correctly and maintains symmetry without needing to treat requests and responses differently.

Example:

request:  ~rovnys -> ~syl -> ~sipsyl -> ~ritpub-sipsyl -> ~mister-ritpub-sipsyl

~rovnys has no record of ~mister-ritpub-sipsyl in the pending requests map and is not a sponsor, so it sends the packet to the ~syl galaxy.
As the request is routed, each relay will enter [[~mister-ritpub-sipsyl ~rovnys] [previous-lane previous-previous-lane]] in its pending map. e.g. ~sypsyl will have [[~mister-ritpub-sipsyl ~rovnys] [(~syl's lane) (~rovnys's lane)]].

response: ~rovnys <- ~syl <- ~sipsyl <- ~ritpub-sipsyl <- ~mister-ritpub-sipsyl

On the way back, the relay's all have entries in their pending map for [~mister-ritpub-sipsyl ~rovnys], so they will send the packet to both lane and the next-lane.
As the request is routed, each relay will enter [[~rovnys ~mister-ritpub-sipsyl] [previous-lane previous-previous-lane]] in its pending map. e.g. ~sypsyl will have [[~rovnys ~mister-ritpub-sipsyl] [(~ritpub-sipsyl's lane) (~mister-ritpub-sipsyl's lane)].

Then, on the next request, the relays have entries in their pending map going both ways. Each relay has an entry in the pending map for [~rovnys ~mister-ritpub-sipsyl] and an entry for [~mister-ritpub-sipsyl ~rovnys]. i.e. after the first request, only the pending map is used to route packets. The result is that all of the relays then try to tighten the route in both directions.

belisarius222:
How sure are you that this is symmetric?  Is it possible for, say, two request packets in a row to take the same trajectory, then one of the relays tightens on the way back?

It might be more efficient for all the relays to tighten bidirectionally all the time, but it's a pretty different design from the original proposal, so we should be careful to think through all the edge cases and make sure it would handle them correctly.

—
~rovnys-ricfer
https://urbit.org

leowini:
It would be possible for a small number of packets to take a tighter route on the way back if the route tightens while the packet is at the destination. However, the route will only tighten if there is a known path to the destination on the tightened route (the packet will only tighten a route to the next lane if it receives a packet from that lane). Essentially, all packets will take the tightest known route to the destination both ways. Is there an issue with breaking strict symmetry in this way?

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2004,86 +2096,82 @@
...
+  ::  +send-packet: fire packet at lane and maybe next
   ::
   ::    If we have no PKI data for a recipient, enqueue the packet and
   ::    request the information from Jael if we haven't already.
   ::
-  ++  send-blob
-    ~/  %send-blob
-    |=  [for=? =ship =blob]
+  ++  send-packet
+    |=  [for=? next-lane=(unit lane) =packet]
```
dnmfarrell, line 2108: As next-lane is a unit, is `for` redundant?

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1462,18 +1586,8 @@
...
-        &(?=(~ origin.packet) !=(%czar (clan:title sndr.packet)))
-      ?:  ?=(%& -.lane)
-        ~
-      ?.  (lte (met 3 p.lane) 6)
-        ~|  ames-lane-size+p.lane  !!
-      `p.lane
-    ::
-    =/  =blob  (encode-packet packet)
-    (send-blob & rcvr.packet blob)
+    (send-packet & `lane packet)
```
dnmfarrell, line 1590: I don't see a caller of `on-hear-forward` which provides `lane`- am I might be missing something?

leowini: on-hear-forward gets called in on-hear-packet, with the same sample as on-hear-packet.
```
 %.  +<
 ::
 ?.  =(our rcvr.packet)
    on-hear-forward
```

pkg/arvo/sys/vane/ames.hoon
leowini: yes I think so I'll get rid of it

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2004,86 +2096,82 @@
...
-        ::    behind a NAT; a planet receiving initial acknowledgment
-        ::    from a star; a planet talking to another planet under
-        ::    another galaxy.
-        ::
-        ?:  ?|  =(our ship)
-                ?&  !=(final-ship ship)
-                    !=(%czar (clan:title ship))
-                ==
+    =/  final-ship  rcvr.packet
+    %-  (trace rot.veb final-ship |.("send-packet: to {<ship>}"))
```
dnmfarrell, line 2107: Should this be `{<final-ship>}` as `ship` is no longer an alias of `final-ship`?

pkg/arvo/sys/vane/ames.hoon
dnmfarrell: Got it, thanks

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -640,11 +675,21 @@
...
       =bug
   ==
+::  $relay-state: state for a relay
+::
+::    pending:  where to route requests
+::    sponsees: lanes for sponsees
+::
++$  relay-state
+  $:  pending=(map ship [(pair lane (unit lane)) expiry=@da])
+      sponsees=(map ship [lane expiry=@da])
```
dnmfarrell, line 691: Wasn't sponsees going to have both rcvr and sndr ships in it?

pkg/arvo/sys/vane/ames.hoon
dnmfarrell: Tracing this back: it seems that +call correctly passes lane when dud is a cell by extracting it from the task `lane.task` on line 1065. Otherwise is passes `lane` (line 1071) which I don't see in the sample at that point - is it passing the type `$lane` and not the value?

leowini: It's still getting the lane and blob from the task: `[lane blob ~]:task`. The dud looks like a stack trace in case of a crash (I think in the arvo kernel).

pkg/arvo/sys/vane/ames.hoon
leowini: yes I'll change that

pkg/arvo/sys/vane/ames.hoon
leowini: Yeah that was the original design. I noticed that the relays only need to know the destination, so I'm pretty sure it's okay to drop the sender. e.g. if a relay knows the lane of ~zod, it shouldn't matter where the relayed packets are coming from. It just sends them to ~zod's lane.

zalberico:
This PR needs an associated issue for symmetric routing.

zalberico:
In discussion with @belisarius222 he's going to focus on this as part of his roadmap work and we'll create an issue then that references the important bits when we get to it. As is, we're going to close this PR for now until we're working on it.

PR #5813, opened 2022-05-31
vere: fixes -X scry/export (broken in previous refactoring)

joemfb:
This restores `-X`, which was broken in #5575.


philipcmonk (approved):
works on my machine

PR #5814, opened 2022-05-31
vere: print ip address in correct order

Fang-:
This was printing the components of IPv4 addresses in backwards order.


pkg/urbit/vere/io/ames.c
```diff
@@ -855,7 +855,7 @@ _ames_forward(u3_panc* pac_u, u3_noun las)
 
     u3l_log("ames: forwarding for %s to %s from %d.%d.%d.%d:%d\n",
             sen_c, rec_c,
-            pip_y[0], pip_y[1], pip_y[2], pip_y[3],
+            pip_y[3], pip_y[2], pip_y[1], pip_y[0],
```
joemfb, line 858: This change is correct so long as the "host byte order" is little-endian. That probably merits the classic comment.

joemfb (approved):
LGTM

PR #5815, opened 2022-06-01
ames: do not produce stale forward lanes

Fang-:
Stale lanes may cause forwarding loops. Imagine the following:
1) Planet A is live. Galaxy B, its indirect sponsor, learns of its route.
2) A goes offline. Another ship, C, is started in its place, at the same route.
3) B receives a packet for A, forwards it to the known route.
4) C received the packet, forwards it to B.
5) Repeat from 3.

Here, we update the forward lane(s) scry used by the runtime to not produce a
peer's lane if they haven't communicated with us in the last hour. Everyone's
supposed to ping their sponsorship chain every 30 seconds. If those aren't
going through, you shouldn't expect to be reachable anyway.

We may or may not want to update +send-blob to match.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2059,6 +2059,10 @@
           ?:  for
             event-core
           (try-next-sponsor sponsor.peer-state)
+        ::  if forwarding, route must not be stale
+        ::
+        ?:  &(for (lth last-contact.qos.peer-state (add now ~h1)))
+          (try-next-sponsor sponsor.peer-state)
         ::
         ?~  route=route.peer-state
           %-  (trace rot.veb final-ship |.("no route to:  {<ship>}"))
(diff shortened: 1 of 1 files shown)


joemfb:
Should this logic also be added to `+send-blob`?

belisarius222 (dismissed):
lgtm

Fang-:
We should make sure this doesn't mess with moon forwarding, and other test cases we should probably be hitting during devstream. We could merge now, since it's Probably Fine™, but also don't want to forget to actually do those tests, just in case.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2055,6 +2059,10 @@
           ?:  for
             event-core
           (try-next-sponsor sponsor.peer-state)
+        ::  if forwarding, route must not be stale
+        ::
+        ?:  &(for (lth last-contact.qos.peer-state (add now ~h1)))
```
joemfb, line 2064: I think this should call `+sub`.

belisarius222: ah yes, this line was designed for ships that move backward through time, like positrons

belisarius222 (approved):
lgtm

PR #5823, opened 2022-06-02
[hotfix] pull-hook: resync subscriptions, slowly

liam-fitzgerald:
Autokicks all subscriptions, slowly over time, so as not to cause too
much stress on the network

- Do we have `eny.bowl` as a real value in an +on-load?
- Is three hours long enough/too long?

### DIFF
--- a/pkg/landscape/lib/pull-hook.hoon
+++ b/pkg/landscape/lib/pull-hook.hoon
@@ -107,12 +107,15 @@
 ::
 +$  state-4  [%4 base-state-3]
 ::
++$  state-5  [%5 base-state-3]
+::
 +$  versioned-state 
   $%  state-0
       state-1
       state-2
       state-3
       state-4
+      state-5
   ==
 ::  +diplomatic: only renegotiate if versions changed
 ::    
@@ -214,7 +217,7 @@
 ++  agent
   |*  =config
   |=  =(pull-hook config)
-  =|  state-4
+  =|  state-5
   =*  state  -
   ^-  agent:gall
   =<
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
Why is this necessary?

Fang- (approved):
Thanks, I hate it. But we do what we must.

For context: this is working around a subscription desync bug that's been around for a while and has seemingly elevated incidence during/around OTAs? Still rather rare though. Next system update will kill the desync bug dead (#5644), but we should get people connected again sooner.

Fang-:
> Do we have `eny.bowl` as a real value in an +on-load?

I see no reason why we wouldn't.

> Is three hours long enough/too long?

Eh, the ota's gonna be spread out a little bit already anyway. I'm more concerned about `%kick` pokes doing all subscriptions at once, which is still a big hit, but I'd be even more concerned about adding special mechanisms here. This is Probably Fine™.

h5gq3:
any progress on this?

liam-fitzgerald:
made redundant by the recent deployment of the gall request queue fix, closing

h5gq3:
see this https://github.com/urbit/urbit/issues/5834#issuecomment-1231461626

joemfb:
If you still have subscription issues after the recent kernel updates, you may need to run `:graph-pull-hook +kick, =nice |` and `:metadata-pull-hook +kick, =nice |` to force restart everything (on top of the new, fixed subscription lifecycle). @liam-fitzgerald can you confirm those are the right commands?

arguser:
Hello, just ran the first two commands and my ship went berserker on CPU use and then unresponsive. Not sure if I should let it work or kill it.

liam-fitzgerald:
let it work for a while

Issue #5748, opened 2022-05-05
Azimuth block gets stuck on many ships following recent OTA


dr-frmr:
This is an issue that I personally experienced on my ship and have seen in several others at this point. The issue is diagnosed by running `+azimuth-block` in dojo. If it's significantly behind (mine is `14.719.179` at time of writing) and doesn't increase regularly, it may be stuck. 

The block being stuck means that you can't communicate with ships whose network keys have changed since the block you're stuck on. This sequence of commands seems to fix the problem.
```
|nuke %eth-watcher
|start %eth-watcher
:azimuth %resub
:azimuth|watch 'http://eth-mainnet.urbit.org:8545' %default
```
I breached recently and notice that there are several ships that I can no longer communicate with. I suspect they may be experiencing this issue without knowing -- not sure what the best way to disseminate this info is.

tacryt-socryp:
Can confirm that this happened to me and that the fix did indeed fix it. This is a major issue. I don't know what part of the state machine is getting messed up, but it's an extremely fundamental issue. We must have properly working networking for any of Ames' constraints to work. @belisarius222

belisarius222:
This seems like a problem with `%eth-watcher`.  Maybe its timer is getting stuck.

@dr-frmr How recent was your breach when this happened?

vvisigoth:
I also was behind (couldn't connect with ~haddef-sigwen who had breached) and I ran these commands. I seem to still be stuck. Any ideas? I'm happy to jump on a call with someone to troubleshoot and collect data. 

Also getting a ton of this in the dojo.

```
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
sivc-de-punt
```

vvisigoth:
Update on this. 

Tried all above commands, checked my timers, set up a personal infura node, pointed eth-watcher at it and my `(lent logs)` remain stuck at 0. I also observe (on the infura dashboard) that no requests have actually hit infura from my ship. And I checked that the instance could hit infura using curl, so it's purely an Urbit problem.

What more can I try to debug this? There's no way to tell how many people are out there who can't communicate because of these log issues, so it's high priority to figure this out. @lukestiles @philipcmonk

vvisigoth:
[...]

PR #5826, opened 2022-06-04
jets: remove +de:sivc punt logging

joemfb:
The "punt" logging pattern in the crypto jets was intended to surface unusual error conditions. But urcrypt and libaes-siv don't clearly distinguish between deterministic decryption failure and other errors (invalid calls, allocation failures, &c). to ensure determinism, so the relevant jets "punt" *all* errors back to the nock interpreter. And `+de:sivc` is used by ames, which means that receiving any invalid ciphertext or stale keys produces noisy, useless output: two lines of `sivc-de-punt` per packet.

The PR just removes the logging, but there are two non-trivial next steps which should be done:

- deterministic decryption failure should not punt

Accomplishing this will require review of libaes-siv and the openssl functions it depends on, and may require patches to distinguish error conditions. This is important for many reasons, not the least of which is that punting is expensive (and therefore a DOS vector).

- receipt of bad packets needs better tracking and logging

Other than the (accidentally) frequent "punt" logging, the ames driver keeps a counter of failed packets and prints every 1K (or every time with `-v`). But if we're receiving lots of invalid packets with a valid ames header, there's a good chance that we or our peer is behind on PKI state. (And if packets from a particular ip:port are consistently invalid, we probably want to stop try to process them for a while.)

### DIFF
--- a/pkg/urbit/include/c/motes.h
+++ b/pkg/urbit/include/c/motes.h
@@ -387,6 +387,7 @@
 #   define c3__ergo   c3_s4('e','r','g','o')
 #   define c3__esh    c3_s3('e','s','h')
 #   define c3__etch   c3_s4('e','t','c','h')
+#   define c3__evil   c3_s4('e','v','i','l')
 #   define c3__ex     c3_s2('e','x')
 #   define c3__exit   c3_s4('e','x','i','t')
 #   define c3__eyre   c3_s4('e','y','r','e')
--- a/pkg/urbit/jets/e/aes_siv.c
+++ b/pkg/urbit/jets/e/aes_siv.c
@@ -149,9 +149,11 @@ _cqea_siv_de(c3_y*   key_y,
     txt_y = u3r_bytes_alloc(0, txt_w, txt);
     out_y = u3a_malloc(txt_w);
 
-    ret = ( 0 != (*low_f)(txt_y, txt_w, dat_u, soc_w, key_y, iv_y, out_y) )
-      ? u3_none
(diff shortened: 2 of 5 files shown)


belisarius222 (approved):
I approve.  This logging annoys more than informs.

If the encryption libraries don't report whether their error was deterministic, then the only guaranteed-correct way to handle that is for Nock to treat any error as nondeteterministic.  Otherwise we could mistakenly think encryption failed deterministically, which would violate Nock.

belisarius222 (approved):
I approve.  This logging annoys more than informs.

If the encryption libraries don't report whether their error was deterministic, then the only guaranteed-correct way to handle that is for Nock to treat any error as nondeteterministic.  Otherwise we could mistakenly think encryption failed deterministically, which would violate Nock.

joemfb:
@belisarius222 I've gone ahead and replaced the "punt" on decryption failure with `bail:evil` (non-deterministic), and special-cased that in the event dispatch logic to elide the error-notification event. This handles both DOS and verbose punt-logging issues.

belisarius222 (approved):
lgtm

joemfb:
@belisarius222 sorry for dripping these changes out. I realized that we never need stack traces for `bail:evil`, so I'm dropping them in `u3m_soft()`. And I decided to track them separately in the ames driver, and print a more specific error message. Pilots have really started depending on `sivc-de-punt` as an indicator of lagging azimuth state (see #5748), so I think we need to replace it with something else.

PTOML (one more)

PR #5827, opened 2022-06-04
ames: %clog less

belisarius222:
Resolves https://github.com/urbit/urbit/issues/6282.

Changes Ames to mark a flow as clogged when 100kB of data has been outstanding for 30 seconds, instead of 5 messages for 30 seconds.  Ames can measure its memory usage quite accurately, aside from deduplication among messages.

To come up with the 100kB number, I figured it would be good to limit Ames to 1GB of memory, assuming up to 10.000 peers.  This still gives each ship up to 100 packets worth of data, which could be 100 small messages or one large one.  I'm hoping that this will end up with similar Ames memory use to what we have right now (maybe a bit more? should test this on a real ship to tune it), but with much less frequent subscription kicking.

Subscription kicking itself triggers a bunch of extra packets and messages, and most ships that are offline are *very* offline, as in they haven't been online for days, so despite 5 packets per ship being common now and 100 per ship after this change, I still think memory usage might not be that different.


tacryt-socryp:
How expensive is running `(met 3 ...)` repetitively in this manner? Not aware of the performance characteristics of its jet.

belisarius222:
The `(met 3)` jet should just look at the size slot in the allocation cell, which says how many bytes the indirect atom has.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2192,14 +2192,15 @@
...
           %+  add  acc
           %+  add
-            ::  in-flight messages
-            ::
-            (sub [next current]:i.pumps)
-          ::  queued messages
-          ::
-          ~(wyt in unsent-messages.i.pumps)
+            =-  (roll - add)
+            (turn ~(tap to unsent-messages.i.pumps) |=(b=@ (met 3 b)))
```
Fang-, line 2196: Consider doing a single pass instead:

```suggestion
            %+  roll  ~(tap to unsent-messages.i.pumps)
            |=  [a=@ b=@ud]
            (add b (met 3 a))
```

Fang- (commented):
Hmm, how do we think the "one big message" case will behave? Feels like being allowed at least five messages gives you some leeway, what happens if we give that up?

belisarius222:
If the first subscription update is over 100kB, then Gall will kick just after it.  As long as the subscription is making progress (i.e. on resubscribe, the first message isn't the same as last time or another big initial-snapshot kind of message), then it should be fine.

If it's common for agents to use subscription patterns that don't make sufficient progress this way, then maybe we should change the clog criteria to "over 100kB total AND at least 2 messages".

Fang-:
> If it's common for agents to use subscription patterns that don't make sufficient progress this way

Well, we did observe /app/treaty doing poorly here. It seems not uncommon to have a subscription path that's "datum at this path, and any changes to it as they happen".  
Agents _could_ require a "last heard" timestamp in the subscription path, and keep a "last updated" timestamp in the datum, only sending out the datum if it's newer. But also that's dumb boilerplate and should be solved for elsewhere/some other way. (And also maybe proliferates ames flows?)

Fang-:
> then maybe we should change the clog criteria to "over 100kB total AND at least 2 messages".

We would be fine to say "at least 5 messages" here, right? In the hopes of not clogging _earlier_ in any case. If it turns out this increases memory usage significantly in practice we can always narrow down later, but seems desirable to keep this change strictly about clogging less, so we can judge changes in behavior more accurately.

belisarius222:
> We would be fine to say "at least 5 messages" here, right? In the hopes of not clogging earlier in any case. If it turns out this increases memory usage significantly in practice we can always narrow down later, but seems desirable to keep this change strictly about clogging less, so we can judge changes in behavior more accurately.

Yeah, good call, let's just do that.

zalberico:
Is there any reason this can't merge? Looks like it's done?

belisarius222:
It requires thorough testing, because it might increase memory use on publisher ships.  If it does, the constants would need to be tuned to use less memory.

zalberico:
Ok - sounds like a good candidate to merge and deploy on our test moons early in the new year. We'll want to also have it deploy to some publishers when it hits marnec. Unless you have a better idea of how we should plan to test and release.

Can you explain the benefit of clogging less? Why do we want this?

belisarius222:
@zalberico This has the potential to significantly reduce overall Ames network traffic.

The reason to %clog in the first place is to reduce memory pressure on the publisher.  This does still need to be done (until we have solid-state subscriptions, which give applications full control over memory use when publishing), but the constants involved have never been tuned.  This PR tunes them to clog less, and it also measures the actual memory use, instead of relying purely on number of messages.  Number of messages is a particularly bad proxy for memory use in the case of chat messages, which are usually under a kilobyte each.

This PR then also resolves a newly made issue: https://github.com/urbit/urbit/issues/6282

jalehman:
Next steps:

1. @yosoyubik is going to test this to ensure that it works and doesn't cause regressions.
2. Get a review from @joemfb 
3. Assuming it clears the above, merge to develop

yosoyubik:
Currently testing this on `~rontul-dantyr-norsyr-torryn`—I created a group there (`~rontul-dantyr-norsyr-torryn/clogger`) if people want to join and clog.

So far nothing stands out as broken/not working. Having another ship join the group and then turning it off before others post >5 (small) messages doesn't trigger the clogging logic and the pending chats are enqueued in the publisher:

Screen Shot 2023-02-05 at 10 09 07 AM

If one of those messages is >100Kb and there are more than 5 in the queue, %ames will emit a %clog to the vane as expected:

Screen Shot 2023-02-05 at 10 27 56 AM

(also confirming that the flow is then properly deleted on both publisher and subscriber)

Testing the real memory impact of this PR can be addressed once we deploy to ~binnec and have a couple of group publishers (UF ships?) synced to it and monitor its %ames memory usage. I don't think we have any tooling for that but a thread could be used to periodically dump that information somewhere, and also update %ames with a task that allows to tune the constants (5 messages; 100Kb of data) as desired—addressing that last bit in a separate PR.

jalehman:
Next step is to get this on ~bitbet-bolbel to find out how well it performs on a ship with real load. @zalberico, can you make sure that @yosoyubik gets access to that ship so he can test and monitor?

zalberico:
@yosoyubik follow up with @dosullivan to get the access you need, you'll need tailscale for that one. @jalehman this ship hosts the urbit community group, but not it's channels.

jalehman:
I see. I think the channels are the important part, right @belisarius222?

If so, we can get Jose access to those ships instead.

belisarius222:
Yes, channels are more likely to cause problems, so they're better to check.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2386,27 +2386,43 @@
...
+        ::  +nuf-memory: is enough memory used to mark as clogged?
+        ::
+        ++  nuf-memory
+          =|  mem=@ud
+          |-  ^-  ?
+          ?~  pumps  |
+          =.  mem
+            ;:  add  mem
+              %+  roll  ~(tap to unsent-messages.i.pumps)
+              |=([a=@ b=@ud] (add b (met 3 a)))
```
joemfb, line 2415: This could be further golfed for efficiency:

```
%+  add
  %-  ~(rep in unsent-messages.i.pumps)
  |=([a=@ b=_mem] (add b (met 3 a)))
?~  unsent-fragments.i.pumps. 0
(met 3 fragment.i.unsent-fragments.i.pumps)
```

that saves the list allocation (`+tap:to`) and one `+add` call (by initializing the `+rep:in` accumulator to `mem`). Using `+rep:in` on a `+qeu` looks strange, but works fine.

joemfb (approved):
this will do

I have an (optional) optimization to suggest:

yosoyubik:
This PR hs been deployed to `~bitbet-bolbel`—missing to get access to the following ships that host UC channels:
```
~darrux-landes
~dopzod
~bitpyx-dildus
```

%ames memory usage prior to the deploy:
bitbet-boldel



(PS: ~wolref-podlex host the "Broadscasts" notebook but I'm assuming that it will not run this)

yosoyubik:
Update on ~bitbel-bolbel after 4 hours:

Screen Shot 2023-02-09 at 6 04 23 PM

belisarius222:
Negligible extra memory use on one of the highest-load ships on the network.  Good sign.  Now that I'm thinking about it, it would be good to see how this affects a ship with heavy %treaty load, which would be any ship distributing software to a lot of other people.  I seem to recall @Fang- running this (or something similar?) on `~paldev` a while ago with good results.

zalberico:
Reminder that bitbel-bolbel hosts zero channels

belisarius222:
Oh, I thought it hosted some. I guess we don't have much of a test yet, then.

yosoyubik:
PR deployed to `~darrux-landes`:

Screen Shot 2023-02-10 at 3 53 49 PM

PR deployed to `~bitpyx-dildus`:

bitpyx

(I'll check again in some time with updates memory logs)

PS: still missing the right permissions for `~dopzod` (have already contacted the higher authorities)

PPS: ran `|ota %disable` on both so let's not forget to turn that on

yosoyubik:
After ~h15, looking good:

`~bitpyx-dildus` (slightly more memory in %peers-known, but seems constant):

Screen Shot 2023-02-11 at 8 13 15 AM

`~darrux-landes` (less memory in peers-know and overall less memory in %ames):

Screen Shot 2023-02-11 at 8 16 30 AM

belisarius222:
Ok, so we have:
```
~darrux-landes: 1.601GB -> 1.589GB
~bitpyx-dildus: 108.9MB -> 109.0MB
```

These changes are small enough that they might even be due to other factors, rather than this PR.  I deem this safe and effective.

zalberico:
Cool don’t forget to enable OTAs on these moons again :)

tacryt-socryp:
Is there evidence that this change does end up with substantially fewer clogs? The memory characteristics seem... so safe as to make me wonder if there's any real change in the underlying behavior.

zalberico:
@belisarius222 @yosoyubik - seeing some comms in uc general that people are having trouble connecting. I confirmed bitbet-bolbel is up, but can you take another look at its status?

zalberico:
darrus-landes crashed, I'm fixing it and bring it back up

zalberico:
I fixed it, I think UF group/channel host ship ~halbex-palheb is offline though and I don't have access to fix that one

yosoyubik:
@zalberico thanks for fixing it!

```
~darrux-landes: 1.601GB -> 1.589GB -> 1.601GB 
~bitpyx-dildus: 108.9MB -> 109.0MB -> 109.0MB
~bitbet-bolbel: 230.0MB ->   ...   -> 246.6MB        
```

~darrux-landes:

Screen Shot 2023-02-13 at 6 32 55 AM

~bitpyx-dildus:
Screen Shot 2023-02-13 at 6 38 37 AM

~bitbet-bolbel:

Screen Shot 2023-02-13 at 6 39 29 AM

joemfb (approved):
this will do

PR #5828, opened 2022-06-04
ames: ping faster

belisarius222:
Previously, despite the %ping app backing off only to 30 seconds, Ames itself could back off up to 120 seconds.  For other ships this might be fine, but for pinging a sponsor, the pings should go through fast enough to keep a firewall pinhole open.  I lowered the interval from 30 to 25 seconds for this reason and special-cased sponsor pings by looking at the duct (never look at the duct, he said).

NOTE: targeted against master for now, since next/arvo looks behind and I want these commits to be easy to see.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -528,6 +528,7 @@
           crypto-core=acru:ames
           =bug
       ==
+      ping=?  ::  is this a sponsor pinging channel?
       ::  her data, specific to this dyad
       ::
       $:  =symmetric-key
@@ -1245,7 +1246,8 @@
     ?>  ?=([@ her=ship *] u.parsed)
     =*  her          her.u.parsed
     =/  =peer-state  (got-peer-state her)
-    =/  =channel     [[our her] now channel-state -.peer-state]
+    =/  ping  ?=([[%gall %use %ping *] *] duct)
+    =/  =channel     [[our her] now channel-state ping -.peer-state]
     =/  peer-core    (make-peer-core peer-state channel)
     |^
     ?:  ?&  ?=([%new *] u.parsed)
@@ -1335,7 +1337,7 @@
       ^+  event-core
       =/  par  (get-peer-state her)
       ?~  par  event-core
-      =/  =channel  [[our her] now channel-state -.u.par]
+      =/  =channel  [[our her] now channel-state ping=| -.u.par]
       =/  peer-core  (make-peer-core u.par channel)
       =/  bones  ~(tap in ~(key by snd.u.par))
       |-  ^+  event-core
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -528,6 +528,7 @@
           crypto-core=acru:ames
           =bug
       ==
+      ping=?  ::  is this a sponsor pinging channel?
```
Fang-, line 531: Putting this in between the "our" and "her" sections is a bit weird, I'd put it before or after `now`.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1245,7 +1246,8 @@
     ?>  ?=([@ her=ship *] u.parsed)
     =*  her          her.u.parsed
     =/  =peer-state  (got-peer-state her)
-    =/  =channel     [[our her] now channel-state -.peer-state]
+    =/  ping  ?=([[%gall %use %ping *] *] duct)
+    =/  =channel     [[our her] now channel-state ping -.peer-state]
```
Fang-, line 1250:
```suggestion
    =/  ping=?       ?=([[%gall %use %ping *] *] duct)
    =/  =channel     [[our her] now channel-state ping -.peer-state]
```

Here and all other `=/  ping` instances below. Ames has some nice verticals going on, let's keep it up.

Fang-: Also there's a comment saying this is a nasty hack in one place, but not in here and the other places.

Fang- (commented):
>inspecting the duct

Dirty & gross. Given that "ping our sponsor to keep a pinhole open" is such an integral part of our networking logic, are we sure we don't just want to move it into ames, or at least make a special ames task for this kind of ping?

belisarius222:
I think there are three options to special-case pinging:
1) add an Ames task (or add a flag to the `%plea` task)
2) handle pinging in Ames
3) inspect the duct

3) is clearly the worst of those.  1) is probably the next easiest to implement.  2) would arguably be the best way to do it -- there's some flexibility in having pinging in userspace, but without the %ping app (or a replacement), Ames doesn't work, which suggests it should be in kernelspace, or at least Ames should privilege it somehow, like how Jael privileges the %azimuth agent as the one it listens to for PKI updates.

The behavior will be the same for any of these, though, so we could at least use this PR to test how much this change will help connectivity, and whether `~25` is indeed the right choice (it's the number Wireguard uses, but maybe we want a shorter interval given the likelihood of slowdowns due to load).

zalberico:
@belisarius222 - can you make a decision on this one so we can either move forward with it (or close).

jalehman:
"It's kind of gross, but it's gross in a fairly harmless way." 
~master-morzod

belisarius222:
Closing in favor of https://github.com/urbit/urbit/pull/6449

PR #5829, opened 2022-06-05
vere: change mapsize to 500 GiB on AAarch64 Linux

botter-nidnul:
Sets the mapsize for AAarch64 Linux in `pkg/urbit/vere/disk.c` to the largest size I can get to work with musl, 500 GiB.

Leaves the Windows mapsize untouched at the old default of 60 GiB because I haven't done any testing to see what the limit is there.


joemfb (approved):
LGTM, thanks

PR #5837, opened 2022-06-09
dojo: say how to beat the %dy-edit-busy escape room

### LINKED ISSUE #1462: Clarify %dy-edit-busy
by adding "press backspace to abort" or similar. (Esc or break might make more sense here semantically, actually.)

dnmfarrell:
It's too easy to get trapped in the Dojo %dy-edit-busy escape room. Just
type something like:

    -build-file /=base/gen/ls/hoon

This modifies the dojo output to tell the user how to get out.

Fixes #1462.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -735,7 +735,6 @@
       ^+  +>+>
       =^  dat  say  (~(transceive sole say) cal)
       ?:  |(?=(^ per) ?=(^ pux) ?=(~ pro))
-        ~&  %dy-edit-busy
         =^  lic  say  (~(transmit sole say) dat)
         =/  tip=@t  'dojo: busy (press backspace to abort)'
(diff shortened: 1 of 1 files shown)


dnmfarrell:
I wonder if this hint should be emitted as an effect, instead of changing the sigpam value?

Fang-:
> I ended up emitting an extra effect with the help text

This is the most real form of output, and as such the best solution. Good call.

> the pattern of using sigpam only to emit %tas.

It's only a pattern in that it's the most common way to easily show debug messages to the user. Of course, we should not be showing debug messages to our users in the first place!

I'd suggest removing the `~&  %dy-edit-busy` entirely, and printing just "dojo: busy (press backspace to abort)" or similar. (As you've done, I just want the "busy" keyword in there also.)

dnmfarrell:
Ahh that's gtk about the sigpam output. Thanks!

Fang- (approved):
Slick, thank you!

PR #5840, opened 2022-06-10
dojo: check =dir exists before switching

### LINKED ISSUE #1559: Dojo should validate =dir path before switching
**Original title:** Can't run generators while within /gen

If you
`~rachus_pilrem:dojo> =dir /=gen`
and then try to to run a generator such as `+ls %`, we get:
```~rachus_pilrem:dojo/gen> +ls %
fame: no /~rachus-salpub-pinsul-battux--nimbex-namled-moglus-pilrem/gen/0/gen/ls
\/ford: core /~rachus-salpub-pinsul-battux--nimbex-namled-moglus-pilrem/gen/~201
  09/gen/ls           
\/ 
```

stuff like `+hello 'yo'` gives the exact same thing. However, doing it this way is fine:

```~rachus_pilrem:dojo> +ls %/gen
> +ls %/gen
  al/hoon
  ask/admins/hoon
  brass/hoon
  bug/hoon
  capitalize/hoon
  cat/hoon
  code/hoon
```
and so and and so forth. This is only an issue while users are inside /gen, it seems. Running generators from inside /app is not a problem.

dnmfarrell:
Scries clay for a list of files beneath the provided path - if the response
is nil then the dir must not exist (clay abhors a vacuum).

Fixes #1559

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -569,8 +569,7 @@
                       ?:  ?=([@ @ ~] pax)  ~[i.pax i.t.pax '0']
                       pax
                   ?:  =(~ .^((list path) %ct pax))
-                    =-  +>(..dy (he-diff %tan - ~))
-                    rose+[" " `~]^~[leaf+"dir does not exist:" (smyt pax)]
+                    +(..dy (he-diff %tan 'dojo: dir does not exist' ~))
                   =.  dir  (need (de-beam pax))
                   =-  +>(..dy (he-diff %tan - ~))
                   rose+[" " `~]^~[leaf+"=%" (smyt (en-beam he-beak s.dir))]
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/dojo.hoon
```diff
@@ -568,6 +568,9 @@
                       ?:  ?=([@ ~] pax)  ~[i.pax %base '0']
                       ?:  ?=([@ @ ~] pax)  ~[i.pax i.t.pax '0']
                       pax
+                  ?:  =(~ .^((list path) %ct pax))
+                    =-  +>(..dy (he-diff %tan - ~))
+                    rose+[" " `~]^~[leaf+"dir does not exist:" (smyt pax)]
```
Fang-, line 573: Any reason to not do the simpler tang?

```suggestion
                    +>(..dy (he-diff %tan 'dir does not exist:' (smyt pax) ~))
```

Fang- (approved):
Sweet, thank you. The tang here can be simplified a little bit, but otherwise looks good.

pkg/arvo/app/dojo.hoon
dnmfarrell: Much cleaner, thanks! Have revised

PR #5841, opened 2022-06-10
dojo: setting eny/now/our shouldn't crash dojo

### LINKED ISSUE #1519: `=our` (or eny, or now) crashes dojo
Running `=our` without any noun (ie, clearing it) crashes + relinks dojo. This happens [here](https://github.com/urbit/arvo/blob/master/app/dojo.hoon?utf8=%E2%9C%93#L466).

That's not very user friendly. It should jump out a few lines earlier for these cases, and either silently do nothing, or warn the user that it's an environment variable and they can't touch that.

Attempting to set it gives you a stack trace with a helpful message in it, but doesn't relink dojo, so your command history remains in tact. Not super clean either, but it's better already.

dnmfarrell:
Changes dojo to emit a usage message and ignore the request.

Fixes #1519

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -532,10 +532,11 @@
       ?:  ?=([%show %3] -.mad)
         (dy-rash %tan (dy-show-source q.mad) ~)
       ?:  ?=(%brev -.mad)
+        ?:  ?=(?(%eny %now %our) p.mad)
+          (dy-rash %tan (cat 3 p.mad ' is immutable') ~)
         =.  var  (~(del by var) p.mad)
         =<  dy-amok
         ?+  p.mad  .
-          $?(%eny %now %our)  !!
           %lib  .(lib ~)
           %sur  .(sur ~)
           %dir  .(dir [[our.hid %base ud+0] /])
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/dojo.hoon
```diff
@@ -532,24 +532,24 @@
       ?:  ?=([%show %3] -.mad)
         (dy-rash %tan (dy-show-source q.mad) ~)
       ?:  ?=(%brev -.mad)
+        ?:  ?=($?(%eny %now %our) p.mad)
+          (dy-rash %tan [(crip "{(scow %tas p.mad)} is immutable")] ~)
```
Fang-, line 536: fyi, it's perfectly acceptable (and in fact, more common) to use `$?`'s shorthand, `?(%etc)`. But `$?` is also fine!

The `[]` here (and below) are superfluous though.

Lastly, both the `+crip` and `+scow` calls here are a bit off. `(scow %tas %something)` is equivalent to `(trip %something)`, and the latter is usually preferred. And if you're going to produce a `@t` tank instead of a `[%leaf tape]` one, you probably want to directly concatenate the two: `(cat 3 p.mad ' is immutable')`.  
(For more complex interpolations, of course, `leaf+"some {(com plex)} string"` is still more ergonomic and preferred.)

(To be clear, this is not a big deal, I'm just nit-picking style here and trying to convey nuances.)

```suggestion
        ?:  ?=(?(%eny %now %our) p.mad)
          (dy-rash %tan (cat 3 p.mad ' is immutable') ~)
```

Fang- (commented):
I don't even remember this issue, haha. Good patch, thank you!

The logic here can be made a little bit tighter, if you'll indulge me.

Fang- (approved):
Oops, sorry for the wait. This is all good!

PR #5843, opened 2022-06-12
vere: support home/end/del on vt terminals

dnmfarrell:
Modifies the terminal to support home, end and delete vt terminal
sequences (as typed by the Gnome Terminal app and others).

### DIFF
--- a/pkg/urbit/include/vere/vere.h
+++ b/pkg/urbit/include/vere/vere.h
@@ -155,6 +155,7 @@
         struct {                            //  escape code control
           c3_o    ape;                      //  escape received
           c3_o    bra;                      //  bracket or O received
+          c3_y    seq;                      //  vt sequence
         } esc;
 
         struct {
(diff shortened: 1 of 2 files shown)


Fang-:
Neat, thank you!

Going to request you re-make these changes on top of `m/drum-sessions`, which has all the latest `term.c` and related changes, and indeed also touches `_term_io_suck_char`. That branch is slated to go out Soon™, no other terminal work will see the light of day before that does, so probably best to just get this straight in there.

It's a little bit awkward how this is just sending ctrl+a and ctrl+e for home and end keypresses. The behavior matches up, but only because drum handles ctrl+a and ctrl+e in a way that matches the expected behavior of home and end. A different handler might have ctrl+a and ctrl+e mean completely different things.

Hesitating on whether it's ok to make this go in as-is. I certainly don't want to _force in_ `[%hom ~]` and `[%end ~]` belts right away. But we are going to have to answer questions about what set of inputs we want to support eventually... (Do we want function keys? Media controls? Joystick events?)

The `%del` one is 100% legitimate though, so at the very least I will accept that change. I'll think over the home and end key support a little bit more, and will get back to you. In the mean time, porting this over to `m/drum-sessions` in its entirety would be a good start.

dnmfarrell:
Thanks for the feedback! I agree. Closing this to submit a revised PR to `m/drum-sessions`.

PR #5845, opened 2022-06-12
term: support vt escape sequences

dnmfarrell:
Modifies the terminal to support vt escape sequences (as emulated by the
Gnome Terminal app and others).

For now it recognizes the delete key - other sequences are ignored as
drum does not (yet) support them.

cc @Fang-

### DIFF
--- a/pkg/urbit/include/vere/vere.h
+++ b/pkg/urbit/include/vere/vere.h
@@ -156,6 +156,7 @@
           c3_o    mou;                      //  M (for mouse event) received
           c3_y    ton_y;                    //  mouse button
           c3_y    col_y;                    //  column coordinate
+          c3_y    seq_y;                    //  vt sequence
         } esc;
 
         struct {                            //  input buffering
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -594,13 +594,15 @@
     ?<  ?=([?(%cru %hey %rez %yow) *] bet)            ::  target-specific
     =.  blt  [q.blt `bet]                             ::  remember belt
     ?-  bet
-      @         (ta-txt bet ~)
-      [%aro *]  (ta-aro p.bet)
-      [%bac *]  ta-bac
-      [%del *]  ta-del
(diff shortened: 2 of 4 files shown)


dnmfarrell:
Hey @Fang- thanks for your comments on #5843. I agree - making term.c convert home/end to ctrl-a/e expands it role from tokenizing inputs to also understanding the behavior of certain tokens (probably not good).
WDYT of this plan:
Make drum map unrecognized tokens to bel, as term does now. Then %hom, %end — whatever —, could be added to vere/term without requiring equivalent changes to drum. That simplifies term's role further to purely tokenizing input. Then the desired behavior can be added to drum later.

Fang- (changes requested):
I like the `%del` fix here and want it in. But I also don't want to commit to adding the home and end keys just yet.

If you can revert adaab02 (a revert commit is fine, that way we keep it accessible in history), and address the other feedback here, then I'd be happy to approve this!

dnmfarrell:
Hey @Fang- thanks for the review! I think I've addressed all the feedback, see what you think.

pkg/urbit/include/vere/vere.h
```diff
@@ -156,6 +156,7 @@
           c3_o    mou;                      //  M (for mouse event) received
           c3_y    ton_y;                    //  mouse button
           c3_y    col_y;                    //  column coordinate
+          c3_y    seq;                      //  vt sequence
```
Fang-, line 159: Oops, forgot to include this in the previous review. We always include the size of the integer in their variable names.

```suggestion
          c3_y    seq_y;                    //  vt sequence
```

Fang- (commented):
Thank you! One small thing I overlooked in my previous review, but otherwise this looks good to go.

Also discovered that the web terminal does _not_ handle the delete key properly, so the real terminal is once again reigning supreme. (^:

pkg/urbit/include/vere/vere.h
dnmfarrell: Ah good catch, I've fixed it up.

Fang- (approved):
All good now. Thank you!

PR #5847, opened 2022-06-14
Pedersen hash jet

martyr-binbex:
This is a jet for the pedersen hash. It copies joshin's hash implementation in zuse from his branch and fixes it to handle atoms bigger than 252 bits (by breaking them into chunks and hashing across them). The jet itself uses the libecc library which is a general purpose elliptic curve library in C.


pkg/arvo/sys/zuse.hoon
```diff
@@ -2155,6 +2155,114 @@
         [r s y.rp]
       ::  general recovery omitted, but possible
       --
+    ::
```
martyr-binbex, line 2158: This hoon function was [originally written](https://github.com/mrdomino/urbit/blob/jo/pedersen/pkg/arvo/sys/zuse.hoon#L2159) by @mrdomino but since it's in a branch in his fork I found it easier to just copy and paste it in here. I updated it to handle atoms bigger than 252 bits.

pkg/urbit/jets/e/pedersen.c
```diff
@@ -0,0 +1,100 @@
+
+/* j/5/pedersen.c
+**
+*/
+#include "all.h"
+#include <urcrypt.h>
+
+static c3_y*
+u3r_bytes_all_works(c3_w* len_w, u3_atom a)
```
frodwith, line 9: Style issue: rename this something meaningful, and make it a local name. Something starting with an underscore. Also, why is this function here? Please explain with a comment.

pkg/urcrypt/Makefile.am
```diff
@@ -32,14 +32,17 @@ liburcrypt_la_CPPFLAGS = -I$(srcdir)/ed25519/src \
 liburcrypt_la_LIBADD = $(LIBCRYPTO_LIBS) \
 		       $(LIBSECP256K1_LIBS) \
 		       $(LIBAES_SIV_LIBS) \
+           $(LIBEC_LIBS) \
```
frodwith, line 35: whitespace looks wrong

pkg/urbit/jets/e/pedersen.c
```diff
@@ -0,0 +1,100 @@
...
+  } else {
+    met_w = *len_w = u3r_met(3, a);
+  }
+  return u3r_bytes_alloc(0, met_w, a);
+}
+
+static c3_y *
+fold_hash(u3_atom x, c3_w *len_acc) {
+  c3_w len_x, pos = 0, ext_pos = 0;
+  c3_y *acc = NULL, *out = NULL, *ext = NULL;
```
frodwith, line 23: these should be static buffers, there is no reason to involve malloc here is there?

pkg/urbit/jets/e/pedersen.c
```diff
@@ -0,0 +1,100 @@
...
+    met_w = *len_w = 1;
+  } else {
+    met_w = *len_w = u3r_met(3, a);
+  }
+  return u3r_bytes_alloc(0, met_w, a);
+}
+
+static c3_y *
+fold_hash(u3_atom x, c3_w *len_acc) {
+  c3_w len_x, pos = 0, ext_pos = 0;
```
frodwith, line 22: please use the naming convention - lex_w, pos_w, exp_w for example.

pkg/urbit/jets/e/pedersen.c
frodwith: by static i mean stack allocated, a la

    c3_y acc_y[32];

pkg/urbit/jets/e/pedersen.c
```diff
@@ -0,0 +1,100 @@
...
+      if ((hed[31] & 0xf0) != 0) {
+        if (ext == NULL) {
+          ext = u3a_malloc(ext_len);
+        }
+        ext[ext_pos] = (hed[31] & 0xf0) >> 4;
+        ext_pos++;
+        hed[31] = hed[31] & 0x0f;
+      }
+    }
+    if (pos == 0) {
```
frodwith, line 42: very minor style issue, but please put extra spaces in the conditionals, and put the constants first in the comparisons, e.g.

    if ( 0 == pos_w ) {

pkg/urbit/jets/e/pedersen.c
```diff
@@ -0,0 +1,100 @@
...
+    } else {
+      if (out == NULL) {
+        out = u3a_malloc(32);
+      }
+      urcrypt_pedersen(acc, 32, hed, len_x - pos, out);
+      acc = out;
+    }
+    if ((len_x - pos) > 32) {
+      pos += 32;
+    } else {
```
frodwith, line 54: no cuddled elses please

pkg/urcrypt/urcrypt/pedersen.c
```diff
@@ -0,0 +1,264 @@
...
+  urcrypt__reverse(b_len, b);
+  if (a_len == 32) {
+    a[0] = a[0] & 0x0f;
+  }
+  if (b_len == 32) {
+    b[0] = b[0] & 0x0f;
+  }
+
+  if (curve_params == NULL) {
+    curve_params = (ec_params *)malloc(sizeof(ec_params));
```
frodwith, line 147: make some kind of urcrypt_pederson_init function (see urcrypt_secp_init and where it is called in vere), and don't use malloc at all.

pkg/urcrypt/urcrypt/pedersen.c
```diff
@@ -0,0 +1,264 @@
...
+    ret = get_curve_params(curve_params);
+    if (ret != 0) {
+      return ret;
+    }
+  }
+
+  if (cp == NULL) {
+    cp = (constant_points *)malloc(sizeof(constant_points));
```
frodwith, line 155: similarly here. actually there are a number of places in this PR where malloc is used unnecessarily.

pkg/arvo/sys/zuse.hoon
```diff
@@ -2155,6 +2155,114 @@
...
+    ++  pedersen
+      ~%  %pedersen  +  ~
```
frodwith, line 2160: I also am not seeing a ~/  %phash  hint anywhere... does this jet currently fire?

pkg/urcrypt/urcrypt/pedersen.c
```diff
@@ -0,0 +1,356 @@
...
+  prj_pt p4;
+} constant_points;
+
+struct urcrypt_pedersen_context_struct {
+  constant_points *cp;
+  ec_params *curve_params;
+};
+
+size_t
+urcrypt_pedersen_context_size()
```
frodwith, line 27: this whole size() concept is unnecessary for pederson -- secp uses it because the size is dynamic. This is a struct with a constant size, so you can do the whole thing with no malloc at all with the sizeof operator.

pkg/urcrypt/urcrypt/pedersen.c
```diff
@@ -0,0 +1,356 @@
...
+  ret = get_curve_params(context->curve_params);
+  if ( 0 != ret ) {
+    return -1;
+  }
+  return get_constant_points(context->curve_params, context->cp);
+}
+
+void
+urcrypt_pedersen_destroy(urcrypt_pedersen_context *context)
+{
```
frodwith, line 353: probably you don't need destroy function at all if the context isn't malloc'd, but you also didn't free the context here.

pkg/urcrypt/urcrypt/pedersen.c
```diff
@@ -0,0 +1,356 @@
...
+static int
+do_fold_hash(urcrypt_pedersen_context *cxt, uint8_t *dat_x, size_t len_x, uint8_t *out) {
+  int ret = 0;
+  int pos = 0, ext_pos = 0, ext_len = (len_x / 32);
+  uint8_t *hed = NULL, *acc = NULL, *ext = NULL;
+  while ( pos < len_x ) {
+    hed = (dat_x + pos);
+    if ( (len_x - pos) >= 32 && (hed[31] & 0xf0) != 0) {
+        if ( ext == NULL ) {
+          ext = calloc(1, ext_len);
```
frodwith, line 267: Do we need to conditionally allocate this inside the loop? Seems since we know its length, we're going to need to allocate it, no? I'm also a little confused by the calloc instead of malloc, and the arguments seem to be in the wrong order (not that that makes a difference).

zalberico:
Given this PR has not seen any action since June and is currently not complete I'm inclined to close it - @belisarius222 I'm assigning this to you to put this on your radar, if it's something that we should clean up and merge let me know, otherwise feel free to close it - we shouldn't leave it as is. Minimally we shouldn't merge it directly into master either way.

martyr-binbex:
> Given this PR has not seen any action since June and is currently not complete I'm inclined to close it - @belisarius222 I'm assigning this to you to put this on your radar, if it's something that we should clean up and merge let me know, otherwise feel free to close it - we shouldn't leave it as is. Minimally we shouldn't merge it directly into master either way.

I never merged it because we (uqbar) may need to make changes to the algorithm for indirect atoms because what I did may be slow in cairo. Also we may switch to the poseidon hash and not need pedersen at all. So we can close this PR. The code remains in my fork if we decide to optimize it and try to merge it in again.

belisarius222:
Roger that.  Thanks for the clarification, @martyr-binbex.  Closing for now.

PR #5850, opened 2022-06-15
groups: don't show url with images

patosullivan:
Fixes an issue where images would be displayed in chat along with the link to the image: 
image

Fixes regression introduced in #5729

### DIFF
--- a/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
+++ b/pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
@@ -34,26 +34,30 @@ interface GraphMentionNode {
   ship: string;
 }
 
-const addEmphasisToMention = (contents: Content[], content: Content, index: number) => {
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/landscape/components/Graph/GraphContent.tsx
```diff
@@ -116,72 +120,80 @@ const codeToMdAst = (content: CodeContent) => {
...
+          children: [
+            {
+              type: 'graph-reference',
+              reference: content.reference
+            }
+          ]
+        }
+      ];
+    } else if ('url' in content) {
+      const images = ['.jpg', '.jpeg', '.png', '.gif'];
```
nerveharp, line 153: + .webp? Your call

patosullivan: yes, good call.

arthyn (approved):
LGTM

PR #5851, opened 2022-06-15
gall: use the latest app list for %mult tasks

### LINKED ISSUE #5308: Apps on fresh ship don't get reloaded until hood does
1) Start a fakezod, with `-B bin/solid.pill` (with urbit 1.6 or earlier, doesn't matter)
2) `|mount %base`
3) Make a change to dojo, `|commit`
4) Observe the clay change notification and `kiln: commit detected`, but no `gall: reloading agents`
5) Make a change to hood, `|commit`
6) Observe the above, plus `gall: reloading` prints
7) Make a change to dojo again, `|commit`
8) Observe all the right printfs, dojo reloading, etc.

When I originally investigated this prior to release, I noticed that gall's `+mo-subscribe-to-agent-builds` gets called many times, with incrementally larger lists of agents that it actually subscribes to. Hood is the very first app it includes, and so is in the very first (and all subsequent) `%mult` request it sends out.

Interestingly, adding `/app/dojo/hoon` to the [set of default sources](https://github.com/urbit/urbit/blob/1e70a1be2a03a1ba60b7d06ce50f12305be62eb0/pkg/arvo/sys/vane/gall.hoon#L389) resolves this issue for dojo, and makes it so step 5 above (which is still required to trigger reloads of other apps) becomes "make a change to either hood or dojo".

All of this seems to me like the `%mult` requests that gall sends out from that arm beyond the first get swallowed, cancelled or otherwise lost. I was not able to figure out why/where at the time.

dnmfarrell:
When %gall installs %hood, %kiln adds %jolt moves for each app on %base
(in desk.bill). %gall queues its own move to update its %mult
subscription with the arvo sources + hood.hoon. This final move ends up
clobbering the %mult path list which has been updated processing the
%jolt moves.

This changes %gall to queue a move to update its %mult list when it
installs an agent. This prevents stale lists of agent paths being sent
to %clay if %gall installs an agent which itself installs other agents.

Fixes #5308

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:c45166ff0f8ab8dc1552bcef519c77c0afa6ca52f8ed1ba31ed632012667d619
-size 8674763
+oid sha256:cdb19d3dab0bd391eda26f72ff5f569a5662df518e00ee82aeed1d1653ad5f91
+size 8662713
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:cdb19d3dab0bd391eda26f72ff5f569a5662df518e00ee82aeed1d1653ad5f91
-size 8662713
+oid sha256:07d3d5191f0b28c49bea21de314476865f3e1ee974d5bed8cedf13f098113921
+size 8662473
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1648,6 +1648,7 @@
     $~  [%vega ~]                                       ::
     $%  [%deal p=sock q=term r=deal]                    ::  full transmission
         [%sear =ship]                                   ::  clear pending queues
+        [%trak ~]                                       ::  update source tracking
         [%jolt =desk =dude]                             ::  (re)start agent
         [%idle =dude]                                   ::  suspend agent
         [%nuke =dude]                                   ::  delete agent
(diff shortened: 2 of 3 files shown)


dnmfarrell:
This was the third fix I tested for this issue. I _thought_ it worked but testing it again, it doesn't. I must have been testing an older fix.

dnmfarrell:
A simpler fix would be to special-case installing %hood in `mo-handle-sys-cor`:
```
    ?: =(%hood dap)
      mo-core
    (mo-subscribe-to-agent-builds now)
```

As I guess %hood is the only agent which installs other agents? I don't like this as much as it makes %hood a special case which would have to be handled if circumstances change, but it does fix the issue too.

belisarius222:
I don't fully understand this issue or its fix.

The fix seems to be for gall to pass a task to itself upon receiving a newly built agent core from clay and adjust its clay subscription when handling that task, instead of adjusting its clay subscription synchronously when receiving the agent core.

Given Arvo's "depth first" move order, this task will be processed before the next %jolt from kiln, so as far as I can tell the only ordering difference from the fix would be that gall adjusts its clay subscription after the first gall activation finishes -- but that shouldn't matter, unless I'm missing something.

What goes wrong with the %mult in the current ordering?

dnmfarrell:
From what I saw, the jolts from kiln enter gall's move list _before_ (mo-receive-core %hood ...) completes, as they are internal to gall. Then mo-subscribe-to-agent-builds runs for %hood and appends a %mult move with the current sources. Arvo's depth first move order kicks in, processes %jolt %acme, and then %jolt %azimuth and so on. Eventually gall's move list is reduced to the original %jolt for %hood, which contains the stale sources list.

dnmfarrell:
Here's a [dump](https://gist.github.com/dnmfarrell/8dd10ae7b97f9e21aedccd0bab96b50d) of gall's move list after installing hood. At the head is the %mult move originating from installing %hood. As the list is flopped before it is returned, it's the last one processed.

belisarius222:
This should be obviated by the agents-in-Clay work.

—
~rovnys-ricfer
https://tlon.io

philipcmonk:
Yes, thank you for the work here, but this specific issue should be resolved now by the Kelvin 417 release.

Also, chalk this bug up as another reason to prioritize #6041, which I believe would have avoided this problem altogether.

PR #5852, opened 2022-06-16
ames: add rift when inserting new peer state

yosoyubik:
Keeping it as a draft until there's an actual fix. There's been some reports of dangling-bone crashes after the fix we deployed. I don't think the commit included here is the cause—if there's a breach, the right rift will be updated for the peer in question, instead of using 0.

My guess is that this is coming from old pre-nonce wire. Currently we don't drop https://github.com/urbit/urbit/pull/5732/commits/91435a92ca7c4d1968ee93ae6e2fbf6a339d55da those since it seemed bad to destroy existing flows, unless the two parties are aware of it, but if those wires belong to out-of-sync bones, the receiver won't have the bone in their ossuary, which will trigger the error.


zalberico:
I think this is no longer relevant after we shipped the gall request queue fix, but let me know if that's not the case.

PR #5853, opened 2022-06-16
Fix for loading buried message and collection list display

willbach:
component: Groups UI

This PR fixes 2 Groups bugs that I fixed in the course of updating EScape.

1. When going to a message that is more than 200 messages deep, the correct message is not fetched and the user is not shown the target message. This fix loads that message and the messages immediately before and after it so that the user is shown the target message.
2. When a collection is loaded in "List" view, the links are not marked as "read" and there are extraneous loading indicators at the top or bottom. Deleted links are also loaded and sometimes shown which causes the display to look/behave incorrectly.

### DIFF
--- a/pkg/interface/src/views/apps/links/LinkWindow.tsx
+++ b/pkg/interface/src/views/apps/links/LinkWindow.tsx
@@ -1,10 +1,10 @@
 import { Box, Col, Text } from '@tlon/indigo-react';
-import { Association, deSig, Graph, Group, isWriter } from '@urbit/api';
+import { Association, BigIntOrderedMap, deSig, Graph, GraphNode, Group, isWriter, markEachAsRead } from '@urbit/api';
 import bigInt, { BigInteger } from 'big-integer';
-import React, {
-  Component, ReactNode
-} from 'react';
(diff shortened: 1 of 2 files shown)


pkg/interface/src/views/apps/links/LinkWindow.tsx
```diff
@@ -122,14 +149,19 @@ class LinkWindow extends Component<LinkWindowProps, {}> {
...
         {/* @ts-ignore calling @liam-fitzgerald on virtualscroller */}
         <GraphScroller
           origin="top"
           offset={0}
           style={style}
-          data={graph}
+          data={orm}
           averageHeight={100}
-          size={graph.size}
+          size={orm.size}
```
liam-fitzgerald, line 164: Does the recreation (and subsequent loss of referential equality), impact render times due to lack of memoisation?

liam-fitzgerald (changes requested):
Thanks for the PR! I have some questions that should be addressed before merge.

zalberico:
No response to Liam's comments from June and work on groups 1 is unlikely to get attention. I'm going to close.

Issue #5854, opened 2022-06-16
Stack traces for Gall apps are unreasonably long


tacryt-socryp:
What can we do to remove the kernel lines from the trace? They're totally unnecessary in 99% of cases.

joemfb:
Remove the `!:` from the vane(s) in question.

tacryt-socryp:
I'm most concerned about clay stack traces while doing development. They're insanely long. @Fang-

PR #5855, opened 2022-06-17
gall: exclude from traces

Fang-:
Removes the `!:` at the top of gall, so that it no longer gets included in traces about agent builds or crashes.

We also refine intentional crashes with `~_`s, so that we still see a crash reason even if we don't get a full trace.  
Unintentional crashes obviously cannot get this treatment, and as such may not provide useful pointers for debugging when they get hit. RFC on this tradeoff. Perhaps there are some more-localized places we still want to put a `!:`?

Lastly, flops the trace for `+on-load` crashes, which were getting printed bottom-first.

Should alleviate #5854 at least partially.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1,5 +1,4 @@
-!:
-::  ::  %gall, agent execution
+::  %gall, agent execution
 !?  163
 ::
 ::::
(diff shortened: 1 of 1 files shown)


belisarius222:
I'd be more inclined to add a few !.'s around smaller pieces of code where errors are clearly in the agent, not in Gall itself.  Otherwise if Gall crashes we'll get less information.

—
~rovnys-ricfer
https://urbit.org

Fang-:
In that case earlier chunks of the gall trace would still be shown even for userspace crashes, no? ie the entire pathway from `+take` or `+call` all the way down to the `!.`.

It almost feels like we want a special tracing mode or hint where we say "disregard earlier `~_`s", or something along those lines. That way, expected/accounted for crashes (like userspace ones) can get full control over the output, while other ones get to stay as normal.  
Looks like `~>  %drop` (`u3t_drop`) is supposed to do something similar? But I can't get it to work, maybe I'm misunderstanding.

tacryt-socryp:
Please get this merged for the upcoming release :(

philipcmonk (approved):
This is a good set of changes.  I would also support removing the `!:` from Clay, I believe it was added in development and then just never got removed.

PR #5857, opened 2022-06-17
behn: fix timers not firing

### LINKED ISSUE #5747: %behn timers sometimes get stuck
%behn timers sometimes get stuck (don't ever fire for a particular agent), and then only fire when ship is ctrl+d'ed and started back up. cc: @dr-frmr @philipcmonk

joemfb:
This PR fixes #5747 and adds regression tests for %behn.

The bug was caused by %behn failing to emit a `%doze` effect (to set the next timer in the runtime) in certain necessary situations (itself caused by failing to clear the runtime state mirror `next-wake`, and unnecessarily comparing pending timers to the current timestamp). It also removes some unnecessary behavior: `%doze` was emitted unnecessarily, and occasionally with intentionally-skewed timestamps.

### DIFF
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -120,6 +120,7 @@
   ++  wake
     |=  error=(unit tang)
     ^+  [moves state]
+    =.  next-wake.state  ~
     ::  no-op on spurious but innocuous unix wakeups
     ::
     ?:  =(~ timers.state)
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/behn.hoon
```diff
@@ -191,10 +189,10 @@
       ?~  next-wake
         event-core
       (emit-doze ~)
-    ::  if :next-wake is in the past or not soon enough, reset it
+    ::  if :next-wake not soon enough, reset it
```
belisarius222, line 192: This comment is now not restrictive enough.  How about:

```
::   set new timer unless vere already has a timer set for that date
```

Fang-:
Running this on ~paldev now, who was consistently affected by the timer bug(s), and it's been behaving properly since applying these changes.

tacryt-socryp:
I'm a lot more worried about people being disconnected from the network semi-permanently than I am about a few weeks of faster log growth.

PR #5858, opened 2022-06-17
behn: refactor to use +abet pattern

### LINKED ISSUE #5747: %behn timers sometimes get stuck

joemfb:
This PR is built from my validation code for #5857. I still wasn't sure that %behn was handling all state transitions correctly, so I wrote a more explicit, declarative state machine to check its results. Once that I was done, I felt the result was clearer then the previous code.

#5857 has the necessary changes to fix #5747, this is additional cleanup that I think is worthwhile.

### DIFF
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -96,7 +96,7 @@
       %wake  (wake(next-wake.state ~) error)
     ==
   ::
-  ::  +take-drip:  XX
+  ::  +take-drip: the future is now, %give the deferred move
   ::
   ++  take-drip
     |=  [num=@ud error=(unit tang)]
@@ -115,7 +115,7 @@
   ::
   +|  %tasks
   ::
-  ::  +drip:  XX
+  ::  +drip: enqueue a future gift (as a vase), %pass ourselves a %wait
   ::
   ++  drip
     |=  vax=vase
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
I requested a comment change, but this LGTM.

pkg/arvo/sys/vane/behn.hoon
```diff
@@ -45,159 +45,109 @@
...
+    |=  vax=vase
+    ^+  this
+    %.  [duct %pass /drip/(scot %ud count.drips.state) %b %wait +(now)]
+    %=  emit
+      movs.drips.state   (~(put by movs.drips.state) count.drips.state vax)
+      count.drips.state  +(count.drips.state)
+    ==
   ::
-  ++  vega  [moves state]
   ::  +wake: unix says wake up; process the elapsed timer and set :next-wake
```
belisarius222, line 129: this looks like it no longer sets .next-wake, since that's now done in +abet

belisarius222 (approved):
LGTM, much cleaner

pkg/arvo/sys/vane/behn.hoon
joemfb: good catch.

PR #5859, opened 2022-06-17
npm: scrubbing every known place that sass existed

arthyn:
Ran into a lot of build/versioning issues with sass and indigo for some reason even though we aren't actually using it.


patosullivan:
I can run bootstrap just fine, but running build:prod from the root of the repo fails, showing a missing dependency error for react-codemirror2. Maybe bootstrap isn't properly cloning/installing git repo packages? After running npm install in the interface package itself I'm able to run build:prod from the root just fine.

Issue #5836, opened 2022-06-08
Full log replay: unavoidable mug mismatches (v1.5 -> v1.8)


aolieman:
**Describe the bug**
I want to do a full event log replay because I let the snapshot of a star get corrupted during a (dirty) reboot. The ship was last run successfully on v1.5 binaries, not with any more recent binaries. During log replay several binary versions (patched v1.5 & v1.8) start printing `mug mismatch` from a specific event onward. It hasn't been possible to do the full replay with a recent binary due to a jet mismatch. I've tried switching from v1.5 to v1.8 just before the mug mismatch starts occurring, but to no avail. Log replay with the v1.8 binary has eventually crashed on each attempt with the following error:
```
play: bail: %exit
play: %hear event on /a/ames failed
```

My attempts to complete the log replay have been informed by these helpful discussions: #5304, #5488, #5528, and #5551.

**To Reproduce**
Steps to reproduce the behaviour:
1. run a ship with the v1.5 binaries until April 2022;
2. corrupt the snapshot and remove north.bin & south.bin from `.urb/chk`;
3. attempt log replay with (patched) v1.5 binaries, and with v1.8 binaries.

**Expected behaviour**
I expected to be able to replay the event log and have my ship end up in the state it was in when it last ran. Fingers crossed that I don't have to breach.

**System:**
 - OS: x86-64, Linux 5.10.28-Unraid
 - Running Urbit in Docker (unofficial image)
 - 12 GB of available RAM
 - Vere v1.5 (see https://github.com/urbit/urbit/issues/5304#issuecomment-971719483)
 - Arvo 420 K (Urbit OS v2.114?)

**Additional context**
I'll try to take you through some of my thinking and the steps I took so far.
<details>
  <summary>Container logs show that the ship wasn't able to start immediately after a forced reboot of the host.</summary>
  
  ```
  2022-04-16T07:56:51.424645531Z ~winlur:dojo> 2022-04-16T07:59:22.016002068Z ~
  2022-04-16T07:59:22.805101152Z boot: core limit: Invalid argument
  2022-04-16T07:59:22.805140711Z loom: mapped 2048MB
  2022-04-16T07:59:28.063002953Z boot: protected loom
  2022-04-16T07:59:28.063033012Z live: loaded: MB/913.571.840
  2022-04-16T07:59:28.077097060Z home: bailing out
  2022-04-16T07:59:28.095745219Z urbit 1.5
[...]

Issue #5304, opened 2021-10-05
Urbit Crashing on Bootup after Dojo Stopped working


mopfel-winrux:
**Describe the bug**
I restarted my pier and urbit booted up as normal. Dojo took forever to load but after it finally did I would get
`gall: not running %hood yet, got %poke`
Whenever a keyboard input occured.

I restated urbit a few times with no such luck and the last time i restarted it I got:
```
~
urbit 1.5
boot: home is /home/XXXX/urbit/XXXXX
loom: mapped 2048MB
lite: arvo formula 79925cca
lite: core 59f6958
lite: final state 59f6958
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/720.781.312
pier: serf unexpectedly shut down
```

I asked for help on ~bitbet-bolbel/urbit-community/ and they recommended I run urbit with gdb and the core. Here is that output:
```
Core was generated by `./urbit-worker serf xxxx 0:0:0:0 128 50000 0'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x000000000042ecf2 in _box_free (box_u=0xc16d3422) at noun/allocate.c:183
183     noun/allocate.c: No such file or directory.
(gdb) thread ap[ply all bt
Invalid thread ID: [ply all bt
(gdb) thread apply all bt

Thread 1 (LWP 1256453):
#0  0x000000000042ecf2 in _box_free (box_u=0xc16d3422) at noun/allocate.c:183
#1  _box_free (box_u=0xc16d3422) at noun/allocate.c:161
#2  u3a_wfree (tox_v=0xc16d342a) at noun/allocate.c:632
#3  0x0000000200000000 in ?? ()
#4  0x000000020c660550 in ?? ()
#5  0x000000000042ed75 in _box_free (box_u=0xc3198154) at noun/allocate.c:240
#6  _box_free (box_u=<optimized out>) at noun/allocate.c:161
#7  u3a_wfree (tox_v=<optimized out>) at noun/allocate.c:632
#8  0x0000000000000000 in ?? ()
(gdb) 
```

**To Reproduce**
Steps to reproduce the behaviour:
1. run `./urbit ~my-planet`

**Expected behaviour**
urbit to load

**System (please supply the following information, if relevant):**
 - OS: Ubuntu 20.03 64-bit
 - Urbit 1.5
 - Can't access

amadeobellotti:
Running
`./urbit -F zod` runs as expected.

Fang-:
[...]

PR #5861, opened 2022-06-17
hoon: fix jet mismatch in +shas

joemfb:
This PR fixes jet mismatch in +shas, introduced with urcrypt in v1.6. `+shas` is used for %gall agent wires, and therefore ends up in %ames flows, so this mismatch has a dramatic and observable effect. It's the cause of the event log replay mug mismatches that effect piers booted before v1.6 and subsequently upgraded. #5304 and #5836 are representative examples. Thanks to @Fang- for tracking down an accessible reproduction, and @frodwith for helping to find the root cause.

This PR cannot be released before #5701.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -3283,7 +3283,8 @@
 ++  shas                                                ::  salted hash
   ~/  %shas
   |=  [sal=@ ruz=@]
-  (shax (mix sal (shax ruz)))
+  =/  len  (max 32 (met 3 sal))
+  (shay len (mix sal (shax ruz)))
 ::
 ++  shax                                                ::  sha-256
   ~/  %shax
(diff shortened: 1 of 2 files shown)


jalehman:
The C portion of this needs to make its way to the new urbit/vere repo. @joemfb , would you mind cherrypicking the relevant commits into a new PR on urbit/vere?

matthew-levan:
I ported the `vere` code over: https://github.com/urbit/vere/pull/309.

jalehman:
Should this retarget next/kelvin/412 @belisarius222 ?

joemfb:
I've updated this PR to resolve the +shas jet mismatch in favor of the jet. `+shas` is used indirectly (via `+shaf`) in moon and comet keys, `+code`, and eyre's sessions. Fixing the jet to match the hoon is simply not worth the trouble.

/cc @frodwith

joemfb:
#6615

PR #5864, opened 2022-06-21
http-api: prevent reuse of old subscription ids

arthyn:
There's a latent bug in the http-api package which will erroneously send subscription responses with the wrong ids because they get reset mistakenly. This moves all setting of `this.lastEventId` to only the `getEventId` function so they stay consistent and ordered.


liam-fitzgerald (approved):
LGTM, good catch

PR #5866, opened 2022-06-23
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>


zalberico:
Thank @naveensrinivasan - though looks like this is currently causing tests to fail:
```
Error: google-github-actions/auth failed with: retry function failed after 1 attempt: the GitHub Action workflow must specify exactly one of "workload_identity_provider" or "credentials_json"! If you are specifying input values via GitHub secrets, ensure the secret is being injected into the environment. By default, secrets are not passed to workflows triggered from forks, including Dependabot.
```

DaneBettis:
I am also hitting this auth issue and I have not (so far as I know) altered any of these files myself.
https://github.com/urbit/urbit/pull/5967
https://github.com/urbit/urbit/actions/runs/3739450106/jobs/6346677937

I'm currently looking for the root cause of this in my PR's CI

DaneBettis:
my branch is based on next/vere and I altered .github/workflows/vere.yml to be identical to the one in master. This seems to have resolved this issue.

So I suspect that rebasing next/vere from master will resolve this particular error.

Just my 2 cents.

jalehman:
A lot has changed with our GitHub actions since this was opened. Going to close, and if it's still something that's needed we can discuss.

PR #5868, opened 2022-06-25
dns: transfer ship.arvo.network domains from ~zod to ~deg

pkova:
As discussed with @Fang- out of band, the Foundation will take ownership of `ship.arvo.network` domains. This PR points `/ted/dns/address` to the Foundation operated `~deg` instead of Tlon operated `~zod`. It also includes some marks for my DNS automation.

I'll coordinate with @Fang- to switch DNS over when the OTA with this commit hits the network.


Fang- (approved):
Yep, this checks out. glhf!

PR #5870, opened 2022-06-27
groups: pre-ota fixes omnibus

arthyn:
This resolves a few issues:

- following paragraphs erroneously being included in blockquotes
- channel descriptions being rendered incorrectly because of latest react-markdown version
- incorrect background in notebook editor on dark mode


patosullivan (approved):
LGTM. Builds and works well on my machine.

PR #5871, opened 2022-06-30
clay: clear ford cache on +load instead of +stay

### LINKED ISSUE #5849: `%clay` fails to compile when `$type` has a nesting change
**Describe the bug**
Before the recent changes to Clay, if you changed `type` in such a way that the old `type` nested in the new `type`, everything compiled fine with no complaints.

Now, if you do something like add a tag `[%foo ~]` to `$hoon`, everything compiles fine until it gets to `%clay`.

**To Reproduce**
1. Add `[%foo ~]` to the `$%` in `$hoon`.
2. `|commit %base`
3. Output:
```
...
>   arvo: +load next
lull: ~daptyl-fodsen
zuse: ~hadmel-rigrel
vane: %ames: ~hosret-lassym
vane: %behn: ~sartes-masnyl
vane: %clay: ~haptyv-natnev
crud: %into event failed
[%poke %into]
%load
- need
(gigantic type that seems to consists entirely of clay data types)
```

**Expected behaviour**
It compiles without error.

**Additional context**
@philipcmonk suggested that the issue is with `+stay` or `+load` in `clay.hoon`. Per his suggestion, I removed the two casts to `flue` in `+stay` and this fixed the bug, but he suggested that there was a better way to fix it in `+load`.  This change must be made and committed before making the change to `$hoon`.

I will be pushing changes to `$hoon` when doccords is released in the near future, so this will need to go out in an OTA before doccords.

philipcmonk:
Best practice is to not remove

fixes #5849

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -4632,12 +4632,55 @@
             [%7 raft-7]
             [%6 raft-6]
         ==
-      +$  raft-13  raft
+      +$  raft-13
+        $:  rom=room-13
+            hoy=(map ship rung-13)
+            ran=rang
+            fad=*
+            mon=(map term beam)
+            hez=(unit duct)
+            cez=(map @ta crew)
+            pud=(unit [=desk =yoki])
+            sad=(map ship @da)
+            bug=[veb=@ mas=@]
+        ==
+      +$  room-13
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Might want a note above `+$  raft-13` about why it's there when `+$  raft` also exists, but otherwise looks good!

PR #5872, opened 2022-07-01
azimuth: fetch snapshot on boot

philipcmonk:
Previously, the initial Azimuth snapshot was stored in Clay and shipped
in the pill.  This causes several problems:

- It bloats the pill
- Updating the snapshot added large blobs to Clay's state.  Even now
  that tombstoning is possible, you don't want to have to do that
  regularly.
- As a result, the snapshot was never updated.
- Even if you did tombstone those files, it could only be updated as
  often as the pill
- And those updates would be sent over the network to people who didn't
  need them

This moves the snapshot out of the pill and refactors Azimuth's
initialization process.  On boot, when app/azimuth starts up, it first
downloads a snapshot from bootstrap.urbit.org and uses that to
initialize its state.  As before, updates after this initial snapshot
come from an Ethereum node directly and are verified locally.

Relevant commands are:

- `-azimuth-snap-state %filename` creates a snapshot file
- `-azimuth-load "url"` downloads and inits from a snapshot, with url
  defaulting to https://bootstrap.urbit.org/mainnet.azimuth-snapshot
- `:azimuth &azimuth-poke-data %load snap-state` takes a snap-state any
  way you have it

Note the snapshot is downloaded from the same place as the pill, so this
doesn't introduce additional trust beyond what was already required.
When remote scry is released, we should consider allowing downloading
the snapshot in that way.

Tagging @joemfb primarily for conceptual review and @yosoyubik for implementation.

Mitigates #5748, possibly solves it


joemfb (approved):
This looks like a massive improvement. We might want to integrate it with boot differently in the future, but those mechanisms aren't ready yet (BYOBS). The poke-handling here is preparation for that anyway.

- Should also automatically tombstone the azimuth-snapshot in %clay when this is released?
- `-azimuth-load` without arguments will be the standard way to restore/reestablish azimuth state?

LGTM

joemfb:
I'm not sure why the test is failing this way -- it's a pre-existing issue on `next/arvo`.

yosoyubik (dismissed):
This is great, very happy about releasing it. LGTM!

philipcmonk:
Tested on ~wicdev-wisryt, and this uncovered another bug, fixed in 43ad38cd8

Issue #5838, opened 2022-06-09
batch comment doccords for ++ and +$ arms


drbeefsupreme:
A batch comment is a list of doccords for different parts of the code written in one block for the sake of readability. A good example of this is in `arvo.hoon`:

```
::  $arch: node identity
::  $axal: fundamental node, recursive (trie)
...
::
+$  arch  (axil @uvI)
++  axal
  |$  [item]
  [fil=(unit item) dir=(map @ta $)]
...
```

For the first release of doccords, we're keeping this simple: only comments for `++` and `+$` arms, written inside the core that contains them. Building this out should leave a skeleton for expanding this to work with comments written outside of the core, comments for cores/chapters/etc written like this, etc, intended for a future release.

drbeefsupreme:
This is working now, but the code is a mess, so I'm rewriting it.

Issue #5785, opened 2022-05-24
Doccords v0 Release


drbeefsupreme:
There are two main issues with doccords release:
 - Memory usage upon upgrade
 - Changes to the binary to jet the changes to the parser

Updating `hoon.hoon` means rebuilding everything, so as an OTA it has a high
memory requirement. Thus we probably ought to wait until @philipcmonk 
completes changes to Ford that are designed help with the OTA memory usage
issue.

I don't really understand what would be involved with changes to the binary. I
should meet with @joemfb to help figure this out.

drbeefsupreme:
#5849 needs to be fixed in a separate OTA that goes out before doccords release.

zalberico:
Last week:
 - The `::` syntax is working!

This week:
 - Knocking out a bunch of small issues
 - Refactoring the parser to clean things up prior to release
 - Jon ran into an issue with the Ford rune parsing, would be helpful to look it over with @Fang- 

By EOW should be ready for code review of `hoon.hoon` changes - Jon will ping the engineering chat when that's ready.

zalberico:
What remains (summary from @drbeefsupreme in team sync):
 - Going through each rune one by one and making sure comments attach as they're supposed to
 - Get the `hoon.hoon` PR ready for review
 - A little bit of refactoring work to get the parser in a good state
 - Then moving to the printing library: sometimes when asking the dojo for a doccords on a type it hangs for a long time, not sure why (may need a timeout here).

drbeefsupreme:
Current branch for `::` syntax changes:

https://github.com/urbit/urbit/tree/jon/doccords-parser-simplify - pardon the unhelpful commit messages, I'm planning on squashing them all into one when merging back into this branch.

It's cooled down from "exploded whitespace parser" back into something that thankfully looks a lot like the parsing for `:>` and `:<` syntax with a handful of new whitespace parsers in `+norm` that fit the pattern of the existing ones.

I'm expecting it will be good for review sometime this week (hopefully tomorrow).

zalberico:
From talking with Jon in 1:1
 - This may require a new runtime version for some jet related stuff (if so we should create an issue with details and link it here)
 - #6052 is separated out because it would require a hoon kelvin version change, but since it doesn't require app developers to change things we should do it and ship it.

zalberico:
Only remaining thing here is @drbeefsupreme needs to go over jet signatures with @joemfb, otherwise we think we're ready for release.

[...]

Issue #5784, opened 2022-05-24
Doccords for standard library functions


drbeefsupreme:
first release ought to come with doccords for the standard library so that you can ask dojo what the standard lib functions do

drbeefsupreme:
This is already partially implemented, in the sense that `hoon.hoon` already gives basic descriptions for many standard library functions and doccords makes them accessible in dojo. But it is accurate that I was not planning on copying the `urbit.org` docs into `hoon.hoon` for the first release.

PR #5873, opened 2022-07-01
hoon, dojo: initial doccords release

drbeefsupreme:
This is a draft since its still a WIP, so there's some code that is obviously sloppy and can be improved, but major features are all working and all tests are passing. A couple minor features still have work to be done.

Short story: you can now wrap an arbitrary `hoon` with `%note` `hoon` containing doccords. You can wrap an arbitrary spec with a `%dict` `spec`, containing doccords. Both end up compiled as `%hint` types.

Surfacing doccords for arbitrary hoons isn't currently implemented in the printing library, but you can print doccords for cores, chapters, faces, arms, arm products, default arm of a core built by an arm, and molds using `# arm/chapter/core/face`, and output values of molds using `?? *foo` in dojo.

OTA is blocked by #5849 which https://github.com/urbit/urbit/pull/5871/commits/172af42def107f93627f5b6aed6a17472fc700e7 is a suggested patch for. I had tried cherry picking this commit, but a fresh fake ship wouldn't compile with it, so I'm probably missing other things from that branch for it to work just yet and so haven't tested the fix with this feature branch. The band-aid in #5849 still works if you'd like to test this out in the meantime.

Examples of doccords syntax can be found in `/lib/dprint.hoon`, `/lib/deco.hoon/` and `/tests/lib/dprint.hoon`. There's a lot of room for change here - I'll work on an exhaustive list of syntax changes and update this PR soon. The biggest difference is that I plan to replace `:<` and `:>` with `::` with some strict rules on what `::` comments should be parsed as formal comments. This might end up being pretty difficult.

All planned features for the first release are essentially complete (#5768 #5769 #5770 #5771 #5773 #5774 #5775 #5776 #5777 #5778 #5779 #5780 #5781 #5782 #5783 #5784 #5785 #5838) - what's mostly left is arguing over syntax and making the code cleaner.

----
Some example usage of printing library (output is colorized in dojo - that section of the printing lib is still rough, it needs another look)

```
>  =dp -build-file %/lib/dprint/hoon
> # dp
.dp
(undocumented)

core containing doccords search and printing utilities
|dprint
(undocumented)

types used by doccords
|dprint-types
(undocumented)

.debug
(undocumented)
```

```
> # dp/dprint
|dprint
core containing doccords search and printing utilities
chapters:
|dprint|printing
contains arms using for printing doccords items

|dprint|searching
contains arms used for looking for docs inside of a type

compiled against:
types used by doccords
|dprint-types
(undocumented)

.debug
(undocumented)
```

```
> # dp/dprint/searching
|searching
contains arms used for looking for docs inside of a type

the entrypoint for finding docs within a type is +find-item-in-type.
arms:
+all-arm-docs
grabs the docs for an arm.

+arm-and-chapter-overviews
returns an overview for a cores arms and chapters

+arm-product-docs
returns 0, 1, or 2 whats for an arm

+arms-as-overview
translate a tome into an overview
...
```

```
> # dp/dprint/searching/find-item-in-type
+find-item-in-type
returns the item to print while searching through topic

this gate is optionally called recursively to find the path (topic) in the
type (sut). once it finds the correct part of the type, it switches to
+signify to describe that part of the type. recursion is turned off (for some
cases) when a hint type is found, in case it is wrapping a match.

product:

(undocumented)

default arm in core:

(undocumented)
```

Doccords from output value of a mold:

```
> ?? [*term *tape]
  [%cell <|ascii symbol|> <|utf8 string as list|>]
```


pkg/arvo/sys/hoon.hoon
```diff
@@ -7198,7 +7200,6 @@
   |=  [p=(pair type note) q=type]
   ^-  type
   ?:  =(%void q)  %void
-  ?:  =(%noun q)  %noun
```
drbeefsupreme, line 7201: leaving myself a reminder to try and remember why I did this and whether I actually still need it

pkg/arvo/sys/hoon.hoon
```diff
@@ -11592,37 +11612,61 @@
...
+          :-  '+'
+          ;~(pfix lus (stag %funk sym))
+          :-  '$'
+          ;~(pfix buc (stag %grog sym))
+          :-  ' '
+          ;~(pfix step (stag %funk (easy %$)))
+      ==
+    ++  shot
+      ;~(plug ;~(sfix line (just `@`10) (punt gap)) (rant ;~(less tine text)))
+    ::
```
drbeefsupreme, line 11633: I knew the parser for batch comments wasn't quite right - looking at it again today, I'm pretty sure I made use of the `def` face in `$whit` incorrectly. I was thinking that that would be where things like docs about the sample of the gate would go. That's true, but I think it's intended for batch comments in general. Right now I have split doccords parsing into two arms, `apex:docs` and `vext:docs` with the latter parsing a `(list whit)` intended for batch comments and the former just a `whit`. But looking to the future of how batch comment parsing ought to work, there should be only one function parsing a single `whit`. I think the new definition of `whit` would look like

```
+$  whit
  $:  lab=(list link)
       boy=(unit (pair cord (list sect))
       def=(map (list link) whit)
       use=(set (list link))
  ==
```
Or maybe `def=(map (list link) (pair cord (list sect)))`. Either way I knew I needed to switch to lists of links, but I need to rewrite more of the batch comment stuff than I had expected for this to be suitable for release.

drbeefsupreme: https://github.com/urbit/urbit/pull/5873/commits/de19456f82b03b41d952b36b5c73b07e1f21ed18

drbeefsupreme:
OK, I ripped out and rewrote the entire parser for doccords since I totally changed `$whit` and everything seems to be working now. I'm actually happy with the parser at this point.

https://github.com/urbit/urbit/blob/jon/doccords/pkg/arvo/lib/deco.hoon

is the best file to look at for style examples. It has everything that works, plus a few more things that parse but don't end up in the compiled type yet (like doccords on the sample of a gate).

The only thing left now is to switch the syntax from `:>` and `:<` to `::` which I haven't really tried to do yet but I'm guessing will just consist of hunting down a bunch of annoying edge cases. See the above link for syntax of how we're avoiding "bad doccords". So I'd say its ready for review at this point @belisarius222 @joemfb

pkg/arvo/app/dojo.hoon
```diff
@@ -679,11 +704,16 @@
...
+                     [%know *]
+                   ?@  p.q.p.a  [(cat 3 '#' mark.p.q.p.a)]~
+                   [(rap 3 '#' auth.p.q.p.a (spat type.p.q.p.a) ~)]~
+                 ::
+                     [%help *]
+                   [summary.crib.p.q.p.a]~
+                 ==
       [%core ^]  `wain`/core
-      [%hold *]  a(p $(a p.a))
+      [%hold *]  $(a (~(play ut p.a) q.a))
```
drbeefsupreme, line 716: Playing `%hold`s (i.e. what we're doing here, not sure if there's an accepted term for this) can take a long time and often this just crashes. Is there a better option? I suppose it could just give up after some chosen depth, but this isn't really user friendly. It ought to "just work". Besides limiting by depth, the only other idea I have is Clay building a cache of doccords.

pkg/arvo/app/dojo.hoon
```diff
@@ -663,6 +679,15 @@
...
           t=(turn `wain`?~(r.hit ~ (to-wain:format q.u.r.hit)) trip)
       ==
+    ::
+    ++  dy-inspect
+      |=  [topics=(list term) sut=type]
+      %+  dy-rash  %mor
+      =+  to-display=(find-item-in-type:dprint (flop topics) sut)
+      ?~  to-display
+        [%tan [%leaf "Could not find help"]~]~
+      [(print-item:dprint u.to-display)]
```
drbeefsupreme, line 689: There's no way to look inside of unnamed cores that aren't built by an arm - which will be all of them at release. I need some kind of syntax to look inside of an unnamed core - maybe just `$` for now

`/foo/hoon`:
```hoon
|%
::  +bar: a bar
++  bar  ~
--
```
then
```
=a -build-file %/foo/hoon
# a$+bar
```

drbeefsupreme:
I spent the last 4 days or so trying various strategies for changing the doccords syntax to `::` and basically came the conclusion that the only right way to do this is to unify whitespace and doccord parsing. Meaning, doccord parsing should happen inside of `+gap` and any callsites for `+gap` need to know how to handle something like a `(unit whit)`. For example, it should be thrown out in the case of something like the whitespace between `++` and the name of an arm, or wrapped around a `hoon` or `spec` where appropriate.

This is mostly because every other way I tried to do this ended up turning into a thicket of edge cases and bespoke code, or would if I kept going, and it wouldn't have been clear that the parser was correct. Language parsers need to be obviously correct, so that reason alone is enough for me to think this is the right way.

I'm going to pair with @Fang- to get a second opinion.

drbeefsupreme:
So @Fang- and I paired last week and got a parser that seems to work without unifying doccords and white space parsing - instead, the doccords parser just inserts a newline into the tape to be parsed if it successfully parses a doccord so that it may be eaten by a `+gap`, and if there are no doccords, `+gap` will eat what is there anyways.

This lead into another issue I had encountered before but ignored at the time since things were still too much in flux to know whether it was a core issue or a fluke of where I was in the process.

The compiler fails here:

https://github.com/urbit/urbit/blob/jon/doccords-add-newline/pkg/arvo/sys/zuse.hoon#L5879-L5886

```hoon
    ++  furl                                            ::  unify changes
      |=  [one=toro two=toro]
      ^-  toro
      ~|  %furl
      ?>  ?&  =(p.one p.two)                            ::  same path
              &(?=(%& -.q.one) ?=(%& -.q.two))          ::  both deltas
          ==
      [p.one [%& (weld p.q.one p.q.two)]]
    --
 ```

with
```
>   arvo: +load next
lull: ~daptyl-fodsen
zuse: ~tardec-bitful
crud: %into event failed
[%poke %into]
%load
find-fork
-find.p.q.one
zuse: build failed
```

If I remove the two postfix comments on the `?>` children, it compiles fine (then fails when it starts compiling `%base`, but let's tackle this first since maybe its the same problem). I suspect the issue is how the compiler handles wut runes. This is how the above expression parses:

```hoon
        { [ p=%furl
              q
            [ %note
                p
              [ %help
                  p
                [ cuff=~[[%funk p=%furl]]
                  crib=[summary='unify changes' details=~]
                ]
              ]
                q
              [ %brts
                  p
                [ %bccl
                    p
                  [ i=[%bcts p=term=%one q=[%like p=~[%toro] q=~]]
                    t=~[[%bcts p=term=%two q=[%like p=~[%toro] q=~]]]
                  ]
                ]
                  q
                [ %kthp
                  p=[%like p=~[%toro] q=~]
                    q
                  [ %sgbr
                    p=[%rock p=%tas q=1.819.440.486]
                      q
                    [ %wtgr
                        p
                      [ %wtpm
                          p
                        ~[
                          [ %note
                              p
                            [ %help
                              p=[cuff=~ crib=[summary='same path' details=~]]
                            ]
                              q
                            [ %dtts
                              p=[%wing p=~[%p %one]]
                              q=[%wing p=~[%p %two]]
                            ]
                          ]
                          [ %note
                              p
                            [ %help
                                p
                              [cuff=~ crib=[summary='both deltas' details=~]]
                            ]
                              q
                            [ %wtpm
                                p
                              ~[
                                [ %wtts
                                  p=[%leaf p=%f q=0]
                                  q=~[[%.y p=2] %q %one]
                                ]
                                [ %wtts
                                  p=[%leaf p=%f q=0]
                                  q=~[[%.y p=2] %q %two]
                                ]
...
```
Note in particular the `%note`s on the `%wtpm` children. I suspect that the `?>` is failing to inform the compiler enough information to know that `p.q.one` exists due to the `%note`s on the conditional statements (which will end up compiling as `%hint` types, so basically I think most wut runes and maybe `.=` aren't treating `%hint` types appropriately.

While the dumps above were produced from the `jon/doccords-add-newline` branch (which changes the syntax to `::`) I can reproduce the issue on the current branch by replacing the `::` in the troublesome comments with `:<`, so it doesn't seem related to our parser changes.

I tried some really basic things - like making it so you walk through `%note`s for `%wtpm` `%wtbr`, `%dtts` and whatever else made sense in `+mint` and `+open`, but it didn't budge. When I was working with @Fang- we tried using the new `?#` but that also didn't help. I only spent about 20 minutes on this so there's plenty of unexplored solution space and I might not have made all the changes to `+mint` and `+open` needed (assuming that's even the right place to look).

joemfb:
Why is `[%note *] q.gen` commented out in `+open:ap`? That line combined with the fixpoint loop in `+chip:ut` should ensure that `%note`'s are expanded out during branch specialization.

drbeefsupreme:
Good catch! I think I commented that out months ago, not really knowing what I was doing, probably thinking that it was telling the compiler to skip over `%note` hoons entirely. Uncommenting it in the current branch works . It runs into a parsing issue on the `::` syntax branch, but that's likely an issue with our new parser.

drbeefsupreme:
This PR is on pause until after assembly. The fix to the wut rune issue above worked, and everything compiles now, but `+gap` is eating prefix comments. There's also still some small improvements and features to make to the printing library. I'll be focusing on userspace permissions in the meantime, but still welcome any comments about syntax improvements etc.

drbeefsupreme:
Finally working on this again. Seem to be making real progress on changing the syntax to `::`. Current branch is https://github.com/urbit/urbit/tree/jon/doccords-gap-conquering (note there's a bunch of junk code in it right now) I think I'll be reaching testing parity with the branch in this PR later this week or early next week, then hopefully merging it into this one after some refactoring.

Then all that's left is some fixes/additions to the printing library and we should be ready for an actual review.

pkg/arvo/app/dojo.hoon
drbeefsupreme: going to be revisiting this now that we're not naming cores in the first release

drbeefsupreme:
OK, `hoon.hoon` is ready for review. I'll start on the printing library tomorrow.

@joemfb @Fang- @belisarius222

pkg/arvo/sys/hoon.hoon
```diff
@@ -11534,114 +11557,175 @@
...
-    ::  into: :: and indent to end of line, consuming following space.
-    ::  indo: :: to end of line, consuming following space.
-    ::  exit: :: to end of line, not consuming following space.
+    ++  rant
+      |*  sec=rule
+      %-  star
+      ;~  pfix
+        (into null)
+        (plus (into sec))
+      ==
```
Fang-, line 11657: Likewise, this and its two callsites feel like a somewhat poor generalization. Can you describe what `+rant` does in a short sentence? If not, may be more clear to just duplicate its logic instead.

What particularly jumps out to me is how it gets called with parsers that start by eating a `step` (or two), even though `into` eats as many `ace`s as it can at the end. ...And `into` is only used here afaict.

pkg/arvo/sys/hoon.hoon
```diff
@@ -11534,114 +11557,175 @@
...
+          :-  '%'
+          ;~(pfix cen (stag %cone bisk:so))
+          :-  '^'
+          ;~(pfix ket (stag %core sym))
+          :-  '_'
+          ;~(pfix cab (stag %door sym))
+          :-  '='
+          ;~(pfix tis (stag %gate sym))
+          :-  '/'
+          ;~(pfix fas (stag %path sym))
```
Fang-, line 11696:
```suggestion
      :~  :-  '|'  ;~(pfix bar (stag %chat sym))
          :-  '.'  ;~(pfix dot (stag %frag sym))
          :-  '+'  ;~(pfix lus (stag %funk sym))
          :-  '$'  ;~(pfix buc (stag %plan sym))
          :-  '%'  ;~(pfix cen (stag %cone bisk:so))
          :-  '^'  ;~(pfix ket (stag %core sym))
          :-  '_'  ;~(pfix cab (stag %door sym))
          :-  '='  ;~(pfix tis (stag %gate sym))
          :-  '/'  ;~(pfix fas (stag %path sym))
```

pkg/arvo/sys/hoon.hoon
```diff
@@ -11534,114 +11557,175 @@
...
+        [%gist c b]
+      =+  docs=~(tap by bat.a)
+      |-
+      ?~  docs  b
+      $(docs t.docs, b [%gist i.docs b])
+    (seam fel)
+  ++  scye                                              ::  prefix doccords only
+    |*  fel=rule
+    ;~(pose ;~(plug apex:docs ;~(pfix gap fel)) ;~(plug (easy *whit) fel))
+  ++  seam                                              ::  postfix doccords too
```
Fang-, line 11725:
```suggestion
  ++  seam                                              ::  doccords
```

Alternatively, "with doccords" here, and "with prefix docs" or similar for `+scye`.

Fang- (commented):
Some thoughts from an initial look. Suggestions are mostly for adding clarity. I do have questions about some of the factoring, and a specification question about summaries, but so far not bad at all!

Haven't looked at the rune parser stuff (everything beyond line 13k) yet, will have to set some more time aside for that.

pkg/arvo/sys/hoon.hoon
```diff
@@ -11534,114 +11557,175 @@
...
     ++  apse
+      %+  knee  *help  |.  ~+
       ;~  pose
-        %+  cook  |=([a=term b=cord] %*(. *whit def (my [a b ~] ~)))
-        (exit fine)
-      ::
-        %+  cook  |=(a=cord %*(. *whit boy `[a ~]))
+        %+  cook  |=(a=cord %*(. *help summary.crib a))
+        ::TODO: if there is a $link then interpret it as the start
+        ::of a prefix comment instead of postfix
```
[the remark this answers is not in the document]
drbeefsupreme: Yeah, but I'm rethinking why and I think I just couldn't figure out the right way to implement it and punted the question til now.

```hoon
|=  ::  .foo: a pat
     ::  .bar: a tar
     [foo=@ bar=*]
[foo bar]
```

which I'd like to attach to the definitions below. The way it is currently written, this could not happen.

So there's two ways to do it - always interpret comments on the same line as postfix comments, so `.foo: a pat` will initially try to attach to `|=`, then it will need to be moved later to `.foo` via some kind of search process. Or say that postfix comments can't have a `$link` and so it must be a prefix comment, and then it will attach to the sample structure via the usual process. The latter sounds easier to understand to me and to implement, but the former is probably how it will look when `(list link)`s are processed anyways and is probably a lot more work to do correctly.

I didn't implement either method yet though, and its at least clear to me that the above comment style really ought to work, and its suprising that it doesn't, so it violates the principle of least surprise.. I'm leaning towards the latter way even though it will have to be ripped out later, and building the former might be grant material.

pkg/arvo/sys/hoon.hoon
```diff
@@ -11534,114 +11557,175 @@
...
+    ++  smol
+      ;~  pfix
+        step
         ;~  plug
-          (into ;~(pfix step line))
-          (rant text)
+          ;~  plug
+            (plus en-link)
+            ;~(sfix ;~(pfix col ace (cook crip (star prn))) (just '\0a'))
+          ==
```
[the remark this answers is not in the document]
drbeefsupreme: I chose to do this style of comment this way because otherwise every comment that starts with `::` then two aces will become a doccord, and I wanted to leave some wiggle room for people to write ordinary comments that don't end up in doccords, since it might not necessarily make sense for them to be attached to whatever they're attaching to. This probably filters out some comments for which this is not the case, but will definitely eliminate a lot of noise that would negatively affect the usability of doccords.

I would estimate based purely on overall impression in the code that comments already written in the batch comment style with a `$link` are useful and accurate, rather than things like notes to the programmer reminding themselves of unfinished work attached to a random atom.

So its sort of a spam prevention measure for usability with the existing codebase, and a way for people who simply don't want to use doccords to write comments like they normally do and avoid it entirely. I don't know whether or not that's the principled way to do this but atm this feels like more of a question of taste than something that really has a right or wrong answer.


For `+larg`, we actually do parse as an empty link, and this signifies to the parser to just attach the `%note` to the following `hoon` or `spec`.


Needing a `col ace` prefix is a mistake I made rewriting the parser, I'll fix that.

pkg/arvo/sys/hoon.hoon
```diff
@@ -11534,114 +11557,175 @@
...
+            ==
+            (just '\0a')
+          ==
+          ;~(pfix null (rant ;~(pfix step teyt)))
+        ==
       ==
     ::
+    ++  myth
+      |*  bod=rule
+      (cook |=([[a=cuff b=cord] c=(list sect)] [a b c]) bod)
```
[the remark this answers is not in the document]
drbeefsupreme: Yeah I feel like I'm doing something wrong elsewhere that makes this even necessary. I know I tried to solve this problem at least a couple times and failed, surely something like this would have been built already if it were actually needed. Your definition is much better in any case. I'll give it another look too try and get rid of this since it seems solvable and I'll probably never be this knowledgable about the Hoon parser again.

pkg/arvo/sys/hoon.hoon
drbeefsupreme: Yeah there's probably some better factoring to do here. I'll give it another look.

drbeefsupreme:
https://github.com/urbit/urbit/pull/5873/commits/df45bfbf84c9ba5a8ff3ad6eceacccbc9a023e32 are TODO items that @joemfb and I came up with while going over `hoon.hoon` today. Not all these need to be addressed before first release, but we should triage them to figure out which ones do, and hopefully answer some of the open questions (mostly around how to appropriately position things to make future contributions easier).

We didn't manage to get to the end - we stopped right as we got to the whitespace parsing among the rune parsers.

pkg/arvo/sys/hoon.hoon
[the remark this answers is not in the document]
drbeefsupreme: https://github.com/urbit/urbit/pull/5873/commits/ca3140a8b3997dd08976f8c09bfd1bdeebd4b14e

pkg/arvo/sys/hoon.hoon
drbeefsupreme: I refactored this a little bit, but didn't remove `+rant`. Even if its only used twice, its creating a `(list sect)` which is one of the central structures in doccords, so I think it deserves its own arm for readability.

https://github.com/urbit/urbit/pull/5873/commits/ca3140a8b3997dd08976f8c09bfd1bdeebd4b14e

pkg/arvo/sys/hoon.hoon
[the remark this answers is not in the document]
drbeefsupreme: See also https://github.com/urbit/urbit/pull/5873/commits/627827f02959b5ddb991799b1e1384467690a0be

drbeefsupreme:
@joemfb and I met today to try to figure out why the `%brbc` branch of `+open` needed to be special-cased to get the annotations to print: https://github.com/urbit/urbit/pull/5873/files#diff-94e19e683e69204abd61d7b5991a01c1db936a982c17b46f8b3d3825497b6d9fR8426-R8432

We didn't solve it, but he suspected that the compiler is not eating the annotation, but it is burying it deeper than the printing library is looking. This might just be a matter of looking deeper inside of nested cores, or it might be somehow related to the lack of a `%make` tag in `%brbc` that the old `+*` had (see #6052). This is more or less the only major issue remaining, and `dprint.hoon` and the changes to `dojo.hoon` are suitable for review at this point.

drbeefsupreme:
Adding a `%-  decorate` into the right spot in `+factory:ax` fixes the `%brbc` issue, but its probably still the wrong thing to do. Helps narrow it down a bit, though. Pretty sure the issue is the hint being applied deeper than expected, as @joemfb had suggested. Hard for me to say whether that's a problem with the compiler or with the printing library? 

https://github.com/urbit/urbit/commit/ef62603e4a32f620bb01cd4a9aaf940aa10c56d2

pkg/arvo/app/dojo.hoon
drbeefsupreme: Since addressing this sort of issue is one of the main goals of the doccords apprenticeship, my opinion now is that its fine to ship this since its an experimental feature seeking a good interface, and its fine to have pain points in the first attempt at an interface.

drbeefsupreme:
The doccords don't show up using the old `+*` syntax either, but some archaeology I did suggests `%made` used to have a bigger role. So I'm not entirely ruling it out.

The fix in https://github.com/urbit/urbit/pull/5873/files#diff-94e19e683e69204abd61d7b5991a01c1db936a982c17b46f8b3d3825497b6d9fR8426-R8432 definitely has some unintended side-effects. If I run

```
> =a %-  ream
  '''
  |$  [item]
  ::    an item
  [thing=cord detail=item]
  '''
  > =b ~(open ap a)
  > =c ?>(?=([%brtr *] b) q.b)
  > ~(open ap c)
```
you end up with
```
> ~(open ap c)
[ %note
  p=[%help p=[cuff=~ crib=[summary='an item' details=~]]]
    q
  [ %brcl
      p
    [ %ktsg
        p
      [ %ktls
        p=[%bust p=%noun]
          q
        [ %note
          p=[%help p=[cuff=~ crib=[summary='an item' details=~]]]
            q
          [   p
            [ %ktts
              p=term=%thing
              q=[%tsgl p=[%$ p=6] q=[%wing p=~[%cord]]]
            ]
              q
            [ %ktts
              p=term=%detail
              q=[%tsgl p=[%$ p=6] q=[%wing p=~[%item]]]
            ]
          ]
        ]
      ]
    ]
      q
    [ %tsls
        p
      [ %note
        p=[%help p=[cuff=~ crib=[summary='an item' details=~]]]
          q
        [   p
          [ %ktts
            p=term=%thing
              q
            [ %cncl
              p=[%tsgr p=[%wing p=~[[%.y p=7]]] q=[%wing p=~[%cord]]]
              q=~[[%$ p=12]]
            ]
          ]
            q
          [ %note
            p=[%help p=[cuff=~ crib=[summary='an item' details=~]]]
              q
            [ %ktts
              p=term=%detail
                q
              [ %cncl
                  p
                [%tsgr p=[%wing p=~[[%.y p=7]]] q=[%wing p=~[%item]]]
                q=~[[%$ p=13]]
              ]
            ]
          ]
        ]
      ]
      q=[%tsls p=[%dtts p=[%$ p=14] q=[%$ p=2]] q=[%$ p=6]]
    ]
```
So we're getting identical `%note`s littered all over the expansion. Not good. I'd prefer the special-casing of `%gist` in `|$` over this, but I think if I stare at this awhile longer a better idea will come out.

joemfb:
The additional `%note`s are because `+decorate:ax` only uses the annotation, it doesn't consume it. You also need to call `+clear:ax` in the right spot (and figure out where that is). It's a difficult convention in that core ...

drbeefsupreme:
> The additional `%note`s are because `+decorate:ax` only uses the annotation, it doesn't consume it. You also need to call `+clear:ax` in the right spot (and figure out where that is). It's a difficult convention in that core ...

I cargo-culted this based off of the fact that `%-  home  %-  decorate` appears several times in the compiler, but I don't see how `+home` is clearing `nut`. I've checked numerous cases at this point (basically just checking what `+factory:ax` and `+open:ap` do with various molds and mold-builders in `hoon.hoon` and confirming there are no duplicated doccords), but since I don't understand why it works I'd appreciate your input @joemfb 

https://github.com/urbit/urbit/commit/42e4839fb1ddc6d801ec160cccb381b60d7d4515

example output:
```
> # list
+list

   item=(* -> *) ~> * -> ?($~ [i=* t=it(*)])

default arm in core:

null-terminated list

mold generator: produces a mold of a null-terminated list of the
homogeneous type {a}.
```
parsing:
```
> =a %-  ream
  '''
  |$  [item]
    ::    null-terminated list
    ::
    ::  mold generator: produces a mold of a null-terminated list of the
    ::  homogeneous type {a}.
    ::
    $@(~ [i=item t=(list item)])
  '''
> ~(open ap a)
[ %brtr
    p
  [ %bccl
      p
    [   i
      [ %bcts
        p=term=%item
        q=[%bcsg p=[%base p=%noun] q=[%bchp p=[%base p=%noun] q=[%base p=%noun]]]
      ]
      t=~
    ]
  ]
    q
  [ %ktcl
      p
    [ %gist
        p
      [ %help
          p
        [ cuff=~
            crib
          [ summary='null-terminated list'
              details
            ~[
              ~[
                [ p=%.y
                    q
                  'mold generator: produces a mold of a null-terminated list of the'
                ]
                [p=%.y q='homogeneous type {a}.']
              ]
            ]
          ]
        ]
      ]
        q
      [ %bcpt
        p=[%base p=%null]
          q
        [ %bccl
            p
          [ i=[%bcts p=term=%i q=[%like p=~[%item] q=~]]
              t
            [   i
              [ %bcts
                p=term=%t
                q=[%make p=[%wing p=~[%list]] q=[i=[%like p=~[%item] q=~] t=~]]
              ]
              t=~
            ]
          ]
        ]
      ]
    ]
  ]
]
```

drbeefsupreme:
I found something that works for the `|$` issue, in the sense that the AST produced by the `+open` hack in https://github.com/urbit/urbit/pull/5873/commits/5fc3b3fd0f781bffe29311adf25601b9ac9352ea#diff-94e19e683e69204abd61d7b5991a01c1db936a982c17b46f8b3d3825497b6d9fR8433-L8429 is reproduced here https://github.com/urbit/urbit/commit/1418e1d9a30c6bb4fc40345f65efcdde1f597292 but making use of `+decorate` and `+clear` instead.

It also does not have the duplicate `%note`s mentioned above. Basically, the `%gist` wrapped by a `%ktcl` turns into a `%note` wrapping the `+open:ap` of `%ktcl`, just like it does on `+$` arms.

`~(open ap c)` from my previous message returns
```
[ %note
  p=[%help p=[cuff=~ crib=[summary='an item' details=~]]]
    q
  [ %brcl
      p
    [ %ktsg
        p
      [ %ktls
        p=[%bust p=%noun]
          q
        [ p=[%ktts p=term=%thing q=[%tsgl p=[%$ p=6] q=[%wing p=~[%cord]]]]
          q=[%ktts p=term=%detail q=[%tsgl p=[%$ p=6] q=[%wing p=~[%item]]]]
        ]
      ]
    ]
      q
    [ %tsls
        p
      [   p
        [ %ktts
          p=term=%thing
            q
          [ %cncl
            p=[%tsgr p=[%wing p=~[[%.y p=7]]] q=[%wing p=~[%cord]]]
            q=~[[%$ p=12]]
          ]
        ]
          q
        [ %ktts
          p=term=%detail
            q
          [ %cncl
            p=[%tsgr p=[%wing p=~[[%.y p=7]]] q=[%wing p=~[%item]]]
            q=~[[%$ p=13]]
          ]
        ]
      ]
      q=[%tsls p=[%dtts p=[%$ p=14] q=[%$ p=2]] q=[%$ p=6]]
    ]
  ]
]
```

My main concern is with the effect `+clear` has on `def`. `+clear` nulls out `nut`, `bug`, and `def`. `+decorate` handles `nut` and `bug`, but not `def`. The only places `def` is used is in `spore` and the `%bchp` case in `+analyze`, and here I am changing the only callsite to `+spore` with `+spore:clear`. But since `+spore` is what handles `$~`, which is ubiquitous, I'd think that if there was a real issue here then it wouldn't get past compiling `hoon.hoon`. `spore:clear` should be running `+spore` with `def` made null.


@joemfb

drbeefsupreme:
I'm currently looking around for `hoon.hoon` PRs that could be bundled together with this one. I started already but its getting late so I'll finish tomorrow.

Based on @joemfb 's comment here https://github.com/urbit/urbit/pull/5939#pullrequestreview-1070285682 I'm guessing that since doccords breaks some jet signatures, it also needs to be a Kelvin release? In which case #6064 should be included as well.

drbeefsupreme:
Friendly poke to @Fang- about reviewing this. Since @joemfb is out for at least a couple weeks I think its all on you unless you call in backup.

I am feeling pretty confident with the `hoon.hoon` changes but nobody has yet to really look at `/lib/dprint.hoon` or the `dojo.hoon` changes closely. Those have the highest probability of needing more than minor changes done IMO (though to be clear I believe they're shippable as is, if we had to).

philipcmonk:
I believe this is ready to go.  I made the following changes:
- Adapted the wip from jon/doccords-vere to a working version here: urbit/vere#114
- Virtualize calls to dprint in dojo so that any crashes are printed stead of clearing your dojo history
- Bump all kelvins
- Includes #6039, which was removed from 417 at the last moment because it needed a hoon kelvin bump
- Put doccords behind a parser flag, defaulting to on.  Per this conversation:

Screenshot from 2023-01-13 17-46-38

PR #5874, opened 2022-07-02
dojo: re-add support for ford runes

sidnym-ladrut:
These changes re-add support for file imports via `%ford` runes (e.g. `/+`, `/-`) to `%dojo`, since their inadvertent absence [post-fusion](https://github.com/urbit/urbit/pull/3060). To minimize changes to `%dojo`'s app data, imports are currently limited to the `/+` and `/-` runes, which import into the `lib` and `sur` variables, respectively.

## Demo ##

```
~zod:dojo> /-  *dns
~zod:dojo> (address.sur [%if *@if])
[%if if=.0.0.0.0]
~zod:dojo> /+  kg=keygen, csv
~zod:dojo> (ownership-wallet-from-ticket.kg.lib ~zod [4 ~.dozzod-dozzod-dozzod-dozzod] ~)
[ type="ownership" seed=...]
~zod:dojo> (print-rows.csv.lib `(list vase)`~[!>('a') !>('b')])
'a'
'b'
<1.tan [* <46.hgz 1.pnw %140>]>
~zod:dojo> /+
~zod:dojo> lib
-find.lib
dojo: hoon expression failed
~zod:dojo> /%  txt  %txt
~zod:dojo> =>(~(grab txt "") (noun))
<||>
~zod:dojo> /=  md  /mar/md
~zod:dojo> =>(~(grow md "") mime)
[[%text %plain ~] p=0 q=0]
```

## Tasks ##

- [X] Add support for importing via `/+` into `lib`.
- [X] Add support for importing via `/-` into `sur`.
- [x] Add support for cable expressions (e.g. `/+  kg=keygen, *csv`).
- [X] Add support for arbitrary Hoon importing via `/=`.
- [X] Add support for mark importing via `/%`.
- [ ] Add support for importing via other Ford runes (?).

Tagging: @Fang-, @joemfb

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -202,12 +202,8 @@
     %+  turn  cables
     |=  cable=cable:clay
     ^-  dojo-source
-    ::  TODO: The following *does* properly add faces to imported cores,
-    ::  but discards all their type information, making them hard to
-    ::  work with
-    ::
-    ::  =+  add-face=?~(face.cable "|=(n=* n)" ;:(weld "|=(n=* ^=(" (trip u.face.cable) " n))"))
-    ::  :^  0  %do  (scan add-face parse-hoon)
+    =+  add-face=?~(face.cable "|*(n=* n)" ;:(weld "|*(n=* ^=(" (trip u.face.cable) " n))"))
+    :^  0  %do  (scan add-face parse-hoon)
     :+  0  %dv  [-.dir `path`[base-path file-path.cable ~]]
   ::
   ++  parse-cable
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -174,8 +174,10 @@
     ::
       ;~  pfix  fas
         ;~  pose
-          (parse-variable (cold %sur hep) ;~(pfix gap (parse-ford %sur)))
-          (parse-variable (cold %lib lus) ;~(pfix gap (parse-ford %lib)))
+          (parse-variable (cold %sur hep) ;~(pfix gap (parse-cables %sur)))
+          (parse-variable (cold %lib lus) ;~(pfix gap (parse-cables %lib)))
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/dojo.hoon
```diff
@@ -193,23 +193,22 @@
...
+    %-  cook  :_  (most ;~(plug com gaw) parse-cable)
+    |=  cables=(list cable:clay)
+    :+  0  %tu
+    ::
+    %+  turn  cables
+    |=  cable=cable:clay
+    ^-  dojo-source
+    ::  TODO: The following *does* properly add faces to imported cores,
+    ::  but discards all their type information, making them hard to
+    ::  work with
```
sidnym-ladrut, line 207: With this version, all imports act like direct/bunt imports (e.g. `/-  *dns`). I was able to use a `%do` command to enable named imports, but my method stripped all type information from the imported files. Any advice/insights on how to fix this would be greatly appreciated!

sidnym-ladrut: I was able to fix this by using a wet gate, but suggestions on how to clean up this part of the code would still be welcome!

belisarius222: I think you’d be better off parsing into just a cable data structure and including the cable itself in Dojo’s state for this session.  You might need to add a new type of dojo-source (or maybe dojo-sink, not sure).

Changing Dojo’s import state should trigger it to send a %warp to Clay to run whatever build is specified by that cable.  The build result will come back as a vase, and you can then use the face in the stored cable data structure to modify the type in the vase to include it in the “dojo subject” — i.e. the subject Dojo uses to run user-specified commands.

A more sophisticated import system would also subscribe to Clay for future versions of this import so the imported file would automatically get rebuilt and loaded into the dojo subject whenever Clay commits to that desk.  I recommend building this as a one-off import command first, though, and only attempting live imports as a second step.  The asynchronicity of live imports makes them more difficult to get right.

PR #5877, opened 2022-07-04
eyre, http-api: add support for noun-based channels

Fang-:
Consider this my Resolution for JSON Independence.

For a long time now, Eyre's channels have communicated over JSON. This has had some unfortunate side-effects. Primarily, to support access over HTTP, userspace developers had to implement Arvo-side JSON conversions for all their relevant marks. This then obscured the relationship between agents and their frontends, and muddied the developer experience there in general.

The contents of this PR have been tested and proven to work, but there are unresolved dev UX and package management questions, primarily on the javascript side. Not to mention, this is based on my new-webterm branch, because webterm felt like the most practical test case for me. Hence: draft PR.

---

There are three parts of work inside:
- 20573ad updates eyre to support multiple channel "modes": `%json` and `%jam`. The latter sends and received `@uw`-encoded jammed nouns in all places where we currently use json.
- 01b00f6 changes the http-api package to use `%jam`-mode channels instead of `%json` ones. This has implications for its outward-facing interface, most notably the handling of subscription updates.
- c0b530c updates nu-webterm to be compatible with the changed http-api, as a proof of concept and indicator of what frontend codebases will need to change/add to use noun-based channels.

As mentioned, there's open questions and concerns for all of these. Let's walk through the current approach and see what is yet unresolved.

### eyre

Eyre now supports multiple channel "modes". The SSE framing is identical between the different modes, but the format of both the SSE events and client-sent requests changes depending on the mode. `%json` mode behaves as expected, unchanged. In `%jam` mode, the format is `@uw`-encoded jammed nouns. For example, the noun `[%poke-ack ~]` becomes the bytestring `0w2RIr.2mIHm.TK7U1`.

For accessing the `%json` mode, the existing channel endpoint (`/~/channel/[etc]`) can be used. To use `%jam` mode instead, access the channel under `/~/channel-jam/[etc]` instead. When a channel is opened, the "selected" mode is associated with it permanently.

Remaining questions/work:
- [ ] That endpoint path is aesthetically bankrupt. We could use the file extension to indicate the mode, but this doesn't _quite_ vibe with the usual URI semantics.
- [ ] That endpoint isn't even implemented as a separate endpoint, instead it resolves back to `/~/channel` during binding lookup, and determines the mode from the original URL later on. This should change in accordance with the answer to the path question above.
- [ ] Connecting to the same channel across multiple modes is currently allowed. But it's untested, and unclear whether this is desirable or not. PUT requests are only supported in the channel's "original" mode. Whatever we end up doing, these two should be consistent. Leaning towards disallowing multi-mode-ing.

### http-api

This PR contains a version of http-api that has been switched over to use noun-based channels exclusively. To handle nouns, we have revived @frodwith's nockjs project. (It's hard-copied in right now, I'll make a package & repo for it under @urbit soon.) To parse and render the `@uw`s, we have extended aura-js.

The http-api interface changes slightly.  
We now give nouns instead of javascript objects for subscription `%fact`s, as well as tang nouns instead of error strings in the poke-nack and watch-nack cases. `quit()` on longer gets called with any arguments, there was no data there anyway (right?).  
Subscriptions now take path nouns* instead of strings, and pokes must be specified as nouns* instead of javascript objects.

\* Notably, to avoid having to expose/use noun construction logic to clients of http-api, we also accept primitive javascript data in place of proper nockjs `Noun`s. We wrap these arguments in a `nockjs.dwim()` call, converting them to nouns in the way you would expect.  
We may still want to expose noun construction helpers, but aside from `@p` string->number conversion I haven't felt too serious a need.

To help clients deal with the nouns they receive over subscriptions, nockjs has been extended with [`enjs` helpers](https://github.com/urbit/urbit/blob/01b00f6c615e3ce563263e07e112e0841692b691/pkg/npm/http-api/src/nockjs/noun.js#L354-L458) that bear some passing resemblance to the ones from zuse we all know ~~and love~~. Some of them parse primitives, others convert lists or sets into javascript arrays, convert tuples into objects, or try any number of parsers in sequence. If they fail to match, they crash.

Remaining questions/work:

- [ ] Main question: do we want to fully deprecate JSON-based channels at some point? I certainly do! But I'm gonna need some buy-in (and nock/noun libraries in more languages).
  - Update the major version, and keep supporting the old one for a little while.
  - Release a separate http-api-nouns package, maintain the two side by side.
  - Support both modes within the same package.
- [ ] In addition to having its `Noun`s api touched up a bit, nockjs should clearly in its own package. But thinking one step further: if it's going to ship with noun<->object conversions, shouldn't it also ship with aura de/serialization? Certainly if you use http-api you're going to want to deal with nockjs.  
      It feels like there's strong relations between these three libraries. How should this be dealt with? Do they get merged, or just re-expose each other's utilities, or expect the developer to figure it out? Notable danger: a `Noun` constructed in http-api doesn't `instanceof` correctly with the `Noun` a client application independently imports.  
      (http-api already re-exposes nockjs's `enjs` stuff, but that's mostly incidental, I wanted to get up-and-running quickly.)
- [ ] Additionally, nockjs uses a different bignum library than aura-js: [`jsbn`](https://www.npmjs.com/package/jsbn) and [`big-integer`](https://www.npmjs.com/package/big-integer) respectively. Apparently `BigInt` exists natively now, and the latter uses that when available. Is native better? Should nockjs be modified to use that? Regardless, these two (and all our other libraries) should probably be consistent, and/or not strictly require their own bignum objects as input.
- [ ] None of the nouns stuff is typed right now. Unclear to me if the noun->object conversions can be made to be typed, especially where we're constructing objects based on the supplied string arguments.
- [ ] Should clients be able to bind per-mark conversion functions _when they subscribe_, so that they don't have to manually check the mark & convert as appropriate?
- [ ] We still use `.json` for scrying. Should probably change that over to `.jam` while we're at it.
- [ ] We may want to pre-render tangs in all the nack cases, or at least provide a utility for rendering tanks.

### webterm

The work here, as expected & desired, isn't actually all that interesting. It's just the diff of moving the conversion logic from the marks into the frontend.

---

cc @arthyn @liam-fitzgerald I'm gonna need some guidance on the http-api questions here. Probably worth meeting up to discuss this soon.

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1292,7 +1292,8 @@
   ::    events since then.
   ::
   +$  channel
-    $:  ::  channel-state: expiration time or the duct currently listening
+    $:  mode=?(%json %jam)
+        ::  channel-state: expiration time or the duct currently listening
         ::
         ::    For each channel, there is at most one open EventSource
         ::    connection. A 400 is issues on duplicate attempts to connect to the
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -67,10 +67,8 @@
 ::  more structures
 ::
 |%
-++  axle
-  $:  ::  date: date at which http-server's state was updated to this data structure
-      ::
-      date=%~2020.10.18
++$  axle
+  $:  %1
       ::  server-state: state of inbound requests
       ::
       =server-state
@@ -119,9 +117,12 @@
   $%  ::  %ack: acknowledges that the client has received events up to :id
       ::
       [%ack event-id=@ud]
-      ::  %poke: pokes an application, translating :json to :mark.
+      ::  %poke: pokes an application, validating :noun against :mark
       ::
-      [%poke request-id=@ud ship=@p app=term mark=@tas =json]
+      [%poke request-id=@ud ship=@p app=term mark=@tas =noun]
+      ::  %poke-json: pokes an application, translating :json to :mark
+      ::
+      [%poke-json request-id=@ud ship=@p app=term mark=@tas =json]
       ::  %watch: subscribes to an application path
       ::
       [%subscribe request-id=@ud ship=@p app=term =path]
@@ -200,11 +201,32 @@
   (sub u.sus ack)
 ::  +parse-channel-request: parses a list of channel-requests
 ::
+++  parse-channel-request
+  |=  [mode=?(%json %jam) body=octs]
+  ^-  (each (list channel-request) @t)
+  ?-  mode
+      %json
+    ?~  maybe-json=(de-json:html q.body)
+      |+'put body not json'
+    ?~  maybe-requests=(parse-channel-request-json u.maybe-json)
+      |+'invalid channel json'
+    &+u.maybe-requests
+  ::
+      %jam
+    ?~  maybe-noun=(bind (slaw %uw q.body) cue)
+      |+'invalid request format'
+    ?~  maybe-reqs=((soft (list channel-request)) u.maybe-noun)
+      ~&  [%miss u.maybe-noun]
+      |+'invalid request data'
+    &+u.maybe-reqs
+  ==
+::  +parse-channel-request-json: parses a json list of channel-requests
+::
 ::    Parses a json array into a list of +channel-request. If any of the items
 ::    in the list fail to parse, the entire thing fails so we can 400 properly
 ::    to the client.
 ::
-++  parse-channel-request
+++  parse-channel-request-json
   |=  request-list=json
   ^-  (unit (list channel-request))
   ::  parse top
@@ -220,7 +242,9 @@
   ?:  =('ack' u.maybe-key)
     ((pe %ack (ot event-id+ni ~)) item)
   ?:  =('poke' u.maybe-key)
-    ((pe %poke (ot id+ni ship+(su fed:ag) app+so mark+(su sym) json+some ~)) item)
+    %.  item
+    %+  pe  %poke-json
+    (ot id+ni ship+(su fed:ag) app+so mark+(su sym) json+some ~)
   ?:  =('subscribe' u.maybe-key)
     %.  item
     %+  pe  %subscribe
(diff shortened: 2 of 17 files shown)


matildepark:
Sure, seems cool

zalberico:
@Fang- does this require your webterm work as a prerequisite? 

Reading through this and trying to understand it, it sounds like the overall goal is to be able to move away from JSON  for eyre channels by supporting using urbit native nouns/jam files alongside JSON (with an eye toward eventually dropping JSON support down the line)?

The argument being that this would make it easier to develop userspace applications because devs wouldn't have to "implement Arvo-side JSON conversions for all their relevant marks"?

Keeping this one open since the directional goal of having things be urbit native where it makes sense seems like a correct one to me, but realistically I suspect it'll be a while before we get to this.

Fang-:
> the overall goal is to be able to move away from JSON for eyre channels by supporting using urbit native nouns/jam files alongside JSON (with an eye toward eventually dropping JSON support down the line)?

That is correct. Dropping JSON fully will be much, much further down the line though.

> this would make it easier to develop userspace applications because devs wouldn't have to "implement Arvo-side JSON conversions for all their relevant marks"?

Both this, and a more cohesive developer experience in general: you will be able to work with the exact same data structures on both the back- and frontend, removing confusion around conversions & formats entirely. This also pushes clients to be "more urbit-y": speaking nouns is guaranteed to be useful forever.

> realistically I suspect it'll be a while before we get to this.

You can't stop me! (;

polwex:
What's missing to get this merged? Happy to help in any capacity.

Fang-:
A couple tricky things, but not _that_ much I don't think? I have some written-but-not-yet-tested improvements to the Eyre side of this. I also tried updating the nockjs bignum library to match aura-js's some time ago, but had trouble with that. Don't think that's a hard requirement, but it might catch library users off-guard.

Other than that, mostly cleanup on the js side? Typescript typing would be great if we can get it. The changes here will need to be ported over to the appropriate repo, we'll need to figure out a release strategy for the noun-based http-api, and yes some of the stuff under "remaining questions/work" in the opening post might be nice to get but also shouldn't be hard-blocking, we can always do updated releases later.

polwex:
How can I help? If you get the repo transition done I'm happy to contribute there.

Fang-:
Depends on what you're good at & want to do! Some suggestions:

If you want a fun challenge, moving nockjs over to use `big-integer` (in place of `jsbn`) might be a good project. Having its bignums be consistent with our other libraries would help a lot with integration.  
Alternatively, it would be cool to try and see if the `enjs` stuff from `http-api/src/nockjs/noun.js` could be made to be Typescript-typed. Maybe it's too dynamic? But I don't actually know. Having it deduce output types properly would be very helpful for all our other Typescript code.

Should be fine to work on top of this branch for now, regardless of repo transitions. We can always cherry-pick/migrate the work later if need be.

jalehman:
Closing in favor of #6396

PR #5880, opened 2022-07-05
vere: build: add ability to create compdb (for Clangd etc.)

midlyx-hatrys:
Running "make compile_commands.json" after configuring does the
needful.

### DIFF
--- a/pkg/urbit/.gitignore
+++ b/pkg/urbit/.gitignore
@@ -3,3 +3,8 @@
 /include/config.h
 /include/ca-bundle.h
 /include/ivory.h
+
+# indexing
+/.cache/
+compile_commands.json
+tags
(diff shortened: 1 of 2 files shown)


midlyx-hatrys:
@joemfb

joemfb (approved):
LGTM

I hadn't seen compdb before this. Are you using this with rtags?

midlyx-hatrys:
> I hadn't seen compdb before this. Are you using this with rtags?

With Clangd.

A quick TL;DR on the C-family indexer landscape:
- The new sort-of-language-agnostic "IDE as a service" hotness is the Language Server Protocol, a.k.a. LSP.  There are LSP clients for all editors and IDEs that are in any way programmable.  Likewise there are LSP servers for most languages, even Hoon.
- All proper C-family indexers are based on an actual compiler front-end (Clang) and take a "compdb" (`compile_commands.json`) as input so that the indexer knows all the actual build flags and stuff and can do its job without guesswork (i.e. ctags it ain't). 
- Rtags is quite nice, but is kept on life support, is a bother to set up, and does not speak LSP.
- The only two maintained C-family LSP servers I'm aware of are Clangd and CCLS.  Basically everybody should just to use Clangd, as it is an official Clang/LLVM tool and is developed kinda more actively than Clang itself these days.

PR #5882, opened 2022-07-07
u3: road stack guard page optimization

joemfb:
The PR builds on the road stack overflow guard page introduced in #5881.

- b630490573 prevents us from faulting twice in a row on the guard page
- 350406c9e1 makes the allocator fallback to cell-at-a-time allocation when there's not enough space for 4k cells
- f0fa528f3b removes/disables overflow checks (now handled by the guard page)
- 9d5b24362e makes `mprotect()` failure always fatal


mcevoypeter (commented):
LGTM. My only two points of feedback were to address the todos you'd left--adding a macro to control the guard page and returning 0 when the guard page cannot be protected--but I just pushed commits that make those changes, which seemed easier than asking you to do it.

joemfb:
@mcevoypeter thanks, this is much better than TODO comments.

PR #5883, opened 2022-07-07
vere: exit correctly on fatal error

joemfb:
This PR tweaks handling of fatal errors in vere (specifically the "king" process). We now track the presence of a fatal error in global state (`u3_Host.xit_i`, already present and fit for purpose), and make sure to fail any attempted prep-for/upgrade in the presence of errors. (The absence of this logic meant that a binary upgrade could succeed even if preparatory replay failed, which could leave the pilot with a binary that refuses to run linked into the pier.)

### DIFF
--- a/pkg/urbit/vere/disk.c
+++ b/pkg/urbit/vere/disk.c
@@ -689,18 +689,14 @@ u3_disk_exit(u3_disk* log_u)
     }
   }
 
-  //  cancel write thread
+  //  try to cancel write thread
+  //  shortcircuit cleanup if we cannot
   //
-  //    XX can deadlock when called from signal handler
-  //    XX revise SIGTSTP handling
-  //
-  if ( c3y == log_u->ted_o ) {
-    c3_i sas_i;
-
-    do {
-      sas_i = uv_cancel(&log_u->req_u);
-    }
-    while ( UV_EBUSY == sas_i );
+  if (  (c3y == log_u->ted_o)
+     && uv_cancel(&log_u->req_u) )
+  {
+    // u3l_log("disk: unable to cleanup\r\n");
+    return;
   }
 
   //  close database
--- a/pkg/urbit/vere/disk.c
+++ b/pkg/urbit/vere/disk.c
@@ -693,7 +693,7 @@ u3_disk_exit(u3_disk* log_u)
   //  shortcircuit cleanup if we cannot
   //
   if (  (c3y == log_u->ted_o)
-     && uv_cancel(&log_u->req_u) )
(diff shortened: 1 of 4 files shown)


pkg/urbit/vere/disk.c
```diff
@@ -689,18 +689,14 @@ u3_disk_exit(u3_disk* log_u)
...
-
-    do {
-      sas_i = uv_cancel(&log_u->req_u);
-    }
-    while ( UV_EBUSY == sas_i );
+  if (  (c3y == log_u->ted_o)
+     && uv_cancel(&log_u->req_u) )
+  {
+    // u3l_log("disk: unable to cleanup\r\n");
+    return;
```
joemfb, line 699: The cancellation spin loop here was invalid -- libuv only lets you cancel worker threads while they're still queued.

pkg/urbit/vere/io/term.c
```diff
@@ -1036,13 +1036,11 @@ u3_term_ef_ctlc(void)
 {
   u3_utty* uty_u = _term_main();
 
-  {
+  if ( uty_u->car_u ) {
     u3_noun wir = u3nt(c3__term, '1', u3_nul);
     u3_noun cad = u3nt(c3__belt, c3__ctl, 'c');
 
     c3_assert( 1 == uty_u->tid_l );
-    c3_assert( uty_u->car_u );
```
joemfb, line 1044: This unnecessary assertion was an unrelated issue that came up while testing.

mcevoypeter (changes requested):
Slowly but surely making our way towards better error handling.

pkg/urbit/vere/disk.c
```diff
@@ -689,18 +689,14 @@ u3_disk_exit(u3_disk* log_u)
...
-  //
-  if ( c3y == log_u->ted_o ) {
-    c3_i sas_i;
-
-    do {
-      sas_i = uv_cancel(&log_u->req_u);
-    }
-    while ( UV_EBUSY == sas_i );
+  if (  (c3y == log_u->ted_o)
+     && uv_cancel(&log_u->req_u) )
```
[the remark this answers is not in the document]
joemfb: Not without dramatic changes elsewhere. This function is called as part of synchronous exit/cleanup, and can be called during error handling. We made all our exit/cleanup functions fully async, then we could always fully cleanup. But this code is changing substantially very soon (mars/urth, log truncation), so it's not worth a major restructuring here.

mcevoypeter (commented):
LGTM. Thanks for addressing the changes I suggested.

PR #5884, opened 2022-07-07
hoon: make ?# wuthax work correctly

Fang-:
Various fixes and improvements to `+ar`, which were needed to make `?#` work correctly.

`?#` now behaves much like `?=`, with the difference that it eats any pre-existing faces it encounters during pattern matching.

Examples follow, comparing behavior between `?=` and `?#`.

```hoon
> =/  a  [x=1 y=2]  ?>  ?=([@ @] a)  a
[x=1 y=2]
> =/  a  [x=1 y=2]  ?>  ?#([@ @] a)  a
[1 2]
> =/  a  [x=1 y=2]  ?>  ?=([@ *] a)  a
[x=1 y=2]
> =/  a  [x=1 y=2]  ?>  ?#([@ *] a)  a
[1 y=2]
::
> =/  a  [x=1 y=2]  ?>  ?=([p=@ *] a)  a
[x=p=1 y=2]
> =/  a  [x=1 y=2]  ?>  ?#([p=@ *] a)  a
[p=1 y=2]
> =/  a  [x=1 y=2]  ?>  ?=([* q=@] a)  a
[x=1 y=q=2]
> =/  a  [x=1 y=2]  ?>  ?#([* q=@] a)  a
[x=1 q=2]
::
> =/  l  (gulf 1 5)  ?>  ?=([@ two=@ three=@ *] l)  l
[i=1 t=[i=two=2 t=[i=three=3 t=~[4 5]]]]
> =/  l  (gulf 1 5)  ?>  ?#([@ two=@ three=@ *] l)  l
[1 two=2 three=3 t=~[4 5]]
```

Should be good to go. `?#` never really worked and so nobody ever really used it. The `#` will remain a signal that this rune is temporary, but it should prove useful already. We may want to consider expanding `?=` with syntax to toggle into (or out of) this behavior.

Things that may still need some attention, but haven't presently been observed to cause problems:
- In `+gain:ar`, the `[%base %noun]` and `%core` - `%cell` cases seem weird, as well as the latter in `+lose:ar`.
- In `+fish:ar`'s `%over` case we may need to guard against the wing not existing.
- We support wings in the pattern, ie `?=(byts [1 2])` and `?#(byts [1 2])`, but it's rarely used, and we might not want to support it, considering those wings may hide complex types we can't match against. Also in the `?#` case it does not actually remove pre-existing faces correctly.

(And no, changing `?=` to secretly be `?#` unfortunately does not trivially just work everywhere.)


joemfb:
```
> =u (~(play ut -:!>(..unit)) (ream '`[(unit) (unit)]`[`1 `2]'))
> u
#t/[u(*) u(*)]
> (ream '?=([^ ^] a)')
[%wtts p=[%bccl p=[i=[%base p=%cell] t=~[[%base p=%cell]]]] q=~[%a]]
> =sp [%bccl [%base %cell] [%base %cell] ~]
> =t (~(play ut %noun) ~(example ax sp))
> t
#t/[[* *] * *]
>
> (~(fuse ut u) t)
#t/[[%~ u=*] %~ u=*]
> (~(crop ut u) t)
#t/[u(*) u(*)]
>
> (ream '?#([^ ^] a)')
[%wthx p=[%cell skin=[%base base=%cell] skin=[%base base=%cell]] q=~[%a]]
> =sk [%cell [%base %cell] [%base %cell]]
> ~(gain ar:ut u sk)
#t/[[%~ u=*] %~ u=*]
> ~(lose ar:ut u sk)
#t/[%~ %~]
```

joemfb:
Booted up this branch and did it for real:

```

> =<  -  !>  =>  `[(unit) (unit)]`[~ ~]  .
#t/[u(*) u(*)]
>
> =<  -  !>  =>  `[(unit) (unit)]`[`1 `2]  ?>(?=([^ ^] .) .)
#t/[[%~ u=*] %~ u=*]
> =<  -  !>  =>  `[(unit) (unit)]`[~ ~]  ?<(?=([^ ^] .) .)
#t/[u(*) u(*)]
>
> =<  -  !>  =>  `[(unit) (unit)]`[`1 `2]  ?>(?#([^ ^] .) .)
#t/[[%~ u=*] %~ u=*]
> =<  -  !>  =>  `[(unit) (unit)]`[~ ~]  ?<(?#([^ ^] .) .)
#t/[%~ %~]
```

joemfb:
My celebration was premature, see #5983

zalberico:
This looks like another good candidate that might be ready to merge?

joemfb:
This is not ready, see my linked draft PR.

joemfb (approved):
this will do

For posterity, I have reservations about the design of `?#` as a replacement for `?=`. In particular, being able to express forks in pattern-matching is very useful and frequently used, but not supported by `$skin` (used for patters by `?#`). We could add forks to it, but then we'd have three separate fork-schemes: `$type`, $spec`, and `$skin`. `$skin` is otherwise used for adding faces, casting, and destructuring (see `=/` and `^=`), and I'm just not convinced that's the same domain as hoon-style pattern matching.

That being said, this PR just makes `?#` work as designed, where it was previously broken. So it should go out to enable further experimentation.

PR #5878, opened 2022-07-05
ames: content distribution (remote scry)

Fang-:
New release candidate PR/branch for the content distribution line of work.

We intend to release this alongside the behn timer fix (#5857) and all other work already merged into next/arvo.

PR #5886, opened 2022-07-08
gall, ames: fix subscription desync (stage ii)

yosoyubik:
Stage II of the deployment of the gall-rq-fix—Stage I: https://github.com/urbit/urbit/pull/5644


yosoyubik:
This is stage II of the deployment of the gall-rq-fix—see https://github.com/urbit/urbit/pull/5644 for state I. Here we send `%cork` pleas when leaving a subscription, and using new flows for future subscription by incrementing `sub-nonce.yoke`. If we detect that a publisher doesn't understand `%corks` (we will get a `naxplanation` for the `%cork`) we add the wire for this bone to a global set that gets processed once a day and resubmits the cork to the publisher—by having one global timer we avoid creating one timer for each `%cork`).

PS: tested locally on fakezods using toy pub/sub agents and Group join/leaving, next step is to put this on live moons and monitor for any issues. Stages I and II state upgrades tested with a pill from `master` but will need to be updated to handle the new state migration after https://github.com/urbit/urbit/pull/5878 and [agent-clay](https://github.com/urbit/urbit/tree/philip/agent-clay) are deployed,

belisarius222:
That branch contains an in-progress overhaul of the clay/gall/kiln upgrade system. Clay will track which agents to run and %pass moves to Gall to install them. Kiln will basically just manage downloads and Kelvin blocking logic.

It should be simpler and more reliable than the current bug-prone upgrade system. It’s still very new, though, and not done yet.

—
~rovnys-ricfer
https://urbit.org

yosoyubik:
I think that if we deployed this in two stages, the best thing is to cherry-pick these commits
- [gall: don't send %cork when leaving a subscription](https://github.com/urbit/urbit/pull/5644/commits/f3a60aef14b697f8118b5fe7038f5ea61a653676)
- [gall: skip adding nonce to new subscriptions](https://github.com/urbit/urbit/pull/5644/commits/fc506805eda27f75411c1170ef81c2e71e60a4cd)
- [gall: use pre-nonce on take](https://github.com/urbit/urbit/pull/5644/commits/955ed073f851896de8d98d46d8a7ab7e6ccca7fa)

from https://github.com/urbit/urbit/pull/5644, and then revert them when releasing stage ii—i think the history is a bit messed up, so this way we avoid dealing with merge conflicts... etc

Fang- (approved):
In the name of @belisarius222, I approve this PR.

PR #5887, opened 2022-07-08
hoon: typed paths

Fang-:
Adds an `$iota` type, which is an atom tagged with its aura, where we only support auras that provide distinct & unique rendering/parsing formats.

Using that, we introduce `$pith`, the "typed path" type. It's a list of `$iota`, to be used in place of `$path`, that serializes in the same way that `$path` does. Parsing infers the elements' auras from their notation, and annotates into `$iota`s accordingly.

Because `%tas` is the common case, we support that as plain atom, instead of `[%tas %my-term]`.

The idea is that using `$pith`-style typed paths throughout the system will let us save on de/serialization costs on both performance and developer ux fronts. We already use paths extensively throughout the system (wires, subscription paths), and it seems likely this will increase in the next iteration of gall's agent model. (No, you didn't miss a memo, we haven't settled on anything yet.)

You'll note that "using `$pith`-style typed paths throughout the system" is not actually happening yet in this PR. We'll work our way up to that. The conversion functions included here (`+pout` and `+pave`) can get us started on a local level.

To make working with `$pith`s more ergonomic, we add special syntax for generating them, as well as generating patterns for matching against them.  
`#/some/~zod/'path'` will parse to `[%some [%p ~zod] [%t 'path'] ~]`.  
`?=(#/some/who=@p/@t my-pith)` will match on that.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -11756,7 +11756,7 @@
       (stag %clsg (more fas stem))
     ==
   ::
-  ++  stem  !:
+  ++  stem
     %+  knee  *hoon  |.  ~+
     %+  cook
       |=  iota=$%([%hoon =hoon] iota)
@@ -11773,7 +11773,7 @@
           :-  '$'      (cold %$ buc)
           :-  '0'^'9'  (slot bisk:so)
           :-  '-'      (slot tash:so)
-          :-  '.'      zust:so
+          :-  '.'      ;~(pfix dot zust:so)
           :-  '~'      (slot ;~(pfix sig ;~(pose crub:so (easy [%n ~]))))
           :-  '\''     (stag %t qut)
           :-  '['      (slip (ifix [sel ser] wide))
@@ -13141,7 +13141,12 @@
                       ^-  spec
                       :+  %bccl
                         [%leaf %tas aura]
-                      [%bcts term [%base %atom aura]]~
+                      :_  ~
+                      :+  %bcts  term
+                      ?+  aura  [%base %atom aura]
+                        %f  [%base %flag]
+                        %n  [%base %null]
+                      ==
                     ;~(plug sym ;~(pfix tis pat mota))
                   ::
                     ::  /constant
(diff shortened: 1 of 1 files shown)


belisarius222:
I'm realizing, upon seeing that example, that there's a tension between two different goals of the path syntax:
1) URL-safe
2) bijective with Hoon atom syntaxes

Specifically, the text encodings might not be URL-safe. I don't think you can have quotes in URLs, and capital letters aren't respected in URLs either iirc.  The `@ta` encoding style, like `~.foo`, could potentially be used for `@ta` path elements, but I'm not sure what a URL-safe encoding of arbitrary `@t` would look like -- and according to [the original URL RFC](https://datatracker.ietf.org/doc/html/rfc1738), 
```
Other characters are unsafe because
   gateways and other transport agents are known to sometimes modify
   such characters. These characters are "{", "}", "|", "\", "^", "~",
   "[", "]", and "`".
   ```

The `@ta` encoding does use `~`, and in practice, a bunch of our HTTP endpoints include `~`, so maybe that RFC is out of date, and `~` is accepted now.  I'm not deeply familiar with the URL validity predicate.

Another option could be to disallow arbitrary `@t` in paths, but it does seem to me that there should be some URL-safe bijective encoding we could use.  It might not exist at the moment, though, in which case it would likely be addressed better as a later line of work.

**Comment by joemfb** [comment]
I'm making an assumption that should be stated explicitly: `$pith` will be used as an internal type, for `wire`, `%clay` paths, in namespace handlers, &c. From outside, paths will be strings or `path` (constructed from URLs, file system paths, network packets, &c.). This perspective is that the serialization is the interface, not the aura taxonomy or whatever.

File: pkg/arvo/sys/hoon.hoon [hunk 1]
```diff
@@ -1870,6 +1870,42 @@
...
++$  knot  @ta                                           ::  ASCII text
++$  noun  *                                             ::  any noun
++$  path  (list knot)                                   ::  like unix path
++$  pith  (list iota)                                   ::  typed urbit path
++$  stud                                                ::  standard name
+          $@  mark=@tas                                 ::  auth=urbit
+          $:  auth=@tas                                 ::  standards authority
+              type=path                                 ::  standard label
+          ==                                            ::
++$  tang  (list tank)                                   ::  bottom-first error
```
joemfb, line 1892: Doesn't this just duplicate the type declarations of `2q`, causing them to be shadowed further down with redundant declarations? @belisarius222

belisarius222: Fixed on the next/kelvin/413 branch

PR #5888, opened 2022-07-08
eyre: fix for %trim bug

bacwyls:
Problem:
by-channel has its own copy of server-state from line 2182. discard-channel returns an altered state, with one channel removed from the state of by-channel.
but the state of by-channel isn't changing with each iteration, so %trim is only removing one channel per invocation instead of all channels.

Solution:
update by-channel on each iteration.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -2178,7 +2178,7 @@
   ::    XX cancel active too if =(0 trim-priority) ?
   ::
   ?:  ?=(%trim -.task)
-    =/  event-args  [[eny duct now rof] server-state.ax]
+    =*  event-args  [[eny duct now rof] server-state.ax]
     =*  by-channel  by-channel:(per-server-event event-args)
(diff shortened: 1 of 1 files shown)


Fang-:
Nice, good find! And +1 on the clear commit message.

I wonder if this could be solved more directly/broadly by just making the `=/  event-args` on line 2181 a `=*  event-args` instead. That way it should get re-evaluated in the way we want on every invocation of `by-channel`. If that works, I think that would be the better patch.

joemfb:
That's a great catch! This pattern of pinning a copy of the state seems suspect (and is used several other places in %eyre).

bacwyls:
> Nice, good find! And +1 on the clear commit message.
> 
> I wonder if this could be solved more directly/broadly by just making the `=/ event-args` on line 2181 a `=* event-args` instead. That way it should get re-evaluated in the way we want on every invocation of `by-channel`. If that works, I think that would be the better patch.

Good idea. I tried this out and it worked as expected.

Fang- (approved):
Cool, this looks good. Thank you!

PR #5889, opened 2022-07-11
vere: use pread/pwrite instead of lseek-then-read/write

ritfed-tabryg:
This is a pretty minor change, but it saves a few lines in events.c and a few thousand syscalls per snapshot. I don't think it's any less portable than the existing code as `pread` & `pwrite` are POSIX.

### DIFF
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -384,11 +384,8 @@ _ce_patch_verify(u3_ce_patch* pat_u)
     c3_w mug_w = pat_u->con_u->mem_u[i_w].mug_w;
     c3_w mem_w[1 << u3a_page];
 
-    if ( -1 == lseek(pat_u->mem_i, (i_w << (u3a_page + 2)), SEEK_SET) ) {
-      fprintf(stderr, "loom: patch seek: %s\r\n", strerror(errno));
-      return c3n;
-    }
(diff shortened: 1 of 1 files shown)


joemfb (changes requested):
This seems like a change worth making, with performance and readability gains. Unfortunately, Windows compatibility rears its head.

You could add `pread()`/`pwrite()` wrappers in `compat/mingw/compat.c` (just making the `lseek()` and `read()`/`write()` calls, and noting that the wrappers statefully modify the file offset and are not thread safe). Or you could write wrappers directly on the win32 apis, which appear support offset arguments. (A search turns up discussions about adding these to mingw, or implementing them for other projects.)

Also, these changes should be made on top of the `next/vere` branch, which already has some refactoring of `events.c`.

zalberico:
@ritfed-tabryg thanks for the PR - if you're able to send this in against next/vere (if it's still relevant there) it's likely something we can use in the near future. As is, I'm closing this.

zalberico:
Why reopen if it's already merged?

joemfb:
Strike that. That's worked for me before, but now it gives an error: "There are no new commits between base branch 'next/vere' and head branch 'pread-pwrite'".

I prefer the status to be clear, closed (red x) PRs being actually accepted and merged is confusing. In any case, the changes will be in the next release.

PR #5894, opened 2022-07-13
naive: uncomment some naive.hoon tests

drbeefsupreme:
I had commented out some of these tests because running the full `+test-rut` can take ~s15 to ~s30 or more depending on hardware. I did not realize that running these tests was actually part of the CI pipeline at the time, so I'm uncommenting them.

This should be uncontroversial, but since it adds a non-trivial amount of time to CI I could see an argument for not including them - but maybe that's just evading the point of CI and it takes however long it takes. Someone more experienced ought to have a better idea than me.

### DIFF
--- a/pkg/arvo/tests/lib/naive.hoon
+++ b/pkg/arvo/tests/lib/naive.hoon
@@ -739,19 +739,19 @@
   ++  gen-rut-jar
     ^~  ^-  (jar @p event)
     =/  filter  ;:  cork
-                    (cury filter-owner %.y)
+                    ::(cury filter-owner %.y)
                     ::(cury filter-proxy %spawn)
-                    (cury filter-nonce %.y)
+                    ::(cury filter-nonce %.y)
                     ::(cury filter-rank %galaxy)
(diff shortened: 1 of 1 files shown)


ynx0:
It's because those commits are in master but not in next/arvo which is your mergebase.

ynx0:
Yeah the reason is because next/arvo is lagging a bit behind master instead of being a "superset" of it, in which case there would be no problem starting off with master. 

I always just use the branch I am targeting as the starting point unless there's something I need from say master, which at that point someone would just need to to merge master into next/arvo and you'd be good.

joemfb:
`next/arvo` is just lagging, `master` should've been merged back into it. I just did so, if you push changes to your branch github will pick up on that.

drbeefsupreme:
I had previously merged next/arvo into this branch to make the unwanted commits disappear at the expense of adding a new merge commit. Decided to try resetting and force-pushing from the original commit instead, and now there are no longer unwanted commits.

philipcmonk (approved):
Sorry for the delay, this looks good

PR #5901, opened 2022-07-15
landscape: new system preferences search

patosullivan:
Closes https://github.com/tloncorp/homestead/issues/376

### DIFF
--- a/pkg/grid/src/preferences/SearchSystemPrefences.tsx
+++ b/pkg/grid/src/preferences/SearchSystemPrefences.tsx
@@ -1,0 +1,115 @@
+import React, { ChangeEvent, useEffect, useState } from 'react';
+import { Link } from 'react-router-dom';
+import fuzzy from 'fuzzy';
+import MagnifyingGlassIcon from '../components/icons/MagnifyingGlassIcon';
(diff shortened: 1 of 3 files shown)


pkg/grid/src/preferences/SearchSystemPrefences.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  const handleBlur = () => {
+    setSearchInput('');
+  };
+
+  useEffect(() => {
+    const results = fuzzy.filter(searchInput, navOptions, { extract: (obj) => obj.title });
+    const matches = results.map((el) => el.string);
+    setMatchingNavOptions(matches);
+  }, [searchInput]);
```
liam-fitzgerald, line 108: should this be thrown into a debounce or similar?

liam-fitzgerald (commented):
looks mostly good, few perf questions

pkg/grid/src/preferences/SearchSystemPrefences.tsx
patosullivan: I don't think it's necessary since we're not hitting the backend.

PR #5902, opened 2022-07-15
http-api: prevent double-processing of events

Fang-:
I had independently discovered and patched #5864, only to discover my branch was just too stale. Still, I think I can add additional value here.

In reconnect scenarios (or wonky network situations) we may receive events we had already heard. Here we make sure to drop those.  
(Not that this library currently does reconnects in that way, but it should probably be made to.)

Also simplifies the getEventId() logic, and clarifies the difference between the different `*EventId` values.


arthyn (approved):
LGTM

PR #5903, opened 2022-07-16
hoon: thread compiler cache through +sell

philipcmonk:
Printing some vases is extremely slow without sharing this cache.
Previously, this was shared "accidentally" because all the compiler
calls were run in the same road; now we do it explicitly.

A pathological case is `!,  *hoon  a=1`.  This change reduces the time
required to run +sell on that (eg by entering it in the dojo) from 43
seconds to 320ms.

Also added a `!.` to `+mean` because without it adding a `!:` to the
top of hoon.hoon creates really bad stack traces in some cases.

We should be on the lookout for this kind of slowdown elsewhere.


jtobin (approved):
Nice one, LGTM.

Issue #5914, opened 2022-07-29
Inputting an malformed ship name to `urbit -F` results in bugged terminal display


bradley-ray:
**Describe the bug**
When creating a fake ship that contains an invalid ship name, the terminal becomes bugged resulting in typed text not appearing and causing issues with formatting. This requires running the `reset` cmd to return to normal. An example of this can be seen in the screenshot below.

**To Reproduce**
Steps to reproduce the behaviour:
1. urbit -F *invalid-ship-name*

**Expected behaviour**
The terminal should return to normal when exiting.

**Screenshots**
![urbit_err](https://user-images.githubusercontent.com/89326133/181667656-4cf86c77-bc54-42c0-bd64-7056de66e3e9.png)

**System (please supply the following information, if relevant):**
 - OS: linux64
 - Urbit: 1.10

**Additional context**
The issue seems to be that `u3_term_log_init()` but when an invalid ship name is used it exits before calling `u3_term_log_exit()`. This could potentially also be happening elsewhere if any other errors occur.

PR #5915, opened 2022-07-29
vere: clean up terminal on invalid fake ship name

bradley-ray:
Currently when creating a fake ship, if an invalid ship name is given,
then the program exits without ever cleaning up the terminal. This
results in  a bugged termianal that requires closing and repopening
or using the `reset` cmd.

This commits adds a call to `u3_term_log_exit()` and `fflush(stdout)`
before calling `exit(1)` to ensure proper cleanup.

Address issue #5914

### DIFF
--- a/pkg/urbit/vere/king.c
+++ b/pkg/urbit/vere/king.c
@@ -578,7 +578,6 @@ _boothack_doom(void)
     if ( u3_nul == whu ) {
       u3l_log("boot: malformed -F ship %s\r\n", u3_Host.ops_u.fak_c);
       u3_king_bail();
-      exit(1);
     }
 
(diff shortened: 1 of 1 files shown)


joemfb (changes requested):
This is a good catch. We put the tty in raw mode for interactivity, but it's obviously important that we restore the original settings on exit.

Instead of explicitly cleaning up here, you can just call `u3_king_bail()`, which will clean up as needed and `exit(1)`.

bradley-ray:
Thanks for pointing this out. I made the change, so it should be good to go.

joemfb (approved):
LGTM, thanks

PR #5916, opened 2022-07-29
landscape: add urbit mock http api to landscape/grid

patosullivan:
No vercel previews yet, but this should be enough to tee that up.

Not every request is mocked yet, just everything necessary to get the app to render and perform basic functions. I figure we'll add more as time goes on.


patosullivan:
Closed in favor of: https://github.com/urbit/garden/pull/1

PR #5918, opened 2022-07-31
eyre: add /~/name endpoint

pkova:
One day we will have userspace permissioning, hopefully preventing evil software from performing evil deeds. This evil software could nevertheless distribute a frontend that steals the login cookie for a ship (XSS). This PR mitigates that risk by adding the `HttpOnly` attribute to the login cookie, preventing Javascript in the browser from accessing it.

There seems to be two ways of doing ship name discovery in the Urbit codebase. The first one is injecting `window.ship` through `session.js` [like this](https://github.com/urbit/urbit/blob/master/pkg/landscape/app/file-server.hoon#L206), and the other is parsing the auth cookie [like this](https://github.com/urbit/urbit/blob/master/pkg/npm/http-api/src/Urbit.ts#L181). This PR adds the endpoint `/~/name` to eyre which responds with the `@p` of the running ship if the request is authenticated. I also included changes to `http-api` to account for not being able to parse the cookie anymore. This new endpoint could also replace the wonky `session.js` stuff but that change can safely be made later.


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -707,44 +707,58 @@
...
+  ::  +handle-name: respond with our @p or 403
+  ::
+  ++  handle-name
+    |=  [authenticated=? =request:http]
+    ^-  (quip move server-state)
+    ?.  authenticated
+      (error-response authenticated url.request 403 ~)
+    ?.  =(%'GET' method.request)
+      (error-response authenticated url.request 405 "may only GET name")
+    %^  return-static-data-on-duct  200  'text/plain'
```
pkova, line 727: Any thoughts on the mime type here? Debated between this and `application/json`, also aware of the upcoming scrying raw nouns stuff.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2532,9 +2549,37 @@
...
+  |=  old=$%(axle axle-old)
   ^+  ..^$
-  ..^$(ax old)
+  ::
+  ~!  %loading
+  ?-    -.old
+      %~2020.10.18
+    =/  [* d=duct *]
+      %+  snag  0
+      %+  skim  bindings.server-state.old
```
pkova, line 2567: This a fairly wonky way to get the `duct` from the `%init` task, any better ideas?

pkg/arvo/sys/lull.hoon
```diff
@@ -1401,6 +1401,9 @@
         ::  gall scry endpoint
         ::
         [%scry ~]
+        ::  internal name endpoint, returns ship @p
+        ::
+        [%name ~]
```
Fang-, line 1406: I would consider renaming this (and the `/~/name` path) to say `host` instead. There may or may not be a world in which you can authenticate as yourself on others' ships, and this naming would be ambiguous.

pkg/arvo/sys/vane/eyre.hoon
Fang-: `text/plain` is simplest, I think this is good.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -767,11 +781,6 @@
       ^-  (unit (unit cage))
       (rof ~ care [our desk da+now] path)
     ::
-    ++  error-response
-      |=  [status=@ud =tape]
-      ^-  (quip move server-state)
-      %^  return-static-data-on-duct  status  'text/html'
-      (error-page status authenticated url.request tape)
```
Fang-, line 774: You might want to just leave this here, and have it call the outer one, so the callsites within this arm remain cleaner.

Alternatively, arguably even better, you could leave this as-is, and wrap just the `+handle-name` callsite in `return-static-data-on-duct`, and include the appropriate `error-page` calls there.

Main thing is just that it's here rn as a helper for this `|^`,  moving it out doesn't seem worth the sample-complexity cost to me. And the newly added callsite isn't particularly involved.

pkg/arvo/sys/vane/eyre.hoon
Fang-: The reason you're getting the duct from the `%init` task is... so that the binding is exactly equal between newly booted ships, and ones that went through this upgrade path?

I probably wouldn't worry about this. The way this system is set up right now, using ducts for "permissions", is flawed in many ways. And it's a binding in the reserved namespace (`/~`), so eyre's the only one who can touch it anyway.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2532,9 +2549,37 @@
...
+      %+  skim  bindings.server-state.old
+        |=  [=binding =duct =action]
+        =(%authentication -.action)
+    =/  new
+    %=  old
+      date  %~2022.7.26
+    ::
+      bindings.server-state
+      ^-  (list [binding duct action])
+      :-  [[~ /~/name] d [%name ~]]  bindings.server-state.old
```
Fang-, line 2576: You'll want to call `+insert-binding`, the ordering in this list does actually matter.

```suggestion
        bindings.server-state
      (insert-binding [[~ /~/name] d [%name ~]] bindings.server-state.old)
```

Fang-: This whole block in/under the `=/  new` could use some indents or further styling love. You can probably get away with just `%=`ing right away, something like:

```hoon
%=  $
  date.old  %~2022.7.26
  bindings.server-state.old  etc
==
```

pkg/npm/http-api/src/Urbit.ts
```diff
@@ -172,17 +189,22 @@ export class Urbit {
...
-    }).then((response) => {
+    }).then(async response => {
       if (this.verbose) {
         console.log('Received authentication response', response);
       }
-      const cookie = response.headers.get('set-cookie');
-      if (!this.ship) {
-        this.ship = new RegExp(/urbauth-~([\w-]+)/).exec(cookie)[1];
+
+      if (response.status !== 204) {
```
Fang-, line 197: Feels like we should just check for `>= 200` and `< 300` here...

Fang- (changes requested):
Thanks! Some comments and suggestions, mostly about the `/~/name` binding. It sucks that we need this, but also think this approach is better than the status quo of "userspace must remember to reimplement a way to get the host ship name". We'll probably want to update the http api to just allow saying `our` at some point.

The cookie change, making it `HttpOnly`, is probably the right thing to do for now. Tempting to say "just don't install malicious software lol", but even friendly software might have been built with malicious js dependencies, etc etc. This is definitely safer.

Tagged @arthyn for review of the http-api, but looks mostly fine to me.

We'll need to come up with a plan for releasing this, since the changes here do break the old http-api, and maybe some hosting logic too.

pkg/arvo/sys/lull.hoon
pkova: Don't like `host`. My idea was for the endpoint to answer the question "what is the name of this ship?", whereas if the question is "what is the host of this ship?" my mind immediately goes to stuff like hosting providers which are unrelated. What do you think about `/~/our`?

pkg/arvo/sys/vane/eyre.hoon
pkova: Yeah that was my idea. 

Do you think we could construct a `duct` here that somehow communicates that it came from an eyre upgrade or just punt and go with `*duct`?

pkova: Put some prints in and found out it's `~[/dill //term/1]` on a fakezod. I assume that's what it always is so I'll just hardcode that.

pkg/arvo/sys/vane/eyre.hoon
pkova: Fixed.

pkg/npm/http-api/src/Urbit.ts
pkova: Fixed.

pkg/arvo/sys/vane/eyre.hoon
pkova: Yeah you're right, my refactor here was a bad idea. Reverted to include `+error-response` inside `+handle-name`.

pkg/arvo/sys/vane/eyre.hoon
Fang-: Ah, yes. It is indeed always that, but it's better to use `outgoing-duct` from the state, which should be equivalent.

pkova: Ah I see, fixed.

pkg/arvo/sys/lull.hoon
Fang-: As discussed out of band, `/~/name` is probably right. Might be worth updating the comment to explicitly say it's the name of the ship the requester is authenticated as though, to avoid confusion about this in the future.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2532,9 +2554,31 @@
...
+  ?-    -.old
+      %~2020.10.18
+    %=  $
+      date.old  %~2022.7.26
+      ::
+      bindings.server-state.old
+      %+  insert-binding
+        [[~ /~/name] outgoing-duct.server-state.old [%name ~]]
+      bindings.server-state.old
+    ==
```
Fang-, line 2577:
```suggestion
    %=  $
      date.old  %~2022.7.26
    ::
        bindings.server-state.old
      %+  insert-binding
        [[~ /~/name] outgoing-duct.server-state.old [%name ~]]
      bindings.server-state.old
    ==
```

Fang- (approved):
Some nits, but otherwise this looks good afaic.

As far as releasing this goes, perhaps we want to write the updated `Urbit.ts` in such a way that it only uses the endpoint if it can't read the cookie? That we we can release that first, let it sit for a while, and then eventually push out the Eyre change.

pkova:
Style nit fixed, modified endpoint description in `/sys/lull` and `Urbit.ts` now tries parsing the cookie, using the new endpoint if that fails.

Fang- (dismissed):
Cool, thanks for the changes!

arthyn (dismissed):
This looks good to me!

zalberico:
@Fang- and @arthyn - this looks like it has a failing test, I'd like to get this over the line and merged since you've both approved it and there was a lot of work here.

@Fang- I'm going to assign this one to you, can we get it passing and merged?

jalehman:
@pkova can you update this by resolving conflicts and updating the base branch? Then we can see where the tests are at and figure out how to finish this up.

pkova:
Rebased and removed the `Urbit.ts` stuff that went in in #6084.

Fang- (approved):
Still good. Let's get it!

tadad:
This PR breaks backwards compatibility, so the next release we should be decrementing a kelvin. See here : https://github.com/urbit/urbit/issues/6342 .

zalberico:
@tadad https://github.com/urbit/urbit/pull/6324 <- the intent of the partial revert in that case was to not break backwards compatibility did you see that or is there still an issue?

tadad:
my point is that the name endpoint also breaks backwards compatibility because it changes the type of a scry. In blog I do a `?-` on the `-.action:eyre` which I scry from `%e /=bindings=`, which broke when this got merged into develop

PR #5923, opened 2022-08-04
ames: recork better

belisarius222:
The previous recork timer queued up %cork messages without sending them.
It also relied on making sure pump timers didn't get set for recork bones.
This was fragile.

The new design enqueues up to one new %cork message per ship during each
recork timer, based on the state of the flow.  If the flow is closing but
there are no outstanding messages in it, then it needs to be recorked.
Flows will be recorked in ascending numerical order by bone.


Fang- (approved):
As discussed out of band, this looks good, and has been tested on a real ship to give better behavior.

PR #5927, opened 2022-08-06
eyre: Add cache

lukechampine:
This fulfills the first milestone of https://urbit.org/grants/eyre-scry. It adds a simple cache to eyre that maps URLs to scry paths, with the usual accessors for updating the cache. A gift is given to the runtime when a watched file changed, and the full cache is sent to the runtime on init.

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1354,6 +1354,11 @@
   ::                                                    ::::
 ++  eyre  ^?
   |%
+  +$  cache-entry
+    $:  auth=?
+    $=  body
+    $%  [%payload =simple-payload:http]
+    ==  ==
   +$  gift
     $%  ::  set-config: configures the external http server
         ::
@@ -1373,6 +1378,9 @@
         ::    not allowed.
         ::
         [%bound accepted=? =binding]
+        ::  notification that a cache entry has changed
+        ::
+        [%grow =path]
     ==
   ::
   +$  task
@@ -1428,6 +1436,9 @@
         ::  %spew: set verbosity toggle
         ::
         [%spew veb=@]
+        ::  remember (or update) a cache mapping
+        ::
+        [%set-response url=@t entry=(unit cache-entry)]
     ==
   ::  +origin: request origin as specified in an Origin header
   ::
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -68,7 +68,7 @@
 ::
 |%
 +$  axle
-  $:  %~2023.2.17
+  $:  %~2023.3.16
       =server-state
   ==
 ::  +server-state: state relating to open inbound HTTP connections
@@ -84,6 +84,9 @@
       ::    the :binding into a (map (unit @t) (trie knot =action)).
       ::
       bindings=(list [=binding =duct =action])
+      ::  cache: mapping from url to versioned entry
+      ::
+      cache=(map url=@t [aeon=@ud val=(unit cache-entry)])
       ::  cors-registry: state used and managed by the +cors core
       ::
       =cors-registry
@@ -672,6 +675,11 @@
           =-  (fall - '*')
           (get-header:http 'access-control-request-headers' headers)
       ==
+    ::  handle requests to the cache
+    ::
+    =/  entry  (~(get by cache.state) url.request)
+    ?:  &(?=(^ entry) ?=(%'GET' method.request))
+      (handle-cache-req authenticated request val.u.entry)
     ::
     ?-    -.action
         %gen
@@ -764,6 +772,32 @@
       (error-response 405 "may only GET name")
     %^  return-static-data-on-duct  200  'text/plain'
     (as-octs:mimes:html (scot %p our))
+    ::
+    ++  error-response
+      |=  [status=@ud =tape]
+      ^-  (quip move server-state)
+      %^  return-static-data-on-duct  status  'text/html'
+      (error-page status authenticated url.request tape)
+    --
+  ::  +handle-cache-req: respond with cached value, 404 or 500
+  ::
+  ++  handle-cache-req
+    |=  [authenticated=? =request:http entry=(unit cache-entry)]
+    |^  ^-  (quip move server-state)
+    ?~  entry
+      (error-response 404 "cache entry for that binding was deleted")
+    ?:  &(auth.u.entry !authenticated)
+      (error-response 403 ~)
+    =*  body  body.u.entry
+    ?-    -.body
+        %payload
+      %-  handle-response
+      :*  %start
+          response-header.simple-payload.body
+          data.simple-payload.body
+          complete=%.y
+      ==
+    ==
     ::
     ++  error-response
       |=  [status=@ud =tape]
@@ -2032,6 +2066,15 @@
           %leave  ~
       ==
     --
+  ::  +set-response: remember (or update) a cache mapping
+  ::
+  ++  set-response
+    |=  [url=@t entry=(unit cache-entry)]
+    ^-  [(list move) server-state]
+    =/  aeon  ?^(prev=(~(get by cache.state) url) +(aeon.u.prev) 1)
+    =.  cache.state  (~(put by cache.state) url [aeon entry])
+    :_  state
+    [outgoing-duct.state %give %grow /cache/(scot %ud aeon)/(scot %t url)]~
   ::  +add-binding: conditionally add a pairing between binding and action
   ::
   ::    Adds =binding =action if there is no conflicting bindings.
@@ -2318,6 +2361,12 @@
     ::  save duct for future %give to unix
     ::
     =.  outgoing-duct.server-state.ax  duct
+    ::  send all cache mappings to runtime
+    ::
+    =/  cache-moves=(list move)
+      %+  turn  ~(tap by cache.server-state.ax)
+      |=  [url=@t cache-val=[aeon=@ud val=(unit cache-entry)]]
+      [duct %give %grow /cache/(scot %u aeon.cache-val)/(scot %t url)]
     ::
     :_  http-server-gate
     :*  ::  hand back default configuration for now
@@ -2328,7 +2377,7 @@
         =<  give-session-tokens
         (per-server-event [eny duct now rof] server-state.ax)
       ::
-        closed-connections
+        (zing ~[closed-connections cache-moves])
     ==
   ::
   ?:  ?=(%code-changed -.task)
@@ -2447,6 +2496,10 @@
       %spew
     =.  verb.server-state.ax  veb.task
     `http-server-gate
+  ::
+      %set-response
+    =^  moves  server-state.ax  (set-response:server +.task)
+    [moves http-server-gate]
   ==
 ::
 ++  take
@@ -2638,7 +2691,8 @@
       +$  axle-any
         $%  [%~2020.10.18 =server-state-0]
             [%~2022.7.26 =server-state-0]
-            [%~2023.2.17 =server-state]
+            [%~2023.2.17 =server-state-1]
+            [%~2023.3.16 =server-state]
         ==
       +$  server-state-0
         $:  bindings=(list [=binding =duct =action])
@@ -2651,6 +2705,18 @@
             ports=[insecure=@ud secure=(unit @ud)]
             outgoing-duct=duct
         ==
+      +$  server-state-1
+        $:  bindings=(list [=binding =duct =action])
+            =cors-registry
+            connections=(map duct outstanding-connection)
+            =authentication-state
+            =channel-state
+            domains=(set turf)
+            =http-config
+            ports=[insecure=@ud secure=(unit @ud)]
+            outgoing-duct=duct
+            verb=@
+        ==
       --
   |=  old=axle-any
   ^+  ..^$
@@ -2659,8 +2725,9 @@
     =,  server-state-0.old
     %=  ..^$
       ax  ^-  axle
-          :*  %~2023.2.17
+          :*  %~2023.3.16
               (insert-binding [[~ /~/name] outgoing-duct [%name ~]] bindings)
+              *(map url=@t [aeon=@ud val=(unit cache-entry)])
               cors-registry
               connections
               authentication-state
@@ -2676,8 +2743,9 @@
     =,  server-state-0.old
     %=  ..^$
       ax  ^-  axle
-          :*  %~2023.2.17
+          :*  %~2023.3.16
               bindings
+              *(map url=@t [aeon=@ud val=(unit cache-entry)])
               cors-registry
               connections
               authentication-state
@@ -2690,6 +2758,24 @@
     ==    ==
   ::
       %~2023.2.17
+    =,  server-state-1.old
+    %=  ..^$
+      ax  ^-  axle
+          :*  %~2023.3.16
+              bindings
+              *(map url=@t [aeon=@ud val=(unit cache-entry)])
+              cors-registry
+              connections
+              authentication-state
+              channel-state
+              domains
+              http-config
+              ports
+              outgoing-duct
+              verb
+    ==    ==
+  ::
+      %~2023.3.16
     ::  enable https redirects if certificate configured
     ::
     =.  redirect.http-config.server-state.old
@@ -2757,6 +2843,14 @@
       %-  =<  request-is-logged-in:authentication
           (per-server-event [eny *duct now rof] server-state.ax)
       %*(. *request:http header-list ['cookie' u.cookies]~)
+    ::
+        [%cache @ @ ~]
+      ?~  aeon=(slaw %ud i.t.tyl)        [~ ~]
+      ?~  url=(slaw %t i.t.t.tyl)        [~ ~]
+      ?~  entry=(~(get by cache) u.url)  [~ ~]
+      ?.  =(u.aeon aeon.u.entry)         [~ ~]
+      ?~  val=val.u.entry                [~ ~]
+      ``noun+!>(u.val)
     ==
   ?.  ?=(%$ ren)
     [~ ~]


pkg/arvo/sys/lull.hoon
```diff
@@ -1163,6 +1163,9 @@
         ::    not allowed.
         ::
         [%bound accepted=? =binding]
+        ::  notification that the clay file associated with url has changed
+        ::
+        [%cache-changed url=@t =beam]
```
lukechampine, line 1168: Is `beam` the right type to use for this? It seems to have the information we care about, but I wasn't sure.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1991,6 +2024,22 @@
...
+    =/  =riff:clay  [q.beam `[%next %x r.beam s.beam]]
+    :_  state
+    [duct %pass /watch-cache/[url] %c %warp our riff]~
+  ::  +del-cache: forget a URL->scry mapping
+  ::
+  ++  del-cache
+    |=  url=@t
+    ^-  [(list move) server-state]
+    =.  cache.state  (~(del by cache.state) url)
+    `state
```
lukechampine, line 2042: note that this doesn't immediately cancel the existing clay subscription. `watch-cache` will discard events that don't have a matching cache entry, so the subscription won't be renewed. But if we're worried about leaking subscriptions, we could explicitly cancel here.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2445,6 +2512,21 @@
...
+    =/  =beam  [[our r.p.u.p.sign q.p.u.p.sign] q.u.p.sign]
+    =/  url=@t  (spat t.wire)
+    ?.  (~(has by cache.server-state.ax) url)
+      `http-server-gate
+    =/  moves
+      :~
+        [duct %give %cache-changed url beam]
+        [duct %pass wire %c %warp our [q.beam `[%next %x r.beam s.beam]]]
+      ==
+    [moves http-server-gate]
```
tacryt-socryp, line 2528: Just a style nit, but I'd inline the moves variable and just put something like this:
```
:_  http-server-gate
:~  [duct %give ...]
    [duct %pass ...]
==
```

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2445,6 +2512,21 @@
       (handle-response http-event)
     [moves http-server-gate]
   ::
+  ++  watch-cache
+    ::
```
tacryt-socryp, line 2516: Unnecessary comment

tacryt-socryp (commented):
Preliminary notes, haven't run this yet.

pkg/arvo/sys/vane/eyre.hoon
lukechampine: I don't get it either, but it seems to be the convention in this file

belisarius222:
@Fang- and I have been discussing this.

We keep learning that Clay is only really good for managing source code.  Other kinds of data, including other files, are better off being stored by Gall agents.  Empirically, @Fang-'s suite of `~paldev` apps and `~tirrel`'s Studio app all manage their own data without using Clay.  This implies that to make it easy on developers, we should directly support agents dynamically binding URLs to data, rather than indirectly by letting them stick data in Clay and tell Eyre to serve the data from Clay.

What's simplest for a userspace dev is to `%pass` a move to Eyre saying "bind URL `/foo/bar` to HTTP response `[%http-response 200 'text/html' ...]`" (some noun representing the response; I don't know the format offhand).  When Eyre hears this, it stores the response in its state and notifies Vere.  Vere then invalidates its own cache.

The Arvo <-> Vere interaction should probably use the scry namespace -- for example, Eyre could bind a scry path like `/e/x/foo/bar/2`, incrementing a revision number each time.  This doesn't mean the agent->Eyre interaction needs to use the scry namespace, though.  For now, agents only support getting scry'd at the current date (and without perfect referential transparency, since it could change during an event).

Even once we have solid-state subscriptions, it's not clear that would make much of a difference.  It's still more effort to set up a publication and tell Eyre to look at it, rather than just directly telling Eyre to change the response for a URL, and that publication probably wouldn't be used by anything other than Eyre in most cases.

Eventually we'll want this sort of interaction for Clay data too, probably.  It would certainly be nice not to have to write an agent just to serve Clay data over the web.  It's worth noting that for the design we're proposing, we could build a "file-server" agent, probably in `%base`, to support serving files from Clay.  I don't know if we need it with the initial release, though, given that people aren't serving data from Clay in the first place.

@tacryt-socryp and @ixv, curious to hear your thoughts on this.  Our inclination is to revise the proposal (sorry @lukechampine, and thanks for the contribution!) to reflect these observations, but you guys would be using this too, so it would be good to get feedback.

tacryt-socryp:
> What's simplest for a userspace dev is to %pass a move to Eyre saying "bind URL /foo/bar to HTTP response [%http-response 200 'text/html' ...]" (some noun representing the response; I don't know the format offhand). When Eyre hears this, it stores the response in its state and notifies Vere. Vere then invalidates its own cache.

I have no personal objection to this. Notably, we decided on this design after talking to you, @philipcmonk, and @joemfb. I'm happy to move toward whatever implementation the core team can agree on, and my explicit understanding after talking to all of you about it was that this particular implementation was agreed upon. If we can come to a concrete singular decision on what implementation is best between all of you, let's get that implemented.

tacryt-socryp:
> Eventually we'll want this sort of interaction for Clay data too, probably. It would certainly be nice not to have to write an agent just to serve Clay data over the web. It's worth noting that for the design we're proposing, we could build a "file-server" agent, probably in %base, to support serving files from Clay. I don't know if we need it with the initial release, though, given that people aren't serving data from Clay in the first place.

In regards to this, sounds good, but as you say, is not important to this grant.

lukechampine:
Storing cache data directly in the `eyre` state sounds good to me. Would certainly simplify a few things. Is the plan to revise the proposal in this thread, or will that happen somewhere else?

tacryt-socryp:
Plan is for me to talk to @belisarius222 and @joemfb and iron out any conflicting opinions, and once I can get any sort of agreement, we'll update the proposal. I apologize for this.

joemfb:
> We keep learning that Clay is only really good for managing source code.

@belisarius222 can you expound on this? %clay is not very suitable for arbitrary outside data (due to marks and certain conversions being required), but it otherwise seems fine for this. The other past concern I'm aware of was the inability to delete/tombstone historical data, but that has been rectified.

I see some things to quibble with in this PR:

- hardcoding the `text/html` MIME type instead of mark-converting to %mime
- having %eyre subscribe to %clay to update bindings instead of letting a %gall agent handle that sort of pattern
- non referentially-transparent URL -> namespace bindings as the only/default option (I'd love a pure function from URL to path, perhaps with cookie-auth represented through `+peek`'s `gang`)

But a basic integration with %clay seems superior to a trivial "filesystem" in %eyre, especially if the output-format can be produced on demand and need not be "pre-materialized". Is this primarily a question of api ergonomics?

Fang-:
I've written up a slightly more fleshed out version of the proposal @belisarius222 described [here](https://gist.github.com/Fang-/1939a5b51fb2f821d960e05b74ad5f9f), please take a look and let me know what you think. (Note that the section on a procedural publication cache may be considered out of scope for the grant work being done here. Still a good idea to account for it in the broader picture though.)

belisarius222:
https://gist.github.com/belisarius222/2c4b0939f987fcc56252f4e011ef17b7

@joemfb and I wrote up some thoughts on this too.

Fang-:
Looks like we're mostly in agreement here. The gist you linked is the more forward-looking side of the story, but the short-term approach it settles on aligns with what I outlined.

I have updated [the section of my gist that talks about authentication](https://gist.github.com/Fang-/1939a5b51fb2f821d960e05b74ad5f9f#authentication). tldr: I think we may be able to support auth checks for cached responses, and I would be happy to do the ancillary work necessary for that.

tacryt-socryp:
Thanks for your posts here. I've read them and understand the rationales. @lukechampine I will be chatting with them in more depth shortly, but it does look like there's consensus on the approach, and it shouldn't be too much of a modification on what you already have from the looks of it.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -634,6 +640,11 @@
           =-  (fall - '*')
           (get-header:http 'access-control-request-headers' headers)
       ==
+    ::  handle requests to the cache
+    ::
+    =/  cache-binding  (~(get by cache-bindings.state) url.request)
+    ?.  ?=(~ cache-binding)
```
tacryt-socryp, line 646: `?^  cache-binding`

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,42 @@
...
+    ?.  authenticated
+      (error-response 403 ~)
+    ?.  =(%'GET' method.request)
+      (error-response 405 "may only GET scries")
+    =/  cache-entry  (~(get of cache.state) path)
+    ?~  cache-entry
+      (error-response 404 "no cache entry for that binding")
+    ?-    -.u.cache-entry
+        %payload
+      =/  payload  simple-payload.u.cache-entry
```
tacryt-socryp, line 739: `=*  payload  simple-payload.u.cache-entry`

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,42 @@
...
+      %-  handle-response
+      :*  %start
+          response-header.payload
+          data.payload
+          complete=%.y
+      ==
+        %scry
+      =/  res=(unit (unit cage))  (rof ~ %cx [our %base da+now] path.u.cache-entry)
```
tacryt-socryp, line 747: The `%scry` path should specify the `case` with a version number, as `da+now` scries cannot be cached.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,42 @@
...
+      ?~  res    (error-response 500 "failed scry")
+      ?~  u.res  (error-response 404 "no scry result")
+      =*  vase   q.u.u.res
+      ?>  ?=(@ q.vase)
+      %^  return-static-data-on-duct  200  'text/html'
+        (as-octs:mimes:html q.vase)
```
tacryt-socryp, line 753: de-dent this line once

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1991,6 +2038,35 @@
       ^-  ?
       &(=(item-binding binding) =(item-duct duct))
     ==
+  ::  +add-cache: remember a cache mapping
+  ::
+  ++  add-cache
+    |=  [=path =cache-entry]
```
tacryt-socryp, line 2044: Note that this trie is meant to be referentially transparent, so we desire to keep the history of all previous cache-entries in this trie slot, but only allow scrying out the most recently added one for now.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1991,6 +2038,35 @@
...
+    ^-  [(list move) server-state]
+    =.  cache.state  (~(put of cache.state) path cache-entry)
+    `state
+  ::  +del-cache: forget a cache mapping
+  ::
+  ++  del-cache
+    |=  =path
```
tacryt-socryp, line 2051: Similarly, deletion should remove the data inside the binding for the cache-entry, leaving a ~ to denote that data was once there but is no longer available.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1991,6 +2038,35 @@
...
+    ^-  [(list move) server-state]
+    =.  cache.state  (~(del of cache.state) path)
+    `state
+  ::  +add-cache-binding: remember a cache binding
+  ::
+  ++  add-cache-binding
+    |=  [url=@t =path]
+    ^-  [(list move) server-state]
+    =.  cache-bindings.state  (~(put by cache-bindings.state) url path)
```
tacryt-socryp, line 2060: the cache-bindings are not meant to be referentially transparent, so this is good as is

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1991,6 +2038,35 @@
...
+    :_  state
+    [duct %give %cache-changed url path]~
+  ::  +del-cache-binding: forget a cache binding
+  ::
+  ++  del-cache-binding
+    |=  url=@t
+    ^-  [(list move) server-state]
+    =.  cache-bindings.state  (~(del by cache-bindings.state) url)
```
tacryt-socryp, line 2068: note that this should also issue a `%cache-changed` move

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2234,6 +2310,12 @@
     ::  save duct for future %give to unix
     ::
     =.  outgoing-duct.server-state.ax  duct
+    ::  send all cache bindings to runtime
+    ::
+    =/  cache-moves=(list move)
```
tacryt-socryp, line 2315: I don't think this is necessary. Maybe I don't understand its purpose?

tacryt-socryp (changes requested):
Overall looks good, just need to make the trie referentially transparent.

pkg/arvo/sys/vane/eyre.hoon
lukechampine: ah, I think I see the rule -- use `=/` when computing new values, and `=*` when referencing existing values. Is that accurate?

pkg/arvo/sys/vane/eyre.hoon
lukechampine: this comes from the initial proposal, which said:

> Eyre will give a %gift to the runtime upon startup and whenever a binding is added or removed from the map.

which I guess is ambiguous about what exactly the startup %gift should contain.

pkg/arvo/sys/vane/eyre.hoon
lukechampine: ah. Would a `jar` be more appropriate than `axal`, then?

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: That's right!

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: Nope, We desire an `+axal` here for fast traversal, likely with a `(list)` inside where each element contains its version number and the appropriate data.

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: Ah, if this is happening on `born` then it totally makes sense. For some reason I thought this was happening on any HTTP request.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1961,6 +2008,40 @@
...
+    ^-  [(list move) server-state]
+    =/  cache-entries  (fall (~(get of cache.state) path) ~)
+    =.  cache.state  (~(put of cache.state) path [cache-entry cache-entries])
+    `state
+  ::  +del-cache: forget a cache mapping
+  ::
+  ++  del-cache
+    |=  =path
+    ^-  [(list move) server-state]
+    =.  cache.state  (~(put of cache.state) path ~)
```
lukechampine, line 2024: @tacryt-socryp is this what you meant by leaving a `~` in the binding?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,42 @@
...
+    ?-    -.cache-entry
+        %payload
+      %-  handle-response
+      :*  %start
+          response-header.simple-payload.cache-entry
+          data.simple-payload.cache-entry
+          complete=%.y
+      ==
+        %scry
+      =/  res=(unit (unit cage))  (rof ~ %cx [our %base ud+case.cache-entry] path.cache-entry)
```
lukechampine, line 747: this only supports scrying from Clay btw. I assume we want to support any kind of scry, but I'm not sure how to go about that.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,42 @@
...
     ==
+ ::  +handle-cache-req: respond with cached value, 404 or 500
+  ++  handle-cache-req
+    |=  [authenticated=? =request:http =path]
+    |^  ^-  (quip move server-state)
+    ?.  authenticated
+      (error-response 403 ~)
+    ?.  =(%'GET' method.request)
+      (error-response 405 "may only GET scries")
+    =/  cache-entries  (fall (~(get of cache.state) path) ~)
```
tacryt-socryp, line 734: Let's treat `~` cache-entries as a different case than there not being an entry present. (Equivalent to the `~` vs `[~ ~]` pattern in scries. `~` will denote that the content does not exist and has never existed, and `[~ ~]` denotes that the content has been deleted.

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: That's fine for now.

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: For now, let's replace the most recent list item with a `~`. (This will require the trie to contain `(list (unit cache-entry))`. This will allow us to have deleted the most recent item and have a different case for *is deleted* vs *does not exist*.

tacryt-socryp (changes requested):
Tiny semantic changes regarding deletion of trie items.

lukechampine:
ok, switched to a `(unit cache-entry)`

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2662,7 +2668,7 @@
       %~2020.10.18
     %_  $
       date.old          %~2022.8.22
-      server-state.old  [-.server-state.old *(axal (list cache-entry)) *(map @t path) +.server-state.old]
+      server-state.old  [-.server-state.old *(axal (list (unit cache-entry))) *(map @t path) +.server-state.old]
```
lukechampine, line 2671: I suppose this could be a `lest`, since there's no longer a way to end up with a `~` in the trie. Then we wouldn't have to use `fall` in a few places. Worth it?

tacryt-socryp: Not worth it imo

pkg/arvo/sys/lull.hoon
```diff
@@ -1147,6 +1147,11 @@
   ::                                                    ::::
 ++  eyre  ^?
   |%
+  +$  cache-entry
+    $%
+      [%scry =path case=@ud]
```
tacryt-socryp, line 1152: I think you should be able to get rid of both of these `case` variables, as they're now implicit in the list ordering in the axal data structure. After, that LGTM and we can get review from other core devs.

lukechampine: hmm, like this?
```diff
         %scry
-      =/  res=(unit (unit cage))  (rof ~ %cx [our %base ud+case.cache-entry] path.cache-entry)
+      =/  case=@ud  (sub (lent cache-entries) 1)
+      =/  res=(unit (unit cage))  (rof ~ %cx [our %base ud+case] path.cache-entry)
       ?~  res    (error-response 500 "failed scry")
```
that seems wrong to me. I don't see why the clay case would necessarily match the cache case.

tacryt-socryp: Ah, I get what you're saying. The case has a different semantic than I thought it did. Let's just remove the one on the simple-payload, then.

lukechampine: done

tacryt-socryp (approved):

tacryt-socryp:
@joemfb @Fang- This is ready for review.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -88,6 +88,12 @@
       ::    the :binding into a (map (unit @t) (trie knot =action)).
       ::
       bindings=(list [=binding =duct =action])
+      ::  cache: mapping from path->entry
+      ::
+      cache=(axal (list (unit cache-entry)))
```
joemfb, line 93: The historical bindings (`(list (unit cache-entry))`) seem to be entirely unaccessible. Is the plan to expose them through %eyre's namespace?. If so, Is the position in the list intended to correspond to a `%ud` case?

Also, will it be possible to delete entries other than the most recent? I'm worried about a space leak here.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2234,6 +2321,12 @@
     ::  save duct for future %give to unix
     ::
     =.  outgoing-duct.server-state.ax  duct
+    ::  send all cache bindings to runtime
+    ::
+    =/  cache-moves=(list move)
+      %+  turn  ~(tap by cache-bindings.server-state.ax)
+      |=  [url=@t =path]
+      [duct %give %cache-changed url path]
```
joemfb, line 2329: What is the expected behavior of the http i/o driver as relates to cache bindings/entries?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1961,6 +2010,44 @@
...
+  ::
+  ++  del-cache
+    |=  =path
+    ^-  [(list move) server-state]
+    =/  cache-entries  (fall (~(get of cache.state) path) ~)
+    ::  replace most recent entry with ~
+    ::
+    =?  cache.state  ?=(^ cache-entries)
+      (~(put of cache.state) path [~ t.cache-entries])
+    `state
```
joemfb, line 2031: Should these also `%give %cache-changed` (maybe depending on the presence of a relevant binding)?

joemfb (commented):
This code is nice and clean.

I have some questions about the requirements that it's implementing:

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: Yes.

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: The plan is that while we have no need to expose them through `%eyre`'s namespace, that they could be exposed in that manner, with the position in the list corresponding to the case (backwards, highest case is most recent in list). I think that previous bindings should be able to be deleted using the `+del-cache` method, perhaps with an optional case. That is not currently implemented, but it should be @lukechampine.

pkg/arvo/sys/vane/eyre.hoon
tacryt-socryp: For now, the behavior is to discard the gift. Can we discuss the intended future behavior of how the HTTP driver should utilize the map that is published to it sometime soon? There are a few viable implementation paths that could make use of that data efficiently, but it will be in a separate grant.

pkg/arvo/sys/vane/eyre.hoon
lukechampine: ok. How are optional arguments usually implemented? `(unit @ud)`? Also, we aren't currently exposing the case anywhere (nor the full list of entries), which makes it pretty annoying for a caller to figure out which case to specify. Perhaps the case could be included in the `%cache-changed` gift? 🤔

tacryt-socryp: `(unit @ud)` should suffice. Including the case in the `%cache-changed` gift is a great idea. Down the line it's likely that the eyre scry namespace will allow querying some of these details as well.

pkg/arvo/sys/vane/eyre.hoon
lukechampine: Should a gift be given even when deleting an entry whose case is not the most recent?

Maybe it's just me who's confused, but the intended behavior/use case of `%cache-changed` is not super clear to me, particularly in relation to the cache/cache-binding distinction.

tacryt-socryp: Imo the semantics of the `%cache-changed` gift are that when anything in the `cache-binding` data structure changes, a gift should be sent. Note that if an item in the `cache` changes that is currently referenced by `cache-binding`, that counts as changing `cache-binding`.

lukechampine: ok, added

tacryt-socryp:
@joemfb @belisarius222 @Fang- This is ready to go in. Let's go.

jalehman:
Should this PR be against `next/arvo` @belisarius222 ?

tacryt-socryp:
Rebased against next/arvo. Thanks @jalehman

tacryt-socryp:
@joemfb @Fang- Hello?

zalberico:
@belisarius222 - I've assigned this one to you to help get over the line and merged, there's a lot of work here that's been approved we should prioritize getting this merged (over the others I've also assigned to you).

Fang-:
Apologies for my extended absence on this issue. I have been partially out of commission due to personal circumstances. I'm going to make a point to take a good look at this, if not by end of week then definitely early next week.

Fang- (commented):
Again, apologies for taking so long on this, @lukechampine.

Code looks good, I mostly have questions (and some careful suggestions) about the setup & semantics of what's being implemented here. I've been out of it for a little while, so forgive if any of that should be obvious. I did read @belisarius222 and mine's proposals again as a quick refresher.

> Let's go.

Reminder that our release pipelines are still rather clogged. Once this passes review we'll get it in as soon as we reasonably can.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,44 @@
...
       (error-page 404 authenticated url.request ~)
     ==
+ ::  +handle-cache-req: respond with cached value, 404 or 500
+  ++  handle-cache-req
+    |=  [authenticated=? =request:http =path]
+    |^  ^-  (quip move server-state)
+    ?.  authenticated
+      (error-response 403 ~)
+    ?.  =(%'GET' method.request)
+      (error-response 405 "may only GET scries")
```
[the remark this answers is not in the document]
tacryt-socryp: Fallback to default flow would be best. @lukechampine

pkg/arvo/sys/lull.hoon
```diff
@@ -1147,6 +1147,11 @@
...
+      [%payload =simple-payload:http]
+    ==
```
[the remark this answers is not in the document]
tacryt-socryp: @lukechampine Public responses are the primary use case.
@Fang- Can you suggest an easy path forward? And yes, we should support alternative desks.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +723,44 @@
...
+          response-header.simple-payload.cache-entry
+          data.simple-payload.cache-entry
+          complete=%.y
+      ==
+        %scry
+      =/  res=(unit (unit cage))  (rof ~ %cx [our %base ud+case.cache-entry] path.cache-entry)
+      ?~  res    (error-response 500 "failed scry")
+      ?~  u.res  (error-response 404 "no scry result")
+      =*  vase   q.u.u.res
+      ?>  ?=(@ q.vase)
```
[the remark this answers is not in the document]
tacryt-socryp: Giving a 500 is the best way forward, easy fix: @lukechampine

pkg/arvo/sys/vane/eyre.hoon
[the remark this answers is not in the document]
tacryt-socryp: Joe requires it, please for the love of God don't make us rehash these philosophical debates.

pkg/arvo/sys/lull.hoon
```diff
@@ -1215,6 +1223,18 @@
...
         [%reject-origin =origin]
+        ::  remember (or update) a cache mapping
+        ::
+        [%add-cache =path =cache-entry]
+        ::  forget a cache mapping
+        ::
+        [%del-cache =path]
+        ::  remember (or update) a cache binding
+        ::
+        [%add-cache-binding url=@t =path]
```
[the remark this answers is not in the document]
tacryt-socryp: The paths resemble URLs. Yes, runtime will map url -> path and scry into Eyre.

tacryt-socryp:
> Reminder that our release pipelines are still rather clogged. Once this passes review we'll get it in as soon as we reasonably can.

No excuses, let's get it in.

pkg/arvo/sys/lull.hoon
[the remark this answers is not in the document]
Fang-: I think simply accepting an `auth=?` as part of the `%add-cache` (as also suggested in my [earlier proposal](https://gist.github.com/Fang-/1939a5b51fb2f821d960e05b74ad5f9f#authentication)) is easiest. You'd store is as part of the `cache-entry`, and eventually give that to the runtime so it, too, knows.

Perhaps:
```suggestion
  +$  cache-entry
    $:  auth=?
    $=  body
    $%  [%scry =path case=@ud]
        [%payload =simple-payload:http]
    ==  ==
```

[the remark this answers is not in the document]
lukechampine: @Fang- It was initially implemented as a `beam`: https://github.com/urbit/urbit/pull/5927/commits/0d34b861df944cf78ad65976a6fb747c02868fe0

I don't recall the exact reason we switched. Perhaps because it was too general; it wouldn't make sense to point a cache at `da+now`, for example.

lukechampine:
Updated and rebased.

@Fang-, should I be making use of the `auth` flag anywhere yet, or is that just for later?

tacryt-socryp:
@Fang- ???

Fang-:
Oh, I ended up talking to @lukechampine about the `auth` question out of band a couple days ago. Copying here for completeness:

> as for the `auth` flag, yes, you'll want to use that wherever responses get resolved from the cache: unauthenticated requests to endpoints/entries that require auth should serve a permission denied or some such.

belisarius222:
@tacryt-socryp I'll be taking over for @Fang-, since he's out for a while to handle family issues.  Let's get @lukechampine paid for his work.  We can do that even before this goes through the full release process, just by getting it into a state where it could plausibly be merged into a release.

As far as I can tell, the only showstopper there is that this has a space leak.  As soon as that's fixed, I'll do a final pass and see if we can mark this as finalized and close out the grant.

Happy to talk about this on a call if you'd like.

lukechampine:
Thanks Ted. Are you referring to what Joe said here? https://github.com/urbit/urbit/pull/5927#discussion_r986392212
Seems like the most direct solution is to add an optional `case` parameter to `+del-cache`, and punt on the question of "how does the caller figure out which `case` to specify?"

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1961,6 +2010,55 @@
...
+      `state
+    ::  no case means the most recent entry
+    ::
+    =/  case=@ud  (fall ucase (sub (lent cache-entries) 1))
+    ?:  (gte case (lent cache-entries))
+      `state
+    ::  replace the entry with ~
+    ::  and give a %cache-changed for each affected binding
+    ::
+    =.  cache-entries  (snap cache-entries case `(unit cache-entry)`~)
```
lukechampine, line 2037: feedback appreciated on this code -- it looks kinda ugly to me, so I suspect there's a more idiomatic approach

belisarius222:
@lukechampine Yes, that's the space leak I'm referring to.  I just read through this code, and most of it looks fine.  Even the parts that are a little clunky, as you pointed out, aren't that bad.  The list of units is an awkward data structure, and I suspect we'll replace that later with something like an ordered map from aeon to cache entry, but since nothing accesses anything other than the latest entry as it is, it should be fine for now.

The issue right now is the space leak, which as was discussed in that comment thread, is currently somewhat annoying to prevent: whoever is sending these cache entries to Eyre would need to store each case itself and keep that data in sync with Eyre's.  If there's anything we've learned in the past couple years of kernel dev, it's that we should avoid the need for this kind of data duplication and synchronization between vanes.

How about this: `+add-cache`, in addition to prepending the new cache entry to the list, also replaces all the other entries in the list with `~`s.  This would certainly be lame in multiple ways, but it's the smallest change I can think of that would prevent the space leak -- no change to the data structures or interface types, and just a very minor change to the implementation.

@lukechampine the `%cache-changed` gift should include the case of the newly added cache entry.  That way a runtime could eventually scry for the entry at the case it heard about, which will succeed unless there's a race condition because a new entry was added in between Vere hearing the notification and Arvo hearing the scry request for the old entry -- while it needs to be handled correctly in Vere, this is very unlikely, and it's not a problem for this phase of the grant.

If these changes are made, I'd be happy to sign off on this as a completed phase.

belisarius222:
Some non-actionable commentary, for posterity:

The `(map path (list (unit cache-entry)))` is both confusing and difficult to manipulate cleanly, which is why the code that woks with it looks "ugly" -- it's trying to do something that's harder than it should be.  I think the lameness in this design is fundamentally an unfortunate side effect of a compromise between present needs and projected future needs.

Taking a step back, as I recall from some months ago, the original idea of splitting Eyre's bindings into two data structures was to have a mutable data structure that points into an immutable data structure.  The immutable data structure could then eventually be something that other systems could scry into.  The thought was that even if that's not strictly required, it might make it easier for the Vere side of this system to stay in sync with the Arvo side, since even if there's some race condition that causes it to be out of date, at least it won't be incorrect.  If it's scryable, Eyre could potentially also serve this data as part of a solid-state publication, which might be easier for a web client to retrieve.

Without this nod toward the future, clearly this whole thing would be a lot simpler.  Eyre could have just one piece of state for this, as a mutable binding: `(map url=@t cache-entry)`.  Could we make something that's almost this easy to manage, but without breaking the futureproofing by having a referentially transparent data structure internally?

In particular, if we're going to have an immutable data structure, Eyre needs its own concept of a numeric revision on that data structure.  This is implicit in the current design, based on the position in the list.  Making it explicit is likely to make it easier to manage.  One idea would be to replace the `(list (unit cache-entry))` with a `((mop @ud cache-entry) lte)`, where the `@ud` is the explicit Eyre case representing how many times this cache entry has been overwritten.

We won't really know what we need here until we have a runtime side of this that actually uses the API.  We can revisit these questions if we run into problems designing the runtime client for this system.

lukechampine:
> How about this: `+add-cache`, in addition to prepending the new cache entry to the list, also replaces all the other entries in the list with `~`s. This would certainly be lame in multiple ways, but it's the smallest change I can think of that would prevent the space leak -- no change to the data structures or interface types, and just a very minor change to the implementation.

I assume `+del-cache` should replace all the entries with `~` as well, then? Also, should `+add-cache` give a `%cache-changed` for the binding(s) affected by the deletion?

belisarius222:
I think the %cache-changed notification to Vere should only be sent if something changes with the latest cache entry, not historical data.  Deleting a

belisarius222:
misfire

belisarius222:
Anyway... *clears throat* Deleting historical entries shouldn't trigger a notification to Vere, I don't think.

tacryt-socryp:
Unless you're deleting an entry that was *currently* bound in the cache, which if adding a new entry deletes the previous entry, would be the case.

belisarius222:
I was thinking the purpose of the effect would be to tell Vere what the new scry path is for this URL, so that Vere knows what to scry for to replace its cache line for that URL.

When we add a new cache entry, it gets prepended to the list, at a new case, and we delete all the previous cache values.  When we do that, we should tell Vere about the new case, so Vere, in order to update its own cache, can scry for the cache entry at the case it heard from us.  I don't think we should need to emit a deletion effect and a second effect for the addition, because I don't think Vere will want to hold anything other than the latest cache entry for any given path, so all Vere needs to know is the latest case, which is achieved with just the effect with the case of the newly bound cache entry.

If Eyre deletes the latest cache entry for a path because someone sent Eyre a %del-cache command, then Eyre should notify Vere so Vere can delete its own cache entry.  Maybe that's a %cache-changed with a null case, or it could be a different effect, like %cache-deleted.

tacryt-socryp:
Sounds good!

lukechampine:
> When we do that, we should tell Vere about the new case

That would imply adding another type of card, right? Since (despite the name) `%cache-changed` pertains to the `cache-bindings`, not the `cache`. Maybe we could have:
```hoon
    ::  notification that a cache entry has changed
    ::
    [%cache-changed case=@ud =path]
    ::  notification that a cache binding has changed
    ::
    [%cache-binding-changed url=@t =path]
```
Also... if we're only going to keep the most recent entry, couldn't we ditch the `list` entirely and only track the current `case=@ud`?

belisarius222:
> Also... if we're only going to keep the most recent entry, couldn't we ditch the list entirely and just track the current case=@ud?

Yes.  While we're at it, I think we can simplify this even further and make it even smaller and easier to think about.

The reason to do anything immutable at all in Eyre for this feature is to make things easier for the runtime to invalidate its cache once it has one.  Expressing requests in referentially transparent terms often helps with this.  So Eyre should notify Vere with a scry path, which Vere can then use to call Arvo's `+peek` arm if it wants the data.  Ideally the `$case` in the scry path Vere uses is numeric, not a date, so it's effectively discrete, not continuous, allowing Vere to deduplicate notifications at the same revision number.

Some of Eyre's responses for cached mutable URLs are direct HTTP responses; others are scry paths (into Clay, for now).  Eyre should abstract over this difference for Vere's benefit.  Vere can't use the Clay scry path directly, even if it has one, since it needs the response to go through Eyre for it to mark-convert the Clay file to an HTTP response.

This all suggests that Eyre needs to maintain its own cone of the scry namespace for mutable URLs.  It will use its own state to resolve requests into this namespace.  If the value for an incoming scry path is a concrete response, it can return that from its state directly; if it's a scry path into Clay or some other vane, Eyre will perform the scry into that other vane, convert the result to an HTTP response, and return that.

Each binding from URL to Eyre scry path needs a numeric revision number.  Eyre could store a single piece of state for all these bindings:
```hoon
::  ...
+$  cache-entry
    $:  auth=?
    $=  body
    $%  [%scry =path]  ::  note I removed a `case=@ud` from this
        [%payload =simple-payload:http]
    ==  ==
::  ...
cache=(map url=@t [aeon=@ud val=cache-entry])
```

The `aeon` in each value in the map is the revision number that Vere will use to scry the value out of Eyre.  Every time Eyre overwrites an entry in this map, it needs to increment this to maintain referential transparency, and it will notify Vere with an effect that looks like this:
```hoon
[%grow =path]
```
The `path` is a fully qualified scry path.  Vere, if it receives such an effect, should check if it already has a cache entry for paths that are identical except for the case.  If the new case is higher, it should replace its cache entry with the new one.

Any `@t` can be encoded in a path:
```hoon
> `path`/eyre/x/3/(scot %t '/foo/bar#id?baz=qux&zip=zap')
/eyre/x/3/~~~2f.foo~2f.bar~23.id~3f.baz~3d.qux~26.zip~3d.zap
```
Eyre will construct this scry path to send to Vere by encoding the mapped URL into the scry path like this.

If an app asks Eyre to delete a URL mapping, Eyre will overwrite the entry for that URL with a simple HTTP response containing an error -- either a 403 or 404, I'm not sure what's best here.  That way Vere will be notified using the same mechanism, and it will be straightforward for an app to get Vere to stop serving something (although there's no guarantee it won't be in Vere's scry cache for the old aeon).

Note that this means the requests to Eyre can also be collapsed into just this:
```
$%  [%set-response url=@t response=(unit cache-entry)]
::  ...
```
or alternatively,
```
$%  [%set-response url=@t =cache-entry]
    [%del-response url=@t]
::  ...
```

lukechampine:
Two questions:

> ```hoon
>    $%  [%scry =path]  ::  note I removed a `case=@ud` from this
> ```

How do I scry from Clay without a `case`?

> The aeon in each value in the map is the revision number that Vere will use to scry the value out of Eyre. 

So would I be adding code to Eyre's `+scry` arm to serve those scries? And would that require a separate `path -> cache-entry` mapping, like the `axal` we have now?

belisarius222:
That path would include the case.

Adding a scry arm to Eyre is not a bad idea, to make sure this plan will work. You shouldn't need a separate mapping, since the scry path will contain an encoded URL, which Eyre can decode to use as the key to look up the cache entry in its state. Eyre would then check that the aeon matches the value for that key before delivering the scry result.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

lukechampine:
ok, I implemented the new approach. I opted for `(unit cache-entry)` rather than storing a 404 payload, and I added the scry. Not sure I got the scry path format right, so pls check. Overall I like how much simpler this is -- almost 50% fewer lines than the previous approach. :)

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1998,6 +2044,17 @@
...
       ==
     --
+  ::  +set-response: remember (or update) a cache mapping
+  ::
+  ++  set-response
+    |=  [url=@t entry=(unit cache-entry)]
+    ^-  [(list move) server-state]
+    =/  aeon
+      =/  cache-val  (~(get by cache.state) url)
+      ?~(cache-val 0 +(aeon.u.cache-val))
```
belisarius222, line 2054: Let's start the aeon at 1, so that aeon 0 can mean there's no mapping yet.  This might be useful if, for example, someone scrys for the latest aeon on a path later.  Clay starts at 1 for this reason.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2585,7 +2654,15 @@
...
         ?=(^ secure.http-config.server-state.old)
     ==
-  ..^$(ax old)
+  ?-  -.old
+      %~2022.12.3  ..^$(ax old)
+  ::
+      %~2020.10.18
+    %_  $
+      date.old          %~2022.12.3
+      server-state.old  [-.server-state.old *(map url=@t [aeon=@ud val=(unit cache-entry)]) +.server-state.old]
```
belisarius222, line 2663: this looks past 80 characters

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2645,6 +2722,13 @@
       %-  =<  request-is-logged-in:authentication
           (per-server-event [eny *duct now rof] server-state.ax)
       %*(. *request:http header-list ['cookie' u.cookies]~)
+    ::
+        [%cache @ @ ~]
+      ?~  aeon=(slaw %u i.t.tyl)  [~ ~]
```
belisarius222, line 2727: I'd align the `[~ ~]` on each of these lines for beautification.

belisarius222:
> 50% fewer lines

That's what I like to hear.  Nice work.

I reviewed the code.  It looks good to me, other than that one comment I left about changing aeons to start at 1, rather than 0, and a couple minor style nits.  At some point I'd like to bikeshed about naming (this code refers to itself as a "cache", but it's not really a cache -- it stores a semantically meaningful mapping from URL to predetermined response), but let's defer that until we have a runtime client for this machinery, since there's a good chance we'll need to redo some of this code to support the runtime anyway.

@lukechampine Once you address this last round of comments, let's move on to the next phase of the grant.

lukechampine:
done.
Given all the churn in these commits, want me to squash them?

belisarius222:
Yeah, I think that makes sense in this case.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

lukechampine:
aight, squashed 👍🏻 

> At some point I'd like to bikeshed about naming (this code refers to itself as a "cache", but it's not really a cache -- it stores a semantically meaningful mapping from URL to predetermined response)

yeah, I noticed that Eyre already has a notion of "bindings," and this does feel more like a binding than a cache; maybe down the road, they could be unified (e.g. by adding a `%response` action)?

lukechampine:
Non-blocking, but I woke up this morning with a thought about referential transparency: can we prevent space leaks by using non-sequential aeons?

Consider what happens when you want to delete an entry. You can set the entry to `~`, but if you're using sequential aeons, you *can't* remove it from the map entirely, because doing so risks breaking referential transparency: if someone re-adds an entry for that url, it will start at `aeon=1` again, so you'll end up with the same scry path bound to different data. Thus, you have to remember every entry's aeon forever.

We can solve this by maintaining a single, shared aeon, rather than one per entry. This is, you increment this aeon whenever *any* entry is added/changed. Every entry still gets a unique aeon, but now we can safely delete entries too. The downside is that the aeon itself becomes less useful: it ceases to be a proper revision number. We still have monotonicity, but you can no longer make assumptions like "this entry's aeon is 6, so it has changed 5 times." Do we *need* that sort of property, though? For a cache like this, where only the current revision is accessible anyway, what is the benefit of sequential per-entry aeons?

belisarius222:
The benefit of per-entry aeons is that each time an entry changes, the runtime could scry for just the changed entry, without invalidating every cache line. That's unlikely to be a problem now, but if a ship is hosting many megabytes across many different files, Vere might need to send all of those files over the IPC connection from Mars to Urth whenever any one file changes -- the amount of invalidated data would grow like O(n) instead of O(1).

That could be avoided by building another layer that compares hashes for each file, or
something like that. I'm not sure it's worth building all this, since the space leak of holding on to deleted paths and their latest aeons is pretty small (and a common price to pay for referential transparency in multiple parts of the system), but you make a good point.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

jalehman:
@lukechampine This can be considered complete for the sake of the grant's initial milestone, but not viable to merge and release until the second milestone involving a Vere implementation is complete. Go ahead and [invoice us](https://airtable.com/shrXXCs1uaxtNSBcg) for the first milestone when you're ready. 

cc @belisarius222 @tacryt-socryp

lukechampine:
Got it, thanks!

lukechampine:
Per discussion with @belisarius222, I've updated the `%cache` scry to always return payloads, never scry paths. This spares vere from having to perform an extra scry when updating its own cache.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2645,6 +2740,39 @@
...
+            :~  ['content-type' 'text/html']
+                ['content-length' (crip (format-ud-as-integer p.data))]
+            ==
+        data=[~ data]
+      =/  cache-entry
+        :*
+          auth=auth.val
+          body=[%payload simple-payload]
+        ==
+      ``noun+!>(val)
```
lukechampine, line 2775: the scry still returns a `cache-entry`, but the body will always be a `%payload`. That feels a little off to me, but it didn't seem worth it to define a separate `cache-scry-result` type. Thoughts?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2827,7 +2827,33 @@
...
+      ::  body is a scry; scry it to get a payload
+      ::
+      ?~  path.body.val                  [~ ~]
+      ?~  bem=(de-beam t.path.body.val)  [~ ~]
+      =/  res=(unit (unit cage))  (rof ~ i.path.body.val u.bem)
+      ?~  res    [~ ~]
+      ?~  u.res  [~ ~]
+      =*  vase   q.u.u.res
+      ?.  ?=(@ q.vase)  [~ ~]
+      =/  data  (as-octs:mimes:html q.vase)
```
belisarius222, line 2843: Eyre should mark-convert this to `%mime` instead of asserting it's an atom and then calling `+as-octs` manually.  Clay files can all be converted to `%mime`.  It will need to scry a `$tube` out of Clay based on the mark found at `p.u.u.res`, scrying it out of the desk found at `q.u.bem` with a `%cc` or `%cf` scry, whichever is easier to use.  A `%cc` scry returns a vase of a `$tube` gate that takes in the vase of an input and returns a vase of a mark-converted output; `%cf` returns a vase of a statically typed gate that takes in an input of one mark and outputs a statically typed value of another mark.

I think `%cf` is better.  Code would look something like this:
```hoon
=+  .^(cast=vase %cf (en-beam -.u.bem /[p.u.u.res]/mime))
(slam cast q.u.u.res)
```

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -589,6 +592,18 @@
...
+++  scry-to-mime
+  |=  [=path rof=roof]
+  ^-  (unit mime)
+  ?~  path  ~
+  ?~  bem=(de-beam t.path)  ~
+  =/  res=(unit (unit cage))  (rof ~ i.path u.bem)
+  ?~  res    ~
+  ?~  u.res  ~
+  =/  cast  .^(vase %cf (en-beam -.u.bem /[p.u.u.res]/mime))
+  `!<(mime (slam cast q.u.u.res))
```
lukechampine, line 606: @belisarius222 how's this?

btw, I noticed that `+handle-scry` does something very similar, but it uses the `%cc` approach.

pkg/arvo/sys/vane/eyre.hoon
lukechampine: Having trouble getting this to work, because `%cf` returns a static gate, not a vase. `+sign-to-json` uses `%cf` with `rof` (rather than a raw `.^`) and that seems to work, but when I try it in `scry-to-mime` it fails with `clay: read-at-aeon fail [desk=%$ care=%f case=[%da p=~2023.3.18..16.53.39..d4f9] path=/noun/mime]`. 😕

belisarius222:
CI failure:
```
vane: %dill: ~tognyx-tapmet
vane: %eyre: ~fabbur-fabted
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
mot: %mean
pier: (0): play: bail

play: bail: %exit
home: bailing out
```

lukechampine:
sry, a typo snuck in somehow. Passing now.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2757,6 +2851,34 @@
...
+      ?~  url=(slaw %t i.t.t.tyl)        [~ ~]
+      ?~  entry=(~(get by cache) u.url)  [~ ~]
+      ?.  =(u.aeon aeon.u.entry)         [~ ~]
+      ?~  val.u.entry                    [~ ~]
+      =*  val  u.val.u.entry
+      ?:  ?=(%payload -.body.val)
+        ``noun+!>(`cache-entry`val)
+      ::  body is a scry path; scry it to get a payload
+      ::
+      =/  res=(unit mime)  (scry-to-mime rof [our %base da+now] path.body.val)
```
belisarius222, line 2866: This doesn't look right to me.  It hard-codes the `%base` desk as the location of the file to be scry'd out, but most use-cases for this will have files stored in other desks.

This suggests the representation of the scry path passed into Eyre should contain a desk and case (it should be possible to tell Eyre to serve an old version of a file too -- no reason not to afaik).  Now we're almost at a `$beam`, just missing `ship`.

So I think we should change the interface here to accept a fully qualified scry path in the `$task` sent to Eyre.  This full path can be used directly to scry a value, and then the `desk` in the path should be used to load the mark conversion to `%mime`.

lukechampine: IIUC the paths are already fully-qualified; `+scry-to-mime` calls `de-beam` on it. I've been using `.^(cors-registry:eyre %ex /=//=/cors)` to test the scry functionality, but since the desk is `%$`, scrying the mark conversion fails. On the other hand, if you try `/=base=/cors`, the Eyre scry fails. So I'm not sure what to do here.

Separately, I noticed that the `cage` returned for a (successful) `/cors` scry is `%noun`, not `%cors-registry`. Scrying a `noun -> mime` mark conversion seems to succeed, but the resulting gate behaves weirdly. Should we check explicitly for `%noun`?

belisarius222: Let's just rip this functionality out of Eyre for now.  It's not that hard for an agent to scry and mark-convert the result, then ask Eyre to use that as the HTTP response.  Halfway future-proofing by allowing the agent to also specify a scry path doesn't meaningfully improve this system, so I think we should ditch it in favor of deploying something that works sooner.

Don't lose the code, for it, though.  When it comes time to expose a scry namespace in response to HTTP requests, we'll need to think this through.  I'm starting to lean toward having this be Clay functionality: ask Clay in the scry path to mark-convert the file for you.

lukechampine: kind of a relief tbh. I kept the tagged union so that we can add a `%scry` case later, is that the right way to go?

belisarius222: Yeah that's fine.

Fang-:
It's not immediately clear to me what the `url=@t` here is expected to be. Just the `/request/path.ext`? Or maybe `domain.com/path.etc`?

In the former case, this doesn't let us assign entries to specific hostnames, even though we might be serving different things on different domains.  
In the latter case, we will probably want a wildcard, in the same way that the existing "bindings" system supports.  
We'll probably want something very similar to `$binding` here, no?

matthew-levan:
Hi @lukechampine will you please rebase this again on `next/kelvin/413`? After doing so, the CI workflows should work with the correct binary version pulled over from `urbit/vere`.

pkg/arvo/sys/lull.hoon
```diff
@@ -1373,6 +1378,9 @@
         ::    not allowed.
         ::
         [%bound accepted=? =binding]
+        ::  notification that a cache entry has changed
+        ::
+        [%grow =path]
```
belisarius222, line 1383: I think we should change `%grow` to `%grew` for this gift, and probably move it to the generic `$gift-arvo` type and then `$>` that here.  Gall will likely use this gift too, and we're adding a `%grow` note to agents so that they can grow their own paths.  `%grew` matches nicely with `%grow`, while implying it's a notification that the path was just grown.

belisarius222: That can be done after merge, to be clear.

belisarius222 (approved):
lgtm

belisarius222:
> We'll probably want something very similar to $binding here, no?

@Fang- @lukechampine Sounds right to me.  Let's switch this to `$binding` if we can, but let's do it in a separate PR, since this is working as-is.

Fang-:
Note that `$binding` just by itself might not be sufficient, since it doesn't include the file extension, which may be relevant.

PR #5928, opened 2022-08-07
vere: make `prep` and `next` commands non-interactive (no tty required)

### LINKED ISSUE #5924: "prep" command requires a tty (for no good reason) and does not accept "--no-tty"
**Describe the bug**
So it looks like Vere 1.10 has stricter error checking.  This is a good thing, but it broke my point systemd services because those services start with "{last-used-vere} prep {pier}", and "prep" now bails out because there's no tty (actually it complained about no tty in 1.9 too, but it used to ignore this and now does not).

So there are two bugs exposed by the improved error checking:
- "prep" wants to be interactive by default (why?)
- there is no way to change that default

Wrapping the "prep" part in screen works around this, so nothing urgent.

**Notify maintainers**
@joemfb

joemfb:
This PR fixes a dumb oversight in the new binary upgrade system. Thanks to @midlyx-hatrys for the report.

Fixes #5924.

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -1409,6 +1409,7 @@ _cw_next(c3_i argc, c3_c* argv[])
 
   u3_Host.pep_o = c3y;
   u3_Host.nex_o = c3y;
+  u3_Host.ops_u.tem = c3y;
 }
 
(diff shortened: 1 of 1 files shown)


mcevoypeter (commented):
LGTM

PR #5930, opened 2022-08-08
gall: cork a subscription's ames flow on-kick

Fang-:
Gall tells ames to %cork flows for subscriptions it has closed.
Receiving a kick also closes a subscription, but gall wasn't issuing a
%cork in that case. We correct that here.

Inlines +mo-handle-ames-response's logic at its only callsite.


belisarius222 (approved):
lgtm

PR #5931, opened 2022-08-09
gall: handle nonces in wires correctly

Fang-:
Gall was not always including the subscription nonce in the wire, when it should have been. Most prominently, when `%leave`ing a subscription due to a crash.

Additionally, we were not initializing the sub-nonce correctly during the upgrade case. This affected ships who have been running pre-release until now, but the artifacts this issue has left behind will get cleaned up by new-and-improved `|doff`.


belisarius222 (approved):
lgtm

PR #5932, opened 2022-08-09
ames: fix "ack crash"

belisarius222:
Stop crashing on benign acks on corked bones.

Also, +on-plea gets called in two very different ways:
1) handling request from local vane to send %plea to peer
2) handling %cork request from another ship, which our local ames has %pass'ed
   to ourselves

In the second case, we shouldn't print misleadingly, or bind a duct in the ossuary.


Fang- (approved):
Looks good to me. As with the other ones, we worked on this together, both understand the problem, and tested this on a live ship. Onward!

PR #5934, opened 2022-08-10
gall: kill old subscriptions on %doff

belisarius222:
`%doff` now kills live subscriptions and any old subscriptions that might be lingering due to prerelease bugs.  We had to modify Ames to ignore `%leave` messages sent on closing or corked bones during a `%doff`.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -432,12 +432,15 @@
   ::  +mo-doff: kill all outgoing subscriptions
   ::
   ++  mo-doff
+    |=  [dude=(unit dude) ship=(unit ship)]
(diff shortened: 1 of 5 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1776,7 +1776,7 @@
       |-  ^+  ap-core
       ?~  subs  ap-core
       =+  [wyr dok]=i.subs
-      ?.  |(?=(~ ship) =(u.ship ship.dok))
```
joemfb: *allocates in `+ap-doff` conditional ;)

Fang- (approved):
Looks good now, let's keep it going.

PR #5935, opened 2022-08-10
Lagoon hoon

sigilante:
This resolves #4455 and #4908 by moving them into `hoon.hoon` and including tests.

It does not, as of this message, include pretty-printing or aura detection yet.

It creates a new section in `hoon.hoon`, `5g: vector and matrix array support`.

We currently have `@lvs` (`@rs` vectors), `@lvd` (`@rd` vectors), and `@lms` (`@rs` matrices).  Obviously at some point we'll need to consider filling these out, but it'll be driven by need to some extent:  hardcore numerics is not in the purview of Urbit.


sigilante:
- [ ] add characteristic unique parsable aura format & parser
- [ ] improve pretty-printing

sigilante:
On characteristic aura syntax:

The atom format has the leftmost byte containing the dimensionality (`1` for `@lv`, `2` for `@lm`), then for `@lm` the number of rows in the next byte.  The remainder are LSB-ordered within the array, i.e. the rightmost entry is the first entry in the array.

We need a way of inputting the values so they are uniquely recognized:

**`@lvs`**:

```
~[.1 .2 .3 .4 .5]
0x1.40a0.0000.4080.0000.4040.0000.4000.0000.3f80.0000
.1._2._3._4._5._
```

```
~[.1]
0x1.3f80.0000
.1._
```

```
`@lvs`~
0x1
.._
```

**`@lvd`**:

```
~[.~1 .~2 .~3 .~4 .~5]
0x1.4014.0000.0000.0000.4010.0000.0000.0000.4008.0000.0000.0000.4000.0000.0000.0000.3ff0.0000.0000.0000
.~1._~2._~3._~4._~5._
```

```
~[.1]
0x1.3ff00.0000.0000.0000
.~1._
```

```
`@lvd`~
0x1
.~_
```

**`@lms`**:

```
~[~[.1 .2 .3] ~[.4 .5 .6]]
0x2.0000.0002.3f80.0000.4000.0000.4040.0000.4080.0000.40a0.0000.40c0.0000
.1_2_3__4_5_6
```

```
~[~[.1 .2] ~[.3 .4] ~[.5 .6]]
0x2.0000.0003.3f80.0000.4000.0000.4040.0000.4080.0000.40a0.0000.40c0.0000
.1_2__3_4__5_6
```

```
~[~[.1]]
0x2.0000.0001.3f80.0000
.1.__
```

```
`@lms`~[~]
0x2.0000.0001
.__
```

@joemfb @belisarius222 please weigh in on this proposed atom syntax—they don't obviously (to me) run afoul of any other atom literal syntax as they stand, and I don't think they'll collide e.g. with `@if`.

The pretty-printer doesn't have to match this, of course, and some don't.

ohAitch:
I certainly vote there not be any unprintable(or unparsable) valid `@l` values :\)
(Relatedly, do we have nans? negative zero? *probably* reusing all the individual float parsers Just Works here)

ohAitch:
> ```
> `@lvs`~
> 0x1
> ._
> ```
This seems not ideal given tuple literal form(e.g. `._1_2_3__`), do we have any other prefix ambiguities?

sigilante:
Well, my idea was to suffix the value with either `_` cab for vectors or `__` cabcab for matrices.

NaNs are unified in the jets (like `@r`) so they should behave normally.

sigilante:
I added a special case for `~` as a `0x0`.

`._` and `.__` are similar to tuple literal form, but don't actually collide.

tacryt-socryp:
Note: since I haven't said it explicitly, I have a business use case for matrix math. :)

sigilante:
Current status:  I have a pretty-printer which works.  The current version distinguishes `hqds` types by the vector angle brackets.

The atom literal parser is being cantankerous about fallthrough.

Let's suppose I have a parser which is supposed to ingest literals of the form

```hoon
`@lvs`..1_.2_.3.4_`     :: <.1 .2 .3.4>
`@lvd`..~1_.~2_.~3.4_`  :: «.~1 .~2 .~3.4»
```

Since the determination of type is done by the embedded value (similar to the
compound tuple entry but not identical), I need the parser to distinguish
between these based on the internal values.

```hoon
|%
++  lago
  ~+
  ;~  pose
    (stag %lvd lago-lvd)
    (stag %lvs lago-lvs)
  ==
::
++  lago-lvd  (cook lagd ;~(pfix dot lago-base-d))
++  lago-lvs  (cook lags ;~(pfix dot lago-base-s))
::
++  lago-base-d
  |=  b=nail
  ((star ;~(plug (just '.') royl-rd:so (just '_'))) b)
++  lago-base-s
  |=  b=nail
  ((star ;~(plug (just '.') royl-rs:so (just '_'))) b)
++  lagd
  |=  a=(list [@t @rd @t])
  ^-  @lvd
  =/  b  (turn a |=(a=[@t @rd @t] +<:a))
  (mix (rep [6 1] b) (lsh [6 (lent b)] 1))
++  lags
  |=  a=(list [@t @rs @t])
  ^-  @lvs
  =/  b  (turn a |=(a=[@t @rs @t] +<:a))
  (mix (rep [5 1] b) (lsh [5 (lent b)] 1))
--
```

and the following should work, for instance:

```hoon
> ((cook lags ;~(pfix dot lago-base-s)) [[1 1] "..5.4_.2.3_"])
[p=[p=1 q=12] q=[~ u=[p=<.5.4 .2.3> q=[p=[p=1 q=12] q=""]]]]
```

But the current code fails to distinguish, in `++lago`, between types of
vectors.  (`tt` is the Dojo pin for the core above.)

```hoon
> (lago:tt [[1 1] "..1_.2_.3_.4_"])
[p=[p=1 q=3] q=[~ u=[p=[%lvd «»] q=[p=[p=1 q=2] q=[i='.' t="1_.2_.3_.4_"]]]]]

> (lago:tt [[1 1] "..~1_.~2_.~3_.~4_"])
[p=[p=1 q=18] q=[~ u=[p=[%lvd «.~1 .~2 .~3 .~4»] q=[p=[p=1 q=18] q=~]]]]
```

How can I get the fallthrough to the next case to work in case of failure?

sigilante:
Hmm, I wonder if I can trick it by `pfix dot sig` then `pose` `royl-rs` and `royl-rd`.

sigilante:
Trying to figure this out (the last comment didn't get anywhere useful):  what makes `++pose` actually decide that a parsing attempt has failed?

sigilante:
```hoon
|%
++  lago
  ~+
  ;~  pfix  dot
    ;~  pose
      (stag %lvs lago-lvs)
      (stag %lvd lago-lvd)
    ==
  ==
::
++  lago-lvd  (cook lagd lago-base-d)
++  lago-lvs  (cook lags lago-base-s)
::
++  lago-base-d
  |=  b=nail
  ((star ;~(plug (just '.') royl-rd:so (just '_'))) b)
++  lago-base-s
  |=  b=nail
  ((star ;~(plug (just '.') royl-rs:so (just '_'))) b)
++  lagd
  |=  a=(list [@t @rd @t])
  ^-  @lvd
  =/  b  (turn a |=(a=[@t @rd @t] +<:a))
  (mix (rep [6 1] b) (lsh [6 (lent b)] 1))
++  lags
  |=  a=(list [@t @rs @t])
  ^-  @lvs
  =/  b  (turn a |=(a=[@t @rs @t] +<:a))
  (mix (rep [5 1] b) (lsh [5 (lent b)] 1))
--
```

the above with pose:

    > (scan "..1_.2.3_" lago:tt)
    [%lvs 0x1401333333f800000]
    > (scan "..~1_.~2.3_" lago:tt)
    {1 3}
    syntax error

the above with simu:

    > (scan "..1_.2.3_" lago:tt)
    {1 3}
    syntax error
    > (scan "..~1_.~2.3_" lago:tt)
    [%lvd 0x140026666666666663ff0000000000000]

the above with plug:

    > (scan "..1_.2.3_" lago:tt)
    [[%lvs 0x1401333333f800000] %lvd 0x1]
    > (scan "..~1_.~2.3_" lago:tt)
    [[%lvs 0x1] %lvd 0x140026666666666663ff0000000000000]

sigilante:
Ah!  But this seems to work:

```hoon
|%
++  lago
  ~+
  ;~  pfix  dot
    ;~  pose
      (stag %lvs lago-lvs)
      (stag %lvd lago-lvd)
    ==
  ==
::
++  lago-lvd  (cook lagd lago-base-d)
++  lago-lvs  (cook lags lago-base-s)
::
++  lago-base-d
  |=  b=nail
  (;~(sfix (most cab ;~(pfix dot royl-rd:so)) cab) b)
++  lago-base-s
  |=  b=nail
  (;~(sfix (most cab ;~(pfix dot royl-rs:so)) cab) b)
++  lagd
  |=  a=(list @rd)
  ^-  @lvd
  (mix (rep [6 1] a) (lsh [6 (lent a)] 1))
++  lags
  |=  a=(list @rs)
  ^-  @lvs
  (mix (rep [5 1] a) (lsh [5 (lent a)] 1))
--
```

sigilante:
What remains:

- [ ] special case parsing
- [ ] `@lms` parsing

belisarius222:
> Working!
> 
> ```
> > ..~1.3_.~2.4_
> «.~1.3 .~2.4»
> 
> > ..1.3_.2.4_
> <.1.3 .2.4>
> 
> > ..~~1.3_.~~2.4_
> ≺.~~1.3 .~~2.4≻
> 
> > ..~~~1.3_.~~~2.4_
> ⋘.~~~1.3 .~~~2.4⋙
> ```

These look like they're not ASCII.  Hoon literals need to all be ASCII.

sigilante:
Question:  shouldn't auras _not_ nest by default?

```
> `@lvs`..1.2_.3.4_
<.1.2 .3.4>

> `@lvd`..1.2_.3.4_
«.~102.40003957748414»
```

(that latter one should be impossible)

sigilante:
> These look like they're not ASCII. Hoon literals need to all be ASCII.

That's an output convention I'm using for the pretty-printer, easy to revert to `<<` or similar.

sigilante:
I'm adding in `@lmd` as well for completeness, then a cleanup pass.

Fang-:
> That's an output convention I'm using for the pretty-printer

tbc, in most cases (and I believe _always_ in the case of atoms) the pretty-printer uses valid hoon syntax to render the values. Just pretty-printing as `..~1.3_.~2.4_` etc should be fine.

sigilante:
That's fine.  I'll leave some kind of export wrapper to `tape`, maybe to CSV or similar then.

sigilante:
Status update:

- I've changed the pretty-printer format to just match the atom syntax.
- I've written the parser for @lmd type.  All pieces work separately, and the whole works, but when I include the `(stag %lmd...)` clause in `++lago` then Hoon won't compile.

```
> (lago-lmd:so [[1 1] ".~1_.~2__.~3_.~4__"])
[p=[p=1 q=19] q=[~ u=[p=..~1_.~2__.~3_.~4__ q=[p=[p=1 q=19] q=""]]]]

> (lago-base-md:so [[1 1] ".~1_.~2__.~3_.~4__"])
[p=[p=1 q=19] q=[~ u=[p=~[~[.~1 .~2] ~[.~3 .~4]] q=[p=[p=1 q=19] q=~]]]]

> |commit %base
>=
arvo: beginning upgrade
arvo: compiling hoon
ride: parsing
ride: compiling
ride: compiled
ride: parsing
ride: compiling
ride: compiled
arvo: compiling next arvo
ride: parsing
crud: %into event failed
[%poke %into]
arvo: call:pith failed
arvo: upgrade failed
syntax error
{1 7}
```

So I'm going to keep noodling on this one and see...

sigilante:
So my guess right now is that there is some sort of collision with the @lmd format that breaks compilation.  `!:` doesn't narrow it down, it's just a flat `[7 1] Syntax error`.

sigilante:
~master-morzod said that it's breaking on parsing `..ride` so that's the next rabbit hole for this... closer.

I may have to revisit the atom literal syntax for matrices if this persists.

sigilante:
https://gist.github.com/sigilante/5c9f38efb52fb256649059158a98c2f3

Update here.  @joemfb @tacryt-socryp

I had a good chat with @joemfb this morning. We have three options now:

1 .Userspace-first. Ship it in /lib. Don't worry about direct atom handling, provide gates to ingest and display prettyprinted version.
2. `zuse.hoon`. Take most of what I have in hoon.hoon, move it to zuse.hoon, and slap a couple of I/O functions across it. A %zuse decrement is fairly straightforward, and userspace will be able to use @lv/@lm types directly.
3. `hoon.hoon`. What I was working on. The trouble is that this gets into politics and metapolitics of Hoon language design, perhaps prematurely, and gets bundled indefinitely for the next Hoon version decrement. Another problem is that the atom literal syntax parser has a subtle collision that breaks compilation (even tho it all works in userspace).

Option three is basically done but untenable to finish and merge. Option two is probably the best because it will allow jetting. It will not, however, have direct atom literal syntax handling.

joemfb:
I endorse option 2 -- putting in %zuse seems like the best combination of supporting reasonably fast iteration while keeping open the option to ship jets.

sigilante:
Current status:  I have moved things over to `%zuse`.  I have upgraded the unit tests.

I would still like to move some parsing/printing functions over into `++la`, and possibly introduce `@lmd` which I never got around to before.  But it's very close now.

sigilante:
@joemfb gonna call this one ready for serious consideration to be complete now.

Jbollenbacher:
First off, great work on this. This is a critical, foundational piece of Hoon, and it will enable a whole new class of apps on urbit. 

I've got a (perhaps naive) question.  On lines 6066 and 6111 of zuse.hoon, you note that you're indexing from 1 rather than 0.  Why?   Is that a design choice or are there technical reasons which make it necessary? 

In my experience, 0-indexing in numerical libraries is generally regarded as preferable by most people who do a lot of numerical work.  For instance, Matlab chose 1-indexing (to make it more laymen-friendly), but it's been widely regarded by proficient users as a poor choice ever since because it often introduces off-by-one errors that need to be corrected with an additional line of code.  I think we should consider sticking to the 0-indexing norm, and only violate it if we have a very compelling, specific reason. 

Also, more generally, I think it's important to highlight that this library is likely going to serve as the foundation of all data-driven software in the future of Urbit. Given that, it's really important to ensure the interface is well-considered, especially for core features that are near-impossible to change later (like indexing).  I'd really love to see the interface reviewed publicly by a few people who regularly do a lot of numerical work.  More eyes will result in a better end product, and it matters a great deal with something this foundational.

sigilante:
> I've got a (perhaps naive) question. On lines 6066 and 6111 of zuse.hoon, you note that you're indexing from 1 rather than 0. Why? Is that a design choice or are there technical reasons which make it necessary?

Aesthetically I prefer the FORTRAN/mathematical school of numerics to the C.

> I'd really love to see the interface reviewed publicly by a few people who regularly do a lot of numerical work.

Good idea.  John Cook or Mike Croucher (Walking Randomly) I believe have both reviewed interfaces for projects or at least have the background.  I'll think of others.

jalehman:
@belisarius222 is going to talk to @joemfb about ways to include jets as a library. This code is dependent on the existence of jets, so that question needs to be resolved.

pkg/arvo/sys/zuse.hoon
```diff
@@ -6084,4 +6084,1498 @@
...
+      |=  n=@ud  ^-  @lvs
+      ~_  leaf+"lagoon-fail"
+      (fill n .1)
+    ::
+    ::    Length of vector
+    ++  length
+      |=  u=@lvs  ^-  @ud
+      ~_  leaf+"lagoon-fail"
+      =/  ell  (met 5 u)
+      ?:  (gth ell 1)  (dec ell)  0
```
belisarius222, line 6122: Can't this always be `(dec ell)`?

sigilante: `` `@lvd`0x0`` is permitted.

sigilante: We can revisit that when we switch to the cell format.

sigilante:
Deferring for rewrite in cell-based format (head tag of metadata, tail atom of array).

PR #5936, opened 2022-08-11
gall: use a fully-formed wire when sending %cork

Fang-:
In response to clog notification from remote ames, we were sending a
%cork to clean up the flow. However, the wire we were using had the /sys
prefix already stripped off. Here, we put it back in.


belisarius222 (approved):
lgtm

PR #5937, opened 2022-08-11
ames: fix comet test

belisarius222:
The test turned out to be broken because we were injecting the wrong rift number in a mocked move.  Fixing that caused the test to pass.


Fang- (approved):
Rubber stamp lgtm.

PR #5938, opened 2022-08-11
gall: scry for nonces

belisarius222:
Adds a `%f` scry to Gall to retrieve the sub-nonce for each agent.


Fang- (approved):
It's Good Enough™.

(We really should clean-ify the gall-internal scry interface at some point, all these strange cares are quite bad imo...)

PR #5939, opened 2022-08-11
hoon: make (slav %tas '') work

liam-fitzgerald:
Updates `(slav %tas '')` to produce `%$`

## Background

`((sane %tas) '') -> %.y` which suggests that this function should work. However `%$` parsing is special cased in a variety of places, so I'm unsure the specifics of how we handle this. It's worth noting that in JSON parsers, you typically to use `(se:dejs:format %tas)` to parse out `@tas`, and the current broken behaviour prevents `%$` from being used correctly.

Updates both the jet and the implementation, so this likely should go out at the next kelvin burn, if approved.


pkg/arvo/sys/hoon.hoon
```diff
@@ -5940,7 +5940,7 @@
     (rush txt ;~(pfix ;~(plug sig dot) urs:ab))
   ::
       %tas
-    (rush txt sym)
+    (rush txt ;~(pose (jest '') sym))
```
joemfb, line 5943: Only changing this here creates a mismatch between `+slaw` and `+slay`.

The comment in `+de-beam` (from @Fang-?) suggests updating `+sym` with something like `;~(pose low (easy %$))`. But I'm not sure if that's too dramatic a change.

joemfb (commented):
I think this kind of change properly needs to wait for a hoon kelvin -- the jet and hoon are not updated atomically, so it necessarily creates a jet mismatch without outside coordination.

Do you want to use `%$` as an empty-string object key in JSON?

joemfb:
@liam-fitzgerald I merged master back into `next/vere`. If you push a change to this branch the extra commits will disappear.

pkg/arvo/sys/hoon.hoon
Fang-: The comment is yours. (;

That would certainly be a very dramatic change, `sym` is used in parsers all over the place and has so far generally been understood to mean "an actual symbol". Allowing the empty string for the empty symbol would be a very breaking change. Adding a `+sim` might be preferable.

joemfb: lol

zalberico:
@joemfb if this is ready we can plan it for release early in the new year alongside a kelvin bump.

drbeefsupreme:
@liam-fitzgerald Joe is out for at least a couple weeks and we'll be burning a Hoon kelvin for the doccords release, which is slated to be cut 2023.1.17. Is this ready? If not, can we find another reviewer besides Joe to get this to the finish line?

@zalberico

zalberico:
Based on the comments it doesn't look like there's consensus to release this or that the current implementation is desirable.

@philipcmonk - defer to your judgement here if it's worth pairing this with the doccords v0, if yes we should set up a CR to make sure it's ready, if no we should close it.

jalehman:
@philipcmonk @joemfb I'd like to get another review on this for whenever we're ready to do the next kelvin burn, that way we can batch it with the next major release. Can one of you take another look at this soon to determine whether or not this is ready or in need of further work?

joemfb:
This is not ready and in need of further work, see https://github.com/urbit/urbit/pull/5939#discussion_r943880778.

PR #5940, opened 2022-08-11
u3: memmap snapshot

mcevoypeter:
`mmap()` the north (heap) segment of the snapshot so that the kernel can swap pages out more efficiently (without using the swap file). With this change, if a page needs to be swapped out due to memory pressure, if that page is clean, then the kernel doesn't need to write it disk because the page is already file-backed. Compare this to the previous situation where the kernel would have to write the page out to the swap file before evicting it from memory.

This feature is enabled by default but can be disabled by passing `-m`/`--no-mmap` at the command line.

## Test Plan
Clean launch and exit has been achieved (on Linux) for both of the following sequences:

```console
$ urbit -F wes
$ urbit --no-mmap wes
$ urbit wes
```

```console
$ urbit --no-mmap -F wes
$ urbit wes
$ urbit --no-mmap wes
```

## `sudo pmap` Output
The following command can be used to examine the worker process memory map (on Linux):
```console
$ sudo pmap $(ps aux | grep 'urbit-worker work' | grep -v 'grep' | awk '{ print $1 }')
```
Note that majority of the mapped pages are backed by `north.bin` when running without `-m`/`--no-mmap` whereas the majority of the mapped pages are anonymous when running with `-m`/`--no-mmap`.

## Additional Notes
- The Windows build is failing because the base address of the loom cannot be `mmap`d for some yet unknown reason.
- `u3_lord_init()` produced a stack buffer overflow when writing to the `arg_c` array on Windows, which was declared as `c3_c arg_c[8]` despite the fact that the terminating `NULL` of the args list was written to `arg_c[8]` . This issue has been resolved alongside the addition of the `-m`/`--no-mmap` flag.

### DIFF
--- a/pkg/urbit/include/noun/events.h
+++ b/pkg/urbit/include/noun/events.h
@@ -94,20 +94,22 @@ u3e_copy(const c3_c* const dir_c);
 //! Load a snapshot copy from a directory into memory.
 //!
 //! @param[in] dir_c  Must contain a valid snapshot.
+//! @param[in] map_o  Don't `mmap()` the snapshot if `c3y`.
 //!
 //! @return c3n  Existing snapshot loaded.
 c3_o
-u3e_load(const c3_c* dir_c);
+u3e_load(const c3_c* dir_c, const c3_o map_o);
 
 //! Start the incremental snapshot system.
 //!
 //! @param[in] dir_c  Directory to run the incremental snapshot system in. Will
 //!                   be created if it doesn't exist.
+//! @param[in] map_o  Don't `mmap()` the snapshot if `c3y`.
 //!
 //! @return c3y  No existing snapshot.
 //! @return c3n  Existing snapshot loaded.
 c3_o
-u3e_live(const c3_c* dir_c);
+u3e_live(const c3_c* dir_c, const c3_o map_o);
 
 //! Disable dirty page tracking allowing read/write of entire loom.
 //!
--- a/pkg/urbit/include/noun/events.h
+++ b/pkg/urbit/include/noun/events.h
@@ -94,22 +94,20 @@ u3e_copy(const c3_c* const dir_c);
 //! Load a snapshot copy from a directory into memory.
 //!
 //! @param[in] dir_c  Must contain a valid snapshot.
-//! @param[in] map_o  Don't `mmap()` the snapshot if `c3y`.
 //!
 //! @return c3n  Existing snapshot loaded.
 c3_o
-u3e_load(const c3_c* dir_c, const c3_o map_o);
+u3e_load(const c3_c* dir_c);
 
 //! Start the incremental snapshot system.
 //!
 //! @param[in] dir_c  Directory to run the incremental snapshot system in. Will
 //!                   be created if it doesn't exist.
-//! @param[in] map_o  Don't `mmap()` the snapshot if `c3y`.
 //!
 //! @return c3y  No existing snapshot.
 //! @return c3n  Existing snapshot loaded.
 c3_o
-u3e_live(const c3_c* dir_c, const c3_o map_o);
+u3e_live(const c3_c* dir_c);
 
 //! Disable dirty page tracking allowing read/write of entire loom.
 //!
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -714,18 +714,6 @@ _ce_patch_apply(u3_ce_patch* pat_u)
 {
   c3_w i_w;
 
-  // resize images
-  //
-  {
-    _ce_image_resize(&u3P.nor_u, pat_u->con_u->nor_w, (c3_y*)u3_Loom);
-
-    // XXX: if the south image grows, we're in trouble because
-    // _ce_image_resize() assumes that an image grows upwards in memory, not
-    // downwards.
-    c3_y* bas_y = ((c3_y*)u3_Loom + u3a_bytes) - (u3P.sou_u.pgs_w * pag_siz_i);
-    _ce_image_resize(&u3P.sou_u, pat_u->con_u->sou_w, bas_y);
-  }
-
   // seek to begining of patch and images
   //
   if ( (-1 == lseek(pat_u->mem_i, 0, SEEK_SET))
@@ -760,6 +748,19 @@ _ce_patch_apply(u3_ce_patch* pat_u)
     u3l_log("apply: %d, %x\n", pag_w, u3r_mug_words(mem_w, pag_wiz_i));
 #endif
   }
+
+  // resize images
+  //
+  {
+    _ce_image_resize(&u3P.nor_u, pat_u->con_u->nor_w, (c3_y*)u3_Loom);
+
+    // XXX: if the south image grows, we're in trouble because
+    // _ce_image_resize() assumes that an image grows upwards in memory, not
+    // downwards.
+    c3_y* bas_y = ((c3_y*)u3_Loom + u3a_bytes) - (u3P.sou_u.pgs_w * pag_siz_i);
+    _ce_image_resize(&u3P.sou_u, pat_u->con_u->sou_w, bas_y);
+  }
+
 }
 
 //! Apply north and south images to memory.
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -772,6 +772,16 @@ _ce_patch_apply(u3_ce_patch* pat_u)
 {
   c3_w i_w;
 
+  // resize images
+  //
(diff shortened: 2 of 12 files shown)


pkg/urbit/noun/events.c
```diff
@@ -704,32 +818,85 @@ _ce_patch_apply(u3_ce_patch* pat_u)
...
-    c3_w pag_w = u3a_outa(ptr_w) >> u3a_page;
-    c3_w blk_w = pag_w >> 5;
-    c3_w bit_w = pag_w & 31;
-    u3P.dit_w[blk_w] &= ~(1 << bit_w);
-
-    ptr_w += stp_ws;
   }
+#undef mark_page_clean
+#undef read_image
+#undef map_image
```
joemfb, line 899: The macros are mostly overkill here: `map_image()` is only performed once, so it doesn't need to be extracted. `read_image()` is a big, expensive i/o operation, so there's no advantage to it being a macro (and the function should use the same argument names and types for clarity). `mark_page_clean()` is a fine macro (although setting up the macro probably uses as many lines as just repeating the bitmap logic).

pkg/urbit/noun/events.c
```diff
@@ -638,18 +641,124 @@ _ce_image_sync(u3e_image* img_u)
...
+//!
+//! @param[in] img_u  Image.
+//! @param[in] pgs_w  New size of the image.
+//! @param[in] bas_y  Base address of the image in memory. Used to establish a
+//!                   new mapping in memory. Should be NULL if no new mappings
+//!                   should be created.
+//!
+//! @n (1) The image is mapped into memory at base address `bas_y`.
+//! @n (2) The image shrunk.
+//! @n (3) The image grew.
```
joemfb, line 710: I've thought more about this style, and I still can't accept it. The nature of these comments is to granularly document specific parts of a function, but grouping them at the top and then referencing each inline forces the reader to manually dereference each comment. That indirection is cognitively costly, and the only reason for it is to make the comments legible to the doxygen machine. Machine-legibility is a good thing, but not worth damaging the readability or clarity of the code.

I object to those kind of tradeoffs in principle, but especially in this case: doxygen comments and generated artifacts are most appropriate for documenting interfaces for external consumption, but these comments are always about specific implementation details.

joemfb (changes requested):
I almost forgot:

pkg/urbit/noun/events.c
mcevoypeter: The intent behind all three macros was increased legibility without polluting the namespace of the file with additional static functions. I agree that `map_image()` is overkill, and I've removed it, but `read_image()` is used twice within the function *and* is not used anywhere else in the file, so I'm reluctant to add another function outside of the scope of `_ce_image_apply()`.

pkg/urbit/noun/events.c
mcevoypeter: I don't think dereferencing each comment is anymore cognitively expensive than some of the other tactics we use throughout the runtime, but at the same time I agree that the benefits don't outweigh the costs in this case. I've reverted back to a more conventional form of inline commenting.

mcevoypeter:
> To workaround the lack of windows compatibility for now, we should just unconditionally set `--no-mmap`.

Done.

> And to simplify the config passing (between processes, and within `events.c`), `--no-mmap` should become a bit-flag in `u3C.wag_w` -- that's already passed between the processes, and can be checked anywhere with `( u3C.wag_w & u3o_mmapf )` (or whatever the enum member is named), without adding module level config or additional function arguments. Also, you can back out the changes to IPC setup. (And that work wasn't wasted, the stack-buffer overflow in `lord.c` is a great catch, and something I would've missed since I wrote it.)

Also done.

Thanks for pointing both out, especially the use of `u3C.wag_w`.

joemfb (changes requested):
Thanks, one more round on the runtime config:

pkg/urbit/noun/events.c
joemfb: This is a good point, and I absolutely agree. There are some pretty cognitively-taxing patterns in those codebase: manual refcounting (and the style that makes it somewhat manageable), a very large surface of function-like macros (for concision or performance), and probably other things too.

pkg/urbit/noun/events.c
joemfb: Can you expound on this? It seems to me that these blocks, whether as macros or static functions, inhabit roughly the same namespace -- the implementation details of this subsystem. Neither impact the external api, and macros have the added strength/difficulty of admitting arbitrary patterns of control flow. And we don't need the optimization of manual inlining for the i/o here (where we arguably want it for the page bitmap).

At the same time, the change I requested here seems to have made the function more ugly -- maybe reversing the branches for handling `north.bin` help?

mcevoypeter: These macros, because they're defined *and* undefined within the same function,
actually occupy an informal, function-local namespace. If they were defined in
the file's namespace (i.e. the static namespace) or, even worse, the global
namespace, they'd be terrible, because any reader of `_ce_image_apply()` would
have to keep in mind the possibility that other functions within the file (or,
even worse, other files), may rely on these macros, which adds cognitive load
for the reader. Admittedly, none of these three macros do anything that a
function can't do. However, because macros can be "scoped" to a particular
function, I'd argue that the slightly higher complexity of macros is justified,
especially because the reader can easily look at the macros and verify that they
don't do anything exotic. Furthermore, they clearly encapsulate the logic of
`_ce_image_apply()`; just by glancing at the following code, it's apparent what
`_ce_image_apply()` does:

```
_ce_image_apply(u3e_image* nor_u, u3e_image* sou_u, c3_o pro_o)
{
#define mark_page_clean(address) /* snip */
#define read_image(image, base_address, should_protect, step_direction) /* snip */
#define map_image(image, base_address, page_protections) /* snip */
  if ( sou_u && sou_u->pgs_w > 0 ) {
    c3_y* bas_y = ((c3_y*)u3_Loom + u3a_bytes) - pag_siz_i;
    read_image(sou_u, bas_y, pro_o, -1);
  }

  if ( nor_u && nor_u->pgs_w > 0 ) {
    if ( u3C.wag_w & u3o_map_snapshot) {
      const c3_i pro_i = ( c3y == pro_o ) ? PROT_READ : PROT_READ | PROT_WRITE;
      map_image(nor_u, (c3_y*)u3_Loom, pro_i);
    }
    else {
      read_image(nor_u, (c3_y*)u3_Loom, pro_o, 1);
    }
  }
#undef mark_page_clean
#undef read_image
#undef map_image
}
```

So the intent here is not performance, but legibility, which I believe is a
legitimate use case of macros. Programming is really just the definition and
subsequent implementation of interfaces stacked one on top of each other 
and nested within each other, and by adding a function-local namespace, we
are able to better express the relationship between parts of an implementation
by only making visible (in this case) to other functions the functionality that
actually concerns them.

pkg/urbit/include/noun/events.h
```diff
@@ -94,20 +94,22 @@ u3e_copy(const c3_c* const dir_c);
...
 //!
 //! @param[in] dir_c  Directory to run the incremental snapshot system in. Will
 //!                   be created if it doesn't exist.
+//! @param[in] map_o  Don't `mmap()` the snapshot if `c3y`.
 //!
 //! @return c3y  No existing snapshot.
 //! @return c3n  Existing snapshot loaded.
 c3_o
-u3e_live(const c3_c* dir_c);
+u3e_live(const c3_c* dir_c, const c3_o map_o);
```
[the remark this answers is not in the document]
mcevoypeter: Good point. I overlooked this when I added the `u3C.wag_w` support.

pkg/urbit/noun/events.c
mcevoypeter: The code snippet above (with `map_image()`) perfectly encapsulates how I broke the problem down in my head as I was writing the code, which, in addition to correctness, is a property that I'd argue all code should have (or at least as much code as possible).

pkg/urbit/noun/events.c
```diff
@@ -638,18 +635,123 @@ _ce_image_sync(u3e_image* img_u)
...
+              img_u->nam_c,
+              ptr_y,
+              strerror(errno));
+      exit(1);
+    }
+
+    if ( dif_ws < 0 ) {
+      off_i += pag_siz_i;
+    }
+    ptr_y += pag_siz_i;
```
joemfb, line 745: These pointer/offset calculations are unnecessary, left over from the per-page mapping loop.

joemfb (changes requested):

joemfb (approved):
LGTM

PR #5941, opened 2022-08-11
gall: disable %doff for now

belisarius222:
We've seen some issues with `|doff`, so we're turning it off for now.  This leaves in the interface and hood generator, so we can enable it again easily.


Fang- (approved):
Straightforward! Looks good to me.

PR #5942, opened 2022-08-12
base: adds generator for getting %ames timer counts

joemfb:
Per ship, for the top N (default `50`) ships .


belisarius222 (approved):
lgtm -- the @ta ship representation took me a minute, but it's reasonable

PR #5947, opened 2022-08-15
gall: revive %hood and %dojo on keypress

belisarius222:
This is a band-aid to work around the upgrade bug where sometimes `%base` agents fail to restart.  Upon receiving a move whose target is `%base` or `%dojo`, instead of enqueueing or dropping the move, force-start `%dojo` and `%hood` and then retry the move.

I've tested this by running `|pass [%g %idle %hood]` and `|pass [%g %idle %dojo]`.  Both cases work fine as far as I can tell.  Earlier versions of this code that didn't roundtrip through Arvo had subscription problems after the revival.  Note that the `/nowhere` wire is important to use with `%jolt` to ensure resulting `%onto` agent restart notifications don't crash Gall.

I've targeted this PR at master, since that's what I based this code off of.  Happy to retarget if needed.


Fang- (dismissed):
We went over this together. Looks good and minimally invasive.

PR #5949, opened 2022-08-16
gall: re-enable doff, only for old-style subs

Fang-:
This is sufficient to bring the normal (non-prerelease-bugged) cases
into the new world.

For the prerelease ships that ran a buggier version of the new gall
subscription logic, we note that the conditional may trigger for the
nonce=1 case where it had already triggered for their
(shouldn't-be-possible) nonce=0 case. This results in a %leave on a wire
that wasn't in use. This no-ops on the publisher side though, and the
flow gets corked right away, so this is considered harmless.


belisarius222 (approved):
lgtm

PR #5950, opened 2022-08-16
gen: add +gall-nonces for reading all next nonces

Fang-:
Shows the list of agent-nonce pairs in ascending-nonce order, so that the (supposedly) most interesting ones are always easily visible.


belisarius222:
> the (supposedly) most interesting ones
lol

belisarius222 (approved):
lgtm

PR #5951, opened 2022-08-17
dns-address: subscribe before poke and fix wire mismatch

pkova:
My newfangled automation of `ship.arvo.network` domains introduced in #5868 exposed a bug in `ted/dns/address`: the thread pokes before subscribing. This often leads to situations where the DNS request is completed before the client subscribes, leaving this thread running forever.

### DIFF
--- a/pkg/arvo/ted/dns/address.hoon
+++ b/pkg/arvo/ted/dns/address.hoon
@@ -30,12 +30,12 @@
   ?.  good
     %+  strand-fail:strandio  %bail-early-self-check
     [>"couldn't access ship on port 80"< ~]
+  ;<  our=@p  bind:m  get-our:strandio
+  ;<  ~       bind:m  (watch:strandio /sub collector-app /(scot %p our))
   ;<  ~       bind:m  (poke:strandio collector-app %dns-address !>([%if if]))
   =/  msg=cord
     (cat 3 'request for DNS sent to ' (scot %p p:collector-app))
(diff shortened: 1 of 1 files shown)


Fang- (dismissed):
Classic. Good catch, thank you!

(Gonna hold off on merging for a little bit, but we'll get it out there soon enough.)

pkg/arvo/ted/dns/address.hoon
```diff
@@ -30,12 +30,12 @@
   ?.  good
     %+  strand-fail:strandio  %bail-early-self-check
     [>"couldn't access ship on port 80"< ~]
+  ;<  our=@p  bind:m  get-our:strandio
+  ;<  ~       bind:m  (watch:strandio /sub collector-app /(scot %p our))
```
tinnus-napbus, line 34:
```suggestion
  ;<  ~       bind:m  (watch:strandio /(scot %p our) collector-app /(scot %p our))
```

tinnus-napbus: it subscribes on a different wire than it expects in take-fact so you never get the response

tinnus-napbus (changes requested):

tinnus-napbus:
the watch was for wire `/sub` but `take-fact` expected wire `/(scot %p our)` so consequently the thread never received the response. I have changed both to `/response` @Fang- pls re-review

Fang- (approved):
Good catches all around, thank you both.

PR #5955, opened 2022-08-18
agentio: fix `+rest` typo, attempt 2

hosted-fornet:
Original post:

> The current code in agentio.hoon has +rest as a copy-paste of +wait, and thus sending a %wait rather than a %rest to Behn.

It was easier to make a new PR to target next/arvo than to rebase (which had conflicts).


tacryt-socryp (approved):
LGTM

PR #5956, opened 2022-08-18
+ames-flows: diagnostic generator to count ames flows

belisarius222:
This should help to diagnose issues with runaway resubscriptions or any Ames flows that have failed to close.


Fang- (approved):
We worked on this together yesterday. Still looks good today!

PR #5957, opened 2022-08-18
Added hoon eval option to Urbit Binary

mopfel-winrux:
As per Urbit Grant: https://urbit.org/grants/hoon-eval

I added the ability for the urbit binary to read some hoon code from stdin and evaluate it using the Ivory pill.

(I'm not sure what to add to this PR to make it official I can do whatever is needed)


pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,73 @@ _cw_intr_win(c3_c* han_c)
...
+  sil_u = u3s_cue_xeno_init_with(ur_fib27, ur_fib28);
+  if ( u3_none == (pil = u3s_cue_xeno_with(sil_u, len_d, byt_y)) ) {
+    printf("*** fail _setup 1\n");
+    exit(1);
+  }
+  u3s_cue_xeno_done(sil_u);
+  if ( c3n == u3v_boot_lite(pil) ) {
+    printf("*** fail _setup 2\n");
+    exit(1);
+  }
```
joemfb, line 1033: This could was copied from a unit test, it'd be better to use the error printfs from the equivalent in `king.c` (`_king_boot_ivory()`). Even better, declare that function so it's accessible here and then just call it.

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,73 @@ _cw_intr_win(c3_c* han_c)
...
+
+  printf("Executing hoon:\n");
```
joemfb, line 1035: Let's just say `"eval:"` here.

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,73 @@ _cw_intr_win(c3_c* han_c)
...
+  }
+
+
+
+  
+  if(0 == u3h(res)){//Succuessful execution print the output
+     u3_pier_tank(0,0,u3k(u3t(res)));
+  }else{
+     u3_pier_punt_goof("Hoon Error", u3k(res)); //print stack trace error
+  }
```
joemfb, line 1054: style: `else` should be on a new line, and there should be spaces around the conditional expression for `if`.

The `"Hoon Error"` caption should also probably just be `"eval"`, it'll print something like:

```
eval: bail: %meme
<stack trace>
```

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,73 @@ _cw_intr_win(c3_c* han_c)
...
+  u3z(res);
+  u3z(gat);
+
+  //printf("finishing\n");
+}
+
+
+
```
joemfb, line 1062: There's some extra whitespace in and around this function, it should be reduced to one line in each case. And the commented-out code can be deleted.

pkg/urbit/noun/vortex.c
```diff
@@ -107,6 +107,31 @@ _cv_nock_wish(u3_noun txt)
...
+      u3A->yot = u3kdb_put(u3A->yot, u3k(txt), u3k(exp));
+    }
+  }
+
+  u3t_event_trace("u3v_wish", 'e');
+
+  u3z(txt);
+  return exp;
+}
+
```
joemfb, line 134: We didn't end up needing this, so I think we should just remove it.

joemfb (changes requested):
This looks good. I have a memory management tweak and a few style points:

pkg/urbit/noun/vortex.c
mopfel-winrux: Removed

pkg/urbit/daemon/main.c
mopfel-winrux: Cleaned up

pkg/urbit/daemon/main.c
mopfel-winrux: Done

pkg/urbit/daemon/main.c
mopfel-winrux: Changed

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,73 @@ _cw_intr_win(c3_c* han_c)
...
+/* _cw_eval_commence(): initialize and run the hoon evaluator
+*/
+static void
+_cw_eval_commence(c3_i argc, c3_c* argv[])
+{
+  //Read from stdin until an EOF is recieved
+  c3_c* line = NULL;
+  size_t len = 0;
+  ssize_t nread;
+  c3_c* evl_c = "";
```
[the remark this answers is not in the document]
mopfel-winrux: Changed

pkg/urbit/daemon/main.c
mopfel-winrux: Changed to be the same error printfs

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,73 @@ _cw_intr_win(c3_c* han_c)
...
+
+  while((nread = getline(&line, &len, stdin)) != -1)
+  {
+     len = asprintf(&evl_c, "%s%s", evl_c, line);
+  }
+
+  free(line);
```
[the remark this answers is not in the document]
mopfel-winrux: Rewrote this to not leak as much memory. Not sure if its 100% leak free though

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,84 @@ _cw_intr_win(c3_c* han_c)
 }
 #endif
 
+c3_c 
+*_cw_eval_get_input(FILE* fp, size_t size)
```
joemfb, line 997: This should have a standard function comment, then  `c3_c*` on the next line, with rest of the signature on the line after.

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,84 @@ _cw_intr_win(c3_c* han_c)
...
+{
+//The size is extended by the input with the value of the provisional
+  c3_c *str;
+  c3_i ch;
+  size_t len = 0;
+
+  str = realloc(NULL, sizeof(*str)*size);//size is start size
+  
+  if(!str)
+    return str;
```
joemfb, line 1007: You should use our `c3_realloc()` wrapper, which asserts that the returned pointer is non-null. Also, the second parameter can just be `size` (I'm not concerned about the size of `char` changing).

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,84 @@ _cw_intr_win(c3_c* han_c)
...
+  
+  while(EOF!=(ch=fgetc(fp))){
```
joemfb, line 1009: This needs a single space inside the parens on each side for conditionals.

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,84 @@ _cw_intr_win(c3_c* han_c)
...
+    str[len++]=ch;
+    if(len==size){
```
joemfb, line 1011: This looks like it will only reallocate once. Seems like it should be `!(len % size)`, or that size should be increased after the reallocation.

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,84 @@ _cw_intr_win(c3_c* han_c)
...
+      str = realloc(str, sizeof(*str)*(size+=16));
+      if(!str)
+        return str;
```
joemfb, line 1014: `c3_realloc()`

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,84 @@ _cw_intr_win(c3_c* han_c)
...
+  //Run the input through a virtualization (u3v_wish) and get the ouput
+  u3_noun gat = u3v_wish("|=(a=@t (sell (slap !>(+>.$) (rain /eval a))))");
+  u3_noun res;
+  {
+    u3_noun sam = u3i_string(evl_c);
+    u3_noun cor = u3nc(u3k(u3h(gat)), u3nc(sam, u3k(u3t(u3t(gat)))));
+    res = u3m_soft(0, u3n_kick_on, cor); // Transfer cor and sam
+  }
+  
+  if(0 == u3h(res)){//Succuessful execution print the output
```
joemfb, line 1064: Spaces around the conditional

joemfb (changes requested):
Thanks for updating this, it looks better and more portable.

I see one issue with reallocation and a few style nits:

pkg/urbit/daemon/main.c
joemfb: This is wrong, I misread the following line.

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,87 @@ _cw_intr_win(c3_c* han_c)
...
+_cw_eval_get_input(FILE* fp, size_t size)
+{
+  c3_c *str;
+  c3_i ch;
+  size_t len = 0;
+
+  str = c3_realloc(NULL, size);//size is start size
+  
+  if( !str )
+    return str;
```
joemfb, line 1008: You can remove these now (allocation failure handled by `c3_realloc()`)

pkg/urbit/daemon/main.c
```diff
@@ -991,6 +993,87 @@ _cw_intr_win(c3_c* han_c)
...
+  
+  while( EOF != (ch=fgetc(fp)) ){
+    str[len++]=ch;
+    if( len == size ){
+      size +=16;
+      str = realloc(str, (size));
+      if(!str)
+        return str;
```
joemfb, line 1016: And here, switch to `c3_realloc()` and remove the conditional.

joemfb (changes requested):

joemfb (approved):
this will do (cf muji)

merging to a staging branch to work around intervening changes on `next/vere` -- thanks @mopfel-winrux!

joemfb:
And now on `next/vere`

PR #5958, opened 2022-08-18
gall: %rake task to clean up dead %leave's

belisarius222:
This sends a `%cork` on any old subscription ducts, which should have been corked already but might not have been if a ship had been running a prerelease version of the gall request queue fix.  h/t @joemfb 

The "all" flag should only be set if you want the ship to try to kill an old subscription at sub-nonce zero.  If your ship didn't run with a particularly old version of the gall request queue fix, then if you set `all=&`, gall will end up creating an extra Ames flow (and hopefully cleaning it up, but better not to find out).

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1792,14 +1792,24 @@
         $(subs t.subs)
       ::
       =/  dud=(set duct)
-        =/  mod  [%gall %use agent-name run-nonce.yoke %out 0 wire]
+        =/  mod=^wire
+          :*  %gall  %use
+              agent-name
+              run-nonce.yoke
+              %out
+              (scot %p p.dock)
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -421,8 +421,20 @@
       (drop (bind (~(get by yokes.state) u.dude) (lead u.dude)))
     |-  ^+  mo-core
     ?~  apps  mo-core
-    =/  ap-core  (ap-yoke:ap:mo-core p.i.apps [~ our] q.i.apps)
```
belisarius222, line 424: I probably didn't need to change this line, but it seems less brittle without actually firing the `+mo-core` arm.

Fang- (approved):
Dry-run tested on ~palfun, it hits all the right cases. Looks good!

PR #5959, opened 2022-08-19
u3: suppress bail:evil printf

joemfb:
Following on #5826, wherein I failed to realize that this would also print 2x per bad packet ...


belisarius222 (approved):
lgtm -- much cleaner, and it'll be good to turn off these printfs for real

PR #5960, opened 2022-08-19
gall: when kicked, register %cork as outstanding

Fang-:
When ames notifies us that our subscription has been kicked, we enqueue
a cork to clean up the flow. Unlike the %leave case, however, we were
not registering the cork in the queue of outstanding comms. We would
eventually get an ack, but not know what for, and erroneously inject
%poke-acks and %watch-acks.

Here we simply add a %cork entry to the queue before sending it.


belisarius222 (approved):
lgtm

PR #5961, opened 2022-08-19
ames: don't skip closing flows in +on-stir

Fang-:
Yes, there is a global timer for closing flows, but all that does is
enqueue a cork message. +on-stir needs to set _pump_ timers for all
flows that might still have messages to send, which includes closing
flows.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1484,10 +1484,9 @@
     =/  snds=(list (list [ship bone message-pump-state]))
       %+  turn  states
       |=  [=ship peer-state]
-      %+  murn  ~(tap by snd)
+      %+  turn  ~(tap by snd)
       |=  [=bone =message-pump-state]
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
lgtm

joemfb:
This PR uses a wire-literal as the base of a runtime-sourced duct -- that seems much more dangerous than doing similar things with "synthetic" ducts in `%gall`. Is this temporary cleanup migration code? If not, I think it'd be worth doing that timer-join another way.

belisarius222:
https://github.com/urbit/urbit/pull/5961/files#diff-2cea4c07b4facff63d912c6b2d14b2b6cc83b356e6d8b7ea74f205c5ae364d08R1507

This is just how Ames works -- all the pump timer ducts are synthetic.  It's not dangerous.  There's an argument to be made that we should keep the original duct that activated that Ames flow, but that would be a broader change, and not related to this PR.

joemfb:
@belisarius222 oh, I didn't realize it was always synthetic. Nevermind!

PR #5962, opened 2022-08-19
gall: don't inject potentially-fake acks

Fang-:
We shouldn't be hitting this case anymore. Injecting events into agents
which they might not expect, and might not even be real, is bad.


belisarius222 (approved):
lgtm

PR #5963, opened 2022-08-19
ames: ensure larva initializes adult's state

Fang-:
Previously, when the larva got to processing enqueued events, it was
doing so without loading state into the adult beforehand, resulting in
incorrect processing of events.

Here, we make the larva call +molt more eagerly, ensuring that the adult
always has its state available when we use it.


belisarius222 (approved):
lgtm

PR #5970, opened 2022-08-23
ames: fix upgrade bugs

belisarius222:
Fixes some issues with Ames's larval upgrade logic that were discovered when upgrading ~zod.


Fang- (approved):
This looks like what we paired on. Good!

PR #5971, opened 2022-08-23
ames: handle failed recork timers

joemfb:
Also fixes the error slog on failed attestation timers.


Fang- (approved):
This looks better to me, thanks!

belisarius222 (approved):
lgtm

PR #5973, opened 2022-08-24
http: check requests for auth using tokens from eyre

Fang-:
Preparatory work for part of #5927, but these changes can go in on their own.

Adds a `%sessions` gift to eyre that it uses to send a set of valid session tokens to then runtime, both on-`%born` and whenever that set changes.

Updates `http.c` to accept those gifts, and use them for checking authentication on the `/~_~/slog` printf stream endpoint by reading the cookie header, if any.  
Previously it used a clunky scry that put the full cookie string in the path. We should probably consider that scry endpoint deprecated now.

@joemfb perhaps you want these commits as separate PRs?


joemfb (commented):
These changes look reasonable to me. I still need to take a closer look at a couple parts. But I have some notes about releasing stuff like this:

This changes the external vane interface, and therefore requires a kelvin. The change is also required; slog streams will stop working on ships with a new binary and an old arvo, as the set of authentication tokens will be null. (Required in some sense, it's not a major feature.) So we can release the vere changes early, without adding a fallback to the authentication scry.

There's also the basic question of where to put changes like this (ie, "pending next kelvin bump"). I think we probably don't want it on next/arvo, but I'm not sure where it should go.

/cc @philipcmonk

jalehman:
This now requires two PRs. @Fang- will cherrypick the vere-related commits off of this to urbit/vere.

jalehman:
@Fang- is this ready to be reviewed? It looks like it, now that the C code is gone.

Fang-:
Yes. This should be fine to go in before/independently of urbit/vere#199, which I still need to address the feedback on. I don't think anything there should affect the interface implemented here.

belisarius222 (approved):

PR #5974, opened 2022-08-25
ames: bump recork timer back up to a day

Fang-:
We had lowered this initially for easier testing, but a day should be sufficient for real-world use.


belisarius222 (approved):
lgtm

PR #5976, opened 2022-08-25
ames: clean & re-set recork timer during %stir

Fang-:
Pre-release ships may have multiple recork timers. This ensures that, after stir, we always have exactly one.


belisarius222 (approved):
lgtm

PR #5977, opened 2022-08-25
ames, gall: no %done on corks

Fang-:
Locally, `%cork` is guaranteed to succeed instantly. The requesting vane doesn't need an ack to safely consider the flow "closed". Ames will now no longer eventually give a `%done` in response to `%cork` tasks.

Subsequently, gall will consider any `%cork`s it sends completed immediately, and no longer sticks it in the outstanding requests queue. It also cleans up the queue's entry whenever it's empty.

Largely untested at the time of writing, but I think this is right. At some point, we'll want to rip out gall's request queue entirely.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -42,9 +42,9 @@
 ::  $move: Arvo-level move
 ::
 +$  move  [=duct move=(wind note-arvo gift-arvo)]
-::  $state-9: overall gall state, versioned
+::  $state-10: overall gall state, versioned
 ::
-+$  state-9  [%9 state]
++$  state-10  [%10 state]
 ::  $state: overall gall state
 ::
 ::    system-duct: TODO document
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -343,11 +356,20 @@
...
         ^-  [bitt boat boar]
         [inbound outbound (~(run by outbound) |=([acked=? =path] nonce=0))]
+      ::
+      ++  spore-9-to-10
+        |=  old=spore-9
+        =-  old(- %10, outstanding -)
+        %-  ~(run by outstanding.old)
+        |=  q=(qeu remote-request-9)
+        %-  ~(gas to *(qeu remote-request))
+        %+  murn  ~(tap to q)
```
belisarius222, line 366: this could just be a `+skip`

Fang-: I believe I tried, but it didn't nest properly.

belisarius222 (approved):
lgtm

PR #5978, opened 2022-08-25
gall: don't add duplicate nonce in +ap-nuke

philipcmonk:
This gets added in +ap-handle-peers already.  This caused outgoing subscriptions to not get cleaned up properly.

Issue introduced in c2d77a5d.


belisarius222 (approved):
lgtm, I saw this fix the issue on a fakezod

PR #5982, opened 2022-08-30
eyre: look for lowercase last-event-id header

Fang-:
Eyre always gets passed request headers in lowercase, so we should search for
the lowercased version of the header.

Arguably `+get-header` should lowercase keys before comparing them, but that's
a more serious behavioral change.

(Notably browser-native EventSource clients don't generally make use of this header, which explains why this has lasted as long as it has. Userland implementations might though!)

ht @joemfb for catching this.


joemfb (approved):
LGTM

At some point, I'd be tempted to change the interface types (and maybe the request itself), to see if we could simply require that header names always be `@tas` or at least `@tas`.

PR #5983, opened 2022-08-31
hoon: further explore ?#, fix bugs

joemfb:
This PR adds `?/` and `?\` as `?#`-based versions of `?-` and `?+`. It also fixes a bug in `+lose:ar` -- the combinatorial explosion turns out to be necessary. Opening as a draft for further discussion/exploration; I'm not sure how we'll want to stage `?#`-related changes, or if the performance implications here are too dire.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -8987,9 +8987,9 @@
                          ref
                        [%atom %$ ~]
             [%atom *]  ref
-            [%cell *]  =+  ^$(skin skin.skin, ref p.ref)
-                       ?:  =(%void -)  %void
-                       (cell - ^$(skin ^skin.skin, ref q.ref))
+            [%cell *]  =/  lef  ^$(skin skin.skin, ref p.ref)
+                       =/  rig  ^$(skin ^skin.skin, ref q.ref)
+                       (fork (cell lef rig) (cell lef q.ref) (cell p.ref rig) ~)
             [%core *]  =+  ^$(skin skin.skin, ref p.ref)
                        ?:  =(%void -)  %void
                        ?.  =(%noun ^skin.skin)


joemfb:
setup: 

```
> =a `[(unit) (unit)]`[~ ~]
> =b `[(unit) (unit)]`[`1 `2]
```

before the change to `+lose:ar`:

```
> 
> ?/(a [~ ~] %no, [~ ^] %right, [^ ~] %left, [^ ^] %both)
mint-vain
dojo: hoon expression failed
> 
> ?/(b [~ ~] %no, [~ ^] %right, [^ ~] %left, [^ ^] %both)
mint-vain
> 
> ?\(b %foobar [~ ~] %no, [~ ^] %right, [^ ~] %left, [^ ^] %both)
mint-vain
dojo: hoon expression failed
>
> ?\(b %foobar [~ ~] %no, [~ ^] %right, [^ ~] %left)
mint-vain
dojo: hoon expression failed
> 
> 
> ?/(b [~ ~] %no, [~ ^] %right, [^ ~] %left)
mint-vain
dojo: hoon expression failed
>>
> ?/(b [~ ^] %right, [^ ~] %left)
%left
> ?/(a [~ ^] %right, [^ ~] %left)
%left
```

after:

```
> ?/(b [~ ~] %no, [~ ^] %right, [^ ~] %left, [^ ^] %both)
%both
> ?/(a [~ ~] %no, [~ ^] %right, [^ ~] %left, [^ ^] %both)
%no
```

Fang-:
This seems probably fine. Have you noticed the `+lose:ar` change making things significantly slower? If not, unclear to me why we would hesitate to just merge this in (and continue kicking the `?#` deployment can).

joemfb:
@Fang- I don't know how big the performance impact is yet. I'm planning to check that by porting something big enough to the new runes to compare (the `hoon` switches in the compiler are probably a pretty good candidate). Better exhaustiveness checking is also an opportunity for more efficient switches -- lots of `+dext`/`+sint` traversals could be combined.

`?/` and `?\` are pretty ugly, but it's less disruptive then adding state to `+vast` to support switching the implementation of `?-` and `?+`. I'm not sure we'd want to ship either option (beyond `?#` itself) without further exploration.

belisarius222:
@joemfb @Fang- What do you think we should do with this PR?  I can't tell from the discussion whether this is something we should just go ahead and try on a few canary ships, or whether the performance concerns are severe enough that we would want to revisit the design before even trying that.

Fang-:
I think the first step is to just do a proper performance comparison in the first place, which hasn't happened yet. That'll give us more information on how to proceed.

If it's "not that bad" or better, then it might be worth trying out a release of `?#` (maybe including `?/` and `?\`) so that developers can play around with them, see how they feel, how they operate in real-world cases. Of course with the caveat that these are intended for deprecation eventually.

belisarius222:
Ok, thanks @Fang-.  I don't think this is high priority at the moment, but it's good to know that the next steps are to test for performance regression, then if it passes, merge it so developers can start experimenting with the placeholder versions of the new runes.

jalehman:
@vvisigoth doing the performa the performance regression test and posting here is a good candidate for the grab bag bounty.

—

~wolref-podlex
https://urbit.org ( https://urbit.org/ )

joemfb:
I've dropped `?/` and `?\` from this PR (that implementation is at `jb/wthx-moar`).

PR #5985, opened 2022-09-01
vere: add pier validation subcommand, sane

Fang-:
There are several ways in which a pier (here, primarily its event log)
must be sane:
- At least the latest epoch directory must exist.
- All epoch directories together must form a contiguous sequence.
- The mug of an epoch's snapshot must match the mug of the last event of
  the preceding epoch.
- The snapshots must be usable.

Since `u3_saga_open` already traverses all the epochs, we add conditional
validation logic to it. A case may be made that this validation needs to
run always on-startup. The validation logic makes sure the above
invariants are maintained, though as currently implemented it only
checks the very latest snapshot. We may or may not want to check every single snapshot contained within the pier this way.

Existing `u3_saga_open` callsites remain unchanged. We simply add a
separate `u3_saga_sane` and hook it up to the same logic under the hood.
It returns a status code, where non-zero values indicate insanity. The
program exits with this status code, for easier programmatic
sanity-checking.

### DIFF
--- a/pkg/urbit/vere/saga.c
+++ b/pkg/urbit/vere/saga.c
@@ -235,32 +235,32 @@ _check_epoc_sanity(u3_epoc* const poc_u, c3_d* const nex_d, c3_l* const mug_l)
     if ( *mug_l != u3r_mug(u3A->roc) ) {
       u3l_log("saga: snapshot for epoch %s not contiguous: mug mismatch\n",
               u3_epoc_path_str(poc_u));
-      return FALSE;
+      return 0;
     }
   }
 
   //  read the next mug from the last event in the epoch
   if ( !u3_epoc_iter_open(poc_u, u3_epoc_last_commit(poc_u)) ) {
     u3l_log("saga: failed to open epoch event log\n");
-    return FALSE;
+    return 0;
   }
   c3_y*  dat_y;
   size_t dat_i;
   if ( !u3_epoc_iter_step(poc_u, &dat_y, &dat_i) ) {
     u3l_log("saga: failed to read last event\n");
     u3_epoc_iter_close(poc_u);
-    return FALSE;
+    return 0;
   }
   u3_epoc_iter_close(poc_u);
   if ( dat_i < 4 ) {
     u3l_log("saga: malformed event, too short for mug\n");
-    return FALSE;
+    return 0;
   }
   *mug_l = dat_y[0]
          | dat_y[1] << 8
          | dat_y[2] << 16
          | dat_y[3] << 24;
-  return TRUE;
+  return 1;
 }
 
 static c3_t
@@ -518,7 +518,7 @@ u3_saga_open(const c3_path* const pax_u, c3_t san_t, u3_meta* const met_u)
   try_list(log_u->epo_u.lis_u = c3_list_init(), goto free_dir_entries);
   u3_epoc *poc_u;
   c3_w* lif_w = &met_u->lif_w;
-  c3_t  pas_t = TRUE;
+  c3_t  pas_t = 1;
   c3_d  nex_d = 0;
   c3_l  mug_l = 0;
   for ( size_t idx_i = 0; idx_i < ent_i; idx_i++ ) {
--- a/pkg/urbit/vere/saga.c
+++ b/pkg/urbit/vere/saga.c
@@ -209,6 +209,65 @@ _cmp_epocs(const void* lef_v, const void* rih_v)
   return len_i == ren_i ? strcmp(lef_c, rih_c) : len_i - ren_i;
 }
 
+static c3_o
+_check_epoc_sanity(      u3_epoc* poc_u,
+                         c3_path* pax_u,
+                   const c3_c*    ent_c,
+                         c3_d*    nex_d,
+                         c3_l*    mex_l)
+{
+  //  check epoch's event range for contiguousness with earlier epochs.
+  //  first epoch may start at arbitrary event nr, so we don't check it.
+  c3_d fir_d = u3_epoc_first_commit(poc_u);
+  if ( (0 != *nex_d) && (*nex_d != fir_d) ) {
+    u3l_log("saga: snapshot for epoch %s not contiguous: "
+            "expected first event of %" PRIu64 ", got %" PRIu64 "\n",
+            u3_epoc_path_str(poc_u), *nex_d, fir_d);
+    return c3n;
+  }
+  *nex_d = u3_epoc_last_commit(poc_u) + 1;
+
+  //  load the snapshot so we can validate its mug
+  if ( 0 != *mex_l ) {
+    //NOTE  careful, will clobber all existing noun references!
+    c3_path_push(pax_u, ent_c);
+    //TODO  want _boot_from_epoc_snapshot but it ends up crashing in events.c
+    u3m_init();
+    u3e_load(c3_path_str(pax_u));
+    u3m_pave(c3n);
+    c3_path_pop(pax_u);
+
+    if ( *mex_l != u3r_mug(u3A->roc) ) {
+      u3l_log("saga: snapshot for epoch %s not contiguous: mug mismatch\n",
+              u3_epoc_path_str(poc_u));
+      return c3n;
+    }
+  }
+
+  //  read the next mug from the last event in the epoch
+  if ( !u3_epoc_iter_open(poc_u, u3_epoc_last_commit(poc_u)) ) {
+    u3l_log("saga: failed to open epoch event log\n");
+    return c3n;
+  }
+  c3_y*  dat_y;
+  size_t dat_i;
+  if ( !u3_epoc_iter_step(poc_u, &dat_y, &dat_i) ) {
+    u3l_log("saga: failed to read last event\n");
+    u3_epoc_iter_close(poc_u);
+    return c3n;
+  }
+  u3_epoc_iter_close(poc_u);
+  if ( dat_i < 4 ) {
+    u3l_log("saga: malformed event, too short for mug\n");
+    return c3n;
+  }
+  *mex_l = dat_y[0]
+         | dat_y[1] << 8
+         | dat_y[2] << 16
+         | dat_y[3] << 24;
+  return c3y;
+}
+
 static c3_t
 _create_metadata_files(const u3_saga* const log_u, const u3_meta* const met_u)
 {
@@ -413,7 +472,7 @@ succeed:
 }
 
 u3_saga*
-_saga_open(const c3_path* const pax_u, u3_meta* const met_u, c3_i* san_i)
+u3_saga_open(const c3_path* const pax_u, c3_o san_o, u3_meta* const met_u)
 {
   u3_saga* log_u = c3_calloc(sizeof(*log_u));
   if ( !(log_u->pax_u = c3_path_fp(pax_u)) ) {
@@ -464,6 +523,7 @@ _saga_open(const c3_path* const pax_u, u3_meta* const met_u, c3_i* san_i)
   try_list(log_u->epo_u.lis_u = c3_list_init(), goto free_dir_entries);
   u3_epoc *poc_u;
   c3_w* lif_w = &met_u->lif_w;
+  c3_o  pas_o = c3y;
   c3_d  nex_d = 0;
   c3_l  mex_l = 0;
   for ( size_t idx_i = 0; idx_i < ent_i; idx_i++ ) {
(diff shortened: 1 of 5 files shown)


pkg/urbit/vere/saga.c
```diff
@@ -375,7 +375,7 @@ u3_saga_new(const c3_path* const pax_u, const u3_meta* const met_u)
 }
 
 u3_saga*
-u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
+_saga_open(const c3_path* const pax_u, u3_meta* const met_u, c3_i* san_i)
```
mcevoypeter, line 378: `static` functions should go in the "Static functions" section of the file, which is alphabetized by `static` function name and which begins with declarations of all of the `static` functions (to enable alphabetization of the function definitions).

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+
+    if ( NULL != san_i ) {
+      //  check epoch's event range for contiguousness with earlier epochs.
+      //  first epoch may start at arbitrary event nr, so we don't check it.
+      //
+      c3_d fir_d = u3_epoc_first_commit(poc_u);
+      if ( (0 != nex_d) && (nex_d != fir_d) ) {
+        u3l_log("sane: epoch snapshot not contiguous: "
+                "expected %" PRIu64 ", got %" PRIu64 ", in %s\n",
+                nex_d, fir_d, ent_c[idx_i]);
```
mcevoypeter, line 444: Can you change the message prefix from `sane:` to `saga:`?

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+      if ( 0 != mex_l ) {
+        //NOTE  careful, will clobber all existing noun references!
+        c3_path_push(log_u->pax_u, ent_c[idx_i]);
+        u3m_init();
+        u3e_load(c3_path_str(log_u->pax_u));
+        u3m_pave(c3n);
+        c3_path_pop(log_u->pax_u);
+
+        if ( mex_l != u3r_mug(u3A->roc) ) {
+          u3l_log("sane: epoch snapshot not contiguous, mug mismatch in %s\n",
```
mcevoypeter, line 461: See message prefix comment above.

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+      //
+      if ( !u3_epoc_iter_open(poc_u, u3_epoc_last_commit(poc_u)) ) {
+        u3l_log("sane: failed to open epoch event log\n");
+        *san_i = 3;
+        goto next;
+      }
+      c3_y*  dat_y;
+      size_t dat_i;
+      if ( !u3_epoc_iter_step(poc_u, &dat_y, &dat_i) ) {
+        u3l_log("sane: failed to read last event\n");
```
mcevoypeter, line 478: See message prefix comment above.

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+        *san_i = 4;
+        u3_epoc_iter_close(poc_u);
+        goto next;
+      }
+      u3_epoc_iter_close(poc_u);
+      if ( dat_i < 4 ) {
+        u3l_log("sane: strange event, too short\n");
```
mcevoypeter, line 485: See message prefix comment above.

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+    if ( NULL != san_i ) {
+      //  check epoch's event range for contiguousness with earlier epochs.
+      //  first epoch may start at arbitrary event nr, so we don't check it.
+      //
+      c3_d fir_d = u3_epoc_first_commit(poc_u);
+      if ( (0 != nex_d) && (nex_d != fir_d) ) {
+        u3l_log("sane: epoch snapshot not contiguous: "
+                "expected %" PRIu64 ", got %" PRIu64 ", in %s\n",
+                nex_d, fir_d, ent_c[idx_i]);
+        *san_i = 1;
```
mcevoypeter, line 445: I'm not a fan of the status codes because they don't seem to add much additional information, especially considering that they're undocumented. I'd be inclined to rely on the `NULL`/not-`NULL` state of the return value to determine whether the event log was sane (if not sane, the return value will be `NULL`) and allow the log messages to communicate *why* the event log was not sane.

pkg/urbit/vere/saga.c
mcevoypeter, line 485: Let's change this log message to be more specific. Something like:
```c
u3l_log("saga: malformed event doesn't have enough space for a mug\n");
```

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+      c3_y*  dat_y;
+      size_t dat_i;
+      if ( !u3_epoc_iter_step(poc_u, &dat_y, &dat_i) ) {
+        u3l_log("sane: failed to read last event\n");
+        *san_i = 4;
+        u3_epoc_iter_close(poc_u);
+        goto next;
+      }
+      u3_epoc_iter_close(poc_u);
+      if ( dat_i < 4 ) {
```
mcevoypeter, line 484: Let's replace `4` with `sizeof(mex_l)`:
```c
if ( dat_i < sizeof(mex_l) ) {
```

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+        //NOTE  careful, will clobber all existing noun references!
+        c3_path_push(log_u->pax_u, ent_c[idx_i]);
+        u3m_init();
+        u3e_load(c3_path_str(log_u->pax_u));
+        u3m_pave(c3n);
+        c3_path_pop(log_u->pax_u);
+
+        if ( mex_l != u3r_mug(u3A->roc) ) {
+          u3l_log("sane: epoch snapshot not contiguous, mug mismatch in %s\n",
+                  ent_c[idx_i]);
```
mcevoypeter, line 462: Instead of just the epoch directory name, let's use the full epoch path via `u3_epoc_path_str(poc_u)`.

pkg/urbit/vere/saga.c
mcevoypeter, line 444: Let's change this to be more specific:
```c
u3l_log("saga: snapshot for epoch %s is not contiguous with the previous epoch:"
        " expected first event to be %" PRIu64 ", instead have %" PRIu64 "\n",
        u3_epoc_path_str(poc_u),
        nex_d,
        fir_d);
```

pkg/urbit/vere/saga.c
```diff
@@ -456,6 +522,21 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+{
+  return _saga_open(pax_u, met_u, NULL);
+}
+
+c3_i
+u3_saga_sane(const c3_path* const pax_u)
+{
+  u3_meta met_u;
+  c3_i    san_i = 0;
+  _saga_open(pax_u, &met_u, &san_i);
```
mcevoypeter, line 536: This leaks the `u3_saga` handle.

pkg/urbit/vere/saga.c
```diff
@@ -424,17 +424,83 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+
+next:
     c3_list_pushb(log_u->epo_u.lis_u, poc_u, epo_siz_i);
     c3_free(poc_u);
     if ( lif_w ) {
       lif_w = NULL;
     }
+    if ( NULL != san_i && 0 != *san_i ) {
+      break;
+    }
```
mcevoypeter, line 503: This should be `goto free_dir_entries`, not `break`, so that we fail if an epoch is not sane instead of continuing to open the event log.

mcevoypeter (changes requested):
Thanks for doing this! It's great to see how well this additional functionality fits into the existing code. I've made a number of comments, which can be bucketed into three categories (ordered from most important to least important):
1. Interface design
2. Memory management
3. Style

pkg/urbit/vere/saga.c
```diff
@@ -456,6 +522,21 @@ u3_saga_open(const c3_path* const pax_u, u3_meta* const met_u)
...
+
+c3_i
+u3_saga_sane(const c3_path* const pax_u)
+{
+  u3_meta met_u;
+  c3_i    san_i = 0;
+  _saga_open(pax_u, &met_u, &san_i);
+  return san_i;
+}
+
```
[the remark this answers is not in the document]
Fang-: Hah, actually had this implemented as `(1)` at first, but ended up deciding against it because it felt like a weird thing to do.

I can see your argument for not wanting random exit statuses. There are already non-sanity-checking cases where we return `NULL` though, I guess we're fine not being able to distinguish those in the do-sanity-check case?

pkg/urbit/vere/saga.c
Fang-: But a mug is defined as always being four bytes? And we're gonna be reading out four bytes. The fact that we're storing it in a variable that has a matching size seems more incidental than something to rely on, if that makes sense. Using `sizeof` feels like it inverts the relationship.

pkg/urbit/vere/saga.c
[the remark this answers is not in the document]
mcevoypeter: Ha, classic. The way I see it, `u3_saga_open()` should return `NULL` when the event log couldn't be opened, whether that's the result of a sanity check or not. I can't think of any cases where you'd want to do X when the event log couldn't be opened for a non sanity check reason and Y when the event log couldn't be opened because the sanity check failed. All that said, I think standardized return statuses would be a huge improvement to Vere, but I think to do so will require strategizing.

pkg/urbit/vere/saga.c
mcevoypeter: I see what you mean. Would it feel different if we used `sizeof(c3_l)`? I know `4` is used for "size of mug" elsewhere in the codebase, but I generally have an aversion to raw literals.

belisarius222:
> as currently implemented it only
checks the very latest snapshot. We may or may not want to check every single snapshot contained within the pier this way.

The ship will run just fine if only old epochs are corrupt, so I don't think a failed integrity check should prevent starting the ship.  As long as replay succeeds, your ship is fine.  Loss of noncritical historical data should show up as a warning, something like an iphone telling you it's having a problem with its icloud backups.

pkg/urbit/vere/saga.c
Fang-: It would not feel different. If we're really that afraid of non-magic magic numbers I'd maybe accept a `MUG_BYTES` constant somewhere... But even then that's somewhat fake because we still manually index `0` through `3` in various places, and other such mug-size-based logic.

pkg/urbit/vere/saga.c
```diff
@@ -209,6 +209,65 @@ _cmp_epocs(const void* lef_v, const void* rih_v)
...
+  *nex_d = u3_epoc_last_commit(poc_u) + 1;
+
+  //  load the snapshot so we can validate its mug
+  if ( 0 != *mex_l ) {
+    //NOTE  careful, will clobber all existing noun references!
+    c3_path_push(pax_u, ent_c);
+    //TODO  want _boot_from_epoc_snapshot but it ends up crashing in events.c
+    u3m_init();
+    u3e_load(c3_path_str(pax_u));
+    u3m_pave(c3n);
```
mcevoypeter, line 237: The issue with `_boot_from_epoc_snapshot()` is that it assumes that `u3m_init()` has been called already (this assumption is undocumented, but I'll explicitly add it to the documentation). To resolve the crash in `events.c`, I think that the following change to `_cw_sane()` will work:
```c
u3m_init();
u3e_live(u3_Host.dir_c);
// ...
u3_saga* const log_u = u3_saga_open(/* snip */);
```

mcevoypeter (changes requested):
Thanks for incorporating the first round of feedback. This next round should be quicker :)

pkg/urbit/vere/saga.c
mcevoypeter: After a second review, I realized that `mex_l` most likely should be `mug_l`, in which case the change that I suggested would result in `if ( dat_i < sizeof(mug_l) ) {`, which maps very nicely into "if `dat_i` is smaller than the size of a mug". I see no problem with this.

In practice at this point in time in this version of the Urbit runtime, I understand that a mug is always four bytes. However, I'd argue a mug *happens* to be 4 bytes but is not defined to be so. In the hypothetical case in which we choose a different hash function or even a different variant of the same `murmur3` hash function, the mug--defined in this case to be a hash of a noun--is not necessarily four bytes. So while using `sizeof()` or a named constant `MUG_BYTES` does not address the use of literal indexes with respect to mugs elsewhere in the codebase, I think it better conveys in code the notion of what a mug is.

zalberico:
@mcevoypeter is this still relevant since we had to close the event log truncation one due to mars/urth? Basically wondering if this will eventually be merged or if we're going to have to write something in the new repo different enough that we should close. Thoughts?

mcevoypeter:
@zalberico This work depended heavily on #5701; since we closed #5701, I'd close this too.

PR #5987, opened 2022-09-03
eyre: add auto-https redirects

tinnus-napbus:
if a cert is configured and a secure port is live it will set the
redirect flag in http-config.state.

When it gets a ++request it will return a 301 redirect to
https://[host]/[path] if:

  1. not already secure
  2. redirect flag set
  3. secure port live
  4. is not requesting /.well-known/acme-challenge/...
  5. the host is in domains.state

It will not happen if forwarded-secured, localhost, local loopback, ip
addresses or domains not in domains.state.

in ++load it checks the secure port is live and a cert is set and
enables it if so (for people who already use in-urbit letencrypt)

%rule %cert tasks also toggle it (only turning it on if secure port
live)

%live tasks also toggle it (only turning it on if cert set)

Have tested with a couple of ships and seems to work fine.

This is useful in combination with pyry's auto arvo.network dns config
system - can finally get rid of reverse proxies entirely.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -551,6 +551,18 @@
       (easy ~)
     ==
   ==
+::  +host-parser: returns [(unit userinfo=@t) (unit port=@ud) host:eyre]
+::
+++  host-parser
+  |^
+  ;~(plug userinfo thor:de-purl:html)
+  ++  userinfo       (punt (cook crip ;~(sfix (star userinfo-char) pat)))
+  ++  userinfo-char  ;~(pose col unreserved sub-delims pct-encoded)
+  ++  unreserved     ;~(pose aln hep dot cab sig)
+  ++  sub-delims     ;~(pose zap buc pam soq pal par tar lus com mic tis)
+  ++  pct-encoded    (cook crip ;~(plug cen hex-char hex-char (easy ~)))
+  ++  hex-char       ;~(pose nud (shim 'a' 'f') (shim 'A' 'F'))
+  --
 ::  +per-server-event: per-event server core
 ::
 ++  per-server-event
@@ -580,7 +592,7 @@
   ::
   ++  request
     |=  [secure=? =address =request:http]
-    ^-  [(list move) server-state]
+    |^  ^-  [(list move) server-state]
     =*  headers  header-list.request
     ::  for requests from localhost, respect the "forwarded" header
     ::
@@ -602,6 +614,35 @@
       [action [authenticated secure address request] ~ 0]
     =.  connections.state
       (~(put by connections.state) duct connection)
+    :: redirect to https if insecure, redirects enabled,
+    :: secure port live, not an acme challenge and host
+    :: is in domains.state
+    ::
+    ?:  ?&  !secure
+            redirect.http-config.state
+            ?=(^ secure.ports.state)
+            ?!  ?=  [* [%'.well-known' %acme-challenge *] *]
+                (parse-request-line url.request)
+            (host-in-domains host)
+        ==
+      =/  location=@t
+        ;:  (cury cat 3)
+          'https://'
+          (need host)
+          ?:  =(443 u.secure.ports.state)
+            ''
+          (crip ":{(a-co:co u.secure.ports.state)}")
+          ?:  ?=([[~ ~] ~] (parse-request-line url.request))
+            '/'
+          url.request
+        ==
+      %-  handle-response
+      :*  %start
+          :-  status-code=301
+          headers=['location' location]~
+          data=~
+          complete=%.y
+      ==
     ::  figure out whether this is a cors request,
     ::  whether the origin is approved or not,
     ::  and maybe add it to the "pending approval" set
@@ -712,6 +753,20 @@
       %^  return-static-data-on-duct  404  'text/html'
       (error-page 404 authenticated url.request ~)
     ==
+    :: test if host header is valid and turf in domains.state
+    ::
+    ++  host-in-domains
+      |=  raw-host=(unit @t)
+      ^-  ?
+      ?~  raw-host  |
+      =/  auth=(unit [* * =host:eyre])
+        (rush u.raw-host host-parser)
+      ?&  ?=(^ auth)
+          ?=(%.y -.host.u.auth)
+          (~(has in domains.state) p.host.u.auth)
+      ==
+    --
+::
   ::  +handle-scry: respond with scry result, 404 or 500
   ::
   ++  handle-scry
@@ -2271,6 +2326,10 @@
       ::
       %live
     =.  ports.server-state.ax  +.task
+    ::  enable http redirects if https port live and cert set
+    ::
+    =.  redirect.http-config.server-state.ax
+      &(?=(^ secure.task) ?=(^ secure.http-config.server-state.ax))
     [~ http-server-gate]
       ::  %rule: updates our http configuration
       ::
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +753,20 @@
       %^  return-static-data-on-duct  404  'text/html'
       (error-page 404 authenticated url.request ~)
     ==
+    :: test if host header is valid and turf in domains.state
+    ::
+    ++  host-in-domains
```
Fang-, line 758: You may be able to deduplicate this with the existing logic in `+get-action-for-binding` and `+host-matches`. Not immediately obvious what the elaborate `+host-parser` is for, isn't the host header always just `host(:port)`?

Fang- (commented):
Hard-coding the ACME challenge path in Eyre doesn't feel quite right to me. We also might not actually need to do this. According to [LE's description of the `http-01` challenge that `/app/acme` uses](https://letsencrypt.org/docs/challenge-types/#http-01-challenge):

> Our implementation of the HTTP-01 challenge follows redirects, up to 10 redirects deep. It only accepts redirects to “http:” or “https:”, and only to ports 80 or 443. It does not accept redirects to IP addresses. When redirected to an HTTPS URL, it does not validate certificates (since this challenge is intended to bootstrap valid certificates, it may encounter self-signed or expired certificates along the way).

Which sounds like it would work over the secure connection just as well. Worth using that if we can get away with it.

Additionally, it looks like the `redirect` field in `http-config` was... never actually used prior to this, which is funny. That field seems like the logical candidate for these changes, except now we're always deriving its value/behavior from other parts of state, instead of having it be un/set explicitly, making it redundant at best.

If we're going to commit to redirecting to https whenever possible, we should just remove the `redirect` from `http-config` and always do the check (secure & cert) on the spot. The alternative is keeping it, and then probably having it set alongside a `%cert` config change. The latter gives you a bit more control, but the former is probably better from a modern http security standpoint. Are there ever cases where a user needs to escape the redirect? Even if the cert is bad, browsers give you the option to proceed anyway right?

pkg/arvo/sys/vane/eyre.hoon
tinnus-napbus: the host-parser is mostly an overly RFC compliant parser for the userinfo portion eg `foo:bar@[host][port]`. The zuse url parsing function that handles it just uses an @ta parser which is inadequate as it doesn't handle uppercase letters and other legal symbols.

The basic issue is I have to concat 'https://' + host + path. When I tested with curl I found the runtime will pass through any insane value for the host header like  `!!!  \\  e\\\\` or w/e without restriction which will result in an invalid location header for the redirect. I wanted to err on the side of only producing a valid redirect location but maybe that's not necessary. Arguably it's still too restrictive as I think the zuse domain parser doesn't handle ipv6 and other weird possibilities.

What you think? Should I just accept any value in the host header and concat it into the location header without caring or should I try simplify the parser by excluding some of the technically legal but unlikely userinfo characters or something else

tinnus-napbus:
> Which sounds like it would work over the secure connection just as well. Worth using that if we can get away with it.

nice yeah I didn't realise it behaved like this, I can remove that exception then

on the question of the redirect field yeah it's redundant, I only used it cos it was already there and I didn't want to change state or the api at all and it would have been weird to have it but not use it.

I feel that it's unlikely you'd not want to force https. There are probably obscure cases like a microcontroller that can't handle tls. If you're doing weird stuff like that you should probably just put urbit behind a more configurable reverse proxy. Otherwise adding it as an option to %cert would be good, although it raises the question of how to transition existing tls cert configurations in +load

tinnus-napbus:
so yeah my vote would be to just get rid of the redirect field

joemfb:
The redirect field was added to the config because that whole structure is sent to the runtime -- the idea was to do this kind of thing directly in `http.c`. If the redirect can be unconditional (in the presence of a configured/working secure server), I think it'd be better to just do it there.

Fang-:
Oh yeah, good call @joemfb, we can have the runtime do redirects as well as an optimization. I think Eyre should continue to do redirects also though, since it may be running within less-fancy runtimes.

I think even in the runtime the "do we have a secure port & a cert" check is possible without explicit configuration, so shouldn't be too hard to add. @tinnus-napbus it's probably outside of scope for this PR, but if you're feeling daring then the matching `http.c` change would be very welcome.

> There are probably obscure cases like a microcontroller that can't handle tls. If you're doing weird stuff like that you should probably just put urbit behind a more configurable reverse proxy.

Proxying to work around this doesn't feel like a satisfying solution. I was gonna suggest to just tell people to bind a subdomain and not set up certs for that, so it won't redirect in the first place, but configuration isn't per-domain yet:

https://github.com/urbit/urbit/blob/4bef3277d5ebfe5340bc8f528b37b80d38ac27b1/pkg/arvo/sys/lull.hoon#L1159-L1164

We should... get around to fixing that, but this too is out of scope here.

pkg/arvo/sys/vane/eyre.hoon
Fang-: My understanding is that userinfo is always excluded from the host header, as per [the http1.1 spec](https://httpwg.org/specs/rfc7230.html#header.host). Is there something I'm misunderstanding?

My inclination would be to just pass the host into redirect field without even inspecting it. If they send bogus, they receive bogus. If we want to reject requests with strange host headers that should probably happen higher up...

tinnus-napbus: oh you're right I didn't realise userinfo was meant to go in a separate header. Yeah I'll just remove that whole check

tinnus-napbus:
> but configuration isn't per-domain yet

the logic I implemented only redirects if the host domain as a turf exists in domains.state. This means a subdomain that isn't configured won't redirect, so you can get around redirects with a subdomain. Idk if this logic is correct actually. It kinda assumes the flow where eyre gets a %turf %rule and then eyre notifies acme and acme does its thing and sends eyre a %cert %rule. With this flow you can reasonably assume domains with certs exist in domains.state. Technically you can just pass eyre a %cert %rule without using acme in which case domains.state will not reflect the domains in the certificate itself. Maybe this is a wrong design.

Fang-:
> I feel you don't want to be redirecting to https if the certificate is not valid for that domain but ...

Yeah my thinking was to just... not care about this? If we have a cert, we assume it to be usable. If you don't want https at all, don't configure a cert, and you won't be redirected.

tinnus-napbus:
ok @Fang- I have simplified it to always redirect if not secure, redirects enabled & secure port active. I still had to do some host parsing cos I need to strip the `:<port>` but it's fairly simple now

Fang- (commented):
Some suggestions for style improvements, and a simplified parser, but otherwise I think this is quite good now.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -551,6 +551,18 @@
...
+++  host-sans-port
+  ;~  sfix
+    %+  cook  crip
+    %-  star
+    ;~  less
+      ;~(plug col (punt dem) ;~(less next (easy ~)))
+      next
+    ==
+    (star next)
+  ==
```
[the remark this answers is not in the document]
Fang-: I guess this means an alternative is doing a `+find` for `":"` on a `+trip`ped hostname and just going off that, instead of full-on parsing...

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -712,6 +749,7 @@
       %^  return-static-data-on-duct  404  'text/html'
       (error-page 404 authenticated url.request ~)
     ==
+::
   ::  +handle-scry: respond with scry result, 404 or 500
   ::
```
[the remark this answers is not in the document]
tinnus-napbus: whoops idk how that got added

pkg/arvo/sys/vane/eyre.hoon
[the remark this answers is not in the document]
tinnus-napbus: I was concerned about ipv6 hosts which are like `https://[FEDC:BA98:7654:3210:FEDC:BA98:7654:3210]:8080/foo/bar`. It's unlikely that you'd get an ipv6 hostname cos urbit doesn't support it but it's probably possible with a reverse proxy. Idk if worth accounting for this possibility

PR #5988, opened 2022-09-05
hoon: full HTML attribute support

arthyn:
Currently sail only supports attributes that are compatible with `@tas` which is unfortunately not compatible with the full HTML spec: https://html.spec.whatwg.org/multipage/syntax.html#attributes-2

To bring it inline I added a type for attribute names `+$  matt  @t` and separated the parsing so that we could apply it correctly.

Also I'm not entirely sure my mask is correct, open to hearing what might be missing.


arthyn:
Closing in favor of doing something in userspace to shove in "invalid" chars into mane rather than writing an HTML parser

PR #5989, opened 2022-09-06
hoon: use static nock formulas for boot and virtualization

joemfb:
This PR cleans up the boot formulas, pill generators, and various virtualization apis, making them avoid unnecessary variation in the nock formulas they use. This avoids pressure on the bytecode compiler and cache, and is just generally more correct. @frodwith has been after me to clean this sort of thing up for a long time. (The change to `+slot` is probably spurious, depending on the precise behavior of an interpreter.)

These changes may help @jtobin with some of his caching work.

I've applied to this to a running ship and tested the apis in question, built the relevant pill types and successfully booted from them all.

/cc @eamsden

### DIFF
--- a/pkg/arvo/gen/solid.hoon
+++ b/pkg/arvo/gen/solid.hoon
@@ -77,10 +77,8 @@
   ~&  %solid-double-loaded
   =/  whole-formula
     =<  +
-    .*  0
-    :+  %7
-      compiler-formula
-    [%9 2 %10 [6 %1 %noun whole-src] [%0 1]]
+    .*  [%noun whole-src]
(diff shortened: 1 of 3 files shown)


frodwith (commented):
`[0 axe]` is the most cromulent of dynamic formulas, since nock's `0` has a static argument. This change just moves the construction of the dynamic formula into explicit nock, which doesn't help, but there is no way to help.

Edit: Just referring to the first commit here (+slot). I can't github.

joemfb:
Confirmed that the change to `+slot` was spurious (per conversation with @frodwith) and removed it.

philipcmonk (approved):
Looks good, and feels very correct to not be regularly generating novel nock formulas.

PR #5994, opened 2022-09-12
garden: fix broken treaty JSON

liam-fitzgerald:
For some reason, a %uv case was added to the JSON serialiser for $case. $case has no such case in the union, and so this does not make sense.

### DIFF
--- a/pkg/garden/lib/treaty.hoon
+++ b/pkg/garden/lib/treaty.hoon
@@ -27,7 +27,6 @@
       %da   s+(scot %da p.c)
       %tas  s+(scot %tas p.c)  
       %ud   (numb p.c)
-      %uv   s+(scot %uv p.c)
     ==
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Yeah this is definitely not supposed to be in yet, it's part of remote scry's line of work (which adds a `%uv` case for, iirc, "commit hash case").

liam-fitzgerald:
how do we feel about merging this straight into master? It's not affecting anything on the network *yet* but I feel pretty strongly that current master should always be buildable

Fang-:
If this change was erroneously merged into master, and is not deployed to the network, I'm fine just merging what is essentially a revert straight into master.

`next/garden` or w/e is the appropriate branch is might do good to have master merged back in, and do a small round of testing and/or diff against mainnet ships, just in case other things slipped by.

PR #5995, opened 2022-09-13
npm: several updates to http-api

patosullivan:


matildepark:
Can you throw #5991 on this?

pkg/npm/http-api/src/Urbit.ts
```diff
@@ -142,7 +144,7 @@ export class Urbit {
     code,
     verbose = false,
   }: AuthenticationInterface) {
-    const airlock = new Urbit(`http://${url}`, code);
+    const airlock = new Urbit(url.startsWith('http') ? url : `http://${url}, code);
```
arthyn, line 147:
```suggestion
    const airlock = new Urbit(url.startsWith('http') ? url : `http://${url}`, code);
```

matildepark: Ah my bad.

PR #6000, opened 2022-09-19
ames: ruin comets not responsive for >6 months

pkova:
The Foundation [has plans](https://github.com/urbit/urbit.org/pull/1794) to provide free short-lived comet hosting. This has the potential to significantly increase the load on popular groups that keep trying to send packets to long-dead comets every two minutes. As far as I understand it the folk wisdom for this has been to `%ruin` comets not seen for six months if needed.

This PR attempts to mitigate the load by `%ruin`ing comets not seen for two weeks in response to a `%trim` event. Two weeks is a lot lower than six months, but this code will only run in response to memory pressure. Also, I heard that [nobody cares](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/jael.hoon#L509).


matildepark:
This would *really* enforce comets as temporary IDs. I think there are probably lots of people who infrequently stop and spin up their comets on their home computers after two weeks and silently failing would look kinda bad. Hmm.

Fang-:
> Also, I heard that [nobody cares](https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/vane/jael.hoon#L509).

To be clear iirc that's supposed to mean "nobody is subscribed to updated for comets", because they can't really breach or change keys, and so nobody (importantly, not ames) has subscriptions open for notifications about the individual comets. "Blasting this out" to all known ducts works around that and ensures ames gets the breach notifications, triggering the cleanup.

Two weeks is indeed very aggressive. I'm also worried about making this run automatically. In particular I'm worried about the DMs and non-chat cases, where longer silences are more likely to be legitimate, and you want communication to keep working across them.

Fang-:
The `%trim` task comes with a priority indicator. We could use that to say something like "only do this for high-priority trims, and scale down the time as the trim priority increases". That way this highly-destructive operation isn't the first line of defense. Of course, the effectiveness of that is contingent on having lower-priority trim handling that actually saves space in most/some cases.

pkova:
I'll go with something else for my use case. Disregarding my comet breaching needs, would this PR be generally useful if I bumped it up to six months?

Fang-:
After talking this over briefly with @philipcmonk: yes, the six-month version of this would be useful, assuming the following change:

In addition to `last-contact.peer-state` being older than six month, none of the entries in `snd.peer-state` may contain entries in `unsent-fragments`.  
By adding that additional filter, we go from "we have not talked to them in six months" to "we have tried and failed to talk to them in six months". In other words, this distinguishes between "inactive" and "offline".

(As for the `%trim` priority levels, we should probably refrain from looking at them until they actually get used, which we'll want to do alongside adding `%trim` handling to all the other vanes that don't have it yet.)

pkova:
Done.

Fang- (changes requested):
The whole structure of this doesn't look quite right to me.

pkova:
Moved the stuff to ames and incorporated the style feedback. 

Note that I didn't push the correct pill here, since `sh/update-solid-pill` doesn't run on my machine and `bin/solid.pill` contains a very different ames from this branch. Let me know if that's a problem.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2136,10 +2137,27 @@
...
+  ++  on-trim
+    ^+  event-core
+    =;  rui=(set @p)
+      (emit duct %pass /ruin %j %ruin rui)
+    =-  (silt (turn - head))
+    %+  skim
+      ~(tap by peers.ames-state)
+    |=  [=ship s=ship-state]
+    ?.  &(=(%pawn (clan:title ship)) =(-.s %known))  %.n
+    ?>  ?=(%known -.s)
```
Fang-, line 2154: Can you not do the `%known` check within the condition?

```suggestion
    ?.  &(=(%pawn (clan:title ship)) ?=(%known -.s))  %.n
```

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2136,10 +2137,27 @@
...
+    ?&  (gth (sub now ~m6) last-contact.qos.s)
+        ::
```
Fang-, line 2156: I've made this mistake before. We have Minutes, but not months. (^:

```suggestion
    ?&  (gth (sub now ~d180) last-contact.qos.s)
      ::
```

Fang- (commented):
Almost there!

re pill: I never use the script, always just build using a fakezod and `.out/pill +solid`, optionally from a non-base desk if you run into migration trouble (which I don't think you should).

pkova:
Fixed the `%known` check and the pill.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2136,10 +2137,27 @@
...
+    =-  (silt (turn - head))
+    %+  skim
+      ~(tap by peers.ames-state)
+    |=  [=ship s=ship-state]
+    ?.  &(?=(%known -.s) =(%pawn (clan:title ship)))  %.n
+    ?&  (gth (sub now ~d180) last-contact.qos.s)
+        ::
+        %+  levy  (turn ~(tap by snd.s) tail)
+        |=  m=message-pump-state
+        ?~(unsent-fragments.m %.y %.n)
```
philipcmonk, line 2159: I believe this condition is reversed, right?  We want to check that unsent-fragments is not null for at least one bone.

I would also tend to use +any:by and ignore the first argument rather than use turn+tap+levy, but that's a stylistic thing.

```suggestion
        %-  ~(any by snd.s)
        |=  [* m=message-pump-state]
        !=(~ unsent-fragments.m)
```

philipcmonk (changes requested):

pkg/arvo/sys/vane/ames.hoon
pkova: Looks like `+any:by` only iterates over the values, so it's even better here. Fixed.

zalberico:
@Fang- I'd also like to get this one over the line, can you take a look and we can get it into a place to merge if that's what we want to do.

Fang- (approved):
Had to look at the condition for like the third time but I'm pretty sure this is indeed the correct implementation of it.

May or may not be slightly easier to read with `(gth (sub now last-contact.qos.s) ~d180)` instead, your call.

But either way this looks good and I'd be happy to release this. Thank you! (:

zalberico:
Looks like there's a conflict and failing test on this, but @Fang- if we can get this passing and green I think we can go ahead and merge it.

joemfb:
The failure is unrelated (just monorepo things).

PR #6001, opened 2022-09-19
u3: add overflow checks to u3r_met()

joemfb:
This PR adds overflow checks (compile time where possible, runtime where not) to the atom measurement function `u3r_met()`. To offset the cost, it also removes so unnecessary branches. These checks could elided if `u3r_met()` returned `c3_d` instead of `c3_w` -- that change should be made, but it'll be more invasive (and have implications for allocation interfaces, which are often driven with measured values).

### DIFF
--- a/pkg/urbit/noun/retrieve.c
+++ b/pkg/urbit/noun/retrieve.c
@@ -1061,10 +1061,7 @@ u3r_met(c3_y  a_y,
         */
         c3_w bif_w, col_w;
 
-        //  .=  35          (add 32 (dec (bex 2))))
-        //  .=  0x7ff.fffe  (rsh [0 5] (sub (bex 32) 35)))
-        //
-        if ( gal_w > 0x7fffffe ) {
+        if ( gal_w > ((UINT32_MAX - 35) >> 5) ) {
           return u3m_bail(c3__fail);
         }
 
--- a/pkg/urbit/noun/retrieve.c
+++ b/pkg/urbit/noun/retrieve.c
@@ -1023,6 +1023,7 @@ u3r_hext(u3_noun  a,
 **   (1 << a_y).
 **
 **   For example, (a_y == 3) returns the size in bytes.
+**   NB: (a_y) must be < 37.
 */
 c3_w
 u3r_met(c3_y  a_y,
--- a/pkg/urbit/noun/retrieve.c
+++ b/pkg/urbit/noun/retrieve.c
@@ -1060,8 +1060,15 @@ u3r_met(c3_y  a_y,
         //
         c3_w bif_w = (gal_w << 5) + c3_bits_word(daz_w) + ((1 << a_y) - 1);
 
+        if ( bif_w < gal_w ) {
+          return u3m_bail(c3__fail);
+        }
+
         return bif_w >> a_y;
       }
+
+      STATIC_ASSERT((UINT32_MAX > ((c3_d)u3a_maximum << 2)),
+                    "met overflow");
 
       case 3: return (gal_w << 2) + ((c3_bits_word(daz_w) +  7) >> 3);
 
--- a/pkg/urbit/noun/retrieve.c
+++ b/pkg/urbit/noun/retrieve.c
@@ -1080,6 +1080,10 @@ u3r_met(c3_y  a_y,
       default: {
         c3_y gow_y = (a_y - 5);
 
+        if ( gow_y > 31 ) {
+          return u3m_bail(c3__fail);
+        }
+
         return ((gal_w + 1) + ((1 << gow_y) - 1)) >> gow_y;
       }
     }
--- a/pkg/urbit/noun/retrieve.c
+++ b/pkg/urbit/noun/retrieve.c
@@ -1056,9 +1056,10 @@ u3r_met(c3_y  a_y,
       case 0:
       case 1:
       case 2: {
-        //  number of bits in [b], rounded up
-        //
-        c3_w bif_w = (gal_w << 5) + c3_bits_word(daz_w) + ((1 << a_y) - 1);
+        /* col_w: number of bits in (daz_w)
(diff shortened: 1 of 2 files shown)


pkg/urbit/noun/retrieve.c
```diff
@@ -1056,24 +1056,24 @@ u3r_met(c3_y    a_y,
...
-        ** bif_w: number of bits in (b)
-        */
-        c3_w bif_w, col_w;
+        //  number of bits in [b], rounded up
+        //
+        c3_w bif_w = (gal_w << 5) + c3_bits_word(daz_w) + ((1 << a_y) - 1);
 
-        col_w = c3_bits_word(daz_w);
-        bif_w = col_w + (gal_w << 5);
+        if ( bif_w < gal_w ) {
```
mcevoypeter, line 1063: Can you explain the motivation behind using `gal_w` as the sentinel for overflow? It strikes me as imprecise, because if I'm following correctly, we should be able to specify the exact maximum value of `gal_w` to avoid overflow.

The largest value that `gal_w` can be without overflow is (call it `x`):
```c
0b0000_0011_1111_1111_1111_1111_1111_1111
```
So `x << 5` is:
```c
0b0111_1111_1111_1111_1111_1111_1110_0000
```
Adding in `32`, which is the greatest number of possible bits in `daz_w`:
```c
0b1000_0000_0000_0000_0000_0000_0000_0000
```
In other words, why aren't we checking that `gal_w` is less than or equal to `0b0000_0011_1111_1111_1111_1111_1111_1111` to ensure that overflow will not occur?

mcevoypeter (changes requested):
Thanks for doing this. I have one comment and one question, but it otherwise looks good to me.

pkg/urbit/noun/retrieve.c
```diff
@@ -1056,24 +1056,24 @@ u3r_met(c3_y    a_y,
...
-              + ((daz_w >> 16) ? 2 : 1);
+        return bif_w >> a_y;
       }
+
+      STATIC_ASSERT((UINT32_MAX > ((c3_d)u3a_maximum << 2)),
+                    "met overflow");
+
+      case 3: return (gal_w << 2) + ((c3_bits_word(daz_w) +  7) >> 3);
+
+      case 4: return (gal_w << 1) + ((c3_bits_word(daz_w) + 15) >> 4);
```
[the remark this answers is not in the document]
joemfb: Interestingly enough, clang and gcc treat these exactly the same: https://godbolt.org/z/fnPfMMWd1.

pkg/urbit/noun/retrieve.c
joemfb: This is a good point, but your numbers aren't quite maximal:

```
> `@ux`(rsh [0 5] (sub (dec (bex 32)) 32))
0x7ff.fffe
> `@ub`(rsh [0 5] (sub (dec (bex 32)) 32))
0b111.1111.1111.1111.1111.1111.1110
```

Or in C

```
(lldb) p/x (((1ULL << 32) - 1) - 32) >> 5
(unsigned long long) $2 = 0x0000000007fffffe
```

Where `(1ULL << 32) - 1` is just `UINT32_MAX`.

joemfb: Actually, this isn't right either. We're rounding up, so the maximum addend isn't `32`, it's `35` (`32 + (1 << 2) - 1`).

pkg/urbit/noun/retrieve.c
```diff
@@ -1056,24 +1056,24 @@ u3r_met(c3_y    a_y,
       case 0:
       case 1:
       case 2: {
-        /* col_w: number of bits in (daz_w)
-        ** bif_w: number of bits in (b)
-        */
-        c3_w bif_w, col_w;
+        //  number of bits in [b], rounded up
+        //
+        c3_w bif_w = (gal_w << 5) + c3_bits_word(daz_w) + ((1 << a_y) - 1);
```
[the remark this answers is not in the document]
joemfb: I tried using `c3_rop()` everywhere, and CI promptly failed. I didn't see anything obviously wrong with what I'd written, but the older way is more intuitive (and less of a change from the previous implementation), not to mention correct. So I reverted.

pkg/urbit/noun/retrieve.c
joemfb: I'd normally use the constant here with a comment, but both gcc and clang constant-fold it, even at `-O0`.

pkg/urbit/noun/retrieve.c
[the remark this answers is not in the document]
mcevoypeter: Just out of curiosity, what did the `c3_rop()` call sites look like that lead to the CI failures?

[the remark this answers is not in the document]
joemfb: `jb/met-with-rop`

pkg/urbit/noun/retrieve.c
```diff
@@ -1061,22 +1061,30 @@ u3r_met(c3_y    a_y,
...
+      case 3: return (gal_w << 2) + ((c3_bits_word(daz_w) +  7) >> 3);
+
+      case 4: return (gal_w << 1) + ((c3_bits_word(daz_w) + 15) >> 4);
+
       default: {
         c3_y gow_y = (a_y - 5);
 
+        if ( gow_y > 31 ) {
+          return u3m_bail(c3__fail);
+        }
```
joemfb, line 1086: On second thought, adding this kind of check feels like a slippery slope. We'll never hit it in practice through a jet (all user-proved bloqs must be checked to be less than 32 already), so it's only guarding against a statically invalid call. Such invariants already abound, and this one is particularly absurd ("i want the length of this atom in terms of `1 << 37`- bit chunks"). Removing in favor of documenting the invariant.

Issue #880, opened 2017-09-04
More on urbit on memory constrained devices.


mikolajpp:
Hi,

It is me again, all talk and no code for as yet ;-) I am figuring out how to best approach "Urbit on android" (code name urbot, see if you like it).
Would it make sense to make the memory limit a dynamic one, set on startup? 
I am not sure if it is even possible, as for now it is a compile time constant. 
Basically I would like to provide user with the ability to choose how much of memory hog his urbit should be, and also enable a wider range of users to run it.

mparaniak:
On a closer look on urbit's u3, it seems entirely infeasible, as the entire memory scheme rests on it. 

Another idea is to have memory-tier system for urbits (urbot does not sound nice, after all):
^urbit - keturbit - low tier devices 100 MB 
%urbit - cenurbit - mid tier 300 MB
@urbit - paturbit - high end 600 MB

Of course in the future we would get
urbit - plain urbit 2000MB as the phone memory get better and better.
 
I plan to do some memory profiling on vanilla urbit and see 
how much it really needs, look at the peak memory, and how memory changes 
under different loads.

chc4:
Making the Loom size a global variable that is set by a launcher before it boots urbit sounds feasible. I think that changing it after you first boot your ship would horribly break, though, because then the checkpoint would no longer fit in memory. I don't actually know anything about how the check pointing works, however.

Getting it working with a static memory size is the important thing, tbh. Everything after that is just a bonus.

cgyarvin:
There are a number of things in the code that assume a fixed (set at
compile time) loom bitwidth and position.  The trouble is that these
constants get used so frequently that you don't even want to think about
hitting memory to pull them out.  You could probably change this to use a
macro indirection.  This would require a pretty solid understanding of the
u3 doc and also strong C skills, but I don't see it as at all impossible.
And I still want most architectures to keep them static.

I really don't think of vere's loom as consuming memory -- I think of it as
consuming virtual address space.  If I'm a process on a 32-bit chip I
should be able to map 2GB, regardless of how much I use.  The question of
the mapping between them is just a question of a working set (though, Unix
kernels being as dumb as they are, I think you could make vere use a lot
less address space by marking junk memory not just with mprotect(), but
also with munmap().  I do realize that this is the attitude of an OS purist
and is probably misplaced in the 21st century.

BTW, I doubt Graal/Truffle plays nicely with Android (Java is one thing,
Java is another) but the jaque G/T interpreter is supposedly coming along
[...]

PR #6002, opened 2022-09-19
u3: allow the loom size to be set at runtime

joemfb:
The PR makes the loom memory-map size a runtime rather than compile-time configuration. The compile-time constants are still there (`u3a_bits`/bytes/words/cells/pages &c), but those are now representational upper-bounds instead of exact limits.

Next steps for this PR include adding commandline flags to set the loom size for both mars and urth processes, and bumping those compile-time upper bounds where possible (which would make #6001 more important).

These changes will need to be reconciled with #5701.

See #880

### DIFF
--- a/pkg/urbit/include/noun/options.h
+++ b/pkg/urbit/include/noun/options.h
@@ -9,7 +9,6 @@
         u3_noun who;                          //  single identity
         c3_c*   dir_c;                        //  execution directory (pier)
         c3_w    wag_w;                        //  flags (both ways)
-        size_t  byt_i;                        //  loom byte-length (<= u3a_bytes)
         size_t  wor_i;                        //  loom word-length (<= u3a_words)
         void (*stderr_log_f)(c3_c*);          //  errors from c code
         void (*slog_f)(u3_noun);              //  function pointer for slog
--- a/pkg/urbit/include/noun/options.h
+++ b/pkg/urbit/include/noun/options.h
@@ -9,6 +9,8 @@
         u3_noun who;                          //  single identity
         c3_c*   dir_c;                        //  execution directory (pier)
         c3_w    wag_w;                        //  flags (both ways)
+        size_t  byt_i;                        //  loom byte-length (<= u3a_bytes)
+        size_t  wor_i;                        //  loom word-length (<= u3a_words)
         void (*stderr_log_f)(c3_c*);          //  errors from c code
         void (*slog_f)(u3_noun);              //  function pointer for slog
         void (*sign_hold_f)(void);            //  suspend system signal regime
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -699,7 +699,7 @@ _ce_image_resize(u3e_image* img_u, c3_w pgs_w)
 {
   if ( img_u->pgs_w > pgs_w ) {
     if ( ftruncate(img_u->fid_i, pgs_w << (u3a_page + 2)) ) {
-      fprintf(stderr, "loom: image truncate %s: %s\r\n",
+      fprintf(stderr, "loom: image (%s) truncate: %s\r\n",
                       img_u->nam_c,
                       strerror(errno));
       c3_assert(0);
@@ -785,7 +785,8 @@ _ce_image_blit(u3e_image* img_u,
   lseek(img_u->fid_i, 0, SEEK_SET);
   for ( i_w = 0; i_w < img_u->pgs_w; i_w++ ) {
     if ( -1 == read(img_u->fid_i, ptr_w, siz_w) ) {
-      fprintf(stderr, "loom: image blit read: %s\r\n", strerror(errno));
+      fprintf(stderr, "loom: image (%s) blit read: %s\r\n",
+                      img_u->nam_c, strerror(errno));
       c3_assert(0);
     }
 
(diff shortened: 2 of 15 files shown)


pkg/urbit/noun/events.c
```diff
@@ -255,9 +255,9 @@ u3e_fault(void* adr_v, c3_i ser_i)
 
   c3_w* adr_w = (c3_w*) adr_v;
 
-  if ( (adr_w < u3_Loom) || (adr_w >= (u3_Loom + u3a_words)) ) {
+  if ( (adr_w < u3_Loom) || (adr_w >= (u3_Loom + u3C.wor_i)) ) {
```
joemfb, line 258: This line is the only change that I find potentially worrisome. We could mark `u3C.wor_i` as `volatile`, but I don't think we should have to since it's only ever set once at initialization.

pkg/urbit/noun/manage.c
```diff
@@ -1713,70 +1718,136 @@ _cm_signals(void)
...
-  mp_set_memory_functions(u3a_malloc, u3a_realloc2, u3a_free2);
+  //  make sure GMP uses our malloc.
+  //
+  mp_set_memory_functions(u3a_malloc, _cm_realloc2, _cm_free2);
 
-  /* Map at fixed address.
-  */
+  //  make sure that [len_i] is a fully-addressible non-zero power of two.
+  //
+  if (  !len_i
```
mcevoypeter, line 1806: Nit, but can we change `!len_i` to `0 != len_i`.

pkg/urbit/noun/manage.c
```diff
@@ -1713,70 +1718,136 @@ _cm_signals(void)
...
 
-    u3l_log("loom: mapped %dMB\r\n", len_w >> 20);
+    u3C.byt_i = len_i;
+    u3C.wor_i = len_i >> 2;
+    u3l_log("loom: mapped %zuMB\r\n", len_i >> 20);
   }
 }
 
-/* u3m_stop(): graceful shutdown cleanup. */
+extern void u3je_secp_stop(void);
```
mcevoypeter, line 1847: Why is this function declaration needed? I'm not seeing a call to `u3je_secp_stop()` below this line in this file in the diff.

pkg/urbit/include/noun/options.h
```diff
@@ -9,6 +9,8 @@
         u3_noun who;                          //  single identity
         c3_c*   dir_c;                        //  execution directory (pier)
         c3_w    wag_w;                        //  flags (both ways)
+        size_t  byt_i;                        //  loom byte-length (<= u3a_bytes)
+        size_t  wor_i;                        //  loom word-length (<= u3a_words)
```
mcevoypeter, line 13: Because these are redundant, I'd opt to remove `byt_i` and instead compute it using `wor_i` since `byt_i` only appears to be used in `u3e_yolo()`.

mcevoypeter (changes requested):
Good stuff. I have only a few comments, and those comments aside, it looks good to me.

pkg/urbit/noun/manage.c
joemfb: I just moved it down so that it's right above the one call to it in `u3m_stop()`.

pkg/urbit/noun/manage.c
joemfb: We could (well, `0 == len_i`), but this is idiomatic C. (I'm responsible for many of the explicit comparisons to zero in this codebase -- those stemmed from a misplaced fear of loobeans. I regret the error.)

pkg/urbit/include/noun/options.h
joemfb: Good point. I had included to avoid the question of overflow checks when shifting words to bytes, but it's always initialized from bytes so that's irrelevant.

joemfb:
Other features depend on IPC changes not yet in this branch. Rather than merge them all in, I'll just make those in a new PR.

pkg/urbit/noun/manage.c
mcevoypeter: Fair, but it's slightly more obvious to use `0 == len_i` since we're compare `len_i` to `1 << (u3a_page + 2)` and `u3a_bytes` anyway, so the pattern of "compare `len_i` to some number using a binary relation" is already established. Like I said though, it's a nit, and it doesn't matter much either way.

joemfb: The second conditional is actually `(len_i & (len_i - 1))` -- "if any bits other than the most significant are set". The whole conditional is checking for violations of our map-size requirements -- size is zero, or not a power of two, or less than our page size, or greater than our compile-time maximum.

joemfb: There are many things that are hard in C, and some of them are avoidable. I'm don't think it's reasonable to place the standard boolean conversions are in either set (which is why I said I regret the error).

Loobeans are another story, but I think the solution is just to quarantine them to the interpreter and relevant jets.

pkg/urbit/noun/events.c
mcevoypeter: I agree that this should be fine. Unless there's a `SEGFAULT` between `u3e_live()` and `u3m_pave()` in `u3m_boot()`, the fault handler won't run until after `u3C.wor_i` has been initialized. We already access `u3P.dit_w` in the fault handler, and that use case is read/write, not read-only like `u3C.wor_i`.

mcevoypeter (approved):
LGTM.

PR #6003, opened 2022-09-19
c3,u3: read() and write() to completion

mcevoypeter:
### Motivation

This PR replaces all instances of `read()` and `write()` (outside of the `compat` directory) with `c3_read()` and `c3_write()`. `c3_read()`/`c3_write()` are wrappers around `read()`/`write()` that correctly handle partial reads/writes. They also handle retries in the case where `read()`/`write()` returns `EINTR`, `EAGAIN`, or `EWOULDBLOCK` up to a specified maximum number of attempts. 

Incorrect handling of partial reads/writes in `events.c` is the suspected cause of our lingering snapshot corruption bug, which has tended to appear when disk space fills up. Additional work will be required to gracefully handle failed calls to `c3_read()` and `c3_write()` in `events.c`, but that work is beyond the scope of this PR.

The inclusion of `c3_read()` and `c3_write()` has the added benefit of rendering `u3_write_fd()` in `term.c` and a big chunk of `_cu_rock_save()` in `urth.c` obsolete. Both of these blocks of code have been removed.

### Test plan

To test, I booted a fake ship, exited, deleted its incremental snapshot, and then rebooted. Replay and terminal output work as expected.
```console
$ urbit -F zod -B solid.pill
$ rm zod/.urb/chk/*.bin
$ urbit zod
```

### DIFF
--- a/pkg/urbit/include/c/defs.h
+++ b/pkg/urbit/include/c/defs.h
@@ -1,6 +1,11 @@
 #ifndef C3_DEFS_H
 #define C3_DEFS_H
 
+#include "c/portable.h"
+#include "c/types.h"
+
+#include <errno.h>
+
   /** Loobeans - inverse booleans to match nock.
   **/
 #     define c3y      0
@@ -156,5 +161,123 @@
         unlink(a);})
 #     define c3_fopen(a, b) ({                                  \
         fopen(a, b);})
+
+      /// Reads the contents of a file into memory.
+      ///
+      /// This function correctly handles the case in which read() reads fewer
+      /// bytes than was requested by retrying until all bytes have been read or
+      /// until an error occurs, whichever comes first.
+      ///
+      /// If interrupted by a signal or if the read() call would block, this
+      /// function will retry up to a specified maximum number of attempts.
+      ///
+      /// @param[in]  fd_i        Open file descriptor to read from.
+      /// @param[out] data_v      Destination buffer.
+      /// @param[in]  data_len_i  Length of `data_v` in bytes.
+      ///
+      /// @return >0  Number of bytes read. Guaranteed to be less than or equal
+      ///             to `data_len_i`. If less than `data_len_i`, then `fd_i`
+      ///             reached EOF.
+      /// @return  0  `fd_i` is at EOF.
+      /// @return <0  Error occurred. The error number is the absolute value of
+      ///             the return value and can be fed into strerror().
+      inline ssize_t
+      c3_read(c3_i fd_i, void* const data_v, const size_t data_len_i)
+      {
+        static const size_t max_attempts_i    = 100;
+        size_t              attempts_i        = 0;
+        c3_y*               ptr_y             = data_v;
+        ssize_t             bytes_remaining_i = data_len_i;
+
+        off_t cur_offset_i = lseek(fd_i, 0, SEEK_CUR);
+        if ( cur_offset_i < 0 ) {
+          return -errno;
+        }
+
+        do {
+          attempts_i++;
+          ssize_t bytes_read_i = read(fd_i, ptr_y, bytes_remaining_i);
+          if ( -1 == bytes_read_i ) {
+            if ( (EINTR == errno || EAGAIN == errno || EWOULDBLOCK == errno)
+                 && attempts_i <= max_attempts_i )
+            {
+              // From the read(2) man page: "On error, -1 is returned, and errno
+              // is set to indicate the error. In this case, it is left
+              // unspecified whether the file position (if any) changes."
+              //
+              // This means that we have to restore the file cursor to the
+              // position that it was at immediately before the call to read()
+              // that failed.
+              if ( lseek(fd_i, SEEK_SET, cur_offset_i) < 0 ) {
+                return -errno;
+              }
+              continue;
+            }
+            else {
+              return -errno;
+            }
+          }
+          else if ( 0 == bytes_read_i ) {
+            break;
+          }
+          else {
+            ptr_y += bytes_read_i;
+            bytes_remaining_i -= bytes_read_i;
+            cur_offset_i += bytes_read_i;
+            attempts_i = 0;
+          }
+        } while ( bytes_remaining_i > 0 );
+
+        return data_len_i - bytes_remaining_i;
+      }
+
+      /// Writes the contents of a buffer to a file.
+      ///
+      /// This function correctly handles the case in which write() writes fewer
+      /// bytes than was requested by retrying until all bytes have been written
+      /// or until an error occurs, whichever comes first.
+      ///
+      /// If interrupted by a signal or if the write() call would block, this
+      /// function will retry up to a specified maximum number of attempts.
+      ///
+      ///
+      /// @param[in] fd_i        Open file descriptor to write to.
+      /// @param[in] data_v      Source buffer.
+      /// @param[in] data_len_i  Length of `data_v` in bytes.
+      ///
+      /// @return >=0  Number of bytes written. Guaranteed to be equal to
+      ///              `data_len_i`.
+      /// @return <0   Error occurred. The error number is the absolute value of
+      ///              the return value and can be fed into strerror().
+      inline ssize_t
+      c3_write(c3_i fd_i, const void* const data_v, const size_t data_len_i)
+      {
+        static const size_t max_attempts_i    = 100;
+        size_t              attempts_i        = 0;
+        const c3_y*         ptr_y             = data_v;
+        ssize_t             bytes_remaining_i = data_len_i;
+        do {
+          attempts_i++;
+          ssize_t bytes_written_i = write(fd_i, ptr_y, bytes_remaining_i);
+          if ( -1 == bytes_written_i ) {
+            // Attempt to read again if we were interrupted by a signal.
+            if ( (EINTR == errno || EAGAIN == errno || EWOULDBLOCK == errno)
+                 && attempts_i <= max_attempts_i )
+            {
+              continue;
+            }
+            else {
+              return -errno;
+            }
+          }
+          else {
+            ptr_y += bytes_written_i;
+            bytes_remaining_i -= bytes_written_i;
+            attempts_i = 0;
+          }
+        } while ( bytes_remaining_i > 0 );
+
+        return data_len_i - bytes_remaining_i;
+      }
 
 #endif /* ifndef C3_DEFS_H */
--- a/pkg/urbit/include/c/defs.h
+++ b/pkg/urbit/include/c/defs.h
@@ -181,55 +181,8 @@
       /// @return  0  `fd_i` is at EOF.
       /// @return <0  Error occurred. The error number is the absolute value of
       ///             the return value and can be fed into strerror().
-      inline ssize_t
-      c3_read(c3_i fd_i, void* const data_v, const size_t data_len_i)
-      {
-        static const size_t max_attempts_i    = 100;
-        size_t              attempts_i        = 0;
-        c3_y*               ptr_y             = data_v;
-        ssize_t             bytes_remaining_i = data_len_i;
-
-        off_t cur_offset_i = lseek(fd_i, 0, SEEK_CUR);
-        if ( cur_offset_i < 0 ) {
-          return -errno;
-        }
-
-        do {
-          attempts_i++;
-          ssize_t bytes_read_i = read(fd_i, ptr_y, bytes_remaining_i);
-          if ( -1 == bytes_read_i ) {
-            if ( (EINTR == errno || EAGAIN == errno || EWOULDBLOCK == errno)
-                 && attempts_i <= max_attempts_i )
-            {
-              // From the read(2) man page: "On error, -1 is returned, and errno
-              // is set to indicate the error. In this case, it is left
-              // unspecified whether the file position (if any) changes."
-              //
-              // This means that we have to restore the file cursor to the
-              // position that it was at immediately before the call to read()
-              // that failed.
-              if ( lseek(fd_i, SEEK_SET, cur_offset_i) < 0 ) {
-                return -errno;
-              }
-              continue;
-            }
-            else {
-              return -errno;
-            }
-          }
-          else if ( 0 == bytes_read_i ) {
-            break;
-          }
-          else {
-            ptr_y += bytes_read_i;
-            bytes_remaining_i -= bytes_read_i;
-            cur_offset_i += bytes_read_i;
-            attempts_i = 0;
-          }
-        } while ( bytes_remaining_i > 0 );
-
-        return data_len_i - bytes_remaining_i;
-      }
+      ssize_t
+      c3_read(c3_i fd_i, void* const data_v, const size_t data_len_i);
 
       /// Writes the contents of a buffer to a file.
       ///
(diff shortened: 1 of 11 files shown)


pkg/urbit/include/c/defs.h
```diff
@@ -157,4 +159,108 @@
...
+              fprintf(stderr,
+                      "read: failed after %lu attempts: %s\r\n",
+                      attempts_i,
+                      strerror(errno));
+              return 0;
+            }
+          }
+          else if ( 0 == bytes_read_i ) {
+            break;
+          }
```
joemfb, line 203: Needing to handle `EOF` here makes me concerned that this signature is insufficient. Reads may correctly produce less than the requested number of bytes, but this function swallows that measurement, which could lead to bogus data being propagated in the target buffer. I'm not sure what the best solution is. The function could return `-errno` on error, `0` on a successful read of the amount requested, or some positive integer on a successful but truncated (due to `EOF`) read (with the invariant that otherwise partial reads have already been retried). Or it could just match the `read()` signature.

Also, almost all calls to this function, already print on error. It seems like this function should either propagate/preserve `errno` for those call sites, or additional take a caption string as a parameter and interpolate that into the error printf.

joemfb: One other point about this signature, from the man page:

>  On error, -1 is returned, and errno is set appropriately. In this case it is left unspecified whether the file position (if any) changes.

That's pretty shocking to me. Combining it with the performance argument in #5889 makes me wonder if we should just always use `pread()` in cases where offsets apply. Otherwise, seems like we would logically need an `lseek()` in this function when we retry on error and the fd has an offset, and I don't know how we would distinguish that case. What do you think?

pkg/urbit/include/c/defs.h
```diff
@@ -157,4 +159,108 @@
...
+              fprintf(stderr,
+                      "write: failed after %lu attempts: %s\r\n",
+                      attempts_i,
+                      strerror(errno));
+              return 0;
+            }
+          }
+          else if ( 0 == bytes_written_i ) {
+            break;
+          }
```
joemfb, line 256: This seems wrong, I think it should just be `continue;` (but could also be removed in favor of the else block below).

pkg/urbit/noun/manage.c
```diff
@@ -113,17 +113,15 @@ _cm_punt(u3_noun tax)
...
 
   if ( sig_l ) {
-    ret_i = write(2, ": ", 2);
-    ret_i = write(2, &sig_l, 4);
+    c3_write(2, ": ", 2);
+    c3_write(2, &sig_l, 4);
   }
 
-  ret_i = write(2, "\r\n", 2);
+  c3_write(2, "\r\n", 2);
```
joemfb, line 124: This function is an "emergency printf", called from within arbitrary (ie, async) signal handlers. I don't think we need or want retry here.

pkg/urbit/noun/urth.c
```diff
@@ -566,50 +566,7 @@ _cu_rock_save(c3_c* dir_c, c3_d eve_d, c3_d len_d, c3_y* byt_y)
...
-      }
-      //  continue partial writes
-      //
-      else {
-        len_d -= ret_i;
-        byt_y += ret_i;
-      }
-    }
-  }
+  c3_assert(c3_write(fid_i, byt_y, len_d));
```
joemfb, line 569: This should not assert no errors, but should handle them with `close(fid_i); return c3n;`.

pkg/urbit/vere/king.c
```diff
@@ -1519,7 +1456,7 @@ _king_copy_file(c3_c* src_c, c3_c* dst_c)
     {
       size_t pag_i = 1 << 14;;
       c3_y*  buf_y = c3_malloc(pag_i);
-      ret_i = _king_copy_raw(src_i, dst_i, buf_y, pag_i);
+      ret_i = c3_read(src_i, buf_y, pag_i) && c3_read(dst_i, buf_y, pag_i) ? 0 : -1;
```
joemfb, line 1459: This needs to be `c3_read()` and `c3_write()`, reformatted to stay within 80-column lines.

joemfb (changes requested):
This is a great catch, and should lead to vastly more reliable snapshots.

I have some concerns about these wrapper-function signatures, and a couple of call sites that need to be tweaked:

pkg/urbit/vere/king.c
mcevoypeter: Good catch. Addressed both the function call and formatting.

pkg/urbit/noun/urth.c
mcevoypeter: Thanks for catching this.

pkg/urbit/include/c/defs.h
```diff
@@ -157,4 +159,108 @@
...
+      /// @param[in]  data_len_i  Length of `data_v` in bytes.
+      ///
+      /// @return 1  Complete read succeeded.
+      /// @return 0  Otherwise.
+      static inline c3_t
+      c3_read(c3_i fd_i, void* const data_v, const size_t data_len_i)
+      {
+        static const size_t max_attempts_i    = 100;
+        size_t              attempts_i        = 0;
+        c3_y*               ptr_y             = data_v;
```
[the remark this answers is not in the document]
mcevoypeter: `<type> *const <name>`  indicates that the pointer identified by `<name>` will not change i.e. `<name>++`, `<name> += 3`, etc are all invalid. So declaring `data_v` as `void* const` indicates to the caller that we're not performing arithmetic on `data_v`. In other words, the `const`ness is tied to the variable name, not to the data that the variable points at, and as a result, creating a new variable that points to the same address does not result in discarding the `const` qualifier.

pkg/urbit/include/c/defs.h
```diff
@@ -157,4 +159,108 @@
...
+      /// @param[in] data_len_i  Length of `data_v` in bytes.
+      ///
+      /// @return 1  Complete write succeeded.
+      /// @return 0  Otherwise.
+      static inline c3_t
+      c3_write(c3_i fd_i, const void* const data_v, const size_t data_len_i)
+      {
+        static const size_t max_attempts_i    = 100;
+        size_t              attempts_i        = 0;
+        const c3_y*         ptr_y             = data_v;
```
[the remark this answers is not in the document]
mcevoypeter: https://github.com/urbit/urbit/pull/6003#discussion_r974756029

pkg/urbit/include/c/defs.h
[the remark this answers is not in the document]
mcevoypeter: So you could argue that it's unnecessary to declare `data_v` as `void* const`, but it provides just a little more context to whoever is reading the function definition.

pkg/urbit/include/c/defs.h
```diff
@@ -157,4 +159,108 @@
...
+      /// If interrupted by a signal or if the read() call would block, this
+      /// function will retry up to a specified maximum number of attempts.
+      ///
+      /// @param[in]  fd_i        Open file descriptor to read from.
+      /// @param[out] data_v      Destination buffer.
+      /// @param[in]  data_len_i  Length of `data_v` in bytes.
+      ///
+      /// @return 1  Complete read succeeded.
+      /// @return 0  Otherwise.
+      static inline c3_t
```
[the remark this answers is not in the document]
mcevoypeter: The arguments seem in that post seem reasonable to me too. I didn't realize that `static inline` generated a separate compilation of the function each time the header it's defined in was included. Thanks for drawing attention to this.

pkg/urbit/include/c/defs.h
```diff
@@ -157,4 +159,108 @@
...
+      ///
+      ///
+      /// @param[in] fd_i        Open file descriptor to write to.
+      /// @param[in] data_v      Source buffer.
+      /// @param[in] data_len_i  Length of `data_v` in bytes.
+      ///
+      /// @return 1  Complete write succeeded.
+      /// @return 0  Otherwise.
+      static inline c3_t
+      c3_write(c3_i fd_i, const void* const data_v, const size_t data_len_i)
```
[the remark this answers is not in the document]
mcevoypeter: Good point about lining up the `c3_write()` signature to match the `c3_read()` signature.

pkg/urbit/include/c/defs.h
mcevoypeter: > Or it could just match the read() signature.

This seems like the cleanest solution. So `-1` will indicate failure (and `errno` will indicate the specific failure), `0` will indicate that `fd_i` was already at EOF, and `>0` will indicate the number of bytes read into `data_v` (which is guaranteed to be in the range `(0, data_len_i]`.

mcevoypeter: > On error, -1 is returned, and errno is set appropriately. In this case it is left unspecified whether the file position (if any) changes.

Wow, this is terrible.

pkg/urbit/include/c/defs.h
mcevoypeter: You're right. I've removed this `else if`.

pkg/urbit/include/c/defs.h
mcevoypeter: I updated `c3_read()` to track the current offset and reset it to its pre-`read()` fail value when a call to `read()` failed. The only added cost for a call to `c3_read()` for which no `read()` calls fail is a single `lseek()` at the beginning used to compute the current offset (which itself should be relatively cheap).

joemfb: That looks like it would work correctly for files that we read entirely, but it significantly limits the applicability of this wrapper function. At the very least, we now have a bunch of redundant `lseek()` calls in `events.c`. And a bunch of refactored call sites in this PR will be wrong, see `ESPIPE` in the `lseek()` man-page.

pkg/urbit/vere/io/unix.c
```diff
@@ -1176,21 +1176,21 @@ _unix_initial_update_file(c3_c* pax_c, c3_c* bas_c)
...
+  if ( red_i != len_ws ) {
+    if ( red_i < 0 ) {
       u3l_log("error reading initial file %s: %s\r\n",
-              pax_c, strerror(errno));
+              pax_c, strerror(-red_i));
     }
     else {
-      u3l_log("wrong # of bytes read in initial file %s: %d %d\r\n",
-              pax_c, len_ws, red_ws);
+      u3l_log("wrong # of bytes read in initial file %s: %d %lu\r\n",
```
joemfb, line 1192: I think all these partial read printfs (here and in `events.c`) should use the `%zu` format specifier.

pkg/urbit/vere/king.c
```diff
@@ -1519,8 +1460,11 @@ _king_copy_file(c3_c* src_c, c3_c* dst_c)
     {
       size_t pag_i = 1 << 14;;
       c3_y*  buf_y = c3_malloc(pag_i);
-      ret_i = _king_copy_raw(src_i, dst_i, buf_y, pag_i);
-      err_i = errno;
+      ret_i        = c3_read(src_i, buf_y, pag_i) == pag_i
+                  && c3_write(dst_i, buf_y, pag_i) == pag_i
+                       ? 0
+                       : -1;
+      err_i        = errno;
```
joemfb, line 1467: I missed this in my earlier review, but this change is just wrong. `pag_i` is just a temporary buffer for the copy, not the length of data to be copied, so we still need the loop. `_king_copy_raw()` should be restored, to something like:

```
static c3_i
_king_copy_raw(c3_i src_i, c3_i dst_i, c3_y* buf_y, size_t pag_i)
{
  ssize_t ret_i;

  do {
    if ( 0 > (ret_i = c3_read(src_i, buf_y, pag_i)) ) {
      return ret_i;
    }

    if ( 0 > (ret_i = c3_write(dst_i, buf_y, (size_t)ret_i)) ) {
      return ret_i;
    }
  }
  while ( ret_i );

  return 0;
}
```

and the callsite here should be something like:

```
if ( 0 > (ret_i = _king_copy_raw(src_i, dst_i, buf_y, pag_i)) ) {
  err_i = -ret_i;
  ret_i = -1;
}
```

joemfb (changes requested):
This looks like the right approach. A couple more things:

pkg/urbit/include/c/defs.h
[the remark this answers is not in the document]
joemfb: I understand the what, but not the why. Conveying to callers that data behind the pointer can/will not be modified seems very valuable, but conveying that the pointer itself will not be changed (because it will be copied first) seems irrelevant. Callers cannot be affected either way.

I've used `const` on scalars in function signatures, as a correctness constraint for the implementation. Even that  seems like it leaks some implementation details (again, because callers can't be impacted either way). But this seems like unnecessary (or even misleading) ceremony.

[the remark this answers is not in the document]
mcevoypeter: This convention adheres to a school of thought in which anything that can be marked `const` should be. The pointer doesn't need to be modified in the function, and so it should be marked `const` to convey that. One word for this is ceremony, but another is discipline. We don't lose anything by marking `data_v` as `void* const` rather than `void*`, but we gain a more explicit explanation of how `data_v` is used.

pkg/urbit/include/c/defs.h
mcevoypeter: After speaking live, we've decided to require that `c3_read()` be called with a seekable file.

pkg/urbit/include/c/defs.h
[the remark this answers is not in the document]
joemfb: But it can only be marked `const` because it's copied before it's modified -- by that logic, everything can be `const` (otherwise known as SSA form)! So it's still not clear to me why there is value in what is being conveyed here.

pkg/urbit/vere/king.c
```diff
@@ -1516,12 +1457,28 @@ _king_copy_file(c3_c* src_c, c3_c* dst_c)
...
+        ret_i = c3_read(src_i, buf_y, sizeof(buf_y));
+        if ( ret_i < 0 ) {
+          break;
+        }
+
+        if ( ret_i == 0 ) {
+          break;
+        }
+
+        ret_i = c3_write(dst_i, buf_y, sizeof(buf_y));
```
joemfb, line 1472: This is still not right; there's no guarantee that the source file is a multiple of any particular buffer size. So the size of the write needs to be specified as the size of the read `(size_t)ret_i`, at least to correctly handle the tail of the file and avoid writing garbage.

joemfb (changes requested):

pkg/urbit/vere/king.c
mcevoypeter: Thanks. Hopefully take three is correct.

pkg/urbit/include/c/defs.h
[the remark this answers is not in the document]
mcevoypeter: Would it feel better if I implemented `c3_read()` using a base pointer and an offset rather than a current pointer? Because in that case, then the pointer wouldn't be modified, and this argument against labeling `data_v` as `const` would no longer be relevant.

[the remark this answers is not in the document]
joemfb: No, the question I've put here in various forms would still remain, unanswered. "If you followed the rule, and it brought you to 'manually convert your code to SSA-form', of what use was the rule?".

More here: https://softwareengineering.stackexchange.com/questions/204500/when-and-for-what-purposes-should-the-const-keyword-be-used-in-c-for-variables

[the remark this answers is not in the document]
mcevoypeter: Pointer variables marked `<type>* const` cannot be reallocated/freed without a cast. Because `data_v` is marked as an output parameter in the interface documentation of `c3_read()`, you could argue that it's obvious that `data_v` won't be reallocated/freed by the `c3_read()` implementation, but marking `data_v` as `const` in this way explains to the caller that the code encapsulates the notion of "`data_v` will not be reallocated/freed", thereby reinforcing the guarantee made by the documentation. Of course, the `const` label *can* easily be overcome by a cast or by creating a new variable (as is done here), but it's unlikely that anyone modifying the implementation of `c3_read()` will be able to overlook both the interface documentation and the fact that `data_v` is marked `const` to reallocate/free the memory owned by `data_v`. In other words, I see absolutely no reason why the `const` label should be omitted given that it communicates valuable information to both the caller and the implementer.

PR #6004, opened 2022-09-20
u3: optimizes u3i_edit (nock opcode 10)

joemfb:
This PR optimizes our implementation of nock opcode `%10` (tree edits, `u3i_edit()`), turning it into a loop (in the style of the "tail recursion modulo cons" optimization) and removing head/tail branches.

These changes are not urgent, but I've had some version of them floating around for awhile now and have meant to open this PR. I was inspired to rewrite this after seeing @eamsden's implementation of tree edits. While I have not benchmarked these changes in detail, I also haven't noticed much of a difference with them. But the changes still seem worthwhile, even if this is not currently a significant hot spot.

These commits are probably best reviewed individually, in order.

/cc @frodwith

### DIFF
--- a/pkg/urbit/noun/imprison.c
+++ b/pkg/urbit/noun/imprison.c
@@ -610,6 +610,7 @@ u3i_edit(u3_noun big, u3_noun axe, u3_noun som)
 
       do {
         u3a_cell*  big_u = u3a_to_ptr(big);
+        u3_noun*     old = (u3_noun*)&(big_u->hed);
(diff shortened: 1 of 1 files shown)


eamsden:
For reference, this is the edit implementation in New Mars: https://github.com/urbit/new-mars/blob/892a490fca1aa19499355ef30115bdf714b6f50e/rust/iron-planet/src/interpreter.rs#L530-L572

jalehman:
@joemfb because this code now targets urbit/vere, I'm going to close it. Please re-open targeting the new repo.

PR #6006, opened 2022-09-20
vere: meld on memory pressure, if --auto-meld

Fang-:
We currently automatically pack the memory when we start running out of space. We have avoided automatically melding, because it may be slow and require a significant amount of memory on the host machine.

Here, we add an --auto-meld flag that enables automatic melding, so that it can be opted into in situations/setups where it makes sense.

Have had some trouble testing this properly, but did get the auto-meld to run at some point.

### DIFF
--- a/pkg/urbit/worker/mars.c
+++ b/pkg/urbit/worker/mars.c
@@ -289,7 +289,10 @@ _mars_sure_feck(u3_mars* mar_u, c3_w pre_w, u3_noun vir)
 
   mar_u->rec_o = c3o(mar_u->rec_o, rec_o);
   mar_u->pac_o = c3o(mar_u->pac_o, pac_o);
-  mar_u->mel_o = c3a(u3_Host.ops_u.mel, c3o(mar_u->mel_o, mel_o));
+
+  if ( u3C.wag_w & u3o_auto_meld ) {
+    mar_u->mel_o = c3o(mar_u->mel_o, mel_o);
+  }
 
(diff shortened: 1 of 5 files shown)


pkg/urbit/worker/mars.c
```diff
@@ -287,6 +289,7 @@ _mars_sure_feck(u3_mars* mar_u, c3_w pre_w, u3_noun vir)
 
   mar_u->rec_o = c3o(mar_u->rec_o, rec_o);
   mar_u->pac_o = c3o(mar_u->pac_o, pac_o);
+  mar_u->mel_o = c3a(u3_Host.ops_u.mel, c3o(mar_u->mel_o, mel_o));
```
joemfb, line 292: `u3_Host.ops_u` isn't propagated from urth to mars. This "works" because a loobean "bunts" to `c3y`, given zero-initialization.

To be correctly propagated, `--auto-meld` should become a bit in `u3C.wag_w`.

joemfb (changes requested):

pkg/urbit/worker/mars.c
joemfb: This needs to be something like `(u3C.wag_w & u3o_auto_meld) ? c3n : c3o(mar_u->mel_o, mel_o)`

joemfb (changes requested):

joemfb (approved):
this will do

PR #6010, opened 2022-09-22
vere: add commandline args to set the loom size

joemfb:
This PR builds on #6002, adding arguments to control the size of the loom in each process (`--loom` and `--urth-loom`). These arguments take a binary exponent to specify the number of bytes: 20 is 1 MB, 31 is 2 GB, and the specified number must be within that range (inclusive). Bumping the compile time limit is blocked on #6001.

This also fixes a couple of initialization issues on `next/vere` -- boot verbosity was reversed, and there was another loobean "bunt" problem in #6006.


Fang-:
This looks largely sane (and thank you for catching the auto-meld initialization), but I'm slightly unclear on why the loom size is specified in bit-size if we end up mapping by number of bytes.

I know we have been talking about this in terms of "n-bit loom" primarily, but surely it's reasonable to want in-between sizes also? I'm thinking of both embedded tiny-loom cases, and the eventual >31-bit loom usage.

joemfb:
These are the acceptable sizes (in bytes):

```
> (turn (gulf 20 31) bex)
~[1.048.576 2.097.152 4.194.304 8.388.608 16.777.216 33.554.432 67.108.864 134.217.728 268.435.456 536.870.912 1.073.741.824 2.147.483.648]
```

Or in `@ux`:

```
~[0x10.0000 0x20.0000 0x40.0000 0x80.0000 0x100.0000 0x200.0000 0x400.0000 0x800.0000 0x1000.0000 0x2000.0000 0x4000.0000 0x8000.0000]
```

Those are the powers of two between 1 MB and 2 GB, inclusive. Smaller than 1 MB is irrelevant, and 2 GB is the current compile-time max (to be addressed separately). Constraining the size to a power of two ensures page alignment (page sizes are always a power of two) and uniform addressability (exactly n-bits of a reference into the address space are used, entirely).

It seems like the cleanest and simplest way to choose one of those is to specify the exponent. Do you have an alternate interface in mind?

Fang-:
> Constraining the size to a power of two ensures page alignment (...) and uniform addressability

But are these hard requirements, or just optimizations? For the former being off-exponent only affects the tail end of the loom right? And for the latter, what are the problems with not completely "filling" the bitsize we use, besides conceptual aesthetic reasons?

I was thinking something along the lines of `--loom-mb 1234` for arbitrary loom sizes, but yeah this makes it harder (with this level of granularity, impossible) to get exact exponent alignment.

tbc I don't feel very strongly either way, just trying to make sure I understand this properly. The exponent argument is a bit opaque, but if you (the user) are needing to use it, you probably know what's up anyway.

joemfb:
Page alignment is required for any virtual memory control. And not just the system page size, but vere's compile-time page size (currently 16KB and required to be some multiple of the system page size), which is used throughout the snapshotting system and to properly restore snapshots across loom sizes (since the snapshot includes pages at either end of the loom).

Powers-of-two are just on obvious constraint -- the conceit of the loom is that it is an address space, and any address space will be defined by the number of bits used by references into it, anything else wastes reference space. It might be possible to work in more variable increments, but it's harder to be confident in it. This is simpler, easier to test, and constrained such that fewer things can go wrong. 

This feature is a significant increase in optionality for operators. It's hard for me to imagine anyone *needing* more granular sizes (or even necessarily needing these smaller sizes, swap is an option, as is different configurations for these memory mappings (like `MAP_NORESERVE`). I don't except these options to be used much except out of curiosity, experiments with more constrained devices or cheaper hosting, and opting into a larger-than-default configuration once that becomes available (my primary motivation).

Fang-:
Ok, this all tracks. Thanks for the further explanation! Bitsize arguments seem fine to me now.

Issue #6009, opened 2022-09-22
ames: %ames-lost-sponsor crash blocks azimuth state updates


tinnus-napbus:
**Describe the bug**

`++on-publ-sponsor` in Ames cannot handle sponsor loss transactions (currently TODO). When I run `-azimuth-load` it eventually crashes with the error below and aborts the entire update. This means I'm stuck on the snapshot block until Ames is updated to handle this tx (or I guess maybe if I ruin the peer, idk).

```
crud: %receive event failed
[%poke %receive]
wire=/public-keys
  bar-stack
~[
  ~[//ames]
  ~[/gall/use/azimuth/0w1.Qh7zU/~tinnus-napbus/lo /dill //term/1]
  ~[/dill //term/1]
  ~[/dill //term/1]
  ~[/dill //term/1]
  ~[//http-client/0v19.mgb70]
]
take: failed
/sys/vane/ames/hoon:<[1.118 3].[1.138 20]>
/sys/vane/ames/hoon:<[1.119 3].[1.138 20]>
/sys/vane/ames/hoon:<[1.123 3].[1.138 20]>
/sys/vane/ames/hoon:<[1.125 3].[1.138 20]>
/sys/vane/ames/hoon:<[1.126 5].[1.136 7]>
/sys/vane/ames/hoon:<[1.127 5].[1.136 7]>
/sys/vane/ames/hoon:<[1.135 32].[1.135 81]>
/sys/vane/ames/hoon:<[1.888 5].[2.117 7]>
/sys/vane/ames/hoon:<[1.890 5].[2.117 7]>
/sys/vane/ames/hoon:<[1.890 9].[1.907 11]>
/sys/vane/ames/hoon:<[1.892 9].[1.907 11]>
/sys/vane/ames/hoon:<[1.900 11].[1.900 61]>
/sys/vane/ames/hoon:<[2.004 7].[2.015 17]>
/sys/vane/ames/hoon:<[2.006 7].[2.015 17]>
/sys/vane/ames/hoon:<[2.007 9].[2.007 44]>
[%ames-lost-sponsor ~tinnus-napbus ~ristyc-ridwyx]
/sys/vane/ames/hoon:<[2.007 42].[2.007 44]>
```

**To Reproduce**
Steps to reproduce the behaviour:
1. `|hi ~ristyc-ridwyx`
2. `-azimuth-load`
3. you get error above and your azimuth state is now stuck on the snapshot block

**Expected behaviour**
[...]

PR #6011, opened 2022-09-22
ames: hotfix for pki sponsorship loss

joemfb:
This PR addresses #6009 by setting a ship's sponsor in %ames to itself when azimuth sponsorship is lost.

Barring a state migration in %ames, the only other obvious option I see is to simply ignore sponsorship loss in %ames. This will likely impact networking more than that would, but it shouldn't entirely prevent it, and it seems better to respect the pki change. Thoughts?

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1997,14 +1997,15 @@
       event-core
     ::  +on-publ-sponsor: handle new or lost sponsor for peer
     ::
-    ::    TODO: handle sponsor loss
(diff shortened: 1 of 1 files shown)


philipcmonk:
I would worry that now anything which loops through sponsorship until it gets to a galaxy might get into an infinite loop.  I think I prefer ignoring the sponsorship change for now.

philipcmonk (approved):
Looks good

Fang- (approved):
Yes, this seems like the simplest & lowest-risk fix.

PR #6015, opened 2022-09-28
mingw: adds (partial) implementations of pread/pwrite

joemfb:
This PR builds on #5889. It's a draft due to conflicts, and should wait to be reconciled with #6003.

The implementations are "partial" because `ReadFile()` and `WriteFile()` update the file position after i/o completion -- these are not spec-conformant implementations, but are fit for our single-threaded use-cases (so long as calls to their non-positioned counterparts aren't intermingled on the same file *description*).

### DIFF
--- a/pkg/urbit/compat/mingw/compat.c
+++ b/pkg/urbit/compat/mingw/compat.c
@@ -144,6 +144,63 @@ int link(const char *path1, const char *path2)
   return -1;
 }
 
+ssize_t pread(int fd, void *buf, size_t count, off_t offset)
+{
+  DWORD len = 0;
+
+  OVERLAPPED overlapped = {0};
+
+  overlapped.OffsetHigh = (sizeof(off_t) <= sizeof(DWORD)) ?
+                          (DWORD)0 : (DWORD)((offset >> 32) & 0xFFFFFFFFL);
+  overlapped.Offset     = (sizeof(off_t) <= sizeof(DWORD)) ?
+                          (DWORD)offset : (DWORD)(offset & 0xFFFFFFFFL);
+
+  HANDLE h = (HANDLE)_get_osfhandle(fd);
+
+  if ( INVALID_HANDLE_VALUE == h ) {
+    errno = EBADF;
+    return -1;
+  }
+
+  if ( !ReadFile(h, buf, count, &len, &overlapped) ) {
+    DWORD err = GetLastError();
+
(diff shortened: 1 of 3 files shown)


pkg/urbit/compat/mingw/compat.c
```diff
@@ -144,6 +144,63 @@ int link(const char *path1, const char *path2)
...
+ssize_t pread(int fd, void *buf, size_t count, off_t offset)
+{
+  DWORD len = 0;
+
+  OVERLAPPED overlapped = {0};
+
+  overlapped.OffsetHigh = (sizeof(off_t) <= sizeof(DWORD)) ?
+                          (DWORD)0 : (DWORD)((offset >> 32) & 0xFFFFFFFFL);
+  overlapped.Offset     = (sizeof(off_t) <= sizeof(DWORD)) ?
+                          (DWORD)offset : (DWORD)(offset & 0xFFFFFFFFL);
```
mcevoypeter, line 156: Why not use a macro to avoid repeating `0xFFFFFFFFL` multiple times?
```c
#define DWORD_MASK 0xFFFFFFFFL
```

pkg/urbit/noun/events.c
```diff
@@ -476,11 +473,9 @@ _ce_patch_write_page(u3_ce_patch* pat_u,
...
-  if ( -1 == lseek(pat_u->mem_i, (pgc_w << (u3a_page + 2)), SEEK_SET) ) {
-    c3_assert(0);
-  }
   if ( (1 << (u3a_page + 2)) !=
-       write(pat_u->mem_i, mem_w, (1 << (u3a_page + 2))) )
+       pwrite(pat_u->mem_i, mem_w, (1 << (u3a_page + 2)),
+              (pgc_w << (u3a_page + 2))) )
   {
     c3_assert(0);
   }
```
mcevoypeter, line 481: Let's print `strerror(errno)` here if `pwrite()` fails.

mcevoypeter (changes requested):
The two comments below aside, it looks good to me.

pkg/urbit/compat/mingw/compat.c
joemfb: This code is entirely consistent with the conventions used throughout `compat.c`. There are many things about that file that I would change if I was going to refactor it, but all that I'm doing here is adding these two wrapper functions.

PR #6016, opened 2022-09-28
garden: respect all zeroes in multipart content

### LINKED ISSUE #6012: Globbing zip files truncates/corrupts them
**Describe the bug**
See: web+urbitgraph://group/~middev/the-forge/graph/~middev/bail-meme-8836/170141184505800909131634257942819110912

.zip files uploaded to a ship via the glob process or fetched from an external site by a gall agent are returned to a web front-end with empty bytes at the end of the file truncated, which can corrupt the file and make it unreadable.

**To Reproduce**
1. Add a zip file to a glob 
2. Glob the files
3. Download the globbed file and diff between it and the original file.
4. Empty trailing bytes are missing

**Expected behaviour**
Globbed files are unchanged by the process

Fang-:
Some files, like zip files, may have trailing zero bytes, which get eaten when stored in reverse in an atom. So instead, we say `content` is an `octs`, passing the true size along.

Fixes #6012.

Not immediately obvious to me if this should go into `next/landscape` or not, considering the only changes there right now are frontend changes.

### DIFF
--- a/pkg/garden/app/docket.hoon
+++ b/pkg/garden/app/docket.hoon
@@ -599,11 +599,11 @@
     ?:  =('desk' name)
       ::  must be a desk with existing charge
       ::
-      ?.  ((sane %ta) body)
-        [desk glob (cat 3 'invalid desk: ' body) err]
-      ?.  (~(has by charges) body)
-        [desk glob (cat 3 'unknown desk: ' body) err]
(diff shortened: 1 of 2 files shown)


liam-fitzgerald (approved):
this doesn't break API and so therefore should go in `next/landscape`

PR #6019, opened 2022-09-29
story: fix story printing

### LINKED ISSUE #6017: +story-list: each line in a commit is printed to the right of the previous
**Describe the bug**
```
> +story-list, =desk %garden

commit: 0v10.blbfu.8a414.qrjqo.ov847.gc6v7.6dpob.49i6s.ndi49.ve74j.a04qr
                                                                        some title

                                                                                  some body
                                                                                           |||
                                                                                              aaaaaaaaa

                                                                                                       bbbbbbb
    ---
```

**To Reproduce**
1. create a story and add a commit message
2. run +story-list on it
3. observe each line in a commit is printed to the right of the last

**Expected behaviour**
should be better formatted

**Notify maintainers**
@ynx0

tinnus-napbus:
`+story-list` produced janky indentation because the `$-(story wain)` functions encoded linefeeds in the cords of the wain and the printer doesn't like this.

Story printing functions have been changed to produce pure wains without linefeeds.

resolves #6017

### DIFF
--- a/pkg/base-dev/lib/story.hoon
+++ b/pkg/base-dev/lib/story.hoon
@@ -38,41 +38,17 @@
 ::
 ::  Canonical textual representation
 ::
-++  tako-to-text
-  |=  [=tako:clay]
-  ^-  tape
-  "commit: {<`@uv`tako>}\0a"
-::
-++  proses-to-text
-  |=  [=proses]
-  ^-  tape
-  =/  proses-list=(list prose)  ~(tap in proses)
-  ?:  ?=(~ proses-list)  ""
-  ?:  ?=([prose ~] proses-list)
-    (prose-to-text i.proses-list)
-  %-  tail
-  %^  spin  `(list prose)`t.proses-list
-    (prose-to-text i.proses-list)
-  |=  [prz=prose state=tape]
-  ^-  [prose tape]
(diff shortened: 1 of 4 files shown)


pkg/base-dev/lib/story.hoon
```diff
@@ -38,41 +38,17 @@
...
+  |=  [=tako:clay =proses]
+  ^-  wain
+  :-  (crip "commit: {<`@uv`tako>}")
+  %-  zing
+  %+  join  `wain`~['|||']
+  %+  turn  ~(tap in proses)
+  |=  prose
+  ^-  wain
+  %-  to-wain:format
+  :((cury cat 3) title '\0a\0a' body)
```
joemfb, line 51: For efficiency reasons, `(rap 3 a b c ~)` is always preferable to `:((cury cat 3) a b c)` -- the former is a single call (and a single allocation inside the jet), the latter is `n-1` calls and allocations. (I'm on a personal mission to broadcast this far and wide.)

tinnus-napbus: whoops for some reason I totally forgot about rap

philipcmonk (approved):
Thanks!

PR #6022, opened 2022-09-30
aqua: make `%swap-files` poke take a desk and add it to pill

dr-frmr:
Previously, attempting to use aqua's `%swap-files` poke would crash, since the scry path was malformed. The desk was also hard-coded to `%work` -- now the user inputs a desk name. I've tested the new poke, and was able to successfully add a desk to aqua's pill and spin up a new fakeship with it, which then properly installed the agents in that desk.


philipcmonk (approved):
Looks good, thanks!

PR #6030, opened 2022-10-14
npm README typo: setings -> settings

0x70b1a5:


belisarius222 (approved):
lgtm

zalberico:
@belisarius222 - I switched the base branch to `next/arvo` and then (accidentally) merged the updates of that branch in instead of rebasing this on top of it (clicked the wrong button in github UI).

When I assign it I'm not looking for a lgtm as much as ownership in getting it to the final state (approved/merged, closed for some other reason, changes needed and X person will make them). In this case, it'd be getting it on `next/arvo` and then merging it to close it.

There's some failing tests here (not sure if that matters) and I annoyingly can't fix the merge with a hard reset/force push because I don't have access to the underlying repo fork.

belisarius222:
Ok, noted. I wasn't sure if I should merge it or let someone else merge it, but when I reviewed it (before your rebase), there was only one commit, which modified only a readme text file -- no code changes at all -- so I felt comfortable approving it at least.

Sounds like in the future, if a PR is a clear win like this, I should check the base branch is what it should be (next/arvo in this case), then go ahead and merge the PR into that branch. Is that correct?

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

matildepark:
Why does a README typo change have to hit next/arvo? Its changes are not contingent on next/arvo changes, and it was just fine being merged into master.

zalberico:
Yes - it's a readme change so in this case it doesn't really matter, but nothing merges into master outside of release (next/arvo).

That way `next/arvo` doesn't need to update master unless something went into it from a release. With runtime in a separate repo (soon) this will be generally true and we have clean releases from next/arvo merges.

zalberico:
> Sounds like in the future, if a PR is a clear win like this, I should check the base branch is what it should be (next/arvo in this case), then go ahead and merge the PR into that branch. Is that correct?

Yep - I think that's the way forward.

zalberico:
Closing in favor of 6083

PR #6031, opened 2022-10-19
Add folders for pill creation.

sigilante:
The current pill library for +brass leaves out /doc (%docs) and /dia (%citadel).

This means that brass pills based on these desks currently break on initial boot.

### DIFF
--- a/pkg/base-dev/lib/pill.hoon
+++ b/pkg/base-dev/lib/pill.hoon
@@ -39,6 +39,8 @@
 ++  file-ovum
   =/  directories=(list path)
     :~  /app    ::  %gall applications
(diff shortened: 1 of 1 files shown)


zalberico:
@belisarius222 - assigning this to you for review and next steps.

Fang-:
Alternative implementation: just include everything, but explicitly ignore the `/tests` directory, and the readme & license & testing files in root. I don't think expanding this list for every new third party dir is something we can keep up in the long run.

Something also to consider, is that #5470 will enable injection of custom packages and such into the boot sequence, without custom pills. "Pill is only for base kernel" could be a valid stance to take.

PR #6035, opened 2022-10-26
arvo: retool for goerli instead of deprecated ropsten

liam-fitzgerald:
Ropsten is so over.
Fortunately, we are so back with a fresh testnet on Goerli.

Went ahead and renamed all instances of `ropsten` to `goerli`

Exceptions/Comments:
- I don't quite understand what's going on in `gen/tx.hoon`, it defaults to ropsten? so I guess it's not actually being used
- `sur/dice.hoon` retains %ropsten to avoid having to write a redundant state upgrade
- The ropsten nix derivations are broken anyway (still refer to `+contracts` being in zuse.hoon), is it worth upgrading the naming there?
- This can't be tested until the corresponding bridge work is finished


liam-fitzgerald:
also: is +launch just the block where azimuth ownership was transferred to ecliptic?

philipcmonk:
+launch is the block where the first contract was deployed.  leaving the nix stuff is fine.

liam-fitzgerald:
sorry, I actually meant +public

philipcmonk:
It looks like probably yes, but i think only gaze uses it, so @Fang- would know.

zalberico:
> * This can't be tested until the corresponding bridge work is finished

What corresponding bridge work - does that exist in an issue or PR somewhere? I don't want this to sit around forever with no path to us merging it. Is anything currently blocking it?

jalehman:
@yosoyubik would you mind reviewing this and, upon your approval, we'll merge?

yosoyubik (approved):
I was not able to do a proper end-to-end test because of [this](https://github.com/urbit/bridge/issues/1096), but I've tested the backend part of it (that concerns this PR) by running a ship with both /app/azimuth and /app/roller hooked up to goerli, inspected their naive states and confirmed that the right contracts are called (both L1 and L2 points are retrieved) by comparing to etherscan so I approve this PR.

Issue #6024, opened 2022-10-05
dojo> term: write failed Bad file descriptor Assertion '0' failed in vere/io/term.c:51


ks0m1c:
I am encountering abort state, I haven't been able to fix with a pack and a meld. The terminal driver appears to be crashing but maybe the error is propagating from deeper in the stack.

This is the [urbit binary](https://github.com/botter-nidnul/urbit/releases/tag/urbit-v1.10-aarch64-linux) I am using on my Raspberry Pi 4B 64bit

** Trace** 
```
~
urbit 1.10
boot: home is /home/pi/urbit/[ship]
loom: mapped 2048MB
lite: arvo formula 11a9e7fe
lite: core 38d4ad4d
lite: final state 38d4ad4d
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/250.560.512
boot: installed 351 jets
vere: checking version compatibility
ames: live on 41475
conn: listening on /home/pi/urbit/[ship]/.urb/conn.sock
http: web interface live on http://localhost:8080
http: loopback live on http://localhost:12321
~[ship]:dojo> term: write failed Bad file descriptor
Assertion '0' failed in vere/io/term.c:51
home: bailing out
Aborted
```

**To Reproduce**
Steps to reproduce the behaviour:
1. SOme have encountered this before (source: [twitter](https://nitter.it/__________1890/status/1575145573908463616#m))

**Expected behaviour**
I expect the ship to boot
**Screenshots**

**System (please supply the following information, if relevant):**
 - OS: "Debian GNU/Linux 11 (bullseye)"
 - Kernel: Linux 5.15.61-v8+ #1579 SMP PREEMPT Fri Aug 26 11:16:44 BST 2022 aarch64 GNU/Linux
 - Vere and Urbit OS versions:  1.10 
 - Ships info: 
```
loom: mapped 2048MB
boot: protected loom
live: loaded: MB/250.560.512
boot: installed 351 jets

urbit: [ship] at event 67949
  disk: live=&, event=67957

lmdb info:
  map size: 536870912000
  page size: 4096
  max pages: 131072000
  number of pages used: 46570
  last transaction ID: 57086
  max readers: 126
  number of readers used: 0
  file size (page): 190750720
  file size (stat): 190750720`

```

alvatar:
I have the same issue:


```
~
urbit 1.10
boot: home is /home/alvatar/togtyr-roplux
loom: mapped 2048MB
lite: arvo formula 11a9e7fe
[...]

PR #6040, opened 2022-10-26
vere: check for permission error on old process kill

pkova:
If you run the urbit binary as root (common in `systemd` services and similar) and happen to run the urbit binary again as a less privileged user the lockfile handling of vere fails. This is a common cause of corruption in situations like #6024. To reproduce, start a fakezod with `sudo`, then run the same fakezod as a normal user.

This PR adds an explicit check for the `kill` permission error, making the protection against dual boot situations more robust.

### DIFF
--- a/pkg/urbit/vere/disk.c
+++ b/pkg/urbit/vere/disk.c
@@ -592,7 +592,15 @@ u3_disk_acquire(c3_c* pax_c)
     else if (pid_w != getpid()) {
       c3_w i_w;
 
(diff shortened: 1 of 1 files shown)


joemfb (approved):
this will do, thanks!

joemfb:
The CI failures here are unrelated and have been fixed on the target branch.

vcavallo:
Old issue, but I just ran into it: `$PIER/.http.ports` may also be owned by `root` in the above-described scenario. `chown`'ing it properly resolves.

PR #6041, opened 2022-10-27
arvo: use breadth-first move order

philipcmonk:
There's a 7-year-old debate about depth-first vs breadth-first move
order, and it flared up again recently, and I think I've convinced
myself that breadth-first is actually importantly better than
depth-first.

Implementation-wise, I think this one-line change is best (only the change
to +emit has any effect, the rest only changes the |verb output).  The main
alternative is to change `run` from `(list plan)` to `(qeu plan)`, which
has better asymptotics if you move trees with large fanout.  However,
(1) this would need to be a deque, which we don't have yet, and (2) even
then the constants are somewhat worse, and I suspect the constants
dominate for the size of move trees we're dealing with.

I've run this on fake ships, booted fake ships from a pill, and run it
on a localhost copy of ~wicdev-wisryt.  All of this is on top of
philip/agent-clay, but that will go out either before or along with this
change.

Some background on the move order debate:

We consider computation inside Arvo to have three stacks:
- the nock stack, which gets deeper as you go deeper into a nock formula
- the duct stack, which gets deeper as you %pass along a duct.  this
  might last through multiple events and is not a true stack because you
  might pop (%give) multiple times from the same point.
- the "bar" stack, which gets deeper as you go deeper in the move tree.
  "bar" refers to its visualization in |verb.  Both %pass and %give push
  bars onto this stack, and each move sent as a result of those moves
  are one level deeper in the stack.

This change affects the bar stack.  Visually, when evaluating
depth-first an event may look like this:

```
["" %unix %belt /d/term/1 ~2022.10.27..06.32.09..30db]
["|" %pass [%dill %g] [[%deal [~zod ~zod] %hood %poke] /] ~[//term/1]]
["||" %give %gall [%unto %poke-ack] i=/dill t=~[//term/1]]
["||" %pass [%gall %g] [[%deal [~zod ~zod] %dojo %poke] /use/hood/0w2.efXKi/out/~zod/dojo/drum/phat/~zod/dojo] ~[/dill //term/1]]
["|||" %give %gall [%unto %poke-ack] i=/gall/use/hood/0w2.efXKi/out/~zod/dojo/drum/phat/~zod/dojo t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w2.efXKi/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["||||" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w2.efXKi/~zod/view/ t=~[/dill //term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w2.efXKi/~zod/view/ t=~[/dill //term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w2.efXKi/~zod/view/ t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w2.efXKi/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["||||" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w2.efXKi/~zod/view/ t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w2.efXKi/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w2.efXKi/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
```

While breadth-first it looks like this:

```
["" %unix %belt /d/term/1 ~2022.10.27..06.32.57..1d2c]
["|" %pass [%dill %g] [[%deal [~zod ~zod] %hood %poke] /] ~[//term/1]]
["||" %give %gall [%unto %poke-ack] i=/dill t=~[//term/1]]
["||" %pass [%gall %g] [[%deal [~zod ~zod] %dojo %poke] /use/hood/0w3.qnAr5/out/~zod/dojo/drum/phat/~zod/dojo] ~[/dill //term/1]]
["|||" %give %gall [%unto %poke-ack] i=/gall/use/hood/0w3.qnAr5/out/~zod/dojo/drum/phat/~zod/dojo t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w3.qnAr5/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w3.qnAr5/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w3.qnAr5/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["|||" %give %gall [%unto %fact] i=/gall/use/hood/0w3.qnAr5/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["||||" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["||||" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w3.qnAr5/~zod/view/ t=~[/dill //term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w3.qnAr5/~zod/view/ t=~[/dill //term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w3.qnAr5/~zod/view/ t=~[/dill //term/1]]
["|||||" %give %dill %blit i=/gall/use/herm/0w3.qnAr5/~zod/view/ t=~[/dill //term/1]]
```

These are the same 15 lines, and they're each at the same depth, but
they're in a different order.  Most code we write is agnostic to this
order, because there are many circumstances where this order gets
inverted compared to our expectations.

The most obvious difference is that the four %facts from dojo to hood
happen one right after the other instead of being mingled with other
moves.  If some of those intermingled moves invoked dojo (which would be
a form of reentrancy) and caused dojo to emit more facts, those facts
would be given to hood *before* the %facts which were already on the
stack to be sent to hood.  If this is textual output, then it will be in
reverse order.  The breadth-first move order fixes this problem
completely by running all the facts that were issued at the same time
before processing the moves that those themselves produced.

This leads to a very important principle: breadth-first ordering
guarantees that moves will be processed in the order they are emitted.
Depth-first ordering constantly violates this, and in the presence of
reentrancy this can cause extremely unexpected results.

The usual argument for depth-first ordering is that it feels like it
should allow greater composability.  In both cases, if A sends moves to
B and C, then A knows that the move to B will happen before the move to
C.  However, in depth-first ordering, we further know that anything that
B causes to happen will also happen before the move to C.  This means
the move to B will "fully complete", even if it delegates some of the
work to another entity.  Of course, this doesn't apply if B needs to
send something over the wire to fully complete.

Another guarantee that depth-first ordering gives you is that the first
move you send will be processed immediately, with no intervening moves.
Breadth-first ordering often will process other moves that were already
in the queue.

Overall, the guarantees that depth-first ordering gives are rarely
useful because they can be violated in practice, while the guarantees
that breadth-first ordering gives are consistent and more useful.

The second commit changes the |verb output to clearly specify the "bar"
stack.  Without this, it's not clear which moves were produced by which
moves.  The way to read this is that a move with stack `ABC` was
produced by a move with stack `AB`, so you can scan upward to find it.
This is the example from above:

```
["" %unix %belt /d/term/1 ~2022.10.27..07.38.32..876d]
["0" %pass [%dill %g] [[%deal [~zod ~zod] %hood %poke] /] ~[//term/1]]
["01" %give %gall [%unto %poke-ack] i=/dill t=~[//term/1]]
["01" %pass [%gall %g] [[%deal [~zod ~zod] %dojo %poke] /use/hood/0wUvCOt/out/~zod/dojo/drum/phat/~zod/dojo] ~[/dill //term/1]]
["013" %give %gall [%unto %poke-ack] i=/gall/use/hood/0wUvCOt/out/~zod/dojo/drum/phat/~zod/dojo t=~[/dill //term/1]]
["013" %give %gall [%unto %fact] i=/gall/use/hood/0wUvCOt/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["013" %give %gall [%unto %fact] i=/gall/use/hood/0wUvCOt/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["013" %give %gall [%unto %fact] i=/gall/use/hood/0wUvCOt/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["013" %give %gall [%unto %fact] i=/gall/use/hood/0wUvCOt/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["0135" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["0136" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["01359" %give %dill %blit i=/gall/use/herm/0wUvCOt/~zod/view/ t=~[/dill //term/1]]
["01359" %give %dill %blit i=/gall/use/herm/0wUvCOt/~zod/view/ t=~[/dill //term/1]]
["01359" %give %dill %blit i=/gall/use/herm/0wUvCOt/~zod/view/ t=~[/dill //term/1]]
["0136:" %give %dill %blit i=/gall/use/herm/0wUvCOt/~zod/view/ t=~[/dill //term/1]]
```

Here we can see that the dill blits are not all caused by the same move,
which was not clear with just bars in the output.  This is not as
aesthetically clean, but the information is frequently important.

NB: this |verb change requires an upgrade to arvo's state, which is not
included in this PR currently.  It's easy to do as a kelvin upgrade
(which this upgrade should do anyway) by updating the +heir type and
writing a state upgrade in +load.

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -1241,7 +1241,7 @@
     ::
     ++  emit
       |=  pan=plan
-      this(run [pan run])
+      this(run (weld run pan ~))
     ::  +loop: until done
     ::
     ++  loop
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -184,7 +184,7 @@
       kel=(list (pair path (cask)))
       fil=(list (pair path (cask)))
   ==
-+$  germ  [vane=term bars=(list duct)]
++$  germ  [vane=term bars=(list [@tD duct])]
 +$  grub
   $:  ::  who: identity once we know it
       ::  eny: entropy once we learn it
@@ -337,6 +337,15 @@
       ==
     ~
   `[[q.p.i q.p.i.t]:p.u.lot u.bem]
+::
+++  rant
+  |=  [a=(list [char=@tD *]) b=tape]
+  ^-  tape
+  ?~  a
+    b
+  =.  b  [char.i.a b]
+  $(a t.a)
+::
 ::
 ++  look
   ~/  %look
@@ -1193,6 +1202,7 @@
             ::
             run=(list plan)
             out=(list ovum)
+            neb=_14
             gem=germ
             dud=(unit goof)
             $=  but  %-  unit
@@ -1275,7 +1285,7 @@
         =*  task  task.note.ball.move
         ::
         ~?  &(!lac.fad !=(%$ vane.gem))
-          :-  (runt [(lent bars.gem) '|'] "")
+          :-  (rant bars.gem "")
           :^  %pass  [vane.gem vane]
             ?:  ?=(?(%deal %deal-gall) +>-.task)
               :-  :-  +>-.task
@@ -1296,7 +1306,7 @@
         =*  task  task.note.ball.move
         ::
         ~?  !lac.fad
-          :-  (runt [(lent bars.gem) '|'] "")
+          :-  (rant bars.gem "")
           [%slip [vane.gem vane] (symp +>-.task) duct]
         ::
         (call duct vane task)
@@ -1316,7 +1326,7 @@
           ?>(?=(^ wire) wire)
         ::
         ~?  &(!lac.fad !=(%$ way) |(!=(%blit +>-.gift) !=(%d vane.gem)))
-          :-  (runt [(lent bars.gem) '|'] "")
+          :-  (rant bars.gem "")
           :^  %give  vane.gem
             ?:  ?=(%unto +>-.gift)
               [+>-.gift (symp +>+<.gift)]
@@ -1397,8 +1407,9 @@
         ;;(waif q.p.task)
       ::
       =.  way  (grow way)
-      %+  push  [way duct bars.gem]
-      ~|  bar-stack=`(list ^duct)`[duct bars.gem]
+      =.  neb  (^mod +(neb) 94)
+      %+  push  [way [(add '!' neb) duct] bars.gem]
+      ~|  bar-stack=`(list ^duct)`[duct (turn bars.gem tail)]
       %.  task
       call:(spin:(plow way) duct eny dud)
     ::  +take: retreat along call-stack
@@ -1415,12 +1426,13 @@
       ::  the caller was a vane
       ::
       =.  way  (grow way)
-      %+  push  [way duct bars.gem]
+      =.  neb  (^mod +(neb) 94)
+      %+  push  [way [(add '!' neb) duct] bars.gem]
       ::
       ::  cons source onto .gift to make a $sign
       ::
       ~|  wire=wire
-      ~|  bar-stack=`(list ^duct)`[duct bars.gem]
+      ~|  bar-stack=`(list ^duct)`[duct (turn bars.gem tail)]
       %.  [wire [vane.gem gift]]
       take:(spin:(plow way) duct eny dud)
     ::  +push: finalize an individual step
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -339,9 +339,12 @@
   `[[q.p.i q.p.i.t]:p.u.lot u.bem]
 ::
 ++  rant
-  |=  [a=(list [char=@tD *]) b=tape]
-  ^-  tape
+  |=  a=(list [char=@tD *])
+  =|  b=tape
   ?~  a
+    b
+  |-  ^-  tape
+  ?~  t.a
     b
   =.  b  [char.i.a b]
   $(a t.a)
@@ -1202,7 +1205,7 @@
             ::
             run=(list plan)
             out=(list ovum)
-            neb=_14
+            neb=_15
             gem=germ
             dud=(unit goof)
             $=  but  %-  unit
@@ -1261,7 +1264,7 @@
           ==
         abet
       ?:  =(~ q.i.run)    :: XX TMI
-        loop(run t.run)
+        loop(run t.run, neb 15)
       =.  dud  ~
       =.  gem  p.i.run
       =^  mov=move  q.i.run  q.i.run
@@ -1273,6 +1276,10 @@
       ^+  this
       ::
       ~?  &(!lac.fad ?=(^ dud))  %goof
+      =.  neb  (^mod +(neb) 94)
+      =/  bar  (add '!' neb)
+      ::
+      |-
       ::
       ?-  -.ball.move
       ::
@@ -1285,7 +1292,7 @@
         =*  task  task.note.ball.move
         ::
         ~?  &(!lac.fad !=(%$ vane.gem))
-          :-  (rant bars.gem "")
+          :-  (rant [bar duct] bars.gem)
           :^  %pass  [vane.gem vane]
             ?:  ?=(?(%deal %deal-gall) +>-.task)
               :-  :-  +>-.task
@@ -1296,7 +1303,7 @@
         ::
         ::  cons source onto wire, and wire onto duct
         ::
-        (call [[vane.gem wire] duct] vane task)
+        (call bar [[vane.gem wire] duct] vane task)
       ::
       ::  %slip: lateral move
       ::
@@ -1306,10 +1313,10 @@
         =*  task  task.note.ball.move
         ::
         ~?  !lac.fad
-          :-  (rant bars.gem "")
+          :-  (rant [bar duct] bars.gem)
           [%slip [vane.gem vane] (symp +>-.task) duct]
         ::
-        (call duct vane task)
+        (call bar duct vane task)
       ::
       ::  %give: return move
       ::
@@ -1326,14 +1333,14 @@
           ?>(?=(^ wire) wire)
         ::
         ~?  &(!lac.fad !=(%$ way) |(!=(%blit +>-.gift) !=(%d vane.gem)))
-          :-  (rant bars.gem "")
+          :-  (rant [bar duct] bars.gem)
           :^  %give  vane.gem
             ?:  ?=(%unto +>-.gift)
               [+>-.gift (symp +>+<.gift)]
             (symp +>-.gift)
           duct.move
         ::
-        (take duct wire way gift)
+        (take bar duct wire way gift)
       ::
       ::  %hurl: action with error
       ::
@@ -1398,7 +1405,7 @@
     ::  +call: advance to target
     ::
     ++  call
-      |=  [=duct way=term task=maze]
+      |=  [bar=@tD =duct way=term task=maze]
       ^+  this
       ?:  ?=(%$ way)
         ~>  %mean.'arvo: call:pith failed'
@@ -1407,15 +1414,14 @@
         ;;(waif q.p.task)
       ::
       =.  way  (grow way)
-      =.  neb  (^mod +(neb) 94)
-      %+  push  [way [(add '!' neb) duct] bars.gem]
+      %+  push  [way [bar duct] bars.gem]
       ~|  bar-stack=`(list ^duct)`[duct (turn bars.gem tail)]
       %.  task
       call:(spin:(plow way) duct eny dud)
     ::  +take: retreat along call-stack
     ::
     ++  take
-      |=  [=duct =wire way=term gift=maze]
+      |=  [bar=@tD =duct =wire way=term gift=maze]
       ^+  this
       ?:  ?=(%$ way)
         ::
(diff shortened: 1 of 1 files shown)


liam-fitzgerald:
+1. Depth-first is extremely unintuitive and was a source of issues in the extremely IO-heavy groups 1 codebase. 

Also, we have a finger-tree based implementation of deques in the content dist branch fwiw

pkg/arvo/sys/arvo.hoon
```diff
@@ -1241,7 +1251,7 @@
     ::
     ++  emit
       |=  pan=plan
-      this(run [pan run])
+      this(run (weld run pan ~))
```
eamsden, line 1254: At the risk of slightly-more-complicated code, but without going to full-on finger trees, a commonly-used trick to get an amortized-kind-of-efficient deque with an SLL is to keep a backend list and a frontend list. decons off the frontend list, cons onto the backend list. When the frontend list is exhausted, reset it to the flop of the backend list and then null the backend list.

This means backend pushes are always O(1), and pops are amortized O(1).

philipcmonk: Oh yeah, I forgot this trick.  That's probably worth doing.

philipcmonk:
@Fang- noticed that we need to print not only the parent's duct stack but also our own to be totally unambiguous.  He also noticed that we can restart the count from `0` on every new plan, which makes the flow easier to read.  I also subjectively feel like starting from 1 looks nicer.  Now our canonical example looks like this:

```
["" %unix %belt /d/term/1 ~2022.11.2..00.18.15..ce5b]
["1" %pass [%dill %g] [[%deal [~zod ~zod] %hood %poke] /] ~[//term/1]]
["11" %give %gall [%unto %poke-ack] i=/dill t=~[//term/1]]
["12" %pass [%gall %g] [[%deal [~zod ~zod] %dojo %poke] /use/hood/0w2.Rh6DI/out/~zod/dojo/drum/phat/~zod/dojo] ~[/dill //term/1]]
["121" %give %gall [%unto %poke-ack] i=/gall/use/hood/0w2.Rh6DI/out/~zod/dojo/drum/phat/~zod/dojo t=~[/dill //term/1]]
["122" %give %gall [%unto %fact] i=/gall/use/hood/0w2.Rh6DI/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["123" %give %gall [%unto %fact] i=/gall/use/hood/0w2.Rh6DI/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["124" %give %gall [%unto %fact] i=/gall/use/hood/0w2.Rh6DI/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["125" %give %gall [%unto %fact] i=/gall/use/hood/0w2.Rh6DI/out/~zod/dojo/1/drum/phat/~zod/dojo t=~[/dill //term/1]]
["1221" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["1231" %give %gall [%unto %fact] i=/dill t=~[//term/1]]
["12211" %give %dill %blit i=/gall/use/herm/0w2.Rh6DI/~zod/view/ t=~[/dill //term/1]]
["12213" %give %dill %blit i=/gall/use/herm/0w2.Rh6DI/~zod/view/ t=~[/dill //term/1]]
["12215" %give %dill %blit i=/gall/use/herm/0w2.Rh6DI/~zod/view/ t=~[/dill //term/1]]
["12311" %give %dill %blit i=/gall/use/herm/0w2.Rh6DI/~zod/view/ t=~[/dill //term/1]]

ohAitch:
The less-clear numerical "bar stack" printout is a symptom of the larger change, which is interleaving the execution of unrelated calls - in status quo depth-first order the bars *do* contain enough information, your proximate cause is the most recent one-less-indented move. That said, I defer to liam on which causes more debugging headache in practice.

Re: deque, the qeu data structure is double ended in *practice*, we just don't have the jets for head-pop and tail-push right?  
Though I'm not sure double-endedness is even necessary here, you have a qeu of moves at the current bar-depth and a qeu of moves at the next bar-depth. If you're at depth 3 you've already executed all moves at depths 0/1/2 and therefore are creating exclusively depth 4 new ones, after you run out of depth 3 you swap the qeus and will be executing depth 4 while pushing depth 5, etc.
("all the extra ~s entail cons allocations which are more expensive in practice than welding lists of size 6 or w/e" is ofc valid)

eamsden:
> Though I'm not sure double-endedness is even necessary here, you have a qeu of moves at the current bar-depth and a qeu of moves at the next bar-depth. If you're at depth 3 you've already executed all moves at depths 0/1/2 and therefore are creating exclusively depth 4 new ones, after you run out of depth 3 you swap the qeus and will be executing depth 4 while pushing depth 5, etc. ("all the extra ~s entail cons allocations which are more expensive in practice than welding lists of size 6 or w/e" is ofc valid)

That's precisely the behavior that my suggestion will have
https://github.com/urbit/urbit/pull/6041#discussion_r1006972901

but without explicitly tracking depth boundaries.

ohAitch:
I mean you end up tracking them somewhere for the debug prints, but yeah whether it's two lists or two queues you don't need them there. And a single list flop per move does sound cheaper than qeu maintenance even asymptotically.

jalehman:
If we're going to merge this we'll need a more thorough testing plan that involves using it on a few live ships. We'll work on coming up with that and beginning the testing in coming days.

philipcmonk:
IMO a tick should be a vane activation, no more or less. So all three of
those happen in separate ticks.

I do think it's reasonable to say the scry function passed into a vane is
taken from when the move is emitted instead of when the move is executed.

belisarius222:
I'm sympathetic to the idea of tracking each vane activation with a single tick, without exceptions.  It's a nice simple mental model.

Ok, so let's say Arvo is on tick 100, then runs the first vane, which produces two moves to Ames and one to Behn.  Now the tick is at 101.  Arvo runs Ames at tick 102, then runs Ames again at tick 103; possibly in parallel, it runs Behn at tick 104.

If Behn scrys into Ames during this activation, Arvo will use a pinned version of Ames just after tick 101, before it runs the two moves.  I'm wondering how this should work.  Behn will call its scry function, which closes around this pinned Ames, at tick 104, since that's the current tick and Behn doesn't know what's going on in other vanes.  Ames will then presumably have its `+scry` gate called at tick 104, which Ames would compare with the tick supplied in its `$vane-sample` along with `now` and `eny`.  Since the `$vane-sample` tick and the tick in the scry path are the same, Ames will treat this as a request for latest.

I think this works.  It's a little weird for vanes to have to compare ticks like that, though, since only the current tick is ever valid to scry at.  I wonder if we'd be better off adding another vane arm (`+read`, perhaps) that just reads a path from the current state -- the vane would know that would only ever be called by Arvo at the current tick, or at the current date if Arvo's `+peek` arm gets called externally.  Arvo's vane scry handler would know to convert requests at the current tick into calls to that new arm.

philipcmonk:
I think technically it'll be cleaner to supply to each vane the tick that
its scry function corresponds to. For example behn would run at tick 104,
but it wouldn't know that, it would just know that the scry function it was
given is at tick 101, and it would use that to scry. Not sure there's any
practical difference though.

As for whether you want a separate +read arm, I don't think it makes a lot
of difference. Technically you could store state from previous ticks and
answer, but I doubt you would ever want to. Still, it feels like there's
going to be overlap between +scry and +read, so it's not clear that it's
worth baking that into the interface.

belisarius222:
Ah ok, so you're thinking Arvo would consider Behn to be running at tick 104, but it wouldn't tell Behn that -- it would tell Behn it's at 101, since that's the tick to which Arvo has pinned Behn's scry handler.  That sounds cleaner than what I had in mind.

philipcmonk:
Fwiw, while those tasks are valuable, I would release just the first one
initially because it's the most likely to have unexpected effects, and it
will be helpful to know they're not due to any of the other tasks.

belisarius222:
Here's a little deque implementation, which might come in handy.  Tested in the dojo.
https://gist.github.com/belisarius222/8fd8b6b9ee6d0c0ab4144a7f11a2af65

midden-fabler:
Closing in favor of https://github.com/urbit/urbit/pull/6775

PR #6044, opened 2022-10-28
Preserve hints in $flow contents

### LINKED ISSUE #6020: ford caches are not portable, but should be
Ford's global cache, $flow, is a referentially transparent cache of builds, and so intuitively every build it contains should be usable by any other ship.  This has obvious appeal: one can imagine providing a fresh ship with a pill containing a primed $flow in order to reduce boot time, or similarly serving builds to ships à la a Nix binary cache e.g. Cachix, for example.

However, the builds, being vases of cores, are generally not portable in the presence of jet hints.  Producing a vase of a core evaluates its outer formula, eliminating any hints it contains in the process.  Hints that depend on the cores these vases contain then miss their targets, breaking jet registration.

The trick to fix this is known, and involves using `(trap vase)` in place of `vase` to defer outer formula evaluation until the core is on the target ship, at which point jet registration can proceed normally.  This technique is already used to get around the problem in Arvo's larval stage and gen/ivory.hoon.

So: to make caches portable, $flow, and the rest of ford that touches $flow, probably must be converted to deal with `(trap vase)` instead of plain `vase`.  And in addition to just transforming the relevant types and code in ford, clay must procure a zuse in the form of a `(trap vase)` rather than the one wrapped in a vase that's presently used (which is constructed in clay, via zapgar).  Presumably this can be provided via Arvo.

jtobin:
Introduces $ming as a synonym for (trap vase), and converts much of clay's internals to operate on mings instead of vases.  These mings preserve the outer formulas of any cores cached in ford's $flow (already evaluated, if cached as vases), and with them any jet hints that they might contain.  The result is that $flow contents can be used on other ships without breaking jet registration.

These changes partially resolve #6020, in that I don't consider flows to yet be truly portable as long as the cache keys (leaks) they contain are unique only up to zuse.  I'll probably address that in another commit here, or in another PR, by adding the zuse kelvin (or possibly subject hash) to the leak.

I've tested these changes by injecting a ford cache containing the build contents of the %base desk in a pill event, so that the cache is present on the booted ship before apps are installed.  Before these changes, :spider's ++on-arvo arm throws a fund (parent not found) printf on install; with these changes, the hint is registered correctly.  No negative operational or performance issues are noticeable on the booted ship.


pkg/arvo/sys/hoon.hoon
```diff
@@ -11484,11 +11485,33 @@
...
+  =/  cel  [%cell p:$:hed p:$:tal]
+  =>  [cel=cel hed=hed tal=tal]
+  |.  ~+
+  [cel [q:$:hed q:$:tal]]
+::  +shut: deferred +slam
+::
+++  shut
+  |=  [gat=ming sam=ming]
+  ^-  ming
+  =/  typ=type  [%cell p:$:gat p:$:sam]
```
belisarius222, line 11508: gat and sam get fired twice.  I think it would be better to fire them once and store their results in local variables, in case firing is slow for some reason (such as dynamically calculated bunts, which happen sometimes).

jtobin: I think you're stuck with two fires in that you want to defer the Nock evaluation itself.  The first fires have to be done to get the subject type for +mint, which is fine, but you don't want to use the Nock that's produced here, as it would be a kind of early evaluation just "packaged up" in the constructed trap.  Firing again defers the whole Nock computation until it's where it's needed.

(My understanding here comes from how +swat works, but could stand a sanity check.)

jtobin:
With the addition of bc79503 this should properly resolve #6020.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4932,6 +5025,37 @@
     |=  raf=raft-11
     ^-  raft-12
     raf(pud [pud.raf 0 0])
+  ::  +raft-12-to-13:
```
belisarius222, line 5028: @jtobin  Since this is likely to be released after the agents-in-clay work, I think it would make more sense to write the migration into the state for that branch.  Might want to retarget this at philip/agent-clay until that makes it into next/arvo, too.

jtobin: No sweat, can do.

jtobin: I've updated the base to philip/agent-clay; there's a conflict arising in the state migration, of course, but it looks to be easily resolvable on merge -- the base version numbers are to be preferred, and the handlers in +raft-12-to-13 probably should clear the flow/flue caches (or maybe wrapping the contents up in traps is safe for the time being).

jtobin: (I'll probably just merge in the branch and resolve it myself to save some hassle.)

zalberico:
Tagging @joemfb for context, @jtobin may need help debugging the above. This is the last bit to prepare this for release.

jtobin:
Per my commentary in a linked issue:

> I've been running a mainnet ship with an alternative kernel including the changes in https://github.com/urbit/urbit/pull/6044 for some days now & have installed groups, talk, etc. It all seems to be performing without issue, and the updated clay tests pass just fine. It should be ready for internal release.

If there *is* a regression here it's likely to be around performance, rather than some show-stopping ship-bricking thing.  We should be able to better-detect that during broader internal testing, but I haven't noticed anything out of the ordinary.

jtobin:
Looks like CI has started failing due to some mark issues.  Investigating..

jtobin:
The mark build issue is more specifically a result of the following mule call in +build-cast failing:

```
=^  new=ming  nub  (build-fit %mar b)
=/  rab  (mule (swat new tsgl/[limb/a limb/%grab]))
```

As a good minimal test case, notice that from dojo `-build-cast %/noun/urb` succeeds, whereas it fails when using `%/urb/noun`.

jtobin:
Ok, the issue with the mark builds failing is resolved.  Carry on!

jtobin:
PSA, @tadad reported an issue when trying to use these changes to preserve the state of an aqua ship.  We're going to delve into it in the new year.

tadad:
Having an issue trying to port `%aqua` snapshots between ships using this branch
Here are some of the `fund` errors:
```
fund: in on, parent 739d418 not found at 7
fund: in event-core, parent 30be5922 not found at 3
fund: in comp, parent bb4c7ad not found at 15
fund: in handle-start-thread, parent 1de8de1 not found at 7
```
The missing jets are coming from random parts of the system: `+on` is in zuse, `+event-core` is from ames, `+comp` is from hoon.hoon, `+handle-start-thread` is from spider, etc. Most jets are still matching

tadad:
actually now that I understand this PR more - my problem wasn't an issue with ford cache portability - I was moving much more than just the ford cache which led to a lot of jet mismatch errors. This PR is probably fine

jalehman:
Until we know of something that this improves directly, we're going to close with the intent to eventually resurrect when needed.

Issue #6032, opened 2022-10-25
gall: unsubscribing from alien ship causes crash


tinnus-napbus:
**Describe the bug**

You can subscribe to an alien ship just fine. However, if you then try to unsubscribe it will crash.

**To Reproduce**

Here is an agent that does nothing but subscribe and unsubscribe:

```hoon
/+  default-agent, dbug
%-  agent:dbug
=+  ~
|_  =bowl:gall
+*  this  .
    def  ~(. (default-agent this %.n) bowl)
++  on-init  on-init:def
++  on-load  on-load:def
++  on-save  on-save:def
++  on-poke
  |=  [=mark =vase]
  ^-  (quip card:agent:gall _this)
  =+  !<([=ship sub=?] vase)
  :: if not .sub, unsubscribe from .ship
  ?.  sub
    :_  this
    [%pass /wire %agent [ship %foobar] %leave ~]~
  :: if .sub, subscribe to .ship
  :_  this
  [%pass /wire %agent [ship %foobar] %watch /foo/bar]~
++  on-watch  on-watch:def
++  on-agent  on-agent:def
++  on-leave  on-leave:def
++  on-arvo  on-arvo:def
++  on-fail  on-fail:def
++  on-peek  on-peek:def
--
```

I subscribe to an alien ship:

```
> :xxx [~sampel &]
```

it appears in wex as unacked:

```
[p=[wire=/wire ship=~sampel term=%foobar] q=[acked=%.n path=/foo/bar]]
```

I then try unsubscribing:

```
> :xxx [~sampel |]
crud: %belt event failed
[%poke %belt]
  bar-stack
~[
  ~[
    /gall/sys/way/~sampel/foobar
    /gall/use/xxx/0w1.5tK8e/out/~sampel/foobar/7/wire
    /dill
    //term/1
  ]
  ~[/gall/use/xxx/0w1.5tK8e/out/~sampel/foobar/7/wire /dill //term/1]
  ~[
    /gall/use/dojo/0w3.KSbZF/out/~tinnus-napbus/xxx/drum_~tinnus-napbus/poke
    /dill
    //term/1
  ]
  ~[
    / gall
      use
      hood
      0w3.KSbZF
      out
      ~tinnus-napbus
      dojo
      drum
      phat
      ~tinnus-napbus
      dojo
    /dill
    //term/1
  ]
  ~[/dill //term/1]
  ~[//term/1]
]
call: failed
[...]

Issue #6025, opened 2022-10-05
ames crash on assertion in ++on-cork


bacwyls:
**Describe the bug**
In development, I had an agent that sent out a watch poke to an agent on an alien (and unbooted) ship, never got a response, then sent a leave poke. this triggers a cork somewhere and then ames crashes [on this assertion](https://github.com/urbit/urbit/blob/9df08c5e727e2f0dcec8772c5edc65286c5cb00a/pkg/arvo/sys/vane/ames.hoon#L1782).

**Suggested fix**
Maybe it would be safe to just return instead of crashing.

**To Reproduce**
I've reproduced it with a stripped out agent which you can find [here](https://github.com/bacwyls/cork-crash)
This agent just subs and unsubs to an agent on an unbooted ship.
To reproduce, install the agent and poke it twice to sub & unsub to an unbooted peer.

**Expected behaviour**
I would expect these pokes to either sit in a queue or be erased from memory.

**Notify maintainers**
[belisarius222](https://github.com/belisarius222)

belisarius222:
Good catch. I think in this case ames should enqueue the %cork message as an alien todo instead of crashing.

—
~rovnys-ricfer
https://tlon.io

belisarius222:
Yes, that PR resolves this issue.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

zalberico:
Can we merge it then?

PR #6046, opened 2022-11-01
ames: prevent crashes when corking an alien flow

liam-fitzgerald:
The +on-cork handler asserts that the peer is known to us. This is the incorrect behaviour, because it will crash when corking a flow to a peer that is still an %alien. This can happen, for instance, when making a gall subscription for the first time and then corking it before the alien naturalises.

Questions:
- Is no-opping the correct move here? I suspect it is not

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1778,8 +1778,9 @@
     |=  =ship
     ^+  event-core
     =/  ship-state  (~(get by peers.ames-state) ship)
-    ::
-    ?>  ?=([~ %known *] ship-state)
+    ::  XX: should go in alien-agenda maybe?
+    ?.  ?=([~ %known *] ship-state)
+      event-core
     =/  =peer-state  +.u.ship-state
     =/  =channel     [[our ship] now channel-state -.peer-state]
     ::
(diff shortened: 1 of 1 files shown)


belisarius222:
Your XX is right -- this should go in alien agenda.  Otherwise the flow will never close, leaving a space leak.

bacwyls:
#6025

philipcmonk:
This is also #6032

zalberico:
What's the action to merge this? @belisarius222 @liam-fitzgerald - move to alien agenda?

belisarius222:
Yes, the code should be changed to enqueue the cork into the alien agenda along with other commands that will be sent over the wire once we can communicate with the other ship.

—
~rovnys-ricfer
https://tlon.io

belisarius222 (approved):
LGTM

zalberico:
@belisarius222 are we good to merge this then? I'll leave you to make the call and click the button.

zalberico:
@yosoyubik to double check that this does not reload agents - if that's the case then we should merge

yosoyubik:
@zalberico @belisarius222 just applied this to a running ship on next/arvo, and the change only reloads the vane, not the agents.

PR #6047, opened 2022-11-01
zuse: change team:title to return false for moons

bacwyls:
in the past, `+team` meant "our / our moon", but it has been primarily used to represent "our".

Moons as having the full permissions of their parents doesn't make a lot of sense anymore.

I've combed through the uses of `+team` throughout urbit/urbit and I'm quite sure that each instance is better off as just "our".

This looks like the more elegant solution (instead of changing each instance of `+team`)

@philipcmonk

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -5004,12 +5004,16 @@
         /(scot %p our)/sein/(scot %da now)/(scot %p who)
     ==
   ::                                                    ::  ++team:title
-  ++  team                                              ::  our / our moon
+  ::   in the past, this meant "our / our moon", but    ::
+  ::   it has been primarly used to represent "our"     ::
+  ::                                                    ::
+  ::   moons as having the full permissions of their    ::
+  ::   parents doesn't make a lot of sense anymore      ::
+  ::                                                    ::
+  ++  team
     |=  [our=ship who=ship]
     ^-  ?
-    ?|  =(our who)
-        &(?=(%earl (clan who)) =(our (^sein who)))
-    ==
+    =(our who)
   --  ::title
 ::                                                      ::
 ::::                      ++milly                       ::  (2k) milliseconds
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -4928,7 +4928,7 @@
         =<  q.q  %-  need  %-  need
         (rof ~ %j `beam`[[our %sein %da now] /(scot %p who)])
       --
-  ::  middle core: for userspace use, with .^
+  ::  middle core: stateless queries for default numeric sponsorship
   ::
   =>  |%
       ::                                                ::  ++clan:title
@@ -4968,7 +4968,7 @@
           %pawn  (end 4 who)
         ==
       --
-  ::  surface core: stateless queries for default numeric sponsorship
+  ::  surface core: for userspace use, with .^
   ::
   |%
   ::                                                    ::  ++cite:title
@@ -5003,17 +5003,25 @@
         %j
         /(scot %p our)/sein/(scot %da now)/(scot %p who)
     ==
+  ::  +team was created with two meanings:
+  ::    A. her / her moon
+  ::    B. whoever should be able to control her ship
+  ::
+  ::  these two things aren't obviously equal anymore,
+  ::  and it's more important for +team to satisfy B than A,
+  ::  so now +team just means "her".
+  ::
+  ::  (ships can definitely be trusted to control themselves)
   ::                                                    ::  ++team:title
-  ::   in the past, this meant "our / our moon", but    ::
-  ::   it has been primarly used to represent "our"     ::
-  ::                                                    ::
-  ::   moons as having the full permissions of their    ::
-  ::   parents doesn't make a lot of sense anymore      ::
-  ::                                                    ::
-  ++  team
-    |=  [our=ship who=ship]
+  ++  team                                              ::  her
+    |=  [her=ship who=ship]
     ^-  ?
-    =(our who)
+    =(her who)
+  ::                                                    ::  ++moon:title
+  ++  moon                                              ::  her moon
+    |=  [her=ship who=ship]
+    ^-  ?
+    &(=(%earl (clan who)) =(her (^sein who)))
   --  ::title
 ::                                                      ::
 ::::                      ++milly                       ::  (2k) milliseconds


philipcmonk:
I think I agree with this change.

The idea way back when was that the moon would relay everything through its planet, but that's not how moons have ended up being used.

A better solution would be a way to tag some ships (presumably your moons, but not necessarily) as having "team" powers, and using that for things like local devices which are actually supposed to be the same as you, if we ever build out that use-case.

joemfb:
I don't think this change makes sense. `+team:title` was added specifically for this use-case (urbit/arvo#221) -- If it's just `=(our who)`, it doesn't make much sense to keep it around and keep calling it.

I continue to think the use-case of moons controlling the planet is an important one, but it obviously hasn't been a high priority to polish it, and many are attached to using moons in other ways. A configurable `team` would be a great feature, and probably not that hard to build.

philipcmonk:
Making this change allows it to be repurposed later on for a configurable team, while going through and changing everything to `=(our who)` makes that a little harder.  I'm fine with accepting this change to leave that door open for later, or removing +team in a kelvin update and telling app devs to change their call-sites.  I think both of those are better than leaving the status quo where moon usage is handicapped by having these unused open permissions.

bacwyls:
For a configurable team, there would need to be some persisted state that the admin can readily change. It seems to me that this would make the most sense as a gall agent, in which case +team would probably still be deprecated in zuse.

I'm not sure such an agent has a good place in urbit core (perhaps better as a third party app or dependency), and if it does, +team in zuse would have to scry from an agent, which seems like poor practice.

It seems pretty clear to me that +team has been misused. I can't find an instance of it where it wouldn't be better off as `=(our src)` until there is something more like a configurable team.

This is pressing now that there are so many moons on the network. Many of these moons are imperfectly secure, and many of them are children of parent ships containing sensitive data.

lukestiles:
> This is pressing now that there are so many moons on the network. Many of these moons are imperfectly secure, and many of them are children of parent ships containing sensitive data.

This seems important. If a moon can be spun up and effectively have a firewall between it and the planet, that would be a good thing.

joemfb:
@bacwyls the state for a configurable `+team` would belong in `%jael`, the gate would scry for the state (much like `+sein:title` and other arms in that core).

bacwyls:
@joemfb I see. I guess my intuition was that the other arms in title that scry from jael are reading implicit data about keys, whereas this would be additional metadata. But considering this kind of change in jael gets a little outside of my wheelhouse.

eamsden:
I believe this change would result in *more* security difficulties later. If `team:title` is functionally equivalent to `|=(@p =(- our)` then programmers will use it that way, and a later re-expansion to a scried set of "team ships" would introduce greater permissions than intended.

Either proper configurable teams should be built now, or apps should replace `(team:title src)` with just `=(our src)`. Maybe we should offer a linter.

philipcmonk:
We shouldn't build configurable teams into the kernel until we have at least one real use for them.  That's exactly what got us into this mess.

> Either proper configurable teams should be built now, or apps should replace `(team:title src)` with just `=(our src)`.

Both of these are significantly more work, and I'm not sure it's worth the effort compared to what else we could be doing.

"If we're not careful we'll have too lax of permissions in the future" is not a convincing argument against fixing too lax of permissions now.  We can still remove +team later on if we make this change.

joemfb:
The one real use-case is "run your planet in the cloud, moons on your devices, control your planet from your devices". This is not a new idea, it was the primary explanation for moons for a few years, and it's currently encoded in the stdlib (`+team:title`, `+cite:title`, and probably elsewhere in similar patterns). It was most real back when we used `:talk` in the terminal, and has become incrementally less so over time. For it to work well, it can't simply be ad-hoc; it needs to be baked into the system in a pretty fundamental way.

I agree that now (ie, this PR) is not the time to spec or build out this feature or use-case more generally. So I guess if we plan to do so, it makes sense to leave `+team:title` around as a stub for it (there's a good chance we could make it work with the same interface). If we're abandoning the use-case, we should just remove the arm (a bigger diff, but not a difficult change).

So I've talked myself full circle: I think a configurable team is worth prioritizing, and these changes are a fine step towards that. (They should probably include a comment to that effect.)

eamsden:
> "If we're not careful we'll have too lax of permissions in the future" is not a convincing argument against fixing too lax of permissions now. We can still remove +team later on if we make this change.

Sorry, this is an e.g. in my argument, not my argument. My argument is that this change amounts to changing the spec of +team, so that we can change it again later, and amounts to planning instability in the API. The counterargument is "the spec for +team is 'whoever should be able to control your ship.'" If that is the case, we need to put that in a comment and make sure the developer docs are very loud that this has changed.

bacwyls:
I'd encourage anyone thinking about this PR to look through urbit/urbit for +team. all ~70 of the occurrences will be affected by this PR. And I think it's important context for the more broad conversation at hand.

In doing so myself, I noticed [an inaccurate use of +team](https://github.com/urbit/urbit/blob/master/pkg/landscape/app/contact-pull-hook.hoon#L57) which is worth noting, but not of immediate importance. note the wrong order of arguments

matildepark:
It's also fairly common to see it in use in third-party userspace development.

bacwyls:
I agree @eamsden that this amounts to changing the formal spec of `+team`, and it's best not to plan API instability.

Developers are already informally using `+team` as "whoever should be able to control your ship",  rather than strictly `our`.

`+team` is not only failing to meet this expectation, but is also poking security holes in apps. So I think this change makes sense as simultaneously moving towards the more prominent informal spec, and patching up the security issue.

What would everyone think of the following?
```hoon
  ::                                                    ::  ++team:title
  ::  whoever should be able to control your ship       ::
  ::  (just you, for now)                               ::
  ::                                                    ::
  ::  -----------------------------------------------   ::
  ::  +team was created to mean "our / our moon", but   ::
  ::  it has only been used for gating access to admin  ::
  ::  priviledges.                                      ::
  ::                                                    ::
  ::  This presents a risk because people are tending   ::
  ::  use moons in low-security contexts.               ::
  ::                                                    ::
  ::  Thus, +team has been temporarily changed to just  ::
  ::  represent "our".                                  ::
  ::                                                    ::
  ::  This change is planned to be in effect until      ::
  ::  +team can represent a configurable set of admin   ::
  ::  ships.                                            ::
  ::                                                    ::
```

eamsden:
From @xiphiness / ~sarpen-laplux in The Forge - core:

> proposal for team change: break the function signature, making it our + now + src, so you could later build some store to scry that manages your team, but for now just asserts =(src our). this way, it forces any app developers who might be using team for moons to be alerted of the change

I think I'd find this preferable to silently changing it underneath people.

matildepark:
Breaking everyone's apps so they know userspace permissioning is happening?

bacwyls:
I feel that the dev ecosystem is still small enough that no one would be blindsided by this change. The means exist to notify all relevant parties before it goes out.

In which case, the vast majority of devs will just shrug, and one or two of them may have to change a line of code to restore functionality. That seems like a good set of outcomes.

as pointed out by ~novlud-padtyv, `+team` in its originally intended functionality is probably still good to have in zuse.

I wonder if it makes sense to offer a new arm in title with the original functionality of `+team`. then any devs who are legitimately affected by the change would be able to just search&replace. I just cant think of another four letter word.

joemfb:
Adding `now` to the sample is a good idea. You could put the old implementation at `^team:title`, using the same pattern as `+sein` and `+saxo`. (While you're at it, the "middle core" and "surface core" comments are wrong, their positions should be reversed.)

bacwyls:
@joemfb like so?
```hoon
  ::  whoever should be able to control your ship       ::
  ::  (just you, for now)                               ::
  ::                                                    ::
  ::  -----------------------------------------------   ::
  ::  +team was created to mean "our / our moon", but   ::
  ::  it has only been used for gating access to admin  ::
  ::  priviledges.                                      ::
  ::                                                    ::
  ::  This presents a risk because people are tending   ::
  ::  to use moons in low-security contexts.            ::
  ::                                                    ::
  ::  Thus, +team has been temporarily changed to just  ::
  ::  represent "our".                                  ::
  ::                                                    ::
  ::  This change is planned to be in effect until      ::
  ::  +team can represent a configurable set of admin   ::
  ::  ships.                                            ::
  ::                                                    ::  ++team:title
```
wasn't sure how to do the comment because the other title arms use the following style
```
  ::                                                    ::  ++team:title
  ++  team                                       ::  our / our moon
```
and all this info wont fit nicely in the margin

bacwyls:
should i go through and change the 62 `+team` calls across `/pkg/`? otherwise this PR will just break tons of code and each maintainer in pkg will have to add `now` to a function call that does nothing with `now`. editing all these other peoples code almost seems worse than breaking a semantic that would affect none of them.

tacryt-socryp:
Doesn't it seem a little absurd to change this functionality rather than expecting devs to just put a normal equality check `=(who ship)`? `+team:title` doesn't have a broken semantic.

bacwyls:
it definitely seems absurd but i dont think it actually is.

philipcmonk:
@tacryt-socryp None of the proposals here are absurd.  It's easy to claim any of them are, but it's  not helpful.  As I see it they are:

- 1: Leave status quo.  Bad because it's an active vulnerability on the network.
- 2: Yell at people to change to `=(our src)`.  Bad because people likely won't change, so it'll remain an active vulnerability on the network.
- 3: Change/add alongside a version of `+team` which includes `now` for people to use which will one day scry for a configurable team.  Bad because it's building complexity for a feature that may never exist in that form.
- 4: Change the name but recommend people to change to `=(our src)`.  Bad because it requires every app developer to change their code, which a is a lot of time that we could be spending on something useful.
- 5: Change +team to be `=(our src)`.  Bad because some people may rely on this including moons (one known case).

As I see it, 4 and 5 are the only acceptable options.  For anyone currently using +team to mean moons, both require them to change code.  For anyone currently using +team to mean "ships with root access", 4 requires a change and 5 does not.  Both break backward compatibility for at least one person, so probably either should be in a kelvin change.  A kelvin change gives us permission to do that, and we can call out what devs should be expected to change.

tacryt-socryp:
If people are issuing moons willy-nilly, doesn't that seem like a *them* problem? There aren't that many app developers at this point. Many could just be asked to take a look at their code for instances of `+team:title`, update, and get on with it.

bacwyls:
>If people are issuing moons willy-nilly, doesn't that seem like a them problem?

No, I think it should be encouraged. Why would we not want people to give out moons? Because it's a security problem? Ok, let's fix the problem.

philipcmonk:
> If people are issuing moons willy-nilly, doesn't that seem like a *them* problem?

I'm not even talking about giving moons to other people, common usage of moons includes as quarantine zones to install apps, places to develop, places to distribute apps, etc.  None of those should have default access to everything on my planet.

This is a legitimate security problem caused by us using and telling people to use +team:title in a particular way.  I can understand a response like "there are other more important security issues", but not a denial of the problem.

> There aren't that many app developers at this point. Many could just be asked to take a look at their code for instances of `+team:title`, update, and get on with it.

That is not an acceptable way to solve a security problem like this because many won't.  You have to make it either default secure or force them to make an explicit choice.

tacryt-socryp:
So the solution is to burn a kelvin to make the semantics of `+team:title` essentially a pass-through for `=(who ship)` because of a poor development practice / bad past messaging regarding its use? Doesn't that seem extreme?

philipcmonk:
If recommended best practice turns out to be insecure, you can't afterward blame it on the devs who believed you.  Security issues often can't be fixed in a backward-compatible way precisely because the old way was insecure.  Backward-incompatible changes happen during kelvin upgrades.

The outstanding question is whether that change should be to the implementation of +team or to (re)move +team.  Work-wise, it's a trolley problem where one branch requires one person to make a change, and the other requires that person to still make a change but also a bunch of other people to make changes.  But at least we don't have to introduce the concept of deprecation.

bacwyls:
## What should the kernel have to say about moons?

Right now, in a roundabout way, the kernel considers all moons to be admins.

Whether you agree with that or not, this is poorly defined, accidental behavior.

Is it not worth a kelvin to clarify this?

Moons are one of the most basic pieces of urbit. We shouldnt all be sitting around debating about what they are. It should be clearly defined in code.

bacwyls:
BTW, I think all moons should definitely NOT be admins. So to me, its even more worth clarifying.

To any question:
>Should X be admin?

The answer should be:
>Does it have to?

bacwyls:
`+team` and its misuse **implicitly** declare that all moons are admins.

### permissions have to be explicit.

`+team` itself is obviously just a simple util, so the kernel is *technically neutral*.

maybe we can leave it alone and spare the kelvin tick but `+team` calls would have to be replaced almost everywhere and everyone would have to agree to stop misusing `+team`, a simple util about moons, it to mean "root"

belisarius222:
How about this:

We (i.e. Tlon, UF, and the other people commenting on this issue) make a lot of noise about how apps should stop using `+team:title` in their request handlers, and we use social pressure to encourage devs to make the change.  This includes:
- email the urbit-dev mailing list with an all-caps title
- post issues on all known github repos, or even PRs that fix the issue
- tweet about this from all major urbit accounts
- DM all known devs on urbit
- after a short grace period, publish a list of apps known to still have this vulnerability, discouraging people from using them until the hole is closed

Each dev will just need to publish one easy change to a few lines of code; it should be quite a quick process.  For the (relatively few) people who have given moons to people they don't trust, if they're worried the moons could commandeer their planets, they could turn off any apps that haven't been fixed, which should show up on the list of unsafe apps.

This approach has the following advantages:
- no code changes to Arvo, meaning it won't be blocked on other work or block ongoing release efforts
- it doesn't break backward compatibility or burn an extra Kelvin
- it begins to establish a cross-organization social system for dealing with security issues
- it's fairly low-effort and the effort can be easily distributed across many people

Redefining or removing `+team:title` entails a degree of paternalism that I think would only be warranted if some of the following counterfactuals were true:
- if the issue were more severe or widespread (such as something that allowed private data exfiltration from arbitrary ships)
- if the fix involved more than a small amount of time for each dev
- if the expected level of general security were higher
- if there were more app devs, some of whom had published closed-source apps or weren't known to us personally

I think it's easy for those of us who've been around the project for a while to underestimate the importance of maintaining backward compatibility in the standard library -- it's only the past year that it's really mattered at all.  I think among other things, we should treat this as an exercise in learning to avoid unneeded API breakage.

bacwyls:
@belisarius222 I think it's a good idea. Trust me I don't want to be the guy responsible for a kelvin tick everyone hates. and I'm willing to put together the +team -> =(our src) changes in urbit/urbit. to be clear, though, there are many occurrences in pkg/arvo/ that would need patching. just maybe not arvo proper. idk how that terminology works.

also: who will be in charge of DMing devs, keeping a list of +team uses, publishing the list, etc? Maybe UF would be willing put someone on it so core devs can core dev in peace.

philipcmonk:
That's a very bad approach to security, and we really should not get into the habit of thinking we can "solve" security problems by having an open line to every developer.

This API is insecure because of how we've advertised it.  When an API is insecure, you must either change the API to become secure or break the API.  It must fail closed.  If you just tell people to review their usage of it, you *will* miss cases of it.  Are you sure you can get all apps updated?  What about random tutorials on blogs?  Even if we could do it for this issue, it absolutely does not scale to more people developing on Urbit.

Broadly, you seem to be saying that our general level of security and backward compatibility are both not very high, but that we should prioritize getting good at backward compatibility before security.  I think that's not true: security should generally be prioritized over backward compatibility.

Additionally, what you're proposing is much more work than just merging this PR.  Option 4 is also at least 10x the work.  There are countless small security problems like this, and we can't afford to spend that much time on each one.  If a change like this takes that many developer-days, kernel development will completely ossify.

4 and 5 are viable in my mind, and I lean toward 5 because the theoretical advantage of "we removed +team" over "we redefined +team" is very dubious in my mind, and the practical advantage is very concrete.  If you're concerned that someone might not realize the definition of +team is changing, feel free to do the same comms blitz.

tacryt-socryp:
The API isn't insecure. `+team:title` is a basic utility for determining whether a ship is our moon or not. It is documented as such. If you hate the word `+team` and want to make a breaking change, change the name to `+moon` and exclude `our` from its return.

Screen Shot 2022-11-04 at 8 16 43 AM

bacwyls:
@tacryt-socryp 
its the same thing, just way more time and effort.

and i still wouldn't call it a breaking change if we just leave something in the place of +team that works *better* for almost everyone.

the strict, "moon", use of +team is not so useful that it must be spared at all costs.

## counterpoint
I dont think i see +team as "our" in k0. maybe sticks around for a while as scar tissue and someone else figures out what to do about it in 300 years when they have a better idea of what the hell is going on. we can even save them a kelvin for that specific purpose, so they can justify it to their martian boss.

belisarius222:
> we really should not get into the habit of thinking we can "solve" security problems by having an open line to every developer.

If we tried to address every security issue by asking userspace devs to change their code, then yes, of course that doesn't work.

There is another side to this, though: we also shouldn't get into the habit of thinking we need to wade in to every userspace security problem and make breaking changes to the kernel to fix their mistakes.  We can't burn a Kelvin every time some app dev makes a typo.

On one end of the spectrum, there are vulnerabilities like Meltdown/Specter, which is widespread, severe, and for which systems software should take full responsibility; on the other end, there could be bug in some app with just a few users, where they, say, used `min` instead of `max` and opened up a security problem: this affects only the one app, and only a few users, and it's a specific problem.  Clearly the kernel should not redefine `min` to be `max` to fix this.

In my view, given the factors I laid out in my previous comment, userspace devs should fix it themselves.  Not all of them will -- but all of them can, with just a few lines of code.

> Broadly, you seem to be saying that our general level of security and backward compatibility are both not very high, but that we should prioritize getting good at backward compatibility before security. I think that's not true: security should generally be prioritized over backward compatibility.

I don't think either security or backcompat should have a strict categorical precedence over the other.  We'll have to use our judgment on a case-by-case basis, and as I wrote earlier, in this particular case my judgment is that the security issue isn't severe enough to be worth breaking backward compatibility.

For now we're just play-acting at both security and backward compatibility, neither of which meaningfully exist.  We're also play-acting that this is a severe security vulnerability, as if we just discovered Specter/Meltdown -- but it's not.  If nobody did anything at all about this, how many people's data would be exfiltrated over the next, say, three months? 

As far as I know, there's no reason to believe this issue has been exploited at all, despite having existed for at least a year or so.  If that's wrong and lots of people are having problems with this, then I will reverse my opinion on this whole issue.  Otherwise, this isn't some desperate situation where we must take direct, vigorous action to save the network.  If one of those does happen, then of course, we'll do whatever we need to do, and backcompat might be the least of our concerns.

> Additionally, what you're proposing is much more work than just merging this PR. Option 4 is also at least 10x the work. There are countless small security problems like this, and we can't afford to spend that much time on each one. If a change like this takes that many developer-days, kernel development will completely ossify.

You claim my way is "much more work", but that's not true for the kernel.  My way requires zero changes to the kernel -- your way requires one.  Asking userspace devs to fix userspace problems scales better than the kernel stepping in to fix the problems for them.  This is what I meant by "paternalism" in my previous message.  Again, I'm not categorically opposed to it, I'm just not convinced it's the right move here.

In any case, much more work has gone into debating this than will go into whatever code change gets made.  This begs the question: why is there a debate about this at all?  Shouldn't it be a clear case of closing a security hole?  

@philipcmonk In my opinion your strongest point is that this security hole can be closed (at least for the people who perform the kernel update) with a one-line change to the kernel, with only minimal negative externalities.  It's a good point.  I think this debate is occurring because developers strongly dislike backward compatibility being broken.  I don't think you've weighed this heavily enough in analyses of how much work these different options involve, because convincing devs to accept breaking kernel changes evidently also requires work.

I think it feels like a loss, or maybe a betrayal, when the kernel breaks backcompat.  In this case it's more psychological than practical, since there's not that much code involved either way.  Interestingly, @tacryt-socryp's approach that replaces `+team:title` with the slightly different `+moon:title` doesn't irk me the way redefining `+team:title` as an equality check on `our` would.  I think the reason is that the resulting API is at least as good as what we have now (it seems cleaner to me -- and less prone to this form of misuse), whereas if we keep a neutered `+team:title` that isn't useful, we lose functionality.

> 4: Change the name but recommend people to change to =(our src). Bad because it requires every app developer to change their code, which a is a lot of time that we could be spending on something useful.
> 5: Change +team to be =(our src). Bad because some people may rely on this including moons (one known case).
> As I see it, 4 and 5 are the only acceptable options. For anyone currently using +team to mean moons, both require them to change code. For anyone currently using +team to mean "ships with root access", 4 requires a change and 5 does not. Both break backward compatibility for at least one person, so probably either should be in a kelvin change. A kelvin change gives us permission to do that, and we can call out what devs should be expected to change.

In this paragraph, you're lumping two different things into "backward compatibility": 1) user-facing app behavior remaining the same, and 2) dev-facing backward compatibility of the kernel's standard library.  To fix this security issue, user-facing app behavior must change, at least for those people who use moons in certain ways for certain apps; as I've expressed here, I remain unconvinced we need to break kernel backward compatibility for this.

bacwyls:
I think theres another important factor here aside from security and backcompat:

### The value and utility of moons, now and indefinitely into the future.

a user asks:
>is my moon automatically a root user to my planet?

we respond with:
>the kernel doesnt say, exactly. but it provides a util called +team and people tend to use it as meaning "admin" because we told them to. but now we think we shouldnt have told them to. but its ok because we never did anything about it. so moons being admins is really a matter of personal taste-

instead of just:
>No

if theres no clear answer about moon perms, they will be insecure forever. and it might just not be clear enough to suggest everyone to cease a bad habit that is rapidly turning into a defacto standard.

also, the organizational nuances of @belisarius222 proposed solution are still unsolved, and no one has taken responsibility for them.

philipcmonk:
> In any case, much more work has gone into debating this than will go into whatever code change gets made. This begs the question: why is there a debate about this at all? Shouldn't it be a clear case of closing a security hole?

Yes, for sure.  We break backward compatibility on every kelvin upgrade, and nobody bats an eye.  The reason this is a debate is it's a relatively accessible decision in a way that eg #6041 is not.  It's classic bikeshedding.

> @philipcmonk In my opinion your strongest point is that this security hole can be closed (at least for the people who perform the kernel update) with a one-line change to the kernel, with only minimal negative externalities. It's a good point. I think this debate is occurring because developers strongly dislike backward compatibility being broken. I don't think you've weighed this heavily enough in analyses of how much work these different options involve, because convincing devs to accept breaking kernel changes evidently also requires work.

If backward compatibility means 99% of uses must change, of what use was the compatibility?

> The API isn't insecure. +team:title is a basic utility for determining whether a ship is our moon or not. It is documented as such.

That's just not true.  We've constantly told people to use it for permissions checks (has anybody used it for anything else?), and we can't now come back and blame devs for believing us.

>  on the other end, there could be bug in some app with just a few users, where they, say, used min instead of max and opened up a security problem: this affects only the one app, and only a few users, and it's a specific problem. Clearly the kernel should not redefine min to be max to fix this.

This a bad example because in that case app devs made a mistake.  For +team, app devs did nothing wrong, they followed recommended best practice.

> in this particular case my judgment is that the security issue isn't severe enough to be worth breaking backward compatibility....  As far as I know, there's no reason to believe this issue has been exploited at all, despite having existed for at least a year or so.

By this measure, Urbit has no significant security issues.  While not the most impactful security issue, it's one of the most obvious and it's often brought of up when talking about security for that reason.

I'm going to try and lay out my argument point-by-point so we can see precisely which steps we disagree on.

A. Using moons as equal in permissions to their planets by default is a bad idea.

B. We recommend the use of `+team:title` for this purpose ([here](https://developers.urbit.org/guides/core/app-school/7-sur-and-marks#permissions) and many times informally), and [many devs use it for that](https://github.com/search?p=1&q=%22team%3Atitle%22+extension%3Ahoon&type=Code).

C. When in conflict, security should be valued over compatibility.  If we're going to larp one of them, it should be security.

D. Changing our recommendation and doing a comms blitz will not catch all cases of `+team:title` being used incorrectly.

E. In general, a comms blitz is insufficient to communicate a security vulnerability that arises from normal usage of an API.  We have to either deprecate or alter the API.

F. Altering the API is trivial, deprecating it takes much more work.

G. This whole issue is far from the most important thing we could be doing, so the trivial change should be strongly preferred as long as it's not worse than status quo.

matildepark:
A and B -> There could be cases where you’d make apps for ships that do use moons as devices though. Then the utility for doing so simply wouldn’t exist.

(I’d even support team working for stars to planets as planets to moons, if someone really wants it to work like that…)

Sorry, I’ll recuse myself.

tacryt-socryp:
Again, if you care enough about the security concern to break backward compatibility, write a utility of another name called `+moon:title` or something that returns `%.y` for any moon of the input ship and makes them easily accessible and add that as well as the `+team` change.

philipcmonk:
> Again, if you care enough about the security concern to break backward compatibility, write a utility of another name called +moon:title or something that returns %.y for any moon of the input ship and makes them easily accessible and add that as well as the +team change.

That's reasonable.  It's easy enough to change +team like in this PR plus adding a `+moon:title` that has the same behavior as `+team:title` currently has.

tacryt-socryp:
I don't even care if it contains the `=(who ship)` check (in other words, it doesn't need the same behavior). It just needs to be an easy to call, useful utility for telling if the ship is a moon of the ship in question.

ashelkovnykov:
I don't see this as a security issue. I agree that `+team:title` is a neutral utility tool. Any use of it as a security check in a situation where moons should not have permissions is a usage bug. Such cases in the kernel should be fixed. Such uses in userspace code should be discouraged.

My preferred solution is a lower-cost version of what @belisarius222 suggests:
- Send out an email to the dev mailing list warning devs to double-check their intent
- Post message in `[battery payload]` dev group
- Add to "best practices" documentation
- Make a note of it for the future (for audits or some such)

TL;DR - SWIM AT OWN RISK

bacwyls:
ig my main point is that the status quo is so bad that it threatens to make moons useless forever if not properly addressed.

i like the idea of adding `+moon:title` a lot.

As a slight improvement on `+team`, I suggest that its documented as "her / her moon" rather than our.

philipcmonk (approved):
Based on the above discussion, I think this is the right solution.

PR #6048, opened 2022-11-02
Improved eval code

mopfel-winrux:
I added the ability for the user pick two flags to pass to the eval subroutine. 

The first is the `-j` flag. This takes the result from the inputted hoon code , `+jam`'s it and outputs it as a hexadecimal string. 

The second is the `-jk` flag. This takes the result from the inputted hoon code, `+jam`'s it and outputs the noun wrapped in the proper `%khan` header (leading zero, noun size, noun) as binary data.

The proper use of the eval is as follows:
`echo "(add 2 2)" | urbit eval -jk > test.jam`

This should be the easiest way to `+jam` a noun for `%khan` use

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -1012,15 +1012,21 @@ static void
 _cw_eval(c3_i argc, c3_c* argv[])
 {
   c3_o jam_l;
+  c3_o khan_l;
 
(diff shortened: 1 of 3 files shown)


jalehman:
Closing this because it's targeting the wrong repo. A new PR has been created.

PR #6051, opened 2022-11-04
Groups/Garden: Add s3 region to s3 settings

patosullivan:
Allows us to use this field in the new garden/grid: https://github.com/urbit/garden/pull/13

We also need to talk about pulling this out of the %landscape desk at some point.


arthyn (approved):
lgtm, @liam-fitzgerald do we need to do any mark versioning for this and also how should these changes get released?

liam-fitzgerald (approved):
LGTM!

Issue #6052, opened 2022-11-04
`+*` is overloaded and the second syntax can mostly be replaced by `|$`


drbeefsupreme:
First, see https://github.com/urbit/developers.urbit.org/issues/218

The "second syntax" I am referring to in the title is the one present in `+boog`:

https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/hoon.hoon#L13256-L13279

The `+*` that is almost always used, at the top of doors, is instead:

https://github.com/urbit/urbit/blob/master/pkg/arvo/sys/hoon.hoon#L13316-L13323

As far as I can tell, these are two completely different use cases, and using `+*` for both of them feels to me like it has to have been some kind of oversight. The only place in the base distribution that uses the second syntax is here: https://github.com/urbit/urbit/blob/master/pkg/base-dev/lib/language-server/complete.hoon#L6-L7

and it seems to work fine by replacing those two lines with:
```hoon
++  option  |$  [item]
  [term=cord detail=item]
```
and everything compiles when removing the `+*` parser from `+boog`.

I'm not entirely sure what the point of this `+*` is. The only real difference between it and `|$` that I can discern is the presence of the `%made` `spec`. I'm not entirely sure what this is for, but I think @joemfb explained it to me once.

I think it is pretty straightforward that having two syntaxes that do totally different things for `+*` is wrong, especially given that the common one isn't even a real arm, while this second one is a real arm. A couple solutions off the top of my head:

1.  Change the second syntax to be a different lus digraph, e.g. `+#` (not suggesting hax specifically)
2.  Remove the second syntax entirely and change the one instance of `+*` being used in the `%base` distribution to use `|$` instead. There is a possibility that some developers are out there is making use of this, so probably we'd want to perform an OTA including a printf saying that it will be deprecated first.

The first one leans back in the direction of having more arm digraphs that we moved away from a few years ago. But it maintains the ability to use the `%made` spec (whose purpose is unknown to me).

The second one makes it impossible to use `%made` except when building a `spec` by hand. Possibly this is fine for now, if `%made` isn't yet serving any purpose.

Until I know what `%made` is for, I don't want to favor one option or the other (and there may be other solutions that didn't occur to me).

philipcmonk:
Relevant: urbit/arvo#365 #1654 #1712

joemfb:
[...]

PR #6054, opened 2022-11-04
language server: change +* to |$

drbeefsupreme:
See #6052. This is the only use of the deprecated syntax of `+*` I was able to find in the repository.

I was going to make the changes to `+boog` to remove the `+*` parsing in the same PR, and that works fine on a ship with only `%base`. But Clay seems to run into issues with this on a regular solid pill fake ship with

```
> .^((set desk) %cd %)
{%bitcoin %base %landscape %webterm %garden %kids}
```

I haven't figured out why yet. Even though `+cat %/lib/language-server/complete/hoon` shows the change in the commit here, it still says its running into a syntax error as though it hadn't changed. I'll be adding the stack trace to #6052. In any case, it makes me pretty sure that we need to change the existing instance of `+*` first before changing `hoon.hoon`.


drbeefsupreme:
Argh branched off of master again and now there's extra stuff, will remake

PR #6055, opened 2022-11-04
language server: change +* to |$

drbeefsupreme:
See https://github.com/urbit/urbit/issues/6052.

This is the only use of the deprecated syntax of `+*` I was able to find in the repository.

I was going to make the changes to `+boog` to remove the `+*` parsing in the same PR, and that works fine on a ship with only `%base`. But Clay seems to run into issues with this on a regular solid pill fake ship with

```
> .^((set desk) %cd %)
{%bitcoin %base %landscape %webterm %garden %kids}
```

I haven't figured out why yet. Even though `+cat %/lib/language-server/complete/hoon` shows the change in the commit here, it still says its running into a syntax error as though it hadn't changed. I'll be adding the stack trace to https://github.com/urbit/urbit/issues/6052. In any case, it makes me pretty sure that we need to change the existing instance of `+*` first before changing `hoon.hoon`, thus the separate PR.

### DIFF
--- a/pkg/base-dev/lib/language-server/complete.hoon
+++ b/pkg/base-dev/lib/language-server/complete.hoon
@@ -3,7 +3,8 @@
 ::
 =/  debug  |
 |%
-+*  option  [item]
+++  option
+  |$  [item]
   [term=cord detail=item]
 ::
 ::  Like +rose except also produces line number


drbeefsupreme:
Per https://github.com/urbit/urbit/issues/6052#issuecomment-1304343002 the reason this doesn't work with the removal of the `+*` parser in `+boog` (even after this change is committed) is because that requires a Kelvin upgrade. But since there's no anticipated Kelvin change for `hoon.hoon` on the horizon as far as I know, it makes sense to push this update separately in the meantime just so we've completely eliminated use of the old `+*` syntax in the default solid pill.

belisarius222 (approved):
LGTM

PR #6057, opened 2022-11-06
kiln: recursive `|rm`

ryjm:
Increments the commit number every time, is there a better way to do this?

image

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -1259,8 +1259,8 @@
   =|  c=(list (unit toro))
   %+  poke-info  "removed: {<a>}"
   =-  %+  roll  -
-  |=  [a=(unit toro) b=(unit toro)]
-  (clap a b furl)
+      |=  [a=(unit toro) b=(unit toro)]
+      (clap a b furl)
   |-  ^-  (list (unit toro))
   =+  b=.^(arch %cy a)
   ?^  fil.b  (snoc c `(fray a))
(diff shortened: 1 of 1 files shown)


ryjm:
ah looks like `furl` might be what i'm looking for.

pkg/arvo/lib/hood/kiln.hoon
```diff
@@ -1256,11 +1256,18 @@
 ::
 ++  poke-rm
   |=  a=path
+  =|  c=(list (unit toro))
+  %+  poke-info  "removed: {<a>}"
+  =-  %+  roll  -
+  |=  [a=(unit toro) b=(unit toro)]
```
belisarius222, line 1262: This line and the next should be indented twice to match the `%+  roll`

zalberico:
@ryjm can you clarify what this is actually for or write up and link it to an issue about what problem this is solving? As is it's hard for someone to evaluate and approve/merge.

belisarius222:
|rm removes files from Clay, just like Unix "rm". It's a cleaner interface if it's recursive, in the sense that it deletes all files whose paths start with the supplied path (like "rm -r"), but it wasn't recursive until this PR.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

zalberico:
@belisarius222 thanks - is this approved/good to merge into next/arvo? We shouldn't be merging straight into master outside of releases.

belisarius222:
Not ready for merge, still needs to handle the case where fil and dir are both non-null (I just talked to Jake about this). Should also target next/arvo, not master

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

zalberico:
@belisarius222 looks like @ryjm made the changes here, can you take a look?

philipcmonk (approved):
This works for me

PR #6063, opened 2022-11-07
vere: implements demand paging

joemfb:
This PR implements demand paging for the loom. Thanks to @ritfed-tabryg for the suggestion, @eamsden for many fruitful design discussions, and @mcevoypeter for the original implementation in #5940.

Demand paging is implemented on linux and macos, and turned on by default (disabled by `--no-demand`). It remains highly desirable on windows (where memory is not overcommitted), but will require changes to our "platform abstraction layer" (lol) to support CoW file-backed mappings, and further changes to the implementation to explicitly remove old mappings before establishing new ones.

The changes here differ from #5940 in key ways. The north segment of the loom (home-road heap, `north.bin`) is still in a CoW file-backed memory map, but that mapping is now reestablished after every snapshot update, ensuring that clean pages are legible to the OS and never swapped. Attempting that was always planned, but the mapping logic in #5940 was staged on/with other irrelevant changes, and needed to be moved to handle re-mapping. Once moved, it became clear it was practically equivalent to coalescing the memory protection changes on snapshot update, so that has been done as well. Mapping requires that the north and south segments be handled separately; this in turn allows the entire north segment to be read in one syscall (retried appropriately as per #6062). With or without demand paging, the number of system calls involved in updating the snapshot has been dramatically reduced. And all of these changes have been made with out compromising the generality of the north/south snapshot system for the loom.

NB: the initial, minimal implementation in #5940 was motivated by an abundance of caution with regards to the efficiency of re-mapping the loom. Further research of virtual memory (and experimentation) with the changes here has made it clear that that was unwarranted.

Next steps include adding windows support, and respecting the `--no-demand` argument in all relevant maintenance commands (ie, `meld`, `next`, `pack`, `prep`, &c).

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -159,6 +159,15 @@ _main_init(void)
   //
   u3_Host.ops_u.has = c3y;
 
+  //  demand paging (ie, file-backed mapping for the loom)
+  //  is not yet supported on windows
+  //
+#ifdef U3_OS_mingw
+  u3_Host.ops_u.map = c3n;
+#else
+  u3_Host.ops_u.map = c3y;
+#endif
+
   u3_Host.ops_u.net = c3y;
   u3_Host.ops_u.lit = c3n;
   u3_Host.ops_u.nuu = c3n;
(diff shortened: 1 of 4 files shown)


eamsden:
:confetti_ball:

Issue #5783, opened 2022-05-24
Printing for spec/type doccords


drbeefsupreme:
Printing these should be easy, the hard part is implementing the doccords for them,  #5768

PR #6064, opened 2022-11-07
hoon: remove old lustar parser

drbeefsupreme:
See #6052. This is completely different from the `+*` used at the top
of doors, and has almost entirely been replaced by `|$`. The exception is
the use of the `%made` spec, not present in `|$`. I do not see an
obvious way to change `|$` to use `%made` since this `+*` parser uses
the name of the arm in the `%made` structure, unless we change the
AST of `|$`.

This commit requires a burning a kelvin on live ships (see #6052), so
that is why this commit also bumps the kelvin down. This opens up
possibilities on how to change `|$` to use `%made` if we decide that
change is warranted.

Blocked by #6055.

Because we might want to change `|$`, and #5873 will also be
changing `+boog`, this is currently marked as a draft and probably won't
need attention again until we're thinking about burning a kelvin on
`hoon.hoon`.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -2,11 +2,11 @@
 ::::    /sys/hoon                                       ::
   ::                                                    ::
 =<  ride
-=>  %140  =>
+=>  %139  =>
 ::                                                      ::
 ::::    0: version stub                                 ::
   ::                                                    ::
-~%  %k.140  ~  ~                                        ::
+~%  %k.139  ~  ~                                        ::
 |%
 ++  hoon-version  +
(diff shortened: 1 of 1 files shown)


zalberico:
Note to self that we should have some mechanism by which we can see all the pending kelvin version PRs so we can be a little smart about merging them together (maybe just a kelvin tag).

drbeefsupreme:
I've confirmed that I'll need to decrement a Hoon kelvin for #5783 so this is just getting folded into that one, since it is already changing the parser involved.

drbeefsupreme:
5b99f74c940ed7785ac37936e6ceae7b6edb5122

PR #6068, opened 2022-11-09
drum, dojo: preserve input history after crash

ryjm:
When a crash is triggered (e.g from a bad scry) `%drum `deletes input history and `%dojo` clears out defined variables. This saves the latter in the `%drum` state instead, retrieving it when the app is relinked, and inserts the old variables into the new session for the former.

### DIFF
--- a/pkg/arvo/lib/hood/drum.hoon
+++ b/pkg/arvo/lib/hood/drum.hoon
@@ -42,6 +42,7 @@
       off=@ud                                           ::  window offset
       kil=kill                                          ::  kill buffer
       inx=@ud                                           ::  ring index
+      hiy=(map gill:gall history)                        :: ancient input
       fug=(map gill:gall (unit target))                 ::  connections
       mir=(pair @ud stub)                               ::  mirrored terminal
   ==                                                    ::
@@ -302,6 +303,8 @@
   ^+  +>
   =+  lag=se-agon
   ?.  (~(has by fug) gyl)  +>.$
+  =/  olt=target  (fall (~(got by fug) gyl) *target)
+  =.  hiy  %+  ~(put by hiy)  gyl  (fall `hit.olt *history)
   =.  fug  (~(del by fug) gyl)
   =.  eel  ?.(pej eel (~(del in eel) gyl))
   =.  +>.$  ?.  &(?=(^ lag) !=(gyl u.lag))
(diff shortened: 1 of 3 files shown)


ryjm:
I'm not sure how to handle the hood state update though, open to suggestions.

zalberico:
@belisarius222 tagging you in this one for review

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -302,6 +303,8 @@
   ^+  +>
   =+  lag=se-agon
   ?.  (~(has by fug) gyl)  +>.$
+  =/  olt=target  (fall (~(got by fug) gyl) *target)
+  =.  hiy  %+  ~(put by hiy)  gyl  (fall `hit.olt *history)
```
Fang-, line 307:
```suggestion
  =/  olt=target  (fall (~(got by fug) gyl) *target)
  =.  hiy         (~(put by hiy) gyl hit.olt)
```

pkg/arvo/lib/hood/drum.hoon
```diff
@@ -42,6 +42,7 @@
       off=@ud                                           ::  window offset
       kil=kill                                          ::  kill buffer
       inx=@ud                                           ::  ring index
+      hiy=(map gill:gall history)                        :: ancient input
```
Fang-, line 45:
```suggestion
      hiy=(map gill:gall history)                       :: ancient input
```

Fang- (commented):
The dojo change is simple and fine. The drum change, I feel a little weird about keeping these things around in state for the whole duration of the session. On the other hand, maybe having the history still there even after you intentionally un- and re-link is kinda nice.

iirc there was some concern about _wanting_ to clear dojo variables in case of disconnect, because "what if those variables caused the disconnect!", but afaict this is entirely theoretical. It's hard for me to imagine a case where this will lock your dojo up. Happy to hear counterarguments, but not gonna hold this back for now.

Some nits, but otherwise should be fine. As for the hood state update, what's wrong with the pattern that's already in-place? Isn't it sufficient to add a `state-5` in drum (yes, tedious), expand drum's `+on-load`, and then add the appropriate `%25` state in hood?

Fang-:
Actually, the alternate approach raised by ~sarpen-laplux [in the forge](web+urbitgraph://group/~middev/the-forge/graph/~middev/guild-hall-8426/170141184505922968510705979284506279936) might be preferable to this. ie handling crashes more gracefully in the first place. We can discuss out of band if you want.

belisarius222:
Yeah, I think ~sarpen's approach should solve this cleanly without potentially introducing drum/dojo synchronization issues.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

zalberico:
> Yeah, I think ~sarpen's approach should solve this cleanly without potentially introducing drum/dojo synchronization issues.

Is there an issue for that? I think the next move is to write up whatever that approach is, link it here, and then close this.

belisarius222:
Joe posted a (rough) writeup earlier in these comments.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

jalehman:
Closing in favor of an eventual contribution that implements ~sarpen-laplux's solution.

PR #6070, opened 2022-11-10
lens: do not jam export output

liam-fitzgerald:
The previous implementation of the /export-all scry jammed it's output, but the runtime jams the result when using -X and -Y. This resulted in the archive being jammed twice, which runs into #5674. This commit removes the jam on the %lens side, which fixes importing if the runtime defers cueing until the event has been injected into lens


zalberico:
I'm going to close this then based on the comment here and on #6071

PR #6071, opened 2022-11-10
fore: remove +cue before import

liam-fitzgerald:
Companion commit to 29ce47a006d27465a8d858a2b28d26ebae81adeb. By removing the double jam and double cue, we work around #5674. Makes imports work again.

### DIFF
--- a/pkg/urbit/vere/io/fore.c
+++ b/pkg/urbit/vere/io/fore.c
@@ -78,7 +78,7 @@ _fore_inject(u3_auto* car_u, c3_c* pax_c)
 static void
 _fore_import(u3_auto* car_u, c3_c* pax_c)
 {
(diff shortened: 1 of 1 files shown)


zalberico:
Thanks @liam-fitzgerald - @philipcmonk, assigning this one to you for review/merge.

joemfb:
@liam-fitzgerald if this (and #6070) just working around #5674, that could be done by just changing the jam call site in `pier.c` and using `u3s_jam_xeno()`. That's probably better, since the bug and workaround would both be isolated to the runtime (and double jamming is pretty cheap, basically just a shift). Unless you already want to change the import interface ...

joemfb:
The workaround suggested above was released in urbit-v1.12 (e1e0ca76d8).

Issue #6050, opened 2022-11-04
ames: fragment crashed %exit


yosoyubik:
```
ames: ~wolref-podlex: hear ~wolref-podlex seq=1 1kb
ames: ~wolref-podlex: sink naxplanation [~wolref-podlex bone=27.366 message-num=1]
ames: ~wolref-podlex fragment crashed %exit
/sys/vane/ames/hoon:<[3.027 5].[3.097 7]>
[bone=27.364 mnum=2 next=1 unsent-messages=0 unsent-fragments=0 any-live=%.n]
/sys/vane/ames/hoon:<[3.020 5].[3.097 7]>
/sys/vane/ames/hoon:<[3.017 5].[3.097 7]>
/sys/vane/ames/hoon:<[2.984 14].[2.984 80]>
/sys/vane/ames/hoon:<[2.951 5].[2.985 7]>
/sys/vane/ames/hoon:<[2.949 5].[2.985 7]>
/sys/vane/ames/hoon:<[2.939 9].[2.939 29]>
/sys/vane/ames/hoon:<[2.939 5].[2.944 7]>
/sys/vane/ames/hoon:<[2.937 5].[2.944 7]>
/sys/vane/ames/hoon:<[2.612 47].[2.612 71]>
/sys/vane/ames/hoon:<[2.612 7].[2.722 9]>
/sys/vane/ames/hoon:<[2.611 7].[2.722 9]>
/sys/vane/ames/hoon:<[2.610 7].[2.722 9]>
/sys/vane/ames/hoon:<[2.607 7].[2.722 9]>
/sys/vane/ames/hoon:<[2.604 7].[2.722 9]>
/sys/vane/ames/hoon:<[2.857 24].[2.857 73]>
/sys/vane/ames/hoon:<[2.857 9].[2.866 18]>
/sys/vane/ames/hoon:<[2.854 9].[2.866 18]>
/sys/vane/ames/hoon:<[2.851 9].[2.866 18]>
/sys/vane/ames/hoon:<[2.848 9].[2.866 18]>
/sys/vane/ames/hoon:<[2.844 9].[2.866 18]>
/sys/vane/ames/hoon:<[2.843 9].[2.866 18]>
/sys/vane/ames/hoon:<[2.745 22].[2.745 63]>
/sys/vane/ames/hoon:<[2.744 13].[2.748 15]>
/sys/vane/ames/hoon:<[2.743 11].[2.749 37]>
[...]

Issue #6075, opened 2022-11-12
ames: ack crashed %exit


tinnus-napbus:
**Describe the bug**
I'm getting a fair few messages like `ames: ~sicdev-pilnup ack crashed %exit` after I upgraded to v1.11 aarch64. I've also seen the following ames crash a couple of times:

```
[%poke %crud]
bar-stack=~[~[//ames]]
call: failed
/sys/vane/ames/hoon:<[1.082 3].[1.113 20]>
/sys/vane/ames/hoon:<[1.084 3].[1.113 20]>
/sys/vane/ames/hoon:<[1.085 3].[1.113 20]>
/sys/vane/ames/hoon:<[1.087 3].[1.113 20]>
/sys/vane/ames/hoon:<[1.088 5].[1.111 7]>
/sys/vane/ames/hoon:<[1.091 5].[1.111 7]>
/sys/vane/ames/hoon:<[1.092 7].[1.095 9]>
/sys/vane/ames/hoon:<[1.094 16].[1.094 60]>
/sys/vane/ames/hoon:<[1.553 5].[1.553 43]>
/sys/vane/ames/hoon:<[1.559 5].[1.577 17]>
/sys/vane/ames/hoon:<[1.560 5].[1.577 17]>
/sys/vane/ames/hoon:<[1.563 5].[1.577 17]>
/sys/vane/ames/hoon:<[1.566 5].[1.577 17]>
/sys/vane/ames/hoon:<[1.660 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.661 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.667 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.675 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.676 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.677 5].[1.712 62]>
[%ames-crash-on-packet-from ~sicdev-pilnup]
/sys/vane/ames/hoon:<[1.678 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.682 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.711 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.712 5].[1.712 62]>
/sys/vane/ames/hoon:<[1.712 10].[1.712 62]>
/sys/vane/ames/hoon:<[2.389 7].[2.418 69]>
[...]

Issue #5979, opened 2022-08-29
ack crashed %exit on recork timer


zalberico:
ames: ~bitpyx-dildus ack crashed %exit

Output like the above appears when the recork timer runs (currently every 20 minutes).

zalberico:
We started seeing these after shipping the GRQF.

It doesn't appear to be causing any obvious issue, but we should understand why they're appearing while the GRQF is fresh in our minds.

philipcmonk:
Took a brief look at my version of this.  You can get a trace with commands `|ames-sift ~bitpyx-dildus` and `|ames-verb %msg` (turn off with `|ames-verb`).  I get one that looks like:

```
[bone=8 message-num=1 meat=[%.n p=[%.n p=[ok=%.n lag=~s0]]]]
decrement-underflow                        
/sys/vane/ames/hoon:<[3.526 19].[3.526 33]> 
/sys/vane/ames/hoon:<[3.526 5].[3.554 12]> 
/sys/vane/ames/hoon:<[3.525 5].[3.554 12]>                              
/sys/vane/ames/hoon:<[3.523 5].[3.554 12]>                  
/sys/vane/ames/hoon:<[3.448 35].[3.448 55]>                             
/sys/vane/ames/hoon:<[3.448 27].[3.448 55]> 
/sys/vane/ames/hoon:<[3.448 7].[3.448 56]>                              
/sys/vane/ames/hoon:<[3.447 5].[3.451 36]>  
/sys/vane/ames/hoon:<[3.443 5].[3.451 36]>                                                                
/sys/vane/ames/hoon:<[3.440 5].[3.451 36]> 
/sys/vane/ames/hoon:<[3.438 5].[3.451 36]> 
/sys/vane/ames/hoon:<[3.433 5].[3.451 36]>  
/sys/vane/ames/hoon:<[3.431 5].[3.451 36]> 
/sys/vane/ames/hoon:<[3.425 5].[3.451 36]>  
/sys/vane/ames/hoon:<[3.423 5].[3.451 36]>
/sys/vane/ames/hoon:<[3.204 14].[3.204 40]>
/sys/vane/ames/hoon:<[3.202 5].[3.208 7]>                               
/sys/vane/ames/hoon:<[3.200 5].[3.208 7]>  
/sys/vane/ames/hoon:<[3.198 5].[3.208 7]>  
/sys/vane/ames/hoon:<[3.148 7].[3.148 42]>  
[...]

PR #6076, opened 2022-11-12
ames: remove num-live from pump-metrics

yosoyubik:
Tested this on two different live moons and saw no communication issues: they can talk to each other, their sponsor, install apps, join/leave/rejoin groups.

Addresses decrement-underflow error seen in #6075 and https://github.com/urbit/urbit/issues/5979#issuecomment-1275723988

Note: this updates `%lull`

### DIFF
--- a/pkg/arvo/app/dbug.hoon
+++ b/pkg/arvo/app/dbug.hoon
@@ -716,7 +716,7 @@
                   'rtt'^(numb (div rtt ~s1))
                   'rttvar'^(numb (div rttvar ~s1))
                   'ssthresh'^(numb ssthresh)
-                  'num-live'^(numb num-live)
+                  'num-live'^(numb ~(wyt by live))
                   'cwnd'^(numb cwnd)
                   'counter'^(numb counter)
               ==
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -657,6 +657,7 @@
   ::    rto: retransmission timeout
   ::    rtt: roundtrip time estimate, low-passed using EWMA
   ::    rttvar: mean deviation of .rtt, also low-passed with EWMA
+  ::    num-live: how many packets sent, awaiting ack
   ::    ssthresh: slow-start threshold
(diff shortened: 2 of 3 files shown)


belisarius222 (dismissed):
LGTM.  I'm sad we have to do this, since it worsens congestion control asymptotics, but it is the fastest way to fix the bug.  A good goal is a refactored Ames that will be so much easier to reason about that tracking num-live won't be bug-prone.

zalberico:
@belisarius222 and @yosoyubik - I want to merge fixes like this once approved into `next/arvo` (and deploy them on the internal canary) unless there's some specific special reason not to.

zalberico:
Does this also resolve #6050 or is that something different?

yosoyubik:
@zalberico #6050 is another issue (we are getting a nack response for a message we don't have in our state anymore). 

Re this PR, once I revert https://github.com/urbit/urbit/pull/6076/commits/b2a0f2fd304f961e4db89499aedefe0be95e6239 (se we don't update %lull yet), it'll be ready to be merged into next/arvo

yosoyubik:
@belisarius222 Just reverted the commit that updated %lull and updated %ames' state—num-live is still in the pump but it's unused. This doesn't reload anything besides the vane itself.

@zalberico Once I get an approval, I'll merge into next/arvo.

zalberico:
Great - thanks, the next part of this is we need to be deploying the tip of next/arvo to our internal canary weekly (and we need kernel devs to be sourcing their OTAs from that canary)

yosoyubik:
@zalberico I can take care of deploying to ~binnec (it's that still the dev canary?)—what has the deployment cadence been to give people a heads up?. Also, do we want to deploy both ames PRs, or do we want a more granular deployment?

yosoyubik:
@belisarius222 checking that this looks good to go into next/arvo—latest commit don't update %lull, so they leave num-live unused (we can take care of the state migration later with some of the other %ames refactoring work, like removing krocs, for example, since it's not really used anyway)

belisarius222 (approved):
lgtm

PR #6077, opened 2022-11-12
WIP 8G Loom

barter-simsum:
In fulfillment of https://urbit.org/grants/loom-pointer-compression

This is a WIP.

`|pack` like migration work needs to be completed to enable users
to migrate from 2G/4G -> 8G. This should be opt-in.

Note, comments beginning with `;;:` are notes to myself and you. They will be
deleted before merge. They might ask a question or explain something.


zalberico:
Thanks for writing this up - assigning @joemfb and @belisarius222 to this for context

nix/pkgs/urbit/default.nix
```diff
@@ -80,6 +81,6 @@ in stdenv.mkDerivation {
 
   meta = {
     debug = enableDebug;
-    arguments = lib.optionals enableDebug [ "-g" ];
+    arguments = lib.optionals enableDebug [ "-g3" ];
```
barter-simsum, line 84: the other one

pkg/urbit/noun/retrieve.c
```diff
@@ -1029,63 +1029,46 @@ c3_w
...
-        return ((gal_w + 1) + ((1 << gow_y) - 1)) >> gow_y;
-      }
-    }
+    return (gal_w << (5 - a_y))
+      + (c3_bits_word(daz_w)
+         + max_y
+         >> a_y);
   }
+  c3_y gow_y = (a_y - 5);
+  return ((gal_w + 1) + ((1 << gow_y) - 1)) >> gow_y;
```
barter-simsum, line 1070: I'm not sure this shouldn't just be `gal_w + 1 >> gow_y;` A fakezod boots and runs fine with that definition. Though compared to the previous implementation, this _does_ return different results. Yes, differing results are encountered in the boot process - and rather quickly - so it's not a matter of incomplete test input.

barter-simsum: @joemfb Is there some hoon I could use to catastrophically crash this if the line referenced were changed to `gal_w + 1 >> gow_y;`? I'm just surprised a fakezod booted fine

barter-simsum: And note comments in https://github.com/urbit/urbit/pull/6077/commits/a10e50ac210a022a7a5c72cd7b6457528ed3259a . In current implementation, this _does_ match exactly

joemfb: You could turn on jet equivalence testing for `+met` (in `jets/tree.c`, change `c3y` to `c3n` for `_140_two_met_a`). Or you could boot a fake ship with one version of `u3r_met()`, then delete the snapshot replay it with the other and watch for the mug mismatches.

joemfb: Generally this kind of thing is hard (but not impossible) to detect from hoon -- you're messing with the laws of physics down here.

barter-simsum: Thanks! Actually, I guess I can't say it's an exact match when I also say:

""
Moreover, this change does not artificially limit the size that gal_w can
be. For instance, in the previous implementation, for a value of a_y=2, gal_w
could not exceed the following without bailing:

`(UINT32_MAX - 35) >> 5` =>
`0x07FFFFFF`

Now, gal_w cannot exceed

`((UINT32_MAX - (32 + max_y)) >> (5 - a_y))` =>
`((UINT32_MAX - 35) >> 3)` =>
`0x1FFFFFFF`
""

Maybe "an exact match" was the actual reason for the prior switch case implementation and not optimization?

barter-simsum: So a jet can bail when the equivalent nock wouldn't? In that case, there shouldn't be anything wrong with my change, though I'm running it through the testing you suggested.

barter-simsum: Ok, confirming that the following doesn't show any mug mismatches:

1) boot fakezod with current branch (`u3r_met` changes)
2) delete chk and bhk
3) recompile with old definition of `u3r_met`
4) recompute events -- no mug mismatches

And vice-versa

Though, given your comments above, I certainly _would_ expect mug mismatches if I replaced the last return with `gal_w + 1 >> gow_y;`

barter-simsum: Aside from generalizing and reducing instruction size, the only functional change the new `u3r_met` introduces is it won't bail as early for `a_y` values of 1 or 2.

For instance, for `a_y=2`, previous would bail when `gal_w` exceeds `0x07FFFFFF`. Now it would bail when `gal_w` exceeds `0x1FFFFFFF` (~4x the allowable length)

pkg/urbit/noun/events.c
```diff
@@ -404,29 +404,30 @@ _ce_patch_delete(void)
 static c3_o
 _ce_patch_verify(u3_ce_patch* pat_u)
 {
-  c3_w i_w, pag_w, mug_w;
-  c3_w      mem_w[pag_wiz_i];
-  size_t    off_i, siz_i = pag_siz_i;
-  ssize_t   ret_i;
+  c3_w  pag_w, mug_w;
+  c3_w  mem_w[pag_wiz_i];     /* ;;: This is a VLA, you know... */
```
joemfb, line 408: That's a very good point. It didn't used to be, but was switched (unintentionally I think) in #5881.

barter-simsum: And obviously the same goes for any other arrays declared with `pag_wiz_i` or `pag_siz_i`. Instances in `_ce_patch_apply`, `_ce_image_fine`, `_ce_image_copy`, etc. In this case, yeah a macro actually is better. Or we could adopt c23 and start using the new `constexpr` type. ;)

mcevoypeter: A `const` variable and a `#define` generate identical assembly when compiled with `-O2` or greater, so it's not `mem_w` isn't a variable-length array in release builds. That said, you're right that a macro is probably preferable.

Here's a simple example demonstrating that `const` and `#define` generate identical assembly at `-O2` and above:

```console
$ gcc --version
gcc (GCC) 12.2.0
Copyright (C) 2022 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
$ cat << EOF > main.c
#if defined(CONST)
static const unsigned int len = 4;
#elif defined(MACRO)
# define len 4
#endif

int
main(int argc, char *argv[])
{
  char arr[len];
  arr[0] = 'a';
  arr[1] = 'b';
  arr[2] = 'c';
  arr[3] = 'd';
  return 0;
}
EOF
$ gcc -DCONST -O2 main.c -o const && objdump -d const > const.objdump
$ gcc -DMACRO -O2 main.c -o macro && objdump -d macro > macro.objdump
$ diff const.objdump macro.objdump
2c2
< const:     file format elf64-x86-64
---
> macro:     file format elf64-x86-64
```

barter-simsum: Yep, confirming that these arrays only show up as a VLA on unoptimized (or -O1 -- didn't test) builds. It's my understanding that gcc/clang tends to make VLA initialization an `alloca` call when it can. But this doesn't even matter much in these use cases as they're functionally identical.

Being a VLA would be a bigger deal if these were globals, because we'd be unnecessarily wasting stack space instead of storing it in bss.

joemfb: All that code is synchronous -- there probably should just be a single, global "page-sized scratch buffer".

barter-simsum:
Retiring this PR for a new one on urbit/vere

PR #6079, opened 2022-11-14
u3: fix memory corruption after meld; enforce guard page invariants

### LINKED ISSUE #6078: Running meld on a running ship can lead to memory corruption
The previous memory corruption issue (#6069) had to do with recentering the guard page when you run out of memory. When this was designed we didn't account for padding between inner roads which means we can end up rarely in a situation where there is no guard page present.

This new one is when you meld you're going to overwrite all persistent state in a loom. As part of this page tracking gets turned off, but since 1.9 this includes the guard page. After a meld until the ship is restarted there is no guard page, this means until we have a hotfix for this, melding on running ship can cause a ship to get corrupted (while continuing to run without a restart after the meld completes).

@joemfb has a release candidate out for 1.12 with this fix and may also include assertions alongside this.

joemfb:
This PR fixes another latent memory-corruption bug introduced in v1.10 (see #6072 for context, the bug was introduced in #5881 and #5882).

`|meld` modifies every page of the persistent state, so tracking the pages that it touches is not worthwhile. It's optimized to turn off page-tracking, in such a way that a subsequent snapshot update will restore it. But this optimization has the unfortunate side effect of removing the guard page (which it pre-dates), for the duration of the process (ie, restarting the ship would restore it, and running `meld` as a subcommand does not risk corruption).

This PR restores the guard page after turning off (all other) page tracking, and maintains or asserts other relevant invariants of the guard page system.

Fixes #6078.

### DIFF
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -568,6 +568,9 @@ _ce_patch_compose(void)
 
     nor_w = (nwr_w + (pag_wiz_i - 1)) >> u3a_page;
     sou_w = (swu_w + (pag_wiz_i - 1)) >> u3a_page;
+
+    c3_assert(  ((gar_pag_p >> u3a_page) >= nor_w)
+             && ((gar_pag_p >> u3a_page) <= (u3P.pag_w - (sou_w + 1))) );
   }
 
 #ifdef U3_SNAPSHOT_VALIDATION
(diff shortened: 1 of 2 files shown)


joemfb:
This first three commits in this PR will be in a hotfix release ASAP (the fourth only pertains to demand paging).

PR #6080, opened 2022-11-14
vere: handle errors during replay

joemfb:
This PR modifies the replay logic (in mars, unreleased) to explicitly enumerate errors during replay, print appropriate messages and/or stack traces, and retry `bail:meme` after `pack` (or `meld`, if run with `--auto-meld`). Targets #6079, as it was in the development of this feature that those issues were discovered.

### DIFF
--- a/pkg/urbit/daemon/main.c
+++ b/pkg/urbit/daemon/main.c
@@ -1300,9 +1300,7 @@ _cw_meld(c3_i argc, c3_c* argv[])
   u3C.wag_w |= u3o_hashless;
   u3m_boot(u3_Host.dir_c, u3a_bytes);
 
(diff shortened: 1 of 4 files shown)


zalberico:
For people finding this because they're trying to run replay and it's crashing because it runs out of memory, you can batch it with `-n`, but that requires manually running it and updating the event log number you want to run to. This simple bash script will automatically do 1M events for you until complete.

1. Save as replay.sh
2. chmod +x replay.sh
3. ./replay.sh
```
#!/bin/bash
num=1000000
while :
do
    ./urbit -n $num ./dovsem-bornyl
    num=$((num+1000000))
done
```

PR #6084, opened 2022-11-15
http-api: use /~/name endpoint if cookie not legible

Fang-:
Subset of #5918 that includes _just_ the `Urbit.ts` changes. This should be able to go out to frontends right now and retain their existing behavior, while preparing them for the full release of #5918 at some point in the future.


arthyn (approved):
lgtm

PR #6085, opened 2022-11-18
clay: document invariants

philipcmonk:
This is an experiment in documenting invariants in Clay.  This was inspired by synchronously reviews of recent changes to Clay, where the most important parts were me constructing arguments that certain guarantees were upheld.

This comment describes the system:

https://github.com/urbit/urbit/blob/45b5257c04f2b3a7d44c50f68be758b5751cb5ea/pkg/arvo/sys/vane/clay.hoon#L34-L77

This also fixes two issues discovered by thinking through these issues in this methodical way:
- `|label` did not result in a `+wake` call, so if a subscriber was waiting on that label to exist, they would not get updated until something else happened on that desk.
- `+park` did not invoke `+tare` for suspended desks, so if a commit to a suspended desk added an item to its commit-in-waiting, then subscribers would not hear about it until something else happened to invoke `+tare`.

This is an experiment because hitherto we have largely not documented these sorts of relationships, even though they're a common source of bugs.  I'm interested in examples of how other projects document this sort of thing.  I'm also using the term "invariant" in a fairly imprecise way -- some of these could be described as true invariants, but I'm not sure that fully encompasses the things that should be documented this way.  There's probably a better term.

In the meantime, I feel comfortable adding these notes to Clay.  Then we'll see over time how useful they are and whether they should be added to other areas of the kernel.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -63,6 +63,21 @@
 ::  close: [wake] >
 ::  open and almost immediately close: [wake] <>
 ::
+::  This system is best used for nonlocal invariants and is not
+::  necessary when a function can guarantee its own invariants.  For
+::  example, consider a set alongside a @ud representing its size.
+::  There is an invariant that any time you add or remove an item from
+::  the set you must update its size.  If you're operating on these
+::  directly, it could be beneficial to tag each line of code which
+::  might modify the set and make it clear where the size is modified.
+::
+::  Sometimes code can be restructured so that many fewer tags are
+::  needed.  In the above example, if the set is modified in many
+::  places, it may be worth factoring out set+size into a data structure
+::  with its own arms for put, del, uni, int, etc.  Then the invariant
+::  only needs to be maintained within that data structure, and call
+::  sites do not need to be tagged.
+::
 ::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
 ::
 ::  Here are the structures.  `++raft` is the formal arvo state.  It's also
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
This adds a new dimension to comments before that I hadn't considered before, very cool idea.

I'd like to point out that at least some of these comments will attach as doccords to the expressions, as it is written (such as all the end of line/post-fix comments `::  [wake]  <`). This opens up the possibility to use doccords to verify that invariants are maintained automatically rather than verifying it by hand. This wouldn't go in #5873 but might be a good future project.

I initially hesitated mentioning this because verification implemented via a comment system felt like it violates the invariant that comments don't change the behavior of code, at least in spirit if not in reality. But I think this is actually fine, since these are just arbitrary tags that are only accurate if the programmer made no mistakes, and wouldn't e.g. prevent it from compiling. It would just be data generated from comments that you could interrogate the type of `clay.hoon` for. Doccords also have a provision for including lines of code, which doesn't do anything yet, but could be used for e.g. unit tests in the future. So these invariant comments would fit into the already existing camp of "doccords as a testing tool".

Fang-:
This is a neat system! I mentioned to @belisarius222 the other day how docs that explain the vanes _only_ in terms of the invariants they aim to provide would be quite helpful. This is slightly more internals-facing than what I was imagining, but still very useful along those same lines.

There's still characters unused for doccords prefix notation, so we could do things like `!wake` in place of `[wake]` if we wanted. The `<` and `>` are also not great aesthetically imo. Maybe you want `wake!` and `wake.` or something, but `!wake!` would just look silly.

> I initially hesitated mentioning this because verification implemented via a comment system felt like it violates the invariant that comments don't change the behavior of code, at least in spirit if not in reality. But I think this is actually fine

For what it's worth, "what if we wrote tests _in_ the doccords" certainly popped into my mind at a couple points in time. If you do this too extensively/verbosely it probably ruins the readability of the whole, but for things you want to be documenting anyway...

drbeefsupreme:
`[foo]` would be fine as a prefix, and might actually be preferable to distinguish it from prefixes intended for batch comments, since you might want to write an invariant doc as part of a batch comment (like `+foo: [wake] a foo`).

What I have in mind is that this invariant documenting system could be parsed as a new kind of `$note` besides a `%help` `note` (which is what all doccords currently parse as). Anyways probably discussion of doccords internals is off-topic in this PR, but we should keep it in mind for bikeshedding the syntax.

philipcmonk:
We'll do it this way for now and can tweak it as we go along

PR #6092, opened 2022-11-21
zuse 417: "Agents in Clay"

### LINKED ISSUE #5897: Upgrade System Fix
The last kelvin change OTA led to multiple issues: either apps would fail to revive after upgrade or fail to suspend/upgrade.

This led to issues where users would get stuck on an old version with confusing notifications about suspended apps or apps blocking upgrades.

The fix required manually suspending apps, installing the OTA, and then reviving apps via the dojo - obviously not good for zero click maintenance.

The locus of improvements required to fix this exist in clay, gall, and kiln. Ted and Phililp are writing the fix for this on philip/agent-clay. We need this fix in prior to the next kelvin change release (remote scry OTA).

https://gist.github.com/belisarius222/2ae74bfb5a40860b59d28970d29b3329

philipcmonk:
This is the planned next arvo release.  It incorporates a bunch of PRs, mentioned in the release notes here: 
https://github.com/urbit/urbit/releases/tag/urbit-os-v2.129-rc1

Additionally, it includes an internal rewrite of the way that desks and agents are marked as running or not, to fix #5897.  This centralizes the logic in Clay, which makes kiln much simpler and allows Gall to remove its larval stage.

As a kelvin upgrade, this includes bumps to the sys.kelvin on each desk.  It also includes significant changes to garden/grid to accommodate the breaking changes.


pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4199,6 +4472,243 @@
...
+    ::
+    =.  sat  (apply-precedence sat)
+    =+  ?:  (lth veb.bug 1)  ~
+        %.  ~  %-  slog
+        %+  turn  sat
+        |=  [=desk =bill]
+        leaf+"goad: output: {<desk>}: {<bill>}"
+    =^  agents  ..abet  (build-agents sat)
+    =.  ..abet
+      (build-marks (turn (skip sat |=([desk =bill] =(bill ~))) head))
```
philipcmonk, line 4553: Screenshot from 2022-11-21 12-02-14

ohAitch: It is certainly at any rate a reasonable request to know *whether* all your mark files compile, to not have any files of marks that don't, to eagerly abort any computation touching non-file values of marks that don't compile, etc

PR #6093, opened 2022-11-22
grid: distinguish app sync from system install

tomholford:
# Context

By using `%kiln-install` instead of `%kiln-sync` for System Updates, this ensures that the `%kids` desk is also updated.

Also, address UX feedback: render the entire source ship's patp to avoid ambiguity. (as opposed to truncating a moon's name).

# Preview

## Render Entire Source's Patp

image

### DIFF
--- a/pkg/grid/src/state/kiln.ts
+++ b/pkg/grid/src/state/kiln.ts
@@ -1,4 +1,13 @@
-import { scryLag, getPikes, Pikes, Pike, kilnUnsync, kilnSync } from '@urbit/api';
+import {
+  scryLag,
+  getPikes,
+  Pikes,
+  Pike,
+  kilnUnsync,
+  kilnSync,
+  kilnUninstall,
+  kilnInstall
+} from '@urbit/api';
 import create from 'zustand';
 import produce from 'immer';
 import { useCallback } from 'react';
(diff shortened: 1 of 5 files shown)


arthyn (approved):
Not sure I entirely understand the semantics of install vs sync, but the moves here look good

philipcmonk (changes requested):
Thanks for the quick turnaround.  This looks good except the one note.

pkg/grid/src/state/kiln.ts
```diff
@@ -33,6 +43,13 @@ const useKilnState = create<KilnState>((set, get) => ({
     const lag = await api.scry<boolean>(scryLag);
     set({ lag });
   },
+  toggleInstall: async (desk: string, ship: string) => {
+    const synced = !!get().pikes[desk].sync;
+    await (useMockData
+      ? fakeRequest('')
+      : api.poke(synced ? kilnUninstall(desk) : kilnInstall(ship, desk)));
```
philipcmonk, line 50: (Whoops, comment didn't save)

For the System Updates case, this needs to install the foreign %kids desk to the local %base desk.  Looks like `kilnInstall` supports this with the optional `local?` argument.

https://github.com/urbit/urbit/blob/1bd0f4be77fcdd6c69799ac78429b08e3552b9d0/pkg/npm/api/hood/lib.ts#L12-L16

tomholford: Good catch, updated the `kilnInstall` usage to match the previous implementation:

https://github.com/urbit/urbit/blob/4834ec59a2a7d73969aa831a92d2c8ad232bd5e4/pkg/grid/src/state/kiln.ts#L59

philipcmonk (approved):
This works!

PR #6094, opened 2022-11-22
jael: send updates to all "general" pubkey subs

Fang-:
As discussed with @philipcmonk out of band, it's strange that the "general" `%public-keys` subscription excludes moons. Jael may not track all moons in existence, but it should certainly still tell you about the ones it knows about. In most cases, those are the only ones you care about anyway.


philipcmonk (approved):
Agree

Issue #6065, opened 2022-11-07
ames: too many open flows


yosoyubik:
There's been a couple of instances of this. One example from my ship:

```
[ship=~wolref-podlex open=1.125 closing=25 in=12 nax=3 corked=17.774]
[ship=~winder-dapsym open=11.953 closing=74 in=3 nax=1 corked=20.971]
[ship=~hiddev-dannut open=1.105 closing=10 in=3 nax=1 corked=3.770]
```
(there's been an increase in ~200 corked flows with ~winder-dapsym in two days)

and from another one:

```
~[
  [ship=~winder-dapsym open=[out-open=9.375 out-closing=7 in=1 nax=0] corked=31.702]
  [ship=~hiddev-dannut open=[out-open=852 out-closing=0 in=1 nax=0] corked=5.248]
  [ship=~wolref-podlex open=[out-open=385 out-closing=10 in=1 nax=0] corked=138.686]
]
```

These ships host big groups (>1k members: Urchan, Hooniverse, Foundation), so it might point to resubscription flows that were not corked properly(?). All these open flows will cause trouble for the hosts ships—there's been reports of [1.2GB in ames](https://habsul.uqbar.lol/the-habsulute/ship-problem)

ryjm:
<img width="720" alt="image" src="https://user-images.githubusercontent.com/2065299/200426091-a24e7956-cd54-44d3-bbc1-70b5585c8935.png">
woah

ryjm:
165mb in ames for me.

zalberico:
Suggestion from @belisarius222 - check to see if these are up to date first. If they are up to date, then this requires more debugging.

belisarius222:
> One solution would be to check in %gall if we get an error handling %watch remote responses, and then pass a %cork move to ames

This sounds right to me.  Good catch.

yosoyubik:
Another cause for the extra flows are re-subscriptions that were not properly %corked. From what I can see this is something that's not a problem anymore—the number of +ames-flows I can see on my ship is not increasing as fast as previously. For example, for a subscription to a resource on my ship, there are 33 flows open, going from sub-nonces 0 (i.e. pre GRQF subscription) to ~16.000, and then jumping to ~46.000 for the latest flow.

This can be addressed by going over all the subscriptions we have and sending a %cork for the ones less than the latest sub-nonce (i.e. the current subscription)—this solution is currently WIP here: https://github.com/urbit/urbit/pull/6129

[...]

PR #6102, opened 2022-11-28
ames, gall: send a %cork if getting a %nack for a %watch

yosoyubik:
Fixes https://github.com/urbit/urbit/issues/6065 by corkingn %nack flows—next would be to address deleting the existing dead flows.


yosoyubik:
@belisarius222 I addressed your style feedback, so this should be good to go into next/arvo.

belisarius222 (approved):
lgtm

Issue #6073, opened 2022-11-10
Kelvin Version Backwards Compatibility


zalberico:

zalberico:
@pkova and @tinnus-napbus is there a WIP PR for this that I can add her?

tinnus-napbus:
not at this stage. We're currently working on the gall and clay sides separately. I have a branch for the clay side: https://github.com/urbit/urbit/tree/tinnus/clay-kelvin-shims

we should have something more organised soon

zalberico:
That's fine to work on them separately - the sooner we can make this visible the better (a draft PR being pushed to is fine). If you're both able to also push comms here that'd be helpful.

Basically write up details/updates here as you hack and we can tag people when necessary to help you get unstuck (preferably at least weekly). It's also healthy to just have some of this communication and progress more accessible/visible on the issues themselves.

belisarius222:
@zalberico I agree

tinnus-napbus:
status update:

basic logic in clay is (theoretically) done & it compiles. Still needs some comments and error messages and possibly a couple of extra bits and pieces but it should basically work. Pyry is just about done on getting gall working so will test actually running a kelvin upgrade with it in the next couple of days

PR #6104, opened 2022-11-28
kelvin shims

tinnus-napbus:
This PR introduces "kelvin shims" which provide backwards-compatibility with old kelvin versions for desks.

- clay allows desks with any old weft in theirs sys.kelvin to be committed.
- clay searches backwards through the %base desk's aeons for a matching zuse, builds it, adds it to a zuse cache and returns it. Everything for that desk is then built against that zuse
- the ford cache keys now include the weft they're built against
- in clay's +lu, agents will be built and extracted to an `agent-any:gall`, which is a tagged union of weft and versioned `agent:gall` like `agent-416:gall` that gall is compatible with. This is provided in the new, modified `%load` task to gall.
- only desks that are gall-compatible can be set `%live`
- the zuse cache is garbage collected at the end of every park so it only contains zuse vases that are in use by one or more desks.
- the initial release of this is intended to be only current-kelvin compatible, there aren't actually shims at this stage
- no changes have been made to spider - since there won't be shims there won't be a problem there unless you try run threads off an old outdated %dead or %held desk, which is highly unusual so probably doesn't matter. Spider shims will have to be implemented when the first proper shim is added in a subsequent kelvin-breaking kernel update.

[see this gist for additional commentary on the changes to clay](https://gist.github.com/tinnus-napbus/5549fbf38d72a8ae482425c9694fa268)

XX @pkova to add further commentary about changes to gall

Hello it's me @pkova

- gall changes to acommodate `agent-any` inside a `yoke`, important to note that gall will not accept anything except the current `[%zuse %416]`
- new types `userspace-note:agent:gall`, `userspace-gift:agent:gall`, `sign-userspace:agent:gall` and `userspace-move` to represent the subset of moves that could ever be used from userspace
- new types `gall-note`, `gall-gift` and `gall-move` to represent an intermediate type which contains moves used by gall that should not be used from userspace (`%huck`, `%cork`)

this is related to #6073


pkova:
This PR is now in a state where we are ready for outside comments.

The aim of kelvin shims is to provide a measure of backwards compatibility for Gall applications. This initial version of kelvin shims contains code in Clay that builds and caches old versions of Zuse, and code in Gall that allows it to run agents of multiple versions in the future. Note that this PR contains the type changes in Gall necessary to write shims in the future, no actual shims are yet included.

pkg/arvo/sys/lull.hoon
```diff
@@ -2472,6 +2478,92 @@
       [%$ %whiz ~]
       [@tas %meta vase]
   ==
++$  note-gall                                           ::  userspace note
```
pkova, line 2481: `note-gall` is the new type for userspace notes. I'm very interested in thoughts on this list, since I'm definitely not familiar with all the notes included here. I constructed this list by starting with the work @Fang- has been doing on userspace permissions, then adding notes until the `%base` desk compiled. Many of these are only used in `/app/helm` and `/app/herm`.

Note that I tried to make this list with bucgar, but always ran into `mint-vain`, even when including a bucsig.

pkova:
Also, `next/arvo` seems to be broken at the moment. A fakezod boots up fine but the dojo does not work. Let me know if we should base this on some other branch.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -699,11 +702,12 @@
...
   ::
   ++  mo-load
-    |=  agents=(list [=dude =beak =agent])
+    |=  agents=(list [=dude =beak =agent-any])
     =.  mo-core
       |-  ^+  mo-core
       ?~  agents  mo-core
-      =/  [=dude =desk]  [dude q.beak]:i.agents
+      =/  [=dude =desk =agent-any]  [dude q.beak agent-any]:i.agents
+      ?>  ?=([%zuse %416] -.agent-any)
```
pkova, line 710: Currently just asserting this everywhere, the `agent-any` type allows us to build shims in the future.

pkova:
@belisarius222 @philipcmonk

pkg/arvo/sys/lull.hoon
```diff
@@ -2472,6 +2478,92 @@
...
+  $%  $:  %a
+    $%  [%sift ships=(list ship)]
+        [%spew veb=(list verb:ames)]  :: set verbosity (we want this?)
+        [%cork =ship]                 :: never userspace
```
belisarius222, line 2485: I'm confused: if they're "never userspace", then why are they in the type?

pkg/arvo/sys/lull.hoon
belisarius222: You'll need to do the same for $gift:agent:gall, so that the agent type won't depend (contravariantly) on the full Arvo gift type.

I'm not sure why you'd get a mint-vain for this, but let's figure that out, because it should work.  One guess as to why it's not working would be that there's a mutual recursion in the types somewhere, where they're defined in terms
of each other, and maybe that's causing a problem -- I wouldn't expect a mint-vain from that, though.

pkg/arvo/sys/lull.hoon
pkova: I made the type `note-gall` for both notes the vane proper emits and notes emitted by userspace. This confusing comment was meant to indicate the former.

Now it's the more reasonable thing, which is that moves in Gall proper (under the `++mo` core) remain `move`, and userspace moves (under the `++ap` core) are `userspace-move`.

pkg/arvo/sys/lull.hoon
pkova: Figured out the bucgar problem and pushed a fix, I had missed some changes to the types of various tasks that I pulled in when merging to `next/arvo`. Doing the gifts next.

belisarius222: 👍

pkova: Gifts are now done.

pkg/arvo/sys/lull.hoon
```diff
@@ -1813,17 +1814,130 @@
...
+          $>(%private-keys gift:jael)
+          $>(%public-keys gift:jael)
+          $>(%turf gift:jael)
+          ::
+          $>(%arow gift:khan)
+      ==
+    +$  userspace-note
+      $%  $:  %a
+        $%  $>(%sift task:ames)
+            $>(%spew task:ames) :: set verbosity (we want this?)
```
belisarius222, line 1862: yes, we want this

pkg/arvo/sys/lull.hoon
```diff
@@ -1813,17 +1814,130 @@
...
+            $>(%cork task:ames) :: gall does this
```
belisarius222, line 1863: I'm not sure whether userspace should ever be allowed to %cork.  I can see the use case for recovery commands, although I'd be inclined to expose those commands as ames tasks rather than letting userspace cork willy-nilly.

pkg/arvo/sys/lull.hoon
```diff
@@ -1813,17 +1814,130 @@
...
+            $>(%stir task:ames) :: ?? helm does this
+            $>(%prod task:ames) :: ?? helm does this
+        ==  ==
+          $:  %b
+        $%  $>(%rest task:behn)
+            $>(%wait task:behn)
+            $>(%huck task:behn) :: gall does htis
```
belisarius222, line 1870: just because gall can %huck doesn't mean userspace should be able to -- since it's manipulating the duct call stack, I don't think userspace should be allowed to do it

pkova:
Changes since last time:

- I had `gift-userspace`, but some stuff (notably `/lib/verb`) needed the vane prefix from `sign-arvo`, so I made `sign-userspace`.
- Migrated the `%base` desk to use `sign-userspace` instead of `sign-arvo`.

pkg/arvo/sys/lull.hoon
pkova: Gall passes a `%huck` from `++ap-kill-down`, eventually ending up in `+ap-move`, where the moves end up in `agent-moves` inside the `+ap` core which is now of type `userspace-move`. I would like the `%huck` moves to end up straight in `moves` under the `++mo` core, unsure of how to achieve this though.

https://github.com/urbit/urbit/blob/31e7a3acba69577ec96cd3eb15ca5f89afa41c5c/pkg/arvo/sys/vane/gall.hoon#L1508-L1515

https://github.com/urbit/urbit/blob/31e7a3acba69577ec96cd3eb15ca5f89afa41c5c/pkg/arvo/sys/vane/gall.hoon#L1201-L1203 

https://github.com/urbit/urbit/blob/31e7a3acba69577ec96cd3eb15ca5f89afa41c5c/pkg/arvo/sys/vane/gall.hoon#L886

https://github.com/urbit/urbit/blob/31e7a3acba69577ec96cd3eb15ca5f89afa41c5c/pkg/arvo/sys/vane/gall.hoon#L194

pkova: As discussed out of band, I made agent-moves an intermediate type `(list gall-move)`, which includes some gifts and notes that are used by the Gall vane but are disallowed for userspace. 
https://github.com/urbit/urbit/blob/f9b1af32f47782bea67394490d612bc350732e43/pkg/arvo/sys/vane/gall.hoon#L48-L68

pkova:
All outstanding issues on the Gall side are now resolved.

tinnus-napbus:
@pkova @belisarius222 the sign-userpace changes break aqua

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -42,9 +42,33 @@
 ::  $move: Arvo-level move
 ::
 +$  move  [=duct move=(wind note-arvo gift-arvo)]
+::
++$  userspace-move  [=duct move=(wind userspace-note:agent userspace-gift:agent)]
```
belisarius222, line 46: what do you think of renaming all the `*-userspace` types to `*-user` instead?

belisarius222: (and `userspace-*` to `user-*`)

pkg/urbit/jets/d/in_apt.c
```diff
@@ -13,21 +13,15 @@ _in_apt(u3_noun a, u3_weak l, u3_weak r)
     u3_noun n_a, l_a, r_a;
     u3x_trel(a, &n_a, &l_a, &r_a);
 
-    if ( (u3_none != l) &&
```
belisarius222, line 16: I'm guessing the changes to this file (in_apt.c) are just git being stale?

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -2012,10 +2090,11 @@
   %=    yoke
       code   ~
       agent
-    :-  %|
     ?:  ?=(%| -.agent.yoke)
-      p.agent.yoke
-    on-save:p.agent.yoke
+      [%| p.agent.yoke]
+    ?-  -.+.agent.yoke
```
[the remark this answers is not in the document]
pkova: Roger, fixed

pkg/arvo/sys/vane/gall.hoon
pkova: Sounds good to me, done

pkova:
Superseded by #6104.

PR #6105, opened 2022-11-28
Fix comment: `+of` in lieu of `|de`

rabsef-bicrym:
I'm not sure if `|de` has some connotation - it's mentioned that this core would be called `de` in prior commits but it doesn't seem to be named that here. Until it switches to `de`, putting it in parity with `+of`


joemfb:
@rabsef-bicrym for future reference, you can always just rebase (or cherry-pick) your commit onto (a local copy of) the target branch (`next/arvo` in this case), then force-push your branch to update the pr. you don't have to open a new one.

also, what you did should've been fine -- `next/arvo` is lagging `master`, but it's not supposed to.

PR #6106, opened 2022-11-28
Fix Comment: `arvo` comment for `|of` reads `|de`

rabsef-bicrym:
I'm not sure if `|de` has some connotation - it's mentioned that this core would be called `de` in prior commits but it doesn't seem to be named that here. Until it switches to `de`, putting it in parity with `+of`

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -415,7 +415,7 @@
     ==
   --
 ::
-::  |de: axal engine
+::  |of: axal engine
 ::
 ++  of
   =|  fat=(axal)


joemfb:
`|de` was refactored in #5694, and then renamed to `|of` in #5408. Thanks for catching the comment.

The CI failures here are a misconfiguration on PRs from forks that I introduced in urbit-v1.11. Not sure what exactly is wrong, I'll make a note to look into that.

joemfb (approved):
this will do

Fang-:
Going forward, please heed the commit format as described in [the hygiene section of the contributing.md](https://github.com/urbit/urbit/blob/master/CONTRIBUTING.md#hygiene). In this case, `lull: fix comment for +of` or similar would be appropriate. Thank you! (:

Fang- (approved):
Still good, thanks for the commit hygiene compliance!

PR #6108, opened 2022-11-28
u3: add comment explaining willoc logic

philipcmonk:
I ran into this and paired with @joemfb on it for a while today.  It appears to work when you remove this line, but it's very slow -- this is important information that we should document in the code.


joemfb (approved):
this will do

zalberico:
Looks like some jobs are failing - short of that I think this can merge? Is there a known reason those jobs are failing?

PR #6110, opened 2022-11-30
Add var loom to docker

yapishu:
This is the modified `start-urbit` script we make use of for NativePlanet's [GroundSeg](https://github.com/Native-Planet/GroundSeg/blob/variable_loom/api/urbit_docker.py); it allows the following additional variables to be set via env var:

`httpPort` -- HTTP server port (default 80)
`loom` -- Loom size (default 31 [2GB])

Making this PR in response to a request by Kenny on the [urbit-dev](https://groups.google.com/a/urbit.org/g/dev/c/Fdoikafh0e0) mailing list


joemfb (commented):
@yapishu thanks for taking care of this!

There are some unnecessary whitespace changes in this diff: blank lines deleted including the conventional newline at the end of the file. Can you revert those?

Also, how well does setting the http port to anything other than 80 work? It look like the image is hardcoded to only expose port 80 over tcp.

yapishu:
@joemfb ah, that didn't occur to me -- as I mentioned this is for the way we use this image, which involves attaching it to a Wireguard docker network rather than forwarding the port to the host's interface. You could leave it there (it works, inside the container) but the typical use case can just set the forwarded port on the host instead.

yapishu:
To clarify, setting the port inside the container is only a necessity if you have multiple ship containers sharing a single docker network, otherwise they will conflict

zalberico:
These images are now managed by the https://github.com/urbit/vere repo

PR #6112, opened 2022-11-30
Revert "removing old Nock specs"

joemfb:
This reverts commit ab7a1bf5647931e93d76a4cfcbb81bff95301d29.

The history of these specs is important and should be preserved. These old docs should never have been deleted.


zalberico:
Changed the base branch to next/arvo - nothing merges to master outside of a release.

@philipcmonk fwiw - the change that did to github is one area I'm confused by merge. Rebase in this case would just cleanly put his nock docs on top of the next/arvo branch (as opposed to adding all the extra crap now seen here and in the diff). Though once we lock master and default to next/arvo this shouldn't be that common?

joemfb:
I targeted master because this is not a shipping change, it's like updating README.md or CONTRIBUTING.md. The reason changing the base branch caused all this stuff to show up is because master has not been merged back into next/arvo.

philipcmonk:
Yes, next/arvo was behind master because of the kernel release (which is an issue that will be resolved as the monorepo splits apart).  Whenever that happens, we should merge master into next/arvo (thanks Mark).  Github keeps the diff cached until you change something (eg make another commit), an easy way is to change the base branch away and then back.

zalberico:
That makes sense - thanks!

zalberico (approved):

PR #6115, opened 2022-11-30
groups: removing notify

arthyn:
This is currently under contention for agent name with new groups, and unused in current groups.


philipcmonk (approved):
This makes sense

Issue #6113, opened 2022-11-30
"too many redirects" after 417 OTA


tomholford:
Yesterday I updated a self-hosted moon and planet to the new kelvin version (417).

Both are running vere 1.13, have updates for `garden`, `landscape`, `webterm`, and `bitcoin` set to the pre-release ship.

The moon update went off without a hitch. Same seemingly for the planet, the dojo output seems nominal. However, receiving a "Too many redirects" error message in the browser when attempting to connect to Grid / Landscape.

Based on the troubleshooting conducted so far, this seems to be a cert issue. Will update this ticket once I have more information.

cc: @zalberico

tomholford:
Update:

In the [release notes](https://github.com/urbit/urbit/releases/tag/urbit-os-v2.129-rc4), there's a line about https redirects:
>- Add automatic redirect to https when available.

Wondering if this might be the root cause of the redirect loop. Looking at the Caddy server logs, seeing ~40 "status: [301](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301)" messages, so a loop potentially between the Urbit binary and Caddy is being triggered.

For context, I am using a docker-compose file like so:
```yml
version: "3.7"

networks:
  web:
    external: true

services:
  caddy:
    image: caddy:2-alpine
    restart: unless-stopped
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - $PWD/Caddyfile:/etc/caddy/Caddyfile
      - caddy_data:/data
      - caddy_config:/config
      - ../globs:/globs
    networks:
      - web
  datder-sonnet:
    image: tloncorp/urbit:v1.13
    restart: unless-stopped
    ports:
      - "8080:80"
      - "34343:34343/udp"
    volumes:
      - ~/urbit/datder-sonnet:/urbit/datder-sonnet
    networks:
      - web
    stdin_open: true
    tty: true
    depends_on:
      - caddy

volumes:
  caddy_data:
    external: true
  caddy_config:
```

And a Caddyfile like so:
```Caddyfile
sonnet.page {
	reverse_proxy ops-datder-sonnet-1:80
}
```

tomholford:
Update:

Looks like this could potentially be related to #5987. Continuing to investigate.

tomholford:
[...]

PR #6116, opened 2022-12-01
Fix host scry

tinnus-napbus:
eyre's %host scry logic only says it's secure if a domain is configured and is not localhost. It's possible to have a certificate without a configured domain so in that case its results are incorrect. I have just removed the domain check


tinnus-napbus:
this arose when troubleshooting #6113 cos it looked like it was insecure but actually it was secure so it was confusing

zalberico:
@Fang- thanks for updating the base branch, when you get a chance to CR if things look good and tests are passing we can approve/merge.

Fang-:
This seems fine, but CI here is hitting what looks like the same thing as #6139. Unclear why that happens, since it doesn't happen on master or next/arvo, and next/arvo has just been merged into this branch...

jalehman:
@tinnus-napbus any concerns with @Fang- 's change?

tinnus-napbus:
whoops, yeah that looks good

Fang- (approved):
Alright, all good then. (:

PR #6117, opened 2022-12-01
hoon: add +lure, for turning a singleton into a list

liam-fitzgerald:
Highly convenient in many point free scenarios. `unit` has `+some` so `list` should have an equivalent


zalberico:
@belisarius222 or @joemfb can you approve this small change if it looks good? We can then merge it into next/arvo so it goes out with the next branched release (after 417)

joemfb:
@liam-fitzgerald I'm not a fan of point-free hoon, but this addition seems harmless to me ;)

@zalberico since approval implies merge and merge implies release, I'm not prepared to approve this PR yet. Changing `hoon.hoon` will require rebuilding everything on all ships, and will add to the (currently unreclaimable) space leak in the cold jet state on each ship. Unless there are other, higher-value changes to hoon ready to ship, I would not release this.

zalberico:
> will add to the (currently unreclaimable) space leak in the cold jet state on each ship

I'm guessing this exists as an issue somewhere, but what is the thing that fixes this? Otherwise all `hoon.hoon` changes are blocked on this?

> Changing hoon.hoon will require rebuilding everything on all ships

The concern here being that this will cause ships to crash from running out of memory? What's necessary for me to be able to release `hoon.hoon` changes without them crashing ships? I think we need to be able to merge these kinds of small things in the weekly releases as soon as they're ready without having to worry about it rebuilding everything being a problem. If something is blocking our ability to do that, it bumps up in my priority to fix.

drbeefsupreme:
This is another candidate `hoon.hoon` change to go out with #5873

philipcmonk (approved):
hoon 139 is imminent (doccords), so this can go in

PR #6122, opened 2022-12-02
[WIP] ames: refactor using the +abet pattern

yosoyubik:

- move the formal interface arms of %ames (+call, +take... to the bottom of the file
- rename nested cores using the two-letter scheme (e.g. peer-core -> po)
- nest +make-sink-message (|sink) and +make-message-pump |pump) into +peer-core, and make-paclet-pump (|pack) into +message-pump, to get access to shared (e.g. channel) state in the peer-core.
- remove gifts, and external +run-message/packet-pump arms, and handled those within the core.
- In an +abet core, (normally) don't explicitly update anything in the outer core that's not part of the core's own (there are a couple of FIXMEs where we need to refactor this)
  - |sink and |pump actually do modify the state of the parent core implicitly by calling each other (e.g. +pump-done abets the |sink core, and thus modifying the peer-core)
- if nested, abetting a core returns the parent core
- if there are gifts, we +abut (`[gifts abet]`) the core—any other name suggestions welcome

TODOs

- shorten xx-xx-state types
- ...
- document invariants

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1159,12 +1159,9 @@
       ^+  event-core
       =/  sink-core  (mi:peer-core bone *message-sink-state)
       =.  peer-core  abet:(call:abed:sink-core %done ok=%.y)
-      ::  XX this shouldn't be call randomly,
-      ::  only after we sink the message,
-      ::  invariant to document
-      =?  peer-core  closing:sink-core  (handle-cork:peer-core bone)
+      ::  handle cork only deals with bones that are in closing
       ::
-      abet:peer-core
+      abet:(handle-cork:peer-core bone)
     ::  failed; send message nack packet
     ::
     ++  send-nack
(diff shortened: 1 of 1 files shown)


yosoyubik:
I'd suggest to start releasing this (prior livenet test on some of my moons, and then ~binnec) once the first of the next-step tasks (remove usage of impure +abet pattern; see https://github.com/urbit/urbit/issues/6103#issuecomment-1346197840) is done, instead of waiting to get all of the tasks completed. This way we can verify that things work as we move along, since the remaining task in the list are not just refactoring but using different (e.g. finger trees) implementations.

jalehman:
That's fine. Let's move this out of draft state, clean it up, and get it into the review pipeline then.

belisarius222:
1. I don't think we need to remove the impure +abet in this first round of refactoring.  A later round can get it.
2. We should make sure to test corks and nacks thoroughly before pushing this out.

yosoyubik:
Instead of directly merging develop or cherrypicking commits, I've redone this in another branch, so closing this in favor of https://github.com/urbit/urbit/pull/6350

PR #6126, opened 2022-12-02
u3: document and enforce snapshot system invariants

joemfb:
This PR fixes an unreleased bug introduced in #6063, whereby snapshots could be corrupted when they were resized.

The bug was caused by a discrepancy between memory protections and page metadata (not unlike #6125), and was introduced while refactoring an earlier version that failed to maintain the invariant that only pages within the snapshot could be clean (harmlessly, because there was no discrepancy). So this PR also includes documentation of the key invariants of the snapshot system. Thanks to @belisarius222 for his assistance formulating and reviewing these invariants. Further localization of these comments (a la #6085) would be highly desirable.

Finally, this PR replaces `static const` page sizes with macros, to avoid the proliferation of VLAs introduced in #5881. Thanks to @barter-simsum for pointing this out in #6077.

### DIFF
--- a/pkg/urbit/noun/events.c
+++ b/pkg/urbit/noun/events.c
@@ -747,34 +747,6 @@ _ce_patch_apply(u3_ce_patch* pat_u)
   }
 }
 
-/* _ce_loom_pure_north(): track clean pages at the bottom of memory.
-*/
-static inline void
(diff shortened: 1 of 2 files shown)


joemfb:
This PR targets a restaging of #6063, to localize the diff. That target branch will be included in the next release.

PR #6129, opened 2022-12-05
ames: add |close-flows

yosoyubik:
Adds `|close-flows`, a %hood generator that sends a %cork to all ships that have nacked a %watch (e.g. we tried to subscribe, and the publisher crashed or rejected our subscription, but we didn't properly %cork the flow). This issue is fixed in https://github.com/urbit/urbit/pull/6102, and this PR depends on it in order to work properly.

The tool runs in dry mode by default, printing the number of flows that can be closed. To actually send the corks, run as:

`|close-flows, =dry |`

(possible fix for https://github.com/urbit/urbit/issues/6065, that will alleviate _some_ memory pressure in %ames)

PS: this updates %lull in its current form by adding a %kroc task to ames—the names is taken but we currently don't use the %kroc gift and the $krocs set so we should probably remove that.

### DIFF
--- a/pkg/arvo/gen/stale-flows.hoon
+++ b/pkg/arvo/gen/stale-flows.hoon
@@ -139,5 +139,3 @@
   ~?  =(%1 veb)  "[bone={<target>} nonce={<nonce>} agent={<app>}] {<path>}"
   (~(put in nacks) [ship target])
 --
-
-
--- a/pkg/arvo/gen/stale-flows.hoon
+++ b/pkg/arvo/gen/stale-flows.hoon
@@ -1,0 +1,145 @@
+::  +stale-flows: prints number of ames flows that can be closed
+::
+::    |stale-flows, =veb %0  :: TODO
+::    |stale-flows, =veb %1  :: flows from nacking initial subscriptions
+::    |stale-flows, =veb %2  :: stale flows that keep (re)trying to connect
+::    |stale-flows, =veb %21 :: ... per app (only forward)
+::    |stale-flows, =veb %3  :: stale resubscriptions
+::    |stale-flows, =veb %31 :: (TODO) ... per app
+::
+=>  |%
+    +$  subs  (jar path [ship bone @])
+    +$  pags  (jar app=term [dst=term =ship =path])  ::  per-agent
+    +$  naks  (set [ship bone])
+    ::  verbosity:
+    ::
+
+    ::
+    +$  veb  ?(%0 %1 %2 %21 %3 %31)
+    ::
+    ++  resubs
+      |=  [=subs =veb]
+      ^-  @
+      ::=/  sorted
+      ::  %+  sort  ~(tap by subs)
+      ::  |=([a=(list *) b=(list *)] (lte (lent a) (lent b)))
+      %+  roll  ~(tap by subs)::sorted
+      |=  [[k=path v=(list [ship bone @])] num=@]
+      ~?  &(=(%3 veb) (gth (lent v) 1))
+        "#{<(dec (lent v))>} stale resubs on {<k>}"
+      ?.  (gth (lent v) 1)  num
+      (add (lent v) num)
+    --
+::
+:-  %say
+|=  $:  [now=@da eny=@uvJ bec=beak]
+        [arg=~ dry=? =veb]
+    ==
+::
+=/  peers-map
+  .^((map ship ?(%alien %known)) %ax /(scot %p p.bec)//(scot %da now)/peers)
+::
+=/  peers=(list ship)
+  %+  murn  ~(tap by peers-map)
+  |=  [=ship val=?(%alien %known)]
+  ?:  =(ship p.bec)
+    ~  ::  this is weird, but we saw it
+  ?-  val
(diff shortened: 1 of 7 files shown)


yosoyubik:
I think that it'd be best to release this at first without the option to close flows (i.e. no %ames or %lull changes), so only printing diagnostics about what can be corked—probably updating what types of diagnostics (ship, apps...) are displayed.

yosoyubik:
Marking this as draft because I'm rethinking how to properly handle this. The generator now combines two types of flows that can be closed:

- nacking %watch pleas—this are stale flows that should have been %corked but were not (https://github.com/urbit/urbit/pull/6102 solves this)
- re-subscriptions that were not properly %corked—another cause of https://github.com/urbit/urbit/issues/6065

The second case can be refactor as a %gall task—similar to +ap-rake (is basically the same but extending it to all subscriptions less than the latest sub-nonce), but that will need to include a new %gall task, which means updating %lull—an update is needed anyway if we want to add am %ames task to %cork specific bones for failed subscriptions (that information is not available in %gall, only in %ames).

Right now this only prints the number of flows that can be closed, which seems an interesting thing on its own, possible as part of a more fully-feature opt-in diagnostic tool, that can inspect some part of state of a ship, possibly adding as well things like azimuth-block, status of /app/azimuth (is it subscribed to %eth-watcher?)... etc. One inmediate benefit of this would be to gather some data about the causes of https://github.com/urbit/urbit/issues/5748

yosoyubik:
@zalberico @belisarius222 thoughts on this?: https://github.com/urbit/urbit/pull/6129#issuecomment-1344457694

tl;dr: 
1. instead of shipping this directly into people's %base desk, we could have an intermediate step where we use it (and extend it) as a diagnostics tool that people can install
2. release this generator (with updates) as part of the %base desk and ask people to run it to gather data
3. add tasks to both %ames and %gall and have something trigger them, cleaning up extra flows.

No. 2 is probably the easiest, less risky option? It's also similar to what was done with %doff and %rake with the GRQF was deployed.

belisarius222:
I like option 2.  Anyone should be able to run this without needing to install extra software, and a generator is fine.  I wouldn't automate running this cleanup unless we find we need to for some reason.

—
Ted Blackman
CTO | Urbit Foundation
~rovnys-ricfer
https://twitter.com/rovnys

yosoyubik:
@belisarius222 I've refactored this to account for all subscriptions that %ames failed to cork and are cluttering its state.

Current state of the PR  (from the latest commit): 

"""
This removes the logic from cleaning up stale subscriptions in %gall,
leaving `+ap-rake` as it was, and moves it to the +on-kroc arm in %ames.

Failed subscriptions from nacking a %watch plea that were not properly 
corked (fixed in https://github.com/urbit/urbit/pull/6102) are a subset of the more general 
"stale re-subscription" issue, so we take care of all stale flows at the 
same time, by focusing on the current  subscription—leaving all other 
[stale flows] to be corked automatically—and checking if it received 
a nack, to subsequently cork it.
"""

I have deployed a small "diagnostics" desk (`|install ~dister-norsyr-torryn %diagnose`; run with `+diagnose!ames`) that uses a generator to inspect the state of ames and print some info about re-subscriptions, flows with too many retries, nacks on initial %watches... so we can gather some info about the state of people's ships—the generator can be incorporated in the %base desk if we think it's useful (maybe as an expansion of the current `+ames-flows`) but thought that making it available right away was best in order to evaluate the problem.

Note: we introduce a new %ames task (%kroc) in %lull, so this reloads everything.

jalehman:
```
> +diagnose!ames
#23.748 flows from %nacking %watches
#6.045 backward flows with >10 retries
#2.894 forward flows with >10 retries
#2.264 stale resubscriptions
```

pkg/arvo/gen/stale-flows.hoon
```diff
@@ -0,0 +1,147 @@
+::  +stale-flows: prints number of ames flows that can be closed
+::
+::    |stale-flows, =veb %0  :: TODO
```
belisarius222, line 3: please fill in this and the other TODO in this block comment

pkg/arvo/gen/stale-flows.hoon
```diff
@@ -0,0 +1,147 @@
...
+::    |stale-flows, =veb %21 :: ... per app (only forward)
+::    |stale-flows, =veb %3  :: stale resubscriptions
+::    |stale-flows, =veb %31 :: (TODO) ... per app
+::
+=>  |%
+    +$  subs  (jar path [ship bone @])
+    +$  pags  (jar app=term [dst=term =ship =path])  ::  per-agent
+    +$  naks  (set [ship bone])
+    ::  verbosity:
+    ::
```
belisarius222, line 15: blank line

belisarius222 (approved):

jalehman:
@yosoyubik has uncovered an issue with this and will be submitting a new PR that reverts it from `develop` ahead of tomorrow's release.

zalberico:
@yosoyubik what’s the issue with it? Can you write up the details? (reverting first is fine)

PR #6136, opened 2022-12-07
clay: add scry for subscribers of a desk

bacwyls:
`.^((set ship) %cs /=escape=/subs)` returns the total set of subscriber ships to the `escape` desk, regardless of the conditions / context of their subscription.

It's filtered out as a `(set ship)` rather than the full `cult` because lull doesn't currently have a `cult:clay` and there's not currently a clear need for the additional data.

I read in the docs that `%cs` is a 'miscellaneous' category and is expected to change over time. So I figured it's a good place to fill the immediate need to surface this data.

When pushing out an OTA it's good to know ahead of time who it will go to / how many ships it will go to. Plus this data is valuable to app developers who want to keep track of who their users are / how many users they have.


zalberico:
@bacwyls thanks for this - @belisarius222, assigning this to you for triage from the UF side.

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4245,6 +4245,21 @@
           %-  wrap:fusion
           (page-to-cage:(aeon-ford yon) u.peg)
         ``cage+[-:!>(*^cage) cage]
+      ::
+          %subs
+        =/  doj=(unit dojo)  (~(get by dos.rom) i.t.pax)
+        ?~  doj  ~
```
belisarius222, line 4251: I think this might be better as `[~ ~]`.  We know the answer -- there are no subscribers.  `~` is for when we don't know the answer.

joemfb:
I think this would be better as `.^((set ship) %cs /=escape=/subs)`, with the desk coming from the beak instead of deeper in the path.

bacwyls:
@joemfb A scry of that form gets caught on this line.
https://github.com/urbit/urbit/blob/ed2190dc196b618d27c066a9cb6ea5655cebbce2/pkg/arvo/sys/vane/clay.hoon#L4210

Either that line of code would need to change or some 'argument' is needed after the tag (e.g. `.^((set ship) %cs /=escape=/subs/foo)`). I'd rather keep a minimal scope for this PR, so it may be best to stick with the requirement.

In that case, the scry could either look like`.^((set ship) %cs /=escape=/subs/~)` or `.^((set ship) %cs %/subs/escape)`. I'm not sure which is preferable.

jalehman:
@bacwyls please resolve the conflict with clay. After you do that, @belisarius222 will run the code to verify that everything's good, at which point he'll approve and merge this.

belisarius222:
This looks good in most cases, but it should refuse to answer at any case other than the current date, to preserve referential transparency.

belisarius222:
I tested this on a fakezod by `.^`ing the following cases:
- real desk, non-null subscriber set (returns set of subscribed ships)
- real desk, null subscriber set (returns `~` without crashing)
- nonexistent desk, current date (returns `~` without crashing)
- real desk, numeric revision (crashes because not current date)
- nonexistent desk, numeric revision (crashes because not current date)

Since I was the original reviewer, I will merge this without a further review once CI finishes.

Issue #6111, opened 2022-11-30
Automatically deploy next/arvo to canary


zalberico:
As part of releasing regularly we need to have next/arvo automatically deploy to our internal canary moon on merge.

I think the correct way to do this is probably with Janeway.

Once we have this in place, every merge into next/arvo should get deployed to the canary. Kernel devs should then be using that canary as their OTA source and keeping an eye on how things are going after their PRs merge and they pick them up.

The specifics can be seen in this doc: https://docs.google.com/document/d/1CNc4QbICMsJDsk3eHUZeQQ5897tE3sQCqChVcOi_aaA/edit

zalberico:
@jtobin are you hitting any issues with this? We don't want to take down binnec prior to the Thursday groups 2 release (since they're using binnec to distribute the beta internally), but the janeway bits I think we should be able to have a good sense about (and what's currently on next/arvo should also be pretty safe).

jtobin:
No issues -- just clued up some travel, so this will probably be done tomorrow.

PR #6146, opened 2022-12-12
ci: add ota-on-merge flow for next/arvo

jtobin:
#6111 

Adds a GitHub action & workflow for making OTA updates to arvo on a remote ship.  The workflow is configured to deploy an OTA to the kernel canary moon (~binnec) on every push to next/arvo, and will go live as soon as this PR is merged.

(Also guts a bunch of deprecated workflows.)


jtobin:
Yeah, I tested the flow by initially having it run on pushes to this branch (instead of next/arvo), and when it did run it would only sync the files to %base (without committing them).  I then retargeted things to run on pushes to next/arvo and removed the `--no-commit` flag when everything seemed ship-shape.

It should be ok, and at minimum nothing should outright explode when it starts running.

zalberico (approved):

joemfb:
Should this deployment be gated on CI success?

zalberico:
Yeah, stuff shouldn't merge with broken CI.

jtobin:
There's no *formal* mechanism here checking for CI success, the working assumption being that stuff just doesn't get merged if it breaks CI.  But it's possible to add some stuff to execute conditionally on the success of another jobs.

Thoughts?

matildepark:
It might introduce a case where it invisibly doesn't deploy something people in the future expect it to because, well, say there's nebulous long-lasting false negative CI results. It would lead people to merge anyway, but then because of lack of institutional knowledge it doesn't go out and models of "what's on canary" diverge between people.

jtobin:
Yeah I'm inclined to agree that gating merges should be sufficient, but don't have terribly strong feelings either way.

zalberico:
Yeah, deploy should happen on merge no matter what. People shouldn't merge if CI is broken. If people merge when CI is broken and it kills the canary (or breaks CI on all figure commits blocking merge) the person that merged that is responsible for fixing.

zalberico:
@jtobin - I’ll send a company wide email today about this (just to make sure non-core devs don’t have binnec unexpectedly set for OTAs), but after that I think we’re good to merge. Once I send that, can you merge this and then keep an eye on it to make sure deploys work as expected?

jtobin:
Sure thing.

zalberico:
Email went out - you're good to merge when ready.

PR #6148, opened 2022-12-12
Master is broken: syntax error in app/azimuth.hoon

tadad:
couldn't build base on my fake ship

### DIFF
--- a/pkg/arvo/app/azimuth.hoon
+++ b/pkg/arvo/app/azimuth.hoon
@@ -4,7 +4,7 @@
     naive,
     dice,
     default-agent,
-/   verb,
+    verb,
     dbug
 ::  To update, run from dojo:
 ::    -azimuth-snap-state %default 'version-0'


belisarius222:
I find it disturbing that the master branch had a syntax error here in the azimuth code.  It should prevent anything from working, right?  I think this suggests a bunch of artifacts are stale, like pills -- or they fail to build and nobody's tried them.

Or maybe it's more that the master branch has diverged from a good state in a release branch, and the artifacts accurately reflect the good state from that other branch.

zalberico:
You can tell in the git blame, it's this: https://github.com/urbit/urbit/commit/0e16d82a46ea569e04326ba497b99f93b7ddc751

We should not be merging anything into master outside of a tested release.

zalberico:
@tadad can you amend your commit with a better commit message?

zalberico:
I'll resubmit with a new commit message - @tadad thanks for flagging this for us!

PR #6156, opened 2022-12-13
jael: also send %ruin breaches to catch-all subs

Fang-:
Previously, fake breaches triggered by a %ruin task would only get sent to subscribers watching for the affected ship specifically. Now, we send them to both those subscribers, and the ones watching for pubkey changes on all ships.


zalberico:
Nice - @philipcmonk if we can get approval on this we can merge and deploy it right to binnec to include in the first weekly release.

philipcmonk (approved):
This is good

PR #6159, opened 2022-12-14
vere: adds dedicated replay command

joemfb:
This PR forward-ports (side-ports?) the synchronous replay implementation from mars/urth (#6080) and anything it depends on, adding a dedicated `play` command (supporting `--loom`, `--no-demand`, `-n`/`--replay-to`, and `-f`/`--full` for full replay without editing the pier and moving/deleting the snapshot).

Ideally this could be automatically used, bypassing the legacy async/multi-process replay implementation. But there are thorny library de/re-initialization problems that will need to be solved first, and this is ready as is.


joemfb:
This now includes #6006 (but only in the dedicate replay command). Turning up all the knobs looks like: `$pier/.run play --auto-meld --loom 32`.

/cc @baudtack

baudtack:
sick, Is there an easy way i can build/get a binary with this already?

PR #6161, opened 2022-12-15
hoon: move +autoname:ax and +name:ap into parser

drbeefsupreme:
`+autoname:ax` and `+name:ap` are only used in the parser (utilized for the irregular form `=foo`), yet they live in the compiler, and while I only skimmed, I'm pretty sure they're the only compiler utilities being called in the parser. So this moves them into the parser.

There is one tradeoff that I think is very slightly negative, but doesn't outweigh the benefit of the parser not relying on the compiler: `+name:ap` had two `+open:ap` callsites, and I just replaced those calls with what `+open:ap` would have done. All it did for `%dbug` was step through it, and for `%tsgl` it desugared it to a `%tsgr`. There's probably other places where it might make sense to call `+open:ap` (like anywhere you step through a `%dbug`) because it happens to do the right thing for a certain tag, but we don't do it - presumably its just not worth the extra indirection.

Since this will cause everything to recompile with no benefit to the end user, it probably makes sense to send this out at the same time as #5873.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -7549,43 +7549,6 @@
           def=*(unit hoon)
       ==
   |_  mod=spec
-  ::
-  ++  autoname
-    ::  derive name from spec
-    ::
-    |-  ^-  (unit term)
-    ?-  -.mod
-      %base  ?.(?=([%atom *] p.mod) ~ ?:(=(%$ p.p.mod) `%atom `p.p.mod))
-      %dbug  $(mod q.mod)
-      %leaf  `p.mod
-      %loop  `p.mod
-      %like  ?~(p.mod ~ ?^(i.p.mod ?:(?=(%& -.i.p.mod) ~ q.i.p.mod) `i.p.mod))
-      %make  ~(name ap p.mod)
-      %made  $(mod q.mod)
-      %over  $(mod q.mod)
-      %name  $(mod q.mod)
-    ::
-      %bcbc  $(mod p.mod)
-      %bcbr  $(mod p.mod)
-      %bccb  ~(name ap p.mod)
-      %bccl  $(mod i.p.mod)
-      %bccn  $(mod i.p.mod)
-      %bcdt  ~
-      %bcgl  $(mod q.mod)
-      %bcgr  $(mod q.mod)
-      %bchp  $(mod p.mod)
-      %bckt  $(mod q.mod)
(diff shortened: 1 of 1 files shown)


drbeefsupreme:
If this is put in with #5873, another case for `%gist` needs to be added.

drbeefsupreme:
Oh yeah, since this changes `+ap`, it breaks its jet signature. So that also needs to be fixed for this to work. I haven't yet learned how to do that.

joemfb (commented):
I don't think these arms should move. `+ap` is the "hoon engine" and `+ax` is the "spec engine", so they're already in their natural places (regardless of the number and location of subsequent call sites). If I were going to change anything like this, I'd be inclined to see what I can factor out of the parser into these engines.

drbeefsupreme:
It's not yet compiling in this commit, but it looks like we'd probably need new `$hoon`, `$skin`, and `$spec` tags to do this. With how much work it took for me to understand the parser/compiler this year still fresh in my mind, I am almost certain I would have found this a lot more confusing than just calling `+autoname` from the parser.

https://github.com/urbit/urbit/commit/2259e0f2316e3489cfbe2857832ff3669e7634b7

If somebody has justification for this being an improvement, I'm happy to hear it. I have no idea if whether its totally normal for parsers to make use of compiler utilities outside of Hoon. If nobody comments in the near future, I'll assume this is not a popular idea, and this PR can be closed.

drbeefsupreme:
So my goal here was to make the parser and compiler a little more self-contained for the sake of future contributors trying to find their way around it. But the approach I started on with https://github.com/urbit/urbit/commit/07c3303bde1a9908cc795cde0fb15567f5d46973 that parses `=foo` as a new kind of token then requires the programmer to go into the compiler anyways to find out that it turns into a `foo` `spec` with a face of the same name. So it doesn't actually reduce the cross-referencing going on - the learner just sees another mysterious tag.

At least by looking at how the product of `+autoname` is used in the parser, we can immediately see what the `=foo` syntax is used for, which is not apparent in https://github.com/urbit/urbit/commit/07c3303bde1a9908cc795cde0fb15567f5d46973. So in fact the programmer doesn't need to go look at the compiler to accurately guess what it does.

And I think @joemfb 's argument that these utilities do in fact belong in the compiler is sound. The benefit of being able to say "everything for working with `spec` is located within `+ax`" and similarly for `hoon` and `+ap` seems to trump the benefit of having the parser be self-contained.

PR #6163, opened 2022-12-15
ames: add %snub

liam-fitzgerald:
Adds .snub to ames-state, a global blocklist for ships. If a packet is
received from a ship that is in the .snub set, it is immediately
dropped. Adds %snub to ames' $task, to allow manipulating this list


belisarius222:
This code looks good.  There does need to be a way to see which ships we've snubbed, too -- is that already scryable somehow?

belisarius222:
Thinking about this again, we should indeed make the data scryable, but even without that, this PR is still a Pareto improvement over the current codebase, so we should release it as soon as we can.

belisarius222 (approved):

philipcmonk (approved):
Added a logging entry because otherwise we'll forget

PR #6169, opened 2022-12-16
hoon: swap =+  foo  |@  ++  $  bar  --  for |*  foo  bar

drbeefsupreme:
`|*  foo  bar` is sugar for `=+  foo  |@  ++  $  bar  --`, and newbies find
the old style confusing. this switches out the `|@` pattern for the `|*`
one, at least in layer <=4. the only ones remaining are `+toad`, `+rune`,
and `+runo`, which are already tweaked in #5873 so we omit them here.

some jets are harmed, so it probably needs to be part of a kelvin release:

```
fund: in in, parent 6a6f885b not found at 7
fund: in apt, parent 3b38f03b not found at 7
```

should probably be merged alongside #5873

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -238,21 +238,18 @@
 ++  corl                                                ::  compose backwards
   |*  [a=$-(* *) b=$-(* *)]
   =<  +:|.((a (b)))      ::  type check
-  =+  c=+<.b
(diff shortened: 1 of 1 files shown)


philipcmonk (approved):
This looks good, I think we should try to get this into hoon 139

philipcmonk:
(don't worry about the couple of small conflicts between this and #5873, i've got a resolution locally)

PR #6178, opened 2022-12-29
changes needed for `%pyro` ships to function properly

tadad:


philipcmonk:
I understand why you need these for %pyro, but they obviously can't work on live ships.  Do you intend to refactor these in such a way that they can be turned on at runtime or something?  Or are you just going to patch them in when you're starting up %pyro ships?

PR #6184, opened 2022-12-31
clay: fix %open %s scry endpoint

### LINKED ISSUE #6181: clay: %open %s scry returns whole +prelude gate
**Describe the bug**

A `%open` `%s` scry returns the whole +prelude gate rather than calling it with the given path and returning the resulting vase.

**To Reproduce**
1. `.^(vase %cs %/open/gen/agents/hoon)`
2. look at type error
3. look at code in `+read-s`: `%open  ``open+!>(prelude:(aeon-ford yon))`

**Expected behaviour**
Should build prelude not return gate

 - OS: Arch linux
 - Vere: v1.15

**Notify maintainers**
If you happen to know who the appropriate maintainers are, consider mentioning them with an @ here.  You may want to use `git blame` to see who has last touched any relevant code.

tinnus-napbus:
%open %s scry currently returns the whole +prelude gate. This fixes it so it correctly builds and returns the prelude in a vase.

resolves #6181

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -4246,7 +4246,7 @@
           (page-to-cage:(aeon-ford yon) u.peg)
         ``cage+[-:!>(*^cage) cage]
       ::
-          %open  ``open+!>(prelude:(aeon-ford yon))
+          %open  ``open+!>((prelude:(aeon-ford yon) t.pax))
           %late  !!  :: handled in +aver
           %case  !!  :: handled in +aver
           %base-tako


philipcmonk:
Do you know if the most recent versions of the language server depend on this behavior?  @ryjm

ryjm:
hmm it looks like the language server assumes it's a vase?

```
  =+  .^(=open:clay %cs /(scot %p our.bow)/[dek]/(scot %da now.bow)/open/foo)
  =/  =type  -:(open loc)

  =.  preludes
    (~(put by preludes) uri type)
```

am i interpreting this correctly? if so it's been broken and this would fix it.

ryjm:
oh i see now, it does depend on this behavior - that's why we give it a dummy path since we assume it ignores it.

ryjm:
scratch that - i just tested, it does break `language-server`.

ryjm:
don'r we also need to change `open:clay` in `lull`? it's type is `$-(path vase)` and should just be a `vase` now.

ryjm:
okay i have a fix for the language server, will PR shortly.

tinnus-napbus:
wait is it *meant* to return a `$-(path vase)`? I assumed this was an error cos that's a weird thing to return rather than just returning the vase. If it's meant to do that maybe I shouldn't change

philipcmonk:
It probably was intentional, with the idea that calling that function
repeatedly is faster than scrying it out each time. But it's very weird,
and generally bad practice because you often leak the entire arvo state by
returning a function (depends if it has the scry function in its state).

jalehman:
This PR is superseded by #6293, closing.

PR #6189, opened 2023-01-05
gitattributes: remove export-ignores

matthew-levan:
We no longer need to omit `pkg/arvo/tests` from this repo's export archives (@jtobin confirmed in chat a couple weeks ago), and I need the folder to be bundled in archives for CI in the [new repo](https://github.com/urbit/vere).

### DIFF
--- a/pkg/arvo/.gitattributes
+++ b/pkg/arvo/.gitattributes
@@ -1,2 +1,0 @@
-.gitattributes export-ignore
-tests export-ignore


jtobin (approved):
Yeah this should be fine.  I haven't been making any direct use of these archives lately myself, but I'm fairly confident nothing is expecting/requiring that the tests subtree be absent.

Issue #5822, opened 2022-06-02
pier: serf unexpectedly shut down


Quodss:
**Description:**
After the last OTA, pier crashes with "pier: serf unexpectedly shut down". When starting the pier after the crash, it would crash again after few seconds.

In these couple of seconds i had time to print +vats. Here are the logs:

```
...
---------------- playback complete ----------------
vere: checking version compatibility
ames: live on 52142
conn: listening on \\.\pipe\urbit-conn-dozreg-toplud
http: web interface live on https://localhost:443
http: web interface live on http://localhost:80
http: loopback live on http://localhost:12321
pier (5773968): live
> +vats
%base
  /sys/kelvin:      [%zuse 418]
  base hash:        0v1v.p3a22.iv754.lt3ot.hpbee.0elcg.orl5e.fakce.r1pb5.sg0kt.46vqk
  %cz hash:         0v10.qk6vo.bpj1l.ckup2.mq161.g50a2.echpp.dh0q9.bmuvb.gvipe.c51sm
  app status:       running
  force on:         ~
  force off:        ~
  publishing ship:  ~
  updates:          tracking
  source ship:      ~zod
  source desk:      %kids
  source aeon:      120
  pending updates:  ~
::
%inet2022
  /sys/kelvin:      [%zuse 418]
  base hash:        0v1h.1odif.kbkos.7cef9.crs82.qdmhf.adl96.26c3n.gs27j.ta6t1.slsdk
  %cz hash:         0v1h.1odif.kbkos.7cef9.crs82.qdmhf.adl96.26c3n.gs27j.ta6t1.slsdk
  app status:       running
  force on:         ~
  force off:        ~
  publishing ship:  ~
  updates:          tracking
  source ship:      ~tocrex-holpen
  source desk:      %inet2022
  source aeon:      13
  pending updates:  ~
::
%studio
  /sys/kelvin:      [%zuse 418]
  base hash:        0vq.c6i3u.2lo4u.ucc6d.vhaia.gu9ic.1htet.gssjp.sdnaa.03e4m.ct81g
  %cz hash:         0vq.c6i3u.2lo4u.ucc6d.vhaia.gu9ic.1htet.gssjp.sdnaa.03e4m.ct81g
  app status:       running
  force on:         ~
  force off:        ~
  publishing ship:  [~ ~tirrel]
  updates:          tracking
  source ship:      ~tirrel
  source desk:      %studio
  source aeon:      19
  pending updates:  ~
::
%landscape
[...]

PR #6190, opened 2023-01-05
ci: remove the mingw build flow

Fang-:
As discussed internally at the end of last year, we are discontinuing Windows support for vere.

The tldr is that we presently don't have the resources to support non-*nix systems at the level of quality we want to deliver. Notably Windows is already very much a second-class citizen (#5822 and others). Windows users may want to run Urbit through WSL instead.

As such, the Window build step should be removed from our CI. Maybe we want to clean up build configuration, logic, compat stuff etc. also, but this PR just aims to have us not get red X's on broken Windows builds.


zalberico (dismissed):

zalberico (approved):
Thanks!

PR #6191, opened 2023-01-05
ci: remove frontend-tests flow

Fang-:
These are no longer being used as much as they were, most of the stuff it was hitting lives in separate repos now. Can be ripped out as such.

(I already went and disabled it in the Github UI.)


zalberico (approved):
Thanks!

PR #6197, opened 2023-01-11
http-api: Optional ship param for subscriptions

ajlamarc:
The [documentation for http-api subscriptions](https://developers.urbit.org/guides/additional/http-api-guide#subscribe) describes a ship parameter for the subscription, defaulting to `api.ship` if not specified.  This is not supported by the JS library (it always uses `api.ship`) which is confusing.

Also, if you wanted to subscribe to different ships in the same application, you would need to toggle `api.ship` somehow.  Rather than fix the documentation, it makes more sense to support this.

Tagging @sigilante, hopefully you can help me get this through soon.  And let me know if I can help update the docs.

### DIFF
--- a/pkg/npm/http-api/src/Urbit.ts
+++ b/pkg/npm/http-api/src/Urbit.ts
@@ -474,9 +474,8 @@ export class Urbit {
       err: () => {},
       event: () => {},
       quit: () => {},
-      app: params.app,
-      path: params.path,
-      ship: params.ship ? params.ship : this.ship,
+      ship: this.ship,
+      ...params,
     };
 
     const message: Message = {
--- a/pkg/npm/http-api/src/Urbit.ts
+++ b/pkg/npm/http-api/src/Urbit.ts
@@ -10,6 +10,7 @@ import {
   AuthenticationInterface,
   PokeInterface,
   SubscriptionRequestInterface,
+  SubscribeOnceOptionalParams,
   headers,
   SSEOptions,
   PokeHandlers,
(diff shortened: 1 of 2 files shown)


sigilante:
Thanks for this.

Why remove the timeout param?

ajlamarc:
> Why remove the timeout param?

I had to combine the new `ship` param and `timeout` into an `options` object so it can be called with any combination of the two.  I think if the function signature stayed as `subscribeOnce<T = any>(app: string, path: string, ship?: string, timeout?: number)` then if you wanted to set a timeout, you would be required to also set the ship.

like this you can call it with `subscribeOnce(app, path, { timeout: 1000 })`

ajlamarc:
I think `subscribeOnce` is buggy: it doesn't emit the `on-leave` event like we would expect.  An example:

```js
return await this.api
    .subscribeOnce(agent, this.subscribePath(key))
    .then((value: string) => {
        this.cache.set(key, value)
        return value
    })
    .catch(() => {
        console.error(`key ${key} not found`)
        return undefined
    })
```
This does retrieve the value, but doesn't close the subscription.  Because of this I get a related error:
```
PUT http://localhost:3000/~/channel/1673798640-9c5c2e
[HTTP/1.1 400 missing 12ms]

Uncaught (in promise) Error: Failed to PUT channel
```

Switching to a (more complicated) use of `api.subscribe()` does emit an `on-leave` and not cause the error:

```js
 // ... inside our get() function
return await this._getValueFromUrbit(key)


// TODO - does this have race conditions?
private async _getValueFromUrbit(key: string): Promise<string> {
    let result = undefined
    return await this.api
        .subscribe({
            app: agent,
            path: this.subscribePath(key),
            err: () => {
                throw new Error(
                    'Tome: the key-value store being used has been removed, or your access has been revoked.'
                )
            },
            event: (value: string) => {
                this.cache.set(key, value)
                result = value
            },
            quit: () => this._getValueFromUrbit(key),
        })
        .then(async (id) => {
            await this.api.unsubscribe(id)
            if (result === null) {
                console.error(`key ${key} not found`)
                result = undefined
            }
            return result
        })
}
```

ajlamarc:
https://github.com/urbit/urbit/blob/bb5bc67c227cd1e1f924b81b951ad47ec9699697/pkg/npm/http-api/src/types.ts#L62

Looking at this comment, do foreign pokes from the `http-api` not work? If so the docs should be updated.  (I haven't been able to get it working).

ajlamarc:
Remote subscriptions do go through, but are infinitely kicked without receiving back any data.

- Should still fix `subscribeOnce` to properly emit the onLeave event
- Either remove `ship` as an option from `poke` and `subscribe`, or make it very clear in the docs that these don't enable foreign contact

PR #6199, opened 2023-01-12
kiln: add `once` flag to |install

philipcmonk:
It's often useful to |merge a desk, but if you're still getting updates from your sync source, you may get overwritten in the future.  In this case, you want to merge and clear the sync source.  With this change, you can do this with:

```
|install ~ship %desk, =once &
```


Fang- (approved):
Cute. This looks fine!

zalberico:
This was reverted and did not ship, details in #6247

PR #6205, opened 2023-01-16
Extend btc-provider API to fetch block transactions

cyclomancer:
**Dependent on [this PR](https://github.com/urbit/urbit-bitcoin-rpc/pull/31) to `urbit/urbit-bitcoin-rpc`.**

- Adds a `block-txs` node RPC request flow to `%btc-provider` that fetches a complete list of raw transactions by blockhash

Volt, an implementation of Bitcoin [Lightning payments on Urbit](https://urbit.org/grants/volt-lightning-on-urbit), needs to find and categorize a transaction spending one of its channels as cooperative, forced, or malicious, and in the last case match the raw transaction to a specific revocation signature. Compared to alternative solutions (making thousands of redundant RPC requests, modifying the existing `getblockinfo` structure, fetching and parsing the entire raw block or verbose transaction JSON) this one has no impact on existing usage, a small code footprint, and provides functionality that will also be useful for future Bitcoin use-cases other than Lightning.

### DIFF
--- a/pkg/bitcoin/app/btc-provider.hoon
+++ b/pkg/bitcoin/app/btc-provider.hoon
@@ -180,6 +180,7 @@
       %tx-from-pos    [%get-tx-from-pos height.act pos.act merkle.act]
       %fee            [%get-fee block.act]
       %psbt           [%update-psbt psbt.act]
+      %block-txs      [%get-block-txs blockhash.act]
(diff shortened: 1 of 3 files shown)


zalberico:
Bitcoin was removed from urbit/urbit

PR #6207, opened 2023-01-17
tests: move to root, out of pkg/arvo

Fang-:
Including these in pkg/arvo has caused many minor problems over the years. We don't want to include them in releases, but this often meant excluding them manually, which was error-prone.

Here, we move them out of /pkg/arvo/tests and into /tests, with an eye towards non-arvo things eventually splitting off into their own repos, and so arvo tests in root being appropriate.

To keep CI from breaking, we add a tiny extra step before we call the `urbit-tests` nix script, that copies the tests back into their old place. This is not _the cleanest_, since the `urbit-tests` script now doesn't work stand-alone...

This also clobbers the github ci view to put the `cp` command as the header/title for that section instead of the more meaningful `nix-build -A urbit-tests`. We could `&&` to show them both, maybe, but would rather just give a manual name-string instead.


philipcmonk:
This doesn't appear to run the tests

Fang-:
I don't get it. If I change the `cp` command back to copying into `pkg/arvo/tests/tests` [then it does run them](https://github.com/urbit/urbit/compare/m/detest...m/detest-rage). Feels like I'm doing something stupid.

Fang-:
Never mind, I am big dumb. Should be better now. 5389898 is a commit off this branch that properly runs the tests. Unclear why they're not _always_ triggered, maybe nix caching shenanigans? Surely this behavior can't be caused by the changes herein?

philipcmonk (dismissed):
This works, just going to put it on next/arvo

philipcmonk (approved):

Issue #6208, opened 2023-01-18
ci: Need to download runtime


philipcmonk:
The runtime has moved into [a new repo](https://github.com/urbit/vere), but I believe that CI here continues to build the binary out of this repository.  This stops working as soon as we have a kelvin update where the binary changes aren't reflected here (yesterday, on next/arvo).

Two solutions that come to mind are including a version identifier in this repo (eg a commit hash, vere version number, or kelvin) or just unconditionally downloading edge.  The latter seems easier and will work for most things.

zalberico:
I like the idea of pulling edge because when we're shipping releases that require vere changes it means we always have the up to date commits, but the risk is for releases where we don't require vere changes.

In the case where we're shipping arvo work that doesn't touch vere, then pulling edge could make things more difficult. That said, the expectation is that develop should remain operable at all times so committing to this is probably a good thing. I think we should probably just do this and if we recognize that vere edge is broken quickly that'll be good (and we can then fix it early).

zalberico:
@belisarius222 tagging you on this - it'd be good to handle this from the foundation side.

belisarius222:
CC'ing @jalehman

philipcmonk:
The main downside of pulling exclusively edge is that if you're doing a kelvin release, your ci won't work until the binary changes hit edge.

I think the right answer is to refer to a specific version of urbit/vere.  It should only need to change every kelvin version anyway, and it gives you the most flexiblity.  That would be something like:
- Here, include a vere-version that includes something like `/edge/v1.17-ed12e552cb`
- Here, download that binary instead of building from this repo
- In urbit/vere, add a manual Github Actions workflow so that you can deploy a specific dev commit

zalberico:
Thanks - that makes sense to me.

philipcmonk:
Fixed by #6210

PR #6210, opened 2023-01-18
ci: Download runtime for unit tests

philipcmonk:
This is not trying to fix nix or ci in general, but it solves #6208.  It introduces `vere-version`, which is the runtime pace and version number to use when running the unit tests.  Generally this only needs to be updated every kelvin, but if you need a specific runtime version (eg a working branch for an upcoming kelvin release), you can add it to this file. 
 To upload a binary from non-develop/release/master branches, manually run the deployment action from urbit/vere#130.

This also deletes various things that broke and are no longer necessary, mostly because vere testing is happening in urbit/vere


philipcmonk:
The tests look like they're getting cached, certainly if you touch pkg/arvo, it'll rerun the tests.

philipcmonk:
Here's an example of it working for next/arvo, which has to pull a specific binary from urbit/vere: https://github.com/urbit/urbit/actions/runs/3953241890/jobs/6769293584

.github/workflows/vere.yml
```diff
@@ -76,122 +74,12 @@ jobs:
...
-          exit $exitcode
-
       - if: ${{ matrix.os == 'ubuntu-latest' }}
         name: run urbit-tests
         run: |
           cp -RL tests pkg/arvo/tests
+          vere="$(cat ./vere-version | sed -e 's/\([^ ]*\) \([^ ]*\)/\1\/\2\/vere-\2/g' | tr -d '\n')"
+          url="$(echo https://bootstrap.urbit.org/vere/${vere}-linux-x86_64)"
+          curl -Lo pkg/arvo/urbit.jam "$url"
+          chmod +x pkg/arvo/urbit.jam
```
Fang-, line 84: The reason for giving it `.jam` should probably be clarified in a comment so we can stop doing this as soon as it's no longer necessary.

(Also consider `vere.jam` instead, for consistency.)

.github/workflows/build.yml
```diff
@@ -16,21 +16,7 @@ on:
       - 'bin/**'
       - 'nix/**'
       - 'default.nix'
-  pull_request:
```
Fang-, line 19: Notably this actives "on" the merge of the PR branch and its target, if I understand the prior behavior and [the docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) correctly. I would expect it to still be desirable to automate running tests against the would-be end result of a merge, so unclear to me why this got removed.

Fang- (commented):
A question that needs addressing and a simple suggestion, but looks good otherwise.

.github/workflows/build.yml
philipcmonk: Ah, i didn't realize that.  Will re-add

Fang- (approved):
This looks good now!

PR #6211, opened 2023-01-19
clay: Drop foreign %wris responses

philipcmonk:
This flow is not supported, and it was causing issues releasing 416.  This change just drops the responses to avoid crashing, but at some point we should either support this flow or reject the request in the first place.


philipcmonk:
Insta-merging because it's already on ~zod (shh don't tell @zalberico)

PR #6215, opened 2023-01-19
garden: removing as it is no longer updated here

arthyn:
This removes garden and it's relevant components now that it is housed at tloncorp/landscape in preparation for renaming.

This has somewhat far-reaching concerns as garden-dev is required for most app flows and documented as being here. We need to account for this and communicate it appropriately as well as maybe come up with a solution for making this easier using something like %herd.


zalberico:
Thanks @arthyn we're actually handling this in another PR where we're removing lots of old stuff. I'm going to close this one in favor of that one. I'll tag you as a reviewer there.

zalberico:
Closing in favor of: https://github.com/urbit/urbit/pull/6217

PR #6224, opened 2023-01-20
http-api: make http api work in node context

pkova:
`http-api` does not work in node because it uses `@microsoft/fetch-event-source`. Luckily this helpful dude [made a PR](https://github.com/Azure/fetch-event-source/pull/28) with a fix. I would expect it to be merged in but we should use his fork for now because `http-api` not working on the server side is untenable.

I had to bump another dependency as well for ESM reasons.


belisarius222:
@pkova what is the latest status on this PR?  I can't tell from the discussion if there's anything blocking merging it.  Seems like there isn't.

arthyn:
This was moved over to urbit/js-http-api

PR #6217, opened 2023-01-20
Remove unused components

tinnus-napbus:
Fixes #6216

We moved Vere out of the urbit repo. As part of this clean up we're also removing a few other unused items and removing bitcoin from the code (and from the default pill: https://github.com/urbit/urbit/issues/6204)

- bitcoin: removed
- garden: new location https://github.com/tloncorp/landscape
- hs: removed
- urbit (vere), docker-image, ent, urcrypt: new repo https://github.com/urbit/vere

PR #6232, opened 2023-01-20
meta: remove npm packages

arthyn:
In continuation with efforts in #6217 this removes npm packages from this repo. This does affect #6224 which will need to be duplicated or moved over to the new repo.

- @urbit/http-api will now live at https://github.com/urbit/js-http-api
- @urbit/api will be somewhat split up and reconstituted as a utils library in another repo. The current iteration is a weird amalgamation of APIs from old groups, garden/landscape, and kiln along with helper utilities. Since most of these apis are for Tlon products we'll take these over and implement a new package.
- @urbit/eslint-config seems mostly irrelevant and continue to exist as is on NPM, but nothing about it is specific enough to urbit that it's worth keeping here. If someone feels strongly they can reconstitute it in it's own repo elsewhere.


tomholford (approved):
LGTM

patosullivan (approved):

PR #6237, opened 2023-01-22
clay: allow scrying out of pending requests

### LINKED ISSUE #6236: clay: incoming clay requests are opaque to userspace
Userspace currently has no way of finding out anything about the pending incoming requests in clay, but these might be useful for metrics, among other things.

Fang-:
Implements a `/cx/[our]//[now]/cult/[desk]` endpoint, for getting a set of pending requests for any given desk. We don't give the $cult for the desk as-is, but instead slim the $roves back down into $raves, remove clay protocol version metadata, and make sure to put our `@p` in place of empty "for" fields.

Also adds a `+desk-requests` generator, for counting the different types of pending requests. It specifies the number of local and incoming-from-foreign requests, and for the latter, how many of those are awaiting the next revision of a desk (read: are currently |sync-ing from the desk).

Tested on my live ships without issue.

Closes #6236.


Fang-:
I did implement it as a `%s` request at first, but it felt like the odd-one-out, especially wrt to the paths `+read-s` expects:

https://github.com/urbit/urbit/blob/6b80aec3c2f539c5ada95551a04a5f39d638c783/pkg/arvo/sys/vane/clay.hoon#L4210-L4214

`+read-buc` felt like a more natural fit: all of the things there are about clay-level metadata moreso than things that can be said to live in the desk's data in some sense. Request status feels more like the former than the latter.

belisarius222:
Hm. Well I just merged another Clay scry PR that was pretty similar, returning the set of ships subscribed to a desk, as a %cs scry. I'm not sure I have a principled answer to where these things should go, then.

I was mostly surprised by the %x in particular, since I associate that with file contents, but maybe I'm just behind the times and there are other +read-buc scrys that are also %x scrys, and I haven't used them enough to remember.

I'm starting to feel a smell that there are too many query details in Clay itself, and maybe we should try to reduce it to one scry that returns general Clay state and one that returns Clay state for a desk -- these diagnostic scrys are proliferating quickly, so I think we're missing some grouping to keep them more organized.

philipcmonk:
I think the first organization is to put all diagnostic scries (especially
anything in +read-s) into +read-buc, so all the disorganization is in one
place.  Then we can try to make that more principled as it becomes more
relevant.  I do think it's important that we don't hold back adding these
scries until we have this organization.

belisarius222:
That sounds good to me.

belisarius222 (dismissed):

belisarius222:
There was a trivial merge conflict in Clay due to recently merged commits.  I resolved it, so once CI passes, I will merge this into develop.

belisarius222 (approved):

PR #6241, opened 2023-01-23
MAINTAINERS.md: improve tagging instructions

### LINKED ISSUE #6240: Add release candidate tagging instructions to maintainers.md
The current docs don't explain this in enough detail.

When we branch release to deploy to marnec, we need to tag it as an rc. `urbit-os-vx.y-rc1`. After this any change that goes into release gets a new tag that increments the rc.

The tag without rc only happens after we ship to the live network (no more changes can come after it).

belisarius222:
Fixes #6240.

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -72,20 +72,27 @@ release branch is cut from `develop` and the process begins again.
 
 ### Tagging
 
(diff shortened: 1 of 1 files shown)


zalberico (approved):
Thanks! Way clearer.

Issue #6231, opened 2023-01-20
pill: we can use the autoprop tooling from byob


Fang-:
#5470, which is presently still unreleased (stuck in mars/urth limbo), comes with an "autoprop" app for auto-building pills and new-style boot props.

This app doesn't really have a hard dependency on any of the changes in that PR, it works perfectly fine without them, except that we can't use the jam files it creates yet. But it can also create pill files, which we _can_ use already!

We don't presently have a good process for uploading new pills (for booting new ships) whenever desk contents change. We do not want to do this at the moment of a release, because this risks booting you into something newer than what your sponsor has. The autoprop app has a five-day delay to accommodate this, but can also be told to re-build pills or props on-demand.

The tool outputs the configured file(s) into the `pier/.urb/put` directory, with the runtime version string in its path, and as such a simple script[1] can be used to upload these to an appropriate place in the bootstrap bucket automatically.

(The app has [a little bit of logic](https://github.com/urbit/urbit/blob/39a104f872736c3b0b879d592302b86a67ad078b/pkg/autoprop/app/autoprop.hoon#L188-L190) for stripping the "vere" out of the version string, but maybe we want to intentionally leave that in at this point.)

All of this already exists, and has been running since that PR, on the `autoprop-builder` instance, and is still live... except its vere version is still 1.10. We should just bring this back up to latest, set it up to build pills, and make a point to update its binary for new releases. If we do that, and maybe update vere to use a new pill url format (#5470 will do this anyway), then we can have nearly fully-automated pill generation.

[1]: <details><summary>`upload-props.sh`</summary>

```bash
#!/bin/bash
inotifywait -e close_write -m -r $1 |
while read -r directory events filename; do
  # skip events on the directories
  if [[ $events == *"ISDIR"* ]]; then
    continue;
  fi
  # props get written to folder with runtime version name
  version=$(basename $directory);
  sleep 2;
  date;
  echo $version/$filename;
  gsutil cp $directory$filename gs://bootstrap.urbit.org/props/$version/$filename;
  echo;
done
```

</details>

zalberico:
tagging @jalehman and @belisarius222 for context - this is the way we can make creating and shipping pills automatic (and something we can ship now)

Fang-:
We'll probably also want to update the autoprop builder to support auto-building new azimuth snapshots.

belisarius222:
[...]

PR #6243, opened 2023-01-23
autoprop: automated pill (& other boot prop) building

Fang-:
This is the arvo part of #5470, cherry-picked and updated to account for recent clay changes.

Primary addition here is `/pkg/autoprop`, a utility for automatically building pills and "prop" files, the latter of which will only become useful once the vere side of #5470 lands. (I am not currently looking at backporting that, but could be convinced to give it a quick shot.) In the immediate term, we intend to use that utility to keep the bootstrap pill up to date.

Secondary additions include a `%prep` task for clay to prime its blob storage (for quicker post-boot desk syncing), and a slightly refactored pill library & generators.

Once we get this onto the moon I've set up for this purpose, we can mark #6231 as done.


zalberico:
Since this isn't a kelvin we should be able to merge this (pending CR and passing tests) and get it in for the next release (currently that will branch after doccords on Jan 31st with a target ship date of Feb 7th).

@belisarius222 and @jalehman for context

philipcmonk (dismissed):
This looks good to me.  CI is failing in a weird place, kind of looks like a transient networking error, so i'll re-run now.

Fang-:
@philipcmonk I noticed I had missed the old-style `%kiln-install` poke sent by dill. iiuc the actual value of the `once` flag doesn't matter here, since it's a purely local install. `&` seems most appropriate regardless though. ptal

philipcmonk:
That change actually got reverted in #6250 because %kiln-install is used in a bunch of places.

philipcmonk (approved):

PR #6246, opened 2023-01-24
ames: put "is ok"/"not responding" behind |ames-verb

### LINKED ISSUE #6245: ames: "not responding"/"is ok" printfs are overly verbose
These have existed forever, but they're only occasionally useful.  Otherwise they add a lot to the scrollback, which makes many bugs unnecessarily hard to debug.  Now that we have the `|ames-verb` system, we should place them behind one of those flags, defaulting to off.

philipcmonk:
Fixes #6245

Respects both `|ames-verb %kay` and `|ames-sift`

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -436,7 +436,7 @@
   +$  address  @uxaddress
   ::  $verb: verbosity flag for ames
   ::
-  +$  verb  ?(%snd %rcv %odd %msg %ges %for %rot)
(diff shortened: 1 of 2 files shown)


belisarius222:
This doesn't break /debug, does it?  It does change the Ames state, but maybe not in a way that /debug cares about.

philipcmonk:
Yeah, debug doesn't read verbosity state

belisarius222:
The code changes all look good.

belisarius222 (approved):
LGTM

PR #6248, opened 2023-01-24
kiln: fix |ota regression

### LINKED ISSUE #6247: `|ota` crashes
#6199 modified the internals of `|install` but didn't make the corresponding changes to `|ota`.  This should be fixed and probably hotfixed to the prerelease on ~marnec

philipcmonk:
Fixes #6247


philipcmonk:
Actually, there are several other things that use %kiln-install, including docket, azimuth, and the landscape frontend.  Better to revert for now, but we should reimplement this feature, maybe with a `%kiln-install-2` or something.

PR #6249, opened 2023-01-24
kiln: Keep %kids desk in sync after kelvin update

### LINKED ISSUE #6244: %kids desks are out of sync with %base desks on most ships
This can most clearly been seen from the network explorer: https://network.urbit.org/?chart=kidsHash

Another way to verify is to run something like `-read [%z ~dinleb-rambep %kids da+now /]` from the dojo. Note that affected ships have the most up to date `%base` desk but are one commit behind on the `%kids` desk.

@philipcmonk is looking into this.

philipcmonk:
This is the fourth option described in #6244.

Fixes #6244

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -1157,7 +1157,7 @@
       ::
       ~>  %slog.(fmt "finished downloading update for {here}")
       =.  let  +(let)
-      ::  If nothing changed, just advance
+      ::  If nothing changed, just ensure %kids is up-to-date and advance
       ::
       ?.  (get-remote-diff our syd now [her sud (dec let)])
         =<  next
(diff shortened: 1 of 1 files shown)


zalberico:
@Fang- (or anyone else on @urbit/kernel) can you review this so we can merge in order to ship in the next release branch?

Fang- (approved):
The change here seems reasonable to me. But I'm a little bit unclear about the activation conditions you described in #6244.

> This relies on the current behavior that when spider reloads, all our syncs restart, which is a little bit incidental, but we can replicate this behavior if it ever changes.

It's not immediately obvious to me how/why this is the case, but iiuc this will cause kids syncs that are currently stuck to restart (and hit this new logic) the next time we push a change to spider, or anything higher up in its dependency tree? Shouldn't this also just trigger on the next sync activation?  
Either way, since this is a one-off trigger we depend on, we shouldn't need to make explicit note of it right?

philipcmonk:
Not sure what you mean about one-off trigger, this is important for every kelvin update that blocks on apps.

zalberico:
Is this good to merge?

Fang- (approved):
We discussed this out of band. I understand the context a bit better now. Still looks good!

PR #6258, opened 2023-01-27
webterm: safer printf rendering

### LINKED ISSUE #6257: webterm: eats scrollback to render printfs
See urbit/vere#168 and urbit/vere#175 for further details. tldr: `S` CSI code bad, simple `\n` good enough for xterm.

Fang-:
As it turns out, the exact behavior of the 'S' CSI command is not "move contents up into scrollback", but rather "delete the top line(s), move other contents up". This behavior leads webterm to eat into outputs whenever it renders slogs.

Xterm(.js), when given a newline character at the bottom of a scroll region, does produce the desired behavior of bumping the top line away into scrollback, instead of obliterating it.

This implementation now diverges from vere's, which is unfortunate, but the alternative is making webterm track the contents of the bottom-most line of the default session, which seems rather ridiculous by comparison.

Fixes #6257.

### DIFF
--- a/pkg/interface/webterm/lib/blit.ts
+++ b/pkg/interface/webterm/lib/blit.ts
@@ -58,15 +58,15 @@ export const showBlit = (term: Terminal, blit: Blit) => {
 
 export const showSlog = (term: Terminal, slog: string) => {
   //  set scroll region to exclude the bottom line,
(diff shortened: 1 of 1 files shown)


tomholford (approved):
LGTM

PR #6259, opened 2023-01-28
dojo: match hoon.hoon %know %hint printing style

Fang-:
Hoon.hoon got the separating `+` added in 13d1c28 (part of #5641), but dojo was never updated to match. Here, we pass it on.

Super tiny change, so figured I should PR against the release candidate. Of course this logic really shouldn't live in two places, but here we are.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -709,7 +709,7 @@
       [%hint *]  ?+    q.p.a  $(a q.a)
                      [%know *]
                    ?@  p.q.p.a  [(cat 3 '#' mark.p.q.p.a)]~
-                   [(rap 3 '#' auth.p.q.p.a (spat type.p.q.p.a) ~)]~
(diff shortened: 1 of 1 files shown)


jalehman:
We're going to keep to the process and not slip anything into the release candidate outside of the normal workflow. Please change this to target develop, and it can proceed from there to the next release candidate and then to eventual release.

jalehman:
@belisarius222 is going to test this in a fakezod dojo. Assuming it works, we'll merge.

belisarius222:
@Fang- I'm trying to test this out, but I'm having trouble figuring out how to get something other than just a mark name as a %hint type.  Syntactically, it looks like `$+` only takes in a `sym` before the second subexpression, so I don't know what I'm missing here.

Fang-:
Yeah I don't think there's syntax support for generating such hints yet. You can manually construct a type obviously but unsure how to do that in such a way dojo's `??` (which calls `+dy-show-type-noun`) picks that up. This is mostly just here for consistency.

belisarius222:
Ok let's merge it, then.

belisarius222 (approved):

PR #6269, opened 2023-02-01
Implement solid-state publications in userspace

### LINKED ISSUE #6268: Implement solid-state publications in userspace
As per here: https://roadmap.urbit.org/project/solid-state-publications

wicrum-wicrun:
Resolves #6268

This PR creates an SSS library. Can be used to test out the ergonomics of this programming model and will be fully supported by the UF and eventually be implemented in kernelspace.

Example usage here, including instructions: https://github.com/wicrum-wicrun/sss/blob/master/urbit/app/simple.hoon


wicrum-wicrun:
@belisarius222 This code is done now as far as I'm concerned. Should I change the PR to non-draft or do you want documentation first?

belisarius222:
@dr-frmr and @rabsef-bicrym please review this -- we can't tag you as official reviewers yet because of github bureaucratic difficulties, but if you could leave your reviews as comments on this PR, that would be appreciated.

tacryt-socryp:
Given that this library is meant to be used by developers, it would be extremely helpful if it were heavily commented.

pkg/base-dev/lib/sss.hoon
```diff
@@ -0,0 +1,280 @@
...
+    [fail rock]:rok.flow
+  ::
+  ++  chit
+    |=  [[aeon=term ship=term dude=term path=paths] =sign:agent:gall]
+    ^+  sub
+    ?>  ?=(%poke-ack -.sign)
+    ?~  p.sign  sub
+    %+  ~(jab by sub)  [(slav %p ship) dude path]
+    |=  =flow
+    ?.  =(aeon.rok.flow (slav %ud aeon))  flow
```
belisarius222, line 77: silent no-op here seems dangerous to me -- at least print, maybe crash

pkg/base-dev/lib/sss.hoon
```diff
@@ -0,0 +1,280 @@
...
+              %poke   %sss-on-rock  :-  on-rock-type  ^-  from
+              [path src.bowl dude rock.rok.flow ^wave]
+      ==  ==
+    --
+  --
+++  du
+  |*  [=(lake) paths=mold]
+  =>
+    |%
+    +$  rule  [rocks=_1 waves=_5]
```
dr-frmr, line 173: I can't figure this out -- what is being tracked here with the constants 1 and 5?

philipcmonk:
I think this basically makes sense.  The real test will be writing real apps in it.

rabsef-bicrym:
I recently wrote an app using this model and it worked well for my purposes.

I'm reviewing my prior commentary and I'm a little dissociated from it now but I guess I had roughly two questions:

1) What sort of events might be emitted in on-rock? It seems like it has its own baked in effects but I guess there could be any number of sides - if there's a train of thought here on why this isn't just something that happens w/ no side effects other than "because you can also have side effects", I'd like to know what it is so I can better conceptualize it.

2) Imagine a game scoring application that sends out waves of rounds of games - think chess tournament. Would the rock be the aggregate score and the waves be the changes? Is that the right way of thinking about it? Like the rock would be everyone's chess ranking score (whatever it's called)?

philipcmonk:
A +wave +wash'es over a +rock, altering it slightly. What does a +wave do to a +tide? Also I think of +rock as a snapshot, ground truth, more than a reset.

pkg/base-dev/lib/sss.hoon
wicrum-wicrun: It crashes now.

wicrum-wicrun:
> I recently wrote an app using this model and it worked well for my purposes.

I'm glad! :)

> I'm reviewing my prior commentary and I'm a little dissociated from it now but I guess I had roughly two questions:
> 
> 1. What sort of events might be emitted in on-rock? It seems like it has its own baked in effects but I guess there could be any number of sides - if there's a train of thought here on why this isn't just something that happens w/ no side effects other than "because you can also have side effects", I'd like to know what it is so I can better conceptualize it.

Because agents should always be allowed to when their state changes. I don't really care what they want to do. Only exceptions are during `+on-save` and `+on-peek` because those are strict *reads*, nothing from the apps perspective changes from them triggering. `+on-rock` is not a read, since it signifies that a state that the agent is interested in changes. Does that track? (For any others reading this, `+on-rock` doesn't exist in the version in this PR, it's equivalent to being poked with the mark `%sss-on-rock`.)

> 2. Imagine a game scoring application that sends out waves of rounds of games - think chess tournament. Would the rock be the aggregate score and the waves be the changes? Is that the right way of thinking about it? Like the rock would be everyone's chess ranking score (whatever it's called)?
> 
> Edit: the alternative in 2 is like passing around raw score all the time and the rock just being the aggregate of all scores, the waves being completed game scores. I like this because it allows more review of the data, say.

Yes, if there's a scorekeeper ship in the system, the rock would be the aggregate score and the waves would be the changes. I would probably make it so that each wave is a `[?(%win %lose %draw) @p]`, but you could do it other ways too, for example by distributing the whole games, including both all moves made by both players. Depends on how deep into replayability you want to go.
 
> Edit 2: I liked about it that I could customize the frequency of the behaviors and how many waves between rocks etc. Keep that please.

Don't worry, it's staying! :) In this current PR it's even user-configurable using an arm (`+rule:du`) instead of having to edit the code!

wicrum-wicrun:
> Given that this library is meant to be used by developers, it would be extremely helpful if it were heavily commented.

Extensive usage instructions are now available at https://github.com/wicrum-wicrun/sss/blob/master/urbit/app/simple.hoon

The user-facing arms in the library itself have also been commented, though not as extensively.

belisarius222:
Docs look great.

Why does the publisher set behn timers?

belisarius222:
Oh, it's because of this:
https://github.com/urbit/urbit/pull/6269/files#diff-1689977305924e45e986421ff1b6ce14b3086328bbdf54c4bdbbc14d6654226cR161

That makes sense.  The publisher is deleting the "sticky scry request" after a certain amount of time, to release itself from the obligation of responding with a result if it hasn't heard a fresh scry request from the subscriber.  I wonder if we could get away with just checking the received date of the latest scry request when the next wave gets published, so the publisher wouldn't have to set timers.  Instead, when it's about to publish a new wave, it looks through the subscriber scry requests and only responds to ones that are fresh enough.  It deletes the other ones.  Each time it hears a new scry request from a subscriber, it overwrites the "latest-heard" field for that subscriber.

wicrum-wicrun:
Yeah, I was thinking of both of those. I realize now that this current version is a trace from when I thought the publisher didn't need to store the `@da` at all. But now that it already does, we may as well not set timers. Pushing new version momentarily.

belisarius222 (approved):
The architecture, code, and documentation all look good to me.  UF will support and maintain this code so that app devs can be confident it works and won't be abandoned.  This is still new code, though, and it's only in experimental use so far, so it's likely we'll uncover some issues with it.  We should be on the lookout for bugs and scaling issues until it's been bulletproofed by actual use for a while.

PR #6271, opened 2023-02-01
Update issue templates

belisarius222:
For creating an issue just to make a PR, and some other use cases, it's convenient to be able to create a blank issue.


jalehman (approved):

Fang-:
This option has always been available at the bottom of the template picker page:

image

PR #6286, opened 2023-02-03
language-server: fix %open %cs scry

ryjm:
Fixes `language-server` in preparation for #6184


belisarius222:
Hm, I didn't realize language server was in the same repo as dojo. I assume this means we should release either both of these PRs simultaneously, or neither.

belisarius222 (dismissed):

jalehman:
@belisarius222 let's create a third PR that updates the requisite updated type definition and has these two branches merged into it.

ryjm:
👍 #6293

Issue #6266, opened 2023-01-31
A desk called '%kids' breaks +vats


ryjm:
**Describe the bug**
I have a desk called `%kids` that is preventing +vats from working (and also breaks apps that do anything similar, like realm's `%bazaar`.

```
> .^((set desk) %cd /=base=)
{ %bitcoin
  ...
  %%kids
  ...
}
```
**To Reproduce**
Steps to reproduce the behaviour:
1. Create a desk with a '%' in it (i.e like `|new-desk '%test'`)
2. Try +vats
3. Observe error

**Expected behaviour**
Doesn't crash on poorly named desks.

vvisigoth:
@Fang- remember what I said about my trepidation wrt to "magic desks"?

belisarius222:
Does Clay not (sane @tas) desk names? Seems like it should, to prevent getting into this situation

vvisigoth:
Came here to say what @belisarius222 said.

%%kids is not a valid `@tas` and thus shouldn't be allowed at all. Where in the stack do we @ryjm -proof this?

ryjm:
Is there a way to turn `+$  desk` into such a validator, or is this why `(sane %tas)` exists?

ryjm:
clearly we need to `symp` more

```
++  symp                                                ::  symbol or empty
  |=  a=*  ^-  @tas
  ?.(&(?=(@ a) ((sane %tas) a)) %$ a)
```

joemfb:
The quick fix is to assert `(sane %tas)` in %clay. Longer term, the answer is to automatically include `+sane` in atom molds (via `+ruth`).

ryjm:
Any ideas for fixing ships who already have illegal desk names?

joemfb:
Write a custom `%clay` migration and just delete them, then revert to the released version.

ryjm:
simple, elegant, urbit

PR #6290, opened 2023-02-04
gen: forbid bad |new-desk names

ryjm:
Minimal change to protect against #6266 

Ideally clay should do this, but we should get one or the other in asap before bad things happen.


jalehman (changes requested):
This assertion should be in clay.

ryjm:
> This assertion should be in clay.

Closing in favor of #6304

PR #6293, opened 2023-02-06
clay, language-server: fix %cs %open scry

ryjm:
Combined PR of #6183 and #6286


belisarius222:
`+open` in Lull should really be deleted.  We don't need an alias for `+vase` here.  That being said, we might want to defer modifying Lull at all until we have a Kelvin release.  I think we could add an inline comment in Lull like `XX delete` without triggering incompatibility, in which case that would be the way to go.  @joemfb right?

ryjm:
I used `XX remove` since there are a few `delete` comments referring to clay operations, ha.

belisarius222:
```
clay: read-at-aeon fail [desk=%base care=%a case=[%da p=~2023.2.9..14.14.55..3eea] path=/lib/wolfur-patch/hoon]
nest-fail
- have
[ i=[p=[@p @tas] q=[[let=@ud hit=nlr([p=@ud q=@uvI]) lab=nlr([p=@tas q=@ud])] tom=nlr([p=@uvI q=#3]) nor=#3 liv=?(%dead %held %live) ren=nlr([p=@tas q=?(%.y %.n)])]]
    t
  it(
    [ p=[@p @tas]
      q=[[let=@ud hit=nlr([p=@ud q=@uvI]) lab=nlr([p=@tas q=@ud])] tom=nlr([p=@uvI q=#3]) nor=^#3.[fil=u(?(%.y %.n)) dir=nlr([p=@ta q=#3])] liv=?(%dead %held %live) ren=nlr([p=@tas q=?(%.y %.n)])]
    ]
  )
]
- need
?(
  %~
  [ i=[[ship=@p desk=@tas] dome=[let=@ud hit=nlr([p=@ud q=@uvI]) lab=nlr([p=@tas q=@ud])] tom=nlr([p=@uvI q=#3]) nor=#3]
    t=it([[ship=@p desk=@tas] dome=[let=@ud hit=nlr([p=@ud q=@uvI]) lab=nlr([p=@tas q=@ud])] tom=nlr([p=@uvI q=#3]) nor=^#3.[fil=u(?(%.y %.n)) dir=nlr([p=@ta q=#3])]])
  ]
)
/lib/wolfur-patch/hoon::[14 3].[14 17]>
/lib/wolfur-patch/hoon::[13 1].[53 3]>
/lib/wolfur-patch/hoon::[12 1].[53 3]>
/lib/wolfur-patch/hoon::[11 1].[53 3]>
/lib/wolfur-patch/hoon::[10 1].[80 3]>
/lib/wolfur-patch/hoon::[9 1].[80 3]>
/lib/wolfur-patch/hoon::[8 1].[80 3]>
/lib/wolfur-patch/hoon::[7 1].[80 3]>
/lib/wolfur-patch/hoon::[6 1].[80 3]>
[%error-building /lib/wolfur-patch/hoon]
clay: %a build failed [%base 2 /lib/wolfur-patch/hoon]
```
@ryjm this is on a fakezod created from the solid pill that's in your branch

jalehman:
Ping @ryjm

> this is on a fakezod created from the solid pill that's in your branch

ryjm:
How did you make the pill? Any chance it was created with old
dependencies? Compiles fine when I apply the changes directly to clay/lull.

belisarius222:
Considering how much trouble we've been having lately with bail:meme handling in Vere, we should be careful for the moment not to change Lull unnecessarily.  Ideally this PR would leave Lull untouched and use `$vase` in the client code instead of `$open:clay`, so that we could deploy this without needing to rebuild vanes and agents.

A second tiny PR could delete `$open:clay` from Lull, and we'll merge that one whenever we're already making other changes to Lull.

belisarius222:
After discussing this with @ryjm, it seems like the language-server might need to keep the current scry for performance reasons -- if we make this switch, language-server would need to run a lot more `.^`s, which are slow.  Considering that this would be a breaking change, and there's not a clear benefit of the proposed API beyond being less "weird", I'm closing this.

It is indeed strange to return a gate from a scry, but Clay already does that with other scry requests, namely mark conversion gates.  I'm not opposed to coming up with a better API that doesn't need to return a gate, but this particular change doesn't seem worth the hassle.

I'm open to other opinions on this.  If people clamor for this change, let's go ahead and do it.  I don't think that's going to happen, though.

PR #6304, opened 2023-02-08
clay: desk name sanity check

ryjm:
Ensures no desk with an illegal name can be created by sanity checking in `%fuse`, `%park`, and `%merge`.

Prevents the cause of #6266 but doesn't fix ships with bad desk names already.

Replaces #6290.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -4827,7 +4827,7 @@
   ::
       %merg                                               ::  direct state up
     ?:  =(%$ des.req)
-      ~&(%merg-no-desk !!)
+      ~|(%merg-no-desk !!)
     ?.  ((sane %tas) des.req)
       ~|([%merg-bad-desk-name des.req] !!)
     =^  mos  ruf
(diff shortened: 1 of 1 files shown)


ryjm:
Any idea for a fix for ships who already have bad desk names?

pkg/arvo/sys/vane/clay.hoon
```diff
@@ -4828,6 +4828,8 @@
       %merg                                               ::  direct state up
     ?:  =(%$ des.req)
       ~&(%merg-no-desk !!)
+    ?.  ((sane %tas) des.req)
+      ~&(%merg-bad-desk-name !!)
```
Fang-, line 4832: Consider including the desk name in the print, and making the print part of the crash trace so it renders more obviously as the cause of the crash.

```suggestion
      ~|([%merg-bad-desk-name des.req] !!)
```

belisarius222:
The thing I'd check here is that every entry point that can create a desk gets checked.  Ideally, there would be one spot where a new desk is created, and it would be checked there, to reduce duplication and make that check trivial.  I'm not sure if there's just one "create desk" pathway in Clay, though, so it might or might not be doable.

Fang-:
Well, `+de` gets initialized with a desk name. Presumably just checking there would do the trick, but is no bueno because desks with invalid names may already exist in the wild? Wouldn't be hard to write a state adapter that just obliterates local (copies of) desks with invalid names, but becomes a little bit less trivial if we want to gracefully handle requests that might have been open to them, or other state/pointers we would want to clean up...

Issue #6303, opened 2023-02-07
eyre: no verbose mode


tadad:
A verbose mode would be very useful for debugging various bugs in `%eyre` - mainly the infamous reconnect bug in landscape

PR #6306, opened 2023-02-08
eyre: add verbosity levels

tadad:
addresses #6303 , adds a verbosity toggle to %eyre with the `%spew` task, and some new printouts to help with debugging. Verbosity has three levels:
`0` : These will always print out. They are legacy printouts from before this PR that would always fire
`1` : includes discarding channels; subscriptions; cancelling subscriptions; cancelling requests; response errors; authentication errors
`2` : logs every http message response - %start, %cancelling, %completed. It is very verbose, sometimes useful.

not partial to these rankings, open to changing the value on all the messages


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2526,10 +2547,10 @@
       ?>  ?=([%behn %wake *] sign)
       ?^  error.sign
         [[duct %slip %d %flog %crud %wake u.error.sign]~ http-server-gate]
-      =/  discard-channel
-        discard-channel:by-channel:(per-server-event event-args)
+      =*  id  i.t.t.wire 
+      ~>  %slog.[0 leaf+"eyre: {(trip id)} cancelling channel due to timeout"]
```
belisarius222, line 2551: This looks like a new print.  I think it should use +trace also.

belisarius222 (changes requested):
A couple minor nits, but this is very close.

belisarius222 (approved):

belisarius222 (changes requested):
Once CI is fixed, I will approve again.

tadad:
Apparently there was another state change between when I started this branch and now. `+load` is working now

belisarius222:
Tests failed, but it looks spurious -- died after 2 minutes in something other than Arvo.  Rerunning.

belisarius222 (approved):

PR #6251, opened 2023-01-25
Kelvin shims

pkova:
Fixes #6073 
Continuing from #6104 and #6239, had to make a new PR since I mistakenly created the previous one from `pkova/urbit`.

PR #6310, opened 2023-02-09
%base: add type restricting userspace moves to subset of arvo moves

pkova:
During the work on #6251 it became apparent that this would be a good opportunity to add a type disallowing userspace from emitting nonsense moves like `%born`. This change is sufficiently orthogonal to warrant a PR of its own.

This PR adds the necessary types to `/sys/lull`, `/sys/vane/gall` and modifies the files in `/app` and `/lib` to use the new type `$sign-user:agent:gall` instead of `sign-arvo`. Some tests will certainly fail but I don't know how to run them locally anymore so I'll fix them when CI fails.

Note that this change will be a breaking change for all Gall apps that have implemented the `+on-arvo` arm. It will also disallow `|pass` from passing arbitrary moves, so if this goes in we will need some other escape hatch.

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1845,7 +1845,7 @@
   +$  suss  (trel dude @tas @da)                        ::  config report
   +$  well  (pair desk term)                            ::
   +$  neat
-    $%  [%arvo =note-arvo]
+    $%  [%arvo =user-note:agent]
         [%agent [=ship name=term] =deal]
         [%pyre =tang]
     ==
@@ -1870,7 +1870,7 @@
     +$  card  (wind note gift)
     +$  note
       $%  [%agent [=ship name=term] =task]
-          [%arvo note-arvo]
+          [%arvo user-note]
           [%pyre =tang]
       ==
     +$  task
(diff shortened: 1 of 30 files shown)


pkova:
@belisarius222

pkg/arvo/sys/lull.hoon
```diff
@@ -1892,6 +1892,154 @@
...
+            $>(%nuke task:jael)
+            $>(%public-keys task:jael)
+            $>(%turf task:jael)
+            $>(%listen task:jael) :: base desk only
+        ==  ==
+          $:  %k
+        $%  $>(%fard task:khan)
+            $>(%lard task:khan) :: kiln does this
+        ==  ==
+      ==
```
pkova, line 2042: I'm very interested in comments of the list of userspace notes here. I created it by looking at Marks preliminary userspace permissioning work, then added all the moves that were needed for `%base` to compile. It is unlikely that I got it right the first time around.

jalehman:
Pyry, please read the [CONTRIBUTING](https://github.com/urbit/urbit/blob/develop/CONTRIBUTING.md) file and make this PR conform to the required format.

- [ ] Create an issue that describes what this PR addresses.
- [ ] Make the branch name conform to the format `i/<issue number>/description`

belisarius222:
> It will also disallow |pass from passing arbitrary moves, so if this goes in we will need some other escape hatch.

This is important, since a user must retain full control over their whole Urbit.  There are three main escape hatches I can think of:
1. A user can always inject an event from Vere (this functionality already exists, but it's not user-friendly)
2. We could add an `ovum` move that app desks could ask to send, which would be an untyped Arvo move.
3. The `%base` desk could have a different type.

I slightly prefer making 1. more user-friendly, since it enforces the cleanest separation between kernelspace and userspace.  Maybe Khan could be helpful here, or we could use Vere's `-I` event injection interface.  2. would be my next choice, since it at least decouples the types between and userspace and kernelspace.   3. seems less than ideal because then there would be effectively three components of the system instead of two, with the third being halfway kernelspace, halfway userspace, with its own types.

PR #6314, opened 2023-02-09
clay: correctly restart syncs when source breaches

### LINKED ISSUE #6313: When breaching OTA sources, the sync doesn't restart correctly
We noticed that ~mister-dister-dozzod-dozzod (which distributes %garden and %webterm) had a continuity error (probably at least a year ago), so we need to breach it.  We decided to test whether OTAs would automatically resume after a breach, and this behavior seems to have broken at some point.  A workaround is to breach twice -- the first breach puts the sync into a stuck mode, but the second one kicks it out to the start.

philipcmonk:
Fixes #6313

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -1095,6 +1095,7 @@
     %+  lard  /init
     =/  m  (strand:rand ,vase)
     ;<  =riot:clay  bind:m  (warp:strandio her sud ~ %sing %y ud+1 /)
+    ?>  ?=(^ riot)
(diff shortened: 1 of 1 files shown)


Fang- (approved):
This looks like what we discussed yesterday. To make sure I understand correctly, I'll reiterate: if a peer breaches, clay gives us `riot=~` responses for all outstanding requests to that peer. In kiln, crashing a thread (that handles the response) causes us to fall back to the "download failed", "retrying sync" logic, which will re-establish the ota source without making any assumptions about it: asks for the latest revision, downloads that, and then awaits newer ones.

Looks good. Superstition eradicated!

PR #6317, opened 2023-02-10
My channels migration

midsum-salrux:
Fixes tloncorp/landscape-apps#1412


pkg/landscape/lib/gladio.hoon
```diff
@@ -78,6 +76,18 @@
...
+  =/  assoc
+    .^(associations:met %gx [(scot %p our.bowl) %metadata-store (scot %da now.bowl) %associations %noun ~])
+  %-  ~(gas by *associations:met)
+  %+  skim
+      ~(tap by assoc)
+    |=  [m=md-resource:met [g=resource:res metdat=metadatum:met]]
+    ?&  =(| hidden.metdat)
+        =(name.g name.resource.m)
+        =(entity.g our.bowl)
+        =(%graph -.config.metdat)
```
arthyn, line 89: I think you want this to also make sure that it's not a chat so you don't pick up the dms. not sure if this will compile since it's a tagged union type:

```suggestion
        &(=(%graph -.config.metdat) !=(%chat module.config.metdat))
```

arthyn (changes requested):
looks good and works well, just one thing to filter the DMs

arthyn (approved):
lgtm

PR #6319, opened 2023-02-12
remove cruft from repo

### LINKED ISSUE #6277: meta: repo cleaning post-monorepo
Seems that there are a number of unused shell scripts, nix scripts, and files that could be removed in PRs akin to: https://github.com/urbit/urbit/pull/6232

belisarius222:
Resolves #6277.

This PR removes the following folders from the `pkg` directory:
- interface
- landscape
- webterm
- autoprop
- npm

It removes the makefile and the `nix` directory.  It also removes nix-specific scripts in the `sh` directory, and it modifies a couple of remaining scripts in that directory to no longer refer to the `nix` directory.

@zalberico If this breaks anything in Tlon's process, please let me know.  It doesn't need to be merged immediately, and it's possible I didn't catch everything or was overzealous in my deletion.  It will be good to merge this soon, though, to complete both the separation of urbit/urbit and urbit/vere and the separation of core from independently distributed apps.

Note that once pkg/webterm and pkg/autoprop are deleted from this repo, UF will need to create new repos for them.  Since we're no longer using urbit/urbit as a monorepo, that seems like the right move to me.  If someone (@Fang- maybe?) has an objection, I'm open to hearing counterarguments.


belisarius222:
image
Looks like CI depends on nix.  I'm not that familiar with the CI architecture -- @mcevoypeter or @philipcmonk, do you have any plans for how to disentangle nix from CI?  I don't know how deep that integration is.

zalberico (changes requested):
I don't think we don't want to do this in a PR like this.

- I think we already removed what was safe to remove up to this point in #6217 
- We should break this down, rather than "remove cruft" we should separate webterm first, then landscape, autoprop etc. and clarify things in well scoped issues (creating the new repos for them as we go and making sure stuff works).
- We should understand what we're removing (nix, CI) etc. and to the extent that you don't understand that we should do the work to figure it out prior to PR. Hash out those details in an issue first.

As is, I'd probably close this, CC @philipcmonk for thoughts.

belisarius222:
Breaking up the removal into smaller steps is reasonable.  I'd like to prioritize separating the non-UF-maintained pkg's, which I believe are landscape, npm, and interface.  @zalberico if I made a PR to just remove those, would that break anything?  If so, what remains to be done to make that separation viable?

philipcmonk:
I agree that we should give things a new home before deleting them.

Nix is indeed used for CI, and while it shouldn't be too hard to port that to regular bash scripts, it will take a little time.  We'll get to it next time we want to make any significant changes, but if someone wants to do it ahead of time, that would be great.  We basically just need to boot a ship and run the series of tests in that script.  I would probably just rewrite that from scratch, perhaps modeling it after some of our other repos that interact with ships in CI (landscape-apps, landscape, and lure come to mind).

zalberico:
I believe removing landscape will break stuff, but defer to @jamesacklin. We're close to being able to kill landscape (old groups) so if we can leave it until that's done (@joemfb is working on that right now) that'd be better.

I don't know about npm and interface, defer to @Fang- or @philipcmonk for an answer there. I don't think this is the most important thing to focus on right now though (for Tlon or UF really), but do agree we want to do it.

jamesacklin:
Removing the %landscape desk wholesale is a bad idea for the time being. 

Groups still needs %s3-store and %settings-store (maybe another, @arthyn would definitely know). 

We will be orphaning %contact-store when @joemfb is done.

belisarius222:
Ok, it seems like it would be a pain for Tlon to move landscape and those other packages out of this repo.  @zalberico As you said, it's not particularly urgent.  Thanks @philipcmonk and @jamesacklin for filling me in.  Closing for now.

sigilante:
Also calling out that `/mar/docket-0.hoon` (and possibly other critical files) are only represented on `landscape` and `webterm` desks.

Issue #6177, opened 2022-12-29
Log growing too fast, ames spamming 'dead' message


gusmacaulay:
I have an issue with my ship log growing unreasonably large - if left to run, it will grow ~2Gb every 24 hours.  I have asked on community help and it was suggested to look at ames.  `|ames-verb %snd` spits out an endless stream of messages of the form "ames: ~sampel-palnet: dead [[3 0] rto=120.000 rtt=910 rttvar=850 ssthresh=1 cwnd=1 num-live=1 counter=2]" 

In which there are a handful of different ship names that come up repeatedly.

I have also looked at the ~debug/ames web tool which shows a bit more detail - here are the results showing two of the different ships found with ames-verb;

![Screenshot from 2022-12-28 12-12-33](https://user-images.githubusercontent.com/1427949/209906364-4d7fb6b1-f535-4006-b0a0-0ca14d85868c.png)
![Screenshot from 2022-12-28 12-25-15](https://user-images.githubusercontent.com/1427949/209906370-aeb7901b-7384-4a61-bb57-07bc52fdd13c.png)

Those screenshots suggest perhaps failed group joins etc. may be the cause, unsubscribing from relevant groups doesn't seem to help.

lukestiles:
Thank you for the screenshots. Can you share the output of +vats as well?

gusmacaulay:
> +vats
%bitcoin
  /sys/kelvin:      [%zuse 417]
  base hash:        0vt.2qfd4.s5bt4.95ia2.o4e82.1ckgr.ltmd5.ddmsb.p9utd.ecsf8.5dqva
  %cz hash:         0vt.2qfd4.s5bt4.95ia2.o4e82.1ckgr.ltmd5.ddmsb.p9utd.ecsf8.5dqva
  app status:       running
  force on:         ~
  force off:        ~
  publishing ship:  [~ ~mister-dister-dozzod-dozzod]
  updates:          remote
  source ship:      ~samzod
  source desk:      %bitcoin
  source aeon:      7
  kids desk:        ~
  pending updates:  ~
::
%verse
  /sys/kelvin:      [%zuse 417]
  base hash:        0vl.vjd30.ivro5.nolk2.srchp.ucegb.it978.9ct4a.1b6kf.cd0al.2s23f
  %cz hash:         0vl.vjd30.ivro5.nolk2.srchp.ucegb.it978.9ct4a.1b6kf.cd0al.2s23f
  app status:       running
  force on:         ~
[...]

PR #6322, opened 2023-02-13
ames: update logs for |close-flows

yosoyubik:
Minor update for the |close-flows work to track the issue described https://github.com/urbit/urbit/issues/6065#issuecomment-1425973349, mostly related to how we log the stale flows, by checking if the stale flow is closing.


yosoyubik:
I'm seeing that several  of the stale flows from re-subscriptions that are still active (have packets in-flight) are already marked as closing, and from the few examples where I can check both sides (publisher and subscriber) the publisher doesn't have the flows anymore. 

Running |close-flows could caused that some of the stale flows we are trying to cork become orphaned (i.e. de-synced from the publisher side) and will keep retrying to send packets on a timer, possibly triggering this https://github.com/urbit/urbit/issues/6177. Until we have a way to 1) fix the issue where the publisher deletes a flow where the subscriber doesn't notice and 2) clean flows for subscribers that have orphaned flows, I would recommend not to run |close-flows.

belisarius222 (approved):

gusmacaulay:
@yosoyubik  is there any work happening to fix the remaining issues with this?

PR #6323, opened 2023-02-13
ames: add %cong task to adjust clog

yosoyubik:
Adjust the variables used for clogging less introduced in https://github.com/urbit/urbit/pull/5827


belisarius222 (approved):

tacryt-socryp:
The fact that this is configurable now is extremely sick!

PR #6324, opened 2023-02-13
eyre: revert inclusion of HttpOnly header for now

pkova:
#5918 got deployed prematurely to ~binnec, without giving app developers a change to update their version of `http-api`. This broke Groups and pretty much every frontend app. This PR removes the `HttpOnly` header from eyre for now, but keeps the new `/~/name` endpoint so people can use the old stuff and the new stuff for a while. We can add the header back later.


Fang- (approved):
This is, indeed, the most minimal reversion. The `/~/name` endpoint should indeed be fine to stay in.

PR #6332, opened 2023-02-14
%clay `$flue` scry

### LINKED ISSUE #6331: %clay : `$dome` scry
currently you can scry out part of the dome, but not all of it. Importantly, I need to be able to scry out the `$flue` to increase `%pyro` performance

tadad:
Resolves #6331 

`%pyro` installs are quite slow because we cannot utilize the host ships's `$flue` cache. By adding the ability to scry this out, we can build desks on ships in *seconds*. I have a demo of this working [here](https://www.youtube.com/watch?v=9QeUFARq1RY).

edit: I added the ability to scry out the entire dome. This also required a renaming of `$dome:clay` to `$domo` in `lull.hoon`, since now the entire `$dome` is in `lull`.


belisarius222:
This changes the return type of a scry, which would necessitate burning a Kelvin.  We might be able to get away with not burning a Kelvin if we added a new scry instead, although it's somewhat questionable because an agent wouldn't be able to specify that it needs to be run in a kernel that has this scry as a capability.

Alternatively, we could merge this into the next Kelvin release.  I think that's better, since the "domes" scry is more useful with this information in it, and duplicating the interface just to add the `$flue` cache strikes me as unclean.  I won't officially approve this PR yet, until we're sure we're burning a Kelvin, but the code looks good.  If we're not burning a Kelvin in this next release, then we will hold onto this work to be release when the next Kelvin goes out.

tadad:
yeah I agree we should burn a kelvin.

I was also thinking maybe we should just scry out the entire dome (we only leave out two pieces from this scry now, `mim` (mime cache), and `wic`(commit in waiting). Unless there is something wrong with scrying the whole thing, it will let us get rid of some duplicated code between clay and lull.

tadad:
just changed it so that you can scry out the entire `$dome` now. This moves `$dome` into `lull.hoon`, which made me have to change the name of `$dome:clay` to `$domo:clay`, which I think is better style

belisarius222 (approved):
LGTM.  We should hold off on merging until we're ready to burn the next Kelvin.

Issue #197, opened 2014-12-01
compile failure


fsteinel:
with commit ac4b62632d02e8197e39f9e7ffd484dc0051428b I get a build failure:
v/sist.c: In Funktion »_sist_rest«:
v/sist.c:816:22: Warnung: Variable »tno_l« gesetzt, aber nicht verwendet [-Wunused-but-set-variable]
   u3_noun     sev_l, tno_l, key_l, sal_l;
                      ^
    CC    v/temp.o
    CC    v/term.o
    CC    v/time.o
    CC    v/unix.o
    CC    v/save.o
    CC    v/walk.o
    CCLD  bin/vere
n/t.o: In function `u3t_samp':
urbit/n/t.c:199: undefined reference to`pthread_threadid_np'
urbit/n/t.c:200: undefined reference to `pthread_threadid_np'
collect2: Fehler: ld gab 1 als Ende-Status zurück
Makefile:371: recipe for target 'bin/vere' failed
make: **\* [bin/vere] Error 1

PR #6336, opened 2023-02-16
Add threads to run arbitrary hoon

ashelkovnykov:
Related to [#197 in urbit/vere](https://github.com/urbit/vere/issues/197).

This PR introduces threads which will be useful to any `conn.c` client. More specifically, these threads will be specifically targeted by the thin and thick clients I plan to produce for the issue linked above.

### DIFF
--- a/pkg/arvo/ted/eval.hoon
+++ b/pkg/arvo/ted/eval.hoon
@@ -1,0 +1,35 @@
+/-  spider
+/+  strandio
+=,  strand=strand:spider
+=>
+|%
++$  deps  (list path)
++$  inpt  $@(cord (pair cord deps))
+--
+^-  thread:spider
+|=  raw=vase
+=/  m  (strand ,vase)
+^-  form:m
+=+  !<(arg=(unit inpt) raw)
+?~  arg
+  (strand-fail:strand %no-input ~)
+?@  u.arg
+  ?~  u.arg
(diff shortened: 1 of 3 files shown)


belisarius222 (dismissed):
I think in the long run, we'd be better off having this take in a Ford source file that would be parsed to a `$pile` and then built according to the standard Ford build semantics, to allow reuse of Ford runes and recursive imports.  As far as I can tell, there's nothing wrong with this as it is, though.

pkg/arvo/sys/hoon.hoon
```diff
@@ -13231,9 +13231,9 @@
           ;~(pfix jump loan)
         ==
       ==
-   ::
-   ::  parses a or [a b c] or a  b  c  ==
-   ++  lynx
+    ::
```
belisarius222, line 13234: Let's move this to a separate PR that will go in next time we modify hoon.hoon

ashelkovnykov:
I plan to implement @belisarius222's feedback later in a separate PR.

I originally opened this PR as a draft because I thought I might change the threads slightly while working on the client which uses them. It seems like there's already demand for these threads, so I'll mark this as a regular PR and makes changes in a separate PR.

I'll also move the `hoon.hoon` style fixes to a separate PR so that this one doesn't force a `%base` recompile.

belisarius222 (approved):

PR #6343, opened 2023-02-18
autoprop: properly keep track of runtime version

Fang-:
Autoprop has a timer that checks whether we're on a new/different runtime periodically. When we detect we are, we re-build the props, so that they may be re-published under that new version identifier.

However, we weren't updating the cached runtime version we store in state, causing every firing of this timer to re-build the props.

Here, we correctly update state if we detect a difference, so that subsequent timers don't trigger redundant rebuilds.


sigilante (approved):
LGTM.

Interesting, I didn't know that you could modify the state this way.

PR #6344, opened 2023-02-21
dill: legible %logs

### LINKED ISSUE #6339: dill: system messages indistinguishable from userspace output
For formal output (explicitly _not_ slogs aka debug printfs), vanes send `%text`, `%talk`, and `%crud` tasks to dill. (Sometimes wrapped in `%flog` for default duct reasons, though maybe changes from #4463 removed the strict need for this.) Some userspace agents that don't provide output through the dill or drum protocols also use this pathway.

What dill does right now is always render these in the default session, through [`+fore`](https://github.com/urbit/urbit/blob/28d3e0610c6fbd73e724fe113ba104f3da538b61/pkg/arvo/sys/vane/dill.hoon#L158). The logic there just dumps the output on the bottom of the screen, ends with a newline, and tells the terminal handler to "please redraw yourself" as a cleanup step. This is coupling that #4463 reduced, but didn't fully remove.

To remove that coupling, we should consider `%text`, `%talk` and `%crud` output as "system logging" output explicitly. In doing so, we should:
- Make dill emit output from those to the runtime directly, without rendering it itself.
- Allow subscribing to dill for this system output.
- Make drum open such a subscription and render the system output into the default session, so as to maintain the status quo.

This will be our first step towards logging reform.  
The big follow-up step is separating dill sessions from the standard process' output, into a standalone "connect to my running urbit's dill" process. This lets the standard process dedicate itself entirely to outputting printfs and system output, at which point its rendering logic can be further simplified and we should be able to pipe its output straight into a log munching tool.  
Secondary to that is the long tail of places where we're using printfs and "system output" for things that shouldn't be going through that pipeline at all.

Fang-:
Adds a `%logs` task to dill, which can be used to un/subscribe to system output.

All of `%crud`, `%talk` and `%text` are now considered "system output" for this purpose. Dill no longer renders those itself. This further uncouples it from drum, dill now makes no assumptions about the default session except the fact that it exists.

Drum has been updated to subscribe to these `%logs`, and will now render those in dill's stead. The visible behavior here should not have changed, they always get rendered into the default session.  
(That's a little bit of a lie. Cases where non-default sessions would experience poke-nacks or watch-nacks (in dill's interactions with the terminal handler agent) would previously get rendered in their associated sessions. I think considering those part of the "global system output" is sane though.)

(This also opens up room for questions around error handling. If the user tries doing something in a non-default session, but it fails, will they be able to see without going back to the default session? This will probably be part of the "long tail" described in #6339.)

Drum also re-implements `|knob`, for making error outputs quieter. @belisarius222 I'm having second thoughts about this one, thinking that maybe this filtering should happen on the dill level still, because it's more "global" of a setting. But doing knob at the edges instead (as in this PR rn) ensures interested subscribers are never missing out on anything, which might be the more important consideration.  
(I'll want to rip out the dill-side knob mechanism if we agree it's no longer needed.)

This intentionally does not auto-include the default duct (which goes to the runtime) in the list of `%logs` subscribers, to avoid current runtimes crying out `kick: lost %logs`. When the time comes, runtimes can inject a `%logs` task (but will need to clean up, or remember they already did so?) or we can update dill to send these to the default ducts as well.

Fixes #6339, see that for further detail.

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1285,7 +1285,7 @@
     ==                                                  ::
   +$  dill-belt                                         ::  arvo input
     $%  belt                                            ::  client input
-        [%cru p=@tas q=(list tank)]                     ::  echo error
+        [%cru p=@tas q=(list tank)]                     ::  errmsg (deprecated)
         [%hey ~]                                        ::  refresh
         [%rez p=@ud q=@ud]                              ::  resize, cols, rows
         [%yow p=gill:gall]                              ::  connect to app
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1232,7 +1232,7 @@
         $>(%trim vane-task)                             ::  trim state
         $>(%vega vane-task)                             ::  report upgrade
         [%verb ~]                                       ::  verbose mode
-        [%knob tag=term level=?(%hush %soft %loud)]     ::  error verbosity
+        [%knob tag=term level=?(%hush %soft %loud)]     ::  deprecated removeme
         session-task                                    ::  for default session
         told                                            ::  system output
     ==                                                  ::
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -1215,26 +1215,26 @@
         [%meld ~]                                       ::  unify memory
         [%pack ~]                                       ::  compact memory
         [%trim p=@ud]                                   ::  trim kernel state
+        [%logs =told]                                   ::  system output
     ==                                                  ::
   +$  task                                              ::  in request ->$
     $~  [%vega ~]                                       ::
     $%  [%boot lit=? p=*]                               ::  weird %dill boot
         [%crop p=@ud]                                   ::  trim kernel state
-        [%crud p=@tas q=(list tank)]                    ::  print error
         [%flog p=flog]                                  ::  wrapped error
         [%heft ~]                                       ::  memory report
         $>(%init vane-task)                             ::  after gall ready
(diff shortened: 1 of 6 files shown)


pkg/arvo/sys/lull.hoon
```diff
@@ -1215,26 +1215,26 @@
...
         [%meld ~]                                       ::  unify memory
         [%pack ~]                                       ::  compact memory
         [%seat =desk]                                   ::  install desk
         [%shot ses=@tas task=session-task]              ::  task for session
-        [%talk p=(list tank)]                           ::  print tanks
-        [%text p=tape]                                  ::  print tape
         $>(%trim vane-task)                             ::  trim state
         $>(%vega vane-task)                             ::  report upgrade
         [%verb ~]                                       ::  verbose mode
         [%knob tag=term level=?(%hush %soft %loud)]     ::  error verbosity
```
belisarius222, line 1235: this is deprecated now that you've moved the verbosity setting to drum, right?

belisarius222:
> Drum also re-implements |knob, for making error outputs quieter. @belisarius222 I'm having second thoughts about this one, thinking that maybe this filtering should happen on the dill level still, because it's more "global" of a setting. But doing knob at the edges instead (as in this PR rn) ensures interested subscribers are never missing out on anything, which might be the more important consideration.

Edge-level filtering seems a little better to me in general.  The only problem I could see with that would be if there's some performance issue with the main logs getting clogged and needing to be able to reject some messages even before they reach the main log.  That's a theoretical problem, not an actual one, though, so I think we should go with your design, and then revise if needed later.

belisarius222:
Architecture questions aside, I reviewed this code and it looks good to me.

belisarius222 (approved):

PR #6350, opened 2023-02-22
ames: refactor

yosoyubik:

- Move the formal interface arms of %ames (+call, +take... to the bottom of the file
rename nested cores using the two-letter scheme (e.g. peer-core -> pe)
nest +make-sink-message (|sink) and +make-message-pump (|pump) into +peer-core, and make-packet-pump (|pack) into +message-pump, to get access to shared state (e.g. channel) in the peer-core.
- Remove gifts, and external +run-message/packet-pump arms, and handled those within the core.
- In an +abet core, (normally) don't explicitly update anything in the outer core that's not part of the core's own (there are a couple of FIXMEs where we need to refactor this)—|sink and |pump actually do modify the state of the parent core implicitly by calling each other (e.g. +pump-done abets the |sink core, and thus modifying the peer-core)
- If nested, abetting a core returns the parent core
- If there are gifts, we +abut ([gifts abet]) the core—any other name suggestions welcome

Because this is a big refactor I'm going to do a couple of passes to check that there are no regressions and also by running it on a couple moons (publisher/subscriber) to also test %nack and %cork logic—will probably adjust the %clog constants to force the resubscription behavior.

Next up:

- Remove usage of impure +abet pattern (see https://github.com/urbit/urbit/issues/6103#issuecomment-1346197840)

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2092,6 +2092,8 @@
       =|  cork-bone=(unit bone)  ::  modified by +on-kroc
       ~%  %event-core  ..$  ~
       |%
+      +|  %helpers
+      ::
       ++  event-core  .
       ++  abet  [(flop moves) ames-state]
       ++  emit  |=(=move event-core(moves [move moves]))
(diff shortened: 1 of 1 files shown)


yosoyubik:
wrong issue in the branch name, closing in favor of https://github.com/urbit/urbit/pull/6351

PR #6351, opened 2023-02-22
ames: refactor

yosoyubik:

- Move the formal interface arms of `%ames` (+call, +take... to the bottom of the file
rename nested cores using the two-letter scheme (e.g. peer-core -> pe)
nest +make-sink-message (`|sink`) and +make-message-pump (`|pump`) into +peer-core, and make-packet-pump (`|pack`) into +message-pump, to get access to shared state (e.g. channel) in the peer-core.
- Remove gifts, and external +run-message/packet-pump arms, and handled those within the core.
- In an +abet core, (normally) don't explicitly update anything in the outer core that's not part of the core's own (there are a couple of FIXMEs where we need to refactor this)—`|sink` and `|pump` actually do modify the state of the parent core implicitly by calling each other (e.g. +pump-done abets the `|sink` core, and thus modifying the peer-core)
- If nested, abetting a core returns the parent core
- If there are gifts, we `+abut` (`[gifts abet]`) the core—any other name suggestions welcome

Because this is a big refactor I'm going to do a couple of passes to check that there are no regressions and also by running it on a couple moons (publisher/subscriber) to also test `%nack` and `%cork` logic—will probably adjust the %clog constants to force the resubscription behavior.

Next up:

- Remove usage of impure +abet pattern (see https://github.com/urbit/urbit/issues/6103#issuecomment-1346197840)


yosoyubik:
Currently testing this on `~locsyr-rabrud-norsyr-torryn`, will report here if any comes up.

yosoyubik:
So far I haven't seen any issues with my tests (a publisher and a subscriber, both running this refactor; the publisher hosting a group with custom %clog constants; the subscribers getting %kicked and the flows %corked) but I'd like to test a couple more things so this resembles more a real livenet ship.

I'll also do another pass at checking all the cores that were refactored (removal of the custom gift processing specially since this is the part that has changed the most).

- There's a group at `~locsyr-rabrud-norsyr-torryn/el-hoyo`. I've tuned the %clog constants for just one message pending, so it'll kick subscriptions more often, which will then send %corks to close those flows. I'll encourage people to join so we get some activity going.

- Installing a couple of apps that use subscriptions (e.g. %rumors).

- Testing %nacks for %watches by crashing on initial subscription

yosoyubik:
Currently running this on ~norsyr-torryn to catch any issues on livenet ships (this is also hosting a couple of groups). All good so far but I'm debugging https://github.com/urbit/urbit/issues/6050 and a variant of https://github.com/urbit/urbit/pull/6364 (for boons) so will push fixes for them when ready.

(Note: this PR also includes this: https://github.com/urbit/urbit/pull/6383)

yosoyubik:
Update: ~norsyr-torryn has been running this since Monday and I haven't had any issues with DMs in Talk, sending/receiving message updates in Groups... etc.

I've also tested that flows are corked properly between different ships, and lastly I deployed the refactor to ~deg and booted a new L2 planet under ~lardeg that had no issues with peer discovery or communication with other ships. At this point I haven't seen any explicit problems so I'd like to mark this as ready to be deployed to ~binnec—if we see something not working we can revert ~binnec's %ames to its previous state and I'll investigate further.

PR #6354, opened 2023-02-23
hoon: fix bug in STL 'prn'

ashelkovnykov:
Since `shim` is inclusive and the max value of an ASCII byte is `0xFF`, input to field `mos` in `shim` for `prn` should be 255.

Also fixed some whitespace errors.


joemfb (approved):
good catch, this will do

ashelkovnykov:
@belisarius222 @jalehman Ready for merge

ashelkovnykov:
@belisarius222 I found it while working on #5566, but I've never hit it. Like @Fang- said, it's almost impossible for someone to be hitting this: I think they would actively have to try.

Issue #257, opened 2015-02-25
:tree navigation widget is too tall


chc4:
If you load any of the library reference sections, for example http://localhost:8080/gen/docs/tree/doc/hoon/library/0, the navigation sidebar extends to the bottom of the screen, and extends below and off the screen when mousing over that section. The navigation widget should probably support either scrolling, or cut off the later elements and have a button to scroll all the contents of the tree.

galenwp:
Also a good one. This is pretty easy to change via CSS. I'll take care of it.

Thanks!

galenwp:
I think this is fixed on `factor`, but I'll be sure to check.

PR #6355, opened 2023-02-23
hoon: minor improvement to +mod and +dvr

ashelkovnykov:
Unify logic for `div` and `mod` in `dvr`, instead of splitting the logic and repeating the same calculation twice on a `dvr` call.

Also add default sample to `dvr`, mirroring `div` and `mod`.

Tested changes successfully against latest code from `develop` (both `urbit/urbit` and `urbit/vere`).

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -50,12 +50,7 @@
   |:  [a=`@`1 b=`@`1]
   ::  quotient
   ^-  @
-  ~_  leaf+"divide-by-zero"
-  ?<  =(0 b)
-  =+  c=0
-  |-
-  ?:  (lth a b)  c
-  $(a (sub a b), c +(c))
+  -:(dvr a b)
 ::
 ++  dvr
   ~/  %dvr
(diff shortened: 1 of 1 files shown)


joemfb:
With this change, the `+mod` jet should also be updated to use `u3m_error()` as the other two do.

ashelkovnykov:
@joemfb Good catch - added a companion PR to `urbit/vere`: [#257](https://github.com/urbit/vere/pull/257). **They must be released together**.

philipcmonk:
Is this a semantic change? Is it possible that the old code would give a different result than the new code? If so, it must be released in a hoon kelvin update so the jet never fires with the wrong nock

ashelkovnykov:
@philipcmonk The only difference should be the one Joe pointed out: the new code gives a different result in the case of "divide by zero" error.

Old code = bail: exit
New code = `u3t_mean("divide-by-zero")`, then bail: exit

What's the protocol for explicitly marking this PR as requiring a hoon kelvin update? Should I dump the kelvin version in this PR, and then all other such changes nicely merge together since they also dump the kelvin version?

joemfb (approved):
Will it's technically true that this code must be released at the same time as the paired jet changes to avoid a mismatch, it's an utterly trivial mismatch (the error string in the trace when dividing by zero). The livenet version of this code already contains such a mismatch -- the `+dvr` jet includes the error string, while the hoon does not. And stack traces are already subject to such mismatches anytime the set of actual jets changes relative to execution history (ie, on replay) -- a problem for another day. I don't think we have to worry about kelvins for this PR specifically.

this will do

ashelkovnykov:
@belisarius222 @jalehman Ready for merge

PR #6377, opened 2023-03-07
Fix stuck %eth-watcher due to get-block-by-number zero-padded number RPC failure.

### LINKED ISSUE #6376: eth-watcher appears to be broken
My ship and many others are seeing this message in the dojo:

```
eth-watcher failed; will retry
%rpc-result-incomplete-batch
[ %a
    p
  [   i
    [ %o
        p
      [ n=[p='jsonrpc' q=[%s p='2.0']]
        l=~
          r
        [   n
          [ p='error'
              q
            [%o p=[n=[p='message' q=[%s p='invalid argument 0: hex number with leading zero digits']] l=~ r=[n=[p='code' q=[%n p=~.-32602]] l={} r={}]]]
          ]
          l={}
          r={[p='id' q=[%s p='block by number']]}
        ]
      ]
    ]
    t=~
  ]
]
'revoke perms for'
'revoke perms for'
eth-watcher failed; will retry
%rpc-result-incomplete-batch
[ %a
    p
  [   i
    [ %o
        p
      [ n=[p='jsonrpc' q=[%s p='2.0']]
        l=~
          r
        [   n
          [ p='error'
              q
            [%o p=[n=[p='message' q=[%s p='invalid argument 0: hex number with leading zero digits']] l=~ r=[n=[p='code' q=[%n p=~.-32602]] l={} r={}]]]
          ]
          l={}
          r={[p='id' q=[%s p='block by number']]}
        ]
      ]
    ]
    t=~
  ]
]
eth-watcher failed; will retry
%rpc-result-incomplete-batch
[ %a
    p
  [   i
    [ %o
        p
      [ n=[p='jsonrpc' q=[%s p='2.0']]
        l=~
          r
        [   n
          [ p='error'
              q
            [%o p=[n=[p='message' q=[%s p='invalid argument 0: hex number with leading zero digits']] l=~ r=[n=[p='code' q=[%n p=~.-32602]] l={} r={}]]]
          ]
          l={}
          r={[p='id' q=[%s p='block by number']]}
        ]
      ]
    ]
    t=~
  ]
]
```

justinlynn:
### Description

Resolves #6376.

Adds ethereum-types:render-hex-minimal and and ethereum-types:hex-bytes-minimal. Modifies RPC for get-block-by-number to render block number using these.

Fixes 16-bit alignment in ethereum block number serialisation which results in stuck eth-watcher on livenet.

### Commentary

I'm new to hoon so I'm sure there's a more elegant way to do this. However, this addressed the issue locally and on the test ships I have available. This is a minimal patch which should not impact other users of num-to-hex or render-hex-bytes until further discussion can be had.

Urbit ID: ~macpem-micsec

### DIFF
--- a/pkg/base-dev/lib/ethereum.hoon
+++ b/pkg/base-dev/lib/ethereum.hoon
@@ -946,8 +946,6 @@
   |=  n=@
   ^-  tape
   %-  prefix-hex
-  ?:  =(0 n)
-    "0"
   ((x-co:co 1) n)
 ::
 ++  address-to-hex
--- a/pkg/base-dev/lib/ethereum.hoon
+++ b/pkg/base-dev/lib/ethereum.hoon
@@ -733,7 +733,7 @@
     ::
         %eth-get-block-by-number
       :-  'eth_getBlockByNumber'
-      :~  (tape (num-to-hex bon.req))
+      :~  (tape (num-to-hex-minimal bon.req))
           b+txs.req
       ==
     ::
@@ -942,6 +942,15 @@
   %-  render-hex-bytes
   (as-octs:mimes:html n)
 ::
+++  num-to-hex-minimal
+  |=  n=@
+  ^-  tape
+  %-  prefix-hex
+  ?:  =(0 n)
+    "0"
+  %-  render-hex-bytes-minimal
+  (as-octs:mimes:html n)
+::
 ++  address-to-hex
   |=  a=address
   ^-  tape
@@ -979,6 +988,12 @@
   |=  a=octs
   ^-  tape
   ((x-co:co (mul 2 p.a)) q.a)
+::
+++  render-hex-bytes-minimal
+  :: atom to string of hex bytes without 0x prefix and dots, minimal representation (no padding).
+  |=  a=octs
+  ^-  tape
+  ((x-co:co 1) q.a)
 ::
 ++  pad-to-multiple
   |=  [wat=tape mof=@ud wer=?(%left %right)]
--- a/pkg/base-dev/lib/ethereum.hoon
+++ b/pkg/base-dev/lib/ethereum.hoon
@@ -948,8 +948,7 @@
   %-  prefix-hex
   ?:  =(0 n)
     "0"
-  %-  render-hex-bytes-minimal
-  (as-octs:mimes:html n)
+  ((x-co:co 1) n)
 ::
 ++  address-to-hex
   |=  a=address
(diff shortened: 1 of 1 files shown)


pkg/base-dev/lib/ethereum.hoon
```diff
@@ -942,6 +942,15 @@
...
   (as-octs:mimes:html n)
 ::
+++  num-to-hex-minimal
+  |=  n=@
+  ^-  tape
+  %-  prefix-hex
+  ?:  =(0 n)
+    "0"
+  %-  render-hex-bytes-minimal
+  (as-octs:mimes:html n)
```
Fang-, line 952: You really don't need this as octs. It seems fine to me to just call `(x-co:co 1)` here directly. That should also take care of the zero case.

Fang- (commented):
The code here could be simpler, but this looks like a quick fix for the problem we're seeing, yes.

We should consider releasing this as a hotfix, considering it's preventing people from loading in newer azimuth state, which breaks networking for newer or recently-rekeyed ships.

pkg/base-dev/lib/ethereum.hoon
justinlynn: Thanks @Fang- as I said before I'm quite new to hoon and urbit, so I'd happily take a direct patch in hoon, just to be sure I understand.

Fang-:
Once the immediate problem is resolved, we'll want to do a pass over the eth rpc stuff and make sure we're omitting leading zeroes in all the places where that's considered invalid. (Which may be all places? But we should confirm that systematically.)

pkg/base-dev/lib/ethereum.hoon
Fang-, line 952: I mean to change it like so:

```suggestion
  ((x-co:co 1) n)
```

And then deleting the `+render-hex-bytes-minimal` arm you added.

arthyn: I think @Fang- is saying this: 

```suggestion
  ?:  =(0 n)
    "0"
  ((x-co:co 1) n)
```

and then you don't need the helper func below

justinlynn: done.

justinlynn: Done.

pkova (approved):

justinlynn (commented):
Thanks for this. Apologies for missing it in your suggestion.

belisarius222 (approved):

PR #6381, opened 2023-03-09
hoon: make sane memory-efficient for large atoms

### LINKED ISSUE #6380: Performance Issue: `sane` uses too much memory for large atoms
I discovered that `((sane %t) data)` would cause my ship to run out of loom when `data` was a 160kb cord. This is unnecessary.
@philipcmonk theorized that `sane` was creating a new slightly smaller atom for each step in the traversal, which added up quickly, and that this could be solved by incrementing a cursor and using `cut` to get each byte.
He was right, and following his suggestion I have modified `sane` to be dramatically more memory efficient. On my dev moon, the new limit is a bit over 4mb.

johnhyde:
### Description

Resolves #6380.

I updated the section of `sane` which checks for `%t` validity to iterate through bytes with an incrementing cursor. This greatly reduces memory load.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -4221,15 +4221,14 @@
   =+  cur=(cut 3 [inx 1] b)
   ?:  &((lth cur 32) !=(10 cur))  |
   =+  tef=(teff cur)
-  ?&  ?|
-          =(1 tef)
+  ?&  ?|  =(1 tef)
           =+  i=1
-        |-  ?|
-            =(i tef)
-            ?&  (gte (cut 3 [(add i inx) 1] b) 128)
-                $(i +(i))
-        ==  ==
-      ==
+          |-  ^-  ?
+          ?|
+              =(i tef)
+              ?&  (gte (cut 3 [(add i inx) 1] b) 128)
+                  $(i +(i))
+      ==  ==  ==
       $(inx +(inx))
   ==
 ::
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -4217,12 +4217,21 @@
         $(inx +(inx))
     ==
   |-  ^-  ?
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -4217,12 +4217,20 @@
...
+  ?&  ?|
+          =(1 tef)
+          =+  i=1
+        |-  ?|
+            =(i tef)
+            ?&  (gte (cut 3 [(add i inx) 1] b) 128)
+                $(i +(i))
+        ==  ==
+      ==
+      $(inx +(inx))
```
joemfb, line 4233: The indentation here is not idiomatic, it should look something like:

```
=+  tef=(teff cur)
?&  ?|  =(1 tef)
        =+  i=1
        |-  ^-  ?
        ?|  =(i tef)
            ?&  (gte (cut 3 [(add i inx) 1] b) 128)
                $(i +(i))
    ==  ==  ==
  $(inx +(inx))
==
```

joemfb (changes requested):
This is a good catch! `+rsh` loops are convenient to write, but horribly inefficient.

I have a style nit:

pkg/arvo/sys/hoon.hoon
johnhyde: Ok, thanks. I was getting mixed up with the indent style for ?- and ?+. But shouldn't `$(inx +(inx))` be at the same indentation level as the `==  ==  ==`?

johnhyde:
Can we squash these commits into a single nice message on merge, or do I need to try to rewrite history to abide by the commit message guidelines?

pkg/arvo/sys/hoon.hoon
joemfb: Yeah, you're right.

joemfb (approved):
PRs are usually merged instead of squashed, so you can use rebase to squash these commits down to one.

this will do

Issue #6154, opened 2022-12-13
|uninstall isn't |suspend-ing agents


sigilante:
The documentation for `|uninstall` states,

> The specified desk will be retained in Clay, but its agents will all be stopped and have their states archived like `|suspend`.

A new ship with a clean desk can `|install` a desk, but if the desk is `|uninstall`ed and `|install`ed a second time, the second `|install` fails because the desk isn't registered as uninstalled.

Similarly, after the desk is `|uninstall`ed, the agent continues to run and has not been `|suspend`ed.

`|suspend` works on its own as expected.

```
> |install our %journal
>=
>   [%0 journal={} log={}]
> :journal +dbug
>=
desk not installed: %journal
> |uninstall %journal
>=
>   [%0 journal={} log={}]
> :journal +dbug
>=
> |install our %journal
>=
>   [%0 journal={} log={}]
> :journal +dbug
>=
gall: stopping %journal
> |suspend %journal
>=
gall: not running %journal yet, got %poke
> :journal +dbug
docket: fetching %ames glob for %journal desk
gall: reviving %journal
>   [%0 journal={} log={}]
> |install our %journal
>=
>=
docket: thread failed;
%watch-ack-fail
watch-ack
/app/docket/hoon:<[162 33].[162 61]>
[%path i=%~.glob t=[i=~.journal t=[i=~.0v0 t=~]]]
/app/docket/hoon:<[162 20].[162 62]>
/app/docket/hoon:<[162 7].[169 53]>
/app/docket/hoon:<[161 7].[169 53]>
/app/docket/hoon:<[160 7].[169 53]>
/app/docket/hoon:<[150 5].[170 7]>
/app/docket/hoon:<[149 3].[171 15]>
/app/docket/hoon:<[148 3].[171 15]>
/sys/vane/gall/hoon:<[1.227 9].[1.227 37]>
>   [%0 journal={} log={}]
> :journal +dbug
```

---

To reproduce:

1. Create a new desk with `|merge`.
2. Clean it up with a `|commit`.
3. `|install` it locally.
4. `:agent +dbug` poke it to confirm it runs.
5. `|uninstall` it.
6. `:agent +dbug` poke it to confirm it is still running (wasn't suspended).

sigilante:
The plot thickens:  I don't know how path-dependent this is, but now I'm also getting this interesting result:

```
> |install our %zettelkasten
>=
>   [%0 journal={} log={}]
> :zettelkasten +dbug
>=
[...]

PR #6382, opened 2023-03-10
kiln: fix uninstall

### LINKED ISSUE #6242: |uninstall doesn't also suspend the app
- |uninstall %bitcoin
- Observe the desk is uninstalled
- Run +vat %bitcoin
- Observe the desk is still marked running
- You need to then run |suspend

This is bad because it means kelvin upgrades will block on the "uninstalled" bitcoin that's still marked running.

MarcusMiguel:
Resolves #6242 / #6154

Handles the `|uninstall` of local desks


philipcmonk (approved):
Looks good!

PR #6388, opened 2023-03-14
gall: bind scry paths from userspace

belisarius222:
This adds three tasks to `+task:agent:gall` so that userspace agents can put data directly into the scry namespace and manage its lifecycle:
- `%grow` to bind a path to a value
- `%tomb` to tombstone a path, i.e. replace the value with its hash
- `%cull` to delete all scry bindings at paths before and including the specified case

Gall ensures referential transparency.  Userspace code has no ability to re-bind a path to a different value once it's been bound.  

I have tested the `%grow` task and Gall's scry handling for the bound value.  ~~I still need to test `%tomb` and `%cull`.  Once I have done that, I will take this out of draft mode.~~

Some TODOs:
- [x] implement scry for hash of value
- [x] implement ~`%y`-style~ `%t`-style listing of scry paths, to support exploration interfaces (or stick them in the bowl)
- [ ] ~add some kind of permissioning scheme -- it's all public to other agents for now~
- [x] test %grow
- [x] test %tomb
- [x] test %cull
- [x] handle scry bindings immediately after agent runs, instead of Gall passing an Arvo move to itself
- [x] support %a scries at the current date (to support pine)
- [x] enforce monotonicity
- [x] when an agent is nuked, delete all the data and store a highwater mark so that you know at which case to restart
- [x] reserve the namespace of `[%$ *]` for the vane and change `%a` to `%x` in that namespace
- [x] smear the `@da`s if several values are bound to the same path within the same event (and verify that they don't overlap with earlier events)

and some questions to resolve:

- [x] guarantee cases are bound monotonically? ("up only")
~NO~ YES because otherwise we can't support `%a` care at current time
- [x] enforce that cases can't be skipped?
NO
- [ ] allow non-numeric cases?
- [x] bump a Kelvin when adding this?
YES
- [x] how to reinflate tombstones when we don't include the `$case` in `%grow`? do we even care to do it for now?
NOT AT THE MOMENT
- [x] store a `@da` together with each aeon?
YES
- [x] emit an effect to Vere for each newly bound path, and re-emit them all on `%born`?
NOT FOR NOW, PREMATURE

Long-term, we'll want a typed interface here instead of untyped, but I think untyped is better for an initial low-level prototype.

There is also [an example app that uses these new notes](https://gist.github.com/belisarius222/4224f4f25c90e3d14d7821a708b02c79).  To run that, you will also need a desk.bill in the desk so it can run.


pkg/arvo/sys/lull.hoon
```diff
@@ -1888,6 +1883,10 @@
       $%  [%agent [=ship name=term] =task]
           [%arvo note-arvo]
           [%pyre =tang]
+      ::
+          [%grow =case =spur =noun]
```
joemfb, line 1887: I think it'll be simpler to remove `=case` here, and just let %gall manage contiguous, monotonic aeons.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -94,7 +95,15 @@
       agent=(each agent vase)
       =beak
       marks=(map duct mark)
+      sky=(map spur path-state)
```
joemfb, line 98: It'd be really nice to have this in the bowl.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -952,14 +1041,19 @@
     ::    We accept %huck to "fake" being a message to a ship but
     ::    actually send it to a vane.
     ::
-    +$  neet
-      $%  neat
+    +$  carp  $+  carp  (wind neet gift:agent)
+    +$  neet  $+  neet
+      $<  %grow
+      $<  %tomb
+      $<  %cull
```
joemfb, line 1048: Style nit, this could just be `$<  ?(%grow %tomb %cull)`

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1906,6 +2053,21 @@
...
+    ?~  ski  ~
+    =/  res=(unit (each noun @uvI))  
+      ?+  -.r.bem  ~
+        %ud  (get:on-path fan.u.ski p.r.bem)
+        %da  ?>(=(p.r.bem now) (bind (ram:on-path fan.u.ski) tail))
+      ==
+    ?.  ?=([~ %& *] res)  ~
+    ``[%noun %noun p.u.res]
+  ::  TODO new care to ask for hash of a sky binding
+  ::
```
joemfb, line 2070: It'd be great to use %z for this hash, and to also support %y for `arch`

joemfb (changes requested):

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1906,6 +2053,21 @@
...
+  ?:  ?=(%a care)
+    ?.  =(p.bem our)  ~
+    =/  yok  (~(get by yokes.state) q.bem)
+    ?~  yok  ~
+    =/  ski  (~(get by sky.u.yok) s.bem)
+    ?~  ski  ~
+    =/  res=(unit (each noun @uvI))  
+      ?+  -.r.bem  ~
+        %ud  (get:on-path fan.u.ski p.r.bem)
+        %da  ?>(=(p.r.bem now) (bind (ram:on-path fan.u.ski) tail))
```
joemfb, line 2065: This shouldn't assert, it should block (`~`) at dates other than `now`. (Alternately, we could store the dates on the entries and support other `%da` cases.)

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -94,7 +95,15 @@
       agent=(each agent vase)
       =beak
       marks=(map duct mark)
+      sky=(map spur path-state)
   ==
+::
++$  path-state
+  $:  bob=(unit @ud)
+      fan=((mop @ud (each noun @uvI)) lte)
```
joemfb, line 103: I think it'd be better if this were `page` instead of `noun`.

pkg/arvo/sys/vane/gall.hoon
wicrum-wicrun: %y is coming!

pkg/arvo/sys/lull.hoon
wicrum-wicrun: I changed this and put the question of reinflation as a question to resolve in the todo.

pkg/arvo/sys/vane/gall.hoon
wicrum-wicrun: %y is interesting because the aeons of different paths may not be in sync with each other. Say that we have these paths bound:
```
/==/0/foo
/==/1/foo
/==/0/foo/bar
```
What should happen if we scry for `gy//==/1/foo`? Should we get `[[~ <foo-hash>] ~]` or `[[~ <foo-hash>] (malt [~.bar ~]~)]`?

An alternative solution would be to force all aeons within the same agent to be in sync, and assume that data bound at one aeon is also bound for all other aeons until the next bound aeon on that same path. Feels icky to layer that sort of interpretation on top of the bindings though.

joemfb:
Wrt `@da` for each aeon, I think something like `fan=((mop @ud (pair @da (each noun @uvI))) lte)` is the right approach. `(mop cass:clay (each noun @uvI))` allows for unnecessary variation between date and aeon.

Note that we have to preserve the referential transparency of %da-cased paths, we need to either a) reject successive growths at a path within a single event, or b) smear the `@da`.

pkg/arvo/sys/vane/gall.hoon
wicrum-wicrun: Sounds good, `%t` is implemented!

pkg/arvo/sys/vane/gall.hoon
wicrum-wicrun: Done! I had to move `+mop` to lull which is certainly debatable (maybe hoon is better even though it's not necessary?) but I figure we can discuss that once we go through the code.

pkg/arvo/sys/vane/gall.hoon
wicrum-wicrun: Done, though I'm a bit unclear on whether we also need a mar/page file? At the moment the `%a` scry resolution looks like this:

```hoon
=/  res=(unit (each page @uvI))   ...
?.  ?=([~ %& *] res)  ~
``[%page !>(p.u.res)]  ::TODO mar/page?
```

Or should the last line be replaced with this instead?

```hoon
``p.u.res(q !>(q.p.u.res))
```

wicrum-wicrun: The latter was decided and has been pushed.

belisarius222:
@wicrum-wicrun To avoid scope creep, I'd prefer to defer implementing permissions until after this initial release.  In particular, there are a couple things we should be careful about with permissions that I'd like to focus on separately, later:
- it would be better not to require userspace devs to materialize the full set of ships who can see some path
- we should make sure data intended to be private doesn't leak through the remote scry network protocol
- we should make sure that whatever encryption scheme we use for remote scry actually works end-to-end, in the sense that a client can retrieve an encryption key and use it to encrypt a request and decrypt its result, without gaining access to any paths that it shouldn't have access to

For reference, encrypted scry was roughly half-written about a year ago:

https://github.com/urbit/urbit/compare/ted/encrypted-scry

wicrum-wicrun:
Ok, reverting that commit!

wicrum-wicrun:
This seems done to me. I and José will merge this into the Remote Scry branch and start testing everything.

pkg/arvo/sys/lull.hoon
```diff
@@ -19,6 +19,412 @@
...
+::
+::
+++  ordered-map  on
+::  +on: treap with user-specified horizontal order, ordered-map
+::
+::  WARNING: ordered-map will not work properly if two keys can be
+::  unequal under noun equality but equal via the compare gate
+::
+++  on
+  ~/  %on
```
joemfb, line 42: This is the kind-of hacky pattern we've historically used in %zuse:

```
~%  %on  ..part  ~
```

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1901,14 +2047,77 @@
...
+    =/  res=(unit (pair @da (each noun @uvI)))
+      ?+  -.r.bem  ~
+        %ud  (get:on-path fan.u.ski p.r.bem)
+        %da  ?.(=(p.r.bem now) ~ (bind (ram:on-path fan.u.ski) tail))
+      ==
+    ?+  res  ~
+      [~ @ %| *]  ``noun/!>(p.q.u.res)
+      [~ @ %& *]  ``noun/!>(`@uvI`(shax (jam p.q.u.res)))
+    ==
+  ::
```
joemfb, line 2120: This whole block should move to the end of `+scry`, so that `|mass`, `%gu`, &c continue to work.

wicrum-wicrun:
I'm closing this since it's been merged into #6406.

pkg/arvo/sys/vane/gall.hoon
wicrum-wicrun: Fixed in the other PR

pkg/arvo/sys/lull.hoon
wicrum-wicrun: Fixed in the other PR

PR #6394, opened 2023-03-15
lib, gen: remove folder whitelist from lib/pill, add option to remove folders 

### LINKED ISSUE #6295: lib/pill should copy all files from a desk into a pill
As it stands, it has a whitelist of directories, but this doesn't mesh well with app desks, which might have files in locations that the `%base` desk hadn't thought about.  It might be good for `+solid` and other pill generators to take an optional argument to remove folders, so app devs can remove /tests and similar directories before deployment, but that shouldn't be hard-coded in `%base`.

See also #6031

silnem2:
Replaces the whitelist of folders in lib/pill with an option to remove folders and adds optional argument `=exc (list spur)` to `+solid` and `+brass`. Resolves #6295 / #6031.


pkg/base-dev/lib/pill.hoon
```diff
@@ -154,8 +153,9 @@
 ++  solid
   ::  sys: root path to boot system, `/~me/[desk]/now/sys`
   ::  dez: secondary desks and their root paths
+  ::  exc: list of desk folders to exlclude
```
belisarius222, line 156: spelling: exlclude

belisarius222 (changes requested):
This looks good, but let's pull out the filtering function for cleanliness.

pkg/base-dev/lib/pill.hoon
silnem2: lmao whoops

pkg/arvo/app/aqua.hoon
```diff
@@ -485,11 +485,10 @@
     =/  =desk  +.val
     =.  userspace-ova.pil
       ::  take all files from a userspace desk
-      =/  all-dirs=(list path)  ~[/]
       :_  ~
       %-  unix-event:pill-lib
-      %+  %*(. file-ovum:pill-lib directories all-dirs)
-      desk  /(scot %p our.hid)/[desk]/(scot %da now.hid)
+      %-  file-ovum:pill-lib  :+  desk
```
belisarius222, line 490: Using a `:+` here isn't very idiomatic.  It should be either `%^` or something like this:
```hoon
%-  file-ovum:pill-lib
[desk /(scot %p our.hid)/[desk]/(scot %da now.hid) ~[/sys]]
```
I think I'd prefer this over the `%^` since the entire argument can be expressed on a single line.

pkg/base-dev/lib/pill.hoon
```diff
@@ -32,25 +32,22 @@
       [/sys/hoon hoon/hoon]
       [/sys/arvo hoon/arvo]
   ==
+::  +dirs: filter for userspace directories to include
+::     
+++  dirs
+  |=  [bas=path exc=(list spur)]
```
belisarius222, line 38: Please add a return type annotation here to make this easier to skim.

belisarius222 (changes requested):
A couple minor nits, but then this will be good to merge.

pkg/arvo/app/aqua.hoon
silnem2: Is excluding /sys correct? I _think_ this change actually represents a slight improvement over the prior implementation, but I have never used `%aqua` so I'm not 100% sure.

belisarius222 (approved):

PR #6396, opened 2023-03-16
eyre: add support for noun-based channels

xiphiness:
Closes #5877. js-http-api related changes moved to https://github.com/urbit/js-http-api/pull/4


Fang-:
As @xiphiness and I discussed out of band, these are the eyre-side changes from #5877 forward-ported to be on top of the latest arvo. See that PR for further discussion/context.

Some of the questions raised therein remain open. Namely:

- [ ] Connecting to the same channel across multiple modes is currently allowed. But it's untested, and unclear whether this is desirable or not. PUT requests are only supported in the channel's "original" mode. Whatever we end up doing, these two should be consistent. Leaning towards disallowing multi-mode-ing.

Fang-:
> Why are PUT requests limited to the channel's original mode? 

This is actually no longer the case as of e11ac8a, and was just an artifact I overlooked in copy-pasting the remaining tasks. See also:

https://github.com/urbit/urbit/blob/e11ac8a1c83245c27c0b2d0c52899569b2223692/pkg/arvo/sys/vane/eyre.hoon#L1440-L1445

belisarius222:
Mixed-mode channels seems fine to me.  It's just a data encoding, right?  I'm inclined to approve this.

jalehman:
@xiphiness we've now got some merge conflicts upstream in `eyre` — would you mind resolving those? CI is now set up to run this correctly on the `next/kelvin/413` branch, so we should get proper test results then.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -198,13 +201,44 @@
...
   ?:  (lte u.sus ack)  0
   (sub u.sus ack)
+::  +find-channel-mode: deduce requested mode from headers
+::
+++  find-channel-mode
+  |=  [met=method:http hes=header-list:http]
+  ^-  ?(%json %jam)
+  =+  ^-  [hed=@t jam=@t]
+    ?:  ?=(%'GET' met)  ['x-channel-format' 'jam']
+    ['content-type' 'application/octet-stream']
```
joemfb, line 211: This should be `application/x-jam`, here and in mar/jam.

PR #6399, opened 2023-03-17
413k

belisarius222:
Draft PR for the Kelvin release scheduled for April 31.

So far, includes https://github.com/urbit/urbit/pull/5887.


belisarius222:
We should address @joemfb's comment on the typed paths PR before merging this:

> Also, I continue to suspect that iota is too strict -- I'm not sure what we gain from it over coin or dime. (Notably, not using coin restricts us from match on %many and %blob, both of where are valid in paths. %many is currently used by arvo to verbosely encode vane and care.

In particular, the fact that Arvo encodes a cell into a single path element suggests that might be worth standardizing.  I have some reticence about this, because the intent of a path is to encode a list of atoms, not arbitrary noun shapes.  That being said, if that will be a widespread use-case, standardizing it into the type of the path itself could be better than forcing each use-case to reinvent its own encoding.

I'm inclined to think we should merge this as-is, see how many use-cases pop up for encoding cells into paths, and see how awkward it is to convert those cells into and out of the paths -- if Arvo can keep the same encoding and it only needs one or two extra lines to convert in or out, then maybe it's not so bad.  (Of course, Arvo will continue to use old-style paths until we rewrite its wire handling to use typed paths, which will also require a nontrivial migration, so it won't come up immediately).

belisarius222:
yes, will fix

belisarius222:
- Merge conflict fixed
- Kelvins bumped in all needed places

Still need to get this PR to use the corresponding urbit/vere branch in CI.  @matthew-levan do you know how to do that?

matthew-levan:
> * Merge conflict fixed
> * Kelvins bumped in all needed places
> 
> Still need to get this PR to use the corresponding urbit/vere branch in CI. @matthew-levan do you know how to do that?

On it, see #6404, will have a PR up to resolve it soon.

Fang-:
idk who needs to hear this or where it's tracked, but we should be sure to include #6474 in the "developer instructions" part of the release notes for this.

PR #6402, opened 2023-03-19
Clay: support links to scry paths

belisarius222:
This PR adds experimental support in Clay for files of a new mark `%link`, which refer to (and include the contents of) a file at a different scry path.  Its desirability should be discussed before merge.  It has been confirmed to work in `-build-file`, at least.

The main use case of this is to increase transparency in dependency management (see https://twitter.com/wispem_wantex/status/1637488407369641984).  A new generator, `+link`, included in this PR, helps to create a link.  This provenance is useful for a few reasons:
- This facilitates code signing.  Even if I retrieved some files in a desk from a relay, I can check the signature from the original publisher.
- A thread could work like `npm i` to copy files from another desk into mine as links.
- This could become one of the layers in a system that allows for specifying behavior (such as a type definition, protocol definition, or library) by absolute scry path.

An alternative to this system would allow a Ford rune to import a fully qualified scry path -- this would likely require changing the datatype for a desk to somehow include the data for files outside its usual beak.  @liam-fitzgerald makes a good point that in practice, userspace devs will want to import an external dependency once within a desk, then re-import the local pointer to that external file in multiple other files.  This suggests that the way this works now is better than importing a file directly in Hoon.  Apparently Deno uses this double-pointer convention.

For now, the signature field is set to `0v0`.  Once remote scry is standard for transmitting Clay data, and probably once we add signatures to Clay's responses over Ames too, we can start enforcing signature validity.

TODOs:
- [ ] Add a scry `$care` to ask for the original scry path
- [ ] Dereference the path in other scrys: `%r`, `%t`,`%x`, `%y`, `%z`
- [ ] Add some kind of support for linking in whole folders
- [ ] have `|cp` preserve provenance?
- [ ] ensure that a `$link` never points to another `$link`


vvisigoth:
Hell yeah

Issue #5788, opened 2022-05-24
Test Scry OTA on Live Net


zalberico:
Before we can ship the Remote Scry OTA work we need to test it.

We've tested a lot already, but here are a few things that remain.

- Simulate a ship behind a NAT/firewall and test cases of publisher/subscriber behind and not behind a firewall
- Torture test with a script that does OTAs repeatedly and let that run (have a publisher and continually make commits on that publisher a bunch of times in a row to push out a lot of OTAs) - we can do this when we're running on the Tlon patps for a real test.
- On Test Net start with an old galaxy and then upgrade the galaxy (we'll know when we go to deploy it to a moon without deploying to that moon's galaxy first we'll have some idea of what happens)

zalberico:
Something were broken because of recent changes (mainly around L2)
- Scripts were patched to be more reliable
- Set up a test net branch off of the content dist PR (PR for this is #5725) that includes an updated Ropsten snapshot for L2
- Added a new galaxy with a star and planets under it
- There's a test net up and running right now with the latest version of everything as it exists on the PR branch. This is subject to change but we want to repro the routing issue and confirm that when fixed it shows up fixed on test net.

zalberico:
The galaxy routing issue we were seeing turned out to be DNS related.

We think we're good to go on further testing.

Ted mentioned he wanted to have a test for an old galaxy - one without remote scry running. Test to make sure it doesn't break on test net.

After that we're ready to move on to the next step in the list (running on moons from ~zod) - check that things look good there.

zalberico:
Target: Finish up test net testing this week, start live net testing next week.

zalberico:
The test net has been running with the latest changes - seems to be working fine with OTA and scry request. Did the galaxy remote scry test with a moon running the update and that galaxy not - worked as expected.

On mainnet tested and things also look good there.

Jose to put the specific flows tested in a doc for us so we can repeat this with encrypted remote scry (and so if we hit a problem on deployment we know the categories we tested).

zalberico:
The testing plan: https://gist.github.com/yosoyubik/f6e0e8d2bb54350f0f25fe806d392e8f

zalberico:
Next step - testing remote scry requests with two planets on live net. Jose has L2 planets we can test with (easier/cheaper to breach if there's an issue).

Run OTA between the L2 planets.

Update a live net galaxy (which one is an open question, preferably one with not a ton of active stars).

[...]

Issue #6404, opened 2023-03-20
Use correct `vere` binaries in CI for `next/kelvin/*` branches


matthew-levan:
Jobs run on `next/kelvin/*` branches should use `vere` builds from the corresponding `next/kelvin/*` branch in `urbit/vere` so that Kelvin releases can be tested in CI automatically.

Issue #305, opened 2015-06-17
TreeStore gives a null for /pub/doc/arvo even after it's loaded


ohAitch:
Displays fine upon navigating away and back

galenwp:
Fixed on demo

PR #6406, opened 2023-03-21
Add Remote Scry Protocol

belisarius222:
Resolves https://github.com/urbit/urbit/issues/5788.

TODOs:
- [x] **find out why the compiler is super slow on this branch** EDIT: doesn't seem much slower than other branches
- [x] make corresponding Vere PR
- [x] link to Vere PR here so CI can pass
  - [x] fix CI errors
    - [x] aqua—untested
    - [x] /tests/sys/grq—etch-shot changed the definition of a packet (or .shot) by adding two new flags (req and sam) 
    - [x] /tests/sys/zuse/balk
- [x] fix Clay compilation errors
- [x] stop instantiating bunted `+pe` core in Ames in the `%hunk` scry handler
- [x] check that -keen works correctly 
  - [ ] and its arguments make sense
  - [x] add a way for %spider to track/cancel scry requests. if the thread is cancelled, tell ames to remove this request
    - currently spider tracks any threads that produce a %keen task to %ames
    - If a thread is done, stops or fails, it will remove its ID from the map of threads it keeps in its state, sending clean up cards (not when is done) to ames to remove the scry request
- [x] include outgoing  scry requests to the ~debug dashboard
- [x] add a command to cancel scry request (|pass might work? maybe not—due to how ames handles ducts from vanes, it might not work, so adding a force flag to %yawn that cancels all the scry requests that any vane has requested)
  - `|wham ~ship path` removes all listeners associated with that scry path and gives them `%miss` gifts
- [x] check that scrying into %gall works—will need to merge https://github.com/urbit/urbit/pull/6388
  - [x] remind @belisarius222 to include his testing app
- [x] write -pine thread to test request at latest
  - %pine currently goes through %ames (see +on-plea) so it costs a disk write in the publisher. 
  - it should work for %clay 
  - [x] check for a requester that knows about %pine, and an old publisher
    - The publisher will crash, either because it expects a %cork plea, or it doesn't support %$ as the recipient vane (see https://github.com/urbit/urbit/pull/6406/commits/8387a28bd2075cb38b54b85dca0f19c3e5d8526c) and the requester sinks the naxplanation as expected
  - [x] check requesting a case for a path that has no data
    - The current behavior is, if when we scry for the case we get `[~ ~]`, we nack, the requester will then emit a %miss to the original initiator of the remote scry 
- for gall (see also https://github.com/urbit/urbit/pull/6388)
  - (if you bind something to %gall, that will be public data; otherwise we need a real permission system for agents). 
  - [x] support %w scry (probably not?) (same as for clay that gives you all the cases: %ud.. etc). use case as %gall, what the latest case that's been bound to this scry path
  - [x] %a to scry into of gall
  - [x] ~%y~ %t scry, to get what paths are bound
  - [x] check what happens if you nuke an app? (it should know how to resume if you revive the app)
- [x] add %miss for %keen (e.g. for malformed scry responses)
- [x] check that getting a scry request for ships that cant' handle them is discarded gracefully
  - probably adding something to %ames for handling it. 
  - add teaser support for remote scry for kelvins prior to the full release, so it checks the packet header and no-ops (it looks like vere is already doing this)
  - check the existing behavior, vere might already drop those packets
- [ ] general testing
  - [ ] packetization unit test (low risk)
  - [ ] ames unit test: 
    - idea is to check moves that come of a %keen task, use then to construct the call to the server ames; see tests/sys/grq.hoon)
    - [x]  1-packet remote-scry request/response
      - ~~currently this is failing for some reason. I get a "can't recenter the guard page" error but it doesn't seem to be the test crashing because its  (as far as I can see) emitting the expected moves~~ CI passes again
    - [ ] client that makes an outgoing scry request where the response is a 3 packet message, with a page size of 2 (will need to construct manually)
  - [x] unit test: have two different ducts ask %ames the same request scry path and check that it's not duplicated, and that the response is sent to both.
  - [ ] diagnostic tools
    - [x] added /fine/ducts/[path]
    - [ ] ... 
  - [x] ames verbosity (verb/sift)—`|ames-verb %fin`
  - [ ] vere verbosity 
- [x] include the size of the scry cache in |mass
- [x] review the code for general cleanliness (record meeting)
  - [x] ames.hoon 
    - [x] (`%keen`) 
    - [x] (`%miss`)
  - [x] ames.c
  - [x] gall.hoon
  - [x] clay.hoon?
- [X] test on fakezods
- [x] test on moons
- [x] stretch goal: make new `[%$ *]` namespace in Gall consistent across all cares (this includes changing arvo)
- [x] requests are signed, but those signatures are never verified
- [x] life and rift in the fine-specific namespace paths are never checked
- [x] (fixed) :spider is special-cased for %keen cancellation (s/b %give %miss)
- [x] error handling is partial and suspect
- [x] namespace paths are parsed/validated in ad-hoc and confusing ways (see +de-omen call in %clay)
- [x] %fine interface is hard to use
- [x] remove dead code path in clay (%warp-index induced read)
- [x] clean up the names of the remote scry interface
- [x] investigate the crash molting from %larva to %adult ames, booting with -x
- [x] Fix Clay nest-fail on `+load` from version `%12` to this PR
- [x] rename `*-etch-keen` in `ames.hoon` to `-etch-wail`
- [ ] figure out what (if anything) to do about invalid `$path`s in Clay, such as README.md, since they can't be fetched using Fine
- [x] remove signature field from scry requests, replacing with a version byte
- [x] check if `.siz` can be safely removed from `$meow` without causing problems with leading zeros


wicrum-wicrun:
The Vere PR is up at urbit/vere#305, both that and Arvo compiles. A simple `|merge` between two ships works and does use remote scry, but of course more extensive testing is needed. Not sure how to force CI to rerun using the correct Vere.

yosoyubik:
Regression bugs have been fixed as the latest commit. 

RE using the correct vere binary: I think for CI to work, if I understand https://github.com/urbit/urbit/issues/6404 correctly, we would need to merge https://github.com/urbit/vere/pull/305 into /next/kelvin/413 in the vere repo?

yosoyubik:
- As of this commit https://github.com/urbit/urbit/pull/6406/commits/8e54f7fcc3cf9fc78b47865d195c044fc6822cea `-keen` is working fine, first time we get a miss but subsequent ones hit the cache.

- I started exposing the scry-state date to the ~debug dashboard since that probably will be useful right away while we are testing. So far I have the backend (i.e. json encoding) almost done—we will probably want to revisit it again once I push my changes to decide what/how we want things exposed—but haven't started with the fronted.

- Deleting scry requests by adding a force flag to %yawn works—I tested this by scrying for future revisions, check that a new %keen request is fired on a timer, and then calling %yawn with the flag on deletes all ducts from the state. Now, we should probably handle notifying all listeners that the scry was cancelled? So far only %clay (and `-yawn`) send %yawn tasks).
  - Do we prefer a separate task instead? Thinking about having the option to removing all listeners as something special outside of the normal "remove me as listener", basically removing the option for a scry client to cancel all other clients
  
- Started looking into `-pine` but ~~there's something I'm not following~~

disregard this, I was getting confused because I was assuming the that prior remote-scry work already accounted for ames-to-ames pleas using %$ as the recipient vane but that work was introduced after. I'm updating the logic to account for that (i.e. -pines work as expected)

yosoyubik:
- /~debug has a minimal integration with the remote scry data—it will need another pass to make everything look as we want, but the data is now exposed there.

- Currently looking into some weird behavior related first contact with aliens (fun!)—it looks like we are mixing up /pump wires somehow? Need to investigate more but merging from a remote desk, and then doing |hi ~ship just hangs after getting a timer for a strange wire—there's also a lot of printfs of "~ship is your neighbor", possibly related.

- Tested scrying from an old ship (i.e. before OTAing remote scry) in two scenarios:
  - With an old interpreter: nothing happens—probably vere just discards this? will need to confirm
  - Running with the new remote-scry enabled vere—this just produces a "fine: miss". After the remote-scry OTA is applied, the receiver will get the scry requests again and serve them as needed.

belisarius222:
Yes, a separate task to kill all requests for a certain path is a good idea -- easier to permission, if nothing else.

yosoyubik:
Latest update:

- Fixed some bugs for `-keen`/`-pine` but current tests work—minor revisiting their arguments; currently just ship and path.
- %spider tracks remote scry requests from any thread that sends %keen/%pine tasks to %ames (by adding them to a `scries.state (map tid [ship path[)`, and stopping any thread will send a %yawn task to %ames to remove the scry from its state.
- Added `|wham` for removing all scry request to a ship for a path. It also notifies all (currently just %spider) listeners and sends appropriate clean up moves—question, what about %gall and %clay
- Extended the `~debug` dashboard with remote-scry data—very crude now, will probably revisit for styling and maybe better display of the current information
- For `%pine`, if there is no case for this path (`[~ ~]`),the publisher will send a `%nack`, and the receiver will emit a %miss to the initiator of the scry request—crashing instead of nacking does the same but it doesn't look that clean?

Next:

- Look into the %gall PR: https://github.com/urbit/urbit/pull/6388

wicrum-wicrun:
Using the agent below we can now scry between two ships! Name the agent `test.hoon` and use these commands:

On publisher:
```
:test [%grow /foo 'LXj']
```
On other ship:
```
`:test [%keen <publisher> /g/x/0/test//foo]`
```

```hoon
/+  default-agent, verb, dbug
::
|%
+$  poke
  $%  [%grow =spur =page]
      [%tomb =case =spur]
      [%cull =case =spur]
      [%graw =spur =(list page)]
      [%keen =ship =spur]
  ==
--
::
%+  verb  &
%-  agent:dbug
^-  agent:gall
|_  =bowl:gall
+*  this  .
    def  ~(. (default-agent this %|) bowl)
++  on-init   on-init:def
++  on-load   on-load:def
++  on-save   on-save:def
++  on-watch  on-watch:def
++  on-poke
  |=  [=mark =vase]
  ^-  (quip card:agent:gall _this)
  :_  this
  =+  !<(=poke vase)
  ?-  -.poke
    %grow  [%pass /grow %grow [spur page]:poke]~
    %tomb  [%pass /tomb %tomb [case spur]:poke]~
    %cull  [%pass /cull %cull [case spur]:poke]~
    %keen  [%pass /keen `note:agent:gall`[%arvo `note-arvo`[%a `task:ames`[%keen [ship spur]:poke]]]]~
    %graw  ~&  now.bowl
           %+  turn  list:poke
           |=  =page
           [%pass /grow %grow spur:poke page]
  ==
::
++  on-arvo
  |=  [=wire =sign-arvo]
  ^-  (quip card:agent:gall _this)
  ?>  ?=([%ames %tune *] sign-arvo)
  ~&  [%received data=data.sign-arvo]
  `this
::
++  on-peek   on-peek:def
++  on-leave  on-leave:def
++  on-agent  on-agent:def
++  on-fail   on-fail:def
--
```

joemfb:
@yosoyubik i worked around the ames compilation issue, although i don't entirely understand it. Judging by the printfs, it was somehow confusing `keens` in `$alien-agenda` and `$peer-state`. Something's off with the namespacing, but I'm not sure what exactly.

I also see the `bail:meme` in the test suite. Trying to get a trace right now. I'll post updates on that and the error handling ASAP.

joemfb:
The `bail:meme` was due to trying to print vase of vases on test failure.

The failure itself is that `%tune` is only give once, not twice as expected. There are two listeners in the state, and the `+rep:in` callback in `+fi-done` is called twice, so I suspect that the move list is being clobbered somehow. The namespacing here looks incomplete to me -- I'm not sure which `+emit` we should be calling from inside `+fi`.

joemfb:
Fixed the missing moves. Now the tests are failing due to mismatched signatures. I'm not seeing this failure locally, which is pretty concerning. Where did the signature value in the test come from @yosoyubik?

yosoyubik:
Thanks @joemfb! Yeah... the aliasing happens inside the `|pe` core, and that uses `=peer-state` so I would expect it to take the keens from that, not the alien? I've tried different permutations of aliasing in different places but all give the same tack error.

I took the signature from the `etch-hunk` core, basically what `make-meow `does, knowing that it would be a one packet response, but I can see that `sift-roar` prints a different signature (I suspect because of the way it assembles the response message, assuming a fixed packet size of 1kb) than the ones that comes in to `on-hear-fine`, and using that for the expected value works.

(Working on updating the unit tests to produce the expected signature using sift-roar)

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -4034,7 +4051,7 @@
             %-  (fi-trace fin.veb |.("done {(spud fi-full-path)}"))
             %-  ~(rep in listeners.keen)
             |=  [=^duct =_fine]
-            fine(event-core (emit:fine duct %give gift))
```
wicrum-wicrun: @yosoyubik Since you removed this and Joe later had to readd it, it seems worth to explicitly explain the pattern:

Since we're accumulating results in new versions of `fine`, we need to call the `+emit` that is available through the accumulated `fine`, not the one that we have in scope in the initial version. Otherwise only the last gift will actually be given, since we're overwriting `event-core`'s state every time we emit.

yosoyubik: Thanks! Yeah, the removal was a spurious one, and it shouldn't have happened—the pattern I had used before (e.g. +fi-error, +fi-clean-up) was making the core in the accumulator the event core, not fine, since there was no emit arm there. I added one now https://github.com/urbit/urbit/pull/6406/commits/5c838a42a795a83f651e703d0d312c08ef4103f0 so hopefully that's more clear.

joemfb:
@yosoyubik thanks, but i'm still worried. Somehow the tests at [790a0de](https://github.com/urbit/urbit/pull/6406/commits/790a0de29251d00dba1745cc2177d16c89488e12) passed locally for me, but failed in CI (due to mismatched signatures). I don't see any sources of non-determinism, so I don't have any explanation for why that would be the case. I'm going to rebuild a fake ship to see if I can reproduce.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3668,13 +3767,500 @@
...
+            |=  [=^duct =_fine]
+            ::  at the moment only %spider keeps state about remote scries
+            ::
+            ?+  duct  fine
+                [[%gall %use %spider @ ship=@ %thread tid=@ *] *]
+              =/  =cage   spider-stop+!>([&7.-.duct |])
+              =/  poke=*  [%0 %m [p q.q]:cage]
+              =/  =plea   [%g /ge/spider poke]
+              (fi-emit:fine duct %pass /fine/unsub %g %plea our plea)
+            ==
```
joemfb, line 3902: This pattern is not sustainable. Other clients need to be able to add/remove such state without upfront coordination through %ames. A %miss should be given on all these ducts.

joemfb (changes requested):

yosoyubik:
Yes, this is concerning, my tests were passing as well. I was looking into it but got distracted with https://github.com/urbit/urbit/pull/6483#issuecomment-1512525973

joemfb:
Tests all still pass for me (macos m1, running vere from urbit/urbit#305) on https://github.com/urbit/urbit/pull/6406/commits/790a0de29251d00dba1745cc2177d16c89488e12. Can somebody run them on linux?

(Boot a fake ship from `bin/solid.pill`, specifying `-A`, mount base, copy in the tests directory, commit, `-test %/tests ~`.)

Just thought of another difference: I tested a fake ~zod, CI tests a fake ~bus. I'll try again ...

joemfb:
~bus also passed.

yosoyubik:
Trying on Linux now

joemfb:
I pushed some more commits (strangely not showing up on this PR for me, looks like there's a github incident). Mostly just minor cleanup and removing unused things.

The big change is removing `%pine`. After discussion with @belisarius222, it's clear that there's no advantage to the current implementation living in kernelspace. It or something like it can be added to userspace at any time, without loss of performance or precision. So we can keep this interface smaller.

yosoyubik:
Commits are showing up now

pkg/arvo/sys/vane/ames.hoon
yosoyubik: Pushed this https://github.com/urbit/urbit/pull/6406/commits/873de612691baba42dbce62349488fa7af4fb6b6 addressing your request, let me know how it looks.

yosoyubik:
It looks like the "ames.c" task—also mentioned in [the vere PR](https://github.com/urbit/vere/pull/305#issuecomment-1494564471)—was already merged here: https://github.com/urbit/urbit/commit/062fbc71d78681329f76e77b2ed1003f2a650d22#diff-193328b9512deb69019a31a06bc4b9deefe0debd0384b39e955d82a3652628aa

One things is `FINE_PAGE` is different (from `16384` to the actual `512`)

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3902,12 +3901,17 @@
...
-            ?.  |(all (~(has in listeners.keen) duct))
-              %.  fine
-              (fi-trace fin.veb |.("unknown listener, skip"))
-            =?  fine  all  (~(rep in listeners.keen) fi-miss)
-            %-  (fi-trace fin.veb |.("deleting {<fi-full-path>}"))
-            fine(listeners.keen ?:(all ~ (~(del in listeners.keen) duct)))
+            ^+  fine
+            ?:  all
+              %-  (fi-trace fin.veb |.("unsub all {<fi-full-path>}"))
+              (~(rep in listeners.keen) (fi-give-tune ~))
```
yosoyubik, line 3907: we need to delete all the listeners here so +fi-abet removes the path from the keen state

joemfb: Thanks, that's my bad.

yosoyubik:
I pushed a first pass at fixing the namespace mismatches:

- Arvo
  - Adds `$yowl`: a parsed $purr ([peep meow])
  - Updated `$purr`: as a `[peep meow]`
  - Renames `$keen` with `$wail` (reduces confusion with ames.hoon usage of `keen` as the `keen-state`)

- [Vere](https://github.com/urbit/vere/pull/305/commits/3c95f4bbfa96747c34d771ea578fc708d00e0196)
  - Renames `_u3_keen` to `_u3_peep`

joemfb:
The arvo changes here merit a kelvin decrement, holding off until that can be wired up correctly in CI.

Issue #6285, opened 2023-02-03
OTA: If a third party app causes OTA failure - alert the user and give instructions


zalberico:
In the last we release we had the following happen:
- Citadel and Gora signaled support for 415 kelvin
- When the 415 base OTA was released users attempted to suspend other apps without 415 support and upgrade
- Citadel and Gora didn't actually work on 415
- The OTA makes an assumption that apps marked supported are expected by the user to work, therefore if those apps don't successfully upgrade, the system rolls back the OTA to a working state.
- There's then a bug where the system attempts to retry the OTA forever
- The user is then confused about why the upgrade failed and is looping

Ideas:
- While it's reasonable to think the user would want to revert the OTA if some important third party app they thought supported it is actually broken, I'm not sure the system should do that. If an app dev signals support for a new version but doesn't confirm it works on that new version that's on the app dev. I'm not sure we should roll back the base OTA because of this.
- Even if we do decide to roll back, we should then stop reattempting the OTA and loudly tell the user in some form: "Apps X and Y failed to upgrade despite claiming support for the new version. Suspend them and upgrade anyway?"

My preference is probably to upgrade base no matter what and if a third party app fails despite claiming support - the user will complain to the dev. I think this is less confusing/more expected than the system doing what it's doing currently.

Fang-:
I'd argue against suspending apps if they fail to upgrade cleanly despite initially signaling the contrary. Without the user explicitly suspending the app, we have no way of knowing whether that's ok or not: it may be in active use for their business, event, or other "sensitive" use-case, which will take precedence over the system update.  
Suspending apps without first consulting the user gives big "Windows will restart in one minute to apply some updates, press [ok] or [restart now]" vibes.

We _do_ clearly need to handle such states better, and most importantly communicate about them to the user properly, so the second option seems good.

ashelkovnykov:
We already have options to control automatic updates per-app, and for the OS itself. It seems like the trick is finding the right balance. What about introducing the idea of "pinned" or "vital" apps, and then:

1. If `%base` updates disabled, never perform automatic updates
2. If `%base` updates enabled:
    1. If all "vital" apps support new kelvin, attempt upgrade once and revert to prev version if one upgarde fails
    2. Otherwise, prompt user to upgrade manually

If while waiting for `2.ii`, the conditions for `2.i` become met, then attempt the upgrade.

PS - We need the option to install old versions of apps, and to install them with automatic updates disabled (so that they don't update the second they're installed)

[...]

PR #6413, opened 2023-03-23
clay: stop failed kelvin upgrades from retrying immediately

### LINKED ISSUE #6407: clay upgrade loop
My ship tried to upgrade to 414, failed, and now appears unable to make progress, since even after Ctrl-Z and restart, it looks like it's handling a behn timer that triggers a kernel update.

I'm not sure why the kernel would try to update more than once, which seems like a Clay bug to me.  I'm more immediately concerned with trying to get my ship (and @jalehman's ship, and possibly others) back up and running ASAP.  If anyone has an idea (maybe an event to inject?), I'd like to hear it.

I think the failure was caused by a non-%base desk that hadn't updated its `/sur/hood/hoon` file to match the one in `%base` and was still referring to `$foam`, which was deleted in this Kelvin.

image

image

philipcmonk:
We were retrying failed kelvin upgrades as many times as we had apps that needed to be suspended, because suspending an app triggers an attempt to run the next kelvin upgrade.  This suspends all those apps in one batch move, and then tries the next kelvin upgrade only once at the end.

Targeting master because we may want to hotfix this ahead of kelvin 414.  If not, this can change to targeting develop, no need to rebase.

Fixes #6407

Partially addresses #6285


belisarius222 (approved):
I won't merge this until we're sure we're ready to push it to the wide network, but LGTM

PR #6423, opened 2023-03-27
hoon: add +zip to jar engine

dr-frmr:
This PR adds a utility to the `jar` engine, `zip`. Like sets and maps, jars can now be converted to lists of key-value pairs.

Since you can already get a listified version of a jar with `tap:by` that has each value-list paired with its key,`zip:ja` treats each element in the jar's value-list as an individual and pairs each one with the key. For example, a jar that had list `[1 2 3 ~]` at key `%a`, and list `[4 5 6 ~]` at key `%b` will be converted into the following list:
`[[%a 1] [%a 2] [%a 3] [%b 4] [%b 5] [%b 6] ~]`


dr-frmr:
@joemfb do you think this should be named `+zip` instead of `+tap`? I kind of like zip as we discussed on Friday.

joemfb (changes requested):
`+zip` sounds good to me.

This implementation looks correct to me -- keys in treap order, lists in-order.

The jet hint is wrong in a couple ways -- you can't ever hint a `|-`, since the trap doesn't escape. And the `~/` rune is for gates, it assumes a parent core at `+7`. See `+wyt:by` for the right way to hint a bare trap.

It'd also be better to only traverse the map once, instead of traversing and allocating in `+tap:by`, then re-traversing/allocating for the "key distribution". There's probably other ways to do that, but it's simplest to copy `+tap:by` and use an accumulator to ensure the correct order. (And once you pin an accumulator outside the trap, `~/` becomes correct.)

Here's how I'd do it:

```hoon
=<  $
~/  %zip
=+  b=`(list _?>(?=([[* ^] *] a) [p i.q]:n.a))`~
|.  ^+  b
?~  a   b
%=  $
  a  r.a
  b  |-  ^+  b
     ?~  q.n.a  ^$(a l.a)
     [[p i.q]:n.a $(q.n.a t.q.n.a)]
==
```

joemfb:
That typecast should also include `p`/`q` faces:

```hoon
=+  b=`(list _?>(?=([[* ^] *] a) [p=p q=i.q]:n.a))`~
```

joemfb (approved):
CI is already broken on this branch, that can be ignored. I've run the changes here locally, this will do.

PR #6425, opened 2023-03-27
sss: Unsubscribing and permissions

### LINKED ISSUE #6409: sss: No way to unsubscribe
If you're subscribed to a path using `lib/sss`, there's no way to unsubscribe.

wicrum-wicrun:
Closes #6409 

Turned into a fairly large rewrite. Still some rough edges in terms of architecture, but the interface should be stable and it works as intended (as far as I can tell after testing), and I wanted to get it released quickly to unblock hodzod and focus on remote scry instead.

Permissions are either set by using `+public:du`, `+secret:du`, `+allow:du` and `+block:du`, or manually by passing a gate to the `+perm:du` arm. Permissions are encoded as `(unit (set ship))`, with `~` indicating public data and `[~ ships]` indicating that only `ships` are allowed. So for example, to only grant access to `~sampel`, call `+perm:du` with `_[~ (silt ~sampel ~)]`. 

It’s also possible to kill a path using `+kill:du`. This deletes the state from the publisher and informs anyone who tries to access it that no more data will come. However, subscribers won’t automatically delete their copy of the state, that’s up to them.

More controversially, you can revive a killed path with `+live:du`. This will start over from the bunt (but at a new aeon, to maintain referential transparency). **Very open to a discussion about this!**

Permissions are maintained even for killed paths, to prevent metadata leaks. A blocked ship can’t tell whether the path is live or dead.

The calling convention changes in a few places. Usage of the new functionality and the new calling conventions are both demonstrated at [wicrum-wicrun/sss](https://github.com/wicrum-wicrun/sss/blob/master/urbit/app/simple.hoon) as before, though some of the cases still need to be properly documented. (All of the new user-facing arms in the library do have comments though.)

**Unfortunately this is not backwards compatible**, since I failed to tag my state in the previous version. State migration would've run a `;;` on the entire subscription/publication state *even when it was already up to date*, which seemed unacceptable. Given the relative novelty and experimental nature of this library, I figured that it's acceptable for things to break. Both @rabsef-bicrym and @dr-frmr have been informed, as they are the main developers using this afaik. Paging @litlep-nibbyt too. Sorry about that, I've made sure that if I change the state again this won't happen.

### DIFF
--- a/pkg/base-dev/lib/sss.hoon
+++ b/pkg/base-dev/lib/sss.hoon
@@ -260,23 +260,16 @@
   ++  block                                  ::  No-ops on public paths.
     |=  [who=(list ship) whence=(list paths)]
     ^-  pubs
-    %+  edit  whence
-    |=  =buoy
-    ?@  tid.buoy  buoy
-    ?~  alo.buoy  buoy
-    %=  buoy
-      alo      `(~(dif in u.alo.buoy) (sy who))
-      mem.tid  (~(dif by mem.tid.buoy) (malt (turn who (late ~))))
-    ==
+    %+  perm  whence
+    |=  old=(unit (set ship))
+    ?~  old  ~  `(~(dif in u.old) (sy who))
   ::                                         ::  Allow ships to paths.
   ++  allow                                  ::  Any public paths will no-op.
     |=  [who=(list ship) where=(list paths)]
     ^-  pubs
-    %+  edit  where
-    |=  =buoy
-    ?@  tid.buoy  buoy
-    ?~  alo.buoy  buoy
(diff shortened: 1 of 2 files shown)


dr-frmr (approved):
https://github.com/uqbar-dao/nectar/pull/5

Already using this update here to great effect. Quitting a path works, as does adding/removing permissions. My one feedback would be to document subscription-permissioning somewhere, perhaps in a separate app to `simple.hoon` in the main SSS repo. I think `simple` could be even simpler, and there could be a second agent that shows the more advanced abilities.

belisarius222 (approved):
LGTM.  I also thought about the killing and permissions semantics for a while, and they sound right to me.

PR #6427, opened 2023-03-28
arvo: add more files to |new-desk generator

### LINKED ISSUE #6426: Make `|new-desk` create more files to increase usefulness
Currently `|new-desk` creates a minimal desk.  However there are several additional files that almost every app needs:

- /mar/bill.hoon
- /mar/mime.hoon
- /mar/json.hoon
- /lib/skeleton.hoon
- /lib/default-agent.hoon
- /lib/dbug.hoon

These files are probably not needed for most library-only distribution desks.

Proposed semantics:

- `|new-desk some-desk` => creates a desk with extra files, applicable to a Gall agent desk
- `|new-desk some-desk, =minimal &` => creates a minimal desk (equivalent to the current behavior)

playfulpachyderm:
Resolves #6426.

Add an optional argument "minimal" to `|new-desk` which preserves current behavior; otherwise, by default, include an additional list of files in the new desk that will help jump-start development of a Gall agent.

### DIFF
--- a/pkg/arvo/gen/hood/new-desk.hoon
+++ b/pkg/arvo/gen/hood/new-desk.hoon
@@ -5,7 +5,7 @@
 :-  %ask
 |=  $:  [now=@da eny=@uvJ bek=beak]
         [=desk ~]
-        [from=$~(%base desk) hard=_| minimal=_|]
+        [from=$~(%base desk) hard=_| gall=_|]
     ==
 ::
 =;  make-new-desk
@@ -34,7 +34,7 @@
         /mar/kelvin/hoon
         /sys/kelvin
       ==
-    =/  extra-files=(list path)  ?:  minimal  [~]
+    =/  extra-files=(list path)  ?.  gall  [~]
       :~
         /mar/bill/hoon
         /mar/mime/hoon
--- a/pkg/arvo/gen/hood/new-desk.hoon
+++ b/pkg/arvo/gen/hood/new-desk.hoon
@@ -5,7 +5,7 @@
 :-  %ask
 |=  $:  [now=@da eny=@uvJ bek=beak]
         [=desk ~]
-        [from=$~(%base desk) hard=_|]
+        [from=$~(%base desk) hard=_| minimal=_|]
     ==
 ::
 =;  make-new-desk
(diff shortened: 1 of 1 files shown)


belisarius222 (dismissed):
LGTM.  Anybody upset by this?  @ryjm @Fang- @rabsef-bicrym let me know if you dislike -- ofherwise I will merge soon

midden-fabler:
I think `|new-desk` should continue producing a minimal desk, and e.g. `|new-desk, =gall &` could include additional files.

pkg/arvo/gen/hood/new-desk.hoon
```diff
@@ -27,12 +27,23 @@
...
-    :~  /mar/noun/hoon
+    =/  common-files=(list path)  :~
+        /mar/noun/hoon
         /mar/hoon/hoon
         /mar/txt/hoon
         /mar/kelvin/hoon
         /sys/kelvin
-    ==
+      ==
+    =/  extra-files=(list path)  ?.  gall  [~]
```
belisarius222, line 37: style: `[~]` -> `~`

belisarius222 (approved):
It would be nice to tweak that style nit at some point, but it's not a blocker.

rabsef-bicrym:
Sorry for my belated reply, but I would agree with @midden-fabler above that the default should be the minimal. I think it's a difference of opinion tho, so I don't have strong feelings either way.

Fang-:
Also late to the party, but would err on the side of stronger minimalism here. "I want this desk to exist" and "I want to use this desk for an app" are two different cases, and the latter need should be served by dependency management solutions. Not to mention, any inclusion that serves the latter case may not be necessary in _all_ cases.

(Given that we don't have a real dependency management story, maybe this is _slightly_ more defensible, but I'd still be against it.)

playfulpachyderm:
I updated it to be minimalist by default before merge (needs `=gall &` to create extra files now), is this ok

playfulpachyderm:
> Also late to the party, but would err on the side of stronger minimalism here. "I want this desk to exist" and "I want to use this desk for an app" are two different cases, and the latter need should be served by dependency management solutions. Not to mention, any inclusion that serves the latter case may not be necessary in _all_ cases.
> 
> (Given that we don't have a real dependency management story, maybe this is _slightly_ more defensible, but I'd still be against it.)

I agree but as you said there doesn't appear to be any dependency management which is why I opened the issue.  Actual dependency management would obviously be better though.  I liked this: https://github.com/urbit/urbit/pull/6402 and would like to help put it together, though my hoon skills are modest at the moment (I just finished App School).

PR #6438, opened 2023-03-30
hoon: fix @da parser (DO NOT MERGE)

mikolajpp:
This turned out during my ongoing work on slaw, scot jets. There is a bug in @da parser. 

In Hoon, there is no year 0 (in the sense that no date atom will ever render to a date with year 0. See `++yore` for details). 
However, the parser admits a date such as `~0.1.1` as valid, while 
the printer will render the resulting date atom as `~1-.1.1'. 

There is also an associated bug: trying to type `~0-.1.1` (note the hep) will result in a decrement error and dojo crash.
This is because hep marks the year as BC, and there is an internal book-keeping calculation which will try to decrement the year (already 0). 

The fix is to disallow year 0 in the date parser.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -5798,7 +5798,7 @@
     ;~  plug
       %+  cook
         |=([a=@ b=?] [b a])
-      ;~(plug dim:ag ;~(pose (cold | hep) (easy &)))
+      ;~(plug dip:ag ;~(pose (cold | hep) (easy &)))
       ;~(pfix dot mot:ag)   ::  month
       ;~(pfix dot dip:ag)   ::  day
       ;~  pose


joemfb (commented):
This is a good catch!

Merging this as is will create a jet mismatch (in our old friend slaw). It should go in a hoon kelvin bump, along with the coming new jet at the new kelvin.

mikolajpp:
Ok, perhaps in this case it should go into the eventual slaw/scot jets PR? There are more problems around parsers, so this is not the only commit that is going to affect `sys/hoon.hoon`.

joemfb:
That sounds good. Please open separate PRs for each of those other fixes for easy review, we'll accumulate/release the changes together with the jet.

mikolajpp:
This PR is superseded by a new PR covering a wider range of parser issues: https://github.com/urbit/urbit/pull/6628

Issue #6457, opened 2023-04-07
`+vats` crashes on nonexistent desks


belisarius222:
from `~finned-palmer`:

```
bail: 4

bail: 2
! /cx/~finned-palmer/alk/~2023.4.7..11.15.34..fa91/sys/kelvin
[%dojo-lame %arvo-response]
> +vats
```

from `~rovnys-ricfer`:

```
bail: 4

bail: 2
! /cx/~rovnys-ricfer/esacpe/~2023.4.7..18.12.53..0bdc/sys/kelvin
[%dojo-lame %arvo-response]
! /cx/~rovnys-ricfer/esacpe/~2023.4.7..18.12.53..0bdc/sys/kelvin
[%dojo-lame %arvo-response]
> +vats
```

`%alk` and `%esacpe` have empty entries in Clay's map of desks, from the user having misspelled `%talk` and `%escape` respectively.  They should show up as `%nonexistent` desks in `+vats`, and `+vats` should not crash under any circumstances.

This is almost certainly a regression due to https://github.com/urbit/urbit/pull/6395.

PR #6458, opened 2023-04-07
+vats: fix crash on nonexistent desks

belisarius222:
Resolves https://github.com/urbit/urbit/issues/6457.


philipcmonk (approved):
Yep, that should do it

PR #6461, opened 2023-04-09
ames: reset timer for naxplanations on corked bones

### LINKED ISSUE #6451: ames: ack crashed %exit after removing a naxplanation flow on closing bones
[Full output](https://gist.github.com/yosoyubik/2fea18cc270b418b35777131b94e883d)

A publisher is, after sending an ack for a %cork, removing the associated flow—and in this case, it had previously also sent nacks (e.g. if it crashed when it received the %watch), so we remove the naxplanation bone as well.

We tell %behn to reset the timer for the backward bone to stop sending %boons, but we don't do that for the naxplanation bone—and that's the one we crashed on, so it looks like we could be trying to send that again, and when we wake up, the flow is gone, so we crash here `?<  =(~ next-wake.state)` because we are doing *message-pump-state.

We have [a check](https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/ames.hoon#L2486) to avoid this kind of thing happening. If we wake up and need to sent some timed out packets, but the bone is corked, we no-op, but... we are not doing that for naxplanations that use a  different bone `(mix 0b10 bone)`, so we should probably be adding that bone to the check.

Now, this error shouldn't repeat for the same ship—and this is what I can see on ~norsyr-torryn. @zalberico, after doing `|ames-verb %odd`, do you see a "remove naxplanation flow" message?

yosoyubik:
Fixes #6451

(testing on ~norsyr-torryn, will open for review after is done)

(Update: I have not seen an instance of the issue this PR is solving on my ship, so I haven't been able to find a repro to test it yet)

(Update 2: [repro found](https://gist.github.com/yosoyubik/ca58b34f0781c22a7d27c562f056d3b1), opening for review)

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -1777,7 +1777,7 @@
           %.  &
           (ev-trace &(dry odd.veb) ship |.((weld "stale %watch plea " log)))
         ::  the current subscription can be safely corked if there
-        ::  is a flow with a naxplanation ack  on a backward bone
+        ::  is a flow with a naxplanation ack on a backward bone
         ::
         =+  backward-bone=(mix 0b10 bone)
         ?.  =(2 (mod backward-bone 4))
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2379,9 +2379,13 @@
                 :_  tang.u.dud
                 leaf+"ames: {<her>} fragment crashed {<mote.u.dud>}"
             abet:(call:abed:(sink-core bone) %hear lane shut-packet ?=(~ dud))
-          ::  benign ack on corked bone
+          ::  benign ack on corked bone (also for naxplanation acks)
           ::
-          ?:  (~(has in corked.peer-state) bone)
+          ?:  ?|  (~(has in corked.peer-state) bone)
```
belisarius222, line 2384: Since this code is repeated, it should be factored out into an `+is-corked` gate.

belisarius222 (changes requested):
Requested a minor cleanup, but this looks like it works, so I'll approve once that little function is extracted.

belisarius222 (approved):

PR #6462, opened 2023-04-11
vanes: remove vane scry nonexistent marks

### LINKED ISSUE #6411: vanes: vane scries should not return marks that do not exist
For two examples, [clay returns a `%file-list` mark in response to a %ct scry](https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/clay.hoon#L4223-L4250), and [gall returns an `%apps` mark in response to a `%ge` scry](https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/gall.hoon#L1962).

This causes problems for %pyro scries.

There are two obvious solutions here:

1. Change the scry marks to `%noun`.
2. Add `mar` files with appropriate arms for handling noun conversion to each possible scry mark.

hosted-fornet:
Resolves #6411 .

Removes marks that have no corresponding `mar` file from scries in vanes, replacing them with `%noun`.

As mentioned in #6411, these nonexistent marks lead to problems in %pyro ship scries. I'm happy to provide more details.

I considered an alternative solution, namely, to add `mar` files for each of the marks that appear in the vane scries. The advantage would be avoiding any possible breaking of userspace. The disadvantage is that we gain nothing by having these `mar` files around, and there are many of them:

```hoon
mass  `(list mass)`
vase  `vase`
dais  ``
tube  ``
nave  ``
cast  `vase`
tako  `tako`
yaki  `yaki`
page  `page
uvi   `@uvI`
uvs   `(list @uv)`
cage  `cage`
open  ``
file-list  `(list path)`
flag  `?`
dome  `domo`
cass  ``
arch  `arch`
sweep  
rang  
cult  
flow  
tire  
tyre  
domes  `cone`
apps  `(set [dude ?])`
nonces  `(map dude @)`
seed
```

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -3956,7 +3956,7 @@
       :~  peers-known+&+known
           peers-alien+&+alien
       ==
-    ``mass+!>(maz)
+    ``noun+!>(maz)
   ::  only respond for the local identity, %$ desk, current timestamp
   ::
   ?.  ?&  =(&+our why)
--- a/pkg/arvo/sys/vane/behn.hoon
+++ b/pkg/arvo/sys/vane/behn.hoon
@@ -250,7 +250,7 @@
     =/  maz=(list mass)
       :~  timers+&+timers.state
       ==
-    ``mass+!>(maz)
+    ``noun+!>(maz)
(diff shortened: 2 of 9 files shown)


belisarius222:
@hosted-fornet Since you're planning on trying a different approach, I will close this for now.

Issue #1775, opened 2019-09-28
Reject invalid number formats in `de-json:html`


courajs:
(details to be added soon, see https://github.com/urbit/urbit/pull/1769)

Issue #5637, opened 2022-03-14
Exponent size limited in JSON reparser


ashelkovnykov:
**Description**

While working on the JSON jets (#5566), @joemfb and I found that the JSON reparser (specifically the call to `+ryld` made by `+ne:dejs:format` on the output of `+json-rn:format`) cannot handle very large exponents. I'm not sure what the exact limit is (it seems to depend on how much memory or loom is available), but it's somewhere in the 9-digits. Attempting to case a JSON number such as `1e987654321` will cause a failure.

**To Reproduce**

```
> (ne:dejs:format [%n '1e9999999999999999999999999999'])
bail: fail
...
> (scan "1e9999999999999999999999999999" json-rn:format)
[%d a=%.y b=[c=1 [d=0 e=0] f=%.y i=9.999.999.999.999.999.999.999.999.999]]
> (royl-cell:so (scan "1e9999999999999999999999999999" json-rn:format))
[%d s=%.y e=--9.999.999.999.999.999.999.999.999.999 a=1]
> (ryld (royl-cell:so (scan "1e9999999999999999999999999999" json-rn:format)))|
bail: fail
...
```

**Expected behaviour**

There's several possible solutions that I can think of:
1. Return infinity (Hoon already does so for large exponents smaller than these: `.~1e9999` -> `.~inf`)
2. Have "safe" versions of the function(s) which return `unit`
3. Limit the accepted size of exponents and document it in the tests/comments
4. Handle really big numbers with something like `BigInteger` from Java

Issue #1776, opened 2019-09-28
Parse unicode escapes properly in `de-json:html`


courajs:
Right now unicode escape sequences (like `\u22ff`) are dropped or rejected.
More details soon, or see https://github.com/urbit/urbit/pull/1769

PR #6463, opened 2023-04-11
zuse: changes in preparation for JSON jets

ashelkovnykov:
Updated version of changes to Arvo from #5566

These changes resolve several issues with how JSON is handled and prepare Arvo for JSON encoding/decoding jets.

Resolves:
#1775
#1776
#5456

Outstanding Issues:
- Haven't re-run the tests from #5566, since the Vere portion of the changes haven't been migrated yet
- @joemfb What's the reason for wanting `+sune`, `+sung`, and `+teff` moved into the `+json` namespace in `%zuse`? Is it just to merge the changes without needing to wait for a `%hoon` kelvin decrement?
    - Is the UF maintaining any branches for banking changes until the next `%hoon`, `%arvo`, and `%lull` decrement?
- Long-standing question over where JSON arms should go in `%zuse`:
    - Should the arms stay in `+html`?
    - If so, should `+html` be renamed?
    - If no, should they go in `+format` or a new arm altogether?
- Related issues:
    - #5637: Limited exponent sizes in the JSON reparser
    - By definition, `cord` is already supposed to be valid UTF-8. Therefore, the correct thing to do would be to move the UTF-8 parsing code into `+sane` (as well as the jets for it), parse any text to `cord` in the parser as before, and then verify that it's valid UTF-8:
      ```
      ;~(pose (sear |=(a=@t ?.(((sane %t) ~ `a)) (cook crip ;~(less doq bas prn)) (cook crip esca))
      ```
        - TODO: make issue for this
    - `+sane` isn't wired up to `+ruth` due to load-bearing aura violations
        - TODO: make issue for this
    - `lib/jose` uses a custom copy of `+en-json` because it needs a custom comparator. If `+en:json:html` is extended to use custom comparators, this cloned code can be removed.
        - TODO: make issue for this
    - Since `+en-json:html` and `+de-json:html` are now deprecated, all uses of them in the kernel should be removed.
        - TODO: make issue for this


belisarius222:
@joemfb says `+sane` is slightly more restrictive about whitespace than standard UTF-8, so we shouldn't change `+teff` in hoon.hoon.  Instead, a modified `+teff` should be used in Zuse for the JSON functions.

joemfb (approved):
this will do

my motivation to not change `+teff` and `+sane` is that `@t` is not exactly utf-8 -- it imposes additional constraints on whitespace (rejecting newlines). that may or may not be a good idea (and it's violated in more than a few places), but it is longstanding and was intentional (i'll see if i can pull up links, maybe from the old urbit-dev list). it's a different kind of change than the others here, and should be decided on for the whole system with context and care. separate tracking issues will be good to have.

ashelkovnykov:
@joemfb Sounds good - I'll make some additional tracking issues and include what context I have now.

@jalehman @belisarius222 I leave this one for you guys to merge whenever and into where ever it needs to go.

PR #6466, opened 2023-04-11
ted/test.hoon: repair ability to run some tests by prefix

dr-frmr:
I believe this is an old feature that has since stopped working. This change removes some old code marked as "vestigial" and replaces it with what seemed to be its intended function, which is to select test arms by matching the beginning of each arm name with the test name entered by the user. I wanted this functionality and it seems to work now, as I use this modified version of `ted/test.hoon` in my fakeship.


dr-frmr:
Oops, accidentally included unrelated symlinks

Issue #6311, opened 2023-02-09
eyre: connection requires frequent refresh to maintain


jalehman:
Users frequently have to refresh the page in order to maintain a connection to eyre. We don't really know why.

@dr-frmr @tadad and @drunkplato are actively looking into this and/or have expressed interest in seeing it resolved. If you have any further context or findings to add, please do so here.

tadad:
so the source of the bug seems to be any time that you don't close a session, but the internet goes out. So sleeping your laptop, backgrounding an app in ios, disconnecting from wifi, etc. all trigger this. @e1block and I have a way to easily reproduce the bug 

First step is adding some extra printouts to %eyre, so finishing up #6303 to help with bug hunting.

matthew-levan:
As far as I understand it (I'm not very familiar with how server-sent events work), the correct flow is something like this:
- A session's underlying internet connection is broken
- Internet connection is later restored
- Client re-establishes an initial connection with the server
- Server continues sending data over the re-established connection

belisarius222:
As @Fang- said, this sounds like it could be a purely client-side issue.  Is there an industry best-practice for handling connectivity loss for a client using server-sent events?  Seems like something that would affect anyone trying to use them in production.

A cursory reading of some SO posts about SSS indicate the browser should automatically try to reconnect.  Not sure if Chrome is just failing here, or if we have something configured awkwardly (a timeout interval set too long, perhaps), or what.

Possibly relevant: https://stackoverflow.com/a/58374643

matthew-levan:
I think we might need to produce reliable reproduction steps for this bug before we can make much more progress. FWIW, I can't reproduce the "forever reconnecting" behavior in Landscape by switching off my network adapter, scrolling far up enough in a chat for Landscape to send a request to the server for more data (and thereby rendering the yellow "Reconnecting..." bar on top of the screen), and then re-enabling my network adapter. In fact, Landscape proceeds to reconnect in this scenario, succeeds, and loads the data it was asking for just fine.

tadad:
<img width="766" alt="Captura de Pantalla 2023-02-13 a la(s) 12 55 12 p  m" src="https://user-images.githubusercontent.com/43223192/218548966-7d6590fe-6b11-4708-8a57-9b9a7297d887.png">
something is really weird here - so the issue stems not from disconnecting - but reconnecting. Somehow on reconnect, something is passing a %cancel-request event directly into %eyre. Must be coming from the frontend? Need to look more

[...]

PR #6471, opened 2023-04-12
eyre: do not clear unacked events in +on-get-request

pkova:
Eyre keeps a queue of unacked events, resending them to a frontend through a SSE stream until they are acked. This behavior is violated in `+on-get-request` where the events are cleared from the queue without waiting to hear the acknowledgement. This is incorrect.

This PR has bearing on #6311, but there are other problems as well. More fixes coming up later.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1348,13 +1348,13 @@
       ::
       =/  heartbeat-time=@da  (add now ~s20)
       =/  heartbeat  (set-heartbeat-move channel-id heartbeat-time)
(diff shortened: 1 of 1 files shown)


pkova:
cc @arthyn

arthyn:
nice!

Fang- (approved):
Yeah. It's not clear to me why this has been the behavior since channels were first implemented. My best guess is there's just some mistaken assumption here that, because we serialize & send the backlog in this flow, we can get rid of it. But sending it doesn't guarantee the client actually receives & processes it.

Backlog/queue should really only be cleared on explicit client ack, and channel deletions/expiry. Good catch @pkova!

PR #6472, opened 2023-04-12
eyre: add support for noun-based channels

Fang-:
Supersedes #6396. Go there to start your point-and-click adventure of the history for this. This PR contains the addition of merge conflicts resolution, and a better jam mime type descriptor.

tldr we let you receive jammed nouns instead of json over eyre channels if you so choose.


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -200,13 +207,44 @@
...
   ?:  (lte u.sus ack)  0
   (sub u.sus ack)
+::  +find-channel-mode: deduce requested mode from headers
+::
+++  find-channel-mode
+  |=  [met=method:http hes=header-list:http]
+  ^-  ?(%json %jam)
+  =+  ^-  [hed=@t jam=@t]
+    ?:  ?=(%'GET' met)  ['x-channel-format' 'application/x-urb-jam']
+    ['content-type' 'application/octet-stream']
```
joemfb, line 217: Can we use `application/x-urb-jam` for PUT as well? Or is that header hard to set in situations we care about?

Fang-: You're right, we can just use the jam mime type here.

joemfb (commented):
this will do

joemfb (approved):

Issue #6477, opened 2023-04-13
gen: `+vats %blocking` doesn't actually show blocking desks


tinnus-napbus:
**Describe the bug**
`+vats %blocking` includes desks with zests `%held` and `%dead` aka suspended and suspended until next kernel update. These do not block kernel updates so should not be listed in `%blocking` output. 

**To Reproduce**
Steps to reproduce the behaviour:
1. run `+vats %blocking` on a ship with non-running desks with outdated kelvins
2. observe they are included in blocking output despite not blocking

**Expected behaviour**
Only desks that are `%live` AND who have no compatible pending updates AND whose current waft is incompatible should be included in `%blocking` output.

**System (please supply the following information, if relevant):**
 - OS: linux64
 - Vere 2.1, latest develop

**Notify maintainers**

@hoclun-rigsep

tinnus-napbus:
here's an example on a fakezod at 414 with a suspended %foo desk at 415:

```
> +vats %blocking
  %foo
  /sys/kelvin:      [%zuse 415]
  base hash:        0vo.9f28r.java2.qnmqf.b3t1l.65mlu.8qsql.hq0cv.fpmdr.vuo1e.iehht
  %cz hash:         0vo.9f28r.java2.qnmqf.b3t1l.65mlu.8qsql.hq0cv.fpmdr.vuo1e.iehht
  app status:       suspended
  force on:         ~
  force off:        ~
  publishing ship:  ~
  updates:          local
  source ship:      ~
  source desk:      ~
  source aeon:      ~
  kids desk:        ~
  pending updates:  ~
::
  %kids
  %kids %cz hash:     0v13.gb7b2.v1h24.upgv3.o743l.jjc2k.8ellh.8keur.4osot.1si86.rll9k
::
```

it also appears to incorrectly pick up %kids

hoclun-rigsep:
Thank you for explaining the logic; I had mindlessly taken the various criteria from the previous implementation. I will implement.

PR #6478, opened 2023-04-13
gen: extend +vats and |suspend to take lists of desks

### LINKED ISSUE #6470: gen: +vats doesn't handle the %kids presentation well
From [this comment](https://github.com/urbit/urbit/pull/6395#issuecomment-1504437791) on #6395:

`+vats` doesn't handles the `%kids` presentation well:
```
+vats
<snip>
::
 %kids %kids %cz hash:     0v14.4jub3.fe4i1.5pfa6.do69s.fqkh7.tf6o9.g7b68.pgdeu.jjec8.u8517::
```
```
+vat %kids
%kids %cz hash:     0v14.4jub3.fe4i1.5pfa6.do69s.fqkh7.tf6o9.g7b68.pgdeu.jjec8.u8517
```

silnem2:
Extends +vats to take an optional list of one or more desks; if +vats is run with a list of desks and a =filt argument the output will contain only the members of the list that match the filter. Also removes the now-redundant +vat generator, updates |suspend to take a list of desks, and refactors +vats and related functions in /sur/hood. Resolves #6470 and #6477.


pkg/arvo/gen/vat.hoon
jalehman, line 1: Instead of removing +vat, I'd rather it output a deprecation notice and be refactored in terms of the newly extended `+vats`. That'll prevent breakage to anyone's workflow and teach them how to use the new capabilities in one fell swoop.

jalehman (commented):
This resolves my previous, although I'll defer to @belisarius222 on an actual code review.

belisarius222 (approved):
I tested this on a fakezod, and it all works right for me.

The non-verbose printing is still wonky, but that behavior is unchanged from the previous version of +vats, so it's not a regression.

belisarius222 (changes requested):
On second thought, the changes to `|suspend`, while I don't think they're risky, do imply a backward-incompatible change to the Kiln API, which someone might be depending on.  There are three options here:
1) wait until we have confirmation that all "systemically important" agents that are clients of Kiln's `%suspend` API are manually notified so they can change their use of this before release
2) split this into two PRs and merge the `|suspend` changes into a later Kelvin branch
3) revert the changes to Kiln's `%kiln-suspend` poke datatype, and instead change the `|suspend` generator to use a new `%kiln-suspend-many` mark, handled in a new `+poke-suspend-many` arm in Kiln.  `+poke-suspend` can be rewritten to call into that arm with a single-element list of desks to avoid code duplication.

Option 1 isn't great.  Options 2 and 3 are both acceptable.  I would prefer option 3, to reduce compatibility work for client apps.

PR #6479, opened 2023-04-13
groups: end of an era

arthyn:
This concludes an era. Old landscape got us to where we are. https://github.com/tloncorp/landscape will now take over the %landscape desk and %garden will be no longer. 

The only code moving over is s3, more details can be seen here: https://github.com/tloncorp/landscape/pull/142

@belisarius222 all references to %garden will need to be changed to %landscape, and people looking to do app dev will want to pull from `desk-dev` in the landscape repo.


jalehman:
@tinnus-napbus this more likely than not has implications for our documentation. 

When do you expect this to merge @arthyn ?

arthyn:
@jalehman sometime before the next kelvin burn, this will be synchronized with it so that we can do some renaming/moves while everything is suspended.

tinnus-napbus:
I will prepare a docs update PR and merge it when the update is deployed

yosoyubik:
Do we also want to remove the /~debug frontend folder? I guess it could be moved somewhere else, but probably it's the only folder that belongs in this repo?

arthyn:
I feel like it's probably right for it to stay in base, unless y'all stopped distributing it with the kernel for some reason.

arthyn:
we're finally ready

yosoyubik:
This PR removes /pkg/interface/dbug. We have two options here:

- restore it
- move it to its own repo (`https://github.com/urbit/dbug`)

arthyn:
personally I think it's own repo would be best so that it could be potentially improved and worked on at it's own pace and not clutter the kernel repo, but I can also understand keeping it colocated here for ease of access for kernel updates.

arthyn:
@yosoyubik what's the difference between `pkg/interface/dbug` and `pkg/arvo/app/debug`?

ashelkovnykov:
@arthyn I think the difference is:
- `app/dbug` = agent for debugging other agents
- `lib/dbug` = library for allowing the `%dbug` agent to reach into your agent(s)
- `interface/dbug` = frontend for using `%dbug` agent from your browser at `<ship URL>/~debug/`

arthyn:
@ashelkovnykov https://github.com/urbit/urbit/tree/develop/pkg/arvo/app/debug seems to be some kind of frontend

tacryt-socryp:
I'm ready for the era to end already. :(

arthyn:
me too, so do we want to only keep the dbug interface here and everything else can go?

arthyn:
@yosoyubik @belisarius222 I think this is good to go now

PR #6480, opened 2023-04-13
clay: add requests permission checks

### LINKED ISSUE #6401: clay: default desk is public
# Issue Description:
New Desks, as created by `|new-desk %desk-name`, are "public" and can be merged despite `%treaty`'s settings.

## Steps to Reproduce:
1. `|new-desk %test`
2. from another ship, `|merge %test ~their %test`
3. enjoy

## Expected Behavior:
Not that.

MarcusMiguel:
Resolves #6401

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -3765,6 +3765,10 @@
     ::  %next is just %mult with one path, so we pretend %next = %mult here.
     ::
         ?(%next %mult)
+      ?.  ?=(~ for)
+      ::  reject if foreign (doesn't work over the network)
+      ::
+        [[~ ~] ..park]
       ::  because %mult requests need to wait on multiple files for each
       ::  revision that needs to be checked for changes, we keep two
       ::  cache maps.  {old} is the revision at {(dec aeon)}, {new} is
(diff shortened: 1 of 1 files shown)


joemfb (commented):
This looks right to me.

It seems like the `?(%next %mult)` case in `+try-fill-sub` also needs a permission check -- I'm not sure if these cases had just been missed, or if there's some obstacle.

@MarcusMiguel have you tried this on a livenet ship?

/cc @Fang- @philipcmonk

MarcusMiguel:
While testing this i discovered what appears to be an issue with `Ames`. I'm digging into that before resuming the tests.

belisarius222:
@MarcusMiguel any luck with the Ames issue?

MarcusMiguel:
Changed the return to `[~ ~]` when the permission is denied for `%many`, which means that we will not store the request to try to fill later (this is the current behaviour in the `%sing` handler).

This PR looks good, I don't think it will break anything.

 I'm going to continue debugging some weird behavior I'm noticing in `Ames`.

jalehman:
@philipcmonk when you get a sec, mind giving this a look?

philipcmonk (approved):
This looks correct to me.

PR #6490, opened 2023-04-18
lull, eyre: actually send events after clogged channel reconnect

### LINKED ISSUE #6311: eyre: connection requires frequent refresh to maintain

pkova:
Fixes #6311.

Eyre maintains a queue of unacknowledged events for each channel. If a SSE client stops acking for more than thirty seconds after 50 unacknowledged events, eyre will "clog" the channel. A clogged channel unsubscribes from all Gall apps but keeps the unacked events around for 12 hours in case of a reconnect. If a client reconnects he should receive some events on the clogged channel in addition to `%kick`/`%quit` events notifying him of the need to resubscribe.

This has never worked as far as I can tell.

What happens is that the client makes a get request on reconnect and everything seems fine, but no events ever appear from any subscriptions that have been clogged. This is not an unusual situation, it's very likely that closing the laptop lid for 10 minutes runs into this problem.

The bug is as follows:

* A channel gets clogged, all Gall subscriptions get removed.
* When a client reconnects unacknowledged events need to be serialized to json.
* To serialize events we need to know which desk the event came from so we can invoke the proper `/mar` file.
* But the channel is clogged, so the information about the desk is long gone:
https://github.com/urbit/urbit/blob/888bad35514bb6332a63476f37472d851f5b62ce/pkg/arvo/sys/vane/eyre.hoon#L1659 https://github.com/urbit/urbit/blob/888bad35514bb6332a63476f37472d851f5b62ce/pkg/arvo/sys/lull.hoon#L1564

This PR addresses the issue by storing the desk directly in the unacknowledged events queue.


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1687,70 +1690,49 @@
...
     ::
     ++  sign-to-channel-event
-      |=  =sign:agent:gall
-      ^-  channel-event
-      ?.  ?=(%fact -.sign)  sign
-      [%fact [p q.q]:cage.sign]
+      |=  [=sign:agent:gall =channel request-id=@ud]
+      ^-  (unit channel-event)
+      ?.  ?=(%fact -.sign)  `sign
+      ?~  desk=(app-to-desk channel request-id)  ~
```
pkova, line 1696: As discussed with @Fang-, we might be able to just assert here. I did notice that crashing here is a road straight to brick city with `behn: timer failed; queue blocked` so I decided to keep this unitized.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1635,7 +1636,7 @@
       ::  if we're clogged, or we ran into an event we can't serialize,
       ::  kill this gall subscription.
       ::
-      =*  msg=tape  "on {(trip channel-id)} for {(trip request-id)}"
+      =*  msg=tape  "on {(trip channel-id)} for {<request-id>}"
```
pkova, line 1639: Strange that these were never caught.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1687,70 +1690,49 @@
...
-      ~%  %sign-to-json  ..part  ~
-      |=  [=channel request-id=@ud =sign:agent:gall]
       ^-  (unit (quip move json))
       ::  for facts, we try to convert the result to json
       ::
       =/  [from=(unit [=desk =mark]) jsyn=(unit sign:agent:gall)]
-        ?.  ?=(%fact -.sign)       [~ `sign]
-        ?:  ?=(%json p.cage.sign)  [~ `sign]
+        ?.  ?=(%fact -.event)       [~ `event]
+        ?:  ?=(%json mark.event)    [~ `[%fact %json !>((json noun.event))]]
```
pkova, line 1722: Doing `(json noun.event)` is a little bit scary here. Is it possible that we end up with a non-json noun in this code path? Not sure how and where the `sign:agent:gall` is verified.

joemfb:
@pkova I want to flag that there are significant, relevant %eyre changes on `next/kelvin/413` (noun channels, a couple migrations), and that branch will be promoted to `develop` shortly. It's probably worth not waiting, and just rebasing this onto there.

pkova:
Done. I also realized that this bug isn't as dire as I originally thought. Kick events have been able to be sent out on clogged channels because they don't need `/mar` files to serialize:
https://github.com/urbit/urbit/blob/f3d2df087a62f0bad1edcbac9b9698bd42252941/pkg/arvo/sys/vane/eyre.hoon#L1802-L1806

This is still better though.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2868,6 +2882,46 @@
...
+               cors-registry
+               connections
+               authentication-state
+               new-channel-state
+               domains
+               http-config
+               ports
+               outgoing-duct
+               0
+    ==     ==
```
Fang-, line 2923: I'd put in a little comment explaining why we're kicking all the subscriptions here. Also, kindly consider `%=` instead of `:*`. Something along the lines of:

```suggestion
    =;  new-channel-sessions
      %=  $
        -.old                            %~2023.4.19
        session.channel-state.old        new-channel-sessions
        outgoing-duct.channel-state.old  [outgoing-duct.channel-state.old 0]
      ==
    %-  ~(run by session.channel-state.old)
    |=  c=channel-3
    =;  new-events
      c(events new-events, unacked ~, subscriptions ~)
    =|  events=(qeu [id=@ud request-id=@ud =channel-event])
    =/  l  ~(tap in ~(key by subscriptions.c))
    |-
    ?~  l  events
    %=  $
      l          t.l
      next-id.c  +(next-id.c)
      events     (~(put to events) [next-id.c i.l %kick ~])
    ==
```

Maybe there's a way to make this cleaner, but I always find explicitly doing the changes only is more legible than reconstructing the whole state.

Fang- (commented):
Again, good catch on this bug. I think the implementation here is mostly correct. Couple of comments though:

pkg/arvo/sys/vane/eyre.hoon
pkova: Unitized this too because of out of band conversation. Avoiding crashing in eyre is good because it lessens the chance of poorly written userspace applications bricking your ship.

pkg/arvo/sys/vane/eyre.hoon
pkova: Done

pkg/arvo/sys/vane/eyre.hoon
pkova: Done.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2868,6 +2887,33 @@
...
+    =|  events=(qeu [id=@ud request-id=@ud =channel-event])
+    =/  l  ~(tap in ~(key by subscriptions.c))
+    |-
+    ?~  l  events
+    %=  $
+      l          t.l
+      next-id.c  +(next-id.c)
+      events     (~(put to events) [next-id.c i.l %kick ~])
+    ==
+      %~2023.4.19
```
joemfb, line 2916: Style nit: this should have a comment separate line before it.

pkova: Done

jalehman:
> Done. I also realized that this bug isn't as dire as I originally thought. Kick events have been able to be sent out on clogged channels because they don't need `/mar` files to serialize:
> 
> ...
> 
> This is still better though.

@pkova what does this improve upon then? Just curious.

PR #6499, opened 2023-04-24
gall: task and scry provenance WIP

tinnus-napbus:
based off of https://github.com/urbit/urbit/pull/4095

this adds an extra `pro=path` field to an agent's `bowl`: the `path` denotes which vane the event came from and the agent if it came from gall. This is populated in all cases - scries/pokes/watches/signs, etc. For a gall agent the path is `/gall/[agent-name]`, for other vanes it's just `/[vane]`, if it's indeterminate it's just `/`. The provenance of events from the local ship can be trusted, the provenance of `deal`s over ames cannot be - it just says whatever the other ship claims.

This adds an additional `pov=path` arg after `lyc=gang` in arvo's `roof`/`room`/`roof`/`rook` scry types, `++look` and relevant machinery. The `++scry` vane arm also takes `pov`. Arvo's external `++peek` doesn't. At this stage, all vanes just use their own name eg `/eyre` except for gall which uses `/gall/[agent-name]` for scries coming from within its agents.

The version of `$ames-request` in gall is bumped to `%1` from its current `%0` version & a `prov` field is added to each case. Since current gall simply crashes without an error message on any `$ames-request-all` that's not `%0`, talking to old galls is tricky. Current approach is to change `contacts.state` from a simple `(set ship)` to a `map` containing info about which ames-request version to use, and versions will be negotiated. We ask the other ship what version(s) they support if we don't know, defaulting to sending %0 until we hear back from them, retrying at most once a day per ship until they're upgraded. If a %1 gall receives a %0 `$ames-request-all`, it just uses an empty `/` provenance`.

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -109,6 +109,7 @@
   |$  [a]
   $~  =>(~ |~(* ~))
   $-  $:  lyc=gang                                      ::  leakset
+          pov=path                                      ::  provenance
           vis=view                                      ::  perspective
           bem=beam                                      ::  path
       ==                                                ::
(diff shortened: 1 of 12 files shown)


tinnus-napbus:
closed in favour of local-only provenance https://github.com/urbit/urbit/pull/6605

Issue #6500, opened 2023-04-24
~/login path should redirect logged in users to /


mcevoypeter:
When I navigate to https://my.ship.domain/~/login after having signed in, I'm prompted to re-enter my access key whereas I'd like to be redirected to https://my.ship.domain.

**Steps to reproduce**
- Navigate to https://my.ship.domain.
- Log in.
- Navigate to https://my.ship.domain/~/login.

**Expected behavior**
https://my.ship.domain/~/login redirects to https://my.ship.domain when I'm already signed in.

mcevoypeter:
cc: @jamesacklin

jamesacklin:
Tagging in @arthyn for investigation

Fang-:
Closed by #6505, which was released yesterday.

PR #6505, opened 2023-04-24
eyre: if logged in and accessing login page, redirect

arthyn:
We don't have a redirect for the login page itself if you're already authenticated. This rectifies that and does the appropriate redirection. Addresses #6500


arthyn:
do I need to update the pill in this PR or is that done after merging?

pkova (approved):
LGTM, I like the 303.

Issue #6307, opened 2023-02-08
Postfix comment on a `=*`  fails to compile


ryjm:
Steps to reproduce:

1.  make a postfix doccords style comment on the same line as a `=*` declaration, i.e in apps at the line `=*  state  -`
2. try to commit
3. witness strange error 

![image](https://user-images.githubusercontent.com/2065299/217615733-2f49937e-d40c-47e6-9a91-c51a34a420fe.png)

ryjm:
Oof I just noticed this also happens on `+*` declarations, which means commenting your core aliases will break compilation.

Seems like this could be related to #6052?

ryjm:
```
[ %brcb
                                  p=[%dbug p=[p=/ q=[p=[p=75 q=7] q=[p=75 q=17]]] q=[%bcts p=term=%bowl q=[%dbug p=[p=/ q=[p=[p=75 q=8] q=[p=75 q=17]]] q=[%like p=~[%bowl] q=~[~[%gall]]]]]]
                                    q
                                  [ i=[p=%this q=[%dbug p=[p=/ q=[p=[p=76 q=17] q=[p=76 q=18]]] q=[%cnts p=~[[%.y p=1]] q=~]]]
                                      t
                                    [ i=[p=%lsp-core q=[%dbug p=[p=/ q=[p=[p=77 q=17] q=[p=77 q=19]]] q=[%cnts p=~[[%.y p=7]] q=~]]]
                                        t
                                      ~[
                                        [ p=%lsp
                                            q
                                          [ %dbug
                                            p=[p=/ q=[p=[p=78 q=17] q=[p=78 q=35]]]
                                              q
                                            [ %cnsg
                                              p=~[[%.y p=1]]
                                              q=[%dbug p=[p=/ q=[p=[p=78 q=21] q=[p=78 q=29]]] q=[%wing p=~[%lsp-core]]]
                                              r=[i=[%dbug p=[p=/ q=[p=[p=78 q=30] q=[p=78 q=34]]] q=[%wing p=~[%bowl]]] t=~]
                                            ]
                                            
```

example ast with no doccords

```
with comments:

 [ %brcb
                                  p=[%dbug p=[p=/ q=[p=[p=75 q=7] q=[p=75 q=17]]] q=[%bcts p=term=%bowl q=[%dbug p=[p=/ q=[p=[p=75 q=8] q=[p=75 q=17]]] q=[%like p=~[%bowl] q=~[~[%gall]]]]]]
                                    q
[...]

PR #6509, opened 2023-04-25
turn off doccord parsing for +*, =*

drbeefsupreme:
see #6307

this turns doccord parsing off for `=*` and `+*`, which was not intended to
be allowed for the first release, but it ended up not compiling if
postfix doccords were put on either of these.

the right way to fix this is to actually implement doccord parsing for
`=*` and `+*`, but at least turning them off seems to fix the compile error here.

94ba944188 changes the signature of `+vang` to add the `doc` flag from
`+vast` and thus changes all the callsites to `+vang` as well.

bff9a0aaa7 changes the parsers for `+*` and `=*`. i considered adding
another arm for `(vang bug wer |)` to avoid repeating it 4 times, but since
this is supposed to be a band-aid and not a permanent fix, i thought it
might be better to avoid adding in new arms that might later be forgotten
when the band-aid is removed. not sure what the right style choice is here.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -11425,11 +11425,12 @@
 ::  +vang: set +vast params
 ::
 ::    bug: debug mode
-::    wer: where we are
 ::    doc: doccord parsing
+::    wer: where we are
+::
 ++  vang
-  |=  [bug=? wer=path doc=?]
-  %*(. vast bug bug, wer wer, doc doc)
+  |=  [f=$@(? [bug=? doc=?]) wer=path]
+  %*(. vast bug ?@(f f bug.f), doc ?@(f & doc.f), wer wer)
 ::
 ++  vast                                                ::  main parsing core
   =+  [bug=`?`| wer=*path doc=`?`&]
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -118,7 +118,7 @@
   |=  [our=ship dir=beam]
   |%
   ++  default-app         %hood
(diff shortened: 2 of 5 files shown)


drbeefsupreme:
i forked off of `master` instead of `develop`, looks like it brought in https://github.com/urbit/urbit/pull/6509/commits/c2c36f42f524fd588c10aaf927832ad5bc922fd9 for free

drbeefsupreme:
> i thought it might be better to avoid adding in new arms that might later be forgotten when the band-aid is removed. not sure what the right style choice is here.

he said, after including an unused arm to balance things out. took out `+lien`

pkg/arvo/sys/hoon.hoon
```diff
@@ -13412,6 +13417,9 @@
     ++  hunk  (most mush loan)                          ::  gapped specs
     ++  jump  ;~(pose leap:docs gap)                    ::  gap before docs
     ++  loaf  ?:(tol tall wide)                         ::  hoon
+    ++  loll  ?:  tol                                   ::  hoon without docs
+                tall:(vang bug wer |)
+              wide:(vang bug wer |)
```
joemfb, line 13422: No need to call `+vang` again here, you can just use `tall(doc |)` or `wide(doc |)`.

joemfb (changes requested):
This makes sense and looks good. One nit:

joemfb (approved):
this will do

belisarius222:
@drbeefsupreme This includes an important bugfix, but the change to the `+vang` sample type, while virtuous, is backward-incompatible.  Do you mind splitting the change to `+vang` into a separate PR?  That way we could merge this PR today and get it deployed with kelvin 413.

belisarius222 (approved):

PR #6515, opened 2023-04-26
fine: next steps

### LINKED ISSUE #6504: fine: next steps
building on #5788 / #6406

- [WIP](https://github.com/urbit/urbit/pull/6526) aqua: test fine integration, run aqua tests
- spider: improve incremental scry state lifecycle mgmt
- spider: use (and maybe fix) `+take-tune` in strandio
- deq: implement +apt
- gall: track `%keen`, send `%yawn` on uninstall/nuke
- hoon: write thru `keens` alias should work (currently using `keens.peer-state`)
- clay: implement permissions checks in scry (move out of fine %hunk namespace)
- fine: `%drip` `%tune`

more to come

wicrum-wicrun:
Closes #6504 

- [ ] aqua: test fine integration, run aqua tests
- [x] spider: improve incremental scry state lifecycle mgmt (WIP, pending review)
- [x] spider: use (and maybe fix) +take-tune in strandio
- [ ] deq: implement +apt
- [x] gall
  - [x] track %keen/%tune
  - [x] send %yawn on nuke
  - [x] send %yawn on uninstall
  - [ ] ~expose open %keens in the bowl~
- [ ] hoon: write thru keens alias should work (currently using keens.peer-state)
- [ ] clay: implement permissions checks in scry (move out of fine %hunk namespace)
- [ ] fine: %drip %tune
- [ ] have Fine perform peer discovery, possibly by having the publisher respond to the requester's galaxy

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1311,6 +1311,7 @@
           :*  wex=boat.yoke                           ::  outgoing
               sup=bitt.yoke                           ::  incoming
               sky=(~(run by sky.yoke) tail)           ::  bindings
+              ken=ken.yoke                            ::  requests
           ==                                          ::
           :*  act=change.stats.yoke                   ::  tick
               eny=eny.stats.yoke                      ::  nonce
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -958,8 +958,18 @@
         moves        moves
       ==
     ::
+    ++  ap-yawn-all
+      ^-  (list card:agent)
+      %-  zing
+      %+  turn  ~(tap by ken.yoke)
+      |=  [=spar:ames wyz=(set wire)]
+      %+  turn  ~(tap in wyz)
+      |=  =wire
+      [%pass wire %arvo %a %yawn spar]
+    ::
     ++  ap-idle
       ?:  ?=(%| -.agent.yoke)  ap-core
+      =>  (ap-ingest ~ |.([ap-yawn-all p.agent.yoke]))
       ap-core(agent.yoke |+on-save:ap-agent-core)
     ::
     ++  ap-nuke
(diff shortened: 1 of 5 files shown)


wicrum-wicrun:
Gall now tracks outstanding keens, removes them on tune and sends yawns on nuke. Forgot to do it on uninstall but should be easy. Exposing them in the bowl also isn't done. Those are for another day though.

yosoyubik:
@joemfb ptal https://github.com/urbit/urbit/pull/6515/commits/cfc1a585598d32e85876f0946c4fac0b92d875a3

belisarius222:
>  expose open %keens in the bowl

that should be reserved for 412k

belisarius222:
let's limit this PR to just the changes made so far -- we'll handle the rest of the todos in separate PRs

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -958,8 +958,18 @@
...
+      %-  zing
+      %+  turn  ~(tap by ken.yoke)
+      |=  [=spar:ames wyz=(set wire)]
+      %+  turn  ~(tap in wyz)
+      |=  =wire
+      [%pass wire %arvo %a %yawn spar]
+    ::
     ++  ap-idle
       ?:  ?=(%| -.agent.yoke)  ap-core
+      =>  (ap-ingest ~ |.([ap-yawn-all p.agent.yoke]))
```
belisarius222, line 972: If an agent gets `%idle`d then woken back up again, does it get notified?  I don't think it does.  Without knowing that it got turned off and back on, if we kill its outbound `keen`s when it gets idled, the agent won't know they got killed, so it might sit there forever without knowing it needs to re-send those requests.

belisarius222: I'm discussing this with @joemfb right now.  I think the best way to handle idling is for Gall to cancel the outgoing scry requests on idle, hold onto the request wires and spars in its state, and then when the agent is turned back on, re-send those requests to Ames.  This should work fine as long as Gall doesn't crash if it happens to receive a scry response in between idling the agent and the agent turning back on -- although Ames itself should handle that, since it remembers which outgoing scry requests are live.

That way the agent doesn't need to be notified, and we don't artificially kill perfectly legitimate requests, but we also avoid sending superfluous network traffic to repeat requests forever on behalf of an idled agent.

wicrum-wicrun: I actually thought about this too, but figured that since we were planning to include keens in the bowl it was fine. The agent could discover it that way. But I agree that this is more polite. (And of course, we're not doing keens in the bowl at this time)

joemfb (approved):
this will do

joemfb:
@yosoyubik at some point, we might want better support multiple, concurrent `%keen` requests in strandio, without requiring unique wires. All the necessary information is in the spar, so this should be pretty straightforward to build. These lower-level implementations seem fine for now.

PR #6518, opened 2023-04-27
ames: make +abet pure

yosoyubik:
WIP

https://github.com/urbit/urbit/pull/6403 got closed, probably due to its previous inclusion in the Remote Scry PR, so we reopen it (as a draft).


yosoyubik:
Tested this on fakezod with toy pub/sub agents and re-subscriptions—the main change in this PR—work: (i.e. flows are corked/deleted appropriately). 

Tests on a live ship also work (minus network lag) and the only issue I found is that if you join a group, leave, and then search for the same group host again, the previously joined group shows up with a "Go", but it's not present on the sidebar, and the host doesn't have the joiner listed as member, and %ames packets are flowing correctly (e.g. no pending-vane acks, hear-last-in-progress messages... etc). My guess is that this is a transient issue with %groups, present already in the live network, and not introduced by this PR (clicking on "Go" shows a bunch of peek-errors in %dojo, and searching for the group host again shows the option to join the group, which happens successfully)

yosoyubik:
This fixes the explicit uses of the impure +abet pattern (updating a part of the state, different that the one the +abet core is in charge off). There are implicit uses of the impure +abet in the |pump and |sink cores by calling each other:

- calling the message sink inside the message pump: 
  - https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/ames.hoon#L3351
- calling the message pump inside the message sink:
  - https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/ames.hoon#L3955
  - https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/ames.hoon#L4023

yosoyubik:
The issue described here https://github.com/urbit/urbit/pull/6518#issuecomment-1536384023 is present on %413 (I have been able to reproduce it on master/%414, for some ships, but not all), it looks like this error is due to a crash happening here: https://github.com/tloncorp/landscape-apps/blob/develop/desk/app/groups.hoon#L695.

Screenshot 2023-05-08 at 4 36 37 PM

In any case, the issue pre-exists the changes of this PR.

yosoyubik:
Based on the [three places](https://github.com/urbit/urbit/pull/6518#issuecomment-1536402164) were |mu and |mi call into each other, the only possible scenario where this could lead to a problem would be if the deferred calls touch a part of the state (in the next arvo event) that needs to be updated in the current arvo event. 

Following the code paths for the different scenarios where this could happen, this seems to not be a case that would arise

When sinking a naxplanation message in `+ha-nack:mi`, we defer the notification of this message being naxplained to the next arvo event, and ack the naxplanation accordingly in the current event. The defer notification to the message pump will be on the target-bone (e.g. `nack-bone=0b10`, `target-bone=(mix 0b10 0b10)=0b0` -> a `%plea`). When +pump-done is called, unless this flow is in closing (in which case, we will no-op) the client vane will be notified on the error.

The other place where the message |sink calls the message |pump is when we sink a %plea with the ok=%.n flag (iirc due to arvo injecting the same event that crashed with a `dud=(unit goof)`, due to a crash on a plea fragment). In this case, sinking this plea will send a naxplanation message in the next arvo event, and a nack in the current one, putting this message in the `nax.state` set. When the ack for the naxplanation comes back, the message pump (in `+pump-done`) will notify the message |sink (in the next arvo event) to drop this nack, which in turn will delete the message from the `nax.state` set.

I've tested this PR in livenet ships that behave without any notable incident for the normal cases of joining, leaving, rejoining groups—which tests most of the cases considered here (acks, nacks, corks...etc). The only case that I still haven't tested is the one caused by a crash on a %plea fragment—I have a ship that was seeing `%bad-foreign-update` crashes so will check if the fragment crashed was also seeing in that scenario.

yosoyubik:
I've simulated a crash on a fragment %plea by just crashing in on-hear, which does the first part of the flow described [here](https://github.com/urbit/urbit/pull/6518#issuecomment-1545863798):

> sinking this plea will send a naxplanation message in the next arvo event, and a nack in the current one, putting this message in the nax.state set.

This other half however is wrong:

> When the ack for the naxplanation comes back, the message pump (in +pump-done) will notify the message |sink (in the next arvo event) to drop this nack, which in turn will delete the message from the `nax.state` set.

When we hear the ack for the naxplanation—from the point if view of the publisher, sent on e.g. bone 15—`+pump-done` is acting on that bone, which triggers the "ack is on "subscription update"" code path, that just no-ops. For %watch %pleas, this is fine because when the %cork is handled, the flow is deleted (or if we crash on the cork, there will be just one nack in `nax.state`) but for %poke %pleas, the set will keep increasing. `nax.state` is only used when sending dupe message-acks—if this message is in the `nax.state` set, and we hear it again, it will send a nack instead.

In any case, it seems that we would need to handle the nack-trace (bone=10) case in +pump-done before the subscription-update one (bone=01), but unsure of the implications of deleting the message from the `nax.state` if we hear another message again, and ack instead of nack—this would imply that the crash happens in a different way when handling the duplicate case?

yosoyubik:
Testing this on a live ship. Nothing shows up, corks, leaving subscriptions... etc work. 

When running on https://github.com/urbit/urbit/pull/6518/commits/7d153ba99ae4839ec386308dd358ac7e331d7a7f and `!!` on `+hear:mi` if `ok=%.y`—basically to simulate a crash only on a fragment, but not on the injected-by-arvo error—the "nack-trace with blank .error for security" is called (in `+ha-plea:handle-sink:mi`, and when we hear the ack for it in `+pump-done:mu` we capture that correctly and clear nack from the `|sink`.

yosoyubik:
I'm not sure how we can replace the hijacking of `+on-plea` to handle these intra-ames calls in the next event. `+on-plea`, because of the presence of `payload=*` in the associated task, gives us the flexibility to pass any type of message.

The cleaner solution here I think would be to have a specific task to handle this calls (`%deep`, as in, it comes from the deep `%ames`?), that explicitly expects one of [the five types](https://github.com/urbit/urbit/blob/i/6103/abet-pure/pkg/arvo/sys/vane/ames.hoon#L2049-L2053) of deferred moves. I've implemented that solution here: https://github.com/urbit/urbit/commit/fec56d94fdb405a8184078fd8f409c2a94df12b4

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2181,6 +2157,49 @@
...
+        ::
+        ?>  ?=([@ =ship *] deep)
+        =/  ship-state  (~(get by peers.ames-state) ship.deep)
+        ?>  ?=([~ %known *] ship-state)
+        =+  peer-core=(abed-peer:pe ship.deep +.u.ship-state)
+        |^  ?-  -.deep
+          %nack  abet:(send-nack-trace [nack-bone message-blob]:deep)
+          %sink  abet:(sink-naxplanation [target-bone naxplanation]:deep)
+          %drop  abet:(clear-nack [nack-bone message-num]:deep)
+          %cork  =~((cork-bone bone.deep) (emit duct %give %done ~))
```
belisarius222, line 2176: better to move the effect emission to inside +on-cork-flow

belisarius222 (approved):
One little comment to address, but this would be good even without that.

PR #6521, opened 2023-04-27
json: used optimized arm for mime -> json

arthyn:
On the tin, this uses the more standard `de:json:html` instead of the internal parsing arm.

### DIFF
--- a/pkg/base-dev/mar/json.hoon
+++ b/pkg/base-dev/mar/json.hoon
@@ -17,7 +17,7 @@
   --
 ++  grab
   |%                                                    ::  convert from
-  ++  mime  |=([p=mite q=octs] (fall (de:json:html (@t q.q)) *^json))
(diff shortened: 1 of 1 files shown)


pkg/base-dev/mar/json.hoon
```diff
@@ -17,7 +17,7 @@
   --
 ++  grab
   |%                                                    ::  convert from
-  ++  mime  |=([p=mite q=octs] (fall (rush (@t q.q) apex:de:json) *^json))
+  ++  mime  |=([p=mite q=octs] (fall (de:json:html (@t q.q)) *^json))
```
joemfb, line 20:
```suggestion
  ++  mime  |=([p=mite q=octs] (fall (de:json (@t q.q)) *^json))
```

joemfb (changes requested):
Thanks!

One nit: the `=,  html` makes this mark a little confusing (`*^json`), but we should be consistent or remove it.

joemfb (approved):

PR #6524, opened 2023-04-28
clay: Make commits fast again

philipcmonk:
Fixes an unreleased bug where commits are much slower than they should be.

I recorded this debugging session in an effort to demonstrate "absolute debugging", which is available [here](https://youtu.be/d1VZcaGWBRM).  If you're going to watch that, I recommend not looking at the diff here -- you want to actually wonder yourself.  If you just want to skip to an explanation of the bug, jump to [1:38:00](https://youtu.be/d1VZcaGWBRM?t=5880), or the eureka moment at [1:32:29](https://youtu.be/d1VZcaGWBRM?t=5549).  The commit message here reflects that as well.


belisarius222 (approved):
I tested this on a fakezod.

PR #6526, opened 2023-04-28
aqua: wip fix ph tests for %413

yosoyubik:
This fixes the issues that disallowed running any tests, but some tests are very slow, or just «behn» (e.g. -ph-keen)


yosoyubik:
Tests involving %clay merges are also very also (or just hang).

joemfb:
The %fine scry in `:aqua` is also wrong: https://github.com/urbit/urbit/blob/701115c7fbe803b248125872ba8351bc48de83b0/pkg/arvo/app/aqua.hoon#L685-L687

yosoyubik:
The latest commit fixes `-ph-keen`. `-ph-child-sync` that does a `|merge` (and syncs `/sur/aquarium.hoon` using `%fine` via a multi-packet response, work) but other threads that deal with %clay (e.g. `-ph-breach-sync`) seem to be very slow, stuck in `+handle-unix-effect` in `/ted/ames/aqua`——I'm running some of this in the background to see if they actually finish.

PR #6532, opened 2023-04-29
vats: force multiline rendering, improve performance

joemfb:
This PR pre-renders the tanks in `+vats` to ensure that they're always wrapped across multiple lines (this is an expedient hack to avoid more custom rendering logic). It also imports the kelvin mark in `sur/hood` to ensure that it's not being built inside scrys, where it can't be cached. And it sorts desks and force on/off lists alphabetically.

/cc @jfranklin9000

PR #6535, opened 2023-05-01
vats: fixes +vat

joemfb:
In #6532 I changed `+report-vat` to produce a `tang` instead of a `tank`, but I forgot that `+vat` still existed.


arthyn (approved):
lgtm

PR #6542, opened 2023-05-02
lull,ames: use `mop` instead of `pha` in `.wan.keens`

wicrum-wicrun:
The implementation of `pha` in lull is wrong and leads to very poor performance. At the moment we don't have a path to fix it, so this PR changes `.wan.keens` to use a `mop` instead. `pha` and `deq` are still kept around for state migration purposes.


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -4027,13 +4312,13 @@
             =.  peer-state    (update-peer-route her peer-state)
             =.  metrics.keen  on-timeout:fi-gauge
             =^  want=(unit want)  wan.keen
-              (pop-left:fi-deq wan.keen)
+              (fall (mole |.([`val.head rest]:(pop:fi-mop wan.keen))) `wan.keen)
```
belisarius222, line 4315: I don't think we should need to virtualize this.  Jumping in and out of a road here could slow things down, so let's look for an alternative that doesn't need virtualization.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -4397,10 +4682,30 @@
...
+    %-  ~(run by peers.old)
+    |=  old=ship-state-13
+    ?:  ?=(%alien -.old)  old
+    old(keens (~(run by keens.old) keen-state-13-to-14))
+  ::
+  ++  keen-state-13-to-14
+    |=  old=keen-state-13
+    ^-  keen-state
+    =-  old(wan wan)
+    %^    (dip-left:(deq:keen-state-13 want) ,wan=((mop @ud want) lte))
```
belisarius222, line 4703: Can we not just `+tap` this to a list of key-value pairs?  I don't think we need the full expressiveness of `+dip`.

pkg/arvo/sys/vane/ames.hoon
belisarius222: That should be less expensive than a road.

belisarius222 (dismissed):
LGTM.  I will merge once some manual testing has been completed.

Issue #6514, opened 2023-04-26
No mark conversions between `mime` <> `noun` <> `jam`


tadad:
At Uqbar we've found that we conversions between jammed nouns and mimes for putting them into unix easily. You can already do this from the dojo using `.foo/noun bar` though doing this from clay would be very convenient.
@joemfb suggested that we do it as follows:
in `mar/noun/hoon`, add conversions to/from `jam`
in `mar/jam/hoon`, adding `++grab` for `%mime`

This implies that whenever you want to put a noun into unix it should be as a `.jam` file, not a `.noun` file

ilyakooo0:
I think this can be closed as complete?

PR #6544, opened 2023-05-02
feat: noun to mime

morelazers:
Addresses #6514.

I tried to add grow to jam in noun, and to mime in jam, but this solution seems to be simpler.


belisarius222:
Please change the mimetype to `application/x-urb-jam` to match `mar/jam/hoon`.  Then this will be good to merge.

joemfb (approved):
this will do

joemfb (approved):
this will do

PR #6545, opened 2023-05-02
clay: virtualize parsing to workaround runaway memoization

joemfb:
This significantly reduces memory usage during userspace rebuilds. It's not exactly clear why; there's some kind of effective space leak in memoization while parsing. This is an easy, low-risk, stop-gap.


philipcmonk (approved):
Definitely interested in learning more about the numbers we're seeing, but this does seem to make a big difference.

On a moon reproduction with only the standard desks, this reduces total ephemeral memory usage for the upgrade event from 808MB to 111MB.

I would not expect the number of desks to have a significant effect on the amount of memory caused by this issue.

Issue #6489, opened 2023-04-17
Apps do not update to latest after binary update


dchem:
**Description**:

After a binary update from `1.9` to `2.1` (skipped two binary releases) and updating `%zuse` from `418` to `414`, applications show pending updates status for older `%zuse` versions. Apps do not update to latest due to pending updates.

`+vats` show `pending updates` with older versions of `%zuse` even though current `%base` is `%zuse 414`: e.g. for an app `%talk`, `pending updates: ~[[%zuse 416] [%zuse 415]]` 

```
> +vat %garden
/sys/kelvin:      [%zuse 417] [%zuse 416] [%zuse 415] [%zuse 414]
base hash:        ~[0v1g.qsorb.rbi48.7julk.q0tgr.556cq.5lr0s.kkl1r.qbc4a.3qiuj.4uobn 0ve.toutn.4g341.qmh43.ckmms.il03a.ijb99.vvos9.7j0mo.fkm02.3qgld]
%cz hash:         0ve.toutn.4g341.qmh43.ckmms.il03a.ijb99.vvos9.7j0mo.fkm02.3qgld
app status:       running
force on:         ~
force off:        ~
publishing ship:  [~ ~mister-dister-dozzod-dozzod]
updates:          remote
source ship:      ~zod
source desk:      %garden
source aeon:      32
kids desk:        ~
pending updates:  ~[[%zuse 416] [%zuse 415]]
```

Other `+vats` show similar output as https://github.com/urbit/urbit/issues/6209#issuecomment-1427130168

**Actions taken**:

1. `|pack` -> no effect
2. `|bump` -> `%base already up-to-date`
3. Suspend, install from new source, revive -> no effect
4. Uninstall, install -> no effect, shows `suspended until next update`

**Actions not yet taken**:

1. `|nuke %<app>` and reinstall

**Questions**
* Is this an unrecoverable state?
* Is not upgrading the binary at each version update and skipping 2-3 release causing this issue?

ajlamarc:
I think this is related to what @belisarius222 [described here](https://gist.github.com/belisarius222/83a3c3af5497fb54cdc1aaa3be36a9bd).


Is an app that signals kelvins 417-414 valid?

dchem:
Doing the command did remove the app dependencies.

Some default apps (`bitcoin`, `garden`, `landscape`) would update to latest version, while others like `talk` and `groups` still shows older version even though it no longer show "pending update"

dchem:
Update:

[...]

Issue #6468, opened 2023-04-11
App activation being incorrectly blocked by old dependency  (Ford?)


ngzax:
**Describe the bug**
I can't activate %pals. Even though my %base desk is at [%zuse 414], %pals, which reports compatibility with [%zuse 414] reports being blocked by needing [%zuse 416]. This makes no sense to me. I have other desks in this same state.

**To Reproduce**
1. suspend %pals
2. it won't un-suspend

**Expected behaviour**
%pals activates.

**Screenshots**
```
> +vat %base
/sys/kelvin:      [%zuse 414]
base hash:        0v14.4jub3.fe4i1.5pfa6.do69s.fqkh7.tf6o9.g7b68.pgdeu.jjec8.u8517
%cz hash:         0v14.4jub3.fe4i1.5pfa6.do69s.fqkh7.tf6o9.g7b68.pgdeu.jjec8.u8517
app status:       running
force on:         ~[[%eth-watcher %.y]]
force off:        ~
publishing ship:  ~
updates:          remote
source ship:      ~darlur
source desk:      %kids
source aeon:      129
kids desk:        %kids
pending updates:  ~

> +vat %pals
/sys/kelvin:      [%zuse 417] [%zuse 416] [%zuse 415] [%zuse 414]
base hash:        0v1u.3iaik.f1r49.9flat.pe2br.7g4j0.7secj.9leoq.idqdc.emp85.vfhmk
%cz hash:         0v1u.3iaik.f1r49.9flat.pe2br.7g4j0.7secj.9leoq.idqdc.emp85.vfhmk
app status:       suspended until next update
force on:         ~
force off:        ~
publishing ship:  ~
updates:          remote
source ship:      ~paldev
source desk:      %pals
source aeon:      28
kids desk:        ~
pending updates:  ~[[%zuse 416]]
```

**System (please supply the following information, if relevant):**
 - OS: linux64
 - Vere 2.1

jfranklin9000:
Try `|bump`?

ngzax:
Thanks, but that just gives me `kiln: %base already up-to-date`

jalehman:
This makes no sense to me either. @belisarius222 we should dig into this sooner than later.

ngzax:
Thanks @jalehman. Let me know if you need me to run any commands or diagnostics to help figure out what the issue is.

ngzax:
and just to be clear, I do have some apps/desks that are OK and some that are not. ¯\(°_o)/¯

[...]

PR #6548, opened 2023-05-03
clay: on update, remove all previous pending updates

### LINKED ISSUE #6537: clay: Pending updates for past kelvins
I've seen a couple reports of this phenomenon, where %base is on eg 414K but has a pending update for 416K.

As background, this is not supposed to be able to happen, because every time we add a kelvin to the pending update list, we filter that it's less than the current one.  Then, when we apply a commit, we always remove the new current kelvin from the pending update list.  Search `wic.dom` in clay to convince yourself of this.

My theory is that these ships have "jumped" a kelvin -- eg, they upgraded from 417K straight to 415K without first installing 416K.  This is generally not supposed to be possible -- if we have pending updates for two kelvins, we apply them consecutively.  However, perhaps there is some combination of blocking apps, outdated runtimes, and switching OTA sources that can cause it.

One theory is:
- %base is at 417K, with pending updates for 416K and 415K not blocked by any apps, but the updates haven't gone through because the runtime was out of date
- upgrade the runtime, but do not kick off the upgrade by clicking in the UI or running `|bump`
- run `|ota ~sponsor` again (sponsor is on 415K)

I believe this would cause us to attempt to upgrade directly to 415K, and it would not clear the pending update for 416K.

I will attempt to reproduce this, but if true, this suggests a few solutions:
- In +park, we delete the current kelvin from the pending update list.  We could also delete all previous kelvins.  This is the smallest cut (a few lines of low-risk code), but accepts that non-consecutive kelvin upgrades can happen.
- We could require that commits to %base have a kelvin no newer than the oldest pending update.  In this case, we would need to direct the user to run `|bump` to stepwise upgrade, or attempt to do it automatically.  This would prevent such "jumps", unless the jump happens "all at once" -- eg you're on 417K and your ship has been not receiving OTAs long enough that when you restart you receive 415K directly.
- We could require that commits to %base have truly consecutive kelvins -- ie the new kelvin is exactly one less than the last one.  This would prevent jumps entirely, which would be a nice guarantee for developers, including kernel developers -- you would be guaranteed to not have to deal with state prior to the last kelvin.

The last two options are an attempt to enforce some form of "stepwisdom" (guaranteed stepwise upgrades) for kelvins.  While thinking about this, I wrote an attack on stepwise upgrades, which I've [published separately](https://gist.github.com/philipcmonk/de5ba03b3ea733387fd13b758062cfce).  Suffice to say, I believe we should reject stepwisdom and use the first solution.

philipcmonk:
Fixes #6537, see discussion there for alternatives.

Tested on a fakezod by reproducing that bug, then applying this fix and attempting to reproduce again.

For existing ships that have past pending updates, this will clear those old updates as soon as they get another kelvin update, so I wouldn't do anything more specific.  Although it shouldn't be necessary, if someone wants to clear their state ahead of time, they can use the debug command `|pass [%c %stir %stay %desk ~]` where `%desk` is the desk to clear all pending updates on.

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -1818,7 +1818,11 @@
       =?  ..park  !?=(%base syd)  wick                  ::  [wick]
       %-  (slog leaf+"clay: wait-for-kelvin, {<[need=zuse/zuse have=kel]>}" ~)
       tare                                              ::  [tare] >
-    =.  wic.dom  (~(del by wic.dom) zuse+zuse)
+    =.  wic.dom
+      %-  ~(gas by *(map weft ^yoki))
+      %+  skip  ~(tap by wic.dom)
+      |=  [w=weft ^yoki]
+      (gte num.w zuse)
     ::
     =/  old-yaki
       ?:  =(0 let.dom)


jfranklin9000:
It fixes these too?
https://github.com/urbit/urbit/issues/6468
https://github.com/urbit/urbit/issues/6489

joemfb (commented):
@philipcmonk it seems like there really isn't support for `sys.kelvin` referencing a layer other than %zuse. If that's right and intended, this looks good to me.

philipcmonk:
I believe it'll resolve those issues as well, yes.

Yes, there are a few places where we compare numbers like this, and all of
them would need to change if we decide to support non-zuse kelvin s.

joemfb (approved):

belisarius222 (approved):

Issue #6552, opened 2023-05-04
ames: ship-state scry crashes if ship is not in peers


midden-fabler:
https://github.com/urbit/urbit/blob/24467176f62a8c7f599eb0a536a5caea3272e830/pkg/arvo/sys/vane/ames.hoon#L4790C1-L4798

Return a `(unit ship-state)`?

PR #6554, opened 2023-05-04
ames: ship-state scry returns a unit

midden-fabler:
Resolves [#6552](https://github.com/urbit/urbit/issues/6552).
no longer crash if scrying for a ship that's not in peers.


midden-fabler:
For uptime monitoring, I need to access data from ship-state frequently. For example, to see if ames is trying to contact a ship, I must first scry for peers and check if that ship is in peers, then scry for ship-state and analyse that.
I can do what I need to, but the initial check is an ergonomic annoyance, and something of a "gotcha" to be guarded against anywhere ship-state is requested.

PR #6558, opened 2023-05-05
i/6541/prop test

### LINKED ISSUE #6541: Property testing for Hoon
Implement property testing for Hoon, according to the following grant: https://urbit.org/grants/kNock-property-testing

Suggested terminilogy:

- A property that some core respects a particular contract that should always hold is called a `%fate`. It is given as a gate.
- A sample that checks the %fate os called a `%quiz`.
- If there is any %quiz that causes the %fate to return `%.n` then we say that the core being tested defies (`%defy`) its fate. We also say that a certain %quiz defies a %fate.
- If a %fate is truly a contract that always holds, i.e. no %quiz exists that defies the fate, then the %fate is `%heed`ed.

Basic design idea:

(How to wrap the tester in a suitable API for programmers to call is a question we defer. It might be that it should be integrated with the `-test` thread or run its own thread or be used via a generator.)

- The tester is a door.
- The sample of the door is a a core `quizzer` whose arms specify gates for generating random quizes.
- The main arm of the tester takes entropy, a `size` and the %fate to be tested (as a vase), and produces a random %quiz using either the supplied quizzer or the default quzzer, supplied with the library.
- The built-in quizzer contains arms for many standard types: arbitrary nouns, atoms, cells, faces and forks. With time there may also be special arms for different atom auras.
- Core and hold types are not supported by default, but the user may name any type with `$+`, giving it a type hint, and supplying an arm in the quizzer with a corresponding name. If a type is a hint, the tester will look for a corresponding arm in the supplied quizzer, then in the default quizzer, and fail if it cannot find one.
- Some common types, for example as lists, sets and maps, are named with hints and have supplied arms in the default quizzer.

hjorthjort:
### Description

Resolves #6541.

Work in progress -- this is mostly request for comments.

Introduces `quiz`, property testing library and generator.

This implements a basic test runner supporting `%atom` `%face` and `%cell` types as gate samples. Other types may be part of the sample but will simply be accepting whatever is already in the sample of the passed gate (usually the default sample) instead of randomized.

Currently uses `~&` but reporting should probably be made cleaner.

### Related

https://urbit.org/grants/kNock-property-testing


pkg/arvo/gen/quiz.hoon
```diff
@@ -0,0 +1,7 @@
+/+  quiz
+:-  %say
+|=  [[now=@da eny=@uvJ bec=beak] [[vax=vase ~] ~]]
+:: Default to using 0 entropy and no special quizzers.
+=+  qui=~(. quiz [eny 100])
+=+  che=^-(? (check:qui vax))
+[%noun che]
```
wicrum-wicrun, line 7: Nit but since you're not doing anything with the intermediate values i'd personally prefer fewer names to keep track of.

```suggestion
[%noun (~(check quiz eny 100) vax)]
```

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,38 @@
+|_  [eny=@uv runs=@ud]
+++  check  |=  [vax=vase]
+  =.  runs  ?:(=(0 runs) 100 runs)
```
wicrum-wicrun, line 3: `=?` is made for this.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,38 @@
...
+  =+  size=1
+  =+  sax=(slot 6 vax)
+  =+  run-i=0
+  =+  rng=~(. og eny)
+  |-
+  =+  sam=(~(fill quiz [size rng]) sax)
+  =+  res=(slam vax sam)
+  ^-  ?
```
wicrum-wicrun, line 11: Typical to put this on the same line as the barhep.

wicrum-wicrun (commented):
Just some code style comments. Structure/semantics wise this looks good to me. Of course eventually you'll want support for minimizing counterexamples and all that but just as a demo of wiring everything up i think this looks very good!

pkg/arvo/lib/quiz.hoon
wicrum-wicrun: That's recarding the kethep, github doesn't display the line i'm referring to very clearly.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,54 @@
...
+      ~&  success-runs+run-i
+      %.y
+  ?:  =(+:res %.y)
+    :: Arbitrarily chosen growth pace.
+    =+  new-size=(add +(size) (div (mul size 2) 21))
+    $(run-i +(run-i), rng +:(rads:rng 1), size new-size)
+  ~&  fail-with-sam+q.sam
+  %.n
+++  quiz  |_  [size=@ud rng=_og]
+  ++  split-rng  |.  ^+  [og og]
```
wicrum-wicrun, line 21: You don't need `|.` here. The arm won't be evaluated until you pull it anyway.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,54 @@
...
+    =^  bits-1  rng  (raws:rng bit-size)
+    [~(. og bits-1) ~(. og (raw:rng bit-size))]
+  ++  fill  |=  [sax=vase]
+    ^+  sax
+    =+  new-rng=+:(rads:rng 1)
+    ?+  p.sax  ~&(warn-unfill-sam+`type`p.sax sax)
+      %noun            sax(q (gen-noun))
+      [%atom p=* q=~]  sax(q (gen-atom p.p.sax))
+      [%atom *]        sax
+      [%cell p=* q=*]  =+  [rng-1 rng-2]=(split-rng)
```
wicrum-wicrun, line 32: If you remove `|.` from the `split-rng` arm you can remove the parentheses around the call.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,54 @@
...
+                       %=  sax
+                          q  :-  q:(fill(rng rng-2) (slot 2 sax))
+                                 q:(fill(rng rng-1) (slot 3 sax))
+                      ==
+      [%face p=* q=*]  sax(q q:(fill [p=q.p.sax q=q.sax]))
+    ==
+  ++  gen-noun  |.
```
wicrum-wicrun, line 39: Same thing here.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,57 @@
+|_  [eny=@uv runs=@ud]
+++  check  |=  [vax=vase]
```
belisarius222, line 2: style: `|=` should go on the next line, indented once fromt the `++` line, like this:
```hoon
++  check
  |=  vax=vase
  =?  runs  =(0 runs)  100
  ...
```
Also, the sample does not need enclosing `[]`s.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,61 @@
...
+      [%atom *]        sax
+      [%cell p=* q=*]  =+  [rng-1 rng-2]=split-rng
+                       %=  sax
+                          q  :-  q:(fill(rng rng-2) (slot 2 sax))
+                                 q:(fill(rng rng-1) (slot 3 sax))
+                      ==
+      [%face p=* q=*]  sax(q q:(fill [p=q.p.sax q=q.sax]))
+    ==
+  ++  gen-noun
+    |.
```
belisarius222, line 42: You don't need the extra core here.  Instead, the `|.` can close around `start-size`.
```hoon
  ++  gen-noun
    =+  start-size=size
    |.  ^-  noun
    ?:  (lte size 1)
      (rad:rng start-size)
    ...
```

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,61 @@
...
+    ?:  (lte size 1)
+      (rad:rng start-size)  :: leafs should be able to make large atoms.
+    =^  ran  rng  (rads:rng 3)
+    ?:  =(0 ran) :: 1/3 chance for a leaf.
+      (rad:rng start-size)
+    ?:  =(1 ran)  :: 1/3 chance for a identical subtrees.
+      =+  subtree=$(size (div size 2))
+      [subtree subtree]
+    =+  [rng-1 rng-2]=split-rng :: 1/3 chance for different subtrees. 
+    :-  $(size (div size 2), rng rng-1)
```
belisarius222, line 54: style: this should either use backstep indentation like this:
```hoon
:-  $(size ...)
$(size ...)
```
or use a `:*` if you really want both lines at the same indentation level, like this:
```hoon
:*  $(size ...)
    $(size ...)
==
```
This is important stylistically because it means the last line of the body of the arm begins one indentation level beyond the `++  gen-atom` on the next line.  This keeps the top-level organization of the file easier to scan quickly.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,61 @@
...
+    |=  sax=vase
+    ^+  sax
+    =+  new-rng=+:(rads:rng 1)
+    ?+  p.sax  ~&(warn-unfill-sam+`type`p.sax sax)
+      %noun            sax(q (gen-noun))
+      [%atom p=* q=~]  sax(q (gen-atom p.p.sax))
+      [%atom *]        sax
+      [%cell p=* q=*]  =+  [rng-1 rng-2]=split-rng
+                       %=  sax
+                          q  :-  q:(fill(rng rng-2) (slot 2 sax))
```
belisarius222, line 36: just like later, either backstep or use `:*`

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,61 @@
...
+      %.y
+  =+  sam=(~(fill quiz [size rng]) sax)
+  =+  res=(slam vax sam)
+  ?:  =(+:res %.y)
+    :: Arbitrarily chosen growth pace.
+    =+  new-size=(add +(size) (div (mul size 2) 21))
+    $(run-i +(run-i), rng +:(rads:rng 1), size new-size)
+  ~&  fail-with-sam+q.sam
+  %.n
+++  quiz  |_  [size=@ud rng=_og]
```
belisarius222, line 21:
```hoon
++  quiz
  |_  [size=@ud rng=_og]
  ++  split-rng
    ...
```

belisarius222:
Minor nits, but this looks good.

hjorthjort:
I'm currently making heavy use of `~&` for output which isn't ideal. I think returning tapes or tanks is a more sensible forward approach, but it becomes a little messier and may be worth saving for a future PR. I'd also be happy to do a design call on neat ways to do this.

Next steps are implementing dropped samples (already in pipeline, was prety simple). After that custom sample generation, implementing `%fork` types, shrinking and some combinators for sample generation -- they all kind of interact with each other.

hjorthjort:
@jalehman thanks for checking in. @wicrum-wicrun made some points about simplifying user-generated inputs that I want to implement this week. I could back this PR up to [223b692](https://github.com/urbit/urbit/pull/6558/commits/223b692dc6647bb5ef816fb37451555140b0aea8) if we want to merge it in a simple stage. Or I could implement the new input generation and ask you to re-review from there.

hjorthjort:
Alright, it was simpler than I thought. This is ready for another review, all that is missing is some tests for shrinking.

hjorthjort:
Pushed with rebase on `develop`

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,180 @@
...
+  =+  drop=0
+  |-  ^-  ?
+  ?:  =(run-i runs)
+      ~&  [success-runs+run-i drops+drop]
+      %.y
+  =/  sam=vase
+    ?~  give
+      (~(fill quiz [size rng]) sax)
+    [p=p.sax q=(u.give size rng)]
+  =+  res=(slum q.vax q.sam)
```
dr-frmr, line 21: I'm working with this library and am wondering if `+mink` works where `+slum` is used here. That would have the benefit of allowing the library to test gates which can crash, and ultimately allow one to run many tests and then print the failing ones along with their inputs at the end.

pkg/arvo/lib/quiz.hoon
```diff
@@ -0,0 +1,195 @@
...
+    =.  tries  +(tries)
+    ?.  =(tries 1.000)
+      $(rng next-rng, size new-size)
+    ~&  %tired
+    (fin run-i drop)
+  =.  tried  (~(put in tried) sam)
+  =+  res=(slum q.vax q.sam)
+  =?  drop  =(res %drop)  +(drop)
+  ?:  |(=(res %drop) =(res %.y))
+    :: Arbitrarily chosen growth pace.
```
hjorthjort, line 38: Move this up

pkg/arvo/lib/quiz.hoon
hjorthjort: I'm not a big fan of the idea of keeping going after a failure is found, partly because runtimes can often increase with size. I prefer just returning (and shrinking) as soon an error is encountered.

That being so, we can probably get away with just using `~|`

```
=+  ^=  res
~| [%quiz-crash-on-input q.sam]
(slum q.vax q.sam)
```

hjorthjort: actually this won't work well, since the program would still crash. virtualizing with `mink` (or `mong` in this case, which just wraps `mink`) makes sense.

Issue #6342, opened 2023-02-17
`%name` binding broke backwards compatibility 


tadad:
[This line here](https://github.com/urbit/urbit/commit/d8ff95db8491f711a5dfaffb7964683fec281e3d#diff-01fd10efd392903ad3f191b1caddb67dba7d57d11997894173665d20f0d42477R1556) in a recently merged PR was a breaking change because the return type of eyre's `/=bindings=` scry changed. This happened to break a scry in `%blog` where I was using the following code:
```
    ?-  -.action
      %gen             [%s (crip "desk: {<desk.generator.action>}")]
      %app             [%s (crip "app: {<app.action>}")]
      %authentication  [%s '%authentication']
      %logout          [%s '%logout']
      %channel         [%s '%channel']
      %scry            [%s '%scry']
      :: %name            [%s '%name'] :: TODO need to add this between releases
      %four-oh-four    [%s '%four-oh-four']
    ==
```
I believe a breaking change means we should burn a kelvin

pkova:
Addressed by #6346.

PR #6561, opened 2023-05-05
eyre: session identities for all

Fang-:
Eyre now grants "guest identities" to unauthenticated clients, and lets them make use of its channels system. For motivation, considerations and other details, please see [the open eyre proposal](https://gist.github.com/Fang-/7f142f73a585cf86f0ecf4f6f7da8a17). For a quick summary, read 0fee4ce.

**This will require a kelvin bump for release, because it contains the following breaking changes:**
- When Eyre watches (on `/http-response/[eyre-id]`) or pokes (with the `%handle-http-request` mark) agents, the value in `src.bowl` is no longer guaranteed to be equal to `our.bowl`. Instead, it will reflect the identity associated with the request.
  - In theory this shouldn't be a problem, because private pages are fine crashing on public access, and public pages shouldn't be asserting this equivalence, but turns out it's not unprecedented: 744dea2
- A new `$action:eyre` tag, `%host`, has been added. (See also #6342.)

Notably, the `authenticated.inbound-request` flag maintains its existing behavior, where it is true iff the requester is logged in as the local ship.  
Given that `src.bowl` can now also tell you this, it's redundant, but we maintain it for backwards compatibility. We may want to mark is for deprecation at some point.

Remaining work:
- [ ] Make the sessions list in the debug dashboard look slightly better.
- [ ] Maybe give the login page a little "current identity" indicator or similar. Probably no use providing a "proceed as guest" link, considering right now you'd only commonly end up on the login page by redirection.
- [ ] Patch the Eyre tests to... account for us sending cookies all of the time now. Tried my hand at this, it's not pretty, should probably just forcefully refactor the tests at this point.
- [ ] Non-blocking, but shouldn't be too hard to get js-http-api ready to support the guest usage pattern.


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -1177,9 +1178,101 @@
...
       ::
-      (lte now expiry-time.u.session)
+      |-
+      =;  nom=@p
+        ?.  =(our nom)  nom
+        $(eny (shas %next-name eny))
+      =/  hat  (end 3^8 (fein:ob our))
+      %^  cat  3  hat
+      %+  rsh  [3 (met 3 hat)]
+      (~(raw og (shas %fake-name eny)) 128)
```
joemfb, line 1217: I think this would be clearer if it concatenated, then truncated. But it works fine as written.

joemfb (approved):
this will do

joemfb:
I fixed CI, and merged `develop` back into `next/kelvin/412`. A push here will update the diff.

pkg/arvo/sys/vane/eyre.hoon
joemfb: If you're looking for a trivial push to update the diff .... `(end 3^8 foo)` is `(end 6 foo)`, and `(end 3^16 foo)` is `(end 7 foo)`.

PR #6563, opened 2023-05-09
eyre: refactor tests

philipcmonk:
This shaves off 1000 lines of testing code while maintaining the same tests.  It reduces boilerplate by introducing "mare", a monad for testing Eyre.  It's very simple (just maintains the current state of Eyre and the current time), but it's easier to build helper functions in this form, and that reduces the immense quantities of copy-and-paste that were in the old tests.  What's there now could surely be improved further, but I think this is a good start.

The underlying mare machinery is not really specific to Eyre, so it would be straightforward to apply this strategy to other vanes.  The work is in creating appropriate helper functions for each vane.  Eyre is undergoing work, so that's the only one I've changed here.  Further, it's not clear that this is the ultimate solution to unit testing vanes. The resulting code is IMO clearer than before, but I wouldn't say it's *clear*.

I've maintained the order of tests to preserve a clean diff.  Additionally, this is targeting m/the-open-eyre because it was spurred by improvements @Fang- made to the tests there, but if that branch gets significantly delayed, it will be worth porting these changes straight to develop.


Fang- (approved):
Sweet, these are good improvements. The line count/boilerplate reduction alone is worth a lot, and this seems probably easier to maintain too. Thank you!

Issue #6571, opened 2023-05-10
gen: `+vats` doesn't work if there's an empty desk recognized by `%cd`


midden-fabler:
Somehow an empty %hark desk was created, which breaks `+vats`.
```
~sampel-midden-fabler:dojo> +vats\
bail: 4
! /cw/~sampel-midden-fabler/hark/~2023.5.10..03.44.14..5feb
[%dojo-lame %arvo-response]
```

joemfb:
I suspect this was caused by #6568.

PR #6574, opened 2023-05-10
clay: allow reads at the null tako

joemfb:
This PR relaxes a constraint introduced in #6568, which is too restrictive for existing use cases.


joemfb:
I've confirmed that this fixes #6571

belisarius222 (approved):

Issue #6579, opened 2023-05-10
behn: bail: %exit


jalehman:
My ship died and, upon attempting to restart, this is what happens:

```
urbit 2.3-95fd7f2
boot: home is /urbit/wolref-podlex
loom: mapped 2048MB
lite: arvo formula 2a2274c9
lite: core 4bb376f0
lite: final state 4bb376f0
loom: mapped 4096MB
boot: protected loom
live: loaded: GB/2.896.527.360
boot: installed 657 jets
---------------- playback starting ----------------
pier: replaying events 2356994909-2356994920
pier: (2356994920): play: done
---------------- playback complete ----------------
vere: checking version compatibility
ames: live on 30298
conn: listening on /urbit/wolref-podlex/.urb/conn.sock
http: web interface live on http://localhost:21969
http: loopback live on http://localhost:12323
pier (2356994931): live
conn: fyrd 0v0
behn: bail 1

[%poke %crud]
  wire
/ fine
  behn
  wake
  ~tocrex-holpen
  c
  x
  0v1u.up5ua.eamt5.ous41.ccndo.s2urh.d74vv.tdug5.oief9.5v1ut.qo7b6
  inet2023
  sys
  kelvin
bar-stack=[i=[i=//newt/0vb.gtnit t=~] t=[i=[i=//behn t=~] t=~]]
take: failed
/sys/vane/ames/hoon:<[4.549 3].[4.568 20]>
/sys/vane/ames/hoon:<[4.550 3].[4.568 20]>
/sys/vane/ames/hoon:<[4.553 3].[4.568 20]>
/sys/vane/ames/hoon:<[4.555 3].[4.568 20]>
/sys/vane/ames/hoon:<[4.556 5].[4.566 7]>
/sys/vane/ames/hoon:<[4.557 5].[4.566 7]>
/sys/vane/ames/hoon:<[4.561 24].[4.561 65]>
/sys/vane/ames/hoon:<[2.169 9].[2.219 74]>
/sys/vane/ames/hoon:<[2.170 9].[2.219 74]>
/sys/vane/ames/hoon:<[2.184 9].[2.219 74]>
/sys/vane/ames/hoon:<[2.185 11].[2.204 13]>
/sys/vane/ames/hoon:<[2.190 11].[2.204 13]>
/sys/vane/ames/hoon:<[2.194 11].[2.204 13]>
/sys/vane/ames/hoon:<[2.195 11].[2.204 13]>
[...]

PR #6580, opened 2023-05-10
ames: on-take-wake no-op if not path for keen

yosoyubik:
Fixes https://github.com/urbit/urbit/issues/6579


yosoyubik:
For unrecoverable ships, I have made a jam.file with the commit that contains the fix for `%ames`.

It can be downloaded here: https://yosoyubik.fra1.digitaloceanspaces.com/fine-fix.jam.

To use it, restart your urbit with: `YOUR_URBIT/.run -I fine-fix.jam`

joemfb (approved):
this will do

Issue #6584, opened 2023-05-11
feature request: generator to scry for rift and life in ames state


belisarius222:
Right now you have to scry out the `$peer-state` from Ames, then filter it, which is a bit annoying.

PR #6586, opened 2023-05-12
gen: add +ames-keys generator

yosoyubik:
Fixes https://github.com/urbit/urbit/issues/6584


belisarius222 (approved):
This could unlink the dojo if you ask about a ship we haven't talked to yet, but it's still an improvement over not having this generator.

PR #6592, opened 2023-05-16
hoon: add pin and awl arms to test and trim subaxes

eamsden:
`+pin`: test if `b` is a sub-axis of `a`
`+awl`: give the remainder axis of `b` under `a` (crash if `b` isn't a sub-axis of `a`)


pkg/arvo/sys/hoon.hoon
```diff
@@ -226,6 +226,33 @@
...
+  ::
+  ::  checks whether {b} addresses a subtree of the tree addressed by {a}.
+  |=  [a=@ b=@]
+  ?<  =(0 a)
+  ?<  =(0 b)
+  |-  ^-  ?
+  ?:  =(a 1)  %&
+  ?:  =((cap a) (cap b))
+    $(a (mas a), b (mas b))
+  %|
```
joemfb, line 241: To ensure that the last sub-hoon is the largest, can you switch this to a `?.`? Also, this needs a single `::` line between the `+pin` and `+awl` arms.

pkg/arvo/sys/hoon.hoon
```diff
@@ -226,6 +226,33 @@
...
+  ::  computes the remainder of axis {b} when navigating to {a}.
+  ::  (crashes if not `(pin a b)`)
+  |=  [a=@ b=@]
+  ?<  =(0 a)
+  ?<  =(0 b)
+  |-  ^-  @
+  ?:  =(a 1)  b
+  ?:  =((cap a) (cap b))
+    $(a (mas a), b (mas b))
+  !!
```
joemfb, line 254: This should be a `?>`.

pkg/arvo/sys/hoon.hoon
```diff
@@ -226,6 +226,33 @@
...
+  ?<  =(0 a)
+  ?<  =(0 b)
+  |-  ^-  ?
+  ?:  =(a 1)  %&
+  ?:  =((cap a) (cap b))
+    $(a (mas a), b (mas b))
+  %|
+++  awl
+  ~/  %awl
+  ::    axis after axis
```
joemfb, line 244: The should also be a blank `::` line after this "doccord summary".

joemfb (changes requested):

eamsden:
just realized `+awl` isn't cvc... don't have any replacement name ideas

pkg/arvo/sys/hoon.hoon
```diff
@@ -226,6 +226,33 @@
...
+++  pin
+  ~/  %pin
+  ::    sub axis test
+  ::
+  ::  checks whether {b} addresses a subtree of the tree addressed by {a}.
+  |=  [a=@ b=@]
+  ?<  =(0 a)
+  ?<  =(0 b)
+  |-  ^-  ?
+  ?:  =(a 1)  %&
```
joemfb, line 238: I think `+pin` should be total over non-zero atoms. For that to be the case, this should be

```
?:  =(a 1)  &
?:  =(b 1)  |
```

joemfb (changes requested):

joemfb (approved):
These look right to me. They're not urgent, and probably won't go into the next release. They're not particularly useful without jets (which I'd be happy to write). Let's defer to @belisarius222 as to when/where to merge.

this will do

joemfb:
@eamsden do you expect to always use these gates together? would it be better to combine them into a single gate producing `(unit axis)`?

eamsden:
> @eamsden do you expect to always use these gates together? would it be better to combine them into a single gate producing `(unit axis)`?

No I expect to use `+pin` by itself quite frequently. `+hub` was frankly only included because it seemed a shame not to when you're already doing all the math.

ashelkovnykov (approved):
Just making sure I understand correctly: `hub` is `mas of b relative to a`, correct?

eamsden:
> Just making sure I understand correctly: `hub` is `mas of b relative to a`, correct?

Roughly yes. Instead of taking off 1 bit it takes off however many bits are in `a` (not counting the MSB).

Notably though it crashes if `b` is not a subaxis of `a`.

eamsden:
@jalehman note that these likely depend on Joe's jets in urbit/vere#397

PR #6598, opened 2023-05-18
eyre: eauth, cross-ship authentication

Fang-:
Eyre now allows people to log in with their own identities on other's ships. For considerations and other details, please see [the mirage proposal](https://gist.github.com/Fang-/41ed84b2a6dd96ca67c6a5dbda1fd35d). For a quick overview of the changes, read 8579b6c.

**This will require a kelvin bump for release, because it contains the following breaking changes:**
- A new `$action:eyre` tag, `%eauth`, has been added.
- The types of `$authentication-state:eyre` and `$identity:eyre` have changed.

This builds on top of the work from #6561, the semantic changes to `src.bowl` described there apply for this work as well. The `authenticated.inbound-request` flag remains the old-school, deprecated-but-it-works way to see if the request comes from our authenticated selves.

Remaining work:
- [x] Make sure we handle all the error cases wrt Ames messages correctly. (Completed, but sort of dependent on #6612.)
- [x] Make the local<->eauth toggle on the login page prettier.
- [x] Add styling to the eauth confirmation page.
- [x] Provide an actual affordance for updating the user-hardcoded eauth hostname/URL.
- [x] Update tests to build again.
- [x] Add tests for the eauth flow.
- [x] Rework to be initiated from the client-side, as abuse protection.
- [x] If this can get merged before the changes from #6561 get released, merge the state migration logic.


joemfb (commented):
I've read this closely, it all looks pretty reasonable. Is there anything in particular that you'd like scrutinized?

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2900,6 +3618,92 @@
...
+        ::NOTE  we would *really* like to be calling on-fail:client here, but
+        ::      we can't, because we don't know the nonce. we *can* say:
+        ::      - if %open was nacked, we will just time out the attempt.
+        ::        - but! we could be showing the user an error sooner...
+        ::      - if %shut was nacked, we optimistically deleted it locally,
+        ::        and the remote session will eventually expire, assuming it
+        ::        goes unused.
+        ::        - but! this may not be the case if we were %shutting because
+        ::          our device/session/cookie was compromised...
+        ::          - but! good hosts shouldn't crash on %shuts at all...
```
joemfb, line 3655: The way to fix this would be to keep a per-ship `qeu` for outstanding plea's. This would let us correlate the nack to a nonce and eagerly send an error response.

The only "normal" case in which we'd get a nack here is if the server has no eauth-url (no one has ever logged into it) and the eauth-flow was begun on the client ship. This doesn't seem like a very priority, but it would be hard to add later, so it'd be best to track the state now.

pkg/arvo/sys/lull.hoon
```diff
@@ -2193,6 +2204,18 @@
     $:  ::  sessions: a mapping of session cookies to session information
         ::
         sessions=(map @uv session)
+        ::  visitors: in-progress incoming eauth flows
+        ::
+        visitors=(map @uv visitor)
+        ::  visiting: outgoing eauth sessions, completed or pending
+        ::
+        visiting=(map ship (map @uv portkey))
```
joemfb, line 2212: to track outstanding pleas: `(map ship (pair (map @uv portkey) (qeu @uv)))`

joemfb (changes requested):
After further discussion with @Fang-, it seems worth explicitly tracking outstanding plea's from the client side

joemfb (approved):
this will do

PR #6519, opened 2023-04-27
`%lick` vane

mopfel-winrux:
This is a PR to include a new vane called `%lick`. This vane allows `%gall` agents to open an IPC communication port on urth. This port speaks in jammed nouns of the form `[=mark =noun]`

The vane has three parts:

1. urbit/urbit modifications: This PR
2. urbit/vere modifications: https://github.com/urbit/vere/pull/358
3. userspace example: https://github.com/mopfel-winrux/lick-agent

TODO:

- [x] Change port name from `@tas` to `$path`
- [x] Write python code that connects to the agent-port and performs `jam`s and `cue`s on data
- [x] Write python code that connects to a HW device and acts as a middleman between the port and device
- [x] Gall should intercept all cards and prefix the agent name to the socket name. It should also remove the name for any `%soak`s

Add state to Gall to track which agent owns which socket (POSTPONED for gall rewrite)

Notes:

Useful links:
- https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/gall.hoon#L1133 for modifying card
- https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/gall.hoon#L602 for modifying soak
- https://github.com/urbit/urbit/blob/develop/pkg/arvo/sys/vane/gall.hoon#L1404 for modifying soak

Prefix agent to note and something to wire to note that its a lick command

Open Design Questions:
- Right now the owner of the ipc port is the last agent that `%spin`ed it. This should probably change to the first one.
- Any agent can `%spit` on an ipc port, even if they didn't %spin it. Do we want to gate `%spit`ing and `%shut`ing to only the agent that spun it?

PR #6605, opened 2023-05-22
gall: implement local provenance

tinnus-napbus:
This PR supercedes https://github.com/urbit/urbit/pull/6499. Having provenance work over the network required a new `$ames-request` version which necessitated negotiation between ships, which ultimately required outbound moves to remote ships to be queued up while version negotiation was underway. Queuing outbound moves introduced enormous complexity, and therefore the whole remote provenance aspect has been scrapped. Since you can't trust the provenance claimed by remote ships, this was already a controversial feature. Having provenance only for local pokes etc should still retain most of the usefulness, especially because it can be trusted. Being able to say "only `/gall/foo` aka local agent `%foo` can poke me" or "I won't accept requests from `/eyre` aka requests from front-ends" lets you lock down an app pretty well, especially with the forthcoming userspace permissions.

This adds an extra `sap=path` field to an agent's `bowl`: the `path` denotes which vane the event came from and the agent if it came from gall. For a local gall agent the path is `/gall/[agent-name]`, for other vanes it's just `/[vane]`, if it's indeterminate it's just `/`. In the case of deals from remote ships, the provenance is just `/`.

This adds an additional `pov=path` arg after `lyc=gang` in arvo's `roof`/`room`/`roof`/`rook` scry types, `++look` and relevant machinery. The `++scry` vane arm also takes `pov`. Arvo's external `++peek` doesn't. At this stage, all vanes just use their own name eg `/eyre` except for gall which uses `/gall/[agent-name]` for scries coming from within its agents.


pkg/arvo/sys/lull.hoon
```diff
@@ -462,7 +462,7 @@
 +$  rift  @ud                                           ::  ship continuity
 +$  mime  (pair mite octs)                              ::  mimetyped data
 +$  octs  (pair @ud @)                                  ::  octet-stream
-+$  sock  (pair ship ship)                              ::  outgoing [our his]
++$  sock  (trel ship ship path)                         ::  [src dest prov]
```
belisarius222, line 465: I'm inclined to think we'd be better off making a new type for this and leaving `$sock` as-is, to reduce the amount of work client code needs to change.

`$sack` could be the name for the new type, perhaps -- it's not used elsewhere afaik.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -630,14 +630,16 @@
         mo-core
       =/  app
         =/  =ship  (slav %p i.t.t.wire)
-        =/  =routes  [disclosing=~ attributing=ship]
+        =/  =routes  [disclosing=~ attributing=[ship /[-.sign-arvo]]]
```
belisarius222, line 633: I wonder if we'll have problems here with signs that get `%drip`'d through Behn, since using this logic they'll show up as coming from Behn despite coming from some other vane originally.

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -630,14 +630,16 @@
...
         (ap-abed:ap dap routes)
       ::
       =.  app  (ap-generic-take:app t.t.t.wire sign-arvo)
       ap-abet:app
     ?>  ?=([%out @ @ *] t.t.wire)
     =/  =ship  (slav %p i.t.t.t.wire)
-    =/  =routes  [disclosing=~ attributing=ship]
+    =/  other-agent  i.t.t.t.t.wire
```
belisarius222, line 640: Do we need to upgrade old wires to this new format somehow?

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1874,6 +1879,18 @@
   ::
   ~|  [%gall-call-failed duct hic]
   =/  =task  ((harden task) hic)
+  =/  prov=path
```
belisarius222, line 1882: Ah ok, I hadn't seen this when I commented "do we need to upgrade wires?"  Now my question is: do we need to upgrade wires in `+take` in addition to here in `+call` -- maybe in `+mo-handle-use` or `+mo-handle-sys`?

belisarius222:
I read through this just now.  Other than the comments I left, it all looks good to me.

pkg/arvo/sys/vane/gall.hoon
tinnus-napbus: existing wire format is sufficient afaict. Maybe I'm misunderstanding what you're suggesting here?

pkg/arvo/sys/lull.hoon
tinnus-napbus: added $sack as recommended

pkg/arvo/sys/vane/gall.hoon
belisarius222: I thought the wire format had changed, but since it hasn't, there's no issue.

joemfb:
This PR broke CI, fixed in [2bae99e](https://github.com/urbit/urbit/pull/6519/commits/2bae99ec2b9a5e22234d560c301b5bd52b905fdd)

PR #6606, opened 2023-05-22
ames: fix incorrect handling of acks in |pack

yosoyubik:
We were not properly handling acks for live packets in the packet pump, and were always instantiating the congestion control pump with the same number (the initial, with all packets) for every iteration.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -3635,6 +3635,11 @@
               =/  acc
                 :*  found=`?`%.n
                     resends=*(list static-fragment)
+                    ::  num-live is still present in pump-metrics but not used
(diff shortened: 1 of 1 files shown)


belisarius222 (dismissed):
This pattern is non-obvious, so it could use some comments describing why it's done this way.  It does make sense, though.

yosoyubik:
I pushed a version of this fix [here](https://github.com/urbit/urbit/commit/3554ab895d8ba989713c229f97b6ce1a8f4f8eca) that doesn't use the unused `num-live` in pump-metrics, removes it from the type definition in `%lull`, and includes the state update in `%ames`—compiles, but still untested for previous state migrations, and on livenet ships.

belisarius222 (approved):

PR #6607, opened 2023-05-22
ames: don't no-op if getting %ack on nack bone

yosoyubik:
Extracted from https://github.com/urbit/urbit/pull/6518/commits/7d153ba99ae4839ec386308dd358ac7e331d7a7f


belisarius222 (approved):
We'll want to do something to delete the space leak that this fixes, where ships were holding onto outbound naxplanations, in a different PR, though.

PR #6609, opened 2023-05-23
ames: remove num-live from pump-metrics

yosoyubik:
Same as https://github.com/urbit/urbit/pull/6606 but with num-live removed from `%lull`—and state migrations in `%ames`, compiles but untested for old state migrations, and livenet environments.


belisarius222:
This code looks good to me.  Is there a reason this is still a draft PR?

yosoyubik:
Marking as ready for review; just wanted to test previous state migration in %ames from old pills, and merging it into a live ship.

PR #6612, opened 2023-05-23
ames: correctly %lose a %boon we crashed on

### LINKED ISSUE #6611: ames: crashing on %boon receives the %boon again
The ames docs say the following:

> If the Arvo event that completed receipt of a %boon message crashes, Ames instead sends the client vane a %lost message indicating the %boon was missed.

Indeed, the implementation confirms that it notifies the client vane if we crashed on this message:

https://github.com/urbit/urbit/blob/77cfbe76c5b516c8b2362b594cf4394a1ec863cc/pkg/arvo/sys/vane/ames.hoon#L4011-L4030

During Eyre development, I'm intentionally crashing during my `%boon` handling. But I do not see the intended behavior described above. Instead, with `|ames-verb %msg`, and a `~&` in Eyre's `+take`, I see the following:

```
ames: ~nec: hear ~nec seq=1 1kb
ames: ~nec: sink boon [~nec bone=120 message-num=1]
[%eyre-take wire=/eauth/plea/0/~nec/0v5gj.e7p19.cna1m %boon dud=%.n]
ames: ~nec fragment crashed %exit
[[... eyre's stack trace from the crash ...]]
take: failed
bar-stack=[i=[i=//http-server/0vr.9jk7g/106/1 t=~] t=[i=[i=//ames t=~] t=~]]
wire=/eauth/plea/0/~nec/0v5gj.e7p19.cna1m
[%poke %hear]
ames: ~nec: hear ~nec seq=1 1kb
ames: ~nec: crashed on sink boon [~nec bone=120 message-num=1]
[%eyre-take wire=/eauth/plea/0/~nec/0v5gj.e7p19.cna1m %boon dud=%.n]
```

Clearly, Ames is detecting the crash. `crashed on sink boon` is accurate. And yet, it just sends us a `%boon` again.

Looking at the code snippet above again, a probable cause presents itself. PR incoming.

Fang-:
Previously, if we noticed `%boon` handling had caused a crash, we would transform any existing `%boon`s into `%lost`s, but still emit a new `%boon` for the message we crashed on.

Now, we make sure to just directly send a `%lost` if sending the `%boon` caused a crash. We drop the existing-moves transformation entirely, assuming it to vestigial from a brief glance at the blame/history. Reviewers, please confirm whether this is correct or not. (It must be, the tests still pass after all. (; )

Fixes #6611 (see there for more deets) and adds a test to make sure it doesn't occur again.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -4018,15 +4018,13 @@
                   ::
                   =/  dat  [her bone=bone message-num=message-num]
                   ?:(ok "sink boon {<dat>}" "crashed on sink boon {<dat>}")
-              =?  moves  !ok
-                ::  we previously crashed on this message; notify client vane
-                ::
-                %+  turn  moves
-                |=  =move
-                ?.  ?=([* %give %boon *] move)  move
-                [duct.move %give %lost ~]
+              ::  if we previously crashed on this message;
+              ::  notify client vane instead of making it try again
               ::
-              =.  peer-core  (pe-emit (got-duct bone) %give %boon message)
+              =.  peer-core
+                %+  pe-emit  (got-duct bone)
+                ?.  ok  [%give %lost ~]
+                [%give %boon message]
               ::  send ack unconditionally
               ::
               (call %done ok=%.y)
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -4018,13 +4018,14 @@
                   ::
                   =/  dat  [her bone=bone message-num=message-num]
                   ?:(ok "sink boon {<dat>}" "crashed on sink boon {<dat>}")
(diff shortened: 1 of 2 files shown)


Fang-:
> assuming it to vestigial from a brief glance at the blame/history.

Well, the jury is still out on this one. 7ca3d96 suggests there's a reentrancy case we need to be concerned with, but also does not inject a `%boon` in the crash case. Not sure I understand it fully, or whether it's still relevant to the Ames implementation as it exists today.

yosoyubik:
I don't understand how how multiple %boons could be added in the case of a crash, but I don't think the conditions for re-entrant %dones has been changed, so restoring the behavior added in https://github.com/urbit/urbit/pull/2115—don't give the %boon to the vane if not ok, and then transform all %boon moves to %lost—seems to be the best solution since this was a regression.

As @Fang- has pointed out, a test for the "previous %boons" scenario seems needed here to fully describe how we can get to that state.

PR #6613, opened 2023-05-24
ames: skip corks if pending acks for naxplanations

### LINKED ISSUE #6608: ames: one-sided flow corking
In implementing #6598, I have run into the following:

1) Sender send a plea.
2) Receiver nacks it.
3) Sender receives a `%done` with an error inside. In response to this, it sends a different plea, immediately (within the same event) followed by a `%cork` on that flow.
4) On the receiver, the flow does not show up in the debug dashboard. But on the sender, it still does:

image

Also of note is that in my test case, the receiver nacks the second plea as well, but the sender never receives a `%done` for this.

Pretty sure I should still be receiving a second `%done`, and that flows should always be removed symmetrically.

yosoyubik:
Fixes #6608 by implementing the solution proposed [here](https://github.com/urbit/urbit/issues/6608#issuecomment-1559498863): 

> to check if we, as a server ames (the one not sending the %cork), are waiting for acks on messages we have sent, don't process the %cork, and wait for the client to send another one.

Testing the error found in https://github.com/urbit/urbit/pull/6598, it removes the flow on both sides, but waiting to hear from @Fang- to confirm that there are no other issues.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -3775,19 +3775,50 @@
               %done  (done ok.task)
             ::
                 %hear
-              ?.  ?|  corked
+              |^  ?:  ?|  corked
                       ?&  %*(corked sink bone (mix 0b10 bone))
                           =(%nack (received bone))
                   ==  ==
-               (hear [lane shut-packet ok]:task)
-              ::  if we %hear a task on a corked bone, always ack
+                ack-on-corked-bone
               ::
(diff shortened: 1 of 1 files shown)


Fang-:
Can confirm that when applied to the recipient ship, this fixes the problem I described in #6608.

belisarius222:
It would be good to have a test confirming that two nacked pleas followed by a cork get processed properly.

joemfb (approved):
this will do

PR #6615, opened 2023-05-24
hoon: 138k

belisarius222:
Draft PR to accumulate upcoming Hoon changes that will go into the next Hoon Kelvin, 138.


joemfb:
CI is failing since it's not using a compatible runtime (ie, one built from `next/hoon/138` in `urbit/vere`).

b3389ef7790e59a4b92e8259419537e67b122bf1 backs out a change from #5884 -- I had suggested that a `^~` be added to `+bool` to improve structural sharing in `$type`, but that compile-time evaluation misses all the relevant jets in `+put:in` (hash ordering, math, &c) if they weren't already registered.

PR #6617, opened 2023-05-24
fine: handle misordered responses

joemfb:
Unbelievably, *fine* has been implicitly depending on hearing response fragments in the correct order. This PR fixes that by sorting them before deserialization, with a test case to prove it. It also limits fine response error messages to the appropriate scope, and refactors the existing fine tests.


joemfb:
Fine could use an ordered map instead of a list for response fragments to reduce these list traversals. But that will require a state migration, and is unlikely to be a bottleneck for now.

belisarius222 (approved):

PR #6621, opened 2023-05-26
hoon: jib and jub wip

eamsden:
Adds +jib (like jab, but call the gate with its bunt if the key is missing) and jub (like jab, but call the gate with `~` if the key is missing and `[~ q.n.a]` if present) to the stdlib 

NOT YET TESTED

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1540,6 +1540,34 @@
     ::
     a(r $(a r.a))
   ::
+  ++  jub                                               ::  add or modify
+    ~/  %jub
+    |*  [key=_?>(?=(^ a) p.n.a) fun=$-((unit _?>(?=(^ a) q.n.a)) _?>(?=(^ a) q.n.a))]
+    ^+  a
+    ::
+    ?~  a  [key (fun ~) ~ ~]
+    ::
+    ?:  =(key p.n.a)
+      a(q.n (fun `q.n.a))
+    ::
+    ?:  (gor key p.n.a)
+      a(l $(a l.a))
+    a(r $(a r.a))
+  ::                                                    :: bunt or modify
+  ++  jib  
+    ~/  %jib
+    |*  [key=_?>(?=(^ a) p.n.a) fun=$-(_?>(?=(^ a) q.n.a) _?>(?=(^ a) q.n.a))]
+    ^+  a
+    ::
+    ?~  a  [key (fun) ~ ~]
+    ::
+    ?:  =(key p.n.a)
+      a(q.n (fun q.n.a))
+    ::
+    ?:  (gor key p.n.a)
+      a(l $(a l.a))
+    a(r $(a r.a))
+  ::
   ++  mar                                               ::  add with validation
     |*  [b=* c=(unit *)]
     ?~  c


pkg/arvo/sys/hoon.hoon
```diff
@@ -1540,6 +1540,34 @@
     ::
     a(r $(a r.a))
   ::
+  ++  jub                                               ::  add or modify
+    ~/  %jub
+    |*  [key=_?>(?=(^ a) p.n.a) fun=$-((unit _?>(?=(^ a) q.n.a)) _?>(?=(^ a) q.n.a))]
+    ^+  a
+    ::
+    ?~  a  [key (fun ~) ~ ~]
```
dr-frmr, line 1548: Both +jub and +jib worked for me if I changed `[key (fun ~) ~ ~]` to `[key^(fun ~) ~ ~]`

dr-frmr (commented):
I find +jub in particular to be much more useful than +jab has been in practice -- this is a very frequent operation, where you want to add an item to a map, but if it already exists, modify it. Usually involves a cumbersome +get then +put

pkg/arvo/sys/hoon.hoon
ashelkovnykov: Yes, the `node` type in `tree` for `map` must be a `pair`.

PS - please make it `(pair key (fun ~))` or `[key (fun ~)]`; I hate the `+` `/` and `^` sugars so much...

ashelkovnykov (approved):
LGTM

xiphiness:
https://github.com/xiphiness/urbit/tree/x/jub
https://github.com/xiphiness/vere/tree/x/jub

In December I worked on jub (I had it named the same), with the modification that also, it returns a unit, deleting if the kv pair is present and the gate returns a ~. I had thought there was an issue with wet gates and the function, but turns out it was with the jet I wrote. I lost a lot of the code, but for core school homework I rewrote both the jet and the hoon code.

eamsden:
if @xiphiness has jets for this I'd like to propose that it be merged once those are ready.

xiphiness:
My jub is a superset of what your jub provides.

I'm not sure about my reference hygiene for one, secondly, for those who know treaps better, in the case of the recursive calls, are my use of puts needed in the case that the recursive call returns an empty map? or is there something cleverer I can do?

I can accommodate your jib too. I just wrote the jet for it, but thinking about it, I don't think it's going to work with the hoon you wrote. Jab is designed to crash if the value doesn't exist, and moreover, it will I believe crash at the building-of-the-gate step if the map is empty. The value of the sample of the gate passed being a `_?>(?=(^ a) q.n.a)` rather than `_?>(?=(^ a) q.n.a)` means I think that it implicitly becomes a wet version of a _?>(?=(^ a) |:(q.n.a q.n.a)), i.e. the sample is constructed by asserting a non-empty map and grabbing the root node. Turning the function into a `$-(* *)` should probably work, as run:by does, but the question then becomes whether or not we add a deletion semantics as with my jub by returning a unit.

But in theory my jub handles all use cases, and the pattern you'd use with your jib can be slightly more verbosely expressed by wrapping the gate `fun` you would put into your jib into a `|=((unit val) ?~(+< `$:fun `(fun u)))`.

xiphiness:
https://github.com/urbit/urbit/pull/6782

jackfoxy:
It would be much cleaner from a user perspective to just deal with primitives `add` (or `ins` for "insert") and `upd` (for update) in `++  by` to act on a key and noun.

Maybe there's a good reason to have `jib` and `jab` requiring gates. These are not the same as `add` and `upd` primitives.

xiphiness:
using the other hypothetical primitives i'd say is a matter of preference, but what's nice about jub is that you can handle the case of presence, absence, and whether to mutate, keep the same, or delete with only a single tree traversal, so all of the other operations can be built in terms of it.

jackfoxy:
You can keep `jib`, `jab`, `jub` and any other function you want, but they are not programming primitives. They are higher order functions.

Not having programming primitives at the standard library level is second guessing everything all application developers could ever do.

I wouldn't be complaining about this if it didn't effect my programming directly, as I foresaw two years ago.

1) Workarounds make the code more complex than necessary.
2) Requiring a gate when all you are doing is moving a noun around is always going to execute slower. Validating hoon's reputation for slowness.
3) Crash is part of the hoon programming paradigm. Not having a proper `ins` and `upd` means you can't program against map in this fashion. 

I have never seen another standard library implementation of map without the usual primitives.

PR #6622, opened 2023-05-29
%hugo

tadad:


jalehman:
@tadad FYI, this PR is going to need an accompanying [UIP](https://github.com/urbit/UIPs) before review and consideration.

PR #6626, opened 2023-05-30
shoe: renders blank tape rows

### LINKED ISSUE #6625: shoe: intentional empty tapes do not render
Empty tapes pushed to shoe don't render because `+row`'s trap stops at the first instance of one. This happens because an empty tape is auto-generated _to indicate_ that all rows had been processed. A different indicator should be used so _intended_ empty tapes render. 

This issue was discovered through work on %talk-cli (https://github.com/tloncorp/landscape-apps/pull/2212), where codeblocks with consecutive linebreaks get cut off.

polrel-witter:
Resolves #6625

To fix, I changed the instance of `+roll` to a `+spin` which increments a counter each time a column contains empty tapes. Also, `+row`'s trap now stops at the point when the null count equals the number of columns, indicating there's nothing left to render.

### DIFF
--- a/pkg/base-dev/lib/shoe.hoon
+++ b/pkg/base-dev/lib/shoe.hoon
@@ -413,23 +413,25 @@
     =/  cobs=(list [wid=@ud (list tape)])
       (turn cows col-as-lines)
     =+  [lin=0 any=|]
+    =+  len=(lent cobs)
     =|  fez=(list sole-effect)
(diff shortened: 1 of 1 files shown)


Fang- (approved):
Took me a little while to grok, but mainly just because the original logic is slightly wack, and so of course that carries over. This looks like the same thing implemented in a less dumb/error-prone way. Approved!

PR #6629, opened 2023-05-31
base: add %runtime-version

pkova:
Adds a new agent to %base called %runtime-version. This agent has a single subscription endpoint `/version` that returns a `%vere-update` `[cur=vere next=(unit vere)]`, where next is non-null if the runtime should be upgraded. You get a a single `%vere-update` as soon as you subscribe and every time a new runtime version is available thereafter. 

This does not yet properly support `soon` or `edge` paces, something like `2.7-de2d39b` will be truncated to `2.7` by `%runtime-version`.

The method of discovering the latest version is essentially doing `curl https://bootstrap.urbit.org/vere/live/last` every thirty minutes. I considered implementing this via remote scry or having everybody subscribing to a pre-determined ship, but I think this solution is the most robust for now. The subscription API is nice and the backend can be changed to be more urbit-native later.


joemfb (dismissed):
I think you should restore the vere mark(s), that type hasn't changed, and it's possible they're already being used somewhere.

Other than that, this looks nice and clean. It'll be great to add pace configuration (maybe this should supersede `$pier/.bin/pace`) and per-pace subscriptions.

pkova:
Decided to make this a thread instead. A subscription API would be nice but propagating the information of when the runtime had been upgraded to userspace proved too much trouble for it's worth. Callers can instead run this thread whenever they need to know the runtime version and/or if the runtime is up to date. I updated my PR https://github.com/tloncorp/landscape/pull/189 to reflect this change.

belisarius222 (approved):

PR #6636, opened 2023-06-02
spider: provide a tang http response when a thread called via http crashes

niblyx-malnus:
(1) In +thread-fail, +thread-clean is called before +thread-http-fail and +cancel-scry meaning neither of the latter two actually run in any meaningful way. `serving` has been replaced with `(~(del by serving.state) tid)` and `scrying` has been replaced with `(~(del by scrying.state) tid)` in the state and therefore we cannot `(~(get by serving.state) tid)` in +thread-http-fail or `(~(get by scrying.state) tid)` in +cancel-scry.

(2) In +thread-http-fail return an informative tang as json instead of an empty 500 http response when the failure results from an internal crash.

### DIFF
--- a/pkg/arvo/app/spider.hoon
+++ b/pkg/arvo/app/spider.hoon
@@ -557,7 +557,8 @@
   ?.  ?=(http-error:spider term)
     %-  (slog tang)
     =/  tube  (convert-tube %tang %json desk bowl)
-    %-  json-response:gen:server
+    :-  [500 [['content-type' 'application/json'] ~]]
+    =-  `(as-octt:mimes:html (en-json:html -))
     o/(malt `(list [key=@t json])`[term+s/term tang+!<(json (tube !>(tang))) ~])
   :_  ~  :_  ~
   ?-  term
--- a/pkg/arvo/app/spider.hoon
+++ b/pkg/arvo/app/spider.hoon
@@ -558,7 +558,7 @@
     %-  (slog tang)
     =/  tube  (convert-tube %tang %json desk bowl)
     :-  [500 [['content-type' 'application/json'] ~]]
-    =-  `(as-octt:mimes:html (en-json:html -))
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/spider.hoon
```diff
@@ -554,9 +554,12 @@
     ~
   %+  give-simple-payload:app:server  u.eyre-id
   ^-  simple-payload:http
-  :_  ~  :_  ~
   ?.  ?=(http-error:spider term)
-    ((slog tang) 500)
+    %-  (slog tang)
+    =/  tube  (convert-tube %tang %json desk bowl)
+    %-  json-response:gen:server
```
belisarius222, line 560: Shouldn't the HTTP status code here still be 500?  We did error out, after all.

niblyx-malnus: Yeah this is a good point. Is [this](https://github.com/urbit/urbit/pull/6636/commits/1818790e357c6c92f6d162b649f571ec073c62ff) an appropriate fix?

pkg/arvo/app/spider.hoon
```diff
@@ -554,9 +554,13 @@
     ~
   %+  give-simple-payload:app:server  u.eyre-id
   ^-  simple-payload:http
-  :_  ~  :_  ~
   ?.  ?=(http-error:spider term)
-    ((slog tang) 500)
+    %-  (slog tang)
+    =/  tube  (convert-tube %tang %json desk bowl)
+    :-  [500 [['content-type' 'application/json'] ~]]
+    =-  `(as-octt:mimes:html (en-json:html -))
```
pkova, line 561: Lets do `as-octs:mimes:html` and `en:json:html` here.

pkova (changes requested):

pkova:
Looks like the old behavior of a thread crashing was the http request hanging forever. This is good.

PR #6640, opened 2023-06-04
hoon: remove rut:by, duplicate of urn:by

ashelkovnykov:


joemfb (commented):
It'll be good to remove this duplication. This pr should target `next/hoon/138`. And I'm inclined to think that `+rut:by` is a better name than `+urn:by`, but I'm open to arguments to the contrary.

ashelkovnykov:
@joemfb Rebased on `next/hoon/138`

One argument is an "argument from convenience": there were only 3 non-test usages of `rut`, but there were 14 non-test usages of `urn`.

Another argument is that there is already another `run`/`rut` pair of gates in the bit arithmetic section which isn't operationally related to this pair of arms.

belisarius222 (approved):

PR #6641, opened 2023-06-05
Unify container generation gates

ashelkovnykov:
Proposal to unify and reduce the different `list`, `map`, and `set` generation gates into just three gates respectively: `limo`, `malt`, and `silt`.


ohAitch:
Any reason not to stick to the two-letter names in common use, and get rid of the four-letter ones? I'm seeing 20 calls to `silt` before this PR, and hundreds to `sy`.

Also, I'm surprised `malt` vs `molt` doesn't break anything - presumably this change will cause a bunch of `[p=p=key q=q=val]` maps - but if it compiles it compiles.

ashelkovnykov:
@ohAitch Confirmed, you were correct that it was adding extra `p=`, `q=` faces. I've added a commit to strip the extra faces.

joemfb:
These container constructors exist for two reasons: ergonomics in "casual" programming, and various wet-gate shenanigans. In normal code, it's best to just cast lists and use `+gas:in`/`+gas:by` for sets and maps. For whatever it's worth, I have never had to use them, and choose not to (outside of messing around in the dojo).

Hoon's wet core types and implementations have gone through many revisions. The last added a namespace-unification step (see `+redo:ut`) which significantly improves basic usability: you used to have to write `(flop (limo [1 2 3 ~]))` or `(flop `(list @)`[1 2 3 ~])`, but now `(flop [1 2 3 ~])` works as expected. For wet gates with more complicated sample types (i believe anytime holds are implicated), `+redo:ut` is not up to the task and the gate author must take extra steps to support "casual" call sites. Compare:

```
> (turn [1 2 3 ~] |=(a=@ +(a)))
~[2 3 4]
> > (spin [1 2 3 ~] 0 |=([a=@ b=@] [+(a) b]))
mull-grow
> (spin (limo [1 2 3 ~]) 0 |=([a=@ b=@] [+(a) b]))
[p=~[2 3 4] q=0]
```

If `+spin` is modified to include `=>  .(a (homo a))`:

```
> (sspin [1 2 3 ~] 0 |=([a=@ b=@] [+(a) b]))
[p=~[2 3 4] q=0]
```

This kind of thing is usually only an issue for lists -- maps and sets are too awkward to manually construct. The ergonomics of the stdlib would be significantly improved by making sure that list-processing wet-gates have appropriate casts and sample-manipulation of this sort. (And their efficiency would be somewhat improved by a rune that allowed casting the subject without simultaneously "editing" it.)

I was going to say that the only time something like `+limo` is required is in some parser combinators, which can't be cast due to `;~` semantics ... But then I tried modifying `+stew` (adding `=>  .(leh (homo leh))`) and my examples (try `(stew ['a' prn] ~)`) started working without it. So let's just say that it's an open question whether `+limo` and friends are *ever* needed given the current type system.

That leaves us with ergonomics. `(sy 1 2 3 ~)` is a lot nicer than `(~(gas in *(set @ud)) 1 2 3 ~)` when experimenting or writing example code. (Again, for product code, I come down on the side of being explicit.)

Switching gears, we have the following constructors:

- `+limo`
- `+homo`
- `+malt`
- `+molt`
- `+silt
- `+ly` (ie, `+le:nl`)
- `+my` (ie, `+my:nl`)
- `+sy` (ie, `+si:nl`)

The last three were the latest added, at the time as `+redo:ut`. `+ly` is equivalent to `+homo`, while `+my` and +sy` looking something like `(molt (homo ...))` and `(silt (homo ...))`.

My inclination would be to improve the stdlib, remove as many of these call sites as possible, and then keep the short names. (I'm not sure why the indirection and duplication of the `+nl` core exists -- I'd be inclined to try to remove it).

That would mean:

- adding `=>  .(foo ^.(homo foo)` to all wet gates accepting lists in their sample (`^.` is better than a call, easier to recognize in nock and compile out)
- removing all now-unnecessary calls to `+limo` (presumably all of them)
- deleting `+limo`, `+malt`, `+molt`, and `+silt`
- redefining `+ly` as an alias to `+homo`, inling `+my:nl` into `+my` and `+si:nl` into `+sy`
- deleting `+nl`.

Alternatly, `+limo`, `+malt`, `+molt`, and `+silt` could stay around as (deprecated?) aliases to their more-modern counterparts (while still removing all unnecessary call sites).

The only downside I can see from this approach is that including casts to `+homo` in stdlib wet gates will increase compile time, as it'll be inferred through during the wet-gate monomorphization check performed at each call site. But I can't think of a good way to quantify that impact without trying it.

Alternately, we could punt this little messy corner of the stdlib into the future. But this kind of approach is what I think is needed to make changes in this area.

ashelkovnykov:
@joemfb I'm okay to:
1. take ownership over this
2. punt into the future (`409`? After the latest batch of changes go live, and the next anticipated batch, too?)

Also gives time for @belisarius222 and @jalehman to comment, since I know that there is a growing movement to change the STL as infrequently as possible

PR #6642, opened 2023-06-05
ames: remove space leak for outbound naxplanations

yosoyubik:
Companion to https://github.com/urbit/urbit/pull/6607, where the space leak was fixed. Built on top of https://github.com/urbit/urbit/pull/6609, to use the state migration to clean up the stale messages in the naxplanation queue.

~~(Note: draft until I test this with a real reproduction)~~

Until https://github.com/urbit/urbit/pull/6607, gets merged, this is the relevant change for this PR:

https://github.com/urbit/urbit/pull/6642/commits/2c854d1285e9baddb660c037f5dcdd11d3d595af


joemfb (approved):
this will do

yosoyubik:
"Always test your code on a live ship". 

The version before https://github.com/urbit/urbit/pull/6642/commits/5fd372dfcb52c1e6a8b71689b20b081b4252847e considered that all target bones of flows we have received (in rcv.peer-state) will be naxplanation we have sent `~(got by snd.peer-state)`. This is wrong, so instead of crashing, we just check if it exists and if not continue processing the next bone.

belisarius222 (approved):

PR #6647, opened 2023-06-07
hoon: fixes a serious bug in +fitz

mikolajpp:
`+fitz` aura compatibility check arm is badly broken due to a typo. With the `+fitz` jet disabled, the aura system is in complete breakdown, 
```
> (fitz 'tas' 'p')
%.y
> (fitz 'ud' 'ui')
%.y
> (fitz 'da' 'p')
%.y
```
.
This PR fixes it. 

Examination of the `+fitz` jet also revealed a separate jet mismatch. 
```
> (fitz 'AB' 'CB')
%.y
> +fitz ['AB' 'CB']
%.n
```
Here, the `+fitz` generator is the Hoon version, while the ship runs on a runtime with the `+fitz` jet enabled.
I am going to address this jet mismatch in a PR in the Vere repository.


mikolajpp:
For the tests to pass, the updated jet from https://github.com/urbit/vere/pull/444 in necessary.

tests/sys/hoon/nest.hoon
```diff
@@ -0,0 +1,41 @@
...
+++  test-fitz
+  ;:  weld
+    %+  expect-eq
+      !>  %.y
+      !>  (fitz ~. ~.tas)
+    ::
+    %+  expect-eq
+      !>  %.y
+      !>  (fitz ~.ud ~.)
+    ::
```
ashelkovnykov, line 12: I'm not familiar with `+fitz`, but based on the logic for bit-length, it seems like it's testing composability: "does `wix` nest in `yaz`?" Is that correct?

If so, it seems like this test should fail, and that line `7258` in `+fitz` should be removed.

mikolajpp: `+fitz` does two things. (1) It checks whether, as specified by the last uppercase bit-width indicator, `wix` _nests_ under `yaz`, size-wise. Auras without bit-width annotation nests under those that have it. (2) It checks whether the two auras nest in the sense of _compatibility_. 

As stated in the [Aura Reference](https://developers.urbit.org/reference/hoon/auras),  _A given aura nests under any aura whose name is a substring or extension of the given aura_,
```
> :-  (~(nest ut [%atom %ud ~]) | [%atom %u ~])
      (~(nest ut [%atom %ud ~]) | [%atom %udx ~])
[%.y %.y]
```
.
 `@tas` and `@ta` are compatible, but `@ud` and `@ux` are not. The aura compatibility is decided upon by string matching, including all letters (lower-case, uppercase) in the test but the last uppercase (and only uppercase) letter of each aura. 

An `@tas` atom can be cast to `@ta` and vice versa. The following all works currently (here it is the `+fitz` jet working)
```
> ^-(@ta `@tas`'ram-pam')
~.ram-pam
> ^-(@tas `@ta`'ram pam')
%ram pam
```
Such reverse nesting sometimes is an issue, sometimes not. However, removing line `7258` would constitute a dramatic change in the aura system. It would break the ability to cast an empty aura `@` to any other aura.
Furthermore, it would make valid castings such as this
```
> ^-(@tas `@ta`'ram-pam')
%ram-pam
```
result in a nest fail error.

This PR is only intended to fix a bug in `+fitz` as it was written. Its current behaviour matches the jet at this point.

Otherwise, the Aura system is known to suffer from inconsistencies and lack of sanity enforcement at present. Fixing it is a possible project we are considering with @joemfb.

ashelkovnykov: Understood, thanks for the very clear elaboration. I wasn't consciously aware that an aura is considered to nest in an aura that is an *extension* of itself - but of course, it does, since as you say that's how we cast down from `@`.

joemfb (approved):
this will do

PR #6652, opened 2023-06-10
dojo: complete-naked-poke for all agents

midden-fabler:
`+complete-naked-poke` now works for all agents.

`:do<tab>`
```
:docket
:dojo
```


joemfb (changes requested):
App source files in %clay aren't necessarily running, so I this completion could give false positives (in a certain sense). You can get the set of running/suspended apps with `.^((set [dude:gall ?]) ge+%/$)`.

joemfb (approved):

PR #6656, opened 2023-06-12
hoon: add rune ?_, equality test w type refinement

xiphiness:
This PR introduces a new rune ?_(a=hoon b=wing), subject to renaming, whose compiled nock is equivalent to .=, but refines the type of b to that of a on equality. Most notably, this introduces new types of polymorphism with wet gates, particularly, I have managed to make work a dependently typed map, which could *perhaps* even be used for dependently typed axals later on.

I haven't included any changes to e.g. +by to fully exploit this yet, though I have tested this out on my own by replacing put and get by:
```hoon
++  put                                               ::  adds key-value pair
  ~/  %put
  |*  [b=* c=*]
  |-  ^+  a
  ?~  a
    [[b c] ~ ~]
  ?:  ?_(b p.n.a)
    ?:  =(c q.n.a)
      a
    a(n [b c])
  ?:  (gor b p.n.a)
    =+  d=$(a l.a)
    ?>  ?=(^ d)
    ?:  (mor p.n.a p.n.d)
      a(l d)
    d(r a(l r.d))
  =+  d=$(a r.a)
  ?>  ?=(^ d)
  ?:  (mor p.n.a p.n.d)
    a(r d)
  d(l a(r l.d))
::
++  get                                               ::  grab value by key
  ~/  %get
  |*  b=*
  |-  ^-  (unit _?>(&(?=(^ a) ?_(b p.n.a)) q.n.a))
  ?~  a
    ~
  ?:  ?_(b p.n.a)
    (some q.n.a)
  ?:  (gor b p.n.a)
    $(a l.a)
  $(a r.a)
```

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -6499,6 +6499,7 @@
     [%wtpt p=wing q=hoon r=hoon]                        ::  ?@  if p is atom
     [%wtsg p=wing q=hoon r=hoon]                        ::  ?~  if p is null
     [%wthx p=skin q=wing]                               ::  ?#  if q matches p
+    [%wtcb p=hoon q=wing]                               ::  ?_  if q equals p
     [%wtts p=spec q=wing]                               ::  ?=  if q matches p
     [%wtzp p=hoon]                                      ::  ?!  loobean not
   ::                                            ::::::  special
@@ -7497,6 +7498,7 @@
   ++  wtsg  |=([sic=hoon non=hoon] (gray [%wtsg puce (blue sic) (blue non)]))
   ++  wthx  |=(syn=skin (gray [%wthx (tele syn) puce]))
   ++  wtts  |=(mod=spec (gray [%wtts (teal mod) puce]))
+  ++  wtcb  |=(sic=hoon (gray [%wtcb (blue sic) puce]))
   --
 ::
 ++  ax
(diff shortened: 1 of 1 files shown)


xiphiness:
actually, suddenly occurs to me this can be added to .= by only modifying +chip, i think. just check if its a %dtts, then check if the latter hoon is a %wing, if not, do nothing, or if so, +cool

xiphiness:
oookay we have some mull-bonks in the map case, i guess, because i was testing it with play rather than mint...

ohAitch:
I would expect adding it to `.=` to break a lot that is relying on it being a pure value comparison, and in particular for the codebase to contain some `?: =(%foo x)` that are not `?: ?=(%foo x)` to specifically avoid type side-effects.

xiphiness:
oh, yeah definitely, and note the various times we use =(~ list) instead of ?~ precisely so we don't refine the type

i think what's happening with the mull-bonk-c is that the lose case is losing too much. basically, we should only lose if ?_(a b) a is a constant

ohAitch:
Yep, apologies if I'm stating the obvious!

xiphiness:
oh, okay, i've figured out the issue.

in the %wtcl case of mull, when it loses in the dox case, the type of e.g. b in (~(get by a) b) is * in the dox case, so when we ?_(b p.n.a) the latter case of the ?: is void. hackish, but i think it might be fine to add another flag to cool that says if its wet, in which case, do not apply ?_ restriction, and use that in the dox case

xiphiness:
oops, i did not add the `how` test in `chip` to my new rune but to wthx. this fixed the issue!

line *9698*
```hoon
    ?:  ?=([%wtcb *] gen)
```
to
```hoon
    ?:  &(how ?=([%wtcb *] gen))
```
did not fix the problem

jalehman:
@xiphiness Not sure the status of this, but if you want to keep moving it forward, this is going to need a [UIP](https://github.com/urbit/UIPs)

PR #6658, opened 2023-06-12
hoon: fix gate currying

### LINKED ISSUE #6655: incorrect type of the sample of a gate produced by ++curr
**Describe the bug**
Sample of a wet gate produced by `++curr` has incorrect type

**To Reproduce**
Type in dojo:
```
=test |=  [d=@ e=^]
[d e]
+<:(curr test [0 0])
```

Product:
`b=e=[0 0]`

**Expected behaviour**
Expected product: `b=d=0`

**Supposed fix**
Change [this line](https://github.com/urbit/urbit/blob/82aae2f9e38ed416057b4e16c8b2d9e466092df4/pkg/arvo/sys/hoon.hoon#LL1890C1-L1890C1) to:
`|*  b=_+<-.a`

Quodss:
### Description

Resolves #6655.

Fixes the sample of the wet gate produced by `++curr`.
Prior to the fix, `++curr` used the same expression for the sample of the returned wet gate as in `++cury`, minus the different face. This is incorrect: `++curr` binds the tail of `a`'s sample, so `b` should use the head of `a`'s sample instead of its tail.

The fix was successfully tested on a fakezod, although I only tested it on the version %139.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1905,17 +1905,17 @@
 ++  corl                                                ::  compose backwards
   |*  [a=$-(* *) b=$-(* *)]
   =<  +:|.((a (b)))      ::  type check
-  |*  c=_+<.b
+  |*  c=_,.+<.b
   (a (b c))
 ::
 ++  cury                                                ::  curry left
   |*  [a=$-(^ *) b=*]
-  |*  c=_+<+.a
+  |*  c=_,.+<+.a
   (a b c)
 ::
 ++  curr                                                ::  curry right
   |*  [a=$-(^ *) c=*]
-  |*  b=_+<-.a
+  |*  b=_,.+<-.a
   (a b c)
 ::
 ++  fore  |*(a=$-(* *) |*(b=$-(* *) (pair a b)))        ::  pair before
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1915,7 +1915,7 @@
 ::
 ++  curr                                                ::  curry right
   |*  [a=$-(^ *) c=*]
-  |*  b=_+<+.a
+  |*  b=_+<-.a
   (a b c)
 ::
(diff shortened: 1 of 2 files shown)


Quodss:
IDK if this belongs to the next/hoon/138 branch, I don't think this a breaking change.

pkg/arvo/sys/hoon.hoon
```diff
@@ -1915,7 +1915,7 @@
 ::
 ++  curr                                                ::  curry right
   |*  [a=$-(^ *) c=*]
-  |*  b=_+<+.a
+  |*  b=_+<-.a
```
joemfb, line 1918: This should actually be `b=,.+<-.a` -- leaving the existing face on prevents us from finding the `$` when handling higher-ordered gates.

The equivalent change is needed for `+cury` (`c=_,.+<+.a`). Can you fix them both here?

joemfb (changes requested):

pkg/arvo/sys/hoon.hoon
Quodss: ok, stripped faces in curry functions and also in ++corl

Quodss:
Regression tests added. I used your examples and added a check for ++corl to see that it also strips the face. I recompiled a fakezod with the proposed changes and ran the tests, no errors raised.

joemfb (approved):
CI was already broken on this branch.

These changes look good, thanks for adding the extra fixes and tests (per https://github.com/urbit/urbit/pull/6658#discussion_r1226980308).

joemfb:
For posterity, the presence of extra faces (ie, the absence of `,` here) was preventing currying of higher-ordered gates. 

```
> ((cury roll (gulf 1 3)) add)
mull-grow
-find.$.+2
> (roll (gulf 1 3) add)
6
> (roll (gulf 1 3) b=add)
mull-grow
-find.$.+2
```

joemfb:
```
> +<+:roll
b=<1.isq [[* *] %~]>
> ,.+<+:roll
<1.isq [[* *] %~]>
```

PR #6665, opened 2023-06-13
arvo: print vane hashes as hashes, not `@p`s

dr-frmr:
This has bothered me for a while now: when compiling vanes, arvo prints the mug hashes of the vane source as a ship name.
Screenshot 2023-06-12 at 3 26 17 PM

This is confusing; it is not immediately clear what these `@p`s mean in this context. The first several times I saw these prints, first getting into Urbit, I assumed these were the source ship for code I was downloading, or something like that. Printing the hashes as `@uv`, in line with how desk source hashes are printed, will make their true nature clearer.

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -1052,7 +1052,7 @@
         ++  smit
           |=  [cap=tape sub=vase pax=path txt=@t]
           ^-  vase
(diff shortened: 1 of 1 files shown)


joemfb (commented):
this change looks fine to me.

it should target `next/kelvin/412`, to avoid unnecessarily rebuilding ~everything. There's currently a conflict between that and `develop`, so you can rebase over or just wait for that to be resolved.

dr-frmr:
Ok I'll rebase

Issue #6668, opened 2023-06-14
Feature request: add by, upd by


jackfoxy:
Create a jetted arm `++  add` in the `++  by` arm that inserts a key/value and crashes if the key already exists.

Create a jetted arm `++  upd` in the `++  by` arm that updates an existing key/value and crashes if the key does not exists.

Update documentation at [put by](https://developers.urbit.org/reference/hoon/stdlib/2i#putby) to indicate that it adds a key/value if the key does not exist and updates the value if it does. The documentation currently reads as if `put` inserts key/values only.

PR #6672, opened 2023-06-15
add:by, upd:by

Quodss:
Addes arms to `by` core requested in #6668 

I wonder if jets are even needed since `has:by` and `put:by` are already jetted and the added arms just call those. But that would be a nice exercise for me to try anyway if no one else will do it.

_(Also, `has:by` calls jetted `get:by` and checks if the result is not null, yet `has` is still jetted. So I might be missing something.)_

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1376,6 +1376,12 @@
   ~/  %by
   =|  a=(tree (pair))  ::  (map)
   |@
+  ++  add                                               ::  crash if key exists or add key-value pair
+    |*  [b=* c=*]
+    ^+  a
+    ?:  (has b)
+      !!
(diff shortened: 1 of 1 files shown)


belisarius222:
@joemfb proposed a new gate to unify these ideas and some of the +jab/+jub/+jib gates: a gate that takes in a map, a key, and a gate from `(unit value)` to `value`.  The gate would return a pair of a flag that will be yes if and only if the value was updated, and the updated map -- or maybe a pair of `(unit value)` and the new map, so that callers can see the newly added value.

This could then be used to implement +add:by and +upd:by efficiently, by asserting in the callback that the original value was null or non-null, respectively.

jalehman:
Closing in light of @joemfb 's forthcoming proposal on the #6621 PR

PR #6673, opened 2023-06-15
eyre: keep json to mime tube warm in handle-scry

pkova:
On most ships the mark conversion tube from json to mime is cold. To see if you're affected, open a talk DM with the dev tools network tab open and refresh the page. On most ships this will take 10-30 seconds to fully load.

This happens because we only scry for the tube, but never tell Clay to cache the tube like we do in `+channel-event-to-json`. This PR makes Eyre give Clay a similar `%warp` move after scrying to keep the tube warm.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -868,25 +868,30 @@
     ?~  u.res  (error-response 404 "no scry result")
     =*  mark   p.u.u.res
     =*  vase   q.u.u.res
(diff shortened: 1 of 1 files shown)


jalehman:
@joemfb will test on his ship, review, and merge.

jalehman:
@arthyn Our plan is to include this in our upcoming next release candidate slated for release either next week or early the following one. If this issue is widespread enough within Talk, would recommend hotfixing there instead.

joemfb (approved):
Tested on my planet, this will do.

PR #6676, opened 2023-06-17
gall: fix typo in +ap-peek causing redundant mark conversion

tinnus-napbus:
this fixes a typo in +ap-peek causing %x scries to always run the mark conversion logic even if the mark you want is the same as what the endpoint produced


joemfb (approved):
Good catch!

PR #6678, opened 2023-06-18
gall, eyre: more tube caching for eyre scries

tinnus-napbus:
Eyre currently caches tubes for scries through its external scry interface. The problem is that two mark conversions are actually done:

(mark produced by agent) -> (mark specified by extension in URL path) -> %mime

Eyre can't cache the tube for the first conversion because it's done internally by gall. Gall can't cache it either because the request is coming through its +scry arm.

This PR does two things:

- adds %r care to gall that presents to the agent as %x but doesn't require the trailing mark and doesn't perform mark conversion

- changes eyre to use %r care for gall agent scries and perform the triple conversion of (endpoint mark -> requested mark -> mime) itself, so it can cache both tubes.

Open questions as to whether:

- it's appropriate to add an obscure care to Gall for this particular purpose. It might not have much utility outside of this
- it's ok to lie to the agent about an %r care being an %x care

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -862,36 +862,46 @@
       (error-response 400 "scry path must start with app name")
     ::  attempt the scry that was asked for
     ::
-    =/  res=(unit (unit cage))
-      (do-scry %gx i.site.req (snoc t.site.req (fall ext.req %mime)))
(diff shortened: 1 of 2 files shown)


belisarius222:
Since we're hoping to build persistent runtime Nock memoization caches soon, we shouldn't need to add new interfaces here.  Closing for now.

PR #6683, opened 2023-06-22
sss: rely on ames acks instead of timers

wicrum-wicrun:
SSS had a problem where, because i had naively implemented the nigh/yore protocol using ames pokes, publishers experienced a very high load, with many pokes coming in per minute.

This instead relies on ames acks from the subscriber, in the following way:

1. Subscriber asks to subscribe.
2. Publisher sends its latest rock `n`.
3. *Only when* the subscriber acks receipt of a rock or a wave `n`, the publisher performs one of the following:
  a. If the ack was a nack, no-op.
  b. If the publisher has wave `n+1`, send it.
  c. If publisher doesn't have wave `n+1`, but it does have a higher rock, send that.
  d. If publisher doesn't have wave `n+1` and it doesn't have a higher rock, store subscriber `~sampel` in a map. Once the next wave is published, send that to all subscribers in the map, and empty it.
5. Repeat step 3.

- [x] Test unsubscribing. The publisher shouldn't continue to send waves after one has been nacked.
- [x] Test how permissions interact with sending out the next wave. Blocked subs shouldn't get waves even if they requested it before being blocked.
- [x] Check tombstoning/closing of paths. It shouldn't put you in an infinite loop.
- [x] Subscribers should get notified if a path is closed or if they get blocked from it.
- [x] Work around that breadth-first move ordering causes on-rock pokes to be processed in the wrong order.

**NOTE 1:** We no longer guarantee that all incoming waves and rocks will generate an on-rock poke. This is because by the time we get some of them, they may be stale and should be ignored. This is caused by ack special-casing in Gall that works around Arvo's depth-first move ordering. This will likely only be an issue when the subscriber and publisher are running on the same ship, and then only while the subscriber is still catching up from a snapshot. It *will* however always receive the final on-rock poke, once it's fully caught up.

**NOTE 2:** While we do perform a state migration from the old version of SSS, this is not backwards compatible with the old SSS network protocol. Old pokes aren't supported, and the new pokes cannot be interpreted using the old version.

### Extra features that should probably be added at the same time
- [x] Add choice to not forget history.
- [x] ~Solid-state versioning?~ **NO**, start a new path instead!

### DIFF
--- a/pkg/base-dev/lib/sss.hoon
+++ b/pkg/base-dev/lib/sss.hoon
@@ -115,15 +115,17 @@
       :_  0/(~(put by sub) current `flow)  :_  ~
       (on-rock-poke current flow ~)
     ::
-    =/  [wave=(unit wave:lake) =flow]
+    =/  [wave=(unit wave:lake) new=(unit flow)]
       ?-  what.res
-        %rock  ?>  |(stale.old (gte aeon.res aeon.old))
+        %rock  ?:  (lte aeon.res aeon.old)  [~ ~]
                [~ [aeon.res | | rock.res]]
-        %wave  ?>  =(aeon.res +(aeon.old))
+        %wave  ?:  (lte aeon.res aeon.old)  [~ ~]
+               ?>  =(aeon.res +(aeon.old))
                [`wave.res [aeon.res | | (wash:lake rock.old wave.res)]]
       ==
-    :_  0/(~(put by sub) current `flow)  :_  ~
-    (on-rock-poke current flow wave)
+    ?~  new  `0/sub
+    :_  0/(~(put by sub) current new)  :_  ~
+    (on-rock-poke current u.new wave)
   ::
   ::  Non-public facing arms below
   ::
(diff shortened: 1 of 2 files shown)


wicrum-wicrun:
I implemented the new retention policy as we discussed. Instead of `[rocks=@ud waves=@ud]` we now have `[horizon=(unit @ud) frequency=@ud]`.

- `frequency` is how often we should make new waves.
- A `horizon` of `[~ a]` means that all rocks older than `a` generations will be deleted. The exception is if this is smaller than `frequency`, since we do always need to keep one rock around.
- A `horizon` of `~` means no cutoff: we never delete the oldest rock, we only delete everything between the oldest and the newest.

pkg/base-dev/lib/sss.hoon
```diff
@@ -313,37 +341,51 @@
...
-    :-  ~[(send nigh/~ src.bowl [dude u.when path]:req)]
-    :-  %0
-    %+  ~(put by pub)  path.req
+      :_  1/pub  :_  ~
+      (send tomb/~ src.bowl dude.req path.req)
+    :_  1/pub  :_  ~
+    =/  last  (fall (pry:rok rok.tid.buoy) *[=key =val]:rok)
+    (send rock/last src.bowl dude.req path.req)
+  ::
+  ++  tell
```
belisarius222, line 350: What's the difference between this and the other +tell arm?

belisarius222: nvm, it's in +du vs in +da

belisarius222 (approved):

PR #6685, opened 2023-06-22
arvo: fix scry permission check

### LINKED ISSUE #6600: Arvo's Vere version scry is too permissive
You can ask at any desk, but it should filter for only the `%$` desk.  This should be fixed in the next Arvo Kelvin.

MarcusMiguel:
Resolves #6600

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -1555,9 +1555,9 @@
         ^-  roon
         |=  [lyc=gang pov=path car=term bem=beam]
         ^-  (unit (unit cage))
-        ?.  ?|  =(our p.bem)
+        ?.  ?&  =(our p.bem)
                 ?=(%$ q.bem)
-                =([%da now] p.r.bem)
+                =([%da now] r.bem)
             ==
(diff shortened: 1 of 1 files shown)


joemfb:
Can you target `next/kelvin/412` (and rebase onto it, there are currently conflicts)? Otherwise, looks good.

pkg/arvo/sys/arvo.hoon
```diff
@@ -1555,9 +1555,9 @@
         ^-  roon
         |=  [lyc=gang pov=path car=term bem=beam]
         ^-  (unit (unit cage))
-        ?.  ?|  =(our p.bem)
+        ?.  ?&  =(our p.bem)
                 ?=(%$ q.bem)
-                =([%da now] p.r.bem)
+                =(now p.r.bem)
```
joemfb, line 1560: This needs to check tag too, otherwise we'd be incorrectly responding to some possible `%ud` and `%uv` cares.

joemfb (changes requested):

joemfb (approved):
thanks!

Issue #6459, opened 2023-04-08
Generator build failures


MnkyPwz:
I've tried to execute generators in dojo but they all seem to fail and I'm uncertain as to the cause. I've even breached, re-installed urbit onto the server but still am encountering the same issue. Curious if anyone might know why this could be happening.

This is an example of me just attempting to do a simple |ping 

```
>   dojo: nest-need
[[now=@da eny=@ bec=[p=@p q=@tas r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud])]] [exp=@ud %~] %~]
>   dojo: nest-have
[[now=@da eny=@uvJ bec=[p=@p q=@tas r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud])]] %~ %~]
> |ping
dojo: generator failure
```

Additionally, this is what I get when I simply attempt to check the version of my urbit instance |version

```
clay: read-at-aeon fail [desk=%base care=%a case=[%da p=~2023.4.8..18.47.21..e69e] path=/gen/hood/version/hoon]
[%file-not-found-mist /gen/hood/version/hoon]
[%error-validating /gen/hood/version/hoon]
[%error-building /gen/hood/version/hoon]
clay: %a build failed [%base 1 /gen/hood/version/hoon]
dojo: %writ fail [i=%~.hand t=/gen/hood/version]
> |version
/gen/hood/version/hoon
%generator-build-fail
```

Any thoughts would be appreciated.

jfranklin9000:
`|ping` takes a `@ud`:
```
[%random-bytes 1 ~fonwed-matnev]
> |ping 1
>=
< ~sampel-palnet: atom: 1 bytes, mug ~fonwed-matnev
```
I'm not sure what it's doing but here's the source:
```
+cat %/gen/hood/ping/hoon
/~sampel-palnet/base/~2023.4.9..00.37.02..003c/gen/hood/ping/hoon
::  ping: send atom
::
::::  /hoon/ping/hood/gen
  ::
/?    310
:-  %say
|=  $:  [now=@da eny=@ bec=beak]
        [[exp=@ud ~] ~]
    ==
=+  rab=(~(raw og eny) (bex exp))
~&  [%random-bytes (met 3 rab) `@p`(mug rab)]
atom+rab
```
I don't see `version.hoon` in `%/gen` (`+version`) or `%/gen/hood` (`|version`).
```
+ls %/gen
< no version/hoon >

+ls %/gen/hood
< no version/hood >
```
[...]

PR #6686, opened 2023-06-23
dojo: better generator errors

midden-fabler:
`%say` and `%ask` generators now skip printing `[now eny beak]` in error messages, making them more legible.

For context see:
[Separate type-checks for head and tail of generator arguments](https://github.com/urbit/arvo/issues/65)
[Generator build failures](https://github.com/urbit/urbit/issues/6459)

`|new-desk` with no arguments is used as an example.

before:
```
>   dojo: nest-need
[ [ now=@da
    eny=@uvJ
      bek
    [ p=@p
      q=@tas
      r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud] [%uv p=@uv])
    ]
  ]
  [desk=@tas %~]
  from=@tas
  hard=?(%.y %.n)
  gall=?(%.y %.n)
]
>   dojo: nest-have
[ [ now=@da
    eny=@uvJ
      bec
    [ p=@p
      q=@tas
      r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud] [%uv p=@uv])
    ]
  ]
  %~
  from=@tas
  hard=?(%.y %.n)
  gall=?(%.y %.n)
]
> |new-desk
dojo: generator failure
```

after:
```
>   dojo: nest-need
[[desk=@tas %~] from=@tas hard=?(%.y %.n) gall=?(%.y %.n)]
>   dojo: nest-have
[%~ from=@tas hard=?(%.y %.n) gall=?(%.y %.n)]
> |new-desk
dojo: generator failure
```

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -846,7 +846,6 @@
       ::  kev: key-value named arguments
       ::  kuv: default keyword arguments
       ::  sam: fully populated sample
-      ::  rog: default gat sample
       ::
       |.  ^-  vase
       =/  gat=vase  (slot 3 q.cay)
@@ -883,10 +882,12 @@
       ::
       =/  sam=vase  :(slop ven poz kev)
       ?.  (~(nest ut p.som) | p.sam)
+        =/  ned=vase  (slot 3 som)
+        =/  hav=vase  (slot 3 sam)
         ~>  %slog.1^leaf+"dojo: nest-need"
-        ~>  %slog.0^(skol p.som)
+        ~>  %slog.0^(skol p.ned)
         ~>  %slog.1^leaf+"dojo: nest-have"
-        ~>  %slog.0^(skol p.sam)
+        ~>  %slog.0^(skol p.hav)
         !!
       (slam gat sam)
     ::


pkova:
Note that we want to have two different nest fails in this situation:

1. Someone has miswritten a generator. Let's say my sample for `/gen/new-desk` is this:

```
|=  $:  [now=@da eny=@uvJ this-is-wrong=@ bek=beak]
        [=desk ~]
        [from=$~(%base desk) hard=_| gall=_|]
    ==
```

In this situation I would like to see an error something like this:
```
>   dojo: invalid generator sample, need
[ now=@da
    eny=@uvJ
      bek
    [ p=@p
      q=@tas
      r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud] [%uv p=@uv])
]
>   dojo: have
[ now=@da
    eny=@uvJ
    this-is-wrong=@
      bec
    [ p=@p
      q=@tas
      r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud] [%uv p=@uv])
]
```

2. The generator arguments are correct but sample is incorrect. Your solution is good here, but maybe it should also say something like `dojo: invalid generator arguments`.

midden-fabler:
We need to first test the nesting of positional arguments and key-value named arguments.
If they nest, but the whole sample doesn't, something must be wrong with the event sample, so we print the default.
This requires providing valid arguments to a badly written generator, as the event sample is injected by dojo.

bad arguments:
```
>   dojo: invalid generator arguments
>   dojo: nest-need
[[desk=@tas %~] from=@tas hard=?(%.y %.n) gall=?(%.y %.n)]
>   dojo: nest-have
[%~ from=@tas hard=?(%.y %.n) gall=?(%.y %.n)]
> |new-desk
dojo: generator failure
```

bad event sample:
```
>   dojo: invalid generator event sample
>   dojo: nest-need
[now=@da eny=@uvJ bec=[p=@p q=@tas r=?([%da p=@da] [%tas p=@tas] [%ud p=@ud] [%uv p=@uv])]]
> |new-desk %base
dojo: generator failure
```

pkg/arvo/app/dojo.hoon
```diff
@@ -883,10 +882,26 @@
...
+        =/  gav=vase  (slot 12 gat)
+        ?.  (~(nest ut p.ven) | p.gav)
+          ::  event arguments
+          ::
+          ~>  %slog.1^leaf+"dojo: invalid generator event sample"
+          ~>  %slog.1^leaf+"dojo: nest-need"
+          ~>  %slog.0^(skol p.ven)
+          ~>  %slog.1^leaf+"dojo: nest-have"
+          ~>  %slog.0^(skol p.gav)
+          !!
```
joemfb, line 894: These were already like this, but I think it'd be better to put the errors in the trace instead of eagerly slogging them. You can do that with `~_`

joemfb (changes requested):
This is great!

I'd like to see a clearer distinction between these error cases. "invalid generator event sample" means that the generator was miswritten (somebody else's fault), but "invalid generator arguments" means it was miscalled (my fault). Not sure what the wording should be ...

pkg/arvo/app/dojo.hoon
midden-fabler: Like so?

midden-fabler:
> This is great!
> 
> I'd like to see a clearer distinction between these error cases. "invalid generator event sample" means that the generator was miswritten (somebody else's fault), but "invalid generator arguments" means it was miscalled (my fault). Not sure what the wording should be ...

The OP has updated examples.
I'm also not sure on wording.
@sigilante - suggestions?

sigilante (approved):
Looks fine to me.

I didn't know about `+skol`, very nice.

sigilante:
Perhaps:

- `invalid generator event sample (generator incorrect)`
- `invalid generator arguments (supplied sample incorrect)`

sigilante (approved):

midden-fabler:
I just found a way someone could mangle a gate that would fail silently. Switching back to draft until I can provide a useful error.

pkg/arvo/app/dojo.hoon
```diff
@@ -882,13 +883,57 @@
...
+        (slam gat sam)
+      ::  something is wrong
+      ::
+      %-  mean
+      ^-  (list tank)
+      =/  cez=type  [%cell %noun [%cell %noun %noun]]
+      ?.  (~(nest ut cez) | p.som)
+        ::  [ven poz kev] can't nest in som
+        ::
+        :~  'dojo: bad gate lost-argument (generator incorrect)'
```
belisarius222, line 896: Please put the dojo-related error messages below the `nest-need` and `nest-have` messages, since those can be very long and the user would have to scroll up to the see the high-level problem.

midden-fabler: Fixed, with the exception of keyword args, which are validated outside the primary nest check part of this arm. I don't know if/how it would be possible to make the high level error appear below need / have for this case.

PR #6693, opened 2023-06-27
ping: Rewrite to support non-NAT mode

### LINKED ISSUE #6650: Sponsor pings are using too much CPU on galaxies
@philipcmonk reported that one galaxy has a fully pegged CPU keeping up with pings from its sponsees.  Each ping is taking 2ms, which is around the minimal Arvo event processing time.

> 2ms/ping is 1500 ships per 30 seconds, and i think we're close to that

philipcmonk:
Ping takes a lot of resources on sponsor ships, because each ship sends a ping every 25 seconds, which is handled within the event loop.  This especially affects Tlon and all hosting providers above a certain scale, because they may have hundreds or thousands of ships under a single galaxy.

One solution is to process pings much faster, probably by handling them in the runtime.  I think this would be a valuable line of work.

The other solution is to ping much less often, which is also good and correct.  Hosted ships generally have public ports, so they don't need to ping every 25 seconds to keep a pinhole open.  They only need to ping when their port or IP changes.

This PR introduces a mode to the `:ping` app which can be manually enabled to declare that you know that you're not behind a NAT.  In this mode, instead of pinging every 25 seconds, you will check your IP address every five minutes (even that's probably too often), and only if it's changed will you ping your sponsors.  So much for IP.

Your port can change only when you restart vere (barring advanced hosting shenanigans). 
 Ames's `%born` handler now kicks the `:ping` app, so that every time you restart your ship, it will ping its sponsors.

To enable this mode, run `:ping %no-nat`.  To go back to the default behavior, use `:ping %nat`.

Regarding the code, this rewrite disentangles the questions of which ships are to be pinged from how they are to be pinged.  Thus, it has three subsystems:
- `+ships`, which monitors Jael to detect breaches of our sponsors and changes in our sponsorship hierarchy.  It uses this to restart the ping process when anything relevant has changed.
- `+nat`, which pings every 25 seconds.
- `+pub`, which checks IP every 5 minutes and pings sponsors as necessary.

Draft because while this code runs correctly on a fakezod, it has not been tested on mainnet, or indeed on any non-galaxy ship (which is the only case that matters!).

Partially resolves #6650 
Includes #6689

### DIFF
--- a/pkg/arvo/app/ping.hoon
+++ b/pkg/arvo/app/ping.hoon
@@ -43,7 +43,7 @@
       $=  plan
       $~  [%nat ~]
       $%  [%nat ~]
-          [%pub ip=(unit @if)]
+          [%pub ip=(unit @t)]
       ==
   ==
 --
@@ -244,18 +244,14 @@
           `state
         ::
         =*  body  q.data.u.full-file.resp
-        =/  ip  (slaw %if (cat 3 '.' (end [3 (dec (met 3 body))] body)))
-        ?~  ip
-          %-  %:  slog
-                'ping: ip check body not an ip:'
-                >q.data.u.full-file.resp<
-                ~
-              ==
+        ?~  body
+          %-  (slog 'ping: ip check body empty' ~)
           `state
         ::
-        ?:  =(ip.plan.state ip)  `state
+        =/  ip  (end [3 (dec (met 3 body))] body)
+        ?:  =(ip.plan.state `ip)  `state
         ::
-        =.  ip.plan.state  ip
+        =.  ip.plan.state  `ip
         (send-pings our)
       ::
       ++  set-timer
--- a/pkg/arvo/app/ping.hoon
+++ b/pkg/arvo/app/ping.hoon
@@ -209,7 +209,7 @@
         ?:  =(our ship)
           ~
         =/  wire  /pub/(scot %uw nonce.state)/ping/(scot %p ship)
-        `u=[%pass wire %agent [our %ping] %poke %noun !>(~)]
+        `u=[%pass wire %agent [ship %ping] %poke %noun !>(~)]
       ::
       ++  take-pings
         |=  [=wire error=(unit tang)]
@@ -257,14 +257,12 @@
       ++  set-timer
         |=  now=@da
         ^-  (quip card _state)
-        ~&  >  %setting
         =/  =wire  /pub/(scot %uw nonce.state)/wait
         [[%pass wire %arvo %b %wait (add ip-timeout now)]~ state]
       ::
       ++  take-wait
         |=  [our=@p now=@da =wire]
         ^-  (quip card _state)
-        ~&  >  %waiting
         ?.  ?=([%pub @ %wait ~] wire)  `state
         ?.  (once i.t.wire)  `state
         ;<  new-state=_state  rind  check-ip
--- a/pkg/arvo/app/ping.hoon
+++ b/pkg/arvo/app/ping.hoon
@@ -344,12 +344,12 @@
     ?+    wire  `state
         [%nat *]
       ?.  ?=(%nat -.plan.state)  `state
-      ?>  ?=(%poke-ack -.sign)
+      ?.  ?=(%poke-ack -.sign)   `state
       (take-ping:nat now.bowl wire p.sign)
     ::
         [%pub *]
       ?.  ?=(%pub -.plan.state)  `state
-      ?>  ?=(%poke-ack -.sign)
+      ?.  ?=(%poke-ack -.sign)   `state
       (take-pings:pub wire p.sign)
     ==
   [cards this]
--- a/pkg/arvo/app/ping.hoon
+++ b/pkg/arvo/app/ping.hoon
@@ -277,7 +277,7 @@
         (set-timer now)
       --
     --
-%+  verb  &
+%+  verb  |
 ^-  agent:gall
 |_  =bowl:gall
 +*  this  .
--- a/pkg/arvo/app/ping.hoon
+++ b/pkg/arvo/app/ping.hoon
@@ -353,7 +353,7 @@
   |=  [=wire =sign-arvo]
   ^-  [(list card) _this]
   =^  cards  state
-    ?+    wire  ((slog 'ping: strange wire' >wire< ~) `state)
+    ?+    wire  `state
         [%jael %delay ~]
       ?>  ?=(%wake +<.sign-arvo)
       ?^  error.sign-arvo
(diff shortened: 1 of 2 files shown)


pkg/arvo/app/ping.hoon
```diff
@@ -9,145 +9,276 @@
...
+    ::  Subsystem to keep track of which ships to ping across breaches
+    ::  and sponsorship changes
+    ::
+    ++  ships
+      |%
+      ++  rind  (^rind card state)
+      ++  kick
+        |=  [our=@p now=@da]
+        ^-  (quip card _state)
+
```
belisarius222, line 81: blank line

pkg/arvo/app/ping.hoon
```diff
@@ -157,24 +288,44 @@
...
+    =^  cards  state  (kick:ships our.bowl now.bowl)
+    [cards this]
+  ?:  =(q.vase %nat)
+    =.  plan.state  [%nat ~]
+    =^  cards  state  (kick:ships our.bowl now.bowl)
+    [cards this]
+  ?:  =(q.vase %no-nat)
+    =.  plan.state  [%pub ~]
+    =^  cards  state  (kick:ships our.bowl now.bowl)
+    [cards this]
```
yosoyubik, line 328: thinking if this could be tweaked a bit to avoid the repeating `=^ cards state`

pkg/arvo/app/ping.hoon
```diff
@@ -9,145 +9,276 @@
...
+      ++  send-pings
+        |=  our=@p
+        ^-  (quip card _state)
+        :_  state
+        %+  murn  ~(tap in ships.state)
+        |=  =ship
+        ?:  =(our ship)
+          ~
+        =/  wire  /pub/(scot %uw nonce.state)/ping/(scot %p ship)
+        `u=[%pass wire %agent [our %ping] %poke %noun !>(~)]
```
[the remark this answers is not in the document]
belisarius222: Probably not, but it illustrates that it's a `(unit)` value.  I think it's fine, stylistically.

pkg/arvo/app/ping.hoon
belisarius222: I just tried to shrink this function in a few different ways and it was harder than I had expected.  I do think it would be cleaner as a `?+  q.vase`, though.

pkg/arvo/app/ping.hoon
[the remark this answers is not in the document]
philipcmonk: It's required unless you add an explicit cast. For short functions like this I often will do this, it feels less noisy than adding a whole other line

pkg/arvo/app/ping.hoon
philipcmonk: Yeah, I can put the whole thing in =^ like on-arvo and on-agent, which would be more stylistically consistent

belisarius222: I think this might work:

```hoon
|=  [=mark =vase]
|^  ?+  q.vase  `this
      %kick    kick  ::  NB: ames calls this on %born
      %nat     kick(plan.state [%nat ~])
      %no-nat  kick(plan.state [%pub ~])
    ==
++  kick
  =^  cards  state  (kick:ships our.bowl now.bowl)
  [cards this]
--
```

philipcmonk: I tried a few options, and I think I'm going to go with this:
```
  =^  cards  state
    ?:  =(q.vase %kick)  :: NB: ames calls this on %born
      (kick:ships our.bowl now.bowl)
    ?:  =(q.vase %nat)
      =.  plan.state  [%nat ~]
      (kick:ships our.bowl now.bowl)
    ?:  =(q.vase %no-nat)
      =.  plan.state  [%pub ~]
      (kick:ships our.bowl now.bowl)
    `state
  [cards this]
```

Everything with `?+` was harder to read, because it needs either queenside or a `|^`, which is a lot of visual indirection for something this simple.  If you wanted to go further, I think you'd stop using `%noun` mark and make a real one, so that it can be a `?-` and is distinct from the "handling a ping" case.  Then you can just `=.  plan.state` a `?-` and do one `kick` after the whole thing.  But again, I think this is too much indirection for something this simple.

Also realized these should be restricted to local pokes, so I added that check.

philipcmonk:
icanhazip.com sometimes gives ipv6 addresses in the browser.  It seems to give ipv4 to `:ping`'s request, and we must have an ipv4 address for urbit to work, but I think this means it's safer to store the text of the response instead of parsing it as an IP.  If the text of the response changes, we interpret that as having possibly moved IP.  I'll make that change.

philipcmonk:
This is the testing regime I performed today:
- Tested a normal moon on my local NAT'd network.  Switched to hotspot, which is a different NAT'd network.  Everything worked like normal.
- Turned this moon onto `%no-nat` mode.  As expected, it soon became unreachable.
- By inspection, I verified that this moon correctly recognized when its IP changed as it changed network.  It pinged its sponsors at that time.
- Started another moon on a server with no NAT.  Worked normally.  Turned on `%no-nat` mode, and this this continued to work.
- Restarted that moon without choosing the port.  Verified that the connection continued to work (this is where I found and fixed a bug where `%no-nat`'s pings were going to ourself instead of our sponsors!).
- Booted a fresh mainnet planet, then booted a moon off of it.
- Breached the planet, and observed that the moon correctly restarted its ping process.

One test I have not explicitly performed is changing the sponsorship hierarchy by escaping.  This is relatively difficult to test because it requires access to at least two stars, ideally under different galaxies.

The test I performed instead is that I hardcoded a different list of ships as the new sponsorship hierarchy, so that when `:ping` reloaded, it would exercise the ship-removal logic.  That logic is only one line, but it's important to at least run every line of code.  It worked correctly: I observed it start to ping only the new ships and not the old, and I observed that the Jael subscription had been updated to have only the new ships.

The other thing that could go wrong is that we could fail to recognize a sponsorship change, but (1) Jael has no changes, and that logic has worked on mainnet for a long time, and (2) we kick the pings on absolutely any response from Jael (this part was tested by the breach test earlier), so we couldn't errantly drop that update.

Based on this, I believe this is ready to go, and I'm marking it for review.

yosoyubik (approved):
LGTM

To add another data point to the tests, I've also tried this PR on two L2 planets behind a NAT and with a public IP and confirmed that everything worked as expected (turning `%no-nat` on the NAT'd planet, and waiting for the pinhole to close makes the ship unreachable; turning it again to `%nat` and communication is re-established; the ship on the public IP works fine on the `%no-nat` mode)

PR #6691, opened 2023-06-26
%yard, a developer commons

sigilante:
`%yard` is a proposed desk to offload some non-essential `%base` functionality and provide ready-to-hand duplicates of essential marks and libraries for developers.

`%yard` is not subject to kelvin versioning.

See discussion at [this gist](https://gist.github.com/sigilante/beddcdfdd9836ede6403d4bb0ac1d43c).

PR #6695, opened 2023-06-28
%base: delete /app/gaze and /ted/naive-csv

pkova:
`/app/gaze` is only used by the network explorer and the code in %base has been outdated for years. The code used by the network explorer lives [here](https://github.com/urbit/network-explorer/blob/master/gaze/gaze.hoon). Tools for aggregating statistics certainly don't belong in %base.

`/ted/naive-csv` is another example of this. This one is more controversial to delete since I'm pretty sure Tlon still uses it somewhere. @drbeefsupreme can you grab this thread to a separate desk? After that we can remove it.


yosoyubik:
I think `naive-csv` is going to be part of the `%yard` desk: https://github.com/urbit/urbit/pull/6691

pkova:
Asked around and it seems like `/ted/naive-csv` is not used by Tlon anymore, so these should be safe to delete. It's a breaking change so we will put it in the kelvin 412 release.

PR #6697, opened 2023-06-28
json: remove deprecated JSON parser calls

### LINKED ISSUE #6474: Remove deprecated JSON parser calls
As of #6463, `+de-json:html` and `+en-json:html` are deprecated. All uses of them should be replaced by `+de:json:html` and `+en:json:html` respectively.

MarcusMiguel:
Resolves #6474


ashelkovnykov (commented):
LGTM, but holding off on approval to resolve one nit and the question below.

This PR makes it safe to completely remove `+en-json:html` and `+de-json:html` - in the kernel. Can we go ahead and delete them, and add it to the release notes for the next kelvin decrement release that userspace devs need to switch to `+en:json:html` and `+de:json:html` respectively? @joemfb @belisarius222 @philipcmonk

belisarius222:
@ashelkovnykov Removing arms from the kernel needs to be done in a Kelvin bump.  That should be in a separate PR.  I think this PR could go in as-is, since it doesn't change any interfaces.  Does that sound right to you?

ashelkovnykov (approved):

ashelkovnykov:
@belisarius222 Sounds good to me

> it doesn't change any interfaces

That's right, only usage changes in this PR

joemfb (approved):
this will do

PR #6700, opened 2023-06-30
ames: Don't check for lane staleness

philipcmonk:
When forwarding packets, we check that we've heard a message along that lane in the last hour, and if we haven't, then we don't use that lane.  This does not work if you enable `%no-nat`, because you'll only receive very occasional pings.  As a result, galaxies will stop forwarding for `%no-nat` ships after one hour.

Those checks were added because we knew that all our children would ping us every 25 seconds, so no messages for one hour was a sure-fire indicator that they weren't online.  This is no longer the case, and the short-circuit is only an optimization anyway.  This PR removes those checks.

If we wanted to re-enable this kind of behavior, we could add a once-a-day ping even when your lane hasn't changed in `%no-nat` mode, and then declare a lane stale if we haven't heard anything in two days.  However, I don't believe this behavior is important.

Tlon has deployed this fix to ~dem and ~fen, which are the two galaxies we host ships under.  Since likely all ships that have enabled `%no-nat` are under these two galaxies, I don't believe this is super urgent to be deployed more broadly -- although if ~zod in particular were affected, we could deploy it there as well.  We should, however, recommend that `%no-nat` not be used until it's deployed.

Re: testing, as mentioned, this is deployed to ~dem and ~fen already, and it resolved the connectivity issues we saw.  It also caused `.^(* %ax /=//=/peers/~ship-name/forward-lane)` to produce a lane instead of `0`, which is what the runtime needs to forward packets.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2758,10 +2758,6 @@
               ?:  for
                 event-core
               (try-next-sponsor sponsor.peer-state)
-            ::  if forwarding, route must not be stale
-            ::
-            ?:  &(for (lth last-contact.qos.peer-state (sub now ~h1)))
-              (try-next-sponsor sponsor.peer-state)
             ::
             ?~  route=route.peer-state
               %-  (ev-trace rot.veb final-ship |.("no route to:  {<ship>}"))
@@ -4977,8 +4973,6 @@
           ~
         [%& gal]~
       =;  zar=(trap (list lane))
-        ?:  (lth last-contact.qos.u.peer (sub now ~h1))
-          $:zar
         ?~  route.u.peer  $:zar
         =*  rot  u.route.u.peer
         ?:(direct.rot [lane.rot ~] [lane.rot $:zar])


belisarius222:
I wonder if there is more risk of routing loops if lanes don't time out.  If not, this seems a lot better than the current behavior.

philipcmonk:
If that happens, we should just add a rule that only galaxies forward packets (if it doesn't already exist)

belisarius222:
Ok.  Now I'm wondering, if we do get a routing loop, is there a manual way to break it?  I would hate to be in a position where the only way out is a code change that we would have to write on the fly.  It does seem at least more likely for a routing loop to happen if routes never time out.  Timing out after a day or two might also be better than indefinite.

philipcmonk:
We can't time out after a day unless we add another timer to :ping.

The manual way to break a routing loop would be to move that ship to
another port, or if you don't have access to that side, just stop
forwarding to that ship.

belisarius222:
Ok, then I think the hardest case to deal with would be if there's a routing loop involving a planet (or moon), where core devs only have access to the galaxy but not the planet.  Breaking the loop would then require us to load new code onto the galaxy to special-case the planet so we drop packets headed toward it.

This could be made into a command-line flag for Vere, where you could pass in a list of ships to avoid forwarding to.  Alternatively, we could add a task to Ames that would let us delete a lane for a ship -- that way, we could just `|pass` that move on the galaxy, and next time Vere scrys into Arvo for a lane for that ship (which it does every couple minutes), it would get an empty list and drop the packet.  Does that sound right?

philipcmonk:
I don't understand how any of that is simpler than not forwarding if you're not a galaxy.

philipcmonk:
Are you waiting on that to merge this?  We just had an issue where a small number of our hosted ships are under ~zod which doesn't have this fix.

belisarius222:
Yes, I've been waiting to merge this until I'm more confident about routing loops.  The problem is that if we do get stuck in a routing loop, it's not that easy to get out of, and if there were suddenly a lot more routing loops on the network, that could effectively take down the network until we have a fix.

In this discussion you've suggested two relatively easy things we could do to prevent routing loops and kill them quickly when they arise:

1) Add a check to Vere so that only galaxies forward packets.
2) Add some kind of functionality to allow a galaxy to stop forwarding to a particular ship.

I think we should build both of those before deploying this PR, since it's hard to know just how many more routing loops would spring up with this change, which makes it hard to evaluate how severe that risk is.  Once we have both of those things in place, even if a bunch of routing loops come up, we could kill them manually quite rapidly.

We need the second change because we can't guarantee that everyone will update their Vere to include the first change, so an old ship could still become part of a routing loop.

Vere change 1) should be trivial.  I'm not sure how to best accomplish change 2).  Could we do it in Arvo somehow?  If we had the ability to manually delete a lane from the Ames vane, that could let us clear a lane that's become part of a loop.  I think that might be enough.  

We could instead lean on the existing `|ruin` functionality to tell a galaxy to ignore some other ship, although that's a bit too permanent for my tastes.

I'm open to ideas on what's the easiest way to make sure we can quickly get ourselves out of routing loops when they arise.

matthew-levan:
Quick update here: we've merged a [PR](https://github.com/urbit/vere/issues/492) to achieve "1. Add a check to Vere so that only galaxies forward packets." but still need a PR for "2. Add some kind of functionality to allow a galaxy to stop forwarding to a particular ship." cc @pkova

pkova:
@belisarius222 says that we could accomplish 

`2. Add some kind of functionality to allow a galaxy to stop forwarding to a particular ship.`

by adding a task to ames to delete all lanes for a specific ship. This could be a good contributor task.

philipcmonk:
I would add this to the existing `%stir` (assorted debugging commandds) task

PR #6703, opened 2023-07-03
mar: better html-related conversions

### LINKED ISSUE #6680: HTML-related mark conversions are a mess
The `%hoon` mark includes a conversion to the `%elem` mark, which no longer exists in `%base` or the pkg/base-dev folder in the repo, but only in `pkg/landscape/mar` and in `%base` in the `/mar/x-htm/elem` folder.  The conversion to `%elem` won't find the `x-htm/elem` mark, and I don't know why that exists.

There is an `%htm` mark in addition to the `%html` mark -- that could be fine, since they're two different file extensions that people use for HTML files, but for some reason the two mark files are subtly different from each other.  My guess is they diverged accidentally over time.

In addition to these marks, there is the `%hymn` mark, which is different from the other ones.  Why do we need this in addition to the `%html` mark?  I can't prove this, but I have a feeling the answer dates back to 2016 and involves Ford "renderers", which were deleted a few years ago.

It strikes me that we should have just an `%html` mark and an `%htm` mark that is effectively a copy of the `%html` mark -- one could even import the other with a `/=` Ford rune and return that to make sure they don't get out of sync again.

See also:
https://stackoverflow.com/questions/1163738/should-you-use-htm-or-html-file-extension-what-is-the-difference-and-which-f

MarcusMiguel:
Resolves #6680

Removes: `%x-htm`, `%elem`, `%x-htm-elem` marks and conversions.

Turns the `%htm` mark into a copy of `%html`.

Adds a `%xhtml` mark that's also a copy of `%html`.

Leaves the `%hymn` manx mark.


belisarius222:
This looks good to me.  Let's retarget this for the next Kelvin, though, in case people are using any of these marks right now.

I just tried to change the target branch to `next/kelvin/412`, but github's UI is spazzing out on me.

PR #6704, opened 2023-07-03
dojo: fix bad gall scries with | on non-null =dir 

### LINKED ISSUE #5879: |i from non-null =dir crashes dojo
Reproduction:
```
=dir /=garden=
|i
```

This will cause dojo to crash, unlink, and relink.  It shouldn't.

I don't know whether this happens with any non-null `=dir`, or just with some more specific ones.

silnem2:
Resolves #5879

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -115,7 +115,7 @@
 ::  |parser-at: parsers for dojo expressions using :dir as working directory
 ::
 ++  parser-at
-  |=  [our=ship dir=beam]
+  |=  [our=ship now=@da dir=beam]
   |%
   ++  default-app         %hood
   ++  hoon-parser         (vang | (en-beam dir))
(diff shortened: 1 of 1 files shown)


belisarius222:
Can you explain how this works, please?  I'm having trouble understanding what the issue was and how this resolves it.

silnem2:
#5879 is caused by the fact that `+to-command`, which is called every time you type `|` followed by a letter into the dojo prompt, contains scries which will return null if the case is anything other than `now`. `+to-command` gets the beak for the scries by modifying `=dir`, but doesn't change the case, so (as @ynx0 notes in the issue) if `=dir` has a specific case the `%gu` scry will cause dojo to crash and relink (e.g.  `|i` on `=dir /=garden=` will trigger the crash, but `|i` on `=dir /=garden` won't).

This PR resolves the issue by having `+to-command` scry using the `now` case unconditionally rather than using the case from `=dir`, which prevents the crash.

belisarius222 (approved):

Issue #6720, opened 2023-07-12
|uninstall and |suspend function incorrectly (kiln)


hanfel-dovned:
**Describe the Bug**
There's some general weirdness with the relationship between kiln's `zyn` state and the `%zest` notes that kiln passes to Clay. This manifests in the following issues:
- `|uninstall` does not prevent the app publisher from pushing changes to the subscriber's desk (at least according to kiln output).
- `|suspend`ing an app *does* prevent the app publisher from pushing changes to the subscriber's desk, in such a way that `|revive`ing doesn't restore, preventing future app updates

**Reproducing the `|uninstall` bug:**
1. Create two fake ships, ~zod and ~dev
2. Install and publish a desk on ~zod via mount/commit/install our/treaty publish
3. Install that desk on ~dev: `|install ~zod %hello`
4. Uninstall the desk on ~dev: `|uninstall %hello`
5. Commit changes to the desk on ~zod
6. Notice in ~dev's terminal that kiln still produces output suggesting that it downloaded the update for this desk from the publisher. (See [this line](https://github.com/urbit/urbit/blob/10ef0c23410bf275718d42d848995ecbcb690dc5/pkg/arvo/lib/hood/kiln.hoon#L1131) in kiln.) However, if you mount the desk and check your filesystem, you'll see that these changes didn't actually get committed.

**Reproducing the `|revive` bug:**
1. Create two fake ships, ~zod and ~dev
2. Install and publish a desk on ~zod via mount/commit/install our/treaty publish
3. Install that desk on ~dev: `|install ~zod %hello`
4. Suspend the desk on ~dev: `|suspend %hello`
5. Revive the desk on ~dev: `|revive %hello`
6. Commit changes to the desk on ~zod
7. Notice in ~dev's terminal that there is no kiln output and that the desk was not updated.

**Expected Behavior**
- As per the [docs](https://operators.urbit.org/manual/os/dojo-tools#uninstall), `|uninstall` should ignore updates from the desk publisher. I would expect that uninstalling an app would terminate my relationship with its distributor and not allow them to put data on my ship anymore, even if that data is inert. It seems likely that this data isn't in fact getting put on the ship, but in that case kiln shouldn't be outputting that it has.
- `|suspend` should not unsubscribe permanently in such a way that `|revive` doesn't recover from. I don't expect to have to reinstall an app after turning it back on in order to get updates.

**System:**
 - Ubuntu 18.04.5
 - `vere 2.11`, `zuse 413k`
[...]

PR #6721, opened 2023-07-13
kiln: have +poke-uninstall unsync desks with %dead zest 

silnem2:
If `|uninstall` is run on a desk with a `zest` of `%dead` (e.g. a desk that has been `|suspend`ed), `+poke-uninstall` will no-op and fail to cancel the desk's syncs if it has any. This PR rewrites `+poke-uninstall` such that it will cancel a desk's syncs irrespective of its `zest`, which was the behavior of `|uninstall` prior to #6382 . I suspect that #6720 is related, but even if not I believe that this change reflects the desired behavior of `|uninstall`. Also fixes a deprecated `%cd` scry in `+poke-suspend`.

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -795,7 +795,7 @@
   %+  turn
     %+  skim  desks
     |=  dek=desk
-    ?:  (~(has in .^((set desk) %cd /(scot %p our)/base/(scot %da now))) dek)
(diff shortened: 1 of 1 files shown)


belisarius222:
This code LGTM.  Because it introduces a breaking change to the behavior of Kiln, it should be part of a Kelvin bump.  I have retargeted this PR at the next/kelvin/412 branch.  It looks like the develop branch has run ahead of that branch, so we should merge develop into next/kelvin/412 and then check that the commits in this PR look right before merging it.

PR #6722, opened 2023-07-14
Implement 'swap' and 'edit'

del-distributed:
This commit introduces two new gates: 'swap' and 'edit' for text processing tapes and cords, respectively.

Both functions take four parameters: a source string, a substring to find, a replacement string, and a boolean that dictates whether to replace the first match or all matches.

Please review for accuracy and efficiency of the implementation. Any suggestions for improvements or optimizations are welcome.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -4184,18 +4184,21 @@
 ++  edit                                                :: cord replace
   |=  [src=@t get=@t set=@t all=?]
   ^-  @t
-  ?:  =(get set)  src
-  =+  cut=(lent (trip get))
+  =/  metg  (met 3 get)
+  =/  mets  (met 3 set)
+  ?~  metg  !!
+  =/  len   (met 3 src)
+  =/  pos  0
   |-
-  =+  i=(find (trip get) (trip src))
-  ?~  i  src
+  ?:  =(pos len)
+    src
+  ?.  =((cut 3 [pos metg] src) get)
+    $(pos +(pos))
+  =.  src
+    (rap 3 ~[(cut 3 [0 pos] src) set (rsh [3 (add pos metg)] src)])
   ?.  all
-    (crip (into (oust [+.i cut] (trip src)) +.i set))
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -4181,6 +4181,21 @@
...
+  |-
+  =+  i=(find (trip get) (trip src))
+  ?~  i  src
+  ?.  all
+    (crip (into (oust [+.i cut] (trip src)) +.i set))
+  ?~  (find (trip get) (trip set))
+    $(src (crip (into (oust [+.i cut] (trip src)) +.i set)))
+  ~&  >>  'infinite loop: set must not contain get'
+  src
+::
```
ashelkovnykov, line 4198: - For cords, using the binary arithmetic operations is always going to be significantly more efficient
- You likely want to crash if `get` is empty
- Since we expect `get` and `set` to not be identical, I think it might actually be better to just never check if they're identical. A user who does this almost certainly has an upstream error that he should correct himself, rather than us correcting it for him silently in `edit` at the cost of a direct comparison for every single usage at every call site.

```
++  edit                                                :: cord replace
  |=  [src=@t get=@t set=@t all=?]
  ^-  @t
  =/  metg  (met 3 get)
  =/  mets  (met 3 set)
  ?~  metg  !!
  =/  len   (met 3 src)
  =/  pos  0
  |-
  ?:  =(pos len)
    src
  ?.  =((cut 3 [pos metg] src) get)
    $(pos +(pos))
  =.  src
    (rap 3 ~[(cut 3 [0 pos] src) set (rsh [3 (add pos metg)] src)])
  ?.  all
    src
  $(pos (add pos mets), len (add mets (sub len metg)))
```

ashelkovnykov (changes requested):

pkg/arvo/sys/hoon.hoon
del-distributed: Ah, bit arithmetic. Enlightening, thank you.

I've omitted ++swap for now, and applied the improvements to ++edit.

PR #6724, opened 2023-07-15
kiln: avoid downloading updates for uninstalled desks

MarcusMiguel:
Resolves the `uninstall` case in #6720

Currently, if a foreign desk is uninstalled, the first update that the publisher commits will produce the following output in dojo (this happens because we asked the other ship to notify us when an update happens before uninstalling the desk, so the data will eventually come):
``` 
kiln: downloading update for %desk from ~publisher/%desk
 ```
However, the update eventually no-ops, and the desk remains unchanged.

This PR solves the problem by checking if the desk is still installed before downloading it.


belisarius222:
Code LGTM.  Retargeting to the next Kelvin since it's a breaking change to Kiln.  The develop branch should be merged into the next kelvin branch before we merge this.

PR #6730, opened 2023-07-18
Use released urbit binaries

ilyakooo0:


ilyakooo0:
@pkova this fixes the regression in CI that we discussed

pkova (approved):

PR #6733, opened 2023-07-19
zuze: remove deprecated arms

MarcusMiguel:
See #6697 for context


ashelkovnykov (changes requested):
This PR includes #6697 as a commit. I would recommend that you remove it, as:
- Including it introduces the potential for a mismatch between the versions of the commit in the two branches, which then needs to be addressed during merge
- It'll pollute the git commit history
- #6697 will make it into the code on its own before this PR anyways

Aside from this, LGTM

MarcusMiguel:
@pkova  still broken: `next/kelvin/412` doesn't have #6697 yet

PR #6738, opened 2023-07-24
ames: consolidate dead flows to a single behn timer

pkova:
As discussed with @yosoyubik and @joemfb out of band.

I tested this with 50 000 dead flows. Without consolidation these flows resulted in a constant 30 % cpu usage. Consolidating the timers led to a CPU usage of almost 0 with a 100 % spike for a few seconds every two minutes.

The retry interval is the normal ~m2, we can make it configurable later.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2406,15 +2406,15 @@
             =.  dead.ames-state.cor  `[~[/ames] /dead-flow `@da`(add now ~m2)]
             (emit:cor duct %pass /dead-flow %b %wait `@da`(add now ~m2))
           %-  ~(rep by peers.ames-state)
-            |=  [[=ship =ship-state] core=_event-core]
-            ^+  event-core
-            =/  peer-state=(unit peer-state)  (get-peer-state:core ship)
-            ?~  peer-state  core
-            %-  ~(rep by snd.u.peer-state)
-            |=  [[=bone =message-pump-state] cor=_core]
-              ?.  =(~m2 rto.metrics.packet-pump-state.message-pump-state)
-                cor
-              abet:(on-wake:(abed-peer:pe:cor ship u.peer-state) bone error)
+          |=  [[=ship =ship-state] core=_event-core]
+          ^+  event-core
+          =/  peer-state=(unit peer-state)  (get-peer-state:core ship)
+          ?~  peer-state  core
+          %-  ~(rep by snd.u.peer-state)
+          |=  [[=bone =message-pump-state] cor=_core]
+          ?.  =(~m2 rto.metrics.packet-pump-state.message-pump-state)
+            cor
+          abet:(on-wake:(abed-peer:pe:cor ship u.peer-state) bone error)
         ::
         ?.  ?=([%recork ~] wire)
           =/  res=(unit ?([%fine her=ship =^wire] [%pump her=ship =bone]))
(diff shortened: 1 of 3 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1756,7 +1756,10 @@
         14+(state-13-to-14:load:adult-core +.u.cached-state)
       =?  u.cached-state  ?=(%14 -.u.cached-state)
         15+(state-14-to-15:load:adult-core +.u.cached-state)
-      =?  u.cached-state  ?=(%15 -.u.cached-state)
+      =^  moz  u.cached-state
+        ?.  ?=(%15 -.u.cached-state)  [~ u.cached-state]
+        ~>  %slog.0^leaf/"ames: init dead flow consolidation timer"
+        :-  [[/ames]~ %pass /dead-flow %b %wait `@da`(add now ~m2)]~
```
joemfb, line 1762: Do we need to duplicate this timer initialization somewhere else to catch fresh boot?

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3976,9 +3993,11 @@
               ::
               =?  peer-core  !=(~ next-wake.state)
                 (pu-emit %b %rest (need next-wake.state))
-              ::  set new timer if non-null
+              ::  set new timer if non-null and not at max-backoff
               ::
               =?  peer-core  ?=(^ new-wake)
+                ?:  =(~m2 rto.metrics.state)
```
joemfb, line 3999: Worth noting here (and above) that the `~m2` literal is important, we don't want to actually use `max-backoff` and accidentally consolidate the `:ping` app flow.

joemfb (commented):
This is a nice, simple approach, and should dramatically reduce the number of retry timers we have to maintain (and therefore the events logged for retry). At some point, I think we should develop "offline" heuristics for a peer and backoff even further.

A couple things:

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2396,6 +2401,21 @@
...
+          %-  ~(rep by peers.ames-state)
+            |=  [[=ship =ship-state] core=_event-core]
+            ^+  event-core
+            =/  peer-state=(unit peer-state)  (get-peer-state:core ship)
+            ?~  peer-state  core
+            %-  ~(rep by snd.u.peer-state)
+            |=  [[=bone =message-pump-state] cor=_core]
+              ?.  =(~m2 rto.metrics.packet-pump-state.message-pump-state)
+                cor
+              abet:(on-wake:(abed-peer:pe:cor ship u.peer-state) bone error)
```
joemfb, line 2417: Style nit: there are a couple extraneous layers of indentation here.

pkova: Fixed

pkg/arvo/sys/vane/ames.hoon
pkova: Added a comment in a3e75959b13e4f2b0c61ca909902e9f1e05b1d76

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -5074,6 +5110,8 @@
     |=  old=ames-state-15
     ^-  ^ames-state
     %=    old
+      cong  [cong.old `[~[/ames] /dead-flow `@da`(add now ~m2)]]
```
joemfb, line 5113: How is this picked up and turned into a move during migration? Seems like it should be initialized to `~` here and then toggled in `+on-born`.

pkova: It happens here: 

https://github.com/urbit/urbit/blob/82d4e2a265adb3aa3e1ab05a9c4365ae885eb026/pkg/arvo/sys/vane/ames.hoon#L1761-L1765

joemfb: I see, thanks.

joemfb (approved):
this will do

PR #6740, opened 2023-07-25
lull, ames: add %tame task to delete a route for a ship

pkova:
I decided to make a new task instead of using `%stir` as suggested by @philipcmonk here: https://github.com/urbit/urbit/pull/6700#issuecomment-1648475369. We need to pass a ship argument (`%stir` takes just a `@tas`) and unbelievable nonsense can erupt in `+load` when you change a type of a task so I think this is better.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2303,6 +2303,21 @@
             =/  rcvr  [ship her-life.channel.peer-core]
             "plea {<sndr rcvr bone=bone vane.plea path.plea>}"
         abet:(on-memo:peer-core bone plea %plea)
+      ::  +on-tame: handle request to delete a route
+      ::
+      ++  on-tame
+        |=  =ship
+        ^+  event-core
+        ?:  =(%czar (clan:title ship))
+          %-  %+  slog
+            leaf+"ames: bad idea to %tame galaxy {(scow %p ship)}, ignoring"
+          ~
+          event-core
+        =/  peer-state=(unit peer-state)  (get-peer-state ship)
+        ?~  peer-state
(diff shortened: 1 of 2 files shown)


belisarius222 (changes requested):
Please rename %bane to something that is not a homonym of %behn.  Otherwise we'll talk about "passing a bane task to ames" and it'll get confusing.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2298,6 +2298,21 @@
             =/  rcvr  [ship her-life.channel.peer-core]
             "plea {<sndr rcvr bone=bone vane.plea path.plea>}"
         abet:(on-memo:peer-core bone plea %plea)
+      ::  +on-bane: handle request to delete a route
+      ::
+      ++  on-bane
+        |=  =ship
+        ^+  event-core
+        ?:  =(%czar (clan:title ship))
```
belisarius222, line 2306: I don't think we need the galaxy check.  We don't use galaxy lanes for anything iirc, but I don't see the benefit of adding logic to deny a request to delete one if it's there.

pkova: From my local testing it appears that if set the `route` of a galaxy to `~` you'll never be able to talk to that galaxy again.

belisarius222:
That's weird. Ames should just use a special galaxy lane when emitting %send effects to Vere, whether it has an IP/port pair for the galaxy or not.

pkova:
I think the reason for that is here: https://github.com/urbit/urbit/blob/65b257b96a3a65f117761cb08d5c17028ac8c137/pkg/arvo/sys/vane/ames.hoon#L2912-L2914

`+try-next-sponsor` just no-ops in `+send-blob` if the route is `~` and the ship is it's own sponsor.

PR #6741, opened 2023-07-26
eyre: add HTTP scrying

lukechampine:
This is an Eyre-side implementation of [UIP 0106](https://github.com/urbit/UIPs/blob/main/UIPS/UIP-0106.md), enabling scrying via browser, `curl`, etc. For example:

```
$ curl localhost:80/_~_/cx/===/sys/kelvin
[%zuse 413]
```

Mark conversion is also supported:
```
$ curl localhost:80/_~_/cx/===/gen/verb/hoon.hymn
<html><head><title>Source</title><script src="//cdnjs.cloudflare.com/ajax/libs/codemirror/4.3.0/codemirror.js"></script><script src="/lib/syntax/hoon.js"></script><link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/codemirror/4.3.0/codemirror.min.css" /><link href="/lib/syntax/codemirror.css" rel="stylesheet" /></head><body><textarea id="src">::  Tell app to print what it&#39;s doing
::
::  For apps that use lib/verb, :app +verb toggles verbosity.
::
:-  %say
|=  [* arg=?(~ [%bowl ~]) ~]
[%verb ?~(arg %loud %bowl)]
</textarea><script>CodeMirror.fromTextArea(src, {lineNumbers:true, readOnly:true})</script></body></html>              
```

As in Dojo, scry paths may be partially- or fully-qualified, and the response's `Cache-Control` is set accordingly.

By default, conversion gates are loaded from `%base`. When scrying Clay (except 'buc' paths), the gate will be loaded from the desk in the scry path. When scrying a Gall agent (not Gall itself), the gate will be loaded from the agent's desk.

Once this is merged, I'll proceed with the Vere-side implementation so that these scries don't require injecting an event.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1,7 +1,7 @@
 !:
 ::  lighter than eyre
 ::
-|=  our=ships
+|=  our=ship
 =,  eyre
 ::  internal data structures
 ::
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1,7 +1,7 @@
 !:
 ::  lighter than eyre
 ::
-|=  our=ship
+|=  our=ships
 =,  eyre
 ::  internal data structures
 ::
@@ -3313,8 +3313,8 @@
   ?~  u  [%| "invalid scry path"]
   ::  perform scry
   ::
-  ?~  res=(rof ~ u.u)  [%| "failed scry"]
-  ?~  u.res            [%| "no scry result"]
+  ?~  res=(rof ~ /eyre u.u)  [%| "failed scry"]
+  ?~  u.res                  [%| "no scry result"]
   =*  mark   p.u.u.res
   =*  vase   q.u.u.res
   ::  convert to mime via ext
@@ -3342,7 +3342,7 @@
         %c
       [%& q.beam]
         %g
-      =/  res  (rof ~ %gd [our q.beam da+now] /$)
+      =/  res  (rof ~ /eyre %gd [our q.beam da+now] /$)
       ?.  ?=([~ ~ *] res)
         [%| "no desk for app {<q.beam>}"]
       [%& !<(=desk q.u.u.res)]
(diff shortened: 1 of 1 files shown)


lukechampine:
One issue I encountered: what to do about non-lowercase scry paths? It ought to be possible to `curl localhost:80/_~_/cx/===/README.md`, but this currently fails with `invalid scry path`. (`readme.md` fails too.)

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -993,6 +1002,107 @@
...
+      |=  a=path
+      ^-  ?
+      =/  vez  (vang | (en-beam [our %base da+now] ~))
+      =/  bek=(unit tyke)  (rush (spat a) ;~(pfix fas gash:vez))
+      ?&
+        ?=([~ [* * * *]] bek)
+        !=(~ i.u.bek)
+        !=(~ i.t.u.bek)
+        !=(~ i.t.t.u.bek)
+      ==
```
lukechampine, line 1071: surely there's a better way...?

joemfb: `?=([~ ^ ^ ^ *] bek)`

matthew-levan:
@belisarius222 @philipcmonk was there a resolution on what to do with non-lowercase-named files in Clay?

lukechampine:
Updated to support scrying via Eyre's scry handler. Now you can do:
```
.^(mime %ex /=//=/'_~_'/cx/~zod/base/~2023.7.19..01.58.53..68f8/sys/kelvin)
```
which is...not the prettiest, but it does work.

You also can't use `===` directly, you have to enclose it in quotes:
```
.^(mime %ex /=//=/'_~_'/cx/'==='/sys/kelvin)
```
(IIUC, this is a Dojo thing -- Dojo parses paths with `poor`, which only allows `=` at the start of a path, not in the middle.)

Anyway, Vere is probably going to be the only consumer of this API, so the ergonomics aren't critical.

lukechampine:
@pkova can you change the target branch to `next/kelvin/411`? GitHub won't let me.

jalehman:
@pkova why did this get closed?

pkova:
Looks like github deleted the `next/kelvin/412` branch and with it autoclosed this.

lukechampine:
fixed conflicts, build is failing though, idk why

pkova:
It's because you forgot the wire in your `rof` calls. I fixed it.

@belisarius222 this is good to go into 411 right? I have the curl examples above work as advertised.

belisarius222:
@pkova Yes, it looks good to me.

PR #6742, opened 2023-07-26
Add a list repeater gate.

sigilante:
A version of `++reap` appropriate for lists that wish to remain flat lists (rather than a list of lists).

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -679,6 +679,13 @@
   ?:  =(~ t.a)  ~
   [i.a $(a t.a)]
 ::
+++  snog                                                ::  glue together a copies of list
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -679,6 +679,13 @@
   ?:  =(~ t.a)  ~
   [i.a $(a t.a)]
 ::
+++  snog                                                ::  glue together a copies of list
+    |*  [a=@ b=(list *)]
+    =/  c  *_b
```
pkova, line 684: What's going with this unused `c` here?

joemfb:
FWIW, I don't think this is worth adding a new name: `(zing (reap n list))` is readable and not that much more inefficient.

PR #6743, opened 2023-07-28
clay: propagate [~ ~] properly from read-x

pkova:
Looks like the introduction of `+read-q` led to a bug where a `[~ ~]` from `+read-q` was not propagated correctly from `+read-x`, instead becoming `~`. This results in stuff like `-read %x our %base da+now /i-dont-exist/txt` blocking forever even though we know the file does not exist.


joemfb (approved):
Good catch!

PR #6749, opened 2023-08-10
eyre: avoid +got'ing a channel that doesn't exist

Fang-:
`+on-gall-response` might detect the symptom of a bug, where we are getting a `%fact` for a channel that has since been deleted. (Meaning that subscription should have been cleaned up, and we shouldn't have received the `%fact`.) We want to issue a `%leave`, but need to take care to do so with the same identity that the subscription was opened on.

Previously, we would forcefully get the identity from the non-existent channel, resulting in a crash. Now, we encode the identity into the wire instead, so that we may retrieve it from there, even when the channel is long gone.

I don't have a repro, so haven't deeply tested this, but @arthyn might. Please give it a go.


arthyn (approved):
lgtm as far as I can tell

PR #6751, opened 2023-08-11
hoon: fix gol type enforcement for trivial %cnts

### LINKED ISSUE #6750: hoon: =* tistar can be used to "escape" some nest checks
Observe:

```hoon
> =x |=  a=*
  ^-  (unit %whatever)
  =*  out  ['not' 'that']
  out

> (x 1)
dojo: failed to process input

> `*`(x 1)
[7.630.702 1.952.540.788]

> -:!>((x 1))
#t/u(%whatever)
```

This does not reproduce if the cast is just `^-  %whatever`. Code that calls such a construction happily compiles, since the _type_ is correct, but will proceed to process with a possibly-misshapen noun, leading to unexpected results, and violent crashes if you ever try pretty-printing to investigate.

Comments from @joemfb:

> `^-` becomes `^+`, and the nest check is enforced via setting `.gol`
`=*` becomes `%tune`, and `.gol` is not consulted
need to go thru `+mint:ut` and check any other cases were `+nice` isn't called on the result, make sure they're legit

joemfb:
Fixes #6750. Looks like the bug was introduced in 1892f28c61656123698a4385c719d2436ab1fa23.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -9238,21 +9238,21 @@
     ::
     ++  mint
       |=  gol=type
+      =-  ?>(?|(!vet (nest(sut gol) & p.-)) -)
       ^-  (pair type nock)
       =+  lug=(find %read hyp)
       ?:  ?=(%| -.lug)  ~>(%mean.'hoon' ?>(?=(~ rig) p.lug))
-      =-  ?>(?|(!vet (nest(sut gol) & p.-)) -)
       (ergo p.lug rig)
     ::
     ++  mull
       |=  [gol=type dox=type]
-      ^-  [type type]
+      =-  ?>(?|(!vet (nest(sut gol) & p.-)) -)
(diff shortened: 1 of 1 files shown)


joemfb:
This PR should probably target `next/hoon/138`. But it fixes a pretty egregious type hole with a very small change, so we may want to pull it forward into %zuse 412. Targeting `develop` for easy testing in lieu of a decision.

ashelkovnykov:
@joemfb Could you elaborate on what the error being fixed is? Do you have a simple example at hand?

pkova:
There's an example in the linked issue.

Note that a migration in [`/app/roller` of all places](https://github.com/urbit/urbit/actions/runs/5828373003/job/15805994909#step:6:1722) fails to compile with this change.

ashelkovnykov:
@pkova Thx, missed that somehow

belisarius222 (approved):

PR #6765, opened 2023-08-21
hoon: added gate to get list of keys to by

baudtack:
ley produces a list of keys instead of a set like key does


baudtack:
@joemfb when you have sec

ashelkovnykov:
This should probably have a jet too, right?

I'm also curious what the performance difference between `~(ley by x)` and `~(tap in ~(key by x))` is (jetted and unjetted).

PR #6767, opened 2023-08-22
eyre: handle old wires correctly in +on-gall-response

pkova:
I decided to run the upcoming 412k release on `~dinleb-rambep`. Something that immediately started occurring on every page refresh was seeing stuff like this in the dojo:
```
eyre: removing watch for non-existent channel 1692289927-d5b4c7 chat
```

It turns out the issue was that the wires being sent to Gall are formatted slightly differently after the 412k eyre changes.
```
:: old format
/channel/subscription/1692289927-d5b4c7/52/~dinleb-rambep/chat
:: new format 
/channel/subscription/1692289927-d5b4c7/52/~dinleb-rambep/chat/~dinleb-rambep
```
This resulted in old subscriptions never being cleaned up and getting some annoying eyre messages on every page refresh. This PR fixes the issue.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -2518,13 +2518,18 @@
       :_  state
       :_  ~
       ^-  move
-      =/  as=^ship
+      =/  [as=@p old=?]
         ?+  t.t.extra  ~|([%strange-wire extra] !!)
-          ~      our  ::  old-style wire
-          [@ ~]  (slav %p i.t.t.extra)
+          ~      [our &]
+          [@ ~]  [(slav %p i.t.t.extra) |]
         ==
+      =/  =wire  (subscription-wire channel-id request-id as ship app)
       %+  deal-as
-        (subscription-wire channel-id request-id as ship app)
+        ::NOTE  we previously used a wire format that had the local identity
+        ::      implicit, instead of explicit at the end of the wire. if we
+        ::      detect we used the old wire here, we must re-use that format
+        ::      (without id in the wire) for sending the %leave.
+        ?:(old (snip wire) wire)
       [as ship app %leave ~]
     ::  +emit-event: records an event occurred, possibly sending to client
     ::


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2518,13 +2518,13 @@
       :_  state
       :_  ~
       ^-  move
-      =/  as=^ship
-        ?+  t.t.extra  ~|([%strange-wire extra] !!)
-          ~      our  ::  old-style wire
-          [@ ~]  (slav %p i.t.t.extra)
+      =+  ?+  t.t.extra  ~|([%strange-wire extra] !!)
+          ~      [as=our is-old-wire=%.y]
+          [@ ~]  [as=(slav %p i.t.t.extra) is-old-wire=%.n]
```
Fang-, line 2523: I would just put the faces on with a cast, slightly cleaner.

```suggestion
      =/  [our=@p old=?]
        ?+  t.t.extra  ~|([%strange-wire extra] !!)
          ~      [our &]
          [@ ~]  [(slav %p i.t.t.extra) |]
```

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -2518,13 +2518,13 @@
...
         ==
+      =/  =wire  (subscription-wire channel-id request-id as ship app)
       %+  deal-as
-        (subscription-wire channel-id request-id as ship app)
+        ?:  is-old-wire  (snip wire)  wire
```
Fang-, line 2527: Worth leaving a comment.

```suggestion
        ::NOTE  we previously used a wire format that had the local identity
        ::      implicit, instead of explicit at the end of the wire. if we
        ::      detect we used the old wire here, we must re-use that format
        ::      (without id in the wire) for sending the %leave.
        ?:(old (snip wire) wire)
```

Fang- (commented):
I'd write this slightly differently, and important to leave a comment here explaining why we snip the wire given to us.

I don't think there's any other places where we strictly need to reuse the old wire format. New pokes and watches are fine using the new format.

(Also, this commit should start with `eyre: `.)

Good catch, and my bad for not testing the upgrade case thoroughly. Thank you.

pkova:
Addressed your comments and fixed the commit message.

Fang- (approved):
Perfect, thank you! (:

PR #6770, opened 2023-08-22
ames: persistently cache `++jim` and remove `++dedup-message`

matthew-levan:
Depends on urbit/vere#496.


belisarius222:
This code LGTM, so once the runtime supports persistent caching, we should merge this.  Note that we'll need this to be part of a Kelvin change since it won't work right on older Veres -- it'll run, but without persistent caching, Ames's deduplication routine is an important memory saver.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -1742,15 +1830,33 @@
...
+        ^-  (qeu queued-event)
+        %-  ~(rep in events)
+        |=  [e=queued-event-17 q=(qeu queued-event)]
+        %-  ~(put to q)  ^-  queued-event
+        ?.  ?=(%call -.e)  e
+        =/  task=task-17  ((harden task-17) wrapped-task.e)
+        %=  e
+          wrapped-task
+          ?.  ?=([%deep %nack *] task)  task
+          =/  msg  (cue message-blob.task)
```
joemfb, line 1853: This should be in a `~+` (with a minimal subject).

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -5293,6 +5368,34 @@
...
+    %-  ~(gas to *(qeu message))
+    %+  turn
+      ~(tap to unsent-messages.m)
+    |=  b=message-blob
+    ^-  message
+     =/  hed
+       ?:  =(1 (end 0 bone))          %plea
+       ?:  =(0 (end 0 (rsh 0 bone)))  %boon
+       %naxplanation
+    ;;(message [hed (cue b)])
```
joemfb, line 5397: And this.

pkg/arvo/sys/lull.hoon
```diff
@@ -904,6 +904,12 @@
   ::    payload: semantic message contents
   ::
   +$  plea  [vane=@tas =path payload=*]
+  ::
+  +$  message
+    $%  [%plea plea]
+        [%boon payload=*]
+        [%naxplanation =message-num =error]
```
joemfb, line 911: Tempted to say this should be `%naxp` to keep it a direct atom.

pkova: `%naxplanation` is the same as in `$ack`, changing that also will be annoying.

pkg/arvo/sys/vane/ames.hoon
pkova: Done

pkg/arvo/sys/vane/ames.hoon
pkova: Done

joemfb: As above. Also, this cue's twice.

pkg/arvo/app/dbug.hoon
```diff
@@ -773,7 +773,7 @@
           'next'^(numb next)
         ::
           :-  'unsent-messages'  ::  as byte sizes
-          (set-array unsent-messages (cork (cury met 3) numb))
+          (set-array unsent-messages (cork jam (cork (cury met 3) numb)))
```
joemfb, line 776: I'm worried that this could be unacceptably slow. We could move `+jim` so that it can be used here, otherwise it's probably best to remove this field or stub it out.

joemfb (commented):
One nit wrt the debug dashboard, otherwise this will do

matthew-levan:
I tested this yesterday with @pkova. We discovered that we can indeed fill the persistent cache with Ames messages using a thread that looks similar to this:
```
/-  spider
/+  strandio
=,  strand=strand:spider
^-  thread:spider
|=  arg=vase
=/  m  (strand ,vase)
^-  form:m
=+  !<([~ arg=@] arg)
=/  ships=(list @p)  (gulf 0 (dec arg))
|-
?~  ships  (pure:m !>('done'))
~&  [%poking i.ships]
=/  =card:agent:gall  [%pass /poke %agent [i.ships %lolo] %poke *cage]
;<  ~  bind:m  (send-raw-card:strandio card)
$(ships t.ships)
```

I'm working on producing benchmarks for Ames message send performance now, using Wireshark. Is using app distribution the best way to go about that? I could setup two fake ships on my local machine, the publisher with the persistent caching code and the other without. Then, I'd publish an app from the publisher ship and `|install` it from the other and time how long it takes to install. Then, I could do the same without any persistent caching and compare the difference.

belisarius222:
You could set up a distribution moon for a private desk, then measure how long it takes for, say, five other ships to complete downloading that desk.  This should be much faster with persistent memoization than with explicit deduplication, which gets slow when there are a lot of outstanding unacked messages in ames.

Making the desk private prevents remote scry from being used -- remote scry doesn't have this problem.  Make sure to add some large text files to the desk, each of which must have unique contents so that Clay can't deduplicate them in its hash-addressed storage.  Maybe copy the /sys folder to /sus and add a comment line to the top of each file.

belisarius222:
Actually /sys might not have enough files in it to make Ames slow.  It could be easier to boot up ships that don't have Tlon groups, then publish an unmodified groups desk privately and time that.

joemfb (approved):
this will do

PR #6773, opened 2023-08-23
arvo: persistently cache scry type checks

matthew-levan:
Depends on urbit/vere#496.

### DIFF
--- a/pkg/arvo/sys/arvo.hoon
+++ b/pkg/arvo/sys/arvo.hoon
@@ -352,7 +352,9 @@
   ?~  dat=(rof lyc pov u.mon)  ~
   ?~  u.dat  [~ ~]
   =*  vax  q.u.u.dat
-  ?.  ?&  ?=(^ ref)
+  ?.  =>  [ref=ref vax=p=p.vax hoon-version=hoon-version wa=wa worm=worm]
+      ~>  %memo./arvo/look                  ::  with memoization
+      ?&  ?=(^ ref)
           =(hoon-version -.ref)
           -:(~(nets wa *worm) +.ref p.vax)
       ==


matthew-levan:
I've tested this and confirmed that the persistent cache size grows as the ship performs various scries. What's a good way to benchmark the performance difference, though?

belisarius222:
Have you tried adding a %bout hint? You could emit one only if the scry path matches certain conditions, to filter for, say, Clay %cc scrys.

matthew-levan:
Cool, thanks @belisarius222. I just added a `%bout` hint, like so:
```hoon
++  look
  ~/  %look
  |=  [rof=roof lyc=gang pov=path]
  ^-  root
  ~/  %in
  |=  [ref=* raw=*]
  ?~  pax=((soft path) raw)  ~
  ?~  mon=(de-omen u.pax)  ~
  ?~  dat=(rof lyc pov u.mon)  ~
  ?~  u.dat  [~ ~]
  =*  vax  q.u.u.dat
  ?:  (%cc vis.u.mon)  ~>  %bout
    ?.  =>  [ref=ref vax=p=p.vax hoon-version=hoon-version wa=wa worm=worm]
        ~>  %memo./arvo/look                  ::  with memoization
        ?&  ?=(^ ref)
            =(hoon-version -.ref)
            -:(~(nets wa *worm) +.ref p.vax)
        ==
      ~>(%slog.[0 leaf+"arvo: scry-lost"] ~)
    [~ ~ q.vax]
  ?.  =>  [ref=ref vax=p=p.vax hoon-version=hoon-version wa=wa worm=worm]
      ~>  %memo./arvo/look                  ::  with memoization
      ?&  ?=(^ ref)
          =(hoon-version -.ref)
          -:(~(nets wa *worm) +.ref p.vax)
      ==
    ~>(%slog.[0 leaf+"arvo: scry-lost"] ~)
  [~ ~ q.vax]
```

What's the better way to do this? This Hoon is definitely nasty and I apologize in advance, but it got the job done for me nonetheless.

The takeaway: at least 2x gain, but more like 5x (at least on this particular scry; see below).

I tested using `=a .^(tube:clay %cc %/txt/mime)` and running it over and over in the `dojo`. With caching, the first one takes ~25us and subsequent calls take ~5us. Without caching, ~25us every time.

Now the question is: when compounded across the various scries that happen in the system almost all the time, what does the overall performance boost look like (on average)?

pkg/arvo/sys/arvo.hoon
```diff
@@ -352,7 +352,9 @@
   ?~  dat=(rof lyc pov u.mon)  ~
   ?~  u.dat  [~ ~]
   =*  vax  q.u.u.dat
-  ?.  ?&  ?=(^ ref)
+  ?.  =>  [ref=ref vax=p=p.vax hoon-version=hoon-version wa=wa worm=worm]
+      ~>  %memo./arvo/look                  ::  with memoization
+      ?&  ?=(^ ref)
           =(hoon-version -.ref)
           -:(~(nets wa *worm) +.ref p.vax)
```
joemfb, line 359: this is a small thing, but we can minimize the subject further here by removing `worm` and changing this to `-:(nets:wa +.ref p.vax)`.

PR #6774, opened 2023-08-23
arvo: add provenance when casting forward move

pkova:
`|verb` is broken on 412k prerelease because we were casting to old-style `[ship ship]` here instead of the correct `[ship ship path]` with provenance included.

Thanks goes to @midden-fabler for the bug report and @joemfb for the fix.


joemfb (approved):
LGTM

Issue #6776, opened 2023-08-24
gall: remote scries targeting tombstones hang forever


del-distributed:
TLDR: Remote scries targeting tombstones hang forever. The issue is addressed in #6743, but the problem persists for me after recompiling clay with the change. I propose #6778 as a partial fix for now.

**To Reproduce**
1. Recompile clay with #6743
2. Register a remote scry path
3. Tombstone it
4. Request it from a secondary ship

**Additional context**

Gall handles `%x` reads as follows:
```
?~  yok=(~(get by yokes.state) q.bem)  ~
    ?:  ?=(%nuke -.u.yok)  ~
    =/  ski  (~(get by sky.u.yok) path)
    ?~  ski  ~
    =/  res=(unit (each page @uvI))
      ?+    -.r.bem  ~
          %ud  (bind (get:on-path fan.u.ski p.r.bem) tail)
          %da
        %-  head
        %^    (dip:on-path (unit (each page @uvI)))
            fan.u.ski
          ~
        |=  [res=(unit (each page @uvI)) @ud =@da val=(each page @uvI)]
        ^-  [new=(unit [@da _val]) stop=? res=(unit _val)]
        :-  `[da val]
        ?:((lte da p.r.bem) |/`val &/res)
      ==
    ?.  ?=([~ %& *] res)  ~        :: change ~ to [~ ~] to fix
    ``p.u.res(q !>(q.p.u.res))
```
It's still unclear to me why the issue arises at all, but changing `~` to `[~ ~]` on the final branch resolves it.

**Post-fix:**

Non-tombstoned response: `dat=[p=path q=[~ [p=%atom q=123.456.789]]]`

Tombstoned response: `dat=[p=path q=~]`

Note here that `q`, which is a `(unit (cask))`, now returns `~` in the tombstone case. Perhaps that's okay. I conceive two alternatives:

a) `dat=[p=path q=[~ ~]`

Conforms with unitized structure of the non-tombstoned case. I'm unsure how to rewrite the handler to produce this, due compiler-enforced typing.

b) `dat=[p=path q=[~ tombstone-hash]`

Clay remembers hashes of tombstoned values. If we wanted to serve them, I think the response mold would have to be reconsidered, since the current `(unit (cask))` would require a reserved term like `%$` to express tombstones, i.e.: `[~ [%$ 0vn.ebj2e]]`

**Local crash**

Locally executed %gx reads crash on tombstones, and the proposed fix does not solve this:

[...]

PR #6778, opened 2023-08-25
gall: fix remote scries hanging on tombstones

del-distributed:
Explained in #6776

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -2280,7 +2280,7 @@
         :-  `[da val]
         ?:((lte da p.r.bem) |/`val &/res)
       ==
-    ?.  ?=([~ %& *] res)  ~
+    ?.  ?=([~ %& *] res)  [~ ~]
     ``p.u.res(q !>(q.p.u.res))
   ::
   ?:  ?&  =(%t care)


belisarius222:
Unfortunately this approach to trying to fetch tombstoned files breaks referential transparency.  The way the remote scry protocol works, a null response means a permanent null.  Tombstones are temporary nulls: any tombstoned file could be reinflated later.

If a ship receives a null response from a publisher ship for a tombstoned file, it would need to interpret that as a permanent null, leading to inconsistencies such as the receiving ship calculating the `%z` folder hash incorrectly by assuming the tombstoned file doesn't exist at that revision.

In general, when a ship handles a scry request that it can't or won't answer, it "blocks" rather than returning a referentially transparent response.  In many parts of the system, a scry result is represented as a `(unit (unit cage))`.  A value of this type can be one of three shapes:
- `[~ ~ value]`, a positive response (e.g. file contents in response to a `%x` scry)
- `[~ ~]`, a permanent negative response (it is known that no such file existed)
- `~`, "blocked" -- we refuse to answer the question

A ship can always refuse to answer a scry, and block instead.  A block is not a referentially transparent response.  A ship can change its mind later and answer the question if asked again, yielding either a positive or negative permanent response -- with the constraint that for any given path, if the ship does give an answer rather than block, it will always give the same answer.  For a nonexistent file, the ship could alternate between blocking and the negative response (perhaps if permissions are being toggled off and on).  For an existent file, the ship could alternate between blocking and the positive response.  But a ship must never switch between negative and positive responses to a given path.

Emitting a negative response in the current scry protocol would mean alternating between positive and negative responses as the file is tombstoned and reinflated, thus violating referential transparency.  The way to avoid this is to block, which for remote scry means not emitting any response at all.

I will note, as you have undoubtedly experienced, @del-distributed, that this is annoying.  As a sender, getting no response at all is indistinguishable from the other ship being offline or otherwise unresponsive, and it's not easy to determine what went wrong that's preventing a desk from downloading.

I think the remote scry protocol should be adjusted to address this issue.  The Ad Fontes proposals include a design for a couple of packet types that indicate different reasons for scry blocking, for example.  Fixing this in the network protocol could potentially be a difficult project, though -- because blocking is not referentially transparent, it becomes harder to protect against things like replay attacks and stale caches, which our current referentially transparent networking defines out of existence.

For now, I recommend workarounds in Clay and Gall.  There are scrys that can query for existence of a path, and in Clay, I believe there is a way to ask whether a file is tombstoned (if not, there needs to be).  If you can run an agent on a publishing ship, a subscriber ship could poke that agent first to get it to ask the vane and poke the subscriber back with the answer.  The subscriber would only try the scry if the value isn't tombstoned.

There are race conditions with this, where the publisher could tombstone the file just after the subscriber asks whether it's tombstoned and just before the scry request packets reach the publisher, causing a failure.

In general, the current remote scry protocol has a number of limitations: it should only be used for ships we're already in contact with, to ask for paths that we know won't block.  Addressing these limitations is high priority for core.

PR #6782, opened 2023-09-02
x/jub: add jub:by

xiphiness:
jub:by is a variant of jab:by, which applies a gate to a value at a certain key. It crashes if it's not there, and also does not allow it to be deleted. This variant allows a create, update, or delete operation with one gate call and one tree traversal. The gate passed to jub takes a unit and returns a unit. The unit it is passed is ~ if the kv pair does not exist, and `val if it does. The unit returned causes a deletion if it is ~, and inserts otherwise.

Would replace https://github.com/urbit/urbit/pull/6621.

Do we want to add the jib mentioned there? See my last comment if so if we want to also add the same deletion semantics.

https://github.com/urbit/vere/pull/510 is the corresponding vere PR.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1564,6 +1564,29 @@
     ::
     a(r $(a r.a))
   ::
+  ++  jub
+    ~/  %jub
+    |*  [b=* fun=$-((unit _?>(?=(^ a) q.n.a)) (unit _?>(?=(^ a) q.n.a)))]
+    =>  .(b ^+(?>(?=(^ a) p.n.a) b))
+    |-  ^-  (tree _?>(?=(^ a) n.a))
+    ?~  a
+      ?~(c=(fun ~) ~ [[b u.c] ~ ~])
+    ?:  =(b p.n.a)
+      ?~  c=(fun `q.n.a)  (del(a ^+($ a)) b)
+      ?:  =(q.n.a u.c)  a
+      a(n [b u.c])
+    ?:  (gor b p.n.a)
+      =+  d=$(a l.a)
+      ?~  d  (put(a r.a) n.a)
+      ?:  (mor p.n.a p.n.d)
+        a(l d)
+      d(r a(l r.d))
+    =+  d=$(a r.a)
+    ?~  d  (put(a l.a) n.a)
+    ?:  (mor p.n.a p.n.d)
+      a(r d)
+    d(l a(r l.d))
+  ::
   ++  mar                                               ::  add with validation
     |*  [b=* c=(unit *)]
     ?~  c


pkg/arvo/sys/hoon.hoon
```diff
@@ -1564,6 +1564,29 @@
...
+      ?:  =(q.n.a u.c)  a
+      a(n [b u.c])
+    ?:  (gor b p.n.a)
+      =+  d=$(a l.a)
+      ?~  d  (put(a r.a) n.a)
+      ?:  (mor p.n.a p.n.d)
+        a(l d)
+      d(r a(l r.d))
+    =+  d=$(a r.a)
+    ?~  d  (put(a l.a) n.a)
```
xiphiness, line 1585: Is there a cleverer way to handle the case that the recursive call returns an empty map than to use put here? Similar case on line 1579

joemfb:
@xiphiness here's the "jub" variant i proposed in #6672:

```
++  job
    |*  [b=* c=$-((unit) *)]
    =>  %=  .
          b  `_?>(?=(^ a) p.n.a)`b
          c  `_?>(?=(^ a) |~((unit _q.n.a) q.n.a))`c
        ==
    |-  ^+  [*? a]
    ?~  a
      [& [[b (c ~)] ~ ~]]
    ?:  =(b p.n.a)
      [| a(q.n (c `q.n.a))]
    ?:  (gor b p.n.a)
      =^(d l.a $(a l.a) [d a])
    =^(d r.a $(a r.a) [d a])
```

The idea of returning a flag kind of breaks down in the presence of deletion, but here's a not-quite-right variant:

```
++  jyb
    |*  [b=* c=$-((unit) (unit))]
    =>  %=  .
          b  `_?>(?=(^ a) p.n.a)`b
          c  `_?>(?=(^ a) |~((unit _q.n.a) *(unit _q.n.a)))`c
        ==
    |-  ^-  (pair (unit ?) _a)
    ?~  a
      ?~  d=(c ~)
        [~ ~]
      [`& [[b u.d] ~ ~]]
    ?:  =(b p.n.a)
      ?~  d=(c `q.n.a)
        [~ (~(del by a) b)]
      [`| a(q.n u.d)]
    ?:  (gor b p.n.a)
      =^(e l.a $(a l.a) [e a])
    =^(e r.a $(a r.a) [e a])
```

xiphiness:
could also `?(%0 %1 %2 %3)`
%0 - deleted
%1 - insertion
%2 - present - no change
%3 - present - change

slightly broken - you mean that you aren't doing the `mor` stuff that `put:by` does/my version does?

xiphiness:
```hoon
  ++  jub
    ~/  %jub
    |*  [b=* fun=$-((unit _?>(?=(^ a) q.n.a)) (unit _?>(?=(^ a) q.n.a)))]
    =>  .(b ^+(?>(?=(^ a) p.n.a) b))
    ^-  [p=?(%sam %ins %del %upd) q=(tree _?>(?=(^ a) n.a))]
    =-  ?~(- %sam^a u)
    |^  ^-  (unit [p=?(%ins %del %upd) q=(tree _?>(?=(^ a) n.a)))
    =|  =tez
    |-  ^-  ^$
    ?~  a
      ?~(c=(fun ~) ~ `ins+unwynd(a [[b u.c] ~ ~]))
    ?:  =(b p.n.a)
      ?~  c=(fun `q.n.a)  `del+unwynd(a (del(a a) b))
      ?:  =(q.n.a u.c)  ~
      `upd+unwind(a a(n [b u.c]))
    ?:  (gor b p.n.a)
      $(a l.a, [%l tez n.a r.a])
    $(a r.a, tez [%r tez n.a l.a])
    ::
    +$  tez
      $@  %top
      $%  [%l ctx=tez par=_?>(?=(^ a) n.a) r=`_a`~]
          [%r ctx=tez par=_?>(?=(^ a) n.a) l=`_a`~]
      ==
    ::
    ::  for use when ordering won't change
    ++  unwind
      |-  ^+  a
      ?@  tez  a
      ?-  -.tez
        %l  $(a [par.tez a r.tez], tez ctx.tez)
        %r  $(a [par.tez l.tez a], tez ctx.tez)
      ==
    ::
    ::  for use when ordering might change
    ::  note: tez is essentially a and a essentially d
    ::  in the normal put/jub
    ++  unwynd
      |-  ^-  (tree a)
      ?@  tez  a
      ?-  -.tez
            %l
        ?~  a  $(a (put(a r.tez) par.tez), tez ctx.tez)
        ?:  (mor par.tez n.a)
          unwind(a [par.tez a r.tez], tez ctx.tez)
        $(a a(r [par.tez r.a r.tez]), tez ctx.tez)
      ::
            %r  $(a [par.tez a r.tez])
        ?~  a  $(a (put(a l.tez) par.tez), tez ctx.tez)
        ?:  (mor par.tez n.a)
          unwind(a [par.tez l.tez a], tez ctx.tez)
        $(a a(l [par.tez l.tez l.a]))
      ==
    --
```

it's basically my jub code written w a zipper and adopting this return flag idea


haven't tested or even tried to make it compile, but having done tree traversal structures that accumulates/returns data like you are with =^, i've often wondered if doing a zipper sort of like this might be more performant. I guess, after thinking about it, it might be faster unjetted, but jetted the normal way is probably faster as you have a native C call stack. alas, i had already written the code and most of this comment by the time i thought of it. this could apply to `put:by` as well. but, I guess in the nonzipper pattern, once you hit an unchanged node in the traversal upwards, every mor check on the way up would have reference equality.

ashelkovnykov:
@xiphiness Your latest comment looks all good (though I'm not the best at Hoon treap logic)

xiphiness:
yes

joemfb:
@pkova this was inadvertently closed by the deletion of `next/hoon/138` in #6615. I've restored the branch for now.

PR #6783, opened 2023-09-04
lull, ames: add %dear task to receive lane from unix

pkova:
Adds a task `%dear` that unix can use to notify arvo about a new lane. Made this for my [mDNS PR](https://github.com/urbit/vere/pull/511) but could be useful in other contexts as well.


joemfb (approved):
this will do

PR #6785, opened 2023-09-06
gall: add scry endpoint to expose agent state

midden-fabler:
UIP: [Expose agent state to enable backup and restore.](https://github.com/urbit/UIPs/blob/main/UIPS/UIP-0110.md)

Adds a `%gv` vane scry that returns a version # and an `$egg`.

Example:
`=a .^([@ud *] %gv /=pals=/$)`

TODO:
- [ ] Add `$egg` to `%lull`


pkova (approved):

joemfb:
The scry endpoint produces `$egg`, tagged with the version number from `$spore`. Ideally there would be a `$%` matching that structure in %lull, to make it easier to upgrade call sites when `$egg` inevitably changes.

PR #6786, opened 2023-09-08
colhax

hanfel-dovned:


belisarius222:
Thank you @hanfel-dovned, but I don't think we'll want to merge this.  There is a point of diminishing returns for fixed-arity runes beyond a certain number of children.  The `:^` four-element tuple rune is already used pretty sparingly, so this rune for five-element tuples would be used even less.

I'm going to close this for now.  If anyone has a strong objection, feel free to comment and I'll consider reopening.

ashelkovnykov:
Agree w/ @belisarius222 (though I just searched for usages of `:^` and was surprised how many hits I got: 90)

PR #6790, opened 2023-09-14
gall: security primitives for encrypted scry

liam-fitzgerald:
This commit adds `%tend` `%germ` and `%snip` to the notes that gall can pass. `%tend` is analogous to `%grow`, except with a security group defined by .coop.

 ### The coop system
A `$coop` is a path, which defines a security context for the portion of the namespace that it prefixes. Each `$coop` receives a symmetric key, which is used to encrypt requests and responses for any key-value pair belonging to a coop.

 ### Network overview
This design requires a single handshake over ames to inform clients what key is to be used. However, this handshake can be made less frequent by including all paths underneath the `%coop` in the response, such that if the user is requesting sibling paths under the same `%coop`, only one handshake is required.

 ### Naming
I am utterly detached to all new names introduced, just trying to get something down

 ### API Design
The most contentious part of this proposal will likely be the split between `%grow` and `%tend`. I assert (rather weakly mind you) that this is more ergonomic for the end user, although there's a strong argument to be made that `%grow` should just take a `(unit coop)`. If this were the case, however, it would muddy the semantics. If the value is encrypted, then the ship,desk,case will be in the coop, else it will be specified in the path.  Worth noting that specifying the `%coop` and the rest of the path seperately seems like it could be unintuitive because the path that it will be bound to is actually `(welp coop path)`

The lifecycles for coops seem straightforward, although worth revisiting the invariants it maintains, and how it handles those invariants. A list of such:
- No nesting (obviously good)
- Crashing on binding publically into a private coop (crashing is bad, do we want to deliver a notification? (See footnote 1))
- Crashing on binding into a coop that doesn't exist (same notes as above)

 ### Key generation
Current implementation is obviously stupid, how should i do it?

 ### Footnotes
 1. Why are the remote scry datastructures notes and not gifts? Forgive me being out of the loop, but we don't actually use the wire for anything anywhere, and remote scry is giving gift anyway.
 2. It's so good to be back


liam-fitzgerald:
Pushed request changes @joemfb inc. merging your two party remote scry work

jalehman:
@liam-fitzgerald the design of this should probably be captured in a UIP, which would make reviewing the conceptual part of it efficient. Thoughts @belisarius222 ?

liam-fitzgerald:
@jalehman https://github.com/urbit/UIPs/pull/31

pkova:
@liam-fitzgerald you're gonna have to update up your migrations here because of the hotfix we did https://github.com/urbit/urbit/pull/6838. You can consolidate your migration stuff to ames state %19 or just make a new ames state %20.

liam-fitzgerald:
Whichever is easier for you

liam-fitzgerald:
You're always long housing me

joemfb (changes requested):
@liam-fitzgerald I've noted a few issues that need to be cleaned up. You also have some outstanding TODOs in this PR.

I'm having a little trouble internalizing the %gall coop/hutch/farm details, it'd be helpful to go over that synchronously. Full disclosure: my agenda there is to see if it'd be possible to handle key distribution through the namespace as well, possibly via the 2-party encrypted remote scry (modulo "pine" and/or a publication schedule).

liam-fitzgerald (commented):
address todos, access control logic.

PR #6804, opened 2023-09-23
dojo: reject pasted bad code

marcusmiguel:
pasting `(add  2 2)`  (double whitespace after 'add')

before:
```
~sampel-palnet:dojo> (add  2 2)
```
now:
```
~sampel-palnet:dojo> (add  
```
%dojo will reject the code from the errored character onwards, and the errored character will be displayed. This is different from the behavior when typing single characters one at a time (the errored character doesn't go through), but I don't see a better solution.

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -1134,8 +1134,14 @@
   ++  he-errd                                           ::  reject update
     |=  [rev=(unit sole-edit) err=@u]  ^+  +>
     =+  red=(fall rev [%nop ~])       ::  required for error location sync
+    =?  red  =(%mor -:red)
+      =/  rel  ;;((list sole-edit) +:red)
+      =/  idx  (find `(list sole-edit)`[%del `@ud`err]~ rel)
+      ?.  ?=(^ idx)
+        red
+      [%mor (oust [+:idx (lent rel)] rel)]
     =^  lic  say  (~(transmit sole say) red)
-    (he-diff %mor [%det lic] [%err err] ~)
+    (he-diff %mor [%det lic] [%err (lent buf.say)] ~)
   ::
   ++  he-pone                                           ::  clear prompt
     ^+  .
@@ -1348,8 +1354,14 @@
     ::  ~&  [%his-clock ler.cal]
     ::  ~&  [%our-clock ven.say]
     =^  dat  say  (~(transceive sole say) cal)
-    ?.  ?&  ?=(%del -.dat)
-            =(+(p.dat) (lent buf.say))
+    =/  len  (lent buf.say)
+    ?.  ?|  ?&  ?=(%del -.dat)
+                =(+(p.dat) len)
+            ==
+            ?&  ?=([%mor ^] dat)
+                =(%del +<-:dat)
+                =(+(;;(@ud +<+:dat)) len)
+            ==
         ==
       +>.$
     =+  foy=(he-dope (tufa buf.say))


belisarius222:
I'd like to note that I think of two other potentially nice behaviors for dojo in case the user pastes in something with a syntax error:
1) roll back the entire paste, rejecting the whole thing, and ring the terminal bell
2) perform the paste, then move the cursor to the first erroring character

I don't have a strong opinion on which one would be best.  The behavior described in this PR seems like a Pareto improvement over what it does now, so in my opinion we should merge it if the code looks good.

marcusmiguel:
My bad, i should have expanded more on this, it's not only when pasting that this PR will have effects; its every time that you send more than one char to the dojo, which is usually when pasting but also i need this to make the debounce work on https://github.com/urbit/webterm/pull/5.

> 1. roll back the entire paste, rejecting the whole thing, and ring the terminal bell

If we do this, the aforementioned PR will not work well, you will be deleting the whole line almost every time that there is an error.

>  2. perform the paste, then move the cursor to the first erroring character

If we do this, it will also make the experience worse, you will have the cursor moving back everytime that the backend gives you a correction. 

If we are going to merge the %webterm PR, then the solution proposed on this PR is the only one that makes the experience the closest to typing in the normal terminal. @belisarius222

belisarius222:
Thanks @marcusmiguel, that all makes sense.  This PR does seem better than those options, then.

I wonder if we should eventually have two different kinds of "insert multiple characters" commands: one atomic, and one not.  The atomic one could be used for pastes, where if the pasted text fails syntax matching, the whole thing gets rolled back.  The non-atomic version would be used for debouncing by remote frontends, and it would work like this PR, where it inserts all the characters before the failure.

pkg/arvo/app/dojo.hoon
```diff
@@ -1134,8 +1134,14 @@
   ++  he-errd                                           ::  reject update
     |=  [rev=(unit sole-edit) err=@u]  ^+  +>
     =+  red=(fall rev [%nop ~])       ::  required for error location sync
+    =?  red  =(%mor -:red)
+      =/  rel  ;;((list sole-edit) +:red)
+      =/  idx  (find `(list sole-edit)`[%del `@ud`err]~ rel)
+      ?.  ?=(^ idx)
+        red
+      [%mor (oust [+:idx (lent rel)] rel)]
```
Fang-, line 1142:
```suggestion
    =?  red  ?=(%mor -.red)
      =/  rel  p.red
      =/  idx  (find [%del err]~ rel)
      ?~  idx  red
      [%mor (scag u.idx rel)]
```

Couple changes for better style:
- If you `?=` to type-check in the conditional, instead of plain equality check, you get typed access to `p.red`, since the type system now knows `red` is the `%mor` case.
- Prefer `.` instead of `:` for simple sub-noun access.
- I don't think you need all these explicit typecasts in the `+find` call.
- `?~  etc` in place of `?.  ?=(^ etc)` is more idiomatic.
- `+scag` takes the first n items of a list, more idiomatic than deleting anything path the nth item.

pkg/arvo/app/dojo.hoon
```diff
@@ -1134,8 +1134,14 @@
...
     =^  lic  say  (~(transmit sole say) red)
-    (he-diff %mor [%det lic] [%err err] ~)
+    (he-diff %mor [%det lic] [%err (lent buf.say)] ~)
```
Fang-, line 1144: Since we cut out all edits/additions beyond the `err`th character in the buffer, isn't this equivalent to the original code?

The original code also has the flexibility to jump to error locations in the middle of the buffer, which may occur if you typed there and then hit return. (Unsure whether that actually hits this same codepath..)

pkg/arvo/app/dojo.hoon
```diff
@@ -1348,8 +1354,14 @@
...
-    ?.  ?&  ?=(%del -.dat)
-            =(+(p.dat) (lent buf.say))
+    =/  len  (lent buf.say)
+    ?.  ?|  ?&  ?=(%del -.dat)
+                =(+(p.dat) len)
+            ==
+            ?&  ?=([%mor ^] dat)
+                =(%del +<-:dat)
+                =(+(;;(@ud +<+:dat)) len)
+            ==
```
Fang-, line 1364:
```suggestion
            ?&  ?=([%mor [%del *] *] dat)
                =(+(p.i.p.dat) len)
            ==
```

You can type-check for the exact thing you want in one go, and then access the `%del` value using faces instead of lark.

Fang- (commented):
I'll disregard the motivations for this stemming from urbit/webterm#5, since per https://github.com/urbit/webterm/issues/4#issuecomment-1734312178 I don't think it's exactly the right thing to do in the general case. However, even taken on its own I think this PR improves dojo ux, so the direction of the change here is good.

That said, some implementation-level comments below.

Somewhat unfortunate that the drum protocol treats pastes as multiple keystrokes instead of a single big insertion. Wondering if there's any cases where the `%mor` heuristic here gives false positives, and if there are any, whether this behavior is undesirable in those cases...

marcusmiguel:
> isn't this equivalent to the original code?

The difference is that in this PR, the first character that causes the error will remain (when receiving a %mor), whereas in the original code, it was deleted.  Had to do that because:  `the drum protocol treats pastes as multiple keystrokes instead of a single big insertion` . The erroring character had to remain so that when the rest of the paste comes, it gets blocked.

> The original code also has the flexibility to jump to error locations in the middle of the buffer

Nice catch, adjustments to this PR would be necessary to preserve this functionality.

@Fang-

marcusmiguel:
> isn't this equivalent to the original code?

The difference is that in this PR, the first character that causes the error will remain (when receiving a %mor), whereas in the original code, it was deleted. So it might be err + 1. Had to do that because:  `the drum protocol treats pastes as multiple keystrokes instead of a single big insertion` . The erroring character had to remain so that when the rest of the paste comes, it gets blocked.

> The original code also has the flexibility to jump to error locations in the middle of the buffer

Nice catch, adjustments to this PR would be necessary to preserve this functionality.

@Fang-

PR #6808, opened 2023-09-28
clay: return all takos in /cs/bloc scry

pkova:
This has the exact same problem as #6807, where the result of the `%bloc` scry excludes takos that are not reachable from the latest aeon.

The `%bloc` scry is only used by `/lib/pill` to pre-populate the clay blob store. The current urbit-v2.12 bootstrap pill is already built with this fix.


joemfb (approved):
this will do

PR #6814, opened 2023-10-03
eyre: prevent login redirect loops

Fang-:
Instead of auto-redirecting the login page if we're already logged in, we simply present it as normal. If we're logged in as the local identity, we present a "continue" button in place of the +code input field. If we're authenticated through eauth, or as a guest, we present a smaller "proceed as" link underneath the login form.

This way, when apps redirect to the login page because the visitor isn't _sufficiently_ authenticated, eyre doesn't just throw them right back into the app with still insufficient creds.

That situation got more common as of the 412 release, where Eyre changes made it possible for the `authenticated` flag of requests (which should now be considered deprecated) to be set to false, while Eyre itself would recognize a user as logged in. Apps that had not been updated could redirect eauth-ed users to the login page, which would then redirect right back into the app, ad infinitum.  
Since this isn't the first time we're dealing with login-redirect loops, and because there are now multiple ways to be authenticated, I figured the best call here was to just stop auto-redirecting upon loading the login page.

<details><summary>screenshots</summary>

Already logged in as the local identity:
image

Not logged in at all (but still assigned a guest identity):
image

Logged in through eauth:
image

These show up regardless of what login tab is selected:
image
</details>

(I experimented briefly with letting apps add a little `?message` query param, to be displayed on the login page, but this didn't seem quite as nice as I first thought. If we're gonna tell the user anything, Eyre should be in control of what's being told.)

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -366,6 +366,8 @@
   }
   button[type=submit] {
     margin-top: 1rem;
+  }
+  button[type=submit], a.button {
     font-size: 1rem;
     padding: 0.5rem 1rem;
     border-radius: 0.5rem;
@@ -373,6 +375,7 @@
     color: var(--white);
     border: none;
     font-weight: 600;
+    text-decoration: none;
   }
   input:invalid ~ button[type=submit] {
     border-color: currentColor;
@@ -380,7 +383,7 @@
     color: var(--gray-400);
     pointer-events: none;
   }
-  span.guest {
+  span.guest, span.guest a {
     color: var(--gray-400);
   }
   span.failed {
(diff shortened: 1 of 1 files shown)


tinnus-napbus (approved):
nice seems like a sensible approach

vcavallo:
I think this may be breaking %mask.

mask internally posts to /~/login and passes the cookie it gets back to the browser. Here is the final bit of %mask's handling of the browser's POST request (which includes a signed challenge from metamask. point being, if you got here, you are allowed to have the cookie):

image

But in getting the response from the browser, the cookies it gets in response are the same (old) cookie it had to begin with:

image

And the session is _not_ logged in.

....I'm drawn to this comment in eyre.hoon:

image

Fang-:
We settled @vcavallo's issue out of band. tldr: it's completely unrelated to this PR. His agent is including a `Cookie` header in the response, but Eyre appends its own `Cookie` header that sets a cookie with the same key, but an "older" value. This is not misbehavior on Eyre's side, but a side-effect of %mask implementation details.

PR #6819, opened 2023-10-06
Revert "hoon: add pin and awl arms to test and trim subaxes"

joemfb:
These new APIs are unused, and I believe there are no plans to use them. They can always be added back if we need them.


eamsden (approved):

joemfb:
Linking #6592 and urbit/vere#397 for posterity.

PR #6820, opened 2023-10-08
fix: change `en-json` to `en:json` and `as-octt` to `as-octs`

matthiasschaub:
### Description

Update occurrences of `en-json` (deprecated) to `en:json` and `as-octt` to `as-octs` in developer desk `yard`.


sigilante:
See https://github.com/urbit/yard/tree/main/desk/lib

PR #6822, opened 2023-10-09
gen: add |gall-lave to print stale bitt.yoke

yosoyubik:
- [ ] add %gall task do the the actual cleaning

On ~norsyr-torryn this shows `"#1.554 stale incoming subscriptions"`, but other publishers will probably have a lot more.

### DIFF
--- a/pkg/arvo/gen/hood/gall-lave.hoon
+++ b/pkg/arvo/gen/hood/gall-lave.hoon
@@ -1,0 +1,51 @@
+::  Deletes all stale incoming gall (re) subscriptions
+::
+::    It runs in dry mode by default, printing the number of stale subscriptions
+::    To actually delete the subscriptions, run with |gall-lave, =dry |
+::
+/=  gall-raw  /sys/vane/gall
+::
+=>  |%
+    +$  subs  (jar [@ta path] [duct bone])
+    --
+::
+:-  %say
+|=  [[now=@da eny=@uvJ bec=beak] arg=~ peer=(unit @p) dry=? veb=?(%1 %2 %3 %x)]
+::
+=/  our-gall  (gall-raw p.bec)
+=/  gall-yokes
+  .^((map dude:gall yoke:our-gall) %gy /(scot %p p.bec)//(scot %da now)/$)
+=;  ducts=(list [duct bone])
+  :-  %helm-gall-lave
+  ~?  dry  "#{<(lent ducts)>} stale incoming subscriptions"
+  ~?  ?=(%1 veb)  ducts
+  dry^~
+::
+%+  roll  ~(tap by gall-yokes)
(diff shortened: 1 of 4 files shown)


yosoyubik:
I've added some other verbose options to the generator:

```hoon
::    |gall-lave, =veb %1  :: for each subscription path, print bone info
::    |gall-lave, =veb %2  :: number of stale subscriptions per agent
::    |gall-lave, =veb %3  :: subscription only present in %gall
::    |gall-lave, =veb %31 :: ... since N are already in closing in %ames
::    |gall-lave, =veb %32 :: ... since N are already corked in %ames
```

If the bone associated with the agent-duct is already in closing/corked, we only delete the incoming subscription in %gall, otherwise %gall will also emit a %kick to the subscriber which in turn should send a %cork to us later.

(note: still untested in live ships)

yosoyubik:
I've updated `|close-flows` because it was missing cases where, the most recent flow in %ames for a given subscription was already deleted from boat.yoke, but we were only corking it if it has received a %nack. 

Now, if the most recent subscription is gone from the outbound subscription map, we mark it also as "corkable".

yosoyubik:
Testing this I discovered that it seems _safe_ to %cork %pokes that have been %nacked—I haven't encountered any weird behaviors by turning on/off the situation when the %poke is %nacked, even after the flow has been removed on both sides... (is this a "[...] a feature not a bug" type of situation?)

yosoyubik:
_(from the latest commit)_

Previously `|close-flows` and `+stale-flows` only used information 
available in `%ames`. This meant that some flows that were stale,
in particular, the latest flow associated with a given subscription,
were only corked if they had received a `naxplanation`, but this was
not always true -- these subscriptions were missing from the
outgoing subscription map, `boat.yoke`.

Both generators are updated to take account of this information, along
with extended and improved output, to match the same number of stale
flows that can be corked in both.

yosoyubik:
Consider an app that uses the same wire when it sends either a `%poke` or a `%watch`. From the point of view of `%ames`, this will create two separate flows, with the same duct for both, with the difference that the `%watch` will have a nonce in the wire. This is not true for pre-nonce subscriptions where nonces where not used. From the point of view of `%ames` these two look the same.

We can infer if this is a `%watch` if there is an entry in either `boat.yoke` or `boar.yoke`... unless we are dealing with an `%ames`/`%gall` desync where `%gall` has removed all information about this subscription, but `%ames` still has an open flow. For this case, we can't know for sure if it's a `%poke` or a `%watch` so in the latest version I decided to only `%cork` for this case, if we have received a `%nack`. This should guarantee that we don't `%cork` a flow for a `%poke`, at the cost of leaving old pre-nonce (re)subscriptions in `%ames`'s state.

Issue #6805, opened 2023-09-25
Snapshot patches are larger than expected


philipcmonk:
Every few minutes, the runtime applies a patch to its on-disk snapshot.  This pauses the process, so the size is important.  In particular, for many large ships (~nibset-napwyn, ~wicdev-wisryt, ~natnex-ronret), this is around 600MB.  If the maximum acceptable pause is 1 second, this requires a disk which can handle 600MB/s of throughput (or maybe twice that, because it writes the patchfile, then applies it?), which is extremely high.

There are several ways to attack this.

Perhaps there's a bug where we're dirtying pages that don't need to be dirtied?  It could be useful to try a debug assertion that when we write a page, it didn't already match.

Otherwise, perhaps we're changing state too often?  And if so, what would cause that?  Before investigating, I would guess this may come from Ames, which retries many packets every couple minutes.

While I wouldn't trust anything too much, initial investigation seems to confirm that as part of the issue.  If you run with `-v` (and turn off `|verb`), you can see how big the patches are, every time they're written.  If you run with `--snap-time 1`, you increase the snapshot frequency to every minute, and on 412K if you run `|pass [%a %stir %dead]`, then you'll enable dead flow consolidation, which will make all dead flows (most Ames retries) retry at the same time.  This way, you get alternating printouts: one covering a time period with ames retries, and one without.

Trying this on a `-L` copy of ~nibset-napwyn, I see that when there is a retry, snapshot size is around 550MB.  Without a retry, if I hit enter a few times (to simulate the various packets it would likely receive if it were not in `-L` mode), it varies between 70-200MB.

This is not conclusive, but it's pretty suggestive.  Eg, this does not eliminate the possibility of a bug where reading is dirtying pages, or a bug where we're writing pages in ephemeral free space, but it does suggest that something about ames retries is causing much of the issue.  It's very possible that something about the retries are dirtying pages for no good reason.  For example, they're incrementing a retry counter; perhaps that should stop incrementing at 100, to avoid changing state most of the time.  Maybe other state is changing, and even if it's not, we may be reallocating identical nouns, freeing the old ones.  Eg perhaps `+abet` should explicitly check whether the `put` is necessary.

Whatever approach we take, we should be able to measure the effect of each change.  The last thing we need is to add complexity without knowing what actually makes a difference.  But there's probably some small things we can do to improve the situation a great deal.

See urbit/vere#429

philipcmonk:
Some assorted measurements on ~nibset:

[...]

PR #6823, opened 2023-10-09
ames: stateless dead flow consolidation

pkova:
Stateless dead flow consolidation to reduce write amplification caused by modifying `next-wake`, `last-sent` and `tries` in ames state. We have seen a 50% reduction in snapshot write IO on a busy publisher ship with this PR applied and dead flow consolidation turned on.

Partially addresses #6805.


joemfb (approved):
this will do

Issue #6768, opened 2023-08-22
+vats is broken for some states


joemfb:
As reported in the UF group:

```
~ship:dojo> +vats|
bail: 4
! /cw/~ship/~zod/~2023.8.22..08.57.30..0953
[%dojo-lame %arvo-response]
> +vats
```

This ship is up-to-date (%base at `0v1s.gbsom.15m55.mpfe9.umbfm.3an52.iaujf.n1c2s.mtfj0.dg0ga.2q87f`).

sethfork:
Also having this issue. I remember seeing this could be related to empty desks

silnem2:
I have this issue as well, but I don't think it's directly related to `+vats`. It looks like a desk called `%~zod` was added to my ship at some point, somehow bypassing the `(sane @tas)` desk name validity check in clay:
``` 
> .^((set desk) %cd /=//=)
[n=%~zod l={%groups %base %quorum %landscape %webterm %garden %rumors %trill %pals %blog %portal} r={%kids %talk}]
```

bradyab:
I'm having this as well, i can't nuke or uninstall the zod desk either:

```
> dojo: |nuke %~zod
nest-need [term=atas %~]
dojo: nest-have [%~zod %~7
dojo: bad-argument (supplied sample incorrect)
```

mikolajpp:
This is interesting. A invalid term `%~zod` would definitely not bypass `+sane`
```
> ((sane %tas) '~zod')
%.n
```
`%~zod` -- a constant atom `~zod` would, but it is a desk `%$`.

When you type `|nuke %~zod` in dojo you are actually trying to nuke the empty desk `%$`, as %~zod is not a term. It is 0 and gets cast down to a `@tas`
```
> `@tas`%~zod
%$
```

When you do a scry to get available desk, there indeed seems to be a desk named '~zod'. To nuke it, one would instead have to 
```
> |nuke '~zod'
```

tinnus-napbus:
fyi `|nuke` operates on agents not desks. You can do `|nuke %the-desk, =desk &` to nuke every agent on a desk but otherwise it's `|nuke %the-agent`

silnem2:
@mikolajpp Yeah, the `(sane %tas)` check was added to `%fuse`, `%park`, and `%merg` in https://github.com/urbit/urbit/pull/6304, but other requests (e.g. `%zest`) can add references to desks that don't exist to clay's map of desks and don't include a sanity check.

[...]

PR #6824, opened 2023-10-10
vats: handle insane desk names

tinnus-napbus:
less-than-ideal fix for #6768 

This makes +vats check desk sanity before trying to scry that desk

Somebody has presumably made an error in their app where they pass Clay a task with the ship name in the desk field. Clay doesn't sanity-check desks in all its tasks eg `%zest` which can result in an insane desk nominally existing in Clay's state. The problem afaics is that arvo's ++de-beam implicitly sanity-checks desks with `(slaw %tas` which means scries fail for insane desks.

Presumably Clay should be updated to sanity-check desks for all tasks but idk how you fix the existing insane entries. Either you need to update clay to expunge insane desks or ++de-beam should stop sanity-checking the desk maybe

Thoughts on proper solution @philipcmonk ?

### DIFF
--- a/pkg/base-dev/sur/hood.hoon
+++ b/pkg/base-dev/sur/hood.hoon
@@ -51,10 +51,11 @@
   =/  deks
     ?~  desks
       %+  sort
-        (sort ~(tap in -.prep) |=([[a=@ *] b=@ *] !(aor a b)))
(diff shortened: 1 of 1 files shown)


tinnus-napbus:
also as an aside vats is horrifically inefficient & complex & needs to be rewritten

belisarius222 (approved):

PR #6825, opened 2023-10-10
lull: add egg-any tagged union for gall scry

pkova:
As suggested by @joemfb in https://github.com/urbit/urbit/pull/6785#issuecomment-1753142434

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -2754,6 +2754,7 @@
             sky=(map spur path-state)
             ken=(jug spar:ames wire)
     ==  ==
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/lull.hoon
```diff
@@ -2754,6 +2754,7 @@
             sky=(map spur path-state)
             ken=(jug spar:ames wire)
     ==  ==
+  +$  egg-any  $%([%15 egg])
```
joemfb, line 2757: Can you add a comment to `$spore` in gall, noting that changes to its version tag need to be replicated here?

PR #6826, opened 2023-10-11
zuse: fix +de:base64:mimes:html

pkova:
`+de:base64:mimes:html` is misbehaving in two scenarios.

1. We do a byte swap on the result at the end of decoding. This means that any leading zeroes that should become trailing zeroes get dropped. An example input for this problem is `(de:base64:mimes:html 'AAD+wg==')`.
2. We pad zeroes at the end for every `=` padding character. This is incorrect behavior, we should instead drop two bits from the end for every padding character. An example input for this problem is `(de:base64:mimes:html '/si=')`.

This PR fixes these issues. Thanks to `~litlep-nibbyt` for discovering the problem and helping with the solution.


sigilante (approved):
LGTM

PR #6832, opened 2023-10-19
ames: prefix abet core arms

yosoyubik:
A common style when using the `+abet` core pattern is to prefix every arm of the core with its two letter (core) identifier—this is specially useful when there are several nested cores, so we can easily know what ++abed/++abet arm we are calling.

Here we do just that (adjusting everything to fit within an 80 column margin)

(Note: draft until I test this on a live ship)


belisarius222 (approved):

yosoyubik:
I've just tested this on a live ship, leaving and joining a couple of groups and nothing weird shows up, so opening up for review.

PR #6835, opened 2023-10-21
lull, ames: do not store jammed messages in unsent-messages

pkova:
The upcoming introduction of persistent nock caching allows us to get rid of the insane `+dedup-message` arm in ames. This wonky function existed to ensure reasonable memory usage in scenarios such as when a busy publisher is sending the exact same `%fact` to many ships. 

Pruning the persistent nock cache with `+dedup-message` removed presents a risk of losing structural sharing in ames `unsent-messages`. This PR refactors `unsent-messages` to store the messages as nouns instead of jammed atoms to fix the problem.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2471,7 +2471,7 @@
           %-  %^  ev-trace  odd.veb  her
               |.("parsing old wire: {(spud wire)}")
           peer-core
-        abet:(on-memo:peer-core bone [%boon payload] %boon)
+        abet:(on-memo:peer-core bone [%boon payload])
       ::  +on-plea: handle request to send message
       ::
       ++  on-plea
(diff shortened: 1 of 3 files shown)


belisarius222 (approved):

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -5261,6 +5368,34 @@
...
+    %-  ~(gas to *(qeu message))
+    %+  turn
+      ~(tap to unsent-messages.m)
+    |=  b=message-blob
+    ^-  message
+     =/  hed
+       ?:  =(1 (end 0 bone))          %plea
+       ?:  =(0 (end 0 (rsh 0 bone)))  %boon
+       %naxplanation
+    ;;(message [hed (cue b)])
```
joemfb, line 5397: We should `~+` calls to `+cue` here and for the events.

PR #6836, opened 2023-10-23
ames: add STUN response handling

yosoyubik:
WIP


yosoyubik:
Current state machine for the %ping app.

(Note: every successful STUN response pokes the ping app with a %stop, but we just ignore them—if in %off plan i.e. subsequent success. Every STUN failure—not hearing a response for more than ~s5 or ~s25—also pokes the ping app with a %kick, but we also discard them, taking care of not ignoring kicks from %born events)

```mermaid
stateDiagram

  state %ping {
    
    state "OFF (only STUN)" as off
    state "NAT (ping ~s25)" as nat
    state "PUB (ping ~m5)" as pub
    state "ONCE (one ping %poke)" as once
    state if_ack <<choice>>
 
    [*] --> nat: on {%born %kick}
    nat --> nat: on %fail
    nat --> pub: on %no-nat
    pub --> nat: on %nat
    pub --> pub: on %fail
   
    nat --> off: on %stop
    pub --> off: on %stop

    off --> nat: on %fail 
    off --> off: on %stop

    off --> [*]

    [*] --> once: different ip/port
    once --> if_ack
    if_ack --> [*]: if %poke-ack -- don't start timers
    if_ack --> once: waiting

  }
```

pkg/arvo/sys/lull.hoon
```diff
@@ -803,6 +804,7 @@
         [%kroc bones=(list [ship bone])]
         $>(%plea vane-task)
         [%deep =deep]
+        [%stun =stun]
```
joemfb, line 807: For future flexibility, this task should include a ship (or possibly a ship and a lane).

joemfb (changes requested):

PR #6838, opened 2023-10-24
ames: do not ignore our peer-state

pkova:
Consider the following scenario:

`~sampel-palnet` is behind a NAT. `/app/ping` will try to ping its sponsor `~sampel` every 25 seconds. `~sampel-palnet` just booted and has no route to `~sampel` so he will try to contact `~sampel` through the galaxy `~pel`.

`~sampel` received the ping. It will respond directly to `~sampel-palnet`, but knowing that the ping came through an indirect route the star will also send the ack through `~sampel-palnet`s sponsorship chain. This is where things go wrong.

`~sampel` will enter `+send-blob`, send the direct ack, but when it enters `+try-next-sponsor` it will encounter a missing peer-state for the sponsor. This sponsor is of course `~sampel` itself. Now `~sampel` will enter this code path : https://github.com/urbit/urbit/blob/10db04b944c6cdb7ed0d4ef3a17c63dc7bf769fa/pkg/arvo/sys/vane/ames.hoon#L3005-L3010

calling `+enqueue-alien-todo` for himself! This is very bad and results in `~sampel-palnet` never being able to receive the ack from `~sampel`. Looks like this bug has been present since 238a36de1de1f2419e005f85c9dab0a4a99b463f.

There are multiple possible approaches to fix this, but I think this one is the simplest. We create a dummy state migration that asks jael for our %public-keys, triggering `+on-publ-full` and eventually `+meet-alien`, clearing all garbage in `packets.todos`.

Thanks to `~sibnes` for reporting this bug.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -559,7 +559,7 @@
 ::    dead:        dead flow consolidation timer and recork timer, if set
 ::
 +$  ames-state
-  $+  ames-state
+  $+  ames-state-17
   $:  peers=(map ship ship-state)
       =unix=duct
       =life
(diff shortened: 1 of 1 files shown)


joemfb (approved):
great catch!

PR #6840, opened 2023-11-04
eyre: simplify the login page ux a little bit

Fang-:
We remove the local<->eauth switching tabs entirely. The only way to see the eauth login page is if the url contains an "eauth" query parameter. (Entering the local ship name into the eauth form still brings you to the normal login screen.)
The "proceed as guest" blurp no longer shows the guest identity, and we clarify the language to indicate that it is not guaranteed to result in access to content.

Both of those changes help simplify the look and feel of the login page. Various users had complained that it was confusing, given that logging into your own ship is still the common case, and not everyone knows what "euauth" is.

For applications redirecting to the login page, it is still recommended to add the ?eauth GET parameter if non-local sessions are supported. For applications that don't it should now be more obvious that logging in through eauth won't do anything (since it's not presented as an option).

We also remove some debugging prints that somehow remained in the javascript.

Screenshot of the page on the eauth screen:

image

Screenshot of the page on the local login screen:

image

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -459,8 +459,6 @@
                 name.focus();
               }
               function doEauth() {
-                console.log('mb get value from event', event);
-                console.log('compare', name.value, our);
                 if (name.value == our) {
                   event.preventDefault();
                   goLocal();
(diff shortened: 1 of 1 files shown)


urcades (commented):
Simplification is looking great — I only have a question regarding what the "log in as guest" button does on both of these screenshots? Does this redirect to the eauth screen? 

If it does, it might make sense to edit the text on the eauth screen to switch the state of the screen to the ordinary/normal login.

Fang-:
No, the "proceed as guest" link simply follows the redirect that would be followed post-login, without actually logging in. In some cases, this may bring you back to the app you came from. In other cases, this may redirect you back to the login page again. (The login page itself has no way of knowing what will happen, so "try to proceed" is the most accurate messaging we can give.)

urcades (approved):
@Fang- makes sense to me, stamped

Issue #6846, opened 2023-11-17
%gall: static agent namespace has fencepost bugs in %cull'ing


joemfb:
from `~rivmud-fabwen`

> found some strange behavior around remote scry, might be a bug?
>
> example:
> * I %grow some data to path /my/data (let's say 4 times)
> * I scry that data at the latest revision `[%ud 3]`
> * It returns the latest version of the data as expected
> * I %cull up to the latest version `[%pass /call/back %cull [%ud 3] /my/data]`
> * I scry the data at version 3, it fails. So far everything appears normal.
> * I %grow to the same path again with new data
> * If I scry versions before 3, it fails as expected
> * I scry version 3 again, and now it returns the updated data I just published
>
> So what I'm seeing is that if I %cull a path, I can republish to the latest version number, meaning that I did the same scry to the same path (including version) and got different data at different times.
>
> Also ran into something else that seems to be a bug too. I can %cull with a `[%ud @ud]` case, but not a `[%da @da]` case. If I try culling by `[%da now.bowl]`, it fails with a crud!.
>
> With the `@da` issue, at first I thought the problem was that it needed an existing version, not just something after the latest, but I was able to %cull with say `[%ud 10]` when something had only been published <10 times, so that isn't the case for `@ud` at least

PR #6847, opened 2023-11-17
gall: fix remote scry bug

pkova:
As reported by @rivmud-fabwen in `~dister-dozzod-lapdeg/battery-payload`.

Consider the case where you `%grow` a path `/foo` 3 times from a gall agent. After this you decide to `%cull` the same path with the case of `[%ud 2]`, removing the cases `[%ud 0]`, `[%ud 1]` and `[%ud 2]`. Everything works as expected so far, however! The next `%grow` that happens for the path `/foo` gets written to case `[%ud 2]`, completely breaking referential transparency. This happens because of a off-by-one error in `+ap-grow`.

That's not all folks. In the previous case if you decide to `%cull` with a case of `[%ud 1]` you also cull the case `[%ud 2]` because of another off-by-one error in `+ap-cull`.

This PR fixes both of the aforementioned issues. I also decided to enforce culling cases to be between the first and last cases that are currently bound, so culling with `[%ud 1.000.000]` with only cases 0, 1 and 2 existing will no-op and print the message: `gall: %your-agent: cull [[%ud p=1.000.000] /foo] out of range, min: 0, max: 2`

Thanks to @yosoyubik for figuring this out with me.

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -1021,7 +1021,7 @@
       =/  ski  (~(gut by sky.yoke) spur *path-state)
       =-  ski(fan (put:on-path fan.ski -< -> &/page))
       ?~  las=(ram:on-path fan.ski)
-        [(fall bob.ski 0) now]
+        [?~(bob.ski 0 +(u.bob.ski)) now]
       :_  (max now +(p.val.u.las))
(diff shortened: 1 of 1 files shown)


joemfb (approved):
this will do

joemfb:
Fixes #6846

PR #6850, opened 2023-11-20
eyre: append set-cookie header rather than clobbering

ixv:
Apps can set-cookies in http responses, but eyre overwrites them. This PR makes eyre append its cookies instead.


pkova (approved):

tacryt-socryp:
This loop logic doesn't make sense to me. What if the set-cookie header isn't the last one in the list? Won't it truncate the remaining headers in that case?

I don't understand. @pkova

ixv:
@tacryt-socryp It wont. All it does is append to the list, unless there's a duplicate (in both key and value), then it just returns the list unchanged.

PR #6852, opened 2023-11-21
gall: migrate sky to versioned path

pkova:
Since we noticed %gall breaking referential transparency in #6847 we have decided to change the interface to incorporate a version tag in the path. This allows us easier migrations if the namespace ever changes in the future.

This means that a path with the format `/~zod/test-agent/2//foo` becomes a path with the format `/~zod/test-agent/2//1/foo`. We will migrate all data from the old unversioned paths to the new versioned paths automatically. Note that since referential transparency is lost we will recover whatever we can from the old `sky` and start the aeons in the path from 1. This means that if you had culled all older revisions and had data at the paths `/~zod/test-agent/10//foo`  `/~zod/test-agent/11//foo` and  `/~zod/test-agent/12//foo`, this data will be migrated to live at  `/~zod/test-agent/1//1/foo`  `/~zod/test-agent/2//1/foo` and  `/~zod/test-agent/3//1/foo` respectively.

Queries against the old unversioned path scheme will block forever once kelvin 411 ships.

This PR also contains some housekeeping so I could do `+$  spore-15  spore`

A few questions. Did I put the version tag in the right place in the namespace hierarchy? Should the version tag itself be something else than just 1?

### DIFF
--- a/pkg/arvo/sys/lull.hoon
+++ b/pkg/arvo/sys/lull.hoon
@@ -2762,7 +2762,7 @@
             sky=(map spur path-state)
             ken=(jug spar:ames wire)
     ==  ==
-  +$  egg-any  $%([%15 egg])
+  +$  egg-any  $%([%15 egg %16 egg])
   +$  bowl                                              ::  standard app state
     $:  $:  our=ship                                    ::  host
             src=ship                                    ::  guest
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -42,9 +42,9 @@
 ::  $move: Arvo-level move
 ::
 +$  move  [=duct move=(wind note-arvo gift-arvo)]
-::  $state-15: overall gall state, versioned
+::  $state-16: overall gall state, versioned
 ::
-+$  state-15  [%15 state]
++$  state-16  [%16 state]
 ::  $state: overall gall state
 ::
 ::    system-duct: TODO document
@@ -150,8 +150,7 @@
 ::  remember to duplicate version tag changes here to $egg-any:gall in lull
 ::
 +$  spore
-  $:  %15
-      system-duct=duct
+  $:  system-duct=duct
       outstanding=(map [wire duct] (qeu remote-request))
       contacts=(set ship)
       eggs=(map term egg)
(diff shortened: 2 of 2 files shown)


pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1062,7 +1062,7 @@
         %+  trace  &
         [leaf+"gall: {<agent-name>}: cull {<[case spur]>} no paths"]~
       =/  fis  (need (pry:on-path fan.u.old))
-      ?.  &((gth yon key.fis) (lte yon key.u.las))
+      ?.  &((gte yon key.fis) (lte yon key.u.las))
```
pkova, line 1065: I also noticed it was my turn to make a off-by-one in my %cull range check :(

pkg/arvo/sys/vane/gall.hoon
```diff
@@ -1791,12 +1791,16 @@
...
         [%pass * ?(%agent %arvo %pyre) *]  $(caz t.caz, fex [i.caz fex])
         [%give *]  $(caz t.caz, fex [i.caz fex])
         [%slip *]  !!
+        ::
+          [%pass * %grow *]
+        $(caz t.caz, ap-core (ap-grow ['1' +<.q.i.caz] +>.q.i.caz))
+          [%pass * %tomb *]
+        $(caz t.caz, ap-core (ap-tomb +<.q.i.caz ['1' +>.q.i.caz]))
+          [%pass * %cull *]
+        $(caz t.caz, ap-core (ap-cull +<.q.i.caz ['1' +>.q.i.caz]))
```
joemfb, line 1803: I wasn't picturing that we would change the paths themselves -- I expected we'd just add an explicit version check in handler for the `%x`, `%w`, `%z`, and `%t` cares.

joemfb (changes requested):
@pkova thanks for taking this on. i had a different approach in mind, we should sync on this asap.

pkg/arvo/sys/lull.hoon
```diff
@@ -2762,7 +2762,7 @@
             sky=(map spur path-state)
             ken=(jug spar:ames wire)
     ==  ==
-  +$  egg-any  $%([%15 egg])
+  +$  egg-any  $%([%15 egg %16 egg])
```
midden-fabler, line 2765: Should be `[[%15 egg] [%16 egg]]`.
#6855 should help ensure `%spore` and `$egg` are kept in sync.

joemfb (approved):
this will do, tyfys

PR #6857, opened 2023-11-23
dojo: fix incorrect beak desk on non-base generator

### LINKED ISSUE #6856: generators: q.bec should contain generator desk, not desk of working directory
**Describe the bug**

When invoking a generator, the `beak` contains information about a desk. When a generator is invoked by the `+desk!gen` syntax, the beak points to the current desk of the dojo, not the desk containing the generator. 

**To Reproduce**
1. Create a new desk with `base-dev`, for instance `%sandbox`
2. Add this generator to `gen` in the new desk: https://gist.github.com/eamsden/2dfe9a3b11d43130a13152a8555f0b02
3. Invoke the generator as in `+sandbox!wrong-beak`

**Observed behaviour**
`desk: %base` is slogged

**Expected behaviour**
`desk: %sandbox` is slogged

**Screenshots**

 - OS: linux x86_64
 - Vere 2.12, Zuse %412
 - Your ship's `%base` hash: `pfbl0`
 
(This behavior is observed on a fakezod)

**Additional context**
The current behavior makes impossible a useful case where a generator may want to load files from its own desk. In my case, I have a build script as a generator for Ares codegen, which should get the Ares codegen source files from the generator's source desk and build them into a `(trap vase)` which can then be jammed and included in Ares.

pkova:
Fixes #6856

### DIFF
--- a/pkg/arvo/app/dojo.hoon
+++ b/pkg/arvo/app/dojo.hoon
@@ -807,7 +807,7 @@
     ::
     ++  dy-run-generator
       !.
-      |=  [cay=cage cig=dojo-config]
+      |=  [cay=cage cig=dojo-config =desk]
       ^+  +>+>
       ?.  (~(nest ut [%cell [%atom %$ ~] %noun]) | p.q.cay)
         ::  naked generator; takes one argument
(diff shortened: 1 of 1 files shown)


eamsden:
@pkova I'll try this out tomorrow

eamsden:
Verified that this fixes #6856, thanks @pkova !

joemfb:
It seems like some generators need access to the current working directory (`+ls`, for example), while others need access to their source directory (the use-case motivating #6856). Prior to the addition of the `+foo!bar` syntax, those were always the same directory; to run `+some-generator out a given `desk`, you had to first `=dir /=desk`, then `+some-generator`. Now that they can diverge, I wonder if generators should be given both beaks.

eamsden:
Both beaks seems like a reasonable thing to do, but the calling convention change would break every extant `%say` generator right?

joemfb:
After further review, `+ls` does *not* use its environment, but always requires an explicit path argument. I don't see anything using the beak as I've described.

I think this change is fine -- if we decide we want access to both beaks, we should introduce new generator interfaces under new tags (%say, %ask, &c).

PR #6858, opened 2023-11-24
hoon: more vase mode arms for (trap vase)

drbeefsupreme:
Added a few gates for working with `(trap vase)`s. One of them, `+swel`, was in `lib/pill.hoon` so I removed it from there and put it into hoon.hoon. We also now have a deferred `+slam` and deferred `+slop`.

These are just the ones I personally needed - are there any other `vase` operations I should make `(trap vase)` versions of while I'm at it?

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -11457,6 +11457,25 @@
     %-  ~(play ut p.vax)
     [%wtgr [%wtts [%leaf %tas -.q.vax] [%& 2]~] [%$ 1]]
   (~(fuse ut p.vax) [%cell %noun %noun])
+::  +shot: deferred +slam
+::
+++  shot
+  |=  [gat=(trap vase) sam=(trap vase)]
+  ^-  (trap vase)
+  =+  :-  ^=  typ  ^-  type
+          [%cell p:$:gat p:$:sam]
+      ^=  gen  ^-  hoon
+      [%cnsg [%$ ~] [%$ 2] [%$ 3] ~]
+  =+  gun=(~(mint ut typ) %noun gen)
+  |.  ~+
+  [p.gun (slum q:$:gat q:$:sam)]
+::  +slew: deferred +slop
+::
+++  slew
+  |=  [hed=(trap vase) tal=(trap vase)]
+  ^-  (trap vase)
+  |.  ~+
+  [[%cell p:$:hed p:$:tal] [q:$:hed q:$:tal]]
 ::  +swat: deferred +slap
 ::
 ++  swat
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -11457,6 +11456,24 @@
...
+  =+  gun=(~(mint ut typ) %noun gen)
+  |.  ~+
+  [p.gun (slum q:$:gat q:$:sam)]
+::  +slew: deferred +slop
+::
+++  slew
+  |=  [hed=(trap vase) tal=(trap vase)]
+  ^-  (trap vase)
+  |.  ~+
+  [[%cell p:$:hed p:$:tal] [q:$:hed q:$:tal]]
```
joemfb, line 11476: This needs a minimized subject outside the `|.` (with `=>  +<` or equivalent). Inside, you should pin `$:hed` and `$:tal` to avoid unnecessarily reevaluating them.

pkg/arvo/sys/hoon.hoon
```diff
@@ -11457,6 +11456,24 @@
...
+::
+++  shot
+  |=  [gat=(trap vase) sam=(trap vase)]
+  ^-  (trap vase)
+  =/  [typ=type gen=hoon]
+    :-  [%cell p:$:gat p:$:sam]
+    [%cnsg [%$ ~] [%$ 2] [%$ 3] ~]
+  =+  gun=(~(mint ut typ) %noun gen)
+  |.  ~+
+  [p.gun (slum q:$:gat q:$:sam)]
```
joemfb, line 11469: This needs a minimized subject outside the `|.` (with `=>  [typ=p.gun +<.$]` or equivalent). To keep that subject maximally minimal, you should also inline the definition of `+slum` inside the `|.`

pkg/arvo/sys/hoon.hoon
```diff
@@ -11465,6 +11482,16 @@
   =/  gun  (~(mint ut p:$:tap) %noun gen)
   |.  ~+
   [p.gun .*(q:$:tap q.gun)]
```
joemfb, line 11484: This is a preexisting condition, but the `|.` here should also get a minimal subject with `=>`.

joemfb (changes requested):
These look reasonable. `+swel` is a little odd, and can be handled by the caller with `+swat` and `+slew`. But it's also convenient...

Some nits:

joemfb:
@drbeefsupreme looking at CI, `+slew` already exists, so you'll need to find another name for it.

Also, please target `next/kelvin/411` with these changes (you can edit this PR, no need for a new one).

drbeefsupreme:
`+swel` is from `lib/pill.hoon`, all I did was move it into `hoon.hoon.` I'm a bit confused about the CI issue - I'm guessing its trying to build a pill with the old version which has the `+swel` arm, and this is conflicting somehow with the new `hoon.hoon` version?

I'll address your other comments later today.

drbeefsupreme:
I renamed `+swel` to `+sock` but CI is still complaining about there being multiple `+swel` arms, even though there are no longer any arms named `+swel` in the whole branch. So I think this confirms its a CI issue.

drbeefsupreme:
Oops, I misread `+slew` as `+swel. Nevermind.

drbeefsupreme:
I renamed `+slew` to `+sock`, but it looks like `lull.hoon` already has a `$sock`, though it appears to be completely unused. I picked `sock` since it has a similar meaning to `swat`, but that was when I thought I was replacing `+swel`, not `+slew`.

Now that I'm actually thinking about names, since all the normal vase mode arms start with `s`, it probably makes sense to pick a different starting letter for `(trap vase)` arms. Then it will be easier to tell at a glance what you're working with.

joemfb (dismissed):
i don't feel strongly about the names. @pkova?

this will do

PR #6871, opened 2023-12-08
Add ;. micdot rune for reversed association of ;: miccol.

sigilante:
`;:` miccol can be counterintuitive for developers, since its order of evaluation is `5 4 3 2 1`.

`;.` micdot reverses the order by flopping it for `1 2 3 4 5`.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -6466,7 +6466,8 @@
     [%sgzp p=hoon q=hoon]                               ::  ~!  type on trace
   ::                                            ::::::  miscellaneous
     [%mcts p=marl:hoot]                                 ::  ;=  list templating
-    [%mccl p=hoon q=(list hoon)]                        ::  ;:  binary to nary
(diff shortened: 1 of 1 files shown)


sigilante:
It's catching on the parser for `;.` so I'll have to debug.

sigilante:
I think some form of the Sail runes are catching it.

Alternatives are to rename it (altho I think `;:`/`;.` a good pair) or to ferret out which Sail case is hitting it first.

sigilante:
Missed a parser change.

```
> ;:(weld ~[1] ~[2] ~[3])
~[1 2 3]

> ;.(weld ~[1] ~[2] ~[3])
~[1 2 3]
```

Still doesn't do what we want for `++corl`.

sigilante:
Got it.  It flips the polarity on `cork` though, which may also be counterintuitive.

```hoon
> `@t`(;:(corl |=(a=@ +(a)) head trip) 'foo')
'g'

> `@t`(;.(cork trip head |=(a=@ +(a))) 'foo')
'g'
```

sigilante:
Too marginal to justify new rune for the time being.

PR #6873, opened 2023-12-10
clay: print download progress for desk updates

yosoyubik:
WIP 
rate

(Note: possible alternative version of https://github.com/urbit/urbit/pull/6867; instead of scrying the rate out of `%ames` we rely on `%ames` notifying any subscribers for a `%keen`—In this case `%clay` while just send the `%rate` `%gift` to `%kiln`. Leaving as a draft so we can compare what approach is more ergonomic)

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -770,7 +770,7 @@
       (clap a b furl)
   |-  ^-  (list (unit toro))
   =+  b=.^(arch %cy a)
-  ?:  ?=([^ ~] b)  (snoc c `(fray a))  
+  ?:  ?=([^ ~] b)  (snoc c `(fray a))
   =?  c  ?=(^ fil.b)  (snoc c `(fray a))
   %-  zing
   %+  turn  ~(tap by dir.b)
@@ -796,8 +796,8 @@
     %+  skim  desks
     |=  dek=desk
     ?:  (~(has in .^((set desk) %cd /(scot %p our)//(scot %da now))) dek)
-      &  
-    ~>  %slog.(fmt "desk does not yet exist: {<dek>}")  |   
+      &
+    ~>  %slog.(fmt "desk does not yet exist: {<dek>}")  |
   |=(=desk [%pass /kiln/suspend %arvo %c %zest desk %dead])
 ::
 ++  poke-sync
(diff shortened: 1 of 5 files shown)


yosoyubik:
rate-all


```
Currently this only supports the view of how many paths are in flight.
The facilities for knowing the the size of any/all paths in the desk are
in place but are still a work in progress.

There is also a thread (-whit) that calculates the size of data at a path
based on the number of fragments and the size of each, as sent by the
publisher.
```

yosoyubik:
demo of usage of global progress rate:

rate-size

yosoyubik:
naive bandwidth demo:

rate-band

jalehman:
This is really great work! A thought that comes to mind:

Having this in the dojo as toggle-able behavior is desirable. It'll also be the case that many applications that eventually leverage fine for pulling content from other nodes will also want to be able to display download bandwidth/progress. 

Providing documentation on how it works as a guide in the docs should accompany its release.

yosoyubik:
Current status supporting both |fine and |mesa (directed messaging):


https://github.com/urbit/urbit/assets/1033194/2374ab27-5cea-48b0-b113-2d4fa587c684

(note: the logic that supports this has not been pushed yet, together with a bunch of fixes and changes, also missing from https://github.com/urbit/urbit/pull/6932)

yosoyubik:
Current status supporting both |fine and |mesa (directed messaging):

https://github.com/urbit/urbit/assets/1033194/bd1f7de6-c87a-4b10-876c-0cecdbc362ab

(PS: the logic that supports this has not been pushed yet, together with a bunch of fixes and changes, also missing from https://github.com/urbit/urbit/pull/6932)

(PPS: i removed the bandwidth print because currently we use the time spent in between hearing responses, which come very fast, and then the bandwidth is a bit misleading—most of the time is spent producing the data, but for that the progress bar just sits idle)

yosoyubik:
This branch was out of date with the latest version of Directed Messaging, also using some idiosyncratic patterns for interfacing between ames and its listeners, if they were subscribed to gifts other than the %tune/%sage noun bounded at a specific path.

Here I handle listeners by tracking what kinds of things they want to hear about this path—so far only adding the %rate gift. There are some awkward things about how we deal with this in |fine since we don't have a %whey namespace, that expose themselves clearly when we deal with the size of a desk in %clay (there could be some refactoring in this area to not ask for the sizes of all files in a desk, and instead just ask one time about the whole desk).

(Note: compiles but untested, it very likely doesn't work yet)

PR #6876, opened 2023-12-13
all: enforce the gang when scrying

pkova:
We've had the gang system for quite a while to set privilege levels when scrying. A gang of `[~ ~]` indicates that you are "root" and vanes should never hide data from you. A gang of `~` indicates that anybody could be asking, vanes should block requests for private data. The gang can also be a set of ships, in which case the permissions have to be checked for each one. The interface has existed for a while but the gang system had been unimplemented until now. This PR goes through every scry callsite and arm in every vane and enforces the gang. Some scry callsites are in vere, see urbit/vere#569.

This fixes a security hole with #6790 where you could use unencrypted remote scry to fetch encrypted private data.

Compiles but as of yet untested.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2006,7 +2006,7 @@
         %+  skim
           ;;  (list [@da duct])
           =<  q.q  %-  need  %-  need
-          (rof ~ /ames %bx [[our %$ da+now] /debug/timers])
+          (rof [~ ~] /ames %bx [[our %$ da+now] /debug/timers])
         |=([@da =duct] ?=([[%ames %recork *] *] duct))
       ::
       =^  moz  u.cached-state
@@ -2078,7 +2078,7 @@
       ++  get-sponsors
         ;;  (list ship)
         =<  q.q  %-  need  %-  need
-        (rof ~ /ames %j `beam`[[our %saxo %da now] /(scot %p our)])
+        (rof [~ ~] /ames %j `beam`[[our %saxo %da now] /(scot %p our)])
       ::
       +|  %tasks
       ::  +on-take-flub: vane not ready to process message, pretend it
@@ -2273,7 +2273,7 @@
         ++  do-rift
           =/  =rift
             =-  ~|(%no-rift (,@ q.q:(need (need -))))
-            (rof ~ /ames %j `beam`[[our %rift %da now] /(scot %p our)])
+            (rof [~ ~] /ames %j `beam`[[our %rift %da now] /(scot %p our)])
           ?:  =(rift rift.ames-state)
             event-core
           ~&  "ames: fixing rift from {<rift.ames-state>} to {<rift>}"
@@ -2298,7 +2298,7 @@
             =/  tim
               ;;  (list [@da ^duct])
               =<  q.q  %-  need  %-  need
-              (rof ~ /ames %bx [[our %$ da+now] /debug/timers])
+              (rof [~ ~] /ames %bx [[our %$ da+now] /debug/timers])
             (skim tim |=([@da hen=^duct] ?=([[%ames ?(%pump %recork) *] *] hen)))
           ::
           ::  set timers for flows that should have one set but don't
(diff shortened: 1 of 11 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -5886,33 +5915,15 @@
...
-        %x  (rof ~ /ames nom)
-      ==
-    ::
-        %g
-      %-  en-hunk
-      ?+  kyr  ~
-        %x  (rof ~ /ames nom)
-      ==
-    ==
+    =+  pem=(rof [~ ~] /ames nom(vis %cp))
```
pkova, line 5920: Well almost all of them. Clay still doesn't respect the gang system properly. This shouldn't matter, but we should test thoroughly to make sure.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -5886,33 +5915,15 @@
...
+    ?.  ?=(^ pem)    ~
+    ?.  ?=(^ u.pem)  ~
+    ~|  u.u.pem
+    =+  per=!<([r=dict:clay w=dict:clay] q.u.u.pem)
+    ?.  =([%black ~ ~] rul.r.per)  ~
+    (en-hunk (rof [~ ~] /ames nom))
```
joemfb, line 5926: Yet another nitpick about control flow ....

This public `/fine/hunk/...` endpoint should not be restricted to only public requests -- asking as ourself (`[~ ~]`) should also work. I think the easiest way to do this would be to move it into a standalone `?:  ?=`, above all the private endpoints.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -4116,97 +4116,104 @@
...
+ ::
+      [%cache @ @ ~]
+    =,  server-state.ax
+    ?~  aeon=(slaw %ud i.t.tyl)        [~ ~]
+    ?~  url=(slaw %t i.t.t.tyl)        [~ ~]
+    ?~  entry=(~(get by cache) u.url)  [~ ~]
+    ?.  =(u.aeon aeon.u.entry)         [~ ~]
+    ?~  val=val.u.entry                [~ ~]
+    ?:  &(auth.u.val !=([~ ~] lyc))    ~ 
+    ``noun+!>(u.val)
```
joemfb, line 4215: Same control flow nitpick: both of these public endpoints should be available to fully-privileged requests. I don't see any simple way to do that without just moving them up into separate `?:  ?=` expressions, above all the private endpoints.

Additionally, both of these endpoints should only be defined for a `$care of `%x` and a `$desk` of `%$`. That constraint was lost in the refactoring.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -4116,97 +4116,104 @@
...
+    =<  eauth-url:eauth:authentication
+    (per-server-event [eny *duct now rof] server-state.ax)
+ ::
+      [%cache @ @ ~]
+    =,  server-state.ax
+    ?~  aeon=(slaw %ud i.t.tyl)        [~ ~]
+    ?~  url=(slaw %t i.t.t.tyl)        [~ ~]
+    ?~  entry=(~(get by cache) u.url)  [~ ~]
+    ?.  =(u.aeon aeon.u.entry)         [~ ~]
+    ?~  val=val.u.entry                [~ ~]
```
joemfb, line 4213: This is an old issue, but all three of these cases should block (`~`), not fail (`[~ ~]`).

joemfb (changes requested):
@pkova thanks for taking this on. I've noted a couple issues, the rest looks good:

pkova:
Pushed up the changes requested, still need to test stuff out. I wonder if the CI running forever is because some scry from vere is now blocking...

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -4116,97 +4116,102 @@
...
-        %rejected  ``noun+!>((~(has in rejected.cors-registry) u.origin))
+    ?~  aeon=(slaw %ud i.t.tyl)        [~ ~]
+    ?~  url=(slaw %t i.t.t.tyl)        [~ ~]
+    ?~  entry=(~(get by cache) u.url)  ~
+    ?.  =(u.aeon aeon.u.entry)         ~
+    ?~  val=val.u.entry                ~
+    ?:  &(auth.u.val !=([~ ~] lyc))    ~ 
+    ``noun+!>(u.val)
+  :: private endpoints
+  ?.  ?=([~ ~] lyc)  ~
```
joemfb, line 4150: Everything after this should be de-dented.

PR #6877, opened 2023-12-15
ames: produce correct larval state tag

yosoyubik:
Producing the wrong tag for the larval state could cause problems when updating ames.hoon because we would be regressing back to a previous state that doesn't nest in our latest. Luckily, ames-states 17 and 18 are identical, so we are only going to retrieve (again) our public keys in +molt and then evolve to adult ames, with the latest tag.

(note: still untested on live ships)


belisarius222:
Good catch.  The Ames larval stage looks worse and worse these days.

yosoyubik:
I talked with @pkova about looking into that first thing next year (:

PR #6879, opened 2023-12-21
kiln: check for desk.bill in +poke-rein

### LINKED ISSUE #6527: Docket fails silently.
If an agent is not listed in `desk.bill`, the desk can list itself as `running` and `|rein` will apparently modify the agent's state.

But since it it not in `desk.bill` it cannot run at all.

While this makes sense, `|rein` or `%docket` should notify the developer when the agent is not eligible to run on a given desk, rather than failing silently (or apparently succeeding but failing).

silnem2:
Currently, the `|rein` generator's behavior can cause confusion when it's used to start agents on desks without a `desk.bill` file. Although it will apparently succeed and the desk's `rein` will be updated (as reflected in the `force on:` field in the `+vats` output for the desk), the specified agent won't actually be started because `+goad` skips desks that don't include a `desk.bill` without providing any indication that it has done so. This adds a check for a `desk.bill` on the desk passed to `+poke-rein`, and if there isn't one, `+poke-rein` will no-op and produce an error message.

Resolves #6527

### DIFF
--- a/pkg/arvo/lib/hood/kiln.hoon
+++ b/pkg/arvo/lib/hood/kiln.hoon
@@ -755,7 +755,10 @@
 ::
 ++  poke-rein
   |=  [=desk =rein]
-  abet:(emit %pass /kiln/rein %arvo %c %rein desk rein)
+  ?:  .^(? %cu /(scot %p our)/[desk]/(scot %da now)/desk/bill)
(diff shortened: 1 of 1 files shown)


sigilante:
This is reasonable. I have had this exact scenario confuse me before.

PR #6881, opened 2023-12-28
add normal-looking aliases to standard library gates

jalehman:
The four-letter names are clever, but in cases where there are well-defined names for concepts in the rest of programming, we should use them.

TODO:
- [ ] Completion pass — what else should be included/excluded?
- [ ] Bikeshed
- [ ] Add documentation of the caliber found in [sequent](https://github.com/jackfoxy/sequent/blob/master/desk/lib/seq.hoon#L137)
- [ ] More bikeshedding

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -690,7 +690,7 @@
   ?~  b  ~
   $(b t.b, a (dec a))
 ::
-++  drop  slag
+++  drop-from  slag
 ::
 ++  snag                                                ::  index
   ~/  %snag
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -467,6 +467,7 @@
   |*  [a=(list) b=*]
   (weld a ^+(a [b]~))
 ::
+++  append  snoc
 ::  +lure: List pURE
 ++  lure
   |*  a=*
@@ -513,11 +514,15 @@
   ?~  a  b
   $(a t.a, b [i.a b])
 ::
+++  reverse  flop
+::
 ++  gulf                                                ::  range inclusive
   |=  [a=@ b=@]
   ?>  (lte a b)
   |-  ^-  (list @)
   ?:(=(a +(b)) ~ [a $(a +(a))])
+::
+++  range  gulf
 ::
 ++  homo                                                ::  homogenize
   |*  a=(list)
@@ -553,6 +558,8 @@
   ?~  a  b
   $(a t.a, b +(b))
 ::
+++  length  lent
+::
 ++  levy
   ~/  %levy                                             ::  all of
   |*  [a=(list) b=$-(* ?)]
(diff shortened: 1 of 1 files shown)


tacryt-socryp:
Fun bikeshed:
Organizing aliases in one place, perhaps in its own file, or at least core, indexed by data structure would be nice as a way for people, particularly new people, to quickly find what they're looking for.

For instance, though perhaps the name shadowing is not ideal:
```
++  list
  |%
  ++  append  snoc
  ++  transform  turn
  ...
  --
::
++  map
  |%
  ...
  --
::
++  set
  |%
  ...
  --
```

jalehman:
This would mean that, assuming your example, they'd be accessed as `append:list`, right?

tacryt-socryp:
Yes, which is very explicit and predictable (good for newcomers). The other advantage is that all the "nicely named" functions would be together, which would make them easier to find and use for newcomers. I imagine =, and =* could be sufficient for making them more ergonomic in practice.

jalehman:
If I had `map` or `set` as arms in that core, would they not conflict with the molds of the same name?

tacryt-socryp:
Yes, hence my comment that the name shadowing in my suggestion isn't ideal... I just can't help but think there's a better naming scheme for things than having no outer core name, like the list utilities, or "by", "in", etc that have no relation to what they do.

jalehman:
Yeah, I think you're on to something here. Making naming more regular across the board at the expense of verbosity seems like a net positive. A dumb way to do it:

```
++  liblist
  |%
  ++  append  snoc
  ++  transform  turn
  ...
  --
::
++  libmap
  |%
  ...
  --
::
++  libset
  |%
  ...
  --
  ```

**Comment by tacryt-socryp** [comment]
One idea... these are all "collections" of things, and potentially could have their type in them as well.

```
++  collections
  |%
  ++  list
    |%
    ++  mold  ^list
    ++  transform  turn
    ...
    --
  ::
  ++  map
    |%
    ++  mold  ^map
    ...
    --
  ...
  --
```

Usage:
```
=,  collections
=/  a=(mold:list @)  (gulf 0 9)
(transform:list a inc)
```

**Comment by belisarius222** [comment]
I've thought for a while we might want to structure the stdlib like this:
```hoon
++  list
  =<  form
  |%
  ++  form  |$  [a]  $?(~ [i=a t=$])
  ++  flop  |*(...)
  ++  turn  |*(...)
  ...
  --
```

This way you can still use `list` as a mold builder without qualifying it, but list-related functions would be inside the `list` core (they'd be in the context of the `form:list` gate).  You'd call `turn:list` unless you first had a `=,  list` to allow unqualified `turn` or a `=*  l list` to allow `turn:l`, as @tacryt-socryp mentioned.

tacryt-socryp:
That sounds great Ted

drbeefsupreme:
@belisarius222 so this is an old, closed issue but I'm not sure where else to comment this. I used your types here extensively for the transaction engine for Nockchain. Not sure if you had a name for it yet, but I've settled on `B-types`, with B for "Blackman" and/or "alternate". There's a PR going in that mention this in the comment at the top. I hope it catches on.

https://github.com/zorp-corp/nockchain/blob/master/hoon/common/tx-engine.hoon

PR #6883, opened 2023-12-30
restore rut:by as an alias of urn:by

jalehman:
## Context

https://github.com/urbit/urbit/pull/6640 removed `rut:by` because it was a duplicate of `urn:by`. 

## Motivation

I think we should break our habit of breaking public interfaces whenever possible. We don't know that there isn't software out there that depends on `rut:by`, and by removing it we force some developer to go and update all their code for no real reason.

This PR makes `rut:by` into an alias of `urn:by`. That means we don't have more surface area to introduce bugs (the reason deleting code is considered "good"), and no one's interface breaks. I've marked it as deprecated in a comment — further in the future we can remove it after giving anyone that relies on this ample time to migrate away from it.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -1588,6 +1588,8 @@
     ?~  a  a
     [n=[p=p.n.a q=(b q.n.a)] l=$(a l.a) r=$(a r.a)]
   ::
+  ++  rut  urn                                          :: ~2023.12.30: deprecated; use urn instead
+  ::
   ++  tap                                               ::  listify pairs
     =<  $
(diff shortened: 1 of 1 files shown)


ashelkovnykov:
I'm weakly opposed to this for the following reasons:
1. We have a very small number of devs, who are of high quality and tuned-in to changes. This won't be a major inconvenience, any more than most kelvin updates are right now.
2. This is actually derivative of the two true issues:
    a. We have no backwards compatibility
    b. We have no user-space "standard library"; our current standard library is for kernel-space

#6881 is on the path to fixing `2b`; @joemfb and I are starting work on what should become a UIP for addressing `2a`.

PR #6888, opened 2024-01-15
(Userspace trawl) Create ++geq/++leq

sigilante:
LaTeX-style names for `++gte`/`++lte`.


sigilante:
Opening for comment.  We actually don't want this as a `/lib`, if we want it then we want them in `/sys/hoon` next to their sources.

PR #6892, opened 2024-01-17
landscape: add public-url-base to s3 config

mrdomino:
Needed for Cloudflare R2 support.

### DIFF
--- a/pkg/landscape/app/s3-store.hoon
+++ b/pkg/landscape/app/s3-store.hoon
@@ -10,13 +10,15 @@
 +$  versioned-state
   $%  state-zero
       state-one
+      state-two
   ==
 ::
 +$  state-zero  [%0 =credentials:zero:past =configuration:zero:past]
-+$  state-one   [%1 =credentials =configuration]
++$  state-one   [%1 =credentials:one:past =configuration:one:past]
++$  state-two   [%2 =credentials =configuration]
 --
 ::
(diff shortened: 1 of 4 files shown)


mrdomino:
@Fang- as discussed.

This is not currently building on my fakezod; it breaks on the `~%`, and when I comment that out, this is the stack trace:

```
[%error-building /app/s3-store/hoon]
/app/s3-store/hoon::[21 1].[157 3]>
/app/s3-store/hoon::[22 1].[157 3]>
/app/s3-store/hoon::[24 1].[157 3]>
/app/s3-store/hoon::[25 1].[157 3]>
/app/s3-store/hoon::[26 1].[157 3]>
/app/s3-store/hoon::[28 1].[157 3]>
/app/s3-store/hoon::[141 15].[141 27]>
- need
[ %2
  credentials=[endpoint=@t access-key-id=@t secret-access-key=@t]
  configuration=[buckets=nlr(@t) current-bucket=@t region=@t public-url-base=@t]
]
- have
[ %1
  credentials=[endpoint=@t access-key-id=@t secret-access-key=@t]
  configuration=[buckets=nlr(@t) current-bucket=@t region=@t]
]
nest-fail
```

Looks like `on-leave` is wrong? But I can't see what to do to fix it. I'm comparing against 02549975334bd001c4e232bce68052ea17288e76 for reference.

PR #6895, opened 2024-01-23
userspace trawl: add parser aliases

tinnus-napbus:
A draft of one approach to aliasing parsers to make them a bit easier.

What do you think of this approach @sigilante ?

also I'm not sure where the best place would be for these - I stuck them in a new lib just to put them somewhere


sigilante:
Yes, these look good.  Maybe a `++whitespace` too, around `gaw` or `gay`.

PR #6896, opened 2024-01-24
test thread: now prints errors/failures in red instead of gray

wispem-wantex:
When something goes wrong it should print in red.  It looks better and adds visual contrast (drawing your eye to the important bit).

Before vs after (test assertion failed):
image
image

Before vs after (build failure):
image
image

### DIFF
--- a/pkg/arvo/sys/vane/clay.hoon
+++ b/pkg/arvo/sys/vane/clay.hoon
@@ -4552,7 +4552,7 @@
       |^  =/  res  (mule |.(read))
           ?:  ?=(%& -.res)  p.res
           %.  [[~ ~] ..park]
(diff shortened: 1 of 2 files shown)


pkova:
Let's rip the clay changes out of this one. We probably want highlighted errors for clay too in the future but it's going to be strange if this is the only error case that clay prints out with red text.

wispem-wantex:
I considered that; what I'm wondering is how to find all error cases.  I only did this one because I ran into it a lot.  Is there an easier way to find all the places where errors are printed besides running into them one by one, saying "Hey, that's an error, it should be in red", and fixing it?

pkova (approved):

PR #6898, opened 2024-01-25
clay: do not return false conversion gate from build-cast

pkova:
@Fang- noticed today that sometimes the clay `%cf` scry returns a mark conversion gate that's complete nonsense:

```hoon
=x .^($-(* *) %cf /=base=/hoon/json)
(x *hoon)
[1.887.072.378 0]
```

`/mar/json` or `/mar/hoon` do not contain the proper arms, yet this scry still works. The gate we get returns the bunt of `hoon` in this case (which is definitely not `json`).

The reason why this happens can be gleaned from the following snippet:

```hoon
=c |%  ++  grab  |%  ++  hello  1  --  --
hoon:grab:c
<1.fji [* <236.trs 51.enr 139.uat 33.rnj 1.pnw %139>]>
```

When trying the `+grab` arm in the mark we don't use `slob` to check whether the arm actually exists, we just trust the tisgal to get us what we want. This PR fixes the bug.

I also decided to remove some unused `+jump` mark stuff I've never seen before.

### DIFF
--- a/pkg/arvo/app/herm.hoon
+++ b/pkg/arvo/app/herm.hoon
@@ -5,7 +5,6 @@
 ::  keep relevant mark conversions in cache for performance
 ::
 /$  blit-to-json  %blit  %json
-/$  json-to-blit  %json  %blit
 /$  json-to-task  %json  %herm-task
 ::
 =,  jael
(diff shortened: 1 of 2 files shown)


pkg/arvo/app/herm.hoon
```diff
@@ -5,7 +5,6 @@
 ::  keep relevant mark conversions in cache for performance
 ::
 /$  blit-to-json  %blit  %json
-/$  json-to-blit  %json  %blit
```
pkova, line 8: The mark that never was...

pkova:
I restored `+jump` by popular demand.

belisarius222 (approved):

PR #6909, opened 2024-02-08
eyre: dynamic responses for empty cache entries

### LINKED ISSUE #6702: eyre: lack of ability to invalidate eyre cache results in broken ship state
**Describe the bug**
%blog uses the eyre cache, which is being used here to serve static responses. You're able to use this to easily add cache entries for / or were at the time this was done, I have not tried to replicate with the latest version. Presumably this would also extend to other important urls such as /apps This would render grid and landscape essentially inaccessible. There is no mechanism for removing cache entries other than by setting an entry with the same path to sig. Doing this did not repair the broken redirection so I was forced to hack in to bunting the cache as a means to clear it, which should have completely resolved the problem. Vere appears to be keeping this cache and sending the responses and so even this drastic reset of the cache did not fix the problem until the runtime was restarted, a fact I only discovered upon accidentally killing the process while trying to look at what version of vere it was using. It appears that this is a result of there being no actual method for telling the cache you no longer want it to handle a given url, the actual request here never getting all the way to eyre and docket making debugging this difficult.

While %blog could probably protect against this, it should not be simple to get into this state. Using the eyre cache for this is arguably an abuse, but if this is intended behavior there needs to be a way to "uncache" something without restarting the runtime.

Fang-:
If the cache entry for any given url was cleared, eyre would recognize that an entry was bound at some point in the past, and serve a 404, instead of falling back to the normal, cacheless, dynamic response.

Here, we make eyre serve from the cache only if there is a full cache entry for the url, and have it proceed into normal request handling logic otherwise.

This way, setting a cached response for some url no longer means giving up dynamic responses on that url for the rest of time.

Recent versions of vere exhibit the same behavior. As such, in practice, this code won't get hit without a similarly-patched version of the runtime. For that, see the sister PR, urbit/vere#603.

Resolves #6702.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -907,11 +907,12 @@
     ::
     ?:  =('/_~_/' (end [3 5] url.request))
       (handle-http-scry authenticated request)
-    ::  handle requests to the cache
+    ::  handle requests to the cache, if a non-empty entry exists
     ::
-    =/  entry  (~(get by cache.state) url.request)
-    ?:  &(?=(^ entry) ?=(%'GET' method.request))
-      (handle-cache-req authenticated request val.u.entry)
+    =/  cached=(unit [aeon=@ud val=(unit cache-entry)])
+      (~(get by cache.state) url.request)
+    ?:  &(?=([~ @ ^] cached) ?=(%'GET' method.request))
+      (handle-cache-req authenticated request u.val.u.cached)
     ::
     ?-    -.action
         %gen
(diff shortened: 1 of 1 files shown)


belisarius222 (approved):
LGTM.  It would be good to test this before release by binding a cache entry, checking that it works, unbinding the entry and handling the request in an agent, checking that works, and then rebinding the cache entry to a new value and making sure the new value is served.

Fang-:
fwiw I ran that test locally, and have been using this version of eyre (& vere) for my local testing of an upcoming feature that relies on the cache. But having a second vote of confidence would be good.

pkova (approved):

PR #6917, opened 2024-02-19
tests: make /ted/test compile again

Fang-:
It appears that bb8ba24 merged changes to /ted/test from #6790 and #6896 incorrectly, as the thread no longer compiled.

Because the former PR essentially included the change made to the latter (printing with slog priority 3 in case of failure), the minimum viable fix is taking the contents of /ted/test from #6790 and using those as-is.

However, that had hidden success prints behind a "quiet" flag, and provided no way of configuring that from outside the thread. So additionally, we expand the ""vase parsing"" performed on the thread's argument, such that an optional _first_ parameter of a quiet=? may be passed.

(Should the merge commit have had green CI checkmark in light of this? Presumably not, but I'll consider that out of scope for this PR.)

(I also have opinions about the output style change inherited from #6790 here, but right now I just want to run my tests, man...)

### DIFF
--- a/pkg/arvo/ted/test.hoon
+++ b/pkg/arvo/ted/test.hoon
@@ -10,7 +10,7 @@
 +$  test       [=beam func=test-func]
 +$  test-arm   [name=term func=test-func]
 +$  test-func  (trap tang)
-+$  args       quiet=_&
++$  args       quiet=?
 --
 =>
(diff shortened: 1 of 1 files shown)


pkova (approved):
wtf, see also #6918

Issue #6920, opened 2024-02-22
hoon: sort runs out of memory


yosoyubik:
```
> =+  (gulf 0 11.384)  =+  (sort - lth)  (lent -)
11.385
```

but

```
> =+  (gulf 0 11.385)  =+  (sort - lth)  (lent -)
(recover: dig: meme)
```

This looks like something wrong with the sort jet.

Another data point, ++skim works fine:

```
> =+  (gulf 0 111.386)  =+  %+  skim  -  |=  a=@  &  (lent -)
111.387
```

Looked at this one together with @pkova

PR #6921, opened 2024-02-22
jael: do not sort remote gifts

pkova:
Last week `~mister-dister-dozzod-dozzod` was corrupted and needs to breach. This is a `~zod` moon with around 30k subscribers and running `|moon-breach` on `~zod` was running out of memory even with an 8gb loom.

The issue is somewhere in our `sort` implementation, see #6920 for a reproduction.

Luckily we can just sidestep this issue in `+public-keys-give`, since we don't need to sort remote `%boons`. This has been tested on a local copy of `~zod` and breaches the moon easily with a 2gb loom.


joemfb (approved):
beautiful

PR #6929, opened 2024-03-05
eyre: do not store localhost as eauth-url

pkova:
Tlon hosting has a login health check that regularly logs into the ship using the `+code`. This login happens from a sidecar and therefore thrashes the eauth-url with localhost constantly. This causes ships on Tlon hosting to have unreliable eauth currently.

Disallowing localhost as the eauth-url seems like the simplest solution to me.

### DIFF
--- a/pkg/arvo/sys/vane/eyre.hoon
+++ b/pkg/arvo/sys/vane/eyre.hoon
@@ -1306,8 +1306,13 @@
         o(session-id session.fex)
       ::  store the hostname used for this login, later reuse it for eauth
       ::
-      =?  endpoint.auth.state  
-        &(?=(^ host) !=('localhost' (fall (rush u.host host-sans-port) '')))
+      =?  endpoint.auth.state
+          ::  avoid overwriting public domains with localhost
+          ::
+          ?&  ?=(^ host)
+          ?|  ?=(~ auth.endpoint.auth.state)
+              !=('localhost' (fall (rush u.host host-sans-port) '')))
+          ==  ==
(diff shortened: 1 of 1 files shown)


tinnus-napbus:
imo this logic should only be applied if `auth.endpoint.auth.state` is non-null cos you still wanna be able to test eauth locally.

Also it should probably short-circuit on a True equality check between `host` and `u.auth.endpoint.auth.state` cos parsers are slow & this will be run on every request despite 99.9% of them having the same host

Fang- (commented):
> Also it should probably short-circuit on a True equality check between `(cat 3 ?:(secure 'https://' 'http://') u.host)` and `u.auth.endpoint.auth.state` cos parsers are slow & this will be run on every request despite 99.9% of them having the same host

Except the code here only runs for every _login attempt_, not every request in general. It's fine.

I do agree that we should be more "eager" in cases where we don't yet have a known public url yet at all. Even in cases where this is undesired, at least it'll be clearly wrong ("i got redirected to localhost!") instead of subtly wrong ("it didn't redirect me at all!").

(This eagerness is probably redundant in practice, considering we [already](https://github.com/urbit/urbit/blob/f0ae78db55362a306ee3cedd78f6e7c3b228f219/pkg/arvo/sys/vane/eyre.hoon#L824-L835) eagerly accept the endpoint from _any_ authenticated request in that scenario, but sharing that intent/behavior here is more consistent.)

Matching suggestion below.

Fang- (approved):
This should be good!

PR #6932, opened 2024-03-11
[WIP] mesa: add Directed Messaging protocol

yosoyubik:
TODO:
- vere [ ]  
  - [x] %memo cache growing out of control and causing crashes
    - seems fixed?
    - [x] test again for bigger messages
    - [ ] %memo cache checks taking longer than expected after crash and reboot
  - [ ] look `dropped packet (authentication)`
    - happens sometimes, gets fixed on re-send 
  - [ ] make forwarding real 
    - [x] expire routes
    - [x] check forward lanes for migrated sponsees, on request that use %ames
  - [x] _real_ jumbo packets
  - [x] merge latest lc/chacha
  - [x] new lss streaming algorithm jets
  - [ ] prime scry cache on first %poke fragment send
    - XX  
  - [x] unified _fake_ driver (ames/mesa.c) 
     -  [ ] make the unified driver real/ polish
  - [x] investigate STUN not working 
- arvo [ ] 
  - [x] initial contact flow, when lane is missing
    - [x] MVP: send to the peer's galaxy lane—route everything through it
    - [ ] make it flexible to send to the sponsorship chain
    - [ ] replicate alien key retrieval from `++send-blob`
  - [ ] lane management/routing
    - fake, using %dear 
    - [x] Fix; port in %ames is different than the "real" one, on a ship not behind a NAT. %ames works fine, but after migrating the peer and using |mesa, we need to manually put the lane in, or we can't communicate
      - [x] test
    - [ ] keeping multiple lanes and sending to all is also messing something up, as in -hi not working, but keeping just the last one we've heard, makes it work again.
    - [x] inject direct/indirect lane for pokes 
      - [x] inject hopcount into %arvo
        - useful for trace logging?
        - XX [ ] replace it with just direct/indirect
    - [x] for %page, if `?=(~ .next)` lane is direct, otherwise add all .next to route.peer
    - [ ] lane eviction
      - currently last 5 lanes are kept.
      - if we hear the same late twice, skip it—it can be added later again if we hear another different lane in between.   
  - [x] fix migrating .keens into +peeks
    - we need to have updated chums.ames-state before we start migrating peeks  
  - [x] add %mesa support in %clay—currently done in the `yu/rate` branch 
  - [x] look into pe-keen for per-peer migrating check
  - [x] look into +peeking for a file that we lack the mark of, causing a block in the queue—the requested crashes in `ev-make-pact` the responder in `[=wire resp=$>(?(%boon %noon) gift)]`—check what other gifts %clay gives to %ames
  - [ ] look into crash in %lull for lane when hearing a `=task  ((harden task) wrapped-task)` 
  - [x] test new lss streaming algorithm
  - [x] scry unification
    - XX how to make test work; we have to move the scry handler into the event-core to avoid calling +rof and do +ev-peek instead
  - [x] weird +scry cancel
    -  seen on a live moon: a scry to the parent planet of a moon, shows up in the path as "~nec"
  - [ ] QoS
 handled properly
  - [ ] encrypted payload path in poke:pact
  - [ ] remove crypto core  
  - [ ] XX remove typed paths
    - [ ] or add missing jets 
  - [ ] trace logs
  - [ ] ~debug dashboard support
  - [ ] test comets
  
- Open questions:
  - [ ] Scrying on ++load
  - [ ] Larval stage—currently the unix duct in ames-state is the only data dependency and the latest version has the duct hardcoded
  - Boot: directly into the latest protocol version, how to do protocol negotiation?
    - [ ] Issue reads?  currently not possible, see %alien +peeks
    - [ ] Try all protocols?
    - [x] currently: boot into %ames, migrate peers manually. 
    - ...    

DONE:

- [ ] vere
  - [x] scry list of fragment using `/hunk/lop=@/len=@' namespace 
  - [x] inject response page into %arvo 
  - inject completed, serialized, and encrypted `%page` using the `%mess-ser` task
    - ~~(it currently uses `%mess %page`)~~ see task bellow
    - XX better to no proliferate task for each possible entry point (one fragment, multi-fragment, encrypted, public... etc)
  - [x] use one entry point for both packet/message %page responses
  - [x] fix segfault caused by incorrect uv_timer_init
    - see https://github.com/urbit/vere/commit/1113971a18327bc2a57726bca7af80c813b71da3  
    - 
- [ ] arvo 
  - [x] store private key in state
  - [x] `|hi ~zod` ($poke via %chum)
  - [x] `-meen ~zod /c/x/1/sys/kelvin`  ($peek via %publ)
  - [x] `-meen ~zod /c/x/1/sys/kelvin [~ key=1 sec=0x0]` ($peek via %shut)
  - [x] `hunk` namespace for batches of fragments 
  - [x] fix wrong "fragment" inject for "fake" jumbo packets
  - [x] meet %alien(s)
  - [x] comet attestation 
    - Instead of sending a special packet (i.e. open-packet in old %ames) we just
read the attestation proof from the comet's namespace. There needs to be
a special case for %alien comets that we pretend to know, so we can add the
path into the pit to track it—this is a bit of a hack and might be fragile because
of the way we need to check that we have bunted the state of the chum comet
  - [x] refactor flow paths (remove `+dire=?(%for %bak)`, use plea/ack-plea boon/ack-boon)

  - [x] rethink %mesa-ser task, %mesa-response gift
  - [ ] test +peek cancel (%yawn/%wham)
  - [x] fix %alien migrationg (fix: re-retrieving public-keys on %prod if %alien or missing_
    - seen on a live moon: ~marzod keys were never received, so we have to manually ask %jaels for them again to switch from %alien to %known, otherwise we can't contact the parent planet.
      - [x] seems fixed—moves were not emitted, although the state not being update to %known from %alien is weird...
        - this was either a stale solid pill, or something previously fixed. now there's only one wire when requesting the keys from %jael, and i haven't seen this again, with the latest mesa-solid-live.pill. 
  - [x] fix state from .peers not being deleted after migration  
  - [x] fix missing /pump timers (at least on first attemp; %prod puts them back)
  - [x] refactor per=[=ship sat=ship-state] in mesa core
    - tasks that include a ship are called in |mesa with the peer already in the door's sample
  - [x] %alien +peeks
    - if we don't have the keys for ships that read into our namespace:
      - how to retrieve their keys so we can respond?
        -  [x] scry from %jael; revisit in the future
        - Every +peek to an alien ship gets preceded by an advisory +poke (e.g. please, get my keys)
        - Make %vere smart so it can detect that aliens are trying to read into our namespace, gather who they are, and send a list of ships for gathering keys that we need.
        - ...
  - [x] For a 6-fragment message, I'm seeing a failure in decrypting the path (%heer crash), then after proding and retrying, the message comes is, |hi succesful, poke acknowleged, entries from the .pit (for both) removed... and then bail: oops on the receiver of the |hi
  - subscriptions
    - [x] %cork
    - [x] %watch
    - [x] %leave
    - [x] %boon
      - [ ] XX repeated, rapid %facts from a %gall agent block
        - [ ] seems fixed? check again 
    - [x] fix truncated %corked flow response
      - see https://github.com/urbit/urbit/pull/6932#issuecomment-2039775496 
    - [x] clog machinery
      - [ ] remove `mem` from the state parameters? 
    - [ ] %naxplanation flow
      - [x] app crash (e.g. +on-leave)
      - [ ] vane (%g/%m) crash
      - reuse `%near` task? 
      - check closing bones   
  - remote scry
    - [x] %keen (both encrypted and public)
    - [x] %yawn
    - [x] %wham 
  - Full end-to-end test
    - Groups
      - [x] install groups
      - [x] search for host
        - [x] search again 
          - XX  seeing `subscribe wire not unique` (for `/gangs/index/~fen`) 
      - [x] join a group
      - [x] receive backlog
      - [x] send a message (as subscriber)
      - [x] send a message (as host)
      - [x] update subscriber
      - [x] leave group
      - [x] kick subscriber
      - [x] rejoin
        - [x] seems to be working, but only sometimes? (more tests needed) 
        - [x] leaving and kicking several times, and rejoining seems to be working, so the things listed bellowes, as far as i can see are %eyre/%gall subscription isssues 
          - [ ] ~~can't find ~host~~ host shows up eventually
          - [ ] seeing `spider crash`
          - [ ] seeing `subscribe wire not unique` (for `/gangs/index/~fen`)
          - [ ] general slowness searching for a group host
          - [ ] ...
      - [ ] load latest version of Groups
        - rejoining a group after leaving/kicking is slow  
    - [x] send DM 
    - [ ] Suspend/Nuke/Revive agents
    - [ ] Install other Apps
      - [x] %file-share
    - [ ] ...
  - system taks
    - [x] %tame
    - [x] %dear
    - [ ] %kroc (probably not)
    - [x] %heed
    - [x] %jilt
    - [x] %stun
    - [x] %prod
    - [x] %sift
    - [x] %snub
    - [x] %spew
    - [ ] %cong
      - [ ] remove `mem` from `$axle`? 
    - [ ] %stir (what to put in here?)
    - [ ] %trim
  - system gifts 
    - [x] %turf
    - [x] %nail
    - [x] %stub
    - ~~%near~~ (only in %ames, captured by the used wire in +take)
    - [x] %saxo
    - [ ] ...
  - Migration
    - 
    - At first, %ames is updated with a:
      - refactor to accommodate the two protocol logic handling: %ames and (wip) %mesa
      - [x] new type of %plea targeting the `$%` vane (currently used for pre-cork ships)
        -  `plea/[vane=%$ path=/mesa payload=%ahoy]`, on a wire controlled by %ames.
      - [x] new intermediate core, for per-peer checks on protocol switching 
    - when on-load is called, with the new version of the vane, %ames (old protocol handling core) will be used by default—nobody will speak the new protocol automatically
    - [x] a generator will allow per-peer attempts to switch to the new protocol by using old-ames to send an `%ahoy` `%plea`. 
      - if the receiver has _not_ updated, it will crash, and the sender will keep trying.
      - if the receiver _has_ updated, we will flag this peer as "speaking the new protocol".
    - [x] any calls into %ames to send a request will use a per-peer check to confirm if we know that the receiver speaks the new protocol
      - if not, the %ames core will be used
        - XX send also the %ahoy %plea to be notified as soon as they switch? 
      - if yes, the %mesa core will be used and all flows will be migrated to the new protocol
    - Migration flow:
      - (pre: only "quiet" flows (i.e. not live) will be migrated; a per-peer timer will check if there are any flows that were live but have achieved quiescence and can be migrated)
      - (XX not true anymore, we migrate all flows, also partial/live naxplanations flows, since we know how to distinguish those based on current != next and lack of state for that message in the packet-pump)
      - A .ship send an %ahoy plea to .peer.
      - .peer receives the %plea, if it doesn't speak the new protocol it will crash (the plea will be resend), otherwise it will migrate any non-live flows and ack the plea.
      - .ship receives the %ack for the %ahoy plea and then migrates any non-live flows.
      - if there are no flows that are still live, .ship and .peer will remove each other from peers.ames-state and eveything will live not at chums.ames-state
      - live flows will be kept in peers.ames-state, and on a per-peer timer we will check if those flows have achieved quiescence.
      - If, after migrating flows, the ack for the %ahoy plea gets lost, the sender of the %ahoy plea could send something, for their point of view, "old", as in, still in peers.ames-state, but which the receiver will have in chums.ames-state. In order to avoid having two bones in two places, we always need to check if something we hear from the old protocol has been migrated, and in that case drop it, since eventually we will receive the %ack for the %ahoy plea, and then migrate the flow to the new protocol.
        -  XX now we have introduced a problem where we have a live flow that won't be migrated but that will never be acked.

- [x] merge last jb/dire [commit](https://github.com/urbit/urbit/commit/7451c4ccc7e8236500b49998b8358399c0480ba3)
- [ ] ... 
- [ ] merge [jb/dire](https://github.com/urbit/urbit/tree/jb/dire) branch to keep git history


yosoyubik:
We ran into an issue decrypting a specific path with the `+crtc:aes` algorithm (the decrypted path doesn't match the one we encrypt). I haven't found any other reproductions with paths of different length... Using chacha instead of AES works fine.

```
> =key `@`0w5ub.Wig2P.yo031.tZgj8.314QW.yDv56.Pe-Hp.fv2Zo.F~jQM 
> =path /m/x/1//flow/6/~fen/poke/~dev/bak/1

:: Encrypting

[%pat 0x58.d075.b0b1.3c07.b32b.23f7.c072.b5b7.b87c.0773.2b33.f7c0.76a0.f76f.6c66.f80c.e341.f1c1.dbc1]
[%tag 0x1bd1.df38.9d35.919d.e8d3.8678.9662.8003]
[%cyf 0x9e75.a089.0786.1021.bf82.b86d.ecc4.5d6e.744f.f313.5cf7.01bc.880c.5a6f.96a3.3e27.96d8.26af]

> `@ux`(seal-path:crypt:mesa key path)
0x1.3ceb.4112.0f0c.2043.7f05.70db.d988.badc.e89f.e626.b9ee.0379.1018.b4df.2d46.7c4f.2db0.4d5e.4100.37a3.be71.3a6b.233b.d1a7.0cf1.2cc5.0007.ec01


:: Decrypting

[%sealed 0x1.3ceb.4112.0f0c.2043.7f05.70db.d988.badc.e89f.e626.b9ee.0379.1018.b4df.2d46.7c4f.2db0.4d5e.4100.37a3.be71.3a6b.233b.d1a7.0cf1.2cc5.0007.ec01]
[%tag 0x1bd1.df38.9d35.919d.e8d3.8678.9662.8003]
[%cyf 0x9e75.a089.0786.1021.bf82.b86d.ecc4.5d6e.744f.f313.5cf7.01bc.880c.5a6f.96a3.3e27.96d8.26af]
[%pat 0xc63b.a099.3fbd.9182.b51e.cd15.f39d.2e8b.663c.cfcb.7c5c.36cb.9473.3959.9fcd.91fa.f0bf.3f52]
[%met-pat 36]
[%keyed 0xdbe5.aab6.9cd7.1199.1520.2d2b.4c8d.5fa4]
[%cue 0]
[%const-cmp 84.967.810.778.138.679.633.494.167.652.306.067.551]


> (open-path:crypt:mesa key enc-path)

/sys/vane/mesa/hoon:<[154 9].[164 27]>
/sys/vane/mesa/hoon:<[155 9].[164 27]>
/sys/vane/mesa/hoon:<[156 9].[164 27]>
/sys/vane/mesa/hoon:<[157 9].[164 27]>
/sys/vane/mesa/hoon:<[158 9].[164 27]>
/sys/vane/mesa/hoon:<[159 9].[164 27]>
/sys/vane/mesa/hoon:<[160 9].[164 27]>
/sys/vane/mesa/hoon:<[161 9].[164 27]>
/sys/vane/mesa/hoon:<[162 9].[164 27]>
/sys/vane/mesa/hoon:<[163 9].[164 27]>
```
cc: @joemfb @lukechampine 

<details>
  <summary>Expand code</summary>

    ```
    ++  encrypt
      |=  [key=@uxI iv=@ msg=@]
      ^+  msg
      (~(en ctrc:aes:crypto key 7 (met 3 msg) iv) msg) :: TODO: chacha8
    ++  decrypt  encrypt
    ::
    ++  seal-path
      |=  [key=@uxI =path]
      ^-  @
      =/  pat  (jam path)
      ~&  pat/`@ux`pat
      =/  tag  ((keyed:blake3:blake:crypto 32^key) 16 (met 3 pat)^pat)
      ~&  tag/`@ux`tag
      =/  cyf  (encrypt (mix key tag) tag pat)
      ~&  cyf/`@ux`cyf
      (jam [tag cyf])
    ::
    ++  open-path
      |=  [key=@uxI sealed=@]
      ^-  path
      =+  ;;([tag=@ cyf=@] (cue sealed))
      ~&  tag/`@ux`tag
      ~&  cyf/`@ux`cyf
      =/  pat  (decrypt (mix key tag) tag cyf)
      ~&  pat/`@ux`pat
      ~&  met-pat/(met 3 pat)
      ~&  keyed/((keyed:blake3:blake:crypto 32^key) 16 (met 3 pat)^pat)
      ~&  cue/(cue pat)
      ?>  (const-cmp tag ((keyed:blake3:blake:crypto 32^key) 16 (met 3 pat)^pat))  :: XX crash!
      ;;(path (cue pat))
    ::
    ++  const-cmp
      |=  [a=@ b=@]
      ^-  ?
      ~&  const-cmp/(~(dif fe 7) a b)
      =(0 (~(dif fe 7) a b))  :: XX jet for constant-time
    ::
    ```
</details>

lukechampine:
Ok, I investigated this, and the root cause was using `@` instead of `byts`. What happened was that, for that particular key and path, the first (last?) byte of the keystream and the plaintext happened to match exactly; thus, when they were xor'd, the result had a leading (trailing?) zero, causing `met 3` returned 1 less than it should have. So when this ciphertext was decrypted, it was missing a byte.

I've rectified this in `jb/dire`, so you can copy from there and it should work fine. Of course, as you observed, it already worked with chacha *anyway*, because the bug is sensitive to exactly which bytes are produced by the keystream. So we can't know for sure that it's fixed without exhaustive testing. But I'm pretty sure. :P

As you might suspect, this same bug also affects encryption of the message bodies. For now, I've cheated by using `met 3` in a few places, but really we should be using the actual payload length. AFAIK that isn't currently exposed anywhere, so either it needs to be added to the packet format, or jam'd alongside the message data and cue'd out before decryption. 🤔 @joemfb

belisarius222:
Note for a future security audit: make sure to test every input and output of our crypto functions for correct handling of leading and trailing zeros.

lukechampine:
As part of the Great Crypto Reform I will be enforcing the use of `byts` and consistent endianness across all crypto arms 😈

yosoyubik:
@lukechampine it seems that that there's a issue with chacha's decryption algorithm. Here's a repro: 

```
[ %ser 4.114.545.892.744.903.819.886.956.731.883.009]

[ %key 39.598.418.646.500.682.561.058.112.490.997.051.307.621.673.044.228.499.795.761.268.322.630.494.928.176]

[ %iv 21.693.197.627.403.879.074.728.363.760.321.977.863.565.466.178.964.571.449.628.575.320.341.700.557.702.648.031.096.773.999.644.078.014.974.484.003.095.594.438.785.271.699.658.539.203.296.257]

=/  cyr  (encrypt:crypt key iv (met 3 ser)^ser)

[ %cyr wid=14 dat=15.139.551.684.194.250.582.242.371.041.754]

=/  decrypt  (decrypt:crypt `@`key iv (met 3 dat.cyr)^dat.cyr)

[%decrypt wid=13 dat=17.499.152.807.266.394.249.611.972.107.777]

~&  ;;([%message %gone] (cue ser))
[%message %gone]

~&  ;;([%message %gon] (cue dat.decrypt))
[%message %gon]
```

joemfb:
@yosoyubik this is the same trailing-zero issue as above. the fix will be to properly preserve/convey the ciphertext length.

My preferred approach is to have a constant `0x1` in the most significant byte, something like this in your example:

```
=/  cyf  (can 3 [wid dat]:cyr [1 0x1] ~)
...
=/  decrypt
  =/  wid  (dec (met 3 cyf))
  ?>  =(0x1 (cut 3 [wid 1] cyf))
  (decrypt:crypt `@`key iv wid^cyf)
```

lukechampine:
Could we make `data:pact` take a `byts` instead of a `@`? Then we could pass the result of `chacha` to it directly.
`+en:data` is already calling `(met 3 dat)` internally anyway. I've seen enough crypto arms at this point to interpret that as a sign that the gate should just take a `byts`.

tests/sys/vane/mesa/mesa-pact.hoon
```diff
@@ -75,18 +73,39 @@
...
+  ?~  u.res
+    ~&  %inner-sig  ~
+  ~&  >  test-mesa-ns/;;([@tas pact:pact:ames.nec] [p q.q]:u.u.res)
+  ~
+::
+++  test-mesa-verify-msg-fail
+  ~?  >  dbug  'test-mesa-verify-msg-fail'
+  ::  4-fragment message
+  ::
+  =/  dat          (crip (reap 3.500 'A'))
```
yosoyubik, line 95: @lukechampine this is the test that fails to verify in `+verify-msg:verifier` here: `?>  =(?=(^ pair) (expect-pair state))`.

Extended output of the fragment, proof, and state of the verifier:

https://gist.github.com/yosoyubik/3a8cbb16b251044787cfc6377d62ec81

lukechampine: yeah I think it's safe to just remove `expect-pair`. Its definition is changing with the new traversal code, and removing it won't break anything in the interim.

yosoyubik: Then we can't make progress from the second fragment; verify-msg just returns `~`, i think this is the same behavior we saw doing real tests between two ships.

This is what we pass into the verifier:

```
>>> [ %no-state
    los
  [ leaves=4
    leaf=2
    cur=[l=2 r=4]
    leaf-queue=~
    parent-stack=~[0xb121.29ae.2118.9091.fac2.fe02.b665.a550.34a6.b38a.8616.767d.58bb.b9ff.b4f0.f4c4]
  ]
    dat
  42.774.436.690.957.487.429.920.950.773.873.837.743.069.182.823.381.674.213.013.050.438.812.601.248.985.070.400.637.813.118.635.372.047.998.325.168.476.746.089.141.563.561.575.575.845.631.803.128.042.257.050.944.270.835.248.713.
450.570.141.561.773.648.949.221.948.896.141.709.476.007.716.544.570.298.404.884.710.638.845.251.054.364.745.800.009.903.364.702.635.294.512.925.491.236.699.482.388.443.051.211.017.739.235.358.046.881.398.192.588.443.622.139.608.4
25.389.256.286.021.617.484.075.643.872.657.595.932.523.297.570.808.530.027.229.409.623.791.287.260.424.113.749.651.212.559.076.019.787.373.071.627.198.660.237.146.761.073.821.286.754.711.039.912.059.575.094.087.308.460.538.134.59
9.112.473.530.495.211.099.812.019.678.147.517.684.145.065.431.265.014.889.562.813.217.646.937.292.592.961.398.237.259.433.775.352.607.549.621.259.048.979.083.511.648.344.561.627.675.718.761.969.477.057.659.780.240.730.794.181.427
.824.163.451.193.240.134.744.953.291.046.723.379.654.408.093.933.751.584.505.323.115.091.485.389.339.898.333.151.803.667.132.108.092.547.858.435.961.640.141.833.312.102.783.131.647.304.455.375.792.420.212.084.590.846.083.814.942.
236.169.055.982.937.657.912.123.941.995.227.632.643.570.563.316.421.761.606.561.685.586.634.696.604.995.442.755.782.291.687.001.057.951.566.316.577.082.807.075.870.228.527.817.254.586.722.127.166.329.827.416.369.649.433.587.598.5
78.679.001.082.798.535.632.485.622.298.868.940.926.981.210.923.046.099.817.129.016.301.734.683.684.535.933.924.317.145.628.841.970.810.600.470.750.224.529.248.412.092.448.613.875.644.552.042.391.315.708.976.979.133.621.677.649.21
3.161.629.846.697.876.281.292.318.273.448.136.465.785.401.512.157.184.475.237.605.098.226.578.381.480.857.012.294.529.170.218.896.533.280.478.886.508.247.386.525.723.208.674.106.432.857.597.106.646.333.283.435.437.396.717.602.202
.934.212.275.822.958.383.738.197.049.075.271.262.078.912.983.459.205.070.726.955.796.257.818.513.538.380.534.226.524.076.285.916.462.990.526.366.868.568.924.354.494.877.939.201.171.247.379.951.751.482.971.692.277.623.261.925.221.
319.542.774.970.140.463.512.274.627.232.266.304.115.884.756.424.247.389.414.888.184.048.320.751.324.153.648.453.537.888.514.329.529.102.755.927.670.896.984.624.999.116.346.726.602.470.106.223.507.849.099.263.994.611.537.375.909.5
62.311.555.178.876.323.008.009.177.522.396.194.468.821.615.572.042.217.497.514.619.462.530.841.240.029.667.033.373.290.432.304.855.746.648.344.165.046.156.023.126.743.777.659.172.286.457.457.624.948.434.445.123.198.758.311.550.19
8.095.587.894.560.738.602.637.852.493.937.447.226.410.036.420.547.040.204.987.632.055.073.249.206.063.155.857.922.343.844.140.259.054.158.355.971.400.302.664.400.861.846.681.781.905.327.219.431.063.376.464.805.447.214.692.380.546
.822.473.964.350.219.488.759.245.785.488.821.114.297.968.066.900.539.394.568.523.757.953.475.444.442.702.208.196.246.588.914.565.678.771.135.791.032.687.372.911.088.562.779.335.671.141.910.736.526.356.920.007.788.852.838.700.199.
878.278.007.418.840.836.845.432.345.994.805.533.851.476.031.124.532.064.719.229.920.393.409.020.525.405.037.855.629.303.460.558.118.745.693.713.320.016.424.430.431.404.526.965.123.617.814.516.219.437.289.232.700.044.985.224.087.3
43.788.447.199.285.547.807.482.610.841.919.630.888.705.772.019.830.018.655.521.290
  pair=~
]
```

That pair=~ looks suspicious...

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -76,5405 +76,8485 @@
...
-                (lth rift.u.parsed rift.peer-state.peer-core)
-            ==
-          ::  ignore events from an old rift
-          ::
-          %-  %^  ev-trace  odd.veb  her
-              |.("dropping old rift wire: {(spud wire)}")
-          event-core
-        =/  =bone
-          ?-(u.parsed [%new *] bone.u.parsed, [%old *] bone.u.parsed)
-        =?  peer-core
```
joemfb, line 7962: @yosoyubik @lukechampine I think the lss problem is here -- this should be `(snag ?((gth wid 4) fag (dec fag)) pairs.lss-proof)`.

joemfb, line 7962: @yosoyubik @lukechampine I think the lss problem is here -- this should be `(snag ?((gth wid 4) fag (dec fag)) pairs.lss-proof)`. (reposting as gh attached this to a random location in the file).

yosoyubik: It looks like it's still happening—no state for leaf=2 and same output for the state of the verifier, wid and cv.

belisarius222:
## directed messaging meeting notes Monday, July 8, 2024

@lukechampine got the LSS transceiver working.  @pkova has been attempting to integrate jumbo frames into the main branch.  So far it's showing a significant (up to 10x) speed improvement, but there's also a bug.  He and @lukechampine will pair on that integration today.

@belisarius222 has been working on route switching.  He and @lukechampine will pair on it tomorrow.  It's compiling, but it's not a full implementation of route switching.  The next steps are to finish writing the route switching logic then try it on fakezods.

belisarius222:
## directed messaging meeting notes Monday, July 15, 2024

@lukechampine noticed packets are being sent twice to galaxies.  Seems like a bug in ames.hoon.  This shouldn't break anything but needs to be fixed at some point.

@lukechampine got four-fragment messages to stop crashing.  There is still an authentication failure, which we are examining in this meeting.

@belisarius222 checked that single-fragment messages still work.

We're leaking memory all over the place right now, so at some point we'll need to revisit the memory lifecycle design.  

Current state of routing:
- trying to debug basic message sending and receiving
- still need to add the next hop to forwarded packets
- jumbo frames still only work if they're the size of the whole message -- we need to figure out some kind of scheme on the publisher for emitting leaf packets that refer to hashes in the next jumbo frame. We could potentially launch without this, but it would limit the maximum message size to around 512MB, since IPC has a limit there. 

@belisarius222 If you run `|ahoy` in both directions, the second one will cause a crash that repeats forever on the receiving ship.  It should just ack it positively.  Could be a good task for @yosoyubik.

@pkova The `+load` arm in Ames has an option that can be uncommented (search for `chums.old`) to clear the whole state, allowing you to restart a test without rebooting whole fakezods.  That has to be done on both ships.

@lukechampine just found a bug in `_mesa_send_jumbo_pieces()` where we seem to be hashing the whole packet, rather than the data payload.  Aaaand he fixed it during the meeting.  Now messages of size 1-4 fragments work.

pkg/arvo/sys/lull.hoon
```diff
@@ -2317,19 +2320,20 @@
...
-        (both (just %peek) name)
-      %+  pick
-        :(both (just %page) name data (many [0 2] lane))
-      :(both (just %poke) name name data)
+      =/  hop  (rand 7)
+      ;:  pick
+        :(both hop (just %peek) name)
+        :(both hop (just %page) name data (many [0 2] lane))
+        :(both hop (just %poke) name name data)
+      ==
```
lukechampine, line 2336: It occurs to me that composing `+pick` in this way results in a significant bias: the first option has a 50% chance of being picked, while the next two only have a 25% chance each. To ensure that the odds are even, `+pick` needs to take a list and generate a random index:
```hoon
    ++  pick
      |*  gens=*
      %.  ^.(limo gens)
      |*  l=(list gen)
      |=  eny=@uvJ
      =^  a  eny  ((rand (lent l)) eny)
      ((snag a l) eny)
```
(@joemfb is there a cleaner way to write this? I tried the obvious approaches but they didn't work.)

Two choices is the most common case, so we probably want to keep an arm like that around, though I'm not sure what to name it -- `+pick2`? `+toss`? `+flip`?

belisarius222:
## Directed Messaging Meeting Notes, Monday July 29, 2024

- Sending large messages is now much slower for some reason, as of the last couple weeks.
- Hashtable deletion has bugs -- `~master-morzod` will investigate.
- Remote scrying broke (possibly), or there could just be a strangeness in the testing environment.
- The new request retry logic crashes.  ~rovnys-ricfer will fix today.
- Should we be caching LSS state, or maybe even fully formed response packets, on the publisher? Re-hashing response packets and recalculating LSS state whenever we send a single response packet is likely quite slow.  If we do want to cache response packets, what kind of data structures and lifecycle do we want for them?  ~rovnys-ricfer and ~watter-parter will look into this today.

yosoyubik:
Some crashes/errors/issues currently being investigated:

- Occasional crash on +ev-authenticate:mesa—a restart fixes it and (apparently) doesn't manifest again
```hoon
key=0x7b91.b96a.6126.76c0.1657.7dc4.2425.915f.5aaa.cc7a.a304.25cf.fa3d.4db8.9691.e087
aut=0x3f6.d7dc.c775.8d3c.5937.eef7.ee56.a73a 
ful=/~met/1/chum/1/~nec/1/0vun98.c6adu.3afbk.vmv70.fgi91.rsa6v.m9t2h.jmi6k.lphhc.p7js4.s7mid.k0lta.ru2bp.dr362
rut=0x9c1f.7520.4105.57b6.9967.79d9.5b73.b0f4.8a9e.c8e7.3b6f.0a9b.266b.129e.b7c5.4f78
```
- Occasional bail:meme on poke-payload producer
- Very aggressive retries, even after full jumbo message has been received
- Memory error (freeing a non-allocated variable? confirm) on peek sender—seeing in `_mesa_send_cb` when calling `_mesa_free_seal`
- General slowness
- Sometimes a peek gets stuck and stops making progress, even after restarting both sender/receiver

yosoyubik:
After a bunch of fixes, current issues are:

- General slowness (see lss_build in _mesa_send_jumbo_pieces)
- Arvo retries reset congestion control metrics, causing a slow restart
  - this is also seen after the OOM kill, but the driver needs to be restarted to empty the cache 
- Injecting an 8MB `(bex (bex 26))` payload fails to `++ev-authenticate`
- OOM kill due to too much (64G) swap memory being used (see _somewhere_ in _mesa_send_jumbo_pieces)

PR #6936, opened 2024-03-15
ping: only ping galaxy

pkova:
`/app/ping` has a few problems. One of them is that we created a state machine hole with the informal pinging changes in the upcoming 411k release. This is a serious, luckily unreleased bug that would make the galaxies lose track of their planets if the planets were restarted more than once after upgrading to 411.

The other one is that the state transitions were triggered for any poke-ack even though we really only care about the galaxy.

This is my attempt at making `/app/ping` only ping the galaxy. In addition, we rip out all of the jael stuff and just scry every time since most people will be informally pinging anyways. `/app/ping` used to have more responsibilities wrt subscribing to our own keys but they should be addressed by #6838.

We also rip out the hacky `%no-nat` stuff that was being used as a workaround in times before informal pings.

### DIFF
--- a/pkg/arvo/app/ping.hoon
+++ b/pkg/arvo/app/ping.hoon
@@ -22,18 +22,6 @@
 ::
 +$  card  card:agent:gall
 ::
-+$  state-2
-  $:  %2
-      ships=(set ship)
-      nonce=@ud
-      $=  plan
-      $~  [%nat ~]
-      $%  [%nat ~]
-          [%pub ip=(unit @t)]
-          [%off ~]
-          [%one ~]
-      ==
-  ==
 +$  state-3
   $:  %3
      mode=?(%formal %informal)
@@ -66,7 +54,7 @@
     ?:  &(!force (gth pokes.state 0) =(ship galaxy.state))
       [~ state]
     :_  state(pokes +(pokes.state), galaxy ship)
-    [%pass /ping/(scot %p ship) %agent [ship %ping] %poke %noun !>(~)]~
+    [%pass /ping %agent [ship %ping] %poke %noun !>(~)]~
 --
 %+  verb  |
 ^-  agent:gall
@@ -111,6 +99,18 @@
         $%  [%nat ~]
             [%pub ip=(unit @t)]
     ==  ==
+  +$  state-2
+    $:  %2
+        ships=(set ship)
+        nonce=@ud
+        $=  plan
+        $~  [%nat ~]
+        $%  [%nat ~]
+            [%pub ip=(unit @t)]
+            [%off ~]
+            [%one ~]
+        ==
+    ==
   ::
   ++  state-0-to-1
     |=  old=state-0
@@ -125,7 +125,14 @@
   ++  state-2-to-3
     |=  old=state-2
     ^-  state-3
-    [%3 %formal 0 ~ (galaxy-for our.bowl bowl)]
+    :*  %3  %formal  0  ~
+    =/  galaxy=(list @p)
+      %+  skim  ~(tap in ships.old)
+      |=(p=@p ?=(%czar (clan:title p)))
+    ?:  =(1 (lent galaxy))
+      -.galaxy
+    (head (flop (^saxo:title our.bowl)))
+    ==
   --
 ::  +on-poke: positively acknowledge pokes
 ::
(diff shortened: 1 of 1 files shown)


pkg/arvo/app/ping.hoon
```diff
@@ -345,6 +120,11 @@
     |=  old=state-1
     ^-  state-2
     old(- %2)
+  ::
+  ++  state-2-to-3
+    |=  old=state-2
+    ^-  state-3
+    [%3 %formal %.n (galaxy-for our.bowl bowl)]
```
joemfb, line 127: This scry in `+on-load` is a problem. Instead, we should either a) pick out the galaxy (there's exactly one, right?) from the set of ships in `old`, or b) initialize to our prefix galaxy: `(head (flop (^saxo:title our.bowl)))`.

pkg/arvo/app/ping.hoon
```diff
@@ -406,62 +154,33 @@
...
-      `state
-    ==
-  [cards this]
+  ?.  ?=([%ping s=@ *] wire)
+    `this
+  ?.  =(galaxy.state (slav %p i.t.wire))
+    `this
+  =.  sent.state  %.n
+  ?.  ?=(%formal mode.state)  `this
+  [[(wait-card now.bowl)]~ this]
```
joemfb, line 163: We need check that this is a `%poke-ack`, and explicitly handle nacks.

I think we could set the timer in %formal mode and when there's an error (`|(?=(%formal mode.state) ?=(^ p.sign.state))`). We might want to also switch to %formal mode on error -- less sure about that.

 If it's nack, we should either a) toggle the state to `%formal` or b) just retry once.

pkg/arvo/app/ping.hoon
```diff
@@ -406,62 +154,33 @@
...
-      ?.  ?=(%one -.plan.state)  `state
-      ?:  ?=(%poke-ack -.sign)   `state
-      :: XX handle error?
-      `state
-    ==
-  [cards this]
+  ?.  ?=([%ping s=@ *] wire)
+    `this
+  ?.  =(galaxy.state (slav %p i.t.wire))
+    `this
```
joemfb, line 160: This should be fine, so long as we only ever set the galaxy at the same time was we emit the ping (outside of `+on-init`).

joemfb (changes requested):

joemfb:
I think we might still have to keep a %jael subscription open for breach notifications. Otherwise, it seems like we won't start pinging the new galaxy until we restart or the driver forces us to (lane change or STUN failure).

pkova:
Ames will notify vere for any sponsor breach: https://github.com/urbit/urbit/blob/32b854e91f9603bd592b734d5cd4c05c4bab533e/pkg/arvo/sys/vane/ames.hoon#L2992-L2996

pkova:
And more relevantly, our sponsor change: https://github.com/urbit/urbit/blob/32b854e91f9603bd592b734d5cd4c05c4bab533e/pkg/arvo/sys/vane/ames.hoon#L3037-L3042

joemfb:
I see, the driver will then inject `%stop`.

pkg/arvo/app/ping.hoon
```diff
@@ -355,48 +135,16 @@
...
-    ?:  =(q.vase %no-nat)
-      =.  plan.state  [%pub ~]
-      (kick:ships our.bowl now.bowl)
+      =?  mode.state  =(+.q.vase %.y)
+        %formal
+      (ping (galaxy-for our.bowl bowl))
+    ::
+    ?:  |(=(q.vase %once) =(q.vase %stop))  :: NB: ames calls this on %once
+      =.  mode.state  %informal
+      (ping (galaxy-for our.bowl bowl))
```
joemfb, line 144: I wonder if we have a race condition here with `%once`, in the case where we have already sent a poke to the galaxy but it hasn't been acknowledged yet. Presumably the galaxy will hear from the correct lane in most cases, but it seems like it's possible for the poke to be sent on lane *A*, the lane to change to *B*, `%once to be injected but no-op (because we've already sent and the galaxy is unchanged), and then the ack to come back.

Online lane changes are presumably pretty rare (ISP reconfiguration, DHCP lease expiration, that sort of thing). I'll need to think about how to handle this scenario, and how important it is.

pkg/arvo/app/ping.hoon
```diff
@@ -48,256 +34,39 @@
...
-        ?:  =(our ship)  cards
-        =/  wire  /one/(scot %uw nonce.state)/ping/(scot %p ship)
-        :_  cards  ^-  card
-        [%pass wire %agent [ship %ping] %poke %noun !>(~)]
-      --
-    --
+  ++  galaxy-for
+    |=  [=ship =bowl:gall]
+    ^-  @p
+    =/  next  (sein:title [our now our]:bowl)
```
yosoyubik, line 52: This needs to be `(sein:title ship [now our]:bowl)` otherwise we bail:meme recursing forever and the first %stun %stop will fail.

pkg/arvo/app/ping.hoon
```diff
@@ -20,284 +20,42 @@
...
-          [%pub ip=(unit @t)]
-          [%off ~]
-          [%one ~]
-      ==
+::
++$  state-3
+  $:  %3
+     mode=?(%formal %informal)
+     pokes=@ud
+     timer=(unit [=wire date=@da])
```
joemfb, line 29: This is not enough information to cancel timers if we decide to do so later. The duct on which we set the timer will be differen ift `!=(our src):bowl`. If we care about getting around this, we need to use the "self-poke trick" to get a consistent duct before we actually set the timer. I don't think we need to care though.

joemfb (approved):
this will do, tyfys

PR #6937, opened 2024-03-16
ames: emit %saxo on ++molt

yosoyubik:
Before this, ships wouldn't start sending informal pings unless the driver was restarted. Now, as soon as we get the 411 OTA, we introduce a dummy state migration that will emit the %saxo gift to unix (same behavior that happens on-born).

(We are only doing this if we did have a unix-duct, if we don't, we skip that since the larval stage will be expecting a %born task from unix anyway, and that code path—`++on-born:ev:adult-core`—is what always sends the sponsorship chain to the driver)

Note: tested on a live ship that applies the state migration; that and booting from a pill will need to be tested once https://github.com/urbit/urbit/pull/6936 is ready.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2046,7 +2046,7 @@
         ::  start sending informal pings to the sponsorship chain
         ::
         ?~  unix-duct.+.u.cached-state
-          ~
+          moz
         ~>  %slog.0^leaf/"ames: retrieving sponsorship chain"
         ^-  (list move)
         :_  moz
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -101,6 +101,7 @@
       rot=`?`%.n  ::  routing attempts
       kay=`?`%.n  ::  is ok/not responding
       fin=`?`%.n  ::  remote-scry
+      sun=`?`%.n  ::  STUN
   ==
 =/  packet-size  13
 =>
@@ -850,7 +851,7 @@
       =life
       =rift
       crypto-core=acru:ames
-      =bug
+      bug=bug-19
       snub=[form=?(%allow %deny) ships=(set ship)]
       cong=[msg=@ud mem=@ud]
   ==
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -2017,7 +2038,22 @@
       ::
       =?  u.cached-state  ?=(%18 -.u.cached-state)
         19+(state-18-to-19:load:adult-core +.u.cached-state)
-      ?>  ?=(%19 -.u.cached-state)
+      =^  moz  u.cached-state
+        ?.  ?=(%19 -.u.cached-state)  [~ u.cached-state]
```
belisarius222, line 2042: shouldn't this preserve `moz` from the earlier upgrade, rather than shadowing those moves with `~`?  This applies to the branch with no unix-duct too.

yosoyubik: Let's see... I think that only happens when we are not doing a state migration (going from %n-1 to %n—%19 to %20—but rather when the larval stage is called with the current state %n = %20). If we are coming from an older version in the cached-state we are going to go through all state migrations in sequence so we will never go into the `?.` branches.

Reading the `+load` arm, it looks like that's never going to happen because we never regress to being a larva from the last adult ames-state... unless we are a larva which will mean we are booting from the first time, and then either we will trigger those timers (`public-keys`, `recork`...) `+on-init`, or in the case of the `%saxo` in `+on-born`.

yosoyubik: But the case of the no unix-duct, yeah you're right, we need to preserve moz; thanks!

yosoyubik:
I've tested this on top of https://github.com/urbit/urbit/pull/6936 (with this[1] and this [change](https://github.com/urbit/urbit/pull/6936#discussion_r1528015945) both applying the 411 OTA and booting from a pill straight into the latest %ames/%ping states and STUN works—%stop is introduced and informal pings start being sent (minus checking the galaxy for the correct lane)

This is the happy-path so more edge cases like [this](https://github.com/urbit/urbit/pull/6936#discussion_r1526708907) will still need to be tested

<details close>
<summary>[1] state-2-to-3 migration</summary>

```
++  state-2-to-3
  |=  old=state-2
  ^-  state-3
  :^  %3  %formal  %.n
  =-  ?>(?=([@ ~] -) i.-)  :: exactly one galaxy in the chain
  (skim ~(tap in ships.old) |=(=ship =((clan:title ship) %czar)))
```

</details>

PR #6939, opened 2024-03-16
vats: fix vats and make faster

tinnus-napbus:
`+vats` is agonisingly slow on my ship. This speeds it up by replacing tape interpolation & pretty-printer usage with explicit printing functions and cord concatenation. It also optimised a bunch of the code & reduces excessive scrying. Additionally, it fixes the %blocking filter to correctly print only the desks that are blocking the oldest pending kernel update.

before: `took s/107.362.567` 
after: `took s/31.124.617`

### DIFF
--- a/pkg/arvo/gen/vats.hoon
+++ b/pkg/arvo/gen/vats.hoon
@@ -17,12 +17,4 @@
         [filt=@tas verb=_|]
     ==
 :-  %tang  ^-  tang
-?.  &(=(~ deks) =(%$ filt))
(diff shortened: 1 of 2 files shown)


pkova (approved):

jfranklin9000:
When called with a list of desks, ex: `+vats %groups %talk %landscape %garden`, the output is flopped (`%garden` is listed first).  On `412` `%groups` is listed first, as expected.  When called as just `+vats` the output looks correct.

PR #6944, opened 2024-03-21
ames: fix chain mop order

pkova:
The `chain` in ames ordered the wrong way around. This problem shows up on line 3335 here: https://github.com/urbit/urbit/blob/49975d9821e775241069ad6e2584a76ee5c84a44/pkg/arvo/sys/vane/ames.hoon#L3330-L3340

`num` there is always either 1 or 2. This causes the key `2` to be overwritten all the time which is obviously not good.

@yosoyubik noticed this when working on new ames, but this is wrong in 411 too.

```hoon
> =mymop ((on ,@ ,[key=@ =path]) lte)
> (pry:mymop (gas:mymop *((mop ,@ ,[key=@ =path]) lte) [2 [0x1 /hello]] [1 [0x1 /hello-2]] ~))
[~ [key=1 val=[key=1 path=/hello-2]]]
```

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -3306,7 +3306,7 @@
         ^+  event-core
         =/  key=@  (shaz eny) :: TODO: check key width
         =/  num=@ud
-          ?~  latest=(pry:on:chain chain.ames-state)
+          ?~  latest=(ram:on:chain chain.ames-state)
             1
           .+(key.u.latest)
         =.  chain.ames-state


joemfb:
Rather than migrate, wouldn't it work to just change line 3335 to call `+ram`?

pkova:
Certainly, but I was assuming a `mop` was chosen here because we want efficient lookup of the head.

joemfb:
`+pry` gets the leftmost leaf and `+ram` gets the rightmost leaf. in either the case, we have roughly the same number of pointers to chase (`(log n)` in the number of entries). Unless I'm missing something else, I think we should just change the one line.

(Because of this bug we may need to tombstone / rotate keys on pre-release ships, that would obviously require a migration.)

pkova:
Done. I find it very unlikely people have been using this system in anger on pre-release ships, so I would be inclined to not bother with messing with the keys.

joemfb (approved):
this will do

PR #6949, opened 2024-03-25
eyre: scry with root gang in http generators

pkova:
#6876 started enforcing the gang system when scrying. Eyre has the capability to register generators to provide responses for http requests. These would now scry with the unauthenticated `~` gang if the http request is unauthenticated, and the root gang `[~ ~]` when authenticated. This broke `/app/acme` letsencrypt certificate renewal that was relying on making `%gx` scries for unauthenticated requests from `/gen/acme/domain-validation`.

This PR allows http generators to scry with the root `[~ ~]` gang. This should be fine since the interface for the generators accepts the `authenticated=?` flag and can use it do whatever dispatch it wants.


joemfb (approved):
good catch, this will do

PR #6952, opened 2024-03-25
hoon.hoon: update $rule to $-(nail edge)

tinnus-napbus:
You can't actually cast rules to the current weird `$rule` typedef. This makes it `$-(nail edge)` as it logically should be.

@joemfb is there some esoteric reason for it not to be `$-(nail edge)`?


tinnus-napbus:
ok I see it all has to be wet to retain the type of the result in the edge

PR #6953, opened 2024-03-26
gall: don't ack %leave for non-running agents

yosoyubik:
In %gall, %pokes to a non-running agent are dropped, and a `%flub` task is sent to %ames to "forget" about the message that has been sent by decreasing the heard (but not acked) sequence number, and removing the pending-ack from its queue.

If the %poke is a `%leave`(s), %gall automatically %acks it (giving a %done to %ames) before sending the %flub, which means that %ames has already remove the pending-ack, and we are going to crash on the %leave, and send a %nack, making the subscriber to retry the %leave forever, every ~m2.

The fix: ~~we just check that the queue is not empty.~~ don't ack %leave for non-running agents

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -4585,10 +4585,12 @@
                 %drop  sink(nax.state (~(del in nax.state) message-num.task))
                 %done  (done ok.task)
                 %flub
-              %=  sink
-                last-heard.state        (dec last-heard.state)
-                pending-vane-ack.state  ~(nap to pending-vane-ack.state)
-              ==
+              =?  pending-vane-ack.state  ?=(^ pending-vane-ack.state)
+                ::  a %leave gets acked in %gall before sending the %flub,
+                ::  so %ames has already removed the pending ack from its queue
+                ::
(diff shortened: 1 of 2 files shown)


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -4585,10 +4585,12 @@
...
-              %=  sink
-                last-heard.state        (dec last-heard.state)
-                pending-vane-ack.state  ~(nap to pending-vane-ack.state)
-              ==
+              =?  pending-vane-ack.state  ?=(^ pending-vane-ack.state)
+                ::  a %leave gets acked in %gall before sending the %flub,
+                ::  so %ames has already removed the pending ack from its queue
+                ::
+                ~(nap to pending-vane-ack.state)
+              sink(last-heard.state (dec last-heard.state))
```
joemfb, line 4593: Rather than patching this up here, I think it'd be better to change %gall to not unconditionally ack `%leave` in the case where the app is suspended.

belisarius222: Yeah this was my immediate reaction too

joemfb (approved):
this will do

PR #6955, opened 2024-03-30
base: %test agent - add desk to argument

midden-fabler:
Adds a desk to the argument of the `%test` agent, so it can now be run against desks other than `%base`.
before: `:test %agents`
after: `:test [%landscape %agents]`


pkova (approved):

joemfb:
I think this would be better if `$command` where `$@(test [desk test])`. Easy to default to `%base` and keep compatibility in this interface (so old CI scripts still run, &c).

midden-fabler:
I tried that initially and ran into issues. I'll give it another go.

PR #6959, opened 2024-04-08
verb+: detailed logging affordances

Fang-:
Primarily: adds a new subscription endpoint to `/lib/verb` (`/verb/events-plus`), which emits a new "verb event" type, `$event-plus:verb`. This type includes many details not present in the `$event:verb` type or the library's slog output, such as the `src.bowl`, n/ack status, mug hashes of included data, and a summary of the produced effects/cards.

Secondarily: adds two pieces of tooling for working with this new endpoint:
- `/app/verb-logger`, which can be set up to watch an agent, and writes out json files for that agent's logs to `pier/.urb/put/verb-logger/[agent]` periodically. (Currently hard-coded at once an hour, can be force-flushed to export earlier. Probably don't run this long-term without a "delete old logs" companion script.)
- A log viewer in the debug dashboard for rendering & inspection of logs of any number of agents on a chronological timeline. This log viewer can import the files output by `/app/verb-logger`, or can ingest events from running agents directly, similar to the pre-existing /lib/verb support in the debug dashboard.

Demo:

https://github.com/urbit/urbit/assets/3829764/8dc72203-2370-4e21-a80d-1b947abf596b

Tertiarily: moves the debug dashboard off of gulp (old and busted) onto vite (new hotness) for its build system. We continue to copy the resulting blobs directly into the pier, because we're mischievous little imps that want to see `%tomb` become load-bearing. (Making `/app/dbug` issue `%tomb` cards is left as an exercise to the reader.)

Of course, usage of this new functionality for any given agent depends on that agent using the new version of `/lib/verb`.

There's probably better ways to visualize the log viewer timeline. There's tons of stuff to be done with color-coding, same-event batches could be displayed differently, etc. Even just a better slice selector would be cool. But the current things is perfectly good enough for basic browsing & insights, so further improvements are out of scope here.

One might be concerned with the additional processing done by `/lib/verb` with this change, but: 1) it's all pattern-matching, mug calculation, and cell construction, all of which should be fast, 2) json conversions don't happen unless you're actively watching, 3) log-viewer only serializes periodically.


patosullivan (approved):
JS/vite related changes look good

pkova (approved):

PR #6960, opened 2024-04-09
Add Easter egg at the Dojo prompt.

sigilante:
It's always irked me that our wordmark `~Urbit` is not valid to type at the Dojo prompt.

Well, no longer is this a problem!

image

Maybe we should UIP this first, but it's a nice little Easter egg.


sigilante:
The code is a little rough in that it simply adds an alternate parsing route to `~Urbit` that directly invokes the targeted thread.  Since it's a special case, it probably has to/should be handled separately everywhere.

jalehman:
What does it look like constructed out of `~` characters instead? 

Would be nice to see something like this printed at boot too.

sigilante:
https://www.asciiart.eu/image-to-ascii and the SVG files from Other Means

sigilante:
I'd actually like to put in a nice little animation in text but that'll require a bit of Dill work.

pkg/arvo/ted/egg.hoon
```diff
@@ -0,0 +1,37 @@
...
+/-  spider
+/+  strandio
+=,  strand=strand:spider
+^-  thread:spider
+|=  arg=vase
+|^
+  =/  m  (strand ,vase)
+  ^-  form:m
+  ;<  =bowl:spider  bind:m  get-bowl:strandio
+  ;<  code=@p  bind:m  (scry:strandio @p /j/code/(scot %p our.bowl))
```
pkova, line 14: Don't need these.

sigilante:
Final version:

```
> -urbit
```

jalehman (approved):

PR #6961, opened 2024-04-10
hoon: adds new atom ops

joemfb:
To be included in hoon %137, as required by #6932.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -959,6 +959,12 @@
           ==
   ==
 ::
+++  clz                                                 ::  leading zeros
+  ~/  %clz
+  |=  [a=bite b=@]
+  =/  c=[=bloq =step]  ?^(a a [a 1])
+  (sub (mul (bex bloq.c) step.c) (met 0 (end a b)))
+::
 ++  dis                                                 ::  binary and
   ~/  %dis
   |=  [a=@ b=@]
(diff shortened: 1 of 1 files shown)


ashelkovnykov (approved):

pkg/arvo/sys/hoon.hoon
```diff
@@ -855,6 +868,19 @@
     [boz (sub len (met boz dat))]
   (swp boz dat)
 ::
+++  rig                                                 ::  convert bloqs
+  ~/  %rig
+  |=  [=bite b=bloq]
+  ^-  step
+  ?@  bite  0
```
ashelkovnykov, line 875: Forgot to ask: why are `$bite` with step size `1` not converted?

joemfb: This is a great question -- it's entirely because of `+hew` and the pattern I've used to compose calls to it in the directed-messaging packet decoder (see #6932). `+hew` takes a `$bite` as both a bloq specifier for the reads it will perform and an initial cursor position. It doesn't make sense to default the cursor to anywhere other than `0`, but the makes it a non-standard (and possibly abusive) interpretation of `$bite`.

For `+rig` in general, it would be simpler to be `$-([bite bloq] bite)` and just return the `bloq` from its sample when the `bite` is an atom, punting on the whole question. But that's less applicable my use-cases in #6963 and packet decoding. I need to play around with it, and possibly split it into two entrypoints.

PR #6963, opened 2024-04-11
lull: adds $plot

joemfb:
This PR adds a new declarative atom builder, supporting incremental and hierarchal composition, with a sample that is a superset of `+can`. Needed by #6932.


ashelkovnykov (approved):
It's `ur3_chop` but for the [2C](https://docs.urbit.org/language/hoon/reference/stdlib/2c) bit arithmetic operations.

PR #6967, opened 2024-04-16
ci: catch khan errors in ci

pkova:
https://github.com/urbit/urbit/pull/6955 broke CI by modifying `/app/test`, but CI still passed. This happened because we are running out tests via khan (click) and we were not catching khan errors, just errors inside the agents and threads.


pkova:
Note that the red cross here is exactly what we want.

Issue #6926, opened 2024-02-26
clay, ames: applying change to widely distributed desk is very slow


Fang-:
I've been suffering this for a while now, but don't think an issue for this has ever been filed, so here we go:

Initiating software updates on ~paldev is a _multi-hour_ affair. `|commit`ing a change on a public desk causes it to `<<sync>>` for a long time, essentially self-DoS-ing.

Observations:
- The amount of time this takes to process seems proportional to how many people have installed from the desk.
  - `|commit`ing for less popular apps is _relatively_ fast, on the order of minutes to tens of minutes.
  - `|commit`ing for popular apps like pals and rumors is extremely slow to resolve, on the order of hours.
- While this is happening, the serf is more or less pegging a CPU core to 100%. The VPS it's on isn't _that_ slow/old (`e2-medium` in gcloud).
- ~~paldev has never run out of memory during a `|commit`, despite sometimes operating close to its memory limit.~~ I should not have written that. It happened.

Happy to help out with additional analysis as desired, but I'm assuming we have good guesses as to where this slowness comes from.

pkova:
My bet is on the following:

Clay creates a new ames flow for every aeon of the desk -> congestion control for all subscribers is reset -> insane amounts of ames effects to notify subscribers of the new commit.

Combine this with the fact that most subscribers will be offline and so ~paldev will retry every two minutes for tons of ships multiplied by tons of commits = 1gb per day event log growth.

bacwyls:
```
    data: [%spot [%sys %vane %ames %hoon 0] [3385 15] 3400 35]
leak: 0x25a3dc918 mug=0 swept=1
    size: B/24
    data: [%spot [%sys %vane %ames %hoon 0] [3397 33] 3397 57]
leak: 0x25a3dc930 mug=0 swept=1
    size: B/24
    data: [%spot [%sys %vane %ames %hoon 0] [3397 11] 3400 35]
leak: 0x25a3dc948 mug=0 swept=1
    size: B/24
    data: [%spot [%sys %vane %ames %hoon 0] [3393 11] 3400 35]
leak: 0x25a3dc960 mug=0 swept=1
    size: B/24
    data: [%spot [%sys %vane %ames %hoon 0] [3392 11] 3400 35]
leak: 0x25a3dc978 mug=0 swept=1
    size: B/32
    data: (null)
leak: 0x25a3dca38 mug=0 swept=1
    size: B/32
    data: (null)
[...]

PR #6970, opened 2024-04-17
ames, gall: more sane clog handling

pkova:
We were investigating https://github.com/urbit/urbit/issues/6926 today when we realised that the `+check-clog` handling in ames is absolutely insane.

Ames maintains a queue of unsent messages for every flow. Publisher ships that send a lot of subscription updates want to be able to cancel their responsibility of sending these updates if the subscriber becomes unresponsive. This functionality is handled by ames `+check-clog` which notifies gall if a peer queues up a certain number of messages or the total size of messages becomes too large (configurable, but 5 messages and 100 000 bytes by default). When gall receives the clog notification it kicks the subscriber. Note that this only applies to %boons not %pleas. This is all reasonable, but here's where things get a little out of control.

`+check-clog` always iterates over every single flow a peer has. Not only that, but `+check-clog` is run _every time a ames retry timer fires._ It doesn't even matter if we're retrying a %plea or a %boon!

This PR changes the clog handling to trigger only when sending a %boon. It also makes gall handle the clog notification for a single agent instead of kicking the subscriber from all running agents.

Committing %rumors on a local copy of ~paldev takes hours without this change, with this it takes 20 seconds.

There's still some fine tuning to do on this PR but the general approach should be significantly more sane.


pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3655,6 +3650,16 @@
           |=  =bone
           ^+  peer-core
           abet:(call:(abed:mi:peer-core bone) %flub ~)
+        ::
+        ++  check-clog
+          |=  [=bone =ship]
+          ^+  peer-core
+          =/  =message-pump-state  (~(got by snd.peer-state) bone)
+          ?:  ?&  !=(0 (end 0 bone))
+                  (gth ~(wyt in unsent-messages.message-pump-state) msg.cong.ames-state)
```
pkova, line 3659: Do we need the byte count check as well? I could put it here but it kinda feels unnecessary.

pkova:
We discussed this today with @yosoyubik and realized we need to correlate gall subscriptions with ames boons to make this work properly. Working on that now.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -3655,6 +3650,16 @@
           |=  =bone
           ^+  peer-core
           abet:(call:(abed:mi:peer-core bone) %flub ~)
+        ::
+        ++  check-clog
+          |=  [=bone =ship]
```
belisarius222, line 3655: we shouldn't need `ship` here, should we? we're already in the peer core

belisarius222:
I'm thinking about how to correlate `%clog` with the `%boon`.  One thing you could try would be to have `%clog` be a per-flow note that Ames can pass to a vane that just gave it a `%boon`.  Ames would pass the `%clog` using the same duct and wire as for passing a `%plea` note, so Gall would receive it on the same duct.

Unfortunately, I don't think that works because the duct doesn't specify the request path for a `%plea` -- the fact that it doesn't has always given me a strange feeling, but I haven't been sure of what to do about it.

Maybe something dumber and more direct would work better, like Gall giving an extra piece of data in the `%boon` gift that Ames will ping-pong back immediately if it gives a `%clog`.  So it wouldn't use the duct -- rather, Gall would send something like `[publisher-agent publisher-path requesting-ship]` to Ames (maybe path-encoded, to avoid type problems) in the `%boon` gift.  Ames would include that same datum back to Gall in the `%clog`, and then Gall would use that to figure out which subscription to kill.  (You might need a subscription nonce in there too.)

This mechanism could be made optional, so Clay would leave the 'extra data' field null, signifying that it doesn't want Ames to check clog.

yosoyubik:
> giving an extra piece of data in the %boon gift 

that was our conclusion yesterday :)

joemfb (approved):
this will do

PR #6979, opened 2024-04-26
lull: use $+ on types in +rand

ryjm:
Makes thread debugging less of a nightmare.

The stacktrace from `-khan-eval '~'` is now 28 lines rather than... 4490.

([full stracktrace](https://littel-wolfur.startram.io/scratch/view/thread-error) if you need proof)


mikolajpp:
I had a look at these annotations: I agree with slapping `$+` on complex types such as `strand-form. Annotating the `strand-bowl` makes it easy not to be confused about which bowl are we talking about. 

However, should the `card` ultimately be annotated here, or rather at the source definition in gall?

Take this comment as a question about a system-wide annotation strategy, especially considering the pretty-printer UIP-0119.

ryjm:
yeah i thought about annotating it in gall but decided it increased the scope of this a bit too heavily, even though i do find those card stack traces annoying in app development - they're just a little bit easier to deal with there bc they tend to be a bit more tightly constrained to specific situations. that's not the case in threads, almost every failure results in a completely unreadable stack trace.

mikolajpp:
I have stumbled upon this PR: https://github.com/urbit/urbit/pull/6649, and its recent incarnation: https://github.com/urbit/urbit/pull/6755.

There are several annotations for basic types, including gall's`card`

tacryt-socryp:
Please just merge something. That other PR has been open for 9 months just to improve stack traces.

Both PRs make meaningful improvements to devex.

pkova (approved):

PR #6981, opened 2024-05-03
hoon: expand =| through ^* to enable constant-folding

joemfb:
I discovered this while helping debug ares-codegen, and was surprised to the current macro-expansion. I can't think of any reason not to make this change. It will help cut down on the overhead of calls into manually deconstructed cores (like `+in`, `+by`, `+nest:ut`, &c).


belisarius222:
We should benchmark before and after this.

PR #6991, opened 2024-05-21
ames: do not clog live peers

pkova:
Consider the following scenario as experienced by `~mopfel-winrux`:

A new ship subscribes to `%hits` on `~mopfel-winrux`. The app sends an initial subscription update that consist of a large list of `%facts`. Five of these `%facts` that Gall sends to Ames get queued as normal. On the 6th `%fact` Ames clogs the flow which is certainly not ideal since we haven't even had the chance to send anything over the network yet! Remember that we are still processing a singular event.

We also noticed that the clogging will happen for the 7th `%fact` and onwards during the same event. This will cause Gall to send duplicate `%kicks` to Ames, but we investigated this with @yosoyubik today and this seems harmless. The `%kick` receiver will immediately `%cork` and the duplicate `%kicks` become no-ops.

This PR adds back the ~s30 check for clogging that used to be active before my #6970. This will mostly prevent scenarios like the above since presumably when we send a big initial subscription update it happens because we just heard a `%watch` from the peer.

### DIFF
--- a/pkg/arvo/sys/vane/ames.hoon
+++ b/pkg/arvo/sys/vane/ames.hoon
@@ -2768,7 +2768,10 @@
         ?~  bone=(bone-ok u.parsed wire rift.peer-state.peer-core)
           event-core
         ::
-        abet:(check-clog:(on-memo:peer-core u.bone [%boon payload]) u.bone id)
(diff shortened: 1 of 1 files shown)


joemfb (approved):
This makes sense. "because we just heard a %watch" is the key insight I was missing yesterday.

PR #7000, opened 2024-05-23
Create /lib/math.

sigilante:
Upstream lives in urbit/numerics repo, but this distributes via `%base`.

### DIFF
--- a/pkg/arvo/lib/math.hoon
+++ b/pkg/arvo/lib/math.hoon
@@ -1,0 +1,3002 @@
+  ::
+::::  Mathematical library
+::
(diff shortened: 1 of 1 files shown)


sigilante:
I am not very happy with the current algorithm on some of the transcendental functions.  I used Taylor expansion instead of Chebyshev functions, which is much slower for smaller rtols.  I will be changing these out in the upstream urbit/numerics repo at some point; when I do things like that, I'll just PR them as updates here so we at least have this distributed.  (This is why I haven't jetted these yet.)

jalehman (dismissed):

jalehman (approved):

PR #7005, opened 2024-05-30
eyre & scry endpoints for double boot protection, and rift to keyfile

### LINKED ISSUE #7013: Double boot protection
It should be difficult to accidentally "double boot" your ship when it's already running. This project has the runtime ask the ship's sponsor what keys it has on file before boot, and if the supplied key matches the current keys, blocks the boot.

ripperi:
Resolves #7013


pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -997,11 +997,128 @@
...
+    ?~  mss
+      %^  return-static-data-on-duct  404  'text/html'
+      %:  error-page
+        404
+        &
+        url.request
+        "Bone {(scow %u u.bone)} of peer {(scow %p u.ship)} not found."
+      ==
+    %^  return-static-data-on-duct  200  'text/plain'
+    (as-octs:mimes:html (scot %u last-acked.u.mss))
```
pkova, line 1121: This should return the rift as well, the aura should be `%uw`, the type should be named and versioned in lull.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -997,11 +997,128 @@
...
+    ?.  ?=([~ ~ %noun *] des)
+      %^  return-static-data-on-duct  404  'text/html'
+      (error-page 404 & url.request "Peer {(scow %p u.ship)} not found.")
+    =/  ship-state
+      !<  ship-state:ames  q.u.u.des
+    ?.  ?=(%known -.ship-state)
+      !!
+    ?~  bone
+      %^  return-static-data-on-duct  200  'text/plain'
+      (as-octs:mimes:html (scot %u (jam [(galaxy-for u.ship) rift.-.+.ship-state life.-.+.ship-state])))
```
pkova, line 1109: %uw aura, the type should be named and versioned in lull.

pkg/arvo/sys/vane/eyre.hoon
```diff
@@ -997,11 +997,142 @@
...
+        &
+        url.request
+        "Bone {(scow %u u.bone)} of peer {(scow %p u.ship)} not found."
+      ==
+    %^  return-static-data-on-duct  200  'text/plain'
+    %-  as-octs:mimes:html
+    %+  scot
+      %uw
+    %-  jam
+    ^-  boot
```
Fang-, line 1128: Any particular reason to `@uw`-encode the responses here and above? I believe eyre uses plain byte buffers in the bodies most other places (except the noun channel SSE streams, into which we can't safely inject arbitrary bytes) so we should prefer sticking to that if we can.

pkg/arvo/sys/vane/ames.hoon
```diff
@@ -6091,6 +6091,33 @@
     ::
         [%protocol %version ~]
       ``noun+!>(protocol-version)
+    ::
+        [%boot ~]
+      =/  who
+        =/  ship  our
+        |-
+        ^-  @p
+        =/  next  (^^sein:title rof /eyre our now ship)
```
Fang-, line 6100:
```suggestion
        =/  next  (^^sein:title rof /ames our now ship)
```

This is ames doing the scry, provenance path should reflect that right?

Fang-: Also, this (and the same bit of logic in eyre) will have to change when directed messaging comes out, right? Because then we properly go to the "direct"/"first" sponsor as opposed to always up to the galaxy level?

If that's the case, would be good to have `TODO` comments on these.

Fang- (commented):
Cheeky drive-by. I lack most of the context, and neither PR provides any besides the code, so forgive me if any of this was discussed elsewhere.

PR #7018, opened 2024-06-14
hoon: optimize %knit expansion

joemfb:
...  for cheaper literals and faster compilation. Opening as a draft as I've only tested this locally on toy examples.

See tocwex/fund#89 for context.


joemfb:
This PR attempts to pre-build tape types (using the hoon AST node `%hand`, otherwise unused), to avoid having the compiler first infer a fixed-width "tuple" type over the whole tape and then downcast to a recursive type. But this type hackery results in a (detected) infinite loop in type addition (`+fuse:ut`) ...

joemfb:
Getting this kind of approach to work will require more and broader work on the compiler than I can prioritize right now. For now, it should be known that all tape literals are slow to compile, and interpolated tapes are slow at runtime. Cord literals and manual cord/tape building works around these problems.

PR #7021, opened 2024-06-16
Revise libtext and libutf8

tinnus-napbus:
a) Changed libtext to use dry gates instead of +curr etc so they can be used in things like +turn.
b) It seems you created a new UTF-8 character list type so you could implement all-script case functions. This seems unnecessary as the stdlib already supports UTF-32. I have renamed the utf-8 lib as just utf & made it support both uft-8 and utf-32. I feel like this is more sensible than a new type but perhaps you have other uses for the utf-8 character list type? Also, the hoon parser is insanely faster at parsing an atom than the list of cells for the character map you had. I have replaced those with a single cord containing the pairs & the library just converts that to lists and build the maps. It now includes the complete set of Common+Simple unicode standard case mappings and it builds almost instantly cos the hoon parser doesn't get bogged down.

Let me know what you think of my alternative UTF lib, I can revert that if it's not what you want.


sigilante:
Re libutf8, I think we do want a UTF-8 native format and tapes are a little weird for that.  UTF-32 works but it's very wasteful for almost all applications, so we want support for both.

sigilante (changes requested):
Adding the UTF-32 tuba/tuft/taft stuff in and working in atoms is fantastic.

I think that tape-style lists of UTF-8 chars is also desirable, so we want that part back in some form.  Can be simpler than the brand/lasso thing if desired.

tinnus-napbus:
I have added back brand/lasso but operating on cords rather than tapes

PR #7025, opened 2024-07-01
Nathan/gzip

nathanlever:
GZIP decompression generator. 
Includes tests.


mikolajpp:
It is great to see the first working implementation of DEFLATE decompression, good job!

I have tested your code with my [hoon-git](https://github.com/mikolajpp/hoon-git/tree/gzip) library and uncovered some problems, which I describe below.

Firstly, the decompression of octs only is quite limited: a frequent use case is expansion of a DEFLATE stream embedded inside a larger data stream. Extraction of the compressed chunk is impossible, since the only way to find out its size is to decompress.

To be able to test your library on the compressed data in git, I have extended it to handle this case by porting it to use the Hoon bytestream library that I have released very recently: https://github.com/mikolajpp/bytestream. You can find the ported code here: https://github.com/mikolajpp/hoon-git/blob/gzip/desk/lib/zlib.hoon (the tests are also there and pass with this modified version). 

Secondly, here you implement only the `gzip` format, while I need the `zlib` format. 
I have extended the code to handle the `zlib` format as well. It is very rough and mostly for testing purposes so I could run decompression on my data. (The only difference is format of the header. Both zlib and gzip use DEFLATE).

Thirdly, it seems that there is some bug or a missing case in your code. While I can decompress the sample test cases you have supplied, and some basic compressed git objects, here is a shell command that decompresses base-64 encoded data stream. With your library, the error that I get when trying to decompress this data is `invalid distance too far back`. 
```
echo -n 'eJwzNDAwMzFR8PF0dvULdmXYuleh1k57+Q/uTVWGi7Z18x39Ev7OEKIkyNXRxddVLzeFYZHLbC7XPJuyoEc3MgWPcM+Ycm6+jokBECikpBZnM1T83j/B0291bpL2PuXCFbpLzx1+7woAfh0oEekBb3icS0+fECwi' | base64 -d | pigz -d
```

I will greatly appreciate it if you could investigate why does this error happen and what could be done to fix it.

nathanlever:
> It is great to see the first working implementation of DEFLATE decompression, good job!
> 
> I have tested your code with my [hoon-git](https://github.com/mikolajpp/hoon-git/tree/gzip) library and uncovered some problems, which I describe below.
> 
> Firstly, the decompression of octs only is quite limited: a frequent use case is expansion of a DEFLATE stream embedded inside a larger data stream. Extraction of the compressed chunk is impossible, since the only way to find out its size is to decompress.
> 
> To be able to test your library on the compressed data in git, I have extended it to handle this case by porting it to use the Hoon bytestream library that I have released very recently: https://github.com/mikolajpp/bytestream. You can find the ported code here: https://github.com/mikolajpp/hoon-git/blob/gzip/desk/lib/zlib.hoon (the tests are also there and pass with this modified version).
> 
> Secondly, here you implement only the `gzip` format, while I need the `zlib` format. I have extended the code to handle the `zlib` format as well. It is very rough and mostly for testing purposes so I could run decompression on my data. (The only difference is format of the header. Both zlib and gzip use DEFLATE).
> 
> Thirdly, it seems that there is some bug or a missing case in your code. While I can decompress the sample test cases you have supplied, and some basic compressed git objects, here is a shell command that decompresses base-64 encoded data stream. With your library, the error that I get when trying to decompress this data is `invalid distance too far back`.
> 
> ```
> echo -n 'eJwzNDAwMzFR8PF0dvULdmXYuleh1k57+Q/uTVWGi7Z18x39Ev7OEKIkyNXRxddVLzeFYZHLbC7XPJuyoEc3MgWPcM+Ycm6+jokBECikpBZnM1T83j/B0291bpL2PuXCFbpLzx1+7woAfh0oEekBb3icS0+fECwi' | base64 -d | pigz -d
> ```
> 
> I will greatly appreciate it if you could investigate why does this error happen and what could be done to fix it.

mikolajpp:
Many thanks for the prompt fix! It seems to be working well now. 

What is your view on integrating your library with the bytestream library? The stream approach is really a must for hoon-git to benefit from your gzip library, since most of the compressed data is embedded in a larger stream. It would also be necessary to slightly modify the gzip jet, so that it operates on `bays` instead `octs`.
We can also expose simple octs decompression arms.

I think we should also bring the commenting style in agreement with current doccords-based recommendations. 

If you are OK with this approach, I can do a PRs against your branch in the urbit and vere repo. From there on you could refactor it to your preference.

nathanlever:
Sounds good, I'm ok with that approach. You can do PRs and then I'll have a look at refactoring.

mikolajpp:
You can find my proposed changes here: https://github.com/mikolajpp/urbit/tree/nathan/gzip. I was not able to make a PR against your branch: for whatever reason your fork is not visible in the PR`s base repository selection. Could you please create a PR from my branch yourself?

mikolajpp:
The matching decompression jets are in this PR: https://github.com/nathanlever/vere/pull/1

mikolajpp:
@nathanlever Any chance you could have a look at these PRs soon?

PR #7034, opened 2024-07-16
hoon: faster implementation of tuple indexing

belisarius222:
I haven't actually tested whether this is faster than what we have, but it should be since it's nonrecursive.  I have tested that the algorithms match what was there, including for zero.

### DIFF
--- a/pkg/arvo/sys/hoon.hoon
+++ b/pkg/arvo/sys/hoon.hoon
@@ -13833,8 +13833,8 @@
         |=(a=axis [%& a])
       ;~  pose
         ;~(pfix lus dim:ag)
-        ;~(pfix pam (cook |=(a=@ (sub (lsh [0 +(a)] 1) 2)) dim:ag))
-        ;~(pfix bar (cook |=(a=@ (sub (lsh [0 +(a)] 1) 1)) dim:ag))
+        ;~(pfix pam (cook |=(a=@ (sub (bex +(a)) 2)) dim:ag))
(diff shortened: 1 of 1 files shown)


pkg/arvo/sys/hoon.hoon
```diff
@@ -13833,8 +13833,8 @@
         |=(a=axis [%& a])
       ;~  pose
         ;~(pfix lus dim:ag)
-        ;~(pfix pam (cook |=(a=@ ?:(=(0 a) 0 (mul 2 +($(a (dec a)))))) dim:ag))
-        ;~(pfix bar (cook |=(a=@ ?:(=(0 a) 1 +((mul 2 $(a (dec a)))))) dim:ag))
+        ;~(pfix pam (cook |=(a=@ (sub (lsh [0 +(a)] 1) 2)) dim:ag))
+        ;~(pfix bar (cook |=(a=@ (sub (lsh [0 +(a)] 1) 1)) dim:ag))
```
joemfb, line 13837: These could be written `(sub (bex +(a)) 2)` and `(sub (bex +(a)) 1)`.

joemfb (changes requested):
Good catch!

a nit:

pkg/arvo/sys/hoon.hoon
belisarius222: oh yeah good point

pkova (approved):

PR #7026, opened 2024-07-02
Add -new-app thread to build from repo /desk

sigilante:
This should call out to a supplied repo, or sigilante/sunrise if none is supplied.

`-new-app` thread:

Create a new app (desk + agent) from a template GitHub repo:

```
-new-app %dawn /sigilante/sunrise
```

The new desk exists with the supplied name and the contents is a copy of the remote repo.

Steps behind the scenes:
1. Retrieve current version of `%sunrise` from [`sigilante/sunrise`](https://github.com/sigilante/sunrise) or from the supplied repo URL. (`/ted/retrieve-latest-commit`, `/ted/retrieve-github`)
2. Add the `/desk` material into Gall.
3. Rename all of the files and internal matches in `/desk` if empty.
4. Copy the modified `/desk` material into Clay.

PR #7035, opened 2024-07-17
lull: do not validate %thread-done in the +strand monad

pkova:
Consider the following thread:

```hoon
/-  spider
/+  *strandio
=,  strand=strand:spider
^-  thread:spider
|=  args=vase
=/  m  (strand ,vase)
;<  =thread-result  bind:m  (await-thread %code !>(~))
?>  ?=(%.y -.thread-result)
=/  res  !<(@p +.thread-result) :: devasing crashes here
(pure:m !>(res))
```
This thread will always crash when trying to devase the result of the child thread. This happens because spider ends up calling `take:eval` for all `%facts` it receives, which in turn ends up in `+validate-mark` in lull. "Validating" the mark in this case involves throwing away the type information so `@p` becomes `*`.

It is quite unfortunate that spider needs to validate marks like this because in this case gall has already taken care of it, but spider also needs to deal with threads being triggered via HTTP. The interface for triggering threads has input and output marks and a desk where the marks files reside. This means that spider has to be responsible for validating marks as well.

Here we special case `%thread-done` in `+validate-mark`. Note that we do not need to special case `%thread-fail` since the return type is not polymorphic.

Thanks to @sigilante for discovering this while working on #7026. Prior art starts at #5697.


pkg/arvo/sys/lull.hoon
```diff
@@ -3489,8 +3489,10 @@
         =.  in.strand-input
           ?~  in.strand-input  ~
           =/  in  u.in.strand-input
-          ?.  ?=(%agent -.in)      `in
-          ?.  ?=(%fact -.sign.in)  `in
+          ?.  ?=(%agent -.in)                  `in
+          ?.  ?=(%fact -.sign.in)              `in
+          ?.  ?=(%fact -.sign.in)              `in
```
joemfb, line 3494: You're checking `%fact` twice here

pkg/base-dev/mar/thread-fail.hoon
```diff
@@ -8,4 +8,4 @@
   ++  noun  err
   --
 ++  grad  %noun
---
+--
```
joemfb, line 11: You removed the unix-idiomatic trailing newline

joemfb (changes requested):
couple nits:

joemfb (approved):
this will do

PR #7043, opened 2024-07-29
hoon: various ?= codepath fixes and improvements

Fang-:
Collection of bugfixes and small improvements @joemfb and I have made to code in the `?=` codepath.

42a9700 and 3199204 are small correctness changes. 16c21d0 is a simple short-circuiting performance optimization.

fea003b is a proper breaking change, see also the change we needed to make to behn. Shouldn't impact too much, but I know for a fact there is code affected by it out there in the wild.


belisarius222:
Since this includes a nontrivial breaking change, it should be deployed as part of a Hoon kelvin update.

PR #7056, opened 2024-08-12
mass: add %mass agent to run & record memory reports

tinnus-napbus:
Depends on https://github.com/urbit/urbit/pull/7037 and https://github.com/urbit/vere/pull/675

Add a %mass agent that runs memory reports via Dill and records the results with the time, size in bytes and base hash at report time, for each entry and sub-entry in the report. It may be configured to run them every `@dr` period, or every `d` days at `h:m`, else can run them ad-hoc. Other agents may subscribe to & scry out reports, either for individual paths in the report like `/arvo/hoon` or for everything. It includes a whitelist for remote ships to run reports, configure polling frequency & get results. JSON conversion methods are included for all requests and updates so requests, subscriptions and scries can be done from a front-end. Old records may be pruned. Records for a particular path can also be scried out as a CSV for external analysis.


pkova:
Not quite, the crash is happening on vere 3.2 `edge` because 411 dill does not expect a `%born` event but 410 does. We'll fix it soon but it's a little tricky since the runtime shouldn't have to know the arvo kelvin it's running.

Note that the `%born` dill crash is harmless, although annoying.

Quodss:
my bad, didn't see the added task in a merged PR

Issue #6794, opened 2023-09-15
ship hangs on hearing about breaches


patosullivan:
I've noticed that my ship will often hang (and drop thousands of ames packets) for approximately 10 to 15 minutes every time it hears about a peer breaching.

@lukestiles and @philipcmonk have noticed the same.

I'm on the latest base and vere. Let me know if you need more info.

lukestiles:
See here https://pastebin.com/zia89XBY. A couple peers were breached.

jalehman:
@pkova @midden-fabler @yosoyubik Can one of you take a look into this when you can?

midden-fabler:
My dister ship on 413 `2q87f`, vere 2.11, was hanging on this before I killed it:
```
ames: breach peer [~mister-dister-midden-fabler ~macbec]
l2-sig-failed
l2-sig-failed
l2-sig-failed
ames: breach peer [~mister-dister-midden-fabler ~nibfeb]
ames: breach peer [~mister-dister-midden-fabler ~witbep-witbep]
ames: breach peer [~mister-dister-midden-fabler ~somlus-savlev]

interrupt
```

yosoyubik:
I've tested if there was an issue with the specific logs/batches so I started from an older snapshot and went over ~two weeks of recently submitted azimuth events (containing the breaches that were involved in this issue) and I could not reproduce the hanging on the test ship. 

I think this points out that the issue is happening on how ames and gall deal with breaches—iterating over all ames flows and all subscriptions. These can certainly happen if the ship that's breaching is part of the "stale ames-flows" issue (e.g. a publisher ship to whom we have subscribed several times but didn't clean those flows properly) but I can't imagine that taking 10-15 minutes since that would entail a large number of flows (I've seen people with around ~50.000 but that long delay in iterating over the forward-flows map would point out to several orders of magnitude bigger I think).

In the case of gall that would be different since old re-subscriptions should have been removed—unless there's something wrong in that case which is worth checking.

Next: reproducing this by breaching a ship and inspecting several of its peers for how they handle it, with sigpams along the way to granularly identify where/if the hanging occurs.

yosoyubik:
Some more notes: 

[...]

PR #7063, opened 2024-08-25
gall: delete agent blocked moves on |nuke

pkova:
I was investigating #6794 by running a local copy of `~finned-palmer`, manually breaching by doing `|pass [%j %ruin [~dinleb-rambep ~ ~]]` and observing the ship hang for a very long time. When I interrupted the ship by doing ctrl+c, the problem became quite evident.

`~finned-palmer` was spending all its' time in `+mo-filter-queue`. Specifically, it was iterating over 450 000 blocked moves to an agent called `%push-notify` trying to find any moves from the ship that just breached. This is an agent from the long defunct `%escape`.

This PR enables us to recover from scenarios where userspace programs do crazy stuff like this. It does potentially break some invariants, but I believe this to be worth it because this is a manual intervention.

@belisarius222 @joemfb

### DIFF
--- a/pkg/arvo/sys/vane/gall.hoon
+++ b/pkg/arvo/sys/vane/gall.hoon
@@ -957,6 +957,10 @@
   ++  mo-nuke
     |=  [prov=path dap=dude]
     ^+  mo-core
+    =?  mo-core  (~(has by blocked.state) dap)
+      =/  num-movs=@  ~(wyt by (~(got by blocked.state) dap))
+      ~>  %slog.0^leaf/"gall: deleted {<num-movs>} blocked moves for {<dap>}"
+      mo-core(blocked.state (~(del by blocked.state) dap))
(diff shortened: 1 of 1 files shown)


patosullivan:
Spoke too soon, the long hangs after breaches are still occurring w/this fix.

pkova:
You need to `|nuke %push-notify` to fix it after applying the code.

patosullivan:
Alright, nuked it, will report back if I see the hanging again.

patosullivan:
Ok, my ship just processed 15 breaches and it was only unresponsive for about ~4 minutes, normally that would be 15 * 15 minutes. I'm guessing it was just unresponsive due to the large number being processed back-to-back?

patosullivan:
Could we also preemptively `|nuke %push-notify` so the effect ships don't all need to be told to do so?

joemfb (approved):
this is good

PR #7057, opened 2024-08-13
Add support for Urbit time intervals less than one second.

sigilante:
1. Add conversion arms to `/sys/hoon` for milliseconds, microseconds, and nanoseconds.

```
::  +msec
::
::  Convert from a value in milliseconds to Urbit relative time, @dr.
::  Examples:
::    > `@ux`(msec 1)
::    0x41.8937.4bc6.a7ef
::    > `@dr`(msec 1)
::    ~s0..0041.8937.4bc6.a7ef
```

2. Add parsing syntax support for intervals in powers of 2¯¹⁰ s.

```
~ms1
```

PR #7067, opened 2024-09-10
Add subsecond parser support.

sigilante:
Split from #7057.


sigilante:
This reduplicates #7057 and adds parser support for subsecond timing intervals in both decimal and binary forms.

```
> (scan "ais1024" crub:so)
[p=~.dr q=16.384]

> `@dr`16.384
~s0..0000.0000.0000.4000

> ~0..ais1024
~s0..0000.0000.0000.4000
```

Barring discussion and tests passing, it is ready to merge.

PR #7074, opened 2024-09-25
dns: improve automatic dns configuration

tinnus-napbus:
This PR moves the `-dns-address` machinery into hood/helm and adds a `|dns-config` %ask generator to set it up. The changes include:

- Eyre
  - add a `/~/ip` endpoint that reflects the requester's IP in the same manner as icanhazip.com so public ip addresses can be discovered automatically without depending on services outside the network
  - adds a `/=domains=` scry endpoint to get the set of currently configured turfs
  - adds a `/=ports=` scry endpoint to get the [insecure=@ud secure=(unit @ud)]` ports configured
  - changes the `%turf` case of the `http-rule:eyre` to include a `%new` case along with the existing `%put` and `%del`. The `%new` case replaces all existing turfs and triggers a new certificate fetch even if nothing changed.

- Acme: remove all self-check steps.

Acme's self-check features created more problems than they solved because:
- If your router didn't support NAT hairpinning or you had some other unusual configuration the self-checks would fail despite your ship actually being accessible on port 80
- If your certificate expired cos your ship had been offline for a while or for some other reason, it couldn't be renewed because the self-checks would fail because requests through Iris simply fail on certificate errors. This wouldn't actually be a problem for the letsencrypt bot cos it just ignores expired/invalid certs and follows redirects.

Instead, I've moved the self-checks into helm & the generator with the option to skip them if you have an unusual setup and know what you're doing. The generator also lets you continue if you're not running on port 80 in case you're behind a load balancer or reverse proxy or whatever. It warns you and forces you to confirm you want to continue though so it should be pretty idiot-proof.

The generator lets you either enter an IP address or else it will try automatically discover it by making a request to `https://zod.urbit.org/~/ip`.

The generator also gives you the option to reset Eyre's domain configuration (using the `%new` `%turf` http-rule if enabled). This solves the problem where Eyre doesn't tell Acme to re-setup the certificate if the turf was already in Eyre. This behaviour had created problems for people trying to troubleshoot certificate problems.

The thread that Helm runs includes much better retry login on the self-checks, trying the domain every 30s for up to 15 minutes. This solves the problem of it failing the first few times simply because the new DNS entry took a few minutes to propagate.

Helm and the generator have very verbose error messages with suggestions of what to do that should cover the vast majority of problems people will encounter. The generator also has a preamble with extra info. This should make it much easier for noobs.


sigilante (approved):

pkova:
You need to base the eyre migration on stuff that's on `next/kelvin/410`, these changes conflict with the double boot protection stuff. Since it's unreleased you can combine the migrations if you want.

tinnus-napbus:
@pkova done

PR #7079, opened 2024-10-16
arvo: scry on vane %load

yosoyubik:
Previously during %load, all vane gates where called with a bunted sample which meant that scrying was not possible. Here we pass the +peek:le arm to the |what update engine—together with .now—that allows reads from the entire namespace.

This work is motivated by the removal of the larval core in ames.hoon in which scries were issued against other vanes (e.g. to retrieve public/private keys from %jael).

Another function of the larval core in %ames was to emit moves for setting up timers or sending gifts to vere. With no larval core, this moves would need to be emitted on %load—this is still not possible and would require a bigger change to arvo.hoon.

@joemfb


pkova:
We decided out of band to go with a modified larval state for the Directed Messaging ames larval stage, so we don't need this feature for now.

PR #7082, opened 2024-11-04
ames: expire routes on a timer

yosoyubik:
Consider the following:

- we achieve a direct route, on the last live flow.
- there are no other live flows, so nothing times out—no /pump timers fire
- we don't trigger any %fine requests, so those don't time out either.
- the only lane in state for that peer never gets modified—that only happens on /pump and /fine %wake timers
- a /forward-lane scry will always produce that direct route, unless sender/receiver ships initiate communication—the route gets also updated if you hear a packet on a different lane than what you have in state.

To fix that, we update the /forward-lane scry to explicitly check if the route is dead, based on the last-contact timestamp, instead of relying on the last value saved in state;

The ames driver has cached the dead lanes, and those lanes only expire on %nail gifts from arvo, but if no party sends a formal packet (e.g. the way to fix eauth communication problems has been to explicitly send a |hi) the lanes will remain there, and will send packets to nowhere.

To expire those lanes, we add a global ~m2 timers that goes over every peer and updates the lane to %dead if it has been more than ~s30 since last contact, and if so, emit a %nail gift to vere that will remove the entry from the cache—this timer will be removed in the Directed Messaging release that explicitly has a global ~m2 timer.

PS: credit to @pkova for finding this out.

PPS: draft until we get some feedback on the solution, and the migration is tested on a live ship.

### DIFF
--- a/bin/solid.pill
+++ b/bin/solid.pill
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:7c50e050b33493d51485c8a8d3b97f24777c420c6675a75a68b753bfbf014c50
-size 7077179
+oid sha256:fc30bad0f7f1d215c995616e30debe551d954612a9b3d24a568c2aa3e7a66e66
+size 7080666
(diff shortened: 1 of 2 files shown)


pkova:
The test failures here are because ames is now not generating the `x is your neighbor` %dill prints. `+update-qos` used to do this so we probably want to preserve the behavior.

pkova (approved):

Issue #7072, opened 2024-09-17
zuse: jet mismatch for X25519 Diffie-Hellman key exchange


newxiphiness:
shar:ed has a jet mismatch.
```hoon
> =my-shar =,  ed:crypto  |=  [pub=@ sek=@]
        ^-  @ux
        =+  exp=(shal (rsh [0 3] b) (suck sek))
        =.  exp  (dis exp (can 0 ~[[3 0] [251 (fil 0 251 1)]]))
        =.  exp  (con exp (lsh [3 31] 0b100.0000))
        =+  prv=(end 8 exp)
        =+  crv=(fra.fq (sum.fq 1 pub) (dif.fq 1 pub))
        (curt prv crv)
> (my-shar (shax 'foo') (shax 'bar'))
0x4fb0.74a6.7aba.56cb.f1be.1e2a.98a0.d5e5.fb8a.009e.1cf8.26f4.d640.b6a4.948c.a7bd
> (shar:ed:crypto (shax 'foo') (shax 'bar'))
0x7d79.c915.9a14.5af7.70fd.7ea9.4352.34fb.f30c.f65b.43b0.3e01.60fd.f8ab.5c39.1476
```
The issue is with the calculation of `crv`. Calculating the montgomery point's X value is "montgomeryX = (edwardsY + 1)*inverse(1 - edwardsY) mod p"  as per https://github.com/orlp/ed25519/blob/master/src/key_exchange.c#L30C29-L30C86). So, we should calculate montgomeryX only with the `y` value of the public key point. Here's a fixed version
```hoon
> =my-shar =,  ed:crypto  |=  [pub=@ sek=@]
        ^-  @ux
        =+  exp=(shal (rsh [0 3] b) (suck sek))
        =.  exp  (dis exp (can 0 ~[[3 0] [251 (fil 0 251 1)]]))
        =.  exp  (con exp (lsh [3 31] 0b100.0000))
        =+  prv=(end 8 exp)
        =.  pub  +:(need (deco pub))
        =+  crv=(fra.fq (sum.fq 1 pub) (dif.fq 1 pub))
        (curt prv crv)
> (my-shar (shax 'foo') (shax 'bar'))
0x7d79.c915.9a14.5af7.70fd.7ea9.4352.34fb.f30c.f65b.43b0.3e01.60fd.f8ab.5c39.1476
```
[...]

PR #7094, opened 2024-12-02
fix shar:ed jet mismatch, add luck:ed, scad:ed, scap:ed, scas:ed, sign-raw:ed, sign-octs-raw:ed

newxiphiness:
- luck:ed and sign-(octs-)raw go together
  - luck:ed generates a public private keypair where the private keypair is generated from SHA512(seed) and the usual EdDSA clamping procedure is done on the first 32 bytes.
  - sign-(octs-)raw allows one to sign from a keypair generated with luck (or, from sca(d/s))

- scad, scas, scap allows one to 'tweak' a public/private keypair with a scalar
  - scad tweaks public and private
  - scas tweaks private
  - scap tweaks public

this also fixes the jet mismatch mentioned in #7072, and potentially other jet mismatches due to sign:ed taking the bit rather than the byte length, whereas the jet uses u3r_bytes_fit to obtain the length to sign (the urcrypt library, moreover, doesn't allow signing at bit-level lengths). it also removes code deduplication generating pub/priv keys via use of luck:ed, and ensures that various length calculuations are derived from the parameters of the EdDSA curve.

a further change is recommended, as our curve parameters are merely in the subject, as opposed to arms, so theoretically, one could obtain a jet mismatch via manipulating the parameters which are hard coded to Ed25519 to some other curve, however that is out of the scope of this PR

### DIFF
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -1090,7 +1090,7 @@
       ::                                                ::  ++deco:ed:crypto
       ++  deco                                          ::  decode point
         |=  s=@  ^-  (unit [@ @])
-        ?>  (lte (met 3 s) cb)
+        ?.  (lte (met 3 s) cb)  ~
         =+  y=(cut 0 [0 (dec b)] s)
         =+  si=(cut 0 [(dec b) 1] s)
         =+  x=(xrec y)
--- a/pkg/arvo/sys/zuse.hoon
+++ b/pkg/arvo/sys/zuse.hoon
@@ -1020,6 +1020,7 @@
   ++  ed
     =>
       =+  =+  [b=256 q=(sub (bex 255) 19)]
+          =+  cb=(rsh [0 3] b)
           =+  fq=~(. fo q)
           =+  ^=  l
                %+  add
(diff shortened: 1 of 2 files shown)


newxiphiness:
i would say, now, this is good to go, i've tested this test suite with the ice flag enabled (or rather, disabled... c3n activates the testing) in the corresponding vere PR. excepting the fact that i haven't added shar:ed to the test file. i haven't managed to find a test vector in an RFC for ed25519 -> curve25519 -> DH

newxiphiness:
related PRs
https://github.com/urbit/vere/pull/748
https://github.com/urbit/urbit/pull/7094
https://github.com/urbit/urcrypt/pull/13

newxiphiness:
on second thought: for scalarmul(-base), we should assure the last byte is less than 127, and rather than punt on the failure of urcrypt library. i’d have to think more about it but one could perhaps imagine a dos vector were the ring signature library used (which is the only place where these jets are ran in injected code) resulting in the very slow +scam:ed being used. previously, this check was not performed in the urcrypt library, resulting in a key mismatch and erroneous behavior beyond the function domain the urcrypt method is well defined under.

will fix as soon as i get a chance and in vere PR, so don’t merge quite yet

newxiphiness:
Okay, just re-ran with ice mode on all the jets with the updated test i've pushed. it should all be good to go, along with the other related PR's.

joemfb (approved):

